PR #18224, opened 2020-10-09
perlapi: Document SvPV_free

khwilliamson:


Leont:
I totally have used this for CPAN code

khwilliamson:
Changed to perlapi

PR #18225, opened 2020-10-09
don't check for Windows NT in tests

xenu:
We dropped support for Win9x a long time ago.

### DIFF
--- a/t/io/fs.t
+++ b/t/io/fs.t
@@ -42,14 +42,12 @@ my $accurate_timestamps =
 	  $wd =~ m#$Config{afsroot}/#
      );
 
-if (defined &Win32::IsWinNT && Win32::IsWinNT()) {
-    if (Win32::FsType() eq 'NTFS') {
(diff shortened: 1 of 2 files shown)


xenu:
On second thought, it will break minitest on win32 (even further).

t/io/fs.t
```diff
@@ -42,14 +42,12 @@ my $accurate_timestamps =
...
-        $accurate_timestamps = 1;
-    }
-    else {
-        $has_link            = 0;
-    }
+if ($^O eq 'MSWin32' && Win32::FsType() eq 'NTFS') {
+    $has_link            = 1;
+    $accurate_timestamps = 1;
+}
+else {
```
Leont, line 49: This else looks wrong to me. Iit will trigger for any non-Windows, right?

xenu: Indeed, one more reason to withdraw this PR :)

PR #18226, opened 2020-10-09
test for use Environment Files command instead of `add-path` command

shogo82148:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


shogo82148:
ooops, it's my mistake

PR #18227, opened 2020-10-10
remove deprecated GitHub Actions Command and some improvements

shogo82148:
This pull request fixes the following warning message.

> https://github.com/Perl/perl5/actions/runs/296812857
>
> The `add-path` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

And replace `actions/checkout@master` into `actions/checkout@v2`.
Because GitHub recommends use tag names rather than branch names.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -352,7 +352,8 @@ jobs:
           echo @"
           C:\strawberry\c\bin
           C:\strawberry\perl\site\bin
-          C:\strawberry\perl\bin" |
+          C:\strawberry\perl\bin
+          "@ |
             Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: Host perl -V
         run: perl -V
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -349,7 +349,10 @@ jobs:
           if (!(Test-Path "C:\strawberry\perl\bin")) {
             choco install strawberryperl
           }
-          echo "C:\strawberry\c\bin`nC:\strawberry\perl\site\bin`nC:\strawberry\perl\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          echo @"C:\strawberry\c\bin
+          C:\strawberry\perl\site\bin
+          C:\strawberry\perl\bin" |
+            Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: Host perl -V
         run: perl -V
       - name: gcc --version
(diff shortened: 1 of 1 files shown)


.github/workflows/testsuite.yml
```diff
@@ -297,11 +296,17 @@ jobs:
...
         with:
           fetch-depth: 10
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ${{ github.workspace }}\choco-cache
+          key: v2-msvc100
       - name: Set up MSVC100
         run: |
+          choco config set cacheLocation "${{ github.workspace }}\choco-cache"
```
atoomic, line 309: Nice improvement

.github/workflows/testsuite.yml
```diff
@@ -335,13 +340,16 @@ jobs:
 
     steps:
       - run: git config --global core.autocrlf false
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
```
atoomic, line 343: Indeed better to use v2 when we can

.github/workflows/testsuite.yml
```diff
@@ -335,13 +340,16 @@ jobs:
...
           fetch-depth: 10
       - name: Set up Perl build environment
         run: |
-          choco install strawberryperl
-          echo "##[add-path]C:\strawberry\c\bin;C:\strawberry\perl\site\bin;C:\strawberry\perl\bin"
+          # skip installing perl if it is already installed.
+          if (!(Test-Path "C:\strawberry\perl\bin")) {
+            choco install strawberryperl
+          }
+          echo "C:\strawberry\c\bin`nC:\strawberry\perl\site\bin`nC:\strawberry\perl\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
atoomic: I think there is an alternate syntax much easier to read. But this is correct

.github/workflows/testsuite.yml
```diff
@@ -373,7 +381,8 @@ jobs:
       CONTINUOUS_INTEGRATION: 1
 
     steps:
-      - run: git config --global core.autocrlf false
```
atoomic, line 376: This line was removed but then added to the very long line later. Wouldn't it be better to preserve it there?

atoomic (changes requested):

.github/workflows/testsuite.yml
shogo82148: Git on windows is not the same thing as git on cygwin.
This line calls the git on windows, and gives no effect git on cygwin.

atoomic (approved):
@shogo82148 this looks good to me, do you mind squashing all of this as a single commit before I can merge it?
thanks

atoomic:
@shogo82148 do not bother squashing, the option is directly available from GitHub.
You do not need to do it, it's now merged.
Thanks for the updated/improved workflow

shogo82148:
Your welcome!

PR #18228, opened 2020-10-10
Smoke me/jkeenan/version warnings 20201010

jkeenan:
https://metacpan.org/source/LEONT/version-0.9928/META.json indicates that the bugtracker for this distribution is rt.cpan.org -- but the repository is blead.  I'll try my luck here first.

If the tests for this distribution are run with warnings enabled, two files generate warnings.  Patches correct this.

Thank you very much.
Jim Keenan

### DIFF
--- a/cpan/version/t/09_list_util.t
+++ b/cpan/version/t/09_list_util.t
@@ -4,8 +4,8 @@
 #########################
 
 use strict;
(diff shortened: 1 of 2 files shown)


Leont:
This PR should go to the [cpan repo](https://github.com/Perl/version.pm/) first.

PR #18230, opened 2020-10-11
win32: use the same manifest for Visual C++ and gcc builds

xenu:
Before this commit, MSVC builds were using `/manifestdependency` flag
instead of embedding `perlexe.manifest`. That was causing issues like
GetVersionEx() returning the wrong version number on Windows 10.


xenu:
BTW, as explained by the comment above them, those `EMBED_EXE_MANI` etc., macros are needed only for Visual C++ 2005 and 2008.

I wasted a few hours on setting up VC++ 2008 env and learning about the role of manifests for those two special compilers. I could've used that time more productively. I really hope we will drop support for them soon.

PR #18221, opened 2020-10-07
Provide strict compliance for DynaLoader and XSLoader

toddr:
Add warnings to tests.

PR #18235, opened 2020-10-12
Fix enough code to get make test_prep working with -Dusedefaultstrict

toddr:
This change fixes enough code that it's possible to run make test_prep when perl is compiled with -Dusedefaultstrict.

There are 2 caveats to this:

 - Does not address XSLoader/DynaLoader already submitted in another PR (#18221)
 - Does not address cpan/Pod-Usage or  cpan/Text-Tabs which continue to be outstanding upstream.

### DIFF
--- a/lib/Getopt/Std.pm
+++ b/lib/Getopt/Std.pm
@@ -1,5 +1,8 @@
 package Getopt::Std;
-require 5.000;
+
+use strict;
+use warnings;
+
 require Exporter;
 
 =head1 NAME
@@ -79,9 +82,9 @@ and version_mess() with the switches string as an argument.
 
 =cut
 
-@ISA = qw(Exporter);
-@EXPORT = qw(getopt getopts);
-$VERSION = '1.12';
+our @ISA = qw(Exporter);
+our @EXPORT = qw(getopt getopts);
+our $VERSION = '1.13';
 # uncomment the next line to disable 1.03-backward compatibility paranoia
 # $STANDARD_HELP_VERSION = 1;
 
@@ -119,6 +122,7 @@ sub getopt (;$$) {
 	        $$hash{$first} = $rest;
 	    }
 	    else {
+            no strict 'refs';
 	        ${"opt_$first"} = $rest;
 	        push( @EXPORT, "\$opt_$first" );
 	    }
@@ -128,6 +132,7 @@ sub getopt (;$$) {
 	        $$hash{$first} = 1;
 	    }
 	    else {
+            no strict 'refs';
 	        ${"opt_$first"} = 1;
 	        push( @EXPORT, "\$opt_$first" );
 	    }
@@ -145,6 +150,7 @@ sub getopt (;$$) {
     }
 }
 
+our ($OUTPUT_HELP_VERSION, $STANDARD_HELP_VERSION);
 sub output_h () {
   return $OUTPUT_HELP_VERSION if defined $OUTPUT_HELP_VERSION;
   return \*STDOUT if $STANDARD_HELP_VERSION;
@@ -251,6 +257,7 @@ sub getopts ($;$) {
 		    $$hash{$first} = $rest;
 		}
 		else {
+            no strict 'refs';
 		    ${"opt_$first"} = $rest;
 		    push( @EXPORT, "\$opt_$first" );
 		}
@@ -260,6 +267,7 @@ sub getopts ($;$) {
 		    $$hash{$first} = 1;
 		}
 		else {
+            no strict 'refs';
 		    ${"opt_$first"} = 1;
 		    push( @EXPORT, "\$opt_$first" );
 		}
--- a/lib/Symbol.pm
+++ b/lib/Symbol.pm
@@ -1,5 +1,8 @@
 package Symbol;
 
+use strict;
+use warnings;
+
 =head1 NAME
 
 Symbol - manipulate Perl symbols and their names
@@ -78,14 +81,12 @@ you reload the C<Foo> module afterwards.
 
 =cut
 
-BEGIN { require 5.005; }
-
 require Exporter;
-@ISA = qw(Exporter);
-@EXPORT = qw(gensym ungensym qualify qualify_to_ref);
-@EXPORT_OK = qw(delete_package geniosym);
+our @ISA = qw(Exporter);
+our @EXPORT = qw(gensym ungensym qualify qualify_to_ref);
+our @EXPORT_OK = qw(delete_package geniosym);
 
-$VERSION = '1.08';
+our $VERSION = '1.09';
 
 my $genpkg = "Symbol::";
 my $genseq = 0;
@@ -99,6 +100,7 @@ my %global = map {$_ => 1} qw(ARGV ARGVOUT ENV INC SIG STDERR STDIN STDOUT);
 #
 sub gensym () {
     my $name = "GEN" . $genseq++;
+    no strict 'refs';
     my $ref = \*{$genpkg . $name};
     delete $$genpkg{$name};
     $ref;
@@ -132,6 +134,7 @@ sub qualify ($;$) {
 }
 
 sub qualify_to_ref ($;$) {
+    no strict 'refs';
     return \*{ qualify $_[0], @_ > 1 ? $_[1] : caller };
 }
 
@@ -150,6 +153,7 @@ sub delete_package ($) {
     }
 
     my($stem, $leaf) = $pkg =~ m/(.*::)(\w+::)$/;
+    no strict 'refs';
     my $stem_symtab = *{$stem}{HASH};
     return unless defined $stem_symtab and exists $stem_symtab->{$leaf};
 
@@ -160,6 +164,7 @@ sub delete_package ($) {
     foreach my $name (keys %$leaf_symtab) {
         undef *{$pkg . $name};
     }
+    use strict 'refs';
 
     # delete the symbol table
 
--- a/lib/Symbol.t
+++ b/lib/Symbol.t
@@ -1,5 +1,8 @@
 #!./perl
 
+use strict;
+use warnings;
+
 BEGIN {
     chdir 't' if -d 't';
     @INC = '../lib';
@@ -87,6 +90,7 @@ use Symbol qw(qualify qualify_to_ref);  # must import into this package too
 
 # tests for delete_package
 package main;
+no warnings 'once';
 $Transient::variable = 42;
 ok( exists $::{'Transient::'}, 'transient stash exists' );
 ok( defined $Transient::{variable}, 'transient variable in stash' );
--- a/lib/locale.pm
+++ b/lib/locale.pm
@@ -1,6 +1,9 @@
 package locale;
 
-our $VERSION = '1.09';
+use strict;
+use warnings;
+
+our $VERSION = '1.10';
 use Config;
 
 $Carp::Internal{ (__PACKAGE__) } = 1;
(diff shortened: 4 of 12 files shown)


lib/utf8.pm
```diff
@@ -1,15 +1,19 @@
 package utf8;
 
-$utf8::hint_bits = 0x00800000;
```
toddr, line 3: I'm aware this might have unexpected side effects but I wasn't able to find any. I'm happy to revert this if there's a good reason for not allowing strict/warnings here. Is this a circular deps issue maybe?

lib/overload.pm
```diff
@@ -1239,7 +1243,7 @@ Put this in F<symbolic.pm> in your Perl library directory:
 
 This module is very unusual as overloaded modules go: it does not
 provide any usual overloaded operators, instead it provides an
-implementation for C<L</nomethod>>.  In this example the C<nomethod>
+implementation for L</C<nomethod>>.  In this example the C<nomethod>
```
khwilliamson: Why this change?  It's my experience that the original method looks better for internal links; the changed-to form for external links.  Sadly.

lib/utf8.pm
khwilliamson: I don't know if this has side effects or not.

BTW I hate that we don't derive the bit number.

khwilliamson: So, apply it and see what breaks on cpan, hopefully nothing

lib/overload.pm
```diff
@@ -1,8 +1,12 @@
 package overload;
 
-our $VERSION = '1.32';
+use strict;
+no strict 'vars'; # Needed for assignment to typeglobs and symbol table
```
Grinnz: This comment is not accurate; assigning to typeglobs and symbol table entries works fine under strict 'vars'.

lib/overload.pm
toddr: That's an oversight. I have no opinion on POD style personally. I will remove it.

lib/overload.pm
toddr: What would you say instead?

lib/utf8.pm
khwilliamson: I can't remember, but isn't this global the way it tells the core that 'use utf8' is in effect?  If so, this would break lots.

Grinnz: I don't know about possibility of circular reference, but applying strict and warnings to this module could be a significant performance hit due to how much it is used. IMO making the other changes and leaving out strict and warnings would be preferable in this specific case. By the way, bytes.pm is a "mirror" of this module so should be updated in the same way.

Grinnz: @khwilliamson This global is still being set a few lines below.

lib/overload.pm
Grinnz: That depends why it's needed here. Strict 'vars' requires you to declare a (scalar, array, or hash) variable with `my` or `our` before using it.

lib/utf8.pm
toddr: I hear what you're saying. I hesitated for similar reasons in all honesty. 

 - require is almost no cost if the module is already loaded. This would only happen the first time utf8.pm is loaded anyway.
 - strict/warnings->import would also only happen once. Why would that be a performance hit?

lib/overload.pm
toddr: yep. nevermind. fixed.

lib/utf8.pm
Grinnz: I'm not talking about the performance hit of activating strict/warnings but of having them active each time this module is imported or its AUTOLOAD is called. Though there are certainly bigger concerns at that point. I certainly don't think it will make a big difference but just questioning the tradeoff for this module in particular, maybe that's why they weren't set before, maybe not.

ext/XS-APItest/APItest_BS
```diff
@@ -1,6 +1,6 @@
 #
 # test that non-empty .bs files get executed (but only once)
-
+no strict 'vars';
```
Grinnz: This doesn't seem like the right thing to do, shouldn't we be testing that these files continue to be executed in a non-strict context?

toddr:
> So, apply it and see what breaks on cpan, hopefully nothing

I was horrified to find out how many inc/utf8.pm files there are on CPAN. Ignoring those leaves us with this:

```
$>git grep utf8::hint_bits|grep -v inc/utf8.pm
A/Acme-Emoticarp/lib/Acme/Emoticarp.pm:    $^H |= $utf8::hint_bits; ## no critic (Bangs::ProhibitBitwiseOperators)
E/Encode/encoding.pm:        require utf8;      # to fetch $utf8::hint_bits;
E/Encode/encoding.pm:        $^H |= $utf8::hint_bits;
F/Fennec/lib/Fennec/Util.pm:    my $with_utf8 = ( caller(0) )[8] & $utf8::hint_bits;
G/Game-Collisions/nytprof/utf8-pm-16-line.html:<tr><td class="h"><a name="3"></a>3</td><td class="c3">1</td><td class="c3"><span title="Avg 400ns">400ns</span></td><td></td><td></td><td class="s">$utf8::hint_bits = 0x00800000;</td></tr>
G/Game-Collisions/nytprof/utf8-pm-16-line.html:<tr><td class="h"><a name="8"></a>8</td><td class="c3">1</td><td class="c0"><span title="Avg 12µs">12µs</span></td><td></td><td></td><td class="s">    $^H |= $utf8::hint_bits;</td></tr>
G/Game-Collisions/nytprof/utf8-pm-16-line.html:<tr><td class="h"><a name="12"></a>12</td><td></td><td></td><td></td><td></td><td class="s">    $^H &amp;= ~$utf8::hint_bits;</td></tr>
b/bare/bare.pm:  my $legal_var = $utf8::hint_bits && $^H & $utf8::hint_bits
l/lambda/lib/lambda.pm:  $^H |= $utf8::hint_bits;
l/lambda/lib/lambda.pm:  $^H &= ~$utf8::hint_bits;
p/perl/cpan/Encode/encoding.pm:        require utf8;      # to fetch $utf8::hint_bits;
p/perl/cpan/Encode/encoding.pm:        $^H |= $utf8::hint_bits;
p/perl/lib/utf8.pm:$utf8::hint_bits = 0x00800000;
p/perl/lib/utf8.pm:    $^H |= $utf8::hint_bits;
p/perl/lib/utf8.pm:    $^H &= ~$utf8::hint_bits;
p/perl/pod/perldata.pod:          (?(?{ (caller(0))[8] & $utf8::hint_bits })
p/perl_mlb/os2/encoding.pm:	require utf8; # to fetch $utf8::hint_bits;
p/perl_mlb/os2/encoding.pm:	$^H |= $utf8::hint_bits;
p/perl_mlb/utf8.pm:$utf8::hint_bits = 0x00800000;
p/perl_mlb/utf8.pm:    $^H |= $utf8::hint_bits;
p/perl_mlb/utf8.pm:    $^H &= ~$utf8::hint_bits;
q/qbit/lib/qbit.pm:    $^H |= $utf8::hint_bits;

```

lib/Getopt/Std.pm
```diff
@@ -119,6 +122,7 @@ sub getopt (;$$) {
 	        $$hash{$first} = $rest;
 	    }
 	    else {
+            no strict 'refs';
```
Grinnz, line 125: Indentation doesn't quite match the following lines.

toddr: Welcome to the already existing world of Mixed tabs/spaces. Can I put you down as a yes vote for doing an eventual mass perltidy of blead? 😛

Grinnz:
> > So, apply it and see what breaks on cpan, hopefully nothing
> 
> I was horrified to find out how many inc/utf8.pm files there are on CPAN.

This shouldn't matter, `$utf8::hint_bits` is still being set with these changes

toddr:
> > > So, apply it and see what breaks on cpan, hopefully nothing
> > 
> > 
> > I was horrified to find out how many inc/utf8.pm files there are on CPAN.
> 
> This shouldn't matter, `$utf8::hint_bits` is still being set with these changes

Yep. I'm just documenting what I found. Agreed. Looks fine to me.

lib/Getopt/Std.pm
Grinnz: I wouldn't go as far as mass perltidy, but making indentation consistent as an isolated commit would be useful IMO

ext/XS-APItest/APItest_BS
toddr: The goal for this commit is to have a perl we can build with strict on everywhere by default. That means this code snippet starts with strict on and has to have it disabled. DynaLoader cannot influence the strictness of this section unless we suck in the BS and then do a quoted eval. but that's a topic for another pull request.

See also https://github.com/Perl/perl5/pull/18221#issuecomment-706311030

toddr: I've removed this file from this commit and am moving BS changes over to #18221

lib/Getopt/Std.pm
```diff
@@ -251,6 +257,7 @@ sub getopts ($;$) {
 		    $$hash{$first} = $rest;
 		}
 		else {
+            no strict 'refs';
```
atoomic, line 260: indentation looks weird, not sure if this is the original or your patch, but they are not matching (same with the next line)

lib/Symbol.pm
```diff
@@ -132,6 +134,7 @@ sub qualify ($;$) {
 }
 
 sub qualify_to_ref ($;$) {
```
atoomic, line 136: maybe we could consider to use function signature instead of prototype? (as part of a next PR)

lib/Symbol.t
```diff
@@ -87,6 +90,7 @@ use Symbol qw(qualify qualify_to_ref);  # must import into this package too
 
 # tests for delete_package
 package main;
+no warnings 'once';
 $Transient::variable = 42;
```
atoomic, line 94: rather than disabling that warning in the global scope either
`$Transient::variable = $Transient::variable = 42;`
or 
`{ no warnings 'once'; $Transient::variable = 42; }`

lib/locale.pm
```diff
@@ -70,7 +73,7 @@ sub import {
                               :numeric :monetary :time) );
         for (my $i = 0; $i < @_; $i++) {
```
atoomic, line 74: not related to this change but maybe consider something like this `foreach my $arg ( @_ )`

lib/overload.pm
```diff
@@ -26,7 +29,7 @@ my %ops_seen;
 sub nil {}
 
 sub OVERLOAD {
-  $package = shift;
+  my $package = shift;
   my %arg = @_;
```
atoomic, line 33: `my ( $package, %args ) = @_;`

lib/utf8.pm
```diff
@@ -1,15 +1,19 @@
...
+use strict;
+use warnings;
 
-our $VERSION = '1.23';
+our $hint_bits = 0x00800000;
```
atoomic, line 6: this is `HINT_UTF8` at minimum we probably want a test to confirm it's in sync with `perl.h`

lib/Symbol.t
toddr: The test is complicated enough that I decided a more directed follow on fix would be a better choice for this.

lib/utf8.pm
toddr: Right. This sounds like a separate case.

atoomic (approved):

lib/Symbol.pm
Grinnz: That's nonsense. Signatures and prototypes are unrelated features.

PR #18236, opened 2020-10-13
Fix t/test.pl and t/TEST to be strict compliant.

toddr:
So tests can run with -Dusedefaultstrict

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -20,7 +20,7 @@
 # will be worked over by t/op/inc.t
 
 $| = 1;
-$Level = 1;
+our $Level = 1;
 my $test = 1;
 my $planned;
 my $noplan;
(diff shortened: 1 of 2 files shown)


t/test.pl
```diff
@@ -1023,7 +1025,7 @@ sub _fresh_perl {
 
     # Use the first line of the program as a name if none was given
     unless( $name ) {
-        ($first_line, $name) = $prog =~ /^((.{1,50}).*)/;
+        my ($first_line, $name) = $prog =~ /^((.{1,50}).*)/;
```
Grinnz: This doesn't look right, it shadows the outer $name instead of updating it.

Grinnz: Probably should be `(my $first_line, $name) = ...`

t/test.pl
```diff
@@ -1435,7 +1438,7 @@ ($@)
 sub new_ok {
     my($class, $args, $obj_name) = @_;
     $args ||= [];
-    $object_name = "The object" unless defined $obj_name;
+    my $object_name = "The object" unless defined $obj_name;
```
Grinnz: Conditional my declaration, will lead to bugs.

t/test.pl
toddr: yep.

t/test.pl
toddr: actually this looks like a bug on closer inspection. `s/object_name/obj_name/g`

Grinnz: I agree, it doesn't really make sense as written, this and the below usage likely should be using `$obj_name`.

toddr: Fixed.

PR #18241, opened 2020-10-14
Fix unicode related modules for strict compliance

toddr:


lib/Unicode/UCD.t
```diff
@@ -1,10 +1,17 @@
...
+
+my $IS_ASCII = (ord("A") == 65) ? 1 : 0;
+my $IS_EBCDIC = (ord("A") == 193) ? 1 : 0;
+
 BEGIN {
-    $::IS_ASCII = (ord("A") == 65) ? 1 : 0;
-    $::IS_EBCDIC = (ord("A") == 193) ? 1 : 0;
     chdir 't' if -d 't';
     @INC = '../lib';
-    require Config; import Config;
```
Grinnz, line 7: Doesn't this prevent Config from getting loaded using the new `@INC` which seems to be the point of this block?

toddr: This boilerplate is all over the place. It's also inconsistent. At this point if strict/warnings loaded, Config will load just fine.

Grinnz: I don't see what this has to do with strict compliance either way. If you want to load Config before this block in cases like this, it should be a coordinated change that also deals with the now pointless change to `@INC` and whatever it was trying to accomplish before.

khwilliamson (approved):

lib/Unicode/UCD.t
karenetheridge: Agreed, let's change just one thing at a time (and that may mean changing other things first to make adding strict/warnings more smooth).

toddr: Config is being moved up because it doesn't compile the way it is written.

```
$>perl -e 'use strict; BEGIN {    require Config;  import Config; print "ok\n" if $Config{'extensions'} }'
Global symbol "%Config" requires explicit package name (did you forget to declare "my %Config"?) at -e line 1.
BEGIN not safe after errors--compilation aborted at -e line 1.

$>perl -e 'use strict; use Config; BEGIN { print "ok\n" if $Config{'extensions'} }'        
ok
```

Grinnz: That's a compile time `strict vars` error - by the time those statements are reached in their runtime, %Config will have been imported. You can work around it by declaring `our %Config` which will refer to the same package variable once imported.

toddr: Or just move the use statement up 😛

Grinnz: That changes behavior, which requires its own justification.

toddr: Is the thing this test is testing affected in any way by when `%Config` is loaded? This concern feels very minor to me. I'm not understanding why we're worrying about it.

Grinnz: I'm not understanding why you are changing this behavior without necessity or relation to the purpose of this PR.

toddr: The commit message is:

>  Fix unicode related modules for strict compliance

Adding strict broke how Config was being used. I fixed that by loading Config earlier so it would work correctly. This change is specifically to address the purpose of the PR.

Grinnz: Fixing strict compliance would involve satisfying strict 'vars'. You have instead made an unrelated change that obviated the issue and requires its own separate justification.

Grinnz: Following up with how @karenetheridge put it, at some point in the future someone may look back and wonder why this was changed around in this unrelated commit. Making changes with specific justifications is very important for historical understanding and bisecting issues.

xsawyerx: I'm sorry, but I disagree here. The change to importing `Config` was done as part of this PR to satisfy this PR. It makes it **very** relevant to it. You may request it done in another commit (which the author of the PR is **not required** to oblige us), but putting it in another PR is entirely unnecessary.

xsawyerx (approved):

xsawyerx:
With @khwilliamson approving it, I consider it ready for merge.

lib/Unicode/UCD.t
Grinnz: It was the incorrect change to satisfy this PR, I have already pointed out the correct action: declaring `our %Config`. I do not think it is appropriate to justify random changes with unexplored side effects just because they also fix the problem.

xsawyerx: Calling this a "random change" is incorrect. Can you detail what unexplored side effect this can have? Calling `our %Config` ignores the direct behavior of loading `Config.pm` which this change maintains.

Grinnz: It does not. It aliases the lexical `%Config` in that scope to `%main::Config`, which after the import at runtime, will be the correct variable.

I'm not sure what you want me to detail; the current change changes when Config is loaded relative to the change to `@INC` and no explanation has been provided for this.

hvds: FWIW, I'd also prefer if this PR made the minimal change @Grinnz proposes. I think we should look separately at how we want to ensure our test files load things from the correct path - the recent 8c892c9 shows how badly it had got out of sync.

It doesn't seem wise to me to make a change in just one test file to the established process of sorting out `@INC` before loading certain things such as Config when that isn't required to satisfy the needs of this PR.

toddr: I'm totally up for sorting out the `@INC` problem first. I'm going to close this PR for the time being.

Grinnz: This discussion has been very confusing to me, so in the hopes that it will assist people in understanding my proposed change in order to make appropriate changes in similar cases:

Importing from Config has the net result of aliasing `%main::Config` (or `%Config` in the current package) to `%Config::Config`, as well as allowing it to subsequently be referenced without fully qualifying under strict 'vars'.

Applying strict 'vars' has caused this parse time error: "Global symbol "%Config" requires explicit package name" because at *parse* time (of the BEGIN block) Config has not yet been imported from, so the strict 'vars' exception for this variable has not yet been applied.

Declaring `our %Config` assures the *parser* that references to `%Config` in that lexical scope are referring to `%main::Config`, which satisfies strict 'vars'. Subsequently at runtime (of the BEGIN block) the variable is aliased by the import before it is accessed so there is no change in behavior.

Issue #14687, opened 2015-05-02
Function lstat behavior case differs between Windows and Unix


p5pRT:
Migrated from [rt.perl.org#124443](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=124443) (status was 'open')

Searchable as RT124443$

p5pRT:
### From rich@richelberger.com

#### Created by rich@richelberger.com



Please refer to RT 90452 for background and original symptom\. \#NYCHackathon

In regular shells\, the path expression ‘foo/…’ should fail and return an empty list\.

\<code>
use strict; use warnings; use 5\.10\.00;
mkdir a;
my @&#8203;result = lstat ‘a/…’;
say for @&#8203;result;
\</code>

In Linux / Mac\, it results in an empty array since it is file not found\.
On Windows\, the result is&#8203;:

$VAR1 = 0;
$VAR2 = 0;
$VAR3 = 16832;
$VAR4 = 0;
$VAR5 = 0;
$VAR6 = 0;
$VAR7 = 0;
$VAR8 = 0;
$VAR9 = 0;
$VAR10 = 0;
$VAR11 = 0;
$VAR12 = '';
$VAR13 = '';

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=medium

Site configuration information for perl 5.20.2:

Configured by rich at Tue Feb 24 17:10:19 EST 2015.

Summary of my perl5 (revision 5 version 20 subversion 2) configuration:
   
  Platform:
    osname=darwin, osvers=14.1.0, archname=darwin-2level
    uname='darwin ananke 14.1.0 darwin kernel version 14.1.0: mon dec 22 23:10:38 pst 2014; root:xnu-2782.10.72~2release_x86_64 x86_64 '
    config_args='-de -Dprefix=/Users/rich/perl5/perlbrew/perls/perl-5.20.1 -Aeval:scriptdir=/Users/rich/perl5/perlbrew/perls/perl-5.20.1/bin'
    hint=recommended, useposix=true, d_sigaction=define
    useithreads=undef, usemultiplicity=undef
    use64bitint=define, use64bitall=define, uselongdouble=undef
    usemymalloc=n, bincompat5005=undef
  Compiler:
    cc='cc', ccflags ='-fno-common -DPERL_DARWIN -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include',
    optimize='-O3',
[...]

Issue #12431, opened 2012-09-21
Win32: -e '"' always returns true


p5pRT:
Migrated from [rt.perl.org#114986](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=114986) (status was 'open')

Searchable as RT114986$

p5pRT:
### From @sisyphus

Subject&#8203;: Win32&#8203;: \-e '"' always returns true
Message\-Id&#8203;: \<5\.16\.0\_1816\_1348202385@&#8203;desktop2>
Reply\-To&#8203;: sisyphus1@&#8203;optusnet\.com\.au
To&#8203;: perlbug@&#8203;perl\.org
From&#8203;: sisyphus1@&#8203;optusnet\.com\.au


This is a bug report for perl from sisyphus1@&#8203;optusnet\.com\.au\,
generated with the help of perlbug 1\.39 running under perl 5\.16\.0\.

Hi\,

This \(raised recently at http&#8203;://www\.perlmonks\.org/index\.pl?node\_id=979494\) 
must surely have come up before ?

It's behaviour that's specific to MS Windows \- for all versions of perl back 
to 5\.6\.2 \(and presumably even further back\)\.

\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#\#

use warnings;
use strict;

print "1&#8203;:  huh?\\n" if \-e '"';
print "2&#8203;:  huh?\\n" if \-e '""';
print "3&#8203;:  huh?\\n" if \-e '"""';
print "4&#8203;:  huh?\\n" if \-e '""""';
print "5&#8203;:  huh?\\n" if \-e '"""""';
print "6&#8203;:  huh?\\n" if \-e '""""""""""""""';
print "7&#8203;:  huh?\\n" if defined \-f '"';
print "8&#8203;:  huh?\\n" if defined \-f '""';
print "9&#8203;:  huh?\\n" if defined \-f '"""';
print "10&#8203;: huh?\\n" if defined \-f '""""';
print "11&#8203;: huh?\\n" if defined \-f '"""""';
print "12&#8203;: huh?\\n" if defined \-f '""""""""""""""';

\_\_END\_\_
Outputs&#8203;:

1&#8203;:  huh?
2&#8203;:  huh?
3&#8203;:  huh?
4&#8203;:  huh?
5&#8203;:  huh?
6&#8203;:  huh?
7&#8203;:  huh?
[...]

Issue #9025, opened 2007-09-10
win32 - file test operators don't work for //?/UNC/server/file filenames


p5pRT:
Migrated from [rt.perl.org#45331](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=45331) (status was 'open')

Searchable as RT45331$

p5pRT:
### From adavies@ptc.com

#### Created by adavies@ptc.com



On win32 you can use UNC pathanmes of the form&#8203;:

\\\\?\\UNC\\server\\folder\\filename

Perl can C\<open> these files ok\, but C\<\-f>\, C\<\-e> etc\. 
incorrectly return false for these filenames\.
However\, they do work for directory names in this format\.

The code in question is in win32/win32\.c's win32\_stat\(\)&#8203;:

\# %\<

 \#if defined\(WIN64\) || defined\(USE\_LARGE\_FILES\)
&nbsp;    res = \_stati64\(path\, sbuf\);
 \#else
&nbsp;    res = stat\(path\, sbuf\);
 \#endif
&nbsp;    sbuf\->st\_nlink = nlink;

 if \(res \< 0\) \{
&nbsp;    /\* CRT is buggy on sharenames\, so make sure it really isn't\.
&nbsp;     \* XXX using GetFileAttributesEx\(\) will enable us to set
&nbsp;     \* sbuf\->st\_\*time \(but note that's not available on the
&nbsp;     \* Windows of 1995\) \*/
&nbsp;    DWORD r = GetFileAttributesA\(path\);
&nbsp;    if \(r \!= 0xffffffff && \(r & FILE\_ATTRIBUTE\_DIRECTORY\)\) \{
&nbsp;        /\* sbuf may still contain old garbage since stat\(\) failed \*/
&nbsp;        Zero\(sbuf\, 1\, Stat\_t\);
&nbsp;        sbuf\->st\_mode = S\_IFDIR | S\_IREAD;
&nbsp;        errno = 0;
&nbsp;        if \(\!\(r & FILE\_ATTRIBUTE\_READONLY\)\)
&nbsp;            sbuf\->st\_mode |= S\_IWRITE | S\_IEXEC;
&nbsp;        return 0;
&nbsp;    \}

\# >%


Note that GetFileAttributesA\(\) does succeed for "\\\\?\\UNC\\server\\file"
format filesnames\.\.\. but the result does not have the
"FILE\_ATTRIBUTE\_DIRECTORY" bit set\.

[...]

Issue #8502, opened 2006-06-27
filetest problem with STDIN/OUT on Windows


p5pRT:
Migrated from [rt.perl.org#39637](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=39637) (status was 'open')

Searchable as RT39637$

p5pRT:
### From @pmqs


\[Forgot to forward this to perlbug\]

Had a problem with my compression modules reported to me yesterday that
boiled down to the \-r & \-w file test operators not working properly with
STDIN/OUT on windows\.

> perl \-e " print \-r \\\*STDIN ? 'Can Read' : 'Cannot Read' "
Cannot Read

> perl \-e " print \-w \\\*STDOUT ? 'Can Write' : 'Cannot Write' "
Cannot Write

My windows box has an ActiveState 5\.8\.8 on it\.

Is this a known issue on windows?

Paul


&nbsp;        
\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_\_ 
The all\-new Yahoo\! Mail goes wherever you go \- free your email address from your Internet provider\. http&#8203;://uk\.docs\.yahoo\.com/nowyoucan\.html

p5pRT:
### From @steve-m-hay

paul\.marquess@&#8203;ntlworld\.com \(via RT\) wrote&#8203;:
> Had a problem with my compression modules reported to me yesterday that
> boiled down to the \-r & \-w file test operators not working properly with
> STDIN/OUT on windows\.
> 
>> perl \-e " print \-r \\\*STDIN ? 'Can Read' : 'Cannot Read' "
> Cannot Read
> 
>> perl \-e " print \-w \\\*STDOUT ? 'Can Write' : 'Cannot Write' "
> Cannot Write
> 
> My windows box has an ActiveState 5\.8\.8 on it\.
> 
> Is this a known issue on windows?

Not to me\, so thanks for the report\!

It isn't mentioned in perlport either\, but it seems to be a Windows thing\.

pp\_ftrread\(\) calls my\_stat\(\)\, which calls PerlLIO\_fstat\(\) on the 
appropriate file descriptor\.  This winds up in win32\_fstat\(\) on Windows\, 
[...]

PR #18242, opened 2020-10-15
Remove EOL Windows Versions From Supported List.

genio:
As discussed on the P5P mailing list here:
https://www.nntp.perl.org/group/perl.perl5.porters/2020/10/msg258453.html

This just removes the declaration that we support the very old versions of Windows that have long since been EOLed.

For reference, here's a short list of problems related to maintaining the EOLed Windows versions:
https://github.com/Perl/perl5/issues/4145
https://github.com/Perl/perl5/issues/6080
https://github.com/Perl/perl5/issues/7410
https://github.com/Perl/perl5/issues/8502
https://github.com/Perl/perl5/issues/9025
https://github.com/Perl/perl5/issues/12431
https://github.com/Perl/perl5/issues/14687

Note that this only removes the requirement that we must keep new versions of Perl fully functional on those old operating systems. It doesn't actually make them immediately non-functional.

### DIFF
--- a/pod/perlport.pod
+++ b/pod/perlport.pod
@@ -2251,17 +2251,15 @@ at L<http://www.cpan.org/src>
 
 =over
 
-=item Windows 2000
-
-=item Windows XP
-
-=item Windows Server 2003
-
-=item Windows Vista
-
-=item Windows Server 2008
+=item Windows Server 2008/R2
 
 =item Windows 7
+
+=item Windows Server 2012
+
+=item Windows 8
+
+=item Windows 10
 
 =back
 
@@ -2372,6 +2370,14 @@ have worked in 5.12, but supporting code has been removed for 5.14:
 =item Windows ME
 
 =item Windows NT4
+
+=item Windows 2000
+
+=item Windows XP
+
+=item Windows Server 2003
+
+=item Windows Vista
 
 =back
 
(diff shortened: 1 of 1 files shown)


pod/perlport.pod
```diff
@@ -2373,6 +2371,14 @@ have worked in 5.12, but supporting code has been removed for 5.14:
 
 =item Windows NT4
 
+=item Windows 2000
```
Grinnz: This section is organized by version of removal and so there should be a new section added instead.

pod/perlport.pod
```diff
@@ -2251,17 +2251,15 @@ at L<http://www.cpan.org/src>
 
 =over
 
-=item Windows 2000
```
Grinnz, line 2254: This section leads with "platforms known to build Perl 5.12." I'm not sure how it should be updated but it clearly should be.

genio: Gah! You're absolutely right. A new 5.34 section should be added?

wchristian:
The documentation itself needs an explanation as to why these decisions would have been made.

wchristian:
> Note that this only removes the requirement that we must keep new versions of Perl fully functional on those old operating systems. It doesn't actually make them immediately non-functional.

Where is this requirement documented?

wchristian:
I've opened another ticket to contain the various questions about OS support brought up by this issue:

- perlport is inconsistent and the policies regarding OS support unclear #18243

genio:
If someone wants to do this, they can steal from here or do a better job

Issue #18249, opened 2020-10-18
ext/POSIX/t/sysconf.t:  Unexpected "unopened filehandle" warning


jkeenan:
I am getting an unexpected warning when I run file `ext/POSIX/t/sysconf.t` in the core distribution with warnings turned on.

To illustrate the problem, consider the following file in which I have stripped away much of the code from `sysconf.t` to illustrate the current problem.  This file does not need to be run from within a git checkout to demonstrate the problem.  I've been able to reproduce the problem with perl-5.16 and later but not with perl-5.14.
```
$ cat second-sysconf.t 
#!perl -w

BEGIN {
    use Config;
    use Test::More;
    plan skip_all => "POSIX is unavailable" if $Config{'extensions'} !~ m!\bPOSIX\b!;
}

use strict;
use File::Spec;
use POSIX;

sub check(@) {
    grep { eval "&$_;1" or $@!~/vendor has not defined POSIX macro/ } @_
}       

my @path_consts_terminal = check qw(
    _PC_MAX_CANON
);

plan( tests =>  4 );

sub _check_and_report {
    my ($sub, $constant, $description) = @_;
    $! = 0;
    my $return_val = eval {$sub->(eval "$constant()")};
    my $errno = $!; # Grab this before anything else changes it.
    is($@, '', $description);

    # We can't test sysconf further without investigating the type of argument
    # provided
    return if $description =~ /sysconf/;

    if (defined $return_val) {
        like($return_val, qr/\A(?:-?[1-9][0-9]*|0 but true)\z/,
             'the returned value should be a signed integer');
    } else {
      SKIP:
        {
            # POSIX specifies EINVAL is returned if the f?pathconf()
            # isn't implemented for the specific path
            skip "$description not implemented for this path", 1
                if $errno == EINVAL && $description =~ /pathconf/;
            cmp_ok($errno, '==', 0, 'errno should be 0 as before the call')
                or diag("\$!: $errno");
        }
    }
}

my $TTY = "/dev/tty";
SKIP: {
    my $n = 2 * 2 * @path_consts_terminal;

    -c $TTY or skip("$TTY not a character file", $n);
    open(TTY, '<', $TTY) or skip("failed to open $TTY: $!", $n);
    -t TTY or skip("TTY ($TTY) not a terminal file", $n);

    my $fd = fileno(TTY);
    like($fd, qr/^\d+$/, "fileno() returned integer");

    # testing fpathconf() on a terminal file
[...]

PR #18250, opened 2020-10-18
sysconf.t: close() takes filehandle, not file descriptor

jkeenan:
Per suggestion from Leon Timmermans in
https://github.com/Perl/perl5/issues/18249.

### DIFF
--- a/ext/POSIX/t/sysconf.t
+++ b/ext/POSIX/t/sysconf.t
@@ -7,6 +7,7 @@ BEGIN {
 }
 
 use strict;
+use warnings;
 use File::Spec;
 use POSIX;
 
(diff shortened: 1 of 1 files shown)


ext/POSIX/t/sysconf.t
```diff
@@ -1,4 +1,4 @@
-#!perl -T
+#!perl -Tw
```
Leont: Any particular reason for using `-w` instead of `use warnings`?

jkeenan: > Any particular reason for using `-w` instead of `use warnings`?

Nothing other than the fact that we already have one switch (`-T`) on the shebang line of that file, so running with warnings only required one extra keystroke.

Grinnz: `-w` does not run with warnings, it imposes warnings on anything that does not express a preference. It should not be used in general.

jkeenan:
On 10/18/20 4:07 PM, Dan Book wrote:
> *@Grinnz* commented on this pull request.
> 
> 
> In ext/POSIX/t/sysconf.t 
> <https://github.com/Perl/perl5/pull/18250#discussion_r507207241>:
> 
>> @@ -1,4 +1,4 @@
> -#!perl -T
> +#!perl -Tw
> 
> |-w| does not run with warnings, it imposes warnings on anything that 
> does not express a preference. It should not be used in general.
> 

OK, p.r. updated.

PR #18255, opened 2020-10-19
embed.h: Add caution about PERL_NO_SHORT_NAMES

khwilliamson:


Leont:
I have been bitten by this before, when 5.10 eliminated the long form of the hv_* functions. Stopped using `PERL_NO_SHORT_NAMES` at that point.

Issue #18170, opened 2020-09-28
%g formatting broken on Ubuntu-18.04, NVSIZE == 8


sisyphus:
This issue is found with perl-5.32.0, all the way back to perl-5.8.9 (AFAICT), there being no such issue on perl-5.6.2
Same problem exists on my Debian Wheezy system, but Windows (Windows 7) and FreeBSD (freebsd-12.0) are fine.
I'm interested in fixing this - I would appreciate any hints on whereabouts in the perl source I should start looking.
For this report, I'll stick to version 5.32.0 and Ubuntu-18.04.

**Description**

"%.${digits}g" formatting is broken for $digits > 17, $Config{nvsize} == 8.
A quick check indicates that this type of issue does NOT occur with -Duselongdouble and -Dusequadmath builds of perl.

**Steps to Reproduce**

For example:
$ perl -le 'printf "%.54g\n", 0.3
0.29999999999999999
$ perl -le 'printf "%.53e\n", 0.3;'
2.99999999999999988897769753748434595763683319091796875e-01

**Expected behavior**

As you can see, the "%.53e" formatting is fine and produces the correct result.
I expected the "%.54g" formatting to produce:

0.299999999999999988897769753748434595763683319091796875

but, either my expectation is way out of line, or the %g formatting is severely crippled.
The significance of the correct output is that it exactly represents (as a base 10 number) the actual value held by the double 0.3.

The "%.54g" formatting works fine in C, using the compiler that built this perl-5.32.0.
Given that the capability for such extended formatting already exists, I'm hopeful that it can be provided to "%g" without too much difficulty.

Cheers,
Rob

**Perl configuration**

```
# perl -V output goes here
```
Summary of my perl5 (revision 5 version 32 subversion 0) configuration:

  Platform:
    osname=linux
    osvers=4.15.0-106-generic
    archname=x86_64-linux
    uname='linux sisyphus5-desktop 4.15.0-106-generic #107-ubuntu smp thu jun 4 11:27:52 utc 2020 x86_64 x86_64 x86_64 gnulinux '
    config_args='-des -Duse64bitall -Dprefix=/home/sisyphus/perl-5.32.0-d'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='cc'
[...]

PR #18262, opened 2020-10-21
perl's %g formatting broken on Ubuntu/Debian when NVSIZE == 8

sisyphus:
I stuffed up the bug report (https://github.com/Perl/perl5/issues/18170), turning it into an inordinately lengthy, messy and frustrating read.
So here is a concise rewrite of that issue - which this PR addresses.

The problem is that Gconvert calls gcvt (qgcvt with -Duselongdouble) on Ubuntu and Debian systems, even though those functions are buggy.
The bug is that gcvt will not output more than 17 digits, and qgcvt will not output more than 21 digits.
This has been reported at https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1899553

We therefore require that, when gcvt/qgcvt are available, Configure probes for this bug, and instead assigns sprintf to Gconvert and $Config{d_Gconvert} when and only when this bug is present.
The proposed changes to the Configure script achieve that aim.

The proposed changes to t/op/sprintf2.t test that this bug has been avoided.

However, in switching from gcvt/qgcvt to sprintf we find that bogus -Wformat-overflow warnings are issued during the compilation of sv.c. (This bug has been reported at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89161)
The proposed changes to sv.c suppress those bogus warnings.

Similarly, there is then a similar -Wformat-overflow warning issued in ext/XS-APItest when APItest.c is compiled.
This time the warnings are warranted.
The proposed changes to ext/XS-APItest/APItest.xs suppress those warnings by ensuring, to the satisfaction of gcc at least, that the overflow cannot occur.

Cheers,
Rob


sisyphus:
If someone can confirm that the failing t/porting/cmp_version.t can be fixed by simply bumping the version in APItest.pm, then I'll push a commit that does exactly that.
(I seem to recall that's all Karl did last time I made changes to APItest.xs.)

Howver, I'm unable to reproduce the failure locally, and I don't want to turn this PR into a shambles by pushing speculative commits that subsequently turn out to be ineffective.

Cheers,
Rob

t/op/sprintf2.t
```diff
@@ -1178,4 +1178,27 @@ if ($Config{intsize} == 4 && $Config{uvsize} > 4 && $Config{sizesize} > 4) {
     is($off2, 1, "offset after 0");
 }
 
+# %g formatting was broken on Ubuntu, Debian and perhaps other systems
+# for a long time. Here we verify that no such breakage still exists.
+# See https://github.com/Perl/perl5/issues/18170
```
hvds, line 1183: This should probably reference #18262 now.

ext/XS-APItest/APItest.xs
```diff
@@ -6802,8 +6802,14 @@ test_Gconvert(SV * number, SV * num_digits)
...
     CODE:
         len = (int) SvIV(num_digits);
-        if (len > 99) croak("Too long a number for test_Gconvert");
-        if (len < 0) croak("Too short a number for test_Gconvert");
+        if (sizeof(NV) == 8) {
+            if(len > 92)
+              croak("Too long a number for test_Gconvert");
+        }
+        else if (len > 91) /* exponent can be 1 digit larger */
+            croak("Too long a number for test_Gconvert");
```
hvds: It would be worth a comment to explain these limits; I think something like the below would be sufficient:
```
  /* bufsize - float_need = 127 - 35 = 92. See comments around the setting of float_need in sv_vcatpvfn_flags() */
```

hvds:
> I'm unable to reproduce the failure locally

What output do you get if you run `porting/cmp_version.t` manually within this branch?

(I'm pretty confident you just need to bump the version of APITest, but it would be even better if the test worked for you.)

ext/XS-APItest/APItest.xs
sisyphus: This is a little bit different to what happened in sv.c.

In sv.c the buffersize was 127 and we said we'll write to the buffer only if the no. of digits is less than buffersize - 35.
In APItest.xs, the buffersize is 100 and we would write to the buffer so long as the no. of digits is less than 100 - 1, which allowed only for the terminating NULL byte. The overflow warning was therefore justified.

All I've done is to allow for a total of 8 bytes - one for the terminating null, one for the radix point, one for a leading '-', two for an 'e-' and three for the exponent digits.
For the long double, I've allowed an extra byte because there can be four exponent digits.
So, it's a case of 100 - 8 = 92 and 100 - 9 = 91.

That was enough to suppress the overflow warning.
If you think I should allow more than those 8 (or 9) bytes just specify that number and I'll accept it without question.
Also, whatever the no. of bytes we decide to allow, I will provide comments that explain just how we arrived at that number.

(BTW, I'll modify my sprintf2.t comments to point to https://github.com/Perl/perl5/pull/18262 , as you've suggested).

sisyphus:
Running the test manually doesn't change anything. 
I presume the reason that I can't reproduce the failure is that Travis is using different source than I am.
I'm just using the source as it appears currently in https://github.com/sisyphus/perl5 , but, although I created my fork just yesterday, there has since been a release of blead and I don't know if that is having some effect.

I'll try to sync my fork with latest blead and see if that can create the failure.

sisyphus:
I didn't actually sync my fork with current blead, but I did ascertain that if I were to do so then I would see the failure that Travis reports.
And I also ascertained that bumping the APItest.pm version was all that was needed to avert the failure.
I'll attend to that tomorrow ... along with the other things that I mentioned previously ... and along with anything else that crops up.

BTW, if it's relevant, test_Gconvert() is tested only in ext/XS-APItest/t/locale.t where it gets called twice as <code>test_Gconvert(4.179, 2)</code>

ext/XS-APItest/APItest.xs
hvds: I don't have a problem with the number, as long as there's some explanation of where it comes from.

t/op/sprintf2.t
tonycoz: I think referring to the bug report is more correct here (which is what he did.)

sisyphus: It may be more correct, but I actually would be happy to refer instead to the PR - simply because https://github.com/Perl/perl5/issues/18170 is such a long, shambolic account (my fault).
Also, this PR begins with a link to #18170 , so the original issue can easily be located.

So, I'm more than happy to proceed with the suggestion that @hvds made, but I won't do so unless given a clear go-ahead. (Personally, I'm not fussed either way.)
I've just now pushed a new APItest.pm (bumped version to 1.13) and APItest.xs (additional clarity) but I've left spritnf2.t as it already was.

sisyphus:
The Windows failures are interesting.

The mingw64 failure highlights the need to have the GNUmakefile and the makefile.mk specify <code>-D__USE_MINGW_ANSI_STDIO</code> by default, if the compiler is gcc.
Both I and Strawberry Perl have been building perl that way beginning with perl-5.26, and there's no problem - but it's something you have to specifically configure, whereas it should be the default.
It's good that this is now being exposed by my proposed sprintf2.t, and I'll provide a PR that fixes those makefiles as soon as (if ever) I can get this PR dealt with.

The cygwin failure is surprising.
The "%.55g" formatting of 0.3 outputs the first 43 digits correctly, then bails out. (Why only 43 digits ?)
But then, cygwin is known for its capacity to surprise.
It's good that this Cygwin bug is also being exposed.
Perhaps someone can figure out how to deal with it.

Good to see no failure on the msvc142 build. Congrats to Microsoft for producing a runtime that can handle this correctly.
Not so good that it took them so long - and I don't currently know how to fix this same bug in the msvc100 build.
I do have msvc100, and I have looked at this type of failing with that runtime before, but I couldn't find a solution

What is most concerning to me is that, with current blead and these proposed changes, I've just started getting some other sprintf2.t failures on Ubuntu-20.04.
Actually, there seems to be an element of intermittency regarding these errors - and I don't really yet know whether my changes are part of the problem or not.

So maybe wait at least until I get this new development sorted before pulling this PR;-)

Here, attached, are these new t/op/sprintf2.t failures:

[sprintf2.t_fail.txt](https://github.com/Perl/perl5/files/5427264/sprintf2.t_fail.txt)

hvds:
> Here, attached, are these new t/op/sprintf2.t failures:
> 
> [sprintf2.t_fail.txt](https://github.com/Perl/perl5/files/5427264/sprintf2.t_fail.txt)

Those look very close to the results you get if you remove the `-w` from the shebang line - but if I do that I see one additional failure (test 96 got 0 expected 1). Assuming the shebang line is intact, does it change if you put an additional `use warnings;` at the top of the test file?

t/op/sprintf2.t
hvds: > Also, this PR begins with a link to #18170 , so the original issue can easily be located.

That was my reasoning too, but I'm happy to defer to @tonycoz .

sisyphus:
Oh, thank you @hvds .
Test 96  was also failing.
Somehow, I managed to have 2 slightly different versions of sprintf2.t - one with the shebang line at line 1, and another with the shebang line at line 2. (The result of some bodgey editing I had done on the Windows box.)
The test failures were occurring if I ran the latter, but not if I ran the former.

I don't know how long it would have taken me to realize what the problem was without your observation.
I never use shebang lines and rarely pay any attention to them at all.

Anyway, from my POV, this PR is in good shape for smoking.
Having fixed the problem with Ubuntu, it correctly identifies those systems in which %g formatting is buggy.

Of the 3 failures reported by the automated tests conducted for this PR, I think that the msvc100 failure might not be readily fixable.

The mingw64 failure is easily fixed by building perl with <code>-D__USE_MINGW_ANSI_STDIO</code> as I've previously mentioned.

I've just had a quick look at what happens on my 5-year-old Cygwin installation (gcc-4.9.2, perl-5.14.4) where <code>perl -V:d_Gconvert</code> reports:
d_Gconvert='sprintf((b),"%.*g",(n),(x))';
With typical contrariness, the cygwin failure would be averted if Gconvert were to use gcvt() instead of sprintf().
I've never successfully built perl on Cygwin (having tried only a few times), and I don't know how to coerce it into using gcvt().
But it seems that such a coercion would remove this failure.
This view is based solely on the output of the attached C program which demonstrates that sprintf() produces the same incorrect output as reported by the sprintf2.t failure - and also demonstrates that gcvt() produces correct output.

[try.c.txt](https://github.com/Perl/perl5/files/5432309/try.c.txt)

Let me know if I should add some comments to sprintf2.t regarding the possibility that this particular test fails
Maybe something like:

\# If this test fails, see https://github.com/Perl/perl5/pull/18262
\# for some ideas regarding the cause of the failure

.

khwilliamson:
On 10/23/20 8:47 PM, sisyphus wrote:
> Of the 3 failures reported by the automated tests conducted for this PR, 
> I think that the msvc100 failure might not be readily fixable.

That release is a likely candidate to have support dropped for

khwilliamson:
On 10/23/20 8:47 PM, sisyphus wrote:
> Oh, thank you @hvds <https://github.com/hvds> .
> Test 96 was also failing.
> Somehow, I managed to have 2 slightly different versions of sprintf2.t - 
> one with the shebang line at line 1, and another with the shebang line 
> at line 2. (The result of some bodgey editing I had done on the Windows 
> box.)
> The test failures were occurring if I ran the latter, but not if I ran 
> the former.
> 
> I don't know how long it would have taken me to realize what the problem 
> was without your observation.
> I never use shebang lines and rarely pay any attention to them at all.
> 
> Anyway, from my POV, this PR is in good shape for smoking.

I have started a smoke of this.  The branch is smoke-me/khw-gconvert

sisyphus:
I've just realized that the changes I've proposed in https://github.com/Perl/perl5/issues/18270 should have been included in this PR.
How could I have not realized this obvious fact earlier ? (I've stuffed up again !)

If I had simply added those changes to this PR, then the automated Windows mingw64 build that the PR instigates should succeed.
The cygwin and msvc results would remain unchanged, as my makefile amendments affect only gcc builds on native windows.
The failure with George's Windows 2000 smoker at http://perl.develop-help.com/?b=smoke-me%2Fkhw-gconvert would therefore still persist, as it's an msvc build.

The absence of those makefile amendments doesn't seem to be having any effect on the smoking of the khw-gconvert branch.
It also seems to me that, with the exception of George's Windows 2000 smoker, all of the failures in that branch are unrelated to this PR.

Is it ok if I now push those makefile amendments to this PR ?
I think it's at least worth verifying that they do fix the mingw64 failure ... but I don't want to compound one stuff up with another, as I've already exceeded my quota wrt that activity.

sisyphus:
I've been looking at the reasons for the cygwin failure.
The Configure process will reject the use of gcvt() because the following probes will fail:
1)<code>gcvt(0.0001, 8, buf)</code> sets buf to "1.e-04" instead of the expected "0.0001";
2)<code>gcvt(0.00009, 8, buf)</code> sets buf to "9.e-05" instead of the expected "9e-05";
3)<code>gcvt(1e34, 8, buf)</code> sets buf to "1.e+34" instead of the expected "1e+34".

Therefore sprintf() gets used.

Nothing wrong with the output being provided by gcvt() - except that (I  believe) it's not correct "%.8g" formatting.

Using gcvt causes a few tests to fail - mostly because of the type of difference between "got" and "expected" that I've just illustrated above.
They would be trivial to fix.

What might not be so trivial to fix (I haven't looked) is the failure of test 29 in cpan/ExtUtils-MakeMaker/t/vstrings.t which offers the diagostics:
"Failed test 'Scientific Notation handled right'  at t/vstrings.t line 102.
Unparsable version '5.e-06' for prerequisite Fake::Scientific treated as 5".

I'm guessing that, on Cygwin, use of gcvt() in place of sprintf() would be regarded as a backwards step.
If, OTOH, the switch to gcvt() would be deemed an improvement to Cygwin, then just let me know and I'll push the necessary changes to this PR, as soon as I've written them.

I also spent some time looking at what happens with msvc100, but fixing it still looks difficult to me. I think you'd have to write a function that provides correct "%g" formatting and use that function instead of anything that's being provided by that MS runtime.

hvds:
> I'm guessing that, on Cygwin, use of gcvt() in place of sprintf() would be regarded as a backwards step.
> If, OTOH, the switch to gcvt() would be deemed an improvement to Cygwin, then just let me know and I'll push the necessary changes to this PR, as soon as I've written them.

From what you've found so far, it sounds to me like it would be a backward step.

sisyphus:
I don't have much to add.
Fixing the issues with Cygwin and MSVC100 will, I think, require a lot of work - more work than I am prepared to undertake.
I think that we have no option than to allow the issue with those 2 systems to continue to exist until someone who is prepared to fix them comes along.
They've existed for a long time, without any complaints - a bit longer shouldn't hurt.

I don't see anything in any of the reports in Karl's smoke-me/khw-gconvert branch that calls for any action from me - except for George's cl: 14.00.50727.762 smoker (https://www.nntp.perl.org/group/perl.daily-build.reports/2020/10/msg259348.html).
This is the same MSVC100 issue as mentioned in the preceding paragraph. I don't have a fix for it.

The only thing left for me to do is to add the altered win32 makefiles to this PR - which I shall shortly do.
It might seem odd that a PR whose subject line claims to be addressing an Ubuntu/Debian issue wants to make changes to some win32 makefiles.
But, as has already been shown, the issue that we considered here extends beyond Ubuntu and Debian.

With the revised makefiles the failure with p5p's mingw64 smoker should be fixed.

Cheers,
Rob

sisyphus:
As expected, t/op/sprintf2.t is exposing the bug on cygwin and msvc100.
The macOS xcode 11 failure is unrelated to this PR.

IMO, this is now fit for merging.

Cheers,
Rob

Configure
```diff
@@ -10445,6 +10445,15 @@ int main()
...
 
+        /* Test for an Ubuntu/Debian bug in gcvt and qgcvt. See:        *
+         * https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1899553 */
+
+	Gconvert((DOUBLETYPE)0.4, 53, 0, buf);
+        if(sizeof(DOUBLETYPE) == 8)
+	  checkit("0.40000000000000002220446049250313080847263336181640625", buf);
+        else
+	  checkit("0.40000000000000000000542101086242752217003726400434971", buf);
+
```
tonycoz: This looks very non-portable.  Given the problem being fixed should we just be checking the length of the resulting string?

khwilliamson:
On 11/6/20 10:31 PM, sisyphus wrote:
> As expected, t/op/sprintf2.t is exposing the bug on cygwin and msvc100.
> The macOS xcode 11 failure is unrelated to this PR.
> 

Should the tests on the failing platforms be marked as Todo?

Configure
sisyphus: I think the lack of portability is not an issue wrt -Dusequadmath builds because they never call Gconvert() anyway.
However, I s'pose we should do the right thing by Double-Double and IEEE-754 128 bit long double builds - and, as you've suggested, that's not currently being done.

IMO, checking the length is a good enough place to start.
I don't know of any cases where the length will be correct && the value will be incorrect.
However, if such cases do exist, then a t/op/sprintf2.t failure should alert us to the fact.

I'll come up with something tomorrow.

sisyphus:
I don't know about p5p TODO policy.
Seems to me that such concealment is not the right thing to do.

But I don't care much about cygwin and builds using old MS toolchains, so I'll comply with any directive that's given in response to Karl's question.

khwilliamson:
We create a ticket for those, with a 5.34 label, to make sure it won't be swept under the carpet.  My view is the old MSVC is not going to get fixed.

khwilliamson:
I think this is getting close to being ready to push. @sisyphus could you rebase to a reasonable set of commits.  For example, the TODOing should come in an order such that those tests fail.

sisyphus:
@khwilliamson I don't understand what you're getting at when you say "the TODOing should come in an order such that those tests fail".
Can you elaborate ?
(Sorry - I've spent a good half-hour wondering, but still haven't seen the light.)

khwilliamson:
On 11/16/20 4:59 PM, sisyphus wrote:
> @khwilliamson <https://github.com/khwilliamson> I don't understand what 
> you're getting at when you say "the TODOing should come in an order such 
> that those tests fail".
> Can you elaborate ?
> (Sorry - I've spent a good half-hour wondering, but still haven't seen 
> the light.)


Sorry for causing you wasted effort.

It's a bad thing for blead to be broken.  And when it happens there is 
much gnashing of teeth and effort to fix it as soon as possible.

One of the reasons it is bad is because later when we are bisecting a 
range of commits for the breaking one, if the bisection happens to land 
on a commit that is bad, it can cause the bisection to fail.

Suppose some later commit causes one of the .t files to fail that here 
need to have a TODO added.  The bisection might stop here instead of the 
real culprit.

So, I try to order my commits so that at no time does the test suite 
exhibit a failure.  If a commit comes from a new author, they will have 
to be added to the AUTHORS file, and I will create a commit to do that, 
and position it, by rebasing before pushing, so that it comes just 
before the one they authored.  Others don't bother with that; I suppose 
we will never try to bisect a porting test.

The bottom line is that you should (re)order and squash your commits so 
that either the commit that causes a .t to fail has the TODO added in 
the same commit, or, if you prefer, you can keep a separate TODO commit, 
but have it ordered before the breaking one.  In theory, that would make 
it easier to later just revert that commit if and when the problem gets 
fixed.  But in practice no one is going to remember that this was a 
separate commit, and they'll just remove the TODO in a separate commit.

If this still doesn't make sense, let me know, and I'll plead with a 
better explainer than me, to respond.

khwilliamson:
Rebased and merged into blead via 808580607a3c1858a97a41b7dc5e2d0ea25b63a6

PR #18267, opened 2020-10-22
Octal constants with "0o" prefix

t-a-k:
I suggest this pr to support octal constants through `0odddd` syntax, as described in Porting/todo.pod.

Modifications consist of two parts:
- octal literals in the script can be specified with `0o` prefix
- `oct()` will accept (simply skip) `0o` or `o` prefix.

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -444,6 +444,7 @@ integer formats:
  0xff                # hex
  0xdead_beef         # more hex
  0377                # octal (only numbers, begins with 0)
+ 0o12_345            # alternative octal (introduced in Perl 5.XXX)
  0b011011            # binary
  0x1.999ap-4         # hexadecimal floating point (the 'p' is required)
 
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -444,7 +444,7 @@ integer formats:
  0xff                # hex
  0xdead_beef         # more hex
  0377                # octal (only numbers, begins with 0)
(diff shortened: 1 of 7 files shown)


khwilliamson:
Thanks for this

Two things though.  The Perl documentation needs to be updated to reflect the new syntax, and I think the third commit doesn't belong with the others, and should be submitted separately.  It is my experience, BTW, that changes to speed up the compilation have no noticeable effect.

t-a-k:
Thank you for comment.
I had pushed updates for documentation.  (For now, they contain `XXX` placeholders to be replaced with Perl version if this PR can be merged.)

> I think the third commit doesn't belong with the others, and should be submitted separately. It is my experience, BTW, that changes to speed up the compilation have no noticeable effect.

I think the third commit is not so unrelated thing, because the reason to make this commit is that the first commit introduced conditional expression like `max = new_octal ? "0o37777777777" : maxima[shift];` which I think is a bit too complex (and, I think, is not so clean) to be executed unconditionally.

khwilliamson:
Please add your doc updates to this PR.  You can say 5.33.5

t-a-k:
> Please add your doc updates to this PR. You can say 5.33.5

Thank you for taking time.
I'd updated `perlfunc` and `perldata`, and added an entry to `perldelta`.
(Other commits in previous push are due to rebasing to recent version of `blead`.)
If there are any inappropriate, please point out.

khwilliamson:
Merged with final commit being 54e31e2485e0fb4822dfa38ac496790a2d877f3c

Issue #18245, opened 2020-10-17
ext/POSIX/t/posix.t: Incorrect use of prototyped testing function


jkeenan:
In the core distribution (), `ext/POSIX/t/posix.t` contains this code:
```
356 SKIP: {
357     skip("locales not available", 26) unless locales_enabled(qw(NUMERIC MONETARY));
358     skip("localeconv() not available", 26) unless $Config{d_locconv};
```
When this file is run with warnings enabled, we get a compile-time error at line 357.
```
$ cd t; ./perl -w -I../lib -I. -I../ext/POSIX/lib ../ext/POSIX/t/posix.t; cd -
1..96
Useless use of a constant ("NUMERIC") in void context at ../ext/POSIX/t/posix.t line 357.
...
```
At line 357, `locales_enabled` is a function imported from `t/loc_tools.pl`.  This function is prototyped.
```
sub locales_enabled(;$) {
    # Returns 0 if no locale handling is available on this platform; otherwise
    # 1.
    #
    # The optional parameter is a reference to a list of individual POSIX
    # locale categories.  If any of the individual categories specified by the
    # optional parameter is all digits (and an optional leading minus), it is
    # taken to be the C enum for the category (e.g., &POSIX::LC_CTYPE).
    # Otherwise it should be a string name of the category, like 'LC_TIME'.
    # The initial 'LC_' is optional.  It is a fatal error to call this with
    # something that isn't a known category to this file.
    #
    # This optional parameter denotes which POSIX locale categories must be
    # available on the platform.  If any aren't available, this function
    # returns 0; otherwise it returns 1 and changes the list for the caller so
    # that any category names are converted into their equivalent numbers, and
    # sorts it to match the expectations of _trylocale.
    #
    # It is acceptable for the second parameter to be just a simple scalar
    # denoting a single category (either name or number).  No conversion into
    # a number is done in this case.
...
```
If I'm reading the prototype correctly, `locales_enabled()` takes a single, optional parameter which must be a scalar.  However, the inline comment suggests that that scalar must be an array reference.  If that is the case then to say:
```
locales_enabled(qw(NUMERIC MONETARY));
```
... would be an error because the function was being provided a list of scalars, not an array ref.

If my reasoning is correct, this patch should clear the warning:
```
diff --git a/ext/POSIX/t/posix.t b/ext/POSIX/t/posix.t
index 25099ea54a..2b317fa894 100644
--- a/ext/POSIX/t/posix.t
[...]

PR #18268, opened 2020-10-23
t/loc_tools.pl: De-prototype to allow for array ref

jkeenan:
Running ext/POSIX/t/posix.t with warnings enabled generated a
compile-time error, "Useless use of a constant ("NUMERIC") in void
context".  Analysis suggested it was a problem with t/loc_tools.pl
function locales_enabled(), which was prototyped to accept an optional
scalar reference but not an array reference.  Experimentation suggested
that the best way out of this problem was not to prototype the function.
Rather, account for the possibility that locales_enabled() would take an
array reference as well as existing cases of scalar and no argument at
all.

For : GH #18245

### DIFF
--- a/t/loc_tools.pl
+++ b/t/loc_tools.pl
@@ -186,7 +186,7 @@ sub valid_locale_categories() {
     return @platform_categories;
 }
 
-sub locales_enabled(;$) {
+sub locales_enabled {
     # Returns 0 if no locale handling is available on this platform; otherwise
     # 1.
     #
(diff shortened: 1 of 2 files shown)


t/loc_tools.pl
```diff
@@ -245,8 +245,13 @@ (;$)
         my @local_categories_copy;
 
         if (ref $categories_ref) {
-            @local_categories_copy = @$$categories_ref;
-            $return_categories_numbers = 1;
+            if (ref $categories_ref eq 'ARRAY') {
+                @local_categories_copy = @$categories_ref;
+            }
+            else {
+                @local_categories_copy = @$$categories_ref;
```
Grinnz, line 252: Not sure what this is supposed to do. If `$categories_ref` is not an arrayref then it will be a string.

Grinnz:
This isn't what the prototype does. It takes an optional scalar argument, and an arrayref is a scalar argument. Regardless there is probably no need for the prototype, but you should make sure no use cases are relying on the prototype. For example `locales_enabled $arrayref, other_stuff` which with this prototype would parse as `locale_enabled($arrayref), other_stuff` but without the prototype would pass the rest as arguments to locales_enabled.

khwilliamson:
Thank you for discovering this.  I would rather, however, fix the root cause, which stemmed from my misunderstanding some things.  I am working on that.

PR #18274, opened 2020-10-29
win32: remove support for disabling USE_LARGE_FILES

xenu:
It was enabled by default on all compilers. I don't think it ever
makes sense to disable it.


win32/GNUmakefile
```diff
@@ -1213,7 +1206,6 @@ CFG_VARS	=					\
 		"useperlio=$(USE_PERLIO)"		\
 		"use64bitint=$(USE_64_BIT_INT)"		\
 		"uselongdouble=$(USE_LONG_DOUBLE)"	\
-		"uselargefiles=$(USE_LARGE_FILES)"	\
```
Leont, line 1216: How does this affect `$Config{uselargefiles}`? This looks like it wouldn't be set anymore, which would be wrong.

xenu: This part of makefile overrides config values from `config.gc` (or `config.vc`), so after this change, `$Config{uselargefiles}` will be set to the value from that file.

PR #18275, opened 2020-10-30
Adding perldocstyle.pod.

jmacdotorg:
Adds `perldocstyle.pod` as a Porting document. (For more information and context, see <https://jmac.org/misc/perldoc/>.)

### DIFF
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -212,7 +212,7 @@ state its assumptions up front--and quickly redirect readers who are
 instead looking for a more basic reference or tutorial.
 
 Reference pages, when appropriate, can precede the DESCRIPTION with a
-SYNOPSIS section that lists, within one more code blocks, some very
+SYNOPSIS section that lists, within one or more code blocks, some very
 brief examples of the referenced feature's use. This section should show
 a handful of common-case and best-practice examples, rather than an
 exhaustive list of every obscure method or alternate syntax available.
@@ -963,8 +963,9 @@ and best practices> should reflect this.
 Even though it comes from many hands and minds, criss-crossing through
 the many years of Perl's lifetime, the language's documentation should
 speak with a single, consistent voice. With few exceptions, the docs
-should avoid explicit first-person statements, or similar self-reference
-to any individual's contributor's philosophies or experiences.
+should avoid explicit first-person-singular statements, or similar
+self-reference to any individual's contributor's philosophies or
+experiences.
 
 Perl did begin life as a deeply personal expression by a single
 individual, and this famously carried through the first revisions of its
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -576,7 +576,7 @@ presents the reader with small, interesting examples that allow them to
 follow along themselves using their own Perl interpreter. The tutorial
 inspires comprehension by letting its readers immediately experience
 (and experiment on) the concept in question. Examples include
-C<perlxstut>, <C<perlpacktut>, and
+C<perlxstut>, C<perlpacktut>, and
 C<perlretut>.
 
 Tutorial man pages must strive for a welcoming and reassuring tone from
@@ -634,7 +634,7 @@ Perl's own best example of a reference work is C<perlfunc>, the
 sprawling man page that details the operation of every function built
 into Perl, with each function's documentation presenting the same kinds
 of information in the same order as every other. For an example of a
-shorter reference on a single topic, see C<perlreref>.
+shorter reference on a single topic, look at C<perlreref>.
 
 Module documentation--including that of all the modules listed in
 L<C<perlmodlib>|perlmodlib>--also counts as reference. They follow
@@ -784,7 +784,7 @@ magazine articles, published elsewhere and with a narrower readership in
 mind. But Perl's own docs should use language as accessible and
 welcoming to as wide an audience as possible.
 
-=head2 Submit only complete documentation
+=head2 Omit placeholder text or commentary
 
 Placeholder text does not belong in the documentation that ships with
 Perl. No section header should be followed by text reading only "Watch
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -78,7 +78,7 @@ particular to Perl's programming manual.
 All of Perl's core documentation uses Pod ("Plain Old Documentation"), a
 simple markup language, to format its source text. Pod is similar in
 spirit to other contemporary lightweight markup technologies, such as
-Markdown and ReStructuredText, and has a decades-long shared history
+Markdown and reStructuredText, and has a decades-long shared history
 with Perl itself.
 
 For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
@@ -158,7 +158,7 @@ patterns its structure after a number of long-time Unix man page
 conventions. (Hence this guide's use of "man page" to refer to any one
 self-contained part of Perl's documentation.)
 
-Adhering to these conventions help Pod formatters present a Perl man
+Adhering to these conventions helps Pod formatters present a Perl man
 page's content in different contexts--whether a terminal, the web, or
 even print. Many of the following requirements originate with
 L<C<perlpodstyle>|perlpodstyle>, which derives its recommendations in
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -92,18 +92,38 @@ L<C<perlpodstyle>|perlpodstyle>.
 
 =head2 Choice of language: American English
 
-Perl's core documentation is written in American English. (The Chinese,
-Japanese, and Korean-language README files included with Perl's source
-distributions provide an exception to this--but these documents fall
-outside the scope of this guide.)
+Perl's core documentation is written in English, with a preference for
+American spelling of words and expression of phrases. That means "color"
+over "colour", "math" versus "maths", "the team has decided" and not
+"the team have decided", and so on.
+
+We name one style of English for the sake of consistency across Perl's
+documentation, much as a software project might declare a four-space
+indentation standard--even when that doesn't affect how well the code
+compiles. Both efforts result in an easier read by avoiding jarring,
+mid-document changes in format or style.
+
+Contributors to Perl's documentation should note that this rule
+describes the ultimate, published output of the project, and does not
+prescribe the dialect used within community contributions. The
+documentation team enthusiastically welcomes any English-language
+contributions, and will actively assist in Americanizing spelling and
+style when warranted.
+
+=head3 Other languages and translations
 
 Community-authored translations of Perl's documentation do exist,
-covering a variety or languages. While the Perl project appreciates
+covering a variety of languages. While the Perl project appreciates
 these translation efforts and promotes them when applicable, it does not
 officially support or maintain any of them.
 
 That said, keeping Perl's documentation clear, simple, and short has a
 welcome side effect of aiding any such translation project.
+
+(Note that the Chinese, Japanese, and Korean-language README files
+included with Perl's source distributions provide an exception to this
+choice of language--but these documents fall outside the scope of this
+guide.)
 
 =head2 Choice of encoding: UTF-8
 
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -681,20 +681,21 @@ wide-ranging discussion laden with many references to the language's
 history, culture, and driving philosophies.
 
 Perl's explainer pages give authors a chance to explore Perl's penchant
-for TMTOWTDI, illustrating alternate and even obscure ways to use the
-language feature under discussion. However, as the remainder of this
-guide discusses, the ideal Perl documentation manages to deliver its
-message clearly and concisely, and not confuse mere wordiness for
-completeness.
+for L<TMTOWTDI|perlglossary/TMTOWTDI>, illustrating alternate and even
+obscure ways to use the language feature under discussion. However, as
+the remainder of this guide discusses, the ideal Perl documentation
+manages to deliver its message clearly and concisely, and not confuse
+mere wordiness for completeness.
 
 =head3 Further notes on documentation modes
 
 Keep in mind that the purpose of this categorization is not to dictate
 content--a very thorough explainer might contain short reference
 sections of its own, for example, or a reference page about a very
-complex function might resemble an explainer in places (c.f. [open]).
-Rather, it makes sure that the authors and contributors of any given man
-page agree on what sort of audience that page addresses.
+complex function might resemble an explainer in places (e.g.
+L<C<open>|perlfunc/open FILEHANDLE,MODE,EXPR>). Rather, it makes sure
+that the authors and contributors of any given man page agree on what
+sort of audience that page addresses.
 
 If a new or otherwise uncategorized man page presents itself as
 resistant to fitting into only one of the four modes, consider breaking
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -152,9 +152,17 @@ readability.
 
 =head2 Document structure
 
-Many of these requirements originate with
+Each individual work of core Perl documentation, whether contained
+within a C<.pod> file or in the Pod section of a standard code module,
+patterns its structure after a number of long-time Unix man page
+conventions. (Hence this guide's use of "man page" to refer to any one
+self-contained part of Perl's documentation.)
+
+Adhering to these conventions help Pod formatters present a Perl man
+page's content in different contexts--whether a terminal, the web, or
+even print. Many of the following requirements originate with
 L<C<perlpodstyle>|perlpodstyle>, which derives its recommendations in
-turn from long-time Unix man page conventions.
+turn from these well-established practices.
 
 =head3 Name
 
@@ -867,7 +875,13 @@ to keep reading for a deeper understanding if they feel so driven.
 Perl's documentation should stay focused on Perl's present behavior,
 with a nod to future directions.
 
-=head3 Keep the past in the past
+=head3 Recount the past only when necessary
+
+=for comment
+The principles of this section caused a lot of lively discussion and
+debate among p5p when first proposed in October 2020. I am keeping the
+recommendations nonspecific, and expect this section to receive a lot of
+further refinement as we start to apply it to core docs.
 
 When some Perl feature changes its behavior, documentation about
 that feature should change too, and just as definitively. The docs have
@@ -881,7 +895,13 @@ boldly when needed. Perl's history remains safe, even when you delete or
 replace outdated information from the current version's docs.
 
 Perl's docs can acknowledge or discuss former behavior when warranted,
-but should otherwise keep it in the past. A manual uncluttered with
+including notes that some feature appeared in the language as of some
+specific version number. Authors should consider applying principles
+similar to those for deprecated techniques, L<as described above|/Show
+the lesser ways when needed>: make the information present, but not
+prominent.
+
+Otherwise, keep the past in the past. A manual uncluttered with
 outdated instruction stays more succinct and relevant.
 
 =head3 Describe the uncertain future with care
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -49,7 +49,7 @@ more welcome and engaged with Perl's documentation, and this in turn
 will help the Perl project itself grow stronger through having a larger,
 more diverse, and more confident population of knowledgeable users.
 
-=head2 Who this guide is for
+=head2 Intended audience
 
 Anyone interested in contributing to Perl's core documentation should
 familiarize themselves with the standards outlined by this guide.
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -536,8 +536,9 @@ explanatory text and code examples.
 =back
 
 Complex functions that need their text divided into subsections (under
-the principles of ["Apply section-breaks and examples generously"]) may
-do so by using sublists, with C<=item> elements as header text.
+the principles of L<"Apply section-breaks and examples
+generously"|/Apply section-breaks and examples generously>) may do so by
+using sublists, with C<=item> elements as header text.
 
 A fictional function "C<myfunc>", which takes a list as an optional
 argument, might have an entry in perlfunc shaped like this:
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -184,12 +184,12 @@ The first few lines of a notional page named C<perlpodexample>:
 Most Perl man pages also contain a DESCRIPTION section featuring a
 summary of, or introduction to, the document's content and purpose.
 
-This section should, one way or another, also make clear the audience
-that the page addresses, especially if it has expectations about the
-reader's prior knowledge. For example, a man page that dives deep into
-the inner workings of Perl's regular expression engine should state its
-assumptions up front--and quickly redirect readers who are instead
-looking for a more basic reference or tutorial.
+This section should also, one way or another, clearly identify the
+audience that the page addresses, especially if it has expectations
+about the reader's prior knowledge. For example, a man page that dives
+deep into the inner workings of Perl's regular expression engine should
+state its assumptions up front--and quickly redirect readers who are
+instead looking for a more basic reference or tutorial.
 
 Reference pages, when appropriate, can precede the DESCRIPTION with a
 SYNOPSIS section that lists, within one more code blocks, some very
@@ -216,15 +216,18 @@ suggested elsewhere in this guide|/Apply one of the four documentation modes>.
 
 =head3 Author and copyright
 
-Perl's core man pages inherit the copyright and license
-information of Perl itself. They should not provide any
-additional such information.
+In most circumstances, Perl's stand-alone man pages--those contained
+within C<.pod> files--do not need to include any copyright or license
+information about themselves. Their source Pod files are part of Perl's
+own core software repository, and that already covers them under the
+same copyright and license terms as Perl itself. You do not need to
+include additional "LICENSE" or "COPYRIGHT" sections of your own.
 
-Man pages do have the option to credit their primary
-author, or include a list of significant contributors. Note that the
-presence of authors' names does not preclude a given page from L<writing
-in a voice consistent with the rest of Perl's documentation|/The
-documentation speaks with one voice>.
+These man pages may optionally credit their primary author, or include a
+list of significant contributors, under "AUTHOR" or "CONTRIBUTORS"
+headings. Note that the presence of authors' names does not preclude a
+given page from L<writing in a voice consistent with the rest of Perl's
+documentation|/The documentation speaks with one voice>.
 
 Note that these guidelines do not apply to the core software modules
 that ship with Perl. These have their own standards for authorship and
@@ -248,16 +251,13 @@ readable as possible, displaying no more complexity than absolutely
 necessary to illustrate the concept at hand.
 
 For the sake of consistency within and across Perl's man pages, all
-examples must adhere to the code-layout principles set out by Damian
-Conway in chapter 2 of  I<Perl Best Practices> (O'Reilly Media, 2005).
-This means four-space indentation, K&R-style block-bracketing, and so
-on. (The rare code examples written expressly to demonstrate alternate
-Perl syntax are exempt from this.)
+examples must adhere to the code-layout principles set out by
+L<C<perlstyle>|perlstyle>.
 
-L<As with CMOS|/Choice of underlying style guide: CMOS>, access to I<Perl
-Best Practices> is not a prerequisite to making Perl documentation
-contributions; the doc team will help contributors standardize their
-code samples when needed.
+Sample code should deviate from these standards only when necessary:
+during a demonstration of how Perl disregards whitespace, for example,
+or to temporarily switch to two-column indentation for an unavoidably
+verbose illustration.
 
 You may include comments within example code to further clarify or label
 the code's behavior in-line. You may also use comments as placeholder
@@ -269,6 +269,10 @@ so:
         # (Do something interesting with $line here.)
         #
     }
+
+Even the simplest code blocks often require the use of example
+variables and subroutines, L<whose names you should choose with
+care/Use meaningful variable and symbol names in examples>.
 
 =head3 Inline code and literals
 
@@ -543,7 +547,7 @@ argument, might have an entry in perlfunc shaped like this:
 
 =head2 Apply one of the four documentation modes
 
-Aside from "meta" documentation such as C<perlhist> or C<perllicense>,
+Aside from "meta" documentation such as C<perlhist> or C<perlartistic>,
 each of Perl's man pages should conform to one of the four documentation
 "modes" suggested by L<I<The Documentation System> by Daniele
 Procida|https://documentation.divio.com>. These include tutorials,
@@ -570,8 +574,10 @@ expected, despite ever-evolving situations.
 A tutorial man page focuses on B<learning>, ideally by I<doing>. It
 presents the reader with small, interesting examples that allow them to
 follow along themselves using their own Perl interpreter. The tutorial
-inspires comprehension by letting its readers immediately experience (and
-experiment on) the concept in question.
+inspires comprehension by letting its readers immediately experience
+(and experiment on) the concept in question. Examples include
+C<perlxstut>, <C<perlpacktut>, and
+C<perlretut>.
 
 Tutorial man pages must strive for a welcoming and reassuring tone from
 their outset; they may very well be the first things that a newcomer to
@@ -595,7 +601,9 @@ better handled by explainers and reference pages.
 
 A cookbook man page focuses on B<results>. Just like its name suggests,
 it presents succinct, step-by-step solutions to a variety of real-world
-problems around some topic.
+problems around some topic. A cookbook's code examples serve less to
+enlighten and more to provide quick, paste-ready solutions that the
+reader can apply immediately to the situation facing them.
 
 A Perl cookbook demonstrates ways that all the tools and techniques
 explained elsewhere can work together in order to achieve practical
@@ -622,10 +630,11 @@ mutually similar subpages--lend themselves well to "random access" by
 a reader who knows precisely what knowledge they need, requiring only
 the minimum amount of information before returning to the task at hand.
 
-Perl's own best example of a reference work is C<perlfunc>, the sprawling
-man page that details the operation of every function built into Perl,
-with each function's documentation presenting the same kinds of
-information in the same order as every other.
+Perl's own best example of a reference work is C<perlfunc>, the
+sprawling man page that details the operation of every function built
+into Perl, with each function's documentation presenting the same kinds
+of information in the same order as every other. For an example of a
+shorter reference on a single topic, see C<perlreref>.
 
 Module documentation--including that of all the modules listed in
 L<C<perlmodlib>|perlmodlib>--also counts as reference. They follow
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -65,9 +65,9 @@ documentation style guides of several open-source technologies
 contemporary with Perl. This has included Python, Raku, Rust, and the
 Linux kernel.
 
-The author intends this early draft as a starting place from which to
-launch a review of Perl's reams of extant documentation, with the
-expectation that those conducting this review should grow and modify
+The author intends to see this guide used as starting place from
+which to launch a review of Perl's reams of extant documentation, with
+the expectation that those conducting this review should grow and modify
 this guide as needed to account for the requirements and quirks
 particular to Perl's programming manual.
 
@@ -230,9 +230,10 @@ whether other standard ones listed by C<perlpodstyle>, or ones specific
 to the page's topic; in either case, render these top-level headings in
 all-capital letters.
 
-Further include as many subsections beneath them as needed to make the
-standards of clarity, accessibility, and cross-reference affinity, L<as
-suggested elsewhere in this guide|/Apply one of the four documentation modes>.
+You may then include as many subsections beneath them as needed to meet
+the standards of clarity, accessibility, and cross-reference affinity
+L<suggested elsewhere in this guide|/Apply one of the four documentation
+modes>.
 
 =head3 Author and copyright
 
@@ -706,7 +707,7 @@ underneath that module's namespace that serve only to hold additional
 documentation. For instance, C<Example::Module>'s reference documentation
 might include a see-also link to C<Example::Module::Cookbook>.
 
-Perl's four man pages about Unicode--comprising a short tutorial, a
+Perl's several man pages about Unicode--comprising a short tutorial, a
 thorough explainer, a cookbook, and a FAQ--provide a fine example of
 spreading a complicated topic across several man pages with different
 and clearly indicated purposes.
@@ -723,8 +724,8 @@ Perl's core documentation must recognize this by making as few
 assumptions as possible about the reader's prior knowledge. While you
 should assume that readers of Perl's documentation are smart, curious,
 and eager to learn, you should not confuse this for pre-existing
-knowledge about any other technology, or even programming in general --
-especially in tutorial or introductory material.
+knowledge about any other technology, or even programming in
+general--especially in tutorial or introductory material.
 
 =head3 Keep Perl's documentation about Perl
 
@@ -990,11 +991,11 @@ your own way to express it in your own contributions. Writing clearly,
 succinctly, and with knowledge of your audience's expectations will get
 you most of the way there, in the meantime.
 
-Every line in the docs--whether English sentence or Perl statement --
-should serve the purpose of bringing understanding to the reader. Should
-a sentence exist mainly to make a wry joke that doesn't further the
-reader's knowledge of Perl, set it aside, and consider recasting it into
-a personal blog post or other article instead.
+Every line in the docs--whether English sentence or Perl
+statement--should serve the purpose of bringing understanding to the
+reader. Should a sentence exist mainly to make a wry joke that doesn't
+further the reader's knowledge of Perl, set it aside, and consider
+recasting it into a personal blog post or other article instead.
 
 Write with a light heart, and a miserly hand. 
 
--- a/Porting/perldocstyle.pod
+++ b/Porting/perldocstyle.pod
@@ -28,7 +28,8 @@ the L<C<perlfunc>|perlfunc> file.
 
 =back
 
-This guide will hereafter refer to user-manual section files as I, per Unix convention.
+This guide will hereafter refer to user-manual section files as I<man
+pages>, per Unix convention.
 
 =head2 Purpose of this guide
 
@@ -83,10 +84,11 @@ with Perl itself.
 For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
 For the sake of reading this guide, familiarity with the Pod syntax for
 section headers (C<=head2>, et cetera) and for inline text formatting
-(C<<< CE<lt>like thisE<gt> >>>) should suffice.
+(C<CE<lt>like thisE<gt>>) should suffice.
 
 Perl programmers also use Pod to document their own scripts, libraries,
-and modules. This use of Pod has its own style guide, outlined by L<C<perlpodstyle>|perlpodstyle>.
+and modules. This use of Pod has its own style guide, outlined by
+L<C<perlpodstyle>|perlpodstyle>.
 
 =head2 Choice of language: American English
 
@@ -192,7 +194,7 @@ exhaustive list of every obscure method or alternate syntax available.
 Pages should conclude, when appropriate, with a SEE ALSO section
 containing hyperlinks to relevant sections of Perl's manual, other Unix
 man pages, or appropriate web pages. Hyperlink each such cross-reference via
-C<<< LE<lt>...E<gt> >>>.
+C<LE<lt>...E<gt>>.
 
 What other sections to include depends entirely upon the topic at hand.
 Authors should feel free to include further C<=head1>-level sections,
@@ -262,16 +264,16 @@ so:
 
 =head3 Inline code and literals
 
-Within a paragraph of text, use C<<< CE<lt>...E<gt> >>> when quoting or
+Within a paragraph of text, use C<CE<lt>...E<gt>> when quoting or
 referring to any bit of Perl code--even if it is only one character
 long.
 
 For instance, when referring within an explanatory paragraph to Perl's
-operator for adding two numbers together, you'd write "C<<< CE<lt>+E<gt> >>>".
+operator for adding two numbers together, you'd write "C<CE<lt>+E<gt>>".
 
 =head3 Function names
 
-Use C<<< CE<lt>...E<gt> >>> to render all Perl function names in monospace,
+Use C<CE<lt>...E<gt>> to render all Perl function names in monospace,
 whenever they appear in text.
 
 Unless you need to specifically quote a function call with a list of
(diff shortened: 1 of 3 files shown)


Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1087 @@
...
+B<program>
+
+The most general descriptor for a stand-alone work made out of
+executable Perl code. Synonymous with, and preferable to, "script".
+
+=item *
+
+B<Raku>
+
+Perl's "sister language", and the current open-source language
```
Grinnz: "current" is too time-sensitive for this document. Perhaps:

Perl's "sister language", developed in a project initially known as "Perl 6" led by Larry Wall, Perl's original designer.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1087 @@
...
+L<C<perlmodlib>|perlmodlib>).
+
+=item *
+
+The hundreds of individually presented reference sections derived from
+the L<C<perlfunc>|perlfunc> file.
+
+=back
+
+This guide will hereafter refer to user-manual section files as I, per Unix convention.
```
Grinnz: This line needs wrapping.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1087 @@
...
+Markdown and ReStructuredText, and has a decades-long shared history
+with Perl itself.
+
+For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
+For the sake of reading this guide, familiarity with the Pod syntax for
+section headers (C<=head2>, et cetera) and for inline text formatting
+(C<<< CE<lt>like thisE<gt> >>>) should suffice.
+
+Perl programmers also use Pod to document their own scripts, libraries,
+and modules. This use of Pod has its own style guide, outlined by L<C<perlpodstyle>|perlpodstyle>.
```
Grinnz: This line needs wrapping.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1087 @@
...
+All of Perl's core documentation uses Pod ("Plain Old Documentation"), a
+simple markup language, to format its source text. Pod is similar in
+spirit to other contemporary lightweight markup technologies, such as
+Markdown and ReStructuredText, and has a decades-long shared history
+with Perl itself.
+
+For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
+For the sake of reading this guide, familiarity with the Pod syntax for
+section headers (C<=head2>, et cetera) and for inline text formatting
+(C<<< CE<lt>like thisE<gt> >>>) should suffice.
```
Grinnz: The multiple-bracket escapes are not actually needed when all of the inner brackets are balanced. `C<CE<lt>like thisE<gt>>` is sufficient for this and similar cases.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1087 @@
...
+
+Before 2019, specifying "Perl 5" was sometimes needed to distinguish
+the language from Perl 6. With the latter's renaming to "Raku", this
+practice became unnecessary.
+
+=item *
+
+B<Perl 6>
+
+See L<Raku|"Raku">.
```
haarg: Section links should usually use the form `L<Raku|/Raku>`. Also, `=item *` items are not linkable. To allow linking to an item, it would need to be in the form `=item Raku`.

Grinnz:
An entry should also be added in Porting/README.pod describing this file

Grinnz:
As you suggested on the mailing list, there is the question of whether this belongs better in Porting/ or as installed documentation in pod/. The latter would certainly be more visible but I don't have a compelling reason either way.

jmacdotorg:
RE /Porting vs. /pod, picking the more humble option seemed like a better starting place (and would leave room for it to be "promoted" later).

I too don't feel especially attached to either location. One could argue that it needn't be as user-visible as the rest of pod/, but then again that's where `perlhack` and `perlpolicy` and such live.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1065 @@
...
+by many people. Maintaining consistency across such a wide swath of work
+presents a challenge; this guide provides a foundation to help mitigate
+this difficulty.
+
+This will help its readers--especially those new to Perl--to feel
+more welcome and engaged with Perl's documentation, and this in turn
+will help the Perl project itself grow stronger through having a larger,
+more diverse, and more confident population of knowledgeable users.
+
+=head2 Who this guide is for
```
FGasper: I raised this concern privately but:

Does CMOS not recommend “whom” as the object of a preposition?

Maybe reword to “Who should read this guide” …

apparluk:
I'm against this proposal.

[perlintro.pod](https://github.com/Perl/perl5/blob/blead/pod/perlintro.pod)
DESCRIPTION, paragraph 6, line 4, mentions different styles

The author's commit and linked to rewrite of perlintro.pod, misses item 2 below, or deemphasises item 2.

theoretical method of scientific research
original social-pedagogical experiences
aesthetic assimilation of reality

One could author however a pod/perlhtmlstyle.pod file.

<hr>

○ <a href="https://jmac.org/misc/perldoc/perlintro.pod">Modified <code>perlintro</code></a>

Porting/perldocstyle.pod
jmacdotorg: I've asked the editor I'm working with to weigh in on this one! Will update.

nwc10:
Thanks for your time and effort in drafting this document - it's clear and concise, and from reading it I learned a few things that I wasn't aware that I didn't know.

I spotted a couple of things I'd like to comment on.

In _FORMATTING AND STRUCTURE_

I read the document slowly from the start, and did a double-take at this sentence:

    After its =encoding utf8 declaration, a Perl man page must present

We've been talking about *documentation* up to here, and suddenly we use a different term, without introducing it. *I* know already that the documentation intends to function as man pages (amongst many formats), but I think that the term needs some sort of introduction first.

Which makes me think - at no point does this guide mention that the "documentation" in the core needs to targeted generating several forms out output - traditional Unix man pages, text files, web pages, which imposes some constraints on flexibility.

In _Document Perl's present_

> The docs have no obligation to keep descriptions of past behavior hanging around, even if attaching clauses like "Prior to version 5.10, [...]".

> Perl's docs can acknowledge or discuss former behavior when warranted, but should otherwise keep it in the past.

Yesterday I needed to look at [UV's threading documentation](http://docs.libuv.org/en/v1.x/threading.html#threads)  for the first time.

I liked how they did it - they had 1 line for the "last changed" or "new since":

    uv_thread_options_t
    
        [3 paragraphs]
    
        New in version 1.26.0.
    
    int uv_thread_create(uv_thread_t* tid, uv_thread_cb entry, void* arg)
    
        Changed in version 1.4.1: returns a UV_E* error code on failure
    
    int uv_thread_create_ex(uv_thread_t* tid, const uv_thread_options_t* params, uv_thread_cb entry, void* arg)
    
        [2 paragraphs]
    
        New in version 1.26.0.

This way, as someone new to the information, I could quickly get an idea of how far back code I wrote should work without problems, and if necessary quickly look up the previous behaviour without needing to search for it.

This seems like a good trade off. Without anything, it's effectively an O(n) scan of previous documentation to assess stability. Whereas with all the "previous behaviour" retained in the current document, it becomes O(n) to read this document, because one has to skip or process probably irrelevant details. The approach UV takes is roughly O(1) for both cases.

Which opens the question - how far back? I feel that the best approach is to share the same versions as the CPAN toolchain maintainers actively support, because this reflects the spread of versions in use, and hence a rough estimate of the range of versions people finding documents on the web might actually be targetting, and the range of versions that code deployed to CPAN might end up used on. Treat there lowest version as the baseline, and when they updated their baseline, remove documentation that references versions that now fall before it.

Porting/perldocstyle.pod
jmacdotorg: The editor says:

> Essentially Felipe is correct that the sentence should be "Whom" in this context, and you are correct that it doesn't sound right and is not normally written that way (mostly due to the ending preposition and also the way English audiences interpret what we read). So to quote my favourite copyediting book, "since being correct is only half the goal, it is perhaps best to revise" (The Copyeditor's  Handbook, Amy Einsohn).  
>
>Would the title, "Intended Audience" be too formal?

I think that's a great idea, and I'll make that suggested change.

FGasper: Along this line:

In light of “whom”’s increasingly esoteric status in common usage, it _may_ be worthwhile for the style guide to recommend that “whom” be avoided, and that phrases where it would otherwise be appropriate should be reworded. This very phrase is a case in point.

Grinnz:
I agree - small acknowledgements of what people need to know about past version support can save lots of time and frustration (and I have added such notes to documentation such as POSIX for this reason), and Perl must continue to be pragmatic in its documentation as in its implementation. But as the section currently says, we need not spend paragraphs describing behavior that has been removed, or to note things which do not have a practical relevance (such as the example recently discussed on the mailing list of when a core module was added). These considerations should all be weighed for a particular instance of documentation.

EDIT: Actually this has reminded me that I have not yet sent a PR for the POSIX documentation that I thought I had, oops!

jmacdotorg:
> We've been talking about _documentation_ up to here, and suddenly we use a different term, without introducing it. _I_ know already that the documentation intends to function as man pages (amongst many formats), but I think that the term needs some sort of introduction first.
> 
> Which makes me think - at no point does this guide mention that the "documentation" in the core needs to targeted generating several forms out output - traditional Unix man pages, text files, web pages, which imposes some constraints on flexibility.

This is a great point. I'm rewording the introductory text of that section this way:

```
Each individual work of core Perl documentation, whether contained
within a C<.pod> file or in the Pod section of a standard code
module, patterns its structure after a number of long-time Unix man page
conventions. (Hence this guide's use of "man page" to refer to any one
self-contained part of Perl's documentation.)

Adhering to these conventions help Pod formatters present a Perl man
page's content in different contexts--whether a terminal, the web, or
even print. Many of the following requirements originate with
L<C<perlpodstyle>|perlpodstyle>, which derives its recommendations in
turn from these well-established practices.
```

jmacdotorg:
Regarding "Document Perl's Present": I am renaming the section in question from "Keep the past in the past" to "Recount the past only when necessary", and otherwise softening up the language within to suggest that version-specific information does have a place in docs, when used with care.

Given the lively discussion and debate this topic has generated between this commentary and the p5p mailing list, I would like to keep this section general, with an invitation to add specific advice once we start applying this style guide to "live" Perl documentation. (And I have added a comment in the Pod saying as much.)

xsawyerx:
> Thanks for your time and effort in drafting this document - it's clear and concise, and from reading it I learned a few things that I wasn't aware that I didn't know.
> 
> I spotted a couple of things I'd like to comment on.
> [...]
> In _Document Perl's present_
> 
> > The docs have no obligation to keep descriptions of past behavior hanging around, even if attaching clauses like "Prior to version 5.10, [...]".
> 
> > Perl's docs can acknowledge or discuss former behavior when warranted, but should otherwise keep it in the past.
> 
> Yesterday I needed to look at [UV's threading documentation](http://docs.libuv.org/en/v1.x/threading.html#threads) for the first time.
> 
> I liked how they did it - they had 1 line for the "last changed" or "new since":
> 
> ```
> uv_thread_options_t
> 
>     [3 paragraphs]
> 
>     New in version 1.26.0.
> 
> [...]
> ```
> 
> This way, as someone new to the information, I could quickly get an idea of how far back code I wrote should work without problems, and if necessary quickly look up the previous behaviour without needing to search for it.

I think this might be a nuanced discussion about language vs. API.

* This pattern is very useful for API documentation, but should it be for the language itself?
* I would suggest it for the Perl API. Counter-argument: Publicize `Devel::PPPort` more, since it helps an author from avoiding functions in older versions.

I think a lot of our assumptions coming in are leading us:

* How many new documentation readers do we expect that would still care about version 5.10? (More experienced developers look for reference, not explanations, and might hsave already read the relevent parts since it's not new code.)
* How far back would it be reasonable for a developer to care about? What about a new developer? How many years far back? If something changed more than once, how many of these should be included?
* Does having this extra information not include any cognitive load when reading documentation? (Take into account, the excellent example you gave is for an API, not for a language. A language feature or keyword would include more information, so you would be reading it, not scanning it.)


> [...]
> 
> Which opens the question - how far back? I feel that the best approach is to share the same versions as the CPAN toolchain maintainers actively support, because this reflects the spread of versions in use, and hence a rough estimate of the range of versions people finding documents on the web might actually be targetting, and the range of versions that code deployed to CPAN might end up used on. Treat there lowest version as the baseline, and when they updated their baseline, remove documentation that references versions that now fall before it.

I'm not sure I agree with this, though.

jmacdotorg:
Committed several edits and clarifications suggested by Sawyer X via this man page's Google Doc version (https://docs.google.com/document/d/1Wlfuk13QDFk3kvBfmI0sqmbrnsnE3rKNJDhw2ROul0w/edit?usp=sharing).

jmacdotorg:
I'd like to check in on the status of this PR, and confirm that the team isn't waiting on me for something.

As far as I can tell, here's the state of this request:

* Three Windows-specific automated tests are failing, which don't seem related to the work at hand; I'm not sure there's much I can do about that?
* There is some discussion between @nwc10 and @xsawyerx regarding strategies for documenting version-specific behavior.
    I did respond to @nwc10's initial comments by rewording the style guide's recommendations, per [my comment of November 1](#issuecomment-720133172). I'm not sure I should treat this as an issue that requires further immediate attention, or if we should consider it a non-blocking topic of ongoing interest for this document.

To the best of my knowledge, then, there's no action for me to take right now. Please advise if I'm overlooking something!

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1094 @@
...
+
+This section should also, one way or another, clearly identify the
+audience that the page addresses, especially if it has expectations
+about the reader's prior knowledge. For example, a man page that dives
+deep into the inner workings of Perl's regular expression engine should
+state its assumptions up front--and quickly redirect readers who are
+instead looking for a more basic reference or tutorial.
+
+Reference pages, when appropriate, can precede the DESCRIPTION with a
+SYNOPSIS section that lists, within one more code blocks, some very
```
DrHyde: one *or* more

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1094 @@
...
+brief examples of the referenced feature's use. This section should show
+a handful of common-case and best-practice examples, rather than an
+exhaustive list of every obscure method or alternate syntax available.
```
DrHyde, line 218: Is it worth putting a hard limit on the length and including guidelines about the composition of a SYNOPSIS? Based on https://github.com/tldr-pages/tldr/blob/master/CONTRIBUTING.md, for example.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1094 @@
...
+documentation that L<puts best practices first|/Lead with common cases
+and best practices> should reflect this.
+
+=head2 The documentation speaks with one voice
+
+Even though it comes from many hands and minds, criss-crossing through
+the many years of Perl's lifetime, the language's documentation should
+speak with a single, consistent voice. With few exceptions, the docs
+should avoid explicit first-person statements, or similar self-reference
+to any individual's contributor's philosophies or experiences.
```
DrHyde: This very document has numerous first-person statements in, such as "Certainly we should include
alternate routes and admonitions as reasonably required". I think this should be clarified to say that authors should avoid the first person *singular*.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1094 @@
...
+
+See L<macOS|/macOS>.
+
+=item macOS
+
+Use this term for Apple's operating system instead of "Mac OS X" or
+variants thereof.
+
+This term is also preferable to "Darwin", unless one needs to refer
+to macOS's Unix layer specifically.
```
DrHyde, line 1025: What about MacOS? Is that an acceptable alternative in the middle of a sentence, following the general English rule of names, ie proper nouns, beginning with an upper-case letter? What about capitalising at the beginning of a sentence?

DrHyde: There should probably be a note about Windows, mentioning capitalising to avoid confusion between Windows the operating system and windows the things on the screen. Maybe "MS Windows" or even "Microsoft Windows". And a note about it never being "M$ Windows" or similar silliness?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1094 @@
...
+For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
+For the sake of reading this guide, familiarity with the Pod syntax for
+section headers (C<=head2>, et cetera) and for inline text formatting
+(C<CE<lt>like thisE<gt>>) should suffice.
+
+Perl programmers also use Pod to document their own scripts, libraries,
+and modules. This use of Pod has its own style guide, outlined by
+L<C<perlpodstyle>|perlpodstyle>.
+
+=head2 Choice of language: American English
```
DrHyde, line 93: But be tolerant about spelling and other minor deviations when accepting pull requests. Instead of bothering a contributor about such things, the doco team should make such minor fix-ups themselves when accepting a contribution.

It is *really hard* to deliberately mis-spell words like "color" for those of us whose native tongue isn't American English.

Porting/perldocstyle.pod
Grinnz: The name of the OS is "macOS" - "MacOS" is a little too close to "Mac OS" which is the name of their older OS. Yes, the names are terrible.

Porting/perldocstyle.pod
jmacdotorg: I would argue that the present text -- "no more than a handful", not "an exhaustive list" -- gets across the spirit of this, without needing to set an arbitrary objective limit.

Porting/perldocstyle.pod
jmacdotorg: Yep, it's "macOS", just like "iPhone", and the same (eye-rolling but trademarked) spelling rules apply.

As for Windows et al, I certainly expect this section to grow quite a bit, especially once the Great Work of reviewing and revising the present core docs gets underway. What's here is just a stub, based on the little bit of test revision I did for this project, plus the computer equipment I typed it all in on. :)

Porting/perldocstyle.pod
jmacdotorg: Many have taken issue with the document's declaration of preferring American English, which clearly came across as much blunter than I intended.

I just added a commit which expands the first paragraph of this section this way:

```
Perl's core documentation is written in English, with a preference for
American spelling of words and expression of phrases. That means "color"
over "colour", "math" versus "maths", "the team has decided" and not
"the team have decided", and so on.

We name one style of English for the sake of consistency across Perl's
documentation, much as a software project might declare a four-space
indentation standard--even when that doesn't affect how well the code
compiles. Both efforts result in an easier read by avoiding jarring,
mid-document changes in format or style.

Contributors to Perl's documentation should note that this rule
describes the ultimate, published output of the project, and does not
prescribe the dialect used within community contributions. The
documentation team enthusiastically welcomes any English-language
contributions, and will actively assist in Americanizing spelling and
style when warranted.
```

DrHyde: 👍

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1115 @@
...
+
+Programmers documenting their own work apart from the Perl project
+itself may also find this guide worthwhile, especially if they wish
+their work to extend the tone and style of Perl's own manual.
+
+=head2 Status of this document
+
+This guide was initially drafted in late 2020, drawing from the
+documentation style guides of several open-source technologies
+contemporary with Perl. This has included Python, Raku, Rust, and the
```
khwilliamson, line 65: I'm unclear that this use of contemporary is proper.  contemporaneous comes to mind, but is unwieldy

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1115 @@
...
+
+=back
+
+=head1 AUTHOR
+
+This guide was initially drafted by Jason McIntosh
+(jmac@jmac.org), under a grant from The Perl Foundation.
+
+=for comment Additional contributors can get listed here (and this
+comment deleted), when there are some.
```
khwilliamson, line 1118: I  waited for the contents of this PR to settle down before reading it myself.  Now that I have, I want to say that I was impressed by the, for want of a better word, professionalism in the writing and tone.  It serves as an excellent example of what it espouses; no hypocrisy here.  The reason I started using Perl was that it was the only thing available at the time that was significantly better than awk, sed, grep and ksh (which itself was signficantly better than the Bourne shell).  But I was captivated by the style in "Programming Perl", and I want our documentation to continue that, speaking in an encouraging and friendly way as someone who happens to know more about this subject than the reader currently does.

Where did the 72 column number come from?
I have nothing really to say about the contents.
Near wars have been fought over whether sentences should have one or two spaces in the input text after the final punctuation (usually a period), although, to my knowledge, no lives have been lost as a result.  There was a fervent discussion some years ago on p5p, the bottom line of which allowed me to continue to use two spaces.  I'm guessing that CMOS specifies a single space.  If so, I'm confident they are wrong, and I suspect that there were not enough people involved who understood the special needs of technical writing, nor the physiology of human eye tracking.

I was pleased that the text uses a comma before the 'and final item in a list'.  There has been a trend, including the New York Times switching over a while back, to omit that.  Those people aren't thinking about the precision needed to express unambiguously lists and sublists

jkeenan:
Since we're due for a monthly development release on Sunday, Dec 20, let's time-box any further evaluation of this pull request.

If anyone (particularly anyone with a commit bit) objects to merging this file into blead **at all**, please comment in this ticket by Sat Dec 19 12:00:00 UTC 2020 (7:00 am America/NewYork on that date).  If I hear no absolute objection and will merge into blead in advance of 5.33.5 release.  If someone thinks this document should not go into blead **at all**, please register that objection and your rationale ASAP.

I am self-assigning this ticket for the purpose of conducting that evaluation/merge this Saturday.

If you have just minor differences about the wording of this document but don't object to its entering the core distribution, please raise them in this ticket and we'll sort them out after the 5.33.5 release.

Thank you very much.
Jim Keenan

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1115 @@
...
+guide discusses, the ideal Perl documentation manages to deliver its
+message clearly and concisely, and not confuse mere wordiness for
+completeness.
+
+=head3 Further notes on documentation modes
+
+Keep in mind that the purpose of this categorization is not to dictate
+content--a very thorough explainer might contain short reference
+sections of its own, for example, or a reference page about a very
+complex function might resemble an explainer in places (c.f. [open]).
```
jkeenan: That abbreviation should be *cf.* rather than *c.f.*, as it's short for the Latin *confer* (https://en.wikipedia.org/wiki/Cf.).  However, I think the better term here would be *e.g.*, as you're providing an example.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1115 @@
...
+
+Outside of its reference pages, most of Perl's manual belongs to this
+mode. This includes the majority of the man pages whose names start with
+"C<perl>". A fine example is C<perlsyn>, the Perl Syntax page, which
+explores the whys and wherefores of Perl's unique syntax in a
+wide-ranging discussion laden with many references to the language's
+history, culture, and driving philosophies.
+
+Perl's explainer pages give authors a chance to explore Perl's penchant
+for TMTOWTDI, illustrating alternate and even obscure ways to use the
```
jkeenan: If *TMTOWTDI* is used only once in this particular document, perhaps a link should be provided to its definition in either perlglossary or perlfaq1.

jmacdotorg: I didn't even know about perglossary--thanks!

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+their work to extend the tone and style of Perl's own manual.
+
+=head2 Status of this document
+
+This guide was initially drafted in late 2020, drawing from the
+documentation style guides of several open-source technologies
+contemporary with Perl. This has included Python, Raku, Rust, and the
+Linux kernel.
+
+The author intends this early draft as a starting place from which to
```
FGasper: Once this merges, is it still an “early draft”?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+=head2 Choice of markup: Pod
+
+All of Perl's core documentation uses Pod ("Plain Old Documentation"), a
+simple markup language, to format its source text. Pod is similar in
+spirit to other contemporary lightweight markup technologies, such as
+Markdown and reStructuredText, and has a decades-long shared history
+with Perl itself.
+
+For a comprehensive reference to Pod syntax, see L<C<perlpod>|perlpod>.
```
FGasper, line 84: Why the C<> within the L<>?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+(Note that the Chinese, Japanese, and Korean-language README files
+included with Perl's source distributions provide an exception to this
+choice of language--but these documents fall outside the scope of this
+guide.)
+
+=head2 Choice of encoding: UTF-8
+
+Perl's core documentation files are encoded in UTF-8, and can make use
+of the full range of characters this encoding allows.
```
FGasper, line 131: Consider adding “… except as this document may restrict use of, e.g., curly quotes or dash characters.”

You might offer one or two specifically-whitelisted characters. Given that quotes, apostrophes, and dashes are all to be ASCII, and that the docs are to be in English, _are_ there any non-ASCII characters that should see regular use in these documents?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+As such, every core doc file (or the Pod section of every core module)
+should commence with an C<=encoding utf8> declaration.
+
+=head2 Choice of underlying style guide: CMOS
+
+Perl's documentation uses the L<Chicago Manual of
+Style|https://www.chicagomanualofstyle.org> (CMOS), 17th Edition, as
```
FGasper, line 139: What happens when a new CMOS edition is published? Will the docs team decide whether it wants to update its standard as each new edition arrives?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+summary of, or introduction to, the document's content and purpose.
+
+This section should also, one way or another, clearly identify the
+audience that the page addresses, especially if it has expectations
+about the reader's prior knowledge. For example, a man page that dives
+deep into the inner workings of Perl's regular expression engine should
+state its assumptions up front--and quickly redirect readers who are
+instead looking for a more basic reference or tutorial.
+
+Reference pages, when appropriate, can precede the DESCRIPTION with a
```
FGasper, line 214: Is “precede” the correct verb here? I usually use “precede” to indicate that the subject itself “goes before” (e.g., “2019 preceded 2020”) rather than that the subject _arranges_ something else “before”, which is what this seems to say.

Maybe “prefix”?

-----
That said … should the SYNOPSIS perhaps be described before the DESCRIPTION (i.e., instead of after)? The overall flow of this section seems to be “first things first”, but we’re inverting that flow here. Seems a bit awkward.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+man pages, or appropriate web pages. Hyperlink each such cross-reference via
+C<LE<lt>...E<gt>>.
+
+What other sections to include depends entirely upon the topic at hand.
+Authors should feel free to include further C<=head1>-level sections,
+whether other standard ones listed by C<perlpodstyle>, or ones specific
+to the page's topic; in either case, render these top-level headings in
+all-capital letters.
+
+Further include as many subsections beneath them as needed to make the
```
FGasper: “Further” seems redundant here. <shrug>

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+Note that these guidelines do not apply to the core software modules
+that ship with Perl. These have their own standards for authorship and
+copyright statements, as found in C<perlpodstyle>.
+
+=head2 Formatting rules
+
+=head3 Line length and line wrap
+
+Each line within a Perl man page's Pod source file should measure 72
+characters or fewer in length.
```
FGasper, line 262: It might do to justify the number 72 here. Are we accommodating old-school terminals, following some other convention, … ?

I personally prefer an 80-column max for code, but it’s worth noting that the Linux kernel just upped that to (I think?) 100.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+=head3 Function names
+
+Use C<CE<lt>...E<gt>> to render all Perl function names in monospace,
+whenever they appear in text.
+
+Unless you need to specifically quote a function call with a list of
+arguments, do not follow a function's name in text with a pair of empty
+parentheses. That is, when referring in general to Perl's C<print>
+function, write it as "C<print>", not "C<print()>".
```
FGasper, line 315: Does this rule also apply to non-builtins? When referring, for example, to Scalar::Util’s `weaken` function, should I write `weaken()` or `weaken`?

Likewise, what if I write about my Sys::Pipe module’s `pipe` function?

FWIW I personally find the parens helpful to disambiguate when I’m talking about “print” as a verb versus the `print()` function/builtin. I don’t get why many prefer them omitted.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+In Pod source, use straight quotes, and not "curly quotes":  "Like
+ this", not “like this”. The same goes for apostrophes:  Here's a
+ positive example, and here’s a negative one.
+
+Render em dashes as two hyphens--like this:
+
+    Render em dashes as two hyphens--like this.
+
+Leave it up to formatters to reformat and reshape these punctuation
+marks as best fits their respective target media.
```
FGasper, line 379: The above two sections seems to contradict the note above that documentation “can make use of the full range of characters [UTF-8] allows”.

It’s also notable that “perlunicook”—maybe unsurprisingly—uses literal curlies and dashes (both m- and n-dashes).

Porting/perldocstyle.pod
FGasper, line 131: _Strictly_ speaking, I think UTF-8 can be said to “allow for” characters that exceed the Unicode range. Should this document be more specific?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+list. (And because it can contain any number of variables, it must be
+the I<last> argument, when present.)
+
+=back
+
+When possible, give scalar arguments names that suggest their purpose
+among the arguments. See, for example, L<C<substr>'s
+documentation|perlfunc/substr>, whose
+listed arguments include C<EXPR>, C<OFFSET>, C<LENGTH>, and C<REPLACEMENT>.
+
```
FGasper, line 367: Should there be a section about use of L<> to refer to modules?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+marks as best fits their respective target media.
+
+=head3 Unix programs and C functions
+
+When referring to a Unix program or C function with its own man page
+(outside of Perl's documentation), include its manual section number in
+parentheses. For example: C<malloc(3)>, or C<mkdir(1)>.
+
+If mentioning this program for the first time within a man page or
+section, make it a cross reference, e.g. C<LE<lt>malloc(3)E<gt>>.
```
FGasper, line 388: It seems a lesser offense to have multiple L<>s than to have all C<>s and no L<>s. Thus, maybe format this section so:
> … (outside of Perl's documentation), refer to it via a cross-reference with its manual section number in parentheses. For example: C<LE<lt>malloc(3)E<gt>>

If a given man page or section refers to the same program or function multiple times, use plain monospace for subsequent references, e.g., C<CE<lt>mkdir(1)E<gt>>.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+The first item of the list should be immediately followed by one or
+more C<XE<lt>...E<gt>> terms listing index-worthy topics; if nothing
+else, then the name of the function, with no arguments.
+
+=item 2.
+
+A C<=for> line, directed at C<Pod::Functions>, containing a one-line
+description of what the function does. This is written as a phrase, led
+with an imperative verb, with neither leading capitalization nor ending
```
FGasper, line 528: Quibble: consider “verb in imperative form”, insofar as “quoted” and “quote” are the same verb in different forms.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+=head3 Cross-references and hyperlinks
+
+Make generous use of Pod's C<LE<lt>...E<gt>> syntax to create hyperlinks
+to other parts of the current man page, or to other documents entirely
+-- whether elsewhere on the reader's computer, or somewhere on the
+internet, via URL.
+
+Use C<LE<lt>...E<gt>> to link to another section of the current man page
+when mentioning it, and make use of its page-and-section syntax to link to
```
FGasper, line 400: The section designator is vulnerable to breakage if a heading is ever renamed. Should there be a note about ensuring, before renaming a section, that no links refer to it by name?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+it up into separate pages. That may mean creating a new "C<perl[...]>"
+man page, or (in the case of module documentation) making new packages
+underneath that module's namespace that serve only to hold additional
+documentation. For instance, C<Example::Module>'s reference documentation
+might include a see-also link to C<Example::Module::Cookbook>.
+
+Perl's four man pages about Unicode--comprising a short tutorial, a
+thorough explainer, a cookbook, and a FAQ--provide a fine example of
+spreading a complicated topic across several man pages with different
+and clearly indicated purposes.
```
FGasper, line 713: What if a fifth one is added? This paragraph will go stale, and given the size of this doc it’s not unlikely that someone will neglect to s/four/five/. Could generalize this to:
> Perl's several man pages about Unicode provide a fine example of spreading a complicated topic across several man pages with different and clearly indicated purposes.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+Perl has grown a great deal from its humble beginnings as a tool for
+people already well versed in C programming and various Unix utilities.
+Today, a person learning Perl might come from any social or
+technological background, with a range of possible motivations
+stretching far beyond system administration.
+
+Perl's core documentation must recognize this by making as few
+assumptions as possible about the reader's prior knowledge. While you
+should assume that readers of Perl's documentation are smart, curious,
+and eager to learn, you should not confuse this for pre-existing
```
FGasper, line 726: “confuse this with” is the usage I’m used to seeing. Is “for” intended?

Maybe “mistake this for” would be stronger?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+
+For example, Perl loves working with filehandles, and as such that word
+appears throughout its documentation. A new Perl programmer arriving at
+a man page for the first time is quite likely to have no idea what a
+"filehandle" is, though. Any Perl man page mentioning filehandles
+should, at the very least, hyperlink that term to an explanation
+elsewhere in Perl's documentation. If appropriate--for example, in the
+lead-in to L<C<open> function's detailed reference|perlfunc/open
+FILEHANDLE,MODE,EXPR>--it can also include a very short in-place
+definition of the concept for the reader's convenience.
```
FGasper, line 769: > Any Perl man page mentioning filehandles
should, at the very least, hyperlink that term to an explanation
elsewhere in Perl's documentation.

Read at face value, this seems to indict practically all existing documentation. When I pull up `perldoc -f open` I see no explanation of what a filehandle is.

Should this be softened somewhat? To be honest I’m not sure what idea is trying to be conveyed here; to a Perl programmer I wouldn’t consider “filehandle” to be jargon-y any more than “string” is to a JavaScript programmer. Admittedly, it’s kind of a gray area.

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+not leaning so hard into overt comedy as to distract and confuse from
+the nonetheless-technical topics at hand.
+
+Like the best written works, Perl's documentation has a soul. Get
+familiar with it as a reader to internalize its voice, and then find
+your own way to express it in your own contributions. Writing clearly,
+succinctly, and with knowledge of your audience's expectations will get
+you most of the way there, in the meantime.
+
+Every line in the docs--whether English sentence or Perl statement --
```
FGasper: Quibble: spaces around the latter double-dash, but none around the former … is that discrepancy intended?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+The name of the programming language is Perl, with a leading capital
+"P", and the remainder in lowercase. (Never "PERL".)
+
+The interpreter program that reads and executes Perl code is named
+"C<perl>", in lowercase and in monospace (as with any other command
+name).
+
+Generally, unless you are specifically writing about the
+command-line C<perl> progam (as, for example, L<C<perlrun>|perlrun>
+does), use "Perl" instead.
```
FGasper, line 1042: Should this section mention “perls” in reference to different builds of the Perl interpreter?

Porting/perldocstyle.pod
```diff
@@ -0,0 +1,1117 @@
...
+parentheses. That is, when referring in general to Perl's C<print>
+function, write it as "C<print>", not "C<print()>".
+
+=head3 Function arguments
+
+Represent functions' expected arguments in all-caps, with no sigils, and
+using C<CE<lt>...E<gt>> to render them in monospace. These arguments
+should have short names making their nature and purpose clear.
+Convention specifies a few ones commonly seen throughout Perl's
+documentation:
```
FGasper, line 323: While I recognize that Perl’s “classical” documentation style looks thus, consider the following advantages of notating functions using sigils instead:

1. It “maps” nicely onto function signatures, a feature which, while still nominally experimental, is coming into greater use in projects like Mojolicious and cPanel.

2. It provides a nicer at-a-glance overview of the purpose of the variables; in addition to their type, you can also give them a name, up-front.

3. It can be actual Perl code, which has obvious benefits.

4. It correlates to the tradition in Unix man pages of giving the variable type as well as a reference name.

It may be out of scope for this document, but personally I’d like to see consideration of an en-masse migration to using sigils in function arguments.

Porting/perldocstyle.pod
FGasper, line 367: Should the section on function names also give a suggested way to name/describe object methods and class methods? I personally prefer:
```
I<OBJ>->mymethod( … )
```

Porting/perldocstyle.pod
jmacdotorg: Sounds good to me, honestly!

Porting/perldocstyle.pod
jmacdotorg: This is a POD idiom found throughout Perl's extant documentation, for when function, command, or other literal term requires both monospace styling and a hyperlink.

Porting/perldocstyle.pod
jmacdotorg: This is an excellent point! I'll tone that down.

Porting/perldocstyle.pod
FGasper: Cool. I suppose the docs team would have as part of its “internal charter”, then, that new CMOS editions would be evaluated as they come out.

Porting/perldocstyle.pod
jmacdotorg: Mm, I feel cautious about piling rules upon rules here. This section is just about encoding, which is UTF-8, the end. Let the presence of beyond-ASCII characters be evaluated case-by-case after that. 💃

Porting/perldocstyle.pod
jmacdotorg: You're right that "precede" by itself is not a transitive verb, but "precede [something] with" makes it so. (Or so confirms my macOS Dictionary, anyway.) But good point about the section ordering in any event; I'll take a swing at it.

Porting/perldocstyle.pod
FGasper: After I read this paragraph I assumed that I could use curly-quotes, which I’m personally used to typing. My idea of adding a generic disclaimer is to prevent people from having the same misunderstanding that I had.

Porting/perldocstyle.pod
jmacdotorg: ("Justification", ha ha.)

The literal answer is that I'm just following in the footsteps of existing Pod-checking tools, and standards that porters suggest I adopt when I did my doc-updating work earlier in the year.

Getting the deeper reason why would be a small research project, and I'm making a note of it for a potential project-level to-do list.

Porting/perldocstyle.pod
jmacdotorg: The text styling should help to disambiguate English verb from function name.

As for parens/non-parens, this is me following apparent precedent set down by existing core documentation, especially `perlfunc`, though the whole collection of Perl documentation is very much of two minds about this.

I do think there should be a standard here, one way or the other. I do prefer the less verbose way, using monospace styling instead of `()` to signal "this is a function". It's less verbose--and it avoids the uncomfortable implication of that the function in question can always be called with no arguments.

Porting/perldocstyle.pod
jmacdotorg: I don't find a contradiction, per previous comments.

My advice against smart quotes and long dashes was based on precedent in extant documentation, so finding a live exception is interesting. I do think that we need a standard here, I'd like to see how other technology-specific style guides handle it. (I expect that "use 'dumb quotes' and double-dashes and let formatters fix it" will continue to be OK.)

Porting/perldocstyle.pod
jmacdotorg: Happily, a style guide needn't patch every possible loophole, and this is one of those times.

Porting/perldocstyle.pod
jmacdotorg: Care to write one and find out? :)

Porting/perldocstyle.pod
jmacdotorg: The first two sentences of `open`'s docs, as of 5.32:

> Associates an internal FILEHANDLE with the external file specified by EXPR. That filehandle will subsequently allow you to perform I/O operations on that file, such as reading from it or writing to it.

Admittedly I'm biased to use this as an example, because I wrote said sentences myself.

I honestly do think "filehandle" is Perlish jargon! According to my own research it's just not a word that comes up in many-if-any other contemporary programming languages, unlike "string".

Porting/perldocstyle.pod
jmacdotorg: This really does sound like a proposal of a solid, early research project of the documentation team, once it's founded, sometime after this doc gets merged. Make a note of it!

Porting/perldocstyle.pod
FGasper: FWIW I’ve never made that assumption—that a function with empty-parens can implicitly be called without arguments—in Perl any more than I’ve made it in C, where man pages frequently do follow function names with empty parens.

Porting/perldocstyle.pod
FGasper: I didn’t read that in `perldoc -f open` as an explanation of what filehandles are in general so much as a description of what that _specific_ filehandle does. I can see it the other way, though. Regardless, there’s no hyperlink, which runs afoul of the literal meaning of the phrase in question, no?

Porting/perldocstyle.pod
FGasper: Sure, I’ll send something. Better before or after this merges?

Porting/perldocstyle.pod
FGasper: If you say “you can use the full range of characters”, doesn’t that literally imply that every character is acceptable?

If, then, you start saying that characters X, Y, and Z are forbidden, then “every character is acceptable” becomes a false statement … right?

Porting/perldocstyle.pod
DrHyde: There is certainly a place for non-ASCII characters in documentation written in English, if you are writing something like (copied/pasted from the doco for one of my modules) ...

```
... the 3 character string "北京市" is stored as 9 bytes:

        北: 0xE5 0x8C 0x97    京: 0xE4 0xBA 0xAC    市: 0xE5 0xB8 0x82
```

As an aside, how does perldoc cope with text that changes direction - if, for example, I had picked القُدس for my example instead of 北京市?

Porting/perldocstyle.pod
jmacdotorg: Since you're not the first person to question this, let's consider it another research point for the near future. (Albeit a fairly simple one, I expect... a matter of seeing how other project style guides tend to handle the question, and doing whatever that is, adapted for Perl.)

Porting/perldocstyle.pod
jmacdotorg: After, please, per reviewer comments.

jmacdotorg:
General comment to @FGasper: I've just made several small changes and fixes based on a handful of your suggestions. Thank you! Per @jkeenan's comment (https://github.com/Perl/perl5/pull/18275#issuecomment-745610312), I'm leaving other style questions as things to discuss further or otherwise resolve post-merge.

khwilliamson:
FWIW, I agree that this is certainly good enough should be merged, and then subject to further tweaks

Porting/perldocstyle.pod
DrHyde: Monospace styling doesn't help when it's all monospaced, in a terminal, reading a manpage.

jkeenan:
@jmacdotorg , I was just about to merge this p.r. into blead when I realized that you've got the main file in the wrong directory. perldocstyle.pod should be in `pod/`, not in `Porting/'.  And once that correction is made, we no longer need a new entry in `Porting/README.pod`, since we're no longer adding new files to that directory.

In the very first post to this p.r., you make reference to your project proposal in https://jmac.org/misc/perldoc/.  In that document, you write:
_The Pod edition of this document is intended for addition to the Perl source distribution–either in /pod or in /Porting._
Apparently no one ever pointed out that the new doc was to go into `pod/`.

So, is there any reason why `perldocstyle.pod` should go into `Porting/` rather than `pod/`?

I will prepare the merge in what I believe to be the correct way but hold off hitting the button until I hear from you.

Thank you very much.
Jim Keenan

jkeenan:
> @jmacdotorg , I was just about to merge this p.r. into blead when I realized that you've got the main file in the wrong directory. perldocstyle.pod should be in `pod/`, not in `Porting/'. And once that correction is made, we no longer need a new entry in `Porting/README.pod`, since we're no longer adding new files to that directory.
> 
> In the very first post to this p.r., you make reference to your project proposal in https://jmac.org/misc/perldoc/. In that document, you write:
> _The Pod edition of this document is intended for addition to the Perl source distribution–either in /pod or in /Porting._
> Apparently no one ever pointed out that the new doc was to go into `pod/`.
> 
> So, is there any reason why `perldocstyle.pod` should go into `Porting/` rather than `pod/`?
> 
> I will prepare the merge in what I believe to be the correct way but hold off hitting the button until I hear from you.
> 
> Thank you very much.
> Jim Keenan

It turns out that there is **a lot** of meta-work needed to add a new `pod/` file and get `make test_porting` to PASS.  What I decided to do was to extract `perldocstyle.pod` from this pull request and make it the subject of a new p.r. in which it was correctly positioned underneath `pod/` and in which all the meta-work needed was completed.

Branch: [smoke-me/jkeenan/jmacdotorg-perldocstyle](https://github.com/Perl/perl5/tree/smoke-me/jkeenan/jmacdotorg-perldocstyle)

My plan now is to merge https://github.com/Perl/perl5/pull/18412 into blead as soon as I get a few smoke-test reports back.  That will supersede this p.r., which I will then close.

Thank you very much.
Jim Keenan

Pull request:  https://github.com/Perl/perl5/pull/18412

jkeenan:
Superseded by https://github.com/Perl/perl5/pull/18412, which was merged into blead in commit 4fb67a9778bdd1bae111bb2b1028c36d74287bb4.

Any further patches or pull requests should be drawn against blead.

Thank you very much.
Jim Keenan

PR #18277, opened 2020-11-01
Fix typo in perlguts

cornett:


jkeenan:
Merged to blead with commit adding contributor to AUTHORS.

Thank you very much.
Jim Keenan

PR #18240, opened 2020-10-14
perlxstut add links and update examples

ghost:
For https://github.com/Perl/perl5/issues/18163.

PR #18278, opened 2020-11-02
xstut.pod changes from @apparluk

khwilliamson:
This PR replaces https://github.com/Perl/perl5/pull/18240

### DIFF
--- a/dist/ExtUtils-ParseXS/lib/perlxstut.pod
+++ b/dist/ExtUtils-ParseXS/lib/perlxstut.pod
@@ -120,16 +120,16 @@ The file Makefile.PL should look something like this:
     WriteMakefile(
 	NAME         => 'Mytest',
 	VERSION_FROM => 'Mytest.pm', # finds $VERSION
-	LIBS         => [''],   # e.g., '-lm'
-	DEFINE       => '',     # e.g., '-DHAVE_SOMETHING'
-	INC          => '',     # e.g., '-I/usr/include/other'
+	LIBS         => [''], # e.g., '-lm'
+	DEFINE       => '', # e.g., '-DHAVE_SOMETHING'
+	INC          => '-I.', # e.g., '-I. -I/usr/include/other'
     );
 
 The file Mytest.pm should start with something like this:
 
     package Mytest;
 
-    use 5.008008;
+    use 5.017007;
     use strict;
     use warnings;
 
@@ -276,9 +276,9 @@ when the test is correct, "not ok" when it is not.
     # so read its man page ( perldoc Test::More ) for help writing this
     # test script.
 
-    is(&Mytest::is_even(0), 1);
-    is(&Mytest::is_even(1), 0);
-    is(&Mytest::is_even(2), 1);
+    is( Mytest::is_even(0), 1 );
+    is( Mytest::is_even(1), 0 );
+    is( Mytest::is_even(2), 1 );
 
 We will be calling the test script through the command "C<make test>".  You
 should see output that looks something like this:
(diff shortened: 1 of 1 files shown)


dist/ExtUtils-ParseXS/lib/perlxstut.pod
```diff
@@ -120,16 +120,16 @@ The file Makefile.PL should look something like this:
...
+	DEFINE       => '', # e.g., '-DHAVE_SOMETHING'
+	INC          => '-I.', # e.g., '-I. -I/usr/include/other'
     );
 
 The file Mytest.pm should start with something like this:
 
     package Mytest;
 
-    use 5.008008;
+    use 5.017007;
```
xsawyerx, line 132: I suggest we make this 5.020000. I don't think it's very likely that someone will use 5.17.1 to 5.17.6 when trying out the XS tutorial.

Objections?

xsawyerx (commented):
Otherwise, I'm good with this.

dist/ExtUtils-ParseXS/lib/perlxstut.pod
apparluk: > I suggest we make this 5.020000.

I concur.

wchristian: The use perl version statement should accurately reflect the minimum perl version required to make the code run correctly.

ghost: https://www.cpan.org/src/README.html

```
Major   Version   Released
5.20    5.20.0    2014-05-27
5.18    5.18.0    2013-05-18
```

<hr>

Copied from the replaced PR.

The define `PERL_NO_GET_CONTEXT` was added to `h2xs.PL`on Dec 6, 2012. 

The define `PERL_NO_GET_CONTEXT` was added to ParseXS's pods on Feb 16, 2014.

Perl `5.008008`  could have been updated to `5.017007` concomitant with either of those commits.

I used 5.017007 in the .pod believing PERL_NO_GET_CONTEXT first appeared in that release.

However it would be unusual to expose the reader to a development release of perl in the .pod, and maybe using instead the >5.017007 next stable perl version would be more sensible.

haarg: `PERL_NO_GET_CONTEXT` provides an optimization. If it is used on earlier perl versions, it just has no impact. There is no reason to change the required version due to it.

ghost: EXAMPLE 1

The reader is instructed to run "`h2xs -A -n Mytest`".

A rendering is presented of that command's generated Mytest.xs.

The rendering contains `#define PERL_NO_GET_CONTEXT`.

Factually if the define first appeared in 5.17.7 then having `5.008008` in the rendered Mytest.pm would not be consistent with what the reader would see in the command's generated file.

wchristian: The prose clearly says: (emphasis mine)

> Finally, the Mytest.xs file should look **something like this**:

There is no requirement for non-function-specific details to be identical, and this being a tutorial it would in fact even be better to: Explain in prose that `PERL_NO_GET_CONTEXT` is an optimization which can show up but is not required for function; and leave the version requirement low.

ghost:
EXAMPLE 1

It's not technically possible to run "`h2xs -A -n Mytest`" and get output with the define prior to a specific Perl version number. As of the current state of the Perl source code it is possible to run that command and get output in Mytest.pm with a newer Perl version number than that number where the define was added, and which also includes that define in Mytest.pm.

> The define PERL_NO_GET_CONTEXT was added to ParseXS's pods on Feb 16, 2014.

The above commit feasibly was made when there may have been coders who still didn't trust threads, consequently compiling Perl without `USE_THREADS="-Dusethreads -Duseithreads"`.

wchristian:
What did you intend to communicate with that comment you just posted?

It would behoove you to try and communicate and explain your intents more clearly when you post comments to issues here.

ghost:
_cf_: https://github.com/Perl/perl5/pull/18278#discussion_r517737700

```diff
+ It is now maintained as part of Perl itself.

  =head2 Last Changed
```

Noting the addition of this line has already been approved by 2 Members, it would be advisable that the merge include an upgrade of the perl version number to a release subsequent to the below commit.

https://github.com/Perl/perl5/commit/3ee084aa8474e0a47583aee63a88beeea5d4742f#diff-1c8d7086354277282e98495f1ff54ef44424e5bbb73f9010cd6149ee771a2c63

The .pod should be accessible to a broad audience.

Grinnz:
Approval by two members does not override the need to address concerns.

wchristian:
@apparluk I asked you to explain [this comment](https://github.com/Perl/perl5/pull/18278#issuecomment-722686181), and i will ask again, because you did not provide any explanation:

What did you intend to communicate with that comment you just posted?

Use your own words in clear prose, do not link to anything else in lieu of writing your own words.

apparluk:
_cf_: https://github.com/Perl/perl5/pull/18278#issuecomment-722686181

For the reader a dissimilarity between the online documentation and output files in the former might be more a matter for concern than in the latter. 

I'm running Perl 5.14.4 why does the documentation show a define not in my Mytest.xs file?

```
bash-4.3$ lynx -dump https://perldoc.perl.org/perlxstut \
> |  perl -ne 'print if $. eq 222 .. 229'
#define PERL_NO_GET_CONTEXT
#include "EXTERN.h"
#include "perl.h"
#include "XSUB.h"

#include "ppport.h"

MODULE = Mytest             PACKAGE = Mytest
bash-4.3$ 
```

[with version upgrade to 5.018000]

I'm running Perl 5.32.0 the documentation shows 5.018000 in Mytest.pm so why does mine read 5.032000?

`lynx -dump https://perldoc.perl.org/perlxstut`

<hr>

The  other content refers to extraneous circumstance at the time of the commit.

khwilliamson:
I looked at this P.R. rebased to latest blead, and most of its proposed changes have otherwise been incorporated.  The few that remain are inconsequentially different.  So I'm closing it.

PR #18247, opened 2020-10-17
replace rpc for tirpc in perlxs.pod

ghost:
For https://github.com/Perl/perl5/issues/18246.

Run rpctest.pl and verify the correctness of the output before merging.

PR #18279, opened 2020-11-02
perlxs changes from @apparluk

khwilliamson:
This replaces PR #18247


xsawyerx (approved):

ghost:
Run rpctest.pl and verify the correctness of the output before merging.

tonycoz:
This is a Linux specific change, non-glibc systems still have the headers visible as `rpc/rpc.h` etc (I checked FreeBSD 13 and OpenIndiana).

khwilliamson:
So does Linux still work with the old scheme, or do we need a Configure probe?

Grinnz:
This is just documentation, no functionality needs to change.

khwilliamson:
Note that this commit is now updated

PR #18285, opened 2020-11-03
perlapi: Document typedef IO

khwilliamson:


pod/perlapio.pod
```diff
@@ -69,52 +69,9 @@ perlapio - perl's IO abstraction interface.
...
-=for apidoc PerlIO_seek
-=for apidoc PerlIO_get_cnt
-=for apidoc PerlIO_get_ptr
-=for apidoc PerlIO_set_ptrcnt
-=for apidoc PerlIO_set_cnt
-=for apidoc PerlIO_get_base
-=for apidoc PerlIO_get_bufsiz
+=for apidoc Ay|        |IO
+An internal Input/Output (C<I/O>) object to process input from, or
+or output to, a file or device.
```
tonycoz: I'm not sure that's a good characterization, they're about as internal as any other SV variant other than INVLIST (which I believe is meant to be internal/re.pm only)

As with other SV types I believe as long as you use the right macros (the Io*() macros for IO) or functions you're following the API.

khwilliamson: The problem is I can't find a public macro or fcn that takes such a parameter.  sv_2io returns one, but then what can you do with it?

khwilliamson: Also GvIO and GvIOp return this, but again what can be done with it?

tonycoz: You can fetch the input and output PerlIO * handles with IoIFP(io) and IoOFP(io), which are used in the standard typemap in `lib/ExtUtils/typemap` and in many CPAN modules.  A few CPAN modules use IoDIRP(io) to get the directory handle.

IO.xs uses IoFLAGS(io) to set taint mode and a small number of other CPAN modules also use it.  A few cpan modules use IoTYPE(io).

There's a few more used for storing format related information which only appear to be used in core.

Unfortunately I don't think any of them are documented.

khwilliamson: How does this revision sound?

tonycoz: It feels a bit vague.  How about:

```
An SV type representing I/O.  May contain input and output L<PerlIO|perlapio> objects or
a C<DIR *> from opendir().

It also contains all of the other state associated with a handle including the flush flag, current line
number and L<format|perlform> related information.
```

though some of the phrasing feels wrong.

PR #18286, opened 2020-11-03
Add most basic __DATA__ example

jidanni:
1. Fixing wrong claim.
2. Adding basic example! (Without which there is no way a user could find out how to get started using __DATA__, aside from searching Google.)

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -606,7 +606,11 @@ X<line> X<file> X<package>
 
 The two control characters ^D and ^Z, and the tokens __END__ and __DATA__
 may be used to indicate the logical end of the script before the actual
-end of file.  Any following text is ignored.
+end of file.  Any following text is ignored, except with __DATA__:
+
+   while (<DATA>) { print; }
+   __DATA__
+   Hello world.
 
 Text after __DATA__ may be read via the filehandle C<PACKNAME::DATA>,
 where C<PACKNAME> is the package that was current when the __DATA__


pod/perldata.pod
```diff
@@ -606,7 +606,11 @@ X<line> X<file> X<package>
 
 The two control characters ^D and ^Z, and the tokens __END__ and __DATA__
 may be used to indicate the logical end of the script before the actual
-end of file.  Any following text is ignored.
+end of file.  Any following text is ignored, except with __DATA__:
```
Grinnz, line 609: This is not really true; the following text is still ignored by the interpreter, reading from DATA is a separate action that happens to reuse the source handle.

pod/perldata.pod
```diff
@@ -606,7 +606,11 @@ X<line> X<file> X<package>
...
+
+   while (<DATA>) { print; }
```
Grinnz, line 611: We shouldn't add examples using while-readline with `$_` since it does not localize that variable, instead suggest:

    while (my $line = <DATA>) { print $line; }

pod/perldata.pod
Grinnz: And also it's not specific to `__DATA__`; the DATA handle will also recognize `__END__` in the top level script (as explained later in the section).

Grinnz: Such text is also "not ignored" by POD parsers, but is still ignored by the interpreter, so I believe that's what this sentence is referring to. Perhaps just adding "by the Perl interpreter" or something to that effect would be better.

pod/perldata.pod
jidanni: I like $line, but I was able to get away with $_ just fine, so users
should be told when they can still use $_, and when they shouldn't.
Else everybody will just use trial and error.

Also regarding "Any following text is ignored."
Well my example proves that it is not --
as far as beginners are concerned.
So if indeed your want to say it is ignored,
then you need to explain more, else it seems
the article was written at 5:00 AM.

So perhaps you can revise accordingly. I will probably screw up
any edit of an edit anyway...

Grinnz: The explanation isn't relevant to this section since it's not a section about using while-readline. We should simply use examples that don't promote bad practices.

The text is always ignored by the interpreter, the original text is correct, though it could probably use more explanation. In other words, text following `__DATA__` or `__END__` is never parsed and executed as part of the perl program.

jidanni: Sounds good. Please make the edit. By the way, I don't recall seeing a minimal `__END__` example anywhere either...

Grinnz:
Suggested alternative in #18287

PR #18287, opened 2020-11-03
perldata - example of using DATA

Grinnz:
As suggested by @jidanni in #18286


jidanni (approved):
Great edit. Even has the `close`.

(And `__END__` is like `__DATA__` as the page already says, so no separate example needed, it turns out.)

PR #18291, opened 2020-11-04
delete metaphoric quotations

ghost:
Seems like this nonsense is finally obsolete.


karenetheridge:
Could you please include more explanation in your your pull request comments?

wchristian:
This arbitrarily removes perl history and character for (at best) marginal gains, and as the emoji reactions above indicate, several people feel there is no clear reason to do this.

As such, at the very least a reason should be argued, in a clear manner.

tonycoz:
> Seems like this nonsense is finally obsolete.
> 
> pod/perlhack.pod
> 
> "..."

That justifies adding quotes to new `*.c` files, not for removing them from existing files.

jkeenan:
Rejected.

apparluk:
People used to sit around in parks on the grass in the 1960s reading those volumes.

Is that really the mindset of the developers of the language or the current contributors?

Issue #18260, opened 2020-10-21
[doc] perlsyn: give example when talking about "my" in "for" initialization section


jidanni:
In man perlsyn

>       There is one minor difference: if variables are declared with "my" in the initialization section of the
>      "for", the lexical scope of those variables is exactly the "for" loop (the body of the loop and the
>      control sections).

OK, but give an example saying if that syntax is
- `for my $i ($i = 1; $i < 10; $i++)` **or**
- `for (my $i = 1; $i < 10; $i++)`

(Answer: the latter is correct. But most users can only guess and then see which one creates an error and pick the other one.)

Grinnz:
The first is invalid perl code for unrelated reasons. There are two 'for's, the C-style and the foreach style. The quoted documentation is in the C-style for loop section.

I agree that an example using `my` would be useful.

jkeenan:
> I agree that an example using `my` would be useful.

Please review https://github.com/Perl/perl5/pull/18292.

Thank you very much.
Jim Keenan

jkeenan:
P.r. referenced above was superseded by commit dba7cf2eab.  Closing ticket.

Thank you very much.
Jim Keenan

PR #18292, opened 2020-11-05
Provide code example for 'my' declared in initialization of 'for' loop

jkeenan:
For: https://github.com/Perl/perl5/issues/18260

### DIFF
--- a/pod/perlsyn.pod
+++ b/pod/perlsyn.pod
@@ -455,8 +455,22 @@ is the same as this:
 There is one minor difference: if variables are declared with C<my>
 in the initialization section of the C<for>, the lexical scope of
 those variables is exactly the C<for> loop (the body of the loop
-and the control sections).
+and the control sections).  To illustrate:
 X<my>
+
+    my $i = 'samba';
+    for (my $i = 1; $i <= 4; $i++) {
+        print "$i\n";
+    }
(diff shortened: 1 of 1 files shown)


Grinnz (approved):
Seems illustrative. I feel like it would be better to show a strict 'vars' error outside the scope than shadowing, but that is harder to show in a concise example.

jkeenan:
> Seems illustrative. I feel like it would be better to show a strict 'vars' error outside the scope than shadowing, but that is harder to show in a concise example.

I thought about that possibility and concluded that if I showed that error, I'd first have to prepend the code with a comment to the effect that `$i` had not already been declared before the `for` loop.  That verbiage seemed excessive.

jkeenan:
> > Seems illustrative. I feel like it would be better to show a strict 'vars' error outside the scope than shadowing, but that is harder to show in a concise example.
> 
> I thought about that possibility and concluded that if I showed that error, I'd first have to prepend the code with a comment to the effect that `$i` had not already been declared before the `for` loop. That verbiage seemed excessive.

No one has objected to this p.r. since early November, so I intend to merge it to blead within 3 days.

Thank you very much.
Jim Keenan

jkeenan:
Handled manually in commit dba7cf2eab.

PR #18298, opened 2020-11-06
fix typo

gtataranni:
small typo in the docs


karenetheridge:
This is a file used by a cpan-first dual-life distribution (presumably in its tests), so it should be patched in its respective repository -- **HOWEVER**, it is just a copy of a file that actually does live in core, so could you patch that instead?  It is pod/perlfunc.pod. Thanks!

tonycoz:
The `cpan/Pod-Perldoc/corpus/perlfunc.pod` change isn't really needed (and should be submitted upstream), with that included the PR can't be merged as is.

Grinnz:
That change is undone by the second commit, it just needs to be squashed.

tonycoz:
Thanks, I squashed these and applied it as 54525b93d52a52326368ccf8a330d5b2d1e56089.

PR #18301, opened 2020-11-06
op.h: Restrict scope of multiconcat symbols to core

khwilliamson:
These are internal only


khwilliamson:
@iabyn any objections?

iabyn:
Looks plausible to me

PR #18302, opened 2020-11-06
keywords.h: Restrict to core

khwilliamson:
These should not be used outside core

### DIFF
--- a/keywords.c
+++ b/keywords.c
@@ -3451,5 +3451,5 @@ unknown:
 }
 
 /* Generated from:
- * f77998a5bc995c1b42d3d080de227ef5f11638bcd329367431d8f193aef2d3cc regen/keywords.pl
+ * c6cfa0000273762d1abe03a08165b4ec1ede8d4898a87e31d17dd55a0b910578 regen/keywords.pl
  * ex: set ro: */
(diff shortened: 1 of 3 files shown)


khwilliamson:
@karenetheridge says that these are used on cpan

karenetheridge:
..but maybe they shouldn't be... there were only two matches on grep.metacpan.org

Leont:
> ..but maybe they shouldn't be... there were only two matches on grep.metacpan.org

Devel::Hook can easily be rewritten not to use it, and warnings::unused includes keywords.h but doesn't appear to do anything with it? (unless I'm missing something)

karenetheridge:
Devel::Hook uses the header definitions to check for the availability of UNITCHECK, but it would be more correct to simply do a version (`$]`) check instead: https://metacpan.org/release/Devel-Hook/source/Hook.xs#L22

warnings::used uses KEY_our and KEY_my:

https://metacpan.org/release/warnings-unused/source/unused.xs#L223
https://metacpan.org/release/warnings-unused/source/unused.xs#L248

khwilliamson:
I'm wondering if we could have something like `#define USE_PERL_KEYWORDS` that a module could use to gain access to them.

PR #18304, opened 2020-11-08
remove email

ghost:


karenetheridge:
This will cause some t/Porting tests to fail, because it checks that all authors of git commits are properly credited. Is there a problem?

ghost:
The problem seems to be finding an author to submit a PR deleting the two lines.

Do you have a solution?

karenetheridge:
The question was why you want to delete those two lines.

ghost:
Sometimes in Denmark you're not supposed to say "could you close the window" but "it's cold in here".

jgamble:
That's a thing in Maine, too, but it doesn't really answer the question. Is there a specific reason you want to remove Eric Lindblad from the list, instead of just removing the e-mail address?

haarg:
There is a test to ensure all of the appropriate authors are listed. Removing this listing causes that test to fail, as can be seen in the checks for this PR. So this change alone is not adequate. There are several potential approaches that could be used to remove the email and adapt the test to pass, but what approach is used will depend on why you want to remove the email address.

ghost:
The author appears in perl5333delta.pod.

The author coded in the language from 2017-2020.

I suppose someone else could submit a PR to delete the AUTHORS and checkAUTHORS.pl lines.

Grinnz:
This list is for authors who have contributed to Perl at any time in the past, it is not a list of current contributors.

apparluk:
I suppose someone else could submit a PR to delete the AUTHORS email and checkAUTHORS.pl line.

karenetheridge:
> I suppose someone else could submit a PR to delete the AUTHORS and checkAUTHORS.pl lines.

To what end?

PR #18306, opened 2020-11-10
symlink for Windows

### LINKED ISSUE #18005: [feature] Implement the symlink function on Windows
I was trying to set GNU Stow on Windows 10 when I saw this: `The symlink function is unimplemented`.
The Windows API equivalent of symlink() is [CreateSymbolicLinkA](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinka) and [CreateSymbolicLinkW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw).

tonycoz:
implement symlinks for Win32.

This implements symlink(), lstat(), readlink() and updates stat() to fetch the volume id and file number into st_dev and st_ino.

One big difference with typical POSIX systems is that Win32 symlink() requires either elevated privileges or developer mode and several core tests assume that symlink() being available means it will succeed.  Core tests were updated to fix that and some other assumptions that code made.

Due to MSVCRT (which gcc builds use) stat not handling symbolic links correctly, I re-wrote win32_stat()... which made it much simpler and eliminated the need for ${^WIN32_SLOPPY_STAT} (and fixed time handling, along with the update to utime)

The API used requires Windows Vista or later, but later commits in the series allow perl to compile and load on pre-Vista (mostly so I could get results from the Windows 2000 smoker), but the win32/stat.t test *doesn't* check for that so fails.  If we do want to fully support pre-Vista windows the test can be updated, but I didn't think it was worth it for now.

The change to Test::Harness had been applied upstream but not released.

Fixes #18005


win32/config_h.PL
```diff
@@ -12,6 +12,8 @@ use File::Compare qw(compare);
 use File::Copy qw(copy);
 use File::Basename qw(fileparse);
 
+${^WIN32_SLOPPY_STAT} = 0;
```
xenu: This very PR removed this variable.

win32/win32.c
```diff
@@ -3166,44 +3479,95 @@ win32_link(const char *oldname, const char *newname)
...
-      /* ERROR_INVALID_FUNCTION - eg. on a FAT volume */
-      errno = EINVAL;
-      break;
+    newfile = PerlDir_mapA(newfile);
+
+    /* are we linking to a directory?
+       CreateSymlinkA() needs to know if the target is a directory,
+       if the oldfile is relative we need to make a relative path
+       based on the newfile
+    */
```
xenu, line 3513: The problem with the heuristics below is that it isn't possible to make a symlink to a non-existent directory. I think it would make sense to always treat destinations that end with a slash (or `/..` or `/.`) as directories.

tonycoz: Done

tonycoz:
This was applied manually as merge commit b277e767f3a3d9af20f138a95c26625cf2dfb53a

PR #18307, opened 2020-11-10
fetch magic on the first stacked filetest, not the last

### LINKED ISSUE #18293: stacked filetest ops stat unexpected “filename containing newline”
Module: core

**Description**

The following MWE script unexpectedly outputs a warning about an “unsuccessful stat on filename containing newline”:

```
#!/usr/bin/perl
use 5.014;
use warnings;
use HTTP::Tiny;

HTTP::Tiny->new->head("https://www.perl.org/");
if ('a' =~ /^(.)$/) {
    if (-f -r $1) {
        say "a";
    }
}
```

A workaround is to replace `-f -r $1` with `-r $1 && -f _`. This is in contradiction to the documentation in `man perlfunc`, which states that the former is just “a form of purely syntactic sugar” for the latter.

**Steps to Reproduce**

Run the above script on Ubuntu Linux (e.g., 20.04).

**Actual behavior**

```
$ ./bug.pl 
Unsuccessful stat on filename containing newline at ./bug.pl line 8.
```

**Expected behavior**

```
$ ./bug.pl 
$ touch a
$ ./bug.pl 
a
```

Perl should try to stat the filename `a` instead of any “filename containing newline”,
and output `a` if and only if file `a` exists and is readable.

**Perl configuration**

I encountered this behaviour on both Perl 5.30.0 as included in Ubuntu Linux 20.04 (x86_64), as well as Perl 5.22.1 as included in Ubuntu Linux 16.04 (x86_64).

tonycoz:
fixes #18293

### DIFF
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -3067,7 +3067,7 @@ S_try_amagic_ftest(pTHX_ char chr) {
     SV *const arg = *PL_stack_sp;
 
     assert(chr != '?');
-    if (!(PL_op->op_private & OPpFT_STACKING)) SvGETMAGIC(arg);
+    if (!(PL_op->op_private & OPpFT_STACKED)) SvGETMAGIC(arg);
 
     if (SvAMAGIC(arg))
     {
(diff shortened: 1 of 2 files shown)


hvds:
Oh, it hadn't even occurred to me the converse flag might already exist.

Looks good to me.

PR #18308, opened 2020-11-10
comp/parser.t count two tests 

brainbuz:
Two lines in comp/parser.t were present to make sure they didn't crash the parser.
I enclosed the relevant section in a block {. 
I added Pass tests immediately after each of the two lines.
Motiviation: I was working on Perl 7 Alpha tests, this section of the code generates warnings that need silencing and it is easier to see all the related code and comment when the section is grouped together in a block. Since each of the two lines is testing something it should be counted as a test.

### DIFF
--- a/t/comp/parser.t
+++ b/t/comp/parser.t
@@ -243,7 +243,7 @@ eval q[
 like($@, qr/Missing right curly/, 'nested sub syntax error' );
 
 eval q[
-    sub { my ($a,$b,$c,$d,$e,$f,$g,$h,$i,$j,$k,$l,$m,$n,$o,$p,$q,$r,$s,$r);
+    sub { my ($a,$b,$c,$d,$e,$f,$g,$h,$i,$j,$k,$l,$m,$n,$o,$p,$q,$r,$s);
 	    sub { my $z
 ];
 like($@, qr/Missing right curly/, 'nested sub syntax error 2' );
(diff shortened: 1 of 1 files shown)


Grinnz:
Not sure I see the point of the block, Nothing within it is lexically scoped.

brainbuz:
Found what I think is a typo in parser.t adding to pull request. The typo causes 'redefined' warnings when warnings are enabled.

The point of the block is that for Perl 7 some warnings need to be disabled, for Perl 5 the only purpose is it groups the code making it more readable. The lines that I added tests for should be counted as tests.

Grinnz:
I don't know what that means. There are no warnings being disabled here. Also, it is unknown what Perl 7 will entail.

brainbuz:
I've removed the braces.

tonycoz:
Rather than `is(1, 1, ...)` wouldn't they be better as `pass(...)`?

brainbuz:
The test doesn't use test::more or test.pl and doesn't have a pass routine defined. 
If it is better form to add a pass subroutine I can do that. 
something like `sub pass { is( 1,1, @_ ) } ` should work

tonycoz:
> 
> 
> The test doesn't use test::more or test.pl and doesn't have a pass routine defined.
> If it is better form to add a pass subroutine I can do that.
> something like `sub pass { is( 1,1, @_ ) } ` should work

Ahh, you're right, sorry for the noise.

I've squashed the braces change into the base commit, rebase and applied your changes to blead as 4281700a59b8da499ae9f1980d60c7e958a6583d and a5fba4f1467e17d87a18ed69501a89ffa05d25ce.

PR #18324, opened 2020-11-16
perlop: Clarify parts of tr///

khwilliamson:


pod/perlop.pod
```diff
@@ -2595,25 +2595,36 @@ C<tr/ACEGIBDFHJ/0246813579/>.
...
-C<tr(+\-*/)/ABCD/>.
-
-Characters may be literals, or (if the delimiters aren't single quotes)
-any of the escape sequences accepted in double-quoted strings.  But
-there is never any variable interpolation, so C<"$"> and C<"@"> are
-always treated as literals.  A hyphen at the beginning or end, or
-preceded by a backslash is also always considered a literal.  Escape
-sequence details are in L<the table near the beginning of this
+not be bracketing quotes; for example, C<tr(aeiouy)(yuoiea)> or
+C<tr[+\-*/]"ABCD">.  This final example shows a way to visually clarify
```
Grinnz, line 2599: This looks pretty weird probably because of the double quotes, maybe something like `tr[+\-*/]|ABCD|` would be clearer?

pod/perlop.pod
```diff
@@ -2595,25 +2595,36 @@ C<tr/ACEGIBDFHJ/0246813579/>.
...
+what is going on for people who are more familiar with regular
+expression patterns than with C<tr>, and who may think forward slash
```
Grinnz, line 2601: I don't think this example shows this any more than any other example not using `/` delimiters. Maybe just say how the use of alternate delimiters can help distinguish it from the common usage of forward slash delimiters for regexes.

pod/perlop.pod
Grinnz: I also think this section should consistently use the word "delimiters" and not "quotes".

pod/perlop.pod
```diff
@@ -2595,25 +2595,36 @@ C<tr/ACEGIBDFHJ/0246813579/>.
...
+delimiters imply that C<tr> is more like a regular expression pattern
+than it actually is.  (Another option might be to use C<tr[...][...]>.)
+
+C<tr> isn't fully like bracketed character classes, just
+(significantly) more like them than it is to full patterns.  For
+example, characters appearing more than once in either list behave
+differently here than in patterns, and C<tr> lists do not allow
```
Grinnz, line 2608: Well it does allow them, they just don't have the special meaning they do in regexes and will warn the same way as when used in a double quoted string.

jkeenan:
Can we move this ticket toward resolution?  There are several unresolved conversations in the p.r., but I suspect the participants can probably negotiate a resolution.

Thank you very much.
Jim Keenan

khwilliamson:
On 12/22/20 4:58 PM, James E Keenan wrote:
> Can we move this ticket toward resolution? There are several unresolved 
> conversations in the p.r., but I suspect the participants can probably 
> negotiate a resolution.
>
This ticket can wait until closer to the end of the development cycle. 
I am working on several things now that can't wait.  So this is lower 
priority.

PR #18325, opened 2020-11-16
Smoke me/khw pr curly

khwilliamson:


regcomp.c
```diff
@@ -12826,9 +12827,9 @@ S_regpiece(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth)
     }
 
     /* Forbid extra quantifiers */
-    if (ISMULT2(RExC_parse)) {
-        RExC_parse++;
-        vFAIL("Nested quantifiers");
+    if (isQUANTIFIER(RExC_parse, RExC_end)) {
+	RExC_parse++;
+	vFAIL("Nested quantifiers");
```
hvds: Did you intentionally change the spaces to tabs here?

inline.h
```diff
@@ -2003,23 +2003,91 @@ S_lossless_NV_to_IV(const NV nv, IV *ivp)
...
-S_regcurly(const char *s)
+S_regcurly(const char *s, const char *e, const char * result[5])
 {
+    /* This function matches a {m,n} quantifier
+     *
+     * It returns FALSE if the input doesn't match such a quantifier;
+     * otherwise, it returns TRUE and optionally returns information about the
+     * parsed input in a 5 element array, if specified, as follows:
+     *  result[0]   points to the closing brace
+     *  result[1]   points to the first byte of the lower bound
```
hvds: Could you clarify why you did not choose to have this evaluate the numbers and return (pointer to end, min value, max value)? At first glance, this seems like a natural place to do that work; it would also reduce the amount of explanation required here quite a bit.

regcomp.c
```diff
@@ -12550,6 +12550,37 @@ S_regbranch(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, I32 first, U32 depth)
     return ret;
 }
 
+U32
+S_get_quantifier_value(pTHX_ RExC_state_t *pRExC_state,
```
hvds, line 12660: I don't understand what this function is doing: why is a null byte "invalid" while nothing else is? Why do we touch `RExC_parse` in the failure cases? Should it be checking that `end` is unchanged by grok_atoUV?

This feels like it is very tightly coupled to what regcurly and regpiece do, in subtle ways.

regcomp.c
```diff
@@ -12622,54 +12651,31 @@ S_regpiece(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth)
...
-            }
+        if (regcurly(RExC_parse, RExC_end, regcurly_return)) {
+            const char * min_start = regcurly_return[1];
+            const char * min_end   = regcurly_return[2];
+            const char * max_start = regcurly_return[3];
+            const char * max_end   = regcurly_return[4];
 
-            assert(*next == '}');
+            assert(min_start);
+            min = get_quantifier_value(pRExC_state, min_start, min_end);
```
hvds, line 12776: Was this supposed to handle `if (min_start == min_end) min = 0; else ...`? It doesn't look like `get_quantifier_value` handles that, but the `{,n}` comment below implies that something should.

regcomp.c
khwilliamson: All I can think of is this is a vim bug.  But no; I'll fix it.

inline.h
khwilliamson: There are a couple reasons.  One is that this is allegedly an inline function and it would be adding too much code for that.  Now that you have pointed it out, I think this function should be removed from being inlined.  I imagine that it is I who inlined it a long time ago.  At that time I did not realize that compilation is not a time critical activity, and this is only called during compilation.  And a future commit will add more lines to it, so I doubt that a compiler actually would choose to obey our inline request anyway.  I now think this function should be moved to regcomp.c.  This would make this reason no longer applicable.  The other calls to it are in toke.c.

The 2nd reason is, before I thought of the first, I actually was implementing it the way you suggest is the natural place, but the problem was that not all possible inputs are legal.  That means there would have to be a way to signal these to the caller; perhaps through an enum, but that makes the return complicated, and the amount of explanation required might actually go up.  The function can't so easily just raise the fatal errors, as it needs the regex pattern struct to be able to generate a proper error message, and this is called from toke.c where that isn't available.  I suppose that could be an optional parameter, but it's now getting more complicated than what the current implementation is

regcomp.c
khwilliamson: I should have commented what this function does, and am abashed that I didn't.
That null should be '0' a zero character.  I will add a test to the suite for this case.
Perhaps there should be at least an assert that 'end' doesn't get changed, but since it is tightly coupled with regpiece(), there shouldn't be a possibility of this.

It is a common paradigm throughout regcomp.c to position RExC_parse prior to outputting a diagnostic so that the <--HERE in that diagnostic comes out in the proper place.  It's not clear to me that every instance of this needs to have a comment about that.

regcomp.c
khwilliamson: The comment is wrong, and should be changed.  Here there should be an assert about min_end.

khwilliamson:
2 revision commits pushed.

regcomp.c
khwilliamson: And I think the position of the HERE in these diagnostics is wrong, but wanted to start with a state where no external behavior change is visible.  I think it should be ...}<--HERE. And I intend to make a later commit change to do that, unless there is a good reason not to.

inline.h
hvds: >  I now think this function should be moved to regcomp.c.

Agreed.

> The 2nd reason is [...] it needs the regex pattern struct to be able to generate a proper error message

I think all the callers that don't ask for results should use a separate function that just takes start and end pointers. Having just two callers to the full regcurly, one living right next to it, should make it easier to manage the complexity of behaviour one (no regexp struct, no results, just tell me if this is the right pattern for a curly) versus behaviour two (do everything, die on error).

khwilliamson: I think you must not be understanding.  Look at the new version I just pushed

hvds: I don't understand what it is I'm supposed to be not understanding, sorry. I've pushed an hv/curly branch with an additional commit as a very rough-and-ready example of what I'm talking about.

khwilliamson: Based on your branch, it was I who misunderstood what you meant.

My PR, as amended by your other comments, is an attempt to set the stage for further changes so that they only have to be made in one place.  That is why I separated out the function that actually computes and verifies the bounds numeric values, instead of having two copies of the same logic.   The error messages now don't really point to the precise error location, and should be fixed.  I left the display as-is in this commit so as to separate it out.  But having a single place for this logic instead of two makes maintenance easier.

My understanding was that you thought it overly complicated that I returned an array of 5 elements, but I don't think your proposal of returning an array of 2 elements plus an extra return parameter is really a significant advantage.

I have no objection  to making a wrapper to hide the extra NULL parameter(s) from all but one call where they aren't NULL, .  If the function were API, I would have done that myself.  But it's only ever going to be called from two files, and I didn't see a need that nicety.

What I propose is to humor me, and let my PR stand, as has been amended by your previous review, and see what direction things go before making any "final" decisions.

hvds: Maybe you could describe what direction you plan to take it, or even include the additional work in this PR?

It seems odd to me, and undesirable, to pass around the start and end pointers of a number knowing that the only thing we'll do with them is convert them to the number they represent, when we could just as effectively do that on the spot.

In my example it would also be practical to replace the two-element array with two individual pointers, in which case the caller could pass `(..., &min, &max)` for a further simplification.

regcomp.c
```diff
@@ -12549,28 +12550,149 @@ S_regbranch(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, I32 first, U32 depth)
...
+    }
+
+    if (s >= e || *s != '}' || ! min_start) {
+        return FALSE;
+    }
+
+    if (result) {
+
+#define RBRACE  0
+#define MIN_S   1
```
atoomic, line 12627: suggestion: use a more explicit name for the macros `MIN_START`, `MIN_END`, ... ?

inline.h
khwilliamson: This is just the first step in a longish sequence of commits that would fix the pointer in any error messages to be a better position., and to relax the current restricted syntax of the curly quantifiers, as has been long planned.

I asked several people to look at your concerns, and none echoed them.  I think it comes down to a matter of taste.  One response was that the parsing and converting to numeric really don't belong together, which is effectively what I had done.  The bottom line for me is to consolidate like-functionality into a single function, so that changes only have to be made in a single place.

PR #18327, opened 2020-11-17
enable warnings for io/sem.t and fix a few other bugs

### LINKED ISSUE #17926: Should semctl() built-in die if not provided with correct type of argument?
Test file `t/io/sem.t` was introduced into the core distribution in commit 64d76282359 in December 2013 and is almost entirely unchanged since then.  When run as part of the test suite, it appears entirely ordinary.
```
$ cd t;./perl harness -v io/sem.t; cd -

ok 1 - acquired semaphore
ok 2 - Initialize all 10 semaphores to zero
ok 3 - Set semaphore 3 to 17
ok 4 - Get current semaphore values
ok 5 - Make sure we get back statuses for all 10 semaphores
ok 6 - Checking value of semaphore 3
ok 7 - Check value via GETVAL
ok
All tests successful.
Files=1, Tests=7,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.02 cusr  0.00 csys =  0.03 CPU)
Result: PASS
```
However, the file was written without using any warnings.  If we simply add a `-w` flag to its shebang line and re-run it, we get:
```
$ cd t;./perl harness -v io/sem.t; cd -

ok 1 - acquired semaphore
ok 2 - Initialize all 10 semaphores to zero
ok 3 - Set semaphore 3 to 17
ok 4 - Get current semaphore values
ok 5 - Make sure we get back statuses for all 10 semaphores
ok 6 - Checking value of semaphore 3
ok 7 - Check value via GETVAL
Argument "ignore" isn't numeric in semctl at io/sem.t line 50.
Argument "ignore" isn't numeric in semctl at io/sem.t line 59.
Use of uninitialized value $semvals in semctl at io/sem.t line 59.
Argument "ignored" isn't numeric in semctl at io/sem.t line 69.
ok
All tests successful.
Files=1, Tests=7,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.01 cusr  0.00 csys =  0.02 CPU)
Result: PASS
```
In the core-p7 branch I've been working on making test files strict- and warnings-compliant.  In many cases, inserting a `no warnings 'numeric'` would be sufficient.  However, in this case, the warning is significant because it points to a flaw in the test file and, potentially, a flaw in `perl` itself.

Here is the first instance of the warning cited above:
```
    ok(semctl($id, "ignore", SETALL, pack("s!*",(0)x$nsem)),
       "Initialize all $nsem semaphores to zero");
```
Now, it's perfectly obvious that `"ignore"` is non-numeric; hence, the warning.  But a case can be made that semctl() should die with a non-numeric argument rather than simply warn.  `perldoc -f semctl` starts:
```
   semctl ID,SEMNUM,CMD,ARG
        Calls the System V IPC function semctl(2).
```
... but has nothing further to say about `SEMNUM` -- though that spelling suggests the argument ought to be numeric.

If, however, we look under the hood, we see that the second argument in the underlying system call *must* be an `int`.  On Linux:
```
$ man 2 semctl
SYNOPSIS
       #include <sys/types.h>
       #include <sys/ipc.h>
       #include <sys/sem.h>

       int semctl(int semid, int semnum, int cmd, ...);

DESCRIPTION
       semctl() performs the control operation specified by cmd on the System V semaphore set iden‐
       tified by semid, or on the semnum-th semaphore of that set.  (The semaphores in  a  set  are
       numbered starting at 0.)
```
On OpenBSD:
```
SYNOPSIS
     #include <sys/sem.h>

     int
     semctl(int semid, int semnum, int cmd, union semun arg);

```
Shouldn't the Perl `semctl()` built-in `die` rather than `warn` if provided the wrong type for its second argument?

Thank you very much.
Jim Keenan

tonycoz:
fixes #17926 and some unicode bug type problems

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5557,6 +5557,7 @@ t/io/inplace.t			See if inplace editing works
 t/io/iofile.t			See if we can load IO::File on demand
 t/io/iprefix.t			See if inplace editing works with prefixes
 t/io/layers.t			See if PerlIO layers work
+t/io/msg.t			See if SysV message queues work
 t/io/nargv.t			See if nested ARGV stuff works
 t/io/open.t			See if open works
 t/io/openpid.t			See if open works for subprocesses
--- a/doio.c
+++ b/doio.c
@@ -3162,7 +3162,7 @@ Perl_do_semop(pTHX_ SV **mark, SV **sp)
     STRLEN opsize;
     const I32 id = SvIVx(*++mark);
     SV * const opstr = *++mark;
-    const char * const opbuf = SvPV_const(opstr, opsize);
+    const char * const opbuf = SvPVbyte(opstr, opsize);
 
     PERL_ARGS_ASSERT_DO_SEMOP;
     PERL_UNUSED_ARG(sp);
--- a/doio.c
+++ b/doio.c
@@ -2999,13 +2999,13 @@ Perl_do_ipcctl(pTHX_ I32 optype, SV **mark, SV **sp)
     {
 	if (getinfo)
 	{
-	    SvPV_force_nolen(astr);
+            SvPV_force_nolen(astr);
 	    a = SvGROW(astr, infosize+1);
 	}
 	else
 	{
 	    STRLEN len;
-	    a = SvPV(astr, len);
+	    a = SvPVbyte(astr, len);
 	    if (len != infosize)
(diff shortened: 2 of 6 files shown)


tonycoz:
Draft since I found some other similar UTF-8 issues

tonycoz:
applied directly

PR #18330, opened 2020-11-17
Improve the Release Manager's Guide

tomhukins:


khwilliamson (commented):
Actually there is a dromedary server that replaces the outdated reference.  So we should direct people to that. @toddr can you give the correct reference

Porting/release_managers_guide.pod
```diff
@@ -166,10 +160,6 @@ pre-release testing, and you may wish to upload to PAUSE via URL.
 Make sure you have a way of sharing files, such as a web server or
 file-sharing service.
 
-Porters have access to the "dromedary" server (users.perl5.git.perl.org),
```
toddr, line 169: Honestly I'd recommend we just suggest a direct PAUSE upload here? do we really need an alternate place to upload the upload?

Grinnz: I believe the intention of this space is a pre-PAUSE sanity check upload as often used by releasers.

tomhukins: Thanks to @toddr I now have access to the new dromedary.  However, it looks like it has no per-user public_html support: this is disabled in `/etc/httpd/conf.d/userdir.conf`.  Although my commit message was wrong - we do have a dromedary server - I suggest keeping the contents of my commit to remove any mention of dromedary, but amending the commit message.  Does that seem sensible?

tomhukins:
I'll merge fe5842c to blead in the next few days unless anyone objects.

PR #18332, opened 2020-11-17
mkheader: Prevent 'print on closed filehandle' messages

jkeenan:
Program 'mkheader' is called from cpan/Unicode-Collate/Makefile.PL,
which was recently revised to run with both strictures and warnings.
Makefile.PL executes mkheader via a 'do'.  This was causing the
following statements to be printed during 'make' in the Perl
5 build process:

printf() on closed filehandle $fh_h at .../cpan/ExtUtils-MakeMaker/lib/ExtUtils/MakeMaker.pm line 1233.
print() on closed filehandle $fh_h at .../cpan/ExtUtils-MakeMaker/lib/ExtUtils/MakeMaker.pm line 1234.

Storing the original filehandle during select(), then restoring it at
the end of the loop, clears up those statements.

### DIFF
--- a/cpan/Unicode-Collate/mkheader
+++ b/cpan/Unicode-Collate/mkheader
@@ -140,7 +140,8 @@ foreach my $tbl (@tripletable) {
     my $init = $tbl->{init};
 
     open my $fh_h, ">$file" or croak "$PACKAGE: $file can't be made";
-    binmode $fh_h; select $fh_h;
+    binmode $fh_h;
+    my $old_fh = select $fh_h;
     my %val;
 
     print << 'EOF';
@@ -195,6 +196,7 @@ EOF
     }
     print "};\n\n";
     close $fh_h;
+    select $old_fh;
 }
 
 1;


Grinnz:
This file is upstream cpan: https://metacpan.org/release/Unicode-Collate

jkeenan:
This patch will ultimately have to go upstream, as Unicode-Collate is cpan-upstream.  However, it took me so long to figure out what the problem was that I would like to get confirmation of my diagnosis here.

I *think* that the problem arose when a new version of Unicode-Collate that had `use strict; use warnings;` added to its Makefile.PL.  When that program runs, it executes `do ./mkheader`.  Without this patch, it appears that even though the filehandle in question was explicitly closed before the end of a loop, it nonetheless remained the `select()`-ed filehandle.

At first I thought that this was a problem when a new version of ExtUtils-MakeMaker was synched into blead, so I filed https://rt.cpan.org/Ticket/Display.html?id=133762.  However, I now think the problem lies entirely within Unicode-Collate.

Thoughts?

Thank you very much.
Jim Keenan

Grinnz:
The analysis makes sense to me. Personally I think a better change would be to modify every `print` and `printf` in that block to reference `$fh_h` so it does not mess with global state at all, even though it would be more tedious.

khwilliamson:
@jkeenan  Several people think there is a better way of accomplishing this.  What do you say?

jkeenan:
I cannot reproduce these unexpected warnings at HEAD in blead.  It appears they vanished somewhere between c49e90e1c5 and 0aee951734.  So we can close this p.r. for the time being.

Thank you very much.
Jim Keenan

haarg:
EUMM added a workaround for the issue. This is still a bug in mkheader that should be fixed.

jkeenan:
> EUMM added a workaround for the issue. This is still a bug in mkheader that should be fixed.

Bug ticket opened for Unicode-Collate at https://rt.cpan.org/Ticket/Display.html?id=133952.  I think this ticket itself is closable.

Thank you very much.
Jim Keenan

jkeenan:
> > EUMM added a workaround for the issue. This is still a bug in mkheader that should be fixed.
> 
> Bug ticket opened for Unicode-Collate at https://rt.cpan.org/Ticket/Display.html?id=133952. I think this ticket itself is closable.
> 

No one has argued that this ticket is not closable, so I'm (re-)closing it now.

Thank you very much.
Jim Keenan

PR #18334, opened 2020-11-18
Improve pattern matching repeating loop termination

khwilliamson:


hvds:
One early comment: while I'm glad you wrote all the information in the final commit message, I don't think a commit message is the right place for it; better would be either to make it part of the docs in the repo, or put it in a ticket attached to this work.

I would like to contribute to review of this, but I'm sure it'll take me a while to get through it all.

khwilliamson:
My plan is to insert a reference to the commit number in the prolog to the function, once that commit number is known.  I have used this paradigm in the past, and I think others have too, to give people a quick wasy to find extra information that really is too complex for being inline with the code.

hvds:
> My plan is to insert a reference to the commit number in the prolog to the function, once that commit number is known. I have used this paradigm in the past, and I think others have too, to give people a quick wasy to find extra information that really is too complex for being inline with the code.

I am not against long commit messages per se, but this does not feel like the right sort of information to go in a commit message. I also do not find the commit message hugely clear, but we can barely review commit messages, and cannot edit them once they get into blead.

In the meantime: it would be helpful if the commit message could at least mention the name of the function that it spends 134 lines talking about.

khwilliamson:
I decided that this should go in the next development release to see what effect it has under wider testing.  I opened #18414 to move the long commit message to, and will keep that ticket open for now, so that any tweaks can be addressed

PR #18339, opened 2020-11-19
Update gitignore files to reflect files in repo

haarg:
There are a number of files excluded using gitignore rules that are
included in the repository. This can lead to confusion if something
other than git tries to read the ignore files.

Add rules to the gitignore files so that these files won't be ignored.
Also move a couple rules for specific dists out of the root gitignore.


karenetheridge (approved):

tonycoz:
Applied as a rebased merge commit in 581f869ed3f672307ccaa86023334fa0622481a4

Thank you.

PR #18347, opened 2020-11-23
Confine scope of SV_CONST to core

khwilliamson:
as well as the constants it uses.  This is unused in cpan


khwilliamson:
closed because another PR for this is already pending, that I forgot

Issue #18341, opened 2020-11-20
Recursive call to mortal_getenv() causes deadlock


mhx:
I've just built blead with a configuration that I've been building quite regularly back in the days:
```
./Configure -des -Uinstallusrbinperl -Dusethreads -DDEBUGGING=both -Doptimize=-ggdb3 -Dprefix=/home/mhx/perl/blead-mem -Dusedevel -Uversiononly -Aappend:ccflags="-DPERL_MEM_LOG -DPERL_MEM_LOG_STDERR -DPERL_MEM_LOG_ENV -DPERL_MEM_LOG_ENV_FD -DPERL_MEM_LOG_TIMESTAMP -DPERL_TRACK_MEMPOOL -DPERL_POISON -DDEBUG_LEAKING_SCALARS -DDEBUG_LEAKING_SCALARS_FORK_DUMP -DPERL_DEBUGGING_MSTATS -Wall -Wextra"
```
I'm pretty certain not all of this is needed for to reproduce the issue, but at least some of the `MEM_LOG` stuff is from a quick glance at the stack trace below.

The first time `miniperl` gets run during the build, it hangs immediately. Here's a stack trace:
```
(gdb) bt
#0  0x00007ffff7f8852b in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007ffff7f811e3 in pthread_mutex_lock () from /lib64/libpthread.so.0
#2  0x00005555557300d2 in Perl_mortal_getenv (str=0x555555b92fff "PERL_MEM_LOG") at inline.h:2608
#3  0x0000555555740882 in S_mem_log_common (mlt=MLT_NEW_SV, n=0, typesize=0, type_name=0x555555b919e8 "", sv=0x555555c89ea8, 
    oldalloc=0x0, newalloc=0x0, filename=0x555555ba2368 "sv.c", linenumber=9366, 
    funcname=0x555555bb5a58 <__FUNCTION__.24917> "Perl_newSVpv") at util.c:5011
#4  0x0000555555740fcc in Perl_mem_log_new_sv (sv=0x555555c89ea8, filename=0x555555ba2368 "sv.c", linenumber=9366, 
    funcname=0x555555bb5a58 <__FUNCTION__.24917> "Perl_newSVpv") at util.c:5150
#5  0x00005555557b40b5 in S_new_SV (my_perl=0x555555c862a0, file=0x555555ba2368 "sv.c", line=9366, 
[...]

PR #18359, opened 2020-11-26
Avoid deadlock with PERL_MEM_LOG

khwilliamson:
Avoid deadlock with PERL_MEM_LOG
     
     This fixes GH #18341
     
     The Perl wrapper for getenv() was changed in 5.32 to allocate memory to
     squirrel safely away the result of the wrapped getenv() call.  It does
     this while in a critical section so as to make sure another thread can't
     interrupt it and destroy it.
     
     Unfortunately, when Perl is compiled for debugging memory problems and
     has PERL_MEM_LOG enabled, that allocation causes a recursive call to
     getenv() for the purpose of checking an environment variable to see how
     to log that allocation.  And hence it deadlocks trying to enter the
     critical section.
     
     There are various solutions.  One is to use or emulate a general semaphore
     instead of a binary one.  This is effectively what
     PL_lc_numeric_mutex_depth does for another mutex, and the code for that
     could be used as a template.
     
     But given that this is an extreme edge case which requires Perl to be
     specially compiled to enable this feature which is used only for
     debugging, a much simpler, if less safe if it were to ever be used in
     production, solution should suffice.  Tony Cook suggested just avoiding
     the wrapper for this particular purpose.

### DIFF
--- a/util.c
+++ b/util.c
@@ -5008,7 +5008,11 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
 
     PERL_ARGS_ASSERT_MEM_LOG_COMMON;
 
-    pmlenv = PerlEnv_getenv("PERL_MEM_LOG");
+    /* Use plain getenv() to avoid potential deadlock with PerlEnv_getenv().
+     * This means that 'pmlenv' is not protected from other threads overwriting
+     * it on platforms where getenv() returns an internal static pointer.  See
+     * GH #18341 */
+    pmlenv = getenv("PERL_MEM_LOG");
     if (!pmlenv)
 	return;
     if (mlt < MLT_NEW_SV ? strchr(pmlenv,'m') : strchr(pmlenv,'s'))


util.c
```diff
@@ -5008,6 +5008,9 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
 
     PERL_ARGS_ASSERT_MEM_LOG_COMMON;
 
+    /* Note that 'pmlenv' is not protected from other threads overwriting it by
+     * calling getenv() themselves on platforms where getenv() returns an
+     * internal static pointer. */
     pmlenv = PerlEnv_getenv("PERL_MEM_LOG");
```
tonycoz: Why not just call getenv() here?  That's what happens anyway, and doesn't cost every other Perl_mortal_getenv() the (minor) cost of a string comparison.

khwilliamson: That is a much better solution than mine.  Updated to do what you say.

tonycoz (approved):

PR #18361, opened 2020-11-27
perlvar - clarify that paragraph mode also discards a single leading newline

### LINKED ISSUE #18352: [doc] paragraph mode discards any number of newlines at the beginning of the file
**Where**

https://perldoc.perl.org/perlvar#$/

**Description**

The document says:

>When `$/` is set to `""` and the entire file is read in with that setting, any sequence of consecutive newlines `"\n\n"` at the beginning of the file is discarded.

On testing, I find that a single newline at the beginning of the file is also discarded:

```bash
$ s='\na\nb\n\n12\n34\n\nhi'

$ printf '%b' "$s" | perl -00 -lnE 'say if $. == 1'
a
b
```

I don't think this is a bug, as this behavior is seen with `awk` too:

```bash
$ printf '%b' "$s" | awk -v RS= 'NR==1'
a
b
```

Grinnz:
Resolves #18352

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -1515,7 +1515,7 @@ be better for something. :-)
 
 Setting C<$/> to an empty string -- the so-called I<paragraph mode> -- merits
 special attention.  When C<$/> is set to C<""> and the entire file is read in
-with that setting, any sequence of consecutive newlines C<"\n\n"> at the
+with that setting, any sequence of one or more consecutive newlines at the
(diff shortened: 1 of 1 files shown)


learnbyexample:
>any sequence of one or more consecutive newlines C<"\n\n">
>at the beginning of the file is discarded.

I'd suggest to remove `C<"\n\n">` because that was part of the confusion.

PR #18365, opened 2020-11-29
Remove my_l?stat from public API

khwilliamson:
They are not documented, and if you try to use them, you get a compile
error.


khwilliamson:
@tonycoz what about your recent stat changes?

tonycoz:
> 
> 
> @tonycoz what about your recent stat changes?

They're unrelated, my changes are to the code these end up calling.

PR #18385, opened 2020-12-06
Make clear Fatal is still not OK

jidanni:
Otherwise people will think we are publishing the final OK accepted way!
Perhaps also note there is still no OK accepted way.

### DIFF
--- a/regen/warnings.pl
+++ b/regen/warnings.pl
@@ -1276,7 +1276,8 @@ As of Perl 5.20, instead of C<< use warnings FATAL => 'all'; >> you can
 use:
 
    use v5.20;       # Perl 5.20 or greater is required for the following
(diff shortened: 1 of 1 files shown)


jidanni:
> Not that we still don't recommend it

s/Not/Note/

Grinnz:
This already seems clear in context to me, since this follows all of the caveats in the text.

jidanni:
> This already seems clear in context to me, since this follows all of the caveats in the text.

No. Compare perlvar:
```
       In Perl 5.20.0 a new copy-on-write system was enabled by default, which
       finally fixes all performance issues with these three variables, and
       makes them safe to use anywhere.
```
So you still need to remind the user that this (new way to write FATAL) doesn't fix anything.
But do please rephrase it. I am a bad writer.

jkeenan:
Addressed in commit 01900a5f23.  Closing ticket.

PR #18388, opened 2020-12-07
Improve "Lost precision" warnings on increment/decrement

### LINKED ISSUE #18333: "Lost precision" warnings on increment/decrement
**Description**

"`Lost precision when incrementing ... by 1`" warning on increment seems to be limited to the case that the target value was positive finite value; infinity or negative large value are not warned.  (Similarly, infinity or positive large value are not warned on decrement.)
I think this is somewhat inconsistent.

**Steps to Reproduce**

```
$ perl -we 'my $x = 10 ** 300; my $y = $x++; printf "%a -> %a\n", $y, $x'
Lost precision when incrementing 100000 (snip) 540160.000000 by 1 at -e line 1.
0x1.7e43c8800759cp+996 -> 0x1.7e43c8800759cp+996
$ perl -we 'my $x = 10 ** 400; my $y = $x++; printf "%a -> %a\n", $y, $x'
Inf -> Inf
$ perl -we 'my $x = -10 ** 300; my $y = $x++; printf "%a -> %a\n", $y, $x'
-0x1.7e43c8800759cp+996 -> -0x1.7e43c8800759cp+996
```
The second and third example above shows that target variable are not changed, but no warnings are generated.

**Expected behavior**

I expect that perl will warn if increment cannot change the value even for infinite or negative values, and for decrement, warn even for infinite or positive values.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 33 subversion 4) configuration:
  Commit id: 185eb2d3f6e86edc749ad9b20f1b6a1e0b231ad7
  Platform:
    osname=linux
    osvers=4.19.0-12-686-pae
    archname=i686-linux
    uname='linux x41 4.19.0-12-686-pae #1 smp debian 4.19.152-1 (2020-10-18) i686 gnulinux '
    config_args='-de -Dusedevel -DEBUGGING'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=undef
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2 -g'
    cppflags='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='8.3.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=4
    doublesize=8
    byteorder=1234
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=12
    longdblkind=3
    ivtype='long'
    ivsize=4
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=4
    prototype=define
  Linker and Libraries:
    ld='cc'
    libpth=/usr/local/lib /usr/lib/gcc/i686-linux-gnu/8/include-fixed /usr/include/i386-linux-gnu /usr/lib /lib/i386-linux-gnu /lib/../lib /usr/lib/i386-linux-gnu /usr/lib/../lib /lib /lib64 /usr/lib64
    libc=libc-2.28.so
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.28'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'


  Compile-time options:
    DEBUGGING
    HAS_TIMES
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_USE_DEVEL
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under linux
  Compiled at Nov 15 2020 18:25:29
  @INC:
    /usr/local/lib/perl5/site_perl/5.33.4/i686-linux
    /usr/local/lib/perl5/site_perl/5.33.4
    /usr/local/lib/perl5/5.33.4/i686-linux
    /usr/local/lib/perl5/5.33.4
```

t-a-k:
Previously, imprecision warnings on increment (`Lost precision when incrementing %f by 1`) were only issued on positive finite values, and, on decrement, only issued on negative finite values.
This commit extends this warnings on both sign and infinite values so that perl will warn on all cases of losing precision.

This fixes #18333.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -8949,9 +8949,14 @@ Perl_sv_inc_nomg(pTHX_ SV *const sv)
     }
     if (flags & SVp_NOK) {
 	const NV was = SvNVX(sv);
-	if (LIKELY(!Perl_isinfnan(was)) &&
-            NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
-	    was >= NV_OVERFLOWS_INTEGERS_AT) {
+        if (NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
+            /* If NVX was NaN, the following comparisons return always false */
+            UNLIKELY(was >= NV_OVERFLOWS_INTEGERS_AT ||
+                     was < -NV_OVERFLOWS_INTEGERS_AT)
+#if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
+            && LIKELY(!Perl_isnan(was))
+#endif
+            ) {
 	    /* diag_listed_as: Lost precision when %s %f by 1 */
 	    Perl_ck_warner(aTHX_ packWARN(WARN_IMPRECISION),
 			   "Lost precision when incrementing %" NVff " by 1",
@@ -9128,9 +9133,14 @@ Perl_sv_dec_nomg(pTHX_ SV *const sv)
     oops_its_num:
 	{
 	    const NV was = SvNVX(sv);
-	    if (LIKELY(!Perl_isinfnan(was)) &&
-                NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
-		was <= -NV_OVERFLOWS_INTEGERS_AT) {
+            if (NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
+                /* If NVX was NaN, these comparisons return always false */
+                UNLIKELY(was <= -NV_OVERFLOWS_INTEGERS_AT ||
+                         was > NV_OVERFLOWS_INTEGERS_AT)
+#if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
+                && LIKELY(!Perl_isnan(was)))
+#endif
+                ) {
 		/* diag_listed_as: Lost precision when %s %f by 1 */
 		Perl_ck_warner(aTHX_ packWARN(WARN_IMPRECISION),
 			       "Lost precision when decrementing %" NVff " by 1",
--- a/sv.c
+++ b/sv.c
@@ -8952,9 +8952,11 @@ Perl_sv_inc_nomg(pTHX_ SV *const sv)
         if (NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
             /* If NVX was NaN, the following comparisons return always false */
             UNLIKELY(was >= NV_OVERFLOWS_INTEGERS_AT ||
-                     was < -NV_OVERFLOWS_INTEGERS_AT)
+                     was < -NV_OVERFLOWS_INTEGERS_AT) &&
 #if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
-            && LIKELY(!Perl_isnan(was))
+            LIKELY(!Perl_isinfnan(was))
+#else
+            LIKELY(!Perl_isinf(was))
 #endif
             ) {
 	    /* diag_listed_as: Lost precision when %s %f by 1 */
@@ -9136,9 +9138,11 @@ Perl_sv_dec_nomg(pTHX_ SV *const sv)
             if (NV_OVERFLOWS_INTEGERS_AT != 0.0 &&
                 /* If NVX was NaN, these comparisons return always false */
                 UNLIKELY(was <= -NV_OVERFLOWS_INTEGERS_AT ||
-                         was > NV_OVERFLOWS_INTEGERS_AT)
+                         was > NV_OVERFLOWS_INTEGERS_AT) &&
 #if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
-                && LIKELY(!Perl_isnan(was)))
+                LIKELY(!Perl_isinfnan(was)))
+#else
+                LIKELY(!Perl_isinf(was))
 #endif
                 ) {
 		/* diag_listed_as: Lost precision when %s %f by 1 */
--- a/sv.c
+++ b/sv.c
@@ -8953,7 +8953,7 @@ Perl_sv_inc_nomg(pTHX_ SV *const sv)
             /* If NVX was NaN, the following comparisons return always false */
             UNLIKELY(was >= NV_OVERFLOWS_INTEGERS_AT ||
                      was < -NV_OVERFLOWS_INTEGERS_AT) &&
-#if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
+#if defined(NAN_COMPARE_BROKEN)
             LIKELY(!Perl_isinfnan(was))
 #else
             LIKELY(!Perl_isinf(was))
@@ -9139,7 +9139,7 @@ Perl_sv_dec_nomg(pTHX_ SV *const sv)
                 /* If NVX was NaN, these comparisons return always false */
                 UNLIKELY(was <= -NV_OVERFLOWS_INTEGERS_AT ||
                          was > NV_OVERFLOWS_INTEGERS_AT) &&
-#if defined(NAN_COMPARE_BROKEN) && defined(Perl_isnan)
+#if defined(NAN_COMPARE_BROKEN)
                 LIKELY(!Perl_isinfnan(was)))
 #else
                 LIKELY(!Perl_isinf(was))
--- a/t/op/inc.t
+++ b/t/op/inc.t
@@ -255,6 +255,26 @@ EOC
 		      "$description under use warnings 'imprecision'");
     }
 
+    # Verify warnings on incrementing/decrementing large values
+    # whose integral part will not fit in NVs. [GH #18333]
+    foreach ([$start_n - 4, '$i++', 'negative large value', 'inc'],
+             ['+Inf' + 0,   '$i++', '+Inf', 'inc'],
+             ['-Inf' + 0,   '$i++', '-Inf', 'inc'],
+             [$start_p + 4, '$i--', 'positive large value', 'dec'],
+             ['+Inf' + 0,   '$i--', '+Inf', 'dec'],
+             ['-Inf' + 0,   '$i--', '-Inf', 'dec']) {
+	my ($start, $action, $description, $act) = @$_;
+	my $code = eval << "EOC" or die $@;
+sub {
+    use warnings 'imprecision';
+    my \$i = \$start;
+    $action;
+}
+EOC
+        warning_like($code, qr/Lost precision when ${act}rementing /,
+                     "${act}rementing $description under use warnings 'imprecision'");
+    }
+
     $found = 1;
     last;
 }
--- a/t/op/inc.t
+++ b/t/op/inc.t
@@ -258,11 +258,7 @@ EOC
     # Verify warnings on incrementing/decrementing large values
     # whose integral part will not fit in NVs. [GH #18333]
     foreach ([$start_n - 4, '$i++', 'negative large value', 'inc'],
-             ['+Inf' + 0,   '$i++', '+Inf', 'inc'],
-             ['-Inf' + 0,   '$i++', '-Inf', 'inc'],
-             [$start_p + 4, '$i--', 'positive large value', 'dec'],
-             ['+Inf' + 0,   '$i--', '+Inf', 'dec'],
-             ['-Inf' + 0,   '$i--', '-Inf', 'dec']) {
+             [$start_p + 4, '$i--', 'positive large value', 'dec']) {
 	my ($start, $action, $description, $act) = @$_;
 	my $code = eval << "EOC" or die $@;
 sub {
@@ -422,5 +418,29 @@ SKIP: {
         skip "the uvsize $Config{uvsize} is neither 4 nor 8", 2;
     }
 } # SKIP
+
+# Incrementing/decrementing Inf/NaN should not trigger 'imprecision' warnings
+# [GH #18333, #18388]
+# Note these tests only check for warnings; t/op/infnan.t has tests that
+# checks the result of incrementing/decrementing Inf/NaN.
+foreach my $infnan ('+Inf', '-Inf', 'NaN') {
+    my $start = $infnan + 0;
+  SKIP: {
+      skip "NV does not have $infnan", 2
+          unless ($infnan eq 'NaN' ? $Config{d_double_has_nan} : $Config{d_double_has_inf});
+      foreach (['$i++', 'inc'],
+               ['$i--', 'dec']) {
+          my ($action, $act) = @$_;
+          my $code = eval <<"EOC" or die $@;
+sub {
+    use warnings 'imprecision';
+    my \$i = \$start;
+    $action;
+}
+EOC
+          warning_is($code, undef, "${act}rementing $infnan under use warnings 'imprecision'");
+      }
+    } # SKIP
+}
 
 done_testing();


sisyphus:
This is a warning that I can't ever recall having seen before.

I built blead using the patched sv.c and t/op/inc.t.

IIU the diagnostics.pm elaboration correctly, it's stating  that the warning will be issued when and only when "the target of ++ or -- is unchanged".

If that's the rule we wish to follow then there's still a bit of work to do - because there are some instances when ++/-- alters the value (by 2), yet the warning is still issued.
For example, the 54-bit integer value 18014398509481982 as a binary string consists of <code>('1' x 53) . '0'</code>
It is exactly represented by the double 1.8014398509481982e+16.

```
use warnings;
my $d = 1.8014398509481982e+16;
printf "%a %.16e\n", $d, $d;
$d++;
printf "%a %.16e\n", $d, $d;

__END__
Outputs:
0x1.fffffffffffffp+53 1.8014398509481982e+16
Lost precision when incrementing 18014398509481982.000000 by 1 at try.pl line 4.
0x1p+54 1.8014398509481984e+16
```

Similarly , when 1.8014398509481982e+16 is decremented:

```
use warnings;
my $d = 1.8014398509481982e+16;
printf "%a %.16e\n", $d, $d;
$d--;
printf "%a %.16e\n", $d, $d;

__END__
Outputs:
0x1.fffffffffffffp+53 1.8014398509481982e+16
Lost precision when decrementing 18014398509481982.000000 by 1 at try.pl line 4.
0x1.ffffffffffffep+53 1.8014398509481980e+16
```

In both cases the value changed, but the warning still appeared.

Another example (there are many) is 1.0808639105689190e16:

```
use warnings;
my $d = 1.0808639105689190e16;
printf "%a %.16e\n", $d, $d;
$d++;
printf "%a %.16e\n", $d, $d;

__END__
Outputs
0x1.3333333333333p+53 1.0808639105689190e+16
Lost precision when incrementing 10808639105689190.000000 by 1 at try.pl line 4.
0x1.3333333333334p+53 1.0808639105689192e+16 - in binary ('1001' x13) . 0
```
and

```
use warnings;
my $d = 1.0808639105689190e16;
printf "%a %.16e\n", $d, $d;
$d--;
printf "%a %.16e\n", $d, $d;

__END__
Outputs
0x1.3333333333333p+53 1.0808639105689190e+16
Lost precision when decrementing 10808639105689190.000000 by 1 at try.pl line 4.
0x1.3333333333332p+53 1.0808639105689188e+16
```

In short, any binary integer of the form <code> '1' . ('X' x 51) . '10'</code> (where the 51 'X' symbols represent all combinations of zeros and ones) will exhibit this behaviour.

I expect (untested)  a similar situation with 65-bit integers ending in '10' on -Duselongdouble builds (where the long double is the 80-bit extended precision type) and 114-bit integers ending in '10' on builds where the nvtype is the 128-bit fully quad type.

But I haven't got any other quibbles at this stage.
Are we prepared to overlook these cases I've outlined ?

I also wondered about what should happen when subnormals are given a ++ or --.
In these cases it seems that the warning is never given. That seems fine since the values always change - though they don't change by exactly 1.

One last thing - the diagnostics.pm elaboration seems to harp back to an earlier time, when the usual ivsize was 4:

```
(W imprecision) The value you attempted to increment or decrement by one
is too large for the underlying floating point representation to store
accurately, hence the target of ++ or -- is unchanged.  Perl issues this
warning because it has already switched from integers to floating point
when values are too large for integers, and now even floating point is
insufficient.  You may wish to switch to using Math::BigInt explicitly.
```

The stuff about "Perl issues this warning because it has already switched from integers to floating point when values are too large for integers" is incorrect when ivsize and nvsize are both 8 .

Cheers,
Rob

tonycoz:
> If that's the rule we wish to follow then there's still a bit of work to do - because there are some instances when ++/-- alters the value (by 2), yet the warning is still issued.
> For example, the 54-bit integer value 18014398509481982 as a binary string consists of `('1' x 53) . '0'`
> It is exactly represented by the double 1.8014398509481982e+16.
> 
> ```
> use warnings;
> my $d = 1.8014398509481982e+16;
> printf "%a %.16e\n", $d, $d;
> $d++;
> printf "%a %.16e\n", $d, $d;
> 
> __END__
> Outputs:
> 0x1.fffffffffffffp+53 1.8014398509481982e+16
> Lost precision when incrementing 18014398509481982.000000 by 1 at try.pl line 4.
> 0x1p+54 1.8014398509481984e+16
> ```

That sounds like a bug in the documentation (and in the original commit message for b88df9907a8d) rather than the implementation.

sisyphus:
OK - thanks @tonycoz.

So, summarising the case for doubles containing integer values we have:
1) For 53-bit integer values, ++/-- will increment/decrement by 1, and not issue the warning.
2) As we work our way through all of the 54-bit integer values that can be expressed exactly as doubles, we will find that ++/-- will alternate between not changing the value at all and changing it by 2 - and in all of those cases it will issue the warning.
3) For 55-bit and greater integer values that can be expressed exactly as doubles, ++/-- will not change the values at all, and the warning will be issued.

Sounds reasonable to me.

The case for Inf and NaN is a little puzzling.
If we ++/-- a NaN we end up with no change (ie it's still a NaN) and no warning.
If we ++/-- an Inf we end up with no change (ie it's still an Inf of the same sign), but this time we do see the warning.

So ... you could argue that ++/-- on a Nan, is not expected to change the value, so the warning is inappropriate.
But then you could also argue that ++/-- on an Inf is not expected to change the value, so the warning is again inappropriate .
(Yet, for the Infs, the warning is issued.)

I think my preference would be to warn for NaN as well as Infs.
I've also an inkling that it should be a slightly different warning, as IMO it's not really related to "precision" - eg something like "Inf and Nan are unaltered by ++ or --".
But I'm not going to quibble over any of that.

And I'll just mention the case of subnormal fractional values again.
With them, ++/-- will always alter the value, and will never issue a warning.
That sounds entirely reasonable. My only concern is that the change in value when a subnormal is incremented/decremented is arguably not exactly one. (It depends on how you argue.)

I'm thinking we could leave that as is ? ... at least until someone complains about it.

The demos:

```
use warnings;
my $d = (2 ** -1025) + (2 ** -1040) + (2 ** - 1074);
printf "%a %.16e\n", $d, $d;
$d++;
printf "%a %.16e\n", $d, $d;

__END__
Outputs
0x1.0002000000008p-1025 2.7814272029656453e-309
0x1p+0 1.0000000000000000e+00
```
and

```
use warnings;
my $d = (2 ** -1025) + (2 ** -1040) + (2 ** - 1074);
printf "%a %.16e\n", $d, $d;
$d--;
printf "%a %.16e\n", $d, $d;

__END__
Outputs
0x1.0002000000008p-1025 2.7814272029656453e-309
-0x1p+0 -1.0000000000000000e+00
```

Cheers,
Rob

hvds:
> The case for Inf and NaN is a little puzzling.
> If we ++/-- a NaN we end up with no change (ie it's still a NaN) and no warning.
> If we ++/-- an Inf we end up with no change (ie it's still an Inf of the same sign), but this time we do see the warning.
> 
> So ... you could argue that ++/-- on a Nan, is not expected to change the value, so the warning is inappropriate.
> But then you could also argue that ++/-- on an Inf is not expected to change the value, so the warning is again inappropriate .
> (Yet, for the Infs, the warning is issued.)
> 
> I think my preference would be to warn for NaN as well as Infs.
> I've also an inkling that it should be a slightly different warning, as IMO it's not really related to "precision" - eg something like "Inf and Nan are unaltered by ++ or --".
> But I'm not going to quibble over any of that.

FWIW, I think it would make more sense _not_ to warn on Inf. It revolves around whether you have an Inf knowingly/deliberately or not - if your code expects to deal with Inf, then you'd expect an increment to give Inf and no warning is needed (and I'd argue there is no loss of precision [1]); if not, you want the warning at the point you inadvertently generated the Inf, not at the point you try to do further work on it.

The case for NaN is much less obvious to me, but my first stab would be to apply the same argument to say it also should not warn.

[1] That's assuming it's a cardinal infinity; I don't think anyone sane would expect simple perl scalars to support infinite ordinals.

t-a-k:
> And I'll just mention the case of subnormal fractional values again.
> With them, ++/-- will always alter the value, and will never issue a warning.
> That sounds entirely reasonable. My only concern is that the change in value when a subnormal is incremented/decremented is arguably not exactly one. (It depends on how you argue.)

I think that there is the same symptom in normal range, not only in subnormal values.
When the increment will step over 2 ** n, the LSB of fractional part is rounded so that the change in value may not be exact one.

```perl
my $d = 3.1;
printf "%a %.17g\n", $d, $d;
my $beforeinc = $d++;
printf "%a %.17g\n", $d, $d;
printf "%.17g\n", $d - $beforeinc;
```
Results:
```
0x1.8cccccccccccdp+1 3.1000000000000001
0x1.0666666666666p+2 4.0999999999999996
0.99999999999999956
```

Anyway, I agree that it is reasonable to issue no warnings in these cases.

sisyphus:
Just trying to think of how we describe the conditions under which the warning is given.
Taking into account the views of @t-a-k and @hvds, I'm thinking:

The warning is emitted if and only if an NV containing a real, integer value does not change by exactly 1 when operated upon by either the <code>++</code> or <code>--</code> operator.

Does that accurately cover our intentions ?

There's no  need to mention the case of real, non-integer values as (I think) they will always be altered by some amount, and never trigger the warning. Right ?
And if we're not going to warn on Inf and NaN then there's no need to mention them, either.

@hvds, I still have a slight preference for a warning being given for Infs and NaN. (Though not the same "loss of precision" warning, I agree.) My reasoning (which I failed to provide earlier) is that incrementing/decrementing an Inf or Nan is a pretty strong indication that the program might not have been thought through all that well - and that's typically the type of scenario under which warnings are given.

Cheers,
Rob

hvds:
> Just trying to think of how we describe the conditions under which the warning is given.
> Taking into account the views of @t-a-k and @hvds, I'm thinking:
> 
> The warning is emitted if and only if an NV containing a real, integer value does not change by exactly 1 when operated upon by either the `++` or `--` operator.
> 
> Does that accurately cover our intentions ?

That sounds reasonable to me, I can't readily think of a counterexample. I'd probably replace "real" with something like "concrete" or "actual", to avoid possible confusion.

> There's no need to mention the case of real, non-integer values as (I think) they will always be altered by some amount, and never trigger the warning. Right ?
> And if we're not going to warn on Inf and NaN then there's no need to mention them, either.
> 
> @hvds, I still have a slight preference for a warning being given for Infs and NaN. (Though not the same "loss of precision" warning, I agree.) My reasoning (which I failed to provide earlier) is that incrementing/decrementing an Inf or Nan is a pretty strong indication that the program might not have been thought through all that well - and that's typically the type of scenario under which warnings are given.

I would not be strongly against a blanket warning for any arithmetic on Inf, which could be turned off by anyone doing so intentionally. It would be regrettable if it could only be disabled by turning off the whole loss-of-precision warning, and singling out inc/dec for such a warning on Inf doesn't seem sensible.

Like you, though, I don't feel strongly about this: I can _imagine_ having a use for such code, but I have never actually had one.

sisyphus:
By "real", I was (perhaps unnecessarily) wishing to indicate that the "integer value" was neither Inf nor NaN .
The mpfr library documentation uses the adjective "ordinary" to describe numbers that are neither Inf nor NaN.

Cheers,
Rob

khwilliamson:
@t-a-k What is the status of this?

t-a-k:
> @t-a-k What is the status of this?

Based on the discussion so far, it seems better to suppress imprecision warnings on Inf/NaN.
`sv.c` in c33ee94 warns on Inf, so I'll update this PR soon.

khwilliamson:
@sisyphus please comment on the revised version

sisyphus:
I think this fixes the the behaviour wrt infinities.
However, I now find that there's another aspect that should probably be corrected.
On a perl whose nvtype is double, for integer values whose absolute value is >= 2 ** 54, I'm seeing the "Lost precision" warning either for <code>--</code> for for <code>++</code> - but not for both operations.
Yet, for these values neither <code>--</code> nor <code>++</code> alters them - which, AIUI, means that *both* operations should emit the warning.
For example, using @t-a-k's fork as it presently stands, I'm getting : 
```
$ ./perl -I./lib -wle '$x = (2 ** 54);  $x--;print "unchanged" if $x == 2 ** 54;'
unchanged
$ ./perl -I./lib -wle '$x = (2 ** 54);  $x++;print "unchanged" if $x == 2 ** 54;'
Lost precision when incrementing 18014398509481984.000000 by 1 at -e line 1.
unchanged
$ ./perl -I./lib -wle '$x = -(2 ** 54);  $x++;print "unchanged" if $x == -(2 ** 54);'
unchanged
$ ./perl -I./lib -wle '$x = -(2 ** 54);  $x--;print "unchanged" if $x == -(2 ** 54);'
Lost precision when decrementing -18014398509481984.000000 by 1 at -e line 1.
unchanged
```
The same behaviour was present in 5.32.0, so I don't think this is a regression introduced by the changes that @t-a-k  has proposed.
Instead, I think it's just something we hadn't noticed earlier.
I presume (untested) that the same type of issue extends to perls whose nvtype is either "long double" or "__float128". 

My apologies (especially to @t-a-k ) for not having detected this earlier.

Cheers,
Rob

t-a-k:
> On a perl whose nvtype is double, for integer values whose absolute value is >= 2 ** 54, I'm seeing the "Lost precision" warning either for `--` for for `++` - but not for both operations.
> Yet, for these values neither `--` nor `++` alters them - which, AIUI, means that _both_ operations should emit the warning.
> For example, using @t-a-k's fork as it presently stands, I'm getting :

I think that something went wrong.  This is what this PR should fix, and after applying commit 66435b24ea or c33ee94ba2 (these two behave same for finite values; differ only for &plusmn;Inf), perl should emit the warning for all these cases:

```
$ ./perl -Ilib -V:git_commit_id
git_commit_id='66435b24ea6b4fb597c18f831558183d4f8b7f22';
$ ./perl -I./lib -wle '$x = (2 ** 54);  $x--;print "unchanged" if $x == 2 ** 54;'
Lost precision when decrementing 18014398509481984.000000 by 1 at -e line 1.
unchanged
$ ./perl -I./lib -wle '$x = (2 ** 54);  $x++;print "unchanged" if $x == 2 ** 54;'
Lost precision when incrementing 18014398509481984.000000 by 1 at -e line 1.
unchanged
$ ./perl -I./lib -wle '$x = -(2 ** 54);  $x++;print "unchanged" if $x == -(2 ** 54);'
Lost precision when incrementing -18014398509481984.000000 by 1 at -e line 1.
unchanged
$ ./perl -I./lib -wle '$x = -(2 ** 54);  $x--;print "unchanged" if $x == -(2 ** 54);'
Lost precision when decrementing -18014398509481984.000000 by 1 at -e line 1.
unchanged
```
These commits also add tests for these cases in `t/op/inc.t`.

sisyphus:
The something that went wrong was probably me.
Looks like I've used the wrong damn source :-(

What's the simplest way for me to grab the "right" source from scratch ?

sisyphus:
Sure makes a difference when I use the right source.
I've just manually applied your patches to current blead (commit 069831c529609317bcba91d61b3bc6568463d8ad) and all looks fine to me.

Cheers,
Rob

PR #18394, opened 2020-12-10
Remove workaround for distros needing dot in @INC

jkeenan:
In commit 19641fd71a (Apr 07 2017), as part of ceasing to pass '.' to
@INC in tests, we added a workaround to t/TEST for CPAN distributions
which ship with core which were not yet fully adapted to the new
regulation on @INC.

All such CPAN distributions have now been adapted.  Hence, we can remove
the workaround.

### DIFF
--- a/t/TEST
+++ b/t/TEST
@@ -76,16 +76,6 @@ my %temp_no_core = (
      '../dist/Unicode-Normalize' => 1,
     );
 
-# temporary workaround Apr 2017.  These need '.' in @INC.
-# Ideally this # list will eventually be empty
-
-my %temp_needs_dot  = map { $_ => 1 } qw(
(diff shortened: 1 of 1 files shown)


jkeenan:
> In commit [19641fd](https://github.com/Perl/perl5/commit/19641fd71ab87a4234f86f69f958b61278dbdade) (Apr 07 2017), as part of ceasing to pass '.' to
> @inc in tests, we added a workaround to t/TEST for CPAN distributions
> which ship with core which were not yet fully adapted to the new
> regulation on @inc.
> 
> All such CPAN distributions have now been adapted. Hence, we can remove
> the workaround.

Are there any objections to this change to t/test.pl?

Thank you very much.
Jim Keenan

Leont:
LGTM

PR #18395, opened 2020-12-11
Distinguish between C- and perly- literals

happy-barney:
to make it clear when `toke.c` works with input characters and when with perly symbols.


khwilliamson:
If I don't hear reasons to the contrary, I will merge this on December 27.

PR #18396, opened 2020-12-12
read-only lock for getenv(), fix PERL_MEM_LOG on all platforms

khwilliamson:
I don't know if the asserts and changing the count of readers in the first commit are a good idea.  Perhaps a panic would be a better action.


Configure
```diff
@@ -14206,6 +14207,82 @@ eval $inlibc
 set getcwd d_getcwd
 eval $inlibc
 
+: check for getenv behavior
```
tonycoz, line 14210: This still performs the test if a value is supplied on the command-line:
```
$ ./Configure -de -Dusedevel -Ud_getenv_preserves_other_thread
....
Checking to see if getenv() preserves a different thread's results
getenv() didn't destroy another thread's buffer
Hmm.  Based on the hints in hints/linux.sh, 
the recommended value for $d_getenv_preserves_other_thread on this machine was "undef"!
    Keep the recommended value? [y]  
...
```

Since this is a relatively expensive test (it builds a program and runs it) it might be good to skip testing if a value is already supplied.  Though @Tux might be the one to ask.

The test itself potentially invokes undefined behaviour, so it might pass at Configure time but the implementation then overwrites in some cases at runtime.

In an ideal world we'd check the documentation of each platform for the equivalent of the "MT-Safe env" from the getenv() documentation for glibc and update the hints files to match, but I don't see us doing that for anything but the major platforms.

It would be good if setting such hints didn't produce a Whoa!

khwilliamson: This looks like a bug

inline.h
```diff
@@ -2618,23 +2618,30 @@ Perl_mortal_getenv(const char * str)
...
+     * mutex, but it is possible for code from, say XS, to not use this mutex,
+     * defeating the safety.
+     * 
+     * getenv() returns, in some implementations, a pointer to a spot in the
+     * **environ array, which could be invalidated at any time by this or
+     * another thread changing the environment.  Other implementations copy the
+     * **environ value to a static buffer, returning a pointer to that.  That
+     * may be a per-process rather than per-thread buffer, in which case this
+     * routine could be destroying another thread's value, and we need an
+     * exclusive lock.  Otherwise, a many-reader lock is ok.  There is a
```
tonycoz: This claims there might be a single buffer, but that isn't what POSIX says:

> The returned string pointer might be invalidated or [Option End]  the string content might be overwritten by a subsequent call to getenv(), [CX] [Option Start] setenv(), unsetenv(), [Option End]
[XSI] [Option Start]  or (if supported) putenv() [Option End]  but they shall not be affected by a call to any other function in this volume of POSIX.1-2017.

It might be a static buffer, it might be a malloced pointer which is then freed on the next call to getenv().

It doesn't (I think) make any difference for this implementation, might it might mislead someone into thinking the pointer from getenv() remains valid but pointing to a changed buffer, when this isn't the case.

khwilliamson: I can generalize it.

It doesn't make a difference to this implementation, provided the probe worked correctly, and that every thread is using this mutex.  Mutexes only work on Posix boxes as long as everyone agrees to use them, and doesn't try an unauthorized access.  So, if another thread calls getenv directly without the mutex, it could be a problem if that getenv destroyed this one.  But we have that problem in existing code, and this mutex should remove it as a problem for all cooperating code.

Configure
tonycoz: To avoid the message typically you'd check the current value, something like:

```
case "$d_getenv_preserves_other_thread" in
"")
(your probe code here)
;;
esac
```

inline.h
khwilliamson: The new version has changed the wording

Configure
khwilliamson: I have added this

khwilliamson:
Failing any further comments, and wanting to get this in the next development release, I have merged this with 03694582f8c247d4a1cc8a7bb8348af0173944d7

PR #18398, opened 2020-12-14
perlfunc: Improve localtime entry

khwilliamson:
This rearranges some paragraphs that really belong together, and
clarifies the result is always in English


Grinnz:
I wonder if POSIX::asctime is locale sensitive. It is basically a specialized version of POSIX::strftime that returns the same format as scalar(localtime) except it has a trailing newline.

khwilliamson:
It returns always English.  Theglibc  manpage says the locale should not be changed by another thread during its execution.  I looked at the glibc source code and it wasn't clear why that would be.  The English is hard-coded in.

khwilliamson:
I've now added notes to the asctime entry

khwilliamson:
This PR was addressed by 62e9eaa93fa2015420d8cdcaf7c1e83d1c846e57 and ff22dd7054d2f46d7451cabb1722dcaa04f8a3d6

PR #18401, opened 2020-12-14
Mark some functions as core-only

khwilliamson:
This is a bunch of these, accumulated into one PR, but it's fine to say some shouldn't be merged.

Note that all this does is move the mention of these to perlintern from perlapi


khwilliamson:
These were separately merged via  e39fb0f and surrounding commits

PR #18407, opened 2020-12-17
Make DynaLoader on MacOS check library existence with dlopen

Leont:
A number of system libraries no longer exist as actual files on Big Sur, even though dlopen will pretend they do, so now we fall back to dlopen if a library file can not be found.

[CI](https://github.com/Perl/perl5/runs/1572311015?check_suite_focus=true) suggests this works on Big Sur but this probably needs human confirmation that it will fix the issue from someone who actually runs Big Sur before being merged.

PR #18408, opened 2020-12-17
Correctly set LC_CTYPE back in POSIX/t/time.t

Leont:
This appears to be a copy-paste mistake

### DIFF
--- a/ext/POSIX/t/time.t
+++ b/ext/POSIX/t/time.t
@@ -86,7 +86,7 @@ if (locales_enabled('LC_TIME')) {
     setlocale(LC_TIME, $orig_time_loc) || die "Cannot setlocale(LC_TIME) back to orig: $!";
 }
 if (locales_enabled('LC_CTYPE')) {
-    setlocale(LC_TIME, $orig_ctype_loc) || die "Cannot setlocale(LC_CTYPE) back to orig: $!";
+    setlocale(LC_CTYPE, $orig_ctype_loc) || die "Cannot setlocale(LC_CTYPE) back to orig: $!";
 }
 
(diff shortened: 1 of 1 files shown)


hakonhagland:
Will this be backported to 5.32.0 ?

I still get [the error](https://github.com/Perl/perl5/pull/18407#issuecomment-747662048) `Cannot setlocale(LC_CTYPE) back to orig`) when trying to install 5.32.0 with perlbrew on macOS Big Sur (after running latest version of patchperl on the source tree).

steve-m-hay:
> Will this be backported to 5.32.0 ?
> 

This has just been backported to maint-5.32, so will be in 5.32.1, which will hopefully be released later this month.

hakonhagland:
Great! So only critical patches are imported to [devel-patchperl](https://github.com/bingos/devel-patchperl) then? Or am I missing something, since [some other patches](https://github.com/bingos/devel-patchperl/commit/3d44f83c8ccc5eba3ad3f449bab2a985464b7726) were recently imported, but this patch was not.

steve-m-hay:
> Great! So only critical patches are imported to [devel-patchperl](https://github.com/bingos/devel-patchperl) then? Or am I missing something, since [some other patches](https://github.com/bingos/devel-patchperl/commit/3d44f83c8ccc5eba3ad3f449bab2a985464b7726) were recently imported, but this patch was not.

Sorry, I'm not familiar with devel-patchperl. Perhaps @bingos can explain?

Leont:
> Great! So only critical patches are imported to devel-patchperl then? Or am I missing something, since some other patches were recently imported, but this patch was not.

Devel::PatchPerl is maintained outside of p5p, you should ask BinGOs there.

PR #18410, opened 2020-12-17
Define integer constants as IV rather than NV in POSIX

t-a-k:
In `POSIX` module, some constants related to floating-point types such as `DBL_DIG`, `FLT_RADIX`, `FP_*` (return values for `fpclassify`), etc. had been defined as NV, but these are actually integer constants.  This PR changes these constants to IV.

This change should not affect the behavior of user programs.


sisyphus:
For the POSIX module to be changing integer constants to NV constants is quite unexpected - so this proposal looks fine to me.

It could perhaps break XS code that does a callback to one of the affected POSIX constants, expecting it to be an NV, when it has been changed to an IV.
But that would be a rare and rather convoluted thing to do ... I mean why would you call (say) POSIX::DBL_DIG from XS code when you can just call DBL_DIG.

Cheers,
Rob

PR #18411, opened 2020-12-18
Fix RT number in perl5260delta.pod

thibaultduponchelle:
Hello,

I propose this documentation change since 129340 is a dead link and 130766 seems to be the corresponding issue and was merged in 5.24.1 and 5.26.0

Thibault


jkeenan:
Merged to blead in commit 34ceef7614.  Thank you very much.

PR #18415, opened 2020-12-20
Document SvPV_shrink_to_cur

khwilliamson:


sv.h
```diff
@@ -1389,6 +1389,15 @@ object type. Exposed to perl code via Internals::SvREADONLY().
 				(char*)saferealloc((Malloc_t)SvPVX(sv), \
 						   (MEM_SIZE)((n)))));  \
 		 } STMT_END
+/*
+=for apidoc Am|void|SvPV_shrink_to_cur|SV* sv
+Trim any trailing unused memory in the PV of C<sv>, which needs to have a real
+C<PV> that is unencombered by things like COW.  Using
```
tomhukins: s/unencombered/unencumbered/

tomhukins (approved):

tonycoz:
I think it could use a warning similar to that of SvPV_renew(), to use it you need to remove COW and it makes the SV non-COWable afterwards (since SvLEN() is too small).

jkeenan:
> I think it could use a warning similar to that of SvPV_renew(), to use it you need to remove COW and it makes the SV non-COWable afterwards (since SvLEN() is too small).

@khw, can you respond to @tonycoz's comment?

Thank you very much.
Jim Keenan

khwilliamson:
@tonycoz I don't see any documentation for SvPV_renew nor is any diagnostic raised when using it.  So I don't understand your comment.

tonycoz:
> 
> 
> @tonycoz I don't see any documentation for SvPV_renew nor is any diagnostic raised when using it. So I don't understand your comment.

It's documented here:

https://github.com/Perl/perl5/blob/8c5e15d0d75a75ec52cba21d188852c305c161fc/sv.h#L1387

and that documentation warns:

```
...  It is generally better to use
C<SvGROW> instead.  This is because C<SvPV_renew> ignores potential issues that
C<SvGROW> handles.  C<sv> needs to have a real C<PV> that is unencombered by
things like COW.  Using C<SV_CHECK_THINKFIRST> or
C<SV_CHECK_THINKFIRST_COW_DROP> before calling this should clean it up, but
why not just use C<SvGROW> if you're not sure about the provenance?
```

khwilliamson:
I thought I had rebased, but hadn't; hence I didn't see that documentation.
But I still don't understand your point.  This commit has many of the same warnings as the 'renew' pod does.  It doesn't say it is better to use something else, because AFAIK, this is the function for the purpose.

tonycoz:
> But I still don't understand your point. This commit has many of the same warnings as the 'renew' pod does. It doesn't say it is better to use something else, because AFAIK, this is the function for the purpose.

I mean in general it's better not to use this macro at all.

khwilliamson:
What's the alternative?

tonycoz:
> 
> 
> What's the alternative?

There's no alternative, in 99% of cases you don't need or want to shrink_to_cur(), not just because the SV needs to be a non-COW PV, but because it prevents future COW on the SV, and if extra space is needed in the future, the memory will need to be reallocated again anyway.

PR #18416, opened 2020-12-20
Document SvPV_renew

khwilliamson:


sv.h
```diff
@@ -1383,6 +1383,17 @@ object type. Exposed to perl code via Internals::SvREADONLY().
 	STMT_START { assert(SvTYPE(sv) >= SVt_PV); \
 		SvCUR_set(sv, (val) - SvPVX(sv)); } STMT_END
 
+/*
+=for apidoc Am|void|SvPV_renew|SV* sv|STRLEN len
+Low level micro optimization of C<L</SvGROW>> which you should generally use
```
jkeenan: What is the antecedent of *which* in the line above: `SvGROW` or `SvPV_renew`?

khwilliamson: See the new version

jkeenan: +1

PR #18422, opened 2020-12-20
Document gv_autoload4

khwilliamson:


gv.h
```diff
@@ -269,6 +269,13 @@ Return the CV from the GV.
 #define gv_fetchmeth(stash,name,len,level) gv_fetchmeth_pvn(stash, name, len, level, 0)
 #define gv_fetchmeth_autoload(stash,name,len,level) gv_fetchmeth_pvn_autoload(stash, name, len, level, 0)
 #define gv_fetchmethod_flags(stash,name,flags) gv_fetchmethod_pv_flags(stash, name, flags)
+
+/*
+=for apidoc gv_autoload4
+Equivalent to C<L</gv_autoload_pvn>>.
```
jkeenan, line 275: Is there a reason why we have two equivalent functions?

khwilliamson: I have no idea; I'm just documenting current behavior

PR #18423, opened 2020-12-20
Document PERLIO_FUNCS_(DECL|CAST)

khwilliamson:


perlio.h
```diff
@@ -63,6 +63,16 @@ typedef PerlIOl *PerlIO;
 #define PerlIO PerlIO
 #define PERLIO_LAYERS 1
 
+/*
+=for apidoc_section $io
+=for apidoc Amu||PERLIO_FUNCS_DECL|PerlIO * func
+Declare C<func> to be a PerlIO function, that is, of type C<PerlIO_funcs>.
```
Leont, line 69: It doesn't define a function, it defines a PerlIO vtable

jkeenan: @khwilliamson, can you respond to @LeonT's question -- and also advise as to the status of this p.r.?

PR #18425, opened 2020-12-20
Correct one misspelled variable name

jkeenan:
Correct one other typo


jkeenan:
This is the first, and simplest, of a series of pull requests I will make to make `lib/B/Deparse-core.t` easier for us mortals to understand and work with.  Coming up: tidying its leading whitespace.

Thank you very much.
Jim Keenan

Issue #18413, opened 2020-12-19
Including a new POD file requires much work, is inadequately documented


jkeenan:
As I learned the hard way today while working on https://github.com/Perl/perl5/pull/18275, including a new POD file in the core distribution is not as simple as creating a correctly formatted file in `.pod` format, placing it in the `pod/` directory, and inserting a new entry into the proper location in the `MANIFEST`.  Because listings to `pod/*.pod` files are found in several places in the core in addition to `MANIFEST`, the task is non-trivial and would benefit from a combination of further documentation and further automation.

What follows is my re-creation of the porting problem I faced this morning for `pod/perldocstyle.pod`.  It's quite detailed, but it has to be, because even to re-create this problem for the purpose of filing this ticket took me three hours.

This will mainly be of interest to Release Managers and to commit-bit holders who apply patches and pull requests from regular contributors.  So if you don't want to get down into the weeds, you can stop reading this ticket here and simply wait for pull requests which I hope to file soon.

Thank you very much.
Jim Keenan

**BEGIN Gory Details**

As an experiment, I created a branch called `jkeenan/phonypod`.  In that branch I created a new POD file, `pod/perlphonypod.pod`, which, per `podchecker` was correctly formatted in POD format.  I also created an entry for that file in `MANIFEST`.
```
$ git show | cat
commit 2b1ee02941a2ec5a2774fe0134ad9269eb1a2339
Author: James E Keenan <jkeenan@cpan.org>
Date:   Sat Dec 19 19:52:16 2020 +0000

    Add dummy copy file for testing

diff --git a/MANIFEST b/MANIFEST
index 6e9f6c64a2..295ab0e12c 100644
--- a/MANIFEST
+++ b/MANIFEST
@@ -5293,6 +5293,7 @@ pod/perlop.pod			Perl operators and precedence
 pod/perlopentut.pod		Perl open() tutorial
 pod/perlpacktut.pod		Perl pack() and unpack() tutorial
 pod/perlperf.pod		Perl Performance and Optimization Techniques
+pod/perlphonypod.pod		Phony POD
 pod/perlpod.pod			Perl plain old documentation
 pod/perlpodspec.pod		Perl plain old documentation format specification
 pod/perlpodstyle.pod		Perl POD style guide
diff --git a/pod/perlphonypod.pod b/pod/perlphonypod.pod
new file mode 100644
index 0000000000..9fbaa47c76
--- /dev/null
+++ b/pod/perlphonypod.pod
[...]

PR #18429, opened 2020-12-21
Better documentation for two Porting/*.pl programs

jkeenan:
In partial satisfaction of https://github.com/Perl/perl5/issues/18413.


khwilliamson (commented):
LGTM but I don't know what I'm talking about

Issue #18405, opened 2020-12-17
[feature] Make signature errors give arg counts


FGasper:
**Describe the solution you'd like**

```
> perl -e'use experimental "signatures"; sub foo ($w) { }; foo()'
Too few arguments for subroutine 'main::foo' at -e line 1.
```
It’d be nice if this instead said something like:
> Too few arguments (0 given, 1 expected) for subroutine …

Likewise, the too-many error could be:
> Too many arguments (2 given, 1 expected) for subroutine …

I’d be happy to try implementing this if it’s an agreeable improvement.

khwilliamson:
On 12/17/20 7:05 AM, Felipe Gasper wrote:
> *Describe the solution you'd like*
> A clear and concise description of what you want to happen.
> 
> |> perl -e'use experimental "signatures"; sub foo ($w) { }; foo()' Too 
> few arguments for subroutine 'main::foo' at -e line 1. |
> 
> It’d be nice if this instead said something like:
> 
>     Too few arguments (0 given, 1 expected) for subroutine …
> 
> Likewise, the too-many error could be:
> 
>     Too many arguments (0 given, 1 expected) for subroutine …
> 
> I’d be happy to try implementing this if it’s an agreeable improvement.
> 
>

LGTM

PR #18432, opened 2020-12-22
Add arg counts to arg-mismatch errors.

FGasper:
Resolves issue #18405.

This is in-progress; if the code change itself looks acceptable I’ll work on the rest.

Thank you!

### DIFF
--- a/pp.c
+++ b/pp.c
@@ -7145,10 +7145,17 @@ PP(pp_argcheck)
     too_few = (argc < (params - opt_params));
 
     if (UNLIKELY(too_few || (!slurpy && argc > params)))
-        /* diag_listed_as: Too few arguments for subroutine '%s' */
-        /* diag_listed_as: Too many arguments for subroutine '%s' */
-        Perl_croak_caller("Too %s arguments for subroutine '%" SVf "'",
-                          too_few ? "few" : "many", S_find_runcv_name());
+
+        /* diag_listed_as: Too few arguments for subroutine '%s' (got %d; expected %d) */
+        /* diag_listed_as: Too few arguments for subroutine '%s' (got %d; expected at least %d) */
+        /* diag_listed_as: Too many arguments for subroutine '%s' (got %d; expected %d) */
+        /* diag_listed_as: Too many arguments for subroutine '%s' (got %d; expected at most %d)*/
+        Perl_croak_caller("Too %s arguments for subroutine '%" SVf "' (got %" UVuf "; expected %s%" UVuf ")",
+                          too_few ? "few" : "many",
+                          S_find_runcv_name(),
+                          argc,
+                          too_few ? (slurpy || opt_params ? "at least " : "") : (opt_params ? "at most " : ""),
+                          too_few ? (params - opt_params) : params);
 
     if (UNLIKELY(slurpy == '%' && argc > params && (argc - params) % 2))
         /* diag_listed_as: Odd name/value argument for subroutine '%s' */
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -218,7 +218,8 @@ XXX Changes (i.e. rewording) of diagnostic messages go here
 
 =item *
 
-XXX Describe change here
+Subroutine argument-count mismatch errors now include the number of
+given and expected arguments.
 
 =back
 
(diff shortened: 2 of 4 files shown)


pp.c
```diff
@@ -7145,14 +7145,13 @@ PP(pp_argcheck)
     too_few = (argc < (params - opt_params));
 
     if (UNLIKELY(too_few || (!slurpy && argc > params)))
-        const char* atleast = (opt_params > 0) ? " at least" : "";
 
         /* diag_listed_as: Too few arguments for subroutine '%s' */
         /* diag_listed_as: Too many arguments for subroutine '%s' */
-        Perl_croak_caller("Too %s arguments (got %d, expected%s %d) for subroutine '%" SVf "'",
+        Perl_croak_caller("Too %s arguments (got %lu, expected%s %lu) for subroutine '%" SVf "'",
```
tonycoz: The two numeric values are UVs, so this should use the UVuf macro to get the format.

FGasper: Fixed. Thank you.

pp.c
```diff
@@ -7145,10 +7145,15 @@ PP(pp_argcheck)
     too_few = (argc < (params - opt_params));
 
     if (UNLIKELY(too_few || (!slurpy && argc > params)))
+
         /* diag_listed_as: Too few arguments for subroutine '%s' */
         /* diag_listed_as: Too many arguments for subroutine '%s' */
-        Perl_croak_caller("Too %s arguments for subroutine '%" SVf "'",
-                          too_few ? "few" : "many", S_find_runcv_name());
+        Perl_croak_caller("Too %s arguments (got %" UVuf ", expected %s%" UVuf ") for subroutine '%" SVf "'",
```
arc: This should probably include an update to pod/perldiag.pod for the changed diagnostic messages (and then a corresponding update to the `diag_listed_as` cookies above).

Maybe it would also be better to move the argument count to the end of the message, as in `Too few arguments for subroutine 'Foo::foo' (got 1, expected at least 2)`.

FGasper: I put the counts after “arguments” so that the “modifier” (i.e., the counts) stayed closer to the referent (“arguments”).

What is the advantage of putting the counts at the end? (I can surmise but want to know your reasoning.)

arc: Well, I first considered it on the basis that it might make the perldiag cookie easier to construct and read. But having thought of it, it seemed to me that it would tend to make the messages read better, by having the big chunky part at the end. I don't have a terribly strong opinion on that part, though.

Leont: I would also agree with arc, but that's also a not terribly strong opinion on my side.

FGasper:
OK, I’ve moved the arg counts to the end, fixed a few bugs in the logic, updated the tests, and updated the perldiag.pod and the `diag_listed_as` comments.

I think this is getting closer to being merge-worthy. Would someone be able to guide on what I should do next?

Thank you!

jkeenan:
> 
> I think this is getting closer to being merge-worthy. Would someone be able to guide on what I should do next?
> 

I have created a smoke-me branch so that we can get some data on the impact of these changes:  smoke-me/jkeenan/FGasper-fg_argcheck_counts

Thank you very much.
Jim Keenan

FGasper:
@jkeenan Forgive my ignorance—and I’m sorry if I missed this somewhere in the contributor docs—but is that smoke-me branch public?

jkeenan:
> @jkeenan Forgive my ignorance—and I’m sorry if I missed this somewhere in the contributor docs—but is that smoke-me branch public?

It's in the main Perl 5 repository:
https://github.com/Perl/perl5/tree/smoke-me/jkeenan/FGasper-fg_argcheck_counts

You can see smoke-test reports for this branch here:
http://perl.develop-help.com/?b=smoke-me%2Fjkeenan%2FFGasper-fg_argcheck_counts

(However, because of the fact that we're only partially recovered from the loss of perl5.test-smoke.org for several months, the aggregator above only shows reports that are transmitted via email.  There are other reports, transmitted via HTTP::Tiny, that are visible only at, say, https://perl5.test-smoke.org/search?selected_osnm=freebsd&report_osnm=freebsd&selected_perl=all.)

Since you've pushed an additional commit to your pull request, that smoke-me branch is somewhat moot now.

Thank you very much.
Jim Keenan

Leont:
> @jkeenan Forgive my ignorance—and I’m sorry if I missed this somewhere in the contributor docs—but is that smoke-me branch public?

You can find the smoke results [here](http://perl.develop-help.com/?b=smoke-me%2Fjkeenan%2FFGasper-fg_argcheck_counts)

Leont:
> I think this is getting closer to being merge-worthy. Would someone be able to guide on what I should do next?

The branch should probably be cleaned up before being unmarked as draft.

PR #18433, opened 2020-12-22
Adapt debugger to warnings on.

jkeenan:
Because this program is still so heavily dependent on (package) global
variables, we'll suppress 'uninitialized value' warnings.  However, all
other warnings can either be surgically suppressed or fixed by
correcting the code generating the warnings.

Increment $VERSION.

### DIFF
--- a/lib/perl5db.pl
+++ b/lib/perl5db.pl
@@ -513,6 +513,8 @@ where it has to go.
 package DB;
 
 use strict;
+use warnings;
+no warnings 'uninitialized';
 
 use Cwd ();
 
@@ -531,7 +533,7 @@ BEGIN {
 use vars qw($VERSION $header);
 
 # bump to X.XX in blead, only use X.XX_XX in maint
-$VERSION = '1.60';
+$VERSION = '1.61';
 
 $header = "perl5db.pl version $VERSION";
 
@@ -870,6 +872,7 @@ BEGIN {
         require threads;
         require threads::shared;
         import threads::shared qw(share);
+        no warnings 'void';
         $DBGR;
         share(\$DBGR);
         lock($DBGR);
@@ -2115,6 +2118,7 @@ sub _DB__handle_c_command {
         for my $j (0 .. $stack_depth) {
             $stack[ $j ] &= ~1;
         }
+        no warnings 'exiting';
         last CMD;
     }
 
@@ -2814,6 +2818,7 @@ sub _DB__handle_l_command {
     my $self = shift;
 
     _cmd_l_main($self->cmd_args);
+    no warnings 'exiting';
     next CMD;
 }
 
@@ -3742,6 +3747,7 @@ sub _handle_t_command {
         . ( ( $trace & 1 )
             ? ( $levels ? "on (to level $DB::trace_to_depth)" : "on" )
             : "off" ) . "\n";
+        no warnings 'exiting';
         next CMD;
     }
 
@@ -3773,6 +3779,7 @@ sub _handle_S_command {
                 print $OUT $subname, "\n";
             }
         }
+        no warnings 'exiting';
         next CMD;
     }
 
@@ -3856,6 +3863,7 @@ sub _handle_dash_command {
 
         # Generate and execute a "l +" command (handled below).
         $DB::cmd = 'l ' . ($start) . '+';
+        no warnings 'exiting';
         redo CMD;
     }
     return;
@@ -3974,6 +3982,7 @@ sub _handle_rc_recall_command {
         # Print the command to be executed and restart the loop
         # with that command in the buffer.
         print {$OUT} $DB::cmd, "\n";
+        no warnings 'exiting';
         redo CMD;
     }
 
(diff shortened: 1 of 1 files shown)


Leont:
Given how fragile the debugger is, I'm slightly worried that «use warnings» will get enabled in places where it shouldn't be. I think I recall such things happening when it was made strict.

I might be completely mistaken though, this will require some careful analysis.

jkeenan:
> Given how fragile the debugger is, I'm slightly worried that «use warnings» will get enabled in places where it shouldn't be. I think I recall such things happening when it was made strict.
> 
> I might be completely mistaken though, this will require some careful analysis.

It is certainly true that modifying the debugger requires careful analysis.  I suspect the real question is: *How well tested is the debugger?*

@shlomif can you comment on this?

Thank you very much.
Jim Keenan

tonycoz:
> > Given how fragile the debugger is, I'm slightly worried that «use warnings» will get enabled in places where it shouldn't be. I think I recall such things happening when it was made strict.
> > I might be completely mistaken though, this will require some careful analysis.
> 
> It is certainly true that modifying the debugger requires careful analysis. I suspect the real question is: _How well tested is the debugger?_

Not as well as it should be, but that's been improving over the years since @shlomif refactored it, since that re-work introduced a mechanism to simplify *testing* the debugger.

Now, that re-work introduced other bugs, but as with any other part of perl the issues get fixed and tests added.

I don't feel that the solution to "the debugger is fragile" is "don't touch it".

In this case, the debugger typically executes code generated for user commands in the context of the current position in the debugged program*, so I think it's unlikely this change will cause any issues.

iabyn:
On Wed, Jan 20, 2021 at 03:12:04PM -0800, Tony Cook wrote:
> Not as well as it should be, but that's been improving over the years since @shlomif refactored it, since that re-work introduced a mechanism to simplify *testing* the debugger.
> 
> Now, that re-work introduced other bugs, but as with any other part of perl the issues get fixed and tests added.
> 
> I don't feel that the solution to "the debugger is fragile" is "don't touch it".
> 
> In this case, the debugger typically executes code generated for user commands in the context of the current position in the debugged program*, so I think it's unlikely this change will cause any issues.

It actually causes a big problem. The sub DB::eval is now has warnings
enabled; this is used to execute code given by the 'x' command, so such
code now warns.

before:

    $ ./perl -Ilib -d -e1
      DB<1> x "foo" + 1
    0  1
      DB<2>

after:

    $ ./perl -Ilib -d -e1
      DB<1> x "foo" + 1
    Argument "foo" isn't numeric in addition (+) at (eval 10)[lib/perl5db.pl:743] line 2.
     at (eval 10)[lib/perl5db.pl:743] line 2.
            eval 'no strict; ($@, $!, $^E, $,, $/, $\\, $^W) = @DB::saved;package main; $^D = $^D | $DB::db_stop;
      "foo" + 1;
    ' called at lib/perl5db.pl line 743
            DB::eval called at lib/perl5db.pl line 3455
            DB::DB called at -e line 1
    0  1
      DB<2>


-- 
Little fly, thy summer's play my thoughtless hand
has terminated with extreme prejudice.
        (with apologies to William Blake)

jkeenan:
Confirming @iabyn's analysis (with some lines rebroken for clarity):
```
$ ./bin/perl -Ilib -d -e 1

Loading DB routines from perl5db.pl version 1.61
Editor support available.

Enter h or 'h h' for help, or 'man perldebug' for more help.

main::(-e:1):	1
  DB<1> x "foo" + 1
Argument "foo" isn't numeric in addition (+) 
 at (eval 12)[/tmp/lib/perl5/5.33.6/perl5db.pl:743] line 2.
 at (eval 12)[/tmp/lib/perl5/5.33.6/perl5db.pl:743] line 2.
	eval 'no strict; ($@, $!, $^E, $,, $/, $\\, $^W) = @DB::saved;
	package main; $^D = $^D | $DB::db_stop;
  "foo" + 1;
' called at /tmp/lib/perl5/5.33.6/perl5db.pl line 743
	DB::eval called at /tmp/lib/perl5/5.33.6/perl5db.pl line 3455
	DB::DB called at -e line 1
0  1
```
At this point I think it's important to distinguish between two questions:

1. *Can* we turn warnings on in the debugger?  That is, would turning warnings on in the debugger give incorrect results?

2. *Should* we turn warnings on in the debugger?  That is, assuming there was no negative impact on the results, would we want to turn warnings on there? 

As to the first question:  So far no one has presented any evidence that turning warnings on in the debugger would give incorrect results.

As to the second question:  Dave's example clearly shows that turning them on might be very annoying.  On the other hand, there are users who repeatedly ask us to implement new warnings.  So there's no one, incontrovertible answer to that question.  It's a question of individual preferences.

Deciding among preferences is something which, in my opinion, a github issue or a bug ticket is not well suited.  So I'm closing this pull request at this time.  (I'll leave the branch in the repository for reference for a while.)  I urge anyone who wishes to continue this discussion to write to the perl5-porters list/newsgroup.

Thank you very much.
Jim Keenan

PR #18435, opened 2020-12-24
Improve GDBM_File

graygnuorg:
This patch implements new functions and improves compatibility with new versions of GDBM.

* ext/GDBM_File/GDBM_File.xs: Define interface methods for
functions in newer GDBM versions.
(GDBM_version): New static method.  Return the version number
(string in scalar, array of numbers in list context).  Provide
heurisics for determining the library version for GDBM prior
to 1.9.
(gdbm_close): Propagate return value from the library call.
(gdbm_DESTROY): Croak if closing the database fails.
(gdbm_UNTIE): New method.
(gdbm_FETCH): Check database validity. Croak if gdbm_fetch returns
error (except GDBM_ITEM_NOT_FOUND).
(gdbm_STORE,gdbm_DELETE): Likewise.
(gdbm_FIRSTKEY,gdbm_NEXTKEY): Likewise.
(gdbm_EXISTS): Check database validity.
(gdbm_errno): New function.
(gdbm_syserrno): New function.
(gdbm_strerror): New function.
(gdbm_clear_error): New function.
(gdbm_needs_recovery): New function.
(gdbm_recover): New function.
(gdbm_count): New function.
* ext/GDBM_File/typemap (gdbm_count_t): Map to T_COUNT.
* ext/GDBM_File/GDBM_File.pm: Document everything. Raise $VERSION.

* MANIFEST: Update.
* AUTHORS: Update.

### DIFF
--- a/ext/GDBM_File/GDBM_File.xs
+++ b/ext/GDBM_File/GDBM_File.xs
@@ -23,21 +23,104 @@ typedef datum datum_key ;
 typedef datum datum_value ;
 typedef datum datum_key_copy;
 
-#if defined(GDBM_VERSION_MAJOR) && defined(GDBM_VERSION_MINOR) \
-    && GDBM_VERSION_MAJOR > 1 || \
-    (GDBM_VERSION_MAJOR == 1 && GDBM_VERSION_MINOR >= 9)
-typedef void (*FATALFUNC)(const char *);
+/* Indexes for gdbm_flags aliases */
+enum {
+    opt_flags = 0,
+    opt_cache_size,
+    opt_sync_mode,
+    opt_centfree,
+    opt_coalesce,
+    opt_dbname,
+    opt_block_size,
+    opt_mmap,
+    opt_mmapsize
+};
+
+/* Names of gdbm_flags aliases, for error reporting.
+   Indexed by opt_ constants above.
+*/
+char const *opt_names[] = {
+    "GDBM_File::flags",
+    "GDBM_File::cache_size",
+    "GDBM_File::sync_mode",
+    "GDBM_File::centfree",
+    "GDBM_File::coalesce",
+    "GDBM_File::dbname",
+    "GDBM_File::block_size",
+    "GDBM_File::mmap",
+    "GDBM_File::mmapsize"
+};    
+
+#ifdef GDBM_VERSION_MAJOR
+# define GDBM_VERSION_GUESS 0
 #else
-typedef void (*FATALFUNC)();
+/* Try educated guess
+ * The value of GDBM_VERSION_GUESS indicates how rough the guess is:
+ *   1 - Precise; based on the CVS logs and existing archives
+ *   2 - Moderate. The major and minor number are correct. The patchlevel
+ *       is set to the upper bound.
+ *   3 - Rough; The version is guaranteed to be not newer than major.minor.
+ */
+# if defined(GDBM_SYNCMODE)
+/* CHANGES from 1.7.3 to 1.8
+ *   1.  Added GDBM_CENTFREE functionality and option.
+ */  
+#  define GDBM_VERSION_MAJOR 1
+#  define GDBM_VERSION_MINOR 8
+#  define GDBM_VERSION_PATCH 3
+#  define GDBM_VERSION_GUESS 1
+# elif defined(GDBM_FASTMODE)
+/* CHANGES from 1.7.2 to 1.7.3
+ *  1.  Fixed a couple of last minute problems. (Namely, no autoconf.h in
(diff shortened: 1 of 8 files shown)


ext/GDBM_File/t/count.t
```diff
@@ -0,0 +1,34 @@
...
+        if $^O =~ /darwin/;
+
+    plan(tests => 3);
+    use_ok('GDBM_File');
+ }
+
+unlink <Op_dbmx*>;
+
+my %h;
+my $db = tie(%h, 'GDBM_File', 'Op_dbmx', GDBM_WRCREAT, 0640);
```
Leont: This test should probably use a tempdir from `File::Temp` (instead of manually unlinking files)

ext/GDBM_File/typemap
```diff
@@ -11,6 +11,7 @@ SDBM_File		T_PTROBJ
 ODBM_File		T_PTROBJ
 DB_File			T_PTROBJ
 DBZ_File		T_PTROBJ
+gdbm_count_t            T_COUNT
```
Leont: Is there any specific reason for not using `T_UV` here?

ext/GDBM_File/t/count.t
graygnuorg: Yes, that's reasonable.  I did it that way because all existing tests do so.

ext/GDBM_File/typemap
graygnuorg: The initial reason was that gdbm_count_t is not a fixed typedef, but is determined at configure time. But I'm going to change it to a fixed type in future versions of gdbm anyway, so yes, it's better to use T_UV  here.

ext/GDBM_File/GDBM_File.xs
```diff
@@ -62,17 +145,114 @@ output_datum(pTHX_ SV *arg, char *str, int size)
...
+        db->dbp = NULL;
+    }
+    return rc;
+}
+
+#if GDBM_VERSION_MAJOR == 1 && GDBM_VERSION_MINOR >= 13
+/* Error-reporting wrapper for gdbm_recover */
+static void
+rcvr_errfun(void *cv, char const *fmt, ...)
+{
```
tonycoz, line 194: This needs a dTHX to build with threaded builds of perl:

ext/GDBM_File/GDBM_File.xs
```diff
@@ -62,17 +145,114 @@ output_datum(pTHX_ SV *arg, char *str, int size)
...
 
 #include "const-c.inc"
 
 MODULE = GDBM_File	PACKAGE = GDBM_File	PREFIX = gdbm_
 
 INCLUDE: const-xs.inc
 
+void
+gdbm_GDBM_version(package)
+        char *package;
```
tonycoz: You don't use `package`, so there's no need to give it a type (no need for the `char *package;` line), which prevents the XS generated initialization, preventing a warning:

```
GDBM_File.c: In function ‘XS_GDBM_File_GDBM_version’:
GDBM_File.c:430:9: warning: unused variable ‘package’ [-Wunused-variable]
  char * package = (char *)SvPV_nolen(ST(0))
```

ext/GDBM_File/GDBM_File.xs
```diff
@@ -105,31 +284,46 @@ gdbm_TIEHASH(dbtype, name, read_write, mode)
...
+        unsigned count
+    CODE:
+        if (count == 0) {
+            if (gdbm_file_close(db))
+                croak("gdbm_close: %s; %s",
+                      gdbm_strerror(gdbm_errno),
+                      strerror(errno));
+	}
+
+
```
tonycoz, line 314: I don't think this is desirable.  If someone unties the hash, gdbm_file_close() is called, but it's also called by DESTROY.

graygnuorg: It is desirable, because it allows to catch eventual errors early. As to your second point, gdbm_file_close can be called multiple times without adverse effects: after the first call it becomes a no-op.

ext/GDBM_File/GDBM_File.xs
graygnuorg: Thanks.

ext/GDBM_File/GDBM_File.xs
tonycoz: ok, thanks

ext/GDBM_File/GDBM_File.xs
graygnuorg: Fixed in new commit. Thank you.

Issue #18389, opened 2020-12-07
[doc] Request better documentation of "my"


jidanni:
```
$ perldoc -f my
    my VARLIST
    my TYPE VARLIST
    my VARLIST : ATTRS
    my TYPE VARLIST : ATTRS
            A "my" declares the listed variables to be local (lexically) to
            the enclosing block, file, or "eval"...
```
OK but be sure to mention when its OK, and when not:

```
$ cat e.pl
use strict;
use warnings 'all';
my $e = 1;
my $e = 1;
{
    $e = 1;
    my $e = 1;
}
$ perl e.pl
"my" variable $e masks earlier declaration in same scope at e.pl line 4.
```
So only one of the three my's is bad.
So *perldoc -f my* needs to tell the *user* why. (Don't tell *me* why.)
Maybe the details are in perlsub. But we aren't using subroutines here,
so the user won't look there on his own.

richardleach:
Some text such as the following? (Which hopefully matches up with the error description in [perldiag.pod](https://github.com/Perl/perl5/blob/blead/pod/perldiag.pod).)
`Variables should not subsequently be redeclared in the same scope or statement.`

khwilliamson:
Please submit a PR

richardleach:
PR coming shortly.

richardleach:
We added additional text in https://github.com/Perl/perl5/commit/920000260c5f3f16e119b8e4840a3c9f89ba3988. Closing this issue.

PR #18437, opened 2020-12-25
perlfunc: explicitly warn against redeclaring lexical variables with my()

richardleach:
Addresses a concern raised in #18389. The new text is based upon the already-existing error description in perldiag.pod.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -4350,6 +4350,17 @@ A L<C<my>|/my VARLIST> declares the listed variables to be local
 more than one variable is listed, the list must be placed in
 parentheses.
 
+Note that with a parenthesised list, L<C<undef>|/undef EXPR> can be used
+as a dummy placeholder, for example to skip assignment of initial
+values:
+
+    my ( undef, $min, $hour ) = localtime;
+
+Redeclaring a variable in the same scope or statement will "shadow" the
+previous declaration, creating a new instance and preventing access to
+the previous one. This is usually undesired and, if warnings are enabled,
+will result in a warning in the C<shadow> category.
+
 The exact semantics and interface of TYPE and ATTRS are still
 evolving.  TYPE may be a bareword, a constant declared
 with L<C<use constant>|constant>, or L<C<__PACKAGE__>|/__PACKAGE__>.  It
@@ -4358,12 +4369,6 @@ currently bound to the use of the L<fields> pragma,
 and attributes are handled using the L<attributes> pragma, or starting
 from Perl 5.8.0 also via the L<Attribute::Handlers> module.  See
 L<perlsub/"Private Variables via my()"> for details.
-
-Note that with a parenthesised list, L<C<undef>|/undef EXPR> can be used
-as a dummy placeholder, for example to skip assignment of initial
-values:
-
-    my ( undef, $min, $hour ) = localtime;
 
 =item next LABEL
 X<next> X<continue>
@@ -8577,6 +8582,11 @@ parentheses.  With a parenthesised list, L<C<undef>|/undef EXPR> can be
 used as a
 dummy placeholder.  However, since initialization of state variables in
 such lists is currently not possible this would serve no purpose.
+
+Redeclaring a variable in the same scope or statement will "shadow" the
+previous declaration, creating a new instance and preventing access to
+the previous one. This is usually undesired and, if warnings are enabled,
+will result in a warning in the C<shadow> category.
 
 L<C<state>|/state VARLIST> is available only if the
 L<C<"state"> feature|feature/The 'state' feature> is enabled or if it is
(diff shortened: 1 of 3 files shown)


pod/perlfunc.pod
```diff
@@ -4348,7 +4348,8 @@ X<my>
 A L<C<my>|/my VARLIST> declares the listed variables to be local
 (lexically) to the enclosing block, file, or L<C<eval>|/eval EXPR>.  If
 more than one variable is listed, the list must be placed in
-parentheses.
+parentheses. Variables should not subsequently be redeclared in the
```
jkeenan: The word *subsequently* there seems superfluous.  I think removing it would remove a whiff of ambiguity from that sentence.

Grinnz:
I don't think this is very informative. If it's going to say people should not do something, it should explain why not, what happens when you do. It also doesn't bear much relation to the rest of the first paragraph which deals with how the syntax works.

My suggestion would be a new paragraph such as:

```
Redeclaring a variable in the same scope or statement will "shadow"
the previous declaration, creating a new instance and preventing
access to the previous one. This is usually undesired and will result
in a warning in the C<shadow> category, if warnings are enabled.
```

Ref: https://perldoc.perl.org/perldiag#%22%25s%22-variable-%25s-masks-earlier-declaration-in-same-%25s

Grinnz:
I also recommend adding the same information to the docs for `state`.

richardleach:
@jkeenan / @Grinnz - thanks for the feedback (and suggested better wording). I'll update this PR later in the holidays.

jkeenan:
> @jkeenan / @Grinnz - thanks for the feedback (and suggested better wording). I'll update this PR later in the holidays.

@richardleach, Are you going to be able to provide this updated PR?  (I'm trying to reduce number of open tickets prior to presumed code freeze.)

Thank you very much.
Jim Keenan

richardleach:
I need to tweak the perldelta text. I'll push again shortly, AFK for now.

pod/perlfunc.pod
richardleach: I subsequently went with Grinnz' wording below, with a minor tweak.

richardleach:
This PR is ready to check over.

pod/perlfunc.pod
```diff
@@ -4359,12 +4370,6 @@ and attributes are handled using the L<attributes> pragma, or starting
 from Perl 5.8.0 also via the L<Attribute::Handlers> module.  See
 L<perlsub/"Private Variables via my()"> for details.
 
-Note that with a parenthesised list, L<C<undef>|/undef EXPR> can be used
```
Grinnz, line 4362: was this paragraph intentionally moved?

Grinnz: (I think the new location is better, but just checking as it's not really relevant)

pod/perlfunc.pod
```diff
@@ -4350,6 +4350,17 @@ A L<C<my>|/my VARLIST> declares the listed variables to be local
...
+Note that with a parenthesised list, L<C<undef>|/undef EXPR> can be used
+as a dummy placeholder, for example to skip assignment of initial
+values:
+
+    my ( undef, $min, $hour ) = localtime;
+
+Redeclaring a variable in the same scope or statement will "shadow" the
+previous declaration, creating a new instance and preventing access to
+the previous one. This is usually undesired and, if warnings are enabled,
+will result in a warning in the C<shadow> category.
```
jkeenan, line 4362: Do we have anything in the test suite that demonstrates a C<shadow> warning being thrown in this situation?

If not, should we?

(Same question applies to line 8589 same file.)

pod/perlfunc.pod
richardleach: Intentional on the basis that (1) the paragraphs for both functions then appear in a more similar-seeming order than before (2) the "gotchas"/see-links-for-more-info paragraphs come last, which I think aids comprehension - especially for newcomers.

pod/perlfunc.pod
richardleach: There are warning for `my` (and `our`) in _t/lib/warnings/pad_ and a warning for a `state` subroutine in _t/op/lexsub.t_

I'm not sure if there are any tests for `state` variable warnings though. (The warnings are of the form `"state" variable $pizza masks earlier declaration in same scope at temp.pl line 4.` and `grep -r '"state" variable' *` produced no matches.)

I'm happy to add a test or two if they genuinely are missing. Should it/they also go in _t/lib/warnings/pad_?

richardleach: I've pushed a commit that adds `state` tests into _t/lib/warnings/pad_. The locations of each new/changed test are pretty arbitrary, so I'm happy to make changes if anyone spots problems or possible improvements.

jkeenan:
Inspected and merged locally as commit 920000260c; pushed to origin; closing ticket.

Thank you very much.
Jim Keenan

PR #18439, opened 2020-12-25
Fix up subsystem termination order

khwilliamson:
During destruction, a list is executed in order.  Here's a couple of
them:
OP_CHECK_MUTEX_TERM; OP_REFCNT_TERM;

If something goes wrong, and the I/O system has already been destructed,
it's hard to figure out what's going on.  This is solved simply by
moving the I/O termination to the end of the list.

### DIFF
--- a/unixish.h
+++ b/unixish.h
@@ -136,29 +136,31 @@ int afstat(int fd, struct stat *statb);
 #define Mkdir(path,mode)   mkdir((path),(mode))
 
 #if defined(__amigaos4__)
-#  define PERL_SYS_INIT_BODY(c,v)					\
-        MALLOC_CHECK_TAINT2(*c,*v) PERL_FPU_INIT; PERLIO_INIT; MALLOC_INIT; amigaos4_init_fork_array(); amigaos4_init_environ_sema();
-#  define PERL_SYS_TERM_BODY()                         \
+#  define AMIGAOS_SYS_TERM_  amigaos4_dispose_fork_array()
+#  define AMIGAOS_SYS_INIT_ STMT_START {                        \
+                                amigaos4_init_fork_array();     \
+                                amigaos4_init_environ_sema();   \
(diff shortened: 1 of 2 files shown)


Leont:
I would expect `PERLIO_TERM` to have to happen before `MALLOC_TERM`, otherwise mymalloc can get confused? (as `PerlIO_teardown()` uses `free()`

jkeenan:
@khwilliamson , @Leont, Can we get an update on the status of this p.r.?

Thank you very much.
Jim Keenan

tonycoz:
Yes, 

> 
> 
> I would expect `PERLIO_TERM` to have to happen before `MALLOC_TERM`, otherwise mymalloc can get confused? (as `PerlIO_teardown()` uses `free()`

mymalloc uses PerlIO under -Dm (which is probably a bug.)

unixish.h
```diff
@@ -136,29 +136,31 @@ int afstat(int fd, struct stat *statb);
...
 
-#ifndef PERL_SYS_INIT_BODY
-#  define PERL_SYS_INIT_BODY(c,v)					\
-        MALLOC_CHECK_TAINT2(*c,*v) PERL_FPU_INIT; PERLIO_INIT; MALLOC_INIT
-#endif
+#define PERL_SYS_INIT_BODY(c,v)					\
+	MALLOC_CHECK_TAINT2(*c,*v) PERL_FPU_INIT; PERLIO_INIT;  \
+        MALLOC_INIT; AMIGAOS_SYS_INIT_;
 
-#ifndef PERL_SYS_TERM_BODY
```
Leont, line 154: I think we still need this `#ifdef`. There are several operating systems that still define their own `PERL_SYS_INIT_BODY` (plan9, dos, vms, os2).

Maybe this can be refactored into some `OS_SYS_INIT_`, I haven't looked into potential ordering issues.

PR #18441, opened 2020-12-27
Use Perl_signbit unconditionally for POSIX::signbit()

t-a-k:
Previously `POSIX::signbit()` had a fallback implementation for the case `Perl_signbit` is not defined, but this code is apparently broken because it returns wrongly non-zero (true) value for `signbit(+0.0)`.
Currently `Perl_signbit` is always defined (either as a macro or as a function), so it seems to be better to use it unconditionally.

Additionally, unused macro `c99_signbit` is deleted, which is also apparently broken because there is no `signbitl()` in C99.

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -280,7 +280,7 @@ static int not_here(const char *s);
 #  define c99_rint	rintq
 #  define c99_round	roundq
 #  define c99_scalbn	scalbnq
-#  define c99_signbit	signbitq
+/* We already define Perl_signbit to signbitq in perl.h. */
 #  define c99_tgamma	tgammaq
 #  define c99_trunc	truncq
 #  define bessel_j0 j0q
@@ -331,9 +331,7 @@ static int not_here(const char *s);
 #  define c99_rint	rintl
 #  define c99_round	roundl
 #  define c99_scalbn	scalbnl
-#  ifdef HAS_SIGNBIT /* possibly bad assumption */
-#    define c99_signbit	signbitl
-#  endif
+/* We already define Perl_signbit in perl.h. */
 #  define c99_tgamma	tgammal
 #  define c99_trunc	truncl
 #else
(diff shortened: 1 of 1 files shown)


jkeenan:
> Previously `POSIX::signbit()` had a fallback implementation for the case `Perl_signbit` is not defined, but this code is apparently broken because it returns wrongly non-zero (true) value for `signbit(+0.0)`.
> Currently `Perl_signbit` is always defined (either as a macro or as a function), so it seems to be better to use it unconditionally.
> 
> Additionally, unused macro `c99_signbit` is deleted, which is also apparently broken because there is no `signbitl()` in C99.

Can you specify more precisely what you mean by *previously* and *currently*?

jkeenan:
> Previously `POSIX::signbit()` had a fallback implementation for the case `Perl_signbit` is not defined, but this code is apparently broken because it returns wrongly non-zero (true) value for `signbit(+0.0)`.
> Currently `Perl_signbit` is always defined (either as a macro or as a function), so it seems to be better to use it unconditionally.
> 
> Additionally, unused macro `c99_signbit` is deleted, which is also apparently broken because there is no `signbitl()` in C99.

@khwilliamson, @doughera88 You have worked on what appears to be a relevant section of `numeric.c`.  Could you take a look at this p.r.?

Thank you very much.
Jim Keenan

t-a-k:
> > Previously `POSIX::signbit()` had a fallback implementation for the case `Perl_signbit` is not defined, but this code is apparently broken because it returns wrongly non-zero (true) value for `signbit(+0.0)`.
> > Currently `Perl_signbit` is always defined (either as a macro or as a function), so it seems to be better to use it unconditionally.
> > Additionally, unused macro `c99_signbit` is deleted, which is also apparently broken because there is no `signbitl()` in C99.
> 
> Can you specify more precisely what you mean by _previously_ and _currently_?

Sorry for my lack of explanation. These are based on this commit (c934219, which is based on blead's dba7cf2e); I mean "before c934219" by _previously_ and "as of c934219" by _currently_.

khwilliamson:
I looked at the code, and this commit looks to be the right thing to do

PR #18443, opened 2020-12-27
perlintern: Document gv_stashsvpvn_cached

khwilliamson:


gv.c
```diff
@@ -1542,19 +1542,24 @@ S_gv_stashpvn_internal(pTHX_ const char *name, U32 namelen, I32 flags)
...
 Returns a pointer to the stash for a specified package, possibly
 cached.  Implements both C<L</gv_stashpvn>> and C<L</gv_stashsv>>.
 
 Requires one of either C<namesv> or C<namepv> to be non-null.
 
-See C<L</gv_stashpvn>> for details on C<flags>.
+If the flag C<GV_CACHE_ONLY> is set, return the stash only if found in the
+cache; see C<L</gv_stashpvn>> for details on the other C<flags>.
 
 Note the sv interface is strongly preferred for performance reasons.
```
tomhukins, line 1555: I realise it's not part of your change, but it might be helpful to replace "the sv interface" with "L</gv_stashsv>" here (unless I've misunderstood what this means).

tomhukins (approved):

tonycoz:
Looks ok except for the relevant test failures.

gv.c
jkeenan: @khwilliamson , could you indicate what you think about @tomhukins's suggestion?

Otherwise, once you get `make test_porting` to pass, this p.r. looks applicable.

Thank you very much.
Jim Keenan

jkeenan:
> Looks ok except for the relevant test failures.

I tried to fix these failures but was unsuccessful. @khwilliamson, can you take a look?

Thank you very much.
Jim Keenan

gv.c
khwilliamson: I finally got a chance to look into this, and looked at the code.

This function takes either an sv or a pv.  If  the sv is furnished, there is a lot less work.  That has to be what was meant here.  I pushed commit 85bf2e910b46818f479636be677d795ac2bc35a1 that includes the original, plus the clarification brought up by @tomhukins

PR #18444, opened 2020-12-28
Run `makedepend` in parallel by using `make`

Corion:
This moves the per-file loop body of `makedepend` into a separate
file named `makedepend_file` and then uses `make` to launch
the `makedepend_file` processes for each target potentially in parallel.

This reduces the time for

  time sh ./makedepend MAKE=make cflags

from 5 seconds to 2 seconds with MAKEFLAGS=-j8

### DIFF
--- a/Makefile.SH
+++ b/Makefile.SH
@@ -1377,6 +1377,7 @@ veryclean:	_verycleaner _mopup _clobber
 _mopup:
 	rm -f *$(OBJ_EXT) *$(LIB_EXT) all perlmain.c opmini.c perlmini.c generate_uudmap$(EXE_EXT) $(generated_headers)
 	-rmdir .depending
+	-rm *.depends
 	-@test -f extra.pods && rm -f `cat extra.pods`
 	-@test -f vms/README_vms.pod && rm -f vms/README_vms.pod
 	-rm -f perl.exp ext.libs $(generated_pods) uni.data opmini.o perlmini.o pod/roffitall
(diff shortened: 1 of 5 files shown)


Corion:
This is a work in progress, but ready for review.

It should not go into the 2021 Perl release, as that is too short a time window to find out which flavours of `make` do not like the syntax I've used.

makedepend.SH
```diff
@@ -235,6 +138,8 @@ if $test -s .deptmp; then
        -e 'h; s/mini\(perlmain\)/\1/p; g' \
        .deptmp >>$mf.new
 else
+    $echo "Should not get here"
+    exit 1
```
tonycoz, line 142: (Interestingly the code block this protects largely goes back to perl 1.0.)

This change however probably shouldn't be in the final patch

Makefile.SH
```diff
@@ -1521,6 +1522,9 @@ $spitshell >>$Makefile <<'!NO!SUBS!'
 depend: makedepend $(DTRACE_H) $(generated_headers)
 	sh ./makedepend MAKE="$(MAKE)" cflags
 
+%.c.depends: %.c
```
tonycoz, line 1525: You should probably be adding `*.depends` to `.gitignore`.

jkeenan:
> This moves the per-file loop body of `makedepend` into a separate
> file named `makedepend_file` and then uses `make` to launch
> the `makedepend_file` processes for each target potentially in parallel.
> 
> This reduces the time for
> 
> time sh ./makedepend MAKE=make cflags
> 
> from 5 seconds to 2 seconds with MAKEFLAGS=-j8

@Corion, would you be in a position to resolve merge conflicts so that we can proceed to test out this p.r.?

Thank you very much.
Jim Keenan

jkeenan:
> > This moves the per-file loop body of `makedepend` into a separate
> > file named `makedepend_file` and then uses `make` to launch
> > the `makedepend_file` processes for each target potentially in parallel.
> > This reduces the time for
> > time sh ./makedepend MAKE=make cflags
> > from 5 seconds to 2 seconds with MAKEFLAGS=-j8
> 
> @Corion, would you be in a position to resolve merge conflicts so that we can proceed to test out this p.r.?
> 

**CORRECTION:** There are no merge conflicts at the present time.  Do you want to move this p.r. from "draft" to active status?

jimk


> Thank you very much.
> Jim Keenan

Corion:
Yes - IMO this should be merged early so we see if/where I broke `Configure` or `make` .

PR #18445, opened 2020-12-28
Remove inheritance from Exporter in ext/ modules

Corion:
Inheriting from Exporter adds several subroutines that are not really
needed by these modules.

The remaining uses of inheritance from Exporter are:

re.pm - this uses export_to_level(), which really, really wants
the inheritance

### DIFF
--- a/ext/Devel-Peek/Peek.pm
+++ b/ext/Devel-Peek/Peek.pm
@@ -3,14 +3,13 @@
 
 package Devel::Peek;
 
-$VERSION = '1.30';
+$VERSION = '1.31';
 $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
 require Exporter;
 require XSLoader;
 
-@ISA = qw(Exporter);
 @EXPORT = qw(Dump mstat DeadCode DumpArray DumpWithOP DumpProg
 	     fill_mstats mstats_fillhash mstats2hash runops_debug debug_flags);
 @EXPORT_OK = qw(SvREFCNT CvGV);
--- a/ext/Amiga-ARexx/ARexx.pm
+++ b/ext/Amiga-ARexx/ARexx.pm
@@ -5,10 +5,7 @@ use strict;
 use warnings;
 use Carp;
 
-require Exporter;
-#use AutoLoader;
-
-our @ISA = qw(Exporter);
+use Exporter 'import';
 
 # Items to export into callers namespace by default. Note: do not export
 # names by default without a very good reason. Use EXPORT_OK instead.
@@ -26,7 +23,7 @@ our @EXPORT_OK = ( @{ $EXPORT_TAGS{'all'} } );
 our @EXPORT = qw(
 );
 
-our $VERSION = '0.05';
+our $VERSION = '0.06';
 
 require XSLoader;
 XSLoader::load('Amiga::ARexx', $VERSION);
--- a/ext/Amiga-Exec/Exec.pm
+++ b/ext/Amiga-Exec/Exec.pm
@@ -5,10 +5,7 @@ use strict;
 use warnings;
 use Carp;
 
-require Exporter;
-#use AutoLoader;
-
-our @ISA = qw(Exporter);
+use Exporter 'import';
 
 # Items to export into callers namespace by default. Note: do not export
 # names by default without a very good reason. Use EXPORT_OK instead.
@@ -26,7 +23,7 @@ our @EXPORT_OK = ( @{ $EXPORT_TAGS{'all'} } );
 our @EXPORT = qw(
 );
 
-our $VERSION = '0.02';
+our $VERSION = '0.03';
 
 require XSLoader;
 XSLoader::load('Amiga::Exec', $VERSION);
--- a/ext/B/B/Concise.pm
+++ b/ext/B/B/Concise.pm
@@ -12,10 +12,9 @@ package B::Concise;
 use strict; # use #2
 use warnings; # uses #3 and #4, since warnings uses Carp
 
-use Exporter (); # use #5
+use Exporter 'import'; # use #5
 
-our $VERSION   = "1.004";
-our @ISA       = qw(Exporter);
+our $VERSION   = "1.005";
 our @EXPORT_OK = qw( set_style set_style_standard add_callback
 		     concise_subref concise_cv concise_main
 		     add_style walk_output compile reset_sequence );
--- a/ext/Errno/Errno_pm.PL
+++ b/ext/Errno/Errno_pm.PL
@@ -2,7 +2,7 @@ use ExtUtils::MakeMaker;
 use Config;
 use strict;
 
-our $VERSION = "1.33";
+our $VERSION = "1.34";
 
 my %err = ();
 
@@ -284,7 +284,7 @@ sub write_errno_pm {
 #
 
 package Errno;
-require Exporter;
+use Exporter 'import';
 use strict;
 
 EDQ
@@ -305,7 +305,6 @@ CONFIG_CHECK_END
     print <<"EDQ";
 our \$VERSION = "$VERSION";
 \$VERSION = eval \$VERSION;
-our \@ISA = 'Exporter';
 
 my %err;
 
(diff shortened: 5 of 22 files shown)


ext/Devel-Peek/Peek.pm
```diff
@@ -3,14 +3,13 @@
 
 package Devel::Peek;
 
-$VERSION = '1.30';
+$VERSION = '1.31';
 $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
 require Exporter;
```
Leont, line 10: This doesn't import `&import`?

Grinnz:
Nothing jumps out to me as problematic here, but it might be worth running a CPAN smoke.

ext/Devel-Peek/Peek.pm
Corion: `Devel::Peek` jumps directly to `&Exporter::import`, and doesn't need the inheritance.

Grinnz:
If you want to be defensive for the case where any of these are to get dual-lifed, you could write `use Exporter 5.57 'import';` to make clear the version of Exporter required.

haarg:
This breaks compatibility with anything using `->export_to_level`.

Corion:
@haarg I don't understand this comment. Is there a (documented) use case to have (say)

    require Devel::Peek;
    Devel::Peek->export_to_level(...);

If so, then yes, this removes (undocumented) functionality. I'd like to understand the usage better - none of these modules are using `->export_to_level` themselves.

haarg:
If you want your module to export from multiple modules to your caller, one way if the modules inherit from Exporter is to use export_to_level:
```perl
package Foo::Bar;
use Devel::Peek ();
use Fcntl ();
sub import {
    Devel::Peek->export_to_level(1);
    Fcntl->export_to_level(1);
}

package Bar;
use Foo::Bar; # get all default exports from Devel::Peek and Fcntl
```
This is limited to just exporting, not the other things done in import methods, so generally tools like Import::Into are better. But in older code, export_to_level is still sometimes expected to be available.

Corion:
At least one module on CPAN is affected by this change:

https://metacpan.org/release/Sun-Solaris-Exacct/source/Exacct.pm#L69

So the way forward would be a deprecation cycle and documentation on how to fix the affected code.

khwilliamson:
@Corion.  It appears you are saying we shouldn't merge this before a deprecation cycle?  Or maybe we do a cpan smoke to find out how much breakage happens.  If it is just a few modules, maybe do PRs on them and merge anyway?

Corion:
Yes, certainly a CPAN smoke to see how much breakage happens is a first step. I would like to see this go forward, but I'm also happy with a full deprecation cycle. Let's first assess the breakage and then decide whether to merge this in 5.36 or afterwards.

jkeenan:
> Yes, certainly a CPAN smoke to see how much breakage happens is a first step. I would like to see this go forward, but I'm also happy with a full deprecation cycle. Let's first assess the breakage and then decide whether to merge this in 5.36 or afterwards.

1. @Corion, this p.r. has merge conflicts which should be resolved before we can proceed further.
2. Bigger problem:  AFAIK we have never had the infrastructure to do a thorough "smoke-against-CPAN" for a branch in the core distribution.  Contributors to cpantesters.org test CPAN releases against (in descending order of frequency) production releases, monthly dev releases and HEAD of blead.  It would be great if we did not have those limitations, but that's where we are.  How do you think we should proceed?

Thank you very much.
Jim Keenan

leonerd:
@jkeenan 
> Bigger problem: AFAIK we have never had the infrastructure to do a thorough "smoke-against-CPAN" for a branch in the core distribution. Contributors to cpantesters.org test CPAN releases against (in descending order of frequency) production releases, monthly dev releases and HEAD of blead. It would be great if we did not have those limitations, but that's where we are. How do you think we should proceed?

Isn't this what the whole BBC ("blead breaks CPAN") infrastructure is about? I've had reports from that before, about my own CPAN modules which blead breaks. For example, https://github.com/Perl/perl5/issues/18376 to which you even replied.

Corion:
@leonerd Yes, but as far as I know, BBC only checks blead/HEAD , not other branches. A good compromise would be IMO to merge this change with/after 5.35.0 and then see what the fallout is from BBC, and potentially revert it later.

jkeenan:
> @jkeenan
> 
> > Bigger problem: AFAIK we have never had the infrastructure to do a thorough "smoke-against-CPAN" for a branch in the core distribution. Contributors to cpantesters.org test CPAN releases against (in descending order of frequency) production releases, monthly dev releases and HEAD of blead. It would be great if we did not have those limitations, but that's where we are. How do you think we should proceed?
> 
> Isn't this what the whole BBC ("blead breaks CPAN") infrastructure is about? I've had reports from that before, about my own CPAN modules which blead breaks. For example, #18376 to which you even replied.

As @Corion said, what infrastructure we have tests blead, not non-blead branches.  And what we have for blead -- as distinct from scheduled dev, prod or maint releases -- we have only because highly skilled and experienced CPANtesters like Andreas, Slaven and Carlos have spent years building up their personally donated operations.

jkeenan:
> @leonerd Yes, but as far as I know, BBC only checks blead/HEAD , not other branches. A good compromise would be IMO to merge this change with/after 5.35.0 and then see what the fallout is from BBC, and potentially revert it later.

@Corion, would you be in a position to resolve the merge conflicts in this branch vis-a-vis blead so that we can proceed to test it out?

Thank you very much.
Jim Keenan

Corion:
@jkeenan Yes - the breakage looks minor and I'll rebase the PR

jkeenan:
> @jkeenan Yes - the breakage looks minor and I'll rebase the PR

Okay, so what I'll do first is create a smoke-me branch from this p.r.  That won't tell us anything about CPAN breakage, but it should flush out any platform/configuration breakage.

Then, after about a week, unless there's well argued objection, we'll merge to blead, after which CPANtesting will begin to produce results.

Thank you very much.
Jim Keenan

jkeenan:
> > @jkeenan Yes - the breakage looks minor and I'll rebase the PR
> 
> Okay, so what I'll do first is create a smoke-me branch from this p.r. That won't tell us anything about CPAN breakage, but it should flush out any platform/configuration breakage.
> 

Here's the branch:
smoke-me/jkeenan/corion/remove-exporter-inheritance

jkeenan:
> > > @jkeenan Yes - the breakage looks minor and I'll rebase the PR
> > 
> > 
> > Okay, so what I'll do first is create a smoke-me branch from this p.r. That won't tell us anything about CPAN breakage, but it should flush out any platform/configuration breakage.
> 
> Here's the branch:
> smoke-me/jkeenan/corion/remove-exporter-inheritance

The smoke-test results are looking good.  (There are some failures in tests not related to this p.r. that are mostly network connectivity issues.)

PSC: @rjbs, @nwc10, @neilb Do you have any objection to merging this (in five days)?

Thank you very much.
Jim Keenan

rjbs:
@jkeenan Nicholas, Neil, and I have no objection to this.

Leont:
> after which CPANtesting will begin to produce results

Extracting that does involve labor, is anyone actively looking out for such breakage?

nwc10:
> https://metacpan.org/release/Sun-Solaris-Exacct/source/Exacct.pm#L69



> At least one module on CPAN is affected by this change:
> 
> https://metacpan.org/release/Sun-Solaris-Exacct/source/Exacct.pm#L69
> 
> So the way forward would be a deprecation cycle and documentation on how to fix the affected code.



> At least one module on CPAN is affected by this change:
> 
> https://metacpan.org/release/Sun-Solaris-Exacct/source/Exacct.pm#L69
> 
> So the way forward would be a deprecation cycle and documentation on how to fix the affected code.

Thanks for digging to find this. In turn, I'm not sure if I made the right call here on a "sample size of 1". The module in question seems to have been broken for 15 years. it's Solaris specific, and when you build with Sun's own `dmake`, it fails due to errors in the `XS`. If you try to build with `gmake`, it fails due to to the `Makefile.PL` post-processing the generated `Makefile` to use `dmake` specific syntax. It's all rather strange. No-one has addressed any of this for more than a decade, so I'm guessing that this module "special biologist word".

I'm hoping that it's unique. If I guessed wrong, I guess I get to help fix the problem.

PR #18446, opened 2020-12-29
Detabify C code indentation.

schwern:
This PR replaces the obsolete mixed tab/space indentation across all the core C code (exceptions below) to just spaces to match the style in perlhack. It's long overdue. Some editors ([VSCode](https://github.com/microsoft/vscode/issues/10339) and [Atom](https://github.com/atom/atom/issues/21371)) conflate tab width with indent width and cannot support 4 character spaces with 8 character tabs; this makes the C code even more difficult to work with.

Tabs were expanded with `expand -i`. Indentation style changes and dealing with other tabs are left to another PR.

warnings.pm and feature.pm's versions were bumped to satisfy cmpVERSION.pl. regen scripts were run as needed to update checksums. There are no other non-whitespace changes.

Code archeologists can ignore the "history sheer" from touching so many files by ignoring whitespace: `git blame -w`. [Github also has ways to ignore whitespace changes](https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing).

Applying a consistent indentation style and dealing with other tabs is left to another PR.

# Exceptions
* cpan/
* vutil.* are part of the version distribution.
* XS files because Porting/cmpVERSION.pl thinks they should have their version bumped. I tried getting it to ignore whitespace, but it relies on `git diff --name-only` and that is ignoring the `-w` flag to ignore whitespace.
* perly.* were left alone because that would require a full Bison run which is outside the scope of this PR.


xenu:
This is something that should've been done a long time ago 👍

regen/opcode.pl
```diff
@@ -78,84 +78,84 @@ BEGIN
...
+                 Perl_pp_untie => ['dbmclose'],
+                 Perl_pp_sysread => {read => '', recv => '#ifdef HAS_SOCKET'},
+                 Perl_pp_sysseek => ['seek'],
+                 Perl_pp_ioctl => ['fcntl'],
+                 Perl_pp_ssockopt => {gsockopt => '#ifdef HAS_SOCKET'},
+                 Perl_pp_getpeername => {getsockname => '#ifdef HAS_SOCKET'},
+                 Perl_pp_stat => ['lstat'],
+                 Perl_pp_ftrowned => [qw(fteowned ftzero ftsock ftchr ftblk
+                                         ftfile ftdir ftpipe ftsuid ftsgid
+                                         ftsvtx)],
```
Grinnz, line 97: lol

schwern: There's no code change, just whitespace. Be sure to tick "Hide whitespace changes".

Grinnz: Sorry, laughing that the previous indentation of this line was a (1) space character followed by tabs.

Grinnz:
GitHub's viewer seems to be confused about whether the NetWare files have non-whitespace changes. I've verified locally that they don't.

RIP tab indentation, but it was never used sensibly and this is a significant improvement.

schwern:
This branch is going to gather conflicts *very* quickly. They can be resolved easily, take the blead version of the file and rerun `expand -i`, but I'd rather not have to repeatedly chase blead.

How do we proceed?

Leont:
> How do we proceed?

Probably this deserves some discussion on the mailing list.

Leont:
> GitHub's viewer seems to be confused about whether the NetWare files have non-whitespace changes. I've verified locally that they don't.

I really wonder what's going on there. At first I assumed it would be a matter of line-endings, but that doesn't appear to be the case.

jkeenan:
On 12/29/20 8:55 PM, Leon Timmermans wrote:
>     How do we proceed?
> 
> Probably this deserves some discussion on the mailing list.
> 

I agree; this should be discussed on list.  When, over the years, I have 
suggested cleaning up leading whitespace in source code files, I have 
gotten considerable push back from major contributors, some of whom are 
still somewhat active.

Thank you very much.
Jim Keenan

khwilliamson:
On 12/29/20 7:52 PM, James E Keenan wrote:
> On 12/29/20 8:55 PM, Leon Timmermans wrote:
>  > How do we proceed?
>  >
>  > Probably this deserves some discussion on the mailing list.
>  >
> 
> I agree; this should be discussed on list. When, over the years, I have
> suggested cleaning up leading whitespace in source code files, I have
> gotten considerable push back from major contributors, some of whom are
> still somewhat active.
> 
> Thank you very much.
> Jim Keenan
> 

The compromise that was adopted then, and is still in effect, was to add 
editor hints to expand tabs, so that as lines get touched they would 
lose their tabs, but un-modified lines would not show up when doing 
searches through history trying to understand how bugs got introduced.

A problem is that vim doesn't necessarily expand tabs in portions of the 
changed line that weren't modified.

Maybe people are now more comfortable with git blame having a white 
space ignore feature.  I have that turned on all the time, and it does 
noticeably slow things down, but it's worth it to me.

hvds:
I'd love to get a patch something like this merged. Ideally, such a patch should also include a test to ensure we don't regress - even more ideally, in a form that can easily be adapted as other files are brought into the fold.

I suspect it will be a bit of a nightmare for some long-lived branches to rebase past - I have at least one such. I hope we can minimize the number of PRs to complete this work, to minimize the cost of such efforts.

schwern:
> I'd love to get a patch something like this merged. Ideally, such a patch should also include a test to ensure we don't regress - even more ideally, in a form that can easily be adapted as other files are brought into the fold.

There are plenty of existing style checkers out there which can be added as automatic checks on PRs. I would suggest taking that route instead of writing more custom software.

However, please don't hold up this PR until that is all in place. Instead, merge #18447 which adds an [EditorConfig](https://editorconfig.org/) file to provide a style hint to most editors. That can be done right now.

> I suspect it will be a bit of a nightmare for some long-lived branches to rebase past

Rebasing past this is straight forward. Run `expand -i` on some files and run a few regen scripts. I can provide a script.

Long lived branches and large refactorings are why this normalization needs to be done. I have a few refactorings I'd like to do which are made far more difficult because of the inconsistent and shifting indentation. It's impossible to keep up-to-date with blead.

schwern:
> The compromise that was adopted then, and is still in effect, was to add editor hints to expand tabs, so that as lines get touched they would lose their tabs, but un-modified lines would not show up when doing searches through history trying to understand how bugs got introduced.

It's been DECADES and the result is a jumbled mix of styles and indentation that just makes everything more complicated for everyone. I'm happy to do the work to make it happen and to make it as painless as possible, but I'm not interested in spending another 10 years debating its merits.

As a code archeologist, I try not to get in the way of people getting work done and improving the code and making it more approachable and easier to work on. Every time I click through a style commit that's a reminder that the code better. I have a Git alias to minimize the impact of whitespace refactoring and file renames: `archeology = blame -w -C -M -n`.

If there's concrete issues, say what they are, let's address them, get this merged, and let's spend our energy on making things better.

schwern:
Here's a bash script to detabify the C code at will. That should make updating any long lived branches easier.

```
#!/bin/bash

shopt -s globstar
shopt -s extglob

# One of the regen scripts needs Devel::Tokenizer::C.
export PERL5LIB=/path/to/Devel::Tokenizer::C

# Expand tab indentation on all code C code and regen scripts.
# Except vutil.*, that's part of version.pm
# Except cpan/ modules, they're maintained on their own.
EXPAND_FILES=$(git ls-files !(cpan)/**/*.[ch] !(vutil).[ch] regen/*.pl regen_perly.pl)
for f in $EXPAND_FILES; do
    echo "Expanding tabs on $f";
    expand -i $f | sponge $f;
done

# Run all the regen scripts to ensure all generated C code is expanded.
# Skip regen_perly.pl, that re-runs the tokenizer which is beyond our scope.
REGEN_FILES=$(git ls-files regen/*.pl)
for f in $REGEN_FILES; do
    echo "Running $f";
    ./perl -Ilib $f;
done
```

schwern:
Actually, all that is unnecessary. To update a branch, use the `ignore-space-change` merge strategy. `git rebase -X ignore-space-change` (or merge).

schwern:
In sum...

* Branches can be updated with `git rebase -X ignore-space-change` (or merge).
* Agree that code archeology serves us and archeology will adapt to changes.
    * Use `git blame -w` and Github ignore whitespace to ignore this commit.
    * `archeology = blame -w -C -M -n` is a nice alias to avoid a multitude of trivial changes.
* Merge #18447 which adds a universal editor style config file to help keep the basic style sane.
* Look into adding a PR check for tabs *after* this is merged.

Leont:
> In sum...
> 
>     * Branches can be updated with `git rebase -X ignore-space-change` (or merge).
> 
>     * Agree that code archeology serves us and archeology will adapt to changes.
>       
>       * Use `git blame -w` and Github ignore whitespace to ignore this commit.
>       * `archeology = blame -w -C -M -n` is a nice alias to avoid a multitude of trivial changes.
> 
>     * Merge #18447 which adds a universal editor style config file to help keep the basic style sane.
> 
>     * Look into adding a PR check for tabs _after_ this is merged.

This should probably be relayed to the list.

schwern:
@Leont That would make the information ephemeral. I'll look into adding it to the patching docs, plus a mention of EditorConfig.

Leont:
> @Leont That would make the information ephemeral. I'll look into adding it to the patching docs, plus a mention of EditorConfig.

porque no los dos

schwern:
@Leont Because people who submit PRs should not also have to join and post on a high traffic mailing list and go over the whole issue again in parallel.

Issue #17331, opened 2019-11-27
[doc] Referencing GitHub issues?


Grinnz:
Currently, https://github.com/Perl/perl5/blob/blead/Porting/how_to_write_a_perldelta.pod#style says that rt.perl.org tickets should be referenced like `perl #43010` or `[perl #43010]`. Similarly, references exist throughout the source code in comments and documentation in that style as well as things like `[rt.perl.org #43010]`, `rt #43010`, `RT #43010`, or omitting the space or hash. What style should be recommended for referencing GitHub issues, and what should be used when converting these existing references to the corresponding GitHub issue number?

My suggestion is `[github #43010]` and `GH #43010` accordingly; reusing `[perl #43010]` while strictly accurate would unfortunately be ambiguous with RT ticket numbers.

Grinnz:
In https://perldoc.pl/5.31.6/perldelta issues were referenced simply as [#43010]. This also seems too ambiguous.

khwilliamson:
I'm fine with this change

tonycoz:
Yeah, the bare ids `[#xxxx]` is mostly my fault, I think I did all of them in the 5.31.6 delta except one, and Karl probably followed my example.

I'm not really fond of any of the possibilities, `[github #xxxx]` or `[gh #xxxx]` implies we're talking about issues in github itself.

Right now I think `[#xxxx]` is the best possiblity, it's distinct from `[perl #xxxx]`, and the 5.31.6 delta mentions we've moved to github.

Mentions in code, including tests should probably refer to gh or github issues.

I can see it being a problem for @xsawyerx for the 5.32 release, he might choose to leave the rt references in 5.31.[0-5] as `[perl #xxxx]` for the 5.32.0 delta or translate them all to github references.

Grinnz:
I agree it's not ideal, but I think it's well established that "GH #XXXXX" refers to a github issue for whatever project you happen to be discussing, given the widespread nature.

Grinnz:
Just for prior art, though of course none of these are in the situation of having to disambiguate from a prior tracker:

* Elasticsearch uses bare issue numbers (https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-7.0.0.html)
* VSCode uses bare issue numbers, and prefix PR for pull requests (https://code.visualstudio.com/updates/v1_40)
* PHP uses bare issue numbers (https://www.php.net/ChangeLog-7.php)

Grinnz:
[...]

PR #18453, opened 2021-01-04
rt.perl.org no longer stores Perl 5 bugs

tomhukins:


Grinnz:
Related: https://github.com/Perl/perl5/issues/17331

PR #18456, opened 2021-01-04
Fix windows compilation

dlaugt:
MSVC gives the following compilation error when PERL_TEXTMODE_SCRIPTS is not defined:
..\toke.c(6701) : error C2065: 'RSFP_FILENO' : undeclared identifier


tonycoz:
I applied one of the intermediate commits as 7e8ad094bab6acb9ea36c1fc61356052a737e31a and updated AUTHORS in 42ee5e7db05ee0d07ff1d182f7c85e4d9c8bbf5e

Issue #18461, opened 2021-01-08
[feature] Make default typemap type-mismatch errors give the passed-in SV’s information


FGasper:
I’m chasing one of these:
```
    (in cleanup) DNS::Unbound::_destroy_context: ctx is not of type struct ub_ctxPtr at /usr/local/cpanel/3rdparty/perl/532/lib/perl5/cpanel_lib/x86_64-linux-64int/DNS/Unbound.pm line 731 during global destruction.
```
Currently DNS::Unbound uses the default T_PTROBJ typemap. I’m about to push up a change to the module to use a custom typemap that will print this instead:
```
DNS::Unbound::_destroy_context: ctx (scalar: 123123) is not of type struct ub_ctxPtr at …
```
I thought I’d throw this up here to gauge interest in making this change to upstream. I’m including a provisional PR.

Thank you!

tonycoz:
Looking at the commit in your repository:

```
const char* valstr = SvROK($arg) ? sv_2pvbyte_nolen($arg) : SvPV_nolen($arg);
```

why differentiate?  (and why call the function directly?)  Also, the object could be blessed into a non-ASCII package name,

Also this could just use the SVf macro:
```
Perl_croak_nocontext(\"%s: %s (%s" SVf ") is not of type %s\",
```

and supply `$arg` instead of `valstr`.  This also handles unicode strings and stash names for blessed objects.

jkeenan:
@FGasper, would you like to respond to @tonycoz's reply?  Otherwise, it looks like we should close this feature request.

Thank you very much.
Jim Keenan

FGasper:
@jkeenan I updated the PR to incorporate @tonycoz’s suggestion a bit back. I didn’t update this thread though .. sorry. :(

PR #18462, opened 2021-01-08
Give SV information in default typemap

FGasper:
Issue #18461

### DIFF
--- a/lib/ExtUtils/typemap
+++ b/lib/ExtUtils/typemap
@@ -217,38 +217,58 @@ T_REF_IV_REF
 	    IV tmp = SvIV((SV*)SvRV($arg));
 	    $var = *INT2PTR($type *, tmp);
 	}
-	else
-	    Perl_croak_nocontext(\"%s: %s is not of type %s\",
+	else {
+		const char* typestr = SvROK($arg) ? \"ref: \" : SvOK($arg) ? \"scalar: \" : \"undef\";
+		const char* valstr = SvROK($arg) ? sv_2pvbyte_nolen($arg) : SvPV_nolen($arg);
+	    Perl_croak_nocontext(\"%s: %s (%s%s) is not of type %s\",
 			${$ALIAS?\q[GvNAME(CvGV(cv))]:\qq[\"$pname\"]},
-			\"$var\", \"$ntype\")
+			\"$var\",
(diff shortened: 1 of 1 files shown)


Leont:
I'm not sure this error message is really more useful than the previous one. Could you explain to me how it would be helpful?

FGasper:
> I'm not sure this error message is really more useful than the previous one. Could you explain to me how it would be helpful?

@Leont By showing what the wrong thing is—not merely that it’s the wrong type of thing—it’ll be easier to determine how we got there in the first place.

Apparently there’s a case where DNS::Unbound has the wrong type of internal object and is sending it to the destructor, leading to this:
```
(in cleanup) DNS::Unbound::_destroy_context: ctx is not of type struct ub_ctxPtr
```
There’s already a filter in place to ensure that global-destruct doesn’t send undef into that _destroy_context() function; it looks like there’s somehow some other non-falsy value that’s being stored.

I’m altering DNS::Unbound to tell me what this value is; I thought it would be useful upstream as well.

Leont:
> By showing what the wrong thing is—not merely that it’s the wrong type of thing—it’ll be easier to determine how we got there in the first place.

I get your intention, but `scalar: 123123` isn't the clearest way of showing what is wrong. If something should have been a reference but isn't, why don't we just say that? And likewise if something is blessed into the wrong class.

FGasper:
@Leont If I take your meaning, you’re looking for different wording?

Maybe:

> Expected `ctx` to be a `struct ub_ctxPtr` reference; received `ARRAY` reference instead.

> Expected `ctx` to be a `struct ub_ctxPtr` reference; received scalar (123123) instead.

> Expected `ctx` to be a `struct ub_ctxPtr` reference; received undef instead.

FGasper:
Feedback from @Leont @tonycoz incorporated. Further thoughts?

Thank you!

jkeenan:
> Issue #18461

Are there any places in the test suite where we exercise the *existing* error messages?

If so, then we may need to modify those tests?

If not, then perhaps we should write tests to exercise them.

Thank you very much.
Jim Keenan

tonycoz:
`ext/XS-Typemap/t/Typemap.t` tests that error messages are produced in a limited set of cases, but not their content

Issue #18469, opened 2021-01-12
Getopt::Long silently overwrites parameter values if same param given twice


epa:
This is a bug report against `Getopt::Long`, a core module.  You can specify named parameters taking an argument, for example

`GetOptions("foo=s")`

This specifies an optional `--foo` parameter which, if given, must be followed by a string value (either as the next argument or as the form `--foo=X`).  It gets stored in `$opt_foo`.  This works:

```
% perl -MGetopt::Long -E 'GetOptions("foo=s") or die; say $opt_foo' -- --foo hello
hello
```

And the `GetOptions()` call returns false if there is a usage error, for example an unrecognized parameter:

```
% perl -MGetopt::Long -E 'GetOptions("foo=s") or die; say $opt_foo' -- --boo
Unknown option: boo
Died at -e line 1.
```

But what about the error of giving the `--foo` parameter twice?  It is declared to take a single scalar.  I would expect passing multiple copies of the option to give a usage error.  (If the script did support multiple values, it would have used the `s@` option specifier instead of plain `s`, and the values would go into the array `@opt_foo`.)  But no usage message is given.  Instead, the earlier values are silently discarded:

```
% perl -MGetopt::Long -E 'GetOptions("foo=s") or die; say $opt_foo' -- --foo hello --foo goodbye
goodbye
```

As far as I can tell this behaviour isn't documented.  I think the right behaviour would be for `GetOptions()` to return false in this case as it does for other usage errors.

Of course, this is a very old and widely used module, so you may not want to risk breaking existing users with such a change.  (Personally, I would say that a usage was already broken, but in backwards compatibility questions you have to tread carefully.)  Therefore, as a less disruptive step, it may be better to print a warning by default, with a configuration option for the new stricter behaviour.

Leont:
First of all, this is the wrong tracker for `Getopt::Long` issues, its issue tracker can be found at https://rt.cpan.org/Public/Dist/Display.html?Name=Getopt-Long

Secondly, I think most people would consider this expected behavior (if only because all other argument parsers act this way), so I don't think a warning-by-default is inappropriate. And indeed, backwards compatibility is an additional reason not to change the default behavior.

If you really want to insist on this behavior, you could use a callback:
```
perl -MGetopt::Long -E 'GetOptions("foo=s" => sub { die "Can only set foo once" if defined $foo; $foo = $_[1] }) or exit 2; say $foo' -- --foo hello --foo world
```

epa:
[...]

PR #18471, opened 2021-01-12
Tests and documentation for Getopt::Long behaviour when the same option appears twice, but is not multi-valued.

epa:
<https://github.com/Perl/perl5/issues/18469>


Grinnz:
CPAN-upstream modules cannot be updated in blead.

Grinnz:
The repository for Getopt::Long is at https://github.com/sciurius/perl-Getopt-Long.

PR #18472, opened 2021-01-13
Call magic on all elements on %SIG delocalization

Leont:
This intends to fix https://www.nntp.perl.org/group/perl.perl5.porters/2021/01/msg258776.html


khwilliamson:
Is this ready to be merged?

Leont:
> Is this ready to be merged?

As far as I'm concerned it is

PR #18478, opened 2021-01-14
Fix typos in conditional macro names in perl.h

t-a-k:
I found some trivial typos in conditional macro names in `perl.h`.
(I don't have relevant platforms to test this change, though.)


sisyphus:
@t-a-k's proposal looks correct to me.

Cheers,
Rob

PR #18479, opened 2021-01-14
Allow empty lower bound in /{,n}/ quantifier; allow blanks inside {...} constructs

khwilliamson:


khwilliamson:
'In effect' seems perfectly reasonable to me.  What don't you like about it?

khwilliamson:
@rwp0 I misread your comment.  Fixed in the final merge.

This PR was merged through other means

PR #18480, opened 2021-01-15
update the links to freenode

gugod:
Hi comitters,

I found two irc URLs that ref to the old hostname of freenome (`"irc.freenode.net"`) and they are replaced with the preferred one now (that  is `"chat.freenode.net"`).


atoomic:
Thanks for the update this makes sense. As seen here https://freenode.net/kb/answer/chat `chat.freenode.net` is the preferred hostname to use.

PR #18482, opened 2021-01-17
perlre: Note the other forms of \k<name>

khwilliamson:
Not all three synonyms were documented.

This also fixes up related comments in regcomp.c to correspond


khwilliamson:
Fixed instead by a44b2be795f4c5f94384c6f6010860588e144b3c

PR #18490, opened 2021-01-18
Fix POD for $Carp::RefArgFormatter

dboehmer:
- Data::Dumper::Dump() MUST be called as class method

- indent example for code markup


Grinnz (approved):
This needs a version bump in Carp.pm and an addition to AUTHORS which can both be done by a committer.

dboehmer:
> This needs a version bump in Carp.pm and an addition to AUTHORS which can both be done by a committer.

At first wasn’t sure how to bump version number but then found previous commit and did the same. Force pushed a single commit doing everything in 1 step.

Grinnz:
Sorry I meant, someone who merges this can do it. But you doing it is fine too.

Issue #18276, opened 2020-10-30
missing error message in perldiag.pod for unicode properties


karenetheridge:
The diagnostic message `Unknown user-defined property name \p{%s}` is not documented in pod/perldiag.pod.
It originates from regcomp.c.

(Maybe not relevant, but it also looks like it changed, at least in some usecases, from `Can't find Unicode property definition "%s"` between 5.28 and 5.30.)

There are other error messages in this file that are also undocumented:
- "Timeout waiting for another thread to define .."
- "Infinite recursion in user-defined property"
- "Error.. in expansion of ..."
- "Insecure private-use override"

apparluk:
The below error message but not the others is found in 5.28.3.

`Can't find Unicode property definition "%s"`

jkeenan:
> The diagnostic message `Unknown user-defined property name \p{%s}` is not documented in pod/perldiag.pod.
> It originates from regcomp.c.
> 
> (Maybe not relevant, but it also looks like it changed, at least in some usecases, from `Can't find Unicode property definition "%s"` between 5.28 and 5.30.)
> 
> There are other error messages in this file that are also undocumented:
> 
>     * "Timeout waiting for another thread to define .."
> 
>     * "Infinite recursion in user-defined property"
> 
>     * "Error.. in expansion of ..."
> 
>     * "Insecure private-use override"

We might be able to work up some documentation by looking for places where these warnings are exercised in the test suite.

* "Unknown user-defined property name"
```
t/re/pat.t
2112:        fresh_perl_like('"q0" =~ /\p{__::Is0}/', qr/Unknown user-defined property name \\p\{__::Is0}/, {}, "GH #17384");

t/re/reg_mesg.t
266: 'm/\p{Is_Is_Any}/' => 'Unknown user-defined property name \p{main::Is_Is_Any}',
```

* "Timeout waiting for another thread to define .."
```
t/re/user_prop_race_thr.t
112:    qr/Thread \d+ terminated abnormally: Timeout waiting for another thread to define "InLongSleep" in regex/,

```
* "Infinite recursion in user-defined property"
```
t/re/regexp_unicode_prop.t
163:        'InRecursedA'   => qr/Infinite recursion in user-defined property (?#
```
* "Error.. in expansion of ..."
```
t/re/regexp_unicode_prop.t
161:                             )"0\t$overflow$utf8_comment" in expansion of (?#
164:                             )"main::InRecursedA" in expansion of (?#
[...]

PR #18491, opened 2021-01-18
Add missing entries to perldiag; GH #18276

khwilliamson:
The ticket mentions yet another message, not addressed in this
commit, "Insecure private-use override".  That message is part of a
hook for a so-far unimplemented module, so it actually doesn't ever get
raised.

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -2204,6 +2204,20 @@ single form when it must operate on them directly.  Either you've passed
 an invalid file specification to Perl, or you've found a case the
 conversion routines don't handle.  Drat.
 
+=item Error %s in expansion of %s
+
+(F) An error was encountered in handling a user-defined property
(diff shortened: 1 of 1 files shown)


pod/perldiag.pod
```diff
@@ -2845,6 +2859,16 @@ not match 8 spaces.
 text.  You should check the pattern to ensure that recursive patterns
 either consume text or fail.
 
+=item Infinite recursion in user-defined property
+
+(F) A user-defined property (L<perlunicode/User-Defined Character
+Properties>) can to depend on the definitions of other user-defined
```
Grinnz, line 2865: "can depend"

jkeenan:
Manually rebased on blead with @Grinnz correction.  Merged to blead, pushed to repository in cac1381071.  Closing ticket.

Thank you very much.
Jim Keenan

PR #18492, opened 2021-01-18
Msys2 upstream

Corion:
This incorporates changes from https://github.com/msys2/MSYS2-packages/blob/master/perl/perl.cygwin-Win32.patch into Perl.

I've split up the PR into several commits to make reviewing easier hopefully. Also, hopefully, bisecting a problem points more closely to the problem than one large patch.

There are separate `.patch`files for `Configure` which I did not apply or touch (yet)

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5263,6 +5263,7 @@ hints/minix.sh				Hints for named architecture
 hints/mips.sh				Hints for named architecture
 hints/mirbsd.sh				Hints for named architecture
 hints/mpc.sh				Hints for named architecture
+hints/msys.sh				Hints for named architecture
 hints/ncr_tower.sh			Hints for named architecture
 hints/netbsd.sh				Hints for named architecture
 hints/newsos4.sh			Hints for named architecture
--- a/MANIFEST
+++ b/MANIFEST
@@ -4124,6 +4124,7 @@ dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/android.pm	CBuilder method
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/cygwin.pm		CBuilder methods for cygwin
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/darwin.pm		CBuilder methods for darwin
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/dec_osf.pm	CBuilder methods for OSF
+dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm		CBuilder methods for Msys
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/os2.pm		CBuilder methods for OS/2
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/Unix.pm		CBuilder methods for Unix
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/VMS.pm		CBuilder methods for VMS
--- a/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm
+++ b/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm
@@ -1,0 +1,33 @@
+package ExtUtils::CBuilder::Platform::msys;
+
+use warnings;
+use strict;
+use File::Spec;
+use ExtUtils::CBuilder::Platform::Unix;
+
+our $VERSION = '0.280234'; # VERSION
+our @ISA = qw(ExtUtils::CBuilder::Platform::Unix);
+
+# TODO: If a specific exe_file name is requested, if the exe created
+# doesn't have that name, we might want to rename it.  Apparently asking
+# for an exe of "foo" might result in "foo.exe".  Alternatively, we should
+# make sure the return value is correctly "foo.exe".
+# C.f http://rt.cpan.org/Public/Bug/Display.html?id=41003
+sub link_executable {
+  my $self = shift;
+  return $self->SUPER::link_executable(@_);
+}
+
+sub link {
+  my ($self, %args) = @_;
+
+  my $lib = $self->{config}{useshrplib} ? 'libperl.dll.a' : 'libperl.a';
+  $args{extra_linker_flags} = [
+    File::Spec->catfile($self->perl_inc(), $lib),
+    $self->split_like_shell($args{extra_linker_flags})
+  ];
+
+  return $self->SUPER::link(%args);
+}
+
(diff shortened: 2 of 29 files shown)


Makefile.SH
```diff
@@ -1542,7 +1543,7 @@ case "$targethost" in
 '') $spitshell >>$Makefile <<'!NO!SUBS!'
 test_prep test-prep: test_prep_pre $(MINIPERL_EXE) $(unidatafiles) $(PERL_EXE) \
 	$(dynamic_ext) $(TEST_PERL_DLL) runtests $(generated_pods) common_build
-	cd t && (rm -f $(PERL_EXE); $(LNS) ../$(PERL_EXE) $(PERL_EXE))
+	cd t && (rm -f $(PERL_EXE) $(LIBPERL); $(LNS) ../$(PERL_EXE) $(PERL_EXE); $(LNS) ../$(LIBPERL) $(LIBPERL))
```
Leont: This change looks dubious to me. It might be correct on msys (though I'd be curious as to why) but I'm not sure we should do it for the other operating systems (that certainly do not need libperl to be in `t/`

ext/NDBM_File/Makefile.PL
```diff
@@ -1,7 +1,16 @@
...
+our @LIBS;
+if( $^O eq 'msys' ) {
+    @LIBS = ("-lgdbm -lgdbm_compat");
+} else {
+    @LIBS = ("-L/usr/local/lib -lndbm", "-ldbm -lucb");
+};
+
 WriteMakefile(
     NAME	=> 'NDBM_File',
-    LIBS => ["-L/usr/local/lib -lndbm", "-ldbm -lucb"],
```
Leont, line 4: This is not how we typically do this. Adding `"-lgdbm -lgdbm_compat"` as a third possible value for `LIBS` would be.

That said, I'm not sure we should do this at all. Instead of abusing gdbm's compat layer, such a perl should just use `GDBM_File`

toddr:
FYI porting/manifest.t is failing right now.

Leont:
> FYI porting/manifest.t is failing right now.

Yeah, `dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm` needs to be added to `MANIFEST`

dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm
```diff
@@ -0,0 +1,33 @@
+package ExtUtils::CBuilder::Platform::msys;
```
Leont, line 1: If this is exactly the same as `ExtUtils::CBuilder::Platform::cygwin`, shouldn't it just inherit from that instead?

Corion:
If the two files are exactly identical then we could do with inheritance. They should be fairly similar environments anyway.

Leont:
Probably #18534 needs to be merged before this one.

ext/NDBM_File/Makefile.PL
Leont, line 4: cygwin uses a `hints/` file for this, msys should probably do exactly the same

demerphq:
@corion this ticket is stalled and in conflict. Do you have any intention to rebase it to blead? It seems like people are cautiously enthusiastic about it.

jkeenan:
> @Corion this ticket is stalled and in conflict. Do you have any intention to rebase it to blead? It seems like people are cautiously enthusiastic about it.

@Corion, ^^

Corion:
I'll look at it - thanks for keeping up with this!

Corion:
Rebased to current blead

ext/DynaLoader/DynaLoader_pm.PL
```diff
@@ -484,6 +484,9 @@ sub dl_findfile {
 	    <<$^O-eq-cygwin>>
             push(@names,"cyg$_.$dl_so")  unless m:/:;
 	    <</$^O-eq-cygwin>>
+	    <<$^O-eq-msys>>
```
Leont, line 487: This commit should also update the version to `'1.55'`

ext/NDBM_File/Makefile.PL
Leont: I suspect this entire patch isn't needed, can we double check that?

MANIFEST
```diff
@@ -3882,6 +3882,7 @@ dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/android.pm	CBuilder method
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/cygwin.pm		CBuilder methods for cygwin
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/darwin.pm		CBuilder methods for darwin
 dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/dec_osf.pm	CBuilder methods for OSF
+dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder/Platform/msys.pm		CBuilder methods for Msys
```
Leont, line 4127: This commit really should be merged with "Support msys in tools" (or possibly there should be a ExtUtils::CBuilder patch).

Corion: I've squashed the two commits together - the EU:CBuilder repo says not to have pull requests for it :)

ext/DynaLoader/DynaLoader_pm.PL
Corion: Done

goyalyashpal:
hi! is this like merged? (as in, sometimes commits get merged using git cli in a way that gh doesn't detect it)

jkeenan:
> hi! is this like merged? (as in, sometimes commits get merged using git cli in a way that gh doesn't detect it)

No, it does not appear to have been merged.  In fact, it has merge conflicts.  @Corion, can you give us an update on the status of this pull request?

Corion:
Rebased. I think the two comments by @Leont still stand, but I can't promise that I can look into them. Maybe the rest should be applied and these two left out?

Leont:
> hi! is this like merged? (as in, sometimes commits get merged using git cli in a way that gh doesn't detect it)

Somehow this work got spread over multiple tickets. There's some additional work [here](https://github.com/Perl/perl5/issues/18489#issuecomment-1673557268) but I haven't cut it into clean patches yet.

khwilliamson:
It would be nice to get this merged. @leont, what needs to be done here besides the rebase

Leont:
> It would be nice to get this merged. @Leont, what needs to be done here besides the rebase

We need to merge some of the patches from #18489 into this branch, that would get us pretty close. But what we really need is someone (on Windows) who can run msys and actually fix issues until the whole thing works.

Corion:
As I have a Windows machine - even if it is not my main machine anymore - I'll see if I can install an msys environment and can build perl there

PR #18493, opened 2021-01-19
Prepare reg_mesg.t for warnings enabled upon invocation

khwilliamson:
Since some of the tests in this file test for default warning behavior,
it must set the warnings to the defaults.

### DIFF
--- a/t/re/reg_mesg.t
+++ b/t/re/reg_mesg.t
@@ -14,6 +14,12 @@ skip_all_without_unicode_tables();
 use strict;
 use open qw(:utf8 :std);
 
+# Show that it works when all warnings are enabled upon invocation.  This file
+# includes tests that the default warnings are enabled by default, and the
+# non-default ones aren't.
+use warnings;
+BEGIN { ${^WARNING_BITS} = undef }  # Kludge to restore default warnings
+
 # Kind of a kludge to mark warnings to be expected only if we are testing
 # under "use re 'strict'"
 my $only_strict_marker = ':expected_only_under_strict';


t/re/reg_mesg.t
```diff
@@ -14,6 +14,12 @@ skip_all_without_unicode_tables();
 use strict;
 use open qw(:utf8 :std);
 
+# Show that it works when all warnings are enabled upon invocation.  This file
+# includes tests that the default warnings are enabled by default, and the
+# non-default ones aren't.
+use warnings;
+BEGIN { ${^WARNING_BITS} = undef }  # Kludge to restore default warnings
```
tonycoz, line 21: As I mentioned but didn't expand on in IRC, this probably isn't suitable for your needs.

If we do implement "warnings by default" by setting `cop_warnings` to `pWARN_ALL` then the `BEGIN {}` block will set warnings back to the current defaults, I don't think that's your intent here but I might be wrong.

I think the better way to do it is to save `${^WARNING_BITS}` in a BEGIN block before any other code, and set `${^WARNING_BITS}` (again in a BEGIN block) to that saved value.

toddr: I'm not clear why simply saying `no warnings` wouldn't work here?

Grinnz: no warnings disables all warnings, including [default warnings](https://perldoc.perl.org/warnings#Default-Warnings-and-Optional-Warnings).

toddr: AFAIK, we're only discussing `${^WARNING_BITS}`

```
$>perl -e 'BEGIN {  print "undef\n" if !defined ${^WARNING_BITS} }'      
undef

$>perl -e 'BEGIN {  no warnings; print "undef\n" if !defined ${^WARNING_BITS} }'
undef

# And just for good measure:
$>perl -w -e 'BEGIN {  no warnings; print "undef\n" if !defined ${^WARNING_BITS} }'
undef
```

Grinnz: The tests rely on the default set of warnings being enabled. `no warnings` disables those, rather than reverting to the default set of warnings, as described in the docs I linked.

khwilliamson: Grinnz is right

@tonycoz my intention was merely to get this file to work for when the switch is flipped to enable warnings, with the expectation that it would need to be revisited later.

tonycoz: ok, I thought you were after default warnings, whether that changed or not.

jkeenan:
This patch looks mergeable.  If there is no further discussion, I'll merge it in 7 days (unless someone else goes ahead first).

Thank you very much.
Jim Keenan

toddr (approved):

tonycoz (approved):

PR #18494, opened 2021-01-19
ExtUtils-CBuilder: Remove image-base generation on Win32/gcc

zmughal:
Switches from generating an image-base address using the basename of the
output file to using GCC's built-in `--enable-auto-image-base` linker
option.

This aligns the linking behaviour for Win32/gcc with that of
ExtUtils-MakeMaker which removed image-base generation in commit
<https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/343d21a453c4d03cf7304dbd4c4dd8180df574ad>.

CC: @ambs, @Leont, @plicease

Connects with: <https://github.com/PerlAlien/Alien-Build/pull/243#issuecomment-758193294>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1356,6 +1356,7 @@ Yves Orton                     <demerphq@gmail.com>
 Zachary Miller                 <zcmiller@simon.er.usgs.gov>
 Zachary Storer                 <zacts.3.14159@gmail.com>
 Zak B. Elep                    <zakame@zakame.net>
+Zakariyya Mughal               <zmughal@cpan.org>
 Zbynek Vyskovsky               <kvr@centrum.cz>
 Zefram                         <zefram@fysh.org>
 Zsbán Ambrus                   <ambrus@math.bme.hu>
--- a/dist/ExtUtils-CBuilder/Changes
+++ b/dist/ExtUtils-CBuilder/Changes
@@ -1,5 +1,12 @@
 Revision history for Perl extension ExtUtils::CBuilder.
 
+0.280236 - 2021-02-12
+
+  Fix:
+
+  - Remove image-base generation on Win32/gcc and instead use GCC's built-in
+    `--enable-auto-image-base` linker option.
+
 0.280235 - 2020-11-01
 
   Fix:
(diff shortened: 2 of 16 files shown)


xenu:
This is a reasonable change, +1.

In the future we should switch to `--high-entropy-va` to make ASLR work, but that would be a bigger patch. We would have to detect whether the linker supports the flag and then put in `$Config{lddconfig}` or somewhere.

zmughal:
I just wanted to note that under MSYS2/MINGW64, the default build options use ASLR in [`/etc/makepkg_mingw64.conf`](https://github.com/msys2/MSYS2-packages/blob/a5855c6458afc6a62628af2ef2a8d5db2d636637/pacman/makepkg_mingw64.conf#L62):
```sh
LDFLAGS="-pipe -Wl,--dynamicbase,--high-entropy-va,--nxcompat,--default-image-base-high"
```

More discussion on this for MINGW at <https://github.com/msys2/MINGW-packages/issues/6674>. However the `mingw-w64-x86_64-perl` package's Config.pm does not have these flags.

ambs:
Please let me know when/if it gets merged, so I can prepare the CPAN release of the module.

jkeenan:
To facilitate evaluation of this pull request, I have created the following smoke-me branch:
smoke-me/jkeenan/zmughal-extutils-cbuilder-win32-gcc-auto-image-base

Could we get people with access to Win32 to test this, preferably with gcc and other compilers?

Thank you very much.
Jim Keenan

xenu:
@ambs  it is merged now.

ambs:
Thank you. Will release EU::CB tomorrow.

ambs:
Done. Released with the same version it is on Blead. Hope that isn't a problem. I can't recall if that can cause problems.

Leont:
I'm not sure this change was complete, really. We're currently doing this bizarre `dlltool;ld; dlltool; ld` dance that should not be necessary anymore, AFAIK we can just do one `ld` now that we have `--enable-auto-image-base`.

And it appears that not doing the simple thing has a unfortunate interaction with antivirus scanners https://github.com/Leont/magic-check/issues/1

Issue #18495, opened 2021-01-19
PerlIO::encoding::fallback corrupts filehandle


maros:
Module: PerlIO::encoding

**Description**
Setting the $PerlIO::encoding::fallback and using PerlIO::ecoding results in a broken filehandle that only returns the first 1024 characters of the file over and over. see https://perldoc.perl.org/PerlIO::encoding#DESCRIPTION

**Steps to Reproduce**
```
use PerlIO::encoding;
use Encode qw(:fallbacks);
$PerlIO::encoding::fallback = FB_PERLQQ; # or anything else like a CODEREF
open my $fh, "<:encoding(utf-8)","some_file.txt"; # some file with > 1024 characters
while (<$fh>) { print } 
```

**Expected behavior**
Print the content of the file, with all octets NOT mapping to utf-8 being replaced with \xHH (with HH being the hex representation of the octet). Instead the first 1024 characters of the file are printed over and over.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 30 subversion 1) configuration:

  Platform:
    osname=linux
    osvers=4.19.0-11-amd64
    archname=x86_64-linux
    uname='linux gh-chroot-2020-11-04-1604493449-kqu4gwl 4.19.0-11-amd64 #1 smp debian 4.19.146-1 (2020-09-17) x86_64 gnulinux '
    config_args='-Dprefix=/opt/gh/perl-5.30.1-live -des -Dusedevel -Duseshrplib -Accflags=-DNO_TAINT_SUPPORT -Accflags=-mtune=core2 -Accflags=-fPIC -Accflags=-DPERL_DISABLE_PMC -Accflags=-ggdb -Dman1dir=none -Dman3dir=none -A'eval:scriptdir=/opt/gh/perl-5.30.1-live/bin''
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='cc'
    ccflags ='-DNO_TAINT_SUPPORT -mtune=core2 -fPIC -DPERL_DISABLE_PMC -ggdb -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
[...]

PR #18496, opened 2021-01-19
Make PerlIO::encoding::fallback more robust

Leont:
This branch
* Forces `$fallback` to always contain `STOP_AT_PARTIAL`, as without it multibyte characters spanning buffer boundaries will not be correctly read
* Disallows setting `$fallback` to a subroutine. This is incompatible with `STOP_AT_PARTIAL` and as such inherently buggy.
* Removes setting `$fallback` from `encoding.xs`, as it's already set from `encoding.pm`.
* Ignores the LEAVE_SRC bit if it's set, as it leads to the issue described in #18495

### DIFF
--- a/ext/PerlIO-encoding/encoding.pm
+++ b/ext/PerlIO-encoding/encoding.pm
@@ -1,7 +1,7 @@
 package PerlIO::encoding;
 
 use strict;
-our $VERSION = '0.28';
+our $VERSION = '0.29';
 our $DEBUG = 0;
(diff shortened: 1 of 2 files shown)


tonycoz (approved):
Looks good to me

jkeenan:
@Leont, per smokers, this looks good to go.  But should we use "Rebase and merge" or "Squash and merge" if merging from the GUI?

Thank you very much.
Jim Keenan

PR #18500, opened 2021-01-20
Set $Config{libpth} properly for MinGW builds

xenu:
Previously the default libpth consisted of just a single folder and
failed to include the directory that contains the majority of the
libraries.

This is a fairly important issue but no one noticed it for two reasons:

1. EU::MM on Windows *always* links XS modules with the libraries from
   $Config{libs}, so you'd notice that linking doesn't work only if you
   needed a library that isn't listed there.

2. Strawberry Perl [has a workaround](https://github.com/StrawberryPerl/Perl-Dist-Strawberry/blob/2112b8a590882e913e98e4aa2dced4f34c4fea79/lib/Perl/Dist/Strawberry/Step/InstallPerlCore.pm#L136) for this issue.

I'm only using MinGW-w64 compilers, so I have no idea how library paths
work on MinGW.org builds. It's possible that the previous libpth worked
fine with them. Either way, this commit only adds new paths to libpth, it
doesn't modify the one that was already there, so it's unlikely it will
break anything.

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -493,6 +493,13 @@ ifeq ($(USE_QUADMATH),define)
 ARCHNAME	:= $(ARCHNAME)-quadmath
 endif
 
+ifeq ($(CCTYPE),GCC)
+GCCVER		:= $(shell $(GCCBIN) -dumpversion)
+GCCVER1		:= $(shell for /f "delims=. tokens=1,2,3" %%i in ('$(GCCBIN) -dumpversion') do echo %%i)
+GCCVER2		:= $(shell for /f "delims=. tokens=1,2,3" %%i in ('$(GCCBIN) -dumpversion') do echo %%j)
+GCCVER3		:= $(shell for /f "delims=. tokens=1,2,3" %%i in ('$(GCCBIN) -dumpversion') do echo %%k)
+endif
+
 # Set the install location of the compiler headers/libraries.
 # These are saved into $Config{incpath} and $Config{libpth}.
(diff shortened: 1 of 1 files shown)


xenu:
Oops, I had accidentally included an unrelated debugging change in this PR. Fixed now.

sisyphus:
I've applied your changes to today's perl-5.33.6 release, and that works fine with all of the ivtype x nvtype permutations available with 32-bit and 64-bit mingw-w64 compilers.
Regarding the compilers, I tested with gcc-8.3.0 (built by Strawberry Perl) and gcc-10.2.0 (built by Brecht Sanders).

I've been using the EXTRALIBDIRS workaround (similar to Strawberry Perl) in my build scripts for so long now, that I had simply forgotten the significance of it.

Should the same fix be made to win32/makefile.mk ?

xenu:
>Should the same fix be made to win32/makefile.mk ?

It should, but

1. I was too lazy to research how to implement the `subst` part in dmake.
2. Quite frankly, I'd rather get rid of Makefile.mk, its existence makes maintenance significantly harder for no benefit at all. It has exactly the same purpose as GNUMakefile, but it uses a much more obscure `make` program. I really don't see the point in keeping it.

xenu:
PS. I have just sent [an e-mail](https://www.nntp.perl.org/group/perl.perl5.porters/2021/01/msg258848.html) to p5p with Makefile.mk removal proposal.

sisyphus:
I've certainly got no objections to the removal of win32/makefile.mk.

Issue #18504, opened 2021-01-22
Adding a try/catch syntax to Perl


leonerd:
I intend to add `try/catch` syntax. This has been one of the most frequently-requested of core syntax features - from both community sources like Freenode #perl and other core developers - so I do not anticipate this will be a contentious or debated idea.

As many of the existing CPAN solutions all follow a very similar theme - including my own [`Syntax::Keyword::Try`](https://metacpan.org/pod/Syntax::Keyword::Try) - I expect I shall follow that pattern, and likewise do not anticipate that this will be considered controversial. `Syntax::Keyword::Try` has been the preferred module by Freenode #perl for a while now, and folks seem happy with it.

I further intend it to be introduced as

```perl
use feature 'try';
```

Towards this goal, I have created a module [`Feature::Compat::Try`](https://metacpan.org/pod/Feature::Compat::Try) to simplify migration. Existing authors can

```perl
use Feature::Compat::Try;
```

in their current code, today, and import a configured version of `Syntax::Keyword::Try` which works on all Perl versions back to 5.16, and conforms exactly to the specification of the thing I want to add to core. (S:K:T additionally contains other features, experimental ideas, and past mistakes, that I do not intend to bring in at this time, hence the configuration).

A brief summary, from its synopsis, is:

```perl
use Feature::Compat::Try;

sub foo
{
   try {
      attempt_a_thing();
      return "success";
   }
   catch ($e) {
      warn "It failed - $e";
      return "failure";
   }
}
```

If anyone wishes to review in more detail the syntax and semantics, please read the documentation and unit-tests of this module. That is what I intend to implement exactly. You do not have to wait until I have a PR to actually implement it to do this - the earlier you do this, the better :)

leonerd:
Cross-posted to perl5-porters@ at
https://www.nntp.perl.org/group/perl.perl5.porters/2021/01/msg258856.html

leonerd:
Work progresses at pace on this branch:

https://github.com/Perl/perl5/tree/leonerd/feature-try

In fact at time of writing, it now passes all the unit tests of `Feature::Compat::Try` except for the one about blessed object exceptions that overload boolification to constant false.

Once I have that working, I shall have a go at the remaining core tasks which didn't affect a CPAN module - such as B::Deparse support. And then it should be ready for review.

Hopefully either by end of weekend, or if not, middle of next week.

leonerd:
There are two points that the current design (docs and tests) leaves entirely unaddressed. Those are

[...]

PR #18505, opened 2021-01-23
An initial attempt at try/catch syntax

leonerd:
This PR adds try/catch notation, of the form specified by #18504; namely

```perl
use feature 'try';

try {
    a_function();
}
catch ($e) {
    warn "An error occurred: $e";
}
```

When merged this is sufficient to pass the unit tests in the `Feature::Compat::Try` module at v0.02.

In addition, at some point once this feature is merged it would be good to have a big trawl through much of the documentation (e.g. starting with `eval` in `perlfunc`) to look for opportunities to draw attention to the new syntax, as well as fixing up examples currently written on `eval` (several of which I expect will contain bugs because of it).


leonerd:
Some comments to reviewers:

* I decided to rename the existing `CXp_TRYBLOCK` and `CxTRYBLOCK` macros to `EVALBLOCK` instead, as they do not relate to the true `try` blocks introduced here. Some back-compat macros exist to avoid breaking existing code, but the new forms should be preferred henceforth. They have identical semantics as before, they were just renamed.
* I added a `CXp_TRY` bit flag to the existing `CXt_EVAL` structure in the context stack, in favour of introducing a whole new type of perhaps `CXt_TRY`. I may still need to do the latter in order to fix the bug/performance issues mentioned in the above PR message, though that would lead to more code in more places being disturbed, so this path was chosen as the minimum viable to get _something_ committed sooner.
* The current implementation does not need to add any new opcodes, but I feel that for performance I will later need to add an `OP_CATCH`, or perhaps for supporting a typed dispatch feature, a few of them.
* I have not at all attempted to fix `B::Deparse` yet. The new syntax does not correctly deparse. While `B::Deparse` does create some output that looks plausible, it is not in fact correct because it hasn't noticed the `OPf_SPECIAL` flag on the `OP_ENTERTRY`. I intend to fix this in a followup commit.

pod/perlsyn.pod
```diff
@@ -282,6 +282,10 @@ The following compound statements may be used to control flow:
 
     PHASE BLOCK
 
+If enabled by the experimental C<isa> feature, the following may also be used
```
tonycoz: by the experimental 'try' feature.

tonycoz:
It looks reasonable to me, though I think the false exception object issue needs to be fixed before it is merged.

pod/perlsyn.pod
leonerd: Fixed.

leonerd:
PTAL. Many updates:

* Have now added an `OP_CATCH` which can distinguish the "exception object is false" case, and added a unit test for it.
* `pp_catch` now also handles the `my VAR = $@` logic, making it faster and less noisy in the optree
* This new opcode is also a handy signal to the still-as-yet unfixed `B::Deparse` module
* `pp_catch` now clears the `$@` variable, to ensure users don't accidentally end up relying on it. They must use the declared var instead

Will also edit the main PR description to match.

leonerd:
PS I shall definitely squash most (or all?) of the commits together before it gets merged.. most of the commit messages shouldn't really stay in the merged copy.

tonycoz:
I wonder, should using try/catch (ideally) be changing `$@` at all?  I mean like:
```
my $x = eval "..."; # some other eval
try {
  # do something with $@
}
catch ($y) {
  # do something with $@
}
```

That said, I don't really expect this to work

Grinnz:
This was discussed in the p5p thread, I believe the answer was "ideally yes but it's way more difficult for the moment"

Grinnz:
Rather, in https://github.com/Perl/perl5/issues/18504

leonerd:
@tonycoz
> I wonder, should using try/catch (ideally) be changing `$@` at all?

That's certainly a potential end goal, yes. As @Grinnz points out currently the related handling of `die` etc.. is somewhat too entangled with `ERRSV` (`$@`) to easily extract. Eventually I could imagine a `CXt_TRY` that stored an SV of its (lexical) exception variable, for `die` to write there instead. But currently that is too much adjustment of other code.

leonerd:
Squashed into fewer commits with better messages; rebased onto current blead.

pp_ctl.c
```diff
@@ -4593,6 +4614,22 @@ PP(pp_leaveeval)
     return retop;
 }
 
+PP(pp_catch)
+{
+    dTARGET;
+
+    if(!SvROK(ERRSV) && !SvTRUE(ERRSV)) {
```
ilmari, line 4621: This won't catch `die 0;`. I think it should use the same logic as [die](https://github.com/perl/perl5/blob/v5.33.6/pp_sys.c#L475), i.e. `if (!SvROK(exsv) && !(SvPV_const(exsv, len), len))`.

leonerd: Turns out that `die 0` was fine, because of the appended "... at FILE line LINE." string. But I've added a unit test to ensure that remains the case.

leonerd:
Trying to find the right balance between forward velocity, and cautiousness on new features. This has been reviewed by a few people now. I have sent it to @iabyn to review but so far there's no word at all from him - not even to acknlowledge the request
I think therefore I'll wait for any other comments from folks for a couple more days, but maybe aim that if nothing else comes up I'll hit the Merge button on Thursday evening

It'd be good to get it tried out in a few 5.33.x point releases ahead of 5.34

xsawyerx:
I don't think this should have been merged. The Perl Steering Committee is still discussing Perl 7 and waited for that discussion to be resolved before addressing new features like this one.

leonerd:
I see. Hmm - well a little late for that now.

I did announce my intentions loudly and clearly on each of of this issue in github, the perl5-porters@ mailing list, and the #p5p IRC channel on irc.perl.org. I gave well over a week of notice and got many OKs from folks all around, so I judged it to be fine.

Was either the form or duration of my notice period somehow inadequate?

wchristian:
I'd like to have an explanation of how this feature could possibly be more contentious or dangerous than anything previously scoped to Perl 7?

(That is the only reason i could imagine for comparing this with Perl 7.)

As an additional question, is that a comment intended to represent PSC or a private opinion?

shadowcat-mst:
@xsawyerx this feature was planned before you announced your original perl7 plan and in fact had had significantly wider public discussion within the community before implementation was started than the perl7 plan did.

This has broad community support and there's no reason to oppose its merging here - if anything you might consider the process that led to this merge to be a model of how do such things successfully by building broad consensus out in the open first.

xsawyerx:
I don't think this feature is contentious or problematic. I don't think the manner in which it was discussed is a problem. I realize this was planned and worked on prior to the Perl 7 announcement. Thank you all for "informing" me of this, but none of these are my issues with it.

The problem is a strategic one relating to a possible release of a Perl 7. If we release Perl 7, we leave Perl 5 with an experimental feature that will not have a stable release approving it. This might be okay and we're discussing it in the Council, but this merge forces us in this direction.

Grinnz:
There are several features currently experimental. If a plan for Perl 7 can't account for that then it will have to be improved.

Leont:
> I don't think this feature is contentious or problematic. I don't think the manner in which it was discussed is a problem. I realize this was planned and worked on prior to the Perl 7 announcement. Thank you all for "informing" me of this, but none of these are my issues with it.

I suspect you were saying what you were saying made sense in the context of a PSC discussion that none of us are part of. I think you should expect to keep being misunderstood as long as the rest of us aren't provided the full context of what you're saying.

> The problem is a strategic one relating to a possible release of a Perl 7. If we release Perl 7, we leave Perl 5 with an experimental feature that will not have a stable release approving it. This might be okay and we're discussing it in the Council, but this merge forces us in this direction.

That could easily be fixed by a sunsetting release (python learned the hard way just how important those are), I don't see how this would be a hard problem to solve.

wchristian:
> There are several features currently experimental. If a plan for Perl 7 can't account for that then it will have to be improved.

Can you make a list of those?

> you should expect to keep being misunderstood as long as the rest of us aren't provided the full context of what you're saying

I wouldn't necessarily go that far, but i agree that when making a claim without a fully grounded justification, some confusion is likely.

And agreed as well, if experimentality of things is an issue, then it seems there should be a sunsetting release with no threads unresolved.

Grinnz:
> > There are several features currently experimental. If a plan for Perl 7 can't account for that then it will have to be improved.
> 
> Can you make a list of those?

They are listed in [perlexperiment](https://github.com/Perl/perl5/blob/blead/pod/perlexperiment.pod#current-experiments) (of which I note, `try` is missing).

xsawyerx:
Good points. I guess I was hasty in my comment on this. I apologize.

PR #18511, opened 2021-01-28
win32: remove makefile.mk

xenu:
Makefile.mk is redundant with GNUmakefile. See [my p5p post](https://www.nntp.perl.org/group/perl.perl5.porters/2021/01/msg258848.html) for more details.

We planned to remove it shortly after the introduction of GNUmakefile but that slipped through the cracks for some reason: #14341


pod/perldelta.pod
```diff
@@ -338,6 +338,9 @@ versions < 3.22 now works again.  This was broken in Perl 5.31.4.
 Building with mingw.org compilers (version 3.4.5 or later) using mingw runtime
 versions >= 3.21 now works (for compilers up to version 5.3.0).
 
+Makefile.mk, and thus support for dmake, has been removed. It is still possible
+to build Perl using nmake (Makefile) and GNU make (GNUmakefile).
+
```
smpeters, line 344: "to build Perl on Windows" might be a bit more clear here

smpeters (commented):
I only have one minor suggested clarification for perldelta.pod

pod/perldelta.pod
xenu: This entry is in the "Windows" section, but I guess additional clarification won't hurt.

Issue #18470, opened 2021-01-12
open pragma with :std and :crlf


mik2:
Module: open

**Description**
```
open ":std", IO => ":crlf"
```
does not affect STDIN or STDOUT, which is in contradiction with the following statement in the documentation:

> :std is not a layer but an additional subpragma. When specified in the import list, it activates an additional functionality of pushing the layers selected for input/output handles to the standard filehandles (STDIN, STDOUT, STDERR)

**Steps to Reproduce**
```
printf "\r\n\n" | perl -Mopen=:std,IO,:crlf -e 'print <>;' | od -c
0000000  \r  \n  \n
0000003
```

**Expected behavior**
the output should be the same as for:
```
printf "\r\n\n" | perl -e 'binmode $_,":crlf" for STDIN, STDOUT; print <>;' | od -c
0000000  \r  \n  \r  \n
0000004
```

**Perl configuration**
```
Summary of my perl5 (revision 5 version 32 subversion 0) configuration:

  Platform:
    osname=linux
    osvers=5.4.43-1-lts
    archname=x86_64-linux-thread-multi
    uname='linux build-edge-x86_64 5.4.43-1-lts #2-alpine smp thu, 28 may 2020 20:13:48 utc x86_64 linux '
    config_args='-des -Dcccdlflags=-fPIC -Dcccdlflags=-fPIC -Dccdlflags=-rdynamic -Dprefix=/usr -Dprivlib=/usr/share/perl5/core_perl -Darchlib=/usr/lib/perl5/core_perl -Dvendorprefix=/usr -Dvendorlib=/usr/share/perl5/vendor_perl -Dvendorarch=/usr/lib/perl5/vendor_perl -Dsiteprefix=/usr/local -Dsitelib=/usr/local/share/perl5/site_perl -Dsitearch=/usr/local/lib/perl5/site_perl -Dlocincpth=  -Doptimize=-Os -fomit-frame-pointer -Duselargefiles -Dusethreads -Duseshrplib -Dd_semctl_semun -Dman1dir=/usr/share/man/man1 -Dman3dir=/usr/share/man/man3 -Dinstallman1dir=/usr/share/man/man1 -Dinstallman3dir=/usr/share/man/man3 -Dman1ext=1 -Dman3ext=3pm -Dcf_by=Alpine -Ud_csh -Dusenm'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
[...]

PR #18516, opened 2021-01-30
Make use open :std push layers without any munging #18470

Leont:
Previously it would only apply the first layer that was either :utf8 or any other layer with and argument (e.g. :encoding(utf8)). Now it will just take the whole argument and apply that. This should solve #18470

### DIFF
--- a/lib/open.pm
+++ b/lib/open.pm
@@ -1,7 +1,7 @@
 package open;
 use warnings;
 
-our $VERSION = '1.12';
+our $VERSION = '1.13';
 
 require 5.008001; # for PerlIO::get_layers()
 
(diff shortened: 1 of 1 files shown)


Grinnz:
Suggest briefly mentioning this at the end of the DESCRIPTION section, i.e.:

```
Before Perl 5.34, C<:std> would only apply the first layer provided that is
either C<:utf8> or has a layer argument, e.g. C<:encoding(UTF-8)>. Since
Perl 5.34 it will apply the same layer stack it provides to C<${^OPEN}>.
```

Grinnz (approved):

khwilliamson:
@leont, any reason to not merge this now?

Leont:
> @Leont, any reason to not merge this now?

No, I had just entirely forgotten about this.

mik2:
The change is absent in 5.34, so the DESCRIPTION should rather mention 5.35 or 5.35.1.

Leont:
> The change is absent in 5.34, so the DESCRIPTION should rather mention 5.35 or 5.35.1.

Fair point

PR #18522, opened 2021-02-03
files in t/opbasic shouldn't load test.pl

tonycoz:
since they're the most basic of tests, it should avoid loading
any modules.

It currently loads Config (conditionally), but there's no way to avoid
that at this point.

### DIFF
--- a/t/opbasic/arith.t
+++ b/t/opbasic/arith.t
@@ -2,8 +2,12 @@
 
 BEGIN {
     chdir 't' if -d 't';
-    require './test.pl';
-    set_up_inc('../lib');
+    # we can't use test.pl here (see the comment below)
+    #require './test.pl';
+    #set_up_inc('../lib');
+    # so inline set_up_inc() and its call to is_miniperl
+    @INC = () if defined &DynaLoader::boot_DynaLoader;
+    unshift @INC, "../lib";
 }
 
 # This file has been placed in t/opbasic to indicate that it should not use


jkeenan:
> since they're the most basic of tests, it should avoid loading
> any modules.
> 
> It currently loads Config (conditionally), but there's no way to avoid
> that at this point.

+1

jkeenan:
> since they're the most basic of tests, it should avoid loading
> any modules.
> 
> It currently loads Config (conditionally), but there's no way to avoid
> that at this point.

Question:  Is this pull request intended to supersede https://github.com/Perl/perl5/pull/18519?

If so, then let's merge this one and thank the OP in 18519 for his identification of the problem.

Thank you very much.
Jim Keenan

tonycoz:
It fixes a different problem to #18519 and does not replace it.

t/opbasic/arith.t
```diff
@@ -2,8 +2,12 @@
 
 BEGIN {
     chdir 't' if -d 't';
-    require './test.pl';
-    set_up_inc('../lib');
+    # we can't use test.pl here (see the comment below)
+    #require './test.pl';
+    #set_up_inc('../lib');
+    # so inline set_up_inc() and its call to is_miniperl
+    @INC = () if defined DynaLoader::boot_DynaLoader;
```
t-a-k: Excuse me for cutting in, but I think `&` is missing here:
```suggestion
    @INC = () if defined &DynaLoader::boot_DynaLoader;
```

tonycoz: Thanks, fixed.

PR #18523, opened 2021-02-03
document how to get a PR as a local branch

tonycoz:


karenetheridge:
You can also fetch all pull requests by adding this to your .git/config under `[origin]`:

            fetch = +refs/pull/*/head:refs/remotes/origin/pull/*

..that is, when you `git fetch origin`, you'll get a bunch of `origin/pull/12345` remote branches pulled down, so you can then do `git checkout pull/12345`, rebase it, merge it to blead manually, etc etc.

tonycoz:
> ..that is, when you `git fetch origin`, you'll get a bunch of `origin/pull/12345` remote branches pulled down, so you can then do `git checkout pull/12345`, rebase it, merge it to blead manually, etc etc.

I was aware of this, but don't use it.  I'll add a mention.

smpeters:
Thank you for this.

PR #18526, opened 2021-02-03
pp_split: calculate flags just once and other minor tweaks

richardleach:
The commits in this PR:
* calculate the flags for `newSVpvn_flags` just once, rather than at every call.
* make `realarray` a bool, rather than an I32.
* move the definition of `multiline` to the only branch that actually uses it.

These changes result in a smaller _pp.o_, although not a smaller _miniperl_/_perl_ (with gcc at -O2).
Simple benchmarks on x64 Linux don't show a change to run time performance.


tonycoz (approved):
Looks good to me

PR #18534, opened 2021-02-07
Upstream cygwin patches

Leont:
Cygwin is currently applying these patches to their perl. I suspect we should apply them ourselves instead.

Note: I didn't write any of these, and can't vouch for them or test them.

### DIFF
--- a/Configure
+++ b/Configure
@@ -5266,6 +5266,16 @@ for thislib in $libswanted; do
 		$test -f "$xxx" && libstyle=shared
 	    fi
 	    if test ! -f "$xxx"; then
+		xxx=$thisdir/lib$thislib.$so$_a
+		$test -f "$xxx" && eval $libscheck
+		$test -f "$xxx" && libstyle="import"
+	    fi
+	    if test ! -f "$xxx"; then
+		xxx=$thisdir/$thislib.$so$_a
+		$test -f "$xxx" && eval $libscheck
+		$test -f "$xxx" && libstyle="import"
+	    fi
+	    if test ! -f "$xxx"; then
 		xxx=$thisdir/lib$thislib.$so
 	        $test -f "$xxx" && eval $libscheck
 		$test -f "$xxx" && libstyle=shared
--- a/README.cygwin
+++ b/README.cygwin
@@ -31,7 +31,7 @@ L<https://www.cygwin.com/>
 
 A recent net or commercial release of Cygwin is required.
 
-At the time this document was last updated, Cygwin 1.7.16 was current.
+At the time this document was last updated, Cygwin 3.0.7 was current.
 
 
 =head2 Cygwin Configuration
@@ -41,8 +41,8 @@ that Perl builds cleanly.  These changes are B<not> required for normal
 Perl usage.
 
 B<NOTE:> The binaries that are built will run on all Win32 versions.
-They do not depend on your host system (WinXP/Win2K/Win7) or your
-Cygwin configuration (binary/text mounts, cvgserver).
+They do not depend on your host system or your
+Cygwin configuration (binary/text mounts, cygserver).
 The only dependencies come from hard-coded pathnames like F</usr/local>.
 However, your host system and Cygwin configuration will affect Perl's
 runtime behavior (see L</"TEST">).
@@ -128,11 +128,10 @@ NOTE: The BerkeleyDB library only completely works on NTFS partitions.
 A port of SysV IPC is available for Cygwin.
 
 NOTE: This has B<not> been extensively tested.  In particular,
-C<d_semctl_semun> is undefined because it fails a Configure test
-and on Win9x the I<shm*()> functions seem to hang.  It also creates
-a compile time dependency because F<perl.h> includes F<<sys/ipc.h>>
-and F<<sys/sem.h>> (which will be required in the future when compiling
-CPAN modules). CURRENTLY NOT SUPPORTED!
+C<d_semctl_semun> is undefined because it fails a Configure test.  It
+also creates a compile time dependency because F<perl.h> includes
+F<<sys/ipc.h>> and F<<sys/sem.h>> (which will be required in the
+future when compiling CPAN modules). CURRENTLY NOT SUPPORTED!
 
 =item * C<-lutil>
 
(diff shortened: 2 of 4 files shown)


Leont:
The auto-image, hints, libsearch and readme patches look non-controversial to me.

The libpth patch had me confused for a moment. It's actually fixing a bug that exists on all platforms that use `Configure`, where include paths that don't end on `/include` will erroneously end up in `libpth` (e.g. `/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/include-fixed` in my particular case). I would have written it differently but it seems correct (and necessary on other platforms as well).

The Win32 patch breaks on Windows, in particular `Unicode.t`. I think we should just leave it out.

hints/cygwin.sh
```diff
@@ -25,12 +25,21 @@ libswanted=`echo " $libswanted " | sed -e 's/ c / /g'`
...
 # - eliminate -lutil, symbols are all in libcygwin.a
 libswanted=`echo " $libswanted " | sed -e 's/ util / /g'`
+ignore_versioned_solibs='y'
+usenm='no'
+libc='/usr/lib/libcygwin.a'
+loclibpth=' '
+glibpth=' '
+plibpth=' '
+libpth=' '
+PATH='.:/usr/bin/'
```
tonycoz: In general I don't think hints should just blindly define things, they should only define them if a value isn't already supplied.

Otherwise `./Configure -Dwhatever` where `whatever` is blindly overridden can get very confusing.

I'm guessing PATH is being set to influence the value set for LDLIBPTH when Makefile.SH is run, since ldlibpthname=PATH.  Of course this is lost if Makefile needs to be rebuilt.

cpan/Win32/t/GetShortPathName.t
```diff
@@ -12,18 +12,21 @@ BEGIN {
...
 Win32::CreateFile($path);
 ok(-f $path);
 
 my $short = Win32::GetShortPathName($path);
-ok($short, qr/^\S{1,8}(\.\S{1,3})?$/);
+ok($short, qr/^C:\\\S{1,8}(\.\S{1,3})?$/);
 ok(-f $short);
+my $long = Win32::GetLongPathName($short);
+ok($long, $path);
+ok(-f $long);
```
tonycoz: The user might not have write access to `c:\` (I ran a smoker like that way back)

Configure
```diff
@@ -4865,18 +4865,21 @@ esac
...
+	   for i in $*; do
+	       j="`$echo $i|$sed 's,/include$,/lib,'`"
+	       if $test -d $j; then
+		   case " $libpth " in
+		   *" $j "*) ;;
+		   *) libpth="$libpth $j";;
+		   esac
+	       fi
+	   done
+	   libpth="`$echo $libpth|$sed 's/^ //'`"
```
Tux: Did the leading whitespace also change? I find this a weird diff and hard to spot the diff and it also looks like the new indentation is wrong

If `include-fixed` is still present here, it should be catered for? (as here the `/include$` is anchored

Configure
```diff
@@ -4865,18 +4865,21 @@ esac
 case "$ccname" in
     gcc)
 	$echo 'extern int foo;' > try.c
-	set X `$cppstdin -v try.c 2>&1 | $awk '/^#include </,/^End of search /'|$cppfilter $grep '/include'`
+	set X `$cppstdin -v try.c 2>&1 | $awk '/^#include </,/^End of search /'|$cppfilter $grep '/include$'`
```
Tux: As a short test on my local linux boxes shows, this would change behavior:
```
$ gcc  -E -v try.c 2>&1 | /usr/bin/awk '/^#include </,/^End of search /'
#include <...> search starts here:
 /usr/lib64/gcc/x86_64-suse-linux/7/include
 /usr/local/include
 /usr/lib64/gcc/x86_64-suse-linux/7/include-fixed
 /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/include
 /usr/include
End of search list.
```
As now `/usr/lib64/gcc/x86_64-suse-linux/7/include-fixed` would be skipped.

Likewise on HP-UX
```
 /wrk/ia64-4.7.2/bin/../lib/gcc/ia64-hp-hpux11.31/4.7.2/include
 /wrk/ia64-4.7.2/bin/../lib/gcc/ia64-hp-hpux11.31/4.7.2/include-fixed
 /usr/local/ia64/include
 /usr/include
```

What is the reason for adding the `$` EOL anchor? Is the **reason** to skip `include-fixed`?

Tux (commented):
Looks ok

hints/cygwin.sh
Leont: > In general I don't think hints should just blindly define things, they should only define them if a value isn't already supplied.

Several of these are also defined for other platforms this way. That said, adding «test -z "$foo" &&» like what's already done below may be useful for some of these

Configure
Leont: > Did the leading whitespace also change? I find this a weird diff and hard to spot the diff and it also looks like the new indentation is wrong

Yeah, fixed that in the other commits, forgot about this one.

> If include-fixed is still present here, it should be catered for? (as here the /include$ is anchored

Yeah, that must be handled somehow, though I'm also not entirely fond of how that's done here.

xenu:
>The Win32 patch breaks on Windows, in particular Unicode.t. I think we should just leave it out.

Yeah, the Win32 part should be skipped. I don't understand the purpose of those changes and it seems leaving them out doesn't break the tests on cygwin.

hints/cygwin.sh
```diff
@@ -62,13 +71,23 @@ case "$osvers" in
         d_inetpton='undef'
 esac
 
+case "$osvers" in
+    2.[0-4].*|1.*)
+        # newlib finitel is buggy before cygwin-2.5.0
+        d_finitel='undef'
+        ;;
+esac
```
xenu: Cygwin is one of those projects that you have to have a very good reason to use a non-latest version. They don't even make them available for download. IMO this part should be skipped.

Configure
```diff
@@ -330,6 +330,7 @@ _o=''
 archobjs=''
 exe_ext=''
 firstmakefile=''
+libprefix='lib'
```
Leont: I think this should be moved to the hints file instead.

Leont:
I cut out the pieces that were controversial or seemed dubious, I think this can be merged in early 5.37

hints/cygwin.sh
```diff
@@ -62,13 +71,23 @@ case "$osvers" in
...
 # compile Win32CORE "module" as static. try to avoid the space.
 if test -z "$static_ext"; then
   static_ext="Win32CORE"
 else
   static_ext="$static_ext Win32CORE"
 fi
 
+# configury should not check for xlocale.h if it gets the API from locale.h
+i_xlocale='undef'
+
```
tonycoz: I think this is obsolete since 68f1b6d70

Leont: Removed it

hints/cygwin.sh
Leont: That sounds reasonable

README.cygwin
```diff
@@ -31,7 +31,7 @@ L<https://www.cygwin.com/>
 
 A recent net or commercial release of Cygwin is required.
 
-At the time this document was last updated, Cygwin 1.7.16 was current.
+At the time this document was last updated, Cygwin 3.0.7 was current.
```
tonycoz, line 34:
```
$ uname -a
CYGWIN_NT-10.0 GANYMEDE 3.3.5(0.341/5/3) 2022-05-13 12:27 x86_64 Cygwin
```

tonycoz (approved):

Issue #18538, opened 2021-02-07
t/op/try.t: Test case where experimental warnings have not been turned off


jkeenan:
In `blead`, all tests in `t/op/try.t` currently run under:
```
no warnings 'experimental::try';
```
That's all well and good -- except that we ought to have at least one test where we demonstrate that a warning is generated and where we capture the warning and verify its content against our expectation thereof.

@leonerd and I have been discussing a patch off-list.  A pull request will come shortly.

Thank you very much.
Jim Keenan

leonerd:
Fix is merged; I think we can close this.

PR #18539, opened 2021-02-07
Test for "experimental::try" warnings being active

jkeenan:
For: https://github.com/Perl/perl5/issues/18538

### DIFF
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -10,10 +10,36 @@ BEGIN {
 use strict;
 use warnings;
 use feature 'try';
+
+
+{
+    my $warnings;
+    BEGIN { $SIG{__WARN__} = sub { $warnings .= shift; }; }
+
+    my $x;
+    my ($ltry, $lcatch) = (__LINE__+1, __LINE__+4);
+    try {
+       $x .= "try";
+    }
+    catch ($e) {
+       $x .= "catch";
+    }
+    is($x, "try", 'successful try/catch runs try but not catch');
+
+    is($warnings, "try/catch is experimental at $0 line $ltry.\n" .
+                  "try/catch is experimental at $0 line $lcatch.\n",
+        'compiletime warnings');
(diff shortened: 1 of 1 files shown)


leonerd:
There's a lot more changed lines to this test than I expected there would be.. you've rearranged and indented almost every part of it. Is that necessary?

jkeenan:
> There's a lot more changed lines to this test than I expected there would be.. you've rearranged and indented almost every part of it. Is that necessary?

Was pasting it into the branch I created yesterday.  Please try 23e4ff743e in that branch.  Diff should now be:
```
$ git diff blead..jkeenan/t-op-try-use-warnings-20210206 -- t/op/try.t
diff --git a/t/op/try.t b/t/op/try.t
index cfc5d82b6b..044dfb6cb7 100644
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -10,6 +10,28 @@ BEGIN {
 use strict;
 use warnings;
 use feature 'try';
+
+{
+    my $warnings;
+    BEGIN { $SIG{__WARN__} = sub { $warnings .= shift; }; }
+
+    my $x;
+    my ($ltry, $lcatch) = (__LINE__+1, __LINE__+4);
+    try {
+       $x .= "try";
+    }
+    catch ($e) {
+       $x .= "catch";
+    }
+    is($x, "try", 'successful try/catch runs try but not catch');
+
+    is($warnings, "try/catch is experimental at $0 line $ltry.\n" .
+                  "try/catch is experimental at $0 line $lcatch.\n",
+        'compiletime warnings');
+    BEGIN { undef $SIG{__WARN__}; }
+}
+
+
 
 {
```
Thank you very much.
Jim Keenan

leonerd:
Much better :)
I'll let you off with the inconsistent indentation because I know you just copied it from the email example I sent which was also wrong ;) I'll fix that up when I merge.

PR #18546, opened 2021-02-08
Jkeenan/pod html convert n test 20210208

jkeenan:
In this pull request I propose to change the interface to a function used inside the test suite for `ext/Pod-Html`.

Function `convert_n_test` currently takes a list of any number of arguments, the first of which is the stem of the basename of a given test file in `t/` and the second of which is the test description.  All remaining arguments are strings in the format `--XXX` where `XXX` is a switch for `Pod::Html::pod2html()`, the principal function within that module and the essence of the command-line utility `pod2html`.

IMO, the fact that `Pod::Html::pod2html()` takes a list of strings which look like command-line options is a bug, not a feature.  However, the ship that bug sailed on left a quarter-century ago.  I'm not proposing to change that function's interface at this date.  But I don't think we should be afraid of changing the interface to testing function `convert_n_test()` to something more rational and Perlish, *i.e.,* a hash reference.

This pull request consolidates all the arguments to `convert_n_test()` other than the first two into a hash reference.  We then parse these arguments into a format acceptable to an internal call to `Pod::Html::pod2html()`.  We adapt all existing files in the module's test suite to this new interface.

I believe that this new interface will make debugging `pod2html` calls easier and will facilitate adding more test files to the test suite.  We will more easily be able to dump the arguments which `convert_n_test()` is passing to `pod2html` before we make that function call.

And, take it from me, we *do* need more tests of Pod-Html.  This [link](https://github.com/Perl/perl5/issues?q=is%3Aissue+label%3Adist%2FPod-Html+is%3Aopen+sort%3Acreated-asc) shows the open issues bearing the 'dist/Pod-Html' link.  Some of these issues date back to 2012 and were some of the first RT tickets I worked on.  The 'dist/Pod-Html` label is a replacement for the META ticket in the old RT system whose name, per @rjbs, was "Pod-Html is a mess."

It's less of a mess now, mainly due to work by Marc Green, Zefram and others.  But it's still messy, and I would like to make the test suite a bit more rational before working on tickets like those in the label URL above.  Please review.

Thank you very much.
Jim Keenan

### DIFF
--- a/ext/Pod-Html/t/pod2html-lib.pl
+++ b/ext/Pod-Html/t/pod2html-lib.pl
@@ -29,7 +29,11 @@ sub rem_test_dir {
 }
 
 sub convert_n_test {
-    my($podfile, $testname, @p2h_args) = @_;
+    my($podfile, $testname, $p2h_args_ref) = @_;
+    if (defined $p2h_args_ref) {
+        die "3rd argument must be hashref"
+            unless ref($p2h_args_ref) eq 'HASH'; # TEST ME
+    }
 
     my $cwd = Pod::Html::_unixify( Cwd::cwd() );
     my ($vol, $dir) = splitpath($cwd, 1);
@@ -41,20 +45,40 @@ sub convert_n_test {
     my $infile   = catpath $vol, $new_dir, "$podfile.pod";
     my $outfile  = catpath $vol, $new_dir, "$podfile.html";
 
-    my %default_args_table = (
+    my %args_table = (
         infile      =>    $infile,
         outfile     =>    $outfile,
         podpath     =>    't',
         htmlroot    =>    '/',
         podroot     =>    $cwd,
     );
-    my @default_args_list = map { "--${_}=" . $default_args_table{$_} }
-        keys %default_args_table;
-    # To add/modify args to p2h, use @p2h_args
-    Pod::Html::pod2html(
-        @default_args_list,
-        @p2h_args,
-    );
+    my %no_arg_switches = map { substr($_,2) => 1 } ( qw|
+           --flush --recurse --norecurse
+           --quiet --noquiet --verbose --noverbose
+           --index --noindex --backlink --nobacklink
+           --header --noheader --poderrors --nopoderrors
+    | );
+    if (defined $p2h_args_ref) {
+        for my $sw (keys %{$p2h_args_ref}) {
+            if ($no_arg_switches{$sw}) {
+                $args_table{$sw} = undef;
+            }
+            else {
+                $args_table{$sw} = $p2h_args_ref->{$sw};
+            }
+        }
+    }
+    my @args_list = ();
+    for my $k (keys %args_table) {
+        if (defined $args_table{$k}) {
+            push @args_list, "--" . $k . "=" . $args_table{$k};
+        }
+        else {
+            push @args_list, "--" . $k;
+        }
+    }
+
+    Pod::Html::pod2html( @args_list );
 
     $cwd =~ s|\/$||;
 
--- a/ext/Pod-Html/t/pod2html-lib.pl
+++ b/ext/Pod-Html/t/pod2html-lib.pl
@@ -41,13 +41,18 @@ sub convert_n_test {
     my $infile   = catpath $vol, $new_dir, "$podfile.pod";
     my $outfile  = catpath $vol, $new_dir, "$podfile.html";
 
+    my %default_args_table = (
+        infile      =>    $infile,
+        outfile     =>    $outfile,
+        podpath     =>    't',
+        htmlroot    =>    '/',
+        podroot     =>    $cwd,
+    );
+    my @default_args_list = map { "--${_}=" . $default_args_table{$_} }
+        keys %default_args_table;
     # To add/modify args to p2h, use @p2h_args
     Pod::Html::pod2html(
-        "--infile=$infile",
-        "--outfile=$outfile",
-        "--podpath=t",
-        "--htmlroot=/",
-        "--podroot=$cwd",
+        @default_args_list,
         @p2h_args,
     );
 
--- a/ext/Pod-Html/t/pod2html-lib.pl
+++ b/ext/Pod-Html/t/pod2html-lib.pl
@@ -52,11 +52,11 @@ sub convert_n_test {
         htmlroot    =>    '/',
         podroot     =>    $cwd,
     );
-    my %no_arg_switches = map { substr($_,2) => 1 } ( qw|
-           --flush --recurse --norecurse
-           --quiet --noquiet --verbose --noverbose
-           --index --noindex --backlink --nobacklink
-           --header --noheader --poderrors --nopoderrors
+    my %no_arg_switches = map { $_ => 1 } ( qw|
+           flush recurse norecurse
+           quiet noquiet verbose noverbose
+           index noindex backlink nobacklink
+           header noheader poderrors nopoderrors
     | );
     if (defined $p2h_args_ref) {
         for my $sw (keys %{$p2h_args_ref}) {
--- a/ext/Pod-Html/t/pod2html-lib.pl
+++ b/ext/Pod-Html/t/pod2html-lib.pl
@@ -62,8 +62,7 @@ sub convert_n_test {
         for my $sw (keys %{$p2h_args_ref}) {
             if ($no_arg_switches{$sw}) {
                 $args_table{$sw} = undef;
-            }
-            else {
+            } else {
                 $args_table{$sw} = $p2h_args_ref->{$sw};
             }
         }
@@ -72,8 +71,7 @@ sub convert_n_test {
     for my $k (keys %args_table) {
         if (defined $args_table{$k}) {
             push @args_list, "--" . $k . "=" . $args_table{$k};
-        }
-        else {
+        } else {
             push @args_list, "--" . $k;
         }
     }
--- a/ext/Pod-Html/t/crossref.t
+++ b/ext/Pod-Html/t/crossref.t
@@ -22,12 +22,11 @@ SKIP: {
     shift @dirs if $dirs[0] eq '';
     my $relcwd = join '/', @dirs;
         
-    convert_n_test("crossref", "cross references", 
-     "--podpath=". File::Spec::Unix->catdir($relcwd, 't') . ":"
-                 . File::Spec::Unix->catdir($relcwd, 'testdir/test.lib'),
-     "--podroot=". catpath($v, '/', ''),
-     "--quiet",
-    );
+    convert_n_test("crossref", "cross references", {
+        podpath => File::Spec::Unix->catdir($relcwd, 't') . ":" . File::Spec::Unix->catdir($relcwd, 'testdir/test.lib'),
+        podroot => catpath($v, '/', ''),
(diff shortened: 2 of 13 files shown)


rjbs:
https://youtu.be/UneS2Uwc6xw?t=58

ext/Pod-Html/t/pod2html-lib.pl
```diff
@@ -41,20 +45,40 @@ sub convert_n_test {
...
-    Pod::Html::pod2html(
-        @default_args_list,
-        @p2h_args,
-    );
+    my %no_arg_switches = map { substr($_,2) => 1 } ( qw|
+           --flush --recurse --norecurse
+           --quiet --noquiet --verbose --noverbose
+           --index --noindex --backlink --nobacklink
+           --header --noheader --poderrors --nopoderrors
+    | );
```
rjbs, line 60: Apropos of nothing, I am surprised by this `map { substr($_,2) => 1 }` form, when you could, I think, just drop the `--` from the source entirely.  It's not a problem, but seems like a small amount more code than is needed.

ext/Pod-Html/t/pod2html-lib.pl
```diff
@@ -41,20 +45,40 @@ sub convert_n_test {
...
+    if (defined $p2h_args_ref) {
+        for my $sw (keys %{$p2h_args_ref}) {
```
rjbs, line 62: If I was supreme monarch, we would use postfix deref for all non-dual-life code.  But I am willing to admit that I am not…

ext/Pod-Html/t/pod2html-lib.pl
```diff
@@ -41,20 +45,40 @@ sub convert_n_test {
...
+            if ($no_arg_switches{$sw}) {
+                $args_table{$sw} = undef;
+            }
+            else {
```
rjbs: While the code is not extremely consistent, in general the pod2html code appears to cuddle elses, and I think we should _try_ to avoid introducing inconsistency.

ext/Pod-Html/t/pod2html-lib.pl
```diff
@@ -41,20 +45,40 @@ sub convert_n_test {
...
+    my @args_list = ();
+    for my $k (keys %args_table) {
+        if (defined $args_table{$k}) {
+            push @args_list, "--" . $k . "=" . $args_table{$k};
+        }
+        else {
+            push @args_list, "--" . $k;
+        }
+    }
+
```
rjbs, line 78: Here is perhaps my only _substantive_ comment:

You have `%no_arg_switches`, which is used to replace any given value for an option in the p2h_args_ref.  Then you build up `@args_list` by treating an undef value as `--foo` rather than `--foo=bar`.  This seems like it is likely to be correct in all existing cases (but I didn't verify), *but* it seems a bit backward to me to do this as opposed to checking it very late.  p2h_args_ref would use TRUE to mean "include this no arg switch" and FALSE to exclude it.  This would make it possible for the defaults, later, to have a `--foo` switch that could be removed.  Then, in the loop at line 72, instead of including an argument based on the definedness of the value in args_table, you'd include an argument based on whether the key is in the `no_arg_switches` lookup.

This isn't a big deal.  I think the code is correct, now.  I think it could later impede further refactoring or updating of this code, just a little, but also that is not a very likely event to plan for.

So, if you think it's worth making the change I suggest:  now you have heard the suggestion and can do so.  If you don't, I think the code as presented is likely fine.  Also, we can refactor more later, if needed.

rjbs (approved):
I provided a few comments, some noise, some stylistic, and one "consider whether you want to make this change."  I think this can be applied as is.

**I did not run the test suite, and am trusting that this code works.**

ext/Pod-Html/t/pod2html-lib.pl
jkeenan: > Apropos of nothing, I am surprised by this `map { substr($_,2) => 1 }` form, when you could, I think, just drop the `--` from the source entirely. It's not a problem, but seems like a small amount more code than is needed.

@rjbs, thanks for taking a look at this.  I'm a bit unclear as to what you're referring to as "the source" here.  Are you referring just to the contents of `%no_arg_switches` here?  Or are you referring to the strings passed as arguments to `Pod::Html::pod2html()`?

Thank you very much.
Jim Keenan

ext/Pod-Html/t/pod2html-lib.pl
jkeenan: > If I was supreme monarch, we would use postfix deref for all non-dual-life code. But I am willing to admit that I am not…

As has so often been the case, because the most advanced version of `perl` that I was permitted to use (while employed) or chose to use (in CPAN modules) was only 5.14, I've never used or learned postfix deref.  :-(

jkeenan:
> I provided a few comments, some noise, some stylistic, and one "consider whether you want to make this change." I think this can be applied as is.
> 
> **I did not run the test suite, and am trusting that this code works.**

The code works ... but I'm going to mull over your comments on `convert_n_test()` a bit more.  I know that I had a very specific reason for writing the code the way I did, but it will take me a while to remember that.

Thank you very much.
Jim Keenan

ext/Pod-Html/t/pod2html-lib.pl
rjbs: What I meant was "remove the double dashes from the strings in the qw(...) operator, and then you won't need to remove them in the map."  To me, this reads like "well, earlier on I was using `$_` itself also in the map, but then I removed it, leaving the code with more complexity than required.

As I said, this isn't really a problem, but feels like a bit of a vestigial appendix that could be excised.

ext/Pod-Html/t/pod2html-lib.pl
jkeenan: > You have `%no_arg_switches`, which is used to replace any given value for an option in the p2h_args_ref. Then you build up `@args_list` by treating an undef value as `--foo` rather than `--foo=bar`. This seems like it is likely to be correct in all existing cases (but I didn't verify), _but_ it seems a bit backward to me to do this as opposed to checking it very late. p2h_args_ref would use TRUE to mean "include this no arg switch" and FALSE to exclude it. This would make it possible for the defaults, later, to have a `--foo` switch that could be removed. Then, in the loop at line 72, instead of including an argument based on the definedness of the value in args_table, you'd include an argument based on whether the key is in the `no_arg_switches` lookup.
> [snip]

My objectives in this refactoring were and are:

* To make `convert_n_test()`'s interface somewhat more sane, by making its third argument a hashref and by keying that hashref on ordinary strings rather than strings that look like command-line switches.

* To make the list of arguments we pass to `Pod::Html::pod2html()` very debuggable.

I believe the first objective has been met.

For the second objective what I want is to be able to debug any call to `convert_n_test()` like this:
```
# %args_table previously populated
print STDERR Dumper(\%args_table);

    my @args_list = ();
    for my $k (keys %args_table) {
        if (defined $args_table{$k}) {
            push @args_list, "--" . $k . "=" . $args_table{$k};
        } else {
            push @args_list, "--" . $k;
        }
    }

print STDERR Dumper(\@args_list);

    Pod::Html::pod2html( @args_list );
```
That is, I want to be able to dump the data being passed to `pod2html()` in either of two ways:  in the form of a lookup table, and/or in the form of the actual list of 'switch-like' strings being bassed to `pod2html()`.  From your verbal description alone I don't see how your suggestion satisfies those objectives.  Could you provide some code?

Thank you very much.
Jim Keenan

rjbs:
```
warn Dumper(\%args) if $DEBUGGING;

my @arg_list;
for my $k (keys %arg) {
  if ($is_bool_option{$k}) {
    push @arg_list, "--$k";
  } else {
    push @arg_list, "--$k=$arg{$k}";
  }
}

warn Dumper(\@arg_list) if $DEBUGGING;

Pod::Html::pod2html(@arg_list);
```

jkeenan: > ```
> warn Dumper(\%args) if $DEBUGGING;
> 
> my @arg_list;
> for my $k (keys %arg) {
>   if ($is_bool_option{$k}) {
>     push @arg_list, "--$k";
>   } else {
>     push @arg_list, "--$k=$arg{$k}";
>   }
> }
> 
> warn Dumper(\@arg_list) if $DEBUGGING;
> 
> Pod::Html::pod2html(@arg_list);
> ```

@rjbs, to the extent that I understand your proposed alternative -- and I grant that I may not -- I disagree with it.

First, you've renamed some of the variables, which tends to obscure the real changes you're suggesting.  Can I assume that your `%is_bool_option` is the same as my `no_arg_switches`?

Can I assume that your `@arg_list` is the same as my `@args_list`?

And now, most crucially, you are using `%args`.  Is it the same as my `%args_table`?  I don't think so.

The point in `%args_table` is to clearly differentiate the arguments being passed to `convert_n_test()` from those ultimately passed to `Pod::Html::pod2html()` inside `convert_n_test()`.  You haven't show how you would populate `%args`, but I suspect it would blur that distinction.

Currently (*i.e.,* in blead), `convert_n_test()` is structured like this:
```
sub convert_n_test {
    my($podfile, $testname, @p2h_args) = @_;
    # where @p2h_args is a list of strings which are written as if they were command-line options
    ...
    
    # define default values
    ...
    
        Pod::Html::pod2html(
        "--infile=$infile",
        "--outfile=$outfile",
        "--podpath=t",
        "--htmlroot=/",
        "--podroot=$cwd",
        @p2h_args,
    );
```
`Pod::Html::pod2html()` takes a list of strings which are written as if they were command-line options.  I think that's a bad interface design.  It should have been a hash or hashref -- and command-line utility `pod2html` should have used `Getopt::Long` internally to create that hash.  But that horse left the barn in the last century.

What really aggravates me is that `convert_n_test()` in blead mushes together (a) default values which it sets internally and (b) arguments passed to `convert_n_test()`.  My refactoring clearly distinguishes the two.  Since I don't know how you propose to populate `%args`, I don't understand how you can maintain that distinction.

ext/Pod-Html/t/pod2html-lib.pl
jkeenan: Addressed in commit 37ad10f992.

ext/Pod-Html/t/pod2html-lib.pl
rjbs: Yes, I think %arg was the same as %arg_table.  I just tried to bang out sufficiently demonstrative code to show what I meant about passing a value to a `--switch` based on something other than the definedness of the hash entry, and to get it done before falling asleep last night.

You're storing "undef" to mean that there's no value.  My suggestion was that instead, I would stick to using a separate list of "this switch is only present or not present".  That way you'd never have a bug where the "infile" option ended up undef, and so `--infile` was passed as a switch with no argument.  Or you could notice an error where some no-arg switch (like `--quiet`) had gotten its value set to 2.  There are some benefits of this which I alluded to.

If you don't agree, it's fine, I think your code is fine to apply as-is (as I said initially).

jkeenan:
Branch merged so that we can work on other Pod-Html-related tickets with benefit of more tractable interface to `convert_n_test()`.  We can consider p.r.s for the internals of that function later.

Thank you very much.
Jim Keenan

PR #18551, opened 2021-02-09
Save and restore errno around DEBUG statements

khwilliamson:
Adding DEBUG statements should not affect errno.  But it could.  I have
in the past wrapped them in save/restores when it's become a problem.
But that is effort that could have been conserved by just always doing
it automatically.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -4484,7 +4484,7 @@ Gid_t getegid (void);
 
 #  define DEBUG__(t, a)                                                 \
         STMT_START {                                                    \
-                if (t) STMT_START {a;} STMT_END;                        \
+          if (t) STMT_START { dSAVE_ERRNO; a; RESTORE_ERRNO;} STMT_END; \
         } STMT_END
 
 #  define DEBUG_f(a) DEBUG__(DEBUG_f_TEST, a)


khwilliamson:
I talked to @tonycoz about this some time ago, and my recollection was that he didn't think this could be done without slowing down the likely case where debugging isn't turned on.  I have now looked at the preprocessed files, and it looks to me like it gets executed only if the check succeeds that this particular debugging flag is set.  Here is a very simple example from regcomp.c.
The source line is:

  `  DEBUG_r(
        RExC_study_chunk_recursed_count++;
    );`

which expands to: 
`    do { if (__builtin_expect((((my_perl->Idebug) & 0x00000200) ? (
   bool
   )1 : (
   bool
   )0),(0))) do { int saved_errno = 
   (*__errno_location ())
   ; (pRExC_state->study_chunk_recursed_count)++;; (
   (*__errno_location ()) 
   = saved_errno);} while (0); } while (0)
`

tonycoz:
IIRC my objections (and they aren't strong objections) were:

1. simple code like your example above becomes a lot more expensive, as you can see in your macro expansion, errno is typically a macro wrapping a function call, and save/restore of the last error number on Win32 are also function calls
2. for cases where debug output is controlled by a second conditional (in addition to the debug flags) you pay the cost of the save/restore even when that second condition is false., ie code like:

```
DEBUG_r( if (my condition) { my code } );
```
would expand to:
```
do { if (__builtin_expect((((my_perl->Idebug) & 0x00000200) ? ( bool )1 : ( bool )0),(0))) do { int saved_errno = (*__errno_location ()) ; if (my condition) { my code }; ( (*__errno_location ()) = saved_errno);} while (0); } while (0)
```
That said, I don't think it's a big deal, unless this code is in a tight loop.

khwilliamson:
This was addressed in a more general way by 50e8f677c30c6af7e17d923d97f6c32fa36be35f.

PR #18552, opened 2021-02-09
Second round of try/catch

leonerd:
This PR does not change the user-visible syntax or semantics of the new `try` feature, but does neaten up several details of internals.

* Adds a `newTRYCATCHOP()` function to call from the action block in `perly.y`, rather than inlining a nontrivial amount of optree code there
* Adds three new opcodes that better define the optree shape of `try/catch` syntax, rather than using the `OPf_SPECIAL` flag on the existing `entertry` op.
* Adds a new `cx_pushtry()` which creates a context stack entry that doesn't change `PL_curstackinfo->si_cxsubix` and therefore the workaround to make `return` in `try` behave correctly is no longer required (thanks @tonycoz).

Due to the new optree shape that the syntax compiles to, it is now much easier to recognise in `B::Deparse`. This has now been implemented and tested too. The new optree shape also adds a small amount of runtime efficiency:

* In the likely case that the `try` block does not throw an exception the `catch` subtree is entirely skipped over and does not need to be tested at all. 
* The two separate starting ops of `enter` + `entertry` have now been combined into a single `entertrycatch`.

Finally, the implementation leaves an easy space for future additions to be added, such as typed `catch(VAR isa CLASS)`.

### DIFF
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -195,4 +195,28 @@ no warnings 'experimental::try';
     }
 }
 
+# return from try is correct even for :lvalue subs
+#   https://github.com/Perl/perl5/issues/18553
+{
+    my $scalar;
+    sub fscalar :lvalue
+    {
+        try { return $scalar }
+        catch ($e) { }
+    }
+
+    fscalar = 123;
+    is($scalar, 123, 'try { return } in :lvalue sub in scalar context' );
+
+    my @array;
+    sub flist :lvalue
+    {
+        try { return @array }
+        catch ($e) { }
+    }
+
+    (flist) = (4, 5, 6);
+    ok(eq_array(\@array, [4, 5, 6]), 'try { return } in :lvalue sub in list context' );
+}
+
 done_testing;


op.c
```diff
@@ -9815,6 +9815,48 @@ Perl_newCONDOP(pTHX_ I32 flags, OP *first, OP *trueop, OP *falseop)
...
+C<OP_PADSV>).  All the optrees are consumed by this function and become part
+of the returned op tree. 
+
+The C<flags> argument is currently ignored.
+
+=cut
+ */
+
+OP *
+Perl_newTRYCATCHOP(pTHX_ I32 flags, OP *tryblock, OP *catchvar, OP *catchblock)
```
tonycoz: flags should be unsigned

inline.h
```diff
@@ -2313,12 +2313,8 @@ Perl_cx_popformat(pTHX_ PERL_CONTEXT *cx)
 
 
 PERL_STATIC_INLINE void
-Perl_cx_pusheval(pTHX_ PERL_CONTEXT *cx, OP *retop, SV *namesv)
+S_push_evalortry_common(pTHX_ PERL_CONTEXT *cx, OP *retop, SV *namesv)
```
tonycoz: inline functions (at least those in inline.h and visible to all XS) should use a `Perl_` prefix rather than the `S_` prefix

tonycoz:
There's some minor issues which don't mean much, but otherwise it could use some tests for the fixes to lvalue function handling.

op.c
leonerd: Hrm.. We don't seem to be very consistent with that:

```
leo@shy:~/src/bleadperl/perl [git]
$ grep -c "I32 flags" embed.fnc 
61

leo@shy:~/src/bleadperl/perl [git]
$ grep -c "U32 flags" embed.fnc 
167
```

In particular, every other "new*OP" function uses an I32:

```
ApdR    |OP*    |newCONDOP      |I32 flags|NN OP* first|NULLOK OP* trueop|NULLOK OP* falseop
ApdR    |OP*    |newFOROP       |I32 flags|NULLOK OP* sv|NN OP* expr|NULLOK OP* block|NULLOK OP* cont
ApdR    |OP*    |newLOGOP       |I32 optype|I32 flags|NN OP *first|NN OP *other
...
```

Perhaps for consistency with the other "new*OP" funcs at least, this should stay I32 too?

inline.h
leonerd: Fixed in 9b080fb469

leonerd:
@tonycoz 
> but otherwise it could use some tests for the fixes to lvalue function handling.

Yes; more tests are always good. I admit in this case though I'm not entirely sure what new case the code now handles that it didn't before - perhaps I should just copy your test case from #18553

leonerd:
Test added by e1083f3ffd

op.c
tonycoz: That's 73% U32, which is a pretty significant majority.

Following existing bad practice for the other OP constructors doesn't sway me much

I do think using an unsigned type is better for flags, but since we don't actually have any flags, let along one that might lead to promotion problems, I don't feel a need to insist on changing this.

leonerd: > Following existing bad practice for the other OP constructors doesn't sway me much

> I do think using an unsigned type is better for flags, but since we don't actually have any flags, let along one that might lead to promotion problems, I don't feel a need to insist on changing this.

Perhaps we should do a cleanup of these in a separate commit, then? Go through all our `I32` flags and see what ought to be `U32` instead.

leonerd:
If there's no further comment on this I'll merge it over the weekend (have already announced as such on #p5p yesterday)

PR #18561, opened 2021-02-12
Smoke me/yves hash and build fixes

demerphq:
This addresses a number of hash related issues and a few minor build issues I noticed along the way.

1. It hard caps the max size of a hash bucket array at 2**28 to prevent various forms of overflow.
2. It fixes the way siphash is set up so that -Accflags=-DPERL_HASH_FUNC_SIPHASH13 works as expected.
3. It ensures that seed and state variables are sized such that they align correctly.
4. Implements S_perl_hash_with_seed() which was not implemented leaving an untested branch in Hash::Util::hash_value() liable to fault, and adds tests.
5. Fixes some build warnings in PerlIO::encoding related to the fallback variable.
6. Fixes t/opbasic/arith.t to not warn and to use the right %Config::Config hash since %Config isn't imported.

Note that the above point 3 is related to issue #18555

### DIFF
--- a/ext/Hash-Util/Changes
+++ b/ext/Hash-Util/Changes
@@ -1,5 +1,14 @@
 Revision history for Perl extension Hash::Util.
 
+0.25, 0.24
+    Build fixes.
+
+0.23
+    Doc Fixes.
+
+0.22, 0.21
+    Build Fixes.
+
 0.20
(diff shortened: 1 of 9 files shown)


Leont:
The `PerlIO::encoding` patches look good to me

PR #18564, opened 2021-02-12
use SIPHASH 1-3 as the default 64 bit hash

demerphq:
Stadtx should be deprecated and removed after this. Siphash 1-3 has much  better statistical behavior, and is peer reviewed.


demerphq:
I am merging this as I want to remove Stadtx from Perl in time for 5.34.

PR #18567, opened 2021-02-12
[#18555] Ensure that PL_hash_seed and PL_hash_state are word aligned

demerphq:
Replace PL_hash_seed and PL_hash_state with PL_hash_seed_w and
PL_hash_state_w, and replace the old vars with defines that cast
their replacements appropriately. This should force linkers to
word align the data.

### DIFF
--- a/hv_func.h
+++ b/hv_func.h
@@ -36,21 +36,24 @@
 
 #if defined(PERL_HASH_FUNC_SIPHASH)
 # define __PERL_HASH_FUNC "SIPHASH_2_4"
-# define __PERL_HASH_WORD_SIZE sizeof(U64)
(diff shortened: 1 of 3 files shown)


hv_func.h
```diff
@@ -82,6 +86,12 @@
 */
 #include "perl_siphash.h"
 
+#define __PERL_HASH_SEED_roundup(x, y)   ( ( ( (x) + ( (y) - 1 ) ) / (y) ) * (y) )
```
afresh1, line 89: Shouldn't need to round if things are correctly sized.

demerphq: We derive the word size from the rounded up byte size so that  we can deal with mixed U64 and U32 data.

PR #18569, opened 2021-02-13
Yves/remove stadtx

demerphq:
Nuff said.


demerphq:
Furthermore on this I checked into the failure in DBIx::Class. It would seem that t/53lean_startup.t uses a set of heuristics (which aren't entirely clear to me) to notice what plug ins are being loaded. The old behavior of caller() meant that those heuristics started noticing that Class::C3::Componentised is loading a module, MRO::Compat, which it has been loading all along. The solution is pretty simple:
```
commit df052966ad4c29f4eb745fd8129a2a80e66c4387 (HEAD -> master)
Date:   Sat Feb 13 13:12:22 2021 +0100

    Ignore MRO::Compat
    
    In newer perls various bugs related to use statements and how
    caller reports them have been fixed. This means that our heuristics
    now notice a dependency of Class::C3::Componentised that it has
    been loading for some time, MRO::Compat, but which due to the bugs
    in caller() and the heuristics in t/53lean_startup.t was not registered
    as a valid plug in.

diff --git a/t/53lean_startup.t b/t/53lean_startup.t
index 9a4ef85..6b7216f 100644
--- a/t/53lean_startup.t
+++ b/t/53lean_startup.t
@@ -136,6 +136,7 @@ BEGIN {
 
     Class::Accessor::Grouped
     Class::C3::Componentised
+    MRO::Compat
     SQL::Abstract::Util
   ));
```
After applying this patch the test pass. It seems to me that the reality is the test code is making faulty assumptions in the first place since it seems it failed to notice this before, even though it was indeed being loaded. The stack frame counter variable $up changes from 3 to 13 in the old to new code because in the old code it thinks the BEGIN block is called from main, but it not, the BEGIN is executed in package Class::C3::Componentized, so when this changed the heuristic for detecting issues gets confused. I dont really understand. If i was testing this I wouldnt do any of this crazy stack walking. I would just track what was required() and then later check to see that it doesn't contain anything I don't expect, without any heuristics about the stack. The stack doesn't contain reliable data and the code doesn't detect when that data is not reliable and do anything about it. It is a very error prone strategy.

Leont:
> Nuff said.

For sake of posterity, can you please actually explain here why this is necessary.

demerphq:
I wrote Stadtx to be fast, but i overlooked some real statistical issues  in how it works. For certain keys it will be far less well distributed than preferable. meaning more bucket splits than necessary, more linked list searches than necessary, and less efficient utilization of available memory. The perl hash data structure degrades relatively well under these conditions, but the tiny speedup of the actual string hashing compared to the other costs isn't worth it. Having less collisions per bucket mitigates the speed differences to a certain extent also. Siphash produces well distributed data, hashes reasonably fast (especially in the 1-3 variant) and is one of the few widely used and peer reviewed PRF's. Stadtx is a project i came up with to see if I could create a fast hash function. There isn't even a question in terms of statistical quality comparison.  If we want to provide a "fast" hash then someone else can select one from the herd, but Stadtx shouldnt be an option.

demerphq:
Also I should mention that Stephen Dolan did some quality research into this. I had actually discovered the same issues independently but he did a much more formal investigation. Credit goes to him for instigating this removal.

jkeenan:
On 2/13/21 7:33 AM, Yves Orton wrote:
> Merged #18569 <https://github.com/Perl/perl5/pull/18569> into blead.
> 

You've gone from a pull request to removal of a documented feature in 
two hours and forty-two minutes.

Can you explain why this should not have been put through a deprecation 
cycle?

Or why it was rushed through so quickly?

Thank you very much.
Jim Keenan

Leont:
> Or why it was rushed through so quickly?

Yes, this. Don't get me wrong, I think this removal is the right decision, but you should at least give the rest of us a chance to come to that conclusion before merging it.

Issue #18570, opened 2021-02-13
t/op/magic.t: minitest failure on FreeBSD


jkeenan:
When invoked by `./miniperl -Ilib` on FreeBSD, one test in `t/op/magic.t` fails with:
```
# Failed test 188 - altering $0 is effective (testing with `ps`) at t/op/magic.t line 840
#      got 'x (miniperl)'
# expected /(?^:^(?:perl: )?x(?: \(perl\))?$)/
```
I originally reported this in https://github.com/Perl/perl5/issues/18547#issuecomment-776351326, but it now appears to have a cause unrelated to the problems discussed in that ticket.

The bug was introduced somewhere in a series of 3 commits in this range:
```
$ git diff 0cf25474a4..0522a9e73a |cat
diff --git a/t/op/magic.t b/t/op/magic.t
index 6283df5ac2..29db4c10a1 100644
--- a/t/op/magic.t
+++ b/t/op/magic.t
@@ -797,15 +797,19 @@ SKIP: {
 	env_is(__NoNeLoCaL => '');
 
     SKIP: {
-	    skip("\$0 check only on Linux and FreeBSD", 2)
-		unless $^O =~ /^(linux|android|freebsd)$/
-		    && open CMDLINE, "/proc/$$/cmdline";
-
-	    chomp(my $line = scalar <CMDLINE>);
-	    my $me = (split /\0/, $line)[0];
-	    is $me, $0, 'altering $0 is effective (testing with /proc/)';
-	    close CMDLINE;
-            skip("\$0 check with 'ps' only on Linux (but not Android) and FreeBSD", 1) if $^O eq 'android';
+	    skip("\$0 check only on Linux, Dragonfly BSD and FreeBSD", 2)
+		unless $^O =~ /^(linux|android|dragonfly|freebsd)$/;
+
+            SKIP: {
+                skip("No procfs cmdline support", 1)
+                    unless open CMDLINE, "/proc/$$/cmdline";
+
+                chomp(my $line = scalar <CMDLINE>);
+                my $me = (split /\0/, $line)[0];
+                is $me, $0, 'altering $0 is effective (testing with /proc/)';
+                close CMDLINE;
+            }
+            skip("No \$0 check with 'ps' on Android", 1) if $^O eq 'android';
             # perlbug #22811
             my $mydollarzero = sub {
               my($arg) = shift;
@@ -815,23 +819,25 @@ SKIP: {
               my $ps = (`ps -o command= -p $$`)[-1];
[...]

PR #18571, opened 2021-02-13
Smoke me/jkeenan/gh 18570 t op magic

jkeenan:
@tomhukins, with this patch `t/op/magic.t` once again PASSes on FreeBSD-12 when invoked via `./miniperl -Ilib t/op/magic.t` and during `make minitest`.  It continues to pass when run with `./perl`.

However, your inline comments suggest that this might need to be tested on other (older ?) FreeBSDs and on Dragonfly BSD.  Can you review?

For: https://github.com/Perl/perl5/issues/18570

Thank you very much.
Jim Keenan

### DIFF
--- a/t/op/magic.t
+++ b/t/op/magic.t
@@ -797,48 +797,48 @@ SKIP: {
 	env_is(__NoNeLoCaL => '');
 
     SKIP: {
-	    skip("\$0 check only on Linux, Dragonfly BSD and FreeBSD", 2)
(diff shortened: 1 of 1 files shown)


tomhukins (approved):

jkeenan:
I'm hoping to get some smoke-test results on Dragonfly BSD.  Once those are in, I will merge this p.r. and close the parent ticket.

Thank you very much.
Jim Keenan

PR #18581, opened 2021-02-14
Update search.cpan.org URLs

jkeenan:
search.cpan.org has (sadly) been gone for several years now.  This patch
updates the remaining important instances of that URL found in the parts
of the core distribution under the control of Perl 5 Porters, i.e.,
excluding the cpan/ directory.

The patch does not attempt to update URLs in "historical" data, e.g.,
older perldeltas.  Nor does it attempt to deal with results produced by
Porting/checkURL.pl.

### DIFF
--- a/ext/mro/mro.pm
+++ b/ext/mro/mro.pm
@@ -12,7 +12,7 @@ use warnings;
 
 # mro.pm versions < 1.00 reserved for MRO::Compat
 #  for partial back-compat to 5.[68].x
-our $VERSION = '1.25';
+our $VERSION = '1.25_001';
 
 require XSLoader;
 XSLoader::load('mro');
--- a/ext/mro/mro.xs
+++ b/ext/mro/mro.xs
@@ -63,8 +63,9 @@ S_mro_get_linear_isa_c3(pTHX_ HV* stash, U32 level)
 
     /* For a better idea how the rest of this works, see the much clearer
        pure perl version in Algorithm::C3 0.01:
-       http://search.cpan.org/src/STEVAN/Algorithm-C3-0.01/lib/Algorithm/C3.pm
-       (later versions go about it differently than this code for speed reasons)
+       https://fastapi.metacpan.org/source/STEVAN/Algorithm-C3-0.01/lib/Algorithm/C3.pm
+       (later versions of this module go about it differently than this code
+       for speed reasons)
     */
 
     if(isa && AvFILLp(isa) >= 0) {
--- a/pod/perlhack.pod
+++ b/pod/perlhack.pod
@@ -1166,7 +1166,7 @@ source, and we'll do that later on.
 Gisle Aas's "illustrated perlguts", also known as I<illguts>, has very
 helpful pictures:
 
-L<https://search.cpan.org/dist/illguts/>
+L<https://metacpan.org/release/RURBAN/illguts-0.49>
 
 =item * L<perlxstut> and L<perlxs>
 
--- a/dist/Devel-PPPort/HACKERS
+++ b/dist/Devel-PPPort/HACKERS
@@ -617,7 +617,7 @@ variable C<DPPP_CHECK_LEVEL> to 1 or 2.
 All of these files could be generated on the fly while building
 C<Devel::PPPort>, but not having the tests in F<t/> will confuse
 TEST/harness in the core. Not having F<PPPort.pm> will be bad for
-viewing the docs on C<search.cpan.org>. So unfortunately, it's
+viewing the docs on C<metacpan.org>. So unfortunately, it's
 unavoidable to put some redundancy into the package.
 
 =head2 Submitting Patches
--- a/dist/IO/IO.pm
+++ b/dist/IO/IO.pm
@@ -7,7 +7,7 @@ use Carp;
 use strict;
 use warnings;
 
-our $VERSION = "1.45_001";
+our $VERSION = "1.46";
 XSLoader::load 'IO', $VERSION;
 
 sub import {
--- a/dist/IO/IO.pm
+++ b/dist/IO/IO.pm
@@ -7,7 +7,7 @@ use Carp;
 use strict;
 use warnings;
 
-our $VERSION = "1.45";
+our $VERSION = "1.45_001";
 XSLoader::load 'IO', $VERSION;
 
 sub import {
@@ -53,7 +53,7 @@ in one go.  The IO modules belonging to the core are:
 
 Some other IO modules don't belong to the perl core but can be loaded
 as well if they have been installed from CPAN.  You can discover which
-ones exist by searching for "^IO::" on L<http://search.cpan.org>.
+ones exist with this query:  L<https://metacpan.org/search?q=IO%3A%3A>.
 
 For more information on any of these modules, please see its respective
 documentation.
--- a/dist/IO/lib/IO/Dir.pm
+++ b/dist/IO/lib/IO/Dir.pm
@@ -18,7 +18,7 @@ use File::stat;
 use File::Spec;
 
 our @ISA = qw(Tie::Hash Exporter);
-our $VERSION = "1.45";
+our $VERSION = "1.46";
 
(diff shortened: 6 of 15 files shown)


dist/PathTools/META.json
```diff
@@ -9,7 +9,7 @@
       "perl_5"
    ],
    "meta-spec" : {
-      "url" : "http://search.cpan.org/perldoc?CPAN::Meta::Spec",
+      "url" : "https://metacpan.org/pod/CPAN::Meta::Spec",
```
Grinnz: These META.json files are generated by EUMM so it's pointless to change them.

dist/if/META.json
```diff
@@ -9,7 +9,7 @@
       "perl_5"
    ],
    "meta-spec" : {
-      "url" : "http://search.cpan.org/perldoc?CPAN::Meta::Spec",
+      "url" : "http://metacpan.org/pod/CPAN::Meta::Spec",
```
Grinnz: Ditto above

dist/PathTools/META.json
jkeenan: But in this case (and the one below), *we* control the code (because "blead-upstream").  So how should these changes be made?

Grinnz: It's not a matter of control but that this will have no effect - these files are regenerated when making a release. This has to be fixed in EUMM or CPAN::Meta.

jkeenan: > It's not a matter of control but that this will have no effect - these files are regenerated when making a release. This has to be fixed in EUMM or CPAN::Meta.

When you say *"making a release"*, do you mean making a release of Perl 5 or a release of one of those distros separately on CPAN?

Grinnz: I mean releases of these distros on CPAN. The META.json files here have no bearing on Perl itself.

jkeenan: Suppose I were to proceed as if I were the person designated to do a CPAN release of PathTools.  I would create a separate git repository for it and copy relevant files from blead (with the exception of files such as `.gitignore`, `MANIFEST` and `MANIFEST.SKIP` (if any) that might need special handling).

Though I haven't been assigned to do a CPAN release of PathTools, I've done all of that locally (including updating `meta_spec:url` in `META.json` as proposed above) and pushed my work to github.  See:  https://github.com/jkeenan/PathTools/.

Now, I'll grant you that, in *my* repository for PathTools, when I say `perl Makefile.PL`, what is written to `Makefile` and to `MYMETA.json` is the search.cpan.org URL.
```
$ bleadperl Makefile.PL
Checking if your kit is complete...
Looks good
Generating a Unix-style Makefile
Writing Makefile for Cwd
Writing MYMETA.yml and MYMETA.json
[PathTools] 608 $ ack 'search\.cpan\.org' .
Makefile
635:	$(NOECHO) $(ECHO) '      "url" : "http://search.cpan.org/perldoc?CPAN::Meta::Spec",' >> META_new.json

MYMETA.json
12:      "url" : "http://search.cpan.org/perldoc?CPAN::Meta::Spec",
```
Furthermore, once I say `make dist` and create a CPAN-ready tarball and then examine that tarball, I see that `make dist` has changed the contents of `META.jsn`:
```
$ pwd
/tmp/PathTools-3.80
$ ack 'search\.cpan\.org' .
META.json
12:      "url" : "http://search.cpan.org/perldoc?CPAN::Meta::Spec",
```
These I would consider bugs (plural) in ExtUtils::MakeMaker.  (No evidence here as yet of bugs in CPAN::Meta::Spec.)

However, I think that's an argument for filing a bug report against ExtUtils::MakeMaker.  Has that been done?

I don't see that as a compelling argument for letting the old URL remain in `dist/PathTools/META.json` in core.

Thank you very much.
Jim Keenan

Grinnz: Your assessment is correct. But there is no reason to edit a META.json by hand, at all, including this one.

pod/perlhack.pod
```diff
@@ -1166,7 +1166,7 @@ source, and we'll do that later on.
 Gisle Aas's "illustrated perlguts", also known as I<illguts>, has very
 helpful pictures:
 
-L<https://search.cpan.org/dist/illguts/>
+L<https://metacpan.org/release/RURBAN/illguts-0.49>
```
Leont, line 1169: Linking to a specific version is probably not the most helpful solution here.

Grinnz: It's kind of the only option because illguts cannot be indexed.

Grinnz: The bigger problem is the contents of illguts is not visible on metacpan. I believe there is a better option but I don't recall it currently.

dist/PathTools/META.json
Leont: I agree it shouldn't be changed here. And as often in toolchain matters, there are some non-obvious difficulties in changing it (though I wouldn't say it's impossible).

pod/perlhack.pod
Grinnz: The contents are visible via st.aticpan.org: https://st.aticpan.org/source/RURBAN/illguts-0.49/index.html

dist/PathTools/META.json
jkeenan: I will revert the changes to the two .json files, as it appears that we will have to wait for changes in toolchain.

pod/perlhack.pod
jkeenan: If we have no better option than mentioning the version number in the hyperlink, then that's preferable IMO to leaving the search.cpan.org link in.

ext/mro/mro.pm
```diff
@@ -12,7 +12,7 @@ use warnings;
 
 # mro.pm versions < 1.00 reserved for MRO::Compat
 #  for partial back-compat to 5.[68].x
-our $VERSION = '1.25';
+our $VERSION = '1.25_001';
```
Leont, line 15: Why this dev-version?

ext/mro/mro.xs
```diff
@@ -63,8 +63,9 @@ S_mro_get_linear_isa_c3(pTHX_ HV* stash, U32 level)
 
     /* For a better idea how the rest of this works, see the much clearer
        pure perl version in Algorithm::C3 0.01:
-       http://search.cpan.org/src/STEVAN/Algorithm-C3-0.01/lib/Algorithm/C3.pm
-       (later versions go about it differently than this code for speed reasons)
+       https://fastapi.metacpan.org/source/STEVAN/Algorithm-C3-0.01/lib/Algorithm/C3.pm
```
Leont, line 66: Or https://metacpan.org/source/STEVAN/Algorithm-C3-0.01/lib/Algorithm/C3.pm

pod/perlhack.pod
Leont: Fair. We may just have to keep updating that link then though.

Grinnz: To be clear, I was suggesting to use the st.aticpan.org URL I linked here, because the metacpan link does not display the contents as search.cpan.org used to do here years ago (so was the original intention).

ext/mro/mro.xs
karenetheridge: Do we need to keep linking to versio 0.01, when the latest version (at present) is 0.11?

ext/mro/mro.pm
jkeenan: > Why this dev-version?

Pull request updated.

jkeenan:
No further objections having been lodged, this branch has been rebased and merged to blead.

Thank you very much.
Jim Keenan

pod/perlhack.pod
karenetheridge: It looks like this change request was missed.

ext/mro/mro.xs
karenetheridge: I believe fastapi is an internal host that shouldn't be linked to directly, so the main metacpan.org link would be preferred here.

PR #18587, opened 2021-02-17
Docs: Encourage SvPVbyte and SvPVutf8 over SvPV.

FGasper:
Per an (admittedly short) p5p discussion.

I don’t know if this is appropriate in Perl’s docs, but I’ve found utf8::downgrade a useful workaround for XS code that uses SvPV.

### DIFF
--- a/pod/perlguts.pod
+++ b/pod/perlguts.pod
@@ -153,27 +153,74 @@ Perl's own functions typically add a trailing C<NUL> for this reason.
 Nevertheless, you should be very careful when you pass a string stored
 in an SV to a C function or system call.
 
-To access the actual value that an SV points to, you can use the macros:
+To access the actual value that an SV points to, Perl's API exposes
+several macros that coerce the actual scalar type into an IV, UV, double,
+or string:
 
-    SvIV(SV*)
-    SvUV(SV*)
-    SvNV(SV*)
-    SvPV(SV*, STRLEN len)
-    SvPV_nolen(SV*)
+=over
 
-which will automatically coerce the actual scalar type into an IV, UV, double,
-or string.
+=item * C<SvIV(SV*)> (C<IV>) and C<SvUV(SV*)> (C<UV>)
 
-In the C<SvPV> macro, the length of the string returned is placed into the
-variable C<len> (this is a macro, so you do I<not> use C<&len>).  If you do
-not care what the length of the data is, use the C<SvPV_nolen> macro.
-Historically the C<SvPV> macro with the global variable C<PL_na> has been
-used in this case.  But that can be quite inefficient because C<PL_na> must
+=item * C<SvNV(SV*)> (C<double>)
+
+=item * Strings are a bit complicated:
+
+=over
+
+=item * Byte string: C<SvPVbyte(SV*, STRLEN len)> or C<SvPVbyte_nolen(SV*)>
+
+If the Perl string is C<"\xff\xff">, then this returns a 2-byte C<char*>.
+
+This is suitable for Perl strings that represent bytes.
+
+=item * UTF-8 string: C<SvPVutf8(SV*, STRLEN len)> or C<SvPVutf8_nolen(SV*)>
+
+If the Perl string is C<"\xff\xff">, then this returns a 4-byte C<char*>.
+
+This is suitable for Perl strings that represent characters.
+
+B<CAVEAT>: That C<char*> will be encoded via Perl's internal UTF-8 variant,
+which means that if the SV contains non-Unicode code points (e.g.,
+0x110000), then the result may contain extensions over valid UTF-8.
+See L<perlapi/is_strict_utf8_string> for some methods Perl gives
+you to check the UTF-8 validity of these macros' returns.
+
+=item * You can also use C<SvPV(SV*, STRLEN len)> or C<SvPV_nolen(SV*)>
+to fetch the SV's raw internal buffer. This is tricky, though; if your Perl
+string
+is C<"\xff\xff">, then depending on the SV's internal encoding you might get
+back a 2-byte B<OR> a 4-byte C<char*>.
+Moreover, if it's the 4-byte string, that could come from either Perl
+C<"\xff\xff"> stored UTF-8 encoded, or Perl C<"\xc3\xbf\xc3\xbf"> stored
+as raw octets. To differentiate between these you B<MUST> look up the
+SV's UTF8 bit (cf. C<SvUTF8>) to know whether the source Perl string
+is 2 characters (C<SvUTF8> would be on) or 4 characters (C<SvUTF8> would be
+off).
+
+B<IMPORTANT:> Use of C<SvPV>, C<SvPV_nolen>, or
+similarly-named macros I<without> looking up the SV's UTF8 bit is
+almost certainly a bug if non-ASCII input is allowed.
+
+When the UTF8 bit is on, the same B<CAVEAT> about UTF-8 validity applies
+here as for C<SvPVutf8>.
+
+=back
+
+(See L</How do I pass a Perl string to a C library?> for more details.)
+
+In C<SvPVbyte>, C<SvPVutf8>, and C<SvPV>, the length of the C<char*> returned
+is placed into the
+variable C<len> (these are macros, so you do I<not> use C<&len>). If you do
+not care what the length of the data is, use C<SvPVbyte_nolen>,
+C<SvPVutf8_nolen>, or C<SvPV_nolen> instead.
+The global variable C<PL_na> can also be given to
+C<SvPVbyte>/C<SvPVutf8>/C<SvPV>
+in this case.  But that can be quite inefficient because C<PL_na> must
 be accessed in thread-local storage in threaded Perl.  In any case, remember
 that Perl allows arbitrary strings of data that may both contain NULs and
 might not be terminated by a C<NUL>.
 
-Also remember that C doesn't allow you to safely say C<foo(SvPV(s, len),
+Also remember that C doesn't allow you to safely say C<foo(SvPVbyte(s, len),
 len);>.  It might work with your
 compiler, but it won't work for everyone.
 Break this sort of statement up into separate assignments:
@@ -181,8 +228,10 @@ Break this sort of statement up into separate assignments:
     SV *s;
     STRLEN len;
     char *ptr;
-    ptr = SvPV(s, len);
+    ptr = SvPVbyte(s, len);
     foo(ptr, len);
+
+=back
 
 If you want to know if the scalar value is TRUE, you can use:
 
@@ -200,7 +249,7 @@ add space for the trailing C<NUL> byte (perl's own string functions typically do
 C<SvGROW(sv, len + 1)>).
 
 If you want to write to an existing SV's buffer and set its value to a
-string, use SvPV_force() or one of its variants to force the SV to be
+string, use SvPVbyte_force() or one of its variants to force the SV to be
 a PV.  This will remove any of various types of non-stringness from
 the SV while preserving the content of the SV in the PV.  This can be
 used, for example, to append data from an API function to a buffer
@@ -3242,6 +3291,66 @@ in ways that your code cares about (see L<perlunicode/The "Unicode Bug">).
 There is no published API for dealing with this, as it is subject to
 change, but you can look at the code for C<pp_lc> in F<pp.c> for an
 example as to how it's currently done.
+
+=head2 How do I pass a Perl string to a C library?
+
+A Perl string, conceptually, is an opaque sequence of code points.
+Many C libraries expect their inputs to be "classical" C strings, which are
+arrays of octets 1-255, terminated with a NUL byte. Your job when writing
+an interface between Perl and a C library is to define the mapping between
+Perl and that library.
+
+Generally speaking, C<SvPVbyte> and related macros suit this task well.
+These assume that your Perl string is a "byte string", i.e., is either
+raw, undecoded input into Perl or is pre-encoded to, e.g., UTF-8.
+
+Alternatively, if your C library expects UTF-8 text, you can use
+C<SvPVutf8> and related macros. This has the same effect as encoding
+to UTF-8 then calling the corresponding C<SvPVbyte>-related macro.
+
+Some C libraries may expect other encodings (e.g., UTF-16LE). To give
+Perl strings to such libraries
+you must either do that encoding in Perl then use C<SvPVbyte>, or
+use an intermediary C library to convert from however Perl stores the
+string to the desired encoding.
+
+Take care also that NULs in your Perl string don't confuse the C
+library. If possible, give the string's length to the C library; if that's
+not possible, consider rejecting strings that contain NUL bytes.
+
+=head3 What about C<SvPV>, C<SvPV_nolen>, etc.?
+
+Consider a 3-character Perl string C<$foo = "\x64\x78\x8c">.
+Perl can store these 3 characters either of two ways:
+
+=over
+
+=item * bytes: 0x64 0x78 0x8c
+
+=item * UTF-8: 0x64 0x78 0xc2 0x8c
+
+=back
+
+Now let's say you convert C<$foo> to a C string thus:
+
+    STRLEN strlen;
+    char *str = SvPV(foo_sv, strlen);
+
+At this point C<str> could point to a 3-byte C string or a 4-byte one.
+
+Generally speaking, we want C<str> to be the same regardless of how
+Perl stores C<$foo>, so the ambiguity here is undesirable. C<SvPVbyte>
+and C<SvPVutf8> solve that by giving predictable output: use
+C<SvPVbyte> if your C library expects byte strings, or C<SvPVutf8>
+if it expects UTF-8.
+
+If your C library happens to support both encodings, then C<SvPV>--always
+in tandem with lookups to C<SvUTF8>!--may be safe and (slightly) more
+efficient.
+
+B<TESTING> B<TIP:> Use L<utf8>'s C<upgrade> and C<downgrade> functions
+in your tests to ensure consistent handling regardless of Perl's
+internal encoding.
 
 =head2 How do I convert a string to UTF-8?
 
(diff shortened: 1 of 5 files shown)


pod/perlguts.pod
```diff
@@ -158,30 +158,34 @@ To access the actual value that an SV points to, you can use the macros:
     SvIV(SV*)
     SvUV(SV*)
     SvNV(SV*)
-    SvPV(SV*, STRLEN len)
-    SvPV_nolen(SV*)
+    SvPVbyte(SV*, STRLEN len)
```
Leont: I'm not sure if ignoring `SvPV`/`SvPV_nolen` entirely here is the right solution.

FGasper: It’s discussed elsewhere, at least., but yeah.

CDB_File had a bug for years where the SV’s internal state dictated how it appeared in the CDB output. I noticed the same problem recently in URI::XSEscape and Net::Curl, and I began to think that pretty much any XS that outputs to something besides Perl probably has more use for the byte/utf8 variants of SvPV, so that a string will translate consistently to C regardless of whether it’s bytes/Latin-1 or UTF-8 internally.

Grinnz: At the very least, people should be advised that the only correct use of SvPV is in conjunction with knowledge of the UTF8 bit.

Tux: I also like to see them kept, if not only to be complete. Just add a warning if desired.
I'm with Grinnz here

FGasper:
Incidentally, YAML::Syck has the problem caused by using SvPV rather than SvPVbyte/SvPVutf8, too: https://github.com/toddr/YAML-Syck/issues/60

tonycoz:
I do think the distinction between UTF-8 flagged and and non-flagged SVs could be integrated into the other discussion of SVs in perlguts.  Right now we have general discussion of SVs, and then a separate discussion of Unicode further down.  The only mention of UTF8 is the example discussing direct buffer access added more years ago than I want to believe.

FGasper:
> I do think the distinction between UTF-8 flagged and and non-flagged SVs could be integrated into the other discussion of SVs in perlguts. Right now we have general discussion of SVs, and then a separate discussion of Unicode further down. The only mention of UTF8 is the example discussing direct buffer access added more years ago than I want to believe.

Clarifying the imperative for SvPV callers to check the string’s encoding would definitely be an improvement.

All the same, the following seem apparent:

1. SvPV leaks the internal string-encoding abstraction—or, at least, makes the caller go out of their way to preserve the abstraction. (Same problem as bytes.pm, but in C.)
2. SvPVbyte and SvPVutf8 preserve that abstraction.
3. All things being equal, it’s better for integrators to prefer “safe”, abstraction-preserving APIs rather than “dangerous”, abstraction-leaking ones.

… in light of which, wouldn’t some manner of “deprioritization” of SvPV (et al.) in the documentation be apropos?

FWIW, personally I’d like to see `SvPV_INTERNAL` be that macro’s canonical name, with a note in documentation about the “historical” `SvPV` alias.

pod/perlguts.pod
```diff
@@ -153,37 +153,73 @@ Perl's own functions typically add a trailing C<NUL> for this reason.
...
+is C<"\xff\xff">, then depending on the SV's internal encoding you might get
+back a 2-byte B<OR> a 4-byte C<char*>.
+Moreover, if it's the 4-byte string, that could come from either Perl
+C<"\xff\xff"> stored UTF-8 encoded, or Perl C<"\xc3\xbf\xc3\xbf"> stored
+as raw octets. To differentiate between these you B<MUST> look up the
+SV's UTF8 bit (cf. C<SvUTF8>) to know whether the source Perl string
+is 2 characters (C<SvUTF8> would be on) or 4 characters (C<SvUTF8> would be
+off).
+
+B<IMPORTANT:> Use of C<SvPV>, C<SvPV_nolen>, or
```
Leont, line 200: This ticket caused me to go through my CPAN modules to update them. And while some of them did need updating (in particular `Crypt::` ones), many did not. My experience is not that `SvPV` was always buggy for a simple reason: it all doesn't matter if your input is ascii.

And frankly, in a lot of cases allowed input is really just ascii.

Grinnz: Perhaps then "is almost certainly a bug if non-ASCII input is allowed." But using either SvPVbyte or SvPVutf8 in this case is also correct, so I think it's still a good idea not to recommend SvPV.

FGasper: I’ll update it, but FWIW I still think SvPVbyte is the “more correct” choice, even if the input is all ASCII, since the conceptual “leak” is still there with SvPV, even if there happens to be no externally observable difference in behaviour.

Leont: Yeah, "is almost certainly a bug if non-ASCII input is allowed" would be much better here. It's not that I disagree with the advice, it's more the tone.

FGasper: Change from the prior history:
```
diff --git a/pod/perlguts.pod b/pod/perlguts.pod
index a97435d9ef..c201be0a6f 100644
--- a/pod/perlguts.pod
+++ b/pod/perlguts.pod
@@ -189,7 +189,8 @@ off).

 B<IMPORTANT:> Use of C<SvPV>, C<SvPV_nolen>, or
 similarly-named macros I<without> looking up the SV's UTF8 bit is
-almost certainly a bug.
+a likely source of character-encoding bugs, unless the string in question
+is always fully UTF-8 invariant.
```

pod/perlguts.pod
FGasper: I’m resolving this, as I think I’ve addressed the concern.

FGasper:
I realized that `SvPVutf8` has its own problem since this is Perl’s “lax” UTF-8, not actual _valid_ UTF-8.

So I added:
```
commit 4ca94ddf03397718a9cc6958ef8956e99dfba4a7
Author: Felipe Gasper <felipe@felipegasper.com>
Date:   Mon Mar 22 07:04:07 2021 -0400

    Add caveat about SvPVutf8 and UTF-8 validity.

diff --git a/pod/perlguts.pod b/pod/perlguts.pod
index c201be0a6f..a5e1a988e5 100644
--- a/pod/perlguts.pod
+++ b/pod/perlguts.pod
@@ -171,10 +171,19 @@ or string:
 
 If the Perl string is C<"\xff\xff">, then this returns a 2-byte C<char*>.
 
+This is suitable for Perl strings that represent bytes.
+
 =item * UTF-8 string: C<SvPVutf8(SV*, STRLEN len)> or C<SvPVutf8_nolen(SV*)>
 
 If the Perl string is C<"\xff\xff">, then this returns a 4-byte C<char*>.
 
+This is suitable for Perl strings that represent characters.
+
+B<CAVEAT>: That C<char*> will be encoded via Perl’s internal "lax" UTF-8
+variant, which accepts various things that the official UTF-8 standard
+forbids. See L<perlapi/is_strict_utf8_string> for some methods Perl gives
+you to ensure that what these macros give you is I<actually> valid UTF-8.
+
 =item * You can also use C<SvPV(SV*, STRLEN len)> or C<SvPV_nolen(SV*)>
 to fetch the SV's raw internal buffer. This is tricky, though; if your Perl
 string
@@ -192,6 +201,9 @@ similarly-named macros I<without> looking up the SV's UTF8 bit is
 a likely source of character-encoding bugs, unless the string in question
 is always fully UTF-8 invariant.
 
+When the UTF8 bit is on, the same B<CAVEAT> about UTF-8 validity applies
+here as for C<SvPVutf8>.
+
 =back
 
 (See L</How do I pass a Perl string to a C library?> for more details.)
```

tonycoz:
> ```
> +B<CAVEAT>: That C<char*> will be encoded via Perl’s internal "lax" UTF-8
> +variant, which accepts various things that the official UTF-8 standard
> +forbids. See L<perlapi/is_strict_utf8_string> for some methods Perl gives
> +you to ensure that what these macros give you is I<actually> valid UTF-8.
> ```

The problem with calling it "lax" is it doesn't really explain anything.  I think it would be better to say if you're input SV has non-Unicode code points, then the result may contain extensions over valid UTF-8.

FGasper:
> > ```
> > +B<CAVEAT>: That C<char*> will be encoded via Perl’s internal "lax" UTF-8
> > +variant, which accepts various things that the official UTF-8 standard
> > +forbids. See L<perlapi/is_strict_utf8_string> for some methods Perl gives
> > +you to ensure that what these macros give you is I<actually> valid UTF-8.
> > ```
> 
> The problem with calling it "lax" is it doesn't really explain anything. I think it would be better to say if you're input SV has non-Unicode code points, then the result may contain extensions over valid UTF-8.

“lax” is [what Larry called it](https://metacpan.org/pod/Encode#UTF-8-vs.-utf8-vs.-UTF8), but that change makes sense. I’ll amend.

pod/perlguts.pod
FGasper: @Leont I made the above change before I saw your last comment in this thread. Does what I did satisfy your concerns, or should I change to `is almost certainly a bug if non-ASCII input is allowed`? I don’t have a strong opinion, though I _slightly_ prefer “UTF-8 invariant” rather than “ASCII” in light of Perl’s EBCDIC support.

Leont: Either is fine to me. The former is more correct, but the latter may be easier to understand for people less deep into character encodings.

Grinnz: A parenthetical `(or non-EBCDIC)` would be sufficient for correctness in the latter case IMO - note that it isn't even UTF-8 being used on these systems anyway.

FGasper: I just went with `is almost certainly a bug if non-ASCII input is allowed`. `perlebcdic` kinda makes clear that EBCDIC folks need to “translate” ASCII in Perl’s documentation to EBCDIC anyway.

FGasper:
I realized that SvPV’s description in `perlapi` was unchanged. So to the section that describes the different SvPV* macros in sv.h I added:
```
+The forms with neither C<byte> nor C<utf8> in their names (e.g., C<SvPV> or
+C<SvPV_nolen>) can expose the SV's internal string buffer. If
+that buffer consists entirely of bytes 0-255 and includes any bytes above
+127, then you B<MUST> consult C<SvUTF8> to determine the actual code points
+the string is meant to contain. Generally speaking, it is probably safer to
+prefer C<SvPVbyte>, C<SvPVutf8>, and the like. See
+L<perlguts/How do I pass a Perl string to a C library?> for more details.
```

sv.h
```diff
@@ -801,7 +801,10 @@ compiler will complain if you were to try to modify the contents of the string,
 (unless you cast away const yourself).
 
 =for apidoc Am|STRLEN|SvCUR|SV* sv
-Returns the length of the string which is in the SV.  See C<L</SvLEN>>.
+Returns the length, in bytes, of the C string which is in the SV.
+Note that this may not match Perl's C<length>; for that, use
+C<SvUTF8(sv) ? sv_len_utf8(sv) : sv_len(sv)>.
```
salva: There is really no need to check the value of `SvUTF8(sv)`, `sv_len_utf8` already does it internally.

sv.h
```diff
@@ -1657,6 +1660,14 @@ see C<L</SvPV_force>>.
 
 The differences between the forms are:
 
+The forms with neither C<byte> nor C<utf8> in their names (e.g., C<SvPV> or
+C<SvPV_nolen>) can expose the SV's internal string buffer. If
+that buffer consists entirely of bytes 0-255 and includes any bytes above
+127, then you B<MUST> consult C<SvUTF8> to determine the actual code points
+the string is meant to contain. Generally speaking, it is probably safer to
```
salva, line 1666: > Generally speaking, it is probably safer

Is that too mild? IMO the usage of `SvPV` should be clearly discouraged, more on the line of...

> **Don't use SvPV** unless you know what you are doing.

pod/perlguts.pod
```diff
@@ -3243,6 +3292,66 @@ There is no published API for dealing with this, as it is subject to
...
+
+At this point C<str> could point to a 3-byte C string or a 4-byte one.
+
+Generally speaking, we want C<str> to be the same regardless of how
+Perl stores C<$foo>, so the ambiguity here is undesirable. C<SvPVbyte>
+and C<SvPVutf8> solve that by giving predictable output: use
+C<SvPVbyte> if your C library expects byte strings, or C<SvPVutf8>
+if it expects UTF-8.
+
+If your C library happens to support both encodings, then of course
```
salva: That would happen very seldom and there are several gotchas associated with doing that.

I think it would be better to just remove this paragraph or at least the "is preferable" part.

salva (approved):
I have made some comments, but all of them are minor issues.

As it is right now this patch is already a big improvement over what we have.

sv.h
FGasper: Thank you for pointing this out! I’m revising.

sv.h
FGasper: I agree, but I’m not sure that’s generally accepted. For now I just want to improve on status quo.

pod/perlguts.pod
FGasper: Revised to:
```
+If your C library happens to support both encodings, then C<SvPV>--always
+in tandem with lookups to C<SvUTF8>!--may be safe and (slightly) more
+efficient.
```

FGasper:
@Grinnz @Leont @khwilliamson Do you see any outstanding issues as of now with this PR?

Thank you!

Grinnz:
LGTM

pod/perlguts.pod
```diff
@@ -153,37 +153,86 @@ Perl's own functions typically add a trailing C<NUL> for this reason.
...
+This is suitable for Perl strings that represent bytes.
+
+=item * UTF-8 string: C<SvPVutf8(SV*, STRLEN len)> or C<SvPVutf8_nolen(SV*)>
+
+If the Perl string is C<"\xff\xff">, then this returns a 4-byte C<char*>.
+
+This is suitable for Perl strings that represent characters.
+
+B<CAVEAT>: That C<char*> will be encoded via Perl's internal UTF-8 variant,
+which means that if the SV contains non-Unicode code points (e.g., U+FFFF),
```
khwilliamson: U+FFFF is a non-character Unicode code point, so this example is wrong.  I suggest 0x110000.  'U+' is strictly wrong here, which is why I used '0x'

khwilliamson: In documentation I've written recently, I've tried to be careful to use U+ only for the 0-10FFFF range

pod/perlguts.pod
```diff
@@ -3243,6 +3292,66 @@ There is no published API for dealing with this, as it is subject to
...
+These assume that your Perl string is a "byte string", i.e., is either
+raw, undecoded input into Perl or is pre-encoded to, e.g., UTF-8.
+
+Alternatively, if your C library expects UTF-8 text, you can use
+C<SvPVutf8> and related macros. This has the same effect as encoding
+to UTF-8 then calling the corresponding C<SvPVbyte>-related macro.
+
+Some C libraries may expect other encodings (e.g., UTF-16LE). To give
+Perl strings to such libraries
+you must either do that encoding in Perl then use C<SvPVbyte>, or
```
khwilliamson, line 3313: Add comma before 'then'

pod/perlguts.pod
```diff
@@ -3243,6 +3292,66 @@ There is no published API for dealing with this, as it is subject to
...
+if it expects UTF-8.
+
+If your C library happens to support both encodings, then C<SvPV>--always
+in tandem with lookups to C<SvUTF8>!--may be safe and (slightly) more
+efficient.
+
+B<TESTING> B<TIP:> Use L<utf8>'s C<upgrade> and C<downgrade> functions
+in your tests to ensure consistent handling regardless of Perl's
+internal encoding.
+
```
khwilliamson, line 3354: I really like this section

sv.h
```diff
@@ -855,8 +857,8 @@ Set the value of the MAGIC pointer in C<sv> to val.  See C<L</SvIV_set>>.
 Set the value of the STASH pointer in C<sv> to val.  See C<L</SvIV_set>>.
 
 =for apidoc Am|void|SvCUR_set|SV* sv|STRLEN len
-Set the current length of the string which is in the SV.  See C<L</SvCUR>>
-and C<SvIV_set>>.
+Sets the current length, in bytes, of the C string which is in the SV.
```
khwilliamson, line 860: I don't understands why this has to be a C string.  The SV may contain embedded NULs, and those are counted as bytes.

sv.h
khwilliamson: I'm OK with the current new wording too

sv.h
khwilliamson: Perhaps an explicit mention that the string may contain NULs and they are counted.

khwilliamson (changes requested):

pod/perlguts.pod
FGasper: Fixed.

pod/perlguts.pod
FGasper: That seems like it would potentially confuse; the separation that the single comma creates between the two options would be weakened by having another comma.

What about:

a) Reformat it as an ordered list (`=item a)`, `=item b)`)

b) Replace the comma before `or` with a semicolon

c) Just add the extra comma, if you think it’s best that way.

?

sv.h
FGasper: Yeah I meant “C string” as distinct from “Perl string”, i.e., what `sv_len_utf8` does. When I looked at it I wasn’t sure if it referred to Perl `length()` or the number of bytes in the PV.

Maybe:
> Sets the current length of the PV inside the SV.

?

pod/perlguts.pod
FGasper: @khwilliamson ^^ Question for you, when you have a chance?

sv.h
khwilliamson: Sets the current length, in bytes, of the PV inside the SV

would be better IMO

FGasper: Done.

pod/perlguts.pod
khwilliamson: I'm trying to see that fix.  I reloaded this PR and it says the commit was in March

FGasper: I haven’t pushed; I was waiting for your thoughts on https://github.com/Perl/perl5/pull/18587#discussion_r607404905

But I‘ll push up now.

Issue #18473, opened 2021-01-13
perl's NaN is -NaN


t-a-k:
**Description**
NaN values generated by perl, such as the result of a string-to-number conversion or the returned value of `POSIX::NAN`, are actually -NaN (i.e. its sign bit is inverted).

As -NaN is also NaN, this seems not to be a serious issue, but inconsistent with underlying C library/compilers.

**Steps to Reproduce**
```
$ perl -wle 'print unpack "H*", pack "F>", 0 + "NaN"'
fff8000000000000
$ perl -MPOSIX -wle 'print "NAN is negative" if POSIX::signbit(POSIX::NAN())'
NAN is negative
```

**Expected behavior**
perl should generate NaNs with their sign bits cleared.

**Perl configuration**
```
$ perl -V
Summary of my perl5 (revision 5 version 33 subversion 6) configuration:
  Commit id: a55134cc382529aee24f81156d7540b61a753544
  Platform:
    osname=linux
    osvers=4.19.0-13-686-pae
    archname=i686-linux
    uname='linux x41 4.19.0-13-686-pae #1 smp debian 4.19.160-2 (2020-11-28) i686 gnulinux '
    config_args='-de -Dusedevel -DEBUGGING'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=undef
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2 -g'
    cppflags='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='8.3.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=4
    doublesize=8
    byteorder=1234
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=12
    longdblkind=3
    ivtype='long'
    ivsize=4
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=4
    prototype=define
[...]

PR #18590, opened 2021-02-17
POSIX: Prefer system's NAN/INFINITY for POSIX::NAN and POSIX::INFINITY

t-a-k:
`POSIX::NAN` and `POSIX::INFINITY` used to return `NV_NAN` and `NV_INF` defined in the perl core respectively, but these might be different from the values defined in the POSIX system header, as these special values might have multiple internal representations.

If respective macro is not defined (eg. on pre-C99 system), they will remain to return `NV_NAN`/`NV_INF`, so there should be no degradation.

This PR addresses (but does not completely resolve) #18473.

### DIFF
--- a/ext/POSIX/Makefile.PL
+++ b/ext/POSIX/Makefile.PL
@@ -95,17 +95,19 @@ END
 #else
                    HUGE_VAL
 #endif
-                           '});
+                           '},
+   # POSIX says these are constant expressions, but if these are not available
+   # we might fallback to NV_INF/NV_NAN which may be not constants.
+   {name=>"INFINITY", type=>"NV", not_constant=>1},
+   {name=>"NAN",      type=>"NV", not_constant=>1});
 
 if ($Config{d_double_has_inf}) {
     push @names,
-        {name=>"INFINITY", type=>"NV", value=>"NV_INF", not_constant=>1},
         {name=>"Inf", type=>"NV", value=>"NV_INF", not_constant=>1},
 }
 
 if ($Config{d_double_has_nan}) {
     push @names,
-        {name=>"NAN", type=>"NV", value=>"NV_NAN", not_constant=>1},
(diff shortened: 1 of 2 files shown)


khwilliamson:
This LGTM, but @sisyphus could you confirm?

sisyphus:
LGTM, too.
I recently noticed that, in the mpfr library, when a NaN mpfr object is converted to a NaN double, the sign of the NaN is not always preserved,
I suggested to the mpfr developers that they might like to fix this, but they declined, saying that the sign of a NaN is meaningless and that IEEE standards agree with their stance.
The surprising thing is that, as I pointed out to them, the current devel mpfr contains an mpfr_total_order() function in which the sign of NaN is meaningful:
```
 -- Function: int mpfr_total_order_p (mpfr_t X, mpfr_t Y)
     This function implements the totalOrder predicate from IEEE 754,
     where −NaN < −Inf < negative finite numbers < −0 < +0 < positive
     finite numbers < +Inf < +NaN.  It returns a non-zero value (true)
     when X is smaller than or equal to Y for this order relation, and
     zero (false) otherwise.  Contrary to ‘mpfr_cmp (x, y)’, which
     returns a ternary value, ‘mpfr_total_order_p’ returns a binary
     value (zero or non-zero).  In particular, ‘mpfr_total_order_p (x,
     x)’ returns true, ‘mpfr_total_order_p (-0, +0)’ returns true and
     ‘mpfr_total_order_p (+0, -0)’ returns false.  The sign bit of NaN
     also matters.
```
I guess their view is that, apart from "total order" considerations, whether a NaN displays with or without a leading sign is irrelevant and meaningless.

However, there's certainly no reason to NOT make changes to POSIX NaN, such that expectations are met ... and I think this PR fits the bill of "meeting those expectations".

Cheers,
Rob

PR #18594, opened 2021-02-18
Alter documentation to reflect behaviour

benkasminbullock:
The name "wanted" chosen for the documentation was misleading, as the
documentation itself pointed out. I've changed it to "found" except in
the one place where "wanted" has to be used.

I've also addressed several other oddities in the documentation.

### DIFF
--- a/ext/File-Find/lib/File/Find.pm
+++ b/ext/File-Find/lib/File/Find.pm
@@ -795,12 +795,12 @@ File::Find - Traverse a directory tree.
 =head1 SYNOPSIS
 
     use File::Find;
-    find(\&wanted, @directories_to_search);
-    sub wanted { ... }
+    find(\&found, @directories_to_search);
+    sub found { ... }
 
     use File::Find;
-    finddepth(\&wanted, @directories_to_search);
-    sub wanted { ... }
+    finddepth(\&found, @directories_to_search);
+    sub found { ... }
 
     use File::Find;
     find({ wanted => \&process, follow => 1 }, '.');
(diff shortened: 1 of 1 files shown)


Leont:
When passed as a named argument it's called `wanted`, so IMO the least surprising thing we can do is to use the same name when passing it as a positional argument.

jkeenan:
> The name "wanted" chosen for the documentation was misleading, as the
> documentation itself pointed out. I've changed it to "found" except in
> the one place where "wanted" has to be used.
> 
> I've also addressed several other oddities in the documentation.

You've submitted one pull request for documentation changes which you earlier cited in three separate issues.  As is already evident, some of your suggested changes are controversial.  I think you would be better off submitting a separate pull request for each of the three problems you described in those tickets.  That might improve your chances of getting some of the changes accepted.

Thank you very much.
Jim Keenan

benkasminbullock:
> You've submitted one pull request for documentation changes which you earlier cited in three separate issues. As is already evident, some of your suggested changes are controversial. I think you would be better off submitting a separate pull request for each of the three problems you described in those tickets. That might improve your chances of getting some of the changes accepted.

No thanks, I've done enough work here.

Leont:
> You've submitted one pull request for documentation changes which you earlier cited in three separate issues. As is already evident, some of your suggested changes are controversial. I think you would be better off submitting a separate pull request for each of the three problems you described in those tickets. That might improve your chances of getting some of the changes accepted.

Yeah, or at least split out into three commits

benkasminbullock:
Sorry but as stated above I'm not willing to spend more time on File::Find.

PR #18595, opened 2021-02-18
Disallow double signs and spaces between a sign and number in my_atof3

### LINKED ISSUE #18584: [doc] perlnumber: "minus plus number" looks like number
**Where**

https://perldoc.perl.org/perlnumber

**Description**

Quoted "minus plus number" behaves like negative number in numeric context.
Seems undocumented. Usually numbers start with either `+` or `-` or none.

(perl v5.32.0)
```
#!/usr/bin/perl

use warnings;
use strict;

$\ = $/;

my $A = "-+3";
my $B = " -+3";
my $C = "- +3";
my $D = "-+ 3";

print map "[$_]", map $_ + 1,
	$A, $B, $C, $D,
	;
```
OUTPUT:
```
Argument "-+3" isn't numeric in addition (+) at minus_plus_number_1.pl line 13.
Argument " -+3" isn't numeric in addition (+) at minus_plus_number_1.pl line 13.
Argument "- +3" isn't numeric in addition (+) at minus_plus_number_1.pl line 13.
Argument "-+ 3" isn't numeric in addition (+) at minus_plus_number_1.pl line 13.
[-2][-2][-2][1]
```

t-a-k:
`Perl_my_atof3` used to pass a substring after the first (optional) sign to `S_strtod`, which causes wrong numifications for strings like `"-+3"` or `"+ 0x123"` (for the latter case, while `Perl_my_atof3 already` had the code to block `"0x"` prefixes, this string will slip through due to the space character in it).

Will fix #18584.

### DIFF
--- a/numeric.c
+++ b/numeric.c
@@ -1648,6 +1648,14 @@ Perl_my_atof3(pTHX_ const char* orig, NV* value, const STRLEN len)
     while (s < send && isSPACE(*s))
         ++s;
 
+#  if defined(NV_INF) || defined(NV_NAN)
+    {
+        char* endp;
+        if ((endp = S_my_atof_infnan(aTHX_ s, FALSE, send, value)))
+            return endp;
+    }
+#  endif
+
     /* sign */
     switch (*s) {
         case '-':
@@ -1663,9 +1671,6 @@ Perl_my_atof3(pTHX_ const char* orig, NV* value, const STRLEN len)
         char* endp;
         char* copy = NULL;
 
-        if ((endp = S_my_atof_infnan(aTHX_ s, negative, send, value)))
-            return endp;
-
         /* strtold() accepts 0x-prefixed hex and in POSIX implementations,
            0b-prefixed binary numbers, which is backward incompatible
         */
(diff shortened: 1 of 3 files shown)


numeric.c
```diff
@@ -1675,14 +1675,18 @@ Perl_my_atof3(pTHX_ const char* orig, NV* value, const STRLEN len)
             return (char *)s+1;
         }
 
+        /* strtod will parse a sign (and skip leading whitespaces) by itself,
+         * so rewind s to the beginning of the string. */
```
hvds: I think it is important to point out is the reason we _should_ let it parse them itself, ie that we'd get an undesirable result if we pointed it at some apparently-leading whitespace _after_ the sign; maybe something like this:
```
        /* We do not want strtod to parse whitespace after the sign, since
         * that would give backward-incompatible results. So we rewind and 
         * let strtod handle the whitespace and sign character itself. */
```

t-a-k: Thank you, but I've forgotten that `"+-inf"` is still wrongly numified, so I will fix it too.

PR #18596, opened 2021-02-20
Ignore microperl and by-products if we build it

Corion:
As long as `Makefile.micro` is a first-class citizen in the Perl distribution, we should also ignore the by-products of building it


jkeenan:
> As long as `Makefile.micro` is a first-class citizen in the Perl distribution, we should also ignore the by-products of building it

Does microperl work **at all**?
Following instructions in `README.micro`, I tried the following on both Linux and FreeBSD.
```
$ make -f Makefile.micro patch_uconfig
perl -MConfig -pi -e "s/^((?:short|int|long(?:dbl|long)?|ptr|double|[iun]v|u?quad|[iu]\d+|fpos|lseek)(?:size|type)|byteorder|d_quad|quadkind|use64.+)=.*/\\1='"'$Config{$1}'"'/g" uconfig.sh
[perlmonger: perl] $ make -f Makefile.micro
cc -c -o uav.o -DPERL_CORE -DPERL_MICRO -DPERL_USE_SAFE_PUTENV -DNO_MATHOMS  av.c
In file included from av.c:20:
In file included from ./perl.h:4043:
In file included from ./iperlsys.h:50:
./perlio.h:32:4: error: "stdio is no longer supported as the default base layer -- use perlio."
#  error "stdio is no longer supported as the default base layer -- use perlio."
   ^
In file included from av.c:20:
In file included from ./perl.h:4077:
In file included from ./regexp.h:296:
./op_reg_common.h:140:30: error: token is not a valid binary operator in a preprocessor
      subexpression
#if RXf_PMf_COMPILETIME != ((nBIT_MASK(_RXf_PMf_SHIFT_COMPILETIME)) \
```
... and many more warnings/failures until:
```
./handy.h:342:24: note: expanded from macro 'nBIT_MASK'
#define nBIT_MASK(n) ((UINTMAX_C(1) << (n)) - 1)
                       ^
./perl.h:1222:26: note: expanded from macro 'UINTMAX_C'
#    define UINTMAX_C(c) UINT32_C(c)
                         ^
./perl.h:1118:27: note: expanded from macro 'UINT32_C'
#    define UINT32_C(x) ((U32_TYPE)x##U)
                          ^
13 errors generated.
*** Error code 1

Stop.
```

jkeenan:
> > As long as `Makefile.micro` is a first-class citizen in the Perl distribution, we should also ignore the by-products of building it
> 
> Does microperl work **at all**?

`pod/perl5180delta.pod` says:
```
Future Deprecations
...
    *   "microperl", long broken and of unclear present purpose, will be
        removed.
```
When will the future arrive?

Corion:
I have another branch at https://github.com/Corion/perl5/tree/corion/microperl in which I restore compiling `microperl` , but I want to take small steps here. I'm not a user of `microperl`, so I don't know what good it is, but it compiles blazingly fast ( 6 seconds ), so I see a use when modifying/testing regex changes for example.

PR #18599, opened 2021-02-21
Correct version information for older perl*delta.pod

### LINKED ISSUE #18598: [doc] Some perl*delta.pod files incorrectly refer to older perl*delta.pod files.
**Where**

* `pod/perl5121delta.pod`
* `pod/perl5122delta.pod`
* `pod/perl5301delta.pod`
* `pod/perl5321delta.pod`

**Description**

Most `perl*delta.pod` files refer to older `perl*delta.pod` files:
> If you are upgrading from an earlier release such as \<*VERSION*>, first read
> L\<perl*VERSION*delta>, which describes differences between *VERSION* and
> *VERSION*.

The ones listed above have incorrect version numbers.

I'll create a pull request to fix this.

Keith-S-Thompson:
Correct version information in some perl*delta.pod files.

Fixes #18598

### DIFF
--- a/pod/perl5121delta.pod
+++ b/pod/perl5121delta.pod
@@ -10,7 +10,7 @@ This document describes differences between the 5.12.0 release and
 the 5.12.1 release.
 
 If you are upgrading from an earlier release such as 5.10.1, first read
-L<perl5120delta>, which describes differences between 5.10.1 and
+L<perl5120delta>, which describes differences between 5.10.0 and
 5.12.0.
 
 =head1 Incompatible Changes
--- a/pod/perl5122delta.pod
+++ b/pod/perl5122delta.pod
@@ -10,7 +10,7 @@ This document describes differences between the 5.12.1 release and
 the 5.12.2 release.
 
 If you are upgrading from an earlier major version, such as 5.10.1,
-first read L<perl5120delta>, which describes differences between 5.10.1
+first read L<perl5120delta>, which describes differences between 5.10.0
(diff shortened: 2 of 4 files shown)


pod/perl5121delta.pod
```diff
@@ -10,7 +10,7 @@ This document describes differences between the 5.12.0 release and
 the 5.12.1 release.
 
 If you are upgrading from an earlier release such as 5.10.1, first read
-L<perl5120delta>, which describes differences between 5.10.1 and
+L<perl5120delta>, which describes differences between 5.10.0 and
```
Leont, line 13: I don't think this is correct. Versioning worked differently before 5.12 than it does now.

Grinnz: perl5120delta does seem to include the changes included in 5.10.1, so this change seems fine to me.

Keith-S-Thompson: perl5120delta.pod itself says:

> This document describes differences between the 5.10.0 release and the
> 5.12.0 release.

so I went by that. But it also says:

> Many of the bug fixes in 5.12.0 are already included in the 5.10.1
> maintenance release.

Does perl5120delta.pod need to be updated (also or instead)? I haven't looked into the details.

Keith-S-Thompson:
Can anyone explain why some tests failed? I don't *think* it could have been related my changes, but do I need to do anything to resolve this?

Grinnz:
They are corelist failures from the recent release - if you rebase your PR upon latest blead they should be resolved.

PR #18607, opened 2021-02-24
Add ability to easily add info to DEBUG output

khwilliamson:
This commit adds two macros that a user can define and recompile Perl to
get every active DEBUG statement to do  something beyond what it would
normally do.

This allows someone to recompile Perl when they need to delve deeper
into fixing a bug without increasing memory use or slowing execution
otherwise.

This could be used to save and restore the errno around debugging, or to print location information about the DEBUG macro calls:
```#dedine DEBUG_PRE   PerlIO_printf(Perl_debug_log, "%s:%d: ", __FILE__, __LINE__);```

or some other variable, like aTHX, or some PL_foo.

I would prefer better names than DEBUG_PRE and POST, suggestions welcome
`

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -4487,16 +4487,22 @@ Gid_t getegid (void);
                               }                                         \
     } STMT_END
 
+#ifndef DEBUG_PRE
+#  define DEBUG_PRE
+#endif
+#ifndef DEBUG_POST
+#  define DEBUG_POST
+#endif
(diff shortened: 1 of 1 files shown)


khwilliamson:
This was addressed by 50e8f677c30c6af7e17d923d97f6c32fa36be35f, which added comments, changed the names slightly from this PR

Issue #10623, opened 2010-09-12
rename, chroot etc. ignore internal encoding


p5pRT:
Migrated from [rt.perl.org#77798](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=77798) (status was 'open')

Searchable as RT77798$

p5pRT:
### From perlbug@plan9.de

#### Created by perlbug@plan9.de



This snippet calls rename with two different paths\, even though the same
string is passed to rename\.

&nbsp;   perl \-e 'my $x = chr 200; rename $x\,0; utf8&#8203;::encode $x; rename $x\,0'

The fact that the internal \(basically invisible to a perl program\)
encoding changes should not change semantics of I/O functions\.

The solution is to use the equivalent of SvPVbyte\, not SvPV\, when passing
paths \(or other 8b\-it data\) to posix functions\.

A cursory examination of pp\_sys shows that at least backtick\, open\,
dbmopen\, sysopen\, truncate\, bind\, setsockopt\, getsockopt\, getpeername\,
stat\, chdir\, chroot\, link\, readlink\, mkdir\, rmdir\, opendir\, system\, exec\,
gethost\*\, getproto\*\, getserv\* etc\. are affected \(I stopped looking\)\.

All those functions silently throw away the crucial information of how
bytes are encoded in a string\. As modules and programs using unicode
become more common\, this problem will become a major issue\.

\(When in doubt\, it always helps to review the discussion about crypt\(\)
which was fixed during 5\.006 times\)\.

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=medium

Site configuration information for perl 5.10.1:

Configured by Marc Lehmann at Wed May  5 10:53:04 CEST 2010.

Summary of my perl5 (revision 5 version 10 subversion 1) configuration:
   
  Platform:
    osname=linux, osvers=2.6.26-2-amd64, archname=amd64-linux
    uname='linux cerebro 2.6.26-2-amd64 #1 smp thu nov 5 02:23:12 utc 2009 x86_64 gnulinux '
[...]

PR #18609, opened 2021-02-28
DRAFT: Sysbinmode

FGasper:
Proof-of-concept of proposed fix for https://github.com/Perl/perl5/issues/10623

### DIFF
--- a/doio.c
+++ b/doio.c
@@ -2283,6 +2283,30 @@ S_exec_failed(pTHX_ const char *cmd, int fd, int do_report)
     }
 }
 
+char*
+_get_syscall_arg(pTHX, SV* perlarg) {
+    HV * const hinthv = PL_hints & HINT_LOCALIZE_HH ? GvHV(PL_hintgv) : NULL;
+
+    if (hinthv) {
+        SV ** const svp = hv_fetchs(hinthv, "sysbinmode/setting", FALSE);
+        if (svp) {
+            const UV setting = (UV) SvUV(*svp);
+            warn("sysbinmode setting %u\n", setting);
+
+            switch (setting) {
+                case 1:
+                    return SvPVbyte_nolen(perlarg);
+                case 2:
+                    return SvPVutf8_nolen(perlarg);
+                default:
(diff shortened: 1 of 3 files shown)


Leont:
Given we're beyond the controversial deadline for 5.34, we should not merge this before 5.35.

(I suspect this was meant as a draft, but it wasn't marked as such)

FGasper:
> Given we're beyond the controversial deadline for 5.34, we should not merge this before 5.35.

Fine w/ me … the issue it fixes is certainly not a new regression. :)

Leont:
This functionality can be implemented as a XS module, and I would argue it should be. That not only allows us to experiment with it, but also allows us to support this on older perl releases.

FGasper:
@Leont I’m happy to do this on CPAN … but how can I affect what exec et al. do in XS? Can you think of an existing module I could look at for an example?

ikegami:
I think this is a step in the wrong direction. We gotta stop telling Perl what encoding to use. [My objection](https://github.com/Perl/perl5/issues/10623#issuecomment-787520455)

FGasper:
@ikegami I’m fine honestly with withholding `:utf8` from this. For myself I’m mostly concerned with the `use bytes`-ish behaviour of taking the PV internals rather than the string’s logical value.

ikegami:
That would be a good first step.

Leont:
> @Leont I’m happy to do this on CPAN … but how can I affect what exec et al. do in XS? Can you think of an existing module I could look at for an example?

In perl, set a %^H (hints hash) entry to enable the new behavior (this should work on any 5.10+)
In XS, set a PL_check entry for all the applicable opcodes. This will check for the hints hash entry, and if enabled in that scope will replace the pp function with your custom function. You custom function will downgrade/encode (a copy of?) the values and then call the original function.

I guess Lexical::Var and Lexical::Importer are examples I know of, but neither is a simple example.

FGasper:
I’m going to implement this via CPAN first as Sys::Binmode.

FGasper:
FYI: https://metacpan.org/pod/Sys::Binmode

PR #18613, opened 2021-03-01
Vulnerabilities in cpan/ modules still affect perl; we should handle them

karenetheridge:
This is a proposed wording change for perlsecpolicy.pod regarding issues in modules shipped under the cpan/ directory, which are dual-life but generally shipped to CPAN first and then copied here.

Since these modules are bundled with perl, any security issues with them can still be seen as security issues of perl itself, as we should not wash our hands of these issues entirely. It would be beneficial to still report the issue to the closed security list, which can then coordinate with the respective author(s) to get the issue fixed, or potentially if the module is not actively maintained, someone from p5p itself can take on the task of fixing the bug and submitting it upstream to the author (and potentially shipping the patched version in perl itself before the cpan version has been released).

### DIFF
--- a/pod/perlsecpolicy.pod
+++ b/pod/perlsecpolicy.pod
@@ -92,8 +92,10 @@ core Perl repository
 =back
 
 Files under the F<cpan/> directory in Perl's repository and release tarballs are
-developed and maintained independently. The Perl security team does not handle
-security issues for these modules.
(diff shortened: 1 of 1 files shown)


nwc10:
PSC reviewed this and I've cherry-picked this to blead (which github doesn't seem to have noticed yet)

Thanks for this. I'm sorry that it got missed.

PR #18619, opened 2021-03-03
Mark Data::Dumper output as UTF-8 when the regex requires it

arc:
When adding a `qr//` to the output, Data::Dumper extracts the SV of the pattern, but was failing to do anything with the UTF8 flag as it then copied it to the return value. This PR fixes that.

### DIFF
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -29,7 +29,7 @@ our ( $Indent, $Trailingcomma, $Purity, $Pad, $Varname, $Useqq, $Terse, $Freezer
 our ( @ISA, @EXPORT, @EXPORT_OK, $VERSION );
 
 BEGIN {
-    $VERSION = '2.176'; # Don't forget to set version and release
+    $VERSION = '2.177'; # Don't forget to set version and release
                         # date in POD below!
 
     @ISA = qw(Exporter);
@@ -1477,7 +1477,7 @@ modify it under the same terms as Perl itself.
 
 =head1 VERSION
 
-Version 2.175
+Version 2.177
 
 =head1 SEE ALSO
 
--- a/dist/Data-Dumper/Dumper.xs
+++ b/dist/Data-Dumper/Dumper.xs
@@ -651,6 +651,10 @@ dump_regexp(pTHX_ SV *retval, SV *val)
 
     assert(sv_pattern);
 
+    if (SvUTF8(sv_pattern)) {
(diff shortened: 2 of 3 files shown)


Leont:
LGTM

Grinnz:
It seems to be accounted for in the test but can you confirm that this would not cause problems for pre-5.14 Perls where the `/u` modifier doesn't exist?

arc:
This is:

- The dist/Data-Dumper directory from this branch
- With Makefile.PL etc from the last CPAN release
- And ppport.h from the latest Devel::PPPort release
- With the latest Test::More installed
- Under Perl 5.10.1

The test can't be made to pass on 5.8 (incomplete UTF-8 support), so I've updated this branch to skip it in that situation.

```
$ perl -wE 'say $^V'
v5.10.1
$ $ make test
PERL_DL_NONLAZY=1 /Users/aaron/.plenv/versions/5.10.1/bin/perl "-MExtUtils::Command::MM" "-e" "test_harness(0, 'blib/lib', 'blib/arch')" t/*.t
t/bless.t ..................... ok
t/bless_var_method.t .......... ok
t/bugs.t ...................... ok
t/deparse.t ................... ok
t/dumper.t .................... ok
t/dumpperl.t .................. ok
t/freezer.t ................... ok
t/freezer_useperl.t ........... ok
t/huge.t ...................... skipped: Need ~10 GiB of core for this test
t/indent.t .................... ok
t/misc.t ...................... ok
t/names.t ..................... ok
t/overload.t .................. ok
t/pair.t ...................... ok
t/perl-74170.t ................ ok
t/purity_deepcopy_maxdepth.t .. ok
t/qr.t ........................ ok
t/quotekeys.t ................. ok
t/recurse.t ................... ok
t/seen.t ...................... ok
t/sortkeys.t .................. ok
t/sparseseen.t ................ ok
t/terse.t ..................... ok
t/toaster.t ................... ok
t/trailing_comma.t ............ ok
t/values.t .................... ok
All tests successful.
Files=26, Tests=803,  1 wallclock secs ( 0.09 usr  0.05 sys +  0.53 cusr  0.10 csys =  0.77 CPU)
Result: PASS
```

khwilliamson:
On 3/3/21 3:14 PM, Aaron Crane wrote:
> The test can't be made to pass on 5.8 (incomplete UTF-8 support), so 
> I've updated this branch to skip it in that situation.

ppport.h has been enhanced in recent times to provide pretty good 
support back to 5.6.1.  But perhaps this lies outside that.

PR #18620, opened 2021-03-03
Add cop_hints_exists_* functions

### LINKED ISSUE #18615: [feature] cop_hints_exists_pvs macro
Per #p5p it would be nice to have a `cop_hints_exists_pvs` macro, since it’s common for modules that manipulate %^H not to care about the actual values in the hash.

Leont:
This adds a set of cop_hints_exists_{pv,pvn,pvs,sv} functions to support checking for the existence of keys in the hints hash of a specific cop without needing to create a mortal copy of said value.

This resolves #18615

### DIFF
--- a/cop.h
+++ b/cop.h
@@ -630,6 +630,58 @@ string/length pair.
     cophh_fetch_sv(CopHINTHASH_get(cop), key, hash, flags)
 
 /*
+=for apidoc Am|bool|cop_hints_exists_pvn|const COP *cop|const char *keypv|STRLEN keylen|U32 hash|U32 flags
+
+Look up the hint entry in the cop C<cop> with the key specified by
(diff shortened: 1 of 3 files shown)


leonerd:
This looks reasonable to me

cop.h
```diff
@@ -228,6 +229,58 @@ string/length pair.
 #define cophh_fetch_sv(cophh, key, hash, flags) \
     Perl_refcounted_he_fetch_sv(aTHX_ cophh, key, hash, flags)
 
+/*
+=for apidoc Amx|SV *|cophh_exists_pvn|const COPHH *cophh|const char *keypv|STRLEN keylen|U32 hash|U32 flags
```
FGasper: Shouldn’t this say `bool` rather than `SV *` for its return type? @Leont

Leont: Yeah, updated that

tonycoz (approved):

PR #18623, opened 2021-03-05
Use IsForeign throughout example

briandfoy:
The example used both IsForeign and Is_Foreign, so I chose the one without the underscore


khwilliamson:
I did read many years ago that not having an underscore made things significantly harder for native speakers of some languages.

But no one seems to have paid much heed.

Issue #18627, opened 2021-03-07
[doc] Missing information about Perl's default sort algorithm


rsFalse:
**Where**

https://perldoc.perl.org/functions/sort
https://perldoc.perl.org/sort
v5.32

**Description**

I think it would be nice to have one sentence in documentation which says which algorithm is used by `sort` function.

dur-randir:
Btw, it's currently noted in perlsec (which isn't an obvious place for such information, but it's still present).

tonycoz:
Fixed by 4e8fb2ba8ecc1c3307d7d4dbd776fe985723f9a3

PR #18629, opened 2021-03-07
Identify the algorithm which 'sort' uses: mergesort

jkeenan:
For:  https://github.com/Perl/perl5/issues/18627

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -7734,6 +7734,9 @@ eliminate any C<NaN>s from the input list.
 
     my @result = sort { $a <=> $b } grep { $_ == $_ } @input;
 
+In this version of F<perl>, the C<sort> function is implemented via the
+mergesort algorithm.
+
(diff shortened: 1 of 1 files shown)


rwp0:
linking to [Wikipedia](https://en.wikipedia.org/wiki/Merge_sort) as `L<mergesort|https://en.wikipedia.org/wiki/Merge_sort>` would be helpful, i think

tonycoz (approved):
I think this is reasonable.

I don't think the algorithm we use is likely to change any time soon, so I don't think we need to add a "this may change at any time" type of warning.

jkeenan:
> I think this is reasonable.
> 
> I don't think the algorithm we use is likely to change any time soon, so I don't think we need to add a "this may change at any time" type of warning.

Merged into blead locally and pushed to origin in 4e8fb2ba8e.

Thank you very much.
Jim Keenan

PR #18633, opened 2021-03-09
Remove trailing whitespace

hginzel:
`# `[`How to remove trailing whitespace of all files recursively?`](https://stackoverflow.com/questions/149057/how-to-remove-trailing-whitespace-of-all-files-recursively)
```sh
export LC_CTYPE=C
export LANG=C
find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | perl -0ne 'print if -T' | xargs -0 sed -Ei 's/[[:blank:]]+$//'
```


Tux:
The trailing whitespace in `Configure` stems from units from `dist` that have never been changed. It *is* possible to remove it, but I have never seen this as only modification from the original code reason enough to make it a modified unit.

FYI all modified units that did have trailing whitespace were cleaned up and a pull request to remove it in the upstream versions too was accepted. Note however that this was done in a process that required way more work than just stripping trailing whitespace.

cpan/Digest-MD5/MD5.pm
```diff
@@ -258,12 +258,12 @@ The above example would print out the message:
 The same checksum can also be calculated in OO style:
 
     use Digest::MD5;
-
```
Leont, line 261: This change is wrong: removing this whitespace will cut the verbatim block into two

Leont:
Files under `cpan/` should not be modified unless there's a pressing reason. Instead patches should be sent to their upstream.

Grinnz:
My recommendation would be to work up independent PRs for different top level directories, as they each tend to have different concerns and these would likely be easier to review, should core contributors determine this worthwhile. (particularly, don't do it for `cpan/` as mentioned)

Leont:
Also, this is failing 37 test files. You really want to double check if you didn't break anything when mass-editing a large codebase.

jkeenan:
> `# `[`How to remove trailing whitespace of all files recursively?`](https://stackoverflow.com/questions/149057/how-to-remove-trailing-whitespace-of-all-files-recursively)
> 
> ```shell
> export LC_CTYPE=C
> export LANG=C
> find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | perl -0ne 'print if -T' | xargs -0 sed -Ei 's/[[:blank:]]+$//'
> ```

This merge request should be rejected.  It is unwise to touch so many files in one pass for so little benefit.

We would be much better off doing these modifications in much smaller batches.

Thank you very much.
Jim Keenan

Leont:
> This merge request should be rejected. It is unwise to touch so many files in one pass for so little benefit.

I agree. An operation like this requires a much more careful approach than running a shell one-liner on the entire codebase.

Issue #18628, opened 2021-03-07
try/catch visibility to caller()


leonerd:
There's an odd quirk about `eval {}` blocks, in that they are visible to `caller()`, and hence turn up in the stack traces thrown by Carp::longmess and friends:

```perl
use Carp;
sub C { Carp::cluck("Boo") }
sub B { eval { C() } } 
sub A { eval { B() } }
A();
__END__

Boo at - line 2.
        main::C() called at - line 3
        eval {...} called at - line 3
        main::B() called at - line 4
        eval {...} called at - line 4
        main::A() called at - line 5
```

That is sortof understandable if you consider that `caller()` is supposed to tell us about successive levels of the "caller stack", where `return` would send us.

But how `caller()` works internally is that it looks down the context stack - which in addition to sub calls and `eval {}` frames, also contains information about `foreach` loops and other fun things. Those are invisible to `caller()`.

A quirk of its current (experimental) implementation means that `try/catch` blocks are also invisible to `caller()`:

```perl
use experimental 'try';
use Carp;
sub C { Carp::cluck("Boo") }
sub B { try { C() } catch($e) {} }
sub A { try { B() } catch($e) {} }
A();
__END__

Boo at - line 3.
        main::C() called at - line 4
        main::B() called at - line 5
        main::A() called at - line 6
```

I think this is a useful property for them to have.

However, I didn't document or unit-test for it [1]. I'd like to propose adding tests and documentation of that to ensure this property remains.

Do we agree?

(cross-posted to perl5-porters@: https://www.nntp.perl.org/group/perl.perl5.porters/2021/03/msg259240.html)

[1]: Actually until I wrote that example just now, I thought they *would* be visible so this mail was going to propose changing it.

leonerd:
With #18634 merged this is now done.

PR #18634, opened 2021-03-09
Document and test that caller() does not see try{} blocks 

leonerd:
As per #18628

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -989,6 +989,11 @@ L<C<undef>|/undef EXPR> if L<C<%^H>|perlvar/%^H> was empty.  Do not
 modify the values of this hash, as they are the actual values stored in
 the optree.
 
+Note that the only types of call frames that are visible are subroutine
+calls and C<eval>. Other forms of context, such as C<while> or C<foreach>
+loops or C<try> blocks are not considered interesting to C<caller>, as they
+do not alter the behaviour of the C<return> expression.
+
 Furthermore, when called from within the DB package in
 list context, and with an argument, caller returns more
 detailed information: it sets the list variable C<@DB::args> to be the
--- a/pod/perlsyn.pod
+++ b/pod/perlsyn.pod
@@ -666,6 +666,13 @@ containing function to return.
         }
     };
 
+As with other control-flow syntax, C<try> blocks are not visible to
+C<caller()> (just as for example, C<while> or C<foreach> loops are not).
+Successive levels of the C<caller> result can see subroutine calls and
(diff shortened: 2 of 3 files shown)


pod/perlfunc.pod
```diff
@@ -989,6 +989,11 @@ L<C<undef>|/undef EXPR> if L<C<%^H>|perlvar/%^H> was empty.  Do not
 modify the values of this hash, as they are the actual values stored in
 the optree.
 
+Note that the only types of call frames that are visible are subroutine
+calls and C<eval>. Other forms of context, such as C<while> or C<foreach>
+loops or C<try> blocks are not considered interesting to C<caller>, as they
+do not alter the behaviour of the C<return> expression.
```
happy-barney, line 995: I'd use (your) sentence "caller() tells you where return would go".

It feels much more cleaner.

leonerd:
@dur-randir You have downvoted the initial comment but not given any indication why.

Could you perhaps give some more wording here to explain your thoughts?

dur-randir:
>Could you perhaps give some more wording here to explain your thoughts?


It's on the list discussion.

leonerd:
@dur-randir 
> It's on the list discussion.

I see; so it is less that you are objecting to the tests and documentation this PR in itself, and more the underlying behaviour that they describe? That behaviour is in bleadperl already - if you think it should be changed perhaps you could make a counter-proposal and suggest that?

This particular PR is only documenting and testing behaviour which is *already* in the code - objecting to this PR won't prevent the behaviour from occurring.

Leont:
If I understand things correctly, `caller` and `return` are in agreement. That sounds fine by me.

leonerd:
@Leont 
> If I understand things correctly, `caller` and `return` are in agreement. That sounds fine by me.

Indeed they are. This PR just adds some documentation and unit-tests for that fact. I didn't think about `caller` in my original implementation, and observed it just happens to work by happy coïncidence.

shadowcat-mst:
@dur-randir caller() seems to generally return 'the thing we would return from' - so eval {} shows up as a frame because 'return' inside an eval only returns from the eval. OTOH 'return' inside a try returns from the containing subroutine, so that's the frame. 

So if https://www.nntp.perl.org/group/perl.perl5.porters/2021/03/msg259257.html is your objection, then I think you're mistaken - they're *not* semantically identical and while they *do* have different behaviour, that's considered a feature of 'try' and both of them behave the same wrt return/caller - as does basically everything else, because caller/return share the relevant logic in the internals so to make them disagree with each other would require special casing one of the two.

As such, I think your worries are well-intentioned but misplaced.

PR #18637, opened 2021-03-10
Fix perlunifaq’s “What if I don't encode?”

FGasper:
I’d ideally like to describe the non-filehandle case as “this is a bug; don’t depend on it”, but at least for now it’s good to fix the information about filehandle output.

### DIFF
--- a/pod/perlunifaq.pod
+++ b/pod/perlunifaq.pod
@@ -63,11 +63,26 @@ positive, but it's best to avoid it.
 
 =head2 What if I don't encode?
 
-Your text string will be sent using the bytes in Perl's internal format. In
-some cases, Perl will warn you that you're doing something wrong, with a
-friendly warning:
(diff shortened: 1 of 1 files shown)


Grinnz:
Seems like an appropriate correction to me. I'd prefer more emphasis on the fact that the use of the internal format for syscalls is a bug, and the "wide character" case also uses the internal format which happens to always approximate UTF-8. (I also think there are other cases where <255 strings are used correctly but the internal format is used otherwise but can't think of them currently.)

FGasper:
> I'd prefer more emphasis on the fact that the use of the internal format for syscalls is a bug

That would be my preference as well; I just wasn’t sure how appropriate it is to describe it thus in official documentation.

khwilliamson:
This is better than what we have, so I merged it.  But further improvements are welcome

PR #18638, opened 2021-03-10
Remove duplicate call to init_globals()

jkeenan:
The first call was added in a 2003 refactoring of the code in commit
99cb6bd822.  It doesn't do anything other than what the second, older
invocation does, so it's superfluous and deletable.

### DIFF
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -2,7 +2,7 @@ package Pod::Html;
 use strict;
 require Exporter;
 
-our $VERSION = 1.26;
+our $VERSION = 1.27;
 our @ISA = qw(Exporter);
 our @EXPORT = qw(pod2html htmlify);
 our @EXPORT_OK = qw(anchorify relativize_url);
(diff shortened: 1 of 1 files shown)


jkeenan:
In `ext/Pod-Html/lib/Pod/Html.pm`, internal subroutine `init_globals()` is called twice, which appears to be unnecessary.
```
259 my $Curdir = File::Spec->curdir;
260 
261 init_globals();
262 
263 sub init_globals {
...
291     $Title = undef;             # title to give the pod(s)
292 }
293 
294 sub pod2html {
295     local(@ARGV) = @_;
296     local $_;
297 
298     init_globals();
299     parse_command_line();
300 
```
Other than the declaration of `init_globals()`, nothing happens between these two calls.  Of these two invocations, the first is newer.  It was added in a refactoring of the code in 2003 in this commit:
```
commit 99cb6bd82261c88aa83ee7d6e9447ada3a00a02e
Author:     Michael G. Schwern <schwern@pobox.com>
AuthorDate: Sun Aug 17 19:44:31 2003 -0700
Commit:     Jarkko Hietaniemi <jhi@iki.fi>
CommitDate: Mon Aug 18 08:46:52 2003 +0000

    Pod::Html  unifying global variable declaration and initialization
```
... but the second commit was left in.  I think the second invocation is the more logically positioned, so I recommend removing the first.  After this change, all tests still PASS.

Thank you very much.
Jim Keenan

oodler577 (approved):
Looks good to me. Even better the net number of lines added is in the negative :)

tonycoz (approved):

PR #18640, opened 2021-03-13
Eliminate double assignment from @_

jkeenan:
In lib/Pod/Html.pm's get_cache(), the arguments from @_ were being read
into the file twice:  once assigned to four distinct scalars, once to an
array.  That array was in turn fed into an internal subroutine -- which
within itself assigned its elements to four distinct scalars.

    sub get_cache {
        my($dircache, $podpath, $podroot, $recurse) = @_;
        my @cache_key_args = @_;

This approach may have made sense back in 1997 when the code first
entered the core distribution.  Some of the four scalars were and are
used within get_cache(), while the array was at that time provided as
argument for two internal subroutines.

This approach, however, is of limited value today.  We will likely to
want to bundle up all these lexical variables into a hash or an object
and just pass a single reference to internal subroutines.  So let's
eliminate the double assignment and eliminate one variable.

Increment $VERSION.

### DIFF
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -2,7 +2,7 @@ package Pod::Html;
 use strict;
 require Exporter;
 
-our $VERSION = 1.26;
+our $VERSION = 1.27;
 our @ISA = qw(Exporter);
 our @EXPORT = qw(pod2html htmlify);
 our @EXPORT_OK = qw(anchorify relativize_url);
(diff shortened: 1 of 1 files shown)


jkeenan:
The problem can be seen by building this demonstration branch: jkeenan/pod-html-demonstrate-dupe-assignment-20210312 -- and then running:
```
cd t; TEST_JOBS=1 ./perl harness ../ext/Pod-Html/t/*.t; cd -
```
Thank you very much.
Jim Keenan

tonycoz (approved):

jkeenan:
Merged locally in commit 165baf7af4 and pushed to blead.

Thank you very much.
Jim Keenan

Issue #14706, opened 2015-05-16
Infinite loop 'while( each { 1=>'a' } )'


p5pRT:
Migrated from [rt.perl.org#125205](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=125205) (status was 'rejected')

Searchable as RT125205$

p5pRT:
### From @KES777

This is a bug report for perl from kes\-kes@&#8203;yandex\.ru\,
generated with the help of perlbug 1\.39 running under perl 5\.18\.2\.

\#\!/usr/bin/perl

use Modern&#8203;::Perl;

my $val = \{ 1 => 'a' \};

\#test 1 OK
while\( my\( $key\, $value \) =  each %\{ $val \} \) \{
&nbsp;    say "$key $value";
\}

\#test 2 OK \(Why this work if "Type of argument to each on reference must be unblessed hashref or arrayref at \./test\.pl line 18\."
while\( my\( $key\, $value \) =  each %$val \) \{
&nbsp;    say "$key $value";
\}

\#test 3 FAIL not Expected
\#while\( my\( $key\, $value \) =  each \( 1 => 'a' \) \) \{
\#    say "$key $value";
\#\}

\#test4 OK as Expected \(see perldoc each\)
my %hash =  %\{ \{ 1 => 'a' \} \}; \#same hash as in prev example&#8203;: %hash =  \( 1 => 'a' \)
while\( my\( $key\, $value \) =  each %hash \) \{
&nbsp;    say "$key $value";
\}

\#test 5 FAIL not expected \(infinite loop\)
while\( my\( $key\, $value \) =  each %\{ \{ 1 => 'a' \} \} \) \{
&nbsp;    say "$key $value";
\}

\#test 6 FAIL not Expected \(infinite loop\)
while\( my\( $key\, $value \) =  each \{ 1 => 'a' \} \) \{
&nbsp;    say "$key $value";
\}


\_\_END\_\_

May you tell me please where I can view status of this BUG report?



---
Flags&#8203;:
&nbsp;    category=core
&nbsp;    severity=critical


---
Site configuration information for perl 5\.18\.2&#8203;:

Configured by Debian Project at Thu Mar 27 18&#8203;:28&#8203;:21 UTC 2014\.

Summary of my perl5 \(revision 5 version 18 subversion 2\) configuration&#8203;:
&nbsp;   
&nbsp;  Platform&#8203;:
[...]

Issue #8051, opened 2005-08-02
each operator loops indefintely on anonymous hash ref


p5pRT:
Migrated from [rt.perl.org#36756](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=36756) (status was 'rejected')

Searchable as RT36756$

p5pRT:
### From siegel@zrz.tu-berlin.de


This is a bug report for perl from anno4000@&#8203;lublin\.zrz\.tu\-berlin\.de\,
generated with the help of perlbug 1\.34 running under perl v5\.8\.1\.




---
When the each operator is applied to an anonymous hash ref like this

&nbsp;    while \( my \( $k\, $v\) = each %\{ \{ a => 1\} \} \) \{ print "$k $v\\n" \}

an endless loop results where one key/value pair is repeatedly returned\.

In test form&#8203;:

&nbsp;    use Test&#8203;::More;
&nbsp;    plan tests => 1;

&nbsp;    my $n = 0;
&nbsp;    while \( each %\{ \{ a => 1 \}\} \) \{
&nbsp;        \+\+ $n \< 2 or last;
&nbsp;    \}
&nbsp;    is $n\, 1\, 'each on anonymous hashref';

Anno




---


---
Flags&#8203;:
&nbsp;    category=core
&nbsp;    severity=low


---
Site configuration information for perl v5\.8\.1&#8203;:

Configured by anno4000 at Wed Oct  1 12&#8203;:15&#8203;:29 MEST 2003\.

Summary of my perl5 \(revision 5\.0 version 8 subversion 1\) configuration&#8203;:
&nbsp;  Platform&#8203;:
&nbsp;    osname=linux\, osvers=2\.0\.36\, archname=i586\-linux
&nbsp;    uname='linux lublin\.zrz\.tu\-berlin\.de 2\.0\.36 \#2 mon jun 26 12&#8203;:24&#8203;:02 mest 2000 i586 unknown '
&nbsp;    config\_args='\-des'
&nbsp;    hint=recommended\, useposix=true\, d\_sigaction=define
&nbsp;    usethreads=undef use5005threads=undef useithreads=undef usemultiplicity=undef
&nbsp;    useperlio=define d\_sfio=undef uselargefiles=define usesocks=undef
&nbsp;    use64bitint=undef use64bitall=undef uselongdouble=undef
&nbsp;    usemymalloc=n\, bincompat5005=undef
&nbsp;  Compiler&#8203;:
[...]

PR #18643, opened 2021-03-15
warn when using each on an anonymous hash

tonycoz:
We've had at least three tickets (#18557, #14706, #8051 from a quick search) over the years where the user has been confused by the behaviour of each on an anonymous array or hash, there's no way to tell if other users have been struck by the same issue.

This change checks the op tree during compilation to try and detect the simple cases, I don't see a way to tell if the hash reference was returned by a function or similar.

If this receives positive feedback (on the idea anyway), I'll add the same check for each on anonymous arrays.

I welcome alternatives to the warning message (or opinions if you think it's fine.)

### DIFF
--- a/op.c
+++ b/op.c
@@ -15350,6 +15350,22 @@ Perl_ck_tell(pTHX_ OP *o)
     return o;
 }
 
+PERL_STATIC_INLINE OP *
+S_last_non_null_kid(OP *o) {
+    OP *last = NULL;
+    if (cUNOPo->op_flags & OPf_KIDS) {
+        OP *k = cLISTOPo->op_first;
+        while (k) {
+            if (k->op_type != OP_NULL) {
+                last = k;
+            }
+            k = OpSIBLING(k);
+        }
+    }
+
+    return last;
+}
+
 OP *
 Perl_ck_each(pTHX_ OP *o)
 {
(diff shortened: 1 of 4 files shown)


Leont:
This sounds reasonable

hvds:
> If this receives positive feedback (on the idea anyway), I'll add the same check for each on anonymous arrays.

Looks good to me too. The 'syntax' class for the warning raised an eyebrow, but looking at existing warnings in that class this seems to fit right in.

tonycoz:
> The 'syntax' class for the warning raised an eyebrow,...

I had trouble selecting a category.

The issue warned about is detected by the syntax, so I selected syntax.  I thought about adding a new category but didn't see an obvious name to use.

tonycoz:
added support for warning on each @{ [ ... ] } and handle more the {} compiling to an enter/leave subtree.

oodler577:
I seem to recall the use of `each` used to be discouraged in the documentation, am I misremembering? The reason is that it has a stateful iterator and if anything causes confusion, it's *that*.

tonycoz:
> I seem to recall the use of `each` used to be discouraged in the documentation, am I misremembering? The reason is that it has a stateful iterator and if anything causes confusion, it's _that_.

perlfunc seems pretty discouraging to me:

```
Thus all uses of C<each> on a single hash or array advance the same
iterator location.  All uses of C<each> are also subject to having the
iterator reset by any use of C<keys> or C<values> on the same hash or
array, or by the hash (but not array) being referenced in list context.
This makes C<each>-based loops quite fragile: it is easy to arrive at
such a loop with the iterator already part way through the object, or to
accidentally clobber the iterator state during execution of the loop body.
It's easy enough to explicitly reset the iterator before starting a loop,
but there is no way to insulate the iterator state used by a loop from
the iterator state used by anything else that might execute during the
loop body.  To avoid these problems, use a C<foreach> loop rather than
C<while>-C<each>.
```

but this change doesn't discourage or encourage each() - it just points out what is apparently a common mistake - we've had three people confused enough by the behaviour to report it as a bug, how many have just been confused and not reported it? (or decided perl is broken)

Also this warns on behaviour different to what `perlfunc -f each` discusses - perlfunc talks about different things that can reset (or otherwise) the iterator on a given hash, but for this PR and the tickets it refers to the problem is calling each on a new hash for each each() call.

jkeenan:
@tonycoz could you resolve the merge conflict in this p.r.?

We then need to locate a reviewer.

Thank you very much.
Jim Keenan

tonycoz:
> @tonycoz could you resolve the merge conflict in this p.r.?

Done, it was purely whitespace.

jkeenan:
> > @tonycoz could you resolve the merge conflict in this p.r.?
> 
> Done, it was purely whitespace.

Thanks. @iabyn, will you be able to review this p.r.?

iabyn:
LGTM

Issue #18639, opened 2021-03-11
(IV_MIN << 0) returns zero under "use integer"


t-a-k:
**Description**
`$x << 0` results in `0` if `use integer` is in force and `$x` is equal to `IV_MIN`.

**Steps to Reproduce**
```
$ perl -wle 'my $x = -(~0 >> 1) - 1; print $x; { use integer; print $x << 0 }'
-2147483648
0
```
`$x` shall be equal to `IV_MIN` here.

**Expected behavior**
`$x << 0` should return `$x` for any IVs.

**Perl configuration**
```
Summary of my perl5 (revision 5 version 33 subversion 8) configuration:
  Commit id: cb174a785cca1bace5d049d08be208784028e671
  Platform:
    osname=linux
    osvers=4.19.0-14-686-pae
    archname=i686-linux
    uname='linux x41 4.19.0-14-686-pae #1 smp debian 4.19.171-2 (2021-01-30) i686 gnulinux '
    config_args='-de -Dusedevel -DEBUGGING'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=undef
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2 -g'
    cppflags='-fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='8.3.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=4
    doublesize=8
    byteorder=1234
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=12
    longdblkind=3
    ivtype='long'
    ivsize=4
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=4
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags =' -fstack-protector-strong -L/usr/local/lib'
[...]

PR #18644, opened 2021-03-16
Fix GH #18639

khwilliamson:


tonycoz:
The change itself looks reasonable, but it could use a test and a better commit message (I can do that if you like.)

khwilliamson:
This was fixed by 4e82c85b1c9c9b30253b8624470da6f20a6c0604 which added a test.

Technically there is still the possibility of undefined behavior.  I should technically test for a zero shift and return without shifting.  But clang ASAN does not detect this as undefined, and I can't imagine actual hardware causing UB.  I have emailed someone who would really know, and will update this as necessary depending on the response.

tonycoz:
> Technically there is still the possibility of undefined behavior.

I don't see any way to get undefined behaviour out of this code.

We only shift when shift >=0 and < IV_BITS, so we don't hit that undefined behaviour.

We only shift unsigned values, which avoids the negative value undefined behaviours, and shifting of unsigned types is well defined when the shift count is in range. (and right shift even for negative signed values for an in range shift count is implementation defined, not undefined)

PR #18650, opened 2021-03-18
add a simple default-enabled autovivification feature

tonycoz:
Unlike other features, the work for this is done at runtime, but we only get to this point when autovivification is needed, so I expect the total cost is small enough not to matter.

The only real way to eliminate all the cost would be to have alternate ops that don't autovivify, since an op flag would have about the same runtime cost.

The runtime-ness of this feature revealed a bug in my original work that reflected feature changes to %^H to PL_compiling.cop_features - the cop_features value wasn't being copied to the next/dbstate op.  This was (fortunately) harmless since current features only work at compile-time (using PL_compiling), and other feature tests that anyone outside of core might want to do need to check the hinthash value from caller().

### DIFF
--- a/regen/feature.pl
+++ b/regen/feature.pl
@@ -41,6 +41,7 @@ my %feature = (
     multidimensional => 'multidimensional',
     bareword_filehandles => 'bareword_filehandles',
     try             => 'try',
+    autovivification => 'autovivification',
 );
 
 # NOTE: If a feature is ever enabled in a non-contiguous range of Perl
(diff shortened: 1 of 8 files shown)


regen/feature.pl
```diff
@@ -824,6 +826,14 @@ =head2 The 'try' feature.
 
 For more information, see L<perlsyn/"Try Catch Exception Handling">.
 
+=head2 The 'autovivification' feature.
+
+This feature enables autovivification of references.  It is enabled by
+default, but can be turned off to disable it.
+
+This feature is available under this name from Perl 5.34 onwards, in
+previous versions it was simple on all the time.
```
Grinnz, line 835: simply*

xenu:
Personally, I expected this to work like `no autovivification` instead of throwing an exception.

Leont:
> Personally, I expected this to work like `no autovivification` instead of throwing an exception.

Yeah. I'm not sure we really need this feature, but fatal non-autovivification is definitely not something I think we need.

tonycoz:
I was mostly thinking of it as a strictness check, throwing when perl would otherwise autovivify, as if the variable was a nil object or similar.

I've made this PR draft for now.

PR #18652, opened 2021-03-20
ExtUtils::ParseXS fix obvious error-message bug

### LINKED ISSUE #16916: ExtUtils::ParseXS release, plus bug with clang++ on 5.14-5.16 on EUPXS 3.04_01+
Migrated from [rt.perl.org#133973](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133973) (status was 'open')

Searchable as RT133973$

mohawk2:
Fix #16916

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -3639,6 +3639,7 @@ dist/ExtUtils-ParseXS/t/lib/PrimitiveCapture.pm			Primitive STDOUT/ERR capturing
 dist/ExtUtils-ParseXS/t/lib/TypemapTest/Foo.pm			ExtUtils::Typemaps tests
 dist/ExtUtils-ParseXS/t/pseudotypemap1				A test-typemap
 dist/ExtUtils-ParseXS/t/typemap					Standard typemap for controlled testing
+dist/ExtUtils-ParseXS/t/XSBroken.xs				Test file for ExtUtils::ParseXS tests
 dist/ExtUtils-ParseXS/t/XSInclude.xsh				Test file for ExtUtils::ParseXS tests
 dist/ExtUtils-ParseXS/t/XSMore.xs				Test file for ExtUtils::ParseXS tests
 dist/ExtUtils-ParseXS/t/XSTest.pm				Test file for ExtUtils::ParseXS tests
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -912,7 +912,7 @@ EOF
   #-Wall: if there is no $self->{Full_func_name} there are no xsubs in this .xs
   #so 'file' is unused
   print Q(<<"EOF") if $self->{Full_func_name};
-##if PERL_VERSION_LT(5, 9, 0)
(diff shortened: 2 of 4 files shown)


jkeenan:
> Fix #16916

The hesitation I have about applying this patch is that currently our test suite does not appear to exercise the "Error: No INPUT definition for type" message which your patch corrects.  Consequently we have no way of demonstrating that the patch improves things.

I hacked up `dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm` --specifically `generate_init()` with `print` statements and ran the test suite.  I was unable to locate any tests where the condition for entering the `if` block were met, *i.e.,* `$inputmap` was always defined.

In your investigations preceding https://github.com/Perl/perl5/issues/16916#issuecomment-544100912, were you able to get the error-message to trigger?  If so, could we re-work that into a unit test that we could use before applying the patch to illustrate the problem, then after applying to demonstrate the solution?

Thank you very much.
Jim Keenan

mohawk2:
@jkeenan I've now added a test, which hopefully helps.

An issue: 63b6fce5821cf1310f5541becaca9834b2a08f31 introduces `PERL_VERSION_LT`, which doesn't exist on older Perls. The now-first commit here changes it to use `PERL_VERSION_LE` which already existed.

After that's resolved (perhaps by merging this), the first thing mentioned in #16916 still applies: the blead version of EU::PXS is 3.42. The CPAN version is 3.35, from 2017. @tsee @khwilliamson (and p5p) have permissions to upload. Could this be fixed?

jkeenan:
> @jkeenan I've now added a test, which hopefully helps.
> 

It does, indeed.  I checked out your branch locally, temporarily reverted the change in ParseXS.pm, built perl and ran `t/001-basic.t`.  It failed as expected.  I then restored the change in ParseXS.pm, rebuilt and re-ran the test.  It now passed, as intended.

(The test file uses a testing subroutine and module credited to me in 2010 that I had completely forgotten about.  Perhaps that's because I got it from @xdg!)

> An issue: [63b6fce](https://github.com/Perl/perl5/commit/63b6fce5821cf1310f5541becaca9834b2a08f31) introduces `PERL_VERSION_LT`, which doesn't exist on older Perls. The now-first commit here changes it to use `PERL_VERSION_LE` which already existed.
> 

As to this internal variable, I'll hope that @tonycoz or @tsee takes a look at that, as that's outside my expertise.  Otherwise, +1 from me.
> After that's resolved (perhaps by merging this), the first thing mentioned in #16916 still applies: the blead version of EU::PXS is 3.42. The CPAN version is 3.35, from 2017. @tsee @khwilliamson (and p5p) have permissions to upload. Could this be fixed?

I have applied a 'cpan-dual-life' label to both this p.r. and to the original bug ticket.  That may help.

Thank you very much.
Jim Keenan

mohawk2:
> > @jkeenan I've now added a test, which hopefully helps.
> 
> It does, indeed. I checked out your branch locally, temporarily reverted the change in ParseXS.pm, built perl and ran `t/001-basic.t`. It failed as expected. I then restored the change in ParseXS.pm, rebuilt and re-ran the test. It now passed, as intended.

As I checked (but am very pleased to have confirmed; more eyes = better). Thanks!

Fingers crossed those mentioned will be able to take a look soon enough.

tonycoz:
It looks good to me.

PR #18656, opened 2021-03-22
Document another case when the "not imported" warning is used

choroba:
The minimal example:

  use strict;
  { our $x; }
  print $x;


choroba:
Do we the warning to be thrown in this case at all?
Investigated because of https://www.perlmonks.org/?node_id=11130084

tonycoz:
Or a my variable whose scope has ended.

I wonder if it's time for this message to change its wording.

choroba:
@tonycoz Can you share an example of the lexical variable case? I can't reproduce it.

tonycoz:
> @tonycoz Can you share an example of the lexical variable case? I can't reproduce it.

You're right, I was thinking of the error from strict vars.

PR #18659, opened 2021-03-24
setsockopt: document OPTVAL being inadvertantly treated as a string

tonycoz:
fixes (as much as is possible) #18642


jkeenan:
> fixes (as much as is possible) #18642

@tonycoz, there's been no discussion of this p.r. in 15 months.  Do you still want us to pursue it?

tonycoz:
We can distinguish between originally-number and originally-string now, so I'll look at a fix rather than documenting the horribleness.

PR #18660, opened 2021-03-24
fix magic and upgraded SV handling for setsockopt()'s OPTVAL

tonycoz:
The code here checked SV flags before fetching magic, potentially
getting confused if magic fetched changed flags.

This also fixes handling for upgraded SVs, but I'm not sure that can
be tested sufficiently portably.

### DIFF
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -2707,13 +2707,14 @@ PP(pp_ssockopt)
     case OP_SSOCKOPT: {
             const char *buf;
             int aint;
+            SvGETMAGIC(sv);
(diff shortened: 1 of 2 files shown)


FGasper:
@tonycoz FWIW, here is my test for handling of setsockopt upgraded SV: https://github.com/FGasper/p5-Sys-Binmode/blob/master/t/net_setsockopt.t

As you mention, it’s a bit cross-platform-rickety. It seems to pass CPAN testers reliably, but of course there are many platforms that that doesn’t test.

PR #18662, opened 2021-03-24
Provide strict compliance for DynaLoader and XSLoader

Leont:
This branch is an alternative to #18221

The main difference between that branch and this is that this one will not break bootstrap files, as described [here](https://github.com/Perl/perl5/pull/18221#issuecomment-705674791).

### DIFF
--- a/dist/XSLoader/t/XSLoader.t
+++ b/dist/XSLoader/t/XSLoader.t
@@ -12,8 +12,6 @@ BEGIN {
         die "Test::More not available\n";
     }
 
-    plan(skip_all => "these tests needs Perl 5.5+") if $] < 5.005;
-
     use Config;
     foreach (qw/SDBM_File GDBM_File ODBM_File NDBM_File DB_File/) {
         if ($Config{extensions} =~ /\b$_\b/) {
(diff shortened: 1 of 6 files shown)


dist/XSLoader/t/XSLoader.t
```diff
@@ -1,6 +1,8 @@
 #!perl -T
 
 use strict;
+use warnings;
```
tonycoz, line 4: This introduces warning during the test process:
```
Subroutine XSLoader::bootstrap_inherit redefined at t/XSLoader.t line ***20.
Subroutine XSLoader::bootstrap_inherit redefined at t/XSLoader.t line ***42.
```

tonycoz (changes requested):
Looks good except for the warnings in `XSLoader.t`, which probably just need `no warnings "redefine";`

dist/XSLoader/t/XSLoader.t
Leont: Good catch, fixed that.

tonycoz (approved):

jkeenan:
Given @tonycoz's approval, locally rebased on blead, tested, merged into blead, re-tested, pushed to origin.  Closing this p.r.

Thank you very much.
Jim Keenan

Issue #18664, opened 2021-03-26
Does Perl formally support exponential notation for octal numbers? 


briandfoy:
**Description**

perldata does not mention a octal floating point form, but apparently it works:

```
$ perl5.32.0 -le 'print 01.4441766521041321410646114314242700334p+1'
3.14159265358979
$ perl5.33.8 -le 'print 0o1.4441766521041321410646114314242700334p+1'
3.14159265358979
```

Is this intended? If so, it should be added to the list of scalar value constructors.

If this is not intended, I guess it should stop doing that.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 32 subversion 1) configuration:
   
  Platform:
    osname=darwin
    osvers=19.3.0
    archname=darwin-2level
    uname='darwin otter.local 19.3.0 darwin kernel version 19.3.0: thu jan 9 20:58:23 pst 2020; root:xnu-6153.81.5~1release_x86_64 x86_64 '
    config_args='-des -Dprefix=/usr/local/perls/perl-5.32.1'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='cc'
    ccflags ='-fno-common -DPERL_DARWIN -mmacosx-version-min=10.15 -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -DPERL_USE_SAFE_PUTENV'
    optimize='-O3'
    cppflags='-fno-common -DPERL_DARWIN -mmacosx-version-min=10.15 -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.62)'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
[...]

PR #18665, opened 2021-03-26
Document octal literals

briandfoy:
If this is officially supported (#18664), here's a patch for the docs.

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -441,12 +441,13 @@ integer formats:
  .23E-10             # a very small number
  3.14_15_92          # a very important number
  4_294_967_296       # underscore for legibility
+ 0b011011            # binary
+ 0377                # octal (only digits, begins with 0)
+ 0o12_345            # alternative octal (introduced in Perl 5.33.5)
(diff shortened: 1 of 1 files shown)


tonycoz:
Did you want to cover binary floating point too?

briandfoy:
Yeah, binary would be fine. I'm waiting for a ruling from somebody that commits Perl to supporting these.

briandfoy:
There's no ruling on this or #18664, so I'll close this rather than let it languish.

khwilliamson:
I think it's premature to close this.  

There are higher priority items on the plates of quite a few people at the moment.  Leaving it open will cause it not to fall through the cracks.

The yearly code freeze is nearly upon us.  This is an item that can be deferred until after that deadline.

toddr:
Actually, aren't doc fixes allowed during this window @khwilliamson ?

briandfoy:
The issue is still open. It's just this PR is closed. I'm not going to keep rebasing it and so on.

khwilliamson:
Yes, which is a reason to keep this in mind

PR #18666, opened 2021-03-28
Enable warnings on «use v5.35»

### LINKED ISSUE #17162: Proposal for enabling warnings with 'use v5.32'
Migrated from [rt.perl.org#134484](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=134484) (status was 'open')

Searchable as RT134484$

Leont:
This is a proof of concept for enabling warnings by default in some future version of perl

~~~shell
leont@leon-desktop:~/Code/perl$ perl -e 'my $foo; print "$foo"'
leont@leon-desktop:~/Code/perl$ perl -e 'use 5.035; my $foo; print "$foo"'
Use of uninitialized value $foo in string at -e line 1.
~~~

This fixes #17162

### DIFF
--- a/lib/warnings.pm
+++ b/lib/warnings.pm
@@ -5,7 +5,7 @@
 
 package warnings;
 
-our $VERSION = "1.52";
+our $VERSION = "1.53";
 
 # Verify that we're called correctly so that warnings will work.
 # Can't use Carp, since Carp uses us!
@@ -580,6 +580,9 @@ warnings - Perl pragma to control optional warnings
     use warnings;
     no warnings;
 
+    # Standard warnings are enabled by use v5.35 or above
+    use v5.35;
+
     use warnings "all";
     no warnings "uninitialized";
 
@@ -644,6 +647,9 @@ The code in the enclosing block has warnings enabled, but the inner
 block has them disabled.  In this case that means the assignment to the
 scalar C<$z> will trip the C<"Scalar value @x[0] better written as $x[0]">
 warning, but the assignment to the scalar C<$y> will not.
+
+All warnings are enabled automatically within the scope of
+a C<L<use v5.35|perlfunc/use VERSION>> (or higher) declaration.
 
 =head2 Default Warnings and Optional Warnings
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -131,6 +131,7 @@ Artiom Morozov                 <artiom@phreaker.net>
 Artur Bergman                  <artur@contiller.se>
 Arvan                          <apritchard@zeus.com>
 Ash Berlin                     <ash@cpan.org>
+Asher Mancinelli               <asher.mancinelli@pnnl.gov>
 Ask Bjørn Hansen               <ask@develooper.com>
 Atsushi Sugawara               <peanutsjamjam@gmail.com>
 Audrey Tang                    <cpan@audreyt.org>
--- a/op.c
+++ b/op.c
@@ -8864,6 +8864,9 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
                 PL_hints |= HINT_STRICT_SUBS;
             if (!(PL_hints & HINT_EXPLICIT_STRICT_VARS))
                 PL_hints |= HINT_STRICT_VARS;
+
+            if (vcmp(use_version, sv_2mortal(upg_version(newSVpvs("5.035000"), FALSE))) >= 0)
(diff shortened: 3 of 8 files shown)


leonerd:
As a draft it's clearly missing tests/docs/etc, but the actual implementation is basically the code I was going to write, yes. You just beat me to it ;)

oodler577:
I haven't looked at the diff, but @atoomic has a whole branch that explored this with the additional effort lead by @jkeenan - maybe have already seen it. Just wanted to make sure that a connection is made here with effort already expended.

Leont:
> I haven't looked at the diff, but @atoomic has a whole branch that explored this with the additional effort lead by @jkeenan - maybe have already seen it. Just wanted to make sure that a connection is made here with effort already expended.

They were trying a rather different thing, and this PR is only like two lines

leonerd:
Now 5.35.1 is ready for contributions again, we should get this ready to go in.

It needs s/33/35/ applying to it, and some tests + docs adding, but otherwise I think we're ready

scottchiefbaker:
This gets a huge +1 from me. I'd love to see it land for the next version of Perl.

leonerd:
(with PSC hat on) Hi - this looks to be going in the right direction, and to my eye it just needs some documentation adding, and setting whatever is the appropriate version number at the time we merge it (i.e. not just 5.35.0). Is there a plan of someone who can finish this off?

epa:
The `-E` command line flag can be understood as `use 5.xyz` where 5.xyz is whatever is the current version of perl.  Does this patch also enable warnings with `-E`?  (Personally I would like it to.)

Leont:
> The -E command line flag can be understood as use 5.xyz where 5.xyz is whatever is the current version of perl. Does this patch also enable warnings with -E? (Personally I would like it to.)

`-E` has always done the equivalent of `use feature ':5.xyz'`, already doesn't enable strict. I don't think it's logical for it to enable warning.

epa:
So this patch makes `use 5.035` enable warnings, but `use feature ':5.035'` will not enable warnings?

cornett:
Took a stab at some documentation:

https://github.com/Perl/perl5/pull/18976

cornett:
If you are interested, you can merge https://github.com/Perl/perl5/pull/18976 onto this branch to add some docs.

Leont:
> If you are interested, you can merge #18976 onto this branch to add some docs.

Sorry I didn't get back to you, will look at it this weekend

Grinnz:
Just noting for posterity a potential inconsistency with how use VERSION's strict enabling works. Not suggesting we should necessarily make it work the same way (because it's a bit weird, and certainly would be complicated for warnings), but that inconsistencies make it harder to explain.

```
use v5.12; # strict is now on
use v5.10; # strict is now off
```

```
use strict;
use v5.10; # strict remains on - this PR is consistent with this, and we need to retain this behavior
```

```
no strict;
use v5.12; # strict remains off (but rjbs proposes changing this)
```

```
no strict 'refs';
use v5.12; # strict refs remain off, strict vars and subs are now on
```

lib/warnings.pm
```diff
@@ -580,6 +580,9 @@ warnings - Perl pragma to control optional warnings
     use warnings;
     no warnings;
 
+    # Standard warnings are enabled by use v5.35 or above
+    use v5.35;
```
Leont, line 584: Possibly we should ignore 5.35 and make it 5.36 for documentation purposes

epa:
The BUGS section in perl.pod still says "The behavior implied by the `use warnings` pragma is not mandatory."  That should change or be removed?

scottchiefbaker:
I would think so... We should definitely recommend that `use warnings` 
is turned on in the documentation.

Maybe: "The behavior implied by the |use warnings| pragma is not 
mandatory, but is highly recommended in helping catch potential errors."

Grinnz:
Yes, that text is quite outdated, but this is rather off topic.

epa:
I mean that this patch set finally fixes that 'bug'.  The entry can be removed from the perl manual page.

rjbs:
@Leont What stands between us and merging this?

lib/warnings.pm
leonerd: I'd suggest leave it at 5.35 for now, and try to remember to change this to 5.36 on the runup to release

leonerd:
I think this looks fine. All the smokers are happy, aside from the cygwin one which is always throwing a wobbly lately so I've taken to just ignoring it.

I say LGTM

PR #18671, opened 2021-03-30
Silence one build-time warning

jkeenan:
Observed in clang 9 and 10.

Partial solution for https://github.com/Perl/perl5/issues/17015

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -12514,7 +12514,7 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
                  && q == fmtstart + 1 /* plain %d, not %....d */
                  && patend >= fmtstart + sizeof(UTF8f) - 1 /* long enough */
                  && *q == '%'
-                 && strnEQ(q + 1, UTF8f + 2, sizeof(UTF8f) - 3))
+                 && strnEQ(q + 1, (UTF8f) + 2, sizeof(UTF8f) - 3))
             {
 		/* The argument has already gone through cBOOL, so the cast
 		   is safe. */


jkeenan:
Given the discussion in https://github.com/Perl/perl5/issues/17015, I'm surprised that a solution to one of the build-time warnings discussed in that ticket was as simple as putting parentheses around one term.

However, [smoke-test results](http://perl.develop-help.com/?b=smoke-me%2Fjkeenan%2Fgh-17015-sv-c-warning) suggest that this did no harm.

hvds:
Oh! I had completely misunderstood what was being warned about here, how embarrassing - and my apologies for repeatedly denigrating clang's warnings based on that misunderstanding.

I now realise that what it is really complaining about is this bit of C, and I have no idea how it would be interpreted:
```
#define UVuf "lu"  /* typically */
#define UTF8f "d%" UVuf "%4p"
(UTF8f + 2)
```

I've never understood C string concatenation; but if the last line is not interpreted the same as `((UTF8f) + 2)`, then perhaps a more correct and complete fix would instead be to move the parens into the macro definition:
```
#define UTF8f ("d%" UVuf "%4p")
```
However I don't know if that risks causing problems for the more common case where we're concatenating it further.

I'll defer to those who know the C (or preprocessor?) spec better.

tonycoz:
> I've never understood C string concatenation; but if the last line is not interpreted the same as `((UTF8f) + 2)`, then perhaps a more correct and complete fix would instead be to move the parens into the macro definition:
> 
> ```
> #define UTF8f ("d%" UVuf "%4p")
> ```
> 
> However I don't know if that risks causing problems for the more common case where we're concatenating it further.
> 
> I'll defer to those who know the C (or preprocessor?) spec better.

I'm pretty sure that would break concatenating it into larger string literals.

The relevant part of the standard (6.4.5 paragraph 5)

> In translation phase 6, the multibyte character sequences specified by any sequence of adjacent character and identically-prefixed string literal tokens are concatenated into a single multibyte character sequence. If any of the tokens has an encoding prefix, the resulting multibyte character sequence is treated as having the same prefix; otherwise, it is treated as a character string literal. Whether differently-prefixed wide string literal tokens can be concatenated and, if so, the treatment of the resulting multibyte character sequence are implementation-defined.

and the `(` between the string literal tokens would make them non-adjacent.

tonycoz (approved):

PR #18674, opened 2021-04-01
Install as /usr/bin/perl5

Corion:
Traditionally, `Configure` asks whether to install the main binary as `/usr/bin/perl`. During `make install`, the binary is then copied to `/usr/bin/perl` after building.

This set of patches changes `Configure` and `installperl` to install the binary as `/usr/bin/perl5` and as `perl5`, in addition. An appropriate new config item is added, `installusrbinperl_major`, should one want to disable this behaviour.


Leont:
Arguably this should default to false if `versiononly` is true.

Corion:
The current behaviour is consistent with how `installusrbin` still defaults to `define` even when `versiononly` is defined. This is weird, but wer are then consistent in the weirdness.

I would consider it more correct to change both, `installusrbinperl` and `installusrbin_major` to default to false when `versiononly` is true.

Leont:
I think this patch takes the wrong path. It should install at `"$installbin/perl$]"`, not as `"/usr/bin/perl$]"`

Corion:
The patch does install a `perl5` into `installbin` by default. It only installs additionally as `/usr/bin/perl5` if `installusrbin_major` is `define`.

Leont:
> The patch does install a `perl5` into `installbin` by default. It only installs additionally as `/usr/bin/perl5` if `installusrbin_major` is `define`.

Then I had misread things, nm :-)

oodler577:
Unless I grossly misunderstand the purpose of this change, I am vehemently opposed to this sort of chicanery. If perl7 comes to pass, it's not going to do the package maintainers and individuals who maintain their won systems any good to change underneath what they expect, i.e., a `perl` - let it be handled where it should be handled. It's worse that not-helpful. This is harmful and infact creates a *third* thing to disambugiate.

Corion:
@oodler577 What is there to disambiguate between `perl5` and `perl7` ? This patch adds a fallback so you can specify in your Perl programs what interpreter you need, independently of when your distribution decides to upgrade its Perl installation from 5 to 7. It even opens up a way for a distribution to provide both versions.

oodler577:
AFAIK, there has been no decision by the PSC moving ahead on what to do here. That's the main issue; it's premature. Fine if this was just getting ahead of the curve or as a POC. I get it. But the spirit of this change seems to assume that there will even be `perl5` or `perl7` binaries; and I don't believe this is even close to being resolved. As an administrator myself and one who has contributed to various package systems over the years (not `perl` or anything majorly related here), I can say though that this change actually complicates the issue. Traditionally this is handled via symlinks, such as how FreeBSD does it. Registering my comments simply for the eventuality that someone will use as an argument, _oh hey look the works already done_. This is not a valid argument (again for the record). TY

Leont:
> AFAIK, there has been no decision by the PSC moving ahead on what to do here. That's the main issue; it's premature.

Not having a decision does not mean having the branch exist is premature; it's only merging it that would be.

oodler577:
> Not having a decision does not mean having the branch exist is premature; it's only merging it that would be.

Correct; you own your own time. However, an invalid argument for merging would be, _the work is already done_. It's a slippery descent into madness if we begin (or continue) to admit this fallaciousness. Very, _very_, slippery. Your time and attention to this matter is appreciated.

duncand:
I strongly oppose having Perl binaries named after specific versions.  There should just be a single "perl" with NO version number, and that binary knows the appropriate action to do implicitly when given Perl code of any age.  If repackagers want to add aliases that's up to them, but the Perl core should only ever install "perl" and that's it.  If it installs any "perl5" that should only be for backwards compatibility if if did so before, and it should NOT be added as a new feature.  And the default setting should remain "perl".  The PSC already formally declared that "use v7;" inside code etc and encouraging use lines is the way we are moving forward, which I wholeheartedly agree with.

Corion:
If PSC decides not to merge Nick's dual-nature Perl interpreter changes (or their moral equivalent), Perl will need to offer an easy approach for distributions and their users to manage the transition from Perl 5 to Perl 7. While this patch predates the announcement of the PSC to `use v7;`, so far no code has appeared for that, and there have been statements from Sawyer that Perl would be 7-by-default.

This change is to ensure a sane transition when no dual-nature Perl interpreter gets merged. I'm happy if a better way is implemented, but the only code I've seen to that effect does not seem to have a PR yet. I'm also happy to rework the change to make the option of a `perl5` (and/or `perl7`) binary off by default and documented as a migration tool.

Leont:
>  If repackagers want to add aliases that's up to them, but the Perl core should only ever install "perl" and that's it.

That would suggest this should be merged but the default should be off?

duncand:
> > If repackagers want to add aliases that's up to them, but the Perl core should only ever install "perl" and that's it.
> 
> That would suggest this should be merged but the default should be off?

The very least I would expect is that the default behavior of Configure is to only make a "perl" and that a "perl5" is only created if someone explicitly opts in.

duncand:
> If PSC decides not to merge Nick's dual-nature Perl interpreter changes (or their moral equivalent), Perl will need to offer an easy approach for distributions and their users to manage the transition from Perl 5 to Perl 7. While this patch predates the announcement of the PSC to `use v7;`, so far no code has appeared for that, and there have been statements from Sawyer that Perl would be 7-by-default.
> 
> This change is to ensure a sane transition when no dual-nature Perl interpreter gets merged. I'm happy if a better way is implemented, but the only code I've seen to that effect does not seem to have a PR yet. I'm also happy to rework the change to make the option of a `perl5` (and/or `perl7`) binary off by default and documented as a migration tool.

Part of the beauty of "use v7;" being formally declared as the decided way forward is that this is already what Perl does and has for a long time.  We DON'T NEED code for "use v7;" because that's how Perl ALREADY behaves.  Any existing Perl interpreter will already do the right thing when a "use vN;" appears, it will already know whether it can run it or otherwise say sorry that's a version I don't support.

Adding perl5 or perl7 binaries now is wholly premature and assumes a path that was already decided against.

Grinnz:
> If PSC decides not to merge Nick's dual-nature Perl interpreter changes (or their moral equivalent), Perl will need to offer an easy approach for distributions and their users to manage the transition from Perl 5 to Perl 7. While this patch predates the announcement of the PSC to `use v7;`, so far no code has appeared for that, and there have been statements from Sawyer that Perl would be 7-by-default.

My understanding is these statements are superceded by [the latest decision](https://www.nntp.perl.org/group/perl.perl5.porters/2021/04/msg259789.html), which was made while Sawyer was still on the steering council and has not been contradicted since. It's also unclear what "code for that" would be, as all required functionality already exists, aside from the version number itself and features we want to add to it.

duncand:
I suggest that this PR would simply be closed/rejected without changes or merging as as far as I can tell it doesn't do anything constructive.  We don't want to encourage greater use of "perl5" binaries than already exist.  And those that already use "perl5" binaries already have the mechanisms in place to do so and can be left to continue working as they do for backwards compatibility.  Not merging means full compatibility with how things already are is maintained, and existing processes continue to work without further maintenance, and we're not encouraging greater "perl5" use.

oodler577:
> I strongly oppose having Perl binaries named after specific versions. There should just be a single "perl" with NO version number, and that binary knows the appropriate action to do implicitly when given Perl code of any age. If repackagers want to add aliases that's up to them, but the Perl core should only ever install "perl" and that's it. If it installs any "perl5" that should only be for backwards compatibility if if did so before, and it should NOT be added as a new feature. And the default setting should remain "perl". The PSC already formally declared that "use v7;" inside code etc and encouraging use lines is the way we are moving forward, which I wholeheartedly agree with.

FWIW, I concur with @duncand's strenuous objection.

oodler577:
> Perl will need to offer an easy approach

this is simply, false

oodler577:
> @oodler577 What is there to disambiguate between `perl5` and `perl7` ? This patch adds a fallback so you can specify in your Perl programs what interpreter you need, independently of when your distribution decides to upgrade its Perl installation from 5 to 7. It even opens up a way for a distribution to provide both versions.

You're adding `perl5` to `perl` and `perl7`. I get math is _hard_, but let's not make it harder.

duncand:
> > What is there to disambiguate between `perl5` and `perl7` ? This patch adds a fallback so you can specify in your Perl programs what interpreter you need, independently of when your distribution decides to upgrade its Perl installation from 5 to 7. It even opens up a way for a distribution to provide both versions.
> 
> You're adding `perl5` to `perl` and `perl7`. I get math is _hard_, but let's not make it harder.

There is nothing special about Perl version 7 aside from marketing.  The answer to the question is exactly the same as how do we distinguish between Perl 5.30 and 5.32.  Either it doesn't matter because the single "perl" will handle all versions, or we disambiguate by the file containing "use 5.30" or "use 5.32" or "use 7" etc.

neilb:
Given the PSC decision that we want Perl to always be a candidate for /usr/bin/perl, and that Perl 7 won't install a perl7 binary, we don't want a perl5 binary either.

This PR was discussed in the [PSC meeting 2021-07-02](https://www.nntp.perl.org/group/perl.perl5.porters/2021/07/msg260858.html), when we decided it can now be closed.

PR #18676, opened 2021-04-01
Simplify the split() documentation by removing the join()s from the examples

scottchiefbaker:
split() is a very basic function, and the documentation should be simple for novices to be able to understand. The split() documentation contains a lot of join() in the examples which only serve to muddle the concepts. This replaces the join() in the example with actual output behind a comment block

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -7871,7 +7871,6 @@ Using split in this fashion is very similar to how
 L<C<qwE<sol>E<sol>>|/qwE<sol>STRINGE<sol>> works.
 
 However, this special treatment can be avoided by specifying
-a separator.  However, this special treatment can be avoided by specifying
 the pattern S<C</ />> instead of the string S<C<" ">>, thereby allowing
 only a single space character to be a separator.  In earlier Perls this
 special case was restricted to the use of a plain S<C<" ">> as the
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -7826,16 +7826,15 @@ to specify a pattern that varies at runtime.
 If PATTERN matches the empty string, the EXPR is split at the match
 position (between characters).  As an example, the following:
 
-    print join(':', split(/b/, 'abc')), "\n";
+    my @x = split(/b/, "abc"); # ("a", "c")
 
-uses the C<b> in C<'abc'> as a separator to produce the output C<a:c>.
+uses the C<b> in C<'abc'> as a separator to produce the list ("a", "c").
 However, this:
 
-    print join(':', split(//, 'abc')), "\n";
+    my @x = split(//, "abc"); # ("a", "b", "c")
 
-uses empty string matches as separators to produce the output
-C<a:b:c>; thus, the empty string may be used to split EXPR into a
-list of its component characters.
+uses empty string matches as separators; thus, the empty string
+may be used to split EXPR into a list of its component characters.
 
 As a special case for L<C<split>|/split E<sol>PATTERNE<sol>,EXPR,LIMIT>,
 the empty pattern given in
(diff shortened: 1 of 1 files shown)


Grinnz:
This looks much clearer and more consistent to me. One thing I would suggest is maybe leaving out all of the semicolons, since these are just showing the "output" of expressions and not what they'll do as a complete statement (which is nothing, in void context).

pod/perlfunc.pod
```diff
@@ -7826,16 +7826,15 @@ to specify a pattern that varies at runtime.
 If PATTERN matches the empty string, the EXPR is split at the match
 position (between characters).  As an example, the following:
 
-    print join(':', split(/b/, 'abc')), "\n";
+    split(/b/, "abc"); # ("a", "c");
 
-uses the C<b> in C<'abc'> as a separator to produce the output C<a:c>.
+uses the C<b> in C<'abc'> as a separator.
```
Grinnz: Suggested by Bram:
```suggestion
uses the C<b> in C<'abc'> as a separator to produce the list ("a", "c").
```

pod/perlfunc.pod
```diff
@@ -7928,23 +7928,21 @@ produces zero fields, regardless of the LIMIT specified.
...
 
-produces the output C<:abc>.  However, a zero-width match at the
+splits into two elements.  However, a zero-width match at the
 beginning of EXPR never produces an empty field, so that:
 
-    print join(':', split(//, ' abc'));
+    split(//, " abc"); # (" ", "a", "b", "c")
 
-produces the output S<C< :a:b:c>> (rather than S<C<: :a:b:c>>).
+splits in to four elements instead of five.
```
Grinnz:
```suggestion
splits into four elements instead of five.
```

pod/perlfunc.pod
```diff
@@ -7906,13 +7906,13 @@ trailing empty fields are stripped (empty leading fields are always
 preserved); if all fields are empty, then all fields are considered to
 be trailing (and are thus stripped in this case).  Thus, the following:
 
-    print join(':', split(/,/, 'a,b,c,,,')), "\n";
+    split(/,/, "a,b,c,,,"); # ("a", "b", "c")
 
-produces the output C<a:b:c>, but the following:
+produces only a three element array.
```
Grinnz:
```suggestion
produces only a three element list.
```

pod/perlfunc.pod
```diff
@@ -7906,13 +7906,13 @@ trailing empty fields are stripped (empty leading fields are always
...
 
-    print join(':', split(/,/, 'a,b,c,,,', -1)), "\n";
+    split(/,/, "a,b,c,,,", -1); # ("a", "b", "c", "", "", "")
 
-produces the output C<a:b:c:::>.
+produces a six element array.
```
Grinnz:
```suggestion
produces a six element list.
```

pod/perlfunc.pod
```diff
@@ -7860,7 +7859,15 @@ S<C<"\x20">>, but not e.g. S<C</ />>).  In this case, any leading
 whitespace in EXPR is removed before splitting occurs, and the PATTERN is
 instead treated as if it were C</\s+/>; in particular, this means that
 I<any> contiguous whitespace (not just a single space character) is used as
-a separator.  However, this special treatment can be avoided by specifying
+a separator.
+
+    my @x = split(" ", "  Quick brown fox\n"); # ("Quick", "brown", "fox")
```
Grinnz: Podcheck found these verbatim lines to be too long

pod/perlfunc.pod
```diff
@@ -7959,20 +7957,11 @@ does B<not> count towards the LIMIT.  Consider the following expressions
...
-    split(/-|(,)/, "1-10,20", 3)
-    # ('1', undef, '10', ',', '20')
-
-    split(/(-)|,/, "1-10,20", 3)
-    # ('1', '-', '10', undef, '20')
-
-    split(/(-)|(,)/, "1-10,20", 3)
-    # ('1', '-', undef, '10', undef, ',', '20')
+    my @x = split(/-|,/    , "1-10,20", 3); # ("1", "10", "20")
+    my @x = split(/(-|,)/  , "1-10,20", 3); # ("1", "-", "10", ",", "20")
```
Grinnz: Podcheck found these verbatim lines to be too long (after the first one)

Grinnz (approved):

pod/perlfunc.pod
```diff
@@ -7860,6 +7859,18 @@ S<C<"\x20">>, but not e.g. S<C</ />>).  In this case, any leading
...
+    my @x = split(" ", "  Quick brown fox\n");
+    # ("Quick", "brown", "fox")
+
+    my @x = split(" ", "RED\tGREEN\tBLUE");
+    # ("RED", "GREEN", "BLUE")
+
+Using split in this fashion is very similar to how
+L<C<qwE<sol>E<sol>>|/qwE<sol>STRINGE<sol>> works.
+
+However, this special treatment can be avoided by specifying
```
khwilliamson, line 7873: This doesn't sound right.  Two adjacent sentences each beginning with 

> However, this special treatment

scottchiefbaker: Good catch I'll tweak the wording.

PR #18688, opened 2021-04-04
feature.pl: stop using "switch" as a demo feature!

rjbs:
It is experimental, it would cause warnings, and frankly, I would like to just inch toward its disappearance.


khwilliamson (approved):

leonerd:
LGTM

Issue #18667, opened 2021-03-28
BBC: Blead Breaks Config::Model


eserte:
DDUMONT/Config-Model-2.141.tar.gz fails on some systems since 5.33.5 or so.
Discussion happened so far in the CPAN module: https://github.com/dod38fr/config-model/issues/26
but it's likely to be a bleadperl issue.

hvds:
Not sure whether to comment here or at config-model, but I've started trying to look into this.

(I tried running the test script with `-Dm`, but the address for which "attempt to free unreferenced scalar" was reported did not appear in any of the alloc or free diagnostics, so I may be misunderstanding what `-Dm` does.)

Investigating one of the SVs for which it reports "attempt to free unreferenced scalar", it appears that this is part of a hash that allows lookup by the numification of a Config::Model object:
```
(gdb) print Perl_sv_dump(PL_tmps_stack[14])
SV = IV(0x5555581efef0) at 0x5555581eff00
  REFCNT = 1
  FLAGS = (ROK)
  RV = 0x5555581e9148
    SV = PVHV(0x55555884e2e0) at 0x5555581e9148
      REFCNT = 1
      FLAGS = (OOK,SHAREKEYS)
      AUX_FLAGS = 0
      ARRAY = 0x555558281180  (0:6, 1:2)
      hash quality = 125.0%
      KEYS = 2
      FILL = 2
      MAX = 7
      RITER = -1
      EITER = 0x0
      RAND = 0xd738dcb0
        Elt "93825001673328" HASH = 0x55b53aca
        SV = IV(0x555558265758) at 0x555558265768
          REFCNT = 1
          FLAGS = (IOK,pIOK)
          IV = 1
        Elt "93825001672968" HASH = 0xbf143ddf
        SV = IV(0x555558265aa0) at 0x555558265ab0
          REFCNT = 1
          FLAGS = (IOK,pIOK)
          IV = 1
$13 = void
(gdb) p /x 93825001673328
$14 = 0x555555e55270
(gdb) print Perl_sv_dump((SV*)0x555555e55270)
SV = PVHV(0x555555e41f40) at 0x555555e55270
  REFCNT = 1
  FLAGS = (OBJECT,OOK,SHAREKEYS)
  STASH = 0x55555675dd00        "Config::Model"
  AUX_FLAGS = 0
  ARRAY = 0x5555581ee4d0  (0:9, 1:5, 2:2)
  hash quality = 103.8%
  KEYS = 9
  FILL = 7
[...]

PR #18690, opened 2021-04-05
Don't zero non-real arrays

hvds:
A minor difference in behaviour introduced by the rewrite of av_extend_guts
in 440c1856 and 399fef93 is that when the SV* array is first acquired, we
now zero its contents regardless of whether that is needed - and add an
extra label and 'goto' to do so.

This commit restores the original behaviour: for non-refcounted arrays such
as @_ and pad arrays this is not needed, and for pads in particular we may
hit this path many times during the course of a program.

@richardleach I mentioned this difference in #18667, but the focus there was on the ary_offset bug. If you think this requires further discussion I can also open a separate ticket for it.

### DIFF
--- a/av.c
+++ b/av.c
@@ -193,11 +193,9 @@ Perl_av_extend_guts(pTHX_ AV *av, SSize_t key, SSize_t *maxp, SV ***allocp,
              * don't get any special treatment here. */
             ary_offset = 0;
             to_null = *maxp+1;
-            goto zero;
         }
 
         if (av && AvREAL(av)) {
-          zero:
             Zero(*allocp + ary_offset,to_null,SV*);
         }
 


richardleach:
> This commit restores the original behavious: for non-refcounted arrays such
> as @_ and pad arrays this is not needed, and for pads in particular we may
> hit this path many times during the course of a program.

The original change was intentional with the thought being that not having to go through the `if (av && AvREAL(av)) { `branch when creating a new array might help performance on balance. Specifically, my thinking was that creating stacks is comparatively rare, people worried about performance try to avoid unnecessarily creating @_, and modern OO code is likely to create lots of short-lived arrays (which would benefit).

I didn't consider many other kind of non-real arrays at the time, because I don't have a good feel for how common they are and what their creation/growth patterns look like. So me having missed something important of that nature is certainly possible.

However, I **did** look at _pad.c_ and noted that there's:
* one call to _av_extend()_: https://github.com/Perl/perl5/blob/blead/pad.c#L2491 - but it looks to pass a real AV which will therefore get Zeroed.
* two calls directly to _av_extend_guts()_ with `NULL` as the first parameter - https://github.com/Perl/perl5/blob/blead/pad.c#L2571 & https://github.com/Perl/perl5/blob/blead/pad.c#L2624 - so `if (av)` is not going to be true inside _av_extend_guts_ and the arrays are going to get Zeroed.

So I'm thinking that if pads don't need to be zeroed, the current behaviour in pad.c would have to be changed.

richardleach:
PS I forgot in the reply above: in _pad.c_, there are also a number of direct `Newx` and `Newxz` calls, including e.g. this one which deliberately avoids calling _av_extend()_: https://github.com/Perl/perl5/blob/blead/pad.c#L242.

(Other places in core also directly allocate their array and fix up the _xpv_ pointers too. (_av_make()_ does, there's something else that doesn't want the default sizing in another file IIRC. Something that might be nice would be a macro for doing Newx/z & pointer fixup, rather than all these places doing it themselves.)

richardleach:
Are `Perl_pad_push` and `Perl_padlist_store` typically the main functions of concern for  run-time performance?

hvds:
>  two calls directly to _av_extend_guts()_ with `NULL` as the first parameter - https://github.com/Perl/perl5/blob/blead/pad.c#L2571 & https://github.com/Perl/perl5/blob/blead/pad.c#L2624 - so `if (av)` is not going to be true inside _av_extend_guts_ and the arrays are going to get Zeroed

Was that a typo? When `if (av)` is false, traditionally the arrays would _not_ be zeroed.

Anyway, I tried instrumenting with:
```
@@ -193,6 +193,13 @@ Perl_av_extend_guts(pTHX_ AV *av, SSize_t key, SSize_t *maxp, SV ***allocp,
              * don't get any special treatment here. */
             ary_offset = 0;
             to_null = *maxp+1;
+            if (av && AvREAL(av)) {
+                ++real_count;
+                real_size += to_null;
+            } else {
+                ++fake_count;
+                fake_size += to_null;
+            }
         }
 
         if (av && AvREAL(av)) {
```
.. and for lack of a better example of realistic code, ran mktables up to the point of perl_destruct:
```
% gdb --args ./miniperl -Ilib lib/unicore/mktables -C lib/unicore -P pod -maketest -makelist -p -w
(gdb) break miniperlmain.c:127
(gdb) run
Breakpoint 1, main (...) at miniperlmain.c:127
127         exitstatus = perl_destruct(my_perl);
(gdb) p real_count
$1 = 23501
(gdb) p fake_count
$2 = 57
(gdb) p real_size
$3 = 171209
(gdb) p fake_size
$4 = 436
```

So colour me convinced - I'll withdraw this PR. Even though the test on i386 is just 4 instructions (including just a 1-byte memory read), it seems clear that performing that test 23558 times to avoid zeroing 436 * sizeof(SV*) bytes is a net loss. If someone has a more realistic workload on which to try the same instrumentation, I'd be interested to see it, but I think I'd expect it to tell essentially the same story.

For future reference: it's _really_ useful if behaviour changes like this are called out by a separate commit, even if that results in a commit that seems trivially small.

Issue #18695, opened 2021-04-06
newAV_x( number) macros to create AV and array?


richardleach:
In various parts of core, SV* arrays are created directly - either for performance or to try to ensure very small array sizes - rather than via _av_extend()_. In other words, multiple variations on the following exist:

```
AV* const some_av = newAV();
Newx( AvALLOC(some_av), ix + 1, SV *);
AvARRAY(some_av) = AvALLOC(some_av);
AvMAX(newpad) = ix;
AvFILLp(newpad) = -1;
```
(There are probably places that do not do this, but could do so in order to save some branches/instructions.)

**Suggestion**
For consistency and future maintainability, it would seem nice to wrap that functionality up in some common macros.

Possible names - intended to imply the amalgamation of _newAV_ and the _Newx_ family of macros:
* newAV_x( ix ) - array elements are not zeroed
* newAV_xz ( ix ) - array elements are zeroed
* plus the cast variant if required, but I'm not sure that it will be

**Note**
Users of such macros might actually want to set AvFILLp to something other than -1, but if they do so soon after the macro, a decent compiler should recognize the assignment in the macro as unnecessary.

**Volunteer**
If people like this idea, I'm happy to implement it, once naming and where to put these macros (_av.h_?) has been agreed upon.

hvds:
> In various parts of core, SV* arrays are created directly - either for performance or to try to ensure very small array sizes - rather than via _av_extend()_. In other words, multiple variations on the following exist:
> 
> ```
> AV* const some_av = newAV();
> Newx( AvALLOC(some_av), ix + 1, SV *);
> AvARRAY(some_av) = AvALLOC(some_av);
> AvMAX(newpad) = ix;
> AvFILLp(newpad) = -1;
> ```

I couldn't find any examples of this pattern after grepping for AvALLOC, could you point at a couple?

If a guarantee on the array size is an important feature of the new macro(s) that should probably also be reflected in the name, perhaps by including an adjective such as "fixed".

richardleach:
> I couldn't find any examples of this pattern after grepping for AvALLOC, could you point at a couple?

I thought there were more than this, but all I could quickly turn up were:

- https://github.com/Perl/perl5/blob/blead/pad.c#L242 and https://github.com/Perl/perl5/blob/blead/pad.c#L2596 look to be variations.
- av_make does it: https://github.com/Perl/perl5/blob/blead/av.c#L421
[...]

PR #18696, opened 2021-04-06
pad.c - less avoidable, inefficient use of av_store

richardleach:
In a few places in _pad.c_, particularly in _Perl_pad_push()_,  `av_store()` is used
to populate array elements very soon after creating the AV via `newAV()`.
Given that proximity, much of the code of `av_store()` can be reasoned to be 
superfluous and therefore its use is inefficient:
* `if (SvRMAGICAL(av))` - never true
* `if (key < 0)`        - never true
* `if (SvREADONLY(av)`  - never true
* `if (!AvREAL(av) && AvREIFY(av))` - never true
* `if (key > AvMAX(av))` - could be true some of the time, but it's easy to `Newx`
 or `Newxz `the SV* array shortly after the `newAV()`, so this test is then not true
* `if (AvFILLp(av) < key)` - final value can be set once, rather than incrementally
* `if (!AvREAL(av))` - never true
* `SvREFCNT_dec(ary[key])` - not necessary when the array is brand new
* `if (SvSMAGICAL(av))` - never true
* `return &ary[key];`   - return value is not used

This PR does two things to avoid such use of `av_store()`:
* Directly creates the SV* arrays for the relevant AVs
* Directly assigns the array elements

Direct array creation does make the code harder to follow, but this could be 
tidied up via macros e.g. https://github.com/Perl/perl5/issues/18695

Note: Some use of `av_store()` elsewhere in _pad.c_ might also be inefficient,
but evaluating occurrences when the AV and SV* array already exist is 
less straightforward.

### DIFF
--- a/pad.c
+++ b/pad.c
@@ -220,17 +220,24 @@ Perl_pad_new(pTHX_ int flags)
 
     Newxz(padlist, 1, PADLIST);
     pad		= newAV();
+    Newxz(AvALLOC(pad), 4, SV *); /* Originally sized to
+                                     match av_extend default */
+    AvARRAY(pad) = AvALLOC(pad);
+    AvMAX(pad) = 3;
+    AvFILLp(pad) = 0; /* @_ or NULL, set below. */
 
     if (flags & padnew_CLONE) {
         AV * const a0 = newAV();			/* will be @_ */
-        av_store(pad, 0, MUTABLE_SV(a0));
+        AvARRAY(pad)[0] = MUTABLE_SV(a0);
         AvREIFY_only(a0);
 
         PadnamelistREFCNT(padname = PL_comppad_name)++;
     }
     else {
         padlist->xpadl_id = PL_padlist_generation++;
-        av_store(pad, 0, NULL);
+        /* Set implicitly through use of Newxz above
+            AvARRAY(pad)[0] = NULL;
+        */
         padname = newPADNAMELIST(0);
         padnamelist_store(padname, 0, &PL_padname_undef);
     }
--- a/pad.c
+++ b/pad.c
@@ -2396,6 +2396,10 @@ Perl_pad_push(pTHX_ PADLIST *padlist, int depth)
         PADNAME ** const names = PadnamelistARRAY((PADNAMELIST *)svp[0]);
         AV *av;
 
+        Newxz( AvALLOC(newpad), ix + 1, SV *);
+        AvARRAY(newpad) = AvALLOC(newpad);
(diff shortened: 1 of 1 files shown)


hvds:
I like it. I think it could be improved by first changing Perl_pad_push to assign the SV to store into `sv` in all cases, falling through to a single `av_store(newpad, ix, sv)` just before the end of the loop; this commit would then have just the one `av_store` to change, and I think the whole would end up a little more legible.

richardleach:
Thanks for the feedback. I'll do that at the next opportunity.

The same sort of analysis might be fruitful throughout _pad.c_.  I had a quick look at _Perl_pad_alloc_, and the first call to _av_fetch()_:
`sv = *av_fetch(PL_comppad, AvFILLp(PL_comppad) + 1, TRUE);`
almost certainly reliably reduces to just a call to _av_store()_ like:
`sv = *av_store(PL_comppad,AvFILLp(PL_comppad) + 1,newSV(0));`
at which point much (but not all) of the discussion above will also apply.

richardleach:
> I like it. I think it could be improved by first changing Perl_pad_push to assign the SV to store into `sv` in all cases, falling through to a single `av_store(newpad, ix, sv)` just before the end of the loop; this commit would then have just the one `av_store` to change, and I think the whole would end up a little more legible.

Ok, if I didn't bodge the rebase, that has been added.

hvds:
Looks good to me, if I can get a +1 from someone else I'll be happy to merge it - @tonycoz, @khwilliamson?

hvds:
> Looks good to me, if I can get a +1 from someone else I'll be happy to merge it - @tonycoz, @khwilliamson?

@tonycoz, khw says he is not comfortable reviewing this, any chance you can take a look?

I understand we have a few more days before things get tight for 5.34.

richardleach:
> @tonycoz, khw says he is not comfortable reviewing this, any chance you can take a look?
> 
> I understand we have a few more days before things get tight for 5.34.

I don't mind this waiting for early in the next dev cycle.

hvds:
Now merged to blead as a043512d92, e92b9ddc13 and 36f923e59e. I'll give it a day or two to smoke before proceeding to #18729.

hvds:
Smoke gave one bad result on a threaded build (http://perl.develop-help.com/raw/?id=268101); this appears to be quite a temperamental smoker (filter arch="aarch64") and the results _look_ like nonsense to me, with apparently random exit codes from a bunch of tests in ExtUtils-Constant and ExtUtils-MakeMaker, but I'm recording it here in case anyone else sees it as a reason to raise a red flag.

Issue #18543, opened 2021-02-07
[feature] strict and warnings optional extensions


Grinnz:
Since they were introduced in early Perl 5, `strict` and `warnings` have enabled the equivalent of `all` when used in the recommended `use strict;` and `use warnings;` forms. There have occasionally been ideas of additional behavior that would make sense for these pragmas to enable, but without enabling them in the massive amount of existing code which currently enable `all` implicitly or explicitly. So I propose that mechanisms be added for "optional" strictures and warnings, which are not enabled by `use strict; use warnings;`.

[strict](https://perldoc.perl.org/strict) has three subpragmas: `vars`, `refs`, and `subs`. There is no `all` category, but a bare `use strict;` enables these three items. So to add an optional stricture, we would need a mechanism to add a subpragma which is *not* enabled by a bare `use strict;`, and users could then simply `use strict 'whatever';` to opt-in to the stricture for that scope. A flag to `use strict` which would enable these additional strictures could be considered, but I do not think it would be a good idea since such a flag would effectively be restricted to whatever it enables the first time it's added.

[warnings](https://perldoc.perl.org/warnings) has hundreds of warning categories organized in a few classifications, as described in [perldiag](https://perldoc.perl.org/perldiag).

* The `(S)` and `(D)` class warnings are enabled by default - they are considered severe enough that all code should be subject to them, unless explicitly disabled.
* The `(W)` class warnings are not enabled by default, but are enabled when under `use warnings 'all';`, which is the same as `use warnings;`.

There is no policy against adding new warnings to these classifications at any time - but there is an expectation for the warnings to be reasonable for the classification they're added to, so as not to cause grief for users who expected certain behavior.

So to add "optional warnings", we would need a new optional warnings classification `(O)` which is *not* enabled by `use warnings 'all';` or `use warnings;`. Yes, this makes the 'all' category name misleading, but I assert that this doesn't really matter, because very few people actually spell it out. An 'optional' category name and a way to enable it in addition to 'all' could be reasonable for users who really want all possible warnings, but this shouldn't necessarily be encouraged, because such 'optional' warnings would be limited to those which are deemed too niche for the userbase at large.

I am purposely describing only the user-facing API design, as the implementation details are outside my expertise, but an important component of making this happen.

leonerd:
An excellent idea. Over the years I have thought up a few things I thought ought to be `strict` flags, and always been stuck for not being able to add them.

[...]

PR #18697, opened 2021-04-07
Feature stringification

tonycoz:
Another simple feature that disables stringifying references.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -31,6 +31,7 @@
 #define PERL_IN_SV_C
 #include "perl.h"
 #include "regcomp.h"
+#include "feature.h"
 #ifdef __VMS
 # include <rms.h>
 #endif
@@ -3105,6 +3106,11 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
 	    if (lp)
 		*lp = len;
 	    SAVEFREEPV(buffer);
+
+            if (!FEATURE_STRINGIFICATION_IS_ENABLED) {
+                Perl_croak(aTHX_ "Attempt to stringify reference under no feature stringification");
+            }
+
 	    return retval;
 	}
     }
--- a/MANIFEST
+++ b/MANIFEST
@@ -5642,6 +5642,7 @@ t/lib/feature/multidimensional	Tests for enabling/disabling $foo{$x, $y} => $foo
 t/lib/feature/nonesuch		Tests for enabling/disabling nonexistent feature
 t/lib/feature/removed		Tests for enabling/disabling removed feature
 t/lib/feature/say		Tests for enabling/disabling say feature
+t/lib/feature/stringification	Tests for enabling/disabling stringification feature
 t/lib/feature/switch		Tests for enabling/disabling switch feature
 t/lib/h2ph.h			Test header file for h2ph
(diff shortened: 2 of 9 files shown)


rjbs:
👍🏽  Would use!

Also, we're gonna have to (soon, I hope) talk about the line between what's strict and what's features and what's what.  This feature has been proposed before as "use strict 'stringification'".  But that's fine we'll sort that out way before vNEXT!

sv.c
```diff
@@ -3105,6 +3106,11 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
 	    if (lp)
 		*lp = len;
 	    SAVEFREEPV(buffer);
+
+            if (!FEATURE_STRINGIFICATION_IS_ENABLED) {
+                Perl_croak(aTHX_ "Attempt to stringify reference under no feature stringification");
+            }
```
ilmari, line 3112: Wouldn't it be better to have this check _before_ doing all the work to stringify the reference?

leonerd:
Excellent start - a much cleaner implementation than my aborted half-attempt at a CPAN module for this.

But I don't think that `feature.pm` is the place to control it. "features" are things like new language keywords or syntax, things we'd like to turn on and make available. Even "indirect" can be seen in this manner - it's a quirk of syntax which in this particular case we hope to be able to turn off one day.

But stringification of references is more like a stricture - as I have hoped before and @rjbs touches on above, would probably be better controlled by `strict.pm`.

scottchiefbaker:
This gets a solid :+1: from me... 

I do agree with @leonerd and @rjbs that this is probably better suited to be in `strict.pm`. There has been exactly zero times I've wanted the string representation of a reference. IMHO this should be automatic as part of `use strict`.

Grinnz:
It shouldn't be made part of default `strict` as we have no way to version that from applying to existing code. But it would be easily possible (interface-wise) to add non-default strictures, as outlined in #18543 (and then `use vX` could at some point add in this stricture, if we so desire).

sv.c
tonycoz: I originally planned to add more information to the message, probably using the information used to produce the typical `Class=TYPE(0xaddress)`, but later decided to keep it simple, depending on any feedback.

tonycoz:
It's mostly done as a feature for demonstration purposes, though both the indirect and bareword_filehandles could also be considered "strict"-ish type things.

Right now we have one free hints bit, and given the high probability that we'll need more feature bundles, I think the currently available hints bit needs to be reserved to expand the feature bundle number.

The simplest solution would be to add another hints word, but that also needs to be stored in COPs, which are amongst the largest OPs in memory, and probably the most common OP in compiled code, so I'm hesitant to expand them any further just yet.

Another option would be to implement them as a feature, so:
```
use strict "stringification";
```

might under the hood be something like:

```
package strict;
...
sub import {
...
  if ($_ eq "stringification") {
    require feature;
    feature->unimport("stringification");
  }
...
}
```

I'm looking at a change to reduce the number of COPs (only cop_line changes from COP to COP in many cases), but I don't know how well that's going to work out.

jkeenan:
@tonycoz, people seemed to like this proposal but have doubts on its implementation.  Given that there are merge conflicts, could you review this p.r. and let us know if you want to move forward?

Thank you very much.
Jim Keenan

PR #18702, opened 2021-04-09
win32.c: make reading UTF-8 characters from the console possible

### LINKED ISSUE #18701: Reading non-ASCII characters from the console on Windows doesn't work when codepage is 65001
**Description**

Reading non-ASCII characters from the console on Windows doesn't work when codepage is 65001. It's caused by a bug in Windows: https://github.com/microsoft/terminal/issues/4551

tl;dr: ReadFile() and ReadConsoleA() return zeros instead of non-ASCII characters when the console codepage is set to 65001. It's broken on all Windows versions, including the latest Windows 10 20H2.

This issue was [discovered](https://www.reddit.com/r/perl/comments/mmd0j5/how_can_i_get_a_simple_filter_program_to_support/) by a reddit user.

I'll submit a PR with a proposed workaround soon.

**Steps to Reproduce**
```
> chcp 65001
> perl -E "while(<>) { printf qq<%vd\n>, $_ }"
<input ąść and press enter>
0.0.0.10
```

**Expected behavior**
```
> chcp 65001
> perl -E "while(<>) { printf qq<%vd\n>, $_ }"
<input ąść and press enter>
196.133.197.155.196.135.10
```
**Perl configuration**

It's broken on all perl versions including blead.

xenu:
Due to a bug in Windows, ReadFile() and ReadConsoleA() (and thus
_read()), return zeros instead of non-ASCII characters when the console
codepage is set to 65001. See this ticket for more details:
https://github.com/microsoft/terminal/issues/4551

This commit works around that bug by using ReadConsoleW() inside
win32_read() when the passed fd points to the console and the console
codepage is set to 65001.

Fixes #18701

### DIFF
--- a/win32/win32.c
+++ b/win32/win32.c
@@ -196,6 +196,10 @@ static const SYSTEMTIME time_t_epoch_base_systemtime = {
 
 #define FILETIME_CHUNKS_PER_SECOND (10000000UL)
 
+#ifdef USE_ITHREADS
+static perl_mutex win32_read_console_mutex;
+#endif
+
 #ifdef SET_INVALID_PARAMETER_HANDLER
 static BOOL silent_invalid_parameter_handler = FALSE;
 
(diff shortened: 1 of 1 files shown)


xenu:
I forgot about Ctrl-Z handling. Unlike ReadFile(), ReadConsole() doesn't do that automatically. It's fixed now.

tonycoz (approved):
Looks good to me.

My only concern is that we'll be calling win32_read() once for every unicode character input, and I *think* that will propagate up through to pp_sysread's read() implementation, pp_sysread will be calling PerlIO_read() for each character input.

I don't see a way to avoid that without making this much more complex.

This will only be a concern for pastes to the console though.

xenu:
win32_read_console() acts just like the standard read(), if you call it with a buffer of length 100, it will try to return 100 bytes. Unless, of course, ENABLE_LINE_INPUT is disabled, but that's a non-default configuration.

The internal algorithm of win32_read_console() is indeed extremely ineffective, but that should **not** propagate. It won't cause its callers to return after reading just a single character.

asfernandes:
Hi @xenu !

We are having the same problem in https://github.com/FirebirdSQL/firebird/issues/7202

Looks like your solution fix our problem.

We use a license similar to MPL. So it's not compatible with GPL, but I'm not sure about Artistic license.

Do you allow us to write our code basing in your solution with appropriate credits as comments in our code?

xenu:
@asfernandes Sure. You have my permission to use the code from this PR under the terms of the whatever license Firebird is using right now.

asfernandes:
> @asfernandes Sure. You have my permission to use the code from this PR under the terms of the whatever license Firebird is using right now.

Thank you!

Issue #18636, opened 2021-03-10
%ENV keys take string internals


FGasper:
```
perl -MData::Dumper -Mstrict -w -Mutf8 -e'my $foo = "é"; %ENV = (); $ENV{$foo} = 123; print Dumper \%ENV; print $$ . $/; exec "bash -c env"'
$VAR1 = {
          "\x{e9}" => '123'
        };
26970
é=123
PWD=/home/pi
SHLVL=0
_=/usr/bin/env
```
Note the `é=123`. Since $foo is code point 233 (0xe9), that env key should actually be the Latin-1 0xe9, not the 2 bytes of U+00E9 in UTF-8.

Latin-1 0xe9 is what happens when you assign $foo as the value rather than the key of %ENV:
```
perl -MData::Dumper -Mstrict -w -Mutf8 -e'my $foo = "é"; %ENV = (); $ENV{foo} = $foo; print Dumper \%ENV; print $$ . $/; exec "bash -c env"'
$VAR1 = {
          'foo' => '�'
        };
26975
PWD=/home/pi
foo=�
SHLVL=0
_=/usr/bin/env
```

jkeenan:
I'm trying to get a handle on your specific complaint here.  Is the problem in assignment to `%ENV` specifically?  Or to hashes in general?  Or is it the way Data::Dumper works?

Consider:
```
$ cat gh-18636-data-dump.pl
#!/usr/bin/env perl
use strict;
use warnings;
use Data::Dump qw(dd);
use utf8;

#perl -MData::Dumper -Mstrict -w -Mutf8 -e'my $foo = "é"; %ENV = (); $ENV{$foo} = 123; print Dumper \%ENV; print $$ . $/; exec "bash -c env"'

my %HASH = ( alpha => 'beta', gamma => 'delta');

print "\%HASH assignment\n\n";

for my $k (sort keys %HASH) { print "$k => $HASH{$k}\n"; }
print "\n";

my $foo = "é";
$HASH{$foo} = 123;
for my $k (sort keys %HASH) { print "$k => $HASH{$k}\n"; }
print "\n";

my $bar = "é";
$HASH{$bar} = 456;
for my $k (sort keys %HASH) { print "$k => $HASH{$k}\n"; }
print "\n";

dd(\%HASH);
print "\n";

##########

print "\%ENV assignment\n\n";

%ENV = ();

my $baz = "é";
$ENV{$baz} = 123;
[...]

PR #18706, opened 2021-04-10
Set %ENV keys using the same byte-string logic as setting %ENV values.

FGasper:
Issue #18636: This extends the work from
613c63b465f01af4e535fdc6c1c17e7470be5aad to %ENV keys. Previously
if you assigned an upgraded string as a key in %ENV, the string’s
internal PV representation was sent to the OS. Now the string is
“soft downgraded” before being given to the OS; if the downgrade
fails--i.e., if the string contains code points above 255--then
a warning is printed, and the string’s utf8 is assigned to the
environment, as happens with %ENV values.

A new internal macro, MgSV, is created to facilitate this work.

### DIFF
--- a/t/op/magic.t
+++ b/t/op/magic.t
@@ -5,7 +5,7 @@ BEGIN {
     chdir 't' if -d 't';
     require './test.pl';
     set_up_inc( '../lib' );
-    plan (tests => 192); # some tests are run in BEGIN block
+    plan (tests => 196); # some tests are run in BEGIN block
 }
 
 # Test that defined() returns true for magic variables created on the fly,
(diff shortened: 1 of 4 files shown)


t/op/magic.t
```diff
@@ -764,6 +764,11 @@ SKIP: {
 	$forced = $ENV{foo} = $chars;
 	ok(!utf8::is_utf8($forced) && $forced eq $bytes, 'ENV store downgrades utf8 in SV');
 	env_is(foo => $bytes, 'ENV store downgrades utf8 in setenv');
+	fail 'chars should still be wide!' if !utf8::is_utf8($chars);
+	$ENV{$chars} = 'widekey';
+	env_is("eh zero \x{A0}" => 'widekey', 'ENV store downgrades utf8 key in setenv');
```
khwilliamson, line 769: FYI for people coming after.  Using \x{A0} ignores EBCDIC.  But it happens to work here, because A0 in both systems requires two bytes to represent in UTF-8, which is all that is being tested here, and not the particular meaning of the code point.  But the pun "eh zero" doesn't apply, but no one should be bothered by that.  The test writing advice we give is to use B6 for things like this.  It is the only code point in all EBCDICs and Latin1 that requires two bytes to represent in UTF-8, and means the same thing everywhere: PARAGRAPH SEPARATOR

khwilliamson: And the test file used A0 already for a similar purpose, and since it works under both systems, never caused a problem, nor was noticed until I saw this just now

FGasper: Good to know … thank you!

PR #18707, opened 2021-04-10
Base refactoring with docs to some t/base/*.t files

kozachkov:
Added `use strict; use warnings` to some t/base/*.t files and comment descriptions about their test plans.

### DIFF
--- a/t/base/if.t
+++ b/t/base/if.t
@@ -1,9 +1,14 @@
 #!./perl
 
+# t/base/if.t - check "if"
+
+use strict;
+use warnings;
+
 print "1..2\n";
 
-# first test to see if we can run the tests.
+# First test to see if we can run the tests.
 
-$x = 'test';
-if ($x eq $x) { print "ok 1 - if eq\n"; } else { print "not ok 1 - if eq\n";}
-if ($x ne $x) { print "not ok 2 - if ne\n"; } else { print "ok 2 - if ne\n";}
+my $x = 'test';
+if ($x eq $x) { print "ok 1 - if eq\n"; } else { print "not ok 1 - if eq\n"; }
+if ($x ne $x) { print "not ok 2 - if ne\n"; } else { print "ok 2 - if ne\n"; }
--- a/t/base/if.t
+++ b/t/base/if.t
@@ -2,13 +2,10 @@
 
 # t/base/if.t - check "if"
 
-use strict;
-use warnings;
-
 print "1..2\n";
 
 # First test to see if we can run the tests.
 
-my $x = 'test';
+$x = 'test';
 if ($x eq $x) { print "ok 1 - if eq\n"; } else { print "not ok 1 - if eq\n"; }
 if ($x ne $x) { print "not ok 2 - if ne\n"; } else { print "ok 2 - if ne\n"; }
--- a/t/base/pat.t
+++ b/t/base/pat.t
@@ -1,9 +1,19 @@
 #!./perl
 
+# t/base/pat.t - check base regex for $_
+
+use strict;
+use warnings;
+
 print "1..2\n";
 
-# first test to see if we can run the tests.
+# First test to see if we can run the tests.
 
 $_ = 'test';
-if (/^test/) { print "ok 1 - match regex\n"; } else { print "not ok 1 - match regex\n";}
-if (/^foo/) { print "not ok 2 - match regex\n"; } else { print "ok 2 - match regex\n";}
+/^test/
+  ? ( print "ok 1 - match regex\n" )
+  : ( print "not ok 1 - match regex\n" );
+
+/^foo/
+  ? ( print "not ok 2 - match regex\n" )
+  : ( print "ok 2 - match regex\n" );
--- a/t/base/pat.t
+++ b/t/base/pat.t
@@ -2,9 +2,6 @@
 
 # t/base/pat.t - check base regex for $_
 
-use strict;
-use warnings;
-
 print "1..2\n";
 
 # First test to see if we can run the tests.
--- a/t/base/rs.t
+++ b/t/base/rs.t
@@ -36,9 +36,6 @@
 # Test $/ by reading file to "our" filehandle variable
 # Test $/ for files from variables like real files
 
-use strict;
-use warnings;
-
 # Test $/
 
 print "1..41\n";
@@ -47,9 +44,9 @@ print "1..41\n";
 # Preparation
 # ======================================================
 
-my $test_count = 1;
-my $teststring = "1\n12\n123\n1234\n1234\n12345\n\n123456\n1234567\n";
-my $teststring2 = "1234567890123456789012345678901234567890";
+$test_count = 1;
+$teststring = "1\n12\n123\n1234\n1234\n12345\n\n123456\n1234567\n";
+$teststring2 = "1234567890123456789012345678901234567890";
 
 # ------------------------------------------------------
 # Remove previous test files if exist
@@ -80,7 +77,7 @@ close TESTFILE or die "error $! $^E closing";
 # ---------------------------------------------
 # Check $/ for $test_string from the 'foo' file
 # ---------------------------------------------
-my $test_count_start = $test_count;  # Needed to know how many tests to skip
+$test_count_start = $test_count;  # Needed to know how many tests to skip
 open TESTFILE, "<./foo";
 binmode TESTFILE;
 test_string(*TESTFILE);
@@ -101,7 +98,7 @@ open TESTFILE, "<./foo";
 binmode TESTFILE;
 test_record(*TESTFILE);
 close TESTFILE;
-my $test_count_end = $test_count;  # Needed to know how many tests to skip
+$test_count_end = $test_count;  # Needed to know how many tests to skip
 
 # -----------------------------------
 # $/ preserved when set to bad value
@@ -137,7 +134,7 @@ if ( $^O eq 'VMS' ) {
     print CREATEFILE "\$EXIT\n";
     close CREATEFILE;
 
-    my $throwaway = `\@\[\]foo`, "\n";
+    $throwaway = `\@\[\]foo`, "\n";
 
     open( TEMPFILE, ">./foo.bar" ) or die "# open failed $! $^E\n";
     print TEMPFILE "foo\nfoobar\nbaz\n";
@@ -147,7 +144,7 @@ if ( $^O eq 'VMS' ) {
 
     $/ = \10;
 
-    my $bar = <TESTFILE>;
+    $bar = <TESTFILE>;
     $bar eq "foo\n"
       ? ( print "ok $test_count\n" )
       : ( print "not ok $test_count\n" );
@@ -180,7 +177,7 @@ if ( $^O eq 'VMS' ) {
 else {
     # Nobody else does this at the moment (well, maybe OS/390, but they can
     # put their own tests in) so we just punt
-    foreach my $test ( $test_count .. $test_count + 3 ) {
+    foreach $test ( $test_count .. $test_count + 3 ) {
         print "ok $test # skipped on non-VMS system\n";
         $test_count++;
     }
@@ -197,7 +194,7 @@ $/ = "\n";
 # For "our" variable
 {
     if ( open our $T, "./foo" ) {
-        my $line = <$T>;
+        $line = <$T>;
         print "# $line\n";
         # length of $teststring2
         length($line) == 40 or print "not ";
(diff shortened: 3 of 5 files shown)


jkeenan:
On behalf of perl5-porters, I would like to thank you for a first-time contribution to the Perl 5 core distribution.

Unfortunately, we won't be able to accept this pull request in its current form.  Here's why.

The test suite is organized into various directories with certain directories -- including `t/base/` -- being reserved for very fundamental concepts that don't depend on our being able to load modules via `use` or `require`.  That means that files in `t/base/` may not contain any `use strict;` or `use warnings;` statements.  See `perldoc perlhack`.  (As a side effect, this means that `git blame` attributes many statements in these test files to Larry Wall's initial commits to Perl 1 back in 1987.)

If perl had strictures or warnings active *by default* (as was proposed for [one version/vision of Perl 7](https://github.com/Perl/perl5/wiki/Defaults-for-v7)), then it's quite possible that the files you have worked on might have to be re-worked to compile with strictures and run warnings-free.  But that's not the current perl's expectation.

We do, however, look forward to your future contributions.

Thank you very much.
Jim Keenan

kozachkov:
@jkeenan Thank you for the response. I removed unnecessary `use strict; use warnings` from these files.

Is the PR is acceptable now in the form of two different commits?

Leont:
> Is the PR is acceptable now in the form of two different commits?

No, it is not. You're changing a bunch of things without good reason, and adding comments that aren't necessarily good value. I would strongly recommend against touching this part of the test suite unless you truly understand what it's trying to test.

hvds:
I'll echo Jim's thanks for your efforts to improve perl here. Based on a quick look, I'm sure there are parts that we would like. But there are quite a few types of changes mixed in here, and it would be easier to review if they were separated out as distinct commits.

The various comments you have added look mostly helpful, and much of the reformatting looks good too (though I'm not personally a fan of the extra spaces in parens you appear to like). I'm less sure about your choice to replace various `if (cond) { ... } else { ... }` constructs with `cond ? ... : ...` - the latter are in principle harder to parse, so in `t/base` when we are still checking whether the absolute basics of the language work correctly we prefer to avoid them (except of course when that is what we are trying to test).

t/base/if.t
```diff
@@ -1,9 +1,11 @@
...
+
 print "1..2\n";
 
-# first test to see if we can run the tests.
+# First test to see if we can run the tests.
 
 $x = 'test';
-if ($x eq $x) { print "ok 1 - if eq\n"; } else { print "not ok 1 - if eq\n";}
-if ($x ne $x) { print "not ok 2 - if ne\n"; } else { print "ok 2 - if ne\n";}
+if ($x eq $x) { print "ok 1 - if eq\n"; } else { print "not ok 1 - if eq\n"; }
```
karenetheridge, line 10: these changes add unwanted trailing whitespace.

t/base/pat.t
```diff
@@ -1,9 +1,16 @@
...
 print "1..2\n";
 
-# first test to see if we can run the tests.
+# First test to see if we can run the tests.
 
 $_ = 'test';
-if (/^test/) { print "ok 1 - match regex\n"; } else { print "not ok 1 - match regex\n";}
-if (/^foo/) { print "not ok 2 - match regex\n"; } else { print "ok 2 - match regex\n";}
+/^test/
+  ? ( print "ok 1 - match regex\n" )
```
karenetheridge, line 11: the `?:` operators should be reserved for strict functional expressions -- that is, they should not be used solely to achieve side effects.

t/base/rs.t
```diff
@@ -1,312 +1,455 @@
...
 $teststring2 = "1234567890123456789012345678901234567890";
 
-# Create our test datafile
-1 while unlink 'foo';                # in case junk left around
+# ------------------------------------------------------
+# Remove previous test files if exist
+# ------------------------------------------------------
+# In case of existing files with the same 'foo' name
+# on some systems with versioning file system (e.g VMS).
+# Read more at pod/perport.pod "System Interaction"
```
karenetheridge: typo in filename... and I'm not sure if the explanation in this comment is correct.

t/base/rs.t
```diff
@@ -1,312 +1,455 @@
...
-  close TESTFILE;
-  1 while unlink qw(foo.bar foo.com foo.fdl);
-} else {
-  # Nobody else does this at the moment (well, maybe OS/390, but they can
-  # put their own tests in) so we just punt
-  foreach $test ($test_count..$test_count + 3) {
-      print "ok $test # skipped on non-VMS system\n";
-      $test_count++;
-  }
+if ( $^O eq 'VMS' ) {
```
karenetheridge, line 95: reformatting large parts of a file for whitespace-only changes means that 'git blame' now reports the wrong commit, which causes more work for bisection and other archaeological endeavours.

karenetheridge (changes requested):
I do not see that there is sufficient value gained, in proportion to the volume of changes.

t/base/if.t
kozachkov: What about symmetry ?

t/base/pat.t
kozachkov: I see. I will return the form.

t/base/rs.t
kozachkov: The typo, yes, thank you. What explanation should be used here instead?

t/base/rs.t
kozachkov: In perlhack.pod there are such lines:

> Use spaces for indentation, not tab characters.
> 
> The codebase is a mixture of tabs and spaces for indentation, and we
> are moving to spaces only.  Converting lines you're patching from 8-wide
> tabs to spaces will help this migration.

That means that potentially there are a lot of **similar places** in perl core where should be done such reformatting "which cases more work for bisection and other archaeological endeavours". It cannot be avoided it we want to have a modern (not slovenly) perl code in the repo.

kozachkov:
> > Is the PR is acceptable now in the form of two different commits?
> 
> No, it is not. You're changing a bunch of things without good reason, and adding comments that aren't necessarily good value. I would strongly recommend against touching this part of the test suite unless you truly understand what it's trying to test.

What is a good reason for you? For me it is:

- a clean indentation and appearance at whole;
- a way to simplify reading test suits just by overview their description especially for newbies, we do want to extend of potentially perl developers count, doesn't it?
- an adding a documentation notes about what is going on and why. In perlhack.pod there are such parts:

> While it's not necessary for documentation changes, new tests or
> trivial patches, it's often worth explaining how your change works.
> Even if it's clear to you today, it may not be clear to a porter next
> month or next year.

"it may not be clear to a porter next month or next year." -- and there are such places there which are "not be clear" and I added a few comments about how it works.

Code and documentation must be improving -- that is "a good reason" for me.

>  I would strongly recommend against touching this part of the test suite unless you truly understand what it's trying to test.

And what if I "truly understand"?

As regards "against touching this part". Well, there are you who can point to mistakes of such changes and there is a chance that some changes may be useful for Perl Community, doesn't it?

kozachkov:
@hvds "But there are quite a few types of changes mixed in here, and it would be easier to review if they were separated out as distinct commits." you right about "mixed changes". I will try to do it more separate in the future.

Leont:
> a clean indentation and appearance at whole;

_Clean indentation and appearance_ is highly subjective. If we all update the source code to what we think is clean, our git history is going to look like a wikipedia edit war. In general, changing indentation of a piece of code that you have no ownership over is bad form IMHO.

Likewise, adding your personal convention of `=====` and `-----` headers when none of these files did so before seems rather inappropriate. It's generally better to stay in the style of the existing code/documentation.

"Talk first, shoot second" is generally speaking a good principle.

> a way to simplify reading test suits just by overview their description especially for newbies, we do want to extend of potentially perl developers count, doesn't it?

I agree with that in theory, I don't agree with this implementation. I don't think these overviews clarify anything.

> an adding a documentation notes about what is going on and why. In perlhack.pod there are such parts:

Prepending «Create test 'foo' file» before a simple open/print/close sequence does not achieve any of that. It doesn't answer any question a competent perl programmer would have about that bit of code.

It's not that I think that the answers you provide are incorrect, it's that they answer the wrong questions; questions that are useless because the answer can easily be deduced from basic reading of the code. Good documentation answers questions that can't be answered that way (typically these are why questions).

> And what if I "truly understand"?

> As regards "against touching this part". Well, there are you who can point to mistakes of such changes and there is a chance that some changes may be useful for Perl Community, doesn't it?

In general, I don't think an outsider is the right person to answer the questions, I do think they can be the right person to ask questions (e.g. "why isn't any of this using strict or warnings" would have been a great question).

rwp0:
### suggestion

> Tests in the t/base/ directory ought to be runnable with plain miniperl.That is, they should not require Config.pm nor should they require anyextensions to have been built. TEST will abort if any tests in thet/base/ directory fail.

clarify further in [README](https://www.github.com/Perl/perl5/tree/blead/t%2FREADME) linking to `perlhack(1)` (on `strict` / `warnings`) or create (perhaps by subdividing the main one) a new one specific to the subdirectory (ie. `t/base`)

Leont:
> or create (perhaps by subdividing the main one) a new one specific to the subdirectory (ie. t/base)

Yeah, I think that would be very useful.

kozachkov:
> > a clean indentation and appearance at whole;
> 
> _Clean indentation and appearance_ is highly subjective. If we all update the source code to what we think is clean, our git history is going to look like a wikipedia edit war. In general, changing indentation of a piece of code that you have no ownership over is bad form IMHO.
> 
> Likewise, adding your personal convention of `=====` and `-----` headers when none of these files did so before seems rather inappropriate. It's generally better to stay in the style of the existing code/documentation.
> 
> "Talk first, shoot second" is generally speaking a good principle.
> 
> > a way to simplify reading test suits just by overview their description especially for newbies, we do want to extend of potentially perl developers count, doesn't it?
> 
> I agree with that in theory, I don't agree with this implementation. I don't think these overviews clarify anything.
> 
> > an adding a documentation notes about what is going on and why. In perlhack.pod there are such parts:
> 
> Prepending «Create test 'foo' file» before a simple open/print/close sequence does not achieve any of that. It doesn't answer any question a competent perl programmer would have about that bit of code.
> 
> It's not that I think that the answers you provide are incorrect, it's that they answer the wrong questions; questions that are useless because the answer can easily be deduced from basic reading of the code. Good documentation answers questions that can't be answered that way (typically these are why questions).
> 
> > And what if I "truly understand"?
> 
> > As regards "against touching this part". Well, there are you who can point to mistakes of such changes and there is a chance that some changes may be useful for Perl Community, doesn't it?
> 
> In general, I don't think an outsider is the right person to answer the questions, I do think they can be the right person to ask questions (e.g. "why isn't any of this using strict or warnings" would have been a great question).

@Leont

I agree that:

- "Clean indentation and appearance is highly subjective." because of that there are pod/perlstyle.pod and pod/perlhack.pod with rules for such situations.
- "If we all update the source code to what we think is clean, our git history is going to look like a wikipedia edit war." that doesn't refer situations which are explained in aforesaid pod files.
- "Likewise, adding your personal convention of ===== and ----- headers when none of these files did so before seems rather inappropriate. It's generally better to stay in the style of the existing code/documentation." You are absolutely right, that's my fault and I removed it.

I disagree that:

- "In general, changing indentation of a piece of code that you have no ownership over is bad form IMHO." in open-source projects with many contributors must be common rules for common practice for it and in perl repo there are such ones in pod/.
- "I agree with that in theory, I don't agree with this implementation. I don't think these overviews clarify anything." you are right about "I don't think" cause you are already know these parts of test code.
- "Prepending «Create test 'foo' file» before a simple open/print/close sequence does not achieve any of that. It doesn't answer any question a competent perl programmer would have about that bit of code." If you look more carefully into the code which was before my changes you will see that this line there are there already. And if we follow it, that we should removed these line from the tests.
- "It's not that I think that the answers you provide are incorrect, it's that they answer the wrong questions; questions that are useless because the answer can easily be deduced from basic reading of the code" it is a "highly subjective" statement because in most cases there is no necessity to write documentation comments as an experienced perl programmer can understand parts of code just by reading them, doesn't it?
- "In general, I don't think an outsider is the right person to answer the questions, I do think they can be the right person to ask questions (e.g. "why isn't any of this using strict or warnings" would have been a great question)." and another "great question" I guess should be "why in this test code it is used indentation with 2 spaces instead of 4 as it's advised in pod/perlstyle.pod?

t/base/pat.t
kozachkov: I found the same using `?:` for tests t/base/num.t however.

jkeenan:
This p.r. has not found good support from core committers.  I propose that we close it in 7 days unless such support emerges.

jkeenan:
> This p.r. has not found good support from core committers. I propose that we close it in 7 days unless such support emerges.

Closing as per schedule.

PR #18709, opened 2021-04-10
Support %#p format

### LINKED ISSUE #18289: Perl's internal printf "%#p" format does not prefix with "0x..."
Platform's printf() prefixes "0x..." on integers and pointers with the `#` flag:

```c
printf("%#x", 15);  // 0xf
printf("%#p", (void *)15); // 0xf
```

Perl's `warn()` uses its own internal implementation which does not do this for `%#p`.

It would be useful if it did.

dmah42:
Prior to this change, the `%#p` format was considered 'invalid'. This change adds support for it similarly to `%#x` and adds a test to confirm it works as intended.

Fixes #18289

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -12472,7 +12472,7 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
 
 	    uv = PTR2UV(args ? va_arg(*args, void*) : argsv);
 	    base = 16;
-            c = 'x';
+            c = 'x';    /* in case the format string contains '#' */
 	    goto do_integer;
 
 	case 'c':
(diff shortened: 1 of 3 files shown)


sv.c
```diff
@@ -12474,6 +12472,7 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
 
 	    uv = PTR2UV(args ? va_arg(*args, void*) : argsv);
 	    base = 16;
+            c = 'x';
```
hvds: I think it's useful to comment the purpose of this, as I did in my example code.

hvds:
Other than my comment (about my comment), this looks good to me.

sv.c
dmah42: Added the comment, and this comment about the comment. :)

PR #18710, opened 2021-04-11
Perl_pad_alloc: refactor away calls to av_fetch()

richardleach:
_Perl_pad_alloc_ has two calls to `av_fetch()` that can be refactored away,
hopefully helping to speed up compilation of Perl applications.

A set of assumptions is used:
* `PL_comppad` is never `SvMAGICAL` or `SvREADONLY`
* `PL_comppad` is always `AvREAL`
* The _key_ values ( `AvFILLp(PL_comppad) + 1` and `retval`) are never negative

These assumptions allow many branches within `av_fetch()` and then `av_store()`
to be pruned, allowing refactoring of the call sites.

1. `sv = *av_fetch(PL_comppad, AvFILLp(PL_comppad) + 1, TRUE);`
is first reduced to:
`sv = *av_store(PL_comppad, AvFILLp(PL_comppad) + 1, newSV(0));`
and then to:
```
        SSize_t key = AvFILLp(PL_comppad) + 1;
        if (key > AvMAX(PL_comppad))
            av_extend(PL_comppad,key);

        sv = AvARRAY(PL_comppad)[key] = newSV(0);

        AvFILLp(PL_comppad) = key;
```

2. `sv = *av_fetch(PL_comppad, retval, TRUE);`
is first reduced to:
```
            sv = ( retval >= AvFILLp(PL_comppad) + 1 ||
                               !AvARRAY(PL_comppad)[retval] )
                    ? *av_store(PL_comppad, retval, newSV(0))
                    : AvARRAY(PL_comppad)[retval]
                    ;
```
and then to:
```
            if ( retval > AvFILLp(PL_comppad) ) {
                if (retval > AvMAX(PL_comppad))
                    av_extend(PL_comppad,retval);

                AvFILLp(PL_comppad) = retval;
                AvARRAY(PL_comppad)[retval] = newSV(0);

            } else if (!AvARRAY(PL_comppad)[retval] ) {

                SvREFCNT_dec(AvARRAY(PL_comppad)[retval]);
                AvARRAY(PL_comppad)[retval] = newSV(0);

            }
            sv = AvARRAY(PL_comppad)[retval];
```
(Although all tests still pass following these commits, merging after any
approval should probably still wait until the next development cycle.)


richardleach:
Note: under the assumptions above, `av_extend_guts()` could be called instead of `av_extend()`.

Also, if this PR isn't considered appropriate, what about an `av_fetch_fast()` and `av_store_fast()` that can be called when the assumptions hold? (For comparison, `Perl_padnamelist_store` also seems to be a cut-down `av_store()` for a specific use case. Possibly that function would become redundant if a compatible `av_store_fast()` came into existence.)

hvds:
> _Perl_pad_alloc_ has two calls to `av_fetch()` that can be refactored away, hopefully helping to speed up compilation of Perl applications.
>
> A set of assumptions is used:
> 
> * `PL_comppad` is never `SvMAGICAL` or `SvREADONLY`
> 
> * `PL_comppad` is always `AvREAL`
> 
> * The _key_ values ( `AvFILLp(PL_comppad) + 1` and `retval`) are never negative

Those sound like quite aggressive assumptions; I'd be looking for at least one (and preferably both) of @tonycoz and @iabyn to confirm how safe they are.

richardleach:
> > _Perl_pad_alloc_ has two calls to `av_fetch()` that can be refactored away, hopefully helping to speed up compilation of Perl applications.
> > A set of assumptions is used:
> > 
> > * `PL_comppad` is never `SvMAGICAL` or `SvREADONLY`
> > * `PL_comppad` is always `AvREAL`
> > * The _key_ values ( `AvFILLp(PL_comppad) + 1` and `retval`) are never negative
> 
> Those sound like quite aggressive assumptions; 

Yeah. They do seem to hold, but I don't have a complete understanding of pad lifecycles, so:

> I'd be looking for at least one (and preferably both) of @tonycoz and @iabyn to confirm how safe they are.

Me too!

tonycoz:
> A set of assumptions is used:
> 
>     * `PL_comppad` is never `SvMAGICAL` or `SvREADONLY`
> 
>     * `PL_comppad` is always `AvREAL`
> 
>     * The _key_ values ( `AvFILLp(PL_comppad) + 1` and `retval`) are never negative

If you're making assumptions you should ASSUME() on them, this will assert on DEBUGGING builds if they don't hold, and hint to the compiler 

> Also, if this PR isn't considered appropriate, what about an av_fetch_fast() and av_store_fast() that can be called when the assumptions hold?

I wonder if an inline av_fetch()/av_store() could work - handling the simple (no magic, positive index(?), no extend) case inline and handing off to a heavy version if more is needed.

With appropriate ASSUME()s it might drop the code for the complex cases and do what you're doing now, without the opaqueness of direct inlining.

richardleach:
Thanks @tonycoz 

> If you're making assumptions you should ASSUME() on them, this will assert on DEBUGGING builds if they don't hold, and hint to the compiler

Will do.

In general, is it safe to assume that pad AVs are always just vanilla AVs?

> I wonder if an inline av_fetch()/av_store() could work - handling the simple (no magic, positive index(?), no extend) case inline and handing off to a heavy version if more is needed.
> 
> With appropriate ASSUME()s it might drop the code for the complex cases and do what you're doing now, without the opaqueness of direct inlining.

I'll give that a try and see what gcc/clang do with it.

richardleach:
Hmm, this inlining+ASSUME approach didn't seem to work with gcc 8.3.0.

I put the following into `inline.h`:

```
SV**
Perl_av_store_simple(pTHX_ AV *av, SSize_t key, SV *val)
{
    SV** ary;

    PERL_ARGS_ASSERT_AV_STORE_SIMPLE;
    assert(SvTYPE(av) == SVt_PVAV);

    if (!SvMAGICAL(av) && !SvREADONLY(av) && AvREAL(av) && key > -1) {
        ary = AvARRAY(av);

        if (AvFILLp(av) < key) {
            AvFILLp(av) = key;
            if (UNLIKELY(key > AvMAX(av))) {
                av_extend(av,key);
                ary = AvARRAY(av);
            }
        }
        else
            SvREFCNT_dec(ary[key]);

        ary[key] = val;
        return &ary[key];
    } else {
        return av_store(av, key, val);
    }
}
```


```
SV**
Perl_av_fetch_simple(pTHX_ AV *av, SSize_t key, I32 lval)
{
    PERL_ARGS_ASSERT_AV_FETCH_SIMPLE;
    assert(SvTYPE(av) == SVt_PVAV);

    if (!SvMAGICAL(av) && !SvREADONLY(av) && AvREAL(av) && key > -1) {

        if ( (key >= AvFILLp(av) + 1) || !AvARRAY(av)[key]) {
            return lval ? av_store_simple(av,key,newSV(0)) : NULL;
        } else {
            return &AvARRAY(av)[key];
        }

    } else {
        return av_fetch(av, key, lval);
    }
}
```

And changed one of the `av_fetch` calls in `Perl_pad_alloc `to an `av_fetch_simple `call:
```
            ASSUME(!SvMAGICAL(PL_comppad));
            ASSUME(!SvREADONLY(PL_comppad));
            ASSUME(AvREAL(PL_comppad));
            ASSUME(AvFILLp(PL_comppad) + 1 > -1);
            ASSUME(retval > -1);
            sv = *av_fetch_simple(PL_comppad, retval, TRUE);

```
But when I look at the disassembly for `Perl_pad_alloc`, the two functions have been inlined but not simplified at all - there's are calls for the full-fat `av_fetch` and `av_store` in there.

If I look at _pad.i_, the ASSUME statements are represented by instances of `((void) (0))`; has the compiler processed the ASSUMEs before it has inlined the functions?

Or is the best that gcc can do to order the inlined function in an optimized manner, but not actually use the ASSUME statements to resolve branches at compilation time? (From googling, `__builtin_expect` is apparently "weaker" than `assume`.)

I tried a build with clang-7 as well, but that didn't even inline the functions!?

The embed.fnc entries are:
```
CipdR	|SV**	|av_fetch_simple|NN AV *av|SSize_t key|I32 lval
Cipd	|SV**	|av_store_simple|NN AV *av|SSize_t key|NULLOK SV *val

```
Maybe changing i -> I will make a difference to clang?

richardleach:
As an aside, I noticed that perl.h defines ASSUME in terms of `__builtin_unreachable` for both _clang_ and _gcc_, but clang also has `__builtin_assume`. Would there be any benefit in using the latter for clang builds that support it?

richardleach:
Ok, changing i->I in _embed.fnc_ does make clang inline the function, but it also doesn't use the `ASSUME`s to cut down the inlined functions.

richardleach:
Or, am I just doing it wrong? 
1. Should I be writing `if (x) { ASSUME() }` to get `__builtin_unreachable` used?
2.  `HAS_BUILTIN_UNREACHABLE` is [only defined on DEBUGGING builds](https://github.com/Perl/perl5/blob/blead/perl.h#L3883) anyway? In which case, non-DEBUGGING builds may be missing out on some compiler optimizations.

khwilliamson:
On 5/30/21 10:16 AM, Richard Leach wrote:
> Or, am I just doing it wrong?
> 
>  1. Should I be writing |if (x) { ASSUME() }| to get
>     |__builtin_unreachable| used?
>  2. |HAS_BUILTIN_UNREACHABLE| is only defined on DEBUGGING builds
>     <https://github.com/Perl/perl5/blob/blead/perl.h#L3883> anyway?
> 
> 

There is a bug in the defines for these.  I'll have a PR shortly you can try

richardleach:
Thanks @khwilliamson. I was just trying out this rearrangement:

```
#if __has_builtin(__builtin_unreachable)
#  define HAS_BUILTIN_UNREACHABLE
#elif (defined(__GNUC__) && (   __GNUC__ > 4                              \
                               || __GNUC__ == 4 && __GNUC_MINOR__ >= 5))
#  define HAS_BUILTIN_UNREACHABLE
#endif

#if __has_builtin(__builtin_assume)
#  define HAS_BUILTIN_ASSUME     
#endif

#ifdef DEBUGGING
#  define ASSUME(x) assert(x)
#else 
#  if defined(__sun) || (defined(__hpux) && !defined(__GNUC__))
#    define ASSUME(x)      /* ASSUME() generates warnings on Solaris */
#    define NOT_REACHED
#  elif defined(HAS_BUILTIN_ASSUME)
#    define ASSUME(x) __builtin_assume(x)
#    define NOT_REACHED
#  elif defined(HAS_BUILTIN_UNREACHABLE)
#    define ASSUME(x) ((x) ? (void) 0 : __builtin_unreachable())
#    define NOT_REACHED                                                       \
        STMT_START {                                                        \
            ASSUME(!"UNREACHABLE"); __builtin_unreachable();                \
        } STMT_END
#  else
#    if defined(_MSC_VER)
#      define ASSUME(x) __assume(x)
#    elif defined(__ARMCC_VERSION) /* untested */
#      define ASSUME(x) __promise(x)
#    else
    /* a random compiler might define assert to its own special optimization
     * token so pass it through to C lib as a last resort */
#      define ASSUME(x) assert(x)
#    endif
#    define NOT_REACHED ASSUME(!"UNREACHABLE")
#  endif
#endif
```

For _clang_, that does expand my ASSUMEs in _pad.i_ (non-DEBUGGING build) to the following:
```
        __builtin_assume(!((PL_comppad)->sv_flags & (0x00200000|0x00400000|0x00800000)));
        __builtin_assume(!((PL_comppad)->sv_flags & (0x08000000|0x00010000)));
        __builtin_assume(((PL_comppad)->sv_flags & 0x40000000));
        __builtin_assume(((XPVAV*) (PL_comppad)->sv_any)->xav_fill + 1 > -1);
```
_clang_ seems to make use of them to optimize the inlined `av_fetch_simple`, but not for the inlining of `av_store_simple` after that.

richardleach:
For _gcc_ (again, non-DEBUGGING) with those revised definitions, the ASSUMEs are expanded to:
```
        ((!((PL_comppad)->sv_flags & (0x00200000|0x00400000|0x00800000))) ? (void) 0 : __builtin_unreachable());
        ((!((PL_comppad)->sv_flags & (0x08000000|0x00010000))) ? (void) 0 : __builtin_unreachable());
        ((((PL_comppad)->sv_flags & 0x40000000)) ? (void) 0 : __builtin_unreachable());
        ((((XPVAV*) (PL_comppad)->sv_any)->xav_fill + 1 > -1) ? (void) 0 : __builtin_unreachable());
```
But _gcc_ wasn't able to make use of them to remove the calls to `av_fetch `or `av_store` in `av_fetch_simple`/`av_store_simple`.

richardleach:
For the record, same _clang_/_gcc_ behaviours as above using @khwilliamson's macro definitions from https://github.com/Perl/perl5/pull/18839.

richardleach:
When I next have time, I'll see what happens if I repeat the ASSUME statements inside `Perl_av_fetch_simple`, to see if that results in `Perl_av_store_simple` then being optimized.

richardleach:
> When I next have time, I'll see what happens if I repeat the ASSUME statements inside `Perl_av_fetch_simple`, to see if that results in `Perl_av_store_simple` then being optimized.

Yeah, that didn't work.

From clang-12, there's a separate [assume](https://releases.llvm.org/12.0.0/tools/clang/docs/AttributeReference.html#assume) attribute that might DWIM: "_This is similar to __builtin_assume but instead of an expression that can be assumed to be non-zero, the assumption is expressed as a string and it holds for the entire function._"

But since clang-12 or -13 are going to be a small proportion of builds, I'm wondering if this isn't a productive avenue for now. Perhaps I should either:
1. Stick with the original PR, but add better comments to show where the inlining came from, or
2. Do use the inline functions, but do not make them capable of punting to the full-fat functions

@tonycoz / @khwilliamson / @hvds : Thoughts?

tonycoz:
Sorry about the rabbit hole I sent you down.

If the compiler isn't optimizing based on the assumes, please try the non-punting inline functions.

These should assert on their pre-conditions (no magic, non-negative index, etc)

richardleach:
> Sorry about the rabbit hole I sent you down.
> 
> If the compiler isn't optimizing based on the assumes, please try the non-punting inline functions.
> 
> These should assert on their pre-conditions (no magic, non-negative index, etc)

No problem, it was interesting to see what the compilers would do. I'll work on the non-punting inline functions as recommended.

richardleach:
Just checking, for the inline functions, should they:
1. Be in _inline.h_ or in _pad.c_? (I can see other locations where they might be useful, so I think _inline.h_ is preferable.)
2. Be `S_` or `Perl_` prefixed?

I tripped over the fact that _autodoc.pl_ is unhappy if something in _inline.h_ has the `I` flag. Should that be permitted?

tonycoz:
In inline.h I think, where they should be Perl_ prefixed (since they're visible to anything that includes inline.h)

WRT the I flag, are they being inlined without it?

richardleach:
> In inline.h I think, where they should be Perl_ prefixed (since they're visible to anything that includes inline.h)

Thanks, I'll update the PR shortly.

> WRT the I flag, are they being inlined without it?

Yes, the non-punting versions are inlined with just the `i` flag in embed.fnc.

richardleach:
Ok folks, how does this PR look now?

hvds:
> Ok folks, how does this PR look now?

For the avoidance of doubt, and with apologies for the delay: I do plan to look at this, but have not yet been able to make time to do so.

inline.h
```diff
@@ -57,6 +57,98 @@ Perl_av_count(pTHX_ AV *av)
...
+=for apidoc av_store_simple
+
+This is a cut-down version of av_store that assumes that the array is
+very straightforward - no magic, not readonly, and AvREAL - and that
+C<key> is not negative. This function MUST NOT be used in situations
+where any of those assumptions may not hold.
+
+If any of those conditions actually turns out
+to be true, this function will punt to av_store. (Assuming that the
+checks haven't been optimized away due to ASSUME()s at the call site.)
```
tonycoz: This is no longer true.

tonycoz (changes requested):

inline.h
richardleach: Thanks Tony. I've pushed an update deleting that comment block.

inline.h
```diff
@@ -57,6 +57,94 @@ Perl_av_count(pTHX_ AV *av)
...
+    assert(key > -1);
+
+    ary = AvARRAY(av);
+
+    if (AvFILLp(av) < key) {
+        AvFILLp(av) = key;
+        if (key > AvMAX(av)) {
+            av_extend(av,key);
+            ary = AvARRAY(av);
+        }
```
tonycoz: If the key is very large, av_extend() can throw a catchable error, AvFILL() should only be updated after av_extend() returns.

tonycoz (changes requested):
sorry, I missed that last time

inline.h
richardleach: Thanks, updated.

tonycoz (approved):

inline.h
```diff
@@ -57,6 +57,94 @@ Perl_av_count(pTHX_ AV *av)
...
+
+This is a cut-down version of av_store that assumes that the array is
+very straightforward - no magic, not readonly, and AvREAL - and that
+C<key> is not negative. This function MUST NOT be used in situations
+where any of those assumptions may not hold.
+
+Stores an SV in an array.  The array index is specified as C<key>.  The
+return value will be C<NULL> if the operation failed or if the value did not
+need to be actually stored within the array (as in the case of tied
+arrays).  Otherwise, it can be dereferenced
```
hvds: I don't think there's actually any way for this to return NULL - all possible failures look to be fatal, and it can't be a tied array - so a chunk of this documentation could be simplified.

inline.h
```diff
@@ -57,6 +57,94 @@ Perl_av_count(pTHX_ AV *av)
...
+to get the C<SV*> that was stored
+there (= C<val>)).
+
+Note that the caller is responsible for suitably incrementing the reference
+count of C<val> before the call, and decrementing it if the function
+returned C<NULL>.
+
+Approximate Perl equivalent: C<splice(@myarray, $key, 1, $val)>.
```
hvds, line 76: Isn't it more like `$myarray[$key] = $val`? The behaviour beyond the end certainly seems so:
```
% perl -wle '@a=(1,2); $a[3]= 3; print join "-", @a'
Use of uninitialized value within @a in join or string at -e line 1.
1-2--3
% perl -wle '@a=(1,2); splice @a, 3, 1, 3; print join "-", @a'
splice() offset past end of array at -e line 1.
1-2-3
```

inline.h
```diff
@@ -57,6 +57,94 @@ Perl_av_count(pTHX_ AV *av)
...
+
+This is a cut-down version of av_fetch that assumes that the array is
+very straightforward - no magic, not readonly, and AvREAL - and that
+C<key> is not negative. This function MUST NOT be used in situations
+where any of those assumptions may not hold.
+
+Returns the SV at the specified index in the array.  The C<key> is the
+index.  If lval is true, you are guaranteed to get a real SV back (in case
+it wasn't real before), which you can then modify.  Check that the return
+value is non-null before dereferencing it to a C<SV*>.
```
hvds, line 119: I think this should also mention, similarly to store_simple, that refcount management is the caller's responsibility.

It might also be worth mentioning that one should not invoke any perl code (ie anything that could modify the av) between calling this and reading the SV**.

inline.h
```diff
@@ -57,6 +57,94 @@ Perl_av_count(pTHX_ AV *av)
...
+Perl_av_fetch_simple(pTHX_ AV *av, SSize_t key, I32 lval)
+{
+    PERL_ARGS_ASSERT_AV_FETCH_SIMPLE;
+    assert(SvTYPE(av) == SVt_PVAV);
+    assert(!SvMAGICAL(av));
+    assert(!SvREADONLY(av));
+    assert(AvREAL(av));
+    assert(key > -1);
+
+    if ( (key >= AvFILLp(av) + 1) || !AvARRAY(av)[key]) {
```
hvds: Why the +1?
```
  if (key > AvFILLp(av) || !AvARRAY(av)[key]) {
```

(I think this could actually be a bug in rare circumstances, if fill hits `SSize_t_MAX`.)

pad.c
```diff
@@ -712,7 +712,7 @@ Perl_pad_alloc(pTHX_ I32 optype, U32 tmptype)
         pad_reset();
     if (tmptype == SVs_PADMY) { /* Not & because this ‘flag’ is 0.  */
         /* For a my, simply push a null SV onto the end of PL_comppad. */
-        sv = *av_fetch(PL_comppad, AvFILLp(PL_comppad) + 1, TRUE);
+        sv = *av_store_simple(PL_comppad, AvFILLp(PL_comppad) + 1, newSV(0));
```
hvds, line 715: Might be worth clarifying in the commit message that replacing a fetch with a store here really is intentional.

In any case, the first line of the commit message is rather longer than recommended; I suggest something more like:
```
Perl_pad_alloc: simplify av_fetch() calls

The nature of PL_comppad allows this.
```

hvds (commented):
Sorry to come in so late; the doc changes are optional and can be made later, but it would be good to fix the +1 and the commit message before merging.

inline.h
richardleach: Yes, agreed. Too much copy and paste from `av_store` left in.

inline.h
richardleach: `Approximate Perl equivalent: C<splice(@myarray, $key, 1, $val)>` is straight out of `av_store`'s documentation. Are you suggesting that instance should also be revised?

inline.h
richardleach: Just thinking out loud: for both functions, given the cut-down nature, would there be any circumstances in which the caller would have to decrement the ref count?

The original `av_fetch` docs don't warn about not doing anything that could modify the av prior to reading the SV**. I guess that any added statement should be replicated there too. Is that a real concern for perl code though, rather than any core/XS users of `av_fetch`/`_simple`?

inline.h
richardleach: It came from me refactoring the following, when we know that `neg` is untrue:
```
    neg  = (key < 0);
    size = AvFILLp(av) + 1;
    key += neg * size; /* handle negative index without using branch */

    /* the cast from SSize_t to Size_t allows both (key < 0) and (key >= size)
     * to be tested as a single condition */
    if ((Size_t)key >= (Size_t)size) {
        if (UNLIKELY(neg))
            return NULL;
        goto emptyness;
    }

    if (!AvARRAY(av)[key]) {
      emptyness:
        return lval ? av_store(av,key,newSV(0)) : NULL;
    }
```
I removed the cast, thinking that it wasn't necessary, but I now think you're right about things going astray if AvFILLp == SSize_t_MAX. (Nice spot.)

So then it should be as you suggest:

`if (key > AvFILLp(av) || !AvARRAY(av)[key]) {` ?

Or since the function assumes `AvREAL(av)`, would it actually be enough just to say:

`if (!AvARRAY(av)[key]) {` ?

pad.c
richardleach: Thanks, I've changed it to your recommendation.

inline.h
hvds: > `Approximate Perl equivalent: C<splice(@myarray, $key, 1, $val)>` is straight out of `av_store`'s documentation. Are you suggesting that instance should also be revised?

Ah, looking at the history of that line, I see the reasoning was described in the commit message (17b0bd7713):
```
    * improve  some of the "perl equivalent" entries; for example
      av_store() is *not* like $myarray[$key] = $val, since it replaces the
      stored SV with a different SV, rather than just updating the current
      SV's value.
```

So I guess leave them both using `splice` as the equivalent for now, I'll think later whether it benefits anyone to add more nuance.

inline.h
hvds: > Just thinking out loud: for both functions, given the cut-down nature, would there be any circumstances in which the caller would have to decrement the ref count?

Not for errors, I think; however the caller may or may not need to increment.

> The original `av_fetch` docs don't warn about not doing anything that could modify the av prior to reading the SV**. I guess that any added statement should be replicated there too. Is that a real concern for perl code though, rather than any core/XS users of `av_fetch`/`_simple`?

My concern is that we don't explicitly say that the `SV**` returned is a pointer directly into the `AvARRAY`, so just based on the documentation this doesn't seem obvious to me. I'd consider it a safer function if it just returned an `SV*`.

inline.h
hvds: > Or since the function assumes `AvREAL(av)`, would it actually be enough just to say:
> 
> `if (!AvARRAY(av)[key]) {` ?

I don't think so, you'd be reading past the bounds of the array.

richardleach: Derp, yes. Brain has disengaged for the night.

inline.h
richardleach: > I think this should also mention, similarly to store_simple, that refcount management is the caller's responsibility.

Looking at this again, does `av_fetch_simple` really need this? Note that `lval` here is basically a flag, not a SV*.



> My concern is that we don't explicitly say that the `SV**` returned is a pointer directly into the `AvARRAY`, so just based on the documentation this doesn't seem obvious to me. I'd consider it a safer function if it just returned an `SV*`.

This is what `av_fetch` does though and I wanted `av_fetch_simple` to be a drop-in replacement.

inline.h
richardleach: I'll change it to `if (key > AvFILLp(av) || !AvARRAY(av)[key]) { `

inline.h
hvds: > > I think this should also mention, similarly to store_simple, that refcount management is the caller's responsibility.
> 
> Looking at this again, does `av_fetch_simple` really need this?

Need? No. Would the docs be improved by such a change? I'm inclined towards yes, but I'm by no means certain - I do understand that too much doc can be as bad as too little.

> Note that `lval` here is basically a flag, not a SV*.

Yeah, that also took me a few beats to understand, and I had considered suggesting renaming it to `want_lval` (and perhaps re-typing it to bool). But if that's valuable it should be a separate change that also covers av_fetch.

> > My concern is that we don't explicitly say that the `SV**` returned is a pointer directly into the `AvARRAY`, so just based on the documentation this doesn't seem obvious to me. I'd consider it a safer function if it just returned an `SV*`.
> 
> This is what `av_fetch` does though and I wanted `av_fetch_simple` to be a drop-in replacement.

Yes, I'm not actually suggesting to change the return type. And again, if the doc is worth clarifying in this way it would be sensible to do it for av_fetch too.

As I mentioned at the start, I think it would be fine to consider all these doc changes separately after merging this. I'd be happy to put together a PR for that.

pad.c
richardleach: Update pushed.

inline.h
richardleach: Update pushed.

inline.h
richardleach: Update pushed.

inline.h
richardleach: I've not made any changes here.

inline.h
richardleach: I've not made any changes here.

richardleach:
@hvds - I've made a number of updates following your comments, but left some of the documentation changes for a possible future PR by yourself. Please let me know if anything further needs attention.

hvds:
> @hvds - I've made a number of updates following your comments, but left some of the documentation changes for a possible future PR by yourself. Please let me know if anything further needs attention.

Yep, looks good to me. I'll merge this tomorrow if @tonycoz doesn't have additional comments.

hvds:
Now merged as 84c7520439 and 310f47b681.

PR #18712, opened 2021-04-12
Replaced pod/ rt.perl.org tickets links to github.com issues links

kozachkov:
For the task from the issue #17331 I replaced all rt.perl.org ticket links to their github.com issue links with issue numbers in square brackets [# 12345] (if it's not ok [# 12345] can be easily replaced to any form of issue number now).

### DIFF
--- a/pod/perl5184delta.pod
+++ b/pod/perl5184delta.pod
@@ -45,12 +45,12 @@ filehandle information is properly reset after a pager is run. [perl #121456]
 =item *
 
 Introduced by
-L<perl #113536|https://rt.perl.org/Public/Bug/Display.html?id=113536>, a memory
+L<[GH #12161]|https://github.com/Perl/perl5/issues/12161>, a memory
 leak on every call to C<system> and backticks (C< `` >), on most Win32 Perls
 starting from 5.18.0 has been fixed.  The memory leak only occurred if you
 enabled pseudo-fork in your build of Win32 Perl, and were running that build on
 Server 2003 R2 or newer OS.  The leak does not appear on WinXP SP3.
-[L<perl #121676|https://rt.perl.org/Public/Bug/Display.html?id=121676>]
+L<[GH #13741]|https://github.com/Perl/perl5/issues/13741>
 
 =back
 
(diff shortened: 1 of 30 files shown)


Grinnz:
Please use the form: `[GH #00000]`

pod/perl5184delta.pod
```diff
@@ -45,12 +45,12 @@ filehandle information is properly reset after a pager is run. [perl #121456]
...
 
 Introduced by
-L<perl #113536|https://rt.perl.org/Public/Bug/Display.html?id=113536>, a memory
+L<[#12161]|https://github.com/Perl/perl5/issues/12161>, a memory
 leak on every call to C<system> and backticks (C< `` >), on most Win32 Perls
 starting from 5.18.0 has been fixed.  The memory leak only occurred if you
 enabled pseudo-fork in your build of Win32 Perl, and were running that build on
 Server 2003 R2 or newer OS.  The leak does not appear on WinXP SP3.
-[L<perl #121676|https://rt.perl.org/Public/Bug/Display.html?id=121676>]
+[L<[#13741]|https://github.com/Perl/perl5/issues/13741>]
```
Grinnz: These are already within square brackets, there should only be one set of brackets either inside or outside the link (moving them inside would be consistent with elsewhere).

Grinnz: Same with the links in perl5200delta, perl5320delta

kozachkov: Aha, I see, will fix. Thank you.

kozachkov:
@Grinnz changed [#0000] as [GH #0000] and moved the brackets.

pod/perl5200delta.pod
```diff
@@ -3591,7 +3591,7 @@ Fixed a crash when destroying a self-referencing GLOB.  [perl #121242]
 =item *
 
 L<IO::Socket> is known to fail tests on AIX 5.3.  There is
-L<a patch|https://rt.perl.org/Ticket/Display.html?id=120835> in the request
+L<[GH #13484]|https://github.com/Perl/perl5/issues/13484> in the request
```
Grinnz: Link label got changed here

Grinnz:
LGTM other than last comment, thanks for the effort!

Leont:
Sounds like a good idea to me

pod/perl5200delta.pod
kozachkov: Fixed it.

Grinnz (approved):

xenu:
LGTM, but could you squash it into a single commit?

kozachkov:
> LGTM, but could you squash it into a single commit?

@xenu I did.

PR #18716, opened 2021-04-14
Moved 2 vars in perlvar to deprecated section

kozachkov:
Moved `${^ENCODING}` and `${^WIN32_SLOPPY_STAT}` to Deprecated section \+ `s/\t/\s+/g`

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -371,7 +371,7 @@ it really means
 
 But don't put
 
-    @foo{$x,$y,$z}	# a slice--note the @
+    @foo{$x,$y,$z}     # a slice--note the @
 
 which means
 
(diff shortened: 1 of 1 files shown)


pod/perlvar.pod
```diff
@@ -835,14 +828,14 @@ by this bit of code:
     sub show_n { print "\$1 is $1; \$2 is $2\n" }
 
     {
-    OUTER:
-	show_n() if $outer =~ m/$pattern/;
+        OUTER:
```
Grinnz: This label seems to be intentionally outdented

kozachkov: Fixed it.

Grinnz: It should remain at the same level as the opening brace.

kozachkov: Ah, I see. Restored.

Grinnz (approved):
Some of the examples were originally using Ratliff style for the closing brace, but this is heavily against perlstyle and inconsistent with the rest of the document so should be changed anyway. LGTM

PR #18717, opened 2021-04-15
Workaround early gcc bug with macros containing ((x) | 0)

### LINKED ISSUE #18655: Unable to complete perl 5.32.1 build on OpenBSD luna88k
**Description**

Attempting to build perl on an OpenBSD luna88k throws an exception when trying to use `miniperl` to `pack 'W*', 168, 769` building Unicode::Normalize. 

http://www.openbsd.org/luna88k.html

Note that this is with the OpenBSD system perl build infrastructure and patches, but we will try to reproduce on a download of p5p's 5.32.1.

If you have suggestions on where to look, I'd appreciate it.  

```
LD_LIBRARY_PATH=/w1/o/cur/src/gnu/usr.bin/perl/obj ./miniperl -Ilib make_ext.pl lib/auto/Unicode/Normalize/Normalize.so           
MAKE="make" LIBPERL_A=libperl.so.22.0 LINKTYPE=dynamic                                                                            
Making header files for XS...                                                                                                     
Malformed UTF-8 character: \xa8 (unexpected continuation byte 0xa8, with no preceding start byte) in pack at ./mkheader           
line 44.                                                                                                                          
Malformed UTF-8 character (fatal) at ./mkheader line 44.                                                                          
Unsuccessful Makefile.PL(dist/Unicode-Normalize): code=65280 at make_ext.pl line 536.                                             
*** Error 2 in /usr/src/gnu/usr.bin/perl (makefile:576 'lib/auto/Unicode/Normalize/Normalize.so')                                 
```

**Steps to Reproduce**

```
LD_LIBRARY_PATH=... miniperl -Ilib -C -E 'say pack "W*", 168, 769' # on an OpenBSD luna88k machine
```

**Expected behavior**

The build completes and I get a working perl, similar to the way it worked with perl 5.30.3.

Specifically that `pack "W*", 168, 769` would return a DIAERESIS with COMBINING ACUTE ACCENT (΅), not throw an exception, the way it does with perl 5.30.3 on the same machine.

```
$ /usr/bin/perl -C -E 'say pack("W*", 168, 769)' | hexdump -C                                                    
00000000  c2 a8 cc 81 0a                                    |.....|                                                               
00000005                                                                                                                          
```

**Perl configuration**

```
# perl -V output goes here as soon as I get it, unfortunately I don't have one of these machines someone in Japan is reporting this remotely.
```

khwilliamson:
Fixes #18655

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -1384,9 +1384,17 @@ or casts
 #   define WIDEST_UTYPE U32
 #endif
 
-/* Use this, where there could be some confusion, as a static assert in macros
- * to make sure that a parameter isn't a pointer. */
-#define ASSERT_NOT_PTR(x) ((x) | 0)
+/* Where there could be some confusion, use this as a static assert in macros
+ * to make sure that a parameter isn't a pointer.  But some compilers can't
+ * handle this.  The only one known so far that doesn't is gcc 3.3.6; the check
+ * below isn't thorough for such an old compiler, so may have to be revised if
+ * experience so dictates. */
+#if  ! defined(__GNUC__)                                                    \
+ || (__GNUC__ > 3 || (__GNUC__ == 3 && __GNUC_MINOR__ > 3))
+#  define ASSERT_NOT_PTR(x) ((x) | 0)
+#else
+#  define ASSERT_NOT_PTR(x) (x)
+#endif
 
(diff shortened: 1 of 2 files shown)


hvds:
Looks reasonable to me.

PR #18719, opened 2021-04-16
Adjust Module::CoreList version in Maintainers.pl

toddr:
This reflects the version that's on CPAN right now.


bingos:
Thumbs up from me.

PR #18727, opened 2021-04-19
Do not cache utf8 offsets for non-canonical lengths

### LINKED ISSUE #18588: Search for empty substring with `index` returns different results with utf8 magic
Searching for an empty substring with `index` returns different results based on the presence or absence of `utf8` (`w`) magic.

Minimal code to reproduce:

```perl
use strict;
use warnings;
use feature qw( say );

my $s = "abc";

my $len = length($s);  # Must come before upgrade.

utf8::upgrade($s) if $ARGV[0];

say index($s, "", $len+0);  # Required.
say index($s, "", $len+1);
```

```
$ perl a.pl 0
3
3

$ perl a.pl 1
3
4
```

Tested with 5.32.0.

Leont:
In particular, if the length is beyond the end, it should not be stored as the end.

This fixes #18588

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -3095,7 +3095,7 @@ S	|int	|sv_2iuv_non_preserve	|NN SV *const sv
 SR	|STRLEN	|expect_number	|NN const char **const pattern
 ST	|STRLEN	|sv_pos_u2b_forwards|NN const U8 *const start \
 		|NN const U8 *const send|NN STRLEN *const uoffset \
-		|NN bool *const at_end
+		|NN bool *const at_end|NN bool *canonical_position
 ST	|STRLEN	|sv_pos_u2b_midway|NN const U8 *const start \
 		|NN const U8 *send|STRLEN uoffset|const STRLEN uend
 S	|STRLEN	|sv_pos_u2b_cached|NN SV *const sv|NN MAGIC **const mgp \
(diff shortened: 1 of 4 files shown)


hvds:
Looking through this commit, I'm slightly confused - when is `canonical_position` ever true?

Leont:
> Looking through this commit, I'm slightly confused - when is `canonical_position` ever true?

`sv_pos_u2b_forwards` contains this piece of code.
````
while (s < send && uoffset) {
    --uoffset;
    s += UTF8SKIP(s);
}
````
So either it stops because we're at the right offset, or because we're at the end of the string (or both, but that's really just a special case of the former).

In the former case, canonical_position would be true, in the latter it wouldn't.

hvds:
> > Looking through this commit, I'm slightly confused - when is `canonical_position` ever true?
> 
> `sv_pos_u2b_forwards` contains this piece of code.
> 
> ```
> while (s < send && uoffset) {
>     --uoffset;
>     s += UTF8SKIP(s);
> }
> ```
> 
> So either it stops because we're at the right offset, or because we're at the end of the string (or both, but that's really just a special case of the former).
> 
> In the former case, canonical_position would be true, in the latter it wouldn't.

Thanks, I see my mistake now - I had (repeatedly) read it as `*canonical_position = uoffset = 0`, rolling two assignments into a single statement rather than `*canonical_position = uoffset == 0`, a single assignment of the result of a comparison.

In part it was natural for me to read it the wrong way because I would always write such a construct as `*canonical_position = (uoffset == 0)`; I think it would be a good idea to add the parens in this case too.

khwilliamson:
I had the same initial reaction as @hvds, but did spot the == when I looked more slowly.  Parens would have helped me see it immediately.  I would be more comfortable with a comment.  Unless a bug fix is face-palm worthy, it wasn't obviously a problem, and won't be again 6 months from now

Leont:
Fair enough, I'll add it

PR #18729, opened 2021-04-19
Perl_clear_defarray: faster array creation via new macro+function

richardleach:
It's moderately common in core to see the SV* array for an AV created in the
following sort of ways:

```
        AV *av = newAV();
        av_extend(av, key);
```
or
```
        AV * av = newAV();
        av_push(av, some_sv);
```

or
```
        AV * av = newAV();
        av_store(av, 0, some_sv);
```

In each case, array creation ultimately occurs via `av_extend()` and involves
steps that we can reason are unnecessary:
* Check if it's a tied array - we know it isn't
* Call `av_extend_guts()`
* Check` if (key < -1)` - we know it isn't (or can make sure it isn't)
* Check `if (key > *maxp)` - we know it is, because there's no array yet
* Check `if (av && *allocp != *arrayp)` - i.e. is there a shifted array.
                         - we know there isn't, because there's no array yet
* Check `if (*allocp)` - if an unshifted array exists
                         - we know it doesn't , because there's no array yet
* Check `if key < 3`, if so set` key == 3`

Some parts of core already work around this (although sometimes because they
really want less than the minimum key size of 3) by calling `Newx/Newxz` directly.
Examples are:
* `av_make` does it obviously: https://github.com/Perl/perl5/blob/blead/av.c#L421
* https://github.com/Perl/perl5/blob/blead/pad.c#L242 and
https://github.com/Perl/perl5/blob/blead/pad.c#L2596 are variations.
* `new_stackinfo` sort of does it: https://github.com/Perl/perl5/blob/blead/scope.c#L76

This PR does two things:
* Introduces a function and some macros around newAV() + SV* array alloc.
* Modifies `Perl_clear_defarray()` to show an example of usage.

Names and implementation are just best-efforts, happy to change upon review!

### DIFF
--- a/av.c
+++ b/av.c
@@ -418,7 +418,7 @@ Perl_av_new_alloc(pTHX_ SSize_t size, bool zeroflag)
     AV * const av = newAV();
     SV** ary;
     PERL_ARGS_ASSERT_AV_NEW_ALLOC;
-    assert(size > 1);
+    assert(size > 0);
 
     Newx(ary, size, SV*); /* Newx performs the memwrap check */
     AvALLOC(av) = ary;
--- a/av.c
+++ b/av.c
@@ -394,6 +394,44 @@ Perl_av_store(pTHX_ AV *av, SSize_t key, SV *val)
 }
 
 /*
+=for apidoc av_new_alloc
+
+Creates a new AV and allocates its SV* array.
+
+This is similar to but more efficient than doing:
+
+    AV *av = newAV();
+    av_extend(av, key);
+
+The zeroflag parameter controls whether the array is NULL initialized.
+
+Note that av_index() takes the desired AvMAX as its key parameter, but
+av_new_alloc() instead takes the desired size (so AvMAX + 1). This
+size must be at least 1.
+
+=cut
+*/
+
+AV *
+Perl_av_new_alloc(pTHX_ SSize_t size, bool zeroflag)
+{
+    AV * const av = newAV();
+    SV** ary;
+    PERL_ARGS_ASSERT_AV_NEW_ALLOC;
+    assert(size > 1);
+
+    Newx(ary, size, SV*); /* Newx performs the memwrap check */
+    AvALLOC(av) = ary;
+    AvARRAY(av) = ary;
+    AvMAX(av) = size - 1;
+
+    if (zeroflag)
+        Zero(ary, size, SV*);
+
+    return av;
+}
+
+/*
 =for apidoc av_make
 
 Creates a new AV and populates it with a list of SVs.  The SVs are copied
--- a/av.h
+++ b/av.h
@@ -109,5 +109,45 @@ Perl equivalent: C<my @array;>.
 #define newAV()	MUTABLE_AV(newSV_type(SVt_PVAV))
 
 /*
+=for apidoc newAV_alloc_x
+
+Similar to newAV(), but a SV* array is also allocated.
+
+This is similar to but more efficient than doing:
+
+    AV *av = newAV();
+    av_extend(av, key);
+
+Note that the actual size requested is allocated. This is unlike
+av_extend(), which takes the maximum desired array index (AvMAX) as its
+"key" parameter, and enforces a minimum value for that of 3.
+
+In other words, the following examples all result in an array that can
+fit four elements (indexes 0 .. 3):
+
+    AV *av = newAV();
+    av_extend(av, 1);
+
+    AV *av = newAV();
+    av_extend(av, 3);
+
+    AV *av = newAV_alloc_x(4);
+
+Whereas this will result in an array that can only fit one element:
+
+    AV *av = newAV_alloc_x(1);
+
+newAV_alloc_x does not initialize the array with NULL pointers.
+newAV_alloc_xz does do that initialization.
+
+These macros MUST NOT be called with a size less than 1.
+
+=cut
+*/
+
+#define newAV_alloc_x(size)  av_new_alloc(size,0)
+#define newAV_alloc_xz(size) av_new_alloc(size,1)
+
+/*
  * ex: set ts=8 sts=4 sw=4 et:
  */
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -4966,8 +4966,6 @@ PP(pp_leavesub)
 void
 Perl_clear_defarray(pTHX_ AV* av, bool abandon)
 {
-    const SSize_t fill = AvFILLp(av);
-
     PERL_ARGS_ASSERT_CLEAR_DEFARRAY;
 
     if (LIKELY(!abandon && SvREFCNT(av) == 1 && !SvMAGICAL(av))) {
@@ -4975,8 +4973,9 @@ Perl_clear_defarray(pTHX_ AV* av, bool abandon)
         AvREIFY_only(av);
     }
     else {
-        AV *newav = newAV();
-        av_extend(newav, fill);
+        const SSize_t size = AvFILLp(av) + 1;
+        /* The ternary gives consistency with av_extend() */
+        AV *newav = newAV_alloc_x(size < 4 ? 4 : size);
         AvREIFY_only(newav);
         PAD_SVl(0) = MUTABLE_SV(newav);
         SvREFCNT_dec_NN(av);
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -4975,6 +4975,7 @@ Perl_clear_defarray(pTHX_ AV* av, bool abandon)
     else {
         const SSize_t size = AvFILLp(av) + 1;
         /* The ternary gives consistency with av_extend() */
+        /* fill can be -1. The ternary gives consistency with av_extend() */
         AV *newav = newAV_alloc_x(size < 4 ? 4 : size);
         AvREIFY_only(newav);
         PAD_SVl(0) = MUTABLE_SV(newav);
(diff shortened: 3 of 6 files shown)


richardleach:
Notes:
1. If accepted, I'll submit separate PRs for each function where it seems that unnecessary CPU cycles can be avoided.
2. https://github.com/Perl/perl5/pull/18696 could be altered to use the new macros/functions also. I've left it is for now, since that PR was submitted first, but am happy to revise.
3. Design-wise, I'm not sure if whether it might be better to do without the macros and/or not provide the option to not Zero the array.

richardleach:
Gah. Just fixing the cat-across-the-keyboard reason for the CI build failure...

av.c
```diff
@@ -393,6 +393,39 @@ Perl_av_store(pTHX_ AV *av, SSize_t key, SV *val)
     return &ary[key];
 }
 
+/*
+=for apidoc av_new_alloc
```
hvds, line 397: As well as the implementation of `zeroflag`, this has some other differences to the newAV+av_extend equivalent that I think should either be corrected or documented. I notice at least:
- lower bound check (size < 0)
- upper bound check (memwrap)
- defaulting to minimum (max(size, 4))

My first guess is that the lack of default minimum should be documented, but the bounds checks should be added.

av.h
```diff
@@ -108,6 +108,26 @@ Perl equivalent: C<my @array;>.
...
+Similar to newAV(), but the SV* array is also allocated.
+
+This is similar to but more efficient than doing:
+
+    AV *av = newAV();
+    av_extend(av, key);
+
+Note that these the actual size requested is allocated. This is unlike
+av_extend(), which enforces a minimum size of 3, and always does a +1
+to the requested/minimum size when allocating.
```
hvds: It adds 1 to the _key_, but that's not the "requested/minimum size" but the index of the maximum element. So the minimum key is 3, but minimum size is 4 (covering indices 0-3).

av.h
```diff
@@ -108,6 +108,26 @@ Perl equivalent: C<my @array;>.
...
+=for apidoc newAV_alloc_x
+
+Similar to newAV(), but the SV* array is also allocated.
+
+This is similar to but more efficient than doing:
+
+    AV *av = newAV();
+    av_extend(av, key);
+
+Note that these the actual size requested is allocated. This is unlike
```
hvds: s/that these the/that the/

pp_hot.c
```diff
@@ -4975,8 +4975,8 @@ Perl_clear_defarray(pTHX_ AV* av, bool abandon)
         AvREIFY_only(av);
     }
     else {
-        AV *newav = newAV();
-        av_extend(newav, fill);
+        /* fill can be -1. The ternary gives consistency with av_extend() */
+        AV *newav = newAV_alloc_xz(fill < 3 ? 4 : fill);
```
hvds: This looks wrong, because `newAV_alloc_xz()` takes a size rather than a key - it should be `fill < 3 ? 4 : fill + 1`. For cleanliness, it would probably be better instead to set `const SSize_t size = AvFILLp(av) + 1; ... newAV_alloc_xz(size < 4 ? 4 : size)`

Note also that the original `av_extend()` call zero-filled the array only because the `AvREIFY_only()` is applied late; a small optimization therefore would be to use `newAV_alloc_x()` instead. (But since that's a change of behaviour, it's probably worth calling out with a separate commit.)

av.c
richardleach: Yes, I had intended that people using this are certain that they will be calling with a known, non-zero size. I will improve the documentation and add an assert for debugging builds.

I thought that Newx - i.e. safesysmalloc - checks for memwrap?

av.h
richardleach: Will fix.

av.h
richardleach: Yes, good point. I'll try to improve the wording.

pp_hot.c
richardleach: Gah, you're right. Managed to get confused by key vs size myself.

The point about `AvREIFY_only()` is a good one. I'll double check: might well have left it out since it was a change, but half recollect trying to just call `newAV_alloc_x()` and being surprised when that broke things.

av.c
hvds: > Yes, I had intended that people using this are certain that they will be calling with a known, non-zero size. I will improve the documentation and add an assert for debugging builds.

Ok.

> I thought that Newx - i.e. safesysmalloc - checks for memwrap?

Ah yes, you're quite right - so the `MEM_WRAP_CHECK_s` in `av_extend_guts` is actually a duplicate, presumably only for the purpose of giving a more specific error message. Maybe that could be combined into a `Newx_s` macro to avoid the duplication.

richardleach: Huh, _safesysmalloc_ does not check for memwrap. I think I intended that people using this function are certain that they are using a non-zero, positive integer that is small enough to never wrap! But I'd initially envisaged the usage being for small, hardcoded sizes, which is obviously not the case for the pp_hot.c function changed in this commit.

I'll add a check for memwrap.

richardleach: D'oh, I must have looked more closely the first time around! Let me look again.

richardleach: Yes, I misremembered where the memwrap check was. (i.e. The `Newx` macros, not the `safesysmalloc` function.)

So I propose to not add an extra check into this PR and we open a separate issue for the `Newx_s` macro you suggest. Sound okay?

hvds: > So I propose to not add an extra check into this PR and we open a separate issue for the `Newx_s` macro you suggest. Sound okay?

Yes.

pp_hot.c
richardleach: I was thinking about some other refactoring breaking with `newAV_alloc_x()`. I've added the change to use that as a separate commit (to remain unsquashed at the end).

richardleach:
@hvds - Have I addressed your review points sufficiently, or do you recommend any more changes?

As far as eventual squashing goes, I was thinking of ending up with three commits:
* One to introduce the new macros/function
* One to change `Perl_clear_defarray` to use `newAV_alloc_xz` (i.e. no change in Zeroing behaviour)
* One to change `newAV_alloc_xz` to `newAV_alloc_x` (the change to no array Zeroing by itself)

hvds:
> @hvds - Have I addressed your review points sufficiently, or do you recommend any more changes?

You've addressed my points excellently; however:
- the assert in Perl_av_new_alloc should be `assert(size >= 1)`
- the `/* fill can be -1. ... */` comment in Perl_clear_defarray no longer makes sense: I suggest removing that fragment, and just leaving `The ternary gives consistency ...`.

> As far as eventual squashing goes, I was thinking of ending up with three commits:
> 
>     * One to introduce the new macros/function
> 
>     * One to change `Perl_clear_defarray` to use `newAV_alloc_xz` (i.e. no change in Zeroing behaviour)
> 
>     * One to change `newAV_alloc_xz` to `newAV_alloc_x` (the change to no array Zeroing by itself)

Yes, that sounds great.

richardleach:
@hvds - thanks for the additional feedback. I've made changes and hopefully squashed correctly.

hvds:
@richardleach I don't think the commit rearrangement worked quite right. I've tried to fix it to what I think you intended - please check branch [hv/clear_defarray](https://github.com/Perl/perl5/tree/hv/clear_defarray) to see if you agree.

* moved the assert fix to the commit that introduced it
* ensured that the last commit only changes *_xz() to *_x()
* (cosmetic/doc) modified the name of the macro parameters in embed.fnc from `key` to `size`
* modified last commit message to shorten the subject line
* also rebased against latest blead

richardleach:
@hvds - thanks for paying such close attention to the details. Yes, your branch does look better than my squashing.

hvds:
Super, now merged as 0b1c19ab1c, d2a9e96003 and 97b09acf9d.

iabyn:
On Wed, May 26, 2021 at 10:10:27AM -0700, Hugo van der Sanden wrote:
> Super, now merged as 0b1c19ab1c, d2a9e96003 and 97b09acf9d.

I find the docs for the new av_new_alloc() function to be very confusing.
Specifically:

    Note that av_index() takes the desired AvMAX as its key parameter, but
    av_new_alloc() instead takes the desired size (so AvMAX + 1). This
    size must be at least 1.

It's suddenly going off on a tangent about two functions/macros not
mentioned before, i.e. av_index() and AvMAX(). I guess it's trying to make
a point about max index != size, but it is (alas) not very clear.

How about just saying something to the effect that the size parameter
causes AvARRAY to be pre-allocated to hold at least elements 0..(size-1) ?


-- 

richardleach:
> How about just saying something to the effect that the size parameter causes AvARRAY to be pre-allocated to hold at least elements 0..(size-1) ?

Sure, how about this, to go above the sentence about the zeroflag parameter?
`The size parameter is used to pre-allocate a SV* array large enough to
hold elements 0..(size-1).`

iabyn:
How about

The size parameter is used to pre-allocate a SV* array large enough to hold at least elements 0..(size-1). size must be at least 1.

We shouldn't guarantee that it won't allocate more than the requested number of elements, for efficiency purposes.

richardleach:
> The size parameter is used to pre-allocate a SV* array large enough to hold at least elements 0..(size-1). size must be at least 1.

Ok. #18959 raised to apply that wording. Thanks for the feedback.

PR #18736, opened 2021-04-21
Fix spaces

jidanni:
Say either
- integer / floating / string, or
- integer/floating/string

not
- integer/floating/ string

### DIFF
--- a/pod/perlnumber.pod
+++ b/pod/perlnumber.pod
@@ -129,7 +129,7 @@ used may lead to breakage of some of the above rules.
 =head1 Flavors of Perl numeric operations
 
 Perl operations which take a numeric argument treat that argument in one
-of four different ways: they may force it to one of the integer/floating/
+of four different ways: they may force it to one of the integer / floating /
 string formats, or they may behave differently depending on the format of
 the operand.  Forcing a numeric value to a particular format does not
 change the number stored in the value.


karenetheridge:
No, this is not correct.

karenetheridge:
Neither of those cases exist. There is a line break in the middle, but that is not the same as introducing a regular 0x20 space.

Grinnz:
Line breaks do get collapsed to a space character by POD parsers, unfortunately. So this change might improve rendering.

jidanni:
Yes. It looks like doo doo on Debian. That's why I'm here.

khwilliamson:
I agree that something should be done.  I think that this would be best, so that formatters would be clear as to what is meant
`of four different ways: they may force it to one of the
integerE<sol>floatingE<sol>string formats, or they may behave differently
`

Grinnz:
Would that not also need a line break and thus introduce whitespace?

khwilliamson:
I don't understand your point.  I actually think that the three items are most clearly displayed as an unbroken sequence on a single line.  

> and/or
for example is best understood unseparated

Grinnz:
Yes, I mean it would need a line break in the POD because otherwise the line would be too long for pod tests. A line break in the POD gets rendered as a space character.

But it could just be split into more lines so that no line break is needed within this term.

oodler577:
Or you can learn how to use commas properly with a helpful semi-colon: 

```
of four different ways: they may force it to one of the integer, floating,
or string formats; or they may behave differently depending on the format of
the operand.
```

jidanni:
Yes, please use @oodler577 's version. Thanks.

jkeenan:
> Yes, please use @oodler577 's version. Thanks.

@oodler577's version implemented in https://github.com/Perl/perl5/pull/18806.  Please review.

jkeenan:
This issued should be resolved as of ffa5817eaa.  Closing ticket.

PR #18738, opened 2021-04-21
workaround unexpected op trees generated by Data::Alias

tonycoz:
A fix is also required to Data::Alias to fix compilation under -DDEBUGGING builds,

A more recent (apparently parser) issue means Data::Alias still won't pass its tests.


tonycoz:
https://rt.cpan.org/Ticket/Display.html?id=130156#txn-1976186

I've supplied a second patch to Data::Alias that I think fixes the problem there, this has the bonus that it fixes the problem for 5.32.0 as well.

tonycoz:
Obsoleted by changes to Data::Alias

PR #18740, opened 2021-04-21
Test new warnings on use 5.033

ashermancinelli:
Hoping to move #18666 along - @Leont @leonerd could you:

* Point me to where documentation should be touched up (I assume somewhere in `perlop.pod`?)
* Suggest additions/changes to the test
* Let me know if this pr is inappropriate/malformed in some way

Cheers! Thanks for the help @leonerd!

### DIFF
--- a/t/op/ver.t
+++ b/t/op/ver.t
@@ -12,7 +12,7 @@ $DOWARN = 1; # enable run-time warnings now
 
 use Config;
 
-plan( tests => 52 );
+plan( tests => 54 );
 
 eval 'use v5.5.640';
 is( $@, '', "use v5.5.640; $@");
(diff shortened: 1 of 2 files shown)


t/op/ver.t
```diff
@@ -267,5 +267,27 @@ sub { $_[0] = v3;
 *{"\3"} = *DATA;
 is( (readline v3), "This is what we expect to see!\n", "v-strings even work in Mordor" );
 
+$DOWARN = 0; # disable warnings just for the following test
```
leonerd: Since `$DOWARN` appears to be a package variable it would be better to surround the entire test sequence in a block, and put

```perl
local $DOWARN = 0;
```
at the top, as this will avoid the need to remember to reset it to 1 again afterwards.

t/op/ver.t
```diff
@@ -267,5 +267,27 @@ sub { $_[0] = v3;
...
+
+# Add a new warning handler to ensure warnings are enabled by default when
+# C<use v5.033> is in play.
+my $handler = $SIG{__WARN__};
```
leonerd: Additionally then this can be avoided; just use

```perl
local $SIG{__WARN__} = ...
```
to set it later

ashermancinelli:
@leonerd Thanks for the feedback!

t/op/ver.t
ashermancinelli: I believe I've resolved these in `ac4a599` - lmk if it's not what you had in mind.

Leont:
The author test is still failing, because you added a different email address than the one you're using for your git commits.

Also, it probably makes sense to squash the first and third commit together.

ashermancinelli:
I've updated the authors and revised the history - thanks @Leont!

Leont:
> Point me to where documentation should be touched up (I assume somewhere in perlop.pod?)

That is actually a bit of a difficult question, to be honest :-/

PR #18746, opened 2021-04-25
op.h: define missing BASEOP comment block fields (op_sibparent, op_targ)

richardleach:
This PR adds descriptions of the `op_sibparent` and `op_targ` fields to the `BASEOP` comment block at the top of _op.h_.

Doubt they are the best possible descriptions, so happy to apply suggestions.

### DIFF
--- a/op.h
+++ b/op.h
@@ -15,7 +15,13 @@
  *			but this is replaced when op is grafted in, when
  *			this op will point to the real next op, and the new
  *			parent takes over role of remembering starting op.)
+ * 	op_sibparent    Pointer to the op's next sibling, or to the parent
+ *			if there are no more siblings.
  *	op_ppaddr	Pointer to current ppcode's function.
+ * 	op_targ		An index into the current pad, identifying an SV
+ *			that is typically used to store the OP's result
+ *			(such as a lexical variable, or a SVs_PADTMP
+ *			temporary intermediate value).
  *	op_type		The type of the operation.
  *	op_opt		Whether or not the op has been optimised by the
  *			peephole optimiser.


Leont:
This title doesn't describe the change well, it does something very different from what I expected (a change to BASEOP itself)

richardleach:
Re-titled, hopefully better.

op.h
```diff
@@ -15,7 +15,12 @@
  *			but this is replaced when op is grafted in, when
  *			this op will point to the real next op, and the new
  *			parent takes over role of remembering starting op.)
+ * 	op_sibparent    Pointer to the op's next sibling, or to the parent
+ *			if there are no more siblings (PERL_OP_PARENT builds)
```
tonycoz: All builds are PERL_OP_PARENT now, so I'm not sure it's worth mentioning.

op.h
```diff
@@ -15,7 +15,12 @@
...
  *	op_ppaddr	Pointer to current ppcode's function.
+ * 	op_targ         Points to the pad position into which the result of
+ *			the op will be stored. (Unless OPf_STACKED is on, in
+ *			which case the result go onto the value stack.)
```
tonycoz: OPf_STACKED usually controls whether an argument arrives on the stack, not whether the result is stacked.

This is otherwise confusing, pp functions typically* won't replace the SV at that entry in the pad, they will typically return that SV, often after using it as a scratch value to avoid creating a new SV each time that particular op is called.

* `pp_refassign` will replace the targ entry, I don't recall any others off hand

op.h
richardleach: Ok. I'll remove that mention.

op.h
richardleach: Hmm, don't know why I got the wrong idea about OPf_STACKED. I had meant that the SV in the pad would be (re)used as a scratch, rather than obliterated, but can see that wording isn't clear enough.

How about this?
"Points to a pad entry (the target) for the current op. The entry typically points to a SV which is (re)used to store the result of the op."

tonycoz: I think that's better.

iabyn:
On Mon, Aug 09, 2021 at 10:16:10PM -0700, Tony Cook wrote:
> @tonycoz commented on this pull request.
> 
> 
> 
> > + * 	op_targ         Points to the pad position into which the result of
> + *			the op will be stored. (Unless OPf_STACKED is on, in
> + *			which case the result go onto the value stack.)
> 
> I think that's better.

But it's not a pointer!

How about:

    An index into the current pad, identifying an SV that is typically
    used to store the OP's result (such as a lexical variable, or an
    SVs_PADTMP temporary intermediate value).

-- 
"Strange women lying in ponds distributing swords is no basis for a system
of government. Supreme executive power derives from a mandate from the
masses, not from some farcical aquatic ceremony."
    -- Dennis, "Monty Python and the Holy Grail"

richardleach:
> An index into the current pad, identifying an SV that is typically used to store the OP's result (such as a lexical variable, or an SVs_PADTMP temporary intermediate value).

Thanks, that's fine with me. I'll update this PR.

richardleach:
@tonycoz  @iabyn - good to merge?

PR #18751, opened 2021-04-29
perlgov: proposed edits for election process

rjbs:
( to be discussed on perl-core )

### DIFF
--- a/pod/perlgov.pod
+++ b/pod/perlgov.pod
@@ -379,11 +379,11 @@ Governance".
 
 =head3 Term
 
-A new Steering Council will be chosen by a Term Election within two weeks after
-each stable feature release (that is, change to C<PERL_REVISION> or
-C<PERL_VERSION>) or after two years, whichever comes first. The council members
-will serve until the completion of the next Term Election unless they are
-removed.
+A new Steering Council will be chosen by a Term Election after each stable
+feature release (that is, change to C<PERL_REVISION> or C<PERL_VERSION>) or
+after two years, whichever comes first. The Term Election will be organized
+within two weeks of the triggering event. The council members will serve until
+the completion of the next Term Election unless they are removed.
 
 =head3 Removal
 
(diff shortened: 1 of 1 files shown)


pod/perlgov.pod
```diff
@@ -395,8 +395,8 @@ entire Steering Council is ever vacant, a Term Election will be held instead.
 
 The Steering Council may defer the Special Election for up to twelve weeks.
 Their intent to do so must be publicly stated to the Core Team.  If any active
-Core Team member objects, the Special Election must be organized within one
-week.
+Core Team member objects within one week, the Special Election must be
+organized within one week.
```
hvds: Given that without deferment the vote must be organized within two weeks, I would have expected an objection to deferment to cause it to fall back to two weeks rather than a tighter one-week schedule.

rjbs: This is fine with me and I have no feelings about it either way.

jkeenan:
Is there anything in this proposed new document that is at variance with @xdg's https://github.com/Perl/perl5/pull/18887?

karenetheridge:
LGTM. I see nothing controversial here.

pod/perlgov.pod
```diff
@@ -427,7 +433,9 @@ anonymously.  After voting is complete, candidates are ranked using the
 Condorcet Internet Voting Service's proportional representation mode.  If a tie
 occurs, it may be resolved by mutual agreement among the tied candidates, or
 else the tie will be resolved through random selection by the Vote
-Administrator.
+Administrator.  Unless there are more candidates than open seats, no vote is
```
xdg: As with code, I find "unless" harder to mentally parse than an "if".  Also, the rest of the document uses the term "vacancy", not "open seats" so I suggest that be changed for consistently.  I have a mild preference that we call this out either at the top of item 2 or as a new item 2 so that this check appears between "nomination" and "ranking" in the document in the same way it will happen in the actual process.

xdg (changes requested):

nwc10:
This PR should be closed, because the branch is out of date?

(The first conflict I found was different text for *perlgov: allow deferral of Special Election*)

rjbs:
Yes, thanks.   I must not have pushed to this branch on GitHub before merging to blead!

Issue #18703, opened 2021-04-09
Cwd hard-codes location of pwd. This causes perl to fail to build on Guix.


Frosku:
Module: Cwd

**Description**
On Guix & NixOS, `/bin/pwd` does not exist, the location of most of what would usually be in `/bin/` is `/run/current-system/profile/bin` (guix) or `/run/current-system/sw/bin/` (nix). This causes building perl to fail:

```
Can't figure out your cwd! at /home/frosku/perl5/perlbrew/build/perl-5.32.1/perl-5.32.1/cpan/ExtUtils-MakeMaker/lib/ExtUtils/MakeMaker.pm line 237.
```

**Steps to Reproduce**
1. Be on Guix or NixOS
2. Download perl sources
3. `./Configure -des -Dcc=gcc -Dprefix=$HOME/perl5/`
4. `make`

**Expected behavior**
Building perl succeeds.

**Perl configuration**
N/A

**Workaround**
Add `(extra-special-file "/bin/pwd" (file-append coreutils "/bin/pwd"))` to `services` block in `/etc/config.scm` and reconfigure system. This will populate `/bin/pwd`.

Grinnz:
The referenced error is thrown by Cwd, which tries a number of different things but hardcodes paths so it works under taint mode 🙄 https://github.com/Perl/perl5/blob/12a8085aee9a1a5ac613205b43ea06406736ac6b/dist/PathTools/Cwd.pm#L150-L164

Probably needs functionality added for this OS

tonycoz:
This should be falling back to Internals::getcwd() in 5.32.1, assuming getcwd(2) is found.

Is `d_getcwd=define` in the config.sh generated by Configure?

It is defined here on guix, but I ran into other problems before I get to where you are (presumably from your better understanding of the platform.)

Frosku:
> This should be falling back to Internals::getcwd() in 5.32.1, assuming getcwd(2) is found.
> 
> Is `d_getcwd=define` in the config.sh generated by Configure?
> 
> It is defined here on guix, but I ran into other problems before I get to where you are (presumably from your better understanding of the platform.)

I've since moved back to nixos because of unrelated issues, I can confirm here it's:

```ini
d_getcwd='define'
```

But I get the same build issue.

Frosku:
I've pushed a PR to include the locations of `pwd` on guix & nixos, similar to what's already used.

tonycoz:
[...]

PR #18752, opened 2021-04-30
Update Cwd.pm to work on Guix & NixOS

Frosku:
refs #18703 

This change fixes Cwd on NixOS/Guix.

### DIFF
--- a/dist/PathTools/Cwd.pm
+++ b/dist/PathTools/Cwd.pm
@@ -3,7 +3,7 @@ use strict;
 use Exporter;
 
 
-our $VERSION = '3.80';
+our $VERSION = '3.81';
 my $xs_version = $VERSION;
 $VERSION =~ tr/_//d;
 
--- a/dist/PathTools/Cwd.pm
+++ b/dist/PathTools/Cwd.pm
@@ -156,7 +156,7 @@ if($^O ne 'MSWin32') {
 		     '/usr/bin/pwd',
 		     '/QOpenSys/bin/pwd', # OS/400 PASE.
 		     '/run/current-system/profile/bin/pwd', # GUIX
-		     '/run/current-system/sw/bin/pwd' #NixOS
+		     '/run/current-system/sw/bin/pwd' # NixOS
 		    ) {
 	if( -x $try ) {
 	    $pwd_cmd = $try;
--- a/dist/PathTools/Cwd.pm
+++ b/dist/PathTools/Cwd.pm
@@ -155,6 +155,8 @@ if($^O ne 'MSWin32') {
     foreach my $try ('/bin/pwd',
 		     '/usr/bin/pwd',
 		     '/QOpenSys/bin/pwd', # OS/400 PASE.
+		     '/run/current-system/profile/bin/pwd', # GUIX
+		     '/run/current-system/sw/bin/pwd' #NixOS
 		    ) {
 	if( -x $try ) {
 	    $pwd_cmd = $try;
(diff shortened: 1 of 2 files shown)


grahamc:
The more correct way to handle this (as a NixOS contributor) is to patch Cwd at build time to use an absolute path to pwd. Please don't merge this.

Frosku:
> The more correct way to handle this (as a NixOS contributor) is to patch Cwd at build time to use an absolute path to pwd. Please don't merge this.

As discussed in IRC, this doesn't work when using any of the common perl version managers (perlbrew/plenv) which do not use nix expressions or nix build tools.

grahamc:
Those tools should also be patched.

Frosku:
> Those tools should also be patched.

This would require actual rewriting of the tool so that it rewrites `Cwd.pm` in place every time it pulls perl.

grahamc:
Unfortunately this will be true for lots of software installed by perlbrew etc, as lots of Perl packages expect programs at specific places in the FS. I wonder what other techniques might work?

Frosku:
> Unfortunately this will be true for lots of software installed by perlbrew etc, as lots of Perl packages expect programs at specific places in the FS. I wonder what other techniques might work?

Typically this is for the same reason where you see it: to avoid breaking under [taint mode](https://perldoc.perl.org/perlsec#Taint-mode), which is an ancient security feature. Short of deprecating taint mode or rewriting perl version managers (plenv/perlbrew) to rewrite the module in place before building, it's very difficult to solve.

Regardless, perl should be able to be built in nixos with or without using nix build tools, because current nix build tools do not have the sophistication to replace something like perlbrew (managing multiple versions, and multiple `local::lib` setups). Tim Toady and all that.

jkeenan:
@tonycoz or @atoomic, could either of you review the discussion in this p.r.?

Thank you very much.
Jim Keenan

dist/PathTools/Cwd.pm
```diff
@@ -155,6 +155,8 @@ if($^O ne 'MSWin32') {
     foreach my $try ('/bin/pwd',
 		     '/usr/bin/pwd',
 		     '/QOpenSys/bin/pwd', # OS/400 PASE.
+		     '/run/current-system/profile/bin/pwd', # GUIX
+		     '/run/current-system/sw/bin/pwd' # NixOS
```
grahamc, line 159: Importantly, Nix's concept of closures is very important and highly valued. It is very common to export a closure and run it elsewhere, where `/run/` doesn't exist in any form. This works "sometimes" but intentionally sidesteps the design philosophy of Nix, creating software artifacts that appear to work in development and then fails on deployment due to cheating like this.

tonycoz:
I suspect from the discussion and from using nixos to test #18791 that Configure should be detecting pwd (or PATH) to save in `config.sh` and Cwd should then using that to find pwd by default.

I think the standard way would be to use confstr() to get the system PATH and search that, but POSIX.pm doesn't expose that, and POSIX.pm isn't available at the point this is called anyway.  I'm pretty sure that would have the same problems that this change does, since the result of `getconf PATH` doesn't change in a child environment.

So from the discussion I don't think this change is appropriate.

jkeenan:
> I suspect from the discussion and from using nixos to test #18791 that Configure should be detecting pwd (or PATH) to save in `config.sh` and Cwd should then using that to find pwd by default.
> 
> I think the standard way would be to use confstr() to get the system PATH and search that, but POSIX.pm doesn't expose that, and POSIX.pm isn't available at the point this is called anyway. I'm pretty sure that would have the same problems that this change does, since the result of `getconf PATH` doesn't change in a child environment.
> 
> So from the discussion I don't think this change is appropriate.

No one has disputed @tonycoz's judgment on this question in the past five months, so I'm closing this p.r.

Leont:
> I suspect from the discussion and from using nixos to test #18791 that Configure should be detecting pwd (or PATH) to save in config.sh and Cwd should then using that to find pwd by default.

This sounds sensible to me

PR #18757, opened 2021-04-30
docs: small but important perlexperiment updates

rjbs:
This updates the `perlexperiment` document to:

* add GitHub issue links for more tickets
* add a few pieces of missing data
* unify the presentation of data just a _little_ more

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -417,6 +417,12 @@ bareword filehandles for older versions of perl.
 
 =head2 The 'try' feature.
 
+B<WARNING>: This feature is still experimental and the implementation may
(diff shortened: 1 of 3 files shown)


jkeenan:
At the risk of being pedantic ... do we have unit tests for each of these experiments, as well as for their `no` variants?

Thank you very much.
Jim Keenan

rjbs:
I don't know.  A lack of tests is a good reason not to take something out of experimental status, but I hope we agree that it doesn't mean we should leave the list of experiments stale.

neilb (commented):
LGTM

PR #18762, opened 2021-05-01
perlfunc - mention that delete can only be used on key-value slices since 5.28

Grinnz:
Key-value hash slices were added in 5.20, but delete could not be called on them until 5.28: https://perldoc.perl.org/perl5280delta#delete-on-key/value-hash-slices


xsawyerx:
Is there a reason not to merge this and have it in 5.34.0?

jkeenan:
> Is there a reason not to merge this and have it in 5.34.0?

I guess we just didn't get to it; merging it into blead now.

PR #18765, opened 2021-05-04
Fix perlop.pod about isa operator's precedence

leonerd:
Tiny one-line docs fix - perhaps we can allow this in before 5.34?


leonerd:
Merging despite 5.34 freeze, having checked with rjbs:

https://www.nntp.perl.org/group/perl.perl5.porters/2021/05/msg260028.html

PR #18766, opened 2021-05-05
perldelta - move split change to other perlfunc changes and add issue link

Grinnz:
This was a change to perlfunc specifically so it belongs with the others


xsawyerx:
LGTM. :+1:

PR #18771, opened 2021-05-06
Data::Dumper: rework Unicode-in-qr support

### LINKED ISSUE #18764: Data::Dumper: Malformed UTF-8 character since 5.33.8
The following script indicates a problem since 5.33.8:
```
use strict;
use warnings;
use Data::Dumper;
use Devel::Peek;

my $utf8_qr = "\x{20ac}"; $utf8_qr = qr{$utf8_qr};
my $s = [$utf8_qr, "\344"];
my $d = Data::Dumper->new([$s])->Dump;
Dump $d;
#binmode STDOUT, ':utf8'; print $d;
```
Output is (with 5.33.8, 5.33.9):
```
SV = PV(0x56407351a490) at 0x5640733eb400
  REFCNT = 1
  FLAGS = (POK,pPOK,UTF8)
  PV = 0x56407341c9a0 "$VAR1 = [\n          qr/\342\202\254/u,\n          '\344'\n        ];\n"\0Malformed UTF-8 character: \xe4\x27\x0a (unexpected non-continuation byte 0x27, immediately after start byte 0xe4; need 3 bytes, got 1) in Dump at /tmp/dd.pl line 11.
 [UTF8 "$VAR1 = [\n          qr/\x{20ac}/u,\n          '\x{0}        ];\n"]
  CUR = 55
  LEN = 64
```
With 5.33.7 it looks OK:
```
SV = PV(0x55dd1ca94ec0) at 0x55dd1c966400
  REFCNT = 1
  FLAGS = (POK,pPOK)
  PV = 0x55dd1c975960 "$VAR1 = [\n          qr/\342\202\254/u,\n          '\344'\n        ];\n"\0
  CUR = 55
  LEN = 64
```
Printing the dumped string gives me a � instead of a-umlaut.

I suspect this could be due to https://github.com/Perl/perl5/pull/18619 --- @arc, will you take a look?

arc:
The previous approach was to upgrade the output to the internal UTF-8
encoding when dumping a regex containing supra-Latin-1 characters. That
has the disadvantage that nothing else generates wide characters in the
output, or even knows that the output might be upgraded.

A better approach, and one that's more consistent with the one taken for
string literals, is to use `\x{…}` notation where needed.

Closes #18764

### DIFF
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -29,7 +29,7 @@ our ( $Indent, $Trailingcomma, $Purity, $Pad, $Varname, $Useqq, $Terse, $Freezer
 our ( @ISA, @EXPORT, @EXPORT_OK, $VERSION );
 
 BEGIN {
-    $VERSION = '2.178'; # Don't forget to set version and release
+    $VERSION = '2.179'; # Don't forget to set version and release
                         # date in POD below!
 
     @ISA = qw(Exporter);
@@ -1478,7 +1478,7 @@ modify it under the same terms as Perl itself.
 
 =head1 VERSION
 
-Version 2.178
+Version 2.179
 
 =head1 SEE ALSO
 
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -394,7 +394,9 @@ sub _dump {
         else {
           $pat = "$val";
         }
-        $pat =~ s <(\\.)|/> { $1 || '\\/' }ge;
+        $pat =~ s <(\\.)|(/)|([^[:ascii:]])> {
+            $1 // (defined $2 ? '\\/' : sprintf '\x{%x}', ord $3)
+        }ge;
         $out .= "qr/$pat/$flags";
     }
     elsif ($realtype eq 'SCALAR' || $realtype eq 'REF'
--- a/dist/Data-Dumper/Dumper.xs
+++ b/dist/Data-Dumper/Dumper.xs
@@ -618,8 +618,10 @@ dump_regexp(pTHX_ SV *retval, SV *val)
     SV *sv_flags = NULL;
     const char *rval;
     const char *rend;
-    const char *slash;
+    const char *p;
     CV *re_pattern_cv = get_cv("re::regexp_pattern", 0);
+    int is_utf8;
+    int increment;
 
     if (!re_pattern_cv) {
         sv_pattern = val;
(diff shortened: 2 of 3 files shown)


dist/Data-Dumper/Dumper.pm
```diff
@@ -394,7 +394,9 @@ sub _dump {
         else {
           $pat = "$val";
         }
-        $pat =~ s <(\\.)|/> { $1 || '\\/' }ge;
+        $pat =~ s <(\\.)|(/)|([^[:ascii:]])> {
+            $1 // (defined $2 ? '\\/' : sprintf '\x{%x}', ord $3)
```
Leont, line 398: This seems like an unnecessarily complicated way to do this? The above line does an "escape `/` unless it's escaped", we don't need to do any such tricks.

AFAICT we can just do `$pat =~ s/([^[:ascii:]])/sprintf '\x{%x}', ord $1/ge`

nwc10:
I think that this revised approach is better. Currently in blead:

    $ git grep sv_utf8_upgrade dist/Data-Dumper/
    dist/Data-Dumper/Dumper.xs:        sv_utf8_upgrade(retval);

*ie* the previous approach introduced the fist call to `sv_utf8_upgrade` anywhere, which doesn't **prove** that it's a bad idea, but certainly doesn't mean that there's a "known to work" approach to steal involving "upgrading".

dist/Data-Dumper/Dumper.xs
```diff
@@ -651,25 +653,34 @@ dump_regexp(pTHX_ SV *retval, SV *val)
...
-    rend = rval+rlen;
-    slash = rval;
     sv_catpvs(retval, "qr/");
 
-    for ( ; slash < rend; slash++) {
-        if (*slash == '\\') {
-            ++slash;
-            continue;
+    rval = SvPV(sv_pattern, rlen);
+    rend = rval+rlen;
```
nwc10, line 661: To be clear, **this** patch doesn't change the ordering, but I thought that **in the general case** it was a subtle bad idea to use `SvUTF8` **before** `SvPV` (and friends) because on a tied variable, the latter will call the relevant tie magic, and **that** can change the state of `SVf_UTF8`.

(And IIRC i wrote deliberately evil tests somewhere once that had a tied variable that returned the same code point, but toggle which representation, to try to trigger this sort of problem.)

What I don't know (easily) is whether a tied value can get this far in, but I don't think that there's any harm here in re-ordering this code so that `SvUTF8` is later, and I'll do that in a followup commit.

Of course, I should then audit the rest of the code for this... it seems to use `DO_UTF8` and always does it after `SvPV`.

nwc10:
I'd prefer to go with the plan in #18793, which is based on your analysis and fix in this PR.

nwc10:
This became #18793. Data::Dumper 2.180 is now on CPAN with the fix.

Issue #16003, opened 2017-06-07
testsuite lib/Benchmark unresponsive on older SPARC Solaris 10 server


p5pRT:
Migrated from [rt.perl.org#131531](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=131531) (status was 'open')

Searchable as RT131531$

p5pRT:
### From dclarke@blastwave.org

#### Created by dclarke@blastwave.org



Please ignore the version of perlbug used to submit this bug\.

I performed a normal compile of Perl 5\.26\.0 from the release 
tarball\.  No issues seen during compile\.

Ran the testsuite and the process seemed fine until the Benchmark
test \.\./lib/Benchmark\.t which seemed to become nonresponsive and
sat running for two days before I hit CTRL\-C to stop it\. 

CPU usage was seen to be at 98% for the entire duration\. 

Some details here are the environment variables in place during
the compile and then testsuite run : 

mimas$ env | sort 
AR=/usr/ccs/bin/ar
AS=/usr/ccs/bin/as
AWK=/usr/local/bin/gawk
BUILD=/usr/local/build
CC=/opt/solarisstudio12\.4/bin/cc
CFLAGS=\-errfmt=error \-erroff=%none \-errshort=full \-xstrconst \-xildoff \-m64 \-xmemalign=8s \-xnolibmil \-Xa \-xcode=pic32 \-xregs=no%appl \-xlibmieee \-mc \-g \-xs \-ftrap=%none \-Qy \-xbuiltin=%none \-xdebugformat=dwarf \-xunroll=1 \-D\_TS\_ERRNO \-D\_POSIX\_PTHREAD\_SEMANTICS \-D\_LARGEFILE64\_SOURCE
COLUMNS=132
CONFIG\_SHELL=/usr/local/bin/bash
CPPFLAGS=\-I/usr/local/include \-D\_TS\_ERRNO \-D\_POSIX\_PTHREAD\_SEMANTICS \-D\_LARGEFILE64\_SOURCE
CXX=/opt/solarisstudio12\.4/bin/CC
[...]

PR #18777, opened 2021-05-08
Gh 16003 benchmark

jkeenan:
This p.r. replaces the patch submitted for investigation in what is now https://github.com/Perl/perl5/issues/16003.

### DIFF
--- a/lib/Benchmark.pm
+++ b/lib/Benchmark.pm
@@ -699,7 +699,7 @@ sub runloop {
         $subref  = _doeval($subcode);
     }
     croak "runloop unable to compile '$c': $@\ncode: $subcode\n" if $@;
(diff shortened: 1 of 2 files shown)


jkeenan:
> This p.r. replaces the patch submitted for investigation in what is now #16003.

Per discussion in https://github.com/Perl/perl5/issues/16003, this p.r. will be withdrawn at the end of the month unless we hear from the OP in that issue.

jkeenan:
> > This p.r. replaces the patch submitted for investigation in what is now #16003.
> 
> Per discussion in #16003, this p.r. will be withdrawn at the end of the month unless we hear from the OP in that issue.

We didn't hear from the OP in the time frame requested, so I'm closing this ticket.

Issue #17095, opened 2019-07-15
Core taint tests should be skipped instead of failing when SILENT_NO_TAINT_SUPPORT is set


p5pRT:
Migrated from [rt.perl.org#134287](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=134287) (status was 'open')

Searchable as RT134287$

p5pRT:
### From burak.gursoy@booking.com

#### Created by burak.gursoy@booking.com



SILENT\_NO\_TAINT\_SUPPORT causes all taint tests spread over lots of places in the 
core test suite to fail\. These shuld be skipped when this option is set as with
the current behavior\, one needs to not runs tests at all all ignore build test
failures for perl altogether\, which in turn might maskk real errors\.

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=medium

Site configuration information for perl 5.28.1:

Configured by root at Mon Jul 15 11:28:20 UTC 2019.

Summary of my perl5 (revision 5 version 28 subversion 1) configuration:
   
  Platform:
    osname=linux
    osvers=3.10.0-862.14.4.el7.x86_64
    archname=x86_64-linux
    uname='linux 9bcba996f4e8 3.10.0-862.14.4.el7.x86_64 #1 smp wed sep 26 15:12:11 utc 2018 x86_64 gnulinux '
    config_args='-des -Dcc=gcc -A ccflags=-DSILENT_NO_TAINT_SUPPORT'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='gcc'
    ccflags ='-DSILENT_NO_TAINT_SUPPORT -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2'
    cppflags='-DSILENT_NO_TAINT_SUPPORT -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='6.3.0 20170516'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
[...]

PR #18778, opened 2021-05-08
Skip all if built with -DSILENT_NO_TAINT_SUPPORT

jkeenan:
What formerly was bug ticket RT # 134287 is now GH issue
https://github.com/Perl/perl5/issues/17095.

This p.r. replaces patches attached to that ticket.


jkeenan:
No one spoke up for this change in five months, and I don't myself consider it good enough to pursue further.  Closing.

Issue #13690, opened 2014-03-25
Can't build perl with gcc4.9 due to UB in sv.c


p5pRT:
Migrated from [rt.perl.org#121505](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=121505) (status was 'open')

Searchable as RT121505$

p5pRT:
### From polacek@redhat.com

#### Created by polacek@redhat.com



<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=high

This perlbug was built using Perl 5.14.3 in the Fedora build system.
It is being executed now by Perl 5.14.3 - Fri Jan 11 13:09:43 UTC 2013.

Site configuration information for perl 5.14.3:

Configured by Red Hat, Inc. at Fri Jan 11 13:09:43 UTC 2013.

Summary of my perl5 (revision 5 version 14 subversion 3) configuration:
   
  Platform:
    osname=linux, osvers=2.6.32-279.9.1.el6.x86_64, archname=x86_64-linux-thread-multi
    uname='linux buildvm-25.phx2.fedoraproject.org 2.6.32-279.9.1.el6.x86_64 #1 smp fri aug 31 09:04:24 edt 2012 x86_64 x86_64 x86_64 gnulinux '
[...]

PR #18781, opened 2021-05-10
Always set -fwrapv in the gcc compiler flags on Win32.

### LINKED ISSUE #18739: [Win32] "-fwrapv" disappears from ccflags with gcc version 10.
This had me a bit puzzled, but I guess that <code>"$(GCCVER1)" geq "5"</code> is doing string comparison, and is therefore untrue when $(GCCVER1) is in the range 10 .. 50.
(I look forward to seeing -fwrapv pop up gain when gcc-51 is released ;-)

AFAICT, the absence of -fwrapv makes no difference, and I do wonder if it was ever actually needed on Windows (or anywhere else).
Is there a simple test case for determining this ?
I'm running Windows 7. Might that make a difference ?

I've read through https://github.com/Perl/perl5/issues/13690, but found it difficult to follow, and I haven't yet managed to reproduce a case that needs that switch.

My belief is that -fwrapv is the default behaviour anyway.
If that view is correct, then I would think that the need to specify that switch must represent a bug in -fwrapv (unless, of course, it is being used to override an earlier instance of -ftrapv).

I've also built 5.33.9 (current devel, released today) without -fwrapv, using gcc-7.2.0, and all tests passed.
```
C:\comp-aux\perl-5.33.9\win32>..\perl -I..\lib -V:archname
archname='MSWin32-x64-multi-thread';

C:\comp-aux\perl-5.33.9\win32>..\perl -I..\lib -V:gccversion
gccversion='7.2.0';

C:\comp-aux\perl-5.33.9\win32>..\perl -I..\lib -V:ccflags
ccflags=' -DWIN32 -DWIN64  -DPERL_TEXTMODE_SCRIPTS -DPERL_IMPLICIT_CONTEXT -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STDIO -fno-strict-aliasing -mms-bitfields';

C:\comp-aux\perl-5.33.9\win32>..\perl -I..\lib -le "print $]"
5.033009
```
AFAICT perl will no longer build with gcc-4 on Windows.
With mingw-w64 port of gcc-4.9.2, I get unresolved symbols during linking when the Win32 (core) module is being built.

I guess there's really nothing wrong with specifying -fwrapv if it's not needed.
Perhaps, since perl is no longer buildable with gcc-4.3 or earlier, all we need do is hard code the -fwrapv switch permanently into the GNUmakefile ?
Or do we really want to include that switch only when it's needed ?

Here's the full 'perl -V' output of the 7.2.0 build:
```
C:\comp-aux\perl-5.33.9\win32>..\perl -I..\lib -V
Summary of my perl5 (revision 5 version 33 subversion 9) configuration:

  Platform:
    osname=MSWin32
    osvers=6.1.7601
    archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -DWIN64  -DPERL_TEXTMODE_SCRIPTS -DPERL_IMPLICIT_CONTEXT
-DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STDIO -fno-strict-aliasing -
mms-bitfields'
    optimize='-s -O2'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='7.2.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='long long'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='g++'
    ldflags ='-s -L"C:\_64\blead-5.33.9\lib\CORE" -L"C:\_64\gcc-mingw-720\mingw6
4\lib" -L"C:\_64\gcc-mingw-720\mingw64\x86_64-w64-mingw32\lib" -L"C:\_64\gcc-min
gw-720\mingw64\lib\gcc\x86_64-w64-mingw32\7.2.0"'
    libpth=C:\_64\gcc-mingw-720\mingw64\lib C:\_64\gcc-mingw-720\mingw64\x86_64-
w64-mingw32\lib C:\_64\gcc-mingw-720\mingw64\lib\gcc\x86_64-w64-mingw32\7.2.0 C:
\_64\gcc-mingw-720\mingw64\x86_64-w64-mingw32\lib C:\_64\msys_720\1.0\local\lib
C:\_64\gcc-mingw-720\mingw64\lib\gcc\x86_64-w64-mingw32\7.2.0
    libs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladvapi3
2 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lversio
n -lodbc32 -lodbccp32 -lcomctl32
    perllibs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladv
api32 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lve
rsion -lodbc32 -lodbccp32 -lcomctl32
    libc=
    so=dll
    useshrplib=true
    libperl=libperl533.a
    gnulibc_version=''
  Dynamic Linking:
    dlsrc=dl_win32.xs
    dlext=dll
    d_dlsymun=undef
    ccdlflags=' '
    cccdlflags=' '
    lddlflags='-mdll -s -L"C:\_64\blead-5.33.9\lib\CORE" -L"C:\_64\gcc-mingw-720
\mingw64\lib" -L"C:\_64\gcc-mingw-720\mingw64\x86_64-w64-mingw32\lib" -L"C:\_64\
gcc-mingw-720\mingw64\lib\gcc\x86_64-w64-mingw32\7.2.0"'


  Compile-time options:
    HAS_TIMES
    HAVE_INTERP_INTERN
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_IMPLICIT_CONTEXT
    PERL_IMPLICIT_SYS
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under MSWin32
  Compiled at Apr 21 2021 10:22:57
  @INC:
    ..\lib
    C:/comp-aux/perl-5.33.9/lib
```
Cheers,
Rob

nwc10:
We need this because we are unfortunately relying on the behaviour of signed
integer overflow. This is undefined behaviour in C (which gcc's optimiser
likes to take advantage of) unless we use this flag.

Previously we had used it conditionally based on gcc version, but that code
failed to consider gcc versions greater than 9. We realise that the flag is
actually present from the minimum gcc that we support, so we should simply
add it always. Fixes #18739.

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -583,15 +583,14 @@ MINIBUILDOPT    += -D__USE_MINGW_ANSI_STDIO
 endif
 
 
-# If you are using GCC, 4.3 or later by default we add the -fwrapv option.
-# See https://github.com/Perl/perl5/issues/13690
-#
-GCCWRAPV := $(shell if "$(GCCVER1)"=="4" (if "$(GCCVER2)" geq "3" echo define) else if "$(GCCVER1)" geq "5" (echo define))
+# Previously, if you were using GCC, 4.3 or later we addeed the -fwrapv option.
+# See https://github.com/Perl/perl5/issues/13690 and 18739
+# However, as documented in README.win32 our minimum Win32 gcc version is 3.4.5.
+# As that supports -fwrapv it's both simpler and more correct to unconditionally
+# add it to the flags:
 
-ifeq ($(GCCWRAPV),define)
 BUILDOPT        += -fwrapv
 MINIBUILDOPT    += -fwrapv
-endif
 
 i = .i
 o = .o


nwc10:
**If this builds and passes tests on Win32** (and my wording is good **fail**), please could someone fast-forward blead with this.

Do we actually have anything smoking gcc on win32?

nwc10:
Sigh, I shouldn't do this on a fast machine that doesn't have a spell checker locally.

sisyphus:
It's fine for me.
I've just applied the patch to GNUmakefile manually by simply removing the 3 lines of code that your patch removed.
LGTM and tests fine with a couple of mingw-w64 compilers (namely gcc-9.3.0 and gcc-10.3.0).
I don't see the point in testing any more rigorously than that - but I can if you like.

I haven't tested to see that -fwrapv doesn't get added when MSVC is used, but it's quite apparent from the code that -fwrapv is not added when CC is 'cl'.
In any case, it seems that 'cl' would simply ignore that flag (and tell us so).

Cheers,
Rob

steve-m-hay (commented):
I worry a little about the comment that Zefram added on #13690 that -fwrapv is broken prior to 4.3: https://github.com/Perl/perl5/issues/13690#issuecomment-544054926
I will try to test this with earlier (www.mingw.org) gccs, but I just found that it doesn't currently build anyway (SE_PRIVILEGE_REMOVED, RRF_RT_REG_DWORD and KEY_WOW64_64KEY are undeclared... - presumably happening since commit edfcb93db2).
Also, there is a typo in the comments: "addeed".

nwc10:
> I worry a little about the comment that Zefram added on #13690 that -fwrapv is broken prior to 4.3: [#13690 (comment)](https://github.com/Perl/perl5/issues/13690#issuecomment-544054926)
> I will try to test this with earlier ([www.mingw.org](http://www.mingw.org)) gccs, but I just found that it doesn't currently build anyway (SE_PRIVILEGE_REMOVED, RRF_RT_REG_DWORD and KEY_WOW64_64KEY are undeclared... - presumably happening since commit [edfcb93](https://github.com/Perl/perl5/commit/edfcb93db2c5e42e47f867b5f2b73a3320a6487e)).
> Also, there is a typo in the comments: "addeed".

Pants! Thanks. I guess I should drag that commit down onto a machine *with* a spell checker and see what else I goofed.

[In 9 days we will celebrate the 13th anniversary of the gcc 4.2.4 release](https://gcc.gnu.org/gcc-4.2/) - the question @rjbs asked on the PSC call was whether we are really concerned with supporting a compiler that old. Does 4.2 even compile "hello world" on post NT 4 systems? I'm not really familiar with how compatible Win32 manages to be, but is supporting this about as relevant as trying to support running on Unix on mixed endian hardware?

steve-m-hay:
> [In 9 days we will celebrate the 13th anniversary of the gcc 4.2.4 release](https://gcc.gnu.org/gcc-4.2/) - the question @rjbs asked on the PSC call was whether we are really concerned with supporting a compiler that old. Does 4.2 even compile "hello world" on post NT 4 systems? I'm not really familiar with how compatible Win32 manages to be, but is supporting this about as relevant as trying to support running on Unix on mixed endian hardware?

The build of 5.34.0 RC1 is also broken. I've just posted to p5p to ask the question should we drop support for www.mingw.org compilers, given how flakey they seem to be: We only had 3.4.5-5.3.0 working anyway; all the later ones already had other issues.

(Yes, gcc-3.4.5 does build a perfectly runnable "hello world" C program on Windows 10.)

Tux:
I do not have HP-UX 10.20 anymore. That system died in a poof of orange smoke, but it had gcc-3.4.6 and no way to build anything newer. Which means that anyone still using HP-UX 10.20, the most recent available gcc compiler is 3.4.6 (as I was the only one making gcc available for 10.20)
I do have HP-UX 11.11, but it is powered off currently. It still has GNU gcc-4.6.1 and no way to build anything newer. 4.6.1 *is* newer than 4.2.4, so that should by no means be a blocker.
The HP-UX 11.23 box won't boot due to disk failure. It had 4.6.1
The HP-UX 11.31 runs gcc 4.7.2 and I got no further in any attempt to build a more recent gcc (I tried several times)

The AIX box that p5p has access to runs gcc-4.2.4 and I have exactly NULL tuits to even try to build a more recent gcc on it

nwc10:
This was only relevant for Win32.

For *nix, it seems that Configure handles all this nicely already:

    # gcc 4.9 by default does some optimizations that break perl.
    # see ticket 121505.
    #
    # The -fwrapv disables those optimizations (and probably others,) so
    # for gcc 4.9 (and later, since the optimizations probably won't go
    # away), add -fwrapv unless the user requests -fno-wrapv, which
    # disables -fwrapv, or if the user requests -fsanitize=undefined,
    # which turns the overflows -fwrapv ignores into runtime errors.
    case "$gccversion" in
    4.[3-9].*|4.[1-9][0-9]*|[5-9].*|[1-9][0-9]*)
        case "$ccflags" in
        *-fno-wrapv*|*-fsanitize=undefined*|*-fwrapv*) ;;
        *) ccflags="$ccflags -fwrapv" ;;
        esac
    esac


so I don't think we need to worry about old gcc for anything that builds with Configure - for *nix "it ain't broken, so don't fix it".

sisyphus:
On Mon, May 10, 2021 at 10:15 PM Nicholas Clark ***@***.***>
wrote

>
> In 9 days we will celebrate the 13th anniversary of the gcc 4.2.4 release
> <https://gcc.gnu.org/gcc-4.2/> - the question @rjbs
> <https://github.com/rjbs> asked on the PSC call was whether we are really
> concerned with supporting a compiler that old.
>

Personally, I'm not.


> Does 4.2 even compile "hello world" on post NT 4 systems?
>
It probably does. The oldest gcc-4.x.x version that I have on Windows 7 is
4.7.0.
But gcc-3.4.5 still compiles basic C programs on that Windows box - though
I very rarely use that compiler.
I did try the following with it:
____________________
C:\_32\C>gcc --version
gcc (GCC) 3.4.5 (mingw special)
Copyright (C) 2004 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

C:\_32\C>type try.c
/* try.c */
#include <stdio.h>

int main(void) {
 unsigned long bar = 1073741823;

 printf("%u\n", bar << 10);
 return 0;
}

C:\_32\C>gcc -o try.exe try.c -fwrapv -fno-strict-aliasing

C:\_32\C>try
4294966272
___________________

It's the same results as I get with gcc-11.1.0 on the same Windows 7 box.

I'm not really familiar with how compatible Win32 manages to be, but is
> supporting this about as relevant as trying to support running on Unix on
> mixed endian hardware?
>
The challenges might be different, but the level of relevance sounds about
the same to me. (However, I don't really feel qualified to be making such
assessments.)

Cheers,
Rob

xenu:
>I worry a little about the comment that Zefram added on #13690 that -fwrapv is broken prior to 4.3:

I tried to find anything about this in google, but to no avail. Anyway, I'm not aware of any valid reasons to use such an old gcc version on Windows. We really should revise our supported compiler/OS version list.

win32/GNUmakefile
```diff
@@ -583,15 +583,14 @@ MINIBUILDOPT    += -D__USE_MINGW_ANSI_STDIO
...
+# Previously, if you were using GCC, 4.3 or later we addeed the -fwrapv option.
+# See https://github.com/Perl/perl5/issues/13690 and 18739
+# However, as documented in README.win32 our minimum Win32 gcc version is 3.4.5.
+# As that supports -fwrapv it's both simpler and more correct to unconditionally
+# add it to the flags:
 
-ifeq ($(GCCWRAPV),define)
 BUILDOPT        += -fwrapv
 MINIBUILDOPT    += -fwrapv
-endif
```
tonycoz, line 594: I think the switch should remain, so you can do `gmake GCCWRAPV=undef` to test building without `-fwrapv`.

The ideal is that we eventually remove the `-fwrapv` entirely, being able to test without it easily is desirable.

nwc10: Good points, and I'd not thought of this. Agree, this is better than what I did.

Given that I don't have access to a Win32 system to test this, and I think that you do, and I think that testing the non-default case requires doing something interactively

1. Could you implement this and test that it works
2. Push it to a branch to replace mine, so that (at least one of) @steve-m-hay @xenu or @sisyphus can check it (better than me)
3. If someone else says yes and no-one says no, it goes to blead.

I feel like I'm turning into a project manager and it's not an enjoyable thought. :-/

sisyphus:
On Tue, May 11, 2021 at 4:16 PM Nicholas Clark ***@***.***>
wrote:

> Good points, and I'd not thought of this. Agree, this is better than what
> I did.
>

Yes, I think that providing a ready option to disable -fwrapv is an
improvement - especially given our uncertainty regarding supposed bugs re
this switch on older gcc versions.
Then we can just leave it to the builder to override the default behaviour
for whatever reason, whenever convenient.

I can't imagine how this patch of Tony's could possibly do anything other
than what is intended, but I've tested with the following compilers:
mingw.org gcc-9.2.0
32 bit mingw-w64 gcc-10.3.0 and gcc-8.3.0
64 bit mingw-w64 gcc-11.1.0 and gcc-8.3.0.

By "tested", I mean that I verified that -fwrapv was being set. This only
takes a few seconds to ascertain, then I'd hit Ctrl-C, run 'make
distclean', and move on to the next compiler.
I also did the same quick test using those same compilers, having specified
"GCCWRAPV=undef'"-  just to be sure that it did remove the -fwrapv switch.

LGTM.

Cheers,
Rob

win32/GNUmakefile
tonycoz: See #18782 (which @sisyphus has found it looks like)

nwc10:
Oooh, this is actually **my** PR. So I don't feel bad in rejecting it and saying #18782 is better.

PR #18785, opened 2021-05-11
Make scalar assignment to undef a compile time error (Closes #18779)

### LINKED ISSUE #18779: Assign to undef detected during run but not compile
The following error message is issued during execution - might it be possible to detect it at compile time so that it can be fixed as soon as possible?  
```
undef = my $a; # Modification of a read-only value attempted
```

richardleach:
Scalar assignment to undef is currently a runtime error:

````
# perl -c -e 'undef = my $foo;'
-e syntax OK
# perl -e 'undef = my $foo;'
Modification of a read-only value attempted at -e line 1.
````
It would be helpful for users to learn that this is invalid
syntax as soon as possible, i.e. at compile time. This
was requested in [[GH#18779](https://github.com/Perl/perl5/issues/18779)].

This PR makes it a compile time error:
````
# ./miniperl -c -e 'undef = my $foo;'
Can't modify undef operator in scalar assignment at -e line 1, near "$foo;"
-e had compilation errors.
````
It does not break the existing list assignment behaviour (even when the
assignment is entirely useless):
````
# ./miniperl -e '(undef) = my $foo;'
# 
````
A grep of CPAN suggests that no (or minimal) breakage is likely.

### DIFF
--- a/op.c
+++ b/op.c
@@ -4291,6 +4291,8 @@ Perl_op_lvalue_flags(pTHX_ OP *o, I32 type, U32 flags)
 
     switch (o->op_type) {
     case OP_UNDEF:
+        if (type == OP_SASSIGN)
+            goto nomod;
 	PL_modcount++;
 	goto do_next;
 
--- a/t/run/fresh_perl.t
+++ b/t/run/fresh_perl.t
@@ -194,8 +194,8 @@ BEGIN failed--compilation aborted at - line 1.
 ########
 BEGIN { undef = 0 }
 EXPECT
-Modification of a read-only value attempted at - line 1.
-BEGIN failed--compilation aborted at - line 1.
+Can't modify undef operator in scalar assignment at - line 1, near "0 }"
+BEGIN not safe after errors--compilation aborted at - line 1.
 ########
 {
     package foo;

### CODE REVIEW

**Comment by richardleach** [comment]
I'm not sure how the perldelta entry for the "Changes to Existing Diagnostics"
section should read. Perhaps something like this:
```
=item *

L<Can't modify %s in %s|perldiag/"Can't modify %s in %s"> (assign to undef)

Attempting to perform a scalar assignment to undef was previously only a
runtime error, triggering the diagnostic
L<Modification of a read-only value attempted|perldiag/"Modification of a read-only value attempted">.
This now triggers a compile time error like
"Can't modify undef operator in scalar assignment".
```

hvds:
@tonycoz could you take a look? It seems straightforward enough to me, but I don't know this area.

@rjbs @nwc10 This should probably get sign-off too, since it's a language change.

tonycoz:
> @tonycoz could you take a look? It seems straightforward enough to me, but I don't know this area.

It looks reasonable to me.

> @rjbs @nwc10 This should probably get sign-off too, since it's a language change.

I'm pretty sure a change at this level doesn't need PSC approval, but I'll admit it's not well defined at this point.

nwc10:
> I'm pretty sure a change at this level doesn't need PSC approval, but I'll admit it's not well defined at this point.

I don't think that it's well defined either at the next level up - I don't know if **one** person on PSC can say "good", or if all 3 have to check. It sort of works if all 3 can guess what the others think.

So, I figure I'm going to guess on both parts, and also attempt to set some precedent. PSC had talked about changes, and when you can detect them:

1. Static tooling
2. Perl interpreter at compile time
3. Perl interpreter at runtime
4. Silent semantic changes that you can't catch

and that the earlier on that list, the better. So, in the spirit of that list:

- Moving the reporting of a fatal error from runtime to compile time is an "optimisation".
- No correct code should be relying on when an error is reported (eg, it should not be relying on an `eval` failing due to a syntax error, or a module compiling with **implicit** fatal errors in code that isn't called.)
- It is OK to make unambiguously similar changes (but if it's code freeze, wait until the freeze is over).

I have not had a chance to look at the actual code change here - there's too much other stuff to do first.

hvds:
> > I'm pretty sure a change at this level doesn't need PSC approval, but I'll admit it's not well defined at this point.
> 
> I don't think that it's well defined either at the next level up - I don't know if **one** person on PSC can say "good", or if all 3 have to check. It sort of works if all 3 can guess what the others think.
> 
> So, I figure I'm going to guess on both parts, and also attempt to set some precedent. PSC had talked about changes, and when you can detect them [...] and that the earlier on that list, the better. So, in the spirit of that list:
> 
>     * Moving the reporting of a fatal error from runtime to compile time is an "optimisation".
> 
>     * No correct code should be relying on when an error is reported (eg, it should not be relying on an `eval` failing due to a syntax error, or a module compiling with **implicit** fatal errors in code that isn't called.)
> 
>     * It is OK to make unambiguously similar changes (but if it's code freeze, wait until the freeze is over).
> 
> 
> I have not had a chance to look at the actual code change here - there's too much other stuff to do first.

Thanks @nwc10; on that basis, and given a) that @tonycoz says it looks reasonable and b) we are nice and early in the 5.36 dev cycle, I think it's a good idea to merge it. I'll plan to do that tomorrow absent objections.

hvds:
Now merged as 2583ef601d.

PR #18792, opened 2021-05-13
style: Detabify the C code... again.

schwern:
Leftovers from 9824c081922f8e3697322536c3da1702e35e45ab and 1604cfb0273418ed479719f39def5ee559bffda2 (see #18446). These don't
seem to be regressions, but never happened. Perhaps they were lost
in the merge.

regen_perly.pl was, again, left alone as it requires rebuilding perly.*
which is beyond a simple indentation change.


feature.h
```diff
@@ -53,132 +53,132 @@
 
 #define FEATURE_FC_IS_ENABLED \
     ( \
-	(CURRENT_FEATURE_BUNDLE >= FEATURE_BUNDLE_515 && \
```
hvds, line 56:
```
   !!!!!!!   DO NOT EDIT THIS FILE   !!!!!!!
   This file is built by regen/feature.pl.
   Any changes made here will be lost!
```

Grinnz: The PR also updates that file.

hvds: Oh, I see that these probably get fixed by the heredoc fixes in regen/feature; that's probably worth calling out as a separate commit, since it is effectively a code change.

Grinnz:
Since we're frozen for 5.34 you should probably expect to have to rebase this in a couple weeks.

hvds:
@schwern blead is unfrozen now that 5.34 is released, would you consider rebasing and pulling the regen/feature.pl change and its side-effects into a separate commit?

khwilliamson:
I needed this applied to one of the files to proceed, so I rebased most of it myself, and it is applied as 1f4fbd3b4b26604673abca2a5f911744e826b1f3.

I'm unsure if it is a good idea to change the regen files to not output tabs.  I found that when I mistakenly went to edit such a generated file, it became immediately obvious that there were tabs in there, and so I knew immediately it was futile to edit such a file.

hvds:
> I'm unsure if it is a good idea to change the regen files to not output tabs. I found that when I mistakenly went to edit such a generated file, it became immediately obvious that there were tabs in there, and so I knew immediately it was futile to edit such a file.

I think we definitely should change them. I think vanishingly few people would notice a tab while editing a file and think "oh, this must be autogenerated".

khwilliamson:
resolved by https://github.com/Perl/perl5/commit/93579e61a7441cd377909544b7b712c860174dbc

Issue #18614, opened 2021-03-02
Data::Dumper silently fails, exiting the script


j-mckim:
When dumping a structure that contains certain Unicode characters, Data::Dumper silently exits the script. The remainder of the structure is not dumped. The rest of the script is not executed.

More seriously, eval{} doesn't intercept the exit.

There's an interaction between Data::Dumper and binmode(). If the latter isn't specified, Dumper works. If it's present, Dumper fails as described.

Versions are: perl 5 version 30, and Data::Dumper version 2.174.

Attached is a short script that triggers the bug.

[bug-data-dumper.pl.txt](https://github.com/Perl/perl5/files/6069516/bug-data-dumper.pl.txt)

jkeenan:
Confirmed; present from at least perl-5.8.4.

hvds:
It is not reproducing for me here.

Do you get the same effect if you replace the literal Unicode character with the corresponding `\x{41f}` escape?

What is the exit code of the script when it fails? (On a Unix system you can check that with `echo $?` directly after the script completes; 0 means it (thinks it) completed ok; 255 usually means a Perl-level crash; any other value usually means a C-level crash.)

Can you catch the failure with a `__DIE__` handler, to get a Perl-level stacktrace? Insert at the start:
```
  use Carp;
  $SIG{__DIE__} = sub { Carp::confess(@_) };
```

Can you run it in a C-level debugger? If it dies with a signal such as SEGV, the debugger should stop at the point the signal is raised;  if so you may be able to get a C-level stacktrace:
```
% gdb -q --args /opt/v5.30.0-d/bin/perl ~/bug-data-dumper.pl.txt
Reading symbols from /opt/v5.30.0-d/bin/perl...done.
(gdb) run
Starting program: /opt/v5.30.0-d/bin/perl /home/hv/bug-data-dumper.pl.txt
$VAR1 = [
          'some stuff',
          " \x{41f}  non-regex with problem char dumps correctly",
          qr/ è® regex with other unicode char dumps correctly /u,
          qr/ Ð regex with problem char; this part is not dumped /u,
          'yet more stuff not dumped'
        ];
Hello! I didn't die!
[Inferior 1 (process 6003) exited normally]
(gdb) where
No stack.
(gdb) exit
% 
```

(This is most useful using a perl built with debugging, but even without that it would give us a clue.)

arc:
I can't reproduce the silent exit:

```
$ perl bug-data-dumper.pl.txt
[...]

PR #18793, opened 2021-05-13
Data::Dumper handles Unicode regex corner cases (GH #18614, GH #18764)

nwc10:
My suggested plan is

1. smoke this
2. merge **adb79bc** into blead before RC2 (so version **2.179**). This reverts the behaviour to how it was on v5.32.0
3. push a dev release of this branch to CPAN to see what the testers think
4. if happy, ship **2.180** to CPAN
5. v5.34.0 ships

The fix for all the bugs feels a bit too risky to put into blead at this time, but we can iterate CPAN release for Data::Dumper much faster than RCs (or v5.34.1). The plan above ensures that

1. There is never a behaviour regression for the version downloaded from CPAN
2. There is a fix immediately available to adopters of v5.34.0, for anyone that needs it

I don't think that we need to change the pure-perl output for `qr//` with Unicode, as (I believe) it wasn't buggy, hence i left it unchanged and adapted the tests.

### DIFF
--- a/dist/Data-Dumper/Changes
+++ b/dist/Data-Dumper/Changes
@@ -6,6 +6,10 @@ Changes - public release history for Data::Dumper
 
 =over 8
 
+=item 2.179_50 (May 14 2021)
+
+Data::Dumper handles Unicode regex corner cases (GH #18614, GH #18764)
+
 =item 2.179 (May 13 2021)
 
 Revert the changes of 2.177 for the v5.34.0 release to avoid a regression.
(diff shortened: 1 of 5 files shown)


nwc10:
XS code for blead unchanged. `$VERSION` now set as 2.179_50

Tests and ppport fun fixed for older perl versions.

rjbs:
@nwc10 I've applied the bottom two commits to blead as (I think!) we discussed.

PR #18802, opened 2021-05-19
perlcommunity - freenode chat is moving to libera

Grinnz:
Context: https://www.kline.sh/


Grinnz:
Probably needless to say, this does not need to go in for 5.34 - the channel is still being set up

Issue #18801, opened 2021-05-19
[doc] `evalbytes`/`unicode_eval` documentation is unclear/confusing


FGasper:
```
  The 'unicode_eval' and 'evalbytes' features
    Together, these two features are intended to replace the legacy string
    "eval" function, which behaves problematically in some instances. They
    are available starting with Perl 5.16, and are enabled by default by a
    "use 5.16" or higher declaration.

    "unicode_eval" changes the behavior of plain string "eval" to work more
    consistently, especially in the Unicode world. Certain (mis)behaviors
    couldn't be changed without breaking some things that had come to rely
    on them, so the feature can be enabled and disabled. Details are at
    "Under the "unicode_eval" feature" in perlfunc.

    "evalbytes" is like string "eval", but operating on a byte stream that
    is not UTF-8 encoded. Details are at "evalbytes EXPR" in perlfunc.
    Without a "use feature 'evalbytes'" nor a "use v5.16" (or higher)
    declaration in the current scope, you can still access it by instead
    writing "CORE::evalbytes".
```
I’m still trying to figure out what these features do, but along the way I’d like to report issues I have grokking the in-place documentation.

> "evalbytes" is like string "eval", but operating on a byte stream that is not UTF-8 encoded.

This doesn’t make a lot of sense … if we’re “eval’ing bytes”, then since UTF-8-encoded text is bytes by definition, why would I not use `evalbytes` for that?

-----
`eval`’s documentation says:
```
                Under the "unicode_eval" feature
                    If this feature is enabled (which is the default under a
                    "use 5.16" or higher declaration), EXPR is considered to
                    be in the same encoding as the surrounding program. Thus
                    if "use utf8" is in effect, the string will be treated
                    as being UTF-8 encoded. Otherwise, the string is
                    considered to be a sequence of independent bytes. Bytes
                    that correspond to ASCII-range code points will have
                    their normal meanings for operators in the string. The
                    treatment of the other bytes depends on if the
                    "'unicode_strings"" feature is in effect.

                    In a plain "eval" without an EXPR argument, being in
                    "use utf8" or not is irrelevant; the UTF-8ness of $_
                    itself determines the behavior.
```
> the string will be treated as being UTF-8 encoded. Otherwise, the string is considered to be a sequence of independent bytes.

What does this mean? How does Perl treat UTF-8-encoded strings any differently from other byte strings?

[...]

PR #18803, opened 2021-05-20
Clarify descriptions of unicode_eval and evalbytes.

FGasper:
Issue #18801


Grinnz:
These are probably reasonable simplifications. I need to have a closer look at the behavior.

FGasper:
For context: https://www.nntp.perl.org/group/perl.perl5.porters/2011/09/msg177373.html

PR #18807, opened 2021-05-21
Perlsphere has been replaced by Planet Perl. Perl Weekly is now on https

davorg:


pod/perlcommunity.pod
```diff
@@ -70,11 +70,11 @@ this site provides quality articles mostly about technical details of Perl.
 Many members of the community have a Perl-related blog on this site. If
 you'd like to join them, you can sign up for free.
 
-=item L<http://perlsphere.net/>
+=item L<httsp://perl.theplanetarium.org/>
```
jkeenan: `s/httsp/https/` -- Yes?

davorg: Oops. Yes. Fixed now.

Issue #18742, opened 2021-04-22
[doc] perlop: Add difference between <=> and cmp super clear paragraph


jidanni:
perlop goes in great detail about <=> and cmp,
```
       Binary "<=>" returns -1, 0, or 1 depending on whether the left argument
       is numerically less than, equal to, or greater than the right argument.
       If your platform supports "NaN"'s (not-a-numbers) as numeric values,
       using them with "<=>" returns undef.  "NaN" is not "<", "==", ">", "<="
       or ">=" anything (even "NaN"), so those 5 return false.  "NaN != NaN"
       returns true, as does "NaN !=" anything else.  If your platform doesn't
       support "NaN"'s then "NaN" is just a string with numeric value 0.

           $ perl -le '$x = "NaN"; print "No NaN support here" if $x == $x'
           $ perl -le '$x = "NaN"; print "NaN support here" if $x != $x'

       (Note that the bigint, bigrat, and bignum pragmas all support "NaN".)

       Binary "cmp" returns -1, 0, or 1 depending on whether the left argument
       is stringwise less than, equal to, or greater than the right argument.

```
but never drives home the point about their difference,

```
$ perl -wle 'print for 10 <=> 2, 10 cmp 2;'
1
-1
```
Please drive home that point.
Yes, even if 99% of readers were supposed to know that back in first
grade, and even if it is on many other man pages.

Perhaps say
```
Here we can see the difference between <=> and cmp,

print 10 <=> 2 #prints 1
print 10 cmp 2 #prints -1

likewise between gt and >, lt and <, etc.
```
and say it right after the paragraph I quoted. Thanks!

khwilliamson:
How about a pull request?

PR #18808, opened 2021-05-22
Add difference between <=> and cmp example

jidanni:
For https://github.com/Perl/perl5/issues/18742


khwilliamson:
I don't see the quote portion you referred to in this commit

PR #18809, opened 2021-05-22
gh18770: stop scanning for substrs after *COMMIT

hvds:
*ACCEPT already avoids this (because it is "ENDLIKE"), but gets a related fix to stop scanning for start class.


khwilliamson:
LGTM, but I am relying on your superior knowledge of the optimizer

demerphq:
LGTM also.

hvds:
Thanks both, I'll merge after I get a test run from rebase.

hvds:
> Thanks both, I'll merge after I get a test run from rebase.

On second thoughts I'll hang fire a bit, since a largish regcomp.c change has just landed.

hvds:
Now merged as 499aa13271.

PR #18813, opened 2021-05-23
Perl_newSVnv: inline relevant subset of sv_setnv

richardleach:
This PR takes its cue from `Perl_newSViv`, which already inlines the relevant subset of `sv_setiv`. This cuts out any function call overhead, a switch statement leading to a `sv_upgrade(sv, SVt_NV)` call, and a touch more bit-twiddling than is necessary.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -9605,10 +9605,12 @@ The reference count for the SV is set to 1.
 SV *
 Perl_newSVnv(pTHX_ const NV n)
 {
-    SV *sv;
+    SV *sv = newSV_type(SVt_NV);
+    (void)SvNOK_on(sv);
 
-    new_SV(sv);
-    sv_setnv(sv,n);
+    SvNV_set(sv, n);
+    SvTAINT(sv);
+
     return sv;
 }
 
--- a/sv.c
+++ b/sv.c
@@ -9626,20 +9626,7 @@ SV is set to 1.
 SV *
 Perl_newSViv(pTHX_ const IV i)
 {
-    SV *sv;
-
-    new_SV(sv);
-
-    /* Inlining ONLY the small relevant subset of sv_setiv here
-     * for performance. Makes a significant difference. */
-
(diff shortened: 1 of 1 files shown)


khwilliamson:
LGTM

If no one ojects by June 15, I'll merge this

sv.c
```diff
@@ -9608,7 +9608,26 @@ Perl_newSVnv(pTHX_ const NV n)
...
 
     new_SV(sv);
-    sv_setnv(sv,n);
+    /* Inlining ONLY the small relevant subset of sv_setnv here
+     * for performance. */
+
+    /* We're starting from SVt_FIRST, so provided that's
+     * actual 0, we don't have to unset any SV type flags
+     * to promote to SVt_IV. */
+    STATIC_ASSERT_STMT(SVt_FIRST == 0);
```
hvds: All this is fine, but I wonder if it is necessary: if we set the type below instead with:
```
  SvFLAGS(sv) = SVt_NV;  /* and optionally ... | SVf_NOK | SVp_NOK */
```
.. then I don't think anything remains reliant on SVt_FIRST does it?

+1 from me to merge this in any case.

richardleach: That comment - which contains a copy&paste error that I'll fix in a sec - and `STATIC_ASSERT_STMT` are only there because they're in `Perl_newSViv`. I figured it would be nice to have visual consistency, but don't think it's necessary to have it. I'm not fussed whether it stays or goes.

richardleach: Now that `newSV_type` is inlined, it's cleaner to just use that, so I've updated the PR to do so.

xenu: I think it would make sense to change `Perl_newSViv` to use `newSV_type` too. It would make it a bit simpler and I expect no change in the performance.

richardleach: Agreed. I'll roll that into this PR shortly.

PR #18815, opened 2021-05-25
perlhacktips: Add callgrind profiling documentation

wolfsage:
One of the benefits of callgrind is you can profile perls that
haven't been built with debugging symbols, which can be real useful
on live-ish datasets


hvds:
+1 Looks like useful information to me; I haven't verified that the information is accurate.

PR #18818, opened 2021-05-25
Khw dumper

khwilliamson:


nwc10:
Looks good. I've rolled this up as part of a Data-Dumper-2.181_50.tar.gz dev release to CPAN. We'll see what the cpan testers make of it over the next couple of days.

nwc10:

PR #18819, opened 2021-05-26
remove MSVC100/vcexpress2010 from CI, at least for now

tonycoz:
Due to failing installation of VC express 2010, that version isn't
being tested anyway, and is just producing errors.

If we work out a solution for the missing download from Microsoft
we can reinstate it, but at this point it's just noise
reducing the utility CI gives us.

This does not make VC2010 unsupported, it's purely a CI change.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -283,50 +283,6 @@ jobs:
 
   #===============================================
 
-  windows-msvc100:
-    name: "Windows msvc100"
-    runs-on: windows-latest
-    timeout-minutes: 120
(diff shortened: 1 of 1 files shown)


jkeenan:
> Due to failing installation of VC express 2010, that version isn't
> being tested anyway, and is just producing errors.
> 
> If we work out a solution for the missing download from Microsoft
> we can reinstate it, but at this point it's just noise
> reducing the utility CI gives us.
> 
> This does not make VC2010 unsupported, it's purely a CI change.

Big +1 from me.

toddr (approved):
LGTM

PR #18821, opened 2021-05-26
Update the glob() documentation so it's more novice friendly

scottchiefbaker:
We throw some heavy concepts (internal C functions?, and filenames with whitespace) at the user in the first two paragraphs. This moves some of that documentation to a separate section and gives simple examples for novices.

Also this normalizes all the function calls to `glob(...)` as per discussion on IRC.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -3370,12 +3370,32 @@ X<glob> X<wildcard> X<filename, expansion> X<expand>
 =for Pod::Functions expand filenames using wildcards
 
 In list context, returns a (possibly empty) list of filename expansions on
-the value of EXPR such as the standard Unix shell F</bin/csh> would do.  In
+the value of EXPR such as the Unix shell Bash would do. In
 scalar context, glob iterates through such filename expansions, returning
-undef when the list is exhausted.  This is the internal function
-implementing the C<< <*.c> >> operator, but you can use it directly.  If
-EXPR is omitted, L<C<$_>|perlvar/$_> is used.  The C<< <*.c> >> operator
-is discussed in more detail in L<perlop/"I/O Operators">.
+L<C<undef>|/undef EXPR> when the list is exhausted. If EXPR is omitted,
+L<C<$_>|perlvar/$_> is used.
+
+    # List context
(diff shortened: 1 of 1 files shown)


pod/perlfunc.pod
```diff
@@ -3388,20 +3401,20 @@ For example, to glob filenames that have an C<e> followed by a space
 followed by an C<f>, use one of:
 
     my @spacies = <"*e f*">;
-    my @spacies = glob '"*e f*"';
-    my @spacies = glob q("*e f*");
+    my @spacies = glob('"*e f*"');
```
Grinnz, line 3411: I don't think these need to be changed.

pod/perlfunc.pod
```diff
@@ -3372,10 +3372,23 @@ X<glob> X<wildcard> X<filename, expansion> X<expand>
...
+    my @perl_files = glob("*.pl *.pm");
+
+    # Scalar context
+    while (my $file = glob("*.mp3")) {
+        # Do stuff
+    }
+
+Glob also supports an altnernate syntax using C<< < >> C<< > >> as
+delimiters. While this syntax is supported, it is recommended that you
+use C<glob()> instead as it is more readable and searchable.
```
Grinnz: Functions should not be followed by empty parentheses in sentences according to the new style guide: https://perldoc.perl.org/perldocstyle#Function-names

pod/perlfunc.pod
```diff
@@ -3415,6 +3428,12 @@ is used as a C<while>/C<for> condition, then the condition actually
 tests for definedness of the expression's value, not for its regular
 truth value.
 
+Internal implemenation details:
+
+In scalar context this is the internal function implementing the
```
Grinnz: Not just in scalar context.

pod/perlfunc.pod
scottchiefbaker: This one I disagree on... I think for consistency all calls should be the same. These are a little harder to read maybe, but they're not different than any of the other glob calls where we used parens.

scottchiefbaker:
Just pinging people. Anything else I can do to land this PR?

PR #18827, opened 2021-05-27
Reveal deeper slices need @{}

jidanni:
Without this additional example, users are sure to assume one must use`@days{'p'}{'x','y'}`!

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -57,9 +57,10 @@ which works much as the word "these" or "those" does in English,
 in that it indicates multiple values are expected.
 X<array>
 
-    @days		# ($days[0], $days[1],... $days[n])
-    @days[3,4,5]	# same as ($days[3],$days[4],$days[5])
-    @days{'a','c'}	# same as ($days{'a'},$days{'c'})
+    @days		# ($days[0], $days[1], ... $days[n])
+    @days[3,4,5]	# same as ($days[3], $days[4], $days[5])
+    @days{'a','c'}	# same as ($days{'a'}, $days{'c'})
+    @{$days{'p'}}{'x','y'} # same as ($days{'p'}{'x'}, $days{'p'}{'y'})
 
 Entire hashes are denoted by '%':
 X<hash>


pod/perldata.pod
```diff
@@ -57,9 +57,10 @@ which works much as the word "these" or "those" does in English,
 in that it indicates multiple values are expected.
 X<array>
 
-    @days		# ($days[0], $days[1],... $days[n])
-    @days[3,4,5]	# same as ($days[3],$days[4],$days[5])
-    @days{'a','c'}	# same as ($days{'a'},$days{'c'})
+    @days		# ($days[0], $days[1], ... $days[n])
```
Grinnz, line 60: Please align these comments with the one below.

jidanni: > Please align these comments with the one below.

I would love to.
But it would be best if you did it for me.
Else, it would tangle up GitHub, in my experience.

smpeters: All that's really needed is a new commit aligning them. If it's messed up in Github, closing and re-opening with a fix would be just fine.

jidanni: To tell you the awful truth, I don't really know what you guys are talking about.
I see a weird diff above. Three red lines and one green.
So maybe those are the comments you want me to align. But maybe I will align them wrong.
Anyway, therefore, I, hereby, authorize you fellows, to align the comments as you please!
(Actually I was going to even throw in another line, showing folks even one more level of depth...)

jidanni: And in fact, all deeper levels require the @{}, but, it seems, one cannot use @{} through and through! As the shallower levels, like the one in the shown before my edit, cannot be made to use it, I think.

smpeters: Screen Shot 2021-05-28 at 11 56 08 AM

As shown here, the start of the end of line comments are not aligned where they all were previously.  Please re-align them.

jidanni:
OK, guessing that pushing "Resolve conversation" would take me to a page where I could put a few spaces into those comments that I was supposed to align.... but it didn't. So I am getting into deeper and deeper water and wish someone would take over.

jidanni (commented):
What a disaster made by me.

tonycoz:
I think this belongs under Slices rather than here, which is just a basic demonstration that `@` means you get a list.

Your new example is probably better done with postfix dereference:

  $days{'p'}->@['x', 'y']

as documented in perlref.

jidanni:
@tonycoz :
I think it belongs in both places, else half the users are sure to miss it.

I also think both formats should be mentioned, yours and mine.

So that makes a total of adding two examples in two places.

khwilliamson:
@jidanni what are your plans for this request

jidanni:
My plans are to leave it up to you guys with more professional judgments than me.

demerphq:
Since Dan was struggling I took the liberty of replacing this PR with:

https://github.com/Perl/perl5/pull/19528

Btw, @tonycoz it is:

```
$days{'p'}->@{'x', 'y'}
```

not 

```
$days{'p'}->@['x', 'y']
```

But I followed your advice and documented the postfix notation.

In order to get the comments to line up and not exceed our line width expectations I had to reduce the leading indent.

Closing this ticket.

PR #18832, opened 2021-05-28
Report Data::Dumper problems to Perl 5 Github issues queue

jkeenan:
In the commit message for 29c66c43dc9b80118284c89a37f090d4f2fada58,
Nicholas Clark wrote:

    Really we should decide *which* bugtracker is canonical for Data::Dumper
    (and will be checked and acted upon) and then record that metadata in this
    Makefile.PL so that [metacpan.org, etc., can] link to it.

Let's provide that information based on the examples of other
blead-upstream distros under `dist`, `dist/Module-CoreList/Makefile.PL`
and `dist/PathTools/Makefile.PL`.

### DIFF
--- a/dist/Data-Dumper/Makefile.PL
+++ b/dist/Data-Dumper/Makefile.PL
@@ -1,10 +1,23 @@
 use strict;
 use warnings;
 use ExtUtils::MakeMaker;
+
+my @extra;
+push @extra, 'LICENSE' => 'perl_5'
+    unless $ExtUtils::MakeMaker::VERSION < 6.31;
+push @extra, 'META_MERGE' => {
+        resources => {
+            repository  => 'git://perl5.git.perl.org/perl.git',
+            bugtracker => 'https://github.com/Perl/perl5/issues',
+            homepage    => "http://dev.perl.org/",
+        },
+    } unless $ExtUtils::MakeMaker::VERSION < 6.46;
+
 WriteMakefile(
     NAME          => 'Data::Dumper',
     VERSION_FROM  => 'Dumper.pm',
     ABSTRACT_FROM => 'Dumper.pm',
     $] <= 5.011000 ? ( INSTALLDIRS => 'perl' ) : (),
     ((grep { $_ eq 'PERL_CORE=1' } @ARGV) ? () : ('DEFINE' => '-DUSE_PPPORT_H')),
+    @extra,
 );


dist/Data-Dumper/Makefile.PL
```diff
@@ -1,10 +1,23 @@
 use strict;
 use warnings;
 use ExtUtils::MakeMaker;
+
+my @extra;
+push @extra, 'LICENSE' => 'perl_5'
+    unless $ExtUtils::MakeMaker::VERSION < 6.31;
```
Grinnz, line 7:
```suggestion
    if eval { ExtUtils::MakeMaker->VERSION('6.31') };
```

This will fail as a numeric comparison for underscore versions.

dist/Data-Dumper/Makefile.PL
```diff
@@ -1,10 +1,23 @@
...
+push @extra, 'META_MERGE' => {
+        resources => {
+            repository  => 'git://perl5.git.perl.org/perl.git',
+            bugtracker => 'https://github.com/Perl/perl5/issues',
+            homepage    => "http://dev.perl.org/",
+        },
+    } unless $ExtUtils::MakeMaker::VERSION < 6.46;
```
Grinnz, line 14:
```suggestion
    } if eval { ExtUtils::MakeMaker->VERSION('6.46') };
```

Same as above.

dist/Data-Dumper/Makefile.PL
jkeenan: This code is modeled on code in other `dist`/*/Makefile.PL` instances.  In all but two of those instances we do *not* use this `eval` syntax.  So I would rather have that change be done in a separate pull request for all of the distributions that need it -- assuming we have a consensus that it is better than the one used above.

Grinnz: Yes, lots of version checks are wrong.

neilb:
We discussed this issue in this week's PSC meeting.

We reckon that, in general, upstream blead distributions should direct users to Perl's issue queue. So in that context, we think that this change is right.

I guess that means the following should happen:

 1. Move all the (open, at least) issues from rt.cpan.org to Perl's github issues
 2. Do a release of DD to CPAN that has the `bugtracker` in metadata pointing to Perl's queue, and then RT will direct them there, and MetaCPAN will too
 3. For extra credit, could also set the _Distribution notes_ on RT (click on "Manage" in RT) to say that this is a core module and so bugs are tracked along with Perl itself.
 4. What have I missed?

jkeenan:
> We discussed this issue in this week's PSC meeting.
> 
> We reckon that, in general, upstream blead distributions should direct users to Perl's issue queue. So in that context, we think that this change is right.
> 
> I guess that means the following should happen:
> 
>     1. Move all the (open, at least) issues from rt.cpan.org to Perl's github issues
> 

There are currently ~~30~~ 28 tickets in [Data-Dumper's upstream bug queue](https://rt.cpan.org/Dist/Display.html?Name=Data-Dumper), each of which is more than two years old.

I don't relish the idea of importing all these tickets into our bug tracker without first reviewing them to see which ones can be handled quickly.  I acknowledge that the fact that these tickets are all so old implies that they were not quickly resolvable till now.

Perhaps we could appoint a task force of people (including at least one person who can work with `Dumper.xs`) to plow through these tickets and reduce the number we'd have to import.

>    [snip]

Thank you very much.
Jim Keenan

jkeenan:
> > [snip]
> 
> There are currently 28 tickets in [Data-Dumper's upstream bug queue](https://rt.cpan.org/Dist/Display.html?Name=Data-Dumper), each of which is more than two years old.
> 
> I don't relish the idea of importing all these tickets into our bug tracker without first reviewing them to see which ones can be handled quickly. I acknowledge that the fact that these tickets are all so old implies that they were not quickly resolvable till now.
> 
> Perhaps we could appoint a task force of people (including at least one person who can work with `Dumper.xs`) to plow through these tickets and reduce the number we'd have to import.
> 

I have now reviewed those 28 tickets and entered comments in most of them.  Some of them appear to be directly closable; others should be rejected.  But IMO we still need "a task force of people (including at least one person who can work with `Dumper.xs`) to plow through these tickets and reduce the number we'd have to import.."

Thank you very much.
Jim Keenan

nwc10:
I agree that it's not perfect, but I think that it's better than what we had. We can build on this.

I've rolled this up as part of a Data-Dumper-2.181_50.tar.gz dev release to CPAN. We'll see what the cpan testers make of it over the next couple of days.

jkeenan:
> I agree that it's not perfect, but I think that it's better than what we had. We can build on this.
> 
> I've rolled this up as part of a Data-Dumper-2.181_50.tar.gz dev release to CPAN. We'll see what the cpan testers make of it over the next couple of days.

Can you say where the source code for that dev release is located?

Thank you very much.
Jim Keenan

nwc10:
https://github.com/nwc10/perl5/tree/Data-Dumper-2.181

But I misnamed my branch - probably it should be named Data-Dumper-2.182, so I'm probably going to correct that soon.

nwc10:

PR #18835, opened 2021-05-29
regcomp.c: Split a too-large-function into two

khwilliamson:
S_regclass() is unwieldy.  This commit splits it into two nearly equal
size parts.  More could be done.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -19319,7 +19319,7 @@ S_optimize_regclass(pTHX_
 
     U8 op = END;    /* The returned node-type, initialized to an impossible
                        one.  */
-    UV value;
+    UV value = 0;
     PERL_UINT_FAST8_T i;
     UV partial_cp_count = 0;
     UV start[MAX_FOLD_FROMS+1] = { 0 }; /* +1 for the folded-to char */
--- a/regcomp.c
+++ b/regcomp.c
@@ -19350,15 +19350,11 @@ S_optimize_regclass(pTHX_
      * any run-time dependencies don't matter */
     if (start[0] == 0 && end[0] == UV_MAX) {
         if (*invert) {
-            op = OPFAIL;
-            *ret = reganode(pRExC_state, op, 0);
+            goto return_OPFAIL;
         }
         else {
-            op = SANY;
-            *ret = reg_node(pRExC_state, op);
-            MARK_NAUGHTY(1);
+            goto return_SANY;
         }
-        return op;
     }
 
     /* Similarly, for /l posix classes, if both a class and its complement
@@ -19370,13 +19366,10 @@ S_optimize_regclass(pTHX_
                 && POSIXL_TEST(posixl, namedclass + 1)) /* its complement */
             {
                 if (*invert) {
-                    op = OPFAIL;
-                    *ret = reganode(pRExC_state, op, 0);
+                    goto return_OPFAIL;
                 }
                 else {
-                    op = SANY;
-                    *ret = reg_node(pRExC_state, op);
-                    MARK_NAUGHTY(1);
+                    goto return_SANY;
                 }
                 return op;
             }
@@ -19460,15 +19453,11 @@ S_optimize_regclass(pTHX_
          * properties). */
         if (partial_cp_count == 0) {
             if (*invert) {
-                op = SANY;
-                *ret = reg_node(pRExC_state, op);
+                goto return_SANY;
             }
             else {
-                op = OPFAIL;
-                *ret = reganode(pRExC_state, op, 0);
+                goto return_OPFAIL;
             }
-
-            return op;
         }
 
         /* If matches everything but \n */
@@ -20183,6 +20172,17 @@ S_optimize_regclass(pTHX_
         }
     }
 
+    return op;
+
+  return_OPFAIL:
+    op = OPFAIL;
+    *ret = reganode(pRExC_state, op, 0);
+    return op;
+
+  return_SANY:
+    op = SANY;
+    *ret = reg_node(pRExC_state, op);
+    MARK_NAUGHTY(1);
     return op;
 }
 
--- a/regcomp.c
+++ b/regcomp.c
@@ -19325,7 +19325,7 @@ S_optimize_regclass(pTHX_
     UV start[MAX_FOLD_FROMS+1] = { 0 }; /* +1 for the folded-to char */
     UV   end[MAX_FOLD_FROMS+1] = { 0 };
     bool single_range = FALSE;
-    UV lowest_cp = 0;
+    UV lowest_cp = 0, highest_cp = 0;
 
     PERL_ARGS_ASSERT_OPTIMIZE_REGCLASS;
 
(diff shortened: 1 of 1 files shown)


regcomp.c
```diff
@@ -17524,8 +17524,8 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
 
     AV* posix_warnings = NULL;
     const bool do_posix_warnings = ckWARN(WARN_REGEXP);
-    U8 op = END;    /* The returned node-type, initialized to an impossible
-                       one.  */
+    U8 op = ANYOF;    /* The returned node-type, initialized the expected type.
+                       */
```
hvds: Does this only affect the two later references to `END`? It would be helpful to split out this change and its knock-on effects to a separate commit, just for ease of reasoning about it.

regcomp.c
```diff
@@ -19170,18 +19170,130 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
...
+         * adjacent nodes to improve efficiency. */
+        op = optimize_regclass(pRExC_state, cp_list,
+                                            only_utf8_locale_list,
+                                            upper_latin1_only_utf8_matches,
+                                            has_runtime_dependency,
+                                            posixl,
+                                            &anyof_flags, &invert, &ret, flagp);
+        RETURN_FAIL_ON_RESTART_FLAGP(flagp);
+
+        /* If optimized to something else, finish up and return */
```
hvds: I think it's worth clarifying that what we're actually checking here is whether one or more nodes have already been emitted.

regcomp.c
```diff
@@ -19291,20 +19408,20 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
...
                     /* Resolve the run-time inversion flag with this possibly
                      * inverted class */
-                    invert = invert ^ already_inverted;
+                    *invert = *invert ^ already_inverted;
 
-                    ret = reg_node(pRExC_state,
-                                   POSIXL + invert * (NPOSIXL - POSIXL));
-                    FLAGS(REGNODE_p(ret)) = classnum;
-                    goto not_anyof;
+                    op = POSIXL + *invert * (NPOSIXL - POSIXL);
```
hvds, line 19413: Is there any reason not to use the more obvious ternary here: `*invert ? NPOSIXL : POSIXL`? (Maybe for a future commit.)

regcomp.c
```diff
@@ -19291,20 +19408,20 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
...
+                    return op;
                 }
             }
         }
 
         /* khw can't think of any other possible transformation involving
          * these. */
         if (has_runtime_dependency & HAS_USER_DEFINED_PROPERTY) {
-            goto is_anyof;
+            return op;
```
hvds, line 19424: This would be more obvious as `return ANYOF;`, unless I've missed some way it could return something else.

regcomp.c
```diff
@@ -19170,18 +19170,130 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
...
+            SvREFCNT_dec(cp_list);;
+            SvREFCNT_dec(only_utf8_locale_list);
+            SvREFCNT_dec(upper_latin1_only_utf8_matches);
+            return ret;
+        }
+    }
+
+    /* Here didn't optimize, or optimized to a specialized ANYOF node.  If the
+     * former, set the particular type */
+    if (op == ANYOF) {
```
hvds: It feels like this block could quite sanely be in the new function too - it appears already to have all the inputs required. That would leave the new function making all the decisions about which op to emit, which would make reasoning about all this slightly easier. The only downside I see is that it probably involves bringing back the "is_anyof" label.

regcomp.c
```diff
@@ -19170,18 +19170,130 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
...
+     * ANYOF node.  The parameter names are the same as the corresponding
+     * variables in S_regclass.
+     *
+     * It returns the new op (ANYOF if no optimization found) and sets *ret to
+     * any created regnode.  If the new op is sufficiently like plain ANYOF, it
+     * leaves *ret unchanged for allocation in S_regclass.
+     *
+     * Certain of the parameters may be updated as a result of the changes herein */
+
+        U8 op = ANYOF; /* The returned node-type, initialized to the unoptimized
```
hvds: It would be good to include the commit to outdent these lines in this PR.

regcomp.c
```diff
@@ -17524,8 +17524,8 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
 
     AV* posix_warnings = NULL;
     const bool do_posix_warnings = ckWARN(WARN_REGEXP);
-    U8 op = END;    /* The returned node-type, initialized to an impossible
-                       one.  */
+    U8 op = ANYOF;    /* The returned node-type, initialized the expected type.
```
hvds: initialized _to_ the expected type?

khwilliamson: will do

regcomp.c
khwilliamson: will do

regcomp.c
khwilliamson: No, unless I misunderstand what you mean.  Calling the new function is skipped when the node cannot be optimized to something besides an ANYOF node; which happens for various reasons.  So deciding which ANYOF node to use has to be outside the block.

hvds: Ah right, fair point.

regcomp.c
khwilliamson: Will do

regcomp.c
khwilliamson: Let me rephrase that.  It is probably true that the new function has the required inputs to make this decision.  But that new function is sometimes skipped, so the decision has to be made as well when the function isn't involved.  Better to not repeat the logic.

regcomp.c
khwilliamson: I now regret this change.  I think it is clearer if the function returns an illegal value (END==0) when it doesn't do an optimization, and so the next iteration will have that.

regcomp.c
khwilliamson: Yes, if this is done, It shouldn't be in this commit, which is trying to avoid doing more than necessary in splitting the function.

Your proposal is clearer.   A reason not to do what you suggest is because this is a repeat of a common paradigm in this file, so it isn't out of the blue, and it works without change if a new type of POSIXL regnode (and inverse) is added, which has long been a possibility and remains so.

hvds:
@khwilliamson I suggest it would be better to separate the whitespace fixup from the comment changes; I've created hv/khwpr to demonstrate that, in case it is helpful.

jkeenan:
@khwilliamson, earlier in the discussion in this p.r. you appear to have been having second thoughts about some of the code.  Given that (and given merge conflicts), would it be better to close out this p.r. and start a fresh one?

Thank you very much.
Jim Keenan

regcomp.c
khwilliamson: changed in later PR to END

PR #18836, opened 2021-05-30
slight speed up of pattern matching

khwilliamson:


hvds:
Looks good to me; it may be worth a comment that `FLAGS(c)` is (presumably) the length of the substring, since that becomes ever less obvious as the macros introduce more layers of abstraction.

regexec.c
```diff
@@ -1884,13 +1884,12 @@ STMT_START {
...
- * at that position. */
-#define REXEC_FBC_FIND_NEXT_UTF8_BYTE_SCAN(byte, COND)      \
+/* This is like the above macro except the function returns NULL if there is no
+ * occurrence, and there is a further condition that must be matched besides
+ * the function */
+#define REXEC_FBC_FIND_NEXT_UTF8_SCAN_COND(f, COND)         \
     while (s < strend) {                                    \
-        s = (char *) memchr(s, byte, strend -s);            \
+        s = (char *) f;                                     \
         if (s == NULL) {                                    \
```
nwc10: I looked at this and found it a bit confusing. In the macros above, `f` is a function that is called, and it's written as `(f)`.

Here it's also (effectively) a function?

And shouldn't it also be written as `(f)`, partly to be consistent, partly because, dammit, macros, it thought that it is usually seen as good defensive practice to assume that what comes in might be an expression which if just interpolated as text, can go wrong thanks to operator precedence?

khwilliamson: It should be (f).  I'm adding comments to try to address the confusion

PR #18841, opened 2021-05-30
Avoid some conditionals in is...UTF8_CHAR()

khwilliamson:
These three functions to determine if the next bit of a string is UTF-8
(constrained in three different ways) have basically the same short
loop.

One of the initial conditions in the while() is always true the first
time around.  By moving that condition to the middle of the loop, we
avoid it for the common case where the loop is executed just once.  This
is when the input is a UTF-8 invariant character (ASCII on ASCII
platforms).

If the functions were constrained to require the first byte pointed to
by the input to exist, the while() could be a do {} while(), and there
would be no extra conditional in calling this vs checking if the next
character is invariant, and if not calling this.  And there would be
fewer conditionals for the case of 2 or more bytes in the character.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -1127,16 +1127,19 @@ Perl_isUTF8_CHAR(const U8 * const s0, const U8 * const e)
      * on 32-bit ASCII platforms where it trivially is an error).  Call a
      * helper function for the other platforms. */
 
-    while (s < e && LIKELY(state != 1)) {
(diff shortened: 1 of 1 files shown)


nwc10:
I couldn't actually figure out "correctness" from just the diff, so I took the original file and made simple edits to re-arrange the functions until I arrived at the same place as you did.

So I believe that it's correct. (As well as passing tests).

"Thanks applied".

PR #18843, opened 2021-05-31
gh18842: fix a pair of sprintf format warnings

hvds:
The format is '%s', so it is not appropriate to wrap the char * arguments in SVfARG().

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -2380,8 +2380,8 @@ S_parse_body(pTHX_ char **env, XSINIT_t xsinit)
 	    (void)Perl_av_create_and_unshift_one(aTHX_ &PL_preambleav,
 						 Perl_newSVpvf(aTHX_
 							       "BEGIN { do {local $!; -f q%c%s/sitecustomize.pl%c} && do q%c%s/sitecustomize.pl%c }",
-							       0, SVfARG(sitelib), 0,
-							       0, SVfARG(sitelib), 0));
+							       0, sitelib, 0,
+							       0, sitelib, 0));
 	    assert (SvREFCNT(sitelib_sv) == 1);
 	    SvREFCNT_dec(sitelib_sv);
 	}


jkeenan:
> The format is '%s', so it is not appropriate to wrap the char * arguments in SVfARG().

@hvds, would you be able and willing to create a smoke-me branch from this p.r.?

Thank you very much.
Jim Keenan

hvds:
I'm not sure how `t/porting/cmp_version.t` can complain on `lib/feature.pm` with `-Dusethreads`, but not with `-Uusethreads`, but I'm going to ignore this:
```
t/porting/checkcfgvar ............................................ ok
# not ok ***05 - lib/feature.pm version ***.65
t/porting/cmp_version ............................................ FAILED at test ***05
t/porting/copyright .............................................. ok
[...]
    porting/cmp_version.t
Failed *** test out of 2544, 99.96% okay.
```

It also looks like something is doing `s/1/***/g` on the output. @xenu is this related to work you're doing on the workflows?

hvds:
> > The format is '%s', so it is not appropriate to wrap the char * arguments in SVfARG().
> 
> @hvds, would you be able and willing to create a smoke-me branch from this p.r.?

@jkeenan I'm able, but it doesn't seem like a useful thing to do - there are no platform-dependent macros in the format string, it's just a straight "%s", so it is a very straightforward fix. In any case, Karl merged this while I was responding.

jkeenan:
> > > The format is '%s', so it is not appropriate to wrap the char * arguments in SVfARG().
> > 
> > 
> > @hvds, would you be able and willing to create a smoke-me branch from this p.r.?
> 
> @jkeenan I'm able, but it doesn't seem like a useful thing to do - there are no platform-dependent macros in the format string, it's just a straight "%s", so it is a very straightforward fix. In any case, Karl merged this while I was responding.

Well then, could you tell me the category of the warnings which this commit should have fixed?

Thank you very much.
Jim Keenan

hvds:
> Well then, could you tell me the category of the warnings which this commit should have fixed?

@jkeenan Ah sorry, it's this pair of "-Wformat=":
```
perl.c: In function 'S_parse_body':
perl.c:2382:64: warning: format '%s' expects argument of type 'char *', but argument 4 has type 'void *' [-Wformat=]
 2382 |                                                                "BEGIN { do {local $!; -f q%c%s/sitecustomize.pl%c} && do q%c%s/sitecustomize.pl%c }",
      |                                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
perl.c:2382:64: warning: format '%s' expects argument of type 'char *', but argument 7 has type 'void *' [-Wformat=]
```

PR #18846, opened 2021-06-01
Rename G_ARRAY to G_LIST

leonerd:
As per my suggestion

  https://www.nntp.perl.org/group/perl.perl5.porters/2021/05/msg260264.html

Note that the three XS modules modified in 699675f are less essential here, and if it's felt excessive to bump the version number of them solely for a no-behaviour change in the .xs files, I can revert that commit.

If merged, I will follow this up with a series of edits and module releases to the various dual-life code under `dist/` which is also using this macro, and suggest to upstream authors of modules under `cpan/` that they do similar changes. (This should be easy as the majority of hits in there are in `Scalar-List-Utils`, maintained by one `PEVANS`, aka me ;) )


cop.h
```diff
@@ -1094,10 +1094,11 @@ struct context {
 
 #define CXINC (cxstack_ix < cxstack_max ? ++cxstack_ix : (cxstack_ix = cxinc()))
 
-#define G_SCALAR	2
-#define G_ARRAY		3
-#define G_VOID		1
-#define G_WANT		3
+#define G_SCALAR        2
+#define G_LIST          3
+#define G_ARRAY         3      /* name prior to 5.35.1 */
```
ilmari: To avoid it accidentally sneaking back in core code, we could do:

```c
#ifndef PERL_CORE
#define G_ARRAY G_LIST
#endif 
```

khwilliamson:
On 6/1/21 8:56 AM, Dagfinn Ilmari Mannsåker wrote:
> 
> 
> In cop.h <https://github.com/Perl/perl5/pull/18846#discussion_r643183457>:
> 
>> @@ -1094,10 +1094,11 @@ struct context {
>   
>   #define CXINC (cxstack_ix < cxstack_max ? ++cxstack_ix : (cxstack_ix = cxinc()))
>   
> -#define G_SCALAR	2
> -#define G_ARRAY		3
> -#define G_VOID		1
> -#define G_WANT		3
> +#define G_SCALAR        2
> +#define G_LIST          3
> +#define G_ARRAY         3      /* name prior to 5.35.1 */
> 
> To avoid it accidentally sneaking back in core code, we cold do:
> 
> #ifndef  PERL_CORE
> #define  G_ARRAY  G_LIST
> #endif  
> 
> —

+1 to this; we've done this before.

I would be against dropping G_ARRAY outside core anytime soon.  There's 
plenty of obsolete constants lying around our headers, and AFAIK no harm 
is done.

Certainly PPPort needs to get the new constant, so distros can change 
with confidence if they choose.

P.S.

I have long toyed with the idea of adding a 'want_list()' to perl

cop.h
leonerd: Good idea. Added in c5b4740f0c. PTAL

ilmari (approved):

leonerd:
mingw64 CI is upset here, but it doesn't look at all related to this change:

```
 Test Summary Report
-------------------
re/uniprops01.t                                                    (Wstat: 2304 Tests: 0 Failed: 0)
  Non-zero exit status: 9
../dist/Net-Ping/t/450_service.t                                   (Wstat: 0 Tests: 26 Failed: 0)
  TODO passed:   9
../ext/IPC-Open3/t/IPC-Open3.t                                     (Wstat: 0 Tests: 45 Failed: 0)
  TODO passed:   25
```

In addition, the other Windows CI was happy.

ext/XS-APItest/APItest.xs
```diff
@@ -3947,7 +3947,7 @@ CODE:
         av_push(av, SvREFCNT_inc(TOPs));
         break;
 
-    case G_ARRAY:
+    case G_LIST:
```
wolfsage, line 3950: I kinda think APItest.xs should use the old stuff since it's testing the API and the API is what we expect non-core modules to use (and possibly steal from APItest.xs's examples)

op.c
```diff
@@ -1663,7 +1663,7 @@ Perl_alloc_LOGOP(pTHX_ I32 type, OP *first, OP* other)
 =for apidoc op_contextualize
 
 Applies a syntactic context to an op tree representing an expression.
-C<o> is the op tree, and C<context> must be C<G_SCALAR>, C<G_ARRAY>,
+C<o> is the op tree, and C<context> must be C<G_SCALAR>, C<G_LIST>,
```
wolfsage, line 1666: Again, will this cause folks to write code that doesn't easily port back to previous perls, or is Devel::PPPort supposed to solve that?

leonerd: It's a trivial oneliner in ppport.h:

```c
#define  G_LIST  G_ARRAY
```

xenu (approved):

PR #18848, opened 2021-06-02
only croak on undef = ... not on (complex... undef) = ...

### LINKED ISSUE #18834: Convert::ASN1 breaks with commit 2583ef601d381a8a3e6e9b5f44db42bac0d537ab
Commit https://github.com/Perl/perl5/commit/2583ef601d381a8a3e6e9b5f44db42bac0d537ab introduced a change that breaks tests on Convert:ASN1 for code that previously worked on all prior versions with errors:

```
PERL_DL_NONLAZY=1 "/home/cpan/bin/perl-blead/bin/perl5.35.1" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t
Can't modify undef operator in scalar assignment at /home/cpan/.cpan/build/Convert-ASN1-0.29-0/blib/lib/Convert/ASN1/_decode.pm line 98, near ");"
BEGIN not safe after errors--compilation aborted at /home/cpan/.cpan/build/Convert-ASN1-0.29-0/blib/lib/Convert/ASN1/_decode.pm line 607.
Compilation failed in require at /home/cpan/.cpan/build/Convert-ASN1-0.29-0/blib/lib/Convert/ASN1.pm line 435.
BEGIN failed--compilation aborted at /home/cpan/.cpan/build/Convert-ASN1-0.29-0/blib/lib/Convert/ASN1.pm line 441.
Compilation failed in require at t/00prim.t line 7.
BEGIN failed--compilation aborted at t/00prim.t line 7.

```

Convert::ASN1
https://github.com/gbarr/perl-Convert-ASN1/blob/9c029e84c2d308c525e2d8e08f1432f670be29c4/lib/Convert/ASN1/_decode.pm#L97

op.c: 
https://github.com/Perl/perl5/blob/0b47b39244b67fe19dc971903ce28a7a637c2a9a/op.c#L4294

I will review the code in Convert::ASN1 I suppose it is possible it was not hitting that in runtime previously in the tests.

tonycoz:
@richardleach I'm not sure whether it was intended to croak on the case below

The previous check would trigger on code like:

  ($x ? $y : undef) = 1;

I also moved the test from t/run/fresh_perl.t to t/run/croak/op and
added a test for the broken case.

fixes #18834

### DIFF
--- a/op.c
+++ b/op.c
@@ -4291,8 +4291,6 @@ Perl_op_lvalue_flags(pTHX_ OP *o, I32 type, U32 flags)
 
     switch (o->op_type) {
     case OP_UNDEF:
-        if (type == OP_SASSIGN)
-            goto nomod;
         PL_modcount++;
         goto do_next;
 
@@ -13675,6 +13673,12 @@ Perl_ck_sassign(pTHX_ OP *o)
 
     if (OpHAS_SIBLING(kid)) {
         OP *kkid = OpSIBLING(kid);
+
+        if (kkid->op_type == OP_UNDEF) {
+            /* diag_listed_as: Can't modify %s in %s */
+            yyerror("Can't modify undef operator in scalar assignment");
+        }
+
         /* For state variable assignment with attributes, kkid is a list op
            whose op_last is a padsv. */
         if ((kkid->op_type == OP_PADSV ||
--- a/t/lib/croak/op
+++ b/t/lib/croak/op
@@ -251,3 +251,16 @@ Type of arg 1 to pop must be array (not private hash) at - line 4, near "%a;"
 Type of arg 1 to shift must be array (not private hash) at - line 5, near "%a;"
 Type of arg 1 to unshift must be array (not private hash) at - line 6, near "1;"
 Execution of - aborted due to compilation errors.
+########
+# NAME detect scalar assignment to undef during compilation
+BEGIN { undef = 0 }
+EXPECT
+Can't modify undef operator in scalar assignment at - line 1, near "0 }"
+BEGIN not safe after errors--compilation aborted at - line 1.
+########
+# NAME but don't trigger on a more complex left-hand expression involving undef
+my $x = 1;
+my $y;
+($x ? $y : undef) = 0;
+EXPECT
+OPTIONS nonfatal
--- a/t/run/fresh_perl.t
+++ b/t/run/fresh_perl.t
@@ -192,11 +192,6 @@ EXPECT
 Illegal division by zero at - line 1.
 BEGIN failed--compilation aborted at - line 1.
 ########
-BEGIN { undef = 0 }
-EXPECT
-Can't modify undef operator in scalar assignment at - line 1, near "0 }"
-BEGIN not safe after errors--compilation aborted at - line 1.
-########
 {
     package foo;
     sub PRINT {


hvds:
> @richardleach I'm not sure whether it was intended to croak on the case below
> 
> The previous check would trigger on code like:
> 
> ($x ? $y : undef) = 1;

@tonycoz do you believe that authors are deliberately writing that expecting it to be equivalent to `die "Modification of a read-only value attempted" unless $x; $y = 1`?

> I also moved the test from t/run/fresh_perl.t to t/run/croak/op and
> added a test for the broken case.
> 
> fixes #18834

My reading of the author's comments on #18834 is that he acknowledges this as a bug: even though his tests did not exercise the case, his expectation was that the fallthrough case was a valid one, and he expected it to call the function on the RHS and discard the result.

So the change as it stands will show up bugs like this by flagging them as compile-time errors. Even if we end up changing this before 5.36 release, I do not think we should rush to do so much earlier than that release.

tonycoz:
I don't know what the author intended,

But the original change #18785 , and the original bug #18779 only discuss the simple case, and the discussion in #18785 I think implies that we fail at compile-time when the code would have failed at runtime, which this code doesn't.

I don't mind if we choose to keep this behaviour, but to my mind it doesn't quite match the description.

richardleach:
@tonycoz I don't mind which behaviour ends up being used. I hadn't come across complex (or really any) cases of people trying to scalar assign to undef, so wasn't aware that they were a thing.

I do share Hugo's concern regarding the programmer's expectation of:
`($x ? $y : undef) = 1;`

It does seem like a way of throwing the value away, like with list assignment to undef, and the fact that it could die might be surprising/unintentional. If the undef occurs only in rarely used code paths, it causing an application to surprisingly die does seem bad IMHO.

Grinnz:
> @tonycoz I don't mind which behaviour ends up being used. I hadn't come across complex (or really any) cases of people trying to scalar assign to undef, so wasn't aware that they were a thing.
> 
> I do share Hugo's concern regarding the programmer's expectation of:
> `($x ? $y : undef) = 1;`
> 
> It does seem like a way of throwing the value away, like with list assignment to undef, and the fact that it could die might be surprising/unintentional. If the undef occurs only in rarely used code paths, it causing an application to surprisingly die does seem bad IMHO.

This of course would be reasonable to warn about at compile time. But making it an error breaks bugward compatibility of scripts that were running perfectly fine because they never actually hit that case at runtime.

richardleach:
> > I do share Hugo's concern regarding the programmer's expectation of:
> > `($x ? $y : undef) = 1;`
> > It does seem like a way of throwing the value away, like with list assignment to undef, and the fact that it could die might be surprising/unintentional. If the undef occurs only in rarely used code paths, it causing an application to surprisingly die does seem bad IMHO.
> 
> This of course would be reasonable to warn about at compile time. But making it an error breaks bugward compatibility of scripts that were running perfectly fine because they never actually hit that case at runtime.

True. But Tony's version would have the same effect on scripts with `undef = $foo;` in an effectively unreachable code block. (I appreciate that's probably splitting hairs.)

Grinnz:
> True. But Tony's version would have the same effect on scripts with `undef = $foo;` in an effectively unreachable code block. (I appreciate that's probably splitting hairs.)

In my opinion that is an acceptable degree, because it's plain that was never going to work, whereas clearly there has been instances where developers assumed (wrongly) a conditional assignment would simply be ignored.

richardleach:
> But making it an error breaks bugward compatibility of scripts that were running perfectly fine because they never actually hit that case at runtime.

I'm a bit grumpy, since this will keep life easier for people who have fragile code at the expense of people who might be about to write new fragile code. But I appreciate that's the trade off of bugward compatibility.

hvds:
> I don't mind which behaviour ends up being used. I hadn't come across complex (or really any) cases of people trying to scalar assign to undef, so wasn't aware that they were a thing.
> 
> I do share Hugo's concern regarding the programmer's expectation of:
> `($x ? $y : undef) = 1;`
> 
> It does seem like a way of throwing the value away, like with list assignment to undef, and the fact that it could die might be surprising/unintentional. If the undef occurs only in rarely used code paths, it causing an application to surprisingly die does seem bad IMHO.

Maybe the better solution, then, would be to make it a legitimate way of throwing a value away, same as in the list case.

richardleach:
> Maybe the better solution, then, would be to make it a legitimate way of throwing a value away, same as in the list case.

That would definitely add consistency.

richardleach:
> Maybe the better solution, then, would be to make it a legitimate way of throwing a value away, same as in the list case.

@Grinnz / @tonycoz - what are your thoughts on that?

Issue #18847, opened 2021-06-02
Perl Install 'make test' Dumps Core on FreeBSD


cjg-cguevara:
**Description**
Perl install 'make test' dumps core on FreeBSD 11, 12, and 13.  Some examples:
Jun  1 03:51:46 cjg-freebsd11 kernel: pid 37335 (perl), jid 0, uid 1001: exited on signal 6 (core dumped)
Jun  1 04:35:59 cjg-freebsd11 kernel: pid 3977 (try), jid 0, uid 1001: exited on signal 10 (core dumped)
Jun  1 00:13:06 cjg-freebsd12 kernel: pid 75501 (try), jid 0, uid 1001: exited on signal 10 (core dumped)
Jun  1 00:25:14 cjg-freebsd12 kernel: pid 88122 (perl), jid 0, uid 1001: exited on signal 6 (core dumped)
Jun  1 01:16:53 cjg-freebsd13 kernel: pid 4114 (try), jid 0, uid 1001: exited on signal 10 (core dumped)
Jun  1 01:27:07 cjg-freebsd13 kernel: pid 17821 (perl), jid 0, uid 1001: exited on signal 6 (core dumped)

**Steps to Reproduce**
wget https://github.com/Perl/perl5/archive/refs/heads/blead.zip
...
./Configure -des -Dusedevel -Duse64bitall
make
make test

tonycoz:
Two SIGABRT (signal 6) are generated intentionally to test CORE::dump (t/op/dump.t)

I didn't see the SIGBUS errors when I tested.

Did any tests fail?

If you want to produce more information:

As root:

```
# mkdir /tmp/core # or elsewhere
# chmod 777 /tmp/core # so anyone can generate a core here
# sysctl kern.corefile="/tmp/core/%N.%P.core"
```
run tests and look for files in /tmp/core:

```
$ ls -l /tmp/core
total 357
-rw-------  1 tony  wheel  9924608 Jun  2 16:26 perl.95561.core
-rw-------  1 tony  wheel  9912320 Jun  2 16:26 perl.95563.core
```
For each core file:
```
$ cd t
$ gdb ./perl /tmp/core/perl.95561.core
...
# this is the backtrace from the CORE::dump() tests
(gdb) bt
#0  0x00000008008d8bda in kill () from /lib/libc.so.7
#1  0x00000000005ee670 in Perl_pp_goto () at pp_ctl.c:3203
[...]

PR #18850, opened 2021-06-02
Skip test of CORE::dump on FreeBSD

jkeenan:
Because on this OS this test writes to /var/log/messages

For: https://github.com/Perl/perl5/issues/18847

### DIFF
--- a/t/op/dump.t
+++ b/t/op/dump.t
@@ -25,6 +25,9 @@ skip_all("only tested on devel builds")
 skip_all("no point in dumping on $^O")
   unless $^O =~ /^(linux|.*bsd|solaris|darwin)$/;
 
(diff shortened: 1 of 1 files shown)


tonycoz:
Both tests in this file dump core, this only skips one of them.

jkeenan:
> Both tests in this file dump core, this only skips one of them.

Updated to skip both tests; two commits squashed.

tonycoz:
All the tests in this file (all 2 of them) dump core, wouldn't this be better as a skip_all along with all the others at the top of the file?

jkeenan:
> All the tests in this file (all 2 of them) dump core, wouldn't this be better as a skip_all along with all the others at the top of the file?

Done; squashed commits.

tonycoz (approved):

PR #18851, opened 2021-06-02
Use G_LIST in dist/threads

leonerd:


khwilliamson:
ppport needs releasing before this can be merged

leonerd:
Oh I thought I checked that `threads` doesn't actually use ppport.h; it was one of the only dists that doesn't

dist/threads/threads.xs
```diff
@@ -46,6 +46,10 @@
 # define CLANG_DIAG_RESTORE_DECL CLANG_DIAG_RESTORE dNOOP
 #endif
 
+#ifndef G_LIST
```
khwilliamson: Remove these lines, and merge this PR

khwilliamson:
Devel::PPPort is no on CPAN with G_LIST.  This PR can now be merged, but after the changes I requested

Leont:
@jdhedden ^

dist/threads/threads.xs
leonerd: Now done by 90e743c0ed

PR #18853, opened 2021-06-03
prevent undefined behaviour at a language level while probing getenv

tonycoz:
This test in Configure tries to probe for undefined behaviour in
getenv(), but provokes undefined behaviour in C/C++ by falling off
the end of a function with a non-void return type.

Without optimization clang++ generated a ud2 instruction here on
amd64 producing an illegal instruction exception.  With optimization
the test case fell off the end and started re-executing main(),
eventually producing a SIGBUS.

Simply dropping the value of getenv() here and returning NULL wasn't
useful, under -O2 the compiler optimized away the getenv() call,
voiding the whole point of the test.

fixes the Configure core dump from #18847

### DIFF
--- a/Configure
+++ b/Configure
@@ -14249,7 +14249,7 @@ $cat >try.c <<EOCP
 void *
 thread_start(void * arg)
 {
-    (void *) getenv("HOME");
+    return (void *) getenv("HOME");
 }
 
 int main() {
@@ -14280,7 +14280,7 @@ int main() {
         exit(2);
     }
 
-    exit(! strcmp(main_buffer, save_main_buffer) == 0);
+    exit(! (strcmp(main_buffer, save_main_buffer) == 0));
 }
 EOCP
 val=


jkeenan:
> This test in Configure tries to probe for undefined behaviour in
> getenv(), but provokes undefined behaviour in C/C++ by falling off
> the end of a function with a non-void return type.
> 
> Without optimization clang++ generated a ud2 instruction here on
> amd64 producing an illegal instruction exception. With optimization
> the test case fell off the end and started re-executing main(),
> eventually producing a SIGBUS.
> 
> Simply dropping the value of getenv() here and returning NULL wasn't
> useful, under -O2 the compiler optimized away the getenv() call,
> voiding the whole point of the test.
> 
> fixes the Configure core dump from #18847

@tonycoz, thank you for looking into this.

I tested your branch on FreeBSD-12 using clang++ as the C-compiler.  Nothing was written to `/var/log/messages` during Configure.  During `make test_harness`, the following records were written:
```
Jun  3 11:16:39 perlmonger kernel: pid 90338 (perl), jid 0, uid 1002: exited on signal 6 (core dumped)
Jun  3 11:16:39 perlmonger kernel: pid 90516 (perl), jid 0, uid 1002: exited on signal 6 (core dumped)
```
... but we expect them because we haven't dealt with `t/op/dump.t` yet.

I have created a smoke branch from yours and am running it through a large configuration that includes clang++:  smoke-me/jkeenan/tonycoz-18847-getenv-sigbus.

Thank you very much.
Jim Keenan

Tux:
The backport restricts to just `U/threads/d_getenv_thread.U` which I already updated (but not committed/pushed) in meta
Looks good. I agree

khwilliamson (approved):

Issue #18830, opened 2021-05-28
[doc] Re-consider use of "grandfather" as a verb


oalders:
**Where**

```
regen/mk_invlists.pl:3181:27:            # support the grandfathered L_ Gc property value
Porting/epigraphs.pod:5076:58:hunters leading the wolf; and winding up the procession, grandfather and the
lib/Unicode/UCD.pm:387:7:    # grandfathered-in 'L_', which retains a single trailing underscore.
lib/Unicode/UCD.pm:2612:14:the old_form grandfathered-in C<"L_">, which is better written as C<"LC">, and
lib/Unicode/UCD.pm:2854:58:name (except for the trailing underscore in the old-form grandfathered-in
lib/Unicode/UCD.pm:2941:53:(except for the trailing underscore in the old-form grandfathered-in general
lib/Unicode/UCD.pm:3278:10:old-form grandfathered-in property C<"L_">, which is better written as C<"LC">,
lib/Tie/Hash.pm:63:63:and allow for selective overwriting of methods.  B<Tie::Hash> grandfathers the
lib/Tie/Scalar.pm:44:34:method is provided as a means of grandfathering, for classes that forget to
lib/Config.t:34:5:my %grandfathers = ( PERL_VERSION       => 'PATCHLEVEL',
lib/Config.t:38:31:while( my($new, $old) = each %grandfathers ) {
regcomp.c:24439:29:     * This special case is grandfathered in: 'L_' and 'GC=L_' are accepted and
pp_pack.c:614:28:      if (code == ','){ /* grandfather in commas but with a warning */
configpm:240:7:    # grandfather PATCHLEVEL and SUBVERSION and CONFIG
pod/perldiag.pod:592:27:The use of backslashes is grandfathered on the right-hand side of a
pod/perl56delta.pod:2734:15:Release 5.005 grandfathered old global symbol names by providing preprocessor
pod/perl561delta.pod:3339:15:Release 5.005 grandfathered old global symbol names by providing preprocessor
pod/perlre.pod:2942:9:This is grandfathered (for \1 to \9) for the RHS of a substitute to avoid
cpan/ExtUtils-MakeMaker/lib/ExtUtils/MakeMaker.pm:2674:15:Release 5.005 grandfathered old global symbol names by providing preprocessor
```

**Description**
[...]

PR #18854, opened 2021-06-03
Replace 'slave' with 'client'

jkeenan:
This commit moves us a step forward on eliminating 'master/slave' terminology.

Its scope is limited to those files within the Perl 5 core distribution which
are not CPAN-upstream.  Within that scope, we leave untouched the use of the
term 'slave' in epigraphs and other literary quotations and its use in data
coming from outside sources (e.g., Unicode code points).

Increment $VERSION in lib/perl5db.pl.

This partially addresses https://github.com/Perl/perl5/issues/18830

### DIFF
--- a/lib/perl5db.pl
+++ b/lib/perl5db.pl
@@ -1225,6 +1225,16 @@ $pidprompt = '';
 our ($client_editor);
 *emacs = $client_editor if $client_editor;    # May be used in afterinit()...
 
+package RenameVariable;
+require Tie::Scalar;
+our @ISA = qw(Tie::Scalar);
+our $message = '$slave_editor deprecated; use $client_editor instead';
+sub FETCH { warn $message; }
+sub STORE { die $message; }
+
+
+package DB;
+
 =head2 READING THE RC FILE
 
 The debugger will read a file of initialization options if supplied. If
--- a/lib/perl5db.pl
+++ b/lib/perl5db.pl
@@ -393,7 +393,7 @@ Controls the output of trace information.
 
 =back
 
-=head4 C<$slave_editor>
+=head4 C<$client_editor>
 
 1 if C<LINEINFO> was directed to a pipe; 0 otherwise.
 
@@ -531,7 +531,7 @@ BEGIN {
 use vars qw($VERSION $header);
 
 # bump to X.XX in blead, only use X.XX_XX in maint
-$VERSION = '1.60';
+$VERSION = '1.61';
 
 $header = "perl5db.pl version $VERSION";
 
@@ -1221,9 +1221,9 @@ else {
 use vars qw($pidprompt);
 $pidprompt = '';
 
-# Sets up $emacs as a synonym for $slave_editor.
-our ($slave_editor);
-*emacs = $slave_editor if $slave_editor;    # May be used in afterinit()...
+# Sets up $emacs as a synonym for $client_editor.
+our ($client_editor);
+*emacs = $client_editor if $client_editor;    # May be used in afterinit()...
 
 =head2 READING THE RC FILE
 
@@ -1504,7 +1504,7 @@ if ($notty) {
 =pod
 
 If there is a TTY, we have to determine who it belongs to before we can
-proceed. If this is a slave editor or graphical debugger (denoted by
+proceed. If this is a client editor or graphical debugger (denoted by
 the first command-line switch being '-emacs'), we shift this off and
 set C<$rl> to 0 (XXX ostensibly to do straight reads).
 
@@ -1512,9 +1512,9 @@ set C<$rl> to 0 (XXX ostensibly to do straight reads).
 
 else {
 
-    # Is Perl being run from a slave editor or graphical debugger?
-    # If so, don't use readline, and set $slave_editor = 1.
-    if ($slave_editor = ( @main::ARGV && ( $main::ARGV[0] eq '-emacs' ) )) {
+    # Is Perl being run from a client editor or graphical debugger?
+    # If so, don't use readline, and set $client_editor = 1.
+    if ($client_editor = ( @main::ARGV && ( $main::ARGV[0] eq '-emacs' ) )) {
         $rl = 0;
         shift(@main::ARGV);
     }
@@ -1584,12 +1584,12 @@ We then determine what the console should be on various systems:
 =back
 
 Several other systems don't use a specific console. We C<undef $console>
-for those (Windows using a slave editor/graphical debugger, NetWare, OS/2
-with a slave editor).
+for those (Windows using a client editor/graphical debugger, NetWare, OS/2
+with a client editor).
 
 =cut
 
-    if ( ( $^O eq 'MSWin32' ) and ( $slave_editor or defined $ENV{EMACS} ) ) {
+    if ( ( $^O eq 'MSWin32' ) and ( $client_editor or defined $ENV{EMACS} ) ) {
 
         # /dev/tty is binary. use stdin for textmode
         $console = undef;
@@ -1603,7 +1603,7 @@ with a slave editor).
 
     # In OS/2, we need to use STDIN to get textmode too, even though
     # it pretty much looks like Unix otherwise.
-    if ( defined $ENV{OS2_SHELL} and ( $slave_editor or $ENV{WINDOWID} ) )
+    if ( defined $ENV{OS2_SHELL} and ( $client_editor or $ENV{WINDOWID} ) )
     {    # In OS/2
         $console = undef;
     }
@@ -1698,7 +1698,7 @@ and if we can.
     _autoflush($OUT);
 
     # Line info goes to debugger output unless pointed elsewhere.
-    # Pointing elsewhere makes it possible for slave editors to
+    # Pointing elsewhere makes it possible for client editors to
     # keep track of file and position. We have both a filehandle
     # and a I/O description to keep track of.
     $LINEINFO = $OUT     unless defined $LINEINFO;
@@ -1725,7 +1725,7 @@ and then call the C<afterinit()> subroutine if there is one.
             print $OUT "\nLoading DB routines from $header\n";
             print $OUT (
                 "Editor support ",
-                $slave_editor ? "enabled" : "available", ".\n"
+                $client_editor ? "enabled" : "available", ".\n"
             );
             print $OUT
 "\nEnter h or 'h h' for help, or '$doccmd perldebug' for more help.\n\n";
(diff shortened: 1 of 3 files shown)


khwilliamson:
I have several concerns about this

I don't think client is necessarily analogous to slavery

Slavery isn't necessarily related to race, which is what I presume this ticket is suggesting.  It is mentioned in the Bible, and no one is going to go expunging references to it from there.

Are there people actually harmed by seeing this term?  If so, my guess is that it would be mostly people in the U.S.  Is this a Ameri-centric issue?  

Slavery exists today.  I was once in a place in Kenya where it had been still legal 60 years earlier!  There are reports of human trafficking all the time.  In my town, some years ago, a Thai restaurant owner had effectively enslaved workers he had imported from back home in Thailand.  

So the origins of the term aren't racist, and isn't used in a racist way in significant contexts.

I have not examined this ticket to see if client is a good-enough term; it it is I don't have any objections to changing.  But I do think we need to examine things on a case-by-case basis

jkeenan:
> I have several concerns about this
>  [snip]
> 
> I have not examined this ticket to see if client is a good-enough term; if it is I don't have any objections to changing. But I do think we need to examine things on a case-by-case basis

@khw, if you read the pull request, you will see that I *did* "examine things on a case-by-case basis.  The majority of instances of the term "slave" in the codebase are left untouched, either because (a) they're CPAN-upstream; (b) they're in literary quotations, *e.g.,* epigraphs; or (c) they're found in source data obtained from outside sources like Unicode.

(The most prominent use of the term "slave" in the CPAN-upstream part of the core distribution is in Net::NNTP.  But even there the term is excisable because the RFC around which that module was originally developed has been superseded by a later one in which the `SLAVE` method was removed completely -- 15 years ago!  See: https://rt.cpan.org/Ticket/Display.html?id=136706)

The "master/slave" dichotomy is has deep roots in the thinking of technologists.  Forty years ago I worked in type shops where typesetters like myself entered keystrokes into a PDP-11-based system that digitally composed the type (Quadex) and then generated instructions to a CRT-based system (Linotype) which actually cast the type onto film or RC paper.  The two systems came from different manufacturers even though the shop could have used a digital keyboarding and composition system from the same manufacturer as the phototypsetter.  So it was common to say that Quadex was the master and Linotron was the slave.  At trade shows salesmen would use this terminology without blinking.

Unfortunately that distinction also mirrored the distinction between the human beings who actually worked on these two systems.  At least initially, all the Quadex operators were white.  The machine tenders, who loaded the phototypesetters with film and then subsequently ran it through developing machines, were primarily black or Latino -- the descendants of enslaved people.  Over time the workers in the latter category taught themselves how to type and then how to set type.  At that point it became obvious to even the most conservative workers in the shop that the "master/slave" dichotomy had to go.  We very quickly swapped in "front end/back end."

Thank you very much.
Jim Keenan

rjbs:
@jkeenan Regarding the debugger, this looks fine to me, but I think it may be worth noting in the documentation for `$client_editor` that `$slave_editor` was once used.  This was a documented piece of the guts, and possibly someone out there is still looking at it.

Another option, which I probably consider very slightly preferable, would be to place a tied variable in the `$slave_editor` slot that carps when read (and returns the value of `$client_editor`) and croaks when assigned to.  That could leave the documentation just as you have it.

In either case, I believe that it is unlikely that the existing variable is being looked at from without, but that's just a hunch and could be wrong.

**On the other hand,** I believe your edit to `perlipc` is a change for the worse.  The term "client server" isn't quite nonsensical, but it's just not accurate here.  It's a subprocess of the server that handles a single client.  It is not, itself, a client.  I would suggest, instead, saying "child process" instead of "slave server".

rjbs (changes requested):
I can't believe that mere days after you said "hey, nice that you can provide all comments and the result in one email," here I am accidentally sending more than one email.  Sorry!

I have provided more detailed feedback in my previous comment.

lib/perl5db.pl
```diff
@@ -1223,8 +1223,19 @@ =head2 SETTING UP THE DEBUGGER GREETING
...
 # Sets up $emacs as a synonym for $client_editor.
 our ($client_editor);
+if ($ENV{EMACS}) { print STDERR "ZZZ: EMACS\n"; }
 *emacs = $client_editor if $client_editor;    # May be used in afterinit()...
 
+package RenameVariable;
+require Tie::Scalar;
+our @ISA = qw(Tie::Scalar);
+our $message = '$slave_editor deprecated; use $client_editor instead';
+sub FETCH { warn $message; }
```
Leont: This should probably return $client_editor

jkeenan: Thanks.  p.r. updated.

rjbs:
I would like to help more, but probably can't right this minute.

I built and tested this branch by hand.

[oops, hit "comment" too soon! adding real content now]

```
  DB<1> p $DB::client_editor
0
  DB<2> p $DB::slave_editor

```

Nothing actually uses the tie class, so this didn't do anything.  But then this happened:

```
  DB<3> tie $DB::slave_editor, 'RenameVariable'
RenameVariable must define either a TIESCALAR() or a new() method at (eval 11)[lib/perl5db.pl:741] line 2.

  DB<4> m RenameVariable
FETCH
STORE
via Tie::Scalar: TIESCALAR
via Tie::Scalar: carp
via Tie::Scalar: confess
via Tie::Scalar: croak
via Tie::Scalar: new
via UNIVERSAL: DOES
via UNIVERSAL: VERSION
via UNIVERSAL: can
via UNIVERSAL: isa
  DB<5> tie $DB::slave_editor, 'RenameVariable'
RenameVariable must define either a TIESCALAR() or a new() method at (eval 12)[lib/perl5db.pl:741] line 2.

  DB<6> sub RenameVariable::TIESCALAR { goto &Tie::Scalar::TIESCALAR }

  DB<7> tie $DB::slave_editor, 'RenameVariable'
RenameVariable must define either a TIESCALAR() or a new() method at (eval 14)[lib/perl5db.pl:741] line 2.
```

I think probably I'm just being thick, but I don't have the time right now to work through my brain haze.

jkeenan:
> [snip]
> 
> 
> ```
>   DB<1> p $DB::client_editor
> 0
>   DB<2> p $DB::slave_editor
> ```
> 

What did you say on the command-line to start the debugger?

> Nothing actually uses the tie class, so this didn't do anything. But then this happened:
> 
> ```
>   DB<3> tie $DB::slave_editor, 'RenameVariable'
> RenameVariable must define either a TIESCALAR() or a new() method at (eval 11)[lib/perl5db.pl:741] line 2.
> 
>   DB<4> m RenameVariable
> FETCH
> STORE
> via Tie::Scalar: TIESCALAR
> via Tie::Scalar: carp
> via Tie::Scalar: confess
> via Tie::Scalar: croak
> via Tie::Scalar: new
> via UNIVERSAL: DOES
> via UNIVERSAL: VERSION
> via UNIVERSAL: can
> via UNIVERSAL: isa
>   DB<5> tie $DB::slave_editor, 'RenameVariable'
> RenameVariable must define either a TIESCALAR() or a new() method at (eval 12)[lib/perl5db.pl:741] line 2.
> 
>   DB<6> sub RenameVariable::TIESCALAR { goto &Tie::Scalar::TIESCALAR }
> 
>   DB<7> tie $DB::slave_editor, 'RenameVariable'
> RenameVariable must define either a TIESCALAR() or a new() method at (eval 14)[lib/perl5db.pl:741] line 2.
> ```
> 
> I think probably I'm just being thick, but I don't have the time right now to work through my brain haze.

As I reported in https://github.com/Perl/perl5/issues/18868, no matter how correct we get the code for a tied variable inside `lib/perl5db.pl`, we're treading dangerously if we don't exercise the code for it in `lib/perl5db.t`.  I've tried to figure out how to test what I'll call the "graphical debugger" block; "brain haze" is an accurate description of how far I've got.

So, in the short run we might be better off simply making *only* a documentation change as you suggested in https://github.com/Perl/perl5/pull/18854#issuecomment-855528948 above.  Would you agree?

Thank you very much.
Jim Keenan

jkeenan:
> > [snip]
> > ```
> >   DB<1> p $DB::client_editor
> > 0
> >   DB<2> p $DB::slave_editor
> > ```
> 
> What did you say on the command-line to start the debugger?
> 
> > Nothing actually uses the tie class, so this didn't do anything. But then this happened:
> > ```
> >   DB<3> tie $DB::slave_editor, 'RenameVariable'
> > RenameVariable must define either a TIESCALAR() or a new() method at (eval 11)[lib/perl5db.pl:741] line 2.
> > 
> >   DB<4> m RenameVariable
> > FETCH
> > STORE
> > via Tie::Scalar: TIESCALAR
> > via Tie::Scalar: carp
> > via Tie::Scalar: confess
> > via Tie::Scalar: croak
> > via Tie::Scalar: new
> > via UNIVERSAL: DOES
> > via UNIVERSAL: VERSION
> > via UNIVERSAL: can
> > via UNIVERSAL: isa
> >   DB<5> tie $DB::slave_editor, 'RenameVariable'
> > RenameVariable must define either a TIESCALAR() or a new() method at (eval 12)[lib/perl5db.pl:741] line 2.
> > 
> >   DB<6> sub RenameVariable::TIESCALAR { goto &Tie::Scalar::TIESCALAR }
> > 
> >   DB<7> tie $DB::slave_editor, 'RenameVariable'
> > RenameVariable must define either a TIESCALAR() or a new() method at (eval 14)[lib/perl5db.pl:741] line 2.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think probably I'm just being thick, but I don't have the time right now to work through my brain haze.
> 
> As I reported in #18868, no matter how correct we get the code for a tied variable inside `lib/perl5db.pl`, we're treading dangerously if we don't exercise the code for it in `lib/perl5db.t`. I've tried to figure out how to test what I'll call the "graphical debugger" block; "brain haze" is an accurate description of how far I've got.
> 
> So, in the short run we might be better off simply making _only_ a documentation change as you suggested in [#18854 (comment)](https://github.com/Perl/perl5/pull/18854#issuecomment-855528948) above. Would you agree?
> 
> Thank you very much.
> Jim Keenan

@rjbs (or PSC) any thoughts on the above?

Thank you very much.
Jim Keenan

khwilliamson:
@rjbs, can we get a resolution on this?

rjbs:
It seems maybe this was waiting on me?
> So, in the short run we might be better off simply making only a documentation change as you suggested in #18854 (comment) above. Would you agree?

Fine with me.

jkeenan:
This p.r. is being closed in favor of https://github.com/Perl/perl5/pull/19227

PR #18859, opened 2021-06-05
Use HOP3() instead of a call to utf8_hop_forward() in Perl_regexec_flags.

nwc10:
The regex engine should be using reghop3() instead of utf8_hop_forward().

Making that replacement reveals that the surrounding ternary is actually the same code as the HOP3() macro, so use that directly to simplify the code.


hvds:
> The regex engine should be using reghop3() instead of utf8_hop_forward().

How do you know? I don't see anything stating that ...

I also don't see any docs explaining why this family of functions exists as something regexp-specific - it appears to be historical, because until the nicely millenial 1aa99e6b6d the reghop* functions hardcoded `PL_regeol` rather than taking a `strend`, but since then they don't appear to have any regexp-specific aspect.

Given that particular call was added by @khwilliamson I wonder whether it was done deliberately to assert `minlenret >= 0` (but I imagine not).

khwilliamson:
I know because I recently tried to unify the twisty little mazes of hops into fewer functions, and it didn't work.  The reason turns out to be that the ones in inline.h, when asked to go off the edge, leave the pointer pointing to the edge.  The ones in regexec.c leave the pointer pointing at the first byte of the closest character to the corresponding edge.  That is, they leave the pointer at a legal position, whereas the inline.h ones may leave it in the middle of a multi-byte character.

@tonycoz and I chatted on irc about this just before the weekend, and we decided perhaps this github issue would be a suitable vehicle to bring this discrepancy to a more public forum and resolution.

I was surprised at the behavior of the inline.h versions, and even more surprised that it wasn't documented.  Tony isn't adverse to changing it, but they are API.  (But again this aspect of how they work isn't documented.)  If someone knows of a situation where the current behavior is preferable, s/he should speak up.

khwilliamson (approved):

tonycoz:
> The ones in regexec.c leave the pointer pointing at the first byte of the closest character to the corresponding edge. That is, they leave the pointer at a legal position, whereas the inline.h ones may leave it in the middle of a multi-byte character.

I wrote them expecting them to be usable in loops like:
```
s = SvPV(sv, len);
e = s + len;
while (s < e) { /* or s != e */
  /* do something with bytes at s */
  ...
  s = utf8_hop_forward(s, 1, e);
}
```

You said at a legal position - could that be before e? after e?  The first would break code like the above.  The second could produce undefined behaviour (as Perl_utf8_length() can if the supplied buffer contains invalid extended UTF-8)

khwilliamson:
I thought I had replied to this, but apparently didn't complete the process

The hop functions shouldn't go outside the bounds, and shouldn't leave the pointer at anything but the start byte of a character.  That means that the API is insufficient for loops like you illustrate.  There needs to be a mechanism to know that something went awry

nwc10:
Dear github, you are really stupid. I moved my branches around, and now you are confused and think that I merged this.

PR #18861, opened 2021-06-05
Handle platforms that don't allow specifying non-int struct bitfields

khwilliamson:


craigberry (commented):
The try.c program from Configure compiles with no warnings or errors on VMS, so I guess that means it does support bitfield structs of varying sizes.

khwilliamson:
Commited by bfe5127121a2046fe9d996a64635f6f08d940528

PR #18869, opened 2021-06-08
Small perly cleanups

happy-barney:
- empty rule to replace `%empty { $$ = NULL; }`
- employ default rule in optional rules
- use optional rule instead of multiple rules in `anonymous`


khwilliamson:
Given how long this author's previous PRs have sat unresponded to, and that complex ones of his worked without problems, I propose a 2 week interval, and if there are no objections within that time, I'll merge it

toke.c
```diff
@@ -2054,7 +2055,6 @@ static int
 S_postderef(pTHX_ int const funny, char const next)
 {
     assert(funny == DOLSHARP
-        || memCHRs("$@%&*", funny)
```
leonerd, line 2131: This removal seems interesting - Do we have any historical insight into why the test is done both with `memCHRs` and with the `==` tests below it? They seem to be duplicates

leonerd:
This mostly looks fine to me, aside from the comment about the `memCHRs()` call at the very end.

toke.c
happy-barney: Historically perly tokens where characters. Recently I replaced those used in perly with `PERLY_*` (to be able to distinguish between actual character literal and return value in toke). I did it one by one and after last one I just forgot to remove now obsoleted test.

iabyn:
Just a general comment about the commit subject and messages - they share the same 'namespace' as all commits and it's not clear what part of the perl distribution the apply to. So for example, I'd suggest changing 'Unify optional rules' to 'Parser: unify optional rules' or similar. Then in the main body of the commit message, make it clear that this is just a cosmetic change which should still generate the same parser.

karenetheridge:
Does perldelta need to be updated for any of these changes?

khwilliamson:
@happy-barney I think this could be merged if you rebased it

Issue #18844, opened 2021-05-31
While configuring Perl, I was asked if my system supports the C.UTF8 locale, should it not be asking about the C.UTF-8 locale?


SteelBlueVision:
Module:

**Description**

**Steps to Reproduce**

**Expected behavior**

**Perl configuration**

```
# perl -V output goes here
```

jkeenan:
Although the original poster supplied no information in the body of the ticket, I believe the question posed in the Subject is valid.  It appears that when running `./Configure` interactively, there are certain prompts on STDOUT which ask the user something about `C.UTF8` where it is reasonable to expect a question about `C.UTF-8`.

**Internally** -- that is, within `./Configure` and what we write to `config.sh` -- we generally delete the hyphen before the `8`.  These user-facing messages seem to be a leakage of that custom out into user-space.

Hence, I believe we can correct what the user sees when running `./Configure` interactively without damaging the information that program detects and stores.  Something like the following:
```
diff --git a/Configure b/Configure
index 6f8ae66f61..c3966b2092 100755
--- a/Configure
+++ b/Configure
@@ -18063,7 +18063,7 @@ case $d_setlocale in
            echo "Your system has setlocale()..." >&4
            $run ./try
            case $? in
-               0) echo "and it seems sane; you don't have a C.UTF8 locale" >&4
+               0) echo "and it seems sane; you don't have a C.UTF-8 locale" >&4
                   d_setlocale="$define"
                   d_setlocale_accepts_any_locale_name="$undef"
                   d_has_C_UTF8="false"
@@ -18073,7 +18073,7 @@ case $d_setlocale in
                   d_setlocale_accepts_any_locale_name="$define"
                   d_has_C_UTF8="false"
                   ;;
-               2) echo "and it seems sane; you have a C.UTF8 locale" >&4
+               2) echo "and it seems sane; you have a C.UTF-8 locale" >&4
                   d_setlocale="$define"
                   d_setlocale_accepts_any_locale_name="$undef"
                   d_has_C_UTF8="true"
@@ -18105,7 +18105,7 @@ $rm_try
         $undef) echo "There may be other ways to set the locale on your system, so we need to ask:" >&4
         ;;
     esac
-    rp="Does your system have the C.UTF8 locale?"
+    rp="Does your system have the C.UTF-8 locale?"
     dflt=n
     . ./myread
     case "$ans" in
@@ -20824,7 +20824,7 @@ if eval $compile_ok; then
 	fi
 else
[...]

PR #18870, opened 2021-06-08
Disambiguate what user sees in interactive Configure

jkeenan:
User should expect to see configuration questions spelled 'UTF-8', not
'UTF8' (even if we store information internally or in config.sh without
the hyphen).

Correct user prompts accordingly.

For: https://github.com/Perl/perl5/issues/18844

### DIFF
--- a/Configure
+++ b/Configure
@@ -18063,7 +18063,7 @@ case $d_setlocale in
            echo "Your system has setlocale()..." >&4
            $run ./try
            case $? in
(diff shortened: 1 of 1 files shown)


khwilliamson (approved):

khwilliamson:
But this needs to go into metaconfig

doughera88:
I'm afraid I don't have any useful insights on this.
-- 
    Andy Dougherty		***@***.***

Tux:
Backported, regenerated and pushed to both meta and blead

PR #18878, opened 2021-06-11
utf8_length: Fix undefined C behavior

khwilliamson:
In C the comparison of two pointers is only legal if both point to
within the same object, or to a virtual element one above the high edge
of the object.

The previous code was doing an addition potentially outside that range,
and so the results would be undefined.

### DIFF
--- a/utf8.c
+++ b/utf8.c
@@ -2372,22 +2372,29 @@ Perl_utf8_length(pTHX_ const U8 *s, const U8 *e)
      * the bitops (especially ~) can create illegal UTF-8.
      * In other words: in Perl UTF-8 is not just for Unicode. */
 
-    if (UNLIKELY(e < s))
-        goto warn_and_return;
     while (s < e) {
-        s += UTF8SKIP(s);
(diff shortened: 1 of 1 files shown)


utf8.c
```diff
@@ -2372,23 +2372,31 @@ Perl_utf8_length(pTHX_ const U8 *s, const U8 *e)
...
-        s += UTF8SKIP(s);
+        Ptrdiff_t expected_byte_count = UTF8SKIP(s);
+
+        if (UNLIKELY(e - s  < expected_byte_count)) {
+            goto warn_and_return;
+        }
+
         len++;
+        s += expected_byte_count;
+        expected_byte_count = UTF8SKIP(s);
```
tonycoz: This line isn't needed.

tonycoz (changes requested):

utf8.c
tonycoz: urr, the line selection was a bit off, that assignment to expected_byte_count isn't needed

khwilliamson:
That line was a relic from a previous incarnation

tonycoz (approved):

Issue #18780, opened 2021-05-10
clang12: Using as compiler generates 30K warnings of `-Wcompound-token-split-by-macro`


jkeenan:
Today I had occasion to try to build perl on FreeBSD-12 with, for the first time clang-12.  Using this version of clang introduced over 30,000 new build-time warnings, all of type `-Wcompound-token-split-by-macro`.
```
$ uname -mrs
FreeBSD 12.2-STABLE amd64

$ ./perl -Ilib -V:config_args
config_args='-des -Dusedevel -Dusethreads -Dcc=clang12';

$ clang12 --version
clang version 12.0.0
Target: x86_64-portbld-freebsd12.2
Thread model: posix
InstalledDir: /usr/local/llvm12/bin

$ ~/bin/perl/report-build-warnings 1e43c13e1f.freebsd.clang12.maketp.output.txt.gz
 1e43c13e1f.freebsd.clang12.maketp.output.txt.gz

  Wcompound-token-split-by-macro           30106
  Wstring-compare                            2

$ zgrep "source =>" 1e43c13e1f.freebsd.clang12.parse.output.txt.gz | sort | uniq -c
  16     source => "../../../hv_func.h",
 144     source => "../../../inline.h",
 208     source => "../../../perl_siphash.h",
  96     source => "../../../sbox32_hash.h",
 272     source => "../../../zaphod32_hash.h",
 122     source => "../../hv_func.h",
1124     source => "../../inline.h",
1586     source => "../../perl_siphash.h",
 732     source => "../../sbox32_hash.h",
2074     source => "../../zaphod32_hash.h",
  36     source => "./const-c.inc",
  12     source => "./const-xs.inc",
  26     source => "./constants.xs",
  86     source => "./hv_func.h",
 774     source => "./inline.h",
  10     source => "./invlist_inline.h",
   6     source => "./numeric.xs",
1118     source => "./perl_siphash.h",
 516     source => "./sbox32_hash.h",
  16     source => "./vutil.c",
  32     source => "./vxs.inc",
1462     source => "./zaphod32_hash.h",
1180     source => "APItest.c",
 506     source => "APItest.xs",
 198     source => "B.c",
 140     source => "B.xs",
  20     source => "Base64.c",
   8     source => "Base64.xs",
  10     source => "Byte.c",
  90     source => "Bzip2.c",
  52     source => "Bzip2.xs",
  10     source => "CN.c",
  10     source => "Call.c",
  16     source => "Call.xs",
[...]

PR #18879, opened 2021-06-11
Fix some build issues with clang in Mingw-w64

Biswa96:


jkeenan:
Thank you for this pull request.  I suspect that this p.r. may be attempting to do too much and that p.r.s addressing separate problems may be better.

Specifically, until we can see the build-time warnings you are getting when compiling with clang on Mingw-w64, we can't evaluate your proposed changes to `perl.h`.  Because `perl.h` is very fundamental to compilation, we have to be very careful before changing it.

We keep track of build-time warnings with a GitHub "label" by that name.  Can you please open a separate Issue for the build-time warnings, pasting in the relevant parts of the `make` (or Windows equivalent) output and the relevant configuration options?

Thank you very much.
Jim Keenan

Biswa96:
I think you've already opened the issue here https://github.com/Perl/perl5/issues/18780

jkeenan:
> I think you've already opened the issue here #18780

As noted in https://github.com/Perl/perl5/issues/18780#issuecomment-878634611 today, I tried out your pull request and found that it eliminated the vast majority of build-time warnings discussed in #18780.  The warnings it does not eliminate appear to come from files under `dist/` rather than from `perl.h`.

I'm not in a position to evaluate the mingw-related parts of this pull request.  Nor can I say whether this is the *best* way to eliminate approximately 28,000 build-time warnings; only that is *a* way to eliminate such warnings.  I don't have access to more recent versions of `clang` (clang13, I think, is relevant).

So a qualified +1 to this p.r.

jkeenan:
@Biswa96, can you confirm that the changes in [this commit](https://github.com/Perl/perl5/commit/88e366d42b36d12b0fed022ac5209f0151150740) are consistent with your proposed changes to `perl.h` and help to eliminate build-time warnings when compiling with clang12?

Reference: https://github.com/Perl/perl5/issues/18780

Thank you very much.
Jim Keenan

jkeenan:
> > I think you've already opened the issue here #18780
> 
> As noted in [#18780 (comment)](https://github.com/Perl/perl5/issues/18780#issuecomment-878634611) today, I tried out your pull request and found that it eliminated the vast majority of build-time warnings discussed in #18780. The warnings it does not eliminate appear to come from files under `dist/` rather than from `perl.h`.
> 
> I'm not in a position to evaluate the mingw-related parts of this pull request. Nor can I say whether this is the _best_ way to eliminate approximately 28,000 build-time warnings; only that is _a_ way to eliminate such warnings. I don't have access to more recent versions of `clang` (clang13, I think, is relevant).
> 
> So a qualified +1 to this p.r.

@xenu, are you in a position to review the mingw-related parts of this p.r.?

Thank you very much.
Jim Keenan

jkeenan:
@Biswa96,

Thank you for providing this pull request, which helped focus discussion of the problem of build-time warnings with clang12.

In https://github.com/Perl/perl5/pull/18984, @tonycoz has provided a more focused version of your patch, which we have applied to blead.

That leaves the 3 mingw-focused commits which were part of your p.r.  If you would like to pursue those further, please open a new p.r. with just mingw-focused commits.

Closing this pull request.  Thank you very much.

Biswa96:
Thank you for the fix. It will take time to test perl in mingw with blead branch because it needs [some patches][1] to adapt mingw style path. Also, if compiled with mingw clang perl crashes after compiling miniperl.exe.

[1]: https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-perl

PR #18880, opened 2021-06-11
perlop - Invocant only needs to be an object or class name for method calls

Grinnz:
The restriction on the left side mentioned here only applies when the right side is a method name, not when it is a subroutine reference.


jkeenan:
> The restriction on the left side mentioned here only applies when the right side is a method name, not when it is a subroutine reference.

+1.  Other reviewers?

Issue #18732, opened 2021-04-20
Porting/corelist.pl should use the official CPAN mirrors


toddr:
It's my understanding that we're moving away from ad-hoc CPAN mirrors. As a result, we need to adjust internal perl code (mostly in Porting/) to use those mirrors exclusively.

Seen running `./perl -Ilib Porting/corelist.pl cpan`:

```
About to fetch 02packages from ftp.funet.fi. This may take a few minutes
```

CC @rjbs

PR #18883, opened 2021-06-12
to check the CPAN, use www.cpan.org

rjbs:
The mirror list is no longer an ongoing concern.  We should rely on the CDN and stop telling users about the list list of mirrors.


rjbs:
I believe this fixes #18732

neilb (approved):
LGTM

PR #18885, opened 2021-06-13
perlgov: re-alphabetize the name list

rjbs:
This just helps with the currently-manual process of setting up voting invites.


xdg (approved):
LGTM

PR #18888, opened 2021-06-15
Fix final-expression semantics for multiple statements in `catch` (fixes #18855)

### LINKED ISSUE #18855: Producing a default value from a do/try/catch block combination fails
**Description**

Even when following the exact code from the 5.34 [perlsyn docs](https://metacpan.org/pod/release/XSAWYERX/perl-5.34.0/pod/perlsyn.pod#Try-Catch-Exception-Handling#highlighter_507101) regarding producing a default value with a try/catch block wrapped in a "do" block, the code fails to produce the value.

**Steps to Reproduce**
```
#!/usr/bin/env perl

use 5.034;
use feature 'try';

my $result = do {
  try {
    1 / 0;
  } catch ($e) {
    warn "Error: $e";
    1;
  }
};

say "'$result'";
```

**Expected behavior**
It outputs:
```
Error: Illegal division by zero at try_catch.pl line 9.
''
```

It should output:
```
Error: Illegal division by zero at try_catch.pl line 9.
'1'
```
EDIT: If you remove/comment the `warn` line, it works as expected.

**Perl configuration**
```
Summary of my perl5 (revision 5 version 34 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=5.11.19-1-manjaro
    archname=x86_64-linux
    uname='linux x1n-manjaro 5.11.19-1-manjaro #1 smp preempt fri may 7 17:34:25 utc 2021 x86_64 gnulinux '
    config_args='-de -Dprefix=/home/user/perl5/perlbrew/perls/perl-5.34.0 -Aeval:scriptdir=/home/user/perl5/perlbrew/perls/perl-5.34.0/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='cc'
    optimize='-O2'
    ccversion=''
    gccversion='10.2.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    libpth=/usr/local/lib /usr/lib
    libc=libc-2.33.so
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.33'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'


  Compile-time options:
    HAS_TIMES
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under linux
  Compiled at Jun  2 2021 12:03:01
  %ENV:
    PERLBREW_HOME="/home/user/.perlbrew"
    PERLBREW_MANPATH="/home/user/perl5/perlbrew/perls/perl-5.34.0/man"
    PERLBREW_PATH="/home/user/perl5/perlbrew/bin:/home/user/perl5/perlbrew/perls/perl-5.34.0/bin"
    PERLBREW_PERL="perl-5.34.0"
    PERLBREW_ROOT="/home/user/perl5/perlbrew"
    PERLBREW_SHELLRC_VERSION="0.91"
    PERLBREW_VERSION="0.91"
  @INC:
    /home/user/perl5/perlbrew/perls/perl-5.34.0/lib/site_perl/5.34.0/x86_64-linux
    /home/user/perl5/perlbrew/perls/perl-5.34.0/lib/site_perl/5.34.0
    /home/user/perl5/perlbrew/perls/perl-5.34.0/lib/5.34.0/x86_64-linux
    /home/user/perl5/perlbrew/perls/perl-5.34.0/lib/5.34.0
```

leonerd:
Bugfix for #18855

Amazingly enough, the fix for `voidnonfinal()` breaking context of `OP_ENTER` doesn't seem to cause any unit-test collateral damage. We might want to test a bit of CPAN as well on this.

### DIFF
--- a/op.c
+++ b/op.c
@@ -2622,8 +2622,13 @@ S_voidnonfinal(pTHX_ OP *o)
         if (type == OP_LINESEQ || type == OP_SCOPE ||
             type == OP_LEAVE || type == OP_LEAVETRY)
         {
-            OP *kid, *sib;
-            for (kid = cLISTOPo->op_first; kid; kid = sib) {
+            OP *kid = cLISTOPo->op_first, *sib;
+            if(type == OP_LEAVE) {
+                /* Don't put the OP_ENTER in void context */
+                assert(kid->op_type == OP_ENTER);
+                kid = OpSIBLING(kid);
+            }
+            for (; kid; kid = sib) {
                 if ((sib = OpSIBLING(kid))
                  && (  OpHAS_SIBLING(sib) || sib->op_type != OP_NULL
                     || (  sib->op_targ != OP_NEXTSTATE
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -232,6 +232,14 @@ no warnings 'experimental::try';
         catch ($e) { 4, 5, 6 }
     };
     ok(eq_array(\@list, [1, 2, 3]), 'do { try } in list context');
+
+    # Regression test related to
+    #   https://github.com/Perl/perl5/issues/18855
+    $scalar = do {
+        try { my $x = 123; 456 }
+        catch ($e) { 789 }
+    };
+    is($scalar, 456, 'do { try } with multiple statements');
 }
 
 # catch as final expression yields correct value
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -247,6 +247,14 @@ no warnings 'experimental::try';
         catch ($e) { 4, 5, 6 }
     };
     ok(eq_array(\@list, [4, 5, 6]), 'do { try/catch } in list context');
+
+    # Regression test 
+    #   https://github.com/Perl/perl5/issues/18855
+    $scalar = do {
+        try { die "Oops" }
+        catch ($e) { my $x = 123; "result" }
+    };
+    is($scalar, "result", 'do { try/catch } with multiple statements');
 }
 
 # try{} blocks should be invisible to caller()


leonerd:
> We might want to test a bit of CPAN as well on this.

I tested all of my syntax modules (`Future::AsyncAwait`, `List::Keywords`, `Object::Pad` and each of `Syntax::Keyword::{Defer,Dynamically,Match,Try}`); all are just fine.

In addition, `Syntax::Keyword::Try` didn't suffer the original bug like #18855 anyway - so once this is fixed for 5.34.1 (hopefully), I can add it to the test suite for `Feature::Compat::Try`.

wolfsage:
I added a warning to scalarseq to see if it's ever hit with an OP_LEAVE and I could not hit it testing a decent amount of cpan. This change makes me nervous, but maybe scalarseq never happens on an OP_LEAVE except with try/catch.

It was introduced with 79072805bf63abe5b5978b5928ab00d360ea3e7f (perl 5.0 alpha 2!) and maybe it made sense back then but now doesn't? I'm really not sure.

Maybe a good candidate for smoking cpan with if we can do that? @rjbs was playing around with some stuff to do this at least.

If we did smoke cpan with it, I'd want to add a warning to somehow let us know if scalarseq on an OP_LEAVE is ever actually hit in the wild (outside of try/catch)...

leonerd:
`scalarseq()` (now renamed `voidnonfinal()`; see #18891) isn't a documented/exposed API function, so there's no way any CPAN code could ever call it directly.

It's only visible indirectly via `newMYSUB()`, `newATTRSUB()`, `block_end()`, and `newFORM()` (yes really). In all my CPAN modules I've never managed to call `block_end()` with an OP_LEAVE toplevel, except the one time I tried to use that in S-K-T and found it broke nonvoid context, so I didn't. I didn't look into more detail on why it broke; I just avoided it. It turns out this is the reason why.

It may be that CPAN authors avoid it in this situation because it didn't work before this bugfix. It's also possible that CPAN authors are applying a workaround. You can fix it manually (and I almost did that here), by

```C
OP *block = block_end(...);
OP *blockenter = cUNOPx(block)->op_first;
blockenter->op_flags = (blockenter->op_flags & ~G_WANT) | (block->op_flags & G_WANT);
```

so maybe take a look around any likely CPAN modules' source for code of that sort of shape.

wolfsage:
Yes I was thinking more specifically of it being used through block_end, etc.

leonerd:
> Yes I was thinking more specifically of it being used through block_end, etc.

grep.cpanauthors.org reveals only 12 places on the whole of CPAN where anyone calls `block_end()`. 

https://grep.cpanauthors.org/search?q=block_end&extension=xs

Of those, I have already tested all of the ones authored by me. The only modules on CPAN where someone other than me calls `block_end()` are:

* ~~Moot (MOOCOW, last release 2018)~~ - *edit* Turns out this is a false hit; that's an XS function coincidentally named, but not the core one.
* List::Gather (ETHER, last release 2018)
* Function::Parameters (MAUKE, last release 2017)
* Switch::Plain (MAUKE, last release 2017)
* Cond::Expr (ETHER, last release 2018)
* ~~GCCJIT (VYF, last release 2016)~~ - *edit* Also a false hit
* Scope::Escape::Sugar (ZEFRAM, last release 2011)

Those remaining 5 cases should be quite easy to iterate through and inspect for possibility of OP_LEAVE to block_end.

leonerd:
### List::Gather
The only callsite is on an op sequence constructed by `op_append_elem(OP_LINESEQ, ...)` and thus must be an OP_LINESEQ, and not a LEAVE.

### Function::Parameters
Same; always OP_LINESEQ.

### Switch::Plain
Has two callsites; one is in the standard `block_start/parse_block/block_end` sequence (and thus will be an OP_LINESEQ, much like all of core perl); the other is - you guessed it - generated by an `op_append_list(OP_LINESEQ, ...)` call.

### Cond::Expr
Only callsite is `block_start/parse_block/block_end`.

### Scope::Escape::Sugar
Would you believe it - it's another `op_prepend_elem(OP_LINESEQ, ...)`.

I think we're fine ;)

wolfsage (approved):
I had checked List::Gather at least, but I've checked those others you found. (grep.metacpan.org failed me so I couldn't find a good list). Wouldn't mind some good smoking before releasing this in 5.34.1 but that's just me.

tonycoz:
This works, and I think it is a solution.

Looking at the other calls to block_end() (which calls scalarseq()), they all op_scope() the block, then wrap that with whatever op (newCONDOP(), newWHILEOP()) and only then call block_end() on the result, so in general scalarseq() doesn't get to mark OP_ENTER as VOID.

One difference between try{} and these other ops is they include code at the top which may introduce lexical variables, or just plain declarations:

  while (CODE) { ... }
  for my $x (...) { ... }
   if (CODE) { ... }

which try doesn't.

try does include a lexical variable in the catch() and the block_end() does include that.  It's possible the ENTER/LEAVE from op_scope() should wrap the catch op instead of just the block, but I wasn't able to produce any misbehaviour from that.

oschwald:
Unfortunately, this doesn't seem to be in 5.34.1-RC1.

Issue #18894, opened 2021-06-15
Clean-up Pod-Html:  a battle plan


jkeenan:
In this ticket I describe a plan to refactor code under `ext/Pod-Html` in the Perl 5 core distribution.

**Rationale**

When, up until October 2019, we used rt.perl.org as our bug tracker, we used to group tickets by creating "Meta" tickets.  Around 2012 Ricardo Signes created one such Meta ticket called [Pod::Html / pod2html is a mess](https://rt.perl.org/Ticket/Display.html?id=112014) [GH 12019](https://github.com/Perl/perl5/issues/12019).  It was then and still is.  The purpose of this refactoring is to significantly reduce the amount of that mess.

There are two major aspects to this messiness.

First, the code in `ext/Pod-Html/lib/Pod/Html.pm` is not fully encapsulated.  State is held in a total of 19 lexical variables declared outside the scope of the main function, `pod2html()`.

Second, the tests in `ext/Pod-Html/t/*.t` cannot be run in parallel.  That's because the test suite is currently set up to create temporary files underneath `ext/Pod-Html` rather than true tempdirs.  This leads to race conditions.  We currently have a workaround in `t/harness` to require the tests in the `ext/Pod-Html/t` to be run one at a time.

**Overall Game Plan**

The two aspects of Pod-Html's messiness should be tackled in reverse order.

First, revise the test suite so that the tests can be run in parallel.

Then encapsulate the code in the main module.  Instead of state being held in lexical variables, hold it in a Pod::Html object.  Read POD and user configuration choices into that object; parse it within that object; compose HTML within that object; write HTML files from within that object.

We won't attempt to fix every existing deficiency in Pod-Html in this project, but we'll be in a much better position to accept pull requests for those deficiencies once the code and tests are cleaner.

Furthermore, in this project we are not attempting to change what Pod-Html actually does.  More specifically, accept for correction of bugs the `pod2html` utility should work exactly the same after this cleanup as before.  But, once again, we'll be in a better position to change the way that utility works going forward.

**Recommendations**

When you are refactoring messy code you are likely to end up with an enormous diff even if, in the end, the functional changes you are making are few.  That will be the case here.  Hence I plan to present this project as a series of **five** pull requests corresponding to five nested branches in my own fork of Perl 5 on github.  The pull requests will implement the following "order of battle":

[...]

PR #18895, opened 2021-06-15
Pod-Html: have tests run in individual tempdirs

jkeenan:
This is the first of five pull requests to be submitted in support of https://github.com/Perl/perl5/issues/18894.

Until now, the tests for this library ran inside the `ext/Pod-Html`
directory in the core distribution. Since the tests necessarily create
files while the harness runs, there was a possibility for race
conditions, which meant that we had to have a workaround in t/harness to
have the tests in this directory not run in parallel.

This commit restructures the testing process so that each test program
gets its own tempdir in which to create files and directories.

The test architecture is now found in `t/lib/Testing.pm`.
This module will export two functions on demand that are used in the
individual test files. They guarantee that all files created when
running a test program will be created in a temporary directory rather
than underneath the core distribution.

`t/pod2html-lib.pl` is now superseded by `t/lib/Testing.pm` and can be
removed. The workaround has been removed from `t/harness`.

This should help in the resolution of https://github.com/Perl/perl5/issues/12271.

Thank you very much.
Jim Keenan


karenetheridge:
Can File::Temp be used to create the tempfiles + tempdirs, or does that module depend on this one? But since they're both in core (since 5.6.1), a circular dependency should still be fine - both will be available via core at some version and either one can be upgraded from cpan with a dependency on the original core version.

jkeenan:
On 6/20/21 7:32 PM, Karen Etheridge wrote:
> Can File::Temp be used to create the tempfiles + tempdirs, or does that 
> module depend on this one? 

File::Temp does not depend on Pod::Html.

jkeenan:
If there are no further comments I will merge this branch into blead within 48 hours, then submit the next pull request in this series.

Thank you very much.
Jim Keenan

jkeenan:
I squashed the two commits in this branch locally before merging to blead, testing, then pushing to origin as merge commit ddf86ab3dc887706544973328ece2259cc4f90bd and code commit 8433f82996ead276b305b5c8e4314d9295e08bbe.

Closing p.r.

Thank you very much.
Jim Keenan

PR #18896, opened 2021-06-16
More tests for new feature try

thibaultduponchelle:
Hello,

This is some more tests for the try/catch feature 😀 

I tried to stress a bit the feature with some (kind of) real use cases (signal, eval, propagation).

### DIFF
--- a/t/op/try.t
+++ b/t/op/try.t
@@ -74,6 +74,95 @@ no warnings 'experimental::try';
     is(f(), "return inside try", 'return inside try');
 }
 
+# eval next/last inside try
+{
+    no warnings "exiting";
+    my $result = "";
+    try {
+        {
+            $result .= "a";
+            eval "next";
+            $result .= "b";
+        }
+    } catch ($e) {
+        $result .= "c";
+    }
+
+    is($result, "a", 'eval "next" (inner block) inside try');
+
+    $result = "";
+    {
+        try {
+            $result .= "d";
+            eval "last";
+            $result .= "e";
+        } catch ($e) {
+            $result .= "f";
+        }
+        $result .= "skipped";
+    }
+
+    is($result, "d", 'eval "last" (outer block) inside try');
+}
+
+# eval a try
+{
+    my $result = "";
+    eval 'try { $result = "try"; die "dying"; } catch($e) { $result .= " and catch"; }';
+
+    is($result, "try and catch", 'eval a try');
+}
+
+# eval a try that eval a try
+{
+    my $result = "";
+    my $eval_a_try = 'try { $result .= " and try inside"; die "dying"; } catch($e) { $result .= " and catch inside"; }';
+    eval 'try { $result .= "try outside"; eval "$eval_a_try"; die "dying"; } catch($e) { $result .= " and catch ouside"; }';
+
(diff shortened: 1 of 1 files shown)


thibaultduponchelle:
@leonerd 😃

jkeenan:
> Hello,
> 
> This is some more tests for the try/catch feature grinning
> 
> I tried to stress a bit the feature with some (kind of) real use cases (signal, eval, propagation).

@leonerd, can you review this p.r.?  I'm inclined to apply it within 7 days unless you indicate otherwise.

t/op/try.t
```diff
@@ -74,6 +74,88 @@ no warnings 'experimental::try';
     is(f(), "return inside try", 'return inside try');
 }
 
+no warnings "exiting";
```
leonerd: Move this inside the scope brackets of the test it applies to

t/op/try.t
```diff
@@ -74,6 +74,88 @@ no warnings 'experimental::try';
...
+        try {
+            $result .= "d";
+            eval "last";
+            $result .= "e";
+        } catch ($e) {
+            $result .= "f";
+        }
+    }
+
+    is($result, "d", 'eval last block outside try');
```
leonerd: ITYM "inside try"

leonerd (changes requested):
Needs a couple of small changes but otherwise looks good.

t/op/try.t
haarg: The block that the last is terminating is outside the try.

jkeenan:
@thibaultduponchelle , could you update your pull request based on the feedback ^^ from @leonerd and @haarg ?  Thanks.

t/op/try.t
thibaultduponchelle: done

t/op/try.t
thibaultduponchelle: I edited the descriptions to make them clearer

thibaultduponchelle:
Thank you @leonerd and @haarg for the review (and thank you @jkeenan). I updated according to the remarks 👍🏼

jkeenan:
> Thank you @leonerd and @haarg for the review (and thank you @jkeenan). I updated according to the remarks 👍🏼

@leonerd and @haarg, can you review the contributor's most recent changes in this pull request?  Thanks.

leonerd:
Overall no huge objection, though in parts I'm unsure quite the value being added by the tests. While "more tests are always good", there gets to a point where we're just testing many minor variations on the same idea, which of course adds size and time.

The signal one at the end, in particular, doesn't actually test signal propagation *through* a `try` boundary, only entirely within on.

jkeenan:
> 
> The signal one at the end, in particular, doesn't actually test signal propagation _through_ a `try` boundary, only entirely within on.

@thibaultduponchelle, can you respond to this point raised by @leonerd ?

Thank you very much.
Jim Keenan

thibaultduponchelle:
@jkeenan I’m going to do the change 👍🏼

thibaultduponchelle:
@jkeenan @leonerd I moved SIGINT handler out of `try {}` block and added an outer and inner block to make it clearer that correct `catch` is reached.

PR #18898, opened 2021-06-17
perlio.c: Add comments

khwilliamson:
The code is using '#ifdef EBCDIC' when it has nothing to do with EBCDIC;
but is a proxy for how things work on IBM boxes that use EBCDIC.  Note
this in the comments.

### DIFF
--- a/perlio.c
+++ b/perlio.c
@@ -1991,7 +1991,7 @@ PerlIOBase_pushed(pTHX_ PerlIO *f, const char *mode, SV *arg, PerlIO_funcs *tab)
             SETERRNO(EINVAL, LIB_INVARG);
             return -1;
         }
-#ifdef EBCDIC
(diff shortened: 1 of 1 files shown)


xenu:
Why can't we just change them to `#ifdef __MVS__`? According to IBM's [documentation](https://www.ibm.com/docs/en/zos/2.4.0?topic=macros-related-platform) that's the macro we should be using to identify z/OS.

khwilliamson:
On 6/17/21 12:22 AM, xenu wrote:
> Why can't we just change them to |#ifdef __MVS__|? According to IBM's 
> documentation 
> <https://www.ibm.com/docs/en/zos/2.4.0?topic=macros-related-platform> 
> that's the macro we should be using to identify z/OS.
> 

That seems to work, so I'll do that instead.  Thanks

Leont (approved):

PR #18899, opened 2021-06-17
Fix some links

thibaultduponchelle:


cpan/Encode/Unicode/Unicode.pm
```diff
@@ -259,7 +259,7 @@ Consider that "division by zero" of Encode :)
 =head1 SEE ALSO
 
 L<Encode>, L<Encode::Unicode::UTF7>, L<http://www.unicode.org/glossary/>,
-L<http://www.unicode.org/unicode/faq/utf_bom.html>,
+L<http://www.unicode.org/faq/utf_bom.html>,
```
Grinnz: This file is maintained on CPAN: https://github.com/dankogai/p5-encode

cpan/Encode/t/Unicode.t
```diff
@@ -25,7 +25,7 @@ use Encode qw(encode decode find_encoding);
 
 #
 # see
-# http://www.unicode.org/unicode/reports/tr19/
+# http://www.unicode.org/reports/tr19/
```
Grinnz: This file is also maintained on CPAN, in the same distribution.

Grinnz (changes requested):

cpan/Encode/Unicode/Unicode.pm
thibaultduponchelle: thank you :) https://github.com/dankogai/p5-encode/pull/158

cpan/Encode/t/Unicode.t
thibaultduponchelle: Same https://github.com/dankogai/p5-encode/pull/158

thibaultduponchelle:
There is a test failure that I need to understand/fix

Grinnz:
I believe that is due to the changes to the txt files in lib/unicore. Some files need to be regenerated. `regen/mk_invlists.pl` and `regen/regcharclass.pl` are the scripts to regen them.

Also, version in lib/Unicode/UCD.pm needs to be bumped, and the changes to the files in cpan/ need to be removed.

thibaultduponchelle:
Yes ! 😁 

Thank you a lot @Grinnz 

(I just updated this PR)

thibaultduponchelle:
Thank you 😃

khwilliamson:
The .txt files in lib/unicore are not perl's to change.  They are copyright by the Unicode Consortium, and permission is not granted to change them.  So this part of the PR should not have been merged.  That said, I don't see a compelling reason to revert.  We are technically in violation of the copyright, but I can't see that Unicode would object to our fixing their mistake.

Grinnz:
Can we submit those link fixes upstream somehow?

khwilliamson:
On 6/17/21 9:33 AM, Dan Book wrote:
> Can we submit those link fixes upstream somehow?
> 
> —

I have notified them of the problem.  Those files are immutable once 
published.  The next edition is due in September.  The drafts of them 
are also wrong.

The best solution would be for Unicode to fix their website to have 
legacy links retained (presuming the links were valid at one point)

thibaultduponchelle:
And there are now redirections...

PR #18903, opened 2021-06-17
local sub

happy-barney:
Consistency complement to "my sub"


happy-barney:
Based on #18869 but different PR in case of "is it perlish enough?" discussions

Grinnz:
What is the use case for this feature?

happy-barney:
@Grinnz primary purpose: consistency (`my sub` already exists)
it replace existing two expressions
```
local *foo = sub { }
```

with little bit easier to read
```
local sub foo { }
```

Usage in production code is unlikely, unlike usage is test code (eg: replaces necessity for `Sub::Override`)

Grinnz:
This is not necessary for consistency. `my` is a declaration, `local` is an operation.

I see the use case for local overrides, though, I'm not sure there's any problem with glob assignment for that purpose.

Leont:
I like it, IMO it's a far more intuitive syntax that glob assignment. Moving people away from having to understand typeglobs is good to me.

happy-barney:
Two follow-ups to this
- should localized function automatically inherit signature of localized function ?
- should localized function be able to address function it localized ?

(questions doesn't need to be addressed in this PR, answer is proposed in define meta operator RFC and subsequent COP)

Grinnz:
> * should localized function automatically inherit signature of localized function ?

This doesn't seem like a good idea.

> * should localized function be able to address function it localized ?

Consistent with current semantics, I think it would make sense if the sub's name within the `local sub` override resolves to the previously existing version of the sub. Also, could it even work any other way? `local` occurs at runtime.

> (questions doesn't need to be addressed in this PR, answer is proposed in define meta operator RFC and subsequent COP)

Can you clarify this? Is there an RFC planned for this feature proposal?

happy-barney:
> > * should localized function automatically inherit signature of localized function ?
> 
> This doesn't seem like a good idea.
> 
> > * should localized function be able to address function it localized ?
> 
> Consistent with current semantics, I think it would make sense if the sub's name within the `local sub` override resolves to the previously existing version of the sub. Also, could it even work any other way? `local` occurs at runtime.
> 
> > (questions doesn't need to be addressed in this PR, answer is proposed in define meta operator RFC and subsequent COP)
> 
> Can you clarify this? Is there an RFC planned for this feature proposal?

Example of syntax to address parent's declaration of function I'd like to use: `${: .. / &foo }->(...)`

RFC is planned, first part part of it (define meta operator) was already posted

nwc10:
This is a language level change - it needs to be an RFC.

Not a very *large* RFC, and the reason for adding it seems useful - a cleaner way to override subroutines, such as for testing. But it's **not** a bug fix.

As to the proposed implementation - the first thing I tried was:

```
$ ./perl -e 'local sub foo {warn}; foo'
$ ./perl -e 'sub foo {die}; local sub foo {warn}; foo'
Warning: something's wrong at -e line 1.
$
```

contrast with scalars:

```
$ ./perl -lwe 'local $foo = "Hello world"; print $foo'
Hello world
$ ./perl -lwe '$foo = "Woof!"; local $foo = "Hello world"; print $foo'
Hello world
$
```

This is inconsistent - the subroutine shouldn't need to exist to be localised.

Lexical subroutines exist. The interaction needs to be tested, and **not do this**:

```
$ ./perl -e 'my sub foo {die}; local sub foo {warn}; foo'
panic: unimplemented op padany (#12) called at -e line 1.
$
```

This works as I would expect, but isn't tested:

```
$ ./perl -e '{ sub foo {die}; local sub foo {warn}; sub bar { foo } } bar()'
Died at -e line 1.
```

The commit as **three** tests. Subroutines are complex with a messy implementation:

* Methods are subroutines, so these also need testing with `local`.
* `AUTOLOAD` and `DESTROY` are subroutines, so can be localised.
* There are internal caches related to CVs, which need to be correctly invalidated for both insertion and removal of a local override.
* overloaded methods are cached specially - these at least need testing
* `DESTROY` is cached specially - this needs checking
* Stub subroutines exist, which affect method lookup, particularly in the presence of overrides. For orthogonality, these ought to exist, and be localisable. *ie*
  * this should not be a syntax error
  * this should be tested, including with inheritance and `AUTOLOAD`
```
$ ./perl -e 'local sub foo; foo'
syntax error at -e line 1, near "sub foo;"
```

There need to be a lot more tests, even if they just validate that expected behaviour is what is observed.

Also

1) There is no documentation
2) This at least needs an experimental warning, if not a feature gate.

I think that the idea seems good, but the implementation isn't there yet.

I'm marking this PR as "draft", because the last commit makes it not suitable for merging. The preceding refactoring commits are viable for review (but I am not competent to do that) - I think that they are in #18869

jkeenan:
This pull request received criticism from two committers.  See:

https://github.com/Perl/perl5/pull/18903#issuecomment-863513196

https://github.com/Perl/perl5/pull/18903#issuecomment-863808112

It has been in Draft status for more than a year and has acquired merge conflicts.

If commenters are correct and this is the sort of language-level change that needs to go through the RFC process, then I recommend this p.r. be closed and a new one opened as needed in the future.

PR #18904, opened 2021-06-17
Put back the old url for unicode.org (in lib/unicore) since there is now a redirection

thibaultduponchelle:
Hello 😀 

I propose this *partial* revert of one commit of #18899 

Since @khwilliamson reported to unicode, there are now redirections on unicode.org 👍🏼 so all the links in lib/unicore could then reuse the initial url. I kept all non lib/unicore links set to the redirection target.

The generated hashes should be the same than before (e.g. in uni_keywords.h) I used the generators again and checked some hashes (that are well back to their old value). EDIT: I checked all hashes and it's good. 

@Grinnz @jkeenan


khwilliamson:
I think we should wait until they respond; or rather give them a few days to respond, to see if something was accidental or what

PR #18905, opened 2021-06-17
Add .gitattributes to tell github to treat perly.{act,h,tab} as generated files

leonerd:
This skips over them in diff views, for example, making code reviews simpler.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 .git_patch export-subst
+perly.act linguist-generated
+perly.h linguist-generated
+perly.tab linguist-generated


Grinnz:
There are probably other generated files that would be useful to add to this list but this is a good start.

Leont:
AFAICT, all of these files are generated:
 * `keywords.h`
 * `regcharclass.h`
 * `charclass_invlists.h`
 * `Porting/Glossary`
 * `embedvar.h`
 * `pod/perlintern.pod`
 * `pod/perltoc.pod`
 * `pod/perlmodlib.pod`
 * `pod/buildtoc`
 * `pod/perluniprops.pod`
 * `pod/perlapi.pod`
 * `mg_raw.h`
 * `perly.tab`
 * `perly.h`
 * `warnings.h`
 * `perly.act`
 * `embed.h`
 * `reentr.h`
 * `reentr.c`
 * `pp_proto.h`
 * `proto.h`
 * `overload.h`
 * `packsizetables.inc`
 * `regnodes.h`
 * `opcode.h`
 * `mg_names.inc`
 * `opnames.h`
 * `l1_char_class_tab.h`
 * `lib/B/Op_private.pm`
 * `lib/overload/numbers.pm`
 * `lib/unicore/uni_keywords.pl`
 * `lib/feature.pm`
 * `lib/warnings.pm`
 * `feature.h`
 * `keywords.c`
 * `overload.inc`
 * `mg_vtable.h`
 * `ebcdic_tables.h`
 * `unicode_constants.h`
 * `uni_keywords.h`
 * `miniperlmain.c`

khwilliamson:
On 6/18/21 7:51 AM, Leon Timmermans wrote:
> AFAICT, all of these files are generated:

Off the top of my head, these on the list are not regenerated

Leont:
> Off the top of my head, these on the list are not regenerated

Yeah, I had removed them once I realized that, but emails don't get updated.

nwc10:
> There are probably other generated files that would be useful to add to this list but this is a good start.

I thought this too, and then wondered why it stalled.

On the basis of "don't let the perfect be the enemy of the good" (or "this is better than what we had") I have applied it.

I wonder whether we can auto-generate `.gitattributes`. **Most** auto-generated files come via `regen/regen_lib.pl`, which writes a standard format "Generated from:" stanza in a comment at the end of the files with one or more sha256 hashes. So maybe adapt `regen.pl` to also generate `.gitattributes`. Or maybe just scan for that stanza brute-force. Yes, the line `.git_patch export-subst` is going to need a special case... :-)

If someone (else) wants to take this further, that would be cool. (Even just write up a suggested plan as an issue, and see if anyone (else) is keen to try it). But I don't have the free time to get to this.

PR #18909, opened 2021-06-20
Update os390 hints

khwilliamson:


jkeenan:
If anyone is using os390 and can evaluate Karl's p.r., please do so!

Thank you very much.
Jim Keenan

Tux (commented):
I really appreciate all the comments!
I didn't test it, but it looks ok to me.

OTOH it seems lik ethis change removes quite a lot of `-D`s. Are all of the deleted one's really not needed?

And yes, I know I am kinda late of my review. I was over 2000 mails behind, and I hate HTML mail

Issue #13198, opened 2013-08-25
[EXPERIMENT] the :win32 pseudolayer


p5pRT:
Migrated from [rt.perl.org#119453](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=119453) (status was 'resolved')

Searchable as RT119453$

p5pRT:
### From @rjbs

perlexperiment lists the :win32 pseudolayer as experimental\, since perl version
I\-don't\-know\-when\.

Its status was discussed a few months ago in this thread&#8203;:

&nbsp;  http&#8203;://www\.nntp\.perl\.org/group/perl\.perl5\.porters/2013/06/msg203089\.html

It looks like the experiment has not gone well\.  Maybe it can be completed and
maybe it can't\.

We should determine its state\, its potential eventual value\, and whether it
really can be gotten to work\.  If not\, we should slate it for deletion\.
Otherwise\, we should sort out steps to get it done\.

\-\- 
rjbs

p5pRT:
### From @rjbs

…at minimum\, we should also issue an experimental\-category warning when this layer is applied\, starting in 5\.21\.early\.

\-\- 
rjbs

p5pRT:
### From @tonycoz

On Wed Feb 26 16&#8203;:27&#8203;:20 2014\, rjbs wrote&#8203;:
> …at minimum\, we should also issue an experimental\-category warning
> when this layer is applied\, starting in 5\.21\.early\.

Here's a patch\.

Tony

p5pRT:
### From @tonycoz

<details><summary>0001-add-a-warning-for-using-the-win32-PerlIO-layer.patch</summary>

```diff
From aa7385ba368304520ab39c4730be9aadc5195e6f Mon Sep 17 00:00:00 2001
From: Tony Cook <tony@develop-help.com>
Date: Tue, 3 Jun 2014 16:56:40 +1000
Subject: [PATCH] add a warning for using the :win32 PerlIO layer

---
 lib/warnings.pm       |  258 +++++++++++++++++++++++++------------------------
 pod/perldiag.pod      |    8 ++
 regen/warnings.pl     |    4 +-
 t/lib/warnings/perlio |   14 +++
 t/op/caller.t         |    4 +-
 warnings.h            |   10 +-
 win32/win32io.c       |    5 +
 7 files changed, 172 insertions(+), 131 deletions(-)

diff --git a/lib/warnings.pm b/lib/warnings.pm
[...]

PR #18910, opened 2021-06-20
Remove :win32 PerlIO layer

Leont:
This removed :win32 as proposed in #13198

Note: should not be merged until PSC has made its final decision


tonycoz:
The diff looks reasonable.

lib/PerlIO.pm
```diff
@@ -235,12 +235,6 @@ like C<:utf8>.  An example of a possible use might be:
 
 A more elegant (and safer) interface is needed.
 
-=item :win32
```
Grinnz, line 238: I wonder if it would be useful to retain this documentation for historical reasons but mention the version it was removed in. (Don't see any reason to retain any of the other docs)

Leont: It might have been if anyone had actually used it, but they didn't.

rjbs:
@Leont Nicholas, Neil, and I all agree this is good to go.

nwc10:
"Thanks applied". Or disapplied. Whatever...

smpeters:
"Chainsawed"?

nwc10:
I always thought that chainsaws were a bit rough. "Precision guided surgical chainsaw" maybe. But I'm not sure if anyone sells them. At least anyone who is shipping to here...

PR #18917, opened 2021-06-22
Make POSIX::strtoul() return UV if possible

t-a-k:
`POSIX::strtoul()` used to return NV if return value does not fit in IV, but this caused loss of precision if unsigned long and NV are both 64-bit.

This PR will fix it by making `POSIX::strtoul()` return UV instead of IV.

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -3581,12 +3581,12 @@ strtoul(str, base = 0)
 	PERL_UNUSED_VAR(base);
 	if (base == 0 || inRANGE(base, 2, 36)) {
             num = strtoul(str, &unparsed, base);
-#if IVSIZE <= LONGSIZE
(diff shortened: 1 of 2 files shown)


khwilliamson:
@sisyphus please review

sisyphus:
On Fri, Jun 25, 2021 at 7:01 AM Karl Williamson ***@***.***>
wrote:

> @sisyphus <https://github.com/sisyphus> please review
>

LGTM.

An (interesting ?) aside is that with perl-5.34.0, irrespective of nvtype.
when both longsize and ivsize are 4, POSIX::strtoul('12379813738877118345')
returns an NV, not an UV.
I don't think that this patch will alter that behaviour.

However, the actual value of that NV (4294967295.0) is correct, and suffers
no loss of precision.
Therefore, I see little point in raising any objection about it.

Cheers,
Rob

PR #18921, opened 2021-06-22
Update perlgov.pod with Steering Council Election Results

xdg:
This commit updates the Perl Steering Council member list with the results of the June 2021 Term Election.


leonerd:
If it counts, LGTM ;)

PR #18922, opened 2021-06-23
Makefile.SH: Fixes for z/OS

khwilliamson:
This removes two things that are causing problems:
    1) a space in a parameter
    2) DPERL_EXTERNAL_GLOB=''


nwc10:
Looks good.

"Thanks applied"

PR #18925, opened 2021-06-23
Multiple-alias syntax for foreach

nwc10:
Implement [RFC 0001 -  *Multiple-alias syntax for foreach*](https://github.com/Perl/RFCs/blob/master/rfcs/rfc0001.md)

I believe that the RFC process now is "this change is subject to normal code review and merging" by **someone competent who is not me**.

I guess we ask *does it implement what the RFCs specified?* in addition to the usual *does the code have bugs or other defects?* and *on balance, does merging this give a net improvement?*


jkeenan:
On 6/23/21 4:57 AM, Nicholas Clark wrote:
> Implement RFC 0001 - /Multiple-alias syntax for foreach/ 
> <https://github.com/Perl/RFCs/blob/master/rfc/rfc0001.md>
> 
> [That link is correct but currently GitHub is reporting it as a 404. 
> Start at https://github.com/Perl/RFCs/ <https://github.com/Perl/RFCs/> 
> and you'll get links that are 404s. I'll edit the text to remove this 
> note once GitHub no longer has its knickers in a twist]
> 

No, the listed link is one character shorter than the link which you get 
by navigating through the above.

Listed:    https://github.com/Perl/RFCs/blob/master/rfc/rfc0001.md
Confirmed: https://github.com/Perl/RFCs/blob/master/rfcs/rfc0001.md

I'm not sure how you generated this email, so I don't know where in p.r. 
18925 to go to fix.  Can you investigate?

> I believe that the RFC process now is "this change is subject to normal 
> code review and merging" by *someone competent who is not me*.
> 

This is curious.  When I go to https://github.com/Perl/perl5/pull/18925 
I see that I have been requested as a code reviewer.  Should I have 
received an email making that request?  AFAICT I did not receive one.

Note that the p.r. has merge conflicts.

Thank you very much.
Jim Keenan

Grinnz:
> This is curious. When I go to #18925 I see that I have been requested as a code reviewer. Should I have received an email making that request? AFAICT I did not receive one.

You aren't currently assigned as a reviewer, those are just github's suggestions for who to assign as a reviewer (which it decides itself based on various factors AIUI).

nwc10:
I don't know what triggered GitHub to send you an e-mail about this, because I don't think that it was anything I explicitly asked for. **It** has suggested you are a potential reviewer, but I didn't take it up on its suggestions.

Worse - I goofed, and fixed things, but it seems that it didn't e-mail you an **updated** version of the comment, so you're replying to things that I already figured out, fixed and edited away. Hateful software.

> On 6/23/21 4:57 AM, Nicholas Clark wrote: Implement RFC 0001 - /Multiple-alias syntax for foreach/ <https://github.com/Perl/RFCs/blob/master/rfc/rfc0001.md> [That link is correct but currently GitHub is reporting it as a 404. Start at https://github.com/Perl/RFCs/ <https://github.com/Perl/RFCs/> and you'll get links that are 404s. I'll edit the text to remove this note once GitHub no longer has its knickers in a twist]
> No, the listed link is one character shorter than the link which you get by navigating through the above. Listed: https://github.com/Perl/RFCs/blob/master/rfc/rfc0001.md Confirmed: https://github.com/Perl/RFCs/blob/master/rfcs/rfc0001.md I'm not sure how you generated this email, so I don't know where in p.r. 18925 to go to fix. Can you investigate?

Yes, **I** figured that the problem was between keyboard and chair about 4 days ago, and fixed it in perl/RFCs@6f2c56a500d2c4fc65f6db64f258dea29762a24e

Specifically, I was getting confused between links that I had generated in the `README.md` that appeared at the bottom of the page, and links that it was generating at the top of the page, based on the files. However, I think I had hit some sort of confusion with force-pushes to my private test repository, where files didn't update immediately (could be wrong), so I was assuming I was seeing the same thing

> I believe that the RFC process now is "this change is subject to normal code review and merging" by *someone competent who is not me*.
> This is curious. When I go to #18925 I see that I have been requested as a code reviewer. Should I have received an email making that request? AFAICT I did not receive one. Note that the p.r. has merge conflicts. Thank you very much. Jim Keenan

(as above) - I didn't know who should be a code reviewer, but wasn't confident that its suggestions were great. Really, it's the grammar changes that I'm least confident about, and would like someone to say "this doesn't obviously add subtle bugs".

The conflict is due to other more recent changes to B::Concise in the lines around the `$VERSION` declaration. I've resolved the conflict, and pushed the whole branch rebased on blead. All the commits are identical.

t/op/for-many.t
```diff
@@ -0,0 +1,388 @@
+#!./perl
+
+BEGIN {
+    chdir 't' if -d 't';
+    require "./test.pl";
```
wolfsage, line 5: It would be nice if `    set_up_inc('../lib');` was here too so you can `./perl -Ilib t/op/for-many.t`

pod/perlsyn.pod
```diff
@@ -549,6 +558,14 @@ followed by C<my>.  To use this form, you must enable the C<refaliasing>
 feature via C<use feature>.  (See L<feature>.  See also L<perlref/Assigning
 to References>.)
 
+As of Perl 5.36, you can iterate over a list of lexical scalars n-at-a-time.
+If the size of the LIST is not an exact multiple of number of iterator
+variables, then on the last iteration the "excess" iterator variables are
+undefined values, much like if you slice beyond the end of an array.  You
```
wolfsage: `...slice beyond the end of an array`

Do you mean: `@x = ("cat"); my @beyond = @x[5,6];`?

If so, I'm not sure this behaves the same. Consider:
```
use strict;
use warnings;

use Data::Dumper;

{
  my @l = ('a'..'c');

  $_ .= $_ for @l[5,6];

  print "slice past end\n";
  print Dumper(\@l);
}

{
  my @l = ('a'..'c', undef);
  for my $x (@l) {
    $x .= $x;
  }

  print "overwrite aliased values in normal for\n";
  print Dumper \@l;
}

{
  my @l = ('a'..'c', undef);

  for my ($x, $y) (@l) {
    $x .= $x;
    $y .= $y;
  }

  print "overwrite aliased values in n-at-a-time\n";
  print Dumper \@l;
}

{
  my @l = ('a'..'c');

  print "Attempting to overwrite aliased values n-at-a-time with not-enough elements\n";
  for my ($x, $y) (@l) {
    $x .= $x;
    $y .= $y;
  }

  print Dumper \@l;
}

```
```
./perl -Ilib undef.pl
for my (...) is experimental at undef.pl line 28.
for my (...) is experimental at undef.pl line 41.
Use of uninitialized value $_ in concatenation (.) or string at undef.pl line 9.
Use of uninitialized value $_ in concatenation (.) or string at undef.pl line 9.
slice past end
$VAR1 = [
          'a',
          'b',
          'c',
          undef,
          undef,
          '',
          ''
        ];
Use of uninitialized value $x in concatenation (.) or string at undef.pl line 18.
overwrite aliased values in normal for
$VAR1 = [
          'aa',
          'bb',
          'cc',
          ''
        ];
Use of uninitialized value $y in concatenation (.) or string at undef.pl line 30.
overwrite aliased values in n-at-a-time
$VAR1 = [
          'aa',
          'bb',
          'cc',
          ''
        ];
Attempting to overwrite aliased values n-at-a-time with not-enough elements
Use of uninitialized value $y in concatenation (.) or string at undef.pl line 43.
Modification of a read-only value attempted at undef.pl line 43.
```

With n-at-a-time, any aliased undefs are readonly and we crash. Is this expected?

The alternative I guess is that we'd grow the array?

pod/perldelta.pod
```diff
@@ -21,11 +21,17 @@ XXX Any important notices here
...
-XXX New core language features go here.  Summarize user-visible core language
-enhancements.  Particularly prominent performance optimisations could go
-here, but most should go in the L</Performance Enhancements> section.
+=head2 iterating over multiple values at a time.
 
-[ List each enhancement as a =head2 entry ]
+As of Perl 5.36, you can iterate over multiple values at a time by specifying
+a list of lexicals within parentheses. For example,
+
+    for my ($key, $value) (%hash) { ... }
```
wolfsage, line 29: You can also do this:
```
 ./perl -Ilib -Mstrict -Mwarnings -Ilib -e 'my %h = map { $_ => $_ } 1..10; for my ($k, $v, $extra) (%h) { print "K: $k, V: $v, E: $extra\n" }'
for my (...) is experimental at -e line 1.
K: 9, V: 9, E: 3
K: 3, V: 6, E: 6
K: 4, V: 4, E: 10
K: 10, V: 2, E: 2
K: 8, V: 8, E: 1
K: 1, V: 7, E: 7
Use of uninitialized value in concatenation (.) or string at -e line 1.
K: 5, V: 5, E: 
```

Should we allow it?

pod/perlsyn.pod
```diff
@@ -574,6 +591,17 @@ Examples:
         # do something which each %hash
     }
 
+    foreach my ($foo, $bar, $baz) (@list) {
+        # do something three-at-a-time
+    }
+
+    foreach my ($key, $value) (%hash) {
+        # iterate over the hash
+        # The hash is eagerly flattened to a list before the loop starts,
```
wolfsage: I'm not sure what `eagerly flattened to a list` means here.
```
use strict;
use warnings;

{

  package NewStdHash;
  require Tie::Hash;

  our @ISA = qw(Tie::StdHash);

  sub FETCH {
    my ($s, $k) = @_;

    my $v =  $s->{$k};

    print "Fetching $k, returning $v\n";

    $v;
  }
}

my %h;
tie %h, 'NewStdHash';

$h{$_} = $_ for (1..5);

for my ($x, $y) (%h) {
  print "X: $x, Y: $y\n";
}

```
```
./perl -Ilib mh.pl
for my (...) is experimental at mh.pl line 27.
Fetching 2, returning 2
X: 2, Y: 2
Fetching 1, returning 1
X: 1, Y: 1
Fetching 5, returning 5
X: 5, Y: 5
Fetching 3, returning 3
X: 3, Y: 3
Fetching 4, returning 4
X: 4, Y: 4
```

pp_hot.c
```diff
@@ -3928,7 +3959,10 @@ PP(pp_iter)
             *itersvp = newSVsv(cur);
             SvREFCNT_dec(oldsv);
         }
-        if (strEQ(SvPVX_const(cur), max))
+
+            if (UNLIKELY(pad_it)) {
+            }
```
wolfsage, line 3989: Maybe a comment to here say why this block is empty or something

pp_hot.c
```diff
@@ -3906,17 +3920,34 @@ PP(pp_iter)
            It has SvPVX of "" and SvCUR of 0, which is what we want.  */
         STRLEN maxlen = 0;
         const char *max = SvPV_const(end, maxlen);
+            bool pad_it = FALSE;
         if (DO_UTF8(end) && IN_UNI_8_BIT)
             maxlen = sv_len_utf8_nomg(end);
-        if (UNLIKELY(SvNIOK(cur) || SvCUR(cur) > maxlen))
+            if (UNLIKELY(SvNIOK(cur) || SvCUR(cur) > maxlen)) {
```
wolfsage: This if block now looks like its under the DO_UTF8 if block above, but it's not and should be spaced back out properly

pp_hot.c
```diff
@@ -3938,12 +3972,21 @@ PP(pp_iter)
     case CXt_LOOP_LAZYIV: /* integer increment */
     {
         IV cur = cx->blk_loop.state_u.lazyiv.cur;
-        if (UNLIKELY(cur > cx->blk_loop.state_u.lazyiv.end))
+            bool pad_it = FALSE;
+            if (UNLIKELY(cur > cx->blk_loop.state_u.lazyiv.end)) {
+                if (LIKELY(!i)) {
             goto retno;
+                }
```
wolfsage, line 4000: There's more confusing white spacing here

pp_hot.c
```diff
@@ -3987,10 +4034,17 @@ PP(pp_iter)
         if (UNLIKELY(inc > 0
                         ? ix > cx->blk_oldsp
                         : ix <= cx->blk_loop.state_u.stack.basesp)
-        )
+                ) {
+                if (LIKELY(!i)) {
             goto retno;
+                }
```
wolfsage, line 4057: And more weird spacing here

pp_hot.c
```diff
@@ -4002,10 +4056,13 @@ PP(pp_iter)
         if (UNLIKELY(inc > 0
                         ? ix > AvFILL(av)
                         : ix < 0)
-        )
+                ) {
+                if (LIKELY(!i)) {
```
wolfsage, line 4077: And more weird spacing

op.c
```diff
@@ -10233,6 +10234,62 @@ Perl_newFOROP(pTHX_ I32 flags, OP *sv, OP *expr, OP *block, OP *cont)
...
+            }
+            padoff = first_padsv->op_targ;
+
+            /* There should be at least one more PADSV to find, and the ops
+               should have consecutive values in targ: */
+            padsv = (UNOP *) OpSIBLING(first_padsv);
+            do {
+                if (!padsv || padsv->op_type != OP_PADSV) {
+                    /* diag_listed_as: Can't use %s for loop variable */
+                    Perl_croak(aTHX_ "Can't use list for loop variable (not a pad2v at %lu)", how_many_more);
```
wolfsage: I'm not sure how I feel about these errors, they seem kind of scary to perl users who accidentally do:
```
alh@emondsfield:~/p5/perl$ ./perl -Ilib -e 'for my ($a,,$b) (@list) { }'
for my (...) is experimental at -e line 1.
Can't use list for loop variable (not a pad2v at 0) at -e line 1.
```
Do we usually include internal jargon for easy to trigger parsing errors?

pod/perldiag.pod
```diff
@@ -2453,6 +2453,12 @@ some time before now.  Check your control flow.  flock() operates on
 filehandles.  Are you attempting to call flock() on a dirhandle by the
 same name?
 
+=item for my (...) is experimental
+
+(S experimental::for_list) This warning is emitted if you use C<for> to
+iterate over a list of lexicals. This syntax is currently experimental and
```
wolfsage: `to iterate over a list of lexicals` is confusing to me and I always think you mean:
`for my $v ($lexical1, $lexical2, $lexical3) {.. }` but maybe that's just me

wolfsage (changes requested):
The C level changes are a bit beyond me - would be nice to have someone like Dave or Tony review them (if they haven't already).

There are a number of whitespace oddities introduced that make the code hard to comprehend that would be nice if they were fixed up.

Have we done benchmarking to see how this impacts normal for loops?

Also worth doing if you haven't:
 * -DDEBUGGING build
 *  Threaded build
 * Look for any new warnings generated by gcc (and maybe clang) 
 * Maybe a new test.valgrind run but that can be painful

Overall I like the feature though I worry about the 'extra undefs if the list isn't quite the right size'.

I ... don't have a solution I like though so I guess I'm just complaining here.

Good stuff!

pod/perlsyn.pod
hvds: I read it as meaning like "slice" in the sense of:
```
  @a = qw{ a b c };
  ($x1, $y1) = @a[0..1];  # ("a", "b")
  ($x2, $y2) = @a[2..3];  # ("c", undef)
  # @a is unchanged
```
.. which makes sense to me.

Whether it should attempt to make lvaluable undefs by extending the array is less obvious to me: I think I prefer it as it is, but I would imagine very rarely wanting to modify the elements. For the cases where you do want to modify them, though, it's sad that there will be no easy way to check whether any undef you get is readonly.

Whatever behaviour we get definitely needs testing though, if it isn't already.

rjbs:
@nwc10 The last I heard, this was waiting on code review.  What's the current status of this branch?  What prevents us from merging it?

Implementation problems?  Lack of tests?  Lack of docs?  Lack of say-so?

leonerd:
Oh if it's ready-pending-review I should definitely give a look-over.

ext/B/t/optree_for.t
```diff
@@ -0,0 +1,191 @@
...
+    unshift @INC, 't';
+    require Config;
+    if (($Config::Config{'extensions'} !~ /\bB\b/) ){
+        print "1..0 # Skip -- Perl configured without B module\n";
+        exit 0;
+    }
+    if (!$Config::Config{useperlio}) {
+        print "1..0 # Skip -- need perlio to walk the optree\n";
+        exit 0;
+    }
```
tonycoz: useperlio cannot be disabled, I suspect this was copied from an existing ext/B test.

pp_hot.c
```diff
@@ -3896,6 +3896,7 @@ PP(pp_iter)
     itersvp = CxITERVAR(cx);
     assert(itersvp);
 
+    {
     switch (CxTYPE(cx)) {
 
     case CXt_LOOP_LAZYSV: /* string increment */
```
tonycoz: This commit doesn't do what it says it does.  It adds the block around the switch (which isn't purely a whitespace change), and doesn't actually indent the block.

Later commits add lines indented further, so the HEAD includes:
```
        STRLEN maxlen = 0;
        const char *max = SvPV_const(end, maxlen);
            bool pad_it = FALSE;
        if (DO_UTF8(end) && IN_UNI_8_BIT)
```

I suspect a `git rebase --ignore-whitespace` bit you here.

t/op/for-many.t
```diff
@@ -0,0 +1,349 @@
...
+}
+
+use strict;
+use warnings;
+
+my @have;
+
```
tonycoz, line 14: You implement and document continue blocks for the new construct, but don't test them.

ext/B/t/optree_for.t
nwc10: Yes. See #19126

pod/perldelta.pod
nwc10: We have to allow it to be consistent.

hash flattens to a list, and then we iterate over the list.

If we start trying to add special cases, then we have to create some rule about "even iteration count only when iterating a hash" and hence we create a new concept of "list derived from a hash" that's a bit different from any other list.

And if we wanted to disallow that, we'd need to disallow this:

    for my ($k, $v, $joker) (%h1, %h2) {
         ....
    }

and then arguably we would need to *not* disallow this:

    for my ($k, $v, $joker) (%h1, "surprise", %h2) {
         ....
    }

and I don't see any winning move down this path.

pod/perlsyn.pod
nwc10: I didn't mean *that*, certainly. The wording was trying to mean "for a hash with 100 keys, you get a 200 item list. It's not a hash iterator"

As to tied hashes, I'd not considered them, but a previous author's comment probably applies:

    C<foreach> probably won't do what you expect if VAR is a tied or other
    special variable.  Don't do that either.

nwc10: Aha, actually that was about VAR, not EXPR. But I suspect that your loop example is weird with one-at-a-time...

```
use strict;
use warnings;

{

  package NewStdHash;
  require Tie::Hash;

  our @ISA = qw(Tie::StdHash);

  sub FETCH {
    my ($s, $k) = @_;

    my $v =  $s->{$k};

    print "Fetching $k, returning $v\n";

    $v;
  }
}

my %h;
tie %h, 'NewStdHash';

$h{$_} = lc $_ for "A" .. "E";

for my $x (%h) {
  print "X: $x\n";
}
```

```
X: E
Fetching E, returning e
X: e
X: C
Fetching C, returning c
X: c
X: A
Fetching A, returning a
X: a
X: D
Fetching D, returning d
X: d
X: B
Fetching B, returning b
X: b
```

keys are copies in the list, values are aliases that only get read side effects  when actually read.

t/op/for-many.t
nwc10: I can't remember exactly what I did, but it wasn't in the boilerplate for `t/op/for.t` and I suspect I copied that.

I've added in in the rebased version, and also added it to `t/op/for.t` when I was adding tests to it.

pod/perlsyn.pod
nwc10: Yes, it's not a slice. I changed the text to this:

```
As of Perl 5.36, you can iterate over a list of lexical scalars n-at-a-time.
You can only iterate over scalars - unlike list assignment, it's not
possible to use C<undef> to signify a value that isn't wanted.  This is a
limitation of the current implementation, and might be changed in the
future.

If the size of the LIST is not an exact multiple of number of iterator
variables, then on the last iteration the "excess" iterator variables are
aliases to C<undef>, as if the LIST had C<, undef> appended as many times as
needed for its length to become an exact multiple.  This happens whether
LIST is a literal LIST or an array - ie arrays are not extended if their
size is not a multiple of the iteration size, consistent with iterating an
array one-at-a-time.  As these padding elements are not lvalues, attempting
to modify them will fail, consistent with the behaviour when iterating a
list with literal C<undef>s.  If this is not the behaviour you desire, then
before the loop starts either explicitly extend your array to be an exact
multiple, or explicitly throw an exception.
```

There are conflicting options on what to do if the array isn't a multiple of the size. I can think of

1. exception before looping
2. grow the array
3. LVALUE undefs
4. RVALUE undefs

Throwing an exception straight away doesn't feel very perlish. Growing the array is inconsistent with current single iterator `foreach` semantics, and isn't an answer for a non-LVALUE list

Whilst undefs are inconsistent with current single iterator `foreach` semantics - there you never get a value that was not in your list expression - I think that RVALUE is better than LVALUE. If we offered LVALUE undefs, then if the loop does modifications expecting that they get stored *somewhere*, then we'd silently discard them after the loop.

No answer is ideal, but this one at least gives the code some idea that it has encountered a situation that it wasn't written to cope with.

pod/perlsyn.pod
nwc10: I changed the text to *The hash is immediately copied to a flat list before the loop starts* which gets rid of the ill defined jargon "eagerly". There might be an ever better phrasing, but I feel that this one is better than what we had, and good enough to go with. Thanks for making me think harder.

pp_hot.c
nwc10: Yes, totally. I came back to the code and thought the same thing. It's in the rebase.

pp_hot.c
nwc10: As @tonycoz spotted (below, on this view) it's a side effect of me not realising that the `git rebase` munged my whitespace. "Fixed in post" - er, on the upcoming rebase.

pp_hot.c
nwc10: Fixed.

pp_hot.c
nwc10: Fixed

op.c
nwc10: Right, in all the other comments I missed the detail of this one

1. we *do* generate panic messages in various places, but they start "panic: "
2. we ought to be able to parse this, and I don't know how...

So this needs a testcase and fixing.

pod/perldiag.pod
nwc10: Good catch. I've changed it to this:

```
(S experimental::for_list) This warning is emitted if you use C<for> to
iterate multiple values n-at-a-time. This syntax is currently experimental
and its behaviour may change in future releases of Perl.
```

because *n-at-a-time* is how it's described in `perlsyn.pod`.

ext/B/t/optree_for.t
nwc10: Gone in the rebase

pp_hot.c
nwc10: It did. Fixed in the rebase, and thanks for correctly diagnosing the problem.

t/op/for-many.t
nwc10: Well spotted. Fixed.

We also didn't seem to have tests for LVALUE aliases in exsting `for` loops, and for attempting to modify RVALUE aliases --  #19133

op.c
nwc10: How - like this:

```
--- a/perly.y
+++ b/perly.y
@@ -1384,7 +1384,7 @@ my_scalar:        scalar

 /* A list of scalars for "for my ($foo, $bar) (@baz)"  */
 list_of_scalars:       list_of_scalars[list] PERLY_COMMA
-                       { parser->in_my = 0; $$ = $list; }
+                       { $$ = $list; }
        |               list_of_scalars[list] PERLY_COMMA scalar
                        {
                          $$ = op_append_elem(OP_LIST, $list, $scalar);
diff --git a/t/op/for-many.t b/t/op/for-many.t
```

ie that was my bug in the grammar.

What had been happening (the panic) was because the internal state had switched back to `parser->in_my` false at the (first) double comma, and so the subsequent variable(s) optree(s) were built as GV lookups, instead of pad lookups. And when those OPs arrived in newFORLOOP, the paranoia checks fired.

So, branch now should be good in 390937f4796971f94123b325d7c484150be6b04c

nwc10:
@wolfsage - GitHub still thinks "changes requested" but I can't work out what it's nagging about. I think I've addressed all the specific comments you made (and everyone else's), although I only marked as "resolved" those that I was confident you would agree with.

pod/perlsyn.pod
leonerd: I think of all the various options, the one you've picked is the least surprising and most perlish. It basically pretends that there were extra `undef`s to fill in the gaps. That is consistent with things like

```perl
my %hash = @arr;
my ($x, $y, $z) = @values;
```

Also your reasoning for readonly RVALUEs seems good - trying to read it you just get a hole, but if you attempted to write it you'd get a complaint.

nwc10:
Pushing a rebase to avoid the conflicts.

pod/perlsyn.pod
wolfsage: Okay, and I mistook how flattening and ties with FETCH worked so I see where I was confused too. Thanks

op.c
wolfsage: Awesome, thanks

pod/perldelta.pod
```diff
@@ -21,11 +21,17 @@ XXX Any important notices here
 
 =head1 Core Enhancements
 
-XXX New core language features go here.  Summarize user-visible core language
-enhancements.  Particularly prominent performance optimisations could go
-here, but most should go in the L</Performance Enhancements> section.
+=head2 iterating over multiple values at a time.
```
wolfsage: nit: Usually these headings don't end with a period

pod/perlsyn.pod
```diff
@@ -282,6 +282,15 @@ The following compound statements may be used to control flow:
 
     PHASE BLOCK
 
+As of Perl 5.36, you can iterate over multiple values at a time by specifying
+a list of lexicals within parentheses
```
wolfsage: nit: This should have a period or maybe a `:`

pod/perlsyn.pod
```diff
@@ -549,6 +558,24 @@ followed by C<my>.  To use this form, you must enable the C<refaliasing>
 feature via C<use feature>.  (See L<feature>.  See also L<perlref/Assigning
 to References>.)
 
+As of Perl 5.36, you can iterate over a list of lexical scalars n-at-a-time.
```
wolfsage: I think this wording `list of lexical scalars` should also be changed here (like it was elsewhere) to `iterate over multiple values` or something. Maybe be consistent with all of them?

pod/perlsyn.pod
```diff
@@ -549,6 +558,24 @@ followed by C<my>.  To use this form, you must enable the C<refaliasing>
...
+You can only iterate over scalars - unlike list assignment, it's not
+possible to use C<undef> to signify a value that isn't wanted.  This is a
+limitation of the current implementation, and might be changed in the
+future.
+
+If the size of the LIST is not an exact multiple of number of iterator
```
wolfsage: `not an exact multiple of number of iterator` -> `not an exact multiple of the number of iterator`

wolfsage (changes requested):
One or two minor doc fixups but seems good to me (I only skimmed the C changes this time)

pod/perldelta.pod
nwc10: Well spotted. Removed - it's now the same as the others in the same document.

pod/perlsyn.pod
nwc10: Similar text used `:` so I added a colon.

pod/perlsyn.pod
nwc10: Yes, consistency is important. I fail!

I changed it to this text - can it be improved further, or are we good enough?

```
As of Perl 5.36, you can iterate over multiple values at a time.
You can only iterate with lexical scalars as the iterator variables - unlike
list assignment, it's not possible to use C<undef> to signify a value that
isn't wanted.  This is a limitation of the current implementation, and might
be changed in the future.
```

pod/perlsyn.pod
nwc10: D'oh!

Well spotted. Fixed.

pod/perlsyn.pod
wolfsage: Wfm, thanks

wolfsage (approved):

nwc10:
I think that the "Sanity" test has failed because I pushed a commit onto the branch as-is, and as it's based on blead from before the recent release, "changed modules with unchanged VERSION" vs "the latest tag" differs (and so fails)

So a rebase is coming soon, to see if I can get to "All tests successful". It's not easy being green.

lib/B/Deparse.pm
```diff
@@ -3954,7 +3954,18 @@ sub loop_common {
 	} else {
 	    $ary = $self->deparse($ary, 1);
 	}
-	if (null $var) {
+        my $iter_targ = $kid->first->first->targ;
```
smpeters, line 3957: I'm not sure how pedantic we want to be here, but the indentation looks off in Github because Deparse.pm is indented with tabs while the code added here is indented with spaces.

nwc10: I'm not sure either.

Usually I try to match the indenting (etc) style of the surrounding code, and I missed that this file had tabs.

Well, ...:

```
$ git show 8b7aa8e4305788f70f89b55541ae664e8ecb17e0 lib/B/Deparse.pm | perl -pe 's/\t/......../g'
commit 8b7aa8e4305788f70f89b55541ae664e8ecb17e0
Date:   Mon Apr 26 11:23:33 2021 +0000

    B::Deparse now handles n-at-a-time for.

diff --git a/lib/B/Deparse.pm b/lib/B/Deparse.pm
index 2d33039a80..12f6a63670 100644
--- a/lib/B/Deparse.pm
+++ b/lib/B/Deparse.pm
@@ -52,7 +52,7 @@ use B qw(class main_root main_start main_cv svref_2object opnumber perlstring
         MDEREF_SHIFT
     );

-$VERSION = '1.58';
+$VERSION = '1.59';
 use strict;
 our $AUTOLOAD;
 use warnings ();
@@ -3954,7 +3954,18 @@ sub loop_common {
 ........} else {
 ........    $ary = $self->deparse($ary, 1);
 ........}
-........if (null $var) {
+        my $iter_targ = $kid->first->first->targ;
+        if ($iter_targ) {
+            # for my ($foo, $bar) () stores the count (less 1) in the targ of
+            # the ITER op.
+            my @vars;
+            my $targ = $enter->targ;
+            while ($iter_targ-- >= 0) {
+                push @vars, $self->padname_sv($targ)->PVX;
+                ++$targ;
+            }
+            $var = 'my (' . join(', ', @vars) . ')';
+        } elsif (null $var) {
             $var = $self->pp_padsv($enter, 1, 1);
 ........} elsif ($var->name eq "rv2gv") {
 ........    $var = $self->pp_rv2sv($var, 1);
```

it does and it doesn't. The line above my new block had a tab, but the first line after it did not.

The whole file is a inconsistent. There are parts of it with tabs, and other parts (I *guess* newer, but I didn't see a good way to check this) that have spaces.

So having discovered that, there doesn't seem to be any "right" answer, so I'm tempted to say that this part can be spaces.

tonycoz: I thought we converted lines we touched to spaces, from perlhack:
```
The codebase is a mixture of tabs and spaces for indentation, and we
are moving to spaces only.  Converting lines you're patching from 8-wide
tabs to spaces will help this migration.
```

PR #18931, opened 2021-06-24
perlnewmod.pod: Remove dead link and make a header line stand out more.

LeamHall:
The link for Ken William's Tutorial is dead, removed it from the document.

In the "Step-by-step" section there's a command line used as a header, and it is less visible than a plain text header. 
 - Added a header and move the command line between the paragraphs.
 - Changed the semi-colons in the command line to "&&" on the advice of DBOOK. 
 - - This allows the commands to visibly fail if there's an error.


LeamHall:
I saw the failure on Windows, is that something I should have fixed beforehand?

PR #18933, opened 2021-06-24
Clean up splitting regcomp.c fcn in two

khwilliamson:
This is a rebased https://github.com/Perl/perl5/pull/18835

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -19319,7 +19319,7 @@ S_optimize_regclass(pTHX_
 
     U8 op = END;    /* The returned node-type, initialized to an impossible
                        one.  */
-    UV value;
+    UV value = 0;
     PERL_UINT_FAST8_T i;
     UV partial_cp_count = 0;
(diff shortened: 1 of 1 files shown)


regcomp.c
```diff
@@ -19443,7 +19448,7 @@ S_optimize_regclass(pTHX_
```
khwilliamson, line 19448: In the original PR @hvds said:
Is there any reason not to use the more obvious ternary here: *invert ? NPOSIXL : POSIXL? (Maybe for a future commit.)
@khwilliamson responded

Yes, if this is done, It shouldn't be in this commit, which is trying to avoid doing more than necessary in splitting the function.

Your proposal is clearer. A reason not to do what you suggest is because this is a repeat of a common paradigm in this file, so it isn't out of the blue, and it works without change if a new type of POSIXL regnode (and inverse) is added, which has long been a possibility and remains so.

====
And that got committed, but without real resolution

PR #18934, opened 2021-06-25
Fix links 3

thibaultduponchelle:


xenu:
>l-optperl.html not found even in webarchive.

http://web.archive.org/web/20041204135741/http://www-106.ibm.com/developerworks/library/l-optperl.html

thibaultduponchelle:
> > l-optperl.html not found even in webarchive.
> 
> http://web.archive.org/web/20041204135741/http://www-106.ibm.com/developerworks/library/l-optperl.html

OK I don't know how I missed it, thank you @xenu  I will update this PR

Grinnz:
You will need to update t/porting/known_pod_issues.dat:

```
# 3) If there remain false positive or problems that you don't plan to fix right
#    now,
#    run this test script by hand, using the following formula (on
#    Un*x-like machines):
#         cd t
#         ./perl -I../lib porting/podcheck.t --regen
# 
#    That should cause all current potential problems to be accepted by
#    the program, so that the next time it runs, they won't be flagged.
```

This should update the known issues in this file from 113 to 115.

jkeenan:
> You will need to update t/porting/known_pod_issues.dat:
> 
> ```
> # 3) If there remain false positive or problems that you don't plan to fix right
> #    now,
> #    run this test script by hand, using the following formula (on
> #    Un*x-like machines):
> #         cd t
> #         ./perl -I../lib porting/podcheck.t --regen
> # 
> #    That should cause all current potential problems to be accepted by
> #    the program, so that the next time it runs, they won't be flagged.
> ```
> 
> This should update the known issues in this file from 113 to 115.

Handled locally, merged, commits squashed, committed to blead:
```
commit cb222c40b553aa18f3f770a448351317a16b2c47 (HEAD -> blead, origin/blead, origin/HEAD)
Author:     Thibault DUPONCHELLE <thibault.duponchelle@gmail.com>
AuthorDate: Fri Jun 25 09:38:56 2021 +0200
Commit:     James E Keenan <jkeenan@cpan.org>
CommitDate: Fri Jun 25 20:59:30 2021 +0000

    Dead link on ccl4, not found anywhere else but in webarchive. Use webarchive for l-optperl.html
    
    oreilly.com.tw no longer exists
    
    oreilly.com.tw no longer exists and jp and cn are give right afterwards
    
    Committer: update t/porting/known_pod_issues.dat
```
Closing ticket.

thibaultduponchelle:
Great 👍 thank you @jkeenan

PR #18938, opened 2021-06-25
fix new libperl.t 'no symbols' warning on darwin

dwmarshall:
A change in October 2019 added handling for the case that 'nm' might
emit the error message 'no symbols' (apparently erroneously. Apple seems
to have included a fix to upstream LLVM to change the message to be a
file reference followed by ': no symbols'. For libperl.t, the message is
typically '../libperl.a:perlapi.o: no symbols'. This change handles both
the old and current behavior.

### DIFF
--- a/t/porting/libperl.t
+++ b/t/porting/libperl.t
@@ -472,8 +472,12 @@ if (defined $nm_err_tmp) {
         while (<$nm_err_fh>) {
             # OS X has weird error where nm warns about
             # "no name list" but then outputs fine.
+            # llvm-nm may also complain about 'no symbols'. In some
+            # versions this is exactly the string "no symbols\n" but in later
+            # versions becomes a string followed by ": no symbols\n". For this
+            # test it is typically "../libperl.a:perlapi.o: no symbols\n"
             if ( $^O eq 'darwin' ) {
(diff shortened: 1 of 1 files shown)


dwmarshall:
See https://github.com/llvm/llvm-project/commit/a9f0025acd2235dccec51cf42d1e604675639343 for the change in `nm` output. There's no way to determine in which version of XCode Apple applied that change.

jkeenan:
> A change in October 2019 added handling for the case that 'nm' might
> emit the error message 'no symbols' (apparently erroneously. Apple seems
> to have included a fix to upstream LLVM to change the message to be a
> file reference followed by ': no symbols'. For libperl.t, the message is
> typically '../libperl.a:perlapi.o: no symbols'. This change handles both
> the old and current behavior.

Could someone who regularly builds and tests perl on a recent Mac check this p.r. out?

Thank you very much.
Jim Keenan

nwc10:
I have CLI access to a Mac. I don't see the warning **without** the patch. I don't see it **with** the patch. Which (at least) means that the patch doesn't mess up *the current behaviour*, which is all that I can test.

Rebased and pushed to blead. So "Thanks applied".

PR #18941, opened 2021-06-26
core-team.json: one-stop shopping for core team members

rjbs:
I am following up on a [recent Core Team thread](https://perl.topicbox.com/groups/perl-core/T53277f85ba05d655-Mc2d4a664b16cdd474d7394ef) where I said it would be useful to have a single canonical source of machine-readable core team members, and work from that.  Here's what I did:

* started with the list of core team member emails from mailing list
* replaced emails of those who had asked me to
* put them in a JSON file (because we have no core YAML and I wanted to make this trivial)
* wrote a program to generate POD from that as proof of concept

I wanted to stick to "person == email" in the data, so I made the doc-generating program use git's mailmap to get real names.  This required making some new entries in the git mailmap.

I left the individual steps (update README, update MANIFEST) distinct because they spell out what has to happen.  If people prefer, this can be all squashed down.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,5 +1,7 @@
 # https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html#_mapping_authors
 Aaron Crane <arc@cpan.org> <perl@aaroncrane.co.uk>
+Abhijit Menon-Sen <ams@toroid.org> <ams@toroid.org>
+Andy Dougherty <doughera@lafayette.edu> <doughera@lafayette.edu>
 Andy Dougherty <doughera@lafayette.edu> Andy Dougherty <doughera.lafayette.edu>
 Andy Dougherty <doughera@lafayette.edu> Andy Dougherty <doughera@fractal.phys.lafayette.edu>
 Andy Dougherty <doughera@lafayette.edu> Andy Dougherty <doughera@lafcol.lafayette.edu>
@@ -7,16 +9,21 @@ Andy Dougherty <doughera@lafayette.edu> Andy Dougherty <doughera@newton.phys.laf
 Audrey Tang <cpan@audreyt.org> Autrijus Tang <unknown>
 Audrey Tang <cpan@audreyt.org> autrijus@ossf.iis.sinica.edu.tw <autrijus@ossf.iis.sinica.edu.tw>
 Ævar Arnfjörð Bjarmason <avar@cpan.org> Ævar Arnfjörð Bjarmason <avarab@gmail.com>
+Chad Granum <exodist7@gmail.com> <exodist7@gmail.com>
 Chip Salzenberg <chip@atlantic.net> Chip <chip@pobox.com>
 Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@ci005.sv2.upperbeyond.com>
 Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@perl.com>
 Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@pobox.com>
 Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@dun.nielsen.com>
 Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@nielsenmedia.com>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> <chris@bingosnet.co.uk>
 Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris BinGOs Williams <chris@bingosnet.co.uk>
 Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris Williams <chris@bingosnet.co.uk>
 Craig A. Berry <craigberry@mac.com> <Craig A. Berry)>
 Craig A. Berry <craigberry@mac.com> <craig.a.berry@gmail.com>
+Craig Berry <craigberry@mac.com> <craigberry@mac.com>
+Dagfinn Ilmari Mannsåker <ilmari@ilmari.org> <ilmari@ilmari.org>
+David Golden <xdg@xdg.me> <xdg@xdg.me>
 David Mitchell <davem@iabyn.com> <davem@fdisolutions.com>
 David Mitchell <davem@iabyn.com> <davem@iabyn.com>
 David Nicol <davidnicol@gmail.com> david nicol <whatever@davidnicol.com>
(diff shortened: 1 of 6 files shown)


Porting/perlgov-team-update
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env perl
+use v5.30.0;
+use warnings;
+use JSON::PP;
+
+my $file = 'Porting/core-team.json';
+my $data = JSON::PP->new->decode(scalar `cat $file`);
```
Grinnz: This will need to either be `decode_json` or set the `utf8` property in the decoder, since it's reading bytes. (Unicode may not always be UTF-8 encoded in JSON, there are other escape mechanisms)

Porting/perlgov-team-update
```diff
@@ -0,0 +1,26 @@
...
+  $pod .= qq{=head2 \u$key Members\n\n=over 4\n\n};
+
+  my @items = map {; "<$_>" } $data->{$key}->@*;
+
+  open(my $fh, '-|', 'git', 'check-mailmap', @items)
+    or die "error running check-mailmap: $!";
+
+  my @lines = <$fh>;
+
+  $pod .= "=item $_\n" for sort @lines;
```
Grinnz, line 21: And this should encode back to UTF-8.

Porting/perlgov-team-update
```diff
@@ -0,0 +1,26 @@
...
+my $data = JSON::PP->new->decode(scalar `cat $file`);
+
+my $pod = q{};
+
+for my $key (qw( active inactive )) {
+  $pod .= qq{=head2 \u$key Members\n\n=over 4\n\n};
+
+  my @items = map {; "<$_>" } $data->{$key}->@*;
+
+  open(my $fh, '-|', 'git', 'check-mailmap', @items)
```
Grinnz, line 22: These args should then be encoded back to UTF-8.

Porting/perlgov-team-update
Grinnz, line 27: Recommend a `use locale` for this sort. Though it also may require decoding from UTF-8 before sort + encoding the final output to UTF-8 for that to work properly.

Porting/perlgov-team-update
Grinnz: The specific failure mode here is: JSON contains a `\uXXXX` escaped character, as well as UTF-8 encoded characters. As a result `$data` will have a mix of UTF-8 encoded text and decoded characters and thus be impossible to correctly encode for the subsequent commands. Therefore when reading JSON bytes, the UTF-8 decoding step is required before (or as part of) decoding.

.mailmap
```diff
@@ -1,71 +1,95 @@
...
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@dun.nielsen.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@nielsenmedia.com>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> <chris@bingosnet.co.uk>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris BinGOs Williams <chris@bingosnet.co.uk>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris Williams <chris@bingosnet.co.uk>
+Craig A. Berry <craigberry@mac.com> <Craig A. Berry)>
+Craig A. Berry <craigberry@mac.com> <craig.a.berry@gmail.com>
+Craig Berry <craigberry@mac.com> <craigberry@mac.com>
+Dagfinn Ilmari Mannsåker <ilmari@ilmari.org> <ilmari@ilmari.org>
+David Golden <xdg@xdg.me> <xdg@xdg.me>
```
xdg, line 27: Could you please also map `dagolden@cpan.org` to `xdg@xdg.me`?

.mailmap
```diff
@@ -1,71 +1,95 @@
...
+Chad Granum <exodist7@gmail.com> <exodist7@gmail.com>
+Chip Salzenberg <chip@atlantic.net> Chip <chip@pobox.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@ci005.sv2.upperbeyond.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@perl.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <chip@pobox.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@dun.nielsen.com>
+Chip Salzenberg <chip@atlantic.net> Chip Salzenberg <salzench@nielsenmedia.com>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> <chris@bingosnet.co.uk>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris BinGOs Williams <chris@bingosnet.co.uk>
+Chris 'BinGOs' Williams <chris@bingosnet.co.uk> Chris Williams <chris@bingosnet.co.uk>
```
haarg, line 21: There's a lot of entries like this in the file that are redundant.  A single entry like:
```
Chris 'BinGOs' Williams <chris@bingosnet.co.uk> <chris@bingosnet.co.uk>
```
will change the name and email of anything matching the email `chris@bingosnet.co.uk`.

Including the second name should only be needed if you want to disambiguate email address like `root@localhost` that have been used by multiple people.

Porting/perlgov-team-update
karenetheridge: alternatively, the file can be read with the :encoding(UTF-8) layer applied.

.mailmap
karenetheridge: blame me for that, not rjbs. he's just sorting the file.

Porting/perlgov-team-update
rjbs: This seems like a bad idea.  It would mean that the sort might do quite a bit of re-ordering if different runners of the script had different locales, no?  Simpler, I think, to just leave things in a single, stable order, even if it's not what any single phone book editor would pick.

Porting/perlgov-team-update
rjbs: I've done as Karen suggested.

rjbs:
I will very likely merge this in the next 48 hours.

Porting/perlgov-team-update
```diff
@@ -0,0 +1,33 @@
...
+my $json = do { local $/; <$fh> };
+
+my $data = JSON::PP->new->decode($json);
+
+my $pod = q{};
+
+for my $key (qw( active inactive )) {
+  $pod .= qq{=head2 \u$key Members\n\n=over 4\n\n};
+
+  my @items = map {; decode('utf-8', "<$_>", Encode::FB_CROAK) }
```
Grinnz: This seems like a double-decode?

Porting/perlgov-team-update
Grinnz: This still needs to be done with the input now being decoded. Pipe open takes byte arguments.

Porting/perlgov-team-update
rjbs: You're right, that was meant to be an encode, but my mind wandered, thanks!

PR #18943, opened 2021-06-27
Three small changes to perlintro.pod

nwc10:
Simplify the text about declaring variables with my. The previous text said both that "you don't have to pre-declare" and "you have to declare them", where it wasn't clear that the former was about types, and the latter about variables. The new phrasing tries to be clearer.

Note that special variables can also be "all capitals", not just "punctuation". Remove the in joke about "line noise" because it adds little, but many people use it as a negative trope about Perl.

Remove mention of given/when. We should not be promoting something we intend to change. We should have done this years ago when we made it experimental.

I wondered about changing "owl" to "alpaca" given that owls aren't camelids, whist alpacas are and feature on another book, but left it for now. Should we do this too?

### DIFF
--- a/pod/perlintro.pod
+++ b/pod/perlintro.pod
@@ -162,7 +162,7 @@ A scalar represents a single value:
 
 Scalar values can be strings, integers or floating point numbers, and Perl
 will automatically convert between them as required.  There is no need
-to pre-declare your variable types, but you have to declare them using
+to pre-declare your variable types - just declare their names using
 the C<my> keyword the first time you use them.  (This is one of the
 requirements of C<use strict;>.)
 
@@ -172,8 +172,9 @@ Scalar values can be used in various ways:
  print "The animal is $animal\n";
  print "The square of $answer is ", $answer * $answer, "\n";
 
-There are a number of "magic" scalars with names that look like
-punctuation or line noise.  These special variables are used for all
+There are a number of "magic" scalars with names that use punctuation symbols,
+and a few that are all uppercase letters.
+These special variables are used for all
 kinds of purposes, and are documented in L<perlvar>.  The only one you
 need to know about for now is C<$_> which is the "default variable".
 It's used as the default argument to a number of functions in Perl, and
@@ -326,9 +327,7 @@ running the program.  Using C<strict> is highly recommended.
 
 =head2 Conditional and looping constructs
 
-Perl has most of the usual conditional and looping constructs.  As of Perl
-5.10, it even has a case/switch statement (spelled C<given>/C<when>).  See
-L<perlsyn/"Switch Statements"> for more details.
+Perl has most of the usual conditional and looping constructs.
 
 The conditions can be any Perl expression.  See the list of operators in
 the next section for information on comparison and boolean logic operators,
--- a/pod/perlintro.pod
+++ b/pod/perlintro.pod
@@ -161,9 +161,8 @@ A scalar represents a single value:
  my $answer = 42;
 
 Scalar values can be strings, integers or floating point numbers, and Perl
-will automatically convert between them as required.  There is no need
-to pre-declare your variable types - just declare their names using
-the C<my> keyword the first time you use them.  (This is one of the
+will automatically convert between them as required.  You have to declare
+them using the C<my> keyword the first time you use them.  (This is one of the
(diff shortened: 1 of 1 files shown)


neilb:
Looks good. I have one small (pedantic) suggestion: maybe change

> There are a number of "magic" scalars

to

> There are a number of "magic" scalar variables

nwc10:
Thanks. I like this suggestion. For most of the initial documentation, I suspect that *scalar variables* is better style wise than *scalars* - **we** know that the latter means the former, but I'm not sure how obvious it is to the intended readership. It doesn't feel like a safe trade off to shave some seeming **small** verbosity for some potential **high** confusion.

pod/perlintro.pod
```diff
@@ -162,7 +162,7 @@ A scalar represents a single value:
 
 Scalar values can be strings, integers or floating point numbers, and Perl
 will automatically convert between them as required.  There is no need
-to pre-declare your variable types, but you have to declare them using
+to pre-declare your variable types - just declare their names using
```
jmacdotorg: Let's simplify this sentence even further:

`You have to declare them using the C<my> keyword the first time you use them.`

pod/perlintro.pod
```diff
@@ -172,8 +172,9 @@ Scalar values can be used in various ways:
  print "The animal is $animal\n";
  print "The square of $answer is ", $answer * $answer, "\n";
 
-There are a number of "magic" scalars with names that look like
-punctuation or line noise.  These special variables are used for all
+There are a number of "magic" scalars with names that use punctuation symbols,
```
jmacdotorg: Making it even more clear:

`Perl defines a number of "magic" scalars with one-character names, often punctuation marks or digits.`

jmacdotorg (commented):
I like all these changes, and added my suggestions for grinding them down a little finer.

Full disclosure: I edited this page as an exercise last year, as part of my style guide research, and all three of the problem areas you found were among those I found as well... and my suggestions now are based on my suggestions then!

pod/perlintro.pod
nwc10: Yes, that's better than mine. Thanks.

pod/perlintro.pod
nwc10: I like your structure better than mine, and yours covers `$a` and `$b` too, which the original does not. I liked Neil's suggestion of *scalar variables* instead of *scalars*. You've done a lot more thinking about style than either Neil or me, so I wondered what your thoughts are on this. I wasn't 100% decided either way.

Also, having got a little further in that document and got to `%ENV` I realised that I'd count that as a variable to mention here. And in typing this comment I also realise that **it** isn't a scalar. So that sort of invalidates the previous sentence saying "scalars" or "scalar variables".

Grinnz: Yes, also `@_`, `@ARGV`, `@INC`, `%INC`, and all of the typeglobs.

jmacdotorg: All excellent points!

Perhaps something like `Perl defines a number of "magic" variables with short names, often single punctuation marks or digits`? That doesn't discount our friends like %ENV.

nwc10: Except then I read the surrounding paragraphs too (which this line-by-line "code" review interface tends to hide) and realised

1. This is, of course, the section on *scalars*
2. The term *scalars* has been defined just above
3. There are references to "special variables" in the corresponding sections on *arrays* and *hashes*
4. But those sections don't use the term *"magic"*

So, I've just pushed a revised change with this:

```
Perl defines a number of special scalars with short names, often single
punctuation marks or digits. These variables are used for all
kinds of purposes, and are documented in L<perlvar>.  The only one you
need to know about for now is C<$_> which is the "default variable".
It's used as the default argument to a number of functions in Perl, and
it's set implicitly by certain looping constructs.
```

which isn't quite what any of us had suggested, and takes out the term *"magic"*.

pod/perlintro.pod
nwc10: If I have this right, that gets us to

```
Scalar values can be strings, integers or floating point numbers, and Perl
will automatically convert between them as required.  You have to declare
them using the C<my> keyword the first time you use them.  (This is one of the
requirements of C<use strict;>.)
```

PR #18944, opened 2021-06-27
Fast determination of most and least set bits in a machine word

khwilliamson:
Modern hardware tends to have machine instructions to find this information.

This series of commits creates functions that return the least and most set bits in the largest word perl thinks the hardware is capable of, and uses them in all but one of the places in core that I am aware of could benefit.  (Perhaps you know of other places.)  The final place has to do with Unicode handling, is more complex, and I'm deferring that until this P.R. is merged.

There are builtins for clang, gcc, and Windows to find the msb; a builtin for Windows for the lsb, and a POSIX function, ffs(), for the latter, but defined only for ints.  Many platforms have a ffsl() additionally, and blead already has probes for that and the other possibilities.  If a platform is lacking, fast bit twiddling, and deBruijn sequences are used instead, so the functions are completely general.

These functions are currently considered core-only, but if there is a benefit for modules to be able to use them, I could easily create public documentation for them.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -1140,6 +1140,7 @@ ATidRp	|bool	|is_utf8_invariant_string_loc|NN const U8* const s	    \
 		|STRLEN len						    \
 		|NULLOK const U8 ** ep
 CTiRp	|unsigned|my_ffs|PERL_UINTMAX_T word
+CTiRp	|unsigned|my_msbit_pos|PERL_UINTMAX_T word
 CTiRp	|unsigned|single_1bit_pos|PERL_UINTMAX_T word
 #ifndef EBCDIC
 CTiRp	|unsigned int|variant_byte_number|PERL_UINTMAX_T word
--- a/embed.fnc
+++ b/embed.fnc
@@ -1139,6 +1139,7 @@ ATdmoR	|bool	|is_utf8_invariant_string|NN const U8* const s		    \
 ATidRp	|bool	|is_utf8_invariant_string_loc|NN const U8* const s	    \
 		|STRLEN len						    \
 		|NULLOK const U8 ** ep
+CTiRp	|unsigned|single_1bit_pos|PERL_UINTMAX_T word
 #ifndef EBCDIC
 CTiRp	|unsigned int|variant_byte_number|PERL_UINTMAX_T word
 #endif
--- a/embed.fnc
+++ b/embed.fnc
@@ -1139,6 +1139,7 @@ ATdmoR	|bool	|is_utf8_invariant_string|NN const U8* const s		    \
 ATidRp	|bool	|is_utf8_invariant_string_loc|NN const U8* const s	    \
 		|STRLEN len						    \
 		|NULLOK const U8 ** ep
+CTiRp	|unsigned|my_ffs|PERL_UINTMAX_T word
 CTiRp	|unsigned|single_1bit_pos|PERL_UINTMAX_T word
 #ifndef EBCDIC
 CTiRp	|unsigned int|variant_byte_number|PERL_UINTMAX_T word
--- a/embed.h
+++ b/embed.h
@@ -331,6 +331,7 @@
 #define my_fflush_all()		Perl_my_fflush_all(aTHX)
 #define my_ffs			Perl_my_ffs
 #define my_fork			Perl_my_fork
+#define my_msbit_pos		Perl_my_msbit_pos
 #define my_popen_list(a,b,c)	Perl_my_popen_list(aTHX_ a,b,c)
 #define my_setenv(a,b)		Perl_my_setenv(aTHX_ a,b)
 #define my_socketpair		Perl_my_socketpair
--- a/embed.h
+++ b/embed.h
@@ -559,6 +559,7 @@
 #define set_context		Perl_set_context
 #define setdefout(a)		Perl_setdefout(aTHX_ a)
 #define share_hek(a,b,c)	Perl_share_hek(aTHX_ a,b,c)
+#define single_1bit_pos		Perl_single_1bit_pos
 #define sortsv(a,b,c)		Perl_sortsv(aTHX_ a,b,c)
 #define sortsv_flags(a,b,c,d)	Perl_sortsv_flags(aTHX_ a,b,c,d)
 #define stack_grow(a,b,c)	Perl_stack_grow(aTHX_ a,b,c)
--- a/embed.h
+++ b/embed.h
@@ -329,6 +329,7 @@
 #define my_exit(a)		Perl_my_exit(aTHX_ a)
 #define my_failure_exit()	Perl_my_failure_exit(aTHX)
 #define my_fflush_all()		Perl_my_fflush_all(aTHX)
+#define my_ffs			Perl_my_ffs
 #define my_fork			Perl_my_fork
 #define my_popen_list(a,b,c)	Perl_my_popen_list(aTHX_ a,b,c)
 #define my_setenv(a,b)		Perl_my_setenv(aTHX_ a,b)
--- a/globvar.sym
+++ b/globvar.sym
@@ -12,6 +12,7 @@ PL_c9_utf8_dfa_tab
 PL_charclass
 PL_check
 PL_core_reg_engine
+PL_deBruijn_bitpos_tab
 PL_EXACTFish_bitmask
 PL_EXACT_REQ8_bitmask
 PL_extended_utf8_dfa_tab
--- a/inline.h
+++ b/inline.h
@@ -610,6 +610,12 @@ Perl_single_1bit_pos(PERL_UINTMAX_T word)
 
 }
 
+#if defined(WIN32) && (defined(PERL_USE_FFS) || defined(PERL_USE_CLZ))
+#  include <intrin.h>
+#  pragma intrinsic(_BitScanReverse,_BitScanReverse64)
+#  pragma intrinsic(_BitScanForward,_BitScanForward64)
+#endif
+
 PERL_STATIC_INLINE unsigned
 Perl_my_msbit_pos(PERL_UINTMAX_T word)
 {
@@ -625,8 +631,6 @@ Perl_my_msbit_pos(PERL_UINTMAX_T word)
     return (PERL_UINTMAX_SIZE * CHARBITS) - 1 - PERL_USE_CLZ(word);
 
 #  else
-#    include <intrin.h>
-#    pragma intrinsic(_BitScanReverse,_BitScanReverse64)
 
     {
         unsigned long Index;
@@ -688,8 +692,6 @@ Perl_my_ffs(PERL_UINTMAX_T word)
     return PERL_USE_FFS(word) - 1;
 
 #  else
-#    include <intrin.h>
-#    pragma intrinsic(_BitScanForward,_BitScanForward64)
 
     {
         unsigned long Index;
--- a/inline.h
+++ b/inline.h
@@ -581,6 +581,23 @@ Perl_is_utf8_invariant_string_loc(const U8* const s, STRLEN len, const U8 ** ep)
     return TRUE;
 }
 
+PERL_STATIC_INLINE unsigned
+Perl_single_1bit_pos(PERL_UINTMAX_T word)
+{
+    /* Given a word known to contain all zero bits except one 1 bit, find and
+     * return the 1's position: 0..63 */
+
+#ifdef PERL_CORE    /* macro not exported */
+    ASSUME(isPOWER_OF_2(word));
+#endif
+
+    /* The position of the only set bit in a word can be quickly calculated
+     * using deBruijn sequences.  See for example
+     * https://en.wikipedia.org/wiki/De_Bruijn_sequence */
+    return PL_deBruijn_bitpos_tab[(word * PERL_deBruijnMagic_)
+                                                    >> PERL_deBruijnShift_];
+}
+
 #ifndef EBCDIC
 
 PERL_STATIC_INLINE unsigned int
--- a/inline.h
+++ b/inline.h
@@ -598,6 +598,32 @@ Perl_single_1bit_pos(PERL_UINTMAX_T word)
                                                     >> PERL_deBruijnShift_];
 }
 
+PERL_STATIC_INLINE unsigned
+Perl_my_ffs(PERL_UINTMAX_T word)
+{
+    /* Find the position (0..63) of the least significant set bit in the input
+     * word */
+
+    ASSUME(word != 0);
+
+    /*  Isolate the lsb;
+     * https://stackoverflow.com/questions/757059/position-of-least-significant-bit-that-is-set
+     *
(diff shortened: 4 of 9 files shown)


sisyphus:
At https://github.com/Perl/perl5/pull/18944/commits/af269b6ec8fcea0bf613e0f2f650ead38333fa6d
we see the following :
```
+        unsigned long * Index;
+
+       PERL_USE_CLZ(&Index, word);
+
+        return Index;
```
and
```
+        unsigned long * Index;
+
+        PERL_USE_FFS(&Index, word);
+
+        return Index;
```
My expectation (unsubstantiated) is that "Index" should be declared in both places as "unsigned long", not "unsigned long *".
At least, I think that would be the case if mingw-w64 compilers were given access to that block of code.
(A brief check reveals that _BitScanForward() and _BitScanReverse() were faulty in mingw-w64 ports of gcc-4.7.0, but they seem to behave correctly from version 4.8.2 onwards.)

IIUC, this code will  currently be visible to recent MS Compilers only - and therefore won't get automatically tested as CI testing for these compilers has been removed.

AFAICT, with the mingw-w64 ports of gcc, the two \#pragma directives are unnecessary, but are happily (and silently) ignored anyway.

nwc10:
I built this on AIX, HP-UX and Solaris (vendor compilers) and it was fine. So sparc64 and ia64 aren't objecting (and ppc64 seems to be more lax than sparc64, so it doesn't really deserve a mention.)

I've not looked at the code, and doubt that I'll have time any time soon (trying to get through the backlog of everything else first. I hope that others can give this a careful check. I'm very glad that you've worked on this, as there might well be other uses for it.)

@craigberry Are you able to give this branch a "spin" on VMS?

khwilliamson:
@sisyphus your expectation was very substantiated.  But, we are testing modern Windows, and CI should have caught what you did.  I don't understand why not. @xenu, any ideas.

Anyway, now even mingw64 has passed with the fix that @sisyphus caught

sisyphus:
I've just built the khw-clz branch for 2 different configurations (ivsize =
4 and ivsize = 8) using mingw-w64.
In perl.h, I'm finding that, on both of those
configurations, PERL_UINTMAX_SIZE is 8 - because (I think) QUADKIND is
true, even when ivsize is 4.

That means that neither of the conditions PERL_UINTMAX_SIZE == INTSIZE and
PERL_UINTMAX_SIZE == LONGSIZE is ever true - because both LONGSIZE and
INTSIZE are *always* 4 on Windows.

I think that for windows, with both modern msvc and mingw-w64 (but not
mingw.org) we simply need to check IVSIZE.
If IVSIZE == 4 we need to call _BitScanReverse and _BitScanForward.
If IVSIZE == 8 we need to call _BitScanReverse64 and _BitScanForward64
The mingw.org builds of gcc don't provide any of those functions, AFAIK.

I'll have a play tomorrow night and see if I can come up with something a
bit more useful.
This strikes me as something that's worth persevering with.

khwilliamson:
On 6/30/21 7:01 AM, sisyphus wrote:
> I've just built the khw-clz branch for 2 different configurations (ivsize =
> 4 and ivsize = 8) using mingw-w64.
> In perl.h, I'm finding that, on both of those
> configurations, PERL_UINTMAX_SIZE is 8 - because (I think) QUADKIND is
> true, even when ivsize is 4.

This is called from code that uses the largest word on the machine, 
regardless of the IV size.  That code is used where otherwise one would 
have to look through a string byte by byte, in situations where we know 
that the special characteristics we are looking for can be done this 
way.  It yields approximately an 8-fold improvement on 64-bit boxes.

I believe your suggestion would break those uses.
> 
> That means that neither of the conditions PERL_UINTMAX_SIZE == INTSIZE and
> PERL_UINTMAX_SIZE == LONGSIZE is ever true - because both LONGSIZE and
> INTSIZE are *always* 4 on Windows.
> 
> I think that for windows, with both modern msvc and mingw-w64 (but not
> mingw.org) we simply need to check IVSIZE.
> If IVSIZE == 4 we need to call _BitScanReverse and _BitScanForward.
> If IVSIZE == 8 we need to call _BitScanReverse64 and _BitScanForward64
> The mingw.org builds of gcc don't provide any of those functions, AFAIK.

I'm thinking we need to look at the long long size on Windows boxes.

khwilliamson:
The word size disparity shows why there was no failure on Windows; it wasn't getting compiled.  Two commits later cause it to compile there (they need to be rebased in) so I'm confident the code is now getting executed.  The mingw CI was "cancelled"  I have no idea what that might mean

sisyphus:
On Mon, Jun 28, 2021 at 3:43 AM Karl Williamson ***@***.***>
wrote:

> This series of commits creates functions that return the least and most
> set bits in the largest word perl thinks the hardware is capable of, and
> uses them in all but one of the places in core that I am aware of could
> benefit. (Perhaps you know of other places.)
>

I'm not sure if you're aware of it, so I'll mention that this could also be
of benefit in the Core module List::Util.

In List::Util::uniqnum, for the case that IVSIZE == NVSIZE == 8, we need to
know if an  IV/UV  can be exactly represented as an NV.
That is, we need to determine whether there are more than 51 bits between
the least significant and most significant set bits of the IV/UV.
We currently answer this with some bit fiddling - see comments and code at
https://metacpan.org/release/PEVANS/Scalar-List-Utils-1.56/source/ListUtil.xs#L1493

Interestingly, I find that the existing L::U code does the job on Windows
as quickly as  using _BitScanReverse64() and _BitScanForward64().
But, of course, the latter method is far more readily digested.

The benchmark script I used is attached.

khwilliamson:
On 7/1/21 8:20 AM, sisyphus wrote:
> On Mon, Jun 28, 2021 at 3:43 AM Karl Williamson ***@***.***>
> wrote:
> 
>  > This series of commits creates functions that return the least and most
>  > set bits in the largest word perl thinks the hardware is capable of, and
>  > uses them in all but one of the places in core that I am aware of could
>  > benefit. (Perhaps you know of other places.)
>  >
> 
> I'm not sure if you're aware of it, so I'll mention that this could also be
> of benefit in the Core module List::Util.
> 
> In List::Util::uniqnum, for the case that IVSIZE == NVSIZE == 8, we need to
> know if an IV/UV can be exactly represented as an NV.
> That is, we need to determine whether there are more than 51 bits between
> the least significant and most significant set bits of the IV/UV.
> We currently answer this with some bit fiddling - see comments and code at
> https://metacpan.org/release/PEVANS/Scalar-List-Utils-1.56/source/ListUtil.xs#L1493
> 
> Interestingly, I find that the existing L::U code does the job on Windows
> as quickly as using _BitScanReverse64() and _BitScanForward64().
> But, of course, the latter method is far more readily digested.
> 
> The benchmark script I used is attached.
> 
> —
>

Maybe the BitScan code isn't getting inlined to a single instruction.

Interestingly, I note that the L:U code makes an assumption that isn't 
necessarily valid.  It assumes it knows how the machine represents the 
negative of a number, namely by complementing and adding 1.  This is 2's 
complement arithmetic.  According to Wikipedia, there are still some 1's 
complement machines out there, and this code would fail on them.

*     https://www.perlmonks.org/?node_id=11113490                       *
* First, identify the lowest set bit and assign its value to an IV.     *
* Note that this value will always be > 0, and always a power of 2.     */
   IV lowest_set = iv & -iv;

My PR doesn't make this assumption.  Instead it complements and adds 1 
explicitly

nwc10:
> According to Wikipedia, there are still some 1's complement machines out there, and this code would fail on them.

It's ages since we've seen any feedback from any system that suggested it was either 1's complement or sign and magnitude. I *think* that there was one, once, over a decade ago, which seemed to be finding bugs in our code (that was already assuming 2s complement), in such a way that I thought it might have been sign and magnitude. But I forget when or what this was. I have some niggling memory that it might have been an Amdahl mainframe, but I've no idea if that is plausible. WIkipedia doesn't suggest anything about this, and I thought that Amdahl's goal was to be compatible with (but cheaper than) IBM, so this seems unlikely.

xenu:
Writing portable code of this kind is a really painful experience, so it's not surprising I found a few issues:

1. I find the code really hard to follow:
- Macros take different arguments depending on a platform.
- I know this doesn't actually happen, but it my_ffs() and my_msbit_pos() can potentially call each other through Perl_single_1bit_pos()
- my_ffs() name is misleading because it behaves differently than ffs (it's 0-indexed).
- PERL_USE_(CLZ|FFS) name IMHO implies that it's a flag, not a function-like macro.

2. MinGW (gcc) shouldn't use Visual C++ intrinsics. They work, but it's a barely supported emulation. It's better to use the native __builtin_xxx functions, just like on unixes.

3. I'm not a fan of using PERL_UINTMAX_T. I don't see what it gains us over UV. In practice, it differs only on 32-bit platforms compiled without 64-bit ints (a pretty rare configuration), but in theory, it could be troublesome. I mean, imagine if gcc suddenly decided to actually follow the standard and they defined uintmax_t as a synonym for unsigned __int128_t. That would completely break our code.

4. As you can see on [godbolt](https://godbolt.org/z/foG5xc58s), for some reason gcc doesn't optimize my_msbit_pos() properly:
```C
unsigned
my_msbit_pos(unsigned long long word)
{
    /* (PERL_UINTMAX_SIZE * CHARBITS) - 1 - PERL_USE_CLZ(word);*/
    return 63 - __builtin_clzll(word);
}
```
compiles to
```assembly
my_msbit_pos(unsigned long long):
        bsr     rdi, rdi
        mov     eax, 63
        xor     rdi, 63
        sub     eax, edi
        ret
```
but if you replace `63 -` with `63 ^` you get much more optimal:
```assembly
my_msbit_pos(unsigned long long):
        bsr     rax, rdi
        ret
```

---

[Here is my proposed implementation](https://github.com/Perl/perl5/files/6750603/scan.txt) of the bitscanning functions that fixes the above issues. It also adds `PERL_GCC_VERSION_GE()` macro because I was really tired of manually writing gcc version checks.

xenu:
>It's ages since we've seen any feedback from any system that suggested it was either 1's complement or sign and magnitude.

Unisys has a thing for weird integer representations. OS 2200 is using 1's complement and MCP is using signed magnitude representation. Both are still actively developed! I don't think they can run Perl, though.

craigberry:
> @craigberry Are you able to give this branch a "spin" on VMS?

Results [here](https://www.nntp.perl.org/group/perl.daily-build.reports/2021/07/msg269483.html)

I think all those compiler messages and test failures were already present.  Sorry for the slow reply, @nwc10. This didn't show up in my GitHub feed for some reason.

khwilliamson:
On 7/1/21 2:34 PM, xenu wrote:
> Writing portable code of this kind is a really painful experience, so 
> it's not surprising I found a few issues:
> 
>  1. I find the code really hard to follow:
> 
>   * Macros take different arguments depending on a platform.

Agreed.

>   * I know this doesn't actually happen, but it my_ffs() and
>     my_msbit_pos() can potentially call each other through
>     Perl_single_1bit_pos()

It can't actually happen because of conditional compilation.  See below 
for why I did this.

>   * my_ffs() name is misleading because it behaves differently than ffs
>     (it's 0-indexed).

Agreed

>   * PERL_USE_(CLZ|FFS) name IMHO implies that it's a flag, not a
>     function-like macro.

I can understand how you might think this.
> 
>  2.
> 
>     MinGW (gcc) shouldn't use Visual C++ intrinsics. They work, but it's
>     a barely supported emulation. It's better to use the native
>     __builtin_xxx functions, just like on unixes.

Fine with me.
> 
>  3.
> 
>     I'm not a fan of using PERL_UINTMAX_T. I don't see what it gains us
>     over UV. In practice, it differs only on 32-bit platforms compiled
>     without 64-bit ints (a pretty rare configuration), but in theory, it
>     could be troublesome. I mean, imagine if gcc suddenly decided to
>     actually follow the standard and they defined uintmax_t as a synonym
>     for unsigned __int128_t. That would completely break our code.

I don't agree here.  I regularly use such boxes for testing, and this 
would harm performance on them that has been in existence for several 
releases, apparently because of a theoretical future possibility.  And 
should that happen, PERL_UNITMAX_T could be redefined so as to not break 
things.

There are various places, such as some pattern matching constructs, 
whose speed would be doubled if we had 16 byte words
> 
>  4.
> 
>     As you can see on godbolt <https://godbolt.org/z/foG5xc58s>, for
>     some reason gcc doesn't optimize my_msbit_pos() properly:
> 
> unsigned
> my_msbit_pos(unsigned  long  long  word)
> {
>      /* (PERL_UINTMAX_SIZE * CHARBITS) - 1 - PERL_USE_CLZ(word);*/
>      return  63  -__builtin_clzll(word);
> }
> 
> compiles to
> 
> my_msbit_pos(unsigned long long):
> bsrrdi,rdi
> moveax,63
> xorrdi,63
> subeax,edi
> ret
> 
> but if you replace |63 -| with |63 ^| you get much more optimal:
> 
> my_msbit_pos(unsigned long long):
> bsrrax,rdi
> ret
> 

Fine, but it has to be documented, as this was not obvious to me, and 
I'm sure not to others as well

> 
> Here is my proposed implementation 
> <https://github.com/Perl/perl5/files/6750603/scan.txt> of the 
> bitscanning functions that fixes the above issues. It also adds 
> |PERL_GCC_VERSION_GE()| macro because I was really tired of manually 
> writing gcc version checks.
> 
> —

I'm mostly fine with your implementation.  (I'm happy to get rid of the 
'my_' prefix in the names I proposed)


The GCC GE macro should be in a separate, first, commit and existing 
code converted to use it.

I disagree about using UV instead of PERL_UINTMAX.  Likely, we will need 
other people to weigh in on this.

The reason I had the functions appear to call each other is to give the 
best possible performance on platforms that have one or the other of the 
builtins.   My claim is that the deBruijn method should be used only if 
neither system builtin/function is available.  Your implementation 
sacrifices that.

For example, it supposedly is common for hardware to have a msb builtin 
and not one for lsb, because one can easily derive a performant version 
of the latter from the former.  Your implementation will instead use the 
deBruijn method in this case; a less desirable outcome.

khwilliamson:
I think I should do some clarification for people who haven't been following in exquisite detail.

There are times where it is advantageous to know the bit position in a word of either the most significant 1 bit, or the least significant one.   Some machines have a single instruction that returns one or the other; some have single instructions for both.  And some compilers can generate code to use those single instructions; many of those need special nudging to do this.  There is also a POSIX function ffs() that will return the lsb of an int.  Presumably that is implemented as fast as the libc writer knew how.

This PR is all about getting this information as efficiently as possible, doing any necessary nudging of the compiler.  xenu came up with a number of improvements to the original PR, but there are two things that need addressing.

On machines that have just one or the other operation, it is possible to derive the other from the implemented one by just a few operations using shifting and the like.  The original PR did this; xenu's revision does not.

And we disagree on whether the operations should be tied to the size of a UV.  We are talking about machine words here, and the original PR uses the widest word available on the platform, even if a UV is not defined to be that wide.  xenu's implementation uses the UV size.

khwilliamson:
Alternative #18997 was merged instead of this

PR #18945, opened 2021-06-28
State that the -s command line switch is discouraged; mention Getopt:Long

nwc10:
We really should discourage the use of -s.

Also clarify that the variables are created in the main package. It's not obvious whether this is obvious - implying that it needs saying.

### DIFF
--- a/pod/perlrun.pod
+++ b/pod/perlrun.pod
@@ -762,7 +762,7 @@ X<-s>
 enables rudimentary switch parsing for switches on the command
 line after the program name but before any filename arguments (or before
 an argument of B<-->).  Any switch found there is removed from @ARGV and sets the
-corresponding variable in the Perl program.  The following program
+corresponding variable in the Perl program, in the main package.  The following program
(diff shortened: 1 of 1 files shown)


neilb:
Could maybe extend the comment to:

> See `L<Getopt::Long>`, which is bundled with Perl, plus there are many other options on CPAN, such as Getopt::Long::Descriptive.

We could be extra helpful, and make that

    L<Getopt::Long::Descriptive|https://metacpan.org/pod/Getopt::Long::Descriptive>

But I don't know if the core doc tries to minimise the number of links like this. Seems safe though?

Otherwise, I think this is good.

nwc10:
Thanks for sanity checking this.

I think I'd prefer to keep it to just `Getopt::Long`. This is supposed to be a reference document on all the (gnarly) command line flags. and it feels to me better to avoid potential "decision paralysis"  by suggesting "do X, or Y, which is better but also needs action Z" versus, "just do X instead. You'll thank us later".

And I don't know about all the nuances of linking, but linking to a module in the core seems always to be done as a simple `L<...>` these days, so I assume that it works well.

PR #18950, opened 2021-06-29
Pod html refactoring 2 of 5

jkeenan:
This is the second of ~two~ five pull requests in support of https://github.com/Perl/perl5/issues/18894.  These p.r.s are intended to make the internals of Pod-Html less messy.  Ultimately Pod::Html will hold a series of methods calls on an object.  In this commit we start to gather together most of the loose lexical variables holding state into an object.  We also move auxiliary functions into a package of their own.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -4324,7 +4324,7 @@ ext/Pod-Html/bin/pod2html	Translator to turn pod into HTML
 ext/Pod-Html/corpus/perlpodspec-copy.pod
 ext/Pod-Html/corpus/perlvar-copy.pod
 ext/Pod-Html/lib/Pod/Html.pm	Convert POD data to HTML
-ext/Pod-Html/lib/Pod/Html/Auxiliary.pm	Helper functions for Pod-Html
+ext/Pod-Html/lib/Pod/Html/Util.pm	Helper functions for Pod-Html
 ext/Pod-Html/t/anchorify.t
 ext/Pod-Html/t/cache.pod
 ext/Pod-Html/t/cache.t
--- a/MANIFEST
+++ b/MANIFEST
@@ -4324,6 +4324,7 @@ ext/Pod-Html/bin/pod2html	Translator to turn pod into HTML
 ext/Pod-Html/corpus/perlpodspec-copy.pod
 ext/Pod-Html/corpus/perlvar-copy.pod
 ext/Pod-Html/lib/Pod/Html.pm	Convert POD data to HTML
+ext/Pod-Html/lib/Pod/Html/Auxiliary.pm	Helper functions for Pod-Html
 ext/Pod-Html/t/anchorify.t
 ext/Pod-Html/t/cache.pod
 ext/Pod-Html/t/cache.t
--- a/Makefile.SH
+++ b/Makefile.SH
@@ -1447,7 +1447,7 @@ _cleaner2:
 	-rmdir lib/TAP/Formatter/File lib/TAP/Formatter/Console
 	-rmdir lib/TAP/Formatter lib/TAP lib/Sys/Syslog lib/Sys lib/Sub
 	-rmdir lib/Search lib/Scalar lib/Pod/Text lib/Pod/Simple
-	-rmdir lib/Pod/Perldoc lib/PerlIO/via lib/PerlIO lib/Perl
+	-rmdir lib/Pod/Perldoc lib/Pod/Html lib/PerlIO/via lib/PerlIO lib/Perl
 	-rmdir lib/Parse/CPAN lib/Parse lib/Params lib/Net/FTP lib/Module/Load
 	-rmdir lib/Module/CoreList lib/Module lib/Memoize lib/Math/BigInt
 	-rmdir lib/Math/BigFloat lib/Math lib/MIME lib/Locale/Maketext
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -574,16 +574,16 @@ sub _save_page {
 sub write_file {
     my ($globals, $output) = @_;
     $globals->{Htmlfile} = "-" unless $globals->{Htmlfile}; # stdout
-    my $FHOUT;
+    my $fhout;
     if($globals->{Htmlfile} and $globals->{Htmlfile} ne '-') {
-        open $FHOUT, ">", $globals->{Htmlfile}
+        open $fhout, ">", $globals->{Htmlfile}
             or die "$0: cannot open $globals->{Htmlfile} file for output: $!\n";
     } else {
-        open $FHOUT, ">-";
+        open $fhout, ">-";
     }
-    binmode $FHOUT, ":utf8";
-    print $FHOUT $output;
-    close $FHOUT or die "Failed to close $globals->{Htmlfile}: $!";
+    binmode $fhout, ":utf8";
+    print $fhout $output;
+    close $fhout or die "Failed to close $globals->{Htmlfile}: $!";
     chmod 0644, $globals->{Htmlfile} unless $globals->{Htmlfile} eq '-';
 }
 
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -220,11 +220,11 @@ This program is distributed under the Artistic License.
 sub feed_tree_to_parser {
     my($parser, $tree) = @_;
     if(ref($tree) eq "") {
-	$parser->_handle_text($tree);
+        $parser->_handle_text($tree);
     } elsif(!($tree->[0] eq "X" && $parser->nix_X_codes)) {
-	$parser->_handle_element_start($tree->[0], $tree->[1]);
-	feed_tree_to_parser($parser, $_) foreach @{$tree}[2..$#$tree];
-	$parser->_handle_element_end($tree->[0]);
+        $parser->_handle_element_start($tree->[0], $tree->[1]);
+        feed_tree_to_parser($parser, $_) foreach @{$tree}[2..$#$tree];
+        $parser->_handle_element_end($tree->[0]);
     }
 }
 
@@ -437,19 +437,7 @@ HTMLFOOT
 
     feed_tree_to_parser($parser, $podtree);
 
-    # Write output to file
-    $globals->{Htmlfile} = "-" unless $globals->{Htmlfile}; # stdout
-    my $fhout;
-    if($globals->{Htmlfile} and $globals->{Htmlfile} ne '-') {
-        open $fhout, ">", $globals->{Htmlfile}
-            or die "$0: cannot open $globals->{Htmlfile} file for output: $!\n";
-    } else {
-        open $fhout, ">-";
-    }
-    binmode $fhout, ":utf8";
-    print $fhout $output;
-    close $fhout or die "Failed to close $globals->{Htmlfile}: $!";
-    chmod 0644, $globals->{Htmlfile} unless $globals->{Htmlfile} eq '-';
+    write_file($globals, $output);
 }
 
 sub get_cache {
@@ -544,6 +532,22 @@ sub _save_page {
 
     my ($file, $dir) = fileparse($modspec, qr/\.[^.]*/); # strip .ext
     $Pages{$modname} = $dir.$file;
+}
+
+sub write_file {
+    my ($globals, $output) = @_;
+    $globals->{Htmlfile} = "-" unless $globals->{Htmlfile}; # stdout
+    my $FHOUT;
+    if($globals->{Htmlfile} and $globals->{Htmlfile} ne '-') {
+        open $FHOUT, ">", $globals->{Htmlfile}
+            or die "$0: cannot open $globals->{Htmlfile} file for output: $!\n";
+    } else {
+        open $FHOUT, ">-";
+    }
+    binmode $FHOUT, ":utf8";
+    print $FHOUT $output;
+    close $FHOUT or die "Failed to close $globals->{Htmlfile}: $!";
+    chmod 0644, $globals->{Htmlfile} unless $globals->{Htmlfile} eq '-';
 }
 
 package Pod::Simple::XHTML::LocalPodLinks;
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -280,50 +280,10 @@ sub pod2html {
     # load or generate/cache %Pages
     unless (get_cache($globals)) {
         # generate %Pages
-        my $pwd = getcwd();
-        chdir($globals->{Podroot}) ||
-            die "$0: error changing to directory $globals->{Podroot}: $!\n";
-
-        # find all pod modules/pages in podpath, store in %Pages
-        # - callback used to remove Podroot and extension from each file
-        # - laborious to allow '.' in dirnames (e.g., /usr/share/perl/5.14.1)
-        Pod::Simple::Search->new->inc(0)->verbose($globals->{Verbose})->laborious(1)
-            ->callback(\&_save_page)->recurse($globals->{Recurse})->survey(@{$globals->{Podpath}});
-
-        chdir($pwd) || die "$0: error changing to directory $pwd: $!\n";
-
-        # cache the directory list for later use
-        warn "caching directories for later use\n" if $globals->{Verbose};
-        open my $cache, '>', $globals->{Dircache}
-            or die "$0: error open $globals->{Dircache} for writing: $!\n";
-
-        print $cache join(":", @{$globals->{Podpath}}) . "\n$globals->{Podroot}\n";
-        my $_updirs_only = ($globals->{Podroot} =~ /\.\./) && !($globals->{Podroot} =~ /[^\.\\\/]/);
-        foreach my $key (keys %Pages) {
-            if($_updirs_only) {
-              my $_dirlevel = $globals->{Podroot};
-              while($_dirlevel =~ /\.\./) {
-                $_dirlevel =~ s/\.\.//;
-                # Assume $Pages{$key} has '/' separators (html dir separators).
-                $Pages{$key} =~ s/^[\w\s\-\.]+\///;
-              }
-            }
-            print $cache "$key $Pages{$key}\n";
-        }
-        close $cache or die "error closing $globals->{Dircache}: $!";
+        %Pages = generate_cache($globals, \%Pages);
     }
 
-    my $input;
-    unless (@ARGV && $ARGV[0]) {
-        if ($globals->{Podfile} and $globals->{Podfile} ne '-') {
-            $input = $globals->{Podfile};
-        } else {
-            $input = '-'; # XXX: make a test case for this
-        }
-    } else {
-        $globals->{Podfile} = $ARGV[0];
-        $input = *ARGV;
-    }
+    my $input = identify_input($globals);
 
     my $podtree = parse_input_for_podtree($globals, $input);
     $globals->{Title} = set_Title_from_podtree($globals, $podtree);
@@ -417,6 +377,65 @@ sub refine_globals {
         $globals->{Htmlfileurl} = unixify($globals->{Htmlfile});
     }
     return $globals;
+}
+
+sub generate_cache {
+    my ($globals, $Pagesref) = @_;
+    my $pwd = getcwd();
+    chdir($globals->{Podroot}) ||
+        die "$0: error changing to directory $globals->{Podroot}: $!\n";
+
+    # find all pod modules/pages in podpath, store in %Pages
+    # - inc(0): do not prepend directories in @INC to search list;
+    #     limit search to those in @{$globals->{Podpath}}
+    # - verbose: report (via 'warn') what search is doing
+    # - laborious: to allow '.' in dirnames (e.g., /usr/share/perl/5.14.1)
+    # - callback: used to remove Podroot and extension from each file
+    # - recurse: go into subdirectories
+    # - survey: search for POD files in PodPath
+    my ($name2path, $path2name) = 
+        Pod::Simple::Search->new->inc(0)->verbose($globals->{Verbose})->laborious(1)
+        ->callback(\&_save_page)->recurse($globals->{Recurse})->survey(@{$globals->{Podpath}});
+    #print STDERR Data::Dumper::Dumper($name2path, $path2name) if ($globals->{Verbose});
+
+    chdir($pwd) || die "$0: error changing to directory $pwd: $!\n";
+
+    # cache the directory list for later use
+    warn "caching directories for later use\n" if $globals->{Verbose};
+    open my $cache, '>', $globals->{Dircache}
+        or die "$0: error open $globals->{Dircache} for writing: $!\n";
+
+    print $cache join(":", @{$globals->{Podpath}}) . "\n$globals->{Podroot}\n";
+    my $_updirs_only = ($globals->{Podroot} =~ /\.\./) && !($globals->{Podroot} =~ /[^\.\\\/]/);
+    foreach my $key (keys %{$Pagesref}) {
+        if($_updirs_only) {
+          my $_dirlevel = $globals->{Podroot};
+          while($_dirlevel =~ /\.\./) {
+            $_dirlevel =~ s/\.\.//;
+            # Assume $Pagesref->{$key} has '/' separators (html dir separators).
+            $Pagesref->{$key} =~ s/^[\w\s\-\.]+\///;
+          }
+        }
+        print $cache "$key $Pagesref->{$key}\n";
+    }
+    close $cache or die "error closing $globals->{Dircache}: $!";
+    return %{$Pagesref};
+}
+
+sub identify_input {
+    my $globals = shift;
+    my $input;
+    unless (@ARGV && $ARGV[0]) {
+        if ($globals->{Podfile} and $globals->{Podfile} ne '-') {
+            $input = $globals->{Podfile};
+        } else {
+            $input = '-'; # XXX: make a test case for this
+        }
+    } else {
+        $globals->{Podfile} = $ARGV[0];
+        $input = *ARGV;
+    }
+    return $input;
 }
 
 sub parse_input_for_podtree {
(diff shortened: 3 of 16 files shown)


ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -2,20 +2,27 @@ package Pod::Html;
 use strict;
 use Exporter 'import';
 
-our $VERSION = 1.28;
-our @EXPORT = qw(pod2html htmlify);
-our @EXPORT_OK = qw(anchorify relativize_url);
+our $VERSION = 1.29;
+eval $VERSION;
```
Grinnz: Preferred mechanism for this: `$VERSION =~ tr/_//d;`

ext/Pod-Html/t/lib/Testing.pm
```diff
@@ -2,11 +2,13 @@ package Testing;
 use 5.10.0;
 use warnings;
 require Exporter;
-our $VERSION = 1.26; # Let's keep this same as lib/Pod/Html.pm
+our $VERSION = 1.27_001; # Let's keep this same as lib/Pod/Html.pm
```
Grinnz, line 5: I'm not sure what this version is for exactly but this won't work as intended. Underscores in numeric literals don't become part of the value. Versions with underscores must be strings (and versions should always be strings anyway, but that's not your fault).

jkeenan: This was corrected in https://github.com/Perl/perl5/pull/18950/commits/0e756f8ee9402244f1edce6a0a4cb47e323f4121 shortly after the p.r. was filed.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > Preferred mechanism for this: `$VERSION =~ tr/_//d;`

That may be your preference, but it is evidently not the preference of people who up until now have contributed to the core distribution.
```
$ ack -l 'eval \$VERSION' cpan dist ext lib |sort
cpan/Compress-Raw-Bzip2/lib/Compress/Raw/Bzip2.pm
cpan/Compress-Raw-Zlib/lib/Compress/Raw/Zlib.pm
cpan/DB_File/DB_File.pm
cpan/ExtUtils-Install/lib/ExtUtils/Install.pm
cpan/ExtUtils-Install/lib/ExtUtils/Installed.pm
cpan/ExtUtils-Install/lib/ExtUtils/Packlist.pm
cpan/File-Fetch/lib/File/Fetch.pm
cpan/File-Path/lib/File/Path.pm
cpan/Filter-Util-Call/Call.pm
cpan/IO-Compress/lib/Compress/Zlib.pm
cpan/Module-Metadata/t/extract-version.t
cpan/Win32/Win32.pm
dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
dist/Thread-Queue/lib/Thread/Queue.pm
dist/Thread-Semaphore/lib/Thread/Semaphore.pm
dist/Time-HiRes/HiRes.pm
dist/threads-shared/lib/threads/shared.pm
dist/threads/lib/threads.pm
ext/Devel-Peek/Peek.pm
ext/Errno/Errno.pm
lib/Compress/Raw/Bzip2.pm
lib/Compress/Raw/Zlib.pm
lib/Compress/Zlib.pm
lib/DB_File.pm
lib/Devel/Peek.pm
lib/Errno.pm
lib/ExtUtils/Install.pm
lib/ExtUtils/Installed.pm
lib/ExtUtils/Packlist.pm
lib/ExtUtils/ParseXS.pm
lib/File/Fetch.pm
lib/File/Path.pm
lib/Filter/Util/Call.pm
lib/Thread.pm
lib/Thread/Queue.pm
lib/Thread/Semaphore.pm
lib/Time/HiRes.pm
lib/threads.pm
lib/threads/shared.pm
```
Since the usage presented is consistent with existing code, I'm going to leave it as it is.  If you want to make a case for your alternative, that should go in a different ticket.

Thank you very much.
Jim Keenan

MANIFEST
```diff
@@ -4324,6 +4324,7 @@ ext/Pod-Html/bin/pod2html	Translator to turn pod into HTML
 ext/Pod-Html/corpus/perlpodspec-copy.pod
 ext/Pod-Html/corpus/perlvar-copy.pod
 ext/Pod-Html/lib/Pod/Html.pm	Convert POD data to HTML
+ext/Pod-Html/lib/Pod/Html/Auxiliary.pm	Helper functions for Pod-Html
```
rjbs: Not a hill I'd die on, but I believe "Util" is a bit more conventional for this sort of thing.  Consider:
```
find . -name '*.pm' | grep Util.pm
```

...but really I'm not bothered, I was just very briefly forced to think, and I dislike thinking. 😉

Makefile.SH
```diff
@@ -1447,7 +1447,7 @@ _cleaner2:
 	-rmdir lib/TAP/Formatter/File lib/TAP/Formatter/Console
 	-rmdir lib/TAP/Formatter lib/TAP lib/Sys/Syslog lib/Sys lib/Sub
 	-rmdir lib/Search lib/Scalar lib/Pod/Text lib/Pod/Simple
-	-rmdir lib/Pod/Perldoc lib/PerlIO/via lib/PerlIO lib/Perl
+	-rmdir lib/Pod/Perldoc lib/Pod/Html lib/PerlIO/via lib/PerlIO lib/Perl
```
rjbs, line 1450: I was confused by this, but the commit message cleared it up sufficiently.  Thanks!

ext/Pod-Html/lib/Pod/Html.pm
rjbs: I have no quarrel with using `eval`, but I believe if you look at the other uses of eval for mangling `$VERSION` you will see this:
```
$VERSION = eval $VERSION;
```

Only evaluating the version will return a value that is then thrown away.  **You should assign the result back to `$VERSION`.**

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -2,20 +2,27 @@ package Pod::Html;
...
+our @ISA = qw(Exporter);
```
rjbs: I believe you were recently removing inheriting from Exporter from core code.  Possibly you plan to do that in a subsequent commit.  If not, though, I can't tell why this would be here.  In this case, the commit message does *not* clarify.  **I believe this line is a mistake** given that you're importing `import` from Exporter above.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
     if(ref($tree) eq "") {
-	$parser->_handle_text($tree);
+        $parser->_handle_text($tree);
     } elsif(!($tree->[0] eq "X" && $parser->nix_X_codes)) {
-	$parser->_handle_element_start($tree->[0], $tree->[1]);
-	feed_tree_to_parser($parser, $_) foreach @{$tree}[2..$#$tree];
-	$parser->_handle_element_end($tree->[0]);
+        $parser->_handle_element_start($tree->[0], $tree->[1]);
+        feed_tree_to_parser($parser, $_) foreach @{$tree}[2..$#$tree];
+        $parser->_handle_element_end($tree->[0]);
```
rjbs, line 226: Work that normalizes indentation will always make me smile.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
-
-my $Recurse;
-my $Quiet;
-my $Verbose;
-my $Doindex;
-
-my $Backlink;
 
-my($Title, $Header);
+my $Podroot;
```
rjbs, line 231: I see that there is also Podroot in %Globals.  Is this variable still required?  (It's difficult to tell from the reviewing interface, so I am rudely punting the question back to you.  If it's required, okay.  If it's now unused, let's delete it!

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
-                # Assume $Pages{$key} has '/' separators (html dir separators).
-                $Pages{$key} =~ s/^[\w\s\-\.]+\///;
-              }
-            }
-            print $cache "$key $Pages{$key}\n";
+sub generate_cache {
+    my ($globals, $Pagesref) = @_;
+    my $pwd = getcwd();
+    chdir($globals->{Podroot}) ||
+        die "$0: error changing to directory $globals->{Podroot}: $!\n";
```
rjbs, line 338: I have an extremely nitpicky point to raise.  Here, you write:
```
chdir($x) ||
    die "...";
```

Later, about twenty lines down, you write this:
```
open ...
  or die "...";
```

This means sometimes you're ending with the operator and sometime starting with the operator, when breaking at an operator.  This is inconsistent.  If I was king of the universe I would put only one of these as acceptable in the style guide (and it would be to put the operator on the second line, not first).  That said, making this coherent across all the code sounds like a slog that should not block moving everything else forward.  So, I raise this as something you might enjoy fixing (as I sometimes enjoy somewhat mind-numbing tasks) or might just prefer to ignore.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
+    # - inc(0): do not prepend directories in @INC to search list;
+    #     limit search to those in @{$globals->{Podpath}}
+    # - verbose: report (via 'warn') what search is doing
+    # - laborious: to allow '.' in dirnames (e.g., /usr/share/perl/5.14.1)
+    # - callback: used to remove Podroot and extension from each file
+    # - recurse: go into subdirectories
+    # - survey: search for POD files in PodPath
+    my ($name2path, $path2name) = 
+        Pod::Simple::Search->new->inc(0)->verbose($globals->{Verbose})->laborious(1)
+        ->callback(\&_save_page)->recurse($globals->{Recurse})->survey(@{$globals->{Podpath}});
```
rjbs, line 350: You're missing an opportunity to use postfix deref, Jim! 😉  (I'm not asking you to do anything here.)

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
+
+    # cache the directory list for later use
+    warn "caching directories for later use\n" if $globals->{Verbose};
+    open my $cache, '>', $globals->{Dircache}
+        or die "$0: error open $globals->{Dircache} for writing: $!\n";
+
+    print $cache join(":", @{$globals->{Podpath}}) . "\n$globals->{Podroot}\n";
+    my $_updirs_only = ($globals->{Podroot} =~ /\.\./) && !($globals->{Podroot} =~ /[^\.\\\/]/);
+    foreach my $key (keys %{$Pagesref}) {
+        if($_updirs_only) {
```
rjbs, line 363: Another tiny style nit:  The code is very inconsistent as to whether a space is put between opening parens and keywords like `if` or `while`.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -228,206 +220,236 @@ This program is distributed under the Artistic License.
...
+        print $cache "$key $Pagesref->{$key}\n";
     }
+    close $cache or die "error closing $globals->{Dircache}: $!";
+    return %{$Pagesref};
+}
 
+sub identify_input {
+    my $globals = shift;
     my $input;
     unless (@ARGV && $ARGV[0]) {
```
rjbs, line 380: It would be nice to extract all the code that looks at `@ARGV` up into some top layer and otherwise pass an arrayref around.  That sounds tedious, though, and like **it shouldn't block this work**.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -692,39 +566,26 @@ sub _save_page {
                                      File::Spec->canonpath($Podroot));
```
rjbs, line 565: This seems to be the only place that uses `$Podroot`.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -453,189 +475,73 @@ $block
...
+    $tests++ if ($globals->{Podroot} eq $_);
 
     # load the cache if its good
     if ($tests != 2) {
         close($cachefh);
         return 0;
     }
 
-    warn "loading directory cache\n" if $Verbose;
+    warn "loading directory cache\n" if $globals->{Verbose};
```
rjbs, line 543: Is your plan, in the end, that these subroutines should be methods on an object?  I ask because clearly all these `warn` calls predicated on Verbose should become calls to `$self->log` or something.  I know we're only on step 2 of 5 though!

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -692,39 +566,26 @@ sub _save_page {
...
-        if( $dirs[0] ) {
-            unshift @dirs, $vol;
-        }
-        else {
-            $dirs[0] = $vol;
-        }
+sub write_file {
+    my ($globals, $output) = @_;
+    $globals->{Htmlfile} = "-" unless $globals->{Htmlfile}; # stdout
+    my $FHOUT;
```
rjbs: Why is this named in all caps?  Convention (in the wild as well as in perlstyle) is to use all capital names for constants.  Sometimes this is expanded to globals.  **This variable is lexical in scope and duration, so should probably be lowercase.**  Possibly I have missed some reason.

ext/Pod-Html/lib/Pod/Html/Auxiliary.pm
```diff
@@ -0,0 +1,292 @@
...
+
+=head1 NAME
+
+Pod::Html::Auxiliary - helper functions for Pod-Html
+
+=head1 SUBROUTINES
+
+=head2 C<parse_command_line()>
+
+TK
```
rjbs: (We're gonna get this in some future commit?)

ext/Pod-Html/lib/Pod/Html/Auxiliary.pm
```diff
@@ -0,0 +1,292 @@
...
+               ? (File::Spec::Unix->curdir())
+               : File::Spec->splitdir($dirs);
+    if (defined($vol) && $vol) {
+        $vol =~ s/:$// if $^O eq 'VMS';
+        $vol = uc $vol if $^O eq 'MSWin32';
+
+        if( $dirs[0] ) {
+            unshift @dirs, $vol;
+        }
+        else {
```
rjbs: Sorry, another style nit: why is this not cuddled, when many/most others are?  (I will comment on the style items entirely at the end of my review.)

ext/Pod-Html/lib/Pod/Html/Auxiliary.pm
```diff
@@ -0,0 +1,292 @@
...
+
+=cut
+
+sub html_escape {
+    my $rest = $_[0];
+    $rest   =~ s/&/&amp;/g;
+    $rest   =~ s/</&lt;/g;
+    $rest   =~ s/>/&gt;/g;
+    $rest   =~ s/"/&quot;/g;
+    $rest =~ s/([[:^print:]])/sprintf("&#x%x;", ord($1))/aeg;
```
rjbs: All these operators line up but the last one, how sad!

ext/Pod-Html/lib/Pod/Html/Auxiliary.pm
```diff
@@ -0,0 +1,292 @@
...
+
+sub trim_leading_whitespace {
+    my ($para) = @_;
+
+    # Start by converting tabs to spaces
+    @$para = Text::Tabs::expand(@$para);
+
+    # Find the line with the least amount of indent, as that's our "base"
+    my @indent_levels = (sort(map { $_ =~ /^( *)./mg } @$para));
+    my $indent        = $indent_levels[0] || "";
```
rjbs: I find this mechanism confusing.  I suppose it relies on the fact that two spaces sorts before three spaces, which is not something that _I_ find obvious.  (I've never been good with less trivial string sorts, though.)  I wonder if this would be clearer with a comment, anyway.

```
# Extract leading runs of spaces from each line, then sort these runs.
# Shorter strings will sort before longer strings (when they're all one
# character), so we can just take the first one as the shortest indent
# prefix.  If there's none, default to an empty string, just to avoid
# uninitialized warnings.
my ($indent) = sort(map { $_ =~ /^( *)./mg } @$para);
$indent //= "";
```

ext/Pod-Html/t/lib/Testing.pm
```diff
@@ -587,6 +605,89 @@ sub _process_diff {
...
+            unless defined $args->{$k};
+    }
+    my $cwd = cwd();
+    my $cache = catfile($cwd, 'pod2htmd.tmp');
+    die("Could not locate file $cache") unless -f $cache;
+    die("Could not locate directory $args->{outdir}") unless -d $args->{outdir};
+    die("'run' element takes integer") unless $args->{run} =~ m/^\d+$/;
+
+    my @cachelines = ();
+    open my $IN, '<', $cache or die "Unable to open $cache for reading";
```
rjbs: (I echo my comments about capitalized variable names here.)

rjbs (changes requested):
Quite a lot of my comments are about style.  Because this commit has moved a lot of code (which is good!) a lot of it looks like new code, and I tend to provide "the style is not consistent" to new code.  But I know you didn't _write_ all this new code, so ignoring style nits to move the structural improvements forward makes sense.

On the other hand, I did make a few comments that should be addressed, especially about $Podroot and eval.

Mostly: looks great!

jkeenan:
On 7/4/21 3:38 PM, Ricardo Signes wrote:
> 
> Quite a lot of my comments are about style. Because this commit has 
> moved a lot of code (which is good!) a lot of it looks like new code, 
> and I tend to provide "the style is not consistent" to new code. But I 
> know you didn't /write/ all this new code, so ignoring style nits to 
> move the structural improvements forward makes sense.
> 
> On the other hand, I did make a few comments that should be addressed, 
> especially about $Podroot and eval.
> 
> Mostly: looks great!
> 

Thank you for this review.  I'm going to assign your comments into three 
buckets:

1. Things that I will deal with right now, in this pull request.

2. Things that I won't deal with in this pull request because they will 
be dealt with in one of the later, already planned pull requests.

3. Things that I won't deal with in this series of pull requests but 
which we should consider once the series is applied and we're starting 
from a cleaner, OO version of ext/Pod-Html/lib/Pod/Html.pm.  I will keep 
a running tab of such concerns at 
http://thenceforward.net/perl/misc/pod-html-remaining-issues.txt.

(Note that if you want a sneak preview of what the code will look like 
after the 5 planned pull requests are applied, go to 
https://github.com/jkeenan/perl5/tree/pod-html-oo-20210512-02/ext/Pod-Html.)

> [snip]
> 
> +ext/Pod-Html/lib/Pod/Html/Auxiliary.pm	Helper functions for Pod-Html
> 
> Not a hill I'd die on, but I believe "Util" is a bit more conventional 
> for this sort of thing. Consider:
> 

Renamed in updated p.r.

> 
> [snip]
> 
> +our $VERSION = 1.29;
> +eval $VERSION;
> 
> I have no quarrel with using |eval|, but I believe if you look at the 
> other uses of eval for mangling |$VERSION| you will see this:
> 
> |$VERSION = eval $VERSION; |
> 
> Only evaluating the version will return a value that is then thrown 
> away. *You should assign the result back to |$VERSION|.*
> 

Corrected in updated p.r.

> 
> [snip]

>   use Exporter 'import';
>   
> -our $VERSION = 1.28;
> -our @EXPORT = qw(pod2html htmlify);
> -our @EXPORT_OK = qw(anchorify relativize_url);
> +our $VERSION = 1.29;
> +eval $VERSION;
> +our @ISA = qw(Exporter);
> 
> I believe you were recently removing inheriting from Exporter from core 
> code. Possibly you plan to do that in a subsequent commit. If not, 
> though, I can't tell why this would be here. In this case, the commit 
> message does /not/ clarify. *I believe this line is a mistake* given 
> that you're importing |import| from Exporter above.
> 

Corrected in updated p.r.

> 
> [snip]

> In ext/Pod-Html/lib/Pod/Html.pm 
> <https://github.com/Perl/perl5/pull/18950#discussion_r663543899>:
> 
>>  
> -my($Title, $Header);
> +my $Podroot;
> 
> I see that there is also Podroot in %Globals. Is this variable still 
> required? (It's difficult to tell from the reviewing interface, so I am 
> rudely punting the question back to you. If it's required, okay. If it's 
> now unused, let's delete it!
> 

At this point, this $Podroot is still needed.  If you look at the 
current (v1.28) ext/Pod-Html/lib/Pod-Html.pm, you will see that this is 
used inside _save_page() at line 689.  At this point in the refactoring, 
I couldn't figure out how to handle this, as _save_page() is used as a 
callback inside that horrid chain of methods at line 327-8.  But this 
will be handled in pull request 5 of 5.

> 
> [snip]
> +    chdir($globals->{Podroot}) ||
> +        die "$0: error changing to directory $globals->{Podroot}: $!\n";
> 
> I have an extremely nitpicky point to raise. Here, you write:
> 
> |chdir($x) || die "..."; |
> 
> Later, about twenty lines down, you write this:
> 
> |open ... or die "..."; |
> 
> This means sometimes you're ending with the operator and sometime 
> starting with the operator, when breaking at an operator. This is 
> inconsistent. If I was king of the universe I would put only one of 
> these as acceptable in the style guide (and it would be to put the 
> operator on the second line, not first). That said, making this coherent 
> across all the code sounds like a slog that should not block moving 
> everything else forward. So, I raise this as something you might enjoy 
> fixing (as I sometimes enjoy somewhat mind-numbing tasks) or might just 
> prefer to ignore.
> 

This we file under: Pod-Html Nits the Picking of Which Will Be Deferred 
until after Pull Request 5 (P-HNtPoWWBDuaPR5)

> 
> [snip]
> 
> +    foreach my $key (keys %{$Pagesref}) {
> +        if($_updirs_only) {
> 
> Another tiny style nit: The code is very inconsistent as to whether a 
> space is put between opening parens and keywords like |if| or |while|.
> 

P-HNtPoWWBDuaPR5

> 
> In ext/Pod-Html/lib/Pod/Html.pm 
> <https://github.com/Perl/perl5/pull/18950#discussion_r663545308>:
> 
>>  
> +sub identify_input {
> +    my $globals = shift;
>       my $input;
>       unless (@ARGV && $ARGV[0]) {
> 
> It would be nice to extract all the code that looks at |@ARGV| up into 
> some top layer and otherwise pass an arrayref around. That sounds 
> tedious, though, and like *it shouldn't block this work*.
> 

I agree.  Similar points raised by oodler in earlier (off-list) review. 
  Let's open an issue once this series of pull requests is applied.

> 
> [snip]
>   
> -    warn "loading directory cache\n" if $Verbose;
> +    warn "loading directory cache\n" if $globals->{Verbose};
> 
> Is your plan, in the end, that these subroutines should be methods on an 
> object? 

Yes.  We will achieve that in pull request 5 of 5.

I ask because clearly all these |warn| calls predicated on
> Verbose should become calls to |$self->log| or something. I know we're 
> only on step 2 of 5 though!
> 

That wasn't part of my original plan, but makes sense.  Again, let's 
open an issue for it once we've got an OO module to look at.

> 
> [snip]

> +    my $FHOUT;
> 
> Why is this named in all caps? Convention (in the wild as well as in 
> perlstyle) is to use all capital names for constants. Sometimes this is 
> expanded to globals. *This variable is lexical in scope and duration, so 
> should probably be lowercase.* Possibly I have missed some reason.
> 

Changed in updated p.r. -- although I've both seen and written a lot of 
code that uppercases all filehandles, whether global or lexical.

> 
> [snip]
> +=head1 NAME
> +
> +Pod::Html::Auxiliary - helper functions for Pod-Html
> +
> +=head1 SUBROUTINES
> +
> +=head2 C<parse_command_line()>
> +
> +TK
> 
> (We're gonna get this in some future commit?)
> 

Yes.  At this point in the refactoring process the subroutines were not 
yet stable.

> 
> [snip]
> +    if (defined($vol) && $vol) {
> +        $vol =~ s/:$// if $^O eq 'VMS';
> +        $vol = uc $vol if $^O eq 'MSWin32';
> +
> +        if( $dirs[0] ) {
> +            unshift @dirs, $vol;
> +        }
> +        else {
> 
> Sorry, another style nit: why is this not cuddled, when many/most others 
> are? (I will comment on the style items entirely at the end of my review.)
> 

P-HNtPoWWBDuaPR5

> 
> [snip]

> +
> +sub html_escape {
> +    my $rest = $_[0];
> +    $rest   =~ s/&/&amp;/g;
> +    $rest   =~ s/</&lt;/g;
> +    $rest   =~ s/>/&gt;/g;
> +    $rest   =~ s/"/&quot;/g;
> +    $rest =~ s/([[:^print:]])/sprintf("&#x%x;", ord($1))/aeg;
> 
> All these operators line up but the last one, how sad!
> 

P-HNtPoWWBDuaPR5


> 
> In ext/Pod-Html/lib/Pod/Html/Auxiliary.pm 
> <https://github.com/Perl/perl5/pull/18950#discussion_r663547215>:
> 
>> +
> +Remove any level of indentation (spaces or tabs) from each code block
> +consistently.  Adapted from:
> +https://metacpan.org/source/HAARG/MetaCPAN-Pod-XHTML-0.002001/lib/Pod/Simple/Role/StripVerbatimIndent.pm
> +
> +=cut
> +
> +sub trim_leading_whitespace {
> +    my ($para) = @_;
> +
> +    # Start by converting tabs to spaces
> +    @$para = Text::Tabs::expand(@$para);
> +
> +    # Find the line with the least amount of indent, as that's our "base"
> +    my @indent_levels = (sort(map { $_ =~ /^( *)./mg } @$para));
> +    my $indent        = $indent_levels[0] || "";
> 
> I find this mechanism confusing. I suppose it relies on the fact that 
> two spaces sorts before three spaces, which is not something that /I/ 
> find obvious. (I've never been good with less trivial string sorts, 
> though.) I wonder if this would be clearer with a comment, anyway.
> 

This was already present when I began this refactoring, so ... 
P-HNtPoWWBDuaPR5

> |# Extract leading runs of spaces from each line, then sort these runs. 
> # Shorter strings will sort before longer strings (when they're all one 
> # character), so we can just take the first one as the shortest indent # 
> prefix. If there's none, default to an empty string, just to avoid # 
> uninitialized warnings. my ($indent) = sort(map { $_ =~ /^( *)./mg } 
> @$para); $indent //= ""; |
> 
> 
> [snip]
> +    my @cachelines = ();
> +    open my $IN, '<', $cache or die "Unable to open $cache for reading";
> 
> (I echo my comments about capitalized variable names here.)

As above.

Thank you very much.
Jim Keenan

rjbs (approved):

PR #18951, opened 2021-06-29
Reformatted Changes as per CPAN::Changes::Spec

neilb:
This means that tools can parse this, and in particular MetaCPAN
will list the changes when you look at the dist/release page for
Data-Dumper.


Grinnz:
This is conflicting with blead (seems to be based on a few versions ago) and the commit message should mention that it's regarding Data::Dumper.

PR #18952, opened 2021-06-29
Reformatted Data-Dumper's Changes file as per CPAN::Changes::Spec

neilb:
This means that tools can parse this, and in particular MetaCPAN
will list the changes when you look at the dist/release page for
Data-Dumper.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -938,7 +938,7 @@ Nathan Kurz                    <nate@valleytel.net>
 Nathan Torkington              <gnat@frii.com>
 Nathan Trapuzzano              <nbtrap@nbtrap.com>
 Neale Ferguson                 <neale@VMA.TABNSW.COM.AU>
-Neil Bowers                    <neil@bowers.com>
+Neil Bowers                    <neilb@neilb.org>
 Neil Watkiss                   <neil.watkiss@sophos.com>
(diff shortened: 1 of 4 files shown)


Grinnz:
podcheck failure does not say what failed, strangely. The other failure is missing AUTHORS entry

Grinnz:
I suspect this file needs to be removed from known_pod_issues: https://github.com/Perl/perl5/blob/blead/t/porting/known_pod_issues.dat#L372

neilb:
Updated, thanks:

* updated AUTHORS to have my new email address as primary
* updated Porting/checkAUTHORS.pl so the email address map in `__DATA__` knows my previous email address
* deleted the entry for `dist/Data-Dumper/Changes` in `t/porting/known_pod_issues.dat`

nwc10:
I cherry-picked this onto my branch https://github.com/nwc10/perl5/tree/Data-Dumper-2.181

It conflicted with my most recent Changes entry, so I fixed that. I also fixed the cosmetic missing newlines that we both spotted, and removed some trailing whitespace that `git diff --color` is very "in-your-face" about.

I'm assuming that this is good (in that, I'm going to build a tarball, test that, if good ship it, and merge my branch into blead)

nwc10:

Issue #18955, opened 2021-07-01
$) and $( should be IOKp but not IOK


FGasper:
```
> perl -MDevel::Peek -e'Dump( my $v = $) )'
SV = PVMG(0x7fe9b782e380) at 0x7fe9b7806ad0
  REFCNT = 1
  FLAGS = (IOK,POK,IsCOW,pIOK,pPOK)
  IV = 20
  NV = 0
  PV = 0x7fe9b6505760 "20 20 505 504 12 61 79 80 81 98 33 100 204 250 395 398 399"\0
  CUR = 58
  LEN = 64
  COW_REFCNT = 1
```
The fact that this is marked IOK tells XS modules (e.g., Sereal) that the IV value is legitimate. That’s _sort_ of true, but it’s inconsistent with Perl’s typical behaviour of invalidating IOK whenever the IV mismatches the PV, which is the case here.

Note that leaving the private flag (IOKp) in place will prevent not-a-number warnings.

FGasper:
This causes https://github.com/Sereal/Sereal/issues/263

atoomic:
note that it's coming from the `void)SvIOK_on(sv);	/* what a wonderful hack! */` setting the IOK in all flags
view https://github.com/Perl/perl5/blob/c683824536c9806b1b970459bf81690e006b59ae/mg.c#L1227

My understanding is the issue is when serializing the data.
We probably should balance pros/cons of having the IOK flag set there

toddr:
Bisecting, you can look at:
```
LD_LIBRARY_PATH=/root/projects/perl ./perl -Ilib -MDevel::Peek -e'Dump( my $v = $) )' 2>&1|egrep '\bIOK\b'
```

If it gives back IOK, then the bug is still there.

At which point you can bisect it to:

```
4bac9ae47b5ad7845a24e26b0e95609805de688a is the first bad commit
commit 4bac9ae47b5ad7845a24e26b0e95609805de688a
Author: Chip Salzenberg <chip@pobox.com>
Date:   Fri Jun 22 15:18:18 2012 -0700
    Magic flags harmonization.
    In restore_magic(), which is called after any magic processing, all of
    the public OK flags have been shifted into the private OK flags.  Thus
    the lack of an appropriate public OK flags was used to trigger both get
    magic and required conversions.  This scheme did not cover ROK, however,
    so all properly written code had to make sure mg_get was called the right
    number of times anyway.  Meanwhile the private OK flags gained a second
[...]

PR #18957, opened 2021-07-02
Only set IOKp, not IOK on $) and $(.

FGasper:
Issue #18955: This will prevent serializers from serializing these
variables as numbers (which loses the additional groups).

This restores behaviour from 5.16.

### DIFF
--- a/mg.c
+++ b/mg.c
@@ -1240,7 +1240,13 @@ Perl_magic_get(pTHX_ SV *sv, MAGIC *mg)
                 Safefree(gary);
             }
         }
-        (void)SvIOK_on(sv);	/* what a wonderful hack! */
+
+        /*
+            Set this to avoid warnings when the SV is used as a number.
+            Avoid setting the public IOK flag so that serializers will
+            use the PV.
+        */
+        (void)SvIOKp_on(sv);	/* what a wonderful hack! */
 #endif
         break;
     case '0':
--- a/t/op/groups.t
+++ b/t/op/groups.t
@@ -51,8 +51,22 @@ sub Test {
     my %basegroup = basegroups( $pwgid, $pwgnam );
     my @extracted_supplementary_groups = remove_basegroup( \ %basegroup, \ @extracted_groups );
 
-    plan 3;
+    plan 4;
 
+    {
+        my @warnings = do {
+            my @w;
+            local $SIG{'__WARN__'} = sub { push @w, @_ };
+
+            use warnings;
+            my $v = $( + 1;
+            $v = $) + 1;
+
+            @w;
+        };
+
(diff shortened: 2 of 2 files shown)


mg.c
```diff
@@ -1240,7 +1240,13 @@ Perl_magic_get(pTHX_ SV *sv, MAGIC *mg)
...
             }
         }
-        (void)SvIOK_on(sv);	/* what a wonderful hack! */
+
+        /*
+            Set this to avoid warnings when the SV is used as a number.
+            Avoid setting the public IOK flag so that serializers will
+            use the PV.
+        */
+        (void)SvIOKp_on(sv);	/* what a wonderful hack! */
```
atoomic, line 1249: maybe keep IOK on if `num_groups == 1` ?

FGasper: I believe that would still be inconsistent with how Perl usually implements PVIV, which is that as soon as the PV mismatches the IV the IOK flag goes off.

In this case `1 1` doesn’t match 1. It does insofar as the interpretation as a list of group IDs, but not at the level of the actual data value.

FGasper: That also wouldn’t fix the issue w/ serialization as it first appeared for us, which was that what JSON and Data::Dumper serialize as "$UID $UID" came out as $UID, which broke tests.

atoomic (approved):

tonycoz:
I see a test for IOKp not being set, but I don't see a test for what the change is actually trying to fix, that serializers treat the value as an integer.

JSON::PP and Data::Dumper are included in core, so either could be used to test it (in op/groups.t)

FGasper:
@tonycoz The only serializations this affects are binary serializations since text-based serializations will (probably?) just use the PV as long as it exists. So neither JSON nor Data::Dumper is likely to be useful in testing this.

Sereal was where I first noticed the problem, but it also affects CBOR::Free. (CBOR::XS isn’t affected because it always prefers the PV.)

Storable is also not affected.

nwc10:
I tested this with Sereal.

Before, I see:

```
perl  -MSereal=encode_sereal,decode_sereal -MData::Dumper -lwe 'my $x = encode_sereal([$(]); print Data::Dumper::Dumper(decode_sereal($x))'
$VAR1 = [
          1147
        ];

```

after I see

```
./perl -MSereal=encode_sereal,decode_sereal -MData::Dumper -lwe 'my $x = encode_sereal([$(]); print Data::Dumper::Dumper(decode_sereal($x))'  $VAR1 = [
          '1147 1147'
        ];

```

OK, so my groups aren't **that** exciting, but I think that this demonstrates that it changes behaviour to something more useful.

I rebased it and pushed it to blead.

PR #18958, opened 2021-07-05
IV vs PV serialisation

nwc10:
Some external serialised data formats such as JSON can distinguish between values that are numbers and values that are strings, and services that communicate in JSON often rely on this distinction being possible. Until now, Perl has a problem with reliably generating output that meets these expectations, because the internal state of `$answer` after these two code paths had been functionally identical:

    my $answer = 42;
    print "The answer is $answer\n";

    my $answer = "42";
    print "That doesn't look right\n"
        unless $answer == 6 * 9;

This branch

* changes `Perl_sv_2pv_flags` so that it does not set the `SVf_POK` flag when caching the string representation of an integer value
* updates the macros `SvPV` etc so that they continue to use the cached value directly (and hence do not start calling into `Perl_sv_2pv_flags` where they had not previously)
* updates `Perl_sv_setsv_flags` to copy the changed flag state accurately

This permits serealisers to reliably distinguish between values first written as numbers, and values first written as strings.

These changes exposed a bug in Data::Dumper (now fixed), where it was already failing to handle NVs and overloaded references for certain cases. All core tests still pass, along with the CPAN modules `Sereal`, `JSON::XS`, `Cpanel::JSON::XS` and all their dependencies.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -425,6 +425,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
 to perform numerical or string comparison which is aware of operator
 overloading.
 
+=item *
+
+Reading the string form of an integer value no longer sets the flag C<SVf_POK>.
+The string form is still cached internally, and still re-read directly by the
+macros C<SvPV(sv)> I<etc> (inline, without calling a C function). XS code that
+already calls the APIs to get values will not be affected by this change. XS
+code that accesses flags directly instead of using API calls to express its
+intent I<might> break, but such code likely is already buggy if passed some
+other values, such as floating point values or objects with string overloading.
+
+This small change permits code (such as JSON serializers) to reliably determine
+between
+
+=over 4
+
+=item *
+
+a value that was initially B<written> as an integer, but then B<read> as a string
+
+    my $answer = 42;
+    print "The answer is $answer\n";
+
+=item *
+
+that same value that was initially B<written> as a string, but then B<read> as an integer
+
+    my $answer = "42";
+    print "That doesn't look right\n"
+        unless $answer == 6 * 9;
+
+=back
+
+For the first case (originally written as an integer), we now have:
+
+    use Devel::Peek;
+    my $answer = 42;
+    Dump ($answer);
+    my $void = "$answer";
+    print STDERR "\n";
+    Dump($answer)
+    
+    
+    SV = IV(0x562538925778) at 0x562538925788
+      REFCNT = 1
+      FLAGS = (IOK,pIOK)
+      IV = 42
+    
+    SV = PVIV(0x5625389263c0) at 0x562538925788
+      REFCNT = 1
+      FLAGS = (IOK,pIOK,pPOK)
+      IV = 42
+      PV = 0x562538919b50 "42"\0
+      CUR = 2
+      LEN = 10
+
+For the second (originally written as a string), we now have:
+
+    use Devel::Peek;
+    my $answer = "42";
+    Dump ($answer);
+    my $void = $answer == 6 * 9;
+    print STDERR "\n";
+    Dump($answer)'
+    
+    
+    SV = PV(0x5586ffe9bfb0) at 0x5586ffec0788
+      REFCNT = 1
+      FLAGS = (POK,IsCOW,pPOK)
+      PV = 0x5586ffee7fd0 "42"\0
+      CUR = 2
+      LEN = 10
+      COW_REFCNT = 1
+    
+    SV = PVIV(0x5586ffec13c0) at 0x5586ffec0788
+      REFCNT = 1
+      FLAGS = (IOK,POK,IsCOW,pIOK,pPOK)
+      IV = 42
+      PV = 0x5586ffee7fd0 "42"\0
+      CUR = 2
+      LEN = 10
+      COW_REFCNT = 1
+
+(One can't rely on the presence or absence of the flag C<SVf_IsCOW> to
+determine the history of operations on a scalar.)
+
+Previously both cases would be indistinguishable, with all 4 flags set:
+
+    SV = PVIV(0x55d4d62edaf0) at 0x55d4d62f0930
+      REFCNT = 1
+      FLAGS = (IOK,POK,pIOK,pPOK)
+      IV = 42
+      PV = 0x55d4d62e1740 "42"\0
+      CUR = 2
+      LEN = 10
+
+(and possibly C<SVf_IsCOW>, but not always)
+
+This now means that if XS code I<really> needs to determine which form a value
+was first written as, it should implement logic roughly
+
+    if (flags & SVf_IOK|SVf_NOK) && !(flags & SVf_POK)
+        serialize as number
+    else if (flags & SVf_POK)
+        serialize as string
+    else
+        the existing guesswork ...
+
+Note that this doesn't cover "dualvars" - scalars that report different
+values when asked for their string form or number form (such as C<$!>).
+Most serialization formats cannot represent such duplicity.
+
+I<The existing guesswork> remains because as well as dualvars, values might
+be C<undef>, references, overloaded references, typeglobs and other things that
+Perl itself can represent but do not map one-to-one into external formats, so
+need some amount of approximation or encapsulation.
+
+=item *
+
+Memory for hash iterator state (C<struct xpvhv_aux>) is now allocated as part
+of the hash body, instead of as part of the block of memory allocated for the
+main hash array.
+
+Nothing else changes - memory for this structure is still allocated only when
+needed, is still accessed via the C<HvAUX()> macro, and the macro should only
+be called when C<SvOOK()> is true. Hashes are still always of type C<SVt_PVHV>,
+hash bodies are still allocated from arenas, and the address of the hash
+doesn't change, because the address is the pointer to the head structure, which
+never moves.
+
+Internally, a second arena (the arena with index 1) is used to allocate larger
+bodies with space for C<struct xpvhv_aux>, the body "upgraded", and the "head"
+structure updated to reflect this (much the same way that the bodies of scalars
+are upgraded). We already re-purpose arenas - arena with index 0 is used for
+C<HE *>s.
+
+None of this affects documented public XS interfaces. The only code changes are
+in F<hv.c> and F<sv.c>. As the rest of the core itself uses these macros but
+needed no changes, likely no code on CPAN will be affected either.
+
 =back
 
 =head1 Selected Bug Fixes
--- a/pod/perlguts.pod
+++ b/pod/perlguts.pod
@@ -307,6 +307,30 @@ in it, you can use the following macros to check the type of SV you have.
     SvNOK(SV*)
     SvPOK(SV*)
 
+Be aware that retrieving the numeric value of an SV can set IOK or NOK
+on that SV, even when the SV started as a string.  Prior to Perl
+5.36.0 retrieving the string value of an integer could set POK, but
+this can no longer occur.  From 5.36.0 this can be used to distinguish
+the original representation of an SV and is intended to make life
+simpler for serializers:
+
+    /* references handled elsewhere */
+    if (SvIsBOOL(sv)) {
+        /* originally boolean */
+        ...
+    }
+    else if (SvPOK(sv)) {
+        /* originally a string */
+        ...
+    }
+    else if (SvNIOK(sv)) {
+        /* originally numeric */
+        ...
+    }
+    else {
+        /* something special or undef */
+    }
+
 You can get and set the current length of the string stored in an SV with
 the following macros:
 
--- a/sv.c
+++ b/sv.c
@@ -4892,6 +4892,12 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
             SvIV_set(dsv, SvIVX(ssv));
             if (sflags & SVf_IVisUV)
                 SvIsUV_on(dsv);
+            if ((sflags & SVf_IOK) && !(sflags & SVf_POK)) {
+                /* Source was SVf_IOK|SVp_IOK|SVp_POK but not SVf_POK, meaning
+                   a value set as an integer and later stringified. So mark
+                   destination the same: */
+                SvFLAGS(dsv) &= ~SVf_POK;
+            }
         }
         SvFLAGS(dsv) |= sflags & (SVf_IOK|SVp_IOK|SVf_NOK|SVp_NOK|SVf_UTF8);
         {
--- a/sv.c
+++ b/sv.c
@@ -3316,11 +3316,14 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
            where the value has subsequently been used in the other sense
            and had a value cached.
            This (somewhat) hack means that we retain the cached stringification,
-           but the existing SvPV() macros end up entering this function (which
-           returns the cached value) instead of using it directly inline.
-           However, the result is that if a value is SVf_IOK|SVf_POK then it
+           but don't set SVf_POK. Hence if a value is SVf_IOK|SVf_POK then it
            originated as "42", whereas if it's SVf_IOK then it originated as 42.
-           (ignore SVp_IOK and SVp_POK) */
+           (ignore SVp_IOK and SVp_POK)
+           The SvPV macros are now updated to recognise this specific case
+           (and that there isn't overloading or magic that could alter the
+           cached value) and so return the cached value immediately without
+           re-entering this function, getting back here to this block of code,
+           and repeating the same conversion. */
         SvPOKp_on(sv);
     }
     else if (SvNOK(sv)) {
--- a/sv.c
+++ b/sv.c
@@ -3309,7 +3309,19 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
         Move(ptr, s, len, char);
         s += len;
         *s = '\0';
-        SvPOK_on(sv);
+        /* We used to call SvPOK_on(). Whilst this is fine for (most) Perl code,
+           it means that after this stringification is cached, there is no way
+           to distinguish between values originally assigned as $a = 42; and
+           $a = "42"; (or results of string operators vs numeric operators)
+           where the value has subsequently been used in the other sense
+           and had a value cached.
+           This (somewhat) hack means that we retain the cached stringification,
+           but the existing SvPV() macros end up entering this function (which
+           returns the cached value) instead of using it directly inline.
+           However, the result is that if a value is SVf_IOK|SVf_POK then it
+           originated as "42", whereas if it's SVf_IOK then it originated as 42.
+           (ignore SVp_IOK and SVp_POK) */
+        SvPOKp_on(sv);
     }
     else if (SvNOK(sv)) {
         if (SvTYPE(sv) < SVt_PVNV)
--- a/sv.h
+++ b/sv.h
@@ -1859,19 +1859,30 @@ scalar.
 #define SvPV_const(sv, len)   SvPV_flags_const(sv, len, SV_GMAGIC)
 #define SvPV_mutable(sv, len) SvPV_flags_mutable(sv, len, SV_GMAGIC)
 
+/* This test is "is there a cached PV that we can use directly?"
+ * We can if
+ * a) SVf_POK is true and there's definitely no get magic on the scalar
+ * b) SVp_POK is true, there's no get magic, and we know that the cached PV
+ *    came from an IV conversion.
+ * For the latter case, we don't set SVf_POK so that we can distinguish whether
+ * the value originated as a string or as an integer, before we cached the
+ * second representation. */
+#define SvPOK_or_cached_IV(sv) \
+    (((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK) || ((SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK)))
+
 #define SvPV_flags(sv, len, flags) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? ((len = SvCUR(sv)), SvPVX(sv)) : sv_2pv_flags(sv, &len, flags))
 #define SvPV_flags_const(sv, len, flags) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? ((len = SvCUR(sv)), SvPVX_const(sv)) : \
      (const char*) sv_2pv_flags(sv, &len, (flags|SV_CONST_RETURN)))
 #define SvPV_flags_const_nolen(sv, flags) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? SvPVX_const(sv) : \
      (const char*) sv_2pv_flags(sv, 0, (flags|SV_CONST_RETURN)))
 #define SvPV_flags_mutable(sv, len, flags) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? ((len = SvCUR(sv)), SvPVX_mutable(sv)) : \
      sv_2pv_flags(sv, &len, (flags|SV_MUTABLE_RETURN)))
 
@@ -1896,16 +1907,16 @@ scalar.
      : sv_pvn_force_flags(sv, &len, flags|SV_MUTABLE_RETURN))
 
 #define SvPV_nolen(sv) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? SvPVX(sv) : sv_2pv_flags(sv, 0, SV_GMAGIC))
 
 /* "_nomg" in these defines means no mg_get() */
 #define SvPV_nomg_nolen(sv) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? SvPVX(sv) : sv_2pv_flags(sv, 0, 0))
 
 #define SvPV_nolen_const(sv) \
-    (SvPOK_nog(sv) \
+    (SvPOK_or_cached_IV(sv) \
      ? SvPVX_const(sv) : sv_2pv_flags(sv, 0, SV_GMAGIC|SV_CONST_RETURN))
 
 #define SvPV_nomg(sv, len) SvPV_flags(sv, len, 0)
--- a/MANIFEST
+++ b/MANIFEST
@@ -5995,6 +5995,7 @@ t/op/sub.t			See if subroutines work
 t/op/sub_lval.t			See if lvalue subroutines work
 t/op/substr.t			See if substr works
 t/op/substr_thr.t		See if substr works in another thread
+t/op/svflags.t			See if POK is set as expected.
 t/op/svleak.pl			Test file for svleak.t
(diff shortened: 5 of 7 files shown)


haarg:
Sub::Quote, JSON::Tiny, and Mojo::JSON also look at the flags (indirectly) to decide how to serialize scalars. They all pass their tests with this branch.

leonerd:
I've raised this PR for further discussion on p5p@
https://www.nntp.perl.org/group/perl.perl5.porters/2021/07/msg260916.html

FGasper:
The intent to preserve initial-stringiness can’t be achieved without making numification not set IOK/NOK.

Consider (on upstream):
```
perl -MCpanel::JSON::XS -e'my $x = "6"; my $y = $x + 0; print encode_json([$x])
[6]
```
The `$x + 0` sets IOK, which causes JSON::XS to output a number.

Basically, in order to preserve initial number/string it seems like we have to:
- prevent stringification from setting POK (done here)
- prevent numification from setting IOK (_not_ done here)

Per @Grinnz via IRC:
>  JSON::XS prefers the string value, JSON::PP and Mojo::JSON and Cpanel::JSON::XS prefer the number value because accidentally successfully numifying a stringy number is less of a false positive

Sereal also prefers the number, as does CBOR::Free. CBOR::XS prefers the string.

leonerd:
@FGasper Sure - we haven't claimed this PR is in any way *sufficient* for making all code able to distinguish. It's just a good start. It helps in some situations and - we hope - doesn't make anything else worse. There are of course many follow-ups that can be done afterwards.

ap:
Just to establish a destination for this journey, I believe it should ideally be the case that after

````perl
my $x = "6";
my $y = $x + 36;
````

the flags on `$x` would be POK=on IOK=off and on `$y` POK=off IOK=on, but `$x` would also have pIOK=on (assuming my understanding of the flags is correct…).

Basically:

* String operators always yield string values, and mathematical operators numeric values
* Intermediate conversions are cached and kept, but they don’t affect the “official” “type” of the scalar

I must add that I have no sense of the feasibility of getting there without too much breakage, but is there consensus that this is the platonic ideal?

**Comment by haarg** [comment]
> The intent to preserve initial-stringiness can’t be achieved without making numification not set IOK/NOK.
> 
> Consider (on upstream):
> 
> ```
> perl -MCpanel::JSON::XS -e'my $x = "6"; my $y = $x + 0; print encode_json([$x])
> [6]
> ```

This PR on its own is not intended to fix how existing serialization code picks types, but to allow new code to be written that can make those choices accurately. Existing serializers need to cope with how perl currently sets its flags, so they need to use heuristics. With this PR merged, any value with POK set is guaranteed to have started as a string. So a POK+IOK value, like in your example, is a string. No heuristic needed. Taking advantage of this would require updates to serializers, who would need to have different behavior on different perl versions.

It would also be worth trying to avoid setting IOK/NOK when numfying strings. That can be addressed separately though.

**Comment by iabyn** [comment]
On Wed, Jul 28, 2021 at 01:46:23AM -0700, Aristotle Pagaltzis wrote:
> Just to establish a destination for this journey, I believe it should ideally be the case that after
> 
> ````perl
> my $x = "6";
> my $y = $x + 36;
> ````
> 
> the flags on `$x` would be POK=on IOK=off and on `$y` POK=off IOK=on, but `$x` would also have pIOK=on (assuming my understanding of the flags is correct…).

Note that currently the IOK vs pIOK flags (and similarly for NOK/pNOK)
are used to distinguish between an accurate and an approximate conversion:

$ p -MDevel::Peek -e'my $x = "6";   my $y = int($x); Dump $x' 2>&1 | grep FLAGS
  FLAGS = (IOK,POK,IsCOW,pIOK,pPOK)

$ p -MDevel::Peek -e'my $x = "6.1"; my $y = int($x); Dump $x' 2>&1 | grep FLAGS
  FLAGS = (NOK,POK,IsCOW,pIOK,pNOK,pPOK)


-- 

**Comment by leonerd** [comment]
General ping here - What's the overall state?

The general feel I get from reading the comments is "this is fine, but not sufficient to give us proper type-intent distinction". Which is what we expected. I don't feel anyone has said "Don't do this", just people have said "This isn't enough".

Is that accurate? If so perhaps we can merge this and continue working

**Comment by iabyn** [comment]
On Tue, Aug 10, 2021 at 02:17:07AM -0700, Paul Evans wrote:
> General ping here - What's the overall state?
> 
> The general feel I get from reading the comments is "this is fine, but not sufficient to give us proper type-intent distinction". Which is what we expected. I don't feel anyone has said "Don't do this", just people have said "This isn't enough".
> 
> Is that accurate? If so perhaps we can merge this an
> 

I'm worried about the performance implications. Various hot places in the
core (such as pp_add()) take optimised paths in the presence of IOK, but
not pIOK.

-- 
The crew of the Enterprise encounter an alien life form which is
surprisingly neither humanoid nor made from pure energy.
    -- Things That Never Happen in "Star Trek" #22

**Comment by leonerd** [comment]
Relating to @iabyn's comment, and a few thoughts of my own on other PRs: Do we have a good way to benchmark these sorts of things? Primarily in terms of CPU time, but also perhaps in memory usage. I've had a few things I've wanted to change but wanted to find out what the {CPU, memory} impact of doing so would be.

**Comment by hvds** [comment]
> Relating to @iabyn's comment, and a few thoughts of my own on other PRs: Do we have a good way to benchmark these sorts of things? Primarily in terms of CPU time, but also perhaps in memory usage. I've had a few things I've wanted to change but wanted to find out what the {CPU, memory} impact of doing so would be.

`Porting/bench.pl` is very good for instruction-level timing on code snippets. AFAIR it'll show pagefaults, but won't directly show changes to memory usage.

**Comment by sisyphus** [comment]
On Wed, Aug 11, 2021 at 6:19 PM iabyn ***@***.***> wrote:

>
> I'm worried about the performance implications. Various hot places in the
> core (such as pp_add()) take optimised paths in the presence of IOK, but
> not pIOK.


Can this concern be alleviated by instead having the core base those
optimisation choices on the presence of the pIOK flag ?

(That's my fallback plan if I find that the loss of the IOK flag affects
those modules of mine that are looking for the IOK flag. In fact, I'm
thinking I should probably just make the switch to the pIOK flag anyway.)

Cheers,
Rob

**Comment by tonycoz** [comment]
> Can this concern be alleviated by instead having the core base those optimisation choices on the presence of the pIOK flag ?

No, pIOK can be set when IV isn't an accurate representation:
```
$ ./perl -Ilib -MDevel::Peek -e 'my $x = "1.1"; my $y = "a" x $x; Dump($x)'
SV = PVNV(0x559f54764340) at 0x559f54788578
  REFCNT = 1
  FLAGS = (NOK,POK,IsCOW,pIOK,pNOK,pPOK)
  IV = 1
  NV = 1.1
  PV = 0x559f547d0610 "1.1"\0
  CUR = 3
  LEN = 10
  COW_REFCNT = 1
```

**Comment by nwc10** [comment]
Rebased, and with better comments in the commit *Update SvPV() etc to avoid calling sv_2pv_flags() for cached IV strings*

**Comment by tonycoz** [comment]
> I'm worried about the performance implications. Various hot places in the core (such as pp_add()) take optimised paths in the presence of IOK, but not pIOK.

As far as I can see this doesn't change when the IOK or pIOK flags are set.

**Comment by nwc10** [comment]
> As far as I can see this doesn't change when the IOK or pIOK flags are set.

This was the intent. The macro code goes from

```
((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK)
? ... do stuff inline
: ... call a function
```

to

```
(((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK)
 || (SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK))
? ... do stuff inline
: ... call a function
```

ie just the flags test changes. If I understand superscalar CPUs correctly, likely this only adds one cycle to the execution, as the two flag tests can dispatch in parallel. Old and new code both depends on the result of (all the) flags tests before it can resolve the branch direction.

I don't know if it's even written down anywhere, but the basic "data structure with cached values and flags" design paradigm seems to be

* have a data structure with space to cache conversions and flags
* have macros which inline **small** code which can **quickly** test the flags for common cases, and where possible use cached values
* else call into a function for the slow path, which can hopefully cache values and set flags to permit fast paths in the future
* arrange all the various SV data structures so that the cache slots are at the same offsets from (stored) pointers, so that the reading code in the macros doesn't need to branch

The last one meaning that at times where the structure is officially

```
+-------+-------+-------+-------+------+
| stash | magic |  cur  |  len  |  IV  |
+-------+-------+-------+-------+------+
^
|
stored pointer

<----------- offset ---------->
```

and so pointer as shown is stored, and the code uses the offset shown, what might actually be allocated is this:

```
                               -+------+-
                                |  IV  |
                               -+------+-
                                ^
                                |
                                real pointer from malloc() or similar                              
^
|
stored pointer points over here
```

and the reading code is always "memory at *stored pointer* plus *offset*" independent of what is actually allocated.

**Comment by tonycoz** [comment]
> and so pointer as shown is stored, and the code uses the offset shown, what might actually be allocated is this:
> 
> ```
>                                -+------+-
>                                 |  IV  |
>                                -+------+-
>                                 ^
>                                 |
>                                 real pointer from malloc() or similar                              
> ^
> |
> stored pointer points over here
> ```
> 
> and the reading code is always "memory at _stored pointer_ plus _offset_" independent of what is actually allocated.

As far as I know this invokes undefined behaviour (a pointer pointing outside an object), but we've been doing it forever.

**Comment by tonycoz** [comment]
Summarizing the various objections:

|Issue|Resolution|
|-|-|
|@FGasper Basically, in order to preserve initial number/string it seems like we have to:... prevent numification from setting IOK (not done here).[1](https://github.com/Perl/perl5/pull/18958#issuecomment-884957815) Two other similar comments. [2](https://github.com/Perl/perl5/pull/18958#issuecomment-888130897) 3|this isn't an issue, the patch allows a serializer to distinguish between values that started as strings and values that started as numbers.|
|@iabyn I'm worried about the performance implications. Various hot places in the core (such as pp_add()) take optimised paths in the presence of IOK, but not pIOK. [1](https://github.com/Perl/perl5/pull/18958#issuecomment-896612066)|This PR doesn't change when IOK is set or unset.|
|(edit) Existing serializers aren't using the correct type|Existing serializers need to be updated|

To distinguish between "started as string" and "started as number" we can simply test:
```
    if (SvIsBOOL(sv)) {
        # it's boolean
    }
    if (SvPOK(sv)) {
        # started as string
    }
    else if (SvNIOK(sv)) {
        # started as number
    }
```

There's no need for extra flags changes.

Are there any other objections to this PR?

**Comment by xenu** [comment]
>I'm worried about the performance implications. Various hot places in the core (such as pp_add()) take optimised paths in the presence of IOK, but not pIOK.

>This PR doesn't change when IOK is set or unset.

POK vs POKp have the same problem. SvPV uses the cached string representation only when SVf_POK is set, so merging this PR will break it.

It seems that in practice "p" flags don't mean anything.

**Comment by xenu** [comment]
Sorry, disregard my comment. This branch changes the behaviour of SvPV(), so what I said isn't true anymore.

File: sv.h [hunk 1]
```diff
@@ -1859,19 +1859,30 @@ scalar.
...
+/* This test is "is there a cached PV that we can use directly?"
+ * We can if
+ * a) SVf_POK is true and there's definitely no get magic on the scalar
+ * b) SVp_POK is true, there's no get magic, and we know that the cached PV
+ *    came from an IV conversion.
+ * For the latter case, we don't set SVf_POK so that we can distinguish whether
+ * the value originated as a string or as an integer, before we cached the
+ * second representation. */
+#define SvPOK_or_cached_IV(sv) \
+    (((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK) || ((SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK)))
```
xenu, line 1871: In what cases testing for just `SVp_POK|SVs_GM == SVp_POK` wouldn't be enough?

I always thought that `SVp_[INP]OK` flags were supposed to mean "this scalar has a string/numeric representation" while `SV_[INP]OK` mean "this scalar has a **canonical** (lossless) string/numeric representation". If my understanding is correct, SvPV should have been testing for `SVp_POK` instead of `SVf_POK` even without the changes introduced in this PR.

FGasper:
@tonycoz
> the patch allows a serializer to distinguish between values that started as strings and values that started as numbers

How can that be claimed in light of this?
```
perl -MCpanel::JSON::XS -e'my $x = "6"; my $y = $x + 0; print encode_json([$x])
[6]
```
In order for Perl to distinguish the above value that started as a string, Perl has to stop setting the IOK … right?

-----
Or is the idea that anything PVIV should be considered a string, and serializers will all just update if Perl is 5.36+?

haarg:
@FGasper Cpanel::JSON::XS will need to be updated to take advantage of this PR. This PR on its own is not intended to fix existing serializers. See @tonycoz's last comment for how they can accurately check the type after this PR.

sv.h
tonycoz: `SvPV()` has been testing only `SVf_POK` for a very long time, since 463ee0b2acb in 1993, if we do change it, I don't think it belongs in this PR.

karenetheridge:
If someone has this branch running somewhere, a test run against JSON::Schema::Modern would be appreciated, as this distribution peeks at the PV, IV and NV flags for deriving "true" string/numeric properties. thanks!

demerphq:
On Thu, 27 Jan 2022, 10:55 Karen Etheridge, ***@***.***>
wrote:

> If someone has this branch running somewhere, a test run against
> JSON::Schema::Modern would be appreciated, as this distribution peeks at
> the PV, IV and NV flags for deriving "true" string/numeric properties.
> thanks
>


FWIW, that never worked reliably. A classic case is "0e0" another is " 0 ",
another is "0000". Variants thereof are also problematic. Sereal had to
address many of these cases with string inspection. IMO the change proposal
can only improve things.

Yves

>

nwc10:
I rebased, resolved conflicts and added two commits @tonycoz offered on a PR to my branch on my fork.

"It's not easy being green"

We're good?

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+
+Reading the string form of an integer value no longer sets the flag C<SVf_POK>.
+The string form is still cached internally, and still re-read directly by the
+macros C<SvPV(sv)> I<etc> (inline, without calling a C function). XS code that
+already the APIs to get values will not be affected by this change. XS code
+that accesses flags directly instead of using API calls to express its intent
+I<might> break, but such code likely is already buggy if passed some other
+values, such as floating point values or objects with string overloading.
+
+This small change permits code (such as JSON serealisers) to reliably determine
```
leonerd: ITYM "serialisers" - you might be thinking of the cutely-misspelled Sereal perl module.

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+    else if (flags & SVf_POK)
+        serealise as string
+    else
+        the existing guesswork ...
+
+Note that this doesn't cover "dualvars" - scalars that report different
+values when asked for their string form or number form (such as C<$!>).
+Most serialisation formats cannot represent such duplicity.
+
+I<the existing guesswork> remains because as well as dualvars, values might
```
leonerd: New sentence should start with a capital: `I<The existing ...`

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+be C<undef>, references, overloaded references, typeglobs and other things that
+Perl itself can represent but do not map one-to-one into external formats, so
+need some amount of approximation or encapsulation.
+
+=item *
```
leonerd, line 544: Is this still related here? This looks to be unrelated change from a different commit elsewhen

sv.c
```diff
@@ -3308,7 +3308,22 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
         Move(ptr, s, len, char);
         s += len;
         *s = '\0';
-        SvPOK_on(sv);
+        /* We used to call SvPOK_on(). Whilst this is fine for (most) Perl code,
+           it means that after this stringification is cached, there is no way
+           to distinguish between values originally assigned as $a = 42; and
+           $a = "42"; (or results of string operators vs numeric operators)
+           where the value has subsequently been used in the other "context"
```
leonerd: I'd try to avoid the word "context" here. Maybe "subsequently been used in the other sense"

sv.c
```diff
@@ -4876,6 +4891,12 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
             SvIV_set(dsv, SvIVX(ssv));
             if (sflags & SVf_IVisUV)
                 SvIsUV_on(dsv);
+            if ((sflags & SVf_IOK) && !(sflags & SVf_POK)) {
+                /* Source was SVf_IOK|SVp_IOK|SVp_POK but not SVf_POK, meaning
+                   an value set as an integer and later stringified. So mark
```
leonerd: "a value"

sv.h
```diff
@@ -1859,19 +1859,30 @@ scalar.
...
+
+#define SvPV_flags(sv, len, flags)              \
```
leonerd: The whitespace change on this line leads to `diff` believing this macro was deleted and added again. I don't think it looks significant or intentional; so if you undo it it'll hopefully clean up the change lines a bit.

pod/perldelta.pod
FGasper: Actually, since [American English is the standard for Perl’s documentation](https://perldoc.perl.org/perldocstyle#Choice-of-language:-American-English), would “serializer” not be apropos?

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
 to perform numerical or string comparison which is aware of operator
 overloading.
 
+=item *
+
+Reading the string form of an integer value no longer sets the flag C<SVf_POK>.
+The string form is still cached internally, and still re-read directly by the
+macros C<SvPV(sv)> I<etc> (inline, without calling a C function). XS code that
+already the APIs to get values will not be affected by this change. XS code
```
FGasper: already _calls_ the APIs?

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+      CUR = 2
+      LEN = 10
+
+(and possibly C<SVf_IsCOW>, but not always)
+
+This now means that if XS code I<really> needs to determine which form a value
+was first written as, it should implement logic roughly
+
+    if (flags & SVf_IOK|SVf_NOK) && !(flags & SVf_POK)
+        serealise as number
```
FGasper: serialize?

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+    else if (flags & SVf_POK)
+        serealise as string
```
FGasper: serialize?

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+    else
+        the existing guesswork ...
```
FGasper, line 533: I know there’s been discussion of this, but FWIW this might be bit nicer if it were wrapped up some new API call … maybe `SvORIGTYPE()`? Then the delta could just mention that API call and _maybe_ some of the flag-manipulation changes, with most of these other details in some sort of .pod document. This discussion may be of especial interest outside the scope of perldelta.

It just seems like a lot of explanation for a changelog.

pod/perldelta.pod
```diff
@@ -383,6 +383,145 @@ with C<..._flags>-suffixed variants.  These expose a simple and consistent API
...
+
+Note that this doesn't cover "dualvars" - scalars that report different
+values when asked for their string form or number form (such as C<$!>).
+Most serialisation formats cannot represent such duplicity.
```
FGasper: serialization?

pod/perldelta.pod
FGasper, line 544: This seems in particular like it ought to be a different section from the much-more-public original-types change.

The audiences for these two things are different: C API callers on the one, Perl-maintainer Jedi wizards on the other.

pod/perlguts.pod
```diff
@@ -307,6 +307,30 @@ in it, you can use the following macros to check the type of SV you have.
...
+        /* originally a string */
+        ...
+    }
+    else if (SvNIOK(sv)) {
+        /* originally numeric */
+        ...
+    }
+    else {
+        /* something special or undef */
+    }
```
FGasper, line 332: Same comment as above: wrapping this up as SvORIGTYPE or some such would be _amazing_ and simplify both the documentation and the lives of folks (like me!) who want to use this new hotness.

FGasper (commented):
Just a few stylistic/docs things, plus a request for a quick-and-easy tool to implement what’s written out here manually.

Thanks!

pod/perldelta.pod
leonerd: I think adding something like that would be useful, but I don't think it needs to hold up this particular PR. We'd verymuch like to get this change in before 5.35.9 (due in 3 days from now). Extending the XS-author API later would be a fine addition which we can do afterwards, perhaps as part of a wider look at "how we do type-like information".

pod/perldelta.pod
nwc10: If you try hard enough you can dereference a` Regexp` as a scalar. (Assignable) LVALUEs are also scalars:

```
$ perl -MDevel::Peek -e '$a = \(substr $^X, 0, 1); Dump $$a'
SV = PVLV(0x20ae4e8) at 0x2085fb0
  REFCNT = 1
  FLAGS = (GMG,SMG)
  IV = 0
  NV = 0
  PV = 0
  MAGIC = 0x209d648
    MG_VIRTUAL = &PL_vtbl_substr
    MG_TYPE = PERL_MAGIC_substr(x)
  TYPE = x
  TARGOFF = 0
  TARGLEN = 1
  TARG = 0x20860e8
  FLAGS = 0
  SV = PV(0x20869c0) at 0x20860e8
    REFCNT = 2
    FLAGS = (POK,pPOK)
    PV = 0x209d6d0 "/usr/bin/perl"\0
    CUR = 13
    LEN = 15
```

I'm not confident that these are the only two cases, so didn't want to claim a definitive list.

demerphq:
On Mon, 5 Jul 2021 at 11:45, Nicholas Clark ***@***.***>
wrote:

> Some external serialised data formats such as JSON can distinguish between
> values that are numbers and values that are strings, and services that
> communicate in JSON often rely on this distinction being possible. Until
> now, Perl has a problem with reliably generating output that meets these
> expectations, because the internal state of $answer after these two code
> paths had been functionally identical:
>
> my $answer = 42;
> print "The answer is $answer\n";
>
> my $answer = "42";
> print "That doesn't look right\n"
>     unless $answer == 6 * 9;
>
> This branch
>
>    - changes Perl_sv_2pv_flags so that it does not set the SVf_POK flag
>    when caching the string representation of an integer value
>    - updates the macros SvPV etc so that they continue to use the cached
>    value directly (and hence do not start calling into Perl_sv_2pv_flags
>    where they had not previously)
>    - updates Perl_sv_setsv_flags to copy the changed flag state accurately
>
> This permits serealisers to reliably distinguish between values first
> written as numbers, and values first written as strings.
>
> These changes exposed a bug in Data::Dumper (now fixed), where it was
> already failing to handle NVs and overloaded references for certain cases.
> All core tests still pass, along with the CPAN modules Sereal, JSON::XS,
> Cpanel::JSON::XS and all their dependencies.
>

Either I have misunderstood the discussion we had on this, and the
explanation above or the patch isn't working as expected:

nicholas/IV-vs-PV:/git_tree/perl$ ./perl -Ilib -MDevel::Peek -e'my $s= " 0
"; my $i=0+$s; Dump($s)'
SV = PVIV(0x55d490dacc08) at 0x55d490daa8e0
  REFCNT = 1
  FLAGS = (IOK,POK,IsCOW,pIOK,pPOK)
  IV = 0
  PV = 0x55d490db64c8 " 0 "\0
  CUR = 3
  LEN = 10
  COW_REFCNT = 1

The result is that $s is IOK and POK, even tho $s started off as a string,
and contains leading and trailing spaces in it.

Is this what you intended? I thought the intent of this patch was to ensure
that code like this would leave $s pIOK but NOT IOK.

Yves

leonerd:
@demerphq There's two sub-problems: Sometimes "stringy" values end up with IOK or NOK accidentally, and sometimes numbers end up with POK accidentally. They are two separate sides of the same overall concern.

This patch only addresses the second kind - it ensures that numbers don't accidentally gain POK as well. So if you see a value with POK you know it was intended to be a string. This patch itself doesn't address the converse problem (of strings gaining IOK|NOK); we'd have to do that separately.

sv.h
nwc10: Well spotted. I try not to let these things sneak out, but this one did.

(I guess my editor decided to be "helpful" and re-indent something)

sv.c
nwc10: Agree both with avoiding *context*, and I can't see a better word than *sense* so went with this.

pod/perldelta.pod
nwc10: Fixed.

sv.c
nwc10: Well spotted. Fixed.

pod/perldelta.pod
nwc10: Whoops yes, good catch. Yes, *calls*

pod/perldelta.pod
nwc10: Fixed.

pod/perldelta.pod
nwc10: Fixed.

pod/perldelta.pod
nwc10: Yes. Fixed.

pod/perldelta.pod
nwc10: Yes, probably we can do better. But for a dev release perldelta, I didn't think we needed to get this perfect.

(As in, I don't think that this should block this PR)

nwc10:
> This patch only addresses the second kind - it ensures that numbers don't accidentally gain POK as well. So if you see a value with POK you know it was intended to be a string. This patch itself doesn't address the converse problem (of strings gaining IOK|NOK); we'd have to do that separately.

I don't think that I can explain it better than that (or add to that usefully, without clarification on what).

pod/perlguts.pod
nwc10: Yes, agree, but I wasn't sure on exactly what, and I didn't see it as a blocker on this PR.

neilb:
Anyone have any reasons why this shouldn't be accepted for inclusion in 5.36?

leonerd (approved):
LGTM

demerphq:
On Fri, 18 Feb 2022 at 15:35, Neil Bowers ***@***.***> wrote:

> Anyone have any reasons why this shouldn't be accepted for inclusion in
> 5.36?
>
It might cause some turbulence, but it will be worth it if it does. I am in
favour of merging it.

cheers,
Yves

ap:
> So if you see a value with POK you know it was intended to be a string. This patch itself doesn't address the converse problem (of strings gaining IOK|NOK); we'd have to do that separately.

Considering that this branch is enough to ensure originally-strings and originally-numbers stay distinguishable as such, is any strong reason left to still do that part?

PR #18959, opened 2021-07-05
av_new_alloc: revise description of size parameter

richardleach:
The parameter is now described more concisely, without resort to other macros or function comparisons.

Perl @iabyn's feedback to #18729


hvds:
LGTM, now merged as 158b05f881.

PR #18962, opened 2021-07-05
Fix syntax error wwhen NAN_COMPARE_BROKEN

khwilliamson:
http://nntp.perl.org/group/perl.perl5.porters/26082


Leont (approved):

nwc10:
I rebased this and pushed it to blead.

PR #18964, opened 2021-07-06
Haiku BFS does not store atime and always return current time

thibaultduponchelle:
There is a limitation in Haiku file system where atime is not stored at all, instead, `stat` always returns the local time.

### DIFF
--- a/lib/File/stat.t
+++ b/lib/File/stat.t
@@ -84,12 +84,20 @@ sub test_X_ops {
             }
             is($@, '', "Overload succeeds $desc");
 
-            if ($^O eq "VMS" && $op =~ /[rwxRWX]/) {
-                is($vwarn, 1, "warning about VMS ACLs $desc");
-            } else {
-                is($rv, eval "-$op \$file", "correct overload $desc")
-                    unless $access;
-                is($vwarn, undef, "no warnings about VMS ACLs $desc");
+            SKIP : {
+                if ($^O eq "haiku" && $op =~ /A/) {
+                    # atime is not stored on Haiku BFS
+                    # and stat always returns local time instead
+                    skip "testing -A $desc_tail on Haiku", 1;
+                }
+
+                if ($^O eq "VMS" && $op =~ /[rwxRWX]/) {
+                    is($vwarn, 1, "warning about VMS ACLs $desc");
+                } else {
+                    is($rv, eval "-$op \$file", "correct overload $desc")
+                        unless $access;
+                    is($vwarn, undef, "no warnings about VMS ACLs $desc");
+                }
             }
 
             # 111640 - File::stat bogus index check in overload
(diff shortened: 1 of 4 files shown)


lib/File/stat.t
```diff
@@ -86,6 +86,10 @@ sub test_X_ops {
 
             if ($^O eq "VMS" && $op =~ /[rwxRWX]/) {
                 is($vwarn, 1, "warning about VMS ACLs $desc");
+            } elsif ($^O eq "haiku" && $op =~ /A/) {
+                # atime is not stored on Haiku BFS
+                # and stat always returns local time instead 
+                note("Not testing -A $desc_tail on Haiku");
```
Leont: That should maybe be a skip instead?

jkeenan:
Are you in a position to submit smoke-test reports for the Haiku OS?  (That might put us in a better position to diagnose/correct bugs.)

Thank you very much.
Jim Keenan

Leont:
> Are you in a position to submit smoke-test reports for the Haiku OS? (That might put us in a better position to diagnose/correct bugs.)

Haiku is not in such a position yet; it fails several tests on blead (I have a fix for another issue laying around).

Leont:
Time::HiRes needs very similar fixups.

lib/File/stat.t
thibaultduponchelle: Fixed

thibaultduponchelle:
> Time::HiRes needs very similar fixups.

I added some skips for Time::HiRes in this PR

thibaultduponchelle:
> Are you in a position to submit smoke-test reports for the Haiku OS? (That might put us in a better position to diagnose/correct bugs.)

@Leont is right about tests failing on haiku.

@jkeenan I can probably run a smoker from time to time, I just sent a first report : https://perl5.test-smoke.org/report/5008053

Until reaching full green test suite (I'm kind of working on it, step by step) I can exclude failing tests to get a green status.

jkeenan:
On 7/8/21 7:52 AM, thibaultduponchelle wrote:
>     Are you in a position to submit smoke-test reports for the Haiku OS?
>     (That might put us in a better position to diagnose/correct bugs.)
> 
> @Leont <https://github.com/Leont> is right about tests failing on haiku.
> 
> @jkeenan <https://github.com/jkeenan> I can probably run a smoker from 
> time to time, I just sent a first report : 
> https://perl5.test-smoke.org/report/5008053 
> <https://perl5.test-smoke.org/report/5008053>
> 

Thank you for that report.

> Until reaching full green test suite (I'm kind of working on it, step by 
> step) I can exclude failing tests to get a green status.
> 

I don't think you need to do this.

The benefit we (mostly non-Haiki users) will derive from your smoke 
tests is:

1. Knowing that we are actively trying to support Perl on this platform.

2. Making sure that changes introduced into Perl 5 blead do not cause 
further test failures -- regressions -- on Haiku.

Hence, in my opinion we should still see what's failing but we don't 
need reports all that frequently.  My suggestion:  Generate a smoke 
report once a month or whenever you manage to reduce the number of failures.

Thank you very much.
Jim Keenan

thibaultduponchelle:
(if you wonder, this PR is ready for review 😆)

PR #18968, opened 2021-07-08
Provide a simple API for testing features enabled

tonycoz:
Inspired by discussion in #p5p.

This calls caller() itself rather than taking hints and hints_hash
parameters so if we end up adding an extra hints word callers won't
need to adjust their code.

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -5,7 +5,7 @@
 
 package feature;
 
-our $VERSION = '1.66';
+our $VERSION = '1.67';
 
 our %feature = (
     fc                   => 'feature_fc',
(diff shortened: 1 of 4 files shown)


lib/feature.pm
```diff
@@ -562,6 +562,57 @@ also does the equivalent of C<use strict>; see L<perlfunc/use> for details.
...
 
+=head1 CHECKING FEATURES
+
+C<feature> provides some simple APIs to check which features are enabled.
+
+=over
+
+=item feature::feature_enabled($feature)
+
+=item feature::feature_enabled($feature, #depth)
```
Grinnz:
```suggestion
=item feature::feature_enabled($feature, $depth)
```

lib/feature.pm
```diff
@@ -562,6 +562,57 @@ also does the equivalent of C<use strict>; see L<perlfunc/use> for details.
 
 =back
 
+=head1 CHECKING FEATURES
+
+C<feature> provides some simple APIs to check which features are enabled.
+
+=over
+
+=item feature::feature_enabled($feature)
```
Grinnz: It produces nicer deeplinks if the =item is just the bare function name, and the way(s) to call it is/are provided in a verbatim block below the heading.

Grinnz:
It's kind of unfortunate that `use feature;` produces an error (though helpful for the usual case, so I don't think we should change it). You would have to `use feature ();` to use these functions, might be worth mentioning.

lib/feature.pm
tonycoz: thanks, fixed.

lib/feature.pm
tonycoz: changed, I've added sample code, and added a note that callers will need to use feature (); if they don't otherwise load it.

thanks for the review.

PR #18971, opened 2021-07-08
Add missing word in 'if's documentation

skirmess:


Grinnz:
"one or more modules" would be a more idiomatic phrase

Grinnz:
Also this section really needs to updated with current recommendations, but that's for another PR.

skirmess:
Thanks, updated. I tried to search if "one or multiple" or "one or more" is better but couldn't find an explanation when to use which.

Grinnz:
"one or more" is a phrase which easily represents a count of items. "one or multiple" is presenting two more separate options: one item, or multiple items.

PR #18976, opened 2021-07-11
Document v5.35 warnings-by-default

cornett:


pod/perl.pod
```diff
@@ -392,9 +392,10 @@ see L<perlvar> for more information.
 Using the C<use strict> pragma ensures that all variables are properly
 declared and prevents other misuses of legacy Perl features.
 
-The C<use warnings> pragma produces some lovely diagnostics. One can
-also use the B<-w> flag, but its use is normally discouraged, because
-it gets applied to all executed Perl code, including that not under
+The C<use warnings> pragma produces some lovely diagnostics.
+These are enabled by default within the scope of C<use v5.35> (or higher).
```
Grinnz: If this is added here, it would also need to mention strict is enabled by `use v5.12` or higher in the paragraph above.

regen/warnings.pl
```diff
@@ -912,7 +915,8 @@ =head1 DESCRIPTION
 The C<warnings> pragma gives control over which warnings are enabled in
 which parts of a Perl program.  It's a more flexible alternative for
 both the command line flag B<-w> and the equivalent Perl variable,
-C<$^W>.
+C<$^W>.  This pragma is enabled automatically by a
```
Grinnz: This should probably be clear that it does the equivalent of `use warnings` or `use warnings 'all'`, since there are a lot of ways to use the pragma.

pod/perlfunc.pod
```diff
@@ -9621,6 +9621,7 @@ C<use VERSION>, possibly removing the C<strict> and C<feature> added by
 C<use VERSION>.  C<use VERSION> does not
 load the F<feature.pm> or F<strict.pm>
 files.
+Similarly, <LC<warnings>> are enabled if C<VERSION> is 5.35.0 or higher.
```
Grinnz: `C<>` here is used for function names or use statements, not needed for a link to the pragma on its own.

```suggestion
Similarly, L<warnings> are enabled if C<VERSION> is 5.35.0 or higher.
```

Grinnz: Also this sentence should be moved before "Any explicit use..." and the following sentences should mention warnings as well.

regen/warnings.pl
```diff
@@ -884,6 +884,9 @@ =head1 SYNOPSIS
     use warnings;
     no warnings;
 
+    # On by default for 5.35 and above
```
Grinnz: Could be a somewhat confusing comment, since there are warnings always on by default already.

Maybe "Standard warnings enabled by use v5.35 or above"?

pod/perl.pod
cornett: Done, thanks.

regen/warnings.pl
cornett: Good idea.

pod/perlfunc.pod
cornett: Fixed.     Note that it now states that use VERSION does not load warnings.pm.  I have no idea if that is true or not.

regen/warnings.pl
cornett: Done

pod/perlfunc.pod
Grinnz: That is accurate.

pod/perl.pod
```diff
@@ -391,10 +391,13 @@ see L<perlvar> for more information.
 
 Using the C<use strict> pragma ensures that all variables are properly
 declared and prevents other misuses of legacy Perl features.
+These are enabled by default within the scope of
+C<L<use v5.12|perlfunc/use VERSION>> (or higher).
 
-The C<use warnings> pragma produces some lovely diagnostics. One can
-also use the B<-w> flag, but its use is normally discouraged, because
-it gets applied to all executed Perl code, including that not under
+The C<use warnings> pragma produces some lovely diagnostics.
```
oodler577, line 399: _lovely_, I do not believe is an appropriate descript since it is meaningless in this context. Can we get a more precise adjective about what is meant? Works like, _actionable_, _informative_, _helpful_, etc. With the audience in mind (a developer), these are the kinds of things I want to be able to know.

pod/perlfunc.pod
```diff
@@ -9608,18 +9608,20 @@ L<C<use>|/use Module VERSION LIST>ing library modules that won't work
...
+with L<C<use strict>|strict>.  
+Similarly, L<warnings> are enabled if C<VERSION> is 5.35.0 or higher.
+Any explicit use of
 C<use strict> or C<no strict> overrides C<use VERSION>, even if it comes
-before it.  Later use of C<use VERSION>
-will override all behavior of a previous
-C<use VERSION>, possibly removing the C<strict> and C<feature> added by
-C<use VERSION>.  C<use VERSION> does not
-load the F<feature.pm> or F<strict.pm>
+before it (likewise for C<use warnings> or C<no warnings>).
```
oodler577: It'd be more clear to not mention `use warnings` here. While maybe technically true, it confuses the point that the default can be changed. So if the net effect is the same, it's probably not necessary information in this section.

regen/warnings.pl
```diff
@@ -884,6 +884,9 @@ =head1 SYNOPSIS
     use warnings;
     no warnings;
 
+    # Standard warnings are enabled by use v5.35 or above
+    use v5.35;
```
oodler577, line 886: Can someone explain to me the purpose of this script and why what seems to be a functional change when the _ask_ here was for documentation?

lib/warnings.pm
```diff
@@ -583,6 +583,9 @@ warnings - Perl pragma to control optional warnings
     use warnings;
     no warnings;
 
+    # Standard warnings are enabled by use v5.35 or above
+    use v5.35;
```
oodler577, line 584: Seems like a functional change to a Perl module? What's the purpose of this kind of change when the _ask_ was for documentation?

oodler577 (changes requested):
Added comments regarding need for clarity. Also questioned functional changes in Perl script and module in a PR that's meant to be for documentation.

regen/warnings.pl
Grinnz: This is the documentation for warnings.pm.

lib/warnings.pm
Grinnz: This is documentation.

pod/perl.pod
Grinnz: Text was already present, so not really in scope for this PR.

lib/warnings.pm
cornett: The ./regen.pl script generates  lib/warnings.pm from regen/warnings.pl

pod/perlfunc.pod
cornett: Sorry, I'm not sure I follow you.  This  is  where `use VERSION` is documented.  The existing documentation mentions that `use v5.12` also enables strictures.  I've just tried to do the same for `use v5.35` and warnings.    I'm happy to take improvements to the wording.

Grinnz:
The test failure is that the warnings.pm (and thus regen/warnings.pl) $VERSION needs to be bumped.

cornett:
Bumped the version, thanks.

lib/warnings.pm
```diff
@@ -5,7 +5,7 @@
 
 package warnings;
 
-our $VERSION = "1.51";
+our $VERSION = "1.53";
```
Leont, line 8: I think only one bump is needed here?

cornett: blead is already at 1.52, so I went to 1.53.   Should it be 1.52?

Grinnz: Well it didn't before but 1.52 is now in the released dev version 5.35.2 so 1.53 is needed. This will also need to be rebased on blead.

Leont: And you should probably also squash the commits when rebasing

pod/perlfunc.pod
Leont: This isn't true, a `no warnings` before a `use 5.36` will enable warnings.

cornett: Thanks.  I removed that statement.

lib/warnings.pm
cornett: I've squashed my commits.   It should ff-merge cleanly onto the warnings-default branch.   I would suggest we do that, close this PR, and add
 any further refinements to https://github.com/Perl/perl5/pull/18666

Leont: I rebased my branch

cornett: Rebased mine as well.

PR #18977, opened 2021-07-12
Pod-Html refactoring 4 of 5

jkeenan:
This is the fourth of five pull requests in support of #18894. These p.r.s are intended to make the internals of Pod-Html less messy. Ultimately Pod::Html will hold a series of methods calls on an object.

This p.r. is concerned solely with the handling of command-line options.

Pod-Html: Simplify the sub which processes options

Pod::Html::Util::parse_command_line() was doing too much.  It should
wrap around GetOptions(), call the usage() routine if needed, then
return a hashref of options to its caller.

Its caller -- Pod::Html::pod2html() -- should do the heavy lifting
involved in processing the options and inserting the data into the
globals hashref.

Hence, we rename Pod::Html::Util::parse_command_line to
Pod::Html::Util::process_command_line() and simplify its functionality.
In lib/Pod/Html.pm we import process_command_line(), call it and pass
its return value onto a new subroutine, process_options(), that
populates $globals.

Document Pod::Html::Util subroutines.

### DIFF
--- a/ext/Pod-Html/lib/Pod/Html/Util.pm
+++ b/ext/Pod-Html/lib/Pod/Html/Util.pm
@@ -47,6 +47,8 @@ Process command-line switches (options).  Returns a reference to a hash.  Will
 provide usage message if C<--help> switch is present or if parameters are
 invalid.
 
+Calling this subroutine may modify C<@ARGV>.
+
 =cut
 
 sub process_command_line {
@@ -86,7 +88,7 @@ sub process_command_line {
 
 =head2 C<usage()>
 
-Display customary Pod::Html usage information.
+Display customary Pod::Html usage information on STDERR.
 
 =cut
 


ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -225,7 +225,9 @@ sub pod2html {
     local $_;
 
     my $globals = init_globals();
-    $globals = parse_command_line($globals);
+    my $opts = process_command_line;
+    $globals = process_options($globals, $opts);
```
oodler577: It's confusing to use the same reference that you're passing in _by reference_ and updating in the sub as the LHS of the assignment. I'd either drop the assignment and LHS `$globals` or do a deep copy and do an actual assignment  - in this case, my preference would be the former (drop the assignment) since it preserves the orignal global semantics the rest of the code assumes.

ext/Pod-Html/lib/Pod/Html/Util.pm
```diff
@@ -30,75 +30,52 @@ Pod::Html::Util - helper functions for Pod-Html
...
-                       'outfile=s'  => \$opt_outfile,
-                       'poderrors!' => \$opt_poderrors,
-                       'podpath=s'  => \$opt_podpath,
-                       'podroot=s'  => \$opt_podroot,
-                       'quiet!'     => \$opt_quiet,
-                       'recurse!'   => \$opt_recurse,
-                       'title=s'    => \$opt_title,
-                       'verbose!'   => \$opt_verbose,
+    my $result = GetOptions(\%opts,
+        'backlink!',
```
oodler577: Minor nit, this can be greatly cleaned up by just throwing the option defs in a `qw//`; e.g.,:

```
my $result = GetOptions(\%opts, qw/backlink! cachedir=s .../);
```

oodler577 (commented):
Overall, much needed change - but I'd consider eliminating the confusion described in my primary comment regarding the LHS assignment of `$globals`.

jkeenan:
On 7/12/21 7:31 PM, oodler577 wrote:
> 
> Overall, much needed change - but I'd consider eliminating the confusion 
> described in my primary comment regarding the LHS assignment of |$globals|.
> 
> 
> In ext/Pod-Html/lib/Pod/Html.pm 
> <https://github.com/Perl/perl5/pull/18977#discussion_r668321995>:
> 
>> @@ -225,7 +225,9 @@ sub pod2html {
>       local $_;
>   
>       my $globals = init_globals();
> -    $globals = parse_command_line($globals);
> +    my $opts = process_command_line;
> +    $globals = process_options($globals, $opts);
> 
> It's confusing to use the same reference that you're passing in /by 
> reference/ and updating in the sub as the LHS of the assignment. I'd 
> either drop the assignment and LHS |$globals| or do a deep copy and do 
> an actual assignment - in this case, my preference would be the former 
> (drop the assignment) since it preserves the orignal global semantics 
> the rest of the code assumes.
> 

This will be cleared up in the 5th of 5 pull requests, when the data 
currently being stored inside $globals will be stored in a proper 
Pod::Html object.

Thank you very much.
Jim Keenan

jkeenan:
On 7/12/21 7:31 PM, oodler577 wrote:
> [snip]
> Overall, much needed change - but I'd consider eliminating the confusion 
> described in my primary comment regarding the LHS assignment of |$globals|.
> 
> 
> Minor nit, this can be greatly cleaned up by just throwing the option 
> defs in a |qw//|; e.g.,:
> 
> |$result = GetOptions(\%opts, qw/backlink! cachedir=s .../); |
> 

This is the sort of change that I think would be better addressed by a 
patch or pull request after we've completed the transformation of the 
main module from one, monolithic subroutine into method calls on an object.

I've added it to the list of concerns at:
http://thenceforward.net/perl/misc/pod-html-remaining-issues.txt

Thank you very much.
Jim Keenan

oodler577:
Very good, thank you!

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -294,6 +296,36 @@ sub init_globals {
     return \%globals;
 }
 
+sub process_options {
+    my ($globals, $opts) = @_;
+
+    @{$globals->{Podpath}}  = split(":", $opts->{podpath}) if defined $opts->{podpath};
```
rjbs:
```suggestion
    $globals->{Podpath}     = [ split(":", $opts->{podpath}) if defined $opts->{podpath} ];
```

I don't _think_ it's intentional that right now you're keeping the same underlying array if this is ever called twice, which is just a tiny bit closer to allowing action at a distance.  Also, un-special-casing this would allow you to slot it down in alpha order.  This is a minor thing, though.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -294,6 +296,36 @@ sub init_globals {
...
+    $globals->{Htmlfile}  =  unixify($opts->{outfile})   if defined $opts->{outfile};
+    $globals->{Poderrors} =          $opts->{poderrors}  if defined $opts->{poderrors};
+    $globals->{Podroot}   =  unixify($opts->{podroot})   if defined $opts->{podroot};
+    $globals->{Quiet}     =          $opts->{quiet}      if defined $opts->{quiet};
+    $globals->{Recurse}   =          $opts->{recurse}    if defined $opts->{recurse};
+    $globals->{Title}     =          $opts->{title}      if defined $opts->{title};
+    $globals->{Verbose}   =          $opts->{verbose}    if defined $opts->{verbose};
+
+    warn "Flushing directory caches\n"
+        if $opts->{verbose} && defined $opts->{flush};
```
rjbs: If you move this `warn` in side the conditional two lines down, you can simplify its trailing `if...`

ext/Pod-Html/lib/Pod/Html/Util.pm
```diff
@@ -30,75 +30,52 @@ Pod::Html::Util - helper functions for Pod-Html
 
 =head1 SUBROUTINES
 
-=head2 C<parse_command_line()>
+=head2 C<process_command_line()>
 
-TK
+Process command-line switches (options).  Returns a reference to a hash.  Will
+provide usage message if C<--help> switch is present or if parameters are
+invalid.
```
rjbs:
```suggestion

Calling this subroutine may modify C<@ARGV>.

```

ext/Pod-Html/lib/Pod/Html/Util.pm
```diff
@@ -30,75 +30,52 @@ Pod::Html::Util - helper functions for Pod-Html
...
-    return $globals;
+    usage("-") if defined $opts{help};  # see if the user asked for help
+    $opts{help} = "";                   # just to make -w shut-up.
+    return \%opts;
 }
 
 =head2 C<usage()>
 
-TK
+Provide usage message on STDERR.
```
rjbs:
```suggestion
Print usage message to STDERR.
```

ext/Pod-Html/lib/Pod/Html/Util.pm
```diff
@@ -149,7 +126,8 @@ END_OF_USAGE
 
 =head2 C<unixify()>
 
-TK
+Provide the full path to a file, taking into account OS-specific
+peculiarities.
```
rjbs:
```suggestion
When passed the full native path to a file, this routine returns it in a unix-like format.
```

rjbs (approved):
I think this looks good.  I made some suggestions that are worth applying, but nothing critical.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > I don't _think_ it's intentional that right now you're keeping the same underlying array if this is ever called twice, which is just a tiny bit closer to allowing action at a distance. Also, un-special-casing this would allow you to slot it down in alpha order. This is a minor thing, though.

I took this approach because in the current production (v5.34.0) release, `ext/Pod-Html/lib/Pod/Html.pm` has this:
```
553 
554     @Podpath  = split(":", $opt_podpath) if defined $opt_podpath;
555 
```
When refactoring a file to replace 20+ lexical variables defined outside of the principal subroutine with `%globals` (which in the next p.r. will become a Pod::Html object), I was taking care to just change one thing at a time.  Hence, the syntax above rather than your suggestion.  By the same rationale, I have not yet changed any of the `Initialcap` spelling of those lexicals (which are now keys within `%globals`) -- even though if I were starting from scratch they would be all `lowercase`.  Since changing those keys from `Initialcap` to `lowercase` would expand the size of the diff without improving functionality, I'd prefer to do that after we've got our Pod::Html object in place.

In any case, when I tried out your suggested replacement code, I got:
```
!perl -Ilib -c lib/Pod/Html.pm
syntax error at lib/Pod/Html.pm line 328, near ") if"
Global symbol "$globals" requires explicit package name (did you forget to declare "my $globals"?) at lib/Pod/Html.pm line 330.
Global symbol "$opts" requires explicit package name (did you forget to declare "my $opts"?) at lib/Pod/Html.pm line 330.
...
```
Perhaps this is what you were aiming for?
```
-    @{$globals->{Podpath}}  = split(":", $opts->{podpath}) if defined $opts->{podpath};
+    $globals->{Podpath}   = (defined $opts->{podpath})
+                            ? [ split(":", $opts->{podpath}) ]
+                            : [];
```

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > If you move this `warn` in side the conditional two lines down, you can simplify its trailing `if...`

Again, this is a place where I wanted to avoid doing too much refactoring too soon.  So I retained the flow of the code in the version which shipped with perl-5.34:
```
572     warn "Flushing directory caches\n"
573         if $opt_verbose && defined $opt_flush;
574     $Dircache = "$Cachedir/pod2htmd.tmp";
575     if (defined $opt_flush) {
576         1 while unlink($Dircache);
577     }
```
So, I'd like to defer this until after we've got our Pod::Html object.  I'll add it to the [pod-html TODO list](http://thenceforward.net/perl/misc/pod-html-remaining-issues.txt).

ext/Pod-Html/lib/Pod/Html/Util.pm
jkeenan: Accepted.  We'll review `@ARGV` when we're done to evaluate this and other comments about that array.

ext/Pod-Html/lib/Pod/Html/Util.pm
jkeenan: Between the time of your initial review and now I committed other changes to the documentation of the helper subroutines in `lib/Pod/Html.pm`.  So you should probably work from those.

ext/Pod-Html/lib/Pod/Html/Util.pm
jkeenan: > 

Between the time of your initial review and now I committed other changes to the documentation of the helper subroutines in `lib/Pod/Html.pm`.  So you should probably work from those.

jkeenan:
@rjbs, you previously approved this p.r., but I'll give you a couple of days to respond to my comments before merging to blead.

Thank you very much.
Jim Keenan

ext/Pod-Html/lib/Pod/Html.pm
rjbs: I just meant to put the `]` before the `if` clause, and not after.  I think your version, or the code as it is, are fine.

rjbs:
@jkeenan Thanks for the ping.  I believe this is all okay.

PR #18982, opened 2021-07-14
Grammar upgrade?

jidanni:
I think I corrected this right...?


Grinnz:
This is not correct.

jidanni:
I get it.
It's like `This does make even top players "slam" baseballs harder.`

Grinnz:
Yes, the sentence is the same by removing "even", and "binmode" is a verb.

PR #18986, opened 2021-07-16
win32sck.c: set WSA_FLAG_OVERLAPPED on new sockets

xenu:
This makes it possible to perform overlapped operations on sockets
created by Perl.

I'm not aware of any downsides to enabling that flag, winsock
documentation says that "most sockets should be created with this flag
set".

Perl libuv bindings require this change to work on Windows:

https://github.com/p5-UV/p5-UV/issues/38
https://github.com/p5-UV/p5-UV/issues/35

### DIFF
--- a/win32/win32sck.c
+++ b/win32/win32sck.c
@@ -618,7 +618,7 @@ open_ifs_socket(int af, int type, int protocol)
     SOCKET out = INVALID_SOCKET;
 
     if ((s = PerlEnv_getenv("PERL_ALLOW_NON_IFS_LSP")) && atoi(s))
-        return WSASocket(af, type, protocol, NULL, 0, 0);
+        return WSASocket(af, type, protocol, NULL, 0, WSA_FLAG_OVERLAPPED);
 
     if (WSCEnumProtocols(NULL, NULL, &proto_buffers_len, &error_code) == SOCKET_ERROR
         && error_code == WSAENOBUFS)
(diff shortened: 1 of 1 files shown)


xenu:
Apparently this change caused some tests to fail, marking as draft for now.

xenu:
Argh, apparently while WSA_FLAG_OVERLAPPED doesn't change the behavior of the socket functions, it does force ReadFile() and WriteFile() into async mode and they need to be called with OVERLAPPED structure. That's unfortunate.

leonerd:
For now it doesn't look like these will be likely in Perl, so I'll just skip the relevant bits on UV on Windows. In any case it only affects functions that try to libuv-wrap sockets created elsewhere. All the stuff in UV that creates its own sockets works fine, so this is relatively cornercase.

jkeenan:
> This makes it possible to perform overlapped operations on sockets created by Perl.
> 
> I'm not aware of any downsides to enabling that flag, winsock documentation says that "most sockets should be created with this flag set".
> 
> Perl libuv bindings require this change to work on Windows:
> 
> [p5-UV/p5-UV#38](https://github.com/p5-UV/p5-UV/issues/38) [p5-UV/p5-UV#35](https://github.com/p5-UV/p5-UV/issues/35)

@xenu, do you still wish to pursue this pull request?

Yaribz:
It would be interesting to have a way to create overlapped Windows sockets from Perl, either with an optional `CORE::socket()` extra parameter for platform-specific options, or a dedicated `Win32::socket()` XS function.

xenu:
I still think that overlapped sockets should be the default, because they're more flexible and they are what third party code usually expects (mostly because `socket()` in winapi creates overlapped sockets). Unfortunately, they can't be (easily) read synchronously with ReadFile(), which in turn means they don't work with read(). 

I'm not sure whether we should rewrite read() to transparently handle overlapped sockets or just give up on this idea. 

>It would be interesting to have a way to create overlapped Windows sockets from Perl, either with an optional CORE::socket() extra parameter for platform-specific options, or a dedicated Win32::socket() XS function.

If we were to go this way, I think it should be a separate function. Adding new parameters to core functions is problematic because of prototypes.

Anyway, this PR is abandoned.

PR #18997, opened 2021-07-23
Alternative fast determination of most and least set bits in a machine word

khwilliamson:
This is an alternative to #18944.  I realized that there are places in core that want a 32 bit version, and places that want a 64 bit one if available, regardless of UV size.  This PR, informed by @xenu 's comments creates a pair of functions for each operation, one for each size.  The next step would be to #define a macro that uses the appropriate one for the UV size on the current platform

Modern hardware tends to have machine instructions to find this information.

This series of commits creates functions that return the least and most set bits in the largest word perl thinks the hardware is capable of, and uses them in all but one of the places in core that I am aware of could benefit. (Perhaps you know of other places.) The final place has to do with Unicode handling, is more complex, and I'm deferring that until this P.R. is merged.

There are builtins for clang, gcc, and Windows to find the msb; a builtin for Windows for the lsb, and a POSIX function, ffs(), for the latter, but defined only for ints. Many platforms have a ffsl() additionally, and blead already has probes for that and the other possibilities. If a platform is lacking, fast bit twiddling, and deBruijn sequences are used instead, so the functions are completely general.

These functions are currently considered core-only, but if there is a benefit for modules to be able to use them, I could easily create public documentation for them.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -664,7 +664,7 @@ Perl_is_utf8_invariant_string_loc(const U8* const s, STRLEN len, const U8 ** ep)
     return TRUE;
 }
 
-/* Below are functions to find the final or only set bit in a word.  On
+/* Below are functions to find the first, last, or only set bit in a word.  On
  * platforms with 64-bit capability, there is a pair for each operation; the
  * first taking a 64 bit operand, and the second a 32 bit one.  The logic is
  * the same in each pair, so the second is stripped of most comments. */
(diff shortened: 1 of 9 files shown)


inline.h
```diff
@@ -664,97 +664,412 @@ Perl_is_utf8_invariant_string_loc(const U8* const s, STRLEN len, const U8 ** ep)
...
+ 
+#endif
+
+}
+
+
+/* Convert the leading zeros count to the bit position of the first set bit.
+ * This just subtracts from the highest position, 31 or 63.  But some compilers
+ * don't optimize this optimally, and so a bit of bit twiddling encourages them
+ * to do the right thing.  It turns out that subracting a smaller non-negative
```
richardleach, line 837: s/subracting/subtracting

khwilliamson: Thanks

khwilliamson:
Applied as bf874180e9a7a8fbb74ab7eaa5c683ecc5a6ffc7..4a1b7bb2bdf3d5a25e306752e4471e21093db314

druud:
Next to finding a set bit position (p), also consider functions that return 2^p directly, like:
```
$n & -$n
```
which becomes a mask when subtracting 1, etc.

See also (the mnemonic column in)
https://en.wikipedia.org/wiki/Find_first_set

PR #19000, opened 2021-07-24
installhtml: new location for imports of helper functions

jkeenan:
As reported by sisyphus in
https://www.nntp.perl.org/group/perl.perl5.porters/2021/07/msg260930.html

### DIFF
--- a/installhtml
+++ b/installhtml
@@ -8,7 +8,8 @@ use File::Path qw(remove_tree);
 use File::Spec::Functions qw(rel2abs no_upwards);
 use Getopt::Long;	# for command-line parsing
 use Cwd;
-use Pod::Html 1.23 qw(anchorify relativize_url);
+use Pod::Html 1.32;
+use Pod::Html::Util 1.32 qw(anchorify relativize_url);
 
 =head1 NAME
 


Grinnz:
Perhaps it would be better to have Pod::Html import these functions and continue to export what it previously had available for export?

jkeenan:
On 7/24/21 12:57 PM, Dan Book wrote:
> Perhaps it would be better to have Pod::Html import these functions and 
> continue to export them?
> 

No, it would not.  The whole point of the series of pull requests I've 
been making is to convert ext/Pod-Html/lib/Pod/Html.pm internally to an 
object-oriented construction, subject to the constraint that it exports 
only one function, pod2html(), which is needed by both the pod2html and 
installhtml utilities.

This will become more evident once all 5 of the p.r.s have been applied; 
we're at 3 of 5.  In the next few days I hope to respond to @rjbs's 
review of p.r. 4 of 5, then create 5 of 5 wherein the object-oriented 
construction will be introduced.

Thank you very much.
Jim Keenan

Grinnz:
Re-exporting functions to preserve the existing documented interface does not conflict with an object-oriented construction.

haarg:
htmlify and anchorify were previously documented functions, so removing them is a compatibility break.

sisyphus:
With that patch applied, the html files seem to be installed ok.
At least, the same html files as were installed with earlier versions of Pod::Html are being installed.

However, with Pod-Html-1.31, I see the following warning repeated many times:
```
Use of uninitialized value $Podroot in string eq at C:/perl-5.35.2-1110/lib/Pod/Html.pm line 372.
```
This warning, which is absent with Pod-Html-1.28, occurs both when installing perl, and when subsequently installing some  modules from the CPAN (eg Win32::IPC) into perl.

I never reference these html files, and I don't know how many there are meant to be.  AFAICT, Strawberry Perl avoids creating and installing them.
Does anyone know how they achieve that ?

jkeenan:
> Perhaps it would be better to have Pod::Html import these functions and continue to export what it previously had available for export?

Point now taken.  Implemented as part of 68327975be.  Thanks.

jkeenan:
I have merged this branch (`installhtml-20210724`) to blead in commit 7e12ee61fa.  This should address the original problem reported by @sisyphus in the p5p post cited above.

This commit does *not* correct the uninitialized value warnings, but I have identified the source of those warnings and they are already fixed in an upcoming pull request.  I have added them to my [pod-html TODO list](http://thenceforward.net/perl/misc/pod-html-remaining-issues.txt).  Concerns raised by @Grinnz and @haarg have been addressed in commit 68327975be0888d6cf8425ec1fb62510c1321310.

Thank you very much.
Jim Keenan

PR #19001, opened 2021-07-24
Save some bitwise & instructions by casting to U8

khwilliamson:
I believe that on any platform that Perl runs on, casting to U8 is the same as ANDing with 0xFF.  This wouldn't be the case if a byte were anything different than 8 bits, but Perl doesn't run on such a platform

This is a series of commits that remove the unnecessary '&'.  In many cases the result is more readable, and serves as an example for future maintainers that '& 0xFF' is unnecessary.

In some cases the code did both a cast and an AND.  This is redundant.

I did not check to see if typical compilers optimized away the instructions, but even if many do; probably not all do, and this isn't less readable

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -1236,18 +1236,18 @@ Use L</UV> to declare variables of the maximum usable size on this platform.
 
 /* byte-swapping functions for big-/little-endian conversion */
 # define _swab_16_(x) ((U16)( \
-         (((U16)(x) & UINT16_C(0x00ff)) << 8) | \
-         (((U16)(x) & UINT16_C(0xff00)) >> 8) ))
+         (((U16)(U8) (x)                   ) << 8) |  \
+         (((U16)     (x) & UINT16_C(0xff00)) >> 8) ))
 
 # define _swab_32_(x) ((U32)( \
-         (((U32)(x) & UINT32_C(0x000000ff)) << 24) | \
+         (((U32)(U8)(x)                   ) << 24) | \
          (((U32)(x) & UINT32_C(0x0000ff00)) <<  8) | \
          (((U32)(x) & UINT32_C(0x00ff0000)) >>  8) | \
          (((U32)(x) & UINT32_C(0xff000000)) >> 24) ))
 
 # ifdef HAS_QUAD
 #  define _swab_64_(x) ((U64)( \
-          (((U64)(x) & UINT64_C(0x00000000000000ff)) << 56) | \
+          (((U64)(U8)(x)                           ) << 56) | \
           (((U64)(x) & UINT64_C(0x000000000000ff00)) << 40) | \
           (((U64)(x) & UINT64_C(0x0000000000ff0000)) << 24) | \
           (((U64)(x) & UINT64_C(0x00000000ff000000)) <<  8) | \
--- a/perl.h
+++ b/perl.h
@@ -4231,11 +4231,11 @@ my_swap16(const U16 x) {
 #  define htovl(x)      vtohl(x)
 #  define htovs(x)      vtohs(x)
 #elif BYTEORDER == 0x4321 || BYTEORDER == 0x87654321
-#  define vtohl(x)	((((x)&0xFF)<<24)	\
-                        +(((x)>>24)&0xFF)	\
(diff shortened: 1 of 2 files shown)


perl.h
```diff
@@ -1224,18 +1224,18 @@ Use L</UV> to declare variables of the maximum usable size on this platform.
 
 /* byte-swapping functions for big-/little-endian conversion */
 # define _swab_16_(x) ((U16)( \
-         (((U16)(x) & UINT16_C(0x00ff)) << 8) | \
-         (((U16)(x) & UINT16_C(0xff00)) >> 8) ))
+         (((U16)(U8) (x)                   ) << 8) |  \
+         (((U16)     (x) & UINT16_C(0xff00)) >> 8) ))
```
xenu, line 1240: IMO it would be more consistent to keep the bitwise AND here, because the next line does a similar operation on different bits.

perl.h
```diff
@@ -1224,18 +1224,18 @@ Use L</UV> to declare variables of the maximum usable size on this platform.
...
 
 # define _swab_32_(x) ((U32)( \
-         (((U32)(x) & UINT32_C(0x000000ff)) << 24) | \
+         (((U32)(U8)(x)                   ) << 24) | \
```
xenu, line 1243: IMO it would be more consistent to keep the bitwise AND here, because the next line does a similar operation on different bits.

perl.h
```diff
@@ -1224,18 +1224,18 @@ Use L</UV> to declare variables of the maximum usable size on this platform.
...
          (((U32)(x) & UINT32_C(0x0000ff00)) <<  8) | \
          (((U32)(x) & UINT32_C(0x00ff0000)) >>  8) | \
          (((U32)(x) & UINT32_C(0xff000000)) >> 24) ))
 
 # ifdef HAS_QUAD
 #  define _swab_64_(x) ((U64)( \
-          (((U64)(x) & UINT64_C(0x00000000000000ff)) << 56) | \
+          (((U64)(U8)(x)                           ) << 56) | \
```
xenu, line 1250: IMO it would be more consistent to keep the bitwise AND here, because the next line does a similar operation on different bits.

perl.h
```diff
@@ -4220,11 +4220,11 @@ my_swap16(const U16 x) {
 #  define htovl(x)      vtohl(x)
 #  define htovs(x)      vtohs(x)
 #elif BYTEORDER == 0x4321 || BYTEORDER == 0x87654321
-#  define vtohl(x)	((((x)&0xFF)<<24)	\
-                        +(((x)>>24)&0xFF)	\
+#  define vtohl(x)	((((U8) (x)) << 24)     \
+                        +((U8) ((x) >> 24))     \
```
xenu, line 4235: IMO it would be more consistent to keep the bitwise AND here, because the next line does a similar operation on different bits.

win32/fcrypt.c
```diff
@@ -357,8 +357,8 @@ des_set_key(des_cblock *key, des_key_schedule schedule)
         PERM_OP (d,c,t,1,0x55555555);
         PERM_OP (c,d,t,8,0x00ff00ff);
         PERM_OP (d,c,t,1,0x55555555);
-        d=	(((d&0x000000ff)<<16)| (d&0x0000ff00)     |
-                 ((d&0x00ff0000)>>16)|((c&0xf0000000)>>4));
+        d= ((((unsigned char) d) <<16)| (d&0x0000ff00)     |
```
xenu, line 360: IMO it would be more consistent to keep the bitwise AND here, because the next line does a similar operation on different bits.

xenu (commented):
Apart from the few places I commented on, LGTM.

perl.h
khwilliamson: Actually it is a fact that it would be more consistent; it isn't just your opinion.

The real question is does the gain in performance outweigh the inconsistency.  You're saying No, and I'm saying I am willing to take your word on that.  I have pushed the rest of the commits to blead, leaving these open for a while longer to see if others wish to comment.  If none do, I'll just abandon these

khwilliamson: no responses, so closing this ticket without further changes

PR #19002, opened 2021-07-25
Document a way to perform `use Module VERSION LIST` on the command-line

book:
The `split(/,/,q{foo,bar})` seems to still happen.

This is the currently documented case:

    $ perl -MO=Deparse '-MList::Util=any,all' -e1
    use List::Util (split(/,/, 'any,all', 0));
    '???';
    -e syntax OK

And these work too (and the split on comma also happens somehow):

    $ perl -MO=Deparse '-MList::Util any,all' -e1
    use List::Util ('any', 'all');
    '???';
    -e syntax OK

    $ perl -MO=Deparse '-MList::Util 1.5' -e1
    use List::Util 1.5;
    '???';
    -e syntax OK

    $ perl -MO=Deparse '-MList::Util 1.5 any,all' -e1
    use List::Util 1.5 ('any', 'all');
    '???';
    -e syntax OK

### DIFF
--- a/pod/perlrun.pod
+++ b/pod/perlrun.pod
@@ -698,6 +698,11 @@ is that B<-MI<MODULE>=number> never does a version check,
 unless C<I<MODULE>::import()> itself is set up to do a version check, which
 could happen for example if I<MODULE> inherits from L<Exporter>.
 
+However, it should be noted that since the syntax C<'-MI<MODULE foo,bar>'>
+shown above will generate C<use MODULE ('foo', 'bar')>, it's possible
+to do a version check using C<'-MI<MODULE VERSION foo,bar>'>, which will
+generate C<use MODULE VERSION ('foo', 'bar')>.
+
 =item B<-n>
 X<-n>
 


Grinnz:
The split actually doesn't happen, this only works because strict is not active.

Grinnz:
-M is a very naive filter, it literally sticks the argument verbatim into a `use` statement. Only when the = is provided are the arguments split into a list of strings.

Grinnz:
I would not approve of the current wording of this because it is B::Deparse that generates that syntax, not -M.

book:
> I would not approve of the current wording of this because it is B::Deparse that generates that syntax, not -M.

Agreed. I was confused by the difference in the `B::Deparse` output. Thanks for the explanation.

If I understand your comment correctly, it's the log message I should rework, not the actual documentation patch, right?

Grinnz:
I think mentioning that the quoted form without = can be used to insert a version check and *quoted* argument list would be useful. But it should be quoted in that case because the = argument splitting doesn't occur.

book:
Re-reading your comments after a day, I now fully understand all of your comments:
* " this only works because strict is not active": no strict 'subs' enables poetry mode, so barewords are treated as strings
* " it is B::Deparse that generates that syntax, not -M": B::Deparse produces the string syntax from the barewords-interpreted-as-strings above
* "it should be quoted in that case because the = argument splitting doesn't occur": with strict, the quotes are indeed required; and without =, no splitting is done

I'll rewrite the patch and the message to be more accurate, and force push on my branch again.

khwilliamson:
@book pinging you about this

book:
@khwilliamson sorry, didn't find time to rewrite this in a satisfactory way yet.

jkeenan:
> @khwilliamson sorry, didn't find time to rewrite this in a satisfactory way yet.

@book, can you provide us with an update on the status of this p.r.?

Thank you very much.
Jim Keenan

jkeenan:
> > @khwilliamson sorry, didn't find time to rewrite this in a satisfactory way yet.
> 
> @book, can you provide us with an update on the status of this p.r.?
> 
> Thank you very much. Jim Keenan

@book, any progress on this?  Or should we close the ticket?

book:
I'll get back to it, after I'm done with #20337.

demerphq:
@book maybe do this first? Its just a doc patch, so it should be easy to get it right and merged.

jkeenan:
> I'll get back to it, after I'm done with #20337.

@book, should we continue to keep this pull request open?

PR #19005, opened 2021-07-27
detect struct stat.st_dev's size and signedness, and return it safely

### LINKED ISSUE #18837: stat() returns wrong dev and rdev values on newer FreeBSD systems
On newer FreeBSD systems (probably starting with version 12) the `dev` and `rdev` values returned by `stat()` are wrong. Here's a quick check comparing with the output of the `stat(1)` utility:

```
$ perl -e '$file = shift; my @p = stat($file); print join(" ", @p[0,6]), "\n"' /tmp; stat -f "%d %r" /tmp
-2029977855 -1
18446744071679573761 18446744073709551615
```

On older FreeBSD systems (I have versions 9 and 10 available, but possibly this works also with version 11) the two printed lines are equal:
```
79 165976
79 165976
```

I guess that the problem is caused by the usage of `mPUSHi`, but newer FreeBSD's `dev_t` is now a `uint64` (see this code line: https://github.com/Perl/perl5/blob/ea852e0ec11bf69a13d2ea959b112905b2895ab7/pp_sys.c#L2917 ). Using `mPUSHu` instead fixes the problem (I tried this with https://github.com/eserte/perl5/commit/3fee3134f3b0849152c990451eb40b5941b5f09f ; pp_sys.c has to be compiled with `-DDEV_T_IS_UNSIGNED`).

A further note: do we have a mPUSH macro which does the correct thing dependind on the type of its argument? Something using `typeof()` or `_Generic` or so?

The problem was noticed in https://github.com/uperl/FFI-C-Stat/issues/5 . FFI::C::Stat does it right, but the test suite compares its results with perl's `stat()`, and fails on newer FreeBSD (@plicease FYI)..

tonycoz:
On FreeBSD dev_t (and hence the st_dev member of struct stat) is an
unsigned 64-bit integer, and the previous simple PUSHi() corrupted
that.

Fixes #18837


NetWare/config.wc
```diff
@@ -1042,6 +1042,8 @@ srand48_r_proto='0'
 srandom_r_proto='0'
 src=''
 ssizetype='int'
+st_dev_sign='4'
```
Leont: I think that should be a 1?

uconfig64.sh
```diff
@@ -850,6 +850,8 @@ srand48_r_proto='0'
 srandom_r_proto='0'
 src='.'
 ssizetype=long
+st_dev_sign='4'
```
Leont: This should be a 1 too?

NetWare/config.wc
tonycoz: Thanks, both fixed

pp_sys.c
```diff
@@ -2915,7 +2915,56 @@ PP(pp_stat)
...
+#  endif
+        else {
+            bool neg;
+            Stat_t s;
+#  if ST_DEV_SIGN < 0
+            neg = PL_statcache.st_dev < 0;
+#  else
+            neg = FALSE;
+#  endif
+            if (neg) {
```
Leont: I don't understand why this numification is hand-rolled here

pp_sys.c
```diff
@@ -2915,7 +2915,56 @@ PP(pp_stat)
     if (max) {
         EXTEND(SP, max);
         EXTEND_MORTAL(max);
+#if ST_DEV_SIZE < IVSIZE || (ST_DEV_SIZE == IVSIZE && ST_DEV_SIGN < 0)
         mPUSHi(PL_statcache.st_dev);
+#elif ST_DEV_SIZE == IVSIZE
+        mPUSHu(PL_statcache.st_dev);
+#else
+#  if ST_DEV_SIGN < 0
```
xenu, line 2923: 1. Why isn't ST_DEV_SIGN a boolean?
2. Is a configure probe for st_dev's signedness really needed? The code for `st_ino` manages to do exactly the same thing without it. Come to think of it, I wonder if it would be possible to merge the handling of them somehow...

tonycoz:
I've re-worked this to not use our own stringification.

It does mean that this change depends on C99 when IVSIZE < sizeof(dev_t), but I think that's reasonable.

jkeenan:
To move discussion of this issue/p.r. forward, I have created this smoke-me branch:
smoke-me/jkeenan/tonycoz/18837-large-dev

@eserte , can you see whether this addresses your original concerns in https://github.com/Perl/perl5/issues/18837?

PR #19006, opened 2021-07-28
av.h - document newAV_alloc_xz

richardleach:


richardleach:
@khwilliamson is this commit a sufficient way of resolving your comment on #p5p that:
```
<khw> 
hydahy, the entry for newAV_alloc_xz in embed.fnc is wrong
The d flag indicates it is documented, but no pod was found for it
```
Or does each `=for apidoc` really need its own text?

av.h
```diff
@@ -110,6 +110,7 @@ Perl equivalent: C<my @array;>.
 
 /*
 =for apidoc newAV_alloc_x
+=for apidoc newAV_alloc_xz
```
khwilliamson, line 113: Use =for apidoc_item newAV_alloc_xz
to get the same write-up to apply to both of them

You might also consider rolling these two into plain newAV, and change that to begin with something like

> These all create a new AV, setting the reference count to 1.

Then compare and contrast the three forms.  This has the advantage of having one entry for all three forms, so there is less writing, and the advantages of the newer ones can be stressed, and it will be less likely for the reader scanning the document to see the first one and stopping there without noticing that these other forms exist which might better suit her/his needs

khwilliamson: SvREFCNT_inc and _dec are examples of this in perlapi.pod

richardleach: Thanks, rolling them together is a great idea. How does the following look? (I'll add some POD markup later.)
```
/*
=for apidoc newAV
=for apidoc_item newAV_alloc_x
=for apidoc_item newAV_alloc_xz

These all create a new AV, setting the reference count to 1. They differ
in when and how an accompanying SV* array is allocated and initialized.

The Perl equivalent is approximately C<my @array;>.

newAV does not allocate a SV* array. 
    AV *av = newAV();

This is very useful when an AV is required, but population of it may be
deferred, or even never actually take place. (Memory is not allocated
unnecessarily.)

Subsequent SV* array allocation would be performed via av_extend. This 
might be called directly:
    av_extend(av, key);

Or it might be called implicitly when the first element is stored:
    (void)av_store(av, 0, sv);

Whether or not any as-yet-untouched array elements are initialized with
NULL pointers by av_extend depends upon whether the array is "real" at 
the time.

In contrast, when an AV is created for immediate population with a known
(or likely) number of elements, it is more efficient to immediately
allocate a SV* array of the necessary size. (This avoids inefficient use
of av_extend and the potential for the first allocation being too small
and then having to resize it.)

For that scenario, newAV_alloc_x and newAV_alloc_xz can be used to create
an AV and allocate a SV* array to fit the specified number of elements.
(As a result, these macros MUST NOT be called with a size less than 1.)

newAV_alloc_x does not initialize the array with NULL pointers - and so 
the expectation is that all should be initialized elsewhere prior to any
potentials reads. newAV_alloc_xz does initialize the array with NULLs.

As an example, the following examples all result in an array that can
fit four elements (indexes 0 .. 3):

    AV *av = newAV();
    av_extend(av, 1);

    AV *av = newAV();
    av_extend(av, 3);

    AV *av = newAV_alloc_xz(4);

    AV *av = newAV_alloc_x(4);

In all but the newAV_alloc_x case, the array elements will contain 
NULL pointers. In that last case, the element contents are undefined.

In contrast, the following examples allocate an SV* array that is only
guaranteed to fit one element:

    AV *av = newAV_alloc_x(1);
    AV *av = newAV_alloc_xz(1);

=cut
*/
```

khwilliamson: lgtm @Grinnz how does the word smithing sound to you?

Grinnz: Each =for needs to be preceded and followed by a newline. Don't know anything about the implementation but grammar looks fine.

khwilliamson:
To be combined, the first line is =for apidoc
The subsequent lines  need to be =for apidoc_item
and no extra line need be used between them.

embed.fnc documents the syntax, or use sv.h as a template.

khwilliamson:
And that is the way the commit already had it

khwilliamson:
Oops, I thought that the commit had been changed to the text, so merged it.  But it was just text in this conversation @richardleach could you please turn that into a commit rebased from blead?

Sorry

iabyn:
I'm a bit concerned that newAV_alloc_x/newAV_alloc_xz are exposing the
internal implementation details of how uninitialised array elements
are indicated.

To put this in context, some years back sprout changed the implementation
from using &PL_sv_undef to using NULL to mark unused AvARRAY() slots. We
got it in the neck then from certain elements of the "you incompetent
fools are breaking everything" brigade.

Officially XS code is not supposed to access the elements of AvARRAY(av)
directly but via an API like av_fetch().

-- 
Standards (n). Battle insignia or tribal totems.

PR #19008, opened 2021-07-29
perlapio: binmode() and apply_layers() require pTHX

khwilliamson:
These two functions, unlike all the rest,  require the thread context
parameter to be explicitly passed.  This commit makes the  documentation
be accurate.

### DIFF
--- a/pod/perlapio.pod
+++ b/pod/perlapio.pod
@@ -62,9 +62,9 @@ perlapio - perl's IO abstraction interface.
   FILE   *PerlIO_findFILE(PerlIO *f);
   void    PerlIO_releaseFILE(PerlIO *f,FILE *stdio);
 
-  int     PerlIO_apply_layers(PerlIO *f, const char *mode,
+  int     PerlIO_apply_layers(pTHX_ PerlIO *f, const char *mode,
                                                     const char *layers);
-  int     PerlIO_binmode(PerlIO *f, int ptype, int imode,
+  int     PerlIO_binmode(pTHX_ PerlIO *f, int ptype, int imode,
                                                     const char *layers);
   void    PerlIO_debug(const char *fmt,...);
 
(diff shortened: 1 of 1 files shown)


Leont:
Question is: is that a documentation bug or a code bug? And if the latter can we still change it.

khwilliamson:
Personally, I think it is a code bug, but am unclear as to if it is too late to change

khwilliamson:
Is it too late to mark this as experimental, as a first step?

Leont:
> Is it too late to mark this as experimental, as a first step?

I don't think that would be helpful. Either we bite the bullet, or we don't.

leonerd:
I wouldn't want to suggest we change the API shape between perl v5.x versions. But when the time comes (hopefully in the not too distant future) that we consider a version bump to v7 I think that might permit this sort of API-neatening shape.

For now I'd suggest it ought to be a documentation issue, but perhaps we can start a list of these sorts of internal API fixes we feel we could get away with during a major version change.

pod/perlapio.pod
```diff
@@ -499,14 +499,14 @@ happened to C<read()> (or whatever) last time IO was requested.
 
 =over 4
 
-=item PerlIO_apply_layers(f,mode,layers)
+=item PerlIO_apply_layers(pTHX_ f,mode,layers)
```
tonycoz: This should probably be `aTHX_` since this shows how you're calling the function, not how the function is declared, as in the SYNOPSIS.

Similarly for the `=item` for  PerlIO_binmode().

pod/perlapio.pod
```diff
@@ -529,9 +529,9 @@ of passing NULL.)
 
 Portable cases are:
 
-    PerlIO_binmode(f,ptype,O_BINARY,NULL);
+    PerlIO_binmode(pTHX_ f,ptype,O_BINARY,NULL);
 and
-    PerlIO_binmode(f,ptype,O_TEXT,":crlf");
+    PerlIO_binmode(pTHX_ f,ptype,O_TEXT,":crlf");
```
tonycoz: This should also use `aTHX_`, since these are examples of calling PerlIO_binmode(), and they would fail to compile with `pTHX_` on a threaded build.

tonycoz (changes requested):

Leont (approved):

PR #19016, opened 2021-08-03
use CLANG_DIAG_IGNORE_STMT instead of GCC_DIAG_IGNORE_STMT

tonycoz:
It turns out gcc wasn't warning on this code, but older gcc (as
included in debian buster) *does* warn on the switch which it
doesn't recognise.

Newer gcc does recognise the -Wstring-compare switch, but it controls
warning on a different construct, so there's no reason to present it
to gcc.


khwilliamson:
I have been effectively using  this patch for some time locally and on smokes, due to the problems which it fixes.  So I'm confident it works, and given the research that Tony has done, there is no downside to applying it.

PR #19018, opened 2021-08-03
Use NV instead of hardcoded 'double' in POSIX::strtol()/strtoul().

t-a-k:
XS code for `POSIX::strtol()` and `POSIX::strtoul()` used to have `newSVnv((double)num)` constructs, but I think that this cast should be `(NV)`.

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -3549,7 +3549,7 @@ strtol(str, base = 0)
             num = strtol(str, &unparsed, base);
 #if IVSIZE < LONGSIZE
             if (num < IV_MIN || num > IV_MAX)
-                PUSHs(sv_2mortal(newSVnv((double)num)));
+                PUSHs(sv_2mortal(newSVnv((NV)num)));
             else
 #endif
                 PUSHs(sv_2mortal(newSViv((IV)num)));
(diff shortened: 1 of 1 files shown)


khwilliamson:
LGTM, @sisyphus how about you

sisyphus:
Yes, I would think that if <code>NVSIZE > 8</code>, then <code>(NV)num</code> DWIMs better than <code>(double)num</code>
And, of course, if <code>NVSIZE == 8</code>, then <code>(NV)num</code> and <code>(double)num</code> are one and the same.

The first thing that puzzles me is:

1) When will the conditions <code>(IVSIZE < LONGSIZE)</code> or <code>(UVSIZE < LONGSIZE)</code> ever be true ?
IME, the answer to that question is "never" - which makes testing of the proposed change rather difficult to perform ;-)
(I assume that IVSIZE/UVSIZE will always be sizeof(IV)/sizeof(UV) and that LONGSIZE is sizeof(long).)

The second thing that puzzles me is:

2) Can <code>newSVnv((NV)arg)</code> and <code>newSVnv(arg)</code> ever produce different results ?
In general, perhaps one form could throw an error that the other doesn't ?? ... though, in this case we know that "arg" is a "long".
Anyway, this second question is largely just idle curiosity as I, too, would include the specific (NV) cast, just to be sure.

t-a-k:
> 1. Can `newSVnv((NV)arg)` and `newSVnv(arg)` ever produce different results ?
>        In general, perhaps one form could throw an error that the other doesn't ?? ... though, in this case we know that "arg" is a "long".

These two seem to be compiled to the same code (I think the difference is only in explicitness of the cast), but the latter seems to trigger `-Wconversion` warning when `LONGSIZE` >= `NVSIZE`.
Though this warning is not enabled by default, I think that implicit casts between floating point and integer are not so good practice in C.

PR #19019, opened 2021-08-03
Create `defer` syntax and `OP_PUSHDEFER` opcode (RFC 0004)

leonerd:
Adds syntax `defer { BLOCK }` to create a deferred block; code that is deferred until the scope exits. This syntax is guarded by

    use feature 'defer';

Adds a new opcode, `OP_PUSHDEFER`, which is a LOGOP whose `op_other` field gives the start of an optree to be deferred until scope exit. That op pointer will be stored on the save stack and invoked as part of scope unwind.

Included is support for `B::Deparse` to deparse the optree back into syntax.

This PR relates to RFC 0004 - https://github.com/Perl/RFCs/blob/master/rfcs/rfc0004.md


pp_ctl.c
```diff
@@ -5434,6 +5457,47 @@ PP(pp_break)
     return cx->blk_givwhen.leave_op;
 }
 
+static void
+invoke_defer_block(pTHX_ void *_arg)
+{
+    OP *start = (OP *)_arg;
+    I32 was_cxstack_ix = cxstack_ix;
```
tonycoz, line 5465: this variable is unused in non-DEBUGGING builds

tonycoz:
Except for the unused variable, it looks good to me.

pp_ctl.c
leonerd: Ahyes. Now fixed in a488435fb4. (I'll squash before merging)

PR #19021, opened 2021-08-03
Only initialize threads::shared interpreter once

Leont:
Previously, the shared interpreter would be recreated every time the bootstrap was run, in the assumption that the bootstrap would only be run once. This assumption isn't necessarily true if multiple non-cloned interpreters exist.

Theoretically there's still a race condition around initialization, but I'm not particularly worried about that.


tonycoz (approved):
looks fine to me

Issue #18754, opened 2021-04-30
[EXPERIMENT] infix "isa" operator


rjbs:
Issue #17200 describes the plan for an infix "isa" operator, which was added to the language in v5.32.0 as an experiment.  This issue tracks the experiment toward its conclusion.

oodler577:
The utility of this not withstanding,  As an outside observer, I have a concern that the way in which it is implemented invites a whole host of additions. `trim` was implemented the same way. So I am not sure if this comment is even approprite here, but wanted to get it out there. The great @leonerd even wrote a very easy to follow recipe for this. On one hand, it's great to provide such an illustration; on the other hand, I don't think y'all want me clogging up the works with "experiments" through this avenue. So I guess I can rephrase as a question, is `isa` implemented in "right" way for current and future feature experiments? This could also be a good thing. And I do not know, which is simply the reason I am asking.

leonerd:
Context on `isa`: It's experimental due to standard "it's new" policy, but I don't have any particular plans to alter it in the future. I suspect it would still be useful to add a `does` operator for doing the `->DOES` role-type query, but that might be a separate issue.

As it stands I believe `isa` can be de-experimentalized after a suitable time - as yet to be decided.

leonerd:
Aside from a tiny correction in documentation, this feature remained unchanged through the entire 5.33 series, and is the same in 5.34.0 as when it was released in 5.32.0. If there are no significant changes by the time we release 5.36.0, I see no reason it needs to remain experimental.

I have no plans to make any further changes to it, so I suspect if the situation remains as it is, we should unmark it around the time of 5.35.6 or so - by my count that would be around January 2022.

leonerd:
A (currently-draft) PR to deêxperimentalise `isa` here: https://github.com/Perl/perl5/pull/19400

robrwo:
I am getting errors in Perl v5.34 with DateTime classes:
```perl
if ($date isa DateTime) { ... }
```
The error is
> Can't locate object method "isa" via package "DateTime" 

I'm having trouble reproducing this in a simple test script, but this error is preventing me from using the feature in one codebase. 

FWIW, it's with DateTime 1.57.

leonerd:
@robrwo Don't forget you still need to `use feature "isa"` to enable the parser to recognise the new operator.

```
$ perl -cE 'say "OK" if [] isa "ARRAY"'
Bareword found where operator expected at -e line 1, near "] isa"
        (Missing operator before isa?)
[...]

PR #19024, opened 2021-08-04
perlfunc 'ref' and UNIVERSAL: document isa operator alternative

Grinnz:
The (still experimental, but hopefully non-experimental by 5.36) isa operator provides an alternative to the cases mentioned here that avoids some of the tedium. https://github.com/Perl/perl5/issues/18754

### DIFF
--- a/lib/UNIVERSAL.pm
+++ b/lib/UNIVERSAL.pm
@@ -1,6 +1,6 @@
 package UNIVERSAL;
 
-our $VERSION = '1.13';
+our $VERSION = '1.14';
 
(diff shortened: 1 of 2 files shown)


leonerd:
LGTM

jkeenan:
> The (still experimental, but hopefully non-experimental by 5.36) isa operator provides an alternative to the cases mentioned here that avoids some of the tedium. #18754

LGTM, too -- but I wonder if we have unit tests in the test suite which precisely demonstrate your additions.

PR #19025, opened 2021-08-05
Porting/sync-with-cpan: handle weird tarball basedir names

ap:
This was failing to map new Text::Tabs releases properly because its distname is Text-Tabs+Wrap and interpolating it into a pattern without quoting causes the `+` to be misinterpreted as a quantifier.

### DIFF
--- a/Porting/sync-with-cpan
+++ b/Porting/sync-with-cpan
@@ -404,7 +404,7 @@ if ($$info {EXCLUDED}) {
 
 FILE: for my $file ( find_type_f( $new_dir )) {
     my $old_file = $file;
-    $file =~ s{^$new_dir/}{};
+    $file =~ s{^\Q$new_dir\E/}{};
 
     next if $EXCLUDED_QQ{$file};
     for my $qr (@EXCLUDED_QR) {


jkeenan:
> This was failing to map new Text::Tabs releases properly because its distname is Text-Tabs+Wrap and interpolating it into a pattern without quoting causes the `+` to be misinterpreted as a quantifier.

Locally I applied this pull request.  Then -- without applying https://github.com/Perl/perl5/pull/19026 -- I attempted these invocations:
```
$ perl Porting/sync-with-cpan Text::Tabs
Cannot find module Text::Tabs.
The available options are listed in the %Modules hash in Porting/Maintainers.pl

$ perl Porting/sync-with-cpan Text::Wrap
Cannot find module Text::Wrap.
The available options are listed in the %Modules hash in Porting/Maintainers.pl

$ perl Porting/sync-with-cpan "Text-Tabs+Wrap"
Cannot find Text::Tabs+Wrap on CPAN
```
So I don't yet see how this solves the problem.

Thank you very much.
Jim Keenan

ap:
Try
````
$ Porting/sync-with-cpan Text-Tabs+Wrap Text::Tabs
````

The first argument is the key by which it’s listed in `Porting/Maintainers.pl` and the second is the module name by which it’s listed in the CPAN index. If you don’t supply the second argument then the script will try to compute it from the first, but there is no direct mapping in case of this dist so it must be passed explicitly.

ap:
Oh, I missed this:

> without applying #19026

That pull request has no effect on this bug here, since the distname hasn’t changed.

Leont:
> Try
> 
> ```
> $ Porting/sync-with-cpan Text-Tabs+Wrap Text::Tabs
> ```
> 
> The first argument is the key by which it’s listed in `Porting/Maintainers.pl` and the second is the module name by which it’s listed in the CPAN index. If you don’t supply the second argument then the script will try to compute it from the first, but there is no direct mapping in case of this dist so it must be passed explicitly.

Sounds like that information should really also be in Maintainers.pl (though not needing it would be even better)

PR #19027, opened 2021-08-05
Porting/core-cpan-diff: Remove '--binary'

jkeenan:
Other things being equal, this program defaults to passing '-u --binary'
to the 'diff' command.  But there is no such option to 'diff' on either
Linux or FreeBSD.  Consequently, running the program spews more than
80,000 lines of error output.  Removing that option from code and
documentation makes the program DWIM.

### DIFF
--- a/Porting/core-cpan-diff
+++ b/Porting/core-cpan-diff
@@ -53,7 +53,8 @@ Usage: $0 [opts] [ -d | -v | -x ] [ -a | module ... ]
 -d/--diff     Display file differences using diff(1), rather than just
               listing which files have changed.
 
---diffopts    Options to pass to the diff command. Defaults to '-u --binary'.
+--diffopts    Options to pass to the diff command. Defaults to '-u --binary'
+              (except on *BSD, where it's just '-u').
 
 -f|force      Force download from CPAN of new 02packages.details.txt file
               (with --crosscheck only).
@@ -125,7 +126,10 @@ sub run {
           if ( $reverse || $use_diff || $diff_opts );
     }
     else {
-        $diff_opts = '-u --binary' unless defined $diff_opts;
+        #$diff_opts = '-u --binary' unless defined $diff_opts;
+        if (! defined $diff_opts) {
+            $diff_opts = ($^O =~ m/bsd$/i) ? '-u' : '-u --binary';
+        };
         usage("can't use -f without --crosscheck") if $force;
     }
 


jkeenan:
@karenetheridge, since you will be the next person using this program, can you take a look at the p.r.?

Thank you very much.
Jim Keenan

rjbs:
@jkeenan I'm confused!  I ran `perl Porting/cpan-core-diff -a` and the program seems to run correctly.  Also, `--binary` was added years ago.  Surely we'd have seen a problem if it was a non-option?

I looked at the --help and man page for diff, which do not list --binary.   So then I ran `diff --binary a b` and got this error:

```
dinah:~$ diff --binary
diff: missing operand after `--binary'
```

This is NOT the "unknown option" error.
```
dinah:~$ diff --binaryxyz
diff: unrecognized option `--binaryxyz'
diff: Try `diff --help' for more information.
```

Then, Heaven help me, I looked at the GNU diffutils source:  https://git.savannah.gnu.org/cgit/diffutils.git/tree/src/diff.c#n165

--binary is an option, although it is not documented.  I do not see it in the FreeBSD source.  Presumably it is in the Win32 diff, given the original commit, but I'm not sure.

Can you tell us more about the circumstance in which you saw these errors?

rjbs (commented):
I believe the premise of this change (there is no --binary in GNU diff) is incorrect, so I am not looking further at the source yet.

jkeenan:
> Other things being equal, this program defaults to passing '-u --binary'
> to the 'diff' command. But there is no such option to 'diff' on either
> Linux or FreeBSD. Consequently, running the program spews more than
> 80,000 lines of error output. Removing that option from code and
> documentation makes the program DWIM.

@rjbs, from your subsequent investigation, it appears as if there is an undocumented `--binary` option in the GNU `diff` program used on Linux, so there it runs without a lot of error messages.

However, on FreeBSD (my most frequent dev platform these days), `diff` is not GNU and does not have a `--binary` option.  Here's the first 20 lines of output from running `perl Porting/core-cpan-diff -a 2>&1 | tee out`:
```
$ cat freebsd.porting.core-cpan-diff.output.head20.txt

Archive::Tar - BINGOS/Archive-Tar-2.40.tar.gz
  upstream is: cpan
diff: unrecognized option `--binary'
usage: diff [-aBbdilpTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            [-I pattern] [-L label] file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            -C number file1 file2
       diff [-aBbdiltw] [-I pattern] [--ignore-case] [--no-ignore-case]
            [--normal] [--strip-trailing-cr] [--tabsize] -D string file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [--strip-trailing-cr]
            -U number file1 file2
       diff [-aBbdilNPprsTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [-I pattern] [-L label]
            [-S name] [-X file] [-x pattern] dir1 dir2
       diff [-aBbditwW] [--expand-tabs] [--ignore-all-blanks]
            [--ignore-blank-lines] [--ignore-case] [--minimal]
```
I'll send you the full output via email.

rjbs:
My suggestion at this point is "add --binary unless $^O eq 'freebsd'".  It was added to solve a problem (it seems), and that can keep working when the switch is available.

jkeenan:
> My suggestion at this point is "add --binary unless $^O eq 'freebsd'". It was added to solve a problem (it seems), and that can keep working when the switch is available.

Patch revised.  Tested with favorable results on Linux, FreeBSD and OpenBSD.  Please re-review.

rjbs (approved):
Looks good to me.

PR #19028, opened 2021-08-05
Porting/sync-with-cpan: document edge cases

jkeenan:
Those edge cases being elements in %Modules whose keys are not spelled
with standard package (Some::Module) syntax.

In the code, make explicit where we are shifting off @ARGV.

This is a partial response to problems observed in
https://github.com/Perl/perl5/pull/19025/.

### DIFF
--- a/Porting/sync-with-cpan
+++ b/Porting/sync-with-cpan
@@ -7,10 +7,13 @@ Porting/sync-with-cpan - Synchronize with CPAN distributions
 =head1 SYNOPSIS
 
     sh ./Configure
+
+Then, in most cases:
+
     perl Porting/sync-with-cpan <module>
 
-where C<module> is the name it appears in the C<%Modules> hash
-of F<Porting/Maintainers.pl>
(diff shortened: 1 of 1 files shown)


ap:
I think listing the specific pieces of data in the POD just creates a future maintenance hazard. The information about Text-Tabs+Wrap in `Porting/Maintainers.pl` was already out of sync with the version of the module in the tree which had been updated manually, and that’s with data used by code. Requiring whoever updates the special-case distributions in the future to also remember to check if the unstructured info in the POD is still correct is even worse. I concur with @Leont’s comment in #19025 that this information should be stored in `Porting/Maintainers.pl` and am preparing an alternative PR that does just that.

ap:
> [I] am preparing an alternative PR that does just that.

Done. Please consider applying #19030 instead of this PR.

bingos:
I concur with @ap it is much better to deal with this programmatically.

jkeenan:
Closed in favor of https://github.com/Perl/perl5/pull/19030.

Issue #19022, opened 2021-08-04
threads::shared is leaking the shared interpeter


Leont:
`threads::shared` uses a shared interpreter to store the shared values in. AFAICT this interpreter is never destroyed or deallocated, it probably should be but the timing is fragile: it must happen after the last interpreter that uses threads::shared is destroyed, but before `PERL_TERM` is called. It seems we're lacking a shutdown hook.

jkeenan:
> this interpreter is never destroyed or deallocated,

How would I go about demonstrating this (on various platforms)?

Thank you very much.
Jim Keenan

Leont:
I guess valgrind would have a field day on it, but it's mainly obvious by the lack of any call to `perl_destroy` or `perl_free` in `shared.xs`

jkeenan:
> I guess valgrind would have a field day on it, but it's mainly obvious by the lack of any call to `perl_destroy` or `perl_free` in `shared.xs`

Thanks.

PR #19031, opened 2021-08-06
Add PL_shutdownhook

Leont:
This is a proposed solution to #19022

This adds a hook that is run early in `PERL_SYS_TERM` to terminate global resources, and uses it to destroy the shared interpreter in `threads::shared`

### DIFF
--- a/dosish.h
+++ b/dosish.h
@@ -31,10 +31,10 @@
  * to work, but must NOT be retained in production code. */
 #ifndef PERL_SYS_TERM_BODY
 #  define PERL_SYS_TERM_BODY()                         \
-    ENV_TERM; USER_PROP_MUTEX_TERM; LOCALE_TERM;       \
-    HINTS_REFCNT_TERM; KEYWORD_PLUGIN_MUTEX_TERM;      \
-    OP_CHECK_MUTEX_TERM; OP_REFCNT_TERM;               \
-    PERLIO_TERM; MALLOC_TERM; 
+    SHUTDOWN_TERM; ENV_TERM; USER_PROP_MUTEX_TERM;      \
+    LOCALE_TERM;INTS_REFCNT_TERM;                      \
+    KEYWORD_PLUGIN_MUTEX_TERM;OP_CHECK_MUTEX_TERM;     \
+    OP_REFCNT_TERM;PERLIO_TERM; MALLOC_TERM;
 #endif
 #define dXSUB_SYS dNOOP
 
--- a/dist/threads-shared/lib/threads/shared.pm
+++ b/dist/threads-shared/lib/threads/shared.pm
@@ -8,7 +8,7 @@ use Config;
 
 use Scalar::Util qw(reftype refaddr blessed);
 
-our $VERSION = '1.69'; # Please update the pod, too.
+our $VERSION = '1.70'; # Please update the pod, too.
 my $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
@@ -196,7 +196,7 @@ threads::shared - Perl extension for sharing data structures between threads
 
 =head1 VERSION
 
-This document describes threads::shared version 1.68
+This document describes threads::shared version 1.70
 
 =head1 SYNOPSIS
 
--- a/dist/threads-shared/shared.xs
+++ b/dist/threads-shared/shared.xs
@@ -1284,6 +1284,21 @@ S_shared_signal_hook(pTHX) {
 }
 #endif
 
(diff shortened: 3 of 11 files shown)


iabyn:
On Fri, Aug 06, 2021 at 07:57:06AM -0700, Leon Timmermans wrote:
> This is a proposed solution to #19022
> 
> This adds a hook that is run early in `PERL_SYS_TERM` to terminate global resources, and uses it to destroy the shared interpreter in `threads::shared`

Any particular reason PL_shutdownhook  is manually declared in perl.h
rather than being in perlvars.h like other global vars?

-- 
Never do today what you can put off till tomorrow.

Leont:
> Any particular reason PL_shutdownhook is manually declared in perl.h rather than being in perlvars.h like other global vars?

I didn't know about that, but that does seem like a better idea. I copied it from something else that also wasn't in perlvars.h (but I forgot what)

Leont:
> Any particular reason PL_shutdownhook is manually declared in perl.h rather than being in perlvars.h like other global vars?

Fixed that.

dosish.h
```diff
@@ -49,7 +49,7 @@
 
 #ifndef PERL_SYS_TERM_BODY
 #  define PERL_SYS_TERM_BODY()                         \
-    HINTS_REFCNT_TERM; KEYWORD_PLUGIN_MUTEX_TERM;      \
+    SHUTDOWN_TERM; HINTS_REFCNT_TERM; KEYWORD_PLUGIN_MUTEX_TERM;      \
```
leonerd: Most minor of nitpicks, but can we reflow these words and reälign the resulting `\` escapes, to keep it neat?

leonerd:
Aside from my comment about reflowing a macro, this LGTM

dosish.h
Leont: We could, but to be honest I was about to open a topic about restructuring `PERL_SYS_TERM_BODY` entirely. There are several definitions of it, but they're 90% identical. That seems suboptimal.

tonycoz:
could you please document PL_shutdownhook?

khwilliamson:
So, are we just waiting for @Leont  to document this before it is ready to merge?

tonycoz:
```
* leont is not sure where/how to document PL_shutdownhook, given that 
          none of the other similar hooks are documented.
```
I'm not so worried about where, more about adding another undocumented public hook.

Things to document:
- when is the hook called?  Do I need to chain to the old hook?
- what do I need to lock to set it? (though a simple API to set (or add/remove a given function pointer) it might be better, since hooks have led to some bugs in the past)

khwilliamson:
@leont, this looks close to being ready

jkeenan:
> @Leont, this looks close to being ready

@LeonT, can we get an update on the status of this pull request?

Leont:
> @Leont, can we get an update on the status of this pull request?

I'm adapting it to tony's suggestion (simple API to set it), will update it soon

demerphq:
@leont last message in this PR is that you were adapting it to feedback. Do you still want to keep this PR? (It seems reasonable to me in principal).

Leont:
> Do you still want to keep this PR? (It seems reasonable to me in principal).

Yeah, it's a problem that needs a solution. I'll try to pick it up again soon.

demerphq:
@Leont I'm happy to get it rebased for you considering you seem busy of late.

Leont:
> I'm happy to get it rebased for you considering you seem busy of late.

It needs much more than a rebase, otherwise this would have been merged already. I'll try to finish this soon.

jkeenan:
> > I'm happy to get it rebased for you considering you seem busy of late.
> 
> It needs much more than a rebase, otherwise this would have been merged already. I'll try to finish this soon.

@Leont, have you been able to move any of this work forward? Thanks.

jkeenan:
Needs a version bump; otherwise, progressing well.

Leont:
> I'm not so worried about where, more about adding another undocumented public hook.

I've been thinking about how to make this prettier, but I now think we should make it non-public instead.

This is needed for `threads::shared`, and `threads::shared` only. Just let it be a private thing.

tonycoz:
> > I'm not so worried about where, more about adding another undocumented public hook.
> 
> I've been thinking about how to make this prettier, but I now think we should make it non-public instead.
> 
> This is needed for `threads::shared`, and `threads::shared` only. Just let it be a private thing.

That's fine, as long as we document that: don't use this, if you do and it breaks you get to keep the pieces.

tonycoz (approved):

PR #19032, opened 2021-08-06
Add a phase_name() macro.

FGasper:


perl.h
```diff
@@ -5427,6 +5427,25 @@ EXTCONST char *const PL_phase_names[] = {
...
+For example, a warning that includes the current interpreter phase
+might look thus:
+
+    const char* phase_name = perl_phase_name(PL_phase);
+    warn("This is weird. (Perl phase: %s)", phase_name);
+
+=cut
+*/
+
+#define perl_phase_name(phase) (PL_phase_names[phase])
```
leonerd: It is very rare (I want to say "never the case", but maybe there are a few) that visible macros take the `perl_` prefix on them. I'd suggest this should be called simply `phase_name()`.

FGasper: @leonerd I was hoping to retain the “namespacing” that `PL_phase` has. `PL_` is a no-no, though.

Things like form() and mess() are hard to search for. Much easier if there’s some sort of prefix on there.

I’ll change if you still think it better, but given that there are interfaces concerning the interpreter that have the `perl_` prefix it seems a bit more helpful to me to leave that on.

perl.h
```diff
@@ -5427,6 +5427,25 @@ EXTCONST char *const PL_phase_names[] = {
...
+    const char* phase_name = perl_phase_name(PL_phase);
+    mess("This is weird. (Perl phase: %s)", phase_name);
+
+=cut
+*/
+
+#define perl_phase_name(phase) (PL_phase_names[phase])
+
 #else
 EXTCONST char *const PL_phase_names[];
```
smpetershm: You'll need a no-op version of your macro here otherwise you'll break code.

FGasper: What would the macro do? Return NULL, empty-string, or … ?

It seems like in this case there’s no concept of a phase name, so anything that wants to access one should fail at compile time … but I’m probably misunderstanding something … ?

smpetershm: I'm hoping you would have the answers to those questions.  You've presented a use case for a new macro, but here's a case where the data you want doesn't exist.

Right now, this change would break things in a Perl where `DOINIT` is not defined.  There's probably some investigating to see where that is defined and what the differences are.

FGasper: Actually, the fix is to move the macro declaration out of that block.

That `DOINIT` check appears to be there so that perl.h works as well in Perl itself as in XS modules. In the former case we need to define the constant, whereas in the latter we just have to declare it (since Perl itself supplies the values).

Thank you!

perl.h
smpetershm: `perl_` has been used in the code in the past to signify "here's our unique twist of an external thing".  It does have meaning and using it here should be avoided.

FGasper: @smpeters `perl_` in the API (cf. `perldoc api`) seems to indicate something that pertains to an interpreter, e.g., `perl_clone`, `perl_alloc`, etc.

I’ve gone and removed the prefix. I still think it would be more sensible left in, but it’s not worth the back-and-forth.

Thanks!

iabyn:
On Fri, Aug 06, 2021 at 12:22:03PM -0700, Felipe Gasper wrote:
> I’ve gone and removed the prefix. I still think it would be more sensible left in, but it’s not worth the back-and-forth.

It needs a prefix to avoid polluting global namespaces (perl.h is included
by all sorts of code) but the prefix should be Perl_ not perl_.Its a macro
pretending to be a perl function (and indeed might well be replaced by a
function at some point in the future).

-- 
    -- The Diary of Samuel Pepys

FGasper:
@iabyn 
> It needs a prefix to avoid polluting global namespaces (perl.h is included by all sorts of code) but the prefix should be Perl_ not perl_.Its a macro pretending to be a perl function (and indeed might well be replaced by a function at some point in the future).

@leonerd @smpetershm Do you agree? `Perl_` is the prefix on the `pTHX`-accepting function behind macros like warn(), form(), and mess(), so that seems a funny choice to me, but whatever. :)

FWIW the API already includes:
- perl_clone
- perl_alloc
- perl_construct
- perl_free
- perl_parse
- perl_run

tonycoz:
> It needs a prefix to avoid polluting global namespaces (perl.h is included by all sorts of code) but the prefix should be Perl_ not perl_.Its a macro pretending to be a perl function (and indeed might well be replaced by a function at some point in the future).

I think this horse has long since bolted and the stable has burned down.

I'm not fond of the namespace pollution from the perl headers, many of the macros currently defined, eg. any `my_*()` function, warn(), croak(), get_context(), block_end(), filter_add(), form(), have the potential to conflict with other code. That's ignoring integrating with code that has similar concepts to perl, `op_*()`, anything else that coincidentally has a "sv", "av" or "hv" type of some sort (or a library named that way.)

I many ways I'd love to see us requiring a prefix on new API (and probably EXT) visible names, though that's going to make using short-ish names like SvISBOOL() (plSvIsBOOL()? PL_SvISBOOL()?) clumsier to use. Assuming new names go into embed.fnc this should be enforceable with a porting test.

Until then, I don't think this name violates existing practice.

If it does become a function, I expect the function will be called Perl_phase_name() and like all of the API macros the "Perl_" won't be included the the name people generally use.

tonycoz:
The commit message still refers to the perl_ prefixed name.

Ideally, the name should have an entry in embed.fnc

FGasper:
Pushed up with `embed.fnc` entry added and commit message revised.

tonycoz:
Ran make regen and squashed it, applied as eb87cedb

Issue #19029, opened 2021-08-05
pod/perlapi.pod missing one entry; build-time warning being generated


jkeenan:
Recently the following build-time warning has begun to be emitted during `make`:
```     
Empty pod for newAV_alloc_x (from av.h) at autodoc.pl line 1215.
```
This warning calls attention to this flaw in auto-generated `pod/perlapi.pod`:
```
  761 =for hackers
  762 Found in file av.h
  763 
  764 =over 4
  765 
  766 =item C<newAV_alloc_x>
  767 X<newAV_alloc_x>
  768 
  769 
  770 
  771 =over 3
  772 
  773  AV*  newAV_alloc_x(SSize_t size)
  774 
  775 =back
  776 
  777 =back
  778 
```
The relevant area in `av.h` is:
```
110 
111 /*
112 =for apidoc newAV_alloc_x
113 =for apidoc newAV_alloc_xz
114 
115 Similar to newAV(), but a SV* array is also allocated.
116 
```
I suspect that this warning was accidentally introduced in: 6e7a7813c39

I further suspect that we've encountered this problem previously.  Does anyone recall the fix?

Thank you very much.
Jim Keenan

khwilliamson:
@richardleach is working on it, but busy

jkeenan:
> @richardleach is working on it, but busy

Using [Devel::Git::MultiBisect::BuildTransitions](https://metacpan.org/pod/Devel::Git::MultiBisect::BuildTransitions), I was able to learn that this problem was cleared up with the following commit:
```
commit 48d21358f7bd06d60aaa4f913cd417b5eafa3fc0
Author:     Richard Leach <richardleach@users.noreply.github.com>
AuthorDate: Sun Aug 8 21:53:33 2021 +0100
Commit:     Karl Williamson <khw@cpan.org>
CommitDate: Fri Aug 27 07:12:53 2021 -0600

    av.h - better document newAV_alloc_x/newAV_alloc_xz

```
Closing ticket.  Thank you very much.

PR #19038, opened 2021-08-08
av.h - better document newAV_alloc_x/newAV_alloc_xz

richardleach:
PR Note: This is a revised version of #19006. The text is as previously proposed, other than I've tried to take into account @iabyn's concern that detailing exactly how array elements are initialized might unnecessarily complicate changes in the future.

### DIFF
--- a/av.h
+++ b/av.h
@@ -98,32 +98,46 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
 
 /*
 =for apidoc newAV
+=for apidoc_item newAV_alloc_x
+=for apidoc_item newAV_alloc_xz
 
-Creates a new AV.  The reference count is set to 1.
+These all create a new AV, setting the reference count to 1. They differ
+in when and how an accompanying SV* array is allocated and initialized.
 
-Perl equivalent: C<my @array;>.
+The Perl equivalent is approximately C<my @array;>.
 
-=cut
-*/
-
-#define newAV()	MUTABLE_AV(newSV_type(SVt_PVAV))
-
-/*
-=for apidoc newAV_alloc_x
-=for apidoc newAV_alloc_xz
-
-Similar to newAV(), but a SV* array is also allocated.
-
-This is similar to but more efficient than doing:
-
+newAV does not allocate a SV* array.
     AV *av = newAV();
+
+This is very useful when an AV is required, but population of it may be
+deferred, or even never actually take place. (Memory is not allocated
+unnecessarily.)
+
+Subsequent SV* array allocation would be performed via av_extend. This
+might be called directly:
     av_extend(av, key);
 
-Note that the actual size requested is allocated. This is unlike
-av_extend(), which takes the maximum desired array index (AvMAX) as its
-"key" parameter, and enforces a minimum value for that of 3.
+Or it might be called implicitly when the first element is stored:
+    (void)av_store(av, 0, sv);
 
-In other words, the following examples all result in an array that can
+Whether or not any as-yet-untouched array elements are initialized by
+av_extend depends upon whether the array is "real" at the time.
+
+In contrast, when an AV is created for immediate population with a known
+(or likely) number of elements, it is more efficient to immediately
+allocate a SV* array of the necessary size. (This avoids inefficient use
+of av_extend and the potential for the first allocation being too small
+and then having to resize it.)
+
+For that scenario, newAV_alloc_x and newAV_alloc_xz can be used to create
+an AV and allocate a SV* array to fit the specified number of elements.
+(As a result, these macros MUST NOT be called with a size less than 1.)
+
+newAV_alloc_x does not initialize the array elements - and so the
+expectation is that all should be initialized elsewhere prior to any
+potentials reads. newAV_alloc_xz does initialize the array elements.
+
+As an example, the following examples all result in an array that can
 fit four elements (indexes 0 .. 3):
 
     AV *av = newAV();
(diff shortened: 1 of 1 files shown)


av.h
```diff
@@ -98,32 +98,46 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
 
-Note that the actual size requested is allocated. This is unlike
-av_extend(), which takes the maximum desired array index (AvMAX) as its
-"key" parameter, and enforces a minimum value for that of 3.
+newAV_alloc_x does not initialize the array elements - and so the
+expectation is that all should be initialized elsewhere prior to any
+potentials reads. newAV_alloc_xz does initialize the array elements.
 
-In other words, the following examples all result in an array that can
+As an example, the following examples all result in an array that can
```
khwilliamson: example occurring twice so close together reads awkwardly

av.h
```diff
@@ -98,32 +98,46 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+might be called directly:
+    av_extend(av, key);
 
-Similar to newAV(), but a SV* array is also allocated.
+Or it might be called implicitly when the first element is stored:
+    (void)av_store(av, 0, sv);
 
-This is similar to but more efficient than doing:
+Whether or not any as-yet-untouched array elements are initialized by
+av_extend depends upon whether the array is "real" at the time.
```
khwilliamson: I have a vague idea of what "real" is, but have never had to deal with it.  I think most readers will have even less of an idea.  Hopefully there is a discussion of this elsewhere that you can link to

khwilliamson (commented):
Otherwise LGTM

jkeenan:
Marginal note:  It appears this branch will clear up the build-time warning reported in https://github.com/Perl/perl5/issues/19029.  I'll double-check once this is merged.

av.h
richardleach: Yes, I'll remove the leading "As an example, " there. Thanks.

av.h
richardleach: Hmm, I've not really seen such a discussion and so only have my own piecemeal picture. Perhaps it would be better for this PR to more vaguely state something like:

"Unused array elements are typically initialized by av_extend, but this does not occur for some special arrays."

Even if that might just leave the reader to wonder what's special about them...

richardleach: @khwilliamson - I've reworded that sentence to avoid specifics. What do you think?

av.h
```diff
@@ -98,54 +98,72 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
-Perl equivalent: C<my @array;>.
+The Perl equivalent is approximately C<my @array;>.
 
-=cut
-*/
+newAV does not allocate a SV* array.
+    AV *av = newAV();
 
-#define newAV()	MUTABLE_AV(newSV_type(SVt_PVAV))
+This is very useful when an AV is required, but population of it may be
```
khwilliamson: I think "but populating it' is better phrasing

av.h
```diff
@@ -98,54 +98,72 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+deferred, or even never actually take place. (Memory is not allocated
+unnecessarily.)
 
-/*
-=for apidoc newAV_alloc_x
-=for apidoc newAV_alloc_xz
+Subsequent SV* array allocation would be performed via av_extend. This
```
khwilliamson: Need to link to first mentions of other operations, and e,g, for subsequent mentions use C<av_extend>

av.h
```diff
@@ -98,54 +98,72 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+might be called directly:
+    av_extend(av, key);
 
-Similar to newAV(), but a SV* array is also allocated.
+Or it might be called implicitly when the first element is stored:
+    (void)av_store(av, 0, sv);
 
-This is similar to but more efficient than doing:
+Unused array elements are typically initialized by av_extend, but this
+is undesirable and will not be the case for some specific arrays.
```
khwilliamson: Where do they go to find the rules about which specific arrays and what

av.h
```diff
@@ -98,54 +98,72 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+
+For that scenario, newAV_alloc_x and newAV_alloc_xz can be used to create
+an AV and allocate a SV* array to fit the specified number of elements.
+(As a result, these macros MUST NOT be called with a size less than 1.)
 
-Note that the actual size requested is allocated. This is unlike
-av_extend(), which takes the maximum desired array index (AvMAX) as its
-"key" parameter, and enforces a minimum value for that of 3.
+newAV_alloc_x does not initialize the array elements - and so the
+expectation is that all should be initialized elsewhere prior to any
```
khwilliamson: s/should/will/
I think

av.h
```diff
@@ -98,54 +98,72 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
 
 /*
 =for apidoc newAV
+=for apidoc_item newAV_alloc_x
+=for apidoc_item newAV_alloc_xz
 
-Creates a new AV.  The reference count is set to 1.
+These all create a new AV, setting the reference count to 1. They differ
+in when and how an accompanying SV* array is allocated and initialized.
```
khwilliamson: how about "They differ in allocating and populating the array of SV*s that always accompanies a non-empty AV."
or something similar

av.h
richardleach: Ok, happy to change that.

av.h
richardleach: Ok. Do the indented mentions need markup though, or are they fine bare?

av.h
richardleach: Those specific arrays will be the non-`AvREAL()` ones, which don't seem to be described anywhere. I'm happy to have a go at documenting "real/unreal" arrays (in _perlguts_ or in code comments?)  but would rather not do as part of this PR.

For now, should it just say:
`Unused array elements are typically initialized by av_extend.`?

av.h
richardleach: Works for me.

av.h
richardleach: I've changed it to something similar.

av.h
richardleach: Or we hold this PR until AvREAL is documented, and then go back to the original text of:
```
Whether or not any as-yet-untouched array elements are initialized by
av_extend depends upon whether the array is "real" at the time.
```

richardleach: I take that back, there are some comments at the top of `av.h`. I'll have a look at converting them to apidoc and expanding them.

richardleach: I've converted this sentence to:
```
Unused array elements are typically initialized by C<av_extend>. (Only
core maintainers should have need to concern themseleves with when that
is not the case. Refer to F<av.h> and F<av.c> for the differences between 
real and fake AVs.)
```
And opened https://github.com/Perl/perl5/pull/19062 as a separate concern for the documentation of real/fake AVs.

PR #19040, opened 2021-08-09
Stable tracking of "boolean intent" across SVs

leonerd:
It is likely this PR should be split into two parts; a low-level "implement static const COW", and a high level "use it to make booleans stable". I'll take suggestions from reviewers as to whether to leave it as one (squashed) commit, form two commits, or even split it into two PRs entirely to implement the to halves.

The overall result is that it is now possible for modules like data printers, formatters, serialisers, etc... to distinguish which Perl values are intended to be booleans, and can take special action on them to render them differently from numbers or strings.

## Low-level

The low-level part is basically this one commit:

Define a third kind of COW state; STATIC

Previously, when IsCOW flag was set there were two cases:
  * SvLEN()==0:
    >  PV is really a shared HEK
 * SvLEN()!=0:
    >  PV is a COW structure with 1..256 refcount stored in its extra final byte

This change adds a third state:

 * SvLEN()==0 && (SvFLAGS() & SVppv_STATIC):
   >  PV is a shared static const pointer and must not be modified

`sv_setsv_flags()`, `sv_setsv_cow()` and `rvpv_dup()` will preserve this state.

`sv_uncow()` will copy it out to a regular string buffer, which means that `SV_THINKFIRST` will undo this and result in a regular mutable copy before any edit operations.

The upshot here is that we can now store static strings inside PVs without having to copy the buffer of them. This can help save memory and CPU time when copying them around. It is likely to be useful independently of the higher-level change; for example when storing large static strings of message text or other data, or as a general-purpose way to pass pointers around in a way that does not get broken even by thread-cloning.

## High Level

The higher part is then everything else that builds atop it. It uses these new static strings to store the `PL_Yes` and `PL_No` buffers inside the boolean constant SVs, and because these strings are copied by pointer reference, these pointers remain stable into any SV intended to be a boolean value, even across threads. A simple pointer comparison can then be used to implement the `SvIsBOOL` macro, which is exposed via `Scalar::Util` to be useful to pureperl code.

### DIFF
--- a/cpan/Scalar-List-Utils/ListUtil.xs
+++ b/cpan/Scalar-List-Utils/ListUtil.xs
@@ -1666,6 +1666,19 @@ PPCODE:
 MODULE=List::Util       PACKAGE=Scalar::Util
 
 void
+isbool(sv)
+    SV *sv
+PROTOTYPE: $
+CODE:
+#ifdef SvIsBOOL
+    SvGETMAGIC(sv);
+    ST(0) = boolSV(SvIsBOOL(sv));
+    XSRETURN(1);
+#else
+    croak("stable boolean values are not implemented in this release of perl");
+#endif
+
+void
 dualvar(num,str)
     SV *num
     SV *str
@@ -2101,7 +2114,7 @@ BOOT:
     HV *lu_stash = gv_stashpvn("List::Util", 10, TRUE);
     GV *rmcgv = *(GV**)hv_fetch(lu_stash, "REAL_MULTICALL", 14, TRUE);
     SV *rmcsv;
-#if !defined(SvWEAKREF) || !defined(SvVOK)
+#if !defined(SvWEAKREF) || !defined(SvVOK) || !defined(SvIsBOOL)
     HV *su_stash = gv_stashpvn("Scalar::Util", 12, TRUE);
     GV *vargv = *(GV**)hv_fetch(su_stash, "EXPORT_FAIL", 11, TRUE);
     AV *varav;
@@ -2118,6 +2131,9 @@ BOOT:
 #endif
 #ifndef SvVOK
     av_push(varav, newSVpv("isvstring",9));
+#endif
+#ifndef SvIsBOOL
+    av_push(varav, newSVpv("isbool",6));
 #endif
 #ifdef REAL_MULTICALL
     sv_setsv(rmcsv, &PL_sv_yes);
--- a/cpan/Scalar-List-Utils/t/boolean-thr.t
+++ b/cpan/Scalar-List-Utils/t/boolean-thr.t
@@ -1,0 +1,38 @@
+#!./perl
+
+use strict;
+use warnings;
+
+use Config ();
+use Scalar::Util ();
+use Test::More
+    (grep { /isbool/ } @Scalar::Util::EXPORT_FAIL) ? (skip_all => 'isbool is not supported on this perl') :
+    (!$Config::Config{usethreads})                 ? (skip_all => 'perl does not support threads') :
+    (tests => 5);
+
+use threads;
+use threads::shared;
+
+Scalar::Util->import("isbool");
+
+ok(threads->create( sub { isbool($_[0]) }, !!0 )->join,
+    'value in to thread is bool');
+
+ok(isbool(threads->create( sub { return !!0 } )->join),
+    'value out of thread is bool');
+
+{
+    my $var = !!0;
+    ok(threads->create( sub { isbool($var) } )->join,
+        'variable captured by thread is bool');
+}
+
+{
+    my $sharedvar :shared = !!0;
+
+    ok(isbool($sharedvar),
+        ':shared variable is bool outside');
+
+    ok(threads->create( sub { isbool($sharedvar) } )->join,
+        ':shared variable is bool inside thread');
+}
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -1083,7 +1083,7 @@ PP(pp_multiconcat)
                     if (!SvOK(right)) {
                         if (ckWARN(WARN_UNINITIALIZED))
                             report_uninit(right);
-                        sv_setsv_flags(left, &PL_sv_no, 0);
+                        sv_setbool(left, FALSE);
                     }
                     else
                         sv_setsv_flags(left, right, 0);
--- a/sv.h
+++ b/sv.h
@@ -462,6 +462,8 @@ perform the upgrade if necessary.  See C<L</svtype>>.
 /* Some private flags. */
 
 
+/* scalar SVs with SVp_POK */
+#define SVppv_STATIC    0x40000000 /* PV is pointer to static const; must be set with SVf_IsCOW */
 /* PVAV */
 #define SVpav_REAL	0x40000000  /* free old entries */
 /* PVHV */
@@ -2008,10 +2010,11 @@ scalar.
     )								\
 )
 
-#define SvIsCOW(sv)		(SvFLAGS(sv) & SVf_IsCOW)
-#define SvIsCOW_on(sv)		(SvFLAGS(sv) |= SVf_IsCOW)
-#define SvIsCOW_off(sv)		(SvFLAGS(sv) &= ~SVf_IsCOW)
-#define SvIsCOW_shared_hash(sv)	(SvIsCOW(sv) && SvLEN(sv) == 0)
+#define SvIsCOW(sv)              (SvFLAGS(sv) & SVf_IsCOW)
+#define SvIsCOW_on(sv)           (SvFLAGS(sv) |= SVf_IsCOW)
+#define SvIsCOW_off(sv)          (SvFLAGS(sv) &= ~(SVf_IsCOW|SVppv_STATIC))
+#define SvIsCOW_shared_hash(sv)  ((SvFLAGS(sv) & (SVf_IsCOW|SVppv_STATIC)) == (SVf_IsCOW) && SvLEN(sv) == 0)
+#define SvIsCOW_static(sv)       ((SvFLAGS(sv) & (SVf_IsCOW|SVppv_STATIC)) == (SVf_IsCOW|SVppv_STATIC))
 
 #define SvSHARED_HEK_FROM_PV(pvx) \
         ((struct hek*)(pvx - STRUCT_OFFSET(struct hek, hek_key)))
--- a/sv.h
+++ b/sv.h
@@ -1052,6 +1052,17 @@ Remove any string offset.
     ((SvFLAGS(sv) & (SVf_POK|SVf_UTF8|SVf_IOK|SVf_NOK|SVf_ROK|SVpgv_GP|SVf_THINKFIRST|SVs_GMG)) == SVf_POK)
 
 /*
+=for apidoc Am|BOOL|SvIsBOOL|SV* sv
+
+Returns true if the SV is one of the special boolean constants (PL_sv_yes or
+PL_sv_no), or is a regular SV whose last assignment stored a copy of one.
+
+=cut
+*/
+
+#define SvIsBOOL(sv)            Perl_sv_isbool(aTHX_ sv)
+
+/*
 =for apidoc Am|U32|SvGAMAGIC|SV* sv
 
 Returns true if the SV has get magic or
@@ -2333,6 +2344,21 @@ See also C<L</PL_sv_yes>> and C<L</PL_sv_no>>.
 */
 
 #define boolSV(b) ((b) ? &PL_sv_yes : &PL_sv_no)
+
+/*
+=for apidoc Am|void|sv_setbool|SV *sv|bool b
+=for apidoc_item |void|sv_setbool_mg|SV *sv|bool b
+
+These set an SV to a true or false boolean value, upgrading first if necessary.
+
+They differ only in that C<sv_setbool_mg> handles 'set' magic; C<sv_setbool>
+does not.
+
+=cut
+*/
+
+#define sv_setbool(sv, b)     sv_setsv(sv, boolSV(b))
+#define sv_setbool_mg(sv, b)  sv_setsv_mg(sv, boolSV(b))
 
 #define isGV(sv) (SvTYPE(sv) == SVt_PVGV)
 /* If I give every macro argument a different name, then there won't be bugs
--- a/MANIFEST
+++ b/MANIFEST
@@ -1977,6 +1977,8 @@ cpan/Scalar-List-Utils/multicall.h		Util extension
 cpan/Scalar-List-Utils/t/00version.t		Scalar::Util
 cpan/Scalar-List-Utils/t/any-all.t		List::Util
 cpan/Scalar-List-Utils/t/blessed.t		Scalar::Util
+cpan/Scalar-List-Utils/t/boolean.t		Scalar::Util
+cpan/Scalar-List-Utils/t/boolean-thr.t		Scalar::Util
 cpan/Scalar-List-Utils/t/dualvar.t		Scalar::Util
 cpan/Scalar-List-Utils/t/exotic_names.t
 cpan/Scalar-List-Utils/t/first.t		List::Util
--- a/MANIFEST
+++ b/MANIFEST
@@ -5830,6 +5830,7 @@ t/op/auto.t			See if autoincrement et all work
 t/op/avhv.t			See if pseudo-hashes work
 t/op/bless.t			See if bless works
 t/op/blocks.t			See if BEGIN and friends work
+t/op/bool.t			Check misc details of boolean values
 t/op/bop.t			See if bitops work
 t/op/caller.pl			Tests shared between caller.t and XS op.t
 t/op/caller.t			See if caller() works
--- a/cpan/Scalar-List-Utils/lib/List/Util.pm
+++ b/cpan/Scalar-List-Utils/lib/List/Util.pm
@@ -16,7 +16,7 @@ our @EXPORT_OK  = qw(
   sample shuffle uniq uniqint uniqnum uniqstr zip zip_longest zip_shortest mesh mesh_longest mesh_shortest
   head tail pairs unpairs pairkeys pairvalues pairmap pairgrep pairfirst
 );
-our $VERSION    = "1.56";
+our $VERSION    = "1.56_001";
 our $XS_VERSION = $VERSION;
 $VERSION =~ tr/_//d;
 
--- a/cpan/Scalar-List-Utils/lib/List/Util/XS.pm
+++ b/cpan/Scalar-List-Utils/lib/List/Util/XS.pm
@@ -3,7 +3,7 @@ use strict;
 use warnings;
 use List::Util;
 
-our $VERSION = "1.56";       # FIXUP
+our $VERSION = "1.56_001";       # FIXUP
 $VERSION =~ tr/_//d;         # FIXUP
 
 1;
--- a/cpan/Scalar-List-Utils/lib/Scalar/Util.pm
+++ b/cpan/Scalar-List-Utils/lib/Scalar/Util.pm
@@ -14,10 +14,12 @@ our @ISA       = qw(Exporter);
 our @EXPORT_OK = qw(
   blessed refaddr reftype weaken unweaken isweak
 
+  isbool
+
   dualvar isdual isvstring looks_like_number openhandle readonly set_prototype
   tainted
 );
-our $VERSION    = "1.56";
+our $VERSION    = "1.56_001";
 $VERSION =~ tr/_//d;
 
 require List::Util; # List::Util loads the XS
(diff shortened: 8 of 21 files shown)


tonycoz:
The `SvLEN() == (STRLEN)-1` change looks really risky to me, I can see it breaking XS code that assumes `SvLEN() != 0` means the PV belongs to that SV.

Scalar::Util::isbool() still needs SvGETMAGIC():
```
$ ./perl -Ilib -MScalar::Util=isbool -le 'my $x; package Foo { sub TIESCALAR { bless {}, shift } sub FETCH { $x } } tie my $y, "Foo"; $x = 1; print isbool($y) ? "Y" : "N"; $x = !0; print isbool($y) ? "Y" : "N"; my $z = $y; print isbool($z) ? "Y" : "N"'
N
N
Y
```

leonerd:
@tonycoz 

> The `SvLEN() == (STRLEN)-1` change looks really risky to me, I can see it breaking XS code that assumes `SvLEN() != 0` means the PV belongs to that SV.

Yeah, I did wonder about that. It might come under general "ABI shape" concerns - e.g. there are other LEN-type fields which take special values to have special meanings. Alternatively, it may be that we need to allocate a flag for "PV is static const", which would be a sub-flag behind IsCOW and LEN == 0, a way to distinguish "static const" from shared HEK.

At that point there wouldn't be many spare flag bits left, but several flags would now only have meaning if there is a real PV in the SV. At which point I wonder about adding a "pvflags" field to the PV_HEAD structure itself. All PV-like SVs become a little bit larger but we gain a number of spare flag positions again. A good trade-off?

> Scalar::Util::isbool() still needs SvGETMAGIC():

Yes I know - I didn't want to bother getting that higher-level wrapping nicely fixed while the underlying mechanism is still broken. Though you've now suggested a unit test so I can at least fix that part.

leonerd:
> At that point there wouldn't be many spare flag bits left, but several flags would now only have meaning if there is a real PV in the SV. At which point I wonder about adding a "pvflags" field to the PV_HEAD structure itself. All PV-like SVs become a little bit larger but we gain a number of spare flag positions again. A good trade-off?

Some further thought here: We could add `U16 pvflags; U16 cowrefcount;` fields to it, and move the current hack of CowREFCNT into there. That would neaten up the way CowREFCNT works, and expand its range from 1-255 to 1-65536. In addition we'd get space for a bunch of PV flags. Some obvious candidates, besides these COW thoughts, are moving OOK out there. It only makes sense with a real PV structure.

Annoyingly we can't move UTF8 or leave space for future thoughts about UNKNOWN|BYTES|TEXT, because that distinction also needs to remain for non-pPOK structures that still have PV'ification overload magic, because meh. :/

But it could be a start as at least we'd have more flags bits spare.

leonerd:
@tonycoz 
> The `SvLEN() == (STRLEN)-1` change looks really risky to me, I can see it breaking XS code that assumes `SvLEN() != 0` means the PV belongs to that SV.

PTAL: completely rewritten (and force pushed) to set LEN==0 and set a new flag when PV is static const. This change looks a bit smaller and neater than the previous one, also. Hopefully somewhat more rugged in the face of XS modules. At least, those modules ought to be using the `SvIsCOW_*` macros.

sv.h
```diff
@@ -1051,6 +1051,17 @@ Remove any string offset.
...
+/*
+=for apidoc Am|BOOL|SvISBOOL|SV* sv
+
+Returns true if the SV is one of the special boolean constants (PL_sv_yes or
+PL_sv_no), or is a regular SV whose last assignment stored a copy of one.
+*/
+
+#define SvISBOOL(sv)                            \
+    (SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) && \
+        (SvPVX(sv) == (char *)&PL_Yes || SvPVX(sv) == (char *)&PL_No))
```
tonycoz: Why take a pointer to the array (vs the default conversion to a pointer to the first character) and cast it?
```
(SvPVX(sv) == PL_Yes || SvPVX(sv) == PL_No))
```
would work too.

I thought it might be an attempt to avoid the -Wstring-compare warning from clang, but I don't see the warning if I use the simpler form.

tonycoz:
It looks good to me, except for the pointer issue with SvISBOOL(), but that's not a functional problem, mostly style.

I thought overloading might be a way to produce undesired behaviour, but in most cases that returns a new SV, and boolean state appears to be propagated.

The only case I can think of where overloading might be an issue is if XS does an SvTRUE() or SvPV() and wants to check whether the value was boolean, but if the primary intent is for serialization then it should have been caught earlier with an SvROK() check.

SvGROW() (and other similar PV-ish functions) appears to remove boolean-ness of an SV, even though it doesn't change the value of other non-RV SVs.  I'm not sure this is a problem, since in most cases a change in value will happen immediately after.

sv.h
leonerd: Oh that'll be because I can never remember exactly how those rules work, so I was probably being overly-cautious. I'll switch it to your simpler suggestion.

leonerd:
@tonycoz 
> SvGROW() (and other similar PV-ish functions) appears to remove boolean-ness of an SV, even though it doesn't change the value of other non-RV SVs. I'm not sure this is a problem, since in most cases a change in value will happen immediately after.

I think this is fine. Even such code as: `my $true = (5==5); $true =~ s/1/1/;` will remove "boolean intent" from `$true`, but that's OK. The most recent operation on the variable had a stringy nature. Even if the value itself didn't measurably change, the intention of it has.

I've now force-pushed another round of commits, having also fixed the capitalization of `SvIsBOOL` to match the other predicate-testing macros. Once CI has chewed on that I'll rewrite the PR description and un-draft it.

I might suggest it could be split in two still though; it depends what reviewers think.

leonerd:
I think this is now ready for review

tonycoz:
Looks good to me, I think it's waiting on the RFC now.

leonerd:
Some benchmarking results:

```
$ git log10
38b5dc294e (HEAD -> blead, perl5/blead) perlfunc 'ref' and UNIVERSAL: document isa operator alternative
...

$ perl Porting/bench.pl --write=../blead.time ./perl
```

compared to this branch:

```
$ perl Porting/bench.pl --read=../blead.time ./perl=stablebool
...
       bleadperl stablebool
       --------- ----------
    Ir    100.00      99.42
    Dr    100.00      99.46
    Dw    100.00      99.51
  COND    100.00      99.52
   IND    100.00     100.50

COND_m    100.00     107.23
 IND_m    100.00     100.30

 Ir_m1    100.00      96.95
 Dr_m1    100.00     100.00
 Dw_m1    100.00     100.00

 Ir_mm    100.00     100.00
 Dr_mm    100.00     100.00
 Dw_mm    100.00     100.00
```

I don't have a feel for how to interpret these numbers. I don't even know if "bigger = better" or not... Some advice, anyone?

khwilliamson:
The numbers look to me like we don't have to be concerned with a performance hit.

I'm in favor of merging

leonerd:
Excellent, thanks @khwilliamson

I guess we're not really waiting for anything else at this point. I'll wave a prod at p5p@ and maybe on the RFC, but if there's no objections I think just merge it and declare the RFC done. Then I can release a new Scalar::Util and folks can get on with updating their serializers.

karenetheridge:
Got a perldelta addition to go with that? :)

karenetheridge:
Which of the tests in  t/op/bool.t  would have failed before this change? They seem to be testing very basic operations which would have passed before -- what am I missing?

salva:
If I am understanding this patch correctly, we are going to have true values that stringify as "1" and false values that stringify as "". I find this quite limiting and inconvenient.

I would like to have booleans with custom string values (i.e. boolean dualvars), so that I don't have to write code as

```perl
print "flag is set to ".($foo ? "true" : "false");
```

leonerd:
@karenetheridge 

> Got a perldelta addition to go with that? :)

Ah good call - I shall add one before merge.

> Which of the tests in t/op/bool.t would have failed before this change? They seem to be testing very basic operations which would have passed before -- what am I missing?

None of them, that's the point. This change doesn't alter any existing behaviour, all it does is makes a new piece of information available about SVs. That information is only exposed to perl code via the new `Scalar::Util::isbool()` function. Pureperl serialisers would likely call that; XS ones can peek at `SvIsBOOL()` in C.

leonerd:
@salva 

> If I am understanding this patch correctly, we are going to have true values that stringify as "1" and false values that stringify as "". I find this quite limiting and inconvenient.

That is exactly how it's worked for the past 30 years.

> I would like to have booleans with custom string values (i.e. boolean dualvars), so that I don't have to write code as
> 
> ```perl
> print "flag is set to ".($foo ? "true" : "false");
> ```

Yes, in an ideal world that would be nice. Unfortunately we have ~30years of accumulated code which says that true stringifies as "1" and false stringifies as empty string. E.g. consider all sorts of unit tests using `Test::More::is`:

```perl
is( somefunc(), "", 'somefunc returned false' );
```

I didn't want to break all of those.

---

As an aside on this point: Overall what we're finding more lately is that Perl stringification really isn't a very useful property. As observed in Corinna/Object::Pad land, for example, we didn't want to define a single "TO_STRING" method for objects, because different situations call for different forms. Sometimes you want to generate valid perl code that can be string-eval()'ed again, sometimes you want something human-readable for debugging. Sometimes you want to present things to end-users. Sometimes you want different objects which happen to contain equal values to appear equal, and sometimes you don't (ie. does refaddr matter?). The more you look into it, the more you come to the conclusion that Perl offering _one single_ stringification form for any value is actually not a good design.

ilmari:
@salva,

> If I am understanding this patch correctly, we are going to have true values that stringify as "1" and false values that stringify as "". I find this quite limiting and inconvenient.

That has _always_ been the behaviour of boolean values in Perl, and is unrelated to this change, which is _only_ about being able to reliably retain and identify the booleanness of values when they get copied between variables.

> I would like to have booleans with custom string values (i.e. boolean dualvars), so that I don't have to write code as
> 
> ```perl
> print "flag is set to ".($foo ? "true" : "false");
> ```

~Please write an elevator pitch to perl5-porters per the [RFC procedure](https://github.com/Perl/RFCs#readme).~

leonerd:
> Please write an elevator pitch to perl5-porters per the [RFC procedure](https://github.com/Perl/RFCs#readme).

Already way ahead of you:

https://github.com/leonerd/Perl-RFCs/blob/rfc-sv-bool-type/rfcs/rfcTODO.md

(See the "Future Scope" and "Rejected Ideas" sections)

cpan/Scalar-List-Utils/ListUtil.xs
```diff
@@ -2119,6 +2132,9 @@ BOOT:
 #ifndef SvVOK
     av_push(varav, newSVpv("isvstring",9));
 #endif
+#ifndef SvIsBOOL
```
ilmari, line 2135: The `#if !defined(SvWEAKREF) || !defined(SvVOK)` condition above also needs `|| !defined(SvIsBOOL)` added.

salva:
@leonerd, @ilmari 

It seems I have not explained well. I am not advocating for changing how `PL_sv_yes` and `PL_sv_no` stringify. 

What I am saying is that if you are going to add support for boolean scalars in Perl, I think it would be a nice to have, to allow those scalars to support custom PVs.

I would like to be able to say...

```perl

use My::Boolean qw(True False);

$foo = False;
say $foo; # ==> prints "False\n"
if ($foo) { # never happens...
}
to_json({foo => $foo}); # goes as JS false
```

leonerd:
@salva Yes. Please read the RFC first. I go over all of this there in order to preëmpt questions such as this.

https://github.com/leonerd/Perl-RFCs/blob/rfc-sv-bool-type/rfcs/rfcTODO.md

ilmari (changes requested):

cpan/Scalar-List-Utils/ListUtil.xs
leonerd: Ahyes. Now added in a488435fb4. (I'll squash it before merge)

salva:
> @salva Yes. Please read the RFC first. I go over all of this there in order to preëmpt questions such as this

@leonerd, I have reread the RFC again and I fail to see the issue I have commented above addressed there. Maybe I have not expressed it clearly yet, so let me expose it in a different way:

I think it would useful to have an XS function for creating dualvar boolean values, similar to the `newSVivpv` you proposed not too long ago:

```C
SV *my_fail = newSVboolpv(0, "Fallo");
SV *my_ok = newSVboolpv(1, "Bien");
```
At least, I know I would use it in my some of my modules where I already use dualvars for values that are actually booleans.

leonerd:
> At least, I know I would use it in my some of my modules where I already use dualvars for values that are actually booleans.

They wouldn't pass the `SvIsBOOL()` test at that point. That test specifically looks at the SvPVX to see if it's exactly one of the two static Yes or No arrays that core perl defines.

Moreover, I think it would be very confusing for users for there to be different kinds of "true" and "false" values; values that are all true or false, but stringify in different ways.

---

Also I feel this is a request for an extension feature - best to use the RFC for that. That doesn't relate to problems in this PR itself.

salva:
> They wouldn't pass the `SvIsBOOL()` test at that point. That test specifically looks at the SvPVX to see if it's exactly one of the two static Yes or No arrays that core perl defines.

Yeah, that's exactly my point, you have gone for an implementation that would not allow one to do that.


> Also I feel this is a request for an extension feature - best to use the RFC for that. That doesn't relate to problems in this PR itself.

Well, I suppose that should have been a comment on the original RFC but I missed the mail where you announced it in p5p :-(

leonerd:
> Well, I suppose that should have been a comment on the original RFC but I missed the mail where you announced it in p5p :-(

https://www.nntp.perl.org/group/perl.perl5.porters/2021/08/msg261172.html

Grinnz:
> > They wouldn't pass the `SvIsBOOL()` test at that point. That test specifically looks at the SvPVX to see if it's exactly one of the two static Yes or No arrays that core perl defines.
> 
> Yeah, that's exactly my point, you have gone for an implementation that would not allow one to do that.

FWIW I disagree with this suggestion. Core booleans should only work one way. You can create boolean objects for other purposes quite easily.

cpan/Scalar-List-Utils/ListUtil.xs
leonerd: That's now handled

tonycoz:
> Well, I suppose that should have been a comment on the original RFC but I missed the mail where you announced it in p5p :-(

The RFC is still a PR, you can still comment there, though I don't expect any substantial changes at this point.

That said, your suggestions fall well outside the intent of the RFC - to provide serializers a mechanism to distinguish boolean values from other values.  Such "bool and string" values would make serializers even harder to implement for perl.

If you don't care about serializers, you can get a similar effect to what you're asking for from overloading, with no changes to perl.

leonerd:
Reviewers, PTAL.

I've added some unit tests around threading, and also added @Leont's suggestion of `sv_setbool()`.

I've had to move where the `SvIsCOW_static()` test happens inside `Perl_sv_setsv_flags`, so that it still applies when non-core XS modules call `sv_setsv()`. Notably, this was otherwise causing the `threads::shared` tests to fail, and I imagine would also make other XS modules fail. This appears now to be fixed.

cpan/Scalar-List-Utils/t/boolean-thr.t
```diff
@@ -0,0 +1,38 @@
...
+use Scalar::Util ();
+use Test::More
+    (grep { /isbool/ } @Scalar::Util::EXPORT_FAIL) ? (skip_all => 'isbool requires XS version') :
+    (!$Config::Config{usethreads})                 ? (skip_all => 'perl does not support threads') :
+    (tests => 5);
+
+use threads;
+use threads::shared;
+
+Scalar::Util->import("isbool");
```
ilmari, line 16: Would it not be neater to just do `use Scalar::Util qw(isbool);` _after_ the `use Test::More …;` incantation?

leonerd: No, because of the `grep` on line 9, still needs to have `Scalar::Util` loaded.
Well, I suppose we could do a second `use`.. hmm.. To be honest I just copied the code structure from some of the other .t files.

ilmari: D'oh, I missed that. Nevermind, it's fine like it is then.

cpan/Scalar-List-Utils/t/boolean-thr.t
```diff
@@ -0,0 +1,38 @@
+#!./perl
+
+use strict;
+use warnings;
+
+use Config ();
+use Scalar::Util ();
+use Test::More
+    (grep { /isbool/ } @Scalar::Util::EXPORT_FAIL) ? (skip_all => 'isbool requires XS version') :
```
haarg: `defined &Scalar::Util::isbool ? (skip_all => 'isbool not supported on this perl') :`

ilmari:
A more informative error message should also be added to the [`export_fail`](https://github.com/Dual-Life/Scalar-List-Utils/blob/fa7b8da283590a72a7db3461c64bc704e8c45d3a/lib/Scalar/Util.pm#L38) method.

cpan/Scalar-List-Utils/t/boolean-thr.t
leonerd: Fixed in bd5eca7756

leonerd:
> A more informative error message should also be added to the [`export_fail`](https://github.com/Dual-Life/Scalar-List-Utils/blob/fa7b8da283590a72a7db3461c64bc704e8c45d3a/lib/Scalar/Util.pm#L38) method.

Fixed in cbf2ab509d

duncand:
This is a comment about https://github.com/leonerd/Perl-RFCs/blob/rfc-sv-bool-type/rfcs/rfcTODO.md which was referred to in comments here, and I don't see a way to leave comments there so I'll do so here.

EDIT:  Comment moved to https://github.com/Perl/RFCs/pull/3 where it belongs, but to which no obvious link was found at the above.

haarg:
The RFC can be commented on at Perl/RFCs#3

ilmari (approved):

nwc10 (commented):
I'd be tempted to implement `SvIsBOOL` as a static inline function instead, because it is evaluating its `sv` argument more than once. I'd keep the same name, but make it a function, which I'm aware means that it's then inconsistent with most of the other `Sv[A-Z]+` "things" which are macros, but they date from a time before decent inline functions.

(Oh, and, if it stays as a macro, every `sv` should be `(sv)` to avoid (other) strange bugs due to precedence.)

I don't know if you'd considered this and had reasons to reject it. I'm aware that making it a function is a bit more complex, and I think might end up needing to go into `inline.h` because it will inevitably then depend on something declared later in the headers.

But going for a function first will avoid creating a subtle trap for anyone calling it on an expression with side effects.

(I've looked at all the other commits in the branch. I think that this is the only thing I could spot to comment on)

khwilliamson:
On 9/6/21 12:58 PM, Nicholas Clark wrote:
> 
> I'd be tempted to implement this as a static inline function instead, 
> because it is evaluating its |sv| argument more than once. I'd keep the 
> same name, but make it a function, which I'm aware means that it's then 
> inconsistent with most of the other |Sv[A-Z]+| "things" which are 
> macros, but they date from a time before decent inline functions.
> 
> (Oh, and, if it stays as a macro, every |sv| should be |(sv)| to avoid 
> (other) strange bugs due to precedence.)
> 
> I don't know if you'd considered this and had reasons to reject it. I'm 
> aware that making it a function is a bit more complex, and I think might 
> end up needing to go into |inline.h| because it will inevitably then 
> depend on something declared later in the headers.
> 
> But going for a function first will avoid creating a subtle trap for 
> anyone calling it on an expression with side effects.
> 
> (I've looked at all the other commits in the branch. I think that this 
> is the only thing I could spot to comment on)
> 

++

leonerd:
@nwc10 Ahyes, the macro is probably complex enough it should be an inline function instead.
I'm unsure about naming a function SvIsBOOL() though - what I might do is call the function something like `Perl_sv_isbool()` and macro-wrap it thusly. Though to be honest at that point maybe people expect the macro to be just `sv_isbool()`?

And yes I've been bitten by `SvTRUE(POPs)` before, so I definitely sympathize with the "expression with side-effects" problem.

leonerd:
@nwc10 What do you think of c1fbc57d1b ?
All I've done is moved the code into a static inline function, but I haven't written any docs on it yet. As it's static inline, I don't think it needs any docs, as surely the "user-facing interface" to it is the SvIsBOOL() macro. But perhaps I could put a short comment to say that's the function body implementing the macro.

nwc10:
c1fbc57 was what I was hoping would work.

I'd be tempted to stick with this, and only explore GCC brace groups/PL_Sv if someone actually demonstrates a platform where the build fails as-is.

I'm kind of assuming that if gcc from 15 years ago is good enough for our general needs (and all the *oldest* *nix vendor compilers that we still have access to) that maybe we're being over cautious in assuming the worst about C toolchains.

pp_hot.c
```diff
@@ -1083,7 +1083,7 @@ PP(pp_multiconcat)
                     if (!SvOK(right)) {
                         if (ckWARN(WARN_UNINITIALIZED))
                             report_uninit(right);
-                        sv_setsv_flags(left, &PL_sv_no, 0);
+                        sv_setbool(left, FALSE);
```
wolfsage, line 1086: Here and in gv.h we go from:
```
sv_setsv_flags(left, &PL_sv_no, 0)
```
to:
```
sv_setsv_flags(left, &PL_SV_no, SV_GMAGIC|SV_DO_COW_SVSETSV)
```
since `sv_setbool(left, FALSE)` does `sv_setsv(sv, boolSV(b))`

... does this matter?

leonerd: It does not. `SV_GAMAGIC` tells `sv_setsv_flags()` to invoke GETMAGIC() on the source; but we know that `PL_sv_no` has no such magic, so this extra flag causes no problem.

In addition, `SV_DO_COW_SVSETSV` is the flag that tells `sv_setsv_flags()` to preserve the COW string and is in fact the exact flag we wanted here - the 0 wouldn't preserve it and hence would break the booleanness of it.

PR #19048, opened 2021-08-13
Avoid passing undefined values to Text::Wrap::wrap().

jkeenan:
Following upgrade to CPAN version 2021.0804 of Text-Tabs+Wrap (9a679b43, Aug
05 2021), we began to get "use of uninitialized value $_ in split" warnings
when running lib/perlbug.t.

### DIFF
--- a/utils/perlbug.PL
+++ b/utils/perlbug.PL
@@ -1096,7 +1096,7 @@ sub _read_report {
     local $Text::Wrap::huge = 'overflow';
     while (<REP>) {
         if ($::HaveWrap && /\S/) { # wrap() would remove empty lines
-            $content .= Text::Wrap::wrap(undef, undef, $_);
+            $content .= Text::Wrap::wrap('', '', $_);
         } else {
             $content .= $_;
         }


jkeenan:
The warnings we're getting during `make test_harness` look like this:
```
===(  561315;298  18982/?  381/?  31/?   1/25 )=========================Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 1.
Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 1.
Use of uninitialized value $lead in concatenation (.) or string at ../lib/Text/Wrap.pm line 53, <REP> line 1.
Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 2.
Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 2.
Use of uninitialized value $lead in concatenation (.) or string at ../lib/Text/Wrap.pm line 53, <REP> line 2.
...
Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 102.
Use of uninitialized value $_ in split at ../lib/Text/Tabs.pm line 19, <REP> line 102.
Use of uninitialized value $lead in concatenation (.) or string at ../lib/Text/Wrap.pm line 53, <REP> line 102.
```
ack-ing the repo for `<REP>` shows that we're getting the problem while running `lib/perlbug.t`, which in turn suggests a correction to `utils/perlbug.PL`.

While we might want to consider upstream changes to check for definedness of values passed to `Text::Wrap::wrap()` -- @ap -- this p.r. should solve the immediate problem.

ap:
Please do not apply this PR!

> which in turn suggests a correction to `utils/perlbug.PL`.

No. It absolutely, unequivocally doesn’t. It’s *good* and I’m *glad* that we’re getting these warnings – they are the only signal I’ve gotten about the fact that I broke Text::Wrap (whose own tests are, unfortunately, woefully inadequate).

What it actually suggests is fixing Text::Wrap. Which I’ve already done, and have had a dev release with the fix already out for a while. I’ll be cutting a proper release with the fix soon. Blead should be updated to that as soon as it’s out (and I intend to submit a PR for that at that point myself also, but feel free to beat me to that).

In the meantime, please don’t patch anything outside of the Text-Tabs+Wrap distribution to deal with this. When upstream breaks downstream, the proper response is to fix upstream – not downstream.

If there is urgency for some reason (is a freeze coming up? is it making CI too noisy? something else?), please let me know and I can push it up my priorities (further; I’ll be getting to it soon anyway).

jkeenan:
I'm okay with waiting for an update upstream.  We can simply hold this p.r. open but mark it do not apply.

Thanks for your quick respons.

ap:
Now pushed as #19068.

jkeenan:
> Now pushed as #19068.

Thanks.  This p.r. is now superseded; closing.

PR #19050, opened 2021-08-14
Pod html refactoring 5 of 5

jkeenan:
This if the fifth (and last) in a series of five pull requests submitted with respect to https://github.com/Perl/perl5/issues/18894.

In this p.r., we at long last complete the conversion of the data structures holding information needed to generate HTML files from documents containing POD.  As noted in https://github.com/Perl/perl5/issues/18894, when we started (in the version which shipped with perl-5.34.0), "the code in `ext/Pod-Html/lib/Pod/Html.pm` [was] not fully encapsulated. State [was] held in a total of 19 lexical variables declared outside the scope of the main function, `pod2html()`."  Now that state is held in a `Pod::Html` object and the exportable function `pod2html()` is simply a wrapper around method calls on that object.  The length of `pod2html()` is reduced from 182 lines to 42 lines, which should make it easier to understand what it's doing.

Once this p.r. is merged to blead, we can evaluate patches for all the small points which we're raised in review of this series of pull requests.

### DIFF
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -24,7 +24,6 @@ use Pod::Html::Util qw(
     relativize_url
 );
 use locale; # make \w work right in non-ASCII lands
-use Data::Dumper;
 
 =head1 NAME
 
@@ -239,12 +238,6 @@ This program is distributed under the Artistic License.
 
 =cut
 
-
-my $Podroot;
-
-#my %Pages = ();                 # associative array used to find the location
-                                #   of pages referenced by L<> links.
-
 sub new {
     my $class = shift;
     return bless {}, $class;
@@ -380,12 +373,10 @@ sub refine_globals {
         # Is the above not just "$self->{Htmlfileurl} = $self->{Htmlfile}"?
         $self->{Htmlfileurl} = unixify($self->{Htmlfile});
     }
-    #return $self;
     return { %{$self} };
 }
 
 sub generate_cache {
-    #my ($self, $Pagesref) = @_;
     my $self = shift;
     my $pwd = getcwd();
     chdir($self->{Podroot}) ||
@@ -427,7 +418,6 @@ sub generate_cache {
         print $cache "$key $self->{Pages}->{$key}\n";
     }
     close $cache or die "error closing $self->{Dircache}: $!";
-    #return %{$Pagesref};
 }
 
 sub _transform {
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -273,7 +273,6 @@ sub pod2html {
 
     # set options for the HTML generator
     my $parser = Pod::Simple::XHTML::LocalPodLinks->new();
-    my $output;
     $parser->codes_in_verbatim(0);
     $parser->anchor_items(1); # the old Pod::Html always did
     $parser->backlink($self->{Backlink}); # linkify =head1 directives
@@ -282,14 +281,14 @@ sub pod2html {
     $parser->htmlfileurl($self->{Htmlfileurl});
     $parser->htmlroot($self->{Htmlroot});
     $parser->index($self->{Doindex});
-    $parser->output_string(\$output); # written to file later
+    $parser->output_string(\$self->{output}); # written to file later
     $parser->pages(\%Pages);
     $parser->quiet($self->{Quiet});
     $parser->verbose($self->{Verbose});
 
     $parser = $self->refine_parser($parser);
     $self->feed_tree_to_parser($parser, $podtree);
-    $self->write_file($output);
+    $self->write_file();
 }
 
 sub init_globals {
@@ -631,7 +630,7 @@ sub feed_tree_to_parser {
 }
 
 sub write_file {
-    my ($self, $output) = @_;
+    my $self = shift;
     $self->{Htmlfile} = "-" unless $self->{Htmlfile}; # stdout
     my $fhout;
     if($self->{Htmlfile} and $self->{Htmlfile} ne '-') {
@@ -641,7 +640,7 @@ sub write_file {
         open $fhout, ">-";
     }
     binmode $fhout, ":utf8";
-    print $fhout $output;
+    print $fhout $self->{output};
     close $fhout or die "Failed to close $self->{Htmlfile}: $!";
     chmod 0644, $self->{Htmlfile} unless $self->{Htmlfile} eq '-';
 }
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -2,7 +2,7 @@ package Pod::Html;
 use strict;
 use Exporter 'import';
 
-our $VERSION = 1.32;
+our $VERSION = 1.33;
 $VERSION = eval $VERSION;
 our @EXPORT = qw(pod2html htmlify);
 our @EXPORT_OK = qw(anchorify relativize_url);
--- a/ext/Pod-Html/lib/Pod/Html.pm
+++ b/ext/Pod-Html/lib/Pod/Html.pm
@@ -25,6 +25,7 @@ use Pod::Html::Util qw(
     relativize_url
 );
 use locale; # make \w work right in non-ASCII lands
+use Data::Dumper;
 
 =head1 NAME
 
@@ -245,15 +246,22 @@ my $Podroot;
 my %Pages = ();                 # associative array used to find the location
                                 #   of pages referenced by L<> links.
 
+sub new {
+    my $class = shift;
+    return bless {}, $class;
+}
+
 sub pod2html {
     local(@ARGV) = @_;
     local $_;
 
-    my $globals = init_globals();
-    my $opts = process_command_line;
-    $globals = process_options($globals, $opts);
+    my $self = Pod::Html->new();
+    $self->init_globals();
 
-    $globals = refine_globals($globals);
+    my $opts = process_command_line;
+    $self->process_options($opts);
+
+    my $globals = $self->refine_globals();
 
     # load or generate/cache %Pages
     unless (get_cache($globals)) {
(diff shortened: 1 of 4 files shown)


ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -558,36 +554,36 @@ sub set_Title_from_podtree {
             (@$podtree == 4 ||
                 (ref($podtree->[4]) eq "ARRAY" &&
                 $podtree->[4]->[0] eq "head1"))) {
-            $globals->{Title} = join("", @{$podtree->[3]}[2..$#{$podtree->[3]}]);
+            $self->{Title} = join("", @{$podtree->[3]}[2..$#{$podtree->[3]}]);
         }
     }
 
-    $globals->{Title} //= "";
-    return html_escape($globals->{Title});
```
Grinnz, line 566: This HTML escaping of the Title appears to get lost.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -239,237 +238,234 @@ This program is distributed under the Artistic License.
...
-        #$globals->{Htmlfileurl} = "$globals->{Htmldir}/" . substr( $globals->{Htmlfile}, length( $globals->{Htmldir} ) + 1);
-        # Is the above not just "$globals->{Htmlfileurl} = $globals->{Htmlfile}"?
-        $globals->{Htmlfileurl} = unixify($globals->{Htmlfile});
-
+        #$self->{Htmlfileurl} = "$self->{Htmldir}/" . substr( $self->{Htmlfile}, length( $self->{Htmldir} ) + 1);
+        # Is the above not just "$self->{Htmlfileurl} = $self->{Htmlfile}"?
+        $self->{Htmlfileurl} = unixify($self->{Htmlfile});
     }
-    return $globals;
+    return { %{$self} };
```
Grinnz: This seems unnecessary since the return value isn't used anymore. Could just `return $self` like some of the others

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -239,237 +238,234 @@ This program is distributed under the Artistic License.
...
+    #     limit search to those in @{$self->{Podpath}}
     # - verbose: report (via 'warn') what search is doing
     # - laborious: to allow '.' in dirnames (e.g., /usr/share/perl/5.14.1)
-    # - callback: used to remove Podroot and extension from each file
     # - recurse: go into subdirectories
     # - survey: search for POD files in PodPath
     my ($name2path, $path2name) = 
-        Pod::Simple::Search->new->inc(0)->verbose($globals->{Verbose})->laborious(1)
-        ->callback(\&_save_page)->recurse($globals->{Recurse})->survey(@{$globals->{Podpath}});
-    #print STDERR Data::Dumper::Dumper($name2path, $path2name) if ($globals->{Verbose});
```
Grinnz, line 398: This comment gets removed but seems to be the only reason Data::Dumper is required when the verbose option is passed. Maybe that could be removed as well.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -239,237 +238,234 @@ This program is distributed under the Artistic License.
...
     }
-    return $globals;
+    return $self;
 }
 
 sub refine_globals {
-    my $globals = shift;
-    require Data::Dumper if $globals->{verbose};
+    my $self = shift;
+    require Data::Dumper if $self->{verbose};
```
Grinnz: Data::Dumper required here.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -239,237 +238,234 @@ This program is distributed under the Artistic License.
...
+    my $class = shift;
+    return bless {}, $class;
+}
 
 sub pod2html {
     local(@ARGV) = @_;
     local $_;
 
-    my $globals = init_globals();
+    my $self = Pod::Html->new();
```
Grinnz, line 250: Could be written `__PACKAGE__->new`

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: Accepted; handled in 1c6c07161d40fc4b5a3762c4a71cbfc8e62e881d

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: Accepted; handled in 34468cc07cbee4d4a1de7ff26d3663a2dabd0fa7

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > This HTML escaping of the Title appears to get lost.

Accepted; handled in 74fb96b24dc12c876b3f2b395418843d6d2d3a47

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: The way it's written is more self-documenting, so I think we should leave it this way for now.

Grinnz (approved):

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -384,36 +384,36 @@ sub refine_globals {
...
     # - verbose: report (via 'warn') what search is doing
     # - laborious: to allow '.' in dirnames (e.g., /usr/share/perl/5.14.1)
     # - callback: used to remove Podroot and extension from each file
     # - recurse: go into subdirectories
     # - survey: search for POD files in PodPath
     my ($name2path, $path2name) = 
-        Pod::Simple::Search->new->inc(0)->verbose($globals->{Verbose})->laborious(1)
-        ->callback(\&_save_page)->recurse($globals->{Recurse})->survey(@{$globals->{Podpath}});
-    #print STDERR Data::Dumper::Dumper($name2path, $path2name) if ($globals->{Verbose});
+        Pod::Simple::Search->new->inc(0)->verbose($self->{Verbose})->laborious(1)
```
rjbs, line 394: Thanks to serendipity for letting me see that there's a boolean property called `laborious` somewhere.  I got a chuckle.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -239,237 +238,233 @@ This program is distributed under the Artistic License.
...
-        or die "$0: error open $globals->{Dircache} for writing: $!\n";
+    warn "caching directories for later use\n" if $self->{Verbose};
+    open my $cache, '>', $self->{Dircache}
+        or die "$0: error open $self->{Dircache} for writing: $!\n";
 
-    print $cache join(":", @{$globals->{Podpath}}) . "\n$globals->{Podroot}\n";
-    my $_updirs_only = ($globals->{Podroot} =~ /\.\./) && !($globals->{Podroot} =~ /[^\.\\\/]/);
-    foreach my $key (keys %{$Pagesref}) {
+    print $cache join(":", @{$self->{Podpath}}) . "\n$self->{Podroot}\n";
+    my $_updirs_only = ($self->{Podroot} =~ /\.\./) && !($self->{Podroot} =~ /[^\.\\\/]/);
```
rjbs, line 409:
```suggestion
    my $_updirs_only = $self->{Podroot} =~ /\.\./ && $self->{Podroot} !~ /[^\.\\\/]/;
```

These kinds of lines can be tricky, so maybe you will not agree that this is a little better.  But if you agree, please feel free to accept this suggestion.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -563,12 +562,12 @@ sub set_Title_from_podtree {
             (@$podtree == 4 ||
                 (ref($podtree->[4]) eq "ARRAY" &&
                 $podtree->[4]->[0] eq "head1"))) {
-            $globals->{Title} = join("", @{$podtree->[3]}[2..$#{$podtree->[3]}]);
+            $self->{Title} = join("", @{$podtree->[3]}[2..$#{$podtree->[3]}]);
```
rjbs, line 556: prime candidate to be postfix

```
join("", $podtree->[3]->@[ 2 .. $podtree->[3]->$#* ]);
```

Or, what I'd actually write:
```
my @want_indices = (2 .. $podtree->[3]->$#*);
… = join("", $podtree->[3]->@[ $want_indices ]);
```

I have refrained from a lot of "but postfix deref!" but in this case it seems to be to be a significant improvement — but only in clarify, not correctness.

ext/Pod-Html/lib/Pod/Html.pm
```diff
@@ -649,6 +646,16 @@ sub write_file {
...
+<<<<<<< HEAD
+=======
+#sub compare {
+#    my ($ha, $hb) = @_;
+#    local $Data::Dumper::Terse  = 1;
+#    local $Data::Dumper::Indent = 0;
+#    local $Data::Dumper::Sortkeys = 1;
+#    return (Dumper($ha) eq Dumper($hb)) ? 1 : 0;
+#}
+>>>>>>> b4b3ed877e (Replace _save_pages() with _transform())
```
rjbs: It appears you have committed a conflicted hunk.  **Surely** this is a mistake!

rjbs (approved):
I made some small suggestions.  If it was me, I would amend the history to remove the merge conflict markers from ever appearing.  They don't demonstrate the development of the code itself, and only show that the programmer made a mistake while building this branch.  Best to drop them, in my humble opinion.

My other small comments not withstanding, this looks great.  Seeing all the `$globals` go away was nice, a real payoff.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > It appears you have committed a conflicted hunk. **Surely** this is a mistake!

This is puzzling.  I don't see this in the code I pushed to github.  Github is saying that the lines in question are "Outdated".

I just rebased the branch on blead.  Can you double-check?

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > prime candidate to be postfix
> 
> ```
> join("", $podtree->[3]->@[ 2 .. $podtree->[3]->$#* ]);
> ```
> 
> Or, what I'd actually write:
> 
> ```
> my @want_indices = (2 .. $podtree->[3]->$#*);
> … = join("", $podtree->[3]->@[ $want_indices ]);
> ```
> 
> I have refrained from a lot of "but postfix deref!" but in this case it seems to be to be a significant improvement — but only in clarify, not correctness.

In keeping with the approach I've been taking throughout this project, I'd like to defer introducing postfix deref (where appropriate) until after we've gotten the OO-ified Pod::Html in blead.  That way, if using that proves at all problematic, we will have isolated the problems.  I will add this to the TODO.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > These kinds of lines can be tricky, so maybe you will not agree that this is a little better. But if you agree, please feel free to accept this suggestion.

Again, following the philosophy of "minimum number of changes needed to get us from Pod-Html as it appears in 5.34.0 to an OO-ified module," I recommend deferring this change until after this p.r. is in blead.  Added to TODO list.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: > Thanks to serendipity for letting me see that there's a boolean property called `laborious` somewhere. I got a chuckle.

The Pod::Simple stuff was the most laborious part of this whole project.

ext/Pod-Html/lib/Pod/Html.pm
jkeenan: @rjbs, I ack-ed the branch for merge-conflict marks and found none.  I also built `perl` at each commit in the branch and ran the tests in `ext/Pod-Html/t/*.t`.  I got no failures.  So unless I hear from you otherwise, I'm going to assume that what you observed poses no barrier to merging this ticket into blead.

rjbs: Your branch includes the commit 47a5a69b5a46d5e567358f2e10594fcca0698b9a which commits merge markers.  A later commit removes them.

jkeenan: > Your branch includes the commit [47a5a69](https://github.com/Perl/perl5/commit/47a5a69b5a46d5e567358f2e10594fcca0698b9a) which commits merge markers. A later commit removes them.

Y'know, I took pains in this project to have small(-ish) commits within each pull request -- but I honestly didn't expect anyone to review on a commit-by-commit basis.  This problem should be repaired now; please double-check.  Thanks for the detailed code review.

rjbs: Looking good!

Reviewing on a "whole diff" versus "commit by commit" basis can be a tough choice.  We were talking about it at $dayjob just last week.  I like commit-by-commit for bigger branches that are altering existing code, *generally*, and whole-diff for new code or a couple small commits.  It's never clear, though.  I wrote a few comments while I was reviewing your work's early commits that I had to then go back and delete in later commits when I saw that my concern was answered by for the end of the commit.

Trade-offs all the way down…

Thanks for your work on this!

PR #19051, opened 2021-08-14
isUTF8 DFA: change while {} to do {} while;

khwilliamson:
This saves a conditional in many cases.  Core Perl doesn't call this
on an empty string, so the first test that it is empty is redundant.

We can't guarantee this for non-core calls, so the conditional is made
explicit for them.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -1481,6 +1481,22 @@ Perl_is_utf8_string_loclen(const U8 *s, STRLEN len, const U8 **ep, STRLEN *el)
     }
 }
 
+/* The perl core arranges to never call the DFA below without there being at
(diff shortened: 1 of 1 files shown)


jkeenan:
> This saves a conditional in many cases. Core Perl doesn't call this
> on an empty string, so the first test that it is empty is redundant.
> 
> We can't guarantee this for non-core calls, so the conditional is made
> explicit for them.

Is there any way we can add tests for this behavior that will illustrate the before/after difference?

khwilliamson:
I don't understand your question.  The change is deliberately designed to prevent the observable behavior from changing

leonerd:
LGTM

PR #19061, opened 2021-08-19
Exporter: Document non-inheriting as default mechanism

Grinnz:
The synopsis currently presents adding Exporter to the inheritance tree as the default way to use it. Modern practice is to only import the 'import' method, so we should present this first. Plus some other cleanups to be consistent with this.

### DIFF
--- a/dist/Exporter/lib/Exporter.pm
+++ b/dist/Exporter/lib/Exporter.pm
@@ -6,7 +6,7 @@ no strict 'refs';
 our $Debug = 0;
 our $ExportLevel = 0;
 our $Verbose ||= 0;
-our $VERSION = '5.76';
+our $VERSION = '5.77';
 our %Cache;
 
 sub as_heavy {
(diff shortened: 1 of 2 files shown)


skirmess:
Is there ever a sane reason to inherit from `Exporter`? Unless there is can't we just remove the examples for the wrong usage? (None of the other modules recommends inheriting from them as an alternative to just `use`ing them. Why is `Exporter` special?)

Whoever reads the pod page of `Exporter` (e.g. me) probably can't make an educated guess on which way to use it and giving us the choice of using the tool wrong doesn't help. :-)

Further down we have this:
> The ideal way to never have to think about that is to use
> C<BEGIN> blocks and the simple import method.  So the first part
> of the L</SYNOPSIS> code could be rewritten as:

And now my question is, if that usage is better, why isn't the synopsis rewritten? I can't see and disadvantages of using it in a `BEGIN` block but the synopsis shows the usage without one.

Grinnz:
If you are using a version of Exporter older than 5.57, or want to use one of the other methods such as export_to_level described elsewhere. Most importantly it needs to be documented because of the prevalence of it in the wild.

leonerd:
If nobody objects to this, then I shall merge it tomorrow. It's purely a documentation change with no actual code (other than the necessary `$VERSION` bumps)

leonerd:
LGTM

PR #19062, opened 2021-08-19
av.h: additional comments regarding real/fake AVs

richardleach:
Supplement commentary on "real" vs "fake" AVs.

### DIFF
--- a/av.h
+++ b/av.h
@@ -18,15 +18,25 @@ struct xpvav {
 
 /* SV*	xav_arylen; */
 
-/* SVpav_REAL is set for all AVs whose xav_array contents are refcounted.
- * Some things like "@_" and the scratchpad list do not set this, to
- * indicate that they are cheating (for efficiency) by not refcounting
- * the AV's contents.
- * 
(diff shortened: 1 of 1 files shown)


richardleach:
Wording for documentation in #19038 has been hampered by there not being much documentation on "real" and "fake" AVs. There only seems to be some incomplete commentary in av.h, which this PR initially tries to supplement.

However, should this be apidoc (despite seeming to be internals-only detail) or covered in _perlguts_?

Are there other examples of fake AVs (or surprising real AVs) worth mentioning? (Are all the stacks fake AVs?)

When/why would a fake AV not be capable of becoming a real array?

What else might maintainers need to know about real/fake AVs?

leonerd:
> Are there other examples of fake AVs (or surprising real AVs) worth mentioning? (Are all the stacks fake AVs?)

As mentioned `@_` is fake. The `backrefs` list on an object or stash is also fake - this contains a list of all the RVs which point to back to that object/stash, which must be set to `undef` if that object/stash is destroyed.

richardleach:
Thanks @leonerd, will add that.

Issue #19063, opened 2021-08-20
Data::Dumper output shouldn’t depend on string internals


FGasper:
Module: Data::Dumper

**Description**
Downgraded string:
```
> perl -MData::Dumper -e'print Dumper("\xc3\x85")'
$VAR1 = 'Å';
```
Upgraded string:
```
> perl -MData::Dumper -e'print Dumper("\xc3\x85\x{100}")'
$VAR1 = "\x{c3}\x{85}\x{100}";
```

**Expected behavior**
If the string internals shouldn’t affect public behaviour then core modules (that don’t exist specifically to facilitate interaction with those internals) should output them the same.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 26 subversion 3) configuration:

  Platform:
    osname=darwin
    osvers=18.7.0
    archname=darwin-thread-multi-2level
    uname='darwin mojave.internal.macports.net 18.7.0 darwin kernel version 18.7.0: mon may 3 20:41:19 pdt 2021; root:xnu-4903.278.68~1release_x86_64 x86_64 '
    config_args='-des -Dprefix=/opt/local -Dscriptdir=/opt/local/bin -Dvendorprefix=/opt/local -Dusemultiplicity=y -Dusethreads -Duseshrplib -Dcc=/usr/bin/cc -Dman1ext=1pm -Dman3ext=3pm -Dinstallstyle=lib/perl5 -Dman1dir=/opt/local/share/man/man1p -Dman3dir=/opt/local/share/man/man3p -Dsitebin=/opt/local/libexec/perl5.26/sitebin -Dsiteman1dir=/opt/local/share/perl5.26/siteman/man1 -Dsiteman3dir=/opt/local/share/perl5.26/siteman/man3 -Dvendorbin=/opt/local/libexec/perl5.26 -Dvendorman1dir=/opt/local/share/perl5.26/man/man1 -Dvendorman3dir=/opt/local/share/perl5.26/man/man3 -Dpager=/usr/bin/less -sR -Dperlpath=/opt/local/bin/perl5.26 -Dstartperl=#!/opt/local/bin/perl5.26 -Acppflags=-I/opt/local/include  -Accflags=-pipe -Os   -Alddlflags=-L/opt/local/lib -Wl,-headerpad_max_install_names   -Aldflags=-L/opt/local/lib -Wl,-headerpad_max_install_names  '
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
[...]

PR #19064, opened 2021-08-20
Provide richer explanation of the `unicode_strings` difference.

FGasper:
The output here kinda exemplifies bug #19063.

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -198,6 +198,51 @@ was extended further in Perl 5.26 to cover L<the range
 operator|perlop/Range Operators>; and was extended again in Perl 5.28 to
 cover L<special-cased whitespace splitting|perlfunc/split>.
 
+To understand this feature better consider the following:
+
+    > perl -MData::Dumper -e'my $str = "\xc3\x85a\tb"; print Dumper(split m[\s], $str)'
+    $VAR1 = 'Åa';
+    $VAR2 = 'b';
+
+C<split> recognizes C<\t> as whitespace and so returns two strings.
+But consider another example:
+
+    > perl -MData::Dumper -e'my $str = "\xc3\x85a\tb\x{100}"; chop $str; print Dumper(split m[\s], $str)'
+    $VAR1 = "\x{c3}";
+    $VAR2 = 'a';
+    $VAR3 = "b";
+
+Here C<split> returned three strings even though $str is the same.
+What gives?
+
+Here C<$str> starts off with an extra character, C<\x{100}>. This exceeds
+255, which requires Perl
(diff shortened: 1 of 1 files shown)


Grinnz:
Heavy details are rather out of place for this document. Perhaps the section https://perldoc.perl.org/perlunicode#The-%22Unicode-Bug%22 could be enhanced with this information and the feature could be updated if it doesn't sufficiently point there?

FGasper:
> Heavy details are rather out of place for this document. Perhaps the section https://perldoc.perl.org/perlunicode#The-%22Unicode-Bug%22 could be enhanced with this information and the feature could be updated if it doesn't sufficiently point there?

Ah, there’s more in that linked section than I’d realized.

I think the verbiage I was looking for—but didn’t find a bit earlier—was (something like) “this fixes a bug in Perl”.

I also think:
> … unless it is clear to Perl that Unicode is desired

… leaves something to be desired in terms of accuracy, since appending wide characters isn’t the only thing that upgrades a string.

khwilliamson:
On 8/19/21 8:52 PM, Felipe Gasper wrote:
>     Heavy details are rather out of place for this document. Perhaps the
>     section https://perldoc.perl.org/perlunicode#The-%22Unicode-Bug%22
>     <https://perldoc.perl.org/perlunicode#The-%22Unicode-Bug%22> could
>     be enhanced with this information and the feature could be updated
>     if it doesn't sufficiently point there?
> 
> Ah, there’s more in that linked section than I’d realized.
> 
> I think the verbiage I was looking for—but didn’t find a bit earlier—was 
> “this fixes a bug in Perl”.
> 
> I also think:
> 
>     … unless it is clear to Perl that Unicode is desired
> 
> … leaves something to be desired in terms of accuracy, since appending 
> wide characters isn’t the only thing that upgrades a string.
> 

There's that word again 'upgrade' which I think clouds what is going on.

But "unless it is clear to Perl" doesn't seem wrong to me.  It doesn't 
say that a wide character is the only or even main way to make it clear 
to Perl.  There is a complete list in one of the pods.  Maybe just 
providing a link would be sufficient.

khwilliamson:
@FGasper  @Grinnz 
where are we with this request

jkeenan:
> @FGasper @Grinnz where are we with this request

There has been no further discussion of this p.r. in the past 6 months, so I propose that we close this ticket.

jkeenan:
> > @FGasper @Grinnz where are we with this request
> 
> There has been no further discussion of this p.r. in the past 6 months, so I propose that we close this ticket.

No further discussion in 3-1/2 months; closing.

PR #19074, opened 2021-08-22
Add an `sv_setrv()` family of functions

leonerd:
These appear to have been missing all this time.

See also https://www.nntp.perl.org/group/perl.perl5.porters/2021/07/msg260924.html

Notes to reviewers: Currently this requires `ref != NULL`, though at a small cost to performance I could be convinced to accept NULL as well, allowing turning off references by doing `sv_setrv(sv, NULL);`. Maybe worth it?


tonycoz:
> Currently this requires `ref != NULL`, though at a small cost to performance I could be convinced to accept NULL as well, allowing turning off references by doing `sv_setrv(sv, NULL);`

I think it should assert on NULL rather than adding another way to sv_setsv_undef().

PR #19078, opened 2021-08-23
pp_tie should completely reset the underlying hash's iterator state.

nwc10:
Previously pp_tie would mangle the hash iterator state - it would reset EITER but not RITER, meaning that if the iterator happened to be partway down a chain of entries, the rest of that chain would be skipped, but if the iterator's next position was the start of a (new) chain, nothing would be skipped.

We don't have space to store the complete older iterator state (and really nothing should be relying on it), so it seems better to correctly reset the iterator (every time), than leave it in a mess just occasionally.

### DIFF
--- a/t/op/tiehash.t
+++ b/t/op/tiehash.t
@@ -72,4 +72,82 @@ package TestIterators {
     is(each %h, undef, "third iterator is undef");
 }
 
+{
+    require Tie::Hash;
+
+    my %h = (
+        lolcat => "OH HAI!",
+        lolrus => "I HAS A BUCKET",
+    );
+
+    my @want = sort keys %h;
+
+    my @have;
+    while (1) {
+        my $k = each %h;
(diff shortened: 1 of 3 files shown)


t/op/tiehash.t
```diff
@@ -0,0 +1,153 @@
...
+    is(each %h, 'beer', "first iterator");
+    is(each %h, 'perl', "second iterator");
+    is(each %h, undef, "third iterator is undef");
+}
+
+{
+    require Tie::Hash;
+
+    my %h = (
+        lolcal => "OH HAI!",
```
richardleach: Potential typo, `lolcal` -> `lolcat`
😆

nwc10: Oh my! Critical bug!

Thanks for spotting that. Fix pushed (and rebased again to blead)

PR #19087, opened 2021-08-27
Reword discussion of /d regexp modifier.

FGasper:
The phrasing as it stood confused UTF8-flagged strings with
“UTF-8 encoded”. The latter term should refer to strings that the
Perl application has actually encode()d, which probably *won’t*
be UTF8-flagged and thus won’t, per /d modifier rules, get the Unicode
treatment.

This also reorders and rewords some parts in an attempt to clarify that
new code should avoid this flag, including use of the 'unicode_strings'
feature to avoid implicit use.

### DIFF
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -678,18 +678,29 @@ X</u>
 
 =head4 /d
 
-This modifier means to use the "Default" native rules of the platform
+B<IMPORTANT:> Because of the unpredictable behaviors this
+modifier causes, only use it to maintain weird backward compatibilities.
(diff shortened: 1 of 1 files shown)


Grinnz:
> This also removes an incorrect statement about only ASCII characters matching in the absence of (the UTF8 flag).

I believe this was clumsily attempting to reference that *character classes* like `\w` and `\d` only match ASCII characters in this instance. That said, the `/a` section sufficiently explains this.

Grinnz (approved):

PR #19088, opened 2021-08-28
Change FIXME to XXX in source code

khwilliamson:
As per a discussion on #irc, XXX stands visually out more than FIXME,
and is more likely to prevent a commit from being made whose intent was
to fix the item beforehand.

It also is more visually arresting when just reading code, and hence
stands a better chance of being addressed later.

There are also TODO lines in the source.  This commit doesn't address
those, in part to see what the response is to this one.

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -1046,7 +1046,7 @@ foreach my $phase (qw(early late)) {
 unless (exists $defines{cc}) {
     # If it fails, the heuristic of 63f9ec3008baf7d6 is noisy, and hence
     # confusing.
-    # FIXME - really it should be replaced with a proper test of
+    # XXX - really it should be replaced with a proper test of
     # "can we build something?" and a helpful diagnostic if we can't.
     # For now, simply move it here.
     $defines{cc} = (`ccache -V`, $?) ? 'cc' : 'ccache cc';
--- a/Porting/how_to_write_a_perldelta.pod
+++ b/Porting/how_to_write_a_perldelta.pod
@@ -196,7 +196,7 @@ for the particular file in core.)
 
 Changes which create B<new> files in F<pod/> go here.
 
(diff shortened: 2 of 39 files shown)


rwp0:
I think `FIXME` and `TODO` are somewhat more standard than `XXX` et al.

Some IDE's I know recognize those well at least.

hvds:
While it might make sense to recommend using XXX "to prevent a commit from being made whose intent was to fix the item beforehand" - and therefore, for example, to change the ones in `Porting/security_template.pod` - it makes no sense to me to change those that have actually been committed.

FIXME has a semantic sense which XXX lacks (and is distinct from TODO, or CHECKME), so I feel this proposed change discards information and reduces quality. And I do not understand why you believe that seeing XXX in committed code is at all more likely to lead to action than FIXME: that seems very unlikely to me.

If any FIXMEs actually have been committed in error, more useful would be to actually fix them, or at least make tickets for them.

jkeenan:
> As per a discussion on #irc, XXX stands visually out more than FIXME,
> and is more likely to prevent a commit from being made whose intent was
> to fix the item beforehand.
> 
> It also is more visually arresting when just reading code, and hence
> stands a better chance of being addressed later.
> 
> There are also TODO lines in the source. This commit doesn't address
> those, in part to see what the response is to this one.

I oppose this change.  Static code analyzers are programmed to report `FIXME`.  Example:  [LGTM](https://lgtm.com/projects/g/Perl/perl5/?mode=tree&ruleFocus=2151900539).

Leont:
I think it would be helpful to define what these words mean, and when to use what. `XXX` in particular appears to mean different things to different people.

oodler577:
Should probably be something less porn-y. Take your pick, there are ~~25~~ 24 other letters in the ASCII alphabet. (update struck out "k". that'd be bad, too)

PR #19091, opened 2021-08-31
remaining_count is marked NULLOK in embed.fnc, so always check it

tonycoz:
Coverity detected the inconsistent NULL checks.

CID 338543

### DIFF
--- a/utf8.c
+++ b/utf8.c
@@ -3237,7 +3237,9 @@ S_to_case_cp_list(pTHX_ const UV original,
              * should be given. */
         }
 
-        *remaining_count = 0;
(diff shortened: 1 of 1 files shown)


jkeenan:
+1.  I'm glad I'm not the only one reading these Coverity reports!  Passes on FreeBSD.

khwilliamson:
I misread this as an issue, not a pull request, so I created and pushed a different fix before I looked at this commit.

But it turns out that code earlier in the function had already done this set, and so the set would be better removed rather than guarded, so I'm keeping my fix

Issue #19097, opened 2021-09-04
FITS_IN_8_BITS can no longer take a parameter that is incremented


craigberry:
blead currently fails to build on VMS due to several compiler warnings that get escalated to link-time errors.  The warnings look like this:

```
CC/DECC/NOANSI_ALIAS /Include=[]/Standard=Relaxed_ANSI/Prefix=All/Obj=.obj/List/Debug/NoOpt/Show=(expansion,include)/machine/float=ieee/ieee=denor
m/NAMES=(SHORTENED)/Define=(PERL_CORE,_USE_STD_STAT=1) VMS.C

  if (toLOWER_L1(*name++) != 'n') return 0;
......^
%CC-W-UNDEFVARMOD, In this statement, the expression "(!((sizeof(*name++)==1)||((U64)((*name++)|0))==((U8)(*name++))))" modifies the variable "name" more than once without an intervening sequence point.  This behavior is undefined.
at line number 11841 in file D0:[craig.blead]vms.c;1

  if (toLOWER_L1(*name++) != 'l') return 0;
......^
%CC-W-UNDEFVARMOD, In this statement, the expression "(!((sizeof(*name++)==1)||((U64)((*name++)|0))==((U8)(*name++))))" modifies the variable "name" more than once without an intervening sequence point.  This behavior is undefined.
at line number 11842 in file D0:[craig.blead]vms.c;1
```

<snip>

```
CC/DECC /Include=[]/Standard=Relaxed_ANSI/Prefix=All/Obj=.obj /NOANSI_ALIAS/float=ieee/ieee=denorm/NAMES=(SHORTENED)/Define=(_USE_STD_STAT=1,"VERS
ION=""3.16""","XS_VERSION=""3.16""")/Include=([--])/List/Debug/NoOpt  BASE64.c /OBJECT=BASE64.obj

		unsigned char uc = index_64[NATIVE_TO_ASCII(*str++)];
............................................^
%CC-W-UNDEFVARMOD, In the initializer for uc, the expression "(((sizeof(*str++)==1)||((U64)((*str++)|0))==((U8)(*str++))))" modifies the variable "str" more than once without an intervening sequence point.  This behavior is undefined.
at line number 360 in file D0:[CRAIG.blead.cpan.MIME-Base64]BASE64.c;1

	    unsigned char uc = index_64[NATIVE_TO_ASCII(*str++)];
........................................^
[...]

PR #19102, opened 2021-09-06
Evaluate arg to FITS_IN_8_BITS just once

khwilliamson:
This had been changed by 9555181b32f9b30122a8ea4e779c2c9916cec9f8 to
evaluating multiple times.

(sizeof() also is called with the parameter, but that doesn't actually
evaluate the parameter.)

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -1449,8 +1449,8 @@ or casts
 #ifndef __COVERITY__
   /* The '| 0' part ensures a compiler error if c is not integer (like e.g., a
    * pointer) */
-#  define FITS_IN_8_BITS(c) (   (sizeof(c) == 1)                        \
-                 || ((WIDEST_UTYPE) ASSERT_NOT_PTR(c)) == ((U8)(c)))
+#  define FITS_IN_8_BITS(c) (   (sizeof(c) == 1)                            \
+                             || ((WIDEST_UTYPE) ASSERT_NOT_PTR(c)) <= 0xFF)
 #else
 #  define FITS_IN_8_BITS(c) (1)
 #endif


khwilliamson:
This change is so obvious, I am wondering what I am missing now that I knew then.  Or was I just not thinking it through then

khwilliamson:
This should fix #19097

nwc10:
This isn't going to work...

For gcc10 this is generating **thousands** of warnings:

```
handy.h:1453:68: warning: comparison is always true due to limited range of data type [-Wtype-limits]
 1453 |                              || ((WIDEST_UTYPE) ASSERT_NOT_PTR(c)) <= 0xFF)
      |                                                                    ^~
utf8.h:214:18: note: in expansion of macro ‘__ASSERT_’
  214 |                 (__ASSERT_(FITS_IN_8_BITS(ch)) ((U8) ASSERT_NOT_PTR(ch)))
      |                  ^~~~~~~~~
utf8.h:214:28: note: in expansion of macro ‘FITS_IN_8_BITS’
  214 |                 (__ASSERT_(FITS_IN_8_BITS(ch)) ((U8) ASSERT_NOT_PTR(ch)))
      |                            ^~~~~~~~~~~~~~
utf8.h:319:16: note: in expansion of macro ‘NATIVE_UTF8_TO_I8’
  319 |             (((NATIVE_UTF8_TO_I8(c) & UTF_IS_CONTINUATION_MASK)             \
      |                ^~~~~~~~~~~~~~~~~
inline.h:2061:20: note: in expansion of macro ‘UTF8_IS_CONTINUATION’
 2061 |             while (UTF8_IS_CONTINUATION(*s))
      |                    ^~~~~~~~~~~~~~~~~~~~
```

because the compiler isn't smart enough to see the short circuit.

(So many warnings that the build blocks on my terminal trying to keep up, and  it rapidly exceeds my 10,000 line scrollback buffer)

This is frustrating - I've played a bit but I can't yet see a hack to get gcc to shut up.

nwc10:
@khwilliamson, could you check #19104 instead, and if happy merge it. That doesn't generate warnings for me.

(Tested on debian's gcc 10 and clang 11)

PR #19105, opened 2021-09-08
Remove duplicate update of iter in Perl_hv_iternext_flags().

nwc10:
Because`iter` points to a structure immediately after of HvARRAY(), it needs
updating if HvARRAY() has moved because it has been expanded.

Code was added for VMS in Aug 2005 to address a specific instance of this
bug with commit 03026e68943709ca:
    [patch@25334] hv.c vms environment fix.
    From: "John E. Malmberg" <wb8tyw@qsl.net>
    Message-ID: <4310F552.8050401@qsl.net>

see https://www.nntp.perl.org/group/perl.perl5.porters/2005/08/msg104261.html

However, this code was duplicated by and hence made redundant with a more
general fix added in March 2014 by commit 339441efdfcbaef5:
    make core safe against HvAUX() realloc

    Since the HvAUX structure is just tacked onto the end of the HvARRAY()
    struct, code like this can do bad things:

        aux = HvAUX();
        ... something that might split hv ...
        aux->foo = ...; /* SEGV! */

    So I've visually audited core for places where HbAUX() is saved and then
    re-used, and re-initialised the var if it looks like HvARRAY() could
    have changed in the meantime.

    I've been very conservative about what might be unsafe. For example,
    destructors or __WARN__ handlers could call perl code that modifies the
    hash.

Hence remove the redundant code.

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -2737,14 +2737,6 @@ Perl_hv_iternext_flags(pTHX_ HV *hv, I32 flags)
     if (!entry && SvRMAGICAL((const SV *)hv)
         && mg_find((const SV *)hv, PERL_MAGIC_env)) {
         prime_env_iter();
-#ifdef VMS
-        /* The prime_env_iter() on VMS just loaded up new hash values
-         * so the iteration count needs to be reset back to the beginning
-         */
-        hv_iterinit(hv);
-        iter = HvAUX(hv);
-        oldentry = entry = iter->xhv_eiter; /* HvEITER(hv) */
-#endif
     }
 #endif
 


nwc10:
If this commit alone (without the rest of the ABH-POC branch) doesn't regress John's debugger example in the linked message, it would be good to get into blead (independent of everything else) as it makes life simpler.

There are 80 commits on that branch, and it's quite unwieldy.

craigberry (commented):
Unfortunately this does not seem to do the trick.  The original problem reported by John M. in 2005 is back, specifically getting nothing when examining %ENV in the Perl debugger the first time, but getting the full contents on the second invocation.

craigberry:
Actually, I mischaracterized the problem.  In 2005 the problem John reported was that `x keys %ENV` in the Perl debugger got a duplicate first entry on the first invocation, but was correct on the second invocation.  The problem now is that %ENV appears empty on the first invocation but is correct on the second invocation.  This is probably not tickled by anything in the test suite, so I need to go check this out with blead and verify this patch is really the source of the problem.  Back soonish.

craigberry:
blead as of v5.35.3-106-ge5a468d336 is ok, so it appears that this goofy "duplicated" code is actually still necessary, at least without something else that's in the ABH-POC code.  Figuring out what the something else is will take more head scratching than I'm going to get to very soon.

nwc10:
Thanks for checking.

Interesting that it's still needed, and a surprise. The new failure behaviour *that %ENV appears empty on the first invocation but is correct on the second invocation* is useful to know.

However, I think I can see how to get a lot more of the structural refactoring improvements into blead **ahead** of "change the hash" (by decoupling them from the commit that changes `HE *` allocation to not use arenas), which **probably** means that the "something else" will be part of that. (It moves all the structures that track iterators around)

If you haven't deleted it and it's not taking up much space, could you keep the build tree for this branch around, as I'll try to write a test case that tickles your observed behaviour. Well, unless you beat me to it :-)

PR #19108, opened 2021-09-08
More tests for assigning to keys %hash

nwc10:
Change some calls from is()/isnt() to cmp_ok() to confirm that values
change in the expected direction, rather than merely being "different".

Re-order some calls to is()/cmp_ok() so that the "expected" value is second.

Don't assume that a hash can always hold `num_buckets()/2` keys.
(The test had previously assumed that assigning this value to keys wouldn't
change the hash size. Instead use the current number of keys, as clearly
the hash is already holding that many keys.)

### DIFF
--- a/t/op/each.t
+++ b/t/op/each.t
@@ -10,8 +10,6 @@ BEGIN {
     set_up_inc('../lib');
 }
 
-plan tests => 55;
-
 my %h;
 $h{'abc'} = 'ABC';
 $h{'def'} = 'DEF';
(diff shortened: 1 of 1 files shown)


jkeenan:
> Change some calls from is()/isnt() to cmp_ok() to confirm that values
> change in the expected direction, rather than merely being "different".
> 
> Re-order some calls to is()/cmp_ok() so that the "expected" value is second.
> 
> Don't assume that a hash can always hold `num_buckets()/2` keys.
> (The test had previously assumed that assigning this value to keys wouldn't
> change the hash size. Instead use the current number of keys, as clearly
> the hash is already holding that many keys.)

I've rarely used Hash::Util, so I'm looking at both `t/op/each.t` and your p.r. with fresh (or, perhaps, naive) eyes.  So I have a question.

`Hash::Util::num_buckets()` is documented as:

> This function returns the total number of buckets the hash holds, or would hold if the array were created. (When a hash is freshly created the array may not be allocated even though this value will be non-zero.)

In the test file as it currently stands, `Hash::Util::num_buckets()` is used without being proceeded by any of the Hash::Util functions that presumably restrict a hash.
```
 68 SKIP: {
 69     skip "no Hash::Util on miniperl", 4, if is_miniperl;
 70     require Hash::Util;
 71     sub Hash::Util::num_buckets (\%);
 72 
 73     my $size = Hash::Util::num_buckets(%h);
```
Does that mean it is okay to use it on "unrestricted" hashes?

Thank you very much.
Jim Keenan

nwc10:
> Does that mean it is okay to use it on "unrestricted" hashes?

"restricted hashes" really are just a hack implemented on top of the regular ("unrestricted") hashes, so most of the functions would do *something* on any hash (whether it makes sense is something else).

But it's a sensible question - its name suggests that it's "an XS module about hashes" (not specific to restricted hashes) so it also became home to some other functions.

I think that the "internal metadata reading" functions make sense on any hash:

* hash_seed
* hash_value
* bucket_stats
* bucket_info
* bucket_array

(but really aren't useful unless you're hacking on `hv.c`)

The only one that might be useful in general purpose code is `hv_store`.

I've not double checked this, but I think that the rest only really make sense for restricted hashes - they either write internal state that only really makes sense for restricted hashes, or they read things (such as the list of permitted keys) that for regular hashes are identical to what built-ins such as `keys` return.

PR #19110, opened 2021-09-09
C11 thread local in OpenBSD shared objects is buggy, so disable it.

nwc10:


nwc10:
@jkeenan - if this fixes it for you, please merge it to blead.

jkeenan:
Tested on threaded builds on openbsd-6.6 and -6.9.  All tests passed.  Rebased on blead, then merged to blead and pushed to origin in 1e0e4aa9919de12bf24246304510aef0e865bd0e.

PR #19113, opened 2021-09-11
Fix a leak when copying a STATIC COW SV to a stringified SV

nwc10:
This is minor a a regression introduced by commit 914bb57489325d34:
    Define a third kind of COW state; STATIC

    ...

    sv_setsv_flags() and sv_setsv_cow() will preserve this state

There was a small omission in the copy code - it didn't handle the case where
the destination SV owned a string buffer already. This is actually
relatively rare - triggering it requires a scalar in code path that is
assigned both strings and booleans. Minimal test case is:

    my $got = 1;
    $got .= "";
    $got = ref "";

cut down from &is in t/comp/parser.t

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -4691,6 +4691,9 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
          * SV_COW_SHARED_HASH_KEYS being set or else we'll break SvIsBOOL()
          */
         else if (SvIsCOW_static(ssv)) {
+            if (SvPVX_const(dsv)) {     /* we know that dtype >= SVt_PV */
+                SvPV_free(dsv);
+            }
             SvPV_set(dsv, SvPVX(ssv));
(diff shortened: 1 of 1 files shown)


nwc10:
It's pure chance that I found this today. I'd figured some weeks ago that we were pretty close to "clean" with `-DPURIFY` and ASAN, so was taking a look at the cause of one of the leaks reported in `t/op/qr.t`

Only to find that when I rebased some changes onto blead, the entire build "blew up" with ASAN errors well before running the tests...

As to the (other) `-DPURIFY` failures - I *almost* have a PR ready for these, so no-need for anyone else to go investigating. I've just run out of time for today (lunch, then `$other`)

leonerd:
Ahyes, good catch. LGTM.
I'll merge when CI reports green.

leonerd (approved):

nwc10:
See #19115 for the (other) `-DPURIFY` fixes.

PR #19115, opened 2021-09-11
Plug PVLV REGEXP body leak

nwc10:
This was leaking a body:

```
#!./perl -w

use strict;

my %h;

sub bar {
    my $r = qr/abc/;
    $_[0] = $$r;
};

bar($h{foo});
```

The leak has actually been present since 8d919b0a35f2b57a but was only exposed by tests added in 2017.

With this leak fixed, and the obscure leak in `hv.c` fixed, the testsuite is clean under `-DPURIFY`


nwc10:
@iabyn it's not your leak, but you've looked at that regexp code more recently than me, so I hope you can spot if I did something stupid.

nwc10:
For reference/to avoid duplication - there's been discussion about two commits on the branch that this PR references over on the PR #19140

nwc10:
These commits were subsumed into #19140

PR #19116, opened 2021-09-12
 Simplify the code related to prime_env_iter()

nwc10:


craigberry:
Thanks for working on this. Adding a comment now mostly so that it will hopefully stay in my GitHub feed long enough for me to remember to test it, which may be a day or a week given other things going on.

craigberry:
Smoke report on VMS with no new test failures is [here](https://www.nntp.perl.org/group/perl.daily-build.reports/2021/09/msg272663.html).  Manual test of `x keys %ENV` in the Perl debugger looks good.

nwc10:
Thanks for the thorough testing. This looks good on VMS and we could put it into blead?

(The GitHub CI suggests that I didn't break Win32)

craigberry:
cygwin was happy on a second run of the CI tests so I think this is ready to merge.

PR #19118, opened 2021-09-12
On VMS, %ENV in scalar context must call prime_env_iter()

nwc10:
Otherwise it will return wrong results, unless other code happens to
have iterated over %ENV. This bug has probably existed forever.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -371,9 +371,18 @@ L</Modules and Pragmata> section.
 
 =over 4
 
-=item XXX-some-platform
+=item C<keys %ENV> on VMS returns consistent results
 
-XXX
+On VMS entries in the C<%ENV> hash are loaded from the OS environment on
+first access, hence the first iteration of C<%ENV> requires the entire
+environment to be scanned to find all possible keys. This initialisation had
+always been done correctly for full iteration, but previously was not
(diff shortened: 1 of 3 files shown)


craigberry:
I will queue up a smoke run of this soonish.  However, note that the GitHub CI reports one failure for "Sanity: Linux -Dusethreads":

```
# Failed test 61 - %ENV lookup, and keys %ENV in scalar context remain consistent at ./test.pl line 1089
#      got '101,101,101'
# expected /(?^:^([1-9][0-9])*,\1,\1$)/
# PROG: 
# $a = $ENV{PATH}; $a = $ENV{q=DCL$PATH=}; $a = keys %ENV; $b = () = keys %ENV; $c = keys %ENV; print qq=$a,$b,$c=
# STATUS: 0
t/op/each ........................................................ FAILED at test 61
```

It looks like there is a slight regex glitch when there are more than two digits in the number of keys.  With 101 keys, we only get the first two digits:

```
$ perl -e '$x=q{101}; $x =~ m/(?^:^([1-9][0-9])*)/; print qq/$1\n/;'
10
```

nwc10:
Ah pants. This is my deserved reward for copy-paste sins. I fixed one, but not the second. Try 0bd57b43ece9505db1f3de7cb276923418da7ee0.

craigberry:
The new test now passes on both VMS and Win32.  Thanks!

PR #19122, opened 2021-09-13
Only expose Internals::getcwd() in miniperl

nwc10:
It's only used during bootstrapping for miniperl when the XS version of Cwd is not yet available, and only needed on platforms that don't already provide their own builtin for getcwd().

It was added in v5.30.0 with the clear warning that

    [it] may be removed or changed without notice

and is not used by any code on CPAN. (Cwd references it, but won't use it once installed as it will have already found an XS implementation (platform specific or generic).

See http://nntp.perl.org/group/perl.perl5.porters/261527

Probably the PSC should approve this PR before it is merged.

### DIFF
--- a/Makefile.SH
+++ b/Makefile.SH
@@ -497,6 +497,15 @@ sh = $SH
 shextract = $shextract
 !GROK!THIS!
 
+# Source files where we build a variant for miniperl:
+mini_special='op perl'
+for file in $mini_special; do
+    mini_special_c="$mini_special_c ${file}mini.c"
+    mini_only_objs="$mini_only_objs ${file}mini\$(OBJ_EXT)"
+    main_only_objs="$main_only_objs ${file}\$(OBJ_EXT)"
+    ctags_exclude="$ctags_exclude --exclude=${file}mini.c"
+done
+
 ## In the following dollars and backticks do not need the extra backslash.
 $spitshell >>$Makefile <<'!NO!SUBS!'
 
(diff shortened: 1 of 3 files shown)


nwc10:
Forgot to note, I think that changes aren't needed in any of the platforms that use their own Makefile template (ie don't use `Makefile.SH`). I think all of these are covered by `%METHOD_MAP` in Cwd.pm, and end up using their own XS implementation, so never reach `Internals::getcwd`.

richardleach:
I'm not sure if this would break Android/Termux builds again. 

(Unfortunately I had to scrap the hardware running my Android smoker and haven't had time to build a Pi-based replacement, so won't be able to test it unless a surprising chunk of free time comes along. )

nwc10:
> I'm not sure if this would break Android/Termux builds again.

Were they using `Cross/Makefile-cross-SH`?

That notes

`# miniperl removed from here, as this is built on HOST`

and that statement has to be accurate, as `Makefile-cross-SH` doesn't contain the logic for `perlmini.c` meaning that it isn't capable of building a  working `miniperl`, as a working `miniperl` requires C level support for `buildcustomize.pl`

Also, unless code is *directly* using `Internals::getcwd` then it won't break, because even the regular `Cwd` support loading on `miniperl` and having a working `getcwd` - as I understand the only thing the "Internals" version supported additionally over the pure Perl fallback was supporting corner cases such as some parent directories not being readable by the current user.

richardleach:
> > I'm not sure if this would break Android/Termux builds again.
> 
> Were they using `Cross/Makefile-cross-SH`?

No, native builds using the Termux toolchain. Configure is just:
` sh Configure -des -Alibpth="/system/lib /vendor/lib"`

> the only thing the "Internals" version supported additionally over the pure Perl fallback was supporting corner cases such as some parent directories not being readable by the current user.

This is exactly the problem there.

nwc10:
> No, native builds

Native builds are still going to use `miniperl` (almost) all the way to make test, and `miniperl` keeps `Internals::getcwd`.

And "real" `perl` will get the XS implementation of Cwd, which is (effectively) the same call as `Internals::getcwd` (just in a proper namespace.

Unless you had local patches that aren't in blead, then I can't see how this is going to change anything, as the native build for anything with `Configure` will use the same code (and approach) on all platforms that it runs on.

richardleach:
> And "real" `perl` will get the XS implementation of Cwd, which is (effectively) the same call as `Internals::getcwd` (just in a proper namespace.

Ok, sounds like it should be fine then. Thanks.

tonycoz (approved):
This looks good to me.

Nicholas suggested this should be approved by the PSC, so I've pinged @rjbs.

rjbs:
Yes, sounds good to me.

rjbs (approved):
(I am here to +1 the concept, not the source, which I'm not great for, but I see that's been reviewed!)

PR #19126, opened 2021-09-14
Remove "skip if no PerlIO" code from tests in ext/

nwc10:
It's been 9 years since we disabled the ability to `Configure` without PerlIO, hence all these bits of conditional logic are dead code and should go.

I left the similar code in tests in `t/` (for now) as *micro*perl isn't configured, and builds without PerlIO (if it builds at all). I'm not sure whether microperl is supposed to attempt tests, and also what the future holds for it, so didn't want to get digressed and delay these improvements.

### DIFF
--- a/ext/B/t/concise.t
+++ b/ext/B/t/concise.t
@@ -42,7 +42,6 @@ like($out, qr/print/, "'-exec' option output has print opcode");
 
 ######## API tests v.60 
 
-use Config;	# used for perlio check
 B::Concise->import(qw( set_style set_style_standard add_callback 
 		       add_style walk_output reset_sequence ));
 
@@ -70,10 +69,7 @@ eval {  walk_output($foo) };
 is ($@, '', "walk_output() accepts obj that can print");
 
 # test that walk_output accepts a HANDLE arg
-SKIP: {
-    skip("no perlio in this build", 4)
-        unless $Config::Config{useperlio};
-
+{
     foreach my $foo (\*STDOUT, \*STDERR) {
 	eval {  walk_output($foo) };
 	is ($@, '', "walk_output() accepts STD* " . ref $foo);
@@ -129,11 +125,8 @@ sub render {
     return $out;
 }
 
-SKIP: {
+{
     # tests output to GLOB, using perlio feature directly
-    skip "no perlio on this build", 127
-	unless $Config::Config{useperlio};
-    
     set_style_standard('concise');  # MUST CALL before output needed
     
     @options = qw(
@@ -413,7 +406,7 @@ $out = runperl ( switches => ["-MO=Concise,-stash=Data::Dumper,-src,-exec"],
 
 SKIP: {
     skip "Data::Dumper is statically linked", 1
-	if $Config{static_ext} =~ m|\bData/Dumper\b|;
+	if $Config::Config{static_ext} =~ m|\bData/Dumper\b|;
     like($out, qr/FUNC: \*Data::Dumper::format_refaddr/,
 	"stash rendering loads package as needed");
 }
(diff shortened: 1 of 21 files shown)


Leont (approved):

ext/B/t/concise.t
```diff
@@ -71,9 +70,6 @@ is ($@, '', "walk_output() accepts obj that can print");
 
 # test that walk_output accepts a HANDLE arg
 SKIP: {
-    skip("no perlio in this build", 4)
-        unless $Config::Config{useperlio};
-
```
jkeenan, line 76: If we're removing the `skip()` function, shouldn't we also remove the `SKIP: { }` block (and de-intent the remaining code as we're doing in `ext/B/t/o.t` farther down)?

ext/B/t/concise.t
```diff
@@ -131,9 +127,6 @@ sub render {
 
 SKIP: {
     # tests output to GLOB, using perlio feature directly
-    skip "no perlio on this build", 127
-	unless $Config::Config{useperlio};
-
```
jkeenan, line 136: As above: If we're removing the `skip()` function, shouldn't we also remove the `SKIP: { }` block (and de-intent the remaining code)?

ext/B/t/showlex.t
```diff
@@ -49,10 +48,6 @@ for $newlex ('', '-newlex') {
 
     print $out if $verbose;
 
-SKIP: {
-    skip "no perlio in this build", 5
-    unless $Config::Config{useperlio};
-
     our $buf = 'arb startval';
     my $ak = B::Showlex::walk_output (\$buf);
```
jkeenan, line 53: Can we de-indent some lines here because we're no longer using a `SKIP:{}` block?

jkeenan (changes requested):

ext/B/t/showlex.t
nwc10: Not that one - GitHub didn't do you any favours with context. The two diff hunks look like this:

```
@@ -49,10 +48,6 @@ for $newlex ('', '-newlex') {

     print $out if $verbose;

-SKIP: {
-    skip "no perlio in this build", 5
-    unless $Config::Config{useperlio};
-
     our $buf = 'arb startval';
     my $ak = B::Showlex::walk_output (\$buf);

@@ -102,4 +97,3 @@ SKIP: {
     $walker->();

 }
-}

```

ie when the `SKIP` block was added, the lines within it weren't indented.

ext/B/t/concise.t
nwc10: Yes, whoops, I should have removed the `SKIP:` label. (I did in other places.) Well spotted- thanks. I did this by amending the existing commit, as I thought it would clutter the history if I did it as another commit.

As to re-indent, I wouldn't argue that we should *automatically* re-indent, if there's a lot of version history that it's going to conceal - ie reading through that would be harder than reading a block.

But in this case, yes, it seems that nothing substantive changed in these tests since 2004, and prior to then there isn't a lto of bug-fix detail in different commits.

So yes, something I should have thought about and done first time. I've done it as a separate commit, so that it's clear what is going on, and re-ordered it immediately after the commit that removes the `SKIP:` labels.

I rebased the branch onto blead and re-pushed. I don't know if that confuses GitHub...

ext/B/t/concise.t
nwc10: As above. (done)

ext/B/t/showlex.t
jkeenan: Yes, I have confirmed that.

jkeenan (approved):

PR #19127, opened 2021-09-14
Use fabs/fabsl/fabsq for NVs in pp_abs

t-a-k:
In many floating point formats, `fabs*()` are rather simple operations such as just clearing the sign bit, and will be slightly faster and compacter than a conditional negation (especially with optimizing compilers).


tonycoz:
With gcc the fabs/fabsl() uses a built-in which compiles down to a single instruction.

I though our long double troublemaker netbsd might have problems, but it appears to have a reasonable fabsl() implementation.

PR #19128, opened 2021-09-14
Test SvIsBOOL using XS::APItest

leonerd:
I recently added SvIsBOOL and tested it via Scalar::Util, but really I should have tested it directly in core.

This PR copies most of the unit tests out of `cpan/Scalar-List-Util` into the XS APItest area, so it's more core-ish. I can then delete most of them from the CPAN side later.

The tests about tied variables aren't copied, because those test a specific part of the Scalar::Util XS wrapping function, rather than the actual core macro.

### DIFF
--- a/ext/XS-APItest/t/boolean-thr.t
+++ b/ext/XS-APItest/t/boolean-thr.t
@@ -1,0 +1,38 @@
+#!./perl
+
+use strict;
+use warnings;
+
+use Config ();
+use if !$Config::Config{usethreads}, 'Test::More',
+    skip_all => "This perl does not support threads";
+
+use Test::More;
+use XS::APItest;
+
+use threads;
+use threads::shared;
+
+ok(threads->create( sub { SvIsBOOL($_[0]) }, !!0 )->join,
+    'value in to thread is bool');
+
+ok(SvIsBOOL(threads->create( sub { return !!0 } )->join),
+    'value out of thread is bool');
+
+{
+    my $var = !!0;
+    ok(threads->create( sub { SvIsBOOL($var) } )->join,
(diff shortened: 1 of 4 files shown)


ext/XS-APItest/t/boolean-thr.t
```diff
@@ -0,0 +1,37 @@
+#!./perl
+
+use strict;
+use warnings;
+
+use Test::More;
+use XS::APItest;
+
+use Config ();
+BEGIN { plan skip_all => "perl does not support threads" unless $Config::Config{usethreads} };
```
Grinnz: Could also be written `use if !$Config::Config{usethreads}, 'Test::More', skip_all => ...;` (done before `use Test::More;`)

leonerd: If the current CI run tests OK, I'll rewrite it as that and squash it down

ext/XS-APItest/t/boolean-thr.t
```diff
@@ -0,0 +1,38 @@
+#!./perl
+
+use strict;
+use warnings;
+
+use Config ();
+use if !$Config::Config{usethreads}, 'Test::More',
+    skip_all => "perl does not support threads";
```
khwilliamson: perl does support threads.  *This* perl does not

khwilliamson (commented):
LGTM

ext/XS-APItest/t/boolean-thr.t
leonerd: Fiiiiine

[test-SvIsBOOL-using-XS-APItest 1aae98d682] Wording fix in test skip message

PR #19099, opened 2021-09-06
Add README.md file in .github directory

rwp0:
Provide formatting with [Markdown](https://daringfireball.net/projects/markdown/) (as is suitable for GitHub repositories)

Hide the file in GitHub-specific [.github](https://github.com/Perl/perl5/tree/blead/.github) directory

Format code, and link to webpages where appropriate

-----

Improve [_main_](https://github.com/Perl/perl5/blob/blead/README) (first-seen) source code documentation readability,
HTML webpage formatting being very easy on readers' eyes

See also: [#17984](https://github.com/Perl/perl5/pull/17984) and #17227

PR #19130, opened 2021-09-15
Rename README to README.md

quick007:
There is literally no reason not to do this, it cleans up the readme, makes it look more presentable, and fixed the parts where they intended formatting but (seemingly?) forgot to use the correct file type


rwp0:
See #19099 

---

the base README is *not* Markdown

jkeenan:
> There is literally no reason not to do this, it cleans up the readme, makes it look more presentable, and fixed the parts where they intended formatting but (seemingly?) forgot to use the correct file type

I strongly oppose this and all attempts to reformat `README` as anything other than plain text.

rwp0:
> > There is literally no reason not to do this, it cleans up the readme, makes it look more presentable, and fixed the parts where they intended formatting but (seemingly?) forgot to use the correct file type
> 
> I strongly oppose this and all attempts to reformat `README` as anything other than plain text.

Right, and therefore I think an additional `README.md` file should be set for GitHub (repository home)

Grinnz:
As stated, this README is not in markdown format, and its filename must not change.

ilmari:
GitHub supports POD too, so if anything we could rewrite (and rename) it to POD, like INSTALL is (which could also be renamed, so GitHub renders it nicely, or maybe we could add `linguist-language=Pod` to `.gitattributes`?)

haarg:
linguist-language is used by GitHub for language stats, but not for deciding how to render a README type file. Adjusting the current README to be markdown without renaming it would be possible if that worked.

rwp0:
> GitHub supports POD too, so if anything we could rewrite (and rename) it to POD, like INSTALL is (which could also be renamed, so GitHub renders it nicely, or maybe we could add `linguist-language=Pod` to `.gitattributes`?)

I initially thought so as well, but the newer Code of Conduct and Security files in the repository, which are in Markdown, and not in Pod is making Markdown the new standard for Github-oriented files.

Grinnz:
There is no standard. POD and markdown are equally viable for Github-oriented files. Markdown is more easily read as a plaintext file and so may be more suited to this particular case. POD is capable of section linking and module linking and so is more suitable for other cases.

PR #19133, opened 2021-09-16
Test that for's iterator aliases the iterated list

nwc10:


t/op/for.t
```diff
@@ -548,12 +550,15 @@ for my $i (reverse (map {$_} @array, 1)) {
 }
 is ($r, '1CBA', 'Reverse for array and value via map with var');
 
-is do {17; foreach (1, 2) { 1; } }, '', "RT #1085: what should be output of perl -we 'print do { foreach (1, 2) { 1; } }'";
+{
+    no warnings 'void';
+    is do {17; foreach (1, 2) { 1; } }, '', "RT #1085: what should be output of perl -we 'print do { foreach (1, 2) { 1; } }'";
```
jkeenan, line 555: Can you re-break this line at the description so that it's not running > 120 characters long?

Otherwise, this is great.

nwc10: Yes, good point. Done and now in blead.

PR #19134, opened 2021-09-16
Untangle the hash delete code from the MRO update code in hv_common

nwc10:
Move the code that removed the entry from the hash ahead of all the code that post-processes that entry to update method caches.

The entry is now completely deleted from the hash before any needed MRO related logic is triggered, and the code that is specific to the chosen hash table implementation is now untangled from the MRO code.

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -1354,13 +1354,6 @@ S_hv_delete_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
             Safefree(key);
 
         sv = d_flags & G_DISCARD ? HeVAL(entry) : sv_2mortal(HeVAL(entry));
-        HeVAL(entry) = &PL_sv_placeholder;
-        if (sv) {
-            /* deletion of method from stash */
-            if (isGV(sv) && isGV_with_GP(sv) && GvCVu(sv)
-             && HvENAME_get(hv))
-                mro_method_changed_in(hv);
-        }
 
         /*
          * If a restricted hash, rather than really deleting the entry, put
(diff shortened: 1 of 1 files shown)


hv.c
```diff
@@ -1257,11 +1257,13 @@ S_hv_delete_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
                             " a restricted hash");
         }
 
+        sv = HeVAL(entry);
```
hvds: Moving this up another 10 lines would allow another 3 `HeVAL()` entries to be replaced.

Note also there's a typo in the commit message: `s/HvVAL/HeVAL/`.

hvds (commented):
The rest looks plausible to me. It might be further improved if the "someone is deleting a package" chunk could be moved out to its own function - I _think_ it would only need to return a value for `mro_changes`.

nwc10:
> It might be further improved if the "someone is deleting a package" chunk could be moved out to its own function - I _think_ it would only need to return a value for `mro_changes`.

I knew that there was a bit more that *could* be done, but I stopped as soon as I had enough untangled to make progress on the Robin Hood stuff.

Having also been looking at code in `mro_core.c` I **think** that it's possible that the call `mro_method_changed_in(hv)` changes no data structures in common with either call `mro_isa_changed_in(hv)` or ` mro_package_moved(NULL, stash, gv, 1)` - *ie* it's possible that those two calls could be moved to the point where currently `mro_changes` is assigned to.

But I didn't have the headspace at the time to investigate further.

If this is actually the case, then the code can get even simpler still.

Bur I don't want to dig into that yet - I have a branch with about 80 commits on it and a bunch of other open things, and I'm really keen to land them/discard them before getting any more "TODO" tasks.

hv.c
nwc10: > Moving this up another 10 lines would allow another 3 `HeVAL()` entries to be replaced.

Why didn't I notice that? :-(

Done - definitely an improvement.

> Note also there's a typo in the commit message: `s/HvVAL/HeVAL/`.

Thanks. Fixed.

I've fixed both of these, rebased on blead and pushed the branch again.

hvds:
LGTM

PR #19135, opened 2021-09-16
untie should release the tie'd hashes iterator state

nwc10:
Iterating a tied hash causes perl to store a copy of the current hash key to track the iteration state, with this stored copy passed as the second parameter to `NEXTKEY`. This internal state is freed immediately when tie hash iteration completes, or if the hash is destroyed, but due to an implementation oversight, it was not freed if the hash was untied. In that case, the internal copy of the key would persist until the earliest of

1. `tie` was called again on the same hash
2. The (now untied) hash was iterated (ie passed to any of `keys`, `values` or `each`)
3. The hash was destroyed.

This inconsistency is now fixed - the internal state is now freed immediately by `untie`.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -418,7 +418,39 @@ files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
 =item *
 
-XXX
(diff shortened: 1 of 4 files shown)


tonycoz:
Looks good to me.

PR #19136, opened 2021-09-16
Was there a way to run CI without a draft PR?

nwc10:


tonycoz:
Yes, just push a branch.

For mirrors you need to set a github secret to run all the jobs - see the workflow.

On the branch page like https://github.com/nwc10/perl5/tree/ASAN-PURIFY-github you'll see a little green dot (running), a tick (success) or a cross in the header immediately above the file list.

Issue #19100, opened 2021-09-06
gcc12: new build-time warnings associated with 1604cfb027


jkeenan:
In https://github.com/Perl/perl5/issues/19098, user @Jamaika1 reports C-level warnings emanating from utf8.c and utf8.h using a devel version of gcc12.  Context is sparse; I've requested more detail.

That prompted me to install a devel version of gcc12 on FreeBSD12 (one slightly earlier than used by reporter).  I encountered build-time warnings that I did not encounter when building on the same platform with gcc10.
```
$ gcc12 --version | head -n 2
gcc12 (FreeBSD Ports Collection) 12.0.0 20210822 (experimental)
Copyright (C) 2021 Free Software Foundation, Inc.
```
I then built Perl 5 blead with `sh ./Configure -des -Dusedevel -Duseithreads -Dcc=gcc12` and parsed the build log for C-level warnings:
```
$ report-build-warnings 756f0910f9.freebsd.gcc12.threaded.maketp.output.txt.gz 
File:  756f0910f9.freebsd.gcc12.threaded.maketp.output.txt.gz

  Warray-bounds                              6
  Wc++-compat                              240
  Wempty-body                                6
  Wmaybe-uninitialized                       2
  Wsign-compare                              4
  Wtype-limits                               2
```
The last previous version of gcc I've build perl with on this machine was gcc10.  At that time I got these build-time warnings:
```
$ report-build-warnings f72873660d.freebsd.gcc10.maketp.output.txt.gz
File:  f72873660d.freebsd.gcc10.maketp.output.txt.gz

  Wc++-compat                              240
  Wempty-body                                6
  Wsign-compare                              4
  Wtype-limits                               2
```
So the `Warraybounds` warnings (6) and the `Wmaybe-uninitialized` warnings (2) are new in either gcc11 or gcc12; the others appear to have existed at least from gcc10.

The `Warraybounds` warnings are these:
```
  {
    char   => 9,
    group  => "Warray-bounds",
    line   => 461,
    source => "hv.h",
    text   => "array subscript 0 is outside array bounds of 'HEK[768614336404564650]' {aka 'struct hek[768614336404564650]'}",
  },
  {
    char   => 9,
    group  => "Warray-bounds",
    line   => 461,
    source => "hv.h",
    text   => "array subscript 0 is outside array bounds of 'HEK[768614336404564650]' {aka 'struct hek[768614336404564650]'}",
  },
...
  {
    char   => 9,
[...]

PR #19137, opened 2021-09-17
gh19100: avoid buggy "maybe used uninitialized" warning

hvds:
New warnings introduced around gcc-10 or -11 assume that the call to
pthread_getspecific() in S_init_tls_and_interp may read through the
my_perl pointer provided; writing to any element of the structure
is enough to shut it up.

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -195,6 +195,12 @@ perl_alloc(void)
     /* Newx() needs interpreter, so call malloc() instead */
     my_perl = (PerlInterpreter*)PerlMem_malloc(sizeof(PerlInterpreter));
 
+#if defined(MULTIPLICITY) && PERL_IS_GCC
+    /* Write to any element to avoid buggy "maybe used uninitialized" warning
+     * (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102329) */
+    my_perl->Istack_sp = 0;
+#endif
+
     S_init_tls_and_interp(my_perl);
 #ifndef PERL_TRACK_MEMPOOL
     return (PerlInterpreter *) ZeroD(my_perl, 1, PerlInterpreter);


jkeenan:
> New warnings introduced around gcc-10 or -11 ...

Probably gcc-11.  I didn't get these with gcc-10.  I was unable to install 11 but in blead am getting them with gcc-12-devel.
```
[perlmonger: make-output] $ ~/bin/perl/parse-one-warnings-group 0f74802e76.freebsd.gcc10.unthreaded.make.output.txt.gz 'Wmaybe-uninitialized'
File:  0f74802e76.freebsd.gcc10.unthreaded.make.output.txt.gz

Warnings group 'Wmaybe-uninitialized' not found at /home/jkeenan/bin/perl/parse-one-warnings-group line 31.
[perlmonger: make-output] $ ~/bin/perl/parse-one-warnings-group 3c71934978.freebsd.gcc12.threaded.make.output.txt.gz 'Wmaybe-uninitialized'
File:  3c71934978.freebsd.gcc12.threaded.make.output.txt.gz

[
  {
    char   => 21,
    group  => "Wmaybe-uninitialized",
    line   => 393,
    source => "thread.h",
    text   => "'my_perl' may be used uninitialized",
  },
  {
    char   => 21,
    group  => "Wmaybe-uninitialized",
    line   => 393,
    source => "thread.h",
    text   => "'my_perl' may be used uninitialized",
  },
]

> assume that the call to
> pthread_getspecific() in S_init_tls_and_interp may read through the
> my_perl pointer provided; writing to any element of the structure
> is enough to shut it up.
```
Your patch silences these warnings on gcc-12.
```
[perlmonger: make-output] $ ~/bin/perl/parse-one-warnings-group  0ce638951b.freebsd.gcc12.threaded.maketp.output.txt.gz 'Wmaybe-uninitialized'
File:  0ce638951b.freebsd.gcc12.threaded.maketp.output.txt.gz

Warnings group 'Wmaybe-uninitialized' not found at /home/jkeenan/bin/perl/parse-one-warnings-group line 31.
```
LGTM

nwc10:
I looked at it and kept thinking "what is the rest of structure initialisation?" and "but couldn't we move the entire structure initialisation earlier?"

and then wondered what `ZeroD` and `Zero` were
and then tried to move code around. and looked at what macros expanded to
and started an edit but only got halfway before

in the end, I think

* `S_init_tls_and_interp` doesn't actually read or write the contents of the recently allocated memory
* `ZeroD` zeros all of it
* nothing else touches the contents
* all the "memory wrap" macros *inside* the other macros can't "trigger"

and I think that ultimately

1. `perl_alloc` should actually be the following:
2. I *think* that this will silence the warning

```PerlInterpreter *
perl_alloc(void)
{
    PerlInterpreter *my_perl = (PerlInterpreter*)PerlMem_calloc(1, sizeof(PerlInterpreter));

    S_init_tls_and_interp(my_perl);
#ifdef PERL_TRACK_MEMPOOL
    INIT_TRACK_MEMPOOL(PL_memory_debug_header, my_perl);
#endif
    return my_perl;
}
#endif /* PERL_IMPLICIT_SYS */
```

(pasting the finished function in seems easier than trying to create a diff or a branch)

If `perl_alloc` is changed to that (and no other changes from blead) does this **also** silence the compiler warning?

nwc10:
If the above comment is correct, then `perl_alloc_using` should change from this:

```
    /* Newx() needs interpreter, so call malloc() instead */
    my_perl = (PerlInterpreter*)(*ipM->pMalloc)(ipM, sizeof(PerlInterpreter));
    S_init_tls_and_interp(my_perl);
    Zero(my_perl, 1, PerlInterpreter);

```

to this

```
    my_perl = (PerlInterpreter*)(*ipM->pCalloc)(ipM, sizeof(PerlInterpreter));
    S_init_tls_and_interp(my_perl);
```


I think that these are the only two locations I can spot with "malloc" followed by "Zero"

Also looking at this made me realise that the expansion of `PERL_SET_CONTEXT` is subtly daft:

```
    STMT_START {						\
        int _eC_;						\
        if ((_eC_ = pthread_setspecific(PL_thr_key, (void *)(t))))	\
            Perl_croak_nocontext("panic: pthread_setspecific (%d) [%s:%d]",	\
                                 _eC_, __FILE__, __LINE__);	\
    } STMT_END
```

What's the first thing that `Perl_croak_nocontext` does?

```
    dTHX;
```

Yay. Reading TLS and getting who knows what...

given that that variable `t` **is** a Perl context, and meant to be the current one, I guess that code (and all its copied variants) should be:

```
            Perl_croak((t), "panic: pthread_setspecific (%d) [%s:%d]",	\
                                 _eC_, __FILE__, __LINE__);	\
```

hvds:
> If `perl_alloc` is changed to that (and no other changes from blead) does this **also** silence the compiler warning?

Yes, that also silences it.

nwc10:
> Yes, that also silences it.

I failed to say - I did build this with threads and without threads, and both passed all tests. Plus looked at the pre-processed output for both. So it's "no obvious bugs", but I wasn't 100% confident that I hadn't missed something.

But if it's both correct and warning free, then I think we should go with it in preference to commit 0ce638951b075ffaf34d8c4d68ca876625503563 because it's both clearer code and calls fewer functions.

(I'm not sure why I spotted this - I guess you've been slogging away at this for several days, and I've come in fresh. But it all seems a bit unfair - you've done all the real work here)

hvds:
> > Yes, that also silences it.
> 
> I failed to say - I did build this with threads and without threads, and both passed all tests. Plus looked at the pre-processed output for both. So it's "no obvious bugs", but I wasn't 100% confident that I hadn't missed something.
> 
> But if it's both correct and warning free, then I think we should go with it in preference to commit [0ce6389](https://github.com/Perl/perl5/commit/0ce638951b075ffaf34d8c4d68ca876625503563) because it's both clearer code and calls fewer functions.

Agreed.

> (I'm not sure why I spotted this - I guess you've been slogging away at this for several days, and I've come in fresh. But it all seems a bit unfair - you've done all the real work here)

I'm happy if we end up with better code; being too close has its drawbacks too.

> What's the first thing that `Perl_croak_nocontext` does?
> 
> ```
>     dTHX;
> ```
> 
> Yay. Reading TLS and getting who knows what...

Yes, that brings me back to the other issue: we need a mechanism to fail (preferably with a message) before the interpreter is fully set up, and after we've started to break it down. [hv/safe-croak](https://github.com/Perl/perl5/compare/blead...hv/safe-croak) was intended to be a start at that, but I'm not sure where to go with it. (And #19100 has had rather too much going on, so this has got lost in that noise.)

jkeenan:
> > Yes, that also silences it.
> 
> I failed to say - I did build this with threads and without threads, and both passed all tests. Plus looked at the pre-processed output for both. So it's "no obvious bugs", but I wasn't 100% confident that I hadn't missed something.
> 
> But if it's both correct and warning free, then I think we should go with it in preference to commit [0ce6389](https://github.com/Perl/perl5/commit/0ce638951b075ffaf34d8c4d68ca876625503563) because it's both clearer code and calls fewer functions.
> 

@nwc10, could you provide a p.r. which implements "it" in the paragraph above, so that we can put "it" through its paces?

Thank you very much.
Jim Keenan

nwc10:
"it" is now at #19141

"it" was

1. `git stash pop`
2. try to write a coherent commit message based on what I wrote above
3. build and pass tests locally

so there might be daft mistakes in it.

hvds:
#19141 is clearly preferable, closing this.

PR #19138, opened 2021-09-17
gh19010: fix returns for Perl_grok_infnan

hvds:
Consistently honour what the docs have always promised:
  If an infinity or a not-a-number is recognized, C<*sp> will point to
  one byte past the end of the recognized string.  If the recognition fails,
  zero is returned, and C<*sp> will not move.

Additionally, restore Perl_grok_number_flags to allowing inf/nan with
trailing garbage only when called with PERL_SCAN_TRAILING; add notes
to the other two core callers to clarify that they always accept such
trailing garbage.

A small number of XS-APItest tests were modified to reflect the stricter
behaviour: "Infin" and "nanx" are now invalid without PERL_SCAN_TRAILING.

### DIFF
--- a/numeric.c
+++ b/numeric.c
@@ -958,7 +958,7 @@ Perl_grok_infnan(pTHX_ const char** sp, const char* send)
         else
             return 0;
     }
-    /* NOTREACHED */
+    NOT_REACHED; /* NOTREACHED */
 
     /* We parsed something valid, s points after it, flags describes it */
   ok_check_space:
--- a/numeric.c
+++ b/numeric.c
@@ -771,15 +771,15 @@ Perl_grok_infnan(pTHX_ const char** sp, const char* send)
         s++; if (s == send || isALPHA_FOLD_NE(*s, 'N')) return 0;
         s++; if (s == send) return 0;
         if (isALPHA_FOLD_EQ(*s, 'F')) {
-            s++;
+            flags |= IS_NUMBER_INFINITY | IS_NUMBER_NOT_INT;
+            *sp = ++s;
             if (s < send && (isALPHA_FOLD_EQ(*s, 'I'))) {
-                int fail =
-                    flags | IS_NUMBER_INFINITY | IS_NUMBER_NOT_INT | IS_NUMBER_TRAILING;
-                s++; if (s == send || isALPHA_FOLD_NE(*s, 'N')) return fail;
-                s++; if (s == send || isALPHA_FOLD_NE(*s, 'I')) return fail;
-                s++; if (s == send || isALPHA_FOLD_NE(*s, 'T')) return fail;
-                s++; if (s == send || isALPHA_FOLD_NE(*s, 'Y')) return fail;
-                s++;
+                int trail = flags | IS_NUMBER_TRAILING;
+                s++; if (s == send || isALPHA_FOLD_NE(*s, 'N')) return trail;
+                s++; if (s == send || isALPHA_FOLD_NE(*s, 'I')) return trail;
+                s++; if (s == send || isALPHA_FOLD_NE(*s, 'T')) return trail;
+                s++; if (s == send || isALPHA_FOLD_NE(*s, 'Y')) return trail;
+                *sp = ++s;
             } else if (odh) {
                 while (s < send && *s == '0') { /* 1.#INF00 */
                     s++;
@@ -787,10 +787,8 @@ Perl_grok_infnan(pTHX_ const char** sp, const char* send)
             }
             while (s < send && isSPACE(*s))
                 s++;
-            if (s < send && *s) {
-                flags |= IS_NUMBER_TRAILING;
-            }
-            flags |= IS_NUMBER_INFINITY | IS_NUMBER_NOT_INT;
+            *sp = s;
+            return flags | (s < send ? IS_NUMBER_TRAILING : 0);
         }
         else if (isALPHA_FOLD_EQ(*s, 'D') && odh) { /* 1.#IND */
             s++;
@@ -798,9 +796,8 @@ Perl_grok_infnan(pTHX_ const char** sp, const char* send)
             while (s < send && *s == '0') { /* 1.#IND00 */
                 s++;
             }
-            if (s < send && *s) {
-                flags |= IS_NUMBER_TRAILING;
-            }
+            *sp = s;
+            return flags | (s < send ? IS_NUMBER_TRAILING : 0);
         } else
             return 0;
     }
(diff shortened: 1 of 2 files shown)


hvds:
I'm not sure who is best placed to review this, but I'm asking @Perl/perl-steering-council for blessing as well since the work requires test changes.

khwilliamson:
@sisyphus what do you think?

rjbs:
(I am not competent to have a ready opinion here.)

hvds:
> (I am not competent to have a ready opinion here.)

I think there are two things needed: someone to agree that it is valid to change the tests in this way, and someone to validate that the code achieves what the commit message claims. I assume PSC approval is appropriate for the test changes, and that our normal review process should handle checking the code.

Here is the essence of the two tests that change:
```
-  [ "Infin",0,                  undef, IS_NUMBER_INFINITY | IS_NUMBER_NOT_INT | IS_NUMBER_TRAILING ],
+  [ "Infin",0,                  undef, 0 ],
+  [ "Infin",PERL_SCAN_TRAILING, undef, IS_NUMBER_INFINITY | IS_NUMBER_NOT_INT | IS_NUMBER_TRAILING ],
...
-   [ "nanx", 0,                  undef, IS_NUMBER_NAN | IS_NUMBER_NOT_INT | IS_NUMBER_TRAILING ],
+   [ "nanx", 0,                  undef, 0 ],
    [ "nanx", PERL_SCAN_TRAILING, undef, IS_NUMBER_NAN | IS_NUMBER_NOT_INT | IS_NUMBER_TRAILING ],
```

Now the docs for `grok_number_flags` say the following, the first part essentially unchanged since @jhi first wrote it in dd5dc04f54, the second part unchanged since @tonycoz added it in 3f7602fa4cd:
```
Recognise (or not) a number.  The type of the number is returned
(0 if unrecognised), otherwise it is a bit-ORed combination of
C<IS_NUMBER_IN_UV>, C<IS_NUMBER_GREATER_THAN_UV_MAX>, C<IS_NUMBER_NOT_INT>,
C<IS_NUMBER_NEG>, C<IS_NUMBER_INFINITY>, C<IS_NUMBER_NAN> (defined in perl.h).
...
C<flags> allows only C<PERL_SCAN_TRAILING>, which allows for trailing
non-numeric text on an otherwise successful I<grok>, setting
C<IS_NUMBER_TRAILING> on the result.
```

So the intent here is that trailing garbage is a fail - we return 0 - unless the caller asks for it with `PERL_SCAN_TRAILING`, and so it never makes sense for a test to call `grok_number_flags` with 0 for flags and expect `IS_NUMBER_TRAILING` set on the return value.

Does that help?

hvds:
@t-a-k you've made some improvements to inf/nan handling in some areas recently, would you be able to take a look at the code here?

t-a-k:
It looks basically good to me, but there are some points to be worried about:
* Trailing spaces after inf/nan are normally ignored (and exempt from "isn't numeric" warnings), but spaces after "`1.#IND`" are not.  Is this intentional?
* `grok_infnan` would eventually have duplicated codes for `while (s < send && isSPACE(*s)) s++; *sp = s; return flags | (s < send ? IS_NUMBER_TRAILING : 0);`. I  suggest to consolidate them into one place, e.g. at the end of this function.

hvds:
> It looks basically good to me, but there are some points to be worried about:
> 
> * Trailing spaces after inf/nan are normally ignored (and exempt from "isn't numeric" warnings), but spaces after "`1.#IND`" are not.  Is this intentional?
> 
> * `grok_infnan` would eventually have duplicated codes for `while (s < send && isSPACE(*s)) s++; *sp = s; return flags | (s < send ? IS_NUMBER_TRAILING : 0);`. I  suggest to consolidate them into one place, e.g. at the end of this function.

Thanks very much @t-a-k, I'd missed the `"1.#IND"` case, and looking more closely showed further cases to consider; that then also led me to a (minor) off-by-one error; and expanding the tests for this then showed up an additonal unrelated inconsistency in `Perl_grok_number_flags` (#19464).

I've pushed two additional commits, one to fix up the code and the second to abstract out the tests so that we check every case both with and without `PERL_SCAN_TRAILING`. In case it's useful I've also pushed a rebased branch [hv/gh19010-rebase](https://github.com/Perl/perl5/tree/hv/gh19010-rebase) with the commits squashed.

Please take another look; I'm interested to know also whether the _motivation_ of this PR makes sense to you, ie the assumption that what the docs promise is sane and useful, and that we should fix the code to match the docs.

t-a-k:
> Please take another look; I'm interested to know also whether the _motivation_ of this PR makes sense to you, ie the assumption that what the docs promise is sane and useful, and that we should fix the code to match the docs.

`grok_infnan` is publicly (in `perlapi`) documented to update `*sp` when it recognized Inf/NaN, so I think that failing to update `*sp` properly should be considered as a bug to be fixed; therefore I agree with the motivation of this PR.

numeric.c
```diff
@@ -771,36 +771,29 @@ Perl_grok_infnan(pTHX_ const char** sp, const char* send)
...
         else if (isALPHA_FOLD_EQ(*s, 'D') && odh) { /* 1.#IND */
             s++;
             flags |= IS_NUMBER_NAN | IS_NUMBER_NOT_INT;
             while (s < send && *s == '0') { /* 1.#IND00 */
                 s++;
             }
-            if (s < send && *s) {
-                flags |= IS_NUMBER_TRAILING;
-            }
+            goto ok_check_space;
```
t-a-k, line 796: Are these `goto ok_check_space;`'s really needed?  Enclosing blocks seem to naturally exit to `ok_check_space:` even without them.

hvds: I think they're valuable for consistency: the intent is that `goto ok_check_space` confirms we have met its prerequisites: "We parsed something valid, s points after it, flags describes it", hence also the `/* NOTREACHED */` to discourage simply falling out of the block. I'd expect any C compiler to generate the same code either way, this is just for the benefit of humans.

t-a-k: Oh, that makes sense.
Then, how about using `NOT_REACHED` macro in addition to `/* NOTREACHED */`?

hvds: > Oh, that makes sense. Then, how about using `NOT_REACHED` macro in addition to `/* NOTREACHED */`?

Ah yes, I had forgotten that we also had a macro for it. Now added, and also updated hv/gh19010-rebase.

hvds:
Thanks @khwilliamson, and thanks @t-a-k for helping to unblock this.

PR #19140, opened 2021-09-17
CI tests for ASAN, `make install` and `PERL_UNICODE`

nwc10:
This branch adds several new CI tests, refactors `testsuite.yml` somewhat, but probably **reduces** the wallclock time for all CI tests to complete.

The "critical path" for all tests to complete is

1. Sanity (23m)
2. cygwin (52m)

This branch

1. replaces `make test` with `make test_harness` in Sanity (`test` does not run tests in parallel) - saving at least 6m
2. replace `make` and `make test` with `make test_prep` and `./perl harness` - saving roughly 1m

`make -j2 test` first needs to (re)check that all the extensions are up to date before getting to the real work - on Cygwin this takes at least 40s.

This likely reduces the total wallclock time to complete all tests by about 10 minutes. The elapsed time seems quite variable, I assume due to load from other VMs on the same host. cygwin is still the slowest test - the addition of the ASAN tests does not change the critical path to completion.

New tests

1. ASAN (so `malloc`/`free` goofs, but also agressive leak checking)
2. `make install`
3. `PERL_UNICODE=""` and UTF-8 locales
4. shared libraries on macOS
5. minitest
5. building **without** `-DDEBUGGING`
6. Adding `MALLOC_PERTURB_` and `MALLOC_CHECK_` to (most) Linux builds

We already had configurations that looked like they were testing `PERL_UNICODE=""` and UTF-8 locales, but these didn't actually do anything special.

This branch incorporates #19115, as without those fixes the ASAN builds fail.


nwc10:
And now cygwin hates me, but for some reason blead is fine.

sv.c
```diff
@@ -6788,8 +6789,10 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
                  * set SvLEN to whatever value was in the now-freed
                  * regex body. The PVX buffer is shared by multiple re's
                  * and only freed once, by the re whose len in non-null */
-                STRLEN len = ReANY(sv)->xpv_len;
+                struct regexp *r = ((XPV*)SvANY(sv))->xpv_len_u.xpvlenu_rx;
```
hvds, line 6811: Unless I've missed a reason, I think this should stick with `struct regexp *r = ReANY(sv);` - I see no benefit to manually expanding the inline function, and some risk if `ReANY` changes again in the future.

Also in the commit message,  I think you intended to follow `with commit df6b4bd56551f2d3:` with that commit's subject line `give REGEXP SVs the POK flag again`.

.github/workflows/testsuite.yml
```diff
@@ -464,3 +464,53 @@ jobs:
...
+      CONTINUOUS_INTEGRATION: 1
+      WORKSPACE: ${{ github.workspace }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # exercise a variety of build options
+        # threads often cause build issues
+        CONFIGURE_ARGS:
+          - "-Dusethreads"
```
hvds: Do we want/need also with and without debugging for this ASAN build? I've been noticing recently that gcc can act quite differently (and sometimes buggily) in the presence of asserts. I've no clue what the cost/benefit ratio looks like though.

.github/workflows/testsuite.yml
```diff
@@ -53,6 +53,10 @@ jobs:
   #                       |__/
   # ascii art small font from https://ascii.co.uk/text
 
+  # "sanity_check" tests -Dusethreads without -DDEBUGGING
+  # "linux-i386" tests -Uusethreads without -DDEBUGGING
+  # All other builds test with -DDEBUGGING
```
hvds: As far as I can see, it's just `linux`, `make install` and `cygwin` that build with DEBUGGING. Was this intended to say "all other _linux_ builds"?

.github/workflows/testsuite.yml
```diff
@@ -145,14 +151,14 @@ jobs:
...
-          MALLOC_PERTURB_=254 MALLOC_CHECK_=3 make -j2
+          MALLOC_PERTURB_=254 MALLOC_CHECK_=3 make -j2 test_prep
       - name: Show Config
         run: |
           LD_LIBRARY_PATH=. ./perl -Ilib -V
           LD_LIBRARY_PATH=. ./perl -Ilib -e 'use Config; print Config::config_sh'
       - name: Run Tests
         run: |
-          MALLOC_PERTURB_=254 MALLOC_CHECK_=3 TEST_JOBS=2 make -j2 test_harness
+          LD_LIBRARY_PATH=`pwd` MALLOC_PERTURB_=254 MALLOC_CHECK_=3 TEST_JOBS=2 ./perl t/harness
```
hvds, line 161: Why "`pwd`" here rather than "." as above? Should the above be changed too? (And similarly for the FIXUP commit.)

hvds (commented):
This changed to draft status while I was looking at it, but I'll dump some comments here anyway.

sv.c
nwc10: > Unless I've missed a reason, I think this should stick with `struct regexp *r = ReANY(sv);` - I see no benefit to manually expanding the inline function, and some risk if `ReANY` changes again in the future.

The diff hunk doesn't really give enough context. What happened was..

I started with code in `sv.c` that reads like this (in blead)

```
            if (isREGEXP(sv)) {
                /* SvLEN points to a regex body. Free the body, then
                 * set SvLEN to whatever value was in the now-freed
                 * regex body. The PVX buffer is shared by multiple re's
                 * and only freed once, by the re whose len in non-null */
                STRLEN len = ReANY(sv)->xpv_len;
                pregfree2((REGEXP*) sv);
                SvLEN_set((sv), len);
                goto freescalar;
            }
```

and I read it, and thought "wait, what's going on here. The comment mentions `SvLEN` but I don't see any sign of `SvLEN`."

So then I had to figure out that the `ReANY` is actually what the `SvLEN` is talking about in the comment - it's not obvious

And then I had to go look up the definition of that

```
PERL_STATIC_INLINE struct regexp *
Perl_ReANY(const REGEXP * const re)
{
    XPV* const p = (XPV*)SvANY(re);

    PERL_ARGS_ASSERT_REANY;
    assert(isREGEXP(re));

    return SvTYPE(re) == SVt_PVLV ? p->xpv_len_u.xpvlenu_rx
                                   : (struct regexp *)p;
}
```

at which point I thought "eh, what's that ternary doing? Sometimes it isn't `SvLEN`"

So then I had to re-read the code in `sv.c`.

Enough to look for more context:

```
        case SVt_PVLV:
            if (LvTYPE(sv) == 'T') { /* for tie: return HE to pool */
                SvREFCNT_dec(HeKEY_sv((HE*)LvTARG(sv)));
                HeNEXT((HE*)LvTARG(sv)) = PL_hv_fetch_ent_mh;
                PL_hv_fetch_ent_mh = (HE*)LvTARG(sv);
            }
            else if (LvTYPE(sv) != 't') /* unless tie: unrefcnted fake SV**  */
                SvREFCNT_dec(LvTARG(sv));
            if (isREGEXP(sv)) {
                /* SvLEN points to a regex body. Free the body, then
                 * set SvLEN to whatever value was in the now-freed
                 * regex body. The PVX buffer is shared by multiple re's
                 * and only freed once, by the re whose len in non-null */
                STRLEN len = ReANY(sv)->xpv_len;
                pregfree2((REGEXP*) sv);
                SvLEN_set((sv), len);
                goto freescalar;
            }
```

and then I had enough to mentally simplify the code in `Perl_ReANY` - it was correct, and it always took one side of the ternary.

And that was a bunch of work, involving two context switches. I didn't find this very readable.

Hence why I deliberately decided to inlinine - based on this experience I felt that it made the code more readable:

```
            if (isREGEXP(sv)) {
                /* SvLEN points to a regex body. Free the body, then
                 * set SvLEN to whatever value was in the now-freed
                 * regex body. The PVX buffer is shared by multiple re's
                 * and only freed once, by the re whose len in non-null */
                struct regexp *r = ((XPV*)SvANY(sv))->xpv_len_u.xpvlenu_rx;
                STRLEN len = r->xpv_len;
                pregfree2((REGEXP*) sv);
                del_body_by_type(r, SVt_REGEXP);
                SvLEN_set((sv), len);
                goto freescalar;
            }
 ```


I also don't think that there's real risk here - this code is "inside" the abstraction of how regexs are store in two possible places, given that it's responsible for correctly freeing up one of those two possible places. If `ReANY` needs to change, then this code needs to be changed too, or at least audited that it's still correct.


So that's how I saw it. It didn't feel like using `ReANY` worked as an abstraction, because it didn't seem to be something you could ignore while trying to understand (or verify) the code - it was getting in the way, rather than making it simpler.

File: .github/workflows/testsuite.yml (hunk 3 above)
Comment (nwc10) [reply .github/workflows/testsuite.yml]: Good catch. Fixed in the (partial) rebase.

Thanks

File: .github/workflows/testsuite.yml (hunk 4 above)
Comment (nwc10) [reply .github/workflows/testsuite.yml:161]: "Because it changes directory, so the lazy approach used previously doesn't work"...

Yes, good catch. I added a commit in the rebase that changes all those `.`s to ```pwd```s

Thanks.

File: sv.c (hunk 1 above)
Comment (hvds) [reply sv.c:6811]: > If `ReANY` needs to change, then this code needs to be changed too, or at least audited that it's still correct.

In such circumstance, I'd have imagined one would primarily rely on grepping for ReANY to find what needs auditing. I'd suggest at least mentioning it in the comment: `SvLEN points to a regex body (the PVLV case in ReANY).`.

(More formally, I think it dubious to assert that SvLEN does the pointing, but I guess that's a reasonable enough shorthand.)

**Comment by nwc10** [comment]
> This changed to draft status while I was looking at it, but I'll dump some comments here anyway.

I changed it to draft because I pushed it thinking that it was all going to be good, and I only realised later that to my horror I'd done something that broke cygwin. It was only the YAML workflow that was really draft - the first two commits are in the existing PR that I referenced.

And then I found that my timing numbers were off.

I think that all that is now fixed.

File: .github/workflows/testsuite.yml (hunk 2 above)
Comment (nwc10) [reply .github/workflows/testsuite.yml]: > Do we want/need also with and without debugging for this ASAN build? I've been noticing recently that gcc can act quite differently (and sometimes buggily) in the presence of asserts. I've no clue what the cost/benefit ratio looks like though.

I don't either. In that, I believe that everything runs in parallel, and we don't seem to saturate the available hardware, meaning that "however many" Linux tests we add, cygwin will still finish last. Between the "ASAN" and "PERL_UNICODE" there are two jobs with `-Dusethreads` and only one with (implied) `-Uusethreads` so I guess moving the precise location of the `-DDEBUGGING` and having one (more) job here with (`no threads` and `no DEBUGGING`) would cover this case.

@rjbs - do you know what accounting policy github has for workflow jobs, and if we're at risk of hitting it and hence having useful jobs denied?

File: sv.c (hunk 1 above)
Comment (nwc10) [reply sv.c:6811]: > I'd suggest at least mentioning it in the comment: `SvLEN points to a regex body (the PVLV case in ReANY).`.

I started tweaking the comments. And then tried to remember exactly what was happening where and when...

It was only a week ago, and I'd already started to forget all the complexity (possibly too clever for its own good, but I can't spot a cleaner way...)

So I've just pushed commit 07e0407ab51f4df8:

```
    Add comments describing how PVLVs store REGEXPs by reference

 inline.h  | 15 +++++++++++++++
 regcomp.c | 10 ++++++++++
 sv.c      | 27 +++++++++++++++++++++++----
 3 files changed, 48 insertions(+), 4 deletions(-)
```

Quite a lot of comments. Do these work better than what we had?

If you think that they are good enough to merge, as part of the rebase I was proposing to re-order the commits so that this commit with comments was immediately after the 2 commits with code changes, and before the commits that only deal with `.github/workflows/testsuite.yml`

File: sv.c [hunk 5]
```diff
@@ -6785,10 +6788,26 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
...
+                 * normally used to store SvLEN instead points to a regex body.
+                 * Retrieving the pointer to the regex body from the correct
+                 * location is normally abstracted by ReANY(), which handles
+                 * both SVt_PVLV and SVt_REGEXP
+                 *
+                 * This code is unwinding the storage specific to SVt_PVLV.
+                 * We get the body pointer directly from the union, free it,
+                 * then set SvLEN to whatever value was in the now-freed regex
+                 * body. The PVX buffer is shared by multiple re's and only
+                 * freed once, by the re whose SvLEN in non-null.
```
hvds, line 6801: s/in/is/

sv.c
hvds: > Quite a lot of comments. Do these work better than what we had?

I think so, yes.

> If you think that they are good enough to merge, as part of the rebase I was proposing to re-order the commits so that this commit with comments was immediately after the 2 commits with code changes, and before the commits that only deal with `.github/workflows/testsuite.yml`

That makes sense to me.

hvds: > If you think that they are good enough to merge

I don't see any blocking issues, but if there's a chance to get eyeballs from someone that knows either of these areas of code better then that'd be a help. If that's unlikely any time soon, then pushing it to see what breaks also works (at least if we're comfortable on the github accounting policy).

.github/workflows/testsuite.yml
hvds: > @rjbs - do you know what accounting policy github has for workflow jobs, and if we're at risk of hitting it and hence having useful jobs denied?

I recall we tripped up on that not so long ago, but I don't remember the details; I think it may have been @atoomic that fixed things that time.

jkeenan:
On 9/23/21 10:47 AM, Hugo van der Sanden wrote:
> 
> 
> In sv.c <https://github.com/Perl/perl5/pull/19140#discussion_r714873816>:
> 
>> @@ -6788,8 +6789,10 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
>                    * set SvLEN to whatever value was in the now-freed
>                    * regex body. The PVX buffer is shared by multiple re's
>                    * and only freed once, by the re whose len in non-null */
> -                STRLEN len = ReANY(sv)->xpv_len;
> +                struct regexp *r = ((XPV*)SvANY(sv))->xpv_len_u.xpvlenu_rx;
> 
>     If you think that they are good enough to merge
> 
> I don't see any blocking issues, but if there's a chance to get eyeballs 
> from someone that knows either of these areas of code better then that'd 
> be a help. If that's unlikely any time soon, then pushing it to see what 
> breaks also works (at least if we're comfortable on the github 
> accounting policy).
> 

Please consider a smoke-me branch as well, so that we can pick up 
additional OSes.

nwc10:
> Please consider a smoke-me branch as well, so that we can pick up additional OSes.

I took the branch that this PR is coupled to, and rebased the commits onto blead as `smoke-me/nicholas/ASAN-PURIFY-github` two days ago. I haven't spotted any problems. (Am I missing them?)

sv.c
nwc10: Pants!

(Well spotted)

Do I really rebase again again to fix this? Or do I just try to remember to fix this at the point that we merge?

nwc10:
We have fails *and* passes for cygwin and i386 (for the same commit)

```
fatal: reference is not a tree: 07e0407ab51f4df89b633c7f5c7babb226161562
```

I think that this is because I

1. accidentally pushed my branch to `origin` (ie this repository)
2. then realised my mistake and deleted it
3. then pushed the updated branch to `nwc10` (where it belongs)

and the CI run was triggered by the first push, for the given SHA1, but at the point that the CI run for the first push started on those two hosts, that SHA1 had been deleted again (but the run wasn't cancelled) hence this error.

ie it's not a problem with the workflow. (which matters, when most of the branch is about workflows)

.github/workflows/testsuite.yml
nwc10: > > @rjbs - do you know what accounting policy github has for workflow jobs, and if we're at risk of hitting it and hence having useful jobs denied?
> 
> I recall we tripped up on that not so long ago, but I don't remember the details; I think it may have been @atoomic that fixed things that time.

I can't find *much* documentation that gives me confidence in an answer.


The only thing that mentions limits (well, free quotas) is https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

I assume that we are "GitHub Free for organizations" and hence that gives us 2,000 minutes per month.

Which works out at 64 minutes daily. Which I think we'd burn through on one push request.

Which must mean that this applies:

> GitHub Actions usage is free for both public repositories and self-hosted runners.

Which seems to imply that it's unlimited. Or more realistically, "unlimited".

Sigh. So I still don't *really* know.

sv.c
nwc10: We don't seem to be getting more eyeballs.

Even if we punt on the workflow changes until we get a better answer on "quota", I'd quite like to get the 3 code commits into blead, because they make the testsuite 100% clean under ASAN (ie no leaks). Right now every time I've done an ASAN build on blead or anything branched from it, I've had to re-confirm that the 3 tests I see failing are the 3 that I know "should" fail. Getting blead clean (and keeping it that way) would be good.

sv.c
hvds: My approach would be to push a fixup commit, and fix it up via rebase just before merging. But that's based on my own workflows - I know I'd always look again at the commits before merging, so I'd spot the "fixup" message and act on it. If you're not confident you'd spot that, it's fine to push it as a normal commit, or leave it out - it's no great harm if we fix it with a normal commit after the fact, just another bit of churn in the logs.

.github/workflows/testsuite.yml
hvds: > > GitHub Actions usage is free for both public repositories and self-hosted runners.
> 
> Which seems to imply that it's unlimited. Or more realistically, "unlimited".

Ah, my recollection of tripping up on this before must have been for the security repo, which is _not_ a public repository. I think we had the mirroring set up such that every CI run for the public repo was duplicated; and that we eventually worked out what was happening and changed the set up so that the duplication no longer occurs.

So I think you're right, that we are "unlimited".

sv.c
hvds: > We don't seem to be getting more eyeballs.
> 
> Even if we punt on the workflow changes until we get a better answer on "quota", I'd quite like to get the 3 code commits into blead

I'd go for it, we're early enough in the release cycle to have lots of time.

sv.c
nwc10: I fixed it before testing the rebase.

PR #19141, opened 2021-09-17
perl_alloc() wants zeroed memory so should use calloc()

nwc10:
The previous code

1) allocated memory with PerlMem_malloc()
2) passed the pointer to S_init_tls_and_interp()
3) called Zero() or ZeroD()
4) optionally invoked INIT_TRACK_MEMPOOL()
5) returned the pointer

ZeroD() and Zero() are equivalent, apart from the return value of the
expression.

The layers of functions and macros obscured what what was actually
happening, and what the ordering dependencies are:

* S_init_tls_and_interp() uses only the address of the pointer
* Zero() zeros the memory
* Only INIT_TRACK_MEMPOOL() touches the contents
* all the "memory wrap" macros inside the other macros can't "trigger"

Hence the order of Zero() and S_init_tls_and_interp() can be swapped,
at which point Zero() immediately follows malloc(), meaning that the two
should be be replaced with calloc().

This simplifies the function considerably.

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -159,9 +159,8 @@ perl_alloc_using(struct IPerlMem* ipM, struct IPerlMem* ipMS,
     PERL_ARGS_ASSERT_PERL_ALLOC_USING;
 
     /* Newx() needs interpreter, so call malloc() instead */
-    my_perl = (PerlInterpreter*)(*ipM->pMalloc)(ipM, sizeof(PerlInterpreter));
+    my_perl = (PerlInterpreter*)(*ipM->pCalloc)(ipM, 1, sizeof(PerlInterpreter));
     S_init_tls_and_interp(my_perl);
-    Zero(my_perl, 1, PerlInterpreter);
     PL_Mem = ipM;
     PL_MemShared = ipMS;
     PL_MemParse = ipMP;
--- a/perl.c
+++ b/perl.c
@@ -193,9 +193,7 @@ perl_alloc(void)
     PerlInterpreter *my_perl = (PerlInterpreter*)PerlMem_calloc(1, sizeof(PerlInterpreter));
 
     S_init_tls_and_interp(my_perl);
-#ifdef PERL_TRACK_MEMPOOL
     INIT_TRACK_MEMPOOL(PL_memory_debug_header, my_perl);
-#endif
     return my_perl;
 }
 #endif /* PERL_IMPLICIT_SYS */
--- a/perl.c
+++ b/perl.c
@@ -190,19 +190,13 @@ Allocates a new Perl interpreter.  See L<perlembed>.
 PerlInterpreter *
 perl_alloc(void)
 {
-    PerlInterpreter *my_perl;
-
-    /* Newx() needs interpreter, so call malloc() instead */
-    my_perl = (PerlInterpreter*)PerlMem_malloc(sizeof(PerlInterpreter));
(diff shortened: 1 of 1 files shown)


jkeenan:
This branch, like [Hugo's](https://github.com/Perl/perl5/pull/19137), eliminates the `-Wmaybe-uninitialized` build-time warnings when I compile it with `gcc12`:
```
$ report-build-warnings 5ab6a5d0ac.freebsd.gcc12.threaded.maketp.output.txt.gz 
File:  5ab6a5d0ac.freebsd.gcc12.threaded.maketp.output.txt.gz

  Warray-bounds                              6
  Wc++-compat                              240
  Wempty-body                                6
  Wsign-compare                              4
  Wtype-limits                               2
```

hvds:
LGTM. We'll want the same change to `perl_alloc_using`, which will be trivial.

perl.c
```diff
@@ -190,19 +190,13 @@ Allocates a new Perl interpreter.  See L<perlembed>.
...
 
     S_init_tls_and_interp(my_perl);
-#ifndef PERL_TRACK_MEMPOOL
-    return (PerlInterpreter *) ZeroD(my_perl, 1, PerlInterpreter);
-#else
-    Zero(my_perl, 1, PerlInterpreter);
+#ifdef PERL_TRACK_MEMPOOL
     INIT_TRACK_MEMPOOL(PL_memory_debug_header, my_perl);
-    return my_perl;
 #endif
```
ilmari: `INIT_TRACK_MEMPOOL` is defined as a no-op if `PERL_TRACK_MEMPOOL` is not defined, so the  `#ifdef` can go now that there's no other code inside it.

nwc10: Thanks. Added as a commit in the rebased branch.

I had *sort-of* noticed that there was something funny going on, as in the `perl_alloc_using` code `INIT_TRACK_MEMPOOL(...)` was used without any guard, but it flitted back out of my mind before I went to investigate what was really going on.

nwc10:
> LGTM. We'll want the same change to `perl_alloc_using`, which will be trivial.

Done in the rebased branch. I *think* that `PERL_IMPLICIT_SYS` is the default configuration on Win32, meaning that GH will build this code for us. But I'm not certain, and I don' think that I can test `PERL_IMPLICIT_SYS` on anything I have access to.

hvds:
> > LGTM. We'll want the same change to `perl_alloc_using`, which will be trivial.
> 
> Done in the rebased branch.

+1

> I _think_ that `PERL_IMPLICIT_SYS` is the default configuration on Win32, meaning that GH will build this code for us. But I'm not certain, and I don' think that I can test `PERL_IMPLICIT_SYS` on anything I have access to.

From this branch's [testsuite/Windows mscv142 (pull_request)](https://github.com/Perl/perl5/pull/19141/checks?check_run_id=3674080351):
```
1299 cl -c [...] -DPERL_IMPLICIT_SYS -Fo..\perl.obj ..\perl.c
```

(Along with huge numbers of warnings, but that's a problem for another day.)

nwc10:
Thanks for the confirmation. I didn't think to look in the build logs.

Are we good to rebase and merge the 3 commits to blead?

hvds:
> Are we good to rebase and merge the 3 commits to blead?

From my point of view, yes.

PR #19146, opened 2021-09-21
Move the HvAUX struct into the end of the hash's body

nwc10:
Memory for hash iterators *etc* is only allocated on demand, as most hashes are used as objects and so are never iterated.

The iterator state (and other extra data) is stored in a `struct xpvhv_aux`. In blead this is allocated as part of the block of memory used to hold the linked lists of `HE *` pointers - the hash's main array. As a consequence, the address of this struct can move if the hash grows, which adds complexity in quite a few places.

Instead, allocate the memory as part of the hash body - this way, the memory doesn't move once it is allocated.

Memory is still only allocated on demand - "upgrade" the body as needed, changing the existing code paths that (re)allocated `HvARRAY()` to instead swap the body.

Larger bodies can still be allocated from arenas - we can use the arena for `SVt_IV`. At one point this arena was used to allocate memory for pointer table hashes, until that allocation was moved into the pointer table code.

ASAN does not find this branch interesting (no regressions, threaded or unthreaded, with `-DPURIFY`)

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -2208,8 +2208,7 @@ S_hv_auxinit(pTHX_ HV *hv) {
             Renew(array, PERL_HV_ARRAY_ALLOC_BYTES(HvMAX(hv) + 1), char);
         }
         HvARRAY(hv) = (HE**)array;
-        SvOOK_on(hv);
-        iter = HvAUX(hv);
+        iter = Perl_hv_auxalloc(aTHX_ hv);
 #ifdef PERL_HASH_RANDOMIZE_KEYS
         if (PL_HASH_RAND_BITS_ENABLED) {
             /* mix in some new state to PL_hash_rand_bits to "randomize" the traversal order*/
--- a/hv.c
+++ b/hv.c
@@ -1954,7 +1954,6 @@ Perl_hfree_next_entry(pTHX_ HV *hv, STRLEN *indexp)
                 /* warning: at this point HvARRAY may have been
                  * re-allocated, HvMAX changed etc */
             }
-            iter = HvAUX(hv); /* may have been realloced */
             iter->xhv_riter = -1; 	/* HvRITER(hv) = -1 */
             iter->xhv_eiter = NULL;	/* HvEITER(hv) = NULL */
 #ifdef PERL_HASH_RANDOMIZE_KEYS
@@ -2262,7 +2261,6 @@ Perl_hv_iterinit(pTHX_ HV *hv)
             HvLAZYDEL_off(hv);
             hv_free_ent(hv, entry);
         }
-        iter = HvAUX(hv); /* may have been reallocated */
         iter->xhv_riter = -1; 	/* HvRITER(hv) = -1 */
         iter->xhv_eiter = NULL; /* HvEITER(hv) = NULL */
 #ifdef PERL_HASH_RANDOMIZE_KEYS
@@ -2378,7 +2376,6 @@ Perl_hv_name_set(pTHX_ HV *hv, const char *name, U32 len, U32 flags)
                 /* The first elem may be null. */
                 if(*this_name) unshare_hek_or_pvn(*this_name, 0, 0, 0);
                 Safefree(this_name);
-                iter = HvAUX(hv); /* may been realloced */
                 spot = &iter->xhv_name_u.xhvnameu_name;
                 iter->xhv_name_count = 0;
               }
@@ -2400,7 +2397,6 @@ Perl_hv_name_set(pTHX_ HV *hv, const char *name, U32 len, U32 flags)
             }
             else if (flags & HV_NAME_SETALL) {
                 unshare_hek_or_pvn(iter->xhv_name_u.xhvnameu_name, 0, 0, 0);
-                iter = HvAUX(hv); /* may been realloced */
                 spot = &iter->xhv_name_u.xhvnameu_name;
             }
             else {
@@ -2546,7 +2542,6 @@ Perl_hv_ename_delete(pTHX_ HV *hv, const char *name, U32 len, U32 flags)
                 : (HEK_LEN(*victim) == (I32)len && memEQ(HEK_KEY(*victim), name, len))
             ) {
                 unshare_hek_or_pvn(*victim, 0, 0, 0);
-                aux = HvAUX(hv); /* may been realloced */
                 if (count < 0) ++aux->xhv_name_count;
                 else --aux->xhv_name_count;
                 if (
@@ -2696,7 +2691,6 @@ Perl_hv_iternext_flags(pTHX_ HV *hv, I32 flags)
             SvREFCNT_dec(HeVAL(entry));
             Safefree(HeKEY_hek(entry));
             del_HE(entry);
-            iter = HvAUX(hv); /* may been realloced */
             iter->xhv_eiter = NULL; /* HvEITER(hv) = NULL */
             HvLAZYDEL_off(hv);
             return NULL;
@@ -2706,10 +2700,6 @@ Perl_hv_iternext_flags(pTHX_ HV *hv, I32 flags)
     if (!entry && SvRMAGICAL((const SV *)hv)
         && mg_find((const SV *)hv, PERL_MAGIC_env)) {
         prime_env_iter();
-        /* The prime_env_iter() on VMS just loaded up new hash values
-         * so HvARRAY() liked has been reallocated
-         */
-        iter = HvAUX(hv);
     }
 #endif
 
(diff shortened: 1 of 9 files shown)


hv.c
```diff
@@ -1460,30 +1451,6 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
...
-    /* before we zero the newly added memory, we
-     * need to deal with the aux struct that may be there
-     * or have been allocated by us*/
-    if (do_aux) {
-        struct xpvhv_aux *const dest = HvAUX(hv);
-        if (SvOOK(hv)) {
-            /* we reset the iterator's xhv_rand as well, so they get a totally new ordering */
-#ifdef PERL_HASH_RANDOMIZE_KEYS
-            dest->xhv_rand = (U32)PL_hash_rand_bits;
-#endif
```
hvds, line 1480: If I read things correctly, this was already a noop (since we've already set it on insert before deciding we need a split). If that's correct, it might be worth removing these 3 lines as a separate commit; if it's not, it probably needs mentioning in the commit message that we now no longer re-randomize on split.

sv.c
```diff
@@ -1417,12 +1414,15 @@ Perl_sv_upgrade(pTHX_ SV *const sv, svtype new_type)
         assert(new_type_details->body_size);
         /* We always allocated the full length item with PURIFY. To do this
            we fake things so that arena is false for all 16 types..  */
+#ifndef PURIFY
```
hvds, line 1429: It isn't clear to me from the commit message why this block and the one below newly become not-PURIFY only. Is that an additional bugfix, an additional optimization, or is it somehow newly required by the rest of this change?

sv.c
```diff
@@ -1496,6 +1507,42 @@ Perl_sv_upgrade(pTHX_ SV *const sv, svtype new_type)
...
+    void *new_body;
+
+    PERL_ARGS_ASSERT_HV_AUXALLOC;
+    assert(SvTYPE(hv) == SVt_PVHV);
+    assert(!SvOOK(hv));
+
+#ifdef PURIFY
+    new_body = new_NOARENAZ(&fake_hv_with_aux);
+#else
+    new_body_inline(new_body, SVt_IV, fake_hv_with_aux);
```
hvds: I think this could do with a comment to explain the `SVt_IV`. I suspect it somehow links with the renaming of that parameter from `svtype` to `root_index` (in the macro, but not the function), but I don't really have a clue. (And I'm not sure `new_body_inline` is still the most appropriate name for the macro.)

hv.c
nwc10: > If I read things correctly

Oh my. Yes, I've been going backwards and forwards for a while trying to be sure about this.

I think that you are correct that it's a no-op for an insert. (Well spotted. I'd not even noticed that this might be relevant.)

But. It doesn't seem to be a no-op where `S_hsplit` is called by `Perl_hv_ksplit`, which is `keys` as an LVALUE where the hash is expanded. So things like `perl -MDevel::Peek -lwe '%h = (1..4); Dump \%h; print scalar each %h; Dump \%h; keys %h = 20; Dump \%h; print scalar each %h; Dump \%h'`

In blead that resets the iterator *and* forces a new random ordering. In this branch, as a result of the commit above it doesn't. (The line `RAND = 0x......` in the rather verbose debug output)

I suspect that the correct thing is to split those 3 lines out as a separate commit that also inserts them into `Perl_hv_ksplit`, along with comments explaining what's going on, but I want to sleep on this idea.

hvds: > It doesn't seem to be a no-op where `S_hsplit` is called by `Perl_hv_ksplit`

Ah right, I hadn't considered lvalue keys().

> I suspect that the correct thing is to split those 3 lines out as a separate commit that also inserts them into `Perl_hv_ksplit`, along with comments explaining what's going on, but I want to sleep on this idea.

That makes sense, if seed reset is really required for this case. It isn't obvious to me that it _is_ required (other than for continuity with current behaviour) for a request to change the number of buckets that doesn't change the actual list of keys.

demerphq:
On Sun, 26 Sept 2021 at 20:58, Nicholas Clark ***@***.***>
wrote:

>
> In hv.c <https://github.com/Perl/perl5/pull/19146#discussion_r716246111>:
>
> > -#ifdef PERL_HASH_RANDOMIZE_KEYS
> -            dest->xhv_rand = (U32)PL_hash_rand_bits;
> -#endif
>
> If I read things correctly
>
> Oh my. Yes, I've been going backwards and forwards for a while trying to
> be sure about this.
>
> I think that you are correct that it's a no-op for an insert. (Well
> spotted. I'd not even noticed that this might be relevant.)
>
> But. It doesn't seem to be a no-op where S_hsplit is called by
> Perl_hv_ksplit, which is keys as an LVALUE where the hash is expanded. So
> things like perl -MDevel::Peek -lwe '%h = (1..4); Dump \%h; print scalar
> each %h; Dump \%h; keys %h = 20; Dump \%h; print scalar each %h; Dump \%h'
>
> In blead that resets the iterator *and* forces a new random ordering. In
> this branch, as a result of the commit above it doesn't. (The line RAND =
> 0x...... in the rather verbose debug output)
>
> I suspect that the correct thing is to split those 3 lines out as a
> separate commit that also inserts them into Perl_hv_ksplit, along with
> comments explaining what's going on, but I want to sleep on this idea.
>
The bucket chain randomization IMO is unnecessary with a good hash with a
128 bit seed like SipHash.

It is not even clear how much uncertainty it adds to the hashing.

consider we have something like this:

01234567
  A
  B

After a split we have 4 possibilities:

0123456789ABCDEF  # case 1, nothing moves, keyorder: AB
  A
  B

0123456789ABCDEF  # case 2, B moves, keyorder: AB
  A      B

0123456789ABCDEF  # case 3, A moves, keyorder: BA
  B      A

0123456789ABCDEF  # case 4, Both moved, keyorder: BA
         B
                  A

The bucket order randomization changes the last case so that it might also
produce:

0123456789ABCDEF  # case 4, Both moved, keyorder: AB
         A
                  B

The point is to make it harder to tell if a pair moved or not.  It is not
clear how well it works, with additional keys in the hash you can tease
apart the possibilities if you have an oracle hash function you can
consult. With this combined with the hash order traversal xor mask you need
more round trips to the oracle to tell if a pair of items collide. I added
this mostly because our hash function and seed allowed attacks on the seed,
which Siphash does not. I'd say that when SipHash is in use that logic
should be disabled. It should only be used in 32bit builds basically. With
a secure hash knowing that two keys collide shouldnt leak anything about
the seed.

cheers,
Yves

-- 

hv.c
nwc10: > That makes sense, if seed reset is really required for this case. It isn't obvious to me that it _is_ required (other than for continuity with current behaviour) for a request to change the number of buckets that doesn't change the actual list of keys.

Agree with your reasoning. But it was easy enough to keep the current behaviour, so (for now) I did, as e9dd222847f9

sv.c
nwc10: > It isn't clear to me from the commit message why this block and the one below newly become not-PURIFY only. Is that an additional bugfix, an additional optimization, or is it somehow newly required by the rest of this change?

It wasn't actually obvious to me either, coming back to this. (The branch is basically a group of cherry-picks of commits from early August). "newly required" is the answer, now documented in the commit message to 6d91a202e2c91c7b. As a side effect of you asking this, I spotted some more things I thought could be cleared up better, so they are also now in that commit.

sv.c
nwc10: > I think this could do with a comment to explain the `SVt_IV`. I suspect it somehow links with the renaming of that parameter from `svtype` to `root_index` (in the macro, but not the function), but I don't really have a clue. (And I'm not sure `new_body_inline` is still the most appropriate name for the macro.)

`SVt_IV` became `#  define HVAUX_ARENA_ROOT_IX   SVt_IV` in `sv.h` and more comments.

Yes, `new_body_inline` wasn't really the best name. It's now `new_body_from_arena` - if there's a better name still, it eludes me.

I also found some outdated comments in `sv.c` to eliminate.

sv.c
hvds: > "newly required" is the answer, now documented in the commit message to [6d91a20](https://github.com/Perl/perl5/commit/6d91a202e2c91c7bcafe475970b333db8863e7d7).

Cool, makes much more sense now thanks. In that commit message, though, `s/as/was/` in "the unreachable code as using".

hvds:
Apologies for going quiet on this; I did take a quick scan through the latest changes, and they look ok, but I haven't yet had time to go back through them more thoroughly. I hope to be able to make time for that in the next few days.

sv.c
nwc10: Well spotted. Fixed in the rebase that I just pushed

hvds:
LGTM, and apologies for the delay.

nwc10:
Thanks for all the useful feedback. I've also added delay by taking too long to respond to your suggestions.

PR #19150, opened 2021-09-22
Remove grandfather as a verb in Tie::Handle, Tie::Hash and Tie::Scalar

oalders:
It's not clear to me that legacy is actually the correct word here as the docs
seem to be expressing that what is provided is actually a fallback for classes
which don't implement their own new() methods. Would fallback be a better way
to describe this? I didn't go with this to start with because the wording
also seemed to imply that the new() methods were in place to support an older
API.

See https://github.com/Perl/perl5/issues/18830 for initial discussion.

### DIFF
--- a/lib/Tie/Handle.pm
+++ b/lib/Tie/Handle.pm
@@ -1,7 +1,7 @@
 package Tie::Handle;
 
 use 5.006_001;
-our $VERSION = '4.2';
+our $VERSION = '4.3';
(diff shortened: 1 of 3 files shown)


nwc10:
"Thanks applied"

(github doesn't seem to have figured out that there are "no conflicts" because there are 0 commits left to merge or cherry-pick)

Issue #19020, opened 2021-08-03
Under dash, Perl provides different results for $?


xsawyerx:
**Description**

Under `dash`, (certain?) signals in calls to `targetsh` produce different values for `$?`.

**Steps to Reproduce**

```sh
# /bin/sh -> dash
$ perl -e'`$^X -e'\''kill "KILL", \$\$'\''`; printf "Raw: $?, Exit: %s, signal: %s\n", ( $? >> 8 ), ( $? & 127 );'
Killed
Raw: 35072, Exit: 137, signal: 0

# /bin/sh -> bash
$ perl -e'`$^X -e'\''kill "KILL", \$\$'\''`; printf "Raw: $?, Exit: %s, signal: %s\n", ( $? >> 8 ), ( $? & 127 );'
Raw: 9, Exit: 0, signal: 9
```

**Long-form**

Perl's  backticks and `system()` call out to shell (specifically to targetsh[1]) when they receive an argument with an escaped variable (`"\$"`). (With `system()`, this only happens when there's only one argument.)

```sh
# Notice PID 24167
$ strace -f perl -e'system("ls \$foo");' 2>&1 | grep execve
execve("/usr/bin/perl", ["perl", "-esystem(\"ls \\$foo\");"], 0x7ffe89dee6c0 /* 33 vars */) = 0
[pid 24167] execve("/bin/sh", ["sh", "-c", "ls $foo"], 0x559db1baae80 /* 33 vars */) = 0
[pid 24168] execve("/bin/ls", ["ls"], 0x558771ffeb78 /* 33 vars */) = 0

# Notice PID 24376
$ strace -f perl -e'`ls\$foo`' 2>&1 | grep execve
execve("/usr/bin/perl", ["perl", "-e`ls\\$foo`"], 0x7ffcf69a68d0 /* 33 vars */) = 0
[pid 24376] execve("/bin/sh", ["sh", "-c", "ls$foo"], 0x55dcf524ce80 /* 33 vars */ <unfinished ...>
[pid 24376] <... execve resumed> )      = 0
[pid 24377] execve("/bin/ls", ["ls"], 0x55edd5930b50 /* 33 vars */) = 0
```

On Debian, `dash` is the default non-interactive shell[2] and on Ubuntu as well[3] and that's what `/bin/sh` (the targetsh) will point to.

[...]

PR #19152, opened 2021-09-23
Add exit code test (GH #19020):

xsawyerx:
This tests that when we kill a process through a shell, we get the correct exit code and signal.

Unfortunately, this fails on some versions of Ubuntu and `dash`, which I believe shows a bug somewhere. This PR **cannot be merged** until that is resolved one way (like fixing it in the core) or another (like turning this into a `TODO` or something). I'm hoping to get some smoke tests from this to provide more information.

There's a `sleep()` included because some systems might delay the signal until the `-e''` execution would already finish.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5916,6 +5916,7 @@ t/op/isa.t			See if isa works
 t/op/join.t			See if join works
 t/op/kill0.t			See if kill works
 t/op/kill0_child		Process tree script that is kill()ed
+t/op/kill9_shell_exit.t	Check that we get the right exit code on shell execution kill signal
 t/op/kvaslice.t			See if index/value array slices work
 t/op/kvhslice.t			See if key/value hash slices work
 t/op/lc.t			See if lc, uc, lcfirst, ucfirst, quotemeta work
--- a/t/op/kill9_shell_exit.t
+++ b/t/op/kill9_shell_exit.t
@@ -1,0 +1,20 @@
+#!./perl -w
+
+BEGIN {
+    chdir 't' if -d 't';
+    require './test.pl';
+    set_up_inc('../lib');
+}
+
+use strict;
+use warnings;
+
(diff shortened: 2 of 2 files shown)


xsawyerx:
Smoke test results also spot the issue:

```
# Failed test 2 - $? (35072) from a KILL signal is 9 at op/kill9_shell_exit.t line 18
t/op/kill9_shell_exit ............................................ FAILED at test 2
#      got "35072"
# expected "9"
# Failed test 3 - OS exit code (shifted) is 0 at op/kill9_shell_exit.t line 19
#      got "137"
# expected "0"
# Failed test 4 - KILL signal (bitwise ANDed exit) is 9 at op/kill9_shell_exit.t line 20
#      got "0"
# expected "9"
```

jkeenan:
There's been no activity in this p.r. since last September.

@xsawyerx, do you intend to pursue this p.r. further?

xsawyerx:
I'm not sure what's left to pursue. This ticket was left unattended and the discussion on the issue itself seems to have petered off.

* There is a bug in Debian's `dash` (which seems like a purposeful change with unintended consequences)
* It was also inherited by Ubuntu
* This bug affects Perl users in some narrow situations
* That narrow situation includes the perl smokers
* I suggested a few options - the most (and maybe only) practical of which is to document this (should probably be in `perlport` and under `system` or backticks in `perlfunc`) and move on
* No additional comments were made on the discussion or on this test PR

This PR includes the test to prove the problem. I would suggest this be added as TODO so it wouldn't break the testing suite but developers would still be notified when this issue is no longer relevant (namely, when the affected `dash` version is not in circulation as much).

I have no intention of pursuing either this PR or the original ticket I opened on it. If no one wishes to act on it (or resolve the discussion on it), you can close it as "ENOFIX".

demerphq:
Anyone want to pick up turning this into a set of TODO tests?

PR #19153, opened 2021-09-26
Add CI tests for `make distclean` and Porting/manicheck

nwc10:
Also leave the checkout `fetch-depth` at its default (1) for most jobs.

### DIFF
--- a/Porting/manicheck
+++ b/Porting/manicheck
@@ -4,7 +4,7 @@
 #  a) files listed in MANIFEST which don't exist
 #  b) files which exist but which aren't in MANIFEST
 
-use strict;
+use v5.14;
 use warnings;
 use File::Find;
 
@@ -17,12 +17,13 @@ for (@files) {
 my %files = map { $_ => 1 } @files;
 find {
     wanted => sub {
-        my $x = $File::Find::name; $x =~ s/^..//;
         return if -d;
         return if $_ eq '.mailmap';
         return if $_ eq '.gitignore';
         return if $_ eq '.gitattributes';
         return if $_ eq '.git_patch';
+
+        my $x = $File::Find::name =~ s!^\./!!r;
         return if $x =~ /^\.git\b/;
         return if $x =~ m{^\.github/};
         print "$x\t\tnot in MANIFEST\n" if !$files{$x};
--- a/Porting/manicheck
+++ b/Porting/manicheck
@@ -7,14 +7,27 @@
 use v5.14;
 use warnings;
 use File::Find;
+use Getopt::Long;
+use constant SKIP => 125;
+
+my $exitstatus;
+GetOptions('exitstatus!', \$exitstatus)
+    or die "$0 [--exitstatus]";
+
+my %files;
+my $missing = 0;
+my $bonus = 0;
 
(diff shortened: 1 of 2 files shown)


nwc10:
@wolfsage - I wasn't sure whether some/all of the steps in the RMG about "check distclean" could now be removed, so didn't change the text. But as this gives us a CI test that will fail if `make distclean` and `perl Porting/manicheck` disagree, perhaps they can go?

Porting/manicheck
```diff
@@ -4,28 +4,57 @@
...
 open my $fh, '<', 'MANIFEST' or die "Can't read MANIFEST: $!\n";
-my @files = map { (split)[0] } <$fh>;
-close $fh;
-for (@files) {
-    print "$_ from MANIFEST doesn't exist\n" if ! -f;
+for my $line (<$fh>) {
+    my ($file) = $line =~ /^(\S+)/;
+    ++$files{$file};
+    next
+        if -f $file;
```
jkeenan: Why is the `next` statement broken into two lines?

nwc10: > Why is the `next` statement broken into two lines?

Because I was writing it in the style i prefer. Which is inconsistent with the rest of the file.

Oops, well spotted. Thanks for catching this.

Fixed in the rebase.

wolfsage:
> @wolfsage - I wasn't sure whether some/all of the steps in the RMG about "check distclean" could now be removed, so didn't change the text. But as this gives us a CI test that will fail if `make distclean` and `perl Porting/manicheck` disagree, perhaps they can go?

For point releases, we don't (ever?) add any new files, so maybe. But what about new releases? (A lot of the releasing happens locally without pushing, and we don't push until after we've uploaded a tarball to cpan, so it could be too late?)

The second distclean check is for the built tarball, so maybe that one should stay regardless? I'm not sure

nwc10:
> > @wolfsage - I wasn't sure whether some/all of the steps in the RMG about "check distclean" could now be removed, so didn't change the text. But as this gives us a CI test that will fail if `make distclean` and `perl Porting/manicheck` disagree, perhaps they can go?
> 
> For point releases, we don't (ever?) add any new files, so maybe. But what about new releases? (A lot of the releasing happens locally without pushing, and we don't push until after we've uploaded a tarball to cpan, so it could be too late?)

I doubt that we add new **installed** files to a maint point release, but we might add new test scripts as a consequence of backporting a commit that fixes a bug. (And if I can think of that case, there are probably other cases I can't think of)

> The second distclean check is for the built tarball, so maybe that one should stay regardless? I'm not sure

If it's for the built tarball, it's testing something we don't/can't test in a git checkout. "More" obviously it would fail if the tarball fails to include a file mentioned in `MANIFEST` (which might be possible without the build itself failing). I think we can build things differently if we're not in a git checkout, and that might mean we create different files, so we do need a check that our `distclean` rules also get these.

Ultimately, I think the answer is "don't remove the steps". It's useful to keep testing them, but the CI tests here mean that they shouldn't ever actually fail again, which saves future release managers the distraction that you got surprised with.

I think that this branch now better than what we had, so is good to merge. Unless someone (myself included) spots something daft I'll merge it if/when the CI run completes.

wolfsage:
Great, thanks!

PR #19159, opened 2021-09-29
Remove NetWare support

ilmari:
The build has been broken since 2009 without anyone noticing.


toke.c
```diff
@@ -12347,37 +12341,18 @@ Perl_yyerror_pvn(pTHX_ const char *const s, STRLEN len, U32 flags)
...
-                    Only for NetWare:
-                    The code below is removed for NetWare because it
-                    abends/crashes on NetWare when the script has error such as
-                    not having the closing quotes like:
-                        if ($var eq "value)
-                    Checking of white spaces is anyway done in NetWare code.
-            */
-#ifndef NETWARE
+                && PL_oldbufptr != PL_bufptr)
+        {
```
nwc10, line 12427: Did you mean to split this onto two lines?

It seems to have confused `git diff`, and hence me, because I see the diff as

```
         else if (  PL_oldbufptr
                 && PL_bufptr > PL_oldbufptr
                 && PL_bufptr - PL_oldbufptr < 200
-                && PL_oldbufptr != PL_bufptr) {
-            /*
-                    Only for NetWare:
-                    The code below is removed for NetWare because it
-                    abends/crashes on NetWare when the script has error such as
-                    not having the closing quotes like:
-                        if ($var eq "value)
-                    Checking of white spaces is anyway done in NetWare code.
-            */
-#ifndef NETWARE
+                && PL_oldbufptr != PL_bufptr)
+        {
             while (isSPACE(*PL_oldbufptr))
                 PL_oldbufptr++;
-#endif
             context = PL_oldbufptr;
             contlen = PL_bufptr - PL_oldbufptr;
         }
```

which has the `+` and `-` interleaved, whereas I *think* the change is really just `-` `-` `-`

nwc10 (commented):
Oh my, that's a lot of code deleted.

I couldn't spot anything daft looking, apart from that part that I commented on, that confused me.

pod/perldelta.pod
```diff
@@ -315,9 +315,11 @@ XXX List any platforms that this version of perl no longer compiles on.
 
 =over 4
 
-=item XXX-some-platform
+=item NetWare
 
-XXX
+Support code for Novell NetWare has been removed.  NetWare was a
+server operating system by Novell.  The port was last updated in July
+2002, and the platform itself in May 2009.
```
nwc10: I realise after hitting "finish" that, *if* the other comment means an `--amend` is needed, then maybe also tweak this text to add that unrelated changes accidentally broke the build for the existing port in Sept 2009, and in 12 years no-one has reported this.

(Otherwise we tweak the text in its own commit)

toke.c
ilmari: The [immediately preceding `elsif`](https://github.com/Perl/perl5/blob/ccb35e8ab9be0d7157065f2583a8e312387e648a/toke.c#L12338-L12348) and [subsequent `if`s](https://github.com/Perl/perl5/blob/ccb35e8ab9be0d7157065f2583a8e312387e648a/toke.c#L12391-L12399) with multi-line conditions have the `{` on a separate line from the `)`, so I took the opportunity to make this block consistent with those.

nwc10: Aha yes. Makes sense.

rjbs:
Putting aside any code review issues, this is okay to merge, per unanimous PSC agreement.

pod/perldelta.pod
ilmari: Because it needed rebasing anyway before merging, I amended the commit to add this text.

PR #19160, opened 2021-09-29
Autogenerate gitattributes

Corion:
Automatically generate the `.gitattributes` file from information gleaned from the programs in `regen/` . Add the necessary information to programs in `regen/` that lack it.


jkeenan:
@Corion, would you like to resolve merge conflicts so that this pull request can move forward?

Also, I think it would be helpful to include an explanation of what purpose a `.gitattributes` file actually serves.  (I myself have never consciously had to create one or reference one.)

Thank you very much.
Jim Keenan

Grinnz:
gitattributes applies arbitrary attributes to files or file globs in the git repo. In this case we are using export-subst causes .git_patch to be evaluated as a template when the repository tree is exported; and the linguist-generated tag tells github's linguist that the files are generated, which it can use to unclutter PR diffs.

khwilliamson:
What should be done with this pull request?

jkeenan:
> What should be done with this pull request?

The original poster did not respond to my query of several months ago, and the ticket has accumulated merge conflicts multiple times.  It should be closed.

Corion:
I think it should be applied some time in the next release cycle. I will rebase it once more, and add a link to the gitattributes manpage ( https://git-scm.com/docs/gitattributes ) to the commit.

I'm not happy with having to rebase it all the time, so if it doesn't get applied in the next release cycle, somebody else can rebase it. Any manual change to .gitattributes makes this PR not apply cleanly.

PR #19166, opened 2021-10-03
Provide SAVESTRLEN() macro

### LINKED ISSUE #19165: SAVESTRLEN() is missing
Many of the savestack types have associated public macros, e.g. SAVEINT, SAVEVPTR, etc...

SAVESTRLEN appears to be lacking.

Core internal code currently makes use of a (lowercased) `save_strlen()` macro that isn't made public.

I suggest renaming it to capitals without the `_`, `SAVESTRLEN`, and making it public.

(I'll probably send a PR to do this tomorrow)

leonerd:
Fixes #19165

Leaving commits separate for now for ease of review; I'll squash them into one before merging.


tonycoz (approved):
Looks good to me.

Issue #17867, opened 2020-06-16
[doc] "Allocation too large" in perldiag


epa:
perldiag describes an error

- Allocation too large: %x
(X) You can't allocate more than 64K on an MS-DOS machine.

But the old 16-bit MS-DOS port was discontinued a long time ago.  A 32-bit build with DJGPP does not have the 64K limitation.  The code raising the error was removed in commit 39c0d7ee897c13cceed4e2b5712cd4e88315336a but it wasn't removed from the documentation.

jkeenan:
@xsawyerx, could you provide an update on the status of this ticket?

(I don't understand why there's a "do not merge" label on this.  Also, we're past the 5.33.1 milestone.)

Thank you very much.
Jim Keenan

jkeenan:
Pull request filed in https://github.com/Perl/perl5/pull/19175

richardleach:
Looks like this was resolved in https://github.com/Perl/perl5/pull/19175.

PR #19175, opened 2021-10-07
Remove documentation of previously removed warning

jkeenan:
Per https://github.com/Perl/perl5/issues/17867, the warning itself was
removed in February 2013.

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -63,10 +63,6 @@ removed in a future Perl version:
     use feature "refaliasing";
     \$x = \$y;
 
-=item Allocation too large: %x
-
-(X) You can't allocate more than 64K on an MS-DOS machine.
-
 =item '%c' allowed only after types %s in %s
 
 (F) The modifiers '!', '<' and '>' are allowed in pack() or unpack() only


Grinnz:
Is it standard practice to remove these entries from perldiag? People may wonder why they can't find an entry for a warning they get on a previous version of Perl, but I would understand a policy of keeping them to those present in the current version.

Grinnz:
Though also, this is present in a pretty old and (I think?) no longer supported architecture of Perl which is another point toward removing it.

jkeenan:
On 10/7/21 7:59 PM, Dan Book wrote:
> Though also, this is present in a pretty old and (I think?) no longer 
> supported architecture of Perl which is another point toward removing it.
> 

Yes, which is why the warning was removed eight years ago.

I can't imagine that there are a non-zero number of people on this 
planet who are making decisions based on this warning.  No one has 
complained about its absence in all these years.  It was simply an 
oversight that the documentation of the warning was not removed when it 
should have been.

jkeenan:
@Grinnz, do you still have concerns about this documentation change?

Does anyone else object?

Thank you very much.
Jim Keenan

Grinnz:
I have no concerns in this particular case.

PR #19179, opened 2021-10-09
Send bug reports to GitHub

jkeenan:
Mail to 'perlbug@perl.org' now simply triggers a response redirecting
sender there.

### DIFF
--- a/dist/IO/lib/IO/Socket.pm
+++ b/dist/IO/lib/IO/Socket.pm
@@ -23,7 +23,7 @@ require IO::Socket::UNIX if ($^O ne 'epoc' && $^O ne 'symbian');
 
 our @ISA = qw(IO::Handle);
 
-our $VERSION = "1.48";
+our $VERSION = "1.49";
 
 our @EXPORT_OK = qw(sockatmark);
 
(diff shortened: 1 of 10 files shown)


dist/IO/lib/IO/Socket.pm
```diff
@@ -916,8 +916,8 @@ L<IO::Socket::IP>
 
 =head1 AUTHOR
 
-Graham Barr.  atmark() by Lincoln Stein.  Currently maintained by the
-Perl Porters.  Please report all bugs to <perlbug@perl.org>.
+Graham Barr. atmark() by Lincoln Stein.  Currently maintained by the Perl
```
rjbs:
```suggestion
Graham Barr.  atmark() by Lincoln Stein.  Currently maintained by the Perl
```

Extreme nitpick:  you've eliminate one double-space-after-full-stop in this paragraph, but not the others.  Let's just leave that be?

rjbs (approved):
I have provided only one very very small comment that needn't stand in the way — but would be easy to apply.  Looks good.

dist/IO/lib/IO/Socket.pm
jkeenan: Accepted.

jkeenan:
Updated with one other 2-character correction.

toddr (approved):

toddr:
I already point the cpan module at p5p so this is perfect. Thanks @jkeenan

PR #19181, opened 2021-10-09
Send bugs to GitHub

jkeenan:
Do not advise sending mail to 'perlbug@perl.org', as that now simply
triggers a response redirecting sender to GitHub.


rjbs (approved):
Seems good to me.  (I wonder what else uses cf_email!  Nothing shocking, upon quick review...)

PR #19191, opened 2021-10-15
Fix Memoize tests, where GDBM_File is involved

graygnuorg:
The tie_gdbm.t test uses O_RDWR | O_CREAT as the flags argument
when tying to GDBM_File.  However, GDBM_File requires this argument
to be a valid combination of GDBM_* constants.  Incidentally,
O_RDWR|O_CREAT = 0x42 = GDBM_NOLOCK|GDBM_WRCREAT, so the test passes,
albeit the use of 0x40 (GDBM_NOLOCK) bit is risky.  Change this to
GDBM_NEWDB, which suits this case best.

In errors.t, tests 4--8, constant value 2 (GDBM_WRCREAT) is used
instead.  Change it to use symbolic value, instead of
a hardcoded constant (again, GDBM_NEWDB).

### DIFF
--- a/cpan/Memoize/t/errors.t
+++ b/cpan/Memoize/t/errors.t
@@ -21,7 +21,7 @@ $n = 4;
 my $dummyfile = './dummydb';
 use Fcntl;
 my %args = ( DB_File => [],
-             GDBM_File => [$dummyfile, 2, 0666],
+             GDBM_File => [$dummyfile, \&GDBM_File::GDBM_NEWDB, 0666],
              ODBM_File => [$dummyfile, O_RDWR|O_CREAT, 0666],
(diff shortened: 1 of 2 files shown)


tonycoz:
in theory this should have been applied upstream first, but Memoize hasn't had a release since 2012, and already has a few customizations in core.

I've opened https://rt.cpan.org/Ticket/Display.html?id=139736 upstream

xenu:
>GDBM_File requires this argument to be a valid combination of GDBM_* constants. 

If GDBM_File requires those constants, they should be documented.

tonycoz:
> If GDBM_File requires those constants, they should be documented.

We could duplicate and adapt the gdbm documentation, but the GDBM_File documentation already refers the reader to the gdbm documentation.

In any case, an issue for this should be separate if it's needed.

PR #19194, opened 2021-10-15
Remove old MSVC++ (pre-VC12) support from Windows Makefiles

steve-m-hay:
Also remove the Platform SDK 2003 SP1/R2 64-bit compiler since (having just
tried it) it no longer builds perl because it doesn't allow mixed code and
declarations.

Also remove mention of the Windows SDK since it doesn't appear to contain
the compiler or linker any more. This page says that the VC++ 2010 compiler
and linker were included in the Windows 7 SDK but then makes no mention of
it for the Windows 8 SDK:
https://docs.microsoft.com/en-us/previous-versions/visualstudio/windows-sdk/ff660763(v=vs.110)
The page notes that it is no longer being updated, but the whole of the
C:\Program Files (x86)\Windows Kits folder on my system (which contains
8.0, 8.10 and 10) does not contain any cl.exe or link.exe.

I've re-worded some of the if/else conditions in the makefiles to pick out
earlier versions rather than later versions. Thus, the current versions
become the "else" case, which means that the next version of Visual C++ is
more likely to work without requiring any changes.

### DIFF
--- a/README.win32
+++ b/README.win32
@@ -63,9 +63,8 @@ that are also supported by perl's makefile.
 =back
 
 The Microsoft Visual C++ compilers are also now being given away free. They
-are available as "Visual C++ 2013-2019 Community Edition" (and also as part of
-the ".NET Framework SDK") and are the same compilers that ship with "Visual
-C++ 2013-2019 Professional".
+are available as "Visual C++ 2013-2019 Community Edition" and are the same
+compilers that ship with "Visual C++ 2013-2019 Professional".
 
 Visual C++ 2013 is capable of B<targeting> XP and Windows Server 2003 but the
 build host requirement is Windows 7/Windows Server 2012. For more details see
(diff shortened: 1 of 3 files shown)


xenu (approved):

tonycoz (approved):

nwc10 (approved):
> Also remove the Platform SDK 2003 SP1/R2 64-bit compiler since (having just tried it) it no longer builds perl because it doesn't allow mixed code and declarations.

Thanks for testing this. I wouldn't have even realised that it needed testing.

> I've re-worded some of the if/else conditions in the makefiles to pick out earlier versions rather than later versions. Thus, the current versions become the "else" case, which means that the next version of Visual C++ is more likely to work without requiring any changes.

I'm not in a position to test any of this, but I looked carefully at all the changes you made, and as best I can tell you didn't make any mistakes.

Issue #9993, opened 2009-11-30
-Dmad: double free or corruption


p5pRT:
Migrated from [rt.perl.org#70934](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=70934) (status was 'resolved')

Searchable as RT70934$

p5pRT:
### From frank.wiegand@gmail.com


This is a bug report for perl from fw@&#8203;hal2\.\(none\)\,
generated with the help of perlbug 1\.39 running under perl 5\.11\.2\.




---
If your perl has \-Dmad\, the following program crashes&#8203;:

$ bleadperl \-we '$x="x" x 257; eval "for $x"'                                                
\*\*\* glibc detected \*\*\* bleadperl&#8203;: double free or corruption \(\!prev\)&#8203;: 0x0000000001dca670 \*\*\*                      
======= Backtrace&#8203;: =========                                                                                     
/lib/libc\.so\.6\[0x7f26dc388db6\]                                                                                   
/lib/libc\.so\.6\(cfree\+0x6c\)\[0x7f26dc38d6fc\]                                                                       
bleadperl\(Perl\_safesysfree\+0x89\)\[0x50d449\]                                                                       
bleadperl\(Perl\_sv\_clear\+0x204d\)\[0x597475\]                                                                        
bleadperl\(Perl\_sv\_free2\+0xa7\)\[0x597746\]                                                                          
bleadperl\(Perl\_parser\_free\+0x8d\)\[0x4733d6\]                                                                       
bleadperl\(Perl\_leave\_scope\+0x363f\)\[0x605bac\]                                                                     
bleadperl\(Perl\_pop\_scope\+0x46\)\[0x5ff4c7\]                                                                         
bleadperl\(Perl\_die\_where\+0x24eb\)\[0x616608\]                                                                       
bleadperl\(Perl\_croak\+0x0\)\[0x5137d6\]                                                                              
bleadperl\(Perl\_vwarn\+0x0\)\[0x5138b2\]                                                                              
bleadperl\(Perl\_yylex\+0x1298c\)\[0x494fac\]                                                                          
bleadperl\(Perl\_yyparse\+0x29e\)\[0x4b6647\]                                                                          
bleadperl\[0x624ca7\]                                                                                              
bleadperl\(Perl\_pp\_entereval\+0xf0f\)\[0x62c43d\]                                                                     
bleadperl\(Perl\_runops\_debug\+0x153\)\[0x5064a0\]                                                                     
[...]

PR #19197, opened 2021-10-15
`for my($k, $v) (%hash)` should not be a syntax error

### LINKED ISSUE #19192: Parsing issue with Mutlivariable for loops
**Description**

The new multi-variable for syntax requires a space after `my` in the for loop

**Steps to Reproduce**
i.e. This is valid:
```perl
for my ($k, $v) (%hash) {
 ...
}
```

This gives an error:
```perl
for my($k, $v) (%hash) {
  ...
}
```
```ERROR: Missing $ on loop variable at (IRC) line 1.```

**Expected behavior**

**Perl configuration**

```
Summary of my perl5 (revision 5 version 35 subversion 5) configuration:

  Platform:
    osname=linux
    osvers=5.10.0-0.bpo.8-amd64
    archname=x86_64-linux
    uname='linux simcop2387.info 5.10.0-0.bpo.8-amd64 #1 smp debian 5.10.46-4~bpo10+1 (2021-08-07) x86_64 gnulinux '
    config_args='-Dprefix=/home/perlbot/perl5/custom/blead-2021-10-15 -de -Dusedevel -Duseshrplib -Dusemymalloc=y -Accflags=-fpie -fPIC -mtune=native -fstack-protector-all -pie -D_FORTIFY_SOURCE=2 -ggdb  -DPERL_EMERGENCY_SBRK -Aldflags=-Wl,-z,now -Wl,-zrelro -Wl,-z,noexecstack -A'eval:scriptdir=/home/perlbot/perl5/custom/blead-2021-10-15/bin''
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef 
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=y
  Compiler:
    cc='cc'
    ccflags ='-fpie -fPIC -mtune=native -fstack-protector-all -pie -D_FORTIFY_SOURCE=2 -ggdb -DPERL_EMERGENCY_SBRK -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-fpie -fPIC -mtune=native -fstack-protector-all -pie -D_FORTIFY_SOURCE=2 -ggdb -DPERL_EMERGENCY_SBRK -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='8.3.0'
    gccosandvers=''   
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define 
    longlongsize=8
    d_longdbl=define  
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'   
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define  
  Linker and Libraries:
    ld='cc'
    ldflags =' -Wl,-z,now -Wl,-zrelro -Wl,-z,noexecstack -fstack-protector-strong -L/usr/local/lib'
    libc=libc-2.28.so 
    so=so
    useshrplib=true   
    libperl=libperl.so
    gnulibc_version='2.28'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef   
    ccdlflags='-Wl,-E -Wl,-rpath,/home/perlbot/perl5/custom/blead-2021-10-15/lib/5.35.5/x86_64-linux/CORE'
    cccdlflags='-fPIC'

  Compile-time options:
    HAS_TIMES
    MYMALLOC
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_USE_DEVEL
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under linux
  Compiled at Oct 15 2021 10:28:07
  %ENV:
    PERLBREW_HOME="/home/perlbot/.perlbrew"
    PERLBREW_MANPATH="/home/perlbot/perl5/perlbrew/perls/perlbot-main/man"
    PERLBREW_PATH="/home/perlbot/perl5/perlbrew/bin:/home/perlbot/perl5/perlbrew/perls/perlbot-main/bin"
    PERLBREW_PERL="perlbot-main"
    PERLBREW_ROOT="/home/perlbot/perl5/perlbrew"
    PERLBREW_SHELLRC_VERSION="0.86"
    PERLBREW_VERSION="0.86"
  @INC:
    /home/perlbot/perl5/custom/blead-2021-10-15/lib/site_perl/5.35.5/x86_64-linux
    /home/perlbot/perl5/custom/blead-2021-10-15/lib/site_perl/5.35.5
    /home/perlbot/perl5/custom/blead-2021-10-15/lib/5.35.5/x86_64-linux
    /home/perlbot/perl5/custom/blead-2021-10-15/lib/5.35.5
```

### PR DESCRIPTION
**Pull request by nwc10**
No-one had thought to test for this explicitly. After all, both
`for my $foo ...` and `for my$foo ...` are valid syntax, so tweaking the
C lexer code to add an optional '(' should work, surely?

The problem was that, *as was*, the lexer code didn't "accept" those two
syntax variants the way the comments would suggest.

`for my $foo ...` was treated as

1) we saw 'my '
2) we saw the dollar sign
3) success!

but `for my$foo ...` was treated as

0) we didn't see 'my ' or 'our '
1) we saw the literal string 'my' which is valid as a package name
2) we saw the dollar sign
3) success!

ie some sort of mangled variant of `for my Dog $spot ...` without 'my'

*but* as the lexer was happy with what it saw, it returned that the input
stream was valid for a "for" token, and control continues to the grammar.
The grammar, of course, didn't make these mistakes, so parsed everything
properly and built the correct optree.

(And, if presented with `for Dog $spot (...)` the grammar wouldn't match,
so all invalid code was correctly rejected)

However, all this came unstuck with `for my($k` because that didn't
mis-tokenise as some crazy package name 'my(', so it reported a syntax error.

Hence rewrite yyl_foreach() to actually tokenise everything correctly.

"Correctly", apart from how scan_word() needs to be called even when we
know we saw neither "my" nor "our', so that "Identifier too long" is reported
instead of "syntax error".

Fixes #19192

### DIFF
--- a/t/comp/parser.t
+++ b/t/comp/parser.t
@@ -8,7 +8,7 @@ BEGIN {
     chdir 't' if -d 't';
 }
 
-print "1..190\n";
+print "1..191\n";
 
 sub failed {
     my ($got, $expected, $name) = @_;
@@ -334,15 +334,29 @@ like($@, qr/BEGIN failed--compilation aborted/, 'BEGIN 7' );
   like($@, qr/Identifier too long/, "too long id in glob ctx");
 
   eval qq[ for $xFC ];
-  like($@, qr/Missing \$ on loop variable/,
+  like($@, qr/^Missing \$ on loop variable /,
        "252 char id ok, but a different error");
   eval qq[ for $xFD; ];
-  like($@, qr/Identifier too long/, "too long id in for ctx");
+  like($@, qr/^Missing \$ on loop variable /, "too long id in for ctx");
 
   # the specific case from the ticket
+  # however the parsing code in yll_foreach has now changed
   my $x = "x" x 257;
   eval qq[ for $x ];
-  like($@, qr/Identifier too long/, "too long id ticket case");
+  like($@, qr/^Missing \$ on loop variable /, "too long id ticket case");
+
+  # as PL_tokenbuf is now PL_parser->tokenbuf, the "buffer overflow" that was
+  # reported in GH #9993 now corrupts some other part of the parser structure.
+  # Currently, that seems to be the line number. Hence this test will fail if
+  # the fix from commit 0b3da58dfdc35079 is reversed. (However, as the later
+  # commit 61bc22580524a6d9 changed the code (now) in yyl_foreach() from
+  # scan_ident() to scan_word(), to recreate the problem one needs to apply
+  # the buggy change to the calculation of the variable `e` in scan_word()
+  # instead.
+
+  my $x = "x" x 260;
+  eval qq[ for my $x \$foo ];
+  like($@, qr/at \(eval \d+\) line 1[,.]/, "line number is reported correctly");
 }
 
 {
--- a/toke.c
+++ b/toke.c
@@ -6931,31 +6931,67 @@ yyl_foreach(pTHX_ char *s)
     if (PL_expect == XSTATE && isIDFIRST_lazy_if_safe(s, PL_bufend, UTF)) {
         char *p = s;
         SSize_t s_off = s - SvPVX(PL_linestr);
-        STRLEN len;
-        bool permit_paren = FALSE;
+        bool paren_is_valid = FALSE;
+        bool maybe_package = FALSE;
+        bool saw_core = FALSE;
+        bool core_valid = FALSE;
 
-        if (memBEGINPs(p, (STRLEN) (PL_bufend - p), "my") && isSPACE(p[2])) {
-            p += 2;
-            permit_paren = TRUE;
+        if (UNLIKELY(memBEGINPs(p, (STRLEN) (PL_bufend - p), "CORE::"))) {
+            saw_core = TRUE;
+            p += 6;
         }
-        else if (memBEGINPs(p, (STRLEN) (PL_bufend - p), "our") && isSPACE(p[3])) {
-            p += 3;
-        }
-
-        p = skipspace(p);
-        /* skip optional package name, as in "for my abc $x (..)" */
-        if (isIDFIRST_lazy_if_safe(p, PL_bufend, UTF)) {
-            permit_paren = FALSE;
-            p = scan_word(p, PL_tokenbuf, sizeof PL_tokenbuf, TRUE, &len);
-            p = skipspace(p);
-        }
-        if (*p != '$' && *p != '\\') {
-            if (permit_paren && *p == '(') {
-                Perl_ck_warner_d(aTHX_
-                                 packWARN(WARN_EXPERIMENTAL__FOR_LIST), "for my (...) is experimental");
-            } else {
-                Perl_croak(aTHX_ "Missing $ on loop variable");
+        if (LIKELY(memBEGINPs(p, (STRLEN) (PL_bufend - p), "my"))) {
+            core_valid = TRUE;
+            paren_is_valid = TRUE;
+            if (isSPACE(p[2])) {
+                p = skipspace(p + 3);
+                maybe_package = TRUE;
             }
+            else {
+                p += 2;
+            }
+        }
+        else if (memBEGINPs(p, (STRLEN) (PL_bufend - p), "our")) {
+            core_valid = TRUE;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: toke.c [hunk 1]
```diff
@@ -6931,29 +6931,67 @@ yyl_foreach(pTHX_ char *s)
...
-        STRLEN len;
-        bool permit_paren = FALSE;
+        /* This might *seem* backwards - surely it can only be a package if
+         * first we find "my" or "our"?
+         * But...
+         * There are tests for `for Reeeeeeeeally_Long_Thing {}`
+         * that actually rely on the call to scan_word happening here in
+         * yyl_foreach and hence croaking with "Identifier too long" *before*
+         * we (instead) croak about the missing  $.
+         * This is nuts. But we test it this way. Go figure. */
```
tonycoz: I wrote that test long ago to match the specific crash case reported in #9993, I don't think it's necessary to make the code convoluted just to keep a test that is no longer applicable.  As long as the incorrect code still errors without producing a buffer overflow, and correct code compiles I think it's all good.

nwc10: > As long as the incorrect code still errors without producing a buffer overflow, and correct code compiles I think it's all good.

Thanks for the insight into why the test is the way that it is.

So I tried to verify that I can still make the test trigger a buffer overflow if I revert the C code change to `scan_ident`, and I can't (directly) because the call was changed to `scan_word` in 2013 by commit 61bc22580524a6d92bf31b2046baa6dcc4e400d4:

```
    Correctly parse class name in 'for my class $foo

    The code in toke.c on encountering 'for my' or for our',
    skips past any optional package name, then checks that the next thing is a
    '$', so that it can give a "Missing $ on loop variable" error if
    necessary.

    However, the code to skip the package name was using scan_ident() rather
    than scan_word(); the latter assumes that the first char was the sigil,
    and starts scanning from the second char onwards. So in something like

        for my a1b $x (...)

    scan_ident() starts scanning at '1b' in 'a1b', thinks it scanning $1 or
    similar, and stops at the first non-digit char, in this case the b.

    The fix is to use parse_word() instead.
```

(which also tweaked the lengths in the regression tests to maintain behaviour.)

I can't reproduce any buffer overflow, even if attempt to reintroduce the original (botched refactoring) bug into `scan_word` instead.

So I think I'm going to re-write the code in `yyl_foreach` to parse only valid syntax, and then "adjust the tests to taste".

nwc10: Oh my.

Done. (Well, "proposed")

But what I found on the way...

t/comp/parser.t
```diff
@@ -334,15 +334,29 @@ like($@, qr/BEGIN failed--compilation aborted/, 'BEGIN 7' );
...
   eval qq[ for $xFC ];
-  like($@, qr/Missing \$ on loop variable/,
+  like($@, qr/^Missing \$ on loop variable /,
        "252 char id ok, but a different error");
   eval qq[ for $xFD; ];
-  like($@, qr/Identifier too long/, "too long id in for ctx");
+  like($@, qr/^Missing \$ on loop variable /, "too long id in for ctx");
 
   # the specific case from the ticket
+  # however the parsing code in yll_foreach has now changed
```
tonycoz, line 343: s/yll_foreach/yyl_foreach/

tonycoz:
Looks good to me except for the typo

PR #19200, opened 2021-10-18
don't overwrite the faked up type details for hv-with-aux

tonycoz:
Detected by coverity as:

340472 Unused value

An assigned value that is never used may represent unnecessary computation, an incorrect algorithm, or possibly the need for cleanup or refactoring.

In Perl_sv_clear: A value assigned to a variable is never used. (CWE-563)

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -6987,8 +6987,8 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
         }
         else {
             arena_index = type;
+            sv_type_details = bodies_by_type + arena_index;
         }
-        sv_type_details = bodies_by_type + arena_index;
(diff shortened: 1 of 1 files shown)


nwc10:
Oops. "Thanks applied".

That would have been "leaking" pointers to bodies allocated from arenas, but would never have shown up under any leak checker. With `-DPURIFY` arenas aren't used, so everything is just `malloc`/`free`, and so bodies are passed to `free` without "thinking too hard" about what the "body details" metadata says to do. Without `-DPURIFY` bodies would themselves leak, but the arena doesn't leak, so that would be correctly freed at interpreter full destruction, and nothing would have realised.

PR #19206, opened 2021-10-20
_BitScanForward64 is available only on 64-bit architectures

### LINKED ISSUE #19205: MSVC++ 32-bit build is broken
The MSVC++ 32-bit build is broken due to using _BitScanForward64() instead of _BitScanForward(), but the obvious fix doesn't completely fix it. I would attach the patch from git format-patch, but Github informs me that they don't support that file type :-/ It's only short anyway, so here it is inline:

```
From 495cc3c660d4b9b0646fccea557f0a64d79ffa05 Mon Sep 17 00:00:00 2001
From: Steve Hay <steve.m.hay@googlemail.com>
Date: Wed, 20 Oct 2021 14:57:33 +0100
Subject: [PATCH] Try to fix MSVC++ 32-bit build


diff --git a/inline.h b/inline.h
index e9b5580fff..161a777917 100644
--- a/inline.h
+++ b/inline.h
@@ -777,7 +777,11 @@ Perl_lsbit_pos64(U64 word)
 
     {
         unsigned long index;
+#    if defined(_WIN64)
         _BitScanForward64(&index, word);
+#    else
+        _BitScanForward(&index, (unsigned long)word);
+#    endif
         return (unsigned)index;
     }
 
-- 
2.31.1.windows.1
```
With that patch I can now build miniperl.exe but it does weird things in make_patchnum.pl, which I've reduced to this (inline again, since .pl files also can't be attached!)

```
use strict;
use warnings;

my $str = <<"ENDOFONE";
# 'lib/Config_git.pl'
Blah=<<'ENDOFTWO';
ENDOFTWO
ENDOFONE

print $1 if $str =~ /'([^']+)'/;
```
That bizarrely prints

```
lib/Config_git.pl'
Blah=<<'ENDOFTWO
```
instead of the expected

```
lib/Config_git.pl
```
but deleting almost any part of $str makes it work as expected.

On a side note, is there any way to change the list of accepted attachment file types? It's quite unhelpful that the Perl bug reporting interface doesn't allow either patch files or Perl scripts to be attached.

xenu:
Fixes #19205

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -730,7 +730,7 @@ Perl_is_utf8_invariant_string_loc(const U8* const s, STRLEN len, const U8 ** ep)
 #  endif
 #endif
 
-#if defined(_MSC_VER) && _MSC_VER >= 1400
+#if defined(_MSC_VER)
 #  include <intrin.h>
 #  pragma intrinsic(_BitScanForward)
 #  pragma intrinsic(_BitScanReverse)
@@ -772,7 +772,7 @@ Perl_lsbit_pos64(U64 word)
 
     return (unsigned) PERL_CTZ_64(word);
 
-#  elif U64SIZE == 8 && defined(_MSC_VER) && _MSC_VER >= 1400
+#  elif U64SIZE == 8 && defined(_MSC_VER)
 #    define PERL_HAS_FAST_GET_LSB_POS64
 
     {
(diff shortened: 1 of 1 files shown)


steve-m-hay:
Ah! I didn't think of fixing it that way ;-) Yes, that gets the build working now (and tests all pass too). Thanks.

jkeenan:
Is this ticket now closable?

PR #19208, opened 2021-10-22
Larger hashes without shared keys

nwc10:
Disable shared hash keys for "large" hashes (that are neither objects nor symbol tables).

Storing these "used just once" keys in the shared string table adds size to it. **Worse** - in any environment that has "copy on write" memory for child process (such as a pre-forking server), the memory pages used for the shared string table rapidly need to be copied as the child process manipulates hashes. Hence if most of the shared string table is such keys that are used only in one place, there is no benefit from re-use within the perl interpreter, but a high cost due to the more pages for the OS to copy.

"Large" is a heuristic - currently the heuristic is that sharing is disabled when adding a key to a hash triggers allocation of more storage, and the hash has more than 42 keys.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -117,7 +117,58 @@ There may well be none in a stable release.
 
 =item *
 
-XXX
+Large hashes no longer allocate their keys from the shared string table.
+
+The same internal datatype (C<PVHV>) is used for all of
+
+=over 4
+
+=item *
+
+Symbol tables
+
+=item *
+
+Objects (by default)
+
+=item *
+
+Associative arrays
+
+=back
+
+The shared string table was originally added to improve performance for blessed
+hashes used as objects, because every object instance has the same keys, so it
+is an optimisation to share memory between them. It also makes sense for symbol
+tables, where derived classes will have the same keys (typically method names),
+and the OP trees built for method calls can also share memory. The shared
+string table behaves roughly like a cache for hash keys.
+
+But for hashes actually used as associative arrays - mapping keys to values -
+typically the keys are not re-used in other hashes. For example, "seen" hashes
+are keyed by object IDs (or addresses), and logically these keys won't repeat
+in other hashes.
+
+Storing these "used just once" keys in the shared string table increases CPU
+and RAM use for no gain. For such keys the shared string table behaves as a
+cache with a 0% hit rate. Storing all the keys there increases the total size
+of the shared string table, as well as increasing the number of times it is
+resized as it grows. B<Worse> - in any environment that has "copy on write"
+memory for child process (such as a pre-forking server), the memory pages used
+for the shared string table rapidly need to be copied as the child process
+manipulates hashes. Hence if most of the shared string table is such keys that
+are used only in one place, there is no benefit from re-use within the perl
+interpreter, but a high cost due to more pages for the OS to copy.
+
+The perl interpreter now disables shared hash keys for "large" hashes (that are
+neither objects nor symbol tables). "Large" is a heuristic - currently the
+heuristic is that sharing is disabled when adding a key to a hash triggers
+allocation of more storage, and the hash has more than 42 keys.
+
+This B<might> cause slightly increased memory usage for programs that create
+(unblessed) data structures that contain multiple large hashes that share the
+same keys. But generally our testing suggests that for the specific cases
+described it is a win, and other code is unaffected.
 
 =back
 


hv.c
```diff
@@ -152,6 +152,12 @@ Perl_he_dup(pTHX_ const HE *e, bool shared, CLONE_PARAMS* param)
 
     PERL_ARGS_ASSERT_HE_DUP;
 
+    /* All the *_dup functions are deemed to be API, despite most being deeply
+       tied to the internals. sv_dup and sv_dup_inc seem to be the only two
+       that are used (and legitmately so) by XS code. Probably the "rest" should
+       be dropped from the API. */
```
hvds: `s/legitmately/legitimately/`. For the comment as a whole, I'm not sure this is the obvious place to put it.

embed.fnc
```diff
@@ -1019,7 +1019,7 @@ po	|struct xpvhv_aux*|hv_auxalloc|NN HV *hv
 Apd	|void	|hv_clear	|NULLOK HV *hv
 : used in SAVEHINTS() and op.c
 ApdR	|HV *	|hv_copy_hints_hv|NULLOK HV *const ohv
-Ap	|void	|hv_delayfree_ent|NN HV *hv|NULLOK HE *entry
+Ap	|void	|hv_delayfree_ent|NULLOK HV *notused|NULLOK HE *entry
```
hvds: In commit message, `s/to mark it/so mark it/`.

hv.c
```diff
@@ -353,6 +352,7 @@ Perl_hv_common_key_len(pTHX_ HV *hv, const char *key, I32 klen_i32,
     return hv_common(hv, NULL, key, klen, flags, action, val, hash);
 }
 
+/* flags might have HVhek_NOTSHARED set. If so, we need to ignore that. */
```
hvds: This comment looks oddly placed, as if it's the only thing worth noting about hv_common.

hv.c
```diff
@@ -1468,6 +1495,8 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
     if (!HvTOTALKEYS(hv))       /* skip rest if no entries */
         return;
 
+    S_large_hash_heuristic(aTHX_ hv, HvTOTALKEYS(hv));
```
hvds: The name doesn't make side-effects obvious, it might be clearer if this were:
```
  /* don't share keys in large simple hashes */
  if (S_large_hash_heuristic(aTHX_ hv, HvTOTALKEYS(hv)))
    HvSHAREKEYS_off(hv);
```
.. or if the name made the possible side-effects clearer. (And obviously the same in hv_ksplit.)

hv.h
```diff
@@ -341,6 +341,13 @@ See L</hv_fill>.
 #define HvPLACEHOLDERS_get(hv)	(SvMAGIC(hv) ? Perl_hv_placeholders_get(aTHX_ (const HV *)hv) : 0)
 #define HvPLACEHOLDERS_set(hv,p)	Perl_hv_placeholders_set(aTHX_ MUTABLE_HV(hv), p)
 
+/* This (now) flags whether *new* keys in the hash will be allocated from the
+ * shared string table. We have a heuristic to call HvSHAREKEYS_off() if a has
```
hvds: `s/has$/hash/`

pod/perldelta.pod
```diff
@@ -89,7 +89,54 @@ There may well be none in a stable release.
...
+are keyed by object IDs (or addresses), and logically these keys won't repeat
+in other hashes.
+
+Storing these "used just once" keys in the shared string table adds size to
+it. B<Worse> - in any environment that has "copy on write" memory for child
+process (such as a pre-forking server), the memory pages used for the shared
+string table rapidly need to be copied as the child process manipulates
+hashes. Hence if most of the shared string table is such keys that are used
+only in one place, there is no benefit from re-use within the perl interpreter,
+but a high cost due to the more pages for the OS to copy.
```
hvds: `s/the more/more/`

pod/perldelta.pod
```diff
@@ -89,7 +89,54 @@ There may well be none in a stable release.
...
+
+The perl interpreter now disables shared hash keys for "large" hashes (that are
+neither objects nor symbol tables). "Large" is a heuristic - currently the
+heuristic is that sharing is disabled when adding a key to a hash triggers
+allocation of more storage, and the hash has more than 42 keys.
+
+This B<might> cause slightly increased memory usage for programs that create
+(unblessed) data structures that contain multiple large hashes that share the
+same keys. But generally our testing suggest that for the specific cases
```
hvds: 'testing suggests' or 'tests suggest'

pod/perldelta.pod
```diff
@@ -89,7 +89,54 @@ There may well be none in a stable release.
...
+tables, where derived classes will have the same keys (typically method names),
+and the OP trees built for method calls can also share memory.
+
+But for hashes actually used as associative arrays - mapping keys to values -
+typically the keys are not re-used in other hashes. For example, "seen" hashes
+are keyed by object IDs (or addresses), and logically these keys won't repeat
+in other hashes.
+
+Storing these "used just once" keys in the shared string table adds size to
+it. B<Worse> - in any environment that has "copy on write" memory for child
```
hvds: I suspect the additional overhead to maintain it is also significant: it feels unlikely that my test case gave so big a time saving from reduced memory usage alone.

I assume some of that cost will be from additional splits on the shared string table, and some will be the additional lookup on each write - always failing, so worst case. Not sure what more there might be.

hvds (commented):
Minor things, mostly typos.

embed.fnc
nwc10: thanks - fixed in rebase. well spotted.

hv.c
nwc10: Thanks. Fixed this typo.

Agree that it's not an obvious place. I split the comment in the rebase, but I'm still not sure if this is the best plan. But I'm not sure what to do instead.

hv.c
nwc10: Agree.

But I'm not sure what to do better. This comment documents a "trip hazard", so I think it needs to stay **somewhere**, but I'm not sure where/how. I seem to be suffering from writers block currently (took a long time to figure out the seemingly simple changes to perldelta), so I thought it better to push the rebase with what I had, rather than stall further.

hv.c
nwc10: Your approach is much better. I went with it. Thanks.

pod/perldelta.pod
nwc10: Done.

pod/perldelta.pod
nwc10: Oops. Done. Thanks.

pod/perldelta.pod
nwc10: Yes, good point.

I reworded the perldelta section based on your comments here. (I also don't know what more there might be). Is what I pushed now good enough?

hv.h
nwc10: Well spotted. Fixed. Thanks.

hv.c
```diff
@@ -2103,14 +2103,12 @@ Perl_hv_undef_flags(pTHX_ HV *hv, U32 flags)
         Safefree(meta);
         HvAUX(hv)->xhv_mro_meta = NULL;
       }
-      if (!HvAUX(hv)->xhv_name_u.xhvnameu_name && ! HvAUX(hv)->xhv_backreferences)
-        SvFLAGS(hv) &= ~SVf_OOK;
```
hvds: Someone else definitely needs to look at this. It isn't clear to me that the `if (SvOOK)` block is safe to reenter - `meta` looks fine, but `HvENAME` and `HvNAME` are less obvious.

hv.c
```diff
@@ -2103,14 +2103,12 @@ Perl_hv_undef_flags(pTHX_ HV *hv, U32 flags)
...
-    }
-    if (!SvOOK(hv)) {
-        Safefree(HvARRAY(hv));
-        HvMAX(hv) = PERL_HASH_DEFAULT_HvMAX;        /* 7 (it's a normal hash) */
-        HvARRAY(hv) = 0;
     }
+
+    Safefree(HvARRAY(hv));
+    HvMAX(hv) = PERL_HASH_DEFAULT_HvMAX;        /* 7 (it's a normal hash) */
+    HvARRAY(hv) = 0;
```
hvds: So previously we would not have done this if there were either a name or backreferences. I'm going to guess we didn't do it before only because it would have lost the aux structure, so the change here is an optimization made possible by having moved the aux structure. If that's correct, it's worth mentioning in the commit message.

If _failure_ to do this would have been a bug, that's also worth mentioning.

hv.c
```diff
@@ -2675,6 +2673,20 @@ Perl_hv_iternext_flags(pTHX_ HV *hv, I32 flags)
...
+           non-NULL.
+           Is this actually a guarantee that we need to make? We should check
+           whether anything is actually relying on this, or if we are simply
+           making work for ourselves.
+           Potentially also we could avoid allocating HvARRAY() automatically
+           in hv_auxinit() */
+
+        char *array;
+        Newxz(array, PERL_HV_ARRAY_ALLOC_BYTES(HvMAX(hv) + 1), char);
+        HvARRAY(hv) = (HE**)array;
```
hvds: How does this change relate to the rest of the commit?

hv.c
```diff
@@ -1768,23 +1764,22 @@ S_hv_free_ent_ret(pTHX_ HV *hv, HE *entry)
 
 
 void
-Perl_hv_free_ent(pTHX_ HV *hv, HE *entry)
+Perl_hv_free_ent(pTHX_ HV *notused, HE *entry)
 {
-    SV *val;
-
-    PERL_ARGS_ASSERT_HV_FREE_ENT;
```
hvds, line 1763: Was this line intended to disappear?

hv.c
```diff
@@ -1768,23 +1764,22 @@ S_hv_free_ent_ret(pTHX_ HV *hv, HE *entry)
...
+    SV *val = hv_free_ent_ret(entry);
     SvREFCNT_dec(val);
 }
 
 
 void
-Perl_hv_delayfree_ent(pTHX_ HV *hv, HE *entry)
+Perl_hv_delayfree_ent(pTHX_ HV *notused, HE *entry)
 {
-    PERL_ARGS_ASSERT_HV_DELAYFREE_ENT;
```
hvds: I assume this ARGS_ASSERT and the one above disappear because they're now empty, but I always find that confusing. Just saying.

hv.c
```diff
@@ -49,7 +49,7 @@ holds the key and hash value.
 static const char S_strtab_error[]
     = "Cannot modify shared string table in hv_%s";
 
-#define HVhek_MASK	0xFF
+#define HVhek_STORAGE_MASK (HVhek_UTF8|HVhek_WASUTF8)
```
hvds: In commit message: `s'(at least) types'(at least) 3 types'`

I note this is the same definition as `hv.h:HVhek_ENABLEHVKFLAGS`, do we want both of them?

hv.c
```diff
@@ -728,12 +722,10 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
                 }
                 else {
                     /* Effectively this is save_hek_flags() for a new version
-                       of the HEK and Safefree() of the old rolled together.
-                       "masked_flags" doesn't have HVhek_NOTSHARED set.
-                       We could be clearer on this. */
-                    HeKFLAGS(entry) = masked_flags | HVhek_NOTSHARED;
+                       of the HEK and Safefree() of the old rolled together. */
+                    HeKFLAGS(entry) ^= HVhek_WASUTF8;
```
hvds: Does this mean we previously would have cleared `HVhek_{FREEKEY,PLACEHOLD,KEYCANONICAL}` but now we don't? What is the implication of that?

hv.c
```diff
@@ -1430,6 +1430,34 @@ S_hv_delete_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
...
+   2) symbol tables
+   3) associative arrays
+
+   shared hash keys benefit the first two greatly, because keys are likely
+   to be re-used between objects, or for constants in the optree
+
+   However, for large associative arrays (lookup tables, "seen" hashes) keys are
+   unlikely to be re-used. Hence having those keys in the shared string table as
+   well as the hash is a memory hit, if they are never actually shared with a
+   second hash. Hence we turn off shared hash keys if a (regular) has gets
```
hvds: `s'has(?!h)'hash'`, still

pod/perldelta.pod
```diff
@@ -99,7 +99,58 @@ There may well be none in a stable release.
...
+
+The perl interpreter now disables shared hash keys for "large" hashes (that are
+neither objects nor symbol tables). "Large" is a heuristic - currently the
+heuristic is that sharing is disabled when adding a key to a hash triggers
+allocation of more storage, and the hash has more than 42 keys.
+
+This B<might> cause slightly increased memory usage for programs that create
+(unblessed) data structures that contain multiple large hashes that share the
+same keys. But generally our testing suggests that for the specific cases
+describe it is a win, and other code is unaffected.
```
hvds: `s'describe'described'`

hv.c
hvds: > I'm not sure what to do instead.

For the "which of these should be API" question, a ticket might be a better place. It's probably no more likely to get acted on though.

hv.c
hvds: It might make more sense either directly by `flags &= ~HVhek_NOTSHARED`, or above the `if (keysv)` block that includes it (in which case it probably needs to clarify that this chunk's job is to sanitize `flags`).

pod/perldelta.pod
hvds: > Is what I pushed now good enough?

Yep, bar a new typo. :)

hvds (commented):
Apologies if there's some duplication here, I'm somewhat losing track.

nwc10:
> Apologies if there's some duplication here, I'm somewhat losing track.

No worries. I must apologise (a little) - I don't think that I'm helping keep things sane because I rebased this branch onto my branch `return-hv_with_aux-to-correct-arena` because I thought it important to have that bug fixed before considering further refactorings. Which might have been OK, but I forgot to say. Bad me.

I have now commented to this effect on #19224 

I might not get a chance to look at your (useful) feedback until tomorrow.

hv.c
nwc10: It is, and agree. Not sure what could be done to improve this.

hv.c
nwc10: No, failure to do so would not have been a bug. Just a pessimisation.

nwc10:
> I might not get a chance to look at your (useful) feedback until tomorrow.

For various reasons it took a lot longer than that. I've how had time to address the first three commits from this branch (to 93fa56bdcfef5bad5acd02fdb34221db81613551) which are the memory exhaustion bug fix branch `return-hv_with_aux-to-correct-arena` on which this branch is based. However, I don't have time (before this evening) to address the rest of your comments and rebase the rest of this branch on that.

I pushed the other branch "early" so that Andreas is able to test it, as that part is a bug fix, not the optimisation.

hv.c
nwc10: No, we wouldn't have cleared them previously.

I agree, this code really isn't clear, but I think it wasn't before, rather than I'm making it worse. `HeKFLAGS(entry)` is 8 bits:

`perl Porting/expand-macro.pl HeKFLAGS` says:

```
# 4 "HeKFLAGS expands to"
(*((unsigned char *)(((A0)->hent_hek)->hek_key)+((A0)->hent_hek)->hek_len+1))
```

(ie assigns to an `unsigned char` LVALUE)

And the flags you mentioned are all explicitly outside of 8 bits:

```
/* the following flags are options for functions, they are not stored in heks */
#define HVhek_FREEKEY	0x100 /* Internal flag to say key is Newx()ed.  */
#define HVhek_PLACEHOLD	0x200 /* Internal flag to create placeholder.
                               * (may change, but Storable is a core module) */
#define HVhek_KEYCANONICAL 0x400 /* Internal flag - key is in canonical form.
                                    If the string is UTF-8, it cannot be
                                    converted to bytes. */
```

I *think* that this resolves your question (very belatedly), but if so, I think you should hit "Resolve conversation", not me. (And if I'm wrong or unclear, please keep me honest by asking further)

hv.c
nwc10: I believe that we addressed this as part of the PR that merged the branch `return-hv_with_aux-to-correct-arena`

hv.c
nwc10: I believe that we addressed this as part of the PR that merged the branch `return-hv_with_aux-to-correct-arena`

pod/perldelta.pod
nwc10: (Finally) fixed in rebase

hv.c
nwc10: Fixed in rebase

hv.c
hvds: What year was this again? :p

I'm guessing a bit here, not helped by github having lost the commit and I suspect corrupted the context of the conversation, but the original question was trying to understand correctness (not per se complaining about clarity). The possibility that the mentioned flags would simply not fit in `HeKFLAGS(entry)` had not occurred to me, so I'm pretty sure that answers my original question, and I'll mark it resolved.

hv.c
nwc10: Created #19409 for this.

hv.c
nwc10: > It might make more sense either directly by `flags &= ~HVhek_NOTSHARED`, or above the `if (keysv)` block that includes it (in which case it probably needs to clarify that this chunk's job is to sanitize `flags`).

I did this, expanding it with the clarification. It works much better there. Thanks for figuring out this improvement.

hv.c
nwc10: So...

I *think* so. But I thought quite a bit about it because it's not obvious what the right answer is. I changed the definition to this:

    #define HVhek_STORAGE_MASK (0xFF & ~HVhek_NOTSHARED)

which for the current 3 (used) bits is equivalent, and wrote a long comment to clarify what the 3 bits are used for, why we don't have any obvious single way to "group" them, and why it (probably) makes sense to have both `HVhek_ENABLEHVKFLAGS` and `HVhek_STORAGE_MASK`.

This is the commit whose message starts *Eliminate "masked_flags" from functions in hv.c*. After rebasing and editing, it is now 4f72a47f3a7931baf20d4d81a086b0f765e0bd62

demerphq:
On Fri, 22 Oct 2021 at 11:44, Nicholas Clark ***@***.***>
wrote:

> Disable shared hash keys for "large" hashes (that are neither objects nor
> symbol tables).
>
> Storing these "used just once" keys in the shared string table adds size
> to it. *Worse* - in any environment that has "copy on write" memory for
> child process (such as a pre-forking server), the memory pages used for the
> shared string table rapidly need to be copied as the child process
> manipulates hashes. Hence if most of the shared string table is such keys
> that are used only in one place, there is no benefit from re-use within the
> perl interpreter, but a high cost due to the more pages for the OS to copy.
>
> "Large" is a heuristic - currently the heuristic is that sharing is
> disabled when adding a key to a hash triggers allocation of more storage,
> and the hash has more than 42 keys.
>

Attached is a benchmark you can use.

./perl -Ilib ../hash_fork.pl --n=100000 --base_n=0

shows a case where the branch is faster and uses less memory.

./perl -Ilib ../hash_fork.pl --n=100000

shows a case where the branch is slower and uses about the same memory.

demerphq:
On Fri, 22 Oct 2021 at 11:44, Nicholas Clark ***@***.***>
wrote:

> Disable shared hash keys for "large" hashes (that are neither objects nor
> symbol tables).
>
> Storing these "used just once" keys in the shared string table adds size
> to it. *Worse* - in any environment that has "copy on write" memory for
> child process (such as a pre-forking server), the memory pages used for the
> shared string table rapidly need to be copied as the child process
> manipulates hashes. Hence if most of the shared string table is such keys
> that are used only in one place, there is no benefit from re-use within the
> perl interpreter, but a high cost due to the more pages for the OS to copy.
>
> "Large" is a heuristic - currently the heuristic is that sharing is
> disabled when adding a key to a hash triggers allocation of more storage,
> and the hash has more than 42 keys.
>
For those not versed in the minutae of the hash engine, this means that the
"desharing" will occur when the bucket array is resized from 64 to 128
buckets. Specifically the hash table implementation will resize the array
on the first collision that occurs when there are more than 42 keys in the
array. If we wanted it to occur earlier the next reasonable number would be
22 and if we wanted it to occur later it would be 85. This also means that
the exact time the split/desharing occurs is non-deterministic and depends
on the hash function in use and the hash seed and the actual keys involved.

Yves

nwc10:
> For those not versed in the minutae of the hash engine, this means that the "desharing" will occur when the bucket array is resized from 64 to 128 buckets. Specifically the hash table implementation will resize the array on the first collision that occurs when there are more than 42 keys in the array. If we wanted it to occur earlier the next reasonable number would be 22 and if we wanted it to occur later it would be 85. This also means that the exact time the split/desharing occurs is non-deterministic and depends on the hash function in use and the hash seed and the actual keys involved.

Yes, exactly, thanks for the important observation that this is non-deterministic.

I figure I should add that. But the "too much detail" version is that, the implementation is careful to permit any individual hash to have a mix of shared and unshared keys. Stopping sharing is just a single flag bit change. This extra implementation work was done to avoid any sort of performance "step" when the threshold kicks in and sharing stops. Without this, stopping sharing would require iterating the hash keys and reallocating each into non-shared storage, which would be purely wasted work for hashes sized just above the threshold.

So I guess I should add a shorter version of that.

demerphq:
On Thu, 17 Feb 2022 at 12:37, Nicholas Clark ***@***.***>
wrote:

> For those not versed in the minutae of the hash engine, this means that
> the "desharing" will occur when the bucket array is resized from 64 to 128
> buckets. Specifically the hash table implementation will resize the array
> on the first collision that occurs when there are more than 42 keys in the
> array. If we wanted it to occur earlier the next reasonable number would be
> 22 and if we wanted it to occur later it would be 85. This also means that
> the exact time the split/desharing occurs is non-deterministic and depends
> on the hash function in use and the hash seed and the actual keys involved.
>
> Yes, exactly, thanks for the important observation that this is
> non-deterministic.
>
> I figure I should add that. But the "too much detail" version is that, the
> implementation is careful to permit any individual hash to have a mix of
> shared and unshared keys. Stopping sharing is just a single flag bit
> change. This extra implementation work was done to avoid any sort of
> performance "step" when the threshold kicks in and sharing stops. Without
> this, stopping sharing would require iterating the hash keys and
> reallocating each into non-shared storage, which would be purely wasted
> work for hashes sized just above the threshold.
>
It took me a while to click to what you are saying here. So it is not that
the hash changes from shared to unshared, but rather that  we simply stop
sharing the keys we put into it after a certain point, thus the hash starts
off as "all keys shared" and then switches to "mixed shared and unshared".
That makes sense.

This got me thinking about the benchmark where the new logic is worse. The
"worse" behavior is where I *want* to build a master hash table with the
keys I am going to use for the other hashes. In this case the new behavior
would cause the master table to not share most of its keys, and thus when I
start creating new hashes with the same keys in them they wont be shared,
and I will use more ram and more time because I have to construct more keys.

Which makes me think maybe we should have a pragmata to control this
behavior and not just make it conditional on the number of keys in the hash.

I envisage four cases maybe?

use keys 'shared';
use keys 'optimistic';
use keys 'pessimistic';
use keys 'unshared';

Shared would be the "old" behavior. Share every key. Optimistic would
basically be the logic you described, except that it would check PL_strtab
to see if the key was already shared, and it was then it would get reused.
Pessimistic would be the logic you described, eg once the bit was flipped
it would not check PL_strtab. "Unshared" would be "dont share any keys
created in this block". I would assume that 'optimistic' would be the
default. Maybe the pragmata would affect the hashes created in their
context, or maybe it would affect the key operations.

This would mean that as a dev I could tune this behavior to the needs and
context of my code. If I know I am going to build a master table then I can
do so with "shared" mode, the new behavior would not kick in and I would
preload PL_strtab with the keys I am going to share, later on I would use
'optimistic' and all my hashes would share the keys that had been
previously created in PL_strtab but would not add new ones from large
hashes. The benchmark case that was slower and fatter could be setup like
this and it would be fast again.

For cases where I know there will be little overlap between keys I would
use "unshared" or "pessimistic" depending on whether I think I would be
creating objects or just writing code that did "large hash aggregations".

This is a bit hand-wavey but do you see where I am going here?

For example, imagine I build a master hash of all the keys from dict.txt, i
then scan a bunch of files and build a per file hash of the counts of the
words therein. The master hash would ensure that i only store one copy of
each word in memory, even though I  might have thousands of files.

Another example would be I am building a small number of hashes with large
numbers of SHA256's,  in this case I dont want  to share any of my keys.

Now modulo both with code that wants to construct large numbers of objects
with a small number of keys in them. Maybe these objects are the values in
the hashes mentioned above.

That basically would give us the four cases I am describing.

Yves

nwc10:
> Attached is a benchmark you can use.

Thanks.

> ./perl -Ilib ../hash_fork.pl --n=100000 --base_n=0
>
> shows a case where the branch is faster and uses less memory.
>
> ./perl -Ilib ../hash_fork.pl --n=100000
>
> shows a case where the branch is slower and uses about the same memory.

Yes, these behaviours make sense. The benchmark code is timing hash creation in the children post-fork.

For the second invocation, where a (perl visible) hash is created in the parent pre-fork (with the same keys), on blead this means that the shared string table already has all those keys, so the (second) hash creation is simply adjusting reference counts on the shared string table. (which will be un-pinging Copy On Write memory pages at the OS level). On the branch, non-shared keys mean that creating hashes means `malloc` for the keys.

For the first invocation, there is no "master" hash in memory, so the first invocation (I believe) is just benchmarking hash creation (for lots of only-seen-once keys), and the branch is winning. I don't think that the forking really has much to do with this.

> This got me thinking about the benchmark where the new logic is worse. The "worse" behavior is where I *want* to build a master hash table with the keys I am going to use for the other hashes. In this case the new behavior would cause the master table to not share most of its keys, and thus when I start creating new hashes with the same keys in them they wont be shared, and I will use more ram and more time because I have to construct more keys.

Yes, this particular behaviour will be worse on the branch. But I think really only this particular use case. Other use cases will be similar, or faster. (eg Hugo's existing code, which doesn't use `fork` - it's a pure win from not creating shared keys that are never re-used)

> Which makes me think maybe we should have a pragmata to control this behavior and not just make it conditional on the number of keys in the hash.
>
> I envisage four cases maybe?
>
> ```
> use keys 'shared';
> use keys 'optimistic';
> use keys 'pessimistic';
> use keys 'unshared';
> ```

I'm not totally sure either for how many cases are useful.

I'd see this as a "future direction" -  I think the behaviour in the branch is a better default than what we have, and hence I'd like to merge it and switch to that behaviour. I think this would (on average, for the vast majority of cases) improve our defaults for running existing code, and that doing this now, and adding flexibility later is better than doing nothing until we have flexibility too.

hvds:
LGTM (assuming a688949277 is still the correct head); if @demerphq has time it'd be good to get review from him too.

What are the next steps here, is there any prospect of getting it into 5.36?

nwc10:
> What are the next steps here, is there any prospect of getting it into 5.36?

Given that

1. it's not a contraversial change (well, no-one has disliked it yet)
2. PSC are explicitly keen to [*pick up the pace*](https://www.nntp.perl.org/group/perl.perl5.porters/2022/03/msg263374.html) which I read as tilting the balance a bit less than ultra-conservative
3. if it does cause problems worse case it's easy to revert, and I think also likely easy to identify these changes as the cause

I propose that if no-one objects in the next 48 hours we merge it. Where "we" is I, unless I fall under a bus before Saturday afternoon.

hvds:
> I propose that if no-one objects in the next 48 hours we merge it. Where "we" is I, unless I fall under a bus before Saturday afternoon.

I'm mildly nervous only because we've seen an unusually high number of BBC tickets recently, and your availability seems to have been quite sporadic. If you think you'll have capacity in the coming weeks should there be any fallout, I'd be good with this.

Note that there is a (probably trivial) conflict with c3c9d6b15f on rebase.

khwilliamson:
On 3/17/22 06:52, Nicholas Clark wrote:
>     What are the next steps here, is there any prospect of getting it
>     into 5.36?
> 
> Given that
> 
>  1. it's not a contraversial change (well, no-one has disliked it yet)
>  2. PSC are explicitly keen to /pick up the pace/
>     <https://www.nntp.perl.org/group/perl.perl5.porters/2022/03/msg263374.html>
>     which I read as tilting the balance a bit less than ultra-conservative
>  3. if it does cause problems worse case it's easy to revert, and I
>     think also likely easy to identify these changes as the cause
> 
> I propose that if no-one objects in the next 48 hours we merge it. Where 
> "we" is I, unless I fall under a bus before Saturday afternoon.
> 
> —


+1

PR #19216, opened 2021-10-29
Fixed possible syntax error in Configure's test program for Inf/NaN bytes

t-a-k:
In `Configure` script, test program (`try.c`) for probing `(double|longdbl)(inf|nan)bytes` seems to lack some preprocessor directives (probably accidentally removed) and caused syntax error if `HAS_LONG_DOUBLE` is not defined.

I'm not sure whether any supported build environment that leave `HAS_LONG_DOUBLE` undefined might exist today.  I have tested this change with `./Configure -Ud_longdbl`.

### DIFF
--- a/Configure
+++ b/Configure
@@ -20777,10 +20777,12 @@ int main(int argc, char *argv[]) {
     memset((char *)&ldinf + 10, '\0', LONG_DOUBLESIZE - 10);
     memset((char *)&ldnan + 10, '\0', LONG_DOUBLESIZE - 10);
 # endif
+#endif
   if (argc == 2) {
     switch (argv[1][0]) {
     case '1': bytes(&dinf, sizeof(dinf)); break;
     case '2': bytes(&dnan, sizeof(dnan)); break;
+#ifdef HAS_LONG_DOUBLE
     case '3': bytes(&ldinf, sizeof(ldinf)); break;
     case '4': bytes(&ldnan, sizeof(ldnan)); break;
 #endif


jkeenan:
> In `Configure` script, test program (`try.c`) for probing `(double|longdbl)(inf|nan)bytes` seems to lack some preprocessor directives (probably accidentally removed) and caused syntax error if `HAS_LONG_DOUBLE` is not defined.
> 
> I'm not sure whether any supported build environment that leave `HAS_LONG_DOUBLE` undefined might exist today. I have tested this change with `./Configure -Ud_longdbl`.

For reasons described below, I am skeptical that this change in `Configure` is needed.

First, let me ask:  Can you provide an actual example of a C-level syntax error in `try.c` when running `Configure`?  (Your original post merely suggests that such an error might be possible.)

For purpose of exploration I created a [jkeenan/gh-19216-configure-longdbl-exploration](https://github.com/Perl/perl5/tree/jkeenan/gh-19216-configure-longdbl-exploration) branch in our repository.  That branch differs from blead only in that I added two `echo` statements to `Configure` for debugging purposes.
```
$ git diff blead..HEAD -- Configure
diff --git a/Configure b/Configure
index dd19ad4e81..53092d6f77 100755
--- a/Configure
+++ b/Configure
@@ -6882,6 +6882,8 @@ $rm_try
 set d_longdbl
 eval $setvar
 
+$echo "WWWWWWWWW $d_longdbl WWWWWWWWW" >&4
+
 : see if ldexpl exists
 set ldexpl d_ldexpl
 eval $inlibc
@@ -20728,6 +20730,8 @@ case "$ans" in
     *)          default_inc_excludes_dot="$define" ;;
 esac
 
+$echo "XXXXXXXXX $d_longdbl XXXXXXXXX" >&4
+
 : Check what kind of inf/nan your system has
 $echo "Checking the kind of infinities and nans you have..." >&4
 $echo "(The following tests may crash.  That's okay.)" >&4

```
The second of these two `echo` statements is just above the creation of the instance of `try.c` which your p.r. proposes to patch.

I first configured with a "regular" invocation: `sh ./Configure -des -Dusedevel`.  I recorded the output from this command and also made a copy of the resulting `config.sh` file.

I then configured with a non-regular, "U" invocation: `sh ./Configure -des -Dusedevel -Ud_longdbl`.  I recorded the output from this command and also made a copy of the resulting `config.sh` file.

Comparing the two outputs from `Configure`, I got:
```
$ grep -n -E '(WWWW|XXXX)' 541c99624a.configure.regular.output.txt 541c99624a.configure.U.output.txt
541c99624a.configure.regular.output.txt:96:WWWWWWWWW define WWWWWWWWW
541c99624a.configure.regular.output.txt:717:XXXXXXXXX define XXXXXXXXX
541c99624a.configure.U.output.txt:99:WWWWWWWWW undef WWWWWWWWW
541c99624a.configure.U.output.txt:716:XXXXXXXXX undef XXXXXXXXX
```
From this I infer that `d_longdbl` is defined by default and only `undef` when you include `-Ud_longdbl` in your invocation of `Configure`.

Assuming that that is correct, then in "regular" builds the block in `try.c` starting at `#ifdef HAS_LONG_DOUBLE` is always entered and executed, while in "U" builds that block is never entered and nothing else needs to be done.  In which case, the code in that instance of `try.c` appears to be fine as it is.

Caveat:  I don't claim to be an expert in either `Configure` or in C.  @doughera88 , @tux, @arc:  can you take a look?

Thank you very much.
Jim Keenan

sisyphus:
@jkeenan, the thing is that if <code>HAS_LONG_DOUBLE</code> is not defined, then that particular try.c effectively becomes:
```
int main(int argc, char *argv[]) {
   double dinf = exp(1e9);
   double dnan = sqrt(-1.0);
  }
  }
  return 0;
}
```
That won't compile because of syntactic errors caused by the 2 unmatched closing brackets.
And if it was made compilable by removing those 2 closing brackets, then it doesn't do anything except return 0.

The approach taken by @t-a-k  addresses this issue such that, if HAS_LONG_DOUBLE is not defined, then try.c effectively becomes:
```
int main(int argc, char *argv[]) {
  double dinf = exp(1e9);
  double dnan = sqrt(-1.0);
  if (argc == 2) {
    switch (argv[1][0]) {
    case '1': bytes(&dinf, sizeof(dinf)); break;
    case '2': bytes(&dnan, sizeof(dnan)); break;
    }
  }
  return 0;
}
```
For the case where HAS_LONG_DOUBLE is defined, then try.c is effectively unchanged by @t-a-k's patch.

t-a-k:
@sisyphus's comment is what I wanted to say.
Thank you, and sorry for my poor explanation.

I'd also note that `Configure` seems to silently fall back to built-in default `double(inf|nan)bytes` when this `try.c` didn't compile, so it might appear to work even with this particular compilation error.

jkeenan:
@Tux, can you take a look at this pull request?  Thanks.

Tux:
I think I agree with @sisyphus ' analysis, so I'm fine with it

khwilliamson:
I am submitting a metaconfig change that corresponds with this

PR #19219, opened 2021-11-02
porting/bidi-control.t: add a porting test for bidi control chars

rjbs:
This is a simple but, I believe, effective test for bidi control characters entering the p5 source.

Obviously, this file needs to be considered specially. :-)

### DIFF
--- a/t/porting/bidi-control.t
+++ b/t/porting/bidi-control.t
@@ -2,12 +2,16 @@
 use v5.20.0;
 use warnings;
 
-use utf8 ();
+use PerlIO::encoding ();
 
 BEGIN {
   @INC = '..' if -f '../TestInit.pm';
   require './test.pl';
 }
+
+my %SKIP_FILE = (
+  'cpan/Encode/t/big5-hkscs.enc' => 1, # weird data file and not really source
+);
 
 use TestInit qw(T); # T is chdir to the top level
 
@@ -42,11 +46,13 @@ unless (@files) {
 }
 
 FILE: for my $file (@files) {
-  open my $fh, '<', "$file"
+  next if $SKIP_FILE{$file};
+
+  open my $fh, '<:encoding(UTF-8)', "$file"
     or die "can't open $file: $!";
 
+  no warnings 'utf8';
   while (defined (my $line = <$fh>)) {
-    utf8::decode($line);
     push @errors, "$file at line $." if $line =~ /\p{Bidi_Control}/;
   }
 }
--- a/t/porting/bidi-control.t
+++ b/t/porting/bidi-control.t
@@ -1,0 +1,61 @@
+#!/usr/bin/perl
+use v5.20.0;
+use warnings;
+
+use utf8 ();
+
+BEGIN {
+  @INC = '..' if -f '../TestInit.pm';
+  require './test.pl';
+}
+
+use TestInit qw(T); # T is chdir to the top level
+
+find_git_or_skip();
+
+my $cdup = `git rev-parse --show-cdup`;
+die "couldn't git-rev-parse --show-cdup\n" if $?;
+chomp $cdup;
+
+if (length $cdup) {
+  chdir $cdup or die "can't chdir to git root: $!";
+}
+
+my @files = `git ls-files`;
+die "couldn't git-ls-files\n" if $?;
+
+chomp @files;
+
+if (0) {
+  # Is this a useful filter?  Should we test every file?  Do we contain binary
+  # files, etc? -- rjbs, 2021-11-02
+  my %is_interesting = map {; $_ => 1 } qw( bat c h pl pm sh t xs );
+  @files = grep {; /\.([a-z]+)\z/ && $is_interesting{$1} } @files;
+}
+
+my @errors;
+
+unless (@files) {
+  fail("Something's wrong: no files to check?!");
+  done_testing;
+  exit;
+}
+
+FILE: for my $file (@files) {
+  open my $fh, '<', "$file"
+    or die "can't open $file: $!";
+
+  while (defined (my $line = <$fh>)) {
+    utf8::decode($line);
+    push @errors, "$file at line $." if $line =~ /\p{Bidi_Control}/;
+  }
+}
+
+if (@errors) {
+  fail("Bidi control characters found in source");
+  diag("Bidi control characters in $_") for @errors;
+} else {
+  pass("No bidi control characters found in source");
+}
+
+done_testing;
--- a/MANIFEST
+++ b/MANIFEST
@@ -6015,6 +6015,7 @@ t/porting/bench/oddentry	a test file for t/porting/bench.t
 t/porting/bench/ret0		a test file for t/porting/bench.t
 t/porting/bench/synerr		a test file for t/porting/bench.t
 t/porting/bench_selftest.t	run Porting/bench.pl's selftest facility
+t/porting/bidi-control.t	Check that source does not contain bidi controls
 t/porting/bincompat.t		Check that {non_,}bincompat_options are ordered
 t/porting/checkcase.t		Check whether we are case-insensitive-fs-friendly
 t/porting/checkcfgvar.t		Check that all config.sh-like files are good


jkeenan:
I generally expect tests under `t/` to be runnable in both of 2 ways:
```
$ cd t; ./perl harness -v porting/bidi-control.t; cd -

ok 1 - No bidi control characters found in source
ok
All tests successful.
Files=1, Tests=1,  2 wallclock secs ( 0.02 usr  0.00 sys +  1.45 cusr  0.20 csys =  1.66 CPU)
Result: PASS
/home/jkeenan/gitwork/perl
```
... and ...
```
$ ./perl -Ilib t/porting/bidi-control.t
Can't locate ./test.pl at t/porting/bidi-control.t line 9.
BEGIN failed--compilation aborted at t/porting/bidi-control.t line 10.
```
We should fix the latter before merging.

Thank you very much.
Jim Keenan

jkeenan:
> I generally expect tests under `t/` to be runnable in both of 2 ways:
> 
> ```
> $ cd t; ./perl harness -v porting/bidi-control.t; cd -
> 
> ok 1 - No bidi control characters found in source
> ok
> All tests successful.
> Files=1, Tests=1,  2 wallclock secs ( 0.02 usr  0.00 sys +  1.45 cusr  0.20 csys =  1.66 CPU)
> Result: PASS
> /home/jkeenan/gitwork/perl
> ```
> 
> ... and ...
> 
> ```
> $ ./perl -Ilib t/porting/bidi-control.t
> Can't locate ./test.pl at t/porting/bidi-control.t line 9.
> BEGIN failed--compilation aborted at t/porting/bidi-control.t line 10.
> ```
> 
> We should fix the latter before merging.
> 
Perhaps:
```
diff --git a/t/porting/bidi-control.t b/t/porting/bidi-control.t
index 17c76f4899..fd9b11d9f8 100644
--- a/t/porting/bidi-control.t
+++ b/t/porting/bidi-control.t
@@ -5,16 +5,15 @@ use warnings;
 use PerlIO::encoding ();
 
 BEGIN {
-  @INC = '..' if -f '../TestInit.pm';
-  require './test.pl';
+  chdir '..' if -f 'test.pl';
+  @INC = ( qw| lib | );
+  require './t/test.pl';
 }
 
 my %SKIP_FILE = (
   'cpan/Encode/t/big5-hkscs.enc' => 1, # weird data file and not really source
 );
 
-use TestInit qw(T); # T is chdir to the top level
-
 find_git_or_skip();
 
 my $cdup = `git rev-parse --show-cdup`;
```

t/porting/bidi-control.t
```diff
@@ -0,0 +1,61 @@
...
+die "couldn't git-ls-files\n" if $?;
+
+chomp @files;
+
+if (0) {
+  # Is this a useful filter?  Should we test every file?  Do we contain binary
+  # files, etc? -- rjbs, 2021-11-02
+  my %is_interesting = map {; $_ => 1 } qw( bat c h pl pm sh t xs );
+  @files = grep {; /\.([a-z]+)\z/ && $is_interesting{$1} } @files;
+}
```
hvds, line 38: My inclination is that it shouldn't be useful: our source code repository should contain source. But I don't think we've ever had an explicit policy in that regard, so we might need to create one. And I don't know what's in the tree today (but see below).

In particular, I think it should be possible to have files with no extension, and it should be ok to test those as well, but a quick check shows we don't have such today. What we do appear to have is 127 different extensions - 89 if you ignore `README.\w+` - and it feels like it would be safer to exclude some extensions from this check (if we really must) than to list only those that should be included. In earlier discussions I suggested to start without exceptions, and add a mechanism for them only at the point we find we really need one.

Peeking at some of the rarer extensions, I note that `os2/diff.configure` has been an empty file since 2014. Several other outliers look to be support files for tests (eg for `IO::Compress`), but a bunch more sure look like code.

hvds, line 38: My gut reaction was that it shouldn't be useful - our source code repository should store store code. But I'm not aware there's ever been policy in this area, and on looking I see some obviously reasonable counterexamples such as test files for `IO::Compress`.

However we have 125 file extensions - 89 if you ignore `README.\w+` - and a high proportion of them appear to be text, with rather more than the short list above being code of some sort. Binary files are TTBOMK added rarely and changed far less often, so if none of the current crop happens to include Bidi controls I think it'd be far safer to leave the creation of an exception mechanism until the point we get a concrete example to base it around.

t/porting/bidi-control.t
```diff
@@ -42,11 +46,13 @@ unless (@files) {
 }
 
 FILE: for my $file (@files) {
-  open my $fh, '<', "$file"
+  next if $SKIP_FILE{$file};
```
hvds, line 49: Ah, I see you found an example, and invented a mechanism. The `if (0)` block seems quite redundant now.

jkeenan:
@khwilliamson, would you like to comment on this p.r.?

t/porting/bidi-control.t
```diff
@@ -0,0 +1,67 @@
...
+
+use PerlIO::encoding ();
+
+BEGIN {
+  @INC = '..' if -f '../TestInit.pm';
+  require './test.pl';
+}
+
+my %SKIP_FILE = (
+  'cpan/Encode/t/big5-hkscs.enc' => 1, # weird data file and not really source
```
khwilliamson, line 13: Big5 is a non-Unicode encoding.  The hk probably stands for Hong Kong.  I haven't investigated further.

t/porting/bidi-control.t
khwilliamson: Why not use file test operator -T.  It will return false on binary files.  Porting/makerel has code for creating an EBCDIC translation.  That works fine for deciding which files to use without any lists.

t/porting/bidi-control.t
```diff
@@ -0,0 +1,67 @@
...
+
+FILE: for my $file (@files) {
+  next if $SKIP_FILE{$file};
+
+  open my $fh, '<:encoding(UTF-8)', "$file"
+    or die "can't open $file: $!";
+
+  no warnings 'utf8';
+  while (defined (my $line = <$fh>)) {
+    push @errors, "$file at line $." if $line =~ /\p{Bidi_Control}/;
```
khwilliamson, line 56: I think this will find harmless characters as well as the problematic ones.  I am unsure which are the ones to be wary of.  But there are only 12 total, so could leave it in until it showed a problem, and decide then if it was a false positive and pare back the test accordingly

jkeenan:
@rjbs can you respond to @khwilliamson's comments on this proposed porting test?

Thank you very much.
Jim Keenan

khwilliamson:
Also, Unicode has some properties for distinguishing between characters that are reasonable to use in computer identifiers, and those which are not.  Not all \w characters are suitable.  Obsolete and archaic characters really shouldn't be used, for example.  Perl already supports  Identifier_Status, which is a two valued property 'Allowed' and 'Restricted', that sums up the other property: Identifier_Type which has values, like Obsolete, for finer grained restrictions.  \p{Identifier_Status =Allowed} removes about 30K \w characters from being in identifiers.

We could change perl to accept only the allowable ones, and I doubt that much if any code would be affected.  Or we could do a use feature to restrict it.

The Bidi characters are a problem when the direction change overwrites the visual space of something else, obscuring what was there.  I tried to get people more involved with Unicode to take an interest in creating an algorithm to detect this.  What came from it was a succinct summary of the problem, but I doubt that anyone will soon do the work needed to do this.

We could prohibit identifiers, again possibly under a use feature, that are of mixed scripts, and biti controls that are not in identifiers in R-to-L scripts.

jkeenan:
> @rjbs can you respond to @khwilliamson's comments on this proposed porting test?
> 
> Thank you very much. Jim Keenan

There has been no activity in this p.r. in 6 months.

@rjbs, do you wish to pursue this p.r. further?

jkeenan:
> > @rjbs can you respond to @khwilliamson's comments on this proposed porting test?
> > Thank you very much. Jim Keenan
> 
> There has been no activity in this p.r. in 6 months.
> 
> @rjbs, do you wish to pursue this p.r. further?

No response, so I'm going to close the ticket.  @rjbs, if you want to pursue this, please re-open.

Thank you very much.
Jim Keenan

PR #19220, opened 2021-11-02
Unencrypted git protocol no longer supported by github.com

jkeenan:
Reference: https://github.blog/2021-09-01-improving-git-protocol-security-github/

### DIFF
--- a/pod/perlgit.pod
+++ b/pod/perlgit.pod
@@ -22,12 +22,9 @@ I<github.com>.
 
 You can make a read-only clone of the repository by running:
 
-  % git clone git://github.com/Perl/perl5.git perl
+  % git clone git@github.com:Perl/perl5.git perl
 
-This uses the git protocol (port 9418).
-
-If you cannot use the git protocol for firewall reasons, you can also
-clone via http:
+If you cannot use that for firewall reasons, you can also clone via http:
 
   % git clone https://github.com/Perl/perl5.git perl
 
--- a/pod/perlhack.pod
+++ b/pod/perlhack.pod
@@ -175,7 +175,7 @@ L<perlgit>.
 You will need a copy of Git for your computer.  You can fetch a copy of
 the repository using the git protocol:
 
-  % git clone git://github.com/Perl/perl5.git perl
+  % git clone git@github.com:Perl/perl5.git perl
 
 This clones the repository and makes a local copy in the F<perl>
 directory.


pod/perlhack.pod
```diff
@@ -175,7 +175,7 @@ L<perlgit>.
 You will need a copy of Git for your computer.  You can fetch a copy of
 the repository using the git protocol:
```
ilmari, line 176: This no longer uses the git protocol, but ssh.

pod/perlgit.pod
```diff
@@ -22,12 +22,9 @@ I<github.com>.
 
 You can make a read-only clone of the repository by running:
 
-  % git clone git://github.com/Perl/perl5.git perl
+  % git clone git@github.com:Perl/perl5.git perl
```
ilmari, line 25: A clone made over ssh is not necessarily read-only, it depends on your user's permissions on the repo. Also, this only works for users with ssh keys registered on github, so it might be better to give the HTTPS example first, since that works anonymously.

OTOH, the bits about creating a fork and a pull request further down do assume the user has a github account, so that might be unnecessary.

Grinnz: I agree, I use HTTPS for all read-only clones. This URL would be https://github.com/Perl/perl5.git

pod/perlhack.pod
```diff
@@ -175,7 +175,7 @@ L<perlgit>.
...
 
-  % git clone git://github.com/Perl/perl5.git perl
+  % git clone git@github.com:Perl/perl5.git perl
```
Grinnz, line 178: Since this is for read access, I believe it should also present the HTTPS URL for ease of access.

Grinnz: My mistake, the HTTPS URL is presented below. Perhaps we should just remove the "git protocol" section here and skip to the HTTPS clone option.

jkeenan:
See revisions, squashed and force-pushed.

pod/perlhack.pod
```diff
@@ -172,16 +172,7 @@ L<perlgit>.
 
 =head2 Read access via Git
 
-You will need a copy of Git for your computer.  You can fetch a copy of
```
Grinnz, line 175: Some of this verbiage could be left in but also maybe isn't necessary in 2021. LGTM either way

jkeenan:
@rjbs, would you like to comment on this pull request?  (Otherwise, since I believe other commenters' concerns have been addressed, I will merge it in a few days.)

pod/perlgit.pod
Grinnz: This seems to have been reverted to the original SSH clone change?

Grinnz: If intentional, some additional explanation would be appropriate: it is not a read-only clone, and it requires you to set up SSH keys for github, and there are no "firewall reasons" relevant to it or the HTTPS clone option.

jkeenan:
Merged in commit a85e04e2281234a61c051f8f3ff63bed7381902c; closing.

jkeenan:
On 11/9/21 5:55 PM, Dan Book wrote:
> 
> 
> In pod/perlgit.pod 
> <https://github.com/Perl/perl5/pull/19220#discussion_r746110528>:
> 
>>  You can make a read-only clone of the repository by running:
>   
> -  % git clone git://github.com/Perl/perl5.git perl
> +  % git clone ***@***.***:Perl/perl5.git perl
> 
> This seems to have been reverted to the original SSH clone change?
> 

I didn't see these comments before merging.  Let's start over.

Issue #19223, opened 2021-11-08
v5.35.5 consumes significantly more memory than v5.35.4


andk:
This Issue has been posted to perl5-porters as https://www.nntp.perl.org/group/perl.perl5.porters/2021/11/msg261832.html by myself. But it did not find its way into the bug tracker, so I figured it might be a better plan to open a ticket here.
----start of posting----
Hi Nicholas,

I have a smoker that runs the cpan shell for about 10 days in a single
process. Its duty is simply to smoke all distros currently on the CPAN
in a random order. I have gathered experiences with this smoker for
about three years: up until v5.35.5 I never observed that it ran out of
memory. Now, with v5.35.5 it did run out of memory. This is
reproducable.

To provide an estimate of the scale of the impact: the VM that runs the
smoker is equipped with 12G memory plus 0.5G swap; for a smoke run with
v5.35.4 munin showed that it had reached an amount of memory committed
of 6G after having run for ten days; with v5.35.5 the smoker was killed
with OOM after about three days; that means in a third of the time we
consumed twice as much memory, so I estimate the scale impact to be
sixfold.

Bisect to the rescue:

  94ee6ed79dbca73d0345b745534477e4017fb990 is the first bad commit                                 
  commit 94ee6ed79dbca73d0345b745534477e4017fb990
  Author: Nicholas Clark <nick@ccl4.org>
  Date:   Fri Aug 20 15:20:13 2021 +0000
  
      Split the XPVHV body into two variants "normal" and "with aux"
      
      Default to the smaller body, and switch to the larger body if we need to
      allocate a C<struct xpvhv_aux> (eg need an iterator).
      
      This restores the previous small size optimisation for hashes used as
      objects.
  
  :100644 100644 ff3729fa5f176adffacb7d2088707fab601782a7 749ede6f5b5d2998a17d18bdabdadc5fb96fa638 M      embed.fnc
  :100644 100644 954eb493d651ab12dcafcf6d6df670b2ba19bcbd 7e8f286859f3cb04dbe2300c1a63d69cba7f56c6 M      hv.c
  :100644 100644 285a004a5f4c6be21370557ff03ac0fd0da5e9b2 c7aeb7de572631a42181ee2056ce6f815114129f M      hv.h
  :100644 100644 0fe96719b0252434423253fc432da2b5b1da1acc 7b65b1390cfe3742f05c2529b8b574e3aa5a39d2 M      proto.h
[...]

Issue #19214, opened 2021-10-28
sv.c: 94ee6ed79db introduced possible code maintainability issue


jkeenan:
Two months back we made changes to `sv.c`, as reported by `git blame`:
```
79072805bf6 (Larry Wall                1993-10-07 23:00:00 +0000  6678) void
5239d5c4bfd (David Mitchell            2010-10-08 16:22:42 +0100  6679) Perl_sv_clear(pTHX_ SV *const orig_sv)
79072805bf6 (Larry Wall                1993-10-07 23:00:00 +0000  6680) {
...
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6984)         if (type == SVt_PVHV && SvOOK(sv)) {
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6985)             arena_index = HVAUX_ARENA_ROOT_IX;
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6986)             sv_type_details = &fake_hv_with_aux;
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6987)         }
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6988)         else {
94ee6ed79db (Nicholas Clark            2021-08-20 15:20:13 +0000  6989)             arena_index = type;
```
We recently received a Coverity scan report which reported a possible code maintainability issue for this stretch of code:
```
*** CID 340738:  Code maintainability issues  (UNUSED_VALUE)
/sv.c: 6985 in Perl_sv_clear()
6979           free_body:
6980     
6981             SvFLAGS(sv) &= SVf_BREAK;
6982             SvFLAGS(sv) |= SVTYPEMASK;
6983     
6984             if (type == SVt_PVHV && SvOOK(sv)) {
>>>     CID 340738:  Code maintainability issues  (UNUSED_VALUE)
>>>     Assigning value "SVt_IV" to "arena_index" here, but that stored value is overwritten before it can be used.
6985                 arena_index = HVAUX_ARENA_ROOT_IX;
6986                 sv_type_details = &fake_hv_with_aux;
6987             }
6988             else {
6989                 arena_index = type;
6990                 sv_type_details = bodies_by_type + arena_index;
```
@nwc10, can you take a look?

Thank you very much.
Jim Keenan

hvds:
[...]

PR #19224, opened 2021-11-09
Return the "hv with aux" PVHV bodies to the correct arena

nwc10:
PVHV body handling was changed recently by commit 94ee6ed79dbca73d:

```
Split the XPVHV body into two variants "normal" and "with aux"

Default to the smaller body, and switch to the larger body if we need to
allocate a C<struct xpvhv_aux> (eg need an iterator).

This restores the previous small size optimisation for hashes used as
objects.
```

as part of changing where memory for `struct xpvhv_aux` is allocated.

The new implementation uses two sizes of "bodies" for PVHVs - the default body is unchanged, but when a hash needs a `struct xpvhv_aux`, it now swaps out the default body for a larger body with space for the struct.

(Previously it would reallocate the memory used for HvARRAY() to make space for the struct there - an approach which requires more tracking of changed pointers.)

That commit was subtly buggy in that on hash deallocation it didn't *return* hash bodies to the correct arena. As-was, it would return both sizes of hash body to the arena for the default (smaller) body. This was due to a combination of two subtle bugs

1) the explicit reset of all SvFLAGS to SVf_BREAK would implicitly clear SVf_OOK. That flag bit set is needed to signal the different body size

2) The code must call del_body() with &PL_body_roots[arena_index], not &PL_body_roots[type] type is SVt_PVHV for both body sizes. arena_index is SVt_PVHV for the smaller (default) bodies, and HVAUX_ARENA_ROOT_IX for the larger bodies.

There were no memory errors or leaks (that would have been detectable by tools such as ASAN or valgrind), but the bug meant that larger hash bodies would never get re-used. So for code that was steady-state creating and destroying hashes with iterators (or similar), instead of those bodies being recycled via their correct arena, the interpreter would need to continuously allocate new memory for that arena, whilst accumulating ever more unused "smaller" bodies in the other arena.

I didn't spot this bug whilst reviewing commit 8461dd8ad90b2bce:

```
don't overwrite the faked up type details for hv-with-aux

CID 340472
```

which is in the same area of the code, but fixes a related problem.

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -2231,18 +2231,15 @@ PERL_STATIC_INLINE U32 S_ptr_hash(PTRV u) {
 static struct xpvhv_aux*
 S_hv_auxinit(pTHX_ HV *hv) {
     struct xpvhv_aux *iter;
-    char *array;
 
     PERL_ARGS_ASSERT_HV_AUXINIT;
 
     if (!SvOOK(hv)) {
-        if (!HvARRAY(hv)) {
+        char *array = (char *) HvARRAY(hv);
+        if (!array) {
             Newxz(array, PERL_HV_ARRAY_ALLOC_BYTES(HvMAX(hv) + 1), char);
-        } else {
-            array = (char *) HvARRAY(hv);
-            Renew(array, PERL_HV_ARRAY_ALLOC_BYTES(HvMAX(hv) + 1), char);
+            HvARRAY(hv) = (HE**)array;
         }
-        HvARRAY(hv) = (HE**)array;
         iter = Perl_hv_auxalloc(aTHX_ hv);
 #ifdef PERL_HASH_RANDOMIZE_KEYS
         if (PL_HASH_RAND_BITS_ENABLED) {
--- a/hv.c
+++ b/hv.c
@@ -2073,21 +2073,15 @@ Perl_hv_undef_flags(pTHX_ HV *hv, U32 flags)
     /* SvOOK() is true for a hash if it has struct xpvhv_aux allocated. That
        structure has several other pieces of allocated memory - hence those must
        be freed before the structure itself can be freed. Some can be freed when
-       a hash is "undefined", but some must persist until it is destroyed.
+       a hash is "undefined" (this function), but some must persist until it is
+       destroyed (which might be this function's immediate caller).
 
-       Until recently (commit 53a41f9c2c0671d8) the memory for struct xpvhv_aux
-       was allocated as part of the block of memory accessed through
-       HvARRAY(hv). This code predates that, and so assumes that HvARRAY(hv)
-       itself cannot actually be freed if the "aux" structure still has live
-       pointers. Hence the code in this block frees what it is logical to free
-       (and NULLs out anything freed), but only clears SVf_OOK if all pointers
-       in the aux structure are now NULL.
-
-       Independent of whether it clears that flag, the structure is left in a
-       logically consistent state - pointers are NULL or point to valid memory,
-       and non-pointer values are correct for an empty hash. The structure state
-       must remain consistent, because if the code below does not clear SVf_OOK
-       then it is still accessible, so might be read again at any point in the
+       Hence the code in this block frees what it is logical to free (and NULLs
+       out anything freed) so that the structure is left in a logically
+       consistent state - pointers are NULL or point to valid memory, and
+       non-pointer values are correct for an empty hash. The structure state
+       must remain consistent, because this code can no longer clear SVf_OOK,
+       meaning that this structure might be read again at any point in the
        future without further checks or reinitialisation. */
     if (SvOOK(hv)) {
       struct mro_meta *meta;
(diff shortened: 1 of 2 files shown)


nwc10:
fixes #19223 and #19214

nwc10:
The cygwin CI check has failed identically on #19225 which is purely a documentation patch. Is github's Cygwin setup dodgy (again?) or is there something bad in blead?

(Either way, this commit doesn't **actually** cause a cygwin regression, even if it looks like it might)

nwc10:
I rebased my branch `larger-hashes-without-shared-keys ` onto the 3 commits of this branch, because these bug fixes felt like a necessary pre-requisite for the performance related changes on that branch. We're reviewing that entire branch as #19208 - maybe it's better to keep all the discussion there.

However (if I have it correct now) - this branch should fix the memory exhaustion without any other changes, so it remains useful.

nwc10:
I believe that I replicated the problem locally, and hence have resolved it.

The problem was that this "fix" (which I shall inline completely, as it will vanish from history soon):

```
commit 3022d46b2c8eeb6aaa609e8e61e84e05e09a8de4
Date:   Tue Nov 9 09:24:06 2021 +0000

    Return the "hv with aux" PVHV bodies to the correct arena

    PVHV body handling was changed recently by commit 94ee6ed79dbca73d:
        Split the XPVHV body into two variants "normal" and "with aux"

        Default to the smaller body, and switch to the larger body if we need to
        allocate a C<struct xpvhv_aux> (eg need an iterator).

        This restores the previous small size optimisation for hashes used as
        objects.

    as part of changing where memory for C<struct xpvhv_aux> is allocated.

    The new implementation uses two sizes of "bodies" for PVHVs - the default
    body is unchanged, but when a hash needs a C<struct xpvhv_aux>, it now
    swaps out the default body for a larger body with space for the struct.

    (Previously it would reallocate the memory used for HvARRAY() to make space
    for the struct there - an approach which requires more tracking of changed
    pointers.)

    That commit was subtly buggy in that on hash deallocation it didn't *return*
    hash bodies to the correct arena. As-was, it would return both sizes of hash
    body to the arena for the default (smaller) body. Hence there were no memory
    errors or leaks (that would have been detectable by tools such as ASAN or
    valgrind), but the bug meant that larger hash bodies would never get
    re-used. So for code that was steady-state creating and destroying hashes
    with iterators (or similar), instead of those bodies being recycled via
    their correct arena, the interpreter would need to continuously allocate
    new memory for that arena, whilst accumulating ever more unused "smaller"
    bodies in the other arena.

    I didn't spot this bug whilst reviewing commit 8461dd8ad90b2bce:
        don't overwrite the faked up type details for hv-with-aux

        CID 340472

    which is in the same area of the code, but fixes a related problem.

    Ironically, defining PURIFY would hide this bug, because that disables the
    use of arenas, instead allocating all bodies explicitly with malloc() and
    releasing them with free(), and free() doesn't need to know the size of the
    memory block, hence it wouldn't matter that the code had that size wrong.

diff --git a/sv.c b/sv.c
index 1b3817e7dd..ad17cbfb68 100644
--- a/sv.c
+++ b/sv.c
@@ -6777,8 +6777,6 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)

     while (sv) {
         U32 type = SvTYPE(sv);
-        U32 arena_index;
-        const struct body_details *sv_type_details;
         HV *stash;

         assert(SvREFCNT(sv) == 0);
@@ -7066,6 +7064,9 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
         SvFLAGS(sv) &= SVf_BREAK;
         SvFLAGS(sv) |= SVTYPEMASK;

+        U32 arena_index;
+        const struct body_details *sv_type_details;
+
         if (type == SVt_PVHV && SvOOK(sv)) {
             arena_index = HVAUX_ARENA_ROOT_IX;
             sv_type_details = &fake_hv_with_aux;
@@ -7076,7 +7077,7 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
         }
         if (sv_type_details->arena) {
             del_body(((char *)SvANY(sv) + sv_type_details->offset),
-                     &PL_body_roots[type]);
+                     &PL_body_roots[arena_index]);
         }
         else if (sv_type_details->body_size) {
             safefree(SvANY(sv));
```

was not sufficient.

The correct fix is now commit da26e4ea186b7bd3, which

1) moves the assignments to SvFLAGS(sv) after the check for SvOOK()
2) uses `arena_index` for the array index to `PL_body_roots`

The "incorrect fix" (the commit inlined just above) didn't do the first, meaning that `SvOOK(sv)` was always false, and hence the code `arena_index = HVAUX_ARENA_ROOT_IX` was always unreachable.

tonycoz:
> The cygwin CI check has failed identically on #19225 which is purely a documentation patch. Is github's Cygwin setup dodgy (again?) or is there something bad in blead?

There's two cygwin issues, one appears to be a bug in Win32.pm or its tests, while the other appears to be a buffer overflow bug in cygwin:

https://cygwin.com/pipermail/cygwin/2021-November/249930.html

PR #19225, opened 2021-11-09
Document `local *STDOUT` as an alternative to `select ... select`

nwc10:
If one needs to temporarily "capture" the output of `print`, for example as part of a unit test, the classic "obvious" approach has been to use `select` before and afterwards, to set a different default output handle and the restore the previous handle.

If one is already in the mindset of `select`, it might not be obvious that `local *STDOUT` is alternative way to achieve the underlying goal, and might be clearer and more robust. Hence add it as a suggestion in the `select` documentation.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -7130,6 +7130,31 @@ methods, preferring to write the last example as:
 (Prior to Perl version 5.14, you have to C<use IO::Handle;> explicitly
 first.)
 
+Whilst you can use C<select> to temporarily "capture" the output of
+C<print> like this:
+
+    {
+        my $old_handle = select $new_handle;
+        
+        # This goes to $new_handle:
+        print "ok 1\n";
+        ...
+        
+        select $old_handle;
+    }
+
+you might find it easier to localize the typeglob instead:
+
+    {
+        local *STDOUT = $new_handle;
+        
+        print "ok 1\n";
+        ...
+    }
+
+The two are not exactly equivalent, but the latter might be clearer, and will
+restore STDOUT if the wrapped code dies.
+
 Portability issues: L<perlport/select>.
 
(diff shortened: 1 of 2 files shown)


pod/perlfunc.pod
```diff
@@ -7130,6 +7130,31 @@ methods, preferring to write the last example as:
...
+you might find it easier to localize the typeglob instead:
+
+    {
+        local *STDOUT = $new_handle;
+        
+        print "ok 1\n";
+        ...
+    }
+
+The two are not exactly equivalent, but the latter might be clearer, and will
```
Grinnz, line 7155: Perhaps mention the specific difference in usability: with the select method, one can still `print STDOUT $str` but replacing the superglobal means the original STDOUT can no longer be used within that (dynamic) scope.

khwilliamson: +1

leonerd:
I have fixed up the merge conflict and also added more words about the difference in my own branch:

https://github.com/leonerd/perl5/tree/local-STDOUT-instead-of-select

jkeenan:
> I have fixed up the merge conflict and also added more words about the difference in my own branch:
> 
> https://github.com/leonerd/perl5/tree/local-STDOUT-instead-of-select

We apparently have two different branches here, one of which is the branch beneath this pull request (and which has merge conflicts), and one which does not yet have a pull request attached.

@nwc10, @LeoNerd, how shall we proceed?

leonerd:
> @nwc10, @leonerd, how shall we proceed?

I don't have enough energy to form an opinion. Tell users something useful otherwise they won't know. Beyond that I don't really mind.

jkeenan:
I've done my best to reconcile the differences between these two branches and have pushed the result in 4d4eaa2cb3.

Still not satisfied?  Open a new p.r.; closing this ticket.

PR #18783, opened 2021-05-11
Fix compilation with MinGW.org

xenu:
Fixes #18510

PR #19226, opened 2021-11-09
_MSC_VER is now always >= 1800 (Visual C++ 2013 / Visual C++ 12.0)

steve-m-hay:


steve-m-hay:
In testing this, I stumbled across #18783, which we probably ought to resolve one way or another too. FWIW, I'm fine with dropping support for the mingw.org compilers.
Two commits on that PR (a644daa and 9994e22) were identified as useful even if we drop mingw.org support. The first of those can be simplified in line with this PR if this PR is approved.

win32/win32.h
```diff
@@ -647,12 +629,6 @@ typedef struct {
     CRITICAL_SECTION lock;
 /* this struct definition breaks ABI compatibility with
  * not using, cl.exe's native VS version specitfic CRT. */
-#  if _MSC_VER >= 1400 && _MSC_VER < 1500
-#    error "This ioinfo struct is incomplete for Visual C 2005"
-#  endif
-/* VS2005 CRT has at least 3 different definitions of this struct based on the
- * CRT DLL's build number. */
-#  if _MSC_VER >= 1500
 #    ifndef _SAFECRT_IMPL
```
smpetershm, line 632: Does this need some re-indentation here?

sisyphus:
One thing I notice is that the removal of <code>#  if _MSC_VER >= 1500</code> at line 655 of win32.h makes the <code>#    ifndef _SAFECRT_IMPL</code> block available to mingw-w64 compilers - unless there's some other encompassing preprocessor block that is excluding mingw-w64 compilers. 
I don't know if that matters.
I'll also point out that the io/socket.t failure reported for this PR is one that I don't see with my own mingw-w64 builds of blead.

steve-m-hay:
> One thing I notice is that the removal of `# if _MSC_VER >= 1500` at line 655 of win32.h makes the `# ifndef _SAFECRT_IMPL` block available to mingw-w64 compilers - unless there's some other encompassing preprocessor block that is excluding mingw-w64 compilers. I don't know if that matters. I'll also point out that the io/socket.t failure reported for this PR is one that I don't see with my own mingw-w64 builds of blead.

Good spot. I will rework this section a little bit and update the PR. We're inside a section marked "#if (!defined(_MSC_VER)) || (defined(_MSC_VER) && _MSC_VER < 1900)", which now basically means "if gcc or _MSC_VER==1800". So WIN32_DYN_IOINFO_SIZE is only defined for _MSC_VER 1800 and the "#ifndef WIN32_DYN_IOINFO_SIZE" block only applies to gcc now so the part that was for _MSC_VER >= 1500 can be removed.

perl.h
```diff
@@ -543,11 +543,6 @@ __typeof__ and nothing else.
 #  endif
 #endif
 
-#if defined(_MSC_VER) && _MSC_VER < 1400
-/* XXX older MSVC versions have a smallish macro buffer */
-#  define PERL_SMALL_MACRO_BUFFER
-#endif
-
```
tonycoz, line 550: It should be possible to eliminate all PERL_SMALL_MACRO_BUFFER references, since the macro was added to support older MSVC and it isn't defined anywhere else.

@khw might have an opinion here.

khwilliamson: I think I oppose removal of this at this time.  I think we may find it useful

khwilliamson:
xenu left his approval, so I'm merging

PR #19230, opened 2021-11-17
Obituary for David H Adler

jkeenan:


pod/perldelta.pod
```diff
@@ -414,8 +414,12 @@ the perldelta of a previous release.
 
 =head1 Obituary
 
-XXX If any significant core contributor or member of the CPAN community has
-died, add a short obituary here.
+David H. Adler (DHA) passed away on November 16, 2021.  David co-founded the
```
briandfoy: I think I'd say something like "in 1997 co-founded NY.pm, the first Perl users group, and in 1998 co-founded Perl mongers".

He may have a had in organizing the New York City Perl Meetup, but I wasn't a part of that effort so my memories are fuzzy on that. Charlie Gonzales might know.

Maybe mention "frobisher" too, since that might be how many people recognize him online.

He also administered the White Camel Awards for several years.

pod/perldelta.pod
```diff
@@ -414,8 +414,13 @@ the perldelta of a previous release.
 
 =head1 Obituary
 
-XXX If any significant core contributor or member of the CPAN community has
-died, add a short obituary here.
+David H. Adler (DHA) passed away on November 16, 2021.  David co-founded 
+New York Perl Mongers, the first chapter of what became a worldwide
```
briandfoy: "In 1997, David co-founded NY.pm, the first Perl users group, and in 1998 cofounded Perl mongers to help establish other user groups across the globe".

It wasn't really Perl Mongers at first. It was NY.pm and then he had to figure out what the .pm would stand for. I'm not really sure how much longer after we had NY.pm that we settled on "Perl mongers". It was a regex for awhile too.

briandfoy (approved):

PR #19231, opened 2021-11-19
t/op/pack.t: Get it to pass under 'make minitest'

jkeenan:
During preparations for release of perl-5.35.6 we noticed that
t/op/pack.t was failing during 'make minitest', though passing under
'make test'.  Certain tests require Math::BigInt, which in turn now
requires Scalar::Util.  The latter is not available under miniperl.  So
we skip certain tests.

Also, Math::BigInt is now under cpan/, not dist/.

Insert one 'note' so that we can see where we are in the file.  Minor
whitespace improvements.

### DIFF
--- a/t/op/pack.t
+++ b/t/op/pack.t
@@ -3,7 +3,7 @@
 BEGIN {
     chdir 't' if -d 't';
     require './test.pl'; require './charset_tools.pl';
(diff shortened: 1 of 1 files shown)


richardleach (approved):
Thanks for fixing this up. LGTM.

PR #19232, opened 2021-11-20
Boolean builtins

leonerd:
Still todo before merge:

- [ ] - mention the module from somewhere else in docs

Still todo sometime:

* Infrastructure for applying callchecker to replace calls to these functions with efficient optrees
* Efficient op version of `isbool`

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -1,0 +1,103 @@
+/*    builtin.c
+ *
+ *    Copyright (C) 2021 by Paul Evans and others
+ *
+ *    You may distribute under the terms of either the GNU General Public
+ *    License or the Artistic License, as specified in the README file.
+ *
+ */
+
+/* This file contains the code that implements functions in perl's "builtin::"
+ * namespace
+ */
+
+#include "EXTERN.h"
+#include "perl.h"
+
+#include "XSUB.h"
+
+XS(XS_builtin_true);
+XS(XS_builtin_true)
+{
+    dXSARGS;
+    if(items)
+        croak_xs_usage(cv, "");
+    XSRETURN_YES;
+}
+
+XS(XS_builtin_false);
+XS(XS_builtin_false)
+{
+    dXSARGS;
+    if(items)
+        croak_xs_usage(cv, "");
+    XSRETURN_NO;
+}
+
+XS(XS_builtin_isbool);
+XS(XS_builtin_isbool)
+{
+    dXSARGS;
+    if(items != 1)
+        croak_xs_usage(cv, "sv");
+
+    SV *sv = ST(0);
+    if(SvIsBOOL(sv))
+        XSRETURN_YES;
+    else
+        XSRETURN_NO;
+}
+
+XS(XS_builtin_import);
+XS(XS_builtin_import)
+{
+    dXSARGS;
+
+    if(!PL_compcv)
+        Perl_croak(aTHX_
+                "builtin::import can only be called at compiletime");
+
+    /* We need to have PL_comppad / PL_curpad set correctly for lexical importing */
+    ENTER;
+    SAVESPTR(PL_comppad_name); PL_comppad_name = PadlistNAMES(CvPADLIST(PL_compcv));
+    SAVESPTR(PL_comppad);      PL_comppad      = PadlistARRAY(CvPADLIST(PL_compcv))[1];
+    SAVESPTR(PL_curpad);       PL_curpad       = PadARRAY(PL_comppad);
+
+    for(int i = 1; i < items; i++) {
+        SV *sym = ST(i);
+        if(strEQ(SvPV_nolen(sym), "import")) goto unavailable;
+
+        SV *ampname = sv_2mortal(Perl_newSVpvf(aTHX_ "&%" SVf, SVfARG(sym)));
+        SV *fqname  = sv_2mortal(Perl_newSVpvf(aTHX_ "builtin::%" SVf, SVfARG(sym)));
+
+        CV *cv = get_cv(SvPV_nolen(fqname), SvUTF8(fqname) ? SVf_UTF8 : 0);
+        if(!cv) goto unavailable;
+
+        PADOFFSET off = pad_add_name_sv(ampname, padadd_STATE, 0, 0);
+        SvREFCNT_dec(PL_curpad[off]);
+        PL_curpad[off] = SvREFCNT_inc(cv);
+        continue;
+
+unavailable:
+        Perl_croak(aTHX_
+                "'%" SVf "' is not recognised as a builtin function", sym);
+    }
+
+    intro_my();
+
+    LEAVE;
+}
+
+void
+Perl_boot_core_builtin(pTHX)
+{
+    newXS_flags("builtin::true",   &XS_builtin_true,   __FILE__, NULL, 0);
+    newXS_flags("builtin::false",  &XS_builtin_false,  __FILE__, NULL, 0);
+    newXS_flags("builtin::isbool", &XS_builtin_isbool, __FILE__, NULL, 0);
+
+    newXS_flags("builtin::import", &XS_builtin_import, __FILE__, NULL, 0);
+}
+
+/*
+ * ex: set ts=8 sts=4 sw=4 et:
+ */
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,0 +1,102 @@
+package builtin 0.001;
+
+use strict;
+use warnings;
+
+# All code, including &import, is implemented by always-present functions in
+# the perl interpreter itself.
+# See also `builtin.c` in perl source
+
+1;
+__END__
+
+=head1 NAME
+
+builtin - Perl pragma to import built-in utility functions
+
+=head1 SYNOPSIS
+
+    use builtin qw( true false isbool );
+
+=head1 DESCRIPTION
+
+Perl provides several utility functions in the C<builtin> package. These are
+plain functions, and look and behave just like regular user-defined functions
+do. They do not provide new syntax or require special parsing. These functions
+are always present in the interpreter and can be called at any time by their
+fully-qualified names. By default they are not available as short names, but
+can be requested for convenience.
+
+Individual named functions can be imported by listing them as import
+parameters on the C<use> statement for this pragma.
+
+=head2 Lexical Import
+
+This pragma module creates I<lexical> aliases in the currently-compiling scope
+to these builtin functions. This is similar to the lexical effect of other
+pragmas such as L<strict> and L<feature>.
+
+    sub classify
+    {
+        my $sv = shift;
+
+        use builtin 'isbool';
+        return isbool($sv) ? "boolean" : "not a boolean";
+    }
+
+    # the isbool() function is no longer visible here
+    # but may still be called by builtin::isbool()
+
+Because these functions are imported lexically, rather than by package
+symbols, the user does not need to take any special measures to ensure they
+don't accidentally appear as object methods from a class.
+
+    package An::Object::Class {
+        use builtin 'true', 'false';
+        ...
+    }
+
+    # does not appear as a method
+    An::Object::Class->true;
+
+    # Can't locate object method "true" via package "An::Object::Class"
+    #   at ...
+
+=head1 FUNCTIONS
+
+=head2 true
+
+    $val = true;
+
+Returns the boolean truth value. While any scalar value can be tested for
+truth and most defined, non-empty and non-zero values are considered "true"
+by perl, this one is special in that L</isbool> considers it to be a
+distinguished boolean value.
+
+This gives an equivalent value to expressions like C<!!1> or C<!0>.
+
+=head2 false
+
+    $val = false;
+
+Returns the boolean fiction value. While any non-true scalar value is
+considered "false" by perl, this one is special in that L</isbool> considers
+it to be a distinguished boolean value.
+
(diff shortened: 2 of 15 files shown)


leonerd:
All the CIs are happy apart from Cygwin, but Cygwin looks to be unhappy about something Unicode-related, not because of this PR.

lib/builtin.t
```diff
@@ -0,0 +1,39 @@
...
+    ok(!isbool(undef), 'undef is not bool');
+    ok(!isbool(1), '1 is not bool');
+    ok(!isbool(""), 'empty is not bool');
+
+    my $truevar  = (5 == 5);
+    my $falsevar = (5 == 6);
+
+    ok(isbool($truevar), '$truevar is bool');
+    ok(isbool($falsevar), '$falsevar is bool');
+}
```
gugod, line 33: Since the result of `isbool` is also a boolean, shall we include these tests as proofs ?

```perl

ok( isbool(isbool(true)), 'isbool() returns a bool' );
ok( isbool(isbool(false)), 'isbool() returns a bool' );
ok( isbool(isbool(undef)), 'isbool() returns a bool' );
ok( isbool(isbool(1)), 'isbool() returns a bool' );
ok( isbool(isbool("")), 'isbool() returns a bool' );
ok( isbool(isbool($truevar)), 'isbool() returns a bool' );
ok( isbool(isbool($false)), 'isbool() returns a bool' );
```

leonerd: Ahyes, a good idea. I don't think it's necessary to test all the possible combinations there, I'll just do one true and one false case.

Added in a4837515be

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+
+Returns the boolean truth value. While any defined but non-zero scalar value
+is considered "true" by perl, this one is special in that L</isbool> considers
+it to be a distinguished boolean value.
+
+=head2 false
+
+    $val = false;
+
+Returns the boolean fiction value. While non-true scalar value is considered
```
wolfsage: `While non-true scalar value is considered` feels like a typo. Maybe `While any non-true scalar value is considered`

wolfsage (commented):
Docs seem good to me. One thing they don't seem to have is a rationale though "Why do these three methods exist?!" Maybe that's fine because it's obvious (to us) but it might not be obvious to newcomers.

lib/builtin.pm
leonerd: That it is. Shall fix.

leonerd:
@wolfsage
> One thing they don't seem to have is a rationale though "Why do these three methods exist?!" Maybe that's fine because it's obvious (to us) but it might not be obvious to newcomers.

There's two parts to that question. Why are each of these present, or why *only* these three?

As to the presence of each of them: They're doing something you can't do any other way in Perl currently, short of adding some utility functions to `Scalar::Util` et.al., and we don't want to keep doing that ad. infinitum.

As to why *only* these three: There won't be only these for very long. Once the base mechanism itself is merged I plan to have a good attack at many more of the functions mentioned in RFC 0009, so by the time it comes to release we'll have somewhere between 10 and 20 of them with any luck.

wolfsage:
Ah sorry, I wasn't clear in my comment. I think having the rationale somewhere in the documentation might be a good idea

builtin.c
```diff
@@ -0,0 +1,105 @@
...
+    ENTER;
+    SAVESPTR(PL_comppad_name); PL_comppad_name = PadlistNAMES(CvPADLIST(PL_compcv));
+    SAVESPTR(PL_comppad);      PL_comppad      = PadlistARRAY(CvPADLIST(PL_compcv))[1];
+    SAVESPTR(PL_curpad);       PL_curpad       = PadARRAY(PL_comppad);
+
+    for(i = 1; i < items; i++) {
+        SV *sym = ST(i);
+        if(strEQ(SvPV_nolen(sym), "import")) goto unavailable;
+
+        /* TODO: I've no idea why the newSVpvf macro isn't visible here */
```
wolfsage: Odd, this compiles fine for me with the short versions

leonerd: Weird. When I try it I get:

```
In file included from perl.h:5744,
                 from builtin.c:15:
builtin.c: In function ‘XS_builtin_import’:
builtin.c:71:34: warning: implicit declaration of function ‘newSVpvf’; did you mean ‘newSVpv’? [-Wimplicit-function-declaration]
   71 |         SV *ampname = sv_2mortal(newSVpvf("&%" SVf, SVfARG(sym)));
      |                                  ^~~~~~~~
embed.h:578:46: note: in definition of macro ‘sv_2mortal’
  578 | #define sv_2mortal(a)  Perl_sv_2mortal(aTHX_ a)
      |                                              ^
builtin.c:71:34: warning: passing argument 2 of ‘Perl_sv_2mortal’ makes pointer from integer without a cast [-Wint-conversion]
   71 |         SV *ampname = sv_2mortal(newSVpvf("&%" SVf, SVfARG(sym)));
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                  |
      |                                  int
...
cc -fstack-protector-strong -L/usr/local/lib -o miniperl \
    opmini.o perlmini.o universalmini.o  gv.o toke.o perly.o pad.o regcomp.o dump.o util.o mg.o reentr.o mro_core.o keywords.o builtin.o hv.o av.o run.o pp_hot.o sv.o pp.o scope.o pp_ctl.o pp_sys.o doop.o doio.o regexec.o utf8.o taint.o deb.o globals.o perlio.o numeric.o mathoms.o locale.o pp_pack.o pp_sort.o caretx.o dquote.o time64.o  miniperlmain.o  -lpthread -lnsl -ldl -lm -lcrypt -lutil -lc 
/usr/bin/ld: builtin.o: in function `XS_builtin_import':
/home/leo/src/bleadperl/perl/builtin.c:71: undefined reference to `newSVpvf'
/usr/bin/ld: /home/leo/src/bleadperl/perl/builtin.c:72: undefined reference to `newSVpvf'
collect2: error: ld returned 1 exit status
```

leonerd: It seems that `embed.h` defines `newSVpvf` only under non-MULTIPLICITY builds:

```c
 394 #ifndef MULTIPLICITY
 395 #define newSVpvf                Perl_newSVpvf
 396 #endif
 397 #define newSVpvn(a,b)           Perl_newSVpvn(aTHX_ a,b)
 398 #define newSVpvn_flags(a,b,c)   Perl_newSVpvn_flags(aTHX_ a,b,c)
 399 #define newSVpvn_share(a,b,c)   Perl_newSVpvn_share(aTHX_ a,b,c)
```

leonerd:
> Ah sorry, I wasn't clear in my comment. I think having the rationale somewhere in the documentation might be a good idea

Sure, that sounds good. Any suggestion on how to word it though?

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+
+builtin - Perl pragma to import built-in utility functions
+
+=head1 SYNOPSIS
+
+    use builtin qw( true false isbool );
+
+=head1 DESCRIPTION
+
+Perl provides several utility functions in a package called C<builtin::>.
```
Grinnz: Minor but the package name is `builtin` whereas `builtin::` could be referred to as the namespace, and it's not only called that but is that!

```suggestion
Perl provides several utility functions in the package namespace C<builtin::>.
```

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+Functions in this namespace can be called by their fully-qualified names, but
+it is often desired to give them shorter name aliases for convenience.
```
Grinnz: This seems like a weird way to say they can be imported, as it suggests aliasing to a different name which isn't implemented. Perhaps can be combined with the following paragraph

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+
+Individual named functions can be imported by listing them as import
+parameters on the C<use> statement for this pragma.
+
+=head2 Lexical Import
+
+This pragma module creates I<lexical> aliases in the currently-compiling scope
+to these builtin functions.
```
Grinnz: Might be helpful context to mention that it operates similarly to keywords provided by the feature pragma.

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+    # Can't locate object method "true" via package "An::Object::Class"
+    #   at ...
+
+=head1 FUNCTIONS
+
+=head2 true
+
+    $val = true;
+
+Returns the boolean truth value. While any defined but non-zero scalar value
```
Grinnz: non-zero may be a bit broad, as '00' and '0e0' are true values (not to mention '0 but true').

lib/builtin.pm
```diff
@@ -0,0 +1,93 @@
...
+it to be a distinguished boolean value.
+
+=head2 isbool
+
+    $bool = isbool($val);
+
+Returns true when given a distinguished boolean value, or false if not. A
+distinguished boolean value is the result of any boolean-returning builtin
+function (such as C<true> or C<isbool> itself), boolean-returning operator
+(such as the C<eq> or C<==> comparison tests), or any variable containing one
```
Grinnz: Also probably good to mention `!` here

lib/builtin.pm
leonerd: And also `""` is non-zero, yet false. Perhaps I won't bother trying to define all that here and just word it differently.

leonerd:
Thanks @Grinnz; docs updated.

Still needs an overall introduction paragraph or two to explain the "builtin" idea though.

leonerd:
Actually on further thought, I think it will become easier to explain the module as a whole later, once it has more functions because we'll have further expanded it as part of RFC 0009. For now it doesn't need that.

builtin.c
```diff
@@ -0,0 +1,104 @@
...
+        XSRETURN_YES;
+    else
+        XSRETURN_NO;
+}
+
+XS(XS_builtin_import);
+XS(XS_builtin_import)
+{
+    dXSARGS;
+    int i;
```
xenu: You can declare `i` inside `for()`.

lib/builtin.pm
```diff
@@ -0,0 +1,95 @@
...
+=head1 FUNCTIONS
+
+=head2 true
+
+    $val = true;
+
+Returns the boolean truth value. While any scalar value can be tested for
+truth and most defined, non-empty and non-zero values are considered "true"
+by perl, this one is special in that L</isbool> considers it to be a
+distinguished boolean value.
```
xenu, line 74: I think it would make sense to mention that `true` and `false` are completely equivalent to `!!1` and `!!0` respectively.

builtin.c
leonerd: Ahyes, I forget that we're now in the lovely modern(!) world of C99. Such heady heights...

leonerd: Done 1a1f12ffb0

wolfsage:
> > Ah sorry, I wasn't clear in my comment. I think having the rationale somewhere in the documentation might be a good idea
> 
> Sure, that sounds good. Any suggestion on how to word it though?

Here's a rough draft, best I got. It'd require restructuring the pod a bit so we can organize builtins by category.
```
=head1 FUNCTIONS

=head2 Boolean operations

Historically Perl hasn't had the concept of real boolean types for true
and false.

Because of this, encoding/decoding Perl data structures into other formats
(like JSON) can have surprising results, or require workarounds to handle
true and false properly (see L<JSON::PP/"JSON::PP::is_bool"> for example).

The functions below allow us to natively create boolean scalars and examine
existing scalars, which we can then use to make encoding systems more
predictable.

=head3 true

=head3 false

=head3 isbool

```

Grinnz:
I think that builtin.pm should have a `$VERSION`. Potentially one could `use builtin X.Y 'foo'` if knowing the version that is available, but the larger use would be if the module is ever possible to make dual-life.

Grinnz:
(note that even strict, warnings, and feature have versions, so it is a precedent)

leonerd:
@Grinnz Ahyes, definitely wants a version number. Added bbadb115eb

And yes, the plan is to offer *a* version of the module as dual-life, from CPAN, as a polyfil for older perls - at least, as far back at 5.18 where lexical subs become possible.

lib/builtin.pm
```diff
@@ -1,4 +1,4 @@
-package builtin;
+package builtin v0.001;
```
Grinnz: If using a v-string version it would be written `v0.1` or ideally `v0.1.0` as leading zeroes in this form are insignificant.

leonerd: Fair point. I'll force-push an amend,

leonerd: Turns out: a v-string requires 3 components to it or else it gets upset. I've instead written it as just `0.001`

win32/GNUmakefile
```diff
@@ -945,6 +945,7 @@ MICROCORE_SRC	=		\
 		..\dump.c	\
 		..\hv.c		\
 		..\av.c		\
+		..\builtin.c	\
 		..\caretx.c	\
```
nwc10, line 949: Adding a new C source file requires similar "cargo cult" in `vms/descrip_mms.template`

(`Makefile.SH`, both win32 makefiles, `vms/descrip_mms.template` and maybe I've missed somewhere else obscure)

shlomif:
FWIW, the changes look fine to me.

PR #19236, opened 2021-11-22
Minor output tweaks to sync-with-cpan

richardleach:
Two minor text changes:
* Explain that _minitest_ should be run in addition to full tests
* Provide the typical commit message format


jkeenan:
@bingos, since you're a heavy user of `sync-with-cpan`, could you take a look at this p.r.?

Thank you very much.
Jim Keenan

khwilliamson:
Since @bingos hasn't replied, I reviewed this P.R. think it is an improvement, and should be merged.

PR #19237, opened 2021-11-22
Forbid using 'goto' to jump into a 'defer' block

leonerd:
Otherwise, once the `defer` block is done the main body of the program will abruptly end, because the `defer`'s optree ends in a NULL `->op_next` pointer.

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -1042,6 +1042,11 @@ error occurs in cases such as these:
     goto G;
     $x + do { G: $y }; # How is + supposed to get its first operand?
 
+=item Can't "goto" into a defer block
+
+(F) A "goto" statement was executed to jump into the scope of a defer
+block.  This is not permitted.
+
 =item Can't "goto" into a "given" block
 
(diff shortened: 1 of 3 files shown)


pod/perldiag.pod
```diff
@@ -1042,6 +1042,11 @@ error occurs in cases such as these:
     goto G;
     $x + do { G: $y }; # How is + supposed to get its first operand?
 
+=item Can't "goto" into a defer block
```
ilmari, line 1045: Should `defer` be quoted, for consistency with `"given"` below?

leonerd: I did think about that, but then the other message doesn't have quotes:

```
=item Can't "%s" out of a defer block
```

Basically it's all a random inconsistent mess.

leonerd: Actually, since those messages are new in 5.35.x, I'll do a second commit after this one to just neaten them all up a bit.

wolfsage (approved):
Seems to fix the problem!

PR #19241, opened 2021-11-23
Quote the word "defer" in exception messages about control flow

leonerd:
These messages are all new in the 5.35.x development cycle anyway, so there's no risk of code out in the wild relying on the exact spelling from any stable version.

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -749,9 +749,9 @@ of C<SV*> scalars containing the arguments.
 integer format can only be used with positive integers, and you attempted
 to compress something else.  See L<perlfunc/pack>.
 
-=item Can't "%s" out of a defer block
+=item Can't "%s" out of a "defer" block
 
-(F) An attempt was made to jump out of the scope of a defer block by using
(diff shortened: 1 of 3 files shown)


pod/perldiag.pod
```diff
@@ -1042,7 +1042,7 @@ error occurs in cases such as these:
     goto G;
     $x + do { G: $y }; # How is + supposed to get its first operand?
 
-=item Can't "goto" into a defer block
+=item Can't "goto" into a "defer" block
 
 (F) A "goto" statement was executed to jump into the scope of a defer
```
ilmari: `C<defer>` here too (and maybe `C<goto>` instead of `"goto"`)?

I don't mean to turn this into an audit of all references to keywords in the document, but while we're nearby …

ilmari (changes requested):

pod/perldiag.pod
leonerd: Done in 54cf2adaef

ilmari (approved):

PR #19244, opened 2021-11-24
Replace sv_2mortal(newSVpv(x,0)) with newSVpvn_flags(x,... ,SVs_TEMP)

richardleach:
This should be more efficient.

(I'm not sure whether there's any further scope to simplifying any of the ternaries because it's safe to assume that the first argument pointer won't be NULL.)

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -3407,7 +3407,7 @@ Perl_mortal_getenv(const char * str)
     ret = getenv(str);
 
     if (ret != NULL) {
-        ret = SvPVX(sv_2mortal(newSVpv(ret, 0)));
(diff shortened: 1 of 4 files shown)


richardleach:
I need to triage Win test failures - one of these changes or because it's being built on D: and may not have short path name support?

Probably the change to  `XST_mPV(i,v`

richardleach:
Yeah, that looks to be it. Will let the tests run and then squash the commits if all is well.

tonycoz:
Is this still intended to have "needs-work"?

richardleach:
I'd left the label while the checks ran again. Removed it now.

PR #19246, opened 2021-11-26
ExtUtils-MakeMaker: add $(LDFLAGS) when linking binary modules

kanavin:


jkeenan:
ExtUtils-MakeMaker is maintained upstream on CPAN.  Its bug tracker is https://rt.cpan.org/Dist/Display.html?Name=ExtUtils-MakeMaker; it's repository is https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker.  So that's where you should submit a pull request.

When you do so, I would recommend providing a rationale for the suggested change, as the revision would apply not only to EU-MM as it appears in blead but also in EU-MM as it gets installed against older versions of Perl and many platforms.

kanavin:
> ExtUtils-MakeMaker is maintained upstream on CPAN. Its bug tracker is https://rt.cpan.org/Dist/Display.html?Name=ExtUtils-MakeMaker; it's repository is https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker. So that's where you should submit a pull request.
> 
> When you do so, I would recommend providing a rationale for the suggested change, as the revision would apply not only to EU-MM as it appears in blead but also in EU-MM as it gets installed against older versions of Perl and many platforms.

Thanks, submitted there https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/405

PR #19247, opened 2021-11-29
Implement builtin:: functions using ops

leonerd:
* Sets a callchecker to rewrite callsites to these new builtin functions into optrees
 * Sets prototypes on the functions - `""` for the constant ones, `"$"` for the unary function
 * Adds new opcode `OP_ISBOOL`
 * Existing XSUBs still need to remain because of cases like `$code = \&builtin::true`
 * `true` and `false` become OP_CONST and thus will const-fold as per normal

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -16,6 +16,13 @@
 
 #include "XSUB.h"
 
+struct BuiltinFuncDescriptor {
+    const char *name;
+    XSUBADDR_t xsub;
+    OP *(*checker)(pTHX_ OP *, GV *, SV *);
+    IV ckval;
+};
+
 XS(XS_builtin_true);
 XS(XS_builtin_true)
 {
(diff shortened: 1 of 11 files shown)


builtin.c
```diff
@@ -34,20 +41,125 @@ XS(XS_builtin_false)
...
+    SAVEFREESV(prototype);
+
+    assert(entersubop->op_type == OP_ENTERSUB);
+
+    entersubop = ck_entersub_args_proto(entersubop, namegv, prototype);
+
+    SV *constval;
+    switch(builtin->ckval) {
+        case 0: constval = &PL_sv_no; break;
+        case 1: constval = &PL_sv_yes; break;
```
ilmari: Do we want an `enum` for these, like we have for the functions?

leonerd: Possibly; if we create any more. I didn't imagine we'd need more than just these two constants at the moment though.

ilmari: Might as well do it straight away, it doesn't cost anything. And there's only one function, and that still/already has an `enum`.

leonerd: Fair; added 65ec621870

builtin.c
```diff
@@ -207,11 +212,18 @@ Perl_boot_core_builtin(pTHX)
...
+            proto = "";
+        else if(builtin->checker == &ck_builtin_func1)
+            proto = "$";
+
+        CV *cv = newXS_flags(builtin->name, builtin->xsub, __FILE__, proto, 0);
         XSANY.any_i32 = builtin->ckval;
 
         if(builtin->checker) {
             cv_set_call_checker_flags(cv, builtin->checker, newSVuv(PTR2UV(builtin)), 0);
+
```
ilmari: Spurious blank line.

leonerd: Thanks; fixed 8e443f3d86

builtin.c
```diff
@@ -91,9 +208,23 @@ XS(XS_builtin_import)
 void
 Perl_boot_core_builtin(pTHX)
 {
-    newXS_flags("builtin::true",   &XS_builtin_true,   __FILE__, NULL, 0);
-    newXS_flags("builtin::false",  &XS_builtin_false,  __FILE__, NULL, 0);
-    newXS_flags("builtin::isbool", &XS_builtin_isbool, __FILE__, NULL, 0);
+    I32 i;
+    for(i = 0; builtins[i].name; i++) {
+        const struct BuiltinFuncDescriptor *builtin = builtins + i;
```
ilmari: Given that `builtins` is declared as an array, and we're using `builtins[i].name` above, `builtins[i]` would be more consistent.

ilmari (changes requested):
How about some tests in `t/perf/optree.t` and `lib/B/Deparse.t`?

builtin.c
leonerd: Fixed.

leonerd:
@ilmari I've added a test in `lib/B/Deparse.t` now. I'll take a look at the optree one.

leonerd:
After a conversation with haarg on IRC I ended up changing the implementation, to just add an individual opcode for `isbool`, rather than the indirect dispatch approach. This is likely nicer all round.

leonerd:
Various tests added; reviewers PTAL.

ilmari (approved):

PR #19252, opened 2021-12-01
Test equivalence of 'true' and 'false' to negated statements

jkeenan:
The documentation for lib/builtin.pm asserts that 'true' is equivalent
to !!1 or !0, and that false is equivalent to !!0 or !1.  Demonstrate
that this is so.

### DIFF
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -112,4 +112,16 @@ package FetchStoreCounter {
     ok($recursecoderef->("rec"), 'true in self-recursive anon sub');
 }
 
+{
+    use builtin qw( true false );
+
+    my $val = true;
+    cmp_ok($val, $_, !!1, "true is equivalent to !!1 by $_") for qw( eq == );
+    cmp_ok($val, $_,  !0, "true is equivalent to  !0 by $_") for qw( eq == );
+
+    $val = false;
+    cmp_ok($val, $_, !!0, "false is equivalent to !!0 by $_") for qw( eq == );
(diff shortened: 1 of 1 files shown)


jkeenan:
@leonerd, if you have any objection to this p.r., can you please let us know?

Otherwise, I plan to merge it within 3 days.

Thank you very much.
Jim Keenan

leonerd:
`is()` perhaps isn't the best test for this. It will use either an `eq` or `==` internally. Perhaps would be best to `cmp_ok` them on *both* of these tests specifically.

jkeenan:
> `is()` perhaps isn't the best test for this. It will use either an `eq` or `==` internally. Perhaps would be best to `cmp_ok` them on _both_ of these tests specifically.

Updated.

lib/builtin.t
```diff
@@ -75,4 +75,16 @@ use warnings;
     ok($recursecoderef->("rec"), 'true in self-recursive anon sub');
 }
 
+{
+    use builtin qw( true false );
+
+    my $val = true;
+    cmp_ok($val, 'eq', !!1, 'true is equivalent to !!1');
```
leonerd: I'd suggest testing both modes;

```perl
cmp_ok($val, $_, !!1, "true is equivalent to !!1 by $_") for qw( eq == );
...
```

jkeenan: Updated.

leonerd:
Looking good. Will approve when the CIs are happy (accepting that cygwin is never happy lately)

leonerd (approved):
LGTM. (Though please squash to merge)

PR #19254, opened 2021-12-02
Set the correct op flags on new OP_ISBOOL

leonerd:


t/perf/opcount.t
```diff
@@ -696,10 +696,19 @@ test_opcount(0, "builtin::true/false are replaced with constants",
                 });
 
 test_opcount(0, "builtin::isbool is replaced with direct opcode",
-                sub { my $x = 123; my $y = builtin::isbool($x); },
+                sub { my $x = 123; my $y; $y = builtin::isbool($x); },
                 {
                     entersub => 0,
                     isbool   => 1,
+                    padsv    => 3, # OA_TARGLEX applies so only 3, not 4
```
ilmari, line 703: Should this also check `sassign => 0`?

leonerd: Well, 1 because `$x` still has one. Just not `$y`. But yes I'll add that. 2df1a36e41

ilmari: You could avoid the `$x` one by leaving it uninitialised, since `isbool(undef)` is perfectly valid.

leonerd: Ohyes, I just wanted something that wouldn't const-fold. That will do too. 5ea6acf8e9

ilmari (approved):

pp.c
```diff
@@ -7215,9 +7215,11 @@ PP(pp_cmpchain_dup)
 PP(pp_isbool)
 {
     dSP;
+    dTARGET;
     SV *arg = POPs;
 
-    PUSHs(boolSV(SvIsBOOL(arg)));
+    sv_setbool(TARG, SvIsBOOL(arg));
```
ilmari: Shouldn't this be `sv_setbool_mg`?

ilmari (changes requested):

pp.c
leonerd: It should. And we should SvGETMAGIC() beforehand as well. I'll add those + some tests for it.

leonerd: Magic added in c2d575565b

ilmari (approved):

PR #19256, opened 2021-12-02
Do not count tests, instead use done_testing()

mbeijen:
this is my first contribution to perl5

Having to count tests is annoying for maintainers.

Also, manually updating count tests is problematic when multiple people
are working on the same code; it causes merge conflicts and recounts.

Sometimes creative workarounds involving BEGIN blocks were used to make
counting the tests a little more bearable. But these tricks are no
longer required.

done_testing() is available since Test::More 0.88 which was released in
2009.

While updating this I've also removed shebang lines if possible - there
were many different shebangs in use, and the files do not have an
execute bit. I've also added 'use strict; use warnings;' if possible.

Some of the tests use 'test.pl' and this is most often not super easy to
change because they use features of 'test.pl' not available in
Test::More; such as fresh_perl_*(); luckily also `test.pl` has
done_testing()!

This PR covers all tests in lib/ that were using Test::More and a plan.

(updated: previously I also made minor other cleanups, which increased the complexity of the PR, probably a bit too much)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -910,6 +910,7 @@ Michael Stevens                <mstevens@etla.org>
 Michael van Elst               <mlelstv@serpens.de>
 Michael Witten                 <mfwitten@gmail.com>
 Michele Sardo
+Michiel Beijen                 <mb@x14.nl>
 Mik Firestone                  <fireston@lexmark.com>
 Mike Doherty                   <mike@mikedoherty.ca>
 Mike Fletcher                  <fletch@phydeaux.org>
--- a/lib/B/Deparse-core.t
+++ b/lib/B/Deparse-core.t
@@ -36,7 +36,6 @@ BEGIN {
 
 use strict;
 use Test::More;
-plan tests => 3904;
 
 use feature (sprintf(":%vd", $^V)); # to avoid relying on the feature
                                     # logic to add CORE::
@@ -434,6 +433,8 @@ SKIP:
     }
     ok($pass, "sanity checks");
 }
+
+done_testing();
 
 __DATA__
 #
--- a/lib/B/Deparse-subclass.t
+++ b/lib/B/Deparse-subclass.t
@@ -4,7 +4,7 @@
 # publicise an API for subclassing B::Deparse they can prevent us from
 # gratuitously breaking conventions that CPAN modules already use.
 
-use Test::More tests => 2;
+use Test::More;
 
 use B::Deparse;
 
@@ -22,3 +22,5 @@ like  $nm->coderef2text(sub { my($a, $b, $c) }),
 like  $nm->coderef2text(sub { my $c; /(??{ $c })/; }),
       qr/\Q(??{\E \$c_groovy/,
      'overriding padname works for renaming lexicals in regexp blocks';
+
+done_testing();
--- a/lib/Benchmark.t
+++ b/lib/Benchmark.t
@@ -8,7 +8,7 @@ BEGIN {
 use warnings;
 use strict;
 our ($foo, $bar, $baz, $ballast);
-use Test::More tests => 213;
+use Test::More;
 
 use Benchmark qw(:all);
 
@@ -634,6 +634,7 @@ is_deeply ([keys %Benchmark::Cache], \@before_keys,
     }
 }
 
+done_testing();
 
 package TieOut;
 
--- a/lib/Config/Extensions.t
+++ b/lib/Config/Extensions.t
@@ -6,7 +6,7 @@ BEGIN {
     }
 }
 use strict;
-use Test::More 'no_plan';
+use Test::More;
 
 BEGIN {use_ok 'Config::Extensions', '%Extensions'};
 
@@ -31,3 +31,5 @@ while (my ($key, $val) = each %Extensions) {
     my $type = $val . '_ext';
     like($Config{$type}, $re, "$key is $type");
 }
+
+done_testing();
--- a/lib/DB.t
+++ b/lib/DB.t
@@ -18,7 +18,7 @@ use Scalar::Util qw( dualvar );
 my $dualfalse = dualvar(0, 'false');
 my $dualtrue = dualvar(1, 'true');
 
-use Test::More tests => 106;
+use Test::More;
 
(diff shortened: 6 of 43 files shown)


jkeenan:
> this is my first contribution to perl5
> 

Welcome!  And thanks for taking the time to contribute to Perl 5.


> Having to count tests is annoying for maintainers.
> 
> Also, manually updating count tests is problematic when multiple people are working on the same code; it causes merge conflicts and recounts.
> 
> Sometimes creative workarounds involving BEGIN blocks were used to make counting the tests a little more bearable. But these tricks are no longer required.
> 

I'm with you so far ... but with an important caveat.  You have made changes to **70** files in the core distribution.  That is a lot for reviewers to plow through.  My eyes glaze over at that number of changes.

I feel that it would have been better to discuss this, either on #p5p or on our mailing list, before touching such a large number of files.  It might have been better to structure this as a **series** of smaller pull requests.

> done_testing() is available since Test::More 0.88 which was released in 2009.
> 
> While updating this I've also removed shebang lines if possible - there were many different shebangs in use, and the files do not have an execute bit. I've also added 'use strict; use warnings;' if possible.
> 

This is where I disagree with you.  (Note that I'm speaking for myself here and other P5P committers may disagree with me.)  I believe it's better to focus on one type of clean-up in a given pull request.  For example, clean-up of shebang lines would be done in one pull request or in a series of smaller pull requests.  Adding `use strict` and `use warnings` would similarly be better done in pull requests focused on each of those.

(I recommend this on the basis of past experience.  Adding `strict` and `warnings` to test files has been surprisingly controversial in the past -- and I speak as someone who, in the no-longer-extant Perl 7 project in 2020, tried to get the entire test suite running with `strict` and `warnings` on by default.  We managed to succeed with `strict` but not with `warnings`.)

> Some of the tests use 'test.pl' and this is most often not super easy to change because they use features of 'test.pl' not available in Test::More; such as fresh_perl_*(); luckily also `test.pl` has done_testing()!
> 
> This PR covers most of the files in ext/ and lib/
> 

In particular, revisions to files using `test.pl` should be done separately from those using `Test::More`.

> I've been listening on #p5p on irc.perl.org and noticed a conversation on this topic; I realized that most tests for perl internals in t/ can also be updated to use 'done_testing();'

I'm frequently available on #p5p U.S. Easter time daytimes (as `kid51`).  Please feel free -- no, more than free, encouraged -- to join the conversations there so that we can have real-time back-and-forth on this.

But, once again, thank you for your efforts and we look forward to working with you on this.

Thank you very much.
Jim Keenan

leonerd:
On @jkeenan 's point here:

> I'm with you so far ... but with an important caveat. You have made changes to 70 files in the core distribution. That is a lot for reviewers to plow through. My eyes glaze over at that number of changes.

> I feel that it would have been better to discuss this, either on #p5p or on our mailing list, before touching such a large number of files. It might have been better to structure this as a series of smaller pull requests.

This change was infact discussed on IRC - with me :) So I was anticipating seeing it.

Though I would agree with jkeenan's comment that it would be better to split this into a few separate changes. `done_testing()` in one commit, and separately look into the strict+warnings. We have some history of that and it turns out for a bunch of files it isn't as simple as we'd hope to just add those, because of decades of accumulated cruft and bad habits in some files.

leonerd:
*Update*: Oh, I see on re-reading that you have in fact done just that. Yes this is looking much better now. I'll press the CI box and see what that makes of things.

leonerd:
A couple of failures on the CI run there that you might want to look into

khwilliamson:
On 12/3/21 9:34 AM, Paul Evans wrote:
> On @jkeenan <https://github.com/jkeenan> 's point here:
> 
>     I'm with you so far ... but with an important caveat. You have made
>     changes to 70 files in the core distribution. That is a lot for
>     reviewers to plow through. My eyes glaze over at that number of changes.
> 
>     I feel that it would have been better to discuss this, either on
>     #p5p or on our mailing list, before touching such a large number of
>     files. It might have been better to structure this as a series of
>     smaller pull requests.
> 
> This change was infact discussed on IRC - with me :) So I was 
> anticipating seeing it.
> 
> Though I would agree with jkeenan's comment that it would be better to 
> split this into a few separate changes. |done_testing()| in one commit, 
> and separately look into the strict+warnings. We have some history of 
> that and it turns out for a bunch of files it isn't as simple as we'd 
> hope to just add those, because of decades of accumulated cruft and bad 
> habits in some files.
> 
> —
>

There are people who think that done_testing is not as bomb-proof as a 
count, and have hence opposed changing tests.

I have never understood their position.

But if you're such a person, now is the time to speak up, with your reasons.

mbeijen:
I updated the PR to reduce the scope; the CI fail should also be gone now.

I also understand that I touched a lot of files, but by keeping the change so trivially simple, I hope it is easily reviewable. Also, the benefit of changing the test files is that we have good test coverage, no need to worry too much about that 😄

leonerd:
Oddly, some of the CI workflows are failing the authors test; but others are not. The failure is:

```
 Michiel W. Beijen  <michiel.beijen@gmail.com> not found in AUTHORS
porting/authors.t .................................................... 
Failed 1/1 subtests 
```

I'm entirely unsure why that would only affect some of them.

jkeenan:
> Oddly, some of the CI workflows are failing the authors test; but others are not. The failure is:
> 
> ```
>  Michiel W. Beijen  <michiel.beijen@gmail.com> not found in AUTHORS
> porting/authors.t .................................................... 
> Failed 1/1 subtests 
> ```
> 
> I'm entirely unsure why that would only affect some of them.

I'll take this ticket for the purpose of (a) fixing up any porting tests; (b) pausing a few days per the concerns @khwilliamson raised above; and (c) merging it into blead.

jkeenan:
@mbeijen, you've got 2 stray lines in `lib/sort.` which are causing these warnings to be thrown:
```
Useless use of addition (+) in void context at ./lib/sort.t line 29.
Argument "*main::3" isn't numeric in addition (+) at ./lib/sort.t line 29.
```
Can you fix?

Thank you very much.
Jim Keenan

mbeijen:
Thanks for noticing! I fixed `lib/sort.t`.

PR #19258, opened 2021-12-03
builtin:: functions for working on weakrefs

leonerd:
Adds three new functions and associated direct opcodes.

Tidies up the previous `enum` that used to index builtin funcs, because they're now all directly associated with opcodes so we might as well just use that.

Also fixes a bug in `B::Deparse` that didn't handle the `OA_TARGMY` optimisation of `isbool` correctly.


lib/builtin.t
```diff
@@ -49,6 +49,22 @@ package FetchStoreCounter {
...
+    my $ref = $arr;
+
+    ok(!isweak($ref), 'ref is not weak initially');
+
+    weaken($ref);
+    ok(isweak($ref), 'ref is weak after weaken()');
+
+    unweaken($ref);
+    ok(!isweak($ref), 'ref is not weak after unweaken()');
+}
```
ilmari, line 70: How about a test that `undef`ing the strong ref also `undef`s the weak one, i.e. that it has the desired practical effect, not just satisfying the predicate function.

lib/builtin.pm
```diff
@@ -16,7 +16,10 @@ builtin - Perl pragma to import built-in utility functions
 
 =head1 SYNOPSIS
 
-    use builtin qw( true false isbool );
+    use builtin qw(
+        true false isbool
+        isweak weaken unweaken
```
ilmari: `weaken unweaken isweak` would be more consistent with the bool ones (also in the docs below), and a more logical order IMO.

leonerd: +1. Done in 2e6da4895d

lib/builtin.t
leonerd: Yeah, I guess. I didn't want to go too far into testing the concept of weakrefs here because I'm sure that'll exist elsewhere, but I guess we could test it enough to demonstrate we've in fact connected the builtin functions to the right core concept. Done in dfa4906e94

t/perf/opcount.t
```diff
@@ -712,4 +712,11 @@ test_opcount(0, "builtin::isbool gets constant-folded",
                     const    => 1,
                 });
 
+test_opcount(0, "builtin::weaken is replaced with direct opcode",
```
ilmari, line 715: Should we test for `unweaken` and `isweak` too, just for completeness?

leonerd: Yeah; why not? Added 37622de526

tonycoz (approved):

builtin.c
```diff
@@ -85,17 +81,46 @@ XS(XS_builtin_func1)
...
             Perl_pp_isbool(aTHX);
             break;
 
+        case OP_ISWEAK:
+            Perl_pp_isweak(aTHX);
+            break;
+
         default:
-            Perl_die(aTHX_ "panic: unhandled ix value %d for xs_builtin_func1()", ix);
+            Perl_die(aTHX_ "panic: unhandled opcode %d for xs_builtin_func1_scalar()", ix);
```
Leont, line 93: That opcode number can generally be turned into a name/description, right?

Leont, line 93: That opcode number can generally be turned into a name/description, right?

PR #19260, opened 2021-12-04
Add builtin::blessed, refaddr and reftype

leonerd:
Give `blessed()` the same `OPp_TRUEBOOL` optimisation that `ref()` has in boolean context, avoiding creating a temporary SV containing the class name only for it to be evaluated for truth. As with `ref()`, the fact that `"0"` is a valid package name yet is false adds an annoying complication to the problem.


lib/builtin.t
```diff
@@ -69,6 +69,30 @@ package FetchStoreCounter {
...
 
+# reference queries
+{
+    use builtin qw( refaddr reftype blessed );
+
+    my $arr = [];
+    my $obj = bless [], "Object";
+
+    is(refaddr($arr), $arr+0, 'refaddr yields same as ref in numeric context');
+    ok(!defined refaddr("not a ref"), 'refaddr yields undef for non-reference');
```
ilmari: `is(…, undef, …)` (as used below) is more useful, since it shows what the returned value actually was.

op.c
```diff
@@ -18263,6 +18263,7 @@ Perl_rpeep(pTHX_ OP *o)
         }
 
         case OP_REF:
+        case OP_BLESSED:
             /* see if ref() is used in boolean context */
```
ilmari: This comment is now out-of-date. Could it be reworded to not refer to the op/function name, but rather _why_ the optimisation is useful?

pp.c
```diff
@@ -7253,6 +7253,75 @@ PP(pp_unweaken)
...
+
+    if(!SvROK(arg) || !SvOBJECT((rv = SvRV(arg)))) {
+        SETs(&PL_sv_undef);
+        RETURN;
+    }
+
+    if((PL_op->op_private & OPpTRUEBOOL) ||
+            ((PL_op->op_private & OPpMAYBE_TRUEBOOL) && (block_gimme() == G_VOID))) {
+        /* refs are always true, but we just have to check for the annoying
+         * cornercase of the package named "0" */
```
ilmari: The truthiness of refs isn't really relevant (neither here nor in `pp_ref`). How about something like `all package names are true, except the annoying corner case of "0"`?

ilmari (changes requested):
Do we want to add `t/perf/opcount.t` tests for these too, for completeness?

leonerd:
Thanks @ilmari; comments updated.

ext/Opcode/Opcode.pm
```diff
@@ -417,6 +417,7 @@ These are a hotchpotch of opcodes still waiting to be considered
     once
 
     rv2gv refgen srefgen ref refassign lvref lvrefslice lvavref
+    blessed reftype refaddr
```
wolfsage: EXTREME NIT: This is `blessed refaddr reftype` in builtins.pm

leonerd: Fixed f146c8ee14

pp.c
```diff
@@ -7253,6 +7253,76 @@ PP(pp_unweaken)
...
+
+    SvGETMAGIC(arg);
+
+    if(!SvROK(arg) || !SvOBJECT((rv = SvRV(arg)))) {
+        SETs(&PL_sv_undef);
+        RETURN;
+    }
+
+    if((PL_op->op_private & OPpTRUEBOOL) ||
+            ((PL_op->op_private & OPpMAYBE_TRUEBOOL) && (block_gimme() == G_VOID))) {
```
wolfsage, line 7270: What is the special casing here for void context? Is blessed in void context even useful?

ilmari: `OPpMAYBE_TRUEBOOL` means that it's part of a logical operator in the final statement in a sub, so we can only optimise it away if the sub is called in void context, since we don't know what the sub return value is used for.

See [the comments for `S_check_for_bool_cxt`](https://github.com/Perl/perl5/blob/v5.35.6/op.c#L16851) for far more detail.

leonerd: Huh, the more I think about it the more I'm not quite sure. I copied that from the similar code in `pp_ref`. Perhaps that needs more staring at.

leonerd: Ah - I see @ilmari has already answered that :)

wolfsage: I'm still confused but I can live with that.

lib/builtin.t
leonerd: Done c5c365a825

leonerd:
Opcount tests added in 03dea871c9

PTAL

ilmari (approved):
LGTM

PR #19263, opened 2021-12-06
Modernise File::Compare a little

JRaspass:
Based on @leonerd 's recent email about "dogfooding" newer features in core I've attempted to modernise File::Compare to the dizzying heights of Perl 5.12. I wasn't sure if one should always bump the version of a lib when changing it but figured since the ISA change is technically user visible we ought to.

The file could definitely be improved more, specially the use of bareword filehandles. But I believe this to be a fairly uncontroversial improvement that removes more lines than it adds which is always good.

 - Move the version to the package line and bump it.
 - Drop superfluous use version.
 - Switch to non-inheritance based Exporter usage.
 - Use defined-or operator.

### DIFF
--- a/lib/File/Compare.pm
+++ b/lib/File/Compare.pm
@@ -1,18 +1,14 @@
-package File::Compare;
+package File::Compare 1.1007;
 
-use 5.006;
-use strict;
+use v5.12;
 use warnings;
-our($VERSION, @ISA, @EXPORT, @EXPORT_OK, $Too_Big);
 
-require Exporter;
+use Exporter 'import';
 
-$VERSION = '1.1006';
-@ISA = qw(Exporter);
-@EXPORT = qw(compare);
-@EXPORT_OK = qw(cmp compare_text);
+our @EXPORT = qw(compare);
+our @EXPORT_OK = qw(cmp compare_text);
 
-$Too_Big = 1024 * 1024 * 2;
+our $Too_Big = 1024 * 1024 * 2;
 
 sub croak {
     require Carp;
@@ -127,8 +123,7 @@ sub compare_text {
 	if @_ == 3 && ref($cmp) ne 'CODE';
 
     # Using a negative buffer size puts compare into text_mode too
-    $cmp = -1 unless defined $cmp;
-    compare($from, $to, $cmp);
+    compare($from, $to, $cmp // -1);
 }
 
 1;
@@ -177,6 +172,3 @@ are equal, 1 if the files are unequal, or -1 if an error was encountered.
 
 File::Compare was written by Nick Ing-Simmons.
 Its original documentation was written by Chip Salzenberg.
-
-=cut
-


jkeenan:
> Based on @leonerd 's recent email about "dogfooding" newer features in core I've attempted to modernise File::Compare to the dizzying heights of Perl 5.12. I wasn't sure if one should always bump the version of a lib when changing it but figured since the ISA change is technically user visible we ought to.
> 
> The file could definitely be improved more, specially the use of bareword filehandles. But I believe this to be a fairly uncontroversial improvement that removes more lines than it adds which is always good.
> 
>     * Move the version to the package line and bump it.
> 
>     * Drop superfluous use version.
> 
>     * Switch to non-inheritance based Exporter usage.
> 
>     * Use defined-or operator.

I have to admit that I haven't completely grokked what @leonerd was getting at with the post about dogfooding.  But my initial impression of these changes is favorable.

However, assuming that others concur with what you have done so far, I would urge you to submit a second commit (within this p.r.) in which you tidy the code up:
* uncuddle `else`s
* convert all leading tabs to whitespace x 4 and fix indents
* etc.

khwilliamson:
On 12/5/21 5:52 PM, James E Keenan wrote:
> I have to admit that I haven't completely grokked what @leonerd 
> <https://github.com/leonerd> was getting at with the post about 
> dogfooding. But my initial impression of these changes is favorable.
> 

We should use new features in our test files and internal scripts; those 
in /ext are candidates.

I would oppose changing things, for "dogfooding" reasons, in dist/ and 
/cpan in such a way that they require a more modern perl to run than 
they currently do.

jkeenan:
On 12/5/21 11:20 PM, Karl Williamson wrote:
> On 12/5/21 5:52 PM, James E Keenan wrote:
>  > I have to admit that I haven't completely grokked what @leonerd
>  > <https://github.com/leonerd> was getting at with the post about
>  > dogfooding. But my initial impression of these changes is favorable.
>  >
> 
> We should use new features in our test files and internal scripts; those
> in /ext are candidates.
> 
> I would oppose changing things, for "dogfooding" reasons, in dist/ and
> /cpan in such a way that they require a more modern perl to run than
> they currently do.
> 

Agreed.

leonerd:
These changes already look good. They can be merged now, or we can wait for further updates. I'm happy either way.

lib/File/Compare.pm
```diff
@@ -1,18 +1,14 @@
-package File::Compare;
+package File::Compare 1.1007;
 
-use 5.006;
```
Leont, line 3: I would favor having a use <version> statement here, it's a self-documenting sort of thing.

Leont:
> We should use new features in our test files and internal scripts; those
in /ext are candidates.

`lib/` is internal the same way `ext/` is. This would be ok if we don't want to dual-life this later on (do we?)

lib/File/Compare.pm
JRaspass: In the case it should be `v5.12` and above the package line so older perls don't die on that syntax first?

khwilliamson:
On 12/6/21 9:28 AM, Leon Timmermans wrote:
>     We should use new features in our test files and internal scripts; those
>     in /ext are candidates.
> 
> |lib/| is internal the same way |ext/| is. This would be ok if we don't 
> want to dual-life this later on (do we?)

I should have included lib/ with ext/.

If we were to dual-life, we could change things back, but my guess if we 
were to do that we wouldn't bother, since people have gotten along 
without it all this long.

lib/File/Compare.pm
jkeenan: > In the case it should be `v5.12` and above the package line so older perls don't die on that syntax first?

Older perls will never die on that syntax because any older perl will only see the version of File::Compare that ships with that perl.  If we apply this p.r. it will be seen by perl-5.36 but not by any earlier version.

For that reason I feel that we do *not* need any `use VERSION;` statement at the top of the file.  If we want to indicate that the file's syntax is expected to implicitly require, day, perl-5.12 syntax, we can document that in a comment or a commit message.

In order to move the process forward, I would like to apply @JRaspass's revisions in about 3 days.

JRaspass: @jkeenan I've restored and bumped the perl version to `v5.12` and left it where it was. This serves two purposes, documenting what level of Perl features this file currently uses, and allows us to drop the `use strict` as we get it from the use version. One day we might even be able to drop the `use warnings` by just bumping said version, that would be nice. Anything that reduces boilerplate for modern perl is very welcome in my book.

Grinnz: Agreed, if the purpose is modernization, appropriate use of `use VERSION`  is a good idea and serves several purposes.

mbeijen: @JRaspass, you wrote:
> One day we might even be able to drop the use warnings by just bumping said version, that would be nice. A

I just wanted to point out that the one day has come! Since 5.35 this is happening now:

```
$ perl -Mv5.34 -E 'say $a'
$ perl -Mv5.35 -E 'say $a'
Use of uninitialized value $a in say at -e line 1.
```

jkeenan:
Since the OP is continuing to push commits to this p.r., I'm de-scheduling the merge.

JRaspass:
> Since the OP is continuing to push commits to this p.r., I'm de-scheduling the merge.

FWIW I'm done tweaking.

PR #19265, opened 2021-12-07
Fix signatures in unicore/mktables

leonerd:
Tidies up some leftover uses of `@_`


lib/unicore/mktables
```diff
@@ -10824,7 +10824,7 @@ sub output_perl_charnames_line ($a, $b) {
     # ordinals, but they are all private use or noncharacters which do not
     # have names, so won't be in this table.
 
-    return sprintf "%05X\n%s\n\n", $_[0], $_[1];
+    return sprintf "%05X\n%s\n\n", $a, $b;
```
ilmari, line 10827: I know this is existing code, but while we're here, please rename these to something descriptive, like `$codepoint` and `$name`.

khwilliamson: I'll do that in a separate commit

PR #19266, opened 2021-12-08
Update compile and bind options for z/OS (os390)

MikeFultonDev:
This update enables us to build EBCDIC static/dynamic
and 31-bit/64-bit addressing mode Perl. The number of
tests that pass is consistent with the baseline before
these updates namely:

For blead.31.dynamic.ebcdic
  Configured using -Dusedl

  Failed 98 tests out of 1939, 94.95% okay.
  Elapsed: 1038 sec
  u=18.13  s=6.04  cu=535.69  cs=178.56  scripts=1939  tests=1035071

For blead.64.dynamic.ebcdic
  Configured using -Dusedl -Duse64bitall 

  Failed 102 tests out of 1941, 94.74% okay.
  Elapsed: 1057 sec
  u=19.49  s=6.49  cu=543.00  cs=181.00  scripts=1941  tests=1053149

These changes also provide the base support to be able to provide
ASCII static/dynamic and 31-bit/64-bit addressing mode Perl.

Description of changes:

Makefile.SH
  Changes were made to the os390*) case specific part of the code.
  Support was added for the 64-bit DLL path because the original
  only had support for 31-bit.

cpan/ExtUtils-MakeMaker/lib/ExtUtils/MM_OS390.pm
  Changes were made to pick up khw's fix that ensures that
  all files are specified after all options because the c99
  options processing requires this organization

hints/os390.sh
  This is the largest set of changes:
   - Compilation and Link options were added for ASCII and 64-bit
   - A z/OS specific check was added to determine if the Perl
     code being built is ASCII or EBCDIC. The check works by
     looking at the first character of the shell script to see
     if it is an ASCII or non-ASCII character. If ASCII, then
     the build is deemed to be ASCII. If not ASCII, it is assumed
     to be EBCDIC.
   - Cleanup was performed to remove code for z/OS systems that are
     no longer supported, simplifying the file (e.g.
        "`uname -v`x`uname -r`" in 02x0[89].*|02x1[0-9].*|[0-9][3-9]x*)
     which would only be true on unsupported, very old pre-z/OS systems
   - The compiler has been changed from xlc to c99. Both are available
     as priced features of the operating system, and the c99 compiler is
     a better 'fit' for options processing, being more consistent with
     c99 on other platforms.
   - Suppressing warning messages for CCN3159 were added because the
     1-bit bitfields flagged due to a smaller-than-int data type are
     harmless.
   - Several feature test macros were added to bring the compilation up
     to a modern level to enable Perl to take advantage of capabilities
     available on all supported z/OS 2.4 and up systems.
   - Removed the -Wl,EDIT=NO because debug information is no longer stored
     in an area that will be loaded into memory, but is now stored in a
     NOLOAD section. So - while this does mean that the binary on disk is a
     little 'fat', what is loaded into memory is not, and it means that
     people can have better problem determination tools available even on
     production Perl distributions.


Leont:
I think you want to squash these commits, they don't make for a clean history on their own.

khwilliamson:
squashed and merged via 0a91ad7bad09fcaf7a49940068a65ce44e753743

PR #19267, opened 2021-12-08
Use the new `builtin::` functions in core unit tests

leonerd:
Avoids loading `Scalar::Util`.

In several places this tidies up logic as now it no longer has to be conditional on not running on miniperl, since the `builtin::` functions are even available there.

### DIFF
--- a/t/op/array.t
+++ b/t/op/array.t
@@ -440,14 +440,13 @@ $::ra = [ bless [], 'A' ];
 pass 'no crash when freeing array that is being cleared';
 
 # [perl #85670] Copying magic to elements
-SKIP: {
-    skip "no Scalar::Util::weaken on miniperl", 1, if is_miniperl;
-    require Scalar::Util;
+{
     package glelp {
-	Scalar::Util::weaken ($a = \@ISA);
-	@ISA = qw(Foo);
-	Scalar::Util::weaken ($a = \$ISA[0]);
-	::is @ISA, 1, 'backref magic is not copied to elements';
+        use builtin 'weaken';
+        weaken ($a = \@ISA);
+        @ISA = qw(Foo);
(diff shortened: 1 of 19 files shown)


ilmari (approved):

t/op/array.t
```diff
@@ -440,14 +440,13 @@ $::ra = [ bless [], 'A' ];
 pass 'no crash when freeing array that is being cleared';
 
 # [perl #85670] Copying magic to elements
-SKIP: {
-    skip "no Scalar::Util::weaken on miniperl", 1, if is_miniperl;
-    require Scalar::Util;
+{
```
wolfsage: This outer bracing is no longer needed I think and should go

t/op/hash.t
```diff
@@ -100,13 +100,12 @@ sub guard::DESTROY {
 }
 
 # Weak references to pad hashes
-SKIP: {
-    skip_if_miniperl("No Scalar::Util::weaken under miniperl", 1);
+{
```
wolfsage, line 103: These outer braces are also no longer needed

wolfsage:
Seems good! The two nits I added are cases where the test file seems to mostly not use braces for separating test cases, so keeping them in adds a little cognitive overhead, but it's fine if you do.

wolfsage (approved):

t/op/array.t
leonerd: Possibly; though I didn't want to disturb more lines than necessary, but I might as well for consistency. Done e076c0cd9a

t/op/hash.t
leonerd: Perhaps not, but I'll keep them in because they still guard the `$ref` lexical later on,

PR #19269, opened 2021-12-09
Use `builtin` in core libs

leonerd:
Various files in `lib/` and `ext/` that aren't dual-life, might as well use the new `builtin` functions, especially if it avoids needing to load the `Scalar::Util` XS module.

This now makes it possible to `use overload` in miniperl.

### DIFF
--- a/lib/File/Copy.pm
+++ b/lib/File/Copy.pm
@@ -44,8 +44,8 @@ sub carp {
 sub _catname {
     my($from, $to) = @_;
     if (not defined &basename) {
-	require File::Basename;
-	import  File::Basename 'basename';
+        require File::Basename;
+        File::Basename->import( 'basename' );
     }
 
     return File::Spec->catfile($to, basename($from));
--- a/lib/File/Copy.pm
+++ b/lib/File/Copy.pm
@@ -7,15 +7,13 @@
 
 package File::Copy;
 
-use 5.006;
+use 5.035007;
 use strict;
 use warnings; no warnings 'newline';
+use builtin 'blessed';
+use overload;
 use File::Spec;
 use Config;
-# During perl build, we need File::Copy but Scalar::Util might not be built yet
-# And then we need these games to avoid loading overload, as that will
-# confuse miniperl during the bootstrap of perl.
-my $Scalar_Util_loaded = eval q{ require Scalar::Util; require overload; 1 };
 # We want HiRes stat and utime if available
 BEGIN { eval q{ use Time::HiRes qw( stat utime ) } };
 our(@ISA, @EXPORT, @EXPORT_OK, $VERSION, $Too_Big, $Syscopy_is_copy);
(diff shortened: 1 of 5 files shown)


lib/File/Copy.pm
```diff
@@ -24,7 +22,7 @@ sub syscopy;
 sub cp;
 sub mv;
 
-$VERSION = '2.37';
+$VERSION = '2.38';
 
 require Exporter;
 @ISA = qw(Exporter);
```
Grinnz, line 28: Indirect import method call here got broken by the import of the 5.035007 feature bundle 😄

leonerd: Fixed in fbf19e9285

hvds:
> Various files in `lib/` and `ext/` that aren't dual-life, might as well use the new `builtin` functions, especially if it avoids needing to load the `Scalar::Util` XS module.
> 
> This now makes it possible to `use overload` in miniperl.

I assume this means some tests no longer need to skip under miniperl, should those be changed by this PR too?

leonerd:
> I assume this means some tests no longer need to skip under miniperl, 

It does indeed. I found and fixed some but maybe there are more still lurking?

> should those be changed by this PR too?

I'm open to suggestions - can either change that in this PR, or do another one.

hvds:
> > I assume this means some tests no longer need to skip under miniperl,
> 
> It does indeed. I found and fixed some but maybe there are more still lurking?
> 
> > should those be changed by this PR too?
> 
> I'm open to suggestions - can either change that in this PR, or do another one.

To my mind the main value of "more things work under miniperl" is that more things can be tested under miniperl. If the change adds value beyond that, it makes sense to consider it in isolation; if not, it would make more sense if it captured that value as part of the change.

hvds:
> To my mind the main value of "more things work under miniperl" is that more things can be tested under miniperl. If the change adds value beyond that, it makes sense to consider it in isolation; if not, it would make more sense if it captured that value as part of the change.

And I should have added: having core using the builtin functions is definitely of value, so in that sense it would be reasonable to consider this PR in isolation and change the tests separately.

leonerd:
@hvds I've had a hunt around but only found one more test in the end, that can be unconditionalised. A lot of the non-miniperl tests require `PerlIO::scalar`, which is unrelated here. And all of the rest are for a good mix of other reasons.

hvds:
> @hvds I've had a hunt around but only found one more test in the end, that can be unconditionalised. A lot of the non-miniperl tests require `PerlIO::scalar`, which is unrelated here. And all of the rest are for a good mix of other reasons.

Thanks for looking.

khwilliamson:
I think utf8::unicode_to_native and native_to_unicode should be accessible from builtin:: as well.  They really don't belong in utf8:: as they have nothing to do with UTF-8.  But we shouldn't remove them from there

leonerd:
@khwilliamson Ohsure there's still loads of functions that need adding to `builtin` along the way. I'm just doing that in small steps.

ilmari (approved):
LGTM

PR #19270, opened 2021-12-10
sv_upgrade: minor AV/HV new_body refactoring

richardleach:
This PR includes 2 commits:
- Removal of a branch that appears to be a no-op. 
- Changing xpvav & xpvhv initialization from Zero() followed by struct member overwrites, to using compound literals.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -1371,13 +1371,6 @@ Perl_sv_upgrade(pTHX_ SV *const sv, svtype new_type)
             AvMAX(sv)	= -1;
             AvFILLp(sv)	= -1;
             AvREAL_only(sv);
-            if (old_type_details->body_size) {
-                AvALLOC(sv) = 0;
-            } else {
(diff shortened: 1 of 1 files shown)


richardleach:
The one concern I have is whether or not it's really portable & safe to initialize the `xmg_u` union like this?

tonycoz:
> The one concern I have is whether or not it's really portable & safe to initialize the `xmg_u` union like this?

It does seem a bit strange to mix designated and non-designated initialization, but it is well defined in C99 and C11.

Designating xmg_u might be clearer though:
```
-                .xmg_stash = NULL, {.xmg_magic = NULL}, .xav_fill = -1,
+                .xmg_stash = NULL, .xmg_u = {.xmg_magic = NULL}, .xav_fill = -1,
```
and works for me.

richardleach:
Thanks, I've done that, rebased and pushed.

PR #19274, opened 2021-12-11
 mktables: Use builtin::refaddr

khwilliamson:
Now that this function is available in miniperl, mktables can use it to
avoid a bunch of visually distracting 'no overloading' calls.


leonerd (approved):
Overall LGTM.
I slightly wonder if this makes problems for maintainers because now no existing released version of perl can run `mktables`, only 5.35.7. But perhaps that's OK?

khwilliamson:
On 12/12/21 6:08 AM, Paul Evans wrote:
> ***@***.**** approved this pull request.
> 
> Overall LGTM.
> I slightly wonder if this makes problems for maintainers because now no 
> existing released version of perl can run |mktables|, only 5.35.7. But 
> perhaps that's OK?
> 
Things in /lib and /ext are required to build perl, and have never, to 
my knowledge, been required to work with older perls.  We would put such 
things in /dist.

lib/unicore/mktables
```diff
@@ -7740,8 +7709,6 @@ sub trace { return main::trace(@_); }
 
                         return if $self->carp_if_locked;
 
-                        my $addr = do { no overloading; pack 'J', $self; };
```
jkeenan, line 7743: This revision is different from all the other revisions in this file, as here we are removing assignment to a variable which we never subsequently use.  Perhaps, for clarity, we could remove this in a separate, previous commit?

jkeenan: Ignore previous comment.  Once I looked at the p.r. *per commit* rather than as a whole, I saw that you found a few other instances of unused variables.

tonycoz (approved):

PR #19276, opened 2021-12-13
perltie: Fix variable name in TIEHASH example

jwilk:


jkeenan:
I don't think there is a strong case for accepting this pull request.  By my count, the syntax `my $self  = shift;` is used over 20 times in `pod/perltie.pod`.  Why should it be changed in this one instance?

Grinnz:
The other instances using $self are not constructors. The other constructor examples all use $class already, this PR makes this example consistent with them. LGTM.

khwilliamson:
Agreed with Grinnz

PR #19281, opened 2021-12-15
POSIX::FLT_ROUNDS should not be an NV constant

t-a-k:
Perl's `POSIX::FLT_ROUNDS` used to be an NV constant, but `FLT_ROUNDS` in C <float.h> is actually an integer and not a constant (C99 seems to say that it reflects the current rounding mode set by `fesetround()`).

This PR will fix this by turning `POSIX::FLT_ROUNDS` into an XS function (actually an alias of existing function for `POSIX::fesetround()`).

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -4438,6 +4438,7 @@ ext/POSIX/lib/POSIX.pod		POSIX extension documentation
 ext/POSIX/Makefile.PL		POSIX extension makefile writer
 ext/POSIX/POSIX.xs		POSIX extension external subroutines
 ext/POSIX/t/export.t		Test @EXPORT and @EXPORT_OK
+ext/POSIX/t/fenv.t		Floating-point rounding mode tests for POSIX
 ext/POSIX/t/iscrash		See if POSIX isxxx() crashes with threads on Win32
 ext/POSIX/t/iv_const.t		See if integer constants of POSIX are IV
 ext/POSIX/t/math.t		Basic math tests for POSIX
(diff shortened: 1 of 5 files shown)


sisyphus:
FWIW (probably very little) I support this PR.
Even on mingw-w64 ports of gcc, it is noted (in comments in float.h) that <code>FLT_ROUNDS</code> should take its value from fegetround().
Of course, they just ignore that and hard code it to "1". Visual Studio 2019 also hard codes it to "1", IIRC.

On Windows, the effect of this PR is to merely change the value of <code>POSIX::FLT_ROUNDS</code> from "1" to an error message informing us that <code>POSIX::FLT_ROUNDS</code> is unavailable.
But that happens only because <code>HAS_FEGETROUND</code> is not defined.
And it turns out that there is no good reason (AFAICS) that <code>HAS_FEGETROUND</code> should not be defined  on Windows - irrespective of whether the Windows perl is built with Visual Studio or gcc.

I've actually tested this (with both Visual Studio 2019 and mingw-w64 port of gcc-11.2.0) where I've defined HAS_FEGETROUND (by patching config.gc, config.vc, config_H.gc and config_H.vc appropriately) and  patching POSIX.xs to <code>#include</code> the fenv.h header file.
With those simple patches, along with the patches provided in this PR, the POSIX extension then provides sane (and significantly enhanced) behaviour on Windows wrt rounding modes, and the POSIX test suite still passes.

I hope this PR is accepted.

PR #19282, opened 2021-12-15
Version tags for `use builtin :ver`

leonerd:
Supports the same syntax/semantics as `use feature`, in that

* `:major.minor` accepted
* `:major.minor.patch` accepted but patch number is ignored - anything goes

Not (yet) integrated into the main `use VERSION` syntax; but we can consider that in future.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -18,7 +18,7 @@
 
 struct BuiltinFuncDescriptor {
     const char *name;
-    int since_ver;
+    int since_ver;  /* if nonzero, included in every version bundle since this */
     XSUBADDR_t xsub;
     OP *(*checker)(pTHX_ OP *, GV *, SV *);
     IV ckval;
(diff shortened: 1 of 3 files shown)


Grinnz:
Like in feature.pm perhaps it would be best not to tie it to "all functions available in version X.Y" as there may be functions we want to remove or leave out for various reasons? And instead specify explicitly what is included in each bundle.

Grinnz:
Corollary: each builtin should document what version of Perl it was first available (or what version of builtin.pm, if it is ever dual-lifed)

leonerd:
@Grinnz 

> it would be best not to tie it to "all functions available in version X.Y" as there may be functions we want to remove or leave out for various reasons

I've adjusted it so if `since_ver` is zero it won't include that function in a bundle. Likely we'll want to use this for some of the rarer functions later - `getcwd` possibly, also likely many of the other bits and pieces stolen from Internals, utf8, or other places.

> each builtin should document what version of Perl it was first available

Docs added.

haarg:
I previously mentioned this on IRC, but if these functions are going to be marked as experimental, then they shouldn't be included in the version tag until they are considered stable.

tonycoz:
obsoleted by #21473 which was merged

PR #19283, opened 2021-12-16
Revert "_MSC_VER is now always >= 1800 (Visual C++ 2013 / Visual C++ 12.0)"

khwilliamson:
Reverts Perl/perl5#19226

The author said it should not have been merged yet


xenu:
Merged #19284 instead.

PR #19285, opened 2021-12-16
Add `finally` blocks to `try/catch` syntax

leonerd:
Permits an optional `finally` block, allowing the often-requested `try/catch/finally` syntax.

Equivalent to using a `defer` block, but many people prefer the top-to-bottom order the code reads this way.


t/op/try.t
```diff
@@ -281,4 +281,73 @@ no warnings 'experimental::try';
...
+}
+
+{
+    sub ff
+    {
+        try {
+            return "return inside try+finally";
+        }
+        catch ($e) {}
+        finally { "last value" }
```
ilmari: Should this test that the `finally` block actually gets invoked on `return` from the `try`?

leonerd: Good thinking. Added in b19ce2fd3d

pod/perlsyn.pod
```diff
@@ -711,6 +712,30 @@ C<eval> blocks, because those affect the way that C<return> would work. Since
 C<try> blocks do not intercept C<return>, they are not of interest to
 C<caller>.
 
+The C<try> and C<catch> blocks may optionally be followed by a third block
+introduced by the C<finally> keyword. This third block is executed after the
+rest of the construction has finished.
```
ilmari:
```suggestion
rest of the construct has finished.
```

leonerd: Done dd19a65632

ilmari (approved):

op.c
```diff
@@ -10723,6 +10724,32 @@ Perl_newDEFEROP(pTHX_ I32 flags, OP *block)
...
 
+/*
+=for apidoc op_wrap_finally
+
+Wraps the given C<block> optree fragment in its own scoped block, arranging
+for the C<finally> optree fragment to be invoked when leaving that block for
+any reason. Both optree fragments are consumed and the combined result is
+returned.
+
+ */
```
wolfsage: This extra whitespace before the closing `*/` causes a warning when compiling:
```
=cut missing? op.c:10735: */
```

pp_ctl.c
```diff
@@ -2494,7 +2499,9 @@ PP(pp_return)
         /* Check for  defer { return; } */
         for(i = cxstack_ix; i > cxix; i--) {
             if(CxTYPE(&cxstack[i]) == CXt_DEFER)
-                Perl_croak(aTHX_ "Can't \"%s\" out of a \"defer\" block", "return");
+                /* diag_listed_as: Can't "%s" out of a "defer" block */
```
wolfsage, line 2502: These diag_listed_as are only a slight lie now, since `"defer"` could be replace with `"finally"`. There's at least one precedent of listing the various forms that can be produced, one after another. It would be nice to add them here and the other places we have diag_listed_as

wolfsage (changes requested):
One compilation error worth fixing. Up to you if you want to update diag_listed_as.

Thanks for this!

op.c
leonerd: Fixed in c33a5ad6f8

pp_ctl.c
leonerd: Ahah; I didn't realise you could list multiple of them at once, but I see we do that for signature arity checking. I'll add the "finally" versions here. 
Done 85500884dc

leonerd:
@wolfsage Fixed - PTAL

pp_ctl.c
wolfsage: Seems good though you didn't also add this to the two cases below? Are they not possible?

PR #19288, opened 2021-12-19
Tie::SubstrHash: rewrite tests to use Test::More

mbeijen:
These tests were still handrolling TAP output. I've dragged the tests
into this century by porting them to Test::More.

### DIFF
--- a/lib/Tie/SubstrHash.t
+++ b/lib/Tie/SubstrHash.t
@@ -1,15 +1,7 @@
-#!/usr/bin/perl -w
-# 
-
-BEGIN {
-    chdir 't' if -d 't';
-    @INC = '.'; 
-    push @INC, '../lib';
-}    
-
-print "1..20\n";
-
 use strict;
+use warnings;
+
+use Test::More;
 
 require Tie::SubstrHash;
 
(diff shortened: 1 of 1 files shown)


jkeenan:
> These tests were still handrolling TAP output. I've dragged the tests into this century by porting them to Test::More.

In tests 12 through 20, the tests which you are replacing explicitly use `==`.  Might it be better to use `cmp_ok ... ==` in the replacements rather than `is`?

mbeijen:
`cmp_ok` has the benefit that it allows you to specify a comparator, for example `cmp_ok(3, '<', 5, "Three is less than five")`.

But with equal comparison the result in case of problems is exactly the same, as for using `is()`. So I think using `is()` is more clear than using `cmp_ok($i, '==', n)`, and there is no difference in what is tested

```
michiel@moon:~/p/perl5$ perl -MTest::More -e 'is(1,2)'
not ok 1
#   Failed test at -e line 1.
#          got: '1'
#     expected: '2'

$ perl -MTest::More -e 'cmp_ok(1, "==", 2)'
not ok 1
#   Failed test at -e line 1.
#          got: 1
#     expected: 2
```

Please note that the `==` operator is also not comparing types, so `cmp_ok(1, "==", "1")` would still be passing.

Grinnz:
String comparison is not numeric comparison though: `is '01', '1e0'` is false, though these are numerically equal.

mbeijen:
That is true. But I'd like to point out that these tests are about counting the number of keys in a hash, or checking the length of something. And surely then the test is about the numerical value, more than if 'scalar keys' returns an int 1 versus the string '01'

Grinnz:
Yes, which makes numerical comparison the more correct option as it will not distinguish between those. But there are plenty of tests which rely on stringification of numbers already.

khwilliamson:
So @grinnz, @jkeenan are you ok that since we have plenty of tests that rely on number stringification that is acceptable to use 'is' here, instead of cmp_ok with '==' ? which is what I think that @mbeijen is indicating

Grinnz:
It's acceptable but it would be preferred to use a numeric comparison as in the tests being replaced.

jkeenan:
> So @Grinnz, @jkeenan are you ok that since we have plenty of tests that rely on number stringification that is acceptable to use 'is' here, instead of cmp_ok with '==' ? which is what I think that @mbeijen is indicating

I'm okay with `is` and will merge the p.r. after final testing.

jkeenan:
Merged to blead in commit 18364759f2; closing ticket.

PR #19297, opened 2021-12-25
Use pull request HEAD commit instead of pr/merge commit

happy-barney:
By default action uses merge commit (created with github account
data). This may cause authors test to fail when contributor's info
differs from github account (eg contributor prefers cpan email when
contributing to perl)

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -37,6 +37,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1000
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: git setup
         run: |
           echo "Pull request authors"
@@ -91,6 +92,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: git cfg + fetch tags
         run: |
           git config diff.renameLimit 999999
@@ -156,6 +158,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libgdbm-dev libdb-dev
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: git cfg + fetch tags
         run: |
           git config diff.renameLimit 999999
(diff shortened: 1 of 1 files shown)


Grinnz:
These actions also run on standard commits, will this work for that?

haarg:
If this is only to fix the authors test, perhaps it would be better to change the test?  An older branch could pass tests, but fail once it was merged to a newer blead.

jkeenan:
> By default action uses merge commit (created with github account data). This may cause authors test to fail when contributor's info differs from github account (eg contributor prefers cpan email when contributing to perl)



> By default action uses merge commit (created with github account data). This may cause authors test to fail when contributor's info differs from github account (eg contributor prefers cpan email when contributing to perl)

Color me "Skeptical" about this commit.  I am a committer who makes commits mainly by pushing from the command-line, just as I did for 7 years before we moved our repository to github.com and pull requests became available.  Some of those commits are my own contributions, but many others are by first-time contributors.  If someone is making their first contribution to the Perl core distribution, I am not going to hold porting test failures against them.  I will gladly do what is needed to get the metadata to pass `make test_porting`.  I will not insist that a first-timer continuously revise a pull request until they get all the porting tests to pass.

As a consequence, I am skeptical about the value of GitHub actions and workflows.  I don't want *anything* that makes my life as a committer more complicated, and I fear that a lot of mussing and fussing with the GitHub stuff will do just that.

I'm not unhappy with the authors test, but if other committers are, then I would recommend taking up @haarg's suggestion above and change the test directly.

mbeijen:
When I recently created my first pull request for perl5, the authors test failed for me. Then I added myself as an author, in a new commit, and the test _still_ failed. I was a little perplexed; the email I added in AUTHORS was exactly the same as the email address I used in my commits. 
The failure was exactly because of the issue described in this PR; my github email was different than the email I chose to use for the commits I pushed. I think this PR was valid and fixed the problem properly; and I think that if failing tests can be avoided, even if they are _merely_ porting tests, that's a win in my opinion.

happy-barney:
@mbeijen it solved only pull request part of the problem. As @Grinnz properly commented, it has to work for commits as well (and I don't have a space to play with that right now)

PR #19302, opened 2021-12-28
File-Find: Replace bareword filehandle with a lexical scalar

xsawyerx:
Bareword filehandles are bad and this module should really remove this one.

### DIFF
--- a/ext/File-Find/lib/File/Find.pm
+++ b/ext/File-Find/lib/File/Find.pm
@@ -3,7 +3,7 @@ use 5.006;
 use strict;
 use warnings;
 use warnings::register;
-our $VERSION = '1.39_01';
+our $VERSION = '1.40';
 use Exporter 'import';
 require Cwd;
 
@@ -322,7 +322,7 @@ sub _find_dir($$$) {
 	$dir_pref= ( $p_dir eq '/' ? '/' : "$p_dir/" );
     }
 
-    local ($dir, $name, $prune, *DIR);
+    local ($dir, $name, $prune);
 
     unless ( $no_chdir || ($p_dir eq $File::Find::current_dir)) {
 	my $udir = $p_dir;
@@ -381,12 +381,13 @@ sub _find_dir($$$) {
 	$dir= $dir_name; # $File::Find::dir
 
 	# Get the list of files in the current directory.
-	unless (opendir DIR, ($no_chdir ? $dir_name : $File::Find::current_dir)) {
+    my $dh;
+	unless (opendir $dh, ($no_chdir ? $dir_name : $File::Find::current_dir)) {
 	    warnings::warnif "Can't opendir($dir_name): $!\n";
 	    next;
 	}
-	@filenames = readdir DIR;
-	closedir(DIR);
+	@filenames = readdir $dh;
+	closedir($dh);
 	@filenames = $pre_process->(@filenames) if $pre_process;
 	push @Stack,[$CdLvl,$dir_name,"",-2]   if $post_process;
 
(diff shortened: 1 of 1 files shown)


jkeenan (approved):

Leont (approved):

shlomif:
@xsawyerx : I suggest moving `my $dh;` closer to the opendir calls: [predeclarations](https://perl-begin.org/tutorials/bad-elements/#declaring_all_vars_at_top)

xsawyerx:
[Edited: Removed a crappy part of my response because I was annoyed. Sorry.]

> @xsawyerx : I suggest moving `my $dh;` closer to the opendir calls: [predeclarations](https://perl-begin.org/tutorials/bad-elements/#declaring_all_vars_at_top)

That's what I did originally. I decided to then stick with the pattern in which this code was written, which included all variable declarations at the top of the function.

xsawyerx:
I've force-pushed a commit that includes the `$dh` variable defined below, prior to being used.

Considering the comment and 2 thumbs up (last I checked), I am complying by making this change. I still think the violation of the coding style of the existing code (where all lexicals are defined together at the beginning of the functions) is the wrong thing to do.

Grinnz:
Not that I have a strong opinion but for the sake of explaining my reaction: I think improving the code style is more important than being consistent with the bad code style in this particular case.

khwilliamson:
On 12/28/21 11:28, Dan Book wrote:
> Not that I have a strong opinion but for the sake of explaining my 
> reaction: I think improving the code style is more important than being 
> consistent with the bad code style in this particular case.
> 
I am mixed about this.  Making a change shouldn't open a can of worms 
about modernizing tangential things.  And by moving it would confuse 
newbies with an inconsistency.

Another option is to leave it where it was, but add a comment that this 
style is no longer recommended

xsawyerx:
> Not that I have a strong opinion but for the sake of explaining my reaction: I think improving the code style is more important than being consistent with the bad code style in this particular case.

Thank you for explaining. I understand and I do think there's logic to this, certainly.

I believe another commit that cleans up some of the code could be useful, but as it stands, this single change creates a situation in which almost all lexicals are defined in one location except *one*. Since that is far less consistent with either direction (one being lexicals defined where used and the other being all defined at once), I think it creates greater cognitive effort this way.

This is why I think a single change for this one particular case is not an improvement on the style, but instead splinters a consistent style (even if I think it's worse) for a single variable.

Of course, I yield to the thumbs up since this is not a hill worth any lives. :) I also appreciate the quick responses and didn't want to then drop the ball on my side.

xsawyerx:
> On 12/28/21 11:28, Dan Book wrote: Not that I have a strong opinion but for the sake of explaining my reaction: I think improving the code style is more important than being consistent with the bad code style in this particular case.
> I am mixed about this. Making a change shouldn't open a can of worms about modernizing tangential things. And by moving it would confuse newbies with an inconsistency.

Agreed on both accounts.

> Another option is to leave it where it was, but add a comment that this style is no longer recommended

Do you mean a comment in the code itself saying "please don't use the style of this code when writing other code"? I don't follow.

khwilliamson:
On 12/28/21 12:08, Sawyer X wrote:
>     On 12/28/21 11:28, Dan Book wrote: Not that I have a strong opinion
>     but for the sake of explaining my reaction: I think improving the
>     code style is more important than being consistent with the bad code
>     style in this particular case.
>     I am mixed about this. Making a change shouldn't open a can of worms
>     about modernizing tangential things. And by moving it would confuse
>     newbies with an inconsistency.
> 
> Agreed on both accounts.
> 
>     Another option is to leave it where it was, but add a comment that
>     this style is no longer recommended
> 
> Do you mean a comment in the code itself saying "please don't use the 
> style of this code when writing other code"? I don't follow.

I indeed meant "please don't use the style of this code when writing 
other code"

And I thre this out as an option; not a recommendation

jkeenan:
Pushed to blead in commit b838a0e335.  Further discussion of best ways to present code samples in documentation can be conducted on list or in a separate pull request.

Thank you very much.
Jim Keenan

PR #19304, opened 2021-12-29
Pod::Html: improve test suite

mbeijen:
- do not count tests
 - use strict and warnings everywhere
 - remove emacs identifiers
 - bump version number

### DIFF
--- a/ext/Pod-Html/t/anchorify-536.t
+++ b/ext/Pod-Html/t/anchorify-536.t
@@ -7,9 +7,6 @@ use Test::More;
 unless ($version == 35 or $version == 36) {
     plan skip_all => "Needed only during 5.36";
 }
-else {
-    plan tests => 3;
-}
 
 my @filedata;
 {
@@ -61,6 +58,8 @@ is_deeply(
     my $relativized_url = relativize_url( $url, "$file.html" );
     is($relativized_url, $expected_relativized_url, "relativize_url() returned expected value");
 }
+
+done_testing;
 
 __DATA__
 =head1 NAME
--- a/ext/Pod-Html/t/anchorify-536.t
+++ b/ext/Pod-Html/t/anchorify-536.t
@@ -1,5 +1,3 @@
-# -*- perl -*-
-
 use strict;
 use warnings;
 use Pod::Html qw( anchorify relativize_url );
--- a/ext/Pod-Html/t/anchorify-536.t
+++ b/ext/Pod-Html/t/anchorify-536.t
@@ -1,6 +1,7 @@
 # -*- perl -*-
 
 use strict;
+use warnings;
 use Pod::Html qw( anchorify relativize_url );
 my ($revision,$version,$subversion) = split /\./, sprintf("%vd",$^V);
 use Test::More;
--- a/ext/Pod-Html/t/anchorify.t
+++ b/ext/Pod-Html/t/anchorify.t
@@ -2,7 +2,7 @@
 
 use strict;
 use Pod::Html::Util qw( anchorify relativize_url );
-use Test::More tests => 3;
+use Test::More;
 
 my @filedata;
 {
@@ -54,6 +54,8 @@ is_deeply(
     my $relativized_url = relativize_url( $url, "$file.html" );
     is($relativized_url, $expected_relativized_url, "relativize_url() returned expected value");
 }
+
+done_testing;
 
 __DATA__
 =head1 NAME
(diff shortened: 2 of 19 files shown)


mbeijen:
This is more changes than I was trying to make :sob: 

What I wanted is to remove the begin block & the 'code before strictures' critic (although I know full well that it's not needed for code to adhere to this). And when I moved the BEGIN block to below the list of 'use' statements, test began to fail because this modules own 't/lib/Testing.pm' could no longer be loaded. So I fixed it by replacing the BEGIN block with a more idiomatic 'use lib'. If you want, I can split up those changes into a separate commit or even a separate PR, just let me know.

jkeenan:
> * do not count tests
> 
>     * use strict and warnings everywhere
> 
>     * no code before strictures
> 
>     * bump version number and remove string eval



> This is more changes than I was trying to make sob
> 
> What I wanted is to remove the begin block & the 'code before strictures' critic (although I know full well that it's not needed for code to adhere to this). And when I moved the BEGIN block to below the list of 'use' statements, test began to fail because this modules own 't/lib/Testing.pm' could no longer be loaded. So I fixed it by replacing the BEGIN block with a more idiomatic 'use lib'. If you want, I can split up those changes into a separate commit or even a separate PR, just let me know.

Yes, I recommend that if you're trying to make *several* kinds of code or documentation improvements across each of *several* different files, that you make only *one* kind of code change in a given commit.  Then, you can either have a given pull request contain only one commit, or you can request a code reviewer to review the code on a *per-commit* basis before trying to grok the whole pull request.

Thank you very much.
Jim Keenan

jkeenan:
> 
> What I wanted is to remove the begin block & the 'code before strictures' critic (although I know full well that it's not needed for code to adhere to this). And when I moved the BEGIN block to below the list of 'use' statements, test began to fail because this modules own 't/lib/Testing.pm' could no longer be loaded. So I fixed it by replacing the BEGIN block with a more idiomatic 'use lib'. If you want, I can split up those changes into a separate commit or even a separate PR, just let me know.

The way much of the code in `ext/Pod-Html/t/*.t` currently appears is the result of refactoring which I proposed in a series of pull requests earlier in this calendar year.  One of the principles of that refactoring was, "First, do no harm."  So while some (actually, *much*) of the code you find there is not the way I would have written the code *from scratch*, it was the least harmful way of making the changes that were really needed.

You may want to look at the difference in `ext/Pod-Html` between, say, perl-5.32.0 and blead in order to get a feel for why the code is the way it is.

mbeijen:
As requested, I've split up the commit in separate ones, each addressing a single subject. I hope this makes it easier for you to review!

Also, I've dropped the 'use lib' instead of the BEGIN block.

You mention your "First, do no harm" principle and of course, I subscribe to that and don't want to cause harm. With these commits I want to make the code base a little bit nicer; and I want to do that in a way that is as unobtrusive as possible.

jkeenan:
> As requested, I've split up the commit in separate ones, each addressing a single subject. I hope this makes it easier for you to review!
> 
> Also, I've dropped the 'use lib' instead of the BEGIN block.
> 
> You mention your "First, do no harm" principle and of course, I subscribe to that and don't want to cause harm. With these commits I want to make the code base a little bit nicer; and I want to do that in a way that is as unobtrusive as possible.

Thanks, it did indeed make my review easier.  Two points ...

1. Minor nit:  the by-commit diffs flag lack of a newline at end of file in these two tests:
```
ext/Pod-Html/t/cache.t
ext/Pod-Html/t/eol.t
```
2. Since (a) this code sits under `ext/`, which means it doesn't get independently released to CPAN; and (b) there are now no substantive changes to the `.pm` files, I think we can eliminate commit decc5ccd12b3edb18ec420374e65bfec51a6d508 entirely.  The only remaining change in any `*.pm` file is from `synch` to `sync` in a comment in `lib/Pod/Html/Util.pm`; I don't think that merits a version bump.

Anyone else want to comment on this p.r.?  Otherwise, after the two points above are resolved we'll be ready to squash, write a consolidated commit message and merge.

Thank you very much.
Jim Keenan

mbeijen:
As requested, I've removed the version bump commit (which also fixed the minor typo in the comment) and all the rest are only changes on files in `t/`

jkeenan:
Rebased on blead, squashed, updated commit message, merged to blead in ab296d81b5.  Closing p.r.

Thank you very much.

PR #19308, opened 2021-12-31
z/OS ASCII Enablement: Manage untagged system header file errno.h by copy and tag

MikeFultonDev:
The system headers on z/OS are in EBCDIC, but are untagged. The bi-modal build requires that files be tagged correctly
   so this patch copies the errno.h file (which may be mounted on a read-only file system) to a local file and then tags
   it as IBM-1047 (EBCDIC) so that it can be processed properly. The longer-term fix for this may be to have z/OS tag
   header files as EBCDIC, although this would not be something Perl could rely on for several years.


MikeFultonDev:
@khwilliamson can you suggest who would be good to review this patch request?

khwilliamson:
The character set looks like it is wrong; mojibake

MikeFultonDev:
darn. sorry - i must've transferred the wrong way. I'll delete this PR and open a new one

PR #19309, opened 2021-12-31
z/OS ASCII Enablement: Manage untagged header errno.h by copy & tag

MikeFultonDev:
The system headers on z/OS are in EBCDIC, but are untagged. The bi-modal build requires that files be tagged correctly
   so this patch copies the errno.h file (which may be mounted on a read-only file system) to a local file and then tags
   it as IBM-1047 (EBCDIC) so that it can be processed properly. The longer-term fix for this may be to have z/OS tag
   header files as EBCDIC, although this would not be something Perl could rely on for several years.

### DIFF
--- a/ext/Errno/Errno_pm.PL
+++ b/ext/Errno/Errno_pm.PL
@@ -2,7 +2,7 @@ use ExtUtils::MakeMaker;
 use Config;
 use strict;
 
-our $VERSION = "1.35";
+our $VERSION = "1.36";
 
 my %err = ();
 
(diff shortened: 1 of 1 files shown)


MikeFultonDev:
@khwilliamson sorry for the churn - who do you recommend review?

khwilliamson:
@leont The only problem I see is not a full path to cp, so could be hacked; but unchanged code doesn't have a full path to echo.

MikeFultonDev:
@khwilliamson are these checks run on z/OS or somewhere else? If somewhere else, I don't understand why code protected by a check for os390 would fail?

MikeFultonDev:
@Leont anything you would like me to change for this patch?

PR #19311, opened 2021-12-31
Specify PERL_USE_SAFE_PUTENV when configuring ASCII z/OS builds

MikeFultonDev:
The Perl code expects that the ``environ`` global variable can be re-allocated to new storage and then have
   entries added and removed from it. This isn't supported by z/OS when in Bi-Modal mode and so    
    _PERL_USE_SAFE_PUTENV_ macro is defined to use the _env_ services to manipulate ``environ`` 
   instead of doing so directly.  It is not clear
   if it is valid to re-allocate storage for the ``environ`` global variable or not, although
   it is worth pursuing with the z/OS development team as a longer-term potential fix.


MikeFultonDev:
@khwilliamson I expect you are the best person to review this change?

MikeFultonDev:
@khwilliamson Are the failing tests 'expected'? My changes are in a z/OS specific file so not sure why they would cause Linux to fail?

PR #19314, opened 2021-12-31
Need to remove ASCII tags on copied compiled executables for z/OS ASCII

MikeFultonDev:
Untag compiled executables (perl and the .so files) after they are copied with
 File::copy because File::copy is marking them as CCSID 819 (ASCII). The longer
 term fix for this could be to provide a syscopy on z/OS which would perform a
 _cp_ which on z/OS copies the tag information as part of the copy.


MikeFultonDev:
@khwilliamson please recommend the right person to review these changes

PR #19315, opened 2021-12-31
z/OS ASCII needs file descriptors on pipes tagged as ASCII

MikeFultonDev:
Add code to change the file descriptor on the file descriptors opened from
 ``Perl_my_popen_list`` and ``Perl_my_popen`` so that the CCSID of the file
 descriptor is 819. This change is only when the code is built for
 ASCII ``#if (__CHARSET_LIB == 1)`` on z/OS ``#if defined(OEMVS)``.

### DIFF
--- a/util.c
+++ b/util.c
@@ -2646,6 +2646,22 @@ Perl_unlnk(pTHX_ const char *f)	/* unlink all versions of a file */
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


MikeFultonDev:
@khwilliamson please recommend the right person to review these pipe code changes

PR #19317, opened 2021-12-31
Enable z/OS to run with '-e' option for runtests

MikeFultonDev:
Change this code so that the unset of MAKEFLAGS is protected because, on z/OS,
 unset will return non-zero if the variable is not set and this then causes the
 shell to fail because it is being run with _-e_. This change is only required
 for z/OS but is harmless to have on all platforms.


MikeFultonDev:
@khwilliamson please recommend someone that can review this change

PR #19318, opened 2021-12-31
Fix problem in makefile for z/OS 64-bit builds

MikeFultonDev:
This change fixes a bug where the value of _use64bitall_ was
  expected to be nothing or something, but in fact, the proper
  test is for ``define|true|[yY]*``.


MikeFultonDev:
@khwilliamson please review

khwilliamson:
Pulled via 979158253287183a813b94e89a988df99e6db713

PR #19322, opened 2022-01-04
dist/Time-HiRes/t/stat.t: provide descriptions for all tests

jkeenan:
Problems with `stat.t` have been reported in https://github.com/Perl/perl5/issues/19321.  This file, however, lacks descriptions for all its tests, so it can be difficult to figure out where you are in the file when debugging.

To facilitate debugging, let's provide rudimentary labels for all tests.

### DIFF
--- a/dist/Time-HiRes/t/Watchdog.pm
+++ b/dist/Time-HiRes/t/Watchdog.pm
@@ -10,30 +10,30 @@ my $watchdog_pid;
 my $TheEnd;
 
 if ($Config{d_fork}) {
-    print("# I am the main process $$, starting the watchdog process...\n");
+    note ("I am the main process $$, starting the watchdog process...");
     $watchdog_pid = fork();
     if (defined $watchdog_pid) {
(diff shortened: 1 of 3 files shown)


dist/Time-HiRes/t/Watchdog.pm
```diff
@@ -10,30 +10,30 @@ my $watchdog_pid;
 my $TheEnd;
 
 if ($Config{d_fork}) {
-    print("# I am the main process $$, starting the watchdog process...\n");
```
tonycoz, line 13: Time::HiRes depends on Test::More without a version number, so note() might not be available (unless you update the dependency in Makefile.PL to say "0.84")

tonycoz (changes requested):

dist/Time-HiRes/t/Watchdog.pm
jkeenan: That seems reasonable, as that's been available since Oct 2008.  p.r. updated.

atoomic (approved):

PR #19333, opened 2022-01-07
Fix “junk” from padding

FGasper:
This isn’t ready yet—I need to add a test—but at least fixes the issue on AlmaLinux 8, where I’ve seen it.


FGasper:
I think this isn’t going to work, alas … it requires special capabilities that a normal process doesn’t have.

PR #19336, opened 2022-01-08
Make spelling of values for 'FILES' consistent

jkeenan:
While developing a program which require-s Porting/Maintainers.pl for
the data in %Modules, I noticed that the entries for two modules'
'FILES' key-value pairs were spelled with a trailing slash.  All other
'FILES' KVPs omitted the trailing slash.

For consistency, let's omit the trailing slash.  Once that's done, two
invocations of "./perl -Ilib Porting/makemeta -y" were needed to make
'make test_porting' happy.


toddr (approved):
LGTM

Leont (approved):

PR #19339, opened 2022-01-10
Add posix_spawn support

### LINKED ISSUE #15355: vfork should be used for spawning external processes
Migrated from [rt.perl.org#128227](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=128227) (status was 'open')

Searchable as RT128227$

Leont:
This fixes #15355

This is not yet including the optimization for single-arg system() that doesn't contain metacharacter. I should probably add that before undrafting this.

It also doesn't include Configure probes for detecting if the platform supports `posix_spawn`, partly because I haven't looked into how to write that, but also because that may be a contentious change and we're quite close to that deadline: it may be a better change for `5.37.1`

### DIFF
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -62,6 +62,13 @@
 # include <sys/syscall.h>
 #endif
 
+/* XXX This should be removed obviously */
+#define HAS_POSIX_SPAWN
+
+#ifdef HAS_POSIX_SPAWN
+#include <spawn.h>
+#endif
+
 /* XXX Configure test needed.
    h_errno might not be a simple 'int', especially for multi-threaded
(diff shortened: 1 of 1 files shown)


jkeenan:
> This fixes #15355
> 
> This is not yet including the optimization for single-arg system() that doesn't contain metacharacter. I should probably add that before undrafting this.
> 
> It also doesn't include Configure probes for detecting if the platform supports `posix_spawn`, partly because I haven't looked into how to write that, but also because that may be a contentious change and we're quite close to that deadline: it may be a better change for `5.37.1`

@Leont, can we get an update on the status of this p.r.?  It has acquired merge conflicts.  Thanks.

PR #19343, opened 2022-01-13
POSIX.pod: fix in the description of "isnan"

t-a-k:
"Note" on the description of "isnan" in POSIX.pod used to say
> Note that you cannot test for "`NaN`-ness" with
> ```
>                  $x == $x
> ```
but actually we _can_ test for `NaN`-ness with equality operators.
("Equality Operators" section in `perlop` already mentions this.)

I also added a small example for this.
I hope this addition will improve clarity.

### DIFF
--- a/ext/POSIX/lib/POSIX.pod
+++ b/ext/POSIX/lib/POSIX.pod
@@ -830,7 +830,7 @@ L<equality operators|perlop/"Equality Operators"> (C<==> or C<!=>), as in
 
     print "x is not a NaN\n" if $x == $x;
 
-since the C<NaN> is not equivalent to anything, B<including itself>.
+since the C<NaN> is not equal to anything, B<including itself>.
 
 See also L</nan>, L</NaN>, L</isinf>, and L</fpclassify>.
 
--- a/ext/POSIX/lib/POSIX.pod
+++ b/ext/POSIX/lib/POSIX.pod
@@ -825,9 +825,10 @@ to use instead.  See L<perlrecharclass/POSIX Character Classes>.
 Returns true if the argument is C<NaN> (not-a-number) [C99].  Added in
 Perl v5.22.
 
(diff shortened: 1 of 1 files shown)


ext/POSIX/lib/POSIX.pod
```diff
@@ -825,11 +825,12 @@ to use instead.  See L<perlrecharclass/POSIX Character Classes>.
 Returns true if the argument is C<NaN> (not-a-number) [C99].  Added in
 Perl v5.22.
 
-Note that you cannot test for "C<NaN>-ness" with
+Note that you can also test for "C<NaN>-ness" with
+L<equality operators|perlop/"Equality Operators"> (C<==> or C<!=>), as in
 
-   $x == $x
+    print "x is not a NaN\n" if $x == $x;
```
atoomic, line 831: consider promoting `say` instead of `print`?

t-a-k: I'd like to use `say` too, but I think it is a bit premature to use it here because it is currently not enabled by default (adding such as `use v5.10;` might be confusing in this one-line code example), and the rest of POSIX.pod (and most of core documentations) still use `print`.

jkeenan:
> "Note" on the description of "isnan" in POSIX.pod used to say
> 
> > Note that you cannot test for "`NaN`-ness" with
> > ```
> >                  $x == $x
> > ```
> 
> but actually we _can_ test for `NaN`-ness with equality operators. ("Equality Operators" section in `perlop` already mentions this.)
> 
> I also added a small example for this. I hope this addition will improve clarity.

+1

Is there any objection to merging this documentation change?

Grinnz:
No objection. I would guess the intent of this was to say that you can't test for NaN by testing if something is == 'NaN', but something got lost in translation. The new text is more correct.

PR #19346, opened 2022-01-16
Discourage `@_` in signatured subroutine

leonerd:
No actual runtime behavioural change, but if we're intending to cause some change at some future version it would be useful to get this warning in anyway to help flush out code that would otherwise be affected by such a future change.


atoomic:
Maybe consider a squash & rebase, before merging this?

leonerd:
This is now rebased, but I don't think we should merge it yet until it's had some more review.

op.c
```diff
@@ -3775,6 +3775,22 @@ S_optimize_op(pTHX_ OP* o)
...
             break;
 
+        case OP_RV2AV:
+        {
+            OP *first = (o->op_flags & OPf_KIDS) ? cUNOPo->op_first : NULL;
+            if(CvSIGNATURE(PL_compcv) &&
+                    first && first->op_type == OP_GV &&
+                    cGVOPx_gv(first) == PL_defgv) {
+                OP *parent = op_parent(o);
+                while(parent && parent->op_type == OP_NULL)
```
atoomic: maybe consider using the existing macro for this 
`while( OP_TYPE_IS(parent, OP_NULL) )`

op.c
```diff
@@ -3775,6 +3775,22 @@ S_optimize_op(pTHX_ OP* o)
             }
             break;
 
+        case OP_RV2AV:
+        {
+            OP *first = (o->op_flags & OPf_KIDS) ? cUNOPo->op_first : NULL;
+            if(CvSIGNATURE(PL_compcv) &&
+                    first && first->op_type == OP_GV &&
```
atoomic: same here `first && first->op_type == OP_GV` can be replaced by `OP_TYPE_IS(first, OP_GV)`

t/op/signatures.t
```diff
@@ -1608,6 +1610,49 @@ while(<$kh>) {
...
+    discouraged_ok 'push',             'push @_, 1';
+    discouraged_ok 'unshift',          'unshift @_, 9';
+    discouraged_ok 'splice',           'splice @_, 1, 2, 3';
+    discouraged_ok 'keys on array',    'keys @_';
+    discouraged_ok 'values on array',  'values @_';
+    discouraged_ok 'each on array',    'each @_';
+    discouraged_ok 'print',            'print "a", @_, "z"';
+    discouraged_ok 'subroutine entry', 'func("a", @_, "z")';
+}
+
```
atoomic, line 1680: consider adding some tests for the reverse `not_discouraged_ok` to avoid regressions in the future
```perl
sub test($one,$two) {
    my $f = sub {
        my $a = $_[0];
    };
}
```

t/op/signatures.t
```diff
@@ -1608,6 +1610,49 @@ while(<$kh>) {
...
+    discouraged_ok 'array element',    '$_[0]';
+    discouraged_ok 'array element',    'my $one = 1; $_[$one]';
+    discouraged_ok 'push',             'push @_, 1';
+    discouraged_ok 'unshift',          'unshift @_, 9';
+    discouraged_ok 'splice',           'splice @_, 1, 2, 3';
+    discouraged_ok 'keys on array',    'keys @_';
+    discouraged_ok 'values on array',  'values @_';
+    discouraged_ok 'each on array',    'each @_';
+    discouraged_ok 'print',            'print "a", @_, "z"';
+    discouraged_ok 'subroutine entry', 'func("a", @_, "z")';
```
atoomic, line 1670: an extra test for `eval`?
`discouraged_ok 'eval',    'eval qq[ $_[0] ]';`
idea for `for`
`discouraged_ok 'for on array',    '1 for @_';`
idea for `__ANON__`
`discouraged_ok 'ANON subroutine' =>  'sub ($x) { goto &__ANON__ }'`
or another subroutine entry
`'sub ($x) { goto __ANON__->(@_) }'`

t/op/signatures.t
```diff
@@ -1608,6 +1610,49 @@ while(<$kh>) {
...
+}
+
+# Warnings can be disabled
+{
+    no warnings 'discouraged';
+    my $warnings = "";
+    local $SIG{__WARN__} = sub { $warnings .= join "", @_ };
+    eval q{ sub($x) { @_ = (1,2,3) } };
+    is($warnings, "", 'No warnings emitted within scope of  no warnings "discouraged"');
```
atoomic, line 1689: maybe consider moving the `no warnings` in the eval too?
```
    sub ($x) { 
        no warnings 'discouraged';
        @_ = ( 1..3 ) 
    }
```

op.c
leonerd: Ahyes, a handy macro. Will do.

op.c
leonerd: Again.

t/op/signatures.t
leonerd: Good idea. Done

t/op/signatures.t
leonerd: The `eval` one is potentially interesting. Not sure about the `for on array` though, because that's really just the same as most of the others. The implementation happens at a lower level than that so if any warnings work they all would.

Not sure what you're aiming at with those `__ANON__` ones.

t/op/signatures.t
leonerd: Done

t/op/signatures.t
atoomic: yeap the `__ANON__` has probably low value and is a dupe of any other subroutine call

t/op/signatures.t
atoomic: Thanks for considering it

t/op/signatures.t
atoomic: thanks for adding the `eval` one

atoomic (approved):
LGTM I cannot wait for starting using it and fix codebase :-)

leonerd:
Having discussed this with PSC, we've decided to make it an `experimental::` warning instead. Since I can't seem to see a way to change what branch name a PR is pointed at, I'll close and recreate the PR

PR #19349, opened 2022-01-18
Stop setting console to "con" on non-MSWin systems

choroba:
The condition has existed since eternity but it never made sense.
Only use "con" if $^O eq 'MSWin32'.

Also, remove MSDOS from the comment: it's the last forgotten mention
of the OS in the whole file.

### DIFF
--- a/lib/perl5db.pl
+++ b/lib/perl5db.pl
@@ -532,7 +532,7 @@ BEGIN {
 use vars qw($VERSION $header);
 
 # bump to X.XX in blead, only use X.XX_XX in maint
-$VERSION = '1.68';
+$VERSION = '1.69';
 
 $header = "perl5db.pl version $VERSION";
 
@@ -1538,11 +1538,11 @@ We then determine what the console should be on various systems:
         undef $console;
     }
 
-=item * Windows or MSDOS - use C<con>.
+=item * Windows - use C<con>.
 
 =cut
 
-    elsif ( -e "con" or $^O eq 'MSWin32' ) {
+    elsif ( $^O eq 'MSWin32' and -e "con" ) {
         $console = "con";
     }
 


choroba:
See [PerlMonks](https://www.perlmonks.org/?node_id=11140562) for the reason.

lib/perl5db.pl
```diff
@@ -1538,11 +1538,11 @@ sub _autoflush {
         undef $console;
     }
 
-=item * Windows or MSDOS - use C<con>.
+=item * Windows - use C<con>.
 
 =cut
 
-    elsif ( -e "con" or $^O eq 'MSWin32' ) {
+    elsif ( $^O eq 'MSWin32' and -e "con" ) {
```
ilmari, line 1545: Does `con` ever _not_ exist on Windows?

choroba: I have no idea and no way to find out 🤷🏽

xenu: Yes, it can be removed using DefineDosDevice() function.

xenu (approved):

lib/perl5db.pl
hadjiprocopis: I don't use windows at all but I think this check:
`elsif ( $^O eq 'MSWin32' and -e "con" ) {`
will always fail unless `-e "con"` does some magic and searches for the DOS special device called CON (short for console). In DOS, CON is a reserved word and can not exist as a file in a straight-forward way. I think the solution will be to simply remove `and -e "con"`. The Unix check further below: `-e "/dev/tty"` makes more sense as Unix is a sensible OS.

Edit: As I understand it, `con` is like STDOUT in `print STDOUT "ahah\n"` (ok not entirely, it is not a filehandle). Or in `bash` `echo ahaha >&2` or `echo ahaha > /dev/stderr` - but I am not an expert on dos+friends. I am just trying to reason why "con" can not be a file (especially in *any, arbitrary* current dir). (last edit 17 mins after posting).

bliako

pryrt: -e shows CON exists under windows:
```
C:\Users\peter.jones\Downloads\TempData>perl -le "print -e 'CON'"
1
C:\Users\peter.jones\Downloads\TempData>dir CON

 Directory of \\.

File Not Found
```

Leont (approved):

lib/perl5db.pl
ikegami: > I think this check: `elsif ( $^O eq 'MSWin32' and -e "con" ) {` will always fail unless `-e "con"` does some magic and searches for the DOS special device called CON (short for console).

`-e` doesn't do any magic, but the system does. `CON`, `NUL`, `LPT1`, etc are deemed to exist in every directory.

hadjiprocopis: got it, thanks

PR #19351, opened 2022-01-19
fix commit calculations in authors porting test

haarg:
For pull requests, the current commit is a merge commit simulated by
github. The first parent will be from blead - anything that has been
commited since the creation of the branch. The second parent will be the
last commit on the branch, which is what we want to test.

The commit range should be all of the commits in the second parent,
excluseding the first parent. We can use HEAD^1 and HEAD^2 for these.
The merge-base does not need to be precomputed, since the .. in the
commit range does that for us.


haarg:
This basically matches the behavior before 4f38b40108cb02e5470a9e2e7065a1f49978aa7f, which changed the commit calculations. It's not obvious what that commit was intending to fix, but it seems pretty plainly wrong to me. This also fixes the problem #19297 tried to address.

PR #19352, opened 2022-01-19
t/porting/authors.t: Windows fixes

xenu:
^ is a shell metacharacter in cmd.exe, so it needs to be quoted.

Also, /dev/null doesn't exist on Windows.

### DIFF
--- a/t/porting/authors.t
+++ b/t/porting/authors.t
@@ -9,6 +9,7 @@ BEGIN {
 
 use TestInit qw(T);    # T is chdir to the top level
 use strict;
+use File::Spec;
 
 find_git_or_skip('all');
 skip_all(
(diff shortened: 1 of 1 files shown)


t/porting/authors.t
```diff
@@ -29,12 +29,13 @@ elsif( $ENV{GITHUB_ACTIONS} && length $ENV{GITHUB_BASE_REF} ) {
     # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables
 
     # we should be on a merge commit, but double check
-    my $branch_head = `git rev-parse "HEAD^2" 2>/dev/null`;
+    my $null = $^O eq 'MSWin32' ? 'nul' : '/dev/null/';
```
ilmari: Spurious trailing slash on `/dev/null/`. But it's better to use `File::Spec->devnull` for max portability.

xenu: Fixed, thanks.

PR #19354, opened 2022-01-20
Adjust Win32/t/Unicode.t for Cygwin

atoomic:
Attempt to fix the Cygwin issue

We need to postpone the function call 

Error reported upstream:
https://github.com/perl-libwin32/win32/pull/33/files

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -1303,6 +1303,9 @@ use File::Glob qw(:case);
     'Win32' => {
         'DISTRIBUTION' => "JDB/Win32-0.58.tar.gz",
         'FILES'        => q[cpan/Win32],
+        'CUSTOMIZED'   => [
+            't/Unicode.t', # https://github.com/perl-libwin32/win32/pull/33/files
+        ]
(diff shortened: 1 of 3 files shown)


xenu:
This test does **not** fail in blead. In fact, it was recently fixed by this commit: 3fe11e08ed12031c30123faa114fa48661ea6226

atoomic:
@xenu view all CI run from pending PRs they are failing in blead due to the lack of wrapper around the function to skip
```
skip($^O eq "cygwin", Win32::GetLongPathName($subdir), $long);
```
should be
```
skip($^O eq "cygwin", sub { Win32::GetLongPathName($subdir) }, $long);
```

The current CI run confirm that this is fixing the issue

atoomic:
Mmm you are right the failure is now fixed, but the test skip should not even try to call the function IMO

atoomic:
Closing this case.

We do not need this fix with 0.58, the smoker are not failing anymore.
Considering using an anonymous sub to avoid running the code to test, should be considered but this is an upstream issue.

PR #19359, opened 2022-01-21
Builtin is experimental; print warnings

leonerd:


atoomic:
notice a few conflicts, please rebase accordingly

t/lib/warnings/builtin
```diff
@@ -0,0 +1,109 @@
+  builtin.c     experimental warnings from builtin functions
+
+__END__
+# op.c - booleans
```
leonerd: Small point but shouldn't this comment say `builtin.c`?

ilmari: Yes, that's a copy-pasteo because I can never remember how many `#`es are required to separate the tests. Amended and force-pushed.

leonerd:
@ilmari's additions LGTM. I can't actually hit the "Approve" button as it's my own PR. :)

ilmari (approved):

PR #19362, opened 2022-01-23
toke.c: Fix potential C sign error

khwilliamson:
Taking (UV)(* char) isn't what was meant.  Instead it should be
       (UV) * (U8 *)(char).
(and the UV cast becomes irrelevant in this case).

If the C compiler defaults char to unsigned, it doesn't matter.  Nor
does it matter if the char value is positive.  But it gives the wrong
result when 'char' means 'signed char' and the value is negative.

Experimentally adding non-ASCII brackets to this code showed up this
error.

### DIFF
--- a/toke.c
+++ b/toke.c
@@ -11473,9 +11473,9 @@ Perl_scan_str(pTHX_ char *start, int keep_bracketed_quoted, int keep_delims, int
                         *to++ = *s++;
                 }
                 /* allow nested opens and closes */
-                else if ((UV)*s == PL_multi_close && --brackets <= 0)
+                else if (*(U8 *) s == PL_multi_close && --brackets <= 0)
                     break;
-                else if ((UV)*s == PL_multi_open)
+                else if (*(U8 *) s == PL_multi_open)
                     brackets++;
                 else if (!d_is_utf8 && !UTF8_IS_INVARIANT((U8)*s) && UTF)
                     d_is_utf8 = TRUE;


toke.c
```diff
@@ -11473,9 +11473,9 @@ Perl_scan_str(pTHX_ char *start, int keep_bracketed_quoted, int keep_delims, int
                         *to++ = *s++;
                 }
                 /* allow nested opens and closes */
-                else if ((UV)*s == PL_multi_close && --brackets <= 0)
+                else if (*(U8 *) s == PL_multi_close && --brackets <= 0)
```
atoomic, line 11476: should not we also fixe the casting just above too?
```
                       && ( ((UV)s[1] == PL_multi_open)
                         || ((UV)s[1] == PL_multi_close) ))
```

khwilliamson:
On 1/24/22 18:55, ℕicolas ℝ. wrote:
> 
> 
> In toke.c <https://github.com/Perl/perl5/pull/19362#discussion_r791294825>:
> 
>> @@ -11473,9 +11473,9 @@ Perl_scan_str(pTHX_ char *start, int keep_bracketed_quoted, int keep_delims, int
>                           *to++ = *s++;
>                   }
>                   /* allow nested opens and closes */
> -                else if ((UV)*s == PL_multi_close && --brackets <= 0)
> +                else if (*(U8 *) s == PL_multi_close && --brackets <= 0)
> 
> should not we also fixed the casting just above too?

Somehow that fix didn't make it into the patch that got pushed.  Thanks 
for finding this

khwilliamson:
I misunderstood your comment.  The commit does make all the changes I believe are needed.  

((UV)s[1]

should work as intended.  The is no extra place for a cast, unlike the changes in this commit

PR #19364, opened 2022-01-24
Tidy snail from mktables

leonerd:
Removes the last warnings ahead of https://github.com/Perl/perl5/pull/19346


khwilliamson:
LGTM. 

 Is there a reason not to squash the commits.   As it is, porting tests will fail between the commits during a bisect.  We may not care, but have it unless there is a reason to split them that I'm unaware of

leonerd:
No reason I just forgot to squash them before pushing. Squashed now ;)

PR #19365, opened 2022-01-24
Suggestions from GW Haywood

davorg:
See https://www.nntp.perl.org/group/perl.perl5.porters/2022/01/msg262533.html


Grinnz:
This is going to require incrementing `$VERSION`. Please also change it to be declared as a string so the trailing zero doesn't get lost.

davorg:
@Grinnz  Thanks for the advice. Done :-)

Grinnz:
Failure is just `Dave Cross  <dave@perlhacks.com> not found in AUTHORS` - you may correct this or the merger can. LGTM

Grinnz:
You are listed in AUTHORS already so Porting/checkAUTHORS.pl needs to be updated with the alternate email used for the commits.

Grinnz (approved):

jkeenan:
Squashed, rebased on blead, merged in cc4fcefcb9.  Closing ticket.  Thank you very much.

PR #19369, opened 2022-01-26
Detect short name clashes in 'use charnames'

DrHyde:
When the user asks for short names for multiple scripts there is a potential for name clashes and confusion. For example, both Arabic and Hebrew have a letter called "alef" - in shouty Unicode terms they are HEBREW LETTER ALEF and ARABIC LETTER ALEF. So what happens when the user asks for short names for both scripts?

```perl
use charnames qw(arabic hebrew);

say "\N{alef}";    # Argh, confusing
```

In this particular example you always get the Hebrew alef, because that has a lower codepoint - 0x5d0 vs Arabic alef's 0x627.

This patch makes charnames.pm emit a warning (at use-time) when this happens.

Questions:

1. Is this an appropriate thing to do? It may be a bit noisier than you expect, as there are some surprising clashes - did you know that there's a LATIN (SMALL|CAPITAL) LETTER GAMMA? I didn't either.
2. To do this we need to load the character database in `import`. Is this acceptable? I assume so, as you're obviously calling `import` so you can use the feature, so you're going to get the memory consumption at some point.
3. The algorithm for finding duplicates in this first draft is naive and inefficient, but it's also _clear_, and in normal circumstances won't be called that often;
4. The text of the warning definitely needs work. It currently warns about all the scripts asked for, even those that _don't_ have clashes, and while it lists one clashing name, it might be useful to list all clashes if there are more of 'em.
5. Even though the issue is about character _naming_ and not character _encoding_, should it only warn if the 'utf8' warnings category is enabled like much of the rest of charnames.pm does?

### DIFF
--- a/lib/_charnames.pm
+++ b/lib/_charnames.pm
@@ -653,6 +653,8 @@ sub import
 {
   shift; ## ignore class name
 
+  populate_txt() unless $txt;
+
   if (not @_) {
     carp("'use charnames' needs explicit imports list");
   }
@@ -711,8 +713,6 @@ sub import
   ## see if at least we can find one letter from each script.
   ##
   if (warnings::enabled('utf8') && @scripts) {
-    populate_txt() unless $txt;
-
     for my $script (@scripts) {
       if (not $txt =~ m/^$script (?:CAPITAL |SMALL )?LETTER /m) {
         warnings::warn('utf8',  "No such script: '$script'");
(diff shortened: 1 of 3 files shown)


khwilliamson:
I haven't had a chance to review the code, but two things that might affect it:

1) The docs say that the order the scripts are listed in the use charnames statement is the order they are searched for the character name.  That it is not happening is a bug, and I think that would be preferable to fix instead.

2) One can also currently specify the script if there is ambiguity:

blead -C -lE 'use charnames qw(arabic hebrew :short);say "\N{ARABIC:ALEF}"'                      
ا

khwilliamson:
s/listed/searched/

DrHyde:
> I haven't had a chance to review the code, but two things that might affect it:
> 
> 1. The docs say that the order the scripts are searched in the use charnames statement is the order they are searched for the character name.  That it is not happening is a bug, and I think that would be preferable to fix instead.

Fixed in 423e766 (with tests in ce7801e). The fix is a bit icky, as the regex engine will always find the first possible match (ie will always prefer Hebrew over Arabic when trying to match `ARABIC|HEBREW`) so there are now more regex lookups as I iterate over the scripts one at a time.

I think that there's still some value in warning the user about the potential name clash but it's less important with that fix in place. If you think just fixing it and not warning is sufficient then I'll back the warning bit out.

Leont:
This looks like it needs a rebase to cut out the noise and prevent a non-ff merge

DrHyde:
> This looks like it needs a rebase to cut out the noise and prevent a non-ff merge

done

Leont:
Can you please reformat the commit messages so that the titles are at most 72 characters. Otherwise they won't be displayed well in most git viewers (including github).

DrHyde:
> Can you please reformat the commit messages ...

Done

PR #19371, opened 2022-01-27
subroutine redefinitions made with local() shouldn't warn

xenu:
The following code will no longer warn:

```
  use warnings 'redefine';
  sub foo {}
  local *foo = sub{};
```

The main purpose of local() is to temporarily redefine stuff, so it
doesn't make sense to warn about it.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -218,7 +218,9 @@ XXX Changes (i.e. rewording) of diagnostic messages go here
 
 =item *
 
-XXX Describe change here
+L<Subroutine %s redefined|perldiag/"Subroutine %s redefined">
+
+Localized subroutine redefinitions no longer trigger this warning.
 
 =back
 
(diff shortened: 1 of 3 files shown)


atoomic:
I really like that idea, this is true that having to disable warning in a scope while using local seems redundant

atoomic:
maybe we should even consider throwing a warning when trying to localize one sub which is not defined…

leonerd:
@atoomic I quite often provide a `local *Some::Subclass::name = sub { ... }` method declaration intended to override one in a base class, within a unit test. That doesn't replace an existing sub at that particular class. I don't think that should warn - it's still perfectly valid.

But overall I like this general concept.

leonerd (approved):
LGTM

atoomic:
Thanks again for that great work!

In order to keep focus on the intent of the Pull Request, I moved the discussion I've started concerning `warning on undefined CVs` to a dedicated thread: 
https://github.com/Perl/perl5/issues/19375

thanks for keeping the discussion on track with the existing PR.

atoomic (approved):

oodler577:
thank you, more like this please 😄

PR #19377, opened 2022-01-28
@_ (aka "snail") in signatured subs is experimental

leonerd:
Yes, currently all of signatures are experimental, but `@_` in them doubly so.

Once we've added this warning, we can remove the "experimental" tag from signatures as a whole, leaving this facet of them experimental while de-experimenting the rest of it.


t/op/signatures.t
```diff
@@ -1608,6 +1613,82 @@ while(<$kh>) {
...
+    sub warnings_from {
+        my ($code, $run) = @_;
+        my $warnings = "";
+        local $SIG{__WARN__} = sub { $warnings .= join "", @_ };
+        my $cv = eval qq{ sub(\$x) { $code }} or die "Cannot eval() - $@";
+        $run and $cv->(123);
+        return $warnings;
+    }
+
+    sub snailwarns_ok {
```
atoomic, line 1628: I like these names

lib/warnings.pm
```diff
@@ -116,6 +116,7 @@ our %Offsets = (
     'experimental::builtin'		=> 148,
     'experimental::defer'		=> 150,
     'experimental::for_list'		=> 152,
+    'experimental::snail_in_signatures'	=> 154,
```
atoomic: what a cool name

pod/perlsub.pod
```diff
@@ -473,13 +478,28 @@ that the caller passed no arguments:
...
+In the current perl implementation, when using a signature the arguments
+are still also available in the special array variable C<@_>.  However,
+accessing them via this array is now discouraged, and should not be
+relied upon in newly-written code as this ability may change in a future
+version.  Code that attempts to access the C<@_> array will produce
+warnings in the C<experimental::snail_in_signatures> category when
+compiled:
+
+    sub f ($x) {
+        print "Arguments are @_";
```
atoomic, line 491: maybe consider making it clear by adding a comment? so one reader can understand the snippet without reading the paragraph above (even if he should)
```
    print "Arguments are @_"; # throws a warning
```

t/op/signatures.t
```diff
@@ -1608,6 +1613,82 @@ while(<$kh>) {
             'f($1)';
 }
 
+# check that various uses of @_ inside signatured subs causes "experimental"
+# warnings at compiletime
+{
+    sub warnings_from {
+        my ($code, $run) = @_;
```
atoomic, line 1620: consider using signature? it's available here, no?
if so then same comment to all added helpers

atoomic (approved):

t/op/signatures.t
leonerd: Ah - for entirely selfish implementation reasons I didn't want to use any signatured subs for the implementation of the unit-test which tests signatured subs. It means I can add all sorts of "print to STDERR" or other messy debugging trickery, and not get fake noise from testing.

atoomic: I understand the fear, I jus noticed that we were using it above in this file

pod/perlsub.pod
leonerd: Added cc1d3fb980

t/op/signatures.t
jkeenan: I do not like any reference to snails for `@_`.  I can remember someone using that terminology at YAPC many years ago -- when I had already been doing Perl for many years -- and thinking, _"What is that guy talking about?"_

When I saw this in my email today, that was still my first reaction -- 22 years on in!  It's too "Perl insider talk" -- for me, and certainly for newcomers.  It's too self-consciously clever and too cute by half.  I object.

leonerd:
Oops. I found a case that doesn't warn yet:

```
$ ./perl -Ilib -Mexperimental=signatures -ce 'sub f($x = shift) { 123 }'
-e syntax OK
```

I'll switch to draft for now to work on that and then re-ready it when that's fixed.

t/op/signatures.t
Grinnz: I don't feel this matters. The only user facing component is the warning category, which the user will not care about unless they've been told to disable it, at which point they will know the name.

leonerd:
Right. That's solved, and actually the implementation is a bit neater now - it's all in the switch block in S_optimize_op, rather than being split between there and the opcheckers.

t/op/signatures.t
leonerd: @jkeenan As to the internal function names in the test file - I don't see that it matters. We have tonnes of "sub foo", "sub wibble" and all sorts else worse in .t files. Internal names within files aren't that interesting.

On the specific note of the warning name - it doesn't need to be discoverable, as the name of the warning category is never printed anywhere. The only place a user will find it is in the docs about the specific situation, where they can be told how to turn it off.

But in case there are better ideas there - do you have any better suggestions? My other potential thought was

```perl
no warnings 'experimental::argsarray_in_signatures';
```

but that's even longer a name.

atoomic: alternate ideas: `defav_in_signatures` or `poparray_in_signatures` ?

khwilliamson:
On 1/28/22 13:34, James E Keenan wrote:
> 
> 
> In t/op/signatures.t 
> <https://github.com/Perl/perl5/pull/19377#discussion_r794871286>:
> 
>> @@ -1608,6 +1613,82 @@ while(<$kh>) {
>               'f($1)';
>   }
>   
> +# check that various uses of @_ inside signatured subs causes "experimental"
> +# warnings at compiletime
> +{
> +    sub warnings_from {
> +        my ($code, $run) = @_;
> +        my $warnings = "";
> +        local $SIG{__WARN__} = sub { $warnings .= join "", @_ };
> +        my $cv = eval qq{ sub(\$x) { $code }} or die "Cannot eval() - $@";
> +        $run and $cv->(123);
> +        return $warnings;
> +    }
> +
> +    sub snailwarns_ok {
> 
> I do not like any reference to snails for |@_|. I can remember someone 
> using that terminology at YAPC many years ago -- when I had already been 
> doing Perl for many years -- and thinking, /"What is that guy talking 
> about?"/
> 
I agree.  It took me a while to figure out what was meant.

If we want to be obscure, "strudel" (used in some places for its name, 
see below) is more appetizing to me.

In Portugal this symbol means french kissing; equally obscure but more 
memorable, once understood, to many.

commat in HTML

Wikipedia says
"The at sign, @, is normally read aloud as "at"; it is also commonly 
called the at symbol, commercial at, address sign or alphasand.[citation 
needed] It is used as an accounting and invoice abbreviation meaning "at 
a rate of" ...

The absence of a single English word for the symbol has prompted some 
writers to use the French arobase[2] or Spanish and Portuguese arroba, 
or to coin new words such as ampersat[3] and asperand,[4] or the 
(visual) onomatopoeia strudel,[5] but none of these have achieved wide use.

Although not included on the keyboard of the earliest commercially 
successful typewriters, it was on at least one 1889 model[6] and the 
very successful Underwood models from the "Underwood No. 5" in 1900 
onward. It started to be used in email addresses in the 1970s, and is 
now routinely included on most types of computer keyboards.

t/op/signatures.t
leonerd: `defav` has a definite meaning to about the 6 people sufficiently familiar with perl core internals, and is just as meaningless to everyone else as `snail` is. Though 4 of those 6 people would point out we don't have a `PL_defav`, what we have is `GvAV(PL_defgv)` and I don't think that will encode as nicely in a warning category ;)

`poparray` I'm not even sure what it's about

leonerd:
From a chat on IRC, a better suggestion may be

```perl
no warnings 'experimental::args_array_with_signatures';
```

Votes? Upvote this comment or suggest your own.

t/op/signatures.t
atoomic: was trying to see the name of the C variable we used for `GvAV(PL_defgv)`
looks like we have a vote going on with a better choice, thanks

ilmari:
I think signature should be singular, i.e. `experimental::args_array_with_signature`.

leonerd:
@ilmari 
> I think signature should be singular,

I thought about that, but the original `use feature 'signatures';` is plural, so it sortof made sense.

PR #19380, opened 2022-01-29
Added function for WIN32

Jamaika1:


Leont:
This commit title doesn't seem to match the change

xenu:
What is the purpose of this change? The modified file is just a template. The actual code that sets the values of those constants is in the makefiles, for example [here](https://github.com/Perl/perl5/blob/ebfc9d5d0d50ffc852e4ae35a78e6a922ada8700/win32/GNUmakefile#L1281).

Jamaika1:
And here is this instruction. Thanks for the hint.

Issue #19382, opened 2022-01-30
Commit 9c91314886 breaks YVES/Sereal-Decoder-4.018.tar.gz on threaded perl


andk:
**Description**

During compilation of YVES/Sereal-Decoder-4.018.tar.gz with a threaded perl built with v5.35.5-32-g9c91314886 the following error is thrown:
``` text
cc -c  -I. -D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2 -O3 -Werror=declaration-after-statement    -DVERSION=\"4.018\" -DXS_VERSION=\"4.018\" -fPIC "-I/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.5-32-g9c91314886/1e0c/lib/5.35.6/x86_64-linux-thread-multi/CORE"  -DNDEBUG -DHAVE_ZSTD Decoder.c

In file included from /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.5-32-g9c91314886/1e0c/lib/5.35.6/x86_64-linux-thread-multi/CORE/perl.h:7135,
                 from Decoder.xs:6:
/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.5-32-g9c91314886/1e0c/lib/5.35.6/x86_64-linux-thread-multi/CORE/inline.h: In function ‘Perl_cop_file_avn’:
/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.5-32-g9c91314886/1e0c/lib/5.35.6/x86_64-linux-thread-multi/CORE/inline.h:3439:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 3439 |     const char *file = CopFILE(cop);
      |     ^~~~~
cc1: some warnings being treated as errors
make: *** [Makefile:368: Decoder.o] Error 1

```

I see same result with v5.35.8-28-gebfc9d5d0d given a threaded perl.

**Steps to Reproduce**

cpan -i Sereal::Decoder

**Expected behavior**

Module should be built, tested and installed

**Perl configuration**

```
Summary of my perl5 (revision 5 version 35 subversion 6) configuration:
  Commit id: 9c913148860b0e83e6149d37e86cdb29663ee812
  Platform:
[...]

PR #19384, opened 2022-01-30
Declaration needs to be at top of block

jkeenan:
For: https://github.com/Perl/perl5/issues/19382

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -3436,17 +3436,19 @@ Perl_cop_file_avn(pTHX_ const COP *cop) {
 
     PERL_ARGS_ASSERT_COP_FILE_AVN;
 
-    const char *file = CopFILE(cop);
-    if (file) {
-        GV *gv = gv_fetchfile_flags(file, strlen(file), GVF_NOADD);
-        if (gv) {
-            return GvAVn(gv);
-        }
-        else
-            return NULL;
-     }
-     else
-         return NULL;
+    {
+        const char *file = CopFILE(cop);
+        if (file) {
+            GV *gv = gv_fetchfile_flags(file, strlen(file), GVF_NOADD);
+            if (gv) {
+                return GvAVn(gv);
+            }
+            else
+                return NULL;
+         }
+         else
+             return NULL;
+    }
 }
 
 #endif
--- a/inline.h
+++ b/inline.h
@@ -3434,21 +3434,20 @@ Perl_sv_isbool(pTHX_ const SV *sv)
 PERL_STATIC_INLINE AV *
 Perl_cop_file_avn(pTHX_ const COP *cop) {
 
+    const char *file = CopFILE(cop);
+
     PERL_ARGS_ASSERT_COP_FILE_AVN;
 
-    {
-        const char *file = CopFILE(cop);
-        if (file) {
-            GV *gv = gv_fetchfile_flags(file, strlen(file), GVF_NOADD);
-            if (gv) {
-                return GvAVn(gv);
-            }
-            else
-                return NULL;
-         }
-         else
-             return NULL;
-    }
+    if (file) {
+        GV *gv = gv_fetchfile_flags(file, strlen(file), GVF_NOADD);
+        if (gv) {
+            return GvAVn(gv);
+        }
+        else
+            return NULL;
+     }
+     else
+         return NULL;
 }
 
 #endif


inline.h
```diff
@@ -3436,17 +3436,19 @@ Perl_cop_file_avn(pTHX_ const COP *cop) {
...
-        if (gv) {
-            return GvAVn(gv);
-        }
-        else
-            return NULL;
-     }
-     else
-         return NULL;
+    {
+        const char *file = CopFILE(cop);
```
Leont: IMO it would make more sense to just move up `const char *file = CopFILE(cop);` to come before `PERL_ARGS_ASSERT_COP_FILE_AVN;`

Leont:
Fundamentally this is a policy decision. Are we ok with C99 leaking out in our headers or not?

Having done a little [grepping](https://grep.metacpan.org/search?size=20&q=Wdeclaration-after-statement&qd=&qft=) I think we should not leak it for now.

Most of the issue is with dists using Module::Install:XSUtil, it's a gift that keeps on giving…

inline.h
jkeenan: Done.  When I now build with `gcc-12`, I can install `Sereal::Decoder` via `cpan` without build failures.  Please review.

tonycoz:
Perl now requires a C compiler that supports mixed declarations and code, upstream configures the compiler to not support that.

I think it should be fixed upstream.

WRT this patch, moving the declaration up means cop is dereferenced before it's asserted on, which is a minor regression (I expect there's similar problems in many other places though.)

Leont:
> Having done a little grepping I think we should not leak it for now.

On second thought, all the other dists only enable it as a warning. If Serial is the only thing declaring it an error, then we don't have much of an urgent issue.

demerphq:
On Mon, 31 Jan 2022, 07:18 Leon Timmermans, ***@***.***>
wrote:

> Perl now requires a C compiler that supports mixed declarations and code,
> upstream configures the compiler to not support that.
>
> I think it should be fixed upstream.
>
> I agree we should, but I'm not sure 4 months until 3.36 is enough time for
> that.
>

I don't mind changing this, but it feels ungenerous to penalize folks like
us who tried to keep our code clean and compliant with the old policy and
make sure our code compiles on as wide a range of compilers as possible.

Why can't the policy be that *headers* have to stay old policy compliant?
The rest of the internals can follow the new policy.

I'd say it's fair to say this is a regression. Sereal was fine until Perl
changed its policy.

Yves

>

jkeenan:
On 1/30/22 17:15, Tony Cook wrote:
> Perl now requires a C compiler that supports mixed declarations and 
> code, upstream configures the compiler to not support that.
> 
> I think it should be fixed upstream.
> 
> WRT this patch, moving the declaration up means cop is dereferenced 
> before it's asserted on, which is a minor regression (I expect there's 
> similar problems in many other places though.)
> 

Would my first commit -- which merely created a block in which the 
declaration was the first statement -- be desirable by itself?

demerphq:
On Mon, 31 Jan 2022, 08:44 James E Keenan, ***@***.***> wrote:

> On 1/30/22 17:15, Tony Cook wrote:
> > Perl now requires a C compiler that supports mixed declarations and
> > code, upstream configures the compiler to not support that.
> >
> > I think it should be fixed upstream.
> >
> > WRT this patch, moving the declaration up means cop is dereferenced
> > before it's asserted on, which is a minor regression (I expect there's
> > similar problems in many other places though.)
> >
>
> Would my first commit -- which merely created a block in which the
> declaration was the first statement -- be desirable by itself?


I'd just split the statement. Decl goes up top, assignment after the macro.

Yves

xenu:
>Why can't the policy be that *headers* have to stay old policy compliant?
The rest of the internals can follow the new policy.

I don't see the point in that. If perl was built using a C99 compiler, that means a C99 compiler can also be used for building modules.

demerphq:
On Mon, 31 Jan 2022, 12:10 xenu, ***@***.***> wrote:

> Why can't the policy be that *headers* have to stay old policy compliant?
> The rest of the internals can follow the new policy.
>
> I don't see the point in that. If perl was built using a C99 compiler that
> means a C99 compiler can also be used for building modules.
>

That is besides the point. Making headers c89 compliant (or whatever) means
modules like Sereal who took cross compilation seriously for years by
enforcing the gcc flag mentioned in this thread do not die like Sereal has
here, and it wouldn't inconvenience anybody who wants to build without the
flag either. The only affected parties would be a small group of perl
internals devs who touch the headers. It wouldn't even affect .c files in
core just .h files.

Yves

xenu:
I don't think it's reasonable to ban C99 (a 23 years old standard!) from headers forever because [a very small number of modules](https://grep.cpanauthors.org/search?q=-Werror%3Ddeclaration-after-statement) uses compiler flags that ban them.

If a module author really needs to keep supporting hypothetical users stuck with ancient compilers, those flags should be either changed to warnings or restricted to a CI environment.

nwc10:
For **all** CPAN it's **only** Sereal which faults this as an error, and Sereal/Sereal#265 was submitted last year to fix this. The other distributions which made it an error are updated. The rest are all warnings.

demerphq:
On Mon, 31 Jan 2022 at 07:42, Nicholas Clark ***@***.***>
wrote:

> For *all* CPAN it's *only* Sereal which faults this as an error, and
> Sereal/Sereal#265 <https://github.com/Sereal/Sereal/pull/265> was
> submitted last year to fix this. The other distributions which made it an
> error are updated. The rest are all warnings.
>

I already said I would make the change. Thanks for the PR, I had not
noticed that.

I'll roll a new Sereal release tomorrow. I have other priorities today.

Yves

demerphq:
On Mon, 31 Jan 2022 at 08:13, demerphq ***@***.***> wrote:

> On Mon, 31 Jan 2022 at 07:42, Nicholas Clark ***@***.***>
> wrote:
>
>> For *all* CPAN it's *only* Sereal which faults this as an error, and
>> Sereal/Sereal#265 <https://github.com/Sereal/Sereal/pull/265> was
>> submitted last year to fix this. The other distributions which made it an
>> error are updated. The rest are all warnings.
>>
>
> I already said I would make the change. Thanks for the PR, I had not
> noticed that.
>
> I'll roll a new Sereal release tomorrow. I have other priorities today.
>
>
Well, it's a week late, but I just pushed to PAUSE now. 4.019, with zstd
updates to 1.5.1

Yves

-- 

jkeenan:
> On 1/30/22 17:15, Tony Cook wrote: Perl now requires a C compiler that supports mixed declarations and code, upstream configures the compiler to not support that. I think it should be fixed upstream. WRT this patch, moving the declaration up means cop is dereferenced before it's asserted on, which is a minor regression (I expect there's similar problems in many other places though.)
> Would my first commit -- which merely created a block in which the declaration was the first statement -- be desirable by itself?

@tonycoz, can I ask you to make an executive decision on this p.r.?  What, if anything, should be committed to blead?

Thank you very much.
Jim Keenan

hvds:
> @tonycoz, can I ask you to make an executive decision on this p.r.? What, if anything, should be committed to blead?

I'm not Tony, but: blead is fine as is, nothing here should be committed. The only module for which the current code caused a problem has had a new version released so that it no longer causes a problem. I'd recommend withdrawing this PR.

ailin-nemui:
hehe, we just stumbled upon this:

https://gitweb.gentoo.org/repo/gentoo.git/commit/net-irc/irssi/files/irssi-1.4.1-no-werror-decl.patch?id=d5bde60aca1e5821685c1488d88edf16228eea84

PR #19386, opened 2022-02-01
Use sha256 instead of sha1 in release announcement

mbeijen:
The last time perl got released, there was this comment:

https://www.nntp.perl.org/group/perl.perl5.porters/2022/01/msg262480.html

> Why use the old ( and somewhat broken ) SHA1 for message digests?

I decided to look into using SHA256;
I notice that the release managers guide gets the SHA1s from the PAUSE
email. So I've made a PR from PAUSE so that they no longer send SHA1s
(or MD5SUMS for that matter) but just SHA256. That change is live
already, see https://github.com/andk/pause/pull/379

So now the next step is to update the perl release documentation to deal
with the new reality that PAUSE mails no longer incluse sha1s!

### DIFF
--- a/Porting/release_announcement_template.txt
+++ b/Porting/release_announcement_template.txt
@@ -10,10 +10,10 @@ favorite CPAN mirror or find it at:
 
 https://metacpan.org/release/[AUTHOR]/perl-5.[VERSION.SUBVERSION]/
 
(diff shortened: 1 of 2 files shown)


atoomic:
excellent thanks for making this happen!

PR #19387, opened 2022-02-02
Fixes for bool because of C99

khwilliamson:


leonerd:
Overall this seems reasonable. `<stdbool.h>` is C99 and we're requiring C99 now, so there ought not be any weird compatibility problems. Though I'm no expert in the uncommon platforms like VMS or so on. Might be useful to get some smoke-testing out of those places too before we commit on this.

The cygwin smoker looks unhappy, but it appears to be our old friend `t/op/alarm.t` and looks unrelated to this change; if anything I'd expect this change to fail at compiletime and it compiled fine on cygwin.

xenu (approved):

xenu:
In C++ booleans are built-in, so it would make sense to include `stdbool.h` conditionally:
```
#ifndef __cplusplus
#  include <stdbool.h>
#endif
```

It's just pedantry and it doesn't really matter. The code will work either way.

PR #19396, opened 2022-02-05
Deprecate downgrading use VERSION below v5.11

leonerd:
Print a deprecation warning if a `use VERSION` requesting a version below v5.11 is encountered, after a `use VERSION` of a higher version is already in scope.

This is because versions v5.11 or higher will enable the strictness hints as if loaded by `use strict`, and earlier versions will clear them. The model for this is complicated by the second set of HINT_EXPLICIT_STRICT_* bits, which have now been decided are a bad design and we'd like to remove them.

Once merged, I will follow this up with another change which actually removes the HINT_EXPLICIT_STRICT bits, freeing up three more hints bits for useful work. Removing the hints bits would slightly change the behaviour in this scenario, therefore we should print a deprecation warning on situations that might be affected.

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -2858,10 +2858,10 @@ last-inclusive range.
                                             "Use of " s " is deprecated")
 #  define deprecate_disappears_in(when,message) \
               Perl_ck_warner_d(aTHX_ packWARN(WARN_DEPRECATED),    \
-                               message ", and will disappear in Perl " when)
+                               message " is deprecated, and will disappear in Perl " when)
 #  define deprecate_fatal_in(when,message) \
               Perl_ck_warner_d(aTHX_ packWARN(WARN_DEPRECATED),    \
-                               message ". Its use will be fatal in Perl " when)
+                               message " is deprecated, and will become fatal in Perl " when)
 #endif
 
 /* Internal macros to deal with gids and uids */
--- a/lib/strict.t
+++ b/lib/strict.t
@@ -3,7 +3,7 @@
 chdir 't' if -d 't';
 @INC = ( '.', '../lib' );
 
-our $local_tests = 6;
+our $local_tests = 7;
 require "../t/lib/common.pl";
 
 eval qq(use strict 'garbage');
@@ -18,8 +18,15 @@ like($@, qr/^Unknown 'strict' tag\(s\) 'foo bar'/);
 eval qq(no strict qw(foo bar));
 like($@, qr/^Unknown 'strict' tag\(s\) 'foo bar'/);
 
-eval 'use v5.12; use v5.10; ${"c"}';
-is($@, '', 'use v5.10 disables implicit strict refs');
+{
+    my $warnings = "";
+    local $SIG{__WARN__} = sub { $warnings .= $_[0] };
+    eval 'use v5.12; use v5.10; ${"c"}';
+    is($@, '', 'use v5.10 disables implicit strict refs');
+    like($warnings,
+        qr/^Downgrading a use VERSION declaration to below v5.11 is deprecated, and will become fatal in Perl 5.40 at /,
+        'use v5.10 after use v5.12 provokes deprecation warning');
+}
 
 eval 'use strict; use v5.10; ${"c"}';
 like($@,
--- a/op.c
+++ b/op.c
@@ -8842,7 +8842,11 @@ Perl_package_version( pTHX_ OP *v )
     op_free(v);
 }
 
-/* grrrr */
+/* Extract the first two components of a "version" object as two 8bit integers
+ * and return them packed into a single U16 in the format of PL_prevailing_version.
+ * This function only ever has to cope with version objects already known
+ * bounded by the current perl version, so we know its components will fit
+ * (Up until we reach perl version 5.256 anyway) */
 static U16 S_extract_shortver(pTHX_ SV *sv)
 {
     SV *rv;
@@ -8867,6 +8871,7 @@ static U16 S_extract_shortver(pTHX_ SV *sv)
 
     return shortver;
 }
+#define SHORTVER(maj,min) ((maj << 8) | min)
 
 void
 Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
@@ -8969,6 +8974,10 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
         }
         /* otherwise they are off */
         else {
+            if(PL_prevailing_version >= SHORTVER(5, 11))
+                deprecate_fatal_in("5.40",
+                    "Downgrading a use VERSION declaration to below v5.11");
+
             if (!(PL_hints & HINT_EXPLICIT_STRICT_REFS))
                 PL_hints &= ~HINT_STRICT_REFS;
             if (!(PL_hints & HINT_EXPLICIT_STRICT_SUBS))
--- a/op.c
+++ b/op.c
@@ -8842,6 +8842,32 @@ Perl_package_version( pTHX_ OP *v )
     op_free(v);
 }
 
+/* grrrr */
+static U16 S_extract_shortver(pTHX_ SV *sv)
+{
+    SV *rv;
+    if(!SvRV(sv) || !SvOBJECT(rv = SvRV(sv)) || !sv_derived_from(sv, "version"))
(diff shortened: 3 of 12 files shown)


leonerd:
I admit that this code does add a `PL_prevailing_version` that becomes visible at compiletime, and the temptation may be there in future to write compiletime code that directly compares this value to some limit in order to make guesses about intended behaviour. That is a slippery slope we'll have to consider on its own merits later on. If we don't want to implement it this way, we could instead just store a single bitflag to remember if the prevailing version is at least 5.11. I rejected that as an idea as it's almost as complex as this solution, while being significantly less flexible in future.

jkeenan:
Tested branch on FreeBSD-12; tests PASS.

lib/strict.t
```diff
@@ -18,8 +18,14 @@ like($@, qr/^Unknown 'strict' tag\(s\) 'foo bar'/);
...
 
-eval 'use v5.12; use v5.10; ${"c"}';
-is($@, '', 'use v5.10 disables implicit strict refs');
+{
+    my $warnings = "";
+    local $SIG{__WARN__} = sub { $warnings .= $_[0] };
+    eval 'use v5.12; use v5.10; ${"c"}';
+    is($@, '', 'use v5.10 disables implicit strict refs');
+    ok($warnings =~ m/^Downgrading a use VERSION declaration to below v5.11 is now deprecated at /,
+        'use 5.01 after use 5.012 provokes deprecation warning');
```
ilmari: The use of different version number formats (especially the truncated one) in the test description from the actual test is confusing.

op.c
```diff
@@ -8941,13 +8974,19 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
         }
         /* otherwise they are off */
         else {
+            if(PL_prevailing_version >= SHORTVER(5, 11))
+                Perl_ck_warner_d(aTHX_ packWARN(WARN_DEPRECATED),
+                    "Downgrading a use VERSION declaration to below v5.11 is now deprecated");
```
ilmari: The "now" is redundant, and the warning should say which release it will be removed in.

op.c
```diff
@@ -8926,6 +8957,8 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
         use_version = sv_2mortal(new_version(use_version));
         S_enable_feature_bundle(aTHX_ use_version);
 
+        U16 shortver = S_extract_shortver(aTHX_ use_version);
+
         /* If a version >= 5.11.0 is requested, strictures are on by default! */
         if (vcmp(use_version,
                  sv_2mortal(upg_version(newSVnv(5.011000), FALSE))) >= 0) {
```
ilmari: Seeing as we now have the `shortver` here, we could avoid constructing a version object just for the comparison.

ilmari (changes requested):
The deprecation should also be documented in `pod/perldeprecation.pod`.

op.c
leonerd: Yes I did think of that as a second step. Not strictly related, but might as well while we're here.

lib/strict.t
leonerd: Oops I think that may be a copy-paste error. I'll fix it up

leonerd:
PTAL. Most comments fixed.

I've mentioned it in perldeprecation.pod but so far I haven't changed the message itself to mention the cutoff date - are there any existing precedents of a message doing that, which I could steal?

ilmari:
> I've mentioned it in perldeprecation.pod but so far I haven't changed the message itself to mention the cutoff date - are there any existing precedents of a message doing that, which I could steal?

The policy of specifying the cut-off date was introcuded in 5.26, so its [perldelta](https://perldoc.perl.org/perl5260delta#Changes-to-Existing-Diagnostics) lists a whole pile of them (which have since mostly been removed, as promised).

There's some handy (hah!) macros for this [in `handy.h`](https://github.com/Perl/perl5/blob/23805bfc0001e743c0dd0a951812bddab402d7e7/handy.h#L2856-L2865).

op.c
```diff
@@ -8926,28 +8957,35 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
...
 
-            if (vcmp(use_version, sv_2mortal(upg_version(newSVpvs("5.035000"), FALSE))) >= 0)
+            if (shortver >= SHORTVER(5, 35))
                 free_and_set_cop_warnings(&PL_compiling, pWARN_ALL);
         }
         /* otherwise they are off */
         else {
+            if(PL_prevailing_version >= SHORTVER(5, 11))
+                deprecate_disappears_in("5.40",
+                    "Downgrading a use VERSION declaration to below v5.11 is deprecated");
```
ilmari: The "disappears" wording seems a bit odd here, I would expect "will be fatal in", but the `deprecate_fatal_in()` wording isn't quite right either. Seeing as those are `PERL_CORE` only, and only `deprecate()` is actually used, feel free to reword them.

leonerd: That's fair. I'll have a think on some better wording.

leonerd:
Message now adjusted. PTAL

handy.h
```diff
@@ -2861,7 +2861,7 @@ last-inclusive range.
                                message ", and will disappear in Perl " when)
 #  define deprecate_fatal_in(when,message) \
               Perl_ck_warner_d(aTHX_ packWARN(WARN_DEPRECATED),    \
-                               message ". Its use will be fatal in Perl " when)
+                               message ", and will become fatal in Perl " when)
```
ilmari: Shouldn't this (and `deprecate_disappears_in`) include the " is deprecated" part of the message too?

leonerd: Ahyes; that's fair. Adjusted and repushed.

pod/perldelta.pod
```diff
@@ -77,6 +77,33 @@ XXX For a release on a stable branch, this section aspires to be:
 
 XXX Any deprecated features, syntax, modules etc. should be listed here.
 
+=head2 Downgrading a C<use VERSION> statement to below v5.11
+
+Attempting to issue a second C<use VERSION> statement that requests a version
+below than C<v5.11> when an earlier statement that requested a version higher
```
wolfsage: `requests a version below than...` doesn't quite make sense. Maybe `requests a version lower than` (later we say `higher than`

pod/perldelta.pod
```diff
@@ -77,6 +77,33 @@ XXX For a release on a stable branch, this section aspires to be:
...
+
+For example:
+
+    use v5.14;
+    say "The say statement is permitted";
+    use v5.8;                               # This will print a warning
+    print "We must use print\n";
+
+This is because of an intended related change to the interaction between
+C<use VERSION> and C<use strict>. Code which wishes to mix versions in this
```
wolfsage: This mentions a relation between VERSION/strict but doesn't say what it is. I suggest:
```
This is because of an intended related change to the interaction between C<use VERSION> and C<use strict>. If you specify a version >= 5.11, strict is enabled automatically for you. If you request a version <= 5.11, strict is disabled for you.

File: pod/perldeprecation.pod [hunk 8]
```diff
@@ -14,6 +14,22 @@ features are available.
 The deprecated features will be grouped by the version of Perl in
 which they will be removed.
 
+=head2 Perl 5.40
+
+=head3 Downgrading a C<use VERSION> to below v5.11
+
+Once Perl has seen a C<use VERSION> declaration that requests a version
+C<v5.11> or above, a subsequent second declaration that requests an earlier
+version will print a deprecation warning. For example,
```
wolfsage, line 23: Once again it'd be nice if this spelled out why we are deprecating this

wolfsage: I see it below in perldiag, but if the reason was in all places we spelled it out (or a link of where to find the reason)

wolfsage (commented):
One bug in the docs I believe, and some nits

pod/perldelta.pod
leonerd: Ahyes.

pod/perldeprecation.pod
leonerd: Much wording added and re-pushed.

wolfsage:
lgtm! Thanks

ilmari (approved):

PR #19400, opened 2022-02-07
Deëxperiment feature isa

leonerd:
This has remained unchanged for two stable releases now, and no further alterations are planned.

Lets stablise it.


Leont (approved):

pod/perldelta.pod
```diff
@@ -55,6 +55,13 @@ Z<>
 The behaviour of code which attempts to do this is no longer specified, and
 may be subject to change in a future version.
 
+=head2 The C<isa> operator is no longer experimental
+
+Introduced in Perl version 5.32.0, this operator has remained unchanged
+since then. The operator is now considered a stable languauge feature.
```
ilmari: Maybe mention that it's included in the `:5.36` feature bundle too?

pod/perlop.pod
```diff
@@ -673,10 +673,6 @@ scalar expression that yields a string class name:
     if( $obj isa "Different::Class" ) { ... }
     if( $obj isa $name_of_class ) { ... }
 
-This is an experimental feature and is available from Perl 5.31.6 when enabled
-by C<use feature 'isa'>. It emits a warning in the C<experimental::isa>
-category.
```
ilmari, line 678: This should still mention that it's only available when the `isa` feature is enabled, and that it's enabled by default under `use v5.36;` or higher, like [`perlfunc/say`](https://perldoc.perl.org/functions/say) does.

ilmari (changes requested):
The `perlexperiment.pod` entry should be moved to the "Accepted experiments` section.

pod/perldelta.pod
ilmari: Or that it's enabled by default under `use v5.36;`, to encourage that rather than manual feature enabling.

ilmari:
If you put "closes #18754" in one of the commit messages, it will automatically close the experiment tracking issue, and be linked from there.

pod/perlop.pod
leonerd: done

leonerd:
All done. PTAL

pod/perlop.pod
ilmari: Feature-gated functions, such as [`say`](https://perldoc.perl.org/functions/say) and [`fc`](https://perldoc.perl.org/functions/fc), refer to `use v5.xx;` rather than feature bundles, so I think the wording here (and in the perldelta entry) should be consistent with that.

leonerd: Ahyes. Further wording fixes to be similar to the `say` docs.

haarg:
Aren't experimental warning categories normally left in place once the feature is stabilized? Otherwise code that tries to disable the warning will start throwing errors.

leonerd:
@haarg Ahyes you're right (e.g. `experimental::lexical_subs`). I'll put the category itself back in place.

ilmari (approved):
LGTM

leonerd:
All three Windows-like smokers (msvc142, mingw64, cygwin) have got upset at this. Each in a totally different way that all seem unrelated to the (mostly-documentation) changes being made here. I think I will merge despite these failures.

PR #19402, opened 2022-02-08
Document that $^S is 1 inside a try block

leonerd:


pod/perlvar.pod
```diff
@@ -1840,7 +1840,7 @@ Current state of the interpreter.
     $^S         State
     ---------   -------------------------------------
     undef       Parsing module, eval, or main program
-    true (1)    Executing an eval
+    true (1)    Executing an eval or try block
```
ilmari, line 1843: This makes it sound like it's only `1` when executing an `eval` block, not a string `eval`. Maybe something like

```
undef     Parsing module, eval string, or main program
true (1)  Executing try block or eval string or block.
```

Grinnz: Or leave out "block": "Executing an eval or try"

Issue #19340, opened 2022-01-11
Segfault in simple program reading from ARGV


sisyphus-ppcg:
**Steps to Reproduce**
Run on your command line.

```bash
perl -e 'map<>,@ARGV' 1 2 < /dev/null
```

**Expected behavior**
No segfault.

**Actual behavior**
```
sisyphus$ perl -e 'map<>,@ARGV' 1 2 < /dev/null
Can't open 1: No such file or directory at -e line 1.
Can't open 2: No such file or directory at -e line 1.
Segmentation fault
```

Note that the segfault happens regardless of whether the files in `@ARGV` exist or not.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 34 subversion 0) configuration:

  Platform:
    osname=linux
    osvers=4.19.128-microsoft-standard
    archname=x86_64-linux
    uname='linux laptop-u12kikmv 4.19.128-microsoft-standard #1 smp tue jun 23 12:58:10 utc 2020 x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dprefix=/home/sisyphus/perl5/perlbrew/perls/perl-5.34.0 -Aeval:scriptdir=/home/sisyphus/perl5/perlbrew/perls/perl-5.34.0/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='9.3.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
[...]

Issue #12315, opened 2012-08-01
Panic in pure-Perl code with vanilla perl-5.16.0 from perlbrew


p5pRT:
Migrated from [rt.perl.org#114372](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=114372) (status was 'open')

Searchable as RT114372$

p5pRT:
### From @shlomif

#### Created by @shlomif



This Bash script causes a panic with vanilla perl\-5\.16\.0 \(installed
using perlbrew \- without threads\) on Mageia Linux 3/Cauldron on x86\-64&#8203;:

\<\<\<\<
\#\!/bin/bash
echo \-e '\[2\,36\,29\,83\]\\n\[30\,81\,55\,15\,68\,26\,71\,29\,6\,46\]\\n\[16\,66\,42\,76\,59\]\\n\[65\,47\,65\,47\,74\,33\,41\,38\,75\,59\]\\n\[64\,55\,39\,26\,33\,66\]\\n' |
&nbsp;    perl \-pe '@&#8203;h\{@&#8203;x=sort\{$a\<=>$b\}/\\d\+/g\}=reverse@&#8203;x;s\#\\d\+\#$h\{$&\}\#g'
>>>>

The code does not make a lot of sense\, but it was a starting point for working
on this Perl Golf challenge&#8203;:

http&#8203;://golf\.shinh\.org/p\.rb?MaxMinSwap

When running it I get&#8203;:

shlomif@&#8203;telaviv1&#8203;:\~$ bash reproduce\.bash
\[\,\,\,\]
\[\,\,\,26\,\,15\,\,6\,29\,\]
\[\,\,\,\,\]
\[\,33\,\,33\,\,47\,41\,47\,\,\]
panic&#8203;: attempt to copy freed scalar 81b2f8 at \-e line 1\, \<> line 5\.

Further lines are not processed\.

Please look into fixing it\.

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=high

Site configuration information for perl 5.16.0:

Configured by shlomif at Wed Aug  1 11:45:44 IDT 2012.

Summary of my perl5 (revision 5 version 16 subversion 0) configuration:
   
  Platform:
    osname=linux, osvers=3.5.0-desktop-1.mga3, archname=x86_64-linux
[...]

PR #19403, opened 2022-02-08
Make map and grep preserve their args lists.

FGasper:
Issue #19340: This prevents an obscure segfault.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -6166,6 +6166,7 @@ t/re/uniprops09.t		Test unicode \p{} regex constructs
 t/re/uniprops10.t		Test unicode \p{} regex constructs
 t/re/user_prop_race_thr.t	Test races in user-defined \p{} under threads
 t/README			Instructions for regression tests
+t/run/argv_free.t		Ensure no conflicts between @ARGV and <>.
 t/run/cloexec.t			Test close-on-exec.
 t/run/dtrace.pl			For dtrace.t
 t/run/dtrace.t			Test for DTrace probes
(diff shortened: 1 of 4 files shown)


Leont:
I don't think leaking really is the best solution we can offer here.

FGasper:
@Leont Given how small ARGV usually is I didn’t think it unreasonable, though obviously preserving the cleanup would be best for all. Do you have any suggestions of where to start? Some way to prevent the SAVEFREESV from nextargv from conflicting with the leave-time free of the saved $_?

Leont:
This should have been a new PR, now my first comment doesn't make sense anymore.

I don't think this is going to fly either. This is one of the more difficult problems that we have, please don't try to quick-fix it.

FGasper:
@Leont I don't mean to oversimplify, certainly. But can you give a bit more detail as to what else needs to happen? This approach doesn't incur leakage and does solve both of the reproductions in the issue thread.

I don't doubt that there are bigger problems in play; I'd just like to have a better sense of those. Thank you for your time.

xenu:
This is a good starting point: https://github.com/Perl/perl5/issues/12315#issuecomment-544033274

xenu:
Also see [this](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=77706) meta ticket for an incomplete list of related tickets (the dependsOn section).

FGasper:
@xenu: Thank you for the clarification and broader context.

I would still think that fixing 30% of the cases (i.e., map and grep) is better than solving none of them. To @demerphq’s point, though, this really is a “stop hitting yourself” thing; there seems to be no reason why normal development would result in such code being written, let alone being put into production.

Anyhow, I‘ll close this.

PR #19404, opened 2022-02-09
String "Inf" should be consistently converted to IV/UV

t-a-k:
I found that using string `"Inf"` in contexts that requires IV/UV is converted to wrong results for the first time:
```
$ perl -wle 'my $x = "Inf"; print $x << 0; print $x << 0; print $x << 0'
0
18446744073709551615
18446744073709551615
```
To make matters worse, `"Inf"` seems to "leak" the previous value of the variable:
```
$ perl -wle 'my $x = 42; $x = "Inf"; print $x << 0; print $x << 0'
42
18446744073709551615
```
This PR will fix this problem (and add testcases).

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -2294,6 +2294,7 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
         if (SvTYPE(sv) == SVt_NV)
             sv_upgrade(sv, SVt_PVNV);
 
+    got_nv:
         (void)SvIOKp_on(sv);	/* Must do this first, to clear any SvOOK */
         /* < not <= as for NV doesn't preserve UV, ((NV)IV_MAX+1) will almost
            certainly cast into the IV range at IV_MAX, whereas the correct
@@ -2418,7 +2419,7 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
             if (ckWARN(WARN_NUMERIC) && ((numtype & IS_NUMBER_TRAILING)))
                 not_a_number(sv);
             S_sv_setnv(aTHX_ sv, numtype);
-            return FALSE;
+            goto got_nv;        /* Fill IV/UV slot and set IOKp */
         }
 
         /* If NVs preserve UVs then we only use the UV value if we know that
--- a/t/op/infnan.t
+++ b/t/op/infnan.t
@@ -578,9 +578,6 @@ cmp_ok('-1e-9999', '==', 0,     "underflow to 0 (runtime) from neg");
 
 # "-Inf" should be converted to IV consistently
 {
-    # Enclosed in a string eval so that majority of infnan.t are run
-    # before loading integer.pm, just in case.
-    eval <<'EOC';
     use integer;
     my $x = '-Inf';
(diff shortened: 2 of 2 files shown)


sv.c
```diff
@@ -2418,7 +2419,7 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
             if (ckWARN(WARN_NUMERIC) && ((numtype & IS_NUMBER_TRAILING)))
                 not_a_number(sv);
             S_sv_setnv(aTHX_ sv, numtype);
-            return FALSE;
+            goto got_nv;        /* Fill IV/UV slot and set IOKp */
```
hvds, line 2422: I appreciate this should be safe with current core, but it takes some thinking to validate that and feels like a trap that could cause problems in the future. A viable alternative might be to duplicate the `(void)SvIOKp_on(sv)` and move the label to line 2353 (ie the `else` branch of `if (SvNVX(sv) < (NV)IV_MAX + 0.5)`.

t-a-k: I think that it should not jump into the `else` branch directly, because NV may be `-Inf` here.  Furthermore, doing so also skips the `NAN_COMPARE_BROKEN` workaround.

hvds: Ah you're quite right, I failed to pick up the right bits of context; looking at the function as a whole this makes sense.

hvds:
This looks good to me, I'll merge in a day or two if there are no other comments.

(We may also want to reconsider what `-Inf` should yield in integer context, but that's not the job of this PR.)

t-a-k:
Sorry to trouble you, but I just added another test to `t/op/infnan.t` for checking conversion of `"-Inf"`.

This test might have to be revised in the future if some change were made as the result of this reconsideration...

> (We may also want to reconsider what `-Inf` should yield in integer context, but that's not the job of this PR.)

t/op/infnan.t
```diff
@@ -576,4 +576,20 @@ cmp_ok('-1e-9999', '==', 0,     "underflow to 0 (runtime) from neg");
     is($x, ' Inf ', "string shall be ' Inf ' after UV/NV conversion");
 }
 
+# "-Inf" should be converted to IV consistently
+{
+    # Enclosed in a string eval so that majority of infnan.t are run
+    # before loading integer.pm, just in case.
+    eval <<'EOC';
```
hvds: Just in case of what? If there's a risk of `use integer` having buggy effects outside its scope, that's not really the responsibility of this test file to cope with. But I see no reason to believe there is such a risk - and certainly we don't use such caution in other much more fundamental test files.

(Also if there were good reason to keep the eval, you should at the very least check for it dying. But far better to remove it.)

t-a-k: > If there's a risk of `use integer` having buggy effects outside its scope, that's not really the responsibility of this test file to cope with.

That makes sense.  I was too worried about becoming the first to `use integer` in `infnan.t`.  I'll push a commit to remove `eval`.

hvds:
Thanks, I rebased, collapsed the last pair of commits, and have now pushed as 7e511f6af1 and 573a2c7380.

PR #19406, opened 2022-02-10
use source::encoding

khwilliamson:
I'm having trouble thinking of diabolical tests for this, so please try some out on this

### DIFF
--- a/lib/source/encoding.pm
+++ b/lib/source/encoding.pm
@@ -1,0 +1,79 @@
+package source::encoding;
+
+use strict;
+use warnings;
+
+our $VERSION = '0.01';
+
+our $ascii_hint_bits = 0x40000000;
+
+sub import {
+    my (undef, $arg) = @_;
+    require utf8;
+    if (lc $arg eq 'utf8') {
+        utf8->import;
+        return;
+    }
+    elsif (lc $arg eq 'ascii') {
+        utf8->unimport;
+        $^H |= $source::encoding::ascii_hint_bits;
+        return;
+    }
+
+    die "Bad argument for source::encoding: $arg";
+}
+
+sub unimport {
+    # If we don't guard against it, somebody will write:
+    #
+    #   use source::encoding 'utf8';
+    #   no source::encoding 'ascii';
+    #
+    # ...and then complain.
+    die q{No arguments permitted to "no source::encoding"} if @_ > 1;
+
+    require utf8;
+    utf8->unimport;
+    $^H &= ~$source::encoding::ascii_hint_bits;
+}
+
+1;
+__END__
+
+=head1 NAME
+
+source::encoding -- Declare Perl script's encoding
+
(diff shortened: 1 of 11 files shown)


toke.c
```diff
@@ -775,6 +777,42 @@ S_cr_textfilter(pTHX_ int idx, SV *sv, int maxlen)
...
 #endif
 
+STATIC void
+S_yyerror_non_ascii_message(pTHX_ const U8 * const s)
+{
+    PERL_ARGS_ASSERT_YYERROR_NON_ASCII_MESSAGE;
+
+    yyerror_pv(Perl_form(aTHX_ "Use of non-ASCII character 0x%02X"
+                               " illegal when 'use source::encoding"
+                               " \"ASCII\"' is in effect", *s), 0);
```
leonerd: I'd suggest putting a linebreak in after the `aTHX_` so you can start the string literal much closer to the left, and hence use two longer lines rather than these three tightly-wrapped ones.

pod/perldiag.pod
```diff
@@ -7357,6 +7357,10 @@ See L<perlunicode/Wildcards in Property Values>.
 it already went past any symlink you are presumably trying to look for.
 The operation returned C<undef>.  Use a filename instead.
 
+=item Use of non-ASCII character 0x%X illegal when 'use source::encoding "ASCII"' is in effect
```
leonerd: The pragma module spells it `"ascii"` in lower-case. I don't have any strong opinion on which case it should be, but this message ought to agree with the module itself.

lib/source/encoding.pm
```diff
@@ -0,0 +1,77 @@
...
+  
+    die "Bad argument for source::encoding: $arg";
+}
+
+sub unimport {
+    $^H &= ~$source::encoding::ascii_hint_bits;
+    utf8->unimport;
+}
+
+our $AUTOLOAD;
```
Leont: Why this autoload here?

khwilliamson: cargo culting

toke.c
khwilliamson: ok

pod/perldiag.pod
khwilliamson: I didn't mention that case doesn't matter.  I think that's the best option.  And I could mention it, or not; but if I don't, then it ought to be consistently cased, yes

Grinnz: I think case should matter. Accepting extra ways to specify the same thing just makes the interface more complex and increases future design debt.

khwilliamson: change to make lower case required

lib/source/encoding.pm
khwilliamson: removed

khwilliamson:
Other proposals have made this obsolete, so am closing it

PR #19412, opened 2022-02-13
Sort tests in t/porting/dual-life.t

hvds:
I notice a couple of openSUSE smokes are giving fails on this test (eg https://perl5.test-smoke.org/report/5015678), but since the order of the tests is determined by `File::Find::find` I don't know which utils it is actually complaining about.

This simple change gives a consistent but non-obvious order: we're sorting on eg
```
../cpan/Pod-Checker/scripts/podchecker.PL
../cpan/podlators/scripts/pod2text
../cpan/IO-Compress/bin/streamzip
../ext/Pod-Html/bin/pod2html 
...
```
rather than what appears in the test legends:
```
Verify -f '../cpan/Pod-Checker/podchecker'
Verify -f '../cpan/podlators/scripts/pod2text'
Verify -f '../utils/streamzip'
Verify -f '../utils/pod2html'
...
```

Since the list changes only rarely, for smoke failures after this is committed that should be enough to work out which file is being complained about by comparing a local verbose run.

### DIFF
--- a/t/porting/dual-life.t
+++ b/t/porting/dual-life.t
@@ -61,7 +61,7 @@ find(
 );
 
 
-for my $f ( @programs ) {
+for my $f ( sort @programs ) {
   $f =~ s/\.\z// if $^O eq 'VMS';
   next if $f =~ $not_installed;
   my $bn = basename($f);


jkeenan:
> I notice a couple of openSUSE smokes are giving fails on this test (eg https://perl5.test-smoke.org/report/5015678), but since the order of the tests is determined by `File::Find::find` I don't know which utils it is actually complaining about.
> 
> This simple change gives a consistent but non-obvious order: we're sorting on eg
> 
> ```
> ../cpan/Pod-Checker/scripts/podchecker.PL
> ../cpan/podlators/scripts/pod2text
> ../cpan/IO-Compress/bin/streamzip
> ../ext/Pod-Html/bin/pod2html 
> ...
> ```
> 
> rather than what appears in the test legends:
> 
> ```
> Verify -f '../cpan/Pod-Checker/podchecker'
> Verify -f '../cpan/podlators/scripts/pod2text'
> Verify -f '../utils/streamzip'
> Verify -f '../utils/pod2html'
> ...
> ```
> 
> Since the list changes only rarely, for smoke failures after this is committed that should be enough to work out which file is being complained about by comparing a local verbose run.

While I think the proposed change does no harm, before applying I think we should first see if @Tux can provide us more information about the tests which are failing.

atoomic (approved):
I see no reasons to do not merge it and wait for someone else?

hvds:
> While I think the proposed change does no harm, before applying I think we should first see if @Tux can provide us more information about the tests which are failing.

I since found that it actually does show the verbose output in the fuller log file, which makes me think that the problem is related to the simultaneous `op/glob.t` failure:
```
# Failed test 12 - Verify -f '../cpan/Pod-Checker/podchecker' at porting/dual-life.t line 70
# Failed test 13 - Verify -f '../cpan/Pod-Usage/pod2usage' at porting/dual-life.t line 70
```

But I think in any case it makes sense to have some sort of order on these tests.

khwilliamson:
On 2/13/22 09:01, James E Keenan wrote:
>     I notice a couple of openSUSE smokes are giving fails on this test
>     (eg https://perl5.test-smoke.org/report/5015678
>     <https://perl5.test-smoke.org/report/5015678>), but since the order
>     of the tests is determined by |File::Find::find| I don't know which
>     utils it is actually complaining about.
> 
>     This simple change gives a consistent but non-obvious order: we're
>     sorting on eg
> 
>     |../cpan/Pod-Checker/scripts/podchecker.PL
>     ../cpan/podlators/scripts/pod2text ../cpan/IO-Compress/bin/streamzip
>     ../ext/Pod-Html/bin/pod2html ... |
> 
>     rather than what appears in the test legends:
> 
>     |Verify -f '../cpan/Pod-Checker/podchecker' Verify -f
>     '../cpan/podlators/scripts/pod2text' Verify -f '../utils/streamzip'
>     Verify -f '../utils/pod2html' ... |
> 
>     Since the list changes only rarely, for smoke failures after this is
>     committed that should be enough to work out which file is being
>     complained about by comparing a local verbose run.
> 
> While I think the proposed change does no harm, before applying I think 
> we should first see if @Tux <https://github.com/Tux> can provide us more 
> information about the tests which are failing.
> 

I believe Tux will be overwhelmed by $work until March 16 at the earliest.

PR #19414, opened 2022-02-13
 Make newSV_type() an inline function

richardleach:
Note: This is an alternative to #19381, following feedback from @xenu.

When a new SV is created and upgraded to a type known at compile time,
using the general-purpose upgrade function (`sv_upgrade`) is clunky. 
Specifically, while uprooting a SV head is lightweight (assuming there are 
unused SVs), `sv_upgrade` is too big to be inlined, contains many branches
that can logically be resolved at compile time for known start & end types,
and the lookup of the correct body_details struct may add CPU cycles.

This PR:
- Adds a new file - _sv_inline.h_, into which are moved many of the definitions
and structures from _sv.c_. This seemed necessary because of the spread
of type definitions across existing header files. 
- Converts `newSV_type` into an inline function and adds to it the logic from
`sv_upgrade` necessary to upgrade a SVt_NULL.

Building on that, the commits in this PR:

- Modify existing calls to `newSV(sv)` followed by an `sv_upgrade(sv, type)` to
just use `newSV_type`, so that they also benefit.
- Replaces calls to `newSV(0)` with `newSV_type(SVt_NULL)`
- Add a new inline function, newSV_type_mortal, to address the absence of 
an efficient way to make a new non-SVt_NULL mortal SV.

With gcc version 10.2.1 on Debian Linux, the resulting perl binary was 25k
larger than blead. (The main commit accounts for almost all of this.)

I used the following trivial benchmark as a gauge of the performance
difference, finding the patched version to be about 30% faster:
`perl -e '$str="A"x64; for (0 .. 1_000_000) { @svs = split //, $str }'`

perf showed numbers in this region for blead:
```
          2,509.68 msec task-clock                #    1.000 CPUs utilized
                 5      context-switches          #    0.002 K/sec
               205      page-faults               #    0.082 K/sec
    10,344,797,368      cycles                    #    4.122 GHz                      (62.29%)
        23,884,608      stalled-cycles-frontend   #    0.23% frontend cycles idle     (62.45%)
     2,800,367,628      stalled-cycles-backend    #   27.07% backend cycles idle      (62.61%)
    32,567,445,991      instructions              #    3.15  insn per cycle
                                                  #    0.09  stalled cycles per insn  (62.70%)
     7,666,288,647      branches                  # 3054.684 M/sec                    (62.70%)
        11,063,728      branch-misses             #    0.14% of all branches          (62.57%)
    13,941,078,593      L1-dcache-loads           # 5554.916 M/sec                    (62.41%)
       149,071,315      L1-dcache-load-misses     #    1.07% of all L1-dcache accesses  (62.25%)
```
with sv_upgrade taking 25% of the run time:
```
  27.82%  perlblead  perlblead           [.] Perl_sv_upgrade
  13.75%  perlblead  perlblead           [.] Perl_sv_clear
  11.20%  perlblead  libc-2.31.so        [.] _int_free
  10.08%  perlblead  libc-2.31.so        [.] malloc
   6.30%  perlblead  libc-2.31.so        [.] _int_malloc
   6.01%  perlblead  perlblead           [.] Perl_newSVpvn_flags
   5.49%  perlblead  perlblead           [.] Perl_sv_setpvn_fresh.part.0
   4.35%  perlblead  perlblead           [.] Perl_safesysmalloc
   4.17%  perlblead  perlblead           [.] Perl_av_clear
   3.28%  perlblead  perlblead           [.] Perl_sv_free2
   1.97%  perlblead  perlblead           [.] Perl_pp_split
```
perf showed numbers in this region for patched:
```
          1,780.95 msec task-clock                #    1.000 CPUs utilized
                 9      context-switches          #    0.005 K/sec
               205      page-faults               #    0.115 K/sec
     7,361,942,324      cycles                    #    4.134 GHz                      (83.22%)
        15,476,039      stalled-cycles-frontend   #    0.21% frontend cycles idle     (83.38%)
     1,205,408,051      stalled-cycles-backend    #   16.37% backend cycles idle      (83.38%)
    27,395,234,336      instructions              #    3.72  insn per cycle
                                                  #    0.04  stalled cycles per insn  (83.38%)
     6,510,317,247      branches                  # 3655.522 M/sec                    (83.38%)
        11,044,449      branch-misses             #    0.17% of all branches          (83.26%)
```
and other functions coming to the fore:
```
  16.61%  perl     perl                [.] Perl_sv_clear
  14.34%  perl     libc-2.31.so        [.] malloc
  13.69%  perl     libc-2.31.so        [.] _int_free
  10.74%  perl     perl                [.] Perl_newSVpvn_flags
   8.48%  perl     libc-2.31.so        [.] _int_malloc
   7.20%  perl     perl                [.] Perl_sv_setpvn_fresh.part.0
   7.02%  perl     perl                [.] Perl_safesysmalloc
   5.39%  perl     perl                [.] Perl_sv_free2
   5.35%  perl     perl                [.] Perl_av_clear
   3.54%  perl     libc-2.31.so        [.] cfree@GLIBC_2.2.5
   2.56%  perl     perl                [.] Perl_pp_split
```
Note: This might be the best-case benchmark, as it is pretty much all about SV creation
and destruction, with little overhead from op dispatch or other functions.

This commit has left the big chunk of body commentary in _sv.c_ somewhat adrift of
e.g. `the bodies_by_type` table. I don't know how best to tidy that up. Hoping for some
feedback and suggestions!

### DIFF
--- a/sv_inline.h
+++ b/sv_inline.h
@@ -1,18 +1,26 @@
-/*    sv.h
+/*    sv_inline.h
  *
- *    Copyright (C) 1991, 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999, 2000,
- *    2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009 by Larry Wall and others
+ *    Copyright (C) 2022 by Larry Wall and others
  *
  *    You may distribute under the terms of either the GNU General Public
  *    License or the Artistic License, as specified in the README file.
  *
  */
 
+/* This file contains the newSV_type and newSV_type_mortal functions, as well as
+ * the various struct and macro definitions they require. In the main, these
+ * definitions were moved from sv.c, where many of them continue to also be used.
+ * (In Perl_more_bodies, Perl_sv_upgrade and Perl_sv_clear, for example.) Code
+ * comments associated with definitions and functions were also copied across
+ * verbatim.
+ *
+ * The rationale for having these as inline functions, rather than in sv.c, is
+ * that the target type is very often known at compile time, and therefore
+ * optimum code can be emitted by the compiler, rather than having all calls
+ * traverse the many branches of Perl_sv_upgrade at runtime.
+ */
 
-/* 2022 */
-/* BLAH BLAH BLAH */
-
-/* This came from perl.h*/
+/* This definition came from perl.h*/
 
 /* The old value was hard coded at 1008. (4096-16) seems to be a bit faster,
    at least on FreeBSD.  YMMV, so experiment.  */
@@ -56,7 +64,7 @@
     } STMT_END
 
 /* Perl_more_sv lives in sv.c, we don't want to inline it.
- * but the function declaration seems to be needed? */
+ * but the function declaration seems to be needed. */
 SV* Perl_more_sv(pTHX);
 
 /* new_SV(): return a new, empty SV head */
@@ -487,6 +495,35 @@ Perl_newSV_type(pTHX_ const svtype type)
                    (unsigned long)type);
     }
 
+    return sv;
+}
+
+/*
+=for apidoc newSV_type_mortal
+
+Creates a new mortal SV, of the type specified.  The reference count for the
+new SV is set to 1.
+
+This is equivalent to
+    SV* sv = sv_2mortal(newSV_type(<some type>))
+and
+    SV* sv = sv_newmortal();
+    sv_upgrade(sv, <some_type>)
+but should be more efficient than both of them. (Unless sv_2mortal is inlined
+at some point in the future.)
+
+=cut
+*/
+
+PERL_STATIC_INLINE SV *
+Perl_newSV_type_mortal(pTHX_ const svtype type)
+{
+    SV *sv = newSV_type(type);
+    SSize_t ix = ++PL_tmps_ix;
+    if (UNLIKELY(ix >= PL_tmps_max))
+        ix = Perl_tmps_grow_p(aTHX_ ix);
+    PL_tmps_stack[ix] = (sv);
+    SvTEMP_on(sv);
     return sv;
 }
 
--- a/sv_inline.h
+++ b/sv_inline.h
@@ -1,0 +1,495 @@
+/*    sv.h
+ *
+ *    Copyright (C) 1991, 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999, 2000,
+ *    2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009 by Larry Wall and others
+ *
+ *    You may distribute under the terms of either the GNU General Public
+ *    License or the Artistic License, as specified in the README file.
+ *
+ */
+
+
+/* 2022 */
+/* BLAH BLAH BLAH */
+
+/* This came from perl.h*/
+
+/* The old value was hard coded at 1008. (4096-16) seems to be a bit faster,
+   at least on FreeBSD.  YMMV, so experiment.  */
+#ifndef PERL_ARENA_SIZE
+#define PERL_ARENA_SIZE 4080
+#endif
+
+/* All other pre-existing definitions and functions that were moved into this
+ * file originally came from sv.c. */
+
+#ifdef PERL_POISON
+#  define SvARENA_CHAIN(sv)     ((sv)->sv_u.svu_rv)
+#  define SvARENA_CHAIN_SET(sv,val)     (sv)->sv_u.svu_rv = MUTABLE_SV((val))
+/* Whilst I'd love to do this, it seems that things like to check on
+   unreferenced scalars
+#  define POISON_SV_HEAD(sv)    PoisonNew(sv, 1, struct STRUCT_SV)
+*/
+#  define POISON_SV_HEAD(sv)    PoisonNew(&SvANY(sv), 1, void *), \
+                                PoisonNew(&SvREFCNT(sv), 1, U32)
+#else
+#  define SvARENA_CHAIN(sv)     SvANY(sv)
+#  define SvARENA_CHAIN_SET(sv,val)     SvANY(sv) = (void *)(val)
+#  define POISON_SV_HEAD(sv)
+#endif
+
+#ifdef PERL_MEM_LOG
+#  define MEM_LOG_NEW_SV(sv, file, line, func)  \
+            Perl_mem_log_new_sv(sv, file, line, func)
+#  define MEM_LOG_DEL_SV(sv, file, line, func)  \
+            Perl_mem_log_del_sv(sv, file, line, func)
+#else
+#  define MEM_LOG_NEW_SV(sv, file, line, func)  NOOP
+#  define MEM_LOG_DEL_SV(sv, file, line, func)  NOOP
+#endif
+
+#define uproot_SV(p) \
+    STMT_START {                                        \
+        (p) = PL_sv_root;                               \
+        PL_sv_root = MUTABLE_SV(SvARENA_CHAIN(p));              \
+        ++PL_sv_count;                                  \
+    } STMT_END
+
+/* Perl_more_sv lives in sv.c, we don't want to inline it.
+ * but the function declaration seems to be needed? */
+SV* Perl_more_sv(pTHX);
+
+/* new_SV(): return a new, empty SV head */
+
+#ifdef DEBUG_LEAKING_SCALARS
+/* provide a real function for a debugger to play with */
+STATIC SV*
+S_new_SV(pTHX_ const char *file, int line, const char *func)
+{
+    SV* sv;
+
+    if (PL_sv_root)
+        uproot_SV(sv);
+    else
+        sv = Perl_more_sv(aTHX);
+    SvANY(sv) = 0;
+    SvREFCNT(sv) = 1;
+    SvFLAGS(sv) = 0;
+    sv->sv_debug_optype = PL_op ? PL_op->op_type : 0;
+    sv->sv_debug_line = (U16) (PL_parser && PL_parser->copline != NOLINE
+                ? PL_parser->copline
+                :  PL_curcop
+                    ? CopLINE(PL_curcop)
+                    : 0
+            );
+    sv->sv_debug_inpad = 0;
+    sv->sv_debug_parent = NULL;
+    sv->sv_debug_file = PL_curcop ? savesharedpv(CopFILE(PL_curcop)): NULL;
+
+    sv->sv_debug_serial = PL_sv_serial++;
+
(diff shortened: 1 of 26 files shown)


richardleach:
Note: xenu looked at preventing inlining when the desired type is not know
at compile time - see patch below - but said that the difference in perl binary
size was only 24 bytes.
```
> git diff
diff --git a/embed.h b/embed.h
index 94a51192ed..eeb9ca7e6a 100644
--- a/embed.h
+++ b/embed.h
@@ -385,7 +385,7 @@
 #define newSV(a)               Perl_newSV(aTHX_ a)
 #define newSVOP(a,b,c)         Perl_newSVOP(aTHX_ a,b,c)
 #define newSVREF(a)            Perl_newSVREF(aTHX_ a)
-#define newSV_type(a)          Perl_newSV_type(aTHX_ a)
+#define newSV_type(a)          (__builtin_constant_p(a) ? Perl_newSV_type(aTHX_ a) : Perl_newSV_type_noinline(aTHX_ a))
 #define newSVhek(a)            Perl_newSVhek(aTHX_ a)
 #define newSViv(a)             Perl_newSViv(aTHX_ a)
 #define newSVnv(a)             Perl_newSVnv(aTHX_ a)
diff --git a/proto.h b/proto.h
index 7c7dd528ba..19ff6a067d 100644
--- a/proto.h
+++ b/proto.h
@@ -7062,6 +7062,9 @@ PERL_CALLCONV int Perl_magic_regdatum_set(pTHX_ SV* sv, MAGIC* mg);
 #define PERL_ARGS_ASSERT_MAGIC_REGDATUM_SET    \
        assert(sv); assert(mg)
 #endif
+
+SV * Perl_newSV_type_noinline(pTHX_ const svtype type);
+
 #ifdef PERL_CORE
 #  include "pp_proto.h"
 #endif
diff --git a/sv.c b/sv.c
index 742fb9c332..f918cbbe33 100644
--- a/sv.c
+++ b/sv.c
@@ -17094,6 +17094,12 @@ Perl_report_uninit(pTHX_ const SV *uninit_sv)
     GCC_DIAG_RESTORE_STMT;
 }
 
+SV *
+Perl_newSV_type_noinline(pTHX_ const svtype type)
+{
+    return Perl_newSV_type(aTHX_ type);
+}
+
 /*
  * ex: set ts=8 sts=4 sw=4 et:
  */
```

demerphq:
On Sun, 13 Feb 2022 at 23:54, Richard Leach ***@***.***>
wrote:

> Note: This is an alternative to #19381
> <https://github.com/Perl/perl5/pull/19381>, following feedback from @xenu
> <https://github.com/xenu>.
>
> When a new SV is created and upgraded to a type known at compile time,
> using the general-purpose upgrade function (sv_upgrade) is clunky.
> Specifically, while uprooting a SV head is lightweight (assuming there are
> unused SVs), sv_upgrade is too big to be inlined, contains many branches
> that can logically be resolved at compile time for known start & end types,
> and the lookup of the correct body_details struct may add CPU cycles.
>
> This PR:
>
>    - Adds a new file - *sv_inline.h*, into which are moved many of the
>    definitions
>    and structures from *sv.c*. This seemed necessary because of the spread
>    of type definitions across existing header files.
>    - Converts newSV_type into an inline function and adds to it the logic
>    from
>    sv_upgrade necessary to upgrade a SVt_NULL.
>
> Building on that, the commits in this PR:
>
>    - Modify existing calls to newSV(sv) followed by an sv_upgrade(sv,
>    type) to
>    just use newSV_type, so that they also benefit.
>    - Replaces calls to newSV(0) with newSV_type(SVt_NULL)
>    - Add a new inline function, newSV_type_mortal, to address the absence
>    of
>    an efficient way to make a new non-SVt_NULL mortal SV.
>
> With gcc version 10.2.1 on Debian Linux, the resulting perl binary was 25k
> larger than blead. (The main commit accounts for almost all of this.)
>
> I used the following trivial benchmark as a gauge of the performance
> difference, finding the patched version to be about 30% faster:
> perl -e '$str="A"x64; for (0 .. 1_000_000) { @svs = split //, $str }'
>
dumbbench concurs:

sv_upgrade_fresh1:~/git_tree/perl$ dumbbench -- ./perl -Ilib -e
'$str="A"x64; for (0 .. 1_000_000) { @svs = split //, $str }'
cmd: Ran 21 iterations (1 outliers).
cmd: Rounded run time per iteration: 3.0000e+00 +/- 5.6e-03 (0.2%)

blead:~/git_tree/perl2$ dumbbench -- ./perl -Ilib -e '$str="A"x64; for (0
.. 1_000_000) { @svs = split //, $str }'
cmd: Ran 24 iterations (4 outliers).
cmd: Rounded run time per iteration: 4.0168e+00 +/- 2.9e-03 (0.1%)

Nice.

cheers,
Yves

-- 

sv_inline.h
```diff
@@ -0,0 +1,527 @@
...
+   unreferenced scalars
+#  define POISON_SV_HEAD(sv)    PoisonNew(sv, 1, struct STRUCT_SV)
+*/
+#  define POISON_SV_HEAD(sv)    PoisonNew(&SvANY(sv), 1, void *), \
+                                PoisonNew(&SvREFCNT(sv), 1, U32)
+#else
+#  define SvARENA_CHAIN(sv)     SvANY(sv)
+#  define SvARENA_CHAIN_SET(sv,val)     SvANY(sv) = (void *)(val)
+#  define POISON_SV_HEAD(sv)
+#endif
```
xenu, line 47: Unless I'm missing something, none of the above defines had to be moved here.

sv_inline.h
```diff
@@ -0,0 +1,527 @@
...
+ *    2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009 by Larry Wall and others
+ *
+ *    You may distribute under the terms of either the GNU General Public
+ *    License or the Artistic License, as specified in the README file.
+ *
+ */
+
+
+/* 2022 */
+/* BLAH BLAH BLAH */
```
xenu: You might want to remove this :)

sv_inline.h
richardleach: `SvARENA_CHAIN` is used in the `uproot_SV` macro below. 

I moved the whole block of `define`s that seemed to belong together, rather than breaking them up.

sv_inline.h
richardleach: Oh yes, that was my placeholder 'cos I wasn't sure what copyright message and descriptive text to include.

I'll change it to the following, unless there are any other recommendations for what else to mention.

```
/*    sv_inline.h
 *
 *    Copyright (C) 2022 by Larry Wall and others
 *
 *    You may distribute under the terms of either the GNU General Public
 *    License or the Artistic License, as specified in the README file.
 *
 */

/* This file contains the newSV_type and newSV_type_mortal functions, as well as
 * the various struct and macro definitions they require. In the main, these
 * definitions were moved from sv.c, where many of them continue to also be used.
 */
```

richardleach: I've pushed that plus an additional chunk of notes & rationale.

xenu:
Two small issues remaining:

- Typo in the commit message: "unprooting". 
- Leftover commented-out code in sv.h (I think it should be uncommented?)

Otherwise LGTM.

richardleach:
> Two small issues remaining:
> 
> * Typo in the commit message: "unprooting".
> * Leftover commented-out code in sv.h (I think it should be uncommented?)
> 
> Otherwise LGTM.

Thanks, I've addressed those two points. (Odd how the waas-commented-out code didn't have an apparent effect, but I can always look at why separately.)

bulk88:
Related to https://github.com/Perl/perl5/pull/22667

I believe this commit needs to be reverted at this point.

I compiled blead perl with GCC 8.3.0 i686 mode. It uses -Os optimization flag.

18 unique copies of struct bodies_by_details inside perl541.dll. The byte values inside each struct field in the array bodies_by_details are NOT considered as hoistable or inlineable  by GCC even though they are declared as ```static const```.

In every .o, unless all callers of static inline Perl_newSV_type(), pass the same sv_type value,  Perl_newSV_type() will not inline.

There are 16 copies of full sized Perl_newSV_type(). Full sized mean that the switch that handles all 17 SV types is there.

In 17 locations total, inside perl541.dll Perl_newSV_type() was totally inlined into its caller, or a not-inlined "one SV type" fast efficient version of Perl_newSV_type() was created.

Unless reproducible benchmarks are demonstrated, or disassembly/assembler deltas can be shown by someone, that justify this commit, or it is #ifdef-ed to exactly to one OS and one verified correct behavior CC with a hard coded version/build number of that CC, this needs to be reverted.

It is not a MSVC only problem problem. Its an every CC problem including GCC. Therefore I recommend reverting it.

bulk88:
Since GCC does not believe ```static const struct body_details bodies_by_type[]``` is RO const.

```
        if(type_details->arena) {
            /* This points to the start of the allocated area.  */
            new_body = S_new_body(aTHX_ type);
            Zero(new_body, type_details->body_size, char);
            new_body = ((char *)new_body) - type_details->offset;
        } else
```
Even in fast efficient "one SV type" versions of ```Perl_newSV_type()```, a call to libc ```memset``` is done, even for SVPV bodies which are only 16 bytes/0x10 bytes big, since GCC doesn't know if struct ```bodies_by_type``` will ask to zero fill 100 MB or 10 bytes. GCC always inlines small fixed const length arg ```memset```  null fill calls, into efficient CPU ```mov``` instructions. For this commit, nope, ```memset``` libc calls are emitted.

There is too much to fix, to get the original authors intent of this branch, after "fixing" this branch, there will be no source code lines left from this branch basically. This merge added nothing of value or no code that can be kept.

C/C++ language is not NodeJS, Java, or CLR/.NET. And C compilers, all of them, won't optimize as if the perl interpreter is a just another javascript file.

richardleach:
> Related to #22667
> 
> I believe this commit needs to be reverted at this point.
> 
> I compiled blead perl with GCC 8.3.0 i686 mode. It uses -Os optimization flag.

Who is building perl with gcc and -Os and what's their use case?

As noted earlier in the PR, this gives a good performance boost to people using the default optimization flags (-O2, no -Os). This, I believe, includes many of the Linux & BSD distro builds. If the majority of non-Windows users are getting the benefit of `newSV_type`, and we don't know of a specific group of users that is suffering, then I don't see the case for reverting.

I certainly DO agree that there's room for improvement around handling of `bodies_by_details`. I was not thrilled back when working on this that it doesn't get compiled away for the "one static SV type" calls. (e.g.` newSV_type(SVt_PV)`) I wondered about moving away from the table model to macros or inline functions, but (a) was wary of making too many structural changes at once (b) ran out of time. I'm happy to pick up on any suggestions when time allows though.

PR #19415, opened 2022-02-14
Yves/revert reverted caller fixes

demerphq:
Revert: commit c0d05305c156c83e4f9f3a207451b3175fbb7f24
Merge: c6b2e294d8 79f75eaa71
Author: David Mitchell <davem@iabyn.com>
Date:   Mon Apr 27 22:04:23 2020 +0100

    [MERGE] Revert BEGIN { caller() } fixups
    
    These commits were intended to fix a problem with stack backtraces
    reporting wrong file and line numbers in nested use's.
    
    A side-effect of the commits was to fix the package name returned by
    caller() too; but quite a few distributions were relying on the old
    behaviour.
    
    So for now, revert to the old behaviour and re-address after 5.32.0 is
    released.
    
    The reverted commits are:
    
    v5.31.6-141-gf2f32cd638 avoid identical stack traces
    v5.31.9-122-gee428a211d docs: clarify effect of $^H, %^H, ${^WARNING_BITS}
    v5.31.9-162-gad89278aa2 fixup to "avoid identical stack traces" - try 2

### DIFF
--- a/op.c
+++ b/op.c
@@ -12043,10 +12043,31 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
             (void)CvGV(cv);
             if (floor) LEAVE_SCOPE(floor);
             ENTER;
+
+            SAVEVPTR(PL_curcop);
+            if (PL_curcop == &PL_compiling) {
+                /* Avoid pushing the "global" &PL_compiling onto the
+                 * context stack. For example, a stack trace inside
+                 * nested use's would show all calls coming from whoever
+                 * most recently updated PL_compiling.cop_file and
+                 * cop_line.  So instead, temporarily set PL_curcop to a
+                 * private copy of &PL_compiling. PL_curcop will soon be
+                 * set to point back to &PL_compiling anyway but only
+                 * after the temp value has been pushed onto the context
+                 * stack as blk_oldcop.
+                 * This is slightly hacky, but necessary. Note also
+                 * that in the brief window before PL_curcop is set back
+                 * to PL_compiling, IN_PERL_COMPILETIME/IN_PERL_RUNTIME
+                 * will give the wrong answer.
+                 */
+                PL_curcop = (COP*)newSTATEOP(PL_compiling.op_flags, NULL, NULL);
+                CopLINE_set(PL_curcop, CopLINE(&PL_compiling));
+                SAVEFREEOP(PL_curcop);
+            }
+
             PUSHSTACKi(PERLSI_REQUIRE);
             SAVECOPFILE(&PL_compiling);
             SAVECOPLINE(&PL_compiling);
-            SAVEVPTR(PL_curcop);
 
             DEBUG_x( dump_sub(gv) );
             Perl_av_create_and_push(aTHX_ &PL_beginav, MUTABLE_SV(cv));
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -1872,6 +1872,10 @@ It has the same scoping as the C<$^H> and C<%^H> variables.  The exact
 values are considered internal to the L<warnings> pragma and may change
 between versions of Perl.
 
+Each time a statement completes being compiled, the current value of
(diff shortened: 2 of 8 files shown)


haarg:
This was previously discussed on [P5P](https://www.nntp.perl.org/group/perl.perl5.porters/2022/01/msg262374.html) and #17663.

I would support getting this merged as soon as possible. The modules that are "broken" by this are either already broken, or trivial to fix, or both. One of the modules broken was mine, and was fixed during the time the patches were live in blead.

The arguments against this that seemed potentially valid were that the correct caller location of a `BEGIN` block should be the `require` that loaded the file. I don't think that behavior is at all useful. Additionally, that isn't what actually happens without this patch. Instead, the caller of a `BEGIN` block reflects data that has been partially overwritten by various stages of compilation. This results in mismatches between package and file/line data, or outer stack frames reportedly being called by inner frames.

op.c
```diff
@@ -12041,12 +12041,33 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
             const I32 oldscope = PL_scopestack_ix;
             dSP;
             (void)CvGV(cv);
-            if (floor) LEAVE_SCOPE(floor);
-            ENTER;
-            PUSHSTACKi(PERLSI_REQUIRE);
-            SAVECOPFILE(&PL_compiling);
-            SAVECOPLINE(&PL_compiling);
+	    if (floor) LEAVE_SCOPE(floor);
```
leonerd: Something weird is going on indentation-wise here. It might be due to mixed tabs-vs-spaces. Make sure the new lines have only spaces, no tabs.

op.c
```diff
@@ -12041,12 +12041,33 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
...
+                 * to PL_compiling, IN_PERL_COMPILETIME/IN_PERL_RUNTIME
+                 * will give the wrong answer.
+                 */
+                PL_curcop = (COP*)newSTATEOP(PL_compiling.op_flags, NULL, NULL);
+                CopLINE_set(PL_curcop, CopLINE(&PL_compiling));
+                SAVEFREEOP(PL_curcop);
+            }
+
+            PUSHSTACKi(PERLSI_REQUIRE);
+	    SAVECOPFILE(&PL_compiling);
```
leonerd: Here also, looks to be mixed tabs/spaces

pod/perlvar.pod
```diff
@@ -1872,6 +1872,10 @@ It has the same scoping as the C<$^H> and C<%^H> variables.  The exact
 values are considered internal to the L<warnings> pragma and may change
 between versions of Perl.
 
+Each time a statement completes being compiled, the current value of
+C<${^WARNING_BITS}> is stored with that statement, and can later be
+retrieved via C<(caller($level))[9]>.
+
```
leonerd, line 1878: Are these three documentation changes (here + the other two lower down) intended to be being added? They don't appear to relate to the current change. Perhaps it's some unrelated collateral from the double-revert?

op.c
demerphq: Nod. Confirmed and fixed.

op.c
demerphq: Confirmed and fixed.

pod/perlvar.pod
demerphq: Yes they are related. See thread starting at:

   https://www.nntp.perl.org/group/perl.perl5.porters/2020/02/msg257128.html

Sayeth Dave Mitchell in 

   https://www.nntp.perl.org/group/perl.perl5.porters/2020/03/msg257214.html
 
The change in behaviour was a side-effect of my fix for recording caller
data correctly, but I think the change makes behaviour actually match the
docs (although they are a bit vague). So I think the bug-fix is correct.

leonerd (approved):
LGTM

khwilliamson:
I think this should go in to 5.36

Issue #19416, opened 2022-02-14
Accidentally creating UNIVERSAL::isa breaks WHATEVER->import and use CLASS


demerphq:
**Description**
Taking a reference to UNIVERSAL::import creates a stub function, which if executed will die. It also breaks the expectation that WHATEVER->import() should not die.

**Steps to Reproduce**

perl  -le'my $import_sub= \&UNIVERSAL::import; Foo->import()'
Undefined subroutine &UNIVERSAL::import called at -e line 1.

perl  -le'BEGIN{ my $import_sub= \&UNIVERSAL::import;} use File::Spec;'
Undefined subroutine &UNIVERSAL::import called at -e line 1.
BEGIN failed--compilation aborted at -e line 1.

**Expected behavior**
No error.

perl  -le'Foo->import()'
perl  -le'use File::Spec qw();'

**Perl configuration**

```
Summary of my perl5 (revision 5 version 24 subversion 1) configuration:
   
  Platform:
    osname=linux, osvers=3.19.0-32-generic, archname=x86_64-linux
    uname='linux styx2 3.19.0-32-generic #37~14.04.1-ubuntu smp thu oct 22 09:41:40 utc 2015 x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dcc=ccache gcc -Dld=gcc -Dprefix=/home/yorton/perl5/perlbrew/perls/perl-5.24.1 -Aeval:scriptdir=/home/yorton/perl5/perlbrew/perls/perl-5.24.1/bin'
    hint=recommended, useposix=true, d_sigaction=define
    useithreads=undef, usemultiplicity=undef
    use64bitint=define, use64bitall=define, uselongdouble=undef
    usemymalloc=n, bincompat5005=undef
  Compiler:
    cc='ccache gcc', ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64',
    optimize='-O2',
    cppflags='-fwrapv -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include'
    ccversion='', gccversion='4.8.4', gccosandvers=''
    intsize=4, longsize=8, ptrsize=8, doublesize=8, byteorder=12345678, doublekind=3
    d_longlong=define, longlongsize=8, d_longdbl=define, longdblsize=16, longdblkind=3
[...]

Issue #19418, opened 2022-02-14
On case insensitive file systems using a module with a case inconsistency which is supposed to export a function does not error.


demerphq:
**Description**
On a case insensitive file system a case mismatch for a use statement that should export will load the module and not throw an error, but it will NOT actually export the function. See also Issue #19417, and Issue #19410

**Steps to Reproduce**
On a case insensitive file system:
perl -le'use LIST::UTIL qw(sum);'

**Expected behavior**
Some kind of exception, preferably that the module name is incorrectly spelled, or at least an exception that the function cannot be exported. 

**Perl configuration**

```
No perl -V available as I do not have a case insensitive file system handy, but the bug is present in 5.35
```

Issue #19417, opened 2022-02-14
Importing a function from a module which does not export does not throw an error


demerphq:
**Description**
Using a module which does not export, such as a class, which does not export does not throw an exception.

**Steps to Reproduce**
perl  -le'use File::Spec qw(catfile);'

**Expected behavior**
Expecting some kind of error which shows that the request function or symbol has not actually been exported.

**Perl configuration**
```
perl -V
Summary of my perl5 (revision 5 version 24 subversion 1) configuration:
   
  Platform:
    osname=linux, osvers=3.19.0-32-generic, archname=x86_64-linux
    uname='linux styx2 3.19.0-32-generic #37~14.04.1-ubuntu smp thu oct 22 09:41:40 utc 2015 x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dcc=ccache gcc -Dld=gcc -Dprefix=/home/yorton/perl5/perlbrew/perls/perl-5.24.1 -Aeval:scriptdir=/home/yorton/perl5/perlbrew/perls/perl-5.24.1/bin'
    hint=recommended, useposix=true, d_sigaction=define
    useithreads=undef, usemultiplicity=undef
    use64bitint=define, use64bitall=define, uselongdouble=undef
    usemymalloc=n, bincompat5005=undef
  Compiler:
    cc='ccache gcc', ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64',
    optimize='-O2',
    cppflags='-fwrapv -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include'
    ccversion='', gccversion='4.8.4', gccosandvers=''
    intsize=4, longsize=8, ptrsize=8, doublesize=8, byteorder=12345678, doublekind=3
    d_longlong=define, longlongsize=8, d_longdbl=define, longdblsize=16, longdblkind=3
    ivtype='long', ivsize=8, nvtype='double', nvsize=8, Off_t='off_t', lseeksize=8
    alignbytes=8, prototype=define
  Linker and Libraries:
    ld='gcc', ldflags =' -fstack-protector -L/usr/local/lib'
[...]

Issue #19410, opened 2022-02-12
Nonexistent::Module->import() gives no error


jimav:
`Nonexistent::Module->import()` **silently succeeds**.  Expecting an error like
 
_Can't locate object method "import" via package "Nonexistent::Module" (perhaps you forgot to load "Nonexistent::Module"?)_

The docs say *import* is an ordinary method and not builtin, but that doesn't seem to be entirely true.   I think this is a bug (calling any method on a non-existent object or class should result in an error); 

**if it can not be fixed, then I suggest updating the perlfunc entry for 'import' to explain what is happening.** 

```
#!/usr/bin/perl
use strict; use warnings;
use Data::Dumper;

$Data::Dumper::Terse = 1;

print "BEFORE:", Dumper(\%Misspelled::);
print "AGAIN :", Dumper(\%Misspelled::);

Misspelled->import();            # NO error

print "AFTER :", Dumper(\%Misspelled::);

Misspelled->othermethod(); # error as expected
```
Output:
```
BEFORE:{}
AGAIN :{}
AFTER :{
  'import' => *Misspelled::import
}
Can't locate object method "othermethod" via package "Misspelled" at /tmp/test.pl line 14.
```

Grinnz:
It is unfortunate, but this is because of:

1. there is no way to conclusively decide if a module has been "loaded", because any perl code can define contents of any package;
2. "use" calls import on whatever you load, so it must succeed even if no such method exists

This is also the source of silent (or not so silent) failures in a circular "use", where the import method is called on a package before its code has been fully loaded.

haarg:
The behavior of `import` is described better in the docs for `use`. Changing this behavior won't be possible, but the docs under import could definitely be improved.

xenu:
I think calling non-existent import **with arguments** is basically always a mistake. Maybe that case should warn? It would also help on case-insensitive file systems:

```
use LIST::UTIL 'any'; # quietly doesn't import anything
```

jimav:
> 2\. "use" calls import on whatever you load, so it must succeed even if no such method exists

Must it?  Could something like
 `&{ $modulename."::import" }(...) if defined(&{ $modulename."::import" })`work?

Edit: Oh never mind -- that won't find 'import' if inherited from another class.   But could "use" reliably determine if 'import' is defined by checking for @ ISA (recursively)?

demerphq:
On Mon, 14 Feb 2022 at 07:40, Jim Avera ***@***.***> wrote:

[...]

PR #19419, opened 2022-02-14
Fix assorted bugs related to not having a UNIVERSAL::import

demerphq:
Since perl 5.0 the methods "import" and "unimport" have been
special cased in gv.c (unimport was removed for a while) to
not produce errors if they are called. This is partly
because

    use Foo;

is defined to be

    BEGIN {
        require Foo;
        Foo->import();
    }

which would blow up if there is no import function defined in
Foo, for instance if it were defining a class and not a package
which exports modules.

This special case can be broken by simple code like

    \&UNIVERSAL::isa

which will create a stub function which then blows up when it is
used. Notably the module "autouse" which is shipped with perl will
trigger this behavior.

A related issue is that if you ask for a function to be exported
from a module that does not have support for exporting there is no
error, eg:

    use File::Spec qw(catfile);

will silently succeed without exporting a catfile function. This is
exacerbated on case insensitive file systems when the module name
is case-mismatched, the use succeeds but the export does not, leading
to confusion, eg:

    use LIst::Util qw(sum);

will load List::Util but will not export the sum function.

This patch defines UNIVERSAL::import() and UNIVERSAL::unimport()
functions. This prevents the "reference to \&UNIVERSAL::import" bug.
The function is defined to be a no-op unless arguments are passed into
the functions, in which case a fatal exception is thrown indicating
that there is likely a problem. The error is modelled after the
error produced by calling a non-existent method or function:

    ./perl -Ilib -le'BEGIN{ my $import_sub= \&UNIVERSAL::import;}
        use File::Spec qw(catfile);'
    Attempt to call UNIVERSAL::import() with arguments via package File::Spec
        (Perhaps you forgot to load "File::Spec"?) at -e line 1.
    BEGIN failed--compilation aborted at -e line 1.

This fixes Issue #19416, Issue #19417, Issue #19418. See also Issue #19410 for
discussion, however this patch does not fix that case (it may not be
fixable.)

Note that the class '_charnames' is special-cased as we have other logic 
which handles when it has not been loaded properly.

### DIFF
--- a/universal.c
+++ b/universal.c
@@ -454,6 +454,29 @@ XS(XS_UNIVERSAL_isa)
     }
 }
 
+XS(XS_UNIVERSAL_import_unimport); /* prototype to pass -Wmissing-prototypes */
+XS(XS_UNIVERSAL_import_unimport)
+{
+    dXSARGS;
+    dXSI32;
+
+    if (items > 1) {
+        char *class_pv= SvPV_nolen(ST(0));
+        if (strEQ(class_pv,"UNIVERSAL"))
+            Perl_croak(aTHX_ "UNIVERSAL does not export anything");
+        /* _charnames is special - ignore it for now as the code that
+         * depends on it has its own "no import" logic that produces better
+         * warnings than this does. */
+        if (strNE(class_pv,"_charnames"))
+            Perl_croak(aTHX_
+                "Attempt to call undefined %s method with arguments via package "
+                "%" SVf_QUOTEDPREFIX " (Perhaps you forgot to load the package?)",
+                ix ? "unimport" : "import", SVfARG(ST(0)));
+    }
+    XSRETURN_EMPTY;
+}
+
+
 XS(XS_UNIVERSAL_can); /* prototype to pass -Wmissing-prototypes */
 XS(XS_UNIVERSAL_can)
 {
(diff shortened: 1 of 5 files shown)


xenu:
I like the idea, but I think the error message is unclear. I think there's no need to mention `UNIVERSAL::import`. Maybe something like "undefined import()" would be better?

demerphq:
@xenu thanks for the suggestion, the latest version of the patch incorporates your feedback.

universal.c
```diff
@@ -433,6 +433,29 @@ XS(XS_UNIVERSAL_isa)
...
+        if (strEQ(class_pv,"UNIVERSAL"))
+            Perl_croak(aTHX_ "UNIVERSAL does not export anything");
+        /* _charnames is special - ignore it for now as the code that
+         * depends on it has its own "no import" logic that produces better
+         * warnings than this does. */
+        if (strNE(class_pv,"_charnames"))
+            Perl_croak(aTHX_
+                "Attempt to call undefined %s with arguments via package %s"
+                " (Perhaps you forgot to load \"%s\"?)",
+                ix ? "unimport" : "import", class_pv, class_pv);
```
ilmari: This will not work correctly for non-ASCII package names. Use `SVf` and `SVfARG` as described in [`perlguts`](https://perldoc.perl.org/perlguts#Formatted-Printing-of-SVs).

demerphq: Fixed. Thanks. I wonder, should it actually use something that would make unprintables visible?

demerphq: FWIW, I have pushed a set of follow up patches adding a way to deal with unprintables.

leonerd (changes requested):
Looks good so far, though there's no `perldiag.pod` for the new message yet.
Also the tests are upset about the wording of the new message - that probably just needs updating.

demerphq:
On Mon, 1 Aug 2022, 18:43 Paul Evans, ***@***.***> wrote:

>
> Looks good so far, though there's no perldiag.pod for the new message yet.
> Also the tests are upset about the wording of the new message - that
> probably just needs updating
>

Fixes and improvements coming!

Thanks,
Yves

>

tonycoz (approved):

haarg (approved):

PR #19420, opened 2022-02-15
Add feature to get extra paired string delimiters

khwilliamson:
This is based on http://nntp.perl.org/group/perl.perl5.porters/262686

No tests are included as of yet.


khwilliamson:
Easiest to just close it and create a new PR #19421

PR #19423, opened 2022-02-15
revert to windows-2019 in GitHub Actions (for now)

xenu:
It seems the changes in windows-latest (windows-2022) image broke our
builds.

Example failure: https://github.com/Perl/perl5/runs/5201005335?check_suite_focus=true

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -326,7 +326,7 @@ jobs:
 
   windows-msvc142:
     name: "Windows msvc142"
-    runs-on: windows-latest
+    runs-on: windows-2019
     timeout-minutes: 120
     needs: sanity_check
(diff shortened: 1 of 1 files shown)


sisyphus:
In my own build environment, current blead is also failing (crashing) in re/regexp_unicode_prop_thr.t.
This is a new development since the devel release of 5.35.8.

However, that's the only commonality I see in the log you linked to. I experience no problems with the other test scripts that failed.

I'll open a separate ticket about it if this failure still persists in the 5.35.9 devel release that's due in a few days time.

xenu:
I actually meant to link to Visual C++ failure which is definitely caused by the image changes (nmake.exe couldn't be found). Those Mingw failures are indeed suspicious, but I'm procrastinating investigating them :P

PR #19424, opened 2022-02-15
Deëxperiment signatures

leonerd:
The long-awaited PR

 * No longer prints "signatures are experimental" warning
 * Removes now-redundant `no warnings "experimental::signatures"` from several support files and unit tests
 * Updates much documentation on the subject of subroutine signatures
 * Adds `signatures` to the `:5.35` feature bundle


pod/perldelta.pod
```diff
@@ -27,6 +27,25 @@ here, but most should go in the L</Performance Enhancements> section.
...
+is included in the C<:5.36> feature bundle, enabled automatically by
+C<use v5.36>:
+
+    use v5.36;
+
+    sub add ($x, $y) {
+      return $x + $y;
+    }
+
+No longer prints a warning.
```
Grinnz: slightly awkward sentence fragment - perhaps this could just be included in the previous paragraph: "is now no longer considered experimental and no longer prints a warning."

leonerd: Reworded and force-pushed.

pod/perldelta.pod
```diff
@@ -27,6 +27,23 @@ here, but most should go in the L</Performance Enhancements> section.
...
+    sub add ($x, $y) {
+      return $x + $y;
+    }
+
+Despite this, certain elements of signatured subroutines remain
+experimental; see below.
+
 =head2 @_ is now experimental within signatured subs
 
 Even though subroutine signatures currently remain experimental, use of the
```
johannessen: I think the description of the section about `@_` in signatured subs also needs to be adjusted.

leonerd: Ahyes. Further rewordings force-pushed.

pod/perlexperiment.pod
```diff
@@ -162,6 +152,15 @@ Using this feature triggers warnings in the category C<experimental::try>.
 The ticket for this experiment is
 L<[perl #18760]|https://github.com/Perl/perl5/issues/18760>
 
+=item Use of C<@_> with subroutine signatures
```
ilmari: Shouldn't this be "within signatured subroutines"?

pod/perlsub.pod
```diff
@@ -104,7 +104,7 @@ like a C<foreach> or a C<while>, the returned value is unspecified.  The
 empty sub returns the empty list.
 X<subroutine, return value> X<return value> X<return>
 
-Aside from an experimental facility (see L</Signatures> below),
+Aside from the "signatures" facility (see L</Signatures> below),
```
ilmari: Now that signatures are no longer experimental, would it not be better to write this in a positive sense, and move the `X<formal parameter>` and  `X<parameter, formal>` tags to the Signatures section?

leonerd: Good idea. Further rewordings pushed.
(I'll definitely have to remember to squash these all up before merge)

pod/perlexperiment.pod
leonerd: Ahyes. Updated, squashed down a bit, and force-pushed. Again.

leonerd:
@tonycoz @khwilliamson - Either of you two regulars want to cast a reviewing eye over this?

lib/feature.pm
```diff
@@ -291,12 +291,6 @@ regardless of what feature declarations are in scope.
 
 =head2 The 'signatures' feature
 
-B<WARNING>: This feature is still experimental and the implementation may
-change or be removed in future versions of Perl.  For this reason, Perl will
-warn when you use the feature, unless you have explicitly disabled the warning:
-
-    no warnings "experimental::signatures";
-
 This enables unpacking of subroutine arguments into lexical variables
```
Ovid: I know you didn't write this, but this is still "programmer speak." I'd write:


```
This assigns subroutine arguments to lexical variables in the subroutine signature.
For example, for this subroutine:

    sub foo ($left, $right) {
        return $left + $right;
    }

Calling C<foo(3, 7);> will assign C<3> and C<7> to C<$left> and C<$right>, respectively,
in the subroutine.
```

pod/perlexperiment.pod
```diff
@@ -162,6 +152,15 @@ Using this feature triggers warnings in the category C<experimental::try>.
...
 L<[perl #18760]|https://github.com/Perl/perl5/issues/18760>
 
+=item Use of C<@_> within subroutine signatures
+
+Introduced in Perl 5.36.0 as part of a reduction in the scope of experimental
+subroutine signatures.
+
+Using the default arguments array (C<@_>) within a subroutine that uses
+signatures will emit a warning in the category
+C<experimental::args_array_with_signatures>.
```
Ovid: I'd duplicate the information mentioned previously in case someone sees this and doesn't see the perldelta mention.

```
Using the default arguments array (C<@_>) within a subroutine that uses
signatures will emit a warning in the category
C<experimental::args_array_with_signatures>.

This includes C<@_>, C<$_[INDEX]>, or using C<shift>, C<unshift>, C<pop>,
or C<push> without a list of arguments.
```

pod/perlsub.pod
```diff
@@ -320,16 +318,14 @@ explicitly.  See L<perlmod/"BEGIN, UNITCHECK, CHECK, INIT and END">
...
 
-Perl has an experimental facility to allow a subroutine's formal
-parameters to be introduced by special syntax, separate from the
-procedural code of the subroutine body.  The formal parameter list
-is known as a I<signature>.  The facility must be enabled first by a
-pragmatic declaration, C<use feature 'signatures'>, and it will produce
-a warning unless the "experimental::signatures" warnings category is
-disabled.
+Perl has a facility to allow a subroutine's formal parameters to be
+introduced by special syntax, separate from the procedural code of the
```
Ovid: `s/introduced/declared/`.

pod/perlsub.pod
```diff
@@ -320,16 +318,14 @@ explicitly.  See L<perlmod/"BEGIN, UNITCHECK, CHECK, INIT and END">
...
+subroutine body.  The formal parameter list is known as a I<signature>.
+The facility must be enabled first by a pragmatic declaration, 
+C<use feature 'signatures'>. This is automatically enabled by a
+C<use v5.36> (or higher) declaration in the current scope.
```
Ovid: Put the important information first.

```
Signatures are automatically available from v5.36 onwards. From version v5.20 to
v5.34, you must explicitly enable them with C<use feature 'signatures'>. This only
enables them for the current scope.
```

regen/feature.pl
```diff
@@ -702,12 +702,6 @@ =head2 The 'postderef' and 'postderef_qq' features
 
 =head2 The 'signatures' feature
 
-B<WARNING>: This feature is still experimental and the implementation may
-change or be removed in future versions of Perl.  For this reason, Perl will
-warn when you use the feature, unless you have explicitly disabled the warning:
-
-    no warnings "experimental::signatures";
-
 This enables unpacking of subroutine arguments into lexical variables
```
Ovid: As above.

```
This assigns subroutine arguments to lexical variables in the subroutine signature.
For example, for this subroutine:

    sub foo ($left, $right) {
        return $left + $right;
    }

Calling C<foo(3, 7);> will assign C<3> and C<7> to C<$left> and C<$right>, respectively,
in the subroutine.
```

Ovid (commented):
My comments are mostly minor nits to improve clarity (I tried to be conservative). Obviously none of this is critical.

lib/feature.pm
leonerd: I'll give it an adjustment.

pod/perlsub.pod
leonerd: It's not so much that they're *automatically* enabled on 5.36; it's that a `use v5.36` will turn them on. I'll see if I can adjust it.

pod/perldelta.pod
```diff
@@ -27,21 +27,37 @@ here, but most should go in the L</Performance Enhancements> section.
...
+warning category is not sufficient to dismiss this.  The new warning is
+emitted with the category name C<experimental::args_array_with_signatures>.
 
 Any subroutine that has a signature and tries to make use of the defaults
 argument array or an element thereof (C<@_> or C<$_[INDEX]>), either
 explicitly or implicitly (such as C<shift> or C<pop> with no argument) will
 provoke a warning at compile-time:
 
-    use experimental 'signatures';
+    use v5.36;
```
duncand, line 60: Was this removed line a typo instead of saying `use feature 'signatures'`?

duncand (approved):
I do not see any problem with these changes, they all look like the right kind of changes.  I am not savvy enough to know if anything is omitted that should be changed, but everything that was changed doesn't seem to be a problem.

pod/perldelta.pod
leonerd: No, it's intentional. The general theme here is that as much as possible we should be using/encouraging `use v5.36` as a general way to turn on signatures (and other features). I thought I might as well tidy up the perldelta entry about this related change while we're here.

duncand: That's not what I meant.  What I meant was that the old version looked like a typo for saying `experimental` rather than `feature`, and I was confirming this.  I agree that we want `use v5.36` now.

Grinnz: It was not a typo. experimental is a recommended module for enabling experimental features.

PR #19425, opened 2022-02-15
Modernise Tie::SubstrHash

JRaspass:
With the code-freeze rapidly approaching I thought I'd see if there's any interest in this patchset I've been toying with. I've tried to break it into logical commits but I will grant you that the top commit could use a bit more breakup. I'm happy to punt this to the next cycle and/or split chunks off if people agree.

The main motivation was to use `builtin::ceil` in anger, but I'm biased :grin:.

It took me a while to unpick the logic of this package, it didn't believe in passing values to subroutines and instead just localised package variables that naturally weren't even explicitly defined due to the lack of strict, very messy. I'm fairly confident in this conversion due to the reasonable test suite this package has but more eyes are always welcome!

I've done a few "style" changes where I felt they improved readability but this is probably very subjective.

I suspect this package gets very little use in the real world and is only in core for hysterical raisins.

### DIFF
--- a/lib/Tie/SubstrHash.pm
+++ b/lib/Tie/SubstrHash.pm
@@ -46,7 +46,7 @@ sub TIEHASH {
     my ($klen, $vlen, $tsize) = @_;
     my $rlen = 1 + $klen + $vlen;
     $tsize = [$tsize,
-	      findgteprime($tsize * 1.1)]; # Allow 10% empty.
+              findgteprime($tsize * 1.1)]; # Allow 10% empty.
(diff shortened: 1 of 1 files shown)


jkeenan:
The first 3 commits in your p.r. -- a4f8, 70cc, 5542 -- are fine by me and could go in right away.

I confess that I haven't been following the discussion about `builtin::` well enough to be able to evaluate your 4th commit -- 3b45.

The 5th commit (8d36) looks good but should get additional eyeballs.

PR #19428, opened 2022-02-16
implement a new command-line flag, -g

xenu:
It's a simpler alias for -0777. It was proposed in [RFC-0011](https://github.com/Perl/RFCs/blob/master/rfcs/rfc0011.md).

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -2098,6 +2098,7 @@ S_parse_body(pTHX_ char **env, XSINIT_t xsinit)
         case 'c':
         case 'd':
         case 'D':
+        case 'g':
         case '?':
         case 'h':
         case 'i':
@@ -3553,6 +3554,11 @@ Perl_moreswitches(pTHX_ const char *s)
         return s;
         NOT_REACHED; /* NOTREACHED */
     }
+    case 'g':
+        SvREFCNT_dec(PL_rs);
+        PL_rs = &PL_sv_undef;
+        sv_setsv(get_sv("/", GV_ADD), PL_rs);
+        return ++s;
 
     case '?':
         /* FALLTHROUGH */
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -64,6 +64,12 @@ C<use v5.36>.
 
 For more detail see L<perlop/Class Instance Operator>.
 
+=head2 -g command-line flag
+
+A new command-line flag, -g, is available. It is a simpler alias for -0777.
+
+For more information, see L<perlrun/-g>.
+
 =head1 Security
 
 XXX Any security-related notices go here.  In particular, any security
--- a/pod/perlrun.pod
+++ b/pod/perlrun.pod
@@ -4,7 +4,7 @@ perlrun - how to execute the Perl interpreter
 
 =head1 SYNOPSIS
 
-B<perl>	S<[ B<-sTtuUWX> ]>
+B<perl>	S<[ B<-gsTtuUWX> ]>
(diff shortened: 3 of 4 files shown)


perl.c
```diff
@@ -2098,6 +2098,7 @@ S_parse_body(pTHX_ char **env, XSINIT_t xsinit)
         case 'c':
         case 'd':
         case 'D':
+        case 'g':
```
oodler577, line 2101: This comment is a last ditch effort to dissuade the use of "-g", which is the VERY FIRST option listed in `gcc --help` under "debugging options" [1]; I strongly implore you all to use something that's more related to what this affects, namely "RS" (awk and is even explicitly referenced in `perlvar`).

My only "better": suggestion is lowercase "r", because it has to be a single character; and "r" may reflect "nothing" compared to the default "RS" of something (\n).

BUT, this is about _NOT_ using **-g** and going with something else. Literally _anything_ else (even `-G`) would be better than **-g** if there is even a sliver of concern regarding confusion and debugging options with those maybe more familiar with `gcc` than `perl`. Please, think of the children. :-)

Update - to further underscore my point, I ran this command over whatever SHA my local perl5 git repo is checked out to,

```
$ git grep "\-g " |wc -l
      64
```

I did a quick glance, and at least most of those lines refer to `gcc`'s `-g`.

[1] `gcc --help`
...
```
Debugging Options
           -g  -glevel  -gcoff  -gdwarf  -gdwarf-version -ggdb
           -grecord-gcc-switches  -gno-record-gcc-switches -gstabs  -gstabs+
           -gstrict-dwarf  -gno-strict-dwarf -gcolumn-info  -gno-column-info
           -gvms  -gxcoff  -gxcoff+  -gz[=type] -fdebug-prefix-map=old=new
           -fdebug-types-section -feliminate-dwarf2-dups
           -fno-eliminate-unused-debug-types -femit-struct-debug-baseonly
           -femit-struct-debug-reduced -femit-struct-debug-detailed[=spec-
           list] -feliminate-unused-debug-symbols  -femit-class-debug-always
           -fno-merge-debug-strings  -fno-dwarf2-cfi-asm -fvar-tracking
           -fvar-tracking-assignments
```

xenu: As I said in the RFC discussion, collisions with single letter flags in other tools are unavoidable. For example, "-r" often means "recursive", "-o" means "output", etc. I don't think it's worth worrying about.

Grinnz: I agree, the risk of confusion with a specific gcc flag seems minimal.

oodler577: Thank you for reconsidering it, which is what I was primarily asking you to do. I hope you both are correct and it ends up being a nothingburger on my part.

Leont: Agreed with xenu and Grinnz

demerphq:
On Wed, 16 Feb 2022, 22:03 Leon Timmermans, ***@***.***>
wrote:

>
> In perl.c <https://github.com/Perl/perl5/pull/19428#discussion_r808009385>
> :
>
> > @@ -2098,6 +2098,7 @@ S_parse_body(pTHX_ char **env, XSINIT_t xsinit)
>          case 'c':
>          case 'd':
>          case 'D':
> +        case 'g':
>
> Agreed with xenu and Grinnz
>

Me too. I like the choice actually, g for "gulp": "read it in gulp mode".
Makes sense to me.

Yves

>

skaji:
I think it is nice `-g` is mentioned in `perl --help` too.

PR #19434, opened 2022-02-17
Remove duplicated code in S_sv_2iuv_common()

t-a-k:
While considering #19404 I noticed that `S_sv_2iuv_common` has almost duplicated code for NV-to-IV/UV conversion.  I think those duplicated conversions should be identical, so converging them into single code will make the code clearer and compact.

No behavioral change is intended in this patch.
I also added a test file to verify this.


khwilliamson:
LGTM

Issue #18363, opened 2020-11-28
[Win64] -Duselongdouble perls built using mingw-w64 port of gcc-10.2.0 prone to crashing.


sisyphus:
My OS is Windows 7.
The compilers I've used are:
https://github.com/brechtsanders/winlibs_mingw/releases/download/10.2.0-11.0.0-9.0.0-r3/winlibs-i686-posix-dwarf-gcc-10.2.0-llvm-11.0.0-mingw-w64-8.0.0-r3.7z
and
https://github.com/brechtsanders/winlibs_mingw/releases/download/10.2.0-11.0.0-9.0.0-r3/winlibs-x86_64-posix-seh-gcc-10.2.0-llvm-11.0.0-mingw-w64-8.0.0-r3.7z

No problems building perl (both nvtype eq 'double' and nvtype eq 'long double') with the former of those two compilers (which is 32-bit).
But using the latter (64-bit) compiler presents crashes under certain conditions -  when and only when perl is built with nvtype of "long double".
When these crashes occur, they present a pop-up box stating that 'perl.exe has stopped working', with no additional diagnostics.

There are no issues if I switch to using mingw-w64 ports of gcc-8.3.0 that were built by the Strawberry Perl developers.

**Description**

The bug prevents the 'make test' step of the perl build process from occurring. As soon as the first of the test files are about to be run, the process crashes.
Also, having installed perl (via 'make install'), attempts at installing additional modules via the 'cpan' utility will crash.
In fact, merely trying to load CPAN.pm causes a crash.
Instead, one has to install modules manually:
perl Makefile.PL
make install

Attempts to run 'make test' again result in the crash.
Also, Inline::C is unusable, with a crash occurring during the attempted compilation of the C code.

**Steps to Reproduce**

This 6-line script drawn from code in CPAN.pm  (search that file for "SAVEOUT") prints out <code>OK so far</code> and then delivers the crash.

<code>{</code>
<code>my $x = *SAVEOUT; # avoid warning</code>
<code>open($x, ">&STDOUT");</code>
<code>print "OK so far\n";</code>
<code>}</code>
<code>print "Still OK\n";</code>

This 5-line variant is fine:

<code>{</code>
<code>open(my $x, ">&STDOUT");</code>
<code>print "OK so far\n";</code>
<code>}</code>
<code>print "Still OK\n";</code>

Having output:

<code>OK so far</code>
<code>Still OK</code>

it then terminates normally.
[...]

Issue #18911, opened 2021-06-21
perl-5.35.1 won't build on Windows 7, using mingw-w64 port of gcc-11.1.0


sisyphus:
Module:
dist/IO

**Description**

Attempting to build the latest devel release (perl-5.35.1) from source on Windows 7 with gcc-11.1.0 fails with:

```
In file included from IO.xs:18:
poll.h:14:12: fatal error: poll.h: No such file or directory
   14 | #  include <poll.h>
      |            ^~~~~~~~
compilation terminated.
gmake[1]: *** [Makefile:348: IO.o] Error 1
gmake[1]: Leaving directory 'C:/comp-1110/perl-5.35.1/dist/IO'
Unsuccessful make(dist/IO): code=512 at ..\make_ext.pl line 584.
make: *** [GNUmakefile:1696: Extensions] Error 2
```

I had no such trouble building blead (5.35.0) two weeks ago on the same system, using the same compiler.
That particular blead was at commit 1c1bafd87.

My first thought was that commit 52f0602 might have been the problem, but that doesn't make sense because, according to <code>git log</code> , commit 52f0602 pre-dates commit 1c1bafd87

**Perl configuration**

Obviously, there's no perl executable. Here, instead is 'miniperl -V'.
```
C:\comp-1110\perl-5.35.1\win32>..\miniperl -I..\lib -V
Summary of my perl5 (revision 5 version 35 subversion 1) configuration:

  Platform:
    osname=MSWin32
    osvers=6.1.7601
    archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -DWIN64 -fdiagnostics-color=never -DPERL_TEXTMODE_SCRIPTS
 -DMULTIPLICITY -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STDIO -fwrap
v -fno-strict-aliasing -mms-bitfields'
    optimize='-s -O2'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='11.1.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=8
    doublesize=8
    byteorder=12345678
[...]

PR #19437, opened 2022-02-18
enable x64 -Duselongdouble build for all mingw-runtime versions.

sisyphus:
See https://github.com/Perl/perl5/issues/18363

Reading through that ticket, I'm no longer sure if the problem with x64 -Duselongdouble builds begins with runtime v7, or runtime v8 or runtime v9.

I could check if need be - but I'm hopeful that the blanket inclusion of all x64 -Duselongdouble builds (irrespective of runtime version) will be quite acceptable.
In saying that, I'm assuming that the directive:
```
typedef NVTYPE NV __attribute__ ((aligned(8)));
```
comes at zero cost if it's not actually needed.
To be clear, it *is* needed for all runtime versions wrt x64 -Dusequadmath builds.
For x64 -Duselongdouble builds, the directive is apparently unnecessary until (somewhere) later than v6.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -2188,7 +2188,7 @@ You probably want to be using L<C</INT2PTR>> instead.
    here, so no allowance is being made for mingw.org
    compilers at this stage. -- sisyphus January 2021
 */
-#if defined(USE_QUADMATH) && defined(__MINGW64__)
+#if (defined(USE_LONG_DOUBLE) || defined(USE_QUADMATH)) && defined(__MINGW64__)
    /* 64-bit build, mingw-w64 compiler only */
    typedef NVTYPE NV __attribute__ ((aligned(8)));
 #else


sisyphus:
I've just double-checked, and the problem does indeed begin with runtime v7, and continues with v8 and (the current latest) v9.

xenu:
I see no connection between what this commit does and the commit message. It's working around a bug, it doesn't enable anything.

sisyphus:
> I see no connection between what this commit does and the commit message. It's working around a bug, it doesn't enable anything.

Full commit message was:
```
    enable x64 -Duselongdouble build for all mingw-runtime versions.

    See https://github.com/Perl/perl5/issues/18363
```
When, using mingw-w64 port of gcc-11.2.0 (mingw runtime v9, I build -Duselongdouble perl 5.34.0 or 5.35.x without this patch, then the perl that is created has some disabilities including:
1)<code>cpan</code> commands crash before outputting anything;
2)<code>make test</code> crashes before any tests are run (both for the perl test suite, and for the test suite of any module one might subsequently be wanting to install);
3) Inline::C scripts crash as soon as the C code has been built - ie just prior to the execution of the perl code;

These are the only disabilities I've encountered. However, I discovered them pretty quickly and haven't gone checking for anything else. (It's already established that, for my purposes at least, such perls are unsable.)
I use the term "enable" in the sense that this patch removes those disabilities.

If you want to wait for a better patch, just say so explicitly, and I'll close this PR.
It's not a big deal to me - it's really no trouble if I have to continue using this patch privately, ad infinitum.
It's just that, between this fix and fixes for https://github.com/Perl/perl5/issues/18911, https://rt.cpan.org/Public/Bug/Display.html?id=141325 and https://github.com/Perl/perl5/pull/19451, the points of divergence between my perl source and official perl source is growing.

I thought this PR might have been sufficiently uncontroversial as to allow it to be enacted, thereby removing one of those points of divergence.

Issue #19168, opened 2021-10-03
trie in variable-length lookbehind is not anchored


hvds:
```
% perl -v | head -2 | tail -1
This is perl 5, version 34, subversion 0 (v5.34.0) built for x86_64-linux

# this correctly fails to match
% perl -E 'no warnings qw{experimental}; "abfoo" =~ /(?=foo)(?<=(a|a))/ and say $1'

# this wrongly matches: it should only match if we ask for `ab`
% perl -E 'no warnings qw{experimental}; "abfoo" =~ /(?=foo)(?<=(a|ac))/ and say $1'
a
```

This should not match; my guess is that the trie code is unaware of whatever we've done to anchor the end of the lookbehind.

I won't get a chance to look at this for a few days: @khwilliamson @demerphq any ideas?

The original work was in 2fe8bdb, I haven't checked if there were later updates.

hvds:
I spent a little time looking at this, but the issue appears to be in the part of regexec.c that I've not understood since it was made iterative in 2006 (the 3600-odd line function S_regmatch). I believe the pattern is compiled correctly.

Note that the `/(?=foo)(?<=(a|ac))/` example above actually extracts the prefix to make `/(?=foo)(?<=a(|c))/`, but the zero-length match in the trie is not the issue: the same occurs with the unprefixed `/(?=foo)(?<=(a|bc))/`.

hvds:
@demerphq any chance you have headspace to look at this one? It's possible that the trie code and the vlb code just need to be a little more aware of each other, but it may go deeper than that.

demerphq:
On Fri, 18 Feb 2022, 23:16 Hugo van der Sanden, ***@***.***>
wrote:

> @demerphq <https://github.com/demerphq> any chance you have headspace to
> look at this one? It's possible that the trie code and the vlb code just
> need to be a little more aware of each other, but it may go deeper than that
>

Sure thing, I'll take a look.

Yves

>

demerphq:
On Sat, 19 Feb 2022 at 02:29, demerphq ***@***.***> wrote:

>
>
> On Fri, 18 Feb 2022, 23:16 Hugo van der Sanden, ***@***.***>
> wrote:
>
>> @demerphq <https://github.com/demerphq> any chance you have headspace to
>> look at this one? It's possible that the trie code and the vlb code just
>> need to be a little more aware of each other, but it may go deeper than that
>>
>
> Sure thing, I'll take a look.
>

[...]

PR #19442, opened 2022-02-20
Yves/fix 19168 variable lookbehind

demerphq:
Issue #19168 - Fix variable length lookbehind matches with alternations
    
    We were not validating that when (?<=a|ab) matched that the right hand
    side of the match lined up with the position of the assertion. Similar
    for (?<!a|ab) and related patterns, eg, (*positive_lookbehind:).
    
    Part of the difficulty here was that the SUCCEED node was serving too
    many purposes, necessitating a new regop LOOKBEHIND_END.
    
    Note these problems do NOT affect lookahead. This logic could use more
    tests, especially for nested lookbehind. I added a couple but my brain
    is mush right now and some tests from someone fresher would be helpful.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -2026,6 +2026,10 @@ EiR	|SV*	|invlist_contents|NN SV* const invlist		    \
 EixRT	|UV	|invlist_lowest|NN SV* const invlist
 ERS	|SV*	|make_exactf_invlist	|NN RExC_state_t *pRExC_state \
 					|NN regnode *node
+ES	|regnode_offset|reg_la_NOTHING	|NN RExC_state_t *pRExC_state \
+					|U32 flags|NN const char *type
+ES	|regnode_offset|reg_la_OPFAIL	|NN RExC_state_t *pRExC_state \
+					|U32 flags|NN const char *type
 ES	|regnode_offset|reg	|NN RExC_state_t *pRExC_state \
 				|I32 paren|NN I32 *flagp|U32 depth
 ES	|regnode_offset|regnode_guts|NN RExC_state_t *pRExC_state          \
--- a/embed.h
+++ b/embed.h
@@ -1096,6 +1096,8 @@
 #define rck_elide_nothing(a)	S_rck_elide_nothing(aTHX_ a)
 #define reg(a,b,c,d)		S_reg(aTHX_ a,b,c,d)
 #define reg2Lanode(a,b,c,d)	S_reg2Lanode(aTHX_ a,b,c,d)
+#define reg_la_NOTHING(a,b,c)	S_reg_la_NOTHING(aTHX_ a,b,c)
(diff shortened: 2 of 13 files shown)


hvds:
Thanks, this looks a lot cleaner than I had feared from your earlier caveats.

I'll try to come up with some additional tests over the next day or two, but in principle this already looks good to me.

demerphq:
Are there any objections to me merging this?

hvds:
I don't have an objection; due to a bereavement, though, I haven't managed to spend the time writing extra tests for it that I had hoped to.

demerphq:
On Wed, 23 Feb 2022 at 12:31, Hugo van der Sanden ***@***.***>
wrote:

> I don't have an objection; due to a bereavement,
>
I am very sorry to hear that. My condolences.


> though, I haven't managed to spend the time writing extra tests for it
> that I had hoped to.
>
Not to worry, we can add them later. I will probably add a few more myself
that I have in my command line history.

Since I have removed the patch that removes variable lookbehind from the PR
we are still in experimental mode so this is purely bug-fix territory.

We also have two options for the experimental status of this feature. This
one

  https://github.com/Perl/perl5/pull/19454

removes experimental status entirely, and this one

  https://github.com/Perl/perl5/pull/19455

keeps the experimental status, but reduces the scope of it to only those
cases where positive lookbehind contains a capturing buffer, thus to only
the cases where a possible future change in implementation might result in
a user visible change of behavior.

But all of that can wait, take care of yourself and your family/friends
first.

Cheers,
Yves


-- 

khwilliamson:
I fixed a typo in a commit message, but otherwise pushed these unchanged, the final commit being cbdefcf186788da38bae658c0796f265d10660c2

demerphq:
On Thu, 24 Feb 2022, 01:48 Karl Williamson, ***@***.***>
wrote:

> I fixed a typo in a commit message, but otherwise pushed these unchanged,
> the final commit being cbdefcf
> <https://github.com/Perl/perl5/commit/cbdefcf186788da38bae658c0796f265d10660c2>
>

Thanks! Sorry for the typo, I thought I spell checked all the commits, but
I guess in the rebase process I missed something.

Yves

>

demerphq:
On Wed, 23 Feb 2022 at 12:49, demerphq ***@***.***> wrote:

> We also have two options for the experimental status of this feature. This
> one
>
>   https://github.com/Perl/perl5/pull/19454
>
> removes experimental status entirely, and this one
>
>   https://github.com/Perl/perl5/pull/19455
>
> keeps the experimental status, but reduces the scope of it to only those
> cases where positive lookbehind contains a capturing buffer, thus to only
> the cases where a possible future change in implementation might result in
> a user visible change of behavior.
>

Hi Hugo, Karl,

Gentle reminder we have a decision to make on this subject...

Cheers,
Yves

PR #19445, opened 2022-02-21
use neutral form for parameter

bekuno:
User $val instead of $sv as for the other parameter in the samples.


demerphq:
LGTM.

jkeenan:
Merged to blead with increment to package version in 8432359dfe1d2f9809293629d155ad0aa46d7ace.  Thank you very much.

PR #19451, opened 2022-02-22
Posix win32 fegetround

sisyphus:
This PR enables POSIX::fegetround() and POSIX::fesetround() on Win32 perls built using mingw-w64 ports of gcc.

It's very simple to extend this PR to cover perls built with Visual Studio.
And I find that it works well with perls built using MSVC141 (Visual Studio 2019).
However, I don't know whether the earlier Microsoft compilers that perl promises to support (beginning with MSVC120) are up to the task.
For that reason I've limited the PR to Win32 perls built using gcc.

If someone can confirm that MSVC120 provides the requisite functionality, then I'll happily expand this PR to provide that additional coverage.


ext/POSIX/POSIX.xs
```diff
@@ -23,6 +23,9 @@ static int not_here(const char *s);
 #include <errno.h>
 #ifdef WIN32
 #include <sys/errno2.h>
+#  ifdef __MINGW32__
+#  include <fenv.h>
+#  endif
```
xenu, line 28: I think it would be better to define `i_fenv` in the config files instead of modifying POSIX. It will have the same effect.

xenu:
>If someone can confirm that MSVC120 provides the requisite functionality, then I'll happily expand this PR to provide that additional coverage.

Visual C++ 2013 (MSVC120) does provide that function.

PS. Could you squash your commits?

xenu:
Also, a side-note: most compilers don't support setting the rounding mode out of the box. GCC requires [-frounding-math](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Optimize-Options.html#index-frounding-math) switch, Visual C++ needs [`#pragma fenv_access(on)`](https://docs.microsoft.com/en-us/cpp/preprocessor/fenv-access?view=msvc-170) while other compilers might need [`#pragma STDC FENV_ACCESS ON`](https://port70.net/~nsz/c/c11/n1570.html#7.6.1).

khwilliamson:
@sisyphus the ball is in your court to respond

sisyphus:
I think I shouldn't have created this PR when I did. (I was fooled by the fact that the POSIX tests all passed ... but they don't test everything.)
I now want to spend some more time on this to ensure that Windows has the same level of functionality in this regard, as occurs on Linux - but it could be a while before that round tuit turns up.
Certainly, defining <code>i_fenv</code> will be one of the changes - which should eliminate the need to alter POSIX.xs.
Thanks @xenu.
But even with that, I still don't get the same rounding capability on Windows as I do on Ubuntu. 
I'll take a punt that the correct thing for me to do now is to close this, and open a new PR for it after I've done the groundwork properly and established that it's as functional on Windows as it is on Linux.

PR #19453, opened 2022-02-22
Perl_gv_init_pvn() must handle a stash holding the value -1

### LINKED ISSUE #19450: BBC: Bleadperl breaks STEVAN/MOP-0.14.tar.gz
**Description**

My smokers have stopped producing PASS reports. Bisect leads me to
``` text
bb5bc97fde9ef2f16cffdfb44d7f42a03311ab2f is the first bad commit
commit bb5bc97fde9ef2f16cffdfb44d7f42a03311ab2f
Date:   Fri Jul 2 12:45:42 2021 +0000

    Don't set SVf_POK in Perl_sv_2pv_flags() when caching the string for an IV.
```
Sample fail reports will appear soon on http://matrix.cpantesters.org/?dist=MOP+0.14

**Steps to Reproduce**

cpan -i STEVAN/MOP-0.14.tar.gz

**Expected behavior**

Should install the module

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 35 subversion 9) configuration:
  Commit id: bb5bc97fde9ef2f16cffdfb44d7f42a03311ab2f
  Platform:
    osname=linux
    osvers=5.15.0-2-amd64
    archname=x86_64-linux
    uname='linux k93msid 5.15.0-2-amd64 #1 smp debian 5.15.5-2 (2021-12-18) x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.8-110-gbb5bc97fde/da1c -Dmyhostname=k93msid -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Uuseithreads -Uuselongdouble -DEBUGGING=-g'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='cc'
    optimize='-O2 -g'
    ccversion=''
    gccversion='11.2.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    libc=libc-2.33.so
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.33'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'


  Compile-time options:
    HAS_TIMES
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_USE_DEVEL
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under linux
  Compiled at Feb 22 2022 09:58:49
  @INC:
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.8-110-gbb5bc97fde/da1c/lib/site_perl/5.35.9/x86_64-linux
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.8-110-gbb5bc97fde/da1c/lib/site_perl/5.35.9
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.8-110-gbb5bc97fde/da1c/lib/5.35.9/x86_64-linux
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.8-110-gbb5bc97fde/da1c/lib/5.35.9

```

nwc10:
As a space optimisation, sub declarations are not stored in the symbol table as a PVGV pointing to a PVCV (ie a real subroutine). Instead a subroutine declaration with a prototype is stored as a string with just the prototype, and a declaration with no prototype is stored as the integer -1

This was implemented in 1998 by commit 8472ac73d6d80294:

    "Sub declaration cost reduced from ~500 to ~100 bytes"

This space optimisation needs to be invisible to regular Perl code. To make it work requires some trickery when handling code such as

     sub foo ($$);
     *foo = [];

The typeglob lookup on foo needs to "upgrade" the symbol table entry to a PVGV, and then create a PVCV in the {CODE} slot with the prototype $$.  (The typeglob is then available so that [] can be assigned to the {ARRAY} slot. For the code above the upgrade happens at compile time, the assignment at runtime)

A regression was introduced by the recently merged commit bb5bc97fde9ef2f1:

    Don't set SVf_POK in Perl_sv_2pv_flags() when caching the string for an IV.

    This permits XS code (such as serialisers) to distinguish between values
    that started as IVs but had a string representation cached, and values that
    started as PVs but had an (exact) integer representation cached.

    As implemented, the change in flags means that Perl_sv_2pv_flags() will be
    entered each time the string for an IV is wanted. The next commit will fix
    SvPV() and the other macros to avoid calling Perl_sv_2pv_flags() more than
    once, restoring the previous behaviour.

Specifically, code which reduces to this:

     sub foo;
     $::{foo} . "";
     eval q{*foo = []};

changed behaviour (to an assertion failure)

Previously the runtime behaviour of the middle line was (simplifying slightly) to convert the IV -1 to the PVIV "-1". The third line would then see this as a(n illegal) prototype "-1", but the code in Perl_gv_init_pvn() would diligently move this across to the PVCV it created, and finish with a structurally valid PVGV and PVCV

With the recent commit, the second line would still convert to PVIV, but no longer set the flag bit SVf_POK. This meant that with this same test case, the SvPOK() test in Perl_gv_init_pvn() was no longer true, and a different code path was taken, resulting in broken assumptions and an assertion failure (when assertions are enabled).

Fixing this regression also fixes a long standing bug, likely present since the first commit in 1998. This pathological code:

     sub foo;
     $::{foo} . "";
     eval q{ sub foo {} };

use to incorrectly warn:

    Prototype mismatch: sub main::foo (-1) vs none at (eval 1) line 1.

because the placeholder -1 for "no prototype" had been converted to a string, that string then copied because it seemed to be a placeholder prototype, and then the warning issued on the mistaken incorrect prototype.

Fixes #19450

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -391,6 +391,52 @@ Perl_gv_init_pv(pTHX_ GV *gv, HV *stash, const char *name, U32 flags)
    gv_init_pvn(gv, stash, name, strlen(name), flags);
 }
 
+/* Packages in the symbol table are "stashes" - hashes where the keys are symbol
+   names and the values are typeglobs. The value $foo::bar is actually found
+   by looking up the typeglob *foo::{bar} and then reading its SCALAR slot.
+
+   At least, that's what you see in Perl space if you use typeglob syntax.
+   Usually it's also what's actually stored in the stash, but for some cases
+   different values are stored (as a space optimisation) and converted to full
+   typeglobs "on demand" - if a typeglob syntax is used to read a value. It's
+   the job of this function, Perl_gv_init_pvn(), to undo any trickery and
+   replace the SV stored in the stash with the regular PVGV structure that it is
+   a shorthand for. This has to be done "in-place" by upgrading the actual SV
+   that is already stored in the stash to a PVGV.
+
+   As the public documentation above says:
+       Converting any scalar that is C<SvOK()> may produce unpredictable
+       results and is reserved for perl's internal use.
+
+   Values that can be stored:
+
+   * plain scalar - a subroutine declaration
+     The scalar's string value is the subroutine prototype; the integer -1 is
+     "no prototype". ie shorthand for sub foo ($$); or sub bar;
+   * reference to a scalar - a constant. ie shorthand for sub PI() { 4; }
+   * reference to a sub - a subroutine (avoids allocating a PVGV)
+
+   The earliest optimisation was subroutine declarations, implemented in 1998
+   by commit 8472ac73d6d80294:
+      "Sub declaration cost reduced from ~500 to ~100 bytes"
+
+   This space optimisation needs to be invisible to regular Perl code. For this
+   code:
+
+         sub foo ($$);
+         *foo = [];
+
+   When the first line is compiled, the optimisation is used, and $::{foo} is
+   assigned the scalar '$$'. No PVGV or PVCV is created.
+
+   When the second line encountered, the typeglob lookup on foo needs to
+   "upgrade" the symbol table entry to a PVGV, and then create a PVCV in the
+   {CODE} slot with the prototype $$ and no body. The typeglob is then available
+   so that [] can be assigned to the {ARRAY} slot. For the code above the
+   upgrade happens at compile time, the assignment at runtime.
+
+   Analogous code unwinds the other optimisations.
+*/
 void
 Perl_gv_init_pvn(pTHX_ GV *gv, HV *stash, const char *name, STRLEN len, U32 flags)
 {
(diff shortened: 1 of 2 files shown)


demerphq:
LGTM.

PR #19455, opened 2022-02-23
Reduce scope of experimental warnings with lookbehind

demerphq:
The only case where we need to warn about variable length lookbehind is
when we are doing postive lookbehind and capturing buffers are involved.

### DIFF
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -1652,13 +1652,27 @@ matches a word that follows a tab, without including the tab in C<$&>.
 Prior to Perl 5.30, it worked only for fixed-width lookbehind, but
 starting in that release, it can handle variable lengths from 1 to 255
 characters as an experimental feature.  The feature is enabled
-automatically if you use a variable length lookbehind assertion, but
-will raise a warning at pattern compilation time, unless turned off, in
-the C<experimental::vlb> category.  This is to warn you that the exact
-behavior is subject to change should feedback from actual use in the
-field indicate to do so; or even complete removal if the problems found
-are not practically surmountable.  You can achieve close to pre-5.30
-behavior by fatalizing warnings in this category.
+automatically if you use a variable length positive lookbehind assertion.
+
+In Perl 5.35.10 the scope of the experimental nature of this construct
+has been reduced, and experimental warnings will only be produced when
+the construct contains capturing parenthesis. The warnings will be
+raised at pattern compilation time, unless turned off, in the
+C<experimental::vlb> category.  This is to warn you that the exact
+contents of capturing buffers in a variable length positive lookbehind
+is not well defined and is subject to change in a future release of perl.
+
+Currently if you use capture buffers inside of a positive variable length
+lookbehind the result will be the longest and thus leftmost match possible.
+This means that
+
+    "aax" =~ /(?=x)(?<=(a|aa))/
+    "aax" =~ /(?=x)(?<=(aa|a))/
+    "aax" =~ /(?=x)(?<=(a{1,2}?)/
+    "aax" =~ /(?=x)(?<=(a{1,2})/
+
+will all result in C<$1> containing C<"aa">. It is possible in a future
+release of perl we will change this behavior.
 
 There is a special form of this construct, called C<\K>
 (available since Perl 5.10.0), which causes the
@@ -1712,16 +1726,44 @@ matches any occurrence of "foo" that does not follow "bar".
 Prior to Perl 5.30, it worked only for fixed-width lookbehind, but
 starting in that release, it can handle variable lengths from 1 to 255
 characters as an experimental feature.  The feature is enabled
-automatically if you use a variable length lookbehind assertion, but
-will raise a warning at pattern compilation time, unless turned off, in
-the C<experimental::vlb> category.  This is to warn you that the exact
-behavior is subject to change should feedback from actual use in the
-field indicate to do so; or even complete removal if the problems found
-are not practically surmountable.  You can achieve close to pre-5.30
-behavior by fatalizing warnings in this category.
+automatically if you use a variable length negative lookbehind assertion.
+
+In Perl 5.35.10 the scope of the experimental nature of this construct
+has been reduced, and experimental warnings will only be produced when
+the construct contains capturing parentheses. The warnings will be
+raised at pattern compilation time, unless turned off, in the
+C<experimental::vlb> category.  This is to warn you that the exact
+contents of capturing buffers in a variable length negative lookbehind
+is not well defined and is subject to change in a future release of perl.
+
+Currently if you use capture buffers inside of a negative variable length
+lookbehind the result may not be what you expect, for instance:
+
+    say "axfoo"=~/(?=foo)(?<!(a|ax)(?{ say $1 }))/ ? "y" : "n";
+
+will output the following:
+
+    a
+    no
+
+which does not make sense as this should print out "ax" as the "a" does
+not line up at the correct place. Another example would be:
+
+    say "yes: '$1-$2'" if "aayfoo"=~/(?=foo)(?<!(a|aa)(a|aa)x)/;
+
+will output the following:
+
+    yes: 'aa-a'
+
+It is possible in a future release of perl we will change this behavior
+so both of these examples produced more reasonable output.
+
+Note that we are confident that the construct will match and reject
+patterns appropriately, the undefined behavior strictly relates to the
+value of the capture buffer during or after matching.
 
 There is a technique that can be used to handle variable length
-lookbehinds on earlier releases, and longer than 255 characters.  It is
+lookbehind on earlier releases, and longer than 255 characters.  It is
 described in
 L<http://www.drregex.com/2019/02/variable-length-lookbehinds-actually.html>.
 
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -27,6 +27,20 @@ here, but most should go in the L</Performance Enhancements> section.
 
 [ List each enhancement as a =head2 entry ]
 
+=head2 Variable length lookbehind is mostly no longer considered experimental.
+
+Prior to this release any form of variable length lookbehind was
+considered experimental. With this release the experimental status has
+been reduced to cover only lookbehind that contains capturing parenthesis.
+This is because it is not clear if
+
+    "aaz"=~/(?=z)(?<=(a|aa))/
+
+should match and leave $1 equaling "a" or "aa". Currently it will match
+the longest possible alternative, "aa". We are confident that the overall
+construct will now match only when it should, we are not confident that we
+will keep the current "longest match" behavior.
+
 =head1 Security
 
 XXX Any security-related notices go here.  In particular, any security
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -7628,6 +7628,26 @@ You can also say
 to apply C</aa> to all regular expressions compiled within its scope.
 See L<re>.
 
+=item Variable length positive lookbehind with capturing is experimental in regex m/%s/
+
+(W) Variable length positive lookbehind with capturing is not well defined. This
+warning alerts you to the fact that you are using a construct which may
+change in a future version of perl. See the
+L<< documentation of Positive Lookbehind in perlre|perlre/"C<(?<=I<pattern>)>" >>
+for details. You may silence this warning with the following:
+
+    no warnings 'experimental::vlb';
+
+=item Variable length negative lookbehind with capturing is experimental in regex m/%s/
+
+(W) Variable length negative lookbehind with capturing is not well defined. This
+warning alerts you to the fact that you are using a construct which may
+change in a future version of perl. See the
+L<< documentation of Negative Lookbehind in perlre|perlre/"C<(?<!I<pattern>)>" >>
+for details. You may silence this warning with the following:
+
+    no warnings 'experimental::vlb';
+
 =item "%s" variable %s masks earlier declaration in same %s
 
 (W shadow) A "my", "our" or "state" variable has been redeclared in the
--- a/regcomp.c
+++ b/regcomp.c
@@ -4444,21 +4444,23 @@ S_rck_elide_nothing(pTHX_ regnode *node)
 
 /* the return from this sub is the minimum length that could possibly match */
 STATIC SSize_t
-S_study_chunk(pTHX_ RExC_state_t *pRExC_state, regnode **scanp,
-                        SSize_t *minlenp, SSize_t *deltap,
-                        regnode *last,
-                        scan_data_t *data,
-                        I32 stopparen,
-                        U32 recursed_depth,
-                        regnode_ssc *and_withp,
-                        U32 flags, U32 depth, bool was_mutate_ok)
-                        /* scanp: Start here (read-write). */
-                        /* deltap: Write maxlen-minlen here. */
-                        /* last: Stop before this one. */
-                        /* data: string data about the pattern */
-                        /* stopparen: treat close N as END */
-                        /* recursed: which subroutines have we recursed into */
-                        /* and_withp: Valid if flags & SCF_DO_STCLASS_OR */
+S_study_chunk(pTHX_
+    RExC_state_t *pRExC_state,
+    regnode **scanp,        /* Start here (read-write). */
+    SSize_t *minlenp,
+    SSize_t *deltap,        /* Write maxlen-minlen here. */
+    regnode *last,          /* Stop before this one. */
+    scan_data_t *data,      /* string data about the pattern */
+    I32 stopparen,          /* treat CLOSE-N as END, see GOSUB */
+    U32 recursed_depth,     /* how deep have we recursed via GOSUB */
+    regnode_ssc *and_withp, /* Valid if flags & SCF_DO_STCLASS_OR */
+    U32 flags,              /* flags controlling this call, see SCF_ flags */
+    U32 depth,              /* how deep have we recursed period */
+    bool was_mutate_ok      /* TRUE if in-place optimizations are allowed.
+                               FALSE only if the caller (recursively) was
+                               prohibited from modifying the regops, because
+                               a higher caller is holding a ptr to them. */
+)
 {
     SSize_t final_minlen;
     /* There must be at least this number of characters to match */
--- a/regcomp.c
+++ b/regcomp.c
@@ -4611,7 +4611,7 @@ S_study_chunk(pTHX_
                     ssc_init_zero(pRExC_state, &accum);
 
                 while (OP(scan) == code) {
-                    SSize_t deltanext, minnext, fake;
+                    SSize_t deltanext, minnext, fake_last_close = 0;
                     I32 f = 0;
                     regnode_ssc this_class;
 
@@ -4624,7 +4624,7 @@ S_study_chunk(pTHX_
                         data_fake.last_closep = data->last_closep;
                     }
                     else
-                        data_fake.last_closep = &fake;
+                        data_fake.last_closep = &fake_last_close;
 
                     data_fake.pos_delta = delta;
                     next = regnext(scan);
@@ -6040,7 +6040,8 @@ S_study_chunk(pTHX_
                    In this case we can't do fixed string optimisation.
                 */
 
-                SSize_t deltanext, minnext, fake = 0;
+                SSize_t deltanext, minnext;
+                SSize_t fake_last_close = 0;
                 regnode *nscan;
                 regnode_ssc intrnl;
                 int f = 0;
@@ -6051,7 +6052,7 @@ S_study_chunk(pTHX_
                     data_fake.last_closep = data->last_closep;
                 }
                 else
-                    data_fake.last_closep = &fake;
+                    data_fake.last_closep = &fake_last_close;
                 data_fake.pos_delta = delta;
                 if ( flags & SCF_DO_STCLASS && !scan->flags
                      && OP(scan) == IFMATCH ) { /* Lookahead */
@@ -6127,7 +6128,7 @@ S_study_chunk(pTHX_
                    length of the pattern, something we won't know about
                    until after the recurse.
                 */
-                SSize_t deltanext, fake = 0;
+                SSize_t deltanext, fake_last_close = 0;
                 regnode *nscan;
                 regnode_ssc intrnl;
                 int f = 0;
@@ -6151,7 +6152,7 @@ S_study_chunk(pTHX_
                     }
                 }
                 else
-                    data_fake.last_closep = &fake;
+                    data_fake.last_closep = &fake_last_close;
                 data_fake.flags = 0;
                 data_fake.substrs[0].flags = 0;
                 data_fake.substrs[1].flags = 0;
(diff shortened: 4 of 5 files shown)


hvds:
First thought here: should it include negative lookbehind too, on the grounds that those captures _can_ come into play within the lookbehind itself? I feel the marginal cost to users of including it would be minimal.

pod/perlre.pod
```diff
@@ -1652,13 +1652,27 @@ matches a word that follows a tab, without including the tab in C<$&>.
...
-the C<experimental::vlb> category.  This is to warn you that the exact
-behavior is subject to change should feedback from actual use in the
-field indicate to do so; or even complete removal if the problems found
-are not practically surmountable.  You can achieve close to pre-5.30
-behavior by fatalizing warnings in this category.
+automatically if you use a variable length positive lookbehind assertion.
+
+In Perl 5.35.10 the scope of the experimental nature of this construct
+has been reduced, and experimental warnings will only be produced when
+the construct contains capturing parenthesis. The warnings will be
```
hvds, line 1659: s/parenthesis/parentheses/

hvds (commented):
The code looks reasonable to me, just one typo.

demerphq:
On Sun, 27 Feb 2022, 00:22 Hugo van der Sanden, ***@***.***>
wrote:

> First thought here: should it include negative lookbehind too, on the
> grounds that those captures *can* come into play within the lookbehind
> itself? I feel the marginal cost to users of including it would be minimal
>

At first i thought the same, but I couldn't come up with a single case
where a capture in a negative lookbehind actually captures, and concluded
capturing in nlb (and probably nla) never captures as the construct fails
if it matches, so the capture buffer is cleared. If you can find a case to
the contrary then I would agree that too should be included, not finding a
case quite possibly could be a lack of imagination on my part.

Happy to have secondary confirmation of my analysis.

Thanks for the follow up.

Cheers
Yves

>

hvds:
> On Sun, 27 Feb 2022, 00:22 Hugo van der Sanden, ***@***.***>
> wrote:
> > First thought here: should it include negative lookbehind too, on the grounds that those captures *can* come into play within the lookbehind itself? I feel the marginal cost to users of including it would be minimal
> 
> At first i thought the same, but I couldn't come up with a single case where a capture in a negative lookbehind actually captures, and concluded capturing in nlb (and probably nla) never captures as the construct fails if it matches, so the capture buffer is cleared. If you can find a case to the contrary then I would agree that too should be included, not finding a case quite possibly could be a lack of imagination on my part.

By "within the lookbehind itself" I meant cases such as `/(?<!(.)\1)/`. Trying it with a handy 5.34 now, I find that is disallowed ("Lookbehind longer than 255 not implemented"). I haven't checked, but I'd guess that because the length of `\1` isn't known it gets treated as infinite.

My inclination would be to include it with the positive case anyway: if it is never useful, the marginal cost drops to zero, while the value of the wiggle room it gives us is slim but non-zero. It would also make both the implementation and the explanation a bit simpler.

hvds:
I've run out of time tonight, but it occurs to me to wonder whether the parno counting also picks up alternatives such as named capture groups (including reuse of a name introduced outside the lookbehind) or the `/(?|...)/` construct (again, where the parno is first introduced outside a lookbehind). I'll try to test that tomorrow.

demerphq:
On Sun, 27 Feb 2022 at 02:20, Hugo van der Sanden ***@***.***>
wrote:

> On Sun, 27 Feb 2022, 00:22 Hugo van der Sanden, *@*.***>
> wrote:
>
> First thought here: should it include negative lookbehind too, on the
> grounds that those captures *can* come into play within the lookbehind
> itself? I feel the marginal cost to users of including it would be minimal
>
> At first i thought the same, but I couldn't come up with a single case
> where a capture in a negative lookbehind actually captures, and concluded
> capturing in nlb (and probably nla) never captures as the construct fails
> if it matches, so the capture buffer is cleared. If you can find a case to
> the contrary then I would agree that too should be included, not finding a
> case quite possibly could be a lack of imagination on my part.
>
> By "within the lookbehind itself" I meant cases such as /(?<!(.)\1)/.
> Trying it with a handy 5.34 now, I find that is disallowed ("Lookbehind
> longer than 255 not implemented"). I haven't checked, but I'd guess that
> because the length of \1 isn't known it gets treated as infinite.
>
Oh! Good play! I had not thought of /that/. Make sense now!

If we allowed negative lookbehind with captures without the experimental
flag and then someone tweaked this logic then we would be back in undefined
territory, so I agree that we should include negative.


> My inclination would be to include it with the positive case anyway: if it
> is never useful, the marginal cost drops to zero, while the value of the
> wiggle room it gives us is slim but non-zero. It would also make both the
> implementation and the explanation a bit simpler.
>
Your example above seals the deal for me, I'll adjust the patch
accordingly. Nice one.

Cheers,
yves
-- 

demerphq:
On Sun, 27 Feb 2022 at 02:21, Hugo van der Sanden ***@***.***>
wrote:

> I've run out of time tonight, but it occurs to me to wonder whether the
> parno counting also picks up alternatives such as named capture groups
> (including reuse of a name introduced outside the lookbehind) or the
> /(?|...)/ construct (again, where the parno is first introduced outside a
> lookbehind). I'll try to test that tomorrow.
>

Not sure I follow this, but looking forward to learning more from you after
you have some rest.

cheers,
Yves

hvds:
I mean something like this:
```
# named capture correctly triggers warning: ok
% ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<d>y)|zz) }x'
Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
  in m/ (?=(?<c>x)) | (?<=(?<d>y)|zz)  <-- HERE / at -e line 1.
% 

# .. even if reused: ok
% ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<c>y)|zz) }x'
Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
  in m/ (?=(?<c>x)) | (?<=(?<c>y)|zz)  <-- HERE / at -e line 1.
% 

# reuse of numbered capture via (?| ...) not ok
% ./miniperl -we 'qr{ (?| (?=(x)) | (?<=(y)|zz) ) }x'
% 
```

That last case needs looking at: it should warn too.

demerphq:
On Sun, 27 Feb 2022 at 14:06, Hugo van der Sanden ***@***.***>
wrote:

> I mean something like this:
>
> # named capture correctly triggers warning: ok
> % ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<d>y)|zz) }x'
> Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
>   in m/ (?=(?<c>x)) | (?<=(?<d>y)|zz)  <-- HERE / at -e line 1.
> %
>
> # .. even if reused: ok
> % ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<c>y)|zz) }x'
> Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
>   in m/ (?=(?<c>x)) | (?<=(?<c>y)|zz)  <-- HERE / at -e line 1.
> %
>
> # reuse of numbered capture via (?| ...) not ok
> % ./miniperl -we 'qr{ (?| (?=(x)) | (?<=(y)|zz) ) }x'
> %
>
> That last case needs looking at: it should warn too
>
Gnash. This is like the 12th time that stupid thing has come back to bite
me. If we had had experimental back in the day it have been marked
experimental, and it would still be marked experimental it has caused so
much trouble.

Thanks for bringing this to my attention. That *deeply* sucks.

Cheers,
Yves

-- 

hvds:
> That *deeply* sucks.

Well, while not ideal, it would not be disastrous if that's the only construct we'd fail to notice: it's a deeply weird thing to do.

Another case to rule in or out is GOSUB; I have a half-memory we might wrap an ersatz capture around it as part of the implementation, but I haven't verified that.

t/porting/known_pod_issues.dat
```diff
@@ -389,6 +389,7 @@ pod/perlandroid.pod	Verbatim line length including indents exceeds 78 by	3
 pod/perlbook.pod	Verbatim line length including indents exceeds 78 by	1
 pod/perldebguts.pod	Verbatim line length including indents exceeds 78 by	24
 pod/perldebtut.pod	Verbatim line length including indents exceeds 78 by	2
+pod/perldiag.pod	Apparent broken link	2
```
khwilliamson: There's something wrong here that should be fixed.

khwilliamson (approved):
Otherwise LGTM

demerphq:
On Mon, 28 Feb 2022, 05:34 Karl Williamson, ***@***.***>
wrote:

>
> In t/porting/known_pod_issues.dat
> <https://github.com/Perl/perl5/pull/19455#discussion_r815494897>:
>
> > @@ -389,6 +389,7 @@ pod/perlandroid.pod	Verbatim line length including indents exceeds 78 by	3
>  pod/perlbook.pod	Verbatim line length including indents exceeds 78 by	1
>  pod/perldebguts.pod	Verbatim line length including indents exceeds 78 by	24
>  pod/perldebtut.pod	Verbatim line length including indents exceeds 78 by	2
> +pod/perldiag.pod	Apparent broken link	2
>
> There's something wrong here that should be fixed
>

Podchecker doesn't seem to grok X<> style index entries, I tried several
things, and none of them worked and I eventually grew tired of wrestling
with it and used the --regen escape hatch. Promise i tried pretty hard to
get it to pass without using that! I'd truly love to know what I could have
done differently if you have advice. The cause is the two L<> references to
the docs for positive and negative lookbehind I added to the perldiag
entries.

Yves

>

demerphq:
On Mon, 28 Feb 2022, 00:24 Hugo van der Sanden, ***@***.***>
wrote:

> That *deeply* sucks.
>
> Well, while not ideal, it would not be disastrous if that's the only
> construct we'd fail to notice: it's a deeply weird thing to do.
>
I agree. But I think I can probably work out a solution, although a priori
I wonder how invasive it might be.

Another case to rule in or out is GOSUB; I have a half-memory we might wrap
> an ersatz capture around it as part of the implementation, but I haven't
> verified that.
>

I imagine gosub out of one of these constructs is forbidden, but I'll check
and if necessary take care if it, gosub in should be fine.

Thanks for the pointers, I'll check up on it.

Yves

>

hvds:
> The cause is the two L<> references to the docs for positive and negative lookbehind I added to the perldiag entries.

Why have two diagnostics rather than one? When I suggested earlier that including negative lookbehind too would "also make both the implementation and the explanation a bit simpler", it was in part with the assumption that things like warning messages would simply lose "positive" from their text.

demerphq:
On Mon, 28 Feb 2022, 21:15 Hugo van der Sanden, ***@***.***>
wrote:

> The cause is the two L<> references to the docs for positive and negative
> lookbehind I added to the perldiag entries.
>
> Why have two diagnostics rather than one? When I suggested earlier that
> including negative lookbehind too would "also make both the implementation
> and the explanation a bit simpler", it was in part with the assumption that
> things like warning messages would simply lose "positive" from their text
>
Because they are different constructs with different issues, and rather
different example cases and explanations. As a user I want more detailed
tailored errors not generic ones. Anyway, lookbehind isn't documented as a
single thing with two flavors but rather as a set of two different
constructs each of which can be spelled three different ways. Which would I
link to? Both? Rewrite the docs totally? Why not DTRT instead?

I don't see any real win with this simpler implementation you propose, it's
a neglible drop in a large ocean. I'd rather have the tiny added
complexity, which is really nothing at all in the grand scheme of things,
if it gives a more useful user experience. We should be optimizing for our
users not us as core devs with this kind of thing.

Also it wouldn't make a difference to the subject of this subthread, there
would still be a bogus report of a broken link, or two, podchecker is just
broken in how it handles X<> index entries and L<> to them. The code, docs
and diags are written, if you want to spend energy on this subject I think
it would better placed fixing the pod checker if I may be so bold to say,
it would be of far greater benefit to the project as a whole than undoing
the work I already have done.

Cheers
Yves

hvds:
> > > The cause is the two L<> references to the docs for positive and negative lookbehind I added to the perldiag entries.
> >
> > Why have two diagnostics rather than one?
> 
> [...] it wouldn't make a difference to the subject of this subthread, there would still be a bogus report of a broken link, or two, podchecker is just broken in how it handles X<> index entries and L<> to them.

Ah, my apologies; I interpreted your original comment above as implying that the problem occurred _because_ there were two entries.

> if you want to spend energy on this subject [...]

I was just trying to help, no offence or sapping of energy was intended.

khwilliamson:
On 2/28/22 06:36, Yves Orton wrote:
> On Mon, 28 Feb 2022, 21:15 Hugo van der Sanden, ***@***.***>
> wrote:
> 
>  > The cause is the two L<> references to the docs for positive and negative
>  > lookbehind I added to the perldiag entries.
>  >
>  > Why have two diagnostics rather than one? When I suggested earlier that
>  > including negative lookbehind too would "also make both the 
> implementation
>  > and the explanation a bit simpler", it was in part with the 
> assumption that
>  > things like warning messages would simply lose "positive" from their text
>  >
> Because they are different constructs with different issues, and rather
> different example cases and explanations. As a user I want more detailed
> tailored errors not generic ones. Anyway, lookbehind isn't documented as a
> single thing with two flavors but rather as a set of two different
> constructs each of which can be spelled three different ways. Which would I
> link to? Both? Rewrite the docs totally? Why not DTRT instead?
> 
> I don't see any real win with this simpler implementation you propose, it's
> a neglible drop in a large ocean. I'd rather have the tiny added
> complexity, which is really nothing at all in the grand scheme of things,
> if it gives a more useful user experience. We should be optimizing for our
> users not us as core devs with this kind of thing.
> 
> Also it wouldn't make a difference to the subject of this subthread, there
> would still be a bogus report of a broken link, or two, podchecker is just
> broken in how it handles X<> index entries and L<> to them. The code, docs
> and diags are written, if you want to spend energy on this subject I think
> it would better placed fixing the pod checker if I may be so bold to say,
> it would be of far greater benefit to the project as a whole than undoing
> the work I already have done.
> 

It appears you think that X<> entries are what gets linked to.  That is 
incorrect, and explains why you couldn't get it to work.

X<> is used for indexing, not linking.  As the docs say:

X<topic name>" -- an index entry
          This is ignored by most formatters, but some may use it for 
building indexes. It always renders as empty-string.

Maybe buildtoc uses them; I've never looked into it.  That would be the 
only place I can think of that uses them in the Perl core.

L<> is used for linking. The target must be a "section" in the linked-to 
page.  A "section" is introduced by =head or =item.

perlpodspec says

Note that section names might contain markup. I.e., if a section
          starts with:

            =head2 About the C<-M> Operator

          or with:

            =item About the C<-M> Operator

          then a link to it would look like this:

          L<somedoc/About the C<-M> Operator>

The podchecker is correctly doing its job here.

If it were me, I would make the link
   L<perlre/Lookaround Assertions>

and make them start reading/skimming there.

demerphq:
On Mon, 28 Feb 2022 at 18:18, Hugo van der Sanden ***@***.***>
wrote:

> The cause is the two L<> references to the docs for positive and negative
> lookbehind I added to the perldiag entries.
>
> Why have two diagnostics rather than one?
>
> [...] it wouldn't make a difference to the subject of this subthread,
> there would still be a bogus report of a broken link, or two, podchecker is
> just broken in how it handles X<> index entries and L<> to them.
>
> Ah, my apologies; I interpreted your original comment above as implying
> that the problem occurred *because* there were two entries.
>
No, sorry if I wasnt clear.


> if you want to spend energy on this subject [...]
>
> I was just trying to help, no offence or sapping of energy was intended.
>

I just meant that if you were thinking of reworking all the docs it might
be better to solve the other problem. I didnt take offense, sorry if it
seemed like it. I just dont think reworking the docs so we could have one
section on lookbehind is worth your or my time. I do think that improving
how podchecker is worth your or my time. It's on my list, but I'm in the
middle of other tasks already.

Yves

demerphq:
On Tue, 1 Mar 2022 at 05:15, Karl Williamson ***@***.***>
wrote:

> On 2/28/22 06:36, Yves Orton wrote:
> > On Mon, 28 Feb 2022, 21:15 Hugo van der Sanden, ***@***.***>
> > wrote:
> >
> > > The cause is the two L<> references to the docs for positive and
> negative
> > > lookbehind I added to the perldiag entries.
> > >
> > > Why have two diagnostics rather than one? When I suggested earlier that
> > > including negative lookbehind too would "also make both the
> > implementation
> > > and the explanation a bit simpler", it was in part with the
> > assumption that
> > > things like warning messages would simply lose "positive" from their
> text
> > >
> > Because they are different constructs with different issues, and rather
> > different example cases and explanations. As a user I want more detailed
> > tailored errors not generic ones. Anyway, lookbehind isn't documented as
> a
> > single thing with two flavors but rather as a set of two different
> > constructs each of which can be spelled three different ways. Which
> would I
> > link to? Both? Rewrite the docs totally? Why not DTRT instead?
> >
> > I don't see any real win with this simpler implementation you propose,
> it's
> > a neglible drop in a large ocean. I'd rather have the tiny added
> > complexity, which is really nothing at all in the grand scheme of things,
> > if it gives a more useful user experience. We should be optimizing for
> our
> > users not us as core devs with this kind of thing.
> >
> > Also it wouldn't make a difference to the subject of this subthread,
> there
> > would still be a bogus report of a broken link, or two, podchecker is
> just
> > broken in how it handles X<> index entries and L<> to them. The code,
> docs
> > and diags are written, if you want to spend energy on this subject I
> think
> > it would better placed fixing the pod checker if I may be so bold to say,
> > it would be of far greater benefit to the project as a whole than undoing
> > the work I already have done.
> >
>
> It appears you think that X<> entries are what gets linked to. That is
> incorrect, and explains why you couldn't get it to work.
>
> X<> is used for indexing, not linking. As the docs say:
>

Yes, I assumed that something that is marked as an index entry can be
linked to. I guess you are saying it doesn't work like that. I am not sure
how I could have known that however.


>
> X<topic name>" -- an index entry
> This is ignored by most formatters, but some may use it for
> building indexes. It always renders as empty-string.
>
> Maybe buildtoc uses them; I've never looked into it. That would be the
> only place I can think of that uses them in the Perl core.
>
> L<> is used for linking. The target must be a "section" in the linked-to
> page. A "section" is introduced by =head or =item.


I had formed the impression that X<> entries were  a way to give
alternative names to the sections. I now understand I was mistaken. Thanks
for explaining that, was driving me bonkers.


>
> perlpodspec says
>
> Note that section names might contain markup. I.e., if a section
> starts with:
>
> =head2 About the C<-M> Operator
>
> or with:
>
> =item About the C<-M> Operator
>
> then a link to it would look like this:
>
> L<somedoc/About the C<-M> Operator>
>
> The podchecker is correctly doing its job here.
>
> If it were me, I would make the link
> L<perlre/Lookaround Assertions>
>
> and make them start reading/skimming there.
>

Thanks for the guidance, ill take a look. I would prefer to link the
correct section instead.

cheers.
Yves

demerphq:
On Tue, 1 Mar 2022 at 08:20, demerphq ***@***.***> wrote:
> Thanks for the guidance, ill take a look. I would prefer to link the correct section instead.

Fixed and repushed.

Cheers,
Yves
-- 

demerphq:
On Sun, 27 Feb 2022 at 14:10, demerphq ***@***.***> wrote:
>
> On Sun, 27 Feb 2022 at 14:06, Hugo van der Sanden ***@***.***> wrote:
>>
>> I mean something like this:
>>
>> # named capture correctly triggers warning: ok
>> % ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<d>y)|zz) }x'
>> Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
>>   in m/ (?=(?<c>x)) | (?<=(?<d>y)|zz)  <-- HERE / at -e line 1.
>> %
>>
>> # .. even if reused: ok
>> % ./miniperl -we 'qr{ (?=(?<c>x)) | (?<=(?<c>y)|zz) }x'
>> Variable length positive lookbehind with capturing is experimental in regex; marked by <-- HERE \
>>   in m/ (?=(?<c>x)) | (?<=(?<c>y)|zz)  <-- HERE / at -e line 1.
>> %

So named captures are just captures, so no need for special tests there.

>>
>> # reuse of numbered capture via (?| ...) not ok
>> % ./miniperl -we 'qr{ (?| (?=(x)) | (?<=(y)|zz) ) }x'
>> %
>>
>> That last case needs looking at: it should warn too
>
> Gnash. This is like the 12th time that stupid thing has come back to bite me. If we had had experimental back in the day it have been marked experimental, and it would still be marked experimental it has caused so much trouble.
>
> Thanks for bringing this to my attention. That *deeply* sucks.

I pushed fixes for this today.  And a bunch of tests for the
experimental warnings that were missing. So this should be fixed now.
Also fixed a couple of other things that were in the way to getting
this fixed.

cheers,
Yves
Ps: Thanks for the reminder. There are a bunch of other branch-reset
related bugs and this reminded me to address them so Ill be doing more
work on this subject in the next days.

hvds:
> I pushed fixes for this today. And a bunch of tests for the experimental warnings that were missing. So this should be fixed now. Also fixed a couple of other things that were in the way to getting this fixed.

LGTM; I commented with a couple of minor opportunities for improvement, but I'm happy enough with it as is.

t/porting/known_pod_issues.dat
demerphq: Fixed now.

pod/perlre.pod
demerphq: fixed.

demerphq:
I totally reworked the patch sequence so that all of the ancillary changes are self contained, and the changes to actually reduce the scope of the warning are in a single self contained last patch. I also added comments as suggested by hugo. I tested each commit one at a time with 'make test_reonly and they all pass. About the only thing I haven't done is spell checked the new comments and commit messages. I will do that later and maybe repush if necessary. I will also prepare a new "remove experimental status" patch based on top of this. I expect to merge this today.

hvds:
> I assume you've thought about this, but how do we know that this op will still be around by the time we come to look at it? Is it just that we happen to know a close op is never modified by in-place optimizations? I don't know, for example, whether the TRIE replacement can ever move or replace a capture.
> 
> If there's an invariant we're relying on here, it would be useful to make sure it's noted somewhere.

Not sure why github didn't capture the line being commented on, it should probably have been in 84950019128:
```
+                *(data->last_close_opp) = scan;
```

demerphq:
On Thu, 3 Mar 2022 at 12:59, Hugo van der Sanden ***@***.***>
wrote:

> I assume you've thought about this, but how do we know that this op will
> still be around by the time we come to look at it? Is it just that we
> happen to know a close op is never modified by in-place optimizations? I
> don't know, for example, whether the TRIE replacement can ever move or
> replace a capture.
>
> If there's an invariant we're relying on here, it would be useful to make
> sure it's noted somewhere.
>
> Not sure why github didn't capture the line being commented on, it should
> probably have been in 8495001
> <https://github.com/Perl/perl5/commit/8495001912886fcd4f9ecece064aae6f78bcbebd>
> :
>
> +                *(data->last_close_opp) = scan;
>
> Yeah, it's a pointer to a pointer to a regnode. The first pointer is
"safe" and we dereference it to look at the scan pointer, which as you
pointed out is not.

Yves

-- 

hvds:
The rest of the branch at dd7fec2322 looks good to me.

demerphq:
Ok, merging, Ill do some follow ups after. This really cleans up the code for the stuff im working on.

PR #19456, opened 2022-02-23
Update perldebug.pod to explain detection of running under the debugger

Ovid:
Perl discussion on P5P, this updates`pod/perldebug.pod` to explain using the `$^P` variable to detect if you're running under the debugger.

### DIFF
--- a/pod/perldebug.pod
+++ b/pod/perldebug.pod
@@ -92,7 +92,7 @@ debugger:
         # running under the debugger
     }
 
-See C<perldoc -v '$^P'> for more information on the variable.
+See L<perlvar/$^P> for more information on the variable.
 
 =head2 Debugger Commands
 
--- a/pod/perldebug.pod
+++ b/pod/perldebug.pod
@@ -84,6 +84,16 @@ Debug a given program using threads (experimental).
 
 =back
 
+If Perl is called with the C<-d> switch, the variable C<$^P> will hold a true
+value. This is useful if you need to know if your code is running under the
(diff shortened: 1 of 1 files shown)


pod/perldebug.pod
```diff
@@ -84,6 +84,16 @@ Debug a given program using threads (experimental).
...
 
+If Perl is called with the C<-d> switch, the variable C<$^P> will hold a true
+value. This is useful if you need to know if your code is running under the
+debugger:
+
+    if ( $^P ) {
+        # running under the debugger
+    }
+
+See C<perldoc -v '$^P'> for more information on the variable.
```
ilmari: This would be better as a link, `L<perlvar/$^P>`.

Ovid: Fixed, thank you.

Ovid:
Tests failed with:

```
porting/args_assert.t ................................................ ok
Ovid  <ovid@cpan.org> not found in AUTHORS
```

I feel awkward just adding my own name there (seems presumptuous). Should I do that, or is there some automated process which handles that?

pod/perldebug.pod
khwilliamson: Please squash your commits.  It would be fine for you to update AUTHORS< though in this case it is more complicated because you have an already existing different email address there.  I will take care of that

khwilliamson:
Never mind, I just squashed and pushed as 2b031f68cbd94fd2378a0554999316c603b81ea5

PR #19459, opened 2022-02-24
Fix regex sets pod to handle more /flags, and update pod

khwilliamson:
The pod didn't reflect restrictions on what can go into such a class.

### DIFF
--- a/pod/perlrecharclass.pod
+++ b/pod/perlrecharclass.pod
@@ -1181,46 +1181,29 @@ closing C<])> characters.
 
 Just as in all regular expressions, the pattern can be built up by
 including variables that are interpolated at regex compilation time.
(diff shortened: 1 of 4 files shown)


jkeenan:
> The pod didn't reflect restrictions on what can go into such a class.

Can you point us to places in the test suite which exercise the code whose documentation needs adjusting?

khwilliamson:
On 2/24/22 13:34, James E Keenan wrote:
>     The pod didn't reflect restrictions on what can go into such a class.
> 
> Can you point us to places in the test suite which exercise the code 
> whose documentation needs adjusting?
> 

t/re /reg_mesg.t
      /regex_sets.t
      /regex_sets_compat.t

PR #19461, opened 2022-02-25
fix __CALLER__ handling in Attribute::Handlers

haarg:
Attribute::Handlers supports a __CALLER__ token when declaring autotie
attributes. This is meant to create the attribute in the caller of the
class it is used in. The only reliable way for this to work requires
creating an import method in the calling class. Instead,
Attribute::Handlers was trying to walk up the call stack from when its
own import would be called. This used to partially work, at least enough
to be deceptive. Checking the caller deeper in the call stack of
Attribute::Handlers would allow the __CALLER__ attribute to work only
for the first time the module using it was called. Any future users
would not re-compile the module, so they would not re-invoke
Attribute::Handlers' import method, and would not get the autotie
attribute defined.

This attempt to find the caller's caller also started failing as of
f6387cff9cb31db4cf18c8641917ea4639ac2b65.

Fix the handling of __CALLER__ by creating an import method in the
caller if it is used, so that users of the calling module will reliably
get the attribute defined.


demerphq:
Thanks, merged.

Issue #19463, opened 2022-02-26
-DXvt crashes in strict.pm


aquanight:
**Description**
Under -DDEBUGGING, the debug flags Xvt (-DXvt or $^D = "Xvt" at runtime) crashes when strict.pm is loaded via require/use.

**Steps to Reproduce**
Perl must be built with -DDEBUGGING
perl -DXvt -Mstrict -e 'print "Hello\n"'

**Expected behavior**
The requested debug output (Xvt traces execution and dumps PADs and reports when CVs are cloned) alongside executed script behavior.

**Actual results**
The last few lines of output are as such:
```
(15244:/usr/share/perl/5.32/strict.pm:28)       nextstate
(15244:/usr/share/perl/5.32/strict.pm:30)       padsv($bits)
(15244:/usr/share/perl/5.32/strict.pm:30)       padsv($inline_all_bits)
(15244:/usr/share/perl/5.32/strict.pm:30)       sassign
(15244:/usr/share/perl/5.32/strict.pm:30)       nextstate
(15244:/usr/share/perl/5.32/strict.pm:31)       anoncode
Pad 0x558289c25a40[0x558289c3fb18] sv:      17 sv=0x558289c25e48
Pad 0x558289c25fe0[0x558289c37438] new:       compcv=0x558289c25fc8 name=0x558289c391c8 flags=0x3
CV undef: cv=0x558289c25fc8 comppad=0x558289c25a40
Pad set_null
Pad undef: cv=0x558289c25fc8 padlist=0x558289c3cc38 comppad=0x558289c25a40

Pad CV clone
  Proto: CV=0x558289c25e48 (ANON), OUTSIDE=0x558289c25a28 (UNIQUE)
    PADLIST = 0x558289c37288
    PADNAME = 0x558289c391c8(0x558289c3c358) PAD = 0x558289c25e60(0x558289c38f08)
         1. 0x0<0> FAKE "$inline_all_bits" flags=0x0 index=16
debugperl: pad.c:1855: S_cv_dump: Assertion `!(((XPVCV*)({ void *_p = (((CV*)(cv))->sv_any); _p; }))->xcv_flags & 0x0008)' failed.
Aborted (core dumped)
```
The line 31 from strict.pm in question is this:
```
*all_bits = sub () { $inline_all_bits };
```
This is inside a BEGIN {} block so it is occurring as soon as strict.pm is require()d (including via 'use').

```
[...]

PR #19468, opened 2022-02-27
Add example of two-stage bisection

jkeenan:
Based on work in https://github.com/Perl/perl5/issues/19463

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -933,7 +933,8 @@ L<GH issue 17333|https://github.com/Perl/perl5/issues/17333>
 
 =item * Problem
 
-In C<-DDEBUGGING> builds, the debug flags C<Xvt> would crash a program when F<strict.pm> was loaded via C<require> or C<use>.
+In C<-DDEBUGGING> builds, the debug flags C<Xvt> would crash a program when
+F<strict.pm> was loaded via C<require> or C<use>.
 
 =item * Solution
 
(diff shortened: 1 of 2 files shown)


jkeenan:
The p.r. documents what invocations were used in each stage of the bisection, but I feel it could use some improvement as to *why we needed a two-stage process*.

demerphq (approved):
You could argue this is a *three* stage bisect, as stage 0 would have been Hugo finding the bracketing versions for it, presumably manually, and then bisecting to the find the commit. 

But if you were to argue that stage 0 wasn't true bisection I wouldn't disagree either. :-)

It does make me think that maybe bisect.pl could learn a new trick. :-)

jkeenan:
> You could argue this is a _three_ stage bisect, as stage 0 would have been Hugo finding the bracketing versions for it, presumably manually, and then bisecting to the find the commit.
> 

In the **EXAMPLES** section we already have one instance where we mention that the investigator used pre-built production releases to narrow the scope over which `Porting/bisect.pl` was asked to iterate.  By that precedent *this* instance counts as two stages (two invocations of the program) rather than three.

hvds (approved):

Porting/bisect-runner.pl
```diff
@@ -927,6 +927,60 @@ =head2 Point in time where code started to segfault is unknown
...
+
+=item * Problem
+
+In C<-DDEBUGGING> builds, the debug flags C<Xvt> would crash a program when F<strict.pm> was loaded via C<require> or C<use>.
+
+=item * Solution
+
+Two-stage solution.  In each stage, to shorten debugging time investigator
+made use of existing set of production releases of F<perl> built with
+C<-DDEBUGGING>.
```
hvds, line 943: For the additional explanation you asked about you could add here something like:

> The first bisect determined the point at which code was introduced to F<strict.pm> that triggered the problem. With an understanding of the trigger, the second bisect then determined at what point such a trigger started causing a crash.

PR #19471, opened 2022-02-28
Make sv_len_utf8_nomg() public

khwilliamson:
The get-magic version of this has long been available, but not the no magic.
It turns out that ppport.h does make both available, so any module
using that would be able to use it anyway, without even noticing that it
isn't supported by core.

I see no reason for this to not be public; the original commit message
did not indicate such a reason, and it is effectively public anyway via
ppport.h.

(This is the only non-publically accessible function that ppport.h
furnishes; I'm working on a patch to have it warn of future instances
creeping in.)


demerphq:
Thanks, I will merge.

Issue #19474, opened 2022-03-01
BBC: Blead Breaks MIROD/Params-Style-0.06.tar.gz


andk:
**Description**

The distribution's test suite started to fail with perl 5.35.9.

Bisect leads to v5.35.8-105-gf6387cff9c (https://github.com/perl/perl5/commit/f6387cf)

Sample fail report: http://www.cpantesters.org/cpan/report/555f4d58-97ca-11ec-9b3a-d44178002d2e

Matrix: http://matrix.cpantesters.org/?dist=Params-Style+0.06

**Steps to Reproduce**

cpan -i MIROD/Params-Style-0.06.tar.gz

**Expected behavior**

Should successfully test and install the module

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: 64f0d0d96084c25955864f7be0672b173a25fb1e
  Platform:
    osname=linux
    osvers=5.15.0-2-amd64
    archname=x86_64-linux-thread-multi-ld
    uname='linux k93msid 5.15.0-2-amd64 #1 smp debian 5.15.5-2 (2021-12-18) x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.9-20-g64f0d0d960/ac75 -Dmyhostname=k93msid -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Duseithreads -Duselongdouble -DEBUGGING=both'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=define
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2 -g'
[...]

PR #19476, opened 2022-03-01
wrap any existing import in Attribute::Handlers' injected import

haarg:
Some modules using Attribute::Handlers autotie feature have their own
import method. When injecting an import method, attempt to wrap anything
that exists, either in the class directly or in a parent class.

We can't detect parent classes at injection time, because @ISA may
change, and this is actually common since users of Exporter will add it
as a parent class at module runtime, while Attribute::Handlers injection
would happen first during compilation.

A better approach could involve using next::can from mro, but that also
requires Sub::Util::set_subname, and Attribute::Handlers is dual life
and currently supports 5.6.

### DIFF
--- a/dist/Attribute-Handlers/lib/Attribute/Handlers.pm
+++ b/dist/Attribute-Handlers/lib/Attribute/Handlers.pm
@@ -88,13 +88,26 @@ sub import {
 
                 if ($attr =~ /\A__CALLER__::/) {
                     no strict 'refs';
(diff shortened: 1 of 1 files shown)


haarg:
This should be a fix for #19474.

I'm not a big fan of this, so alternative approaches would be welcome. But this module is a horrible mess already, so I don't feel too bad about adding more ugliness.

demerphq:
Thanks, passes test, ill merge.

PR #19477, opened 2022-03-01
improved .mailmap and AUTHORS file logic

demerphq:
This change updates AUTHORS with multiple missing contributors to the
project. It updates .mailmap so that the preferred email for each
developer matches what is in AUTHORS (sometimes changing the AUTHORS
entry, sometimes changes the .mailmap entry or entries to achieve this
goal. It also adds a new tool, Porting/UpdateAuthors.pl which can be
used to automatically update the relevant data.

The general idea is that if you run Porting/UpdateAuthors.pl any never
developers will be added to AUTHORS, and new entries will be added to
the .mailmap file. The aim is that the only data we need to have to
manage both files is the data in both files and the commit history
itself. There is NOT embedded special datasets in the tool.

The idea is that *every* developer, be it author or committer has at
least one entry in .mailmap, and that the vast majority of the email
data we have for commits is mapped to corresponding entry in the AUTHORS
file. Furthermore the .mailmap file is expected to have an entry for
every distinct email ever used by a developer on the project.

If you run Porting/UpdateAuthors.pl the tool will analyze all the
commits in the projects history, and update either or both the .mailmap
and AUTHORS file appropriately. If an existing developer adds a commit
which either has the same name or email as is listed already then it
will DTRT and update the .mailmap with the appropriate mappings. If a
new developer adds a commit then it will DTRT and update both the
.mailmap and the AUTHORS file with that developers details.

Every single existing distinct name/email combination has been added
to the mailmap, and matched to AUTHORS data. In some case where there
was conflicting or missing data I used an educated guess, contacted
the developer myself, or verified the email with the email provider
(eg Google).

wip

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5479,6 +5479,7 @@ Porting/sync-with-cpan		Sync with CPAN
 Porting/timecheck.c		Test program for the 2038 fix
 Porting/timecheck2.c		Test program for the 2038 fix
 Porting/todo.pod		Perl things to do
+Porting/updateAUTHORS.pl	Tool to automatically update AUTHORS and .mailmap from git log data
 Porting/valgrindpp.pl		Summarize valgrind reports
 Porting/vote_admin_guide.pod	Perlgov Vote Administrator guide
 pp.c				Push/Pop code
--- a/Porting/README.pod
+++ b/Porting/README.pod
@@ -44,7 +44,9 @@ Check source code for ANSI-C violations.
 
 =head2 F<checkAUTHORS.pl>
 
-Used by F<t/porting/authors.t> to ensure the F<AUTHORS> list is up to date.
+Used by F<t/porting/authors.t> to ensure the F<AUTHORS> list is up to
+date. See also L<< /"F<updateAUTHORS.pl>" >> for a way to automatically
+fix issues found by this tool.
 
 =head2 F<checkcfguse.pl>
 
@@ -380,6 +382,13 @@ The tasks we think are smaller or easier are listed first.  Anyone is welcome
 to work on any of these, but it's a good idea to first contact
 I<perl5-porters@perl.org> to avoid duplication of effort, and to learn from
 any previous attempts.
+
+=head2 F<updateAUTHORS.pl>
+
+This script will automatically update AUTHORS and create .mailmap entries
+based on the git commit log history. If F<checkAUTHORS.pl> complains
+during testing you should run this. It will automatically fix most if not
+all AUTHORS related test fails.
 
 =head2 F<valgrindpp.pl>
 
--- a/Porting/exec-bit.txt
+++ b/Porting/exec-bit.txt
@@ -66,6 +66,7 @@ Porting/newtests-perldelta.pl
 Porting/perlhist_calculate.pl
 Porting/sort_perldiag.pl
 Porting/sync-with-cpan
+Porting/updateAUTHORS.pl
 Porting/valgrindpp.pl
 Cross/generate_config_sh
 Cross/warp
--- a/Porting/updateAUTHORS.pl
+++ b/Porting/updateAUTHORS.pl
@@ -1,0 +1,659 @@
+#!/usr/bin/env perl
+package Porting::updateAUTHORS;
+use strict;
+use warnings;
+use Getopt::Long qw(GetOptions);
+use Pod::Usage qw(pod2usage);
+use Data::Dumper;
+use Encode qw(encode_utf8 decode_utf8 decode);
+
+# The style of this file is determined by:
+#
+# perltidy -w -ple -bbb -bbc -bbs -nolq -l=80 -noll -nola -nwls='=' \
+#   -isbc -nolc -otr -kis -ci=4 -se -sot -sct -nsbl -pt=2 -fs  \
+#   -fsb='#start-no-tidy' -fse='#end-no-tidy'
+
+# Info and config for passing to git log.
+#   %an: author name
+#   %aN: author name (respecting .mailmap, see git-shortlog(1) or git-blame(1))
+#   %ae: author email
+#   %aE: author email (respecting .mailmap, see git-shortlog(1) or git-blame(1))
+#   %cn: committer name
+#   %cN: committer name (respecting .mailmap, see git-shortlog(1) or git-blame(1))
+#   %ce: committer email
+#   %cE: committer email (respecting .mailmap, see git-shortlog(1) or git-blame(1))
+#   %H: commit hash
+#   %h: abbreviated commit hash
+#   %s: subject
+#   %x00: print a byte from a hex code
+
+my %field_spec= (
+    "an" => "author_name",
+    "aN" => "author_name_mm",
+    "ae" => "author_email",
+    "aE" => "author_email_mm",
+    "cn" => "committer_name",
+    "cN" => "committer_name_mm",
+    "ce" => "committer_email",
+    "cE" => "committer_email_mm",
+    "H"  => "commit_hash",
+    "h"  => "abbrev_hash",
+    "s"  => "commit_subject",
+);
+
+my @field_codes= sort keys %field_spec;
+my @field_names= map { $field_spec{$_} } @field_codes;
+my $tformat= join "%x00", map { "%" . $_ } @field_codes;
+
+sub _make_name_author_info {
+    my ($author_info, $commit_info, $name_key)= @_;
+    (my $email_key= $name_key) =~ s/name/email/;
+    my $email= $commit_info->{$email_key};
+    my $name= $commit_info->{$name_key};
+
+    my $line= $author_info->{"email2line"}{$email}
+        // $author_info->{"name2line"}{$name};
+
+    $line //= sprintf "%-31s<%s>",
+        $commit_info->{$name_key}, $commit_info->{$email_key};
+    return $line;
+}
+
+sub _make_name_simple {
+    my ($commit_info, $key)= @_;
+    my $name_key= $key . "_name";
+    my $email_key= $key . "_email";
+    return sprintf "%s <%s>", $commit_info->{$name_key},
+        lc($commit_info->{$email_key});
+}
+
+sub read_commit_log {
+    my ($author_info, $mailmap_info)= @_;
+    $author_info ||= {};
+    open my $fh, qq(git log --pretty='tformat:$tformat' |);
+
+    while (defined(my $line= <$fh>)) {
+        chomp $line;
+        $line= decode_utf8($line);
+        my $commit_info= {};
+        @{$commit_info}{@field_names}= split /\0/, $line, 0 + @field_names;
+
+        my $author_name_mm= _make_name_author_info($author_info, $commit_info,
+            "author_name_mm");
+
+        my $committer_name_mm=
+            _make_name_author_info($author_info, $commit_info,
+            "committer_name_mm");
+
+        my $author_name_real= _make_name_simple($commit_info, "author");
+
+        my $committer_name_real= _make_name_simple($commit_info, "committer");
+
+        _check_name_mailmap(
+            $mailmap_info, $author_name_mm, $author_name_real,
+            $commit_info,  "author name"
+        );
+        _check_name_mailmap($mailmap_info, $committer_name_mm,
+            $committer_name_real, $commit_info, "committer name");
+
+        $author_info->{"lines"}{$author_name_mm}++;
+        $author_info->{"lines"}{$committer_name_mm}++;
+    }
+    return $author_info;
+}
+
+sub read_authors {
+    my ($authors_file)= @_;
+    $authors_file ||= "AUTHORS";
+
+    my @authors_preamble;
+    open my $in_fh, "<", $authors_file
+        or die "Failed to open for read '$authors_file': $!";
+    while (defined(my $line= <$in_fh>)) {
+        chomp $line;
+        push @authors_preamble, $line;
+        if ($line =~ /^--/) {
+            last;
+        }
+    }
+    my %author_info;
+    while (defined(my $line= <$in_fh>)) {
+        chomp $line;
+        $line= decode_utf8($line);
+        my ($name, $email);
+        my $copy= $line;
+        $copy =~ s/\s+\z//;
+        if ($copy =~ s/<([^<>]*)>//) {
+            $email= $1;
+        }
+        elsif ($copy =~ s/\s+(\@\w+)\z//) {
+            $email= $1;
+        }
+        $copy =~ s/\s+\z//;
+        $name= $copy;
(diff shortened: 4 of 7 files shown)


haarg:
This seems to result in a lot of overly specific entries in .mailmap that shouldn't normally be required by git. For example,
```
Chip Salzenberg <chip@pobox.com> Chip <chip@pobox.com>
Chip Salzenberg <chip@pobox.com> chip <chip@pobox.com>
Chip Salzenberg <chip@pobox.com> Chip Salzenberg <chip@pobox.com>
```
These could all be replaced by a single
```
Chip Salzenberg <chip@pobox.com> <chip@pobox.com>
```
Is there some reason why the extra entries are needed?

demerphq:
On Tue, 1 Mar 2022 at 16:50, Graham Knop ***@***.***> wrote:

> This seems to result in a lot of overly specific entries in .mailmap that
> shouldn't normally be required by git. For example,
>
> Chip Salzenberg ***@***.***> Chip ***@***.***>
> Chip Salzenberg ***@***.***> chip ***@***.***>
> Chip Salzenberg ***@***.***> Chip Salzenberg ***@***.***>
>
> These could all be replaced by a single
>
> Chip Salzenberg ***@***.***> ***@***.***>
>
> Is there some reason why the extra entries are needed?
>
Yeah, it made managing the file and the auto update process much easier. I
dont think we should care about that. It was a deliberate design decision
to make it easier to manage. See the email I just posted for more details.

But in simple terms I did this:

git log --pretty="format:%aN <%aE> %an <%ae>%n%cN <%cE> %cn <%ce>" | sort |
uniq >> .mailmap

and then cleaned up the mess. I tried simplifying but it just made life
more complicated. For instance, how do I know its "safe" to reduce to jus
an email on the right with an entry like this:

Ævar Arnfjörð Bjarmason ***@***.***> Ævar Arnfjörð Bjarmason) (via RT <
***@***.***>

In this case we definitely dont want to map every commit that is marked as
coming from ***@***.*** to avars name.  And there are several
such cases. So I would have had to maintain a secondary list of exceptions.
The intent here was that there would be no side-band data involved.
Everything needed is in .mailmap, or AUTHORS. No secondary set of data
should be involved. That includes whitelists of emails that require a name.

cheers,
Yves

demerphq:
On Tue, 1 Mar 2022 at 16:59, demerphq ***@***.***> wrote:
>
> On Tue, 1 Mar 2022 at 16:50, Graham Knop ***@***.***> wrote:
>>
>> This seems to result in a lot of overly specific entries in .mailmap that shouldn't normally be required by git. For example,
>>
>> Chip Salzenberg ***@***.***> Chip ***@***.***>
>> Chip Salzenberg ***@***.***> chip ***@***.***>
>> Chip Salzenberg ***@***.***> Chip Salzenberg ***@***.***>
>>
>> These could all be replaced by a single
>>
>> Chip Salzenberg ***@***.***> ***@***.***>
>>
>> Is there some reason why the extra entries are needed?
>
> Yeah, it made managing the file and the auto update process much easier. I dont think we should care about that. It was a deliberate design decision to make it easier to manage. See the email I just posted for more details.
>
> But in simple terms I did this:
>
> git log --pretty="format:%aN <%aE> %an <%ae>%n%cN <%cE> %cn <%ce>" | sort | uniq >> .mailmap
>
> and then cleaned up the mess. I tried simplifying but it just made life more complicated. For instance, how do I know its "safe" to reduce to jus an email on the right with an entry like this:
>
> Ævar Arnfjörð Bjarmason ***@***.***> Ævar Arnfjörð Bjarmason) (via RT ***@***.***>
>
> In this case we definitely dont want to map every commit that is marked as coming from ***@***.*** to avars name.  And there are several such cases. So I would have had to maintain a secondary list of exceptions. The intent here was that there would be no side-band data involved. Everything needed is in .mailmap, or AUTHORS. No secondary set of data should be involved. That includes whitelists of emails that require a name.

After chatting with Graham in IRC I thought id add a point to my
response to his question. The tool automanages BOTH the .mailmap file
AND the AUTHORS. So nobody should have to edit either in 99% of the
cases. Having the .mailmap file be explicit most of the time means
that the logic is much simpler, and I dont see why we should care
about some extra lines in the mailmap file. In general the AUTHORS
file is "leading". For example if you remove an email from someones
name in authors and then rerun the script it should update the
preferred emails in the .mailmap for all that persons entires to be
"<unknown>".

Yves




-- 

MANIFEST
```diff
@@ -5479,6 +5479,7 @@ Porting/sync-with-cpan		Sync with CPAN
 Porting/timecheck.c		Test program for the 2038 fix
 Porting/timecheck2.c		Test program for the 2038 fix
 Porting/todo.pod		Perl things to do
+Porting/updateAUTHORS.pl	Tool to automatically update AUTHORS and .mailmap from git log data
```
atoomic, line 5482: do we need to ship this tool? 
Would it be used outside the GitHub repo?
note that `.mailmap` is not even listed in the MANIFEST

Porting/README.pod
```diff
@@ -381,6 +383,13 @@ to work on any of these, but it's a good idea to first contact
 I<perl5-porters@perl.org> to avoid duplication of effort, and to learn from
 any previous attempts.
 
+=head2 F<updateAUTHORS.pl>
+
+This script will automatically update AUTHORS and create .mailmap entries
+based on the git commit log history. If F<checkAUTHORS.pl> complains
```
atoomic, line 389: as stated here this cannot work outside a `git repo` no?

Porting/exec-bit.txt
```diff
@@ -3,69 +3,70 @@
 # 
 # List files (or glob patterns) one per line relative to the repo root
 # Lines starting with a pound and empty lines will be ignored
-Configure
-Makefile.SH
-Policy_sh.SH
```
atoomic: why so much noise in this file? seems like we are using a smarter sort :-)

Porting/updateAUTHORS.pl
```diff
@@ -0,0 +1,639 @@
+#!/usr/bin/env perl
+package Porting::updateAUTHORS;
+use strict;
+use warnings;
```
atoomic, line 4: let's be crazy and consider using function signature too :-)

Porting/updateAUTHORS.pl
```diff
@@ -0,0 +1,639 @@
...
+    my $line = $author_info->{"email2line"}{$email}
+            // $author_info->{"name2line"}{$name};
+
+    $line //= sprintf "%-31s<%s>",
+        $commit_info->{$name_key}, $commit_info->{$email_key};
+    return $line;
+}
+
+sub _make_name_simple {
+    my ($commit_info, $key)= @_;
```
atoomic, line 63: consider tidy the file?

atoomic (approved):
LGTM I would prefer we do not ship the devel tools only used during development
I understand they were added as they are part of the doc, but I doubt they would work outside a git repo and having no `.mailmap` ship makes the tool useless IMO outside devel context

MANIFEST
demerphq: "do we need to ship this tool"

Probably not, but that applies to a number of things in Porting and IMO is orthogonal to this PR.

"Would it be used outside the GitHub repo?"
No, but this point would also apply to Porting/checkAUTHORS.pl as well wouldn't it? Nit: its a git repo, github just hosts the repo. :-)

Maybe we should add a new directory that is excluded from tarballs. But I can also imagine that is a can of worms. It certainly doesn't cause any *harm* shipping them.

Also tools in Porting aren't installed or anything, but they are included in the build tarball, so I think its like shipping any of our build tools. We have many tools in there that would be useless without git, eg, Porting/bisect.

"note that .mailmap is not even listed in the MANIFEST"
That suggests there might be a bug in the MANIFEST tests and an oversight I guess. But also I think something that should be addressed in a separate issue/PR.

Porting/README.pod
demerphq: Yes. Same as checkAUTHORS.pl itself. My assumption is that checkAUTHORS.pl doesn't run if its not in a git repo, so it wouldn't complain and people wouldn't be directed to use updateAUTHORS.pl :-)

Porting/exec-bit.txt
demerphq: I just used "sort" from the shell. I can remove this patch and do it in a separate PR if you prefer. But in vim its really easy to just call sort on a list. :-)

Porting/updateAUTHORS.pl
demerphq: I don't really like function signatures and I don't use them in code I write myself. I don't think we need to use every feature in perl just because they are there. Also I would like this script to "just work" with whatever perl happens to be installed. There is a chicken and egg problem with some of our Perl based build tooling, sometimes people want to run it from older perls than the perl they are building, and I think those tools should just work if possible.

Porting/updateAUTHORS.pl
demerphq: Is there something specific you object to? I left hug my assignment operators deliberately, and as long as there is no standard format for our scripts and this is a new script I don't see why this script shouldn't use my style. The rule is "new scripts Author chooses the style, otherwise follow the style of the file".  I can run perltidy with my favorite options and include the perltidy specifier in the file however if you like, but AFAIK it won't change that line of code. :-)

I raised the idea of having a standard perltidy format as in the name of consistency I would be willing to give up my style eccentricities if others give up theirs, but it was not well received.

karenetheridge:
Is this script intended to be run after every commit, or just at release time? I thought the entire reason why we started discussing the checkAUTHORS.pl script was that having Porting tests fail when a new contributor sent a PR caused unnecessary friction, which led us to discussing that we might as well just generate a new authors file at release time based on information we already had in git.

> The idea is that every developer, be it author or committer has at least one entry in .mailmap

Why is this needed? .mailmap is for indicating that two email addresses (or names) are actually one person. It should not be necessary to map something to itself, when that entry is already in git.

Porting/exec-bit.txt
demerphq: I dropped this commit. Ill do it as its own pr.

Porting/updateAUTHORS.pl
demerphq: I ran perltidy on it, and included my perltidy options in the script. I couldn't decide what was best tho, putting it in the pod so it can be easily C&P executed, or putting it in a comment. For now I put it in a comment at the top of the file. Thanks for the feedback, as much as I like my style the file wasn't consistent, and now it is. Thanks for the feedback!

demerphq:
On Wed, 9 Mar 2022 at 04:33, Karen Etheridge ***@***.***>
wrote:

> Is this script intended to be run after every commit, or just at release
> time? I thought the entire reason why we started discussing the
> checkAUTHORS.pl script was that having Porting tests fail when a new
> contributor sent a PR caused unnecessary friction, which led us to
> discussing that we might as well just generate a new authors file at
> release time based on information we already had in git.
>
Yeah, but that doesn't make as much sense as it sounds because humans are
messy and email addresses and names are under human control. You can use
the --author switch and put whatever you like there. After a great deal of
effort trying to do it the other way I realized It's better to have the
test fail and then have the author run the  update tool and if they are
happy with the results commit and we are done.

Eg, someone commits with --author="Muhahahah ***@***.***>",
what should we do with that?

It can ALSO be done any time someone wants to change their authoritative
data. Eg, if you change your email address in authors and run updateAUTHORS
then the mailmap file will be updated appropriately to map your old email
addresses to your new. The idea is to keep everything in sync and minimize
pain.


> The idea is that every developer, be it author or committer has at least
> one entry in .mailmap
>
> Why is this needed? .mailmap is for indicating that two email addresses
> (or names) are actually one person. It should not be necessary to map
> something to itself, when that entry is already in git.
>
I actually already answered this a couple of times already (see the reply
to graham on the emails and comments I have posted on this). But I will
recap once more.

The idea is to make the .mailmap more or less act as a binding table
between the AUTHORS file and the names actually in the commit log. The name
on the left is a foreign key into AUTHORS, the name on the right is a
foreign key into the commit log.  Essentially it says "every name/email
tuple in the commit log must map to an entry in AUTHORS" or be specifically
excluded from it by being "unknown".

There are all kinds of edge cases where the "you don't need it" blows up,
and where the code cant heuristically make the right decision. So doing it
this way ends up simplifying everything and making it obvious in the change
log. It also means that when people view the commit log on github they see
the corrected name for all commits.

You can't just think about normal cases with this problem, you need to
consider the weird edge cases which can come up because the author and
committer name in the commit log, and the data in AUTHORs are free form
unvalidated data fields totally under human control and can easily get out
of sync with each other.

Cases in point: we were missing multiple people from AUTHORS but nobody
noticed. We had multiple conflicting entries in .mailmap but nobody
noticed. We had .mailmap entries that conflicted with AUTHORs but nobody
noticed. We had many people who are in AUTHORS but had commits with weird
names/emails in the commit log without any corresponding data in the
mailmap file. So this cleans up the *whole* mess at once.

I said on #p5p when I pushed this that working on this at first sounded
like "oh that is a simple hack" and it turned out to be harder than a lot
of regex features and bugs I have worked on over the years. This patch took
a great deal more effort than it looks. Not only was the script rewritten
pretty much from scratch multiple times, but I also had to personally
contact about 10 devs and validate and research on our commit history to
get this up to speed and ensure that we knew that everything was complete.
In short if you think "why isn't this simply like 'X'" then respectfully
you haven't studied the problem properly. Please trust me when I say this
is a messy problem and this is a pretty reasonable solution which deals
with multiple issues related to this problem at once. Please also see the
other replies I have made, and the posts I have made on the subject.

cheers,
Yves

-- 

MANIFEST
demerphq: I created https://github.com/Perl/perl5/issues/19512 and https://github.com/Perl/perl5/issues/19511 to address these points.

Porting/README.pod
demerphq: See https://github.com/Perl/perl5/issues/19512

MANIFEST
atoomic: thanks, this makes sense to extend the discussion to other files too

Porting/exec-bit.txt
atoomic: no problem with it, was trying to understand what was happening and then realized it lol
fine as part of this change or as a later commit both work for me

Porting/updateAUTHORS.pl
atoomic: interesting feedback, I was mainly teasing you 
personally I love using function signature: code shorter, and helps catching some incorrect calls

Porting/updateAUTHORS.pl
atoomic: thanks for considering it, I know we do not have a good tidy policy in the whole codebase for C, Perl...

PR #19479, opened 2022-03-02
perl.c: zero stacks on creation

demerphq:
I think not initializing these to zero is a false economy, Felipe Gasper brought
this up that it made debugging issues with the stack confusing. We only do this once
per process and zeroing memory is prety cheap so it make sense to me to zero
the stacks on startup.


leonerd:
I agree it's a false economy.

LGTM

leonerd (approved):

khwilliamson:
agreed

PR #19480, opened 2022-03-02
Add macros to assert at compile time that an argument is either a pointer or a literal string

khwilliamson:


demerphq (approved):
LGTM! Good idea, I like the ASSERT_IS_PTR() trick.

PR #19487, opened 2022-03-03
regcomp.c: add comment that last_close_opp is dangerous

demerphq:
We use last_close_opp currently to track what close ops we have
encountered during study. Some of our optimizations may remove
the close op afterwards. So the "inner" pointer is not safe to
dereference. I will push a patch in the future to clean this up
somehow or another.


hvds:
Thanks, merged as be22c34e02.

Issue #19472, opened 2022-02-28
Many warnings reading empty in-memory file


jimav:
If a read-write (open mode "+>") in-memory file is created and immediately read, many warnings appear. 

**Steps to reproduce**
```
#!/usr/bin/perl
use strict; use warnings; use v5.20;

open XX, "+>", "/tmp/testfile" or die $!;
seek XX,0,0 or die $!;
my @XXbuf = <XX>;  # no warnings (as expected)
say "Read (@XXbuf) from XX";

open YY, "+>", \(my $string) or die $!;
#print YY "stuff";
seek YY,0,0 or die $!;
my @YYbuf = <YY>;  # ******* MANY "Use of uninitialized value in <HANDLE>"
say "Read (@YYbuf) from YY";
```

**Expected behavior**
Expecting a silent EOF indicator, like a regular file.

P.S. truncate does not seem to work on in-memory files, but over-writing does.

**Perl configuration**
```
Summary of my perl6 (revision 5 version 30 subversion 0) configuration:
  
  Platform:
    osname=linux
    osvers=4.19.0
    archname=x86_64-linux-gnu-thread-multi
    uname='linux localhost 4.19.0 #1 smp debian 4.19.0 x86_64 gnulinux '
[...]

PR #19488, opened 2022-03-03
Fix GH Issue #19472: read warnings from open($fh,">",\(my $x))

demerphq:
We produce all kinds of warnings if someone opens a scalar reference
that is undef. Prior to this we handled write operations ok, at
least in blead, but read operations would produce a plethora of
warnings. To me this analagous to treating an undef var as hash
and trying to read from it, we autovivify the undef var to be
a hash. So in this case we should just "autovivify" the referenced
scalar to be an empty string.

Eg before this patch:

./perl -Ilib -wle'open my $fh,"+>", \(my $v); my @x=<$fh>; print 0+@x'
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
Use of uninitialized value $fh in <HANDLE> at -e line 1.
0

After it:
./perl -Ilib -wle'open my $fh,"+>", \(my $v); my @x=<$fh>; print 0+@x'
0

### DIFF
--- a/doio.c
+++ b/doio.c
@@ -639,6 +639,16 @@ Perl_do_open6(pTHX_ GV *gv, const char *oname, STRLEN len,
                 goto say_false;
             }
 #endif /* USE_STDIO */
+            if (SvROK(*svp) && !sv_isobject(*svp)) {
+                /* if they pass in a reference and its not an object
+                 * and the reference is to undef, "autovivify" it to
+                 * the empty string. See GH Issue #19472
+                 */
+                SV *sv= SvRV(*svp);
+                if (!SvOK(sv) && !SvREADONLY(sv))
+                    sv_setpvs(MUTABLE_SV(sv),"");
+            }
+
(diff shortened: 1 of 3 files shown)


Leont (approved):

Grinnz:
Seems reasonable

demerphq:
Added docs and more tests, and fixed the stupid typos from previous version, turned out I tested with the wrong command line and didn't notice yesterday. Double checked I did it right this time. :-)

demerphq:
Tweaked yet again to deal with references to read only scalars.

demerphq:
On Thu, 3 Mar 2022 at 18:17, Dan Book ***@***.***> wrote:
>
> Seems reasonable

Thanks. Anyone object to this being merged?

Yves


-- 

PR #19490, opened 2022-03-04
Update Devel-PPPort to match last 3.67 release

atoomic:


jkeenan:
Question: At the risk of being pedantic ... since Devel-PPPort is found under `dist/`, shouldn't that mean that it should be changed in blead first, and only then get a CPAN release?

Or does the special nature of this library warrant a different approach?

atoomic:
Merging to blead does not provide the benefit of testing all flavors of perl….
Devel-PPPort is designed to work on older versions of Perl… and I think this is kind of an exception how we are developing  it.
Now I agree I could have probably merged it to blead before release but this could require some extra when bumping the release tag.

jkeenan:
> Merging to blead does not provide the benefit of testing all flavors of perl…. Devel-PPPort is designed to work on older versions of Perl… and I think this is kind of an exception how we are developing it. Now I agree I could have probably merged it to blead before release but this could require some extra when bumping the release tag.

Thanks; seems reasonable.  Can you look at these CPANtesters FAILs?
http://matrix.cpantesters.org/?dist=Devel-PPPort%203.66;perl=5.35.9;reports=1

khwilliamson:
This should wait until we get a version that passes smoke tests

atoomic:
@jkeenan indeed these issues are reported to https://github.com/Dual-Life/Devel-PPPort/issues/219
I m investigating them this only happen with the last development releases.
Strangely this does not fail when the tests are run from blead, but fail the second you use blead to run the bistro test suite...

atoomic:
Also note that these reported issues are not specific to the release of 3.66 
I can reproduce the same issue with 3.62 which is a ~2 yo release.

Going to bisect, something recently changed in blead is triggering this issue.

atoomic:
The failures are coming from this change https://github.com/Perl/perl5/commit/a460925186154b270b7a647a4f30b2f01fd97c4b
which alter `Perl_utf8n_to_uvchr_msgs` which is used by `Perl_utf8_to_uvchr_buf_helper` for `Perl_utf8_to_uvchr`
 The call `utf8_to_uvchr_buf('A', -1)` is now behaving differently and do not throw warning as earlier

jkeenan:
> The failures are coming from this change [a460925](https://github.com/Perl/perl5/commit/a460925186154b270b7a647a4f30b2f01fd97c4b) which alter `Perl_utf8n_to_uvchr_msgs` which is used by `Perl_utf8_to_uvchr_buf_helper` for `Perl_utf8_to_uvchr` The call `utf8_to_uvchr_buf('A', -1)` is now behaving differently and do not throw warning as earlier

@khwilliamson, can you take a look?  Thanks.

atoomic:
Here is a potential fix 

https://github.com/Dual-Life/Devel-PPPort/issues/219#issuecomment-1059581475

haarg:
If the plan is to continue maintaining Devel::PPPort outside of core, shouldn't it be moved to `cpan/`? Alternatively, everything in `dist/` should be tested against older perl versions.

PR #19491, opened 2022-03-04
Fix definition of ITEM_NOT_FOUND for pre-1.13 versions.

jkeenan:
Signed-off-by: James E Keenan <jkeenan@cpan.org>

### DIFF
--- a/ext/GDBM_File/GDBM_File.xs
+++ b/ext/GDBM_File/GDBM_File.xs
@@ -145,14 +145,14 @@ output_datum(pTHX_ SV *arg, char *str, int size)
 #define gdbm_setopt(db,optflag,optval,optlen) not_here("gdbm_setopt")
 #endif
 
-#ifndef GDBM_ITEM_NOT_FOUND
-# define GDBM_ITEM_NOT_FOUND GDBM_NO_ERROR
+#if GDBM_VERSION_MAJOR == 1 && GDBM_VERSION_MINOR < 13
+/* Prior to 1.13, only gdbm_fetch set GDBM_ITEM_NOT_FOUND if the requested
+   key did not exist.  Other similar functions would set GDBM_NO_ERROR instead.
+   The GDBM_ITEM_NOT_FOUND existed as early as in 1.7.3 */
+# define ITEM_NOT_FOUND()  (gdbm_errno == GDBM_NO_ERROR || gdbm_errno == GDBM_ITEM_NOT_FOUND)
+#else
+# define ITEM_NOT_FOUND()  (gdbm_errno == GDBM_ITEM_NOT_FOUND)
 #endif
-
-/* Prior to 1.13, gdbm_fetch family functions set gdbm_errno to GDBM_NO_ERROR
-   if the requested key did not exist */
-#define ITEM_NOT_FOUND()                                                \
-    (gdbm_errno == GDBM_ITEM_NOT_FOUND || gdbm_errno == GDBM_NO_ERROR)
 
 #define CHECKDB(db) do {                        \
     if (!db->dbp) {                             \
--- a/ext/GDBM_File/GDBM_File.pm
+++ b/ext/GDBM_File/GDBM_File.pm
@@ -363,7 +363,7 @@ require XSLoader;
 );
 
 # This module isn't dual life, so no need for dev version numbers.
-$VERSION = '1.19';
+$VERSION = '1.20';
 
 XSLoader::load();
 


ext/GDBM_File/GDBM_File.xs
```diff
@@ -145,15 +145,15 @@ output_datum(pTHX_ SV *arg, char *str, int size)
 #define gdbm_setopt(db,optflag,optval,optlen) not_here("gdbm_setopt")
 #endif
 
-#ifndef GDBM_ITEM_NOT_FOUND
-# define GDBM_ITEM_NOT_FOUND GDBM_NO_ERROR
+#if GDBM_VERSION_MAJOR == 1 && GDBM_VERSION_MINOR < 13
+/* Prior to 1.13, only gdbm_fetch set GDBM_ITEM_NOT_FOUND if the requested
+   key did not exist.  Other similar function wouls set GDBM_NO_ERROR instead.
```
steve-m-hay: Typos: Should say "...functions would..."

ext/GDBM_File/GDBM_File.xs
```diff
@@ -145,15 +145,15 @@ output_datum(pTHX_ SV *arg, char *str, int size)
...
+   The GDBM_ITEM_NOT_FOUND existeds as early as in 1.7.3 */
```
steve-m-hay: Typo: Should say "...existed..."

ext/GDBM_File/GDBM_File.xs
```diff
@@ -145,15 +145,15 @@ output_datum(pTHX_ SV *arg, char *str, int size)
 #define gdbm_setopt(db,optflag,optval,optlen) not_here("gdbm_setopt")
 #endif
 
-#ifndef GDBM_ITEM_NOT_FOUND
-# define GDBM_ITEM_NOT_FOUND GDBM_NO_ERROR
+#if GDBM_VERSION_MAJOR == 1 && GDBM_VERSION_MINOR < 13
+/* Prior to 1.13, only gdbm_fetch set GDBM_ITEM_NOT_FOUND if the requested
```
steve-m-hay, line 149: Are there versions earlier than 1.0 that are likely to be in use? E.g. What if someone is using 0.9?...

steve-m-hay (commented):
I don't know anything about GDBM so I can't really say whether the change is correct, but it looks reasonable aside from the minor comments I've added.

ext/GDBM_File/GDBM_File.xs
jkeenan: Done in next commit to branch; also next typo.

ext/GDBM_File/GDBM_File.xs
jkeenan: The code and the comment were originally composed by one of the two GNU project maintainers of GDBM (whom I've contacted off-list re this ticket).  In addition, releases with a `0` major version, if they ever existed, are no longer available for download at the [GNU site](http://ftp.gnu.org/gnu/gdbm/).  So I think we're safe here.

jkeenan: Corrected typos; corrected $VERSION; squashed commits and force-pushed smoke-me branch.

Leont:
> I don't know anything about GDBM so I can't really say whether the change is correct, but it looks reasonable aside from the minor comments I've added.

Same, @graygnuorg is the person to ask

Leont (approved):

Leont:
It fixes the problem on my machine

jkeenan:
> It fixes the problem on my machine

Thanks.  Could you cast a vote for this in the maint-votes branch?

@steve-m-hay, the commit listed in the maint-votes branch will have some merge conflicts and need a `$VERSION` bump.  I did this in this pull request, but I'll leave it up to you to figure out the best way to apply it to maint-5.34 and the next RC tarball.

graygnuorg:
Yes, this should fix the bug. GDBM_* are enum constants, not defines, so that #ifdef is not able to catch them.

steve-m-hay:
On Sat, 5 Mar 2022 at 00:25, James E Keenan ***@***.***>
wrote:

> It fixes the problem on my machine
>
> Thanks. Could you cast a vote for this in the maint-votes branch?
>
> @steve-m-hay <https://github.com/steve-m-hay>, the commit listed in the
> maint-votes branch will have some merge conflicts and need a $VERSION
> bump. I did this in this pull request, but I'll leave it up to you to
> figure out the best way to apply it to maint-5.34 and the next RC tarball.
>
Now merged. I took the commit before it too (
aacd2398e766500cb5d83c4d76b642fcf31d997a) as the simplest way to avoid
merge conflicts. I also took  the commit after it (
9fa00226c78c14bc34a562b90346047d2dcdd3b1), which fixes the same spelling
mistakes that I spotted in review! I hadn't realized they'd already been
fixed in blead.

PR #19492, opened 2022-03-04
Fix possible null pointer dereferences in Perl_my_strtod()

t-a-k:
I found that `Perl_my_strtod()` apparently dereferences null pointer when `Perl_strtod` macro is not defined.

I couldn't test this, because I couldn't find when this code is actually used (even if configured with `-Ud_strtod`).


hvds:
LGTM, though it's a bit surprising we have no test that exercises it. It may be worth adding a simple one to XS-APItest.

t-a-k:
> It may be worth adding a simple one to XS-APItest.

Thanks for comment.  I've pushed a commit to add a test using XS::APItest.

hvds:
Thanks @t-a-k, I was about to merge but I see @khwilliamson has beat me to it.

PR #19493, opened 2022-03-05
hv.c: ALWAYS perform in_collision check

demerphq:
This is a critical patch. Any perl built with the previous code
without PERL_HASH_RANDOMIZE_KEYS would never perform a key split!

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -849,6 +849,7 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
         HeKEY_hek(entry) = save_hek_flags(key, klen, hash, flags);
     }
     HeVAL(entry) = val;
+    in_collision = cBOOL(*oentry != NULL);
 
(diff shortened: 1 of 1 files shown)


demerphq:
Since this is a critical fix for builds without PERL_HASH_RANDOMIZE_KEYS I am merging it myself immediately.

hvds:
You only avoid PERL_HASH_RANDOMIZE_KEYS by specifying PERL_PERTURB_KEYS_DISABLED, and in the section that mentions that option `INSTALL` says:
```
B<Unless stated otherwise these options are considered experimental or
insecure and are not recommended for production use.>
```

So arguably it isn't _so_ critical that it can't wait for a second pair of eyes (and would maybe not even meet our standard criteria for backporting). The fix does seem obviously correct though.

atoomic:
Also note that even if this was critical you would not be able to fix all installed perl... I'm not sure to understand the urge of merging. Anyway thanks for the fix, it LGTM.

PR #18095, opened 2020-08-27
Cannot reproduce run with PERL_HASH_SEED when using PERL_PERTURB_KEYS=DETERMINISTIC

atoomic:
This is either an issue in the documentation or a bug in the source code itself. 
After reading the documentation about PERL_PERTURB_KEYS I lean to the second option.
view `PERL_PERTURB_KEYS` section from https://github.com/Perl/perl5/blob/blead/pod/perlrun.pod for more details.

PERL_PERTURB_KEYS can be one of the three possible values:
- 0: NO 
- 1: RANDOM
- 2: DETERMINISTIC

When using `RANDOM` the doc in `perlrun/PERL_PERTURB_KEYS` makes it clear that `the order may not be repeatable in a following program run even if the PERL_HASH_SEED has been specified.`

But `NO` and `DETERMINISTIC` modes are supposed to be `repeatable` when preserving PERL_HASH_SEED.
As shown by the additional unit test this is not the case for `DETERMINISTIC`. (you may have to run it multiple times)

This pull request adds three commits:

1. move `runperl_and_capture` helper to a generic location
2. add `t/run/runenv_hashseed.t` to show the limitation when using `DETERMINISTIC` mode
3. provide a fix candidate to honor `PERL_HASH_SEED` under `DETERMINISTIC` mode

This is clearly a change that needs to be reviewed by @demerphq before being merged.

Thanks to @jkeenan for discovering this issue while working on https://github.com/atoomic/perl/issues/250

PR #19494, opened 2022-03-05
perlrun.pod and INSTALL: document PERL_PERTURB_KEYS better

demerphq:
Expand on the documentation of PERL_PERTURB_KEYS and PERL_HASH_SEED.
Specifically note that any change in the environment may change the
key order under PERL_PERTURB_KEYS=DETERMINISTIC.

See PR #18095 for related discussion.


hvds (approved):
LGTM

PR #19496, opened 2022-03-05
perl.c: Remove run time docs and redundant support for -DH

demerphq:
In 2017 in fcd573e77ec68fbe3936ac1381654581fba8a64f Dave Mitchell
removed support for the -DH flag but missed this.

Part of the -D options parsing is positionally sensitive, so in order to
be able to remove the 'H' option entirely I had to use a placeholder
character, and I chose '?'. This is because it is not isWORDCHAR(),
which means the preceding logic prevents it from being chosen anyway,
and thus it should be safe.

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -3360,7 +3360,6 @@ Perl_get_debug_opts(pTHX_ const char **s, bool givehelp)
       "  r  Regular expression parsing and execution\n"
       "  x  Syntax tree dump\n",
       "  u  Tainting checks\n"
-      "  H  Hash dump -- usurps values()\n"
       "  X  Scratchpad allocation\n"
       "  D  Cleaning up\n"
       "  S  Op slab allocation\n"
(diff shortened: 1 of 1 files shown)


perl.c
```diff
@@ -3384,7 +3383,13 @@ Perl_get_debug_opts(pTHX_ const char **s, bool givehelp)
 
     if (isALPHA(**s)) {
         /* if adding extra options, remember to update DEBUG_MASK */
```
hvds, line 3385: Should this be "if adding or removing options"? `DEBUG_MASK` still has 0x2000 set, though it looks unlikely to cause any problems.

hvds (approved):
LGTM, but possible followup for DEBUG_MASK.

perl.c
demerphq: FWIW, I plan to reuse 0x2000. So there isn't much point in doing anything about it. I can change the comment when I push my PR to replace the '?' with a new letter.

PR #19498, opened 2022-03-06
Rename builtins to snake_case

leonerd:
Adds the underscore to make the names `is_bool` and `is_weak`.

Also includes back-compat aliases as the old names for now, because there's already CPAN code which uses these. They don't have to live too long; we can remove them in 5.37.1.


demerphq:
On Sun, 6 Mar 2022 at 02:55, Paul Evans ***@***.***> wrote:

> Adds the underscore to make the names is_bool and is_weak.
>
> Also includes back-compat aliases as the old names for now, because
> there's already CPAN code which uses these. They don't have to live too
> long; we can remove them in 5.37.1.
>
>   https://github.com/Perl/perl5/pull/19498
> Commit Summary
>
>    - d5fad83
>    <https://github.com/Perl/perl5/pull/19498/commits/d5fad835bd2f3671f751f7777bb22805443013e6>
>    Rename is{bool,weak} to is_{bool,weak}
>    - 07963f2
>    <https://github.com/Perl/perl5/pull/19498/commits/07963f2a489147a5c25bfcb51e9a9e69b5d77b24>
>    Also re-add isbool and isweak as (temporary) aliases for compatibility
>
> You missed the tests in lib/B/Deparse.t

Cheers,
Yves

haarg:
Do we really want to maintain the non-snake_case aliases into a stable release, when they have only existed in dev releases? Anything using them on CPAN is relying on pre-release features and it should be fine to break them.

leonerd:
@haarg I don't mind too much either way, I just know I already have some CPAN modules that opportunistically check for `isbool` if it exists. It's purely a single extra line in `builtin.c` to add those, and we can remove them in 5.37.1 anyway. But if folks really object to that I can just not add it now.

demerphq:
LGTM

demerphq (approved):

leonerd:
Force-pushed without the back-compat aliases in

haarg (approved):

PR #19499, opened 2022-03-06
improve test and debug hash randomization, and related improvements

demerphq:
This replaces PR #18095, please review that PR for related discussion that lead to this PR.

This PR includes the following:

1. Updates to test.pl used by the PR.
2. Updates to our SBOX32 hash initialization to be more robust as revealed by the original PR from @atoomic 
3. A reworking of the PL_hash_rand_bits update logic so it is wrapped in consistent macros which include the capability to produce debug output
4. Add a new -Dh debug switch to enabled the diagnostics introduced in point 3.
5. Add a new test which fairly thoroughly tests the hash randomization logic and seed initialization, including verifying that a bit change to the seed actually results in a significant change in hash behavior. The depends on all of the changes mentioned above.

The test is based on work by @atoomic which was apparently the result of observations by @jkeenan. Thanks to both for bringing this to my attention.

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -49,6 +49,62 @@ holds the key and hash value.
 static const char S_strtab_error[]
     = "Cannot modify shared string table in hv_%s";
 
+#define DEBUG_HASH_RAND_BITS (DEBUG_h_TEST)
+
+/* Algorithm "xor" from p. 4 of Marsaglia, "Xorshift RNGs"
+ * See also https://en.wikipedia.org/wiki/Xorshift
+ */
+#if IVSIZE == 8
+/* 64 bit version */
+#define XORSHIFT_RAND_BITS(x)   \
+STMT_START {                    \
+    (x) ^= (x) << 13;           \
+    (x) ^= (x) >> 17;           \
+    (x) ^= (x) << 5;            \
+} STMT_END
+#else
+/* 32 bit version */
+#define XORSHIFT_RAND_BITS(x)   \
+STMT_START {                    \
+    (x) ^= (x) << 13;           \
+    (x) ^= (x) >> 7;            \
+    (x) ^= (x) << 17;           \
+} STMT_END
+#endif
+
+#define UPDATE_HASH_RAND_BITS_KEY(key,klen)                             \
+STMT_START {                                                            \
+    XORSHIFT_RAND_BITS(PL_hash_rand_bits);                              \
+    if (DEBUG_HASH_RAND_BITS) {                                         \
+        PerlIO_printf( Perl_debug_log,                                  \
+            "PL_hash_rand_bits=%016"UVxf" @ %s:%-4d",                   \
+            (UV)PL_hash_rand_bits, __FILE__, __LINE__                   \
+        );                                                              \
+        if (DEBUG_v_TEST && key) {                                      \
+            PerlIO_printf( Perl_debug_log, " key:'%.*s' %"UVuf"\n",     \
+                    (int)klen,                                          \
+                    key ? key : "", /* silence warning */               \
+                    (UV)klen                                            \
+            );                                                          \
+        } else {                                                        \
+            PerlIO_printf( Perl_debug_log, "\n");                       \
+        }                                                               \
+    }                                                                   \
+} STMT_END
+
+#define MAYBE_UPDATE_HASH_RAND_BITS_KEY(key,klen)                       \
+STMT_START {                                                            \
+    if (PL_HASH_RAND_BITS_ENABLED)                                      \
+        UPDATE_HASH_RAND_BITS_KEY(key,klen);                            \
+} STMT_END
+
+
+#define UPDATE_HASH_RAND_BITS()                                         \
+    UPDATE_HASH_RAND_BITS_KEY(NULL,0)
+
+#define MAYBE_UPDATE_HASH_RAND_BITS()                                   \
+    MAYBE_UPDATE_HASH_RAND_BITS_KEY(NULL,0)
+
 #ifdef PURIFY
 
 #define new_HE() (HE*)safemalloc(sizeof(HE))
@@ -851,15 +907,17 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
     HeVAL(entry) = val;
     in_collision = cBOOL(*oentry != NULL);
 
+
 #ifdef PERL_HASH_RANDOMIZE_KEYS
     /* This logic semi-randomizes the insert order in a bucket.
      * Either we insert into the top, or the slot below the top,
      * making it harder to see if there is a collision. We also
      * reset the iterator randomizer if there is one.
      */
+
+
     if ( *oentry && PL_HASH_RAND_BITS_ENABLED) {
-        PL_hash_rand_bits++;
-        PL_hash_rand_bits= ROTL_UV(PL_hash_rand_bits,1);
+        UPDATE_HASH_RAND_BITS_KEY(key,klen);
         if ( PL_hash_rand_bits & 1 ) {
             HeNEXT(entry) = HeNEXT(*oentry);
             HeNEXT(*oentry) = entry;
@@ -884,11 +942,7 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
                              pTHX__VALUE);
         }
         */
-        if (PL_HASH_RAND_BITS_ENABLED) {
-            if (PL_HASH_RAND_BITS_ENABLED == 1)
-                PL_hash_rand_bits += (PTRV)entry + 1;  /* we don't bother to use ptr_hash here */
-            PL_hash_rand_bits= ROTL_UV(PL_hash_rand_bits,1);
-        }
+        MAYBE_UPDATE_HASH_RAND_BITS_KEY(key,klen);
         HvAUX(hv)->xhv_rand= (U32)PL_hash_rand_bits;
     }
 #endif
@@ -1466,11 +1520,7 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
      * second from top. After each such insert we rotate the hashed value. So we can
      * use the same hashed value over and over, and in normal build environments use
      * very few ops to do so. ROTL32() should produce a single machine operation. */
-    if (PL_HASH_RAND_BITS_ENABLED) {
-        if (PL_HASH_RAND_BITS_ENABLED == 1)
-            PL_hash_rand_bits += ptr_hash((PTRV)a);
-        PL_hash_rand_bits = ROTL_UV(PL_hash_rand_bits,1);
-    }
+    MAYBE_UPDATE_HASH_RAND_BITS();
 #endif
     HvARRAY(hv) = (HE**) a;
     HvMAX(hv) = newsize - 1;
@@ -1498,8 +1548,7 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
                  * and use the new low bit to decide if we insert at top,
                  * or next from top. IOW, we only rotate on a collision.*/
                 if (aep[j] && PL_HASH_RAND_BITS_ENABLED) {
-                    PL_hash_rand_bits+= ROTL32(HeHASH(entry), 17);
-                    PL_hash_rand_bits= ROTL_UV(PL_hash_rand_bits,1);
+                    UPDATE_HASH_RAND_BITS();
                     if (PL_hash_rand_bits & 1) {
                         HeNEXT(entry)= HeNEXT(aep[j]);
                         HeNEXT(aep[j])= entry;
@@ -1568,6 +1617,7 @@ Perl_hv_ksplit(pTHX_ HV *hv, IV newmax)
         hsplit(hv, oldsize, newsize);
 #ifdef PERL_HASH_RANDOMIZE_KEYS
         if (was_ook && SvOOK(hv) && HvTOTALKEYS(hv)) {
+            MAYBE_UPDATE_HASH_RAND_BITS();
             HvAUX(hv)->xhv_rand = (U32)PL_hash_rand_bits;
         }
 #endif
@@ -2256,12 +2306,7 @@ S_hv_auxinit(pTHX_ HV *hv) {
         }
         iter = Perl_hv_auxalloc(aTHX_ hv);
 #ifdef PERL_HASH_RANDOMIZE_KEYS
-        if (PL_HASH_RAND_BITS_ENABLED) {
-            /* mix in some new state to PL_hash_rand_bits to "randomize" the traversal order*/
-            if (PL_HASH_RAND_BITS_ENABLED == 1)
-                PL_hash_rand_bits += ptr_hash((PTRV)array);
-            PL_hash_rand_bits = ROTL_UV(PL_hash_rand_bits,1);
-        }
+        MAYBE_UPDATE_HASH_RAND_BITS();
         iter->xhv_rand = (U32)PL_hash_rand_bits;
 #endif
     } else {
--- a/perl.c
+++ b/perl.c
@@ -1672,27 +1672,7 @@ perl_parse(pTHXx_ XSINIT_t xsinit, int argc, char **argv, char **env)
 #ifndef MULTIPLICITY
     PERL_UNUSED_ARG(my_perl);
 #endif
-#if (defined(USE_HASH_SEED) || defined(USE_HASH_SEED_DEBUG)) && !defined(NO_PERL_HASH_SEED_DEBUG)
-    {
-        const char * const s = PerlEnv_getenv("PERL_HASH_SEED_DEBUG");
-
-        if (s && strEQ(s, "1")) {
-            const unsigned char *seed= PERL_HASH_SEED;
-            const unsigned char *seed_end= PERL_HASH_SEED + PERL_HASH_SEED_BYTES;
-            PerlIO_printf(Perl_debug_log, "HASH_FUNCTION = %s HASH_SEED = 0x", PERL_HASH_FUNC);
-            while (seed < seed_end) {
-                PerlIO_printf(Perl_debug_log, "%02x", *seed++);
-            }
-#ifdef PERL_HASH_RANDOMIZE_KEYS
-            PerlIO_printf(Perl_debug_log, " PERTURB_KEYS = %d (%s)",
-                    PL_HASH_RAND_BITS_ENABLED,
-                    PL_HASH_RAND_BITS_ENABLED == 0 ? "NO" : PL_HASH_RAND_BITS_ENABLED == 1 ? "RANDOM" : "DETERMINISTIC");
-#endif
-            PerlIO_printf(Perl_debug_log, "\n");
-        }
-    }
-#endif /* #if (defined(USE_HASH_SEED) ... */
-
+    debug_hash_seed(false);
 #ifdef __amigaos4__
     {
         struct NameTranslationInfo nti;
@@ -2324,6 +2304,8 @@ S_parse_body(pTHX_ char **env, XSINIT_t xsinit)
         Perl_drand48_init_r(&PL_internal_random_state, seed());
     }
 #endif
+    if (DEBUG_h_TEST)
+        debug_hash_seed(true);
 
     /* Set $^X early so that it can be used for relocatable paths in @INC  */
     /* and for SITELIB_EXP in USE_SITECUSTOMIZE                            */
@@ -3375,6 +3357,8 @@ Perl_get_debug_opts(pTHX_ const char **s, bool givehelp)
       "  L  trace some locale setting information--for Perl core development\n",
       "  i  trace PerlIO layer processing\n",
       "  y  trace y///, tr/// compilation and execution\n",
+      "  h  Show (h)ash randomization debug output"
+                " (changes to PL_hash_rand_bits)\n",
       NULL
     };
     UV uv = 0;
@@ -3382,14 +3366,22 @@ Perl_get_debug_opts(pTHX_ const char **s, bool givehelp)
     PERL_ARGS_ASSERT_GET_DEBUG_OPTS;
 
     if (isALPHA(**s)) {
-        /* if adding extra options, remember to update DEBUG_MASK */
-        /* Note that the ? indicates an unused slot. As the code below
+        /* NOTE:
+         * If adding new options add them to the END of debopts[].
+         * If you remove an option replace it with a '?'.
+         * If there is a free slot available marked with '?' feel
+         * free to reuse it for something else.
+         *
+         * Regardles remember to update DEBUG_MASK in perl.h, and
+         * update the documentation above AND in pod/perlrun.pod.
+         *
+         * Note that the ? indicates an unused slot. As the code below
          * indicates the position in this list is important. You cannot
          * change the order or delete a character from the list without
          * impacting the definitions of all the other flags in perl.h
          * However because the logic is guarded by isWORDCHAR we can
          * fill in holes with non-wordchar characters instead. */
-        static const char debopts[] = "psltocPmfrxuU?XDSTRJvCAqMBLiy";
+        static const char debopts[] = "psltocPmfrxuUhXDSTRJvCAqMBLiy";
 
         for (; isWORDCHAR(**s); (*s)++) {
             const char * const d = strchr(debopts,**s);
--- a/sbox32_hash.h
+++ b/sbox32_hash.h
@@ -1390,71 +1390,20 @@
 #define case_1_SBOX32(hash,state,key) /**/
 #endif
 
-#define XORSHIFT96_set(r,x,y,z,t) STMT_START {          \
-    t = (x ^ ( x << 10 ) );                             \
-    x = y; y = z;                                       \
-    r = z = (z ^ ( z >> 26 ) ) ^ ( t ^ ( t >> 5 ) );    \
-} STMT_END
-
 #define XORSHIFT128_set(r,x,y,z,w,t) STMT_START {       \
     t = ( x ^ ( x << 5 ) );                             \
     x = y; y = z; z = w;                                \
     r = w = ( w ^ ( w >> 29 ) ) ^ ( t ^ ( t >> 12 ) );  \
 } STMT_END
 
-#define SBOX32_SCRAMBLE32(v,prime) STMT_START {  \
-    v ^= (v>>9);                        \
-    v ^= (v<<21);                       \
-    v ^= (v>>16);                       \
-    v *= prime;                         \
-    v ^= (v>>17);                       \
-    v ^= (v<<15);                       \
-    v ^= (v>>23);                       \
-} STMT_END
-
 #ifndef SBOX32_CHURN_ROUNDS 
-#define SBOX32_CHURN_ROUNDS 5
-#endif
-#ifndef SBOX32_SKIP_MASK
-#define SBOX32_SKIP_MASK 0x3
+#define SBOX32_CHURN_ROUNDS 128
 #endif
 
 #define _SBOX32_CASE(len,hash,state,key) \
     /* FALLTHROUGH */ \
     case len: hash ^= state[ 1 + ( 256 * ( len - 1 ) ) + key[ len - 1 ] ];
 
-
-SBOX32_STATIC_INLINE void sbox32_seed_state96 (
-    const U8 *seed_ch,
-    U8 *state_ch
-) {
-    const U32 *seed= (const U32 *)seed_ch;
-    U32 *state= (U32 *)state_ch;
-    U32 *state_cursor = state + 1;
-    U32 *sbox32_end = state + 1 + (256 * SBOX32_MAX_LEN);
-    U32 s0 = seed[0] ^ 0x68736168; /* sbox */
-    U32 s1 = seed[1] ^ 0x786f6273; /* hash */
-    U32 s2 = seed[2] ^ 0x646f6f67; /* good */
-    U32 t1,t2,i;
-
-    /* make sure we have all non-zero state elements */
-    if (!s0) s0 = 1;
-    if (!s1) s1 = 2;
-    if (!s2) s2 = 4;
-
-    /* Do a bunch of mix rounds to avalanche the seedbits
-     * before we use them for the XORSHIFT rng. */
-    for ( i = 0; i < SBOX32_CHURN_ROUNDS; i++ )
-        SBOX32_MIX3(s0,s1,s2,"SEED STATE");
-
-    while ( state_cursor < sbox32_end ) {
-        U32 *row_end = state_cursor + 256; 
-        for ( ; state_cursor < row_end; state_cursor++ ) {
-            XORSHIFT96_set(*state_cursor,s0,s1,s2,t1);
-        }
-    }
-    XORSHIFT96_set(*state,s0,s1,s2,t2);
-}
 
 SBOX32_STATIC_INLINE void sbox32_seed_state128 (
     const U8 *seed_ch,
@@ -1464,8 +1413,8 @@ SBOX32_STATIC_INLINE void sbox32_seed_state128 (
     U32 *state= (U32 *)state_ch;
     U32 *state_cursor = state + 1;
     U32 *sbox32_end = state + 1 + (256 * SBOX32_MAX_LEN);
-    U32 s0 = seed[0] ^ 0x68736168; /* sbox */
-    U32 s1 = seed[1] ^ 0x786f6273; /* hash */
+    U32 s0 = seed[1] ^ 0x786f6273; /* sbox */
+    U32 s1 = seed[0] ^ 0x68736168; /* hash */
     U32 s2 = seed[2] ^ 0x646f6f67; /* good */
     U32 s3 = seed[3] ^ 0x74736166; /* fast */
     U32 t1,t2,i;
@@ -1475,9 +1424,23 @@ SBOX32_STATIC_INLINE void sbox32_seed_state128 (
     if (!s1) s1 = 2;
     if (!s2) s2 = 4;
     if (!s3) s3 = 8;
-    
+
     /* Do a bunch of mix rounds to avalanche the seedbits
      * before we use them for the XORSHIFT rng. */
+    for ( i = 0; i < SBOX32_CHURN_ROUNDS; i++ )
+        SBOX32_MIX4(s0,s1,s2,s3,"SEED STATE");
+
+    s0 ^= ~seed[3];
+    s1 ^= ~seed[2];
+    s2 ^= ~seed[1];
+    s3 ^= ~seed[0];
+
+    /* make sure we have all non-zero state elements, again */
+    if (!s0) s0 = 8;
+    if (!s1) s1 = 4;
+    if (!s2) s2 = 2;
+    if (!s3) s3 = 1;
+    
     for ( i = 0; i < SBOX32_CHURN_ROUNDS; i++ )
         SBOX32_MIX4(s0,s1,s2,s3,"SEED STATE");
 
@@ -1758,16 +1721,6 @@ SBOX32_STATIC_INLINE U32 sbox32_hash_with_state(
         case 0: break;
     }
     return hash;
-}
-
-SBOX32_STATIC_INLINE U32 sbox32_hash96(
-    const U8 *seed_ch,
-    const U8 *key,
-    const STRLEN key_len
-) {
-    U32 state[SBOX32_STATE_WORDS];
-    sbox32_seed_state96(seed_ch,(U8*)state);
-    return sbox32_hash_with_state((U8*)state,key,key_len);
 }
 
 SBOX32_STATIC_INLINE U32 sbox32_hash128(
--- a/t/test.pl
+++ b/t/test.pl
@@ -817,10 +817,15 @@ sub runperl {
 # Nice alias
 *run_perl = *run_perl = \&runperl; # shut up "used only once" warning
 
--# Run perl with specified environment and arguments, return (STDOUT, STDERR)
+# Run perl with specified environment and arguments, return (STDOUT, STDERR)
+# set DEBUG_RUNENV=1 in the environment to debug.
 sub runperl_and_capture {
-  local *F;
   my ($env, $args) = @_;
+
+  my $STDOUT = tempfile();
+  my $STDERR = tempfile();
+  my $PERL   = $^X;
+  my $FAILURE_CODE = 119;
 
   local %ENV = %ENV;
   delete $ENV{PERLLIB};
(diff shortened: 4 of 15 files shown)


t/test.pl
```diff
@@ -922,28 +922,38 @@ sub _num_to_alpha{
 my %tmpfiles;
 END { unlink_all keys %tmpfiles }
 
+
+# IMPORTANT NODE: tempfile() may be used as a module names in our tests
```
hvds: `s/NODE/NOTE/; s/names/name/`

(Also more generally, saying "NOTE: " is already enough to clue that it's important, doesn't really need "IMPORTANT" there too.)

t/test.pl
```diff
@@ -922,28 +922,38 @@ sub _num_to_alpha{
...
+my $max_file_chars = 3;
 sub tempfile {
-    while(1){
-	my $try = (-d "t" ? "t/" : "")."tmp$$";
-        my $alpha = _num_to_alpha($tempfile_count,2);
+    # if you change the format returned by tempfile() you MUST change
+    # the $::tempfile_regex define above.
+    my $try_prefix = (-d "t" ? "t/" : "")."tmp_"._num_to_alpha($$);
+    while(my $alpha = _num_to_alpha($tempfile_count,$max_file_chars)) {
         last unless defined $alpha;
```
hvds: There's a redundant test now. But I'd be inclined to write:
```
  while (1) {
    my $alpha = _num_to_alpha($tempfile_count, $max_file_chars);
    last unless defined $alpha;
```
to keep the constructs in use minimal: the scoping of a lexical introduced in the COND of a while seems like the sort of thing to avoid relying on here.

t/test.pl
```diff
@@ -922,28 +922,38 @@ sub _num_to_alpha{
...
-        $try = $try . $alpha;
+        my $try = $try_prefix . "_" . $alpha;
         $tempfile_count = $tempfile_count + 1;
 
-	# Need to note all the file names we allocated, as a second request may
-	# come before the first is created.
+        # Need to note all the file names we allocated, as a second request
+        # may come before the first is created. Also we are avoiding ++ here
+        # so we aren't using the normal idiom for this kind of test.
 	if (!$tmpfiles{$try} && !-e $try) {
```
hvds, line 1021: Please fix the indentation of this block to match the fixed comment above it.

t/test.pl
```diff
@@ -922,28 +922,38 @@ sub _num_to_alpha{
...
+# so the result must be restricted to only legal characters for a module
+# name.
+
 # A regexp that matches the tempfile names
-$::tempfile_regexp = 'tmp\d+[A-Z][A-Z]?';
+$::tempfile_regexp = 'tmp_[A-Z]+_[A-Z]+';
 
 # Avoid ++, avoid ranges, avoid split //
 my $tempfile_count = 0;
+my $max_file_chars = 3;
```
hvds, line 1007: I don't know how portable it is to assume you can write up to 26^3 files in a directory. It might be worth considering setting a limit on `$tempfile_count` instead, and letting the file_chars grow as needed for that. That'll also help limit the mess from a runaway script.

Trying out the test file from a later commit, it appears to be generating 1200 files, which already sounds quite high - I can certainly imagine filesystems having a limit as low as 1024, though I don't know what actual limits exist in the systems we target. (Off the top of my head, I could imagine any of VMS, AmigaOS or Z/OS having lower limits.)

Another approach would be to leave the old 676 limit alone, but provide a new function here:
```
sub clear_tempfiles {
  unlink_all keys %tmpfiles;
  %tmpfiles = ();
}
```
.. and for the new test file to invoke that after each (successful?) septuplet of tests.

t/test.pl
```diff
@@ -817,6 +817,64 @@ sub runperl {
...
+    return (0, "Failure in child.\n") if ($?>>8) == $FAILURE_CODE;
+
+    open my $stdout, '<', $STDOUT
+    or return (0, "Couldn't read $STDOUT file: $!");
+    open my $stderr, '<', $STDERR
+    or return (0, "Couldn't read $STDERR file: $!");
+    local $/;
+    # Empty file with <$stderr> returns nothing in list context
+    # (because there are no lines) Use scalar to force it to ''
+    my $out= scalar <$stdout>;
```
hvds: Now it's a scalar assignment (instead of return of a list), the comment and explicit `scalar` are redundant.

t/test.pl
```diff
@@ -817,6 +817,64 @@ sub runperl {
...
+    if ($out eq $bogus) {
+        die "panic: something went wrong with runperl_and_capture";
+    }
+    my $err= scalar <$stderr>;
+    if ($ENV{DEBUG_RUNENV}) {
+        print "OUT: $out\n";
+        print "ERR: $err\n";
```
hvds, line 857: I recommend printing with explicit delimiters around `$out` and `$err`, so whitespace stands out.

t/test.pl
```diff
@@ -817,6 +817,64 @@ sub runperl {
...
+    }
+    return ($out, $err);
+  } elsif (defined $pid) {                      # child
+    for my $k (sort keys %$env) {
```
hvds, line 863: Is the `sort` there for hash-randomization consistency? It'd be worth commenting that, since otherwise it looks redundant.

t/test.pl
```diff
@@ -817,6 +817,64 @@ sub runperl {
...
+      $ENV{$k} = $env->{$k};
+    }
+    if ($ENV{DEBUG_RUNENV}) {
+        print "Child Process $$ Executing:\n$PERL @$args\n";
+    }
+    open STDOUT, '>', $STDOUT or exit $FAILURE_CODE;
+    if (open STDERR, '>', $STDERR) {
+        exec $PERL, @$args
+            or die "Failed to exec!";
```
hvds: The original code falls through to the print and exit in this case. It isn't clear that a die() instead will find any STDERR to write to.

hvds:
@demerphq I'm working my way through this, but it occurs to me that we may want to give clearer and more stable promises about what we will and won't do with the hash functionality - because you are our only real expert in this area, it can sometimes feel like things are added or removed pretty much at your whim.

As far as I know, the main public-facing information on this is the section "Algorithmic Complexity Attacks on Hashes" in `INSTALL`, which says "unless stated otherwise these options are considered experimental or insecure and are not recommended for production use" (and does not anywhere "state otherwise", that I can see), and "from time to time we change which functions we support, and which function is default".

Can we offer something more stable, such as describing criteria used for evaluation that would lead us to remove hashing functions, or change the default? Should we consider a deprecation period before removal of a hashing function? Are any of these "considered experimental" options ever going to become non-experimental?

There might also be value in giving a table of more specific recommendations in INSTALL for a mix of security/speed/memory tradeoffs - for the build I use for maths research, for example, the security requirements are nil, but speed/memory tradeoffs are of great interest. That might even allow some of the gorier details to be moved out of this document.

hvds:
@demerphq I've reached the SBOX changes now, do you have any handy code that tests and demonstrates how effectively bits are propagated and randomized by this code? It might also be of interest to have a compile option to instrument how often the `/* make sure we have all non-zero state elements */` cases are hit - anything significantly over 1/2^32 per word would presumably signal a problem.

sbox32_hash.h
```diff
@@ -1390,94 +1390,57 @@
...
 ) {
-    const U32 *seed= (const U32 *)seed_ch;
+    U32 *seed= (U32 *)seed_ch;
     U32 *state= (U32 *)state_ch;
     U32 *state_cursor = state + 1;
     U32 *sbox32_end = state + 1 + (256 * SBOX32_MAX_LEN);
     U32 s0 = seed[0] ^ 0x68736168; /* sbox */
     U32 s1 = seed[1] ^ 0x786f6273; /* hash */
     U32 s2 = seed[2] ^ 0x646f6f67; /* good */
+    U32 s3 = seed[3] ^ 0x74736166; /* fast */
```
hvds: `s[0], s[1]` comments are the wrong way round, FWIW.

hv.c
```diff
@@ -49,6 +49,46 @@ holds the key and hash value.
 static const char S_strtab_error[]
     = "Cannot modify shared string table in hv_%s";
 
+#define DEBUG_HASH_RAND_BITS (DEBUG_h_TEST)
+
+#define UPDATE_HASH_RAND_BITS_KEY(key,klen) STMT_START {                \
+    PL_hash_rand_bits *= 17;                                            \
+    PL_hash_rand_bits = ROTL_UV(PL_hash_rand_bits,11);                  \
```
hvds: This will update 0 to 0 (which the replaced code would not), was that intended?

demerphq:
On Sun, 6 Mar 2022 at 14:45, Hugo van der Sanden
***@***.***> wrote:
>
> @demerphq I'm working my way through this, but it occurs to me that we may want to give clearer and more stable promises about what we will and won't do with the hash functionality - because you are our only real expert in this area, it can sometimes feel like things are added or removed pretty much at your whim.

I don't think there is really. Ideally we would require bare Siphash
2-4. But that would mean a performance hit, and it would mean we
couldn't support 32 bit builds.

I would prefer we found a different 32 bit hash, but finding one that
is both secure and not a performance dog is a challenge.

Anyone could plug in a different hash function if they wanted really,
all the infra is there.

I dont change things like this without evidence and a basis to do so.
In this case Nicholas R. wrote a test that showed that minimal bit
changes to the seed would not affect the results of SBOX32.  I dug
into it and the 128 bit initializer did not have the issue so I
switched. Not much more to be said really.

> As far as I know, the main public-facing information on this is the section "Algorithmic Complexity Attacks on Hashes" in INSTALL, which says "unless stated otherwise these options are considered experimental or insecure and are not recommended for production use" (and does not anywhere "state otherwise", that I can see), and "from time to time we change which functions we support, and which function is default".

I dont think we can or should make any commitments in this area. If
someone found a flaw in Siphash and we had to replace it with
something else we would. If someone pointed out a peer reviewed 32 bit
"secure" hash that had decent performance profile I would drop
Zaphod32 in an eyeblink.

>
> Can we offer something more stable, such as describing criteria used for evaluation that would lead us to remove hashing functions, or change the default?

The main criteria are security and performance. If someone published a
faster but still generally considered secure function we would switch
IMO. (Why not?)

I dont know what you mean by "stable".

> Should we consider a deprecation period before removal of a hashing function?

Not in my opinion, that is in my opinion *strongly* NO. We
specifically document we can and will change hashing functions when we
feel the need. Why would you have a deprecation cycle on something
known to be insecure?  That doesn't make sense to me. These aren't
features we promise, they are internal implementation details. It
would be like deprecating the logic we use to process UTF8 or
something like that, so long as it produces correct results it's not
the users business how it works. We could migrate our associative
array implementations to something that doesn't use hashing at all and
remove the hash function entirely if we wanted (ignoring XS). The only
commitment we make is that our associative arrays (aka hashes) work
well and are reasonably robust to hostile input.

For a while we were maintaining something like six different hash
function implementations.  That just makes a lot of things difficult,
and as far as we know nobody used them. I removed the older deprecated
ones once Siphash became the defacto industry gold standard for
secure-but-fast hash functions.

As it is we basically offer 6 options and we give the user some tuning
options as well.  They are:

    PERL_HASH_FUNC_SIPHASH
    PERL_HASH_FUNC_SIPHASH13
    PERL_HASH_FUNC_ZAPHOD32

We also allow all three to be "fronted" by SBOX32 which is very fast,
but cannot handle arbitrary length keys, and whose security potential
is in theory "perfect", but in practice its quality depends on how
well it is initialized, and the setup time and memory grows per byte
it can handle. The default is to use SBOX32 fronting for SIPHASH 1-3
on 64 bit builds and SBOX32 fronting ZAPHOD32 for 32 bit builds. We
only offer Zaphod32 because we need /something/  for 32 bit builds. (I
would love it if we "just" dropped support for that build frankly.)

Someone who cared about security over all else and who was building on
64 bit would use SIPHASH (2-4) without SBOX32. Someone who cares about
security should not be using 32 bit builds.

SBOX32 can be configured to "front" for the others for a specific
number of characters, by default we do 24, but you can configure it up
255.

>Are any of these "considered experimental" options ever going to become non-experimental?

I believe you are referring to the sentence:

B<Unless stated otherwise these options are considered experimental or
insecure and are not recommended for production use.>

I believe that that sentence is overly general and refers to things
like compiling with a fixed hash seed, or with key perturbing off, or
in the past with one of the older known to be insecure hash functions.
I would say all the options we advertise in INSTALL are not
experimental, but they may be of varying levels of security.

I think that is pretty much CYA type language. Basically if you
compile with a hard coded seed and no perturbing if someone hacks you
dont come crying to us.

> There might also be value in giving a table of more specific recommendations in INSTALL for a mix of security/speed/memory tradeoffs - for the build I use for maths research, for example, the security requirements are nil, but speed/memory tradeoffs are of great interest. That might even allow some of the gorier details to be moved out of this document.

When I read the document I think that is what it does. It says we have
chosen a balance between speed and security. It gives an option for
more secure but slower SIPHASH aka 2-4 without SBOX, the default
Siphash 1-3 with SBOX, and it gives you options for how long a string
you want to use SBOX, the longer the string the  more memory it will
use. SBOX is about as fast as can be, one array lookup and one xor per
character. It's hard to beat that, at least theoretically, OTOH cpu
pipeline stalls and caching might mean that register based hashes are
faster. YMMV.  Siphash is parameterized, we could offer Siphash 4-8 if
people were really uptight[1]. We could offer SHA256 if we wanted, or
SHA3 but it would be very slow.

I dont think there is really more we can offer or say. I would say 99%
of people aren't competent to make these decisions. I dont think it is
something we should be asking our users generally to decide on, I
think we already offer more options than anyone else does language
wise. I think most folks just say "you get Siphash 2-4 and that is
it."

cheers,
yves
The numbers for siphash refer to the number of mix rounds per block of
data read and the number of finalization rounds performed before
returning the final result. So 2-4 means every block performs 2
Siprounds, and before we return the final result performs 4.  1-3
means 1 sipround per block, and 3 before we return the result. The
higher the numbers the slower things get, with the first number more
greatly improving security, but more greatly decreasing performance.

-- 

demerphq:
On Sun, 6 Mar 2022 at 15:11, Hugo van der Sanden ***@***.***>
wrote:

> @demerphq <https://github.com/demerphq> I've reached the SBOX changes
> now, do you have any handy code that tests and demonstrates how effectively
> bits are propagated and randomized by this code? It might also be of
> interest to have a compile option to instrument how often the /* make
> sure we have all non-zero state elements */ cases are hit - anything
> significantly over 1/2^32 per word would presumably signal a problem.
>
I think you are mistaking the initializer for a hash function.  It is not
the hash function. It is a table generator.  That code takes a seed and
expands it out into tables of 256 x U32s per character we use SBOX on. Eg,
1+24 * 256  U32s (the +1 is for the empty string.)

The reason it checks for 0's in the initialization is it uses a 128 bit
marsaglia random number generator, which (like many hash and RNGs) has a 0
attractor, that is, if the initial state is 0 then it will simply output 0
every time. As long as any one of the state registers (s0, s1, s2, s3) is
non-zero it will output 2**128-1 values from 0..2**32-1. it creates a ring.
If you expanded the ring it would contain 3 x 0, and 4 x every other number
from 1..2*32-1. In an abundance of caution we do not check for "all state
registers being zero" to trigger the "override" to prevent the zero
attractor, we simply assert each cannot be zero at all.

So imagine you ran perl with a user supplied seed of all zeros.  Or
actually in this case imagine you supplied it with a seed of
0x68736168786f6273646f6f6774736166. Then s0, s1, s2, s3 would be 0, and the
marsaglia generator would produce SBOX tables containing all zeros. (At
least for the initial step.) So we detect this case and would set them to
some constants instead.  After we churn we mix in the seed again, but with
bits reversed. If in the insanely unlikely chance we ended up with any of
the state registers being 0 we unzero them and churn again.

The logic in the initializer simply ensures that the seed is well mixed
into the state vector, and that during the mixing process it does not end
up zero. We mix the user supplied seed into the marsaglia state twice.
Once xored with some constants, and then later we mix them in again but
with bits inverted. After each we check that the result is not 0 vectors.
In between both steps we churn, that is we burn off values produced by the
marsaglia generator.  What this does is ensure that the state is relatively
"rich" and does not look anything like the input state, and has a decent
mix of bits. Our hash function uses a variable number of bits from the
generated values, the churn and mix process ensures that we do not set up
the SBOX with mostly 0 bits. (Which is possible but extremely unlikely.)

I dont have much to offer beyond suggesting that you go read the literature
on these things. We could replace the initializer with SHA-1, or Siphash,
or whatever. All you need is something that is not slow, and will work with
32 bit and 64 builds and which supports a large seed. IMO this is good
enough. If someone wants to attack the SBOX they have to make their attack
work in 24 bytes without knowing the 128 bit seed and without knowing more
than the key order exposed by a keys operation, eg they do not get to see
the bits of the result of the hash function, just its effect on the data,
and with perturbing in place it is even harder, they dont even get the same
key order for the same keys.

If you revert the relevant patch and run the test you will see the test
fails. With the patch it does not. That is good enough justification to me
that this is better. If you can find something even better to initialize
with later on then all the power to you, I'll even buy you a beer, but
perfect is the enemy of good enough, and IMO over analyzing this is
counterproductive. The main question to me is "is it better than what it
replaced" and on that point I am confident that it is. I dont think
delaying this patch while you do this research makes sense, what we are
using now has an obvious hole (that atoomic found) and we should fix it
ASAP. If you find something better later on I'd be happy to help integrate
it, but for now I think this patch provides important improvements to the
SBOX initializer and the testing and debugging of the hash function and the
key perturbing logic and we shouldnt wait to get it into blead.

cheers,
yves

demerphq:
So cygwin is failing under "random" presumbly it doesnt do any address randomization like most OS'es do. Not sure if we can fix that. It means that its RANDOM mode behaves the same as "deterministic" does elsewhere.  We rely on malloc and address space randomization allowing us to use the address of the objects we operate on as a (poor) source of entropy. No much we can do about it that isn't the case. So ill adjust the test to skip on cygwin tomorrow. Now its bed time.

hv.c
```diff
@@ -49,6 +49,46 @@ holds the key and hash value.
...
+            );                                                          \
+        } else {                                                        \
+            PerlIO_printf( Perl_debug_log, "\n");                       \
+        }                                                               \
+    }                                                                   \
+} STMT_END
+
+#define UPDATE_HASH_RAND_BITS_MIX_KEY(mix,key,klen) STMT_START {        \
+    if (PL_HASH_RAND_BITS_ENABLED == 1)                                 \
+        PL_hash_rand_bits ^= (mix);                                     \
```
hvds: This appears to (approximately) replace a `+=`, so it seems like the effect would now be more localized. Is that intended?

hvds:
> > I've reached the SBOX changes now, do you have any handy code that tests and demonstrates how effectively bits are propagated and randomized by this code? It might also be of interest to have a compile option to instrument how often the /* make sure we have all non-zero state elements */ cases are hit - anything significantly over 1/2^32 per word would presumably signal a problem.
>
> I think you are mistaking the initializer for a hash function.

Ah, indeed I was, I should have realised that.

> If you revert the relevant patch and run the test you will see the test fails. With the patch it does not.

Ok, I'll examine the test.

> I dont think delaying this patch while you do this research makes sense

I was hoping to be able to do a responsible job of reviewing the code without having to do extensive research and become a hashing expert - that's not something I have time for in any case. That's why I was hoping you might have a tool that demonstrates what _you're_ looking at in evaluating one version of the code against another.

Maybe I'm asking the wrong question, or asking it the wrong way - what is it _you_ look at? How do _you_ satisfy yourself that changes like this haven't, for example, opened a new security hole?

As well as the changes in the sbox initialization, the way PL_hash_rand_bits gets updated has also substantially changed without any obvious justification for the choices made there. It's difficult to know if I should worry about the fact that changing `hash_rand_bits -> rotl(hash_rand_bits + 1, 1)` to `hash_rand_bits -> rotl(hash_rand_bits * 17, 11)` means we now take 0 to 0, or the fact that changing `hash_rand_bits += mix` to `hash_rand_bits ^= mix` would have a more localized effect.

hvds (commented):
I've run out of time for now, understanding the new test is next.

demerphq:
On Sun, 6 Mar 2022 at 19:08, Hugo van der Sanden ***@***.***>
wrote:

> I've reached the SBOX changes now, do you have any handy code that tests
> and demonstrates how effectively bits are propagated and randomized by this
> code? It might also be of interest to have a compile option to instrument
> how often the /* make sure we have all non-zero state elements */ cases are
> hit - anything significantly over 1/2^32 per word would presumably signal a
> problem.
>
> I think you are mistaking the initializer for a hash function.
>
> Ah, indeed I was, I should have realised that.
>
> If you revert the relevant patch and run the test you will see the test
> fails. With the patch it does not.
>
> Ok, I'll examine the test.
>
> I dont think delaying this patch while you do this research makes sense
>
> I was hoping to be able to do a responsible job of reviewing the code
> without having to do extensive research and become a hashing expert -
> that's not something I have time for in any case. That's why I was hoping
> you might have a tool that demonstrates what *you're* looking at in
> evaluating one version of the code against another.
>
I appreciate that, and you are a good reviewer, and I do appreciate your
reviews, you have helped me catch many things that I have overlooked.  All
I meant was that there is a nearly infinite range of alternatives to the
changes I made, and unless you have an alternative proposal you should just
take my word on it that the one I proposed is not unreasonable. At a
certain point if you are not expert enough to understand or criticize the
change based on your own knowledge then you should just give the person the
benefit of the doubt.  For instance right now I feel like I have to give
you a personal tutorial on hash and RNG construction to convince you I know
what i am doing. I dont think I should feel like that, and I dont think the
PR process is meant for that. I  wouldn't expect Nicholas or Dave M or Karl
to have to explain every little decision they made in the areas they know
better than me, I would assume that they knew what they were doing. It
doesn't mean I would not scrutinize their work for things that I might
spot, having a second set of eyes is always helpful, but I wouldn't second
guess their domain knowledge and expect them to provide a tutorial on the
things I dont know about before I would approve.

FWIW, Generally I am just using my past experience studying and testing a
wide range of hash functions, random number generators, and the modest
amount of time I have spent analyzing and playing with encryption
algorithms. I am not formally educated in this area, I am self taught with
all the upsides and downsides that brings with it. On the other hand I have
probably spent more time reviewing hash function implementations and
playing with and testing different hash functions and random number
generators than most people around. Not the professionals who do it for a
living of course, but probably more than the majority of people who are
not. I happen to enjoy it and do it for fun. I spent an inordinate amount
of time studying Siphash and SHA3 for instance. SHA-3 is really freaking
cool.

In the case of the initialization I was depending on my experience working
with Xor-shift random number generators and how they work, and combining
that with my experience looking into hash algorithm seed procedures.

I do not have a tool to assess the PL_hash_rand_bits because I do not think
that it demands that level of scrutiny. We could probably get away by
having a table with a bunch of constants we xor the riter with and just
cycle through them. An attacker wouldn't know which we are using at any
given time and it would probably be good enough for the purposes the logic
is intended for. I dont have a tool to assess the randomness of the SBOX
because I haven't bothered, XorShift-128 is pretty well studied and tested
and while not the best RNG it does reasonably well, and most importantly
has a large state vector and is fast enough we can generate the 6145 random
numbers at the top of each process without anybody noticing or caring.
Ideally we would use something cryptographically strong to generate the
SBOX tables, but that also introduces portability and performance issues,
especially portability issues. I am not sure that folks would be keen for
us to bake in SHA1 or SHA256 or into perl core just to initialize the SBOX
lookup tables.


> Maybe I'm asking the wrong question, or asking it the wrong way - what is
> it *you* look at? How do *you* satisfy yourself that changes like this
> haven't, for example, opened a new security hole?
>
Well, I ran the tests LOTS of times, and I eyeballed output, but also to a
certain extent it is just basic math and logic. Mixing the seed in once is
probably enough, mixing it in twice in different ways is an abundance of
caution and increases the chance that any bit changes in the seed will
cause the state to end up sufficiently different that the resulting tables
we generate are different enough that we pass the tests. It is not uncommon
practice to do something like this in hashing functions. Siphash for
instance mixes in the block read both before and after each sipround, first
into v0 and then into v3. It also initializes its state by XORing the seed
into its state vector two different ways. The initialization I did is
intended to be much the same idea.


> As well as the changes in the sbox initialization, the way
> PL_hash_rand_bits gets updated has also substantially changed without any
> obvious justification for the choices made there. It's difficult to know if
> I should worry about the fact that changing hash_rand_bits ->
> rotl(hash_rand_bits + 1, 1) to hash_rand_bits -> rotl(hash_rand_bits *
> 17, 11) means we now take 0 to 0, or the fact that changing hash_rand_bits
> += mix to hash_rand_bits ^= mix would have a more localized effect
>
The +1 followed by a 1 bit rotate was something I was never entirely happy
about. In the time since I wrote that code I learned a lot and I was even
less satisfied with it so since I was touching this code anyway I switched
to something that mixes the bits up better and at the same time uses less
overhead. Notice I removed the ptrhash() call as well.

Addition is "nice" in a sense for things like this as it cascades. But it
also accumulates (duh). Consider what happens if you start with 0 and add 1
repeatedly. Slowly all the bits will be turned on. If you rotate left after
each addition you'll get something a pattern like this: (using an 8 bit
state)

   0000 0000
   0000 0010
   0000 0110
   0000 1110
   0001 1110
   0011 1110
   0111 1110
   1111 1110
   1111 1111
   0000 0000

So we get a cycle with a period of 9 and a not very random pattern. The
accumulation is visible on the first 8 steps, the cascade is visible on the
last step. The cascade of addition makes it "nicer" than XOR in some
contexts as it may affect other bits than the ones that are both set as the
carry bit is propagated through.But it also bad in the sense that  the
accumulation tends to lead to all set bits. XOR doesn't do that.

So adding 1 is actually a pretty crappy way to do things. Now we dont
always add 1, at least not in the recommended  "RANDOMIZED" mode, and we
dont start off with the state at 0, and we have 64 (or 32) bits not 8. But
you get the idea. Adding some randomly chosen large odd constant would
produce a more randomlike behavior than 1. Using 1 originally was really in
hindsight a terrible choice.

I switched to xor because I also switched to multiplication and
multiplication typically affects more bits across the state than addition
does, especially when the addition has a small set of set bits like 1 does.
Why *= 17? Mostly because it is odd, and a prime, and it came to mind. We
could switch to a different larger prime and it wouldn't hurt. (Probably I
will do so in the update to the patch) Xor also doesn't cascade so there is
no "accumulation" effect. There are many hash functions based around
Xor-Multiply.  We could also probably use a combination of addition and
xor. There are lots of hash functions that are based on the use of
addition, xor and rotation. Siphash is one of them. Why did I rotate by 11?
Because I know that most of our hash tables use mostly the low bits of the
hash value and the PL_hash_rand_bits, and because I know that the high bits
of an integer multiplied by an odd number under module 2^K tend to behave
randomly.  Also because I know that when we are mixing in addresses from
malloc they usually have their low bits set to 0. all of this combines such
that the high bits of the PL_hash_rand_bits tend to be more "random" than
the low ones. Rotating them down moves those more random bits so they are
available to the task we use them for (xoring with riter) and because they
are then available to affect the next round of mutations. I also chose 11
because it is not an even divisor of 64 or 32 so in my mind it increases
the chance we will have a longer sequence before we start repeating
ourselves.

The main point here is that both the old and new operations are reversible
(multiplication by an odd number modulo K is reversible, as is xoring any
number), which means they dont lead to collisions on their own, just like
the operations they replaced.

Remember, this logic is highly opaque to an "attacker" as they dont know
how many times it has been performed, and since it happens when a collision
happens it is somewhat random already due to the randomized seeding of the
hash algorithm already. And they dont know the seed we start off with. And
in random mode we mix in secondary data like the hash value of the item, or
the address of the array. So there are a lot of unknowns for the attacker
to deal with. The objective here is to get something random enough that
when we walk the bucket array we do not walk it in an order known to an
outsider, it doesn't need to be perfect in terms of security. It needs to
avoid certain problems, like not ending up and staying 0, but other than
that something pseudo random and hard to discern by an outsider is IMO
sufficient.

So are there alternatives? Sure there are. We could use an Xorshift or LCG,
or any number of other RNG's at the cost of having to perform more cpu ops
every time the code is executed.

Having said that while walking through this it occurs to me I can tweak the
logic to be "harder" still than it is. I saw you posted a bunch of
feedback, and ill go through that next and post a follow up.

As for how I assessed this stuff? I eyeballed the state of
PL_hash_rand_bits over many runs and saw that the bits 'dance' well enough
for what I think we need, and saw that it passed the tests that atoomic and
I put together over *many* runs. (I ran the test in a loop for a long time)
That is part of the reason I added -Dh and the reason I made the test run
many times and not simply once. Really we want to run it millions of times
to be really confident, but we need to balance what we do (at every level
of this logic) with the practical constraints. Having the tests take an
hour each time just to test the hash key order randomization is not really
acceptable.

Also please keep in mind what I said earlier, this is not encryption here,
it is not creating message digests or anything that requires NIST level
security, it is not even at the level of casual random number generation,
this is merely hardening to make it more difficult for an attacker to
determine our hash seed via leakage of raw key order (especially when using
the less than secure Zaphod32), and to ensure that users do not (again)
start to depend on or assume a specific key order from our hashes. We do
this as much to prevent users from baking raw key order lists into their
tests so that we can change the hash function when we need and so we can
flush out weird bugs as we do to make an attacker's life more difficult.
It doesn't have the need to be cryptographically hard. It just needs to be
good enough.

cheers,
Yves

demerphq:
On Sun, 6 Mar 2022 at 20:15, Hugo van der Sanden ***@***.***>
wrote:

>
> I've run out of time for now, understanding the new test is next.
>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820221153>:
>
> > @@ -922,28 +922,38 @@ sub _num_to_alpha{
>  my %tmpfiles;
>  END { unlink_all keys %tmpfiles }
>
> +
> +# IMPORTANT NODE: tempfile() may be used as a module names in our tests
>
> s/NODE/NOTE/; s/names/name/
>
> (Also more generally, saying "NOTE: " is already enough to clue that it's
> important, doesn't really need "IMPORTANT" there too.)
>

Ok, changed.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820221682>:
>
> > +    while(my $alpha = _num_to_alpha($tempfile_count,$max_file_chars)) {
>          last unless defined $alpha;
>
> There's a redundant test now. But I'd be inclined to write:
>
>   while (1) {
>     my $alpha = _num_to_alpha($tempfile_count, $max_file_chars);
>     last unless defined $alpha;
>
> to keep the constructs in use minimal: the scoping of a lexical introduced
> in the COND of a while seems like the sort of thing to avoid relying on
> here.
>

I really hate while(1) {} constructs as I then have to analyze the code to
see if there is a halting condition that will actually fire, and I avoid
them as much as possible if there is a simpler and more obvious way to
express the code that shows that the loop will actually stop and when.
Having said that, if we are avoiding modern style in this code then I will
restore it to the old form.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820221909>:
>
> >          $tempfile_count = $tempfile_count + 1;
>
> -	# Need to note all the file names we allocated, as a second request may
> -	# come before the first is created.
> +        # Need to note all the file names we allocated, as a second request
> +        # may come before the first is created. Also we are avoiding ++ here
> +        # so we aren't using the normal idiom for this kind of test.
>  	if (!$tmpfiles{$try} && !-e $try) {
>
> Please fix the indentation of this block to match the fixed comment above
> it.
>
This is an artifact of how you are viewing the diff. The code is lined up.
The old code uses tabs, the new spaces. The new standard is spaces.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820222251>:
>
> >
>  # Avoid ++, avoid ranges, avoid split //
>  my $tempfile_count = 0;
> +my $max_file_chars = 3;
>
> I don't know how portable it is to assume you can write up to 26^3 files
> in a directory. It might be worth considering setting a limit on
> $tempfile_count instead, and letting the file_chars grow as needed for
> that. That'll also help limit the mess from a runaway script.
>
https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory

Says that pretty much any modern FS can handle far far more than that. 626
is simply not enough.

I am doubtful that anything that might have trouble with this number of
files can build perl.

Also, dont forget this function is called for other purposes, like creating
module names. (Blech). And the limit is enforced whether or not the files
are actually written to disk, so IMO it's fine to have a higher limit.

But I'll implement a version of your suggestion below anyway.

Trying out the test file from a later commit, it appears to be generating
> 1200 files, which already sounds quite high - I can certainly imagine
> filesystems having a limit as low as 1024,
>
I find it unlikely, at least not for anything that can compile a modern
perl. But it would be interesting to find out, but it wont be this PR. :-)


> though I don't know what actual limits exist in the systems we target.
> (Off the top of my head, I could imagine any of VMS, AmigaOS or Z/OS having
> lower limits.)
>
> Another approach would be to leave the old 676 limit alone, but provide a
> new function here:
>
> sub clear_tempfiles {
>   unlink_all keys %tmpfiles;
>   %tmpfiles = ();
> }
>
> .. and for the new test file to invoke that after each (successful?)
> septuplet of tests.
>
Sure, I'll do that. I called it "unlink_tempfiles" but same thing.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820223236>:
>
> > +    # Empty file with <$stderr> returns nothing in list context
> +    # (because there are no lines) Use scalar to force it to ''
> +    my $out= scalar <$stdout>;
>
> Now it's a scalar assignment (instead of return of a list), the comment
> and explicit scalar are redundant.
>
I fairly significantly rewrote that sub after reading your feedback on this.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820223519>:
>
> > +        print "OUT: $out\n";
> +        print "ERR: $err\n";
>
> I recommend printing with explicit delimiters around $out and $err, so
> whitespace stands out.
>
No, I don't think that is appropriate here. The output could be many lines,
adding quotes wouldn't really help except for whitespace at the start or
end. Also this is debug code for people *writing* tests, I dont think it is
needed.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820223731>:
>
> > +    or return (0, "Couldn't read $STDERR file: $!");
> +    local $/;
> +    # Empty file with <$stderr> returns nothing in list context
> +    # (because there are no lines) Use scalar to force it to ''
> +    my $out= scalar <$stdout>;
> +    if ($out eq $bogus) {
> +        die "panic: something went wrong with runperl_and_capture";
> +    }
> +    my $err= scalar <$stderr>;
> +    if ($ENV{DEBUG_RUNENV}) {
> +        print "OUT: $out\n";
> +        print "ERR: $err\n";
> +    }
> +    return ($out, $err);
> +  } elsif (defined $pid) {                      # child
> +    for my $k (sort keys %$env) {
>
> Is the sort there for hash-randomization consistency? It'd be worth
> commenting that, since otherwise it looks redundant.
>
Sure, I'll add a comment.


>
> In t/test.pl
> <https://github.com/Perl/perl5/pull/19499#discussion_r820224216>:
>
> > +    if ($ENV{DEBUG_RUNENV}) {
> +        print "OUT: $out\n";
> +        print "ERR: $err\n";
> +    }
> +    return ($out, $err);
> +  } elsif (defined $pid) {                      # child
> +    for my $k (sort keys %$env) {
> +      $ENV{$k} = $env->{$k};
> +    }
> +    if ($ENV{DEBUG_RUNENV}) {
> +        print "Child Process $$ Executing:\n$PERL @$args\n";
> +    }
> +    open STDOUT, '>', $STDOUT or exit $FAILURE_CODE;
> +    if (open STDERR, '>', $STDERR) {
> +        exec $PERL, @$args
> +            or die "Failed to exec!";
>
> The original code falls through to the print and exit in this case. It
> isn't clear that a die() instead will find any STDERR to write to.
>
The exec only happens if we successfully dup STDERR, so the die would be
written to the file and picked up later.

Having said that, I have reworked the sub pretty significantly.


>
> In sbox32_hash.h
> <https://github.com/Perl/perl5/pull/19499#discussion_r820240481>:
>
> >      U32 s0 = seed[0] ^ 0x68736168; /* sbox */
>      U32 s1 = seed[1] ^ 0x786f6273; /* hash */
>      U32 s2 = seed[2] ^ 0x646f6f67; /* good */
> +    U32 s3 = seed[3] ^ 0x74736166; /* fast */
>
> s[0], s[1] comments are the wrong way round, FWIW.
>

That is ancient code that I didnt change in this patch. But sure, I'll fix
it.


>
> In hv.c <https://github.com/Perl/perl5/pull/19499#discussion_r820242766>:
>
> > +    PL_hash_rand_bits *= 17;                                            \
> +    PL_hash_rand_bits = ROTL_UV(PL_hash_rand_bits,11);                  \
>
> This will update 0 to 0 (which the replaced code would not), was that
> intended?
>
PL_hash_rand_bits should never get to 0. But I will adjust this logic
anyway.


>
> In hv.c <https://github.com/Perl/perl5/pull/19499#discussion_r820253271>:
>
> > +        );                                                              \
> +        if (DEBUG_v_TEST && key) {                                      \
> +            PerlIO_printf( Perl_debug_log, " key:'%.*s' %"UVuf"\n",     \
> +                    (int)klen,                                          \
> +                    key ? key : "", /* silence warning */               \
> +                    (UV)klen                                            \
> +            );                                                          \
> +        } else {                                                        \
> +            PerlIO_printf( Perl_debug_log, "\n");                       \
> +        }                                                               \
> +    }                                                                   \
> +} STMT_END
> +
> +#define UPDATE_HASH_RAND_BITS_MIX_KEY(mix,key,klen) STMT_START {        \
> +    if (PL_HASH_RAND_BITS_ENABLED == 1)                                 \
> +        PL_hash_rand_bits ^= (mix);                                     \
>
> This appears to (approximately) replace a +=, so it seems like the effect
> would now be more localized. Is that intended?
>
Yes it was.

But I have removed the bulk of this logic and replaced it with a simple
XORSHIFT. Instead of exploiting address randomization I just use seed() at
the beginning which should also fix the cygwin issue. It means a bunch of
the "poor mans" bit twiddling we discussed in a previous mail in this
thread has been removed completely and replaced by a fairly well known RNG
implementation. So no custom malarky anymore.

Please see the latest version of the patch.

cheers,
yves

hv.c
```diff
@@ -49,6 +49,52 @@ holds the key and hash value.
...
+                    (UV)klen                                            \
+            );                                                          \
+        } else {                                                        \
+            PerlIO_printf( Perl_debug_log, "\n");                       \
+        }                                                               \
+    }                                                                   \
+} STMT_END
+
+#define UPDATE_HASH_RAND_BITS()                                         \
+    UPDATE_HASH_RAND_BITS_KEY(NULL,0)
```
atoomic, line 103: should not this also check for `PL_HASH_RAND_BITS_ENABLED` 
this could also simplified some usages of `UPDATE_HASH_RAND_BITS`

hv.c
```diff
@@ -1466,11 +1510,7 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
      * second from top. After each such insert we rotate the hashed value. So we can
      * use the same hashed value over and over, and in normal build environments use
      * very few ops to do so. ROTL32() should produce a single machine operation. */
-    if (PL_HASH_RAND_BITS_ENABLED) {
-        if (PL_HASH_RAND_BITS_ENABLED == 1)
-            PL_hash_rand_bits += ptr_hash((PTRV)a);
-        PL_hash_rand_bits = ROTL_UV(PL_hash_rand_bits,1);
-    }
+    UPDATE_HASH_RAND_BITS();
```
atoomic: we have lost the `PL_HASH_RAND_BITS_ENABLED` check?

hv.c
atoomic: mm probably need to use it in `UPDATE_HASH_RAND_BITS_KEY` itself

perl.c
```diff
@@ -3375,21 +3357,31 @@ Perl_get_debug_opts(pTHX_ const char **s, bool givehelp)
...
-        /* if adding extra options, remember to update DEBUG_MASK */
-        /* Note that the ? indicates an unused slot. As the code below
+        /* NOTE:
+         * If adding new options add them to the END of debopts[].
+         * If you remove an option replace it with a '?'.
+         * If there is a free slot available marked with '?' feel
+         * free to reuse it for something else.
+         *
+         * Regardles remember to update DEBUG_MASK in perl.h, and
+         * update the documentation above AND in pod/perlrun.pod.
```
atoomic, line 3376: did you forgot to bump `DEBUG_MASK`?

atoomic (commented):
Asked a few questions during review, otherwise LGTM

perl.c
demerphq: Not needed, the mask wasn't adjusted when -DH was removed, and I repurposed its bit.

demerphq:
@atoomic: 

I have reviewed the code and I realized why I was getting confused. To me, the creator of this logic, perturb keys relates to the logic where we do not always insert a key at the top of the bucket chain, and is a part but a distinct part of hash key order randomization. In theory you should be able to keep hash key order randomization, but disable perturbing the key order in a bucket. However we only have one "knob" to control this, and it is name is PL_HASH_RAND_BITS_ENABLED and is controlled from the environment with PERL_PERTURB_KEYS.  (Which in itself is actually pretty confusing, it should probably be called PERL_HASH_KEY_RANDOMIZATION_MODE, and the env var should be similarly named. Ah the wonders of hindsight :-) ). 

You could argue that when PERL_PERTURB_KEYS=0 it should leave PL_hash_rand_bits as 0 so that you can see the native keyorder in the buckets, however historically and currently it will be initialized based on the hash seed that is in use. So while it wont change for the duration of the process, and keys still wont expose the true bucket order. 

Because of this entanglement of different, although related, concepts I got confused in my last simplification, and removed the checks. It didn't break the tests, so I didn't notice, and really with the new implementation it wouldn't have made a difference except for wasting some CPU cycles. Nevertheless with the latest revision I have preserved the older logic of determining when to update the bits, and to not do so during when PERL_PERTURB_KEYS=0.

I may put together a new patch that disentangles things, or at least changes the internally used define, but I would do it is part of a new PR and will not include any such changes in this one.

@hv:

I believe I have resolved all your concerns.  All of the bit flipping logic changes you were asking about have been replaced by clean clear well known and understand RNG updates.  I believe I have also addressed all your concerns about the perl test code changes and the changes to sbox32_hash.h as well.

Thanks both for your reviews.

atoomic:
Thanks @demerphq for these improvements and long explanations :-) 
++

demerphq:
On Thu, 10 Mar 2022 at 03:26, ℕicolas ℝ. ***@***.***> wrote:

> Thanks @demerphq <https://github.com/demerphq> for these improvements and
> long explanations :-)
> ++
>
You're welcome!

Yves

-- 

Issue #19497, opened 2022-03-05
Enh. request: Catch @{ bareword[index] } as invalid else update docs


jimav:
This may not be a bug, but rather a documentation problem.  Or a user problem...

Is `@{ bareword[index] }` intended to be valid syntax?   I couldn't find in _perlref_ an explanation of what such a construct means.

In fact it is accepted (in v5.30.0) and means an array slice, more conventionally written as
`@{arrayname}[index]` or just `@arrayname[index]`.

If the former is not correct syntax, then it would be great if perl could catch it, so as to save users from silent bugs if they have a typo like in the following:
```
use strict; use warnings;
my @aryofarys = ([]);
# ...
@{  aryofarys[0] } = (1,2,3);
#  ^typo omitting the $ sigl
```
On the other hand if `@{ name[index] }` is correct syntax, then perhaps  _perlref_ could be augmented in the "barewords" discussion to point out its intended meaning; or the "slices" discussion in _perldata_.

Grinnz:
I believe there's a previous issue discussing this syntax but can't find it

xenu:
https://github.com/Perl/perl5/issues/18800

demerphq:
On Sat, 5 Mar 2022 at 21:13, Jim Avera ***@***.***> wrote:
>
> This may not be a bug, but rather a documentation problem. Or a user problem...
>
> Is @{ bareword[index] } intended to be valid syntax? I couldn't find in perlref an explanation of what such a construct means.

Yes it is. It is the demarcated version of a variable name. Eg,
strings are demarcated, we have a symbol to show where the start and
stop. So are array subscripts and hash subscripts.  Perl allows you to
use implicitly delimited variable names, eg where you have whitespace
or other non-variable letter syntax following the variable name, but
the syntax is especially dangerous in interpolation where such
constraints are not necessarily in place. Even in normal code there
are places where you MUST delimit your variable name with a space, eg,
consider:

sub func { return "blah" }
print $fh func;

if you were to remove the space between $fh and func() it would become
a reference to a different variable $fhfunc.

This is especially troublesome in interpolation.

Demarcated variable names are the solution. You can inject a
demarcated variable into any non variable containing string without
changing the meaning of the variable name. You can also remove nearly
all the spaces from a program if you ONLY use demarcated variable
names.

Consider the string "[0]"

If I stick $array[0] at the start of that string I will turn it into
$array[0][0].

"$array[0][0]" would NOT mean $array[0] . "[0]".

[...]

Issue #18800, opened 2021-05-18
Syntax @{hash{@elems}} treated as @hash{@elems}


Grinnz:
**Description**
The syntax `@{hash{@elems}}` gets parsed as `@hash{@elems}` for some reason, and does not throw an error under strict.

**Steps to Reproduce**
```
use strict;
my %hash = (foo => 42, bar => 'baz');
print "$_\n" for @{hash{'foo','bar'}};
```

**Expected behavior**
It would be expected that `hash{@elems}` would be used as the name of the array, which would surely result in either a syntax or strict 'subs' or strict 'refs' error depending how far it gets.

**Perl configuration**
Occurs at least back to Perl 5.8.9

Grinnz:
This also seems to apply to other constructs like `${hash{foo}}`

wolfsage:
This seems like the right behaviour to me. Why do you think it's not?

Grinnz:
Can you explain? The portion inside the braces is `hash{@elems}` which doesn't seem like a valid identifier for an array to me.

wolfsage:
Is it an identifier for an array? It looks like a hash slice which is how it's behaving and how I expect it to behave.

For example, `@what` is the same as `@{what}`.

Doesn't it follow that `@what{qw(hash keys)}` would be the same as `@{what{qw(hash keys)}}` ?

Grinnz:
Why would it? `@what` is only equivalent to `@{what}` when the braces only contain a bareword; otherwise it is a block which evaluates to the name of the variable (and generally throws a strict refs error if it doesn't evaluate to an array reference).

xenu:
To make @Grinnz' point more clear, this is what `hash{@elems}` evaluates to outside of `@{}`:

```
> perl -MO=Deparse,-p -e 'sub hash; hash{@elems}'
hash({@elems});
sub hash ;
-e syntax OK
```

wolfsage:
Yeah I'm not sure I agree with this being wrong. We explicitly check for and allow this behaviour in toke.c, and even warn if it's ambiguous:
```
 Ambiguous use of %c{%s[...]} resolved to %c%s[...]
    Ambiguous use of %c{%s{...}} resolved to %c%s{...}
        (W ambiguous) You wrote something like "${foo[2]}" (where foo
        represents the name of a Perl keyword), which might be looking for
        element number 2 of the array named @foo, in which case please write
        $foo[2], or you might have meant to pass an anonymous arrayref to the
[...]

PR #19500, opened 2022-03-06
Document demarcated variable form better

demerphq:
This updated pod/perldata.pod to explain the demarcated variable form better and more completely. We have always documented it but did not specify it included array subscripts or hash subscripts and we did not document that it is the *required* form for mutli-character caret variables, and that it is the required form when doing lookups into such a variable. 

It also updated pod/perlre.pod to include references to @{^CAPTURE}, one of the notable examples of a non scalar multi-character caret variable. Previously it was only documented in perlvar.pod.

It also updates pod/perlvar.pod to include links back to pod/perldata.pod on the demarcated variable form. And a few minor enhancements to the docs on @{^CAPTURE} spelling out the necessity of using that form in interpolation. 

See GH Issue #19497 and GH Issue #18800 for examples of people being confused by this notation.

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -503,12 +503,14 @@ and POSIX::setlocale() has been called, the character used for the
 decimal separator is affected by the LC_NUMERIC locale.
 See L<perllocale> and L<POSIX>.
 
-As in some shells, you can enclose the variable name in braces to
-disambiguate it from following alphanumerics (and underscores).
-You must also do
-this when interpolating a variable into a string to separate the
-variable name from a following double-colon or an apostrophe, since
-these would be otherwise treated as a package separator:
+=head3 Demarcated variable names using braces
+
+As in some shells, you can enclose the variable name in braces as a
+demarcator to disambiguate it from following alphanumerics and
+underscores or other text. You must also do this when interpolating a
+variable into a string to separate the variable name from a following
+double-colon or an apostrophe since these would be otherwise treated as
+a package separator:
 X<interpolation>
 
     $who = "Larry";
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
 $0 and the $s variables in the (presumably) non-existent package
 C<who>.
 
-In fact, a simple identifier within such curlies is forced to be
-a string, and likewise within a hash subscript.  Neither need
-quoting.  Our earlier example, C<$days{'Feb'}> can be written as
-C<$days{Feb}> and the quotes will be assumed automatically.  But
-anything more complicated in the subscript will be interpreted as an
-expression.  This means for example that C<$version{2.0}++> is
-equivalent to C<$version{2}++>, not to C<$version{'2.0'}++>.
+In fact, a simple identifier within such curly braces is forced to be a
+string, and likewise within a hash subscript. Neither need quoting. Our
+earlier example, C<$days{'Feb'}> can be written as C<$days{Feb}> and the
+quotes will be assumed automatically. But anything more complicated in
+the subscript will be interpreted as an expression. This means for
+example that C<$version{2.0}++> is equivalent to C<$version{2}++>, not
+to C<$version{'2.0'}++>.
+
+There is a similar problem with interpolation with text that looks like
+array or hash access notation. Placing a simple variable like C<$who>
+immediately in front of text like C<"[1]"> or C<"{foo}"> would cause the
+variable to be interpolated as accessing an element of C<@who> or a
+value stored in C<%who>:
+
+    $who = "Larry Wall";
+    print "$who[1] is the father of Perl.\n";
+
+would attempt to access index 1 of an array named C<@who>. Again, using
+braces will prevent this from happening:
+
+    $who = "Larry Wall";
+    print "${who}[1] is the father of Perl.\n";
+
+will be treated the same as
+
+    $who = "Larry Wall";
+    print $who . "[1] is the father of Perl.\n";
+
+This notation also applies to more complex variable descriptions,
+such as array or hash access with subscripts. For instance
+
+    @name = qw(Larry Curly Moe);
+    print "Also ${name[0]}[1] was a member\n";
+
+Without the braces the above example would be parsed as a two level
+array subscript in the C<@name> array, and under C<use strict> would
+likely produce a fatal exception, as it would be parsed like this:
+
+    print "Also " . $name[0][1] . " was a member\n";
+
+and not as the intended:
+
+    print "Also " . $name[0] . "[1] was a member\n";
+
+A similar result may be derived by using a backslash on the first
+character of the subscript or package notation that is not part of
+the variable you want to access. Thus the above example could also
+be written:
+
+    @name = qw(Larry Curly Moe);
+    print "Also $name[0]\[1] was a member\n";
+
+however for some special variables (multi character caret variables) the
+demarcated form using curly braces is the B<only> way you can reference
+the variable at all, and the only way you can access a subscript of the
+variable via interpolation.
+
+Consider the magic array C<@{^CAPTURE}> which is populated by the
+regex engine with the contents of all of the capture buffers in a
+pattern (see L<perlvar> and L<perlre>). The B<only> way you can
+access one of these members inside of a string is via the braced
+(demarcated) form:
+
+    "abc"=~/(.)(.)(.)/
+        and print "Second buffer is ${^CAPTURE[1]}";
+
+is equivalent to
+
+    "abc"=~/(.)(.)(.)/
+        and print "Second buffer is " . ${^CAPTURE}[1];
+
+Saying C<@^CAPTURE> is a syntax error, so it B<must> be referenced as
+C<@{^CAPTURE}>, and to access one of its elements in normal code you
+would write C< ${^CAPTURE}[1] >. However when interpolating in a string
+C<"${^CAPTURE}[1]"> would be equivalent to C<${^CAPTURE} . "[1]">,
+which does not even refer to the same variable! Thus the subscripts must
+B<also> be placed B<inside> of the braces: C<"${^CAPTURE[1]}">.
+
+The demarcated form using curly braces can be used with all the
+different types of variable access, including array and hash slices. For
+instance code like the following:
+
+    @name = qw(Larry Curly Moe);
+    local $" = " and ";
+    print "My favorites were @{name[1,2]}.\n";
+
+would output
+
+    My favorites were Curly and Moe.
 
 =head3 Special floating point: infinity (Inf) and not-a-number (NaN)
 
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -1279,6 +1279,23 @@ until the end of the enclosing block or until the next successful
 match, whichever comes first.  (See L<perlsyn/"Compound Statements">.)
 X<$+> X<$^N> X<$&> X<$`> X<$'>
 X<$1> X<$2> X<$3> X<$4> X<$5> X<$6> X<$7> X<$8> X<$9>
+X<@{^CAPTURE}>
+
+The C<@{^CAPTURE}> array may be used to access ALL of the capture buffers
+as an array without needing to know how many there are. For instance
+
+    $string=~/$pattern/ and @captured = @{^CAPTURE};
+
+will place a copy of each capture variable, C<$1>, C<$2> etc, into the
+C<@captured> array.
+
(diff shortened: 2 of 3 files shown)


pod/perldata.pod
```diff
@@ -520,13 +522,94 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+
+    $who = "Larry Wall";
+    print $who, "[1] is the father of the Perl programming language\n";
+
+but
+
+    $who = "Larry Wall";
+    print "$who[1] is the father of the Perl programming language\n";
+
+would attempt to lookup index 1 in the C<@who> array.
```
karenetheridge: "lookup" is only a noun; the verb is "look up".

pod/perldata.pod
```diff
@@ -520,13 +522,94 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+        and print "Second buffer is ", ${^CAPTURE}[1];
+
+Saying C< @^CAPTURE > is a syntax error, so it B<must> be referenced as
+C< @{^CAPTURE} >, and to access one of its elements in normal code you
+would write C< ${^CAPTURE}[1] >. When interpolating in a string however
+C<"${^CAPTURE}[1]" > would be equivalent to C< ${^CAPTURE} . "[1]" >,
+which does not even refer to the same variable! Thus the subscripts must
+B<also> be placed B<inside> of the braces: C< "${^CAPTURE[1]}" >.
+
+This form applies to all the different types variable lookup, including
```
karenetheridge: lookups plural?

pod/perlvar.pod
```diff
@@ -937,7 +947,13 @@ See also L<<< /$<I<digits>> ($1, $2, ...) >>>, L</%{^CAPTURE}> and
 L</%{^CAPTURE_ALL}>.
 
 Note that unlike most other regex magic variables there is no single
-letter equivalent to C<@{^CAPTURE}>.
+letter equivalent to C<@{^CAPTURE}>. Also be aware that when
+interpolating subscripts of this array you B<must> use the demarcated
+variable form, for instance
+
+    print "@{^CAPTURE[0]}"
```
karenetheridge: Wouldn't that be `${^CAPTURE[0]}`?

pod/perlvar.pod
```diff
@@ -25,16 +25,26 @@ to hold data captured by backreferences after a regular expression
...
 
 Since Perl v5.6.0, Perl variable names may also be alphanumeric strings
-preceded by a caret.  These must all be written in the form C<${^Foo}>;
-the braces are not optional.  C<${^Foo}> denotes the scalar variable
+preceded by a caret.  These must all be written using the demarcated
+variable form using curly braces such as C<${^Foo}>;
+the braces are B<not> optional.  C<${^Foo}> denotes the scalar variable
 whose name is considered to be a control-C<F> followed by two C<o>'s.
+(See L<perldata/"Demarcated variable names using braces"> for more
+information on this form spelling a variable name or lookup).
```
karenetheridge: The wording here is weird; I'm not quite sure what it's saying.

demerphq:
On Sun, 6 Mar 2022 at 20:16, Karen Etheridge ***@***.***>
wrote:

>
> In pod/perlvar.pod
> <https://github.com/Perl/perl5/pull/19500#discussion_r820275193>:
>
> > +(See L<perldata/"Demarcated variable names using braces"> for more
> +information on this form spelling a variable name or lookup).
>
> The wording here is weird; I'm not quite sure what it's saying.
>
Responding here to all your reviews: Super useful, thanks! I'll fix all the
issues you raised.

I  was trying to differentiate between spelling a variable name by itself
and subscripted access to a composite variable, or a slice. Eg,
differentiating between $foo or %bar or @array and $bar{whatever}, or
$array[93]. Any suggestions for me?

cheers,
Yves
-- 

pod/perldata.pod
demerphq: I ditched lookup/"look up" entirely and replaced it with access. Changes pushed. Thanks!

pod/perldata.pod
demerphq: Fix pushed! Thanks.

pod/perlvar.pod
demerphq: Yep, fixed. Thanks!

pod/perlvar.pod
demerphq: Changed to:
```
See L<perldata/"Demarcated variable names using braces"> for more
information on this form of spelling a variable name or specifying
access to an element of an array or a hash
```
hopefully that makes more sense!

pod/perldata.pod
```diff
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+earlier example, C<$days{'Feb'}> can be written as C<$days{Feb}> and the
+quotes will be assumed automatically. But anything more complicated in
+the subscript will be interpreted as an expression. This means for
+example that C<$version{2.0}++> is equivalent to C<$version{2}++>, not
+to C<$version{'2.0'}++>.
+
+There is a similar problem with interpolation with text that looks like
+an array subscript or hash subscript. Placing a simple variable like
+C<$who> immediately in front of text like C<"[1]"> or C<{foo}> would
+cause the variable to be interpolated as accesing an element of C<@who>
```
khwilliamson: accessing

pod/perldata.pod
```diff
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+or a value stored in C<%who>, and again using braces will prevent this
+from happening.
+
+    $who = "Larry Wall";
```
khwilliamson, line 539: This looks like it might exceed the verbatim column limit.  You only have to indent one space to get verbatim treatment, and if won't look bad with most formatters either.

pod/perlvar.pod
```diff
@@ -25,16 +25,27 @@ to hold data captured by backreferences after a regular expression
...
 begin with C<^_> (caret-underscore).  No
 name that begins with C<^_> will acquire a special
 meaning in any future version of Perl; such names may therefore be
 used safely in programs.  C<$^_> itself, however, I<is> reserved.
 
+Note that you also B<must> use the demarcated form to access subscripts of
+variables of this type when interpolating, for instance to access the first
+element of the C< @{^CAPTURE} > variable inside of a double quoted string
+you would write C< "${^CAPTURE[0]}" > and NOT C< "${^CAPTURE}[0]" > which
+would reference the C< ${^CAPTURE} > scalar variable and not the magic
```
khwilliamson: This makes it sound like this scalar variable exists,  Perhaps
"would mean to reference a scalar variable named C< ${^CAPTURE} > and not the magic"

pod/perlre.pod
```diff
@@ -1279,6 +1279,22 @@ until the end of the enclosing block or until the next successful
 match, whichever comes first.  (See L<perlsyn/"Compound Statements">.)
 X<$+> X<$^N> X<$&> X<$`> X<$'>
 X<$1> X<$2> X<$3> X<$4> X<$5> X<$6> X<$7> X<$8> X<$9>
+C<@{^CAPTURE}>
```
khwilliamson: This should be X<>

pod/perlre.pod
```diff
@@ -1279,6 +1279,22 @@ until the end of the enclosing block or until the next successful
...
+
+The C<@{^CAPTURE}> array may be used to access ALL of the capture buffers
```
khwilliamson, line 1284: I had no idea this existed.  I could have used it over the years

pod/perlre.pod
```diff
@@ -1279,6 +1279,22 @@ until the end of the enclosing block or until the next successful
...
+as an array without needing to know how many there are. For instance
+
+    $string=~/$pattern/ and @captured = @{^CAPTURE};
+
+will place a copy of each capture variable, C<$1>, C<$2> etc, into the
+C<@captured> array without needing to know how many there are nor to
```
khwilliamson: without needing ...are 
is redundant with the line before the example, and I think should just be deleted

karenetheridge:
LGTM!

pod/perlre.pod
```diff
@@ -1279,6 +1279,22 @@ until the end of the enclosing block or until the next successful
...
+have to spell out each one independently. Note that when interpolating
```
khwilliamson: Note that ...
ishould be in a new paragraph, as it is starting a new subtobpic

pod/perldata.pod
```diff
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
 
-In fact, a simple identifier within such curlies is forced to be
-a string, and likewise within a hash subscript.  Neither need
-quoting.  Our earlier example, C<$days{'Feb'}> can be written as
-C<$days{Feb}> and the quotes will be assumed automatically.  But
-anything more complicated in the subscript will be interpreted as an
-expression.  This means for example that C<$version{2.0}++> is
-equivalent to C<$version{2}++>, not to C<$version{'2.0'}++>.
+In fact, a simple identifier within such curly braces is forced to be a
+string, and likewise within a hash subscript. Neither need quoting. Our
```
khwilliamson, line 526: hash subscript?

pod/perldata.pod
```diff
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+    $who = "Larry Wall";
+    print $who, "[1] is the father of the Perl programming language\n";
+
+but
+
+    $who = "Larry Wall";
+    print "$who[1] is the father of the Perl programming language\n";
+
+would attempt to access index 1 in the C<@who> array.
+
```
khwilliamson, line 552: It's easier to just say how I think reordering things would be, rather than describe it
`
There is a similar problem with interpolation with text that looks like
an array subscript or hash subscript. Placing a simple variable like
C<$who> immediately in front of text like C<"[1]"> or C<{foo}> would
cause the variable to be interpolated as accessing an element of C<@who>
or a value stored in C<%who>:

 $who = "Larry Wall";
 print "$who[1] is the father of the Perl programming language\n";

would attempt to access index 1 in an array named C<@who>.
Again, using braces will prevent this from happening:

 $who = "Larry Wall";
 print "${who}[1] is the father of the Perl programming language\n";

will be treated the same as

 $who = "Larry Wall";
 print $who, "[1] is the father of the Perl programming language\n";
`

pod/perldata.pod
demerphq: fixed.

pod/perldata.pod
demerphq: This is the only piece of your feedback I am not following. The line is 72 chars long with the existing 4 spaces, and passes test. I don't want this patch to be the only part that is indented only to 1 character, there are many verbatim sections in the file, it would look weird.

pod/perlvar.pod
demerphq: applied.

pod/perlre.pod
demerphq: fixed.

pod/perlre.pod
demerphq: It's not that old, but not that new either. I added it 2016 :-) But obviously not documenting it in perlre was a mistake.

pod/perlre.pod
demerphq: deleted.

pod/perldata.pod
demerphq: i think this was existing text. so i left it alone.

pod/perldata.pod
demerphq: Applied exactly as you recommended (barring possible typos while I do so).

pod/perlre.pod
demerphq: Done.

pod/perldata.pod
```diff
@@ -520,13 +522,95 @@ C<$who::0>, and a C<$who's> variable.  The last two would be the
...
+would attempt to access index 1 of an array named C<@who>. Again, using
+braces will prevent this from happening:
+
+    $who = "Larry Wall";
+    print "${who}[1] is the father of the Perl programming language\n";
+
+will be treated the same as
+
+    $who = "Larry Wall";
+    print $who, "[1] is the father of the Perl programming language\n";
```
t-a-k: Shouldn't this be `print $who . "[1] is the father of ...`?
With commas `print` will receive a list of strings, which is not the same as variable interpolations.

demerphq: Yes, of course you are right! I was thinking purely in the sense of the side-effect, but in the docs we should be more specific! Thanks. Fixed, and pushed!

Issue #19504, opened 2022-03-07
UTF8 $0 patch is producing test failures.


demerphq:
**Description**
PR #19334 changed how we handle utf8 $0. But it is failing tests.

**Steps to Reproduce**
./perl t/op/magic.t
...
```
ok 104 - compare $0 to UTF8-flagged
ok 105 - compare $0 to non-UTF8-flagged
not ok 106 - /proc cmdline shows as expected (compare to UTF8-flagged)
# Failed test 106 - /proc cmdline shows as expected (compare to UTF8-flagged) at t/op/magic.t line 476
#      got '�'
# expected /(?^u:\A�\0)/
not ok 107 - /proc cmdline shows as expected (compare to non-UTF8-flagged)
# Failed test 107 - /proc cmdline shows as expected (compare to non-UTF8-flagged) at t/op/magic.t line 477
#      got '�'
# expected /(?^:\A�\0)/
ok 108 - warning after assignment of wide character
ok 109 - .. and the warning is about a wide character
ok 110 - .. and the UTF-8 version is written
not ok 111 - .. and /proc cmdline shows that
# Failed test 111 - .. and /proc cmdline shows that at t/op/magic.t line 497
#      got "hahaĀhoho"
# expected "hahaĀhoho\000"
ok 112 - $0 from wide -> local non-wide: no warning
ok 113 - one warning: wide -> local wide
ok 114 - one warning: non-wide -> local wide
```
**Expected behavior**
No test failures. 
**Perl configuration**
```
$ ./perl -Ilib -V
Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: 7ea8b04b5a0e6952b7ffd5a8fd96468b72da6bea
  Platform:
    osname=linux
    osvers=4.15.0-166-generic
    archname=x86_64-linux-thread-multi
    uname='linux psykopsis 4.15.0-166-generic #174-ubuntu smp wed dec 8 19:07:44 utc 2021 x86_64 x86_64 x86_64 gnulinux '
    config_args='-Dusethreads -Doptimize=-g -d -Dusedevel -Dcc=ccache gcc -Dld=gcc -DDEBUGGING'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
[...]

PR #19505, opened 2022-03-07
t/op/magic.t - Issue #19504 - fixup tests for utf8 $0

demerphq:
Since PR #19334 was merged the tests for it are failing. This should fix
them. I am taking the perspective based on the broken previous tests
that that on some systems /proc/$$/commandline returns a trailing null.
This fixes the code to read /proc/$$/comandline to chop the null off if
it is there so the tests can be agnostic to that.

At the same time I have fixed the tests a bit to not use regexen where
they aren't needed and to avoid a test count issue in the skip logic,
and also to produce better errors if the code to read
/proc/$$/commandline fails.

### DIFF
--- a/t/op/magic.t
+++ b/t/op/magic.t
@@ -465,16 +465,22 @@ EOP
   is ($0, $char_with_plain_pv, 'compare $0 to non-UTF8-flagged');
 
   my $linux_cmdline_cr = sub {
-    open my $rfh, '<', "/proc/$$/cmdline" or skip "open: $!", 1;
-    return do { local $/; <$rfh> };
+    my $skip = shift // 1;
+    open my $rfh, '<', "/proc/$$/cmdline"
+      or skip "failed to read '/proc/$$/cmdline': $!", $skip;
+    my $got = do { local $/; <$rfh> };
+    $got=~s/\0\z//;
+    return $got;
   };
 
   SKIP: {
-    skip "Test is for Linux, not $^O", 2 if $^O ne 'linux';
-
-    my $slurp = $linux_cmdline_cr->();
-    like( $slurp, qr<\A$char_with_utf8_pv\0>, '/proc cmdline shows as expected (compare to UTF8-flagged)' );
-    like( $slurp, qr<\A$char_with_plain_pv\0>, '/proc cmdline shows as expected (compare to non-UTF8-flagged)' );
+    my $skip_tests = 2;
+    skip "Test is for Linux, not $^O", $skip_tests if $^O ne 'linux';
+    my $slurp = $linux_cmdline_cr->($skip_tests);
+    is( $slurp, $char_with_utf8_pv,
+        '/proc cmdline shows as expected (compare to UTF8-flagged)' );
+    is( $slurp, $char_with_plain_pv,
+        '/proc cmdline shows as expected (compare to non-UTF8-flagged)' );
   }
 
   my $name_unicode = "haha\x{100}hoho";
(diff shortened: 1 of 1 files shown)


t/op/magic.t
```diff
@@ -465,16 +465,21 @@ EOP
   is ($0, $char_with_plain_pv, 'compare $0 to non-UTF8-flagged');
 
   my $linux_cmdline_cr = sub {
-    open my $rfh, '<', "/proc/$$/cmdline" or skip "open: $!", 1;
-    return do { local $/; <$rfh> };
+    my $skip= shift // 1;
```
leonerd: Needs whitespace fix - ` = ` on both sides of `=`

t/op/magic.t
```diff
@@ -465,16 +465,21 @@ EOP
...
+    open my $rfh, '<', "/proc/$$/cmdline"
+      or skip "failed to read '/proc/$$/cmdline': $!", $skip;
+    my $got= do { local $/; <$rfh> };
```
leonerd: ditto

leonerd (changes requested):

demerphq:
On Mon, 7 Mar 2022 at 14:52, Paul Evans ***@***.***> wrote:

>
> In t/op/magic.t
> <https://github.com/Perl/perl5/pull/19505#discussion_r820719480>:
>
> > @@ -465,16 +465,21 @@ EOP
>    is ($0, $char_with_plain_pv, 'compare $0 to non-UTF8-flagged');
>
>    my $linux_cmdline_cr = sub {
> -    open my $rfh, '<', "/proc/$$/cmdline" or skip "open: $!", 1;
> -    return do { local $/; <$rfh> };
> +    my $skip= shift // 1;
>
> Needs whitespace fix - = on both sides of =
>
> In t/op/magic.t
> <https://github.com/Perl/perl5/pull/19505#discussion_r820720747>:
>
> > @@ -465,16 +465,21 @@ EOP
>    is ($0, $char_with_plain_pv, 'compare $0 to non-UTF8-flagged');
>
>    my $linux_cmdline_cr = sub {
> -    open my $rfh, '<', "/proc/$$/cmdline" or skip "open: $!", 1;
> -    return do { local $/; <$rfh> };
> +    my $skip= shift // 1;
> +    open my $rfh, '<', "/proc/$$/cmdline"
> +      or skip "failed to read '/proc/$$/cmdline': $!", $skip;
> +    my $got= do { local $/; <$rfh> };
>
> ditto
>
>
> Those aren't bugs. I left hug my assignment operators, and since we dont
have an approved style it shouldn't matter.

But i pushed a fix to match the style in the file to get this moving.

Yves

t/op/magic.t
```diff
@@ -465,16 +465,22 @@ EOP
   is ($0, $char_with_plain_pv, 'compare $0 to non-UTF8-flagged');
 
   my $linux_cmdline_cr = sub {
-    open my $rfh, '<', "/proc/$$/cmdline" or skip "open: $!", 1;
-    return do { local $/; <$rfh> };
+    my $skip = shift // 1;
+    open my $rfh, '<', "/proc/$$/cmdline"
+      or skip "failed to read '/proc/$$/cmdline': $!", $skip;
+    my $got = do { local $/; <$rfh> };
+    $got=~s/\0\z//;
```
leonerd, line 472: FYI I'd probably write this as

```perl
return do { local $/; <$rfh> } =~ s/\0\z//r;
```

these days. It's a core perl unit test so we can rely on having 5.14+ features now. ;)

leonerd (approved):
LGTM

t/op/magic.t
demerphq: Old habits die hard. :-)

PR #19508, opened 2022-03-08
Implement `builtin::indexed` (RFC 0016)

leonerd:
As per https://github.com/Perl/RFCs/blob/main/rfcs/rfc0016.md

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -224,7 +224,7 @@ XS(XS_builtin_import)
 
     if(!PL_compcv)
         Perl_croak(aTHX_
-                "builtin::import can only be called at compiletime");
+                "builtin::import can only be called at compile time");
 
     /* We need to have PL_comppad / PL_curpad set correctly for lexical importing */
     ENTER;
--- a/builtin.c
+++ b/builtin.c
@@ -197,6 +197,55 @@ static OP *ck_builtin_func1(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
     return newUNOP(opcode, wantflags, argop);
 }
 
+XS(XS_builtin_indexed)
+{
+    dXSARGS;
+
+    switch(GIMME_V) {
+        case G_VOID:
+            Perl_ck_warner(aTHX_ packWARN(WARN_VOID),
+                "Useless use of %s in void context", "builtin::indexed");
+            XSRETURN(0);
+
+        case G_SCALAR:
+            Perl_ck_warner(aTHX_ packWARN(WARN_SCALAR),
+                "Useless use of %s in scalar context", "builtin::indexed");
+            ST(0) = sv_2mortal(newSViv(items * 2));
+            XSRETURN(1);
+
+        case G_LIST:
+            break;
+    }
+
+    SSize_t retcount = items * 2;
+    EXTEND(SP, retcount);
+
+    /* Copy from [items-1] down to [0] so we don't have to make
+     * temporary copies */
+    for(SSize_t index = items - 1; index >= 0; index--) {
+        /* Copy, not alias */
+        ST(index * 2 + 1) = sv_mortalcopy(ST(index));
+        ST(index * 2)     = sv_2mortal(newSViv(index));
+    }
+
+    XSRETURN(retcount);
+}
+
+static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
+{
+    const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
+
+    warn_experimental_builtin(builtin->name, false);
+
+    SV *prototype = newSVpvs("@");
+    SAVEFREESV(prototype);
+
+    assert(entersubop->op_type == OP_ENTERSUB);
+
+    entersubop = ck_entersub_args_proto(entersubop, namegv, prototype);
+    return entersubop;
+}
+
 static const char builtin_not_recognised[] = "'%" SVf "' is not recognised as a builtin function";
 
 static const struct BuiltinFuncDescriptor builtins[] = {
@@ -214,6 +263,9 @@ static const struct BuiltinFuncDescriptor builtins[] = {
     { "builtin::reftype",  &XS_builtin_func1_scalar, &ck_builtin_func1, OP_REFTYPE  },
     { "builtin::ceil",     &XS_builtin_func1_scalar, &ck_builtin_func1, OP_CEIL     },
     { "builtin::floor",    &XS_builtin_func1_scalar, &ck_builtin_func1, OP_FLOOR    },
+
+    /* list functions */
+    { "builtin::indexed", &XS_builtin_indexed, &ck_builtin_funcN, 0 },
     { 0 }
 };
 
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,4 +1,4 @@
-package builtin 0.003;
+package builtin 0.004;
 
 use strict;
 use warnings;
@@ -167,6 +167,34 @@ numerical argument.
 
 Returns the largest integer value less than or equal to the given numerical
 argument.
+
+=head2 indexed
+
+    @ivpairs = indexed(@items)
+
+Returns an even-sized list of number/value pairs, where each pair is formed
+of a number giving an index in the original list followed by the value at that
+position in it.  I.e. returns a list twice the size of the original, being
+equal to
+
+    (0, $items[0], 1, $items[1], 2, $items[2], ...)
+
+Note that unlike the core C<values> function, this function returns copies of
+its original arguments, not aliases to them. Any modifications of these copies
+are I<not> reflected in modifications to the original.
+
+    my @x = ...;
+    $_++ for indexed @x;  # The @x array remains unaffected
+
+This function is primarily intended to be useful combined with multi-variable
+C<foreach> loop syntax; as
+
+    foreach my ($index, $value) (indexed LIST) {
+        ...
+    }
+
+In scalar context this function returns the size of the list that it would
+otherwise have returned, and provokes a warning in the C<scalar> category.
 
 =head1 SEE ALSO
 
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -645,6 +645,13 @@ C<builtin::> namespace, which is currently experimental. The existence
 or nature of the function may be subject to change in a future version
 of Perl.
 
+=item builtin::import can only be called at compile time
+
+(F) The C<import> method of the C<builtin> package was invoked when no code
+is currently being compiled. Since this method is used to introduce new
+lexical subroutines into the scope currently being compiled, this is not
+going to have any effect.
+
 =item Callback called exit
 
 (F) A subroutine invoked from an external package via call_sv()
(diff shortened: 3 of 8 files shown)


lib/builtin.pm
```diff
@@ -168,6 +168,31 @@ numerical argument.
...
+=head2 indexed
+
+    @kvlist = indexed(@list)
+
+Returns an even-sized list of number/value pairs, formed by taking successive
+values from the list of arguments and prefixing each one with an integer
+giving its position within the original list. I.e. returns a list twice the
+size of the original, being equal to
+
+    (0, $list[0], 1, $list[2], 2, $list[2], ...)
```
wolfsage: -(0, $list[0], 1, $list[2], 2, $list[2], ...)
    +(0, $list[0], 1, $list[1], 2, $list[2], ...)

lib/builtin.pm
```diff
@@ -168,6 +168,31 @@ numerical argument.
 Returns the largest integer value less than or equal to the given numerical
 argument.
 
+=head2 indexed
+
+    @kvlist = indexed(@list)
+
+Returns an even-sized list of number/value pairs, formed by taking successive
+values from the list of arguments and prefixing each one with an integer
```
wolfsage: prefixing sounds to me like "1value" "2value". I think this whole paragraph is kind of overly confusing. It doesn't  mention array indices which I think are well understood and might make this clearer?


```
Returns an even-sized list of index/value pairs for the given list, in order. For example:

    my @list = qw(first second third);
    my @indexed = indexed @list; # (0, first, 1, second, 2, third);

File: lib/builtin.pm [hunk 3]
```diff
@@ -168,6 +168,31 @@ numerical argument.
 Returns the largest integer value less than or equal to the given numerical
 argument.
 
+=head2 indexed
+
+    @kvlist = indexed(@list)
```
demerphq: Calling an @array a "list" just propagates the list/array confusion. This should not use the word list on either side.

   @iv = indexed(@array);
   @iv = indexed("foo","bar","baz","bop");
   
 would be better IMO. Or even
 
   @idx_val = indexed(@array);
   @idx_val = indexed("foo","bar","baz","bop");
  
since this is a doc.

lib/builtin.pm
```diff
@@ -168,6 +168,31 @@ numerical argument.
...
+
+Returns an even-sized list of number/value pairs, formed by taking successive
+values from the list of arguments and prefixing each one with an integer
+giving its position within the original list. I.e. returns a list twice the
+size of the original, being equal to
+
+    (0, $list[0], 1, $list[1], 2, $list[2], ...)
```
demerphq: Same problem here.

wolfsage:
I didn't see aliased vs copied mentioned in the RFC, but I find this being copies instead of aliasing surprising personally

lib/builtin.pm
```diff
@@ -168,6 +168,31 @@ numerical argument.
...
+its original arguments, not aliases to them. Any modifications of these copies
+are I<not> reflected in modifications to the original.
+
+    my @x = ...;
+    $_++ for indexed @x;  # The @x array remains unaffected
+
+This function is primarily intended to be useful combined with multi-variable
+C<foreach> loop syntax; as
+
+    foreach my ( $index, $value ) ( indexed LIST ) {
```
demerphq: here it DOES makes sense, since its grammar.

demerphq (changes requested):
I really don't think our docs should be calling an 'array' @list. We have enough confusion related to the difference between lists and arrays, the docs shouldn't make it worse.

builtin.c
```diff
@@ -197,6 +197,55 @@ static OP *ck_builtin_func1(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
...
+
+    switch(GIMME_V) {
+        case G_VOID:
+            Perl_ck_warner(aTHX_ packWARN(WARN_VOID),
+                "Useless use of builtin::indexed in void context");
+            XSRETURN(0);
+            break;
+        case G_SCALAR:
+            Perl_ck_warner(aTHX_ packWARN(WARN_SCALAR),
+                "Useless use of builtin::indexed in scalar context");
```
ilmari: The existing `Useless use of sort in scalar context` warning is in the `void` category. Should that be changed?

Also, the `perldiag` entry should be updated to `Useless use of %s in scalar context` and the wording adjusted accordingly.

ilmari (changes requested):

leonerd:
@wolfsage 
> I didn't see aliased vs copied mentioned in the RFC, but I find this being copies instead of aliasing surprising personally

The RFC says:

> All values are copies, unlike `values ARRAY`.

builtin.c
leonerd: > The existing Useless use of sort in scalar context warning is in the void category. Should that be changed?

Possibly. Actually I wonder about splitting that out into its own PR; adding a "scalar" category and moving the existing one first, then rebase this PR atop that.

> Also, the perldiag entry should be updated to Useless use of %s in scalar context and the wording adjusted accordingly.

I expected it would but it didn't fail for me locally so I left it alone. It seems the smokers have found it though.

leonerd: > > The existing Useless use of sort in scalar context warning is in the void category. Should that be changed?
> 
> Possibly. Actually I wonder about splitting that out into its own PR; adding a "scalar" category and moving the existing one first, then rebase this PR atop that.

Done https://github.com/Perl/perl5/pull/19510

wolfsage:
> @wolfsage
> 
> > I didn't see aliased vs copied mentioned in the RFC, but I find this being copies instead of aliasing surprising personally
> 
> The RFC says:
> 
> > All values are copies, unlike `values ARRAY`.

Ah... okay. Still weird but wfm

leonerd:
@wolfsage
> Ah... okay. Still weird but wfm

I think it's reasonable. Perhaps sometime off in the future when we work out how to apply attributes to function calls we could expand this by doing

```perl
for my ($idx, $val) (indexed :alias @array) { ... }
```

Or

```perl
for my ($idx, $val) ((:alias) indexed @array) { ... }
```

Or somesuch spelling. But overall I think it's best to be "safest by default", i.e. copies, and let people opt-in if they really want aliases once it becomes possible.

leonerd:
PTAL. Changes made and rebased atop blead which now has 'scalar' warning category already.

lib/builtin.pm
```diff
@@ -168,6 +168,31 @@ numerical argument.
...
+Returns an even-sized list of number/value pairs, where each pair is formed
+of a number giving an index in the original list followed by the value at that
+position in it.  I.e. returns a list twice the size of the original, being
+equal to
+
+    (0, $item[0], 1, $item[1], 2, $item[2], ...)
+
+Note that unlike the core C<values> function, this function returns copies of
+its original arguments, not aliases to them. Any modifications of these copies
+are I<not> reflected in modifications to the original.
```
wolfsage, line 184: I wonder if we need to say more here. For example, this will have lasting effects:
```
ref($_) && push(@$_, "foo") for indexed(@l);
```

wolfsage:
We could also maybe mention that `indexed %hash` is perfectly valid, as is some of the other cases that you're testing for like `indexed "a".."z"`. The docs make it seem very `@array` centric...

This might mostly be doable with a bunch of simple examples at the end of the document, like:

```
Some examples:

@ivpairs = indexed 'a'..'c';      # 0, a, 1, b, 2, c
@ivpairs = indexed "a", "b", "c"; # 0, a, 1, b, 2, c
@ivpairs = indexed %hash;         # 0, key1, val1, 1, key2, val2, ...
```

neilb:
The doc says:

    @ivpairs = indexed(@items)

And then says this is equivalent to:

    (0, $item[0], 1, $item[1], 2, $item[2], ...)

But that should be `$items[0]`, `$items[1]`, etc (i.e. "items" and not "item").

leonerd:
> @ivpairs = indexed %hash;         # 0, key1, val1, 1, key2, val2, ...

Ah but in that specific case it won't be. It'll be

```
@ivpairs = indexed %hash;  # 0, key1, 1, val1, 2, key2, 3, val3, ...
```

i.e. unlikely to be useful to anybody. I don't feel it particularly useful to document the behaviour on a hash - it's just a natural consequence of hash in list context, and `indexed` on a list.

pod/perldiag.pod
```diff
@@ -645,6 +645,13 @@ C<builtin::> namespace, which is currently experimental. The existence
 or nature of the function may be subject to change in a future version
 of Perl.
 
+=item builtin::import can only be called at compile time
+
+(F) The C<import> method of the C<builtin> package was invoked when no code
+is currently being compiled. Since this method is used to introduce new
+lexical subroutines into the code currently being compiled, this is not going
```
ilmari: `s/code/scope/`?

builtin.c
```diff
@@ -197,6 +197,55 @@ static OP *ck_builtin_func1(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
...
+    switch(GIMME_V) {
+        case G_VOID:
+            Perl_ck_warner(aTHX_ packWARN(WARN_VOID),
+                "Useless use of %s in void context", "builtin::indexed");
+            XSRETURN(0);
+            break;
+        case G_SCALAR:
+            Perl_ck_warner(aTHX_ packWARN(WARN_SCALAR),
+                "Useless use of %s in scalar context", "builtin::indexed");
+            mXPUSHi(items);
```
ilmari: Shouldn't this return the length of the would-be-returned list, i.e. `items * 2`?

wolfsage:
> > @ivpairs = indexed %hash;         # 0, key1, val1, 1, key2, val2, ...
> 
> Ah but in that specific case it won't be. It'll be
> 
> ```
> @ivpairs = indexed %hash;  # 0, key1, 1, val1, 2, key2, 3, val3, ...
> ```

This is what I get for testing it, and then writing the comment by hand instead of copying my output....

> i.e. unlikely to be useful to anybody. I don't feel it particularly useful to document the behaviour on a hash - it's just a natural consequence of hash in list context, and `indexed` on a list.

Questions always arise, sometimes it's nice to head them off, but I'm okay if we don't document it.

lib/builtin.pm
demerphq: personally i wouldn't consider that a modification of the argument. if you put water in a glass have you modified the glass?

wolfsage: It's "obvious" that modifying a reference will modify a reference, but at the same time "Any modifications of these copies
are I<not> reflected in modifications to the original." seems very much like "Your data will be unchanged".

People *will* be surprised.

builtin.c
leonerd: Oooh, probably yes.

Re-pushed.

ilmari: Do we want a test for it as well?

leonerd: Huh, yes. Good call as it turns out it wasn't correctly implemented anyway. Now fixed, along with test and docs - 9666de1f11

ilmari (approved):

leonerd:
Rebased and squashed a little further since ilmari's review.

@demerphq Does this look good to you now?

demerphq (approved):
LGTM. There is  a use of @list in the tests still, but that is nitpicking :-) The docs look good to me. Thanks!

leonerd:
> LGTM. There is a use of @list in the tests still, but that is nitpicking :-)

Ahyes, I'll just fix those up before I merge.

PR #19509, opened 2022-03-08
Increase NV precision in sv_dump and Devel::Peek::Dump

t-a-k:
I think it will be better that `Devel::Peek::Dump` dump NVs to enough precision to distinguish any adjacent floating point numbers.

Previously, some LSBs in floating point numbers were lost in `Dump` output, so very close (but numerically not equal) floating point numbers (e.g. 1.0 and 1.0000000000000003 if NV is "double") might produce indistinguishable dumps.

### DIFF
--- a/dump.c
+++ b/dump.c
@@ -1747,6 +1747,33 @@ const struct flag_to_name regexp_core_intflags_names[] = {
     {PREGf_ANCH_GPOS,       "ANCH_GPOS,"},
 };
 
+/* Minimum number of decimal digits to preserve the significand of NV.  */
+#ifdef USE_LONG_DOUBLE
+#  ifdef LDBL_DECIMAL_DIG
+#    define NV_DECIMAL_DIG      LDBL_DECIMAL_DIG
+#  endif
+#elif defined(USE_QUADMATH) && defined(I_QUADMATH)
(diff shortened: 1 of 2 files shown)


demerphq (approved):
LGTM. Would it be helpful to show the hex form also?

atoomic (approved):

demerphq:
Merged. Thanks!

t-a-k:
Thanks for merging!

> Would it be helpful to show the hex form also?

I agree, but I'm afraid that adding the hex form might break existing tests and programs (including `Peek.t`, which seems to expect the `NV =` line to contain a single decimal number only), so I think the hex form is better to be enabled only by some opt-in method.

demerphq:
On Thu, 10 Mar 2022 at 09:58, TAKAI Kousuke ***@***.***>
wrote:

> Thanks for merging!
>
> Would it be helpful to show the hex form also?
>
> I agree, but I'm afraid that adding the hex form might break existing
> tests and programs (including Peek.t, which seems to expect the NV = line
> to contain a single decimal number only), so I think the hex form is better
> to be enabled only by some opt-in method.
>
I was thinking of it as a separate line. Nothing wrong with changing the
tests but if it's too much effort I understand.

Cheers,
Yves

-- 

PR #19514, opened 2022-03-10
Add a configure probe for setenv() and enable PERL_USE_SAFE_PUTENV on the platforms that have it

### LINKED ISSUE #19399: Should build using `-DPERL_USE_SAFE_PUTENV` on Linux
This compiletime flag affects the way perl interacts with the `putenv()` libc call.

Without it, subtle memory corruption happens if both perl and some other C library attempt to `putenv()` within the same program. This corruption is only detected at destruction time, and only on thready+debugging builds.

An example of this can be seen in https://rt.cpan.org/Ticket/Display.html?id=140138

I believe the solution is simply that Perl should build using this flag all the time.

xenu:
Fixes #19399


xenu:
PS. Apart from the usual platforms, I've tested this change on AIX and HP-UX. There were no problems.

leonerd:
Looks like the test failure was due to the trailing NUL byte in `$0`, which we fixed a while ago.

Can you rebase this on current blead and try again?

It's too close to a 5.36 release to put this in now, but it would be nice to get it lined up ready to start in 5.37.1.

demerphq:
See https://github.com/Perl/perl5/issues/19504#issuecomment-1061325725 for details on the null byte issue with $0.

xenu:
>Looks like the test failure was due to the trailing NUL byte in $0, which we fixed a while ago.

The new test has actually exposed a real problem in my branch. The cause of the failure was that `$0` was getting truncated. Perl doesn't overwrite `environ` when `PERL_USE_SAFE_PUTENV` is defined, and because of that, the maximum length of `$0` was limited to the length of `argv`.

I have updated the branch with a more comprehensive fix and also some refactoring, see the commit messages for more details. The non-`PERL_USE_SAFE_PUTENV` code path was completely removed.

Tux (commented):
For the `Configure` part, I see that this probe already exists upstream, but has not been included because it isnt/wasnt used. If I would add it it `metaconfig.h` and regenerate `Configure` and friends, it'll include the original documentation too.

Tux:
FWIW, here is the part of the original code that is already available:
```
?S:d_setenv:
?S:     This variable conditionally defines the HAS_SETENV
?S:     symbol, which indicates to the C program that setenv()
?S:     is available to change or add an environment variable.
?S:.
?C:HAS_SETENV:
?C:     This symbol is defined when setenv() is available to change or
?C:     add an environment variable.
?C:.
?H:#$d_setenv HAS_SETENV
?H:.
?LINT: set d_setenv
: do we have setenv?
$cat >try.c <<EOC
#$i_stdlib I_STDLIB
#ifdef I_STDLIB
#include <stdlib.h>
#endif
int main (void) {
    return setenv ("foo", "bar", 1);
    }
EOC
cyn=setenv
set d_setenv
eval $trylink
```
(somewhat modified to fit)
Note that this actually treis to compile its use instead of just checking for its existence.

xenu:
I don't quite understand. As far as I can tell, there are no `setenv` probes in the Perl's metaconfig fork. Do you mean that the probe should be copied from the upstream dist repository?

Tux:
There are no `setenv` probes in the perl-part of `metaconfig`, but there is one in the upstream part of what is used for meta/dist
If the above code (also) works for you, it is easier for me to add that probe than to add a new probe that differs.

xenu:
Sure, that works too.

PR #19521, opened 2022-03-11
whitespace fixups for "define\t"

demerphq:
'#define' has seven characters, so following it with a \t makes it look
like '#define ' when it is not, which then frustrates attempts to find
where a given define is. If you *know* then you do a

    git grep -P 'define\s+WHATEVER'

but if don't or you forget, you can get very confused trying to find
where a given define is located. This fixes all such cases so they
actually are 'define WHATEVER' instead.

View this patch with -w and you will see "no changes".

If this patch is getting your way with blame analysis use the -w option.


haarg (approved):

Tux:
Given `config_h.SH` is a generated file, it requires a different approach. I think for the `Configure` part it boils down to
Two changes in Metaconfig:
```
diff --git a/U/modified/Config_h.U b/U/modified/Config_h.U
index 3558da8..254c495 100644
--- a/U/modified/Config_h.U
+++ b/U/modified/Config_h.U
@@ -89,7 +89,7 @@ echo "Extracting $CONFIG_H (with variable substitutions)"
 ?X:    This sed command from Hallvard B Furuseth <h.b.furuseth@usit.uio.no>
 ?X:    changes it to
 ?X:            /*#define FOO / **/
-sed <<!GROK!THIS! >$CONFIG_H -e 's!^#undef\(.*/\)\*!/\*#define\1 \*!' -e 's!^#un-def!#undef!'
+sed <<!GROK!THIS! >$CONFIG_H -e 's!^#undef\(.*/\)\*!/\*#define\1 \*!' -e 's!^#un-def!#undef!' -e's!define\t!define !'
 /* This file was produced by running the config_h.SH script, which
  * gets its values from $CONFIG_SH, which is generally produced by
  * running Configure.
diff --git a/U/protos/Protochk.U b/U/protos/Protochk.U
index d6f6e6c..e64e196 100644
--- a/U/protos/Protochk.U
+++ b/U/protos/Protochk.U
@@ -93,7 +93,7 @@ while test $# -ge 2; do
     shift 2
 done
 cat >> try.c <<'EOCP'
-#define        _(args) args
+#define _(args) args
 EOCP
 echo "$foo" >> try.c
 ?X: Just so we have something to compile.
```
and one in core:
```
diff --git a/Porting/config_h.pl b/Porting/config_h.pl
index 935834990d..360ff4799c 100755
--- a/Porting/config_h.pl
+++ b/Porting/config_h.pl
@@ -78,6 +78,8 @@ unless ($ch[0] =~ m/THIS IS A GENERATED FILE/) {
     push @ch, ";;\nesac\n";
     }

+s/^\s*#\s*define\t\s*/#define /gm for @ch;
+
 open  $ch, ">", $cSH or die "Cannot write $cSH: $!\n";
 print $ch @ch;
 close $ch;
```

jkeenan:
This merge request has languished since March 2022 and has developed merge conflicts. It received one Approval but then attracted dissents.  @demerphq, if you would like to move this request forward, could you resolve the merge conflicts?  (Otherwise, I think we should close the ticket.) Thanks.

Tux:
I have just pushed my side of the story.
Please verify if this does what the people who care wanted as a result.
If so, I am happy and we can close this issue

PR #19524, opened 2022-03-11
Add `builtin::created_as_...` funcs (draft RFC 0017)

leonerd:
As per RFC in https://github.com/Perl/RFCs/tree/main/rfcs/rfc0017.md

### DIFF
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -20,6 +20,7 @@ builtin - Perl pragma to import built-in utility functions
         true false is_bool
         weaken unweaken is_weak
         blessed refaddr reftype
+        created_as_string created_as_number
         ceil floor
         trim
     );
@@ -154,6 +155,54 @@ a means to test for referential identity or uniqueness.
 Returns the basic container type of the referent of a reference, or C<undef>
 for a non-reference. This is returned as a string in all-capitals, such as
 C<ARRAY> for array references, or C<HASH> for hash references.
+
+=head2 created_as_string
+
+    $bool = created_as_string($val);
+
+Returns a boolean representing if the argument value was originally created as
+a string. It will return true for any scalar expression whose most recent
+assignment or modification was of a string-like nature - such as assignment
+from a string literal, or the result of a string operation such as
+concatenation or regexp. It will return false for references (including any
+object), numbers, booleans and undef.
+
+It is unlikely that you will want to use this for regular data validation
+within Perl, as it will not return true for regular numbers that are still
+perfectly usable as strings, nor for any object reference - especially objects
+that overload the stringification operator in an attempt to behave more like
+strings. For example
+
+    my $val = URI->new( "https://metacpan.org/" );
+
+    if( created_as_string $val ) { ... }    # this will not execute
+
+=head2 created_as_number
+
+    $bool = created_as_number($val);
+
+Returns a boolean representing if the argument value was originally created as
+a number. It will return true for any scalar expression whose most recent
+assignment or modification was of a numerical nature - such as assignment from
+a number literal, or the result of a numerical operation such as addition. It
+will return false for references (including any object), strings, booleans and
+undef.
+
+It is unlikely that you will want to use this for regular data validation
+within Perl, as it will not return true for regular strings of decimal digits
+that are still perfectly usable as numbers, nor for any object reference -
+especially objects that overload the numification operator in an attempt to
+behave more like numbers. For example
+
+    my $val = Math::BigInt->new( 123 );
+
+    if( created_as_number $val ) { ... }    # this will not execute
+
+While most Perl code should operate on scalar values without needing to know
+their creation history, these two functions are intended to be used by data
+serialisation modules such as JSON encoders or similar situations, where
+language interoperability concerns require making a distinction between values
+that are fundamentally stringlike versus numberlike in nature.
 
 =head2 ceil
 
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,4 +1,4 @@
-package builtin 0.004;
+package builtin 0.005;
 
 use strict;
 use warnings;
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -94,6 +94,73 @@ package FetchStoreCounter {
     is(blessed(bless [], "0") ? "YES" : "NO", "NO", 'blessed in boolean context handles "0" cornercase');
 }
 
+# created_as_...
+{
+    use builtin qw( created_as_string created_as_number );
+
+    # some literal constants
+    ok(!created_as_string(undef), 'undef created as !string');
+    ok(!created_as_number(undef), 'undef created as !number');
+
+    ok( created_as_string("abc"), 'abc created as string');
+    ok(!created_as_number("abc"), 'abc created as number');
+
+    ok(!created_as_string(123),   '123 created as !string');
+    ok( created_as_number(123),   '123 created as !number');
+
+    ok(!created_as_string(1.23),   '1.23 created as !string');
+    ok( created_as_number(1.23),   '1.23 created as !number');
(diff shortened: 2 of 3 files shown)


lib/builtin.t
```diff
@@ -94,6 +94,58 @@ package FetchStoreCounter {
...
+    ok( created_as_string($just_str), 'def created as string');
+    ok(!created_as_number($just_str), 'def created as number');
+
+    my $just_num = 456;
+    ok(!created_as_string($just_num), '456 created as string');
+    ok( created_as_number($just_num), '456 created as number');
+
+    # variables reused
+    my $originally_str = "1";
+    my $str_as_num = $originally_str + 0;
```
ilmari: We should test that this one is actually a num too. And vice versa for `$num_as_str` below.

leonerd: Added

rwp0:
I think was_ is better

haarg:
> I think was_ is better

Can you elaborate?

lib/builtin.t
```diff
@@ -94,6 +94,62 @@ package FetchStoreCounter {
...
+    ok( created_as_number($originally_num), 'IV reused as PV created as number');
+    ok( created_as_string($num_as_str), 'New string from IV created as string');
+    ok(!created_as_number($num_as_str), 'New string from IV created as !number');
+
+    # magic
+    local $1;
+    "hello" =~ m/(.*)/;
+    ok(created_as_string($1), 'magic string');
+}
+
```
wolfsage, line 163: How do we expect this to behave? Right now it's somewhat surprising:
```
$ cat ex.pl 
#!./perl -I lib

use strict;
use warnings;

use builtin qw(created_as_number created_as_string);
no warnings qw(experimental);

my $x = 2**10000000;

if (created_as_number($x)) {
  print "created as number\n";
}

if (created_as_string($x)) {
  print "created as string\n";
}

my $z = "$x";

if (created_as_number($x)) {
  print "created as number\n";
}

if (created_as_string($x)) {
  print "created as string\n";
}
```
```
created as number
created as string
```

Here's one case where using the value as a string modifies it behind the scenes

rwp0:
> > I think was_ is better
> 
> Can you elaborate?

I mean naming the functions `was_string` and `was_number` instead of `created_as` to match the ones with `is_`

khwilliamson:
On 3/14/22 13:32, Elvin Aslanov wrote:
>         I think was_ is better
> 
>     Can you elaborate?
> 
> I mean naming the functions |was_string| and |was_number| instead of 
> |creates_as| to match the ones with |is_|
> 

"was_" implies past tense, something that no longer is true.  And these 
well could still be true.  created_as or born_as are more accurate

lib/builtin.t
haarg: This seems like a bug. @nwc10 could you take a look at this? It looks like the branches in  `Perl_sv_2pv_flags` for Inf/NaN or when USE_LOCALE_NUMERIC is not defined are still setting SvPOK_on.

leonerd: Ooooh oops, that looks like it might be a bug :/ That's supposed to remain numeric

rwp0:
> On 3/14/22 13:32, Elvin Aslanov wrote:
> >         I think was_ is better
> > 
> >     Can you elaborate?
> > 
> > I mean naming the functions |was_string| and |was_number| instead of 
> > |creates_as| to match the ones with |is_|
> > 
> 
> "was_" implies past tense, something that no longer is true.  And these 
> well could still be true.  created_as or born_as are more accurate
> 

I don't agree that `was` means something is no longer true.

Cf. `He was, he is, and he will be.`

Anyway `created_as_` is long, and not Perly to me, unlike `was_` which feels highly Perl as an idiom.

---

`born_string` is still better than `created_as_string`

leonerd:
> Anyway `created_as_` is long, and not Perly to me, unlike `was_` which feels highly Perl as an idiom.

That is rathermuch by design though. We're trying to discourage most people from using these functions most of the time. Unless you're writing a JSON/YAML/MsgPack/Tangence/etc.. serialiser or other such similar code, you really don't want to be going near them. The point is to make them technically available to those who **need** them, but be sufficiently unpalatable to those who merely **want** them.

lib/builtin.t
leonerd: Turns out that was indeed a bug in core perl; #19540 fixed it. This PR has been rebased on that and I've added some more tests for those cases.

With that in place, your test case now prints:

```
created as number
created as number
```

lib/builtin.pm
```diff
@@ -154,6 +155,54 @@ Returns the basic container type of the referent of a reference, or C<undef>
...
+
+    my $val = Math::BigInt->new( 123 );
+
+    if( created_as_number $val ) { ... }    # this will not execute
+
+While most Perl code should operate on scalar values without needing to know
+their creation history, these two functions are intended to be used by data
+serialisation modules such as JSON encoders or similar situations, where
+language interoperability concerns require making a distinction between values
+that are fundamentally stringlike verses numberlike in nature.
```
gregoa: s/verses/versus/ # I suppose

leonerd: Thanks. Fixed and pushed; should be squashed before merging.

leonerd:
As it didn't get in in time for 5.35.10, this will now have to wait for 5.37.x.

leonerd:
PSC discussed this and it seems we'd still like to merge it before 5.35.11. It simply adds new experimentally-marked functions to the entirely-experimental `builtin` namespace, so the risk of breakage to existing programs is low to non-existent.

https://www.nntp.perl.org/group/perl.perl5.porters/2022/03/msg263425.html

ilmari (approved):

lib/builtin.pm
```diff
@@ -154,6 +155,32 @@ Returns the basic container type of the referent of a reference, or C<undef>
...
+
+    $bool = created_as_number($val);
+
+Returns a boolean representing if the argument value was originally created as
+a number. It will return true for any scalar expression whose most recent
+assignment or modification was of a numerical nature - such as assignment from
+a number literal, or the result of a numerical operation such as addition. It
+will return false for references, strings, booleans and undef.
+
+While most Perl code should operate on scalar values without needing to know
```
khwilliamson, line 201: This paragraph applies to both =head2 sections, but is included only in the second.

khwilliamson: I think the two doc commits should be s uashed.  (Something is wrong and I can't type the letter that comes between p and r in this text box; I've seen this before, it happens on both Linux and Windows boxes.)

The title of the commit message would be better to follow the paradigm of the 2nd one, so that someone perusing the history of commits will know what this commit refers to

khwilliamson (approved):

lib/builtin.pm
```diff
@@ -154,6 +155,54 @@ Returns the basic container type of the referent of a reference, or C<undef>
 for a non-reference. This is returned as a string in all-capitals, such as
 C<ARRAY> for array references, or C<HASH> for hash references.
 
+=head2 created_as_string
+
+    $bool = created_as_string($val);
+
+Returns a boolean representing if the argument value was originally created as
+a string. It will return true for any scalar expression whose most recent
+assignment or modification was of a stringy nature - such as assignment from a
```
jkeenan: Since this documentation will presumably be around for a long time to come, I'd suggest `s/stringy/string-like/g` to give a less casual tone.

lib/builtin.pm
leonerd: Yeah I tend to squash little docs-fix commits down before actually merging. I just usually leave them visible during review so reviewers can see it.

lib/builtin.pm
leonerd: Done in 8c82e520c2

PR #19525, opened 2022-03-11
Fix order of builtin:: warnings tests

leonerd:
The variables were assigned in a different order to the funcs themselves, meaning they didn't line up and were confusing to read.


haarg (approved):

ilmari (approved):
Oops, I messed that up when creating this test. Thanks for fixing!

Issue #19519, opened 2022-03-11
BBC: Blead Breaks Set::CrossProduct


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.42 running under perl 5.35.10.

-----------------------------------------------------------------
[Please describe your issue here]

BBC: Blead Breaks Set::CrossProduct

Please see http://matrix.cpantesters.org/?dist=Set::CrossProduct

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.35.10:

Configured by cpan at Tue Mar  8 09:26:05 EST 2022.

Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: f70b3f77e0c82e8e1b599de1542ef342631cb4f5
  Platform:
    osname=openbsd
    osvers=7.0
    archname=OpenBSD.amd64-openbsd
    uname='openbsd cjg-openbsd7 7.0 generic#5 amd64 '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    optimize='-O2'
    cppflags='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='OpenBSD Clang 11.1.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags ='-Wl,-E  -fstack-protector-strong -L/usr/local/lib'
    libpth=/usr/lib /usr/local/lib
[...]

Issue #19518, opened 2022-03-11
BBC: Blead Breaks Regexp::Grammars


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.42 running under perl 5.35.10.

-----------------------------------------------------------------
[Please describe your issue here]

BBC: Blead Breaks Regexp::Grammars

Please see http://matrix.cpantesters.org/?dist=Regexp::Grammars

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.35.10:

Configured by cpan at Tue Mar  8 09:26:05 EST 2022.

Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: f70b3f77e0c82e8e1b599de1542ef342631cb4f5
  Platform:
    osname=openbsd
    osvers=7.0
    archname=OpenBSD.amd64-openbsd
    uname='openbsd cjg-openbsd7 7.0 generic#5 amd64 '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    optimize='-O2'
    cppflags='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='OpenBSD Clang 11.1.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags ='-Wl,-E  -fstack-protector-strong -L/usr/local/lib'
    libpth=/usr/lib /usr/local/lib
[...]

Issue #19517, opened 2022-03-11
BBC: Blead Breaks Excel::Writer::XLSX


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.42 running under perl 5.35.10.

-----------------------------------------------------------------
[Please describe your issue here]

BBC: Blead Breaks Excel::Writer::XLSX

Please see http://matrix.cpantesters.org/?dist=Excel::Writer::XLSX

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.35.10:

Configured by cpan at Tue Mar  8 09:26:05 EST 2022.

Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: f70b3f77e0c82e8e1b599de1542ef342631cb4f5
  Platform:
    osname=openbsd
    osvers=7.0
    archname=OpenBSD.amd64-openbsd
    uname='openbsd cjg-openbsd7 7.0 generic#5 amd64 '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    optimize='-O2'
    cppflags='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='OpenBSD Clang 11.1.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags ='-Wl,-E  -fstack-protector-strong -L/usr/local/lib'
[...]

PR #19527, opened 2022-03-12
Revert "Fix GH Issue #19472: read warnings from open($fh,">",\(my $x))"

demerphq:
This reverts commit 8b03aeb95ab72abdb2fa40f2d1196ce42f34708d.

This is causing BBC breakage, and its unimport and grey zone enough that
we can pick it up in 5.37 when we have more time to deal with it.

### DIFF
--- a/doio.c
+++ b/doio.c
@@ -639,16 +639,6 @@ Perl_do_open6(pTHX_ GV *gv, const char *oname, STRLEN len,
                 goto say_false;
             }
 #endif /* USE_STDIO */
-            if (SvROK(*svp) && !sv_isobject(*svp)) {
-                /* if they pass in a reference and its not an object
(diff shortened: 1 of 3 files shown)


jkeenan:
I built a perl from this pull request and attempted to install each of the following 4 CPAN modules which are failing against blead:

[Excel::Writer::XLSX](https://github.com/Perl/perl5/issues/19517)

[Regexp::Grammars](https://github.com/Perl/perl5/issues/19518)

[Set::CrossProduct](https://github.com/Perl/perl5/issues/19519)

[Test::Smoke](https://github.com/Perl/perl5/issues/19520)

Each PASSed, so +1 to this reversion pull request.  (I'll apply it to blead on March 14 unless someone beats me to it.)

PR #19528, opened 2022-03-12
Add documentation on slices on an array or hash element

demerphq:
In PR #18827 Dan Jacobson requested we provide examples for slices on an
array element or hash element, he struggled to get the patch right so I
am pushing this as a replacement.

The feedback from the PR #18827 discussion suggested including a brief
introduction to postfix notation so I added that as well as a reference
to the perlref manpage.


Grinnz:
These additions seem out of place to me and add confusion, since this documentation is not about references.

PR #19530, opened 2022-03-13
hv_macro.h: only use direct memory loads on little endian machines

demerphq:
Using the shift logic should work correctly anywhere, using direct
memory loads will only work correctly on little endian machines.


demerphq:
Unless someone objects I will merge this on the weekend.

Issue #19484, opened 2022-03-03
`"AB" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x` doesn't match


ikegami:
`"AB" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x` doesn't match.

---

The following is the example from the documentation for `(*ACCEPT)`:

```
$ perl -e'CORE::say "AB" =~ /(A (A|B(*ACCEPT)|C) D)(E)/x || 0;'
1
```

No surprises there. But if we add a modifier to the alternation...

```
$ perl -e'CORE::say "AB" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x || 0;'
0
```

It doesn't even match `ABDE`!

```
$ perl -e'CORE::say "ABDE" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x || 0;'
0
```

In both cases, a match isn't even attempted.

```
$ perl -Mre=debug -e'CORE::say "AB" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x || 0;'
...
anchored "A" at 0..0 floating "DE" at 1..9223372036854775807 (checking floating) minlen 4
String shorter than min possible regex match (2 < 4)
0
Freeing REx: "(A (A|B(*ACCEPT)|C)+ D)(E)"
```

```
$ perl -Mre=debug -e'CORE::say "ABDE" =~ /(A (A|B(*ACCEPT)|C)+ D)(E)/x || 0;'
...
anchored "A" at 0..0 floating "DE" at 1..9223372036854775807 (checking floating) minlen 4
Matching REx "(A (A|B(*ACCEPT)|C)+ D)(E)" against "ABDE"
Intuit: trying to determine minimum start position...
  doing 'check' fbm scan, [1..3] gave -1
  Did not find floating substr "DE"...
Match rejected by optimizer
0
Freeing REx: "(A (A|B(*ACCEPT)|C)+ D)(E)"
```

---

Perl 5.34.0 built for Ubuntu on WSL using perlbrew.

ikegami:
Source: https://stackoverflow.com/q/71333798/589924

demerphq:
On Thu, 3 Mar 2022 at 09:35, Eric Brine ***@***.***> wrote:

> Source: https://stackoverflow.com/q/71333798/589924
>
>
> Interesting, off the top of my head I would say this is a regression from
when Dave reworked the fixed string logic.

Ill take a look.

[...]

PR #19531, opened 2022-03-13
Fix issue #19484 - ACCEPT in (...)+ doesn't work as expected, plus other ACCEPT related bugs.

demerphq:
This PR fixes Issue #19484, along with some ancillary issues related to ACCEPT and look-behind. 

It also adds a bunch of tests.

It also add some additional debugging output for use re Debug => ALL to debug minlen/stopmin/delta issues.

It also does some code cleanup and commenting to make things easier to understand.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -5681,6 +5681,11 @@ S_study_chunk(pTHX_
                                    : f)
                                   , depth+1, mutate_ok);
 
+                if (data && data->flags & SCF_SEEN_ACCEPT) {
+                    if (mincount > 1)
+                        mincount = 1;
+                }
+
                 if (flags & SCF_DO_STCLASS)
                     data->start_class = oclass;
                 if (mincount == 0 || minnext == 0) {
--- a/regcomp.c
+++ b/regcomp.c
@@ -4631,7 +4631,7 @@ STATIC SSize_t
 S_study_chunk(pTHX_
     RExC_state_t *pRExC_state,
     regnode **scanp,        /* Start here (read-write). */
-    SSize_t *minlenp,
+    SSize_t *minlenp,       /* used for the minlen of substrings? */
     SSize_t *deltap,        /* Write maxlen-minlen here. */
     regnode *last,          /* Stop before this one. */
     scan_data_t *data,      /* string data about the pattern */
@@ -4646,20 +4646,42 @@ S_study_chunk(pTHX_
                                a higher caller is holding a ptr to them. */
 )
 {
-    SSize_t final_minlen;
-    /* There must be at least this number of characters to match */
-    SSize_t min = 0;
-    I32 pars = 0, code;
-    regnode *scan = *scanp, *next;
-    SSize_t delta = 0;
+    /* vars about the regnodes we are working with */
+    regnode *scan = *scanp; /* the current opcode we are inspecting */
+    regnode *next = NULL;   /* the next opcode beyond scan, tmp var */
+    regnode *first_non_open = scan; /* FIXME: should this init to NULL?
+                                       the first non open regop, if the init
+                                       val IS an OPEN then we will skip past
+                                       it just after the var decls section */
+    I32 code = 0;           /* temp var used to hold the optype of a regop */
+
+    /* vars about the min and max length of the pattern */
+    SSize_t min = 0;    /* min length of this part of the pattern */
+    SSize_t stopmin = OPTIMIZE_INFTY; /* min length accounting for ACCEPT
+                                         this is adjusted down if we find
+                                         an ACCEPT */
+    SSize_t delta = 0;  /* difference between min and max length
+                           (not accounting for stopmin) */
+    SSize_t final_minlen = 0; /* final_minlen that we return  */
+
+    /* vars about capture buffers in the pattern */
+    I32 pars = 0;       /* count of OPEN opcodes */
+    I32 is_par = OP(scan) == OPEN ? ARG(scan) : 0; /* is this op an OPEN? */
+
+    /* vars about whether this pattern contains something that can match
+     * infinitely long strings, eg, X* or X+ */
     int is_inf = (flags & SCF_DO_SUBSTR) && (data->flags & SF_IS_INF);
     int is_inf_internal = 0;		/* The studied chunk is infinite */
-    I32 is_par = OP(scan) == OPEN ? ARG(scan) : 0;
-    scan_data_t data_fake;
-    SV *re_trie_maxbuff = NULL;
-    regnode *first_non_open = scan;
-    SSize_t stopmin = OPTIMIZE_INFTY;
-    scan_frame *frame = NULL;
+
+    /* scan_data_t (struct) is used to hold information about the substrings
+     * and start class we have extracted from the string */
+    scan_data_t data_fake; /* temp var used for recursing in some cases */
+
+    SV *re_trie_maxbuff = NULL; /* temp var used to hold whether we can do
+                                   trie optimizations */
+
+    scan_frame *frame = NULL;  /* used as part of fake recursion */
+
     DECLARE_AND_GET_RE_DEBUG_FLAGS;
 
     PERL_ARGS_ASSERT_STUDY_CHUNK;
--- a/regcomp.c
+++ b/regcomp.c
@@ -5715,6 +5715,16 @@ S_study_chunk(pTHX_
                     delta += (minnext + deltanext) * maxcount
                              - minnext * mincount;
                 }
+
+                if (data && data->flags & SCF_SEEN_ACCEPT) {
+                    if (flags & SCF_DO_SUBSTR) {
+                        scan_commit(pRExC_state, data, minlenp, is_inf);
+                        flags &= ~SCF_DO_SUBSTR;
+                    }
+                    if (stopmin > min)
+                        stopmin = min;
+                    DEBUG_STUDYDATA("after-whilem", data, depth, is_inf);
+                }
                 /* Try powerful optimization CURLYX => CURLYN. */
                 if (  OP(oscan) == CURLYX && data
                       && data->flags & SF_IN_PAR
--- a/regcomp.c
+++ b/regcomp.c
@@ -6514,11 +6514,10 @@ S_study_chunk(pTHX_
             if (OP(scan)==ACCEPT) {
                 /* m{(*ACCEPT)x} does not have to start with 'x' */
                 flags &= ~SCF_DO_STCLASS;
-                if (data) {
+                if (data)
                     data->flags |= SCF_SEEN_ACCEPT;
-                    if (stopmin > min)
-                        stopmin = min;
-                }
+                if (stopmin > min)
+                    stopmin = min;
             }
         }
         else if (OP(scan) == COMMIT) {
--- a/regcomp.c
+++ b/regcomp.c
@@ -4694,7 +4694,6 @@ S_study_chunk(pTHX_
             first_non_open=regnext(first_non_open);
     }
 
-
   fake_study_recurse:
     DEBUG_r(
         RExC_study_chunk_recursed_count++;
@@ -6312,6 +6311,7 @@ S_study_chunk(pTHX_
                                       last, &data_fake, stopparen,
                                       recursed_depth, NULL, f, depth+1,
                                       mutate_ok);
+
                 if (scan->flags) {
                     if (   deltanext < 0
                         || deltanext > (I32) U8_MAX
--- a/regcomp.c
+++ b/regcomp.c
@@ -1343,16 +1343,18 @@ S_debug_show_study_flags(pTHX_ U32 flags, const char *open_str,
 
 static void
 S_debug_studydata(pTHX_ const char *where, scan_data_t *data,
-                    U32 depth, int is_inf)
+                    U32 depth, int is_inf,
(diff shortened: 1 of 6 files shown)


regcomp.c
```diff
@@ -5715,6 +5715,16 @@ S_study_chunk(pTHX_
                     delta += (minnext + deltanext) * maxcount
                              - minnext * mincount;
                 }
+
+                if (data && data->flags & SCF_SEEN_ACCEPT) {
+                    if (flags & SCF_DO_SUBSTR) {
```
hvds: Should we also check for `SCF_DO_STCLASS` here?

regexec.c
```diff
@@ -8975,7 +8978,7 @@ NULL
                 sayNO;
 
           curlym_do_B: /* execute the B in /A{m,n}B/  */
-            if (ST.Binfo.count < 0) {
+            if (!is_accepted && ST.Binfo.count < 0) {
```
hvds: Seems like after this block we can still execute the following diagnostic ("CURLYM trying tail") and the subsequent block (testing for "CURLYM Fast bail"), which I'm guessing we shouldn't if `is_accepted`.

regexec.c
```diff
@@ -9019,21 +9022,22 @@ NULL
             if (ST.me->flags) {
```
hvds, line 9036: Similarly, if `ST.me->flags` is false, we'll fall through to the `PUSH_STATE_GOTO`, which I'm guessing we shouldn't. (Guessing at all this stuff, took me quite a few minutes to dig out what `ST.me` was.)

regexec.c
```diff
@@ -8397,21 +8398,29 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
...
+                             ? NEXTOPER(cursor)
+                             : regnext(cursor)
+                ){
+                    if ( OP(cursor) != CLOSE )
+                        continue;
+
+                    n = ARG(cursor);
+
+                    if ( n > lastopen )
+                        break;
```
hvds: If I read it correctly, this is an additional change to that described in the commit: previously it would have continued in this case. Can you clarify what it implies here if we see a `CLOSE(n)` with `n > lastopen`? Is making it break simply an additional optimization? At first glance it seems like this loop could correctly walk past `OPEN(lastopen+1) ... CLOSE(lastopen+1)` with a `CLOSE(lastopen)` still to come.

regcomp.c
```diff
@@ -6745,6 +6744,17 @@ S_study_chunk(pTHX_
...
     DEBUG_STUDYDATA("pre-fin", data, depth, is_inf, min, stopmin, delta);
 
+    if (stopmin < min) {
+        /* stopmin might be shorter than min if we saw an (*ACCEPT). If
+        this is the case then it means this pattern is variable length
+        and we need to ensure that the delta accounts for it. delta
+        represents the difference between min length and max length for
+        this part of the pattern. */
+        if (delta < min - stopmin)
+            delta = min - stopmin;
```
hvds: This is unexpected, I was expecting it to be an unconditional `delta += min - stopmin;`.

For `/ (A (*ACCEPT) | BC){2} D{0,4} /x` I think I'd expect before this block min=2 (for assumed match on AA), delta=6 (for match on BCBCDDDD), stopmin=1 (for match on A). After this block I'd expect min=1, delta=7, since the possible longest match has not changed.

t/re/re_tests
```diff
@@ -2086,6 +2086,8 @@ AB\s+\x{100}	AB \x{100}X	y	-	-
 /(A(A|B(*ACCEPT)|(?>C))D)(E)/	ABCDE	y	$&	AB	-	# ACCEPT with SUSPEND in pattern
 /(A(A|B(*ACCEPT)|C(?<=C))D)(E)/	ABCDE	y	$1-$2	AB-B	-	# ACCEPT with IFMATCH in pattern
 /(A(A|B(*ACCEPT)|C(?<=C))D)(E)/	ABCDE	y	$&	AB	-	# ACCEPT with IFMATCH in pattern
+/(A(A|B(*ACCEPT)|C(?<!Q))D)(E)/	ABCDE	y	$1-$2	AB-B	-	# ACCEPT with UNLESSM in pattern
+/(A(A|B(*ACCEPT)|C(?<!Q))D)(E)/	ABCDE	y	$&	AB	-	# ACCEPT with UNLESSM in pattern
```
hvds, line 2093: These don't match the commit message: the ACCEPT is not _inside_ UNLESSM.

hvds (commented):
It may also be worth adding tests to `t/re/opt.t` to more specifically declare the expected minlen, maxlen etc that we expect to be discovered for these cases.

regcomp.c
```diff
@@ -4644,20 +4646,41 @@ S_study_chunk(pTHX_
...
 {
-    SSize_t final_minlen;
-    /* There must be at least this number of characters to match */
-    SSize_t min = 0;
-    I32 pars = 0, code;
-    regnode *scan = *scanp, *next;
-    SSize_t delta = 0;
+    /* vars about the regnodes we are working with */
+    regnode *scan = *scanp; /* the current opcode we are inspecting */
+    regnode *next = NULL;   /* the next opcode from scan, tmp var */
```
khwilliamson, line 4651: I think s/from/beyond/ is clearer

khwilliamson (commented):
It would take me a lot of time to familiarize myself with the optimizer in order to adequately review these changes.  Except for a trivial note about a comment, it all looks reasonable to me.  But I'm close to having zero knowledge about this area.

regcomp.c
demerphq: No, not as far as I can tell. Although while verifying this I found a problem caused by different code. The start class stuff should be fine as long as the min-length logic is correct. I will push an additional commit to deal with that.

regexec.c
demerphq: Yes, right. Good catch.

regexec.c
demerphq: Yes, also true, I will reorganize that logic a bit when i push.

regexec.c
demerphq: That should read "continue" not "break". Will be fixed in the repush.

regcomp.c
demerphq: Yes, concur. Will fix. It turns out it doesn't matter much, as the delta is only used for fixed string, and fixed string is disabled by accept. But you are correct in your analysis. While debugging the case you showed however I found a different issue with the {2}, but that is a different bug and will get its own commit.

t/re/re_tests
demerphq: Test renamed, we are checking that the LOOKBEHIND_END does not break accept capture close logic.

regcomp.c
demerphq: changed.

demerphq:
I did a force push with the changes rebased. I had to do a rebase anyway as there were conflicts, sorry about that Hugo. I have changed pretty much everything mentioned in feedback, and added a couple of extra commits that came up while digging into Hugo's questions.

demerphq:
@hvds waiting on you on this one mate.

demerphq:
I have added the release blocker tag to this PR because it fixes some bugs related to not throwing experimental warnings about variable look-behind with capturing. Something like this:
```
/(?<=(foo|bar(*ACCEPT)|baz)bop)/
```
is actually variable length, and can match a string that is either 3 or 6 characters long. Without this patch the optimizer thinks all of the options are 6 characters long and therefore does not produce experimental warnings like it should.

regcomp.c
demerphq: So the reason (*ACCEPT) doesn't affect stclass (eg ANYOF style) in the optimizer in the same way it does substrings is because *ACCEPT affects the minimum length string that the pattern might match, but it doesn't affect the things it might start with. The only real effect it might have on startclass would be to allow it to accept the empty string. Put another way, (*ACCEPT) means that things to its right cant be mandatory, but they can still be something that starts a match. The only case where this is not the case is when ACCEPT would cause the pattern to match the empty and in that case there shouldnt be a startclass at all, and it should be handled via the minlen logic. This is why a couple of the later patches in this PR are there, to ensure that minlen is set properly.  BTW, those patches wouldn't exist if you hadn't asked this question so thanks!

Consider a case like this (using charclasses here to prevent this being turned into a mandatory substring question).

`/[aA]*[bB]/`

it should have a stclass of `ANYOF[aAbB]` and have a minlen of 1.

But now if we insert an `(*ACCEPT)` in the middle:

`/[aA]*(*ACCEPT)[bB]/`

it has a minlen of 0, and will match any string and the optimizer already understands that if the pattern can match the empty string then there is no startclass at all.

Change the * to a +, and the ACCEPT doesn't affect the stclass at all, it will be the same with or without the ACCEPT, any `ANYOF[aA]`.

Consider a case like `/ [aA]* ( [bB] | [cC](*ACCEPT) | [dD]*) [eE] /x`, this should have a stclass of `ANYOF[aAbBcCdDeE]`. The ACCEPT doesn't affect the fact that 'e' might be a legit start character for a matching string.

With mandatory substrings however the case is different. If I have `/a(?:b+(*ACCEPT))?c/` then the fixed string should be 'ab', c will not be part of a successful match. Compare with `/ab+c/`, where we have two mandatory strings to choose from 'ab' and 'bc'. 

With `/a* ( b | c | d ) e/` the mandatory substring would be 'e', but if i inject an accept into the alternation like this: `/a* (b | c (*ACCEPT) | d ) e/` then 'e' is now not mandatory, if it goes through the 'c' branch then 'e' will not be part of a match.

hvds: Thanks, I'd overlooked the fact that minlen=0 makes stclass irrelevant: I guess that was implied by your previous comment, but I didn't pick up on it.

(I'd also recommend putting backticks round code fragments: one of your examples above is being parsed as URL markup in github's web interface, because it matches `[...](...)`.)

t/re/reg_mesg.t
```diff
@@ -747,6 +747,12 @@ my @experimental_vlb = (
     '/(?| (?=(foo)) | (?<!(foo)|p) )/x'
             => 'Variable length negative lookbehind with capturing' .
                ' is experimental {#} m/(?| (?=(foo)) | (?<!(foo)|p) ){#}/',
+    '/(?<!(foo|bloop(*ACCEPT)|bar)baz)/'
+            => 'Variable length negative lookbehind with capturing' .
+               ' is experimental {#} m/(?<!(foo|bloop(*ACCEPT)|bar)baz){#}/',
+    '/(?<=(foo|bloop(*ACCEPT)|bar)baz)/'
+            => 'Variable length positive lookbehind with capturing' .
+               ' is experimental {#} m/(?<=(foo|bloop(*ACCEPT)|bar)baz){#}/',
```
hvds: I would have expected this to test something where the ACCEPT was what changed it from fixed to variable length: `/(?<=(ab|c(*ACCEPT)d))/`, and perhaps also the reverse: `/(?<=(ab|cd(*ACCEPT)efg))/` (though if the latter were wrongly to warn I'd have no problem living with it).

hvds (approved):
LGTM, the one new comment is optional.

regcomp.c
demerphq: Good tip. I updated the comment, but ill keep it in mind for the future. :-)

t/re/reg_mesg.t
demerphq: Ah good suggestion. Ill tweak that and push a fix now for rebase afterwards. (I'm learning. ;-)

demerphq: Pushed.

demerphq: BTW, I didn't add `/(?<=(ab|cd(*ACCEPT)efg))/` as it does warn as you suspected it might, and I want to add a new issue to make it not warn. But I don't want to burden this PR with that fix as it could be serious work and I'd really like to see this released in 5.36. I just had some ideas how to make this better, but perfect is the enemy of good enough and all that.

hvds (approved):

t/re/reg_mesg.t
hvds: > BTW, I didn't add `/(?<=(ab|cd(*ACCEPT)efg))/` as it does warn as you suspected it might, and I want to add a new issue to make it not warn. But I don't want to burden this PR with that fix as it could be serious work and I'd really like to see this released in 5.36. I just had some ideas how to make this better, but perfect is the enemy of good enough and all that.

Yep, that's fine by me.

hvds:
> I have added the release blocker tag to this PR because it fixes some bugs related to not throwing experimental warnings about variable look-behind with capturing. Something like this:
> 
> ```
> /(?<=(foo|bar(*ACCEPT)|baz)bop)/
> ```
> 
> is actually variable length, and can match a string that is either 3 or 6 characters long. Without this patch the optimizer thinks all of the options are 6 characters long and therefore does not produce experimental warnings like it should.

Given how broken ACCEPT has been up until now, it seems unlikely anyone has actually tried to use constructs like that in the years since it was added, so while I can see the argument I'm dubious about calling it a release blocker.

While the individual commits are beautifully small, in aggregate this is still quite a chunky change, so if we do decide to merge for 5.36 we should probably do so as soon as possible and keep a close eye out for any BBC impact.

demerphq:
Based on @rjbs  mail that we should be able to merge fixes that are unlikely to cause problems to anything other than the feature they pertain to, which this patch is I am merging now. Thanks for the review @hvds and @khwilliamson.

PR #19534, opened 2022-03-13
runenv_hashseed.t - rework to deal with probabilistic failures

demerphq:
The old version of this test had a 1/128 chance of failing the
RANDOM key change test each time we try it. We run it 50 times, so
there is a %39 chance of failing it at least once. It is actually
surprising we haven't seen test fails before today. The key change
test also has a similar problem, albeit with a much lower
probability of failing the test.

Today I got the first report of a failed test from Tux.

This patch reworks the test to reduce the odds of both tests to
50/131072 (across all runs). Which is about 0.03%, while low this
is still non-zero and since we cant have a fractional number of
test fails we allow for each type of test to fail once before we
consider it a true failure.

### DIFF
--- a/t/run/runenv_hashseed.t
+++ b/t/run/runenv_hashseed.t
@@ -1,8 +1,45 @@
 #!./perl
+
+# Test that setting PERL_HASH_SEED and PERL_PERTURB_KEYS in different
+# combinations works as expected, and that changing the values provided
+# produces the expected results
 #
-# Tests for Perl run-time environment variable settings
+# We do this by first executing Perl with a given PERL_PERTURB_KEYS
+# mode, and then extract the randomly chosen PERL_HASH_SEED it ran under
+# from its debug output which was printed to STDERR, and then use it for
+# further tests. This allows the tests to be robust to the choice of hash
+# function and seed sizes that might be in use in the perl being tested.
+# We do not ask perl to output any keys on this run, as our subsequent
+# runs will use different environment variables (specifically
+# PERL_HASH_SEED) which will change any key order results we see.
 #
-# $PERL5OPT, $PERL5LIB, etc.
+# We then execute perl a further three times and ask perl to build a
+# hash with a specific number of buckets and a specific set of keys. We
+# then have perl print the raw keys to STDOUT.
+#
+# For two of these three runs we supply the same seed, and both of those
+# times we supply the same perturb mode, but in different ways, once as
+# a name and once as a digit. The debug output should be identical in
+# both cases regardless of mode. For PERL_PERTURB_KEYS mode 0=NO, and
+# 2=DETERMINISTIC the key order should match. For mode 1=RANDOM the key
+# order should differ the vast majority of the time, however the test is
+# probabilistic and occasionally may result in the same key order.
+#
+# The third run we supply a different seed, with a 1 bit difference, but
+# with the same PERL_PERTURB_KEYS mode. In this case we expect the key
+# order to differ for all three modes, but again the test is
+# probabilistic and we may get the same key order in a small percentage
+# of the times we try this.
+#
+# To address the probabilistic nature of these tests we run them
+# multiple times and count how many times we get the same key order.
+# Most times this should be zero, but occasionally it might be higher.
+# Therefore we use a threshold $allowed_fails to determine how many
+# times the key order may be unchanged before we consider the tests
+# actually failed. We also use a largish number of keys in a hash with
+# a large number of buckets, which means we produce a lot a large temp
+# files as we test, so we aggressively clean them up as we go.
+
 
 BEGIN {
     chdir 't' if -d 't';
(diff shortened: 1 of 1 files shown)


t/run/runenv_hashseed.t
```diff
@@ -42,24 +46,45 @@ sub my_runperl_and_capture {
...
+        'NO',
+        'RANDOM',
+        'DETERMINISTIC'
+    ); # 0, 1 and 2 respectively
+
+my $repeat = 50; # if this changes adjust the comments below.
+
+plan tests => ((@modes-1) * 6 * $repeat) + #non RANDOM modes
+              (5 * $repeat) + 1 + # RANDOM mode
+              @modes; # all modes
```
hvds: I'd recommend always breaking long expressions _before_ an operator - the trailing '+' here is difficult to pick out.
```
plan tests => ((@modes-1) * 6 * $repeat) #non RANDOM modes
              + (5 * $repeat) + 1 # RANDOM mode
              + @modes; # all modes
```

t/run/runenv_hashseed.t
```diff
@@ -42,24 +46,45 @@ sub my_runperl_and_capture {
...
+# Note the keys(%h)=100_000 means perl will allocate 1<<17 buckets for
+# the hash, IOW, 131072 buckets.
 my $print_keys = [
     '-Dh', '-I../lib',
     (is_miniperl() ? () : '-MHash::Util=hash_traversal_mask,num_buckets'),
     '-e',
-    'my %h; @h{"A".."Z", "a".."z"}=(); @k=keys %h;'.
+    'my %h; keys(%h)=100_000; @h{"A".."Z", "a".."z"}=(); @k=keys %h;'.
       ' print join ":", 0+@k, ' .
       (is_miniperl() ? '' : 'num_buckets(%h),hash_traversal_mask(%h),') .
```
hvds: `hash_traversal_mask` has no prototype, so this call fails to see a ref as first argument and immediately returns an empty list. Test still passes for me (and probably runs a bit faster) if `\%h` is passed.

t/run/runenv_hashseed.t
```diff
@@ -76,9 +101,8 @@ for my $mode (@modes) {
...
 
         if ( $mode eq 'RANDOM' ) {
-            isnt( $out1, $out2,
-                "$descr results in different key order with the same keys"
-            );
+            # there is a 1/131072 chance that we got the same result.
+            $random_same++ if $out1 eq $out2;
         }
         else {
             is( $out1, $out2,
```
hvds: For the SKIP block just below: doesn't seem like the mode RANDOM are really "skipped" tests. I'd be inclined to change it to:
```
  for (1) {
    next if $mode eq 'RANDOM';  # not testing rand bits
```
... with a corresponding adjustment to the plan.

Also, `@$rand_bits_data`  appear to be empty for mode NO - is that expected? The relevant tests should probably also be either skipped or (as with RANDOM) not tested at all in that case.

For mode DETERMINISTIC it seems they are empty if perl is compiled without debugging. I think those should be true skips, and it should probably be a fail if they are empty when we expect them not to be.

demerphq:
FWIW, this or something like it needs to be applied before we release 5.36. This old version of the test has a non-zero chance of failing on a perfectly functional perl due to the probabilistic nature of what it is testing.

t/run/runenv_hashseed.t
```diff
@@ -126,11 +150,28 @@ for my $mode (@modes) {
         my $new_seed = "0x" . join '', @chars;
```
hvds: At line ---118/+++142 above, it would be super to reduce this to under 80 chars.

The construction of `$new_seed` seems a bit tortuous, I think it could be shortened to:
```
  my $new_seed = $seed;
  substr($new_seed, -1) =~ tr{0-9a-f}{1-9a-f0};
```

Is there a desire to avoid some of those constructs in this testfile?

t/run/runenv_hashseed.t
```diff
@@ -126,11 +150,28 @@ for my $mode (@modes) {
...
         isnt $new_seed, $seed, "$descr got a different seed";
```
hvds: This isn't a test of hash functionality, you could simply die if they're not different.

t/run/runenv_hashseed.t
```diff
@@ -126,11 +150,28 @@ for my $mode (@modes) {
...
+        }
     }
+
+    if ($mode eq "RANDOM") {
+        # There is a 50/131072 chance that we got the same result.
+        # Rounded up that means we should allow at most 1 case where
+        # the keys were the same.
+        ok( $random_same <= 1,
+            "$descr_mode results in same key order at most once")
+            or diag("Key order was the same $random_same/$repeat times");
```
hvds: Here and below, it might be worth having a longer diag() to clarify that a failure here is expected to occur by chance about 1 in XXX runs, so is not a problem unless it happens consistently. Or maybe set the limit to 2 instead of 1.

(OTTOMH I think XXX is probably 131072^2/comb(50,2) ~= 1.4e7, which would mean it's quite likely someone, somewhere will see this eventually.)

t/run/runenv_hashseed.t
```diff
@@ -1,8 +1,20 @@
...
 #
-# Tests for Perl run-time environment variable settings
+# Tests for PERL_PERMUTE_KEYS and PERL_HASH_SEED
 #
-# $PERL5OPT, $PERL5LIB, etc.
+# For PERL_PERMUTE_KEYS=NO or PERL_PERMUTE_KEYS=DETERMINISTIC we should
+# get the same key order each time.
+#
+# For PERL_PERMUTE_KEYS=RANDOM the test is probabilistic and there is a
+# 1/131072 chance per repeat that we might get the same key order.
```
hvds: I'd like to add an extra clarification here:
```
# For PERL_PERMUTE_KEYS=RANDOM we should get a _different_ key order each time,
# but the test is probabilistic and there is a 1/131072 chance per repeat that
# we might get the same key order.
```

t/run/runenv_hashseed.t
demerphq: Sure if you prefer that I'll fix it up.

t/run/runenv_hashseed.t
demerphq: Doh. I can't believe I didn't notice that. Grrrr. Thanks!

t/run/runenv_hashseed.t
demerphq: Regarding the skip. I had the same thought but wanted to keep the patch minimally changed from @atoomic's original. But now that you suggest ill make the change. 

Regarding the running under DEBUGGING part, do you know a reliable way to tell if we are running under DEBUGGING in a test? If you do I am all for it, otherwise I'd rather leave it. I am not aware of a proper %Config key for that. (Its on my todo list to address actually, along with some other defines, but I am waiting on @neilb 's patch to add the taint stuff so I can crib from what he did.)

t/run/runenv_hashseed.t
demerphq: Nice trick! this is code from @atoomic's original patch that I didn't change at all. No reason to avoid tr here at all. Ill make the change.

t/run/runenv_hashseed.t
demerphq: Fair enough.

t/run/runenv_hashseed.t
demerphq: If you think 2 fails in 50 is likely then I'll set it to 2. :-) I'm going to reach out to you privately about this one though, i have some questions.

t/run/runenv_hashseed.t
demerphq: Sure. Will do.

t/run/runenv_hashseed.t
hvds: > Regarding the running under DEBUGGING part, do you know a reliable way to tell if we are running under DEBUGGING in a test? If you do I am all for it, otherwise I'd rather leave it. I am not aware of a proper %Config key for that. (Its on my todo list to address actually, along with some other defines, but I am waiting on @neilb 's patch to add the taint stuff so I can crib from what he did.)

Oh, I assumed there would be an obvious way to test for DEBUGGING (short of trying to run `perl -DU - e1` and checking stderr), but if there isn't one right now I'm fine with deferring that until there is.

But I think it would already make sense to have it skip rather than pass if there's no data.

demerphq:
Repushed with all your feedback incorperated @hvds, thanks!

t/run/runenv_hashseed.t
```diff
@@ -16,121 +31,193 @@ skip_all_without_config('d_fork');
...
         # and once more, these two should be the same
-        my ( $out2, $err2, $err_got_data2, $rand_bits_data2 )
+        my ( $out2, $err2, $seed2, $err_got_data2, $rand_bits2 )
             = my_runperl_and_capture( $run_opts, $print_keys );
 
-        if ( $mode eq 'RANDOM' ) {
-            isnt( $out1, $out2,
-                "$descr results in different key order with the same keys"
-            );
+        is( $err, $err2,
```
hvds: Was the intent here to compare `($err1, $err2)`?

t/run/runenv_hashseed.t
```diff
@@ -16,121 +31,193 @@ skip_all_without_config('d_fork');
...
+            "$descr debug output was consistent between runs"
+        );
+        if( $seed1 ne $seed) {
+            die "Seed wasn't consistent on retry 1";
         }
-        else {
-            is( $out1, $out2,
-                "$descr results in the same key order each time"
-            );
+        if( $seed2 ne $seed) {
```
hvds: Given the `$err2` output is the same (whether you end up comparing to `$err` or `$err1`), this test is redundant. Unless you also intended to remove the hash seed when parsing it out, in which case line 51 needs to change.

t/run/runenv_hashseed.t
```diff
@@ -16,121 +31,193 @@ skip_all_without_config('d_fork');
...
+    }
+
+    # There is a small chance we got the same result. So allow a certain
+    # number of fails.
+    ok( $seed_change_same <= $allowed_fails,
+        "$descr_mode same key order with munged seed no more " .
+        "than $allowed_fails times" )
+        or diag(
+            "Key order was the same $random_same/$repeat times with",
+            "a munged seed. This test is probabilistic so if the number",
```
hvds: Given the audience for this, I suggest `s/munged/different/`.

t/run/runenv_hashseed.t
```diff
@@ -1,8 +1,23 @@
...
+# For PERL_PERMUTE_KEYS=RANDOM the test is probabilistic and there is a
+# 1/131072 chance per repeat that we might get the same key order.
+#
+# For all the test modes there is also a chance that we might get the same
+# key order when we munge the hash seed, albeit a much lower chance.
+#
+# We do 50 repeats, so we have 50/131072 chance of getting the same
+# results for the probabilistic tests, which we round up to 1 since the
+# count of test fails can not be a fraction. So we allow 1 try per batch
+# of repeats to produce the same result before we fail the test.
```
hvds: This is out of date now, we use `$allowed_fails`.

t/run/runenv_hashseed.t
```diff
@@ -16,121 +31,193 @@ skip_all_without_config('d_fork');
...
+        'DETERMINISTIC'
+    ); # 0, 1 and 2 respectively
+
+my $repeat = 50; # if this changes adjust the comments below.
+my $min_buckets = 100_000;
+my $actual_buckets = 1;
+$actual_buckets *= 2 while $actual_buckets <= $min_buckets;
+my $key_expr = '"A" .. "Z", "a" .. "z", 0 .. 9';
+my @keys = eval $key_expr;
+my $allowed_fails = 2; # FIXME: compute from @keys and $actual_buckets
```
hvds: Until then, maybe add a comment that we expect this to give a false negative about 1 in 1.1e11 times. You can add "Hugo says" on there if you like. :)

hvds (commented):
LGTM, remaining items are minor nits.

t/run/runenv_hashseed.t
demerphq: No its to compare that tested debug output corresponds to the original output so that we are sure we are testing the same thing. I could change this to a die but I dont see a problem if it stays as an is(). It tests that we can actually set the hash seed.

t/run/runenv_hashseed.t
demerphq: Ack. Willdo.

t/run/runenv_hashseed.t
demerphq: Right, ill update the comment.

t/run/runenv_hashseed.t
demerphq: Sure, ill add something like that. I am not sure exactly what the odds are, given that some changes to the mask might not actually change the order. Eg, if nothing is in the first bucket then changes to only the low bit of the mask wont change the order of the keys, for instance if the mask was 0x0 and 0x1. That was part of the reason I added more keys. I think stating a specific probability is not helpful unless we know we are right.  That was part of the reason I changed some of the other wording.

hvds (commented):
Not sure why this would pass if %ENV is not the same. Other than that, all looks good.

t/run/runenv_hashseed.t
```diff
@@ -36,51 +58,75 @@ use warnings;
 
 # enable DEBUG_RUNENV if you want to see what is being returned
 # by the executed perl.
-sub my_runperl_and_capture {
-    my ($opts_hash, $cmd_array) = @_;
+sub my_runperl {
+    my ($cmd_array, $perturb, $set_seed) = @_;
+    my $opts_hash= {
+        PERL_HASH_SEED_DEBUG => 1,
+        PERL_PERTURB_KEYS => $perturb
```
hvds, line 65: (My comment appears to have gone awol from last review, trying again)

I thought the intent was for the two runs to have identical ENV, but you're setting this as a number on one run and as a string on the other. I don't understand why that isn't causing the test to fail.

rjbs:
At this point, what is the reason that this is not merged?

jkeenan:
> At this point, what is the reason that this is not merged?

This is the sort of patch that a very small number of people are competent to review.  One of the few, @hvds, raised a concern in a comment, but has not gotten a response.

t/run/runenv_hashseed.t
haarg: The numbers and strings are both allowed values of `PERL_PERTURB_KEYS`. The test is verifying that the name and its corresponding string behave the same.

hvds:
> At this point, what is the reason that this is not merged?

The only reason from my point of view is that due to family issues, @demerphq has apparently not had a chance to respond to my comment on his most recent rework.

t/run/runenv_hashseed.t
hvds: > The numbers and strings are both allowed values of `PERL_PERTURB_KEYS`. The test is verifying that the name and its corresponding string behave the same.

Yes, but the message from the rest of this test file is that they _should not_ behave the same, because any change in the incoming `%ENV` should cause the hash randomization to give different results. So providing an ENV key as a number one time and as a keyword the next should not give comparable results.

demerphq:
On Sun, 3 Apr 2022 at 03:03, Hugo van der Sanden ***@***.***>
wrote:

>
> In t/run/runenv_hashseed.t
> <https://github.com/Perl/perl5/pull/19534#discussion_r841136740>:
>
> > @@ -36,51 +58,75 @@ use warnings;
>
>  # enable DEBUG_RUNENV if you want to see what is being returned
>  # by the executed perl.
> -sub my_runperl_and_capture {
> -    my ($opts_hash, $cmd_array) = @_;
> +sub my_runperl {
> +    my ($cmd_array, $perturb, $set_seed) = @_;
> +    my $opts_hash= {
> +        PERL_HASH_SEED_DEBUG => 1,
> +        PERL_PERTURB_KEYS => $perturb
>
> The numbers and strings are both allowed values of PERL_PERTURB_KEYS. The
> test is verifying that the name and its corresponding string behave the
> same.
>
> Yes, but the message from the rest of this test file is that they *should
> not* behave the same, because any change in the incoming %ENV should
> cause the hash randomization to give different results. So providing an ENV
> key as a number one time and as a keyword the next should not give
> comparable results.
>

In older perls the sensitivity is to having different keys in %ENV, as the
hash value for each key was mixed into the random state, with the most
recent patches we no longer do this, but there is still a sensitivity to
the /number/ of keys in %ENV, each key causes one iteration of the XORSHIFT
RNG.  FWIW, the value of an environment variable has never been relevant.

I am back online, the funeral is over and I am working through my backlog.
Sorry for my prolonged absence, a bunch of stuff conspired together to make
it very difficult for me to do any p5p work at all for a few weeks there.

cheers,
Yves

hvds:
> In older perls the sensitivity is to having different keys in %ENV, as the hash value for each key was mixed into the random state, with the most recent patches we no longer do this, but there is still a sensitivity to the /number/ of keys in %ENV, each key causes one iteration of the XORSHIFT RNG. FWIW, the value of an environment variable has never been relevant.

Ah cool, that was never clear to me.

(Wrt older perls, I may also at some point have misparsed "the hash value for each key" as meaning "the value in the hash for each key". It might be useful to prefer "the hash of a key" or "the hashed value" to reduce the chance of such errors.)

hvds (approved):

demerphq:
On Mon, 4 Apr 2022 at 15:37, Hugo van der Sanden ***@***.***>
wrote:

> In older perls the sensitivity is to having different keys in %ENV, as the
> hash value for each key was mixed into the random state, with the most
> recent patches we no longer do this, but there is still a sensitivity to
> the /number/ of keys in %ENV, each key causes one iteration of the XORSHIFT
> RNG. FWIW, the value of an environment variable has never been relevant.
>
> Ah cool, that was never clear to me.
>
> (Wrt older perls, I may also at some point have misparsed "the hash value
> for each key" as meaning "the value in the hash for each key". It might be
> useful to prefer "the hash of a key" or "the hashed value" to reduce the
> chance of such errors.)
>
Ah good point. Too close to the trees to see the forest. :-)

Yves

-- 

demerphq:
Note that last force push was just to squash the versions together before I merge. No changes made.

hvds:
> Note that last force push was just to squash the versions together before I merge. No changes made.

Thanks, now merged as af3d9fd639.

demerphq:
Thanks @hvds you beat me to the punch. I was waiting for a last sanity check to complete before I merged. I removed the release blocker label since it is now merged.

hvds:
> Thanks @hvds you beat me to the punch. I was waiting for a last sanity check to complete before I merged.

Ah, I had already done a series of sanity checks before realising you'd want to squash the commits yourself. :)

> I removed the release blocker label since it is now merged.

Not a big issue, but I would assume that is not necessary: for making a release we care about _open_ issues that are release blockers (and if you click on the label, it defaults to listing only open issues). It also discards the (minor) historical information that it _was_ a release blocker - it doesn't work so well since the label is not version-specific, but it could be useful if for example someone wanted to double-check that all the closed release blockers had actually been fixed.

PR #19537, opened 2022-03-13
perldelta for removal of empty sort

hvds:
This reflects the change in 78cc98885f, "make a sort with zero args
a compile-time err".

(Should this commit also remove the preceding "XXX" boilerplate?)

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -59,6 +59,13 @@ XXX For a release on a stable branch, this section aspires to be:
 
 [ List each incompatible change as a =head2 entry ]
 
+=head2 sort with no arguments is now a compile-time error
+
+C<@a = sort;> used to be equivalent to C<@a = ();>, which wasn't very
+useful; it is now a compile-time error. The old behaviour has been
+removed to make room for future features. Sorting an empty array (ie
+C<@empty = (); @a = sort @empty;>) continues to be valid syntax.
+
 =head1 Deprecations
 
 XXX Any deprecated features, syntax, modules etc. should be listed here.


hvds:
Oh, I just realised that this was already in perl5358delta.pod, I'd forgotten that perldelta.pod is not cumulative up to a release. So ignore this.

pod/perldelta.pod
```diff
@@ -59,6 +59,13 @@ XXX For a release on a stable branch, this section aspires to be:
 
 [ List each incompatible change as a =head2 entry ]
 
+=head2 sort with no arguments is now a compile-time error
+
+C<@a = sort;> used to be equivalent to C<@a = ();>, which wasn't very
+useful; it is now a compile-time error. The old behaviour has been
+removed to make room for future features. Sorting an empty array (ie
```
khwilliamson, line 66: Perhaps things have changed, but the correct way to spell this is I<i.e.>,

pod/perldelta.pod
```diff
@@ -59,6 +59,13 @@ XXX For a release on a stable branch, this section aspires to be:
 
 [ List each incompatible change as a =head2 entry ]
 
+=head2 sort with no arguments is now a compile-time error
+
+C<@a = sort;> used to be equivalent to C<@a = ();>, which wasn't very
```
khwilliamson, line 64: Technically, the C<> should be surrounded by S<>, but it doesn't matter here since it is first in the paragraph, so won't wrap.

pod/perldelta.pod
```diff
@@ -59,6 +59,13 @@ XXX For a release on a stable branch, this section aspires to be:
...
+useful; it is now a compile-time error. The old behaviour has been
+removed to make room for future features. Sorting an empty array (ie
+C<@empty = (); @a = sort @empty;>) continues to be valid syntax.
+
```
khwilliamson, line 68: S<> should probably be added.

I have no substantive issues with this P.R.  LGTM

PR #19540, opened 2022-03-15
always prevent setting POK flag when NV values are used as strings

haarg:
Since PR #18958, values that start as IVs will not get their POK flags
set when they are used as strings. This is meant to aid in
serialization, allowing the "original" type of a value to be preserved.

For NV values, the POK flag was already usually not being set, because
the string form of a float could change based on the locale changing.
However, for Inf and NaN values, the POK flag would still be enabled.
Also, POK would be set for all floats if USE_LOCALE_NUMERIC was not
defined.

Update Perl_sv_2pv_flags to only enable the POKp flag when storing the
PV for Inf or NaN values, or all NVs when USE_LOCALE_NUMERIC is not
defined.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -3005,7 +3005,7 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
             len = S_infnan_2pv(SvNVX(sv), s, size, 0);
             if (len > 0) {
                 s += len;
-                SvPOK_on(sv);
+                SvPOKp_on(sv);
             }
             else {
                 /* some Xenix systems wipe out errno here */
@@ -3025,7 +3025,7 @@ Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
 #ifndef USE_LOCALE_NUMERIC
                 SNPRINTF_G(SvNVX(sv), s, SvLEN(sv), NV_DIG);
 
-                SvPOK_on(sv);
+                SvPOKp_on(sv);
 #else
                 {
                     bool local_radix;
--- a/t/op/svflags.t
+++ b/t/op/svflags.t
@@ -10,7 +10,7 @@ BEGIN {
 # Tests the new documented mechanism for determining the original type
 # of an SV.
 
-plan tests => 4;
+plan tests => 12;
 use strict;
 use B qw(svref_2object SVf_IOK SVf_NOK SVf_POK);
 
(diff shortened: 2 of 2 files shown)


haarg:
This is meant to fix the issue @wolfsage noted on #19524 that prevents a potential `created_as_string` from being reliable on Inf values.

cc: @nwc10

leonerd (approved):
LGTM

sisyphus:
Heh ... it's a pity this didn't come up a coupla weeks ago.
I knew what was failing in http://www.cpantesters.org/cpan/report/dc17e330-900b-11ec-bfc9-d1f1448276d4 but I didn't exactly know why.
I thought it must have been some in-house patch that OpenBSD 7.0 was using.
But *now* I can see that it arose because USE_LOCALE_NUMERIC was not a compile-time option.

One question (and sorry for the diversion), but can you tell me how the absence of this compile-time option is reflected in %Config ?
Obviously, I'm assuming that there *is* some %Config setting that reflects this case. However, I couldn't find it.

khwilliamson:
On 3/17/22 06:56, sisyphus wrote:
> Heh ... it's a pity this didn't come up a coupla weeks ago.
> I knew what was failing in 
> http://www.cpantesters.org/cpan/report/dc17e330-900b-11ec-bfc9-d1f1448276d4 
> <http://www.cpantesters.org/cpan/report/dc17e330-900b-11ec-bfc9-d1f1448276d4> 
> but I didn't exactly know why.
> I thought it must have been some in-house patch that OpenBSD 7.0 was using.
> But /now/ I can see that it arose because USE_LOCALE_NUMERIC was not a 
> compile-time option.
> 

I had not been following that closely.  Does this just mean you wasted 
some time, or is there something needed to be done?

> One question (and sorry for the diversion), but can you tell me how the 
> absence of this compile-time option is reflected in %Config ?
> Obviously, I'm assuming that there /is/ some %Config setting that 
> reflects this case. However, I couldn't find it.

Finding what locale things are enabled is complicated, so I wrote 
t/loc_tools.pl.  That ay have a sub you could use, or you can examine it 
to see what it does

sisyphus:
On Sat, Mar 19, 2022 at 4:14 AM Karl Williamson ***@***.***>
wrote:

> On 3/17/22 06:56, sisyphus wrote:
> > Heh ... it's a pity this didn't come up a coupla weeks ago.
> > I knew what was failing in
> >
> http://www.cpantesters.org/cpan/report/dc17e330-900b-11ec-bfc9-d1f1448276d4
> > <
> http://www.cpantesters.org/cpan/report/dc17e330-900b-11ec-bfc9-d1f1448276d4>
>
> > but I didn't exactly know why.
> > I thought it must have been some in-house patch that OpenBSD 7.0 was
> using.
> > But /now/ I can see that it arose because USE_LOCALE_NUMERIC was not a
> > compile-time option.
> >
>
> I had not been following that closely. Does this just mean you wasted
> some time, or is there something needed to be done?
>
>
It was just that I had spent some time puzzling over why it was that a
behaviour that I thought had disappeared with the arrival of perl-5.20.0
was now showing up on a system perl-5.32.1 (in OpenBSD-7.0).
And it was just down to the absence of USE_LOCALE_NUMERIC, which was a
known issue (but not known by me until a day or so ago).

It's good that this has all been fixed up.

When overloading arithmetic operators that accept both strings and NVs, I
need to know which slot to use when the SV is both POK and NOK, because
they can contain different values.
Unless I'm on a pre-5.20 perl or a perl built without USE_LOCALE_NUMERIC,
then I just use the value in the PV slot, because it will have been a
string first. Otherwise, I need to work out whether it was originally a
string or originally a number.
IOW, I need to keep a close eye on the flags of NVs - which I do.

> One question (and sorry for the diversion), but can you tell me how the
> > absence of this compile-time option is reflected in %Config ?
> > Obviously, I'm assuming that there /is/ some %Config setting that
> > reflects this case. However, I couldn't find it.
>
> Finding what locale things are enabled is complicated, so I wrote
> t/loc_tools.pl. That ay have a sub you could use, or you can examine it
> to see what it does


Thanks for the pointer, Karl.
I was also pointed to the Config::bincompat_options() function which
returns an array that I could grep for USE_LOCALE_NUMERIC. (That was on
perlmonks, where I posted last night after having at that time received no
response here.)
But, in the absence of a %Config key, the simplest portable way (as in
fewest keystrokes) in a perl script is to check the condition (`$^X -V` =~
/\bUSE_LOCALE_NUMERIC\b/).
Or so it seems to me ....

Cheers,
Rob

>

PR #19523, opened 2022-03-11
Add MANIFEST.SKIP file

Leont:
This replaces the lists of files to skip in `Porting/manicheck`, `t/porting/manifest.t` and `Porting/makerel` with a centralized and standardized `MANIFEST.SKIP` file.
    
This would also allow us to add a make manifest target, but I left that for a follow-up PR.

PR #19542, opened 2022-03-17
Add and use MANIFEST.SKIP file for Porting files

atoomic:
This change is merging the two ideas from @demerphq  #19513 and @Leont  #19523
by relying on a traditional MANIFEST.SKIP file.

Case #19523 simplifies the way we can easily exclude common noise
Case #19513 provides a mechanism to avoid shipping not necessary
files as part of the tarball

This is adding a new 'Porting/Manifest.pm' file which provides helpers
to list files from MANIFEST taking into account the 'MANIFEST.SKIP'
using ExtUtils::Manifest.


Leont:
I think merging these branches into one bigger branch is the opposite of what we need here. This should have been at least three IMNSHO.

atoomic:
note: I preserved your original commits and built this on top of your changes
IMO the `MANIFEST.SKIP` change by itself is incomplete without integrating the effort made by @demerphq to adjust the tooling and tests.

jkeenan:
@atoomic, this p.r. has been inactive for over two years and has acquired numerous merge conflicts.  It appears to me that we're not enthusiastic about applying it.  Can it be closed?  Thanks.

jkeenan:
> @atoomic, this p.r. has been inactive for over two years and has acquired numerous merge conflicts. It appears to me that we're not enthusiastic about applying it. Can it be closed? Thanks.

No response from OP or other commenters.  Closing.

PR #19545, opened 2022-03-19
Perl_reg_named_buff_fetch: simplify newSVsv(&PL_sv_undef)

richardleach:
Specifically, `newSVsv(&PL_sv_undef)` reduces to just `newSV_type(SVt_NULL)`.

This seems to be the only instance of `newSVsv(&PL_<some constant>)` in core.


demerphq:
LGTM - Merging.

PR #19558, opened 2022-03-23
gh19557: restore match_end on early bailout

hvds:
After 271c3af797, early bailout from the inner one of a pair of nested
lookbehinds would leave the desired match_end pointing at the wrong
place, so the outer lookbehind could give the wrong answer.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -9484,6 +9484,7 @@ NULL
 
                 /* If the lookbehind doesn't start in the actual string, is a
                  * trivial match failure */
+                match_end = ST.prev_match_end;
                 if (logical) {
                     logical = 0;
                     sw = 1 - cBOOL(ST.wanted);
--- a/t/re/re_tests
+++ b/t/re/re_tests
@@ -2073,8 +2073,10 @@ AB\s+\x{100}	AB \x{100}X	y	-	-
 /(?=(\w))(?<!^)/	1234	y	$1	2	-	# Lookahead with zero width negative lookbehind
 /(\w)(?<!^)/	1234	y	$1	1	-	# Negative lookbehind with zero width assertion
 /(?=foo)(?<=(a??))/	afoo	y	$1	a	-	# Postive lookbehind with minmod (leftmost longest)
-/(?=foo)(?<=(a{0,2}))/	aafoo	y	$1	aa	-	# Postive lookbehind with min/max (leftmost longest)
-/(?=foo)(?<=(|a|aa))/	aafoo	y	$1	aa	-	# Postive lookbehind with alternation (leftmost longest)
+/(?=foo)(?<=(a{0,2}))/	aafoo	y	$1	aa	-	# Positive lookbehind with min/max (leftmost longest)
+/(?=foo)(?<=(|a|aa))/	aafoo	y	$1	aa	-	# Positive lookbehind with alternation (leftmost longest)
+(?<=(?<!yz)b)c	aabcdef	y	$&	c	-	# Nested lookbehind without early bailout
+(?<=(?<!yz)b)c	abcdef	y	$&	c	-	# Nested lookbehind with early bailout
 
 # Keep these lines at the end of the file
 # pat	string	y/n/etc	expr	expected-expr	skip-reason	comment


hvds:
@demerphq would it be a better idea to store and access `match_end` as a regmatch_state variable directly, as we do with a bunch of the other state variables? ie rename `prev_match_end` to `match_end` and replace most references to `match_end` with `ST.match_end`. That feels like it would reduce the risk of bugs like this.

khwilliamson:
This PR LGTM.  I think @hvds idea about using the stacked version always is a good idea

hvds:
Hmm, got a wierd one-off test failure for "macOS (catalina) xcode 12 (-Uusethreads) " in https://github.com/Perl/perl5/runs/5661251847?check_suite_focus=true:

```
re/charset.t ......................................................... ok
Use of uninitialized value $_ in substitution (s///) at re/anyof.t line 79.
Use of uninitialized value $_ in substitution (s///) at re/anyof.t line 80.
Use of uninitialized value $_ in substitution (s///) at re/anyof.t line 83.
Use of uninitialized value $_ in substitution (s///) at re/anyof.t line 84.
re/fold_grind_8.t .................................................... ok
re/fold_grind_a.t .................................................... ok
re/fold_grind_aa.t ................................................... ok
re/fold_grind_d.t .................................................... ok
re/fold_grind_l.t .................................................... ok
re/fold_grind_T.t .................................................... skipped: Couldn't find a UTF-8 turkic locale
Use of uninitialized value $expected in substitution (s///) at re/anyof.t line 879.
Use of uninitialized value $expected in concatenation (.) or string at re/anyof.t line 881.
Use of uninitialized value $expected in pattern match (m//) at re/anyof.t line 885.
# Failed test 434 - Verify compilation of (?8)[\x{103}] displays as  at re/anyof.t line 890
#      got "EXACT_REQ8 <\\x{103}>"
# expected undef
re/fold_grind_u.t .................................................... ok
```

Looking at a normal run that test gives me:
```
ok 434 - Verify compilation of (?8)[\x{103}] displays as EXACT_REQ8 <\x{103}>
```

So it seems that somehow the testscript got undef for $expected here, which seems insane. (Maybe the uninit warnings in re/fold_grind_8.t above are also related.) @khwilliamson does this make any sense to you?

hvds:
> I think @hvds idea about using the stacked version always is a good idea

I had a go at this, but I'm not sure that the relevant state object is available when we get to LOOKBEHIND_END. Certainly my simplistic attempt to hack it (using `st->u.ifmatch.match_end`) resulted in lots of test failures.

> This PR LGTM.

Thanks; I'll wait a couple of days in case @demerphq shows up, but if not I'll merge it.

khwilliamson:
That one-off test looks like the compilation is failing or is returning unexpected results; I pushed to blead a new version of anyof.t that outputs the details should this ever happen again.

hvds:
Now merged as 9e3e2820c0.

demerphq:
On Wed, 23 Mar 2022 at 19:55, Hugo van der Sanden ***@***.***>
wrote:
>
> I think @hvds idea about using the stacked version always is a good idea
>
> I had a go at this, but I'm not sure that the relevant state object is
available when we get to LOOKBEHIND_END. Certainly my simplistic attempt to
hack it (using st->u.ifmatch.match_end) resulted in lots of test failures.
>
> This PR LGTM.
>
> Thanks; I'll wait a couple of days in case @demerphq shows up, but if not
I'll merge it.

Sorry about that, some personal issues got in the way.

FWIW, I too thought I could do the kind of hack you tried there and was
foiled. The problem is the regex engine stack does not work like you might
think. It exists only to serve backtracking, and the top of the stack does
not represent the current state. The state of the regex engine is the
context vars in the sub. When a regop which can backtrack executes it
injects a stack frame which preserves whatever context vars it will use,
and then uses the context vars for its purposes. Only the regops listed in
the second section of regcomp.sym with definitions like this:

IFMATCH         A:FAIL

and their auto generated helper states/regops are allowed to interact with
the stack. The regex engine arranges that when the IFMATCH_A or
IFMATCH_A_fail execute the top frame of the stack will be the one created
by IFMATCH, but no other regop may make that assumption.

Lookbehind is an "independent" subexpression, like several others, it
throws away its backtracking when it is done. Before I added LOOKBEHIND_END
it would have been terminated by a SUCCEED regop. Such regops are type
"END", which mean they have no "next regop", and cause the regex engine to
unwind the stack.

./perl -Ilib -Mre=Debug,ALL,STACK
-le'"boardboardx"=~/(?<=(foo|board){1,3})x/'

shows what happens with the stack.  The bottom looks like this:
  10 <boardboard> <x>        |   5|      18:LOOKBEHIND_END(0)
                             |   5|      SUCCEED: subpattern success...
  10 <boardboard> <x>        |   5|      pop (yes) WHILEM_B_max[Y]
  10 <boardboard> <x>        |   4|     pop (yes) WHILEM_A_max
  10 <boardboard> <x>        |   3|    pop (yes) WHILEM_A_pre
  10 <boardboard> <x>        |   2|   pop (yes) CURLYX_end[Y]
  10 <boardboard> <x>        |   1|  pop (yes) IFMATCH_A[Y]
  10 <boardboard> <x>        |   0| 20:EXACT <x>(22)
  11 <boardboardx> <>        |   0| 22:END(0)

Where you can see that when the LOOKBEHIND_END is executed the top stack
frame is for WHILEM_B_max, and IFMATCH_A is burried 5 levels deep.  If we
did the alignment test in IFMATCH_A we would have thrown away the
backtracking state from inside of the lookbehind, eg, much much too late.

So the alignment check must in LOOKBEHIND_END, and it must be against a
context var, as LOOKBEHIND_END has no reliable access to the stack frame
correlating with IFMATCH_A. In fact the general pattern is that the stack
exists to unwind changes to the global state, and everything else uses the
global state.

I havent reviewed the patch closely. but at a quick glance it made sense to
me.

Cheers,
yves



-- 

PR #19568, opened 2022-03-30
Add support for ARM64 in CPAN Win32

p-b-o:
Add support for ARM/ARM64 architecture in cpan/Win32 module.
Fix for test cpan/Win32/t/Names.t to handle this platform.

---

Implemented function Win32::GetChipArch(). Win32::GetChipName() uses a
deprecated member of struct SYSTEM_INFO ("An obsolete member that is
retained for compatibility")

Details: https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info

In cpan/Win32/t/Names.t, added explicit values possible for
Win32::GetChipName() and Win32::GetChipArch().

---

Ran tests on arm64, x64 platforms.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1099,6 +1099,7 @@ Philip Newton                  <pne@cpan.org>
 Philippe Bruhat (BooK)         <book@cpan.org>
 Philippe M. Chiasson           <gozer@ActiveState.com>
 Pierre Bogossian               <bogossian@mail.com>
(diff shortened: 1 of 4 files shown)


Leont:
This module is maintained externally, you should open a PR at https://github.com/perl-libwin32/win32 instead

p-b-o:
Thanks, I'll do it there.

Issue #19569, opened 2022-03-30
Perl Storable module should ask for Test::More version >= 0.82


kbulgrien:
This is a bug report for perl from kevinb@systemsdesignusa.com,
generated with the help of perlbug 1.35 running under perl v5.8.8.

-----------------------------------------------------------------

Apologies if this is the incorrect reporting forum.  I've not run across a
module that points issues to the perl list issue tracker.  Also it was not
clear at all what "category" to mention in perlbug.  Forging ahead:

A recent build of https://metacpan.org/pod/Storable 3.25, on a very old
system with perl 5.8.8, incurred a testing error due to Test::More being
too old.

Storable should assert a Test::More minimum version of 0.82 since
  https://metacpan.org/pod/Test::More#COMPATIBILITY
indicates that note first appeared in Test::More 0.82.

The supporting test log snippet:

```
t/canonical...........Unquoted string "note" may clash with future reserved
word at t/canonical.t line 37.
String found where operator expected at t/canonical.t line 37, near "note
"Will use Digest::MD5""
        (Do you need to predeclare note?)
syntax error at t/canonical.t line 37, near "note "Will use Digest::MD5""
Execution of t/canonical.t aborted due to compilation errors.
# Looks like your test died before it could output anything.
1..8
dubious
        Test returned status 255 (wstat 65280, 0xff00)
DIED. FAILED tests 1-8
        Failed 8/8 tests, 0.00% okay
```

This error is due to a Test::More that has not yet introduced the "note" key
word.  Installing Test-Simple-1.302188 resolved the test issue.

```
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl v5.8.8:

Configured by nobody at Thu Jun 22 10:41:06 PDT 2006.

Summary of my perl5 (revision 5 version 8 subversion 8) configuration:
  Platform:
    osname=sco, osvers=5.0.7, archname=i586-pc-sco3.2v5.0
    uname='sco_sv keats 3.2 5.0.7 i386 '
    config_args='-Dcc=gcc -Dprefix=/usr -Dmake=gmake -Dhintfile=sco_5_0_7.sh -Dmksymlinks -d -e -s'
    hint=recommended, useposix=true, d_sigaction=define
    usethreads=undef use5005threads=undef useithreads=undef usemultiplicity=undef
    useperlio=define d_sfio=undef uselargefiles=undef usesocks=undef
    use64bitint=define use64bitall=undef uselongdouble=define
[...]

PR #19570, opened 2022-03-30
Test::More::note() needs version 0.82

jkeenan:
For: https://github.com/Perl/perl5/issues/19569, as reported by
kbulgrien.

### DIFF
--- a/dist/Storable/Makefile.PL
+++ b/dist/Storable/Makefile.PL
@@ -29,7 +29,7 @@ WriteMakefile(
                 'ExtUtils::MakeMaker' => '6.31',
             },
             TEST_REQUIRES => {
-                'Test::More' => '0.41',
+                'Test::More' => '0.82',
             },
            )
         : () ),


Grinnz:
My only concern with this is Test::More also depends on Storable, so I would want to make sure this would not create a dependency cycle on ancient Perls when trying to upgrade these modules.

Grinnz (approved):
I've checked and Test::More's dependency on Storable is versionless, so there should be no upgrade cycle.

jkeenan:
> My only concern with this is Test::More also depends on Storable, so I would want to make sure this would not create a dependency cycle on ancient Perls when trying to upgrade these modules.

I would never have guessed it!  However, it seems that that dependency is on the `Test-Simple` *distribution*, not on the `Test::More` *module*.
```
$ cd cpan/Test-Simple/

[perlmonger: Test-Simple] $ ack Storable .
t/Test2/modules/IPC/Driver/Files.t
243:    Storable::store({}, $fn);
251:    Storable::store(bless({}, 'Test2::Event::FakeEvent'), $fn);
259:    Storable::store(bless({}, 'Test2::API'), $fn);
267:    Storable::store(bless({}, 'Foo'), $fn);

lib/Test2/IPC/Driver/Files.pm
13:use Storable();
185:        Storable::store($e, $file);
349:    my $obj = Storable::retrieve($file);

lib/Test2/API/InterceptResult.pm
9:use Storable     qw/dclone/;
287:structure is duplicated. This uses C<dclone> from L<Storable> to achieve the

lib/Test2/API/InterceptResult/Event.pm
11:use Storable qw/dclone/;
```
I think we can live with that.

khwilliamson (commented):
My review was requested, but I have no insight or knowledge into its correctness

Grinnz:
> > My only concern with this is Test::More also depends on Storable, so I would want to make sure this would not create a dependency cycle on ancient Perls when trying to upgrade these modules.
> 
> I would never have guessed it! However, it seems that that dependency is on the `Test-Simple` _distribution_, not on the `Test::More` _module_.

There is no distinction - dependencies are always on modules, and satisfied by a distribution.

Grinnz:
The issue I mentioned would occur if both Test::More and Storable had dependencies on each others' distribution, on versions of module which are not present in core in a particular version of Perl (i.e. 5.8.9 does not ship with Test::More 0.82). But since the dependency on Storable is versionless, it is satisfied by default on any Perl that ships with Storable.

jkeenan:
Addressed by https://github.com/Perl/perl5/pull/19570; closing ticket.  Thanks.

Issue #19573, opened 2022-03-30
IPC::Open3 - Fails if STDOUT is not opened to fileno 1 in caller


dfskoll:
If STDOUT is not open with fileno 1, then IPC::Open3 can fail.  For example, if STDOUT is opened for in-memory I/O, then its fileno is -1 and the file descriptors in the child process are not duplicated properly.  The attached test program illustrates the issue; expected output is:
```
Without messing with STDOUT:
Received: Hello
With STDOUT changed to in-memory I/O:
Received: Hello
```

but actual output is:

```
Without messing with STDOUT:
Received: Hello
With STDOUT changed to in-memory I/O:
Received: NOTHING RECEIVED
```

Open3 should make sure in the child that STDIN/STDOUT/STDERR are open with fileno 0, 1 and 2 before it dups the file descriptors.
[open3-bug-test_pl.txt](https://github.com/Perl/perl5/files/8383789/open3-bug-test_pl.txt)

dfskoll:
The attached patch seems to fix it, but I haven't tested extensively.  It's patched against the version shipped with Debian 11; the patch applies right after the `untie` calls in the child process.
[open3-patch.txt](https://github.com/Perl/perl5/files/8383915/open3-patch.txt)

dfskoll:
This patch (against the version of `IPC::Open3` shipped with Debian 11) seems to fix it.  Haven't done extensive testing.
[open3-patch.txt](https://github.com/Perl/perl5/files/8383943/open3-patch.txt)

Grinnz:
As I alluded to in the PR, most things cannot be expected to function correctly if the program has closed STDIN, STDOUT, or STDERR, and then opened filehandles in their place (in-memory or otherwise). Programs should instead reopen the global handles to `File::Spec->devnull` if they are attempting to daemonize.

dfskoll:
I disagree.  There are many use-cases for opening STDOUT to an in-memory string.  In our specific case, it's a Plack service that is accumulating its result in memory before sending it to the browser.

The problem is that IPC-Open3 conflates STDOUT with file descriptor 1.  Once we remove that assumption, it can easily be made to work without affecting existing code (see https://github.com/Perl/perl5/pull/19582) So I see absolutely no downside to fixing it.

Grinnz:
I would recommend looking at how [Capture::Tiny](https://metacpan.org/pod/Capture::Tiny) accomplishes such things without breaking forks.

dfskoll:
[...]

PR #19575, opened 2022-03-31
Make IPC-Open3 work as expected with localized or open-for-memory-io STDIN/STDOUT

dfskoll:
This makes Open3 work as expected (at least, as **I** expected) even with a localized STDOUT or a STDOUT open for in-memory I/O.

See https://github.com/Perl/perl5/issues/19573

### DIFF
--- a/ext/IPC-Open3/lib/IPC/Open3.pm
+++ b/ext/IPC-Open3/lib/IPC/Open3.pm
@@ -257,6 +257,16 @@ sub _open3 {
 		untie *STDOUT;
 		untie *STDERR;
 
+		if ((fileno(STDIN)//-1) != 0) {
+                        eval { close(STDIN); };
+                        open(STDIN, '<&=', '0');
+		}
+
+		if ((fileno(STDOUT)//-1) != 1) {
+                        eval { close(STDOUT); };
+                        open(STDOUT, '>&=', '1');
+		}
+
 		close $stat_r;
 		require Fcntl;
 		my $flags = fcntl $stat_w, &Fcntl::F_GETFD, 0;
--- a/ext/IPC-Open3/t/IPC-Open3.t
+++ b/ext/IPC-Open3/t/IPC-Open3.t
@@ -256,7 +256,7 @@ SKIP: {
     }
 }
 
-# Test that STDOUT open for in-memory I/O does not caus trouble.
+# Test that STDOUT open for in-memory I/O does not cause trouble.
 # Github issue https://github.com/Perl/perl5/issues/19573
 {
         my $buffer;
--- a/ext/IPC-Open3/t/IPC-Open3.t
+++ b/ext/IPC-Open3/t/IPC-Open3.t
@@ -14,7 +14,7 @@ BEGIN {
 }
 
 use strict;
-use Test::More tests => 45;
+use Test::More tests => 47;
 
 use IO::Handle;
 use IPC::Open3;
@@ -219,12 +219,9 @@ foreach my $handle (qw (DUMMY STDIN STDOUT STDERR)) {
     };
     is($@, '', "No errors with localised $handle");
     cmp_ok($pid, '>', 0, "Got a pid with localised $handle");
-    if ($handle eq 'STDOUT') {
-	is(<$out>, undef, "Expected no output with localised $handle");
-    } else {
-	like(<$out>, qr/\A# $handle\r?\n\z/,
-	     "Expected output with localised $handle");
-    }
+
+    like(<$out>, qr/\A# $handle\r?\n\z/,
+         "Expected output with localised $handle");
     waitpid $pid, 0;
 }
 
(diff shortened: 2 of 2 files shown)


jkeenan:
> This makes Open3 work as expected (at least, as **I** expected) even with a localized STDOUT or a STDOUT open for in-memory I/O.
> 
> See #19573

@dfskoll, I'd like to evaluate your pull request in a test-driven-development manner, *i.e.,* I'd like to run your added tests, see them fail, then add your changes to the `*.pm` file, re-run the tests and see them pass.

To do so, I recommend you (i) reformulate the pull request as two commits, the first strictly for the test and the second for the source code; (ii) give your branch a name other than `blead`, something like `ipc-open3-localized` perhaps.

Thank you very much.
Jim Keenan

ext/IPC-Open3/lib/IPC/Open3.pm
```diff
@@ -257,6 +257,16 @@ sub _open3 {
 		untie *STDOUT;
 		untie *STDERR;
 
+		if ((fileno(STDIN)//-1) != 0) {
+                        eval { close(STDIN); };
```
Grinnz, line 261: This is a huge hammer with global program effects at a distance and I don't think it's a good idea. This problem only occurs when STDIN or STDOUT (or STDERR) is closed, and I would consider that the bug in any such program - for the common task of daemonization, one should instead reopen them to `File::Spec->devnull`. Perhaps the documentation could mention this.

Grinnz: To put it more simply, if STDIN is not fd 0, then this line is closing an arbitrary filehandle we don't own. I don't consider that acceptable behavior.

dfskoll: This is happening in the _child_ process, which is about to `exec` another program.  If STDIN/STDOUT do not refer to UNIX fds 0/1 respectively, then there is _no way_ to set up the pipes correctly.

I consider it a serious defect that `open3` can fail because of what STDIN/STDOUT refer to in the parent process.

I would accept changing the patch to only open STDIN/STDOUT if they have no associated fd (meaning they're open for in-memory I/O), but I think that's wrong.

Your remarks about global effects at a distance are not relevant because as mentioned, we're about to exec a program anyway.

Grinnz: Yes, the defect is that STDIN/STDOUT have been broken by the parent process.

dfskoll:
On Fri, 01 Apr 2022 15:21:18 -0700
James E Keenan ***@***.***> wrote:

[...]

> To do so, I recommend you (i) reformulate the pull request as two
> commits, the first strictly for the test and the second for the
> source code; (ii) give your branch a name other than `blead`,
> something like `ipc-open3-localized` perhaps.

OK; I can do that.  I'll rework.

Regards,

Dianne.

ext/IPC-Open3/lib/IPC/Open3.pm
Grinnz: The global effect is relevant because closing the handle in the child process affects the underlying handle.

dfskoll: Actually, another option would be to `fdopen` our own private handles to fd 0 and 1 respectively, and use those (rather than STDIN/STDOUT) to set up the pipes.

Basically, we need a way to get Perl to set up pipes using POSIX file descriptors 0 and 1 regardless of how the caller might have mucked about with STDIN and STDOUT.  Because ultimately it's the file descriptors that matter, not the Perl filehandles.

It's the weekend, but I'll look at this on Monday.

Grinnz: Keep in mind also that IPC::Open3 does not always exec:

```
Note if you specify - as the command, in an analogous fashion to open(my $fh, "-|") the child process will just be the forked Perl process rather than an external command.
```

dfskoll:
On Fri, 01 Apr 2022 16:34:40 -0700
Dan Book ***@***.***> wrote:

> Yes, the defect is that STDIN/STDOUT have been broken by the parent
> process.

That should not make open3 fail.  A parent process may have very
good reasons for redirecting STDOUT to in-memory I/O but still want
to use IPC-Open3.

As I mentioned in another comment, I believe the proper fix is for open3
not to refer to STDIN or STDOUT at all (which are global file handles that
might be open to who knows where), but rather open its own private file
handles to file descriptors 0 and 1, like this:

```
   my ($private_stdin, $private_stdout);

   open($private_stdin, '<&=', '0');
   open($private_stdout, '>&=', '1');

   # Use $private_stdin and $private_stdout to set
   # up pipes rather than STDIN and STDOUT
```

I will rework the patch to work this way instead.

Regards,

Dianne.

dfskoll:
On Fri, 01 Apr 2022 16:38:32 -0700
Dan Book ***@***.***> wrote:

> Keep in mind also that IPC::Open3 does not always exec:

OK.  In the case where it doesn't exec, then definitely STDIN and STDOUT
in the child process must be reset to refer to fd 0 and 1 respectively,
or else the documentation should me amended to say that if the file
descriptors have been changed then open3 won't work.

Regards,

Dianne.

dfskoll:
This PR is replaced with and superseded by https://github.com/Perl/perl5/pull/19582

dfskoll:
Hi, everyone,

OK, I've made a new PR with changes suggested by James Keenan.

The new PR is at https://github.com/Perl/perl5/pull/19582

It has two commits that just update the test file so you can see
that it failes, followed by a commit that fixes the tests.

This version does not touch STDIN or STDOUT.  Instead, it uses
completely private Perl filehandles that are connected to POSIX
file descriptors 0 and 1, respectively.

One thing to note is that it's UNIX-specific, because it opens
filehandles to /dev/null.  I have no access to non-UNIX systems,
so not sure how to handle it on those systems.

Regards,

Dianne.

Issue #16867, opened 2019-03-03
Assertion failure in S_regmatch (regexec.c:7279)


p5pRT:
Migrated from [rt.perl.org#133885](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133885) (status was 'open')

Searchable as RT133885$

p5pRT:
### From @dur-randir

#### Created by @dur-randir



While fuzzing perl v5\.29\.8\-21\-gde59f38ed9 built with afl and run
under libdislocator\, I found the following program

s\(\(?\{my sub f\}\)\(\)\)00

to cause an assertion failure&#8203;:

perl&#8203;: regexec\.c&#8203;:7295&#8203;: ssize\_t S\_regmatch\(regmatch\_info \*\, char \*\,
regnode \*\)&#8203;: Assertion \`o\->op\_type == OP\_NEXTSTATE || o\->op\_type ==
OP\_DBSTATE || \(o\->op\_type == OP\_NULL && \( o\->op\_targ == OP\_NEXTSTATE
|| o\->op\_targ == OP\_DBSTATE \) \)' failed\.

GDB stack trace is following&#8203;:

\#0  \_\_GI\_raise \(sig=sig@&#8203;entry=6\) at \.\./sysdeps/unix/sysv/linux/raise\.c&#8203;:50
\#1  0x00007ffff7c25535 in \_\_GI\_abort \(\) at abort\.c&#8203;:79
\#2  0x00007ffff7c2540f in \_\_assert\_fail\_base \(fmt=0x7ffff7d87ee0
"%s%s%s&#8203;:%u&#8203;: %s%sAssertion \`%s' failed\.\\n%n"\,
&nbsp;    assertion=0x10ec280 \<\.str\.318> "o\->op\_type == OP\_NEXTSTATE ||
o\->op\_type == OP\_DBSTATE || \(o\->op\_type == OP\_NULL && \( o\->op\_targ ==
OP\_NEXTSTATE || o\->op\_targ == OP\_DBSTATE \) \)"\, file=0x10ca1e0 \<\.str\.2>
"regexec\.c"\, line=7295\, function=\<optimized out>\) at assert\.c&#8203;:92
\#3  0x00007ffff7c330f2 in \_\_GI\_\_\_assert\_fail \(
&nbsp;    assertion=0x10ec280 \<\.str\.318> "o\->op\_type == OP\_NEXTSTATE ||
[...]

Issue #17279, opened 2019-11-08
heap-buffer-overflow in Perl_my_atof3


dur-randir:
This is a bug report for perl from sergey.aleynikov@gmail.com,
generated with the help of perlbug 1.41 running under perl 5.31.6.

[Please describe your issue here]

While fuzzing perl v5.31.5-213-g9bec17d7c built with afl and run
under libdislocator, I found the following program

0=~/\p{nV:-0}/

to cause heap-buffer-overflow. ASAN diagnostics are:

==44466==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000001375 at pc 0x000000db90b1 bp 0x7ffc4c3fe2f0 sp 0x7ffc4c3fe2e8
READ of size 1 at 0x602000001375 thread T0
    #0 0xdb90b0 in Perl_my_atof3 /home/afl/afl-asan/numeric.c:1564:14
    #1 0x816b5d in Perl_parse_uniprop_string /home/afl/afl-asan/regcomp.c:23624:24
    #2 0x88ce21 in S_regclass /home/afl/afl-asan/regcomp.c:17210:44
    #3 0x8673b8 in S_regatom /home/afl/afl-asan/regcomp.c:13538:19
    #4 0x849b82 in S_regpiece /home/afl/afl-asan/regcomp.c:12404:11
    #5 0x849b82 in S_regbranch /home/afl/afl-asan/regcomp.c:12324
    #6 0x7a6080 in S_reg /home/afl/afl-asan/regcomp.c:12026:10
    #7 0x78122f in Perl_re_op_compile /home/afl/afl-asan/regcomp.c:7738:9
    #8 0x55c406 in Perl_pmruntime /home/afl/afl-asan/op.c:8089:6
    #9 0x752a87 in Perl_yyparse /home/afl/afl-asan/perly.y:1260:23
    #10 0x614bfc in S_parse_body /home/afl/afl-asan/perl.c:2529:9
    #11 0x60a9b6 in perl_parse /home/afl/afl-asan/perl.c:1820:2
    #12 0x5352bd in main /home/afl/afl-asan/perlmain.c:132:18
    #13 0x7f7b3ec4509a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #14 0x43ccb9 in _start (/home/afl/afl-asan/perl+0x43ccb9)

0x602000001375 is located 0 bytes to the right of 5-byte region [0x602000001370,0x602000001375)
[...]

PR #19577, opened 2022-04-01
Accept \p{nv=-0}

khwilliamson:
Some platforms already ignore the minus sign.  The Unicode standard says
the minus sign should be ignored always.

(But always doing so would cause it to not properly get the only
negative numeric value character in Unicode; I have submitted a ticket
to Unicode about this.  Nevertheless, it is in keeping with the letter
and spirit of the standard for the leading minus to be ignored for -0)

Note that -0 not being treated as legal was being tested for.  This was the result of a fuzzing operation trying this out and getting assertion failures (#16867), and not because the old behavior was desired.  When this P.R is merged, I'll open a ticket to create a new test for that failure.  It's not immediately obvious to me, since it was generated by fuzzing.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -24679,6 +24679,15 @@ S_parse_uniprop_string(pTHX_
                     break;
                 }
             }
+
+            /* Turn nv=-0 into nv=0.  These should be equivalent, but vary by
+             * underling libc implementation. */
+            if (   i == name_len - 1
+                && name[name_len-1] == '0'
+                && lookup_name[j-1] == '-')
+            {
+                j--;
+            }
         }
     }
     else {  /* No '=' */
--- a/t/lib/croak/regcomp
+++ b/t/lib/croak/regcomp
@@ -179,8 +179,3 @@ qr/(?<=(?<!x)x)\K/;
 qr/(?<!(?<!x)x)\K/;
 EXPECT
 OPTIONS nonfatal
-########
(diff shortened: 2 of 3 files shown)


khwilliamson:
github changed the issue to 16867 when I typed in the CORRECT ISSUE NUMBER which is https://github.com/Perl/perl5/issues/17279 which I got this time by giving it a full link which it then abbreviated

ilmari:
> the only negative numeric value character in Unicode;

In case anyone else was curious about this, the character is:

U+0F33 TIBETAN DIGIT HALF ZERO
༳
Category: No (Number, Other); East Asian width: N (neutral)
Unicode block: 0F00..0FFF; Tibetan
Numeric value: -1/2
Bidi: L (Left-to-Right)

khwilliamson:
On 4/1/22 08:50, Dagfinn Ilmari Mannsåker wrote:
>     the only negative numeric value character in Unicode;
> 
> In case anyone else was curious about this, the character is:
> 
> U+0F33 TIBETAN DIGIT HALF ZERO

And if you're really curious, read
https://www.babelstone.co.uk/Blog/2007/04/numbers-that-dont-add-up-tibetan-half.html

The --0.5 one is especially controversial, but I didn't see any articles 
about it in particular.  Apparently there is zero (no pun intended) 
evidence of it being actually used.

Unicode is more strict nowadays about what they encode.

Issue #19554, opened 2022-03-22
BBC: Blead breaks DYACOB/Convert-Number-Roman-0.01.tar.gz


andk:
**Description**

Test diagnostics:

``` text
Use of unassigned code point or non-standalone grapheme for a delimiter is not allowed at ../lib/Convert/Number/Roman.pm line 106, near ""
Use of unassigned code point or non-standalone grapheme for a delimiter is not allowed at ../lib/Convert/Number/Roman.pm line 162, near "=~ "
Compilation failed in require at t/numbers.t line 23.
BEGIN failed--compilation aborted at t/numbers.t line 23.
t/numbers.t .. 
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 96/96 subtests 
```

Bisect leads to:
``` text
32b87797e986f5d99836e16ea6b9d9ff5a56d3be is the first bad commit
commit 32b87797e986f5d99836e16ea6b9d9ff5a56d3be
Author: Karl Williamson <khw@cpan.org>
Date:   Sat Feb 12 21:47:13 2022 -0700

    toke.c: merge loops, multi-byte delim
```

**Steps to Reproduce**

cpan -i DYACOB/Convert-Number-Roman-0.01.tar.gz

**Expected behavior**

The module should test successfully and be installed

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
  Commit id: 32b87797e986f5d99836e16ea6b9d9ff5a56d3be
  Platform:
    osname=linux
    osvers=5.16.0-5-amd64
    archname=x86_64-linux
    uname='linux k93msid 5.16.0-5-amd64 #1 smp preempt debian 5.16.14-1 (2022-03-15) x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.35.9-157-g32b87797e9/da1c -Dmyhostname=k93msid -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Uuseithreads -Uuselongdouble -DEBUGGING=-g'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
[...]

PR #19579, opened 2022-04-01
toke.c: Return to pre-existing gcb rules

khwilliamson:
As stated in the C comments added in this commit, it has always worked
to say s/mark1/mark2/ where the marks are literal characters.  This is
incorrect because the marks should be considered modifiers of the '/'
delimiters.  (The substitute should be changed to do something like
s/\x{foo/\x{bar}/.)  But that may not happen in time for 5.36.  So this
commit reverts to  the behavior prior to 32b8779 for the ASCII delimiters that
have always worked this way.  Non-ASCII delimiters will continue to work
the new way, which could break code expecting the old.  But use of
non-ASCII delimiters is pretty rare, and the users of that could be
expected to have more knowledge of Unicode and would be less likely to
do something contrary to the Unicode Standard than someone using the
typical ASCII delimiters.

### DIFF
--- a/toke.c
+++ b/toke.c
@@ -11514,12 +11514,21 @@ Perl_scan_str(pTHX_ char *start, int keep_bracketed_quoted, int keep_delims, int
                  * incremented above 0, so will come here at the first closing
                  * delimiter.
                  *
-                 * Only grapheme delimiters are legal. */
-                if (   UTF  /* All Non-UTF-8's are graphemes */
-                    && UNLIKELY(! is_grapheme((U8 *) start,
-                                              (U8 *) s,
-                                              (U8 *) PL_bufend,
-                                              close_delim_code)))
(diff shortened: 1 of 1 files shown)


jkeenan:
Was there a specific commit at which we moved from the "pre-existing rules" to the existing rules?

khwilliamson:
I revised the commit message to include the breaking commit number

rjbs:
As I commented on #19554, I think we should leave blead as is and view this as as correction that does not need a long warning period.

demerphq:
Just to be clear, @rjbs you are saying we should close this PR and not apply?

rjbs:
That's my personal opinion.  I could certainly be outvoted, and possibly swayed.

demerphq:
Its been a long time with no activity on this, so I am going to close the ticket. @khwilliamson if you object please reopen.

Issue #19550, opened 2022-03-21
BBC: Blead breaks SISYPHUS/Math-MPFR-4.21.tar.gz, SISYPHUS/Math-GMPf-0.47.tar.gz, SISYPHUS/Math-Float128-0.15.tar.gz


eserte:
t/NOK_and_POL.t in Math-MPFR-4.21 started to fail with perl 5.35.10:
```
2 != 1

2 != 1

3 != 1

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

4 != 2

5 != 3

5 != 3
t/NOK_and_POK.t .................. 
Failed 20/27 subtests 
```

Math-GMPf-0.47 and Math-Float128-0.15 have a same-named test script, which also started to fail.

eserte:
@sisyphus FYI

jkeenan:
Bisecting with the following invocation:
```
perl Porting/bisect.pl --start=v5.35.9 --end=v5.35.10 --module=Math::MPFR
```
... pointed to the following commit:
```
b1e3ba27cdcf1b4e2558f05ed4cc6780d2f208e1 is the first bad commit
commit b1e3ba27cdcf1b4e2558f05ed4cc6780d2f208e1
Author: Graham Knop <haarg@haarg.org>
Date:   Tue Mar 15 15:19:03 2022 +0100
Commit:     Paul Evans <leonerd@leonerd.org.uk>
CommitDate: Tue Mar 15 15:44:55 2022 +0000


    always prevent setting POK flag when NV values are used as strings
```
@haarg, @leonerd, can you take a look?

hvds:
The test is specifically checking the NOK and POK flags on Inf and NaN values, so it is unsurprising that this change results in test failures.

Because most of the tests are checking the value of an accumulating count, it also fails on all of those tests after the first true failure: I suspect there are only 2 real test failures here.

My guess is that only the tests will need to change, but @sisyphus will need to compare against whatever he's actually trying to achieve here; I'll create an issue for the distro.

hvds:
Created https://github.com/sisyphus/math-mpfr/issues/4

haarg:
These tests are checking what the internal flags are after using values in various forms. This is sensitive adjustments to how core handles the flags, which have been intentionally changed in 5.35.10. I'd say that these dists need to have their tests adjusted.

[...]

PR #19580, opened 2022-04-01
fix SVf_POK getting set on NVs when copying

haarg:
When copying an SV with SVp_POK set, the SVf_POK flag gets enabled on
the destination SV in all cases, to allow some of the COW logic to work
properly. The flag is then meant to be disabled later in the routine, if
the original SV was not SVp_POK. This was happening for IOK values, but
not NOK values. This would only impact a few values (Inf, -Inf, NaN) as
the string form of other floats is sensitive to locale so would not be
stored. In the past, values like this would also always have SVf_POK
set, but this has recently been changed to allow tracking the origin
type of a value, exposing this bug.

Fix this by checking for this case and unsetting the flag as
appropriate for SVf_NOK values, the same as was previously being done
for SVf_IOK values.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -4547,6 +4547,16 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
         }
         if (sflags & SVp_NOK) {
             SvNV_set(dsv, SvNVX(ssv));
+            if ((sflags & SVf_NOK) && !(sflags & SVf_POK)) {
+                /* Source was SVf_NOK|SVp_NOK|SVp_POK but not SVf_POK, meaning
+                   a value set as floating point and later stringified, where
+                  the value happens to be one of the few that we know aren't
+                  affected by the numeric locale, hence we can cache the
+                  stringification. Currently that's  +Inf, -Inf and NaN, but
+                  conceivably we might extend this to -9 .. +9 (excluding -0).
+                  So mark destination the same: */
+                SvFLAGS(dsv) &= ~SVf_POK;
+            }
         }
         if (sflags & SVp_IOK) {
             SvIV_set(dsv, SvIVX(ssv));
--- a/t/op/svflags.t
+++ b/t/op/svflags.t
@@ -10,7 +10,7 @@ BEGIN {
 # Tests the new documented mechanism for determining the original type
 # of an SV.
 
-plan tests => 12;
+plan tests => 16;
 use strict;
 use B qw(svref_2object SVf_IOK SVf_NOK SVf_POK);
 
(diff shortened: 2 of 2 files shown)


haarg:
Fixes #19550

sisyphus:
The bug that this PR aims to fix is also reflected in the newly added builtins <code>created_as_string()</code> and <code>created_as_number()</code> :
```
use strict;
no warnings 'experimental';
use builtin qw(created_as_string created_as_number);

my $orig = 99 ** (99 ** 99); # Inf
eval{"$orig"};
my $copy = $orig;

print "ok 1\n" if created_as_number($orig);
print "not ok 2\n" if created_as_string($copy);
```
Outputs (on current blead):
```
ok 1
not ok 2
```

sisyphus:
I've now had time to check this PR on Windows.
It seems fine to me.
It fixes the odd behaviour described in #19550, the demo script I provided above then behaves as I expect, and the failures in Math-MPFR-4.21, Math-GMPf-0.47 & Math-Float128-0.15 no longer occur.

sv.c
```diff
@@ -4547,6 +4547,16 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
         }
         if (sflags & SVp_NOK) {
             SvNV_set(dsv, SvNVX(ssv));
+            if ((sflags & SVf_NOK) && !(sflags & SVf_POK)) {
+                /* Source was SVf_NOK|SVp_NOK|SVp_POK but not SVf_POK, meaning
+                   a value set as floating point and later stringified, where
+                  the value happens to be one of the few that we know aren't
+                  affected by the numeric local, hence we can cache the
```
khwilliamson: s/local/locale/

sv.c
```diff
@@ -4547,6 +4547,16 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
         }
         if (sflags & SVp_NOK) {
             SvNV_set(dsv, SvNVX(ssv));
+            if ((sflags & SVf_NOK) && !(sflags & SVf_POK)) {
```
demerphq, line 4550: maybe I am being predantic but I would have written that as 
```
if ((sflags & (SVf_NOK|SVf_POK)) == SVf_NOK) { 
```

khwilliamson:
On 4/6/22 06:14, Yves Orton wrote:
> 
> 
> In sv.c <https://github.com/Perl/perl5/pull/19580#discussion_r843873428>:
> 
>> @@ -4547,6 +4547,16 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
>           }
>           if (sflags & SVp_NOK) {
>               SvNV_set(dsv, SvNVX(ssv));
> +            if ((sflags & SVf_NOK) && !(sflags & SVf_POK)) {
> 
> maybe I am being predantic but I would have written that as
> 
> |if ((sflags & (SVf_NOK|SVf_POK)) == SVf_NOK) { |

I would have written it this way too.

sv.c
haarg: This is copying the exact code in the block after it, just with different flags. I don't want to change the IV block when not needed, but I also want this to match in terms of style.

I think this entire code path is in need of refactoring, so I would prefer to make it obvious that this is meant to match its sibling block.

demerphq: Fair enough. Ill pretend you have promised to the refactor.

demerphq (approved):
Approved on the assumption you are going to follow up with a refactor patch which does it the efficient way. :-)

PR #19581, opened 2022-04-01
Fix whitespace noise in Locale::MakeText

toddr:
Bump to 1.31 for release to CPAN.

### DIFF
--- a/dist/Locale-Maketext/ChangeLog
+++ b/dist/Locale-Maketext/ChangeLog
@@ -1,8 +1,12 @@
 Revision history for Perl suite Locale::Maketext
 
+2022-04-14
+   * Release 1.31 to CPAN
+   * White space cleanup.
+
 2022-01-14
    * Release 1.30 to CPAN
-   * Provide alternate methods allowlist and denylist for 
+   * Provide alternate methods allowlist and denylist for
      whitelist and blacklist, respectively
 
 2020-01-19
--- a/dist/Locale-Maketext/lib/Locale/Maketext.pm
+++ b/dist/Locale-Maketext/lib/Locale/Maketext.pm
@@ -25,7 +25,7 @@ BEGIN {
 }
 
 
-our $VERSION = '1.30';
+our $VERSION = '1.31';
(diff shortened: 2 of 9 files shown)


atoomic (approved):

atoomic:
need a version bump view: `porting/cmp_version.t `

atoomic (changes requested):
please bump version

jkeenan:
@toddr, I am very grateful that you have taken on the task of preparing Locale-Maketext for an updated release on CPAN -- but I think you're going about this backwards.

As is indicated by its positioning under `dist/` in the core distribution, Locale-Maketext is **blead-first**, not CPAN-first.  Changes should be made in the core distribution first, tested and verified there, then propagated out to CPAN.  If one follows this procedure, and if the changes entail modification of a `*.pm` or `*.xs` file, then `$VERSION` needs to be bumped in blead.  It can then be released on CPAN without further modification of `$VERSION`.

What you have done is to take Locale-Maketext version 1.30 from blead and then made more changes in the CPAN version -- changes which include changes in a `*.pm` file.  As @atoomic has pointed out, this means that if you try to import CPAN back into blead (as per this pull request), you need a version bump.  But I really hate to see unnecessary "devel version" numbers like 1.30_01 in the core distribution, particularly as we get close to our annual production release.  It just seems sloppy.

So why not simply make your whitespace changes in blead, bump `$VERSION` in `dist/Locale-Maketext/lib/Locale/Maketext.pm` to `1.31`, then release *that* to CPAN with no further modification of `$VERSION`.  That way, blead and CPAN end up with the same `$VERSION`.

toddr:
@jkeenan Bumped to 1.31 and changed the commit message.

atoomic (approved):
This LGTM and seems to have fixed concerns raised by @jkeenan 
thanks for the update

jkeenan:
toddr:lmt-130 branch rebased on blead, merged into blead, pushed to repo in commit f5469426bb.  Closing this ticket.  Thank you very much.

PR #19584, opened 2022-04-04
Add the new flow-control keywords to perlfunc.pod

leonerd:


demerphq:
LGTM. 

One minor comment, maybe it would be good to split this patch into two. One to make the list of control keywords a list of one per line to make the list diff friendly, and then a second to add the new items. Lists like this are a pain to read through diffs. This content isn't for a human as far as I understand it, so it shouldn't matter how its structured. Not a big deal but since you are changing this it might be a good time to take the opportunity to make it more diff friendly.

leonerd:
I'll reshape it in this PR; I don't see much need to do two separate PRs for it. Also added `isa` which I realised I'd forgotten.

Force-pushed; PTAL.

demerphq:
FWIW, I didnt mean you should make two PR's just to break the PR into two commits, one to reorder, and then one to add the new stuff. I guess "patch" is a bit ambiguous.

demerphq (approved):

PR #19591, opened 2022-04-06
`use VERSION` in docs

leonerd:
A sweep of all the `pod/*.pod` files, looking for places to update to mention `use VERSION`.

 * Split out the main docs in `perlfunc.pod` into its own section dedicated to the `use VERSION` syntax
 * Link to that from more places in the docs
 * Change code examples that used to start `use strict; use warnings;` into `use v5.36`
 * Updated v-less mentions of `use 5.0dd` into `use v5.dd`

### DIFF
--- a/pod/perldebtut.pod
+++ b/pod/perldebtut.pod
@@ -421,8 +421,8 @@ For more on references see L<perlref> and L<perlreftut>
 Here's a simple program which converts between Celsius and Fahrenheit, it too
 has a problem:
 
- #!/usr/bin/perl -w
- use strict;
+ #!/usr/bin/perl
+ use v5.36;
 
  my $arg = $ARGV[0] || '-c20';
 
--- a/pod/perldsc.pod
+++ b/pod/perldsc.pod
@@ -284,11 +284,11 @@ of the array pointed to by $AoA. If you wanted the C notion, you could
 write C<< $AoA[$i]->$* >> to explicitly dereference the I<i'th> item,
 reading left to right.
 
-=head1 WHY YOU SHOULD ALWAYS C<use strict>
+=head1 WHY YOU SHOULD ALWAYS C<use VERSION>
 
 If this is starting to sound scarier than it's worth, relax.  Perl has
-some features to help you avoid its most common pitfalls.  The best
-way to avoid getting confused is to start every program with:
+some features to help you avoid its most common pitfalls.  One way to avoid
+getting confused is to start every program with:
 
     use strict;
 
@@ -309,6 +309,19 @@ because you were accidentally accessing C<@aref>, an undeclared
 variable, and it would thereby remind you to write instead:
 
     print $aref->[2][2]
+
+Since Perl version 5.12, a C<use VERSION> declaration will also enable the
+C<strict> pragma.  In addition, it will also enable a feature bundle,
+giving more useful features.  Since version 5.36 it will also enable the
+C<warnings> pragma.  Often the best way to activate all these things at
+once is to start a file with:
+
+    use v5.36;
+
+In this way, every file will start with C<strict>, C<warnings>, and many
+useful named features all switched on, as well as several older features
+being switched off (such as L<C<indirect>|feature/The 'indirect' feature>).
+For more information, see L<perlfunc/use VERSION>.
 
 =head1 DEBUGGING
 X<data structure, debugging> X<complex data structure, debugging>
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -703,7 +703,7 @@ Portability issues: L<perlport/-X>.
 To avoid confusing would-be users of your code with mysterious
 syntax errors, put something like this at the top of your script:
 
-    use 5.010;  # so filetest ops can stack
+    use v5.10;  # so filetest ops can stack
 
 =item abs VALUE
 X<abs> X<absolute>
@@ -2082,8 +2082,8 @@ versions of Perl with mysterious syntax errors, put this sort of thing at
 the top of your file to signal that your code will work I<only> on Perls of
 a recent vintage:
 
-    use 5.012;	# so keys/values/each work on arrays
-    use 5.018;	# so each assigns to $_ in a lone while test
+    use v5.12;	# so keys/values/each work on arrays
+    use v5.18;	# so each assigns to $_ in a lone while test
 
 See also L<C<keys>|/keys HASH>, L<C<values>|/values HASH>, and
 L<C<sort>|/sort SUBNAME LIST>.
@@ -3821,7 +3821,7 @@ versions of Perl with mysterious syntax errors, put this sort of thing at
 the top of your file to signal that your code will work I<only> on Perls of
 a recent vintage:
 
-    use 5.012;	# so keys/values/each work on arrays
+    use v5.12;	# so keys/values/each work on arrays
 
 See also L<C<each>|/each HASH>, L<C<values>|/values HASH>, and
 L<C<sort>|/sort SUBNAME LIST>.
@@ -6406,7 +6406,7 @@ versions of Perl with mysterious failures, put this sort of thing at the
 top of your file to signal that your code will work I<only> on Perls of a
 recent vintage:
 
-    use 5.012; # so readdir assigns to $_ in a lone while test
+    use v5.12; # so readdir assigns to $_ in a lone while test
 
 =item readline EXPR
 
@@ -8343,7 +8343,7 @@ conversion flag.  Should you instead prefer an exception, do this:
 If you would like to know about a version dependency before you
 start running the program, put something like this at its top:
 
-    use 5.014;  # for hh/j/t/z/ printf modifiers
+    use v5.14;  # for hh/j/t/z/ printf modifiers
 
 You can find out whether your Perl supports quads via L<Config>:
 
@@ -8454,7 +8454,7 @@ L<C<srand>|/srand EXPR> uses that for the seed; otherwise it
 it returns the seed.  To signal that your code will work I<only> on Perls
 of a recent vintage:
 
-    use 5.014;	# so srand returns the seed
+    use v5.14;	# so srand returns the seed
 
 If L<C<srand>|/srand EXPR> is not called explicitly, it is called
 implicitly without a parameter at the first use of the
@@ -9924,7 +9924,7 @@ versions of Perl with mysterious syntax errors, put this sort of thing at
 the top of your file to signal that your code will work I<only> on Perls of
 a recent vintage:
 
-    use 5.012;	# so keys/values/each work on arrays
+    use v5.12;	# so keys/values/each work on arrays
 
 See also L<C<keys>|/keys HASH>, L<C<each>|/each HASH>, and
 L<C<sort>|/sort SUBNAME LIST>.
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -5119,7 +5119,7 @@ L<C<our>|/our VARLIST> declaration.  This applies immediately--even
 within the same statement.
 
     package Foo;
-    use strict;
+    use v5.36;  # which implies "use strict;"
 
     $Foo::foo = 23;
 
@@ -5136,7 +5136,7 @@ This works even if the package variable has not been used before, as
 package variables spring into existence when first used.
 
     package Foo;
-    use strict;
+    use v5.36;
 
     our $foo = 23;   # just like $Foo::foo = 23
 
@@ -5147,7 +5147,7 @@ long as there is no variable with that name is already in scope, you can then
 reference the package variable again even within the same statement.
 
     package Foo;
-    use strict;
+    use v5.36;
 
     my  $foo = $foo; # error, undeclared $foo on right-hand side
     our $foo = $foo; # no errors
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -9820,6 +9820,11 @@ different implementations, file naming conventions, and other
 infrastructure, this feature is now little used in practice and should be
 avoided in newly-written code.
 
+Care should be taken when using the C<no VERSION> form, as it is I<only>
+meant to be used to assert that the running Perl is of a earlier version
+than its argument and I<not> to undo the feature-enabling side effects
+of C<use VERSION>.
+
 =item utime LIST
 X<utime>
 
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -9653,8 +9653,6 @@ X<use> X<module> X<import>
 
 =item use Module
 
-=item use VERSION
-
 =for Pod::Functions load in a module at compile time and import its namespace
 
 Imports some semantics into the current package from the named module,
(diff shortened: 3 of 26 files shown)


pod/perlipc.pod
```diff
@@ -1632,8 +1619,7 @@ using select() to do a timed-out wait for I/O.  To do something similar
 with TCP, you'd have to use a different socket handle for each host.
 
  #!/usr/bin/perl
- use strict;
- use warnings;
+ use v5.3;6
```
haarg: typo

pod/perlobj.pod
```diff
@@ -751,6 +751,16 @@ appending "::" to it, like we saw earlier:
 
   my $file = new File:: $path, $data;
 
+Indirect object syntax is only available when the C<"indirect"> named
+feature is enabled. This is enabled by default, but can be disabled if
+requested.  This feature is present in older feature version bundles,
+but was removed from the :5.36 bundle; so a
```
haarg: Probably want `C<:5.36>`

pod/perlunicook.pod
```diff
@@ -696,17 +695,15 @@ When run, the following program produces this nicely aligned output:
...
 
-Here's that program; tested on v5.14.
+Here's that program.
 
  #!/usr/bin/env perl
  # umenu - demo sorting and printing of Unicode food
  #
  # (obligatory and increasingly long preamble)
  #
+ use v5.36;
```
haarg, line 705: This code uses prototypes, so this change would break it.

pod/perlfunc.pod
```diff
@@ -9644,6 +9644,65 @@ Breaks the binding between a variable and a package.
...
+Perl version is 5.35.0 or higher, L<warnings> are enabled.  Later use of
+C<use VERSION> will override all behavior of a previous C<use VERSION>,
+possibly removing the C<strict>, C<warnings>, and C<feature> added by it.
+C<use VERSION> does not load the F<feature.pm>, F<strict.pm>, or
+F<warnings.pm> files.
+
+In the current implementation, any explicit use of C<use strict> or
+C<no strict> overrides C<use VERSION>, even if it comes before it.
+However, this may be subject to change in a future release of Perl, so new
+code should not rely on this fact.  It is recommended that a
```
haarg: This is a policy change, not just a doc change. Doesn't seem appropriate to make without an actual deprecation.

pod/perlipc.pod
leonerd: Fixed

pod/perlobj.pod
leonerd: Fixed.

pod/perlunicook.pod
leonerd: Ugh. :( How annoying.

I wonder if I should revert it, or fix it up with `:proto(...)` attributes. Any preference?

pod/perlfunc.pod
leonerd: Hrm, it is a change but I'm fairly sure we already mentioned that, as part of adding the version-downgrade deprecation warning:

```
$ ./perl -e 'use v5.12; use v5'
Downgrading a use VERSION declaration to below v5.11 is deprecated, and will become fatal in Perl 5.40 at -e line 1.
```

pod/perlunicook.pod
haarg: Possibly just remove the prototypes? The only adjustment needed would be to change one of the calls to `colwidth` to explicitly pass `$_`.

leonerd: Fixed 5a03b3b9b5

scottchiefbaker:
Dang those are some hefty changes. I did a brief pass through all of them and nothing other than what @haarg mentioned stood out to me. Looks good to me.

pod/perlfunc.pod
```diff
@@ -9644,6 +9644,57 @@ Breaks the binding between a variable and a package.
...
+use of C<use strict> or C<no strict> overrides C<use VERSION>, even if it
+comes before it.  Later use of C<use VERSION> will override all behavior of
+a previous C<use VERSION>, possibly removing the C<strict>, C<warnings>,
+and C<feature> added by C<use VERSION>.  C<use VERSION> does not load the
+F<feature.pm>, F<strict.pm>, or F<warnings.pm> files.
+
+Specifying VERSION as a numeric argument of the form 5.024001 should
+generally be avoided as older less readable syntax compared to
+v5.24.1. Before perl 5.8.0 released in 2002 the more verbose numeric
+form was the only supported syntax, which is why you might see it in
```
ilmari: Might see it in what?

pod/perlintro.pod
```diff
@@ -102,6 +102,12 @@ enable both C<strict> and C<warnings>:
   #!/usr/bin/perl
   use v5.35;
 
+In addition to enabling the C<strict> and C<warnings> pragmata, this
+declaration will also activate a "feature bundle"; a collection of named
```
ilmari: Should this link to `feature`?

pod/perlstyle.pod
```diff
@@ -16,6 +16,13 @@ disable.  The B<-w> flag and C<$^W> variable should not be used for this
 purpose since they can affect code you use but did not write, such as
 modules from core or CPAN.
 
+A short and concise way to arrange for this is to use the
+L<C<use VERSION>|perlfunc/use VERSION> syntax, requesting a version 5.36
+or above, which will enable both the C<strict> and C<warnings> pragmata (as
+well as several other useful named features).
```
ilmari: This should link to `feature`, like `strict` and `warnings` do in the preceding paragraph.

pod/perlobj.pod
```diff
@@ -751,6 +751,16 @@ appending "::" to it, like we saw earlier:
 
   my $file = new File:: $path, $data;
 
+Indirect object syntax is only available when the C<"indirect"> named
+feature is enabled. This is enabled by default, but can be disabled if
```
ilmari: This should link to the docs for the `indirect` feature.

Grinnz:
I think there's a lot of important and good changes here. My only two concerns:

* We have to make sure to emphasize that `use 5.12` is not the same as `use v5.12`, and will break.
* I am somewhat concerned about the user that will read the latest Perl docs on the web, and try `use v5.36` in their Perl 5.24 and then get frustrated by the inability to continue running examples.

pod/perlref.pod
```diff
@@ -232,6 +232,10 @@ better to use the direct method invocation approach:
     $menubar = $main->Frame(-relief              => "raised",
                             -borderwidth         => 2)
 
+This indirect object syntax is only available when C<use feature "indirect">
+is in effect, and that is not the case when C<use v5.36> (or higher) is
```
ilmari: Should these link to the `feature` and `use VERSION` docs?

pod/perldsc.pod
```diff
@@ -310,6 +310,19 @@ variable, and it would thereby remind you to write instead:
...
+C<strict> pragma.  In addition, it will also enable a feature bundle,
+giving more useful features.  Since version 5.36 it will also enable the
+C<warnings> pragma.  Often the best way to activate all these things at
+once is to start a file with:
+
+    use v5.36;
+
+In this way, every file will start with C<strict>, C<warnings>, and many
+useful named features all switched on, as well as several older features
+being switched off (such as C<indirect>).  For more information, see
```
ilmari: This should link to the `feature` docs.

ilmari (changes requested):

skington:
Now that we've got `v5.36` hard-coded in a bunch of places, do we need to say "replace `use v5.36` with `use v5.38`" in the instructions about cutting a new release? Because otherwise we might forget, and we'll have code that says "use this specific recent version" rather than "use the latest version".

Grinnz:
No; the use of `v5.36` is specifically because that will be the first (stable) version declaration that enables warnings.

pod/perlfunc.pod
rjbs: I agree with @leonerd.  We have put in this deprecation specifically because we believe we are going to change the behavior.

pod/perlfunc.pod
leonerd: Oops, a line seems to have got lost in a copypaste. Have now restored it.

leonerd:
@ilmari Thanks for the comments. I've added more cross-linkage and addressed the other IRC comments. PTAL

leonerd:
> Now that we've got `v5.36` hard-coded in a bunch of places, do we need to say "replace `use v5.36` with `use v5.38`" in the instructions about cutting a new release? Because otherwise we might forget, and we'll have code that says "use this specific recent version" rather than "use the latest version".

That shouldn't be necessary. E.g. 5.14 added quite a few "use 5.014" or "use v5.14" examples, and those remained. The point of v5.36 here is that's the version that adds strict+warnings. We shouldn't be updating the versions hereafter, unless specific ones need specific tidying up (e.g. because 5.38 adds some builtins or whatever might come along).

leonerd:
@Grinnz 

>     * We have to make sure to emphasize that `use 5.12` is not the same as `use v5.12`, and will break.

Hmm. There are some words about that already, in the main `use VERSION` part of perlfunc. Do you think I should add more - there or elsewhere?

>     * I am somewhat concerned about the user that will read the latest Perl docs on the web, and try `use v5.36` in their Perl 5.24 and then get frustrated by the inability to continue running examples.

That's always been the case though; every perl version has some docs about a feature new in that version, which won't work when you copypaste the examples into older ones. Admittedly this set of changes does that in more places, but it's not fundamentally a new kind of problem. Specifically the failure in this case looks like:

```
$ perl
use v5.36;
Perl v5.36.0 required--this is only v5.34.0, stopped at - line 1.
BEGIN failed--compilation aborted at - line 1.
```

so I think it's fairly forgiving to end-users about what the issue is.

Grinnz:
> @Grinnz
> 
> > ```
> > * We have to make sure to emphasize that `use 5.12` is not the same as `use v5.12`, and will break.
> > ```
> 
> Hmm. There are some words about that already, in the main `use VERSION` part of perlfunc. Do you think I should add more - there or elsewhere?

I don't see anything about this specifically? Just about `use 5.012` vs `use v5.12`, which I don't have a strong opinion on. But it might be nice to mention `use 5.12` is `use v5.120`. But I didn't realize this was mostly just moving the existing docs around, so this doesn't need to be done in this PR.

ilmari (approved):

pod/perlfunc.pod
```diff
@@ -9796,17 +9755,71 @@ or no unimport method being found.
...
+(We try not to do this more than we have to.)
+
+Symmetrically, C<no VERSION> allows you to specify that you want a version
+of Perl older than the specified one.  Historically this was added during
+early designs of the Raku language (formerly "Perl 6"), so that a Perl 5
+program could begin
+
+    no 6;
+
+to declare that it is not a Perl 6 program.  As the two languages have
```
wolfsage, line 9818: I'm a little concerned that we're dropping the note that this does *not* disable feature bundles

pod/perlipc.pod
```diff
@@ -1455,8 +1444,7 @@ well, which is probably why it's spread to other systems.)
 Here's the code:
 
     #!/usr/bin/perl
```
wolfsage, line 1449: What are the chances these changes in the docs introduce breakage to the sample code beneat them? (Do we have a test that ensures these all compile/work...?)

wolfsage (commented):
There's that small thought of "If we remove 'use strict; use warnings;" from all of the samples, folks who read these docs but are trying to work with older perl might just remove the "use v5.36" to make their code work and not put in strict/warnings, but that's probably silly....

Anyway mostly seems good from a quick skim though, as noted, I worry about maybe samples breaking because of changes introduced by 'use v5.36'....

pod/perlintro.pod
```diff
@@ -102,6 +102,12 @@ enable both C<strict> and C<warnings>:
...
+features that enable many of the more recent additions and changes to the
+language, as well as occasionally removing older features found to have
+been mistakes in design and removed.
```
rjbs: I reckon "removed" should be "discouraged".  The feature flag means it's still there. 😉

pod/perlstyle.pod
```diff
@@ -16,6 +16,13 @@ disable.  The B<-w> flag and C<$^W> variable should not be used for this
 purpose since they can affect code you use but did not write, such as
 modules from core or CPAN.
 
+A short and concise way to arrange for this is to use the
```
rjbs: Redundant.  Just say concise.

pod/perldsc.pod
```diff
@@ -284,11 +284,11 @@ of the array pointed to by $AoA. If you wanted the C notion, you could
 write C<< $AoA[$i]->$* >> to explicitly dereference the I<i'th> item,
 reading left to right.
 
-=head1 WHY YOU SHOULD ALWAYS C<use strict>
+=head1 WHY YOU SHOULD ALWAYS C<use strict> or C<use VERSION>
```
rjbs: I'd drop "use strict".  v5.12 was 12 years ago.  We can just say to use the version bundle… or "use the version bundle and if you can't, at *least* use strict, good grief!"

pod/perlthrtut.pod
```diff
@@ -852,37 +852,36 @@ things we've covered.  This program finds prime numbers using threads.
...
-  32
-  33 my $stream = Thread::Queue->new(3..1000, undef);
-  34 check_num($stream, 2);
+   4 use v5.36;
+   5
+   6 use threads;
+   7 use Thread::Queue;
+   8
+   9 sub check_num {
+  10     my ($upstream, $cur_prime) = @_;
```
rjbs: Wanna stick that into a signature?

pod/perldebtut.pod
```diff
@@ -421,8 +421,8 @@ For more on references see L<perlref> and L<perlreftut>
 Here's a simple program which converts between Celsius and Fahrenheit, it too
 has a problem:
 
- #!/usr/bin/perl -w
- use strict;
+ #!/usr/bin/perl
+ use v5.36;
```
rjbs, line 425: Any removal of -w is music to my eyes

pod/perlipc.pod
```diff
@@ -1210,8 +1202,7 @@ network terminators here because Unix domain sockets are guaranteed
 to be on the localhost, and thus everything works right.
 
     #!/usr/bin/perl -T
```
rjbs, line 1207: We wanna drop that -T while we're in here?

pod/perlunicook.pod
```diff
@@ -756,17 +751,17 @@ Here's that program.
      printf " €%.2f\n", $price{$item};
  }
 
- sub pad($$$) {
+ sub pad {
      my($str, $width, $padchar) = @_;
```
rjbs: A great time to _also_ consider moving to signatures.

pod/perlfunc.pod
rjbs: Adding that back seems sensible to me.

rjbs (changes requested):
I made just a few requests/suggestions, but I think this is basically good to go!

pod/perlthrtut.pod
leonerd: Ah go on then. I've renumbered all the lines once already, one more time won't hurt. Much.

pod/perlipc.pod
leonerd: Now I thought about it, but later in the file there's lots of discussion about why you should put `-T` and run in taint mode.. :(

I didn't want to fight that battle at the same time in the same PR

pod/perlfunc.pod
leonerd: Ahyes I think that went missing in the copypaste movements. I'll return it back somewhere sensible.

leonerd:
Righty. Another round of updates, rebasing, and squashing. PTAL.

rjbs:
LGTM!

leonerd:
Rik says "LGTM" here and on IRC but hasn't ticked the box on github. Will merge this anyway despite the UI's objections.

PR #19597, opened 2022-04-08
large hashes: temporarily disable new optimization

rjbs:


demerphq:
I am not sure that the compiler will optimize this away.  I would prefer we define a macro:

```
#ifdef USE_UNSHARED_KEYS
#define LARGE_HASH_HEURISTIC(hv,newmax) S_large_hash_heuristic(aTHX_ hv, newmax)
#else
#define LARGE_HASH_HEURISTIC(hv,newmax) 0
#endif
```
and then replace the calls to the S_large_hash_heuristic wit LARGE_HASH_HEURISTIC()

I can do the patch if you want.

demerphq:
See https://github.com/Perl/perl5/pull/19598 for alternative approach.

PR #19598, opened 2022-04-08
hv.* - disable unshared keys except when requested

demerphq:
This disables the new unshared hash key logic by default. People
who wish to try it out can enable it at build time with

    -DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES

in the configure process.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -36,6 +36,67 @@ distinctions, this can be useful when serialising or encoding Perl data
 structures for export to other languages or data formats where such a
 distinction does matter.
 
+=head2 Support for not sharing all keys in large hash tables.
+
+This release includes support to build Perl so that it does not share
+all keys in hash tables as has been historically the case since Perl
+v5.6. The shared key behavior was added to Perl originally to support
+having many objects with the same keys without having to store each key
+many times. For objects that might be created hundreds or thousands of
+times this key sharing behavior can result in using significantly less
+memory. However key sharing comes at the cost of having to maintain an
+internal "master hash table" (PL_str_tab), and having to do two store
+operations per key stored in a hash table. It also can result in
+surprising consequences with programs that fork, causing the master hash
+table to be COWed (copy on write) into the memory space for each forked
+process which can, in worst case, greatly increase the memory
+utilization of such scripts.
+
+The new behavior is an attempt to have our cake and eat it too. Small
+hash tables will continue to share keys, but once the hash table is used
+to store more than a certain number of keys (currently 64) it will not
+share the new keys that are added. This should benefit scripts that
+build large hash tables, especially those that do so in forked sub-
+processes. The unshared mode is triggered when the hash table is resized
+which is non-deterministic and depends on the underlying hash function
+that perl has been built with and the seed used by the current process
+so the exact circumstances when this new mode will be enabled for a
+given hash will vary between different invocations of a script.
+
+We are uncertain of the exact effects of this new mode and are keen to
+get field reports about the consequences of enabling it. We know that in
+many cases not sharing keys speeds up scripts in general, and may result
+in a reduction in memory consumption. Code that builds large hashes
+where the keys are stored in only one hash, for instance where the keys
+are message digests, will likely benefit from this change by being
+faster and using less memory. However there are also circumstances where
+not sharing keys can significantly increase memory consumption, such as
+where large numbers of keys are stored in multiple hash tables at once.
+We have data to suggest that not sharing keys will reduce COW churn and
+overall memory consumption in scripts that fork and construct large hash
+tables, however such scripts may also encounter a modest speed penalty.
+Overall the consequences of not sharing keys will vary depending on
+workload, however we believe that broadly speaking enabling this feature
+will be beneficial for most of our users. We would like to get data from
+the field to validate our assumptions.
+
+We are not enabling the new mode by default in this release as we want
+to gather more data and obtain feedback from the field about how it
+affects people's scripts, but our intention is to ship the next release
+of Perl, version 5.38, with some form of this functionality enabled. We
+are hopeful that people will try to build their perls with this feature
+and provide us feedback on the results. This will help guide our next
+steps. So please try it and send feedback to the porters mailing list.
+
+You can enable this new feature by passing
+
+    -Accflags='-DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES'
+
+to F<Configure> during the Configuration process.
+
+Big thanks to Nicholas Clark for all the hard work in implementing this
+new feature.
+
 =head1 Security
 
 XXX Any security-related notices go here.  In particular, any security
--- a/hv.c
+++ b/hv.c
@@ -1576,7 +1576,12 @@ S_hv_delete_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
 
    This is a heuristic. There might be a better answer than 42, but for now
    we'll use it.
+
+   NOTE: Configure with -Accflags='-DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES'
+   to enable this new functionality.
 */
+
+#ifdef PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
 static bool
 S_large_hash_heuristic(pTHX_ HV *hv, STRLEN size) {
     if (size > 42
@@ -1588,6 +1593,7 @@ S_large_hash_heuristic(pTHX_ HV *hv, STRLEN size) {
     }
     return FALSE;
 }
+#endif
 
 STATIC void
 S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
@@ -1624,7 +1630,7 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
         return;
 
     /* don't share keys in large simple hashes */
-    if (S_large_hash_heuristic(aTHX_ hv, HvTOTALKEYS(hv)))
+    if (LARGE_HASH_HEURISTIC(hv, HvTOTALKEYS(hv)))
         HvSHAREKEYS_off(hv);
 
 
@@ -1720,7 +1726,7 @@ Perl_hv_ksplit(pTHX_ HV *hv, IV newmax)
         }
 #endif
     } else {
-        if (S_large_hash_heuristic(aTHX_ hv, newmax))
+        if (LARGE_HASH_HEURISTIC(hv, newmax))
             HvSHAREKEYS_off(hv);
         Newxz(a, PERL_HV_ARRAY_ALLOC_BYTES(newsize), char);
         xhv->xhv_max = newsize - 1;
--- a/hv.h
+++ b/hv.h
@@ -28,6 +28,13 @@
 #   define PERL_HASH_ITER_BUCKET(iter)      (((iter)->xhv_riter) ^ ((iter)->xhv_rand))
 #endif
 
+#ifdef PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
+#define LARGE_HASH_HEURISTIC(hv,new_max) S_large_hash_heuristic(aTHX_ (hv), (new_max))
+#else
+#define LARGE_HASH_HEURISTIC(hv,new_max) 0
+#endif
+
+
 /* entry in hash value chain */
 struct he {
     /* Keep hent_next first in this structure, because sv_free_arenas take
--- a/perl.c
+++ b/perl.c
@@ -1958,6 +1958,9 @@ S_Internals_V(pTHX_ CV *cv)
 #  ifdef PERL_USE_SAFE_PUTENV
                              " PERL_USE_SAFE_PUTENV"
 #  endif
+#  ifdef PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
+                             " PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES"
+#  endif
 #  ifdef SILENT_NO_TAINT_SUPPORT
                              " SILENT_NO_TAINT_SUPPORT"
 #  endif


hvds:
LGTM

hvds:
There will be knock-on effects on perldelta too, of course.

demerphq:
Added a patch to silence build warnings when PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES is not defined, and a patch to add a perldelta entry for it.

@hvds I'd like to rebase this on top of the most recent build changes. But you have objected to that in the past. This PR will have test failures unrelated to its changes until it is rebased. Do you mind if I rebase?

hvds:
> @hvds I'd like to rebase this on top of the most recent build changes. But you have objected to that in the past. This PR will have test failures unrelated to its changes until it is rebased. Do you mind if I rebase?

Not at all, rebasing tends to cause problems for unresolved review comments - they get unanchored from the changes that prompted them, and sometimes github gets very confused as a result (probably more so for a long-running PR, such that the pre-rebase commits get gc'd). For me, that usually means effectively starting the review from scratch again which is much more costly (potentially hours) than simply looking at what changed in response to a particular review comment.

There are no outstanding review comments here (and it's a small enough change that it would probably be no huge deal if there were).

demerphq:
Thanks, rebased and pushed. You ok with the perldelta change?

hvds:
> You ok with the perldelta change?

Not hugely: it _reduces_ memory use for any key that is used only once, so headlining it as "faster, but more memory hungry" seems misleading, and likely to cause people who could benefit from it to ignore it.

Here's some proposed alternative wording:
```
=head2 Experimental support for faster hashing.

This release includes support to build Perl so that it no longer shares
all keys in larger hashes. This should give a speed benefit in all cases
(unless memory pressure kicks in), and should reduce memory use for keys 
that are I<not> repeated in any other hash, but increase memory use for
keys that I<are> repeated in another hash. The gains may be particularly
significant for forking applications that build large hash tables.

This new functionality is disabled by default. You can enable it by passing

    -Accflags='-DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES'

to F<Configure>. We welcome and encourage people to test this new mode
and provide feedback on their experience with it so we can make decisions
about enabling it by default in a future release, or even allowing it to
be controlled on a per-hash basis.
```

demerphq:
On Thu, 14 Apr 2022 at 03:19, Hugo van der Sanden ***@***.***>
wrote:
>
> You ok with the perldelta change?
>
> Not hugely: it reduces memory use for any key that is used only once, so
headlining it as "faster, but more memory hungry" seems misleading, and
likely to cause people who could benefit from it to ignore it.

Well I have no problem rewording, but I think your rewording is also
problematic. :-)

The problem is the behavior of this patch swings one way or another
depending on what you are doing. If you have several large hashes that
contain the same keys then yes it does use more memory for those keys. I
provided code to prove it, a benchmark which used 40% more memory WITH the
patch. It also is NOT faster in ALL cases. In fact in some cases it is
slower. I also provided a benchmark to show that too.

Consider if I have a bunch of keys, and I have them stored in K hashes,
unless I am lucky and the key is one of the ones that gets stored in the
PL_str_tab, we will have to store K copies, so it will use K times the
memory it used to.

See:
***@***.***

If it does have to store K copies then it will have to do the work to
create K HEKs and K mallocs and K copies of the data, which will be slower
than just incrementing the refcount.

So the problem is the results are very much mixed depending on the workload.

>
> Here's some proposed alternative wording:
>
> =head2 Experimental support for faster hashing.
>
> This release includes support to build Perl so that it no longer shares
> all keys in larger hashes. This should give a speed benefit in all cases

I provided benchmarks where this is not the case.

See:
***@***.***

> (unless memory pressure kicks in), and should reduce memory use for keys
> that are I<not> repeated in any other hash, but increase memory use for
> keys that I<are> repeated in another hash. The gains may be particularly
> significant for forking applications that build large hash tables.

I guess we need to be more neutral in the language we choose. I'll give it
another go and mix in some of your proposed language.

>
> This new functionality is disabled by default. You can enable it by
passing
>
>     -Accflags='-DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES'
>
> to F<Configure>. We welcome and encourage people to test this new mode
> and provide feedback on their experience with it so we can make decisions
> about enabling it by default in a future release, or even allowing it to
> be controlled on a per-hash basis.

BTW, normally I would provide links to nntp.perl.org, but I think it
doesn't like my mails for some reason, perhaps because the posts had
attachments and HTML markup for code, and they render as just empty posts.
However if you use gmail and search for the message id using the
rfc822msgi: prefix it should find them.

The relevant content from the two mails is below, I am not attaching the
scripts as it seems to upset github and nntp.perl.org, i have sent them to
you (Hugo) in a separate email.

----
./perl -Ilib ../hash_fork.pl --n=100000 --base_n=0

shows a case where the branch is faster and uses less memory.

./perl -Ilib ../hash_fork.pl --n=100000

shows a case where the branch is slower and uses about the same memory.
----
$ ./perl -Ilib ~/scratch/shared_keys.pl; perl ~/scratch/shared_keys.pl
This is perl 5.035011
Took 6.29 seconds to process 9421 files
$VAR1 = {
          'Anonymous' => 1009956,
          'KernelPageSize' => 380,
          'MMUPageSize' => 380,
          'Private_Clean' => 56,
          'Private_Dirty' => 1012736,
          'Pss' => 1012870,
          'Referenced' => 1015536,
          'Rss' => 1015536,
          'Shared_Clean' => 2744,
          'Size' => 1023204
        };
This is perl 5.034001
Took 7.93 seconds to process 9421 files
$VAR1 = {
          'Anonymous' => 722968,
          'KernelPageSize' => 380,
          'MMUPageSize' => 380,
          'Private_Clean' => 2676,
          'Private_Dirty' => 722968,
          'Pss' => 725729,
          'Referenced' => 728432,
          'Rss' => 728432,
          'Shared_Clean' => 2788,
          'Size' => 736220
        };

Since github doesnt seem to deal well with attachments I have not included
them in this version of the post.

cheers
yves


--

hvds:
> The problem is the behavior of this patch swings one way or another depending on what you are doing. If you have several large hashes that contain the same keys then yes it does use more memory for those keys. I provided code to prove it, a benchmark which used 40% more memory WITH the patch. It also is NOT faster in ALL cases. In fact in some cases it is slower. I also provided a benchmark to show that too.

Yes, I had seen your benchmark showing a case that uses more memory: I tried to make that possibility, and the reason for it, clear in my wording.

I'd missed that you also had a benchmark showing it can be slower, I'll try to look at that more closely tomorrow.

hvds:
> $ ./perl -Ilib ~/scratch/shared_keys.pl; perl ~/scratch/shared_keys.pl
> This is perl 5.035011
> Took 6.29 seconds to process 9421 files
...
> This is perl 5.034001
> Took 7.93 seconds to process 9421 files

@demerphq Could you clarify what the version numbers represent here? They don't look like two perls differing only in whether the large_hash_heuristic is enabled.

> ./perl -Ilib ../hash_fork.pl --n=100000
> shows a case where the branch is slower and uses about the same memory.

Do you have figures for that handy? My machine is not great for getting consistent timings.

rjbs:
It would, I think, be good to get this merged before .10, right?

hvds:
> It would, I think, be good to get this merged before .10, right?

@rjbs Before .11, right? :) The primary content, yes (assuming you're happy for this to replace your #19597). The wording for perldelta doesn't need to be agreed for that, and could be moved to a separate PR if necessary.

rjbs:
@hvds My PR was the simplest thing that could address the problem, and I am perfectly happy for it to be replaced by some other solution.

demerphq:
Ill update the perldelta today and hopefully we can apply it together. Id prefer to get it in as one PR.

demerphq:
@hvds you asked for better timings, here you go. Summary: the benchmark takes 6.74 seconds with shared keys, and 8.13 without. There was some fluctuation, as is expected with a benchmark, I saw the shared keys run as fast as 5.5 seconds, and the unshared keys as fast as 7.9. Thus for this benchmark unshared keys is about 20% slower.

With shared hash keys:
```
$ git log -1 --oneline; ./perl -Ilib ../hash_fork.pl --n=500000; ./perl -Ilib -V | grep -P "^\s+PERL_USE_UNSHARED_KEYS" || echo "using shared hash keys"; ./perl -Ilib -V | grep 'config_args'
4f07281f6f (HEAD -> yves/disable_by_default_unshared_hashes, origin/yves/disable_by_default_unshared_hashes) perldelta.pod - document PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
2870606: Size: 131700 Rss: 117836 Pss: 97486
2870609: Size: 131708 Rss: 117868 Pss: 97446
2870605: Size: 131700 Rss: 117832 Pss: 97463
2870608: Size: 131700 Rss: 117868 Pss: 97493
2870611: Size: 131704 Rss: 117868 Pss: 97473
2870604: Size: 131700 Rss: 117868 Pss: 97893
2870607: Size: 131700 Rss: 117868 Pss: 101015
2870610: Size: 131708 Rss: 117868 Pss: 102862
Running 8 procs with 500000 keys with 10 items took 6.74 seconds in total
using shared hash keys
```

With the PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES defined:
```
$ git log -1 --oneline; ./perl -Ilib ../hash_fork.pl --n=500000; ./perl -Ilib -V | grep -P "^\s+PERL_USE_UNSHARED_KEYS" || echo "using shared hash keys"; ./perl -Ilib -V | grep 'config_args'
4f07281f6f (HEAD -> yves/disable_by_default_unshared_hashes, origin/yves/disable_by_default_unshared_hashes) perldelta.pod - document PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
2870639: Size: 131352 Rss: 117748 Pss: 62891
2870637: Size: 131344 Rss: 117748 Pss: 63023
2870643: Size: 131328 Rss: 117748 Pss: 64357
2870636: Size: 131344 Rss: 117748 Pss: 64570
2870640: Size: 131344 Rss: 117748 Pss: 67889
2870641: Size: 131332 Rss: 117748 Pss: 70897
2870638: Size: 131344 Rss: 117748 Pss: 71558
2870642: Size: 131332 Rss: 117748 Pss: 83417
Running 8 procs with 500000 keys with 10 items took 8.13 seconds in total
    PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
    config_args='-Dusethreads -Doptimize=-g -d -Dusedevel -Dcc=ccache gcc -Dld=gcc -DDEBUGGING -Accflags=-DPERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES'
```

demerphq:
The updated and rebased branch includes changes to show the define in -V, and to perldelta.

demerphq:
@rjbs your thoughts on the perldelta.pod change would be appreciated.

pod/perldelta.pod
```diff
@@ -36,6 +36,62 @@ distinctions, this can be useful when serialising or encoding Perl data
...
+v5.6. The shared key behavior was added to Perl originally to support
+having many objects with the same keys without having to store each key
+many times. For objects that might be created hundreds or thousands of
+times this key sharing behavior can result in using significantly less
+memory. However key sharing comes at the cost of having to maintain an
+internal "master hash table" (PL_str_tab), and having to do two store
+operations per key stored in a hash table. It also can result in
+surprising consequences with programs that fork, causing the master hash
+table to be COWed (copy on write) into each forked processes memory
+space which can, in worse case, greatly increase the memory utilization
```
hvds: `s/processes/process's/`
`s/in worse case/one_of('in the worst case', 'worst case')/`

pod/perldelta.pod
```diff
@@ -36,6 +36,62 @@ distinctions, this can be useful when serialising or encoding Perl data
...
+consumption. We also have data to suggest that it will reduce COW churn
+in scripts that use fork and build large hash tables in child processes
+for aggregation purposes. There is also data which suggests that in some
+cases it may result in a modest speed penalty. So the consequences are
+mixed. However overall we believe that enabling it will have positive
+consequences.
+
+We are not enabling the new mode by default in this release as we want
+to gather more data and obtain feedback from the field about how it
+affects peoples scripts, but our intention is to ship the next release
```
hvds: `s/peoples/people's/`

pod/perldelta.pod
```diff
@@ -36,6 +36,62 @@ distinctions, this can be useful when serialising or encoding Perl data
...
+mode will be enabled for a given hash will vary between different
+invocations of a script.
+
+We are uncertain of the exact effects of this new mode and are keen to
+get field reports about the consequences of enabling it. We know that in
+many cases it speeds up code that builds large hash tables. We also know
+that in some extreme situations it can significantly increase memory
+consumption. We also have data to suggest that it will reduce COW churn
+in scripts that use fork and build large hash tables in child processes
+for aggregation purposes. There is also data which suggests that in some
```
hvds: Same phrase "for aggregation purposes" here: I still don't really know what you mean by it, but for this context it is fine to talk about the purpose if it is reasonably obvious what kind of mechanics it implies.

pod/perldelta.pod
```diff
@@ -36,6 +36,62 @@ distinctions, this can be useful when serialising or encoding Perl data
...
+surprising consequences with programs that fork, causing the master hash
+table to be COWed (copy on write) into each forked processes memory
+space which can, in worse case, greatly increase the memory utilization
+of such scripts.
+
+The new behavior is an attempt to have our cake and eat it too. Small
+hash tables will continue to share keys, but once the hash table is
+used to store more than a certain number of keys (currently 64) it will
+not share the new keys that are added. This should benefit scripts that
+build large hash tables for aggregation purposes, especially those that
```
hvds: I don't know what "for aggregation purposes" means, and in the context of this sentence I think it would be better to talk about mechanics - we want the feedback to tell us about the purposes that benefit or lose out.

Could we perhaps replace that phrase with "in which most keys are unique", or to spell it out "in which most keys are not repeated in any other hash"? I'd like some wording that would prompt a reader to think "oh, my 100M-key hash of seen values might benefit from that" since that's the use-case where I saw huge savings (without any forking).

hvds (commented):
Thanks @demerphq, I like the balance of the updated delta, just minor comments remain.

demerphq:
@hvds I am heading to bed now. Why don't you push a patch on top of this branch with the fixes you like. I am fine with everything you have said here.  Then I can review in the morning and get this merged and out of our hair.

demerphq:
I think I applied the changes you requested. I ended up substantially rewording one of the paragraphs, no more mention of aggregation.

hvds:
LGTM

PR #19599, opened 2022-04-08
Extra B::PADNAME helpers: GEN and IsUndef

atoomic:
This commit adds two new helpers to B::PADNAME.
- IsUndef to check if the PADNAME is PL_padname_undef
- GEN to access to the xpadn_gen slot of the pad.

### DIFF
--- a/ext/B/B.xs
+++ b/ext/B/B.xs
@@ -2252,13 +2252,14 @@ MODULE = B	PACKAGE = B::PADNAME	PREFIX = Padname
 	sv_U32p | STRUCT_OFFSET(struct padname, xpadn_low)
 #define PN_cop_seq_range_high_ix \
 	sv_U32p | STRUCT_OFFSET(struct padname, xpadn_high)
+#define PN_xpadn_gen_ix \
+	sv_I32p | STRUCT_OFFSET(struct padname, xpadn_gen)
 #define PNL_refcnt_ix \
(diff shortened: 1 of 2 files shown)


toddr (approved):
Without this helper, it's impossible to round trip this information.

khwilliamson (commented):
This looks t me, but I don't have any expertise in this area

khwilliamson:
looks good to me that is; I need a new keyboard

rjbs:
@leonerd Can this go in before v5.35.11?  I don't see why not?

ext/B/B.xs
```diff
@@ -2305,6 +2307,14 @@ PadnamePV(pn)
 	SvUTF8_on(TARG);
 	XPUSHTARG;
 
+bool
+PadnameIsUndef(padn)
+       B::PADNAME      padn
+    CODE:
+        RETVAL = padn == &PL_padname_undef ? TRUE : FALSE;
```
leonerd: Why the ternary? This can simply be

```
RETVAL = padn == &PL_padname_undef;
```

atoomic: true we can simplify this, this was more a style to make it clear but I'm fine adjusting it to the short version

atoomic: changed to use the short version as part of b98236f7

leonerd (approved):
LGTM

leonerd:

PR #19600, opened 2022-04-08
Add B helpers to check if the Sv is a boolean

atoomic:
Bool are using PVNV. It makes it more convenient
to provide these helpers at the top level for any SVs.

So we can easily check if the SV is a boolean and check
if it's true or false.

Note: alternatively we could consider a single helper to return the boolean value:
- 1 if bool & Yes
- 0 if bool & No
- undef if not bool

### DIFF
--- a/ext/B/B.pm
+++ b/ext/B/B.pm
@@ -20,7 +20,7 @@ sub import {
 # walkoptree comes from B.xs
 
 BEGIN {
-    $B::VERSION = '1.83';
+    $B::VERSION = '1.84';
     @B::EXPORT_OK = ();
 
     # Our BOOT code needs $VERSION set, and will append to @EXPORT_OK.
@@ -593,6 +593,26 @@ C<REFCNT> (corresponding to the C function C<SvREFCNT>).
 
 =item FLAGS
 
+=item IsBOOL
+
+Returns true if the SV is a boolean (true or false).
+You can then use C<TRUE> to check if the value is true or false.
+
+    my $something = ( 1 == 1 ) # boolean true
+                 || ( 1 == 0 ) # boolean false
+                 || 42         # IV true
+                 || 0;         # IV false
+    my $sv = B::svref_2object(\$something);
+
+    say q[Not a boolean value]
+        if ! $sv->IsBOOL;
+
+    say q[This is a boolean with value: true]
+        if   $sv->IsBOOL && $sv->TRUE_nomg;
+
+    say q[This is a boolean with value: false]
+        if   $sv->IsBOOL && ! $sv->TRUE_nomg;
+
 =item object_2svref
 
 Returns a reference to the regular scalar corresponding to this
@@ -600,6 +620,24 @@ B::SV object.  In other words, this method is the inverse operation
 to the svref_2object() subroutine.  This scalar and other data it points
 at should be considered read-only: modifying them is neither safe nor
 guaranteed to have a sensible effect.
+
+=item TRUE
+
+Returns a boolean indicating hether Perl would evaluate the SV as true or
+false.
+
+B<Warning> this call performs 'get' magic. If you only want to check the
+nature of this SV use C<TRUE_nomg> helper.
+
+This is an alias for C<SvTRUE($sv)>.
+
+=item TRUE_nomg
+
+Check if the value is true (do not perform 'get' magic).
+Returns a boolean indicating whether Perl would evaluate the SV as true or
+false.
+
+This is an alias for C<SvTRUE_nomg($sv)>.
 
 =back
 
--- a/ext/B/B.xs
+++ b/ext/B/B.xs
@@ -1398,12 +1398,12 @@ aux_list(o, cv)
 
 
 
-MODULE = B	PACKAGE = B::SV
+MODULE = B	PACKAGE = B::SV         PREFIX = Sv
 
 #define MAGICAL_FLAG_BITS (SVs_GMG|SVs_SMG|SVs_RMG)
 
 U32
-REFCNT(sv)
+SvREFCNT(sv)
 	B::SV	sv
     ALIAS:
 	FLAGS = 0xFFFFFFFF
@@ -1417,11 +1417,23 @@ REFCNT(sv)
 	RETVAL
 
 void
-object_2svref(sv)
+Svobject_2svref(sv)
 	B::SV	sv
     PPCODE:
 	ST(0) = sv_2mortal(newRV(sv));
 	XSRETURN(1);
+
+bool
+SvIsBOOL(sv)
+    B::SV   sv
+
+bool
+SvTRUE(sv)
+    B::SV   sv
+
+bool
+SvTRUE_nomg(sv)
+    B::SV   sv
 	
 MODULE = B	PACKAGE = B::IV		PREFIX = Sv
 
--- a/MANIFEST
+++ b/MANIFEST
@@ -4247,6 +4247,7 @@ ext/B/Makefile.PL	Compiler backend makefile writer
 ext/B/O.pm		Compiler front-end module (-MO=...)
 ext/B/t/b.t		See if B works
 ext/B/t/B/success.pm	Test module for ext/B/t/o.t
+ext/B/t/bool.t		See if B works for bool
 ext/B/t/concise.t	See whether B::Concise works
 ext/B/t/concise-xs.t	See whether B::Concise recognizes XS functions
 ext/B/t/f_map			code from perldoc -f map
--- a/ext/B/t/bool.t
+++ b/ext/B/t/bool.t
@@ -1,0 +1,59 @@
+#!./perl
+
+BEGIN {
+    unshift @INC, 't';
+    require Config;
+    if (($Config::Config{'extensions'} !~ /\bB\b/) ){
+        print "1..0 # Skip -- Perl configured without B module\n";
+        exit 0;
+    }
+}
+
+use strict;
+use warnings;
+
+use B;
+use Test::More;
+
+$|  = 1;
+
+{
+    note "testing true";
+    my $bool = ( 1 == 1 );
+    my $sv = B::svref_2object(\$bool);
+    ok $sv->IsBOOL, "got a boolean";
+    ok $sv->TRUE_nomg, "TRUE_nomg is true";
+    ok $sv->TRUE, "TRUE is true";
+}
+
+{
+    note "testing false";
+    my $bool = ( 1 == 0 );
+    my $sv = B::svref_2object(\$bool);
+
+    ok $sv->IsBOOL, "got a boolean";
+    ok !$sv->TRUE_nomg, "TRUE_nomg is false";
+    ok !$sv->TRUE, "TRUE is false";
+}
+
+{
+    note "not a boolean";
+    my $iv = 42;
+    my $sv = B::svref_2object(\$iv);
+
+    ok !$sv->IsBOOL, "not a boolean";
+    ok $sv->TRUE_nomg, "TRUE_nomg is true";
+    ok $sv->TRUE, "TRUE is true";
+}
+
+{
+    note "not a boolean";
+    my $iv = 0;
+    my $sv = B::svref_2object(\$iv);
+
+    ok !$sv->IsBOOL, "not a boolean";
+    ok !$sv->TRUE_nomg, "TRUE_nomg is false";
+    ok !$sv->TRUE, "TRUE is false";
+}
+
+done_testing();


toddr (approved):

leonerd:
Not sure I get this. The `builtin::` namespace already provides an `is_bool` function that can be called from pure perl. There's no need to add it this to `B`. Or have I somehow missed the point here?

atoomic:
B serves as an authoritative interface for the internal and needs to be up to date to parse these new additions.
It should be in blead before releasing 5.36.0

toddr:
> Not sure I get this. The `builtin::` namespace already provides an `is_bool` function that can be called from pure perl. There's no need to add it this to `B`. Or have I somehow missed the point here?

The builtin functions against an SV. It does not help though with an arbitrary B::SV when walking a tree.

demerphq:
On Tue, 12 Apr 2022, 04:22 ℕicolas ℝ., ***@***.***> wrote:

> B serves as an authoritative interface for the internal and needs to be up
> to date to parse these new additions.
> It should be in blead before releasing 5.36.
>

+1

Yves

Leont:
The camel-case seems inconsistent with the rest of the API

toddr:
> The camel-case seems inconsistent with the rest of the API

If there is a consistent naming scheme, I'd say we go with something like `is_offset` ?

atoomic:
Thanks for your feedbacks, I applied a few suggestions:
* use SvTRUE
* rename functions to `is_bool`, `is_bool_yes` and `is_bool_no`

rjbs:
@leonerd Can this go in before v5.35.11?  I don't see why not?

ext/B/B.pm
```diff
@@ -593,6 +593,18 @@ C<REFCNT> (corresponding to the C function C<SvREFCNT>).
 
 =item FLAGS
 
+=item is_bool
+
+Returns true if the SV is a boolean (PL_Yes or PL_No).
+
+=item is_bool_yes
+
+Returns true if the SV is a boolean and is PL_Yes.
```
leonerd: "... is a boolean and true". Most regular SVs that are boolean and true will not be exactly `PL_sv_yes`

ext/B/B.xs
```diff
@@ -1422,6 +1422,30 @@ object_2svref(sv)
     PPCODE:
 	ST(0) = sv_2mortal(newRV(sv));
 	XSRETURN(1);
+
+bool
+is_bool(sv)
+    B::SV   sv
+    CODE:
+    RETVAL = SvIsBOOL(sv) ? TRUE : FALSE;
```
leonerd: Why the ternary? This can simply be

```
RETVAL = SvIsBOOL(sv);
```

Ditto the other two functions below.

ext/B/B.pm
```diff
@@ -593,6 +593,18 @@ C<REFCNT> (corresponding to the C function C<SvREFCNT>).
...
+
+=item is_bool_no
+
+Returns true if the SV is a boolean and is PL_No.
+
```
tonycoz: The code doesn't check whether the SV is PL_Yes or PL_No, it checks whether the SV is a boolean (which any SV can be) and for the second two, whether that value is true or false.

ext/B/B.xs
leonerd: Actually for that matter, why does it have its own name? Why can it not simply be the SvIsBOOL itself?

```
bool
SvIsBOOL(sv)
    B::SV  sv
```

with no code body at all. E.g. compare to the `SvIV` below.

ext/B/B.xs
```diff
@@ -1422,6 +1422,30 @@ object_2svref(sv)
...
+    OUTPUT:
+        RETVAL
+
+bool
+is_bool_yes(sv)
```
leonerd: I'm not sure why this is useful (also the `no` below it). If anyone cares to check they can test it from perl-land just as easily

```perl
  if($b_sv->IsBOOL && $b_sv->TRUE) { ... }
```

toddr: When you are serializing the information, you need a way to query the B object so it round trips properly in another process.

atoomic: I think this provides a better abstraction, where you could change internal without changing all consumers
if this would change.

ext/B/B.xs
atoomic: that could work

ext/B/B.pm
atoomic: going to update it as suggested

ext/B/B.pm
atoomic: going to adjust the description to match the suggestion for is_bool_yes

ext/B/B.xs
leonerd: Yes. But why do you need to add a new XS method on the B::SV class (`is_bool_yes`) which just returns information that is the `and` combination of two other methods that already exist (`IsBOOL and TRUE`)?

leonerd: > I think this provides a better abstraction, where you could change internal without changing all consumers
if this would change.

This isn't about internals. IsBOOL returns true for any boolean. TRUE returns true for any value that perl considers true. Thus the combination `$sv->IsBOOL and $sv->TRUE` is simply the combination of those two facts. It doesn't need a special method.

This would be like adding a special method to ask if a given SV is numerical and the value 1. Why bother? We can just check those two facts, separately, using the regular query methods. No need to special-case it.

atoomic: This is assuming that this will always be the way to check if the boolean is true or false.

I was under the assumption that this could change overtime and we could start using one shared SV for true or false at one point or whatever other optimization we decide in the future.

If we think that `IsBOOL and TRUE` is always going to work, then indeed we can consider removing the `is_bool_yes` and `is_bool_no` helpers. I was trying to provide abstraction for the user which should not track how this is implemented behind.

This would add more boilerplate to code parsing a B tree and trying to check booleans. But I'm fine with that.

ext/B/B.xs
```diff
@@ -1422,6 +1422,30 @@ object_2svref(sv)
...
+    B::SV   sv
+    CODE:
+    RETVAL = SvIsBOOL(sv) && SvTRUE(sv) ? TRUE : FALSE;
```
tonycoz: Aside from @leonerd 's comment (which I agree with), this will invoke magic if any on the SV if it started out as a bool.

Other B accessors correctly don't do that, since the intent is to fetch the internal state of the SV.

This (or a replacement TRUE method if the combos are removed) should use SvTRUE_nomg() rather than SvTRUE().

ext/B/B.xs
atoomic: note that this should not use `SvIsBool` as there is no prefix for the `B::SV` package set
so I gave a try of using 
```
bool
IsBOOL(sv)
    B::SV   sv
```

this is now failing with
```
./perl: symbol lookup error: ../lib/auto/B/B.so: undefined symbol: IsBOOL
```

note sure how to fix it. If you have any ideas or suggestion I'm glad to take it
thanks

ext/B/B.xs
leonerd: I was imagining that there's already a `->TRUE` method which is a wrapper of `SvTRUE()` (though, as @tonycoz points out it really should be using `SvTRUE_mg()`. If not, then I'd much prefer adding that one as it's more universally useful - it would apply to any kind of SV then.

But yes, have no fear of `SvTRUE()` and friends ever changing - they are as much a core part of the basic SV API as something like `SvIV` is.

ext/B/B.xs
atoomic: ok need to setup the prefix... but then this become a mess
```diff
diff --git a/ext/B/B.xs b/ext/B/B.xs
index 878fb0dba5..4f75bec4f0 100644
--- a/ext/B/B.xs
+++ b/ext/B/B.xs
@@ -1398,12 +1398,12 @@ aux_list(o, cv)



-MODULE = B     PACKAGE = B::SV
+MODULE = B     PACKAGE = B::SV         PREFIX = Sv

 #define MAGICAL_FLAG_BITS (SVs_GMG|SVs_SMG|SVs_RMG)

 U32
-REFCNT(sv)
+SvREFCNT(sv)
        B::SV   sv
     ALIAS:
        FLAGS = 0xFFFFFFFF
@@ -1417,7 +1417,7 @@ REFCNT(sv)
        RETVAL

 void
-object_2svref(sv)
+Svobject_2svref(sv)
        B::SV   sv
     PPCODE:
        ST(0) = sv_2mortal(newRV(sv));
@@ -1428,7 +1428,7 @@ SvIsBOOL(sv)
     B::SV   sv

 bool
-is_bool_yes(sv)
+Svis_bool_yes(sv)
     B::SV   sv
     CODE:
     RETVAL = SvIsBOOL(sv) && SvTRUE(sv);
@@ -1436,7 +1436,7 @@ is_bool_yes(sv)
         RETVAL

 bool
-is_bool_no(sv)
+Svis_bool_no(sv)
     B::SV   sv
     CODE:
     RETVAL = SvIsBOOL(sv) && !SvTRUE(sv);
```

atoomic: Would you really prefer this flavor?

ext/B/B.xs
atoomic: yeap I can use SvTRUE_nomg... 
my first implementation was using a PVX check as the `Perl_sv_isbool` does.
But as suggested I changed it to `SvTRUE` and you are right `SvTRUE_nomg` is a better choice
I wonder if the `PVX` check was not the correct way to do it
```
 PERL_STATIC_INLINE bool
 Perl_sv_isbool(pTHX_ const SV *sv)
 {
     return SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) &&
         (SvPVX_const(sv) == PL_Yes || SvPVX_const(sv) == PL_No);
 }
```

ext/B/B.xs
Leont: Using PREFIX generally only makes sense if you don't have a CODE body

ext/B/B.xs
leonerd: The `SvPVX_const` test is how the implementation of SvIsBOOL itself is implemented (along with the IsCOW_static check). That function is the one that implements the macro that everyone *else* should be using.

atoomic:
Hi here is another iteration via 24204f27c73c1682d8a51e549369f96a3e621c5e trying to use most suggestions from this PR:
Note that it has the following issues:
- add the Sv prefix to B::SV to use a no body helper for IsBOOL
- IsBOOL is now back to camel case after requesting a no body implementation
- back to original PVX check rather than SvTRUE to avoid triggering magic

Question:
if we decide to remove the `is_bool_yes` and `is_bool_no` helpers, would not this trigger some magic
while trying to call `$sv->TRUE` on it?

leonerd:
> if we decide to remove the is_bool_yes and is_bool_no helpers, would not this trigger some magic
while trying to call $sv->TRUE on it?

That's why you'd have `$sv->TRUE` call the SvTRUE_mg macro:

```
bool
SvTRUE(sv)
    B::SV sv
    CODE:
      RETVAL = SvTRUE_nomg(sv);
    OUTPUT:
      RETVAL
```

Having done that, any caller who cares can check `$sv->IsBOOL and $sv->TRUE` so you no longer need your (still incorrectly implemented) is_bool_yes / no.

toddr:
> so you no longer need your (still incorrectly implemented) is_bool_yes / no.

Point of clarification. Are you suggesting that this existing code is incorrectly implemented?

```C
PERL_STATIC_INLINE bool
Perl_sv_isbool(pTHX_ const SV *sv)
{
    return SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) &&
        (SvPVX_const(sv) == PL_Yes || SvPVX_const(sv) == PL_No);
}
```

leonerd:
No. Please read again my comment:

> The SvPVX_const test is how the implementation of SvIsBOOL itself is implemented (along with the IsCOW_static check). That function is the one that implements the macro that everyone else should be using.

The code you quoted is the *implementation* of SvIsBOOL. It is permitted to use its own internal trickery. Everyone else should call that macro instead, so as to be abstracted away from its implementation.

leonerd:
To clarify further: `SvIsBOOL`'s current implementation relies on the internals of how real-boolean SVs work - they are COW-static SVs, whose PV pointer is exactly a pointer to the `PL_Yes` or `PL_No` static string buffers. If we have a COW-static SV whose PV pointer is one of these, then we know we have a kind of boolean.

Any other user code, knowing *that* we have a real-boolean (by calling the above macro) does not need to know anything about these internals. The only thing that distinguishes it, beyond this, is whether the boolean is true or false. For that the user can call the plain ol' `SvTRUE()` or `SvTRUE_nomg()` macros to test for SV truth exactly the same way as they've always done for the past 20 years.

I will say once more in case it is still unclear: There is **absolutely no need** for any code *except the implementation of `SvIsBOOL`* to be making any sort of pointer comparison against the `PL_Yes` or `PL_No` static arrays.

ext/B/B.xs
haarg: I'm not particularly a fan of `is_bool_yes` or `is_bool_no`, but more importantly:

B is not the place for abstractions. If an abstract understanding was appropriate, it would be available through normal perl ops. When the internals change, B should reflect that. If you were relying on those internals, your code should break.

leonerd (changes requested):
The `IsBOOL` part looks fine now.

The `is_bool_yes` would be better done by making a new `TRUE` in similar style to `IsBOOL` and letting the user test `$sv->IsBOOL and $sv->TRUE` rather than having this special-case method just for that test. Similar thoughts for `is_bool_no`.

As commented; make sure to call `SvTRUE_nomg()` to avoid the magic effects of the regular `SvTRUE` macro.

atoomic:
I think following @toddr advice a better implementation would be to use a single helper `check_bool` or whatever name
that could return three different values:
- undef: when it's not a boolean 
- PL_sv_yes: when it s a boolean and true
- PL_sv_no: when it s a boolean and false 

So then this would be easy to check, and avoid multiple calls and helpers
```perl
my $bool_state = $sv->check_bool;
if ( defined $bool_state ) {
   say $bool_state == 1 ? 'true' : 'false';
} else {
... not a boolean
}
```

tonycoz:
> I think following @toddr advice a better implementation would be to use a single helper `check_bool` or whatever name that could return three different values:
> 
>     * undef: when it's not a boolean
> 
>     * PL_sv_yes: when it s a boolean and true
> 
>     * PL_sv_no: when it s a boolean and false

I think you're trying to over-optimize to the point of providing a bad UI.

It's a UI perl itself uses for mutators, eg read, readline, but in those examples the underlying state changes due to the call itself, which isn't the case here.

Ideally I think the methods should do one thing and not try to overload their return values in this way.

atoomic:
I gave it a try via ad594907a94cbdd20adf6c4d0448f1caed2af174 and IMO this looks much cleaner and less noise added to the B::SV package. One single helper instead of three...
Definitively open to any opinion/feedbacks...

iabyn:
I dislike this intensely. It is mixing two very different concepts - to  retrieve the type of a variable and retrieve the value of a variable.

leonerd:
I agree with @iabyn and @tonycoz here.

The B interface is supposed to be about as low-level and lightweight as possible; providing a perl interface layer onto basically the same "SV field accessor" macros that the C/XS layer already uses. C/XS already uses function-like macros called SvIsBOOL and SvTRUE to answer the questions you are trying to ask of it. Therefore the B interface layer really should provide the same. (That it previously didn't provide a `->TRUE` method already is perhaps an oversight of the current API).

The helper wrapper you provide *could* be useful in your use-case, but if you need it for your own use it would be easy to make as a one-line injected method in pureperl on top of the `->IsBOOL` and `->TRUE` accessors I've already suggested and given implementations for:

```perl
sub B::SV::check_bool ($sv) { return $sv->IsBOOL ? $sv->TRUE : undef }
```

ext/B/B.xs
toddr: > Aside from @leonerd 's comment (which I agree with), this will invoke magic if any on the SV if it started out as a bool.

@tonycoz: Is it possible to have a bool SV with magic on it?

atoomic:
Thanks for your feedback, I've adjusted the PR by providing the `IsBOOL`, `TRUE` and `TRUE_nomg`.
TBH it's not clear to me once we have established that `IsBOOL` is true, if it's wrong to use `TRUE` instead of `TRUE_nomg`. Would not a boolean never contain magic? What would be the cases?

Waiting for another review.

leonerd:
Any kind of SV can have magic on it. Don't forget that variables are SVs, and many kinds of magic apply to the variable, not the value. For example, a `tie`d lexical.

ext/B/B.pm
```diff
@@ -601,6 +621,24 @@ to the svref_2object() subroutine.  This scalar and other data it points
 at should be considered read-only: modifying them is neither safe nor
 guaranteed to have a sensible effect.
 
+=item TRUE
+
+Returns a boolean indicating hether Perl would evaluate the SV as true or
```
leonerd, line 626: Typo "whether"

leonerd (approved):
That's looking much better now. All the additions are simply wrappers for the same-named underlying C macros, as is the usual B style.

Approved, subject to one docs typo fix.

ext/B/B.pm
atoomic: thanks, fixed by e2f9e9f9

ext/B/B.pm
```diff
@@ -593,6 +593,26 @@ C<REFCNT> (corresponding to the C function C<SvREFCNT>).
 
 =item FLAGS
 
+=item IsBOOL
+
+Returns true if the SV is a boolean (true or false).
+You can then use C<TRUE> to check if it's true or false
```
toddr: s/it's/it is/ ?

toddr (approved):

ext/B/B.pm
atoomic: fixed as part of c62e2787

leonerd:

PR #19602, opened 2022-04-09
gdbm: make sure basic error codes are always defined

graygnuorg:
* ext/GDBM_File/Makefile.PL: Don't protect basic error codes by
ifdefs: they are not macros in recent versions of GDBM. Besides,
they are known to exist in any GDBM version.

### DIFF
--- a/ext/GDBM_File/Makefile.PL
+++ b/ext/GDBM_File/Makefile.PL
@@ -25,7 +25,12 @@ push @names, {
              GDBM_SNAPSHOT_SAME GDBM_SNAPSHOT_SUSPICIOUS);
 
 # Basic error codes - these are supported by all versions of gdbm
-push @names, qw(
+push @names, {
+    name  => $_,
+    value => "$_",
+    type  => "IV",
+    macro => 1
+} foreach qw(
     GDBM_NO_ERROR
     GDBM_MALLOC_ERROR
     GDBM_BLOCK_SIZE_ERROR


demerphq:
LGTM, but will wait a day or two before I merge just to see if anyone has any additional feedback.  Does this need a version bump as well?

jkeenan:
> LGTM, but will wait a day or two before I merge just to see if anyone has any additional feedback. 

Over the past couple of years we've had a number of problems with GDBM_File in the way it interacts with the version of `gdbm` that it finds on various platforms.  @graygnuorg has usually been the person to solve this problem for us, so I have a high degree of confidence in this p.r.

However, I think it would benefit from smoke-testing so that we see what happens on platforms not represented in our github CI.  To that end, I have created the `smoke-me/jkeenan/gray/gdbm-macro` branch and encourage our smoke-testers to give it a spin.

> Does this need a version bump as well?

The fact that it's passing the `make test_porting` tests in CI suggests that no version bump is needed.  The change is only in the `Makefile.PL`.

demerphq:
On Sat, 9 Apr 2022 at 15:29, James E Keenan ***@***.***>
wrote:

> LGTM, but will wait a day or two before I merge just to see if anyone has
> any additional feedback.
>
> Over the past couple of years we've had a number of problems with
> GDBM_File in the way it interacts with the version of gdbm that it finds
> on various platforms. @graygnuorg <https://github.com/graygnuorg> has
> usually been the person to solve this problem for us, so I have a high
> degree of confidence in this p.r.
>
Fair enough!


> However, I think it would benefit from smoke-testing so that we see what
> happens on platforms not represented in our github CI. To that end, I have
> created the smoke-me/jkeenan/gray/gdbm-macro branch and encourage our
> smoke-testers to give it a spin.
>
Good idea.


> Does this need a version bump as well?
>
> The fact that it's passing the make test_porting tests in CI suggests
> that no version bump is needed. The change is only in the Makefile.PL.
>
Interesting, I guess that makes sense since it is in ext/, unlike cpan/ or
dist/ where a change to Makefile.PL would still need a version bump.

Yves

-- 

jkeenan:
[Smoke test results](http://perl.develop-help.com/?b=smoke-me%2Fjkeenan%2Fgray%2Fgdbm-macro) were satisfactory.  Some smokers had failures not related to this change and previously observed on those machines.

Rebased on blead, manually merged and pushed to origin in 
0f875b7c4180f248f0fec87711de88c22cbc2009.

Closing this ticket.

Thanks again @graygnuorg

PR #19604, opened 2022-04-09
EnhanceExtUtils::Constant to work for z/OS

khwilliamson:


jkeenan:
ExtUtils-Constant falls under `cpan/` in the core distribution -- meaning that it is primarily maintained on CPAN.  So shouldn't this pull request go first to https://rt.cpan.org/Dist/Display.html?Name=ExtUtils-Constant ?

khwilliamson:
That we already have a customized version of this module for 1 2/3 years tells me it isn't likely to be re-released any time soon.  I merely added new customizations, and submitted patches to the RT queue of the distribution, as the instructions for customizations say to do.

khwilliamson:
And, if I were the maintainer, I wouldn't see the need to do a new release for this issue

Issue #16478, opened 2018-03-26
Regexes which interpolate list elements incorrectly parsed


p5pRT:
Migrated from [rt.perl.org#133027](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133027) (status was 'open')

Searchable as RT133027$

p5pRT:
### From david@cantrell.org.uk

#### Created by david@cantrell.org.uk



If you interpolate a list element \- ie $x\[\.\.\.\] into a regex\,
It appears to not always be parsed correctly\. I would expect
this code to print "bar" four times\. Instead it prints&#8203;:

bar
bar
foo
bar

$x\[1\]    = "foo"; $\_ = "foo"; s/$x\[1\]/bar/;    print "$\_\\n";
$x\[10\]   = "foo"; $\_ = "foo"; s/$x\[10\]/bar/;   print "$\_\\n";
$x\[100\]  = "foo"; $\_ = "foo"; s/$x\[100\]/bar/;  print "$\_\\n";
$x\[1000\] = "foo"; $\_ = "foo"; s/$x\[1000\]/bar/; print "$\_\\n";

NB I didn't discover this myself\. User 'ernix' on stackoverflow
did\, but I couldn't see anything for it in rt\.perl\.org\. Here's his
original message&#8203;:

https://stackoverflow.com/questions/49492181/regex-substitution-using-a-list-element-with-just-3-digits-index-doesnt-work-as

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=medium

Site configuration information for perl 5.27.11:

Configured by dc at Mon Mar 26 16:20:20 BST 2018.

Summary of my perl5 (revision 5 version 27 subversion 11) configuration:
  Commit id: 811612a11efb1ebc131370e8238d3512779354f8
  Platform:
    osname=linux
    osvers=4.9.0-6-amd64
    archname=x86_64-linux-thread-multi
    uname='linux chimera-dev-vm 4.9.0-6-amd64 #1 smp debian 4.9.82-1+deb9u3 (2018-03-02) x86_64 gnulinux '
    config_args='-de -Duse64bitall -Dusethreads -Dprefix=/home/dc/perl-blead -Dusedevel'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
[...]

PR #19614, opened 2022-04-10
toke.c: S_intuit_more: Cleanup

khwilliamson:
"This is terrifying code"

At $work there was a story about hardware that collectively handled perhaps billions of phone calls a day.  The interrupt handler was long  spaghetti code which had a single comment /* Subtle */.  Everyone was afraid to touch it, instead creating workarounds for its bugs.  Eventually Management decided to spend the money to rewrite it.

I think this is similar to a portion of S_intuit_more in toke.c, as expressed in #16478.  I have spent some hours going through the code, commenting it and adding white space to make it more legible.  In the process, I discovered some wasted effort, and some missing things.  This PR should make it less terrifying for people to work on

### DIFF
--- a/toke.c
+++ b/toke.c
@@ -4464,7 +4464,7 @@ S_scan_const(pTHX_ char *start)
  *   {4,5} (any digits around the comma) returns FALSE
  * if we're in a pattern and the first char is a [
  *   [] returns FALSE
- *   [SOMETHING] has a funky algorithm to decide whether it's a
+ *   [SOMETHING] has a funky heuristic to decide whether it's a
  *      character class or not.  It has to deal with things like
  *      /$foo[-3]/ and /$foo[$bar]/ as well as /$foo[$\d]+/
  * anything else returns TRUE
@@ -4477,22 +4477,37 @@ S_intuit_more(pTHX_ char *s, char *e)
 {
     PERL_ARGS_ASSERT_INTUIT_MORE;
 
+    /* If recursed within brackets, there is more to the expression */
     if (PL_lex_brackets)
         return TRUE;
-    if (*s == '-' && s[1] == '>' && (s[2] == '[' || s[2] == '{'))
+
+    if (    *s == '-'
+        &&  s[1] == '>'
+        && (s[2] == '[' || s[2] == '{'))
+    {
         return TRUE;
-    if (*s == '-' && s[1] == '>'
-     && FEATURE_POSTDEREF_QQ_IS_ENABLED
-     && ( (s[2] == '$' && (s[3] == '*' || (s[3] == '#' && s[4] == '*')))
-        ||(s[2] == '@' && memCHRs("*[{",s[3])) ))
+    }
+
+    if (   *s == '-'
+        && s[1] == '>'
+        && FEATURE_POSTDEREF_QQ_IS_ENABLED
+        && (   (s[2] == '$' && (    s[3] == '*'
+                                || (s[3] == '#' && s[4] == '*')))
+            || (s[2] == '@' && memCHRs("*[{", s[3])) ))
         return TRUE;
+
     if (*s != '{' && *s != '[')
         return FALSE;
+
+    /* quit immediately from any errors from now on */
     PL_parser->sub_no_recover = TRUE;
+
+    /* Here is '{' or '['.  Outside patterns, they're always subscripts */
     if (!PL_lex_inpat)
         return TRUE;
 
-    /* In a pattern, so maybe we have {n,m}. */
+    /* In a pattern, so maybe we have {n,m}, in which case, there isn't more to
+     * the expression. */
     if (*s == '{') {
         if (regcurly(s, e, NULL)) {
             return FALSE;
@@ -4500,14 +4515,17 @@ S_intuit_more(pTHX_ char *s, char *e)
         return TRUE;
     }
 
-    /* On the other hand, maybe we have a character class */
-
+    /* Here is '[': maybe we have a character class.  Examine the guts */
     s++;
+
+    /* '^' implies a character class; An empty '[]' isn't legal, but it does
+     * mean there isn't more to come */
     if (*s == ']' || *s == '^')
         return FALSE;
 
+    /* Find matching ']' */
     const char * const send = (char *) memchr(s, ']', e - s);
-    if (!send)		/* has to be an expression */
+    if (! send)		/* has to be an expression */
         return TRUE;
 
     /* If the construct consists entirely of one or two digits, call it a
@@ -4516,7 +4534,7 @@ S_intuit_more(pTHX_ char *s, char *e)
         return TRUE;
     }
 
-    /* this is terrifying, and it works */
+    /* this is terrifying, and it mostly works.  See GH #16478 */
 
     int weight;
 
@@ -4529,10 +4547,15 @@ S_intuit_more(pTHX_ char *s, char *e)
         weight = 2;
     }
 
+    /* Unsigned version of current character */
     unsigned char un_char = 255;
-    char seen[256];
-    Zero(seen,256,char);
 
+    /* Keep track of how many multiple occurrences of the same character there
+     * are */
+    char seen[256];
+    Zero(seen, 256, char);
+
+    /* Examine each character in the construct */
     for (; s < send; s++) {
         unsigned char last_un_char = un_char;
         un_char = (unsigned char)*s;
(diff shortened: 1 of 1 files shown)


khwilliamson:
I did not analyze why the particular weight factors were chosen.  And #16478 indicates this doesn't work as well as claimed

toke.c
```diff
@@ -4427,15 +4427,25 @@ S_intuit_more(pTHX_ char *s, char *e)
 
     if (PL_lex_brackets)
         return TRUE;
-    if (*s == '-' && s[1] == '>' && (s[2] == '[' || s[2] == '{'))
+
+    if (    *s == '-'
```
ilmari: I know the code mostly use s `*s` when only looking at one byte, but in places like this where it's looking several bytes ahead using `s[1]` etc., I think it would be clearer to use `s[0]` for the current byte.

toke.c
```diff
@@ -4470,76 +4482,112 @@ S_intuit_more(pTHX_ char *s, char *e)
...
+
+            /* If it is something like 'a-' or '0-', it is more likely to
+             * be a character class */
+            if (memCHRs("aA01! ", last_un_char))
                 weight += 30;
-            if (memCHRs("zZ79~",s[1]))
+
+            /* If it is something like '-Z' or '-7' (for octal) or '-9' it
+             * is more likely to be a character class */
+            if (memCHRs("zZ79~", s[1]))
```
ilmari: These comments really should mention the least obvious one of these: `[!-~]` (all printable ASCII characters).

toke.c
```diff
@@ -4623,6 +4623,11 @@ S_intuit_more(pTHX_ char *s, char *e)
          * one way or another. */
         if (un_char != '\\') {
             seen[un_char]++;
+
+            /* If wrapped around, just leave it at the maximum value */
+            if (seen[un_char] == 0) {
+                seen[un_char]--;
+            }
```
ilmari: IMO this would be clearer as 
```c
/* Avoid wrapping around */
if (seen[un_char] < 255)
    seen[un_char]++;
```

toke.c
```diff
@@ -4470,76 +4482,112 @@ S_intuit_more(pTHX_ char *s, char *e)
...
-
-        /* First char is dollar; lean very slightly to it being a subscript */
+    else if (*s == '$') { /* If the first char is dollar; lean very slightly to
+                             it being a subscript */
         weight = -1;
     }
-    else {
-        /* Otherwise, lean a little more towards it being a character class */
+    else { /* Otherwise, lean a little more towards it being a character class.
+            */
```
hvds: Interesting, I much prefer the layout of these comments before the change - when I'm cleaning up gnarly code I'd tend to make exactly the opposite edit (except for the useless blank line at 4474).

toke.c
```diff
@@ -4458,9 +4468,11 @@ S_intuit_more(pTHX_ char *s, char *e)
     if (!send)		/* has to be an expression */
         return TRUE;
 
-    /* this is terrifying, and it works */
+    /* this is terrifying, and it mostly works.  See GH #16478 */
 
     int weight;
+
+    /* Unsigned versions of current character, and the penultimate one */
```
hvds: s/penultimate/previous/

I'd expect "the penultimate one" to be the last-but-one character of the string.

hvds (commented):
I'm a bit concerned that this isn't just cleanup, but also makes a variety of functional changes. Can we list all those in one place (perhaps in the ticket), so that we have the chance to more carefully consider which patterns will be affected by them, and thus what sort of breakage this might lead to in the wild?

ilmari:
> I'm a bit concerned that this isn't just cleanup, but also makes a variety of functional changes. Can we list all those in one place (perhaps in the ticket), so that we have the chance to more carefully consider which patterns will be affected by them, and thus what sort of breakage this might lead to in the wild?

I share this concern, and I think the functional changes should be a separate follow-on PR from the cleanup, so that we can confirm that the cleanup didn't break anything before we merge the changes.

Also, some tests demonstrating the changed behaviour would be nice.

toke.c
khwilliamson: Thank you, I failed to spot this

toke.c
khwilliamson: Agreed

toke.c
khwilliamson: I agree.  I was trying to avoid changing the code unnecessarily.  But now I feel free to

toke.c
khwilliamson: You're right

toke.c
```diff
@@ -4470,76 +4482,113 @@ S_intuit_more(pTHX_ char *s, char *e)
...
                      && s[1]
-                     && memCHRs("[#!%*<>()-=",s[1]))
+                     && memCHRs("[#!%*<>()-=", s[1]))
             {
-                if (/*{*/ memCHRs("])} =",s[2]))
+                /* Here we have what could be a punctuation variable.  If the
+                 * next character after it is a closing bracket, it makes it
+                 * quite likely to be that, and hence a subscript.  If it is
+                 * something else, more mildly a subscript */
+                if (/*{*/ memCHRs("])} =", s[2]))
```
khwilliamson, line 4621: I don't understand why the comment `/*{*/`

toke.c
```diff
@@ -4493,7 +4493,7 @@ S_intuit_more(pTHX_ char *s, char *e)
 
     /* Keep track of how many multiple occurrences of the same character there
      * are */
-    char seen[256];
+    U8 seen[256];
```
khwilliamson: This change breaks a test in pat.t, showing that the code worked by happenstance.  That test wanted 2000 'b' characters in a row to be treated as a character class.  No one, outside poor generated code, would write the same character multiple times inside a character class: it would reduce to just the first occurrence.  And the code intended to make such things to be interpreted as subscripts, by penalizing it from being a character class with increasing slope for each added redundant character.  But because this array was declared signed, after you got to 127, the next increment would be to -126 or -127 (which I believe is undefined behavior) and that would **subtract** instead of adding additional penalty.  This would continue to -1, then more undefined behavior to get to 0, and the cycle repeats.  The 2000 was arbitrarily chosen; other repeat factors would have resulted in different results

toke.c
```diff
@@ -4446,7 +4447,10 @@ S_intuit_more(pTHX_ char *s, char *e)
     if (s[0] != '{' && s[0] != '[')
         return FALSE;
 
+    /* quit immediately from any errors from now on */
```
khwilliamson, line 4511: I don't know why this is here

toke.c
```diff
@@ -4446,7 +4447,10 @@ S_intuit_more(pTHX_ char *s, char *e)
     if (s[0] != '{' && s[0] != '[')
```
khwilliamson, line 4508: I don't understand this line.  Apparently it is implementing the prolog comment

>  * if we're outside a pattern and it's not { or [, then return FALSE

But we haven't tested yet if we're inside a pattern

toke.c
```diff
@@ -4428,30 +4428,30 @@ S_intuit_more(pTHX_ char *s, char *e)
...
+    if (s[0] != '{' && s[0] != '[')
         return FALSE;
 
     PL_parser->sub_no_recover = TRUE;
     if (!PL_lex_inpat)
         return TRUE;
 
     /* In a pattern, so maybe we have {n,m}. */
-    if (*s == '{') {
+    if (s[0] == '{') {
```
khwilliamson, line 4525: I don't understand this.  regcurly had very restricted syntax, so the authors could have imagined that anything that matched that would necessarily mean a character class.  But the syntax for regcurly has been relaxed somewhat, and the absence of it being regcurly doesn't mean it can't be a character class

khwilliamson: Hmm, github attached my comment to the wrong line.

toke.c
hvds: > I don't understand why the comment `/*{*/`

Pretty sure this is to avoid the `}` in the string messing up bracket matching in editors.

khwilliamson:
Yes the cleanup should be split from any changes

But now that I understand it better, I see how brittle it is, and that some tests pass just because the arbitrarily chosen numbers happened to work.

One thing that I see missing is context.  If this is called immediately when parsing `$a[...]` it would be helpful for this routine to know that there exists a `@a` which would lead the initial bias more towards a subscript

I'd like to know why we aren't getting more reports of failures from the field

toke.c
khwilliamson: Actually on a twos complement box, it goes from 127 to -128, which will over the course of many cycles will cause the weight to become more and more negative; exactly the opposite of what the code intends.

toke.c
khwilliamson: @hvds response is that it is for editors who bounce between matching brackets and don't know to disregard those in string literals, seems quite plausible.  Thanks

toke.c
khwilliamson: The breakage actually stemmed from a different change, but this remains, and shows the brittleness of the code

toke.c
khwilliamson: addressed in commentary commit

khwilliamson:
The original pull request is now replaced by other commits which don't change the current behavior

toke.c
khwilliamson: The new PR changes this a bit; and it just seems natural to me to do it similarly in this kind of context

toke.c
```diff
@@ -4681,6 +4679,7 @@ S_intuit_more(pTHX_ char *s, char *e)
         }   /* End of switch */
 
         seen[un_char]++;
+        first_time = false;
```
hvds: This is vulnerable to future addition of a 'continue' further up; an alternative would be to put it in the loop continuation instead:
```
-  for (; s < send; s++) {
+  for (; s < send; s++, first_time = false) {
```

toke.c
```diff
@@ -4477,11 +4477,20 @@ S_intuit_more(pTHX_ char *s, char *e)
...
+
     /* If recursed within brackets, there is more to the expression */
     if (PL_lex_brackets)
         return TRUE;
 
-    /* If begins with '->' ... */
+    /* If begins with '->' ...
+     *
+     * khw: This potentially reads s[1], s[2], and s[3] without checking that
+     * those exist. */
```
hvds: Is the string not known to be nul-terminated?

toke.c
```diff
@@ -4609,10 +4632,12 @@ S_intuit_more(pTHX_ char *s, char *e)
             if (s[1]) {
                 if (memCHRs("wds]", s[1]))
                     weight += 100;  /* \w \d \s => strongly charclass */
+                    /* khw: Why not \W \D \S \h \v, etc as well? */
```
hvds, line 4632: Pretty sure at least the negations (`\W \D \S`) were added after this code was last touched.

toke.c
```diff
@@ -4625,7 +4650,10 @@ S_intuit_more(pTHX_ char *s, char *e)
 
           case '-':
             /* If it is something like '-\', it is more likely to be a
-             * character class */
+             * character class.
+             *
+             * khw: The rest of the conditionals in this 'case' really should
+             * be subect to an 'else' of this condition */
```
hvds: "subject"

toke.c
```diff
@@ -4685,6 +4718,15 @@ S_intuit_more(pTHX_ char *s, char *e)
             break;
         }   /* End of switch */
 
+        /* khw: 'seen' is declared as a char.  This ++ can cause it to wrap.
+         * This gives different results with compilers for which a plain 'char'
+         * is actually unsigned, versus those where it is signed.  I believe it
+         * is undefined behavior to wrap a 'signed'.  I think it should be
+         * instead declared an unsigned int to make the chances of wrapping
+         * essentially zero.
```
hvds, line 4723: I think making it unsigned would be good. Having it wrap at 8 bits just means the 256'th occurrence adds 0 to the weight, which is no great loss - making them uints doesn't make any difference to the heuristic in any realistic scenario.

hvds (approved):
My preference would be to avoid trailing comments that are multi-line.

toke.c
khwilliamson: My original P.R added them, but you (rightly) objected to it  changing behavior, so blead has never checked for them.  Doing so would be in a new P.R.

toke.c
khwilliamson: Thanks

toke.c
khwilliamson: You're right; I wasn't thinking

toke.c
khwilliamson: This is just a summary of the function.

toke.c
khwilliamson: I never would have thought of this..  Changing

toke.c
khwilliamson: It makes sense to me now.

Issue #11568, opened 2011-08-08
4 pod casemapping errors: s/(\w+)/\u\L$1/g is always wrong


p5pRT:
Migrated from [rt.perl.org#96592](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=96592) (status was 'open')

Searchable as RT96592$

p5pRT:
### From tchrist@perl.com

These are all in error&#8203;:

&nbsp;    perldata\.pod&#8203;:        s/\(\\w\+\)/\\u\\L$1/g;   \# "titlecase" words
&nbsp;    perlfaq4\.pod&#8203;:    $string =~ s/\(\[\\w'\]\+\)/\\u\\L$1/g;
&nbsp;    perlop\.pod&#8203;:    substr\($str\, \-30\) =~ s/\\b\(\\p\{Alpha\}\+\)\\b/\\u\\L$1/g;
&nbsp;    perlretut\.pod&#8203;:string\. The regexps C\<\\L\\u$word> or C\<\\u\\L$word> convert the first

They don't work because you cannot guarantee a correct titlecase
mapping if you first send it through lowercase\.  There are no 
roundtrip guarantees with Unicode casemapping\.

Here are two places where you get an error doing it the way
the pods erroneously suggest\, but there are others&#8203;:

&nbsp;    orig  => İ is 0130
&nbsp;       lc => i̇ is 0069\.0307
&nbsp;    tc    => İ is 0130
&nbsp;    tc lc => İ is 0049\.0307     \(wrong answer\)

&nbsp;    orig  => ẞ is 1E9E
&nbsp;       lc => ß is 00DF
&nbsp;    tc    => ẞ is 1E9E
&nbsp;    tc lc => Ss is 0053\.0073    \(wrong answer\)

The correct approach requires something more like

&nbsp;    s/\\b\(\\w\)\(\\w\*\)\\b/\\u$1\\L$2/g;  \# "titlecase" "words"

Because casemapA\(string\) is never guaranteed to be the
same as casemapA\(casemapB\(string\)\)\.  

\-\-tom

\#\!/usr/bin/env perl

use utf8;

use v5\.14;
use strict;
use warnings;
use open qw\(&#8203;:std :encoding\(UTF\-8\)\);
use charnames qw\(&#8203;:full\);

my @&#8203;chars = \(
&nbsp;    "\\N\{LATIN CAPITAL LETTER I WITH DOT ABOVE\}"\,
[...]

PR #19616, opened 2022-04-11
Use proper titlecase technique in pods

khwilliamson:
This closes GH #11568.

The faq pod in that ticket needs more extensive updating.  Ticket [https://github.com/perl-doc-cats/perlfaq/issues/100](url) has been filed


Grinnz (approved):

Grinnz:
for reference, this is because the original mechanism of lowercase-then-ucfirst does not work for some unicode text that doesn't cleanly round trip between upper and lower case.

PR #19618, opened 2022-04-12
Update IO-Compress to 2.106 and add a way to skip utils/ files that need 64 bit ints 

demerphq:
On 32 bit perls IO-Compress/bin/zipdetails fails t/porting/utils.t via its copy into the utils directory. This patch adds a way for scripts that end up in utils to advise they require 64 bit perls.

```
 # Failed test 83 - utils/zipdetails compiles at porting/utils.t line 85
 #      got "Integer overflow in hexadecimal number at utils/zipdetails line 1432.
 Integer overflow in hexadecimal number at utils/zipdetails line 2247.
 Integer overflow in hexadecimal number at utils/zipdetails line 2248.
 Integer overflow in hexadecimal number at utils/zipdetails line 2249.
 Integer overflow in hexadecimal number at utils/zipdetails line 2250.
 Integer overflow in hexadecimal number at utils/zipdetails line 2251.
 Integer overflow in hexadecimal number at utils/zipdetails line 2252.
 Integer overflow in hexadecimal number at utils/zipdetails line 2253.
 Integer overflow in hexadecimal number at utils/zipdetails line 2254.
 utils/zipdetails syntax OK
 "
 # expected "utils/zipdetails syntax OK
 "
 t/porting/utils .................................................. FAILED at test 83
```

This patch adds a way that utils of this form can exit early with a
warning containing the magic phrase "requires 64 bit integers", in
which case the compile test will be skipped.

I will push a corresponding patch to https://github.com/pmqs/zipdetails
which is the upstream repository for cpan/IO-Compress/bin/zipdetails
which is hosted at https://github.com/pmqs/IO-Compress

This problem seems to have started when IO-Compress 2.105 was merged
in 131aabd8eb09322f5fc1ad33ed607d45390ee16d.

This includes the fix to IO-Compress 2.106, and should now pass tests on 32 bit perls (it does on my local test builds).

This also includes a fix for our github pipelines that was caused by the fix for the git CVE-2022-24765.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -211,7 +211,9 @@ jobs:
       # actions/checkout@v2 doesn't work in a container, so we use v1.
       - uses: actions/checkout@v1
       - name: fix git remote credential
-        run: git remote set-url origin "https://$GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY"
+        run: |
+          git config --global --add safe.directory /__w/perl5/perl5
+          git remote set-url origin "https://$GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}@github.com/$GITHUB_REPOSITORY"
       - name: git cfg + fetch tags
         run: |
           git config diff.renameLimit 999999
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -634,7 +634,7 @@ use File::Glob qw(:case);
     },
 
     'IO-Compress' => {
-        'DISTRIBUTION' => 'PMQS/IO-Compress-2.105.tar.gz',
+        'DISTRIBUTION' => 'PMQS/IO-Compress-2.106.tar.gz',
         'MAIN_MODULE'  => 'IO::Compress::Base',
         'FILES'        => q[cpan/IO-Compress],
         'EXCLUDED'     => [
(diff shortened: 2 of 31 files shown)


pmqs:
@demerphq I just know that I'll mess about with the message you added in `zipdetails`. I can make sure that the string "_requires 64 bit integers_" is still present, but an exclusion list feels more robust.

demerphq:
On Tue, 12 Apr 2022 at 14:10, Paul Marquess ***@***.***>
wrote:

> @demerphq <https://github.com/demerphq> I just know that I'll mess about
> with the message you added in zipdetails. I can make sure that the string
> "*requires 64 bit integers*" is still present, but an exclusion list
> feels more robust.
>

Not sure I follow, but as long as we apply the patch to perl and you
include the magic phrase when we test I have no objections whatsoever. :-)

cheers,
Yves

-- 

demerphq:
Given that this PR changes very little actual core code, and that is only a porting test, primarily involves an update to IO-Compress which involved patches that required co-ordination with its maintainer, and fixes issues that break our entire CI pipeline, eg:

```
Run git remote set-url origin "***[github.com/$GITHUB_REPOSITORY](http://github.com/$GITHUB_REPOSITORY)"
fatal: unsafe repository ('/__w/perl5/perl5' is owned by someone else)
To add an exception for this directory, call:

git config --global --add safe.directory /__w/perl5/perl5
Process completed with exit code 128.
```

I am merging it right now. Note it passed *all* tests before I did this. The final forced push just updated the commit message for the fix for the above bug.

PR #19627, opened 2022-04-15
Update IO.xs to conform to ppport.h 3.68 recommendations

toddr:
While updating IO for release to CPAN, I discovered this advice from `ppport.h`. I'd prefer this go to blead before releasing to CPAN but I understand the situation. Looking for advice here.


demerphq (approved):
LGTM

PR #19628, opened 2022-04-15
Update Dumper.xs to conform to ppport.h 3.68 recommendations

toddr:


dist/Data-Dumper/Dumper.xs
```diff
@@ -335,7 +339,7 @@ esc_q_utf8(pTHX_ SV* sv, const char *src, STRLEN slen, I32 do_utf8, I32 useqq)
                 * first byte */
                 increment = (k == 0 && *s != '\0') ? 1 : UTF8SKIP(s);
 
-                r = r + my_sprintf(r, "\\x{%" UVxf "}", k);
+                r = r + my_snprintf(r, "\\x{%" UVxf "}", k);
```
demerphq, line 342: I dont think this can be right. my_snprintf() has the signature:

```
len = my_snprintf(ptr, max_len, format, ...);
```

eg there is already an example in the file:

```
len = my_snprintf(tmpbuf, sizeof(tmpbuf), "%" IVdf, SvIV(val));
```

dist/Data-Dumper/Dumper.xs
```diff
@@ -922,7 +926,7 @@ DD_dump(pTHX_ SV *val, const char *name, STRLEN namelen, SV *retval, HV *seenhv,
 		
 		ilen = inamelen;
 		sv_setiv(ixsv, ix);
-                ilen = ilen + my_sprintf(iname+ilen, "%" IVdf, (IV)ix);
+                ilen = ilen + my_snprintf(iname+ilen, "%" IVdf, (IV)ix);
```
demerphq, line 929: Same issue with snprintf here. Also for both cases, shouldnt you really be checking if it wrote what you expect?

Also from the manpages for snprintf:

> The  functions  snprintf()  and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was truncated due to this limit, then the return value is the number of  characters  (excluding the terminating null byte) which would have been written to the final string if enough space had been available.  Thus, a return value of size or more means that the output was truncated.

So you should be checking if that happened. eg, it should look something like:

```
len = my_sprintf(iname+ilen, max_len, "%" IVdf, (IV)ix);
if (len >= max_len) croak("panic: buffer too small");
ilen += len;
```
how you compute max_len I leave to you. :-)

demerphq (changes requested):

dist/Data-Dumper/Dumper.xs
toddr: This change is purely a recommendation from ppport.h it sounds like I need to mention it in #19632

```diff
Analysis completed (13 warnings)
Suggested changes:
--- ./Dumper.xs
+++ ./Dumper.xs.patched
@@ -3,6 +3,10 @@
 #include "perl.h"
 #include "XSUB.h"
 #ifdef USE_PPPORT_H
+#define NEED_load_module
+#define NEED_my_strlcpy
+#define NEED_PL_parser
+#define NEED_vload_module
 #  define NEED_my_snprintf
 #  define NEED_my_sprintf
 #  define NEED_sv_2pv_flags
@@ -335,7 +339,7 @@
                 * first byte */
                 increment = (k == 0 && *s != '\0') ? 1 : UTF8SKIP(s);
 
-                r = r + my_sprintf(r, "\\x{%" UVxf "}", k);
+                r = r + my_snprintf(r, "\\x{%" UVxf "}", k);
                 continue;
             }
 
@@ -530,7 +534,7 @@
         PUSHMARK(SP);
         XPUSHs(val);
         PUTBACK;
-        count = call_sv((SV*)re_pattern_cv, G_ARRAY);
+        count = call_sv((SV*)re_pattern_cv, G_LIST);
         SPAGAIN;
         if (count >= 2) {
             sv_flags = POPs;
@@ -922,7 +926,7 @@
 		
 		ilen = inamelen;
 		sv_setiv(ixsv, ix);
-                ilen = ilen + my_sprintf(iname+ilen, "%" IVdf, (IV)ix);
+                ilen = ilen + my_snprintf(iname+ilen, "%" IVdf, (IV)ix);
 		iname[ilen++] = ']'; iname[ilen] = '\0';
                 if (style->indent >= 3) {
 		    sv_catsv(retval, totpad);
@@ -1665,7 +1669,7 @@
                         sv_catsv(retval, style.sep);
 		    }
 		    SvPVCLEAR(valstr);
-		    if (gimme == G_ARRAY) {
+		    if (gimme == G_LIST) {
 			XPUSHs(retval);
 			if (i < imax)	/* not the last time thro ? */
 			    retval = newSVpvs_flags("", SVs_TEMP);
@@ -1681,7 +1685,7 @@
 	    }
 	    else
 		croak("Call to new() method failed to return HASH ref");
-	    if (gimme != G_ARRAY)
+	    if (gimme != G_LIST)
 		XPUSHs(retval);
 	}
 

```

PR #19630, opened 2022-04-15
Update Time-HiRes.xs to conform to ppport.h 3.68 recommendations

toddr:


atoomic (approved):

atoomic:
Should not the distro also use `ppport.h`?

dist/Time-HiRes/HiRes.xs
```diff
@@ -20,6 +20,7 @@ extern "C" {
 #include "XSUB.h"
 #include "reentr.h"
 #ifdef USE_PPPORT_H
+#define NEED_ck_warner
 #  include "ppport.h"
```
atoomic, line 24: note that `ppport.h` is not included in the distro

toddr: https://github.com/Perl/perl5/pull/19640/files

PR #19631, opened 2022-04-15
Update Unicode-Normalize.xs to conform to ppport.h 3.68 recommendations

toddr:


demerphq (commented):
I dont really get this patch, it doesnt change anything but add an include to ppport that it doesnt seem to use, what am I missing?

demerphq:
This package doesnt include ppport.h

toddr:
From what I can tell, this is the only dist module that does not include ppport.h. It seemed like we should add it here but I'm also happy for this to be closed as: NOPE!

demerphq:
Well you need to add ppport.h to the dist if this patch is to build. :-) Ill leave whether it makes sense to someone who knows ppport.h better than I do.

toddr:
> Well you need to add ppport.h to the dist if this patch is to build. :-) Ill leave whether it makes sense to someone who knows ppport.h better than I do.

That's not how ppport.h works in any of the other dist/ modules?

demerphq:
On Sat, 16 Apr 2022, 02:06 Todd Rinaldo, ***@***.***> wrote:

> Well you need to add ppport.h to the dist if this patch is to build. :-)
> Ill leave whether it makes sense to someone who knows ppport.h better than
> I do.
>
> That's not how ppport.h works in any of the other dist/ modules
>
Sure, they all include a copy.

Do a

  find . -name ppport.h

Cheers,
Yves

>

toddr:
These are generated files
```
$>git ls-files|grep ppport.h
dist/Devel-PPPort/ppport_h.PL
```

demerphq:
>These are generated files

Not really sure what your point is.  This PR is failing because it can't find ppport.h, because your patch doesn't add it. Each dist that uses ppport.h needs to have its own copy. When you use Deve::PPPort it injects the latest copy into the repo.

Leont:
> Not really sure what your point is. This PR is failing because it can't find ppport.h, because your patch doesn't add it. Each dist that uses ppport.h needs to have its own copy.

It really just needs the file to exist. It might make more sense to have them be empty (but existent) files, and only add the ppport.h when releasing to CPAN.

demerphq:
Provided we are confident that it doesn't get released without one I can live with that, but it doesn't fill me with warmth and joy, I'd prefer we just generate the correct version so we can test it is up to date.

But the real question for this PR is why include it at all if it's not needed.

khwilliamson:
Having looked at this, I now realize that U:N should have been using ppport.h for a long time.  That fixes a bunch of UTF-8 bugs and holes that this otherwise could be subject to.  I'll work on a full PR for it

Any module on cpan that calls UTF-8 perl functions should be using Devel::PPPort to get the modern treatment.

toddr:
Closing. now #19637 has been opened for @khwilliamson to look at.

Leont:
> Provided we are confident that it doesn't get released without one I can live with that, but it doesn't fill me with warmth and joy, I'd prefer we just generate the correct version so we can test it is up to date.

We are actually already doing what I described for several dists. However, I just discovered we already have a script (mkppport) to generate them in core (with mkppport.list as configuration), so my point is probably moot. We should just add these dists to the configuration (and in some cases always make the dist use ppport)

demerphq:
@Leont good catch, although note its `mkppport.lst`, no 'i'.

demerphq:
Following up from the comment from @leont I have created https://github.com/Perl/perl5/pull/19640

demerphq:
On Sat, 16 Apr 2022, 02:11 Todd Rinaldo, ***@***.***> wrote:

> These are generated files
>
> $>git ls-files|grep ppport.h
> dist/Devel-PPPort/ppport_h.PL
>
>
Check the test fail, it can't find ppport.h, you need to add one. You can't
include a file that doesn't exist.

Yves

>

khwilliamson:
superceded by #19640

Issue #19603, opened 2022-04-09
t/re/uniprops02.t failures when NV is "doubledouble" long double


sisyphus:
For quite some time, t/re/uniprops02.t has been failing some tests whenever I build perl such that the type of NV is the "doubledouble" long double.

The first test that fails is:
```
not ok 30410 - couldn't compile /\p{Numeric_Value:3.12e-03}/; line 69958: Can't find Unicode property definition "Numeric_Value:3.12e-03" in regex; marked by <-- HERE in m/\p{Numeric_Value:3.12e-03} <-- HERE / at (eval 26922) line 2.
```
This is in response to the following line of code in lib/unicore/TestProp.pl:
```
Expect(1, 73684, '\p{Numeric_Value:3.12e-03}', "");
```
On other systems, that line of the generated TestProp.pl specifies <code>3.13e-03</code> instead of <code>3.12e-03</code>.
(The test passes if the value of 3.13e-03 is used.)

On this "doubledouble" sysyem of mine, the value of 3.12e-03 (or equivalent) is repeated in various other tests throughout TestProp.pl, causing those tests to fail in every instance.
Perhaps this is indicative of a problem with the way that the test values are generated by lib/unicore/mktables ?

But there's another problem, too.
For values of 3.125e-03 (or equivalent), which are also repeated in various tests throughout TestProp.pl, these tests are failing (in the same way) on the "doubledouble" build, but passing on other systems.

Sometimes "doubledoubles" are not assigned correctly (ie can be off by one or two ULPs), but this doesn't seem to be relevant here.
In fact, one value that is off by three ULPs (3.13e-03) is handled correctly, while the one that is assigned correctly (3.125e-03) fails.

I guess it's not critical that all perl bugs related to this very rare type of NV be fixed.
OTOH, perl would be no worse off if it could be fixed.
Please let me know if there's something further I can do to help with this.

**Steps to Reproduce**

$ perl -le 'use re; my $re = qr[\p{Numeric_Value: 3.12e-03}];'
Can't find Unicode property definition "Numeric_Value: 3.12e-03" in regex; marked by <-- HERE in m/\p{Numeric_Value: 3.12e-03} <-- HERE / at -e line 1.
$ perl -le 'use re; my $re = qr[\p{Numeric_Value: 3.125e-03}];'
[...]

PR #19634, opened 2022-04-17
refactored and improved minimal perfect hash code (regen/mph.pl)

demerphq:
Some time back I wrote regen/mph.pl for @khwilliamson to use in parsing our unicode properties. I had completely forgotten this code and rediscovered it when debugging https://github.com/Perl/perl5/issues/19603 (I wrote it at a time when I was under great stress after my mom passed away).

I noticed some odd things when it ran and decided to clean it up, fix the issues I saw and generally reacquaint myself with the code. While digging into it I realized it was missing certain optimizations and ran very very slowly. The version of regen/mk_invlists.pl in blead right now takes about 11.5 minutes to run on my laptop, about 9.5 of that coming from regen/mph.pl

I fixed the major of the cause of the slowness by removing "use bignum" which had been added to make it work on 32 bit perls, and then added some optimizations and improvements. The new version of regen/mk_invlists runs in about 3.5 minutes, with about 1 minute due to mph.pl and at the same time it produces a significantly (~20%) smaller blob for the keys.

Some context: mph.pl constructs a two level minimal perfect hash, and it constructs a minimal covering string of split keys. The two level minimal perfect hash algorithm involves building a special type of hash tale with exactly the number of buckets as there are keys.  The minimal covering set of split keys is a way of compressing the keys such that they are represented as a prefix and a suffix. For our use case (And many use cases) it turns out that by splitting the keys we can save a lot of space by reusing the prefixes and suffixes or parts of them, for multiple keys. 

This PR includes several improvements to both algorithms. 

The improvements to the minimal perfect hash construction means it should execute in nearly O(N) time, and the improvements to the minimal covering string algorithm mean it produces a smaller buffer to represent the keys. 

The algorithm for the minimal covering string algorithm "_squeeze" is by a former colleague of mine Ilya Sashcheka who I have unfortunately lost contact with and no longer have an email for, but I have added him to the AUTHORS file none the less. He gave me permission to use this code back when we worked together but it just never made it into core until now. 

The combination of the two algorithms means we can represent the entire data set in about 65% of the size of the raw keys themselves. This is substantially smaller than the size of code we would produce if we used Devel::Tokenizer:C which would be much larger than the 104237 bytes of raw keys we need to represent.

The net result of this patch is that it takes about 3.5 minutes to run regen/mk_invlists.pl and it produces a key buffer that is 8635 verus the current 10617 bytes. IOW, it saves about 7 minutes and 2k of ram.  As a bonus this version of the code is easier to work on and maintain as I have converted it to use OO for the public API, and has way more comments about how it works.

Included in this PR is a fix for t/op/magic.t that i noticed while working on the PR.


regen/mph.pl
```diff
@@ -32,160 +112,382 @@ package MinimalPerfectHash;
...
+    # solution for the hard cases. The buckets can be divided into three
+    # levels of difficulty to solve "hard", "medium" and "trivial".
+    #
+    # * Hard buckets have more than one item in them.
+    # * Medium buckets have one item whose hash is above $MAX_SEED2.
+    # * Trivial buckets have one item whose hash is not above $MAX_SEED2.
+    #
+    # Each type of bucket uses a different algorithm to solve. Note that
+    # a "classical" two level hash would only have "hard" and "trivial"
+    # buckets, but since we supporint having a larger hash value than we
```
haarg: typo

demerphq: Fixed.

khwilliamson (changes requested):
LGTM, but it needs to regen mk_invlist to not fail porting tests.  The CI won' run without that fix.  I did it on my box locally, and all tests passed.

demerphq:
I modified the old last commit to include the mk_invlist.pl changes, and I added an additional sanity check patch that validates the split keys buffer and regenerated again.

PR #19635, opened 2022-04-17
Update Test::Harness to 3.44

Leont:
This includes changes to support taint-free perl and removes ASCII-isms to better work on EBCDIC

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -1080,7 +1080,7 @@ use File::Glob qw(:case);
     },
 
     'Test::Harness' => {
-        'DISTRIBUTION' => 'LEONT/Test-Harness-3.42.tar.gz',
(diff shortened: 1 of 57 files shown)


neilb:
LGTM

neilb (commented):
LGTM

khwilliamson (approved):

Leont:
@jkeenan This probably should not have been merged, given that ASAN builds apparently break.

The quick fix is to skip that one test, I think I'll do that for now.

jkeenan:
On 4/18/22 15:00, Leon Timmermans wrote:
> @jkeenan <https://github.com/jkeenan> This probably should not have been 
> merged, given that ASAN builds apparently break.
> 

Sorry about that.  I checked matrix.cpantesters.org before hitting the 
button.  Test-Harness is about the last place I would have expected ASAN 
complications.
> The quick fix is to skip that one test, I think I'll do that for now.
> 

Good enough.  Thanks.

PR #19636, opened 2022-04-17
regexec.c: Rmv obsolete recursion checks: GH #8369

### LINKED ISSUE #8369: Can exceed regex complex recursion limit
Migrated from [rt.perl.org#38717](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=38717) (status was 'stalled')

Searchable as RT38717$

khwilliamson:
There is no recursion to exceed limits of.

This fixes #8369

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -8782,18 +8782,8 @@ NULL
             DEBUG_EXECUTE_r(Perl_re_exec_indentf( aTHX_  "WHILEM: failed, trying continuation...\n",
                 depth)
             );
-          do_whilem_B_max:
-            if (cur_curlyx->u.curlyx.count >= REG_INFTY
-                && ckWARN(WARN_REGEXP)
-                && !reginfo->warned)
-            {
-                reginfo->warned	= TRUE;
-                Perl_warner(aTHX_ packWARN(WARN_REGEXP),
-                     "Complex regular subexpression recursion limit (%d) "
-                     "exceeded",
-                     REG_INFTY - 1);
-            }
 
+          do_whilem_B_max:
             /* now try B */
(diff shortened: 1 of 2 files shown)


regexec.c
```diff
@@ -8804,22 +8794,6 @@ NULL
...
-                {
-                    reginfo->warned	= TRUE;
-                    Perl_warner(aTHX_ packWARN(WARN_REGEXP),
-                        "Complex regular subexpression recursion "
-                        "limit (%d) exceeded",
-                        REG_INFTY - 1);
-                }
-                cur_curlyx->u.curlyx.count--;
-                CACHEsayNO;
-            }
```
hvds, line 8821: I think this removes too much: the outer block looks to me like it is checking whether we've reached the max allowed for this (minimal) repetition. I can have a dig to verify, but I expect @demerphq will know.

If I'm correct, then we should really have a test that fails when you remove this. (I'm a bit surprised we don't have any tests of the warning, for that matter.)

khwilliamson: Thank you, I believe you are correct.

khwilliamson:
The test suite only exercises this code via the fold_grind tests, and only then because it is using a ?? quantifier

regexec.c
hvds: @khwilliamson I think this got merged without fixing the above over-removal, please could you verify?

khwilliamson:
Thanks for finding this.  Fixed by bdadb244dbc743bdb32c9b5f33d0940d55b93f8d

I'll change the issue title to indicate we need tests for this

khwilliamson:
I instead created a new ticket #19781

PR #19638, opened 2022-04-17
fix confusing configure question

cmburn:
Hi!
I feel odd over a pull request this small, but the question
`When you set your locale to something besides C.UTF8, does it do so, or just pretend to?`
is on the confusing side, since it's unclear if it's asking if it actually does, or if it pretends to. I'm aware this won't be committed, however the file itself says to submit pull requests anyway.

### DIFF
--- a/pod/perlmodlib.PL
+++ b/pod/perlmodlib.PL
@@ -204,7 +204,8 @@ double quotes should be used instead of single quotes).
       'find { wanted => sub { print canonpath $_ if /\.pm\z/ },
       no_chdir => 1 }, @INC'
 
-(The -T is here to prevent '.' from being listed in @INC.)
+(The -T is here to prevent @INC from being populated by C<PERL5LIB>,
+C<PERLLIB>, and C<PERL_USE_UNSAFE_INC>.)
 They should all have their own documentation installed and accessible
 via your system man(1) command.  If you do not have a B<find>
 program, you can use the Perl B<find2perl> program instead, which
--- a/pod/perlsec.pod
+++ b/pod/perlsec.pod
@@ -235,8 +235,8 @@ Unix-like environments that support #! and setuid or setgid scripts.)
 
 =head2 Taint mode and @INC
 
-When the taint mode (C<-T>) is in effect, the environment variables
-C<PERL5LIB> and C<PERLLIB>
++When the taint mode (C<-T>) is in effect, the environment variables
++C<PERL5LIB>, C<PERLLIB>, and C<PERL_USE_UNSAFE_INC>
 are ignored by Perl.  You can still adjust C<@INC> from outside the
 program by using the C<-I> command line option as explained in
 L<perlrun|perlrun/-Idirectory>.  The two environment variables are


jkeenan:
> Hi! I feel odd over a pull request this small, but the question `When you set your locale to something besides C.UTF8, does it do so, or just pretend to?` is on the confusing side, since it's unclear if it's asking if it actually does, or if it pretends to. I'm aware this won't be committed, however the file itself says to submit pull requests anyway.

We will try to summon `Configure` experts: @Tux @doughera88

khwilliamson:
We'll absolutely make a clarification.   But I'm not sure your proposal is sufficiently helpful.   Here
are the salient points:

> Some` systems fake their locales, so that they pretend to accept a locale
     change, but they either do nothing, making everything the C locale, or
     on some systems there is a a second locale "C-UTF-8" that can be
     switched to.  Configure probes have been added to find such systems. 
 Anything running the current musl library has this characteristic, as does OpenBSD and its kin, as
they view locales as security risks. 

So how to boil that down to a comprehensible question to override the Configure probe?  One possibility, is to add something like

> Accept the default if this question doesn't make much sense to you

With this added information, do you have a different wording?

cmburn:
Hmm... not quite, maybe something like
`Some systems do not actually change locales when told to do so, in particular OpenBSD and systems with a musl libc. Does your system fall in this category?

Accept the default if this question doesn't make much sense to you`

I use OpenBSD systems a good deal (including the system I'm building for), as well as musl systems if I'm running in Docker or something, I might not think of that. It might be good to define those systems! I'm not attached to anything here, so obviously feel free to do whatever works best :)

khwilliamson:
reopening to track the progress of this

khwilliamson:
woops, won't let me reopent

doughera88:
On Mon, Apr 18, 2022 at 06:29:52AM -0700, cmburn wrote:
> Hmm... not quite, maybe something like
> `Some systems do not actually change locales when told to do so, in particular OpenBSD and systems with a musl libc. Does your system fall in this category?

I agree that a generic 'some systems' is often too vague to be helpful,
but more specific lists of systems are impractical to maintain, and we
have no real way to know if they will be useful or correct in a couple
of years.  Possibly including a specific version (e.g. OpenBSD X.Y)
would be helpful -- it would at least give a reader 5 years from now a
sense that the data in the question is old.

-- 
    Andy Dougherty		***@***.***
    Dept. of Physics
    Lafayette College, Easton PA 18042

khwilliamson:
On 4/19/22 08:44, doughera88 wrote:
> On Mon, Apr 18, 2022 at 06:29:52AM -0700, cmburn wrote:
>  > Hmm... not quite, maybe something like
>  > `Some systems do not actually change locales when told to do so, in 
> particular OpenBSD and systems with a musl libc. Does your system fall 
> in this category?
> 
> I agree that a generic 'some systems' is often too vague to be helpful,
> but more specific lists of systems are impractical to maintain, and we
> have no real way to know if they will be useful or correct in a couple
> of years. Possibly including a specific version (e.g. OpenBSD X.Y)
> would be helpful -- it would at least give a reader 5 years from now a
> sense that the data in the question is old.

The text here is wrong.  musl on any box had this problem (and maybe 
still does; I don't know their philosophy on this).

So a version for that would be good.

It's unlikely OpenBSD will ever change, because they think locales are a 
security hole, and a major point of their OS is to be secure.

PR #19642, opened 2022-04-18
Customize a few modules to work with z/OS

khwilliamson:
z/OS works with both ASCII and EBCDIC character sets; that it works with ASCII was unknown to many people (including me).  We've had contributions this release cycle to get the ASCII mode to work.  There are still several failures in the test suite on it.  due to some peculiarities in the OS.  This commit makes minor corrections to a few /cpan modules to bring the failures down to just a few.

Some modules assumed EBCDIC for z/OS.  This commit series changes them to test for the character set, not the OS.

Some locales on z/OS improperly make some space characters into also being graphic.  A commit changes this so graphic characters can't be spaces.  We already do something similar for Windows.

getnameinfo() on z/OS requires both its return buffers to be non-NULL.  POSIX allows one to be.  A commit in this branch causes it to be called with both non-NULL.

podlators, in spite of testing for EBCDIC, includes some ASCII-isms.  Since I already had to customize the module, I included a commit to remove these.

If I were the maintainer of any of these modules, I would think it not worth it to release a new version just to accommodate a new platform.  So I have not yet issued pull requests for the modules here.  But, as Perl core, it makes sense to accommodate a new platform.

### DIFF
--- a/cpan/Socket/Socket.pm
+++ b/cpan/Socket/Socket.pm
@@ -3,7 +3,7 @@ package Socket;
 use strict;
 { use v5.6.1; }
 
-our $VERSION = '2.032';
+our $VERSION = '2.033';
 
 =head1 NAME
 
--- a/cpan/Socket/Socket.xs
+++ b/cpan/Socket/Socket.xs
@@ -733,8 +733,13 @@ static void xs_getnameinfo(pTHX_ CV *cv)
 #endif
 
 	err = getnameinfo((struct sockaddr *)sa, addr_len,
-			want_host ? host : NULL, want_host ? sizeof(host) : 0,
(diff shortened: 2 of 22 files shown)


Leont:
> If I were the maintainer of any of these modules, I would think it not worth it to release a new version just to accommodate a new platform. So I have not yet issued pull requests for the modules here. But, as Perl core, it makes sense to accommodate a new platform.

No, you must make pull requests for those modules, otherwise your changes will be overwritten again if they release an update.

haarg:
The version numbers used here need to be changed. When making changes to `cpan/` modules, the version numbers should be underscore versions. `2.032_01`, not `2.033`.

Issue #19643, opened 2022-04-18
cpan/Test-Harness/t/harness.t: Test failure under Address Sanitizer


jkeenan:
https://github.com/Perl/perl5/pull/19635 requested synching of the latest CPAN version of Test-Harness into blead.  [CPANtesters results](http://matrix.cpantesters.org/?dist=Test-Harness) were good and the p.r. was favorably reviewed, so I hit the "rebase and merge" button.

Uh-oh!  Didn't see those CI failures under ASAN.  @LeonT is working on this.  Here, for the record, is the failure on my own machine.
```
$ uname -mrs
Linux 5.4.0-107-generic x86_64

$ ./perl -Ilib -V:config_args
config_args='-des -Dusedevel -DDEBUGGING -Dcc=clang -fsanitize=address -Dld=clang -fsanitize=address';

$ cd t;./perl harness ../cpan/Test-Harness/t/harness.t; cd -
../cpan/Test-Harness/t/harness.t .. 84/133 AddressSanitizer:DEADLYSIGNAL
=================================================================
==101927==ERROR: AddressSanitizer: SEGV on unknown address 0x03e800018e27 (pc 0x7fa0f167d40b bp 0x7fff4e180b60 sp 0x7fff4e180938 T0)
==101927==The signal is caused by a READ memory access.
    #0 0x7fa0f167d40b in kill /build/glibc-sMfBJT/glibc-2.31/signal/../sysdeps/unix/syscall-template.S:78
    #1 0x9829e9 in Perl_apply /home/jkeenan/gitwork/perl/doio.c:2719:26
    #2 0x94eddf in Perl_pp_chown /home/jkeenan/gitwork/perl/pp_sys.c:3744:28
    #3 0x73f685 in Perl_runops_debug /home/jkeenan/gitwork/perl/dump.c:2634:23
    #4 0x56e663 in S_run_body /home/jkeenan/gitwork/perl/perl.c
    #5 0x56de8e in perl_run /home/jkeenan/gitwork/perl/perl.c:2628:9
    #6 0x4e0bec in main /home/jkeenan/gitwork/perl/perlmain.c:110:9
    #7 0x7fa0f165e0b2 in __libc_start_main /build/glibc-sMfBJT/glibc-2.31/csu/../csu/libc-start.c:308:16
    #8 0x438bdd in _start (/home/jkeenan/gitwork/perl/perl+0x438bdd)

AddressSanitizer can not provide additional info.
[...]

PR #19644, opened 2022-04-18
Skip Test::Harness t/harness.t SEGV test

Leont:
This is a workaround for #19643.

This will need a better fix later on, but for now let's unblock the CI

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -1093,6 +1093,9 @@ use File::Glob qw(:case);
                 t/lib/if.pm
                 ),
         ],
+        'CUSTOMIZED' => [
+            't/harness.t'
+        ]
     },
 
     'Test::Simple' => {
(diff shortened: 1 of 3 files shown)


jkeenan:
> This is a workaround for #19643.
> 
> This will need a better fix later on, but for now let's unblock the CI

LGTM on my ASAN build.
```
$ cd t;./perl harness -v ../cpan/Test-Harness/t/harness.t; cd -

ok 1 - HARNESS_ACTIVE env variable should be set
[snip]
ok 81 - An object of class 'TAP::Parser' isa 'TAP::Parser'
ok 82 - ... and the status line should be correct
ok 83 - ... and the output should be correct
ok 84 - ... and the status line should be correct
ok 85 - ... and the output should be correct
ok 86 # skip Skipping for now because of ASAN failures
ok 87 - ... and the status line should be correct

[snip]
ok 133 - _add_descriptions: different exts, two already cooked
ok
All tests successful.
Files=1, Tests=133,  1 wallclock secs ( 0.07 usr  0.00 sys +  0.66 cusr  0.23 csys =  0.96 CPU)
Result: PASS
```

PR #19645, opened 2022-04-19
mg.c, Cwd.pm - Empty path is the same as "." which is forbidden under taint

demerphq:
Having a relative path, including ".", is forbidden under taint. On *nix
    an empty PATH or an empty PATH component is equivalent to a PATH of ".",
    so they should be forbidden as well.
    
    Note that on Windows the current working directory is ALWAYS checked
    first if you try to execute something that does not specify its path,
    regardless of the PATH.
    
    I do not know what happens on VMS and I do not have access to a
    VMS environment to test. There are totally different codepaths for
    VMS as well. This patch does not (or rather should not) change
    behavior for VMS.

### DIFF
--- a/dist/PathTools/Cwd.pm
+++ b/dist/PathTools/Cwd.pm
@@ -3,7 +3,7 @@ use strict;
 use Exporter;
 
 
-our $VERSION = '3.84';
+our $VERSION = '3.85';
 my $xs_version = $VERSION;
 $VERSION =~ tr/_//d;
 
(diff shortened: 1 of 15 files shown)


demerphq:
Dangit. Both of these patches are revert reverts. I fat fingered this morning and merged it to blead instead of updating this PR.  Apparently the patch breaks local::lib and possibly other things, and given how close we are to 5.36 it might be prudent to wait till 5.37 to deal with it.

haarg:
local::lib has been fixed to deal with this change.

dist/PathTools/Cwd.pm
```diff
@@ -192,8 +192,14 @@ sub _backtick_pwd {
...
     # finding 'sh' in the PATH.
-    my @localize = grep exists $ENV{$_}, qw(PATH IFS CDPATH ENV BASH_ENV) if $^O ne "amigaos";
+    my @localize = grep exists $ENV{$_}, qw(IFS CDPATH ENV BASH_ENV) if $^O ne "amigaos";
     local @ENV{@localize} if @localize;
+    # empty PATH is the same as "." on *nix, so localize it to /something/
+    # we won't *use* the path as code above turns $pwd_cmd into a specific
+    # executable, but it will blow up anyway under taint. We could set it to
+    # anything absolute. Perhaps "/" would be better.
+    local $ENV{PATH}= "/usr/bin"
+        if $^O ne "vms" and $^O ne "amigaos";
```
hvds, line 202: > This patch does not (or rather should not) change behavior for VMS.

@demerphq Note that this _is_ a change of behaviour under VMS: we no longer localize `$ENV{PATH}` to undef, but we also don't localize it to `'/usr/bin'`.

I think the change to line 195 is not needed, and should probably be undone.

(Sorry I didn't think to look at this until after it was merged.)

demerphq: @hvds: see https://github.com/Perl/perl5/pull/20576

PR #19646, opened 2022-04-19
GH16319: avoid recursion parsing 'pack' template

hvds:
A template with many open brackets or open parentheses could
overflow the stack, modify the parsing loop to avoid that.


hvds:
I think this should work, but it should probably be deferred until after ~~5.38~~ 5.36 release.

khwilliamson:
I hope you mean s/5\.38/5.36

hvds:
> I hope you mean s/5.38/5.36

Oops yes, sorry.

demerphq (approved):

PR #19649, opened 2022-04-20
The new bool type is missing the usual cohort of utility functions, also add detailed tests for the bool internals.

demerphq:
The new bool "type" does not have the usual complement of utility
functions and macros. It only has one encapsulating function, which is
perfectly reasonable for most use cases where one wants to test if an
SV* is a bool, but does a bit too much if one is working on a
serialization tool which is likely to want to unroll a nice chunk of the
logic.

The new type also lacks the usual cohort of utility functions to create
new bool SV's or set SV to be a bool efficiently.

This patch adds the following functions:

    newSVbool(const bool bool_val)
    newSV_true()
    newSV_false()

    sv_set_true(SV *sv)
    sv_set_false(SV *sv)
    sv_set_bool(SV *sv, const bool bool_val)

All of these functions are currently simple wrappers around existing functions like sv_setsv() and newSVsv(). My assumption is that over time we might see the functions unrolled like we have done with sv_set_undef() and similar cases so that they are more efficient than using newSVsv() or sv_setsv(), the key point is we expose the API in a sane way, later one we can make them fast.

And the utility following macros for efficiently testing if an SV is IOK and POK in one step.

    SvPIOK(sv)
    SvPIOK_off(sv)
    SvPIOK_on(sv)

The following three are intended very specifically for writing
serialization code like for Sereal, where it is reasonable to want to
unroll the logic contained in Perl_sv_isbool() and SvTRUE(). They are
documented as "you dont want to use this except under special
circumstances". The assumption is that a caller has already validated the SV is PIOK and can then use these specialty functions.

    PIOK_sv_isbool(sv)
    PIOK_sv_isbool_true(sv)
    PIOK_sv_isbool_false(sv)

In regard to the latter set of three macros I am well aware of the API debates that are occurring in https://github.com/Perl/perl5/pull/19600 and I fully agree with the API opinions expressed by @iabyn and @leonerd about the *B* modules and Perl level coding. However in a module like Sereal or perhaps JSON::XS having lower level encapsulations of the logic is pretty important. 

There is no way that Sereal will ever execute Perl_sv_isbool() for every SV it serializes, it simply would duplicate too much logic that already needs to be in place, and I assume that any XS code author of a serialization module would say the same thing. Furthermore the way the "value" of a bool is represented internally required validation against a hard coded constant value, unlike every other type where a member of the struct can be accessed to find its value.  So for performance reasons we need to do something that otherwise might be a bit "dirty". This is better than the alternative of XS coders using their own unrolled code and having it break in a future release. As such we need to encapsulate the task of saying "is this a false bool" and "is this a true" bool in an efficient and streamlined way. If Perl does not expose a macro to do this then people will roll their own. So this is an attempt to Do The Right Thing, while strongly discouraging the misuse of the macros. 

The fact that I am proposing these macros for the XS layer for specialized purposes is NOT in any way me supporting these types of API to the B package. B is always going to be slow and unrolling logic like this makes no sense. In a serialization package like Sereal where performance is a primary design objective I believe the case is different.  

Thanks for your consideration.


ext/XS-APItest/t/boolean.t
```diff
@@ -43,4 +43,8 @@ use XS::APItest;
     # We won't test AELEM but it's likely to be the same
 }
 
+
+is(test_bool_internals(),0,"Bulk test internal bool related APIs");
```
leonerd: The prevailing style in this file puts spaces after commas; this should read

```
is(test_bool_internals(), 0, "Bulk test internal bool related APIs");
```

sv.c
```diff
@@ -4662,6 +4662,67 @@ Perl_sv_set_undef(pTHX_ SV *sv)
...
+Perl_sv_set_false(pTHX_ SV *sv)
+{
+    PERL_ARGS_ASSERT_SV_SET_FALSE;
+    sv_setsv(sv, &PL_sv_no);
+}
+
+/*
+=for apidoc sv_set_bool
+
+Equivalent to C<sv_setsv(sv, &PL_sv_no)>, but may be made more
```
leonerd: no or yes. ;)

sv.h
```diff
@@ -746,6 +746,40 @@ Unsets the PV status of an SV.
...
+=for apidoc Am|void|SvPIOK_on|SV* sv
+Tells an SV that is a string and a number in one operation. Equivalent
+to C<SvIOK_on(sv); SvPOK_on(sv);> but more efficient.
+
+=for apidoc Am|void|SvPIOK_off|SV* sv
+Unsets the PV and IV status of an SV in one operation. Equivalent to
+C<SvIOK_off(sv); SvPK_off(v);> but more efficient.
+
+=for apidoc Am|bool|PIOK_sv_isbool|SV* sv
+Checks if a PIOK sv is a bool. Note that it is the callers
```
leonerd: "caller's" - requires apostrophe

sv.h
```diff
@@ -746,6 +746,40 @@ Unsets the PV status of an SV.
...
+useful in specialized logic where the flags have already been checked to
+be correct. Almost always you should be using the function
+C<sv_isbool(sv)> instead.
+
+=for apidoc Am|bool|PIOK_sv_isbool_true|SV* sv
+Checks if a PIOK sv is a true bool. Note that it is the callers
+responsibility to ensure that the sv is SvPIOK. This macro is only
+useful in specialized logic where the flags have already been checked to
+be correct. This is NOT what you should use to check if an SV is "true",
+for that you should be using the function C<SvTRUE(sv)> instead.
```
leonerd: s/function/macro/ perhaps? I don't know if we usually bring attention to how many of our "API functions" are in fact macros.

sv.h
```diff
@@ -746,6 +746,40 @@ Unsets the PV status of an SV.
 Tells an SV that it is a string and disables all other C<OK> bits.
 Will also turn off the UTF-8 status.
 
+=for apidoc Am|U32|SvPIOK|SV* sv
```
leonerd: I like the idea of this family of test/set macros, but unfortunately the naming is problematic.

We already have a set of macros `SvNIOK` that test if an SV has `NOK` **or** `IOK`, so to have a similarly-named one `SvPIOK` which tests for both `POK` **and** `IOK` is going to cause confusion. Is there some way we can work in the words "and" or "both" or somesuch into its name?

leonerd (changes requested):
Comments left above.

Overall: idea is good, some names need adjustment, some docs wording fixes also.

demerphq:
Thanks. Will fix. Also the test fails, whatever they are.

leonerd:
The two Linux/i386 machines fail with:

```
Undefined subroutine &main::test_bool_internals called at t/boolean.t line 47.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 12.
```

Weird that all the 64bit machines are fine. I wonder what happened...

demerphq:
Yeah, ill build a 32 bit perl and check it out. Might not be fixed up till tomorrow. But ill get to it all. THanks for the feedback!

demerphq:
Just pushed a bunch of changes based on your feedback here and comments in chat.  Hope this is moving toward a place you are more comfortable with.

I havent built 32 bit yet. I need to do a manual configure to do that and I figured i could push and see if there were any fails on github while i do it

sv.h
demerphq: Renamed them to SvIandPOK (this font doesnt make it as obvious as I chose that), there arent any words that i can think of that start with Iand (that is and 'I' not an 'L') but there a bunch that start with Pand so it seemed more obvious the 'and' is a separate word.

ext/XS-APItest/t/boolean.t
demerphq: Fixed, I changed all the cases I noticed in all the files i touched to use a space after comma. (I may have missed some tho I guess)

sv.c
demerphq: dumb-de-dumb-dumb. fixed :-)

sv.h
demerphq: yeah, I am terrible with apostrophe, and I havent managed to get hunspell to do them right either. :-( sorry about that. fixed.

sv.h
demerphq: tried to avoid both terms in all the examples I checked. I think there was one case I couldnt avoid tho. (i forget which, sleepy now)

demerphq:
BTW, summary of what i changed from the main fixup commit:

* Fixed up yves style assignment.
* Fixed up whitespace after commas.
* typos and verbiage, pod C<> and B<> tags.
* renamed SvPIOK() to SvIandPOK()
  I tried SvPandIOK() and it seems less intuitive what it means
  as "Pand" is a proper prefix to various words, whereas "Iand" is not,
  at least in English. (That is an I not an L)
* Created a wrapper SvBoolFlagsOK() to abstract the actual flags in use.
* Renamed the PIOK_ prefixed functions to BOOL_INTERNALS_ prefixes.

inline.h
```diff
@@ -3451,8 +3451,7 @@ Perl_mortal_getenv(const char * str)
 PERL_STATIC_INLINE bool
 Perl_sv_isbool(pTHX_ const SV *sv)
 {
-    return SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) &&
-        (SvPVX_const(sv) == PL_Yes || SvPVX_const(sv) == PL_No);
+    return SvBoolFlagsOK(sv) && BOOL_INTERNALS_sv_isbool(sv);
```
leonerd: As the rest of this PR is purely additions, I think the rest of it is fine to go in to 5.35.11 at this late hour.

However, these three lines aren't *exactly* equivalent any more because of the slightly different way the flags tests are done. I'd feel more comfortable approving it without this part to it - we can maybe add a comment though to remind ourselves to update it to using this form when we start development for 5.37.

leonerd (changes requested):

inline.h
demerphq: Sure, done. Although I have to admit I don't really understand your concern here, at the mathematical / logical level they are exactly equivalent operations if they weren't on some machine then something would be very very wrong with that machine. I suspect that an optimizing compiler will change the less efficient form into the more efficient form anyway.

FWIW I ran this:
```
PERL_STATIC_INLINE bool
Perl_sv_isbool(pTHX_ const SV *sv)
{
#ifdef DEBUGGING
    bool n = (SvBoolFlagsOK(sv) && BOOL_INTERNALS_sv_isbool(sv));
    bool o = (SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) &&
            (SvPVX_const(sv) == PL_Yes || SvPVX_const(sv) == PL_No));
    assert(n==o);
#endif
    return SvIOK(sv) && SvPOK(sv) && SvIsCOW_static(sv) &&
        (SvPVX_const(sv) == PL_Yes || SvPVX_const(sv) == PL_No);

}
```
and no asserts fired, as expected. Anyway, I have pushed a patch as your requested. I used an ifdef 0 to leave the new code with a comment that we will switch in 5.37.

demerphq:
BTW, @leonerd could you please hit resolve on the issues you have raised that have been cleared please? It doesn't feel right for me to do it myself.

sv.h
leonerd: A better name, even if the font github uses makes it look like the word "land". In a real code editor with a proper font I'm sure it looks fine.

inline.h
leonerd: Oh nice - I like the `#if 0` approach to this.
Thanks.

sv.c
```diff
@@ -4662,6 +4662,67 @@ Perl_sv_set_undef(pTHX_ SV *sv)
...
 
+/*
+=for apidoc sv_set_true
+
+Equivalent to C<sv_setsv(sv, &PL_sv_yes)>, but may be made more
+efficient in the future. Doesn't handle set magic.
+
+The perl equivalent is C<$sv = !0;>.
+
+Introduced in perl 5.35.11.
```
leonerd, line 4673: The 5.35.11 ship has sailed now; so this is inaccurate. We may consider a 5.35.12, or if not we'll have to wait for 5.37 now. :/ I hope it's the former but it is an open discussion.

leonerd (approved):
Aside from open question of the version number in the docs, I'm generally happy with this.

I'd suggest hold off on merging it for now though. Technically we're in the "no changes before 5.36" freeze, but we may manage to work something out.

atoomic:
@leonerd any chance we can apply the same treatment to #19600 and #19599?
and have them merged before 5.36, so we include and provide these APIs on Day One.
Thanks

demerphq:
Rebased and squashed the main fixups, and updated the commit message to reflect those changs. I did not squash the commit that enables the new mode. Please review the message for the first commit especially as I did the update quickly due to lack of time. It is likely I will not be able to look into this PR until tomorrow Singapore time.

leonerd:

PR #19652, opened 2022-04-21
regen/mph.pl - make sure the author of _squeeze() has a commit in the log

demerphq:
This commit is actually by the committer, and is intended to ensure that someone looking for what the author wrote can find it and that his name ends up in the 5.36 authors list, etc. It took me a while to get a email address for him or I would have done this in eda35008b17e739922 which is where his work on the _squeeze() split key algorithm was added. Credit where credit is due and all of that. Thanks Ilya.


demerphq:
@sas4eka this is for you bud. thanks again!

PR #19653, opened 2022-04-21
Correction to commit ecef3a65d2

steve-m-hay:
Don't make everything writable when $opts{e} is not set


khwilliamson:
My apologies for not saying I would be working on this; as I broke it, and I have access to the hardware to test any changes.

In looking at the code, I realized that the original commit didn't make sense.  It was thinking an earlier incarnation of the code was still in effect.  There is no reason for either EBCDIC or ASCII builds to have the files (beyond the exception list) writable.  So simply reverting that commit was all that was needed.  I did that and tested it on an EBCDIC box; without reading my mail to find that you @steve-m-hay had issued a PR.

PR #19660, opened 2022-04-24
perlapi: Document SvPVXtrue

khwilliamson:


sv.h
```diff
@@ -2024,13 +2024,25 @@ scalar.
 #  define SvPVbytex_nolen(sv) ((PL_Sv = (sv)), SvPVbyte_nolen(PL_Sv))
 #endif /* __GNU__ */
 
+/*
+=for apidoc Am|bool|SvPVXtrue|SV * sv
+
+Returns a boolean as to whether C<sv> contains a PV that is considered TRUE or
+not.  FALSE is returned if C<sv> doesn't contain a PV, or if the PV it does
+contain is length 1 byte, and points to an (illegal) low memory address given
+by the representation of the digit 0 (hence is 0x30 on ASCII platforms).
```
leonerd: That isn't what it does. That description would match code that looked like

```c
(sv)->sv_u.svu_pv != '0'
```

The code in this macro dereferences the pointer; this is the good ol'

```c
*ptr != '0'
```

for a regular string comparison.

A more accurate description would be that it 

    returns TRUE for any PV of length greater than one, or length exactly one whose character
    is not exactly '0'. Returns FALSE for length zero or the exact string "0".

leonerd (changes requested):

sv.h
leonerd: Much improved now :)

sv.h
```diff
@@ -1985,13 +1985,24 @@ scalar.
...
 #endif /* __GNU__ */
 
+/*
+=for apidoc Am|bool|SvPVXtrue|SV * sv
+
+Returns a boolean as to whether or not C<sv> contains a PV that is considered
+TRUE.  FALSE is returned if C<sv> doesn't contain a PV, or if the PV it does
+contain is zero length, or consists of just the single character '0'.  Every
+other PV value is considered TRUE.
+
```
leonerd, line 1997: I don't know if it's standard practice to point out, but this macro may evaluate the `sv` argument more than once. I know that bit me before on the `SvTRUE()` macro, by doing `SvTRUE(POPs)`. It might be worth adding a note here too?

leonerd: I see you've added :)

leonerd (approved):
LGTM

PR #19663, opened 2022-04-25
Enable perl to be compiled using mingw-w64 runtimes 9 and later

sisyphus:
Hi,
Up to and including perl-5.36.0, per source does not compile with mingw-w64 compilers that have mingw runtime versions >= 9.
See #18911.
This PR request is part of the fix that I've been using, without issue, for the last few months.
The other part is a patch to Socket.xs. (I'll update this PR with a link to that Socket bug report once it has been created - ie shortly after submitting this PR.)

According to my testing, this PR can be applied at any time (after release of 5.36.0, of course) without breaking anything that isn't already broken.
But it won't actually fix anything until Socket has been updated to blead with the Socket.xs patch.

Similarly, the corrected Socket distro can be uploaded to blead at any time without breaking anything that isn't already broken, but will also fail to fix anything until this PR has been pushed. It's perhaps safest to upload the corrected Socket distro to blead first though, as I said, my testing indicates that it doesn't matter which is done first.

Cheers,
Rob


sisyphus:
The Socket bug report is at:
https://rt.cpan.org/Ticket/Display.html?id=142390

The patch that I've requested be made to Socket,xs is:
```
--- Socket.xs_socket_poll       2022-04-25 15:08:37 +1000
+++ Socket.xs   2022-04-25 19:37:46 +1000
@@ -98,6 +98,11 @@

 #endif

+/* MinGW-w64 ports of gcc with mingw runtime version >= 9 already
+ * provide inet_pton() and inet_ntop() in ws2tcpip.h.
+ */
+#if(!defined(__MINGW64_VERSION_MAJOR) || __MINGW64_VERSION_MAJOR < 9)
+
 static int inet_pton(int af, const char *src, void *dst)
 {
   struct sockaddr_storage ss;
@@ -146,6 +151,8 @@
     return dst;
 }

+#endif /* mingw runtime version 9 exclusions */
+
 #define HAS_INETPTON
 #define HAS_INETNTOP
 #endif
```

win32/config.gc
```diff
@@ -307,8 +307,8 @@ d_ilogbl='define'
 d_inc_version_list='undef'
 d_index='undef'
 d_inetaton='undef'
-d_inetntop='undef'
-d_inetpton='undef'
+d_inetntop='define'
```
Leont, line 310: Won't that cause issues on older versions? Or did those already have these functions but just needed to include winsock2.h

sisyphus: I wondered about that for the first time as I was creating that PR ... but I didn't really follow it through because I had found no issue with it in the testing I had done. (This morning I again ran a build of patched blead using gcc-8.3.0, runtime 6.0.0, including 'make test' ... just to be sure.)

The thing is that the current blead version of Socket.xs unconditionally provides inet_ntop() and inet_pton() for win32.
It then unconditionally defines both HAS_INETNTOP and HAS_INETPTON for win32 (at lines 149 and 150).
Aside: I can't actually find any C/XS code that looks for or utilises those 2 definitions. They seem to be purposeless.

I don't think that the contents of config.gc and config_H.gc (wrt these symbols) has any influence on the functionality of perl, irrespective of gcc version and mingw runtime version.
But the absence of these config.gc definitions might conceivably result in tests being unnecessarily skipped.

So these changes to config.gc and config_H.gc are not really part of the fix for this issue - and sorry for not having been aware enough to mention that.
They are, IMO. changes that could have been made (and should have been made) a while back.
It seems like an opportune moment to introduce them now.

Cheers,
Rob

Leont (approved):

PR #19667, opened 2022-04-25
Davem/refactor warnings pl

iabyn:
Refactor regen/warnings.pl. The src for this tool was a mess: it started out 20 years ago as a small utility, then grew. Lots of badly-named global variables, no comments about what subs do etc. This series of commits cleans things up, refactors, adds comments etc. In theory it makes no functional changes, and indeed as run currently it generates exactly the same warnings.h and lib/warning.pm files as before.

I think this can be safely merged before 5.36.0. It's a tool that's only used when developing the perl interpreter (and then only when adding a new warnings category), so even if were broken, it would make no different to an end user of 5.36.0.

### DIFF
--- a/regen/warnings.pl
+++ b/regen/warnings.pl
@@ -165,12 +165,10 @@ my %NAME_TO_VALUE; # ('NAME'       => index_number,       ....);
 sub valueWalk
 {
     my ($tre, $v_list) = @_;
-    my @list = () ;
     my ($k, $v) ;
 
     foreach $k (sort keys %$tre) {
         $v = $tre->{$k};
-        die "duplicate key $k\n" if defined $CATEGORIES{$k} ;
         die "Value associated with key '$k' is not an ARRAY reference"
             if !ref $v || ref $v ne 'ARRAY' ;
 
--- a/regen/warnings.pl
+++ b/regen/warnings.pl
@@ -152,11 +152,19 @@ my %VALUE_TO_NAME; # (index_number => [ 'NAME', version ], ...);
 
 my %NAME_TO_VALUE; # ('NAME'       => index_number,       ....);
 
-my %v_list = () ;
+###########################################################################
+
+# Generate a hash with keys being the version number and values
+# being a list of node names with that version, e.g.
+#
+# { '5.008' => [ 'all', 'closure', .. ], 5.021' => .... }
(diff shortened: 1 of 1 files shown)


hvds (approved):
All looks good to me. In commit messages:
  `s/orderValues()/valueWalk()/` in "eliminate broken dup check"
  `s/on;y/only/` in "eliminate global %v_list var"

Now that `%tree` has been renamed, it would also be nice to rename the various in-function `$tre` variables to `$tree`.

leonerd (approved):
LGTM

I'm in two minds about whether to merge it before 5.36 or not. As you observe, it's only used as part of the regen process when we add a new warning, which we're not going to do before 5.36. So on the one hand adding it now won't affect anything, but on the other hand there's no urgent need to.

We could merge it now, or apply the "defer-next-dev" tag to remind us to do it as part of 5.37 startup.

Leont (approved):
Given the fact that verifying it doesn't change any output is trivial, I don't really see a reason to postpone it.

iabyn:
On Thu, Apr 28, 2022 at 05:51:12AM -0700, Paul Evans wrote:
> @leonerd approved this pull request.
> 
> LGTM
> 
> I'm in two minds about whether to merge it before 5.36 or not. As you
> observe, it's only used as part of the regen process when we add a new
> warning, which we're not going to do before 5.36. So on the one hand
> adding it now won't affect anything, but on the other hand there's no
> urgent need to.

I forgot to mention that I want to do some moderate-scale reworking of
warnings.pm and its pod (starting now, for release after 5.36) which
depends on the warnings.pl changes, so having that merged now makes it
easier for me to maintain my private warnings branch before 5.36.

-- 
The Enterprise is involved in a bizarre time-warp experience which is in
some way unconnected with the Late 20th Century.
    -- Things That Never Happen in "Star Trek" #14

iabyn:
rebased, fixed the commit msgs and s/tre/tree as suggested by Hugo

hvds:
> rebased, fixed the commit msgs and s/tre/tree as suggested by Hugo

+1

iabyn:
I've now merged this. I'm not going to be very available for the next week or so, so if it causes any issues, feel free to revert it.

PR #19669, opened 2022-04-26
Synch with CPAN version 2.34

jkeenan:
Since we're in code freeze, assigning to production release manager for merge at his discretion (either before 5.36.0 or after 5.37.0).

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -250,7 +250,7 @@ use File::Glob qw(:case);
     },
 
     'CPAN' => {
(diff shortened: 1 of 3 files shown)


mbeijen:
Looking at the change set, the current (in blead) release is cpan 2.33 marked as 'TRIAL', and the to-be version is 2.34 which has this one additional fix: https://github.com/andk/cpanpm/commit/16c5f944fde41392c32b005312bd7ef6e36a322c

I'd prefer if we have the non-trial 2.34 CPAN release in perl 5.36.0

Leont (approved):

Issue #19664, opened 2022-04-25
[doc] perlsub "Constant Functions" section out of date


haarg:
**Where**
perlsub - Constant Functions

**Description**
The "Constant Functions" documentation includes several examples and descriptions that are out of date.

This code is documented as working but issuing a deprecation warning:
```perl
BEGIN {
    my $x = 79907;
    *RT_79908 = sub () { $x };
    $x++;
}
print RT_79908(); # prints 79907
```
This was true in perl 5.22, but it will now fail to compile.

The next example includes:
```perl
# Warns.  Future Perl versions will stop inlining it.
BEGIN {
    my $x;
    $x = 54321;
    *ALSO_INLINED = sub () { $x };
}
```
This also fails to compile.

iabyn:
PR  #19674 contains my proposed fix to the docs.

iabyn:
fixed by f8ed0a9151; closing.

PR #19674, opened 2022-04-28
update const sub docs for stuff that now croaks

iabyn:
GH #19664

Some types of constant sub candidates emitted deprecation warnings
from 5.22 onwards, and croaked from 5.32 onwards. The perlsub docs knew
about the warnings; this commit updates them for the croaks.

At the same time, change the code examples to no longer use the RT
ticket number of the original issue. For one thing, RT numbers are
obsolete, and for another, it was being too cute - setting a variable to
one less than the RT number, then showing that the constant value
hadn't been incremented by the $x++. And using the number for the sub
names. The examples just became noisy with random 5-digit numbers
everywhere.

I consider this a 5.36.0 candidate

### DIFF
--- a/pod/perlsub.pod
+++ b/pod/perlsub.pod
@@ -1733,29 +1733,31 @@ has no other references.  Only the first example here will be inlined:
         *NOT_INLINED = sub () { $var };
     }
 
-A not so obvious caveat with this (see [RT #79908]) is that the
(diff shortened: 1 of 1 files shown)


leonerd (approved):
LGTM.

And yes - a docs fix like this is fine for 5.36.

haarg (approved):

Issue #19657, opened 2022-04-22
Collection of issues related to new procedure for disabling 'taint', including 5.36 blockers.


demerphq:
**Description**
The new procedure for disabling taint support has several issues, some of which are merely annoying technical debt, some of which are show-stoppers.

Background: The logic to support taint mode, even when taint mode is not enabled, impose a non-trivial performance burden on all perl code executed. Conservative estimates for disabling it at compile time are a 3%-5% speedup, but some workloads may see significantly more performance win. In Perl 5.18 options were provided to disable taint support at compile time, the 'NO_TAINT_SUPPORT' define and the 'SILENT_NO_TAINT_SUPPORT' define. Enabling either removes taint related logic from the compile process, NO_TAINT_SUPPORT results in a perl which will throw a fatal exception when the `-t` option is used to enable taint, and SILENT_NO_TAINT_SUPPORT builds a perl which silently ignores the `-t` switch, so scripts previously using it will continue to run, albeit without taint protections.

In PR #19541 a new procedure was added to control these features. This involved adding support for a 'taint_support' fake define which can be used with Configure via `-Utaint_support` to "undefine" this fake define. This then defines 'SILENT_NO_TAINT_SUPPORT'. There does not seem to be a way to enable the safer 'NO_TAINT_SUPPORT' mode other than the "old way". 

1. Using the old way of `-Accflags='-DNO_TAINT_SUPPORT'` or `-Accglags='-DSILENT_NO_TAINT_SUPPORT'` results in a perl whose %Config is incorrect and which states that $Config{taint_support}='define', even though that is not true. This is *definitely* a show-stopper.
2. There is no documentation of 'taint_support' in INSTALL. If this is an "official" setting then INSTALL should include documentation. It is minimally documented in perldelta. This is not a show-stopper but is unfortunate, arguably the new feature is incomplete without it.
3. The new method does not provide way to enable the safer "NO_TAINT_SUPPORT". I think this is a show stopper.
4. perl -V does not show this setting. This is not a show stopper as the NO_TAINT_SUPPORT and SILENT_NO_TAINT_SUPPORT defines will be shown in the define setting, but I think it is unfortunate and unhelpful.
[...]

PR #19679, opened 2022-05-01
Revert "Add Configure question for taint support"

leonerd:
This reverts commit 39f8eb4a21670e6b973dcfc86d0b1339064f5642 which was added in PR #19541

This is because of a variety of issues discussed https://github.com/Perl/perl5/issues/19657 and at the PSC meeting 064 2022-04-22 https://www.nntp.perl.org/group/perl.perl5.porters/2022/04/msg263670.html


jkeenan:
To evaluate this p.r. I built blead at fc995cfb17c56da9b8762746fc42f10ef7f9fc0a with my regular configuration options, then built @leonerd's branch the same way, then compared their respective `config.sh` files.  Results:
```
$ diff -w -I "# Configuration time" -I api_subversion -I api_version -I api_versionstring -I archlib -I archlibexp -I cf_time -I installarchlib -I installprivlib -I installsitearch -I installsitelib -I patchlevel -I perlpath -I privlib -I privlibexp -I sitearch -I sitearchexp -I sitelib -I sitelibexp -I startperl -I subversion -I version -I version_patchlevel_string -I PERL_API_VERSION -I PERL_API_SUBVERSION config.sh.blead config.sh.leonerd
1125d1124
< taint_support='define'
```
Looks good.

Just for reference, here's a similar diff starting with `v5.34.1`, thereby showing us our "net configuration changes from 5.34" assuming this p.r. is merged:
```
$ diff -w -I "# Configuration time" -I api_subversion -I api_version -I api_versionstring -I archlib -I archlibexp -I cf_time -I installarchlib -I installprivlib -I installsitearch -I installsitelib -I patchlevel -I perlpath -I privlib -I privlibexp -I sitearch -I sitearchexp -I sitelib -I sitelibexp -I startperl -I subversion -I version -I version_patchlevel_string -I PERL_API_VERSION -I PERL_API_SUBVERSION config.sh.v5.34.1 config.sh.leonerd
224a225,226
> d_ffs='define'
> d_ffsl='define'
431a434,435
> d_nl_langinfo_l='define'
> d_non_int_bitfields='define'
598a603
> d_strxfrm_l='define'
611a617
> d_thread_local='define'
986a993
> perl_thread_local='_Thread_local'
1097a1105,1106
> st_dev_sign='1'
> st_dev_size='8'
1215a1225
> xlocale_needed='define'
1221,1222c1231,1232
< PERL_VERSION=34
< PERL_SUBVERSION=1
---
> PERL_VERSION=35
> PERL_SUBVERSION=12
```

leonerd:
This is still needing some review. @neilb , @rjbs , anyone else?

Leont:
I think this missed perlsec.pod, that needs a documentation update to also be reverted.

neilb:
As @Leont noted, perlsec.pod needs the section on "Checking for taint suppot" removing, and the reference to that section earlier on.

The second paragraph in section "Strictness and warnings" of perlmodstyle should be removed.

In perlipc.pod remove this text:
```
Note that perl can be built without taint support,
in which case -T silently does nothing
(see L<perlsec> for how to check if your perl support taint checking).
```

replace it with: "Note that perl can be built without taint support. There are two different modes: in one, -T will silently do nothing. In the other mode -T results in a fatal error". 

In perllocale.pod, remove this:
```
See L<perlsec> for how to tell if your perl supports taint checking.
```

In perlrun.pod, remove this:
```This has no effect if your perl was built without taint support
(see L<perlsec>).
```

and this:

```
If your perl has been built without taint support, then this option
has no effect. See L<perlsec> for how to check whether your perl
supports taint checking.
```

leonerd:
@neilb I believe that's all the docs changes you listed. PTAL.

Should I squash these before merge, or leave them as two commits?

pod/perlipc.pod
```diff
@@ -1132,9 +1132,9 @@ even if we aren't running setuid or setgid.  This is always a good idea
 for servers or any program run on behalf of someone else (like CGI
 scripts), because it lessens the chances that people from the outside will
 be able to compromise your system.
-Note that perl can be built without taint support,
-in which case -T silently does nothing
-(see L<perlsec> for how to check if your perl support taint checking).
+Note that perl can be built without taint support.  THere are two
```
Leont: "THere"

Leont: Shouldn't the preceding sentence be removed entirely?

leonerd: Hrm, that was in Neil's suggested text. I think it's still technically true, it's just you have go to the long way round with `-Accflags=...` rather than using a configure question.

rjbs:
At this point, what remains before this is merged?  IMHO, this is the only _known_ blocker.

PR #19689, opened 2022-05-03
Move sv.h functions from inline.h to sv_inline.h

richardleach:
Move `-- sv.h --` functions from _inline.h_ into _sv_inline.h_, so that all functions inlined from _sv.*_ are in one place. This was suggested by @khw.


khwilliamson (approved):

khwilliamson:
rebased and merged via 2356f8bb5077e29623007d815dab58f43473a56d

Issue #19690, opened 2022-05-03
t/op/magic.t failing on AlmaLinux 8.


toddr:
I built using:
```
./Configure -des -Dusedevel -Duseithreads -Doptimize="-O2 -pipe -fstack-protector -fno-strict-aliasing"
```

The verbose output is a little hard to read but...
```
t/op/magic ... 1..208
ok 1 - *SIG appears to be defined at the outset
ok 2 - *^OPEN appears to be defined at the outset
ok 3 - *^TAINT appears to be defined at the outset
...
ok 99 - Assigning $^I does not clobber $^O
ok 100 - test $^P = 0
ok 101 - test $^P = 1
ok 102 - Set process name through argv[0] ( 136098 pts/0    S+     0:00 Good Morning, Dave                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           )
ok 103 - Set process name through prctl() ( 136098 pts/0    S+     0:00 Good Morning, D)
ok 104 - compare $0 to UTF8-flagged
ok 105 - compare $0 to non-UTF8-flagged
# Failed test 106 - /proc cmdline shows as expected (compare to UTF8-flagged) at op/magic.t line 480
#      got "?\000                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           "
not ok 106 - /proc cmdline shows as expected (compare to UTF8-flagged)
# expected "?"
# Failed test 107 - /proc cmdline shows as expected (compare to non-UTF8-flagged) at op/magic.t line 482
[...]

PR #19691, opened 2022-05-04
Make t/op/magic.t pass on AlmaLinux 8’s kernel.

FGasper:
Issue #19690

### DIFF
--- a/t/op/magic.t
+++ b/t/op/magic.t
@@ -469,7 +469,11 @@ EOP
     open my $rfh, '<', "/proc/$$/cmdline"
       or skip "failed to read '/proc/$$/cmdline': $!", $skip;
     my $got = do { local $/; <$rfh> };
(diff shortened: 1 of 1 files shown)


rwp0 (commented):
would be good to explicitly mention kernels/distros that do so in the comment as a short example

FGasper:
> would be good to explicitly mention kernels/distros that do so in the comment as a short example

I don’t disagree, but I don’t know the full spate of systems that have been tested here. I’ll add a couple notes, though.

rwp0 (approved):
that's better, thanks

toddr (approved):

Leont (approved):

leonerd (approved):
LGTM

leonerd:
This has four approves now, and is a 5.36 release blocker. Can you merge it now please?

PR #19692, opened 2022-05-05
Fix a typo in a comment in Carp.pm

nwc10:
Bump $CARP::VERSION and $Carp::Heavy::VERSION

Probably not worth "merging" this until 5.37.1 (ie cherry-pick it then)

### DIFF
--- a/dist/Carp/lib/Carp.pm
+++ b/dist/Carp/lib/Carp.pm
@@ -179,7 +179,7 @@ BEGIN {
             ? do { require "overload.pm"; _fetch_sub overload => 'mycan' }
             : \&UNIVERSAL::can;
 
-        # _blessed is either UNIVERAL::isa(...), or, in the presence of an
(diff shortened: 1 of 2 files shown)


khwilliamson (approved):

khwilliamson:
I believe this qualifies as something you can push without a PR

rwp0 (approved):
a change of a letter

Leont (approved):

nwc10:
> I believe this qualifies as something you can push without a PR

Nomally yes, and I would have. But we're in code freeze, so it's supposed to be ask the ~~pumpking~~ PSC first. And this sort of minor change *isn't* something that I would have liked someone *else* to apply when I was trying to create a stable release. It's just congnative load. (I believe that I'm the only person in this thread (so far) who has cut any release, let alone a stable release. Stable releases are scary)

I pushed it so that it didn't get forgotten.

PR #19701, opened 2022-05-09
ExtUtils::CBuilder - print out commands being run in usable form

haarg:
When not set to be quiet, the system commands being run are printed out
before running them. This can assist with diagnosing issues. It is
implied that this is a command that could be run by a user.

Since the command is internally run with system using list form, it does not
involve the shell, and does not need any extra quoting. The printed out
command however, would be run by a user through a shell. Just joining
the arguments with spaces will not quote them correctly for use through
a shell.

Add a new method for quoting a string for the shell. And use this method
when printing out the command line that we are running.


khwilliamson (commented):
lgtm

Issue #19704, opened 2022-05-10
disabling bareword filehandles erroneously forbids package method calls


ilmari:
**Description**

When the `bareworld_filehandles` feature is disabled (e.g. under `use v5.35;`), package method calls are also forbidden where a bareword filehandle would be.

**Steps to Reproduce**

```
$ perl -le 'no feature "bareword_filehandles"; sub foo { "foo" } print main->foo'
Bareword filehandle "main" not allowed under 'no feature "bareword_filehandles"' at -e line 1.
Execution of -e aborted due to compilation errors.
```
It seems to be interpreting the bareword `const` op under the `entersub` as a bareword filehandle, even though that's actually the argument to `method_named`:

```
$ perl -MO=Concise -e 'no feature "bareword_filehandles"; sub foo { "foo" } print main->foo'
Bareword filehandle "main" not allowed under 'no feature "bareword_filehandles"' at -e line 1.
-e had compilation errors.
9  <@> leave[1 ref] KP/REFC ->(end)
1     <0> enter ->2
2     <;> nextstate(main 4 -e:1) :us,*,&,{,$,fea=6 ->3
8     <@> print K ->9
3        <0> pushmark ->4
7        <1> entersub[t1] KRS/TARG,STRICT ->8
4           <0> pushmark ->5
5           <$> const(PV "main") s/BARE ->6
6           <.> method_named(PV "foo") ->7
```

**Expected behavior**

Same as with the feature enabled:
```
$ perl -le 'sub foo { "foo" } print main->foo'
foo
```

**Perl configuration**

```
Summary of my perl5 (revision 5 version 35 subversion 10) configuration:
   
  Platform:
    osname=linux
    osvers=5.10.0-12-amd64
    archname=x86_64-linux
    uname='linux ilmari-thinkpad 5.10.0-12-amd64 #1 smp debian 5.10.103-1 (2022-03-07) x86_64 gnulinux '
    config_args='-de -Dprefix=/home/ilmari/perl5/perlbrew/perls/35.10 -Duseshrplib -Dman1dir=none -Dman3dir=none -Dusedevel -Aeval:scriptdir=/home/ilmari/perl5/perlbrew/perls/35.10/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
[...]

PR #19705, opened 2022-05-11
Put `bareword_filehandles` feature back into the v5.35 bundle

leonerd:
This is due to bug #19704, where it doesn't work properly.

It's far too close to 5.36 release to have a go at fixing it now, so the best we can do put the named feature back into the version bundle and see about fixing it next time.

We should probably also add some release notes somewhere saying that explicitly disabling it will run into this same bug, so probably advise don't do that.


wolfsage:
https://github.com/perl/perl5/commit/d5c835da0bd504cf8b12ef8b4f4f0795bd314751 was the initial removal, which also changed some documentation in regen/feature.pl:
```
- This feature is enabled under this name from Perl 5.34 onwards.  In
- previous versions it was simply on all the time.
+ This behavior was always present in versions before Perl 5.34.  In Perl 5.34,
+ it was made controllable with the C<feature> pragma, but was on by default.
+ It is not present in the C<:5.36> feature bundle, so C<use v5.36> disables
+ this feature.
```
I believe we need to revert that documentation change as well

leonerd:
Ahyes. Rebased, reverted documentation in `feature.pm` and force-pushed.

I'll squash it to a single commit before merge.

wolfsage (approved):

wolfsage:
Thanks, LGTM

PR #19707, opened 2022-05-11
perlapi: Document call_atexit

khwilliamson:


Leont:
It's already implied, but it may be useful to be explicit that when using threads this will run when the current or any descending thread ends (so it may run multiple times).

khwilliamson:
@leont, thanks fixed

PR #19556, opened 2022-03-22
build: add configurations to compile perl for windows/arm64

niyas-sait:
This change is adding a few missing configurations to compile `Perl` for windows/arm64 targets.

All tests are not passing yet on windows/arm64 but this could be a starting point.

```
Test Summary Report
-------------------
porting/customized.t                                               (Wstat: 0 Tests: 22 Failed: 22)
  Failed tests:  1-22
porting/pending-author.t                                           (Wstat: 0 Tests: 1 Failed: 1)
  Failed test:  1
porting/pod_rules.t                                                (Wstat: 0 Tests: 5 Failed: 4)
  Failed tests:  2-5
  Parse errors: Bad plan.  You planned 7 tests but ran 5.
porting/regen.t                                                    (Wstat: 65280 Tests: 12 Failed: 1)
  Failed test:  12
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 42 tests but ran 12.
../cpan/CPAN-Meta-YAML/t/30_yaml_spec_tml.t                        (Wstat: 512 Tests: 0 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output
../cpan/CPAN-Meta-YAML/t/31_local_tml.t                            (Wstat: 512 Tests: 0 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output
../cpan/CPAN-Meta-YAML/t/32_world_tml.t                            (Wstat: 512 Tests: 0 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output
../cpan/Term-Cap/test.pl                                           (Wstat: 512 Tests: 45 Failed: 2)
  Failed tests:  16, 22
  Non-zero exit status: 2
../cpan/Win32/t/Names.t                                            (Wstat: 256 Tests: 16 Failed: 1)
  Failed test:  6
  Non-zero exit status: 1
../ext/IPC-Open3/t/IPC-Open3.t                                     (Wstat: 0 Tests: 45 Failed: 0)
  TODO passed:   25
Files=2745, Tests=1152241, 1996 wallclock secs (81.53 usr + 10.36 sys = 91.89 CPU)
```

PR #19709, opened 2022-05-11
Win32: synch with CPAN version 0.59

jkeenan:
As discussed in:
https://www.nntp.perl.org/group/perl.perl5.porters/2022/05/msg263707.html

As explained in the p5p list thread starting at the URL above, since we're in code freeze, this merge request should be completed only once perl-5.37.0 has been released and our next development cycle has begun.  However, I'm assigning this to the production release manager for his consideration.

### DIFF
--- a/cpan/Win32/Win32.pm
+++ b/cpan/Win32/Win32.pm
@@ -8,7 +8,7 @@ package Win32;
     require DynaLoader;
 
     @ISA = qw|Exporter DynaLoader|;
-    $VERSION = '0.58';
+    $VERSION = '0.59';
     $XS_VERSION = $VERSION;
     $VERSION = eval $VERSION;
 
@@ -306,6 +306,8 @@ sub PRODUCT_EDUCATION_N                      () { 0x7A } # Windows 10 Education 
 
 sub PRODUCT_UNLICENSED                       () { 0xABCDABCD } # product has not been activated and is no longer in the grace period
 
+sub PROCESSOR_ARCHITECTURE_ARM64   ()   { 12 }     # ARM64
+sub PROCESSOR_ARCHITECTURE_ARM     ()   { 5 }      # ARM
 sub PROCESSOR_ARCHITECTURE_AMD64   ()   { 9 }      # x64 (AMD or Intel)
 sub PROCESSOR_ARCHITECTURE_IA64    ()   { 6 }      # Intel Itanium Processor Family (IPF)
 sub PROCESSOR_ARCHITECTURE_INTEL   ()   { 0 }      # x86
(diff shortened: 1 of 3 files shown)


p-b-o:
Showing that this failure is the only one left for passing testsuite on Windows on Arm: https://github.com/Perl/perl5/pull/19556#issuecomment-1124586208

Leont:
Should this be a 5.36 blocker? It looks to me like it should be

Leont (approved):

jkeenan:
On 5/14/22 08:34, Leon Timmermans wrote:
> Should this be a 5.36 blocker? It looks to me like it should be
> 

This is going to be the production release manager's call, but I argue 
that it should *not* be a 5.36.0-release-blocker.

As far as I can tell, until three days ago we (P5P) were completely 
unaware of the existence of the Linaro-Microsoft-Qualcomm team's efforts 
to build applications on the OS/Platform combination of MSWin32 and ARM 
(as mentioned in 
https://www.nntp.perl.org/group/perl.perl5.porters/2022/05/msg263707.html). 
  While perl has been built on that operating system and that platform 
*separately*, I don't think we've ever previously received smoke-test 
reports or bug tickets about building perl on their *combination*.  So 
we're not currently set up to do ongoing QA on this combination.

That being the case, to designate this p.r. as a release blocker would 
mean delaying the release of 5.36 for all other OS/Platform 
combinations.  The benefit is too small to warrant that delay -- 
especially given the fact that for decades the common expectation has 
been that no one will build and install perl on Windows "right out of 
the box," i.e., directly from the P5P source code as opposed to some 
packaging of perl such as ActivePerl or Strawberry Perl.

We (P5P) certainly have been remiss with respect to Perl on Win32. 
AFAICT, right now there is exactly *one* person on the planet generating 
smoke-test reports for Perl 5 blead on Win32.  Just one.  Moreover, one 
of our principal smoke-test aggregators has for some reason simply 
stopped reporting results on Win32 
(https://rt.cpan.org/Ticket/Display.html?id=142752).  With regard to 
Windows we are pretty much flying blind.

We need to correct these problems in our QA apparatus.  Simultaneously 
we need to establish more regular communication with the people 
contributing to https://github.com/perl-libwin32/win32/.  But I don't 
think we should further delay perl-5.36.0 on this account.

p-b-o:
As @jkeenan said, it's not worth blocking the release planned.

Our goal here was simply to contribute for this *new* platform, before investigating more  on this. It's a difficult situation, because there is no public free machine (any CI runner) available now. On our side, we use Azure with Linaro's account (arm64 machines have been available for a few weeks), but of course, it's not free.

It's a problem we have with the different projects we are contributing to, because accepting changes people can't check by themselves is not an ideal situation.

PR #19712, opened 2022-05-11
perlapi: Document several *_dup functions

khwilliamson:
I don't know what to say about the 'param' parameter


Leont:
Should these be public API? I can't think of any use-case for them outside of perl_clone

khwilliamson:
I'm fine with making them internal.  What about sv_dup() which I thought might not have the same treatment?

Leont:
>  What about sv_dup() which I thought might not have the same treatment?

Yeah, `sv_dup`/`sv_dup_inc` are the only ones that makes sense as an API IME.

PR #19716, opened 2022-05-11
document SSNEW and kin

khwilliamson:


rjbs:
@khwilliamson I renamed this PR just so it was easier to understand in the PR list.

PR #19724, opened 2022-05-13
perlapi: Document get_op_(descs|names)

khwilliamson:


khwilliamson:
Since we will be deprecating these in 5.37, and the furnished documentation is hence wrong, I see no need to follow through and fix it al up

PR #19734, opened 2022-05-16
Loosen regex in tests which validate process name.

toddr:
This addresses some kernel inconsistencies on how the
process name can be represented in the /proc file. This
specifically addresses a failure on AlmaLinux 8.


toddr:
I don't know how but the original fix (https://github.com/Perl/perl5/pull/19691) we applied to this test for AlmaLinux 8 did not work consistently. We think the regex should be loosened to this to correct the problem.

FGasper (approved):
I don't know that there will always be a NUL, but it seems a reasonable guess.

leonerd (approved):
LGTM

Issue #19726, opened 2022-05-15
Regex: parsing/notation issues under /xx modifier


wollmers:
IMHO the issues are related. That's why I submit them under one main issue.

All examples use the same boilerplate:

```
#!perl

use strict;
use warnings;
use utf8;

use 5.034.001;

# use 5.026;    # /xx modifier needs 5.26
# use 5.018;    # regex_sets need 5.18
no warnings "experimental::regex_sets";

use open qw( :std :encoding(UTF-8) );
```

## 1. DON'T WARN:  "\x{0027} " is more clearly written simply as "'" in regex;

**Description**

Warns on use of `"\x{0027} "` in a character class.

**Steps to Reproduce**

Example 1 (parses within comment):

```
# matches without warning
my $apo_x_comment  = qr/

       # \x{0027} # APOSTROPHE (Other_Punctuation)
       ' # \x{0027} # APOSTROPHE (Other_Punctuation)

/xx;

if ("\x{0027}" =~ m/[$apo_x_comment]/xx) { say '$apo_x_comment matches' }

# warns
=pod
"\x{0027} " is more clearly written simply as "'" in regex; marked by <-- HERE in m/
    (?[
        \p{alpha}
        + \p{gc=Mark}
        + \p{Join_Control}
        + [(?^uxx:

       # \x{0027}  <-- HERE # APOSTROPHE (Other_Punctuation)
       ' # \x{0027} # APOSTROPHE (Other_Punctuation)

)]
    ])
/ at words.pl line 77.
"\x{0027} " is more clearly written simply as "'" in regex; marked by <-- HERE in m/
    (?[
        \p{alpha}
        + \p{gc=Mark}
        + \p{Join_Control}
        + [(?^uxx:

       # \x{0027} # APOSTROPHE (Other_Punctuation)
       ' # \x{0027}  <-- HERE # APOSTROPHE (Other_Punctuation)

)]
    ])
/ at words.pl line 77.
=cut

my $word_x_comment = qr/
    (?[
        \p{alpha}
        + \p{gc=Mark}
        + \p{Join_Control}
        + [$apo_x_comment]
    ])
/xx;
```

Example 2 (warns):

```
# matches without warning
my $apo_05 = qr/
        \x{0027} # APOSTROPHE (Other_Punctuation)
        \x{05F3} # HEBREW PUNCTUATION GERESH (Other_Punctuation)
        \x{2018} # LEFT SINGLE QUOTATION MARK (Initial_Punctuation)
[...]

PR #19735, opened 2022-05-16
perlre: Clarify /xx pod

khwilliamson:
This is a result of GH #19726

### DIFF
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -474,15 +474,18 @@ You can use L</(?#text)> to create a comment that ends earlier than the
 end of the current line, but C<text> also can't contain the closing
 delimiter unless escaped with a backslash.
 
-A common pitfall is to forget that C<"#"> characters begin a comment under
-C</x> and are not matched literally.  Just keep that in mind when trying
-to puzzle out why a particular C</x> pattern isn't working as expected.
+A common pitfall is to forget that C<"#"> characters (outside a
+bracketed character class) begin a comment under C</x> and are not
+matched literally.  Just keep that in mind when trying to puzzle out why
+a particular C</x> pattern isn't working as expected.
+Inside a bracketed character class, C<"#"> retains its non-special,
+literal meaning.
 
 Starting in Perl v5.26, if the modifier has a second C<"x"> within it,
-it does everything that a single C</x> does, but additionally
-non-backslashed SPACE and TAB characters within bracketed character
-classes are also generally ignored, and hence can be added to make the
-classes more readable.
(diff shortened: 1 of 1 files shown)


pod/perlre.pod
```diff
@@ -492,6 +492,10 @@ may be easier to grasp than the squashed equivalents
     /[d-eg-i3-7]/
     /[!@"#$%^&*()=?<>']/
 
+Note that this unfortunately doesn't mean that your bracketed classes
+can extend over multiple lines.  The closing bracket must be on the same
+line as the opening one.
```
hvds: That reads like it's a syntax error, but it actually just means newline will be one of the characters in the class. I think the second sentence could read something like: "Unless the closing bracket is on the same line as the opening one, the newline character will be part of the class just as if you'd written `C<\n>`." (Sadly that also means we'd have back-compat problems if we wanted it to ignore newlines in the future.)

The placement of the `/xx` docs directly after "A common pitfall is to forget that `"#"` characters begin a comment under `/x` ..." may also explain some of the additional confusion - the parenthetical "... (outside a bracketed character class, which is unaffected by `/x`) ..." a couple of paragraphs earlier can easily get lost here. It's not immediately obvious to me how to rearrange things to minimize the opportunity for such confusion.

khwilliamson:
On 5/16/22 18:12, Hugo van der Sanden wrote:
> Unless the closing bracket is on the same line as the opening one, the 
> newline character will be part of the class just as if you'd written 
> |C<\n>|.

Much better than what I wrote


I repushed with much more detail

pod/perlre.pod
```diff
@@ -474,15 +474,18 @@ You can use L</(?#text)> to create a comment that ends earlier than the
...
+character under C</x>.
 
 Starting in Perl v5.26, if the modifier has a second C<"x"> within it,
-it does everything that a single C</x> does, but additionally
-non-backslashed SPACE and TAB characters within bracketed character
-classes are also generally ignored, and hence can be added to make the
-classes more readable.
+it increases the effect of a single C</x>.  The only difference is that
+inside bracketed character classes, non-backslashed SPACE and TAB
+characters classes are also generally ignored, and hence can be added to
```
hvds: `s/classes //`

hvds (approved):
LGTM other than the spurious word.

I'd love to see ".. are also generally ignored" replaced with something more prescriptive some time, but I don't think it urgent to do as part of this PR.

pod/perlre.pod
hvds: ^^ that was line 487 "characters classes"

Issue #12301, opened 2012-07-29
Bleadperl v5.17.0-408-g3c13cae breaks DGL/re-engine-RE2-0.10.tar.gz


p5pRT:
Migrated from [rt.perl.org#114302](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=114302) (status was 'resolved')

Searchable as RT114302$

p5pRT:
### From @andk

git bisect


---
commit 3c13cae629d936c43bca9d992cc445d93287af8e
Author&#8203;: David Mitchell \<davem@&#8203;iabyn\.com>
Date&#8203;:   Tue Dec 13 12&#8203;:00&#8203;:12 2011 \+0000

&nbsp;    add op\_comp field to regexp\_engine API

sample fail report


---
http&#8203;://www\.cpantesters\.org/cpan/report/a83658ae\-d8ef\-11e1\-bfde\-e2730e58c4a0

I could reproduce Reini's result with

&nbsp;  BSD&#8203;::Resource&#8203;::setrlimit\(BSD&#8203;::Resource&#8203;::RLIMIT\_AS\(\)\,3\_000\_000\_000\,3\_000\_000\_000\);

Without that the test t/02\.chars\.t allocated memory like mad until
triggering oom\-killer\.

perl \-V


---
Summary of my perl5 \(revision 5 version 17 subversion 0\) configuration&#8203;:
&nbsp;  Commit id&#8203;: 3c13cae629d936c43bca9d992cc445d93287af8e
&nbsp;  Platform&#8203;:
&nbsp;    osname=linux\, osvers=3\.2\.0\-3\-amd64\, archname=x86\_64\-linux
&nbsp;    uname='linux k83 3\.2\.0\-3\-amd64 \#1 smp thu jun 28 09&#8203;:07&#8203;:26 utc 2012 x86\_64 gnulinux '
&nbsp;    config\_args='\-Dprefix=/home/src/perl/repoperls/installed\-perls/perl/v5\.17\.0\-408\-g3c13cae/165a \-Dmyhostname=k83 \-Dinstallusrbinperl=n \-Uversiononly \-Dusedevel \-des \-Ui\_db \-Uuseithreads \-Uuselongdouble \-DDEBUGGING=\-g'
&nbsp;    hint=recommended\, useposix=true\, d\_sigaction=define
&nbsp;    useithreads=undef\, usemultiplicity=undef
[...]

PR #19736, opened 2022-05-17
perlapi: Mark pregfree as internal

khwilliamson:
People should be doing the more general SvREFCNT_dec (which is what this currently expands into) instead of thinking that because this exists they need to handle compiled pattern freeing specially.

There are zero uses of this in CPAN

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -1968,7 +1968,7 @@ CiTop	|struct regexp *|ReANY	|NN const REGEXP * const re
 Apdh	|I32	|pregexec	|NN REGEXP * const prog|NN char* stringarg \
 				|NN char* strend|NN char* strbeg \
 				|SSize_t minend |NN SV* screamer|U32 nosave
-Ap	|void	|pregfree	|NULLOK REGEXP* r
+Cp	|void	|pregfree	|NULLOK REGEXP* r
 Cp	|void	|pregfree2	|NN REGEXP *rx
 : FIXME - is anything in re using this now?
 EXp	|REGEXP*|reg_temp_copy	|NULLOK REGEXP* dsv|NN REGEXP* ssv


hvds:
I assume it is ok, but I'll note:
- I don't see anything in the comments at the top of `embed.fnc`, nor in perlguts "Internal Functions" that explains what "C" means;
- there are also zero uses of this function in core, as far as I can see, which makes me wonder if there's any remaining reason for it to exist at all. (It has been that way since 288b8c02c5 in Jan 2008 as far as I can see.)

khwilliamson:
On 5/17/22 06:29, Hugo van der Sanden wrote:
> I don't see anything in the comments at the top of |embed.fnc|, nor in 
> perlguts "Internal Functions" that explains what "C" means;

Lines 84-96 in embed.fnc explain the C flag

khwilliamson:
On 5/17/22 06:29, Hugo van der Sanden wrote:
> there are also zero uses of this function in core, as far as I can see, 
> which makes me wonder if there's any remaining reason for it to exist at 
> all. (It has been that way since 288b8c0 
> <https://github.com/Perl/perl5/commit/288b8c02c5ee89a2978a1b9e56ed255c53beb793> 
> in Jan 2008 as far as I can see.)

Yes, but there's still time to get pod changes into 5.36; not for nearly 
anything else.  And this is a first step

hvds:
> On 5/17/22 06:29, Hugo van der Sanden wrote: I don't see anything in the comments at the top of |embed.fnc|, nor in perlguts "Internal Functions" that explains what "C" means;
> Lines 84-96 in embed.fnc explain the C flag

Oops, so sorry - my repo was still on 5.30-ish after a recent bisect.

hvds (approved):

khwilliamson:
A related issue is re_compile().  The comments say it is "public-ish" but the second parameter, the flags that stem from /i /s, etc aren't documented, and there are zero cpan uses.  That tells me it is unusable publicly as-is.  We could write a function that given the flag names returns the proper bits and make that public, or we could remove this from public consumption.  I don't like leaving it in limbo, though

hvds:
> A related issue is re_compile(). The comments say it is "public-ish" but the second parameter, the flags that stem from /i /s, etc aren't documented, and there are zero cpan uses. That tells me it is unusable publicly as-is. We could write a function that given the flag names returns the proper bits and make that public, or we could remove this from public consumption. I don't like leaving it in limbo, though

After finding background for this in 40b6423cb7 and #12301, I checked the code in https://github.com/dgl/re-engine-RE2 and it does indeed still call `Perl_re_compile` - are you perhaps excluding `*.cc` in your CPAN greps?

In that code, it's simply passing on the flags it was called with (via `(regexp_engine)my_engine.comp`) if it gets something it can't compile itself. The flags for that call are documented in `perlreapi` (but I haven't checked how up-to-date they are). I think we could document the flags for `re_compile` simply with a reference to that document, I don't see any obvious use for a function that translates flag names to bits.

khwilliamson:
I was looking only in *.xs, as otherwise there are too many false positives from ppport.h.  I tidn't know about .cc

But in this case, I can documentthe flag bits.

PR #19739, opened 2022-05-18
perlapi: Clean up some AV documentation

khwilliamson:
The consolidation of newAV with newAV_alloc_xz? included nonsensical
juxtapositions.  This cleans that up, adds a bit of markup, moves the
function implementing the newAV_alloc ones to internal, as there are no
CPAN uses, and we prefer the macro interface.

### DIFF
--- a/av.h
+++ b/av.h
@@ -112,64 +112,94 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
 =for apidoc_item newAV_alloc_x
 =for apidoc_item newAV_alloc_xz
 
-These all create a new AV, setting the reference count to 1. They differ
-in the allocation and population of the array of SV*s that always
-accompanies a non-empty AV.
+These all create a new AV, setting the reference count to 1.
 
-The Perl equivalent is approximately C<my @array;>.
+As background, an array consists of three things:
 
-newAV does not allocate a SV* array.
-    AV *av = newAV();
+=over
 
-This is very useful when an AV is required, but populating it may be
-deferred, or even never actually take place. (Memory is not allocated
-unnecessarily.)
+=item 1.
 
-Subsequent SV* array allocation would be performed via C<L</av_extend>>.
-This might be called directly:
-    av_extend(av, key);
+A data structure containing information about the array as a whole, such as its
+size and reference count.
 
-Or it might be called implicitly when the first element is stored:
+=item 2.
+
+A C language array of pointers to the individual elements.  These are treated
+as pointers to SVs, so all must be castable to SV*.
+
+=item 3.
+
+The individual elements themselves.  These could be, for instance, SVs and/or
+AVs and/or HVs, etc.
+
+=back
+
+An empty array need only have the first data structure, and all these functions
+create that.  They differ in what else they do, as follows:
+
+=over
+
+=item C<newAV> form
+
+=for comment
+'form' above and below is because otherwise have two =items with the same name,
+can't link to them.
+
+This does nothing beyond creating the whole-array data structure.
+The Perl equivalent is approximately S<C<my @array;>>
+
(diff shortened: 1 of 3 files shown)


av.h
```diff
@@ -112,64 +112,96 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+C<size> must be at least 1.
+
+=back
 
 The following examples all result in an array that can fit four elements
 (indexes 0 .. 3):
 
-    AV *av = newAV();
-    av_extend(av, 1);
-
```
richardleach, line 159: This example was intended to point out, albeit opaquely, that av_extend() forces a minimum pointer array size. I don't hugely object to this change though.

av.h
```diff
@@ -112,64 +112,96 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
 
-Or it might be called implicitly when the first element is stored:
     (void)av_store(av, 0, sv);
 
 Unused array elements are typically initialized by C<av_extend>. (Only
 core maintainers should have need to concern themseleves with when that
-is not the case. Refer to F<av.h> and F<av.c> for the differences between
+is not the case.
+
+Refer to F<av.h> and F<av.c> for the differences between
```
richardleach: I'm not sure that having this spacing helps. If the reader doesn't know what the differences between real and fake AVs are, they might now wonder if the instruction to "Refer to F<av.h>..." is a separate topic to that of pointer initialization.

av.h
```diff
@@ -112,64 +112,96 @@ If all you need is to look up an array element, then prefer C<av_fetch>.
...
+C<size> must be at least 1.
+
+=item C<newAV_alloc_xz> form
 
-For that scenario, newAV_alloc_x and newAV_alloc_xz can be used to create
-an AV and allocate a SV* array to fit the specified number of elements.
-(As a result, these macros MUST NOT be called with a size less than 1.)
+This is C<newAV_alloc_x>, but initializes each pointer in it to NULL.  This
+gives added safety to guard against them being read before being set.  If you
+know now what those elements should be, instead use L</C<av_make>>.
```
richardleach, line 185: The advice about using `av_make` could apply to more than just this form. Would it be better as a trailing "see also" statement?

richardleach:
This all looks good to me. Some suggestions above, but they are all optional.

av.h
khwilliamson: It was too opaque in my opinion, and led to questions about how much do you get by default, and sort of committed us to never lowering the default below 4.  That's why I omitted it.  Language explicitly stating there is a default minimum might be helpful.  I don't know

av.h
khwilliamson: I don't understand your point.  This is the only case where initialization is desired.

av.h
khwilliamson: Ok, I removed the spacing.   I didn't understand this point.  Do we need to even have mention in the API about things that don't apply to most people?  Could this be in a regular comment somewhere, like after the =cut.  I don't know, and am deferring to you

av.h
richardleach: I just had in mind that if the reader knows the elements, they might just want to go straight to `av_make`, rather than create the AV and then separately populate it with e.g. `av_push`. Otherwise, I agree that from an initialization standpoint, it makes sense where you've put it.

av.h
richardleach: Yes, I think the "Refer to" sentence could be a regular comment. It doesn't seem necessary for it to be in the apidoc, as it could unnecessarily confuse, but I like having it somewhere for newbie porters (like me) to follow if they want to learn the subtleties of AVs in core.

av.h
khwilliamson: ok

av.h
khwilliamson: done

PR #19743, opened 2022-05-19
Document clone_params_(del|new)

khwilliamson:


Leont:
IMHO, only modules implementing threads should need such a thing, and the fact that some other code on CPAN uses it means we have leaky abstractions elsewhere (in particular, the `CLONE` method doesn't get `CLONE_PARAMS*` as parameter).

And TBH, threading modules need enough knowledge of the internals of CLONE_PARAMS that this function is not terribly useful.

I'm not sure this interface is good enough to be API.

sv.c
```diff
@@ -16030,6 +16039,24 @@ Perl_clone_params_del(CLONE_PARAMS *param)
     }
 }
 
+/*
+=for apidoc clone_params_new
+
+Allocate and initialize a C<CLONE_PARAMS> structure for the
+S<C<PerlInterpreter *>> parameter C<to> based on data contained in the C<from>
+S<C<PerlInterpreter *>> parameter, returning a pointer to the new structure.
```
Leont, line 16371: I'm not sure this is right. And even if it is, I'm not sure this should be API because it's surprisingly unhelpful.

khwilliamson: If sv_dup is made API, and this isn't, how to resolve the fact that sv_dup has a parameter that may not be NULL that one of these fills, and the other frees?   I see on cpan some uses that fill that structure manually, which is diving even more into the internals.

Leont: > If sv_dup is made API, and this isn't, how to resolve the fact that sv_dup has a parameter that may not be NULL that one of these fills, and the other frees? I see on cpan some uses that fill that structure manually, which is diving even more into the internals.

And they're buggy for doing that.

`svt_dup` gets a `CLONE_PARAMS*` argument, but somehow `CLONE` doesn't. Recreating a new one for an existing clone session can easily lead to wrong results. I have a patch for that on my todo list. Nothing should be using these functions except thread modules (`threads`, `Thread::Csp`).

jkeenan:
There doesn't seem to be enough consensus to warrant merging this p.r.  It now has merge conflicts.  Discussion petered out more than a year ago.  Can we close this ticket?

jkeenan:
> There doesn't seem to be enough consensus to warrant merging this p.r. It now has merge conflicts. Discussion petered out more than a year ago. Can we close this ticket?

No one has spoken up for keeping this p.r. open.  Closing.

PR #19744, opened 2022-05-19
perlapi: Document re_compile

khwilliamson:


regcomp.c
```diff
@@ -6938,7 +6938,22 @@ Perl_pregcomp(pTHX_ SV * const pattern, const U32 flags)
...
 #endif
 
-/* public(ish) entry point for the perl core's own regex compiling code.
+/*
+=for apidoc re_compile
+
+Compile the regular expression pattern C<pattern>, returning a pointer to the
+compiled object for later matching with the internal regex engine.
+
+If C<pattern> is already a C<REGEXP>, this function, does nothing but return a
```
hvds, line 6947: Second comma is spurious.

regcomp.c
```diff
@@ -6938,7 +6938,22 @@ Perl_pregcomp(pTHX_ SV * const pattern, const U32 flags)
...
+pointer to the input.  Otherwise the PV is extracted and treated like a string
+representing a pattern.  See L<perlre>.
+
+The possible flags for C<rx_flags> are documented in L<perlreapi>.  Their names
+all begin with C<RXf_>.
+
+=cut
+
+ * public(ish) entry point for the perl core's own regex compiling code.
```
hvds, line 6956: I think we can probably lose the '(ish)' now.

hvds (approved):
LGTM; I'm tempted also to add something like:

This function is typically used by a custom regexp engine C\<.comp()\> function to hand off to the core regexp engine those patterns it doesn't want to handle itself (typically passing through the same flags it was called with). In almost all other cases, a regexp should be compiled by calling L\<pregcomp\> to compile using the currently active regexp engine.

khwilliamson:
changes as @hvds suggested

PR #19755, opened 2022-05-20
perlapi: Document hv_ksplit

khwilliamson:


hv.c
```diff
@@ -1679,6 +1679,15 @@ S_hsplit(pTHX_ HV *hv, STRLEN const oldsize, STRLEN newsize)
     } while (i++ < oldsize);
 }
 
+/*
+=for apidoc hv_ksplit
+
+Attempt to grow the hash C<hv> so it has at least C<newmax> entries available.
```
Leont: I think this is about buckets not entries? And that is begging the question "should it be public API", because that's really internal to how hashes work.

khwilliamson: I originally wanted to say buckets, but it looked like they weren't equivalent, but your comment got me to dig deeper, and you are right.

There are 15 + Devel::PPPort distros that use this function, so I thought it should be considered API

khwilliamson: I meant there are 15 distros + PPPort.

Another option is to mark it as experimental

Leont: I suppose that means the cat is out of the bag already, so I guess we should just accept that.

khwilliamson: Is there some phrasing we can add to tell them to try some other options first?

tonycoz: I think it's a reasonable function to call, it's no different to doing:
```
keys %hv = newmax;
```
which is implemented by calling hv_ksplit() (see Perl_magic_setnkeys() in mg.c)

It might be worthwhile including that equivalence in the hv_ksplit() documentation.

khwilliamson: suggestion pushed

Leont (approved):

PR #19762, opened 2022-05-20
perlapi: Document start_subparse

khwilliamson:


iabyn:
is_format means that a perl format is being parsed (i.e. a
    format NAME =
    FORMLIST
    .
thinngy), since internally that is compiled into a sub

khwilliamson:
I'm wondering if you were looking at the old wording, which after research I had revised to 

> f C<is_format> is non-zero, the input is to be considered a picture format;
else a C<sub>.

I took the name from Programming Perl. 

Please suggest better wording if this isn't good enough

khwilliamson:
That should have been

`If C<is_format> is non-zero, the input is to be considered a picture format;
else a C<sub>.`

iabyn:
On Mon, May 23, 2022 at 12:53:10PM -0700, Karl Williamson wrote:
> I'm wondering if you were looking at the old wording

I just saw the commend that said words to the effect that "I'm not sure what
this is".

> That should have been
> 
> `If C<is_format> is non-zero, the input is to be considered a picture format;
> else a C<sub>.`

I haven't come across that terminology before: perlform talks about
'picture lines' within a format, but not 'picture format'. How about


`If C<is_format> is non-zero, the input is to be considered a format sub
(a specialised sub used to implement perl's C<format> feature); else a
normal C<sub>.`


-- 
That he said that that that that is is is debatable, is debatable.

khwilliamson:
Accepted

PR #19769, opened 2022-05-23
Data-Dumper: dump booleans as booleans

haarg:
Use the support for tracking booleans to dump boolean values as !!1 and
!!0.

### DIFF
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -1448,7 +1448,7 @@ modify it under the same terms as Perl itself.
 
 =head1 VERSION
 
-Version 2.184
+Version 2.186
 
 =head1 SEE ALSO
 
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -18,6 +18,7 @@ use 5.008_001;
 require Exporter;
 
 use constant IS_PRE_516_PERL => $] < 5.016;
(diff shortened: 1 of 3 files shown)


rwp0:
Good idea, but  can't we name them more clearly simply as `true` and `false`?

`!!1` and `!!0` isn't nice even for dumping IMHO

haarg:
An advantage of `!!1` and `!!0` is that they are compatible with older perl versions. Newer perls do include `builtin::true` and `builtin::false`. They currently trigger warnings, so if they were used it would probably require delaying making this change until the builtin features are marked stable.

leonerd:
To be honest, even when `builtin::true` et.al. are considered stable, I'd still suggest dumping these as `!!0` and `!!1` by default, for compatibility with older perls, which will still be able to load this output. Maybe by an option users can request a more human-readable output at the expensive of that compat.

khwilliamson:
Is there a reason to delay applying this?

tonycoz (approved):

nicomen:
Given AFAIK Dumper produces a string, and you want to let it output !!0 and !!1 as false and true for eval-ing a new output in an older perl, would it make sense to output `!!0 # false` and `!!1 # true` when `$Data::Dumper::Indent` > 1 for readability?

rwp0:
> Maybe by an option users can request a more human-readable output at the expensive of that compat.

Please let us have such an option, since there is already many Dumper options this could fit well.

PR #19785, opened 2022-05-29
Deprecate 5 functions which are trivial accessors for public variables

nwc10:
These functions are obsolete, and stemmed from a long-solved shared library linkage problem in Windows.

Remove the use of two in Opcode.xs. This leaves a couple of uses on cpan of a couple of them. But people should just use PL_opnames[], for example, instead of the function here that returns the address of that array.


khwilliamson (approved):

nwc10:
I noticed that I needed to ammend the commit `opset_len in Opcode.xs is a constant, so should be a static const` to remove many now-unneeded `dMY_CXT` declarations. The C compiler is (usefully) warning about the unused variable declaration.

Issue #19780, opened 2022-05-28
ext/XS-APItest/t/boolean*.t: new test failures with PERLIO=stdio


jkeenan:
I am observing new test failures in `ext/XS-APItest/t/boolean-thr.t` and `ext/XS-APItest/t/boolean.t` when those program are run through the harness with `PERLIO=stdio`.

As seen on FreeBSD-13, threaded build, at v5.37.0-49-ge1980ad278:
```
$ cd t;TEST_JOBS=1 PERLIO=stdio ./perl harness -v ../ext/XS-APItest/t/boolean*.t;
cd -
../ext/XS-APItest/t/boolean-thr.t .. 
ok 1 - value in to thread is bool
ok 2 - value out of thread is bool
ok 3 - variable captured by thread is bool
ok 4 - :shared variable is bool outside
ok 5 - :shared variable is bool inside thread
# ok: SvIOK(sviv) && !SvIandPOK(sviv)
# ok: SvPOK(svpv) && !SvIandPOK(svpv)
# ok: SvIOK(sviv) && !SvBoolFlagsOK(sviv)
# ok: SvPOK(svpv) && !SvBoolFlagsOK(svpv)
# Testing 'newSVsv(&PL_sv_yes) and newSVsv(&PL_sv_no)'
# ok: SvCUR(true_sv) == 1
# ok: SvCUR(false_sv) == 0
# ok: SvLEN(true_sv) == 0
# ok: SvLEN(false_sv) == 0
# ok: SvIV(true_sv) == 1
# ok: SvIV(false_sv) == 0
# ok: SvIsCOW(true_sv)
# ok: SvIsCOW(false_sv)
# ok: strEQ(SvPV_nolen(true_sv),"1")
# ok: strEQ(SvPV_nolen(false_sv),"")
# ok: SvIOK(true_sv)
# ok: SvIOK(false_sv)
# ok: SvPOK(true_sv)
# ok: SvPOK(false_sv)
# ok: SvBoolFlagsOK(true_sv)
# ok: SvBoolFlagsOK(false_sv)
# ok: SvTYPE(true_sv) >= SVt_PVNV
# ok: SvTYPE(false_sv) >= SVt_PVNV
# ok: SvBoolFlagsOK(true_sv) && BOOL_INTERNALS_sv_isbool(true_sv)
# ok: SvBoolFlagsOK(false_sv) && BOOL_INTERNALS_sv_isbool(false_sv)
[...]

PR #19793, opened 2022-05-30
APItest.xs: flush output in test_bool_internals_func

Leont:
With the explicit flush this test will fail when PERLIO=stdio, because stdio defaults to block-based buffering on pipes and therefore the output will come out wrong.


Tux:
leont++ ; # immediate action and fix! (plus explain)

Tux (approved):

jkeenan:
This fixes https://github.com/Perl/perl5/issues/19780. @Leont, thanks!

PR #19795, opened 2022-05-31
XSLoader: convert from Test::More to internal test helpers

haarg:
The XSLoader tests need to play with the XS bits of various modules,
which can interfere with testing modules like Test::More. For example,
Test::More now loads Time::HiRes. This results in redefinition warnings,
and could lead to more serious problems.

Avoid this by creating some test helpers inside the XSLoader test, and
using those rather than Test::More.

The helpers implemented include roughly the same features used by the
test itself, so that the impact on the rest of the test code is minimal.

### DIFF
--- a/dist/XSLoader/t/XSLoader.t
+++ b/dist/XSLoader/t/XSLoader.t
@@ -7,11 +7,6 @@ use Config;
 
 my $db_file;
 BEGIN {
-    if (not eval "use Test::More; 1") {
-        print "1..0 # Skip: Test::More not available\n";
-        die "Test::More not available\n";
-    }
-
     use Config;
     foreach (qw/SDBM_File GDBM_File ODBM_File NDBM_File DB_File/) {
         if ($Config{extensions} =~ /\b$_\b/) {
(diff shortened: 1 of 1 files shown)


haarg:
There may be better approaches to this issue. The helpers are enough to cover what is in the test script, but end up being a rather large amount of code compared to the rest of the file.

dist/XSLoader/t/XSLoader.t
```diff
@@ -21,6 +16,103 @@ BEGIN {
...
+    print STDOUT "ok $tests";
+    print STDOUT " - $name"
+        if defined $name;
+    print "\n";
+    return $ok;
+}
+sub is ($$;$) {
+    my ($got, $want, $name) = @_;
+
+    my $ok
```
wolfsage, line 40: This makes my head hurt but totally works

wolfsage (approved):
Nice, thank you

dist/XSLoader/t/XSLoader.t
haarg: If you have a suggestion for an improvement, I'd welcome it. All the negatives do hurt the comprehensibility.

wolfsage: Maybe this: Could be even clearer with more ifs etc...

```
   my $different = (defined $want xor defined $got)
                 || (defined $got && $got ne $want);
    my $ok = ! $different;
```

dist/XSLoader/t/XSLoader.t
```diff
@@ -21,6 +16,102 @@ BEGIN {
...
+    my ($message) = @_;
+    $message =~ s/\n?\z/\n/;
+    $message =~ s/^/# /gm;
+    print STDERR $message;
+}
+END {
+    if (!defined $planned_tests) {
+        $planned_tests = $tests;
+        print STDOUT "1..$planned_tests\n";
+    }
```
tonycoz: This is a little dangerous. If the file is changed to not use a plan, which appears to be the intent of this code, this would allow the file to pass even if one of the tests in it does an early exit.

I think this should die if there's no plan set.

If we want to support no-plan type testing, the code above should die and we add:
```
sub done_testing {
    $planned_tests = $tests;
}
```
which ensures execution reaches the call to done_testing() and behaves like no-plan testing in modern Test::More.

haarg: I'd agree that no_plan style testing is dangerous. Since this test doesn't need it, I've just removed this and replaced it with an error.

dist/XSLoader/t/XSLoader.t
haarg: I didn't love your proposal, but I've replaced this with something that should be clearer.

wolfsage: Yes yours is great, thanks

tonycoz (approved):

PR #19797, opened 2022-05-31
Fix building with clang-cl

clemenswasser:
As mentioned in https://lists.llvm.org/pipermail/llvm-dev/2015-July/088122.html
and https://github.com/llvm/llvm-project/issues/24625 building with clang-cl currently fails due to the declaration of __PL_nan_u.
By declaring it like this, cl and clang-cl are happy to parse it.

### DIFF
--- a/win32/win32.h
+++ b/win32/win32.h
@@ -284,8 +284,8 @@ MSVC_DIAG_RESTORE
    importing __PL_nan_u across DLL boundaries in size in the importing DLL
    will be more than the 8 bytes it will take up being in each XS DLL if
    that DLL actually uses __PL_nan_u */
-extern const __declspec(selectany) union { unsigned __int64 __q; double __d; }
-__PL_nan_u = { 0x7FF8000000000000UI64 };
+union PerlNan { unsigned __int64 __q; double __d; };
+extern const __declspec(selectany) union PerlNan __PL_nan_u = { 0x7FF8000000000000UI64 };
 #define NV_NAN ((NV)__PL_nan_u.__d)
 
 /* The CRT was rewritten in VS2015. */
--- a/.mailmap
+++ b/.mailmap
@@ -408,6 +408,7 @@ Claes Jacobsson <claes@surfar.nu> Claes Jacobsson <claes@surfar.nu>
 Claes Jacobsson <claes@surfar.nu> Claes Jakobsson <claes@surfar.nu>
 Claes Jacobsson <claes@surfar.nu> Claes Jakobsson <claes@versed.se>
 Claudio Ramirez <nxadm@cpan.org> Claudio Ramirez <nxadm@cpan.org>
+Clemens Wasser <clemens.wasser@gmail.com> Clemens Wasser <clemens.wasser@gmail.com>
 Clinton A. Pierce <clintp@geeksalad.org> Clinton A. Pierce <clintp@geeksalad.org>
 Clinton A. Pierce <clintp@geeksalad.org> Clinton Pierce <cpierce1@ford.com>
 Clinton Gormley <unknown> Clinton Gormley <unknown>
--- a/AUTHORS
+++ b/AUTHORS
@@ -283,6 +283,7 @@ Chunhui Teng                   <cteng@nortel.ca>
 Claes Jacobsson                <claes@surfar.nu>
 Clark Cooper                   <coopercc@netheaven.com>
 Claudio Ramirez                <nxadm@cpan.org>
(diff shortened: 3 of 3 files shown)


jkeenan:
> As mentioned in https://lists.llvm.org/pipermail/llvm-dev/2015-July/088122.html and [llvm/llvm-project#24625](https://github.com/llvm/llvm-project/issues/24625) building with clang-cl currently fails due to the declaration of __PL_nan_u. By declaring it like this, cl and clang-cl are happy to parse it.

**List:** Do we have someone who is familiar with use of clang-cl on Windows who could evaluate this p.r.?

@clemenswasser, that issue in the LLVM project appears to have been closed before you submitted this p.r.  Can you explain the status of that issue?

clemenswasser:
> @clemenswasser, that issue in the LLVM project appears to have been closed before you submitted this p.r. Can you explain the status of that issue?

The Issue was closed as `WONTFIX`, because they concluded, that they don't want to change clang.
Here the relevant comment https://github.com/llvm/llvm-project/issues/24625#issuecomment-980948611:
> Hacking around this in clang is not the right solution because it involves calculating linkage incorrectly in a lot of places which erodes some of clang's inherent value. The problem solves itself if the perl people change their union to have a class-name.

clemenswasser:
@jkeenan I'm really confused about the build failure.
The exact same MSVC compiler (29.30145) compiles it without errors: https://godbolt.org/z/17EMWPMa4

win32/win32.h
```diff
@@ -284,8 +284,8 @@ MSVC_DIAG_RESTORE
    importing __PL_nan_u across DLL boundaries in size in the importing DLL
    will be more than the 8 bytes it will take up being in each XS DLL if
    that DLL actually uses __PL_nan_u */
-extern const __declspec(selectany) union { unsigned __int64 __q; double __d; }
-__PL_nan_u = { 0x7FF8000000000000UI64 };
+union U { unsigned __int64 __q; double __d; };
+extern const __declspec(selectany) U __PL_nan_u = { 0x7FF8000000000000UI64 };
```
tonycoz: This looks like a reasonable solution based on the discussion, but I think it could use a more unique class name than "U", maybe "PerlNan".

tonycoz (changes requested):

win32/win32.h
clemenswasser: Yes, for sure. I wanted to change the name, but forgot about it. Thanks for reminding me 😅

tonycoz (approved):

tonycoz:
Did you test build this with Visual Studio, or the command-line tools?

clemenswasser:
This shouldn't make any difference. This compiles for me locally with MSVC and clang-cl. The godbolt link above also confirms that this compiles with MSVC (I've even selected the same version as is used in your CI). I'll try running the CI build locally and see if I can fix this.

clemenswasser:
Found and fixed the issue. The problem was a missing `union` before the type of `__PL_nan_u`, which is required in C but not in C++.

clemenswasser:
@tonycoz or should I declare `PerlNan` with `typedef union`? Which method is used more across the perl codebase?

tonycoz:
> This shouldn't make any difference. This compiles for me locally with MSVC and clang-cl. The godbolt link above also confirms that this compiles with MSVC (I've even selected the same version as is used in your CI). I'll try running the CI build locally and see if I can fix this.

I was thinking in terms of doing a full build, I had linker problems using:
```
nmake  CCTYPE=MSVC142 CC=clang-ci LINK32=lld-link
```
I haven't tried hard to diagnose it.

> should I declare PerlNan with typedef union? Which method is used more across the perl codebase?

This type is used only for this purpose, I think it can just be `union PerlNan`.

clemenswasser:
> I was thinking in terms of doing a full build, I had linker problems using:

Oh, I probably should have stated this more clearly 😅. This only fixes the parsing of the Perl headers for **external code** when using clang-cl. So I haven't actually tried building Perl with clang-cl, this PR is only about the use of the headers in other code.

khwilliamson:
failing on mac

clemenswasser:
> failing on mac

How is this possible? I'm only changing `win32.h`.

tonycoz:
> How is this possible? I'm only changing `win32.h`.

It's a timing test which is likely unrelated, I've re-run the CI job.

tonycoz:
> I've re-run the CI job.

Which passed, I've merged into blead.

PR #19802, opened 2022-05-31
feature - document that use utf8 should probably be used for non-ascii delimiters

Grinnz:
and actually it seems required: the parser doesn't recognize \xAB and \xBB as a pair though these technically should be interpreted as the correct characters (U+00AB/U+00BB) without use utf8. but this is not necessary to document and may or may not be considered a bug

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -5,7 +5,7 @@
 
 package feature;
 
-our $VERSION = '1.73';
+our $VERSION = '1.74';
 
 our %feature = (
     fc                      => 'feature_fc',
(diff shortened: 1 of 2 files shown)


khwilliamson:
This isn't true.  These two characters work fine without a 'use utf8'.  There are tests in the suite just for them, and I tried it again just now.

But all three of the following must be true:
The file has to be encoded as bytes (my editor defaults to UTF-8)
The pattern must be compiled within the scope of 'unicode_strings' or have some indication that Unicode mode is used, such as a /u modifier or a \b{gcb} element.
The pattern must be compiled within the scope of 'extra_paired_delimiters'

Grinnz:
Sorry you're correct, I made a stupid mistake in my test! But it doesn't affect the doc patch, anyway.

khwilliamson:
I guess I don't understand why.  If the file isn't UTF-8, you shouldn't say 'use utf8'

Grinnz:
I am fairly confident every usage of this feature will be in a UTF-8 encoded file, in which case the most common failure mode would be missing "use utf8" because it is not mentioned as necessary.

Grinnz:
Updated to specifically mention its use for UTF-8-encoded source files

PR #19803, opened 2022-05-31
Remove support for Ultrix

khwilliamson:
Ultrix has been removed.  Ultrix was the native Unix-like operating
system for various Digital Equipment Corporation machines.  Its final
release was in 1995.


jkeenan:
The only part of this p.r. about which I have any doubts are the inline comments in [locale.c](https://github.com/Perl/perl5/pull/19803/files#diff-f7d64c37bffc437cd59ec53833a08ce5adec3e7b03436c7ed8cb11c789b26ac0).  Could we get some additional eyeballs on that diff?

khwilliamson:
Given no responses, merged

PR #19805, opened 2022-06-01
ci: add build for Windows on Arm

p-b-o:
Since Github Actions does not offer environment for Windows on Arm, we
can only compile this target, and not run any tests.

Since perl build uses some binaries, we must build in two steps.
First, create those binaries (x64), and then compile perl itself for arm64.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -375,6 +375,40 @@ jobs:
           cd win32
           nmake CCTYPE=MSVC142 test
 
+  #
+  #          _           _                                               __   _  _
+  # __      _(_)_ __   __| | _____      _____        __ _ _ __ _ __ ___  / /_ | || |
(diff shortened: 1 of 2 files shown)


p-b-o:
Please find a run (only for this new job) [here](https://github.com/pbo-linaro/perl5/runs/6685489262). You can see on this [line](https://github.com/pbo-linaro/perl5/runs/6685489262#step:4:1814) that compiled binary is arm64.

p-b-o:
Polite ping :) @xenu @nwc10

Leont:
What will this test in practice? Can it actually fail in ways that the existing Windows CI doesn't?

p-b-o:
This checks builds work under this platform. Alas, since there is no runner available today for Windows on Arm, it's not possible to run tests.

However, checking the build is a first step and we can ensure no regression is added to build system.

khwilliamson:
@leont, are you satisfied with the response

PR #19814, opened 2022-06-02
(Optimization:Regexec) Replaced recppush/regcppop with memcpy

AnFunctionArray:
Improved performance of complex recursive regular expressions with a lot of capture groups.

Determined bottleneck with Intel VTune and fixed by using faster built-in memcpy instead of a loop.

Currently in my test cases performance seems to be 2x.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -218,7 +218,6 @@ static void S_setup_eval_state(pTHX_ regmatch_info *const reginfo);
 static void S_cleanup_regmatch_info_aux(pTHX_ void *arg);
 static regmatch_state * S_push_slab(pTHX);
 
-#define REGCP_PAREN_ELEMS 3
 #define REGCP_OTHER_ELEMS 3
 #define REGCP_FRAME_ELEMS 1
 /* REGCP_FRAME_ELEMS are not part of the REGCP_OTHER_ELEMS and
@@ -228,8 +227,11 @@ STATIC CHECKPOINT
 S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
 {
     const int retval = PL_savestack_ix;
-    const int paren_elems_to_push =
-                (maxopenparen - parenfloor) * REGCP_PAREN_ELEMS;
+    /* Number of bytes about to be stored in the stack */
+    const SSize_t paren_bytes_to_push = sizeof(*rex->offs) * (maxopenparen - parenfloor);
+    /* Number of savestack[] entries to be filled by the paren data */
+    /* Rounding is performed in case we are few elements short */
+    const int paren_elems_to_push = (paren_bytes_to_push + sizeof(*PL_savestack) - 1) / sizeof(*PL_savestack);
     const UV total_elems = paren_elems_to_push + REGCP_OTHER_ELEMS;
     const UV elems_shifted = total_elems << SAVE_TIGHT_SHIFT;
     I32 p;
@@ -238,9 +240,9 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
     PERL_ARGS_ASSERT_REGCPPUSH;
 
     if (paren_elems_to_push < 0)
-        Perl_croak(aTHX_ "panic: paren_elems_to_push, %i < 0, maxopenparen: %i parenfloor: %i REGCP_PAREN_ELEMS: %u",
+        Perl_croak(aTHX_ "panic: paren_elems_to_push, %i < 0, maxopenparen: %i parenfloor: %i",
                    (int)paren_elems_to_push, (int)maxopenparen,
-                   (int)parenfloor, (unsigned)REGCP_PAREN_ELEMS);
+                   (int)parenfloor);
 
     if ((elems_shifted >> SAVE_TIGHT_SHIFT) != total_elems)
         Perl_croak(aTHX_ "panic: paren_elems_to_push offset %" UVuf
@@ -249,31 +251,35 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
                    (unsigned long)maxopenparen,
                    (long)parenfloor);
 
-    SSGROW(total_elems + REGCP_FRAME_ELEMS);
-
     DEBUG_BUFFERS_r(
         if ((int)maxopenparen > (int)parenfloor)
-            Perl_re_exec_indentf( aTHX_
+            Perl_re_exec_indentf(aTHX_
                 "rex=0x%" UVxf " offs=0x%" UVxf ": saving capture indices:\n",
                 depth,
                 PTR2UV(rex),
                 PTR2UV(rex->offs)
             );
     );
-    for (p = parenfloor+1; p <= (I32)maxopenparen;  p++) {
-/* REGCP_PARENS_ELEMS are pushed per pairs of parentheses. */
-        SSPUSHIV(rex->offs[p].end);
-        SSPUSHIV(rex->offs[p].start);
-        SSPUSHINT(rex->offs[p].start_tmp);
-        DEBUG_BUFFERS_r(Perl_re_exec_indentf( aTHX_
-            "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "\n",
-            depth,
-            (UV)p,
-            (IV)rex->offs[p].start,
-            (IV)rex->offs[p].start_tmp,
-            (IV)rex->offs[p].end
-        ));
-    }
+
+    SSGROW(total_elems + REGCP_FRAME_ELEMS);
+
+    /* memcpy the offs inside the stack - it's faster than for loop */
+    memcpy(&PL_savestack[PL_savestack_ix], rex->offs + parenfloor + 1, paren_bytes_to_push);
+    PL_savestack_ix += paren_elems_to_push;
+
+    DEBUG_BUFFERS_r(
+        for (p = parenfloor + 1; p <= (I32)maxopenparen; p++) {
+            Perl_re_exec_indentf(aTHX_
+                "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "\n",
+                depth,
+                (UV)p,
+                (IV)rex->offs[p].start,
+                (IV)rex->offs[p].start_tmp,
+                (IV)rex->offs[p].end
+            );
+        }
+    );
+
 /* REGCP_OTHER_ELEMS are pushed in any case, parentheses or no. */
     SSPUSHINT(maxopenparen);
     SSPUSHINT(rex->lastparen);
@@ -366,25 +372,36 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
                 PTR2UV(rex->offs)
             );
     );
-    paren = *maxopenparen_p;
-    for ( ; i > 0; i -= REGCP_PAREN_ELEMS) {
-        SSize_t tmps;
-        rex->offs[paren].start_tmp = SSPOPINT;
-        rex->offs[paren].start = SSPOPIV;
-        tmps = SSPOPIV;
-        if (paren <= rex->lastparen)
-            rex->offs[paren].end = tmps;
-        DEBUG_BUFFERS_r( Perl_re_exec_indentf( aTHX_
-            "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "%s\n",
-            depth,
-            (UV)paren,
-            (IV)rex->offs[paren].start,
-            (IV)rex->offs[paren].start_tmp,
-            (IV)rex->offs[paren].end,
-            (paren > rex->lastparen ? "(skipped)" : ""));
-        );
-        paren--;
-    }
+    /* substract remaining elements from the stack */
+    PL_savestack_ix -= i;
+
+    /* static assert that offs struc size is not less than stack elem size */
+    STATIC_ASSERT_STMT(sizeof(*rex->offs) >= sizeof(*PL_savestack));
+
+    /* calculate actual number of offs/capture groups stored */
+    /* by doing integer division (leaving potential alignment aside) */
+    i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
+
+    /* calculate paren starting point */
+    /* i is our number of entries which we are subtracting from *maxopenparen_p */
+    /* and we are storing + 1 this to get the beginning */
+    paren = *maxopenparen_p - i + 1;
+
+    /* restore them */
+    memcpy(rex->offs + paren, &PL_savestack[PL_savestack_ix], i * sizeof(*rex->offs));
+
+    DEBUG_BUFFERS_r(
+        for (; paren <= *maxopenparen_p; ++paren) {
+            Perl_re_exec_indentf(aTHX_
+                "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "%s\n",
+                depth,
+                (UV)paren,
+                (IV)rex->offs[paren].start,
+                (IV)rex->offs[paren].start_tmp,
+                (IV)rex->offs[paren].end,
+                (paren > rex->lastparen ? "(skipped)" : ""));
+        }
+    );
 #if 1
     /* It would seem that the similar code in regtry()
      * already takes care of this, and in fact it is in
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,6 +67,7 @@ Alexander D'Archangel          <darksuji@gmail.com>
 Alexander Foken
 Alexander Gernler              <alexander_gernler@genua.de>
 Alexander Gough                <alex-p5p@earth.li>
+Alexander Nikolov              <sasho648@gmail.com>
 Alexander Hartmaier            <abraxxa@cpan.org>
 Alexander Klimov               <ask@wisdom.weizmann.ac.il>
 Alexander Smishlajev           <als@turnhere.com>


AnFunctionArray:
@khwilliamson Does that mean the tests failing are for author only - I thought there were alignment issues in 64 bit.

That's why I closed this by myself. Anyway I'll be happy if this actually gets accepted.

regexec.c
```diff
@@ -260,20 +259,8 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
                 PTR2UV(rex->offs)
             );
     );
-    for (p = parenfloor+1; p <= (I32)maxopenparen;  p++) {
-/* REGCP_PARENS_ELEMS are pushed per pairs of parentheses. */
-        SSPUSHIV(rex->offs[p].end);
-        SSPUSHIV(rex->offs[p].start);
-        SSPUSHINT(rex->offs[p].start_tmp);
-        DEBUG_BUFFERS_r(Perl_re_exec_indentf( aTHX_
```
hvds, line 268: I think it would be wise to retain the debug info, here and below: this will be zero-cost on a non-debugging perl and minimal cost on a debugging perl when the relevant debug flag is not set. Simplest would be to move the loop inside the preceding `DEBUG_BUFFERS_r(...)`:
```
--- a/regexec.c
+++ b/regexec.c
@@ -258,6 +258,15 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
                 PTR2UV(rex),
                 PTR2UV(rex->offs)
             );
+        for (p = parenfloor+1; p <= (I32)maxopenparen;  p++) {
+            Perl_re_exec_indentf( aTHX_
+                "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "\n",
+                depth, (UV)p,
+                (IV)rex->offs[p].start,
+                (IV)rex->offs[p].start_tmp,
+                (IV)rex->offs[p].end
+            );
+        }
     );
     memcpy(&PL_savestack[PL_savestack_ix], rex->offs + parenfloor + 1, paren_elems_to_push * sizeof(*PL_savestack));
     PL_savestack_ix += paren_elems_to_push;
```

regexec.c
```diff
@@ -229,7 +228,7 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
 {
     const int retval = PL_savestack_ix;
     const int paren_elems_to_push =
-                (maxopenparen - parenfloor) * REGCP_PAREN_ELEMS;
+        (sizeof(*rex->offs) * (maxopenparen - parenfloor)) / sizeof(*PL_savestack);
     const UV total_elems = paren_elems_to_push + REGCP_OTHER_ELEMS;
```
hvds: You are mixing units in the calculation of `total_elems`, so the `SSGROW` call below will use an incorrect argument. As a result, the savestack will not be grown as much as is needed.

hvds (changes requested):
I like the general idea here, but more care is needed in the calculations - I think you need to retain the original calculation of `paren_elems_to_push`, and convert that to units of `PL_savestack` elements in a separate variable.

regexec.c
AnFunctionArray: I don't quite get that? - I think I solved potential alignment issues with my second patch.

https://github.com/Perl/perl5/pull/19814/commits/57e118d4dc8fb5ab5d2e1c1197d791b6ebba3b06

hvds: It is not an alignment issue, but a size issue (or units issue).

Previously, `paren_elems_to_push` and `total_elems` were both measured in units of `sizeof(IV)`. With this change, `paren_elems_to_push` is now measured in units of `sizeof(ANY)`. So `total_elems` is now a nonsense value calculated by adding two things measured in different units.

AnFunctionArray: Aren't they the same thing?

AnFunctionArray: I guess not but I measure the elements needed in the stack nothing more -  taking the stack element size.

AnFunctionArray: I guess if you mean in terms of code clarity and the later if for debug printing then I may just add another variable.

hvds: Looking again, I may be totally wrong (I think I misinterpreted the `SSGROW`), and that `paren_elems_to_push` and `total_elems` were always actually measured in units of `sizeof(ANY)`.

I'm looking further.

hvds: Ok, I was wrong, my apologies for the confusion.

It would probably be useful to add a comment something like:
```
    /* Number of savestack[] entries to be filled by the paren data */
    const int paren_elems_to_push = ...
```

hvds:
I think it would also be useful to reorder things slightly for clarity: in `regcppush` have the `SSGROW()` immediately next to the `memcpy` and `SSPUSHxxx` calls (and correspondingly for `regcppop`), and have the complex `DEBUG_BUFFERS_r()` chunk either before or after those.

regexec.c
AnFunctionArray: Added and reworked debugging prints.

AnFunctionArray:
Can I get tests run please - I think nothing brakes with this (and I hope the debug prints are alright although I believe there aren't tests for that)?

regexec.c
```diff
@@ -260,20 +259,8 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
...
-        DEBUG_BUFFERS_r(Perl_re_exec_indentf( aTHX_
-            "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "\n",
-            depth,
-            (UV)p,
-            (IV)rex->offs[p].start,
-            (IV)rex->offs[p].start_tmp,
-            (IV)rex->offs[p].end
-        ));
-    }
+    memcpy(&PL_savestack[PL_savestack_ix], rex->offs + parenfloor + 1, paren_elems_to_push * sizeof(*PL_savestack));
```
hvds: Oh, with the alignment fix I think this can read up to `sizeof(*PL_savestack) - 1` bytes beyond the end of `rex->offs[]`.

I think it should be:
```
  const int paren_bytes_to_push = sizeof(*rex->offs) * (maxopenparen - parenfloor);
  const int paren_elems_to_push = (paren_bytes_to_push + sizeof(*PL_savestack) - 1) / sizeof(*PL_savestack);
...
  memcpy(&PL_savestack[PL_savestack_ix], &rex->offs[parenfloor + 1], paren_bytes_to_push);
  PL_savestack_ix += paren_elems_to_push;
```

(It also seems more consistent to refer to `&rex->offs[parenfloor + 1]`.)

I assume we do not need guards against numeric overflow on those calculations, but maybe they should more correctly be Size_t or SSize_t rather than int.

regexec.c
```diff
@@ -248,19 +249,35 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
...
             );
     );
+
+    SSGROW(total_elems + REGCP_FRAME_ELEMS);
+
+    /* memcpy the offs inside the stack - it's faster than for loop */
     memcpy(&PL_savestack[PL_savestack_ix], rex->offs + parenfloor + 1, paren_elems_to_push * sizeof(*PL_savestack));
     PL_savestack_ix += paren_elems_to_push;
+
+    DEBUG_BUFFERS_r(
```
hvds, line 270: You don't need DEBUG_BUFFERS_r() inside DEBUG_BUFFERS_r(). The idea was also to combine this with the previous debug:
```
    DEBUG_BUFFERS_r(
        if ((I32)maxopenparen > (I32)parenfloor) {
            Perl_re_exec_indentf(aTHX_
                "rex=0x%" UVxf " offs=0x%" UVxf ": saving capture indices:\n",
                depth,
                PTR2UV(rex),
                PTR2UV(rex->offs)
            );
            for (p = parenfloor + 1; p <= (I32)maxopenparen; p++) {
                Perl_re_exec_indentf(aTHX_
                    "    \\%" UVuf ": %" IVdf "(%" IVdf ")..%" IVdf "\n",
                    depth,
                    (UV)p,
                    (IV)rex->offs[p].start,
                    (IV)rex->offs[p].start_tmp,
                    (IV)rex->offs[p].end
                );
            }
        }
    );
```

(I've also tweaked it to consistently cast to I32 here.)

regexec.c
```diff
@@ -366,25 +353,10 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
...
-            (IV)rex->offs[paren].start,
-            (IV)rex->offs[paren].start_tmp,
-            (IV)rex->offs[paren].end,
-            (paren > rex->lastparen ? "(skipped)" : ""));
-        );
-        paren--;
-    }
+    PL_savestack_ix -= i;
+    i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
+    paren = *maxopenparen_p - i;
```
hvds: Making this `paren = *maxopenparen_p - i + 1;` saves a little bit of effort later.

regexec.c
```diff
@@ -353,10 +370,31 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
                 PTR2UV(rex->offs)
             );
     );
+    /* substract remaining elements from the stack */
     PL_savestack_ix -= i;
+
+    /* calculate number of offs/capture groups stored */
     i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
```
hvds: This creates a problem the other way: if, for example, `sizeof(*rex->offs) = 32, sizeof(*PL_savestack) = 64`, then we could end up saving 7 sets of parens, but then trying to restore 8 sets.

It is possible that we could change the definition of the guard value to be `SAVEt_REGCONTEXT | (paren_count << SAVE_TIGHT_SHIFT)`, but I'm not sure what else might be relying on it - we'd at least need to change the handling at `scope.c:1492`, which would mean the code there would need to know about `sizeof(*rex->offs)`.

Another option would be to push an extra element just to store this count, but then this PR becomes a net loss when there are just 0 or 1 parens to push (which is likely the commonest case).

AnFunctionArray: I don't think this is a problem because of the way integer division works. Unless I'm missing something.

hvds: Using the same 32/64 sizes (sorry, those were intended to be bit sizes, I mean 4/8 bytes), if we had 7 sets of parens to save, we will have stored them in 4 savestack entries. `i = (4 * 8) / 4` at this point will then calculate that there are 8 sets of parens to restore.

AnFunctionArray: Well the original code was measuring the parens in N elements of the stack. Meaning the parens will always be greater or equal in size than a stack element. If you are to be that pedantic.

hvds: I'm sorry, I don't understand your comment. The original code would save 7 sets and restore 7 sets, which is fine. The new code would save 7 sets and attempt to restore 8 sets, which seems not fine.

AnFunctionArray: My comment was because it's pedantic in my case - but ok I guess if like offs have 8 bit elements for some reason

hvds: Ok, I did partially mess up again: I was taking `sizeof(*rex->offs)` as equal to `sizeof(SSize_t)`, when it is actually `3 * sizeof(SSize_t)`. The calculation can go wrong only if `sizeof(ANY)` can be greater than this, so if both ANY and SSize_t are sized as a power of 2, it would need ANY to be at least 4 times the size of SSize_t before it goes wrong - which seems somewhat unlikely, but not beyond the bounds of reason. (Anyone else know if that is possible?)

Beyond that is the potential for future problems, eg if `struct regexp_paren_pair` or `struct any` change in the future: there is nothing that would point the author of such a change to look at this code, it could simply start going wrong in nasty subtle ways.

So I think at the very least it needs a static assert that `sizeof(*rex->offs) >= sizeof(*PL_savestack)`. If someone else signs that off as sufficient, I could live with that (but if the assert ever does fail on some platform, I fear it will be hard to fix).

AnFunctionArray: Added static assert for this.

regexec.c
AnFunctionArray: Done

regexec.c
AnFunctionArray: You are right - I corrected that also.

regexec.c
AnFunctionArray: I remove the nested DEBUG_BUFFERS_r - I didn't combine them though because I think the first is informing about the push operation and the second reporting its status.

regexec.c
```diff
@@ -235,6 +235,10 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
     I32 p;
     DECLARE_AND_GET_RE_DEBUG_FLAGS;
 
+    /* static assert that offs struc size is not less than stack elem size */
+
+    sizeof(char[sizeof(*rex->offs) >= sizeof(*PL_savestack) ? 1 : -1]);
```
hvds: See `STATIC_ASSERT_DECL` and `STATIC_ASSERT_STMT` in `perl.h`

This should ideally be used at the point we rely on it for the calculation in `regcppop`, something like:
```
     /* calculate number of offs/capture groups stored */
+    STATIC_ASSERT_STMT(sizeof(*rex->offs) >= sizeof(*PL_savestack));
     i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
```

regexec.c
```diff
@@ -377,13 +381,13 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
     i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
 
     /* calculate paren starting point (-1) */
-    paren = *maxopenparen_p - i;
+    paren = *maxopenparen_p - i + 1;
```
hvds: The comment no longer needs '(-1)'.

hvds (commented):
When you've addressed the latest comments, it would probably be a good time to squash the commits and rebase.

regexec.c
hvds: This still needs to move to reccppop

regexec.c
```diff
@@ -366,25 +373,31 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
...
+
+    /* calculate paren starting point */
+    paren = *maxopenparen_p - i + 1;
+
+    /* restore them */
+    memcpy(rex->offs + paren, &PL_savestack[PL_savestack_ix], i * sizeof(*rex->offs));
+
+    DEBUG_BUFFERS_r(
+        for (; paren <= *maxopenparen_p; ++paren) {
+            DEBUG_BUFFERS_r(Perl_re_exec_indentf(aTHX_
```
hvds: I missed this - still double DEBUG_BUFFERS_r here.

regexec.c
hvds: Not sure if I marked this unresolved last time, but it still needs to move.

AnFunctionArray: It's a static assert I can move it if you like.

AnFunctionArray: @hvds Ok moved it as requested.

hvds (approved):
I remain nervous about the reliance on relative structure sizes, but I can live with this if someone else agrees that they're happy with it.

To recap: in `regcppush` this change packs the capture groups to save (`sizeof(*rex->offs)` each, currently equal to `3 * sizeof(SSize_t)`) into a calculated number of savestack slots (`sizeof(*PL_savestack)` each, currently `sizeof(ANY)`). In `regcppop` we then discover how many capture groups should be restored by doing the reverse calculation - which can get the wrong result if there was enough spare space in the final savestack slot to fit another one. I.e. it can go wrong if we ever have `sizeof(*PL_savestack) > sizeof(*rex->offs)`.

That risk is now managed by a static assert:
```
    STATIC_ASSERT_STMT(sizeof(*rex->offs) >= sizeof(*PL_savestack));
```
My remaining concern is that if the assert ever fails on some platform, we have no good way to fix it for them. @tonycoz, @iabyn, @leonerd, anyone want to weigh in for or against?

tonycoz:
I don't see why the static assert is needed, if we assume for argument that Size_t is 1 byte (making sizeof(regexp_paren_pair) == 3) and sizeof(*PL_savestack) is 8, then the calculation of `paren_elems_to_push` is still the correct 1.

I don't think the assertion can fail currently, though that would change if we added NV to `union any` (__float128 on 32-bit systems)

hvds:
> I don't see why the static assert is needed, if we assume for argument that Size_t is 1 byte (making sizeof(regexp_paren_pair) == 3) and sizeof(*PL_savestack) is 8, then the calculation of `paren_elems_to_push` is still the correct 1.
> 
> I don't think the assertion can fail currently, though that would change if we added NV to `union any` (__float128 on 32-bit systems)

@tonycoz The issue is with the reverse calculation in `recppop` (which is why I was at pains to have the static assert moved to the correct place):
```
    /* calculate number of offs/capture groups stored */
    i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
```

With your example sizes, this would calculate 2 as the number of paren groups to restore, regardless of whether 1 or 2 paren groups had been saved in `regcppush`.

AnFunctionArray:
If I need to do anything - please tell me - I would like this to be pushed forward personally.

hvds:
> If I need to do anything - please tell me - I would like this to be pushed forward personally.

I'm hoping that at some point @tonycoz will get a chance to look at my most recent response. If @iabyn gets a chance to look at the PR as well, that would be super.

tonycoz:
> > If I need to do anything - please tell me - I would like this to be pushed forward personally.
> 
> I'm hoping that at some point @tonycoz will get a chance to look at my most recent response. If @iabyn gets a chance to look at the PR as well, that would be super.

Sorry, you were so obviously correct I didn't think I needed to respond.

My comment about whether the assertion can fail still applies - only if we add NV or another similarly large type to `union any`.  Note that `perl.h` has provision to allow an external definition of union any, though I doubt it has seen any recent use.

iabyn:
On Thu, Jul 28, 2022 at 09:41:30AM -0700, Hugo van der Sanden wrote:
> I'm hoping that at some point @tonycoz will get a chance to look at my
> most recent response. If @iabyn gets a chance to look at the PR as well,
> that would be super.

It's still on my todo list.

-- 
Diplomacy is telling someone to go to hell in such a way that they'll
look forward to the trip

regexec.c
demerphq: just curious, can it be made into a compile time assert somehow?

hvds: > just curious, can it be made into a compile time assert somehow?

Yes, per my comment above: "See STATIC_ASSERT_DECL and STATIC_ASSERT_STMT in perl.h".

regexec.c
```diff
@@ -228,8 +227,9 @@ STATIC CHECKPOINT
 S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
 {
     const int retval = PL_savestack_ix;
-    const int paren_elems_to_push =
-                (maxopenparen - parenfloor) * REGCP_PAREN_ELEMS;
+    const SSize_t paren_bytes_to_push = sizeof(*rex->offs) * (maxopenparen - parenfloor);
+    /* Number of savestack[] entries to be filled by the paren data */
+    const int paren_elems_to_push = (paren_bytes_to_push + sizeof(*PL_savestack) - 1) / sizeof(*PL_savestack);
```
demerphq, line 234: I think the preceding comment here should note that this is rounding in disguise would be in order. I had to stare at this a while to understand what it was doing.

regexec.c
```diff
@@ -366,25 +370,33 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p _pDEPTH)
...
-        paren--;
-    }
+    /* substract remaining elements from the stack */
+    PL_savestack_ix -= i;
+
+    /* static assert that offs struc size is not less than stack elem size */
+    STATIC_ASSERT_STMT(sizeof(*rex->offs) >= sizeof(*PL_savestack));
+
+    /* calculate number of offs/capture groups stored */
+    i = (i * sizeof(*PL_savestack)) / sizeof(*rex->offs);
```
demerphq, line 383: I keep staring at this and wondering what happens when `i=1` and `sizeof(PL_savestack) == 4` and `sizeof(*rex->offs) == 12`.  Doesnt it mean that `i` ends ups as `0`? `(1*4)/ 12 == 0`. Is that intended?

I am not necessarily questioning the correctness of this logic, but I do feel it could be a touch more self explaining or use a bit more detail in the comments. It seems that line wants to change the i units of i, but it doesnt explain how the case of a fractional unit is dealt with. Is it just the case that i cannot be 1 when sizeof(rex->ofs) is >= sizeof(PL_savestack)? or is it the case that we always round up and somewhere after we add 1? Its not clear to me if that is what is happening with `*maxopenparen_p - i + 1` in the following statement.

demerphq (changes requested):
I think at least this could use some more detailed explanation in the comments why it is correct. I had to stare at a couple of parts to understand what it is doing.

regexec.c
AnFunctionArray: I see ok I'll try to add more detailed comments as soon as I have some free time.

demerphq: Thanks. For what its worth I really appreciate you writing this patch, its a great idea.

regexec.c
AnFunctionArray: Actually I had the tree laying around so added comments earlier - please check them out and tell me if they are sufficient.

AnFunctionArray: It's +1 because it's like that on other places (and in regcppop originally I believe).

demerphq: The comments are helpful, but its still not clear to me from the code alone why `i` ending up as 0 would be ok, or why it would never happen.

AnFunctionArray: > I keep staring at this and wondering what happens when `i=1` and `sizeof(PL_savestack) == 4` and `sizeof(*rex->offs) == 12`. Doesnt it mean that `i` ends ups as `0`? `(1*4)/ 12 == 0`. Is that intended?
> 
> I am not necessarily questioning the correctness of this logic, but I do feel it could be a touch more self explaining or use a bit more detail in the comments. It seems that line wants to change the i units of i, but it doesnt explain how the case of a fractional unit is dealt with. Is it just the case that i cannot be 1 when sizeof(rex->ofs) is >= sizeof(PL_savestack)? or is it the case that we always round up and somewhere after we add 1? Its not clear to me if that is what is happening with `*maxopenparen_p - i + 1` in the following statement.

This scenario is impossible since earlier (in regcppush) we align the elements.

In this case `i` would be equal to at least 3.

tonycoz:
It looks reasonable to me at this point.

I did a performance test using the code from the reproduction of #20411 (with a smaller source file).

It made a huge difference in the number of instructions, from 9.8e9 to 2.6e9 instructions, from 5e6 branch misses to 3.6e6 branch misses.
<details><summary>raw perf output</summary>

```
blead:

Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

          1,334.39 msec task-clock                #    0.998 CPUs utilized          
                 5      context-switches          #    0.004 K/sec                  
                 1      cpu-migrations            #    0.001 K/sec                  
           122,041      page-faults               #    0.091 M/sec                  
     6,244,474,923      cycles                    #    4.680 GHz                    
     9,844,500,190      instructions              #    1.58  insn per cycle         
       847,094,914      branches                  #  634.816 M/sec                  
         5,295,250      branch-misses             #    0.63% of all branches        

       1.336585169 seconds time elapsed

       1.203539000 seconds user
       0.131949000 seconds sys


Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

          1,352.82 msec task-clock                #    1.000 CPUs utilized          
                 4      context-switches          #    0.003 K/sec                  
                 1      cpu-migrations            #    0.001 K/sec                  
           122,047      page-faults               #    0.090 M/sec                  
     6,269,318,583      cycles                    #    4.634 GHz                    
     9,847,125,497      instructions              #    1.57  insn per cycle         
       847,748,958      branches                  #  626.652 M/sec                  
         5,119,632      branch-misses             #    0.60% of all branches        

       1.353153777 seconds time elapsed

       1.208993000 seconds user
       0.144118000 seconds sys
 Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

          1,331.40 msec task-clock                #    1.000 CPUs utilized          
                 3      context-switches          #    0.002 K/sec                  
           122,048      page-faults               #    0.092 M/sec                  
     6,218,556,628      cycles                    #    4.671 GHz                    
     9,844,846,499      instructions              #    1.58  insn per cycle         
       847,218,356      branches                  #  636.337 M/sec                  
         5,008,442      branch-misses             #    0.59% of all branches        

       1.331663734 seconds time elapsed

       1.211690000 seconds user
       0.119969000 seconds sys

19814:

 Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

            449.84 msec task-clock                #    0.917 CPUs utilized          
                 2      context-switches          #    0.004 K/sec                  
           122,019      page-faults               #    0.271 M/sec                  
     2,070,819,713      cycles                    #    4.603 GHz                    
     2,567,054,451      instructions              #    1.24  insn per cycle         
       510,284,522      branches                  # 1134.359 M/sec                  
         3,573,565      branch-misses             #    0.70% of all branches        

       0.490348496 seconds time elapsed

       0.333654000 seconds user
       0.116578000 seconds sys

Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

            436.59 msec task-clock                #    0.999 CPUs utilized          
                 0      context-switches          #    0.000 K/sec                  
           122,017      page-faults               #    0.279 M/sec                  
     2,043,129,622      cycles                    #    4.680 GHz                    
     2,567,131,711      instructions              #    1.26  insn per cycle         
       510,323,315      branches                  # 1168.880 M/sec                  
         3,630,996      branch-misses             #    0.71% of all branches        

       0.436871340 seconds time elapsed

       0.344714000 seconds user
       0.092191000 seconds sys
Performance counter stats for '../../perl5/perl -I/home/tony/dev/perl/git/perl5/lib parse.pl ../shorter.pp':

            438.77 msec task-clock                #    0.999 CPUs utilized          
                 0      context-switches          #    0.000 K/sec                  
           122,020      page-faults               #    0.278 M/sec                  
     2,054,471,588      cycles                    #    4.682 GHz                    
     2,565,659,212      instructions              #    1.25  insn per cycle         
       509,973,832      branches                  # 1162.293 M/sec                  
         3,678,948      branch-misses             #    0.72% of all branches        

       0.439053740 seconds time elapsed

       0.314207000 seconds user
       0.124877000 seconds sys
```

</details>

demerphq:
@khwilliamson @tonycoz @demerphq - what do we need to get this merged finally?

demerphq:
I see we have ack already, and @tonycoz seems happy, ill do a further review and assuming im happy with it ill merge it.

demerphq (approved):

demerphq:
This was merged as 66964e141c6594291e27649d190bc6ca573861f2

I made two whitespace changes to the commit, one to the commit text and one to remove a spurious whitespace change, and since there had been so much chatter on this before I decided not to ask @AnFunctionArray to do it himself.  I then manually pushed to blead. 

Closing this as merged.

demerphq:
BTW, thanks for the patch!

AnFunctionArray:
> BTW, thanks for the patch!

Well no problem - I've definitely had fun doing this after my crazy shenanigans - trying to add a way to preserve capture groups after recursion of a pattern (by adding new syntax) - but this and added multithreading plus some micro optimisations (like removing `(??{})`) - which are by nature slow (because of the pattern recompiling each time) - kinda did the same thing for my project and I'm happy :).

PR #19817, opened 2022-06-02
Make fc(), qr//i thread-safe on participating platforms

khwilliamson:
A long standing bug in Perl that has gone undetected is that the array
is global that is created when changing locales and tells fc() and qr//i
matching what the folds are in the new locale.

What this means is that any program only has one set of fold definitions
that apply to all threads within it, even if we claim that the locales
are thread-safe on the given platform.  One possibility for this going
undetected so long is that no one is using locales on multi-threaded
systems much.  Another possibility is that modern UTF-8 locales have the
same set of folds as any other one.

It is a simple matter to make the fold array per-thread instead of
per-process, and that solves the problem transparently to other code.

I discovered this stress-testing locale handling under threads.  That
test will be added in a future commit.


inline.h
```diff
@@ -3042,6 +3042,8 @@ Perl_foldEQ_locale(const char *s1, const char *s2, I32 len)
 {
     const U8 *a = (const U8 *)s1;
     const U8 *b = (const U8 *)s2;
+    dTHX;   /* XXX pTHX_ for this, but would have to make all similar fcns the
```
Leont: IMO, `dTHX` is generally a bad idea. This really should be using `pTHX_` instead.

PR #19819, opened 2022-06-03
Rename variable in EXTEND_HWM_SET

### LINKED ISSUE #19818: EXTEND_HWM_SET() should not declare "ix"
We need to be a lot more careful in macros to avoid using variable names likely to appear in the interpolated arguments. See https://perlmonks.org/?node_id=11144355 for an example, which I reckon was unreasonably hard to debug:
```
                EXTEND(SP, ix == 3 ? 2*n : n);
```

I've no idea what the compiler is actually trying to do with this in:
```
#  define EXTEND_HWM_SET(p, n)                      \
        STMT_START {                                \
            SSize_t ix = (p) - PL_stack_base + (n); \
            if (ix > PL_curstackinfo->si_stack_hwm) \
                PL_curstackinfo->si_stack_hwm = ix; \
        } STMT_END
#  define EXTEND(p,n)   STMT_START {                                    \
                         EXTEND_HWM_SET(p, n);                          \
                         if (UNLIKELY(_EXTEND_NEEDS_GROW(p,n))) {       \
                           sp = stack_grow(sp,p,_EXTEND_SAFE_N(n));     \
                           PERL_UNUSED_VAR(sp);                         \
                         } } STMT_END
```
but I suspect we're in nasal demon territory.

hvds:
"ix" is a common variable name in XS code, and may be passed in as part of a parameter.

Fixes #19818, could be considered for backporting (affects v5.28 and later).

### DIFF
--- a/pp.h
+++ b/pp.h
@@ -316,11 +316,11 @@ Does not use C<TARG>.  See also C<L</XPUSHu>>, C<L</mPUSHu>> and C<L</PUSHu>>.
  * requested to be extended (which is likely to be less than PL_stack_max)
  */
 #if defined DEBUGGING && !defined DEBUGGING_RE_ONLY
-#  define EXTEND_HWM_SET(p, n)                      \
-        STMT_START {                                \
-            SSize_t ix = (p) - PL_stack_base + (n); \
-            if (ix > PL_curstackinfo->si_stack_hwm) \
-                PL_curstackinfo->si_stack_hwm = ix; \
+#  define EXTEND_HWM_SET(p, n)                                     \
+        STMT_START {                                               \
+            SSize_t extend_hwm_set_ix = (p) - PL_stack_base + (n); \
+            if (extend_hwm_set_ix > PL_curstackinfo->si_stack_hwm) \
+                PL_curstackinfo->si_stack_hwm = extend_hwm_set_ix; \
         } STMT_END
 #else
 #  define EXTEND_HWM_SET(p, n) NOOP


khwilliamson:
I didn't understand what the problem was until I looked at the perlmonks report.

I believe this  problem goes away if the macro made internal copies of its parameters.  I think we should be stressing doing that, as otherwise, calling it with an expression with side effects gives unexpected results, and its trivial with  STMT_START ... STMT_END to avoid this.

To this end, I have written #19823.  After we come to agreement on the wording for that, I would like this PR to become whatever that wording ends up as

hvds:
> I didn't understand what the problem was until I looked at the perlmonks report.
> 
> I believe this problem goes away if the macro made internal copies of its parameters. I think we should be stressing doing that, as otherwise, calling it with an expression with side effects gives unexpected results, and its trivial with STMT_START ... STMT_END to avoid this.
> 
> To this end, I have written #19823. After we come to agreement on the wording for that, I would like this PR to become whatever that wording ends up as

Could you clarify here what you think the specific problem was in the PerlMonks issue? I don't see an "expression with side-effects".

It appeared to me that what I saw was equivalent to the `ix` on the RHS of `SSize_t ix = (p) - PL_stack_base + (n);` (after argument substition of `ix == 3 ? 2*n : n` for `n`) being compiled as a reference to the (as yet uninitialized) `ix` declared on the LHS. But I'm pretty vague on precisely how this should work, so it's entirely possible I'm misinterpreting what I saw.

Could you also clarify what you think this PR should change to? Because reading #19823 it seems like you're proposing that it should be an identical change (or, optionally, use a static inline function instead - which I'd be all for, if we think that's a change we can safely make).

khwilliamson:
Your fix works for the problem in perlmonks, but we should be writing macros that are bomb-proof whenever possible.

I am saying we should add 
`            SV **   extend_hwm_set_p = p;           \
            SSize_t extend_hwm_set_n = n;           \
`
and refer to those local variables inside the block.  Then `ix` wouldn't have to change at all, but it is better practice to change its name as suggested.

As I write this, realize a flaw in what I had said  Macros can be a poor substituted for C++ templates, where the type can vary.  My proposal only works if the type of the formal parameters is always going to be the same in every call

hvds:
> Your fix works for the problem in perlmonks, but we should be writing macros that are bomb-proof whenever possible.
> 
> I am saying we should add ` SV ** extend_hwm_set_p = p; \ SSize_t extend_hwm_set_n = n; \` and refer to those local variables inside the block. Then `ix` wouldn't have to change at all, but it is better practice to change its name as suggested.

Thanks, that makes the second part clearer.

For the first part, I was confused though by the assertion in #19823 that "The name of the declared copy can be anything in this case, even being the same as the formal parameter name.", which seemed to imply that it would be safe to write:
```
#define my_single_parameter(sv) STMT_START { \
    SV *   sv = sv; \
...
```

I would not expect the above to be safe: I'd expect the text `sv` to be replaced on both sides of the assignment by the parameter value.

> As I write this, realize a flaw in what I had said Macros can be a poor substituted for C++ templates, where the type can vary. My proposal only works if the type of the formal parameters is always going to be the same in every call

Right, and under those restrictions there would appear to be no impediment to using the clearer, safer option of a static inline function.

Given the very widespread use of `EXTEND()`, my proposed change seems like something that could safely be backported; anything that changes the type-checking on the arguments would need much more extensive testing.

khwilliamson:
On 6/5/22 08:53, Hugo van der Sanden wrote:
> For the first part, I was confused though by the assertion in #19823 
> <https://github.com/Perl/perl5/pull/19823> that "The name of the 
> declared copy can be anything in this case, even being the same as the 
> formal parameter name.", which seemed to imply that it would be safe to 
> write:
> 
> |#define my_single_parameter(sv) STMT_START { \ SV * sv = sv; \ ... |
> 
> I would not expect the above to be safe: I'd expect the text |sv| to be 
> replaced on both sides of the assignment by the parameter value.
> 

You are right; I was wrong.

khwilliamson:
On 6/5/22 08:53, Hugo van der Sanden wrote:
> Given the very widespread use of |EXTEND()|, my proposed change seems 
> like something that could safely be backported; anything that changes 
> the type-checking on the arguments would need much more extensive testing.

Agreed.  I'm fine with merging this and backporting it

iabyn:
LGTM

hvds:
Now merged as c31ae9a243.

I don't know much about ppport.h, is that a possible approach to making this available instead of backporting?

PR #19820, opened 2022-06-03
GH19478: applying taint is no reason to mess with pos

hvds:
25fdce4a16 introduced a chunk in sv_magic() to "force pos to be stored
as characters, not bytes" whenever any magic was applied to a string
marked UTF8.

It is not clear why a random call to sv_magic(), eg to mark a string as
tainted, needs to do this - it would seem more logical for the check to
happen either earlier (when the string first qualifies as SvMAGICAL(sv)
&& DO_UTF8(sv)) or later (eg on mg_find).

Experimentally remove this block - it appears to cause no test failures,
and allows the new test cases to pass.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -5825,16 +5825,6 @@ Perl_sv_magic(pTHX_ SV *const sv, SV *const obj, const int how,
         }
     }
 
-    /* Force pos to be stored as characters, not bytes. */
-    if (SvMAGICAL(sv) && DO_UTF8(sv)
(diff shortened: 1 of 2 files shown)


khwilliamson (approved):

khwilliamson:
I think we should apply this now and see what happens

hvds:
> I think we should apply this now and see what happens

I was hoping we might still get a comment from @iabyn, but you're right. I've pushed it as  f757874dac.

tonycoz:
This has been applied, so closing.

PR #19827, opened 2022-06-06
Refactor warning-generating code in two different places into one common function

leonerd:
The new combined function, named `S_warn_elem_scalar_context()`, is much more meaningful and useful to become exported in an intended subsequent change, than the previous (poorly-named) `S_op_pretty()` would have been.

### DIFF
--- a/op.c
+++ b/op.c
@@ -1839,36 +1839,72 @@ S_op_varname(pTHX_ const OP *o)
     return S_op_varname_subscript(aTHX_ o, 1);
 }
 
+/*
+
+Warns that an access of a single element from a named container variable in
+scalar context might not be what the programmer wanted. The container
+variable's (sigiled, full) name is given by C<name>, and the key to access
+it is given by the C<SVOP_sv> of the C<OP_CONST> op given by C<o>.
+C<is_hash> selects whether it prints using {KEY} or [KEY] brackets.
+
+C<is_slice> selects between two different messages used in different places.
+ */
 static void
-S_op_pretty(pTHX_ const OP *o, SV **retsv, const char **retpv)
-{ /* or not so pretty :-) */
+S_warn_elem_scalar_context(pTHX_ const OP *o, SV *name, bool is_hash, bool is_slice)
+{
+    SV *keysv;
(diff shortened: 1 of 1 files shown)


op.c
```diff
@@ -1839,36 +1839,72 @@ S_op_varname(pTHX_ const OP *o)
...
 
+/*
+
+Warns that an access of a single element from a named container variable in
+scalar context might not be what the programmer wanted. The container
+variable's (sigiled, full) name is given by I<name>, and the key to access
+it is given by the C<SVOP_sv> of the C<OP_CONST> op given by I<o>.
+I<is_hash> selects whether it prints using {KEY} or [KEY] brackets.
+
+I<is_slice> selects between two different messages used in different places.
```
khwilliamson: AFAIK, almost all other markup uses C<> for formal parameter names.  I don't know how important consistency is, especially since this doesn't actually go into pod

op.c
```diff
@@ -1839,36 +1839,72 @@ S_op_varname(pTHX_ const OP *o)
...
 static void
-S_op_pretty(pTHX_ const OP *o, SV **retsv, const char **retpv)
-{ /* or not so pretty :-) */
+S_warn_elem_scalar_context(pTHX_ const OP *o, SV *name, bool is_hash, bool is_slice)
+{
+    SV *keysv;
+    const char *keypv = NULL;
+
+    const char lbrack = is_hash ? '{' : '[';
+    const char rbrack = is_hash ? '}' : ']';
```
khwilliamson, line 1859: refactoring could use just one conditional here.  Likely the compiler will do it under optimization, and if not, likely not a performance concern.  But other places in the core would use is_hash as an index into a 2 character string "{[" and same for the terminator; or even use math to have a single 4 character string like "{{[]"

khwilliamson (approved):

op.c
leonerd: I mostly just copypasted this from one of the existing sites of the previous version of the code already. I did think about indexing into a string though.

op.c
leonerd: Ahyes; fixed and force-pushed.

PR #19830, opened 2022-06-07
Fix indentation in op.c to not be misleading

leonerd:
gcc did not previously warn about this code, but in my branch where I split the giant `op.c` file into some smaller parts, it started warning:

```
peep.c: In function ‘S_aassign_scan’:
peep.c:1292:9: warning: this ‘else’ clause does not guard... [-Wmisleading-indentation]
 1292 |         else
      |         ^~~~
peep.c:1295:13: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘else’
 1295 |             o = o->op_sibparent; /* try parent's next sibling */
      |             ^
```

I suspect it just gave up when it was in `op.c` because the file was so long, but in the smaller `peep.c` it noticed the problem.

I've also generally reïndented the body of the file so it now matches the code.

This is a **whitespace-only** change and should not cause any changes in behaviour.

### DIFF
--- a/op.c
+++ b/op.c
@@ -16278,9 +16278,8 @@ S_aassign_scan(pTHX_ OP* o, bool rhs, int *scalars_p)
             if (o == effective_top_op)
                 effective_top_op = next_kid;
         }
-        else
-            if (o == effective_top_op)
-                effective_top_op = o->op_sibparent;
+        else if (o == effective_top_op)
+            effective_top_op = o->op_sibparent;
         o = o->op_sibparent; /* try parent's next sibling */
     }
     o = next_kid;
--- a/op.c
+++ b/op.c
@@ -16089,202 +16089,201 @@ S_aassign_scan(pTHX_ OP* o, bool rhs, int *scalars_p)
     int all_flags = 0;
 
     while (1) {
-    bool top = o == effective_top_op;
-    int flags = 0;
-    OP* next_kid = NULL;
+        bool top = o == effective_top_op;
+        int flags = 0;
+        OP* next_kid = NULL;
 
-    /* first, look for a solitary @_ on the RHS */
-    if (   rhs
-        && top
-        && (o->op_flags & OPf_KIDS)
-        && OP_TYPE_IS_OR_WAS(o, OP_LIST)
-    ) {
-        OP *kid = cUNOPo->op_first;
-        if (   (   kid->op_type == OP_PUSHMARK
-                || kid->op_type == OP_PADRANGE) /* ex-pushmark */
-            && ((kid = OpSIBLING(kid)))
-            && !OpHAS_SIBLING(kid)
-            && kid->op_type == OP_RV2AV
-            && !(kid->op_flags & OPf_REF)
-            && !(kid->op_private & (OPpLVAL_INTRO|OPpMAYBE_LVSUB))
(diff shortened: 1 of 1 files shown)


op.c
```diff
@@ -16281,8 +16281,7 @@ S_aassign_scan(pTHX_ OP* o, bool rhs, int *scalars_p)
         else
             if (o == effective_top_op)
                 effective_top_op = o->op_sibparent;
```
ilmari: Would it not be even clearer to also pull the `if` up on the same line as the `else`?

leonerd:
The apparent diff here is large but you can ignore the changes in indentation by using

https://github.com/Perl/perl5/pull/19830/files?diff=unified&w=1

whereupon the only visible change is two deleted blank lines, and a joining of separate `else .. if` into a single `else if` on the same line.

ilmari (approved):

jkeenan:
> gcc did not previously warn about this code, but in my branch where I split the giant `op.c` file into some smaller parts, it started warning:

Which version(s) of gcc are you using here?

Also, is this on Linux or some other OS?

Thank you very much.
jimk

leonerd:
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Debian 11.3.0-3' --with-bugurl=file:///usr/share/doc/gcc-11/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-11 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-11-6lTGZp/gcc-11-11.3.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-11-6lTGZp/gcc-11-11.3.0/debian/tmp-gcn/usr --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.3.0 (Debian 11.3.0-3) 
```

PR #19835, opened 2022-06-08
Split peep from op.c

leonerd:
`op.c` was getting too large - over 300k bytes and 18k lines. This PR begins the work of splitting it into smaller sections, by removing the (topdown and peephole) optimizers and optree finalizer into their own new file, `peep.c`.

As discussed in this perl5-porters@ thread:

https://www.nntp.perl.org/group/perl.perl5.porters/2021/12/msg262118.html


peep.c
```diff
@@ -1,3 +1,23 @@
...
+ *
+ *    Copyright (C) 1991-2022 by Larry Wall and others
+ *
+ *    You may distribute under the terms of either the GNU General Public
+ *    License or the Artistic License, as specified in the README file.
+ *
+ */
+
+/*
+ * 'Programmers waste enormous amounts of time thinking about, or worrying
```
Leont: Having a non-standard quote feels more awkward to me than having no quote, TBH

leonerd: Ah; fair. I can remove it and queue it up with the other recently-added files to find a real LotR quote if that's preferred.

Leont:
How about this quote from the chapter "The Departure of Boromir"

> Aragorn sped on up the hill. Every now and again he bent to the ground. Hobbits go light, and their footprints are not easy even for a Ranger to read, but not far from the top a spring crossed the path, and in the wet earth he saw what he was seeking.
> 'I read the signs aright,' he said to himself. 'Frodo ran to the hill-top. I wonder what he saw there? But he returned by the same way, and went down the hill again.'

leonerd:
Quote changed.

leonerd:
Are there any objections to merging this soon? Being a large move of lots of code out of `op.c` it has a danger of causing more merge conflicts the longer we keep it open. I'm therefore keen to merge it soon.

If nobody says otherwise, I'd like to hit "merge" by end of the weekend.

op.h
```diff
@@ -64,6 +64,12 @@ typedef PERL_BITFIELD16 Optype;
     U8		op_private;
 #endif
 
+#define OpTYPE_set(o,type)                      \
```
khwilliamson, line 67: The new guidance about STMT_START { ... } STMT_END talk about how to bomb-proof such macros.  I suspect it isn't an issue here as long as long as both  o and type are not going to be expressions or have side effects.  I'm pointing this out to be sure you consider it.

khwilliamson: I did not verify that there were no substantial changes introduced by splitting the files.  I view it likely there weren't; so given that proviso it LGTM

leonerd: Since it's easy enough to defend against that multiple evaluation case here by allocating new variables, I'll just add it anyway.

PR #19837, opened 2022-06-08
Convert '!!' to cBOOL()

khwilliamson:
I believe the '!!' is somewhat obscure; I for one didn't know about it
for years of programming C, and it was buggy in one compiler, which is why
cBOOL was created, I believe.

This commit dates from before we moved to C99 where we can simply cast
to (bool), and cBOOL() has been rewritten to do that.  But the vast
majority of code uses cBOOL(), and this commit brings the remainder of
the core .[ch] files into uniformity.


ericherman (approved):
My knee-jerk response not very positive when I read the message, but looking at the diff, I think it is an improvement to code clarity.

rwp0 (approved):

tonycoz (approved):

PR #19838, opened 2022-06-08
A better error message for `try {} catch {}` missing its (VAR) (fixes #19811)

### LINKED ISSUE #19811: `catch` without `var` could have a more helpful error message
Users who are familiar with more generic `try/catch` syntax may be used to a plain `catch { ... }` block and thus forget to put the `($var)` part on. On 5.36 this results in a very vague error message:

```
$ perl5.36.0 
use feature 'try';   

try { A() }
catch { B() }  

syntax error at - line 5, near "catch {"
Execution of - aborted due to compilation errors.
```

It would be much more useful to new users to catch (pardon the phrasing) this particular trap, if the message said something like:

```
catch blocks require a (VAR) at - line 5, near "catch {"
```

leonerd:
As suggested in https://github.com/Perl/perl5/issues/19811


rwp0 (approved):

leonerd:
Does anyone want to review this? It already has one green tick - if no other objections by endofplay today I shall merge it

PR #19846, opened 2022-06-09
Add a PERL_MEM_LOG=c option that prints more levels of C backtrace

leonerd:
Setting just `PERL_MEM_LOG=s` is almost useless, because basically every SV appears to be allocated by calls to `Perl_newSV_type()` from `inline.h` line 369 (or whatever).

This new flag requests that the `Perl_c_backtrace` facility is used to obtain more information about the C-level caller context of the SV allocation operation, which will likely be of more help when reading the debug log.

### DIFF
--- a/pod/perlhacktips.pod
+++ b/pod/perlhacktips.pod
@@ -1699,6 +1699,8 @@ determine whether to log the event, and if so how:
 
     $ENV{PERL_MEM_LOG} =~ /m/           Log all memory ops
     $ENV{PERL_MEM_LOG} =~ /s/           Log all SV ops
+    $ENV{PERL_MEM_LOG} =~ /c/           Additionally log C backtrace for
+                                        new_SV events
     $ENV{PERL_MEM_LOG} =~ /t/           include timestamp in Log
     $ENV{PERL_MEM_LOG} =~ /^(\d+)/      write to FD given (default is 2)
 
@@ -1713,6 +1715,10 @@ Since the logging doesn't use PerlIO, all SV allocations are logged and
 no extra SV allocations are introduced by enabling the logging.  If
 compiled with C<-DDEBUG_LEAKING_SCALARS>, the serial number for each SV
 allocation is also logged.
+
+The C<c> option uses the C<Perl_c_backtrace> facility, and therefore
+additionally requires the Configure C<-Dusecbacktrace> compile flag in
+order to access it.
 
 =head2 DDD over gdb
 
--- a/util.c
+++ b/util.c
@@ -5218,9 +5218,10 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
             PERL_UNUSED_RESULT(PerlLIO_write(fd, buf, len));
 #ifdef USE_C_BACKTRACE
             if(strchr(pmlenv,'c') && (mlt == MLT_NEW_SV)) {
-                /* TODO: get caller package in here when we work out how */
                 len = my_snprintf(buf, sizeof(buf),
-                        "  caller at %s line %d\n",
+                        "  caller %s at %s line %d\n",
+                        /* CopSTASHPV can crash early on startup; use CopFILE to check */
+                        CopFILE(PL_curcop) ? CopSTASHPV(PL_curcop) : "<unknown>",
                         CopFILE(PL_curcop), CopLINE(PL_curcop));
                 PERL_UNUSED_RESULT(PerlLIO_write(fd, buf, len));
 
--- a/util.c
+++ b/util.c
@@ -5107,7 +5107,7 @@ Perl_debug_hash_seed(pTHX_ bool via_debug_h)
 /* -DPERL_MEM_LOG_SPRINTF_BUF_SIZE=X: size of a (stack-allocated) buffer
  * the Perl_mem_log_...() will use (either via sprintf or snprintf).
  */
-#define PERL_MEM_LOG_SPRINTF_BUF_SIZE 128
+#define PERL_MEM_LOG_SPRINTF_BUF_SIZE 256
 
 /* -DPERL_MEM_LOG_FD=N: the file descriptor the Perl_mem_log_...()
  * writes to.  In the default logger, this is settable at runtime.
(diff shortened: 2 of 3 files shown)


pod/perlhacktips.pod
```diff
@@ -1695,6 +1695,8 @@ determine whether to log the event, and if so how:
 
     $ENV{PERL_MEM_LOG} =~ /m/           Log all memory ops
     $ENV{PERL_MEM_LOG} =~ /s/           Log all SV ops
+    $ENV{PERL_MEM_LOG} =~ /c/           Additionally log C backtrace for
+                                        new_SV events
```
ilmari, line 1703: How about how about `$ENV{PERL_MEM_LOG} =~ /c(\d*)/` to log the specified number of levels of backtrace (default 3)?

leonerd: I could add that but that'll complicate the sanitisation code in `handy.h` relating to it.

util.c
```diff
@@ -5227,6 +5227,12 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
             PERL_UNUSED_RESULT(PerlLIO_write(fd, buf, len));
 #ifdef USE_C_BACKTRACE
             if(strchr(pmlenv,'c') && (mlt == MLT_NEW_SV)) {
+                /* TODO: get caller package in here when we work out how */
```
ilmari: [`CopSTASHPV`](https://perldoc.perl.org/perlapi#CopSTASHPV)?

leonerd: I tried `CopSTASHPV` and it just immediately SEGV'ed. I see other calls to it that apparently work so I'm not sure what I'm doing differently.

pod/perlhacktips.pod
```diff
@@ -1710,6 +1712,10 @@ no extra SV allocations are introduced by enabling the logging.  If
 compiled with C<-DDEBUG_LEAKING_SCALARS>, the serial number for each SV
 allocation is also logged.
 
+The C<c> option uses the C<Perl_c_backtrace> facility, and therefore
+additionally requires the C<-DUSE_C_BACKTRACE> compile flag in order to
```
tonycoz: This should probably refer to the Configure -Dusecbacktrace flag instead, which will check for the required library support.

util.c
```diff
@@ -5225,6 +5225,32 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
...
+
+                Perl_c_backtrace *bt = Perl_get_c_backtrace(aTHX_ 3, 3);
+                Perl_c_backtrace_frame *frame;
+                UV i;
+                for (i = 0, frame = bt->frame_info;
+                        i < bt->header.frame_count;
+                        i++, frame++) {
+                    len = my_snprintf(buf, sizeof(buf),
+                            "  frame[%d]: %p %s at %s +0x%lx\n",
+                            i,
```
tonycoz: `%d` expects `int`, but i is a `UV`.

util.c
```diff
@@ -5225,6 +5225,32 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
...
+                            frame->addr,
+                            frame->symbol_name_size && frame->symbol_name_offset ? (char *)bt + frame->symbol_name_offset : "-",
+                            frame->object_name_size && frame->object_name_offset ? (char *)bt + frame->object_name_offset : "?",
+                            (char *)frame->addr - (char *)frame->object_base_addr);
+                    PerlLIO_write(fd, buf, len);
```
tonycoz: The PerlIO_write() should be wrapped in PERL_UNUSED_RRSULT() to avoid
```
In file included from perl.h:4117,
                 from util.c:26:
iperlsys.h:802:41: warning: ignoring return value of ‘write’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
  802 | #  define PerlLIO_write(fd, buf, count) write((fd), (buf), (count))
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
iperlsys.h:802:41: note: in definition of macro ‘PerlLIO_write’
  802 | #  define PerlLIO_write(fd, buf, count) write((fd), (buf), (count))
      |                                         ^~~~~
```

util.c
leonerd: Ahyes; I'll use `UVuf`

leonerd:
Comments addressed.

pod/perlhacktips.pod
khwilliamson: I agree with Leonerd

khwilliamson (approved):
I haven't looked at the code in a while, but given that the first level is useless, always coming from the inline.h function, could that be solved by creating a macro that adds __FILE__ and __LINE__ to the call?  Also, I hadn't known about Perl_deb() which outputts file, line, and pid.  I haven't tried it but I'm guessing the file and line are the perl values rather than the C one.

util.c
tonycoz: The memory logging functions can be called really early, for threaded builds before PL_stashpad is initialized.  CopSTASH() is `PL_stashpad[(c)->cop_stashoff]` for threaded builds, so it segfaults.

We could make CopSTASH() check that PL_stashpad is non-NULL, but then every other call which doesn't need that check will also be paying the cost.

Using CopFILE() as a proxy for checking for the validity of CopSTASHPV() seemed to work:
```
diff --git a/util.c b/util.c
index fa2d129186..6069c80cfe 100644
--- a/util.c
+++ b/util.c
@@ -5229,8 +5229,9 @@ S_mem_log_common(enum mem_log_type mlt, const UV n,
             if(strchr(pmlenv,'c') && (mlt == MLT_NEW_SV)) {
                 /* TODO: get caller package in here when we work out how */
                 len = my_snprintf(buf, sizeof(buf),
-                        "  caller at %s line %d\n",
-                        CopFILE(PL_curcop), CopLINE(PL_curcop));
+                        "  caller at %s line %d stash %s\n",
+                        CopFILE(PL_curcop), CopLINE(PL_curcop),
+                        CopFILE(PL_curcop) ? CopSTASHPV(PL_curcop) : "<unknown>");
                 PERL_UNUSED_RESULT(PerlLIO_write(fd, buf, len));
 
                 Perl_c_backtrace *bt = Perl_get_c_backtrace(aTHX_ 3, 3);
```

I didn't see a crash with non-threaded builds.

leonerd: Ahyes, that seems to do it.

pod/perlhacktips.pod
ilmari: Fair enough, that can be a future enhancement if someone feels a strong need.

PR #19848, opened 2022-06-10
A batch of  improving locale handling commits

khwilliamson:
It turns out that there are a lot of issues with locale handling: bugs, inefficiencies, unnecessarily crippled functionality on some platforms.  Several years ago I started working on this.  GH #19828 started working on handy.h to improve that side of things.  This PR is the first major one for locale.c proper.


tonycoz:
> this. GH #19628 started working on handy.h

Is that the right ticket id?

locale.c
```diff
@@ -330,17 +330,17 @@ S_category_name(const int category)
 
 /* Windows requres a customized base-level setlocale() */
 #ifdef WIN32
-#  define my_setlocale(cat, locale) win32_setlocale(cat, locale)
+#  define porcelain_setlocale(cat, locale) win32_setlocale(cat, locale)
```
tonycoz, line 368: git usage of "porcelain" is pretty ambiguous.

https://stackoverflow.com/questions/6976473/what-does-the-term-porcelain-mean-in-git

From that, porcelain *commands* are high level commands that depend on the plumbing (which are the low-level commands).

But the various `--porcelain` switches and values are used for high level commands to produce stable, script readable output.

khwilliamson: Ticket changed to #19828, which already used 'porcelain'.  I'll change the wording, but since it's already in blead, I'll push this as-is, and do it all at the same time

PR #19861, opened 2022-06-14
Fix some issues in t/loc_tools.pl

khwilliamson:


tonycoz (approved):
Looks fine to me

t/loc_tools.pl
```diff
@@ -589,7 +591,7 @@ ($)
     }
 
     die "Couldn't restore locale '$save_locale'"
-        unless setlocale(&POSIX::LC_CTYPE(), $save_locale);
+                            unless setlocale(&POSIX::LC_CTYPE(), $save_locale);
```
jkeenan, line 594: I don't understand why you are increasing the indent so much in the line above.

jkeenan (commented):
I've never spent much time in `t/loc_tools.pl`, so I've never encountered most of the problems you're addressing in this pull request.  Can you provide links, for example, to any of the crashes in locales-handling that you mention in the commits.

t/loc_tools.pl
khwilliamson: It is a style used in some spots of the core.   If a line is broken so as to make it fit in the recommended line width, a hanging indent is created to make the second line end close to that max width.  It makes that line more obviously a continuation, and the eye doesn't have to move so far to the left.  I use it sometimes when it seems like the right thing.  I don't know of any discussions about this style; I certainly didn't originate it.

khwilliamson:
I'll add a few current examples in the merge commit message

khwilliamson:
Merged via 2fbb1a06df7219e842b3a2a0c9dcbab10de433fc

PR #19864, opened 2022-06-15
allow building with -DPERL_MEM_LOG on Win32

tonycoz:
This appears to have been broken for a while, and became more broken
with 75acd14e, which made newSV_type() inline.

This will also prevent warnings about calls to undeclared functions
on systems that don't need symbols to be exported.


leonerd (approved):
LGTM

khwilliamson:
LGTM

PR #19868, opened 2022-06-16
Fix previous PR by running Porting/updateAUTHORS.pl

nicomen:


Leont:
This rather looks like it needs a rebase

nicomen:
I rebased before committing again to avoid any potential conflicts

nicomen:
The script Prting/updateAUTHORS.pl run dragged in a few new names in addition to mine so I was worried there were changes I would mess up if I just pushed a commit with several new names without rebasing.

Leont:
> I rebased before committing again to avoid any potential conflicts

Two of those commits are already in blead, and a third is a merge commit. Only one of these four commits should be in the PR.

Don't argue, just rebase when you're asked to.

nicomen:
Ok, I'm not trying to argue. I tried to explain what I did. I thought it was a bit weird that running updateAUTHORS.pl here locally before any rebase dragged in many new names. I thought it was because I had some partial commits already from the last rebase (before previous PR). So I tried to keep it clean by including the new changes in blead in my branch to get a smaller diff. I think I understand now that I should've included the sole commit of running updateAUTHORS.pl. I did try to re-open the original PR, but that didn't seem to work.

I wonder though should I perhaps have avoided dragging in the other names that came with the script run and only kept mine?

nicomen:
I see now what added to my confusion for me. Karl fixed the AUTHORS thing in between me figuring it out and I messed up. Sorry.

PR #19870, opened 2022-06-16
Storable - maintain boolean values

haarg:
Add a new type to storable for boolean values. Use it to encode booleans
when possible. Decode as booleans which will maintain their boolean
status when possible.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -3985,6 +3985,7 @@ dist/Storable/t/attach.t		Check STORABLE_attach doesn't create objects unnecessa
 dist/Storable/t/attach_errors.t		Trigger and test STORABLE_attach errors
 dist/Storable/t/attach_singleton.t	Test STORABLE_attach for the Singleton pattern
 dist/Storable/t/blessed.t		See if Storable works
+dist/Storable/t/boolean.t		See if Storable works
 dist/Storable/t/canonical.t		See if Storable works
 dist/Storable/t/circular_hook.t		Test thaw hook called depth-first for circular refs
 dist/Storable/t/code.t			See if Storable works
--- a/dist/Storable/Makefile.PL
+++ b/dist/Storable/Makefile.PL
@@ -11,11 +11,14 @@ use warnings;
 use ExtUtils::MakeMaker 6.31;
 use Config;
 
+my $defines = $ENV{PERL_CORE} ? q[] : q[-DUSE_PPPORT_H];
+
 WriteMakefile(
     NAME                => 'Storable',
     AUTHOR              => 'Perl 5 Porters',
     LICENSE             => 'perl',
     DISTNAME            => "Storable",
+    DEFINE              => $defines,
     PREREQ_PM           =>
       {
           XSLoader => 0,
--- a/dist/Storable/Storable.xs
+++ b/dist/Storable/Storable.xs
@@ -16,8 +16,7 @@
 #include <perl.h>
 #include <XSUB.h>
 
-#ifndef PERL_VERSION_LT
-# if !defined(PERL_VERSION) || !defined(PERL_REVISION) || ( PERL_REVISION == 5 && ( PERL_VERSION < 10 || (PERL_VERSION == 10 && PERL_SUBVERSION < 1) ) )
+#ifdef USE_PPPORT_H
 #   define NEED_PL_parser
 #   define NEED_sv_2pv_flags
 #   define NEED_load_module
@@ -25,7 +24,6 @@
 #   define NEED_newCONSTSUB
 #   define NEED_newSVpvn_flags
 #   define NEED_newRV_noinc
-# endif
 #include "ppport.h"             /* handle old perls */
 #endif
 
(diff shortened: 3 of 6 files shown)


haarg:
This is a more conventional alternative to #19871 for supporting boolean values in Storable.

I am not primarily a C programmer, so there may be some mistakes here.

dist/Storable/Storable.xs
```diff
@@ -2452,6 +2455,17 @@ static int store_scalar(pTHX_ stcxt_t *cxt, SV *sv)
             pv = SvPV(sv, len);		/* We know it's SvPOK */
             goto string;			/* Share code below */
         }
+#ifdef SvIsBOOL
+    } else if (SvIsBOOL(sv)) {
+        /* Artificially store bools as 'long' scalars. Strings of length 0 or
+         * 1 shouldn't otherwise be encoded in this format, so it can serve as
+         * a marker that they are booleans.
+         */
```
tonycoz: This doesn't match the code, I assume it came from the other branch.

tonycoz:
The failing tests here with DEBUGGING builds are from failing assertions:
```
ASSERT FAILED ("Storable.xs", line 7350): old and new retrieve dispatch table have same size
ASSERT FAILED ("Storable.xs", line 7352): SX_LAST entry correctly initialized in old dispatch table
```
Unfortunately Storable uses it's own ASSERT() macro which writes to stdout, so they don't show in the github test logs.

Adding:
```
diff --git a/dist/Storable/Storable.xs b/dist/Storable/Storable.xs
index 31abf07079..f3e67ebb15 100644
--- a/dist/Storable/Storable.xs
+++ b/dist/Storable/Storable.xs
@@ -1453,6 +1453,7 @@ static const sv_retrieve_t sv_old_retrieve[] = {
     (sv_retrieve_t)retrieve_other,	/* SX_SVUNDEF_ELEM not supported */
     (sv_retrieve_t)retrieve_other,	/* SX_REGEXP */
     (sv_retrieve_t)retrieve_other,  	/* SX_LOBJECT not supported */
+    (sv_retrieve_t)retrieve_other,  	/* SX_BOOLEAN not supported */
     (sv_retrieve_t)retrieve_other,  	/* SX_LAST */
 };
```
and all tests pass for me.

haarg:
Tests are now passing.

One note about the implementation of this. It uses a boolean type marker and then the next byte encodes the true or false value. However, the existing implementation can encode the immortal booleans (PL_sv_yes, PL_sv_no), and each has its own type marker. It would be possible to use separate type markers for the mortal boolean copies as well for symmetry. I have no particular opinion on this.

tonycoz:
> It would be possible to use separate type markers for the mortal boolean copies as well for symmetry.

Given we have other types of "small encodings" and over 200 unused type markers I think it's a worthwhile optimization.

dist/Storable/Storable.xs
```diff
@@ -5880,6 +5898,50 @@ static SV *retrieve_integer(pTHX_ stcxt_t *cxt, const char *cname)
...
+    HV *stash;
+
+    TRACEME(("retrieve_boolean_true (#%d)", (int)cxt->tagnum));
+
+    sv = newSVsv(&PL_sv_yes);
+    stash = cname ? gv_stashpv(cname, GV_ADD) : 0;
+    SEEN_NN(sv, stash, 0);  /* Associate this new scalar with tag "tagnum" */
+
+    TRACEME(("boolean true"));
+    TRACEME(("ok (retrieve_boolean at 0x%" UVxf ")", PTR2UV(sv)));
```
tonycoz: This and the _false version should probably log the new function names to match all the other similar logging.

haarg: Fixed.

tonycoz (approved):

MANIFEST
```diff
@@ -3985,6 +3985,7 @@ dist/Storable/t/attach.t		Check STORABLE_attach doesn't create objects unnecessa
 dist/Storable/t/attach_errors.t		Trigger and test STORABLE_attach errors
 dist/Storable/t/attach_singleton.t	Test STORABLE_attach for the Singleton pattern
 dist/Storable/t/blessed.t		See if Storable works
+dist/Storable/t/boolean.t		See if Storable works
```
leonerd, line 3988: Maybe this comment needs updating? Though a lot of them are just copied so perhaps it's fine.

dist/Storable/Makefile.PL
```diff
@@ -11,11 +11,14 @@ use warnings;
 use ExtUtils::MakeMaker 6.31;
 use Config;
 
+my $defines = $ENV{PERL_CORE} ? q[] : q[-DUSE_PPPORT_H];
```
leonerd, line 14: Why the weird `q[...]` quoting? A regular `""` string would be fine

dist/Storable/Storable.xs
```diff
@@ -5882,6 +5898,50 @@ static SV *retrieve_integer(pTHX_ stcxt_t *cxt, const char *cname)
...
+ * Retrieve boolean true copy.
+ */
+static SV *retrieve_boolean_true(pTHX_ stcxt_t *cxt, const char *cname)
+{
+    SV *sv;
+    HV *stash;
+
+    TRACEME(("retrieve_boolean_true (#%d)", (int)cxt->tagnum));
+
+    sv = newSVsv(&PL_sv_yes);
```
leonerd, line 5913: I *was* going to comment that there is a new `newSVbool` function you can use here, but then writing it this way means this part of the code is portable to older perls before core bools anyway and still does (as close to) the right thing there. So this is fine.

leonerd (approved):
Overall LGTM, give or take a couple of small comments. But it's fine as-is.

PR #19871, opened 2022-06-16
Storable - support boolean values via special case hack

haarg:
Add a special case to Storable to always encode boolean values as 'long'
format. This should never normally happen for strings of length 0 or 1.
This is a valid encoding of these values, even on old versions of
Storable. On older versions, they will decode as "1" and "", the same as
they would have without this change.

When decoding 'long' strings, recognise these abnormally encoded
strings, and decode them as boolean values.

This allows round tripping booleans through Storable without changing
anything about the format.

It would be possible for an existing stored blob to have these abnormal
encodings, but they would have to be manually constructed. And even in
that case, the decoded values would still have the same string value.
They would just additionally be flagged as booleans.

### DIFF
--- a/dist/Storable/Storable.xs
+++ b/dist/Storable/Storable.xs
@@ -16,8 +16,7 @@
 #include <perl.h>
 #include <XSUB.h>
 
-#ifndef PERL_VERSION_LT
-# if !defined(PERL_VERSION) || !defined(PERL_REVISION) || ( PERL_REVISION == 5 && ( PERL_VERSION < 10 || (PERL_VERSION == 10 && PERL_SUBVERSION < 1) ) )
+#ifdef USE_PPPORT_H
 #   define NEED_PL_parser
 #   define NEED_sv_2pv_flags
 #   define NEED_load_module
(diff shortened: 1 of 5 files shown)


haarg:
I am not primarily a C programmer, so there may be some mistakes here.

This is one option for handling boolean values in Storable. It's definitely a hack, but supports booleans without needing to change the format.

#19870 is an alternative approach using a more conventional format extension.

dist/Storable/Storable.xs
```diff
@@ -5650,8 +5664,16 @@ static SV *get_lstring(pTHX_ stcxt_t *cxt, UV len, int isutf8, const char *cname
...
+    /* 0 length strings should generally never be encoded in this 'long'
+     * format. this should only happen if artifically forced to do so, which
+     * only happens for booleans.
+     */
+    if (len == 0) {
+#ifdef sv_setbool
+        sv_setbool(sv, FALSE);
+#else
         SvPVCLEAR(sv);
+#endif
```
tonycoz, line 5676: get_lstring() is used for fetching all 3 string layouts, not just the long forms.

haarg: Well that's a pretty silly mistake. I could certainly get this working, but I'm not sure if it's really worth it. This is rather weird hack to begin with.

PR #19872, opened 2022-06-16
Third batch of locale commits

khwilliamson:
These collapse several bits of duplicate functionality in locale.c, make it easier to read, add debugging info, add new API functions for XS writers to determine if locale-related items (such as a time string) are in UTF-8, add nl_langinfo emulation for Windows, and and fix bugs that existed only on certain platforms and/or Configure options.


khwilliamson:
superseded by #19945

PR #19881, opened 2022-06-22
point out that non-permission bits for mkdir's mode are non-portable

### LINKED ISSUE #12082: mkdir in Perl 5.8.8 Fails to Set the Sticky Bit on FreeBSD 7.1
Migrated from [rt.perl.org#112760](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=112760) (status was 'open')

Searchable as RT112760$

tonycoz:
In particular the sticky bit may or may not be set by the system
mkdir() implementation.

Fixes #12082

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -4326,6 +4326,8 @@ a restrictive MODE and give the user no way to be more permissive.
 The exceptions to this rule are when the file or directory should be
 kept private (mail files, for instance).  The documentation for
 L<C<umask>|/umask EXPR> discusses the choice of MODE in more detail.
+If bits in MODE other than the permission bits are set, the result may
+be implementation defined, per POSIX 1003.1-2008.
 
 Note that according to the POSIX 1003.1-1996 the FILENAME may have any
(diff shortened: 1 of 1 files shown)


khwilliamson (approved):

pod/perlfunc.pod
```diff
@@ -4326,6 +4326,8 @@ a restrictive MODE and give the user no way to be more permissive.
 The exceptions to this rule are when the file or directory should be
 kept private (mail files, for instance).  The documentation for
 L<C<umask>|/umask EXPR> discusses the choice of MODE in more detail.
+If bits in MODE other than the permission bits are set, the result may
+be implementation defined, per POSIX 1003.1-2008.
```
jkeenan, line 4330: Would it be desirable and possible to supply a URL to the location in POSIX where this is discussed?

tonycoz: It's here: https://pubs.opengroup.org/onlinepubs/9699919799/functions/mkdir.html

I'm not sure on desirable.

jkeenan: Okay.  Thanks for the link.

PR #19886, opened 2022-06-24
[doc] perlvar.pod: Capitalize Perl

rwp0:
Capitalize Perl where it is referred to the language rather than the interpreter.

Format one direct interpreter reference as code.

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -156,7 +156,7 @@ test.  Outside a C<while> test, this will not happen.
 
 C<$_> is a global variable.
 
-However, between perl v5.10.0 and v5.24.0, it could be used lexically by
+However, between Perl v5.10.0 and v5.24.0, it could be used lexically by
 writing C<my $_>.  Making C<$_> refer to the global C<$_> in the same scope
 was then possible with C<our $_>.  This experimental feature was removed and is
 now a fatal error, but you may encounter it in older code.
@@ -203,13 +203,13 @@ this variable, doing so is generally discouraged, although it can be
 invaluable for some testing purposes.  It will be reset automatically
 across C<fork()> calls.
 
-Note for Linux and Debian GNU/kFreeBSD users: Before Perl v5.16.0 perl
+Note for Linux and Debian GNU/kFreeBSD users: Before Perl v5.16.0 Perl
 would emulate POSIX semantics on Linux systems using LinuxThreads, a
 partial implementation of POSIX Threads that has since been superseded
 by the Native POSIX Thread Library (NPTL).
 
 LinuxThreads is now obsolete on Linux, and caching C<getpid()>
-like this made embedding perl unnecessarily complex (since you'd have
+like this made embedding Perl unnecessarily complex (since you'd have
 to manually update the value of $$), so now C<$$> and C<getppid()>
 will always return the same values as the underlying C library.
 
@@ -259,12 +259,12 @@ to ps(1) (assuming the operating system plays along).  Note that
 the view of C<$0> the other threads have will not change since they
 have their own copies of it.
 
-If the program has been given to perl via the switches C<-e> or C<-E>,
+If the program has been given to C<perl> via the switches C<-e> or C<-E>,
 C<$0> will contain the string C<"-e">.
 
-On Linux as of perl v5.14.0 the legacy process name will be set with
+On Linux as of Perl v5.14.0 the legacy process name will be set with
 C<prctl(2)>, in addition to altering the POSIX name via C<argv[0]> as
-perl has done since version 4.000.  Now system utilities that read the
+Perl has done since version 4.000.  Now system utilities that read the
 legacy process name such as ps, top and killall will recognize the
 name you set when assigning to C<$0>.  The string you supply will be
 cut off at 16 bytes, this is a limitation imposed by Linux.
@@ -476,10 +476,10 @@ sanity-checked.
 
 The C<$OLD_PERL_VERSION> form was added in Perl v5.20.0 for historical
 reasons but its use is discouraged. (If your reason to use C<$]> is to
-run code on old perls then referring to it as C<$OLD_PERL_VERSION> would
+run code on old Perls then referring to it as C<$OLD_PERL_VERSION> would
 be self-defeating.)
 
-Mnemonic: Is this version of perl in the right bracket?
+Mnemonic: Is this version of Perl in the right bracket?
 
 =item $SYSTEM_FD_MAX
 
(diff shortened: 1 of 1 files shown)


pod/perlvar.pod
```diff
@@ -203,13 +203,13 @@ this variable, doing so is generally discouraged, although it can be
...
 across C<fork()> calls.
 
-Note for Linux and Debian GNU/kFreeBSD users: Before Perl v5.16.0 perl
+Note for Linux and Debian GNU/kFreeBSD users: Before Perl v5.16.0 Perl
 would emulate POSIX semantics on Linux systems using LinuxThreads, a
 partial implementation of POSIX Threads that has since been superseded
 by the Native POSIX Thread Library (NPTL).
 
 LinuxThreads is now obsolete on Linux, and caching C<getpid()>
-like this made embedding perl unnecessarily complex (since you'd have
```
richardleach, line 212: AFAICT this is literally talking about embedding the perl interpreter, so this instance should not be capitalized.

pod/perlvar.pod
```diff
@@ -476,10 +476,10 @@ sanity-checked.
 
 The C<$OLD_PERL_VERSION> form was added in Perl v5.20.0 for historical
 reasons but its use is discouraged. (If your reason to use C<$]> is to
-run code on old perls then referring to it as C<$OLD_PERL_VERSION> would
```
richardleach, line 479: This seems to definitely be referring to version of the interpreter, not the language.

pod/perlvar.pod
```diff
@@ -773,15 +773,15 @@ The name used to execute the current copy of Perl, from C's
 C<argv[0]> or (where supported) F</proc/self/exe>.
 
 Depending on the host operating system, the value of C<$^X> may be
-a relative or absolute pathname of the perl program file, or may
-be the string used to invoke perl but not the pathname of the
-perl program file.  Also, most operating systems permit invoking
+a relative or absolute pathname of the C<perl> program file, or may
+be the string used to invoke Perl but not the pathname of the
```
richardleach, line 777: It is the interpreter that is invoked here.

pod/perlvar.pod
```diff
@@ -578,7 +578,7 @@ Then
 
 would allocate a 64K buffer for use in an emergency.  See the
 L<INSTALL> file in the Perl distribution for information on how to
-add custom C compilation flags when compiling perl.  To discourage casual
```
richardleach, line 581: This is definitely talking about compiling the interpreter, not about compiling Perl code using the interpreter. The existing capitalization should stay.

pod/perlvar.pod
```diff
@@ -773,15 +773,15 @@ The name used to execute the current copy of Perl, from C's
...
-perl program file.  Also, most operating systems permit invoking
+a relative or absolute pathname of the C<perl> program file, or may
+be the string used to invoke Perl but not the pathname of the
+Perl program file.  Also, most operating systems permit invoking
 programs that are not in the PATH environment variable, so there
 is no guarantee that the value of C<$^X> is in PATH.  For VMS, the
 value may or may not include a version number.
 
 You usually can use the value of C<$^X> to re-invoke an independent
-copy of the same perl that is currently running, e.g.,
```
richardleach, line 784: Again, an independent copy of the same _perl interpreter_

pod/perlvar.pod
```diff
@@ -1779,9 +1779,9 @@ variables:
         close $pipe or die "bad pipe: $?, $!";
     };
 
-When perl executes the C<eval()> expression, it translates the
```
richardleach, line 1782: The interpreter executes expressions.

pod/perlvar.pod
```diff
@@ -2170,7 +2170,7 @@ Each time a statement completes being compiled, the current value of
 C<$^H> is stored with that statement, and can later be retrieved via
 C<(caller($level))[8]>.
 
-When perl begins to parse any block construct that provides a lexical scope
```
richardleach, line 2173: The interpreter does the parsing

pod/perlvar.pod
```diff
@@ -2337,7 +2337,7 @@ Reflects if taint mode is on or off.  1 for on (the program was run with
 B<-T>), 0 for off, -1 when only taint warnings are enabled (i.e. with
 B<-t> or B<-TU>).
 
-Note: if your perl was built without taint support (see L<perlsec>),
```
richardleach, line 2340: This sentence is implicitly "If your [interpreter ] was built without.."

pod/perlvar.pod
```diff
@@ -2347,9 +2347,9 @@ This variable was added in Perl v5.8.0.
 =item ${^SAFE_LOCALES}
 X<${^SAFE_LOCALES}>
 
-Reflects if safe locale operations are available to this perl (when the
+Reflects if safe locale operations are available to this Perl (when the
 value is 1) or not (the value is 0).  This variable is always 1 if the
-perl has been compiled without threads.  It is also 1 if this perl is
```
richardleach, line 2352: It's the interpreter that gets compiled with or without threads.

richardleach (commented):
I'm not sufficiently familiar with the naming conventions to be certain about all the changes in this PR, but many instances clearly (IMHO) do refer to the interpreter, not the language, and therefore should not be changed. I've called out some specific instances in review comments.

rwp0:
I think `perl` means the interpreter command, the executable binary referred to with its switches and arguments.

I also think it should always be formatted as code to avoid confusion, and that the other references are about the language, and should therefore be capitalized as a proper name.

I might be wrong but I submitted the PR having that view in mind.

I'm for capitalizing Perl where possible.

------

Some examples:

> I  pass `PERL5LIB` environment variable to `perl` in my projects

> I embed Perl in my projects

pod/perlvar.pod
rwp0: I think it's about embedding Perl as a language, not as a  particular interpreter.

pod/perlvar.pod
rwp0: they may coincide, but usually the version is of the language

pod/perlvar.pod
rwp0: hmm.. I could `C<perl>` here as well but it'd be repeated and unnecessary, though invoking a language isn't something wrong

pod/perlvar.pod
rwp0: yeah, then it should be written as compiling the **Perl interpreter**?

richardleach:
I don't disagree with your point about formatting `perl` as code, but we clearly have different opinions about whether it is the language or interpreter being referred to your PR. I'll defer to committers views.

Grinnz:
Other than "old perls", I agree with @richardleach 's interpretation: where it is talking about perl as an entity that executes code, that refers to the interpreter, not the language which is a set of behaviors.

pod/perlvar.pod
Grinnz: I agree this instance is referring to the version of the Perl language.

PR #19893, opened 2022-06-25
Update STMT_START..END guidance

khwilliamson:
My mental model of how the C preprocessor did things was flawed.  This
commit makes things clearer.  It needs some polishing

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -740,110 +740,27 @@ as in
  if (x) STMT_START { ... } STMT_END else ...
 
 Note that you can't return a value out of this construct and cannot use it as
-an operand to the comma operator, which limit its utility.  If you need
-those features, instead use C language C<static inline> functions.
-But those won't work where the types of the inputs can vary depending on where
-called from; you'd have to create a separate function for each combination of
-parameter types.
+an operand to the comma operator.  These limit its utility.  
 
-Subtle bugs can occur due to this being a macro with names evaluated in
-the context of the caller, unlike a function.  These bugs can be eliminated by
-treating this more like a function.  This means that you must only use
-variables inside it that are either formal parameters or are declared
-internally within the construct.  Thus
+But, a value could be returned by constructing the API so that a pointer is
+passed and the macro dereferences this to set the return.  If the value can be
+any of various types, depending on context, you can handle that situation in
+some situations by adding the type of the return as an extra accompanying
+parameter:
 
-=over
+ #define foo(param, type)  STMT_START {
+                              type * param; *param = do_calc; ...
+                           } STMT_END
 
-=item *
+This could be awkward, so consider instead using a C language C<static inline>
+function.
 
-B<All> references to outside-the-construct variables must be through formal
-parameters.
-
-=item *
-
-All formal parameters should be used only once.  If a value needs to be used in
-more than one place, instead declare a variable internally, initialized with
-the formal parameter.  Then use that internal variable freely.  For example
-
- #define warn_if_odd(n)                                     \
-                STMT_START {                                \
-                    int n_ = n;                             \
-                                                            \
-                    if (n_ % 2 != 0) {                      \
-                        my_warn(aTHX_ "%d is odd\n", n_);   \
-                    }                                       \
-                } STMT_END
-
-This prevents issues when the macro is called with a parameter with side
-effects.
-
-Note that the internal variable must have a different name than the formal
-parameter one.  Using a trailing underscore is legal in C, and unobtrusive.
-C<But> see below for why you should choose a different naming convention if
-there are multiple parameters.
-
-=item *
-
-If there are multiple formal parameters, the names of the internal copies need
-to be something that won't clash with any names outside the macro.  This
-prevents subtle bugs when the caller uses one parameter in terms of another.
-
- #define my_best_shot(a, b)                                 \
-                STMT_START {                                \
-                    UV     my_best_shot_a = a;              \
-                    UV     my_best_shot_b = b;              \
-                                                            \
-                    Size_t ix;                              \
-                                                            \
-                    ...                                     \
-                                                            \
-                } STMT_END
-
- my_best_shot(a + b + ix, a - b - ix);
-
-Without the declarations, the macro call at the end of the example would be
-problematic.
-
-This example uses the macro name as part of each variable name that corresponds
-to a formal parameter.  If this convention is followed everywhere, there is no
-possibility of name collisions.
-
-Note that the C<ix> within the construct doesn't interfere with the outside
-C<ix>, but only because we have declared and initialized variables that are
-copies of all the formal parameters.
-
-Also note that if you forget and use plain C<a> or C<b> in some places, things
-may well compile, and have subtle run-time bugs.
-
-For a concrete example, see L<https://perlmonks.org/?node_id=11144355>.
-
-=item *
-
-Pass in the type for any declarations if it isn't known at macro definition
-time.  Suppose that the parameters to C<my_best_shot> in the example above
-didn't have to be UV.  This can be handled by changing the signature, something
-like this:
-
- #define my_best_shot(a, a_type, b, b_type)                 \
-                STMT_START {                                \
-                    a_type my_best_shot_a = a;              \
-                    b_type my_best_shot_b = b;              \
-                                                            \
-                    Size_t ix;                              \
-                                                            \
-                    ...                                     \
-                                                            \
-                } STMT_END
-
- my_best_shot(a + b + ix, int, a - b - ix, long);
-
-You are out of luck if the signature can't be changed to include the types.
-Best then is to document that the macro cannot safely be called with parameters
-with side effects.  And any internally-declared variables should have names
-that won't clash with any outside names.  C<ix> should then be
-C<my_best_shot_ix>, for example.
-
-=back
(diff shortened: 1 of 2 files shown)


perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
-                STMT_START {                                \
-                    int n_ = n;                             \
-                                                            \
-                    if (n_ % 2 != 0) {                      \
-                        my_warn(aTHX_ "%d is odd\n", n_);   \
-                    }                                       \
-                } STMT_END
+If the parameter is a tied variable or otherwise has side effects, that can
+present a problem, as it's likely the caller is expecting any side effect to
+happen just once per call.
```
hvds: >If the parameter is a tied variable or otherwise has side effects, that can
>present a problem, as it's likely the caller is expecting any side effect to
>happen just once per call.

This is C code, so "tied variable" is meaningless. I'd suggest:

  If the parameter is an expression with side effects, such as C<i++>, that can
  present a problem - since macro invocations look just like function calls, the
  caller may well expect any side effect to happen just once per call.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
+ #define my_best_shot(a, b)     ...
 
-=item *
+ my_best_shot(a + b + ix, a - b - ix);
 
-If there are multiple formal parameters, the names of the internal copies need
-to be something that won't clash with any names outside the macro.  This
-prevents subtle bugs when the caller uses one parameter in terms of another.
+This cross referring can cause unexpected substitutions when the macro is
+expanded.
```
hvds: >If the macro has multiple parameters, the call could have some of the actual
>parameters be expressions containing another parameter:
>
> #define my_best_shot(a, b)     ...
>
> my_best_shot(a + b + ix, a - b - ix);
>
>This cross referring can cause unexpected substitutions when the macro is
>expanded.

Can you give an example of such an unexpected substitution? As far as I understand it, any reference to `a` or `b` in the macro body will be stringwise replaced by those expressions - the `a` and `b` as formal parameter names don't appear at all in what the preprocessor hands to the compiler.

If this can't actually cause problems it would be better to remove reference to this issue rather than confuse matters.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
-                } STMT_END
+Note that it would be a bad idea in the above to declare just plain C<ix>.
 
- my_best_shot(a + b + ix, a - b - ix);
+        Size_t  ix;     // WRONG!
 
-Without the declarations, the macro call at the end of the example would be
-problematic.
+A caller could have done S<C<#define ix  0>>, and then our statement above
+would expand to
```
hvds: >A caller could have done S<C<#define ix  0>>, and then our statement above
>would expand to
>    Size_t 0;

That would give a compile-time error, an example that goes wrong silently might be more convincing.

But I'm also not sure that trying to extend this document to cover protection against arbitrary macros is wise: I'm nervous even about the earlier mention of the `#define else ...` possibility, even if only to discount it. I think the variable masking issue is clear, and enough to motivate standardized longer variable names in macros - we don't need to list every other case it could protect against.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
+        Size_t  0;
 
-Note that the C<ix> within the construct doesn't interfere with the outside
-C<ix>, but only because we have declared and initialized variables that are
-copies of all the formal parameters.
+Besides this initialization, don't use any variable inside the macro that isn't
+also declared inside it.  If this were a function that would be a syntax error.
+But in a macro it would use any macro of the same name closest in the call
+chain.  Even if successful, that makes it harder to understand what the code is
+doing.
```
hvds: I'm all for recommending this, but if it becomes a requirement (as implied by the wording here) we have a lot of code that will need very substantive changes.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
-may well compile, and have subtle run-time bugs.
+One thing a macro can do that a function can't is to not have to declare the
+type of a parameter, since the type is known from the surrounding context.  Or
+the parameter doesn't have to be a variable; it could be some code to insert,
+for example.  Those circumstances may mean that the techniques here won't be
+available.
 
-For a concrete example, see L<https://perlmonks.org/?node_id=11144355>.
+But, the declaration/initialization can be skipped if both the following are
+true:
```
hvds: >But, the declaration/initialization can be skipped if both the following are
>true:

I think this section may need some rewriting. We probably need to distinguish between a parameter that we want to treat as a value - i.e. treating it like the parameter to a function call - and one where we are more specifically taking advantage of the stringy nature of the preprocessor, eg to accept a chunk of code that we may invoke conditionally (as in the `act_on_oob` example), to use it in preprocessor string concatenation, or to use it as the name of a type.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
+Also, if the parameter does have a type, the macro signature can include an
+extra parameter giving the type.
+
+ #define warn_if_odd(n, type)                               \
+    STMT_START {                                            \
+        type n__warn_if_odd = n;                            \
                                                             \
-                } STMT_END
+        if (n__warn_if_odd % 2 != 0) {                      \
+            my_warn(aTHX_ "%d is odd\n", n__warn_if_odd);   \
```
hvds: That's likely to be a printf format error. :)

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
 
-You are out of luck if the signature can't be changed to include the types.
-Best then is to document that the macro cannot safely be called with parameters
-with side effects.  And any internally-declared variables should have names
-that won't clash with any outside names.  C<ix> should then be
-C<my_best_shot_ix>, for example.
+Note that C<type> didn't have to have a disambiguating name.  It would still
+work even if somewhere in the calling change C<type> was defined to be
+something; or if the call to the macro passed C<type> as the first positional
+parameter.
```
hvds: >It would still
>work even if somewhere in the calling change C<type> was defined to be
>something; or if the call to the macro passed C<type> as the first positional
>parameter.

I don't understand either part of this sentence.

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
 
 =back
 
+Following these rules should bomb-proof your macro against outside influences.
```
hvds: Now _there's_ a bold claim. Probably too bold, especially if you include hostile macros among those influences.

perl.h
khwilliamson: done

perl.h
```diff
@@ -737,110 +737,156 @@ as in
...
+One possible technique is to include the name of the macro as part of each
+variable name.  Suppose you have a macro named C<Foo_Bar> that has two
+variables that would normally be named C<sv> and C<len>.  You could instead
+name them C<sv__foo_bar> and C<len__foo_bar>, respectively, and pretty much
+guarantee that those names won't ever clash with anything else.  In this
+example naming convention, the macro name is appended to the variable name with
+two underscores, which makes the variable name stand out more, and indicates
+that the macro name is a modifier, and not the important part.  It also
+lowercases the macro name to make the result more like a typical variable name;
+this assumes that there isn't another macro named C<foo_bar>.
```
tonycoz: It's not an issue for C, but since we build with C++: "Each identifier that contains a double underscore __ or begins with an underscore followed by an uppercase letter is reserved to the implementation for any use."

tonycoz:
Does most of this belong in perlhacktips.pod rather than perlapi?  Most of it isn't specific to STMT_START/END.

pod/perlhacktips.pod
```diff
@@ -309,6 +309,123 @@ C<-xc99>
...
+
+This is harder to read than plain C<foo>, but it is pretty much
+guaranteed that a caller will never naively use C<foo_svpv_> (and run
+into problems).  (The lowercasing makes it clearer that this is a
+variable, but assumes that there won't be two elements whose names
+differ only in the case of their letters.)  The trailing underscore
+makes it even more unlikely to clash, as those, by convention, signify a
+private variable name.  (Note that C reserves, for its own use, variable
+names containing two underscores in a row, or beginning with an
+underscore.  It's undefined behavior for anyone else to do that.)
```
demerphq: A really common solution in our code is to variables that AlTeRnAtInG case. I have not seen trailing underbars used often but i have seen the alternating case form used a bunch. I think Nick Clark liked the style.

demerphq (approved):
I think you could mention the AlTeRnAtInG case form. But i like this.

pod/perlhacktips.pod
khwilliamson: I'm hesitant to mention this.  I find those difficult to read myself; having to stop and unfocus my eyes so as to read the whole word in a slurp.  I read once that non-native-English speakers have a hard time with PascalCase in general.

And it doesn't solve the problem of macro X calling macro Y and both use the same variable name.  That requires something unique to the instance.  The solution cited is easier to read than ALtErNaTeCaSe, and works except if X gets called recursively.  Thus it is both easier to read and works in more cases.

I've thought about some way to use the line number in the file of the declare statement as part of the the variable name, like we do when we emulate static assertions.  But haven't come up with a good solution,.

perl.h
khwilliamson: revised

perl.h
khwilliamson: revised

perl.h
khwilliamson: done

Issue #19857, opened 2022-06-13
perl: op.c:14769: Perl_find_lexical_cv: Assertion `PARENT_PAD_INDEX(name)' failed.


tonycoz:
Module:

**Description**

Running this code:
```
use builtin 'ceil';

my $x;
BEGIN {
    my $e = <<'EOS';
sub { ceil($_[0]) }
EOS
    $x = eval $e;
}
```

in a DEBUGGING build throws:

```
$ ./perl -Ilib ../ceilassert.pl 
perl: op.c:14769: Perl_find_lexical_cv: Assertion `PARENT_PAD_INDEX(name)' failed.
Aborted
```

Backtrace:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c65537 in __GI_abort () at abort.c:79
#2  0x00007ffff7c6540f in __assert_fail_base (
    fmt=0x7ffff7dce128 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x55555594579a "PARENT_PAD_INDEX(name)", 
    file=0x55555593e331 "op.c", line=14769, function=<optimized out>)
    at assert.c:92
#3  0x00007ffff7c74662 in __GI___assert_fail (
    assertion=0x55555594579a "PARENT_PAD_INDEX(name)", 
    file=0x55555593e331 "op.c", line=14769, 
    function=0x5555559482c0 <__PRETTY_FUNCTION__.33> "Perl_find_lexical_cv")
    at assert.c:101
#4  0x00005555555dd638 in Perl_find_lexical_cv (off=1) at op.c:14769
#5  0x0000555555660eee in yyl_keylookup (s=0x555555c09a2a "($_[0]) }\n\n;", 
    gv=0x0) at toke.c:8805
#6  0x0000555555663a6e in yyl_try (s=0x555555c09a26 "ceil($_[0]) }\n\n;")
    at toke.c:9319
#7  0x0000555555664ed9 in Perl_yylex () at toke.c:9621
#8  0x000055555567ab7e in Perl_yyparse (gramtype=258) at perly.c:357
#9  0x0000555555851412 in S_doeval_compile (gimme=2 '\002', 
    outside=0x555555bf6e08, seq=201, hh=0x0) at pp_ctl.c:3613
#10 0x0000555555858fd3 in Perl_pp_entereval () at pp_ctl.c:4580
[...]

PR #19895, opened 2022-06-27
Add builtin::export_lexically

leonerd:
Adds a `builtin::` function for performing the lexical-export behaviour, so that new code can do the same trickery that core `builtin::` itself achieves.

### DIFF
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -358,6 +358,41 @@ TODO: {
     is($storecount, 1, 'is_tainted() invokes STORE magic');
 }
 
+# Lexical export
+{
+    my $name;
+    BEGIN {
+        use builtin qw( export_lexically );
+
+        $name = "message";
+        export_lexically $name => sub { "Hello, world" };
+    }
+
+    is(message(), "Hello, world", 'Lexically exported sub is callable');
+    ok(!__PACKAGE__->can("message"), 'Exported sub is not visible via ->can');
+
+    is($name, "message", '$name argument was not modified by export_lexically');
+
+    our ( $scalar, @array, %hash );
+    BEGIN {
+        use builtin qw( export_lexically );
+
+        export_lexically
+            '$SCALAR' => \$scalar,
+            '@ARRAY'  => \@array,
+            '%HASH'   => \%hash;
+    }
+
+    $::scalar = "value";
+    is($SCALAR, "value", 'Lexically exported scalar is accessible');
+
+    @::array = ('a' .. 'e');
+    is(scalar @ARRAY, 5, 'Lexically exported array is accessible');
+
+    %::hash = (key => "val");
+    is($HASH{key}, "val", 'Lexically exported hash is accessible');
+}
+
 # vim: tabstop=4 shiftwidth=4 expandtab autoindent softtabstop=4
 
 done_testing();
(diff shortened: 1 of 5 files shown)


lib/builtin.t
```diff
@@ -342,6 +342,37 @@ TODO: {
...
+    }
+
+    is(message(), "Hello, world", 'Lexically exported sub is callable');
+    ok(!__PACKAGE__->can("message"), 'Exported sub is not visible via ->can');
+
+    our ( $scalar, @array, %hash );
+    BEGIN {
+        use builtin qw( lexically_export );
+
+        # TODO: support a list
```
ilmari: It does.

pod/perldiag.pod
```diff
@@ -2275,6 +2275,12 @@ as a goto, or a loop control statement.
 (W exiting) You are exiting a substitution by unconventional means, such
 as a return, a goto, or a loop control statement.
 
+=item Expected a %s reference in lexically_export
```
ilmari: Other occurrences of `%s reference` in `perldiag` leave the article as part of the `%s`.

pod/perldiag.pod
```diff
@@ -2275,6 +2275,12 @@ as a goto, or a loop control statement.
...
+
+(F) A call to L<builtin/lexically_export> associates new names with items
+given by reference.  Every second argument must be a reference to a kind
+of item that matches the sigil in the name it is to be associated with.
```
ilmari: The general style is to first explain what you did wrong, then explain why. Maybe something like:

> The type of a reference given to L<builtin/lexically_export> did not match the sigil of the preceding name.

pod/perldiag.pod
```diff
@@ -3345,6 +3351,14 @@ The number of items in a hash can be obtained by doing:
 
     scalar(keys %hash);
 
+=item lexically_export can only be called at compile time
+
+(F) A call to L<builtin/lexically_export> creates new names in the lexical
+scope of the code currently being compiled.  This function can therefore
+only be called when a scope is being compiled (i.e. from a C<BEGIN> block
+in that scope).  If there is no code currently being compiled, this function
+will have no effect and so calling it is not permitted.
```
ilmari: Same here. How about something like this?

> (F) L<builtin/lexically_export> was called at runtime. Because it creates new names in the lexical scope
> currently being compiled, it can only be called from a C\<BEGIN\> block in that scope.

ilmari (changes requested):

leonerd:
Thanks @ilmari. Comments addressed and force-pushed.

ilmari (approved):
LGTM

tonycoz:
#19857 is still a problem here:
```
tony@venus:.../git/perl5$ ./perl -Ilib -Mbuiltin=lexically_export -e 'BEGIN { lexically_export abc => sub {}; } BEGIN { eval "abc()" }'
Built-in function 'builtin::lexically_export' is experimental at -e line 1.
perl: op.c:13404: Perl_find_lexical_cv: Assertion `PARENT_PAD_INDEX(name)' failed.
Aborted
```

ilmari:
> #19857 is still a problem here:

Seeing as that's a problem with lexical subs in general, not specifically with exporting them, I don't think that should block this.

leonerd:
P5P requests that this be expanded into a full RFC. I am now in the process of writing that. Until that's agreed, we should not merge this (but it can be referenced as a prototype implementation)

leonerd:
RFC now written and PR raised: https://github.com/Perl/RFCs/pull/20

PR #19898, opened 2022-06-28
PathTools/Cwd.xs: define SYSNAME/SYSNAME_LEN for OS390 only

mtelka:
This fixes the following warning on illumos based platforms:

    Cwd.xs:27:0: warning: "SYSNAME" redefined
     #define SYSNAME "$SYSNAME"

    In file included from ../../perl.h:1111:0,
                     from Cwd.xs:8:
    /usr/include/sys/param.h:184:0: note: this is the location of the previous definition
     #define SYSNAME 9  /* # chars in system name */

Also, both SYSNAME and SYSNAME_LEN are used later in the code only in a case
the EBCDIC is defined so it is safe to have defined them for EBCDIC only.


khwilliamson (changes requested):
It turns out that os390 can run ASCII also nowadays.  We have just converted a bunch of EBCDIC tests to use `#ifdef OS390` so this should be that too, and the later use you refer to.  Otherwise, this looks fine to me

mtelka:
> It turns out that os390 can run ASCII also nowadays. We have just converted a bunch of EBCDIC tests to use `#ifdef OS390` so this should be that too, and the later use you refer to. Otherwise, this looks fine to me

Done.

PR #19901, opened 2022-06-28
B::Concise - when private flags are really the kid count

richardleach:
It's not apparent what the value "2" represents in the following example:
`sassign vKS/2`
as it doesn't seem to be a valid private flag set via _regen/op_private_.
The number of kids therefore seems to be the most likely thing it represents.

**I'm not 100% certain if this is correct, so please could the committer check!**

### DIFF
--- a/ext/B/B/Concise.pm
+++ b/ext/B/B/Concise.pm
@@ -14,7 +14,7 @@ use warnings; # uses #3 and #4, since warnings uses Carp
 
 use Exporter 'import'; # use #5
 
-our $VERSION   = "1.006";
+our $VERSION   = "1.007";
 our @EXPORT_OK = qw( set_style set_style_standard add_callback
 		     concise_subref concise_cv concise_main
 		     add_style walk_output compile reset_sequence );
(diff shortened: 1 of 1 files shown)


ext/B/B/Concise.pm
```diff
@@ -1489,6 +1489,10 @@ They're opcode specific, and occur less often than the public ones, so
 they're represented by short mnemonics instead of single-chars; see
 B::Op_private and F<regen/op_private> for more details.
 
+Note that a number after a '/' might refer to the number of child OPs
+if the OP has not been privatized by a check routine. (e.g. sassign
+has 2 child OPs in the example above.)
+
```
tonycoz: I don't what it means for an op to be "privatized".

richardleach: Neither do I. That terminology is straight out of op.h:
````
 *	op_private	Flags peculiar to a particular operation (BUT,
 *			by default, set to the number of children until
 *			the operation is privatized by a check routine,
 *			which may or may not check number of children).
 ````

richardleach: That description of `op_private` goes back at least as far as [perl 5.0 alpha 2](https://github.com/Perl/perl5/commit/79072805bf63abe5b5978b5928ab00d360ea3e7f)

tonycoz:
Some ops have the argument count set in op_private even after the ck is done, see the beginning of the flag definitions in `regen/op_private` (which also discusses this in more detail).  Any OP that uses MAXARG (and there's a few) is using the argument count from op_private.

B::Concise specifically uses the number when the label is `-`, and only the argument count fields have label set to `-`.

So the note you're adding can be more definitive - the number after / is the argument count.

richardleach:
Thanks Tony. I've revised the text to mention the argument count and give MAXARG as an example.

PR #19903, opened 2022-06-29
Optimize hand-coded fabs-like operations

t-a-k:
There were some constructs like `(nv > 0 ? nv : -nv)` to compute absolute values of NVs, but it will be better to use `Perl_fabs(nv)` for them to get compacter and (slightly) faster code, just like #19127.

Furthermore, one instance of such construct could be completely optimized out by transforming overall expression into a static assertion.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -2187,11 +2187,12 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
                     SvIV_set(sv, I_V(SvNVX(sv)));
                     if ((NV)(SvIVX(sv)) == SvNVX(sv))
                         SvIOK_on(sv);
-                    /* Assumption: first non-preserved integer is < IV_MAX,
-                       this NV is in the preserved range, therefore: */
-                    if (!(U_V(Perl_fabs(SvNVX(sv))) < (UV)IV_MAX)) {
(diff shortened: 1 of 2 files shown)


khwilliamson (approved):
But I'm not an expert here

sv.c
```diff
@@ -2187,11 +2187,12 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
...
-                    /* Assumption: first non-preserved integer is < IV_MAX,
-                       this NV is in the preserved range, therefore: */
-                    if (!(U_V(Perl_fabs(SvNVX(sv))) < (UV)IV_MAX)) {
-                        Perl_croak(aTHX_ "sv_2iv assumed (U_V(Perl_fabs(SvNVX(sv))) < (UV)IV_MAX) but SvNVX(sv)=%" NVgf " U_V is 0x%" UVxf ", IV_MAX is 0x%" UVxf "\n", SvNVX(sv), U_V(Perl_fabs(SvNVX(sv))), (UV)IV_MAX);
-                    }
+                    /* There had been runtime checking for
+                       "U_V(Perl_fabs(SvNVX(sv))) < (UV)IV_MAX" here to ensure
+                       that this NV is in the preserved range, but this should
+                       be always true if the following assertion is true: */
+                    STATIC_ASSERT_STMT(((UV)1 << NV_PRESERVES_UV_BITS) <=
```
tonycoz, line 2194: For 64-bit UV and long doubles, NV_PRESERVES_UV_BITS is 64, since that's the size of the left operand this produces undefined behaviour.

t-a-k: In such a case (where NV_PRESERVES_UV_BITS &ge; number of bits in UV),  `NV_PRESERVES_UV` should be defined so that this static assertion is not compiled anyway.

tonycoz: You're right, I should have looked outside the context to see the `#ifdef`

PR #19905, opened 2022-06-29
Rename token types for keywords to add KW_... prefix

leonerd:
Some of the token types represent simple keywords; some of them do not. It's easier to read and work out what's going on if all the simple keyword ones have a common prefix; `KW_...` in this case.

Additionally I've renamed the four `sub`-related keywords to have a bit more structure to them.

Also added comments.


leonerd:
I should add: This is mostly just general tidying up, but it's also useful ahead of my work to add a `method` keyword; because it will have to follow a similar structure for `KW_SUB`

leonerd:
cygwin failed for unrelated reasons - our dear friend `t/op/alarm.t`

khwilliamson (approved):

PR #19908, opened 2022-06-30
Create and use a `HvHasAUX` predicate macro

leonerd:
Previously, the `SvOOK` macro and associated flag was (ab)used to indicate whether an HV has the AUX structure on it.

This PR creates a new macro to use with HVs, rather than confusingly checking `SvOOK`. The new flag macro occupies the same bit position, as we can't easily change that without potentially breaking CPAN modules, etc... But at least this way, core code has a clearer intent when reading it.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -1206,7 +1206,7 @@ Perl_hv_auxalloc(pTHX_ HV *hv) {
 
     PERL_ARGS_ASSERT_HV_AUXALLOC;
     assert(SvTYPE(hv) == SVt_PVHV);
-    assert(!SvOOK(hv));
+    assert(!HvHasAUX(hv));
 
 #ifdef PURIFY
     new_body = new_NOARENAZ(&fake_hv_with_aux);
@@ -1230,7 +1230,7 @@ Perl_hv_auxalloc(pTHX_ HV *hv) {
 #endif
 
     SvANY(hv) = (XPVHV *) new_body;
-    SvOOK_on(hv);
+    SvFLAGS(hv) |= SVphv_HasAUX;
     return HvAUX(hv);
 }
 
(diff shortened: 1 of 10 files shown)


richardleach:
This makes sense to me and I can't see anything glaringly wrong with the changes. (Minor comments added in a couple of places.)

sv.c
```diff
@@ -1230,7 +1230,7 @@ Perl_hv_auxalloc(pTHX_ HV *hv) {
 #endif
 
     SvANY(hv) = (XPVHV *) new_body;
-    SvOOK_on(hv);
+    SvFLAGS(hv) |= SVphv_HasAUX;
```
richardleach, line 1233: Is it worth creating `HvHasAUX_on` and related macros for consistency and simplicity?

sv.h
```diff
@@ -422,9 +422,8 @@ perform the upgrade if necessary.  See C<L</svtype>>.
                                            including PVLV-as-regex. See
                                            isREGEXP().
                                        */
-#define SVf_OOK		0x02000000  /* has valid offset value. For a PVHV this
-                                       means that a hv_aux struct is present
-                                       after the main array */
+#define SVf_OOK		0x02000000  /* has valid offset value */
+#define SVphv_HasAUX    SVf_OOK     /* PVHV has an hv_aux struct after the main array */
```
richardleach: I'm not clear on what the "main array" means here. Seems like the `xpvhv_aux` struct is the final member of the `xpvhv_with_aux` struct, which will be allocated from an appropriately-sized xpv arena?

sv.c
leonerd: I thought about it, but as this is the one line of all the source code that actually sets the flag, I decided it wasn't worth it. Also no code ever removes it; once it's set it remains there until the HV is destroyed.

sv.h
leonerd: I'll be honest - I have no idea myself. I just cloned the wording from the previous version of the comment :) I guess I could just remove the words "after the main array" - it's still clear without that.

leonerd:
Commits somewhat squashed, and wording of comment updated.

khwilliamson (approved):

PR #19911, opened 2022-06-30
add SO_INCOMING_CPU and SO_INCOMING_NAPI_ID socket options constants to Socket module

Yaribz:
SO_INCOMING_CPU (gettable since Linux 3.19, settable since Linux 4.4)
SO_INCOMING_NAPI_ID (gettable since Linux 4.12)


Leont:
This dist is maintained upstream, and this issue should be filed in its [own bugtracker](https://rt.cpan.org/Dist/Display.html?Name=Socket)

leonerd:
I've now added these to Socket 2.035; coming soon to CPAN

PR #19913, opened 2022-07-01
perl.h: Add %format strings for 32 bit quantities

khwilliamson:
In order to portably print a variable declared as I32 or U32, it is
usually cast to (IV) or (UV) and then the %Ivdf, etc format is used.  We
can avoid needing these casts by creating formats for just the I32 and
U32 variables, which is what this commit does.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -2414,6 +2414,57 @@ typedef UVTYPE UV;
 #endif
 
 /*
+=for apidoc   AmnU||U32of
+=for apidoc_item  ||U32uf
+=for apidoc_item  ||U32xf
+=for apidoc_item  ||U32Xf
+
+These symbols define the format strings used for printing variables declared
+as U32; respectively as octal, unsigned, hex (lowercase C<a-f>), and hex
+(uppercase C<a-f>).
+
+=cut
+*/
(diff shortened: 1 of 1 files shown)


regcomp.c
```diff
@@ -21322,7 +21321,7 @@ S_regtail_study(pTHX_ RExC_state_t *pRExC_state, regnode_offset p,
         DEBUG_PARSE_MSG("");
         regprop(RExC_rx, RExC_mysv, REGNODE_p(val), NULL, pRExC_state);
         Perl_re_printf( aTHX_
-                      "~ attach to %s (%" IVdf ") offset to %" IVdf "\n",
+                      "~ attach to %s (%zd) offset to %td\n",
                       SvPV_nolen_const(RExC_mysv),
                       (IV)val,
                       (IV)(val - scan)
```
tonycoz: The formats were changed here, but the casts weren't removed.

khwilliamson: Hmm.  That wasn't supposed to be part of this commit.  I just wanted to make sure no one had a problem with the definitions I came up with, before in later commits starting to use them.

Issue #19661, opened 2022-04-24
ExtUtils-ParseXS-3.44: xsubpp seems to output false-positive warnings


twata1:
Module:
ExtUtils-ParseXS-3.44

**Description**

Hello,

The following link is the background.

[DBD-MariaDB-1.22: Build warnings](https://github.com/perl5-dbi/DBD-MariaDB/issues/169)

As shown below, the **xsubpp** command outputs warnings when I run the **gmake** command.
According to https://github.com/perl5-dbi/DBD-MariaDB/issues/169#issuecomment-1107666139 here, it seems to be a false positive warnings.

```
Warning: duplicate function definition 'do' detected in MariaDB.xs, line 104
Warning: duplicate function definition 'rows' detected in MariaDB.xs, line 229
Warning: duplicate function definition 'last_insert_id' detected in MariaDB.xs, line 248
```

**Steps to Reproduce**

0. Start the mysqld in advance for testing.
```
C:\mysql\bin>mysqld --defaults-file="..\my.ini" --console
```
1. Install ExtUtils::ParseXS 3.44 beforehand. (The following command confirms ExtUtils::ParseXS 's version.)
```
C:\home\sunlight2\DBD-MariaDB-1.22>module-version ExtUtils::ParseXS

The version of ExtUtils::ParseXS in C:\home\sunlight2\perl5\lib\perl5 is 3.44
```
2. Download DBD-MariaDB-1.22.tar.gz from [metacpan ](https://metacpan.org/dist/DBD-MariaDB)and extract it and run ``make Makefile.PL`` and ``gmake``.
```Perl
C:\home\sunlight2\DBD-MariaDB-1.22>perl Makefile.PL


PLEASE NOTE:

For 'make test' to run properly, you must ensure that the
database user 'root' can connect to your MariaDB or MySQL server
and has the proper privileges that these tests require such
as 'drop table', 'create table', 'drop procedure', 'create procedure'
as well as others.

mysql> grant all privileges on test.* to 'root'@'localhost' identified by 's3kr1t';

You can also optionally set the user to run 'make test' with:

perl Makefile.PL --testuser=username

I will use the following settings for compiling and testing:

  cflags       (mysql_config) = -IC:\Strawberry\c\bin\..\include\mysql50716
  libs         (mysql_config) = -LC:\Strawberry\c\bin\..\lib -lmysql
  mysql_config (guessed     ) = mysql_config
  testdb       (default     ) = test
  testhost     (default     ) =
[...]

PR #19920, opened 2022-07-04
Smoke me/jkeenan/parsexs gh19661 20220704

jkeenan:


jkeenan:
This p.r. responds (somewhat) to the issues raised in https://github.com/Perl/perl5/issues/19661.  The first commit is adapted from a test file suggested by @choroba in that ticket; the other commits are my extensions of that line of thought.

This p.r. does add a print-out of a TODO test when you run `make test_harness`.  That's aesthetically displeasing to me, especially since I can't predict when that test would PASS and therefore be un-TODO-able.  So I'm not going to fight to the death for this p.r. to be merged.  If it's not accepted in a reasonable period of time, I'll withdraw it.

jkeenan:
> This p.r. responds (somewhat) to the issues raised in #19661. The first commit is adapted from a test file suggested by @choroba in that ticket; the other commits are my extensions of that line of thought.
> 
> This p.r. does add a print-out of a TODO test when you run `make test_harness`. That's aesthetically displeasing to me, especially since I can't predict when that test would PASS and therefore be un-TODO-able. So I'm not going to fight to the death for this p.r. to be merged. If it's not accepted in a reasonable period of time, I'll withdraw it.

I will close this pull request within 14 days unless we get additional support for committing it.

demerphq (approved):

demerphq:
I think this should be merged. The underlying problem is a bit tricky, but imo we should deal with it more gracefully than we do.

PR #19921, opened 2022-07-04
Neaten Padname flag contants

leonerd:
The constants were named `PADNAMEt_*` suggesting their purpose is a type enumeration. They're not. They're bitflags, so the name `PADNAMEf_*` would be more appropriate.

Also updated the `B` and `B::Deparse` modules to make use of the new modern names (and avoid the old `SVpad_*` flags).

### DIFF
--- a/pad.h
+++ b/pad.h
@@ -318,6 +318,7 @@ Restore the old pad saved into the local variable C<opad> by C<PAD_SAVE_LOCAL()>
 #define PadnameIsOUR(pn)	cBOOL((pn)->xpadn_ourstash)
 #define PadnameOURSTASH(pn)	(pn)->xpadn_ourstash
 #define PadnameTYPE(pn)		(pn)->xpadn_type_u.xpadn_typestash
+#define PadnameHasTYPE(pn)      cBOOL(PadnameTYPE(pn))
 #define PadnamePROTOCV(pn)	(pn)->xpadn_type_u.xpadn_protocv
 #define PadnameREFCNT(pn)	(pn)->xpadn_refcnt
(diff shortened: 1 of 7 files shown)


ilmari:
There appears to be at least one [CPAN module](https://grep.cpanauthors.org/search?q=PADNAMEt_&extension=xs) using the `PADNAMEt_*` constants.

leonerd:
@ilmari Hah - typical that it would happen to be one of mine :)
OK I suppose I'll do those back-compat wrappers too.

pad.h
```diff
@@ -348,6 +348,11 @@ Restore the old pad saved into the local variable C<opad> by C<PAD_SAVE_LOCAL()>
 #define SVpad_STATE		PADNAMEf_STATE
 #define SVpad_TYPED		PADNAMEf_TYPED
 #define SVpad_OUR		PADNAMEf_OUR
+#define PADNAMEt_OUTER          PADNAMEf_OUTER
+#define PADNAMEt_STATE          PADNAMEf_STATE
+#define PADNAMEt_LVALUE         PADNAMEf_LVALUE
+#define PADNAMEt_TYPED          PADNAMEf_TYPED
+#define PADNAMEt_OUR            PADNAMEf_OUR
```
ilmari: To avoid accidental use in-core, these should be wrapped in `#ifndef PERL_CORE`.

leonerd: Huh. Wow it turns out quite a few bits of existing code in core were still even using those old names from the previous round of renaming. I'll fix those up too.

ilmari: If I were being _really_ nitpicky I would do the commits in the opposite order, but I shan't ;-)

ilmari (approved):

Issue #19882, opened 2022-06-22
select() reacts improperly to UTF8-flagged strings


FGasper:
**Description**
Perl’s `select()` has the “Unicode bug”.

The impact is low, but the fix is trivial.

The potential for breakage seems low enough that this one is OK to fix.

**Steps to Reproduce**
```
#!/usr/bin/env perl

use strict;
use warnings;

use Test::More;

use File::Temp;

my $mask;
my @fhs;

while (!defined $mask) {
    push @fhs, scalar File::Temp::tempfile();
    vec( my $curmask, fileno($fhs[-1]), 1 ) = 1;

    if ($curmask =~ m<[\x80-\xff]>) {
        $mask = $curmask;
    }
}

print "FD: " . fileno($fhs[-1]) . $/;

my $result = select $mask, undef, undef, 0;
print "error: $!" if $result == -1;

is($result, 1, 'file is readable (downgraded input to select)');

utf8::upgrade($mask);

$result = select $mask, undef, undef, 0;

is($result, 1, 'file is readable (upgraded input to select)');

done_testing;
```

**Expected behavior**
The `strace` output of the 2nd one-liner should match that of the first one.

**Perl configuration**
This happens in blead as of now.

This fixes it:
```
diff --git a/pp_sys.c b/pp_sys.c
index a3bc0794a2..6bfa98f14f 100644
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -1165,7 +1165,10 @@ PP(pp_sselect)
                 Perl_croak_no_modify();
         }
         else if (SvIsCOW(sv)) sv_force_normal_flags(sv, 0);
-        if (!SvPOK(sv)) {
+        if (SvPOK(sv)) {
+            if (SvUTF8(sv)) sv_utf8_downgrade(sv, FALSE);
+        }
+        else {
             if (!SvPOKp(sv))
                 Perl_ck_warner(aTHX_ packWARN(WARN_MISC),
                                     "Non-string passed as bitmask");
```

jkeenan:
> **Description** Perl’s `select()` has the “Unicode bug”.
> 
> The impact is low, but the fix is trivial.
> 
> The potential for breakage seems low enough that this one is OK to fix.
[snip]> 



@tonycoz, @iabyn Can you take a look at this ticket?

Leont:
Looks good to me

tonycoz:
It seems reasonable, I think it could use a test (it isn't a PR though)

[...]

PR #19922, opened 2022-07-04
Make 4-argument select() handle UTF8-flagged strings correctly.

FGasper:
Issue #19882

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -412,6 +412,14 @@ setsockopt() now uses the mechanism added in 5.36 to better
 distinguish between numeric and string values supplied as the
 C<OPTVAL> parameter.  [github #18761]
 
+=item *
+
+4-argument C<select()> now rejects strings with code points above
+255. Additionally, for code points 128-255, this operator will now always
+give the corresponding octet to the OS, regardless of how Perl stores
+such code points in memory. (Previously Perl leaked its internal string
+storage to the OS.)
+
 =back
 
 =head1 Known Problems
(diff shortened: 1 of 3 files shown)


pod/perldelta.pod
```diff
@@ -363,7 +363,7 @@ files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
 =item *
 
-XXX
+4-argument select now properly handles UTF8-flagged strings.
```
leonerd: C<select()>

Also, this doesn't really say what it does with those strings. What is the correct semantics here? I'm not sure offhand I can imagine what "properly" means in this case

t/op/sselect.t
```diff
@@ -106,16 +106,21 @@ select ($tie, undef, undef, $tie);
 ok("no crash from select $numeric_tie, undef, undef, $numeric_tie");
 
 SKIP: {
-    require POSIX;
+    my $SKIP_CR = sub {
+        skip shift, 4;
+    };
+
+    eval 'require POSIX' or do {
```
Leont: Why quotes and not braces?

FGasper: No good reason. I think I thought the braces form caused a broken test, but that’s probably a mistake. Updated.

pod/perldelta.pod
tonycoz: I think this is resolved with the new perldelta message.

FGasper:
Conflict fixed.

PR #19924, opened 2022-07-05
Remove bareword_filehandles from the :5.38 feature bundle 

ilmari:
Commit c5327cb77526e4e4c7bb83e24be20ac48e85b1fa fixed the bug that
caused it to be put back in the :5.36 feature bundle, so let's pull it
out again.

This broke a hint-bit fiddling test that got missed when `HINT_FEATURE_MASK`
was widened to 4 bits in 5d1739474d967de1ab8a8f88aa5eff250dbc0eab, so fix that
(and add a comment indicating which constants the magic number is meant to be).


leonerd:
Seems overall a good idea, but still wants a `perldelta.pod` entry.

ilmari:
> Seems overall a good idea, but still wants a `perldelta.pod` entry.

Done, under "Updated Modules and Pragmata"

Leont (approved):

leonerd (approved):

rwp0:
Thanks, @ilmari .

PR #19926, opened 2022-07-05
Share duplicated aux data in regex patterns

khwilliamson:
Bracketed character classes may contain auxiliary data consisting of
inversion lists.  Some of these lists can be quite large, a couple
thousand UV-sized words; the worst case is much larger, but these are
the more likely worst case numbers.  If the same class is used a second
time or more in a pattern, the savings from sharing are considerable.
And it's not uncommon for such repeated classes to occur.

The official Unicode/Posix character classes are already in shared data,
so prior to this commit if you had a bunch of \w's in a pattern the only
duplication would be the array containing pointers to the bulk of the
data.  This commit eliminates even that.  But if someone said [\w,] for
example, that custom character class would have a copy of the \w
inversion list (plus the neglible comma) in each case.  That copy as of
Unicode 14.0 is about 1500 UV words.  Now any extra [\w,] classes would
share the same data as the first one in the pattern.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -20563,22 +20563,108 @@ S_set_ANYOF_arg(pTHX_ RExC_state_t* const pRExC_state,
 
     PERL_ARGS_ASSERT_SET_ANYOF_ARG;
 
-    if (     PL_regkind[OP(node)] == ANYOF
-        && ! runtime_defns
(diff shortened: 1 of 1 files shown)


demerphq:
Cool.

Yves

PR #19928, opened 2022-07-06
Unixish.h, doshish.h: Reorder terminations; simplify

khwilliamson:
The IO and memory terminations need to be after other things.  Add a
comment so that future maintainers won't make the mistakes I did.

Also refactor to that amiga os doesn't have a separate list to get out
of sync

I suspect that the amiga termination should be moved to earlier in
the sequence, but absent any evidence; I'm leaving it unchanged.

vms destruction was missing a bunch of things and I didn't see any
reason to have special handling, so I changed it to just use the
standard.

### DIFF
--- a/unixish.h
+++ b/unixish.h
@@ -136,28 +136,34 @@ int afstat(int fd, struct stat *statb);
 #define Mkdir(path,mode)   mkdir((path),(mode))
 
 #if defined(__amigaos4__)
-#  define PERL_SYS_INIT_BODY(c,v)					\
-        MALLOC_CHECK_TAINT2(*c,*v) PERL_FPU_INIT; PERLIO_INIT; MALLOC_INIT; amigaos4_init_fork_array(); amigaos4_init_environ_sema();
-#  define PERL_SYS_TERM_BODY()                         \
-    HINTS_REFCNT_TERM; KEYWORD_PLUGIN_MUTEX_TERM;      \
-    OP_CHECK_MUTEX_TERM; OP_REFCNT_TERM; PERLIO_TERM;  \
-    MALLOC_TERM; LOCALE_TERM; USER_PROP_MUTEX_TERM;    \
-    ENV_TERM;                                          \
-    amigaos4_dispose_fork_array();
+#  define PLATFORM_SYS_TERM_  amigaos4_dispose_fork_array()
+#  define PLATFORM_SYS_INIT_ STMT_START {                       \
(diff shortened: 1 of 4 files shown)


unixish.h
```diff
@@ -136,28 +136,34 @@ int afstat(int fd, struct stat *statb);
...
 #if defined(__amigaos4__)
-#  define PERL_SYS_INIT_BODY(c,v)					\
-        MALLOC_CHECK_TAINT2(*c,*v) PERL_FPU_INIT; PERLIO_INIT; MALLOC_INIT; amigaos4_init_fork_array(); amigaos4_init_environ_sema();
-#  define PERL_SYS_TERM_BODY()                         \
-    HINTS_REFCNT_TERM; KEYWORD_PLUGIN_MUTEX_TERM;      \
-    OP_CHECK_MUTEX_TERM; OP_REFCNT_TERM; PERLIO_TERM;  \
-    MALLOC_TERM; LOCALE_TERM; USER_PROP_MUTEX_TERM;    \
-    ENV_TERM;                                          \
-    amigaos4_dispose_fork_array();
+#  define AMIGAOS_SYS_TERM_  amigaos4_dispose_fork_array()
```
Leont: I think it makes more sense later on if this would be renamed to something like `PLATFORM_SYS_INIT`/`PLATFORM_SYS_TERM`, that way it can easily be reused on for other platforms if we ever need it.

Leont:
This whole area is a mess and I hate it. This makes it slightly better, but I believe it ultimately needs bigger maintenance.

unixish.h
khwilliamson: good idea

Leont (approved):

craigberry (changes requested):
This fails to link on VMS as-is ("undefined symbol PERL_SYS_TERM_BODY referenced") because unixish.h is not included.  It's only included when DOSISH, VMS, PLAN9, __VOS__ AND __HAIKU__ are false.  So the macro might have to be moved to perl.h or something.

khwilliamson:
@leont, as a result of @craigberry comments, I moved the definitions to perl.c

Leont:
> @Leont, as a result of @craigberry comments, I moved the definitions to perl.c

I think that's ok.

craigberry (approved):

craigberry:
Sorry to be so slow, but this now compiles and introduces no new test failures on VMS.

Issue #19874, opened 2022-06-17
UBsan failures on 5.36


dur-randir:
This is a bug report for perl from sergey.aleynikov@gmail.com,
generated with the help of perlbug 1.42 running under perl 5.36.0.

-----------------------------------------------------------------
[Please describe your issue here]

When running test suite for a perl 5.36.0 built with ubsan, the following tests fail:

run/fresh_perl.t with 
```
# Failed test 2 - $cusp = ~0 ^ (~0 >> 1); at ./test.pl line 1159
#      got "pp.c:2726:7: runtime error: signed integer overflow: -9223372036854775808 - 1 cannot be represented in type \'long int\'\npp.c:2824:9: runtime error: negation of -9223372036854775808 cannot be represented in type \'long int\'; cast to an unsigned type to negate this value to itself\n7 0 0 8 !"
```
and re/pat.t with
```
# Failed test 1044 - gh17730: should not crash at ./test.pl line 1159
#      got "regcomp.c:6757:15: runtime error: signed integer overflow: 9223372036854775807 + 2 cannot be represented in type \'long int\'\nok"
# expected "ok"
```

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.36.0:

Configured by s.aleinikov at Fri Jun 17 14:53:53 MSK 2022.

Summary of my perl5 (revision 5 version 36 subversion 0) configuration:

  Platform:
    osname=linux
    osvers=5.10.0-13-amd64
    archname=x86_64-linux
    uname='linux dev.crazypanda.ru 5.10.0-13-amd64 #1 smp debian 5.10.106-1 (2022-03-17) x86_64 gnulinux '
    config_args='-de -Dprefix=/home/s.aleinikov/perlbrew/perls/perl-5.36.0-asan -Dman1dir=none -Dman3dir=none -DDEBUGGING -Accflags=-fsanitize=address -fsanitize=undefined -ggdb3 -Aldflags=-Wl,--no-as-needed -lasan -lubsan -Aeval:scriptdir=/home/s.aleinikov/perlbrew/perls/perl-5.36.0-asan/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
[...]

Issue #19851, opened 2022-06-10
Lots of ASAN errors


khwilliamson:
**Description**

I ran clang-14, and it showed many many errors in our test suite.  These are the  call points.

breakbreakgv.c:871:24: runtime error: applying zero offset to null pointer
breakbreaksv.c:489:18: runtime error: applying zero offset to null pointer
breakgv.c:871:24: runtime error: applying zero offset to null pointer
breaksv.c:489:18: runtime error: applying zero offset to null pointer
defaultgv.c:871:24: runtime error: applying zero offset to null pointer
defaultsv.c:489:18: runtime error: applying zero offset to null pointer
givengv.c:871:24: runtime error: applying zero offset to null pointer
givensv.c:489:18: runtime error: applying zero offset to null pointer
gv.c:2453:48: runtime error: applying zero offset to null pointer
gv.c:287:42: runtime error: applying zero offset to null pointer
gv.c:2880:44: runtime error: applying zero offset to null pointer
gv.c:331:17: runtime error: applying zero offset to null pointer
gv.c:4138:10: runtime error: applying zero offset to null pointer
gv.c:4144:13: runtime error: applying zero offset to null pointer
gv.c:871:24: runtime error: applying zero offset to null pointer
gv.c:952:57: runtime error: applying zero offset to null pointer
gv.c:959:32: runtime error: applying zero offset to null pointer
gv.c:976:32: runtime error: applying zero offset to null pointer
okaygv.c:871:24: runtime error: applying zero offset to null pointer
oksv.c:489:18: runtime error: applying zero offset to null pointer
oksv.c:6659:26: runtime error: applying zero offset to null pointer
op.c:11715:15: runtime error: applying zero offset to null pointer
op.c:12428:27: runtime error: applying zero offset to null pointer
op.c:16426:12: runtime error: applying non-zero offset 8 to null pointer
op.c:3999:47: runtime error: applying zero offset to null pointer
op.c:3999:59: runtime error: applying zero offset to null pointer
op.c:4446:23: runtime error: applying zero offset to null pointer
pad.c:2215:5: runtime error: applying zero offset to null pointer
perl.c:3035:37: runtime error: applying zero offset to null pointer
pp.c:5315:31: runtime error: applying zero offset to null pointer
pp.c:963:37: runtime error: applying zero offset to null pointer
[...]

PR #19930, opened 2022-07-06
Fix several undefined behaviour issues detected by -fsanitize=undefined

tonycoz:
Fixes both issues mentioned in #19874.

Fixes all but the unaligned access issues described in #19851.

Draft for now, hoping for an update from @hvds on the FIXME patch.

### DIFF
--- a/cop.h
+++ b/cop.h
@@ -813,9 +813,9 @@ struct block_loop {
                 ? &GvSV((c)->blk_loop.itervar_u.gv)     \
                 : (SV **)&(c)->blk_loop.itervar_u.gv)
 
-#define CxLABEL(c)	(0 + CopLABEL((c)->blk_oldcop))
-#define CxLABEL_len(c,len)	(0 + CopLABEL_len((c)->blk_oldcop, len))
-#define CxLABEL_len_flags(c,len,flags)	(0 + CopLABEL_len_flags((c)->blk_oldcop, len, flags))
+#define CxLABEL(c)	(CopLABEL((c)->blk_oldcop))
+#define CxLABEL_len(c,len)	(CopLABEL_len((c)->blk_oldcop, len))
+#define CxLABEL_len_flags(c,len,flags)	((const char *)CopLABEL_len_flags((c)->blk_oldcop, len, flags))
 #define CxHASARGS(c)	(((c)->cx_type & CXp_HASARGS) == CXp_HASARGS)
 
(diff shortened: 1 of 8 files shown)


hvds:
> Draft for now, hoping for an update from @hvds on the FIXME patch.

I'm hoping for a comment from @demerphq on that patch (originally [here](https://github.com/Perl/perl5/issues/19874#issuecomment-1159011222)), but I think it's probably good.

khwilliamson:
I would have thought the casts would need to include 'const' to avoid assignments

tonycoz:
> I would have thought the casts would need to include 'const' to avoid assignments

No, code like:

  (type)(lvalue) = somevalue;

is ill-formed.  Old versions of gcc allowed it as an extension.
```
tony@venus:.../perl/git$ gcc -c castlvalue.c
castlvalue.c: In function ‘f’:
castlvalue.c:5:14: error: lvalue required as left operand of assignment
    5 |     (int *)x = y;
      |              ^
castlvalue.c: In function ‘g’:
castlvalue.c:9:12: error: lvalue required as left operand of assignment
    9 |     (int)z = a;
      |            ^
tony@venus:.../perl/git$ clang-14 -c castlvalue.c
castlvalue.c:5:5: error: assignment to cast is illegal, lvalue casts are not supported
    (int *)x = y;
    ^~~~~~~~ ~
castlvalue.c:9:5: error: assignment to cast is illegal, lvalue casts are not supported
    (int)z = a;
    ^~~~~~ ~
2 errors generated.
tony@venus:.../perl/git$ cat castlvalue.c
int *x;
int z;

void f(int *y) {
    (int *)x = y;
}

void g(int a) {
    (int)z = a;
}
```

khwilliamson (approved):

hvds:
Commit message for the "FIXME" can be something like `study_chunk: avoid integer overflow in calculation of delta`.

hvds:
@tonycoz - @demerphq has now commented on the `study_chunk` patch in #19874, so please remove it from this PR; he is working up a new fix.

demerphq:
I created https://github.com/Perl/perl5/pull/19947 related to the regex engine parts of this.

tonycoz:
> @tonycoz - @demerphq has now commented on the `study_chunk` patch in #19874, so please remove it from this PR; he is working up a new fix.

Done.

Leont (approved):

PR #19932, opened 2022-07-07
Revert "study_chunk: temporary underflow guard for scan_commit"

khwilliamson:
This reverts commit d23733db32b12b784b2bf07cae6d645569636f6d.
As it says in its message it should be reverted so as to surface
more underlying issues, early in the development cycle.


hvds (approved):

hvds:
Sadly I didn't include a ticket number in the original commit referencing the "multiple issues" that were seen. If anyone is able to find an original ticket or p5p conversation, that would be useful.

PR #19933, opened 2022-07-07
Emit __VA_ARGS__ macros in embed.h

leonerd:
Now we're using C99, we can safely use the `__VA_ARGS__` expansion in these variable-list macros.

Unfortunately we can't just emit them unconditionally, because much existing CPAN code exists that thinks it can call e.g. `warn()` without an aTHX_ in scope (because they don't #define PERL_NO_GET_CONTEXT).

Therefore, we have to guard these new macro forms by

      ... || defined(PERL_CORE)

and continue to emit the "..._nocontext()" variants at the end of the file, as we previously did.

It's not a great solution but it at least means we can use `croak()`, `warn()`, et.al. within perl core source now.

### DIFF
--- a/embed.h
+++ b/embed.h
@@ -101,15 +101,17 @@
 #define ck_entersub_args_list(a)	Perl_ck_entersub_args_list(aTHX_ a)
 #define ck_entersub_args_proto(a,b,c)	Perl_ck_entersub_args_proto(aTHX_ a,b,c)
 #define ck_entersub_args_proto_or_list(a,b,c)	Perl_ck_entersub_args_proto_or_list(aTHX_ a,b,c)
-#ifndef MULTIPLICITY
-#define ck_warner		Perl_ck_warner
-#define ck_warner_d		Perl_ck_warner_d
+#if !defined(MULTIPLICITY) || defined(PERL_CORE)
+#define ck_warner(a,...)	Perl_ck_warner(aTHX_ a,__VA_ARGS__)
+#endif
+#if !defined(MULTIPLICITY) || defined(PERL_CORE)
+#define ck_warner_d(a,...)	Perl_ck_warner_d(aTHX_ a,__VA_ARGS__)
 #endif
 #define clear_defarray(a,b)	Perl_clear_defarray(aTHX_ a,b)
 #define cop_fetch_label(a,b,c)	Perl_cop_fetch_label(aTHX_ a,b,c)
 #define cop_store_label(a,b,c,d)	Perl_cop_store_label(aTHX_ a,b,c,d)
-#ifndef MULTIPLICITY
-#define croak			Perl_croak
+#if !defined(MULTIPLICITY) || defined(PERL_CORE)
+#define croak(...)		Perl_croak(aTHX_ __VA_ARGS__)
 #endif
 #define croak_memory_wrap	Perl_croak_memory_wrap
 #define croak_no_modify		Perl_croak_no_modify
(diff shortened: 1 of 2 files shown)


leonerd:
> (because they don't #define PERL_NO_GET_CONTEXT).

As an extension to this I might see if I can get it to work around that as well, by providing the macros if that is defined, and only offering the `_nocontext()` variants if not. But that's a different change for a later PR.

khwilliamson:
This successfully compiled with an IBM XL C/C++ compiler in c99 mode for z/OS EBCDIC

embed.h
```diff
@@ -101,15 +101,17 @@
 #define ck_entersub_args_list(a)	Perl_ck_entersub_args_list(aTHX_ a)
 #define ck_entersub_args_proto(a,b,c)	Perl_ck_entersub_args_proto(aTHX_ a,b,c)
 #define ck_entersub_args_proto_or_list(a,b,c)	Perl_ck_entersub_args_proto_or_list(aTHX_ a,b,c)
-#ifndef MULTIPLICITY
-#define ck_warner		Perl_ck_warner
-#define ck_warner_d		Perl_ck_warner_d
+#if !defined(MULTIPLICITY) || defined(PERL_CORE)
+#define ck_warner(a,...)	Perl_ck_warner(aTHX_ a,__VA_ARGS__)
+#endif
+#if !defined(MULTIPLICITY) || defined(PERL_CORE)
```
Leont, line 107: Why the `!defined(MULTIPLICITY)`? Shouldn't that definition work regardless?

Leont:
> Therefore, we have to guard these new macro forms by `defined(PERL_CORE)`

I would prefer a define to explicitly ask for these macros in XS modules (not sure about the right name though).

embed.h
leonerd: It probably should but maybe that will be surprising to XS authors? Up until now, it's been available in third-party XS but only when `!defined(MULTIPLICITY)`. It's a silly reason, but I didn't want to break anything. If we want to revisit that we could as a separate idea.

leonerd:
> I would prefer a define to explicitly ask for these macros in XS modules (not sure about the right name though).

They've already been available to XS modules, when not built with `MULTIPLICTY`. It's annoyingly inconsistent but that is how it's been for a while

embed.h
Leont: > Up until now, it's been available in third-party XS but only when !defined(MULTIPLICITY). It's a silly reason, but I didn't want to break anything. If we want to revisit that we could as a separate idea.

But if we don't make it accessible to `!defined(MULTIPLICITY)`, none of this is actually useful.

Leont: > Up until now, it's been available in third-party XS but only when !defined(MULTIPLICITY). It's a silly reason, but I didn't want to break anything. If we want to revisit that we could as a separate idea.

But if we don't make it available to `defined(MULTIPLICITY)`, none of this is actually useful so what would be the point?

leonerd: This commit makes them visible within core, finally. Lots of code in perl core currently does `Perl_warn(aTHX_ ...)` because it cannot `warn(...)`. This commit would fix that.
A future commit could also remove the `defined(MULTIPLICTY) &&` part of this condition to make them available elsewhere as well. That would need more testing though to see if it breaks any CPAN modules.
I wanted to take the first step of making it available within core first so we can use it in our own code.

Leont: Fair, I'll leave it for a follow-up

Leont (approved):

PR #19934, opened 2022-07-07
qr/\K*/ now works without hanging; rmv check

khwilliamson:
This removes the diagnostic that was put into place to forbid saying
\K*, and similar, as the code now handles the case reasonably.


khwilliamson:
@hvds, if it helps about #19932, I found this commit in the same workspace on my box as that one, and from the same time.  So they are likely related.

hvds (approved):

PR #19935, opened 2022-07-08
Add API functions `newAVav()` and `newAVhv()`

leonerd:
As discussed in https://www.nntp.perl.org/group/perl.perl5.porters/2022/02/msg263150.html there seemed to be interest in functions to create a new AV by coping an existing AV (or HV). Here are those functions.

Named `newAVav` and `newAVhv` in order to match the existing naming scheme of things like `newSVsv` and also the recently-documented `newHVhv` (which wasn't documented at the time of the original mailing list posting, so I hadn't noticed it exists).


av.c
```diff
@@ -449,6 +449,113 @@ Perl_av_make(pTHX_ SSize_t size, SV **strp)
...
+    if(UNLIKELY(!oav))
+        return newAV();
+
+    if(LIKELY(!SvRMAGICAL(oav))) {
+        return av_make(av_count(oav), AvARRAY(oav));
+    }
+
+    AV *ret = newAV();
+
+    /* avoid ret being leaked if croak when calling magic below */
```
Leont, line 478: This feels like a rather awkward way to do this. Can't we just mortalize it, and then ref-increment it on success?

leonerd: Hmm that might work too. I just stole this structure from the `Perl_av_make` function right above it. It seemed simple enough.

Leont: Fair enough

bulk88:
For the record a major tied hash performance bug is in this PR because of not understanding the Perl C API.
```
    HE *he;
    while((he = hv_iternext(ohv))) {
        if(tied) {
            av_push(ret, newSVsv(hv_iterkeysv(he)));
            av_push(ret, newSVsv(hv_iterval(ohv, he)));
        }
        else {
            av_push(ret, newSVhek(HeKEY_hek(he)));
            av_push(ret, HeVAL(he) ? newSVsv(HeVAL(he)) : &PL_sv_undef);
        }
    }
```

```
Returns the key as an C<SV*> from the current position of the hash
iterator.  The return value will always be a mortal copy of the key.  Also
see C<L</hv_iterinit>>.

=cut
*/

SV *
Perl_hv_iterkeysv(pTHX_ HE *entry)
{
    PERL_ARGS_ASSERT_HV_ITERKEYSV;

    return newSVhek_mortal(HeKEY_hek(entry));
}
```

richardleach:
> For the record a major tied hash performance bug is in this PR because of not understanding the Perl C API.

Sounds like this should be tracked as a new issue. 

Also for `newAVav`, I've been meaning to ask why empty elements in a source are copied across as `PL_sv_undef`, rather than `NULL`. Deliberate design choice or unintended quirk that can be revisited?

PR #19938, opened 2022-07-09
Rename `CVf_METHOD` to `CVf_NOWARN_AMBIGUOUS`

leonerd:
Renaming as suggested in https://www.nntp.perl.org/group/perl.perl5.porters/2022/07/msg264282.html

I decided on `NOWARN_AMBIGUOUS` rather than `NOWARN_RESOLVE` because it feels a bit closer related to the actual warning it causes not to happen.

### DIFF
--- a/cv.h
+++ b/cv.h
@@ -107,7 +107,10 @@ See L<perlguts/Autoloading with XSUBs>.
           : 0                                           \
         )
 
-#define CVf_METHOD	0x0001	/* CV is explicitly marked as a method */
+/* CV has the `:method` attribute. This used to be called CVf_METHOD but is
+ * renamed to avoid collision with an upcoming feature */
+#define CVf_NOWARN_AMBIGUOUS	0x0001
+
 #define CVf_LVALUE	0x0002  /* CV return value can be used as lvalue */
 #define CVf_CONST	0x0004  /* inlinable sub */
 #define CVf_ISXSUB	0x0008	/* CV is an XSUB, not pure perl.  */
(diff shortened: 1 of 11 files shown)


tonycoz:
Looks reasonable to me, there's two CPAN distributions that mention it, one was last released in 2000, the other only uses it for older perls.

leonerd:
@tonycoz 
> Looks reasonable to me, there's two CPAN distributions that mention it, one was last released in 2000, the other only uses it for older perls.

Fairly sure one of my own dists cares about CvMETHOD - Object::Pad. Did your search not find this?

https://metacpan.org/release/PEVANS/Object-Pad-0.66/source/lib/Object/Pad.xs#L1227

tonycoz:
> Fairly sure one of my own dists cares about CvMETHOD - Object::Pad. Did your search not find this?

I used:

https://grep.metacpan.org/search?size=20&q=CVf_METHOD&qd=&qft=*.xs

which only found ObjStore and Sub-Filter.  Searching .c now also finds mod_perl.

leonerd:
Ah; that was the flag bit itself. The associated macros find a few more:

https://grep.metacpan.org/search?q=CvMETHOD&qd=&qft=*.xs

Still, all of those will be fine with the back-compat macros I'm leaving in place.

atoomic (approved):
Worth doing a followup with the CPAN distro using the Cvf_METHOD
or maybe add it to PPPort

leonerd:
> Worth doing a followup with the CPAN distro using the Cvf_METHOD or maybe add it to PPPort

Possibly, but for now the back-compat wrappers in here will handle it anyway.

PR #19943, opened 2022-07-11
Implement OP_PADSV_STORE - combined sassign/padsv OP

richardleach:
This commit introduces a new OP to replace simple cases
of OP_SASSIGN and OP_PADSV.

For example, `my $x = 1` is currently implemented as:

```
1  <;> nextstate(main 1 -e:1) v:{
2  <$> const(IV 1) s
3  <0> padsv[$x:1,2] sRM*/LVINTRO
4  <2> sassign vKS/2
```
But now will be turned into:

```
1  <;> nextstate(main 1 -e:1) v:{
2  <$> const(IV 1) s
3  <1> padsv_store[$x:1,2] vKMS/LVINTRO
```

This intended to be a transparent performance optimization.
It should be applicable for RHS optrees of varying complexity.

Notes:
* This optimization was suggested by @leonerd as a good project for getting feet wet.
* This was my first foray into Deparse.pm; no idea if this was the most appropriate way of supporting this new op.
* Ideas for separate optimizations of simple assignments of CONSTs, PADSVs, and UNDEF into single ops are already in the works.

Performance:
The following unrealistic code shows the op-related speedup:
```
my ($l0, $l1, $l2, $l3, $l3, $l5, $l6, $l7, $l7, $l9);
for (0..10_000_000) {
    $l0 = $l1 = $l2 = $l3 = $l4 = $l5 = $l6 = $l7 = $l8 = $l9 = 1 
}
```
Blead:
```
            870.42 msec task-clock                #    0.985 CPUs utilized          
                 1      context-switches          #    0.001 K/sec                  
               202      page-faults               #    0.232 K/sec                  
     3,800,528,668      cycles                    #    4.366 GHz                    
        79,206,368      stalled-cycles-frontend   #    2.08% frontend cycles idle   
           217,927      stalled-cycles-backend    #    0.01% backend cycles idle    
    15,523,885,258      instructions              #    4.08  insn per cycle         
                                                  #    0.01  stalled cycles per insn
     3,030,769,392      branches                  # 3481.972 M/sec                  
                 0      branch-misses             #    0.00% of all branches        

       0.883580599 seconds time elapsed
```

Patched:
```
            721.04 msec task-clock                #    0.983 CPUs utilized          
                 1      context-switches          #    0.001 K/sec                  
               197      page-faults               #    0.273 K/sec                  
     3,163,626,846      cycles                    #    4.388 GHz                    
        51,457,402      stalled-cycles-frontend   #    1.63% frontend cycles idle   
                30      stalled-cycles-backend    #    0.00% backend cycles idle    
    12,563,719,698      instructions              #    3.97  insn per cycle         
     2,470,736,436      branches                  # 3426.638 M/sec                  
                 0      branch-misses             #    0.00% of all branches        

       0.733857229 seconds time elapsed
```
The extent to which users will notice a benefit in real code will depend on things like how often the optimized op is used and what path is taken through `Perl_sv_setsv`.


pp_hot.c
```diff
@@ -123,6 +123,53 @@ PP(pp_and)
...
+    if (op->op_private & OPpLVAL_INTRO) {
+        if (!(op->op_private & OPpPAD_STATE)) {
+            save_clearsv(padentry);
+        }
+    }
+    if (op->op_private & OPpDEREF) {
+        targ = vivify_ref(targ, op->op_private & OPpDEREF);
+    }
+
+    /* Inlined, simplified pp_assign from here */
```
tonycoz: pp_sassign

peep.c
```diff
@@ -3839,6 +3839,59 @@ Perl_rpeep(pTHX_ OP *o)
...
+                 * Besides the op and sibling context, which carry over
+                 * the SASSIGN will have KIDS & STACKED set. */
+
+                OpTYPE_set(o, OP_PADSV_STORE);
+
+                o->op_flags = (
+                    OPf_MOD | OPf_KIDS | OPf_STACKED |
+                    (o->op_flags & OPf_WANT) |
+                    (o->op_flags & OPf_PARENS)
+                );
```
tonycoz: SLABBED should be retained - it means the OP was allocated from the SLAB attached to a CV during compilation.  Clearing it may result in an invalid free() later.

pp_hot.c
```diff
@@ -123,6 +123,53 @@ PP(pp_and)
...
+
+PP(pp_padsv_store)
+{
+    dSP;
+    OP * const op = PL_op;
+    SV** const padentry = &PAD_SVl(op->op_targ);
+    SV* targ = *padentry; /* lvalue to assign into */
+    SV* const val = TOPs; /* RHS value to assign */
+
+    /* !OPf_STACKED is handled by pp_padsv2padsv */
```
tonycoz: `pp_padsv2padsv` - hopefully this should be something else.

pp_hot.c
richardleach: Thanks, updated in my local copy until I fixup the ASAN errors and rebase.

peep.c
richardleach: Thanks, updated in my local copy until I fixup the ASAN errors and rebase.

I'm genuinely unsure whether to propagate the REF and SPECIAL flags or not. (`pp_sassign` doesn't use them, but I don't know if anything else would check them.)

pp_hot.c
richardleach: Thanks, tweaked in my local copy until I fixup the ASAN errors and rebase.

(pp_padsv2padsv is my placeholder name for a follow-on optimization, which was originally going to be in the same PR.)

peep.c
richardleach: Now I remember, looking at it again, SLABBED is already carried over from the original _sassign_ OP because it's a separate bitfield in `BASEOP`, not part of `op_flags`.

Leont: `(o->op_flags & OPf_WANT) | (o->op_flags & OPf_PARENS)` equals `(o->op_flags & (OPf_WANT | OPf_PARENS))`

richardleach: Merged in working copy

peep.c
```diff
@@ -1226,6 +1226,9 @@ S_finalize_op(pTHX_ OP* o)
     PERL_ARGS_ASSERT_FINALIZE_OP;
 
     do {
+if (o->op_type == OP_FREED) {
+PerlIO_stdoutf("BUGGER! 0x%p is ->op_type == OP_FREED\n", o);
+}
```
tonycoz: This should probably be removed :)

peep.c
```diff
@@ -2717,6 +2720,13 @@ Perl_rpeep(pTHX_ OP *o)
       redo:
 
         /* oldoldop -> oldop -> o should be a chain of 3 adjacent ops */
+/*
+PerlIO_stdoutf("\noldoldop is 0x%p, oldop is 0x%p\n", oldoldop, oldop);
+if (oldoldop)
+PerlIO_stdoutf("oldoldop->op_next is 0x%p\n", oldoldop->op_next);
+if (oldop)
+PerlIO_stdoutf("oldop->op_next is 0x%p\n", oldop->op_next);
+*/
```
tonycoz: This too

peep.c
richardleach: Gah. Can't believe I missed these debugging statements. Thanks for your eagle eyes!

peep.c
richardleach: Removed

lib/B/Deparse.pm
```diff
@@ -3092,6 +3092,18 @@ sub pp_isa { binop(@_, "isa", 15) }
 sub pp_sassign { binop(@_, "=", 7, SWAP_CHILDREN) }
 sub pp_aassign { binop(@_, "=", 7, SWAP_CHILDREN | LIST_CONTEXT) }
 
+sub pp_padsv_store {
+    my $self = shift;
+    my ($op,$cx,$forbid_parens,@args)=@_;
```
leonerd: Most minor of style notes - this wants whitespacing

```perl
my ($op, $cx, $forbid_parens, @args) = @_;
```

peep.c
```diff
@@ -3839,6 +3839,65 @@ Perl_rpeep(pTHX_ OP *o)
...
+               ) {
+
+                /* SASSIGN's bitfield flags, such as op_moresib and
+                 * op_slabbed, will be carried over unchanged. */
+                OpTYPE_set(o, OP_PADSV_STORE);
+
+                /* Explicitly craft the new OPs op_flags, carrying
+                 * some bits over from the SASSIGN */
+                o->op_flags = (
+                    OPf_MOD | OPf_KIDS | OPf_STACKED |
```
leonerd: Why are we preserving `OPf_MOD` here? There's no need to explicitly mark that a `padsv_store` operation is a modification - that's already given :) The `pp_padsv_store` function doesn't test that bit at all.

peep.c
```diff
@@ -3839,6 +3839,65 @@ Perl_rpeep(pTHX_ OP *o)
...
+                    (o->op_flags & (OPf_WANT|OPf_PARENS))
+                );
+
+                /* Reset op_private flags */
+                o->op_private = 0;
```
leonerd: This assignment is pointless given the line below.

pp_hot.c
```diff
@@ -123,6 +123,53 @@ PP(pp_and)
...
+    SV** const padentry = &PAD_SVl(op->op_targ);
+    SV* targ = *padentry; /* lvalue to assign into */
+    SV* const val = TOPs; /* RHS value to assign */
+
+    /* !OPf_STACKED is not handled by this OP */
+    assert(op->op_flags & OPf_STACKED);
+
+    /* Inlined, simplified pp_padsv here */
+    if (op->op_private & OPpLVAL_INTRO) {
+        if (!(op->op_private & OPpPAD_STATE)) {
```
leonerd: These two conditions could be simplified to

```c
if(op->op_private & (OPpLVAL_INTRO|OPpPAD_STATE) == OPpLVAL_INTRO)
```

as a way to test both flags with a single conditional

t/perf/opcount.t
```diff
@@ -788,4 +788,14 @@ test_opcount(0, "builtin::is_tainted is replaced with direct opcode",
                     is_tainted => 1,
                 });
 
+# sassign + padsv combinations are replaced by padsv_store
+test_opcount(0, "sassign + padsv replaced by padsv_store",
+                sub { my $x = undef; my $y; my $z = $y = 3; },
+                {
+                    entersub => 0,
```
leonerd: Probably no need to count `entersub` in here - I suspect that's just been copied from the previous test

regen/opcodes
```diff
@@ -94,6 +94,7 @@ transr		transliteration (tr///)	ck_match	is"	S
 # Lvalue operators.
 
 sassign		scalar assignment	ck_sassign	s2	S S
+padsv_store	padsv scalar assignment	ck_sassign	sT1	S S
```
leonerd: Ordering-wise, I'd be tempted to put this right under `padsv`; on the assumption that at some future point we'll also do some similar logic for a newly-added `padav_assign` and `padhv_assign` as well

peep.c
```diff
@@ -3839,6 +3839,65 @@ Perl_rpeep(pTHX_ OP *o)
...
+
+                /* Take relevant private flasgs from the PADSV */
+                o->op_private = (lval->op_private &
+                                (OPpLVAL_INTRO|OPpPAD_STATE|OPpDEREF));
+
+                /* Take the targ from the PADSV */
+                o->op_targ = lval->op_targ;
```
leonerd: I'd consider

```c
/* Steal the targ from the OP_PADSV */
o->op_targ = lval->op_targ; lval->op_targ = 0;
```

peep.c
```diff
@@ -3839,6 +3839,65 @@ Perl_rpeep(pTHX_ OP *o)
...
+                oldoldop->op_next = o;
+                /* Even when (rhs != oldoldop), rhs might still have a
+                 * relevant op_next ptr to lval. This is definitely true
+                 * when rhs is OP_NULL with a LOGOP kid (e.g. orassign).
+                 * There may be other cases. */
+                if (rhs->op_next == lval)
+                    rhs->op_next = o;
+
+                /* Do not trash the pad when op_free calls Perl_pad_free */
+                lval->op_targ = 0;
```
leonerd: This could then be removed if moved above.

peep.c
```diff
@@ -3839,6 +3839,65 @@ Perl_rpeep(pTHX_ OP *o)
...
+
+                /* Fixup op_next ptrs */
+                /* oldoldop can be arbitrarily deep in the RHS OP tree */
+                oldoldop->op_next = o;
+                /* Even when (rhs != oldoldop), rhs might still have a
+                 * relevant op_next ptr to lval. This is definitely true
+                 * when rhs is OP_NULL with a LOGOP kid (e.g. orassign).
+                 * There may be other cases. */
+                if (rhs->op_next == lval)
+                    rhs->op_next = o;
```
leonerd, line 3887: Ah; if this is the case (and yes the weird branch structure in LOGOPs make it complicated), then it's best not to free the op. Instead use

```c
op_null(lval);
```

To turn the op into one of those "ex-padsv" things you see in Concise output sometimes. Being a OP_NULL'ed op the optree finalizer will skip over it in all the actual op chaining.

leonerd (changes requested):

lib/B/Deparse.pm
richardleach: Thanks. Will do.

peep.c
richardleach: True. Will remove.

pp_hot.c
richardleach: I'd left it as it was for easy comparison with `pp_padsv`, but happy to make this change.

peep.c
richardleach: I don't yet have a good understanding of when things other than the `pp_` code - e.g. other parts of the optimization/fixup routines, or _B::Deparse_ - need the `op_flags`, so it seemed like a reasonable precaution to set it. If there's no need for it, I can take it out.

peep.c
richardleach: Will do.

peep.c
richardleach: Will do.

t/perf/opcount.t
richardleach: Yeah, copied from the preceding tests.

regen/opcodes
richardleach: Ok, will do.

peep.c
richardleach: I did try `op_null(lval);` originally, but had problems with memory leakage. (Presumably because it was not linked to elsewhere in the optree?)  I'll give it another go; two questions:

1. Does this mean that `OP_PADSV_STORE `should remain a BINOP, not be a UNOP? 
2. If so, do I still need to do all of the fixing up of pointers that is in the current iteration of _peep.c_, or will e.g. some of the `op_next` pointers be automatically updated by the optree finalizer?

richardleach: > 2\. If so, do I still need to do all of the fixing up of pointers that is in the current iteration of _peep.c_, or will e.g. some of the `op_next` pointers be automatically updated by the optree finalizer?

That doesn't seem to happen. The OP_PADSV looks to still be in the `op_next` chain despite being a null op. e.g.
```
$ ./perl -Dt -e 'my $boop = 1'

EXECUTING...

(-e:0)	enter
(-e:0)	nextstate
(-e:1)	const(IV(1))
(-e:1)	null
(-e:1)	padsv_store
(-e:1)	leave e.g.
```
So seems like I still have to fix up the op_next pointers manually. In which case I'm not clear on why `op_null(lval)` is more desirable than `op_clear(lval)`, unless there are potentially some more pointers in the RHS that might point to the ex-PADSV and potentially not redirected by the optimization.

richardleach: I've pushed a commit that leaves OP_PADSV_STORE as a BINOP and does `op_null(lval)`, but still has the manual `op_next` fix-ups. See what you think?

leonerd: > * Does this mean that `OP_PADSV_STORE `should remain a BINOP, not be a UNOP?

Not quite sure I follow the question. `regen/opcodes` should call it an UNOP. If you manually create one you should `newUNOP(OP_PADSV_STORE, ...)`.

But if you took an existing BINOP and did `OpTYPE_set(o, OP_PADSV_STORE)` that is fine. A BINOP is larger in memory than an UNOP. All that means is that we're now storing an UNOP in a little memory slab that was accidentally too large for it. No huge problem.

> * If so, do I still need to do all of the fixing up of pointers that is in the current iteration of _peep.c_, or will e.g. some of the `op_next` pointers be automatically updated by the optree finalizer?

You can leave it in memory where it is, so this is all fine. But to answer the question: nothing in the finalizer will *change* any existing op_next pointers (apart from to unthread any `OP_NULL`s). If you did change something and it's already been threaded up you'll have to go chasing all the pointers yourself. Which is why it's usually best to change an op inplace, so no pointer hunting required.

richardleach: Hmm, I'm probably not following correctly, but what I've observed is:
1. 
* If I make OP_PADSV_STORE an UNOP, call `OpTYPE_set(o, OP_PADSV_STORE)`, fixup `rhs` to have no sibs, and then `op_null(lval)`, there are memory leaks detected by ASAN. To get rid of the leaks, I had to `op_free(lval)`. (If I left `rhs` untouched, `rpeep` would barf.)
versus:
* If I make OP_PADSV_STORE a BINOP, call `OpTYPE_set(o, OP_PADSV_STORE)`, and then `op_null(lval)`, that seems fine.

2. The finalizer definitely does not unthread the ex-PADSV `OP_NULL` from the `op_next` pointer chain. I have to do that manually. (I can't see any code in `finalize_op` that would do this. Maybe `op_prune_chain_head` does this for OP_NULLs in some but not all locations?)

richardleach: Ok, you were definitely right that it should work with OP_PADSV_STORE being an UNOP, which this PR now uses again, along with `op_null(lval)`. I think what must have gone wrong on my first attempts was that I though I had to manually fix-up the RHS, but that wasn't actually necessary and it cast the `lval` adrift, leading to the leakage.

I think this PR is ready to re-review.

richardleach:
Any further comments on this one, or does it look good to merge?

tonycoz (approved):

leonerd (changes requested):
All that's here looks good, but one tiny thing is missing.

regen/op_private needs to remark that the new opcode makes use of the `OPp_DEFER` bits. Add it to the list around line 416.

richardleach:
Thanks for spotting that. I'll take a look later today. Might not be straightforward - regen complains that:
`addbits(): bit 4 of padsv_store already specified (OPpTARGET_MY) at ./regen/op_private line 417.`

(I can't remember offhand if _B::Deparse_ needs the `OPpTARGET_MY` flag.)

richardleach:
Assuming all CI tests pass, OP_PADSV_CONST doesn't seem to need the `OPpTARGET_MY` flag, so I've taken that out and added in the `OPpDEREF` flag bits.

richardleach:
After working on #20036 and looking a bit more at this PR, I realised that the basic optimization will not need to support `OPpDEREF` and have removed that branch from `pp_padsv_store`.

Note: It also seemed simple enough to add support for assigning simple `padsv-padsv-sassign` optrees, which I originally did. However, this involved turning `OP_PADSV_STORE` into an `UNOP_aux` and stashing the second `padoffset` in the `op_aux` slot. That worked fine, but there was no way to make _Deparse.pm_ work (short of hacking around in _B.xs_, which definitely seems like a bad move.) So I dropped that commit.

Hopefully, UNOPs (and possibly other op types) will soon get an `op_src` member, at which point the `padsv-padsv-sassign` commit can be tweaked to use it and submitted as a separate PR.

t/perf/opcount.t
```diff
@@ -788,4 +788,13 @@ test_opcount(0, "builtin::is_tainted is replaced with direct opcode",
                     is_tainted => 1,
                 });
 
+# sassign + padsv combinations are replaced by padsv_store
+test_opcount(0, "sassign + padsv replaced by padsv_store",
+                sub { my $x = undef; my $y; my $z = $y = 3; },
+                {
+                    padsv    => 1,
+                    padsv_store  => 3,
+                    undef    => 1,
```
leonerd: Tiny whitespace alignment issue here. Make the longest one have a single space, and align the others at the `=>` to match.

leonerd (changes requested):
One tiny whitespace issue but otherwise this all looks good now.

Will you be writing a perldelta.pod entry?

t/perf/opcount.t
richardleach: Thanks. I took out the `undef` assignment, since #20077 will change it, and aligned the spaces on the remainder.

leonerd (approved):
Excellent, looking good now.
This can be merged assuming CI is happy.

t/perf/opcount.t
richardleach: Re: perldelta.pod, an entry per optimization seems like too much. Would a placeholder that covers multiple (and could link to individual PRs or commits) suffice? Something like: 
"Multiple optree optimizations have been applied in order to speed up common operations."

PR #19946, opened 2022-07-13
(try to) eliminate some coverity scan complaints

tonycoz:


jkeenan:
@tonycoz, is there any way that we can simulate/imitate the Coverity scan so that we can see the status before and then after applying these patches?

tonycoz:
I did some searching, and I couldn't find anything that would let us distinguish such testing builds from the mainline blead builds.

When my script submits a build it does supply a description (which is currently the result of `git describe`) which I thought we might be able to use, but I don't see that available anywhere in the defect view.  It would be useful to identify exactly which build defects would be found in too.

Otherwise we could submit the testing builds on a separate project, but that doesn't provide a simple way to compare against the mainline build.

jkeenan:
> _ No description provided. _



> I did some searching, and I couldn't find anything that would let us distinguish such testing builds from the mainline blead builds.
> 
> When my script submits a build it does supply a description (which is currently the result of `git describe`) which I thought we might be able to use, but I don't see that available anywhere in the defect view. It would be useful to identify exactly which build defects would be found in too.
> 
> Otherwise we could submit the testing builds on a separate project, but that doesn't provide a simple way to compare against the mainline build.

Okay, then we might as well merge now and let Coverity tell us the results.

PR #19947, opened 2022-07-13
Fix various ubsan revealed issues in the build process and regex engine.

demerphq:
See #19874 for related discussion. I only addressed build warnings and the regex engine.

locale.c - silence warnings about unused S_calculate_LC_ALL() under non USE_POSIX_2008_LOCALE builds
regexec.c - avoid (bogus) compiler warnings about possibly uninitialized pointer dereference and make code clearer
regcomp.h - fix non aligned pointer value stored in the regex opcode directly.
regcomp.c - deal with overflow of delta during stopmin logic related to (*ACCEPT)
regen/embed_lib.pl - deal with comments in `#else` statements better and fix handling for `#endif`

Also for the regex engine stuff I added a bunch of comments.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -6820,18 +6820,88 @@ S_study_chunk(pTHX_
     assert(!frame);
     DEBUG_STUDYDATA("pre-fin", data, depth, is_inf, min, stopmin, delta);
 
+    /* is this pattern infinite? Eg, consider /(a|b+)/ */
+    if (is_inf_internal)
+        delta = OPTIMIZE_INFTY;
+
+    /* deal with (*ACCEPT), Eg, consider /(foo(*ACCEPT)|bop)bar/ */
     if (min > stopmin) {
-        /* stopmin might be shorter than min if we saw an (*ACCEPT). If
-        this is the case then it means this pattern is variable length
-        and we need to ensure that the delta accounts for it. delta
-        represents the difference between min length and max length for
-        this part of the pattern. */
-        delta += min - stopmin;
+        /*
+        At this point 'min' represents the minimum length string we can
+        match while *ignoring* the implication of ACCEPT, and 'delta'
+        represents the difference between the minimum length and maximum
+        length, and if the pattern matches an infinitely long string
+        (consider the + and * quantifiers) then we use the special delta
+        value of OPTIMIZE_INFTY to represent it. 'stopmin' is the
+        minimum length that can be matched *and* accepted.
+
+        A pattern is accepted when matching was successful *and*
+        complete, and thus there is no further matching needing to be
+        done, no backtracking to occur, etc. Prior to the introduction
+        of ACCEPT the only opcode that signaled acceptance was the END
+        opcode, which is always the very last opcode in a regex program.
+        ACCEPT is thus conceptually an early successful return out of
+        the matching process. stopmin starts out as OPTIMIZE_INFTY to
+        represent "the entire pattern", and is ratched down to the
+        "current min" if necessary when an ACCEPT opcode is encountered.
+
+        Thus stopmin might be smaller than min if we saw an (*ACCEPT),
+        and we now need to account for it in both min and delta.
+        Consider that in a pattern /AB/ normally the min length it can
+        match can be computed as min(A)+min(B). But (*ACCEPT) means
+        that it might be something else, not even neccesarily min(A) at
+        all. Consider
+
+             A  = /(foo(*ACCEPT)|x+)/
+             B  = /whop/
+             AB = /(foo(*ACCEPT)|x+)whop/
+
+        The min for A is 1 for "x" and the delta for A is OPTIMIZE_INFTY
+        for "xxxxx...", its stopmin is 3 for "foo". The min for B is 4 for
+        "whop", and the delta of 0 as the pattern is of fixed length, the
+        stopmin would be OPTIMIZE_INFTY as it does not contain an ACCEPT.
+        When handling AB we expect to see a min of 5 for "xwhop", and a
+        delta of OPTIMIZE_INFTY for "xxxxx...whop", and a stopmin of 3
+        for "foo". This should result in a final min of 3 for "foo", and
+        a final delta of OPTIMIZE_INFTY for "xxxxx...whop".
+
+        In something like /(dude(*ACCEPT)|irk)x{3,7}/ we would have a
+        min of 6 for "irkxxx" and a delta of 4 for "irkxxxxxxx", and the
+        stop min would be 4 for "dude". This should result in a final
+        min of 4 for "dude", and a final delta of 6, for "irkxxxxxxx".
+
+        When min is smaller than stopmin then we can ignore it. In the
+        fragment /(x{10,20}(*ACCEPT)|a)b+/, we would have a min of 2,
+        and a delta of OPTIMIZE_INFTY, and a stopmin of 10. Obviously
+        the ACCEPT doesn't reduce the minimum length of the string that
+        might be matched, nor affect the maximum length.
+
+        In something like /foo(*ACCEPT)ba?r/ we would have a min of 5
+        for "foobr", a delta of 1 for "foobar", and a stopmin of 3 for
+        "foo". We currently turn this into a min of 3 for "foo" and a
+        delta of 3 for "foobar" even though technically "foobar" isn't
+        possible. ACCEPT affects some aspects of the optimizer, like
+        length computations and mandatory substring optimizations, but
+        there are other optimzations this routine perfoms that are not
+        affected and this compromise simplifies implementation.
+
+        It might be helpful to consider that this C function is called
+        recursively on the pattern in a bottom up fashion, and that the
+        min returned by a nested call may be marked as coming from an
+        ACCEPT, causing its callers to treat the returned min as a
+        stopmin as the recursion unwinds. Thus a single ACCEPT can affect
+        multiple calls into this function in different ways.
+        */
+
+        if (OPTIMIZE_INFTY - delta >= min - stopmin)
+            delta += min - stopmin;
+        else
+            delta = OPTIMIZE_INFTY;
         min = stopmin;
     }
 
     *scanp = scan;
-    *deltap = is_inf_internal ? OPTIMIZE_INFTY : delta;
+    *deltap = delta;
 
     if (flags & SCF_DO_SUBSTR && is_inf)
         data->pos_delta = OPTIMIZE_INFTY - data->pos_min;
@@ -6862,7 +6932,9 @@ S_study_chunk(pTHX_
 }
 
 /* add a data member to the struct reg_data attached to this regex, it should
- * always return a non-zero return */
+ * always return a non-zero return. the 's' argument is the type of the items
+ * being added and the n is the number of items. The length of 's' should match
+ * the number of items. */
 STATIC U32
 S_add_data(RExC_state_t* const pRExC_state, const char* const s, const U32 n)
 {
--- a/regcomp.h
+++ b/regcomp.h
@@ -112,6 +112,45 @@ typedef struct regexp_internal {
 
 /* this is where the old regcomp.h started */
 
+
+/* Define the various regnode structures. These all should be a multiple
+ * of 32 bits large, and they should by and large correspond with each other
+ * in terms of naming, etc. Things can and will break in subtle ways if you
+ * change things without care. If you look at regexp.h you will see it
+ * contains this:
+ *
+ * struct regnode {
+ *   U8  flags;
+ *   U8  type;
+ *   U16 next_off;
+ * };
+ *
+ * This structure is the base unit of elements in the regexp program. When
+ * we increment our way through the program we increment by the size of this
+ * structure, and in all cases where regnode sizing is considered it is in
+ * units of this structure.
+ *
+ * This implies that no regnode style structure should contain 64 bit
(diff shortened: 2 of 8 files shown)


regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
+        use the special delta value of OPTIMIZE_INFTY to represent it.
+        stopmin is a special case which represents the minimum length we
+        can match without having to match the entire pattern.
+
+        In particular stopmin might be smaller than min if we saw an
+        (*ACCEPT). If this is the case then it means this pattern is
+        variable length and we need to ensure that the delta accounts
+        for it properly. Consider that in a pattern /AB/ normally the
+        min length it can match is computed as min(A)+min(B). But (*ACCEPT)
+        means that it might be something else, not even necesarily
```
hvds, line 6842: necessarily

regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
+        min(A) at all. Consider
+
+             A= /(foo(*ACCEPT)|x+)/
+             B= /whop/
+             AB= /(foo(*ACCEPT)|x+)bar/
```
hvds, line 6847: Should be `AB = /(foo(*ACCEPT)|x+)whop/`

regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
+
+        The min for A is 1 for "X" and the delta for A is OPTIMIZE_INFTY
+        for "xxxxx...". The min for B is 4 for "whop", and the delta is 0.
+        The min for AB is 3 for "foo", and the delta is OPTIMIZE_INFTY for
+        "xxxxx...bar".
+
+        Another example is /(foo(*ACCEPT)|x+)bar/ where we would have a
+        delta of OPTIMIZE_INFTY, a stopmin of 3, and a min of 4. This
+        should result in a delta of OPTIMIZE_INFTY and a min of 3.
```
hvds: This example doesn't seem to add anything, it gives the same results as the preceding "whop" example/

regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
+
+        In something like /(dude(*ACCEPT)|irk)x{10,20}/ we would have a
+        delta of 10, a stopmin of 4 and a min of 13. This should result
+        in a delta of 20 and a min of 3.
```
hvds: I'm not sure this makes sense - the final min should be 4, I think. I also don't understand "we would have a delta of 10 ... This should result in a delta of 20".

regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
+        */
+        if (OPTIMIZE_INFTY - delta >= min - stopmin)
+            delta += min - stopmin;
```
hvds: I think this needs `... else delta = OPTIMIZE_INFTY;` (and maybe setting some is_inf* too).

regcomp.c
```diff
@@ -6820,18 +6820,52 @@ S_study_chunk(pTHX_
...
         min = stopmin;
     }
 
     *scanp = scan;
-    *deltap = is_inf_internal ? OPTIMIZE_INFTY : delta;
+    *deltap = delta;
```
hvds: Ah! This was the motivation for my original version of the fix. I note that this logic originated from the initial introduction of `is_inf_internal` by Ilya in 1998 (commit aca2d49724). At that point it was line 714!

regcomp.h
```diff
@@ -112,6 +112,42 @@ typedef struct regexp_internal {
 
 /* this is where the old regcomp.h started */
 
+
```
hvds, line 115: In the commit message: `s/64 aligned/64 bit aligned/`; `s/that is stored/that it is stored/`; `s/which part/which is part/`

regcomp.h
```diff
@@ -112,6 +112,42 @@ typedef struct regexp_internal {
...
+ *   U16 next_off;
+ * };
+ *
+ * This structure is the base unit of elements in the regexp program. When
+ * we increment our way through the program we increment by the size of this
+ * structure, and in all cases where regnode sizing is considered it is in
+ * units of this structure.
+ *
+ * This implies that no regnode style structure should contain 64
+ * aligned members. Since the base regnode is 32 bits any member might
```
hvds: 64 bit aligned

regcomp.h
```diff
@@ -150,7 +186,12 @@ struct regnode_p {
     U8	flags;
     U8  type;
     U16 next_off;
-    SV * arg1;
+#if PTRSIZE == 8
```
hvds: The comment to `struct regnode_p` just above could also mention the caveats around 64-bit pointers.

regcomp.h
```diff
@@ -226,6 +267,8 @@ struct regnode_charclass_posixl {
  * allocated in terms of multiples of a single-argument regnode.  SSC nodes can
  * have a pointer field because there is no alignment issue, and because it is
  * set to NULL after construction, before any cloning of the pattern */
+/* NOTE - this "regnode" is not used in the program itself, so you do not need
+ * to worry about member alignment issues here */
```
hvds: This is already said in part in the preceding comment. It might be clearer to replace these extra lines with a slight augmentation of what precedes:
```
  [...] Also, there is no alignment issue,
  * because this regnode is declared or allocated as a complete unit and never
  * used in the program itself, to the compiler takes care of alignment. [...]
```

regcomp.h
```diff
@@ -282,13 +325,25 @@ struct regnode_ssc {
 #undef ARG2
 
 #define ARG(p) ARG_VALUE(ARG_LOC(p))
-#define ARGp(p) ARG_VALUE(ARGp_LOC(p))
+#if PTRSIZE == 8
+#define ARGp(p) (SV*)(((UV)ARG_VALUE(ARGp_PTR_HIGH_LOC(p))) << 32 | ((UV)(ARG_VALUE(ARGp_PTR_LOW_LOC(p)))))
```
hvds: Note that this and `ARGp_SET` newly evaluate their `p` argument multiple times. Any chance of switching them to be inline functions instead?

hvds (commented):
I'll leave @khwilliamson to comment on the fold_array and locale.c changes.

regcomp.h
```diff
@@ -150,7 +186,12 @@ struct regnode_p {
     U8	flags;
     U8  type;
     U16 next_off;
-    SV * arg1;
+#if PTRSIZE == 8
+    U32 ptr_high;
+    U32 ptr_low;
+#else
```
tonycoz: Instead of this, why not just `char ptr_bytes[sizeof(SV*)]` and memcpy() the pointer value in and out?

ARGp() would need to be (or use) an inline function or gcc brace group, but I think it would simplify things otherwise, and make fewer assumptions about pointer sizes.

The compiler can also optimize this to a simple (unaligned) access, eg.

```
#include <string.h>

typedef unsigned char U8;
typedef unsigned short U16;

struct regnode_p {
    U8	flags;
    U8  type;
    U16 next_off;
    char arg1_bytes[sizeof(void *)];
};

static inline void
ARGp_SET(struct regnode_p *n, void *p) {
    memcpy(n->arg1_bytes, &p, sizeof(p));
}

static inline void *
ARGp(const struct regnode_p *n) {
    void *p;
    memcpy(&p, n->arg1_bytes, sizeof(p));

    return p;
}

int x;

void foo(struct regnode_p *n) {
    ARGp_SET(n, &x);
}

void *bar(struct regnode_p *n) {
    return ARGp(n);
}
```
With gcc -O2, foo and bar are:
```
foo:
.LFB2:
        .cfi_startproc
        leaq    x(%rip), %rax
        movq    %rax, 4(%rdi)
        ret
        .cfi_endproc
...
bar:
.LFB3:
        .cfi_startproc
        movq    4(%rdi), %rax
        ret
        .cfi_endproc
```

As you can see, while it's undefined behaviour for us to access the unaligned data in C, the compiler has more freedom.

regcomp.h
demerphq: Id rather not, and save the energy to eliminate regnode_p entirely. We shouldnt be putting pointers into the program like this as they wont be handled right, i believe the use case this one is put to is "special" and avoids those issues, but really we should just convert to the "Safe" api and use the data[] array.

regcomp.h
demerphq: I dont really understand your argument. Why would using memcpy to copy 8 bytes be better than assigning two 32 bit words? 

Ideally we Just Shouldn't Have 64 bit members in our regnodes. Id rather we just eliminate the need than make our implementation better, this is just a work around for now until I have the time to address this properly and rip it out entirely. If had been paying attention when this was introduced I would have blocked it and directed Karl to the data[] array and related functionality.

regcomp.c
demerphq: About the "4" yes, I changed this example from "foo" to "dude" and missed a number that needed to be updated as a consequence.

The part about delta is confusing because it is also off by one. But consider the min and delta /during/ the processing of this sub do NOT account for ACCEPT. The code this comment is part of is where that accounting is done. The comment is intended to describe what the initial values would be, and what we would expect as a result.

regcomp.c
demerphq: Ah, yes, good catch.

regcomp.h
demerphq: Ah. I guess I will do that in a fixup before final merge so I dont clobber this feedback with a rebase :-)

regcomp.h
demerphq: +1

regcomp.h
demerphq: I decided to rework that paragraph entirely as it was too dense.

regcomp.h
demerphq: Also note that its only used in a very few places, I dont think we need to worry about it getting reused.

regcomp.h
demerphq: I was insufficiently caffeinated when I wrote my original reply.  Roger, willco. :-)

regcomp.h
demerphq: Ended up using the approach @tonycoz suggested, so this will use inline functions now.

regcomp.c
demerphq: Yeah, in a future patch I think we should introduce some macros/inline functions for delta updates. I suspect there are other bugs hidden in our code that our test suite doesnt tickle. There is a conceptual issue BTW, we use "max length we can store" to represent "infinity". Which in theory could result in us turning something bounded into something unbounded.

regcomp.h
demerphq: Please don't resolve this until I squash the patches.

demerphq:
I've pushed a bunch of comment changes and fixup patches to resolve all the issues that dont need a rebase/squash.

hvds:
> I've pushed a bunch of comment changes and fixup patches to resolve all the issues that dont need a rebase/squash.

Updates all look good to me, except I'm not sure if the inline functions should go to `inline.h` (with an appropriate prefix).

regcomp.h
```diff
@@ -181,17 +184,25 @@ struct regnode_1 {
...
     U8	flags;
     U8  type;
     U16 next_off;
-#if PTRSIZE == 8
-    U32 ptr_high;
-    U32 ptr_low;
-#else
-    SV *ptr;
-#endif
+    char arg1_sv_ptr_bytes[sizeof(void *)];
```
tonycoz: I used `sizeof(void *)` in my example code to keep it short(er), but this should be `sizeof(SV*)`.

demerphq: Ack. Fixed.

demerphq:
FWIW, inline.h doesn't feel right to me, these aren't general purpose and should only be available in the regex engine.

regcomp.h
demerphq: Squashed now.

demerphq:
I squashed down all the fixup patches now, and fixed the commit message @hvds pointed out.

khwilliamson:
inline functions don't have to go into inline.h.  Only those that are needed by multiple files should go there.

regcomp.h
khwilliamson: There's no need to be concerned with regnode_p and performance.  It is compile only, removed before execution

I had already decided that it should use a mechanism like data[], but it should use a new array that can just be wiped out at the end of compilation, since data[] is persistent.  There is no way IIUC to remove individual elements of it.

khwilliamson (approved):

tonycoz (approved):

demerphq:
On Thu, 14 Jul 2022 at 19:27, Karl Williamson ***@***.***>
wrote:

>
> In regcomp.h
> <https://github.com/Perl/perl5/pull/19947#discussion_r921407585>:
>
> > +    U32 ptr_high;
> +    U32 ptr_low;
> +#else
>
> There's no need to be concerned with regnode_p and performance. It is
> compile only, removed before execution
>
> I had already decided that it should use a mechanism like data[], but it
> should use a new array that can just be wiped out at the end of
> compilation, since data[] is persistent. There is no way IIUC to remove
> individual elements of it.
>
I believe if the element is an SV and NULL then nothing would break.

Yves

-- 

demerphq:
Pushed a comment fixup, and then a squash again. Will merge after it passes CI.

demerphq:
@hvds waiting on you to hit approve.

hvds (approved):

PR #19948, opened 2022-07-13
Mark S_setlocale_failure_panic_i() as currently unused

leonerd:
It otherwise provokes a compiler warning because the function isn't actually called anywhere. This warning becomes fatal on Windows compilers.


leonerd:
Actually this bug has now been fixed. This can be closed.

PR #19951, opened 2022-07-14
getcwd() is fixed from dragonfly bsd 6.2

### LINKED ISSUE #16525: cwd_enoent.t fails on DragonFly BSD
Migrated from [rt.perl.org#133141](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133141) (status was 'open')

Searchable as RT133141$

tonycoz:
While working on this I found a problem where osvers wasn't being populated properly on both default 5.8 and 6.2 installs, so a fix to Configure is included for that.

@Tux is the Configure change sane?

Fixes #16525

### DIFF
--- a/Configure
+++ b/Configure
@@ -3494,7 +3494,10 @@ EOM
 			osvers="$3"
 			;;
 		dragonfly) osname=dragonfly
-			osvers="$3"
+                        case "$3" in
+                            [0-9]*) osvers="$3" ;;
+                            *) osvers="$2" ;;
+                        esac
 			;;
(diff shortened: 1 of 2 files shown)


jkeenan (commented):
Test change looks okay.  I'll let Tux comment on Configure.

atoomic (approved):

atoomic:
@Tux Configure changed here too

Tux:
Configure part backported as https://github.com/Perl/metaconfig/commit/5ee0f028677255e6cf5b743b240ef08959293097

PR #19952, opened 2022-07-14
Fix slowness in unicode regex-subst

pengvado:
Applying a regex-substitution to a unicode string is much slower than it should be.
`perf` says that the vast majority of the cpu-time is in Perl_utf8_length, called from Perl_pp_subst. (That's 98% of cpu-time in this isolated benchmark, and I have some real-world scripts where it's as high as 40%.)
This is not a bug in utf8 length caching, rather it's a case that hasn't implemented caching.

The solution isn't to add caching, because pp_subst doesn't really need to know utf8 length at all. It was only using that info to compute an upper bound on number of matches so that perl can die instead of hang if a bug in perl core leads to the regex engine getting stuck in an infinite loop. But that bound doesn't need to be tight, so we can use length-in-bytes as an upper bound on length-in-characters.

I had to change one regression test, which expected Perl_utf8_length to print a warning about a malformed string. Of course it doesn't print if Perl_utf8_length isn't called. The test was associated with a (very old) commit making it safe to regex-process malformed strings, it doesn't reject them, so this doesn't change the functionality of the test. Also, Perl_utf8_length only checks for a very limited subset of the ways that a utf8 string can be malformed, so anything that actually needs validity can't depend on this warning.

```
use Benchmark qw(:all);
my $str = "a" x 10000;
timethis(-5, sub { $str =~ s/a/b/r; }, "ascii s///");
utf8::upgrade($str);
timethis(-5, sub { $str =~ s/a/b/r; }, "utf s///");
```
before patch, on a Sandybridge cpu:
ascii s///:  5 wallclock secs ( 5.25 usr +  0.00 sys =  5.25 CPU) @ 2205538.29/s (n=11579076)
  utf s///:  5 wallclock secs ( 5.21 usr +  0.00 sys =  5.21 CPU) @ 37167.95/s (n=193645)
after patch:
ascii s///:  5 wallclock secs ( 5.31 usr +  0.00 sys =  5.31 CPU) @ 2208935.22/s (n=11729446)
  utf s///:  5 wallclock secs ( 5.12 usr +  0.00 sys =  5.12 CPU) @ 1664148.83/s (n=8520442)

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1110,6 +1110,7 @@ Lincoln D. Stein <lstein@cshl.org> Lincoln Stein <lstein@formaggio.cshl.org>
 Lincoln D. Stein <lstein@cshl.org> Lincoln Stein <lstein@genome.wi.mit.edu>
 Linda Walsh <unknown> Linda Walsh <unknown>
 Lionel Cons <lionel.cons@cern.ch> Lionel Cons <lionel.cons@cern.ch>
+Loren Merritt <pengvado@videolan.org> Loren Merritt <pengvado@videolan.org>
 Louis Strous <louis.strous@gmail.com> Louis Strous <louis.strous@gmail.com>
 Lubomir Rintel <lkundrak@v3.sk> Lubomir Rintel (GoodData) <lubo.rintel@gooddata.com>
 Lubomir Rintel <lkundrak@v3.sk> Lubomir Rintel <lkundrak@v3.sk>
--- a/AUTHORS
+++ b/AUTHORS
@@ -827,6 +827,7 @@ Lesley Binks                   <lesley.binks@gmail.com>
 Lincoln D. Stein               <lstein@cshl.org>
 Linda Walsh
 Lionel Cons                    <lionel.cons@cern.ch>
+Loren Merritt                  <pengvado@videolan.org>
 Louis Strous                   <louis.strous@gmail.com>
 Lubomir Rintel                 <lkundrak@v3.sk>
 Luc St-Louis                   <luc.st-louis@ca.transport.bombardier.com>
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -4250,7 +4250,6 @@ PP(pp_subst)
     STRLEN len;
     int force_on_match = 0;
     const I32 oldsave = PL_savestack_ix;
-    STRLEN slen;
     bool doutf8 = FALSE; /* whether replacement is in utf8 */
 #ifdef PERL_ANY_COW
     bool was_cow;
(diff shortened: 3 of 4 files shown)


khwilliamson (approved):

jkeenan:
Several questions:

* Is there any thing equivalent to `perf` for non-Linux systems?
* [According to Wikipedia](https://en.wikipedia.org/wiki/Sandy_Bridge), the chip on which you tested was discontinued in 2013.  Can this be tested on more recent chips?
* My own testing on FreeBSD-12 suggests that while there is a vast improvement in performance for the `utf8::upgrade`d strings, there is a slight slowdown for the ASCII strings.
```
ascii
blead           branch
1720804         1685235
1808897         1685235
1784850         1705570

utf
blead           branch
28408           1419169
28408           1408562
28408           1392009
```
I do not claim to be a benchmarking expert, but I think this raises some questions.

richardleach:
I had a quick go at benchmarking on a Debian Linux system running on an AMD Zen 3 CPU. Using the Benchmark script above, I initially got similar results to Jim:

blead:
```
ascii s///:  5 wallclock secs ( 5.00 usr +  0.00 sys =  5.00 CPU) @ 4399565.60/s (n=21997828)
  utf s///:  6 wallclock secs ( 5.23 usr +  0.00 sys =  5.23 CPU) @ 49057.74/s (n=256572)
```
patched
```
ascii s///:  5 wallclock secs ( 5.24 usr +  0.00 sys =  5.24 CPU) @ 4336108.21/s (n=22721207)
  utf s///:  5 wallclock secs ( 5.32 usr +  0.00 sys =  5.32 CPU) @ 3660772.56/s (n=19475310)
```
But if I took Benchmark out of it and just ran "perf stat" a bunch of times on the following ascii case:
```
my $str = "a" x 10000;
for (1 .. 100_000_000) {
    $str =~ s/a/b/r;
}
```
...sometimes blead was a bit faster and sometimes the patched version was a bit faster. So the performance on ascii strings seemed to be within the realm of normal noise for me.

richardleach:
PS For alternatives to `perf`, I'm not sure if there's anything as easy to use for e.g. FreeBSD. Googling suggests hwpmc or dtrace, but it's not obvious to me how to replicate `perf`-like usage easily.

Depending upon your CPU:
* Intel's VTune [supports FreeBSD](https://www.intel.com/content/www/us/en/develop/documentation/vtune-help/top/set-up-analysis-target/freebsd-targets/freebsd-system-setup.html), but there'd be a learning curve. It's commercial software, but they use to and probably still do give out free licenses for eligible OSS contributors. 
* AMD's uprof [also supports FreeBSD](https://developer.amd.com/amd-uprof/) and is free for anyone to use, but again there'll be a learning curve.

PR #19955, opened 2022-07-14
newAVav/newAVhv - use recent inline AV functions

richardleach:
Apologies to @leonerd for not commenting on #19935 in time. The new _av.c_ functions can be made more efficient by using new-ish inline functions and/or doing array assignments directly.

The commits are:
1. The `newAV_alloc_x(z)` macros are more efficient here than calling `av_extend()`, especially in newAVav, where there's no need to `Zero()` the allocation.
2. Since `ret` is a new, simple AV, we could use `av_push_simple` in place of `av_push`. I made the assignments and `AvFILLp` management explicit instead, but wouldn't be too surprised if a good compiler did similar with `av_push_simple`. Also, since most of the guts of `av_make` seemed to be in `newAVav`, seemed like we may as well do the simple case inline.
3. Again, since `ret` is a new, simple AV, we could use `av_push_simple` in place of `av_push`, but I went with doing things explicitly for clarity.

Wasn't sure whether it would be useful to hoist the `if (tied)` check outside of the loop in `newAVhv`, creating two separate loops instead, but perhaps that's something to leave the compiler to decide?

### DIFF
--- a/av.c
+++ b/av.c
@@ -523,7 +523,7 @@ Perl_newAVhv(pTHX_ HV *ohv)
 
     bool tied = SvRMAGICAL(ohv) && mg_find(MUTABLE_SV(ohv), PERL_MAGIC_tied);
 
-    U32 nkeys = hv_iterinit(ohv);
+    Size_t nkeys = hv_iterinit(ohv);
     /* This number isn't perfect but it doesn't matter; it only has to be
      * close to make the initial allocation about the right size
      */
@@ -538,12 +538,12 @@ Perl_newAVhv(pTHX_ HV *ohv)
     HE *he;
     while((he = hv_iternext(ohv))) {
         if(tied) {
-            av_push(ret, newSVsv(hv_iterkeysv(he)));
-            av_push(ret, newSVsv(hv_iterval(ohv, he)));
+            av_push_simple(ret, newSVsv(hv_iterkeysv(he)));
+            av_push_simple(ret, newSVsv(hv_iterval(ohv, he)));
         }
         else {
-            av_push(ret, newSVhek(HeKEY_hek(he)));
-            av_push(ret, HeVAL(he) ? newSVsv(HeVAL(he)) : &PL_sv_undef);
+            av_push_simple(ret, newSVhek(HeKEY_hek(he)));
+            av_push_simple(ret, HeVAL(he) ? newSVsv(HeVAL(he)) : &PL_sv_undef);
         }
     }
 
--- a/av.c
+++ b/av.c
@@ -466,14 +466,10 @@ Perl_newAVav(pTHX_ AV *oav)
 {
     PERL_ARGS_ASSERT_NEWAVAV;
 
-    U32 count = av_count(oav);
+    Size_t count = av_count(oav);
 
     if(UNLIKELY(!oav) || count == 0)
         return newAV();
-
-    if(LIKELY(!SvRMAGICAL(oav))) {
-        return av_make(av_count(oav), AvARRAY(oav));
-    }
 
     AV *ret = newAV_alloc_x(count);
 
@@ -482,9 +478,17 @@ Perl_newAVav(pTHX_ AV *oav)
     PL_tmps_stack[++PL_tmps_ix] = (SV *)ret;
     SSize_t ret_at_tmps_ix = PL_tmps_ix;
 
-    for(U32 i = 0; i < count; i++) {
-        SV **svp = av_fetch(oav, i, 0);
-        av_push(ret, svp ? newSVsv(*svp) : &PL_sv_undef);
+    Size_t i;
+    if(LIKELY(!SvRMAGICAL(oav) && AvREAL(oav) && (SvTYPE(oav) == SVt_PVAV))) {
+        for(i = 0; i < count; i++) {
+            SV **svp = av_fetch_simple(oav, i, 0);
+            av_push_simple(ret, svp ? newSVsv(*svp) : &PL_sv_undef);
+        }
+    } else {
+        for(i = 0; i < count; i++) {
+            SV **svp = av_fetch(oav, i, 0);
+            av_push_simple(ret, svp ? newSVsv(*svp) : &PL_sv_undef);
+        }
     }
 
     /* disarm leak guard */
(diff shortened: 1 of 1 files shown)


av.c
```diff
@@ -521,31 +524,34 @@ Perl_newAVhv(pTHX_ HV *ohv)
...
+            AvARRAY(ret)[i++] = newSVsv(hv_iterval(ohv, he));
         }
         else {
-            av_push(ret, newSVhek(HeKEY_hek(he)));
-            av_push(ret, HeVAL(he) ? newSVsv(HeVAL(he)) : &PL_sv_undef);
+            AvARRAY(ret)[i++] = newSVhek(HeKEY_hek(he));
+            AvARRAY(ret)[i++] = HeVAL(he) ? newSVsv(HeVAL(he)) : &PL_sv_undef;
         }
     }
+    AvFILLp(ret) = i - 1;
```
atoomic: note that this could lead to set it to `-1`?

atoomic: probably not, we abort earlier after checking the number of entries, if it s `= 0`

richardleach: Yeah, in `newAVav()`, that assignment shouldn't be reached if `count == 0`. I think it could be reached in the `newAVhv()` case if there were actually no keys in `ohv`, but `AvFILLp(ret)` really should `== -1` if that's the case.

richardleach: It's safe to say that just using `av_push_simple` and letting the compiler do its best does reduce the likelihood of human thinko errors in e.g. setting `AvFILLp(ret). I can change the commits to do that if people prefer.

av.c
```diff
@@ -466,28 +466,31 @@ Perl_newAVav(pTHX_ AV *oav)
...
-        av_push(ret, svp ? newSVsv(*svp) : &PL_sv_undef);
+    U32 i;
+    if(LIKELY(!SvRMAGICAL(oav) && AvREAL(oav) && (SvTYPE(oav) == SVt_PV))) {
+        for(i = 0; i < count; i++) {
+            SV **svp = av_fetch_simple(oav, i, 0);
+            AvARRAY(ret)[i] = svp ? newSVsv(*svp) : &PL_sv_undef;
+        }
+    } else {
+        for(i = 0; i < count; i++) {
+            SV **svp = av_fetch(oav, i, 0);
```
leonerd, line 489: Because this is a magic-invoking fetch there's a chance it might throw an exception. If it does, the `AvFILL` has not yet been updated, so none of the initial elements in the array will be freed when the AV itself is destroyed; they will leak.

To fix this you'll have to `AvFILLp(ret)++` after every successful addition.

av.c
```diff
@@ -466,28 +466,31 @@ Perl_newAVav(pTHX_ AV *oav)
...
 
-    U32 count = av_count(oav);
-
-    av_extend(ret, count);
-
-    for(U32 i = 0; i < count; i++) {
-        SV **svp = av_fetch(oav, i, 0);
-        av_push(ret, svp ? newSVsv(*svp) : &PL_sv_undef);
+    U32 i;
+    if(LIKELY(!SvRMAGICAL(oav) && AvREAL(oav) && (SvTYPE(oav) == SVt_PV))) {
```
leonerd: Do you mean `SvTYPE == SVt_PVAV` here?

av.c
```diff
@@ -521,31 +524,34 @@ Perl_newAVhv(pTHX_ HV *ohv)
...
+    SSize_t i = 0;
     while((he = hv_iternext(ohv))) {
+        if (i > AvMAX(ret))
+            av_extend(ret, i);
+
         if(tied) {
-            av_push(ret, newSVsv(hv_iterkeysv(he)));
-            av_push(ret, newSVsv(hv_iterval(ohv, he)));
+            AvARRAY(ret)[i++] = newSVsv(hv_iterkeysv(he));
+            AvARRAY(ret)[i++] = newSVsv(hv_iterval(ohv, he));
```
leonerd: Same comments here about die during tied fetch of key / value. Made worse by the fact that the `hv_iternext` is part of the while condition of the loop itself.

Either this wants to be two entire sections, based on `tied`, or use `av_push_simple()` entirely.

leonerd (changes requested):
Overall the direction seems good, but some problems with potential memory leak if tied fetch on oav / ohv throws an exception.

All of these might be more easily fixed by use of `av_push_simple()` as suggested by other comments, or by moving the assignment of `AvFILLp()` into a per-iteration `AvFILLp(ret)++`.

av.c
richardleach: Thanks, I'd not thought deeply enough about destruction. I'll change this to use `av_push_simple()`.

av.c
richardleach: Yarp.

av.c
richardleach: Thanks, I'd not thought deeply enough about destruction. I'll change this to use `av_push_simple()`.

richardleach:
I've changed the code to use `av_push_simple()`, rather than manually manage assignments and the value of `AvFILLp`. Squashed and all tests pass.

av.c
```diff
@@ -466,24 +466,22 @@ Perl_newAVav(pTHX_ AV *oav)
 {
     PERL_ARGS_ASSERT_NEWAVAV;
 
-    if(UNLIKELY(!oav))
+    U32 count = av_count(oav);
```
tonycoz: Not that you introduced it, but this should be `Size_t` and not `U32`.  Similarly below.

richardleach: Thanks Tony. Tweaked, squashed, pushed.

tonycoz (approved):

av.c
```diff
@@ -521,29 +523,27 @@ Perl_newAVhv(pTHX_ HV *ohv)
 
     bool tied = SvRMAGICAL(ohv) && mg_find(MUTABLE_SV(ohv), PERL_MAGIC_tied);
 
-    AV *ret = newAV();
+    Size_t nkeys = hv_iterinit(ohv);
+    /* This number isn't perfect but it doesn't matter; it only has to be
+     * close to make the initial allocation about the right size
+     */
+    AV *ret = newAV_alloc_xz( nkeys ? nkeys * 2 : 2);
```
leonerd, line 530: Tiniest of tiny whitespace issues here - `newAV_alloc_xz(nkeys ? ...`

leonerd:
LGTM besides the whitespace issue. I'll merge it and just fix that up on blead

Issue #16500, opened 2018-04-09
I18N::LangTags::List name fails for az-*


p5pRT:
Migrated from [rt.perl.org#133093](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133093) (status was 'open')

Searchable as RT133093$

p5pRT:
### From martin@senfdax.de

#### Created by martin@senfdax.de

This is a bug report for perl from martin@&#8203;senfdax\.de\,
generated with the help of perlbug 1\.40 running under perl 5\.26\.0\.


\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-
I think that the sub name does not respond corret to 'az\-Latn'\, 
'az\-Arab' and 'az\-Cyrl'\. The results returns the "subform fallback" 
because the string 'az\-Arab' in the POD contains a uppercase char while 
sub name\(\) invokes lc\(\) on $\_\[0\]

 > perl \-E 'use I18N&#8203;::LangTags&#8203;::List; $I18N&#8203;::LangTags&#8203;::List&#8203;::Debug = 1; 
say I18N&#8203;::LangTags&#8203;::List&#8203;::name\("az\-Latn"\)'
Input&#8203;: \{az\-latn\}
Shaving off&#8203;: \-latn leaving az
Output&#8203;: \{Azerbaijani\}\{\-latn\}
Azerbaijani \(Subform "latn"\)

<details><summary>Perl Info</summary>

```

Flags:
     category=library
     severity=low
     module=I18N::LangTags

Site configuration information for perl 5.26.0:

Configured by martin at Wed May 31 13:14:31 CEST 2017.

Summary of my perl5 (revision 5 version 26 subversion 0) configuration:

   Platform:
     osname=linux
     osvers=4.4.0-78-generic
     archname=x86_64-linux-multi
     uname='linux psyche 4.4.0-78-generic #99-ubuntu smp thu apr 27 
15:29:09 utc 2017 x86_64 x86_64 x86_64 gnulinux '
     config_args='-de 
-Dprefix=/home/martin/perl5/perlbrew/perls/perl-5.26.0 -Duseshrplib 
-Dusemultiplicity 
[...]

PR #19959, opened 2022-07-16
I18N:LangTags: Fix casing issue GH #16500

khwilliamson:
This was caused by errors in the data.  A few entries weren't
lowercase like all the rest are.


jkeenan:
LGTM

jkeenan:
> LGTM

Actually, if you could add descriptions to the two new unit tests, that would be good.

Issue #19962, opened 2022-07-17
[doc] builtin should document the behaviour of `weaken` and `unweaken` on non-references


tobyink:
**Where**
builtin

**Description**
builtin should document what `weaken` and `unweaken` do when called on non-references.

(Pretty sure the answer is that they're a no-op.)

tonycoz:
They both croak for defined non-references.

jkeenan:
> They both croak for defined non-references.

Please review https://github.com/Perl/perl5/pull/19965 for tests and documentation.

PR #19965, opened 2022-07-18
Test and document non-reference argument to weaken and unweaken

jkeenan:
For: https://github.com/Perl/perl5/issues/19962

### DIFF
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -122,13 +122,14 @@ currently but will be removed in a later version.
 Weakens a reference. A weakened reference does not contribute to the reference
 count of its referent. If only weakened references to a referent remain, it
 will be disposed of, and all remaining weak references to it will have their
-value set to C<undef>.
+value set to C<undef>.  Dies when provided with a defined non-reference.
 
 =head2 unweaken
 
     unweaken($ref);
 
 Strengthens a reference, undoing the effects of a previous call to L</weaken>.
+Dies when provided with a defined non-reference.
 
 =head2 is_weak
 
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -67,6 +67,18 @@ package FetchStoreCounter {
     weaken($ref);
     undef $arr;
     is($ref, undef, 'ref is now undef after arr is cleared');
+
+    $ref = 'GH 19965';
+    {
+        local $@;
+        eval { weaken($ref); };
+        like($@, qr/^Can't weaken a nonreference/, "Can't weaken a defined nonreference");
+    }
+    {
+        local $@;
+        eval { unweaken($ref); };
+        like($@, qr/^Can't unweaken a nonreference/, "Can't unweaken a defined nonreference");
+    }
 }
 
 # reference queries


lib/builtin.t
```diff
@@ -67,6 +67,30 @@ package FetchStoreCounter {
...
+        local $@;
+        eval { unweaken($ref); };
+        like($@, qr/^Can't unweaken a nonreference/, "Can't unweaken a defined nonreference");
+    }
+
+    $ref = undef;
+    {
+        local $@;
+        eval { weaken($ref); };
+        ok(! $@, "Can weaken an undefined nonreference");
```
tonycoz: You can call weaken on an undefined value, but it doesn't weaken it (that has no meaning).

tonycoz:
My main issue is whether the current inconsistent behaviour (shared by Scalar::Util::weaken) is what we want.  @leonerd ?

lib/builtin.pm
```diff
@@ -122,13 +122,14 @@ currently but will be removed in a later version.
 Weakens a reference. A weakened reference does not contribute to the reference
 count of its referent. If only weakened references to a referent remain, it
 will be disposed of, and all remaining weak references to it will have their
-value set to C<undef>.
+value set to C<undef>.  Dies when provided with a defined non-reference.
```
leonerd, line 125: Hrm, it's all a bit weird here. This distinction goes right the way down to the actual `sv_rvweaken` function, which complains about defined but non-referential values, but lets undef itself have a special pass and just ignores it. That just is what the function has always done, but it still feels kinda weird.

Given that, I wonder if we should specifically also point out that undef is permitted but ignored.

Also I tend to avoid the verb "die" in actual documentation - that's the perl operator name for it, but I usually describe it as "throws an exception".

lib/builtin.t
```diff
@@ -67,6 +67,18 @@ package FetchStoreCounter {
     weaken($ref);
     undef $arr;
     is($ref, undef, 'ref is now undef after arr is cleared');
+
```
leonerd, line 70: Plus if we're going to document that `undef` is permitted, it probably want a test here to check it doesn't throw

leonerd:
Actually, since `builtin::weaken` is new and experimental it doesn't have to follow the same weird sloppy interface of the prior art of `sv_rvweaken()` and `Scalar::Util::weaken`. We do have an opportunity to tidy it up - make it more consistent somehow. Either more things should be permitted (with perhaps a warning), or fewer things should and it should actually throw more often. The current random mix is not a great combination.

hvds:
> Actually, since `builtin::weaken` is new and experimental it doesn't have to follow the same weird sloppy interface of the prior art of `sv_rvweaken()` and `Scalar::Util::weaken`. We do have an opportunity to tidy it up - make it more consistent somehow. Either more things should be permitted (with perhaps a warning), or fewer things should and it should actually throw more often. The current random mix is not a great combination.

This PR doesn't seem like the right place for that discussion, should it happen on the RFC? My 2c: if the current behaviour is "die if it's something we can't weaken, except silently ignore undef" that seems entirely useful and perlish, but I wouldn't be unhappy if it were also to die on undef. I do think dying on defined non-references is the right thing to do.

jkeenan:
I'm going to withdraw this pull request because it appears that there are issues that go well beyond documentation and test coverage.  Questions like "What should the behavior of XXX be?" are best handled on the mailing list.

Issue #16595, opened 2018-06-23
pod/buildtoc: generates warnings and errors in pod/perltoc.pod


p5pRT:
Migrated from [rt.perl.org#133296](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133296) (status was 'open')

Searchable as RT133296$

p5pRT:
### From @jkeenan

The program 'pod/buildtoc' is nearly the last step run by 'make'\.  It 
generates 'pod/perltoc\.pod' from all the other \.pod files in the core 
distribution\.  If\, however\, you run the 'podchecker' utility over 
'pod/perltoc\.pod'\, you will get both WARNINGs and ERRORs\.  For example\, 
in perl\-5\.26\.0\, you get&#8203;:

\#\#\#\#\#
$ perldoc \-l perltoc
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod

$ podchecker \`perldoc \-l perltoc\`
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20657 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20675 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20677 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20815 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20817 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
\*\*\* WARNING&#8203;: empty section in previous paragraph at line 20833 in file 
/home/jkeenan/perl5/perlbrew/perls/perl\-5\.26\.0/lib/5\.26\.0/pod/perltoc\.pod
[...]

PR #19966, opened 2022-07-18
buildtoc - fix inner links when generating perltoc.pod

haarg:
When generating perltoc.pod, the headings taken from all of the various
pod documents included in core. These headings can have arbitrary
content, including links. A link with a second but without a name won't
work once it's moved into the perltoc.pod file, so convert any links
like this to include the name of the pod document they are being taken
from.

### DIFF
--- a/pod/buildtoc
+++ b/pod/buildtoc
@@ -237,6 +237,8 @@ sub podset {
 	    s/^\s*\*\s*//;
 	    s/\n/ /g;
 	    s/\s+$//;
+	    # make sure inner links include the target page
+	    s{L<(?:[^|>]+\|)?\K/}{$pod/}g;
 	    next if /^[\d.]+$/;
 	    next if $pod eq 'perlmodlib' && /^ftp:/;
 	    $OUT .= ", " if $initem;


haarg:
This is meant to fix #16595, which I believe was incorrectly closed with [a change filed ](https://rt.cpan.org/Ticket/Display.html?id=143622) on `Test-Harness`. The Pod in TAP::Harness is completely valid, it just gets broken when partially copied to a different file. The responsibility to deal with that lies with core, not `Test-Harness`.

jkeenan:
I checked out this branch, configured and called `make test_porting`, but then got this failure:
```
$ cd t; ./perl harness porting/exec-bit.t; cd -
porting/exec-bit.t .. 1/? # Failed test 118 - tarball will chmod +x ../pod/buildtoc at porting/exec-bit.t line 60
# Remove the exec bit or add '../pod/buildtoc' to Porting/exec-bit.txt
porting/exec-bit.t .. Failed 1/135 subtests 

Test Summary Report
-------------------
porting/exec-bit.t (Wstat: 0 Tests: 135 Failed: 1)
  Failed test:  118
Files=1, Tests=135,  0 wallclock secs ( 0.03 usr  0.01 sys +  0.07 cusr  0.03 csys =  0.14 CPU)
Result: FAIL
```
This p.r. *does,* however clear up the two ERRORs reported in https://github.com/Perl/perl5/issues/16595 when one calls `podchecker pod/perltoc.pod`.

jkeenan:
Most recent commit clears up porting test problem.  +1 to merge.  Thanks.

Issue #19983, opened 2022-07-22
BBC: Math-FastGF2-0.07 triggers "stack smashing detected"


andk:
**Description**

With 1ef9039bcc the tests for DMALONE/Math-FastGF2/Math-FastGF2-0.07.tar.gz were starting to fail. Sample test report with v5.37.2: http://www.cpantesters.org/cpan/report/21cf2db8-0954-11ed-aef9-e4323abf57b8

**Steps to Reproduce**

cpan -i DMALONE/Math-FastGF2/Math-FastGF2-0.07.tar.gz

**Expected behavior**

Tests should succeed and module should be installed.

**Observed behavior**
``` text
PERL_DL_NONLAZY=1 "/tmp/basesmoker-reloperl-a8HE/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t
t/Cauchy.t ........ ok
t/Math-FastGF2.t .. ok
*** stack smashing detected ***: terminated
t/Matrix.t ........ 
Failed 166/195 subtests 
*** stack smashing detected ***: terminated
t/Vandermonde.t ... 
All 19 subtests passed 

Test Summary Report
-------------------
t/Matrix.t      (Wstat: 134 (Signal: ABRT, dumped core) Tests: 29 Failed: 0)
  Non-zero wait status: 134
  Parse errors: Bad plan.  You planned 195 tests but ran 29.
t/Vandermonde.t (Wstat: 134 (Signal: ABRT, dumped core) Tests: 19 Failed: 0)
  Non-zero wait status: 134
  Parse errors: No plan found in TAP output
Files=4, Tests=109,  1 wallclock secs ( 0.06 usr  0.05 sys +  0.29 cusr  0.16 csys =  0.56 CPU)
Result: FAIL
Failed 2/4 test programs. 0/109 subtests failed.
make: *** [Makefile:1065: test_dynamic] Error 255
```

**Perl configuration**

```
# perl -V output goes here

Summary of my perl5 (revision 5 version 37 subversion 2) configuration:
  Commit id: 7f9b65d969a1c8b44028cacbb9e88596d1e5a8a3
  Platform:
    osname=linux
    osvers=5.15.0-41-generic
    archname=x86_64-linux
[...]

PR #19990, opened 2022-07-24
Add assert to SvPV-ish macros

khwilliamson:
This is to see how many CPAN modules call these with a differently sized
'len' than the specified 'STRLEN'.  See
https://github.com/Perl/perl5/issues/19983

### DIFF
--- a/sv.h
+++ b/sv.h
@@ -1902,19 +1902,22 @@ END_EXTERN_C
     (((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK) || ((SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK)))
 
 #define SvPV_flags(sv, len, flags)                                          \
+   (__ASSERT_(sizeof(len) == sizeof(STRLEN))                                \
    Perl_SvPV_helper(aTHX_ sv, &len, flags, SvPVnormal_type_,                \
-                    Perl_sv_2pv_flags, FALSE, 0)
+                    Perl_sv_2pv_flags, FALSE, 0))
 #define SvPV_flags_const(sv, len, flags)                                    \
+   (__ASSERT_(sizeof(len) == sizeof(STRLEN))                                \
    ((const char*) Perl_SvPV_helper(aTHX_ sv, &len, flags, SvPVnormal_type_, \
                                    Perl_sv_2pv_flags, FALSE,                \
-                                   SV_CONST_RETURN))
+                                   SV_CONST_RETURN)))
 #define SvPV_flags_const_nolen(sv, flags)                                   \
    ((const char*) Perl_SvPV_helper(aTHX_ sv, NULL, flags, SvPVnormal_type_, \
                                    Perl_sv_2pv_flags, FALSE,                \
                                    SV_CONST_RETURN))
 #define SvPV_flags_mutable(sv, len, flags)                                  \
+   (__ASSERT_(sizeof(len) == sizeof(STRLEN))                                \
     Perl_SvPV_helper(aTHX_ sv, &len, flags, SvPVnormal_type_,               \
-                     Perl_sv_2pv_flags, FALSE, SV_MUTABLE_RETURN)
+                     Perl_sv_2pv_flags, FALSE, SV_MUTABLE_RETURN))
 
 #define SvPV_nolen(sv)                                                      \
     Perl_SvPV_helper(aTHX_ sv, NULL, SV_GMAGIC, SvPVnormal_type_,           \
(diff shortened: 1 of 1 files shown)


jkeenan:
> This is to see how many CPAN modules call these with a differently sized 'len' than the specified 'STRLEN'. See #19983

Can you be a bit more specific as to what your battle plan is here?

For example ... are you proposing to include this in our next monthly development release, such that the volume of CPANtesters failure reports would indicate the scope of the problem?

khwilliamson:
On 7/24/22 11:08, James E Keenan wrote:
>     This is to see how many CPAN modules call these with a differently
>     sized 'len' than the specified 'STRLEN'. See #19983
>     <https://github.com/Perl/perl5/issues/19983>
> 
> Can you be a bit more specific as to what your battle plan is here?
> 
> For example ... are you proposing to include this in our next monthly 
> development release, such that the volume of CPANtesters failure reports 
> would indicate the scope of the problem?
> 
That would be one possibility; I hoped that people's comments would 
indicate if that would be a good idea

sv.h
```diff
@@ -1902,19 +1902,22 @@ END_EXTERN_C
     (((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK) || ((SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK)))
 
 #define SvPV_flags(sv, len, flags)                                          \
+   (__ASSERT_(sizeof(len) == sizeof(STRLEN))                                \
```
tonycoz, line 1905: Why `__ASSERT_()` (which really isn't a name we should define, let alone use) instead of `assert(...),`?

Unfortunately we can't use a static assert here :(

khwilliamson:
On 7/24/22 17:30, Tony Cook wrote:
> 
> 
> In sv.h <https://github.com/Perl/perl5/pull/19990#discussion_r928326673>:
> 
>> @@ -1902,19 +1902,22 @@ END_EXTERN_C
>       (((SvFLAGS(sv) & (SVf_POK|SVs_GMG)) == SVf_POK) || ((SvFLAGS(sv) & (SVf_IOK|SVp_POK|SVs_GMG)) == (SVf_IOK|SVp_POK)))
>   
>   #define SvPV_flags(sv, len, flags)                                          \
> +   (__ASSERT_(sizeof(len) == sizeof(STRLEN))                                \
> 
> Why |__ASSERT_()| (which really isn't a name we should define, let alone 
> use) instead of |assert(...),|?

Because it already exists, has since 5.19.7, is documented as API, and 
has been backported.  I don't know why the original has the leading 
underscores which make it undefined behavior.

We could create a synonym that is legal for us to use.   ASSERT_ comes 
to mind, just to avoid propagating this one further, but the former has 
a longish history to it.

We can't use plain

  assert(),

  because on non-DEBUGGING builds, that is a syntax error.

khwilliamson:
On 7/24/22 11:08, James E Keenan wrote:
>     This is to see how many CPAN modules call these with a differently
>     sized 'len' than the specified 'STRLEN'. See #19983
>     <https://github.com/Perl/perl5/issues/19983>
> 
> Can you be a bit more specific as to what your battle plan is here?
> 
> For example ... are you proposing to include this in our next monthly 
> development release, such that the volume of CPANtesters failure reports 
> would indicate the scope of the problem?

I think it should go in blead for now, to see what breaks immediately, 
and then if a bunch does, revert it.  If not, then decide about a 
development release.

tonycoz:
> We can't use plain assert(), because on non-DEBUGGING builds, that is a syntax error.
> […](#)

Plain assert(), as used in a comma expression:
```
(assert(foo),expected result)
```
has been valid syntax since 2014 when the bug was fixed in 11f9ab1a291.

sv.h
khwilliamson: Bram++  pointed me to a discussion about using a static assert expression.  I had a go at that, and am stymied.

Yes, one can create such a thing for C, but apparently not for C++, without using lambda expressions.  But I don't consider this a show stopper.  For C++ compilations, just expand to nothing

The problem I run into is many compilation warnings because we enable C++ compat warnings.  AFAICT the pragmas to turn those off around these are statements, not an expression

khwilliamson:
So what should we do about __ASSERT_?

tonycoz:
> So what should we do about `__ASSERT_`?

I'd suggest documenting it as discouraged, don't use it in new code.

khwilliamson:
I think we would need to backport your plain assert() fix

demerphq:
@tonyc I think we should remove all the uses of `__ASSERT_` from our code if we don't want it to be used. When I need to create a new macro I look for one that does something similar and crib from it. If all the examples use `__ASSERT_` then people will continue to use it.  BTW, just for the record could you explain why that spelling/macro is problematic? Are identifiers like that reserved?

tonyc:
@demerphq  I think you pinged the wrong GH user ;)

khwilliamson:
symbols containing multiple underscores in a row or beginning with an underscore are reserved for the implementation of C itself.  There's generally no problem unless our names collide, which is very unlikely.  Still, it is better practice to not use those.

I do think C has reserved too large a class.

khwilliamson:
He meant @tonycoz

tonycoz:
> @tonyc I think we should remove all the uses of `__ASSERT_` from our code if we don't want it to be used.

I was mostly considering it unnecessary churn, but removing it avoids technical debt, so removing it really isn't that unnecessary.

jkeenan:
@khwilliamson this p.r. has had no activity or discussion since December 2022.  My impression is that people were not enthusiastic about it.  Can we close it?

PR #19993, opened 2022-07-25
my_strerror(): Avoid leak

khwilliamson:
This code does a savepv(), and then before returning, a SAVEFREEPV() of
that already saved PV.  The result is a leak.

Remove the SAVEFREEPV() and fix up its only two callers to free its
return when done, and update the comments.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5137,11 +5137,16 @@ Perl__is_in_locale_category(pTHX_ const bool compiling, const int category)
 char *
 Perl_my_strerror(pTHX_ const int errnum)
 {
-    /* Returns a mortalized copy of the text of the error message associated
-     * with 'errnum'.  It uses the current locale's text unless the platform
-     * doesn't have the LC_MESSAGES category or we are not being called from
-     * within the scope of 'use locale'.  In the former case, it uses whatever
-     * strerror returns; in the latter case it uses the text from the C locale.
+    /* Returns a newly malloc'd copy of the text of the error message
+     * associated with 'errnum'.
+     *
+     * The caller must arrange for the return value to be freed after it is no
+     * longer needed.
+     *
+     * If not called from within the scope of 'use locale', it uses the text from
+     * the C locale.  If Perl is compiled to not pay attention to LC_MESSAGES,
+     * it uses whatever strerror() returns.  Otherwise the text is
+     * derived from the locale LC_MESSAGES is in.
      *
      * The function just calls strerror(), but temporarily switches, if needed,
      * to the C locale */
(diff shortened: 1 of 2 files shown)


tonycoz:
In what way does this leak?  The `SAVEFREEPV()` should ensure the memory is eventually freed.

The change does free earlier, which is an improvement, but I don't see how it leaks.

Since both callers immediately `sv_setpv()` the result of the `my_strerror()` it might be better pass that SV to be modified.  I can prepare such a change if you prefer.

Leont:
> In what way does this leak?

I suspect this confusion is because `save_pv*` sounds like it would interact with the save stack, instead of being our `strdup`

khwilliamson:
On 7/27/22 20:01, Leon Timmermans wrote:
>     In what way does this leak?
> 
> I suspect this confusion is because |save_pv*| sounds like it would 
> interact with the save stack, instead of being our |strdup|
> 

I have always thought something allocated with savepv() needed to be 
freed with Safefree().  I'm wrong?

tonycoz:
`SAVEFREEPV()` pushes a `SAVEt_FREEPV` entry on the save stack, when that's popped, the supplied PV is freed with `Safefree()`:
```
        case SAVEt_FREEPV:
            a0 = ap[0];
            Safefree(a0.any_ptr);
            break;
```

khwilliamson:
I am uncertain who is confused here.  

The difference listing in this commit is misleading, not showing enough context.   But I did say in the commit message why the problem.  `errstr` is created by a `savepv()`, so it is malloc'd memory that needs to be freed at some point.  Prior to this commit,  the `SAVEFREEPV` made a copy of that copy to return to the calling function.  Yes that copy of a copy will be freed, but the original copy,  `errstr` is left dangling, hence leaking.  That is my mental model of what is going on.  I've written a lot of code based on that model, so if it is wrong, I just don't know what to do.

There are other ways to solve this leak, of course. But, this seemed the easiest, and I thought sufficient, since this function is private to the perl core called in two places (one of which is OS2) both in the same file.  That is unlikely to change.  If this were called more generally, I would have used the `SAVEFREEPV` method for safety.

A future commit will extend the interface, returning the UTF-8ness of `errstr` as well.

tonycoz:
SAVEFREEPV does not make a copy of the string pointed at by the supplied pointer, only a copy of the pointer itself.

It pushes the pointer supplied onto the save stack, at some point later that pointer is supplied to `Safefree()`, from perlguts:
```
=item C<SAVEFREEPV(p)>

The chunk of memory which is pointed to by C<p> is Safefree()ed at the
end of I<pseudo-block>.
```

PR #19998, opened 2022-07-27
Minor optype neatenings

leonerd:
A few miscellaneous fixings of small details, ahead of some larger work that I intend to do to tidy up the way some of these op types work.

### DIFF
--- a/dump.c
+++ b/dump.c
@@ -1255,7 +1255,7 @@ S_do_op_dump_bar(pTHX_ I32 level, UV bar, PerlIO *file, const OP *o)
         /* with ITHREADS, consts are stored in the pad, and the right pad
          * may not be active here, so skip */
         S_opdump_indent(aTHX_ o, level, bar, file, "SV = %s\n",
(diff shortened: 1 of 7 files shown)


leonerd:
I have some larger things to work on that are depending on this to be merged first.
If there are no objections I shall merge it on Wednesday.

jkeenan:
> I have some larger things to work on that are depending on this to be merged first. If there are no objections I shall merge it on Wednesday.

Do you know/suspect if there are any important CPAN distros that might be affected by these changes?

bram-perl (approved):
For commit 'Define the remaining convenience cMETHOP* macros' :  one thing *I* would do is include the list of the added methods in the commit message (and with what they match, basically the list of added `#define` but a bit more readable)

leonerd:
> Do you know/suspect if there are any important CPAN distros that might be affected by these changes?

None of this should be visible to CPAN modules at all. It's minor adjustments of deep internals; no interface-visible changes.

Issue #18981, opened 2021-07-14
# parsed incorrectly in qr/\Q...\E/x


choroba:
**Where**
perlre

**Description**
The documentation currently states:
> Use of /x means that if you want real whitespace or "#" characters in the pattern (outside a bracketed character class, which is unaffected by /x), then you'll either have to escape them (using backslashes or \Q...\E) o

But in reality, `\Q...\E` doesn't help:
```perl
"a" =~ /( \Q # abc \E) /x;
```
leads to 
```
Unmatched ( in regex; marked by <-- HERE in m/(  <-- HERE \ \#\ abc\ \\E\)\ / 
```
The documentation should be fixed, unless it's a bug in the implementation (I'm not sure).

choroba:
It seems it's not a bug in the documentation but in the implementation.
```perl
print "a#b" =~ /( \Q\#\E )/x; # no output
```
It seems there's no way how to match # inside of \Q...\E.

choroba:
Reference: https://stackoverflow.com/q/68378838/1030675

Grinnz:
Relevant: https://github.com/Perl/perl5/issues/13257

Grinnz:
Note that this is not a problem for `\Q$var\E`, only for literal patterns within the `\Q` escape sequence.

Issue #11145, opened 2011-02-21
Bug report: error with \L, \l, \U and \u operators


p5pRT:
Migrated from [rt.perl.org#84578](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=84578) (status was 'open')

Searchable as RT84578$

p5pRT:
### From g+i@gameintellect.com

Hello Perl maintainers\,

 does the operators \\L\, \\l\, \\U\, and \\u have right to left associativity\,
or vice versa? I think\, the operators must have right to left associativity as =\.
Have the operators such property as a priority?

print "\\u\\LdD\\n";   \# Dd    It seems\, first works \\L\, then \\u\. Right\.
print "\\u\\la\\n";    \# A     First \\l\, then \\u\. Right
print "\\l\\ua\\n";    \# a     First \\u\, then \\l\. Right
print "\\L\\udD\\n";   \# Dd    It seems\, first works \\L\, then \\u\. I think\, it's a bug\!
print lc "\\udD\\n";  \# dd    Yes\, the result differ from previous line\!
print "\\LdD\\udD\\n"; \# dddd  It seems\, first works \\u\, then \\L\, hm\.\.\.

print "\\L\\Ua\\n";    \# Syntax error\, oops\!
print "\\U\\La\\n";    \# Syntax error
print "\\L\\La\\n";    \# Syntax error
print "\\u\\la\\n";    \# A     Right
print "\\u\\ua\\n";    \# A     Right

\-\-
Regards\,
 Serge

p5pRT:
### From @abigail

On Mon\, Feb 21\, 2011 at 06&#8203;:57&#8203;:58AM \-0800\, Serge wrote&#8203;:
> \# New Ticket Created by  Serge 
> \# Please include the string&#8203;:  \[perl \#84578\]
> \# in the subject line of all future correspondence about this issue\. 
> \# \<URL&#8203;: http&#8203;://rt\.perl\.org/rt3/Ticket/Display\.html?id=84578 >
> 
> 
> Hello Perl maintainers\,
> 
>  does the operators \\L\, \\l\, \\U\, and \\u have right to left associativity\,
> or vice versa? I think\, the operators must have right to left associativity as =\.
> Have the operators such property as a priority?


All that's being said about priorities in "Gory details of parsing quoted
constructs" in perlop is&#8203;:

&nbsp;  All operations above are performed simultaneously\, left to right\.

Which is vague enough that any behaviour described below can be explained
[...]

Issue #8848, opened 2007-03-26
unpack use internal string representation (utf8)


p5pRT:
Migrated from [rt.perl.org#42102](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=42102) (status was 'resolved')

Searchable as RT42102$

p5pRT:
### From powerman@powerman.asdfGroup.com

#### Created by powerman@powerman.asdfGroup.com



pcalc> $s1 = $s = "\\xAA\\xBB\\xCC"; utf8&#8203;::upgrade $s1
pcalc> x map \{sprintf "%x"\, $\_\} unpack "CCC"\, $s
$VAR1 = 'aa';
$VAR2 = 'bb';
$VAR3 = 'cc';
pcalc> x map \{sprintf "%x"\, $\_\} unpack "CCC"\, $s1
$VAR1 = 'c2';
$VAR2 = 'aa';
$VAR3 = 'c2';
pcalc> x map \{sprintf "%x"\, $\_\} unpack "n"\, $s
$VAR1 = 'aabb';
pcalc> x map \{sprintf "%x"\, $\_\} unpack "n"\, $s1
$VAR1 = 'c2aa';

Actually I got this issue by using JSON&#8203;::XS and Compress&#8203;::Zlib\.
I've received HTTP reply from web server\, packed it into JSON\,
transfer to another part of my application\, it unpack from JSON
\(at this point my bytes become marked 'UTF8' as Devel&#8203;::Peek show\)
and Compress&#8203;::Zlib fail to ungzip this HTTP reply because it use&#8203;:

&nbsp;    sub \_removeGzipHeader
&nbsp;        \.\.\.
&nbsp;        unpack \('CCCCVCC'\, $$string\);

JSON&#8203;::XS author say it's bug in perl and ask me to send bugreport\.

pcalc> $s = "\\xAA\\xBB\\xCC"; $s1=from\_json\(to\_json\(\[$s\]\)\)\->\[0\];
pcalc> Dump $s
SV = PV\(0x1040cc40\) at 0x1051c060
&nbsp;  REFCNT = 1
&nbsp;  FLAGS = \(POK\,pPOK\)
&nbsp;  PV = 0x105e04d8 "\\252\\273\\314"\\0
&nbsp;  CUR = 3
&nbsp;  LEN = 4
pcalc> Dump $s1
SV = PVMG\(0x1051c4e8\) at 0x1051c084
&nbsp;  REFCNT = 1
[...]

PR #19999, opened 2022-07-27
perlop: clarify \U, \L, \F behaviour

mfontani:
... as they do not stop "at \E or end of string" but are also
stopped by another \U, \L or \F.

See also:
    https://www.nntp.perl.org/group/perl.perl5.porters/2022/07/msg264490.html

### DIFF
--- a/pod/perlop.pod
+++ b/pod/perlop.pod
@@ -1790,9 +1790,9 @@ X<\l> X<\u> X<\L> X<\U> X<\E> X<\Q> X<\F>
 
     \l		lowercase next character only
     \u		titlecase (not uppercase!) next character only
-    \L		lowercase all characters till \E or end of string
-    \U		uppercase all characters till \E or end of string
-    \F		foldcase all characters till \E or end of string
+    \L		lowercase all characters till \U, \F, \E or end of string
+    \U		uppercase all characters till \L, \F, \E or end of string
+    \F		foldcase all characters till \U, \L, \E or end of string
     \Q          quote (disable) pattern metacharacters till \E or
                 end of string
     \E		end either case modification or quoted section
--- a/t/op/lc.t
+++ b/t/op/lc.t
@@ -17,7 +17,7 @@ BEGIN {
 
 use feature qw( fc );
 
-plan tests => 139 + 2 * (5 * 256) + 17;
+plan tests => 139 + 2 + 2 * (5 * 256) + 17;
 
 is(lc(undef),	   "", "lc(undef) is ''");
 is(lcfirst(undef), "", "lcfirst(undef) is ''");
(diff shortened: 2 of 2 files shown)


khwilliamson:
People may find https://github.com/Perl/perl5/issues/11145 instructive

mfontani:
@khwilliamson my goal here is to "at least" document (and have tests for) the current non-stacking behaviour, so that it's clearer that's what's happening / isn't as nebulous as before.

It's hopefully a simple change to merge.

bram-perl:
Looking at some tickets; this change in docs was also suggested in #19670 
but a reply (from @khwilliamson) was "It's worse then that";
Unfortunately the reply didn't specify in what way it worse/in what way documenting it like that would be incorrect :(

khwilliamson:
But, I added a follow-up immediately after to that reply that did give concrete examples:

Here are some tickets https://github.com/Perl/perl5/issues/8848 https://github.com/Perl/perl5/issues/11145 https://github.com/Perl/perl5/issues/18981

iabyn:
On Wed, Jul 27, 2022 at 04:46:51AM -0700, Marco Fontani wrote:
> ... as they do not stop &quot;at \E or end of string&quot; but are also
> stopped by another \U, \L or \F.

Given that we're currently unclear as to what the correct behaviour should
be in all circumstances, and we may well change things once we are clear,
I don't think we want to be fixing the current behaviour in the
documentation.

-- 
everyone's satisfaction.

mfontani:
Sounds good, closing this then.

bram-perl:
Having documented some(/all?) the quirks/caveats/... for case modifiers[^1] it puts me in a better position to comment.

The text is mostly correct but based on the **current** behavior it's missing the caveats:

1. `\U\lfoo` is transformed into `\l\Ufoo`;
2. `\L\ufoo` is transformed into `\u\Lfoo`;
3. A `\E` which follows immediately after a `\U`, `\L`, `\F`, `\Q`, `\u`, `\l` causes both symbols to be completely ignored. Examples:
    * `\Ufoo\L\Ebar` is parsed as `\Ufoobar` -> There no longer is a `\L` or `\E`!
    * applying 1. and 2.: `\Ufoo\L\u\Ebar` is first transformed into `\Ufoo\u\L\Ebar` and then into `\Ufoo\ubar` -> Again there is no longer a `\L` or `\E`

So if after 1., 2., 3. there are still two occurrences of `\U`, `\L`, `\F` in the string then the second occurrence will terminate the first `\U`, `\L`, `\F`.

Also note: a `\L` also ends a previous `\L`, i.e. `\LFOO\LBAR\EBAZ` is equal to `foobarBAZ` (same applies for `\U` and `\F`)

[^1]: see #20042

Issue #19989, opened 2022-07-24
${10}, ${11}, ... does not work when 'use utf8' and 'use strict' is in use


bram-perl:
**Description**

This is basically the same issue as #12948, in #12948 fixes were applied[^1] but these
only work when `use utf8` is not in use[^2].

In addition to that: under `use utf8` it allows things like `${010}` which were also forbidden in #12948.

**Steps to reproduce**

1. Using `${10}`

        $ ./perl -Ilib -le 'use strict; use utf8; "abcdefghijkl" =~ m/(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)/ or die; print ${10};'
        Can't use string ("10") as a SCALAR ref while "strict refs" in use at -e line 1.

        $ ./perl -Ilib -le 'use strict;           "abcdefghijkl" =~ m/(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)/ or die; print ${10};'
        j

2. Using `${010}`

        $ ./perl -Ilib -e 'use utf8; print ${010};'
        $ ./perl -Ilib -e '          print ${010};'
        Numeric variables with more than one digit may not start with '0' at -e line 1.

**Additional information**

Looking at the second set of changes applied commit 60267e1d0e12bb5bdc88c62a18294336ab03d4b8 (L10174-10187) with a bit of context (L10166-10173) shows that the changes are made inside the `else` block that belongs to `if (is_utf8) {` which is why it falls back to the old (incorrect) behaviour under `use utf8`.

        10166        if (is_utf8) {
        10167            const STRLEN skip = UTF8SKIP(s);
        10168            STRLEN i;
        10169            d[skip] = '\0';
        10170            for ( i = 0; i < skip; i++ )
        10171                d[i] = *s++;
        10172        }
        10173        else {
        10174            *d = *s++;
        10175            /* special case to handle ${10}, ${11} the same way we handle ${1} etc */
        10176            if (isDIGIT(*d)) {
        10177                bool is_zero= *d == '0' ? TRUE : FALSE;
        10178                char *digit_start= d;
        10179                while (s < PL_bufend && isDIGIT(*s)) {
        10180                    d++;
        10181                    if (d >= e)
        10182                        Perl_croak(aTHX_ "%s", ident_too_long);
        10183                    *d= *s++;
        10184                }
        10185                if (is_zero && d - digit_start > 1)
        10186                    Perl_croak(aTHX_ ident_var_zero_multi_digit);
[...]

Issue #19986, opened 2022-07-23
`${00}` should not be allowed


bram-perl:
**Description**

In commit 60267e1d0e12bb5bdc88c62a18294336ab03d4b8 (by @demerphq ) for #12948 code was added to disallow `$00`, `$000`, `${000}`, ...
The perldiag entry added for it:

        Numeric variables with more than one digit may not start with '0'
                (F) The only numeric variable which is allowed to start with a 0 is $0,
                and you mentioned a variable that starts with 0 that has more than one digit.
                You probably want to remove the leading 0, or if the intent was to express a
                variable name in octal you should convert to decimal.

Using blead (b5df4e081):

        $ ./perl -e '$0;'
        $ ./perl -e '${0};'
        $ ./perl -e '$00;'
        Numeric variables with more than one digit may not start with '0' at -e line 1.
        $ ./perl -e '${00};'
        $ ./perl -e '${000};'
        Numeric variables with more than one digit may not start with '0' at -e line 1.
        $ ./perl -e '$000;'
        Numeric variables with more than one digit may not start with '0' at -e line 1.

Behaviour of `$00` and `${00}` is inconsistent.
I believe the intent was to also disallow `${00}`.

Looking at the code there appears to be an off-by-one error;

In toke.c S_scan_ident there are two similar loops to handle this case:

        10113    if (isDIGIT(*s)) { /* handle $0 and $1 $2 and $10 and etc */
        10114        bool is_zero= *s == '0' ? TRUE : FALSE;
        10115        char *digit_start= d;
        10116        *d++ = *s++;
        10117        while (s < PL_bufend && isDIGIT(*s)) {
        10118            if (d >= e)
        10119                Perl_croak(aTHX_ "%s", ident_too_long);
        10120            *d++ = *s++;
        10121        }
        10122        if (is_zero && d - digit_start > 1)
        10123            Perl_croak(aTHX_ ident_var_zero_multi_digit);
        10124    }
        10125    else {  /* See if it is a "normal" identifier */
        10126        parse_ident(&s, &d, e, 1, is_utf8, FALSE, TRUE);
        10127    }
        10128    *d = '\0';

and

        10173        else {
        10174            *d = *s++;
        10175            /* special case to handle ${10}, ${11} the same way we handle ${1} etc */
        10176            if (isDIGIT(*d)) {
        10177                bool is_zero= *d == '0' ? TRUE : FALSE;
        10178                char *digit_start= d;
[...]

Issue #12948, opened 2013-05-07
On RHS of s///, ${9} works but ${10} does not


p5pRT:
Migrated from [rt.perl.org#117907](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=117907) (status was 'open')

Searchable as RT117907$

p5pRT:
### From @epa

#### Created by @epa



I found this surprising&#8203;:

&nbsp;    \#\!/usr/bin/perl
&nbsp;    use 5\.016;
&nbsp;    use warnings;
&nbsp;    use strict;
&nbsp;    $\_ = 'a' x 10;
&nbsp;    s/\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)/$\{9\}/ or die;
&nbsp;    say $\_;
&nbsp;    $\_ = 'a' x 10;
&nbsp;    s/\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)\(a\)/$\{10\}/ or die;
&nbsp;    say $\_;

Why is $\{9\} accepted on the RHS of a substitution but $\{10\} not?
They should both be or neither be\.

The same applies in ordinary code&#8203;:

&nbsp;    say $\{9\};    \# ok
&nbsp;    say $\{10\};   \# fails
&nbsp;    

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=low

Site configuration information for perl 5.16.3:

Configured by Red Hat, Inc. at Thu Apr 11 09:48:29 UTC 2013.

Summary of my perl5 (revision 5 version 16 subversion 3) configuration:
   
  Platform:
    osname=linux, osvers=2.6.32-358.2.1.el6.x86_64, archname=x86_64-linux-thread-multi
    uname='linux buildvm-08.phx2.fedoraproject.org 2.6.32-358.2.1.el6.x86_64 #1 smp wed feb 20 12:17:37 est 2013 x86_64 x86_64 x86_64 gnulinux '
[...]

PR #20000, opened 2022-07-27
toke.c - improve handling of $00 and ${00}

demerphq:
In 60267e1d0e12bb5bdc88c62a18294336ab03d4b8 I patched toke.c to refuse
`$00` but did not properly handle `${00}` and related cases when the code
was unicode. Part of the reason was the confusing macro
VALID_LEN_ONE_IDENT() which despite its name does not restrict what it
matches to things which are one character long.

Since the VALID_LEN_ONE_IDENT() macro is used in only one place and its
name and placement is confusing I have moved it back into the code
inline as part of this fix. I have also added more comments about what
is going on, and moved the related comment directly next to the code
that it affects. If it moved out of this code then we should think of a
better name and be more careful and clear about checking things like
length. I would argue the logic is used to parse what might be called a
variable "description", and thus it is not identical to code which might
validate an actual parsed variable name. Eg, ${^Var} is a description of
the variable whose "name" is "\026ar". The exception of course is $^
whose name actually is "^".

A byproduct of this change is that the logic to detect duplicated
leading zeros is now quite a bit simpler.

This includes more tests for leading zero checks.

See Issue #12948, Issue #19986, and Issue #19989.

### DIFF
--- a/toke.c
+++ b/toke.c
@@ -10076,25 +10076,14 @@ Perl_scan_word(pTHX_ char *s, char *dest, STRLEN destlen, int allow_package, STR
     return s;
 }
 
-/* Is the byte 'd' a legal single character identifier name?  'u' is true
- * iff Unicode semantics are to be used.  The legal ones are any of:
- *  a) all ASCII characters except:
- *          1) control and space-type ones, like NUL, SOH, \t, and SPACE;
- *          2) '{'
- *     The final case currently doesn't get this far in the program, so we
- *     don't test for it.  If that were to change, it would be ok to allow it.
- *  b) When not under Unicode rules, any upper Latin1 character
- *  c) Otherwise, when unicode rules are used, all XIDS characters.
- *
- *      Because all ASCII characters have the same representation whether
- *      encoded in UTF-8 or not, we can use the foo_A macros below and '\0' and
- *      '{' without knowing if is UTF-8 or not. */
-#define VALID_LEN_ONE_IDENT(s, e, is_utf8)                                  \
-    (isGRAPH_A(*(s)) || ((is_utf8)                                          \
-                         ? isIDFIRST_utf8_safe(s, e)                        \
-                         : (isGRAPH_L1(*s)                                  \
-                            && LIKELY((U8) *(s) != LATIN1_TO_NATIVE(0xAD)))))
 
+/* scan s and extract an identifier ($var) from it if possible
+ * into dest.
+ * XXX: This function has subtle implications on parsing, and
+ * changing how it behaves can cause a variable to change from
+ * being a run time rv2sv call or a compile time binding to a
+ * specific variable name.
+ */
 STATIC char *
 S_scan_ident(pTHX_ char *s, char *dest, STRLEN destlen, I32 ck_uni)
 {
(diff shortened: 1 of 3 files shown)


toke.c
```diff
@@ -10172,22 +10185,16 @@ S_scan_ident(pTHX_ char *s, char *dest, STRLEN destlen, I32 ck_uni)
...
-                    Perl_croak(aTHX_ ident_var_zero_multi_digit);
-            }
             d[1] = '\0';
         }
     }
+
+    /* make sure that we dont have a var starting with two zero digits
+     * which is not allowed as it introduces confusion with octal and
+     * decimal. */
+    if (*d == '0' && *s == '0')
```
bram-perl: This check doesn't look entirely right..
What it will do is disallow `${00}` (and `${000}`, `${0000}`, ...) but the earlier check (lines 10117-10127) disallows things like: `$00`, `$010`, `$09324`, ...  i.e. the second digit doesn't have to be zero.

demerphq:
On Wed, 27 Jul 2022 at 14:10, Bram ***@***.***> wrote:

>
> In toke.c <https://github.com/Perl/perl5/pull/20000#discussion_r930983187>
> :
>
> >              d[1] = '\0';
>          }
>      }
> +
> +    /* make sure that we dont have a var starting with two zero digits
> +     * which is not allowed as it introduces confusion with octal and
> +     * decimal. */
> +    if (*d == '0' && *s == '0')
>
> This check doesn't look entirely right..
> What it will do is disallow ${00} (and ${000}, ${0000}, ...) but the
> earlier check (lines 10117-10127) disallows things like: $00, $010, $09324`,
> ... i.e. the second digit doesn't have to be zero
>
 Doh. Good catch, thanks bram. Fixup coming.

Yves
-- 

toke.c
demerphq: pushed a fixup patch with more tests.

toke.c
```diff
@@ -10172,24 +10185,27 @@ S_scan_ident(pTHX_ char *s, char *dest, STRLEN destlen, I32 ck_uni)
...
+	char *digit_start= d;
+	while (s < PL_bufend && isDIGIT(*s)) {
+	    d++;
+	    if (d >= e)
+		Perl_croak(aTHX_ "%s", ident_too_long);
+	    *d= *s++;
+	}
+	if (is_zero && d - digit_start > 1)
+	    Perl_croak(aTHX_ ident_var_zero_multi_digit);
+        d[1] = '\0';
```
bram-perl: Mixed tabs/spaces indentation?

demerphq: I cleaned and squashed the commits so there shouldn't be any mixed tabs. Do you still see any?

demerphq: Ah, the new compare function shows it. Thanks for the catch, seems to be cleaned up now.

bram-perl (commented):
The code looks good to me;
Just one additional suggestion: maybe amend the commit description and mention something about the change in behavior for `${10}`, ${11}`, ... (i.e. that they now behave as expected under `use strict`)

demerphq:
I wanted to see if my understand of scan_ident was correct, so I made it handle binary and hex eg, `${0x10}` and `${0b10000}`. Both of these end up returning the same GV that `$16` and `${16}` do. If that proves controversial then we can drop that patch.

bram-perl:
> I wanted to see if my understand of scan_ident was correct, so I made it handle binary and hex eg, `${0x10}` and `${0b10000}`. Both of these end up returning the same GV that `$16` and `${16}` do. If that proves controversial then we can drop that patch.

I think it would be best to drop it;
Even tho just some small comments:
- there is a `0o` prefix for octal (since perl v5.34.0)  (`0o10` = 8)
- in the commit message of the patch you correctly note that `$0x10` is not allowed; one side remark on that: `$::0x10` (for some reason) is allowed.

demerphq:
On Thu, 28 Jul 2022 at 20:39, Bram ***@***.***> wrote:

> I wanted to see if my understand of scan_ident was correct, so I made it
> handle binary and hex eg, ${0x10} and ${0b10000}. Both of these end up
> returning the same GV that $16 and ${16} do. If that proves controversial
> then we can drop that patch.
>
> I think it would be best to drop it;
> Even tho just some small comments:
>
>    - there is a 0o prefix for octal (since perl v5.34.0) (0o10 = 8)
>
> Ah, ok, ill add that and push this part in a separate PR for consideration
separately.

>
>    - in the commit message of the patch you correctly note that $0x10 is
>    not allowed; one side remark on that: $::0x10 (for some reason) is
>    allowed.
>
> It is not so much that $0x10 is not allowed as it is that it will be (and
should be?) parsed as $0 x 10 as $0 is the only var allowed to start with
0. That $::0x10 is not parsed similarly is because (I guess) that in
$::0x10 the 0 is not the first character of the identifier, which is
actually $main::0x10. Notice it does not resolve to a digit var, it
resolves to a var named "0x10", unlike ${0x10} which resolves to the var
$16. As far as I know the name after the colons does not have to follow
IDFIRST conventions.

perl -MO=Concise -e'print $::0x10'
6  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 1 -e:1) v:{ ->3
5     <@> print vK ->6
3        <0> pushmark s ->4
-        <1> ex-rv2sv sK/1 ->5
4           <$> gvsv(*0x10) s ->5

$ perl -MO=Concise -MDevel::Peek -e'Dump(${0x10})'
6  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 44 -e:1) v:{ ->3
5     <2> Devel_Peek_Dump vK/1 ->6
4        <1> rv2sv sK/1 ->5
-           <@> scope sK ->4
-              <;> ex-nextstate(main 45 -e:1) v ->3
3              <$> const(IV 16) s ->4
-e syntax OK

Anyway, I'll tweak the PR.

Yves


-- 

demerphq:
K. Removed the controversial patch.

Issue #19997, opened 2022-07-25
Evaluation order of `EXPR->(...)` is confusing


leonerd:
While Perl doesn't strictly *guarantee* evaluation order, it usually does pretty well at not surprising you, doing the right thing in most circumstances. In particular, when fetching elements out of array or hash references, or invoking methods:

```
eval: (do { say "First"; \my @a })->[say "Second"]
First
Second

eval: (do { say "First"; \my %h })->{say "Second"}
First
Second

eval: (do { say "First"; undef })->meth(say "Second")
First
Second
ERR: Can't call method "meth" on an undefined value at (eval 15) line 1.
```
(Ignore the fact that it died in this last example; the fact remains it evaluated the LHS and printed `"First"` before it evaluated the RHS and printed `"Second"`)

This neatness does not hold when invoking code references:

```
eval: (do { say "First"; sub {} })->(say "Second")
Second
First
```

The casual observer would expect that it would print "First" then "Second" in that order, as it does for the other examples; but yet it does not.

This may be regarded as "just a quirk", but it does have an annoying consequence when we consider https://github.com/Perl/RFCs/pull/23#issuecomment-1194661578. A satisfactory solution in the optional-chaining case might involve changing the way that regular (i.e. non-optional) code ref invocation also works, to fix this evaluation order problem.

leonerd:
I suspect fixing this would be relatively simple, involving a flag on `OP_ENTERSUB` to tell it to look at the other end of the stack for the CV to invoke, and shuffling the way the optree is generated in these cases.

However, having done that it remains to be seen whether there's any breakage in core or CPAN, from code that somehow relies on this odd order of evaluation. There *probably* isn't much of interest, but I wouldn't be surprised if some weird cornercase somewhere somehow makes use of it.

garu:
It may be important to realize that the same behaviour happens on regular subroutine calls too, not just refs:

```
▶ perl -Mv5.36 -E ' my $x; $x->(die q(oops!)) '
oops! at -e line 1.

▶ perl -Mv5.36 -E ' some_function(die q(oops!)) '
oops! at -e line 1.
```

Both of them should have, in my opinion, failed with _"Can't use an undefined value as a subroutine reference"_ and _"Undefined subroutine &main::some_funcion called"_, respectively, without having called arguments at all. Which is exactly what would have happened if no side-effect argument had been passed.

[...]

PR #20001, opened 2022-07-27
Remove OPf_SPECIAL flag from OP_ENTERSUB

ilmari:
It was being set for lvalue sub assignments and method calls implied by `use` and attributes on lexical variables, but nothing appears to ever have used that flag for anything.

The last thing that actually used it was `do SUBNAME(LIST)` notation, which signalled that `B::Deparse` had to emit a `do ` before the subname. That functionality was removed in 5.20 (commit 8c74b41425572faeb638f1269025b59d0785794f), but `B::Deparse` missed the memo.

This frees up the `OPf_SPECIAL` flag for some other purpose, see https://github.com/Perl/perl5/issues/19997#issuecomment-1196987459.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -358,6 +358,13 @@ actual behaviour (it suppresses a warning that would otherwise be generated
 about ambiguous names), in order to be less confusing with a possible upcoming
 feature.
 
+=item *
+
+The C<OPf_SPECIAL> flag is no longer set on the C<OP_ENTERSUB> op
+constructed to call the C<VERSION>, C<import> and C<unimport> methods
+as part of a C<use> statement and attribute application, nor when
+assigning to an C<:lvalue> subroutine.
+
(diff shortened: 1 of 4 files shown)


leonerd (approved):
LGTM, presuming CI is happy.

jkeenan:
Branch passes tests on FreeBSD-12, using clang10 as compiler.

pod/perldelta.pod
```diff
@@ -358,6 +358,13 @@ actual behaviour (it suppresses a warning that would otherwise be generated
 about ambiguous names), in order to be less confusing with a possible upcoming
 feature.
 
+=item *
+
+The C<OPf_PRIVATE> flag is no longer set on the C<OP_ENTERSUB> op
```
leonerd: s/OPf_PRIVATE/OPf_SPECIAL/

PR #20003, opened 2022-07-27
vutil.c: Simplify locale handling

khwilliamson:
I read the code over and realized that there was a much simpler way to do things.


Leont:
Changes to vutil should probably be applied to [version.pm](https://github.com/Perl/version.pm/) first.

khwilliamson:
ok, but note that version is already customized in blead

Leont:
> ok, but note that version is already customized in blead

Sadly, that's permanent: https://metacpan.org/dist/version/source/lib/version.pm#L15-60

PR #20008, opened 2022-07-28
File::Find: fix "follow => 1" on Windows

### LINKED ISSUE #19995: File-Find-1.39 (and later) breaks Module::Pluggable and Module::Find
Module:
File::Find
Version 1.37 (which shipped with perl-5.32.x) is ok. I can't find version 1.38 in any stable perl release.
The problem is arising only with version 1.39 (which shipped with perl-5.34.0) and later.

**Description**

Module::Pluggable and Module::Find fail tests because of the new changes wrt Win32 symlinks support in File::Find.
Both MSVC-built and MinGW-built perls are affected in the same way.

**Steps to Reproduce**

cpan -i Module::Pluggable
cpan -i Module::Find

The question is:
Should Module::Pluggable and Module::Find be patched to accommodate these recent changes to File::Find, or should File::Find be patched such that it doesn't break these 2 cpan modules in the first place ?

The Module-Pluggable bug report is at https://github.com/simonwistow/Module-Pluggable/issues/24.
This patch to Module-Pluggable-Object.pm allows Module::Pluggable to build successfully against these problematic versions of File::Find:
```
--- Object.pm_orig	2022-07-25 11:42:24 +1000
+++ Object.pm	2022-07-25 12:12:02 +1000
@@ -66,6 +66,7 @@
     $self->{'on_instantiate_error'} ||= sub { my ($plugin, $err) = @_; carp "Couldn't instantiate $plugin: $err"; return 0 };
 
     # default whether to follow symlinks
+    $self->{'follow_symlinks'} = 0 if ($^O eq 'MSWin32' && !exists $self->{'follow_symlinks'});
     $self->{'follow_symlinks'} = 1 unless exists $self->{'follow_symlinks'};
 
     # check to see if we're running under test
```
There's a bug report for Module::Find at https://github.com/crenz/Module-Find/issues/9 .
No patch for that one, yet, AFAIK.

I'm on a Windows 7 machine - which might be relevant.
I'm not sure which Windows version is being used by the others who have hit this issue.
As the latest Strawberry Perl is still at 5.32.1 (File-Find-1.37), the number of these "others" is probably quite small.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 36 subversion 0) configuration:

  Platform:
    osname=MSWin32
    osvers=6.1.7601
    archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -DWIN64 -D_WIN32_WINNT=0x0601 -fdiagnostics-color=never -
DPERL_TEXTMODE_SCRIPTS -DMULTIPLICITY -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_M
INGW_ANSI_STDIO -fwrapv -fno-strict-aliasing -mms-bitfields'
    optimize='-O2'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='11.3.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='long long'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='g++'
    ldflags ='-s -L"C:\perl-5.36.0\lib\MSWin32-x64-multi-thread\CORE" -L"C:\_64\
winlibs-gcc-1130\mingw64\lib" -L"C:\_64\winlibs-gcc-1130\mingw64\x86_64-w64-ming
w32\lib" -L"C:\_64\winlibs-gcc-1130\mingw64\lib\gcc\x86_64-w64-mingw32\11.3.0"'
    libpth=C:\_64\winlibs-gcc-1130\mingw64\lib C:\_64\winlibs-gcc-1130\mingw64\x
86_64-w64-mingw32\lib C:\_64\winlibs-gcc-1130\mingw64\lib\gcc\x86_64-w64-mingw32
\11.3.0 C:\_64\msys_1130\1.0\local\lib
    libs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladvapi3
2 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lversio
n -lodbc32 -lodbccp32 -lcomctl32
    perllibs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladv
api32 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lve
rsion -lodbc32 -lodbccp32 -lcomctl32
    libc=
    so=dll
    useshrplib=true
    libperl=libperl536.a
    gnulibc_version='10.0.0'
  Dynamic Linking:
    dlsrc=dl_win32.xs
    dlext=dll
    d_dlsymun=undef
    ccdlflags=' '
    cccdlflags=' '
    lddlflags='-shared -s -L"C:\perl-5.36.0\lib\MSWin32-x64-multi-thread\CORE" -
L"C:\_64\winlibs-gcc-1130\mingw64\lib" -L"C:\_64\winlibs-gcc-1130\mingw64\x86_64
-w64-mingw32\lib" -L"C:\_64\winlibs-gcc-1130\mingw64\lib\gcc\x86_64-w64-mingw32\
11.3.0"'


  Compile-time options:
    HAS_TIMES
    HAVE_INTERP_INTERN
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_IMPLICIT_SYS
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
  Built under MSWin32
  Compiled at May 28 2022 11:14:45
  @INC:
    C:/perl-5.36.0/site/lib/MSWin32-x64-multi-thread
    C:/perl-5.36.0/site/lib
    C:/perl-5.36.0/lib/MSWin32-x64-multi-thread
    C:/perl-5.36.0/lib
```

xenu:
File::Find's code expects unix-style paths and it manipulates them using
basic string operations. That code is very fragile, and ideally we
should make it use File::Spec, but that would involve rewriting almost
the whole module.

Instead, we made it convert backslashes to slashes and handle drive
letters.

Note from xenu: this commit was adapted from the PR linked in this
blogpost[1]. I have squashed it, written the commit message and slightly
modified the code.

[1] - https://www.nu42.com/2021/09/canonical-paths-file-find-way-forward.html

Fixes #19995

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -4265,6 +4265,7 @@ ext/File-DosGlob/DosGlob.xs		Win32 DOS-globbing module
 ext/File-DosGlob/lib/File/DosGlob.pm	Win32 DOS-globbing module
 ext/File-DosGlob/t/DosGlob.t		See if File::DosGlob works
 ext/File-Find/lib/File/Find.pm	Routines to do a find
+ext/File-Find/t/correct-absolute-path-with-follow.t
 ext/File-Find/t/find.t		See if File::Find works
 ext/File-Find/t/lib/Testing.pm		Functions used in testing File-find
 ext/File-Find/t/taint.t		See if File::Find works with taint
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -172,6 +172,7 @@ use File::Glob qw(:case);
         'DISTRIBUTION' => 'SMUELLER/AutoLoader-5.74.tar.gz',
         'FILES'        => q[cpan/AutoLoader],
         'EXCLUDED'     => ['t/00pod.t'],
+        'CUSTOMIZED'   => ['t/02AutoSplit.t'],
     },
 
     'autouse' => {
--- a/cpan/AutoLoader/t/02AutoSplit.t
+++ b/cpan/AutoLoader/t/02AutoSplit.t
@@ -149,8 +149,12 @@ foreach (@tests) {
 
   if ($args{Files}) {
     $args{Files} =~ s!/!:!gs if $^O eq 'MacOS';
+    $args{Files} =~ s!\\!/!g if $^O eq 'MSWin32';
     my (%missing, %got);
-    find (sub {$got{$File::Find::name}++ unless -d $_}, $dir);
+    find(
+        sub { (my $f = $File::Find::name) =~ s!\\!/!g; $got{$f}++ unless -d $_ },
+        $dir
+    );
     foreach (split /\n/, $args{Files}) {
       next if /^#/;
       $_ = lc($_) if $Is_VMS_lc;
--- a/ext/File-Find/lib/File/Find.pm
+++ b/ext/File-Find/lib/File/Find.pm
@@ -43,11 +43,21 @@ sub contract_name {
     return $abs_name;
 }
 
+sub _is_absolute {
+    return $_[0] =~ m|^(?:[A-Za-z]:)?/| if $Is_Win32;
+    return substr($_[0], 0, 1) eq '/';
+}
+
+sub _is_root {
+    return $_[0] =~ m|^(?:[A-Za-z]:)?/\z| if $Is_Win32;
+    return $_[0] eq '/';
+}
+
 sub PathCombine($$) {
     my ($Base,$Name) = @_;
     my $AbsName;
 
-    if (substr($Name,0,1) eq '/') {
+    if (_is_absolute($Name)) {
         $AbsName= $Name;
     }
     else {
@@ -123,6 +133,7 @@ sub is_tainted_pp {
     return length($@) != 0;
 }
 
+
 sub _find_opt {
     my $wanted = shift;
     return unless @_;
@@ -183,19 +194,17 @@ sub _find_opt {
 
         ($topdev,$topino,$topmode,$topnlink) = $follow ? stat $top_item : lstat $top_item;
 
-        if ($Is_Win32) {
-            $top_item =~ s|[/\\]\z||
-              unless $top_item =~ m{^(?:\w:)?[/\\]$};
-        }
-        else {
-            $top_item =~ s|/\z|| unless $top_item eq '/';
-        }
+        # canonicalize directory separators
+        $top_item =~ s|[/\\]|/|g if $Is_Win32;
+
+        # no trailing / unless path is root
+        $top_item =~ s|/\z|| unless _is_root($top_item);
 
         $Is_Dir= 0;
 
         if ($follow) {
 
-            if (substr($top_item,0,1) eq '/') {
+            if (_is_absolute($top_item)) {
                 $abs_dir = $top_item;
             }
             elsif ($top_item eq $File::Find::current_dir) {
@@ -304,11 +313,7 @@ sub _find_dir($$$) {
     my $tainted = 0;
     my $no_nlink;
 
-    if ($Is_Win32) {
-        $dir_pref
-          = ($p_dir =~ m{^(?:\w:[/\\]?|[/\\])$} ? $p_dir : "$p_dir/" );
-    } elsif ($Is_VMS) {
-
+    if ($Is_VMS) {
         # VMS is returning trailing .dir on directories
         # and trailing . on files and symbolic links
         # in UNIX syntax.
@@ -319,7 +324,7 @@ sub _find_dir($$$) {
         $dir_pref = ($p_dir =~ m/[\]>]+$/ ? $p_dir : "$p_dir/" );
     }
     else {
-        $dir_pref= ( $p_dir eq '/' ? '/' : "$p_dir/" );
+        $dir_pref = _is_root($p_dir) ? $p_dir : "$p_dir/";
     }
 
     local ($dir, $name, $prune);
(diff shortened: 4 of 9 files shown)


xenu:
Note for reviewers: one of the commits converts tabs to spaces, make sure to hide whitespace changes.

sisyphus:
This PR fixes the issue with Module::Pluggable. Thanks @xenu, for digging through all of that.
However, it turns out that the issue with Module::Find is a different one - and not addressed by this PR..

The failing test script in Module::Find is  t/07-symlinks.t.
That file tests to see whether symlinks are available or not by running:
```
SKIP: {
    eval { symlink($dirName, $linkName) };
    skip "Symlinks not supported on this system", 13 if $@;
.....
```
On my Windows systems, with perl-5.32 (File-Find-1.37), $@ is set and all 13 tests are skipped - as is the correct thing to do, IIUC.
But with perl-5.34, 5.36, and this PR, $@ is unset and it is therefore assumed (incorrectly) that symlinking is available.
Of course, that symlink() call has returned 0, but it seems that a return of 0 merely signifies that the call failed - not that symlinking is unavailable.
The symlink documentation says:
```
>perldoc -f symlink
    symlink OLDFILE,NEWFILE
            Creates a new filename symbolically linked to the old filename.
            Returns 1 for success, 0 otherwise. On systems that don't
            support symbolic links, raises an exception. To check for that,
            use eval:

                my $symlink_exists = eval { symlink("",""); 1 };

            Portability issues: "symlink" in perlport.
```
which is the behaviour that t/07-symlinks.t is checking for.
Unfortunately that documentation is, as of File-Find-1.39, no longer correct on Windows systems that don't support symlinks.

Cheers,
Rob

Leont:
That check in Module::Find isn't quite what the docs say it should be, and should be changed IMO.

sisyphus:
> That check in Module::Find isn't quite what the docs say it should be, and should be changed IMO

But even if it strictly adheres to the docs, I still get different results between 5.32 and this PR.

With perl-5.32:
```
C:\>perl -e "$x = symlink('',''); print $x;"
The symlink function is unimplemented at -e line 1.

C:\>
```
With this PR:
```
C:\>perl -e "$x = symlink('',''); print $x;"
0
C:\>
```
IIU the symlink() docs correctly, we should be getting an exception in both cases - given that symlinks are unsupported on this system of mine.

Cheers,
Rob

Leont:
> IIU the symlink() docs correctly, we should be getting an exception in both cases - given that symlinks are unsupported on this system of mine.

The OS supports it, they're just not enabled. Hence symlink returns 0.

xenu:
The best ways to check if you have the required permissions to create a symlink on Windows are:

- Attempt to create a symlink and check if `$!` gives you "operation not permitted" code
- [Win32::IsSymlinkCreationAllowed](https://metacpan.org/pod/Win32#Win32::IsSymlinkCreationAllowed())

blues1875:
I have never used symlinks on Windows before, but I find I can create them on my Windows 10 machine (as Administrator). I did a basic test of File::Find using my Strawberrry Perl 5.32 and 5.36 installations, with File::Find 1.37 and 1.40 respectively. This is what I found:

- File::Find 1.37
  - follow unset: find follows the symlink
  - follow => 0: find follows the symlink
  - follow => 1: find follows the symlink

 - File::Find 1.40
   - follow unset: find does not follow the symlink
   - follow => 0: find does not follow the symlink
   - follow => 1: find follows the symlink

It looks to me like the functionality is correct for 1.40.
I see the documentation for 1.37 has the warning: "This is a no-op on Win32" for the 'follow' option. I'm not sure if that means that the module is not expected to follow symlinks or that the option doesn't work. It seems that the functionality to follow symlinks does work (on Windows 10 at least), but the option to disable it doesn't.
Interestingly, although the behaviour seems to have been fixed in 1.40, the caveat in the documentation is still there.

xenu:
>the caveat in the documentation is still there.

I noticed it too, which is why this PR removes it :)

sisyphus:
> The OS supports it, they're just not enabled. Hence symlink returns 0.

Ok - If no-one beats me to it, I'll update https://github.com/crenz/Module-Find/issues/9 accordingly.
Use of  Win32::IsSymlinkCreationAllowed() is apparently only available beginning perl-5.34.0 but I think it's possible to make use of symlink() in a way that does what is needed here, anyway.

Cheers,
Rob

xenu:
>Use of Win32::IsSymlinkCreationAllowed() is apparently only available beginning perl-5.34.0

That's also the version where support for symlinks was introduced. Basically, when you're on Windows and `$Config{d_symlink}` is true, `Win32::IsSymlinkCreationAllowed()` is guaranteed to be present.

```
my $can_symlinks_be_used = sub {
  return 0 if !$Config{d_symlink};

  if ($^O eq 'MSWin32') {
    require Win32;
    return Win32::IsSymlinkCreationAllowed();
  }
  else {
    return 1;
  }
}->();
```

sisyphus:
Can/will this fix be backported to future 5.34.x and 5.36.x releases ?

PR #20011, opened 2022-07-28
Add (to sv.c) and use new Perl_newSVhek_mortal function

richardleach:
This is a small wrapper function intended to replace use of
    `sv_2mortal(newSVhek(...))`

This provides some small optimization, since `sv_2mortal` performs
a number of checks for conditions that we know cannot be true.


leonerd (approved):
LGTM

PR #20017, opened 2022-07-30
Add PERL_MAGIC_extvalue

Leont:
A new magic type `PERL_MAGIC_extvalue` has been added. This is available for use like `PERL_MAGIC_ext`, but is a value magic: upon localization the new value will not be magical.


regen/mg_vtable.pl
```diff
@@ -202,8 +202,10 @@ BEGIN
                 desc => 'Extra data for restricted hashes' },
      arylen_p => { char => '@', value_magic => 1,
                    desc => 'To move arylen out of XPVAV' },
-     ext => { char => '~', desc => 'Available for use by extensions',
+     ext => { char => '~', desc => 'Variable magic available for use by extensions',
               readonly_acceptable => 1 },
+     value => { char => '^', desc => 'Value magic available for use by extensions',
```
leonerd: I can't help thinking it might want to be called `extvalue`, to point out it's for extensions (i.e. non-core) use. And maybe can the other one be renamed `extvar`, with a back-compat alias?

((I'm lately more of the opinion that things that are badly named ought to be renamed to better names with back-compat wrappers...))

pod/perlguts.pod
```diff
@@ -1562,12 +1565,14 @@ uppercase letter is typically used to represent some kind of composite type
...
-there is no way for normal perl code to corrupt this private information
-(unlike using extra elements of a hash object).
+The C<PERL_MAGIC_ext>, C<PERL_MAGIC_value> and C<PERL_MAGIC_uvar> magic types
+are defined specifically for use by extensions and will not be used by perl
+itself.  Extensions can use C<PERL_MAGIC_ext> or C<PERL_MAGIC_value> magic to
+'attach' private information to variables (typically objects).  This is
+especially useful because there is no way for normal perl code to corrupt this
+private information (unlike using extra elements of a hash object). Unlike
+C<PERL_MAGIC_ext> and C<PERL_MAGIC_uvar>, C<PERL_MAGIC_value> is a value magic,
+meaning that on localization the new value will not be magical.
```
leonerd: The wording of this final sentence feels a bit round-about. Maybe start with "C<PERL_MAGIC_value> is value magic (unlike C<PERL_MAGIC_ext> and C<PERL_MAGIC_uvar>), meaning that ..."

leonerd (changes requested):
Overall good direction. I'm unsure about the naming of it though.

Also, I note that `ext/XP-APItest` has some tests for `PERL_MAGIC_ext` (in t/magic.t). Perhaps one could be added for this new valuemagic as well?

pod/perlguts.pod
Leont: Yeah that does sound better

regen/mg_vtable.pl
Leont: Personally I think `ext` is a terrible name/prefix that doesn't actually explain what it does (though it's still better than `uvar`)

Leont:
> Also, I note that `ext/XP-APItest` has some tests for `PERL_MAGIC_ext` (in t/magic.t). Perhaps one could be added for this new valuemagic as well?

I didn't know we had such tests, yeah that sounds like a good idea.

Leont:
> I didn't know we had such tests, yeah that sounds like a good idea.

Added that

ext/XS-APItest/t/magic.t
```diff
@@ -15,6 +15,13 @@ sv_magic_foo($sv, $foo);
 is mg_find_foo($sv), $foo, 'foo magic attached';
 ok !mg_find_bar($sv), '... but still no bar magic';
 
+{
+	local $sv = 'Emu';
+	sv_magic_foo($sv, $foo);
+	is mg_find_foo($sv), $foo, 'foo magic attachedi to localized value';
```
leonerd: Typo "attached"

regen/mg_vtable.pl
leonerd: `PERL_MAGIC_custom(_value)`? There's precedent in `OP_CUSTOM`

Leont: Not including 'value' definitely would be a downgrade to me, but TBH I'd prefer to keep it the way it is unless someone else also disagrees with the current name.

Leont:
I've renamed it to extvalue

leonerd (approved):
LGTM lets ship this puppy

PR #20019, opened 2022-07-30
regex engine - rework node sizing logic and related functionality

demerphq:
Currently there is a lot of brittle and opaque logic related to regex regnode (aka regop) sizing. 

* many objects share common node structures structures (eg, multiple different regnodes use `struct regnode_1`) and related macros.
*  there is a very confusing concept called `NEXTOPER(p)` heavily used in the regcomp.c (and somewhat used in regexec.c), this function looks like it does something like "find the next node in the compiled program which follows p", but in fact resolves to `p+1`, and is often used like this: `NEXTOPER(NEXTOPER(p))` to compute `p+2` for cases where the code assumes the regnode is of size 2. 
*  there is another concept called `regnext` which is a different kind of "next operator" for a node. The similarity in names with `NEXTOPER()` make it quite confusing what the difference is.
*  We have a utility array called `PL_regarglen` which can be used to look up how big a given opcode should be, but it does not handle all node types, only those of a constant size.
* There are other macros which can be used to find how big a variable length regnode is, but there is nothing programmatic that knows how to handle all regnode types.
* There is no programmatic way to know if a given regnode is variable length, it is only implied by the code. 

All of these factors make it difficult and error prone to adjust the sizes of regnodes in the compiled regex program and add new functionality, and understand the code.

This patch sequence attempts to resolve these issues: 

* Adds a new function `Perl_regnode_after()` which can be used to reliably find the regnode after a given regnode regardless of its type. 
* Replaces `NEXTOPER()` with `REGNODE_AFTER()` (and a set of specialty variants of `REGNODE_AFTER()`) which should distinguish it from `regnext()`.
* Adds logic to validate the special case variants of `REGNODE_AFTER()` which are used for performance reasons (no need to do run time array looks when we know the result will be zero at compile time).
* Adds a new per-nodetype typedef so that code that it is easy to distinguish between nodes types in the code. For instance if the code is operating on a structure used by the `ANYOF` operator then it can use the typedef `tregnode_ANYOF` instead of `struct charclass` which is shared with multiple other regnodes so that when the given implementation of a node type changes it is easy to find the relevant logic.
* Rename all the `PL_reg` vars which track meta-data about the regnode structures so that they start `PL_regnode_` and add underbars to make them more readable.
* Adds  a new regnode metadata variable `PL_regnode_arg_len_varies[]` which can be used to identify node types whose length is reflected by their internal state. 
* Moves `Perl_regnext()` into reginline.h along with `Perl_regnode_after()` and `Perl_check_regnode()`. The idea being that reginline.h contains inline functions which are strictly used by the regex engine only.

Various sundry changes to facilitate the above changes. 

With these patches it is relatively easy to change the definition of a given regnode and to be able to find the code related to it, and detect what code has broken because it makes unwarranted assumptions about regnode sizes or types. Before this PR when sizes went wrong a bunch of pointer math would end up producing the wrong answer and mayhem would ensue, but nary an assert would fire to show where things went wrong. With it the issues are obvious.  -- This PR *doesnt* include any regnode sizing changes, but a follow up patch does, and these comments are based on experience of resizing the BRANCH nodetype, which is one of the weirder edge cases, and this patch was essential to allowing that to be possible.

The changes in this PR inter mesh with each other. For instance, adding typedefs means we can provide REGNODE_AFTER_type(p,type) which essentially turns into p+=sizeof(type) for the cases where we know we are working with a single well known regnode type.


jkeenan:
I see several branches with similar names.  How does yours compare to the other branches listed here:
```
$ git branch -a | grep regnode_typedef
  remotes/bram-perl/yves/regnode_typedefs
  remotes/origin/yves/regnode_typedefs
  remotes/pengvado/yves/regnode_typedefs
```

pod/perlreguts.pod
```diff
@@ -216,39 +216,113 @@ types.
 
 =head3 What regop is next?
 
-There are three distinct concepts of "next" in the regex engine, and
-it is important to keep them clear.
+There are two distinct concepts of "next regnode" in the regex engine,
+and it is important to keep them distinct in your thinking as they
+overlap conceptually in many places, but where they dont overlap the
```
jkeenan: missing an apostrophe; s/dont/don't/

pod/perlreguts.pod
```diff
@@ -216,39 +216,113 @@ types.
...
+current one has successfully matched, but in some cases this may not be
+true. In loop control and branch control regnode types the regnext may
+signify something special, for BRANCH nodes C<regnext> is the
+next BRANCH that should be executed if the current one fails execution,
+and some loop control regnodes set the regnext to be the end of the loop
+so they can jump to their cleanup if the current iteration fails to match.
 
 =back
 
+Most regnode types do not create a branch in the exection flow, and
```
jkeenan: typo: should be `exception`

regcomp.h
```diff
@@ -438,11 +438,68 @@ struct regnode_ssc {
...
 #define NODE_STEP_REGNODE	1	/* sizeof(regnode)/sizeof(regnode) */
 #define EXTRA_STEP_2ARGS	EXTRA_SIZE(struct regnode_2)
 
-#define	NEXTOPER(p)	((p) + NODE_STEP_REGNODE)
-#define	PREVOPER(p)	((p) - NODE_STEP_REGNODE)
+/* core macros for computing the regnode after this one, there is also
+ * a function which is more powerful called Perl_regnode_after(), see
+ * regcomp.c.
+ *
+ * Note that regnodes can be thought of as nodes in binary tree like structure.
```
jkeenan: Recommend a hyphen between `tree` and `like`:  `tree-like`

jkeenan (changes requested):

regcomp.h
```diff
@@ -438,11 +438,68 @@ struct regnode_ssc {
...
+ * The first "child" of a node is its immediate successor in the program, the
+ * second "child" of a node, if it has one, will be stored as an index in the
+ * regnode struct, usually in a field marked next_off (but sometimes somewhere
+ * else).  To access the immediate successor of a reegnode you should use one
```
khwilliamson: typo: regnode

regcomp.h
```diff
@@ -438,11 +438,68 @@ struct regnode_ssc {
...
+ * So for instance if you know you are dealing with a known node type
+ * then you should use REGNODE_AFTER_type(NODE,TYPE).
+ *
+ * If you have a regnode pointer and nothing else use: REGNODE_AFTER_dynamic(NODE)
+ *
+ * If you have a regnode pointer and you have already extracted its opcode use:
+ * REGNODE_AFTER_opcode(NODE,OPCODE).
+ *
+ */
+#define	_REGNODE_AFTER_plus(p,extra)    ((p) + NODE_STEP_REGNODE + (extra))
```
khwilliamson: We probably shouldn't be adding non-conforming symbol names.  `REGNODE_AFTER_PLUS_` should do

regcomp.h
```diff
@@ -438,11 +438,68 @@ struct regnode_ssc {
...
+/* under DEBUGGING we check that all REGNODE_AFTER style macro did the
```
khwilliamson: *macros

khwilliamson (approved):
Thank you for this.  

I only found trivial issues.

demerphq:
On Sat, 30 Jul 2022 at 21:29, James E Keenan ***@***.***>
wrote:

> I see several branches with similar names. How does yours compare to the
> other branches listed here:
>
> $ git branch -a | grep regnode_typedef
>   remotes/bram-perl/yves/regnode_typedefs
>   remotes/origin/yves/regnode_typedefs
>   remotes/pengvado/yves/regnode_typedefs
>
> Those are other peoples copies of the branch which they have pulled into
their repo, and then you have pulled from them into yours. Likely theirs is
old. The canonical branch is origin/yves/regnode_typesdef.

Yves
-- 

pod/perlreguts.pod
demerphq: execution actually. :-)

demerphq:
Hi, I integrated and rebased the requested changes. I also added a new comment with more and better docs. No code changes other than requested in the rebase. Please review the very last commit in the PR.

demerphq:
A follow up question from this PR for me is whether we should unify the regnode information data into a structure and replace all the PL_regnode_... arrays with something unified. We could even use packed integers and reduce the memory footprint involved, not sure if that would be a win or not.

demerphq:
@jkeenan and @khwilliamson need approvals from you guys again.

khwilliamson (approved):

jkeenan (approved):

PR #20021, opened 2022-07-30
Fix typo in CvMETHOD_off definition

Leont:
The definition uses the wrong name for the argument (off instead of cv), and therefore causes compilation failures.


leonerd (approved):
Oops, my bad.
LGTM

PR #20022, opened 2022-07-30
newSvsv_flags - support SVs_TEMP flag, then make use of it

richardleach:
Prior to this commit, a new mortal SV couldn't be created in one
call, instead the following would be necessary:
`    sv_2mortal(newSVsv(...))`

Between the two function calls and unnecessary checks performed by
`sv_2mortal`, it is more efficient for `newSVsv_flags` to be able to
directly add the SV it has just created to the temp stack. This mirrors
the `SVs_TEMP` flag handling from `newSVpvn_flags`.

A number of instances of the old pattern have then been updated to
use the new functionality.

### DIFF
--- a/hv.c
+++ b/hv.c
@@ -692,7 +692,7 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
                     }
                     if (TAINTING_get)
                         TAINT_set(SvTAINTED(keysv));
-                    keysv = sv_2mortal(newSVsv(keysv));
+                    keysv = newSVsv_flags(keysv, SV_GMAGIC|SV_NOSTEAL|SVs_TEMP);
                     mg_copy(MUTABLE_SV(hv), val, (char*)keysv, HEf_SVKEY);
                 } else {
                     mg_copy(MUTABLE_SV(hv), val, key, klen);
--- a/mro_core.c
+++ b/mro_core.c
@@ -1190,7 +1190,7 @@ S_mro_gather_and_rename(pTHX_ HV * const stashes, HV * const seen_stashes,
                             }
                         }
                         else {
-                            subname = sv_2mortal(newSVsv(namesv));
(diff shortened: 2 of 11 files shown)


leonerd:
I'm unclear how this differs from `sv_mortalcopy(sv)`. Perhaps the two could be combined somehow?

richardleach:
Aha, excellent point! It turns out that `Perl_newSVsv_flags` (with this change) and `Perl_sv_mortalcopy_flags` would be almost identical. The only difference is that the former checks for `(!old)` and `(SvTYPE(old) == (svtype)SVTYPEMASK)` but the latter doesn't. 

There's also some flag differences when it comes to calling macros:
```
#define newSVsv(sv) newSVsv_flags((sv), SV_GMAGIC|SV_NOSTEAL)
```
vs
```
#define sv_mortalcopy(sv) \
        Perl_sv_mortalcopy_flags(aTHX_ sv, SV_GMAGIC|SV_DO_COW_SVSETSV)
```
So, it looks like there'd be no need to add `SVs_TEMP` support to `Perl_newSVsv_flags` in this PR, since the instances of `sv_2mortal(newSVsv(...))`  could be changed to use `sv_mortalcopy_flags` ... _as long as_ the old SV* will always be valid and not point to a freed string. 

Any thoughts on whether those conditions might be reliably met?

richardleach:
`Perl_sv_setsv_flags`, called by both functions, does it's own checking for null or freed sv's. That brings up a seemingly-unfortunate difference in behaviour, making it harder to know what's the correct route for this PR.

If the old SV is `NULL`:
 * `Perl_newSVsv_flags` will return `NULL`
 * `Perl_sv_mortalcopy_flags` will (via `sv_setsv_flags`) return `&PL_sv_undef`
  
If the old SV has been freed:
 * `Perl_newSVsv_flags` will warn and return `NULL`
* `Perl_sv_mortalcopy_flags` will (via `sv_setsv_flags`) croak.

richardleach:
Not sure the effort of trying to square `Perl_newSVsv_flags` and `Perl_sv_mortalcopy_flags` is worth it. Closing this PR.

PR #20023, opened 2022-08-01
pod/perlsub.pod -  Make three links working after pod2html conversion

HaraldJoerg:
Severity: minor

A space after the separator between text and name as in L<< text | name >>
is retained in the href element created by pod2html: href="base/ name".
Three of these occur in perlsub.pod, the links are broken in the
perlsub presentations on metacpan, perldoc.pl and github.

This patch removes a set of such spaces.

I guess this could also (or instead) be fixed in pod2html and maybe
other POD converters by making them eliminate leading spaces after
the separator?

### DIFF
--- a/pod/perlsub.pod
+++ b/pod/perlsub.pod
@@ -294,8 +294,8 @@ C<SETLINEBUF>, C<SYSOPEN>, C<TELL>, C<UNREAD>, C<UTF8>, C<WRITE>
 
 =item documented in L<perlfunc>
 
-L<< C<import> | perlfunc/use >>, L<< C<unimport> | perlfunc/use >>,
-L<< C<INC> | perlfunc/require >>
+L<< C<import>|perlfunc/use >>, L<< C<unimport>|perlfunc/use >>,
+L<< C<INC>|perlfunc/require >>
 
 =item documented in L<UNIVERSAL>
 


haarg (approved):

jkeenan:
It's not yet clear to me that the changes in the source code will, in and of themselves, have the effect which @HaraldJoerg is aiming for.

First, one administrative comment.  The pull request as originally submitted came from a new email address from an existing AUTHOR, so `make test_porting` failed.  I corrected this and pushed this branch to origin:  `jkeenan/HaraldJoerg/pod-fixes`.

Second, when assessing claims of incorrect POD-to-HTML conversion, it's tempting to lay the blame on the `pod2html` utility that ships with the core distribution.  However, we (Perl 5 Porters and people who have worked on the underlying Pod::Html module) don't generate or publish the HTML code which you see at metacpan.org, perldoc.pl or github.  We don't know where in the HTML-generation process those sites use `pod2html`; they might not use that utility at all (though that's unlikely)!

All that we can do is (i) generate an HTML text file by running `pod2html` over a source file in blead; (ii) generate a second HTML text file by running `pod2html` over that same source file in the pull request; (iii) `diff`-ing the two generated files; and (iv) making a judgment call as to whether applying the patch will result in "better looking" HTML once various downstream users use these revised inputs.

Doing that with the revisions in this p.r. gives the following `diff`:
```
$ diff perlsub.blead.html perlsub.branch.html 
216c216
< <p><a><code>import</code> </a>, <a><code>unimport</code> </a>, <a><code>INC</code> </a></p>
---
> <p><a><code>import</code></a>, <a><code>unimport</code></a>, <a><code>INC</code></a></p>
```
The patch eliminates the wordspaces between instances of `</code>` and `</a>`.  Does this seem reasonable?  Yes.  Can we absolutely guarantee that, once the downstream web pages are updated, the links will be correct?  No.

Part of the problem is that this line of code in `pod/perlsub.pod` is the only line in that file that has that particular pattern of POD-formatting codes.

I think we should apply the porting-test-corrected version of this p.r., but we should bear in mind that it might not fully solve the problem.

Thank you very much.
Jim Keenan

haarg:
The output you posted does not include the `href` links on the `<a>` tags. If you include the `--podpath pod` option, it should try to generate the `href` links, which will demonstrate the problem. Without this fix, it will still not include the `href` option because it can't find the file with a space in its name.

The difference can also be demonstrated using `perldoc -o html pod/perlsub.pod`. Without this fix, it will generate links like `<a href="https://metacpan.org/pod/%20perlfunc#use"`, which has an extraneous `%20`.

jkeenan:
> The output you posted does not include the `href` links on the `<a>` tags. If you include the `--podpath pod` option, it should try to generate the `href` links, which will demonstrate the problem. Without this fix, it will still not include the `href` option because it can't find the file with a space in its name.
> 
> The difference can also be demonstrated using `perldoc -o html pod/perlsub.pod`. Without this fix, it will generate links like `<a href="https://metacpan.org/pod/%20perlfunc#use"`, which has an extraneous `%20`.

I tried the approach you suggested in the second paragraph.  This gives good results, so I'll merge the (porting-test-corrected) branch.  Thanks.

HaraldJoerg:
Thanks @jkeenan for fixing checkAuthors.pl!  I was not aware of that test (and also had forgotten that I'm mentioned in AUTHORS so it didn't occur to me to verify the mail address).

jkeenan:
Superseded by https://github.com/Perl/perl5/pull/20025, which has been merged into blead.  Thanks for your contribution; closing this p.r.

Grinnz:
> Second, when assessing claims of incorrect POD-to-HTML conversion, it's tempting to lay the blame on the `pod2html` utility that ships with the core distribution. However, we (Perl 5 Porters and people who have worked on the underlying Pod::Html module) don't generate or publish the HTML code which you see at metacpan.org, perldoc.pl or github. We don't know where in the HTML-generation process those sites use `pod2html`; they might not use that utility at all (though that's unlikely)!

For posterity, none of those sites use pod2html or Pod::Html. They use Pod::Simple directly.

PR #20028, opened 2022-08-01
Perl_reg_named_buff_fetch - simpler retarray creation & push

richardleach:
`retarray` is a straightforward array created and populated with
`SvIVX(sv_dat)` number of elements within this function. As the
number of elements is known and all elements will be used, a
correctly-sized array can be created using `newAV_alloc_x`.

Since it's a straightforward array, `av_push_simple` can be used in
place of `av_push`. This will be more efficient.


demerphq:
Thanks. There is lots of that in the regex engine. Maybe you should make a tutorial list on what the new functions look like and we can do some mass conversions.

demerphq (approved):

PR #20030, opened 2022-08-02
Configure should avoid looping infinitely repeating the same question

nwc10:
Configure's helper function ./myread is intended to loop until it gets an
acceptable answer. For a couple of cases, an empty string is not acceptable
(eg 'Where is your C library?', if all automated attempts at answering this
have failed). In these cases, if Configure's standard input is /dev/null (or
closed), the shell read returns an empty string, and ./myread repeats the
question.

Before this commit, it would loop infinitely (generating continuous terminal
output). With this commit, we add a retry counter - if it asks the same
question to many times, it aborts Configure. This means that unattended
./Configure runs should now always terminate, as termination with an error is
better than spinning forever.

### DIFF
--- a/Configure
+++ b/Configure
@@ -2215,6 +2215,7 @@ $startsh
 xxxm=\$dflt
 $myecho
 ans='!'
+counter=42
 case "\$fastread" in
 yes) case "\$dflt" in
 	'') ;;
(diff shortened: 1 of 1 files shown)


nwc10:
Applies cleanly all the way back to perl-5.000.
(Strictly actually perl 5.0 alpha 8, which first added Configure)

Tux (approved):

Tux:
Approved. Go ahead, I'll merge it to meta afterwards

Tux:
Backported to meta

PR #20031, opened 2022-08-02
Create and use a `refcount_is()` unit-test helper function

leonerd:
Extracts some common behaviour from unit tests into one standard place.

The use of `Internals::SvREFCNT` isn't great here, but at least it's now in fewer places where we can fix them all up later.

### DIFF
--- a/t/op/eval.t
+++ b/t/op/eval.t
@@ -514,7 +514,7 @@ END_EVAL_TEST
     my $t;
     my $s = "a";
     $s =~ s/a/$t = \%^H;  qq( qq() );/ee;
-    is(Internals::SvREFCNT(%$t), $count_expected, 'RT 63110');
+    refcount_is $t, $count_expected, 'RT 63110';
 }
 
 # make sure default arg eval only adds a hints hash once to entereval
@@ -531,9 +531,9 @@ END_EVAL_TEST
     # test that the CV compiled for the eval is freed by checking that no additional 
     # reference to outside lexicals are made.
     my $x;
-    is(Internals::SvREFCNT($x), 1, "originally only 1 reference");
+    refcount_is \$x, 1+1, "originally only 1 reference"; # + 1 to account for the ref here
     eval '$x';
-    is(Internals::SvREFCNT($x), 1, "execution eval doesn't create new references");
+    refcount_is \$x, 1+1, "execution eval doesn't create new references"; # + 1 the same
 }
 
 fresh_perl_is(<<'EOP', "ok\n", undef, 'RT #70862');
--- a/t/op/inccode.t
+++ b/t/op/inccode.t
@@ -247,16 +247,17 @@ shift @INC;
     my $data = [];
     unshift @INC, sub { $die, $data };
 
-    my $initial_sub_refcnt = &Internals::SvREFCNT($die);
-    my $initial_data_refcnt = &Internals::SvREFCNT($data);
+    # + 1 to account for prototype-defeating &... calling convention
+    my $initial_sub_refcnt = &Internals::SvREFCNT($die) + 1;
+    my $initial_data_refcnt = &Internals::SvREFCNT($data) + 1;
 
     do "foo";
-    is(&Internals::SvREFCNT($die), $initial_sub_refcnt, "no leaks");
-    is(&Internals::SvREFCNT($data), $initial_data_refcnt, "no leaks");
+    refcount_is $die, $initial_sub_refcnt, "no leaks";
+    refcount_is $data, $initial_data_refcnt, "no leaks";
 
     do "bar";
-    is(&Internals::SvREFCNT($die), $initial_sub_refcnt, "no leaks");
-    is(&Internals::SvREFCNT($data), $initial_data_refcnt, "no leaks");
+    refcount_is $die, $initial_sub_refcnt, "no leaks";
+    refcount_is $data, $initial_data_refcnt, "no leaks";
 
     shift @INC;
 }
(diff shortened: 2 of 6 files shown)


t/test.pl
```diff
@@ -476,6 +476,16 @@ ($$$@)
     _ok($pass, _where(), $name, @mess);
 }
 
+sub refcount_is {
+    my (undef, $expected, $name, @mess) = @_;
+    my $got = &Internals::SvREFCNT($_[0]) + 1; # +1 to account for the & calling style
+    my $pass = $got == $expected;
+    unless ($pass) {
+        unshift @mess, "#      got $got references, expected $expected\n";
```
bram-perl: The other messages (from `is`, `isnt`, `cmp_ok`, ...) all appear to be multiline, i.e.
```
not ok 1 - foobar
# Failed test 1 - foobar at bla.t line 9
#      got "foo"
# expected "bar"
1..1
```
This message has the 'got' and 'expected' on the same line which is a bit inconsistent.

t/op/eval.t
```diff
@@ -531,9 +531,9 @@ END_EVAL_TEST
     # test that the CV compiled for the eval is freed by checking that no additional 
     # reference to outside lexicals are made.
     my $x;
-    is(Internals::SvREFCNT($x), 1, "originally only 1 reference");
+    refcount_is \$x, 1+1, "originally only 1 reference";
```
bram-perl: It's ok but it's a bit misleading: the message says "originally only 1 reference" and the check is for 2 references.
(I understand why; it's just a bit amusing/misleading :) )

t/op/inccode.t
```diff
@@ -247,16 +247,17 @@ shift @INC;
...
-    my $initial_data_refcnt = &Internals::SvREFCNT($data);
+    # + 1 to account for prototype-defeating &... calling convention
+    my $initial_sub_refcnt = &Internals::SvREFCNT($die) + 1;
+    my $initial_data_refcnt = &Internals::SvREFCNT($data) + 1;
 
     do "foo";
-    is(&Internals::SvREFCNT($die), $initial_sub_refcnt, "no leaks");
-    is(&Internals::SvREFCNT($data), $initial_data_refcnt, "no leaks");
+    refcount_is $die, $initial_sub_refcnt, "no leaks";
+    refcount_is $data, $initial_data_refcnt, "no leaks";
```
bram-perl, line 256: Hmm... not saying it's wrong but can you double check this?

In t/op/eval.t you replaced:
```
    is(Internals::SvREFCNT($x), 1, "originally only 1 reference");
```
with:
```
    refcount_is \$x, 1+1, "originally only 1 reference";
```
i.e. argument changed from `$x` to `\$x`.

In this code however you replaced
```
    is(&Internals::SvREFCNT($die)...
```
with:
```
    refcount_is $die
```
i.e. argument remained unchanged.

Without further testing I would expect one of them to be wrong..

t/test.pl
leonerd: Ahyes, will fix.

t/op/eval.t
leonerd: I'll put a trailing comment to explain it.

t/op/inccode.t
leonerd: Yes. So in the latter case, the original `Internals::SvREFCNT` function was being called by prototype-defeating ampersand notation, `&FUNC...` and therefore the arguments should be unchanged.

In the former case, the SvREFCNT has an annoying `\[$@%]` prototype, which causes the actual XS function to see a new reference to the variable which was named at the callsite. As the new `refcount_is` function does not have that prototype, we must create the reference manually with the `\` operator.

leonerd:
Comments addressed and force-pushed; PTAL

bram-perl (approved):

t/test.pl
```diff
@@ -476,6 +476,17 @@ ($$$@)
     _ok($pass, _where(), $name, @mess);
 }
 
+sub refcount_is {
+    my (undef, $expected, $name, @mess) = @_;
```
jkeenan, line 482: IIRC our policy is that `t/test.pl` is strictly for the use of the core distribution and is therefore **not** documented with POD.  But perhaps some inline comments would be helpful here.  In particular, could you explain why the first argument to a call of this function is set to `undef`?

leonerd: Ahyes; that's fair. I'll pop something in before I merge it.

jkeenan:
Can this new function be used in these tests as well?
```
t/op/gv.t

1027 package HTTP::MobileAttribute::Plugin::Locator {
1028     use constant LOCATOR_GPS => 1;
1029     ::ok defined &{__PACKAGE__."::LOCATOR_GPS"},
1030         'defined &{"name of constant"}';
1031     ::ok Internals::SvREFCNT(${__PACKAGE__."::"}{LOCATOR_GPS}),
1032        "stash elem for slot is not freed prematurely";
1033 }

t/op/ref.t

733     fresh_perl_is(<<'EOF', 'ok', { stderr => 1 }, 'array with 1 weak ref');
734 no warnings 'experimental::builtin';
735 use builtin qw(weaken);
736 my $r = [];
737 Internals::SvREFCNT(@$r, 9);
738 my $r1 = $r;
739 weaken($r1);
740 print "ok";
741 EOF

+ 3 similar instances in same file
```

leonerd:
No, I looked around at all the tests and updated the easy ones. The rest of them are doing *weird* things with SvREFCNT. Sometimes it is even used as a mutator to *change* the refcount of SVs.

PR #20032, opened 2022-08-02
Perl_newRV_noinc - explicitly simplify, convert to inline func

richardleach:
`Perl_newRV_noinc` creates a new `SVt_IV`, then calls `sv_setrv_noinc`,
which will check if the `SVt_IV` is `SvTHINKFIRST` (it won't be) and
if it is types other than `SVt_IV` (it won't be). If those checks
and associated branches are removed, all that remains is some flag
twiddling and setting the `sv_u.svu_rv` pointer.

A decent compiler *might* figure that all out and simplify
`Perl_newRV_noinc` right down to the essentials, but if we do that
directly, the entire function is small enough to move to _sv_inline.h_.

### DIFF
--- a/sv_inline.h
+++ b/sv_inline.h
@@ -929,5 +929,30 @@ Perl_SvPV_helper(pTHX_
 }
 
 /*
+=for apidoc newRV_noinc
+
+Creates an RV wrapper for an SV.  The reference count for the original
+SV is B<not> incremented.
+
+=cut
+*/
+
+PERL_STATIC_INLINE SV *
+Perl_newRV_noinc(pTHX_ SV *const tmpRef)
+{
+    SV *sv = newSV_type(SVt_IV);
+
+    PERL_ARGS_ASSERT_NEWRV_NOINC;
+
+    SvTEMP_off(tmpRef);
+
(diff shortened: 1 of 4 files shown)


sv_inline.h
```diff
@@ -928,6 +928,34 @@ Perl_SvPV_helper(pTHX_
...
+Perl_newRV_noinc(pTHX_ SV *const tmpRef)
+{
+    SV *sv = newSV_type(SVt_IV);
+
+    PERL_ARGS_ASSERT_NEWRV_NOINC;
+
+    SvTEMP_off(tmpRef);
+
+    /* inlined, simplified sv_setrv_noinc(sv, tmpRef); */
+    assert(!SvTHINKFIRST(sv));
```
Leont: I don't understand why we would keep these asserts here.

richardleach: I was in two minds about not having them. Left them in as sort of a visual aid to the reader to show why the simplification was possible, but granted it's not that hard to figure out. Am happy to remove them.

richardleach: I've removed the asserts.

Leont (approved):

PR #20034, opened 2022-08-03
regex engine - replace many attribute arrays with one     

demerphq:
This replaces `PL_regnode_arg_len`, `PL_regnode_arg_len_varies`,
`PL_regnode_off_by_arg` and `PL_regnode_kind` with a single `PL_regnode_info`
array, which is an array of `struct regnode_meta`, which contains the same
data but as a struct. Since `PL_regnode_name` is only used in debugging
builds of the regex engine we keep it separate. If we add more debug
properties it might be good to create a `PL_regnode_debug_info[]` to hold
that data instead.

This means when we add new properties we do not need to modify any
secondary sources to add new properites, just the struct definition
and `regen/regcomp.pl`

### DIFF
--- a/globvar.sym
+++ b/globvar.sym
@@ -57,11 +57,8 @@ PL_phase_names
 PL_ppaddr
 PL_reg_extflags_name
 PL_reg_intflags_name
-PL_regnode_arg_len
-PL_regnode_arg_len_varies
-PL_regnode_kind
+PL_regnode_info
 PL_regnode_name
-PL_regnode_off_by_arg
 PL_revision
 PL_runops_dbg
 PL_runops_std
--- a/regcomp.c
+++ b/regcomp.c
@@ -2950,7 +2950,7 @@ S_make_trie(pTHX_ RExC_state_t *pRExC_state, regnode *startbranch,
         case EXACTFU:
         case EXACTFLU8: folder = PL_fold_latin1; break;
         case EXACTF:  folder = PL_fold; break;
-        default: Perl_croak( aTHX_ "panic! In trie construction, unknown node type %u %s", (unsigned) flags, PL_regnode_name[flags] );
+        default: Perl_croak( aTHX_ "panic! In trie construction, unknown node type %u %s", (unsigned) flags, REGNODE_NAME(flags) );
     }
 
     trie = (reg_trie_data *) PerlMemShared_calloc( 1, sizeof(reg_trie_data) );
@@ -5168,7 +5168,7 @@ S_study_chunk(pTHX_
                                 }
                                 Perl_re_printf( aTHX_  "(First==%d,Last==%d,Cur==%d,tt==%s,ntt==%s,nntt==%s)\n",
                                    REG_NODE_NUM(first), REG_NODE_NUM(prev), REG_NODE_NUM(cur),
-                                   PL_regnode_name[trietype], PL_regnode_name[noper_trietype], PL_regnode_name[noper_next_trietype]
+                                   REGNODE_NAME(trietype), REGNODE_NAME(noper_trietype), REGNODE_NAME(noper_next_trietype)
                                 );
                             });
 
@@ -5266,7 +5266,7 @@ S_study_chunk(pTHX_
                               depth+1, SvPV_nolen_const( RExC_mysv ), REG_NODE_NUM(cur));
                             Perl_re_printf( aTHX_  "(First==%d, Last==%d, Cur==%d, tt==%s)\n",
                                REG_NODE_NUM(first), REG_NODE_NUM(prev), REG_NODE_NUM(cur),
-                               PL_regnode_name[trietype]
+                               REGNODE_NAME(trietype)
                             );
 
                         });
@@ -6788,7 +6788,7 @@ S_study_chunk(pTHX_
 
         else if (OP(scan) == REGEX_SET) {
             Perl_croak(aTHX_ "panic: %s regnode should be resolved"
-                             " before optimization", PL_regnode_name[REGEX_SET]);
+                             " before optimization", REGNODE_NAME(REGEX_SET));
         }
 
         /* Else: zero-length, ignore. */
@@ -21390,7 +21390,7 @@ S_reginsert(pTHX_ RExC_state_t *pRExC_state, const U8 op,
     PERL_UNUSED_CONTEXT;
     PERL_UNUSED_ARG(depth);
 /* (REGNODE_TYPE((U8)op) == CURLY ? EXTRA_STEP_2ARGS : 0); */
-    DEBUG_PARSE_FMT("inst"," - %s", PL_regnode_name[op]);
+    DEBUG_PARSE_FMT("inst"," - %s", REGNODE_NAME(op));
     assert(!RExC_study_started); /* I believe we should never use reginsert once we have started
                                     studying. If this is wrong then we need to adjust RExC_recurse
                                     below like we do with RExC_open_parens/RExC_close_parens. */
@@ -21474,7 +21474,7 @@ S_regtail(pTHX_ RExC_state_t * pRExC_state,
             Perl_re_printf( aTHX_  "~ %s (%zu) %s %s\n",
                 SvPV_nolen_const(RExC_mysv), scan,
                     (temp == NULL ? "->" : ""),
-                    (temp == NULL ? PL_regnode_name[OP(REGNODE_p(val))] : "")
+                    (temp == NULL ? REGNODE_NAME(OP(REGNODE_p(val))) : "")
             );
         });
         if (temp == NULL)
@@ -21567,7 +21567,7 @@ S_regtail_study(pTHX_ RExC_state_t *pRExC_state, regnode_offset p,
             Perl_re_printf( aTHX_  "~ %s (%zu) -> %s\n",
                 SvPV_nolen_const(RExC_mysv),
                 scan,
-                PL_regnode_name[exact]);
+                REGNODE_NAME(exact));
         });
         if (temp == NULL)
             break;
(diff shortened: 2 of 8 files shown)


dur-randir:
blead
``` 
 1,873,148,624,472      cycles                                                        (74.97%)
 2,900,892,460,205      instructions              #    1.55  insn per cycle           (75.05%)
    22,855,830,376      cache-misses                                                  (75.04%)
     9,753,618,600      branch-misses                                                 (74.94%)

      64.483203280 seconds time elapsed
````

origin/pull/20034
```
 1,864,054,984,220      cycles                                                        (74.96%)
 2,895,803,310,996      instructions              #    1.55  insn per cycle           (74.79%)
    23,121,151,012      cache-misses                                                  (75.08%)
     9,736,196,892      branch-misses                                                 (75.17%)

      65.270244613 seconds time elapsed
```

This is for the whole test suite, results are roughly the same.

**Comment by jkeenan** [comment]
> blead
> 
> ```
>  1,873,148,624,472      cycles                                                        (74.97%)
>  2,900,892,460,205      instructions              #    1.55  insn per cycle           (75.05%)
>     22,855,830,376      cache-misses                                                  (75.04%)
>      9,753,618,600      branch-misses                                                 (74.94%)
> 
>       64.483203280 seconds time elapsed
> ```
> 
> origin/pull/20034
> 
> ```
>  1,864,054,984,220      cycles                                                        (74.96%)
>  2,895,803,310,996      instructions              #    1.55  insn per cycle           (74.79%)
>     23,121,151,012      cache-misses                                                  (75.08%)
>      9,736,196,892      branch-misses                                                 (75.17%)
> 
>       65.270244613 seconds time elapsed
> ```
> 
> This is for the whole test suite, results are roughly the same.

For the non-specialists among us ... can you explain how this data is derived and what its significance is?

**Comment by jkeenan** [comment]
> This replaces `PL_regnode_arg_len`, `PL_regnode_arg_len_varies`, `PL_regnode_off_by_arg` and `PL_regnode_kind` with a single `PL_regnode_info` array, which is an array of `struct regnode_meta`, which contains the same data but as a struct. Since `PL_regnode_name` is only used in debugging builds of the regex engine we keep it separate. If we add more debug properties it might be good to create a `PL_regnode_debug_info[]` to hold that data instead.
> 

Are any of the functions you're replacing likely to have been touched in XS code on CPAN?  What would we `grep` for to anticipate possible breakage?

(Also, all tests PASS in a typical build on FreeBSD.)

**Comment by dur-randir** [comment]
> can you explain how this data is derived 

This is the output from `TEST_JOBS=30 perf stat -e cycles,instructions,cache-misses,branch-misses make test_harness` It uses Linux profiler 'perf' to record 4 hardware-level counters from all processes in the test suite. 

> what its significance is

'Cycles' is the total code runtime, more precise than the 'time elapsed', as it's not affected by frequency scaling. Difference in 'instructions' and 'branch-misses' shows whether any significantly different code paths are taken. And 'cache-misses' is to check the hypothesis that this change won't hit any obscure corners that could lead to a slowdown (as a cache miss is very costly).

**Comment by demerphq** [comment]
On Wed, 3 Aug 2022 at 21:22, James E Keenan ***@***.***>
wrote:

> This replaces PL_regnode_arg_len, PL_regnode_arg_len_varies,
> PL_regnode_off_by_arg and PL_regnode_kind with a single PL_regnode_info
> array, which is an array of struct regnode_meta, which contains the same
> data but as a struct. Since PL_regnode_name is only used in debugging
> builds of the regex engine we keep it separate. If we add more debug
> properties it might be good to create a PL_regnode_debug_info[] to hold
> that data instead.
>
> Are any of the functions you're replacing likely to have been touched in
> XS code on CPAN? What would we grep for to anticipate possible breakage?
>

They aren't functions but global variables.  PL_regnode_arg_len[] etc. And
they contain data useful only to the regex engine about how its opcodes
function, it is hard to imagine why someone would use them in the wild.

cheers,
Yves

-- 

**Comment by demerphq** [comment]
On Wed, 3 Aug 2022 at 21:57, Sergey Aleynikov ***@***.***>
wrote:

> can you explain how this data is derived
>
> This is the output from TEST_JOBS=30 perf stat -e
> cycles,instructions,cache-misses,branch-misses make test_harness It uses
> Linux profiler 'perf' to record 4 hardware-level counters from all
> processes in the test suite.
>
> what its significance is
>
> 'Cycles' is the total code runtime, more precise than the 'time elapsed',
> as it's not affected by frequency scaling. Difference in 'instructions' and
> 'branch-misses' shows whether any significantly different code paths are
> taken. And 'cache-misses' is to check the hypothesis that this change won't
> hit any obscure corners that could lead to a slowdown (as a cache miss is
> very costly).
>

When you do this do you set PERL_HASH_SEED to a consistent mode? Eg via
PERL_HASH_SEED=0 or equivalent recipes (0 is special it implies
PERL_PERTURB_KEYS=0)?

That I believe would make your tests results more comparable. Perhaps not
100% accurate compared to random mode. But without the above there will be
a certain level of random noise in the kind of numbers you are talking
about, as hashes as split at different times (we only split on collision
these days), and as strings get hashed into different orders, etc.

Even with this I believe there are various tests that use rand() in our
test codebase that do not faciliate an external SRAND() call.

I wonder if we should add an env var so we can SRAND from the env?  We
might have to deal with the rand() special case under forking: if you
rand()/srand() before fork the childrens srand state will be consistent, if
you do not then each child will get their own state when they call
rand()/srand() the first time.

It would be nice to have a "deterministic test mode" where we could
guarantee that two runs of the test suite would do exactly the same thing.

cheers,
Yves

-- 

**Comment by dur-randir** [comment]
>When you do this do you set PERL_HASH_SEED to a consistent mode? Eg via
>PERL_HASH_SEED=0 or equivalent recipes (0 is special it implies
>PERL_PERTURB_KEYS=0)?

No, this was more of a quick and rough estimate. Running only regex-related tests or just a number of specific tests that hit those arrays would be even better, but I don't know this area of code so I don't know how to select them. I've repeated some runs with PERL_HASH_SEED=0, but it still gives different instructions count, so it doesn't eliminate all noise.

>It would be nice to have a "deterministic test mode" where we could
>guarantee that two runs of the test suite would do exactly the same thing.

It'd especially be great for fuzzing, as to achieve this I use lots of LD_PRELOAD redefined functions, but it's really cumbersome, so I wasn't using this technique here.

PS: op/each.t fails under PERL_HASH_SEED=0

**Comment by demerphq** [comment]
On Thu, 4 Aug 2022 at 10:42, Sergey Aleynikov ***@***.***>
wrote:

> When you do this do you set PERL_HASH_SEED to a consistent mode? Eg via
> PERL_HASH_SEED=0 or equivalent recipes (0 is special it implies
> PERL_PERTURB_KEYS=0)?
>
> No, this was more of a quick and rough estimate. Running only
> regex-related tests or just a number of specific tests that hit those
> arrays would be even better, but I don't know this area of code so I don't
> know how to select them.
>

make test_reonly


> I've repeated some runs with PERL_HASH_SEED=0, but it still gives
> different instructions count, so it doesn't eliminate all noise.
>

No, there will be some tests using srand(). I guess something like:

$ echo "BEGIN{srand(1234);}" > lib/ForceSrand.pm
$ export PERL5OPT="-MForceSrand"

would sort it out.


> It would be nice to have a "deterministic test mode" where we could
> guarantee that two runs of the test suite would do exactly the same thing.
>
> It'd especially be great for fuzzing, as to achieve this I use lots of
> LD_PRELOAD redefined functions, but it's really cumbersome, so I wasn't
> using this technique here.
>
> PS: op/each.t fails under PERL_HASH_SEED=0
>
Yep, I have a PR in preparation already. :-)

cheers,
Yves


-- 

**Comment by demerphq** [comment]
On Thu, 4 Aug 2022 at 11:56, demerphq ***@***.***> wrote:

>
> $ echo "BEGIN{srand(1234);}" > lib/ForceSrand.pm
>
>
$ echo "BEGIN{srand(1234);} 1;" > lib/ForceSrand.pm

 should work, but it doesn't because lib gets cleaned during make. When I
get time later ill try to put something together.

Yves

**Comment by dur-randir** [comment]
Still get ~1% noise even with the following `TEST_ARGS='re/*.t ext/re/t/*.t' TESTFILE=harness PERL5OPT="-I/home/s.aleinikov/perl-git/lib -MForceSrand" PERL_HASH_SEED=0 TEST_JOBS=1 taskset 0x30 perf stat -e cycles,cache-misses,branch-misses ./runtests choose`

pull/20034
``` 
   206,540,119,336      cycles
     2,861,646,883      cache-misses
       764,086,003      branch-misses
```

blead
```
   208,722,675,581      cycles
     2,852,989,346      cache-misses
       764,943,621      branch-misses
```

demerphq:
On Thu, 4 Aug 2022 at 14:05, Sergey Aleynikov ***@***.***>
wrote:

> Still get ~1% noise even with the following TEST_ARGS='re/*.t
> ext/re/t/*.t' TESTFILE=harness PERL5OPT="-I/home/s.aleinikov/perl-git/lib
> -MForceSrand" PERL_HASH_SEED=0 TEST_JOBS=1 taskset 0x30 perf stat -e
> cycles,cache-misses,branch-misses ./runtests choose
>

One last option: PERL_INTERNAL_RAND_SEED=123


> pull/20034
>
>    206,540,119,336      cycles
>      2,861,646,883      cache-misses
>        764,086,003      branch-misses
>
> blead
>
>    208,722,675,581      cycles
>      2,852,989,346      cache-misses
>        764,943,621      branch-misses
>
>
> So very slightly more cache misses, but the difference is less than 1%, so
less than the noise.

Yves
-- 

dur-randir:
>One last option: PERL_INTERNAL_RAND_SEED=123

It hasn't improved stability.

khwilliamson (approved):

demerphq:
On Thu, 4 Aug 2022 at 16:36, Sergey Aleynikov ***@***.***>
wrote:

> One last option: PERL_INTERNAL_RAND_SEED=123
>
> It hasn't improved stability
>

FWIW, I pushed this: https://github.com/Perl/perl5/pull/20052

Yves

-- 

demerphq:
And #20052 is now merged. PERL_RAND_SEED=1 PERL_INTERNAL_RAND_SEED=2 PERL_HASH_SEED=0 should now produce the most repeatable results from our test suite that we can provide.

PR #20035, opened 2022-08-03
Add tests for `do { ... } while 0;` and `do { ... } until 1;`

nwc10:
Astoundingly, even though we had the perl code as a comment in the C code
which handles this special case in Perl_newLOOPOP(), we didn't actually have
a regression test for it. Father C added a test case for a variant in commit
fc39925ca702f4c8 in Dec 2013, where the value for until is a constant, but
constant folding fails, but the "simple" version never actually had a test
case.

This commit adds the missing test cases.

### DIFF
--- a/t/cmd/mod.t
+++ b/t/cmd/mod.t
@@ -1,6 +1,6 @@
 #!./perl
 
-print "1..13\n";
+print "1..15\n";
 
 print "ok 1\n" if 1;
 print "not ok 1\n" unless 1;
@@ -54,3 +54,19 @@ print "not ok 12\n" if $x > 0;
 # This used to cause a segfault
 $x = "".("".do{"foo" for (1)});
 print "ok 13\n";
+
+$x = 0;
+do { ++$x } while 0;
+if ($x == 1) {
+    print "ok 14\n";
+} else {
+    print "not ok 14 # $x\n";
+}
+
+$x = 0;
+do { ++$x } until 1;
+if ($x == 1) {
+    print "ok 15\n";
+} else {
+    print "not ok 15 # $x\n";
+}


t/cmd/mod.t
```diff
@@ -1,6 +1,6 @@
 #!./perl
 
-print "1..13\n";
+print "1..15\n";
```
leonerd, line 3: Whenever I encounter one of these in a core `.t` file I replace it with `done_testing` at the end of the file.

leonerd (commented):
Test logic looks good.
Personally I'd take the opportunity to modernize the test file with the usual gubbins while we're there (require `test.pl`, use the various test helpers, etc). But up to you if you want to here - if not I won't object ;)

t/cmd/mod.t
nwc10: That would be tricky in that file. It's too early to use `test.pl`

nwc10:
@ilmari (as you might already know, but I did not until going for a big dig), the C code in `newLOOPOP` also used to handle `OPf_SPECIAL` being set on `ENTERSUB`, because `do foo() while 0` was subject to the same "run once" special casing as `do { ... } while 0`. Father C removed that in c96f4d3199da99e66c9172234f7dd6ff18ea8456. (It was no longer needed after your exorcism of do sub in 8c74b41425572faeb638f1269025b59d0785794f)

But as best I can tell, the (other) code that set `OPf_SPECIAL` you removed recently started out in 5.000 on the internal implicit call to "import"/"unimport", and then just carefully got cargo-culted to all other places that added implicit calls, without anyone ever realising that nothing further took any notice of that flag. The code in `newLOOPOP` was only reachable for postfix `while` and `until`, which just aren't relevant for internal calls.

t/cmd/mod.t
leonerd: Oh hmm... OK that's fair ;)

leonerd (approved):

PR #20040, opened 2022-08-04
t/op/sselect.t: skip 4 tests under miniperl

jkeenan:
Commit 85907e6fd46 (July 22 2022) added unit tests to t/op/sselect.t but
failed to take into consideration that this file is run as part of 'make
minitest'.  When we run that command, we get this result:

    t/op/sselect .... FAILED--unexpected output at test 17

... which is due in part to a quirk in t/TEST.  To avoid this problem we
should SKIP the block of tests in question if we're running with
miniperl.

### DIFF
--- a/t/op/sselect.t
+++ b/t/op/sselect.t
@@ -106,6 +106,7 @@ select ($tie, undef, undef, $tie);
 ok("no crash from select $numeric_tie, undef, undef, $numeric_tie");
 
 SKIP: {
+    skip "Can't load modules under miniperl", 4 if is_miniperl;
     my $SKIP_CR = sub {
         skip shift, 4;
     };


jkeenan:
Commit 85907e6fd46 (July 22 2022) added unit tests to `t/op/sselect.t` but failed to take into consideration the fact that this file is run as part of `make minitest`.  When we run that command, we get this result:
```
t/op/sselect ................... FAILED--unexpected output at test 17
```
... and as a result `make minitest` fails.
```
...
Failed 1 test out of 362, 99.72% okay.
	op/sselect.t
...
u=1.93  s=0.38  cu=62.01  cs=8.85  scripts=362  tests=462529
*** Error code 1

Stop.
```
When we run this test file with `miniperl`, we get inelegant results.
```
$ make minitest_prep
...
$  ./miniperl t/op/sselect.t
1..23
ok 1 - select undef  $blank $blank 0
[snip]
ok 15 - RT120102
ok 16
ok 17 # skip Failed to load POSIX.pm: Can't locate POSIX.pm in @INC (you may need to install the POSIX module) (@INC contains: . ../lib .) at t/op/sselect.t line 117.

ok 18 # skip Failed to load POSIX.pm: Can't locate POSIX.pm in @INC (you may need to install the POSIX module) (@INC contains: . ../lib .) at t/op/sselect.t line 117.

ok 19 # skip Failed to load POSIX.pm: Can't locate POSIX.pm in @INC (you may need to install the POSIX module) (@INC contains: . ../lib .) at t/op/sselect.t line 117.

ok 20 # skip Failed to load POSIX.pm: Can't locate POSIX.pm in @INC (you may need to install the POSIX module) (@INC contains: . ../lib .) at t/op/sselect.t line 117.
```
(The program itself is reported to pass.)

The language in the `make minitest` failure message for `t/op/sselect.t` comes from `t/TEST`:
```
 723                     else {
 724                         # module tests are allowed extra output,
 725                         # because Test::Harness allows it
 726                         next if $test =~ /^\W*(cpan|dist|ext|lib)\b/;
 727                         $failure = "FAILED--unexpected output at test $next";
 728                         last;
 729                     }
```
I believe it's the `@INC contains: . ../lib` which triggers this error.  While we might fiddle with that block in `t/TEST`, I believe the simpler and more immediate solution is to `SKIP` the test block when using `miniperl`.

t/op/sselect.t
```diff
@@ -106,65 +106,68 @@ select ($tie, undef, undef, $tie);
...
-    my $mask;
-
-    for (my $f=0; $f<100; $f++) {
-        my $fd = POSIX::dup(fileno \*STDOUT);
-
-        if (!defined $fd) {
-            $SKIP_CR->("dup(STDOUT): $!");
-            last UTF8TEST;
+    skip "Can't load modules under miniperl", 4, if is_miniperl;
+    SKIP: {
```
Leont: This nesting of `SKIP` blocks is unnecessary

Leont: This `SKIP` block is nested and hence not needed.

jkeenan: I believe this is fixed in next commit; please re-review.

jkeenan:
@Leont, other reviewers:  Is the current state of this branch acceptable to merge?

Leont:
> @Leont, other reviewers: Is the current state of this branch acceptable to merge?

It makes more sense to merge the two commits, but otherwise I approve.

t/op/sselect.t
```diff
@@ -106,6 +106,7 @@ select ($tie, undef, undef, $tie);
 ok("no crash from select $numeric_tie, undef, undef, $numeric_tie");
 
 SKIP: {
+    skip "Can't load modules under miniperl", 4, if is_miniperl;
```
FGasper: There’s a superfluous comma here after `4`.

I guess if Perl doesn’t complain about it, though, it’s no big deal.

FGasper (approved):

t/op/sselect.t
jkeenan: > There’s a superfluous comma here after `4`.
> 
> I guess if Perl doesn’t complain about it, though, it’s no big deal.

Thanks; will fix.

PR #20044, opened 2022-08-05
dl_dyld.xs should not create a `mode` variable only to ignore it

nwc10:
This unused variable dates back to when the file was first added. The code
was structured as a static C wrapper emulating the dlopen() API with the
native APIs, and then a copy of the existing XS code that called dlopen().
However, there is (and was) no need to keep it exactly this way, as the
wrapper is only visible inside this file, meaning that it does not need to
conform to any externally defined prototype.


tonycoz (approved):
Looks sane to me.

PR #20047, opened 2022-08-05
Errno pm modeline 202208050957

jkeenan:


ext/Errno/Errno_pm.PL
```diff
@@ -469,7 +473,6 @@ under the same terms as Perl itself.
 
 =cut
 
-# ex: set ro:
 ESQ
```
bram-perl, line 479: At the risk of starting bikeshedding:  what I would probably do is replace it with something like:

```
ESQ

    # use string concatenation to prevent vi/vim from parsing the next line as a modeline for *this* file
    print "# e" . "x: set ro\n";

```

That should be more then enough to stop the modeline parsing since vi looks for the exact phrase `ex:`
(and then it can stay near the end of the file/or anywhere in the file for that matter)

Either way: if you go with your change then it at least should have a comment on why it's placed where it is.

ext/Errno/Errno_pm.PL
```diff
@@ -272,6 +274,9 @@ sub write_errno_pm {
 
     # Write Errno.pm
 
+    open OUT, '>', $errno_tmp or die "Cannot open $errno_tmp: $!";
+    select OUT;
```
Leont, line 281: Instead of this select, can't we just `print $out <<"EDQ"`

jkeenan: I tried this approach.  The problem is that the generated file is not nicely contained within one single heredoc.  There are 9 separate `print` statements that are used to generate `Error.pm`.
```
280:    print <<"EDQ";
297:        print <<"CONFIG_CHECK_END";
306:    print <<"EDQ";
320:	print "\t$err => $err{$err},\n";
323:print <<'ESQ';
368:    print "\t",$k,"\n    )],\n";
371:	print "    WINSOCK => [qw(\n";
374:	print "\t",$k,"\n    )],\n";
377:    print ");\n\n";
379:    print <<'ESQ';
```
The `select` enables us to avoid typing `print $OUT` in each of those statements.  I suspect that was Gurusamy Sarathy's reasoning when he introduced the `select` back in eab60bb1f2e in 1998.

ext/Errno/Errno_pm.PL
jkeenan: I will accept your point about adding a comment (now squashed into 3rd commit), but otherwise I'm going to stick with the approach I've taken.  If you disagree, then I recommend you file your own p.r. and we'll let the PSC choose.

bram-perl: Do I think that"s a bad call? Yes.
Do I care enough about it to file my own pr and to 'waste' PSC time on this? No.

(Why I consider it a bad call: it makes the code more complex + it's not guaranteed to work. If someone configures there modeline to look at 51 lines then it'll be broken for them.)

demerphq: I second @bram-perl's point.There are any number of ways you can obfuscate the modeline. I suspect a simple escape on the colon would do it actually, which would be even easier than brams proposal.  A crafty way would be to replace "ex" with `${\'ex'}` :-)

FWIW, @jkeenan I think this is a pretty reasonable request that is too minor to justify escalation to the PSC. There are a handful of simple ways you can obfuscate the modeline so a modeline parser doesn't treat it as /the/ modeline for the file and I find it hard to understand why you object. If it were my patch id be grateful for the suggestion.

PR #20049, opened 2022-08-05
Document postderef_qq support for ->$#* interpolation

zmughal:
This was added in commit <https://github.com/Perl/perl5/commit/ff25e5dbbad6ccf83f2e2a874a3c90294ea8cb47>.

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -272,9 +272,10 @@ regardless of what feature declarations are in scope.
 =head2 The 'postderef' and 'postderef_qq' features
 
 The 'postderef_qq' feature extends the applicability of L<postfix
-dereference syntax|perlref/Postfix Dereference Syntax> so that postfix array
-and scalar dereference are available in double-quotish interpolations. For
-example, it makes the following two statements equivalent:
+dereference syntax|perlref/Postfix Dereference Syntax> so that
+postfix array dereference, postfix scalar dereference, and
+postfix array highest index access are available in double-quotish interpolations.
+For example, it makes the following two statements equivalent:
 
   my $s = "[@{ $h->{a} }]";
   my $s = "[$h->{a}->@*]";
--- a/regen/feature.pl
+++ b/regen/feature.pl
@@ -687,9 +687,10 @@ regardless of what feature declarations are in scope.
 =head2 The 'postderef' and 'postderef_qq' features
 
 The 'postderef_qq' feature extends the applicability of L<postfix
-dereference syntax|perlref/Postfix Dereference Syntax> so that postfix array
-and scalar dereference are available in double-quotish interpolations. For
-example, it makes the following two statements equivalent:
+dereference syntax|perlref/Postfix Dereference Syntax> so that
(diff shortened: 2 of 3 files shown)


lib/feature.pm
```diff
@@ -271,8 +271,9 @@ regardless of what feature declarations are in scope.
 
 The 'postderef_qq' feature extends the applicability of L<postfix
 dereference syntax|perlref/Postfix Dereference Syntax> so that postfix array
-and scalar dereference are available in double-quotish interpolations. For
-example, it makes the following two statements equivalent:
+and scalar dereference and postfix array length access are available in
+double-quotish interpolations. For example, it makes the following two
```
jkeenan: The "X and Y and Z" syntax is sub-optimal English style.  You probably want:
```
dereference syntax|perlref/Postfix Dereference Syntax> so that postfix array,
scalar dereference and postfix array length access are available in
double-quotish interpolations.
```

regen/feature.pl
```diff
@@ -683,8 +683,9 @@ =head2 The 'postderef' and 'postderef_qq' features
 
 The 'postderef_qq' feature extends the applicability of L<postfix
 dereference syntax|perlref/Postfix Dereference Syntax> so that postfix array
-and scalar dereference are available in double-quotish interpolations. For
-example, it makes the following two statements equivalent:
+and scalar dereference and postfix array length access are available in
+double-quotish interpolations. For example, it makes the following two
```
jkeenan: Same as above.

jkeenan (changes requested):

lib/feature.pm
zmughal: I was going for it to parse as:
```
  (postfix array and scalar dereference) and (postfix array length access)
```
Perhaps some other way of rearranging it might work:
```
  postfix array/scalar dereference and postfix array length access
```
?

zmughal: Or use separate sentences like I did in `perlref.pod`.

zmughal: I'll probably go with

> postfix array dereference, postfix scalar dereference, and postfix array length access

as this is the least ambiguous despite the extra length.

jkeenan:
The English style of the changes is now good.  But I'd like someone else to review the content of the changes, as I myself have never used this Perl syntax.  @rjbs perhaps?

khwilliamson:
This addition is needed, but it is misleading.  The number returned is related to the array length, but it is one off.  The number returned is the index of the highest element in the array.  So if the array has 10 elements, 0..9, this returns 9.  This needs to be clarified

zmughal:
I have updated the text again with @khwilliamson's suggestions. I think I might need to do a rebase again?

khwilliamson:
Unfortunately, use of the word 'last' is ambiguous.  It can mean the previous, or the final.  I suggest using 'highest' instead.

I don't think rebasing is required at this point.

khwilliamson:
Thanks

This was merged via 365f3e0f038557f283dc48f11d9dd1d2eb0dfb8b

PR #20052, opened 2022-08-06
Add support for PERL_RAND_SEED env var

demerphq:
The idea of this env var is to provide consistent behavior to scripts that call `srand()` with no arguments, either implicitly or explicitly. This is implemented in such a way that each process in the process tree should get unique but repeatable initialization.

Eg

```
$ PERL_RAND_SEED=123 ./perl -Ilib -le'print map { chr(65+rand 26) } 1..10'
HKYDDJQVSD
$ PERL_RAND_SEED=123 ./perl -Ilib -le'print map { chr(65+rand 26) } 1..10'
HKYDDJQVSD
```

Should produce the same results in each time. Similarly if the program forks() or starts subthreads the Nth forked child should get the same initialization each time, and each child should be seeded differently from the parent similarly to how the seed would normally be randomly initialized. In the below example the `| sort -n` sorts the results by which fork they come from so that it is more obvious they are producing the same results 

```
$ PERL_RAND_SEED=123 ./perl -Ilib -le'for my $n (1..5) { my $pid= fork(); if (!$pid) { print $n,map { chr(65+rand 26) } 1..10; exit} }' | sort -n
1ZFKXOTACST
2IXQXRCLXGV
3YZSNSMOBHX
4BNSLRHINMT
5SFCPDTHIRV

$ PERL_RAND_SEED=123 ./perl -Ilib -le'for my $n (1..5) { my $pid= fork(); if (!$pid) { print $n,map { chr(65+rand 26) } 1..10; exit} }' | sort -n
1ZFKXOTACST
2IXQXRCLXGV
3YZSNSMOBHX
4BNSLRHINMT
5SFCPDTHIRV
```
The exact seeding is deliberately unspecified, although in practice when no sub processes are involved it should be equivalent to calling `srand($ENV{PERL_RAND_SEED})` at the top of the script, however each fork() or thread->create() call could change the exact value the mother process uses if the use of `rand()` or `srand()` occurs post fork. The main point is that if the program is executed twice it should produce the exact same results from `rand()` and `srand()` assuming all other factors are equal:

```
PERL_RAND_SEED=1 ./perl -Ilib -le'for my $n (1..5) { my $pid= fork(); if ($pid) { push @pids, $pid} else { print join ", ",$n,map { sprintf "%08x", srand() } 1..10; exit} } waitpid $_,0 for @pids; printf "%08x\n", srand() for 1..5'
1, 80a42501, 8484e763, 1579e1b7, 00adb710, 7fb26ad7, 6ffae7c3, 53515602, 4bb4738a, 7d6f7c68, cea49639
2, 8dcef731, 36372a05, bd8b947e, 2f824c7c, a7b2103a, 53117ee0, e5655046, a2db49f0, b740ac42, fbdf19e6
3, 7de9d161, bf7f82e3, ab5b6b32, 01944dac, 8c10be3c, e1390937, c3eeae38, d31d00ac, 121f9c28, d2c77045
4, be5614a4, e497f265, 41b0be08, 98433330, 3a8cc862, d39bede2, 6ad8a4d1, b615c5f0, dbd39605, 8b7f330c
5, 8a3616fb, 55c3036f, 8071ceb8, 8d74df0b, c4233c61, d646fb76, 2c2fc3a2, 5f26d86f, 0144353a, 7b85efab
00000001
2c6f5bd0
25b2331a
19f91cb2
77877125
```

The new env var does NOT change behavior of scripts using srand() with an argument.

This can be combined with `PERL_HASH_SEED` and `PERL_INTERNAL_RAND_SEED` to produce a repeatable test run of of code that does not hard initialize the random number generator. We document that users cannot expect consistency except in the most constrained circumstance of a given perl executable, but in practice the results should be stable (all other things considered) which we porters can exploit to test the feature in the perl test suite.

This includes a few patches for issues that came up while I testing this with our test suite along with PERL_HASH_SEED=0 and PERL_INTERNAL_RAND_SEED=123.

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -2960,6 +2960,31 @@ last-inclusive range.
 
 #endif
 
+/* These are simple Marsaglia XOR-SHIFT RNG's for 64 and 32 bits. These
+ * RNG's are of reasonable quality, very fast, and have the interesting
+ * property that provided 'x' is non-zero they create a cycle of 2^32-1
+ * or 2^64-1 "random" like numbers, with the exception of 0. Thus they
+ * are very useful when you want an integer to "dance" in a random way,
+ * but you also never want it to become 0 and thus false.
+ *
+ * Obviously they leave x unchanged if it starts out as 0. */
+
+#define PERL_XORSHIFT64(x)      \
+STMT_START {                    \
+    (x) ^= (x) << 13;           \
+    (x) ^= (x) >> 17;           \
+    (x) ^= (x) << 5;            \
+} STMT_END
+
+/* 32 bit version */
+#define PERL_XORSHIFT32(x)           \
+STMT_START {                    \
+    (x) ^= (x) << 13;           \
+    (x) ^= (x) >> 7;            \
+    (x) ^= (x) << 17;           \
+} STMT_END
+
+
 #endif  /* PERL_HANDY_H_ */
 
 /*
--- a/handy.h
+++ b/handy.h
@@ -2967,21 +2967,75 @@ last-inclusive range.
  * are very useful when you want an integer to "dance" in a random way,
  * but you also never want it to become 0 and thus false.
  *
- * Obviously they leave x unchanged if it starts out as 0. */
+ * Obviously they leave x unchanged if it starts out as 0.
+ *
+ * We have two variants just because that can be helpful in certain
+ * places. There is no advantage to either, they are equally bad as each
+ * other as far RNG's go. Sufficiently random for many purposes, but
+ * insufficiently random for serious use as they fail important tests in
+ * the Test01 BigCrush RNG test suite by L’Ecuyer and Simard. (Note
+ * that Drand48 also fails BigCrush). The main point is they produce
+ * different sequences and in places where we want some randomlike
+ * behavior they are cheap and easy.
+ *
+ * Marsaglia was one of the early researchers into RNG testing and wrote
+ * the Diehard RNG test suite, which after his death become the
+ * Dieharder RNG suite, and was generally supplanted by the Test01 suite
+ * by L'Ecruyer and associates.
+ *
+ * There are dozens of shift parameters that create a pseudo random ring
+ * of integers 1..2^N-1, if you need a different sequence just read the
+ * paper and select a set of parameters. In fact, simply reversing the
+ * shift order from L/R/L to R/L/R should result in another valid
+ * example, but read the paper before you do that.
+ *
+ * PDF of the original paper:
+ *  https://www.jstatsoft.org/article/download/v008i14/916
+ * Wikipedia:
+ *  https://en.wikipedia.org/wiki/Xorshift
+ * Criticism:
+ *  https://www.iro.umontreal.ca/~lecuyer/myftp/papers/xorshift.pdf
+ * Test01:
+ *  http://simul.iro.umontreal.ca/testu01/tu01.html
+ * Diehard:
+ *  https://en.wikipedia.org/wiki/Diehard_tests
+ * Dieharder:
+ *  https://webhome.phy.duke.edu/~rgb/General/rand_rate/rand_rate.abs
+ *
+ */
 
-#define PERL_XORSHIFT64(x)      \
+/* 32 bit version */
+#define PERL_XORSHIFT32_A(x)    \
 STMT_START {                    \
-    (x) ^= (x) << 13;           \
-    (x) ^= (x) >> 17;           \
-    (x) ^= (x) << 5;            \
+    (x) ^= ((x) << 13);         \
+    (x) ^= ((x) >> 17);         \
+    (x) ^= ((x) << 5);          \
+} STMT_END
+
+/* 64 bit version */
+#define PERL_XORSHIFT64_A(x)    \
+STMT_START {                    \
+    (x) ^= ((x) << 13);         \
+    (x) ^= ((x) >> 7);          \
+    (x) ^= ((x) << 17);         \
 } STMT_END
 
 /* 32 bit version */
-#define PERL_XORSHIFT32(x)           \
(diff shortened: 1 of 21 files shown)


handy.h
```diff
@@ -2960,6 +2960,53 @@ last-inclusive range.
...
+ * */
+
+#define PERL_XORSHIFT32_A(x)    \
+STMT_START {                    \
+    (x) ^= ((x) << 13);         \
+    (x) ^= ((x) >> 17);         \
+    (x) ^= ((x) << 5);          \
+} STMT_END
+
+/* 32 bit version */
```
tonycoz: Presumably this should be "64 bit version", going by the name and the usage when `IVSIZE == 8` below.

handy.h
```diff
@@ -2960,6 +2960,53 @@ last-inclusive range.
...
+#define PERL_XORSHIFT32_B(x)    \
+STMT_START {                    \
+    (x) ^= ((x) << 5);          \
+    (x) ^= ((x) >> 27);         \
+    (x) ^= ((x) << 8);          \
+} STMT_END
+
+#define PERL_XORSHIFT64_B(x)    \
```
tonycoz, line 3034: `PERL_XORSHIFT64_B()` is defined here but isn't used anywhere.

Should `PERL_SRAND_OVERRIDE_NEXT_CHILD()` in perl.h be using it on 64-bit platforms?

demerphq: I made the PL_srand_override be a U32 because most of the rest of the seeding logic seemed to use a U32.  So I am only using the _B variant of the 32 bit. I figured that if we supply a 32 bit _B variant "for completeness" we should supply a 64 bit as well even if we dont use it right now. I realize "completeness" is a threadbare excuse, if someone thought we shouldn't bother I would have no problem removing it.

handy.h
demerphq: I will check the comment. Thanks!

handy.h
demerphq: I enhanced the comments to explain, but as I said, im fine with removing the 64_B variant if you want.

handy.h
demerphq: comment fixed.

pod/perlfunc.pod
```diff
@@ -8484,6 +8484,16 @@ combinations to test comprehensively in the time available to it each run.  It
...
 
+The C<PERL_RAND_SEED> environment var may be used to cause C<srand()> called
+without arguments to return a consistent seed each time it is called (whether
+called explicitly or implicitly). The exact seeding is unspecified, but
+using different values for C<PERL_RAND_SEED> should produce different
+results. This is intended for debugging and is only guaranteed to produce
+consistent results between invocations of the same perl executable. Note
+that this env variable is read once during process startup, changing it
+during the program flow will not affect the currently running process.
+See L<perlrun> for more details.
```
tonycoz: Should this mention that the value of `PERL_RAND_SEED` needs to be a non-negative integer?  (the perlrun change does mention it)

Also "environment var", "env variable", should these be spelled in full (or be consistent)?

demerphq: I guess since you bring it up then yes. :-) Ill make both changes later today.

tonycoz (approved):

demerphq:
BTW that last push is a rebase of the previous doc patch fixes, along with some whitespace cleanup i forgot when i was hacking on this. I will merge once it passes CI thanks to @tonycoz for the approval.

Tux:
FYI : these new attributes have been added to Perl::Config::V as they
are defined in the portability section(s) - https://github.com/Tux/Config-Perl-V/commit/506e8490669446fb10fbeb24eb9f8c3baad7449c
Also merged to perl CORE in 3eb80770405d1994612fe1040f8587f61d6f22cd

demerphq:
On Thu, 25 Aug 2022 at 14:19, H.Merijn Brand ***@***.***>
wrote:

> FYI : these new attributes have been added to Perl::Config::V as they
> are defined in the portability section(s) - ***@***.***
> <https://github.com/Tux/Config-Perl-V/commit/506e8490669446fb10fbeb24eb9f8c3baad7449c>
> Also merged to perl CORE in 3eb8077
> <https://github.com/Perl/perl5/commit/3eb80770405d1994612fe1040f8587f61d6f22cd>
>

 Can you expand on that a bit? Is there something i should have done?

Yves

-- 

Tux:
No, this is all automated on my side. You could have pinged me though.
This reply was merely to make you notice that changes in those sections might have consequences in other modules, and as Config::Perl::V is CORE, awareness is good :)
Config::Perl::V takes all attributes that alter the compile-time behavior (those are reported with `perl -V`) and creates a unique checksum, so one can verify that the configuration of two perls are exactly the same. In order to do so it needs to know all defines that ever (could have been) used.

demerphq:
On Thu, 25 Aug 2022 at 15:18, H.Merijn Brand ***@***.***> wrote:
>
> No, this is all automated on my side. You could have pinged me though.
> This reply was merely to make you notice that changes in those sections might have consequences in other modules, and as Config::Perl::V is CORE, awareness is good :)
> Config::Perl::V takes all attributes that alter the compile-time behavior (those are reported with perl -V) and creates a unique checksum, so one can verify that the configuration of two perls are exactly the same. In order to do so it needs to know all defines that ever (could have been) used.

We should talk. We are not reporting nearly anything like all the
options that are relevant, and we are not very good about determining
if they affect binary compatibility. I have a "rainy day hack" todo
item to auto extract every define so that what is reported by -V is
actually correct. We were wrong in various regards for quite a while
without noticing. :-(

Yves



-- 

Tux:
Then let this be our wakeup call :)

See also https://github.com/Tux/Config-Perl-V/blob/master/xt/10_check-core.t

PR #20055, opened 2022-08-06
Errno.t: provide labels for tests lacking them

jkeenan:
If any of these test descriptions are misleading, please suggest improvements.

### DIFF
--- a/ext/Errno/t/Errno.t
+++ b/ext/Errno/t/Errno.t
@@ -14,29 +14,33 @@ BAIL_OUT("No errno's are exported") unless @Errno::EXPORT_OK;
 my $err = $Errno::EXPORT_OK[0];
 my $num = &{"Errno::$err"};
 
-is($num, &{"Errno::$err"});
+is($num, &{"Errno::$err"},
+    'element in @Errno::EXPORT_OK found via sub call');
 
 $! = $num;
-ok(exists $!{$err});
+ok(exists $!{$err}, 'entry in %! reflects current value of $!');
 
 $! = 0;
-ok(! $!{$err});
+ok(! $!{$err}, 'entry in %! reflects the current value of $!');
 
-ok(join(",",sort keys(%!)) eq join(",",sort @Errno::EXPORT_OK));
+ok(join(",",sort keys(%!)) eq join(",",sort @Errno::EXPORT_OK),
+    'keys of %! match keys of @Errno::EXPORT_OK');
 
 eval { exists $!{[]} };
-ok(! $@);
+ok(! $@, "no exception recorded in %! when element's key is '[]'");
 
 eval {$!{$err} = "qunckkk" };
-like($@, qr/^ERRNO hash is read only!/);
+like($@, qr/^ERRNO hash is read only!/,
+    "can't assign to ERRNO hash: 'ERRNO hash is read only!'");
 
 eval {delete $!{$err}};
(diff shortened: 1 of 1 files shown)


ext/Errno/t/Errno.t
```diff
@@ -14,29 +14,33 @@ BAIL_OUT("No errno's are exported") unless @Errno::EXPORT_OK;
...
-ok(! $!{$err});
+ok(! $!{$err}, 'no true value for error in %!');
 
-ok(join(",",sort keys(%!)) eq join(",",sort @Errno::EXPORT_OK));
+ok(join(",",sort keys(%!)) eq join(",",sort @Errno::EXPORT_OK),
+    'keys of %! match keys of @Errno::EXPORT_OK');
 
 eval { exists $!{[]} };
-ok(! $@);
+ok(! $@, "no error from most recent 'eval'");
```
tonycoz: If you look back at the commit that introduced this test, it's checking that exists isn't throwing an exception when the key isn't a sub ref or a name.

Maybe `'$!{[]} doesn't throw an exception'` would be better.

ext/Errno/t/Errno.t
```diff
@@ -14,29 +14,33 @@ BAIL_OUT("No errno's are exported") unless @Errno::EXPORT_OK;
...
+is($num, &{"Errno::$err"},
+    'element in @Errno::EXPORT_OK found via sub call');
 
 $! = $num;
-ok(exists $!{$err});
+ok(exists $!{$err}, 'error found in %!');
 
 $! = 0;
-ok(! $!{$err});
+ok(! $!{$err}, 'no true value for error in %!');
```
tonycoz: These describe what the ok() is doing, but don't really describe the point of the test - in each case whether the entry in `%!` reflects the current value of `$!`.

ext/Errno/t/Errno.t
jkeenan: Is this the commit which you were referring to?
```
commit f8ed3476f18eeb645feeb028cc05ab8e72bae078
Author:     Rafael Garcia-Suarez <rgarciasuarez@gmail.com>
AuthorDate: Mon Oct 25 15:31:21 2004 +0000
Commit:     Rafael Garcia-Suarez <rgarciasuarez@gmail.com>
CommitDate: Mon Oct 25 15:31:21 2004 +0000

    Fix [perl #32130] Errno.pm must not pass references to "prototype"
```
If so, could I use that commit message as the basis for a better test description?

ext/Errno/t/Errno.t
jkeenan: Adapted test descriptions in 80f77bb58d; please re-review.

ext/Errno/t/Errno.t
tonycoz: Sure, but realize it isn't testing whether references are being passed to prototype (a user of the module doesn't care about that the implementation), but whether `$!{...}` is throwing an exception.

ext/Errno/t/Errno.t
tonycoz: Good enough.

ext/Errno/t/Errno.t
jkeenan: Please review ac7e1be4ec.  If it's okay, I'll squash and merge.  Thanks.

tonycoz (approved):

PR #20056, opened 2022-08-06
Reduce tech debt in define related logic in perl.c and hv_func.h

demerphq:
hv_func.h is chocker full of reserved identifiers, this renames them so they start with PVT and not underbars.

perl.c is chocker full of define checks for non-existent defines related to the hash function. This removes them. At the same time it replaces the fragile mechanism used currently with something a little more likely to be future proof.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -450,30 +450,39 @@ storage with 256 32-bit random values as well. In practice the RNG we use
 for seeding the SBOX32 storage is very efficient, and populating the table
 required for hashing even fairly long keys is negligible as we only do it
 during startup. By default we build with SBOX32 enabled, but you can change
-that by setting
+that by setting the C<PERL_HASH_USE_SBOX32_ALSO> in the Configure process,
+with something like this
 
-   PERL_HASH_USE_SBOX32_ALSO
+   -Accflags='-DPERL_HASH_USE_SBOX32_ALSO=0'
 
-to zero in configure. By default Perl will use SBOX32 to hash strings 24 bytes
-or shorter, you can change this length by setting
+or alternatively you can use the simple define C<PERL_HASH_NO_SBOX32> like this:
 
-    SBOX32_MAX_LEN
+    -Accflags='-DPERL_HASH_NO_SBOX32'
 
-to the desired length, with the maximum length being 256.
+By default Perl will use SBOX32 to hash strings 24 bytes
+or shorter, you can change this length by setting C<SBOX32_MAX_LEN>
+to the desired length, with the maximum length being 256. For example with
+this:
+
+    -Accflags='-DSBOX_MAX_LEN=128'
 
 As of Perl 5.18 the order returned by keys(), values(), and each() is
 non-deterministic and distinct per hash, and the insert order for
 colliding keys is randomized as well, and perl allows for controlling this
 by the PERL_PERTURB_KEYS environment setting. You can disable this behavior
-entirely with the define
+entirely with the define C<PERL_PERTURB_KEYS_DISABLED> with
 
-    PERL_PERTURB_KEYS_DISABLED
+    -Accflags='-DPERL_PERTURB_KEYS_DISABLED'
 
 You can disable the environment variable checks and compile time specify
-the type of key traversal randomization to be used by defining one of these:
+the type of key traversal randomization to be used by defining either
+C<PERL_PERTURB_KEYS_RANDOM> or C<PERL_PERTURB_KEYS_DETERMINISTIC> with
 
-    PERL_PERTURB_KEYS_RANDOM
-    PERL_PERTURB_KEYS_DETERMINISTIC
+    -Accflags='-DPERL_PERTURB_KEYS_RANDOM'
+
+or
+
+    -Accflags='-DPERL_PERTURB_KEYS_DETERMINISTIC'
 
 Since Perl 5.18 the seed used for the hash function is randomly selected
 at process start, which can be overridden by specifying a seed by setting
@@ -489,9 +498,9 @@ DETERMINISTIC in this context means "if everything else is kept the same
 the same results should be observed".
 
 You can change this behavior so that your perl is built with a hard coded
-seed with the define
+seed with the define C<NO_HASH_SEED> by providing to Configure
 
-    NO_HASH_SEED
+    -Accflags='-DNO_HASH_SEED'
 
 Note that if you do this you should modify the code in hv_func.h to specify
 your own key. In the future this define may be renamed and replaced with one
--- a/hv_func.h
+++ b/hv_func.h
@@ -34,40 +34,40 @@
 #include "sbox32_hash.h"
 
 #if defined(PERL_HASH_FUNC_SIPHASH)
-# define __PERL_HASH_FUNC "SIPHASH_2_4"
-# define __PERL_HASH_WORD_TYPE U64
-# define __PERL_HASH_WORD_SIZE sizeof(__PERL_HASH_WORD_TYPE)
-# define __PERL_HASH_SEED_BYTES (__PERL_HASH_WORD_SIZE * 2)
-# define __PERL_HASH_STATE_BYTES (__PERL_HASH_WORD_SIZE * 4)
-# define __PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
-# define __PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_2_4_with_state((state),(U8*)(str),(len))
+# define PVT__PERL_HASH_FUNC "SIPHASH_2_4"
+# define PVT__PERL_HASH_WORD_TYPE U64
+# define PVT__PERL_HASH_WORD_SIZE sizeof(PVT__PERL_HASH_WORD_TYPE)
+# define PVT__PERL_HASH_SEED_BYTES (PVT__PERL_HASH_WORD_SIZE * 2)
+# define PVT__PERL_HASH_STATE_BYTES (PVT__PERL_HASH_WORD_SIZE * 4)
+# define PVT__PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
+# define PVT__PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_2_4_with_state((state),(U8*)(str),(len))
 #elif defined(PERL_HASH_FUNC_SIPHASH13)
-# define __PERL_HASH_FUNC "SIPHASH_1_3"
-# define __PERL_HASH_WORD_TYPE U64
-# define __PERL_HASH_WORD_SIZE sizeof(__PERL_HASH_WORD_TYPE)
-# define __PERL_HASH_SEED_BYTES (__PERL_HASH_WORD_SIZE * 2)
-# define __PERL_HASH_STATE_BYTES (__PERL_HASH_WORD_SIZE * 4)
-# define __PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
-# define __PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_1_3_with_state((state),(U8*)(str),(len))
+# define PVT__PERL_HASH_FUNC "SIPHASH_1_3"
+# define PVT__PERL_HASH_WORD_TYPE U64
+# define PVT__PERL_HASH_WORD_SIZE sizeof(PVT__PERL_HASH_WORD_TYPE)
+# define PVT__PERL_HASH_SEED_BYTES (PVT__PERL_HASH_WORD_SIZE * 2)
+# define PVT__PERL_HASH_STATE_BYTES (PVT__PERL_HASH_WORD_SIZE * 4)
+# define PVT__PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
+# define PVT__PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_1_3_with_state((state),(U8*)(str),(len))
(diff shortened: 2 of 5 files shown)


jkeenan:
CI failing on `t/porting/bincompat.t`; please check.

jkeenan:
> hv_func.h is chocker full of reserved identifiers, this renames them so they start with PVT and not underbars.
> 
> perl.c is chocker full of define checks for non-existent defines related to the hash function. This removes them. At the same time it replaces the fragile mechanism used currently with something a little more likely to be future proof.

Can you explain how these problems meet some definition of "technical debt"?  What is it that is fragile and how will these commits reduce that fragility?

demerphq:
It is technical debt if we have code like this:

```
#ifdef SOME_DEFINE_THAT_DOESNT_EXIST
...
#endif
```

It is debt in the sense that it makes people wonder "what about this define" and then have to search to find out in fact the define doesn't exist and isn't used. 

It is also technical debt because there is double entry involved here if we wish to prevent errors and oversight. Years ago we removed most of the hash functions and related defines that were mentioned in perl.c, yet nothing noticed that we had removed them. Part of the problem is that in perl.c we had to enumerate every possible define related to the hash function selection, even though we can only use one hash function during build time and even if more than one of them were defined only one of them would take effect.

BY restructuring this so that we introduce the new `PERL_HASH_FUNC_DEFINE` which is defined to be a string which says which `PERL_HASH_FUNC_...` was defined we keep the bookkeeping for which hash function is in use in hv_func.h where they belong. If we add a new hash function, or remove one of the ones we currently support, then we will have to update hv_func.h, and if we do it appropriately then perl.c wont need to be touched.  It will just use `PERL_HASH_FUNC_DEFINE` and everything should Just Work Out Fine.

The need to do duplicate bookkeeping for something in our code is a canonical example of technical debt.  By removing it we have to do half the work in future that we might have, and we remove the chance that an error occurs if we do so. If you look closely the old code in perl.c would not report that we were using `PERL_HASH_FUNC_SIPHASH13`, but it would report that we were using `PERL_HASH_FUNC_ONE_AT_A_TIME` if we still had support for doing so. In fact the code in perl.c could end up reporting that we are using both `PERL_HASH_FUNC_ONE_AT_A_TIME` and `PERL_HASH_FUNC_SDBM` at the same time when in fact we are actually using `PERL_HASH_FUNC_SIPHASH13`. Try adding `-Accflags='-DPERL_HASH_FUNC_ONE_AT_A_TIME' -Accflags='-DPERL_HASH_FUNC_SDBM' -Accflags='-DPerl_HASH_FUNC_SIPHASH13'` to your configure invocation, I bet that what perl -V says doesnt line up with what `PERL_HASH_SEED_DEBUG` says, when configure and make have finished try running

```
PERL_HASH_SEED_DEBUG=1 ./perl -Ilib -V
```

in the build directory. If I have that ccflags stuff right above you should see Perl -v claim to be using both SDBM and ONE_AT_A_TIME at the same time, which is impossible, and the PERL_HASH_SEED_DEBUG output will say something about SBOX32 and SIPHASH which is what is actually happening. With my patch we should "get it right" and have the debug output match -V.

demerphq:
Latest patch should fix the bincompat.t issues along with a few other improvements, mostly to the docs in INSTALL.

Leont:
> Can you explain how these problems meet some definition of "technical debt"? What is it that is fragile and how will these commits reduce that fragility?

I think the commit messages explain that?

jkeenan:
`t/porting/bincompat.t` continues to fail one test at bfc193dd4acd40a4f5a5da179be68bf004a52416; see https://github.com/Perl/perl5/runs/7710849649?check_suite_focus=true.

```
$ cd t;./perl harness -v porting/bincompat.t; cd -
...
not ok 2 - C source code has non_bincompat_options in sorted order
# Failed test 2 - C source code has non_bincompat_options in sorted order at porting/bincompat.t line 18
#      got "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
# expected "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
```
If we insert some line breaks into that output for readability, we get:
```
not ok 2 - C source code has non_bincompat_options in sorted order
# Failed test 2 - C source code has non_bincompat_options in sorted order at porting/bincompat.t line 18
#      got "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV 
PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 
PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV 
USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"

# expected "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV 
PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV 
PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 
USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
```
The "expected" is not alphabetically sorted; the "got" is.  That's weird, isn't it?  Usually we would write a test to expect alphabetical sorting and fail the test if what we got was not.

So I'm not defending the way this test is written.  However, since the point of the test is apparently to enforce alphabetical order on certain entries in the source code, we can get the test to pass if we apply this patch to the branch in the pull request.
```
diff --git a/perl.c b/perl.c
index 0555c1296c..133b5d4392 100644
--- a/perl.c
+++ b/perl.c
@@ -1960,6 +1960,15 @@ S_Internals_V(pTHX_ CV *cv)
 #  ifdef PERL_EXTERNAL_GLOB
                              " PERL_EXTERNAL_GLOB"
 #  endif
+#  ifdef PERL_HASH_FUNC_DEFINE
+/* note that this is different from the others, PERL_HASH_FUNC_DEFINE
+ * is a string which says which define was defined. */
+                             " " PERL_HASH_FUNC_DEFINE
+#  endif
+#  ifdef PERL_HASH_USE_SBOX32
+                             " PERL_HASH_USE_SBOX32"
+#  endif
+
 #  ifdef PERL_IS_MINIPERL
                              " PERL_IS_MINIPERL"
 #  endif
@@ -1996,15 +2005,6 @@ S_Internals_V(pTHX_ CV *cv)
 #  ifdef PERL_USE_SAFE_PUTENV
                              " PERL_USE_SAFE_PUTENV"
 #  endif
-#  ifdef PERL_HASH_FUNC_DEFINE
-/* note that this is different from the others, PERL_HASH_FUNC_DEFINE
- * is a string which says which define was defined. */
-                             " " PERL_HASH_FUNC_DEFINE
-#  endif
-#  ifdef PERL_HASH_USE_SBOX32
-                             " PERL_HASH_USE_SBOX32"
-#  endif
-
 #  ifdef PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
                              " PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES"
 #  endif
```

bram-perl:
> Latest patch should fix the bincompat.t issues along with a few other improvements, mostly to the docs in INSTALL.

You've removed that commit?
That is: this PR doesn't contain the https://github.com/Perl/perl5/commit/ef6b2b73e7bba637f101cc96f2d59dba7dab76f3 commit?
(If that's intended no problem, just double checking)

demerphq:
On Sun, 7 Aug 2022 at 15:32, James E Keenan ***@***.***>
wrote:

> t/porting/bincompat.t continues to fail one test at bfc193d
> <https://github.com/Perl/perl5/commit/bfc193dd4acd40a4f5a5da179be68bf004a52416>;
> see https://github.com/Perl/perl5/runs/7710849649?check_suite_focus=true.
>
> $ cd t;./perl harness -v porting/bincompat.t; cd -
> ...
> not ok 2 - C source code has non_bincompat_options in sorted order
> # Failed test 2 - C source code has non_bincompat_options in sorted order at porting/bincompat.t line 18
> #      got "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
> # expected "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32 USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
>
> If we insert some line breaks into that output for readability, we get:
>
> not ok 2 - C source code has non_bincompat_options in sorted order
> # Failed test 2 - C source code has non_bincompat_options in sorted order at porting/bincompat.t line 18
> #      got "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV
> PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32
> PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV
> USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
>
> # expected "PERL_COPY_ON_WRITE PERL_DONT_CREATE_GVSV
> PERL_MALLOC_WRAP PERL_OP_PARENT PERL_PRESERVE_IVUV PERL_USE_DEVEL PERL_USE_SAFE_PUTENV
> PERL_HASH_FUNC_SIPHASH13 PERL_HASH_USE_SBOX32
> USE_LOCALE USE_LOCALE_CTYPE USE_PERL_ATOF USE_THREAD_SAFE_LOCALE"
>
> The "expected" is not alphabetically sorted; the "got" is. That's weird,
> isn't it? Usually we would write a test to expect alphabetical sorting and
> fail the test if what we got was not.
>
> So I'm not defending the way this test is written. However, since the
> point of the test is apparently to enforce alphabetical order on certain
> entries in the source code, we can get the test to pass if we apply this
> patch to the branch in the pull request.
>
> diff --git a/perl.c b/perl.c
> index 0555c1296c..133b5d4392 100644
> --- a/perl.c
> +++ b/perl.c
> @@ -1960,6 +1960,15 @@ S_Internals_V(pTHX_ CV *cv)
>  #  ifdef PERL_EXTERNAL_GLOB
>                               " PERL_EXTERNAL_GLOB"
>  #  endif
> +#  ifdef PERL_HASH_FUNC_DEFINE
> +/* note that this is different from the others, PERL_HASH_FUNC_DEFINE
> + * is a string which says which define was defined. */
> +                             " " PERL_HASH_FUNC_DEFINE
> +#  endif
> +#  ifdef PERL_HASH_USE_SBOX32
> +                             " PERL_HASH_USE_SBOX32"
> +#  endif
> +
>  #  ifdef PERL_IS_MINIPERL
>                               " PERL_IS_MINIPERL"
>  #  endif
> @@ -1996,15 +2005,6 @@ S_Internals_V(pTHX_ CV *cv)
>  #  ifdef PERL_USE_SAFE_PUTENV
>                               " PERL_USE_SAFE_PUTENV"
>  #  endif
> -#  ifdef PERL_HASH_FUNC_DEFINE
> -/* note that this is different from the others, PERL_HASH_FUNC_DEFINE
> - * is a string which says which define was defined. */
> -                             " " PERL_HASH_FUNC_DEFINE
> -#  endif
> -#  ifdef PERL_HASH_USE_SBOX32
> -                             " PERL_HASH_USE_SBOX32"
> -#  endif
> -
>  #  ifdef PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES
>                               " PERL_USE_UNSHARED_KEYS_IN_LARGE_HASHES"
>  #  endif
>
>
>
If you check the previous patch was exactly that. I dont think that is the
right fix for this. Now I understand what is going on I can remove YET MORE
technical debt. This is all essentially brain dead.

Yves

-- 

demerphq:
@bram-perl thanks! I think I managed to mess this up! Fixing now. Beer++

demerphq:
@bram-perl thanks to you I have fixed it thanks. Also I have moved the defines out of the "non bincompat list". Since the hash function isn't linkable and will be compiled into certain XS code we have to assume that changes in these defines can break binary compatibility.  I think we should think about whether that really makes sense, I dont see why it does.

demerphq:
So unless someone objects ill merge this soon.

Leont (approved):

demerphq:
Thanks @Leont . Merging.

PR #20059, opened 2022-08-07
Locale fallback cleanup

bram-perl:
Cleanup some code in locale.c: `Perl_init_i18nl10n` which sets/prints the fallback locale that was used.

Note: this does alter the output of the message slightly, if this is not wanted then that can easily be patched.

Old output:
```
	$ LC_ALL= LANG=C LC_PAPER=aa_BB perl -e1
	(...)
	perl: warning: Falling back to the standard locale ("C").
```

New output:
```
	$ LC_ALL= LANG=C LC_PAPER=aa_BB ./perl -e1
	(...)
	perl: warning: Falling back to a fallback locale (LANG) ("C").
```

I.e. the output now includes the ENV variable that contains the fallback locale (in the output above: `LANG`)

(Suggestions for a better name for `trial_locales_struct` are also welcome)

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -3535,7 +3535,8 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
         const char* fallback_desc;
         const char* fallback_name;
     } trial_locales_struct;
-    struct trial_locales_struct trial_locales[5]; /* 5 = 1 each for "", LC_ALL, LANG, "", C */
+    /* 5 = 1 each for "", LC_ALL, LANG, (Win32) system default locale, C */
+    struct trial_locales_struct trial_locales[5];
     unsigned int trial_locales_count;
     const char * const lc_all     = PerlEnv_getenv("LC_ALL");
     const char * const lang       = PerlEnv_getenv("LANG");
@@ -3559,18 +3560,6 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
     /* current locale for given category; should have been copied so aren't
      * volatile */
     const char * curlocales[NOMINAL_LC_ALL_INDEX + 1];
-
-#  ifdef WIN32
-
-    /* In some systems you can find out the system default locale
-     * and use that as the fallback locale. */
-#    define SYSTEM_DEFAULT_LOCALE
-#  endif
-#  ifdef SYSTEM_DEFAULT_LOCALE
-
-    const char *system_default_locale = NULL;
-
-#  endif
 
 #  ifndef DEBUGGING
 #    define DEBUG_LOCALE_INIT(a,b,c)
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -3514,6 +3518,8 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
                     }
                 }
 
+                trial_locales[i].fallback_desc = "the system default locale";
+                trial_loalces[i].fallback_name = system_default_locale;
```
khwilliamson: typo

locale.c
```diff
@@ -3701,14 +3653,33 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
...
-                .fallback_desc = "the system default locale",
-                .fallback_name = "", /* overwritten at the top of the loop */
-            };
+             * fallback possibility). */
+            {
+                const char *system_default_locale = NULL;
+
+                /* Note that this may change the locale, but we are going to do
+                 * that anyway */
+                system_default_locale = setlocale_c(LC_ALL, "");
```
khwilliamson: As I read this code, I think it would be clearer if the declaration and initialization were combined.

khwilliamson (changes requested):
In the first commit message, I think the XXX name is fine
In the second commit message, 'tries' should be 'trial'

khwilliamson:
otherwise LGTM

locale.c
bram-perl: (Fixed)

Good catch; I had only tested the final version on windows but I've now tested the intermediate version as well.

locale.c
bram-perl: Code updated

locale.c
```diff
@@ -3734,14 +3747,14 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
...
-            if (strEQ(trial_locales[i], "C")) {
+            if (strEQ(trial_locales[i].trial_locale, "C")) {
                 description = "the standard locale";
                 name = "C";
             }
 
 #  ifdef SYSTEM_DEFAULT_LOCALE
 
-            else if (strEQ(trial_locales[i], "")) {
+            else if (strEQ(trial_locales[i].trial_loalce, "")) {
```
khwilliamson: I earlier missed this typo

khwilliamson:
I'm happy to merge this, but will wait if @bram-perl  wants to wait for other feedback.  

(My experience is that this such feedback is unlikely for this area of the code)

locale.c
bram-perl: Fixed

(And lesson learned.. I should've tested every commit on windows instead of just the last one :( )

bram-perl:
Some feedback on the change of the warning message is still welcome/needed..
That is: is the change acceptable? Or is it a bad idea and should the old message be preserved?

There are two changes in ouptut:

* example 1:
    * old: `perl: warning: Falling back to the standard locale ("C").`
    * new: `perl: warning: Falling back to a fallback locale (LANG) ("C").`
* example 2:
    * old: `perl: warning: Falling back to a fallback locale ("en_US").`
    * new: `perl: warning: Falling back to a fallback locale (LANG) ("en_US").`

jkeenan:
> Some feedback on the change of the warning message is still welcome/needed.. That is: is the change acceptable? Or is it a bad idea and should the old message be preserved?
> 
> There are two changes in ouptut:
> 
>     * example 1:
>       
>       * old: `perl: warning: Falling back to the standard locale ("C").`
>       * new: `perl: warning: Falling back to a fallback locale (LANG) ("C").`
> 
>     * example 2:
>       
>       * old: `perl: warning: Falling back to a fallback locale ("en_US").`
>       * new: `perl: warning: Falling back to a fallback locale (LANG) ("en_US").`

The proposed new messages do seem awkward, as they contain redundant words.  If you're "falling back" from some (presumably intended) locale to another locale, that second locale is a *fallback* locale by definition.

Based on past experience, changes in warnings or error messages should always be done cautiously, as people tend to write tests to detect them.

bram-perl:
> The proposed new messages do seem awkward, as they contain redundant words. If you're "falling back" from some (presumably intended) locale to another locale, that second locale is a _fallback_ locale by definition.

That wording has been in place since 2014.
Sure it can be changed but that would change the message even more.

> Based on past experience, changes in warnings or error messages should always be done cautiously, as people tend to write tests to detect them.

This message is shown when perl is started, not when some code is executed. (it's also completely unrelated to `use warnings;`, `-w`, ...)
It is possible that people wrote tests for this but  I think that is less likely.. (I could be wrong ofc)
To test it they would need to do something like:

    my $foo = qx/perl -e1 2>&1/;
    if ($foo =~ m/......./) {
        ....
    }

bram-perl:
I've updated the output so it behaves exactly the same as before;
So there should be no visible change from these commits.

khwilliamson (approved):
I'll push whenever you've decided there's been a long enough time for suggested names, and change the first commit's message to not have the XXX

bram-perl:
> I'll push whenever you've decided there's been a long enough time for suggested names, and change the first commit's message to not have the XXX

Done

PR #20063, opened 2022-08-07
OP_AELEMFASTLEX_STORE - combined sassign/aelemfast_lex

richardleach:
This commit introduces a new OP to replace simple cases of OP_SASSIGN
and OP_AELEMFAST_LEX. (Similar concept to GH #19943)

For example, `my @ary; $ary[0] = "boo"` is currently implemented as:
```
    7     <2> sassign vKS/2 ->8
    5        <$> const[PV "boo"] s ->6
    -        <1> ex-aelem sKRM*/2 ->7
    6           <0> aelemfast_lex[@ary:1,2] sRM ->7
    -           <0> ex-const s ->-
```

But now will be turned into:
```
    6     <1> aelemfastlex_store[@ary:1,2] vKS ->7
    5        <$> const(PV "boo") s ->6
    -        <1> ex-aelem sKRM*/2 ->6
    -           <0> ex-aelemfast_lex sRM ->6
    -           <0> ex-const s ->-
```

This is intended to be a transparent performance optimization.
It should be applicable for RHS optrees of varying complexity.

Performance:
The following unrealistic code shows the op-related speedup:
```
my (@l0, @l1, @l2, @l3, @l4, @l5, @l6, @l7, @l8, @l9);
for (0..10_000_000) {
    $l0[0] = $l1[1] = $l2[2] = $l3[3] = $l4[4] = $l5[5] = $l6[6] = $l7[7] = $l8[8] = $l9[9] = 1 
}
```
Blead:
```
          1,069.74 msec task-clock                #    0.999 CPUs utilized          
                15      context-switches          #    0.014 K/sec                  
               200      page-faults               #    0.187 K/sec                  
     4,746,898,970      cycles                    #    4.437 GHz                    
        91,116,259      stalled-cycles-frontend   #    1.92% frontend cycles idle   
             1,534      stalled-cycles-backend    #    0.00% backend cycles idle    
    17,304,048,596      instructions              #    3.65  insn per cycle         
                                                  #    0.01  stalled cycles per insn
     3,350,806,733      branches                  # 3132.369 M/sec  
```  
Patched:
```
            713.06 msec task-clock                #    0.999 CPUs utilized          
                10      context-switches          #    0.014 K/sec                  
               200      page-faults               #    0.280 K/sec                  
     3,153,196,578      cycles                    #    4.422 GHz                    
        24,931,297      stalled-cycles-frontend   #    0.79% frontend cycles idle   
                50      stalled-cycles-backend    #    0.00% backend cycles idle    
    12,603,789,338      instructions              #    4.00  insn per cycle         
     2,550,753,544      branches                  # 3577.208 M/sec   
```

Note: Name is a bit unwieldy, suggestions welcome. (@leonerd - I think this PR is something different to what you had in mind as OP_PADAV_STORE?)


richardleach:
Rebased after #20077 merge.

pp_hot.c
```diff
@@ -169,6 +169,57 @@ PP(pp_padsv_store)
...
+    /* !OPf_STACKED is not handled by this OP */
+    assert(op->op_flags & OPf_STACKED);
+    assert(key >= 0);
+
+    /* Inlined, simplified pp_aelemfast here */
+    assert(SvTYPE(av) == SVt_PVAV);
+    SV** svp;
+
+    /* inlined av_fetch() for simple cases ... */
+    if (!SvRMAGICAL(av) && key >= 0 && key <= AvFILLp(av)) {
```
leonerd: Does this condition need to test `key >= 0`? You already `assert()`ed on that earlier, and rpeep only rewrote it to this opcode if so.

leonerd (commented):
Overall direction looks good. One small point

leonerd:
> (@leonerd - I think this PR is something different to what you had in mind as OP_PADAV_STORE?)

Indeed. For `OP_PADAV_STORE` I was thinking more to optimise code like `my @arr; @arr = (1,2,3);` by rewriting the entire `OP_AASSIGN` into `OP_PADAV` - analogous to the SASSIGN/PADSV case you already did.

pp_hot.c
richardleach: Oh yeah, good point. I'll remove that test.

richardleach:
This shouldn't be merged until 20114 is fixed for Memoize::Once; possibly the peephole code in this PR should be amended to handle the RHS actually being the `op_other` of e.g. an `ORASSIGN`.

richardleach:
It doesn't seem like any change akin to https://github.com/Perl/perl5/commit/c5bbf81a67393c24727402cfeca0bfe002966ee0 is needed here for Memoize::Once.

I think this is ready to merge.

demerphq:
@richardleach you can merge yourself no?

richardleach:
> @richardleach you can merge yourself no?

Yes, I'll do it when I get a moment, assuming no-one shouts "Nooo, don't doo iiitt" between now and then.

iabyn:
On Wed, Sep 07, 2022 at 08:28:03AM -0700, Richard Leach wrote:
> > @richardleach you can merge yourself no?
> 
> Yes, I'll do it when I get a moment, assuming no-one shouts "Nooo, don't doo iiitt" between now and then.
> 

Not so much a shout, as a delayed clearing of the throat.

Two points.

1) I don't see why you've skipped optimising negative key values. I added
support for negative keys to AELEMFAST a while back, on the grounds that
literal keys like $a[-1] etc were a lot more likely to appear in code than
$a[255]. The only change needed in pp_aelemfastlex_store() to support it
is, like in pp_aelemfast(), to do a check for key >= 0 and skip the short
cut if it's not (meaning the array fetch will be done by av_fetch()
instead of by directly accessing AvARRAY[key]). But you still get the
benefit of a single op rather than 2 ops with intermediate pushing and
popping of values.

2) I think the code which tests for and emits the "Useless assignment to a
temporary" is not needed. This condition could (although generally
shouldn't) happen when the LHS of an assignment is an arbitrary
expression, but when the LHS is the result of an array fetch, it can never
be an SvTEMP with a refcount of 1 unless something has gone horribly wrong.
Maybe replace that code with an assert that's not SvTEMP with a refcount
of 1.

Other than that, it looks good!

-- 
"There's something wrong with our bloody ships today, Chatfield."
    -- Admiral Beatty at the Battle of Jutland, 31st May 1916.

richardleach:
Thanks for the feedback, @iabyn. On:

1) Yep, that's true. I probably over-simplified, thinking that negative values would still go the slow route through av_fetch.
2) Aha!

I'll prepare a PR that addresses both suggestions.

richardleach:
Not sure if I interpreted point 2 correctly, but `assert(!(SvTEMP(targ) && SvREFCNT(targ) == 1)` fails in multiple places in the test suite.

richardleach:
`assert(!(SvTEMP(targ) && SvREFCNT(targ) == 1 && !SvSMAGICAL(targ)));` doesn't get tripped by any tests, but I'm still unsure of whether or not it's safe to exchange the full `if()` condition for it. Guidance welcome!

iabyn:
On Wed, Sep 21, 2022 at 04:23:09PM -0700, Richard Leach wrote:
> `assert(!(SvTEMP(targ) && SvREFCNT(targ) == 1 && !SvSMAGICAL(targ)));` doesn't get tripped by any tests, but I'm still unsure of whether or not it's safe to exchange the full `if()` condition for it. Guidance welcome!

Yeah, I forgot about tied arrays. With those, av_fetch() can return a
magical temp value which when modified, triggers a call to STORE(). But 
that type of temp wouldn't trigger a warning. So asserting

    ! (SvTEMP(targ) && SvREFCNT(targ) == 1 && !SvMAGICAL(av))

should be ok. With a comment which explains that the assert is checking
that a "Useless assignment to a temporary" warning should never need to be
emitted here.

-- 

richardleach:
Thanks. I've updated https://github.com/Perl/perl5/pull/20331 accordingly.

PR #20068, opened 2022-08-08
Make PerlIO::get_layers accept IO references #20058

### LINKED ISSUE #20058: PerlIO::get_layers doesn't handle iorefs
Module:
PerlIO
**Description**

PerlIO::get_layers returns an empty list when called with ioref.
**Steps to Reproduce**

`perl -mPerlIO -E 'my @ls = PerlIO::get_layers(*STDOUT{IO}); say("[@ls]")'`
**Expected behavior**

Return the list of layers.

Leont:
`PerlIO::get_layers` used to not work with IO references, now it does.

```
perl -MPerlIO -E 'my @ls = PerlIO::get_layers(*STDOUT{IO}); say("[@ls]")'
```

This fixes #20058

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -381,6 +381,15 @@ assigning to an C<:lvalue> subroutine.
 XXX Important bug fixes in the core language are summarized here.  Bug fixes in
 files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
+=over 4
+
+=item * PerlIO::get_layers will now accept IO references too
+
+Previously it would only take glob references or names of globs. Now it will
+also accept IO references.
+
+=back
+
 [ List each fix as an =item entry ]
 
 =over 4
--- a/universal.c
+++ b/universal.c
@@ -732,7 +732,7 @@ XS(XS_PerlIO_get_layers)
     {
         SV *	sv;
         GV *	gv;
-        IO *	io;
+        IO *	io = NULL;
         bool	input = TRUE;
         bool	details = FALSE;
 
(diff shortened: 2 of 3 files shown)


universal.c
```diff
@@ -775,12 +775,15 @@ XS(XS_PerlIO_get_layers)
         }
 
         sv = POPs;
-        gv = MAYBE_DEREF_GV(sv);
 
-        if (!gv && !SvROK(sv))
-            gv = gv_fetchsv_nomg(sv, 0, SVt_PVIO);
+        if (SvROK(sv) && SvTYPE(SvRV(sv)) == SVt_PVIO)
+            io = (IO*)SvRV(sv);
```
tonycoz: `MAYBE_DEREF_GV()` does `SvGETMAGIC()` on sv, but now we're checking flags on sv without calling `SvGETMAGIC()`

pod/perldelta.pod
```diff
@@ -372,6 +372,15 @@ assigning to an C<:lvalue> subroutine.
 XXX Important bug fixes in the core language are summarized here.  Bug fixes in
 files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
+=over 4
+
+=item * PerlIO::get_layers will now accept IO references too
+
+Previously it would only take glob references or names of globs. Now it will
+also take accept IO references.
```
tonycoz: s/also take accept IO/also accept IO/ ?

universal.c
Leont: Good point, I guess I should swap the first two options around

tonycoz: That should work.

tonycoz (approved):

Issue #20010, opened 2022-07-28
remove username from repo


ghost:
I want to have my name, e-mail address and username, removed from the repo.

affected files

.mailmap
AUTHORS
Porting/checkAUTHORS.pl

Would a PR removing the relevant lines suffice, or if any of the above files are auto generated, and would be recreated, are there special instructions to remove the relevant lines?

If special instructions are indicated please provide these.

Leont:
Given two of those are generated from git history, it's currently not possibly. But I guess it could be made possible (by using a hash of some sort to filter you out).

@demerphq ideas?

ghost:
I neglected to mention the file pod/perl5340delta.pod.

ghost:
Presently there are no commits to the repo, consequently no git history, associated with this username.

`https://github.com/Perl/perl5/commits?author=`

Perhaps a Contributor would oblige and remove these lines, which would not result in a git history associated with this username.

rwp0:
@apparluk Sorry, but why would you like to remove your name from the authors?

Leont:
> Perhaps a Contributor would oblige and remove these lines, which would not result in a git history associated with this username.

Like I said before, `.mailmap` and `AUTHORS` are generated files, so there's no point to manually removing them: they'd be added back on regeneration.

`Porting/checkAUTHORS.pl` is the source of the link between the two addresses, I'm not sure we can remove that without things breaking. I suspect we need to block your email first if we don't want to break tests.

The `perl5340delta` entry can be fixed easily enough though.

ghost:
Update nmake for VS 2017  
committed on Jun 14, 2017

I don't think there's any code I wrote in the repo since then.

Leont:
b998bd1a4a2c4e35a119ea73c7650a26e0060950

ghost:
If removal of the content it is too much trouble, I could just use a different GitHub account and e-mail address.

I presume in the foregoing that a merged commit - using a different GitHub account and e-mail address - would overwrite the current information in .mailmap, AUTHORS and Porting/checkAUTHORS.pl.

Leont:
> If removal of the content it is too much trouble, I could just use a different GitHub account and e-mail address.

I'm sure it can be done, but it requires code and someone will need to write that first.

> I presume in the foregoing that a merged commit - using a different GitHub account and e-mail address - would overwrite the current information in .mailmap, AUTHORS and Porting/checkAUTHORS.pl.

[...]

Issue #19867, opened 2022-06-16
Need to skip AUTHOR testing for github CI sanity


khwilliamson:
A first time author will not have their workflow run on GH without approval.  If someone does approve it, it will automatically fail because the person isn't in the AUTHORS file.

So their patch can't be tested without them being an accepted author.  This doesn't make sense.  The obvious solution is to skip AUTHOR testing in such cases.

tonycoz:
One option might be to modify `authors.t` to skip on a pull_request event.

Note that if they do test the change locally they'll see the test failure anyway, and they'll see the failure on push to their fork.

haarg:
Wouldn't this also result in blead failing CI if a PR was merged without adding the author?

Leont:
Yeah what tony and haarg said

khwilliamson:
Can we make it so that failing any porting test doesn't stop the rest of the tests from running?  Yes we know that this has to be fixed before merging, but at least we'd find out soon what else might be wrong.

Just now we had a failure because a version didn't get bumped, and the rest of the tests didn't run.

haarg:
That definitely seems worth changing. The version and author tests shouldn't prevent the rest of the tests from running. Ideally, they also would only show up as failures on a single action, and not any of the later checks.

bram-perl:
> The version and author tests shouldn't prevent the rest of the tests from running. Ideally, they also would only show up as failures on a single action, and not any of the later checks.

In the PR I just created (#20123)  that should already be the case for the version check: it needs the full history and that's only available during the 'sanity check'.
The author tests failure will (in the current state) still show up in each build configuration but changes are being made to it which may/will result in it needing full history (which means they can only run during 'sanity check' since only that has the full history)

bram-perl:
Only running the authors test in the sanity check run is done in PR #20124

demerphq:
On Sat, 20 Aug 2022 at 17:46, Bram ***@***.***> wrote:
>
> Only running the authors test in the sanity check run is done in PR #20124

I dont get this. Am I missing something? At least in my branch the
authors test runs fine in all the tests.  Why do we have to restrict
it to the sanity check when there is a full clone? As far as I
understand it checks the commits it has for new authors, if it has
less authors to check it shouldnt matter.

Yves

demerphq:
Bram has explained what I was missing: sometimes we end up with a
merge commit and sometimes we dont.

[...]

PR #20070, opened 2022-08-08
Replace checkAUTHORS.pl with updateAUTHORS.pl/updateAUTHORS.pm, allow users to be excluded

demerphq:
In #20010 a user has asked that we do not list their name or email any of our latest files in the repo. 

This was a bit problematic as updateAUTHORS.pl would just add the user back if we did so. I added functionality to updateAUTHORS.pl to support us using hashed contributor details as a way to prevent users from being added by our automation. Unfortunately this then meant that checkAUTHORS.pl and updateAUTHORS.pl had different idea about what should be in the files. There are other issues, the checkAUTHORS.pl file contains a lot of data in a hard to reuse or manage form, and in plaintext, (although slightly munged so its not a direct scraper target).  In the end it made a lot more sense to me to simply drop checkAUTHORS.pl and replace it with updateAUTHORS.pl for all use cases.

This patch sequence adds support for a bunch of features that checkAUTHORS.pl had that updateAUTHOR.pl did not, almost uniformly related to reporting so that the tool is now nearly feature equivalent with checkAUTHORS.pl.

There are a few differences:
1. The old tool  supported saving the commit history to a file and then piping into the checkAUTHORS.pl, this mode is no longer supported, although if we need it we can bring it back, but with a certain amount of bodge. The reason for the bodge is that updateAUTHORS.pm expects its log data in a very particular format that is not supported out of the box by git log. You have to construct a long and awkward format string to provide for log to use, so it is not quite so simple to pipe the results to a file. If we really need it we can figure out a solution. An upside of this is that the tool doesn't expect filenames as arguments, instead it expects a get ref spec just like git log would do so it shouldnt really be needed to pipe the results.
2. The old tool supported querying github. updateAUTHORS.pl doesnt support this as far as I remember. I dont think we were using it anyway. We can teach updateAUTHORS.pl to do the same thing if someone remembers why were doing this.
3. The old tools supported a couple of options that IMO no longer make sense. Specifically the  "--missing" option is redundant. --validate/--tap does the same thing.
4. The old tool had some (out of date) information on committers which I have abandonded. We currently do not track this. I was thinking if we want to track it we should add a * to peoples AUTHOR data and then add a footnote explaining they are committers. Then we can use the AUTHORS file as the primary source of this data. 

This PR adds two new files
1. Porting/updateAUTHORS.pm which is the brains of the updateAUTHORS.pl tool separated from the command line part of the two so that the POD for using the tool could be separated from the pod of how the tools works. 
2. Porting/exclude_contrib.txt. This contains a list of SHA-256 in base64 of the email details of contributors (commit author/email) that we should not include in .mailmap or AUTHORS.

This PR removes two files:
1. Porting/checkAUTHORS.pl - replaced by Porting/updateAUTHORS.pl
2. t/porting/pending-author.t - I couldnt figure out what this was for, and it doesnt seem to be required with the updateAUTHORS.pl workflow. If someone can explain what it was used for better than the test file itself then I would have no problem adding it back.

This PR removes the personal details as requested in #20010 

The latest version of updateAUTHORS.pl has support for additional reports, and I expect as I have time I will add more. IMO it is much easier to do this with updateAUTHORS.pl than it was with checkAUTHORS.pl (because of the reliance on the git standard log formats). I plan to add a "who-works-on-what" to show file data by author. Eg, it might show that @khwilliamson, myself and @hvds do most of the changes to the regex engine.

I have marked this PR as draft for now, as i expect to make changes as I get feedback, but the patch it pretty stable, review appreciated!


Leont:
> t/porting/pending-author.t - I couldnt figure out what this was for, and it doesnt seem to be required with the updateAUTHORS.pl workflow. If someone can explain what it was used for better than the test file itself then I would have no problem adding it back.

If there are pending changes, it checks if the current author is listed as author. This is to prevent the test-suite from succeeding before making one's first commit but failing after it.

demerphq:
On Mon, 8 Aug 2022 at 23:10, Leon Timmermans ***@***.***>
wrote:

> t/porting/pending-author.t - I couldnt figure out what this was for, and
> it doesnt seem to be required with the updateAUTHORS.pl workflow. If
> someone can explain what it was used for better than the test file itself
> then I would have no problem adding it back.
>
> If there are pending changes, it checks if the current author is listed as
> author.
>
What is a pending change?

Yves

-- 

demerphq:
On Tue, 9 Aug 2022 at 09:09, demerphq ***@***.***> wrote:
>
> On Mon, 8 Aug 2022 at 23:10, Leon Timmermans ***@***.***> wrote:
>>
>> t/porting/pending-author.t - I couldnt figure out what this was for, and it doesnt seem to be required with the updateAUTHORS.pl workflow. If someone can explain what it was used for better than the test file itself then I would have no problem adding it back.
>>
>> If there are pending changes, it checks if the current author is listed as author.
>
> What is a pending change?

To expand on my failure to understand this concept, authors.t checks
that the AUTHORS file is up to date. That means it checks that if you
ran updateAUTHORS.pl nothing would change given the commits that are
in the repo when it runs.

When pending-author.t runs what does it check?

Our tests run in two scenarios relevant to this discussion, on peoples
local box while developing and on CI when the code is pushed.

When this executes on my local box what is it testing that authors.t
doesnt test? When this runs on CI when the code has been pushed what
is it testing that authors.t doesnt test?

I cannot work out a reasonable understanding of what a "pending
change" is in this context. Either the commit is in the commit log and
it is being tested or it is not. If it is then it gets tested by
authors.t, if the commit isn't in the commit log then it isn't pending
it is non-existent as far our testing goes.

cheers,
Yves

-- 

Leont:
> What is a pending change?

Uncommited changes

demerphq:
@Leont  - Ok, when I first read pending-author i think i misunderstood what it does and got myself in a muddle, the term "pending" really threw me. I should have just ignored that and looked at the code more closely. 

So it is looking to see if there are uncommitted changes that if they *were* committed with the current users details would result in a test failure. I dont see why we need another test file for this, we can just put it into the authors.t testing. I will add that.

demerphq:
Ok, authors.t now does the equivalent of what pending authors used to do, arguably better as it now checks (nearly) all the things that git would. I also added copious verbiage if the tests fail.

Porting/updateAUTHORS.pl
```diff
@@ -1,661 +1,639 @@
...
-        chomp $line;
-        push @authors_preamble, $line;
-        if ($line =~ /^--/) {
-            last;
+            my $author_known= $self->known_contributor($author_name,$author_email);
+            my $committer_known= $self->known_contributor($committer_name,$committer_email);
+            if (is( $author_known && $committer_known, 1,
+                "Uncommitted changes are by a known contributor?"))
+            {
+                diag "Testing uncommtted changes! Remember to commit before you push!"
```
Leont: I don't think it is helpful for this to `diag` at all.

bram-perl:
Some early feedback without really looking at the code yet:

* commit `Porting/updateAUTHORS.pl - add support for excluding contributors`:
    * based on the commit description I got a feeling that the commit tries to do too much at once (i.e. consider splitting the commit?)
    * `Porting/checkAUTHORS.pl which does not do the right thing these days
anyway.`: Is it possible/sensible to include a small description of how/when it doesn't do the right thing?
*  commit `remove redundant data from .mailmap`:
    * can you describe in the commit message what you consider redundant data?
    * and the reason why this matters?  [I'm not disputing that there - probably - is a good reason for it; but I like seeing it in the commit message]

demerphq:
On Thu, 11 Aug 2022 at 12:56, Bram ***@***.***> wrote:
>
> Some early feedback without really looking at the code yet:
>
> commit Porting/updateAUTHORS.pl - add support for excluding contributors:
>
> based on the commit description I got a feeling that the commit tries to do too much at once (i.e. consider splitting the commit?)

Ok.

> Porting/checkAUTHORS.pl which does not do the right thing these days anyway.: Is it possible/sensible to include a small description of how/when it doesn't do the right thing?

Sure. I think it was that it doesn't understand the new
Porting/exclude_contrib.txt but ill update the commit.

> commit remove redundant data from .mailmap:
>
> can you describe in the commit message what you consider redundant data?

When the LHS and RHS of a .mailmap line are the same.

> and the reason why this matters? [I'm not disputing that there - probably - is a good reason for it; but I like seeing it in the commit message]

Actually this is a controversial point. Some folks seem to prefer us
not having it in the .mailmap. It also makes the mailmap larger.

I personally dont think it matters and think that removing the entries
results in us losing capabilities, specifically the ability to
recognize when someone has manually updated the AUTHORS file to change
their preferred email or their name. I only implemented it because it
came up multiple times and I got tired with discussing it. To
compensate I added options to change name via the command line.

Yves

-- 

Porting/updateAUTHORS.pl
```diff
@@ -613,32 +646,39 @@ =head1 FUNCTIONS
       $other_name, $other_email,
       $line_num ]
 
-=item read_and_update($authors_file, $mailmap_file)
+=item new()
```
bram-perl: *commit Porting/updateAUTHORS.pl - refactor to be OO*:

`sub new` is missing in this commit
It appears it was added in a later commit ("Porting/updateAUTHORS.pl - add support for excluding contributors") but the basic/initial version should be in this commit (or at least that's what I would assume)

Also: should this be:
```
=item new($authors_file, $mailmap_file, $exclude_file)
```
since these are the required parameters? (at least based on the POD)

Porting/exclude_contrib.txt
```diff
@@ -0,0 +1,13 @@
++Qhq3JfXxVLUkhm5ywYu11wxK21xoqELiMJXYr9y3L4
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

A header in this file describing it's content/purpose would be nice..
A complete example would be nice too, what I mean by that: show the plain text entry that is excluded and the base64_sha256 string.

Also: it starts out with a bunch of info already and from the commit it's difficult to tell which items it is excluding (i.e. to what plain text this maps)

(Personally I would prefer if it started empty/with a header and then contents added in it in a next commit so the link between the data and what it excludes is a bit more obvious in history)

Porting/updateAUTHORS.pl
```diff
@@ -68,6 +69,32 @@ sub _make_name_simple {
...
+sub _register_author {
+    my ($self, $name, $type)= @_;
+    my $digest= $self->_keeper_digest($name)
+        or return;
+
+    $self->{author_info}{"lines"}{$name}++
+        and return;
+    my $munged= _clean_name($name);
+    if (my $h= $self->{add_to_exclude_file}) {
+        $h->{$digest} //= $munged;
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

It's not (yet) obvious to me what this code does :/
Digging deeper into it I think I get what it does but I still find it a bit confusing:

* it first adds it into the `$self->{author_info}` which will later be used to create the AUTHORs file (or at least that's what I think/assume)
* if the `add_to_exclude_file` option is set (to a hash) then the digest is also added in it and that will (somehow) be used later to exclude the info

So the first part: add it to AUTHORs, the second part: exclude it from AUTHORs.. I guess it's sort of works but at first glance it looks like a contradiction..
(And in this particular commit ("Porting/updateAUTHORS.pl - add support for excluding contributors ")    the second part - excluding from AUTHORs - doesn't appear to happen? In `update_authors` it just prints everyting that is in `$self->{author_info}`)

Also somewhat related: the `add_to_exclude_file` option doesn't appear to be documented?

Porting/updateAUTHORS.pl
```diff
@@ -507,17 +593,93 @@ sub read_and_update {
...
+    my ($self, $name)= @_;
+    return $self->{changed_file}{$name};
+}
+
+sub unchanged_file {
+    my ($self, $name)= @_;
+    return $self->changed_file($name) ? 0 : 1;
+}
+
+sub new {
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

(Part of this `sub new` belongs in the previous commit)

Porting/updateAUTHORS.pl
```diff
@@ -526,14 +688,64 @@ sub main {
...
     pod2usage(-verbose => 2) if $show_man;
 
-    read_and_update($authors_file, $mailmap_file);
+    my $self= __PACKAGE__->new(
+        authors_file => $authors_file,
+        mailmap_file => $mailmap_file,
+        exclude_file => $exclude_file,
+        add_to_exclude_file => ($exclude ? {} : undef),
+        validate     => $validate,
+        verbose      => $verbose,
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

Options `add_to_exclude_file`, `validate`, `verbose` missing in the POD for `new()`

Porting/updateAUTHORS.pl
```diff
@@ -526,14 +688,64 @@ sub main {
...
+    );
+    my $changed= $self->read_and_update();
+    if ($validate) {
+        eval <<'EOF_TEST'
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

code was adapted in later commits and eval removed in "t/porting/authors.t - detect if committing uncommitted changes might" 
but personally I prefer  `eval { ... }` over `eval '...'`

t/porting/authors.t
```diff
@@ -16,31 +16,7 @@ skip_all(
...
-    # we should be on a merge commit, but double check
-    my $null = File::Spec->devnull;
-    my $branch_head = `git rev-parse "HEAD^2" 2>$null`;
-    chomp $branch_head;
-
-    # gives the history of the branch being merged, excluding what it is
-    # merged into
-    $revision_range = '"HEAD^1..HEAD^2"'
-        if $branch_head;
-}
```
bram-perl: *commit Porting/updateAUTHORS.pl - add support for excluding contributors*:

It's unclear to me why this code was removed and/or if it is OK to remove it..
Explaining why it's ok to remove it/why it was removed (in the commit message) might be helpful

(I also have no idea what problem this code tried to solve, would need to check the history to see why it was added)

AUTHORS
```diff
@@ -12,10 +12,42 @@
 # is strictly forbidden.  (Passive distribution with the Perl source
 # code kit is, of course, allowed.)
 #
-# This should contain the preferred addresses.  Alternate addresses are in
-# Porting/checkAUTHORS.pl.
+# This file is one of three files that we use to keep track of our
+# contributors, and is coupled with C<.mailmap> and
+# C<Porting/exclude_contrib.txt> in being used to manage our contributor
```
bram-perl: *commit Improved header comment in AUTHORS file*:

The `in being used` this sentence reads a bit strange (but I'm not native so it might just be me)

AUTHORS
```diff
@@ -12,10 +12,42 @@
...
+# their visible presence on GitHub and casual git operations shows as
+# indicated in this file. If the person is listed WITHOUT an email
+# address then the email in git tool will be shown as unknown. This may
+# be used because contributors choose to withdraw from the project and
+# are unreachable but still deserve recognition, or even possibly
+# because they are deceased.
+#
+# Every entry in this file should also appear on the left hand side of
+# at least one entry in F<.mailmap> (this may change), and if you
+# modify a persons name or email (but not both together)then
```
bram-perl: *commit Improved header comment in AUTHORS file*:

missing space between `)` and `then`

Porting/updateAUTHORS.pl
```diff
@@ -1,939 +1,438 @@
 #!/usr/bin/env perl
-package Porting::updateAUTHORS;
```
bram-perl, line 2: *commit checkAUTHORS.pl -> updateAUTHORS.pl in all cases*:

This commit "checkAUTHORS.pl -> updateAUTHORS.pl in all cases " is a bit misleading and I personally would prefer if it was split.

It moves code from Porting/updateAUTHORS.pl to Porting/updateAUTHORS.pm but that's not really mentioned in the commit message; IMO the move from updateAUTHORS.pm to updateAUTHORS.pm deserves it's own commit so that reviewing/viewing the actual changes becomes much easier to see (and review).

(Small note: I did not really review the "checkAUTHORS.pl -> updateAUTHORS.pl in all cases " commit (yet) because it's a bit more difficult to review to the code move)

Porting/updateAUTHORS.pm
```diff
@@ -158,7 +158,10 @@ sub read_commit_log {
 
     my $last_commit_info;
     my $cmd= qq(git log --pretty='format:$tformat' $numstat $commit_range);
-    open my $fh, "$cmd |";
+    $cmd=~s/'/"/g if $^O=~/Win/;
+    open my $fh, "-|", $cmd
```
bram-perl, line 216: *commit Win32 support*:

consider using mutli-args pipe open? it might avoid the need to mangle with the quotes (but should be tested)
i.e.: (untested)

    open my $fh, "-|", "git", "log", "--pretty=format:$tformat", $numstat, $commit_range;

Porting/updateAUTHORS.pl
```diff
@@ -94,72 +96,178 @@ sub main {
...
-
-            diag "\nFiles need updating! ",
-                 "You probably just need to run\n\n",
-                 "   Porting/updateAUTHORS.pl\n\n",
-                 "and commit the results."
-                 if $self->changed_count;
-            done_testing();
-            1
-EOF_TEST
-        eval $code or warn "Error in code: $@\n", do {
```
bram-perl: *commit t/porting/authors.t - detect if committing uncommitted changes might*:

Removing the `eval` looks unrelated to the " t/porting/authors.t - detect if committing uncommitted changes might break authors test" commit:

(I do support removing the `eval` but might need to be done in another commit to make it easier to see the actual changes needed for "committing uncommitted changes")

Porting/updateAUTHORS.pl
```diff
@@ -94,72 +96,178 @@ sub main {
...
 
+sub uncommitted_changes {
+    my ($author_name,$author_email, $committer_name, $committer_email)= @_;
+    my $props= "";
+    my $extra= "";
+    my @git_env_keys=  map {
+            /^GIT_(AUTHOR|COMMITTER)_(NAME|EMAIL)\z/
+            ? "$_='$ENV{$_}'"
+            : ()
+        } sort keys %ENV;
```
bram-perl: *commit t/porting/authors.t - detect if committing uncommitted changes might*:

hmm... are you sure `GIT_AUTHOR_NAME`, ... are available via `%ENV`?
I would expect them to be set when running a git-hook (i.e. pre-commit) but in other circumstances I wouldn't expect these to be set...
Checking my ENV at least shows none of those vars to be set..

(It appears t/porting/pending-author.t uses `git config --get user.name` etc which is what I would expect)

Porting/updateAUTHORS.pl
```diff
@@ -94,72 +96,178 @@ sub main {
...
     elsif ($self->{show_who}) {
         $self->print_who();
+        return 0;
     }
     return 0;    # 0 for no error - intended for exit();
 }
 
 exit(main()) unless caller;
 
+sub uncommitted_changes {
```
bram-perl: *commit t/porting/authors.t - detect if committing uncommitted changes might*:

I'm guessing this commit ("t/porting/authors.t - detect if committing uncommitted changes might") is/should be coupled with the removal of t/porting/pending-author.t but it isn't..  which makes it difficult to see if the new code behaves the same or not..

Porting/updateAUTHORS.pl
```diff
@@ -94,72 +96,178 @@ sub main {
...
+    my ($author_name,$author_email, $committer_name, $committer_email)= @_;
```
bram-perl: *commit t/porting/authors.t - detect if committing uncommitted changes might*:

Hmm... when I look at the yves/exclude_contrib branch I don't really see this code being called..
So is it ever called?

AUTHORS
```diff
@@ -12,41 +12,9 @@
...
-#
-# Lastly some users have contributed to the project but choose not to be
-# listed here. We respect these requests, and support them. Simply
-# delete the entries from this file and .mailmap and then run
-#
-#   Porting/updateAUTHORS.pl --exclude
-#
-# and new exclude records will be added to F<Porting/exclude_contrib.txt>
-# to prevent the author from being readded by F<Porting/updateAUTHORS.pl>
-# in the future.
```
bram-perl: *commit Add a way to rename authors properly*:

This text was added in commit "Improved header comment in AUTHORS file " of this branch and now removed again;
Are you sure you want to remove it?

(I'm not against the removal but the removed text was more informative)

Porting/updateAUTHORS.pl
```diff
@@ -94,72 +96,178 @@ sub main {
...
         $self->{in_reverse}= !$self->{in_reverse};
         $self->report_stats(
             "file_stats", "lines_delta", "lines_added", "lines_removed",
             "commits"
         );
+        return 0;
     }
     elsif ($self->{show_who}) {
         $self->print_who();
+        return 0;
```
bram-perl, line 224: *commit t/porting/authors.t - detect if committing uncommitted changes might*:

This commit added a bunch of `return 0`; in each elsif block.
Should the final `return 0` at the end of the function be changed into `return 1`? If it reached then it may indicate something is wrong? (I didn't fully check)

(Note: I can't leave a comment on the `return 0` at the end of the function since GitHub then shows the comment in the wrong commit...)

bram-perl (commented):
**Note**: I reviewed this commit per commit and not by final changes..
GitHub UI doesn't really make clear on which commit the comment was added tho :(

Some of the comments apply to code that was later changed, so it might be a bit confusing when looking at the current code..

Porting/updateAUTHORS.pl
demerphq: ok, ill remove it except under TEST_VERBOSE. FWIW, i think it makes sense to point out they aren't necesssarily testing what will be pushed up stream.  But its not a hill i want to die on.

Porting/updateAUTHORS.pl
demerphq: regarding missing new() i fixed that commit in my last patch sequence (not pushed just yet, still refining it). But its true i did not add docs for new(). Ill add that to the todo list.

Porting/exclude_contrib.txt
demerphq: Heh. Another issue  I have already cleaned up in my unpushed cleanup (motivated by your previous comment about too much in one patch). I added a header, although it doesn't contain an example, ill add that to the todo list. But I dont want it to start empty. I removed some heuristics that were used to ignore some specific cases in our history and replaced them with exclude file entries which simplified some of the logic i had to change anyway to use exclusions.

Porting/updateAUTHORS.pl
demerphq: Ok, some good catches here. I cleaned it up a bit later on in the patch sequence so but i also incorporated some of your suggestions.

Porting/updateAUTHORS.pl
demerphq: yeah, sorted

Porting/updateAUTHORS.pl
demerphq: yep, also sorted.

Porting/updateAUTHORS.pl
demerphq: the reason I was evaling was some kind of silly desire not to load Test::More if I didn't use it so eval block wouldnt work. It was dumb doing that, which is why i removed it later in the PR.

t/porting/authors.t
demerphq: when i was rebasing some stuff got out of order and missquashed and i didnt notice. this patch shouldnt be there in the sequence.

AUTHORS
demerphq: ok, noted.

Porting/updateAUTHORS.pl
demerphq: yeah, i restructured that commit significantly this evening.

Porting/updateAUTHORS.pm
demerphq: its working now. :-) i dont really want to mess with it right now as i dont have a windows box to test on and I have to wait for the linux stuff to finish CI first.

Porting/updateAUTHORS.pl
demerphq: The test is meant to test what would happen if you did a commit. If you read the docs for git commit I think you will find that my code is a faithful rendition of how it actually works, leaving the case that git commit accepts those as parameters, and also has fallback to using various host name and similar things which I didnt implement.

```
$ GIT_AUTHOR_NAME="Joe <blogs@com>" git commit -a -m'testing'
[detached HEAD fe96690796] testing
 Author: Joe blogs@com <demerphq@gmail.com>
 1 file changed, 1 insertion(+)
 create mode 100644 Die.pm
```
`user.name` is the way most sane people configure their set up, but it is not what git commit would look at first. The environment variables and commiter/author specific config variables take precedence. Try it. 

```
git config author.name Testing
git commit -a -m'Testing'
```

you will see that the author name was set to Testing even through you already have a user.name that says something else.

Arguably t/porting/pending-author.t does the wrong thing.

Porting/updateAUTHORS.pl
demerphq: yep, ill fix, that squash mess up has caused me a headache.

AUTHORS
demerphq: Once the .mailmap doesnt hold the canoncal data some of that comment became redundant.

Porting/updateAUTHORS.pl
demerphq: i added the returns because each action is supposed to do one thing and return, and the if statement was getting long.  I didnt /intend/ for the final return to be 1, but its a good idea. :-)

Porting/updateAUTHORS.pl
bram-perl: Ok

(Next time consider using `require`(+`import`) instead of `eval`+`use` )

Porting/updateAUTHORS.pl
bram-perl: Right, I wasn't thinking about that case.

What is (or was since it may have been cleaned up already) is that in this particular commit ("t/porting/authors.t - detect if committing uncommitted changes might") : 
* updateAUTHORS.pl : `sub uncommited_changes`: this only (appears) to look at the ENV keys and never at the git config. (But later I noticed that `uncommited_changes` of updateAUTHORS.pl - currently - isn't called)
* updateAUTHORS.pm: `sub current_git_name_email`: this first looks at the ENV keys and then at the git config (which is correct)

demerphq: in the latest code uncommitted_changes() is called when the test fails. I probably should give it a better name, like "error_msg_for_uncommitted_changes". it is provided the data returned by current_author_name_email() and current_committer_name_email(). It only inspects the environment to detect if the environment /might/ have affected the data it received.

Porting/updateAUTHORS.pl
demerphq: the removal of checkAUTHORS is now very late in the sequence and is not part of a commit with any other logic changes. (or should not, i might have missed something i guess)

Porting/updateAUTHORS.pl
demerphq: Ill fix the names. I was tired when i did that. with the latest patch updateAUTHORS.pl calls P::updateAUTHORS->uncommitted_changes() to get a git status, if there is nothing uncommitted then all is well, if there is something uncommitted then it checks the user details, and if those are missing from AUTHORS and etc, then it complains by calling App::Porting::updateAUTHORS::uncommited_changes() to produce diagnostics.

demerphq:
BTW, I am still not entirely satisfied with a couple of aspects of this PR, it might be wise to wait until I say so before anyone wastes their time reviewing code I am going to kill.

demerphq:
Hi @bram-perl I have update the code and include I think all of your feedback. With the amount that I have rebased, squashed, and otherwise munged the sequence I am going to mark all your previous feedback as resolved (as I hope it is) and start fresh. The patch sequence should be much cleaner now with much of the interim stuff squashed out. Please let me know what you find that I overlooked. I have been staring at it so long I have become desensitized. :-)

demerphq:
That last push just removes a function made obsolete by the checkAUTHORS removal.

.mailmap
```diff
@@ -59,6 +59,8 @@
 ########################################################################
 A. C. Yardley <yardley@tanet.net> A. C. Yardley <yardley@tanet.net>
 A. Sinan Unur <nanis@cpan.org> A. Sinan Unur <nanis@cpan.org>
+A. Sinan Unur <nanis@cpan.org> A. Sinan Unur <sinan@unur.com>
```
bram-perl, line 55: *commit  .mailmap - add missing entries using updateAUTHORS.pl*:

Small typo in commit message: "updateAUTHORS.pl cares about .mail and checkAUTHORS.pl" -> should be '.mailmap'

perl.h
```diff
@@ -400,7 +400,7 @@ Now a no-op.
 #    define __attribute__always_inline__      __attribute__((always_inline))
 #  endif
 #endif
-#if defined(HASATTRIBUTE_VISIBILITY) && !defined(_WIN32)
+#if defined(HASATTRIBUTE_VISIBILITY) && !defined(_WIN32) && !defined(__CYGWIN__)
```
bram-perl: *commit perl.h - see if this makes cygwin happy*:

commit is unrelated to this PR so you may want to remove it before merging

Porting/updateAUTHORS.pm
```diff
@@ -100,8 +99,8 @@ sub read_commit_log {
 }
 
 sub read_authors {
-    my ($self, $authors_file)= @_;
-    $authors_file ||= "AUTHORS";
+    my ($self)= @_;
+    my $authors_file= $self->{authors_file};
```
bram-perl, line 315: *commit updateAUTHORS.pm - use object to store state and avoid passing args*:

Small inconsistency:

In the old code `read_authors` uses 'AUTHORS' as a default value for `$authors_file`
In the new code `read_authors` has no default value

In the old code `update_authors` uses 'AUTHORS' as a default value for `$authors_file`
In the new code `update_authors` uses 'AUTHORS' as a default value for `$authors_file`

The same applies for `$mailmap_file` in `read_mailmap` and `update_mailmap`

I would somewhat expect `read_...` and `update_...` to be consistent (i.e. both have a default or neither has a default)

Porting/updateAUTHORS.pm
```diff
@@ -0,0 +1,588 @@
...
+__END__
+
+=head1 NAME
+
+Porting::updateAUTHORS - Library to automatically update AUTHORS and .mailmap based on commit data.
+
+=head1 SYNOPSIS
+
+    my $updater= Porting::updateAUTHORS->new(%opts);
+    $updater->read_and_update();
```
bram-perl: *commit updateAUTHORS.pl - split into module Porting/updateAUTHORS.pm*:

I would probably make that:
```
   my $self= Porting::updateAUTHORS->new(
        authors_file => ....,
        mailmap_file => ....,
    );
```

To make it a bit clearer what opts it expect.

(Note: if you do alter the commit you may wish to also add `use Porting::updateAUTHORS;` in the SYNOPSIS; it's added in a later commit so it's good in the final result)

Porting/updateAUTHORS.pm
```diff
@@ -540,32 +567,32 @@ records with the contents:
...
 mailmap file name. Returns nothing but may modify the AUTHORS file
 or the .mailmap file. Requires that both files are editable.
 
-=item read_commit_log($authors_info, $mailmap_info)
+=item read_commit_log()
 
 Read the commit log and find any new names it contains.
 
-=item read_authors($authors_file)
+=item read_authors()
```
bram-perl: *commit updateAUTHORS.pm - use object to store state and avoid passing args*:

For `read_authors` and `read_mailmap`: consider updating the description of the methods;
For `read_authors` current description is:

    Read the AUTHORS file and return data about it.

The description is not wrong per-se but it also updates the internal object state.

(If *I* read that 'returns data' then I *may* start wondering if I need to do something with that data and/or if I need to include it in later method calls)

Porting/updateAUTHORS.pl
```diff
@@ -15,30 +15,38 @@ package App::Porting::updateAUTHORS;
...
 
-    my $self= Porting::updateAUTHORS->new(
-        authors_file => $authors_file,
-        mailmap_file => $mailmap_file,
-    );
+    my $self= Porting::updateAUTHORS->new(%opts);
+
+    $self->read_and_update();
 
-    $self->read_and_update($authors_file, $mailmap_file);
```
bram-perl: *commit updateAUTHORS.pl - rework to use the hash of options to GOL*:

This change belongs in commit "updateAUTHORS.pm - use object to store state and avoid passing args"

Porting/updateAUTHORS.pm
```diff
@@ -156,29 +161,48 @@ sub update_authors {
...
+            }
+            print $out_fh encode_utf8($author), "\n"
+                or die "Failed to print to '$authors_file_new': $!";
+        }
+        close $out_fh
+    }
+    if ($new_raw_text ne $old_raw_text) {
+        $self->{changed_count}++;
+        $self->{changed_file}{$authors_file}++;
+        return 1 unless $self->{include_contrib};
```
bram-perl: *commit updateAUTHORS.pm - don't update files if they don't change*:

This line (`return 1 unless $self->{include_contrib};`) probably belongs in a later commit (I don't know yet which one; I haven't gotten that far yet).

Porting/updateAUTHORS.pm
```diff
@@ -244,21 +271,44 @@ sub update_mailmap {
...
+            print $out encode_utf8($line), "\n"
+                or die "Failed to print to '$mailmap_file': $!";
+        }
+        close $out;
+    }
+    if ($new_raw_text ne $old_raw_text) {
+        $self->{changed_count}++;
+        $self->{changed_file}{$mailmap_file}++;
+
+        return 1 unless $self->{include_contrib};
```
bram-perl: *commit updateAUTHORS.pm - don't update files if they don't change*:

This line (`return 1 unless $self->{include_contrib};`) probably belongs in a later commit (I don't know yet which one; I haven't gotten that far yet).

Porting/updateAUTHORS.pm
```diff
@@ -156,29 +161,48 @@ sub update_authors {
...
+        binmode $out_fh;
+        print $out_fh $new_raw_text;
+        close $out_fh
+            or die "Failed to close '$authors_file_new': $!";
+        rename $authors_file_new, $authors_file
+            or die "Failed to rename '$authors_file_new' to '$authors_file':$!";
+        return 1;
+    }
+    else {
+        return 0;
```
bram-perl, line 454: *commit updateAUTHORS.pm - don't update files if they don't change*:

Looking at the return values (of `update_authors` and `update_mailmap`) I started wondering what they mean but it appears they're not documented?

Return value 0 is clear, the content is still up to date/no changes/...
Return value 1 is a bit less clear (due to the `$self->{include_contrib}`) that is in the commit: it appears to mean old and new is not the same. However depending on `include_contrib` the file is or isn't updated..)

Porting/updateAUTHORS.pm
```diff
@@ -244,21 +271,44 @@ sub update_mailmap {
...
+    {
+        open my $out, ">", \$new_raw_text
+            or die "Failed to open scalar buffer for write:$!";
+        foreach
+            my $line (@$mailmap_preamble, __sorted_hash_keys($mailmap_hash),)
+        {
+            print $out encode_utf8($line), "\n"
+                or die "Failed to print to '$mailmap_file': $!";
+        }
+        close $out;
```
bram-perl: *commit updateAUTHORS.pm - don't update files if they don't change*:

Return value of `close $out` isn't checked? (In `update_authors` it does check the return value, so it just looks a bit inconsistent to check it one place but not the other)

Porting/updateAUTHORS.pm
```diff
@@ -156,29 +161,48 @@ sub update_authors {
...
+                or die "Failed to print to '$authors_file_new': $!";
+        }
+        close $out_fh
+    }
+    if ($new_raw_text ne $old_raw_text) {
+        $self->{changed_count}++;
+        $self->{changed_file}{$authors_file}++;
+        return 1 unless $self->{include_contrib};
+
+        warn "Updating '$authors_file'\n" if $self->{verbose};
```
bram-perl, line 440: *commit updateAUTHORS.pm - don't update files if they don't change*:

Idle wondering: does this change (`if $self->{verbose}`) in `update_authors` and `update_mailmap` belong in commit "updateAUTHORS.pl - add support for verbose"?

(Now the code in this commit is using an undocumented/not yet existing option)

Porting/updateAUTHORS.pl
```diff
@@ -20,6 +20,8 @@ package App::Porting::updateAUTHORS;
     man
     authors_file=s
     mailmap_file=s
+
+    verbose+
```
bram-perl, line 27: *commit updateAUTHORS.pl - add support for verbose*:

Documentation in updateAUTHORS.pm should also be updated for the verbose option.

Porting/updateAUTHORS.pm
```diff
@@ -70,7 +70,14 @@ sub read_commit_log {
     my $author_info= $self->{author_info}   ||= {};
     my $mailmap_info= $self->{mailmap_info} ||= {};
 
-    open my $fh, qq(git log --pretty='tformat:$tformat' |);
+    my $commit_range= $self->{commit_range};
```
bram-perl, line 208: *commit updateAUTHORS.p[lm] - add support for selecting the commit range*:

Options 'commit_range' and 'numstat' are not documented in updateAUTHORS.pm

Porting/updateAUTHORS.pl
```diff
@@ -126,6 +151,60 @@ =head1 OPTIONS
...
+email in the AUTHORS file should use C<< <unknown> >>.
+
+For example:
+
+  Porting/updateAUTHORS.pl --exclude-contrib="Joe B <b@joe.com>"
+
+Would remove all references to "Joe B" from F<AUTHORS> and F<.mailmap>
+and add the required entires to F<Porting/exclude_contrib.txt> such that
+the contributor would never be automatically added back, and would be
+automatically removed should someone read them manually.
```
bram-perl, line 498: *commit updateAUTHORS.p[lm] - add exclusion support*:

Small typo: 'read' -> 'readd'

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
 
-    return undef;
+    $self->update_exclude_file();
+
+    return $self->changed_count();
+}
+
+sub read_exclude_file {
+    my ($self, $file_name)= @_;
+    $file_name ||= $self->{exclude_file};
```
bram-perl: *commit updateAUTHORS.p[lm] - add exclusion support*

This looks a bit inconsistent: `read_exclude_file` takes an optional filename as argument but `read_authors` and `read_mailmap` do not.

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
+    if ($exclude_text ne $self->{exclude_file_text_orig}) {
+        $self->{changed_count}++;
+        $self->{changed_file}{ $self->{exclude_file} }++;
+        warn "Updating '$self->{exclude_file}'\n" if $self->{verbose};
+
+        my $tmp_file = "$self->{exclude_file}.new";
+        open my $out_fh, ">", $tmp_file
+            or die "Cant open '$tmp_file' for append: $!";
+        print $out_fh $exclude_text;
+        close $out_fh;
```
bram-perl: *commit updateAUTHORS.p[lm] - add exclusion support*:

No error checking on `close $out_fh;`
(`update_authors` does do error checking on `close`; having one method do it and the other looks inconsistent)

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
+        rename $tmp_file, $self->{exclude_file}
+            or die "Failed to write to exclude file '%s': $!";
```
bram-perl: *commit updateAUTHORS.p[lm] - add exclusion support*:

Incorrect error message: it's a rename not a write + (`%s` in the error message won't really work)

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
+
+sub unchanged_file {
+    my ($self, $name)= @_;
+    return $self->changed_file($name) ? 0 : 1;
 }
 
 sub new {
-    my ($class, %opts)= @_;
-    return bless \%opts, $class;
+    my ($class, %self)= @_;
```
bram-perl, line 949: *commit updateAUTHORS.p[lm] - add exclusion support*

Docs in updateAUTHORS.pm need to be updated for the new options

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
     $self->add_new_mailmap_entries()
         and $self->update_mailmap();
 
-    return undef;
+    $self->update_exclude_file();
+
+    return $self->changed_count();
+}
+
+sub read_exclude_file {
```
bram-perl, line 862: *commit updateAUTHORS.p[lm] - add exclusion support*

Should the docs in updateAUTHORS.pm be updated for this function?

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
+        $exclude_digest->{$line}++ if length($line);
+    }
+    close $in_fh;
+    $self->{exclude_file_text_head}= $head;
+    $self->{exclude_file_text_orig}= $orig;
+
+    return $exclude_digest;
+}
+
+sub update_exclude_file {
```
bram-perl, line 897: *commit updateAUTHORS.p[lm] - add exclusion support*

Should the docs in updateAUTHORS.pm be updated for this function?

Porting/updateAUTHORS.pm
```diff
@@ -817,6 +904,148 @@ sub _exclude_contrib {
...
+    "name"           => "Name",
+    "pos"            => "Pos",
+    "num_files"      => "NFiles",
+    "lines_added"    => "L++",
+    "lines_removed"  => "L--",
+    "lines_delta"    => "L+-",
+    "binary_changed" => "Bin+-",
+);
+
+sub report_stats {
```
bram-perl, line 1069: *updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has*:

Should the docs in updateAUTHORS.pm be updated for this function?

Porting/updateAUTHORS.pm
```diff
@@ -189,6 +260,22 @@ sub read_commit_log {
...
+            $self->{who_stats}{$committer_name_mm_canon}{applied}++;
+        }
+    }
+    $self->finalize_commit_info($last_commit_info) if $last_commit_info;
+    if (!$commits_read) {
+        if ($self->{commit_range}) {
+            die "No commits in range '$self->{commit_range}'\n";
+        }
+        else {
+            die "Panic! There are no commits!\n";
```
bram-perl, line 292: *commit updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has*:

This might be better suites in commit "updateAUTHORS.p[lm] - add support for selecting the commit range "?

Porting/updateAUTHORS.pm
```diff
@@ -70,7 +70,14 @@ sub read_commit_log {
     my $author_info= $self->{author_info}   ||= {};
     my $mailmap_info= $self->{mailmap_info} ||= {};
 
-    open my $fh, qq(git log --pretty='tformat:$tformat' |);
+    my $commit_range= $self->{commit_range};
+    my $commits_read= 0;
+
+    my $numstat= $self->{numstat} ? "--numstat" : "";
+
+    my $last_commit_info;
```
bram-perl, line 213: *commit updateAUTHORS.p[lm] - add support for selecting the commit range*:

This might be better suited in commit "updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has"?

(In this commit `$last_commit_info` isn't used)

Porting/updateAUTHORS.pl
```diff
@@ -374,31 +374,77 @@ =head1 DESCRIPTION
...
-commit. If they are not then update the AUTHORS or .mailmap files as is
-appropriate and run the tool again. Typically you shouldn't need to do
-either unless you are changing the default name or email for a user. For
-instance if a person currently listed in the AUTHORS file whishes to
-change their preferred name or email then change it in the AUTHORS file
-and run the script again. I am not sure when you might need to directly
-modify .mailmap, usually modifying the AUTHORS file should suffice.
+message or it will update the files as necessary, possibly not at all
+if there is no need to do so. If the C<--validate> option is provided
+the the content will not be updated and instead the tool will act as a
```
bram-perl, line 626: *commit updateAUTHORS.pl - improved pod:*

There is no `--validate` option (it's added later in commit "updateAUTHORS.pl - add support for --tap ")

Porting/updateAUTHORS.pl
```diff
@@ -374,31 +374,77 @@ =head1 DESCRIPTION
...
+=head1 RECIPES
+
+  \$ perl Porting/updateAUTHORS.pl --who --from=v5.31.6 --to=v5.31.7
+  \$ perl Porting/updateAUTHORS.pl --who v5.31.6..v5.31.7
+  \$ perl Porting/updateAUTHORS.pl --rank --percentage --from=v5.31.6
+  \$ perl Porting/updateAUTHORS.pl --thanks-applied --from=v5.31.6
+  \$ perl Porting/updateAUTHORS.pl --tap --from=v5.31.6
+  \$ perl Porting/updateAUTHORS.pl --files --from=v5.31.6
+  \$ perl Porting/updateAUTHORS.pl --activity --from=v5.31.6
+  \$ perl Porting/updateAUTHORS.pl --chainsaw v5.31.6..HEAD
```
bram-perl: *commit updateAUTHORS.pl - improved pod*:

Escaping `$` shouldn't be needed in POD;
(Escaping it causes it to be rendered as `\$ perl ...` (at least in `perldoc`))

Porting/updateAUTHORS.pm
```diff
@@ -554,12 +665,156 @@ sub read_and_update {
...
+
+    my $self= bless \%self, $class;
+
+    if (my $ary= $self->{exclude_contrib}) {
+        $self->_exclude_contrib($_) for @$ary;
+    }
+
+    $self->read_exclude_file();
+
+    die Dumper(\%self) if $self{dump_opts};
```
bram-perl, line 964: *commit: updateAUTHORS.p[lm] - add exclusion support*

This (`$self->{dump_opts}`) belongs in commit *updateAUTHORS.pl - Add debugging option*
Also:
- I wouldn't expect a debug option to `die`
- `dump_opts` not documented in POD of uodateAUTHORS.pm

Porting/updateAUTHORS.pl
```diff
@@ -25,6 +25,7 @@ package App::Porting::updateAUTHORS;
     exclude_missing|exclude
     exclude_contrib=s@
     exclude_me
+    dump_opts
```
bram-perl, line 31: *commit updateAUTHORS.pl - Add debugging option *

Missing docs?

Porting/updateAUTHORS.pl
```diff
@@ -102,7 +105,46 @@ sub main {
...
+            sprintf "%s is listed in AUTHORS", _clean_name($_))
+            for sort keys %{$self->{missing_author}||{}};
+
+        my $uncommitted_files= $self->git_status_porcelain;
+        if ($uncommitted_files) {
+            my ($author_name, $author_email)= $self->current_author_name_email();
+            my ($committer_name, $committer_email)= $self->current_committer_name_email();
+
+            ok($author_name && $author_email,"git knows your author name and email.");
+            ok($committer_name && $committer_email,"git knows your author name and email.");
```
bram-perl: *commit updateAUTHORS.pl - add support for --tap *

Wrong test description (checking for `$committer_name` while message is "author name")

Porting/updateAUTHORS.pl
```diff
@@ -102,7 +105,46 @@ sub main {
...
+
+            my $author_known= $self->known_contributor($author_name,$author_email);
+            my $committer_known= $self->known_contributor($committer_name,$committer_email);
+            if (is( $author_known && $committer_known, 1,
+                "Uncommitted changes are by a known contributor?"))
+            {
+                diag "Testing uncommtted changes! Remember to commit before you push!"
+                    if $ENV{TEST_VERBOSE};
+            } else {
+                diag uncommitted_changes($author_name,$author_email,$committer_name, $committer_email);
```
bram-perl: *commit updateAUTHORS.pl - add support for --tap*:

I'm guessing this should be: `diag error_advice_for_uncommitted_changes(...)`?

Porting/updateAUTHORS.pl
```diff
@@ -149,6 +191,94 @@ sub main {
 
 exit(main()) unless caller;
 
+sub error_advice_for_uncommitted_changes {
+    my ($author_name, $author_email, $committer_name, $committer_email)= @_;
+    my $props= "";
+    my $extra= "";
+    my @git_env_keys=  map {
```
bram-perl: *commit updateAUTHORS.pl - add support for --tap*:

I would add an extra comment in this function;

Something like: (a first draft, may come up with something better later)
```
Check if one of the ENV keys influenced $author_name, $author_email, $committer_name, $committer_email to provide a more meaningful help message.

In this function the git config isn't checked: this is already done by the calling code and result of those checks are included in $author_name, ...
```

Porting/updateAUTHORS.pl
```diff
@@ -529,16 +562,13 @@ =head1 DESCRIPTION
...
 
 =head2 CHANGING A CONTRIBUTORS CANONICAL NAME OR EMAIL
 
-Manually modify F<AUTHORS> so that the name or email is correct. Run the
-tool with no arguments, it should update C<.mailmap> appropriately.
-Check the changes before you commit them. If you wish to change BOTH
-files at the same time you should do it in two steps, running the tool
-in between each step. Check the changes and then commit them.
+Use the C<--change-name-for-name> and related options. This will do
+things "properly" and update all the files.
```
bram-perl, line 646: *commit updateAUTHORS.p[lm] - Add a way to rename authors properly*:

An example would be nice

t/porting/authors.t
```diff
@@ -16,31 +16,7 @@ skip_all(
...
 
-# This is the subset of "pretty=fuller" that checkAUTHORS.pl actually needs:
-print qx{git log --pretty=format:"Author: %aN <%aE>" $revision_range | $^X Porting/checkAUTHORS.pl --tap -};
+exec("$^X Porting/updateAUTHORS.pl --validate");
```
bram-perl: *commit checkAUTHORS.pl - delete and replace with updateAUTHORS.pl*:

I think the `$revision_range` logic should stay and that `$revision_range` should be included in the `Porting/updateAUTHORS.pl` call.

Reading the comments it looks like it was added for a reason and I would expect that reason to still be there..

bram-perl (commented):
Done a commit-per-commit review.

(I may have gotten carried away a bit during reviewing - hope you don't mind :) )

Porting/updateAUTHORS.pm
demerphq: No, the verbose patch should have been before this.

Porting/updateAUTHORS.pl
demerphq: Meh. I did it, but this .pm file is ONLY used by updateAUTHORS.pl, and its not really meant for reuse. I dont feel like it needs a CPAN "well used api" level of documentation of internal features.

Porting/updateAUTHORS.pm
demerphq: Both are internal use only, and I dont think they need to be explicitly documented. I did mention "commit_range" I did not mention numstat.

However I added a paragraph pointing that anyone using updateAUTHORS.pm should consult the docs and implementation for sub man in updateAUTHORS.pl

Porting/updateAUTHORS.pm
demerphq: Well, not to me. 

If the update process updates nothing because the commit range evaluates to nothing then it does not surprise me that much and i dont necessarily expect to be told about it.

If the stats for a given commit range are empty and i get no stats I would like to be told that i am looking at no data.

Porting/updateAUTHORS.pl
demerphq: that came from checkAUTHORS.pl

Porting/updateAUTHORS.pl
demerphq: yep.

Porting/updateAUTHORS.pl
demerphq: I added /some/ verbiage, but i am not sure it is really necessary. Anyone tracing the code flow there would know what is up. Regardless I worked this a fair bit.

```
not ok 6 - Uncommitted changes are by a known contributor?
#   Failed test 'Uncommitted changes are by a known contributor?'
#   at Porting/updateAUTHORS.pl line 133.
#          got: '0'
#     expected: '1'
# 
# There are uncommitted changes in the working directory and your git
# credentials are new to us. We think that git thinks your credentials
# are as follows (git may use defaults we don't guess properly):
# 
#     Author Name     = bob
#     Author Email    = demerphq@gmail.com
#     Committer Name  = Yves Orton
#     Committer Email = demerphq@gmail.com
# 
# You have configured the following relevant git config settings:
# 
#     author.name      = bob
#     user.name        = Yves Orton
#     user.email       = demerphq@gmail.com
# 
# 
# Your committer and author details differ. You may want to review your
# git configuration.
# 
# 
# To resolve this you can perform one or more of these steps:
# 
#     1. Remove the uncommitted changes, including untracked files that
#        show up in
# 
#             git status
# 
#        if you wish to REMOVE UNTRACKED FILES and DELETE ANY CHANGES
#        you can
# 
#             git clean -dfx
#             git checkout -f
# 
#         BE WARNED: THIS MAY LOSE DATA.
# 
#     2. You are already configured in git and you just need to add
#        yourself to AUTHORS and other infra: commit the changes in the
#        working directory, including any untracked files that you plan to
#        add (the rest should be removed), and then run
# 
#             Porting/updateAUTHORS.pl
# 
#        to update the AUTHORS and .mailmap files automatically. Inspect
#        the changes it makes and then commit them once you are
#        satisfied. This is your option to decide who you will be known
#        as in the future!
# 
#     3. You are already a contributor to the project but you are committing
#        changes on behalf of someone who is new. Run
# 
#             Porting/updateAUTHORS.pl
# 
#        to update the AUTHORS and .mailmap files automatically. Inspect
#        the changes it makes and then commit them once you are satisfied.
#        Make sure the conributor is ok with the decisions you make before
#        you merge.
# 
#     3. You are already an author but your git config is broken or
#        different from what you expect, or you are a new author but you
#        havent configured your git details properly, in which case you
#        can use something like the following commands:
# 
#             git config --set user.name "Some Name"
#             git config --set user.email "somewhere@provider"
# 
#        If you are known to the project already this is all you need to
#        do. If you are not then you should perform option 2 or 4 as well
#        afterwards.
# 
#     4. You do not want to be listed in AUTHORS: commit the changes,
#        including any untracked unignored files, and then run
# 
#             Porting/updateAUTHORS.pl --exclude
# 
#        and commit the changes it creates. This test should pass once
#        those commits are created. Thank you for your contributions.
```

t/porting/authors.t
demerphq: This is the only feedback you gave that I disagree with.

The change to remove this logic was deliberate and motivated by the fact that I am pretty sure that incremental checks of this data are error prone and have lead to data being missed from AUTHORS. When I first implemented updateAUTHORS I ended up adding multiple users that were not detected as missing by the old logic, and the only way i could explain their exclusion was faulty range input in the author tests.

IMO given that the cost to check the entire commit graph is negligible, and is obviously correct we should use it unless there is a clear reason we should do something different. Using these env values may or may not produce correct results, it is not clear, not using them gives correct results, thus we shouldn't use them.

I guess maybe we should do something special with the travis case but I would like to see what fails and how /exactly/ before I said what it should be. There is more than one way to do things. :-)

Eg, I would like to see some reports of this failing as it is coded here before we patch it to deal with the travis case.

bram-perl: A suggestion: can the removal of the revision logic be taken out of this PR and a new one created for it?
This to get the bulk of the changes merged and then the revision range logic can be discussed/investigated/troubleshooted/removed/.. separately?

I currently have done no testing with github actions so I can only go by the comments in the code..
I may/will do some testing with it but that may take a couple of days
What I somewhat "fear" (based on the comments) is that github actions add a dummy merge commit (in some cases?) and that this contains a "bogus" username/emailaddress which (obviously) shouldn't be included in AUTHORS/.mailmap/...

demerphq:
Hi @bram-perl I have incorporated most of your feedback, either literally as you suggested, or in some case in spirit but i went a different way.

The only points I did not go along with were the following two:

1. extensive docs in updateAUTHORS.pm. First, updateAUTHORS.pm is the "brain" of "updateAUTHORS.pl", and many things are documented in the .pl script that apply to the .pm file. So I added a paragraph saying that people should be familiar with both if they want to hack on this, but to put it simply I don't think the docs for the internals of a Porting/ script need to be at CPAN quality levels, and definitely that they aren't shouldn't be a blocker to move this ball forward now. This code isn't intended for reuse (outside of the Porting directory anyway) and its already better documented and structured than what it replaces, likely I will be one of the few who work on it, and frankly I'd like to put it to bed for now for a while.  Patches welcome of course if you think otherwise, and I might even do a follow up PR with more docs when I have more energy, but right now I am growing exhausted with this PR and I would like to get it merged and rest up on it a bit before I do anything else. Doc patches are sometimes best written by the person who didnt write the code anyway, as the have a better eye for what is obvious or not.
2. The control logic for authors.t, I think we should try to avoid commit range scanning in our testing as I believe it is fragile and likely to break. I am open to the idea that we need to do something special for TRAVIS but i would like to see what breaks before I say what it is. My feeling is right now are doing it wrong, and we should consider an alternative strategy, but i need data about how the test fails without the TRAVIS special case.

So if you don't mind I will let you push a follow PR for point 1. For point 2, I am open to doing something, but I would really like to know how this fails when that travis logic fires, which is simplest done by merging, waiting for it to fail (if it does) and then looking at the failure and then dealing with it. If folks are ok with that I would like to proceed that way.

Porting/updateAUTHORS.pm
demerphq: its a debugging tool for me or someone else debugging the tool, i dont plan to document it. same for numstat.

Porting/updateAUTHORS.pl
demerphq: Nope. Omitted not missing.

bram-perl:
1. The docs: ok,you did document some of it so I wasn't sure how extensive you wanted the docs so I just added a comment when I spotted something. Not wanting to write/have extensive docs for it is acceptable (at least to me it is).

2. As suggested earlier: I would just take it out of this PR and submit it as a separate PR. (I think the error in the revision logic is already know, and that PR #20113 is the fix for it but the reasons why it's needed - if it's needed at all - should be better documented)

I'll try todo a final review tonight on this PR;

t/porting/authors.t
demerphq: > can the removal of the revision logic be taken out of this PR and a new one created for it?

Roger. Wilco.

demerphq: And done (not the second PR). But I really do want to debug this without the range logic. I have a feeling it wont be that long before we have patches that fail locally that do not fail in CI.

demerphq:
Re: POD docs for the module. I would /like/, long run, to have complete docs for this, but I don't see those docs as a necessary precondition to getting the PR merged.  Rather I would like to get this merged and then incrementally improve it over time, ideally with the support from others who care about these things. IMO after the PR (very much with your help) updateAUTHORS.pl will be a better tool than we had with either itself or checkAUTHORS.pl before this PR. We can incrementally make it even better as we like once its merged.

demerphq:
BTW, I am deliberately pushing commits that fail the authors test right now, i will force push the fix for it aftewards

jkeenan:
The subject of this pull request indicates it's a "DRAFT".  However, it's not categorized as a draft on github.  Could that be clarified?

demerphq:
@bram-perl fwiw i can see some issues where the commit messages are a little out of sync with the contents. ill fix that today and repush.

bram-perl:
> @bram-perl fwiw i can see some issues where the commit messages are a little out of sync with the contents. ill fix that today and repush.

ok, just let me know when it's done and I'll review

demerphq:
@jkeenan I think I will remove the "DRAFT" relatively soon, its pretty close to being ready for merge now IMO, so not sure there is any point in getting github to think it is one.

demerphq:
@bram-perl I think I fixed all the issues I noticed when I reviewed. Please go ahead, ill avoid any force pushes until i hear from you. Thanks for your time!

bram-perl:
It's looking good, there are some minor changes I would still do (moving things from one commit to another); but rather then burden you with them I did these myself using fixup commits[^1].

Applying them is slightly trickier but following steps should work assuming:
- your current tree is clean (no uncommited changes, no untracked files, ..)
- your first commit is e6d0e3e
- your last commit is 06460a3

In your perl5 repo: (i.e. origin=github.com/perl/perl5)
```
git checkout yves/exclude_contrib
git remote add bram https://github.com/bram-perl/perl5
git fetch bram
git rebase -i 7aa33fb
```
in the editor put:
```
pick e6d0e3e testsuite.yml - fix and rename "authors" CI workflow to "authors_involved_debug"
pick 673d0f5 .mailmap - add missing entries using updateAUTHORS.pl
pick f1ebb11 updateAUTHORS.pl - split into module Porting/updateAUTHORS.pm
pick 3175823 fixup! updateAUTHORS.pl - split into module Porting/updateAUTHORS.pm
pick 6ac1a3b fixup! updateAUTHORS.pl - split into module Porting/updateAUTHORS.pm
pick bdb8db5 fixup! updateAUTHORS.pl - split into module Porting/updateAUTHORS.pm
pick fa0934c updateAUTHORS.pm - use object to store state and avoid passing args
pick 016ab00 fixup! updateAUTHORS.pm - use object to store state and avoid passing args
pick d3b46b6 fixup! updateAUTHORS.pm - use object to store state and avoid passing args
pick f734e7f updateAUTHORS.pl - use Getopt::Long in "hash ref" mode
pick 4617c69 updateAUTHORS.pl - add support for verbose
pick d131c5c updateAUTHORS.pm - don't update files if they don't change
pick 8521c85 fixup! updateAUTHORS.pm - don't update files if they don't change
pick c6a70a9 fixup! updateAUTHORS.pm - don't update files if they don't change
pick f07e026 fixup! updateAUTHORS.pm - don't update files if they don't change
pick 4187e3b updateAUTHORS.pl - add support for a commit range
pick 09834f4 fixup! updateAUTHORS.pl - add support for a commit range
pick 6e098ff updateAUTHORS.pm - use _file suffix for methods that read/update files
pick 3a21d07 updateAUTHORS.p[lm] - add exclusion support
pick 7761494 fixup! updateAUTHORS.p[lm] - add exclusion support
pick 5691a27 fixup! updateAUTHORS.p[lm] - add exclusion support
pick 615cd2c fixup! updateAUTHORS.p[lm] - add exclusion support
pick f35118a updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has
pick a4fcfb7 fixup! updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has
pick 21d6fb6 fixup! updateAUTHORS.p[lm] - add support for reports like checkAUTHORS.pl has
pick 19ebf33 updateAUTHORS.pm - Fix windows support for git operations
pick 0751ba3 updateAUTHORS.pm - use fc() for sorting, and improved subs to do so
pick 6d82b90 updateAUTHORS.pl - Add debugging option
pick 62db29e AUTHORS - update the blurb
pick 9f5c7f9 updateAUTHORS.pl - add support for --tap
pick cb19cf0 updateAUTHORS.pl - improved pod explanations
pick eec890a updateAUTHORS.pm - remove redundant data from .mailmap
pick 9fd8bda fixup! updateAUTHORS.pm - remove redundant data from .mailmap
pick 3d29923 updateAUTHORS.p[lm] - Add a way to rename authors properly
pick 6659c97 checkAUTHORS.pl - delete and replace with updateAUTHORS.pl
pick e6ca45b AUTHORS, .mailmap - resolve issue #20010
pick dbb255d *FIXUP* reports - make sure we turn on numstat when required
pick 5817f37 *FIXUP* for reports patch
pick 3cf49b8 *FIXUP* - remove commented out debug code (redundant mailmap)
pick 1fe51f2 updateAUTHORS.p[lm] - more generic update skipping
pick 87befb8 update_authors.t - test updateAUTHORS.pl
pick 06460a3 updateAUTHORS.p[lm] - filter out "unknown" authors
```
close the editor, the rebase will/should give 5 conflicts, fix them using:
```
git checkout --theirs . && git add . && git rebase --continue
```
(don't worry about the conflicts, there are fixup to fix it up)

And when that is all done the final step is:
```
git rebase -i 7aa33fb --autosquash
```


[^1]: some of the fixup are duplicated to fix things (again) after a  the `git checkout --theirs`

bram-perl:
one thing to consider (that I did not change):

updateAUTHORS.pl in it's final version contains:
```
sub main {
    ....
    my $changed= $self->read_and_update();
    ...
    return $changed; # 0 means nothing changed
}

exit(main()) unless caller;
```

Meaning when it did make changes the exit code will be 1.
That seems a bit strange.

demerphq:
Thanks for the feedback and commits. Ill review them and the instructions you left and get them merged up. Regarding the exit code my thought was that things like this:

```
$ Porting/updateAUTHORS.pl --no-update || echo "Files need updating"
$ Porting/updateAUTHORS.pl || echo "Files were changed"
```

might be useful behavior. It should be documented tho. Ill tweak that when i integrate your fixups.

demerphq:
BTW the test fails are because I didnt realize the tests were run under a shallow clone. Not sure what I can do about that except skip them in CI or change the CI workflow to use a full clone.

bram-perl:
> Thanks for the feedback and commits. Ill review them and the instructions you left and get them merged up.

If you should run into troubles let me know (or come find me on IRC)

>  Regarding the exit code my thought was that things like this:
> 
> ```
> $ Porting/updateAUTHORS.pl --no-update || echo "Files need updating"
> $ Porting/updateAUTHORS.pl || echo "Files were changed"
> ```
> 
> might be useful behavior. It should be documented tho. Ill tweak that when i integrate your fixups.

If it's intended behavior then it's OK.
(The `|| echo "Files were changed"` is a bit simplistic tho, it's really: `|| echo "Files were changed or something went wrong"`)

bram-perl:
> BTW the test fails are because I didnt realize the tests were run under a shallow clone. Not sure what I can do about that except skip them in CI or change the CI workflow to use a full clone.

The 'sanity check' in the CI runs with a full clone so it should be able to do the tests there.
Which would mean the other builds should be able to skip it since the sanity check already verified it.
Can the test detect it's run with a shallow clone? (and then skip the tests?)

I've also been thinking about CI/authors test in general but that's more a discussion to have in #19867 and not in this PR

demerphq:
On Sat, 20 Aug 2022, 11:37 Bram, ***@***.***> wrote:

> Thanks for the feedback and commits. Ill review them and the instructions
> you left and get them merged up.
>
> If you should run into troubles let me know (or come find me on IRC)
>
> Regarding the exit code my thought was that things like this:
>
> $ Porting/updateAUTHORS.pl --no-update || echo "Files need updating"
> $ Porting/updateAUTHORS.pl || echo "Files were changed"
>
> might be useful behavior. It should be documented tho. Ill tweak that when
> i integrate your fixups.
>
> If it's intended behavior then it's OK.
> (The || echo "Files were changed" is a bit simplistic tho, it's really: ||
> echo "Files were changed or something went wrong")
>

True. We can use a specific exit code to be able to disambiguate.

Yves

>

bram-perl:
A shallow clone can be detected by checking: `git rev-parse --is-shallow-repository` or by checking if the file `.git/shallow` exists.
Note that the `git rev-parse` needs a recent enough version of git;
With a recent version:
```
$ git rev-parse --is-shallow-repository
false
```
With an old version:
```
$ git rev-parse --is-shallow-repository
--is-shallow-repository
```

What Porting/cmpVERSION.pl appears to check are the git tags:
```
unless (length $tag_to_compare) {
    die "$0: Git found, but no Git tags found\n"
	unless $tap;
    print "1..0 # SKIP: Git found, but no Git tags found\n";
    exit 0;
}
```

An alternative (that is also in Porting/cmpVERSION.pl) is to check if the perl-1 commit exists:
```
    if (-f '.git') {
        # the hash of the initial commit in perl.git (perl-1.0)
	my $commit = '8d063cd8450e59ea1c611a2f4f5a21059a2804f1';
	my $out = `git rev-parse --verify --quiet '$commit^{commit}'`;
	chomp $out;
	if($out eq $commit) {
            ++$found;
	}
    }
```

bram-perl:
> A shallow clone can be detected by checking

Just a heads up: I'm working on a PR for this (#20124) because CI runs for PRs are currently red (since commit 16dd3f70cc16005d5af7146385733a7c945fb67e was merged on blead).

demerphq:
Hey @bram-perl I applied your changes, fixup-merged some of my own, and perltidied the sequence. I have two minor additional changes I want to merge that I will do right now (in the next 30 minutes), but i think then it will be good to go.

Re: #20124 thanks for picking that up. I guess my bad on merging that. :-(

demerphq:
Oh, btw, for the question of what to do with the shallow clones I went with the "alternative" strategy of checking if the commit id exists much as you described.

demerphq:
Ok, that last push moved the creation of the update_authors.t test script to early in the sequence so that i could test every commit in the sequence using:

```
git rebase --exec 'Porting/updateAUTHORS.pl' origin/blead
git rebase --exec 'TEST_JOBS=8 make -j8 test_porting' origin/blead
```
And both pass clean. The code is tidied according to the previously documented rules. It has tests for at least some of its functions (and ill add more after it gets merged.) Im thinking if it passes test now it is good to merge. @bram-perl really appreciate all your help.

demerphq:
I lied. That last push was the last push. I needed to filter .bak files out of the porting tests. Better to have that at the beginning of the PR. Really it could be another PR, but I don't see the point given the patch sequence will be rebased anyway, same for the first few patches in the sequences frankly.

demerphq:
I'm gunna stop saying I'm stopping pushing patches to this branch. But I'd like to merge it so I stop pushing patches to the branch. :-)

bram-perl (approved):

AUTHORS
```diff
@@ -1396,7 +1396,6 @@ Ville Skyttä                   <scop@cs132170.pp.htv.fi>
 Vincent Pit                    <perl@profvince.com>
 Vishal Bhatia                  <vishal@deja.com>
 Vitali Peil                    <vitali.peil@uni-bielefeld.de>
-vividsnow                      <vividsnow@gmail.com>
 vividsnow                      @vividsnow
```
Leont: I suspect it's the other one that you want to remove?

demerphq: I assumed vividsnow wanted to hide their email by using the github id, there are a few other cases in AUTHORS like that. I dont have a preference which we use or keep. I just checked the history and the @vividsnow came last, and was part of an automated update which i guess went wrong, so I guess you are right. Ill reverse that change.

demerphq: Reversed. Thanks for the catch @Leont.

demerphq:
rebased to deal with the conflicts in .mailmap

PR #20072, opened 2022-08-09
Ensure cmd.exe is used in pl2bat tests

kenneth-olwing:
Tests done in cpan/ExtUtils-PL2Bat/t/make_executable.t are geared to work with cmd.exe; a user having something else as the COMSPEC (e.g. in my case, TCC (see [jpsoft.com](jpsoft.com)) may well get confusing errors. Hence, ensure that COMSPEC is set to cmd.exe.


Leont:
Changes in `cpan/` should be done upstream first. I've just released a new ExtUtils::PL2Bat to CPAN with your fix, assuming no problems show up we will update the version in core as well.

kenneth-olwing:
Aha. Thanks for your help!

ken1

PR #20073, opened 2022-08-09
Define a `CvREFCOUNTED_ANYSV` flag

leonerd:
If this flag is set, then the `CvXSUBANY(cv).any_sv` pointer will have its reference count decremented when the CV itself is freed. This is useful for XS extensions that wish to store extra data in here. Without this flag, such extensions have to resort to using magic with a 'free' function to perform this work.

As discussed in https://www.nntp.perl.org/group/perl.perl5.porters/2022/08/msg264572.html

### DIFF
--- a/cv.h
+++ b/cv.h
@@ -134,6 +134,7 @@ See L<perlguts/Autoloading with XSUBs>.
 #define CVf_LEXICAL	0x10000 /* Omit package from name */
 #define CVf_ANONCONST	0x20000 /* :const - create anonconst op */
 #define CVf_SIGNATURE   0x40000 /* CV uses a signature */
+#define CVf_REFCOUNTED_ANYSV 0x80000 /* CvXSUBANY().any_sv is refcounted */
 
 /* This symbol for optimised communication between toke.c and op.c: */
 #define CVf_BUILTIN_ATTRS	(CVf_NOWARN_AMBIGUOUS|CVf_LVALUE|CVf_ANONCONST)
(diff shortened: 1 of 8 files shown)


cv.h
```diff
@@ -227,6 +228,32 @@ See L<perlguts/Autoloading with XSUBs>.
...
+
+=for apidoc m|bool|CvREFCOUNTED_ANYSV|CV *cv
+
+If true, indicates that the C<CvXSUBANY(cv).any_sv> member contains an SV
+pointer whose reference count should be decremented when the CV itself is
+freed.
+
+Any CV that wraps an XSUB has an C<ANY> union that the XSUB function is free
+to use for its own purposes.  It may be the case that the code wishes to store
+an SV in C<any_sv> member of this union.  By setting this flag, this SV
```
wolfsage: nit: `in C<any_sv>` -> `in the C<any_sv>`

wolfsage (commented):
One small doc nit.

Otherwise, it'd be nice to have tests...

leonerd:
> One small doc nit.

Wording fixed.

> Otherwise, it'd be nice to have tests...

Yeeeah. Hrm. A test could be written but that's basically going to be a lot of lines added in probably XS-APItest. We don't have much existing framework for testing this kind of thing - even no equivalent of Test::Refcount in core until I recently added it.

pad.c
```diff
@@ -2201,6 +2204,8 @@ S_cv_clone(pTHX_ CV *proto, CV *cv, CV *outside, HV *cloned)
     if (UNLIKELY(CvISXSUB(proto))) {
         CvXSUB(cv)    = CvXSUB(proto);
         CvXSUBANY(cv) = CvXSUBANY(proto);
+        if(CvREFCOUNTED_ANYSV(cv))
```
bram-perl: (nitpicking) Missing whitespace between `if` and `(`

sv.c
```diff
@@ -14633,6 +14633,9 @@ S_sv_dup_common(pTHX_ const SV *const ssv, CLONE_PARAMS *const param)
                 } else if (CvCONST(dsv)) {
                     CvXSUBANY(dsv).any_ptr =
                         sv_dup_inc((const SV *)CvXSUBANY(dsv).any_ptr, param);
+                } else if (CvREFCOUNTED_ANYSV(dsv)) {
+                    CvXSUBANY(dsv).any_sv =
+                        sv_dup_inc((const SV *)CvXSUBANY(dsv).any_sv, param);
```
bram-perl, line 14638: There was some speculation that this code might already be broken for threads (that is if dup is called for a CV which uses `CvXSUBANY`).

Is that still the case? Should that be fixed as well? (possibly as a separate commit/PR);

i.e. should the code be updated with something like:
```
              else {
                    CvXSUBANY(dsv).any_sv =
                        sv_dup((const SV *)CvXSUBANY(dsv).any_sv, param);
              }
```

(`else` because the if condition that is checked is `if (!CvISXSUB(dsv)) {`)

leonerd: > Is that still the case? Should that be fixed as well? (possibly as a separate commit/PR);
> 
> i.e. should the code be updated with something like:
> 
> ```
>               else {
>                     CvXSUBANY(dsv).any_sv =
>                         sv_dup((const SV *)CvXSUBANY(dsv).any_sv, param);
>               }
> ```

Unfortunately it can't be. The `ANY` is a union, so there's no telling which field (if indeed any at all) are actually in use. In particular we can't distinguish between this being an SV pointer, some other (maybe-malloc()'ed) pointer, an integer, or just memory junk. Only the behaviour of the code itself will know.

This in fact is why I gated all the logic behind a new flag. New code sets the flag only when it is definitely using the `.any_sv` field to store an SV pointer, so we can rely on that fact. Without the flag we cannot make any assumptions.

Leont: > There was some speculation that this code might already be broken for threads (that is if dup is called for a CV which uses CvXSUBANY).

Yeah, in `sv_dup` it will need to `sv_dup_inc`, and `cv_clone` will need to `SvREFCNT_inc`

bram-perl: Ok, that somewhat make sense.

Next question then: can/should/could it warn/croak/... if dup is called when `CvISXSUB(dsv)` is true?
The goal appears to be to 'duplicate an SV of any type' but for a SV which has an `ANY` that isn't really the case since whatever is in `ANY` doesn't get duplicated..

leonerd: > Yeah, in `sv_dup` it will need to `sv_dup_inc`, and `cv_clone` will need to `SvREFCNT_inc`

Which it now does, in the newer commits.

leonerd: > Next question then: can/should/could it warn/croak/... if dup is called when `CvISXSUB(dsv)` is true?
> The goal appears to be to 'duplicate an SV of any type' but for a SV which has an `ANY` that isn't really the case since whatever is in `ANY` doesn't get duplicated..

It cannot. It's perfectly fine to `sv_dup` a CV containing an XSUB. And it cannot reliably detect if the XSUB happens to be using the CvXSUBANY union at all; and if it is using it, which field, and to what it points. There's already going to be much code which happens to do safe things, and happens to do unsafe things, and we cannot tell which is which. I don't think a warning is possible.

bram-perl (approved):

Leont:
There is the awkward bit where you can either use this or aliases, but not both. I'm wondering if ParseXS should do something slightly smarter with that.

Related to that, this should also update the `CvXSUB` method in `ext/B/B.xs`, `Perl_do_sv_dump` and `Perl_dump_sub_perl` in dump.c; for better debuggability of `CvXSUBANY`.

leonerd:
I've added an entry to `perldelta.pod` as well, before merge.

wolfsage:
Thanks for the tests!

PR #20074, opened 2022-08-10
Test dist/ modules against older perls

tonycoz:
Add testing dist/ modules against older perls.

This currently uses the perldocker/perl-testers images rather than the official perl images since perl-testers includes the older versions, while the official perl images don't.

Unfortunately perl-testers doesn't include threaded perls, but it's an improvement over no testing at all.

I currently only test against a small subset of perls, but this is easily changed.

### DIFF
--- a/Porting/test-dist-modules.pl
+++ b/Porting/test-dist-modules.pl
@@ -1,0 +1,1087 @@
+#!perl
+# this should be perl 5.8 compatible, since it will be used
+# with old perls while testing dist modules on those perls
+use strict;
+use warnings;
+use File::Temp "tempdir";
+use ExtUtils::Manifest "maniread";
+use Cwd "getcwd";
+
+-f "Configure"
+  or die "Expected to be run from a perl checkout";
+
+my $manifest = maniread();
+
+my $start = getcwd()
+  or die "Cannot fetch current directory: $!\n";
+
+# get ppport.h
+my $pppdir = test_dist("Devel-PPPort");
+
+my $pppfile = "$pppdir/ppport.h";
+
+-f $pppfile
+  or die "No ppport.h found in $pppdir\n";
+
+opendir my $distdir, "dist" or die;
+my @dists = sort { lc $a cmp lc $b } grep { /^\w/ && $_ ne "Devel-PPPort" } readdir $distdir;
+closedir $distdir;
+
+# https://github.com/Perl/version.pm claims CPAN is upstream
+@dists = grep { $_ ne "version" } @dists;
+
+# Safe is tied pretty heavily to core
+# in any case it didn't seem simple to fix
+@dists = grep { $_ ne "Safe" } @dists;
+
+for my $dist (@dists) {
+    test_dist($dist);
+}
+
+sub test_dist {
+    my ($name) = @_;
+
+    print "*** Testing $name ***\n";
+    my $dir = tempdir( CLEANUP => 1);
+    system qq(cp -a "dist/$name/." "$dir/.")
+      and die "Cannot copy files";
+    chdir $dir
+      or die "Cannot chdir $dir: $!";
+    if ($pppfile) {
+        system "cp", $pppfile, "."
+          and die "Cannot copy $pppfile to .\n";
+    }
+    if ($name eq "IO" || $name eq "threads" || $name eq "threads-shared") {
+        write_testpl();
+    }
+    unless (-f "Makefile.PL") {
+        print "  Creating Makefile.PL for $name\n";
+        my $key = "ABSTRACT_FROM";
+        my @parts = split /-/, $name;
+        my $last = $parts[-1];
+        my $fromname;
+        for my $check ("$last.pm", join("/", "lib", @parts) . ".pm") {
+            if (-f $check) {
+                $fromname = $check;
+                last;
+            }
+        }
+        $fromname
+          or die "Cannot find ABSTRACT_FROM for $name\n";
+        my $value = $fromname;
+        open my $fh, ">", "Makefile.PL"
+          or die "Cannot create Makefile.PL: $!\n";
+        # adapted from make_ext.pl
+        printf $fh <<'EOM', $name, $fromname, $key, $value;
+use strict;
+use ExtUtils::MakeMaker;
+
+# This is what the .PL extracts to. Not the ultimate file that is installed.
+# (ie Win32 runs pl2bat after this)
+
+# Doing this here avoids all sort of quoting issues that would come from
+# attempting to write out perl source with literals to generate the arrays and
+# hash.
+my @temps = 'Makefile.PL';
+foreach (glob('scripts/pod*.PL')) {
+    # The various pod*.PL extractors change directory. Doing that with relative
+    # paths in @INC breaks. It seems the lesser of two evils to copy (to avoid)
+    # the chdir doing anything, than to attempt to convert lib paths to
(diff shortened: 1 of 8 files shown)


Leont:
This is a good idea

jkeenan:
Is the new program written in such a way that it could be called in environments other than GH continuous integration (assuming appropriate older `perl` executables and libraries were present)?

khwilliamson:
On 8/9/22 18:14, Tony Cook wrote:
> Add testing dist/ modules against older perls.
> 
> This currently uses the perldocker/perl-testers images rather than the 
> official perl images since perl-testers includes the older versions, 
> while the official perl images don't.
> 
> Unfortunately perl-testers doesn't include threaded perls, but it's an 
> improvement over no testing at all.
> 
> I currently only test against a small subset of perls, but this is 
> easily changed.
> 

Thanks to @pali, Devel::PPPort is tested back to 5.6 something in CI.  I 
b elieve it includes threaded.

tonycoz:
> Is the new program written in such a way that it could be called in environments other than GH continuous integration (assuming appropriate older `perl` executables and libraries were present)?

I think so.

It doesn't try to install dependencies so they would need to be installed first.  I know at least one module depends on `parent` which isn't a `dist/` module, but that's already installed in the `perl-testers` images.

bram-perl:
For the " XSLoader.t: fix local skip() implementation" commit: can you specify in the commit message what was wrong and what it fixed?
(Looking at the code I do see what's wrong/what it fixed but I like that information to be in the commit message so that looking at the code isn't required)

tonycoz:
> For the " XSLoader.t: fix local skip() implementation" commit: can you specify in the commit message what was wrong and what it fixed?

I've updated the commit message for that commit and for the `test dist/ modules against various older perls in CI` commit.

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+
+# get ppport.h
+my $pppdir = test_dist("Devel-PPPort");
+
+my $pppfile = "$pppdir/ppport.h";
+
+-f $pppfile
+  or die "No ppport.h found in $pppdir\n";
+
+opendir my $distdir, "dist" or die;
```
bram-perl: It would be better to include `$!` in the die message..
(It's very unlikely to fail but if it does fail then knowing why will be very useful)

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+my @dists = sort { lc $a cmp lc $b } grep { /^\w/ && $_ ne "Devel-PPPort" } readdir $distdir;
```
bram-perl, line 30: Just an observation (feel free to leave the code as is): skipping `Devel-PPPort` is done in this code;
`version` and `Safe` are excluded later;
That these are treated differently looks a bit inconsistent (but as said I'm fine with leaving it as-is)
One final observation: it doesn't include the reason on why `Devel-PPPort` is excluded, consider adding a comment explaining the why? (The why does become obvious when fully reading the code, it was already tested on line 19)

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+for my $dist (@dists) {
+    test_dist($dist);
+}
+
+sub test_dist {
+    my ($name) = @_;
+
+    print "*** Testing $name ***\n";
+    my $dir = tempdir( CLEANUP => 1);
+    system qq(cp -a "dist/$name/." "$dir/.")
```
bram-perl: why `system EXPR` and not system LIST`? i.e.

    system("cp", "-a", "dist/$name/.",  "$dir/.");

(unlikely to matter tho since `$name`/`$dir` is unlikely to contain special characters that would be treated differently by the shell)

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+      and die "Cannot copy files";
```
bram-perl: consider including `$!` and/or `$?`

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+    chdir $dir
+      or die "Cannot chdir $dir: $!";
+    if ($pppfile) {
+        system "cp", $pppfile, "."
+          and die "Cannot copy $pppfile to .\n";
```
bram-perl, line 58: consider including `$!` and/or `$?`

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+        close $fh;
+    }
+    system "$^X Makefile.PL"
+      and die "$name: Makefile.PL failed\n";
+
+    my $verbose = 0;
+    system "make test TEST_VERBOSE=$verbose"
+      and die "$name: make test failed\n";
+
+    system "make install"
```
bram-perl: Should it include a PREFIX on where to install?
Unless I'm missing something it will be installing in the the default location and not inside a tmpdir?
(That is: the make install may/will change the running system and could - in theory - impact later tests?)
(Also: writing to the default location might require root permissions?)

Porting/test-dist-modules.pl
```diff
@@ -0,0 +1,1087 @@
...
+      and die "$name: make install failed\n";
+
+    chdir $start
+      or die "Cannot return to $start: $!\n";
+
+    $dir;
+}
+
+sub write_testpl {
```
bram-perl, line 152: This needs a bigger explanation of why this is needed/done..

I.e. something like:

    Some dists (IO, threads, threads-shared) include a t/test.pl file when they're released on CPAN.
    In the repo/in core however there is no t/test.pl in the module dir (i.e. dist/IO/t/test.pl does not exist) and instead there it uses the t/test.pl.
    To test these tests on an older perl we need to add a 't/test.pl' that is suitable for using in the dist dir.
    (blead/core t/test.pl can't be used since it's using newer features not available on older perls)

I do think this is very messy but I suppose cleaning this up isn't as easy/straight-forward..
(I also slightly wonder how/who/where the t/test.pl is added to the dist when releasing to CPAN..)

Porting/test-dist-modules.pl
Leont: Agreed

Porting/test-dist-modules.pl
Leont: I don't really think either would help in this case. $! should be empty, and $? will just contain the return value of the `cp`, which isn't informative at all. `cp` will probably print an error message itself that is helpful.

Porting/test-dist-modules.pl
tonycoz: They're installed so later tests are impacted.

This was written for use within the docker images which are setup to allow installation with the default permissions.

I can see this script being extended to allow use outside of that environment (and allowing invocation like `someperl Porting/test-dist-modules.pl Storable` rather than populating from `readdir()`) but I don't aim to do that in this PR.

Porting/test-dist-modules.pl
bram-perl: In my experience that really depends on what happened..
If `cp` encountered an error then yes it will print an informative message; but if something killed the cp command then it will not print a message (and in that case having `$?` in the error is useful since then it's clear that it was killed and with what signal)

(but I'm also ok with leaving as is if it's considered not to be worth it)

Porting/test-dist-modules.pl
bram-perl: Ok, fair enough

PR #20076, opened 2022-08-10
Add what-to-do when push after merge fails

bram-perl:
The instructions on what to do when a `git push` after a `git merge --no-ff` failed were missing/weren't clear.

This PR adds a small section in perlgit.pod on what can be done to preserve the merge.

(Context: @khwilliamson attempted to merge a branch with a merge commit but the merge commit got lost due to a `git pull --rebase` :-( )

### DIFF
--- a/pod/perlgit.pod
+++ b/pod/perlgit.pod
@@ -785,6 +785,27 @@ When describing the merge commit, explain the purpose of the branch, and
 keep in mind that this description will probably be used by the
 eventual release engineer when reviewing the next perldelta document.
 
+If the subsequent I<push> fails then you must be careful on how you I<rebase>.
+If you use
+
+  % git rebase p5p/blead
+
+or
+
+  % git pull --rebase
+
+then your carefully created merge commit will be lost! To avoid this you
+can use:
+
+  % git fetch p5p
+  % git rebase --rebase-merges p5p/blead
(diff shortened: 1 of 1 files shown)


pod/perlgit.pod
```diff
@@ -785,6 +785,23 @@ When describing the merge commit, explain the purpose of the branch, and
...
+
+or
+
+  % git pull --rebase
+
+then your carefully created merge commit will be lost! To avoid this you
+can use:
+
+  % git fetch p5p
+  % git rebase --preserve-merges p5p/blead
```
tonycoz: --preserve-merges was removed in git 2.34, --rebase-merges should be used instead.

https://github.com/git/git/blob/master/Documentation/RelNotes/2.34.0.txt

bram-perl: Thanks;

Looks like it was added in git 2.18 ( https://github.com/git/git/blob/master/Documentation/RelNotes/2.18.0.txt ),
`--preserve-merges` got deprecated in git 2.22 ( https://github.com/git/git/blob/master/Documentation/RelNotes/2.22.0.txt )

Text was updated, reference to `--preserve-merges` kept for those with an older git version.

tonycoz (approved):

PR #20078, opened 2022-08-10
Langinfo.t: Handle systems without LC_ALL

khwilliamson:


jkeenan:
This looks good, but I have to ask:  What platforms which we support lack `LC_ALL`?

jkeenan:
FWIW, branch builds and test passes on OpenBSD-6.9.

khwilliamson:
On 8/10/22 12:39, James E Keenan wrote:
> This looks good, but I have to ask: What platforms which we support lack 
> |LC_ALL|?
> 

I don't remember.  I started this work 2+ years ago, and paused work on 
it 1.5 years ago.  I was doing extensive testing on lots of different 
platforms, and I do remember encountering ones without LC_ALL, but I now 
have no idea which

bram-perl (approved):

PR #20082, opened 2022-08-11
Use optype casting macros

leonerd:
Use these macros where appropriate for consistency and ease of searching. Also some cases can be made less verbose (when the argument is `o`, `kid` or `PL_op`).

Should have no CPAN-visible change


leonerd:
Will also take reviewers' suggestions on whether they should all be squashed into one commit, or kept as-is.

bram-perl (approved):

PR #20083, opened 2022-08-11
win32: Remove trailing backslash from `INST_TOP`

bram-perl:
When `INST_TOP` contains a trailing backslash then things go horribly wrong.

Example (output slightly altered for readability):

	C:\...> gmake INST_TOP=C:\Perl\blead\perl\
	...
	..\miniperl.exe -I..\lib config_sh.PL
		"INST_TOP=C:\Perl\blead\perl\"
		"INST_VER="
		"INST_ARCH="
		"archname=MSWin32-x64-multi-thread"
		"cc=gcc"
		"ld=g++"
		"ccflags= -DWIN32 -DWIN64 ...."
		...
	Use of uninitialized value $opt{"static_ext"} in split at config_sh.PL line 57.
	...
	Can't open -DWIN32: No such file or directory at config_sh.PL line 335.
	...

	..\miniperl.exe -I..\lib ..\configpm --chdir=..
	Use of uninitialized value $t in string eq at ..\configpm line 345.
	...
	written lib/Config.pod
	syntax error at lib/Config_heavy.pl line 165, near "x;"
	Compilation failed in require at ..\configpm line 1144.
	gmake: *** [GNUmakefile:1195: ..\lib\Config.pm] Error 255

-> The trailing backslash in 'INST_TOP' caused the double quote (in
   `miniperl.exe config_sh.PL`) to be escaped which messes up the
   rest of the arguments/the argument parsing leading to the errors.

Avoid the errors by removing the trailing backslash.

(Tested on Windows 10 with GNU Make v4.2.1)

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -539,6 +539,9 @@ CPANDIR		= ..\cpan
 PODDIR		= ..\pod
 HTMLDIR		= .\html
 
+# Strip trailing backslash from INST_TOP
+override INST_TOP := $(INST_TOP:\=)
+
 INST_SCRIPT	= $(INST_TOP)$(INST_VER)\bin
 INST_BIN	= $(INST_SCRIPT)$(INST_ARCH)
 INST_LIB	= $(INST_TOP)$(INST_VER)\lib


khwilliamson:
@xenu, please take a quick look at this

sisyphus:
The same problem exists with win32/Makefile when building with Visual Studio and nmake.
Should the win32/Makefile also be fixed for this ? ( I don't know what that fix would look like.)

BTW, @bram-perl's  proposed patch to win32/GNUmakefile worked fine for me, too.

Cheers,
Rob

bram-perl:
It would be nice to have it (also) fixed in the win32/Makefile but I can't test such a change since I don't have Visual Studio/nmake.

going by the other items in the win32/Makefile the `$(INST_TOP:\=)` **should** work the same but I have no idea if the `override` is needed or what the equivalant is..

An **untested** patch for win32/Makefile: (copied from `incpath` in `CFG_VARS`)
```
diff --git a/win32/Makefile b/win32/Makefile
index 7351a22..ceb8374 100644
--- a/win32/Makefile
+++ b/win32/Makefile
@@ -807,7 +807,7 @@ STATIC_EXT  = Win32CORE
 DYNALOADER     = ..\DynaLoader$(o)
 
 CFG_VARS       =                                       \
-               "INST_TOP=$(INST_TOP)"                  \
+               "INST_TOP=$(INST_TOP:"=\")"             \
                "INST_VER=$(INST_VER)"                  \
                "INST_ARCH=$(INST_ARCH)"                \
                "archname=$(ARCHNAME)"                  \
```

but I suspect that will still result in a double backslash in 
* `INST_SCRIPT`
* `INST_BIN`
* `INST_LIB`
* `INST_ARCHLIB`
* `INST_COREDIR`
* `INST_HTML`

So if the override is not possible then a (bit of an uglier) patch might work: (again, **untested**)
```
diff --git a/win32/Makefile b/win32/Makefile
index 7351a22..3810198 100644
--- a/win32/Makefile
+++ b/win32/Makefile
@@ -396,12 +396,12 @@ CPANDIR           = ..\cpan
 PODDIR         = ..\pod
 HTMLDIR                = .\html
 
-INST_SCRIPT    = $(INST_TOP)$(INST_VER)\bin
+INST_SCRIPT    = $(INST_TOP:=\)$(INST_VER)\bin
 INST_BIN       = $(INST_SCRIPT)$(INST_ARCH)
-INST_LIB       = $(INST_TOP)$(INST_VER)\lib
+INST_LIB       = $(INST_TOP:=\)$(INST_VER)\lib
 INST_ARCHLIB   = $(INST_LIB)$(INST_ARCH)
 INST_COREDIR   = $(INST_ARCHLIB)\CORE
-INST_HTML      = $(INST_TOP)$(INST_VER)\html
+INST_HTML      = $(INST_TOP:=\)$(INST_VER)\html
 
 #
 # Programs to compile, build .lib files and link
@@ -807,7 +807,7 @@ STATIC_EXT  = Win32CORE
 DYNALOADER     = ..\DynaLoader$(o)
 
 CFG_VARS       =                                       \
-               "INST_TOP=$(INST_TOP)"                  \
+               "INST_TOP=$(INST_TOP:"=\")"             \
                "INST_VER=$(INST_VER)"                  \
                "INST_ARCH=$(INST_ARCH)"                \
                "archname=$(ARCHNAME)"                  \

```

xenu:
This has a small drawback of not working with filenames that begin with a space, but I'm sure we can live with that. Doing this more properly would be very complicated.

xenu:
>going by the other items in the win32/Makefile the $(INST_TOP:\=) should work the same

It doesn't, unfortunately. In nmake it removes every occurence of `\` in the variable, as documented [here](https://docs.microsoft.com/en-us/cpp/build/reference/using-an-nmake-macro?view=msvc-170#-macro-substitution).

PR #20084, opened 2022-08-12
Fix sporadic test failure in t/io/eintr.t (+ cleanup)

bram-perl:
The test would sporadically fail; troubleshooting this revealed that
the size of the pipe buffer could vary between pipes which this code
didn't take into account.

What the code did:

1. Create two pipes ($in, $out)
2. Get buffer size of '$out' pipe
3. Do something that should block (using the buffer size)
4. Create two new pipes ($in, $out)
5. Do something that should block (using the buffer size)

The problem: the pipe created in 4. does not always have the same buffer size as the pipe created in 1.

Adding some debug info in the test:

	# pipe buf size: 8193
	ok 4 - read/die: read status
	ok 5 - read/die: close status
	# pipe buf size: 65537
	not ok 6 - print/close: sig handler close status

During test 6 it reused the smaller buffer size and wrote data
to the new pipe. It expected this to block but the buffer of the pipe
used for test 6 is 65536 bytes. End result: the pipe buffer is not
full (only 8194 bytes were written) so the test does not block which
causes the alarm not to trigger which causes the test failure.

Fix all this by reinitialising `$pipe_buf_size` when creating new pipes.

(Observed by @khwilliamson , diagnosed with help of @wolfsage )

+

Apply some small cleanup to t/io/eintr.t

### DIFF
--- a/t/io/eintr.t
+++ b/t/io/eintr.t
@@ -19,16 +19,16 @@ use Config;
 
 my $piped;
 eval {
-	pipe my $in, my $out;
-	$piped = 1;
+    pipe my $in, my $out;
+    $piped = 1;
 };
 if (!$piped) {
-	skip_all('pipe not implemented');
-	exit 0;
+    skip_all('pipe not implemented');
+    exit 0;
 }
 unless (exists  $Config{'d_alarm'}) {
-	skip_all('alarm not implemented');
-	exit 0;
+    skip_all('alarm not implemented');
+    exit 0;
 }
 
 # XXX for some reason the stdio layer doesn't seem to interrupt
@@ -36,8 +36,8 @@ unless (exists  $Config{'d_alarm'}) {
 # hang.
 
 if (exists $ENV{PERLIO} && $ENV{PERLIO} =~ /stdio/  ) {
-	skip_all('stdio not supported for this script');
-	exit 0;
+    skip_all('stdio not supported for this script');
+    exit 0;
 }
 
 # on Win32, alarm() won't interrupt the read/write call.
@@ -55,8 +55,8 @@ if ($^O eq 'VMS' || $^O eq 'MSWin32' || $^O eq 'cygwin' || $^O =~ /freebsd/ || $
      ($^O eq 'darwin' && $osmajmin < 9) ||
     ((int($]*1000) & 1) == 0)
 ) {
-	skip_all('various portability issues');
-	exit 0;
+    skip_all('various portability issues');
+    exit 0;
 }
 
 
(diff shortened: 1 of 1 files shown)


wolfsage (approved):

bram-perl:
(force pushed to fix a small typo in commit message)

t/io/eintr.t
```diff
@@ -55,24 +55,39 @@ if ($^O eq 'VMS' || $^O eq 'MSWin32' || $^O eq 'cygwin' || $^O =~ /freebsd/ || $
...
+    # consistently failing. At exactly 0x100000 it started passing
+    # again. Now we're asking the kernel what the pipe buffer is, and if
+    # that fails, hoping this number is bigger than any pipe buffer.
+    $pipe_buf_size = eval {
+        use Fcntl qw(F_GETPIPE_SZ);
+        fcntl($out, F_GETPIPE_SZ, 0);
+    };
+    if ($@ or not $pipe_buf_size) {
+        $pipe_buf_size = 0xfffff;
+        diag("fcntl F_GETPIPE_SZ failed, falling back to $pipe_buf_size");
```
Leont: I don't think this should be a diag, as it's not an issue that warrents user intervention. It should probably be a note instead.

bram-perl: If you insist on `note()` then I will change it but I'm a bit in doubt;

With `note()` the message is only shown when it's being run in verbose mode..
But if the code gets the pipe buf size wrong then it could lead to test failures and with `note()` the output is then **not** include the report making it harder to troubleshoot/debug (since it may not always happen...)

With `diag()`: if the o/s doesn't support F_GETPIPE_SZ (are there any?) then it will always show the message leading to cluttered output..


(I suppose ideally the diag message would be shown when there was a test failure but that would be slightly trickier..)

Leont: > I suppose ideally the diag message would be shown when there was a test failure

Yeah, I think that's the better way to look at it. One shouldn't diag unless things are failing.

t/io/eintr.t
```diff
@@ -73,18 +73,27 @@ sub fresh_io {
...
+        # but instead these set $! and make fcntl() return undef.
+        fcntl($out, F_GETPIPE_SZ, 0) or die "$!\n";
     };
     if ($@ or not $pipe_buf_size) {
+        my $err = $@;;
+        chomp $err;
         $pipe_buf_size = 0xfffff;
-        diag("fcntl F_GETPIPE_SZ failed, failing back to $pipe_buf_size");
+        $pipe_buf_err = "fcntl F_GETPIPE_SZ failed" . ($err ? " ($err)" : "") .
+                        ", failing back to $pipe_buf_size";
```
khwilliamson: I don't understand "failing back"

demerphq: its a fancy way of saying "defaulting to $pipe_buf_size". :-)

PR #20091, opened 2022-08-15
Fourth batch of locale commits

khwilliamson:
The new commits deal largely with Perl_langinfo(), which is a wrapper for libc nl_langinfo().  That function is missing from some platforms, notably Windows.,  This series of commits extends our  emulation  to better cover the components missing from before them.  The CODESET item detection is much improved on such systems.  This allows us to reliably determine if some item is UTF-8 or not.  New API functions are created to allows XS code to directly access the values, instead of having to go through POSIX.

Redundant implementations of mbtowc, mbrtwoc, strftime, and localeconv are collapsed into one, eliminating  some bugs found in one but not the other


khwilliamson:
Pushed via f7968da43ed14ca0a69dfd121cf9423d35ad5448

PR #20092, opened 2022-08-15
Clean up diag.t a bit

khwilliamson:
These commits remove from the exceptions list messages that no longer show up as errors.
It removes from that list messages that are of the form 'Usage: %s', and 'Function foo not implemented in this version of Perl'  There are a bunch of such exceptions, and each form can be collapsed down to one.  And I believe the messages are self-explanatory enough that they don't need further text in perldiag.

### DIFF
--- a/plan9/plan9.c
+++ b/plan9/plan9.c
@@ -106,7 +106,7 @@ void setprotoent(int a)
 
 void setservent(int a)
 {
-    croak("Function \"setservent\"  not implemented in this version of perl.");
+    croak("Function \"setservent\" not implemented in this version of perl.");
 }
 
 void endnetent()
--- a/t/porting/diag.t
+++ b/t/porting/diag.t
@@ -243,6 +243,12 @@ sub standardize {
   elsif ($name =~ m/^(panic|Usage): /) {
     $name = "$1: \%s";
   }
+  else {
+    $name =~ s/ (*plb:^Function\ ")
+                (\w+)
+                (*pla:"\ not\ implemented\ in\ this\ version\ of\ perl\.)
+              /%s/gx;
+  }
 
   return $name;
 }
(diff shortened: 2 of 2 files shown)


bram-perl (approved):
Looks good to me,

There might be a couple more messages that can be removed but may need some double-checking.
As a trial I adapted the code in t/porting/diag.t and included:
```
diff --git a/t/porting/diag.t b/t/porting/diag.t
index 91a8176..d8aacd0 100644
--- a/t/porting/diag.t
+++ b/t/porting/diag.t
@@ -229,6 +229,19 @@ while (my $file = <$fh>) {
 }
 close $fh or die $!;

+
+foreach my $key (sort { $a cmp $b } keys %entries) {
+    next, if $key =~ m/^#/;
+    next, if $key eq "__CATEGORIES__";
+    next, if $key eq "";
+    next, if $entries{$key}{seen};
+
+    if ($entries{$key}{todo}) {
+        print STDERR "# Unseen todo item: $key\n";
+    }
+}
+
+
 # Standardize messages with variants into the form that appears
 # in perldiag.pod -- useful for things without a diag_listed_as annotation
 sub standardize {
```

Running with that code:
```
# Unseen todo item: %s in regex m/%s/
# Unseen todo item: %sCompilation failed in require
# Unseen todo item: Malformed UTF-8 character (unexpected non-continuation byte 0x%x, immediately after start byte 0x%x)
# Unseen todo item: execl not implemented!
# Unseen todo item: socketpair not implemented!
# Unseen todo item: win32_get_osfhandle() TBD on this platform
# Unseen todo item: win32_open_osfhandle() TBD on this platform
```

And removing those items from `__DATA__` still makes t/porting/diag.t pass

(If the trial code is correct it could be adapted to an actual test (which fails if it finds "unseen" items), when time permits I'll try to do that - unless someone else does it first)

PR #20097, opened 2022-08-16
Fix g++ builds (/ locale.c: Fix trial_locales declaration)

bram-perl:
Commit e57aaf750b3d2461053eff690d0954bc8fb9d7f3 broke builds with g++
because the declaration of `trial_locales` was incorrect. It contained
an extra 'struct' keyword which is wrong.

Errors on g++:

	locale.c:3539:12: error: using typedef-name 'trial_locales_struct' after 'struct'
	     struct trial_locales_struct trial_locales[5];
            ^
	locale.c:3537:7: note: 'trial_locales_struct' has a previous declaration here
	     } trial_locales_struct;

(Build tested with gcc-12, g++-12, gcc-4.9, g++-4.9)

PR #20098, opened 2022-08-16
Add 'g++' build in the GitHub workflow

bram-perl:
Also do a build with 'g++' since it can catch more errors.

(I accidentally broke builds with g++ and that commit got merged
 on blead; smokers then reported build failures with g++ but it would
 have been nicer if I knew this before it got merged.)


bram-perl:
Note: I marked this as DRAFT because the g++ builds on blead need to be fixed **before** this can be merged.

(PR to fix g++ builds: #20097 )

PR #20100, opened 2022-08-16
perl.h - remove visibility attribute on cygwin

demerphq:
There is no attribute visibility on cygwin apparently, and the warnings
seem to overwhelm the cygwin build process.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -400,8 +400,12 @@ Now a no-op.
 #    define __attribute__always_inline__      __attribute__((always_inline))
 #  endif
 #endif
-#if defined(HASATTRIBUTE_VISIBILITY) && !defined(_WIN32)
-/* On Windows, instead of this, we use __declspec(dllexport) and a .def file */
+#if defined(HASATTRIBUTE_VISIBILITY) && !defined(_WIN32) && !defined(__CYGWIN__)
+/* On Windows instead of this, we use __declspec(dllexport) and a .def file
+ * Cygwin works by exporting every global symbol, see the definition of ldflags
+ * near the end of hints/cygwin.sh and the visibility attribute doesn't appear
+ * to control that.
+ */
 #  define __attribute__visibility__(x) __attribute__((visibility(x)))
 #endif
 


bram-perl:
Just an observation: I'm not convinced this is/was the real failure of the cygwin build failures..
Looking at two CI logs:
- https://github.com/Perl/perl5/runs/7845406610?check_suite_focus=true - in this one the cygwin build took 7m47s, looking at the output shows that it's full of the "visibility attribute not supported" warnings
- https://github.com/Perl/perl5/runs/7792317735?check_suite_focus=true - in this one the cygwin build was aborted after 1h53m, log also shows the "visibility attribute not supported" warnings **and** it also shows warnings in cygwin.c. It **appears** it really failed (/got stuck) when running  `PATH=/home/runneradmin/work:.:/usr/bin/ ./miniperl.exe -w -Ilib -Idist/Exporter/lib -MExporter -e '<?>` 

So I'm not sure if it this PR will really fix the CI issues; the visbility attributes warning looks more like a red herring.

**Back to the PR**: fixing(/silencing) the warning is good; the change looks good to me but I'm not really familiar with the visibility attribute/with the build process on cygwin

demerphq:
On Tue, 16 Aug 2022, 13:05 Bram, ***@***.***> wrote:

> Just an observation: I'm not convinced this is/was the real failure of the
> cygwin build failures..
> Looking at two CI logs:
>
>    - https://github.com/Perl/perl5/runs/7845406610?check_suite_focus=true
>    - in this one the cygwin build took 7m47s, looking at the output shows that
>    it's full of the "visibility attribute not supported" warnings
>    - https://github.com/Perl/perl5/runs/7792317735?check_suite_focus=true
>    - in this one the cygwin build was aborted after 1h53m, log is also shows
>    the "visibility attribute not supported" warnings *and* it also shows
>    warnings in cygwin.c. It *appears* it really failed (/got stuck) when
>    running PATH=/home/runneradmin/work:.:/usr/bin/ ./miniperl.exe -w
>    -Ilib -Idist/Exporter/lib -MExporter -e '<?>
>
> So I'm not sure if it this PR will really fix the CI issues; the visbility
> attributes warning looks more like a red herring.
>
> *Back to the PR*: fixing(/silencing) the warning is good; the change
> looks good to me but I'm not really familiar with the visibility
> attribute/with the build process on cygwin
>

The warnings drown out any ability to debug further so let's get rid of
them first.

Yves

>

bram-perl:
>     * https://github.com/Perl/perl5/runs/7792317735?check_suite_focus=true - in this one the cygwin build was aborted after 1h53m, log also shows the "visibility attribute not supported" warnings **and** it also shows warnings in cygwin.c. It **appears** it really failed (/got stuck) when running  `PATH=/home/runneradmin/work:.:/usr/bin/ ./miniperl.exe -w -Ilib -Idist/Exporter/lib -MExporter -e '<?>`

Taking another look at it: it appears there is a "show timestamp" button in the log viewer,
Using it:
```
Thu, 11 Aug 2022 17:27:37 GMT Run path %GITHUB_WORKSPACE%\cygwin\bin;%GITHUB_WORKSPACE%\cygwin\usr\bin
...
Thu, 11 Aug 2022 17:28:41 GMT PATH=/home/runneradmin/work:.:/usr/bin/ ./miniperl.exe -w -Ilib -Idist/Exporter/lib -MExporter -e '<?>' || sh -c 'echo >&2 Failed to build miniperl.  Please run make minitest; exit 1'
Thu, 11 Aug 2022 19:21:27 GMT Terminate batch job (Y/N)? 
```

In that particular build it really is/was the `miniperl` that hang for 1h29m

(PS: sorry for hijacking this PR for a bit)

tonycoz:
The cygwin build failures were caused by a still unfixed bug in cygwin.

See https://sourceware.org/pipermail/cygwin/2022-August/252043.html

The fix here for the visibility warnings is the same as one I was considering, so I think it's good.

@khwilliamson applied 5ee7770b70fd48652f905014e47b07eb031ed5bc to workaround the cygwin bug until it's fixed.

tonycoz:
> The fix here for the visibility warnings is the same as one I was considering, so I think it's good.

Thinking further here - the WIN32 exception has a comment describing why it is an exception, but this PR doesn't add one for Cygwin.

Cygwin works by exporting every global symbol (see the definition of `ldflags` near the end of `hints/cygwin.sh`), and the visibility attribute doesn't appear to control that.

demerphq:
On Wed, 17 Aug 2022 at 02:59, Tony Cook ***@***.***> wrote:
>
> The fix here for the visibility warnings is the same as one I was considering, so I think it's good.
>
> Thinking further here - the WIN32 exception has a comment describing why it is an exception, but this PR doesn't add one for Cygwin.
>
> Cygwin works by exporting every global symbol (see the definition of ldflags near the end of hints/cygwin.sh), and the visibility attribute doesn't appear to control that.

Ok, ill add a comment.

Yves
-- 

demerphq:
@tonycoz comment updated.

PR #20103, opened 2022-08-17
only clear the stream error state in readline() for a good nextargv()

### LINKED ISSUE #20060: IO::Handle::error() can lose read errors
As reported by Ian Jackson in https://bugs.debian.org/1016369 IO::Handle::error() can report 0 when there was an error on the filehandle.

A test case is

`perl -MIO::Handle -e 'open X, "<", "." or die $!; $_ = <X>; printf "%s %s\n", X->error(), $!;'`

which returns "0 Is a directory" for me on Linux although all reads on the filehandle fail with EISDIR.

I've debugged this a bit and it looks like the error gets cleared in Perl_do_readline() around pp_hot.c:3295 or so with a PerlIO_clearerr() call.  I think this is related to reading past EOF or something like that?

While trying to treat a directory as a regular file is something of a corner case (and the result is rather platform dependent), Ian argues that this is an indication of a more general problem where EIO and similar errors can go unnoticed if user code is relying on the IO::Handle::error() method to report them. Of course, creating simple test cases for such errors is difficult if not impossible.

I see #12782 is very similar but hasn't really received much attention. Is this something that should at least be documented as a caveat if it's not fixable?

```
-----------------------------------------------------------------
---
Flags:
    category=library
    severity=medium
    module=IO::Handle
---
Site configuration information for perl 5.37.3:

Configured by ntyni at Sat Aug  6 17:54:25 BST 2022.

Summary of my perl5 (revision 5 version 37 subversion 3) configuration:
  Commit id: 12d173c94b050c244dbb4e2162e29834a9ac3aff
  Platform:
    osname=linux
    osvers=5.17.0-1-amd64
    archname=x86_64-linux
    uname='linux carme 5.17.0-1-amd64 #1 smp preempt debian 5.17.3-1 (2022-04-18) x86_64 gnulinux '
    config_args='-des -Dusedevel'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='cc'
    optimize='-O2'
    ccversion=''
    gccversion='11.2.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    libc=libc-2.33.so
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.33'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'


---
@INC for perl 5.37.3:
    lib
    /usr/local/lib/perl5/site_perl/5.37.3/x86_64-linux
    /usr/local/lib/perl5/site_perl/5.37.3
    /usr/local/lib/perl5/5.37.3/x86_64-linux
    /usr/local/lib/perl5/5.37.3

---
Environment for perl 5.37.3:
    HOME=/home/ntyni
    LANG=C
    LANGUAGE (unset)
    LC_CTYPE=fi_FI.UTF-8
    LD_LIBRARY_PATH (unset)
    LOGDIR (unset)
    PATH=/usr/local/bin:/usr/bin:/bin:/usr/games
    PERL_BADLANG (unset)
    SHELL=/bin/zsh

### PR DESCRIPTION
**Pull request by tonycoz**
This would previously clear the stream error state in any case
where sv_gets() failed and the error state was set.

Since this now leaves the error state on the stream, the close()
calls in the test updated by this commit would fail, changing its
output.  Since the result of those closes didn't seem related
to the purpose of the test, I changed it not throw an error on
either close() failing.

Fixes #20060

### DIFF
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -3334,14 +3334,19 @@ Perl_do_readline(pTHX)
                 || SNARF_EOF(gimme, PL_rs, io, sv)
                 || PerlIO_error(fp)))
         {
-            PerlIO_clearerr(fp);
             if (IoFLAGS(io) & IOf_ARGV) {
                 fp = nextargv(PL_last_in_gv, PL_op->op_flags & OPf_SPECIAL);
-                if (fp)
+                if (fp) {
                     continue;
+                }
                 (void)do_close(PL_last_in_gv, FALSE);
             }
             else if (type == OP_GLOB) {
+                /* clear any errors here so we only fail on the pclose()
+                   failing, which should only happen on the child
+                   failing
+                */
+                PerlIO_clearerr(fp);
                 if (!do_close(PL_last_in_gv, FALSE)) {
                     Perl_ck_warner(aTHX_ packWARN(WARN_GLOB),
                                    "glob failed (child exited with status %d%s)",
--- a/t/lib/warnings/pp_hot
+++ b/t/lib/warnings/pp_hot
@@ -237,8 +237,8 @@ $a = <FOO> ;
 use warnings 'io' ;
 $a = <FOO> ;
 $a = <FH> ;
-close (FH) or die $! ;
-close (FOO) or die $! ;
+close (FH);
+close (FOO);
 unlink $file ;
 EXPECT
 Filehandle FH opened only for output at - line 5.
--- a/t/op/readline.t
+++ b/t/op/readline.t
@@ -6,7 +6,7 @@ BEGIN {
     set_up_inc('../lib');
 }
 
-plan tests => 32;
+plan tests => 36;
 
 # [perl #19566]: sv_gets writes directly to its argument via
 # TARG. Test that we respect SvREADONLY.
@@ -287,6 +287,18 @@ is ${^LAST_FH}, undef, '${^LAST_FH} after readline undef';
     like $w, qr/^readline\(\) on unopened filehandle y at .*\n(?x:
                 )Undefined value assigned to typeglob at .*\n\z/,
         '[perl #123790] *x=<y> used to fail an assertion';
+}
+
+SKIP:
+{
+    skip_without_dynamic_extension("IO", 4);
+    my $tmpfile = tempfile();
+    open my $fh, ">", $tmpfile
+        or die "Cannot open $tmpfile: $!";
+    ok(!$fh->error, "no error before we try to read");
+    ok(!<$fh>, "fail to readline file opened for write");
+    ok($fh->error, "error after trying to readline file opened for write");
+    ok(!close($fh), "closing the file should fail");
 }
 
 __DATA__

### CODE REVIEW

**Comment by bram-perl** [comment]
The code change looks good but what I'm missing:

* a perldelta entry (I'm not sure when these are normally added): perl's own test suite was "impacted" by the bug fix so it's possible(/likely?) other code out there will be too; so I think it certainly should be mentioned in perldelta
* a test for the new behavior (i.e. that the error doesn't get cleared when it shouldn't)
* a test for the old behavior (i.e. that the error gets cleared when it should)


For that last one:  
What I tried was to completely remove `Perl_clearerr()` call in `Perl_do_readline` and then tried to come up with a test case where it would behave differently.
So far I've not been successful.. All seem to work as expected...
Digging a bit deeper: it **appears** that  `nextargv()` will call a `do_open`-variant and these all **appear** to call `S_openn_cleanup` which **appears** to call `PerlIO_clearerr()` when the open succeeded.. 
In other words: I don't know when the `Perl_clearerr()` in `Perl_do_readline` is actually needed.. Can anyone come up with a test case?

**Comment by tonycoz** [comment]
> * a perldelta entry (I'm not sure when these are normally added): perl's own test suite was "impacted" by the bug fix so it's possible(/likely?) other code out there will be too; so I think it certainly should be mentioned in perldelta

I never include perldelta updates in PRs - they're almost guaranteed to conflict with other updates to perldelta.  But I could talk about it here, maybe something like:

```
(under incompatible changes)
=head2 readline() no longer clears the stream error flag

C<readline()>, also spelled C<< <> >>, would clear the handle's error flag after an error occurred on the stream.

The error flag is now only cleared when processing starts on a new file when reading the C<ARGV> stream.

Since the error flag is no longer cleared calling `close()` on the stream may fail and if the stream was not explicitly
closed, the implicit close of the stream may produce a warning.

```

> * a test for the new behavior (i.e. that the error doesn't get cleared when it shouldn't)

This is reasonable, I'll look at it.

> * a test for the old behavior (i.e. that the error gets cleared when it should)

I don't see a way to do that, since the error is only cleared for nextargv() we'd need to populate `@ARGV` with a file that can be opened successfully for reading, but then produces an error (not eof) when read.

**Comment by bram-perl** [comment]
> > * a test for the old behavior (i.e. that the error gets cleared when it should)
> 
> I don't see a way to do that, since the error is only cleared for nextargv() we'd need to populate `@ARGV` with a file that can be opened successfully for reading, but then produces an error (not eof) when read.

Doing some more testing:

On my system reading from /dev/ttyS3: open succeeds, read fails:

        $ strace perl -ne '' /dev/ttyS3
        ...
        open("/dev/ttyS3", O_RDONLY)            = 3
        ioctl(3, SNDCTL_TMR_TIMEBASE or SNDRV_TIMER_IOCTL_NEXT_DEVICE or TCGETS, 0x7ffce37139c0) = -1 EIO (Input/output error)
        lseek(3, 0, SEEK_CUR)                   = -1 ESPIPE (Illegal seek)
        fstat(3, {st_mode=S_IFCHR|0660, st_rdev=makedev(4, 67), ...}) = 0
        fcntl(3, F_SETFD, FD_CLOEXEC)           = 0
        read(3, 0x633640, 8192)                 = -1 EIO (Input/output error)
        close(3)                                = 0

So open succeeds, read fails so that is a good test case.


**Testing with my own fh:**

old perl:

        $ perl  -MIO::Handle -wle 'open my $fh, "<", "/dev/ttyS3" or die "open failed: $!";  <$fh> or warn "Readline failed: $!";  print $fh->error();'
        Readline failed: Input/output error at -e line 1.
        0

your PR:

        $ ./perl -Ilib -MIO::Handle -wle 'open my $fh, "<", "/dev/ttyS3" or die "open failed: $!";  <$fh> or warn "Readline failed: $!";  print $fh->error();'
        Readline failed: Input/output error at -e line 1.
        1

blead *without* `PerlIO_clearerr()` in `Perl_do_readline`:

        $ ./perl -Ilib -MIO::Handle -wle 'open my $fh, "<", "/dev/ttyS3" or die "open failed: $!";  <$fh> or warn "Readline failed: $!";  print $fh->error();'
        Readline failed: Input/output error at -e line 1.
        1


So that looks good.


**Testing with ARGV:**


old perl:

        $ perl -MIO::Handle -wle '<ARGV> or warn "Readline failed: $!";  print ARGV->error();' /dev/ttyS3
        Readline failed: Input/output error at -e line 1.
        -1

your PR:

        $ ./perl -Ilib -MIO::Handle -wle '<ARGV> or warn "Readline failed: $!";  print ARGV->error();' /dev/ttyS3
        Readline failed: Input/output error at -e line 1.
        -1

blead *without* `PerlIO_clearerr()` in `Perl_do_readline`:

        $ ./perl -Ilib -MIO::Handle -wle '<ARGV> or warn "Readline failed: $!";  print ARGV->error();' /dev/ttyS3
        Readline failed: Input/output error at -e line 1.
        -1

No change in behaviour.

`ARGV->error()` returns -1 because ARGV is closed (`Perl_do_readline` closes it on error`).

Which really means: as far as I can tell you can't check if reading from ARGV failed by using `ARGV->error()` since that will either return 0 (no error) or -1 (closed/invalid handle) never 1.

So again I'm at the state where I don't see when `PerlIO_clearerr()` in `Perl_do_readline` does something useful/something observable :/

**TLDR:**

I (still) question if/when doing a `PerlIO_clearerr()` inside `if (fp) {` is needed.

**Comment by bram-perl** [comment]
After writing the previous message I started wondering about something else: why does `close` return an error?

<details>
 <summary>Old message</summary>

Again testing with /dev/ttyS3 (where open succeeds but read fails with EIO):

        $ cat tt.pl
        open my $fh, "<", "/dev/ttyS3" or die "open failed: $!";
        <$fh> or warn "Readline failed: $!";
        close $fh or warn "Close failed: $!";

        $ ./perl tt.pl
        Readline failed: Input/output error at tt.pl line 4.
        Close failed: Input/output error at tt.pl line 5.

So according to that *I* would expect that `close` returned an error;
So let's check it using `strace`:

        $ strace ./perl tt.pl
        ...
        open("/dev/ttyS3", O_RDONLY|O_CLOEXEC)  = 3
        ioctl(3, SNDCTL_TMR_TIMEBASE or SNDRV_TIMER_IOCTL_NEXT_DEVICE or TCGETS, 0x7ffd96ae9050) = -1 EIO (Input/output error)
        lseek(3, 0, SEEK_CUR)                   = -1 ESPIPE (Illegal seek)
        fstat(3, {st_mode=S_IFCHR|0660, st_rdev=makedev(4, 67), ...}) = 0
        read(3, 0x986520, 8192)                 = -1 EIO (Input/output error)
        write(2, "Readline failed: Input/output er"..., 53) = 53
        close(3)                                = 0

The `close` system call returned 0, i.e. no error.
So why are we reporting an error on close?

From `perldoc -f close`:

        Closes the file or pipe associated with the filehandle, flushes
        the IO buffers, and closes the system file descriptor. Returns
        true if those operations succeed and if no error was reported by
        any PerlIO layer.

=>
- 'flushes the IO buffers': it was a read fh so there are no buffers to flush
- 'closes the system file descriptor': closing the file descriptor succeeded (as shown by strace)
- 'no error was reported by any PerlIO layer': that could mean anything I guess..
</details>

The question: would one expect that a failed `read` results in a failed `close`?
Especially when the actual close syscall succeeded?

**Update** doing some more testing shows the same behavior when writing to a file.
That is: when the write syscall returns an error then Perl's close will report that error even when the actual close syscall succeeded.

So applying the same behavior to read looks OK.

<details>
  <summary>Write example</summary>

Using a tmpfs of 1M:
```
$ mkdir /tmp/bla
$ mount -o size=1M  -t tmpfs tmpfs /tmp/bla
$ perl -wle 'open my $fh, ">", "/tmp/bla/foo"; print $fh "a" x (2 ** 19) or warn "print 1 failed"; print $fh "a" x (2 ** 19) or warn "print 2 failed"; close $fh or warn "close failed: $!"'
close failed: No space left on device at -e line 1.

$ strace perl -wle 'open my $fh, ">", "/tmp/bla/foo"; print $fh "a" x (2 ** 19) or warn "print 1 failed"; print $fh "a" x (2 ** 19) or warn "print 2 failed"; close $fh or warn "close failed: $!"'
...
write(3, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"..., 8192) = 8192
write(3, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"..., 8192) = 8192
write(3, "a\n", 2)                      = -1 ENOSPC (No space left on device)
close(3)                                = 0
```

Note:
- at the system level: errors reported on the last `write`, no error on `close`;
- at the 'Perl'-level no errors were reported on `print`, only on `close`;

</details>

**Review by bram-perl — APPROVED** [review]
(I still wonder if/when `PerlIO_clearerr` is needed but that shouldn't hold back this PR)

**Comment by tonycoz** [comment]
Looking at the flow through that code, do_open6() and nextargv() it looks like the clearerr() is only needed for globs:

- nextargv() doesn't call do_close(), it just calls do_open6(), which calls openn_setup() which `PerlIO_close()`s the input stream instead of calling do_close() or io_close().  In nextargv() new PerlIO objects are created, presumably with the error flag cleared, so the clearerr() I moved has no effect.
- the glob closing code does call do_close(), but always treats such a failure as a child process exit failure, so it probably should keep on doing clearerr()
- the remaining cases are for user opened files/sockets/pipes, in which case we should preserve the error.

Note that the implicit close warning is only ever produced for an IO object going out of scope, not for a file reopened:
```
open FH, "<", "foo";
# do something to put $fh into error
open FH, "<", "bar"; # stream was in error state, but no warning
```
I've re-worked it to only clear the error for glob(), which I think is correct(-ish), but should only have an effect for `miniperl`.

demerphq:
On Thu, 18 Aug 2022 at 02:39, Tony Cook ***@***.***> wrote:

>
>    - a perldelta entry (I'm not sure when these are normally added):
>    perl's own test suite was "impacted" by the bug fix so it's
>    possible(/likely?) other code out there will be too; so I think it
>    certainly should be mentioned in perldelta
>
> I never include perldelta updates in PRs - they're almost guaranteed to
> conflict with other updates to perldelta.
>
IMO it would be nice to have a solution to that. If perldeltas changes were
added as new filelets and assembled later it might make people more
inclined to add new perldeltas. That would avoid any conflicts and make it
easier to see what changes came from what PR's.

Eg, instead of changing perldelta you create perldelta_20103.pod, it
contains the delta changes from PR #20103, this one. When we do the release
process we construct perldelta from the pieces.

cheers,
Yves



-- 

bram-perl:
> I've re-worked it to only clear the error for glob(), which I think is correct(-ish), but should only have an effect for `miniperl`.

Do you have an example of when it's needed?
(I did try several things - with miniperl - but so far I've not yet found a case where it *appears* to make a difference, so I'm wondering what I'm missing..)

tonycoz:
> Do you have an example of when it's needed?

 I don't see a specific case - it would have to be read() failing before sv_gets() has read anything, and the cases where read() fails where we don't otherwise handle it (EINTR can happen and is handled).

On POSIX systems I don't think it can happen unless something else is badly wrong, none of EAGAIN, EFAULT, EBADF, EIO, EISDIR should happen.

It might be different on non-POSIX systems.

PR #20106, opened 2022-08-17
Perldelta for GH#19943 (OP_PADSV_STORE)

richardleach:
The wording has been left generic, since other such optimizations
are in the PR queue and could just be tacked on the end later.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -100,7 +100,9 @@ There may well be none in a stable release.
 
 =item *
 
-XXX
+Additional optree optimizations for common OP patterns. For example, multiple
+simple OPs replaced by a single streamlined OP, so as to be more efficient at
+runtime. L<[GH #19943]|https://github.com/Perl/perl5/pull/19943>.
 
 =back
 


pod/perldelta.pod
```diff
@@ -91,16 +91,12 @@ as an updated module in the L</Modules and Pragmata> section.
 
 =head1 Performance Enhancements
 
-XXX Changes which enhance performance without changing behaviour go here.
-There may well be none in a stable release.
-
-[ List each enhancement as an =item entry ]
-
```
leonerd: Usually we leave these bits in for individual updates. Deleting that chunk is part of the release process.

pod/perldelta.pod
```diff
@@ -91,16 +91,12 @@ as an updated module in the L</Modules and Pragmata> section.
...
 =over 4
 
 =item *
 
-XXX
+Additional optree optimizations for common OP patterns.
+L<[GH #19943]|https://github.com/Perl/perl5/pull/19943>.
```
leonerd: Might be worth expanding on that a little more, just to say something like "common patterns of multiple small ops are combined into fewer larger ones, so as to be more efficient at runtime".

pod/perldelta.pod
richardleach: Ok, restored it

pod/perldelta.pod
richardleach: I've elaborated.

leonerd (approved):
LGTM

PR #20107, opened 2022-08-17
Add a PadnameREFCNT_inc() macro

leonerd:
Implemented as a static inline function call, so that it can return the padname pointer itself.  This would allow use in expressions such as

```c
ptr->field = PadnameREFCNT_inc(pn);
```

That makes it similar to the familiar `SvREFCNT_inc()` macro.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -3430,6 +3430,13 @@ Perl_cop_file_avn(pTHX_ const COP *cop) {
 
 #endif
 
+PERL_STATIC_INLINE PADNAME *
+Perl_padname_refcnt_inc(PADNAME *pn)
+{
+    PadnameREFCNT(pn)++;
+    return pn;
+}
+
 /*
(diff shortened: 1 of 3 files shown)


ilmari (approved):

inline.h
```diff
@@ -3430,6 +3430,13 @@ Perl_cop_file_avn(pTHX_ const COP *cop) {
 
 #endif
 
+PERL_STATIC_INLINE PADNAME *
+Perl_padname_refcnt_inc(PADNAME *pn)
```
leonerd, line 3434: Sudden wondering whether this should be named `S_` instead of `Perl_`, as it's only here to make one of the macros work.

leonerd: Eh; most of the rest of them are `Perl_...` and those seem fine. I'll leave it as is.

PR #20111, opened 2022-08-17
win32: check if CCHOME is correctly set

bram-perl:
When building on Windows one must provide a value for 'CCHOME'.
The value of 'CCHOME' is used to set the value of 'CCLIBDIR' and 'CCINCDIR'.
(These eventually become `$Config{libpth}` and `$Config{incpath}`.)

When 'CCHOME' is incorrectly set the build mostly works; what fails
is building cpan/Win32 and this then fails with non-obvious errors such as:

	Warning (mostly harmless): No library found for -luserenv
	Warning (mostly harmless): No library found for -lwinhttp
	...
	g++ Win32.def -o ....
	C:/.../ld.exe: ... undefined reference to `WinHttpCrackUrl'
	...
	gmake[1]: *** [Makefile:472: ..\..\lib\auto\Win32\Win32.dll] Error 1
	gmake[1]: Leaving directory 'C:/Perl/perl5/cpan/Win32'
	Unsuccessful make(cpan/Win32): code=512 at ..\make_ext.pl line 584.

Make this a bit more obvious by checking if the path(s) specified in 'CCLIBDIR' and 'CCINCDIR' actually exist.

If these do not exist then stop the build process at an early stage with a clear(er) error.

Example error:

	..\miniperl.exe -I..\lib config_sh.PL ...
	'CCLIBDIR' contains the following non-existing paths:
	        C:\MinGW\lib
	        C:\MinGW\x86_64-w64-mingw32\lib
	        C:\MinGW\lib\gcc\x86_64-w64-mingw32\8.3.0
	Did you provide a correct value for the 'CCHOME' option?

	(This check can be skipped by using the SKIP_CCHOME_CHECK=1 option)
	gmake: *** [GNUmakefile:1176: ..\config.sh] Error 1

Some notes:
- This check can be skipped/overridden by adding 'SKIP_CCHOME_CHECK=1'
  on the make cmd line. (I added this because I don't know if there are
  valid use cases for having a non-existing path in 'libpth'/'incpath')
- My first idea was to do this in the Makefile/GNUmakefile at a very early
  stage but that is more complex to do and the resulting code is messy/ugly.
  So instead I opted to do it in 'config_sh.PL' since this is the
  first(/only) place that uses these vars.
- I (only) tested on Windows 10 using GNU Make and gcc

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -1111,7 +1111,8 @@ CFG_VARS	=					\
 		"LINK_FLAGS=$(subst ",\",$(LINK_FLAGS))"\
 		"optimize=$(subst ",\",$(OPTIMIZE))"	\
 		"ARCHPREFIX=$(ARCHPREFIX)"		\
-		"WIN64=$(WIN64)"
+		"WIN64=$(WIN64)"                        \
+		"SKIP_CCHOME_CHECK=$(SKIP_CCHOME_CHECK)"
 
 #
 # Top targets
--- a/win32/Makefile
+++ b/win32/Makefile
@@ -834,7 +834,8 @@ CFG_VARS	=					\
 		"default_inc_excludes_dot=$(DEFAULT_INC_EXCLUDES_DOT)"	\
 		"LINK_FLAGS=$(LINK_FLAGS:"=\")"		\
 		"optimize=$(OPTIMIZE:"=\")"		\
-		"WIN64=$(WIN64)"
+		"WIN64=$(WIN64)"                        \
+		"SKIP_CCHOME_CHECK=$(SKIP_CCHOME_CHECK)"
 
 #
 # Top targets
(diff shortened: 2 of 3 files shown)


tonycoz (approved):

sisyphus:
So ... here's an interesting little ramification of this PR that I hadn't foreseen.
When doing my scripted builds of perl I like to add the paths to various 3rd party libraries (not needed to build perl itself) via the EXTRALIBDIRS option.
That way, perl will find those libraries by default when I subsequently build modules that depend upon them
Sometimes I might build perl before I've actually built those 3rd party libraries - and a path I've specified in EXTRALIBDIRS does not yet exist.
When that happens, I now get (during the 'gmake' phase) eg :
```
'CCLIBDIR' contains the following non-existing paths:
        C:\_64\msys_1220\1.0\local\lib
Did you provide a correct value for the 'CCHOME' option?

(This check can be skipped by using the SKIP_CCHOME_CHECK=1 option)
make: *** [GNUmakefile:1176: ..\config.sh] Error 1
```
@bram-perl, congrats for having the foresight to provide the SKIP_CCHOME_CHECK=1 option. It's not something that I would inevitably have thought of providing.
And I just wanted to let you know that this option has already proved its usefulness .... though I'm not entirely sure that mine is the purpose for which you envisaged it ;-)

Cheers,
Rob

PR #20113, opened 2022-08-18
fix t/porting/authors.t detection of merge commits

haarg:


demerphq (approved):

demerphq:
Ive hit approve, but i question this logic at a different level. I would like us to just do a full code scan. I am almost certain that doing these checks incrementally misses stuff. See https://github.com/Perl/perl5/pull/20070 where I want to delete this logic entirely.

bram-perl:
The change looks ok but the commit message could do with some improving.
I.e. include what was wrong with the old check; (I know what was wrong but I would prefer that the reason is clear from the commit message so that it's recorded in the history)

demerphq:
@haarg id like to rebase on top of this once it is merged, could you fix the commit message like @bram-perl requested so we can get it merged?

bram-perl:
@demerphq if you don't want to wait on @haarg then I suppose you could merge it on blead and alter the commit message before pushing.

Commit message I would use would be something like:
```
`git rev-parse HEAD^2` was used to detect if HEAD was a merge commit.
When HEAD is not a merge commit then HEAD^2 produces an error;
When HEAD is a merge commit then HEAD^2 points to the second parent.

The problem was that `git rev-parse "HEAD^2"` also produced output on STDOUT
which caused `$revision_range` to be set to something bogus.

Example:
    $ git rev-parse HEAD^2 2>/dev/null ; echo $?
    HEAD^2
    128
    $ git log HEAD^1..HEAD^2
    fatal: ambiguous argument 'HEAD^1..HEAD^2': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'

The fix is to use --verify which doesn't give output on STDOUT for something invalid.
It does print an error on STDERR (for invalid things) which can be silenced using -q
    $ git rev-parse --verify HEAD^2
    fatal: Needed a single revision

    $ git rev-parse -q --verify HEAD^2
    $

For an actual merge commit:
    $ git rev-parse -q --verify e91a97a^2
    bee6601bbf238b5366ce11cd41f6b095bfeebfd8

(i.e. working as expected)
```

demerphq:
merged as 16dd3f70cc16005d5af7146385733a7c945fb67e

PR #20115, opened 2022-08-18
diag.t: Subdivide exception list

khwilliamson:
Some of the entries in this list of diagnostics are ones that people
just haven't gotten around to documenting.

Patches welcome!

But a very few are ones that, in my opinion, don't ever need to be
documented because the text of the diganostic is sufficiently
explanatory in and of itself.

This commit revises the comments explaining the exception list, and
moves the ones I think are sufficiently self-explanatory to the front.

REVIEWERS: See if you agree that these don't need more explanations

### DIFF
--- a/t/porting/diag.t
+++ b/t/porting/diag.t
@@ -519,19 +519,34 @@ sub check_message {
     return $ret;
 }
 
-# Lists all missing things as of the inauguration of this script, so we
-# don't have to go from "meh" to perfect all at once.
-# 
-# PLEASE DO NOT ADD TO THIS LIST.  Instead, write an entry in
-# pod/perldiag.pod for your new (warning|error).  Nevertheless,
-# listing exceptions here when this script is not smart enough
-# to recognize the messages is not so bad, as long as there are
-# entries in perldiag.
-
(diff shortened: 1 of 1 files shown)


t/porting/diag.t
```diff
@@ -510,19 +510,34 @@ sub check_message {
...
+# blank line.
+#
+# The first set are entries whose text we think is fully explanatory and don't
+# need further elaboration in perldiag.  This set should consist of very few
+# entries.
+#
+# The second set, after the blank line, consists of TODO entries.  (There are
+# actually two subsets described below.)  This list should basically be just
+# those entries that otherwise would have generated an error upon inauguration
+# of this script, so we didn't have to go from "meh" to perfect all at once.
```
jkeenan: In comments and documentation, I generally prefer `s/script/program/` (2 lines farther on as well) -- but I recognize that here you're preserving existing language.

bram-perl (approved):
LGTM, it should be rebased tho (and merge commit dropped);

t/porting/diag.t
khwilliamson: Changed

PR #20117, opened 2022-08-18
Clean up cygwin.c some

khwilliamson:
This series of commits simplifies cygwin.c by, for example, folding nearly identical functions into a single one.  The use of locales to convert between UTF 8/16 is eliminated; instead using the core functions that do that explicitly.

However **The changes are not executed in our test suite**.  Are they actually used by cygwin?  @sisyphus

### DIFF
--- a/cygwin/cygwin.c
+++ b/cygwin/cygwin.c
@@ -262,19 +262,33 @@ XS(XS_Cygwin_winpid_to_pid)
     XSRETURN_UNDEF;
 }
 
-XS(XS_Cygwin_win_to_posix_path)
+/* The conversion between Posix and Windows paths is essentially the same in
+ * either direction, so a common function is used, with which direction passed
+ * in.
+ *
+ * These numbers are chosen so can be or'd with absolute flag to get 0..3 */
+typedef enum {
+    to_posix = 0,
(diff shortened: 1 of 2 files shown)


demerphq:
@sisyphus do you have any comments on this ticket?

demerphq (approved):

sisyphus:
>@sisyphus do you have any comments on this ticket?

Nothing meaningful to add.
Cygwin is not my goto build, and I can see no reason to refrain from merging this.

Because I'm on Windows (and have Cygwin installed), I've started routinely performing x64   "-des", "-Duselongdouble" and "-Dusequadmath" builds of the latest perl devel release every month.
I do this mainly so I can then check that my math modules are behaving as intended on that platform.
If there's some reason that any of these changes should not be made, I guess I've therefore got some chance of uncovering the problem ... but likely only after it has been introduced into a devel release ;-)

Cheers,
Rob

demerphq:
Merging. Lets see what happens.

khwilliamson:
An assert(0) in the code does not fail, and makes me think this code is completely obsolete.  I have not had a chance to look at the cygwin source to find out under what conditions it gets called, if any.

PR #20118, opened 2022-08-18
makedef: Export certain symbols

### LINKED ISSUE #20090: t/porting/globvar.t: fcbef0e88b causes test failures on OpenBSD unthreaded builds
In a [smoke test report on OpenBSD-6.9](https://perl5.test-smoke.org/report/5021519), I observe new and unexpected test failures in `t/porting/globvar.t`.  These failure appear to be happening on **unthreaded** builds only.
```
# Failed test 7 - 'PL_numeric_name' is defined, but we do not export it at porting/globvar.t line 81
# Failed test 8 - 'PL_numeric_radix_sv' is defined, but we do not export it at porting/globvar.t line 81
# Failed test 9 - 'PL_numeric_standard' is defined, but we do not export it at porting/globvar.t line 81
# Failed test 10 - 'PL_numeric_underlying' is defined, but we do not export it at porting/globvar.t line 81
# Failed test 11 - 'PL_numeric_underlying_is_standard' is defined, but we do not export it at porting/globvar.t line 81
[2022-08-14 13:06:50+0000] [known failed test] ../t/porting/globvar.t......................................
*** Error 5 in /home/jkeenan/p5smoke/perl-current (makefile:842 'test_harness')
```
In my OpenBSD smoke testing I configure like this:
```
-Dusedevel -Accflags="-DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE -fno-delete-null-pointer-checks"
```
... to better simulate the `ccflags` setting of the platform's "vendor" `perl`. (@afresh1)  The smoke-testing configuration then creates a matrix by adding `-Duseithreads` and/or `-DDEBUGGING`.  I can reproduce the test failures locally.

Bisecting with this invocation:
```
perl Porting/bisect.pl \
-Accflags="-DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE -fno-delete-null-pointer-checks" \
--start=1b5a220440d875d11b90b09072c10e1de8c9420a \
--end=bf2a3dae9f4f828fd1f2f8aaf4769f96520c9552 \
--target=t/porting/globvar.t
```
... points to this commit:
```
fcbef0e88b4eade993ab05b0c51334391a265233 is the first bad commit
commit fcbef0e88b4eade993ab05b0c51334391a265233
Author: Karl Williamson <khw@cpan.org>
Date:   Mon Aug 1 16:42:31 2022 -0600
CommitDate: Tue Aug 9 18:05:59 2022 -0600

    intrpvar.h: Fix when certain locale symbols get defined
    
    Certain interpreter variables are needed only in certain Configurations.
    This commit removes some that aren't needed everywhere to be generated
    only where they are needed, and generates some that weren't being used
    that are or will be needed everywhere by follow-up commits.

 intrpvar.h | 12 ++++--------
 1 file changed, 4 insertions(+), 8 deletions(-)
```

khwilliamson:
This fixes #20090

These symbols were changed by fcbef0e to be defined in more
Configurations, but as @bram-perl pointed out, makedef.pl didn't also
change to export them in those expanded situations.

### DIFF
--- a/makedef.pl
+++ b/makedef.pl
@@ -550,11 +550,6 @@ unless ($define{USE_LOCALE_COLLATE}) {
 unless ($define{USE_LOCALE_NUMERIC}) {
     ++$skip{$_} foreach qw(
 		    PL_numeric_local
-		    PL_numeric_name
-		    PL_numeric_radix_sv
-		    PL_numeric_standard
-                    PL_numeric_underlying
-                    PL_numeric_underlying_is_standard
                     PL_underlying_numeric_obj
 			 );
 }


jkeenan (commented):
Following @bram-perl's remarks, I created a branch for smoke-testing, [smoke-me/jkeenan/fix-20090-20220818](https://github.com/Perl/perl5/compare/smoke-me/jkeenan/fix-20090-20220818), in which I simply commented out the entire `unless` block.  While your more selective pruning of the entries in `%skip` is probably correct, I'd like to see what happens in those smoke-tests before proceeding.

jkeenan:
> Following @bram-perl's remarks, I created a branch for smoke-testing, [smoke-me/jkeenan/fix-20090-20220818](https://github.com/Perl/perl5/compare/smoke-me/jkeenan/fix-20090-20220818), in which I simply commented out the entire `unless` block. While your more selective pruning of the entries in `%skip` is probably correct, I'd like to see what happens in those smoke-tests before proceeding.

The [smoke-test results](http://perl.develop-help.com/?b=smoke-me%2Fjkeenan%2Ffix-20090-20220818) from my branch are looking good, so I suspect your branch is good to merge as well.  But when you do so, can you include in the commit message an explanation as to why we are retaining `PL_numeric_local` and `PL_underlying_numeric_obj` in the `%skip`, rather than simply removing that entire `unless` block?  Thanks.

bram-perl:
For `PL_numeric_local`: it appears that got renamed in 2017 in commit 892e646556f1997b0384dc508d3223c2c2a7be74;
what was missing in that commit tho was also renaming it in `makedef.pl`;
In commit d36adde059ed1c4f7af210b4f9fc3a7bd2d7d343 ` PL_numeric_underlying` was added in `makedef.pl` but it didn't remove `PL_numeric_local`
TLDR: `PL_numeric_local` should be removed as well (but in a separate commit) 

For `PL_underlying_numeric_obj`: in intrpvar.h
- Before:
```
#ifdef USE_LOCALE_NUMERIC
#  ifdef USE_POSIX_2008_LOCALE
PERLVARI(I, underlying_numeric_obj, locale_t, NULL)
#  endif
#endif
```
- Now:
```
#if defined(USE_LOCALE_NUMERIC) && defined(USE_POSIX_2008_LOCALE)
PERLVARI(I, underlying_numeric_obj, locale_t, NULL)
#endif
```
so that didn't really change; only the condition got merged.
Which means it should be exclude when `USE_LOCAL_NUMERIC` isn't defined *or* when `USE_POSIX_2008_LOCALE` isn't defined.
Makedef.pl (after this patch) contains:
```
...
unless ($define{USE_POSIX_2008_LOCALE})
{
    ++$skip{$_} foreach qw(
        PL_C_locale_obj
        PL_scratch_locale_obj
        PL_underlying_numeric_obj
    );
}
...
unless ($define{USE_LOCALE_NUMERIC}) {
    ++$skip{$_} foreach qw(
		    PL_numeric_local
                    PL_underlying_numeric_obj
			 );
}
```
So that is OK.

The only thing that **might** be a bit clearer could be to add it as (untested!):
```
unless ($define{USE_POSIX_2008_LOCALE} && $define{USE_LOCALE_NUMERIC})
{
    ++$skip{PL_underlying_numeric_obj};
}
```
so that it's (more) obvious that it matches the same condition as in 'intrpvar.h'

bram-perl (approved):

khwilliamson:
I think Bram explained this adequately.  This commit only removes the skips for the things that were failing the test, leaving the rest to (correctly) be skipped.   

I didn't take Bram's suggestion because I think that is less clear.  It turns out that either of those cause the skip to be skipped.  It should read 

` unless ($define{USE_POSIX_2008_LOCALE} ||  $define{USE_LOCALE_NUMERIC})`

But there is already another block for the NUMERIC, and it is combined with something else and I think that makes sense.

It might be a goal to cause this to be automatically generated by some program from the header files.

PR #20121, opened 2022-08-19
Math-Complex: fix typo (remove redundant "an")

pjacklam:
This fixes CPAN RT #85369

PR #20120, opened 2022-08-19
Math-Complex: list all imported constants in POD

pjacklam:
The constant 'pi4' (= 4*pi) is imported, but is missing from the list of
imported constants in the SYNOPSIS.

This fixes CPAN RT #102544

PR #20122, opened 2022-08-19
Math-Complex: fix syntax error in example

pjacklam:
This fixes CPAN RT #131037

### DIFF
--- a/cpan/Math-Complex/lib/Math/Complex.pm
+++ b/cpan/Math-Complex/lib/Math/Complex.pm
@@ -1847,7 +1847,7 @@ number mathematicians call C<j> such that:
 
 is a simple matter of writing:
 
-	$j = ((root(1, 3))[1];
+	$j = (root(1, 3))[1];
 
 The I<k>th root for C<z = [r,t]> is given by:
 


jkeenan:
@pjacklam, thank you for your continued efforts in maintaining Math-Complex.

Since this library is "CPAN upstream," we usually expect the maintainer(s) of such libraries to do revisions upstream and make a new CPAN release.  After that point, someone who has a commit bit to the core distribution runs the program `Porting/sync-with-cpan`, which handles merging the updated CPAN library into core.

Hence, I'm a bit surprised to see this p.r., as well as https://github.com/Perl/perl5/pull/20120 and https://github.com/Perl/perl5/pull/20121, as it suggests you want to do revisions to Math-Complex in the core distribution *first*, rather than after a CPAN release (which, among other things, would generate valuable CPANtesters results).

Can you clarify?

Thank you very much.
Jim Keenan

jkeenan:
> @pjacklam, thank you for your continued efforts in maintaining Math-Complex.
> 
> Since this library is "CPAN upstream," we usually expect the maintainer(s) of such libraries to do revisions upstream and make a new CPAN release. After that point, someone who has a commit bit to the core distribution runs the program `Porting/sync-with-cpan`, which handles merging the updated CPAN library into core.
> 

Hmm, I see that Math-Complex is indeed maintained [upstream on CPAN](https://metacpan.org/pod/Math::Complex), but that its last CPAN release was in 2012, done by someone who is no longer (to the best of my knowledge), and does not have a listed repository.  So how we should proceed in this is unclear.  @rjbs, @book, @leonerd, do you have thoughts about this?

pjacklam:
I have contacted the maintainers of Math-Complex several times though the years asking for co-maint, so I can release new versions of the Math-Complex distribution, but have I never received any reply. I contacted a PAUSE admin to ask for co-maint, but was told to contact the maintainers, which, as I mentioned, I had already done several times without success. So I gave up.

I noticed that the Math-Complex version in core Perl is 1.59_02, newer than the one on CPAN, so I thought I might contribute though pull requests to core Perl. While I know this is not the right way to develop a module that has CPAN as upstream, I don't know how else to do it.

jkeenan:
> I have contacted the maintainers of Math-Complex several times though the years asking for co-maint, so I can release new versions of the Math-Complex distribution, but have I never received any reply. I contacted a PAUSE admin to ask for co-maint, but was told to contact the maintainers, which, as I mentioned, I had already done several times without success. So I gave up.
> 
> I noticed that the Math-Complex version in core Perl is 1.59_02, newer than the one on CPAN, so I thought I might contribute though pull requests to core Perl. While I know this is not the right way to develop a module that has CPAN as upstream, I don't know how else to do it.

Thanks for taking the initiative on this.  I suspect the Perl Steering Council and PAUSE administrators will have to discuss this.  (That, however, should not keep you from continuing to work on these p.r.s.)

jkeenan:
Math-Complex is now maintained in `dist/` rather than `cpan/`.  Handled manually in 17f4c42e6e.  Closing.

PR #20123, opened 2022-08-20
CI: Special case porting test in 'sanity check'

### LINKED ISSUE #19867: Need to skip AUTHOR testing for github CI sanity

bram-perl:
Add a special case for the porting test in the sanity check of the
CI run.

**Before:**

When a porting test failed (such as t/porting/authors.t, t/porting/cmp_version.t, ...) then it would skip all the other test configurations.

**Now:**

When a porting test fails:
- result of sanity check is failure (i.e. it's red)
-  other test configurations are run and these will also run the porting tests. (These may or may not fail, depending on the test)

When a non-porting test fails:
- result of sanity check is failure (i.e. it's red)
- sanity check still runs the porting tests
- other test configurations are skipped

Note: in the `if`conditions of the other test configurations it was needed to include `always()` because otherwise GitHub uses an implicit `success() &&` (which check the result of the sanity check - which is unwanted if only the porting tests failed)

Fixes #19867

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -102,15 +102,16 @@ jobs:
         env:
           CONFIGURE_ARGS: "-Dusethreads"
       - name: Build
+        id: build
         run: |
           make -j2 test_prep
       - name: Show Config
         run: |
           ./perl -Ilib -V
           ./perl -Ilib -e 'use Config; print Config::config_sh'
-      - name: Run Tests
+      - name: Run Tests (excluding t/porting)
         run: |
-          TEST_JOBS=2 ./perl t/harness
+          TEST_JOBS=2 ./perl t/harness -nre='^porting/'
 
       # Set a variable for dependent jobs to signal if full testsuite is enabled
       - name: "Check if EXTENDED_TESTING is set"
@@ -127,6 +128,11 @@ jobs:
             echo "::set-output name=run_all_jobs::true"
           fi
 
+      - name: Run Porting Tests (t/porting)
+        if: always() && steps.build.outcome == 'success'
+        run: |
+          TEST_JOBS=2 ./perl t/harness -re='^porting/'
+
   #  _ _
   # | (_)_ _ _  ___ __
   # | | | ' \ || \ \ /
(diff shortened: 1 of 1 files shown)


bram-perl:
Note: the CI run fails due to the authors tests but that is unrelated to this PR.
The solution (and explanation) for that is in PR #20124

demerphq:
Looks good to me. But obviously it would be good to see the CI on top of #20124

demerphq:
I'm really inclined to say we should just teach updateAUTHORS to ignore new entries like "users.noreply.github.com" unless someone runs it with a special option. This this problem would go away and we wouldnt need the machinations elsewhere to deal with it.

```
#   Failed test 'Is mailmap_file '.mailmap' up to date?'
#   at Porting/updateAUTHORS.pl line 126.
#   File '.mailmap' changes:
#     would add: Bram <perl-rt@wizbit.be> Bram <109858694+bram-perl@users.noreply.github.com>
```

bram-perl:
> I'm really inclined to say we should just teach updateAUTHORS to ignore new entries like "users.noreply.github.com" unless someone runs it with a special option.

Go for it, but keep these things in mind:
- AUTHORs that only have commits with a 'users.noreply.github.com' should still be credited (but their emailaddress can/should be empty)
- It is not enough to remove the `$revision_range` logic in t/porting/authors.t (it'll be needed in less cases but it can/will still be needed in some)

demerphq:
On Mon, 22 Aug 2022 at 19:17, Bram ***@***.***> wrote:

> I'm really inclined to say we should just teach updateAUTHORS to ignore
> new entries like "users.noreply.github.com" unless someone runs it with a
> special option.
>
> Go for it, but keep these things in mind:
>
>    - AUTHORs that only have commits with a 'users.noreply.github.com'
>    should still be credited (but their emailaddress can/should be empty)
>
> I was only thinking of  .mailmap changes.


>
>    - It is not enough to remove the $revision_range logic in
>    t/porting/authors.t (it'll be needed in less cases but it can/will still be
>    needed in some)
>
> Understood. I'd like to learn more about what would be required. :-)

cheers,
Yves
-- 

demerphq (approved):

PR #20125, opened 2022-08-20
Perl 5.37.3

neilb:
I've currently lost my commit bit, so submitting the 5.37.3 as a PR.

Neil


Leont:
"[perldelta WIP](https://github.com/Perl/perl5/pull/20125/commits/fc91b6f948ee9aa41beb14368723601c1f835d57)" sounds like it should be renamed.

khwilliamson:
oops didn't see that comment

neilb:
> "perldelta WIP" sounds like it should be renamed.

Gah. I had renamed it as part of a rebase, but that then got aborted, and I forgot to rename it in the successful rebase.
Can I change the name of a commit after the fact? If so, how do I do that?

PR #20128, opened 2022-08-21
The idea of this PR is to give the dev a little bit more control over our CI testing

demerphq:
Karl asked for a way to run CI with a restricted set of targets, which is pretty easy to do on the command line. This PR is intended to allow some magic files to be included in the root directory, RUNTESTS_TEST_ARGS and RUNTESTS_HARNESS_ARGS, which is intended to be a way to make runtests use the desired harness target with the desired arguments.

Files of the form RUNTESTS_[A-Z]_ARGS are forbidden from being listed in the manifest, and if the manifest checks run they will get unhappy if the files are either present on disk or listed explicitly. Eg, unless somone is extremely careless we should never merge these files to blead.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -203,6 +203,11 @@ jobs:
       matrix:
         CONFIGURE_ARGS:
           - "-Dusedevel"
+        TEST_TARGET:
+          - "test"
+          - "test_harness"
+          - "test_porting"
+          - "test_reonly"
 
     steps:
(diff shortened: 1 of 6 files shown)


tonycoz:
Can't the dev just add:
```
  TEST_FILES: "sometestfilepaths"
```

to the env block at the top of `testsuite.yml`?

demerphq:
@tonycoz yeah, i guess they could. i havent tested it tho, but ill  close this anyway.

Issue #20081, opened 2022-08-11
Windows builds built with gcc -O2 works on Win10 but fails on Win11


kenneth-olwing:
_**The below was done with blead@c52f54c8521c6155984dba0a0675d3220170468b and using a standard MSI install of Strawberry  Perl 5.32.1.1 as the toolchain.**_

**Description**
Building is done out-of-the box with:
```
cd win32
gmake -f GNUmakefile CCHOME=c:\Strawberry\c test
```
The build seems fine. Also, on a Win10 machine tests passes.
However, on my Win11 box the tests fails with:
```
Test Summary Report
-------------------
re/reg_mesg.t                                                      (Wstat: 65280 (exited 255) Tests: 339 Failed: 0)
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
re/regexp_unicode_prop.t                                           (Wstat: 65280 (exited 255) Tests: 0 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 1110 tests but ran 0.
re/regexp_unicode_prop_thr.t                                       (Wstat: 65280 (exited 255) Tests: 4 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 1110 tests but ran 4.
re/subst.t                                                         (Wstat: 65280 (exited 255) Tests: 236 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 278 tests but ran 236.
re/subst_wamp.t                                                    (Wstat: 65280 (exited 255) Tests: 236 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 278 tests but ran 236.
re/substT.t                                                        (Wstat: 65280 (exited 255) Tests: 236 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 278 tests but ran 236.
win32/popen.t                                                      (Wstat: 2304 (exited 9) Tests: 0 Failed: 0)
  Non-zero exit status: 9
  Parse errors: No plan found in TAP output
../dist/Safe/t/safeops.t                                           (Wstat: 65280 (exited 255) Tests: 71 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 418 tests but ran 71.
../dist/Storable/t/utf8.t                                          (Wstat: 65280 (exited 255) Tests: 5 Failed: 0)
  Non-zero exit status: 255
[...]

PR #20131, opened 2022-08-21
Make it possible to turn off (or lengthen time for) the watchdog in the t\win32\popen.t test

kenneth-olwing:
In my case when building on Win11, the t\win32\popen.t test often fails due to a watchdog timing out. Turning off (or really lengthening) the timer, it completes successfully. So far, this seems prevalent on Win11, have not been able to replicate on Win10.
As the root cause for why it takes a longer time to succeed is unknown, I suspect that the watchdog code itself is causing a problem. Hence, I require the possibility to turn it off which this change enables.

For further discussion on when this was identified, see #20081 (this problem seems not to be directly coupled with the issue discussed there).

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -495,6 +495,7 @@ Keith Thompson <keith.s.thompson@gmail.com> Keith Thompson <kst@mib.org>
 Ken Neighbors <unknown> Ken Neighbors <perlbug@perl.org>
 Ken Williams <ken@mathforum.org> <kenahoo@gmail.com>
 Ken Williams <ken@mathforum.org> Ken Williams <ken.williams@thomsonreuters.com>
+Kenneth Ölwing <knth@cpan.org>
 Kent Fredric <kentfredric@gmail.com> Kent Fredric <kentnl@cpan.org>
 Kevin Brintnall <kbrint@rufus.net> kevin brintnall <kbrint@rufus.net>
 Kevin Ryde <user42@zip.com.au> Kevin Ryde <perlbug-followup@perl.org>
--- a/AUTHORS
+++ b/AUTHORS
@@ -783,7 +783,6 @@ Ken Williams                   <ken@mathforum.org>
 Kenichi Ishigaki               <ishigaki@cpan.org>
 Kenneth Albanowski             <kjahds@kjahds.com>
 Kenneth Duda                   <kjd@cisco.com>
-Kenneth Olwing                 <knth@cpan.org>
 Kenneth Ölwing                 <knth@cpan.org>
 Kent Fredric                   <kentfredric@gmail.com>
 Keong Lim                      <Keong.Lim@sr.com.au>
--- a/AUTHORS
+++ b/AUTHORS
@@ -783,6 +783,8 @@ Ken Williams                   <ken@mathforum.org>
 Kenichi Ishigaki               <ishigaki@cpan.org>
 Kenneth Albanowski             <kjahds@kjahds.com>
 Kenneth Duda                   <kjd@cisco.com>
+Kenneth Olwing                 <knth@cpan.org>
+Kenneth Ölwing                 <knth@cpan.org>
 Kent Fredric                   <kentfredric@gmail.com>
 Keong Lim                      <Keong.Lim@sr.com.au>
 Kevin Brintnall                <kbrint@rufus.net>
--- a/t/win32/popen.t
+++ b/t/win32/popen.t
@@ -13,7 +13,15 @@ BEGIN {
 
 # [perl #77672] backticks capture text printed to stdout when working
 # with multiple threads on windows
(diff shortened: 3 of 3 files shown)


khwilliamson:
Consider instead using the existing mechanism 

`my $time_out_factor = $ENV{PERL_TEST_TIME_OUT_FACTOR} || 1;

  $time_out_factor = 1 if $time_out_factor < 1;

  watchdog(20 * $time_out_factor);
`
This was originally created so that someone working on a slow platform could set this environment variable to say this runs twic as slow as normal, or 3 times, etc.  But it is applicable to your case and doesn't introduce a new, specialized variable

kenneth-olwing:
Thanks, I didn't know about that.

I see two differences though...one is that just setting the time factor doesn't allow me to *completely* disable the watchdog. The other is ofc that setting the time factor will change the time factor in every test that uses it. I don't think my machine is slow by any measure, this is the only test affected and it shouldn't really take 20+ secs for the code to run - hence, I believe there is something else at play. Which, of course I don't have a real solution for :-/

But I'll have a go at that approach see if can get it to run reliably. BRB.

khwilliamson:
On 8/21/22 12:19, Kenneth Ölwing wrote:
> Thanks, I didn't know about that.
> 
> I see two differences though...one is that just setting the time factor 
> doesn't allow me to /completely/ disable the watchdog. The other is ofc 
> that setting the time factor will change the time factor in every test 
> that uses it. I don't think my machine is slow by any measure, this is 
> the only test affected and it shouldn't really take 20+ secs for the 
> code to run - hence, I believe there is something else at play. Which, 
> of course I don't have a real solution for :-/

I was under the impression that you need to bump up the time temporarily 
to figure things out, and then you could drop it down again.  What I was 
suggesting was a way that you could do that, and leave the change in, in 
case it is needed in the future, and it would make the code work on a 
slow machine.  As for the other affected tests, they can have their 
timeout increased for now without any effect, as they aren't using their 
max time now.

So this mechanism was to be used for slow machines, but it seems to me 
like it could be used for your current purpose, and the change left in. 
  Just bump the factor to such a large value that the watchdog doesn't 
time out.  I've used this method at points in the past to debug similar 
situations.

kenneth-olwing:
Hi, thanks for your insight,

Well, on the whole, I have to admit that I don't have the foggiest on how to fix any problem causing the test to run unnecessarily long or even hang. Basically, it shouldn't and if it does then there is clearly a more insidious problem that the watchdog simply hides. And, since I can't get it to ill-behave by hanging, now the watchdog kills it before it finishes.

But yes, prolonging the timeout does appear to help, so that's good. But in the end I think it would be a good idea to simply be able to turn off the watchdog functionality entirely and try to figure things out without that distraction.

I've pushed a new change to do it the lengthening way for now. I've withdrawn my changes to AUTHORS/.mailmap...Tried using Porting/updateAuthors.pl but pushing those changes gives me a conflict with .mailmap that I just can't figure out...any hints on what I should do to make all test actions happy?

ken1

demerphq:
On Mon, 22 Aug 2022 at 07:44, Kenneth Ölwing ***@***.***>
wrote:

> Hi, thanks for your insight,
>
> Well, on the whole, I have to admit that I don't have the foggiest on how
> to fix any problem causing the test to run unnecessarily long or even hang.
> Basically, it shouldn't and if it does then there is clearly a more
> insidious problem that the watchdog simply hides. And, since I can't get it
> to ill-behave by hanging, now the watchdog kills it before it finishes.
>
> But yes, prolonging the timeout does appear to help, so that's good. But
> in the end I think it would be a good idea to simply be able to turn off
> the watchdog functionality entirely and try to figure things out without
> that distraction.
>
> I've pushed a new change to do it the lengthening way for now. I've
> withdrawn my changes to AUTHORS/.mailmap...Tried using
> Porting/updateAuthors.pl but pushing those changes gives me a conflict with
> .mailmap that I just can't figure out...any hints on what I should do to
> make all test actions happy?
>
Your base image isn't against the latest blead but against an older patch
in blead, and since then there were some very large cleanups to the
.mailmap file so there is a sort of "commit horizon" where you will have
conflicts with .mailmap. You just need to make sure your base patch is
applied to a commit *after* the cleanup. (This problem is not specific to
the .mailmap file, any large list might have the same problem if the same
type of changes were applied to it.)

Assuming "origin" is the github master repo, and assuming you have NOT
included any .mailmap changes you should do the following:

git co $your_branch
git fetch
git rebase origin/blead # without any .mailmap changes
Porting/updateAUTHORS.pl # should apply cleanly.

And it should all pass green.

Yves


-- 

kenneth-olwing:
Thanks, now on track.

ken1

kenneth-olwing:
Closing this as #20134 will assume the desired effect.

Issue #20132, opened 2022-08-21
BBC: Blead Breaks Hash::MoreUtils


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.42 running under perl 5.37.3.

-----------------------------------------------------------------
[Please describe your issue here]

BBC: Blead Breaks Hash::MoreUtils

Please see http://matrix.cpantesters.org/?dist=Hash::MoreUtils

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.37.3:

Configured by cpan at Thu Aug 18 11:49:39 EDT 2022.

Summary of my perl5 (revision 5 version 37 subversion 3) configuration:
  Commit id: 52ed379edc11b00708b403ff03aaf5e3a1a2eb81
  Platform:
    osname=linux
    osvers=5.15.60-0-lts
    archname=x86_64-linux
    uname='linux cjg-alpine3 5.15.60-0-lts #1-alpine smp tue, 16 aug 2022 09:10:36 +0000 x86_64 linux '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='10.3.1 20211027'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
[...]

Issue #20114, opened 2022-08-18
BBC: Blead Breaks Variable::Magic, Devel::Caller


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.42 running under perl 5.37.3.

-----------------------------------------------------------------
[Please describe your issue here]

BBC: Blead Breaks Variable::Magic

Please see http://matrix.cpantesters.org/?dist=Variable::Magic

[Please do not change anything below this line]
-----------------------------------------------------------------
---
Flags:
    category=core
    severity=low
---
Site configuration information for perl 5.37.3:

Configured by cpan at Thu Aug 18 01:16:20 EDT 2022.

Summary of my perl5 (revision 5 version 37 subversion 3) configuration:
  Commit id: bbae2b23b3ff14ae001b0b2b0ba757540aaae4b8
  Platform:
    osname=openbsd
    osvers=7.1
    archname=OpenBSD.amd64-openbsd
    uname='openbsd cjg-openbsd7 7.1 generic#3 amd64 '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    optimize='-O2'
    cppflags='-fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='OpenBSD Clang 13.0.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags ='-Wl,-E  -fstack-protector-strong -L/usr/local/lib'
[...]

PR #20133, opened 2022-08-22
rpeep: don't apply padsv_store and padrange together

richardleach:
As originally committed, the OP_PADSV_STORE optimization interacted
negatively with OP_PADRANGE. The simple fix applied by this commit is
to modify peep(), such that the OP_PADSV_STORE optimization is not
applied when OP_PADRANGE has already been applied.

Existing tests did not pick up the problems, which were identified via
Blead-Breaks-CPAN reports. Additional tests have thus been included.

### DIFF
--- a/t/op/lex_assign.t
+++ b/t/op/lex_assign.t
@@ -227,6 +227,35 @@ is($@, '', 'ex-PVBM assert'.$@);
     cmp_ok($diff, '<',  2,  "time delta is small");
 }
 
+# GH #20132 and parts of GH ##20114
+# During development of OP_PADSV_STORE, interactions with OP_PADRANGE
+# caused BBC failures not picked up by any pre-existing core tests.
+# (Problems only arose in list context, the void/scalar tests have been
+# included for completeness.)
+eval {
+    my $x = {}; my $y;
+    keys %{$y = $x};
+    1;
+};
+is($@, '', 'keys %{$y = $x}');
+
+eval {
+    my $x = {}; my $y;
+    my $foo = keys %{$y = $x};
+    1;
+};
+is($@, '', 'my $foo = keys %{$y = $x}');
+
+eval {
+    my $x = {}; my $y;
+    my @foo = keys %{$y = $x};
+    1;
+};
+is($@, '', 'my @foo = keys %{$y = $x}');
+
+fresh_perl_is('my ($x, $y); (($y = $x))', '', {}, '(($y = $x))');
+fresh_perl_is('my ($x, $y); my $z= (($y = $x))', '', {}, 'my $z= (($y = $x))');
+fresh_perl_is('my ($x, $y); my @z= (($y = $x))', '', {}, 'my @z= (($y = $x))');
 
 done_testing();
 
--- a/peep.c
+++ b/peep.c
@@ -3854,6 +3854,8 @@ Perl_rpeep(pTHX_ OP *o)
             if (!(o->op_private & (OPpASSIGN_BACKWARDS|OPpASSIGN_CV_TO_GV))
                  && lval && (lval->op_type == OP_PADSV) &&
                 !(lval->op_private & OPpDEREF)
+                 /* skip if padrange has already gazumped the padsv */
+                 && (lval == oldop)
                ) {
 
                 /* SASSIGN's bitfield flags, such as op_moresib and
@@ -3876,6 +3878,7 @@ Perl_rpeep(pTHX_ OP *o)
                 o->op_targ = lval->op_targ; lval->op_targ = 0;
 
                 /* Fixup op_next ptrs */
+                assert(oldop->op_type == OP_PADSV);
                 /* oldoldop can be arbitrarily deep in the RHS OP tree */
                 oldoldop->op_next = o;
 


richardleach:
Fixes #20132 (Hash::MoreUtils)
Fixes recent additions to #20114 (Mixin::Event::Dispatch, Test::Weaken)

jkeenan:
> Fixes #20132 (Hash::MoreUtils) Fixes recent additions to #20114 (Mixin::Event::Dispatch, Test::Weaken)

Confirmed ... but I note that it does not fix the current (unpatched) CPAN versions of Variable::Magic and Devel::Caller (https://github.com/Perl/perl5/issues/20114).

t/op/lex_assign.t
```diff
@@ -227,6 +227,17 @@ is($@, '', 'ex-PVBM assert'.$@);
     cmp_ok($diff, '<',  2,  "time delta is small");
 }
 
+# GH #20132 and parts of GH ##20114
+# During development of OP_PADSV_STORE, interactions with OP_PADRANGE
+# caused BBC failures not picked up by any pre-existing core tests.
+eval {
+    my $x = {}; my $y;
+    print keys %{$y = $x};
```
bram-perl: I'm not a fan of using `print` in the test..

Sure, it *should* print nothing but if it does print something then it may lead to inconsistent results between `make test` and `make test_harness`
(It **appears** that `make test` will fail if it encounters unexpected output on stdout where as `make test_harness` doesn't **appear** to care very much)

richardleach: Just doing `keys %{$y = $x};` didn't seem to trigger the bug, nor did assigning the result. I'll double check and then have a think, but am very open to suggestions for alternatives.

richardleach:
> Confirmed ... but I note that it does not fix the current (unpatched) CPAN versions of Variable::Magic and Devel::Caller (#20114).

Thanks for confirming. Yeah, those two are definitely needing patching themselves to be aware of the new OP.

t/op/lex_assign.t
bram-perl: `keys %{$y = $x}` --> keys is in void context
`$foo = keys %{$y = $x}` --> keys in scalar context
`print keys %{$y = $x}` --> keys in list context

What does seem to crash is when using it in list context,
i.e.:
`@foo = keys %($y = $x};`

richardleach: Do you get an actual crash (e.g. segfault/panic) with `@foo = keys %($y = $x};` or just the warning _Can't use an undefined value as a HASH reference at -e line 1._? I only see the latter on blead, but that's also how v5.32 behaves.

richardleach: No, I was just being stoopid. It does throw the error on blead but not v5.32.  I'll swap that in as the test.

richardleach: Also made the test changes/additions as recommended on IRC. See what you think?

bram-perl: The tests looks OK.

I can't comment on the actual code change tho, I'm not familiar with that area.

Only thing I'm now wondering about is if the comment message can be improved..
Looking at the commit message:
> As originally committed, the OP_PADSV_STORE optimization interacted negatively with OP_PADRANGE.

That makes me wonder how they interacted and how/why that interaction became negative..
I don't know if there is a possibility (or a need for that matter) to elaborate a bit more about it in the commit message?
(Skimming the conversation on p5p: there was talk about the stack but I'm not sure what conclusion it reached)

(The real question: if you read that commit message again in a year or so is it then clear enough to know what was wrong?)

richardleach: I've updated the commit message with some more gory details. Thanks for the reviews.

tonycoz (approved):

demerphq:
> Just doing keys %{$y = $x}; didn't seem to trigger the bug, nor did assigning the result. I'll double check and then have a think, but am very open to suggestions for alternatives.

Triggers for me. I think you keep forgetting that context matters here:

```
./perl -le'my $x={}; my $y; my @ar= keys %{$y = $x}; print $y'
Can't use an undefined value as a HASH reference at -e line 1.

./perl -le'sub f{} my $x={}; my $y; f(keys %{$y = $x}); print $y'
Can't use an undefined value as a HASH reference at -e line 1.
```

You need to use an array assignment, or use a subroutine to impose list context to trigger the buggy behavior.

Also keys isn't required. Same bug, and an assert failure:

```
$ ./perl -le'sub f{} my $x={}; my $y; f(%{$y = $x}); print $y'
perl: inline.h:360: Perl_POPMARK: Assertion `(PL_markstack_ptr > PL_markstack) || !"MARK underflow"' failed.
Aborted (core dumped)
```

All you have to do is dereference in list context.

Yves

richardleach:
> > Just doing keys %{$y = $x}; didn't seem to trigger the bug, nor did assigning the result. I'll double check and then have a think, but am very open to suggestions for alternatives.
> 
> Triggers for me. I think you keep forgetting that context matters here:

Yes, I had forgotten context at that point. Bram has already put me straight.

Both the _Can't use an undefined value_ and assertion failure paths are hit by tests included in this PR.

Bram suggested that I make the tests `use strict` safe, which I'll do and then merge.

PR #20134, opened 2022-08-22
Consolidate PERL_TEST_TIME_OUT_FACTOR to watchdog()

khwilliamson:
This changes test.pl watchdog() to always consider this potential
setting of an environment variable, and removes the distributed uses.

This means that no code needs to change when tests start failing on a
slow platform due to timing out; or when you need to temporarily
increase the timeout of a test for debugging.  This will correspondingly
increase the timeout of all tests, but who cares for debugging purposes.

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -1724,7 +1724,11 @@ sub warning_like {
     }
 }
 
-# Set a watchdog to timeout the entire test file
+# Set a watchdog to timeout the entire test file.  The input seconds is
+# multiplied by $ENV{PERL_TEST_TIME_OUT_FACTOR} (default 1; minimum 1).
+# Set this in your profile for slow boxes, or use it to override the timeout
+# temporarily for debugging.
+#
 # NOTE:  If the test file uses 'threads', then call the watchdog() function
 #        _AFTER_ the 'threads' module is loaded.
 sub watchdog ($;$)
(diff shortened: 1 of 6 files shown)


t/test.pl
```diff
@@ -1733,8 +1737,16 @@ ($;$)
     my $method  = shift || "";
     my $timeout_msg = 'Test process timed out - terminating';
 
+    # Accept either spelling
+    my $timeout_factor = $ENV{PERL_TEST_TIME_OUT_FACTOR}
```
demerphq, line 1741: FWIW, I was thinking of taking this a teeny step further.
```
my $script= $0;
$script=~s!/!_!g;
my $timeout_factor = $ENV{"PERL_TTF_$script"} || $ENV{PERL_TEST_TIME_OUT_FACTOR}
```

That is where we can set an environment variable to make a specific test script do this. So we have two level of settings "give all my tests extra time" and "give this script extra time".

demerphq (approved):
Approving because its "good enough", but i think it could be a teeny bit better still... Up to you if you want to do it in this thread tho.

t/test.pl
kenneth-olwing: For the record, in case you feel inclined to change it further, I still think it would be nice to be able to turn the watchdog off altogether. E.g.  make an immediate return if the 'timeout' resolves to '0' or negative or so.

khwilliamson:
I think Yves idea is good, but I think that should go in a separate pull request.  I don't think it's a wise idea to turn off the watchdog globally, so that would have to come as part of an individual script PR

demerphq:
FWIW, I don't think we should decide what is wise or not in this context. That substitutes our imagination for someone else's and they might dream bigger than we do. If someone wants to run the test suite with no watchdogs why should we care so long as they get to keep the parts when it breaks? Anyway, if you think the per-script logic should be in a separate PR lll whack one together when i get a chance if you dont beat me to it.

PR #20136, opened 2022-08-23
Change optimization level for Win32 builds

kenneth-olwing:
As described in #20081, building with the toolchain in Strawberry 5.32.1 (gcc 8.3.0), something in -O2 optimization stably causes fatal errors (i.e. process dying hard) when building (on either Windows 10/11) and running the resulting perl on Windows 11. For safety until the cause can be found,, this PR shifts to -Os instead which, again stably, makes the problems disappear.

As an aside, I think it's odd to have a debug build do -O2 optimizations; instead suggesting -Og which is documented to provide a good debugging experience.

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -608,11 +608,16 @@ LIBFILES	+= -lquadmath
 endif
 
 ifeq ($(CFG),Debug)
-OPTIMIZE	= -g -O2
+# According to https://gcc.gnu.org/onlinedocs/gcc-8.3.0/gcc/Optimize-Options.html
+# -Og should provide some optimizations while still giving convenient debugging
+OPTIMIZE	= -g -Og
 LINK_DBG	= -g
 DEFINES		+= -DDEBUGGING
 else
-OPTIMIZE	= -O2
+# In https://github.com/Perl/perl5/issues/20081 it is found that the previous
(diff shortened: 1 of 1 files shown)


kenneth-olwing:
@jkeenan Unsure, but apparently you tagged stuff before, and I assume this should be a 'distro-win32' or so?

@xenu Not sure what the protocol is for asking for a review otherwise, and I'm sure you have a lot to do, but if possible a quick review would be helpful so I know it's seen...

Thanks,
ken1

demerphq (approved):

demerphq:
Its been a while since I did any win32 hacking, but this looks reasonable to me. I also think its weird to do -O2 in a debugging build, although I do remember in the distant past that most win32 dev was done on VC/Visual Studio, and its debugging capabilities were different than on linux, and I wonder if this is from then. I vaguely remember something about being able to get VC to write a debugging info file that would allow you to debug at higher optimizations levels much like you would normally. I could easily imagine this being from then, and -Og being a better default.

sisyphus:
I just want to add my support to the changing of default optimization level from -O2 to -Os.
My reasoning is simple - I experience cases where -O2 optimization causes problems, but -Os optimization is, IME. always trouble-free.
It makes perfect sense to me that the trouble-free one should be the default.

Cheers,
Rob

PR #20137, opened 2022-08-23
Prefer %ll over %L for formatting long long ints

t-a-k:
Configure used to try `%L`, `%q`, and then `%ll` in this order to find the modifier for formatting 64-bit integers on 32-bit machines, but `%L` for integer conversions (such as in `%Ld`) seems to be a non-standard GNU extension.
I think it will better to try `%ll` before `%q` and `%L`, as `%ll` is standardized by C99.

### DIFF
--- a/Configure
+++ b/Configure
@@ -21657,28 +21657,6 @@ EOCP
 	fi
 fi
 
-if $test X"$sPRId64" = X -a X"$quadtype" != X; then
-	$cat >try.c <<EOCP
-#include <sys/types.h>
-#include <stdio.h>
-int main() {
-  $quadtype q = 12345678901;
-  printf("%Ld\n", q);
-}
-EOCP
-	set try
-	if eval $compile; then
-		yyy=`$run ./try`
-		case "$yyy" in
-		12345678901)
-			sPRId64='"Ld"'; sPRIi64='"Li"'; sPRIu64='"Lu"';
-			sPRIo64='"Lo"'; sPRIx64='"Lx"'; sPRIXU64='"LX"';
-			echo "We will use %Ld."
-			;;
-		esac
-	fi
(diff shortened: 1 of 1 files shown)


demerphq:
Doesnt this need @Tux ?

Tux:
Yes, it does. I am trying real hard to catch up currently

Tux:
I'd like to see test-runs on AIX, HP-UX and Solaris.
Sadly my resources are currently off-line

Tux:
Otherwise I think this change looks good! Thank you

demerphq:
@tux, does that mean we can merge this PR?

Tux:
HP-UX and AIX are back on-line. Anyone has time to check them?

khwilliamson:
I tested this on aix with -DUSE_MORE_BITS, and unfortunately all the formats for 32-bit items look like this:
` #define I32df                           /**/`

Tux:
I have no idea what the final verdict is. Can I merge this now?

khwilliamson:
It doesn't work on aix. @t-a-k please look at this

t-a-k:
Sorry for delayed response, but I don't have an access to AIX hosts so a bit more information will be helpful.

Does the test pass if configured without `-DUSE_MORE_BITS`?

khwilliamson:
I tried again on AIX, and it worked!  I don't know what went wrong before.  Sorry.

I think this should be merged in 5.39.1

t-a-k:
> I tried again on AIX, and it worked!

I'm relieved to hear that.

Anyway this PR should only affect formats for 64-bit integers.  If any 32-bit formats such as `I32df` went wrong, I'm worried that there might be another issue, possibly related to #20415.

PR #20143, opened 2022-08-23
Locale changes for "$!"

khwilliamson:
This series of commits improves the handling of system error messages that come from strerror().

It turns out that strerror() can return mojibake if the LC_CTYPE locale and the LC_MESSAGES locale aren't set to the same thing.  They are now synced for the duration of the routine, if necessary.

There have long been lots of preprocessor conditionals depending on  the platform's capabilities.  These are now refactored so that the various implementations are separate short functions, which are easier to understand when not interrupted by those directives.

These commits and previous ones have now completely replaced some large and cumbersome functions, which are also removed in this series.


locale.c
```diff
@@ -6680,19 +6724,15 @@ Perl_my_strerror(pTHX_ const int errnum)
...
-              "Strerror returned; saving a copy: '"),
-              print_bytes_for_locale(errstr, errstr + strlen(errstr), 0),
-              PerlIO_printf(Perl_debug_log, "'\n")));
-
-#endif   /* End of does have locale messages */
+    DEBUG_STRERROR_RETURN(errstr);
 
     SAVEFREEPV(errstr);
     return errstr;
 }
```
tonycoz: > So this commit separates them out into 5 different
> functions, with the result that it's clear what is going on in each.

This commit splits out the ends of the function definitions but the start of the 5 different function definitions aren't split out until the next commit.

mg.c
```diff
@@ -945,6 +931,14 @@ Perl_magic_get(pTHX_ SV *sv, MAGIC *mg)
...
                 PerlProc_GetOSError(sv, dwErr);
                 fixup_errno_string(sv);
+
+#     ifdef USE_LOCALE
+                if (   IN_LOCALE
+                    && get_win32_message_utf8ness(SvPV_nomg_const_nolen(sv)))
+                {
+                    SvUTF8_on(sv);
+                }
+#     endif
```
tonycoz, line 941: Isn't fixup_errno_string() already doing something similar to this?

That said the selected locale doesn't seem to change the messages anyway:
```
#!perl
use Devel::Peek;
use POSIX qw(LC_ALL setlocale);
use locale;
my $loc = setlocale(LC_ALL, "Japanese.utf8");
print $loc, "\n";
unlink "Doesn't exist";
my $x = $^E;
Dump($x);
my $y = $!;
Dump($y);
```

result (I have the Japanese language pack installed):
```
C:\Users\tony\dev\perl\git\perl\win32>..\perl ..\..\errortest.pl
Japanese_Japan.utf8
SV = PVMG(0x17d4ba24d18) at 0x17d4ba0c860
  REFCNT = 1
  FLAGS = (NOK,POK,IsCOW,pNOK,pPOK,UTF8)
  IV = 0
  NV = 2
  PV = 0x17d4d53fb78 "The system cannot find the file specified"\0 [UTF8 "The system cannot find the file specified"]
  CUR = 41
  LEN = 43
  COW_REFCNT = 1
SV = PVMG(0x17d4ba24d48) at 0x17d4d4a2580
  REFCNT = 1
  FLAGS = (NOK,POK,IsCOW,pNOK,pPOK)
  IV = 0
  NV = 2
  PV = 0x17d4d4ea798 "No such file or directory"\0
  CUR = 25
  LEN = 27
  COW_REFCNT = 1
```

locale.c
khwilliamson: revised

mg.c
khwilliamson: IIUC, you're suggesting moving the check into the fixup function. I think it's slightly better the current way.  Moving it there would mean that would have an #ifdef for Windows, instead of here, and that would be the only case where that function would care about UTF8ness.  But that function is called from another place where UTF8ness has already been determined before the call.  And that would mean an inconsistent use of that function.  So I think it's better to keep that function not worrying about UTF-8 in any circumstance, and leave the needed special case here.

It has been my experience that most language packs do not come with the error strings in anything but English.  I just checked on the only box currently available to me located in a country where the native language isn't English, and the message was in English even so.  These commits were developed a couple of years ago, and I did find boxes where some locales did have translations for these.  Perhaps they were @Tux 's boxes which are currently offline

khwilliamson:
Closed via 69bc7167fa24b1e8d3f810ce465d84bdddf413f6

mg.c
tonycoz: The messages from $^E seem to depend on the system language, when I changed the session language to Japanese in the control panel (which required a log out / in), I instead got:
```
C:\Users\tony\dev\perl\git\perl\win32>..\perl -w ..\..\errortest.pl
Japanese_Japan.utf8
SV = PVMG(0x20e5d2c6f78) at 0x20e5d2acff0
  REFCNT = 1
  FLAGS = (NOK,POK,IsCOW,pNOK,pPOK,UTF8)
  IV = 0
  NV = 2
  PV = 0x20e5ed05038 "??????????????????"\0 [UTF8 "??????????????????"]
  CUR = 18
  LEN = 20
  COW_REFCNT = 1
SV = PVMG(0x20e5d2c6fa8) at 0x20e5ecae468
  REFCNT = 1
  FLAGS = (NOK,POK,IsCOW,pNOK,pPOK)
  IV = 0
  NV = 2
  PV = 0x20e5ed06db8 "No such file or directory"\0
  CUR = 25
  LEN = 27
  COW_REFCNT = 1
```

This wasn't locale dependent. commenting out the setlocale() made no difference.

As to the fixup function, it was more that the extra code added here and the fixup function have a very strong overlap, the new code comes close to setting UTF-8 in all the circumstances the fixup function does., so perhaps the call to the fixup could be eliminated.

An alternative would be to make PerlProc_GetOSError() (aka win32_str_os_error()) always return the UTF-8 version of the message, and flag its return value that way (since it works with an SV rather than a char *).

khwilliamson: @leont, do you have an opinion on the 2nd paragraph of Tony's comment

PR #20147, opened 2022-08-24
Add PUSHpvs("literal") macro family

leonerd:
This set of PUSH-style macros takes a string literal argument and pushes it to the stack, optionally mortalizing it and/or extending the stack.

Previously, the best alternative was

    mPUSHp("literal", 7);

which required the author to visually count the number of characters in the string literal (7 in this case). These new macros fit the similar pattern familiar to many functions such as `newSVpvs`, which takes a string literal and counts it directly.

### DIFF
--- a/pp.h
+++ b/pp.h
@@ -201,6 +201,10 @@ C<TARG>, so C<dTARGET> or C<dXSTARG> should be called to declare it.  Do not
 call multiple C<TARG>-oriented macros to return lists from XSUB's - see
 C<L</mPUSHp>> instead.  See also C<L</XPUSHp>> and C<L</mXPUSHp>>.
 
+=for apidoc Am|void|PUSHpvs|"literal string"
+A variation on C<PUSHp> that takes a literal string and calculates its size
+directly.
+
 =for apidoc Am|void|PUSHn|NV nv
 Push a double onto the stack.  The stack must have room for this element.
 Handles 'set' magic.  Uses C<TARG>, so C<dTARGET> or C<dXSTARG> should be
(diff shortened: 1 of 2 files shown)


pp.h
```diff
@@ -472,13 +488,15 @@ Does not use C<TARG>.  See also C<L</XPUSHu>>, C<L</mPUSHu>> and C<L</PUSHu>>.
 #define PUSHs(s)	(*++sp = (s))
 #define PUSHTARG	STMT_START { SvSETMAGIC(TARG); PUSHs(TARG); } STMT_END
 #define PUSHp(p,l)	STMT_START { sv_setpvn(TARG, (p), (l)); PUSHTARG; } STMT_END
+#define PUSHpvs(s)      PUSHp(""s"", sizeof(s)-1)
```
ilmari: C++ requires spaces between concatenated string literals. But even better, use the [`STR_WITH_LEN`](https://perldoc.perl.org/perlapi#STR_WITH_LEN) macro.

ilmari: D'oh, I just realised that that can't be used as an argument to a macro, so it would have to be written out as

```c
STMT_START { Perl_sv_setpvn(aTHX_ STR_WITH_LEN(s)); PUSHTARG; } STMT_END
```

leonerd: Ahyes, spaces added and force-pushed.

I thought about the `STR_WITH_LEN` macro but as you say, there's an annoying macro-nesting problem to it. :( The current implementation is a little copy-pastey on the sizeof part, but avoids copypasting the far more subtle body of each variation on the PUSHp macro.

ilmari (approved):

PR #20148, opened 2022-08-24
POSIX: mbrlen,mbtowc: Use SvPVbyte not SvPV

khwilliamson:
It's better form to explicitly use the 'byte' form.  Just above, we
tested that the byte form is valid.

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -3200,7 +3200,7 @@ mblen(s, n = ~0)
             }
             else {
                 size_t len;
-                char * string = SvPV(byte_s, len);
+                char * string = SvPVbyte(byte_s, len);
                 if (n < len) len = n;
 #ifdef USE_MBRLEN
                 RETVAL = (SSize_t) mbrlen(string, len, &PL_mbrlen_ps);
@@ -3244,7 +3244,7 @@ mbtowc(pwc, s, n = ~0)
             }
             else {
                 size_t len;
-                char * string = SvPV(byte_s, len);
+                char * string = SvPVbyte(byte_s, len);
                 if (n < len) len = n;
                 RETVAL = mbtowc_(&wc, string, len);
                 if (RETVAL >= 0) {


FGasper:
- What are the odds that this changeset would break existing applications?
- Can you add tests to confirm the expected behaviour?

richardleach:
I'm not familiar with all the SvPV* functions, so I'd probably be confused in any case, but given:
```
            if (! sv_utf8_downgrade_nomg(byte_s, TRUE)) {
                SETERRNO(EINVAL, LIB_INVARG);
                RETVAL = -1;
            }
            else {
                size_t len;
                char * string = SvPVbyte(byte_s, len);
```
Seems like:
* `sv_utf8_downgrade_nomg` would downgrade the string if it's UTF-8?
* If so, we only enter the `else` block if the string wasn't UTF-8 or has been successfully downgraded?

sv.h says:
```
C<SvPVbyte> is like C<SvPV>, but converts C<sv> to byte representation first if
currently encoded as UTF-8.  If C<sv> cannot be downgraded from UTF-8, it
croaks.```
* Therefore `SvPVbyte` would do unnecessary work in examining the not-a-UTF-8 string before doing the same job as SvPV? 

If that's all actually the case, then it doesn't seem to me like this PR is necessary.

**Comment by FGasper** [comment]
@richardleach: SvPV is an abstraction leak: Perl’s internal storage of the logical string influences what C receives when reading the SV. It’s basically `use bytes` for C, though it’s occasionally legitimately useful for C code to know about Perl’s internals in this way.

The problem is that applications might _depend_ on that abstraction leak. See the POD for [Sys::Binmode](https://metacpan.org/pod/Sys::Binmode) for a fuller discussion.

**Comment by richardleach** [comment]
I infer  that in the commit message: "Just above, we tested that the byte form is valid." refers to:
```
if (! sv_utf8_downgrade_nomg(byte_s, TRUE)) {
```
Doesn't this mean that `char * string` will be assigned the same thing regardless of whether `SvPV` or `SvPVbyte` is used?

FGasper:
@richardleach Oops! Indeed, I'd missed that.

So, this appears just to be a cosmetic change, @khwilliamson?

khwilliamson:
I intended it to be a cosmetic change.  But was making sure.  I also wanted to let you know that it has sunk in with me anyway to avoid using plain SvPV

richardleach:
LGTM then.

I've looked at **Perl_SvPV_helper** for the first time ever and seen how much it gets reduced at compile time. Pretty nice.

I'm guessing that `SvPOK_byte_nog(sv)` should always be true here? If so, I wonder if common compilers can figure that out, or if we might want to help them avoid that check here with an `assume(SvPOK_byte_nog(sv));` ?

richardleach (approved):

FGasper (approved):
LGTM

PR #20150, opened 2022-08-24
Improve switching to a new LC_CTYPE locale

khwilliamson:
This series of commits caches the LC_CTYPE value so it doesn't need to be recalculated

And uses our built-in fold tables for the C and UTF-8 locales.  This saves work and avoids bugs in vendors' locale definitions

A new diagnostic is raised when someone switches to, say an ISO 2022, multibyte locale that is not UTF-8.  These are unsupported, and can crash the interpreter.  The message is thus marked as S

And new verbose debugging info is added.


khwilliamson:
Closed via e4bbbfe02b9e9aae521b164eba0e518ca478945f

PR #20151, opened 2022-08-25
Perl_sv_setsv_flags: return after invlist_clone case

richardleach:
In the "first_ switch" statement in **Perl_sv_setsv_flags**, there's this case:
```
    case SVt_INVLIST:
        invlist_clone(ssv, dsv);
        break;
```
It's not obvious clear that there's any point in this case continuing rather than returning from the function. This was briefly discussed on IRC earlier this year,  but no-one pointed out anything wrong with it or any reason for it, and it seemed a case of "if it ain't broke, don't fix it".

But looking at it again, it definitely seems inefficient. **S_initialize_invlist_guts** (called from **Perl_invlist_clone**, both in _regcomp.c_), does `SvPOK_on(invlist)`.

**Perl_invlist_clone** does some stuff and then copies the PV buffer from the old invlist to the new one:
`    Copy(SvPVX(invlist), SvPVX(new_invlist), physical_length, char);`
and then returns to **Perl_sv_setsv_flags**.

There, since `(sflags & SVp_POK)` is presumably true from the original inversion list, the function is going to try to (swipe or COW or) copy the PV buffer again!

Changing the `break;` to a `return;` doesn't seem to break any existing tests.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -4156,7 +4156,9 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
     SV_CHECK_THINKFIRST_COW_DROP(dsv);
     dtype = SvTYPE(dsv); /* THINKFIRST may have changed type */
 
-    /* There's a lot of redundancy below but we're going for speed here */
+    /* There's a lot of redundancy below but we're going for speed here
+     * Note: some of the cases below do return; rather than break; so the
+     * if-elseif-else logic below this switch does not see all cases. */
 
     switch (stype) {
     case SVt_NULL:
@@ -4244,7 +4246,7 @@ Perl_sv_setsv_flags(pTHX_ SV *dsv, SV* ssv, const I32 flags)
 
     case SVt_INVLIST:
         invlist_clone(ssv, dsv);
-        break;
+        return;
     default:
         {
         const char * const type = sv_reftype(ssv,0);


bram-perl:
> This was briefly discussed on IRC earlier this year, but no-one pointed out anything wrong with it or any reason for it, and it seemed a case of "if it ain't broke, don't fix it".

That's not how I remember the conversation (but I don't have logs of it)
IIRC the reason for `break;` instead of the `return;` was something like: today the remainder of the code (after the switch that is) is indeed irrelevant for `SVt_INVLIST` but a `break` was chosen to be 'future-safe' if the code after the switch is changed. [to be clear: that's rephrasing what I remember in my own words, that's not the exact thing that was said)

(To be clear: I'm *not* arguing against the change, only providing some context; I have no opinion on the actual change since I'm not familiar with it)

khwilliamson:
On 8/28/22 15:16, Bram wrote:
>     This was briefly discussed on IRC earlier this year, but no-one
>     pointed out anything wrong with it or any reason for it, and it
>     seemed a case of "if it ain't broke, don't fix it".
> 
> That's not how I remember the conversation (but I don't have logs of it)
> IIRC the reason for |break;| instead of the |return;| was something 
> like: today the remainder of the code (after the switch that is) is 
> indeed irrelevant for |SVt_INVLIST| but a |break| was chosen to be 
> 'future-safe' if the code after the switch is changed. [to be clear: 
> that's rephrasing what I remember in my own words, that's not the exact 
> thing that was said)
> 
> (To be clear: I'm /not/ arguing against the change, only providing some 
> context; I have no opinion on the actual change since I'm not familiar 
> with it)
> 

What I said at the time is effectively that early returns are risky from 
a future maintenance point of view.  Its not really an issue for short 
functions, but this has extensive code after the switch.

I wrote the invlist code, and I had no idea if the stuff that came later 
needed to be done or not; there might be magic handling that needed to 
be done, so I didn't take a chance.  And someday someone could come 
along, not realize there is a return much earlier in the function, and 
expect the new code to be executed.

That said, somebody more familiar with how SV handling all fits together 
would know immediately that there's no problem.  So my concern isn't 
strongly held.

richardleach:
You both have much better recollection skills than I. Reading the above reminded me that was indeed how the discussion went.

I don't really mind whether this PR is merged or closed for the reasons above. It was just created to capture that there seems to be that double PV copy, which doesn't seem intentional.

bram-perl:
> And someday someone could come along, not realize there is a return much earlier in the function, and expect the new code to be executed.

An observation: there are several other cases in the switch that already do a `return;` so at least for some cases/some conditions it already skips the remainder of the code.

Two suggestions for this PR/commit:
- add a comment at the start of the `switch` (I guess somewhere on line 4161) that several `cases` call `return;` (and not `break;`)
- slightly alter the commit message and add the reasoning why it was a `break;` (i.e. what @khwilliamson said)

khwilliamson:
I want to make sure I'm clear that I do not  in general oppose early returns; it is just this has very extensive code after the switch.  I'm fine with @bram-perl suggestions

richardleach:
Thanks, both. I'll revise with Bram's suggestions.

richardleach:
@bram-perl / @khwilliamson - The revised commit message and additional comment are hopefully close to what you'd wanted.

bram-perl:
> @bram-perl / @khwilliamson - The revised commit message and additional comment are hopefully close to what you'd wanted.

Looks ok to me

PR #20152, opened 2022-08-25
run/locale.t: skip illegal locale test when all are legal

### LINKED ISSUE #20149: t/run/locale.t: recent commit causes test failure on OpenBSD with '*.UTF-8' locales
We have a new smoke-test rig for OpenBSD, but it is reporting a failure in `t/run/locale.t` I haven't seen before.  Please see [this report](https://perl5.test-smoke.org/report/5021937) for which [this is the logfile](https://perl5.test-smoke.org/report/5021937).

I can reproduce this problem on the OpenBSD-6.9 VM which I myself used for smoke-testing.
```
$ export LC_CTYPE=nl_NL.UTF-8 && sh ./Configure -des -Dusedevel && make test_prep
$ ./perl -Ilib t/run/locale.t
# locales available: C C.UTF-8 POSIX
ok 1 - /il matching of [bracketed] doesn't skip POSIX class if fails individ char
ok 2 - /l matching of [bracketed] doesn't skip non-first POSIX class
ok 3 # skip didn't find a suitable locale
ok 4 - check that setlocale overrides startup
not ok 5 - check that illegal startup environment falls back
# Failed test 5 - check that illegal startup environment falls back at ./test.pl line 1189
#      got ''
# expected /(?^:Falling back to the standard locale)/
# PROG: 
# STATUS: 0
1..5
```
Many tests in this file are skipped on OpenBSD.  What are numbered as tests 4 and 5 above are tests 41 and 42 in a similar build on FreeBSD with the same LC_CTYPE export.

Now, this locale *is* reported to be available on OpenBSD:
```
$ locale -a | ack 'nl_NL\.UTF-8'                                                 
nl_NL.UTF-8
```
... even though the test output above states: `# locales available: C C.UTF-8 POSIX`

And I get the same test failures with `LC_CTYPE=en_US.UTF-8`:
```
$ export LC_CTYPE=en_US.UTF-8 && sh ./Configure -des -Dusedevel && make test_prep && ./perl -Ilib t/run/locale.t
...
ok 4 - check that setlocale overrides startup
not ok 5 - check that illegal startup environment falls back
# Failed test 5 - check that illegal startup environment falls back at ./test.pl line 1189
#      got ''
# expected /(?^:Falling back to the standard locale)/
# PROG: 
# STATUS: 0
1..5
```
This test failure is probably a side effect of recent commit c3b01b791bb308adb1c4c7bbbe7590fef79b9dd0.
```
commit c3b01b791bb308adb1c4c7bbbe7590fef79b9dd0
AuthorDate: Sat Aug 20 11:13:39 2022 -0600
CommitDate: Tue Aug 23 13:54:08 2022 -0600

    Use Windows system default locale
    
    This fixes #20054
    
    In order to make Perl programs more portable, the perl locale system on
    Windows emulates POSIX behavior.  But Windows has a default system
    locale, missing from other platforms.  This really should be considered
    as a fallback when the POSIX behavior doesn't work.
    
    And it was, for a few releases, as noted in the ticket.  But
    inadvertently broken in 5.28; and this commit now restores it.
```
On OpenBSD-6.9, I configured and built at the commit immediately preceding the one above.  The test PASSed, with the following output:
```
# locales available: C C.UTF-8 POSIX
ok 1 - /il matching of [bracketed] doesn't skip POSIX class if fails individ char
ok 2 - /l matching of [bracketed] doesn't skip non-first POSIX class
ok 3 # skip didn't find a suitable locale
ok 4 - check that setlocale overrides startup
1..4
```
Other relevant tickets: https://github.com/Perl/perl5/issues/20054; https://github.com/Perl/perl5/pull/20127.

Thank you very much.

khwilliamson:
OpenBSD believes that locales are so inherently insecure that it only
internally has two locales: C and C.UTF-8, which it knows how to treat
securely.

However, to preserve at least the illusion of portability, the locale
setting operations do not fail whatever garbage is passed as the locale
name.  (There may be unacceptable strings, but any reasonable locale
name and a lot of unreasonable ones are accepted.)  If the input name
looks like it was meant to have "UTF-8" in its name, the locale is set
to C.UTF-8.  Otherwise, it is set to C.  Early releases had bugs in
switching between the two.

I wrote a Configure probe some years ago to detect this situation.

I forgot about this fact in adapting Bram's test case.  It should be
skipped on such platforms.

This fixes #20149

### DIFF
--- a/t/run/locale.t
+++ b/t/run/locale.t
@@ -571,6 +571,10 @@ SKIP: {   # GH #20085
 }
 
 SKIP: {   # GH #20054
+    skip "Even illegal locale names are accepted", 1
+                    if $Config{d_setlocale_accepts_any_locale_name}
+                    && $Config{d_setlocale_accepts_any_locale_name} eq 'define';
+	
     my @lc_all_locales = find_locales('LC_ALL');
     my $locale = $lc_all_locales[0];
     skip "LC_ALL not enabled on this platform", 1 unless $locale;


jkeenan:
LGTM
```
[openbsd69: perl] $ uname -mrs
OpenBSD 6.9 amd64
[openbsd69: perl] $ gitcurr
openbsd
[openbsd69: perl] $ git describe
v5.37.2-334-g8f8790e38b
[openbsd69: perl] $ cd t;./perl harness -v run/locale.t; cd - 

ok 1 - /il matching of [bracketed] doesn't skip POSIX class if fails individ char
ok 2 - /l matching of [bracketed] doesn't skip non-first POSIX class
ok 3 # skip didn't find a suitable locale
ok 4 - check that setlocale overrides startup
ok 5 # skip Even illegal locale names are accepted
ok
All tests successful.
Files=1, Tests=5,  0 wallclock secs ( 0.03 usr  0.00 sys +  0.15 cusr  0.12 csys =  0.30 CPU)
Result: PASS
```

Issue #20146, opened 2022-08-24
Printing package names as raw bytes can hide subtle bugs


leonerd:
My XS code did:

```c
#define mPUSHpvs(s)   mPUSHp("" s "", sizeof(s))
...
      mPUSHpvs("Future::Exception");
      ...
      call_method("new", G_SCALAR);
```

Which resulted in an error message:

```
Can't locate object method "new" via package "Future::Exception" (perhaps you forgot to load "Future::Exception"?) at ...
```

Despite the fact that I am **sure** I definitely loaded the package and the method is available.

The ultimate bug? I forgot to subtract 1 from `sizeof(s)` to account for the trailing nul. So the macro ought to have been:

```c
#define mPUSHpvs(s)   mPUSHp("" s "", sizeof(s)-1)
```

That is certainly a bug in my code. However, Perl's literal printing of the package name hid the bug from me. It throws raw bytes directly at the terminal:

```
$ perl -e '"ab\0cd"->new' 2>&1 | xxd
00000000: 4361 6e27 7420 6c6f 6361 7465 206f 626a  Can't locate obj
00000010: 6563 7420 6d65 7468 6f64 2022 6e65 7722  ect method "new"
00000020: 2076 6961 2070 6163 6b61 6765 2022 6162   via package "ab
00000030: 0063 6422 2028 7065 7268 6170 7320 796f  .cd" (perhaps yo
00000040: 7520 666f 7267 6f74 2074 6f20 6c6f 6164  u forgot to load
00000050: 2022 6162 0063 6422 3f29 2061 7420 2d65   "ab.cd"?) at -e
00000060: 206c 696e 6520 312e 0a                    line 1..
```

Note specifically that it printed a literal zero NUL byte, between "ab" and "cd".

This sort of printing ought to be escaped. If instead it had done, then messages like

```
Can't locate object method "new" via package "Future::Exception\0" (perhaps you forgot to load "Future::Exception\0"?)
```

would yield a vital clue as to what's gone wrong.

leonerd:
@demerphq I recall you recently created a "print this string with escapes and stuff", which might be handy to solve this very problem

demerphq:
On Thu, 25 Aug 2022 at 09:40, demerphq ***@***.***> wrote:
>
> On Wed, 24 Aug 2022 at 16:18, Paul Evans ***@***.***> wrote:
> >
> > @demerphq I recall you recently created a "print this string with escapes and stuff", which might be handy to solve this very problem
>
[...]

PR #20153, opened 2022-08-25
sv.c - add a _QUOTEDPREFIX version of SVf, UTF8f, and HEKf for use in…

demerphq:
… error messages.

These new formats are intended to be used in error messages where
we want to show the contents of a string without any possible
hidden characters not rendering in the error message, and where
it would be unreasonable to show every character of the string
if it is very long.

A good example would be when we want to say that a class name is
illegal. Consider:

    "Foo\0"->thing()

should not throw an error message about "Foo" being missing, the fact
there is a null in there should be visible to the developer.
Similarly if we had

    ("x" x 1000_000)->thing()

we also wouldn't want to throw a 1MB error message.

Currently this patch restricts it to 256 characters, eliding the middle of the string
when it is too long, but the docs are such that we can change that if we wish, 
I suspect something like 100 would be more reasonable. You can override the define
PERL_QUOTEDPREFIX_LEN to a longer value in Configure if you wish.

Example usage:

    other= newSVpvs("Some\0::Thing\n");
    sv_catpvf(msg_sv,"%" SVfQUOTEDPREFIX, SVfARG(other));

Should append

    "Some\0::Thing\n"

to the msg_sv. If it were very long it would have elipses infixed:

    "long prefix".."long suffix"

This patch changes the error messages related to method calls. I haven't
changed every place that this could happen yet, just the main ones.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -4152,8 +4152,28 @@ out there, Solaris being the most prominent.
 
 #define SVfARG(p) ((void*)(p))
 
+/* Render an SV as a quoted and escaped string suitable for an error message.
+ * Only shows the first PERL_QUOTEDPREFIX_LEN characters, and adds ellipses if the
+ * string is too long.
+ */
+#ifndef PERL_QUOTEDPREFIX_LEN
+# define PERL_QUOTEDPREFIX_LEN 256
+#endif
+#ifndef SVf_QUOTEDPREFIX
+#  define SVf_QUOTEDPREFIX "5p"
+#endif
+
+/* like %s but runs through the quoted prefix logic */
+#ifndef PVf_QUOTEDPREFIX
+#  define PVf_QUOTEDPREFIX "1p"
+#endif
+
 #ifndef HEKf
 #  define HEKf "2p"
+#endif
+
+#ifndef HEKf_QUOTEDPREFIX
+#  define HEKf_QUOTEDPREFIX "7p"
 #endif
 
 /* Not ideal, but we cannot easily include a number in an already-numeric
@@ -4162,19 +4182,28 @@ out there, Solaris being the most prominent.
 #  define HEKf256 "3p"
 #endif
 
+#ifndef HEKf256_QUOTEDPREFIX
+#  define HEKf256_QUOTEDPREFIX "8p"
+#endif
+
 #define HEKfARG(p) ((void*)(p))
 
 /* Documented in perlguts
  *
- * %4p is a custom format
+ * %4p and %9p are custom formats for handling UTF8 parameters.
+ * They only occur when prefixed by specific other formats.
  */
 #ifndef UTF8f
 #  define UTF8f "d%" UVuf "%4p"
+#endif
+#ifndef UTF8f_QUOTEDPREFIX
+#  define UTF8f_QUOTEDPREFIX "d%" UVuf "%9p"
 #endif
 #define UTF8fARG(u,l,p) (int)cBOOL(u), (UV)(l), (void*)(p)
 
 #define PNf UTF8f
 #define PNfARG(pn) (int)1, (UV)PadnameLEN(pn), (void *)PadnamePV(pn)
+
 
 #ifdef PERL_CORE
 /* not used; but needed for backward compatibility with XS code? - RMB
@@ -8121,7 +8150,7 @@ Allows one ending \0
 #define PERL_PV_ESCAPE_NONASCII     0x000400
 #define PERL_PV_ESCAPE_FIRSTCHAR    0x000800
 
-#define PERL_PV_ESCAPE_ALL            0x001000
+#define PERL_PV_ESCAPE_ALL          0x001000
 #define PERL_PV_ESCAPE_NOBACKSLASH  0x002000
 #define PERL_PV_ESCAPE_NOCLEAR      0x004000
 #define PERL_PV_PRETTY_NOCLEAR      PERL_PV_ESCAPE_NOCLEAR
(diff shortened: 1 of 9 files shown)


leonerd:
Worth noting that this would be a good fix for #20146

mg.c
```diff
@@ -1706,7 +1706,7 @@ Perl_magic_setsig(pTHX_ SV *sv, MAGIC *mg)
             if (sv) {
                 SV *tmp = sv_newmortal();
                 Perl_ck_warner(aTHX_ packWARN(WARN_SIGNAL), "No such signal: SIG%s",
-                                            pv_pretty(tmp, s, len, 0, NULL, NULL, 0));
+                                            pv_pretty(tmp, s, len, 0, NULL, NULL,0));
```
leonerd: This (unrelated?) whitespace change seems to have snuck in by mistake.

perl.h
```diff
@@ -4152,30 +4152,59 @@ out there, Solaris being the most prominent.
 
 #define SVfARG(p) ((void*)(p))
 
+/* Render an SV as a quoted and escaped string suitable for an error message.
+ * Only shows the first PERL_QUOTEDPREFIX_LEN characters, and adds elipses if the
```
leonerd: s/elipses/ellipsis/

pod/perlguts.pod
```diff
@@ -3021,6 +3021,16 @@ You can use this to concatenate two scalars:
  SV *var3 = newSVpvf("var1=%" SVf " and var2=%" SVf,
                      SVfARG(var1), SVfARG(var2));
 
+=for apidoc Amnh||SVf_QUOTEDPREFIX
+
+C<SVf_QUOTEDPREFIX> is similar to C<SVf> except that it restricts the
+number of the characters printed, showing at most the first
+C<PERL_QUOTEDPREFIX_LEN> characters of the argument, and rendering it with
+double quotes and with the contents escaped using double quoted string
+escaping rules. If the string is longer than this then elipses "..."
```
leonerd: s/elipses/ellipsis/

pod/perlguts.pod
```diff
@@ -3021,6 +3021,16 @@ You can use this to concatenate two scalars:
...
+will be appended after the trailing quote. This is intended for error
+messages where the string is assumed to be a class name.
```
leonerd, line 3032: "or method"... or presumably some other bits and pieces as well..? Might be worth trying to be a bit more general in the wording here.

sv.c
```diff
@@ -12132,6 +12133,8 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
 
         char c;                       /* the actual format ('d', s' etc) */
 
+        bool escape_it   = FALSE;     /* if this is a string should we quote and escape it? */
+
```
leonerd, line 12137: Extraneous blank line

sv.c
```diff
@@ -12523,23 +12541,40 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
...
-             *                  arg rather than a char* arg.
+             *                  arg rather than a char* arg. Use C<SVfARG()>
+             *                  to set up the argument properly.
              *                  (This was previously %_).
              *
-             * %-<num>p         Ditto but like %.<num>s (i.e. num is max width)
+             * %1p       (PVf_QUOTEDPREFIX). Like raw %s, but it is escaped
+             *                  and quoted.
+             *
+             * %5p       (SVfQUOTEDPREFIX) Like SVf, but length restricted, escaped
```
leonerd: Underscore in `SVf_QUOTEDPREFIX`

sv.c
```diff
@@ -12523,23 +12541,40 @@ Perl_sv_vcatpvfn_flags(pTHX_ SV *const sv, const char *const pat, const STRLEN p
...
+             * %7p       (HEKf_QUOTEDPREFIX) ... but escaped and quoted.
              *
              * %3p       (HEKf256) Ditto but like %.256s
+             * %8p       (HEKf256_QUOTEDPREFIX) ... but escaped and quoted
              *
              * %d%lu%4p  (UTF8f) A utf8 string. Consumes 3 args:
              *                       (cBOOL(utf8), len, string_buf).
              *                   It's handled by the "case 'd'" branch
              *                   rather than here.
+             * %d%lu%9p  (UTF8fQUOTEDPREFIX) .. but escaped and quoted.
```
leonerd: Underscore in `UTF8f_QUOTEDPREFIX`

leonerd (commented):
Overall: This seems like a useful ability to add.

I've put a few small comments.

One larger overall comment: The width numbers used in the format string seem quite arbitrary associations that are all picked out individually. Would a bitfield make more sense here; using some numerical flags to indicate the various parts of behaviour. That might make it easier to add new ones in future.

mg.c
demerphq: ack. willfix.

pod/perlguts.pod
demerphq: yeah, sorry, i think i forgot to spellcheck this commit. Ill take  care of those.

sv.c
demerphq: that one is intentional, the list is divided into groups of different declarations. with whitespace separators.

pod/perlguts.pod
demerphq: done

sv.c
demerphq: yep, good catch.

sv.c
demerphq: yep, sorted.

pod/perlguts.pod
demerphq: yep, done.

demerphq:
@leonerd : Just repeating what we discussed in chat, there is some awkardness with the formats used that might come up if we change existing meanings even though it would make sense to restructure it a bit, so we should do that in a separate PR from this one .  

Anyway, I think I included all the changes you requested except one about whitespace that was deliberate.  Also I have added some docs about why we are hijacking the %p format like this.

leonerd (approved):
LGTM

PR #20159, opened 2022-08-25
locale.c: Improve newnumeric()

khwilliamson:
This function is called when LC_NUMERIC is changed.  This series of commits reorders things so as to perform less work by performing things earlier that later things depend on.


khwilliamson:
merged via 992e38c75c3458a29260860388d37ee4204e864e

PR #20160, opened 2022-08-25
locale.c: Improve newcollate

khwilliamson:
newcollate is called when LC_COLLATE is set to a new locale.  This series of commits improves its operation

Some platforms require LC_CTYPE and LC_COLLATE to be the same locale in order to operate properly.  These commits now bring LC_CTYPE into sync with LC_COLLATE for the duration of the calculations.

More and better debugging statements are added

Memory is allocated when a string is collated.  The amount is roughly a linear function of the string length, varying depending on the locale.  To save some recalculations, perl calculates and saves coefficients for the current locale's equation once, revising them if necessary.  Prior to this commit, this  somewhat expensive calculation was done every time the locale changed.  Now it is deferred until the first time it is needed.


jkeenan:
There are no tests in this pull request.  Do we already have tests that exercise the interaction between LC_CTYPE and LC_COLLATE?

khwilliamson:
A later commit in the series introduces tests

khwilliamson:
Closed via cd55125d69f5f698ef7cbdd650cda7d2e59fc388

PR #20163, opened 2022-08-26
gv.c - use SVf_QUOTEDPREFIX in error message

demerphq:
I overlooked this case when SVf_QUOTEDPREFIX was introduced

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -916,8 +916,8 @@ S_gv_fetchmeth_internal(pTHX_ HV* stash, SV* meth, const char* name, STRLEN len,
                 } else {
                     Perl_ck_warner(aTHX_ packWARN(WARN_SYNTAX),
                         "While trying to resolve method call %.*s->%.*s()"
-                        " can not locate package \"%" SVf "\" yet it is mentioned in @%.*s::ISA"
-                        " (perhaps you forgot to load \"%" SVf "\"?)",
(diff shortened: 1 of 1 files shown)


gv.c
```diff
@@ -916,8 +916,8 @@ S_gv_fetchmeth_internal(pTHX_ HV* stash, SV* meth, const char* name, STRLEN len,
                 } else {
                     Perl_ck_warner(aTHX_ packWARN(WARN_SYNTAX),
                         "While trying to resolve method call %.*s->%.*s()"
-                        " can not locate package \"%" SVf "\" yet it is mentioned in @%.*s::ISA"
-                        " (perhaps you forgot to load \"%" SVf "\"?)",
+                        " can not locate package %" SVf_QUOTEDPREFIX " yet it is mentioned in @%.*s::ISA"
+                        " (perhaps you forgot to load %" SVf_QUOTEDPREFIX "?)",
                          (int) hvnamelen, hvname,
                          (int) len, name,
```
leonerd, line 922: Since you're using `SVf_...` in the format, should these not be `SVfARG()`ed?

leonerd: Oops sorry, I've reread the code in more detail. They already are :)

leonerd (approved):
On a second, closer reading of the code, LGTM. Please ignore my earlier comment.

gv.c
demerphq: :-)

Issue #17397, opened 2019-12-28
NULL pointer dereference in S_pending_ident()


fcambus:
Hi,

While fuzzing Perl 5.30.1 with Honggfuzz, I found a NULL pointer dereference in the S_pending_ident() function, in toke.c.

Attaching a reproducer (gzipped so GitHub accepts it):  [test01.pl.gz](https://github.com/Perl/perl5/files/4007258/test01.pl.gz)

Issue can be reproduced by running:

```
perl test01.pl
```

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==13609==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x0000005857d1 bp 0x7ffd300e8150 sp 0x7ffd300e6f80 T0)
==13609==The signal is caused by a READ memory access.
==13609==Hint: address points to the zero page.
    #0 0x5857d0 in S_pending_ident /home/fcambus/perl-5.30.1/toke.c:9111:17
    #1 0x5857d0 in Perl_yylex /home/fcambus/perl-5.30.1/toke.c:4903:13
    #2 0x5d21cc in Perl_yyparse /home/fcambus/perl-5.30.1/perly.c:340:34
    #3 0x54cfa0 in S_parse_body /home/fcambus/perl-5.30.1/perl.c:2531:9
    #4 0x54cfa0 in perl_parse /home/fcambus/perl-5.30.1/perl.c:1822:2
    #5 0x4df38c in main /home/fcambus/perl-5.30.1/perlmain.c:126:10
    #6 0x7f1b520c11e2 in __libc_start_main /build/glibc-4WA41p/glibc-2.30/csu/../csu/libc-start.c:308:16
    #7 0x437bfd in _start (/home/fcambus/perl-5.30.1/perl+0x437bfd)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/fcambus/perl-5.30.1/toke.c:9111:17 in S_pending_ident
==13609==ABORTING
```

jkeenan:
What happens if you rewrite the test program to include:
```
use strict;
use warnings;
```
... and then re-fuzz it?

Thank you very much.
Jim Keenan

jkeenan:
FWIW, from at least perl-5.6.2 to perl-5.8.9, the test program fails to compile.
```
$ perlbrew use perl-5.6.2
$ perl ghi-17397-test01.pl 
[...]

Issue #16944, opened 2019-04-09
Segfault in Perl_ck_shift (op.c:12664)


p5pRT:
Migrated from [rt.perl.org#134012](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=134012) (status was 'new')

Searchable as RT134012$

p5pRT:
### From @dur-randir

#### Created by @dur-randir



While fuzzing perl v5\.29\.9\-63\-g2496d8f3f7 built with afl and run
under libdislocator\, I found the following program

s\,\,$0\[sub\{m\]\]\]\]\,\}>0\,shift

to segfault\. GDB stack trace is following

\#0  0x00005555555d8062 in Perl\_ck\_shift \(o=0x555555b77440\) at op\.c&#8203;:12664
\#1  0x00005555555b6c74 in Perl\_newOP \(type=165\, flags=0\) at op\.c&#8203;:6174
\#2  0x000055555566fc44 in Perl\_yyparse \(gramtype=258\) at perly\.y&#8203;:1192
\#3  0x00005555555ec726 in S\_parse\_body \(env=0x0\, xsinit=0x5555555a11f8
\<xs\_init>\) at perl\.c&#8203;:2531
\#4  0x00005555555ea9f8 in perl\_parse \(my\_perl=0x555555b4c260\,
xsinit=0x5555555a11f8 \<xs\_init>\, argc=2\, argv=0x7fffffffe1c8\, env=0x0\)
at perl\.c&#8203;:1822
\#5  0x00005555555a113b in main \(argc=2\, argv=0x7fffffffe1c8\,
env=0x7fffffffe1e0\) at perlmain\.c&#8203;:126

This is present a regression between 5\.16 and 5\.18\, bisect points to

commit 9ffcdca1f504cb09088413c074b35af4b7f247e3
Author&#8203;: Father Chrysostomos \<sprout@&#8203;cpan\.org>
Date&#8203;:   Mon Nov 12 23&#8203;:04&#8203;:16 2012 \-0800

&nbsp;    Don’t leak subs containing syntax errors

&nbsp;    I fixed this for BEGIN blocks earlier\, but missed the fact that
&nbsp;    all subs are affected\.

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=low

Site configuration information for perl 5.29.9:

Configured by dur-randir at Wed Feb 27 14:51:01 MSK 2019.

[...]

Issue #16057, opened 2017-06-29
unknow address reference in __memmove_ssse3_back


p5pRT:
Migrated from [rt.perl.org#131677](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=131677) (status was 'open')

Searchable as RT131677$

p5pRT:
### From @Mipu94

Hi there\,
I’ve just found out one more issue\. The poc attached below triggered a
crash because register $rdi reference to unknow andress in
function \_\_memmove\_ssse3\_back\.


This is perl 5\, version 27\, subversion 1 \(v5\.27\.1\-37\-g4c95ee9\) built for
x86\_64\-linux

Program received signal SIGSEGV\, Segmentation fault\.
\[\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-registers\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\]
RAX&#8203;: 0x0
RBX&#8203;: 0x7f6c90 \-\-> 0x7f6c78 \-\-> 0x7f6cd8 \-\-> 0xc70098 \-\-> 0xc700b0 \-\->
0x7f6e10 \(\-\-> \.\.\.\)
RCX&#8203;: 0xb60 \('\`\\x0b'\)
RDX&#8203;: 0x0
RSI&#8203;: 0x54fb61 \-\-> 0x73303432250012
RDI&#8203;: 0x1
RBP&#8203;: 0x1
RSP&#8203;: 0x7fffffffdf58 \-\-> 0x4ceb11 \(\<Perl\_sv\_setpv\+97>&#8203;:   mov
rax\,QWORD PTR \[rbx\]\)
RIP&#8203;: 0x7ffff720b844 \(\<\_\_memmove\_ssse3\_back\+9252>&#8203;:       mov    BYTE
PTR \[rdi\-0x1\]\,dl\)
R8 : 0x29 \('\)'\)
R9 : 0x2
R10&#8203;: 0x30 \('0'\)
R11&#8203;: 0x7ffff7241350 \-\-> 0xfffca4f0fffc9d30
R12&#8203;: 0x54fb60 \-\-> 0x7330343225001200
R13&#8203;: 0x0
[...]

PR #20168, opened 2022-08-26
Stop parsing on first syntax error.

demerphq:
We try to keep parsing after many types of errors, up to a (current)
maximum of 10 errors. Continuing after a semantic error (like
undeclared variables) can be helpful, for instance showing a set of
common errors, but continuing after a syntax error isn't helpful
most of the time as the internal state of the parser can get confused
and is not reliably restored in between attempts. This can produce
sometimes completely bizarre errors which just obscure the true error,
and has resulted in security tickets being filed in the past.

This patch makes the parser stop after the first syntax error, while
preserving the current behavior for other errors. An error is considered
a syntax error if the error message from our internals is the literal
text "syntax error". This may not be a complete list of true syntax
errors, we can iterate on that in the future.

This fixes the segfaults reported in Issue https://github.com/Perl/perl5/issues/17397, and https://github.com/Perl/perl5/issues/16944 and
likely fixes other "segfault due to compiler continuation after syntax
error" bugs that we have on record, which has been a recurring issue
over the years.

The PR also includes a fix to another segfault/assert (Issue #16057) 
related prototypes on BEGIN blocks, which is in this PR because it
originally looked related to the stop on first error problem, and given it 
/is/ related to "stopping segfaults during compilation" it seems reasonable
to save some work and keep it in this PR.

Sorry for the weird wrapping of this ticket.


pp_ctl.c
```diff
@@ -1669,8 +1669,16 @@ Perl_qerror(pTHX_ SV *err)
...
         Perl_warn(aTHX_ "%" SVf, SVfARG(err));
-    if (PL_parser)
+
+    if (PL_parser) {
+        STRLEN len;
+        char *err_pv= SvPV(err,len);
         ++PL_parser->error_count;
+        if (memBEGINs(err_pv,len,"syntax error"))
+        {
+            PL_parser->error_count |= 128;
```
demerphq: I will abstract this in a future fixup of this patch.

t/comp/retainedlines.t
```diff
@@ -4,7 +4,7 @@
 
 # Uncomment this for testing, but don't leave it in for "production", as
 # we've not yet verified that use works.
-# use strict;
+#use strict;
```
demerphq: will fix this too.

t/comp/retainedlines.t
```diff
@@ -55,7 +55,7 @@ sub check_retained_lines {
     # Is there a more efficient way to write this?
     my @expect_lines = (undef, map ({"$_\n"} split "\n", $prog), "\n", ';');
 
-    my @keys = grep {!$seen{$_}} grep { /eval/ } keys %::;
+    my @keys = grep {!$seen{$_}} grep { /eval/ } sort keys %::;
```
demerphq: Required to get consistent results when the test fails.

t/comp/retainedlines.t
```diff
@@ -67,7 +67,7 @@ sub check_retained_lines {
     for (0..$#expect_lines) {
 	is ($got_lines[$_], $expect_lines[$_], "Line $_ is correct");
     }
-    $seen{$keys[0]}++;
+    $seen{$_}++ for @keys;
```
demerphq: Again, required to prevent one test failure cascading into the following tests.

t/comp/retainedlines.t
```diff
@@ -120,14 +120,15 @@ foreach my $flags (0x0, 0x800, 0x1000, 0x1800) {
 	    "evals that don't define subroutines are correctly cleaned up");
     }
 
+    @before = grep { /eval/ } keys %::;
```
demerphq: Will remove this.

khwilliamson:
I'm somewhat leery of this.

I wrote a Lisp interpreter in Snobol for a school assignment.  Snobol stops at the first error.  This was using punch cards and the turnaround time was 8-12 hours during the day, dropping to .5 hr at 3am.  It was awful.  I have hated compilers that don't try to recover ever since.

I can see giving up the current section of code if there are several errors in a few adjacent lines.  But why not then skip ahead some looking for a semi colon immediately followed by a new line, and continue trying from there?

demerphq:
Hey @khwilliamson can I ask you to try it before you pass judgement? My experience is that the storm of error messages from perl getting confused is really unhelpful, and results in old-times teaching new-comers to "go through it and look for the lowest line number" and such things. I have been using it for some hacking and i have been quite pleased with it. At first it was a little jarring, I'm used to having to sift through a heap of ridiculous and meaningless errors to find the one that is relevant and with the patch it "throws your eyeballs" a bit not having all that crap there, but I got quickly used to it and when i go back to the old rules it throws me the other way now. :-)

If you have thoughts how to safely restart parsing at a semicolon then i think you could do a follow up patch, but given the vagaries of parsing perl IMO that might not be quite as useful as you think: consider code like: `if if (/;/) { ... }`. It seems like something that is pretty darn hard to do well in perl, so we shouldnt try and stop giving users bogus errors.

Also, I am happy to make this a configurable option with whatever default we want. If people like the storm of hallucinatory errors that perl produces from common syntax errors then they are welcome to build with them, so long as I am welcome to build without them. :-)

demerphq:
Also note that this patch includes a revert of a patch from https://github.com/Perl/perl5/issues/16300 which caused breakage with Module::Install. We need to decide what to do about that.

demerphq:
I have added a workaround patch for the issue in Module::Install::DSL. We convert INIT blocks from that namespace to be BEGIN blocks. I thought about the added restriction of "INIT blocks in an eval", but it didnt seem necessary. With that i think in theory this PR should be "ok" to go and not cause havok in the CPAN river.

hvds:
> I'm somewhat leery of this.
[...]
> I can see giving up the current section of code if there are several errors in a few adjacent lines. But why not then skip ahead some looking for a semi colon immediately followed by a new line, and continue trying from there?

@khw note that "stop on first error" has for quite a while been advocated by @iabyn, if I remember correctly - we're not good at ensuring everything is restored to a valid state after an error, and the attempt to continue after errors has been the source of numerous security issues in the past. (That said, I think they were all _rejected_ as security issues, because they needed code from an untrusted source to exploit - but they also cost us a lot of effort to analyse.)

For me it is second nature to use the strategy @demerphq mentions - to scan a screed of garbage on the screen for the lowest-mentioned line number - but I'm always aware when doing so that a) I'm making up for perl's failings in doing so, and b) that someone new to perl probably won't know about that strategy.

demerphq:
> @khw note that "stop on first error" has for quite a while been advocated by @iabyn, if I remember correctly

Thanks @hvds - he actually pointed me at the code that started me down this rabbit hole. This task has lead to a bunch of bugs and old issues being identified and fixed.

demerphq:
FWIW #20181 and #20182 need to be applied in that order, and then the code from this can be applied on top. Ill rebase as things get merged.

demerphq:
@leonerd you expressed some interest in this, it is now out of draft and ready for merge.

toke.c
```diff
@@ -12644,22 +12653,39 @@ Perl_yyerror_pvn(pTHX_ const char *const s, STRLEN len, U32 flags)
...
     }
-    if (s == NULL || PL_error_count >= 10) {
-        const char * msg = "";
+    if ( s == NULL ||
+         PL_error_count >= PERL_STOP_PARSING_AFTER_N_ERRORS
+    ) {
         const char * const name = OutCopFILE(PL_curcop);
+        SV * errsv = NULL;
+        U8 raw_error_count= PERL_PARSE_ERROR_COUNT(PL_error_count);
+        U8 syntax_error= PERL_PARSE_IS_SYNTAX_ERROR(PL_error_count);
```
leonerd: U8? Surely a `bool`?

demerphq: Its a bitmap test on PL_error_count which is U8. I can convert it to bool on the fly, but there doesn't seem much point.  It boils down to:

```
 U8 syntax_error = PL_error_count & 128;
```

demerphq: Anyway, changed it to a bool.

demerphq:
squashed them down to a single patch now.

leonerd (approved):
Overall seems a reasonable direction. I'm not hugely a fan of special-casing the exception message "syntax error"; but I gather this is just a first-step in the direction of having a better mechanism - such as a dedicated abort-the-parse function.

demerphq:
>but I gather this is just a first-step in the direction of having a better mechanism - such as a dedicated abort-the-parse function.

Yes, indeed, that will come once I start getting some feedback. Please let me know if you encounter something that should stop the parse but doesnt.

I will merge this!

demerphq:
On Sun, 28 Aug 2022 at 20:05, Karl Williamson ***@***.***>
wrote:

> I'm somewhat leery of this.
>
> I wrote a Lisp interpreter in Snobol for a school assignment. Snobol stops
> at the first error. This was using punch cards and the turnaround time was
> 8-12 hours during the day, dropping to .5 hr at 3am. It was awful. I have
> hated compilers that don't try to recover ever since.
>
I can see giving up the current section of code if there are several errors
> in a few adjacent lines. But why not then skip ahead some looking for a
> semi colon immediately followed by a new line, and continue trying from
> there?
>

 I think you misunderstand the patch. It doesnt stop on first error. it
stop on first *syntax* error, which is where the compiler gets totally
confused. For other errors the old behavior is preserved.

Personally I *strongly* disagree with continuing after a syntax error. The
storm of bogus errors from restarting the parse obscures the true error in
a storm of halucinatory rubbish. I dont think trying to find the first
semicolon is reliable. Perl syntax is too messed up. At my old job I had to
do a lot of hand holding of people new to perl and one of the things they
complained about was the rubbish errors. Its a bit embarrasing when you
have to train people to ignore most of the mesages and root through them to
find the gem hidden in a pile of dung.

Why dont you try the branch and see what you think? So far I have found it
quite nice. When I mess up and make a syntax error there is one error
message, and it's always correct.  I dont have to stare at 9 other errors
which are figments of perl imagination. In fact I found it a little
confusing at first as I am so used to ignoring most of the error messages
Perl produces that when it gave me a single error that was correct it threw
me my a bit (in a good way): "Where did all the garbarge go?".

I think until you have tried the branch you should reserve judgement, i
dont think extrapolation from snobol is a reasonable thing to do. Perl is a
very different language.

Anyway, if you have ideas on how to sanely restart the parser somewhere
else then go for it.

Yves

-- 

PR #20169, opened 2022-08-27
retainedlines.t - deterministic results and fixup tests under failure

demerphq:
The existing logic is only "correct" when all the tests pass. When they
fail the tests are revealed to be non-determinstic. The other problem is
that once a single test case fails and leaks an entry in the stash the
following tests are all contaminated and trigger "false failures". The
combination makes things look much more broken than they are.

This patch fixes the tests so that they are deterministic and the
"leaking" between test cases is stopped.

### DIFF
--- a/t/comp/retainedlines.t
+++ b/t/comp/retainedlines.t
@@ -55,7 +55,15 @@ sub check_retained_lines {
     # Is there a more efficient way to write this?
     my @expect_lines = (undef, map ({"$_\n"} split "\n", $prog), "\n", ';');
 
-    my @keys = grep {!$seen{$_}} grep { /eval/ } keys %::;
+    # sort in decreasing number so that $keys[0] is the from the most
+    # recent eval. In theory we should only have one, but if something
+    # breaks we might have more than one, and keys will return them in a
+    # random order, so if we dont do this then failing tests will have
+    # inconsistent results from run to run.
+    my @keys = map { $_->[0] }
+               sort { $b->[1] <=> $a->[1] }
+               map { (!$seen{$_} and /eval (\d+)/) ? [ $_, $1 ] : ()  }
+               keys %::;
 
     is ((scalar @keys), 1, "1 new eval");
 
@@ -67,7 +75,10 @@ sub check_retained_lines {
     for (0..$#expect_lines) {
 	is ($got_lines[$_], $expect_lines[$_], "Line $_ is correct");
     }
-    $seen{$keys[0]}++;
+    # if we are "leaking" evals we only want to fail the current test,
+    # so we need to mark them all seen (older code only marked $keys[0]
+    # seen and this caused tests to fail that actually worked properly.)
+    $seen{$_}++ for @keys;
 }
 
 my $name = 'foo';


jkeenan:
> The existing logic is only "correct" when all the tests pass. When they fail the tests are revealed to be non-determinstic. The other problem is that once a single test case fails and leaks an entry in the stash the following tests are all contaminated and trigger "false failures". The combination makes things look much more broken than they are.
> 

Is there a GH issue or other URL where we can observe this problem and then corroborate that the p.r. fixes it?

demerphq:
On Sat, 27 Aug 2022 at 15:12, James E Keenan ***@***.***> wrote:
>
> The existing logic is only "correct" when all the tests pass. When they fail the tests are revealed to be non-determinstic. The other problem is that once a single test case fails and leaks an entry in the stash the following tests are all contaminated and trigger "false failures". The combination makes things look much more broken than they are.
>
> Is there a GH issue or other URL where we can observe this problem and then corroborate that the p.r. fixes it?

The PR for #20170 commit:dcc2422a6520a2f4bab8703bc96ac4776dda7d72
contains changes that will show t/comp/retainedlines.t failing tests.
Run it a few times. It will show random and widespread breakage. The
patch in this PR commit:080ae9b8b9629d9ea2fba723d4bbdcc5e530be45 will
make it stable will show that the only failures are due to the "eval
fail case".

git fetch
git reset --hard origin/blead
git cherry-pick dcc2422a6520a2f4bab8703bc96ac4776dda7d72
make -j8
./perl -Ilib t/comp/retainedlines.t
./perl -Ilib t/comp/retainedlines.t
./perl -Ilib t/comp/retainedlines.t

git cherry-pick 080ae9b8b9629d9ea2fba723d4bbdcc5e530be45
./perl -Ilib t/comp/retainedlines.t



-- 

jkeenan:
> On Sat, 27 Aug 2022 at 15:12, James E Keenan ***@***.***> wrote: The existing logic is only "correct" when all the tests pass. When they fail the tests are revealed to be non-determinstic. The other problem is that once a single test case fails and leaks an entry in the stash the following tests are all contaminated and trigger "false failures". The combination makes things look much more broken than they are. Is there a GH issue or other URL where we can observe this problem and then corroborate that the p.r. fixes it?
> The PR for #20170 commit:dcc2422a6520a2f4bab8703bc96ac4776dda7d72 contains changes that will show t/comp/retainedlines.t failing tests. Run it a few times. It will show random and widespread breakage. The patch in this PR commit:080ae9b8b9629d9ea2fba723d4bbdcc5e530be45 will make it stable will show that the only failures are due to the "eval fail case". git fetch git reset --hard origin/blead git cherry-pick [dcc2422](https://github.com/Perl/perl5/commit/dcc2422a6520a2f4bab8703bc96ac4776dda7d72) make -j8 ./perl -Ilib t/comp/retainedlines.t ./perl -Ilib t/comp/retainedlines.t ./perl -Ilib t/comp/retainedlines.t git cherry-pick [080ae9b](https://github.com/Perl/perl5/commit/080ae9b8b9629d9ea2fba723d4bbdcc5e530be45) ./perl -Ilib t/comp/retainedlines.t
> […](#)
> -- perl -Mre=debug -e "/just|another|perl|hacker/"

Thanks.  If and when this commit gets merged, let's include URL to comment above so that we have the evidence for future lookup.

tonycoz (approved):

Issue #20161, opened 2022-08-26
Evals that die under the debugger may or may not leak eval data depending on how they die.


demerphq:
**Description**
If eval string is used when the debugger is enabled then we may or may not leak debugging meta data in the stash depending on how the eval fails. 

When `$^P` is set to an integer with the `0x400` bit set (and maybe other bits) the debugger mode is enabled and when code is compiled a copy of the lines that were compiled is stashed away in `%::` in a slot named `"_<(eval $eval_ctr)"`, we make attempts to arrange that this slot is deleted if the eval fails. However there are a variety of ways that the compilation phase of eval may fail. If the failure triggers a `Perl_croak()` then we leak the debugging data and do not delete the entry.

Two easy ways to see this are to `eval "BEGIN{die}"`, and another is trigger 10 or more syntax errors in the eval text, which will cause the parser to trigger a croak. Any error pathway that triggers a croak may cause this outcome and there are a handful of special cases where the parser /will/ croak.

Thus `"1+"` will fail compile in a way that cleans up `"_<(eval $eval_ctr)"`, but `"1+;1+;1+;1+;1+;1+;1+;1+;1+;1+;"` will trigger a croak and the `"_<(eval $eval_ctr)"` will be left behind.

`t/comp/retainedlines.t` attempts to test this behavior, but it only uses very simple code to eval, and it doesnt trigger a croak, which exposes the issue. I noticed this because I am trying to make the parser stop after the first syntax error it encounters, and this is easiest done by croaking, but that exposes the underlying bug in this ticket which exists with or without the new logic.

**Steps to Reproduce**
This should output "0 0", it outputs "0 1" instead.

```
perl -le'for (1,10) { $^P=0x400; eval "1+;" x $_; my $count=0; /eval/ and $count++ for keys %::; push @ret,$count } print "@ret"'
0 1
```

**Expected behavior**
We should clean up the entries consistently. The output from the above script should be `"0 0"`.

Or I guess alternatively we shouldn't care about these "stash droppings". Although that seems like a great way to make the debugger leak memory. On the other hand, its not clear to me if we clean up these droppings if the eval is successful, so I am not sure if we should care. It feels like we should, but maybe not.

**Perl configuration**
```
$ ./perl -Ilib -V
Summary of my perl5 (revision 5 version 37 subversion 4) configuration:
  Derived from: ebc2efe6583f7cf7ea29a7f94f7b7887a75ad3a7
[...]

PR #20170, opened 2022-08-27
Tests for Issue #20161

demerphq:
This is to demonstrate the issue reported in GH Issue #20161,
that we leak debug data from eval statements that croak during
compilation.

We fail test when the parser croaks during compilation, but the old code
did not trigger a croak. With this change the test fails in a large
number of ways, many of which are false positives. A follow up patch
will remove the false positives, but still leave failing tests.


demerphq:
Note that if this is applied it really should be applied on top of https://github.com/Perl/perl5/pull/20169 otherwise the tests will emit a large number of bogus failures along with the true failures and produce inconsistent results from run to run.

demerphq:
Not needed anymore.Supplanted by other PR's.

PR #20171, opened 2022-08-27
Add a perldelta for SVf_QUOTEDPREFIX changes to error messages

demerphq:
I forgot to add a delta in the original patch.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -241,13 +241,71 @@ L<Locale '%s' is unsupported, and may crash the interpreter.message|perldiag/"Lo
 
 =head2 Changes to Existing Diagnostics
 
-XXX Changes (i.e. rewording) of diagnostic messages go here
-
 =over 4
 
 =item *
 
-XXX Describe change here
+Error messages that output class or package names have been modified to
+output double quoted strings with various characters escaped so as to
+make the exact value clear to a reader. The exact rules on which
+characters are escaped may change over time but currently are that
+printable ASCII codepoints, with the exception of C<"> and C<\>, and
+unicode word characters whose codepoint is over 255 are output raw, and
+any other symbols are escaped much as Data::Dumper might escape them,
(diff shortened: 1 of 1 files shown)


pod/perldelta.pod
```diff
@@ -226,13 +226,65 @@ XXX L<message|perldiag/"message">
...
 
-XXX Changes (i.e. rewording) of diagnostic messages go here
-
 =over 4
 
 =item *
 
-XXX Describe change here
+Error messages that output class or package names have been modified to
+output double quoted strings with non-word characters escaped. If a
```
leonerd: Perhaps you mean "non-printable"? Numbers and other punctuation are usually considered "non-word" but they're still handled literally:

```
$ bleadperl -e '"a+b0c\nd"->new'
Can't locate object method "new" via package "a+b0c\nd" (perhaps you forgot to load "a+b0c\nd"?) at -e line 1.
```

demerphq: Actually digits and underbars are word chars in perl:

```
$ perl -le'print "1"=~/\w/ ? "word" : "not word"'
word
```

But yes, its a good point. I will rephrase it. The logic is actually a bit more complicated. If the codepoint is ASCII then it will be escaped if it is non printable or a double quote or backslash, and if it is larger than 255 it will be escaped if it is not a wordchar. 

Not exactly the easiest thing to sum up. We dont want to make error messages unreadable for people working with unicode package names in other languages, and in theory identifiers should match IDFIRST IDREST+, which basically means "word characters", along of course with the colon, which we don't want to escape. Anyway, I am not certain that preciseness here is necessary, ill figure out a different wording.

demerphq: Anyway, new wording pushed.

demerphq:
@leonerd ping. Rebased/squashed the fixup.

leonerd (approved):
LGTM

PR #20177, opened 2022-08-28
op.c - Restrict nested eval/BEGIN blocks to a user controllable maximum

### LINKED ISSUE #20176: segfault with BEGIN and eval
**Description**
Our guards against stackoverflow from recursion dont cover the case of a BEGIN inside of an eval. With appropriately crafted code we can trigger a segfault inside of an eval.

**Steps to Reproduce**
```
./perl -le'sub f{ eval "BEGIN{ f() }" } f()'
```

**Expected behavior**
Some kind of catchable exception, not a segfault.

**Perl configuration**
```
$ ./perl -Ilib -V
Summary of my perl5 (revision 5 version 37 subversion 4) configuration:
  Commit id: 6dacde6e70aa89d33e9b44e2c00050775261d8fb
  Platform:
    osname=linux
    osvers=5.14.0-1049-oem
    uname='linux oncidium 5.14.0-1049-oem #56-ubuntu smp fri aug 12 10:23:08 utc 2022 x86_64 x86_64 x86_64 gnulinux '
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='gcc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-g'
    ccversion=''
    gccversion='9.4.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='gcc'
    libc=libc-2.31.so
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.31'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'
    lddlflags='-shared -g -L/usr/local/lib -fstack-protector-strong'


  Compile-time options:
    DEBUGGING
    HAS_TIMES
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_HASH_FUNC_SIPHASH13
    PERL_HASH_USE_SBOX32
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_TRACK_MEMPOOL
    PERL_USE_DEVEL
    PERL_USE_SAFE_PUTENV
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
    USE_REENTRANT_API
    USE_THREAD_SAFE_LOCALE
  Built under linux
  Compiled at Aug 28 2022 10:54:12
  %ENV:
    PERLBREW_CONFIGURE_FLAGS="-de -Dcc=ccache\ gcc -Dld=gcc"
    PERLBREW_HOME="/home/yorton/.perlbrew"
    PERLBREW_MANPATH="/home/yorton/perl5/perlbrew/perls/perl-5.34.1/man"
    PERLBREW_PATH="/home/yorton/perl5/perlbrew/bin:/home/yorton/perl5/perlbrew/perls/perl-5.34.1/bin"
    PERLBREW_PERL="perl-5.34.1"
    PERLBREW_ROOT="/home/yorton/perl5/perlbrew"
    PERLBREW_SHELLRC_VERSION="0.88"
    PERLBREW_VERSION="0.88"
  @INC:
    lib
    /usr/local/lib/perl5/site_perl/5.37.4/x86_64-linux-thread-multi
    /usr/local/lib/perl5/site_perl/5.37.4
    /usr/local/lib/perl5/5.37.4/x86_64-linux-thread-multi
    /usr/local/lib/perl5/5.37.4
```

demerphq:
Nested BEGIN blocks can cause us to segfault by exhausting
the C stack. Eg:

    perl -le'sub f { eval "BEGIN { f() }" } f()'

will segfault. This adds a new interpreter var PL_eval_begin_nest_depth
to keep track of how many layer of eval/BEGIN we have seen, and a new
reserved variable called ${^MAX_NESTED_EVAL_BEGIN_BLOCKS} which can be
used to raise or lower the limit. When set to 0 it blocks BEGIN entirely,
which might be useful from time to time.

This fixes https://github.com/Perl/perl5/issues/20176

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -528,6 +528,38 @@ C<-Accflags=-DNO_PERL_HASH_ENV>.
 The C<PERL_HASH_SEED_DEBUG> environment variable can be disabled by
 configuring perl with C<-Accflags=-DNO_PERL_HASH_SEED_DEBUG>.
 
+=head3 MISCELLANEOUS CONFIG
+
+Perl uses various defines to control defaults for its behavior. These
+values are chosen to represent "sane" config, but users can override
+these values in their builds if they wish. This is a list of such
+settings.
+
+=over 2
+
+=item PERL_MAX_NESTED_EVAL_BEGIN_BLOCKS_DEFAULT
+
+This define is used to control the default maximum number of nested
+eval/BEGIN statements, and in this context require should be
+understood to be a special form of eval so this means require/BEGIN
+and "use" statements as well.
+
+Currently each C<BEGIN> block inside of an C<eval EXPR> or C<require>
+operation will use a fairly high number of frames of the perl internal
+C stack, and this value is used to prevent stack overflows. Normally
+it is defaulted to 1000 but the default can be configured to another
+value, for instance 100, like this
+
+  -Accflags='-DPERL_MAX_NESTED_EVAL_BEGIN_BLOCKS_DEFAULT=100'
+
+
+If you don't know what this is then it is safe to ignore it. Do not
+configure this to 0 or another very low value, it will break a lot of
+code. If you want to set it to a low value use the run time variable
+C<${^MAX_NESTED_EVAL_BEGIN_BLOCKS}> instead.
+
+=back
+
 =head3 SOCKS
 
 Perl can be configured to be 'socksified', that is, to use the SOCKS
--- a/op.c
+++ b/op.c
@@ -10717,10 +10717,72 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
     if (*name == 'B') {
         if (strEQ(name, "BEGIN")) {
             const I32 oldscope = PL_scopestack_ix;
+            SV *max_nest_sv = NULL;
+            IV max_nest_iv;
             dSP;
             (void)CvGV(cv);
             if (floor) LEAVE_SCOPE(floor);
             ENTER;
+
+            /* make sure we don't recurse too deeply into BEGIN blocks
+             * but let the user control it via the new control variable
+             *
+             *   ${^MAX_NESTED_EVAL_BEGIN_BLOCKS}
+             *
+             * Note this *looks* code like when max_nest_iv is 1 that it
+             * would block the following code:
+             *
+             * BEGIN { $n |= 1; BEGIN { $n |= 2; BEGIN { $n |= 4 } } }
+             *
+             * but it does *not*, this code will happily execute when
+             * the nest limit is 1. The reason is revealed in the
+             * execution order. If we could watch $n in this code we
+             * would see the follow order of modifications:
+             *
+             * $n |= 4;
+             * $n |= 2;
+             * $n |= 1;
+             *
+             * This is because nested BEGIN blocks execute in FILO
+             * order, this is because BEGIN blocks are defined to
+             * execute immediately they are closed. So the innermost
+             * block is closed first, and it executes, which would the
+             * eval_begin_nest_depth by 1, it would finish, which would
+             * drop it back to its previous value. This would happen in
+             * turn as each BEGIN was terminated.
+             *
+             * The *only* place these counts matter is when BEGIN in
+             * inside of some kind of eval, either a require or a true
+             * eval. Only in that case would there be any nesting and
+             * would perl try to execute a BEGIN before another had
+             * completed.
+             *
+             * Thus this logic puts an upper limit on module nesting.
+             * Hence the reason we let the user control it, although its
+             * hard to imagine a 1000 level deep module use dependency
+             * even in a very large codebase. The real objective is to
+             * prevent code like this:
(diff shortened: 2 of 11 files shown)


INSTALL
```diff
@@ -528,6 +528,31 @@ C<-Accflags=-DNO_PERL_HASH_ENV>.
...
+=over 2
+
+=item PERL_MAX_NESTED_EVAL_BEGIN_BLOCKS_DEFAULT
+
+This define is used to control the default maximum number of nested
+eval/BEGIN statements. Currently each C<BEGIN> block inside of an C<eval
+EXPR> operation will uses a certain number of frames of the perl
+internal C stack, and this value is used to prevent stack overflows.
+Normally this is 1000 but it can be configured to another value, for
+instance 100, like this
```
tonycoz: This doesn't just limit the number of levels of BEGIN within eval, but the nesting levels of BEGINs in general.

This includes `use ModuleName;` since that's just `BEGIN { require ModuleName; ModuleName->import }`.

So the name of the variables and the documentation only talks about eval, but it has a more general effect.

op.c
```diff
@@ -10717,11 +10717,32 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
...
+                sv_setiv(max_nest_sv, PERL_MAX_NESTED_EVAL_BEGIN_BLOCKS_DEFAULT);
+            max_nest_iv = SvIV(max_nest_sv);
+            if (max_nest_iv < 0) {
+                max_nest_iv = PERL_MAX_NESTED_EVAL_BEGIN_BLOCKS_DEFAULT;
+                sv_setiv(max_nest_sv, max_nest_iv);
+            }
+
+            if (PL_eval_begin_nest_depth >= max_nest_iv) {
+                Perl_croak(aTHX_ "Too many nested BEGIN blocks, maximum of %d allowed",
+                             (int)max_nest_iv);
```
tonycoz: You could use `IVdf` here to avoid the cast.

INSTALL
demerphq: >This doesn't just limit the number of levels of BEGIN within eval, but the nesting levels of BEGINs in general.

No, that isn't correct. It looks like that but it does not work like that.

Consider this code:

```
BEGIN { $a++; BEGIN { $b++; BEGIN{ $c++ }  } } }
```
this code will complete parsing the innermost BEGIN first. So it increments the counter, executes the BEGIN, and then the  counter is decremented. It then finishes parsing the second BEGIN, and does the same thing. That is it is equivalent to this code:

```
BEGIN { $c++; }
BEGIN { $b++; }
BEGIN { $a++; }
```

so even though there are three nested BEGIN's textually, they are executed in the order of completion and thus the inner necessarily finishes before the outer.

Only when the code inside the BEGIN does an eval (or require, same thing) does this counter get incremented without being immediately decremented.

BEGIN {} blocks are executed *immediately* after parsing and then thrown away.  This means they don't nest without an eval inside them. Be aware that require is just a special type of eval.

> This includes use ModuleName; since that's just BEGIN { require ModuleName; ModuleName->import }.

Yes, it is true that affects such case, because 'require' is actually an eval internally that rethrows its exceptions.

Ill tweak the wording anyway, and maybe add a comment. The fact that BEGIN{} dont nest due to execution order unless they contain an eval/required is not obvious. I actually wrote tests to check this and then realized my expectations are broken.

op.c
demerphq: sure. willdo.

demerphq:
@tonycoz you good with the changes?

demerphq:
Repushed with the commits squashed.

tonycoz (approved):

Issue #20155, opened 2022-08-25
Locale patch breaks dist/threads/t/kill.t by triggering attribute.pm XS version mismatch errors.


demerphq:
**Description**
Bisect tells me that ever since a7ff7aca38 I have been seeing test failures from dist/threads/t/kill.t:

`
attributes object version 0.35 does not match $attributes::VERSION 0 at lib/XSLoader.pm line 112.
Compilation failed in require at lib/Thread/Queue.pm line 19.
BEGIN failed--compilation aborted at lib/Thread/Queue.pm line 19.
Compilation failed in require at dist/threads/t/kill.t line 31.
BEGIN failed--compilation aborted at dist/threads/t/kill.t line 36.
`

This is the patch details:

```
git log -1 --stat a7ff7aca3826600e5832ffe161fe075182f19a82
commit a7ff7aca3826600e5832ffe161fe075182f19a82
Author: Karl Williamson <khw@cpan.org>
Date:   Thu Mar 11 12:24:56 2021 -0700

    locale.c: Use setlocale() for init, not P2008
    
    We have found bugs in the POSIX 2008 libc implementations on various
    platforms.  This code, which does the initialization of locale handling
    has always been very conservative, expecting possible failures due to
    bugs in it or the libc implementations, and backing out if necessary to
    a crippled, but workable state, if something goes wrong.
    
    I think we should use the oldest, most stable locale implementation in
    these circumstances

 locale.c | 18 ++++++++++++++----
 1 file changed, 14 insertions(+), 4 deletions(-)
```
FWIW, I looked at the patch but beyond maybe an off by one error overwriting memory somewhere it is not clear to me why this patch would lead to this error. It is also not clear why it would fail for me and not in CI. I would be happy to provide any additional details required to debug.

**Steps to Reproduce**
```
$ git reset --hard a7ff7aca3826600e5832ffe161fe075182f19a82
HEAD is now at a7ff7aca38 locale.c: Use setlocale() for init, not P2008

$ make -j8 test_prep > /tmp/test_prep.log 2>&1; ./perl -Ilib dist/threads/t/kill.t
attributes object version 0.35 does not match $attributes::VERSION 0 at lib/XSLoader.pm line 112.
Compilation failed in require at lib/Thread/Queue.pm line 19.
BEGIN failed--compilation aborted at lib/Thread/Queue.pm line 19.
Compilation failed in require at dist/threads/t/kill.t line 31.
[...]

PR #20178, opened 2022-08-28
Don't end up in global locale upon thread destruction

khwilliamson:
The POSIX 2008 locale API introduces per-thread locales.  But the
previous global locale system is retained, probably for backward
compatibility.

Prior to this commit, there was a bug in which, when a thread
terminates, the master thread was switched into the global locale.  That
meant that that thread was no longer thread-safe with regards to
locales.

This bug stems from the fact that perl assumes that all you need to do
to switch between threads (or embedded interpreters) is to change out
aTHX.  Indeed much effort was expended in crafting perl to make this the
case.  But it breaks down in the case of some alien library that keeps
per-thread information.  That library needs to be informed of the
switch.  In this case it is libc keeping per-thread locale information.
We change the thread context, but the library still retains the old
thread's locale.

One cannot be using a given locale object and successfully free it.
Therefore the code switches to the global locale (which isn't
deletable) before freeing.  There was no apparent need to do more
switching, as the thread is in the process of dying.  What I was unaware
of is that it is the parent thread pretending to be the dying one for the
purposes of destruction.  So switching to the global locale affected the
parent, leaving it there.

The parent thread called the locale.c thread locale termination
function, and then called the perl.c perl_destruct() on the thread.  This
commit moves all the code for thread destruction from perl.c into the
locale.c code, and calls it.  Thus the thread initiation and termination
is moved into locale.c

The thread termination is also called from thread.c.  This cleans up a
dying thread.  The perl.c call is needed for thread0 and
non-multiplicity builds.  A check is done to prevent duplicate work.

This commit adds a new per-interpreter variable which maps aTHX to its
locale.  This is used to get the terminating thread's locale instead of
the master.  And the master locale is switched back to at the end.

This commit is incomplete.  Something similar needs to be done for
Windows where the libc knows the per-thread locale.

I'm unsure of if this is the full correct approach.  It only works for
thread termination.  Perhaps a better solution would be to change the
locale every time aTHX is changed.  PERL_SET_INTERP, PERL_SET_CONTEXT,
and PERL_SET_THX all seem to do the aTHX change, and I can't figure out
when you would prefer one over the other.  But maybe one of them should
then arrange also to change the locale when aTHX is changed.

Perhaps you can think of other libraries and functions that have a
similar problem that also would need something like this.

This commit causes https://github.com/Perl/perl5/issues/20155 to go away.  The triggering failure is merely
a symptom of the deeper problem.  A proper test will need to be done in
XS.

### DIFF
--- a/dist/threads/threads.xs
+++ b/dist/threads/threads.xs
@@ -241,11 +241,20 @@ S_ithread_clear(pTHX_ ithread *thread)
     S_block_most_signals(&origmask);
 #endif
 
+    int save_veto = PL_veto_switch_non_tTHX_context;
+
     interp = thread->interp;
     if (interp) {
         dTHXa(interp);
 
+        /* We will pretend to be a thread that we are not by switching tTHX,
+         * which doesn't work with things that don't rely on tTHX during
+         * tear-down, as they will tend to rely on a mapping from the tTHX
+         * structure, and that structure is being destroyed. */
+        PL_veto_switch_non_tTHX_context = true;
+
         PERL_SET_CONTEXT(interp);
+
         S_ithread_set(aTHX_ thread);
 
         SvREFCNT_dec(thread->params);
@@ -262,6 +271,8 @@ S_ithread_clear(pTHX_ ithread *thread)
     }
 
     PERL_SET_CONTEXT(aTHX);
+    PL_veto_switch_non_tTHX_context = save_veto;
+
 #ifdef THREAD_SIGNAL_BLOCKING
     S_set_sigmask(&origmask);
 #endif
--- a/dist/threads/lib/threads.pm
+++ b/dist/threads/lib/threads.pm
@@ -5,7 +5,7 @@ use 5.008;
 use strict;
 use warnings;
 
-our $VERSION = '2.29';      # remember to update version in POD!
+our $VERSION = '2.30';      # remember to update version in POD!
 my $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
--- a/embed.fnc
+++ b/embed.fnc
@@ -1663,8 +1663,8 @@ Xp	|void	|set_numeric_standard
 Cp	|bool	|_is_in_locale_category|const bool compiling|const int category
 Apd	|void	|switch_to_global_locale
 Apd	|bool	|sync_locale
-ApxT	|void	|thread_locale_init
-ApxT	|void	|thread_locale_term
+Apx	|void	|thread_locale_init
+Apx	|void	|thread_locale_term
 ApdO	|void	|require_pv	|NN const char* pv
 Apd	|void	|packlist	|NN SV *cat|NN const char *pat|NN const char *patend|NN SV **beglist|NN SV **endlist
 #if defined(PERL_USES_PL_PIDSTATUS) && defined(PERL_IN_UTIL_C)
--- a/embed.fnc
+++ b/embed.fnc
@@ -3442,12 +3442,15 @@ S	|const char*|my_langinfo_i|const int item			\
 				|NULLOK utf8ness_t * utf8ness
 #    endif
 #    ifdef DEBUGGING
-S	|const char *|get_LC_ALL_display
 SR	|char *	|my_setlocale_debug_string_i			\
 			    |const unsigned cat_index		\
 			    |NULLOK const char* locale		\
 			    |NULLOK const char* retval		\
 			    |const line_t line
+#    endif
+#    if  defined(DEBUGGING)						    \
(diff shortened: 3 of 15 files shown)


jkeenan:
> 
> This commit causes #20155 to go away. The triggering failure is merely a symptom of the deeper problem. A proper test will need to be done in XS.

This gets a quick build-time error on OpenBSD.
```
$ uname -mrs
OpenBSD 6.9 amd64

$ cc --version | head -1
OpenBSD clang version 10.0.1 

$ sh ./Configure -des -Dusedevel && make test_prep
...
cc -c -DPERL_CORE -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings locale.c
locale.c:6960:24: error: 'void' must be the first and only parameter if specified
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
                       ^
locale.c:6960:1: error: conflicting types for 'Perl_thread_locale_term'
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
^
./proto.h:4353:20: note: previous declaration is here
PERL_CALLCONV void      Perl_thread_locale_term(pTHX_ const bool caller_is_thread_termination);
                        ^
locale.c:6960:29: error: parameter name omitted
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
                            ^
3 errors generated.
*** Error 1 in target 'locale.o'
*** Error 1 in /home/jkeenan/gitwork/perl (makefile:246 'locale.o': @`sh  cflags "optimize='-O2'" locale.o`  locale.c)
```
(It did, however, pass all tests on FreeBSD-12.)

jkeenan:
> (It did, however, pass all tests on FreeBSD-12.)

That was a *threaded* build on FreeBSD.  But it looks like this is failing more generally on *unthreaded* builds.  Unthreaded build on FreeBSD:
```
cc -c -DPERL_CORE -DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings locale.c
locale.c:6960:24: error: 'void' must be the first and only parameter if specified
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
                       ^
locale.c:6960:1: error: conflicting types for 'Perl_thread_locale_term'
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
^
./proto.h:4353:20: note: previous declaration is here
PERL_CALLCONV void      Perl_thread_locale_term(pTHX_ const bool caller_is_thread_termination);
                        ^
locale.c:6960:29: error: parameter name omitted
Perl_thread_locale_term(pTHX, const bool caller_is_thread_termination)
                            ^
3 errors generated.
*** Error code 1

Stop.
```
Threaded build on OpenBSD in progress -- but it's gotten past the `locale.c` failure point in the unthreaded build.

Leont:
My opinion on *"should this be API or not"* greatly depends on *"do I need this to correctly support locales with [Thread::Csp](https://metacpan.org/dist/Thread-Csp)"*. I have the impression the answer is "yes"

dist/threads/threads.xs
```diff
@@ -667,8 +668,6 @@ S_ithread_run(void * arg)
     MUTEX_UNLOCK(&thread->mutex);
     MUTEX_UNLOCK(&MY_POOL.create_destruct_mutex);
 
-    thread_locale_term();
```
Leont: I think this is wrong, or if it's right it needs more explanation of why it's right.

`S_ithread_clear` will be called from other (OS) threads, whereas S_ithread_run will be run from the interpreter's own thread. This is an essential distinction.

khwilliamson: we agree its wrong

Leont:
Currently the code will destroy the current locale on the thread that calls `perl_destroy`. Because of threads.pm architecture this will often not be the thread matching the interpreter (ithread objects are refcounted, and perl_destroy is called in the last thread referencing it).

I can think of two possible solution:
1. perl_destroy never destroys thread-local data like the locale.
  This does beg the question how we will destroy it.
3. threads.pm will always call perl_destroy from the matching thread.
  This will require a rearchitecture of thread destruction in threads.pm

To be honest, doing both may be the best idea.

khwilliamson:
#20361 supersedes this

Issue #16418, opened 2018-02-13
Bleadperl breaks OALDERS/LWP-Protocol-https-6.07.tar.gz on Windows


p5pRT:
Migrated from [rt.perl.org#132863](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=132863) (status was 'resolved')

Searchable as RT132863$

p5pRT:
### From @chorny

#### Created by @chorny



https\_proxy\.t hangs on 5\.27\.8\, did not hang on 5\.27\.6\, 5\.27\.1\, 5\.26\.0\.
All tests of https\_proxy\.t pass\, it just does not exit\.

I'm using Windows XP SP3 Russian 32\-bit\.

Also reported here&#8203;: https://github.com/libwww-perl/LWP-Protocol-https/issues/52

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=high

Site configuration information for perl 5.27.8:

Configured by strawberry-perl at Tue Feb 13 01:56:34 2018.

Summary of my perl5 (revision 5 version 27 subversion 8) configuration:

  Platform:
    osname=MSWin32
    osvers=5.1.2600
    archname=MSWin32-x86-multi-thread-64int
    uname='Win32 strawberry-perl 5.27.8.1-beta1 #1 Tue Feb 13 01:53:53
2018 i386'
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='gcc'
    ccflags =' -s -O2 -DWIN32 -D__USE_MINGW_ANSI_STDIO
-DPERL_TEXTMODE_SCRIPTS -DPERL_IMPLICIT_CONTEXT -DPERL_IMPLICIT_SYS
-DUSE_PERLIO -fwrapv -fno-strict-aliasing -mms-bitfields'
    optimize='-s -O2'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='7.1.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=4
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=12
    longdblkind=3
    ivtype='long long'
    ivsize=8
[...]

Issue #1537, opened 2000-03-28
[BUG] can't exit 0 from CHECK{}


p5pRT:
Migrated from [rt.perl.org#2754](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=2754) (status was 'open')

Searchable as RT2754$

p5pRT:
### From tchrist@chthon.perl.com

From a CHECK\{\}\, you cannot exit\(0\)\.  You may exit \!0\, but not 0\.
If you put this in /tmp/a and run it&#8203;:

&nbsp;    \#BEGIN \{ warn "testing exit from BEGIN"; exit \}
&nbsp;    \#BEGIN \{ warn "testing exit N from BEGIN"; exit 1 \}

&nbsp;    \#INIT \{ warn "testing exit from INIT"; exit \}
&nbsp;    \#INIT \{ warn "testing exit N from INIT"; exit 2 \}

&nbsp;    CHECK \{ warn "testing exit from CHECK"; exit \}
&nbsp;    \#CHECK \{ warn "testing exit N from CHECK"; exit 3 \}

&nbsp;    \#END \{ warn "testing exit from END"; exit \}
&nbsp;    \#END \{ warn "testing exit N from END"; exit 4 \}

&nbsp;    print "i am now the main program\\n";
&nbsp;    warn "testing exit 5 from main";
&nbsp;    exit 5;

&nbsp;    die "XXX";

You will get&#8203;:

&nbsp;    % perl /tmp/a
&nbsp;    testing exit from CHECK at /tmp/a line 7\.
&nbsp;    i am now the main program
&nbsp;    testing exit 5 from main at /tmp/a line 14\.
&nbsp;    Exit 5

If you switch the comment on the two CHECKs\, you get

&nbsp;    % perl /tmp/a
&nbsp;    testing exit N from CHECK at /tmp/a line 8\.
&nbsp;    Exit 3

\-\-tom

p5pRT:
### From @floatingatoll

\[tchrist@&#8203;chthon\.perl\.com \- Tue Mar 28 03&#8203;:39&#8203;:19 2000\]&#8203;:

> From a CHECK\{\}\, you cannot exit\(0\)\.  You may exit \!0\, but not 0\.
> If you put this in /tmp/a and run it&#8203;:
> 
>     \#BEGIN \{ warn "testing exit from BEGIN"; exit \}
>     \#BEGIN \{ warn "testing exit N from BEGIN"; exit 1 \}
[...]

PR #20181, opened 2022-08-29
Resolve #2754 and #16300 - introduce a workaround for Module::Install::DSL and fixup return code for perl_parser()

### LINKED ISSUE #2754: [PATCH]Not OK: perl v5.7.0 +DEVEL7414 on dos-djgpp djgpp
Migrated from [rt.perl.org#4504](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=4504) (status was 'resolved')

Searchable as RT4504$

demerphq:
In 0301e899536a22752f40481d8a1d141b7a7dda82 Zefram fixed a long standing bug #2754, and properly defined the exit code from perl_parse(). This however broke Module::Install::DSL as can be seen in issue #16300. The patch was reverted in 857320cbf85e762add18885ae8a197b5e0c21b69.  Module::Install is not easily upgradable due to its design, and there are still a lot of modules out there with the buggy use of INIT. 

However, it turns out that the bugfix from #2754 allows us to fix other issues which we want to address. 

So this PR reverts Zeframs revert and creates a simple workaround for the Module::Install::DSL issue. INIT blocks in the Module::Install::DSL namespace will magically be treated as BEGIN blocks. This means that they will be executed inside of the eval that Module::Install::DSL creates, and executed as expected, at the same time a warning is produced to show the workaround has been triggered.

The check for this is quite efficient so it should not be noticeable outside of the special case it is meant to work around.

This will enable us to fix Issue: #20161 and potentially also apply PR #20168

### DIFF
--- a/op.c
+++ b/op.c
@@ -10711,14 +10711,17 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
 {
     const char *const colon = strrchr(fullname,':');
     const char *const name = colon ? colon + 1 : fullname;
+    int is_module_install_hack = 0;
 
     PERL_ARGS_ASSERT_PROCESS_SPECIAL_BLOCKS;
 
     if (*name == 'B') {
-        if (strEQ(name, "BEGIN")) {
+        module_install_hack:
+        if (strEQ(name, "BEGIN") || is_module_install_hack) {
             const I32 oldscope = PL_scopestack_ix;
             dSP;
             (void)CvGV(cv);
+            is_module_install_hack = 0;
             if (floor) LEAVE_SCOPE(floor);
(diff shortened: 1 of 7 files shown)


demerphq:
Turns this touches on other reverted logic from what i can tell. The last two commits are related to Win32 issues. See https://github.com/Perl/perl5/issues/16418

tonycoz (approved):

demerphq:
I will merge this as is, if and when the MI workaround is no longer needed we can remove it.

tonycoz:
This links to #2754 but that seems unrelated, I suspect it means rt2754 aka #1537

PR #20182, opened 2022-08-29
Resolve issue #20161: Evals that die under the debugger may or may not leak eval data

### LINKED ISSUE #20161: Evals that die under the debugger may or may not leak eval data depending on how they die.

demerphq:
We were leaking debug data, and not testing if we did. Assuming #20181 is applied, then we can apply this and fix #20161 which will enable use to merge #20168 if we want. This is rebased on top of #20181. Once that is applied only the most recent patch in this sequence will be relevant to this PR. 

This patch changes the code to use try_yyparse() when compiling evals, while preserving the existing behavior for require.

### DIFF
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3752,7 +3752,55 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
 
     if (PL_unitcheckav) {
         OP *es = PL_eval_start;
-        call_list(PL_scopestack_ix, PL_unitcheckav);
+        if (in_require) {
+            call_list(PL_scopestack_ix, PL_unitcheckav);
+        } else {
+            /* TODO: are we sure we shouldn't do JMPENV_PUSH in
+             * when `in_require` is true? */
+            int ret=0;
+            dJMPENV;
+            JMPENV_PUSH(ret);
+            switch (ret) {
+            case 0:
+                /*
+                 * Doesn't seem like PUSHMARK(SP)/ENTER
+                 * is needed here. */
+
+                call_list(PL_scopestack_ix, PL_unitcheckav);
+                /* Nor LEAVE here. */
+                break;
+            case 3: {
+                /* call_list failed and threw an error */
+
+                /* Restore PL_OP */
+                PL_op = saveop;
+
+                SV *errsv = ERRSV;
+                if (!*(SvPV_nolen_const(errsv))) {
+                    /* This happens when using:
+                     * eval qq# UNITCHECK { die "\x00"; } #;
+                     */
+                    sv_setpvs(errsv, "Unit check error");
+                }
+
+                /* We're returning so POP our JMPENV */
+                /* NOTE: in `S_try_yyparse` the default for ret=3 is to
+                 *       break which falls back to the `JMPENV_POP`
+                 *       after the switch. In this code we're returning
+                 *       early so we must POP it outrself. */
+                JMPENV_POP;
+
+                if (gimme != G_LIST) PUSHs(&PL_sv_undef);
+                PUTBACK;
+                return FALSE;
+              }
+            default:
+                JMPENV_POP;
+                JMPENV_JUMP(ret);
+                NOT_REACHED; /* NOTREACHED */
+            }
+            JMPENV_POP;
+        }
         PL_eval_start = es;
     }
 
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3686,12 +3686,13 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
 
     CALL_BLOCK_HOOKS(bhk_eval, saveop);
 
-    /* note that yyparse() may raise an exception, e.g. C<BEGIN{die}>,
-     * so honour CATCH_GET and trap it here if necessary */
-
-
+    /* we should never be CATCH_GET true here, as our immediate callers should
+     * always handle that case. */
+    assert(!CATCH_GET);
     /* compile the code */
-    yystatus = (!in_require && CATCH_GET) ? S_try_yyparse(aTHX_ GRAMPROG) : yyparse(GRAMPROG);
+    yystatus = (!in_require)
+               ? S_try_yyparse(aTHX_ GRAMPROG)
+               : yyparse(GRAMPROG);
 
     if (yystatus || PL_parser->error_count || !PL_eval_root) {
         PERL_CONTEXT *cx;
@@ -4577,6 +4578,8 @@ PP(pp_entereval)
      */
     if (CATCH_GET)
         return docatch(Perl_pp_entereval);
+
+    assert(!CATCH_GET);
 
     gimme = GIMME_V;
     was = PL_breakable_sub_gen;
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3750,7 +3750,7 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
         }
     }
 
-    if (PL_unitcheckav) {
+    if (PL_unitcheckav && av_count(PL_unitcheckav)>0) {
         OP *es = PL_eval_start;
         if (in_require) {
             call_list(PL_scopestack_ix, PL_unitcheckav);
--- a/cop.h
+++ b/cop.h
@@ -116,8 +116,8 @@ typedef struct jmpenv JMPENV;
             int i = 0;                                                  \
             JMPENV *p = PL_top_env;                                     \
             while (p) { i++; p = p->je_prev; }				\
-            Perl_deb(aTHX_ "JMPENV_PUSH pre level=%d at %s:%d\n",       \
-                         i,  __FILE__, __LINE__);                       \
+            Perl_deb(aTHX_ "JMPENV_PUSH pre level=%d in %s at %s:%d\n", \
+                         i,  SAFE_FUNCTION__, __FILE__, __LINE__);      \
         });                                                             \
         cur_env.je_prev = PL_top_env;					\
         JE_OLD_STACK_HWM_save(cur_env);                                 \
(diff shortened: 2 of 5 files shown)


bram-perl:
I don't know if this is a good idea..
Reading https://perldoc.perl.org/perlinterp#Exception-handing (as was linked by @iabyn) I got the impression that `CATCH_GET` was done the way it is for optimizations..

Now the code will always use `try_yyparse()` (for eval) which I'm **guessing** is going to add some overhead..
What I don't know is how much that overhead is..

What I'm also wondering about: is this the only way to fix it? Skimming the code it appears that before commit f9bddea7d2a0d824366014c8ee6ba57e7dedd8c3 it first called `SAVEDELETE` and then called `doeval` while the current code first does `doeval` (which can die) and only then calls `SAVEDELETE`.. Would switching the order (i.e. calling `SAVEDELETE` before `doeval`) fix it too?

demerphq:
@bram-perl I dont think you are right TBH. But I guess @iabyn will have to speak up. He was the one who suggested the CATCH_GET was superfluous after all, and removing it does fix the leaking debug data. I dont think it changes the performance signature at all, that concern relates to execution not compilation.

Note that the added overhead is that of adding a call to JMPENV_PUSH()/JMPENV_POP which boils down to some bookkeeping and a call to Perl_procsetjmp(). In the grand scheme of doing an eval that is a tiny thing. Also because this is compilation we push an env and then pop it, they wont cascade except with regard to BEGIN, and that is already a problem that is being tackled in a different PR.

Also it doesnt make sense to me, @iabyn added try_yyparse, which is now no longer used. It makes a lot more sense to me that it should be used than it should be there totally unused.

bram-perl:
> Also it doesnt make sense to me, @iabyn added try_yyparse, which is now no longer used. It makes a lot more sense to me that it should be used than it should be there totally unused.

Just for reference: reason for that is commit d7e3f70f30811328d2f6ae57e5892deccf64d0b2. <details>

That commit added an earlier 'catchable runloop' which resulted in `S_doeval_compile` to always be called with a false value for `CATCH_GET`.

I *suspect* (but did not verify) that an alternative fix could've been to add a 'catchable runloop' further down in `S_doeval_compile` namely in the block:
```
    if (PL_unitcheckav) {
        OP *es = PL_eval_start;
        call_list(PL_scopestack_ix, PL_unitcheckav);
        PL_eval_start = es;
    }
```

</details>

demerphq:
On Mon, 29 Aug 2022 at 21:31, Bram ***@***.***> wrote:

> Also it doesnt make sense to me, @iabyn <https://github.com/iabyn> added
> try_yyparse, which is now no longer used. It makes a lot more sense to me
> that it should be used than it should be there totally unused.
>
> Just for reference: reason for that is commit d7e3f70
> <https://github.com/Perl/perl5/commit/d7e3f70f30811328d2f6ae57e5892deccf64d0b2>.
>
>
> That commit added an earlier 'catchable runloop' which resulted in
> S_doeval_compile to always be called with a false value for CATCH_GET.
>
> I *suspect* (but did not verify) that an alternative fix could've been to
> add a 'catchable runloop' further down in S_doeval_compile namely in the
> block:
>
>     if (PL_unitcheckav) {
>         OP *es = PL_eval_start;
>         call_list(PL_scopestack_ix, PL_unitcheckav);
>         PL_eval_start = es;
>     }
>
>
Alternate fix to d7e3f70 you mean?

BTW, the other thing that makes me thing this is the right thing to do is
the chain of bugs this revealed and fixes as a whole. Every step of the way
fixing this we have fallen into old reverted bugfixes. IMO that suggests
that what we are doing is essentially using a bunch of bugs to cancel each
other out, which further suggests to me we are on the right track. Also,
while admittedly this last point verges on the superstitious, it felt like
a good sign that *all* the windows and cygwin tests passed after these
patches. I have seen so many CI heisenbugs that I found that to be quite
encouraging, I almost wonder if this stuff has been at the root of other
issues and we never realized.

cheers,
Yves

-- 

bram-perl:
For `UNITCHECK` see https://github.com/Perl/perl5/issues/20161#issuecomment-1230930457

The test only works correct by accident (imo).

demerphq:
On Tue, 30 Aug 2022 at 00:22, Bram ***@***.***> wrote:
>
> For UNITCHECK see #20161 (comment)
>
> The test only works correct by accident (imo).

Ok. Ill take a look at it later.

Yves

-- 

khwilliamson:
A more general comment.  It was suggested on some of my commit messages that it should dereference the issue's name so that the reader doesn't have to each time it is read.  That effectively could be to copy/paste.

pp_ctl.c
```diff
@@ -3752,7 +3752,55 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
 
     if (PL_unitcheckav) {
         OP *es = PL_eval_start;
-        call_list(PL_scopestack_ix, PL_unitcheckav);
+        if (saveop->op_type == OP_REQUIRE) {
```
bram-perl: Could/should be: `if (in_require) {`

(for consistency with the ` yystatus = ...`)

demerphq: fix pushed.

bram-perl (approved):

PR #20183, opened 2022-08-29
perl.h - add DEBUG_force() macro, to force debug for DEBUG_l() style …

demerphq:
…macros

If you temporarily want to force on a DEBUG_l({...}) style macro you
have to change a fair bit of code, which is error prone. DEBUG_force()
lets you replace any DEBUG_x() macro with DEBUG_force() which will force
the debug block to execute (under DEBUGGING).


tonycoz:
How is this different to PERL_DEB()?

demerphq:
I didn't know about PERL_DEB() when i wrote this.  Closing.

PR #20184, opened 2022-08-29
perlinterp.pod - minor enhancements of the docs about JMPENV_ macros

demerphq:
Added a bit more on the levels and return codes. 1 is actually used
in the outmost call, or at least there is code to support it. Also
added "level" info, since that is used in -Dl.

### DIFF
--- a/pod/perlinterp.pod
+++ b/pod/perlinterp.pod
@@ -194,11 +194,11 @@ achieved via a C<JMPENV_JUMP()>. In particular:
 
 =over
 
-=item * perl-level exit() and internals my_exit()
+=item * level 2: perl-level exit() and internals my_exit()
 
 These unwind all stacks, then perform a JMPENV_JUMP(2).
 
-=item * perl-level die() and internals croak()
+=item * level 3: perl-level die() and internals croak()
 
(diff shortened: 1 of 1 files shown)


iabyn:
On Mon, Aug 29, 2022 at 11:57:37AM -0700, Yves Orton wrote:
> Added a bit more on the levels and return codes. 1 is actually used
> in the outmost call, or at least there is code to support it. Also
> added &quot;level&quot; info, since that is used in -Dl.

A couple of comments:

This seems to be mixing up levels and jump types. 2 and 3 are types - 
the argument to JMPENV_JUMP() to indicate whether its an exit or restart.
On the other hand, level (as displayed by -Dl) is how many nested
JMPENV_PUSH()'s there have been. So the docs for JMPENV_JUMP shouldn't be
referring to 0,1,2,3 as "levels".

JMPENV_JUMP(1) is completely unused as far as I'm aware. The code to catch
it in perl_run() is (I strongly suspect) completely vestigial, and has
been that way since 5.000. If it serves any purpose, I don't know what it
is (or was). It should probably be removed.


-- 
Standards (n). Battle insignia or tribal totems.

PR #20185, opened 2022-08-29
cop.h - show a bit more debug out in JMPENV macros, and minor fixes

demerphq:
Eg, use (v) consistently. Line up backslashes. Show the return from
JMPENV_PUSH().

### DIFF
--- a/cop.h
+++ b/cop.h
@@ -110,26 +110,30 @@ typedef struct jmpenv JMPENV;
 
 #define dJMPENV		JMPENV cur_env
 
-#define JMPENV_PUSH(v) \
+#define JMPENV_PUSH(v)                                                  \
     STMT_START {							\
-        DEBUG_l({							\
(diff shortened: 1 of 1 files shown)


cop.h
```diff
@@ -109,27 +109,36 @@ typedef struct jmpenv JMPENV;
...
+#else
+#define iDECL
+#endif
 
-#define JMPENV_PUSH(v) \
+#define JMPENV_PUSH(v)                                                  \
     STMT_START {							\
-        DEBUG_l({							\
-            int i = 0; JMPENV *p = PL_top_env;				\
+        iDECL;                                                          \
```
tonycoz: This could just be `PERL_DEB(int i = 0);`

demerphq: Oh, nice, I didnt know that one. Willdo.

tonycoz (approved):

PR #20186, opened 2022-08-29
toke.c - dont just return the function, assign it to an intermediary

demerphq:
So we can debug it before we return. an optimization compiler should
make them the same thing anyway, but under -Og it is helpful to be able
to be able to see the return before we return it.

### DIFF
--- a/toke.c
+++ b/toke.c
@@ -12531,6 +12531,7 @@ Perl_abort_execution(pTHX_ const char * const msg, const char * const name)
         Perl_croak(aTHX_
                 "%sExecution of %s aborted due to compilation errors.\n", msg, name);
     }
+
     NOT_REACHED; /* NOTREACHED */
 }
 
@@ -12547,14 +12548,16 @@ int
 Perl_yyerror(pTHX_ const char *const s)
 {
     PERL_ARGS_ASSERT_YYERROR;
-    return yyerror_pvn(s, strlen(s), 0);
+    int r = yyerror_pvn(s, strlen(s), 0);
+    return r;
 }
 
 int
 Perl_yyerror_pv(pTHX_ const char *const s, U32 flags)
 {
     PERL_ARGS_ASSERT_YYERROR_PV;
-    return yyerror_pvn(s, strlen(s), flags);
+    int r = yyerror_pvn(s, strlen(s), flags);
+    return r;
 }
 
 int


tonycoz:
This seems like the type of change I might make temporarily rather than something to commit.

Assuming `finish` didn't do what I wanted.

demerphq:
On Wed, 31 Aug 2022 at 07:01, Tony Cook ***@***.***> wrote:

> This seems like the type of change I might make temporarily rather than
> something to commit.
>
Well, personally i dont see why we wouldn't make it permanent. If it got
used to debug this once then why not again in the future? My experience is
the past tends to repeat itself....  What would the harm be?

Yves

>
>

-- 

ericherman:
>an optimization compiler should make them the same thing anyway

Indeed. And trivial to confirm with [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:4,endLineNumber:11,positionColumn:4,positionLineNumber:11,selectionStartColumn:4,selectionStartLineNumber:11,startColumn:4,startLineNumber:11),source:'%23include+%3Cstddef.h%3E%0A%23include+%3Cstring.h%3E%0Aint+yyerror_pvn(const+char+*const,+size_t,+int)%3B%0A%0Aint%0Afoo_orig(const+char+*const+s)%0A%7B%0A++++return+yyerror_pvn(s,+strlen(s),+0)%3B%0A%7D%0A%0Aint%0Afoo_next(const+char+*const+s)%0A%7B%0A++++int+r+%3D+yyerror_pvn(s,+strlen(s),+0)%3B%0A++++return+r%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g121,deviceViewOpen:'1',filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O2',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+12.1+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) if you wish.

ericherman (approved):

PR #20192, opened 2022-08-30
[doc] Update File::Copy synopsis

rwp0:

### DIFF
--- a/lib/File/Copy.pm
+++ b/lib/File/Copy.pm
@@ -7,7 +7,7 @@
 
 package File::Copy;
 
-use 5.035007;
+use 5.035008;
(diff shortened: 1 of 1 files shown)


lib/File/Copy.pm
```diff
@@ -7,7 +7,7 @@
 
 package File::Copy;
 
-use 5.035007;
+use 5.035008;
```
bram-perl, line 10: You have bumped the wrong version number.
(Should be bumping `$VERSION`)

rwp0: haha, I'm sorry, thanks :-)

PR #20203, opened 2022-08-31
[doc] Update synopsis for User::grent and User::pwent

rwp0:


karenetheridge:
I would suggest squashing these commits together.

rwp0:
> I would suggest squashing these commits together.

I'm sorry, some day I'll learn to do that as well with IntelliJ IDEA, and until then I'd delete this.

PR #20205, opened 2022-08-31
Update synopses on User:: and Sys:: modules

rwp0:


rwp0:
Should I submit PR's individually?

ie. a PR per module:

- `User::grent`
- `User::pwent`
- `Sys::Syslog`

I checked that all of these modules above point to Perl GitHub repository (ie. integrated with Perl distribution) on MetaCPAN, so submitted the PR here.

Grinnz:
This seems fine, these are all core maintained modules and the changes are related.

PR #20209, opened 2022-09-01
Draft: Whitespace Cleanup Manually Maintained Header Files

demerphq:
This patch sequences cleans up various minor nits in our manually managed header files (eg, *.h in the root of the repo)  , and then introduces a whitespace fixup patch to put all our header files into a standardized format, with defines using as consistent formatting as possible, multiline macros definitions having consistently formatted line continuation markers and other things like that. Similarly structs and unions are consistently formatted with the `*` character from pointer declarations being associated with the name not the type like we would normally for variables, and with the member declarations lined up properly and with comments at consistent positions in the definition. In other words the last patch in this sequence should ONLY make whitespace changes.

I wanted to have this patch so we could start a conversation about whether we should use some kind of tool to enforce a consistent format. (I think we should.) I find the inconsistencies in the formatting makes the header files harder to read, and i think programs are better at it than we are, so at least for header files I thought i would see what people think.


bram-perl:
I think the first commits can be merged as-is,

For the whitespace cleanup commit, skimming the files:

- op.h doesn't look good:
  - lines 11-40 are misaligned
  - lines 111-167: comments that continue on the next line are no longer aligned
- perl.h:
  - lines ~~5223-5555~~ 5523-5555
  - 5556-5599
  - 5605-5636
  - 5642-5681
  - 5842-5849
- utfebcdic.h:
  - lines 30-32
 
Other things **I** am not fond of:

(Taking a random example)
regexp.h
```
        struct {
            /* this first element must match u.yes */
            struct regmatch_state   *prev_yes_state;
            U32                     lastparen;
            U32                     lastcloseparen;
            CHECKPOINT              cp;

        }                            branchlike;
```
The name of the struct is too far to the right (again IMO)

(Update: fix typo in line numbers of perl.h)

khwilliamson:
On 9/1/22 10:40, Bram wrote:
> I think the first commits can be merged as-is,
> 
> For the whitespace cleanup commit, skimming the files:
> 
>   * op.h doesn't look good:
>       o lines 11-40 are misaligned
>       o lines 111-167: comments that continue on the next line are no
>         longer aligned
>   * perl.h:
>       o lines 5223-5555
>       o 5556-5599
>       o 5605-5636
>       o 5642-5681
>       o 5842-5849
>   * utfebcdic.h:
>       o lines 30-32
> 
> Other things *I* am not fond of:
> 
> (Taking a random example)
> regexp.h
> 
> |struct { /* this first element must match u.yes */ struct 
> regmatch_state *prev_yes_state; U32 lastparen; U32 lastcloseparen; 
> CHECKPOINT cp; } branchlike; |
> 
> The name of the struct is too far to the right (again IMO)

agreed

demerphq:
On Thu, 1 Sept 2022 at 20:42, Karl Williamson ***@***.***>
wrote:

> On 9/1/22 10:40, Bram wrote:
> > I think the first commits can be merged as-is,
> >
> > For the whitespace cleanup commit, skimming the files:
> >
> > * op.h doesn't look good:
> > o lines 11-40 are misaligned
> > o lines 111-167: comments that continue on the next line are no
> > longer aligned
> > * perl.h:
> > o lines 5223-5555
> > o 5556-5599
> > o 5605-5636
> > o 5642-5681
> > o 5842-5849
> > * utfebcdic.h:
> > o lines 30-32


I think I understand what caused these, I may have fixed some, but i
think not all, or maybe not any, will fix later today.


> > The name of the struct is too far to the right (again IMO)
>
> agreed
>

That should be fixed in the latest push. Which is not perfect in a number a
ways that I am aware of already. ;-)  Particularly "fixing" the indent on
multiline STMT_START macros.

Feel free to review but if it regards the latter point I'm already aware
and on it.

Yves

-- 

demerphq:
I have pushed a significant revamp of the white space cleanups. This also includes sundry changes to infra that is whitespace sensitive or sensitive to line continuation markers in our header files. 

I dont think this is "good enough" to merge yet, but i do think its "good enough" to hear folks comments. One issue for me is lining up things can lead to comments straying past the 80 column boundary. I havent decided the right thing to do there, and some of the choices the code makes may not be ideal.

regen/regcomp.pl
```diff
@@ -586,36 +586,39 @@ sub print_reg_extflags_name {
...
+                next unless $line =~ /<</;    # Bit defines use left shift
                 if ( $val & $newval ) {
                     my @names= ( $define, $reverse{$newval} );
                     s/PMf_// for @names;
                     if ( $names[0] ne $names[1] ) {
-                        die sprintf
-                            "ERROR: both $define and $reverse{$newval} use 0x%08X (%s:%s)",
+                        die sprintf "ERROR: both %s and %s use 0x%08X (%s:%s)",
+                            $define, $reverse{$newval},
                             $newval, $orig, $_;
```
bram-perl: *commit regen/regcomp.pl - make robust to line continuations in header files*

I think the `$_` be changed to `$line`?

ext/XS-APItest/t/locale.t
```diff
@@ -118,7 +118,7 @@ SKIP: {
         chomp;
         next unless / - \d+ $ /x;
         s/ ^ \# \s* define \s*//x;
-        m/ (.*) \  (.*) /x;
+        m/ (\S+) \s+ (.*) /x;
```
bram-perl: *commit XS-APItest/t/locale.t - deal with indented values properly*

It would be nice to amend the commit message with an example of what failed to parse
and/or what parsed incorrectly..

Also: I think it fixes more then just indented values?
I'm assuming (not verified) that before this commit something like: `#define FOO "A B C"` would get parsed incorrectly

gv.h
```diff
@@ -225,9 +231,13 @@ Return the CV from the GV.
...
                                    already; used also by gv_init_* */
-#define GV_ADDWARN      0x04    /* add, but warn if symbol wasn't already there */
+#define GV_ADDWARN  \
+    0x04    /* add, but warn if symbol wasn't already there */
+
                 /*      0x08       UNUSED */
-#define GV_NOINIT       0x10    /* add, but don't init symbol, if type != PVGV */
+#define GV_NOINIT   \
+    0x10    /* add, but don't init symbol, if type != PVGV */
+
```
bram-perl: *commit (2) break long define*

not a fan of these, probably better to make the comment multi-line

handy.h
```diff
@@ -1808,8 +1814,12 @@ END_EXTERN_C
...
      *  Latin1, which utilities and bootstrapping don't deal with */
 #   define _IS_NON_FINAL_FOLD_ONLY_FOR_USE_BY_REGCOMP_DOT_C(c) 0
-#   define _HAS_NONLATIN1_SIMPLE_FOLD_CLOSURE_ONLY_FOR_USE_BY_REGCOMP_DOT_C_AND_REGEXEC_DOT_C(c) 0
-#   define _HAS_NONLATIN1_FOLD_CLOSURE_ONLY_FOR_USE_BY_REGCOMP_DOT_C_AND_REGEXEC_DOT_C(c) 0
+#   define _HAS_NONLATIN1_SIMPLE_FOLD_CLOSURE_ONLY_FOR_USE_BY_REGCOMP_DOT_C_AND_REGEXEC_DOT_C(c)    \
+        0
+
+#   define _HAS_NONLATIN1_FOLD_CLOSURE_ONLY_FOR_USE_BY_REGCOMP_DOT_C_AND_REGEXEC_DOT_C(c)   \
+        0
+
```
bram-perl: *commit (2) break long define*

Not liking these..
**But** I think the good news is that these can just be dropped, they appear to be unused..
There was a variant of these that was used but khw renamed these in commit c62fdeb784c7643c90d2ea8c2ec0f03a548da338 so these just appear to be a left-over

op.h
```diff
@@ -1149,7 +1153,9 @@ C<sib> is non-null. For a higher-level interface, see C<L</op_sibling_splice>>.
 #  define TR_UNMAPPED           (UV)-1
 #  define TR_DELETE             (UV)-2
 #  define TR_R_EMPTY            (UV)-3  /* rhs (replacement) is empty */
-#  define TR_OOB                (UV)-4  /* Something that isn't one of the others */
+#  define TR_OOB    \
+       (UV)-4  /* Something that isn't one of the others */
+
```
bram-perl: *commit  (2) break long define*

Split the comment instead

perl.h
```diff
@@ -1623,7 +1627,9 @@ Use L</UV> to declare variables of the maximum usable size on this platform.
 #define PERL_MULTICONCAT_IX_PLAIN_LEN 2 /* non-utf8 constant string length */
 #define PERL_MULTICONCAT_IX_UTF8_PV   3 /* utf8 constant string */
 #define PERL_MULTICONCAT_IX_UTF8_LEN  4 /* utf8 constant string length */
-#define PERL_MULTICONCAT_IX_LENGTHS   5 /* first of nargs+1 const segment lens */
+#define PERL_MULTICONCAT_IX_LENGTHS \
+    5 /* first of nargs+1 const segment lens */
+
```
bram-perl: *comment (2) break long define*

Split the comment

perl.h
```diff
@@ -6868,8 +6931,12 @@ typedef struct am_table_short AMTS;
...
 #define PERLDBf_NAMEEVAL        0x100   /* Informative names for evals */
 #define PERLDBf_NAMEANON        0x200   /* Informative names for anon subs */
-#define PERLDBf_SAVESRC         0x400   /* Save source lines into @{"_<$filename"} */
-#define PERLDBf_SAVESRC_NOSUBS  0x800   /* Including evals that generate no subroutines */
+#define PERLDBf_SAVESRC \
+    0x400   /* Save source lines into @{"_<$filename"} */
+
+#define PERLDBf_SAVESRC_NOSUBS  \
+    0x800   /* Including evals that generate no subroutines */
+
```
bram-perl: *commit (2) break long define*

Also: split the comment

XSUB.h
```diff
@@ -186,7 +186,8 @@ is a lexical C<$_> in scope.
     I32 ax = XS_SETXSUBFN_POPMARK;  \
     SV **mark = PL_stack_base + ax - 1; dSP; dITEMS
 
-#define dXSTARG SV * const targ = ((PL_op->op_private & OPpENTERSUB_HASTARG)    \
+#define dXSTARG                                                     \
+    SV * const targ = ((PL_op->op_private & OPpENTERSUB_HASTARG)    \
     ? PAD_SV(PL_op->op_targ) : sv_newmortal())
```
bram-perl: *commit (3) break multiline define*

Not entirely sure of these..
I might be tempted to indent the ternary in this case

i.e. something like:
```
#define dXSTARG                                                     \
    SV * const targ = ((PL_op->op_private & OPpENTERSUB_HASTARG)    \
        ? PAD_SV(PL_op->op_targ) : sv_newmortal())
```

(but I suppose it might be out of scope for this comment)

cop.h
```diff
@@ -499,22 +499,27 @@ string C<p>, creating the package if necessary.
...
+       (CopFILE(c)      \
        ? GvSV(gv_fetchfile(CopFILE(c))) : NULL)
-#  define CopFILEAV(c)          (CopFILE(c) \
+#  define CopFILEAV(c)  \
+       (CopFILE(c)      \
        ? GvAV(gv_fetchfile(CopFILE(c))) : NULL)
-#  define CopFILEAVx(c)         (assert_(CopFILE(c))    \
+#  define CopFILEAVx(c)     \
+       (assert_(CopFILE(c)) \
        GvAV(gv_fetchfile(CopFILE(c))))
```
bram-perl: *commit (3) break multiline define*

I may also have indented in this case but maybe with just one space because the first `(` is not yet closed.
So something like:
```
#  define CopFILEAVx(c)     \
       (assert_(CopFILE(c)) \
        GvAV(gv_fetchfile(CopFILE(c))))
```
or
```
#  define CopFILEAVx(c)     \
       (assert_(CopFILE(c)) \
           GvAV(gv_fetchfile(CopFILE(c))))
```

cv.h
```diff
@@ -283,15 +285,17 @@ CvNAME_HEK(CV *sv)
 /* helper for the common pattern:
    CvNAMED(sv) ? CvNAME_HEK((CV *)sv) : GvNAME_HEK(CvGV(sv))
 */
-#define CvGvNAME_HEK(sv) (                                      \
+#define CvGvNAME_HEK(sv)                                        \
+    (                                                           \
     CvNAMED((CV*)sv) ?                                          \
         ((XPVCV*)MUTABLE_PTR(SvANY((SV*)sv)))->xcv_gv_u.xcv_hek \
         : GvNAME_HEK(CvGV( (SV*) sv))                           \
     )
```
bram-perl: *commit (3) break multiline define*

Here too I would indent

EXTERN.h
```diff
@@ -22,32 +22,32 @@
 #  if defined(WIN32) && !defined(PERL_STATIC_SYMS)
     /* miniperl should not export anything */
 #    if defined(PERL_IS_MINIPERL)
-#      define EXT extern
+#      define EXT          extern
 #      define dEXT
-#      define EXTCONST extern const
-#      define dEXTCONST const
+#      define EXTCONST     extern const
```
bram-perl: *commit (4) join short continuations*

I think you got the commit message wrong;
Looking at the next commit that has the message: '(5) line up simple defs' and that looks like a better message for this commit.
(And 'commit (4) join short continuations' looks like a better message for (5) line up simple defs')

bram-perl (commented):
Preamble:
- I reviewed commit-per-commit;
- Some commits I merely skimmed and didn't fully read
- I know some of the changes that I left comments on was changed again in a later commit (basically resolving the comment); I still left them in place tho just in case they might be useful for other cases.

regen/regcomp.pl
demerphq: Good catch. Thanks, fixed with the latest force push.

ext/XS-APItest/t/locale.t
demerphq: I updated the commit message. I suspect you are right, but the code in practice doesn't have such things. It was just that if you changed a define from `define foo bar` into `define foo__  bar` (assume the underbars are spaces here) the parse code would see the define name as `foo__` (that is 2 spaces) and then fail. Such white-space sensitivity is unhelpful. :-) And as you said, had it had a more complex value it would have gotten things worse.

gv.h
demerphq: I agree. I thought i had sorted this out properly in my most recent push, but checking my most recent push I think i missed something. Working on it.

handy.h
demerphq: Added a commit to drop them entirely.

op.h
demerphq: Yep, agreed. Working on it.

perl.h
demerphq: agreed.

perl.h
demerphq: agreed. :-)

XSUB.h
demerphq: I can try to come up with some heuristics, but there is a mixture of cases, some of which expect things to line up precisely, some which could do with an indent like you suggest. If i can figure out a more or less reliable heuristic to distinguish them I am game, but so far anything i tried like that seemed to help a small handful of cases and make the rest worse. This isn't done by hand obviously. :-)

cop.h
demerphq: Same comment as to your previous feedback. If I can figure out a way to tell when it is something like this:
```
#define X (foo
          |bar
          |baz)
```
from cases where the indent would make sense then I will. In general however i found preserving the existing line up as much as possible has produced the widest range of reasonable results.

cv.h
demerphq: Or collapse the parens entirely maybe. I can try something with this, its a special enough case it shouldnt cause much problems (compared to the previous two points).

EXTERN.h
demerphq: Oh yes, thanks, good catch. I was reordering the process a bit and the commit labels got out of synch. Fixed now. Good catch.

demerphq:
I hacked late last night, and i see that the latest push has some broken artifacts. Shoulda packed it in earlier. Just a FYI I know about it and im fixing.

cop.h
```diff
@@ -790,7 +790,7 @@ struct block_eval {
 
 /* blk_u16 bit usage for eval contexts: */
 
-#define CxOLD_IN_EVAL(cx)	(((cx)->blk_u16) & 0x3F) /* saved PL in_eval */
+#define CxOLD_IN_EVAL(cx)	(((cx)->blk_u16) & 0x3F) /* saved PL_in_eval */
```
bram-perl: *commit cop.h - fix variable name PL_in_eval*

I would change the commit message to something like:
- `commit cop.h - fix variable name PL_in_eval in comment`   OR
- `commit cop.h - fix variable name in comment (PL_in_eval)`

handy.h
```diff
@@ -1805,12 +1805,6 @@ END_EXTERN_C
 #   define isQUOTEMETA_(c)      (FITS_IN_8_BITS(c) && ! isWORDCHAR_L1(c))
 #   define _IS_IN_SOME_FOLD_ONLY_FOR_USE_BY_REGCOMP_DOT_C(c) isALPHA_L1(c)
 
-    /*  And these aren't accurate at all.  They are useful only for above
```
bram-perl: *commit handy.h - remove unused macros*

`_IS_IN_SOME_FOLD_ONLY_FOR_USE_BY_REGCOMP_DOT_C` should also be removed

intrpvar.h
```diff
@@ -998,8 +998,8 @@ PERLVARA(I, op_exec_cnt, OP_max+2, UV)	/* Counts of executed OPs of the given ty
                                            If PERL_TRACE_OPS is enabled, we'll dump
                                            a summary count of all ops executed in the
                                            program at perl_destruct time. For
-                                           profiling/debugging only. Works only if
-                                           DEBUGGING is enabled, too. */
+                                           profiling and debugging only. Works only
+                                           if DEBUGGING is enabled, too. */
```
bram-perl: *commit intrpvar.h - reword comment a bit to not use a slash for "and"*

I would probably repalce `Works only if DEBUGGING is enabled, too` with `Works only when DEBUGGING is enabled`, I think that is easier to read

Thinking some more: `For profiling and debugging only.` I would make that `Only for profiling and debugging`.

And one final thing I would be tempted to do: use `and/or` instead of `and`. (`profiling and debugging` could be read that it should only be used when you're debugging the profiling.. might be a bit of a stretch)

(PS: I don't care very hard about these; so if you decided not to change them that is good too)

zaphod32_hash.h
```diff
@@ -133,12 +133,17 @@ void zaphod32_seed_state (
...
+     *
+     * PI value in hex from here:
+     *
+     * http://turner.faculty.swau.edu/mathematics/materialslibrary/pi/pibases.html
+     *
+     * Ensure that the three state vectors are nonzero regardless of
+     * the seed. The idea of these two steps is to ensure that the 0
+     * state comes from a seed utterly unlike that of the value we
+     * replace it with.
+     */
```
bram-perl: *commit zaphod32_hash.h - rework comment to be more legible*

Reading this comment for the very first time leaves a confusing impression.
The start of the comment (`hex expansion of PI, ...´) lack context which made it very confusing.

I think something like this would be better:
```
    /* Ensure that the three state vectors are nonzero regardless of
     * the seed. The idea of these two steps is to ensure that the 0
     * state comes from a seed utterly unlike that of the value we
     * replace it with. This is accomplished by XOR-ing the seed
     * with PI in hex form (hex PI = 3.2[43f6...])
     */
```

The link it points to also appears to be dead so I would simply drop it from the code.

regen/regcomp.pl
```diff
@@ -701,6 +701,15 @@ sub print_reg_intflags_name {
                 chomp($next_line);
                 $line .= $next_line;
             }
+            if ( $line =~ m!/\*! ) {
+                while ( $line !~ m!\*/!) {
+                    defined(my $next_line= <$in_fh>)
+                        or last;
+                    chomp($next_line);
+                    $next_line=~s/^\s*/ /;
```
bram-perl: *commit regen/regcompl - deal with multiline comment*

(Typo in commit title? `regcompl` -> `regcomp.pl`?)

Being completely unfamiliar with regen/regcomp.pl: there is not enough context - for me - to understand why it matters. I can understand what the code does but I don't know why the multiline comment is relevant.

One thing I do wonder about, there are cases where comments are written as:
```
    /* foo
     * bar
     * baz */
```
Reading this code right it would make `$line` equal to `/* foo* bar* baz */`
Does that matter? No idea (since I lack the context)
Might be better to change the regex to something like (untested!): `s/^\s*(?:\*\s*)?/ /;` ?

XSUB.h
```diff
@@ -159,35 +159,36 @@ is a lexical C<$_> in scope.
 
 #define dAX const I32 ax = (I32)(MARK - PL_stack_base + 1)
 
-#define dAXMARK                         \
```
bram-perl: *commit (2) break multiline define*

I'm not sure what 'break multiline define' means..

gv.h
```diff
@@ -55,8 +55,11 @@ struct gp {
               &(GvXPVGV(_gvname_hek)->xiv_u.xivu_namehek);                                  \
             }))
 #  define GvNAME_get(gv)        ({ assert(GvNAME_HEK(gv)); (char *)HEK_KEY(GvNAME_HEK(gv)); })
+
 #  define GvNAMELEN_get(gv)     ({ assert(GvNAME_HEK(gv)); HEK_LEN(GvNAME_HEK(gv)); })
+
 #  define GvNAMEUTF8(gv)        ({ assert(GvNAME_HEK(gv)); HEK_UTF8(GvNAME_HEK(gv)); })
+
 #else
```
bram-perl: *commit (3) break long define*

This looks a bit bogus, either the commit message or the change..

util.h
```diff
@@ -70,7 +70,7 @@ typedef U64TYPE perl_drand48_t;
 #else
 
 struct PERL_DRAND48_T {
-    U16 seed[3];
+    U16     seed[3];
```
bram-perl: *commit (6) reformat structs*

That looks unnecessary, the struct only has one member so why the extra whitespace?
There also doesn't appear to be anything in the neighborhood with which it does line-up..
(It doesn't appear to line up with the later `typedef struct`s in the file)

handy.h
```diff
@@ -2874,7 +2877,8 @@ enum mem_log_type {
 #define Safefree(d)     safefree(MEM_LOG_FREE((Malloc_t)(d)))
 #endif
 
-/* assert that a valid ptr has been supplied - use this instead of assert(ptr)  *
+/* assert that a valid ptr has been supplied
+   - use this instead of assert(ptr) *
  * as it handles cases like constant string arguments without throwing warnings *
  * the cast is required, as is the inequality check, to avoid warnings          */
```
bram-perl: *commit (7) break remaining long line comments*

This also needs some checking:
The comment already continues on the next lines but these satart with `*`.
It appears each comment line also ends with `*`? Are there other places that do that?
In either case: if the `*` at the end is kept it should line up since right now it's confusing

(Also: if the first line of the comment is too long then surely the subsequent lines are also to long but these aren't split?)

hv.h
```diff
@@ -8,7 +8,8 @@
  *
  */
 
-/* These control hash traversal randomization and the environment variable PERL_PERTURB_KEYS.
+/* These control hash traversal randomization and
+   the environment variable PERL_PERTURB_KEYS.
  * Currently disabling this functionality will break a few tests, but should otherwise work fine.
```
bram-perl: *commit (7) break remaining long line comments*

Same as above, comment is already continued on next line but the next line starts with `*` where as the split line doesn't)

util.h
```diff
@@ -176,22 +176,30 @@ typedef struct {
...
-#define HSf_POPMARK         0x00000040                  /* popmark mode or you must supply ax and items */
-#define HSf_IMP_CXT         0x00000080                  /* ABI, threaded/MULTIPLICITY, pTHX_ present */
-#define HSm_INTRPSIZE       0xFFFF0000                  /* ABI, interp struct size */
-/* A mask of bits in the key which must always match between a XS mod and interp.
+#define HSf_POPMARK         0x00000040                  /* popmark mode or you
+                                                     must supply ax and items */
+#define HSf_IMP_CXT         0x00000080                  /* ABI,
+                                         threaded/MULTIPLICITY, pTHX_ present */
+#define HSm_INTRPSIZE       0xFFFF0000                  /* ABI, interp struct
+                                                           size */
```
bram-perl: *commit (7) break remaining long line comments*

Same as before, needs to line up

XSUB.h
```diff
@@ -16,47 +16,45 @@
 /*
 =for apidoc_section $XS
 
-F<xsubpp> compiles XS code into C.  See L<perlutil/xsubpp>.
```
bram-perl: *commit (8) reflow_multiline_comments*

Define 'reflow'?

utf8.h
```diff
@@ -1051,8 +1050,7 @@ this macro matches
  * U+10FFFF: \xF4\x8F\xBF\xBF   \xF9\xA1\xBF\xBF\xBF    max legal Unicode
  * U+110000: \xF4\x90\x80\x80   \xF9\xA2\xA0\xA0\xA0
  * U+110001: \xF4\x90\x80\x81   \xF9\xA2\xA0\xA0\xA1
- */
-#define UTF_START_BYTE_110000_          UTF_START_BYTE(PERL_UNICODE_MAX + 1, 21)
+ */#define UTF_START_BYTE_110000_          UTF_START_BYTE(PERL_UNICODE_MAX + 1, 21)
```
bram-perl: *commit (8) reflow_multiline_comments*

This is bogus.

utf8.h
```diff
@@ -1288,8 +1286,8 @@ point's representation.
 #define UNICODE_GREEK_SMALL_LETTER_FINAL_SIGMA  0x03C2
 #define UNICODE_GREEK_SMALL_LETTER_SIGMA        0x03C3
 #define GREEK_SMALL_LETTER_MU                   0x03BC
-#define GREEK_CAPITAL_LETTER_MU                 0x039C                      /* Upper and title case
-                                                           of MICRON */
+#define GREEK_CAPITAL_LETTER_MU                 0x039C                      \
+/* Upper and title case of MICRON */
```
bram-perl: *commit (8) reflow_multiline_comments*

Also doesn't look great

cop.h
demerphq: Willdo.

handy.h
demerphq: dioh, i removed and then forgot to update the shell variable that i use for resetting back to and nuked it. :-( 

Will redo, thanks and sorry for wasting your time with this again.

intrpvar.h
demerphq: I don't mind rewording it. Ill give it a go.

zaphod32_hash.h
demerphq: Ok, willdo. That comment is ancient and id like us to drop zaphod hash sometime anyway, its only there for 32 bit only builds. Everybody else should be on 64 bit Siphash.

regen/regcomp.pl
demerphq: Good questions, I can update the code with a comment to explain, and harden the code further to deal with star style comments, although in practice we don't use them on these defines it might save someone some surprise in the future. Below is some context:
  
<details>

The code in question is trying to parse out defines to build up documentation and introspection datasets in regnodes.h so we can show the flags that are set on a pattern in re 'debug' output. If you do a `git grep NAUGHTY` you will see a good example. The loop in question is looking for things like this:

```
#define PREGf_XXX 0x100 /* the xxx flags */
```

And it wants to turn into a tuple like this `["XXX", 0x100, "the xxx flags"]` for further processing.

However if and when the comment gets wrapped to this:

```
#define PREGf_XXX 0x100 /* the xxx 
                           flags */
```

The old logic would fail to match entirely and the given flag. With the new logic we would handle it correctly, first by noticing the define at all, and second by unwrapping the comment. You are correct that if the comments were "star style comments" that the unwrap wouldn't work. 

But currently they are all one line, the only reason they get wrapped are they are overlong, and my script wraps them and when it does so it does not add stars, so this change armors against that wrapping to come in the PR. It wouldn't hurt to armor this logic against start style comments as well, so i will add that, but strictly speaking it isn't necessary for this as the comments in question already exist and dont use that style.

Just FYI, regen/regcomp.pl takes the node and state definition data in regcomp.sym and combines it with metadata from a few of the other regex header files, and then produces regnodes.h, which is what the regex engine actually uses for processing. PREGf flags are special flags that allow the regex engine to determine how to run the pattern and some subtleties like that. For instance PREGf_GPOS is set if the regex engine detects that a pattern might go hyperexponential in performance (IIRC). 

```
$ ./perl -Ilib -Mre=Debug,DUMP,FLAGS -e'qr/\Gfoo/'
Compiling REx "\Gfoo"
Final program:
   1: GPOS (2)
   2: EXACT <foo> (4)
   4: END (0)
anchored "foo" at 0..0 (checking anchored noscan) anchored(GPOS) GPOS:0 minlen 3 
r->extflags: USE_INTUIT_NOML USE_INTUIT_ML IS_ANCHORED 
r->intflags: NOSCAN GPOS_FLOAT RECURSE_SEEN 
Freeing REx: "\Gfoo"
```
the logic we are discussing is required output the `r->intflags` and `r->extflags` lines.
</details>

gv.h
demerphq: Hrm. Good catch ill dig into that, it shouldnt happen.

util.h
demerphq: Reformat structs just applies the same layout logic to all structs.  Its not just lining them up, its also deciding which tab stop they are lined up on, and I believe I deliberately did this to ensure there was at least one empty tab stop in between type and field. I can undo it, but i found it made things more legible with structs with 3 character types, which is fairly common for us, also it provides a bit more leeway to add a longer type without needing to reindent the struct again.  

If there was consensus we should do something else I am fine with that, but i made this choice because it opened up the struct definitions and made them "breathe" a bit which I think makes them more readable and comprehensible. I guess we could say "less than K lines uses a minimal indent", with K being 3 or something, and for structs with more members we would use the current logic. We have some structs that have dozens of integer type fields, and i found that when they were "close packed" (eg allow a minimum of one space between type and field) they became an eyesore to read.

In short, while I agree this might be controversial, this is "not a bug". :-)

I can try re-pushing without this deliberate rule and we can take a look, but it will affect every struct. Which is a point worth considering with these changes in the PR, the logic might make a few minor cases slightly "worse" in some aesthetic sense while fixing a whole bunch of other cases and being a net win.  That is a the problem with standardizing things, it can be hard to codify rules that lead to the desired outcome in every case across all our code.  

Also I guess at the end we can just rebase-interactive some of the controversial parts, but I'm hoping to automate the vast bulk of this so that if we were to apply it, and someone had changes that turned into conflict hell, we could run the script over their pre-image headers to get the same format, which then could be merged with far less conflicts.

handy.h
demerphq: good catch. the right hand side '*'s are confusing the line break logic. 

BTW, while i realize its been tedius to review, this is super useful feedback, I try to manual review all the changes for things like this, but doing it over and over is a bit mind numbing and its easy to miss a few things in the forest of changes.

hv.h
demerphq: yep. willfix.

util.h
demerphq: Ah. This one is actually a "difficult case" of a few issues piling up together that I am already working on. The first unforunate  decisions is that the line up code is lining up values like '0x0000040' with defines for code expressions. This is then forcing the columns wider than they should be, which is then eating up the space for where the comment goes, which is then leaving the comment wrapping logic only 20 columns to work with, and this comment contains "threaded/MULTIPLICITY", which is a very long word, so then it cant line it up with the rest without breaking the line limit, and then it gives up.

FWIW, this is not the last step, and the last step is intended to try to help with this case. However, I am trying to approach this a different way, and come up with a better partitioning algorithm/heuristic so lists of numbers like this (which very often have sidebar comments) are lined up independently and thus put the name and value tighter together so that there is more space to wrap the side-bar comment. Having said that there comes a certain point where I might address this by just removing the "/" in "threaded/MULTIPLICITLY" so this can be wrapped more easily.

XSUB.h
demerphq: Basically it is running Text::Autoformat on all non-tabular looking comment paragraphs.  One downside is that TA does not do "two spaces after period". Which is these days less common than it once was. I didn't bother to filter out "this paragraph already fits", as it seemed to do a good enough job and it complicated things a fair bit to add it. I debated this one. On one hand consistent formatting is good, on the other hand unnecessary change is bad. I narrowly decided that there are enough cases where all the lines in an existing paragraph fit into the margin so technically dont need to be changed, but where the layout has been "wonked" for some reason such that lines are highly irregular lengths and difficult to read that I should just format everything so it was consistently styled and wrapped by the same algorithm.  Text::Autoformat is a bit smarter than most wrapping algorithms and leads to more readable text IMO than Text::Wrap.

utf8.h
demerphq: Indeed it is. Interesting nothing complained about this when i compiled. I wonder why. Ill definitely fix this.

utf8.h
demerphq: Sigh, thanks. I missed that. This is logic that fixed one weird edge case (long `#if` statement with super long side bar comment) breaking something that it shouldnt be applied to. Will fix.

XSUB.h
demerphq: Ah, i guess you are being thrown off because this step includes two steps, "break multiline define" and "fixup line continuations and related ws", as do most of the steps following it. I should break the first one out to its own step. "Break multiline define" means take multiline defines which have a value (replacement) segment on the define line itself, and move that first line down. So this

```
#define Foo(x,y,z) STMT_START { \
```
turns into this:
```
#define Foo(x,y,z) \
    STMT_START { \
```
as this affects line continuations it is followed by a second step which cleans up the line continuations. But the first time the line continuations are changed should be an explicit step.

khwilliamson:
I didn't look at per-commit, but at the differences between this and blead.

Overall, this is a big improvement over what we now have, IMO.  Here are the exceptions to that I see

It is controversial whether sentences should be separated by 1 vs 2 blanks after the terminating punctuation mark.  I don't think therefore it is a good idea to defacto make a single space our standard.  Thus I dont agree with the removal of double-spaces in XSUB.h, for example, or in a number of comments throughout.

There are a bunch of places where things are needlessly shifted to the right, so that something that used to fit in 80 columns no longer does.  I can't get github to show this property, as it squashes strings of white space, but an example is  in cop.h lines 793,795 changing to 802,804.  

And there are places with the opposite problem, like lines 733-735 of inline.h where comments were squeezed to the left so that there's only a single space separating them from what they are commenting.  That lost visual separation is helpful.

After seeing so many of the above, I gave up and looked at a diff with ignoring all white space changes.  It looks like a bunch of this is machine generated, so the algorithm could be improved.  We have this change, which makes no sense

```871,872c891,892
   < #define ANYOF_UNIPROP   ((ANYOF_POSIXL_MAX)+4)  /* Used to indicate a Unicode
   <                                                    property: \p{} or \P{} */
   ---
   > #define ANYOF_UNIPROP                           ((ANYOF_POSIXL_MAX)+4)          \
   > /* Used to indicate a Unicode property: \p{} or \P{} */
```

creating a continuation line just for a comment

In inline.h 2096,2101c2097,2102 suddenly the comments lines are inappropriately indented.  vim does that to me sometimes.




`

demerphq:
@khwilliamson thanks for the feedback, ill review in detail later. 

ALL: I messed up and fat finger pushed a broken version of the PR, ill fix it up later, possibly as much as week from now as I will be pretty busy the coming week and I dont know how much hacking time ill get. Any feedback though I will follow up on.

demerphq:
On Thu, 1 Sept 2022, 18:40 Bram, ***@***.***> wrote:

> I think the first commits can be merged as-is,
>
> For the whitespace cleanup commit, skimming the files:
>
>    - op.h doesn't look good:
>       - lines 11-40 are misaligned
>       - lines 111-167: comments that continue on the next line are no
>       longer aligned
>    - perl.h:
>       - lines 5223-5555
>       - 5556-5599
>       - 5605-5636
>       - 5642-5681
>       - 5842-5849
>    - utfebcdic.h:
>       - lines 30-32
>
> Yeah, I found some bugs that I think would explain this. I'll review.



>    -
>
> Other things *I* am not fond of:
>
> (Taking a random example)
> regexp.h
>
>         struct {
>             /* this first element must match u.yes */
>             struct regmatch_state   *prev_yes_state;
>             U32                     lastparen;
>             U32                     lastcloseparen;
>             CHECKPOINT              cp;
>
>         }                            branchlike;
>
> The name of the struct is too far to the right (again IMO)
>

Not just you. I pushed that because I sometimes notice issues there I don't
in a local diff. I think I fixed that locally. I'll repush at some point.

Please let me know any other examples that annoy you. Likely I will want to
fix them. But perhaps wait my next push to see what fixes I have queued up
already.

Yves

>

khwilliamson:
After thinking about this some more, I don't think we should be adjusting white space in comments other than to wrap text and expand any tabs.

I do have a perl function that does that in a WIP workspace that I could polish up.

demerphq:
> After thinking about this some more, I don't think we should be adjusting white space in comments other than to wrap text and expand any tabs.

@khwilliamson at first I thought the same, but over time names have been mass adjusted, text changed, etc, with sometimes unfortunate results.  I dont mind excluding "block comments" from my code, as for the stuff i really care about (making the code and comments readable) mostly concerns "hanging comments".

But FWIW, I have taken the time to write my own text reflow around Text::Wrap, (i am working with damian conway to make Text::Autoformat better with C comments but for now i have switched away from it). As part of that i have hacked support for "two spaces after period". and some other stuff. ..

I just did a push with the latest state, and it seems to be much improved these days over how it looked in earlier days. Maybe take a check? 

BTW, id love to hear why you think we should "just leave it alone". Is there a specific reason? What about when the comment is over 80 columns because something was renamed with a search and replace? Wouldnt it be better to have a consistent format as much as possible?

My thought was that once the output of this was stable and more or less agreed on, id make it into a Porting tool. Then when people like you or me (or whomever) make mass changes that might mess things up we can just run the script.  Then if you add a new define the tool will realign for you, or whatever.

khwilliamson:
I said that I don't object to wrapping.

The problem I see in adjusting comment blanks in general is that you don't know what the author was doing that actually was planned, like creating a table or other construct where things in adjacent rows line up vertically for clarity or to make things easier to read.

demerphq:
Ok, so only touch a block comment if it actually contains a line that is longer than desired?

Anyway, for what its worth i put some effort into making it avoid tables. Also anything in a comment that "verbatim style" is left alone. A block of whitespace indented text inside of a large comment will be left alone also. I was thinking if we made this a permanent thing we could also have some special notation to say "this leave this please".

khwilliamson:
On 10/28/22 08:50, Yves Orton wrote:
> Ok, so only touch a block comment if it actually contains a line that is 
> longer than desired?

Your proposal below likely stops my concerns

> 
> Anyway, for what its worth i put some effort into making it avoid 
> tables. Also anything in a comment that "verbatim style" is left alone. 
> A block of whitespace indented text inside of a large comment will be 
> left alone also. I was thinking if we made this a permanent thing we 
> could also have some special notation to say "this leave this please".

This may be good enough

khwilliamson:
I finally started to review this closely, but I want to compare it with blead, and it needs to be rebased.  I should now have some time so if @demerphq rebases it, I will look at it soon thereafter.

demerphq:
@khwilliamson I dropped this for a bit. Ill get it regenerated onto the latest version of blead ASAP.

XSUB.h
```diff
@@ -23,32 +23,31 @@ Some variables below are flagged with 'u' because Devel::PPPort can't currently
...
+Variable which is setup by C<xsubpp> to indicate the class name for a C++ XS
+constructor.  This is always a C<char*>.  See C<L</THIS>>.
 
 =for apidoc Amnu|type|RETVAL
-Variable which is setup by C<xsubpp> to hold the return value for an
-XSUB.  This is always the proper type for the XSUB.  See
-L<perlxs/"The RETVAL Variable">.
+Variable which is setup by C<xsubpp> to hold the return value for an XSUB.
+This is always the proper type for the XSUB.  See L<perlxs/"The RETVAL
+Variable">.
```
demerphq: I need to fix this.

demerphq: fixed.

demerphq:
Oh, hrm, that chain of reflow C comments at the end means something is wrong. I will check what it is.

demerphq:
and fixed.

jkeenan:
With the merger of https://github.com/Perl/perl5/pull/20442 into blead today, is this pull request still needed?  If so, merge conflicts will need to be resolved and the p.r.'s status will have to get out of `Draft`.

demerphq:
On Fri, 30 Dec 2022 at 22:14, James E Keenan ***@***.***>
wrote:

> With the merger of #20442 <https://github.com/Perl/perl5/pull/20442> into
> blead today, is this pull request still needed? If so, merge conflicts will
> need to be resolved and the p.r.'s status will have to get out of Draft.
>
I'll get this rebased. FWIW, yes, it is still a work in progress or even
just a discussion starter, which is why it is stalled in draft status. A
nice chunk of it goes away with #20442 however.

Yves

bulk88:
I would appreciate the tool. Less thinking/time for everyone, what the "right way" is for the p5p repo.

PR #20210, opened 2022-09-01
Math-Complex: correct typos in documentation

pjacklam:
- Correct typo where "pi" should be "phi".

- Spherical coordinates should use the same order of the coordinates
  everywhere, i.e., rho, theta, phi.

### DIFF
--- a/dist/Math-Complex/lib/Math/Trig.pm
+++ b/dist/Math-Complex/lib/Math/Trig.pm
@@ -466,15 +466,15 @@ B<All angles are in radians>.
 
 B<Cartesian> coordinates are the usual rectangular I<(x, y, z)>-coordinates.
 
-Spherical coordinates, I<(rho, theta, pi)>, are three-dimensional
+Spherical coordinates, I<(rho, theta, phi)>, are three-dimensional
 coordinates which define a point in three-dimensional space.  They are
 based on a sphere surface.  The radius of the sphere is B<rho>, also
 known as the I<radial> coordinate.  The angle in the I<xy>-plane
 (around the I<z>-axis) is B<theta>, also known as the I<azimuthal>
 coordinate.  The angle from the I<z>-axis is B<phi>, also known as the
-I<polar> coordinate.  The North Pole is therefore I<0, 0, rho>, and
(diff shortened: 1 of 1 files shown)


pjacklam:
This fixes [CPAN RT #136653](https://rt.cpan.org/Ticket/Display.html?id=136653).

Grinnz:
This distribution is maintained on CPAN, though permissions have been transferred to P5P, it would need to be moved from cpan/ to dist/ before changes can be made to it in blead.

Grinnz:
requires https://github.com/Perl/perl5/pull/20215

jkeenan:
Can we have someone handle the *content* of this pull 

> * Correct typo where "pi" should be "phi".
> 
>     * Spherical coordinates should use the same order of the coordinates
>       everywhere, i.e., rho, theta, phi.

Can we have someone handle the *content* of this pull request?

(I or someone else can handle the rebasing, etc., separately once the content is approved.)

jkeenan:
Having wrapped my head around the [Wikipedia article on spherical coordinates](https://en.wikipedia.org/wiki/Spherical_coordinate_system) -- specifically, around the sidebar on *Spherical coordinates (r, θ, φ) as often used in mathematics:* -- I believe this patch is correct and should be applied.  I will do so within 7 days unless there is some objection.

shlomif:
@jkeenan : the changes look right to me too (and I studied that in the Technion).

demerphq:
Personally I think @Grinnz has a point. I dont think we need to be as strict as he says, we can patch it in cpan/ and THEN move it to dist, provided we actually move it to dist/ in the end.

demerphq:
OH, we have already moved it to dist.

demerphq:
@jkeenan why wait?

jkeenan:
> @jkeenan why wait?

Among other reasons ... at this time of year many people are not reading their email.  So giving people (including the OP) a time frame in which to respond does not hurt.

(I could give other reasons, but then we would be deep into a policy discussion which should take place on the perl5-porters mailing list.)

jkeenan:
Handled manually and merged to blead in 7b4ead7ffc.  Closing ticket.

PR #20211, opened 2022-09-01
locale.c: Add debugging check

khwilliamson:
If we run across the REPLACEMENT CHARACTER, its likely a mismatch
between LC_CTYPE and another category.  This panics on debugging builds.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -1407,7 +1407,7 @@ S_stdize_locale(pTHX_ const int category,
     return retval;
 }
 
-#if defined(USE_POSIX_2008_LOCALE)
+#if defined(USE_POSIX_2008_LOCALE) || ! defined(LC_ALL)
 
 STATIC
 const char *
@@ -1464,7 +1464,7 @@ S_calculate_LC_ALL(pTHX_ const char ** individ_locales)
 
     /* First calculate the needed size for the string listing the categories
      * and their locales. */
-    for (i = 0; i < LC_ALL_INDEX_; i++) {
+    for (i = 0; i < NOMINAL_LC_ALL_INDEX; i++) {
 
 #  ifdef USE_QUERYLOCALE
         const char * entry = querylocale_l(i, cur_obj);
@@ -1484,7 +1484,7 @@ S_calculate_LC_ALL(pTHX_ const char ** individ_locales)
     SAVEFREEPV(Newxz(aggregate_locale, names_len, char));
 
     /* Then fill it in */
-    for (i = 0; i < LC_ALL_INDEX_; i++) {
+    for (i = 0; i < NOMINAL_LC_ALL_INDEX; i++) {
         Size_t new_len;
 
 #  ifdef USE_QUERYLOCALE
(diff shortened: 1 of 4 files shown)


locale.c
```diff
@@ -2748,6 +2748,20 @@ S_save_to_buffer(const char * string, const char **buf, Size_t *buf_size)
                           *buf));
             }
 
+#    ifdef DEBUGGING
```
demerphq, line 2741: FWIW, I would prefer we were consistent about whether and when we outdent ifdefs, personally i would not outdent this, and not outdenting seems to be the standard (Although i guess some of these hits could be from code where no outdent was required).

```
~/git_tree/perl$ git grep -P '^#\s+ifdef DEBUGGING' | wc -l
25
~/git_tree/perl$ git grep -P '^#ifdef DEBUGGING' | wc -l
208
```

locale.c
```diff
@@ -2748,6 +2748,20 @@ S_save_to_buffer(const char * string, const char **buf, Size_t *buf_size)
...
+     * locale as whatever is being worked on */
+    if (UNLIKELY(instr(string, REPLACEMENT_CHARACTER_UTF8))) {
+        dTHX_DEBUGGING;
+
+        locale_panic_(Perl_form(aTHX_
+                                "Unexpected REPLACEMENT_CHARACTER in '%s'\n%s",
+                                string, get_LC_ALL_display()));
+    }
+
+#    endif
```
demerphq, line 2753: Same issue with outdenting. Other than that question, which is debatable and even arguable controversial, the patch LGTM.

demerphq (approved):

locale.c
khwilliamson: From perlhack:

>   4-wide indents for code, 2-wide indents for nested CPP "#define"s,

I think the wording on that is wrong, it meant to include any nested CPP directives.  Otherwise it doesn't make sense

It doesn't much matter on files that don't have much nesting.  But  locale.c has a lot due to the vagaries of various platforms.  When I changed that to have indenting, it made it a lot clearer

> § git grep -P '^#\s+(if|el|endif)' | wc -l 
7290
git grep -P '^#(if|el|endif)' | wc -l
17141

jkeenan:
On a threaded debugging build on FreeBSD-12, with clang10 as the compiler, build failed early:
```
$ sh ./Configure -des -Dusedevel -Duseithreads -DDEBUGGING && make test_harness
...
cc -c -DPERL_CORE -DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -g -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings miniperlmain.c
cc -pthread -Wl,-E  -fstack-protector-strong -L/usr/local/lib -o miniperl  opmini.o perlmini.o universalmini.o  gv.o toke.o perly.o pad.o regcomp.o dump.o util.o mg.o reentr.o mro_core.o keywords.o builtin.o hv.o av.o run.o pp_hot.o sv.o pp.o scope.o pp_ctl.o pp_sys.o peep.o doop.o doio.o regexec.o utf8.o taint.o deb.o globals.o perlio.o numeric.o mathoms.o locale.o pp_pack.o pp_sort.o caretx.o dquote.o time64.o  miniperlmain.o  -lpthread -ldl -lm -lcrypt -lutil
ld: error: undefined symbol: get_LC_ALL_display
>>> referenced by locale.c:2758
>>>               locale.o:(S_save_to_buffer)
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1

Stop.
[perlmonger: perl] $ clang --version
FreeBSD clang version 10.0.1 (git@github.com:llvm/llvm-project.git llvmorg-10.0.1-0-gef32c611aa2)
Target: x86_64-unknown-freebsd12.3
```

jkeenan:
Okay, build and test are now completing on the replacement branch with debugging build.
```
$ uname -mrs
FreeBSD 12.3-RELEASE-p1 amd64
$ gitcurr
replacement
$ git describe
v5.37.2-424-gbe556cf0e4
$ ./perl -Ilib -V:config_args
config_args='-des -Dusedevel -Duseithreads -DDEBUGGING';

Test Summary Report
All tests successful.
```

PR #20215, opened 2022-09-01
Move Math-Complex from cpan/ to dist/

khwilliamson:
This module is now being maintained by p5p now.


leonerd:
This seems to be fine... Are we waiting on any specific comments or reviews? How does PAUSE permissions for this alter - do we need to talk to the CPAN admins about it?

Grinnz:
P5P has first-come on the modules, so permissions can be given to a releaser as needed. Someone(tm) should release the distribution to CPAN once it has been moved since dist/ entries are not indexed in perl releases.

t/porting/customized.dat
```diff
@@ -4,17 +4,7 @@
...
 Locale::Maketext::Simple cpan/Locale-Maketext-Simple/lib/Locale/Maketext/Simple.pm 57ed38905791a17c150210cd6f42ead22a7707b6
-Math::Complex cpan/Math-Complex/lib/Math/Complex.pm 66f28a17647e2de166909ca66e4ced26f8a0a62e
-Math::Complex cpan/Math-Complex/t/Complex.t 17039e03ee798539e770ea9a0d19a99364278306
-Math::Complex cpan/Math-Complex/t/Trig.t 508f8e27373c08228be13ca5d42b28812ab0e020
-Math::Complex cpan/Math-Complex/t/underbar.t 97e7b9615658eefc67a710d4b258349cc5bace63
-Memoize cpan/Memoize/Memoize.pm 902092ff91cdec9c7b4bd06202eb179e1ce26ca2
-Memoize cpan/Memoize/t/errors.t bcd0c02a4bc47dfda07a97d265d7210849871659
-Memoize cpan/Memoize/t/expmod_t.t a1c3d03fd3ec2e7c6c835f02316475756e1b16b7
-Memoize cpan/Memoize/t/speed.t 89fe8c8928f0164c9ed898b5a427dbbc814e9976
-Memoize cpan/Memoize/t/tie_gdbm.t d81c4f6208f6c925539b3878bff9a8ff1b73a37e
```
bram-perl: You may need to double check the list of removed customizations, it removes some that are unrelated to Math::Complex
Also, in #20216 @tonycoz noted that customization of 'Filter::Util::Call' is still present (**I** did not verify)

Leont:
Are we sure we want to do this? IMO we should generally prefer to have modules on CPAN unless there's a reason to do otherwise (e.g. when it's tied to core development the way Devel::PPPort is). Delegating maintenance of other modules is the more manageable option IME.

jkeenan:
> Are we sure we want to do this? IMO we should generally prefer to have modules on CPAN unless there's a reason to do otherwise (e.g. when it's tied to core development the way Devel::PPPort is). Delegating maintenance of other modules is the more manageable option IME.

If @pjacklam is willing to be the maintainer of this module on CPAN, then I would endorse that approach.  He would be able to apply patches and test them out against older Perl versions directly, without having to wait around for us (P5P) to approve his patches.

Only complication is that we haven't heard from the module's current CPAN maintainer in years.  That's why I think the Perl Steering Council should discuss this with PAUSE administrators.

khwilliamson:
On 9/2/22 14:48, James E Keenan wrote:
>     Are we sure we want to do this? IMO we should generally prefer to
>     have modules on CPAN unless there's a reason to do otherwise (e.g.
>     when it's tied to core development the way Devel::PPPort is).
>     Delegating maintenance of other modules is the more manageable
>     option IME.
> 
> If @pjacklam <https://github.com/pjacklam> is willing to be the 
> maintainer of this module on CPAN, then I would endorse that approach. 
> He would be able to apply patches and test them out against older Perl 
> versions directly, without having to wait around for us (P5P) to approve 
> his patches.
> 
> Only complication is that we haven't heard from the module's current 
> CPAN maintainer in years. That's why I think the Perl Steering Council 
> should discuss this with PAUSE administrators.
> 
>
I thought @pjacklam said that the PAUSE administrators were not 
responding or not responding favorably to his wanting to be comaint.,

If I am right, that is unconscionable, and the system needs immediate reform

pjacklam:
The PAUSE admins replied that I should contact the maintainers again. However, as I had contacted the maintainers in 2015 and again in 2020 without success, I didn't see any point in contacting them yet again. So I gave up Math-Complex and created my own modules for elementary functions, spherical trigonometry etc. (I have a background in applied mathematics and numerical analysis and work with air navigation).

The reason why I have submitted pull requests recently is that I noticed Math-Complex in core (1.59_02)  is ahead of the version on CPAN (1.59). I thought my pull requests might be a way to get some progress on Math-Complex.

It is not important for me to be the one in charge of Math-Complex. I just wish there was someone, not necessarily me. If I remember correctly, @shlomif also wanted co-maint some years ago. Perhaps he is still interested?

khwilliamson:
@leonerd @rjbs @book 
We need some action on this

rjbs:
@khwilliamson You should just move this to dist, thanks.  If we get a strong contender for fixing things, we can move it back (easy) or just accept a bunch of fixes to the library.

[ edit:  I originally wrote "to cpan", which was just a brain-to-fingers error ]

rjbs (approved):

khwilliamson:
On 9/30/22 08:26, Ricardo Signes wrote:
> @khwilliamson <https://github.com/khwilliamson> You should just move 
> this to cpan, thanks. If we get a strong contender for fixing things, we 
> can move it back (easy) or just accept a bunch of fixes to the library.

It already is in cpan.  I presume you meant move it to dist

t/porting/customized.dat
khwilliamson: fixed

PR #20216, opened 2022-09-02
Net-Ping: No longer need to TODO some tests when on FreeBSD

jkeenan:
See Net-Ping maintainer's comment:
https://github.com/rurban/Net-Ping/pull/18#issuecomment-1234223370

### DIFF
--- a/t/porting/customized.dat
+++ b/t/porting/customized.dat
@@ -12,8 +12,8 @@ Math::Complex cpan/Math-Complex/t/Trig.t 508f8e27373c08228be13ca5d42b28812ab0e02
 Math::Complex cpan/Math-Complex/t/underbar.t 97e7b9615658eefc67a710d4b258349cc5bace63
 Net::Ping dist/Net-Ping/t/000_load.t deff5dc2ca54dae28cb19d3631427db127279ac2
 Net::Ping dist/Net-Ping/t/001_new.t 7b24e05672e22edfe3e6b5cc0277f815efe557e5
-Net::Ping dist/Net-Ping/t/010_pingecho.t 218d7a9ee5b6d03ba2544210acaf6585f8dc5503
-Net::Ping dist/Net-Ping/t/450_service.t f6578680f2872d7fc9f24dd75388d55654761875
+Net::Ping dist/Net-Ping/t/010_pingecho.t 74437379673c8ba82f81574fd7e652f41a65c993
+Net::Ping dist/Net-Ping/t/450_service.t 4963cd2240f1e583aa950f2bbe8c478ceba3b949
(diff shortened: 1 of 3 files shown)


jkeenan:
Anyone with access to FreeBSD:  Please smoke-test this branch:
https://github.com/Perl/perl5/tree/smoke-me/jkeenan/untodo-net-ping-tests-freebsd-20220901

t/porting/customized.dat
```diff
@@ -4,21 +4,15 @@
 AutoLoader cpan/AutoLoader/t/02AutoSplit.t bb90cda13b88599ad45de4b45799d5218afcb6d8
 ExtUtils::Constant cpan/ExtUtils-Constant/lib/ExtUtils/Constant/Base.pm 7560e1018f806db5689dee78728ccb8374aea741
 ExtUtils::Constant cpan/ExtUtils-Constant/t/Constant.t 165e9c7132b003fd192d32a737b0f51f9ba4999e
-Filter::Util::Call pod/perlfilter.pod 545265af2f45741a0e59eecdd0cfc0c9e490c1e8
```
tonycoz: This customization is still present in `pod/perlfilter.pod`.

jkeenan: This is curious. After changing the two Net-Ping tests I ran make test_harness and got a test failure in t/porting/customized.t. I opened up t/porting/customized.dat and saw the instructions about re-running ./perl -I../lib porting/customized.t --regen, after which make test_porting PASSed.

Now, I admit that I did not look at the changes in customized.dat before pushing the branch and creating the merge request. But the fact that there were lots of changes in customized.dat other than those for Net-Ping implies that the --regen program had not been run in a long time and that make test_porting was failing to recognize that.

Does that logic seem correct? If so, how should we proceed?

jkeenan: This may be a side effect of the fact that `pod/perlfilter.pod` originates on CPAN: https://metacpan.org/release/RURBAN/Filter-1.64/source/perlfilter.pod.

jkeenan:
This depends on https://github.com/Perl/perl5/pull/20237, and it will probably need rebasing on blead and some refinement.

jkeenan:
> This depends on #20237, and it will probably need rebasing on blead and some refinement.

https://github.com/Perl/perl5/pull/20237 having been merged into blead, this pull request has been rebased on blead and the customized database has been regenerated.  Assuming its second round of CI and smoke-testing results are good, it should be ready to merge.

t/porting/customized.dat
jkeenan: > This customization is still present in `pod/perlfilter.pod`.

I believe this should be resolved with the merge of https://github.com/Perl/perl5/pull/20237 and subsequent refinements to this p.r.

tonycoz (approved):

PR #20217, opened 2022-09-02
Configure: remove nsl from libswanted

danielnachun:
`libnsl` has been deprecated in `glibc` since at least `glibc` 2.27 and has been superseded by `libnsl2`.  In Homebrew we have been migrating packages that were previously using `libnsl` from `glibc` to `libnsl2`. In doing so we've found that while Configure checks for `libnsl`, there is no dynamic linkage we can find in any binaries to `libnsl` whether it is provided by `glibc` or `libnsl`.

It does not make sense to try to link to a library while configuring that will not actually be linked to in the final binary, which makes me think this is an oversight.  Perhaps `libnsl` was at one point linked to by an XS module that is no longer installed by default?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -335,6 +335,7 @@ Daniel P. Berrange             <dan@berrange.com>
 Daniel Perrett                 <perrettdl@googlemail.com>
 Daniel S. Lewart               <lewart@uiuc.edu>
 Daniel Yacob                   <perl@geez.org>
+danielnachun                   <daniel.nachun@gmail.com>
 Danny R. Faught                <faught@mailhost.rsn.hp.com>
 Danny Rathjens
 Danny Sadinoff                 <danny-cpan@sadinoff.com>
--- a/Configure
+++ b/Configure
@@ -1505,7 +1505,7 @@ archname=''
 usereentrant='undef'
 : List of libraries we want.
 : If anyone needs extra -lxxx, put those in a hint file.
-libswanted="cl pthread socket bind inet nsl ndbm gdbm dbm db malloc dl ld"
+libswanted="cl pthread socket bind inet ndbm gdbm dbm db malloc dl ld"
 libswanted="$libswanted sun m crypt sec util c cposix posix ucb bsd BSD"
 : We probably want to search /usr/shlib before most other libraries.
 : This is only used by the lib/ExtUtils/MakeMaker.pm routine extliblist.


Leont:
Linking to libnsl goes way back to perl 1.0 back in 1987. I'm sure it made sense then, and odds are there will be some traditional unices where this is still necessary, but it's a poor default in 2022

danielnachun:
> Linking to libnsl goes way back to perl 1.0 back in 1987. I'm sure it made sense then, and odds are there will be some traditional unices where this is still necessary, but it's a poor default in 2022

That was what I guessed.  There is already code for adding optional dependencies such as SOCKS with the `usesocks` option so I could refactor this to have a `usensl` option that adds `nsl` if specified by the user.  That leaves the option of linking to `libnsl` open if it is actually needed, without including it as a default.

danielnachun:
I pushed a commit to add `usensl` as an option to enable linking `libnsl` if needed.

Leont:
> I pushed a commit to add usensl as an option to enable linking libnsl if needed.

I don't think that is a good solution; as that would mean you can't build perl without passing that option on affected platforms.

A better solution would be to add it in the hints files of affected platforms (if any, we're not sure of that at this stage).

danielnachun:
> > I pushed a commit to add usensl as an option to enable linking libnsl if needed.
> 
> I don't think that is a good solution; as that would mean you can't build perl without passing that option on affected platforms.
> 
> A better solution would be to add it in the hints files of affected platforms (if any, we're not sure of that at this stage).

Yes that sounds like a better solution.  Is there a systematic way for us to determine which platforms still need to link `libnsl`?  I can see that some platforms explicitly remove `libnsl` from `libswanted` in their hints files so we can be certain those don't need to link `libnsl`.

One thing that is puzzling to me is that I can't find any place in the source tree where the `libnsl` header files (which all begin with `yp` as NIS was once called Yellow Pages) are included.  If there were references in the code then we could work out which platforms would need it.  I took a look at the `Net::NIS` module in CPAN and it does include `yp_prot.h` and `ypclnt.h` from `libnsl` in its `NIS.xs`.  But `Net::NIS` isn't included in the `ext` folder here as a default module, and its `Makefile.PL` adds `-lnsl` to `LIBS`, so it does its own checking for `libnsl`.

Leont:
> Yes that sounds like a better solution. Is there a systematic way for us to determine which platforms still need to link libnsl? I can see that some platforms explicitly remove libnsl from libswanted in their hints files so we can be certain those don't need to link libnsl.

I would guess that in the days before dynamic loading and Name Service Switch were common, this was the only way to make  `getpwent` and related functions use NIS on such platforms. Odds are it doesn't break anything released in the past 20+ years, but we'll only know when we try. I wouldn't worry too much about it.

danielnachun:
Okay.  Since I'm in a first time contributor approval is needed to run the workflows and we can see if anything actually breaks.

danielnachun:
Apologies I will have to get approval for the workflow to run again.  If there are other basic steps aside from updating the AUTHORS file I am happy to fix those before having this run again.

danielnachun:
All tests here have passed, so it looks like this didn't break anything that is tested in CI.

khwilliamson:
I'm afraid this appears to break Solaris builds.  (I didn't bisect)

Undefined			first referenced
 symbol  			    in file
h_errno                             pp_sys.o  (symbol belongs to implicit dependency /lib/amd64/libnsl.so.1)
ld: fatal: symbol referencing errors. No output written to miniperl

Leont:
Could you try it with this patch?
```
--- hints/solaris_2.sh
+++ hints/solaris_2.sh
@@ -63,6 +63,10 @@ esac
 set `echo " $libswanted " | sed -e 's@ ld @ @' -e 's@ malloc @ @' -e 's@ ucb @ @' -e 's@ sec @ @' -e 's@ crypt @ @'`
 libswanted="$*"
 
+# Add libnsl for networking support
+set `echo " $libswanted " | sed -e 's@ inet @ inet nsl @'`
+libswanted="$*"
+
 # Look for architecture name.  We want to suggest a useful default.
 case "$archname" in
 '')
```

khwilliamson:
I'm afraid I misled you.  I don't have access to such a machine.  I was just reporting a smoke report http://perl.develop-help.com/?b=smoke-me%2Fkhw-alpine

I suppose you or I could create a smoke-me branch, but I'd be just the middleman, so I think it better if you do.

Leont:
Pushed as `smoke-me/leont/nsl-solaris`

bram-perl:
> Pushed as `smoke-me/leont/nsl-solaris`

That looks to be working;
On blead: `make miniperl` fails
On your smoke-me branch: `make miniperl` succeeds (and so does `make` and `make test_harness`)

Leont:
> Pushed as smoke-me/leont/nsl-solaris

Merged

PR #20219, opened 2022-09-02
Update Hash::Util synopsis

rwp0:
Add a single `my` into hash definition line.

### DIFF
--- a/ext/Hash-Util/lib/Hash/Util.pm
+++ b/ext/Hash-Util/lib/Hash/Util.pm
@@ -42,7 +42,7 @@ our @EXPORT_OK  = qw(
 BEGIN {
     # make sure all our XS routines are available early so their prototypes
     # are correctly applied in the following code.
-    our $VERSION = '0.29';
+    our $VERSION = '0.30';
     require XSLoader;
     XSLoader::load();
 }
@@ -92,7 +92,7 @@ Hash::Util - A selection of general-utility hash subroutines
                      hash_traversal_mask
                    );
 
-  %hash = (foo => 42, bar => 23);
+  my %hash = (foo => 42, bar => 23);
   # Ways to restrict a hash
   lock_keys(%hash);
   lock_keys(%hash, @keyset);


demerphq:
Normally we leave my out of example code, as it obstructs from the core point being demonstrated. In this case I guess its pretty minor, but i think there is some value in having consistency in this area.

rwp0:
@demerphq

I usually copy synopsis code into my editor and start modifying (customizing) it when working with Perl modules so for me at least it's useful not to forget scoping variables, and as mentioned many Perl modules have `my` in their synopses, except few core ones which I aim to correct for documentation consistency.

demerphq:
On Fri, 2 Sept 2022 at 17:14, Elvin Aslanov ***@***.***>
wrote:

> @demerphq <https://github.com/demerphq>
>
> I usually copy synopsis code into my editor and start modifying
> (customizing) it when working with Perl modules so for me at least it's
> useful not to forget scoping variables, and as mentioned many Perl modules
> have my in their synopses, except few core ones which I aim to correct
> for documentation consistency.
>

Ok, but don't be surprised if people close the tickets as "won't fix".
Last time it came up there was consensus the other way on this point.

Yves
-- 

Grinnz:
Synopses and examples should aim to present modern code practices which now certainly includes declaring variables with my.

demerphq:
On Fri, 2 Sept 2022 at 17:22, Dan Book ***@***.***> wrote:
>
> Synopses and examples should aim to present modern code practices which now certainly includes declaring variables with my.

Is that written down somewhere?

-- 

rwp0:
> On Fri, 2 Sept 2022 at 17:22, Dan Book ***@***.***> wrote: Synopses and examples should aim to present modern code practices which now certainly includes declaring variables with my.
> Is that written down somewhere?
> […](#)
> -- perl -Mre=debug -e "/just|another|perl|hacker/"

The [Synopsis](https://perldoc.perl.org/perlpodstyle#SYNOPSIS) section in [perlpodstyle](https://perldoc.perl.org/perlpodstyle) has a relevant description:

> #### **SYNOPSIS**
> A short usage summary for **programs** and functions. This section is mandatory for section 3 pages. For Perl module documentation, it's usually convenient to have the contents of this section be a **verbatim block** showing some (brief) examples of **typical ways the module is used**.

I think by _verbatim block_ and _typical ways_ we mean the exact copy, and by scoping 🙂

Grinnz:
> On Fri, 2 Sept 2022 at 17:22, Dan Book ***@***.***> wrote: Synopses and examples should aim to present modern code practices which now certainly includes declaring variables with my.
> Is that written down somewhere?
> […](#)
> -- perl -Mre=debug -e "/just|another|perl|hacker/"

https://perldoc.perl.org/perldocstyle#Lead-with-common-cases-and-best-practices

demerphq:
> I think by verbatim block and typical ways we mean the exact copy and by scoping 🙂

Verbatim block is a POD technical term. It means a block of text that
is indented and is meant to be output with no processing, and visually
identical to what was indented, hence the term "verbatim".  Compare
with paragraph mode which is meant to be wrapped and formatted and
where POD directives like B<bold> and C<code> are to be respected. So
I dont think you can rely on that unfortunately.

Dan posted a link to this:

   https://perldoc.perl.org/perldocstyle#Lead-with-common-cases-and-best-practices

but I dont see anything that settles the matter. The text says to
document the better way of doing things, using my variables declared
line by line or in a block above their assignment is an equivalently
acceptable way to do things. I dont mind if the policy has changed
(Although I think it will have unfortunate effects on the readability
of some of our examples), but I would like to see it clearly
expressed, and I am not aware of anything that changes that status
quo.

Anyway, there are a myriad of other committers who can overrule me.
This is just my $0.02.

cheers,
Yves

-- 

PR #20220, opened 2022-09-02
Update POSIX Synopsis

rwp0:
Add `my` to examples.

Not sure if POSIX.pm's version should be incremented since a separate file.


jkeenan:
Probably need rebase on blead to get `make test_porting` to pass.  Otherwise looks good.

Grinnz:
It's in ext/ so it's not released as a separate distribution on CPAN, but ext/ is still organized into distributions, so I do believe a version bump across the distribution is still warranted when updating the documentation.

rwp0:
Rebased, `POSIX.pm` module version bumped.

PR #20224, opened 2022-09-02
Update Data::Dumper synopsis

rwp0:
Add `my` to synopsis.


rwp0:
Any feedback on this PR?
Looks like all is good, the two test failures are on macOS and seem unrelated.

rwp0:
Is Dumper developed separately from Perl core?

jkeenan:
On 9/24/22 21:07, Elvin Aslanov wrote:
> Is Dumper developed separately from Perl core?
> 

In the Perl core distribution, you find Data-Dumper under dist/.  That 
means it's "dual life, but blead upstream."  That is, the core 
distribution's blead (main) branch is the place development should occur 
first.  Then, periodically (or perhaps intermittently), releases are 
made to CPAN so that older perls can obtain the benefit of improvements 
in core.

jkeenan:
I updated $VERSION in the POD, then locally merged and pushed in a0a15662a9.  Closing ticket.

PR #20225, opened 2022-09-02
Update Carp.pm

rwp0:
Add `my` to synopsis.


Grinnz:
All versions in the Carp dist should be updated together

rwp0:
@Grinnz thanks, updating `Carp::Heavy` as well, since that's the only other module left in the distribution.

rwp0:
Increased `Carp::Heavy` module's version as well by rebasing.

PR #20227, opened 2022-09-02
Start cleaning up mutexes for locale and functions that access global memory 

khwilliamson:
This series of commit first cleans up some locale initialization gotchas, and adds asserts.

And it cleans up where a few locale-related definitions are done

It converts the locale mutex to a general semaphore from a binary one, which allows for simplification in switching the LC_NUMERIC locale.

And certain mutex macros now use the newly-general one


khwilliamson:
Merged via 3df28905fbf380b6a2ae5645d4129a69ae0d296a

PR #20230, opened 2022-09-02
sv.c - add BODYLESS_IV/NV fast code to sv_setXv functions

richardleach:
For "BODYLESS" IVs and NVs, type up/downgrades can be easily done
without having to call `sv_upgrade`. (Which will also not convert a
BODYLESS_NV back down to an IV, even when it is harmless to do so.)

This commit adds "BODYLESS" handling to `sv_setiv` and `sv_setnv`, which
previously lacked it, and "BODYLESS_NV" handling to `sv_setsv_flags`,
which previously only special-cased "BODYLESS_IV"s.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -1435,10 +1435,23 @@ Perl_sv_setiv(pTHX_ SV *const sv, const IV i)
 
     SV_CHECK_THINKFIRST_COW_DROP(sv);
     switch (SvTYPE(sv)) {
+#if NVSIZE <= IVSIZE
     case SVt_NULL:
+    case SVt_NV:
+        SET_SVANY_FOR_BODYLESS_IV(sv);
+        SvFLAGS(sv) &= ~SVTYPEMASK;
+        SvFLAGS(sv) |= SVt_IV;
+        break;
+#else
+    case SVt_NULL:
+        SET_SVANY_FOR_BODYLESS_IV(sv);
+        SvFLAGS(sv) &= ~SVTYPEMASK;
+        SvFLAGS(sv) |= SVt_IV;
+        break;
     case SVt_NV:
         sv_upgrade(sv, SVt_IV);
         break;
+#endif
     case SVt_PV:
         sv_upgrade(sv, SVt_PVIV);
         break;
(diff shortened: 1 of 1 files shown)


demerphq (approved):

richardleach:
Before I merge this, anyone got any thoughts on how doing BODYLESS_NVs in `Perl_sv_setsv_flags` requires preservation of `SVs_TEMP`, but the BODYLESS_IV code doesn't give a hoot? 

Is it just because coincidentally we have a test for a temp SV that gets assigned an NV and not an equivalent for IVs?

And should the new code in `Perl_sv_setiv` & `Perl_sv_setnv` be made to preserve `SVs_TEMP`?

richardleach:
I looked again and think the BODYLESS_IV code in `Perl_sv_setsv_flags` _should_ do `SvOK_off`, rather than clearing **all** the existing flags. I'll modify this PR so that happens.

`Perl_sv_setiv` & `Perl_sv_setnv` are fine as-is.

richardleach:
I've done the above - make `sv_setsv_flags` preserve flags like `SVs_TEMP` in the fast cases. @demerphq  - do you want another look over before I merge?

For posterity, what breaks in the BODYLESS_NV case if `SVs_TEMP` is not preserved:
```
            /* If SVs_TEMP is not preserved, t/lib/warnings/util has a
             * failure because of:
             *    perl: op.c:4915: OP *S_fold_constants(OP *const):
             *        Assertion `SvIMMORTAL(sv)' failed.
```
(I haven't traced through to find out what SV(s) result in that failure.)

richardleach:
@demerphq - do you want a final re-look through this before I merge?

demerphq:
On Sat, 17 Sept 2022, 23:16 Richard Leach, ***@***.***> wrote:

> @demerphq <https://github.com/demerphq> - do you want a final re-look
> through this before I merge?
>

Looks ok to me at the eyeball level, but I can't test till Thursday.  I'm
fine if you merge it without waiting.

Yves

>

richardleach:

richardleach:
No further comments received on this one since Yves said it looked okay, so am merging.

demerphq:
Thanks Richard!

Issue #18965, opened 2021-07-06
The readme file is not up to the standard [doc] 


wphtech:
**Where**

https://github.com/Perl/perl5/blob/blead/README

**Description**

I think it should be redone with proper markdown styling since it is the face of the perl5 language. Please take a look at

https://github.com/rakudo/rakudo

https://github.com/rust-lang/rust

https://github.com/golang/go

Grinnz:
Duplicate of https://github.com/Perl/perl5/issues/17227

PR #20232, opened 2022-09-03
Add a .github/README.md

bram-perl:
This is based on the existing README and is an almost exact copy.
Some minor formatting changes to improve the rendering but the text
remains the same.

t/porting/copyright.t was also adjusted to check the copyright in
.github/README.md

Note: this can be considered a (very) minor first step in adding a more
      informative README for Github. This first step is needed because
      not having a formatted README is blocking some other changes (rendering
      the platform specific READMEs as POD in GitHub).

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -1,0 +1,128 @@
+```
+Perl is Copyright (C) 1993, 1994, 1995, 1996, 1997, 1998, 1999, 2000,
+2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012,
+2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022
+by Larry Wall and others.
+All rights reserved.
+```
+
+# ABOUT PERL
+
+Perl is a general-purpose programming language originally developed for
+text manipulation and now used for a wide range of tasks including
+system administration, web development, network programming, GUI
+development, and more.
+
+The language is intended to be practical (easy to use, efficient,
+complete) rather than beautiful (tiny, elegant, minimal).  Its major
+features are that it's easy to use, supports both procedural and
+object-oriented (OO) programming, has powerful built-in support for text
+processing, and has one of the world's most impressive collections of
+third-party modules.
+
+For an introduction to the language's features, see pod/perlintro.pod.
+
+For a discussion of the important changes in this release, see
+pod/perldelta.pod.
+
+There are also many Perl books available, covering a wide variety of topics,
+from various publishers.  See pod/perlbook.pod for more information.
+
+
+# INSTALLATION
+
+If you're using a relatively modern operating system and want to
+install this version of Perl locally, run the following commands:
+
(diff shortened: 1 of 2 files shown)


demerphq:
I would love to see this done by a script, with a test that when the read README is updated that we know the .md version has too.

demerphq:
Also we have tools to check this stuff in the regen/ tree. It feels to me like that would be a good fit, it already has a mechanism to hash the input files to determine if the output files need to be changed, and if we updated the .md files on make regen it wouldnt be a bad thing.

bram-perl:
> I would love to see this done by a script, with a test that when the read README is updated that we know the .md version has too.

Short answer: No.

Long answer:
The README file isn't really changing; the only thing that changes in it is the copyright year and there is a test to cover that.
Eventually I would also like the .github/README.md to be much more informative and much more relevant for people viewing the repo on GitHub, if you compare the perl5 repo with other repo's then it clear that the current thing is not very user-friendly/doesn't offer a lot of relevant information. (See #18965 for some examples)
So in the long run I expect that README and .github/README.md will differ since they serve a different purpose.
The README file is not suitable for that since that's what included in the releases and that file shouldn't change. For including it in a release the current README is good enough, for showing it on GitHub it - in my opinion - isn't.

demerphq:
On Sat, 3 Sept 2022 at 10:24, Bram ***@***.***> wrote:

> I would love to see this done by a script, with a test that when the read
> README is updated that we know the .md version has too.
>
> Short answer: No.
>
> Long answer:
> The README file isn't really changing; the only thing that changes in it
> is the copyright year and there is a test to cover that.
> Eventually I would also like the .github/README.md to be much more
> informative and much more relevant for people viewing the repo on GitHub,
> if you compare the perl5 repo with other repo's then it clear that the
> current thing is not very user-friendly/doesn't offer a lot of relevant
> information. (See #18965 <https://github.com/Perl/perl5/issues/18965> for
> some examples)
> So in the long run I expect that README and .github/README.md will differ
> since they serve a different purpose.
> The README file is not suitable for that since that's what included in the
> releases and that file shouldn't change. For including it in a release the
> current README is good enough, for showing it on GitHub it - in my opinion
> - isn't.
>

I think if you put this explanation in the commit message it would be
helpful to future people who encounter this should you somehow get diverted
in reaching the conclusion of your plan, which I support.  So if you put a
curated version of this paragraph in the commit and I'd be happy to merge
it for you.

Yves

-- 

demerphq (approved):

rwp0:
Excellent, it was long awaited, although I'd like the titles (headings) not to be all uppercase.

I had a similar PR: https://github.com/Perl/perl5/pull/19099

Also, having `README.pod` instead would have been better I think (we could do that in the future).

Markdown is good, but we already have POD, so why not using that if GitHub supports rendering POD as good as Markdown?

Grinnz:
I think using pod format would be a good idea particularly so we can add documentation links to it easily in the future. Otherwise I think this is great.

PR #20235, opened 2022-09-03
Move GitHub files into ".github" directory

rwp0:
These might not need to go into [CPAN's Perl release](https://metacpan.org/dist/perl/source/pod).

Keeping all GitHub files in a single place is good I think.

I'd also like to convert the two Markdown files to POD using [Pod::Markdown](https://metacpan.org/pod/Pod::Markdown) but that's to be done in the future best.

Removed these from `MANIFEST` as well.

`MANIFEST` used to say:
- `SECURITY.md`:			Add Security Policy for **GitHub**
- `CODE_OF_CONDUCT.md`: Information on where to find the Standards of Conduct

---

There's also [`.travis.yml`](https://github.com/Perl/perl5/blob/blead/.travis.yml) file, not sure if that can also go into `.github` directory.

### DIFF
--- a/t/porting/regen.t
+++ b/t/porting/regen.t
@@ -109,8 +109,8 @@ foreach (@progs) {
     note("./perl $args");
     my $command = "$^X $args";
     system $command
-        and die <<~"HINT";
+        and die <<~'HINT';
     Hint:  A failure in this file can often be corrected by running:
-     ./perl -Ilib $_
+     make regen
 HINT
 }


bram-perl:
Hmm.. I don't think I'm fully in favor for this..

I think SECURITY.md should stay in the root and should be included in the release.
It's a lot easier to find then pod/perlsecpolicy.pod.
(Having a SECURITY.md file in the release might also be more uniform with other software (I'm guessing here tho, I did not check))


For CODE_OF_CONDUCT.md: I don't care where that ends up.

rwp0:
I think having all Perl documentation as POD instead of Markdown, and GitHub-specific files in `.github` directory is more uniform for Perl.

We can link to `perlsecpolicy`, `perlhack`, `perlgov` etc in README or somewhere else for people to see.

Grinnz:
> There's also [`.travis.yml`](https://github.com/Perl/perl5/blob/blead/.travis.yml) file, not sure if that can also go into `.github` directory.

.travis.yml is no longer relevant as github actions are used instead. It could be removed but that's not relevant to this PR.

Issue #20228, opened 2022-09-02
make test_porting does not flag need to update t/porting/customized.dat


jkeenan:
(This issue stems from an [observation by Tony Cook](https://github.com/Perl/perl5/pull/20216#discussion_r961317924) in his review of https://github.com/Perl/perl5/pull/20216.)

Suppose I build blead as I did today at 741a5c7396a0ca90a22ea8d8e0761c70c14b0a77, and then run `make test_porting`.  All tests will PASS, including, in particular, `t/porting/customized.t`.
```
porting/customized.t ...... ok
```

Now, in the absence of any *deep* study of that test program, I would assume that that means that the data in `t/porting/customized.dat` is up-to-date -- which in turn means that all the `CUSTOMIZED` entries in `Porting/Maintainers.pl` are up-to-date and that I do *not* need to regenerate that database with this command:
```
$ cd t; ./perl -I../lib porting/customized.t --regen; cd -
```
... which appears at the top of `t/porting/customized.dat`.

That assumption turns out to be incorrect.  When I run the command above with the `--regen` switch, I get:
```
ok 1 - regenerated data file
1..1
...
$ git status
On branch blead
Your branch is up to date with 'origin/blead'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   t/porting/customized.dat

no changes added to commit (use "git add" and/or "git commit -a")

$ git diff |cat
diff --git a/t/porting/customized.dat b/t/porting/customized.dat
index dae5af2446..0c2c70e25a 100644
--- a/t/porting/customized.dat
+++ b/t/porting/customized.dat
@@ -4,17 +4,11 @@
 AutoLoader cpan/AutoLoader/t/02AutoSplit.t bb90cda13b88599ad45de4b45799d5218afcb6d8
 ExtUtils::Constant cpan/ExtUtils-Constant/lib/ExtUtils/Constant/Base.pm 7560e1018f806db5689dee78728ccb8374aea741
 ExtUtils::Constant cpan/ExtUtils-Constant/t/Constant.t 165e9c7132b003fd192d32a737b0f51f9ba4999e
-Filter::Util::Call pod/perlfilter.pod 545265af2f45741a0e59eecdd0cfc0c9e490c1e8
[...]

PR #20237, opened 2022-09-03
perlfilter.pod: Retain entry in customized database

jkeenan:
`pod/perlfilter.pod` is (for historical reasons, apparently) maintained
upstream on CPAN as part of the Filter distribution.  As such, (i) it
should, generally speaking, only be edited upstream and synched into
core; (ii) needs to be referenced in several places in the
`Filter::Util::Call` element in `%Maintainers::Modules`
(`Porting/Maintainers.pl`); (iii) needs to have an entry in
`t/porting/customized.dat` (the "customized database") to protect it from
being "overwritten" or accidentally removed.

For all of the above conditions to work together, the
`filter_customized()` subroutine in `t/porting/customized.t` needs to be
able to construct a regex which recognizes `pod/perlfilter.pod`'s
weirdness.  Prior to this patch it was failing to do so.  Thus when
`pod/perlfilter.pod` was accidentally edited in commit d84bd0bd47, the
need to regenerate the customized database was not recognized.  Later,
review of a different pull request which required regeneration of the
customized database spotted the unintentional deletion of the entry for
`pod/perlfilter.pod` from `t/porting/customized.dat`.

Addressing this problem is inevitably going to be kludgy, but it will be
less kludgy if done in `t/porting/customized.t` than in
`Porting/Maintainers.pl`.  This patch modifies the former file to
guarantee that the string 'pod/perlfilter.pod' is matched.  After that
modification, `cd t; ./perl -I../lib porting/customized.t --regen; cd -`
was run.  This accomplished the (previously overlooked) removal of
certain entries for `Memoize` in the customized database while leaving the
entry for `pod/perlfilter.pod` in that database intact.

If at some point in the future we decide that `pod/perlfilter.pod` should
be maintained directly in core, we can remove this kludge and simplify
the `Filter::Util::Call` element in `%Maintainers::Modules` in
`Porting/Maintainers.pl`.

For: https://github.com/Perl/perl5/issues/20228.

Should (partially) unblock https://github.com/Perl/perl5/pull/20216.

### DIFF
--- a/t/porting/customized.dat
+++ b/t/porting/customized.dat
@@ -10,11 +10,6 @@ Math::Complex cpan/Math-Complex/lib/Math/Complex.pm 66f28a17647e2de166909ca66e4c
 Math::Complex cpan/Math-Complex/t/Complex.t 17039e03ee798539e770ea9a0d19a99364278306
 Math::Complex cpan/Math-Complex/t/Trig.t 508f8e27373c08228be13ca5d42b28812ab0e020
 Math::Complex cpan/Math-Complex/t/underbar.t 97e7b9615658eefc67a710d4b258349cc5bace63
-Memoize cpan/Memoize/Memoize.pm 902092ff91cdec9c7b4bd06202eb179e1ce26ca2
-Memoize cpan/Memoize/t/errors.t bcd0c02a4bc47dfda07a97d265d7210849871659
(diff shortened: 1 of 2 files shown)


jkeenan:
Can either the requested reviewers or anyone else take a look at this pull request?  Thanks.

khwilliamson (approved):

tonycoz (approved):

PR #20239, opened 2022-09-03
Improve rendering of platform specific READMEs on GitHub

bram-perl:
On GitHub the platform specific README files (README.win32, ...) were rendered as plain text even tho there are using the POD format.

Fix this so that GitHub now renders these as POD which makes them easier to read.


rwp0:
Excellent, may consider adding Emacs modeline as well. That's also recognized by GitHub but also mainly by Emacs editor.

README.ko
```diff
@@ -1,9 +1,10 @@
-=encoding utf8
+# vim: syntax=pod
 
 이 파일을 내용 그대로 읽고 있다면 우스꽝스러운 문자는 무시해주세요.
 이 문서는 POD로 읽을 수 있도록 POD 형식(F<pod/perlpod.pod> 문서를
 확인하세요)으로 작성되어 있습니다.
 
+=encoding utf8
```
Leont, line 7: Why did you move this down? I think that's incorrect.

bram-perl: It's in the commit message of  https://github.com/Perl/perl5/pull/20239/commits/24a9eff1532e63a82acdf738d0286113e518d479
```
 Move '=encoding utf8' after preamble

All the READMEs start with a preamble describing that the file is
written in the POD format.

When the '=encoding' line is put before the preamble then the
preamble is also rendered when viewing the file with `perldoc`
(and/or when viewing the file as POD).

Fix: move '=encoding' to the place where the actual POD starts.
```

I did test this with:
- perldoc
- podchecker
- viewing it on GitHub

bram-perl: You can also view it on https://github.com/bram-perl/perl5/blob/render-as-pod/README.ko
And while I don't speak Korean it does look correct (and copy/paste to google translate is able to detect the language as Korean and translate it..)

demerphq (approved):

rwp0 (approved):

README.ko
demerphq: @Leont you still have reservations on this? It makes sense to me. I will merge it if i dont hear from you soon to the contrary.

Leont: No, it all makes sense to me now

Leont (approved):

PR #20243, opened 2022-09-03
todo.pod: Remove version number from todo item

bram-perl:
Porting/todo.pod contains one todo item that mentions a version number.
That particular version number has already been bumped 90 times on blead...

I think it's time to stop bumping it and replace it with a generic text.
(It started as v5.14.0, then got bumped to v5.18.0 and then got bumped
 with every release...)


khwilliamson:
Is the bumping done automatically?  If so, we need to stop that

PR #20244, opened 2022-09-04
Delete .travis.yml

rwp0:
This file seems to be no longer needed since switching to GitHub Actions for CI.

See also the discussion at https://github.com/Perl/perl5/pull/20235#issuecomment-1236168934


rwp0:
Removed the file from the `MANIFEST` listing as well.

haarg (approved):

PR #20247, opened 2022-09-04
Yves/simple header cleanups

demerphq:
This PR fixes a handful of minor formatting nits in our header files. It should show no changes under diff -w.

### DIFF
--- a/intrpvar.h
+++ b/intrpvar.h
@@ -1067,7 +1067,8 @@ PERLVAR(I, wcrtomb_ps, mbstate_t)
 /* Enough space for the reserved byte, 1 for a potential leading 0, then enough
  * for the longest representable integer plus an extra, the 3 flag characters,
  * and NUL */
-PERLVARA(I, mem_log, 1 + 1 + TYPE_DIGITS(UV) + 1 + 3 + 1, char)
+#define PERL_MEM_LOG_ARYLEN (1 + 1 + TYPE_DIGITS(UV) + 1 + 3 + 1)
+PERLVARA(I, mem_log, PERL_MEM_LOG_ARYLEN,  char)
 #endif
 
 /* The most recently seen `use VERSION` declaration, encoded in a single
--- a/intrpvar.h
+++ b/intrpvar.h
@@ -950,7 +950,7 @@ PERLVAR(I, Xpv,		XPV *)		/* (unused) held temporary value */
 
 /* name of the scopes we've ENTERed. Only used with -DDEBUGGING, but needs to be
    present always, as -DDEBUGGING must be binary compatible with non.  */
-PERLVARI(I, scopestack_name, const char * *, NULL)
+PERLVARI(I, scopestack_name, const char **, NULL)
 
 PERLVAR(I, debug_pad,	struct perl_debug_pad)	/* always needed because of the re extension */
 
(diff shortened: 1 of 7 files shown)


intrpvar.h
```diff
@@ -1067,7 +1067,8 @@ PERLVAR(I, wcrtomb_ps, mbstate_t)
 /* Enough space for the reserved byte, 1 for a potential leading 0, then enough
  * for the longest representable integer plus an extra, the 3 flag characters,
  * and NUL */
-PERLVARA(I, mem_log, 1 + 1 + TYPE_DIGITS(UV) + 1 + 3 + 1, char)
+#define PERL_MEM_LOG_ARYLEN (1 + 1 + TYPE_DIGITS(UV) + 1 + 3 + 1)
```
jkeenan, line 1070: Is `PERL_MEM_LOG_ARYLEN` going to be used anyplace other than in the line immediately following its define?  If not, what is the rationale for this change?

perlvars.h
```diff
@@ -265,7 +265,8 @@ PERLVARA(G, hash_seed_w, PERL_HASH_SEED_WORDS, PVT__PERL_HASH_WORD_TYPE) /* perl
 PERLVARA(G, hash_state_w, PERL_HASH_STATE_WORDS, PVT__PERL_HASH_WORD_TYPE) /* perl.c and hv.h */
 #endif
 #if defined(PERL_USE_SINGLE_CHAR_HASH_CACHE)
-PERLVARA(G, hash_chars, (1+256) * sizeof(U32), unsigned char) /* perl.c and hv.h */
+#define PERL_SINGLE_CHAR_HASH_CACHE_ELEMS ((1+256) * sizeof(U32))
+PERLVARA(G, hash_chars, PERL_SINGLE_CHAR_HASH_CACHE_ELEMS, unsigned char) /* perl.c and hv.h */
 #endif
```
jkeenan, line 270: Similar to the above question ... Is `PERL_SINGLE_CHAR_HASH_CACHE_ELEMS` going to be used anyplace other than in the line immediately following its define? If not, what is the rationale for this change?

jkeenan (commented):
Other than those two questions, looks good.

intrpvar.h
Leont: Yeah, the benefit over `sizeof(PL_mem_log)` isn't quite clear to me either

demerphq: @leont - i just wanted to get the complex expression out of the PERLVARA() declaration. I dont care what the complex expression is really, and I dont think changing it is in scope of this patch.
@jkeenan - no, but that doesn't mean we shouldnt make it available.

Ultimately this is just a prep patch that makes it easier to do futher cleanups on this file. It shouldn't matter nor hurt anything to do so.

perlvars.h
demerphq: I thought I did explain in the commit message: it makes it easier to parse the file. The secondary answer is that it is always better to hide a complex constant expression behind a named identifier, it adds comprehensibility and it allows the complex expression to be reused. 

Indeed, I could easily reuse this define in the hash code, I am fairly sure I have duplicated this logic twice. 

BTW, check the blame log for added this line of code in the first place. Arguably I should have used a define in the first place! ;-)

demerphq:
BTW, general comment, this is part of a series of work i am doing to clean up our header files. Some of the patches here have been made to make it easier to follow up changes.  Especially the PERLVAR ones, which make it easier to parse the file which in turn allows us to do other things, like clean up the indentation of the file.

intrpvar.h
demerphq: So @Leont and @jkeenan anything further on this? Can i resolve this?

perlvars.h
demerphq: @jkeenan ping.

intrpvar.h
jkeenan: If @Leont is satisfied, then I have no further question.

perlvars.h
jkeenan: As above.

intrpvar.h
Leont: I'd vote -0, so go ahead.

PR #20253, opened 2022-09-05
Fix POSIX::strxfrm()

khwilliamson:
This function takes an SV containing a PV.  The encoding of that PV is
based on the locale of the LC_CTYPE locale.  It really doesn't make
sense to collate based off of the sequencing of a different locale, which
prior to this commit it would do if the LC_COLLATION locale were
different.


demerphq (approved):

bram-perl:
Is it possible to add test case for this?

khwilliamson:
I don't see a good way for such a test case.  The results of strxfrm() are opaque, varying by platform and potentially by release.  And finding a locale that was sufficiently different from another locale would be hard to do portably.  The function now is just a wrapper around the function that is used all the time by perl code, and so is tied to that working or not.  That will be tested more in a later commit that runs a bunch of locales in parallel threads

bram-perl:
If a test case is not possible, is an example - that could be included in the commit message - possible?
That is: can you add a - non portable - example in the commit message where the output of `POSIX::strxfrm` is different before/after the commit?

khwilliamson:
@bram-perl  the commit message has now been rewritten and includes an example from glibc.

bram-perl (approved):

PR #20254, opened 2022-09-05
locale.c: Initialize variable to avoid runtime check; also PL_underlying_numeric_ob

khwilliamson:
Initializing the first means we no longer have to tess that it is non-NULL
each time it is accessed; the second is just better form.


demerphq (approved):

demerphq:
Huh, code looked fine, but obviously its failing tests...

PR #20256, opened 2022-09-05
locale.c More new_ctype() refactoring

khwilliamson:
Merge commit e4bbbfe02b9e9aae521b164eba0e518ca478945f refactored this
function some.  Most of the commits in that series dated to before when
we could assume C99.  In re-reading the result, I saw some opportunities
to take advantage of C99, by, for example, moving declarations closer to
their use.

I also hadn't previously noticed that when changing to the C locale (a
frequent occurrence), various things that we being recalculated are
determinable at compile time.  So this commit returns early under this
circumstance.

And, an obsolete comment is removed

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -1784,6 +1784,7 @@ Perl_set_numeric_underlying(pTHX)
 STATIC void
 S_new_ctype(pTHX_ const char *newctype)
 {
+    PERL_ARGS_ASSERT_NEW_CTYPE;
 
 #  ifndef USE_LOCALE_CTYPE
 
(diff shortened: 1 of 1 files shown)


jkeenan:
Currently failing a porting test.
```
$ cd t;./perl harness porting/cmp_version.t; cd -
porting/cmp_version.t .. 1/28 # not ok 21 - ext/XS-APItest/APItest.pm version 1.25
# not ok 22 - lib/B/Deparse.pm version 1.68
porting/cmp_version.t .. Failed 2/28 subtests 

Test Summary Report
-------------------
porting/cmp_version.t (Wstat: 0 Tests: 28 Failed: 2)
  Failed tests:  21-22
```
Probably needs to be rebased on blead.

bram-perl (approved):

PR #20257, opened 2022-09-05
peep.c:OP_PADSV_STORE - skip SASSIGN created by Memoize::Once

richardleach:
Memoize::Once produces a non-standard SASSIGN which should not and
need not be optimized. Because this SASSIGN has only one child,
something that hasn't been seen elsewhere, that can be used as a
heuristic for skipping this type of SASSIGN in rpeep().


richardleach:
Fixes the Memoize::Once portion of #20114.

PR #20259, opened 2022-09-05
pp_ctl.c - move unitcheck in eval handler to a function similar to S_try_yyparse

demerphq:
This creates S_try_run_unitcheck(), which is similar to S_try_yyparse(),
in that it runs the subs in PL_unitcheckav() inside of a JMPENV_PUSH()
and JMPENV_POP wrapper so that any call to Perl_croak() inside of a
unitcheck block executed as part of eval compilation ends back in the
right place.

Both of these functions are private to pp_ctl.c and not listed in
embed.c and are used to trap the setjmp/longjmp calls and handle
exceptions during eval.

This fixes a cryptic warning about a variable being used across setjmp
in pp_ctl.c doeval_compile(), which turns out to be evalcv. By wrapping
the JMPENV_PUSH logic in a light wrapper these issues are moved one
layer down the stack and avoided.

### DIFF
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3547,6 +3547,32 @@ S_try_yyparse(pTHX_ int gramtype)
     return ret;
 }
 
+/* Run PL_unitcheckav in a setjmp wrapper via call_list.
+ * Returns:
+ *   0: unitcheck blocks ran without error
+ *   3: a unitcheck block died
+ */
+STATIC int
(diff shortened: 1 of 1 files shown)


demerphq:
Note the verbiage in the commit is slightly different and better than that of the PR itself.

bram-perl (approved):

Issue #20261, opened 2022-09-06
Failed eval in sort causes segfault


sisyphus-ppcg:
**Description**
An eval that results in a syntax error inside of a `sort` causes a segmentation fault. This reproduces on v5.36.0 and v5.30.0 but does not segfault on a very old version of Perl (v5.16.3) so it was introduced somewhere between those versions.

**Steps to Reproduce**
Run the following code:

```perl
@a = sort{eval"("}1,2
```

**Expected behavior**
No segfault.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 36 subversion 0) configuration:

  Platform:
    osname=linux
    osvers=5.10.0-13-cloud-amd64
    archname=x86_64-linux-gnu
    uname='linux 5918bdee5462 5.10.0-13-cloud-amd64 #1 smp debian 5.10.106-1 (2022-03-17) x86_64 gnulinux '
    config_args='-Darchname=x86_64-linux-gnu -Duse64bitall -Duseshrplib -Dvendorprefix=/usr/local -des'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2'
    cppflags='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='10.2.1 20210110'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    ldflags =' -fstack-protector-strong -L/usr/local/lib'
[...]

PR #20263, opened 2022-09-06
t/op/eval.t - Test for eval string inside of sort

demerphq:
Things that die in eval in sort have from time to time caused trouble.
This tests that dieing inside of a BEGIN or UNITCHECK inside of an eval
inside of sort does not segfault.

See discussion in GH Issue #20261.

### DIFF
--- a/t/op/eval.t
+++ b/t/op/eval.t
@@ -6,7 +6,7 @@ BEGIN {
     set_up_inc('../lib');
 }
 
-plan(tests => 149);
+plan(tests => 152);
 
 eval 'pass();';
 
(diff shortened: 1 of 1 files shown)


t/op/eval.t
```diff
@@ -725,5 +725,8 @@ pass("eval in freed package does not crash");
         'eval "BEGIN     { eval q(UNITCHECK { die; }); print q(A-) }";',
     ) {
         fresh_perl_is($line . ' print "ok";', "A-ok", {}, "No segfault: $line");
+        my $sort_line= 'my @x= sort { ' . $line . ' } 1,2;';
+        fresh_perl_is($line . ' print "ok";', "A-ok", {}, "No segfault: $line");
+
     }
```
jkeenan, line 730: Blank line 730 is superfluous; otherwise +1.

jkeenan (changes requested):

demerphq:
I accidentally merged this while fixing the broken cmp_version tests. I will have to remove the newline in a new PR.

Issue #20214, opened 2022-09-01
Win32: (gcc) build without `USE_MINGW_ANSI_STDIO` fail at t/op/sprintf2.t


bram-perl:
On Windows 10 - using with gcc/gmake - t/op/sprintf2.t fails when build is done  **without** `USE_MINGW_ANSI_STDIO`:[^1]
```
# Failed test 1704 - sprintf( "%.54g", 0.3 ) renders correctly at op/sprintf2.t line 1197
#      got "0.29999999999999999"
# expected eq "0.299999999999999988897769753748434595763683319091796875"
```

Looking at:
- #18170
- commit b14ae3c3b36461b9e396451434fba1dff9a69bb8 (by @sisyphus )

I'm guessing this failure is somewhat expected on builds not using `USE_MINGW_ANSI_STDIO`?
Can/should this test be skipped when build is done without `USE_MINGW_ANSI_STDIO`?
Is there an easy way to detect this in the test?

**Steps to reproduce**

1. Disable `USE_MINGWS_ANSI_STDIO` in win32/GNUmakefile, i.e.:
```
diff --git a/win32/GNUmakefile b/win32/GNUmakefile
index 81ded707b9..df5f417978 100644
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -142,7 +142,7 @@ USE_PERLIO  := define
 # (s)printf formatting of numbers, whereas the MS runtime might not.
 # This option has no effect on MSVC builds.
 #
-USE_MINGW_ANSI_STDIO := define
+#USE_MINGW_ANSI_STDIO := define

 #
 # Comment this out if you want the legacy default behavior of including '.' at
```
2. Build
3. Run t/op/sprintf2.t

**Result:**

Test failure:
```
C:\Perl\perl-current\t>.\perl.exe harness op/sprintf2.t
op/sprintf2.t .. 1437/? # Failed test 1704 - sprintf( "%.54g", 0.3 ) renders correctly at op/sprintf2.t line 1197
#      got "0.29999999999999999"
# expected eq "0.299999999999999988897769753748434595763683319091796875"
op/sprintf2.t .. Failed 1/1704 subtests
        (less 30 skipped subtests: 1673 okay)
```

**Expected result**

No test failure (/test skipped/...)

[^1]: I noticed this on my Windows 10 smoker due to a recent change in Test::Smoke which now builds without `USE_MINGW_ANSI_STDIO`.

sisyphus:
[...]

Issue #19495, opened 2022-03-05
MS Windows: %Config doesn't provide the mingw runtime version.


sisyphus:
**Description**

AFAICS, on Windows, there is currently no %Config key that stores the value of the mingw runtime version of the gcc toolchain that built perl.
Given that recent  mingw runtimes (from v7 to the current v9) are throwing up issues, I think it would be nice to have a %Config key that holds that version number of the mingw runtime.

Would it be ok to make use of the existing, but unused, <code>gnulibc_version</code> key ?
Or do we need to create a new key ? (If so ... ideas on what it should be called would be welcome.)

Further, I think it would be smart to have this key and its value displayed automatically by <code>perl -V</code> so that we can readily see this value in cpantester/smoker reports.

I'm quite happy for someone else to implement these features.
I'm also happy to have a crack at writing that implementation myself ... though some hints on how to get this key added to the default <code>perl -V</code> output would be handy.

For reference, here's the 'perl -V' output for my perl-5.34.0, gcc-10.2.0, mingw runtime version 8.0.0.
# perl -V output goes here
```
Summary of my perl5 (revision 5 version 34 subversion 0) configuration:

  Platform:
    osname=MSWin32
    osvers=6.1.7601
    archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -DWIN64 -fdiagnostics-color=never -DPERL_TEXTMODE_SCRIPTS
 -DPERL_IMPLICIT_CONTEXT -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STD
IO -fwrapv -fno-strict-aliasing -mms-bitfields'
    optimize='-s -O2'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='10.2.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long long'
    ivsize=8
    nvtype='double'
    nvsize=8
[...]

PR #20265, opened 2022-09-07
Add mingw runtime version to %Config

sisyphus:
This follows on from https://github.com/Perl/perl5/issues/19495 and https://github.com/Perl/perl5/issues/20214 .

Currently, the version of the mingw runtime which shipped with the mingw-w64 toolchain that built perl cannot be determined by examining %Config.
The only way I know of currently extracting that information is via an XS subroutine that returns that info.
This PR adds the mingw runtime version to %Config, as $Config{mingwrt_version}.

Passes all tests for me on Ubuntu and Windows (for both mingw-built and msvc-built perls).

However, the value displayed on Ubuntu is less than perfect.
The <code>perl -V</code> output displays <code>mingwrt_version='undef'</code>, whereas I would prefer to see either <code>mingwrt_version=undef</code> or <code>mingwrt_version=''</code>.
I don't know which is correct - I welcome feedback on that.

Also, on Ubuntu, I get:
```
$ perl -V:mingwrt_version
mingwrt_version='UNKNOWN';
```
I thought <code>'UNKNOWN'</code> was meant to signify that the specified hash key does not exist.
Some thoughts on how to tidy up that piece of inconsistency would also be welcome.

A couple of my commit messages aren't quite right - but rebasing and squashing should take care of that ... if we ever get to that stage.

Now, let's see what CI throws up


sisyphus:
Nope ... don't like this PR at all.
In a day or two, I'll present a much less ambitious effort that simply sets $Config{mingwrt_version} appropriately, when and only when the OS is Windows && the compiler is gcc.
I'll leave the more complex exercise of having that mingwrt_version info displayed by <code>perl -V</code> to a later date.

Cheers,
Rob

bram-perl:
> Nope ... don't like this PR at all.

For what it's worth: my first impression of the PR was that it wasn't too bad but I haven't had an opportunity yet to test it out. (And/or to think about some of the issues raised)

sisyphus:
One problem  (that I'll still face even if I do come up with the simplified PR I've mooted) is what to do about the documentation of $Config{mingwrt_version}.
There is no Config.pod that I can simply patch, because Config.pod is generated, along with Config.pm and Config_heavy.pl, by configpm.
And configpm draws on Porting/Glossary (a file that is not to be edited by hand), Configure, myconfig.SH and the generated config.sh.

It would be tempting to just leave $Config{mingwrt_version} as undocumented.

Anyway, I'll try to wade through all of that - and maybe ask for some guidance from p5p if I get stuck.

Update: I've just restored the mingwrt-version branch in case deleting it was the wrong thing to do.

PR #20267, opened 2022-09-07
Fix GNU/{Hurd,kFreeBSD,kNetBSD} hint files in several modules

ntyni:
The idiom of

  do './hints/linux.pl' or die $@;

broke with https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/367 which made $self a lexical variable so it is no longer visible in a file loaded with 'do'.

This silently makes the loaded Linux hints a no-op, leading to missing symbols, broken modules and test failures in at least NDBM_File and ODBM_File on GNU/Hurd, as reported by Samuel Thibault in https://bugs.debian.org/1018289 .

The replacement of eval `cat hints/linux.pl` is not very sophisticated but should be enough for this, and alternatives seem overly verbose.

Bug-Debian: https://bugs.debian.org/1018289

### DIFF
--- a/dist/Storable/hints/gnukfreebsd.pl
+++ b/dist/Storable/hints/gnukfreebsd.pl
@@ -1,1 +1,1 @@
-do './hints/linux.pl' or die $@;
+eval `cat hints/linux.pl` or die $@;
--- a/dist/Storable/hints/gnuknetbsd.pl
+++ b/dist/Storable/hints/gnuknetbsd.pl
@@ -1,1 +1,1 @@
-do './hints/linux.pl' or die $@;
+eval `cat hints/linux.pl` or die $@;
--- a/ext/DynaLoader/hints/gnukfreebsd.pl
+++ b/ext/DynaLoader/hints/gnukfreebsd.pl
@@ -1,1 +1,1 @@
-do './hints/linux.pl' or die $@;
+eval `cat hints/linux.pl` or die $@;
--- a/ext/DynaLoader/hints/gnuknetbsd.pl
+++ b/ext/DynaLoader/hints/gnuknetbsd.pl
@@ -1,1 +1,1 @@
-do './hints/linux.pl' or die $@;
+eval `cat hints/linux.pl` or die $@;
--- a/ext/NDBM_File/hints/gnu.pl
+++ b/ext/NDBM_File/hints/gnu.pl
@@ -1,1 +1,1 @@
-do './hints/linux.pl' or die $@;
+eval `cat hints/linux.pl` or die $@;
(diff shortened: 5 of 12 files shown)


ntyni:
Verified to fix the issue on GNU/Hurd. GNU/kFreeBSD and GNU/kNetBSD are dead AFAIK.

Leont:
To be honest, I'm wondering if we shouldn't revert that change in MakeMaker instead.

jkeenan:
> To be honest, I'm wondering if we shouldn't revert that change in MakeMaker instead.

In my experience, it can take a long time for changes in modules managed by PTC to reach the consensus needed for approval and new CPAN release.  The change in MakeMaker which triggered this problem was made in [October 2020](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/5490957e1b65009602369a08003b55f1df575b2a), so any discussion upstream about its reversion is likely to be involved and drawn-out.

I would not like to have the problems pointed out by @ntyni linger on.  His proposal is reasonable and we'll get relatively quick feedback via the Debian testing setup.

So I'm in favor of applying this merge request now.

Leont:
> In my experience, it can take a long time for changes in modules managed by PTC to reach the consensus needed for approval and new CPAN release. The change in MakeMaker which triggered this problem was made in [October 2020](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/commit/5490957e1b65009602369a08003b55f1df575b2a), so any discussion upstream about its reversion is likely to be involved and drawn-out.
>
> I would not like to have the problems pointed out by @ntyni linger on. His proposal is reasonable and we'll get relatively quick feedback via the Debian testing setup.

I'm quite sure ntyni tested it already. I don't like to add a workaround only to have to revert it not much later when the real fix has been committed. It's not like Debian is going to ship a development release, as long as we manage to ship a new MakeMaker in within 6 months we're fine.

hy-ntyni:
>  I don't like yo add a workaround only to have to revert it not much later when the real fix has been committed.

Thanks for looking at this. Fine by me of course. I can certainly add the workaround as a Debian patch for now and remove it later when a fixed EU::MM has trickled downstream to us.

Not sure if you're quite decided yet so not closing this PR, but feel free to do that at your convenience.

jkeenan:
> > I don't like to add a workaround only to have to revert it not much later when the real fix has been committed.
> 
> Thanks for looking at this. Fine by me of course. I can certainly add the workaround as a Debian patch for now and remove it later when a fixed EU::MM has trickled downstream to us.
> 
> Not sure if you're quite decided yet so not closing this PR, but feel free to do that at your convenience.

Okay, since @Leont has submitted a [patch](https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/423) upstream for ExtUtils::MakeMaker and since @ntyni has the option of patching Debian, we'll close this p.r. for now.

PR #20269, opened 2022-09-08
perlhacktips: Add a section on symbol naming

khwilliamson:


pod/perlhacktips.pod
```diff
@@ -309,6 +309,54 @@ C<-xc99>
...
+Indeed, the perl core has more than a few instances of using
+implementation-reserved symbols.  (These are gradually being changed.)
+But your code might stop working any time that the implementation
+decides to use a name you already had chosen, potentially many years
+before.  
+
+It's best then to:
+
+B<Don't begin a symbol name with an underscore, nor have two adjacent
+underscores>.
```
jkeenan: For better grammatical flow, I suggest:
```
B<Don't begin a symbol name with an underscore or with two adjacent underscores.>

**Comment by khwilliamson** [comment]
On 9/7/22 19:21, James E Keenan wrote:
> 
> 
> In pod/perlhacktips.pod 
> <https://github.com/Perl/perl5/pull/20269#discussion_r965420569>:
> 
>> +symbol beginning with an underscore, not just ones followed by a
> +capital.  We care about C++ because C<hdr> files need to be compilable by
> +it, and some people do all their development using a C++ compiler.
> +
> +The consequences of failing to do this are probably none.  Unless you
> +stumble on a name that the implementation uses, things will work.
> +Indeed, the perl core has more than a few instances of using
> +implementation-reserved symbols.  (These are gradually being changed.)
> +But your code might stop working any time that the implementation
> +decides to use a name you already had chosen, potentially many years
> +before.
> +
> +It's best then to:
> +
> +B<Don't begin a symbol name with an underscore, nor have two adjacent
> +underscores>.
> 
> For better grammatical flow, I suggest:
> 
> |B<Don't begin a symbol name with an underscore or with two adjacent 
> underscores.> |

That's wrong.  You can't have two adjacent underscores anywhere.

**Comment by bram-perl** [comment]
> That's wrong. You can't have two adjacent underscores anywhere.

I think it might be best to split into two lines then, that is something like:

```
 It's best then to:
- Don't begin a symbol name with an underscore; (i.e. don't use: _FOOBAR)
- Don't use two adjacent underscores (i.e. don't use: FOO__BAR)
```
(with or without the examples)


*or* to reverse the sentence a bit, i.e.:
```
Don't use two adjacent underscores and don't begin a symbol name with an underscore
```

Another change I would suggest: use `adjacent/consecutive underscores` (instead of `adjacent underscores`)

bram-perl (approved):

jkeenan (approved):

PR #20270, opened 2022-09-08
locale.c Use functions to initialize certain locale categories; save some conditionals

khwilliamson:
This series of commits includes #20256 as its requirement

The rest of the series initializes an object, and changes the initialization process to call the specialty functions for all the locale categories that them to do the initialization, instead of duplicating the logic in the initialization section.  This allows for the removal of a couple of conditionals that otherwise would need to be executed each time the locale changes.


bram-perl (approved):

khwilliamson:
Merged via c1fa8ffb1b03a17736a6f98a97f7283ca6feba28

PR #20271, opened 2022-09-08
Win32 stat() didn't handle AF_UNIX socket files

### LINKED ISSUE #20204: Win32: stat/unlink fails on UNIX sockets
`dist/IO/t/io_unix.t` contains a test that creates a UNIX socket, does some tests and then attempts to `unlink` the file.
That `unlink` call fails and that results in a stray tmp file.

A reduced example case:

```
#!./perl

use IO::Socket;

my $PATH = "sock-$$";

my $listen = IO::Socket::UNIX->new(Local => $PATH, Listen => 0);
close $listen;

my @x = stat($PATH) or warn "stat $PATH failed: $!";
unlink $PATH or die "unlink $PATH failed: $!";
```

Running on Windows 10:
```
$ .\perl -I..\lib foo.t
stat sock-6816 failed: Invalid argument at foo.t line 10.
unlink sock-6816 failed: Invalid argument at foo.t line 11.
```

See @tonycoz comment: https://github.com/Perl/perl5/pull/20179#issuecomment-1232810008
> I expect this needs to be fixed in the core code, at least in win32_stat() and maybe in Perl_apply() (which does unlink() amongst other ops).
>
> I'll be looking at win32_stat() tomorrow, but being windows, I expect it to be moderately ugly (see https://github.com/golang/go/issues/33357 for example).

Other references:
- #20179
- #17575 
- #17429

tonycoz:
Unfortunately both symbolic links and sockets can only be "statted" by opening with FILE_FLAG_OPEN_REPARSE_POINT which obviously doesn't follow symbolic links.

So to find if a chain of symbolic links points to a socket, is a broken chain, or loops, we need to follow the chain ourselves.

Fixes #20204

### DIFF
--- a/win32/win32.c
+++ b/win32/win32.c
@@ -1488,12 +1488,76 @@ translate_ft_to_time_t(FILETIME ft) {
 
 typedef DWORD (__stdcall *pGetFinalPathNameByHandleA_t)(HANDLE, LPSTR, DWORD, DWORD);
 
+/* Adapted from:
+
+https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_reparse_data_buffer
+
+Renamed to avoid conflicts, apparently some SDKs define this
+structure.
+
+Hoisted the symlink and mount point data into a new type to allow us
+to make a pointer to it, and to avoid C++ scoping issues.
+
+*/
+
+typedef struct {
+    USHORT SubstituteNameOffset;
+    USHORT SubstituteNameLength;
+    USHORT PrintNameOffset;
+    USHORT PrintNameLength;
+    ULONG  Flags;
+    WCHAR  PathBuffer[MAX_PATH*3];
+} MY_SYMLINK_REPARSE_BUFFER, *PMY_SYMLINK_REPARSE_BUFFER;
+
+typedef struct {
+    USHORT SubstituteNameOffset;
+    USHORT SubstituteNameLength;
+    USHORT PrintNameOffset;
+    USHORT PrintNameLength;
+    WCHAR  PathBuffer[MAX_PATH*3];
+} MY_MOUNT_POINT_REPARSE_BUFFER;
+
+typedef struct {
+  ULONG  ReparseTag;
+  USHORT ReparseDataLength;
+  USHORT Reserved;
+  union {
+    MY_SYMLINK_REPARSE_BUFFER SymbolicLinkReparseBuffer;
+    MY_MOUNT_POINT_REPARSE_BUFFER MountPointReparseBuffer;
+    struct {
+      UCHAR DataBuffer[1];
+    } GenericReparseBuffer;
+  } Data;
+} MY_REPARSE_DATA_BUFFER, *PMY_REPARSE_DATA_BUFFER;
+
+#ifndef IO_REPARSE_TAG_SYMLINK
+#  define IO_REPARSE_TAG_SYMLINK                  (0xA000000CL)
+#endif
+#ifndef IO_REPARSE_TAG_AF_UNIX
+#  define IO_REPARSE_TAG_AF_UNIX 0x80000023
+#endif
+#ifndef IO_REPARSE_TAG_LX_FIFO
+#  define IO_REPARSE_TAG_LX_FIFO 0x80000024
+#endif
+#ifndef IO_REPARSE_TAG_LX_CHR
+#  define IO_REPARSE_TAG_LX_CHR  0x80000025
+#endif
+#ifndef IO_REPARSE_TAG_LX_BLK
+#  define IO_REPARSE_TAG_LX_BLK  0x80000026
+#endif
+
 static int
-win32_stat_low(HANDLE handle, const char *path, STRLEN len, Stat_t *sbuf) {
+win32_stat_low(HANDLE handle, const char *path, STRLEN len, Stat_t *sbuf,
+               DWORD reparse_type) {
     DWORD type = GetFileType(handle);
     BY_HANDLE_FILE_INFORMATION bhi;
 
     Zero(sbuf, 1, Stat_t);
+
+    if (reparse_type) {
+        /* Lie to get to the right place */
+        type = FILE_TYPE_DISK;
+    }
 
     type &= ~FILE_TYPE_REMOTE;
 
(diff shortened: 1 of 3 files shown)


bram-perl:
Just a small heads up: I'm planning on doing some testing with this and some reviewing;

(so if possible: wait with merging until I had a chance to test/review)

tonycoz:
> (so if possible: wait with merging until I had a chance to test/review)

That's no problem.

One test is failing, though I don't see why yet.

tonycoz:
> One test is failing, though I don't see why yet.

It succeeds sometimes, but mostly fails.  I haven't managed to get it to fail locally, and DrMemory (like valgrind) isn't showing anything relevant that I can see.

tonycoz:
> It succeeds sometimes, but mostly fails.

Or it was, I can't get it to fail in CI at all now.

sisyphus:
FWIW, this 20204-stat-socket branch is fine for me when I build it on Windows 11 using gcc-12, where _WIN32_WINNT is 0x0a00 - though some of the TODO tests fail, and the only archtype I've tested is MSWin32-x64-multi-thread.

With the same archtype and same compiler on Windows 7, t/win32/stat.t doesn't run to completion.
When run outside the Test::Harness, it terminates with:
```
....
ok 46 - filename '"' shouldn't exist
ok 47 # skip Cannot create unix socket
Label not found for "last SKIP" at ./test.pl line 535.
```
Something awry with the <code>SKIP</code> construction ?
It also fails some TODO tests.

Of course, there's a couple of questionable practices being undertaken on Windows 7:
1) AIUI, Windows 7 is technically not supported any more;
2) Using a compiler that has _WIN32_WINNT set to 0xa00 (Windows 10) on a Windows 7 machine seems a bit dubious.

But this has not been a problem before with gcc-12 builds of perl on Windows 7 - so long as optimization is set to <code>-Os</code> (which I think has just recently become the default optimization setting in blead's win32/GNUmakefile).

Cheers,
Rob

tonycoz:
The failing TODO tests are for bugs that had been reported against Win32 stat() before I re-worked it, unfortunately they seem to be Windows limits (or bugs, I guess).

win32/win32.c
```diff
@@ -1585,22 +1677,150 @@ win32_stat_low(HANDLE handle, const char *path, STRLEN len, Stat_t *sbuf) {
...
                     FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
                     NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
+    if (handle == INVALID_HANDLE_VALUE) {
+        /* AF_UNIX sockets need to be opened as a reparse point, but
+           that will also open symlinks rather than following them.
+
+           There may be other reparse points that need similar
+           treatment.
+        */
+        handle = S_follow_symlinks_to(aTHX_ path, &reparse_type);
```
bram-perl, line 1816: ~This part of the commit doesn't make sense to me...~

<details>
  <summary>Outdated comment</summary>

Reading https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#symbolic-link-behavior

What I **expect** when `path` is a symlink:
- if the `FILE_FLAG_OPEN_REPARSE_POINT` flag is not included then it points to the target of the symlink (i.e. to the destination, i.e. to the real file)
- if the `FILE_FLAG_OPEN_REPARSE_POINT` flag is included then it points to th e symlink

In this code the `FILE_FLAG_OPEN_REPARSE_POINT` flag is **not** included in the `CreateFileA` call so I would expect `handle` to refer to the target file (i.e. not the symlink) but then the comment and code reads: `S_follow_symlinks_to` which doesn't make sense: if it was a symlink then `handle` already points to the destination file.. (assuming I read the MS docs correct)~
</details>

**Update**: Ok, it does make a bit of sense. It's only a bit confusing.

It first calls `CreateFileA` without `FILE_FLAG_OPEN_REPARSE_POINT`.
That either returns a valid handle (to the target file) or return an invalid handle.
The invalid handle can mean three different things:
- there is no such file with that name
- there is a file with that name but it's a symlink and points to a non-existing file
- there is a file with that name but it's a unix socket

The code doesn't really check which case applies and (sort of) assumes it might be a symlink and calls `S_follow_symlinks_to` which calls `readlink` to follow the link (if it's a link I guess);
Eventually it calls `CreateFileA` again with the `FILE_FLAG_OPEN_REPARSE_POINT` flag set.

bram-perl: What I have/had initially in mind for this code was something like: (pseudo-code obviously):

```
    handle =
        CreateFileA(path, FILE_READ_ATTRIBUTES,
                    FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
                    NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT, NULL);
    if (handle != INVALID_HANDLE_VALUE) {
         result = GetFileInformationByHandleEx(handle, FileBasicInfo, &basicInfo, ...)
         if (baiscInfo.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) {

              DeviceIoControl(handle, FSCTL_GET_REPARSE_POINT, ....)
              ....
              if (reparse_type == IO_REÄRSE_TAG_SYMLINK) {
                  handle =
                      CreateFileA(path, FILE_READ_ATTRIBUTES,
                             FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
                             NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);

              }
         }
    }
```

That is: first open with `FILE_FLAG_OPEN_REPARSE_POINT` and if it's a symlink then call `CreateFileA` again but this time without the `FILE_FLAG_OPEN_REPARSE_POINT` flag so that it resolves the symlink.

bram-perl: An alternative to the current code could also be to change `S_follow_symlinks_to` to not follow the symlink chain and instead check the `reparse_type`.
That is:
- if `CreateFileA(...., FILE_FLAG_BACKUP_SEMANTICS, ...)` fails
- *and* if `CreateFileA(...., FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINTS, ...)` succeeds
- then check if `reparse_type` is 'symlink'. If it is return ENOENT

win32/win32.c
```diff
@@ -1488,13 +1488,77 @@ translate_ft_to_time_t(FILETIME ft) {
...
+win32_stat_low(HANDLE handle, const char *path, STRLEN len, Stat_t *sbuf,
+               DWORD reparse_type) {
     DWORD type = GetFileType(handle);
     BY_HANDLE_FILE_INFORMATION bhi;
 
     Zero(sbuf, 1, Stat_t);
 
+    if (reparse_type) {
+        /* Lie to get to the right place */
+        type = FILE_TYPE_DISK;
```
bram-perl, line 1559: In what case was this needed?
When I test with a unix socket it seems `GetFileType()` returns `FILE_TYPE_DISK`.
Did you encounter a case in which it returned something else?

win32/win32.c
bram-perl: Thinking some more.. I guess that won't do the right thing if it's a symlink to a unix socket..
(I'll do some testing tomorrow)

tonycoz: I do the non-reparse-point open first because I expect it to be by far the more common case, and it does the right thing for win32_stat() for valid symlink chains.

I went direct for win32_readlink() mostly out of laziness - I didn't want to include the CreateFile/DeviceIOControl/WideCharToMultiByte() code here (though maybe a win32_readlink_handle() would help), and win32_readlink() does the job.

I don't really consider this code path to be performance critical.

And yes, a symlink chain to a socket fails the non-reparse CreateFile().

win32/win32.c
tonycoz: This part may be unnecessary, I ass-umed that it would fail, but didn't test.

Thanks for the review.

demerphq (approved):
This passes the eyeball test for me. I dont have a win32 box to play with it on however. I need to fix that.

PR #20273, opened 2022-09-08
sv_setsv_cow - directly create desired SVt_COW SV

richardleach:
It's more efficient to directly create the desired SV type than create a SVt_NULL and then call sv_upgrade to get the desired type.


richardleach:
This is a really small and straightforward change. Given the number of open PRs, I'm just going to merge it.

PR #20275, opened 2022-09-08
toke.c/regcomp.c: Simplified handling of new SVs

richardleach:
The 3 commits in this PR reflect the fact that calling generic function/macros on new SVs is often inefficient. 
* In `S_scan_heredoc`, `SvGROW` is replaced by `sv_grow_fresh` and `sv_setpvn` by `sv_setpvn_fresh`, depending upon the if/else branch.
* Also in `S_scan_heredoc`, `SvPVCLEAR` is inlined and simplified.
* In `S_regclass`, another `SvPVCLEAR` is inlined and simplified.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -18444,8 +18444,8 @@ S_regclass(pTHX_ RExC_state_t *pRExC_state, I32 *flagp, U32 depth,
                                 * points. */
                                 AV * this_string = (AV *) av_shift( strings);
                                 STRLEN cp_count = av_count(this_string);
-                                SV * final = newSV(cp_count * 4);
-                                SvPVCLEAR(final);
+                                SV * final = newSV(cp_count ? cp_count * 4 : 1);
+                                SvPVCLEAR_FRESH(final);
 
                                 /* Create another string of sequences of \x{...} */
                                 while (av_count(this_string) > 0) {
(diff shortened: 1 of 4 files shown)


demerphq:
Maybe it would be better to introduce a SvPVCLEAR_FAST() or something? I added this macro specifically because I didnt want to remember the correct way to do $x= ""; Unrolling that seems like a step backwards even if I agree with the intent.

richardleach:
Yeah, that was my first thought but I wasn't sure if the low number of compatible uses justified it. I'm happy to do it for the reasoning you outlined though.

`SvPVCLEAR_FRESH` would keep with the `_fresh` naming convention used by other function in _sv.c_ and it can have the same constraints - only to be used on a freshly minted SVt_PV.

Sound okay?

demerphq:
Yeah, not gunna bike-shed the name, I just thought it was ironic as I am pretty sure I created the macro originally because i needed it in the regex engine, so possibly *because* of the macros you were making not use it. :-)

richardleach:
Ok, I've revised this PR with a new SvPVCLEAR_FRESH.

Also moved #20274 into this PR as an extra commit, since the new macro makes that change more straightforward.

richardleach:
@demerphq - you okay with this PR now?

demerphq:
Looka good to me.

Thanks!

Yves

PR #20279, opened 2022-09-09
locale.c: Change macro name 'porcelain' => 'posix'

khwilliamson:
When I wrote this I tought porcelain in git meant something different than it does.  But what this macro does is to hide the non-POSIX vs POSIX differences from its callers.


jkeenan:
> When I wrote this I tought porcelain in git meant something different than it does. But what this macro does is to hide the non-POSIX vs POSIX differences from its callers.

LGTM

bram-perl (approved):

rwp0 (approved):

PR #20280, opened 2022-09-10
locale.c: Don't use LC_ALL unless platform has it

khwilliamson:
There was one function where LC_ALL could be used even if not available on the platform.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6540,12 +6540,11 @@ Perl_thread_locale_init()
     /* On Windows, make sure new thread has per-thread locales enabled */
     _configthreadlocale(_ENABLE_PER_THREAD_LOCALE);
 
-#  else
+#  endif
 
     /* This thread starts off in the C locale */
     Perl_setlocale(LC_ALL, "C");
 
-#  endif
 #endif
 
 }
--- a/locale.c
+++ b/locale.c
@@ -6530,21 +6530,29 @@ Perl_thread_locale_init()
 
     dTHX_DEBUGGING;
 
-
      DEBUG_L(PerlIO_printf(Perl_debug_log,
-            "new thread, initial locale is %s; calling setlocale\n",
-            setlocale(LC_ALL, NULL)));
-
+                           "new thread, initial locale is %s;"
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -6531,21 +6531,27 @@ Perl_thread_locale_init()
 
     dTHX_DEBUGGING;
 
-
      DEBUG_L(PerlIO_printf(Perl_debug_log,
-            "new thread, initial locale is %s; calling setlocale\n",
-            setlocale(LC_ALL, NULL)));
-
+                           "new thread, initial locale is %s;"
+                           " calling setlocale(LC_ALL, \"C\"\n",
```
bram-perl: the message in the debug log is missing a closing `)`
i.e. it's logging:
```
new thread, initial locale is ........; calling setlocale(LC_ALL, "C"
```

locale.c
```diff
@@ -6531,21 +6531,27 @@ Perl_thread_locale_init()
...
+                           get_LC_ALL_display()));
 #  ifdef WIN32
 
     /* On Windows, make sure new thread has per-thread locales enabled */
     _configthreadlocale(_ENABLE_PER_THREAD_LOCALE);
 
-#  else
+#  endif
+#  if defined(LC_ALL)
```
bram-perl, line 6543: This (also) alters the behavior on Windows, I'm assuming that's intended but in that case it would be good to explicitly mention it in the commit message.

Before this commit:
- Win32 it called `_configthreadlocale(_ENABLE_PER_THREAD_LOCALE);`
- on everything else: it called `Perl_setlocale(LC_ALL, "C")`

Now the `Perl_setlocale(LC_ALL, "C")` (or the variant with the for loop) is also called on Windows.

locale.c
khwilliamson: fixed

locale.c
khwilliamson: That was a bug in Windows, but I  overlooked that I was fixing it.  Anyway, two commits are now done, so the messages are explicit

bram-perl (approved):

PR #20281, opened 2022-09-10
Mitigate unsafe threaded locales

khwilliamson:
This series of commits lays the groundwork for, and then implements a new set of macros and functions to do locale changing and querying for platforms where perl is compiled with threads, but the platform doesn't have thread-safe locale handling.  These add a modicum of thread-safety to such platforms

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -7148,41 +7148,6 @@ the plain locale pragma without a parameter (S<C<use locale>>) is in effect.
                                     MUTEX_DESTROY(&PL_locale_mutex);        \
                                 } STMT_END
 #endif
-#if ! (   defined(USE_LOCALE)                                                 \
-       &&    defined(USE_LOCALE_THREADS)                                    \
-       && (  ! defined(USE_THREAD_SAFE_LOCALE)))
-#else
-
-   /* Here, we will need critical sections in locale handling, because one or
-    * more of the above conditions are true.  This could be because the
-    * platform doesn't have thread-safe locales, or that at least one of the
-    * locale-dependent functions in the core isn't thread-safe.  The latter
-    * case is generally because they return a pointer to a static buffer, which
-    * may be per-process instead of per-thread.  There are supposedly
-    * re-entrant, safe versions for all of them Perl currently uses (which the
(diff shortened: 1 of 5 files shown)


perl.h
```diff
@@ -7148,17 +7148,10 @@ the plain locale pragma without a parameter (S<C<use locale>>) is in effect.
                                     MUTEX_DESTROY(&PL_locale_mutex);        \
                                 } STMT_END
 #endif
-#if ! (   defined(USE_LOCALE)                                               \
+#if ! (   defined(USE_LOCALE)                                                 \
        &&    defined(USE_LOCALE_THREADS)                                    \
        && (  ! defined(USE_THREAD_SAFE_LOCALE)))
```
bram-perl: This looks incorrect;

Before the code was:
```
#if ! (   defined(USE_LOCALE)                                               \
       &&    defined(USE_LOCALE_THREADS)                                    \
       && (  ! defined(USE_THREAD_SAFE_LOCALE)))
        ...
#else
         /* Here, we will need critical sections in locale handling, because one or
          * more of the above conditions are true.  This could be because the
#endif
```

This commits drops the `#else` but doesn't alter the `#if`so the code becomes:
```
#if ! (   defined(USE_LOCALE)                                               \
       &&    defined(USE_LOCALE_THREADS)                                    \
       && (  ! defined(USE_THREAD_SAFE_LOCALE)))
         /* Here, we will need critical sections in locale handling, because one or
          * more of the above conditions are true.  This could be because the
#endif
```
which looks wrong.

I think the `!` should be dropped inside the `#if` so that it becomes:

```
#if (   defined(USE_LOCALE)                                               \
       &&    defined(USE_LOCALE_THREADS)                                    \
       && (  ! defined(USE_THREAD_SAFE_LOCALE)))
         /* Here, we will need critical sections in locale handling, because one or
          * more of the above conditions are true.  This could be because the
         ...
#endif
```
but when that change is made then the comments no longer make sense the `one or more of the above conditions are true` part that it.)

I think it might be easier to keep the `#else` (the entire block is removed in the next commit anyway)

khwilliamson: ok

PR #20282, opened 2022-09-10
locale.c: Refactor internal debugging function

khwilliamson:
setlocale_debug_string() variants  now use Perl_form, a function I
didn't know existed when I originally wrote this code.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -179,6 +179,8 @@ static const char C_thousands_sep[] = "";
 #ifdef USE_LOCALE
 
 #  ifdef DEBUGGING
+#    define setlocale_debug_string_i(index, locale, result)                 \
+            my_setlocale_debug_string_i(index, locale, result, __LINE__)
 #    define setlocale_debug_string_c(category, locale, result)              \
                 setlocale_debug_string_i(category##_INDEX_, locale, result)
 #    define setlocale_debug_string_r(category, locale, result)              \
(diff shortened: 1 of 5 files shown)


locale.c
```diff
@@ -6471,56 +6473,66 @@ Perl_sync_locale()
...
+                     locale_quote, locale, locale_quote,
+                     retval_quote, retval, retval_quote);
 }
 
 #endif
 
 void
 Perl_thread_locale_init()
 {
-    /* Called from a thread on startup*/
```
bram-perl: Why the removal of that line? Seems unrelated to the commit/PR

locale.c
```diff
@@ -6471,56 +6473,66 @@ Perl_sync_locale()
...
     }
 
-    assert(strlen(ret) < sizeof(ret));
+#  ifdef USE_LOCALE_THREADS
+#    define THREAD_FORMAT "%p:"
+#    define THREAD_ARGUMENT aTHX_
+#  else
+#    define THREAD_FORMAT
+#    define THREAD_ARGUMENT
+#  endif
```
bram-perl, line 6527: I wonder if this should be followed by a `#undef THREAD_FORMAT`, `#undef THREAD_ARGUMENT`..

That is
```
# ifdef ...
#   define THREAD_FORMAT ...
# endif

    return ....

# undef THREAD_FORMAT
# undef THREAD_ARGUMENT
```

Right now it's leaving these defined; I don't know what the recommended practice is tho..

locale.c
```diff
@@ -6471,56 +6473,66 @@ Perl_sync_locale()
...
-    static char ret[1024];
-    assert(cat_index <= NOMINAL_LC_ALL_INDEX);
+    const char * locale_quote;
+    const char * retval_quote;
 
-    my_strlcpy(ret, "setlocale(", sizeof(ret));
-    my_strlcat(ret, category_names[cat_index], sizeof(ret));
-    my_strlcat(ret, ", ", sizeof(ret));
+    assert(cat_index <= NOMINAL_LC_ALL_INDEX);
+    PERL_UNUSED_ARG(line);  /* Currently unused */
```
bram-perl: That doesn't seem to be true?
`line` is used on line 6525

khwilliamson: You're right

locale.c
khwilliamson: rebase -i error on my part

PR #20283, opened 2022-09-11
regcomp.c - use _simple functions where obviously safe to do so

richardleach:
Sort of a follow-up to https://github.com/Perl/perl5/pull/20028.
This PR changes:
* `av_store` -> `av_store_simple`
* `av_push` -> `av_push_simple`
where it seems safe to do so. (i.e. In the same region as the AV being operated upon was created.)

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -2444,7 +2444,7 @@ S_dump_trie(pTHX_ const struct _reg_trie_data *trie, HV *widecharmap,
         depth+1, "Match","Base","Ofs" );
 
     for( state = 0 ; state < trie->uniquecharcount ; state++ ) {
-        SV ** const tmp = av_fetch( revcharmap, state, 0);
+        SV ** const tmp = av_fetch_simple( revcharmap, state, 0);
         if ( tmp ) {
             Perl_re_printf( aTHX_  "%*s",
                 colwidth,
@@ -2553,7 +2553,7 @@ S_dump_trie_interim_list(pTHX_ const struct _reg_trie_data *trie,
             );
         }
         for( charid = 1 ; charid <= TRIE_LIST_USED( state ) ; charid++ ) {
-            SV ** const tmp = av_fetch( revcharmap,
+            SV ** const tmp = av_fetch_simple( revcharmap,
                                         TRIE_LIST_ITEM(state, charid).forid, 0);
             if ( tmp ) {
                 Perl_re_printf( aTHX_  "%*s:%3X=%4" UVXf " | ",
@@ -2603,7 +2603,7 @@ S_dump_trie_interim_table(pTHX_ const struct _reg_trie_data *trie,
     Perl_re_indentf( aTHX_  "Char : ", depth+1 );
 
     for( charid = 0 ; charid < trie->uniquecharcount ; charid++ ) {
-        SV ** const tmp = av_fetch( revcharmap, charid, 0);
+        SV ** const tmp = av_fetch_simple( revcharmap, charid, 0);
         if ( tmp ) {
             Perl_re_printf( aTHX_  "%*s",
                 colwidth,
@@ -2779,10 +2779,10 @@ is the recommended Unicode-aware way of saying
             SvCUR_set(zlopp, kapow - flrbbbbb);				   \
             SvPOK_on(zlopp);						   \
             SvUTF8_on(zlopp);						   \
-            av_push(revcharmap, zlopp);					   \
+            av_push_simple(revcharmap, zlopp);					   \
         } else {							   \
             char ooooff = (char)val;                                           \
-            av_push(revcharmap, newSVpvn(&ooooff, 1));			   \
+            av_push_simple(revcharmap, newSVpvn(&ooooff, 1));			   \
         }								   \
         } STMT_END
 
@@ -2841,7 +2841,7 @@ is the recommended Unicode-aware way of saying
             tmp = newSVpvn_utf8(STRING(noper), STR_LEN(noper), UTF);	\
         else                                                    \
             tmp = newSVpvn_utf8( "", 0, UTF );			\
-        av_push( trie_words, tmp );                             \
+        av_push_simple( trie_words, tmp );                             \
     });                                                         \
                                                                 \
     curword++;                                                  \
@@ -3738,7 +3738,7 @@ S_make_trie(pTHX_ RExC_state_t *pRExC_state, regnode *startbranch,
                             /* if this is the first state there is no common prefix
                              * to extract, so we can exit */
                             if ( state == 1 ) break;
-                            tmp = av_fetch( revcharmap, ofs, 0);
+                            tmp = av_fetch_simple( revcharmap, ofs, 0);
                             ch = (U8*)SvPV_nolen_const( *tmp );
 
                             /* if we are on count 2 then we need to initialize the
@@ -3752,7 +3752,7 @@ S_make_trie(pTHX_ RExC_state_t *pRExC_state, regnode *startbranch,
                                         depth+1,
                                         (UV)state));
                                 if (first_ofs >= 0) {
-                                    SV ** const tmp = av_fetch( revcharmap, first_ofs, 0);
+                                    SV ** const tmp = av_fetch_simple( revcharmap, first_ofs, 0);
                                     const U8 * const ch = (U8*)SvPV_nolen_const( *tmp );
 
                                     TRIE_BITMAP_SET_FOLDED(trie,*ch, folder);
@@ -3772,7 +3772,7 @@ S_make_trie(pTHX_ RExC_state_t *pRExC_state, regnode *startbranch,
                     /* This state has only one transition, its transition is part
                      * of a common prefix - we need to concatenate the char it
                      * represents to what we have so far. */
-                    SV **tmp = av_fetch( revcharmap, first_ofs, 0);
+                    SV **tmp = av_fetch_simple( revcharmap, first_ofs, 0);
                     STRLEN len;
                     char *ch = SvPV( *tmp, len );
                     DEBUG_OPTIMISE_r({
@@ -3825,7 +3825,7 @@ S_make_trie(pTHX_ RExC_state_t *pRExC_state, regnode *startbranch,
                    ) {
                    U32 word = trie->wordcount;
                    while (word--) {
-                       SV ** const tmp = av_fetch( trie_words, word, 0 );
+                       SV ** const tmp = av_fetch_simple( trie_words, word, 0 );
                        if (tmp) {
                            if ( STR_LEN(convert) <= SvCUR(*tmp) )
                                sv_chop(*tmp, SvPV_nolen(*tmp) + STR_LEN(convert));
@@ -9130,7 +9130,7 @@ Perl_reg_named_buff_all(pTHX_ REGEXP * const r, const U32 flags)
                 }
             }
             if (parno || flags & RXapif_ALL) {
-                av_push(av, newSVhek(HeKEY_hek(temphe)));
+                av_push_simple(av, newSVhek(HeKEY_hek(temphe)));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


richardleach:
Out of interest, I blindly changed all instances of `av_store`, `av_fetch`, and `av_push` in _regcomp.c_ to their `_simple` equivalents to see if tests would fail. They did - clearly some AVs in there are READONLY, or magic, or have keys <0:
```
t/re/opt ......................................................... ok
t/re/overload .................................................... ok
t/re/pat ......................................................... ok
t/re/pat_advanced ................................................ # Failed test 1662 - GH \#17734 at ./test.pl line 1189
#      got 'perl: ./inline.h:136: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `!SvREADONLY(av)\' failed.\nAborted'
# expected /(?^:Done)/
# PROG: 
# no warnings "experimental::uniprop_wildcards";
#                          my $re = q<[[\p{name=/[Y-]+Z/}]]>;
#                          eval { "\N{BYZANTINE MUSICAL SYMBOL PSILI}"
#                                 =~ /$re/ }; print $@ if $@; print "Done\n";
# STATUS: 34304
FAILED at test 1662
t/re/pat_advanced_thr ............................................ skipped
t/re/pat_psycho .................................................. ok
t/re/pat_psycho_thr .............................................. skipped
t/re/pat_re_eval ................................................. perl: ./inline.h:135: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `!SvMAGICAL(av)' failed.
FAILED--expected 508 tests, saw 437
t/re/pat_re_eval_thr ................

t/re/reg_mesg .................................................... perl: ./inline.h:138: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `key > -1' failed.
FAILED--no leader found
t/re/reg_namedcapture ............................................ ok
t/re/reg_nc_tie .................................................. ok
t/re/reg_nocapture ............................................... ok
t/re/reg_pmod .................................................... ok
t/re/reg_posixcc ................................................. ok
t/re/regex_sets .................................................. perl: ./inline.h:138: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `key > -1' failed.
FAILED--no leader found
t/re/regex_sets_compat ....

t/op/svflags ..................................................... ok
t/op/svleak ...................................................... perl: ./inline.h:138: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `key > -1' failed.
FAILED--expected 151 tests, saw 74

lib/charnames .................................................... perl: ./inline.h:136: SV **Perl_av_fetch_simple(AV *, ssize_t, I32): Assertion `!SvREADONLY(av)' failed.
FAILED--no leader found
```
So any further changes will need a domain expert to go through each instance in turn.

demerphq:
Those errors are one reason im a touch hesitant about inline functions. It doesnt show the place the inline function is called, it shows the inline function, so its a bit hard to know what is going on here. Could you change it so that the assert happens BEFORE the function is called so we can see where these are coming from? :-(

demerphq:
FWIW, you can almost certainly change any instance used against the vars revcharmap, trie_words, stack and name_list. The others I am not sure about.

richardleach:
> FWIW, you can almost certainly change any instance used against the vars revcharmap, trie_words, stack and name_list.

Nope, one or more of them causes assertions. I'll sniff out which one(s).

richardleach:
> > FWIW, you can almost certainly change any instance used against the vars revcharmap, trie_words, stack and name_list.
> 
> Nope, one or more of them causes assertions. I'll sniff out which one(s).

`stack` is the only one that asserts and then only if some of the instances of `av_fetch(stack` in `S_handle_regex_sets` are replaced by `av_fetch_simple(stack`. The incompatibility is that `av_fetch_simple` needs its second parameter (`key`) to be > -1, but that's not always the case here.

Might be easier to show with an annotated diff:
```
diff --git a/regcomp.c b/regcomp.c
index 43dbc19662..173da20cdc 100644
--- a/regcomp.c
+++ b/regcomp.c
@@ -17164,6 +17164,8 @@ redo_curchar:
                     /* If the top entry on the stack is an operator, it had
                      * better be a '!', otherwise the entry below the top
                      * operand should be an operator */
+/*seems fine*/
+    assert(top_index > -1);
                     if (   ! (top_ptr = av_fetch(stack, top_index, FALSE))
                         || (IS_OPERATOR(*top_ptr) && SvUV(*top_ptr) != '!')
                         || (   IS_OPERAND(*top_ptr)
@@ -17303,7 +17305,8 @@ redo_curchar:
             case '+':
             case '-':
             case '^':
-
+/* Not okay. top_index is not always > -1 here */
+    assert(top_index > -1);
                 /* These binary operators should have a left operand already
                  * parsed */
                 if (   top_index - fence < 0
@@ -17335,6 +17338,8 @@ redo_curchar:
 
                 /* The operator on the stack is supposed to be below both its
                  * operands */
+/* Not okay. (top_index - 2) is not always > -1 here */
+    assert((top_index -2 ) > -1);
                 if (   ! (stacked_ptr = av_fetch(stack, top_index - 2, FALSE))
                     || IS_OPERAND(*stacked_ptr))
                 {
@@ -17432,6 +17437,8 @@ redo_curchar:
 
                 /* If what's already at the top of the stack is another '!",
                  * they just cancel each other out */
+/* top_index is not always > -1 here */
+    assert(top_index > -1);
                 if (   (top_ptr = av_fetch(stack, top_index, FALSE))
                     && (IS_OPERATOR(*top_ptr) && SvUV(*top_ptr) == '!'))
                 {
@@ -17463,6 +17470,8 @@ redo_curchar:
                 /* If the top entry on the stack is an operator, it had better
                  * be a '!', otherwise the entry below the top operand should
                  * be an operator */
+/* seems okay */
+    assert(top_index > -1);
                 top_ptr = av_fetch(stack, top_index, FALSE);
                 assert(top_ptr);
                 if (IS_OPERATOR(*top_ptr)) {
@@ -17662,6 +17671,8 @@ S_dump_regex_sets_structures(pTHX_ RExC_state_t *pRExC_state,
     else {
         PerlIO_printf(Perl_debug_log, "Stack: (fence=%d)\n", (int) fence);
         for (i = stack_top; i >= 0; i--) {
+/* No failures here, but is it tested under just -DDEBUGGING? */
+    assert(i > -1);
             SV ** element_ptr = av_fetch(stack, i, FALSE);
             if (! element_ptr) {
             }
```

richardleach:
I've pushed additional the changes that didn't cause test failures.

richardleach:
Ok, think I've been though every `av_push`, `av_store`, and `av_fetch` in _regcomp.c_.  All that can be converted to `_simple` without triggering asserts or failures in the test suite have been so converted in these (unsquashed) commits.

I'm not saying that I think these changes should definitely be merged, the intention was just to do some legwork for the regex engine gurus.

demerphq:
They look ok to me. Curious if we should have and av_fetch_top for the -1
case?

Yves

richardleach:
> They look ok to me. 

Ok, I'll squash them.

> Curious if we should have and av_fetch_top for the -1 case?

Something like this?
```
SV**
Perl_av_fetch_top(pTHX_ AV *av, SSize_t key, I32 lval)
{

    PERL_ARGS_ASSERT_AV_FETCH;
    assert(SvTYPE(av) == SVt_PVAV);
    assert(!SvMAGICAL(av));
    assert(!SvREADONLY(av));
    assert(AvREAL(av));
    assert(key == -1);

    if (AvFILLp(av) < 0)
        return NULL;
    else if (AvFILLp(av) == 0)
        key = 0;
    else
        key = AvFILLp(av) - 1;

    if (!AvARRAY(av)[key]) {
        return lval ? av_store_simple(av,key,newSV_type(SVt_NULL)) : NULL;
    }

    return &AvARRAY(av)[key];
}
```
We'd have to know in advance that it was the -1 case though, in order to use the correct function? Or are you thinking of some kind of wrapper that checks for the -1 case?

demerphq:
On Sat, 17 Sept 2022, 23:40 Richard Leach, ***@***.***> wrote:

> They look ok to me.
>
> Ok, I'll squash them.
>
> Curious if we should have and av_fetch_top for the -1 case?
>
> Something like this?
>
> SV**
> Perl_av_fetch_top(pTHX_ AV *av, SSize_t key, I32 lval)
> {
>
>     PERL_ARGS_ASSERT_AV_FETCH;
>     assert(SvTYPE(av) == SVt_PVAV);
>     assert(!SvMAGICAL(av));
>     assert(!SvREADONLY(av));
>     assert(AvREAL(av));
>     assert(key == -1);
>
>     if (AvFILLp(av) < 0)
>         return NULL;
>     else if (AvFILLp(av) == 0)
>         key = 0;
>     else
>         key = AvFILLp(av) - 1;
>
>     if (!AvARRAY(av)[key]) {
>         return lval ? av_store_simple(av,key,newSV_type(SVt_NULL)) : NULL;
>     }
>
>     return &AvARRAY(av)[key];
> }
>
> We'd have to know in advance that it was the -1 case though, in order to
> use the correct function? Or are you thinking of some kind of wrapper that
> checks for the -1 case?
>

I was thinking that we have at least some code that wants to do the fetch
equivalent of pop() by fetching the -1 index. I was thinking that providing
an av_fetch_top() that could be used instead, and maybe wouldn't need a
guard to make sure the array was nonempty, eg if it return null in such
cases.

So the dev would have to opt in, I wasn't thinking of entangling it with
any other operation. It just caught my eye in this thread because this case
broke your new function, so I'd use av_fetch_top() in those cases.

Maybe it should be av_fetch_top_simple(). Whatever. :-)

Yves

iabyn:
On Thu, Sep 15, 2022 at 04:30:34PM -0700, Richard Leach wrote:
> Ok, think I've been though every `av_push`, `av_store`, and `av_fetch` in _regcomp.c_.  All that can be converted to `_simple` without triggering asserts or failures in the test suite have been so converted in these (unsquashed) commits.
> 
> I'm not saying that I think these changes should definitely be merged, the intention was just to do some legwork for the regex engine gurus. 

Does this set of "simple" fetch/stores result from

1) changing everything to _simple() and reverting the ones which break,
or,
2) examining every occurrence of fetch()/store() in the source and by the
exercise of intellect, determining which ones are safe to simplify

?

Because I would be very unhappy with this commit if (1) was the case.

-- 
"Do not dabble in paradox, Edward, it puts you in danger of fortuitous wit."
    -- Lady Croom, "Arcadia"

richardleach:
> Does this set of "simple" fetch/stores result from 1) changing everything to _simple() and reverting the ones which break, or, 2) examining every occurrence of fetch()/store() in the source and by the exercise of intellect, determining which ones are safe to simplify ? 

The original commit was (2), since limited changes were only made close to where AVs were created.

@demerphq then advised of a list of variables for which changes to `_simple()` functions should be safe, but we found that some calls using`stack` have a `key` value less than `0`, which is incompatible with `av_fetch_simple`. So all uses - even ones that didn't cause any test failures - of `av_fetch_simple` with `stack` in `S_handle_regex_sets` were reverted back to `av_fetch`.

So the changes are then:
* on variables thought to be compatible, and
* ones that haven't caused test failures.

> Because I would be very unhappy with this commit if (1) was the case.

I understand if you want this reverting for further analysis.

iabyn:
On Mon, Sep 26, 2022 at 01:06:12PM -0700, Richard Leach wrote:
> So the changes are then:
> * on variables thought to be compatible, and
> * ones that haven't caused test failures.
> 
> > Because I would be very unhappy with this commit if (1) was the case.
> 
> I understand if you want this reverting for further analysis.

Well, this seems like a minor optimisation on the regex compilation code,
which is (in general) less performance critical than regex execution
(regexec.c). So the performance gain versus risk of breakage balance
should be much more skewed towards avoiding breakage.

Unless its completely obvious from inspecting just the nearby code that
the array access *must* be simple, I'd avoid hacking it. The very fact
that some of the changes broke existing tests is a good indication that it
wasn't obvious. And the fact that tests currently pass isn't a sure
indication that it hasn't broken anything on CPAN.

Some years ago I introduced the PADRANGE op, which merges multiple PADMY
ops into a single op for expressions such as:

    my ($x,$y.$z) = ...;

However, it turned out that my assumption that all the pad target indices
would always be sequential was an assumption that a CPAN module managed to
break. So the moral of this story is to be risk averse unless it's really
justified!

-- 
"Do not dabble in paradox, Edward, it puts you in danger of fortuitous wit."
    -- Lady Croom, "Arcadia"

PR #20288, opened 2022-09-12
regen/regcomp.pl - add more detail on what the script does and list its inputs properly

demerphq:
Previously we were only showing one of the inputs that generates regnodes.h, now we show all of them. Also add some paragraphs explaining a bit more detail what the files contain and how we process.


regen/regcomp.pl
```diff
@@ -17,6 +18,34 @@
...
+# the regex engine to show things like which flags were set during
+# compilation. In some cases we transform the C code in the header files
+# into perl code which we execute to C<eval()> the contents. For instance
+# in a situation like this:
+#
+#   #define RXf_X 0x1   /* the X mode */
+#   #define RXf_Y 0x2   /* the Y mode */
+#   #define RXf_Z (X|Y) /* the Z mode */
+#
+# this script might end up eval()ing something like C<0x1> and then
```
jkeenan, line 40: s/^this script/This script/

regen/regcomp.pl
```diff
@@ -17,6 +18,34 @@
...
+# F<regcomp.sym> defines the opcodes and states used in the regex
+# engine, it also includes documentation on the opcodes. This script
+# parses those definitions out and turns them into typedefs, defines,
+# and data structures, and maybe even code which the regex engine can
+# use to operate.
+#
+# F<regexp.h> and op_reg_common.h contain defines C<RXf_xxx> and
+# C<PREGf_xxx> that are used in flags in our code. These defines are
+# parsed out and data structures are created to allow the debug mode of
+# the regex engine to show things like which flags were set during
```
jkeenan, line 31: (At the risk of being pedantic) consider `show things such as which`.

Otherwise +1.

tonycoz:
Looks good to me, with jkeenan's suggestions.

regen/regcomp.pl
khwilliamson: I disagree, `this script` is not beginning a sentence

khwilliamson (approved):

regen/regcomp.pl
jkeenan: > I disagree, `this script` is not beginning a sentence

Well, then, perhaps we could place ellipses at the end of the phrase above the code sample and at the beginning of the phrase after the code sample.  (This is a suggestion, not a requirement.)

jkeenan: @khwilliamson, are you okay with my revised suggestion above?

jkeenan:
In order to resolve this ticket, I'm accepting @khwilliamson's point in one instance and applying my grammatical correction in another instance.  Rebased on blead and merged into blead locally.  Pushed as 51135966bc.

demerphq:
@jkeenan thanks for picking this up, I lost track of it.

regen/regcomp.pl
demerphq: Ah willdo in the future, sorry i missed this feedback!

PR #20289, opened 2022-09-12
Fix defined PERL_CORE

Jamaika1:


tonycoz:
As it says at the top of the file, this file is generated and shouldn't be manually edited.

Jamaika1:
In perl.h is default define MULTIPLICITY for USE_ITHREADS. My question is how to use USE_ITHREADS?

Leont:
I'm not sure what this change is trying to fix, but I'm pretty sure it's the wrong fix anyway.

jkeenan:
> I'm not sure what this change is trying to fix, but I'm pretty sure it's the wrong fix anyway.

What @LeonT said.

PR #20293, opened 2022-09-12
Fix t/porting/authors.t breakgage for people with non-ASCII names

ilmari:
The lack of decoding of git config values in Porting/updateAUTHORS.pm would make it erroneously claim that anyone with a non-ASCII character in their name was an unknown author.

The decoding and encoding in updateAUTHORS.pl is not strictly necessary, since that's just passing things through to the terminal, but let's be consistent.


leonerd (approved):
LGTM

PR #20294, opened 2022-09-12
Set aTHX to NULL after freeing its data

khwilliamson:
It's an invitation to disaster to leave a pointer around pointing to freed space.  This keeps any code from referencing that space.


Leont:
I'm not sure this is really meaningful, given that most API calls will not use the implicit context (but use an explicit one instead). The function that calls `S_ithread_clear` will still have that same pointer.

PR #20299, opened 2022-09-13
Revert "Don't use POSIX 2008 locales on cygwin"

tonycoz:
This reverts commit 5ee7770b70fd48652f905014e47b07eb031ed5bc.

Claimed to be and apparently fixed:
 https://cygwin.com/pipermail/cygwin/2022-September/252178.html

### DIFF
--- a/cygwin/cygwin.c
+++ b/cygwin/cygwin.c
@@ -160,7 +160,6 @@ leave:
 char*
 wide_to_utf8(const wchar_t *wbuf)
 {
-    dTHX;
     char *buf;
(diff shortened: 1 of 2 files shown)


bram-perl:
I wonder.. should this just be reverted or should it be conditional on the cygwin version being used?
I'm guessing the hints change is still needed for older cygwin versions?

khwilliamson:
I agree with @bram-perl

khwilliamson:
But we should fix things to take advantage of the cygwin changes

PR #20300, opened 2022-09-14
Small docs clarification on MRO resolve function

leonerd:
Points out that the linear MRO function returns a list of strings.

Before this it wasn't clear to me whether it's a list of string names, or actual HV pointers to stashes.

### DIFF
--- a/ext/mro/mro.pm
+++ b/ext/mro/mro.pm
@@ -12,7 +12,7 @@ use warnings;
 
 # mro.pm versions < 1.00 reserved for MRO::Compat
 #  for partial back-compat to 5.[68].x
-our $VERSION = '1.26';
+our $VERSION = '1.27';
 
 require XSLoader;
 XSLoader::load('mro');
--- a/ext/mro/mro.xs
+++ b/ext/mro/mro.xs
@@ -14,7 +14,8 @@ static const struct mro_alg c3_alg =
 =for apidoc mro_get_linear_isa_c3
 
 Returns the C3 linearization of C<@ISA>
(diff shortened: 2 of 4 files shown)


ilmari (changes requested):
I think the [perlapi](https://perldoc.perl.org/perlapi#mro_get_linear_isa) entries (in `mro_core.c`) should be updated as well.

leonerd:
Good plan. Updated. (Should squash before merge)

ilmari:
There's two entries in `mro_core.c`: `mro_get_linear_isa` as well as `mro_get_linear_isa_dfs`, and I just noticed there's `mro_get_linear_isa_c3` in `ext/mro/mro.xs`.

Maybe the MRO-specific ones shouldn't have all the details, and just link to `mro_get_linear_isa` or `perlmroapi`?

leonerd:
Eh; I don't feel like a big docs rewrite at this point. This is a 4th(?)-level dependency of something I happen to be working on. I'll just add that they're strings, to each of the locations.

leonerd:
Ugh, if I update ext/mro/mro.xs then Porting/cmp_version.t gets upset that I didn't bump the version number. I'm not sure I want to bump a version number purely to add a tiny clarification note on the docs of a function almost nobody cares about :P

ilmari:
> I'm not sure I want to bump a version number purely to add a tiny clarification note on the docs of a function almost nobody cares about :P

Are version numbers that expensive? ;-P

demerphq:
I updated the version and force pushed this with a better formatted commit message.

demerphq (approved):

Issue #20307, opened 2022-09-16
^D doesn't always work in the perl debugger


x-yuri:
Module: lib/perl5db.pl

**Description**

<kbd>^D</kbd> doesn't work at the beginning of the session, after `n`, `s`, and maybe in some other cases.

From what I can see that happens because generally `$DB::single` is [reset](https://github.com/Perl/perl5/blob/v5.37.3/lib/perl5db.pl#L3106-L3107) before processing a command. But at the beginning of a session it doesn't get to that line. After `n` and `s` `DB::DB()` is supposedly set to be executed [again](https://github.com/Perl/perl5/blob/v5.37.3/lib/perl5db.pl#L3865-L3901) (`$DB::single = 1/2`).

**Steps to Reproduce**

`perl -de 'print 1; print 2'`, <kbd>^D</kbd>

**Expected behavior**

It runs the script to the end.

**Actual behavior**
It stops on the `print 2` statement.

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 36 subversion 0) configuration:
... (I will provide the rest if need be)
```

tonycoz:
I suspect 80c1f1e45e8 changed the behaviour of this, with blead:
```
tony@venus:.../git/perl5$ ./perl -Ilib -lde 'print 1; print 2; '

Loading DB routines from perl5db.pl version 1.76
Editor support available.

Enter h or 'h h' for help, or 'man perldebug' for more help.

main::(-e:1):   print 1; print 2; 
  DB<1> 1
main::(-e:1):   print 1; print 2; 
  DB<1> 2
tony@venus:.../git/perl5$
```
I hit ^D only once, and since the eof flag is now left on, the second attempt to read also return eof and the run proceeds.

You said you expected the script to run to the end, but the documentation indicates both `q` and `^D` should *exit* the debugger, and q immediately exits without running the rest of the script.

x-yuri:
> You said you expected the script to run to the end, but the documentation indicates both q and ^D should exit the debugger, and q immediately exits without running the rest of the script.

Oh, I meant it should stop, one way or another. Immediate exit sounds fine to me.

> I hit ^D only once, and since the eof flag is now left on, the second attempt to read also return eof and the run proceeds.

Can you elaborate on the eof flag? Where it's set/reset?

I'm not sure I understand your description, but for me `^D` at the beginning, and after `n`/`s` acts as if it's `n`. After you give it some other command, `^D` makes it exit.

[...]

PR #20308, opened 2022-09-16
perldebug: clarify the readline support

x-yuri:
Also I don't see any traces that [`Term::ReadKey`][a] is used (`rg ReadKey /usr/share/perl5 /usr/lib/perl5`). I'm running `perl-5.36.0` under Linux. And it doesn't seem like installing it makes any difference. Is it needed only under Windows and for it to get used you need [`Term::ReadLine::Perl5`][b]? But the former is a dependency of the latter, so that should be handled automatically. Which means under Windows `Term::ReadLine::Perl5` is preferred over `Term::ReadLine::Gnu`?

In other words a couple of corrections might be needed. Another one: Perl -> Perl5.

[a]: https://metacpan.org/pod/Term::ReadKey
[b]: https://metacpan.org/pod/Term::ReadLine::Perl5

### DIFF
--- a/pod/perldebug.pod
+++ b/pod/perldebug.pod
@@ -1145,7 +1145,7 @@ use only, and as such are subject to change without notice.
 
 As shipped, the only command-line history supplied is a simplistic one
 that checks for leading exclamation points.  However, if you install
-the L<Term::ReadKey> and L<Term::ReadLine> modules from CPAN (such as
+the L<Term::ReadKey> and one of the L<Term::ReadLine::*> modules from CPAN (such as
 L<Term::ReadLine::Gnu>, L<Term::ReadLine::Perl>, ...) you will
 have full editing capabilities much like those GNU I<readline>(3) provides.
 Look for these in the F<modules/by-module/Term> directory on CPAN.


jkeenan:
> Also I don't see any traces that [`Term::ReadKey`](https://metacpan.org/pod/Term::ReadKey) is used (`rg ReadKey /usr/share/perl5 /usr/lib/perl5`). I'm running `perl-5.36.0` under Linux. And it doesn't seem like installing it makes any difference. Is it needed only under Windows and for it to get used you need [`Term::ReadLine::Perl5`](https://metacpan.org/pod/Term::ReadLine::Perl5)? But the former is a dependency of the latter, so that should be handled automatically. Which means under Windows `Term::ReadLine::Perl5` is preferred over `Term::ReadLine::Gnu`?
> 
> In other words a couple of corrections might be needed. Another one: Perl -> Perl5.

I'm confused by your comment.  If Term::ReadKey is indeed *not* used, then why did you not delete the mention of it in your patch?

Also, is this pull request at all related to the issue you filed today: https://github.com/Perl/perl5/issues/20307 ?

x-yuri:
> I'm confused by your comment. If Term::ReadKey is indeed not used, then why did you not delete the mention of it in your patch?

Well, I made a change. But than a couple of thoughts came to mind. And while I'm certain about the change, I'm not sure about the rest:

1. `Term::ReadKey` [might be][a] needed only on Windows:

    >  Version 2.30.01: Added handling of arrows, page up/down, home/end, insert/delete keys under Win32. These keys emit xterm-compatible sequences. Works with Term::ReadLine::Perl.

    In that case for it to be used you need `Term::ReadLine::Perl5`. But `Term::ReadLine::Perl5` depends on `Term::ReadKey` anyway. As such the latter doesn't need to be mentioned. Unless I'm missing something.

2. Which makes me think, if `Term::ReadLine::Perl5` was created to overcome issues with `Term::ReadLine::Gnu` on Windows. Another possibility to improve the docs.

3. Shouldn't `Term::ReadLine::Perl` be changed to `Term::ReadLine::Perl5`? [`Term::ReadLine::Perl`][b] seems to exist, but old and maybe abandoned.

I think we could discuss here if any of the above assumptions are true and should be implemented in this PR. Or I'm open to your other suggestions.

> Also, is this pull request at all related to the issue you filed today: https://github.com/Perl/perl5/issues/20307?

No, not related. I don't have a solution for that one in mind.

[a]: https://metacpan.org/pod/Term::ReadKey
[b]: https://metacpan.org/dist/Term-ReadLine-Perl

tonycoz:
`Term::ReadLine::Perl` (no `5`) uses `Term::ReadKey` to read the terminal size, falling back to an ioctl(), the `5` version always uses `Term::ReadKey` to read the terminal size.

> Shouldn't Term::ReadLine::Perl be changed to Term::ReadLine::Perl5? [Term::ReadLine::Perl](https://metacpan.org/dist/Term-ReadLine-Perl) seems to exist, but old and maybe abandoned.

The plain `Perl` version is passing on CPAN testers, and on many platforms, perhaps the author doesn't think it needs an update.

I think the change adds a little clarity (you don't need all of the Term::RL::* modules, just one).

jkeenan:
@x-yuri, more than a year has elapsed since there last was activity in this pull request.  No one has spoken up in favor of this request, so I'm going to label the ticket 'Closable?'.  I am self-assigning it for the purpose of closing it within 1 or 2 weeks unless there is a strong argument for keeping it open.

tonycoz (approved):

tonycoz:
I think the change as is is an improvement.

Installing Term::Key is desirable so I think leaving the suggestion to install it is fine.

khwilliamson:
I think this should be merged.  @tonycoz spoke up in favor of it, but @jkeenan says no one has.  I don't understand

jkeenan:
> I think this should be merged. @tonycoz spoke up in favor of it, but @jkeenan says no one has. I don't understand

@khwilliamson, in his Oct 18 comment @tonycoz expressed some doubts about the patch.  Subsequently I reviewed the ticket, did not find any *unambiguous* support for the patch and applied the "Closable?" label.  Tony then clarified that he was in favor of the patch.  At that point I removed the "Closable?" label.  So, if both  you and Tony support the patch, you can merge.

PR #20309, opened 2022-09-16
CI: Improve dist-modules (group output + continue on error)

bram-perl:
Highlights:

- Group the output per dist (before there was no grouping so all output was in one big block)
- Continue when one dist fails (before it would stop after the first failure)

(Details: see commit message)

Example runs:
- no failure: https://github.com/bram-perl/perl5/actions/runs/3070729590/jobs/4960754562
- 1 failure: https://github.com/bram-perl/perl5/actions/runs/3070682723/jobs/4960655118
- 2 failures: https://github.com/bram-perl/perl5/actions/runs/3070645332/jobs/4960579443

### DIFF
--- a/Porting/test-dist-modules.pl
+++ b/Porting/test-dist-modules.pl
@@ -7,8 +7,12 @@ use File::Temp "tempdir";
 use ExtUtils::Manifest "maniread";
 use Cwd "getcwd";
 
+$|++;
+
 -f "Configure"
   or die "Expected to be run from a perl checkout";
+
+my $github_ci = $ENV{'GITHUB_SHA'} ? 1 : 0;
 
 my $manifest = maniread();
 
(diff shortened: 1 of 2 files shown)


Porting/test-dist-modules.pl
```diff
@@ -7,17 +7,32 @@
 use ExtUtils::Manifest "maniread";
 use Cwd "getcwd";
 
+$|++;
+
 -f "Configure"
   or die "Expected to be run from a perl checkout";
 
+my $github_ci = $ENV{'GITHUB_SHA'} ? 1 : 0;
+
```
tonycoz, line 16: Instead of making the continue on failure and github grouping output both depend on `$ENV{'GITHUB_SHA'}` wouldn't it be better to make the first be controlled by a command-line switch?

bram-perl:
Closing this PR since the changes are included in #20385.

PR #20316, opened 2022-09-19
Add builtin::load() as specified by RFC 0006

wolfsage:
See https://github.com/Perl/RFCs/blob/main/rfcs/rfc0006.md for more info.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -63,6 +63,30 @@ static void S_finish_export_lexical(pTHX)
     LEAVE;
 }
 
+XS(XS_builtin_load);
+XS(XS_builtin_load)
+{
+    dXSARGS;
+
+    if (items != 1) {
+        croak_xs_usage(cv, "arg");
+    }
+
+    SV *module = TOPs;
+
+    if (!SvOK(module)) {
+        Perl_croak(aTHX_ "Undefined argument to builtin::load");
+    }
+
+    module = newSVsv(module);
+
+    if (!SvPOK(module))
+        SvPV_force_nolen(module);
+
+    Perl_load_module(aTHX_ PERL_LOADMOD_NOIMPORT, module, NULL, NULL, NULL);
+
+    XSRETURN(1);
+}
 
 XS(XS_builtin_true);
 XS(XS_builtin_true)
@@ -500,6 +524,7 @@ static const struct BuiltinFuncDescriptor builtins[] = {
     { "builtin::floor",      &XS_builtin_func1_scalar, &ck_builtin_func1, OP_FLOOR      },
     { "builtin::is_tainted", &XS_builtin_func1_scalar, &ck_builtin_func1, OP_IS_TAINTED },
     { "builtin::trim",       &XS_builtin_trim,         &ck_builtin_func1, 0 },
+    { "builtin::load",       &XS_builtin_load,         &ck_builtin_func1, 0 },
 
     { "builtin::created_as_string", &XS_builtin_created_as_string, &ck_builtin_func1, 0 },
     { "builtin::created_as_number", &XS_builtin_created_as_number, &ck_builtin_func1, 0 },
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,4 +1,4 @@
-package builtin 0.008;
+package builtin 0.009;
 
 use strict;
 use warnings;
@@ -26,6 +26,7 @@ builtin - Perl pragma to import built-in utility functions
         trim
         is_tainted
         export_lexically
+        load
     );
 
 =head1 DESCRIPTION
@@ -317,6 +318,21 @@ Z<>
 This must be called at compile time; which typically means during a C<BEGIN>
 block. Usually this would be used as part of an C<import> method of a module,
 when invoked as part of a C<use ...> statement.
+
+=head2 load
+
+    load($module);
+
+This loads a named module from the inclusion paths (C<@INC>).  EXPR must be
+a string that provides a module name.  It cannot be omitted, and providing
+an invalid module name will result in an exception.
+
+The effect of C<load>-ing a module is the same as C<require>-ing, down to
+the same error conditions when the module does not exist, does not compile,
+or does not evalute to a true value.
+
+C<load> can't be used to require a particular version of Perl, nor can it
+be given a bareword module name as an argument.
 
 =head1 SEE ALSO
 
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -413,6 +413,40 @@ TODO: {
     is($HASH{key}, "val", 'Lexically exported hash is accessible');
 }
 
+# Load
+{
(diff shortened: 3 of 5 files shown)


builtin.c
```diff
@@ -63,6 +63,30 @@ static void S_finish_export_lexical(pTHX)
...
+    if (!SvOK(module)) {
+        Perl_croak(aTHX_ "Undefined argument to builtin::load");
+    }
+
+    module = newSVsv(module);
+
+    if (!SvPOK(module))
+        SvPV_force_nolen(module);
+
+    Perl_load_module(aTHX_ PERL_LOADMOD_NOIMPORT, module, NULL, NULL, NULL);
```
Leont, line 86: I think only one `NULL` value is necessary here (a second would have been if flags hadn't been `PERL_LOADMOD_NOIMPORT`).

builtin.c
```diff
@@ -63,6 +63,30 @@ static void S_finish_export_lexical(pTHX)
     LEAVE;
 }
 
+XS(XS_builtin_load);
+XS(XS_builtin_load)
+{
+    dXSARGS;
+
+    if (items != 1) {
+        croak_xs_usage(cv, "arg");
```
Leont, line 72: `s/arg/module/`?

builtin.c
```diff
@@ -63,6 +63,30 @@ static void S_finish_export_lexical(pTHX)
...
+    }
+
+    SV *module = TOPs;
+
+    if (!SvOK(module)) {
```
tonycoz, line 77: SvOK() isn't valid before checking for magic:
```
tony@venus:.../git/perl5$ cat ../20316.pl
use v5.36.0;
use builtin 'load';
my $x;
tie my $y, TestTie => \$x;
$x = "Storable";
eval { load $y; 1 } or print $@;
print ">>$y<<\n";

package TestTie;
sub TIESCALAR {
    bless $_[1], $_[0];
}
sub FETCH {
    ${$_[0]}
}

tony@venus:.../git/perl5$ ./perl -Ilib ../20316.pl
Built-in function 'builtin::load' is experimental at ../20316.pl line 6.
Undefined argument to builtin::load at ../20316.pl line 6.
>>Storable<<
tony@venus:.../git/perl5$
```

You can `SvGETMAGIC(module)` just before the SvOK() and then use newSVpv_nomg() to duplicate it while avoiding an extra magic call.

builtin.c
```diff
@@ -63,6 +63,30 @@ static void S_finish_export_lexical(pTHX)
...
+    }
+
+    module = newSVsv(module);
+
+    if (!SvPOK(module))
+        SvPV_force_nolen(module);
+
+    Perl_load_module(aTHX_ PERL_LOADMOD_NOIMPORT, module, NULL, NULL, NULL);
+
+    XSRETURN(1);
```
tonycoz, line 88: What value are you intending to return here?  Right now it is returning the module name.  (which confused me when I had an extra FETCH call.)

tonycoz (changes requested):

builtin.c
Leont: I assumed that was deliberate, but yeah it is a bit confusing and a comment about it may help.

leonerd: Perhaps you wanted `XSRETURN_YES`?

haarg: While I don't think the return value from `require` is particularly useful, I think I'd rather have this function return the same thing. It would be much easier to explain or document this function if it behaved the same as `require` aside from the input value.

wolfsage: I don't know that we can easily get at the return value of Perl_load_module?

Returning 1 I think makes the most sense, rather than "Whatever the package returned"...

leonerd: > I don't know that we can easily get at the return value of Perl_load_module?
> 
> Returning 1 I think makes the most sense, rather than "Whatever the package returned"...

That's not what `XSRETURN(1)` means.

It means "the return value is the 1 most recent items pushed to the SV stack". I.e. whatever was at the top of stack at the time.

I'd suggest it ought to just return "true", because that's what `require` would return for any module that has the "require on me just returns true" flag set - see other RFC/discussion.

wolfsage: Yeah I know the current behaviour is wrong. I'm talking about "What do we want" :)

leonerd: Ah; yes it is an unfortunate omission of the original RFC that we forgot to specify what the result should be. My vote is on a simple "yes". Any else with other opinions?

haarg: I would strongly prefer it to return the same value as `require`.

The RFC states:
> The effect of C<load>-ing a module is the same as C<require>-ing, down to
> the same error conditions when the module does not exist, does not compile,
> or does not evalute to a true value.

While this doesn't state anything about the return value, it implies that `load` should behave the same in every way except for its input.

haarg:
This doesn't seem to do any validation in the module name, as the RFC states it should. It can currently be given a string like `/foo/bar/Blah` and it will attempt to load `/foo/bar/Blah.pm`.

wolfsage:
> This doesn't seem to do any validation in the module name, as the RFC states it should. It can currently be given a string like `/foo/bar/Blah` and it will attempt to load `/foo/bar/Blah.pm`.

I think this part of the RFC needs more discussion. This correctly fails to find the module, same as if you passed it to require, which seems sufficient to me. Is adding more code and error strings useful at this point? Or could this also wait until we have proper exceptions, etc etc

haarg:
There are actually extra checks to prevent loading paths that start with `/` or contain `..`, so the worst cases are protected against.

I think it should reject anything that isn't a valid package name. That's what the RFC states.

builtin.c
ghost: As it is runtime thing returning a module name could be useful to load a class and call a method (chain) on it.
`load($class)->new->foo->bar;`

tonycoz:
`load()` or whatever it ends up being called isn't especially performance critical - is it essential that it is implemented as XS?

The validation of the module name would be simpler in Perl, the magic wouldn't be an issue, and returning the result of require would be much simpler too, if we went that way.

If we want these to be available automatically gv_fetchpvn_flags() could load builtin.pm just as it does the tie modules for %! and a few other variables, or give load() an XS stub.

The RFC specifies a "builtin" but that may be over specification in this case.

wolfsage:
As tony states, perhaps this doesn't need to be XS. I'm closing this to not block someone else from attempting to implement this at some point as I don't see getting to it soon

haarg:
Trying to implement this in perl will make it rather difficult to get errors reported from the correct location. This has always been a problem with Module::Runtime.

haarg:
Trying to write `load` in perl will make it hard to get errors reported from the correct location. This has always been a problem with Module::Runtime. CORE::require can't be called via goto, so you end up needing to capture and manipulate the error from require, or just live with it giving a useless error location.

PR #20318, opened 2022-09-19
eliminate an unused call to SvPVX() at the end of SvPVCLEAR_FRESH()

tonycoz:
The value was simply being discarded, which caused build warnings.

@richardleach

### DIFF
--- a/sv.h
+++ b/sv.h
@@ -2311,7 +2311,6 @@ that already have a PV buffer allocated, but no SvTHINKFIRST.
             *(SvEND(sv))= '\0';                         \
             (void)SvPOK_only_UTF8(sv);                  \
             SvTAINT(sv);                                \
-            SvPVX(sv);                                  \
         } STMT_END
 
 #define SvSHARE(sv) PL_sharehook(aTHX_ sv)


richardleach:
Thanks for pointing this out. `SvPVCLEAR` returns a char *, so I thought that `SvPVCLEAR_FRESH` should too. I don't understand why there are build warnings for `SvPVCLEAR_FRESH` but not `SvPVCLEAR` though? Is it literally because one is just a macro, but a function underlies the other?

demerphq:
On Mon, 19 Sept 2022, 14:16 Richard Leach, ***@***.***> wrote:

> Thanks for pointing this out. SvPVCLEAR returns a char *, so I thought
> that SvPVCLEAR_FRESH should too. I don't understand why there are build
> warnings for SvPVCLEAR_FRESH but not SvPVCLEAR though? Is it literally
> because one is just a macro, but a function underlies the other
>

Yes. Sort of.

It not that it's a macro, it's that it's wrapped in STMT_START, which means
it's actually a do {} while (0) loop, and thus it can't return a value.

You could rewrite it without the STMT_START and leave the semicolon off the
last statement, but that would be fragile and something like

if (foo) ...

would not work as expected. Usually if a macro like this assigns you need
to pass in the value it assigns to as a parameter so that it can continue
to be inside of the STMT_START wrapper.

Yves

>

demerphq:
On Mon, 19 Sept 2022, 14:36 demerphq, ***@***.***> wrote:

>
>
> On Mon, 19 Sept 2022, 14:16 Richard Leach, ***@***.***>
> wrote:
>
>> Thanks for pointing this out. SvPVCLEAR returns a char *, so I thought
>> that SvPVCLEAR_FRESH should too. I don't understand why there are build
>> warnings for SvPVCLEAR_FRESH but not SvPVCLEAR though? Is it literally
>> because one is just a macro, but a function underlies the other
>>
>
> Yes. Sort of.
>
> It not that it's a macro, it's that it's wrapped in STMT_START, which
> means it's actually a do {} while (0) loop, and thus it can't return a
> value.
>
> You could rewrite it without the STMT_START and leave the semicolon off
> the last statement,
>

Actually, scratch that. You can't make that work here either as it's a
multistatement macro.

Yves

>

tonycoz:
> Actually, scratch that. You can't make that work here either as it's a multistatement macro. Yves

Every statement is an expression statement, it could be turned into a comma expression.

Or it could just be an inline function, and avoid the possible macro argument multiple evaluation.

demerphq:
On Mon, 19 Sept 2022, 14:57 Tony Cook, ***@***.***> wrote:

> Actually, scratch that. You can't make that work here either as it's a
> multistatement macro. Yves
>
> Every statement is an expression statement, it could be turned into a
> comma expression.
>

Ah. I had burned that concept out of my mind. Good point.

> Or it could just be an inline function, and avoid the possible macro
> argument multiple evaluation
>

True. The only downside of that is asserts then become semi useless as they
report the inline function location and not where the function is used.
I've noticed that debugging issues with inline functions is a pita, at
least with gcc. It would be nice if we could have our cake and eat it too.
:-)

Yves

>

tonycoz:
#20322 replaces this

PR #20320, opened 2022-09-19
Bisect runner errors 20220919

jkeenan:
This pull request emerged out of discussions between @hvds and myself.  We would like to make use of `Porting bisect.pl` more user-friendly, specifically by providing more relevant error output.  Currently, on most failures the program issues a `pod2usage` call which dumps the (now extensive) documentation found in `Porting/bisect-runner.pl`.  In this p.r., we trim back the error output for the most frequent source of program failure by simply relying on the program's own call to `GetOptions()` to warn the user of command-line errors.

Please review.  Assuming there are no objections, I will hold off on merging this until after tomorrow's dev release.

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -54,35 +54,35 @@ my %defines =
 push @paths, qw(/usr/local/lib /lib /usr/lib)
         unless $linux64;
 
-unless(GetOptions(\%options,
-                  'target=s', 'make=s', 'jobs|j=i', 'crash', 'expect-pass=i',
-                  'expect-fail' => sub { $options{'expect-pass'} = 0; },
-                  'clean!', 'one-liner|e=s@', 'c', 'l', 'w', 'match=s',
-                  'no-match=s' => sub {
-                      $options{match} = $_[1];
-                      $options{'expect-pass'} = 0;
-                  },
-                  'force-manifest', 'force-regen', 'setpgrp!', 'timeout=i',
-                  'test-build', 'validate',
-                  'all-fixups', 'early-fixup=s@', 'late-fixup=s@', 'valgrind',
-                  'check-args', 'check-shebang!', 'usage|help|?', 'gold=s',
-                  'module=s', 'with-module=s', 'cpan-config-dir=s',
-                  'test-module=s', 'no-module-tests',
-                  'A=s@',
-                  'D=s@' => sub {
-                      my (undef, $val) = @_;
-                      if ($val =~ /\A([^=]+)=(.*)/s) {
-                          $defines{$1} = length $2 ? $2 : "\0";
-                      } else {
-                          $defines{$val} = '';
-                      }
-                  },
-                  'U=s@' => sub {
-                      $defines{$_[1]} = undef;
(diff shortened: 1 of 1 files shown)


demerphq:
On Mon, 19 Sept 2022, 17:41 James E Keenan, ***@***.***>
wrote:

> This pull request emerged out of discussions between @hvds
> <https://github.com/hvds> and myself. We would like to make use of Porting
> bisect.pl more user-friendly, specifically by providing more relevant
> error output. Currently, on most failures the program issues a pod2usage
> call which dumps the (now extensive) documentation found in Porting/
> bisect-runner.pl. In this p.r., we trim back the error output for the
> most frequent source of program failure by simply relying on the program's
> own call to GetOptions() to warn the user of command-line errors.
>
> Please review. Assuming there are no objections, I will hold off on
> merging this until after tomorrow's dev release
>
Makes sense to me, but i don't understand why you removed the pod2usage
entirely, although I do like that you removed the unless. If you changed it
to verbose=>0, it seems like it would be better. Also personally I'd expect
it to support a --man option which *would* show the full docs.

I'd expect it to be something like this:

GetOptions(...) or pod2usage(exitval=>255);
pod2usage(exitval=>255, verbose=>1) if $man;

See the sample usage in the Getopt::Long docs for using it together with
Pod::Usage.

Sorry I can't review this inline just now, it's a pain on a phone.

Cheers,
Yves

hvds:
> Makes sense to me, but i don't understand why you removed the pod2usage entirely, although I do like that you removed the unless. If you changed it to verbose=>0, it seems like it would be better. Also personally I'd expect it to support a --man option which *would* show the full docs.

Even the `verbose => 0` output is quite lengthy, and it is a 27-lines long list of 13 example usages rather than an actual definition of how it can be used.

It isn't obvious how best to serve both the experienced and the inexperienced user here, but I'd expect the typical inexperienced user already to be quite knowledgeable. In any case, if `GetOptions()` fails, I would expect its error message to be by far the most immediately useful information.

> Also personally I'd expect it to support a --man option which *would* show the full docs.

That's already available with both `--help` and `--usage`.

hvds (approved):

demerphq:
On Mon, 19 Sept 2022, 19:30 Hugo van der Sanden, ***@***.***>
wrote:

> Makes sense to me, but i don't understand why you removed the pod2usage
> entirely, although I do like that you removed the unless. If you changed it
> to verbose=>0, it seems like it would be better. Also personally I'd expect
> it to support a --man option which *would* show the full docs.
>
> Even the verbose => 0 output is quite lengthy, and it is a 27-lines long
> list of 13 example usages rather than an actual definition of how it can be
> used.
>
> It isn't obvious how best to serve both the experienced and the
> inexperienced user here, but I'd expect the typical inexperienced user
> already to be quite knowledgeable. In any case, if GetOptions() fails, I
> would expect its error message to be by far the most immediately useful
> information.
>
> Also personally I'd expect it to support a --man option which *would*
> show the full docs.
>
> That's already available with both --help and --usage.
>

Ok. I'm good then.

Yves

>

PR #20322, opened 2022-09-19
SvPVCLEAR_FRESH - change from macro to inline function

richardleach:
This is to prevent warnings due to the char * frequently being unused.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -1958,6 +1958,7 @@ Apd	|void	|sv_setpv	|NN SV *const sv|NULLOK const char *const ptr
 Apd	|void	|sv_setpvn	|NN SV *const sv|NULLOK const char *const ptr|const STRLEN len
 Apd	|void	|sv_setpvn_fresh|NN SV *const sv|NULLOK const char *const ptr|const STRLEN len
 Apd	|char  *|sv_setpv_bufsize|NN SV *const sv|const STRLEN cur|const STRLEN len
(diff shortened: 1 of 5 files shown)


richardleach:
This is an alternative to https://github.com/Perl/perl5/pull/20318

embed.fnc
```diff
@@ -1958,6 +1958,7 @@ Apd	|void	|sv_setpv	|NN SV *const sv|NULLOK const char *const ptr
 Apd	|void	|sv_setpvn	|NN SV *const sv|NULLOK const char *const ptr|const STRLEN len
 Apd	|void	|sv_setpvn_fresh|NN SV *const sv|NULLOK const char *const ptr|const STRLEN len
 Apd	|char  *|sv_setpv_bufsize|NN SV *const sv|const STRLEN cur|const STRLEN len
+Cipdx	|char  *|sv_setpv_freshbuf|NN SV *const sv
```
tonycoz: the "d" claims there is documentation, but there isn't, hence:
```
./miniperl -Ilib autodoc.pl
no docs for sv_setpv_freshbuf
```

richardleach: Yves wanted to check over the original SvPVCLEAR logic for a couple of points when he gets back, which might influence the contents of documentation (and the function). Is it acceptable to just drop the "d" for now and re-add later in this cycle?

tonycoz: Sure.

richardleach: Ok, dropped the "d" from _embed.fnc_ for now.

PR #20323, opened 2022-09-20
make expr parameter to newLOOPOP() NN, it was required anyway

tonycoz:
Coverity picked up that while we checked that expr was non-NULL in newLOOPOP(), the call to new_logop() dereferenced it anyway, the address of expr is passed to new_logop(), which sees it as firstp (so firstp == &expr), that is then derefed to first (first == expr) and then switches on first->op_type, hence dereferencing expr.

Since both callers always pass in an expr value, and the code would have crashed anyway with a NULL expr, make it required.


leonerd (approved):
LGTM

PR #20331, opened 2022-09-21
AELEMFASTLEX_STORE - support negative keys, skip unnecessary check

richardleach:
This commit:
* Adds support for negative keys, as per the original `AELEMFAST_LEX`
* Changes an `if()` check for a "_useless assignment to a temporary_" 
into an assert, since this condition should never be true when the
LHS is the result of an array fetch.

Both changes address feedback from @iabyn in #20063.


richardleach:

richardleach:
These two changes were suggested by and discussed with @iabyn. No-one else has commented in a month, so I'm merging them.

PR #20333, opened 2022-09-22
locale.c Change function to return a string, not print

khwilliamson:
This makes some print statements less awkward, and is more flexible, which will be used in future commits


jkeenan:
Build-time failure during CI.

PR #20341, opened 2022-09-23
Add mutexes to protect various libc calls

khwilliamson:
The vast majority of libc calls are thread-safe, but a substantial number aren't, even ones claimed to be re-entrant.  `mbrlen(),` for example, may require both the locale and environment to remain unchanged by other threads during its execution, or undefined behavior (like segfaults)  ensues.  I went through the POSIX specification looking for such calls, and also looked at likely suspects in my Ubuntu man pages.  It turns out that glibc violates POSIX in a few cases, allowing thread-unsafe behavior even when the Standard prohibits it.  A future commit will change perlxs to include my findings.  This series of commits implements them by defining mutexes to prohibit concurrent execution of multiple collaborating threads during critical sections.  Like all mutexes, code can simply not opt-in to using it, but this at least causes all the perl core to conform.  And the future perlxs changes will document what XS code should do.

Another example is most time functions.  `tzset()` is called behind the scenes in many of them.  It sets `tzname[],` which in many implementations is a global variable.  Thus, `tzset()` needs to be executed under control of a mutex until `tzname` is safely copied to a per-interpreter variable.  And hence, `asctime_r` isn't really thread safe without a mutex surrounding it.

On unthreaded perls, all these mutexes resolve to no-ops.

I certainly may have missed, in my search, items that should have mutexes.  Those can be added whenever any surfaces

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -1752,7 +1752,12 @@ my_tzset(pTHX)
 #endif
         fix_win32_tzenv();
 #endif
+    TZSET_LOCK;
     tzset();
+    TZSET_UNLOCK;
+    /* After the unlock, another thread could change things, but this is a
+     * problem with the Posix API generally, not Perl; and the result will be
+     * self-consistent */
 }
 
 MODULE = SigSet		PACKAGE = POSIX::SigSet		PREFIX = sig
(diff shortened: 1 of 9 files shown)


khwilliamson:
merged via e31c458ba106743af1db34149b866b2c5f9d406b

PR #20337, opened 2022-09-22
Deprecate smartmatch

book:
This is a follow-up from <http://nntp.perl.org/group/perl.perl5.porters/264678>.

This pull request replaces `experimental::smartmatch` by `deprecated::smartmatch`.
The goal is to be more explicit that smartmatch is a failed experiment, and its
use is discouraged.

This also remove some uses of `~~` in the core code.

PR #20343, opened 2022-09-23
Remove ~~ from autodoc.pl

book:
Following up on #20337, I've looked for and only found one use of `~~` in the core scripts and modules.

This patch replaces the `~~` with equivalent code.

### DIFF
--- a/autodoc.pl
+++ b/autodoc.pl
@@ -1369,11 +1369,12 @@ sub docout ($$$) { # output the docs for one function group
             my @base_args = $items[0]->{args}->@*;
             my $base_thread_context = $items[0]->{flags} =~ /T/;
             for (my $i = 1; $i < @items; $i++) {
-                no warnings 'experimental::smartmatch';
                 my $item = $items[$i];
+                my $args_are_equal = $item->{args}->@* == @base_args
+                  && !grep $item->{args}[$_] ne $base_args[$_], keys @base_args;
                 $need_individual_usage = 1
                                     if    $item->{ret_type} ne $base_ret_type
-                                    || ! ($item->{args}->@* ~~ @base_args)
+                                    || !  $args_are_equal
                                     ||   (   $item->{flags} =~ /T/
                                           != $base_thread_context);
                 my $ret_length = length $item->{ret_type};


autodoc.pl
```diff
@@ -1369,11 +1369,13 @@ ($$$)
             my @base_args = $items[0]->{args}->@*;
             my $base_thread_context = $items[0]->{flags} =~ /T/;
             for (my $i = 1; $i < @items; $i++) {
-                no warnings 'experimental::smartmatch';
                 my $item = $items[$i];
+                my $identical_args_count;
```
leonerd: Should probably initialise this to `= 0` in case there are none, to avoid the warning of "use of uninitialized value in !="

autodoc.pl
```diff
@@ -1369,11 +1369,13 @@ ($$$)
...
+                $item->{args}[$_] eq $base_args[$_] and $identical_args_count++
+                  for 0 .. $#base_args;
```
leonerd: FYI: This is fine as-is, but don't forget that `List::Util::all` exists. A neater way to write this could be

```perl
use List::Util 'all';
my $args_are_equal = @baseargs == $item->{args}->@* and
    all { $baseargs[$_] eq $item->{args}{$_} } keys @baseargs;
```

leonerd (changes requested):
Needs a `= 0`

autodoc.pl
book: I've actually picked your other suggestion and force pushed a new commit.

autodoc.pl
book: Thanks. I just stole that code!

book:
It seems I can't use `List::Util`, because `autodoc.pl` is run via `miniperl`.

```
./miniperl -Ilib autodoc.pl
Can't locate List/Util.pm in @INC (you may need to install the List::Util module) ...
```

leonerd:
> Turns out I can't use `List::Util`, because `autodoc.pl` is run via `miniperl`.

Ah; yeah that is unfortunate. You could inline a little helper version commented:

```perl
# Can't use List::Util because this runs via miniperl
sub any(&@) {
  my $code = shift;
  $code->($_) or return 0 for @_;
  return 1;
}
```

Or failing that, write the test in some other way. For a one-off it might be shorter to inline some other logic, such as your counting code from before.

leonerd (approved):
LGTM

PR #20344, opened 2022-09-23
locale.c: Save output of emulate_setlocale in buffer

khwilliamson:
Depending on Configuration and platform and details of the current request, the value returned could be pointing to a system static buffer, or be a temporary freeable upon LEAVE.  This commit standardizes it to a known per-interpreter buffer that can be properly freed at termination.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -683,17 +683,22 @@ S_less_dicey_bool_setlocale_r(pTHX_ const int cat, const char * locale)
 #  define common_emulate_setlocale(i, locale)                               \
                  emulate_setlocale_i(i, locale, YES_RECALC_LC_ALL, __LINE__)
 
-#  define setlocale_i(i, locale)     common_emulate_setlocale(i, locale)
+#  define setlocale_i(i, locale)                                            \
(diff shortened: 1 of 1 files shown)


tonycoz (commented):
I'm lost in a maze of macros, my main concerns are:
- whether it's possible for `save_to_buffer()` to be called with `buf == PL_stdize_locale_buf` (causing undefined behaviour in `Copy()`/`memcpy()`)
- whether the locale parameter can ever be `PL_stdize_locale_buf`, but then have the allocation `realloc()`ed out from under it.

khwilliamson:
I was planning to do this in a future commit, but now have added a commit that treats calling save_to_buffer() as a no-op when the string is the same address as the buffer.

IIUC your second comment.  I don't think that the locale parameter to emulate_setlocale() can ever be PL_stdize_locale_buf.  I could add an assert for that.

tonycoz:
I think the assert() is a good idea.

khwilliamson:
assert added

tonycoz (approved):

PR #20347, opened 2022-09-24
Reclaim unused bits in ${^WARNING_BITS}

book:
`no warnings 'experimental::foo';` is effectively a no-op if the `foo` feature is not experimental any more.

So, stop reserving bits in `${^WARNING_BITS}` for them, and silently do nothing when they are enabled or disabled via `use warnings` or `no warnings`.

Warnings turned into no-op in this commit:
* experimental::alpha_assertions
* experimental::bitwise
* experimental::isa
* experimental::lexical_subs
* experimental::postderef
* experimental::script_run
* experimental::signatures


haarg (approved):

Leont (approved):

regen/warnings.pl
```diff
@@ -1102,6 +1126,11 @@ =head2 "Negative warnings"
     no warnings  'experimental';
     use warnings 'experimental::somefeature';
 
+As experimental features become regular features of Perl,
+the corresponding warnings are not thrown anymore.
```
leonerd: Warnings aren't "thrown", they are "printed".

regen/warnings.pl
```diff
@@ -1102,6 +1126,11 @@ =head2 "Negative warnings"
...
+For backwards compatibility, it is still possible to disable or enable
```
leonerd: Hrm. I'd say "it is still possible to request turning on or off these warnings, but doing so does exactly nothing".

leonerd (changes requested):
Code looks fine. I'd suggest a couple of minor adjustments to wording of docs.

leonerd:
One of the CI runs fails a single test, `dist/IO/t/io-poll.t` but that doesn't look anyhow related to this change. I'd be minded to ignore it,.

regen/warnings.pl
book: Instead of:

> It is still possible to request turning on or off these warnings, but doing so does exactly nothing.

how about:

> It is still possible to request turning on or off these warnings, but doing so has no effect.

The avoidance of repetition might be a French stylistic thing, though.

leonerd (approved):
LGTM

Issue #16856, opened 2019-03-01
Add some const qualifiers


p5pRT:
Migrated from [rt.perl.org#133874](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133874) (status was 'open')

Searchable as RT133874$

p5pRT:
### From peter@eisentraut.org

This avoids compiler warnings if something including these header
files \(via perl\.h\) is compiled using \-Wcast\-qual\.

p5pRT:
### From peter@eisentraut.org

<details><summary>0001-Add-some-const-qualifiers.patch</summary>

```diff
From 06cec305baa2e694f98bdf98bf408ca90476196d Mon Sep 17 00:00:00 2001
From: Peter Eisentraut <peter@eisentraut.org>
Date: Fri, 1 Mar 2019 15:52:54 +0100
Subject: [PATCH] Add some const qualifiers

This avoids compiler warnings if something including these header
files (via perl.h) is compiled using -Wcast-qual.
---
 hv_func.h       | 10 +++++-----
 inline.h        |  4 ++--
 sbox32_hash.h   |  4 ++--
 stadtx_hash.h   |  2 +-
 zaphod32_hash.h |  2 +-
 5 files changed, 11 insertions(+), 11 deletions(-)

diff --git a/hv_func.h b/hv_func.h
index 111de93f1e..d6b8450f85 100644
--- a/hv_func.h
+++ b/hv_func.h
@@ -39,26 +39,26 @@
 # define __PERL_HASH_SEED_BYTES 16
 # define __PERL_HASH_STATE_BYTES 32
 # define __PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
-# define __PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_2_4_with_state((state),(U8*)(str),(len))
+# define __PERL_HASH_WITH_STATE(state,str,len) S_perl_hash_siphash_2_4_with_state((state),(const U8*)(str),(len))
 #elif defined(PERL_HASH_FUNC_SIPHASH13)
 # define __PERL_HASH_FUNC "SIPHASH_1_3"
 # define __PERL_HASH_SEED_BYTES 16
 # define __PERL_HASH_STATE_BYTES 32
 # define __PERL_HASH_SEED_STATE(seed,state) S_perl_siphash_seed_state(seed,state)
[...]

PR #20352, opened 2022-09-26
Free PL_splitstr properly with threads and duplicate flags

tonycoz:
Fixes a problem found while looking at #16856.

An alternative for the threads case might be to just set splitstr to NULL in child threads.

### DIFF
--- a/intrpvar.h
+++ b/intrpvar.h
@@ -496,7 +496,7 @@ PERLVAR(I, warnhook,	SV *)
 /* switches */
 PERLVAR(I, patchlevel,	SV *)
 PERLVAR(I, localpatches, const char * const *)
-PERLVARI(I, splitstr,	const char *, " ")
+PERLVARI(I, splitstr,	char *, NULL)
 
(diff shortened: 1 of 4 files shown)


jkeenan:
Possible reviewers for this p.r.: @khwilliamson @iabyn @richardleach anyone else?

richardleach:
I'm not terribly familiar with run switches, threads, or the save stack. In particular, I've no idea when the cloned interpreter SAVEPV will be freed. (Thread destruction?)

Those limitations aside, I can see what these commits are accomplishing and it looks fine AFAICT.

tonycoz:
PL_splitstr is released in perl_destruct() which is called for each thread, but when the thread is duplicated in perl_clone() the pointer itself is just copied, leading to the double-free the first patch fixes.

Raw use of threads with -F (as in my test case) seems unlikely to me, but I could see an end-user one-liner that uses some internally complex module running into this.

demerphq (approved):

PR #20353, opened 2022-09-26
ensure curlocales[] is initialized

tonycoz:
Coverity complains that the call to Safefree() now at line 5098 could be called with an uninitialized value for curlocales[i], which appears to be possible if the trial_locales loop just below this change fails to find a locale.

Fixes CID 184451


khwilliamson:
I believe this is a false positive, but the path of least resistance is to just do it.  This only happens at start-up and in Threads::Shared, I believe, so performance isn't an issue

Issue #20296, opened 2022-09-13
$SIG{__DIE__} does not get reliably triggered by compilation errors.


demerphq:
**Description**
During compilation Perl "saves up errors", so that it can show as many errors as possible to the user as it can. When it reaches a threshold (in theory 1 for syntax errors and 10 for other errors but in practice sometimes 11 depending on the error), it will trigger a Perl_croak and stop compilation, this in turn will feed the *list* of errors encountered to the `$SIG{__DIE__}` handler, however if it runs off the end of the code without reaching the maximum it will not call Perl_croak when inside of an eval, and instead will stuff the error directly into `$@`,  *bypassing* `$SIG{__DIE__}`.   Thus `$SIG{__DIE__}` is inconsistently called by compilation errors. 

**Steps to Reproduce**
This demonstrates code with two errors, neither of them syntax errors, which fail the compilation but do not trigger `$SIG{__DIE__}`:
```
$ perl -le'use strict; print $]; $SIG{__DIE__}= sub { print STDERR "in __DIE__"; $_[0] }; eval q($z+$y); print "\$@=",$@;'
5.037004
$@=Global symbol "$z" requires explicit package name (did you forget to declare "my $z"?) at (eval 1) line 1.
Global symbol "$y" requires explicit package name (did you forget to declare "my $y"?) at (eval 1) line 1.
```

This demonstrates code with one error, a syntax error, which fails compilation and does trigger `$SIG{__DIE__}`:
```
norole:yorton@oncidium:master:~/git_tree/HTML-Mason-1.59$ perl -le'use strict; print $]; $SIG{__DIE__}= sub { print STDERR "in __DIE__"; $_[0] }; eval q(1=); print "\$@=",$@;'
5.037004
in __DIE__
$@=syntax error at (eval 1) line 1, at EOF
```

And curiously die inside of a BEGIN triggers `$SIG{__DIE__}` twice:
```
perl -le'print $]; $SIG{__DIE__}= sub { print STDERR "in __DIE__"; $_[0] }; eval "BEGIN{die}"; print "\$@=",$@;'
5.037004
in __DIE__
in __DIE__
$@=Died at (eval 1) line 1.
BEGIN failed--compilation aborted at (eval 1) line 1.
```

**Expected behavior**
[...]

Issue #20346, opened 2022-09-23
BBC: Blead breaks MAUKE/Quote-Code and others


andk:
**Description**

Starting with v5.37.3-256-geb54d46f72 tests started to fail for MAUKE/Quote-Code-1.0102.tar.gz

Sample fail report: http://www.cpantesters.org/cpan/report/2afa4b80-39d7-11ed-a607-5f869d66406b

Git bisect for this fail lead me to

``` text
commit eb54d46f7264ff7af62c409d8a6ab984a5a34f57
Author: Yves Orton <demerphq@gmail.com>
Date:   Fri Aug 26 18:26:14 2022 +0200

    Stop parsing on first syntax error.
```

**Steps to Reproduce**

cpan -i MAUKE/Quote-Code-1.0102.tar.gz

**Expected behavior**

Should build, test and install the pkg

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 37 subversion 4) configuration:
  Commit id: f2582f5b18658f945a763f2edc110cdc7c5220e7
  Platform:
    osname=linux
    osvers=5.4.0-125-generic
    archname=x86_64-linux-thread-multi
    uname='linux k93focal 5.4.0-125-generic #141-ubuntu smp wed aug 10 13:42:03 utc 2022 x86_64 x86_64 x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93focal/v5.37.4/d119 -Dmyhostname=k93focal -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Duseithreads -Uuselongdouble -DEBUGGING=both'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2 -g'
[...]

Issue #20291, opened 2022-09-12
BBC: Blead Breaks HTML::Mason


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.37.4.

-----------------------------------------------------------------

BBC: Blead Breaks HTML::Mason

Please see http://matrix.cpantesters.org/?dist=HTML::Mason

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.37.4:

Configured by cpan at Mon Sep 12 10:08:25 EDT 2022.

Summary of my perl5 (revision 5 version 37 subversion 4) configuration:
  Commit id: dcf2c66b4ba2e50408461d7ac13a4782ba26bf1b
  Platform:
    osname=linux
    osvers=5.15.64-0-lts
    archname=x86_64-linux-thread-multi
    uname='linux cjg-alpine3 5.15.64-0-lts #1-alpine smp mon, 05 sep 2022 08:02:49 +0000 x86_64 linux '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall -Duseithreads'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='11.2.1 20220219'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
[...]

PR #20357, opened 2022-09-28
Consistent Compilation Error Behavior.

demerphq:
This is a set of patches to make our rules for errors during compilation consistent. Currently they are a bit random. 

When the stop on first error change was introduced I didnt understand about yyquit() so I did some hacky logic to handle syntax errors properly and it means that the way eval errors were reported wasnt consistent, this makes it more consistent.

This also ensures that `$SIG{__DIE__}` is called whenever the eval fails to compile without actually generating a croak(), and ensures that any errors produced by the eval are passed into the handler as they would be for other errors. 

See issue https://github.com/Perl/perl5/issues/20296


jkeenan:
On FreeBSD-12, using `clang10` as the compiler, I built my customary threaded `perl` from this branch.  I then tried to install all the CPAN modules which have been cited in the 2 BBC tickets: https://github.com/Perl/perl5/issues/20291 and https://github.com/Perl/perl5/issues/20346.
```
./bin/cpanm HTML::Mason Quote::Code BenchmarkAnything::Schema Mojolicious::Plugin::RenderCGI Quote::Ref Sub::Documentation
```
I should note that I wasn't really expecting this p.r. to clear up all those failures, as this is a situation where we know that some of the CPAN code is sub-optimal.  None of the above modules succeeded, but there was one new result: a test in Mojolicious itself failed, thereby preventing us from getting to its reverse dependencies.  That failure looks like this:
```
t/mojo/template.t .......................... 1/? 
#   Failed test 'exception in nested template'
#   at t/mojo/template.t line 778.
#                   'Bareword "bar" not allowed while "strict subs" in use at template line 1.
# Context:
#   143:   # Use a local stack trace for compile exceptions
#   144:   my $compiled = $self->compiled;
#   145:   unless ($compiled) {
#   146:     my $code = $self->_compile->code;
#   147:     monkey_patch $self->namespace, '_escape', $self->escape;
#   148:     return Mojo::Exception->new($@)->inspect($self->unparsed, $code)->trace->verbose(1)
#   149:       unless $compiled = eval $self->_wrap($code, @_);
#   150:     $self->compiled($compiled);
#   151:   }
#   152: 
#   153:   # Use a real stack trace for normal exceptions
# Traceback (most recent call first):
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 148, in "Mojo::Template::__ANON__"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
#   File "template", line 3, in "Mojo::Template::render"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "Mojo::Template::Sandbox::__ANON__"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "(eval)"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
#   File "t/mojo/template.t", line 769, in "Mojo::Template::render"
# '
#     doesn't match '(?^u:test\n\nBareword "bar".+in use at template line 1\.)'
# Looks like you failed 1 test of 226.
t/mojo/template.t .......................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/226 subtests 
```
At least as of two days ago, this test is passing in blead.  A test failure in Mojolicious is certainly as much of a blocker as a test failure in HTML::Mason.  Needs more investigation.

bram-perl:
Relevant issue: https://github.com/Perl/perl5/issues/20296

demerphq:
@jkeenan the underlying nature of this problem is that what happens in a given string eval is only incidentally deterministic, and from the point of view of a single error is non-deterministic. Thus whether a given eval error will trigger someones `$SIG{__DIE__}` depends on the larger context of the code being evaled and the presence of *other* errors. 

This patch makes the behavior "strongly" deterministic, in that the `$SIG{__DIE__}` will fire for ALL eval STRING errors. 

This will very likely surface existing bugs in `$SIG{__DIE__}` related code, bugs that could be triggered by supplying just the right different test case than the ones they were using. It is an expected outcome of fixing indeterminate behavior. The best we can do is help people fix their code to deal with it and make things strongly deterministic in the future. I guess we could introduce a round of warnings, and then fix it in a following cycle.

Anyway, ill check out the issue that @bram-perl referenced.

demerphq:
@jkeenan fwiw:

```
t/mojo/template.t .......................... 1/? 
#   Failed test 'exception in nested template'
#   at t/mojo/template.t line 778.
#                   'Bareword "bar" not allowed while "strict subs" in use at template line 1.
# Context:
#   143:   # Use a local stack trace for compile exceptions
#   144:   my $compiled = $self->compiled;
#   145:   unless ($compiled) {
#   146:     my $code = $self->_compile->code;
#   147:     monkey_patch $self->namespace, '_escape', $self->escape;
#   148:     return Mojo::Exception->new($@)->inspect($self->unparsed, $code)->trace->verbose(1)
#   149:       unless $compiled = eval $self->_wrap($code, @_);
#   150:     $self->compiled($compiled);
#   151:   }
#   152: 
#   153:   # Use a real stack trace for normal exceptions
# Traceback (most recent call first):
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 148, in "Mojo::Template::__ANON__"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
#   File "template", line 3, in "Mojo::Template::render"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "Mojo::Template::Sandbox::__ANON__"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "(eval)"
#   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
#   File "t/mojo/template.t", line 769, in "Mojo::Template::render"
# '
#     doesn't match '(?^u:test\n\nBareword "bar".+in use at template line 1\.)'
```

seems to suggest their test pattern is to restrictive. If they remove the "test\n\n" prefix they would get a test that pases in both cases.

jkeenan:
> @jkeenan fwiw:
> 
> ```
> t/mojo/template.t .......................... 1/? 
> #   Failed test 'exception in nested template'
> #   at t/mojo/template.t line 778.
> #                   'Bareword "bar" not allowed while "strict subs" in use at template line 1.
> # Context:
> #   143:   # Use a local stack trace for compile exceptions
> #   144:   my $compiled = $self->compiled;
> #   145:   unless ($compiled) {
> #   146:     my $code = $self->_compile->code;
> #   147:     monkey_patch $self->namespace, '_escape', $self->escape;
> #   148:     return Mojo::Exception->new($@)->inspect($self->unparsed, $code)->trace->verbose(1)
> #   149:       unless $compiled = eval $self->_wrap($code, @_);
> #   150:     $self->compiled($compiled);
> #   151:   }
> #   152: 
> #   153:   # Use a real stack trace for normal exceptions
> # Traceback (most recent call first):
> #   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 148, in "Mojo::Template::__ANON__"
> #   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
> #   File "template", line 3, in "Mojo::Template::render"
> #   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "Mojo::Template::Sandbox::__ANON__"
> #   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 160, in "(eval)"
> #   File "/usr/home/jkeenan/.cpanm/work/1664387589.71629/Mojolicious-9.27/blib/lib/Mojo/Template.pm", line 163, in "Mojo::Template::process"
> #   File "t/mojo/template.t", line 769, in "Mojo::Template::render"
> # '
> #     doesn't match '(?^u:test\n\nBareword "bar".+in use at template line 1\.)'
> ```
> 
> seems to suggest their test pattern is to restrictive. If they remove the "test\n\n" prefix they would get a test that pases in both cases.

Okay, I have prepared a pull request for Mojolicious which I will submit once this p.r. is merged into blead.

haarg:
I believe the test failure in Mojo is pointing to a real bug in this PR. Test case:
```
perl -e'use strict; use warnings; $SIG{__DIE__} = sub { die $@ }; eval { eval "bar"; warn "after strict error in eval [$@]"; }; if ($@) { print "outer eval failed [$@]\n" }'
```
The inner eval will fail during compilation due to strict. We should expect that the "after strict error" message will be printed. That is what happens on stable and blead. With this PR, we instead get the "outer eval failed" message. If you remove the `__DIE__` handler, it produces the "after strict error" message as expected.

haarg:
Also, the behavior is not consistent for all compile time errors:
```
perl -e'use strict; use warnings; $SIG{__DIE__} = sub { die $@ }; eval { eval "]"; warn "after compile error in eval [$@]"; }; if ($@) { print "outer eval failed [$@]\n" }'
```
This is a syntax error rather than a strict error. It consistently produces the "after compile error" message, with or without this PR.

bram-perl:
> Anyway, ill check out the issue that @bram-perl referenced.

I didn't mean it as something to check out, I **assumed** this PR was fixing #20296 but I might have been mistaken..
(I have not yet looked at the changes in the PR)

demerphq:
@haarg i will check into your test case. I am focusing on getting https://github.com/Perl/perl5/pull/20398 sorted out first tho. I will follow up on this one once that has been pushed. (I suspect they might be related subtly)

demerphq:
@bram-perl I got confused, i didnt realize that was my original bug report.

demerphq:
@haarg: as regards to your also, yes, indeed the whole point of this patch is to make this consistent. I am now testing a fix for the test case you showed, thanks for that!

demerphq:
@haarg i added your test case to my patch, and reordered when the sig die handler is called so it is handled in the right scope. Thanks for the pointer.

haarg:
I can confirm that this now handles my test cases and the Mojo test correctly.

demerphq:
@haarg @jkeenan am I correct in thinking this is good to go now?

demerphq:
I am going to merge this so it can be part of  5.37.9 and so we can shake out all the bugs related to early compilation halts in one release (5.38). We already have multiple such patches in this sequence. So I am merging this now.

PR #20364, opened 2022-09-30
Add comments to perly.y to explain token ordering

leonerd:
Explains why KW_PACKAGE and KW_USE_or_NO appear to have the package name and version barewords in the wrong order.

It took me a very long time to work this out so I'd like to think this isn't wasted effort :)


wolfsage:
Might be useful to mention yylex() is the one popping them back off in reverse order, but otherwise seems fine

wolfsage (approved):

PR #20365, opened 2022-10-01
Enable `use feature 'module_true'`

chromatic:
Per RFC 18, whenever `use feature 'module_true';` is enabled in a scope, any file required with `require` has an implicit return value of true and will not trigger the "did not return a true value" error condition.

Note that this implementation also propagates all features of the previous COP into the ex-nextstate COP created when leaving a block. This makes it possible to find and examine the last COP of a block (even when nulled) to discover any hints or features in effect at the end of the block.

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -4,8 +4,7 @@
 # Any changes made here will be lost!
 
 package feature;
-
-our $VERSION = '1.76';
+our $VERSION = '1.77';
 
 our %feature = (
     fc                      => 'feature_fc',
@@ -20,6 +19,7 @@ our %feature = (
     evalbytes               => 'feature_evalbytes',
     signatures              => 'feature_signatures',
     current_sub             => 'feature___SUB__',
+    module_true             => 'feature_module_true',
     refaliasing             => 'feature_refaliasing',
     postderef_qq            => 'feature_postderef_qq',
     unicode_eval            => 'feature_unieval',
@@ -37,8 +37,8 @@ our %feature_bundle = (
     "5.23"    => [qw(bareword_filehandles current_sub evalbytes fc indirect multidimensional postderef_qq say state switch unicode_eval unicode_strings)],
     "5.27"    => [qw(bareword_filehandles bitwise current_sub evalbytes fc indirect multidimensional postderef_qq say state switch unicode_eval unicode_strings)],
     "5.35"    => [qw(bareword_filehandles bitwise current_sub evalbytes fc isa postderef_qq say signatures state unicode_eval unicode_strings)],
-    "5.37"    => [qw(bitwise current_sub evalbytes fc isa postderef_qq say signatures state unicode_eval unicode_strings)],
-    "all"     => [qw(bareword_filehandles bitwise current_sub declared_refs defer evalbytes extra_paired_delimiters fc indirect isa multidimensional postderef_qq refaliasing say signatures state switch try unicode_eval unicode_strings)],
+    "5.37"    => [qw(bitwise current_sub evalbytes fc isa module_true postderef_qq say signatures state unicode_eval unicode_strings)],
+    "all"     => [qw(bareword_filehandles bitwise current_sub declared_refs defer evalbytes extra_paired_delimiters fc indirect isa module_true multidimensional postderef_qq refaliasing say signatures state switch try unicode_eval unicode_strings)],
     "default" => [qw(bareword_filehandles indirect multidimensional)],
 );
 
@@ -867,6 +867,14 @@ The complete list of accepted paired delimiters as of Unicode 14.0 is:
  🢫  🢪    U+1F8AB, U+1F8AA RIGHT/LEFTWARDS FRONT-TILTED SHADOWED WHITE
                           ARROW
 
+=head2 The 'module_true' feature
+
+This feature removes the need to return a true value at the end of a module
+loaded with C<require> or C<use>. Any errors during compilation will cause
+failures, but reaching the end of the module when this feature is in effect
+will prevent C<perl> from throwing an exception that the module "did not return
+a true value".
+
 =head1 FEATURE BUNDLES
 
 It's possible to load multiple features together, using
@@ -944,8 +952,8 @@ The following feature bundles are available:
             state unicode_eval unicode_strings
 
   :5.38     bitwise current_sub evalbytes fc isa
-            postderef_qq say signatures state
-            unicode_eval unicode_strings
+            module_true postderef_qq say signatures
+            state unicode_eval unicode_strings
 
 The C<:default> bundle represents the feature set that is enabled before
 any C<use feature> or C<no feature> declaration.
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4761,6 +4761,7 @@ PP(pp_entereval)
 
 PP(pp_leaveeval)
 {
+    dSP;
     SV **oldsp;
     U8 gimme;
     PERL_CONTEXT *cx;
@@ -4804,9 +4805,16 @@ PP(pp_leaveeval)
         }
         PL_curcop = old_pl_curcop;
 
-        failed = !(gimme == G_SCALAR
-                        ? (module_true || SvTRUE_NN(*PL_stack_sp))
-                        : PL_stack_sp > oldsp);
+        /* failed stays false */
+        if (module_true) {
+            POPs;
+            PUSHs(&PL_sv_yes);
+        }
+        else {
+            failed = !(gimme == G_SCALAR
+                            ? SvTRUE_NN(*PL_stack_sp)
+                            : PL_stack_sp > oldsp);
+        }
     }
 
     /* the cx_popeval does a leavescope, which frees the optree associated
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3763,7 +3763,6 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
     }
 
     /* Compilation successful. Now clean up */
-
     LEAVE_with_name("evalcomp");
 
     CopLINE_set(&PL_compiling, 0);
@@ -4766,11 +4765,11 @@ PP(pp_leaveeval)
     U8 gimme;
     PERL_CONTEXT *cx;
     OP *retop;
-    int failed;
+    int failed = 0;
     CV *evalcv;
     bool keep;
-
     PERL_ASYNC_CHECK();
+    int module_true = 0;
 
     cx = CX_CUR();
     assert(CxTYPE(cx) == CXt_EVAL);
@@ -4778,11 +4777,6 @@ PP(pp_leaveeval)
     oldsp = PL_stack_base + cx->blk_oldsp;
     gimme = cx->blk_gimme;
 
-    /* did require return a false value? */
-    failed =    CxOLD_OP_TYPE(cx) == OP_REQUIRE
-             && !(gimme == G_SCALAR
-                    ? SvTRUE_NN(*PL_stack_sp)
-                    : PL_stack_sp > oldsp);
 
     if (gimme == G_VOID) {
         PL_stack_sp = oldsp;
@@ -4791,6 +4785,29 @@ PP(pp_leaveeval)
     }
     else
         leave_adjust_stacks(oldsp, oldsp, gimme, 0);
+
+    /* did require return a false value?
+       make sure to initialize failed, whatever circumstance this was called
+       but check if use feature 'module_true' is enabled where it matters
+     */
+    if (CxOLD_OP_TYPE(cx) == OP_REQUIRE) {
+        COP *old_pl_curcop = PL_curcop;
+        if (PL_op->op_flags & OPf_KIDS && OP_TYPE_IS_OR_WAS(PL_op, OP_LEAVEEVAL)) {
+            const OP *kid = cLISTOPx(cUNOPx(PL_op)->op_first)->op_first;
+
+            for (; kid; kid = OpSIBLING(kid)) {
+                if (OP_TYPE_IS_OR_WAS(kid, OP_NEXTSTATE)) {
+                    PL_curcop = cCOPx(kid);
+                    module_true = FEATURE_MODULE_TRUE_IS_ENABLED;
+                }
+            }
+        }
+        PL_curcop = old_pl_curcop;
+
+        failed = !(gimme == G_SCALAR
+                        ? (module_true || SvTRUE_NN(*PL_stack_sp))
+                        : PL_stack_sp > oldsp);
+    }
 
     /* the cx_popeval does a leavescope, which frees the optree associated
      * with eval, which if it frees the nextstate associated with
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4807,7 +4807,7 @@ PP(pp_leaveeval)
         PL_curcop = old_pl_curcop;
 
         if (module_true) {
-            POPs;
+            (void)POPs;
             PUSHs(&PL_sv_yes);
             failed = false;
         }
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4788,7 +4788,7 @@ PP(pp_leaveeval)
     /* did require return a false value?
        check if use feature 'module_true' is enabled where it matters
      */
-    bool failed = false;
+    int failed = 0;
 
     if (CxOLD_OP_TYPE(cx) == OP_REQUIRE) {
         bool module_true = false;
@@ -4809,7 +4809,7 @@ PP(pp_leaveeval)
         if (module_true) {
             (void)POPs;
             PUSHs(&PL_sv_yes);
-            failed = false;
+            failed = 0;
         }
         else {
             failed = !(gimme == G_SCALAR
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4766,11 +4766,9 @@ PP(pp_leaveeval)
     U8 gimme;
     PERL_CONTEXT *cx;
     OP *retop;
-    int failed = 0;
     CV *evalcv;
     bool keep;
     PERL_ASYNC_CHECK();
-    int module_true = 0;
 
     cx = CX_CUR();
     assert(CxTYPE(cx) == CXt_EVAL);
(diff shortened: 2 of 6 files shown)


chromatic:
I'm not sure if this PR should include changes to `B::Concise`, so I'm happy to receive feedback either way.

leonerd:
I should take a look at this as hopefully we can move forward with the RFC soon

haarg:
The PR as written is not clear about what the return value from require should be. @rjbs said in an email to p5p that it was adjusted "based on suggestions from Dave M". But the [last email from Dave](https://www.nntp.perl.org/group/perl.perl5.porters/2022/06/msg264198.html) agreed with my argument that when impacted by this feature, require should always return a simple true value.

leonerd:
@haarg 
> The PR as written is not clear about what the return value from require should be.

Ahyes; it doesn't appear that this PR does anything about that. I'd imagine it should be fairly easy to fix with

```
if(module_true) {
  POPs;
  PUSHs(&PL_sv_yes);
}
```

chromatic:
> Ahyes; it doesn't appear that this PR does anything about that. I'd imagine it should be fairly easy to fix with

I considered that and am happy to make the change.

I didn't do that (yet), because I wondered what users should expect if `module_true` is in effect *and* there is already a return value (especially if it evaluates to a non-true value). I think we can manage expectations with proper documentation, but I didn't want to decide that unilaterally without discussion.

leonerd:
> I didn't do that (yet), because I wondered what users should expect if `module_true` is in effect _and_ there is already a return value (especially if it evaluates to a non-true value). I think we can manage expectations with proper documentation, but I didn't want to decide that unilaterally without discussion.

The module in question would have to have opted-in to `module_true` anyway, so it would be free to make that decision. I'd say it's fine to force a "PL_sv_yes" result on those cases, to avoid surprises later on.

chromatic:
All tests pass for me with this change. I'm open to revising the logic of the conditionals for the sake of clarity.

pp_ctl.c
```diff
@@ -4762,27 +4761,23 @@ PP(pp_entereval)
...
     U8 gimme;
     PERL_CONTEXT *cx;
     OP *retop;
-    int failed;
+    int failed = 0;
     CV *evalcv;
     bool keep;
-
     PERL_ASYNC_CHECK();
+    int module_true = 0;
```
leonerd: Since we're using C99 now don't forget this can be `bool` - we're only ever using it in a truthy capacity so it's better to be `bool` and assign `true` or `false`

pp_ctl.c
```diff
@@ -4762,27 +4761,23 @@ PP(pp_entereval)
 
 PP(pp_leaveeval)
 {
+    dSP;
     SV **oldsp;
     U8 gimme;
     PERL_CONTEXT *cx;
     OP *retop;
-    int failed;
+    int failed = 0;
```
leonerd: As C99 permits we can make this a `bool`, we can also move it lower down to just before it's first needed.

pp_ctl.c
```diff
@@ -4792,6 +4787,36 @@ PP(pp_leaveeval)
...
+            for (; kid; kid = OpSIBLING(kid)) {
+                if (OP_TYPE_IS_OR_WAS(kid, OP_NEXTSTATE)) {
+                    PL_curcop = cCOPx(kid);
+                    module_true = FEATURE_MODULE_TRUE_IS_ENABLED;
+                }
+            }
+        }
+        PL_curcop = old_pl_curcop;
+
+        /* failed stays false */
```
leonerd: I'd bring the `failed` variable down here. Start with

```c
bool failed;
```

then inside the true branch, put `failed = false` so it's clear both sides of the if will assign a value into it.

leonerd (commented):
Overall seems a good direction. I see you've adjusted the code to overwrite whatever actual return value from the module and replace it with a simple PL_sv_yes. Can you add a unit test for this? Try a module that specifically does something like `"some_true_value"` at the end, and assert that the `require` of that module does not yield that string, but simply the boolean true value.

pp_ctl.c
chromatic: Good point. Done.

pp_ctl.c
chromatic: Also done.

chromatic:
> Overall seems a good direction. I see you've adjusted the code to overwrite whatever actual return value from the module and replace it with a simple PL_sv_yes. Can you add a unit test for this? Try a module that specifically does something like `"some_true_value"` at the end, and assert that the `require` of that module does not yield that string, but simply the boolean true value.

I've added this test, but it fails now. I think there's some additional stack manipulation necessary here that I haven't figured out yet. Any suggestions?

leonerd:
@chromatic 
> I've added this test, but it fails now. I think there's some additional stack manipulation necessary here that I haven't figured out yet. Any suggestions?

The more I look into it the less sure I am of what's going on. In particular, I'm not even sure that `pp_leaveeval` is the right place to be putting this code. I've put lots of debug warnings into that code to watch what's going on, and it doesn't get triggered by:

```
$ cat module-ending-string.pm 
#!/usr/bin/perl
use v5.37.5;
use feature 'module_true';
"a true value here"

$ ./perl -Mwarnings -Ilib -E 'use lib "."; my $ret = require "module-ending-string.pm"; 1'
```

Moreover because I don't think these get eval-wrapped:

```
$ ./perl -Mwarnings -Ilib -MO=Concise -cE 'use lib "."; my $ret = require "module-ending-string.pm"; 1'
7  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 84 -e:1) v:%,us,{,fea=15 ->3
5     <1> padsv_store[$ret:84,85] vKS/LVINTRO ->6
4        <1> require sK/1 ->5
3           <$> const[PV "module-ending-string.pm"] s ->4
-        <0> ex-padsv sRM*/LVINTRO ->5
6     <;> nextstate(main 85 -e:1) v:%,us,{,fea=15 ->7
-     <0> ex-const v ->7
-e syntax OK
```

That's just a `pp_require` - so surely the code to check what's going on with the `module_true` feature needs to appear somewhere in the code called from that, and not `pp_leaveeval`.

I suspect the fact your tests are passing is more to do with the fact that each individual test is eval-wrapped and manages to otherwise make something pass, in a way that isn't directly related to the `require` itself.

demerphq:
@chromatic / @leonerd - this behavior is triggered by the `S_pop_eval_context_maybe_croak()` call in `pp_leaveeval`. The "failed" parameter being 1 triggers this error.

pp_ctl.c
demerphq: That isn't correct, it should not be a bool, it is an int. failed is the 'action' parameter to S_pop_eval_context_maybe_croak().

demerphq:
@chromatic it feels weird giving this feedback as you are the reason I hack perl core. Don't know if you remember "use caution". But whatever. 

I don't think this is the right way to deal with this. I think you need to deal with this through the eval context data and S_pop_eval_context_maybe_croak(). If you embed the feature state into the context when a context is pushed, then when the context is popped in action mode == 1 then you can use the eval context data entirely to determine if you are in a require and to determine if you should be picky about the true value or not.  I haven't closely looked at your patch except that given what I know right this second it doesn't at all look like what I expect. I would expect most of this logic to be in S_pop_eval_context_maybe_croak() and the related logic which pushes an eval context (doeval_compile and related code). I almost am willing to try to write this for you if you want... I happen to have been working on related logic recently.

chromatic:
@leonerd I agree about the `require` behavior, but I disagree that the `eval` wrapping is masking the right behavior. Try your code examples with `use` statements or `-M` flags.

@demerphq I welcome the help. This implementation is tricky for a couple of reasons:

- we want the effect to take place *after* leaving a compilation unit
- there may be no COPs accessible at the end of the compilation unit (if there are no expressions or non-declarations inside the compiled code; a module that has only `sub` statements, for example)
- the return value of the compiled module has to be manipulated from outside the compilation process

Essentially, any implementation has to ask "was the `module_true` feature enabled at the end of the outermost scope of a compiled file", and I think that means reconstructing the context in which compilation ended.

The `true` implementation on CPAN handles this with global variables and toggles, so it looks a lot easier than doing it in the core.

chromatic:
Never mind this most recent commit; I made a typo in the test and thought I'd found an interesting compiler edge case.

demerphq:
@chromatic and @leonerd I played around with this a fair bit this morning and I have gotten /some/ understanding of /some/ of the problem. I have pushed yves/chr_rfc-18-module-true to the master repo with my working code.  With that built (ignore the copious debug it produces during build) you can do this:

```
./perl -Dx -Ilib t/comp/require.t
```

And you will get a bunch of output like this:

```
in require
has kids
is leave
checking for feature
     
262  nextstate COP(0x557669c6f278) ===> 263 [const 0x557669c6f318]
     PARENT ===> 261 [lineseq 0x557669c6f238]
     FLAGS = (VOID,SLABBED,MORESIB)
     LINE = 7
     PACKAGE = "feature"
     SEQ = 4294967278
checking 557669c6f278
checking 557669c6fe30
checking 557669c74600
checking 557669c74388
checking 557669c74110
checking 557669c73e98
checking 557669c73c20
checking 557669c739a8
checking 557669c73730
checking 557669c734b8
checking 557669c73240
checking 557669c7aba8
checking 557669c7a930
checking 557669c7a6b8
checking 557669c7a440
checking 557669c7a1c8
checking 557669c79f50
checking 557669c79cd8
checking 557669c79a60
checking 557669c797e8
checking 557669c79570
checking 557669c792f8
checking 557669c79080
checking 557669c78e08
checking 557669c78b90
checking 557669c78918
checking 557669c78678
checking 557669c78448
checking 557669c782f8
checking 557669c781a8
checking 557669c77df0
checking 557669c77ca0
checking 557669c77c00
module_true is not enabled
SV = IV(0x557669c7bf38) at 0x557669c7bf48
  REFCNT = 1
  FLAGS = (IOK,READONLY,PROTECT,pIOK)
  IV = 1
     
686  leaveeval UNOP(0x557669c67d50) ===> [0x0]
     PARENT ===> [0x0]
     TARG = 1
     FLAGS = (SCALAR,KIDS,SLABBED,SAVEFREE)
     PRIVATE = (REFC,0x1)
     REFCNT = 1
     |   
687  +--lineseq LISTOP(0x557669c67d88) ===> 686 [leaveeval 0x557669c67d50]
         FLAGS = (SCALAR,KIDS,PARENS,SLABBED,SAVEFREE)
         |   
688      +--nextstate COP(0x557669c67dc8) ===> 689 [pushmark 0x557669c67e68]
         |   FLAGS = (VOID,SLABBED,MORESIB)
         |   LINE = 3
         |   PACKAGE = "bleah"
         |   SEQ = 119
         |   
690      +--return LISTOP(0x557669c67e28) ===> 686 [leaveeval 0x557669c67d50]
             FLAGS = (UNKNOWN,KIDS,SLABBED)
             |   
689          +--pushmark OP(0x557669c67e68) ===> 250 [const 0x557669c67ec8]
             |   FLAGS = (SCALAR,SLABBED,MORESIB)
             |   
250          +--const SVOP(0x557669c67ec8) ===> 690 [return 0x557669c67e28]
                 FLAGS = (SCALAR,SLABBED)
                 SV = PV("some_true_value"\0)
in require
has kids
is return
checking for feature
     
688  nextstate COP(0x557669c67dc8) ===> 689 [pushmark 0x557669c67e68]
     PARENT ===> 687 [lineseq 0x557669c67d88]
     FLAGS = (VOID,SLABBED,MORESIB)
     LINE = 3
     PACKAGE = "bleah"
     SEQ = 119
checking 557669c67dc8
module_true is not enabled
SV = PV(0x557669c57c60) at 0x557669c62800
  REFCNT = 1
  FLAGS = (POK,IsCOW,READONLY,PROTECT,pPOK)
  PV = 0x557669c75580 "some_true_value"\0
  CUR = 15
  LEN = 17
  COW_REFCNT = 0
not in require
SV = IV(0x557669c5c390) at 0x557669c5c3a0
  REFCNT = 1
  FLAGS = (IOK,READONLY,PROTECT,pIOK)
  IV = 22
ok 2 - use feature 'module_true' works even if module has an explicit true return value ()
not ok 3 - use feature 'module_true' replaces any explicit return value with a true value () <some_true_value>
```

what this shows is that with the test code using `return "some_true_value"` the last op that is executed is the OP_RETURN not OP_LEAVEEVAL, pp_return tail-calls pp_leaveeval, I guess as an optimization. In that case the existing opcode tree munging that @chromatic was doing doesnt work. The -Dx debugging flag is helpful for understanding what is happening. So I changed the code to deal with OP_RETURN differently, and it find what I think should be the correct COP, but when it runs the check to see if the feature is running it does not return true. I do not know why yet. 

A couple of minor notes. I pretty much unrolled all of the changes @chromatic made to the 'failed' variable, and move the logic to check whether the feature is enabled up. I do not think it is correct moving that logic down, it needs to happen before any stack munging happens. 

I also have removed the logic to munge the stack, while I was debugging finding out if the feature is enabled.

I have to do chores now, but hopefully this is helpful to someone more familiar with the optree logic.

chromatic:
Ah, the `OP_RETURN` case makes sense. I don't see your commits though @demerphq : is [this branch in the perl5 repo](https://github.com/Perl/perl5/compare/blead...yves/chr_rfc-18-module-true) the right one?

demerphq:
Sorry fat fingered what i pushed there. Sorry. I have pushed the right thing now.

demerphq:
I did some more debugging. It seems that when I change the failing test code from `use feature "module_true";` to `use v5.37;` it does detect the feature with my branch. So the problem is there. My latest push to the branch passes test (With lots of debug output). I will dig into why `use feature "module_load";` does not work, but `use v5.37;` does.

demerphq:
BTW, regarding "correcting the return"? What do you want to do in list context? My code fixing a failing list context require by pushing a PL_sv_yes, but it feels icky. How do you standardize a list return? Replace the entire list with a single PL_sv_yes?

demerphq:
And now pushed without the debug crap. Still trying to understand why `use feature "module_load";` does not trigger this behavior, but `use v5.37;` does.

demerphq:
Ive pushed further debugging to see what is happening with the features bits. I dont get it, i dont see a single cop with a non-zero cop_features bits, and i havent found anywhere that copies the cop_features bits out of PL_compiling into the PL_curcop. Is this the only feature that affects behavior at run time and not compile time? Could it be that we simply overlooked copying this data over? It seems possible if this is the first that changes behavior at run time.

demerphq:
Annnd fixed in my branch I believe. I had to add a patch to ensure that the cop_features data was available at run time.

demerphq:
ON irc @nwc10 messaged that he remembered dealing with the same thing, and that he created a patch that includes the change from mine:
https://github.com/nwc10/perl5/commit/d5f88b34531704a752a34ccf475e0ad62775f6c4

Apparently @tonycoz also was involved in a similar issue. Tony you remember anything about this?

lib/feature.pm
```diff
@@ -37,8 +37,8 @@ our %feature_bundle = (
     "5.23"    => [qw(bareword_filehandles current_sub evalbytes fc indirect multidimensional postderef_qq say state switch unicode_eval unicode_strings)],
     "5.27"    => [qw(bareword_filehandles bitwise current_sub evalbytes fc indirect multidimensional postderef_qq say state switch unicode_eval unicode_strings)],
     "5.35"    => [qw(bareword_filehandles bitwise current_sub evalbytes fc isa postderef_qq say signatures state unicode_eval unicode_strings)],
-    "5.37"    => [qw(bitwise current_sub evalbytes fc isa postderef_qq say signatures state unicode_eval unicode_strings)],
-    "all"     => [qw(bareword_filehandles bitwise current_sub declared_refs defer evalbytes extra_paired_delimiters fc indirect isa multidimensional postderef_qq refaliasing say signatures state switch try unicode_eval unicode_strings)],
+    "5.37"    => [qw(bitwise current_sub evalbytes fc isa module_true postderef_qq say signatures state unicode_eval unicode_strings)],
```
Leont, line 40: This assumes the feature will not go through an experimental phase. I'm not sure that's deliberate.

chromatic: Good point. I have mixed feelings about that. There might be edge cases in the implementation we should consider (`require` called in list context, for example). On the other hand, making this an experimental feature is a lot of boilerplate to remove a single line at the end of a module.

tonycoz:
```
tony@venus:.../git/perl5$ cat blahsub.pm
use feature "module_true";;
sub foo { 1 }
""
tony@venus:.../git/perl5$ cat blahsub2.pm
use v5.37;
sub foo { 1 }
""
tony@venus:.../git/perl5$ ./perl -Ilib -I. -le '$x = require "blahsub.pm"; print $x'
blahsub.pm did not return a true value at -e line 1.
tony@venus:.../git/perl5$ ./perl -Ilib -I. -le '$x = require "blahsub2.pm"; print $x'
1
tony@venus:.../git/perl5$ git log -n1
commit 22747d94c925953d807ea3aacd3c223994b83693 (HEAD -> 20365-module_true)
Author: chromatic <chromatic@wgz.org>
Date:   Sun Oct 23 14:44:01 2022 -0700

    [squashme] Fix type of `failed`
    
    Fix tests.
    
    If you can explain why using a different test file here produces the desired
    results, you get many virtual Internet points.
tony@venus:.../git/perl5$ ./perl -v | head -n2

This is perl 5, version 37, subversion 5 (v5.37.5 (v5.37.4-79-g22747d94c9)) built for x86_64-linux

```
I don't see why there's a difference.

tonycoz:
Also:
```
tony@venus:.../git/perl5$ cat ret2.pm
use v5.37;
return "";
tony@venus:.../git/perl5$ ./perl -Ilib -I. -le '$x = require "ret2.pm"; print $x'
ret2.pm did not return a true value at -e line 1.
```

I think there's also a problem with the stack handling, sp is initialized by the `dSP` at the top of pp_leaveeval(), but leave_adjust_stacks() can invalidate that.  You could move the `dSP;` into the POP/PUSH block, but you may as well do `*PL_stack_sp = &PL_sv_undef;`, which is what I think it compiles down to with -O2 if the `dSP;` is in that block.

demerphq:
Check out my branch @tonycoz, it works properly. It needs the cop_features change to make the feature test work at run time.

demerphq:
@tonycoz pretty much all of your feedback is addressed in my version of chromatics patch at https://github.com/Perl/perl5/compare/blead...yves/chr_rfc-18-module-true

tonycoz:
> Check out my branch @tonycoz, it works properly. It needs the cop_features change to make the feature test work at run time.

ahh, sorry, I'll look there.

demerphq:
Just so folks in this ticket have context, @tonycoz commented on my patch here: https://github.com/Perl/perl5/commit/d6b8be71272aafbbaa17bb27be4ac484f8726651

I guess i should just turn mine into a PR so its easier to work with. I just didnt want to steal @chromatic's thunder or credit for this.

demerphq:
Hi. I have taken the liberty of creating https://github.com/Perl/perl5/pull/20446 to replace this PR. That patch works. I have preserved @chromatic's ownership. I will close this PR as it easier to work with that one as a native branch in the master repo.

lib/feature.pm
demerphq: I didnt address this in https://github.com/Perl/perl5/pull/20446 @Leont do you want to raise it again there?

PR #20367, opened 2022-10-03
locale.c: Display thread in DEBUG statements

khwilliamson:
This makes it easier to understand what's going on in threaded perls.


tonycoz (approved):
The only problem I see is the argument to %p isn't `void *`, but we do that everywhere.

PR #20369, opened 2022-10-06
Use `LINE_Tf` consistently for formatting `line_t` values

t-a-k:
Before this change, formatting `line_t` values used to be done with various specifiers, such as `%ld` (with `(long)` cast), `UVuf` (with `(UV)` cast), or plain `%d` (without cast).
I think this inconsistent so that I will suggest to use `LINE_Tf` thoroughly.

This will also silence build warnings (on 32-bit DEBUGGING build) around `line_t`, like these:
```
locale.c:527:28: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘line_t’ {aka ‘const long unsigned int’} [-Wformat=]
deb.c:91:14: warning: comparison of integer expressions of different signedness: ‘long int’ and ‘long unsigned int’ [-Wsign-compare]
```

No functional changes are intended.

### DIFF
--- a/deb.c
+++ b/deb.c
@@ -87,17 +87,18 @@ Perl_vdeb(pTHX_ const char *pat, va_list *args)
 #ifdef DEBUGGING
     const char* const file = PL_curcop ? OutCopFILE(PL_curcop) : "<null>";
     const char* const display_file = file ? file : "<free>";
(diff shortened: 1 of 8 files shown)


khwilliamson:
LGTM, but needs to be rebased

t-a-k:
Thank you for comment.  I've rebased and force-pushed the changes.

PR #20372, opened 2022-10-08
Fix compiler/pattern warnings

khwilliamson:
These are harmless, but a distraction

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4258,7 +4258,7 @@ EGdp	|HV *	|gv_stashsvpvn_cached					\
 #if defined(PERL_IN_HV_C)
 Sx	|void	|clear_placeholders					\
 				|NN HV *hv				\
-				|U32 items
+				|const U32 placeholders
 S	|void	|hsplit 	|NN HV *hv				\
 				|STRLEN const oldsize			\
 				|STRLEN newsize
--- a/embed.fnc
+++ b/embed.fnc
@@ -5488,7 +5488,7 @@ CRip	|bool	|check_regnode_after					\
 				|const STRLEN extra
 CRip	|regnode *|regnext	|NULLOK const regnode *p
 CRip	|regnode *|regnode_after|NULLOK const regnode *p		\
-				|bool varies
+				|const bool varies
 # if defined(DEBUGGING)
 Ep	|void	|debug_peep	|NN const char *str			\
 				|NN const RExC_state_t *pRExC_state	\
--- a/embed.fnc
+++ b/embed.fnc
@@ -3615,8 +3615,8 @@ CMdp	|U8 *	|uvoffuni_to_utf8_flags 				\
 				|UV flags
 Cp	|U8 *	|uvoffuni_to_utf8_flags_msgs				\
 				|NN U8 *d				\
-				|UV input_uv				\
-				|const UV flags 			\
+				|const UV input_uv			\
+				|UV flags				\
 				|NULLOK HV **msgs
 CDbp	|U8 *	|uvuni_to_utf8	|NN U8 *d				\
 				|UV uv
--- a/proto.h
+++ b/proto.h
@@ -6863,7 +6863,7 @@ Perl_gv_stashsvpvn_cached(pTHX_ SV *namesv, const char *name, U32 namelen, I32 f
 #endif
 #if defined(PERL_IN_HV_C)
 STATIC void
-S_clear_placeholders(pTHX_ HV *hv, U32 items);
+S_clear_placeholders(pTHX_ HV *hv, const U32 placeholders);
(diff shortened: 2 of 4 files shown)


jkeenan:
> These are harmless, but a distraction

When I build blead at HEAD (v5.37.4-90-gf607bd2c83), then apply only the change to `t/op/require_errors.t` and run that file, I get no change in output.  Should I?

khwilliamson:
Did you run it on Windows?

jkeenan:
On 10/8/22 10:08, Karl Williamson wrote:
> Did you run it on Windows?
> 

No, on FreeBSD.

khwilliamson:
The commit message indicates this only occurs on  Windows

Leont:
Personally, I don't particularly like declaring `const` for integer arguments: if it doesn't affect the interface, it shouldn't be part of the interface.

And the compiler is silly for warning about this, I don't see how such a warning would ever be helpful.

khwilliamson:
Declaring a parameter const is an extra bit of information to the reader of the code, and the programmer while writing it that this thing isn't supposed to change.  

And it is pretty much silly for the compiler to complain, except it indicates a potential programmer oversight.  But I learned long ago, when lint was the only checker, that it saved me time in the long run to just give in to its invalid complaints, so I wouldn't have to deal with them again.  But if I didn't, my cognitive load would be increased, for no good reason

tonycoz:
> Declaring a parameter const is an extra bit of information to the reader of the code, and the programmer while writing it that this thing isn't supposed to change.

Which compiler warned here?  I don't see neither the newSVbool() nor the regnode_after() warning with VS2022.

khwilliamson:
At least some of them came from https://m-l.org/~perl/smoke/perl/mswin32-x64-multi-thread-win10/mswin32-win10-smoke-me/Karl%20Williamson/log4fec6e83d97275c709d4b4c49331a79a096b5106.log.gz

tonycoz:
Thanks.

I'd be more inclined to remove the const in these cases rather than add it - the const is an implementation detail of the function and shouldn't really be visible in the declaration (I expect it shows in perlapi too.)

Unfortunately this warning can occur for [useful cases](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4028?view=msvc-170) too, so I don't think we can disable it.

khwilliamson:
@tonycoz,  IIRC the end result of what you are saying is that we need to forbid and remove const declarations to non-pointer variables.  Is that your intent?

tonycoz:
> @tonycoz, IIRC the end result of what you are saying is that we need to forbid and remove const declarations to non-pointer variables. Is that your intent?

Only for the parameters in the function declaration/definition ("formal parameters").

While @Leont mentioned integer parameters, I don't think it's limited to integers, but for any type parameter where the const qualifier is on the parameter itself, eg.

```
void f(int x); /* fine */
void g(const int x); /* causes MSVC problems */
void h(const char *x); /* fine, x points to const, but isn't const itself */
void m(char const *x); /* fine, has exactly the same type as h() */
void j(const char * const x); /* x itself is const, but this only matters inside j() itself, causes MSVC problems */
void k(char const * const x); /* has exactly the same type as j() */
```

demerphq (approved):

demerphq:
@khwilliamson this is in conflict and stalled. Would you like me to rebase it onto blead so we can decide if we want to apply it? @tonycoz do you have an opinion on what to do with this PR?

demerphq:
I resolved conflicts on this only, I have no opinion on whether it should be applied.

jkeenan:
> I resolved conflicts on this only, I have no opinion on whether it should be applied.

Thanks for the conflict resolution.  There doesn't really seem to be enough consensus on this patch to move forward.  Should we close the ticket?

jkeenan:
> > I resolved conflicts on this only, I have no opinion on whether it should be applied.
> 
> Thanks for the conflict resolution. There doesn't really seem to be enough consensus on this patch to move forward. Should we close the ticket?

There has been no further discussion in this ticket since the end of July.  It still seems that we do not have strong support for the changes proposed in this p.r.  I propose that we close the p.r. in 7 days unless someone makes a good case for it.

jkeenan:
> > > I resolved conflicts on this only, I have no opinion on whether it should be applied.
> > 
> > 
> > Thanks for the conflict resolution. There doesn't really seem to be enough consensus on this patch to move forward. Should we close the ticket?
> 
> There has been no further discussion in this ticket since the end of July. It still seems that we do not have strong support for the changes proposed in this p.r. I propose that we close the p.r. in 7 days unless someone makes a good case for it.

Closing as per plan.  Thanks for all who contributed to the discussion.

PR #20380, opened 2022-10-09
handy.h: Set macro to false if can't ever be true

khwilliamson:
It's unlikely that perl will be compiled with out the LC_CTYPE locale category being enabled.  But if it isn't, there is no sense in having per-interpreter variables for various conditions in it, and no sense having code that tests those variables.

This commit changes a macro to always yield 'false' when this is disabled, adds a new similar macro, and changes some occurrences that test for a variable to use the macros instead of the variables.  That way the compiler knows these to conditions can never be true.

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -1880,7 +1880,13 @@ END_EXTERN_C
 #define toUPPER_LATIN1_MOD(c) ((! FITS_IN_8_BITS(c))                       \
                                ? (c)                                       \
                                : PL_mod_latin1_uc[ (U8) (c) ])
-#define IN_UTF8_CTYPE_LOCALE PL_in_utf8_CTYPE_locale
+#ifdef USE_LOCALE_CTYPE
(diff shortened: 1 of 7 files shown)


jkeenan:
> It's unlikely that perl will be compiled with out the LC_CTYPE locale category being enabled. But if it isn't, there is no sense in having per-interpreter variables for various conditions in it, and no sense having code that tests those variables.
> 

Are there any Configure options we could use to test the "isn't" case?  Or platforms that would typically show this?

khwilliamson:
On 10/9/22 08:48, James E Keenan wrote:
>     It's unlikely that perl will be compiled with out the LC_CTYPE
>     locale category being enabled. But if it isn't, there is no sense in
>     having per-interpreter variables for various conditions in it, and
>     no sense having code that tests those variables.

To test this, I add to Configure -DNO_LOCALE_CTYPE

tonycoz (approved):

PR #20385, opened 2022-10-12
fix clockid_t compilation failure on macos system perl (and some other things)

### LINKED ISSUE #20362: Time::HiRes build error, 'clockid_t' redefined, Mac OS Darwin 19.6.0
While attempting `install Time::HiRes` from CPAN command line, I got the following error:

    `HiRes.xs:95:13: error: typedef redefinition with different types ('int' vs 'enum clockid_t')`

I am running Mac OS Darwin Kernel Version 19.6.0

Problem occurs during: 
    `CPAN.pm: Building A/AT/ATOOMIC/Time-HiRes-1.9764.tar.gz`

Compilation args and error messages are:
```
/usr/bin/perl /System/Library/Perl/5.18/ExtUtils/xsubpp  -typemap /System/Library/Perl/5.18/ExtUtils/typemap -typemap typemap  HiRes.xs > HiRes.xsc && mv HiRes.xsc HiRes.c
cc -c   -g -pipe -fno-common -DPERL_DARWIN -fno-strict-aliasing -fstack-protector -Os   -DVERSION=\"1.9764\" -DXS_VERSION=\"1.9764\"  -iwithsysroot "/System/Library/Perl/5.18/darwin-thread-multi-2level/CORE"  -DTIME_HIRES_NANOSLEEP -DTIME_HIRES_CLOCK_GETTIME -DTIME_HIRES_CLOCK_GETTIME_EMULATION -DTIME_HIRES_CLOCK_GETRES -DTIME_HIRES_CLOCK_GETRES_EMULATION -DTIME_HIRES_CLOCK_NANOSLEEP -DTIME_HIRES_CLOCK_NANOSLEEP_EMULATION -DUSE_PPPORT_H HiRes.c
**HiRes.xs:95:13: error: typedef redefinition with different types ('int' vs 'enum clockid_t')**
typedef int clockid_t;
            ^
**/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/incl**ude/time.h:172:3: note: previous definition is here
} clockid_t;
  ^
1 error generated.
```

tonycoz:
Fixes #20362 

This is built on top of @bram-perl's PR #20309 and extends it in a few ways, fixing some of the limitations of the dist-modules work that I ran out of energy for the first time around:

- continuing on error now depends on a command-line flag
- added a help option to test-dist-modules.pl
- added an option to install to a temp directory instead of to perl, so it can be used for system perls
- test dist/ modules against the macos system perl (to reproduce #20362)
- allow the dist-modules tests to run when test_porting fails, to match other tests
- threads and threads-shared (which weren't properly tested with the non-threaded perls used by dist-modules) are now passed the flag needed to make them use ppport.h
- added headers bundled by the threads and threads-shared CPAN distributions (in test-dist-modules.pl, as with test.pl)
- log system calls for builds, and better report failures
- supply the module name, not the dist name as NAME in the generated Makefile.PL
- optionally accept list of distributions to test on the test-dist-modules.pl command-line

And fixes that came about from being able to test on macos, and to test the threads code:
- fix building threads (a recent change broke it)
- fix the clockid_t probe problem, which was caused by the compiler not finding the system perl's CORE headers, and fix another problem when building against the darwin system perl
- make ExtUtils::CBuilder also properly tell the compiler where to find the system perl's CORE headers, which allows ExtUtils::ParseXS to test successfully.

### DIFF
--- a/Porting/test-dist-modules.pl
+++ b/Porting/test-dist-modules.pl
@@ -38,6 +38,13 @@ if ($separate) {
       "$install_base/lib/perl5";
 }
 
+my %dist_config = (
+    # these are defined by the modules as distributed on CPAN
+    # I don't know why their Makefile.PLs aren't in core
+    "threads"        => [ "DEFINE=-DHAS_PPPORT_H" ],
+    "threads-shared" => [ "DEFINE=-DHAS_PPPORT_H" ],
+   );
+
 my $start = getcwd()
   or die "Cannot fetch current directory: $!\n";
 
@@ -178,7 +185,11 @@ EOM
 
     my $verbose = $github_ci && $ENV{'RUNNER_DEBUG'} ? 1 : 0;
     my $failed = "";
-    if (system($^X, "Makefile.PL", @config)) {
+    my @my_config = @config;
+    if (my $cfg = $dist_config{$name}) {
+        push @my_config, @$cfg;
+    }
+    if (system($^X, "Makefile.PL", @my_config)) {
         $failed = "Makefile.PL";
         die "$name: Makefile.PL failed\n" unless $continue;
     }
--- a/Porting/test-dist-modules.pl
+++ b/Porting/test-dist-modules.pl
@@ -1288,6 +1288,11 @@ sub watchdog ($)
 #  define CLANG_DIAG_RESTORE
 #endif
 
+/* Added in 5.38 */
+#ifndef PERL_SRAND_OVERRIDE_NEXT_PARENT
+#  define PERL_SRAND_OVERRIDE_NEXT_PARENT()
+#endif
+
 #endif
 -- shared.h --
 #ifndef _SHARED_H_
--- a/Porting/test-dist-modules.pl
+++ b/Porting/test-dist-modules.pl
@@ -118,6 +118,9 @@ sub test_dist {
     if ($name eq "threads" || $name eq "threads-shared") {
         write_threads_h();
     }
+    if ($name eq "threads-shared") {
+        write_shared_h();
+    }
     unless (-f "Makefile.PL") {
         print "  Creating Makefile.PL for $name\n";
         my $key = "ABSTRACT_FROM";
@@ -230,6 +233,11 @@ sub write_testpl {
 # threads and threads-shared bundle this file, which isn't needed in core
 sub write_threads_h {
     _write_from_data("threads.h");
+}
+
+# threads-shared bundles this file, which isn't needed in core
+sub write_shared_h {
+    _write_from_data("shared.h");
 }
 
 # file data read from <DATA>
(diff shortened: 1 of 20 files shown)


tonycoz:
> * allow the dist-modules tests to run when test_porting fails, to match other tests

Removed this, as #20304 did it better and is has been applied to blead.

Rebased to resolve conflicts.

Porting/test-dist-modules.pl
```diff
@@ -115,6 +115,9 @@ sub test_dist {
     if ($name eq "IO" || $name eq "threads" || $name eq "threads-shared") {
         write_testpl();
     }
+    if ($name eq "threads" || $name eq "threads-shared") {
```
bram-perl, line 126: *commit  CI/dist-modules: add threads.h for threads/threads-shared:*

There appears to be some typos in the commit message:
```
...
Since we now how two files, rather than including them as (fairly
large for t/test.pl) here-docs, move the content to __DATA__.
...
```

Since we "now how two files"?

Porting/test-dist-modules.pl
```diff
@@ -292,14 +292,26 @@ sub run {
...
  -c | -continue
      Continue processing after failures
      Devel::PPPort must successfully build to continue.
  -s | -separate
      Install to a work path, not to perl's site_perl.
  -h | -help
      Display this message.
+
+Optional distnames should be names of the distributions under dist/ to
+test.  If omitted all of the distributions under dist/ are tests.
```
bram-perl: `... are tests.` should probably be `... are tested.`?

Porting/test-dist-modules.pl
```diff
@@ -1288,6 +1288,11 @@ ($)
 #  define CLANG_DIAG_RESTORE
 #endif
 
+/* Added in 5.38 */
+#ifndef PERL_SRAND_OVERRIDE_NEXT_PARENT
+#  define PERL_SRAND_OVERRIDE_NEXT_PARENT()
```
bram-perl, line 1293: *commit  CI/dist-modules: define PERL_SRAND_OVERRIDE_NEXT_PARENT for CPAN use*

I sort of wonder if this is the right fix..
I would expect it to also be a problem when a new version of 'threads' is ever released on CPAN.. won't it fail for the older perls?
That is: I would expect a fix in dist/threads/threads.xs
(Or should Devel::PPPort take care of this?, i.e. defining it to a no-op on older perls?)

tonycoz: Well, hopefully whoever releases threads tests on any older perl before release.

Ideally threads.h (and the threads, threads-shared Makefile.PLs) would be core - dist/ modules are at least theoretically perl upstream.

We could add it to Devel::PPPort, but I don't see much use for this outside of threads.

We could add it to threads.xs, but dist/threads/ tends to put those types of portability definitions in threads.h.

Porting/test-dist-modules.pl
tonycoz: Fixed.

Porting/test-dist-modules.pl
tonycoz: Thanks, fixed.

Porting/test-dist-modules.pl
bram-perl: I see..

(In other words: it's all messy because the files needed to make a CPAN-release are not included in the perl5 repo..)

bram-perl (approved):

tonycoz:
Applied manually to blead

PR #20386, opened 2022-10-12
locale.c: Silence unused var warning on dragonfly

khwilliamson:
This just moves some code out of #ifdefs so that the compiler sees it, decides it is always false, and almost certainly won't generate any code for it, but stops warning.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -3132,13 +3132,9 @@ S_my_localeconv(pTHX_ const int item, const locale_utf8ness_t locale_is_utf8)
     PERL_UNUSED_ARG(item);
     PERL_UNUSED_ARG(locale_is_utf8);
 
-#    ifdef USE_LOCALE_NUMERIC
-
     /* When there is a nl_langinfo, we will only be called for localeconv
      * numeric purposes. */
     const bool is_localeconv_call = true;
-
-#    endif
 
 #  else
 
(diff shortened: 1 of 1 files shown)


jkeenan:
> This just moves some code out of #ifdefs so that the compiler sees it, decides it is always false, and almost certainly won't generate any code for it, but stops warning.

@khwilliamson, thanks for looking into this.  However, what is the warning that this pull request clears up?

I just built my customary threaded perl on FreeBSD-12 at HEAD of blead and got (as expected) only this warning:
```
$ parse-build-warnings f08b7a33d9.freebsd.threaded.maketp.output.txt.gz
File:  f08b7a33d9.freebsd.threaded.maketp.output.txt.gz

[
  {
    char   => 1,
    group  => "Wunused-function",
    line   => 5946,
    source => "locale.c",
    text   => "unused function 'S_get_displayable_string'",
  },
]
```
I then built your branch with the same config_args and got this warning:
```
$ parse-build-warnings f08b7a33d9.freebsd.threaded.maketp.output.txt.gz
File:  f08b7a33d9.freebsd.threaded.maketp.output.txt.gz

[
  {
    char   => 1,
    group  => "Wunused-function",
    line   => 5946,
    source => "locale.c",
    text   => "unused function 'S_get_displayable_string'",
  },
]
```
So the warning was not cleared up (though its triggering line had moved in `locale.c`).  Can you clarify?

khwilliamson:
I was wrong about the distro I saw this on.  I is dragonfly, and this fixed it.

PR #20387, opened 2022-10-12
locale.c: Fix Debug statement on netbsd

khwilliamson:
Other platforms declare the nl_item typedef an int, but this one makes it a long.  To portably output its value, cast it to a long and use the %ld format.


jkeenan:
@khwilliamson would it be possible to make this into a smoke-me branch?  We could then get feedback from Carlos's reports on NetBSD?

tonycoz (approved):

khwilliamson:
http://perl.develop-help.com/?b=smoke-me%2Fkhw-netbsd_warning

khwilliamson:
The smoke reports verified this worked

PR #20388, opened 2022-10-12
t/test.pl: Add ability to cancel an 'alarm' watchdog

khwilliamson:
A watchdog affects the entire rest of the file.  As more tests get added
to a test file, they could end up triggering the watchdog.  It's trivial
to cancel a watchdog that uses SIGALRM.  This commit adds the capability
to do so; checking that what gets cancelled actually was set via this
method.

The other watchdog methods remain uncancellable.  Patches welcome.
These were created so that one could set a watchdog for cases where we
are testing timers and alarms.  Those situations aren't as common.
Hence this commit allows cancellation in most cases, but not all.

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -1731,10 +1731,12 @@ sub warning_like {
 #
 # NOTE:  If the test file uses 'threads', then call the watchdog() function
 #        _AFTER_ the 'threads' module is loaded.
+{ # Closure
+    my $watchdog;
+    my $watchdog_thread;
+
 sub watchdog ($;$)
 {
-    CORE::state $watchdog;
-    CORE::state $watchdog_thread;
     my $timeout = shift;
 
     # If cancelling, use the state variables to know which method was used to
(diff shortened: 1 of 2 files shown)


tonycoz:
The only problem I see here is systems without alarm() (Win32) won't have a watchdog in this particular case.

t/test.pl
```diff
@@ -1733,16 +1733,42 @@ sub warning_like {
 #        _AFTER_ the 'threads' module is loaded.
 sub watchdog ($;$)
 {
+    CORE::state $watchdog;
+    CORE::state $watchdog_thread;
```
tonycoz: From the top of test.pl:
```
# Do not rely on features found only in more modern Perls here, as some CPAN
# distributions copy this file and must operate on older Perls. Similarly, keep
# things, simple as this may be run under fairly broken circumstances.
```
I think that means we shouldn't be using state here.

khwilliamson: Changed to a closure

tonycoz (approved):

PR #20393, opened 2022-10-13
Ensure 'ok #' test output includes a location

leonerd:
If no test name is supplied it can be hard to work out where the failure was, if the output is something like

    ...
    ok 100
    ok 101
    ok 102
    Segmentation fault

There's no easy correlation with the lines in the .t file, short of counting all the calls.

By adding the caller file/line number, at least we can get a rough idea of where the test script got to when it failed

    ...
    ok 102 - [at foo.t line 1234]
    Segmentation fault

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -246,7 +246,7 @@ sub _ok {
         $name =~ s/#/\\#/g;
 	$out = $pass ? "ok $test - $name" : "not ok $test - $name";
     } else {
-	$out = $pass ? "ok $test" : "not ok $test";
+	$out = $pass ? "ok $test - [$where]" : "not ok $test - [$where]";
     }
 
     if ($TODO) {


Leont:
I'm not sure this is a good idea in the general case. Most of the time it really is noise, it's only in this particular case of a segfault that it's actually desirable. And in many cases it will be wrong/unhelpful (because it should really refer to something upstack to clarify anything).

What we should do instead is actually add descriptions.

leonerd:
> What we should do instead is actually add descriptions.

Oh I definitely agree there - the tests should have proper descriptions so you can find stuff and know what's broken. This would just be a quick initial first step so as to at least get me moving while I add those.

khwilliamson:
On 10/13/2022 3:32 PM, Paul Evans wrote:
>
>     What we should do instead is actually add descriptions.
>
> Oh I definitely agree there - the tests should have proper 
> descriptions so you can find stuff and know what's broken. This would 
> just be a quick initial first step so as to at least get me moving 
> while I add those.
>
Why not use this iff there isn't a description

jkeenan:
> > What we should do instead is actually add descriptions.
> 

I agree with @LeonT here.  If you want to see the impact of this, do what I did: compare blead and Paul's branch.
```
$ cd t; TEST_JOBS=1 ./perl harness -v op/*.t 2>&1 |tee /tmp/blead.out; cd -

$ cd t; TEST_JOBS=1 ./perl harness -v op/*.t 2>&1 |tee /tmp/leonerd.ensure.out; cd -

$ cd /tmp; vimdiff blead.out leonerd.ensure.out
```
Do you really consider that an improvement?  Wouldn't it be better to have some switch that you could turn on on a per-file basis as needed?

> Oh I definitely agree there - the tests should have proper descriptions so you can find stuff and know what's broken. This would just be a quick initial first step so as to at least get me moving while I add those.

Yes, the tests should have proper descriptions.  I got my start with working on the core distribution 10 years ago doing just that -- and for that @rjbs and @dagolden thrust a commit bit on me.  But the lesson I draw from that is that it's not a question of **whether** we should write proper descriptions but **who** should write those descriptions.

**We should not have our senior developers writing test descriptions.**  That's a misallocation of our scarce human resources.  **We should be recruiting new contributors to writing these descriptions,** thereby introducing them to working on the core distribution -- getting them listed in `AUTHORS` and possibly earning a commit bit.  Our senior developers should be **mentoring** newcomers rather than themselves taking on projects which mentored newcomers could take on.

khwilliamson:
I have run into this issue myself, and it took quite a bit of effort to find the source of the failing tests.  Therefore I'm eager to have some automatic way to narrow down which test is failing.

How about an environment variable that if set would induce this behavior and only for tests without a name.  PERL_DEBUG_FULL_TEST is in use currently and could be made to also enable this; or another variable could be introduced

leonerd:
Wow. I didn't imagine there'd be anywhere near this amount of objection. There's almost zero CPU-time cost, and it generates output that honestly in 99% of cases no user ever sees. You only ever see the output when you're looking at one specific file.

Still if people don't want it I'm happy to abandon it. I just felt it useful since I *needed* to do it to find my bug. If nobody else thinks this is useful I'll leave it as a one-line comment somewhere to tell people how to do it for themselves instead.

leonerd:
@jkeenan 
> **We should not have our senior developers writing test descriptions.**

On the one hand: yes.

On the other hand: I'm stuck unable to make progress on `use feature 'class'` because I can't add the `:isa` syntax to the parser because `toke.c` is took concerned with subroutine attributes and my attempt to move the code out causes it to spinlock *somewhere* in `t/op/attrs.t` but exactly where I can't tell because all I know is it fails after `ok 38`. Now if you want to arrange for someone to put test descriptions in that file, please go ahead. I need it to make further progress. My quick hack to get the caller file + line number at least allowed me to find that test is

```
ok 38 - [at t/op/attrs.t line 123]
```

Already it's immediately helpful output.

I fully agree this is no decent substitute for having proper test descriptions. However it *does* get me out of an immediate hole I'm otherwise stuck in.

khwilliamson:
I have run into this issue myself, and it took quite a bit of effort to find the source of the failing tests.  Therefore I'm eager to have some automatic way to narrow down which test is failing.

How about an environment variable that if set would induce this behavior and only for tests without a name.  PERL_DEBUG_FULL_TEST is in use currently and could be made to also enable this; or another variable could be introduced

Tux:
I would love to see this extra info, most likely under `prove -v` for all lines that do not have a comment.
As @khwilliamson and @leonerd I have run into this multiple times

demerphq:
@khwilliamson note the that the patch from @leonerd does only change the case where there is no name. 

I am more or less in favour of this patch. To address the issue that @Leont raises about the information being wrong maybe add a way to show the full call stack as a diagnostic for passing tests.

tonycoz:
I think this is a good change, and I don't think it should depend on a runtime flag like an environment variable.

It doesn't interfere with manually written test descriptions since it only applies where those descriptions haven't been provided.

It provides significant utility both when the test process crashes in finding the location where the developer needs to look when trying to track down the cause of a failure.

It does not prevent adding descriptions to tests that do not have descriptions.

It's only real flaw is the cases where longer call stacks might be useful, which could be added to this, or changed in a separate commit, but this change is a significant improvement over the current situation, at least until every test has a description (and we can make that fail instead of substituting in a description).

demerphq:
That makes it 5 to 2 for merging... I say go for it.

demerphq:
I think the consensus is that this is good, so i will merge it.

PR #20397, opened 2022-10-14
Better handling of builtin CV attributes

leonerd:
The previous code would handle subroutine attributes directly against `PL_compcv` as a side-effect of merely parsing the syntax in `yyl_colon()`, an unlikely place for anyone to find it. This complicates the way the parser works.

The new structure creates a new function to apply all the builtin attributes out of an attribute list to any given CV, and invokes it from the parser at a slightly better time.


leonerd:
The only failing test was `dist/Time-HiRes/t/alarm.t` on one of the MacOSX boxes; which I'm going to put down to being flaky and unrelated here.

perly.y
```diff
@@ -709,7 +709,11 @@ proto
 subattrlist
 	:	empty
 	|	COLONATTR THING
-			{ $$ = $THING; }
+			{
+			  if($THING && !PL_parser->sig_seen)
+			      $THING = apply_builtin_cv_attributes(PL_compcv, $THING);
```
happy-barney: better assign directly to `$$` than to named reference

leonerd: Oh, fair. Actually probably better yet is to make a local variable out of it;

```C
OP *attrlist = $THING;
...
```

leonerd: Adjusted and force-pushed. PTAL

toke.c
```diff
@@ -5989,10 +5963,8 @@ yyl_colon(pTHX_ char *s)
                    rejected) by a package's attribute routines, but is
                    justified by the performance win for the common case
                    of applying only built-in attributes.) */
-                else
-                    attrs = op_append_elem(OP_LIST, attrs,
```
bram-perl, line 5993: The comment above this block no longer make sense and should be updated.

leonerd: Now removed by d88cfc6975.

Needs squash before merge.

tonycoz (approved):

Issue #20396, opened 2022-10-14
try_yyparse breaks Attribute::Handlers


demerphq:
**Description**
In response to  https://github.com/Perl/perl5/issues/20161 we applied https://github.com/Perl/perl5/commit/c304acb49dada68ec331d50f8af45f0dda83ba6a which seems to cause trouble with certain types of Attribute::Handlers. Specifically when handler contains attributes which die during eval to perl `:Sig(s, foo)` we end up malbehaving in different ways depending on the exact code involved. 

Disabling c304acb49dada68ec331d50f8af45f0dda83ba6a (or reverting it) fixes the problem, but it is not clear to me if it is the cause of the problem, or whether it just exposes another problem. I think possibly the latter, but i am not sure.

I will post more information about what i know about this ticket as replies.

**Steps to Reproduce**

Code that fails with c304acb, note the `s, ARRAY`, but works in older perls (perhaps accidentally)
```
$ cat ~/scratch/t_eixo_bad.pl 
use Eixo::Base::Util;
sub concat : Sig(s, ARRAY) { }
warn "Done\n";
```

Code that works with c304acb and in older perls, note the `i, ARRAY`.
```
$ cat ~/scratch/t_eixo_ok.pl 
use Eixo::Base::Util;
sub concat : Sig(i, ARRAY) { }
warn "Done\n";
```

Reduced module with just a little of debug output to see what is going on:
```
$ cat ~/scratch/lib/Eixo/Base/Util.pm 
package Eixo::Base::Util;
use strict;
use warnings;

use Attribute::Handlers;
use Carp qw(cluck);
use Data::Dumper;

sub UNIVERSAL::Sig :ATTR(CODE) { 
    #cluck("In 'Sig': ", Dumper(\@_));
}
1;
```

with this I can reproduce the issues with:
```
./perl -Ilib -I/home/yorton/scratch/lib ~/scratch/t_eixo_bad.pl
```
from inside of a perl build directory.

**Expected behavior**
We shouldn't have problems from these examples. 

**Perl configuration**
```
Summary of my perl5 (revision 5 version 37 subversion 5) configuration:
  Derived from: ac9ba9d6a3979c516130176f25df5d0ac898a553
  Ancestor: f08b7a33d99b9b204260e108bc1789c5392aff36
  Platform:
    osname=linux
    osvers=5.14.0-1052-oem
    archname=x86_64-linux-thread-multi
[...]

PR #20398, opened 2022-10-14
pp_ctl.c - make sure we dont leak PL_restartop from failed compile

demerphq:
Fix GH Issue #20396, try_yyparse() breaks Attribute::Handlers.

Reduced test case is:

```
    perl -e'CHECK { eval "]" }'
```

which should not assert or segfault.

In c304acb49 we made it so that when doeval_compile() is executed and it calls yyparse() inside of an eval any exceptions that occur during the parse process are trapped by try_yyparse() so that exection would return to doeval_compile(). This was done so that post eval compilation cleanup logic could be handled similarly regardless of whether Perl_croak() was called or not. However the logic to setup PL_restartop was not adjusted accordingly.

The opcode that calls doeval_compile() setups an eval context data before it calls doeval_compile().  This data includes the "retop" which is used to return control to after the eval should it die and is set to the be the evaling opcodes op_next. When Perl_die_unwind() is called it sets PL_restartop to be the "retop" of the of the current eval frame, and then does a longjmp, on the assumption it will end up inside of a "run loop enabled jump enviornment", where it restarts the run loop based on the value of PL_restartop, zeroing it aftewards.

After c304acb49 however, a die inside of try_yyparse the die_unwind returns control back to the try_yyparse, which ignores PL_restartop, and leaves it set. Code then goes through the "compilation failed" branch and execution returns to PL_restartop /anyway/, as PL_op hasn't changed and pp_entereval returns control to PL_op->op_next, which is what we pushed into the eval context anyway for the PL_restartop.

The end result of this however is that PL_restartop remains set when we enter perl_run() for the first time. perl_run() is a "run loop enabled jump enviornment" which uses run_body() to do its business, such that when PL_restartop is NULL it executes the just compiled body of the program, and when PL_restartop is not null it assumes it must be in the eval handler from an eval from the main body and it should recontinue. The leaked PL_restartop is thus executed instead of the main program body and things go horribly wrong.

This patch changes it so that when try_yyparse traps an exception we restore PL_restartop back to its old value. Same for its partner PL_restartjmpenv. This is fine as they have been set to the values from the beginning of the eval frame which we are part of, which is now over.

### DIFF
--- a/dist/Attribute-Handlers/lib/Attribute/Handlers.pm
+++ b/dist/Attribute-Handlers/lib/Attribute/Handlers.pm
@@ -4,9 +4,9 @@ use Carp;
 use warnings;
 use strict;
 our $AUTOLOAD;
-our $VERSION = '1.02'; # remember to update version in POD!
+our $VERSION = '1.03'; # remember to update version in POD!
 # $DB::single=1;
-
+my $debug= $ENV{DEBUG_ATTRIBUTE_HANDLERS} || 0;
 my %symcache;
 sub findsym {
 	my ($pkg, $ref, $type) = @_;
@@ -241,7 +241,8 @@ sub _apply_handler_AH_ {
 	my ($declaration, $phase) = @_;
 	my ($pkg, $ref, $attr, $data, $raw, $handlerphase, $filename, $linenum) = @$declaration;
 	return unless $handlerphase->{$phase};
-	# print STDERR "Handling $attr on $ref in $phase with [$data]\n";
+        print STDERR "Handling $attr on $ref in $phase with [$data]\n"
+            if $debug;
 	my $type = ref $ref;
 	my $handler = "_ATTR_${type}_${attr}";
 	my $sym = findsym($pkg, $ref);
@@ -249,12 +250,29 @@ sub _apply_handler_AH_ {
 	no warnings;
 	if (!$raw && defined($data)) {
 	    if ($data ne '') {
-		my $evaled = eval("package $pkg; no warnings; no strict;
-				   local \$SIG{__WARN__}=sub{die}; [$data]");
-		$data = $evaled unless $@;
+                # keeping the minimum amount of code inside the eval string
+                # makes debugging perl internals issues with this logic easier.
+                my $code= "package $pkg; my \$ref= [$data]; \$data= \$ref; 1";
+                print STDERR "Evaling: '$code'\n"
+                    if $debug;
+                local $SIG{__WARN__} = sub{ die };
+                no strict;
+                no warnings;
+                # Note in production we do not need to use the return value from
+                # the eval or even consult $@ after the eval - if the evaled code
+                # compiles and runs successfully then it will update $data with
+                # the compiled form, if it fails then $data stays unchanged. The
+                # return value and $@ are only used for debugging purposes.
+                # IOW we could just replace the following with eval($code);
+                eval($code) or do {
+                    print STDERR "Eval failed: $@"
+                        if $debug;
+                };
 	    }
 	    else { $data = undef }
 	}
+
+        # now call the handler with the $data decoded (maybe)
 	$pkg->$handler($sym,
 		       (ref $sym eq 'GLOB' ? *{$sym}{ref $ref}||$ref : $ref),
 		       $attr,
@@ -300,7 +318,7 @@ Attribute::Handlers - Simpler definition of attribute handlers
 
 =head1 VERSION
 
-This document describes version 1.02 of Attribute::Handlers.
+This document describes version 1.03 of Attribute::Handlers.
 
 =head1 SYNOPSIS
 
--- a/perl.c
+++ b/perl.c
@@ -2701,6 +2701,11 @@ perl_run(pTHXx)
 #ifndef MULTIPLICITY
     PERL_UNUSED_ARG(my_perl);
 #endif
+    /* perl_parse() may end up starting its own run loops, which might end
+     * up "leaking" PL_restartop from the parse phase into the run phase
+     * which then ends up confusing run_body(). This leakage shouldn't
+     * happen and if it does its a bug. */
+    assert(!PL_restartop);
 
     oldscope = PL_scopestack_ix;
 #ifdef VMS
--- a/perl.c
+++ b/perl.c
@@ -2783,6 +2783,14 @@ S_run_body(pTHX_ I32 oldscope)
     PERL_SET_PHASE(PERL_PHASE_RUN);
 
     if (PL_restartop) {
+#ifdef DEBUGGING
+        /* this complements the "EXECUTING..." debug we emit above.
+         * it will show up when an eval fails in the main program level
+         * and the code continues after the error.
+         */
+        if (!DEBUG_q_TEST)
+          PERL_DEBUG(PerlIO_printf(Perl_debug_log, "\nCONTINUING...\n\n"));
+#endif
         PL_restartjmpenv = NULL;
         PL_op = PL_restartop;
         PL_restartop = 0;
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -3527,16 +3527,39 @@ Perl_find_runcv_where(pTHX_ U8 cond, IV arg, U32 *db_seqp)
 }
 
 
-/* Run yyparse() in a setjmp wrapper. Returns:
+/* S_try_yyparse():
+ *
+ * Run yyparse() in a setjmp wrapper. Returns:
  *   0: yyparse() successful
  *   1: yyparse() failed
  *   3: yyparse() died
+ *
+ * This is used to trap Perl_croak() calls that are executed
+ * during the compilation process and before the code has been
+ * completely compiled. It is expected to be called from
+ * doeval_compile() only. The parameter 'caller_op' is
+ * only used in DEBUGGING to validate the logic is working
+ * correctly.
+ *
+ * See also try_run_unitcheck().
+ *
  */
 STATIC int
-S_try_yyparse(pTHX_ int gramtype)
+S_try_yyparse(pTHX_ int gramtype, OP *caller_op)
 {
-    int ret;
+    /* if we die during compilation PL_restartop and PL_restartjmpenv
+     * will be set by Perl_die_unwind(). We need to restore their values
+     * if that happens as they are intended for the case where the code
+     * compiles and dies during execution, not where it dies during
+     * compilation. PL_restartop and caller_op->op_next should be the
+     * same anyway, and when compilation fails then caller_op->op_next is
+     * used as the next op after the compile.
+     */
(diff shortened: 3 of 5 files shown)


dist/Attribute-Handlers/lib/Attribute/Handlers.pm
```diff
@@ -249,12 +249,16 @@ sub _apply_handler_AH_ {
...
 	if (!$raw && defined($data)) {
 	    if ($data ne '') {
-		my $evaled = eval("package $pkg; no warnings; no strict;
-				   local \$SIG{__WARN__}=sub{die}; [$data]");
-		$data = $evaled unless $@;
+                my $code= "package $pkg; my \$ref= [$data]; \$data= \$ref; 1";
+                local $SIG{__WARN__} = sub{ die };
+                no strict;
+                no warnings;
+                eval($code);
```
bram-perl: I would probably move the Attribute-Handlers change to it's own commit. (These make the code more readable but shouldn't be required for the actual bug fix)

(and/or maybe add something like: `# print STDERR "Evaling code: $code\n";` since it could be useful when debugging. (only suggesting it because there already are commented `print STDERR` statements in the code, a debug of verbose flag might be more useful but might be overkill))

perl.c
```diff
@@ -2770,6 +2775,10 @@ S_run_body(pTHX_ I32 oldscope)
     PERL_SET_PHASE(PERL_PHASE_RUN);
 
     if (PL_restartop) {
+#ifdef DEBUGGING
+        if (!DEBUG_q_TEST)
+          PERL_DEBUG(PerlIO_printf(Perl_debug_log, "\nRESTARTOP...\n\n"));
+#endif
```
bram-perl: Did you intend to include this?
(The debug message itself doesn't tell very much; if it was intended: can the message be improved?)

t/op/eval.t
```diff
@@ -723,10 +723,10 @@ pass("eval in freed package does not crash");
         'eval "UNITCHECK { eval q(UNITCHECK { die; }); print q(A-) }";',
         'eval "UNITCHECK { eval q(BEGIN     { die; }); print q(A-) }";',
         'eval "BEGIN     { eval q(UNITCHECK { die; }); print q(A-) }";',
+        'CHECK { eval "]" } print q"A-";'
     ) {
         fresh_perl_is($line . ' print "ok";', "A-ok", {}, "No segfault: $line");
         my $sort_line= 'my @x= sort { ' . $line . ' } 1,2;';
-        fresh_perl_is($line . ' print "ok";', "A-ok", {}, "No segfault: $line");
-
+        fresh_perl_is($line . ' print "ok";', "A-ok", {}, "No segfault inside sort: $line");
```
bram-perl: Hmm.. this doesn't make sense (but neither did the original code)
I think it should be: `fresh_perl_is($sort_line . ' print.... )`
Right now it appears to be calling `fresh_perl_is($line...`) twice

t/op/eval.t
```diff
@@ -723,10 +723,10 @@ pass("eval in freed package does not crash");
         'eval "UNITCHECK { eval q(UNITCHECK { die; }); print q(A-) }";',
         'eval "UNITCHECK { eval q(BEGIN     { die; }); print q(A-) }";',
         'eval "BEGIN     { eval q(UNITCHECK { die; }); print q(A-) }";',
+        'CHECK { eval "]" } print q"A-";'
```
bram-perl: Might be good to also test: `INIT { eval "]" } ...` since that  was observed (by @haarg) to segfault as well.

pp_ctl.c
```diff
@@ -3724,6 +3724,12 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
      * always handle that case. */
     assert(!CATCH_GET);
     /* compile the code */
+
+    /* if we die during compilation PL_restartop and PL_restartjmpenv
+     * will be set. We need to restore their values if that happens. */
```
bram-perl: Consider changing it to something like:
```
    /* if we die during compilation PL_restartop and PL_restartjmpenv
     * will be set (by Perl_die_unwind). We need to restore their values if that happens. */
```

pp_ctl.c
```diff
@@ -3733,10 +3739,19 @@ S_doeval_compile(pTHX_ U8 gimme, CV* outside, U32 seq, HV *hh)
...
-         * compilation, so the EVAL CX block has already been popped, and
-         * various vars restored */
-        if (yystatus != 3) {
+        if (yystatus == 3) {
+            /* try_yyparse() died and we trapped the error. We need to restore
+             * the old PL_restartjmpenv and PL_restartop values. */
+            assert(!in_require); /* not in a require */
+            assert(PL_restartop == PL_op->op_next); /* we expect these to match */
+            PL_restartjmpenv = restartjmpenv;
+            PL_restartop = restartop;
```
bram-perl: Hmm.. right now I'm not yet convinced if this is the proper place to fix it.

What comes to mind:
- why is the fix not in `S_try_yyparse`?
- if `S_tty_yyparse` can cause the breakage then won't the same breakage be possible via `S_try_run_unitcheck`? (I haven't tested it yet)

dist/Attribute-Handlers/lib/Attribute/Handlers.pm
demerphq: Sure. NO problem. Ironically i deleted the print STDERR stuff I had in there before pushing. Ill put it back. :-)

perl.c
demerphq: Yeah, I did, but i can tweak it. That shouldnt happen, and i wanted to make it obvious when it does.

t/op/eval.t
demerphq: Doh. Good catch.

t/op/eval.t
demerphq: Yeah, willdo.

pp_ctl.c
demerphq: Sure.

pp_ctl.c
demerphq: I debated putting it in try_yyparse, but decided to leave it as a very clean light wrapper around yyparse.  I can do it the other way. Ill check S_try_rununitcheck when I get a chance.

dist/Attribute-Handlers/lib/Attribute/Handlers.pm
demerphq: Done in the latest push.

t/op/eval.t
demerphq: Fixed in the lastest push.

pp_ctl.c
demerphq: Fixed in latest push.

pp_ctl.c
demerphq: I moved it to try_yyparse(), and added equivalent logic to S_try_rununitcheck(). I couldn't find a test case for the UNITCHECK scenario but if I added an assert about PL_restartop and I could see it "leaking" into S_try_rununitcheck where I believe it shouldn't so I added it to that as well.

bram-perl:
On a non debugging build these commits cause two build time warnings:
```
pp_ctl.c: In function ‘S_try_yyparse’:
pp_ctl.c:3536:39: warning: unused parameter ‘caller_op’ [-Wunused-parameter]
 3536 | S_try_yyparse(pTHX_ int gramtype, OP *caller_op)
      |                                   ~~~~^~~~~~~~~
pp_ctl.c: In function ‘S_try_run_unitcheck’:
pp_ctl.c:3576:3***: warning: unused parameter ‘caller_op’ [-Wunused-parameter]
 3576 | S_try_run_unitcheck(pTHX_ OP* caller_op)
      |                           ~~~~^~~~~~~~~
```

bram-perl:
For `S_try_run_unit_check`:

**Without** your latest change:

- without -DDEBUGGING:
```
$ ./perl -wle 'INIT { eval "UNITCHECK { die; } print qq#A#;"; print qq#B#; print $@; } print "C";'
B
Died at (eval 1) line 1.
UNITCHECK failed--call queue aborted.

B

Segmentation fault
```
- with -DDEBUGGING
```
$ ./perl -wle 'INIT { eval "UNITCHECK { die; } print qq#A#;"; print qq#B#; print $@; } print "C";'
B
Died at (eval 1) line 1.
UNITCHECK failed--call queue aborted.

B

perl: pp_hot.c:5113: Perl_pp_leavesub: Assertion `((cx)->cx_u.cx_subst.sbu_type & 0xf) == 9' failed.
Aborted
```

**With** your latest change:

- without -DDEBUGGING
```
$ ./perl -wle 'INIT { eval "UNITCHECK { die; } print qq#A#;"; print qq#B#; print $@; } print "C";'
B
Died at (eval 1) line 1.
UNITCHECK failed--call queue aborted.

C
```
- with -DDEBUGGING
```
$ ./perl -wle 'INIT { eval "UNITCHECK { die; } print qq#A#;"; print qq#B#; print $@; } print "C";'
B
Died at (eval 1) line 1.
UNITCHECK failed--call queue aborted.

C
```

(Or in other words; that might be a suitable test case to add)

demerphq:
@bram-perl fixed and pushed. Thanks for the test case.

demerphq:
Unless someone objects I will merge this bug fix tomorrow without approval, or sooner if someone acks the patch.

dist/Attribute-Handlers/lib/Attribute/Handlers.pm
```diff
@@ -241,20 +241,33 @@ sub _apply_handler_AH_ {
...
-		$data = $evaled unless $@;
+                # keeping the minimum amount of code inside the eval string
+                # makes debugging perl internals issues with this logic easier.
+                my $code= "package $pkg; my \$ref= [$data]; \$data= \$ref; 1";
+                print STDERR "Evaling: '$code'\n"
+                    if $debug;
+                local $SIG{__WARN__} = sub{ die };
+                no strict;
+                no warnings;
+                # Note we do not need
```
tonycoz: What don't we need?

tonycoz:
Except for the incomplete comment it looks good to me.

dist/Attribute-Handlers/lib/Attribute/Handlers.pm
demerphq: Oh. Drat. Will fix. Thanks!

demerphq: Fixed. Thanks! Replaced with:
```
+                # Note in production we do not need to use the return value from
+                # the eval or even consult $@ after the eval - if the evaled code
+                # compiles and runs successfully then it will update $data with
+                # the compiled form, if it fails then $data stays unchanged. The
+                # return value and $@ are only used for debugging purposes.
+                # IOW we could just replace the following with eval($code);
```

PR #20400, opened 2022-10-14
Shorten Pull Requests title and body for IRC bot

atoomic:
As suggested by bram

This is reducing the body and title length sent to the irc bot.

### DIFF
--- a/.github/workflows/irc-notifications.yaml
+++ b/.github/workflows/irc-notifications.yaml
@@ -8,6 +8,7 @@ on: [push, pull_request]
 permissions: {}
 jobs:
   notify-irc:
+    timeout-minutes: 20
     runs-on: ubuntu-latest
     # only on main repo
     if: ( github.event.pull_request.head.repo.full_name == 'Perl/perl5' || github.repository == 'Perl/perl5' )
@@ -31,6 +32,7 @@ jobs:
           github_ref: ${{ github.event.ref }}
 
       - name: Setup commit message SUMUP env
+        id: sumup
         env:
           TXT: ${{ join(github.event.commits.*.message, '\n') }}
           C1: ${{ github.event.commits[0].message }}
@@ -60,7 +62,7 @@ jobs:
           my @l = split( "\n", $txt );
           my $max = 5;
           @l = ( @l[0..$max], "..." ) if @l > $max;
-          @l = grep { $_ !~ m{^EOF} } @l;
+          @l = grep { $_ !~ m{^EOS} } @l;
           say( join( "\n", @l ) );
           EOS
 
@@ -70,9 +72,9 @@ jobs:
 
           # -------------------------------------
           echo "# setup SUMUP environment variable"
-          echo 'SUMUP<<EOF' >> $GITHUB_ENV
+          echo 'SUMUP<<EOS' >> $GITHUB_ENV
           perl parse.pl     >> $GITHUB_ENV
-          echo 'EOF'        >> $GITHUB_ENV
+          echo 'EOS'        >> $GITHUB_ENV
 
           # -------------------------------------
           echo "# done"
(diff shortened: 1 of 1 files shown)


jkeenan:
We'll have to get our IRC bot working again before we can properly evaluate this.

atoomic:
The goal here is to fix the issues from the bot, by avoiding sending body or title too long.
I performed some tests and this is now good to merge

.github/workflows/irc-notifications.yaml
```diff
@@ -107,6 +107,41 @@ jobs:
...
+          my ( $key, $len ) = @ARGV;
+          $key //= ''; $len //= 50;
+          my $txt = substr( $ENV{$key} || "Unknown $key", 0, $len );
+          $txt =~ s{\\n}{\n}g; $txt =~ s{\\t}{ }g; $txt =~ s{\t}{ }g;
+          $txt =~ s{^EOS}{}mg;
+          say $txt;
+          EOS
+
+          # -------------------------------------
+          echo 'TITLE<<EOF'       >> $GITHUB_ENV
```
haarg: Seems like this should be using `::set-output` rather than the environment.

haarg:
The messages sent via the GitHub actions go to #p5p-commits. I haven't been in that channel, so I don't know what issues there may be there.

The messages sent to #p5p use an entirely different mechanism. There is a web hook contacting a [Notifico](https://n.tkte.ch/) service. That service seems broken for several things currently, including p5p, metacpan, and mojo.

.github/workflows/irc-notifications.yaml
atoomic: note: the `$GITHUB_ENV` is still documented and works fine
any reasons to stop using it?
view https://docs.github.com/en/enterprise-server@3.4/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

atoomic:
@haarg I ve enable both IRC workflows.

Initially I enabled notifico service, then I've added this built-in workflow so we can stop using notifico.
If notifico is not working for us anymore, we should add new expected notifications to this workflow or a new one.

notifico service / status is not related to that workflow and should not block the change here.

bram-perl:
> The messages sent via the GitHub actions go to #p5p-commits. I haven't been in that channel, so I don't know what issues there may be there.

The problem is (or was) that when opening a pull request it would include the entire body of the PR in the IRC message to the #p5p-commits channel.
That is fine for pull requests where the body is small but it broke badly when the body was large: it would connect; attempt to send the message and get killed for 'Excess flood'. It then continue to repeat that cycle for 6 hours and then it eventually failed the 'irc-notification' workflow.

So the end results was a workflow failure.
One such example: https://github.com/Perl/perl5/actions/runs/3236344540/jobs/5302023255
But more examples can be found via https://github.com/Perl/perl5/actions/workflows/irc-notifications.yaml?query=is%3Afailure


(And I suppose some of the above message could/should be included in the commit message)

.github/workflows/irc-notifications.yaml
```diff
@@ -107,6 +107,41 @@ jobs:
...
+        run: |
+          # -------------------------------------
+          echo "# script parse.pl"
+          cat <<'EOS' > parse.pl
+          use v5.14; use strict; use warnings;
+          my ( $key, $len ) = @ARGV;
+          $key //= ''; $len //= 50;
+          my $txt = substr( $ENV{$key} || "Unknown $key", 0, $len );
+          $txt =~ s{\\n}{\n}g; $txt =~ s{\\t}{ }g; $txt =~ s{\t}{ }g;
+          $txt =~ s{^EOS}{}mg;
```
bram-perl, line 127: It's non-obvious to me why this line (`$txt =~ s{^EOS}{}mg;`) is needed..
Who/what/where/when is `EOS` is added in the environment var?

bram-perl (commented):
I also propose adding a `timeout-minutes: 60` (or 30 or 120 or something) in the workflow.
Default timeout is 6 hours; having it attempt to deliver the message for that long seems a bit excessive..

(The CI jobs that do the build/test/.. are using `timeout-minutes:120`)

.github/workflows/irc-notifications.yaml
haarg: Outputs just seem like the more appropriate way to communicate between these actions. And I was wrong about `::set-output`, as it is deprecated. `$GITHUB_OUTPUT` would be the appropriate thing to use for outputs. But none of this is a big deal. If it's more annoying to use outputs, it doesn't need to be changed.

.github/workflows/irc-notifications.yaml
haarg: It seems like an attempt to avoid premature termination of the output by removing the string used as a delimiter. But the delimiter used to wrap the output of this string is `EOF`, not `EOS`.

atoomic: Correct I'm using a `heredoc` to define the environment variable with `EOS`.. (need to fix if non consistent, that s a mistake... EOS, EOF, END, ... )
the goal is to avoid a premature end, if the commit message contains a line starting with that `token`.

atoomic:
> I also propose adding a `timeout-minutes: 60` (or 30 or 120 or something) in the workflow. Default timeout is 6 hours; having it attempt to deliver the message for that long seems a bit excessive..
> 
> (The CI jobs that do the build/test/.. are using `timeout-minutes:120`)

yeap that s a great idea, will add it too, I think 15 to 30 minutes is more than we need

.github/workflows/irc-notifications.yaml
atoomic: I understand, not against using `$GITHUB_OUTPUT`. I'm not sure how to consume it later in other workflow without some research. Will have a look.

.github/workflows/irc-notifications.yaml
atoomic: this is now fixed and `EOS` is used everywhere

.github/workflows/irc-notifications.yaml
atoomic: giving it a try using 79728f9c044f61881c5c6eb25404c1245967512c 
this is going to require a longer syntax where it's used but should be fine and cleaner, thanks for the suggestion
`$steps.sumup.output.SUMUP` instead of `$env.SUMUP`

atoomic: @haarg after multiple attempts I was able to find the proper way to exchange variables between steps & jobs.
But I cannot make it work using multilines variables like the one currently used by sumup.
I'm going to keep using `GITHUB_ENV` for now, patch welcome.
thanks

atoomic:
this is now ready for review/merge

.github/workflows/irc-notifications.yaml
atoomic: here is probably the solution to use multiline variables for an output variable...
https://trstringer.com/github-actions-multiline-strings/
honestly the current usage of ENV make it much easier to maintain at this point

bram-perl:
Personally I would still like the commit message to be updated to explain the why (i.e. to avoid Excess flood);
other then that it LGTM

bram-perl (approved):

atoomic:
> Personally I would still like the commit message to be updated to explain the why (i.e. to avoid Excess flood);
> other then that it LGTM

I reworked the commit message to be more explicit on the intent

.github/workflows/irc-notifications.yaml
atoomic: feel free to reopen the discussion if needed

toddr (approved):

PR #20404, opened 2022-10-16
locale.c: Silence '-Wchar-subscripts' build-time warning on C++ builds

khwilliamson:
This warning is bogus, but better to avoid it.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -2336,8 +2336,10 @@ S_new_ctype(pTHX_ const char *newctype)
         if (bad_count == 2 && maybe_utf8_turkic) {
             bad_count = 0;
             *bad_chars_list = '\0';
-            PL_fold_locale['I'] = 'I';
-            PL_fold_locale['i'] = 'i';
+
+            /* The casts are because otherwise some compilers warn */
+            PL_fold_locale[ (U8) 'I' ] = 'I';
+            PL_fold_locale[ (U8) 'i' ] = 'i';
             PL_in_utf8_turkic_locale = TRUE;
             DEBUG_L(PerlIO_printf(Perl_debug_log, "%s is turkic\n", newctype));
         }


jkeenan:
> This warning is bogus, but better to avoid it.

@khwilliamson, can you provide the config_args you used and the warning you observed?

jkeenan:
On a hunch based on spotting the string `DEBUG` in your patch, I decided to compare my (i) simple and (ii) threaded debugging builds on OpenBSD in *blead*, with simple and threaded+debugging builds in the *smoke-me/khw-openbsd* branch.  In *either* branch, I got an `-Wunused-function` warning during `make test_prep` in a simple (*i.e.,* non-threaded, non-debugging) build that went away in a threaded+debugging build.

**blead:  simple**
```
$ /home/jkeenan/bin/perl/parse-build-warnings  4e362c6516.openbsd.maketp.output.txt.gz  
File:  4e362c6516.openbsd.maketp.output.txt.gz

[
  {
    char   => 1,
    group  => "Wunused-function",
    line   => 5946,
    source => "locale.c",
    text   => "unused function 'S_get_displayable_string'",
  },
  {
    char   => 42,
    group  => "Wunused-parameter",
    line   => 351,
    source => "./const-c.inc",
    text   => "unused parameter 'iv_return'",
  },
]
```
**blead: threaded debugging**
```
$ /home/jkeenan/bin/perl/parse-build-warnings  4e362c6516.openbsd.threaded.>
File:  4e362c6516.openbsd.threaded.debugging.maketp.output.txt.gz

[
  {
    char   => 42,
    group  => "Wunused-parameter",
    line   => 351,
    source => "./const-c.inc",
    text   => "unused parameter 'iv_return'",
  },
]
```
**smoke-me/khw-openbsd: simple**
```
$ /home/jkeenan/bin/perl/parse-build-warnings a7e2dd56b2.openbsd.khw-openbsd.maketp.output.txt.gz                                                                      <
File:  a7e2dd56b2.openbsd.khw-openbsd.maketp.output.txt.gz

[
  {
    char   => 1,
    group  => "Wunused-function",
    line   => 5953,
    source => "locale.c",
    text   => "unused function 'S_get_displayable_string'",
  },
  {
    char   => 42,
    group  => "Wunused-parameter",
    line   => 351,
    source => "./const-c.inc",
    text   => "unused parameter 'iv_return'",
  },
]
```
**smoke-me/khw-openbsd: threaded debugging**
```
$ parse-build-warnings a7e2dd56b2.openbsd.threaded.debugging.khw-openbsd.maketp.output.txt.gz
File:  a7e2dd56b2.openbsd.threaded.debugging.khw-openbsd.maketp.output.txt.gz

[
  {
    char   => 42,
    group  => "Wunused-parameter",
    line   => 351,
    source => "./const-c.inc",
    text   => "unused parameter 'iv_return'",
  },
]
```
(The `./const-c.inc` warnings above are from a cpan-upstream module and outside the scope of the current discussion.)

This makes me wonder whether the patch in this p.r. is actually accomplishing what the Subject of the ticket suggests it is doing.

khwilliamson:
Sorry for the poor commit message.  Updated.

jkeenan:
> Sorry for the poor commit message. Updated.

Which configure arguments?  Which smoke test report (URL)?  I haven't observed any of these warnings in my own OpenBSD builds and I don't have enough information to reproduce them.

tonycoz:
> > Sorry for the poor commit message. Updated.
> 
> Which configure arguments? Which smoke test report (URL)? I haven't observed any of these warnings in my own OpenBSD builds and I don't have enough information to reproduce them.

https://www.nntp.perl.org/group/perl.daily-build.reports/2022/10/msg285932.html

It requires a C++ build, as character constants are of type int in C, but type char in C++.

The warning is technically correct, but fairly pointless in this case since values here are always positive.

The warning is intended to prevent programmers from supplying a char type value as a subscript, since char may be signed, possibly resulting in a negative subscript (which is bad).

gcc has open tickets on problems with this warning:
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99950
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94182

tonycoz (approved):
Maybe the comment could be expanded to mention C++ compilers complain about the char subscript.

jkeenan:
>  [snip]
> https://www.nntp.perl.org/group/perl.daily-build.reports/2022/10/msg285932.html
> 
> It requires a C++ build, as character constants are of type int in C, but type char in C++.
> 

@Tonycoz, Thanks, that gave me the information I needed to reproduce the warning.  It turns out that the warning is not specific to OpenBSD, though I was able to reproduce it there compiling with clang++13.  I was able to reproduce it on Linux (Debian 11) using `clang++-13` as the compiler (threaded build) (adjusting ticket subject and labels accordingly).

**blead**
```
$ parse-one-warnings-group 4e362c6516.linux.threaded.clang++-13.maketp.output.txt.gz Wchar-subscripts
File:  4e362c6516.linux.threaded.clang++-13.maketp.output.txt.gz

[
  {
    char   => 27,
    group  => "Wchar-subscripts",
    line   => 2339,
    source => "locale.c",
    text   => "array subscript is of type 'char'",
  },
  {
    char   => 27,
    group  => "Wchar-subscripts",
    line   => 2340,
    source => "locale.c",
    text   => "array subscript is of type 'char'",
  },
]
```
**smoke-me/khw-openbsd**
```
$ parse-one-warnings-group 3233666865.linux.threaded.clang++-13.khw-openbsd.maketp.output.txt.gz Wchar-subscripts
File:  3233666865.linux.threaded.clang++-13.khw-openbsd.maketp.output.txt.gz

Warnings group 'Wchar-subscripts' not found at /home/jkeenan/bin/perl/parse-one-warnings-group line 31.
```
So, once the commit message is updated to reflect that this was observed with clang++, this p.r. should be mergeable.  Thanks for the report and the investigation.

khwilliamson:
Merged via 08c664df7c07619c9a7ab8039b4d42f20a3d9adf with both comments and commit message updated

PR #20405, opened 2022-10-17
perlsub - replace uses of $a and $b in examples

Grinnz:
replace $a and $b in examples to avoid recommending their use, since they aren't treated as normal variables by strict. also replace $c and $d when used in the same examples so it looks consistent.


demerphq (approved):

demerphq:
Do I need to merge for you?

jkeenan (approved):

PR #20406, opened 2022-10-17
Refactor regen/opcode.pl little bit

happy-barney:
This PR is preparation for removal of "unimplemented" `#ifdef`s from `PP()` functions

- move misc generators into functions for better readability
- use c99 array init for better readability (even generated files should be readable)

Contains one POC - demonstration of usage of module which should be part of stdlib - [Syntax::Construct](https://metacpan.org/pod/Syntax::Construct)

Following steps
- extend existing `PP(unimplemented_op`) with support for more categories than just `socket`
- implementation function (`PL_ppaddr`) will be provided by platform
- `opcode.pl` will provide only default values
- bonus - mechanism will allow deliberate exclusion of some opcodes (eg: local filesystem access)

### DIFF
--- a/regen/opcode.pl
+++ b/regen/opcode.pl
@@ -19,6 +19,9 @@
 use v5.26;
 use warnings;
 
+sub generate_opcode_h;
+sub generate_opcode_h_prologue;
+
 my $restrict_to_core = "if defined(PERL_CORE) || defined(PERL_EXT)";
 
 BEGIN {
@@ -920,7 +923,7 @@ require './regen/op_private';
 #use Data::Dumper;
 #print Dumper \%LABELS, \%DEFINES, \%FLAGS, \%BITFIELDS;
 
-print $oc "#$restrict_to_core\n\n";
+generate_opcode_h;
 
 # Emit defines.
 
@@ -1235,5 +1238,13 @@ OP_PRIVATE::print_B_Op_private($oprivpm);
 
 foreach ($oc, $on, $pp, $oprivpm) {
     read_only_bottom_close_and_rename($_);
+}
+
+sub generate_opcode_h {
+    generate_opcode_h_prologue;
+}
+
+sub generate_opcode_h_prologue {
+    print $oc "#$restrict_to_core\n\n";
 }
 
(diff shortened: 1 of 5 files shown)


regen/opcode.pl
```diff
@@ -1309,3 +1267,46 @@ sub generate_opnames_h_opcode_enum {
...
+	print $on <<~'EO_OP_IS_COMMENT';
+
+	/* the OP_IS_* macros are optimized to a simple range check because
+	    all the member OPs are contiguous in regen/opcodes table.
+	    opcode.pl verifies the range contiguity, or generates an OR-equals
+	    expression */
+	EO_OP_IS_COMMENT
+
+	my %OP_IS_SOCKET;					 # /Fs/
+	my %OP_IS_FILETEST;					 # /F-/
```
demerphq: comment is indented more than the ones that follow it

regen/opcode.pl
```diff
@@ -1336,3 +1325,17 @@ sub generate_pp_proto_h {
...
+		by          => 'regen/opcode.pl',
+		from        => "data in\nregen/op_private and pod embedded in regen/opcode.pl",
+		style       => '#',
+		file        => 'lib/B/Op_private.pm',
+		copyright   => [2014 .. 2014],
+	});
+
+	OP_PRIVATE::print_B_Op_private($oprivpm);
+
+    read_only_bottom_close_and_rename($oprivpm);
```
demerphq, line 1352: mixed indentation?

demerphq (changes requested):
I like the way you broke this up into small patches. One thing id suggest "for next time" is to separate the two refactorings (output and script) into two series. With the number of changes to opcode.h its a bit hard to tell if its all good or not. It looks ok, but it would be nice to be able to see that the refactored code could produce the old file, and then see the changes to the refactored code to make it out the new file separately.

Anyway, i noticed two whitespace glitches, other than that the code looks reasonable and nicely refactored.  Happy to approve once those glitches are sorted.

regen/opcode.pl
happy-barney: original code indented comments were aligned with tabs, untabified

regen/opcode.pl
happy-barney: tab vs space indentation, untabified

demerphq (approved):

demerphq:
Thanks for the patch, applied!

Issue #20389, opened 2022-10-12
Double-exception `defer` collision loses information


leonerd:
(largely copied from https://rt.cpan.org/Ticket/Display.html?id=144760)

When a `defer` block throws an exception due to stack unwind of an existing exception, we entirely lose one of them:

```
$ cat exception-collision.pl 
#!/usr/bin/perl

use v5.36;
use experimental 'defer';

sub f
{
   local $SIG{__WARN__} = sub {
      print STDERR "WARN: $_[0]";
   };
   local $SIG{__DIE__} = sub {
      print STDERR "DIE: $_[0]";
   };

   defer {
      die "Incidental failure during shutdown";
   }

   die "The real reason your program failed";
}

eval { f(); 1 } or
   print STDERR "EVAL failure: $@";
```

```
$ perl -Mblib exception-collision.pl
DIE: The real reason your program failed at exception-collision.pl line 19.
DIE: Incidental failure during shutdown at exception-collision.pl line 16.
EVAL failure: Incidental failure during shutdown at exception-collision.pl line 16.
```

The outer EVAL has totally lost the information. It didn't even print as a warning.

Moreover, the real failure got lost, leaving us only the incidental failure. In a real program, whatever the real failure was may well have caused some knock-on effects meaning the defer block won't run either. We've now forgotten it.

leonerd:
Largely copying notes from the XS module version of this bug:

Solving this is going to be difficult in two ways:

 1) Reliably detecting that the stack-unwind is happening due to an exception unwind

    This is doable in a hacky way currently with `SvTRUE(ERRSV)` but it's prone to false positives by any code which does simply `$@ = 1`.

    Core perl does not have a reliable mechanism to know this, but does at least have a comment stating that we *know* we don't have such a mechanism. https://github.com/Perl/perl5/blob/1932805f63d6181a90fe36fabbdc3755a78b072f/pp_ctl.c#L1776-L1782

    Solving this bug would require adding such a mechanism, and may be useful in other cases too.

 2) Invoking the defer block with an EVAL context around it.

    This may be possible with some hackery involving `SETJMP` and creating a new jmpenv. It may need some of the internal static functions in `pp_ctl.c` but fortunately all of the core-related machinery to do with defer is in the same file, so it is visible there.

[...]

PR #20407, opened 2022-10-17
Add `PL_throwing`, true during exceptional stack unwind

leonerd:
This PR adds a mechanism by which code that runs during stack unwind (`finally` and `defer` blocks at the Perl layer, `SAVEDESTRUCTOR*` at the XS layer) can inspect whether that stack unwind is happening because of a thrown exception. Or, at least, whether *a* stack unwind because of a thrown exception is happening - not necessarily "this" one.

Such a mechanism is a prerequisite to solving #20389, and also hinted at by the comment https://github.com/Perl/perl5/blob/1932805f63d6181a90fe36fabbdc3755a78b072f/pp_ctl.c#L1776-L1782

Still undecided:

 * How it should behave during nested (normal) unwinds invoked as part of an exceptional one
 * Whether a pure-perl wrapping should be added; perhaps by a interpreter-global variable `${^THROWING}`

### DIFF
--- a/ext/XS-APItest/t/throwing.t
+++ b/ext/XS-APItest/t/throwing.t
@@ -2,7 +2,12 @@
 
 use v5.36;
 
-use Test::More;
+# Test::More doesn't have fresh_perl_is() yet
+
+BEGIN {
+    require '../../t/test.pl';
+};
+
 use XS::APItest;
 
 package RunOnDestruct {
@@ -49,6 +54,30 @@ sub run_on_destruct :prototype(&) {
     ok(!XS::APItest::PL_throwing, 'PL_throwing false after eval returned');
 }
 
+# PL_throwing during string-eval()
+{
+    my $var;
+    eval q(
+        my $guard = run_on_destruct { $var = 0 + XS::APItest::PL_throwing };
+        die "Oopsie";
+    ); # ignore the error
+    is($var, 1, 'var is true after die in string-eval');
+    ok(!XS::APItest::PL_throwing, 'PL_throwing false after string-eval returned');
+}
+
+# It's (probably) not possible to observe PL_throwing=true during a BEGIN,
+# CHECK or INIT phaser, but we can test END
+fresh_perl_is(
+    <<'EOF',
+use strict;
+use XS::APItest;
+END { printf "PL_throwing=%s\n", XS::APItest::PL_throwing ? "true" : "false" }
+die "Oopsie\n";
+EOF
+    "Oopsie\nPL_throwing=true",
+    { stderr => 1 },
+    'PL_throwing during END block');
+
 # PL_throwing during normal defer
 {
     use experimental 'defer';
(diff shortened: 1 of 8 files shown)


demerphq:
Shouldn't PL_throwing be true when we unwind due to an exception whether or not we are inside of an eval?

leonerd:
> Shouldn't PL_throwing be true when we unwind due to an exception whether or not we are inside of an eval?

Ah, yes it should. I'll move the assignment further out.

leonerd:
More tests added; @demerphq PTAL

pp_ctl.c
```diff
@@ -1744,8 +1744,15 @@ Perl_die_unwind(pTHX_ SV *msv)
...
 
+    bool was_throwing = PL_throwing;
+    PL_throwing = true;
+
     if (in_eval) {
         I32 cxix;
+        /* We can't use the regular SAVEBOOL() mechanism for this because we
+         *   are currently unwinding the save stack. We will have to do it the
+         *   old-fashioned way
+         */
```
demerphq, line 1755: i think this comment should move up with the code no?

ext/XS-APItest/t/throwing.t
```diff
@@ -0,0 +1,159 @@
...
+}
+
+# PL_throwing at top-level defer
+fresh_perl_is(
+    <<'EOF',
+use strict;
+use experimental 'defer';
+use XS::APItest;
+defer { printf "PL_throwing=%s\n", XS::APItest::PL_throwing ? "true" : "false" }
+die "Oopsie\n";
```
demerphq, line 113: I think you might want to make this into a loop and wrap these two lines in CHECK/UNITCHECK/BEGIN/INIT/END blocks and make sure that they all the do the right thing. I also think you might want to check nesting the blocks, Eg, `X { eval "Y { defer ...; die ..; }" ... } ...` where X and Y are the list of phase blocks and validate that they all work as you expect. We have found that these combinations can lead to weird results before. Another thing to check is that they work under the debugger. I would take a look at t/op/eval.t and t/comp/retainedlines.t for examples.  Also see the tests in https://github.com/Perl/perl5/pull/20398

demerphq:
@leonerd i know this is a draft, but it conflicts and seems stalled. Do you have plans to pick this up again? Can we at least get it updated to latest blead?

demerphq:
Since some of the conflicts where from my embed related patches I took the liberty of rebasing and resolving the conflicts to save you the hassle @leonerd. I didn't change anything else.

leonerd:
@demerphq Much thanks :) I'll take a look again and see what the next steps ought to be

demerphq:
This is conflicting over the perldelta. I guess there isnt much point in fixing it until tomorrow.

jkeenan:
@leonerd there has been no discussion in this draft p.r. in a year-and-a-half, and the code has re-acquired merge conflicts.  Do you still plan to pursue it?  Thanks.

Issue #20411, opened 2022-10-18
Some obscure Segmentation fault on a long file


AnFunctionArray:
I still haven't exactly figured out why but here is some debug info nevertheless:

```
2117462 <1];%ne> <xtern char>   |493641|  1269:SUSPEND(1291)
2117462 <1];%ne> <xtern char>   |493642|   1271:CURLYN[62]{0,INFTY}(1289)
                             |493642|   ANYOF[0-9@A-Z_a-z] can match 5 times out of 2147483647...
2117467 <xtern> < char s_zo>    |493643|    1289:SUCCEED(0)
                             |493643|    SUCCEED: subpattern success...
2117467 <xtern> < char s_zo>    |493641|  1291:SUCCEED(0)
                             |493641|  SUCCEED: subpattern success...
2117467 <xtern> < char s_zo>    |493640| 1293:BOUND(1294)
2117467 <xtern> < char s_zo>    |493640| 1294:CLOSE61 'identifierpure'(1296)
                             |493640| END: EVAL trying tail ... (cur_eval=1a8a38ee0)
2117467 <xtern> < char s_zo>    |493641|  4017:CLOSE190 'identf'(4019)
2117467 <xtern> < char s_zo>    |493641|  4019:LOGICAL[2](4020)
2117467 <xtern> < char s_zo>    |493641|  4020:EVAL(4023)
EVAL/GOSUB: Matching embedded REx "" against " char s_zone5_master__trigger_17_0018552c[1];%nextern char s"...
2117467 <xtern> < char s_zo>    |493642|   1:NOTHING(2)
2117467 <xtern> < char s_zo>    |493642|   2:END(0)
                             |493642|   END: EVAL trying tail ... (cur_eval=1a8a38ee0)
2117467 <xtern> < char s_zo>    |493643|    4023:SUSPEND(4029)
2117467 <xtern> < char s_zo>    |493644|     4025:STAR(4027)
                             |493644|     POSIXD[\s] can match 1 times out of 2147483647...
2117468 <tern > <char s_zon>    |493645|      4027:SUCCEED(0)
                             |493645|      SUCCEED: subpattern success...
2117468 <tern > <char s_zon>    |493643|    4029:CLOSE189 'identiiferfast'(4031)
                             |493643|    END: EVAL trying tail ... (cur_eval=1a8a32fa0)
2117468 <tern > <char s_zon>    |493644|     4037:LOGICAL[1](4038)
[...]

Issue #20104, opened 2022-08-17
WHILEM: matched 65535 out of 0..65535


AnFunctionArray:
So I was wondering why it wasn't matching at a an obvious match - apparently I had reached the limit wow.

AnFunctionArray:
I don't know if it's a bug but it's an interesting corner case at least.

AnFunctionArray:
The obvious solution is to use recursion at the end like: `(?<pattern>(something)*+(?&pattern)?+)`.

demerphq:
Recently we merged a patch sequence that means that it should be trivial to replace all of these short integer counts with proper sized ones. @khwilliamson @hvds

AnFunctionArray:
@demerphq Can you clarify - I recently run into the same problem again.

AnFunctionArray:
Ok - I'm not sure but it looks like this solution works:

`(?<pattern>(?&something)*+((?&something)(?&pattern))?+)`

PR #20412, opened 2022-10-19
Fixed signed mgs_ix added to pointer

AnFunctionArray:
Which in certain cases results in subtraction which is not intended.

Fixed by making all instances of mgs_ix to be unsigned - U32 #20411

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -71,6 +71,7 @@ Alexander D'Archangel          <darksuji@gmail.com>
 Alexander Foken
 Alexander Gernler              <alexander_gernler@genua.de>
 Alexander Gough                <alex-p5p@earth.li>
+Alexander Nikolov              <sasho648@gmail.com>
 Alexander Hartmaier            <abraxxa@cpan.org>
 Alexander Klimov               <ask@wisdom.weizmann.ac.il>
 Alexander Smishlajev           <als@turnhere.com>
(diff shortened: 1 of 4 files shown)


tonycoz:
If SSNEW() is returning a number that can't be represented in an I32 then several other things are going to be broken, if nothing else PL_savestack_max is likely to overflow..

AnFunctionArray:
@tonycoz Yeah well tests run fine on my commit:

https://github.com/AnFunctionArray/perl5/actions/runs/3278656896/jobs/5397296932

Plus this fixes the seg-fault on me - before it crashes - after it doesn't.


I think what I've created suffices before the actual problem is inside `save_magic_flags` and `SSPTR`.

AnFunctionArray:
@tonycoz If you are so sure my patch is broken - I would suggest to put your money where you mouth is and allow the tests to run.

AnFunctionArray:
But I do kinda agree it would be better to handle this in a bit more depth - so it can not only not segfault but this


https://github.com/Perl/perl5/issues/20104

To be also somehow configurable (although I'm not sure if this is rather an issue with the regex  - since a workaround (given there by me) generally works).


**However this patch as I've stated fixed the segfault and perl gracefully exits prematurely when the match counter is hit**

AnFunctionArray:
@tonycoz Although tbh I don't think it should even get to this case currently with the tests that are present (since it wasn't reflected before anyway).

The issue is that I wasn't able to exactly pinpoint the problematic set of conditions for it to happen. 

I tried creating a sample code but currently it only crashes with my full program.

tonycoz:
I believe this is already covered by #20473

PR #20413, opened 2022-10-19
overload: only disable builtin warnings where blessed is called

haarg:
Until recently, overload did not enable or disable any warnings.
Instead, it would rely on the value of $^W, or the -w flag. When
switching to use the builtin::blessed function, the code was changed to
disable 'experimental::builtin' warnings. Since no other warnings are
changed, the current active warning set is combined with this one
disabled category, and baked in for the entire module. The currently
active warning set is determined by $^W, so whatever its value was when
overload was compiled would get baked in, and any runtime changes to $^W
would no longer have any impact.

The particular warning that would most commonly be emitted by overload
is the sub redefinition warning, in case an overload was being
overwritten. Various modules, such as JSON::PP and Cpanel::JSON::XS,
were expecting to be able to overwrite overloads without any warnings.
They would set $^W to false to prevent them. After the "no warnings
'experimental::builtin';" line was added, it became impossible to
disable these warnings at runtime.

Move the disabling of the experimental warnings to a smaller scope. This
means that the rest of the module is restored to its previous state of
having warnings controlled by $^W.

Also cleans up the writespace in overload.pm.

### DIFF
--- a/lib/overload.pm
+++ b/lib/overload.pm
@@ -3,7 +3,7 @@ package overload;
 use strict;
 no strict 'refs';
 
-our $VERSION = '1.35';
+our $VERSION = '1.36';
 
 our %ops = (
     with_assign         => "+ - * / % ** << >> x .",
--- a/lib/overload.pm
+++ b/lib/overload.pm
@@ -30,152 +30,152 @@ my %ops_seen;
 sub nil {}
 
 sub OVERLOAD {
-  my $package = shift;
-  my %arg = @_;
-  my $sub;
-  *{$package . "::(("} = \&nil; # Make it findable via fetchmethod.
-  for (keys %arg) {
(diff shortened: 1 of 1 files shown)


haarg:
An alternative to this would be to add a `use warnings;` or `no warnings;` at the top of the file. That would at least make the warnings issued consistent. I don't think the current state of the $^W value getting baked in at compile time is reasonable.

leonerd:
Oh that's messy - probably my fault when I added the `builtin::blessed` stuff. I didn't realise `$^W` had that silly behaviour.

Reducing the scope like this seems fine enough to avoid the immediate breakage problem, though indeed perhaps we should (separately) have another look at the wider logic here and see if we can do it better.

leonerd (approved):
LGTM

PR #20415, opened 2022-10-19
Configure `I32df`, `U32uf`, etc. in `config.h` rather than in `perl.h`

t-a-k:
`I32df`, `U32uf`, and similar macros had been defined in `perl.h` using preprocessor conditionals, but determining wheter `I32 `is `int` or `long` is pretty hard with preprocessor when `INTSIZE` == `LONGSIZE`.
I think it is better to make `Configure` script determine whether `%d` or `%ld` shall be used to format `I32` value, as it should know exact underlying type of `I32`.

Without this change, `U32` is typedef'ed to `unsigned long` and `U32uf` is defined to `"u"` for `-Duse64bitint` build on 32-bit host, which leads a lot of build warnings about mismatches between `LINE_Tf` (= `U32uf`) format and `line_t` (= `U32`) argument.


tonycoz (approved):

khwilliamson:
I think these should be squashed into a single commit; but otherwise LGTM

t-a-k:
I'm sorry that there is a mistake in `uconfig64.sh`. I'll fix it (and rebase to current blead).

khwilliamson:
@t-a-k Please squash the commits into one, and it will get merged

demerphq:
Doesnt some of this need to be coordinated with @Tux ?

jkeenan:
Do we have any way of testing the changes to the two files in this p.r. pertaining to the plan9 OS?

tonycoz:
> Doesnt some of this need to be coordinated with @Tux ?

As long as you let him know, I believe we can just commit, and he'll backport.

> Do we have any way of testing the changes to the two files in this p.r. pertaining to the plan9 OS?

If anyone is interested in setting up a VM http://9legacy.org/download.html but I'm not.

Tux:
> Doesnt some of this need to be coordinated with @Tux ?

Yes it does. The change to Configure and related files looks fine, except for the omission of the entries in the Glossary.
If you(plural) add those, go ahead and I'll backport

t-a-k:
> If you(plural) add those, go ahead and I'll backport

I've pushed a commit (2694c7d8603467619fbf7a0d12a2a6dc3ff66424) to add `Porting/Glossary` entries.
(Another push is to update `perldelta`.)

If this is okay, I will squash the commits into one.

PR #20417, opened 2022-10-20
t/test.pl: Fix threads signalling watchdog cancellation

khwilliamson:
Commit 4712a4c94e5974bba8ef97d68aadf2b99c0b4ff4 added the ability to cancel a watchdog timer.  Unfortunately it doesn't work properly when the timer method was using the thread signalling cancellation method. This problem only shows up on slow systems under load, so the flaw wasn't immediately discovered.  That commit lowered the incidence level significantly.

This commit solves the problem by having the child thread detach itself, after it sets up the needed code to handle a signal from the parent. The parent thread signals the child, as currently when it is done processing, but waits for the child being detached.  That is a cue to the parent that it is safe to signal the child.

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -1932,7 +1932,7 @@ sub watchdog ($;$)
         while (     $watchdog_thread->is_running()
                && ! $watchdog_thread->is_detached())
         {
-            sleep 0;    # Relinquishes control, with minimal delay
+            'threads'->yield();
         }
 
         return;
--- a/t/test.pl
+++ b/t/test.pl
@@ -1907,6 +1907,10 @@ sub watchdog ($;$)
 
                 $SIG{'KILL'} = sub { threads->exit(); };
 
+                # Detach after the signal handler is set up; the parent knows
+                # not to signal until detached.
+                'threads'->detach();
+
                 # Execute the timeout
                 my $time_left = $timeout;
                 do {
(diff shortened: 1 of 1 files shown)


t/test.pl
```diff
@@ -1743,6 +1743,16 @@ ($;$)
...
     if ($timeout == 0) {
         if ($watchdog_thread) {
+
+            # Don't kill the watchdog until it has set up the signal handler.
+            # It tells us that it has done so by detaching itself immediately
+            # afterwards.  (The 'is_running()' is a fail-safe.)
+            while (     $watchdog_thread->is_running()
+                   && ! $watchdog_thread->is_detached())
+            {
+                sleep 1;
```
demerphq: Is that a long time? sleep 0.1 would be more reactive, and wouldnt burden the system that much more in the grand scheme of things.

demerphq:
Nice catch btw.

t/test.pl
khwilliamson: I didn't realize that a non-integer was possible.  But I tried it, and it didn't work, and then I looked at perlfunc, and it takes an integer argument.  We don't probably want to load some module in t/test.pl that could sleep less than a second.

demerphq:
Yeah, sorry, i was being dumb. to do sub-second sleep you need Time::HiRes. Its a pity, I wonder if it would  be reasonable to merge that functionality into core, seems like a useful thing to have out of the box.

t/test.pl
demerphq: Yeah, my bad. That would need Time::HiRes, and of course we cant load that in this code. Sorry for the noise.

tonycoz (approved):

khwilliamson:
I want to make sure that sleep 0 is acceptable.

tonycoz:
> I want to make sure that sleep 0 is acceptable.

`threads->yield()` might be better.

khwilliamson:
yield() it is

khwilliamson:
merged via dd8f4e6b1cdd0b2bce8bb0601054d575b69b4265

PR #20421, opened 2022-10-21
README: make the GitHub and root directory match

rjbs:
…by using a symlink!


leonerd (approved):
LGTM

book (approved):

rwp0 (approved):

Grinnz:
The symlink is not necessary - github will use the root README.md, the only reason to also have one in .github is so it doesn't need to be in root.

Grinnz:
Oh I missed that the root README is not being renamed to README.md. I'm not sure this will work with the symlink.

Grinnz:
It seems like it is working at https://github.com/Perl/perl5/tree/42557cf44325545afd1c1455ebfeecbc2d23157c but I'm not sure if that is indicative (I've been fooled before)

PR #20423, opened 2022-10-21
Fix fegetround on s390x with clang

tstellar:
Clang's s390x backend hard codes FLT_ROUNDS to one.  This fixes fenv.t test failures with building with clang on s390x.

### DIFF
--- a/ext/POSIX/POSIX.xs
+++ b/ext/POSIX/POSIX.xs
@@ -1021,7 +1021,7 @@ static NV my_log2(NV x)
 /* XXX nexttoward */
 
 /* GCC's FLT_ROUNDS is (wrongly) hardcoded to 1 (at least up to 11.x) */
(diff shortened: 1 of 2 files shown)


jkeenan:
> Clang's s390x backend hard codes FLT_ROUNDS to one. This fixes fenv.t test failures with building with clang on s390x.

@tstellar you have one test failure in a Porting (metadata) test.  Can you go into the file cited below, increment the value of `$VERSION` to be `2.08`, then commit and force push?
```
ext/POSIX/lib/POSIX.pm
our $VERSION = '2.07';
```
List: While this change looks plausible to me, I wonder how we would test it on s390x.

khwilliamson:
I think we have to take @tstellar's word for it.

tstellar:
I did  a test build with the patch here: https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/clang-built-f37/build/4962967/

PR #20424, opened 2022-10-22
fix ExtUtils::ParseXS compatibility with perl < 5.8.8

haarg:


tonycoz (approved):
This looks sane, but will probably need a version bump.

Issue #20425, opened 2022-10-22
Perl_no_op() warnings call Perl_warner() more than once, which interferes with warnings fatal and __WARN__ hooks


philiprbrenan:
Without warnings enabled we get a helpful message telling us the name of the missing bareword such as  `Do you need to predeclare id_deeply?` in the following:
```
#!/usr/bin/perl
use v5.30;
sub pick{}

id_deeply pick;

# Bareword found where operator expected at ... line 5, near "id_deeply pick"
# 	(Do you need to predeclare id_deeply?)
# syntax error at ... line 5, near "id_deeply pick"
```

But with `warnings FATAL` enabled this helpful message is omitted. 

```
#!/usr/bin/perl
use v5.30;
use warnings FATAL => qw(all);
sub pick{}

id_deeply pick;

# Bareword found where operator expected at ... line 6, near "id_deeply pick"
```

Please could you provide this useful information even when warnings are enabled?  Doing so would save programmer time when it is not immediately obvious which of the two words is the one causing the problem.

demerphq:
Hi thanks for the report. I agree with you, this is improper behavior I have pushed a patch to fix this: #20427. I have also done an audit for other cases, and I didn't spot any of the same form (eg adding warning data in parenthesized hints after the main warning message). This is problematic not just for fatal warnings but also for `$SIG{__WARN__}` which gets called twice, with parts of the overall message, from what is supposed to be a single warning.

demerphq:
thanks, this has been fixed in blead and will be available in the next release of perl.

PR #20427, opened 2022-10-23
Make sure missing operator and other "Perl_no_op" related warnings are dispatched properly.

demerphq:
Currently we construct a message like this:
```
Bareword found where operator expected at -e line 1, near "foo bar"
	(Do you need to predeclare foo?)
```
from code like this:
```
use strict; 
use warnings; 
sub bar {}
foo bar;
```
This is problematic as it means we call `$SIG{__WARN__}` multiple times with parts of the same error message:
```
$ perl -e'BEGIN{$SIG{__WARN__}= sub { print STDERR "warn: $_[0]" };} use strict; use warnings; sub bar {} foo bar;'
warn: Bareword found where operator expected at -e line 1, near "foo bar"
warn: 	(Do you need to predeclare foo?)
syntax error at -e line 1, near "foo bar"
```
It is also problematic as it means that with fatalized warnings the first fragment triggers the upgrade to a die, and we never get to the second part of the error message. (As is implied by the fact that `$SIG{__WARN__}` is called twice.)

This PR fixes problem by emitting the parenthesized "hint" message as part of the main diagnostic, eg like this:
```
$ ./perl -Ilib -e'BEGIN{$SIG{__WARN__}= sub { print STDERR "warn: $_[0]" };} use strict; use warnings; sub bar {} foo bar;'
warn: Bareword found where operator expected (Do you need to predeclare "foo"?) at -e line 1, near "foo bar"
syntax error at -e line 1, near "foo bar"
```

At the same time we quote the subject (var or bareword) of the hint, and we also strip whitespace from its start. The old code was sloppy and didn't do so consistently. This change is included in the main patch, as it affects the same test files, and there are a lot of them and I didn't want to manually adjust them several times. 

This PR also include a couple of auxiliary patches to allow some of the involved tests files to be run from the root directory of the repo. It looks like a few of our test files were broken in this sense when we removed `.` from `@INC`.

This resolves GH Issue #20425

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -218,6 +218,23 @@ XXX Changes (i.e. rewording) of diagnostic messages go here
 
 =item *
 
+Various warnings that used to produce parenthesized hints underneath the
+main warning message and after its "location data" were chanaged to put
+the hint inline with the main message. For instance:
+
+ Bareword found where operator expected at -e line 1, near "foo bar"
+     (Do you need to predeclare foo?)
+
+will now look like this but as one line:
+
+ Bareword found where operator expected (Do you need to predeclare
+ foo?) at -e line 1, near "foo bar"
+
+as a result such warnings will no longer trigger C<$SIG{__WARN__}>
+twice, and the hint will be visible when fatal warnings is in effect.
+
+=item *
+
 XXX Describe change here
 
 =back
--- a/t/lib/croak/toke
+++ b/t/lib/croak/toke
@@ -2,8 +2,7 @@ __END__
 # NAME foo found where operator expected
 myfunc 1,2,3
 EXPECT
-Number found where operator expected at - line 1, near "myfunc 1"
-	(Do you need to predeclare myfunc?)
+Number found where operator expected (Do you need to predeclare "myfunc"?) at - line 1, near "myfunc 1"
 syntax error at - line 1, near "myfunc 1"
 Execution of - aborted due to compilation errors.
 ########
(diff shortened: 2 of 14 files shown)


tonycoz:
This makes the `$SIG{__WARN__}` behaviour more consistent, but I do think it makes the extra note, which was formatted to better stand out, less readable.

demerphq:
I dunno, personally while i agree its a bit more "dense" I think it makes more sense and likely more comprehensible overall. Having the location data *before* the explanation is exceptional (no other warning does that), and obviously it doesn't show at all with `use warnings 'fatal';`, leaving aside the double call to `$SIG{__WARN__}`.

Personally i don't think the fairly significant changes to the error logic to fake up the "old style" and correct the other issues is worth the effort. (Well unless it was in the context of reworking ALL of our errors to go through a standard set of code paths instead of the plethora of helpers and the like.)

Note we have many cases of parenthesized error hints being on the same line as the main error message, so this isnt precedent setting at all, in fact it removes the one exception to the rule that exists. (It wouldnt surprise me if the other ones were also multi-call at one point and changed to single call for similar reasons to this ticket.)

tonycoz:
If they were changed from multiple dispatch of the primary warning and any note, perhaps that was a mistake too.

The primary audience of a diagnostic message is a person, imagine if:
```
In file included from ../../perl.h:7735,
                 from re_comp.c:132:
re_comp.c: In function ‘my_reg_named_buff_fetch’:
../../inline.h:108:14: warning: ‘ret’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  108 |     ary[key] = val;
      |     ~~~~~~~~~^~~~~
re_comp.c:8969:9: note: ‘ret’ was declared here
 8969 |     SV *ret;
      |         ^~~

```
was all one line.

Perhaps notes could be pushed onto an AV `PL_warn_notes`, so Perl_vwarn() would still add the location to the first line, and add the notes as extra lines.  `PL_warn_notes` would be cleared when a warning was dispatched.

demerphq:
> If they were changed from multiple dispatch of the primary warning and any note, perhaps that was a mistake too.

Given there is only one error message of this style left, this one, I seems to me the evidence supports the opposite conclusion, that this one is the mistake. I started doing an audit, we have a fair number with inline hints. Just one with a separate line hint.  Personally i am fine with us bulk reworking the error messages to be more readable. But if I was to do that I would first make them all consistent (this patch), /then/ work on a proposal to change them all. 

While sure, we could add some kind of PL_exception_hints variable, (note that hints are supplied for fatal exceptions also), and we could make all the exceptions that fire use it, the patch would be large, and the CPAN impact large too. I think if we are going to do this we should do it with more community involvement and consensus, and realize that it is likely a multi-release project to get right.  I also think /until/ we do it wholesale, we should make this error consistent. 

```
perl -le'foo->bar()'
Can't locate object method "bar" via package "foo" (perhaps you forgot to load "foo"?) at -e line 1.
```
this a perfect example. If we change this people *will* notice. I wanted to remove the hint entirely, people prostested, even though it is duplicative and often outright unhelpful. I do agree it would be less helpful if it was in the "indented" style, but i know for sure a bunch of things in the eco-system would notice if this was changed. 

> The primary audience of a diagnostic message is a person,

While I agree, I don't think comparing a C compiler, with static compilation, no `eval`, no `$SIG{__DIE__}`, no `$SIG{__WARN__}` is particularly apposite, while sure they are both compiling code and returning diagnostics, the context is so hugely different the comparison IMO breaks down.  The fact that Perl mixes compilation with execution means that the direct audience for an error might be the `$SIG{__WARN__}` hook that is pushing those warnings into a logging system, or munging them in some way, or the internals converting the warning into a fatal exception, or converting those text exceptions into exception objects. That means a standardized form, that is both machine processable and human friendly (ish, humans are easier to train than computers) form, has real benefits to everyone.  Having one particular case different just because isn't helpful at an eco system level even if it might be argued that it is more readable for a given user. 

I get what you are suggesting here. I think its worthy. I just dont think its appropriate to tangle it up with this bug and PR. I really think we should make this exception consistent with the rest for 5.38, in the same release we can make an announcement we will massively change the error messages in 5.40 (it could be the 40th birthday present to the community!) and then do your idea, but across all exceptions. (FWIW, I think a way to turn OFF hints would be helpful to many users as well.) 

If we did it right we might even provide a module that people can use to test for perl's exceptions in a way that will future proof them from changes in the messages themselves. Eg, we could provide a special test module or something that lets the code use symbolic names, and then we can change the text whenever we like. If the correct way to test if a string contains a given perl exception was 

```
check_exception(UNDECLARED,$string)
```

then we could free ourselves from CPAN breakage when we improve the error system.

Anyway, my point is, I have been thinking a lot about error messages, handling, and etc. I really want to improve it as I feel it is an obstacle to new perl devs, and I know some of the issue make working with Perl at scale more difficult than it should be (Perl actually works at scale pretty well, but error collection and processing are one place where there is a lot of friction.)  My feeling is that generally our compile errors are generally crappy, we could do a lot better, but there is a whole jenga pile of work to make it really feasible to do anything more than address the inconsistencies and make minor changes. So for now I think the consistency argument should beat the readability argument (which is a matter of taste) until we get there.  I do think we should go there tho.

tonycoz (approved):

demerphq:
Thanks @tonycoz, I will follow up with introducing a standard way to handler compiler error hints in some follow up patches.

Issue #18713, opened 2021-04-13
[doc] perldoc -f bless output is not very reader- or newbie-friendly


wchristian:
I don't know yet where exactly this is, mainly noting it so i don't forget.

After watching a newbie struggling with constructing a valid bless incantation i checked the pod for bless and found: The documentation for bless is currently two rather dense and short paragraphs and no synopsis.

For better reader-friendliness i think it would help to change the docs by:

  - adding a line break after the first sentence
  - adding a synopsis with a handful of common ways to use it

Unless there's heavy opposition or somebody else is faster than me, i'm happy to write this up later. :)

PR #20428, opened 2022-10-23
Clarify perldoc bless per #18713

joemcmahon:
- Specifically calls out the possible invocations
 - Breaks out the specific use cases and why to use them
 - Notes the things to avoid and why in clearer language
 - Update .mailmap so that tests all pass

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -931,24 +931,63 @@ X<bless>
 
 =for Pod::Functions create an object
 
-This function tells the thingy referenced by REF that it is now an object
-in the CLASSNAME package.  If CLASSNAME is an empty string, it is
-interpreted as referring to the C<main> package.
-If CLASSNAME is omitted, the current package
-is used.  Because a L<C<bless>|/bless REF,CLASSNAME> is often the last
-thing in a constructor, it returns the reference for convenience.
-Always use the two-argument version if a derived class might inherit the
-method doing the blessing.  See L<perlobj> for more about the blessing
-(and blessings) of objects.
+C<bless> tells Perl to mark the item referred to by C<REF> as an
+object in a package.  The two-argument version of C<bless> is
+always preferable unless there is a specific reason to I<not>
+use it.
 
-Consider always blessing objects in CLASSNAMEs that are mixed case.
-Namespaces with all lowercase names are considered reserved for
-Perl pragmas.  Builtin types have all uppercase names.  To prevent
-confusion, you may wish to avoid such package names as well.
-It is advised to avoid the class name C<0>, because much code erroneously
-uses the result of L<C<ref>|/ref EXPR> as a truth value.
+=over
 
-See L<perlmod/"Perl Modules">.
+=item * Bless the referred-to item into a specific package
+(recommended form):
+
+    bless $ref, $package;
+
+The two-argument form adds the object to the package specified
+as the second argument.
+
+=item * Bless the referred-to item into package C<main>:
+
+    bless $ref, "";
+
+If the second argument is an empty string, C<bless> adds the
+object to package C<main>.
+
+=item * Bless the referred-to item into the current package (not
+inheritable):
+
+    bless $ref;
+
+If C<bless> is used without its second argument, the object is
+created in the current package. The second argument should
+always be supplied if a derived class might inherit a method
+executing C<bless>. Because it is a potential source of bugs,
+one-argument C<bless> is discouraged.
+
+=back
+
+See L<perlobj> for more about the blessing (and blessings) of
+objects.
+
+L<C<bless>|/bless REF,CLASSNAME> returns its first argument, the
+supplied reference, as the value of the function; since C<bless>
+is commonly the last thing executed in constructors, this means
+that the the reference to the object is returned as the
+constructor's value and allows the caller to immediately use
+this returned object in method calls.
+
+C<CLASSNAME> should always be a mixed-case name, as
+all-uppercase and all-lowercase names are meant to be used only
+for Perl builtin types and pragmas, respectively. Avoid creating
+all-uppercase or all-lowercase package names to prevent
+confusion.
+
+Also avoid <Cbless>ing things into the class name C<0>; this
+will cause code which (erroneously) checks the result of
+L<C<ref>|/ref EXPR> to see if a reference is C<bless>ed to fail,
+as "0", a falsy value, is returned.
+
+See L<perlmod/"Perl Modules"> for more details.
 
 =item break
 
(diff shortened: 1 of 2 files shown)


pod/perlfunc.pod
```diff
@@ -931,24 +931,40 @@ X<bless>
...
-
-Consider always blessing objects in CLASSNAMEs that are mixed case.
-Namespaces with all lowercase names are considered reserved for
-Perl pragmas.  Builtin types have all uppercase names.  To prevent
-confusion, you may wish to avoid such package names as well.
-It is advised to avoid the class name C<0>, because much code erroneously
-uses the result of L<C<ref>|/ref EXPR> as a truth value.
-
-See L<perlmod/"Perl Modules">.
+C<bless> tells Perl to mark C<REF> as an object in a package.
```
Grinnz: This is not correct - the referent is marked as an object, not the reference.

pod/perlfunc.pod
```diff
@@ -931,24 +931,40 @@ X<bless>
...
+The two-argument version of C<bless> is always preferred unless there is a specific reason to I<not> use it.
+
+=over
+
+=item * Bless a reference into a specific package (recommended form):
```
Grinnz: as above.

pod/perlfunc.pod
```diff
@@ -931,24 +931,40 @@ X<bless>
...
+
+    bless $ref, $package;
+
+The two-argument form adds the object to the package specified as the second argument.
+
+=item * Bless a reference into package C<main>:
```
Grinnz: as above.

pod/perlfunc.pod
```diff
@@ -931,24 +931,40 @@ X<bless>
...
+
+    bless $ref, "";
+
+If the second argument is an empty string, C<bless> adds the object to package C<main>.
+
+=item * Bless a reference into the current package (not inheritable):
```
Grinnz: as above.

pod/perlfunc.pod
```diff
@@ -931,24 +931,40 @@ X<bless>
...
+
+See L<perlobj> for more about the blessing (and blessings) of objects.
+
+L<C<bless>|/bless REF,CLASSNAME> returns its first argument, the supplied reference, as the value of the function; since C<bless> is commonly the last thing executed in constructors, this returns the reference to the object by default and allows the caller to immediately use this returned object in method calls.
+
+C<CLASSNAME> should always be a mixed-case name, as all-uppercase and all-lowercase names are meant to be used only for Perl builtin types and pragmas, respectively. Avoid creating all-uppercase or all-lowercase package names to prevent confusion.
+
+Also avoid <Cbless>ing things into the class name C<0>; this will cause code which (erroneously) checks the result of L<C<ref>|/ref EXPR> to see if a reference is C<bless>ed to fail, as "0", a falsy value, is returned.
+
+See L<perlmod/"Perl Modules"> for more details.
```
Grinnz, line 990: this is going to fail podcheck as the lines need to be under 78 characters when formatted (they will end up needing to be shorter since this is in the middle of a list in perlfunc).

pod/perlfunc.pod
joemcmahon: Changed to to `the referred-to item`.

pod/perlfunc.pod
joemcmahon: Changed to to `the referred-to item`.

pod/perlfunc.pod
joemcmahon: Changed to to `the referred-to item`.

pod/perlfunc.pod
joemcmahon: Got it -- I'm changing all these to `the referred-to item`. Fixed here.

pod/perlfunc.pod
tonycoz: You marked this resolved, but there's still very long lines.

joemcmahon: I ran it through podcheck without issues, but I'll shorten everything to 64 characters.

joemcmahon: I see the issue; there are lines other than this specific one that are longer that that; checking every line.

khwilliamson:
On 10/23/22 21:47, Joe McMahon wrote:
> 
> 
> In pod/perlfunc.pod 
> <https://github.com/Perl/perl5/pull/20428#discussion_r1002862026>:
> 
>> +
> +    bless $ref;
> +
> +If C<bless> is used without its second argument, the object is created in the current package. The second argument should always be supplied if a derived class might inherit a method executing C<bless>. Because it is a potential source of bugs, one-argument C<bless> is discouraged.
> +
> +=back
> +
> +See L<perlobj> for more about the blessing (and blessings) of objects.
> +
> +L<C<bless>|/bless REF,CLASSNAME> returns its first argument, the supplied reference, as the value of the function; since C<bless> is commonly the last thing executed in constructors, this returns the reference to the object by default and allows the caller to immediately use this returned object in method calls.
> +
> +C<CLASSNAME> should always be a mixed-case name, as all-uppercase and all-lowercase names are meant to be used only for Perl builtin types and pragmas, respectively. Avoid creating all-uppercase or all-lowercase package names to prevent confusion.
> +
> +Also avoid <Cbless>ing things into the class name C<0>; this will cause code which (erroneously) checks the result of L<C<ref>|/ref EXPR> to see if a reference is C<bless>ed to fail, as "0", a falsy value, is returned.
> +
> +See L<perlmod/"Perl Modules"> for more details.
> 
> I ran it through podcheck without issues, but I'll shorten everything to 
> 64 characters.
> 

podcheck only cares about lines that are to be output verbatim.  Those 
begin with a blank space in the first column.

The length of all others is irrelevant; the pod formatters, since 
they're not verbatim, will typically wrap them at the displayed width.

pod/perlfunc.pod
joemcmahon: All lines trimmed to 64 characters, plus or minus a few. The really long lines are all definitely gone.

joemcmahon:
@khwilliamson Re wrapping, I was pretty sure that was the case, but there was no point arguing about it. In any case, themost recent version of the new text renders properly and is quite a bit clearer and more prescriptive of good practice.

demerphq:
I resolved the conflict in .mailmap and forced pushed this for you @joemcmahon. I think the new wording is good and I will merge if it passes CI.

demerphq (approved):

demerphq:
@Grinnz you have any follow up on this before i merge it?

demerphq:
And forced push again with the correct fix. :-)

Grinnz (approved):

demerphq:
Thanks, applied.

PR #20434, opened 2022-10-24
locale.c: Use buffer instead of SAVEFREEPV

khwilliamson:
This eliminates the use of temporaries here, as long suggested to do by the comments

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -2633,8 +2633,9 @@ S_wrap_wsetlocale(pTHX_ const int category, const char *locale)
     const char * result = Win_wstring_to_utf8_string(wresult);
     WSETLOCALE_UNLOCK;
 
-    SAVEFREEPV(result); /* is there something better we can do here? */
-
+    SAVEFREEPV(result); /* is there something better we can do here?  Answer:
+                           Without restructuring, returning a unique value each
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -2578,10 +2578,12 @@ S_wrap_wsetlocale(pTHX_ const int category, const char *locale)
...
 
-    const char * result = Win_wstring_to_utf8_string(wresult);
-    SAVEFREEPV(result); /* is there something better we can do here? */
-
-    return result;
+    const char * result = (const char *) Win_wstring_to_utf8_string(wresult);
+    const char * retval = save_to_buffer(result,
+                                         &PL_win32setlocale_buf,
+                                         &PL_win32setlocale_bufsize);
+    Safefree(result);
```
tonycoz: Wouldn't it be better to modify Win_wstring_to_byte_string() to accept the buf/bufsize pointers avoid the extra allocation/copy/free?

You'd need to extract the allocation part of save_to_buffer() out to a new function that Win_wstring_to_byte_string() would call.

khwilliamson: The code failed anyway.  This PR is now just a documentation change that things fail

tonycoz (approved):

PR #20435, opened 2022-10-24
Fix multiple typos in C files

atoomic:


charclass_invlists.h
```diff
@@ -434684,7 +434684,7 @@ static const char * const UNI_lb_values[] = {
 	"infixnumeric",
 	"Inseparable",
 	"inseparable",
-	"Inseperable",
+	"Inseparable",
```
tonycoz: a) this file is generated
b) the mis-spelling here is specified by [Unicode](http://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt).  Note the correct spelling is also available.

regexec.c
```diff
@@ -3583,7 +3583,7 @@ Perl_regexec_flags(pTHX_ REGEXP * const rx, char *stringarg, char *strend,
     regnode *c;
     char *startpos;
     SSize_t minlen;		/* must match at least this many chars */
-    SSize_t dontbother = 0;	/* how many characters not to try at end */
+    SSize_t don'tbother = 0;	/* how many characters not to try at end */
```
tonycoz: This is wrong, C doesn't allow `'` in variable names.

Further down too.

uconfig.h
```diff
@@ -3594,7 +3594,7 @@
 
 /* HAS_WCSXFRM:
  *	This symbol, if defined, indicates that the wcsxfrm routine is
- *	available to tranform a wide character string for wcscmp().
+ *	available to transform a wide character string for wcscmp().
```
tonycoz, line 3597: This change should be made in `config_h.SH`.  This change will be overridden on a make regen.

win32/perlhost.h
```diff
@@ -108,7 +108,7 @@ class CPerlHost
     };
 
 /* IPerlMemParse */
-    /* Assume something else is using locks to mangaging serialize
+    /* Assume something else is using locks to manage serialize
        on a batch basis
```
tonycoz: serialization?

charclass_invlists.h
atoomic: change reverted, thanks

regexec.c
atoomic: indeed... I should have seen it. Fixed

uconfig.h
atoomic: updated with other generated files

win32/perlhost.h
atoomic: applied the suggestion

tonycoz:
Please `make test_porting` before pushing again.

uconfig.h
Tux: I updated this in `metaconfig/U/perl/d_wcsxfrm.U` as 87e76cbf7e9ee95ecd1bfe81e6ddd31351ada25f
will be included after the next regen

jkeenan:
Superseded by https://github.com/Perl/perl5/pull/20646, which was merged into blead in 953398c3fd421376c66841d7658f8b97328e2726 (Dec 29 2022); closing.

PR #20439, opened 2022-10-25
allow porting/corelist.t to be run from root

demerphq:
Currently it errors out in the root dir.

### DIFF
--- a/t/porting/corelist.t
+++ b/t/porting/corelist.t
@@ -1,7 +1,9 @@
 #!perl -w
 
 # Check that the current version of perl exists in Module-CoreList data
-
+BEGIN {
+    push @INC, "." if -e "TestInit.pm";
+}
 use TestInit qw(T);
 use strict;
 use Config;


jkeenan (commented):
This certainly works.  The syntax is, however, slightly different from that in other cases where we want to do the same thing.  For example:
```
t/porting/utils.t
19-BEGIN {
20:    @INC = '..' if -f '../TestInit.pm';
21-}
```
Or more generally:
```
$ ack -C1 'TestInit\.pm' t
...
```
Is this a situation where we ought to follow one uniform syntax?

demerphq:
@jkeenan I think if you poke into it a touch further you will see that there are a small handful of patterns. a) scripts which were originally created to run from root, which were then modified to run correctly from 't', and b) scripts which were created to run from t c) scripts like b which were modified to run correctly from root, and d) scripts which originally ran in both but were broken by "." being removed from `@INC` back in the day. e) others I havent thought of...

Which pattern a given script follows seems to just depend on it authors. Eg, some scripts detect they are in root and cd into 't', some detect they are in 't' can cd into root, some replace chdir with `@INC` modification, etc.  

So for instance any script which has `require "test.pl"` in it expects to run from "t". Anything with `require './t/test.pl';` expects to be run from the root dir. 

If we went through and standarized all of that stuff then I suppose we could get to the point where this would all be a boilerplate recipe, but until then I suspect that there will be a handful of different fixes depending on the use case, and we shouldnt expect consistency. 

Lastly, I suspect some fixes are lazy. For instance `@INC=".."` is a bit different from `push @INC,".."`, in that it might be a bit more flexible with the user using -I on the command line when running the test script. Its not that annoying to have to do:

```
    ./perl -Ilib t/op/whatever.t
```

to make "whatever" work. (for instance maybe whatever.pl pushed "../lib" into @INC), but it is annoying when no matter how you run the script from the root dir it fails. :-)

Sorry to give such a complex answer to such a simple question.

OH, also, another point is that TestInit I think is intended to work around a bunch of these issues and it used to work when "." was in `@INC`, but doesn't when it is not.

demerphq:
@jkeenan  and to answer your question more specifically, I dont think it matters that we are consistent in this regard, although it would be better if we are, but i would argue that the `add to @INC` patterns are more flexible than the `overwrite @INC` patterns. 

Ideally we would have a standard boiler plate that ensures:

a) the script works properly with or without "." in `@INC`.
b) the script works properly when run from the t dir or from the root dir of the repo.

How reasonable it is to get there I dont know.

Issue #20438, opened 2022-10-25
cop_features not copied into optree


demerphq:
**Description**
Unlike hints we do not copy cop_features out of PL_compiling into the constructed nextstate ops. This means feature bit checks do not work at run time. Currently it seems we have no feature that perform run time decision making, but we have at least one queued up that does need to do so. https://github.com/Perl/perl5/pull/20365

It seems to be an oversight as the hints data is properly copied. I already have a patch for this, but not prepped for a PR yet. See  https://github.com/Perl/perl5/commit/6f97620ca1caf27246acb5faccdaa0ed3986e61f for example. Also @nwc10 identified the same issue and created https://github.com/nwc10/perl5/commit/d5f88b34531704a752a34ccf475e0ad62775f6c4 which i am testing as a cherry-pick to blead right now.

The following p5p irc is relevant:
```
(07:43:23) dmq: so i got to the bottom of the module_load problem.
(07:44:09) dmq: for some reason `use feature 'module_load';` does not set the features flag correctly but `use v5.37;` does.
(07:44:38) dmq: does that happen to ring a bell for anybody?
(08:19:34) dmq: is it possible that none of the existing features until now required *run* time decision making instead of compile time?
(08:19:52) dmq: i cant find a single COP that has a set cop_features field.
(08:20:38) dmq: except PL_compiling itself.
(08:32:06) Nicholas: Belive this is the case. It's a bug that hadn't been fixed IIRC. TonyC might remember more.
(08:34:19) Nicholas: There was this https://github.com/nwc10/perl5/commit/d5f88b34531704a752a34ccf475e0ad62775f6c4
(08:34:27) Nicholas: on that https://github.com/nwc10/perl5/commits/two-personalities
(08:34:42) Nicholas: but I can't remmeber if that commit is the same thing as what you seem to be asking
(08:35:16) Nicholas: and I really don't have time to dig further
(08:37:43) Nicholas: (you are not going mad. it is quite possibly what you think)
(08:45:02) dmq: I just found it!
(08:45:12) dmq: thanks for the confirmation nicholas.
(08:46:04) dmq: see 6f97620ca1caf27246acb5faccdaa0ed3986e61f
(08:46:21) dmq: https://github.com/Perl/perl5/commit/6f97620ca1caf27246acb5faccdaa0ed3986e61f
(08:47:03) dmq: i found one special case where cop_features was copied over for subs only.
(08:47:11) dmq: and only in a special edge case.
[...]

PR #20440, opened 2022-10-25
Default the COP's cop_features to the bitmask of the default-on featu…

demerphq:
…res.

The assumption is that cop_features holds the bitmask of the currently enabled feature set, and hence that S_magic_sethint_feature() can incrementally update it as keys for features are set in/cleared from %^H.

However, this overlooks one problem - the default set of features enabled (keys) is not empty, the default for %^H *is* empty, but as long as %^H is empty the internals knows that it is *not* running with a custom feature set, so avoids the complexity of setting up %^H.

There's a C test for features that knows about this, but it was only correctly handling (Perl) compile-time tests for features. That hasn't been noticed before, because all the "default on" features are compile-time only.

This commit fixes things so that run-time default-on features also work properly, by setting each COP's cop_features to the value of the bits set at the COP's compile time - ie the same approach as cop_hints.

### DIFF
--- a/cop.h
+++ b/cop.h
@@ -558,6 +558,8 @@ string C<p>, creating the package if necessary.
 #define CopHINTHASH_get(c)	((COPHH*)((c)->cop_hints_hash))
 #define CopHINTHASH_set(c,h)	((c)->cop_hints_hash = (h))
 
+#define CopFEATURES_setfrom(dst, src) ((dst)->cop_features = (src)->cop_features)
+
 /*
 =for apidoc   Am|SV *|cop_hints_fetch_pv |const COP *cop|const char *key              |U32 hash|U32 flags
 =for apidoc_item|SV *|cop_hints_fetch_pvn|const COP *cop|const char *key|STRLEN keylen|U32 hash|U32 flags
--- a/regen/feature.pl
+++ b/regen/feature.pl
@@ -305,8 +305,13 @@ for (@HintedBundles) {
 print $h <<'EOH';
 #define FEATURE_BUNDLE_CUSTOM	(HINT_FEATURE_MASK >> HINT_FEATURE_SHIFT)
 
-#define CURRENT_HINTS \
+/* this is preserved for testing and asserts */
+#define OLD_CURRENT_HINTS \
     (PL_curcop == &PL_compiling ? PL_hints : PL_curcop->cop_hints)
+/* this is the same thing, but simpler (no if) as PL_hints expands
+   to PL_compiling.cop_hints */
+#define CURRENT_HINTS \
+    PL_curcop->cop_hints
 #define CURRENT_FEATURE_BUNDLE \
     ((CURRENT_HINTS & HINT_FEATURE_MASK) >> HINT_FEATURE_SHIFT)
 
--- a/feature.h
+++ b/feature.h
@@ -43,8 +43,13 @@
 #define FEATURE_BUNDLE_537	7
 #define FEATURE_BUNDLE_CUSTOM	(HINT_FEATURE_MASK >> HINT_FEATURE_SHIFT)
 
-#define CURRENT_HINTS \
+/* this is preserved for testing and asserts */
+#define OLD_CURRENT_HINTS \
     (PL_curcop == &PL_compiling ? PL_hints : PL_curcop->cop_hints)
+/* this is the same thing, but simpler (no if) as PL_hints expands
+   to PL_compiling.cop_hints */
+#define CURRENT_HINTS \
+    PL_curcop->cop_hints
 #define CURRENT_FEATURE_BUNDLE \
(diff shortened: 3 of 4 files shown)


demerphq:
See also https://github.com/Perl/perl5/issues/20438

regen/feature.pl
```diff
@@ -302,6 +302,8 @@ sub longest {
     print $h "#define FEATURE_BUNDLE_$key	", $count++, "\n";
 }
 
+# I think that the test for (CURRENT_HINTS & HINT_LOCALIZE_HH) can actually
```
leonerd: "I" in a comment in committed code is rather unclear. Either put your name, or add a signature line to the comment to explain who you are who are thinking this.

leonerd (commented):
Overall the change looks good.

The change also ensures that generated COPs contain the `->cop_features` copied from the compiling state. That's a key useful change that we're after now, but the commit message doesn't really mention that part. It might be useful to mention that specifically as well so it's clearer we intend to add that too.

Or maybe split it into a separate commit to draw more attention to it.

regen/feature.pl
demerphq: Ill fix that.

demerphq: Fixed.

cop.h
```diff
@@ -558,6 +558,8 @@ string C<p>, creating the package if necessary.
 #define CopHINTHASH_get(c)	((COPHH*)((c)->cop_hints_hash))
 #define CopHINTHASH_set(c,h)	((c)->cop_hints_hash = (h))
 
+#define CopFEATURES_setfrom(c, o) ((c)->cop_features = (o)->cop_features)
```
leonerd: Tiniest of nitpics: isn't it traditional to use `dst` and `src` for these two arguments?

demerphq:
@leonerd: I split out the logic copying the cop_features data by integrating the patch from @tonycoz which uses a macro for the copy.

I also added a patch which simplifies the CURRENT_HINTS macro, which was doing an unnecessary ternary to avoid indirection on the PL_curcop pointer.

cop.h
demerphq: This seems in keeping with the style of the file, and is harmless, even though generally id agree with you that it would be more comprehensible.

demerphq: Fixed.

demerphq:
I pushed some fixes to update the version mentioned in @nwc10 patch since it is now a bit out of date.

tonycoz:
> The assumption is that cop_features holds the bitmask of the currently enabled feature set, and hence that S_magic_sethint_feature() can incrementally update it as keys for features are set in/cleared from %^H.
> 
> However, this overlooks one problem - the default set of features enabled (keys) is not empty, the default for %^H _is_ empty, but as long as %^H is empty the internals knows that it is _not_ running with a custom feature set, so avoids the complexity of setting up %^H.

"The assumption is that cop_features holds the bitmask of the currently enabled feature set" - cop_features is intended to reflect the current state of the feature* keys in `%^H`, this is only matches the current set of enabled features when the feature bundle (in cop_hints) is set to FEATURE_BUNDLE_CUSTOM.

I thought this change might break feature.pm setting features when switching from a feature bundle to a custom set, but feature.pm clears every %^H feature, even those not currently set (hence setting cop_features to zero) before setting the feature flags for that bundle.

"but as long as %^H is empty" - cop_hints, not $^H is looked at here.  %^H isn't only used for features, so this wouldn't work anyway.

```
use v5.36; # bundle set to FEATURE_BUNDLE_535, cop_features/%^H ignored for features
use feature 'indirect'; # bundle set to FEATURE_BUNDLE_CUSTOM, cop_features has meaning, feature.pm sets appropriate $^H{feature...} keys
use v5.34; # bundle set to FEATURE_BUNDLE_527, cop_features/%^H have features set to 535 bundle features + indirect, but are ignored
```

So I think the set-default-cop_features change has no real effect, and may confuse someone looking at the behaviour of the code.

One thing we could do here is eliminate feature bundles entirely, since they I assume they originally existed to avoid a costly %^H lookup.  We no longer pay that cost, and it would save the (currently always compile-time) cost of checking for the bundle, and get 4 hints bits back.  This has some complexities though.

The copying of the feature bits from PL_compiling to the new state op is still needed and unrelated to my comments above.

demerphq:
Sorry, some of the terms used here are confusing me. When you say "default bundle" you mean the patch from @nwc10? So you think we should just drop @nwc10's patch and only keep yours and mine? I am fine with that. It looked reasonable to me, but if you are convinced otherwise im fine with that. I had the impression from @nwc10 and the patch that if we apply it we can remove some other logic aftewards. But im not fussed. I just want the cop_features copy done properly.

demerphq:
Ive removed the default features patch.

tonycoz (approved):

PR #20441, opened 2022-10-25
release_managers_guide.pod - emphasize that the Module::CoreList needs updating after a version bump

demerphq:
People seem to overlook this step, and it seems to be out of order as well. This moves it up in the text and emphasizes it with an IMPORTANT tag.

### DIFF
--- a/Porting/release_managers_guide.pod
+++ b/Porting/release_managers_guide.pod
@@ -461,6 +461,15 @@ You may also need to regen opcodes:
 
  $ ./perl -Ilib regen/opcode.pl
 
+B<IMPORTANT>: When the version number is bumped, you must also update
+Module::CoreList (as described below in L<"update Module::CoreList">) to
+reflect the new version number. In particular if you do not run the
+
+    $ ./perl -Ilib Porting/corelist.pl cpan
+
+step or its equivalent (see the actual text in the section), then tests
+will fail.
+
 Test your changes:
 
  $ git clean -xdf   # careful if you don't have local files to keep!
@@ -483,9 +492,6 @@ At this point you may want to compare the commit with a previous bump to
 see if they look similar.  See commit f7cf42bb69 for an example of a
 previous version bump.
 
-When the version number is bumped, you should also update Module::CoreList
-(as described below in L<"update Module::CoreList">) to reflect the new
-version number.
 
 =head3 update INSTALL
 
@@ -1616,4 +1622,3 @@ L<http://www.xray.mpe.mpg.de/mailing-lists/perl5-porters/2009-05/msg00608.html>,
 plus a whole bunch of other sources, including private correspondence.
 
 =cut
-


haarg:
This seems to be trying to address how the porting tests were failing in blead recently, after the 5.37.5 release. But I don't think it does so properly.

The section being update here is the [Bump the version number](https://github.com/Perl/perl5/blob/blead/Porting/release_managers_guide.pod#bump-the-version-number) section, which begins with the test "Do not do this yet for a BLEAD-POINT release!". This section is regarding bumping the version prior to making a stable release. For blead releases, the version is bumped immediately after the release is made, not immediately before. This is shown in the [bump version](https://github.com/Perl/perl5/blob/blead/Porting/release_managers_guide.pod#bump-version) section.

For blead releases, according to the guide, Module::CoreList is not immediately updated with the new perl version. This is meant to allow time for a new release of Module::CoreList to be made with information about the new blead release. Adding an entry for the not-yet-released version would be inappropriate, because if a Module::CoreList release was made it would contain inaccurate information. Instead, Module::CoreList is meant to be updated "[the day after](https://github.com/Perl/perl5/blob/blead/Porting/release_managers_guide.pod#building-a-release---the-day-after)"

During the 5.37.5 release, as far as I can tell Module::CoreList was updated according to the guide. But the "day after" update was not made.

This does seem problematic though. Following the guide will leave the tests in blead failing for at least a day. It seems like either something should be changed in the release process for Module::CoreList, or something in the porting test.

demerphq:
So i guess I should close this PR then? Or leave it open for discussion on the latter issue?

Corion:
Personally, as a BLEAD-POINT releaser, I've always disliked this "skip this step" part of the guide, which makes me use `goto` (well, `gosub` ) instead of stepping through the list in a linear fashion.

Maybe we should write a tool to generate the RMG with only the appropriate sections, and insert the "bump Module::CoreList" parts at the appropriate place, instead of having the reader jump around?

As I usually do my releases on a weekend, I found that BINGOS is quite quick in releasing the updated Module::CoreList, but I admit that the "long" time between the Perl dev release and Module::CoreList is not great.

demerphq:
@corion good idea with the tooling. i was actually wondering how much of the release manager guide could be automated. it seems like a lot of steps to mess up on.

Corion:
I have [App-rmg-tools](https://github.com/Corion/App-rmg-tools) , which gives me a dashboard and (printed) next steps to do. Maybe later it will make sense to automate these steps, but there are a lot of steps that require human judgement.

The shell script `bleadperl-installtest.sh` automates some of the "follow-up" steps after building a release tarball, by "manually" installing DBI , DBD::SQLite and Inline.

demerphq:
@Corion it would be nice if we could figure out a way to automate away all the human judgment. Would you be interested in outlining what would be involved in an issue? (Hope you are well btw)

toddr:
Don't we already have a tool to make a specific guide for a release?

```pod
This script is run as:

 perl Porting/make-rmg-checklist \
     --version [5.x.y-RC#] > /tmp/rmg.pod
```

demerphq:
Closing unapplied.

PR #20442, opened 2022-10-25
Assorted manual whitespace cleanups and related changes

demerphq:
This is a set of manual adjustments to WS in our headers and in a few cases to our code which parses headers to harden them against whitespace change.  It consists primarily of WS changes to the headers, or minor adjustments to how we handle certain cases in our headers.

### DIFF
--- a/EXTERN.h
+++ b/EXTERN.h
@@ -9,10 +9,11 @@
  */
 
 /*
- * EXT  designates a global var which is defined in perl.h
- * dEXT designates a global var which is defined in another
- *      file, so we can't count on finding it in perl.h
- *      (this practice should be avoided).
+ * EXT:  designates a global var which is defined in perl.h
+ *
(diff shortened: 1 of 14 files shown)


demerphq:
Note this contains the manual changes from my other WS related PR, which is distinct from this. I wanted to get these in as they all improve legibility or functionality IMO.

jkeenan:
@demerphq, the merged conflicts which GH flagged last week in `intrpvar.h` appear to have been resolved, but there are 5 conflicts remaining in `regen/regcomp.pl`.  Could you take a look so that we can merge this p.r.?

Thank you very much.

demerphq:
@jkeenan done. I was able to drop two of the patches as they are no longer necessary due to the use of regen/HeaderParser.pm. Thanks for the heads up.

PR #20445, opened 2022-10-26
Add a diag_listed_as comment to perly.y

demerphq:
This way if the message is changed in the future we will notice in the diagnostics tests. 

This patch is two step as I have more modern bison than the one used for the last build. The first step is just the bison changes, the second is the bison changes with the new comment. 

Considering this is a fairly big version jump for bison this might bring other benefits or performance changes, presumably benefits.   (NOTE: after creating this i noticed someone had done a push with a more recent bison already, so the changes arent as big as they were before.)


leonerd (approved):
LGTM

PR #20449, opened 2022-10-27
Add `xV_FROM_REF` macros

leonerd:
These macros do the moral equivalent of `(AV *)SvRV(ref)` or similar for other types, while also `assert()`ing that it's correct to do so. Not much used in core but very handy in lots of XS modules and similar.

### DIFF
--- a/handy.h
+++ b/handy.h
@@ -108,6 +108,34 @@ blindly casts away const.
 #define MUTABLE_IO(p)	((IO *)MUTABLE_PTR(p))
 #define MUTABLE_SV(p)	((SV *)MUTABLE_PTR(p))
 
+/*
+=for apidoc_section $SV
+=for apidoc   Am |AV *|AV_FROM_REF|SV * ref
+=for apidoc_item |CV *|CV_FROM_REF|SV * ref
+=for apidoc_item |HV *|HV_FROM_REF|SV * ref
+
+The C<I<*>V_FROM_REF> macros extract the C<SvRV()> from a given reference SV
+and return a suitably-cast to pointer to the referenced SV. When running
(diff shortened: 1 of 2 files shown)


leonerd:
Actually, suddenly not sure I like this as it stands because it evaluates `ref` multiple times. I'll draft it for now until I can fix that.

leonerd:
OK; only works when `defined(PERL_USE_GCC_BRACE_GROUPS)` but overall I think it's still a win.

handy.h
```diff
@@ -108,6 +108,34 @@ blindly casts away const.
...
 #define MUTABLE_SV(p)	((SV *)MUTABLE_PTR(p))
 
+/*
+=for apidoc_section $SV
+=for apidoc   Am |AV *|AV_FROM_REF|SV * ref
+=for apidoc_item |CV *|CV_FROM_REF|SV * ref
+=for apidoc_item |HV *|HV_FROM_REF|SV * ref
+
+The C<I<*>V_FROM_REF> macros extract the C<SvRV()> from a given reference SV
+and return a suitably-cast to pointer to referred SV. When running under
```
khwilliamson: the referred to SV
or
the referenced SV

handy.h
```diff
@@ -108,6 +108,34 @@ blindly casts away const.
...
+C<-DDEBUGGING> assertions are also applied that check that I<ref> is
```
khwilliamson: I realize that the modern style is to drop a lot of commas, but my training was to separate clauses by commas.  The 'When...GGING>" is a clause that I would put a comma after

khwilliamson (approved):
A couple of nits in the pod.  But the actual code looks fine to me

leonerd:
Thanks; adjusted docs and force-pushed

PR #20450, opened 2022-10-28
Fix -Wsign-compare warning in `op.c`

t-a-k:
In 32-bit build (where IV is not wider than U32), comparing `PL_eval_begin_nest_depth` and `max_nest_iv` would generate a warning:

    op.c:10817:42: warning: comparison of integer expressions of different signedness: ‘U32’ {aka ‘long unsigned int’} and ‘IV’ {aka ‘long int’} [-Wsign-compare]
    10817 |             if (PL_eval_begin_nest_depth >= max_nest_iv) {
          |                                          ^~

This PR adds a cast to silence this warning.

### DIFF
--- a/op.c
+++ b/op.c
@@ -10818,7 +10818,10 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
                 sv_setiv(max_nest_sv, max_nest_iv);
             }
 
-            if (PL_eval_begin_nest_depth >= max_nest_iv) {
+            /* (UV) below is just to silence a compiler warning, and should be
+             * effectively a no-op, as max_nest_iv will never be negative here.
+             */
+            if (PL_eval_begin_nest_depth >= (UV)max_nest_iv) {
                 Perl_croak(aTHX_ "Too many nested BEGIN blocks, maximum of %" IVdf " allowed",
                              max_nest_iv);
             }


jkeenan:
On 10/28/22 10:15, TAKAI Kousuke wrote:
> In 32-bit build (where IV is not wider than U32), comparing 
> |PL_eval_begin_nest_depth| and |max_nest_iv| would generate a warning:
> 
> |op.c:10817:42: warning: comparison of integer expressions of different 
> signedness: ‘U32’ {aka ‘long unsigned int’} and ‘IV’ {aka ‘long int’} 
> [-Wsign-compare] 10817 | if (PL_eval_begin_nest_depth >= max_nest_iv) { 
> | ^~ |
> 
> This PR adds a cast to silence this warning.
> 

Can you provide the ./Configure arguments you are using when you see 
this warning?

t-a-k:
> Can you provide the ./Configure arguments you are using when you see this warning?

I didn't do anything special:
```
% ./Configure -de -DEBUGGING -Dusedevel
```
on 32-bit x86 Linux (Debian 11 with GCC 10.2.1).
Note that this warning will not be issued on 64-bit build (`-Duse64bitint` or on 64-bit targets).

op.c
```diff
@@ -10814,7 +10814,9 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
                 sv_setiv(max_nest_sv, max_nest_iv);
             }
 
-            if (PL_eval_begin_nest_depth >= max_nest_iv) {
+            /* (UV) below is just to silence a compiler warning, and should be
+             * effectively no-op as max_nest_iv will never be negative here */
```
khwilliamson: Change to
effectively a no-op,

khwilliamson (approved):
Otherwise LGTM

op.c
khwilliamson: s/effectively no-op/effectively a no-op,/

t-a-k: Thank you for your advise.  I've pushed a commit to fix it.

khwilliamson:
@t-a-k  Please squash the two commits, and we'll merge the result

t-a-k:
I'd squashed two commits, with rebasing to the current blead.

Issue #20419, opened 2022-10-20
Problems with exposing is_utf8()


sblondeel:
While migrating from an old system I had to add utf8::upgrade($line) after each

$line = <$fh>

type code snippet.

**Module: utf8**

**Description**

If a scalar $var is undef utf8::upgrade($var) should not turn it into q{}.

**Steps to Reproduce**

```perl

use Data::Dumper;

my $var;

print Dumper($var);

utf8::upgrade($var);

print Dumper($var);
```

***Output***

```
$ perl /tmp/utf8.pl 
$VAR1 = undef;
Use of uninitialized value in subroutine entry at /tmp/utf8.pl line 7.
$VAR1 = '';
```

***Expected output***

```
$ perl /tmp/utf8.pl 
$VAR1 = undef;
Use of uninitialized value in subroutine entry at /tmp/utf8.pl line 7.
$VAR1 = undef;
Use of uninitialized value in subroutine entry at /tmp/utf8.pl line 9.
```

**Expected behavior**

An undef value should still be undef after being upgraded to the utf8 internal
Perl representation of scalars.

undef is a special value of scalars, unrelated to any encoding scheme.

This breaks loops like

```perl
while (defined $line) {

  ...

  $line = <$fh>;
  utf8::upgrade($line);
}
```

Note: the utf8::upgrade call is in itself a circumvent of other Perl
bugs/module bugs I have reported.

**Perl configuration**
```
Summary of my perl5 (revision 5 version 28 subversion 1) configuration:
   
  Platform:
    osname=linux
    osvers=4.9.0
    archname=x86_64-linux-gnu-thread-multi
    uname='linux localhost 4.9.0 #1 smp debian 4.9.0 x86_64 gnulinux '
[...]

PR #20451, opened 2022-10-28
utf8::upgrade: Don't coerce undef arg

khwilliamson:
This fixes GH #20419

### DIFF
--- a/universal.c
+++ b/universal.c
@@ -593,11 +593,21 @@ XS(XS_utf8_upgrade)
         croak_xs_usage(cv, "sv");
     else {
         SV * const sv = ST(0);
-        STRLEN	RETVAL;
+        STRLEN	RETVAL = 0;
         dXSTARG;
 
-        RETVAL = sv_utf8_upgrade(sv);
-        XSprePUSH; PUSHi((IV)RETVAL);
+        XSprePUSH;
+        if (UNLIKELY(! sv)) {
+            XSRETURN_UNDEF;
+        }
+
+        SvGETMAGIC(sv);
+        if (UNLIKELY(! SvOK(sv))) {
+            XSRETURN_UNDEF;
+        }
+
+        RETVAL = sv_utf8_upgrade_nomg(sv);
+        PUSHi( (IV) RETVAL);
     }
     XSRETURN(1);
 }
--- a/lib/utf8.pm
+++ b/lib/utf8.pm
@@ -5,7 +5,7 @@ use warnings;
 
 our $hint_bits = 0x00800000;
 
-our $VERSION = '1.24';
+our $VERSION = '1.25';
 our $AUTOLOAD;
 
 sub import {
@@ -114,6 +114,8 @@ sequence in the native encoding (Latin-1 or EBCDIC) to UTF-8. The
 logical character sequence itself is unchanged.  If I<$string> is already
 upgraded, then this is a no-op. Returns the
 number of octets necessary to represent the string as UTF-8.
+Since Perl v5.38, if C<$string> is C<undef> no action is taken; prior to that,
+it would be converted to be defined and zero-length.
 
 If your code needs to be compatible with versions of perl without
 C<use feature 'unicode_strings';>, you can force Unicode semantics on
--- a/lib/utf8.t
+++ b/lib/utf8.t
@@ -669,6 +669,9 @@ for(__PACKAGE__) {
 	eval { utf8::upgrade($_) };
 	is $@, "", 'no error with utf8::upgrade on read-only COW';
 }
+
+is(utf8::upgrade(undef), undef, "Returns undef for undef input"); # GH #20419
+
 # This one croaks, but not because the scalar is read-only
 eval "package \x{100};\n" . <<'END'
     for(__PACKAGE__) {
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -45,6 +45,12 @@ XXX For a release on a stable branch, this section aspires to be:
 
 [ List each incompatible change as a =head2 entry ]
 
+=head2 L<C<utf8::upgrade()>|utf8/Utility functions>
+
+Starting in this release, if the input string is C<undef>, it remains
+C<undef>.  Previously it would be changed into a defined, zero-length
+string.
+
 =head1 Deprecations
 
 XXX Any deprecated features, syntax, modules etc. should be listed here.


jkeenan:
Looking at this p.r., I had the hunch that it would be the sort of change that would pass all its tests in the core test suite but stumble on something on CPAN.  So today I built a threaded `perl` using the `smoke-me/khw-upgrade` branch on FreeBSD-12 with my customary configuration options.  I then used `cpanm` to try to install 500 modules from an old 'CPAN River' list.  I got some failures that were not surprising because I've been getting them for years on this platform.  However, I got a peculiar test-time failure in Dancer.

But first, for a baseline, I installed Dancer against a very recent commit in blead.

**Blead**

FreeBSD-12 / threaded build / blead at fd7d660c03 (Oct 24 2022)

Dancer-1.3513 (latest CPAN release) installed successfully via `cpan`, albeit with a few test-time warnings.
```
[perlmonger: blead] $ ./bin/perl -v | head -2 | tail -1
This is perl 5, version 37, subversion 6 (v5.37.6 (v5.37.5-39-gfd7d660c03)) built for amd64-freebsd-thread-multi
[perlmonger: blead] $ ./bin/perl -Ilib -V:config_args
config_args='-des -Dusedevel -Uversiononly -Dprefix=/home/jkeenan/testing/blead -Dman1dir=none -Dman3dir=none -Duseithreads -Doptimize=-O2 -pipe -fstack-protector -fno-strict-aliasing';
[perlmonger: blead] $ ./bin/perl -Ilib -MDancer -E 'say $Dancer::VERSION;'
1.3513
```
Warnings observed during `make test` (probably not relevant to this ticket)
```
t/08_session/01_load.t .............................. ok
Use of uninitialized value in subroutine entry at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/FileUtils.pm line 32.
t/08_session/02_dependency_check.t .................. ok

t/12_response/03_charset.t .......................... ok
Wide character in print at /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/site_perl/5.37.6/HTTP/Server/Simple/PSGI.pm line 132, <DATA> line 16.
Wide character in print at /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/site_perl/5.37.6/HTTP/Server/Simple/PSGI.pm line 132, <DATA> line 16.
t/12_response/04_charset_server.t ................... ok
```

**smoke-me/khw-upgrade branch**

FreeBSD-12 / threaded build / blead at 6ea2a51564 (Oct 28 2022)
```
[perlmonger: khw-upgrade] $ ./bin/perl -v | head -2 | tail -1
This is perl 5, version 37, subversion 6 (v5.37.6 (v5.37.5-66-g6ea2a51564)) built for amd64-freebsd-thread-multi

[perlmonger: khw-upgrade] $ ./bin/perl -Ilib -V:config_args
config_args='-des -Dusedevel -Uversiononly -Dprefix=/home/jkeenan/testing/smoke-me/khw-upgrade -Dman1dir=none -Dman3dir=none -Duseithreads -Doptimize=-O2 -pipe -fstack-protector -fno-strict-aliasing';

$ ./bin/perl -Ilib -MDancer -c
Can't locate Dancer.pm in @INC (you may need to install the Dancer module) (@INC contains: lib /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/site_perl/5.37.6/amd64-freebsd-thread-multi /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/site_perl/5.37.6 /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/5.37.6/amd64-freebsd-thread-multi /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/5.37.6).
BEGIN failed--compilation aborted.
```
Below, `$THISPROVE` is the `bin/prove` installed with the `bin/perl` from the `smoke-me/khw-upgrade` branch.
```
[perlmonger: Dancer-1.3513] $ $THISPROVE -vb t/12_response/10_error_dumper_without_clone.t 
t/12_response/10_error_dumper_without_clone.t .. Devel::Hide hides Clone.pm
Can't locate Clone.pm in @INC (hidden)
BEGIN failed--compilation aborted at /home/jkeenan/testing/smoke-me/khw-upgrade/lib/perl5/site_perl/5.37.6/HTTP/Headers.pm line 8.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Response.pm line 14.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Response.pm line 14.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/SharedData.pm line 8.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/SharedData.pm line 8.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Request.pm line 13.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Request.pm line 13.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Route.pm line 13.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Route.pm line 13.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Route/Registry.pm line 8.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/Route/Registry.pm line 8.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/App.pm line 12.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer/App.pm line 12.
Compilation failed in require at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer.pm line 10.
BEGIN failed--compilation aborted at /usr/home/jkeenan/.cpanm/work/1666973947.78519/Dancer-1.3513/blib/lib/Dancer.pm line 10.
Compilation failed in require at t/12_response/10_error_dumper_without_clone.t line 17.
BEGIN failed--compilation aborted at t/12_response/10_error_dumper_without_clone.t line 17.
Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 

Test Summary Report
-------------------
t/12_response/10_error_dumper_without_clone.t (Wstat: 65280 (exited 255) Tests: 0 Failed: 0)
  Non-zero exit status: 255
Files=1, Tests=0,  1 wallclock secs ( 0.05 usr  0.00 sys +  0.16 cusr  0.04 csys =  0.25 CPU)
Result: FAIL
```
I got the same warnings in `make test` for Dancer in this branch as I did in blead.

jkeenan:
Per conversation with @haarg in irc.perl.org#p5p: This test failure is not due to the change @khwilliamson is proposing in this branch.  Since I was testing Dancer via `cpanm`, it was installing the latest version of Dancer's prerequisites, including HTTP::Headers version 6.44.  The "blead" I was using was several days old and already had HTTP::Headers version 6.43.  When I upgraded "blead" to 6.44, I got the same test failure in Dancer there that I did in Karl's branch.

universal.c
```diff
@@ -593,10 +593,13 @@ XS(XS_utf8_upgrade)
         croak_xs_usage(cv, "sv");
     else {
         SV * const sv = ST(0);
-        STRLEN	RETVAL;
+        STRLEN	RETVAL = 0;
         dXSTARG;
 
-        RETVAL = sv_utf8_upgrade(sv);
+        if (LIKELY(sv && SvANY(sv))) {
```
demerphq: I would have said that this should be `sv && SvOK(sv) && !SvROK(sv)`, but it might be that `SvANY(sv)` has the same result. I think i would prefer we check the flags. :-)

demerphq: I think i might make it return undef if it does nothing.

demerphq: Maybe it should just be `sv && SvOK(sv)`, eg, consider overloaded refs, but including `SvROK(sv)` helped me find what are arguably bugs in our test suite. (calling it on a qr// object).

khwilliamson: FWIW, not knowing any better, I  copied the template from pp_defined

demerphq: Expect it to look like this:

```
diff --git a/universal.c b/universal.c
index 0b12a231b5..931487364e 100644
--- a/universal.c
+++ b/universal.c
@@ -596,11 +596,14 @@ XS(XS_utf8_upgrade)
         STRLEN RETVAL = 0;
         dXSTARG;
 
-        if (LIKELY(sv && SvANY(sv))) {
+        XSprePUSH;
+        if (LIKELY(sv && SvOK(sv))) {
             RETVAL = sv_utf8_upgrade(sv);
+            PUSHi((IV)RETVAL);
+        } else {
+            PUSHs(newSV(0));
         }
 
-        XSprePUSH; PUSHi((IV)RETVAL);
     }
     XSRETURN(1);
 }
```

Eg, it should return undef if the argument is undef. I dont think testing SvANY() is correct, it should just test SvOK().

```
    "SvOK"
        Returns a U32 value indicating whether the value is defined. This is
        only meaningful for scalars.

            U32  SvOK(SV* sv)
```

The `SvANY`  in pp_defined is a bit surprising to me, maybe its faster? Weird. Why wouldnt SvOK() be implemented that way if its faster?  I know i have seen code that basically turns off the SvOK() flag to "undef it" temporarily.

Anyway, we definitely dont want to test `SvROK()` like i thought at first. We have a bunch of regex test code that expects to be able to upgrade the pattern. It just happens to work, but it means our tests double compile the pattern.

demerphq: Also see https://github.com/Perl/perl5/pull/20453 which I found by applying the above patch but also including `!SvROK(sv)` in the guard clause.

tonycoz: You need to SvGETMAGIC() before checking flags.

You can then use sv_utf8_upgrade_nomg() to avoid double get magic.

demerphq: BTW, I reviewed pp_defined again, and I think you have misread the code it contains. SvANY() as a check is *definitely* not sufficient to determine if an SV is undef (and I question its use in pp_defined at all actually). It checks to the see if the SV has something in the any_ptr slot. Any SV that does not cannot be defined, but an SV that does may still not be defined. If you review the logic further it falls through to a bunch of logic that is meant to handle the different types, which I think is why it does the SvANY check early so that those checks are "safe", and then it falls through to further logic which finally does the SvOK().

Since your code is dealing only with SV's (and not CV's or AV's or what not, as you can not pass such from perl to XS directly, they have to come in SV (RV) wrapper) at the XS level the correct way to see if an SV is defined is to simply call SvOK(), the SvANY check is not sufficient, and unnecessary if you have to call SvOK() anyway.

I have to say I find the code in defined() fairly odd, i would have expected a fairly minimal wrapper around SvOK().  But I am 100% certain it is not the code to copy for your use case.

khwilliamson:
I have tried to incorporate the comments

universal.c
```diff
@@ -593,11 +593,21 @@ XS(XS_utf8_upgrade)
...
-        RETVAL = sv_utf8_upgrade(sv);
-        XSprePUSH; PUSHi((IV)RETVAL);
+        XSprePUSH;
+        if (LIKELY(sv)) {
+            SvGETMAGIC(sv);
+            if (LIKELY(SvOK(sv))) {
+                RETVAL = sv_utf8_upgrade_nomg(sv);
+                PUSHi( (IV) RETVAL);
+            } else {
+                PUSHs(newSV(0));
```
tonycoz: This leaks, since the SV isn't made mortal (tested with `./perl -e 'my $x; while (1) { utf8::upgrade($x); }'`)

That isn't a problem for PUSHi() since it uses TARG.

You can simply use `XSRETURN_UNDEF;` here, and immediately below.

tonycoz:
This could probably use a test in `lib/utf8.t`.

khwilliamson:
Changed

tonycoz:
I don't see fixes for the leak nor tests, did you forget to push?

khwilliamson:
I had pushed to the wrong fork

tonycoz (approved):

demerphq (approved):

demerphq:
Does this need a perldelta entry?

khwilliamson:
Yes, and a change to its own pod.  Now done

demerphq:
On Fri, 11 Nov 2022, 22:17 Karl Williamson, ***@***.***>
wrote:

> Yes, and a change to its own pod. Now done
>

Did you forget to push? I don't see it.

Yves

>

PR #20453, opened 2022-10-28
re/fold_grind.pl - rework so we don't double compile patterns

demerphq:
We were calling utf8::upgrade() on qr// objects, which works, in that it stringifies the pattern, and then upgrades the string, and the string can be used to match against. But it double compiles the pattern, which is a bit unnecessary. For example

```
    $pat= qr/\x{DF}/;
    utf8::upgrade($pat); # $pat is now a string!
    "ss"=~/$pat/;        # and now $pat gets compiled again
```

works, but it does twice the work needed.

This was found by making utf8::upgrade() ignore refs.

### DIFF
--- a/t/re/fold_grind.pl
+++ b/t/re/fold_grind.pl
@@ -770,7 +770,8 @@ foreach my $test (sort { numerically } keys %{$tests_ref}) {
                                                   || ($charset eq 'd' && $utf8_pattern)
                                                   ||  $charset =~ /a/);
           my $upgrade_pattern = "";
-          $upgrade_pattern = ' utf8::upgrade($p);' if ! $pattern_above_latin1 && $utf8_pattern;
+          $upgrade_pattern = ' utf8::upgrade($rhs);'
+            if ! $pattern_above_latin1 && $utf8_pattern;
 
           my $lhs = join "", @x_target;
           my $lhs_str = eval qq{"$lhs"}; fail($@) if $@;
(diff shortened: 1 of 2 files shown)


khwilliamson:
There is significant logic in pattern matching that varies depending on whether the pattern is in UTF-8 versus not.  This code attempts to test all of it.   Thus, it is by design that any pattern that can be compiled without using UTF-8 is tested that way, and also tested when it is UTF-8.

demerphq:
@khwilliamson yes i know. I didnt change any of that. I just made it do it with one qr// compilation and not two.

the old code did this:

```
     my $qr = qr/ WHATEVER /;
     utf8::upgrade($qr);
     $str =~ $qr;
```

at the final step `$qr` is a string, (with the utf8 bit set), not a qr// object, as the utf8::upgrade() call will have stringified the qr// object, then turned on the utf8 flag.. Thus it has to be compiled a second time when it is matched against.

I changed the code to do something like this:

```
     my $empty_string = "";
     utf8::upgrade($empty_string);
     my $qr = qr/ WHATEVER $empty_string/;
     $str =~ $qr;
```

this should mean that the pattern is utf8 when it is compiled, and that it gets compiled once.  In some cases I didnt use the empty string but a non-empty string. Shouldn't matter.

Eg:

```
$ perl -le'my $empty_string= ""; utf8::upgrade($empty_string); print "ss"=~/$empty_string\xDF/i'
1
```

The point is the code shouldnt utf8::upgrade a compiled pattern, and then compile it again, it should construct the desired patter the first time around.

This patch reduces the run time of the fold_grind.t tests by about half a second, which is about %12.

khwilliamson (approved):
I had skimmed the PR, and jumped to the wrong conclusion.

LGTM

PR #20455, opened 2022-10-30
Win32API::File - Remove "Buffer Sizes" links

alanhaggai:
The POD in Win32API::File links to a non-existing "Buffer Sizes" section.


jkeenan:
> The POD in Win32API::File links to a non-existing "Buffer Sizes" section.

The [Win32API-File distribution](https://metacpan.org/pod/Win32API::File) is primarily maintained upstream on CPAN.  Hence, bug reports for its components should be filed with [that distribution's bug tracker](https://rt.cpan.org/Dist/Display.html?Name=Win32API-File).  You can use the GUI interface at that URL or send email to bug-Win32API-File [at] rt.cpan.org.

@chorny

PR #20457, opened 2022-10-30
cop.h - add support for refcounted cop_file under threads and cop_warnings always.

demerphq:
We have a weird bifurcation of the cop logic around threads. With threads we use a char * cop_file member, without it we use a GV * and replace cop_file with cop_filegv.

The GV * code refcounts filenames and more or less efficiently shares the filename amongst many opcodes. However under threads we were simplify copying the filenames into each opcode. This is because in theory opcodes created in one thread can be destroyed in another. I say in theory because as far as I know the core code does not actually do this. But we have tests that you can construct a perl, clone it, and then destroy the original, and have the copy work just fine, this means that opcodes constructed in the main thread will be destroyed in the cloned thread. This in turn means that you can't put SV derived structures into the op-tree under threads. Which is why we can not use the GV * stategy under threads.

As such this code adds a new struct/type RCPV, which is a refcounted string using shared memory. This is implemented in such a way that code that previously used a char * can continue to do so, as the refcounting data is located a specific offset before the char * pointer itself. This also allows the len data to embedded "into" the PV, which allows us to expose macros to acces the length of what is in theory a null terminated string.

    struct rcpv {
        UV      refcount;
        STRLEN  len;
        char    pv[1];
    };
    typedef struct rcpv RCPV;

The struct is sized appropriately on creation in rcpv_new() so that the pv member contains the full string plus a null byte. It then returns a pointer to the pv member of the struct. Thus the refcount and length and embedded at a predictable offset in front of the char *, which means we do not have to change any types for members using this.

We provide three operations: rcpv_new(), rcpv_copy() and rcpv_free(), which roughly correspond with newSVpv(), SvREFCNT_inc(), SvREFCNT_dec(), and a handful of macros as well. We also expose SAVERCPVFREE which is similar to SAVEGENERICSV but operates on pv's constructed with rcpv_new().

Currently I have not restricted use of this logic to threaded perls. We simply do not use it in unthreaded perls, but I see no reason we couldn't normalize the code to use this in both cases, except possibly that actually the GV case is more efficient.

Note that rcpv_new() does NOT use a hash table to dedup strings. Two calls to rcpv_new() with the same arguments will produce two distinct pointers with their own refcount data.

Refcounting the cop_file data was Tony Cook's idea.

**ADDENDUM:** It turned out to be very simple to convert cop_warnings to use the same rcpv infrastructure under the hood, so this PR also includes cop_warnings as well.


demerphq:
@tonycoz this is for you.

op.c
```diff
@@ -15242,6 +15242,119 @@ Perl_dup_warnings(pTHX_ STRLEN* warnings)
...
+        return NULL;
+
+    if (flags & RCPVf_USE_STRLEN)
+        len = strlen(pv);
+
+    rcpv = (RCPV *)PerlMemShared_malloc(sizeof(struct rcpv) + len + 1);
+    rcpv->len = len;
+    rcpv->refcount = 1;
+    if (!rcpv)
+        croak_no_mem();
```
tonycoz: The NULL check belongs before the accesses.

cop.h
```diff
@@ -506,14 +520,76 @@ string C<p>, creating the package if necessary.
...
+Returns the refcount for a pv created with C<rcpv_new()>. 
+
+=for apidoc Am|RCPV *|RCPV_LEN|char *pv
+Returns the length of a pv created with C<rcpv_new()>. 
+
+=cut
+*/
+
+struct rcpv {
+    UV      refcount;
```
tonycoz: `UV refcount` makes the reference count 64-bit on a 32-bit platform with `-Duse64bitint`, it could be STRLEN instead.

tonycoz:
Why not just put the RCPV pointer into cop_file? (probably renamed)

op.c
demerphq: Willfix.

cop.h
demerphq: willfix

demerphq:
> Why not just put the RCPV pointer into cop_file? (probably renamed)

The main reason is that it would require more code to change than I think is necessary or reasonable. For instance in scope.c we can use SAVEPPTR with a RCPV pv without needing to create a bespoke SAVEt_ for RCPV *. The scope.c logic and headers are all manually maintained, so adding new stuff is highly error prone and requires modifying a lot of comments and code. Ditto for the B code, doing the change like this meant I didn't need to change B at all. That seems like a good thing.

The other point is more general, I have a feeling this logic could long term be used in places where we store char * pointers and where it would be awkward to change their type. 

Also I want to look into using them for warning strings too, and i think there it will be awkward to change type. 

I am not /against/ using the type directly, but I feel this is a time where being parsimonious with the amount of change is a good thing.

demerphq:
BTW, i will rebase squash a bunch of these patches once review i complete.

demerphq:
@tonycoz it now includes the cop_warnings stuff. I might do more work on that soon, but it demonstrates how the pv API is helpful. It is not perfect, but it works.

cop.h
```diff
@@ -506,14 +520,77 @@ string C<p>, creating the package if necessary.
...
-#  define CopFILE_set(c,pv)	((c)->cop_file = savesharedpv(pv))
-#  define CopFILE_setn(c,pv,l)  ((c)->cop_file = savesharedpvn((pv),(l)))
+#  define CopFILE_set_x(c,pv)       ((c)->cop_file = rcpv_new((pv),0,RCPVf_USE_STRLEN))
+#  define CopFILE_setn_x(c,pv,l)    ((c)->cop_file = rcpv_new((pv),(l),0))
+#  define CopFILE_free_x(c)         ((c)->cop_file = rcpv_free((c)->cop_file))
+#  define CopFILE_copy_x(dst,src)   ((dst)->cop_file = rcpv_copy((src)->cop_file))
+
+#  define CopFILE_debug(c,t,rk)    if (0 && (c)->cop_file)                      \
+                                    PerlIO_printf(Perl_debug_log,               \
+                                        "%-14s THX:%p OP:%p PV:%p rc: %6d fn: '%.*s' at %s line %d\n", \
```
tonycoz: The %6d could be %6zu and avoid the cast to int for RCPV_REFCOUNT().

op.h
```diff
@@ -618,6 +618,7 @@ typedef enum {
  * The same mutex is used to protect the refcounts of the reg_trie_data
  * and reg_ac_data structures, which are shared between duplicated
  * regexes.
+ * The same mutext is used to protect the refcounts for RCPV objects.
```
tonycoz: mutext -> mutex

sv.c
```diff
@@ -15129,6 +15130,18 @@ Perl_ss_dup(pTHX_ PerlInterpreter *proto_perl, CLONE_PARAMS* param)
...
+             * we copy the SV, but dont change the **SV. But in this thread
+             * the target of the **SV could be something from the *other* thread.
+             * So how can this possibly work correctly? */
+            break;
+        case SAVEt_RCPV_FREE:
+            pv = (char *)POPPTR(ss,ix);
+            TOPPTR(nss,ix) = rcpv_copy(pv);
+            ptr = POPPTR(ss,ix);
+            (void)rcpv_copy(*((char **)ptr));
+            TOPPTR(nss,ix) = ptr;
```
tonycoz, line 15143: Should this OP_REFCNT_LOCK?

I see what you mean about the (probably wild) pointer issue, perhaps this should croak instead.

A real fix would require tracking every object across the clone, and keeping the object base pointer and the offset to the pointer we're working on in the save record.  That probably isn't practical.

tonycoz:
> I am not /against/ using the type directly, but I feel this is a time where being parsimonious with the amount of change is a good thing.

I'm mostly concerned with downstream code that tries to access cop_file/cop_warnings directly, though from looking today, there isn't much doing that.

Devel::Size will silently but mostly harmlessly not count the space needed for the size/refcount fields for cop_file, though it does appear to handle the value being shared across COPs correctly.

Syntax::Keyword::Try appears to access cop_warnings safely.

Please consider this concern resolved.

cop.h
demerphq: Ah, thanks, will fix.

sv.c
demerphq: > Should this OP_REFCNT_LOCK?

It locks in rcpv_copy when it does the refcount increment. Is that sufficient you think?

> I see what you mean about the (probably wild) pointer issue, perhaps this should croak instead.

I might try it just to see if anything blows up. I kinda wonder about a number of the cases in that code, it seems like several of them would have this problem.

tonycoz: > > Should this OP_REFCNT_LOCK?
> 
> It locks in rcpv_copy when it does the refcount increment. Is that sufficient you think?

Doh, of course.  Please ignore me on that then.

demerphq:
@nwc10 - regarding Devel::Size, is the repo handy anywhere so i can patch it to handle the new RCPVs?

tonycoz (approved):

demerphq:
@tonycoz thanks - i have squashed the fixups, and ill let CI run for a while before i merge.

PR #20458, opened 2022-10-30
Memoize: exclude CPAN-only toolchain from core

ap:
What it says on the tin.


jkeenan:
> What it says on the tin.

LGTM

PR #20459, opened 2022-10-30
gv.c - add ${^FORCE_UPGRADE} for triggering upgrade by concatenation

demerphq:
This is a readonly global variable which contains an upgraded empty string (eg UTF8-on). When concatenated into a string it (like any other upgraded string) causes the resulting string to also be upgraded. Similar to calling utf8::upgrade() on the string, but suitable for use in double quoted strings or regex patterns or other places where adding a utf8::upgrade() call might be awkward.

Assuming this gets merged I can think of some test code that could be cleaned up by using this, and I definitely would use it in one-liners and what not as well.

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -2211,6 +2211,16 @@ S_gv_magicalize(pTHX_ GV *gv, HV *stash, const char *name, STRLEN len,
                 if (memEQs(name, len, "\005NCODING"))
                     goto magicalize;
                 break;
+            case '\006':
+                if (memEQs(name, len, "\006ORCE_UPGRADE")) { /* ${^FORCE_UPGRADE} */
+                    /* empty string which is marked as UTF8 on */
+                    SV *sv= GvSVn(gv);
+                    sv_setpvs(sv,"");
+                    SvUTF8_on(sv);
+                    SvREADONLY_on(sv);
+                }
+                break;
+
             case '\007':	/* $^GLOBAL_PHASE */
                 if (memEQs(name, len, "\007LOBAL_PHASE"))
                     goto ro_magicalize;
--- a/lib/utf8.pm
+++ b/lib/utf8.pm
@@ -5,7 +5,7 @@ use warnings;
 
 our $hint_bits = 0x00800000;
 
-our $VERSION = '1.24';
+our $VERSION = '1.25';
 our $AUTOLOAD;
 
 sub import {
@@ -132,6 +132,13 @@ For example:
   utf8::upgrade($x);
   $x =~ /ss/i;       # matches
   my $z = uc($x);    # converts to "SS"
+
+As of 5.37.6 there is also the global variable C<${^FORCE_UPGRADE}>
+which contains an empty string which when concatenated with any other
+string causes the result to be an upgraded string. This can be helpful
+in some contexts where explicitly calling this function might be
+awkward. A similar effect can also be produced by using a C<\N{U+...}>
+style escape for one of the characters in the string.
 
 B<Note that this function does not handle arbitrary encodings>;
 use L<Encode> instead.
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -27,6 +27,11 @@ here, but most should go in the L</Performance Enhancements> section.
 
 [ List each enhancement as a =head2 entry ]
 
+=head2 New global variable C<${^FORCE_UPGRADE}>
+
+This variable contains the empty string, and can be used to force a string
+or pattern to be upgraded via concatenation.
+
 =head1 Security
 
 XXX Any security-related notices go here.  In particular, any security
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -504,6 +504,19 @@ mode is turned on.  See L<perlrun|perlrun/-a> for the B<-a> switch.  This
 array is package-specific, and must be declared or given a full package
 name if not in package main when running under C<strict 'vars'>.
 
+=item ${^FORCE_UPGRADE}
+
+This contains an empty string which when concatenated with any other
+string will cause the result to be an upgraded string. Eg:
+
+    print "ss"=~/\xDF/i ? "yes" : "no";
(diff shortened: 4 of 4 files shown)


tonycoz:
It seems like it would encourage the pre-5.12/unicode_strings model of Perl Unicode.

(and why do we upgrade when appending a zero-length string?)

Grinnz:
Anyone who sufficiently understands the Perl unicode model to correctly use this does not need it (thus it would only provide a footgun for people who shouldn't be using it), and Perl does not guarantee anything about how concatenation affects the internal state of a string.

Grinnz:
To clarify my comment; forcing upgrades on strings is only useful for working around bugs, and though such bugs are common even within Perl core, I don't feel it would be prudent to present additional features suggesting that it is a reliable and coherent operation to those unfamiliar with the Unicode string model, which I would conservatively estimate is 99% of Perl users. And regarding this specific operation, it would be perfectly reasonable and likely that Perl would choose differently in different versions how the internal storage of a string should be affected by appending an empty string of any sort.

Leont:
> It seems like it would encourage the pre-5.12/unicode_strings model of Perl Unicode.

Yeah, that.

demerphq:
> It seems like it would encourage the pre-5.12/unicode_strings model of Perl Unicode.

I don't really follow. This is just the same as using a `\N{U+...}` escape in a string, except it doesn't inject anything into the string. Any user could do the same thing by saying `my $UTF8= ""; utf8::upgrade($UTF8);` under every version of perl that supports Unicode.  So I dont get it. Why would it be a problem to have a var pre-available like this given that anyone can construct one anytime?  I imagine its use likely to be oneliners and maybe fresh_perl() tests, if that.

> (and why do we upgrade when appending a zero-length string?)

We upgrade whenever we append a UTF8-on string with a non-UT8-string. It doesn't matter that its empty. 

It would be strange it if it didn't work like this actually. Consider code like this:
```
my $str = "\x{100}";  
my $alpha= "\xDF" . $str;
chop($str);
chop($alpha);
my $bravo= "\xDF" . $str;
print "ss"=~/$alpha/i ? "yes" : "no";
print "ss"=~/$bravo/i ? "yes" : "no";
```
it would be pretty weird if those didn't output the same thing don't you think?

leonerd:
I have to say I am equally unsure why this would be a useful feature. This appears to be a baked-in variable that contains an empty string. It's hard to explain to people how this variable differs from the simple `""`.

The only real difference comes in terms of deep internals that most end-user programmers ought to not be touching - or really even have any awareness of. It's possible this value is useful in a few special-case situations like when you are writing unit tests that check your XS code correctly handles string values of both kinds of internal encoding, but aside from that special-purposes time I don't know when it would ever be useful to have.

demerphq:
@leonerd yes, I was thinking this would mostly be useful for testing, especially perl itself.  It is distinct from the normal empty string, its the unicode empty string.  Anyway, if people really think its a problem having this then we dont have to have it, but i would have rewritten a bunch of our tests to use it if we had it, and reduced the complexity of the generated test code while doing so. (Which makes understanding and debugging what you broke easier.) Part of the problem is that many of these tests are fresh_perl type, so being brief is helpful, and at the same time "putting it in a module" or whatnot isnt really an attractive option. I mean, personally who cares if we have a variable like this? Its in reserved namespace, it is helpful to perl core devs (at least me to start, and i bet others over time), and it is as useful as utf8::upgrade() is, and we expose that.

tonycoz:
> > It seems like it would encourage the pre-5.12/unicode_strings model of Perl Unicode.
> 
> I don't really follow. This is just the same as using a `\N{U+...}` escape in a string, except it doesn't inject anything into the string. Any user could do the same thing by saying `my $UTF8= ""; utf8::upgrade($UTF8);` under every version of perl that supports Unicode. So I dont get it. Why would it be a problem to have a var pre-available like this given that anyone can construct one anytime? I imagine its use likely to be oneliners and maybe fresh_perl() tests, if that.
> 
> > (and why do we upgrade when appending a zero-length string?)
> 
> We upgrade whenever we append a UTF8-on string with a non-UT8-string. It doesn't matter that its empty.
> 
> It would be strange it if it didn't work like this actually. Consider code like this:
> 
> ```
> my $str = "\x{100}";  
> my $alpha= "\xDF" . $str;
> chop($str);
> chop($alpha);
> my $bravo= "\xDF" . $str;
> print "ss"=~/$alpha/i ? "yes" : "no";
> print "ss"=~/$bravo/i ? "yes" : "no";
> ```
> 
> it would be pretty weird if those didn't output the same thing don't you think?


They do return the same thing under the model we (try to) educate people to use:
```
tony@venus:.../git/perl5$ cat ../20459.pl
use v5.12;
my $str = "\x{100}";
my $alpha= "\xDF" . $str;
#chop($str);
$str = "";  # no longer UTF8 on
chop($alpha);
my $bravo= "\xDF" . $str;
print "ss"=~/$alpha/i ? "yes" : "no";
print "ss"=~/$bravo/i ? "yes" : "no";
tony@venus:.../git/perl5$ ./perl -Ilib -l ../20459.pl 
yes
yes
```

I'm not really suggesting we break compatibility with non-unicode_strings code by changing the behaviour of appending an upgraded empty string, but appending this value only has an effect on non-unicode_strings code, a model we tell our users to avoid.

If tests are made easier with this type of value, they can add `utf8::upgrade(my $upgrade = "");` at the top and use that value.

Or write a trivial function that returns the upgraded string.

iabyn:
On Mon, Oct 31, 2022 at 05:18:39PM -0700, Tony Cook wrote:
> If tests are made easier with this type of value, they can add `utf8::upgrade(my $upgrade = "");` at the top and use that value.
> 
> Or write a trivial function that returns the upgraded string.

Or just substr("\x{100}",0,0)

-- 
everyone's satisfaction.

demerphq:
I like this idea, but others dont. closing and deleting.

Issue #20384, opened 2022-10-11
BBC: Blead breaks XS-Parse-Keyword, Devel-Chitin, Devel-WatchVars, others


andk:
**Description**

Starting with v5.37.4-92-g35458d36d3  tests started to fail for PEVANS/XS-Parse-Keyword-0.25.tar.gz

Sample fail reports:

- http://www.cpantesters.org/cpan/report/a7af5df2-493e-11ed-a119-5c71d9fbdcec # Solaris
- http://www.cpantesters.org/cpan/report/eb2e4352-4968-11ed-b9a4-86c3d88efda4 # Linux

Git bisect for this fail lead me to

```  text
35458d36d3c9fcf47064e4379e15276057b0bca5 is the first bad commit
commit 35458d36d3c9fcf47064e4379e15276057b0bca5
Author: Felipe Gasper <felipe@felipegasper.com>
Date:   Mon Sep 12 08:58:00 2022 -0400

    Compile anonymous subs as anoncode without srefgen.

```

**Steps to Reproduce**

cpan  -i  PEVANS/XS-Parse-Keyword-0.25.tar.gz

**Expected behavior**

Should build, test and install the pkg

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 37 subversion 5) configuration:
  Commit id: 35458d36d3c9fcf47064e4379e15276057b0bca5
  Platform:
    osname=linux
    osvers=5.19.0-2-amd64
    archname=x86_64-linux
    uname='linux k93msid 5.19.0-2-amd64 #1 smp preempt_dynamic debian 5.19.11-1 (2022-09-24) x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.4-92-g35458d36d3/da1c -Dmyhostname=k93msid -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Uuseithreads -Uuselongdouble -DEBUGGING=-g'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
[...]

PR #20462, opened 2022-10-31
Fail attempts to pass anonsub to \$ prototype.

FGasper:
Issue #20384

### DIFF
--- a/t/comp/proto.t
+++ b/t/comp/proto.t
@@ -18,7 +18,7 @@ BEGIN {
 # strict
 use strict;
 
-print "1..215\n";
+print "1..216\n";
 
 my $i = 1;
 
@@ -892,3 +892,24 @@ $_ = sub ($$$$$$$) {};
 @_ = (1, 2, 3, prototype(), 4, 5, 6);
 print "not " unless "@_" eq '1 2 3 $$$$$$$ 4 5 6';
 print "ok ", $i++, " - [perl #123514] (got @_)\n";
+
+{
+    my $weird_failure = q<>;
+
+    if (eval 'sub scalarref (\$) { }; scalarref( sub {} ); 1') {
+        print "not ok $i: Unexpected scalarref success!\n";
+    }
+    else {
+        if ($@ !~ m/anonymous subroutine/ || $@ !~ m/scalarref/) {
+            $weird_failure = $@;
+
+            $weird_failure =~ s/^/# /m;
+
+            print "# Unexpected error (or none):$/$weird_failure";
+        }
+
+        print( ($weird_failure ? 'not ok' : 'ok') . " $i - anonsub passed to \\\$\n" );
+    }
+
+    $i++;
+}
--- a/op.c
+++ b/op.c
@@ -2124,6 +2124,7 @@ Perl_scalarvoid(pTHX_ OP *arg)
         case OP_REF:
         case OP_REFGEN:
         case OP_SREFGEN:
+        case OP_ANONCODE:
         case OP_DEFINED:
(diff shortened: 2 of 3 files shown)


t/comp/proto.t
```diff
@@ -892,3 +892,16 @@ $_ = sub ($$$$$$$) {};
...
 print "ok ", $i++, " - [perl #123514] (got @_)\n";
+
+{
+    $i++;
+    my $err = q<>;
+
+    eval 'sub scalarref (\$) { }; scalarref( sub {} )';
+    if ($@ !~ m/anonymous subroutine/ || $@ !~ m/scalarref/) {
+        $err = $@;
+        warn;
```
demerphq: personally id feel better if this was `print "$_\n" for split /\n/, $@;` or `diag` if its available.

FWIW, if I had written this it would look more like this (untested):
```
my $unexpected_err = "";
my $ok = eval 'sub scalarref (\$) { }; scalarref( sub {} ); 1';
my $err = !$ok ? $@ : "";
if ($err !~ m/anonymous subroutine/ || $err !~ m/scalarref/) {
   $unexpected_err = $err;
}
print( (($ok || $unexpected_err) ? 'not' : '') . "ok $i - anonsub passed to \\\$\n",
   map { "#$_\n" } split /\n/, $unexpected_err );
```
I dont mean to style nit, its probably fine as is, but it makes me leery using $@ in any way other than a plain copy. It also makes me leery not checking that eval returns true, and outputing the error on STDERR when the test output is on STDOUT means that it might be emitted out of order which makes life difficult. Using a single print statement with commented output means it should be output all together in one chunk in the correct order and not upset the harness.  The other changes are just to make intent clear here. When I first read this it was obvious what '$err' "meant" in the code. 

None of this is absolutely necessary to change I suspect. But it would make me feel better if you integrated some of the points I raise here.

FGasper: > It also makes me leery not checking that eval returns true

Does checking `$@` not substitute?

FGasper: Tweaked, @demerphq. Thanks!

demerphq: Historically you couldn't rely on `$@` being true when an eval fails. The only reliable way to tell that you should look at `$@` is if the eval returned a true value.  Maybe this has been fixed enough my concern is irrelevant, but at the same time, using the robust formulation is obviously correct, whereas not using it open up that tiny wedge of doubt. So i just always use the robust formulation, then i don't have to worry. :-)

t/comp/proto.t
```diff
@@ -892,3 +892,20 @@ $_ = sub ($$$$$$$) {};
...
+    $i++;
+    my $failure = q<>;
+
+    if (eval 'sub scalarref (\$) { }; scalarref( sub {} ); 1') {
+        print "not ok $i: Unexpected scalarref success!\n";
+    }
+    else {
+        if ($@ !~ m/anonymous subroutine/ || $@ !~ m/scalarref/) {
+            $failure = $@;
+            print "Unexpected error (or none): $failure";
```
demerphq: this can confuse the harness. Stick a # at the start, if $failure can be multiline stick a # in front of each.

FGasper: Fixed.

demerphq (approved):

FGasper:
@demerphq I added a 2nd commit to this branch that fixes a related breakage, also reported in the referent issue (#20384). Do you mind looking at it?

Thank you!

atoomic (approved):

leonerd (approved):
LGTM

PR #20463, opened 2022-10-31
Prefer scalar assignment to get caller's first return value

richardleach:
Multiple forms of syntax can be used to obtain a package name from `caller`, which emits this as its first return value, and assign that name to a lexical scalar.

The following each achieve the same result, but with varying efficiency:
* `sub callme { my $package = caller(2); ...}`
* `sub callme { my ($package) = caller(2); ...}`
* `sub callme { my $package = (caller(2))[0]; ...}`

In the first example, `pp_caller` determines only the package name and pushes it to the stack. In the other two examples, the other 10 of `caller`'s return values are calculated and pushed onto the stack, before being discarded.

This commit changes non-CPAN-first instances of the latter two forms in core to the first form.

### DIFF
--- a/dist/Env/lib/Env.pm
+++ b/dist/Env/lib/Env.pm
@@ -1,6 +1,6 @@
 package Env;
 
-our $VERSION = '1.05';
+our $VERSION = '1.06';
 
 =head1 NAME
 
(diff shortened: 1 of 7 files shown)


richardleach:
Rebased

demerphq (approved):

Issue #20461, opened 2022-10-31
Leaking locales under address sanitizer


demerphq:
**Description**
When building with the address sanitizer I can see leaking locale objects. This is preventing me from debugging leaks that I caused in https://github.com/Perl/perl5/pull/20457 which is how i noticed, see https://github.com/Perl/perl5/actions/runs/3361810659/jobs/5572900746

**Steps to Reproduce**
./Configure -Dusedevel -Doptimize=-g -d -DDEBUGGING -Dusethreads -Accflags=-DPURIFY -Dcc='gcc -fsanitize=address' -Dld='gcc -fsanitize=address' -Dprefix="$HOME/perl-blead"
make

Build blows up.

**Expected behavior**
Perl should build in this configuration in blead.

**Perl configuration**

```
$ ./Configure -Dusedevel -Doptimize=-g -d -DDEBUGGING -Dusethreads -Accflags=-DPURIFY -Dcc='gcc -fsanitize=address' -Dld='gcc -fsanitize=address' -Dprefix="$HOME/perl-blead"
$ make
# Build fails...
$ ./miniperl -Ilib -V
Can't locate Config.pm in @INC (you may need to install the Config module) (@INC contains: lib .).
BEGIN failed--compilation aborted.

=================================================================
==1706939==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 348 byte(s) in 1 object(s) allocated from:
    #0 0x7ff5edcc4808 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x7ff5ed83e1b9 in __newlocale /build/glibc-SzIz7B/glibc-2.31/locale/newlocale.c:200
    #2 0x55c87a1077ef in S_emulate_setlocale_i /home/yorton/git_tree/perl/locale.c:1251
    #3 0x55c87a10b230 in Perl_set_numeric_standard /home/yorton/git_tree/perl/locale.c:1970
    #4 0x55c87a10b039 in S_new_numeric /home/yorton/git_tree/perl/locale.c:1946
    #5 0x55c87a1109a9 in S_new_LC_ALL /home/yorton/git_tree/perl/locale.c:2447
    #6 0x55c87a119a3a in Perl_init_i18nl10n /home/yorton/git_tree/perl/locale.c:5162
    #7 0x55c8798ec2d6 in perl_construct /home/yorton/git_tree/perl/perl.c:466
[...]

PR #20464, opened 2022-10-31
locale.c - add comments about PERL_DESTRUCT_LEVEL avoiding "leaks"

demerphq:
Some of the locale data is deliberately not destroyed during shutdown. Setting PERL_DESTRUCT_LEVEL=2 causes them to be cleaned up. If someone ends up debugging under the address sanitizer and ends up at this logic then the comments will help them avoid some frustration.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -1247,7 +1247,12 @@ S_emulate_setlocale_i(pTHX_
                                    line, basis_obj));
         }
 
-        /* Ready to create a new locale by modification of the exising one */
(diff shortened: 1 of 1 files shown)


demerphq:
See https://github.com/Perl/perl5/issues/20461

richardleach:
LGTM, although why not have the two comment blocks be more identical? (The text in `S_new_numeric` seems more definitive.)

demerphq:
@richardleach - done.

khwilliamson (approved):

PR #20466, opened 2022-11-01
Devel-PPPort: fix STMT_START and STMT_END to not warn on clang

haarg:
Since 7169efc77525df70484a824bff4ceebd1fafc760, perl's core STMT_START and STMT_END macros no longer try to use brace groups, due to the warnings they can generate and their very limited usefulness.

That commit also changed Devel::PPPort to remove the use of brace groups in STMT_START/STMT_END. That led to errors in older perls, so it was partly reverted in e08ee3cb66f362c4901846a46014cfdfcd60326c. Since then, various other macros have been improved to properly work with brace groups enabled or disabled.

We can now remove the brace group using variant of STMT_START/STMT_END, which will silence the warnings from clang.

### DIFF
--- a/dist/Devel-PPPort/parts/inc/misc
+++ b/dist/Devel-PPPort/parts/inc/misc
@@ -299,17 +299,12 @@ __UNDEF_NOT_PROVIDED__  PERL_GCC_BRACE_GROUPS_FORBIDDEN
 
 #undef STMT_START
 #undef STMT_END
-#if defined(VOIDFLAGS) && defined(PERL_USE_GCC_BRACE_GROUPS)
-#  define STMT_START    (void)( /* gcc supports ``({ STATEMENTS; })'' */
-#  define STMT_END      )
-#else
-#  if defined(VOIDFLAGS) && (VOIDFLAGS) && (defined(sun) || defined(__sun__)) && !defined(__GNUC__)
+#if defined(VOIDFLAGS) && (VOIDFLAGS) && (defined(sun) || defined(__sun__)) && !defined(__GNUC__)
 #    define STMT_START  if (1)
 #    define STMT_END    else (void)0
-#  else
+#else
 #    define STMT_START  do
 #    define STMT_END    while (0)
-#  endif
 #endif
 
 __UNDEFINED__  boolSV(b)    ((b) ? &PL_sv_yes : &PL_sv_no)
--- a/dist/Devel-PPPort/PPPort_pm.PL
+++ b/dist/Devel-PPPort/PPPort_pm.PL
@@ -756,7 +756,7 @@ package Devel::PPPort;
 use strict;
 use vars qw($VERSION $data);
 
-$VERSION = '3.68';
+$VERSION = '3.69';
 
 sub _init_data
 {
--- a/dist/Devel-PPPort/parts/inc/mess
+++ b/dist/Devel-PPPort/parts/inc/mess
@@ -58,17 +58,18 @@ NEED_vmess
 #  else
 #    define D_PPP_FIX_UTF8_ERRSV_FOR_SV(sv) STMT_START {} STMT_END
 #  endif
-#  define croak_sv(sv)                         \
-    STMT_START {                               \
-        SV *_sv = (sv);                        \
-        if (SvROK(_sv)) {                      \
-            sv_setsv(ERRSV, _sv);              \
-            croak(NULL);                       \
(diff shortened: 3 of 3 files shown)


demerphq:
this is failing CI, but it looks ok to me, not sure what is up.

demerphq (approved):

dist/Devel-PPPort/parts/inc/misc
```diff
@@ -299,18 +299,8 @@ __UNDEF_NOT_PROVIDED__  PERL_GCC_BRACE_GROUPS_FORBIDDEN
...
-#  if defined(VOIDFLAGS) && (VOIDFLAGS) && (defined(sun) || defined(__sun__)) && !defined(__GNUC__)
-#    define STMT_START  if (1)
-#    define STMT_END    else (void)0
-#  else
-#    define STMT_START  do
-#    define STMT_END    while (0)
-#  endif
-#endif
+#define STMT_START do
+#define STMT_END while (0)
```
atoomic: Wasn't there any good reasons to prefer the other flavor when possible?

khwilliamson: I thought there were multiple implementations depending on the compiler (besides the brace groups).  Perhaps those went away with C99?

haarg: I've restored the other variant of this. I'm not certain of the exact necessity of it, but it seems reasonable to keep.

haarg:
After going over the history of this, I noted that this had previously been changed and then reverted. Since then, there have been more changes to make the brace groups less needed.

I've restored the alternate non-brace group implementation of `STMT_START`/`STMT_END`, and updated the commit message to reflect the history of this.

dist/Devel-PPPort/parts/inc/call
```diff
@@ -54,7 +54,7 @@ __UNDEFINED__ PERL_LOADMOD_IMPORT_OPS   0x4
     (   (cond)                                      \
      && (_errsv = ERRSV)                            \
      && (SvROK(_errsv) || SvTRUE(_errsv))           \
-     && (croak_sv(_errsv), 1));                     \
+     && (({ croak_sv(_errsv); }), 1));              \
```
demerphq: This seems weird to me, wouldnt it be simpler to just define a croak_sv_ret() that returns 1? It would never actually do that as it would longjump, but wouldn't it be a simpler way to avoid the need for the list or the brace groups? It strikes me as something that is likely useful in other marcos.

haarg: In core, `croak_sv` is a function. Devel::PPPort provides a version of it, but it's just a macro using `STMT_START`/`STMT_END`. That feels wrong. Maybe it could be changed like so:
```
PERL_STATIC_INLINE void D_PPP_CROAK_SV(pTHX_ SV *sv) {
    SV *_sv = (sv);
    if (SvROK(_sv)) {
        sv_setsv(ERRSV, _sv);
        croak(NULL);
    } else {
        D_PPP_FIX_UTF8_ERRSV_FOR_SV(_sv);
        croak("%" SVf, SVfARG(_sv));
    }
}
#  define croak_sv(sv) D_PPP_CROAK_SV(sv)
```

demerphq: My point is that I think the (croak_sv(),1) is there because Perl_croak_sv() does not return a value. It is not a true function, but rather a procedure with a void return. So you cant use it in an expression by itself like the one we are looking at. If we create a Perl_croak_sv_ret() which returns 0, or 1, or whatever (it doesnt matter it will never return). Then the compile will be happy and there would be no need for (croak_sv(),1).

Eg, if a function is void then i think 

```
int x = void_function();
```

is an error but

```
int x = (void_function(),1);
```
is not.

I think that the function you propose would have the same problem, even if it works around the issue with PPPort providing a macro form of it.

demerphq: Frankly it seems weird to me at a certain level that the croak() functions dont return something. It make them usable in expressions without hijinks, and the return can be casted away where we dont need it. I guess that boat has long sailed now tho.

haarg:
I changed this to define `croak_sv` as a macro to a static inline function. Checking over the history, Devel::PPPort has been changed to provide several other functions in the same way, so it seems it should be safe.

This allows removing the brace groups from STMT_START/STMT_END without any other problems.

PR #20467, opened 2022-11-01
Add a `any_padix` member to the ANY union

leonerd:
Needed the `PADOFFSET` typedef to be moved first.

### DIFF
--- a/pad.h
+++ b/pad.h
@@ -13,9 +13,6 @@
 
 /* offsets within a pad */
 
-typedef SSize_t PADOFFSET; /* signed so that -1 is a valid value */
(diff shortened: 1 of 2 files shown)


tonycoz:
Do we need this?

We already have `any_ssize` which has the same type, if we hide  it behind a macro we can use the same entry.

Though right now we don't use `any_size`, `any_ssize`, `any_strlen` (even for SAVESTRLEN()).

demerphq:
@leonerd if this isnt needed as @tonycoz suggests can we close it please?

jkeenan:
@leonerd, it does not appear that there is support for applying this p.r.  Unless you intend to return to development on it, could you close it?  Thanks.

Issue #20468, opened 2022-11-01
1102a6f006 breaks IO-Socket-SSL tests, prevents CPANtester reports from blead


jkeenan:
Today I attempted to install Task::CPAN::Reporter against blead.  The installation failed due to failures in two test programs in the IO-Socket-SSL test suite.
```
$ uname -mrs
Linux 5.10.0-18-amd64 x86_64

$ bleadperl -v | head -2 | tail -1
This is perl 5, version 37, subversion 6 (v5.37.6 (v5.37.5-90-g6848aea2c8)) built for x86_64-linux

$ bleadperl -V:config_args
config_args='-des -Dusedevel -Uversiononly -Dprefix=/home/jkeenan/testing/blead -Dman1dir=none -Dman3dir=none';

$ cd ~/testing/blead
$ ./bin/cpan IO::Socket::SSL

# [failures in two tests]

$ bleadperl -Iblib/lib t/public_suffix_lib_encode_idn.t
Segmentation fault
$ bleadperl -Iblib/lib t/public_suffix_lib_libidn.t
Segmentation fault
```
Bisecting with this invocation:
```
perl Porting/bisect.pl --start=v5.37.5 --module=IO::Socket::SSL
```
... pointed to recent commit 1102a6f006f2bb626d2c7f1c5b7d360c28518129
```
1102a6f006f2bb626d2c7f1c5b7d360c28518129 is the first bad commit
commit 1102a6f006f2bb626d2c7f1c5b7d360c28518129
Author: chromatic <chromatic@wgz.org>
Date:   Thu Oct 27 11:24:58 2022 +0200
Commit:     Yves Orton <demerphq@gmail.com>
CommitDate: Tue Nov 1 09:46:16 2022 +0100


    Enable `use feature 'module_true'`
```
@chromatic, @demerphq, can you take a look?

demerphq:
Ill pick this up, i shepherded it into core.

jkeenan:
On a threaded build on FreeBSD-12 just now I got failures in **3** test programs, not just the 2 in Linux:

```
$ uname -mrs
FreeBSD 12.3-RELEASE-p1 amd64

$ ./bin/perl -v | head -2 | tail -1
This is perl 5, version 37, subversion 6 (v5.37.6 (v5.37.5-93-g5f112c0ac6)) built for amd64-freebsd-thread-multi

$ ./bin/perl -Ilib -V:config_args
[...]

PR #20470, opened 2022-11-01
pp_ctl.c - teach module_true how to deal with blocked requires

demerphq:
If require encounters an @INC hook that blocks the require it ends up with a STUB node as the optree for the LEAVEEVAL. This was causing the module_true logic to segfault. Guarding against the OP_STUB node fixes the problem.

This affected the tests for IO::Socket::SSL. Thanks to Graham Knopp for the reduced case, and James Keenan for reporting it in https://github.com/Perl/perl5/issues/20468.

This should fix the problem.

### DIFF
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4867,24 +4867,26 @@ PP(pp_leaveeval)
              * feature state out of the COP data it contains.
              */
             if (check) {
-                const OP *kid = cLISTOPx(check)->op_first;
-                const OP *last_state = NULL;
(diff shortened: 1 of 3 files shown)


jkeenan:
With this branch I can install both IO-Socket-SSL and Module-Pluggable in the Linux and FreeBSD environments mentioned in https://github.com/Perl/perl5/issues/20468.  (Someone other than me should review the C-code.)  Thanks.

haarg (approved):

demerphq:
fixed up the require tests hopefully

PR #20471, opened 2022-11-01
See if we can reduce our CI jobs in half by not running on push and pull_request

demerphq:
This is a test, trying to stop the tests running twice for PR's against branches hosted in the repo.  The patch makes it so we only run CI on pull_request, or on pushes to blead or maint.

### DIFF
--- a/.github/workflows/irc-notifications.yaml
+++ b/.github/workflows/irc-notifications.yaml
@@ -1,5 +1,5 @@
 name: "Push Notification"
-on: [push, pull_request]
+on: [pull_request]
 # add create for tracking tags
 
 # IRC colors: https://modern.ircdocs.horse/formatting.html
--- a/.github/workflows/irc-notifications.yaml
+++ b/.github/workflows/irc-notifications.yaml
@@ -1,5 +1,5 @@
 name: "Push Notification"
-on: [pull_request]
+on: [push, pull_request]
 # add create for tracking tags
 
 # IRC colors: https://modern.ircdocs.horse/formatting.html
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -52,11 +52,11 @@ name: testsuite
 on:
   push:
     branches:
-      - "blead"
-      - "maint-*"
+      - "**"
     tags-ignore:
       - "*"
   pull_request:
+      - "*:*"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/irc-notifications.yaml
```diff
@@ -1,5 +1,5 @@
 name: "Push Notification"
-on: [push, pull_request]
+on: [pull_request]
```
atoomic: maybe `push` should only do a very simple test suite run?

demerphq: I think if we do it on pull_request we dont need it on push. but i can restore that.

tonycoz:
What does the unlabeled `- "*:*"` match on?

From looking at the [workflow documentation](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow) there can be branches, types, paths subkeys but I don't see anything on a pattern directly under "pull_requests".

The testsuite workflow didn't run for the second commit here, I don't know if the rule change indicates it shouldn't have run or if the workflow failed to parse.

demerphq:
@tonycoz IIRC i had a discussion with @bram and he suggested that my original patch works for PR's in the repo, but not PR's out of the repo. I believe the idea was that that pattern would trigger CI for PR's from another repo, but not from our own. 

@bram do you recall?

jkeenan:
> @tonycoz IIRC i had a discussion with @bram and he suggested that my original patch works for PR's in the repo, but not PR's out of the repo. I believe the idea was that that pattern would trigger CI for PR's from another repo, but not from our own.
> 
> @bram do you recall?

Discussion in this Draft pull request lapsed in November.  If we want to proceed, how should we do so?

demerphq:
On Tue, 17 Jan 2023, 22:10 James E Keenan, ***@***.***> wrote:

> @tonycoz <https://github.com/tonycoz> IIRC i had a discussion with @bram
> <https://github.com/bram> and he suggested that my original patch works
> for PR's in the repo, but not PR's out of the repo. I believe the idea was
> that that pattern would trigger CI for PR's from another repo, but not from
> our own.
>
> @bram <https://github.com/bram> do you recall?
>
> Discussion in this Draft pull request lapsed in November. If we want to
> proceed, how should we do so?
>

I think it would be great if we could improve this. Just not sure how.

Yves

>

PR #20472, opened 2022-11-01
FindBin.pm - Handle FIFOs / Named Pipes

DabeDotCom:
Before:  (Dies)

     bash% perl -MFindBin <( echo 'print "$FindBin::Bin\n"' )
     Cannot find current script '/dev/fd/63' at FindBin.pm line 167.
     BEGIN failed--compilation aborted at FindBin.pm line 167.

After:   (Behaves like '-e' and '-')

     bash% perl -MFindBin <( echo 'print "$FindBin::Bin\n"' )
     /path/to/current/directory

### DIFF
--- a/dist/FindBin/lib/FindBin.pm
+++ b/dist/FindBin/lib/FindBin.pm
@@ -90,7 +90,7 @@ our @EXPORT_OK = qw($Bin $Script $RealBin $RealScript $Dir $RealDir);
 our %EXPORT_TAGS = (ALL => [qw($Bin $Script $RealBin $RealScript $Dir $RealDir)]);
 our @ISA = qw(Exporter);
 
-our $VERSION = "1.53";
+our $VERSION = "1.54";
 
 # needed for VMS-specific filename translation
 if( $^O eq 'VMS' ) {
--- a/dist/FindBin/lib/FindBin.pm
+++ b/dist/FindBin/lib/FindBin.pm
@@ -111,9 +111,9 @@ sub init
  *Dir = \$Bin;
  *RealDir = \$RealBin;
 
- if($0 eq '-e' || $0 eq '-')
+ if($0 eq '-e' || $0 eq '-' || -p $0)
   {
-   # perl invoked with -e or script is on C<STDIN>
+   # perl invoked with -e or script is on C<STDIN> or a fifo (named pipe)
    $Script = $RealScript = $0;
    $Bin    = $RealBin    = cwd2();
    $Bin = VMS::Filespec::unixify($Bin) if $^O eq 'VMS';


dist/FindBin/t/from-fifo.t
```diff
@@ -0,0 +1,44 @@
...
+###  There may be a better way to `POSIX::mkfifo()` then use `runperl`  ###
+###  from `../../t/test.pl`, but all that's not available when the      ###
+###  `dist/*` tests are being run...  I just stole these helper subs!   ###
+###########################################################################
+
+sub _print   { local($\, $", $,) = (undef, ' ', ''); print STDOUT @_; }
+sub skip_all { if (@_) { _print "1..0 # Skip @_\n" } else { _print "1..0\n" }; exit(0); }
+
+skip_all('/bin/bash required') unless -f "/bin/bash";
+
```
jkeenan: Whether we want FindBin to handle FIFOs and named pipes is something that really ought to be discussed **first** on the Perl5-Porters mailing list.

But even if we do want to add this functionality, having some of the regression tests depend on using `bash` as our shell **will not fly.**  Perl builds on many operating systems, not all of which support `bash` and we use quite a few more shells than just `bash`.  So this p.r. is not acceptable in its current form.

I recommend you make your case for this new functionality by posting to the perl5-porters mailing list.

Tux: protest! Linux is moving towards `/usr/bin` as default location and not `/bin`. Though most Linux distributions *still* support `/bin`, that test should be at least testing for `/usr/bin` (in addition to `/bin`)
This is a relative fresh rockylinux8:
```
rocky8: $ locate bin/bash
/usr/bin/bash
/usr/bin/bashbug
/usr/bin/bashbug-64
```
Hence I agree with @jkeenan

demerphq: > Whether we want FindBin to handle FIFOs and named pipes is something that really ought to be discussed first on the Perl5-Porters mailing list.

FWIW, @jkeenan I don't agree. The code currently errors without this check, and work with it.  This doesn't strike me as anything more than a bug fix. If this breaks a platform, or changes behavior in an undesirable way then you might have a point, but nobody has suggest that it would. 

I mean, go ahead and start a thread if you wish, I'm not saying you mustn't or shouldn't, I am merely saying this doesn't seem like something that needs more than a few commiters saying "yep doesn't seem like it can break anything".  (Note it is not clear we are at that threshold yet, but that is a separate point.) It is not a language feature change or whatnot, its simply the modernizing of a very very old piece of code.  Its the kind of thing that if the context were different (better tests, and you hadn't already replied) I might have merged without thinking too much about. It definitely doesn't trigger a "woah this could have a huge impact and so lets discuss it widely before we do it" in my head. More like a "maybe it could cause issues on an esoteric system, but sufficiently useful that we should find out".

I could totally see a patch like this (with better tests) being ptiched, and accepted in a few hours of chatter on the irc channel.

> But even if we do want to add this functionality, having some of the regression tests depend on using bash as our shell will not fly.

Totally agree.  The tests need to work in a way that is shell and platform agnostic. This code is used widely. We need to make sure the new test doesn't do something weird on some platform, like block on access or some other weird behavior that nobody is expecting.

demerphq:
Hi @DabeDotCom can you add a test for this new functionality please? Id like to unstall this PR.

khwilliamson:
@DabeDotCom we'd like to move forward with this PR, but need a test

PR #20474, opened 2022-11-02
improve storable hook error reporting and allow hooks to work with blessed regexps

### LINKED ISSUE #19964: Unexpected object type (8) in store_hook()
**Where**
Storable

**Description**
When running Storable::freeze on some structure, I got the following error, which is completely unhelpful and undocumented:
"Unexpected object type (8) in store_hook()"

I think it is necessary, if Storable throws off exceptions like this, that they be properly documented so that we know what corrective action to take.

tonycoz:
Fixes #19964 

This does not attempt to fix the feature request to add support for globs in hooks as suggested in #19964.

### DIFF
--- a/dist/Storable/t/blessed.t
+++ b/dist/Storable/t/blessed.t
@@ -44,7 +44,7 @@ use Storable qw(freeze thaw store retrieve fd_retrieve);
    'long VSTRING' => \(my $lvstring = eval "v" . 0 x 300),
    LVALUE         => \(my $substr  = substr((my $str = "foo"), 0, 3)));
 
-my $test = 14;
+my $test = 18;
 my $tests = $test + 41 + (2 * 6 * keys %::immortals) + (3 * keys %::weird_refs);
 plan(tests => $tests);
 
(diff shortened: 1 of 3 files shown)


dist/Storable/t/blessed.t
```diff
@@ -414,3 +414,54 @@ is(ref $t, 'STRESS_THE_STACK');
...
+    {
+        package RegexpHooked;
+        sub STORABLE_freeze {
+            ++$hook_called;
+            "$_[0]";
+        }
+        sub STORABLE_thaw {
+            my ($obj, $cloning, $serialized) = @_;
+            ++$hook_called;
+            $$obj = ${ qr/$serialized/ };
```
demerphq, line 455: Wow. I am shocked that works. IN C sure, but in perl, im pretty surprised.  I guess it is safe because storable isnt backported?

demerphq (approved):

dist/Storable/t/blessed.t
tonycoz: Backported?  I wrote that a while ago, and from the condition it worked in 5.12, and it tested ok in 5.24 in CI.

There's code in sv_setsv() to handle REGEXP assignment, at least in blead.

It looks like deliberate support was added in f0826785082.

demerphq: Oh wait. It doesn't actually work properly. `$obj` is not blessed. but `qr/.../` is. I guess that is expected?

demerphq: Thanks for the pointer. Interesting what people do with functions you create. (I created reg_temp_copy) I didnt expect this use case at all. :-) Makes perfect sense tho.

Issue #20391, opened 2022-10-12
gmake test: Windows 11, gcc 12: Use of uninitialized value $_ in substitution (s///) at ..\pod\buildtoc line 215, <$fh> chunk 2


hakonhagland:
I am trying to build blead on Windows 11 using MinGW-w64 and gcc 12.2 from https://winlibs.com/ (using the MSVCRT runtime library) following a similar approach as in https://github.com/Perl/perl5/issues/20096. See also https://github.com/Perl/perl5/issues/20383.

After downloading WinLibs to `C:\Winlibs64-Gcc12-msvcrt`, I needed to rename gnu make binary to `gmake.exe`:
```
> copy C:\Winlibs64-Gcc12-msvcrt\mingw64\bin\mingw32-make.exe C:\Winlibs64-Gcc12-msvcrt\mingw64\bin\gmake.exe
```
This is gcc version 12.2.0, and gnu make version 4.3:
```
c:\PerlWinLib> git clone https://github.com/Perl/perl5.git
[...]

c:\PerlWinLib>cd perl5\win32

c:\PerlWinLib\perl5\win32>where gcc
C:\Winlibs64-Gcc12-msvcrt\mingw64\bin\gcc.exe

c:\PerlWinLib\perl5\win32>gcc --version
gcc (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

C:\PerlWinLib\perl5\win32>where gmake
C:\Winlibs64-Gcc12-msvcrt\mingw64\bin\gmake.exe

c:\PerlWinLib\perl5\win32>gmake --version
GNU Make 4.3
Built for x86_64-w64-mingw32
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
To build perl I needed to set the environment variable `CCHOME`:
```
c:\PerlWinLib\perl5\win32>set CCHOME=C:\Winlibs64-Gcc12-msvcrt\mingw64
```
Now, I was able to build perl:
```
[...]

PR #20475, opened 2022-11-02
buildtoc: Add error check

khwilliamson:
If something goes wrong, this wouldn't have known it


jkeenan:
> If something goes wrong, this wouldn't have known it

Would it be possible for you to provide an example of the error which this p.r. is intended to correct?

khwilliamson:
It's clearly wrong to read a file in a loop and not check for EOF.

But I discovered this in tracking down errors like this in #20391

> Use of uninitialized value $_ in substitution (s///) at ..\pod\buildtoc line 219, <$fh> chunk 2.

which is repeated over and over

PR #20480, opened 2022-11-03
<> to readline documentation updates

scottchiefbaker:
I believe using `readline()` is considered best practice now, so this is a simple doc fix to update from `<>` to `readline()`

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -4584,7 +4584,7 @@ Reading from a file:
         or die "Can't open < input.txt: $!";
 
     # Process every line in input.txt
-    while (my $line = <$fh>) {
+    while (my $line = readline($fh)) {
         #
         # ... do something interesting with $line here ...
         #


scottchiefbaker:
If `<>` to `readline()` doc fixes are OK and recommended I'll submit a couple more as well.

pod/perlfunc.pod
```diff
@@ -4584,7 +4584,7 @@ Reading from a file:
         or die "Can't open < input.txt: $!";
 
     # Process every line in input.txt
-    while (my $line = <$fh>) {
+    while (my $line = readline($fh)) {
```
demerphq, line 4587: This should be `while (defined(my $line = readline($fh))) {`  BUt otherwise seems reasonable unless this is documenting `<>` itself.

demerphq (changes requested):

Grinnz:
I would hesitate to call it "Best practice". personally I always use readline() for explicitness but `<>` is heavily used in the wild and I don't see any reason at this point to suggest against it, except to point out that readline() is the same operation and less ambiguous, as `<>` can also be parsed as glob() depending on its contents.

pod/perlfunc.pod
Grinnz: it actually doesn't matter; the implicit defined check is applied to the optree regardless of whether it was parsed from `readline` or `<>`.

demerphq: Oh, i sit corrected:
```
$ perl -MO=Deparse -e'while (my $line = <>) { ... }'
while (defined(my $line = readline ARGV)) {
    do {
        die 'Unimplemented'
    };
}
-e syntax OK
```

thanks.

demerphq (approved):

scottchiefbaker:
From a novice perspective I always teach `readline($fh)` over `<$fh>` so I want to clean up the docs to reflect this. Readline is preferred, but `<>` isn't going anywhere. They're aliases of each other essentially.

demerphq:
FWIW, at the internals level <> is an alias for readline(). Its not quite correct to say "of each other". The run time internals don't know about <> at all, it is implemented via code generation during the parse phase.

demerphq:
I'll wait for a second approve as @Grinnz has a point.  I personally lean towards teaching the long form first, so i would apply this, but it is not the case that readline is preferred over <>, the latter is very heavily used.

leonerd (approved):
LGTM

The two Windows smokers got upset about something in t/op/fork.t and I can't imagine this one-line docs change has anything to do with it. I expect something entirely unrelated.

scottchiefbaker:
@leonerd there are a couple other locations in the code that use `<$filehandle>` instead of `readline()` that I'd like to update. It's been my understanding that `readline()` is preferred in 2022 for readability. Is that correct? Would more doc updates to reflect `readline()` be accepted?

PR #20481, opened 2022-11-03
Revert patches to try to get t/op/fork.t to pass on win32. 

demerphq:
This reverts commit f8552c1a7e2b27e9c88f12e4569bbdf7218d0f27.

This is to test if it is the cause of op/fork.t breakage on Win32.


demerphq:
Served its purpose. Closing without merge.

PR #20483, opened 2022-11-04
pod/perlhacks.pod - document TEST_ARGS and make win32 perl use it too

demerphq:
Win32 has had TEST_SWITCHES and TEST_FILES for ages. Unix has TEST_ARGS. This extends Win32 to support TEST_ARGS as well. I havent made the contrary change to the unix code, anyone using TEST_SWITCHES is already using it on Win32.

### DIFF
--- a/pod/perlhack.pod
+++ b/pod/perlhack.pod
@@ -880,12 +880,28 @@ it doesn't redirect stderr to stdout.
 Note that under Win32 F<t/harness> is always used instead of F<t/TEST>,
 so there is no special "test_harness" target.
 
+Under the Unix build process you may use the TEST_ARGS and TEST_FILES
+parameters to pass arguments through to the underlying harness call.
+This means that for instance you could do
+
+    make test_harness TEST_ARGS="-v -re pat"
+
+which would make, and then run the test harness in verbose mode over
+files which contain "pat". Or you could do
+
+    make test_harness TEST_ARGS="-torture" TEST_FILES="op/*.t"
+
+and run torture tests on files matching the glob "op/*.t".
+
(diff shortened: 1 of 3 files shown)


demerphq:
@tonycoz you are the main win32 dev these days i think, could you check this please?

pod/perlhack.pod
```diff
@@ -880,13 +880,25 @@ it doesn't redirect stderr to stdout.
...
 
+Under the Unix build process you may use the TEST_ARGS parameter to
+pass arguments through to the underlying harness call. This means
+that for instance you could do
+
+    make test_harness TEST_ARGS="-v -re pat"
+
+which would make, and then run the test harness in verbose mode over
+files which contain "pat".
+
```
tonycoz: Unix also supports TEST_FILES.

win32/Makefile
```diff
@@ -1325,12 +1325,12 @@ test-prep : all utils ../pod/perltoc.pod
 
 test : test-prep
 	cd ..\t
```
tonycoz, line 1327: `win32/GNUmakefile` also needs to be updated (`win32/Makefile` is used by `nmake` part of MSVC, `win32/GNUmakefile` is used by GNU make)

Otherwise it looks fine.

demerphq:
@tonycoz thanks for the feedback, ill pick it up today.

pod/perlhack.pod
demerphq: thanks,add a note about that.

win32/Makefile
demerphq: Added this too.

tonycoz (approved):

PR #20487, opened 2022-11-05
buildtoc: Fix to work on native line endings 

khwilliamson:
\r is clearer and potentially more portable than \015

### DIFF
--- a/pod/buildtoc
+++ b/pod/buildtoc
@@ -186,7 +186,7 @@ my ($inhead1, $inhead2, $initem);
 sub podset {
     my ($pod, $file) = @_;
 
-    open my $fh, '<:raw', $file or my_die "Can't open file '$file' for $pod: $!";
+    open my $fh, '<:', $file or my_die "Can't open file '$file' for $pod: $!";
 
     local *_;
     my $found_pod;
--- a/pod/buildtoc
+++ b/pod/buildtoc
@@ -206,7 +206,7 @@ sub podset {
     local $/ = '';
 
     while(<$fh>) {
-	tr/\015//d;
+	s/\r$//;
 	if (s/^=head1 (NAME)\s*/=head2 /) {
 	    unhead1();
 	    $OUT .= "\n\n=head2 ";


Leont:
I don't understand why we need to do this, but if we really need to do this I think the better solution is to just open the files with `:crlf` as mode.

khwilliamson:
I added text to the commit message giving more details as to the failure reasons.

I don't understand why `:crlf` is better.  I was unaware of plain : until I started digging for this.  But what we want is a layer that will take whatever the native line ending is and convert it to the form that perl understands.  I would prefer a layer named `:native`, as being clearer.  But`crlf` seems to me to be too wedded to a particular implementation.  I think it's better to have a layer whose name doesn't imply a particular underlying encoding

Grinnz:
`:crlf` is in the default layers on Windows (which plain `:` would apply regardless of open modes currently in scope). There is no other case where "native line endings" should need translation, now that classic MacOS is unsupported.

Grinnz:
The main complication is that the line ending format is decided by the file, not the OS. So the default layers would not help in the case where you open Windows-formatted files on a Linux system. In that case you can either append the `:crlf` layer yourself or remove the extraneous `\r` bytes manually. I don't know enough about the use case of this script to know whether that's a relevant concern.

demerphq:
I think @Grinnz has it correct in his last comment. This code is designed to read files that are encoded with either unix or windows line endings and to do the the right thing with both. This looks pretty reasonable to me. Using :crlf would be right if all the input files were in windows format.

Personally i like the idea of input discipline that just strips whatever line ending there are, be they crlf or unix regardless of the platform the code runs on.

demerphq:
@khwilliamson personally i think the intent here would be clearer if you restored :raw and then used `s/\R\z//; # strip linebreaks whatever they are`. I wont educate you on \R, as I think you added our support for it. :-)

Grinnz:
From a look through the code, it intentionally avoids stripping line endings as the lines are appended as is to the output. The only thing it does is normalize them by removing the extranous `\r` bytes.

khwilliamson:
I haven't yet looked at the output.  If the input is opened to strip off \r, the output should correspondingly put them back.

But the code attempting to remove \r is currently useless.  The paragraph slurp functionality doesn't see two \n\n in a row, so it reads the whole file (IIRC) as a single paragraph.

The paragraph slurping is not likely to be the only code in core that is expecting only a single \n as the line separator.  And in fact, there are a bunch of failures in our test suite if git is not configured to make line endings strictly \n.  This is just the first one I'm looking at, in part because it fails to build a proper toc as a result.  Another obvious one is checksums.  The porting tests that compare those fail because of the extra \r.  If those are opened with stripping off \r, the tests pass.  But maybe the better solution is to skip those on such systems.

I haven't looked at the other failures

jkeenan:
This pull request about `buildtoc` has not attracted support from other committers.  Do we want to proceed with it or should it be closed?

tonycoz (approved):

demerphq (approved):

PR #20488, opened 2022-11-05
locale.c: Omit final ';' in LC_ALL composite locale

khwilliamson:
When locales for categories differ, LC_ALL needs a way to express the individual components.  Prior to this commit, the aggregation ended with a semi-colon, which is unnecessary.  This commit removes that.

It also moves a loop counter declaration into the loops, now that we have C99, and avoids an extra strlen() call.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -1660,19 +1660,19 @@ S_calculate_LC_ALL(pTHX_ const char ** individ_locales)
      * categories, adding new ones as they show up on obscure platforms.
      */
 
-    unsigned int i;
     Size_t names_len = 0;
     bool are_all_categories_the_same_locale = TRUE;
     char * aggregate_locale;
     char * previous_start = NULL;
     char * this_start = NULL;
-    Size_t entry_len = 0;
+    SSize_t entry_len = 0;
+    SSize_t prev_entry_len;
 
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -1683,16 +1683,14 @@ S_calculate_LC_ALL(pTHX_ const char ** individ_locales)
...
     }
 
-    names_len++;    /* Trailing '\0' */
-
     /* Allocate enough space for the aggregated string */
     SAVEFREEPV(Newxz(aggregate_locale, names_len, char));
 
     /* Then fill it in */
-    for (i = 0; i < NOMINAL_LC_ALL_INDEX; i++) {
+    for (unsigned i = 0; i < NOMINAL_LC_ALL_INDEX; i++) {
```
demerphq, line 1693: I know it is legal, but could we spell this out fully? it makes me a bit leery to see a raw unsigned, even if it is legal.

demerphq (changes requested):

locale.c
jkeenan: @khwilliamson, have you had a chance to think about @demerphq's question?

khwilliamson:
I will replace this PR with an updated one at some point

PR #18, opened 2018-10-14
Add test of out-of-source build to .travis.yml

krichter722:

PR #20496, opened 2022-11-08
ParseXS improvements

demerphq:
This includes a number of ParseXS changes. A version bump, tests that perlxs.pod has the correct version, improvements to parsing cpp processor directives so XS files will be less whitespace sensitive, and enhanced support for aliases.

### DIFF
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -1881,7 +1881,7 @@ sub fetch_para {
         #   others:    ident (gcc notes that some cpps have this one)
         || $self->{lastline} =~ /^\#[ \t]*
                                   (?:
-                                        (?:if|ifn?def|elif|else|endif|
+                                        (?:if|ifn?def|elif|else|endif|elifn?def|
                                            define|undef|pragma|error|
                                            warning|line\s+\d+|ident)
                                         \b
@@ -1892,7 +1892,7 @@ sub fetch_para {
     )
     {
       last if $self->{lastline} =~ /^\S/ && @{ $self->{line} } && $self->{line}->[-1] eq "";
-      if ($self->{lastline}=~/^#[ \t]*(if|ifn?def|elif|else|endif)\b/) {
+      if ($self->{lastline}=~/^#[ \t]*(if|ifn?def|elif|else|endif|elifn?def)\b/) {
         my $type = $1; # highest defined capture buffer, "if" for any if like condition
         if ($type =~ /^if/) {
           if (@{$self->{line}}) {
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -48,7 +48,7 @@ our @EXPORT_OK = qw(
 
 ##############################
 # A number of "constants"
-
+our $DIE_ON_ERROR;
 our ($C_group_rex, $C_arg);
 # Group in C (no support for comments or literals)
 $C_group_rex = qr/ [({\[]
@@ -104,6 +104,7 @@ sub process_file {
     typemap         => [],
     versioncheck    => 1,
     FH              => Symbol::gensym(),
+    die_on_error    => $DIE_ON_ERROR, # if true we die() and not exit() after errors
     %options,
   );
   $args{except} = $args{except} ? ' TRY' : '';
@@ -133,6 +134,8 @@ sub process_file {
   $self->{WantVersionChk} = $args{versioncheck};
   $self->{WantLineNumbers} = $args{linenumbers};
   $self->{IncludedFiles} = {};
+
+  $self->{die_on_error} = $args{die_on_error};
 
   die "Missing required parameter 'filename'" unless $args{filename};
   $self->{filepathname} = $args{filename};
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -1800,11 +1800,17 @@ sub fetch_para {
     $self->_process_module_xs_line($1, $2, $3);
   }
 
+  # count how many #ifdef levels we see in this paragraph
+  # decrementing when we see an endif. if we see an elsif
+  # or endif without a corresponding #ifdef then we dont
+  # consider it part of this paragraph.
+  my $if_level = 0;
   for (;;) {
     $self->_maybe_skip_pod;
 
     $self->_maybe_parse_typemap_block;
 
+    my $final;
     if ($self->{lastline} !~ /^\s*#/ # not a CPP directive
         # CPP directives:
         #    ANSI:    if ifdef ifndef elif else endif define undef
@@ -1825,6 +1831,31 @@ sub fetch_para {
     )
     {
       last if $self->{lastline} =~ /^\S/ && @{ $self->{line} } && $self->{line}->[-1] eq "";
+      if ($self->{lastline}=~/^#[ \t]*(if|ifn?def|elif|else|endif)\b/) {
+        my $type = $1; # highest defined capture buffer, "if" for any if like condition
+        if ($type =~ /^if/) {
+          if (@{$self->{line}}) {
+            # increment level
+            $if_level++;
+          } else {
+            $final = 1;
+          }
+        } elsif ($type eq "endif") {
+          if ($if_level) { # are we in an if that was started in this paragraph?
+            $if_level--;   # yep- so decrement to end this if block
+          } else {
+            $final = 1;
+          }
+        } elsif (!$if_level) {
+          # not in an #ifdef from this paragraph, thus
+          # this directive should not be part of this paragraph.
+          $final = 1;
+        }
+      }
+      if ($final and @{$self->{line}}) {
+        return 1;
+      }
+
       push(@{ $self->{line} }, $self->{lastline});
       push(@{ $self->{line_no} }, $self->{lastline_no});
     }
@@ -1838,6 +1869,9 @@ sub fetch_para {
 
     chomp $self->{lastline};
     $self->{lastline} =~ s/^\s+$//;
+    if ($final) {
+      last;
+    }
   }
 
   # Nuke trailing "line" entries until there's one that's not empty
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -31,6 +31,7 @@ use ExtUtils::ParseXS::Utilities qw(
   analyze_preprocessor_statements
   set_cond
   Warn
+  WarnHint
   current_line_number
   blurt
   death
@@ -1312,24 +1313,71 @@ sub get_aliases {
 
   # Parse alias definitions
   # format is
-  #    alias = value alias = value ...
+  #    alias = value Pack::alias = value ...
+  # or
+  #    alias => other
+  # or
+  #    alias => Pack::other
+  # or
+  #    Pack::alias => Other::alias
 
-  while ($line =~ s/^\s*([\w:]+)\s*=\s*(\w+)\s*//) {
-    my ($alias, $value) = ($1, $2);
+  while ($line =~ s/^\s*([\w:]+)\s*=(>?)\s*([\w:]+)\s*//) {
+    my ($alias, $is_symbolic, $value) = ($1, $2, $3);
     my $orig_alias = $alias;
+
+    blurt( $self, "Error: In alias definition for '$alias' the value may not"
+                  . " contain ':' unless it is symbolic.")
+        if !$is_symbolic and $value=~/:/;
 
     # check for optional package definition in the alias
     $alias = $self->{Packprefix} . $alias if $alias !~ /::/;
 
-    # check for duplicate alias name & duplicate value
-    Warn( $self, "Warning: Ignoring duplicate alias '$orig_alias'")
-      if defined $self->{XsubAliases}->{$alias};
+    if ($is_symbolic) {
+      my $orig_value = $value;
+      $value = $self->{Packprefix} . $value if $value !~ /::/;
+      if (!defined $self->{XsubAliases}->{$value}) {
+        blurt( $self, "Error: Unknown alias '$value' in symbolic definition for '$orig_alias'");
+      }
+      $value = $self->{XsubAliases}->{$value};
+    }
 
-    Warn( $self, "Warning: Aliases '$orig_alias' and '$self->{XsubAliasValues}->{$value}' have identical values")
-      if $self->{XsubAliasValues}->{$value};
+    # check for duplicate alias name & duplicate value
+    my $prev_value = $self->{XsubAliases}->{$alias};
+    if (defined $prev_value) {
+      if ($prev_value eq $value) {
+        Warn( $self, "Warning: Ignoring duplicate alias '$orig_alias'")
+      } else {
+        Warn( $self, "Warning: Conflicting duplicate alias '$orig_alias'"
+                     . " changes definition from '$prev_value' to '$value'");
+        delete $self->{XsubAliasValues}->{$prev_value}{$alias};
+      }
+    }
+
+    # Check and see if this alias results in two aliases having the same
+    # value, we only check non-symbolic definitions as the whole point of
+    # symbolic definitions is to say we want to duplicate the value and
+    # it is NOT a mistake.
+    unless ($is_symbolic) {
+      my @keys= sort keys %{$self->{XsubAliasValues}->{$value}||{}};
+      if (@keys) {
+        @keys= map { "'$_'" }
+               map { my $copy= $_;
+                     $copy=~s/^$self->{Packprefix}//;
+                     $copy
+                   } @keys;
+        WarnHint( $self,
+                  "Warning: Aliases '$orig_alias' and "
+                  . join(", ", @keys)
+                  . " have identical values",
+                  !$self->{XsubAliasValueClashHinted}++
+                  ? "If this is deliberate use a symbolic alias instead."
+                  : undef
+        );
+      }
+    }
 
     $self->{XsubAliases}->{$alias} = $value;
-    $self->{XsubAliasValues}->{$value} = $orig_alias;
+    $self->{XsubAliasValues}->{$value}{$alias}++;
   }
 
   blurt( $self, "Error: Cannot parse ALIAS definitions from '$orig'")
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -11,7 +11,7 @@ use Symbol;
 
 our $VERSION;
 BEGIN {
-  $VERSION = '3.47';
+  $VERSION = '3.48';
   require ExtUtils::ParseXS::Constants; ExtUtils::ParseXS::Constants->VERSION($VERSION);
   require ExtUtils::ParseXS::CountLines; ExtUtils::ParseXS::CountLines->VERSION($VERSION);
   require ExtUtils::ParseXS::Utilities; ExtUtils::ParseXS::Utilities->VERSION($VERSION);
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -873,7 +873,7 @@ EOF
        "        (void)$self->{newXS}(\"$self->{pname}\", XS_$self->{Full_func_name}$self->{file}$self->{proto});\n");
     }
 
-    for my $operator (keys %{ $self->{OverloadsThisXSUB} }) {
+    for my $operator (sort keys %{ $self->{OverloadsThisXSUB} }) {
       $self->{Overloaded}->{$self->{Package}} = $self->{Packid};
       my $overload = "$self->{Package}\::($operator";
       push(@{ $self->{InitFileCode} },
@@ -881,7 +881,7 @@ EOF
     }
   } # END 'PARAGRAPH' 'while' loop
 
-  for my $package (keys %{ $self->{Overloaded} }) { # make them findable with fetchmethod
+  for my $package (sort keys %{ $self->{Overloaded} }) { # make them findable with fetchmethod
     my $packid = $self->{Overloaded}->{$package};
     print Q(<<"EOF");
 #XS_EUPXS(XS_${packid}_nil); /* prototype to pass -Wmissing-prototypes */
@@ -968,7 +968,7 @@ EOF
 #
 EOF
 
-  if (%{ $self->{Overloaded} }) {
+  if (keys %{ $self->{Overloaded} }) {
     # once if any overloads
     print Q(<<"EOF");
 #    /* register the overloading (type 'A') magic */
@@ -976,7 +976,7 @@ EOF
 #    PL_amagic_generation++;
 ##endif
 EOF
-    for my $package (keys %{ $self->{Overloaded} }) {
+    for my $package (sort keys %{ $self->{Overloaded} }) {
       # once for each package with overloads
       my $fallback = $self->{Fallback}->{$package} || "&PL_sv_undef";
       print Q(<<"EOF");
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -1311,6 +1311,9 @@ sub get_aliases {
   my ($line) = @_;
   my ($orig) = $line;
 
+  # we use this later for symbolic aliases
+  my $fname = $self->{Packprefix} . $self->{func_name};
+
   # Parse alias definitions
   # format is
   #    alias = value Pack::alias = value ...
@@ -1335,10 +1338,13 @@ sub get_aliases {
     if ($is_symbolic) {
       my $orig_value = $value;
       $value = $self->{Packprefix} . $value if $value !~ /::/;
-      if (!defined $self->{XsubAliases}->{$value}) {
+      if (defined $self->{XsubAliases}->{$value}) {
+        $value = $self->{XsubAliases}->{$value};
+      } elsif ($value eq $fname) {
+        $value = 0;
+      } else {
         blurt( $self, "Error: Unknown alias '$value' in symbolic definition for '$orig_alias'");
       }
-      $value = $self->{XsubAliases}->{$value};
     }
 
     # check for duplicate alias name & duplicate value
@@ -1359,6 +1365,10 @@ sub get_aliases {
     # it is NOT a mistake.
     unless ($is_symbolic) {
       my @keys= sort keys %{$self->{XsubAliasValues}->{$value}||{}};
+      # deal with an alias of 0, which might not be in the XsubAlias dataset
+      # yet as 0 is the default for the base function ($fname)
+      push @keys, $fname
+        if $value eq "0" and !defined $self->{XsubAlias}{$fname};
       if (@keys) {
         @keys= map { "'$_'" }
                map { my $copy= $_;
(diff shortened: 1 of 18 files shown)


jkeenan (commented):
When I build this branch up through this commit 0b3c72025a, I start to get failures in `dist/ExtUtils-ParseXS/t/002-more.t;`:
```
$ cd t;./perl harness -v ../dist/ExtUtils-ParseXS/t/002-more.t; cd -

ok 1 - require ExtUtils::ParseXS;
Warning: Conflicting duplicate alias 'return_1' changes definition from '0' to 'RET_1' in XSMore.xs, line 194
ok 2 - Create an output file
ok 3 - ExtUtils::CBuilder::compile() returned true value
ok 4 - Make sure XSMore.o exists
ok 5 - ExtUtils::CBuilder::link() returned true value
ok 6 - Make sure XSMore.so exists
ok 7 - No error message recorded, as expected
ok 8 - ExtUtils::ParseXS::errors()
ok 9 - the BOOT keyword
ok 10 - the INCLUDE keyword
ok 11 - the PROTOTYPES keyword
ok 12 - the PROTOTYPE keyword
ok 13 - the ATTRS keyword
ok 14 - ATTRS with prototype
not ok 15 - the CASE keyword (1)

#   Failed test 'the CASE keyword (1)'
#   at t/002-more.t line 78.
#          got: '*XSMore::return_1'
#     expected: '1'
ok 16 - the CASE keyword (2)
ok 17 - ALIAS with prototype (1)
```

jkeenan:
When building this branch on FreeBSD with `clang10` as the compiler, `make` is failing here:
```
Writing Makefile for DynaLoader
"../../miniperl" "-I../../lib" DynaLoader_pm.PL DynaLoader.pm
rm -f DynaLoader.xs
cp dl_dlopen.xs DynaLoader.xs
"../../miniperl" "-I../../lib" "../../lib/ExtUtils/xsubpp" -noprototypes -typemap '/usr/home/jkeenan/gitwork/perl/ext/DynaLoader/../../lib/ExtUtils/typemap'  DynaLoader.xs > DynaLoader.xsc
Error: 'endif' with no matching 'if' in DynaLoader.xs, line 193
*** Error code 1

Stop.
make[1]: stopped in /usr/home/jkeenan/gitwork/perl/ext/DynaLoader
"../../miniperl" "-I../../lib" "../../lib/ExtUtils/xsubpp" -noprototypes -typemap '/usr/home/jkeenan/gitwork/perl/ext/DynaLoader/../../lib/ExtUtils/typemap'  DynaLoader.xs > DynaLoader.xsc
Error: 'endif' with no matching 'if' in DynaLoader.xs, line 193
*** Error code 1

Stop.
make[1]: stopped in /usr/home/jkeenan/gitwork/perl/ext/DynaLoader
Unsuccessful make(ext/DynaLoader): code=256 at make_ext.pl line 584.
*** Error code 25

Stop.
```

demerphq:
On Tue, 8 Nov 2022, 22:06 James E Keenan, ***@***.***> wrote:

>
> When I build this branch up through this commit, I start to get failures
> in dist/ExtUtils-ParseXS/t/002-more.t;:
>
> $ cd t;./perl harness -v ../dist/ExtUtils-ParseXS/t/002-more.t; cd -
>
> ok 1 - require ExtUtils::ParseXS;
> Warning: Conflicting duplicate alias 'return_1' changes definition from '0' to 'RET_1' in XSMore.xs, line 194
> ok 2 - Create an output file
> ok 3 - ExtUtils::CBuilder::compile() returned true value
> ok 4 - Make sure XSMore.o exists
> ok 5 - ExtUtils::CBuilder::link() returned true value
> ok 6 - Make sure XSMore.so exists
> ok 7 - No error message recorded, as expected
> ok 8 - ExtUtils::ParseXS::errors()
> ok 9 - the BOOT keyword
> ok 10 - the INCLUDE keyword
> ok 11 - the PROTOTYPES keyword
> ok 12 - the PROTOTYPE keyword
> ok 13 - the ATTRS keyword
> ok 14 - ATTRS with prototype
> not ok 15 - the CASE keyword (1)
>
> #   Failed test 'the CASE keyword (1)'
> #   at t/002-more.t line 78.
> #          got: '*XSMore::return_1'
> #     expected: '1'
> ok 16 - the CASE keyword (2)
> ok 17 - ALIAS with prototype (1)
>
>
Strange. I will dig. Thanks for the report.

Yves

>

demerphq:
FWIW, I pushed this as a smoke-me branch to see what kind of reports we get with it:

smoke-me/yves_parsexs

jkeenan (commented):
In the commit message for 589b5f18a882b7b759175826a08022d22714743f, what does it mean for output to be 'deterministic'? What would be the benefit of this?

jkeenan (commented):
Commit 34781067e5c44b97cf43571e48885e9a124c485c introduces new functionality:  "symbolic alias".  [Will need review by someone more familiar with actual XS.]
Commit message indicates commit does several different things at once.  Could this be confusing or difficult to maintain?

jkeenan (commented):
Can you provide an example of where (i) in current blead (or 5.36) we get warnings that you consider to be inadequate (or fail to get warnings that would be helpful) and (ii) after merging this branch we will get better warnings?

Can you provide a simple illustration of how WarnHint() will improve warnings emitted by EU::PXS?

demerphq:
Hi @jkeenan, thanks for the feedback.

> what does it mean for output to be 'deterministic'?

the code builds an array of code for output using a raw keys statement which produce a non-deterministic ordering of their keys.
<details> 

Its much like this:
```
$ perl -le'%h=("a".."z"); print keys %h'
eysmiagkcwquo
$ perl -le'%h=("a".."z"); print keys %h'
qiekmuwgayosc
$ perl -le'%h=("a".."z"); print keys %h'
yiksqwucemgao
```
thus the output is not deterministic, and from run to run it might change.  Sorting the keys means the data is output in a consistent order every time.  Which means that with no input change the output should be identical. As far as I could tell that wasn't true for the code involved.  To be honest the change didn't break any tests, so it is possible something is sorting this data elsewhere and also providing deterministic output, but I couldn't be sure. Adding the `sort` at worst is a no-op change, and at best ensures deterministic output.
</details>

> Commit message indicates commit does several different things at once

Ah, I see how you come to that conclusion. What it means is something like "in the process of adding this functionality I necessarily took care of some other edge case issues relating to aliases that were affected by my change, while at the same time correcting their logic to be consistent.". I can change the commit message if you want, it seemed clear to me.  

<details>
Before we start here is a brief explanation of what the ALIAS mechanism is for and does at a high level. It is common in XS to have a number of functions which require standard boiler plate around a smaller set of "internal code". This is common practice, for instance all the logic to validate arguments might be identical for all the methods in a package, similarly code to generate constants from header values might do the same thing.  So XS provides a mechanism where it will use one C function for multiple Perl functions. When the C function is called an integer value (usually called `ix` IIRC)  is set to signal to the C code which Perl function was called (or rather which `ix` value is associated with the function name that was called). Thus there is a level of indirection involved on the mapping.  Consider the following XS sub header:

```
void
do(dbh)
   SV *dbh
ALIAS:
    do_fast = 1
    do_slow = 2
```

This would create one C/XS function, and three perl function entries for `do()`, `do_fast()` and `do_slow()`, which when called will call the same underlying C/XS function with the `ix` value being 0, 1, or 2 respectively. The C/XS code can then switch on `ix` and perform the desired operation appropriately.

This means collisions are possible, both at the name level (a name might be listed twice) or the value level (a value might be listed twice). The ParseXS code has historically tried to warn when either happened (modulo some bugs in its implementation). So for instance the old code will warn you if you have something like this:

```
ALIAS:
   foo = 1
   foo = 1
```

it will tell you it is "ignoring" the second "foo". However this is not entirely correct (and in fact is only true because the values are the same), if you feed it:

```
ALIAS:
   foo = 1
   foo = 2
```

it will again tell you that is ignoring the second foo, but it doesn't actually /ignore/ the second foo, it *overwrites* the value of foo so it ends up as 2 (and along the way gets confused about what values have been used).  Since my objective with the symbolic aliases is to allow an engineer to specify "this alias is deliberately a dupe of that other alias (or sub) and thus should not produce a warning", I necessarily had to change the logic for warnings and while I was doing so I noticed the issues with the existing warnings.

ParseXS also has warnings about multiple names having the same value. So for instance

```
void
do(dbh)
   SV *dbh
ALIAS:
    do1 = 1
    do2 = 1
```

you will get a warning about do1 and do2 having the same value, which was actually my core objective, as I have legit reasons to create a dupe and I don't want warnings about them, and there is no way to turn the warning off.  So I introduced syntax like this:

```
void
do(dbh)
   SV *dbh
ALIAS:
    do1 => do
    do2 => do
```

which would be the same as this:

```
void
do(dbh)
   SV *dbh
ALIAS:
    do1 = 0
    do2 = 0
```

except it would not throw warnings. I chose `=>` because the grammar requires a separator that does not match `[\w:]`. I chose to do 'symbolic aliasing' because I think that when people do this kind of thing intentionally they want to specify "these names should have the same index", where-as if I had simply added a new separator to signal intentionality but kept the values still "numeric" then it would increase the maint burden if the `ix` for the group were to change.  

Interestingly there is another long standing bug in the dupe value warning code in that

```
void
do(dbh)
   SV *dbh
ALIAS:
    do1 = 0
```

historically does not warn, duping the default value of the base function (0) only warns on the second alias. From the point of view of the internals of ParseXS I can understand how this bug came about, but it is inconsistent with the behavior of other alias values, so I fixed it to warn similarly to how it would handle any other value collision.

Anyway, adding this new form of alias declaration resulting in new edge cases being added to the warning detection for some of this, for instance consider this:

```
void
do(dbh)
   SV *dbh
ALIAS:
    do1 = 1
    do2 => do1
    do3 = 1 
```

do2 should not warn that it is a dupe of 1, but do3 should. While working on these cases I noticed the other issues. Since the all involve the same underlying logic I dealt with them all at once.  As far as I intended and understand nothing that used to warn will stop warning, and in only one condition, where a developer is using the default value of 0 for an alias (possibly without knowing) are any new warnings added. On the other hand the warnings, if they were changed at all, should be more informative.
</details>

> Can you provide an example of where ... we will get better warnings?
> how WarnHint() will improve warnings emitted by EU::PXS?

Well part of the intent of this PR is to provide a way to silence unwanted warnings. If I want to alias "foo" and "bar" to the same value internally (maybe for backcompat or something like that) I should be able to do so in a way that xsubpp does not warn, especially when those warnings will be visible to end users who cant do anything about them. None of the core code intentionally does this type of aliasing (probably to avoid warnings), so there are no examples in /core/, and since I added new syntax to the XS module grammar when one does want to do this none of our existing core code would exercise this. However you can see it in the tests. Similarly for the hint message.

<details>
You can see pretty much all the changes in one view with the tests in t/001_basic.t and t/XSAlias.t, This is the XS:

```
void
do(dbh)
   SV *dbh
ALIAS:
    dox = 1
    lox => dox
    pox = 1
    pox = 2
    docks = 1
    dachs => lox
    xunx = 0
    xukes => do
```

and the test output:

```
  is $stderr,
    "Warning: Aliases 'pox' and 'dox', 'lox' have"
    . " identical values of 1 in XSAlias.xs, line 9\n"
    . "    (If this is deliberate use a symbolic alias instead.)\n"
    . "Warning: Conflicting duplicate alias 'pox' changes"
    . " definition from '1' to '2' in XSAlias.xs, line 10\n"
    . "Warning: Aliases 'docks' and 'dox', 'lox' have"
    . " identical values of 1 in XSAlias.xs, line 11\n"
    . "Warning: Aliases 'xunx' and 'do' have identical values"
    . " of 0 - the base function in XSAlias.xs, line 13\n",
    "Saw expected warnings from XSAlias.xs";
```
which expands out to
```
Warning: Aliases pox and dox, lox have identical values of 1 in XSAlias.xs, line 9
    (If this is deliberate use a symbolic alias instead.)
Warning: Conflicting duplicate alias pox changes definition from 1 to 2 in XSAlias.xs, line 10
Warning: Aliases docks and dox, lox have identical values of 1 in XSAlias.xs, line 11
Warning: Aliases xunx and do have identical values of 0 - the base function in XSAlias.xs, line 13
```
the result of using WarnHint is visible on the first two lines. The parenthesized message about using a symbolic reference to avoid the warning is the "Hint".  

The fix for the incorrect claim about "ignoring" a definition is visible on the third line. Previously the message would not show the value nor show it changing. 

The fix for the "ignoring" message overall to show all aliases and to show the value is visible on the first and fourth lines, previously we would not show aliases nor would we show the value, and the fix to now show a warning when the default value of 0 is duplicated is visible on the last line. I chose to make the symbolic alias hint "show once" so as not to clutter up the warning list. The point is to call the devs attention to the new feature so they can change their code not to warn if they are intending to create a dupe, IMO we don't need to do so every case there is an alias collision.

If I change all the `=>` style alias entries to the equivalent '=' entry 

```
void
do(dbh)
   SV *dbh
ALIAS:
    dox = 1
    lox = 1
    pox = 1
    pox = 2
    docks = 1
    dachs = 1
    xunx = 0
    xukes = 0
```
then we get this on an older perl:
```
Warning: Aliases 'lox' and 'dox' have identical values in XSAlias.xs, line 8
Warning: Aliases 'pox' and 'lox' have identical values in XSAlias.xs, line 9
Warning: Ignoring duplicate alias 'pox' in XSAlias.xs, line 10
Warning: Aliases 'docks' and 'pox' have identical values in XSAlias.xs, line 11
Warning: Aliases 'dachs' and 'docks' have identical values in XSAlias.xs, line 12
Warning: Aliases 'xukes' and 'xunx' have identical values in XSAlias.xs, line 14
```

You can see the outright error on line 3 (the value is not being ignored, the old value is being overwritten), and if you compare closely you will notice the warning about xunx and do is missing.  The other point is that even though I *meant* to alias `lox` and `dox` I got a warning I don't want! 

</details>

Hopefully that explains things sufficiently. :-)

jkeenan:
Smoke testing results are satisfactory.  I was able to do a *limited* amount of testing of CPAN distros with XS code against a perl built from this branch; results also satisfactory.

At this point what we could use is some review of the changes by someone who (unlike me) has actually *written* XS code.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
```diff
@@ -1838,12 +1869,18 @@ sub fetch_para {
...
+    }
   }
 
   # Nuke trailing "line" entries until there's one that's not empty
   pop(@{ $self->{line} }), pop(@{ $self->{line_no} })
     while @{ $self->{line} } && $self->{line}->[-1] eq "";
 
+  #use Data::Dumper;
+  #warn Dumper("late",$self->{line},$self->{line_no}, $self->{lastline});
+
```
tonycoz: left over debugging

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
```diff
@@ -1825,6 +1831,31 @@ sub fetch_para {
     )
     {
       last if $self->{lastline} =~ /^\S/ && @{ $self->{line} } && $self->{line}->[-1] eq "";
+      if ($self->{lastline}=~/^#[ \t]*(if|ifn?def|elif|else|endif)\b/) {
```
tonycoz: The upcoming C++23 and C23 standards also add `#elifdef`, `#elifndef`, though I'm not sure you need any of the `#else` variants here.

https://en.cppreference.com/w/cpp/preprocessor
https://en.cppreference.com/w/c/preprocessor

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
```diff
@@ -1312,24 +1313,71 @@ sub get_aliases {
...
+  #    Pack::alias => Other::alias
+
+  while ($line =~ s/^\s*([\w:]+)\s*=(>?)\s*([\w:]+)\s*//) {
+    my ($alias, $is_symbolic, $value) = ($1, $2, $3);
     my $orig_alias = $alias;
 
+    blurt( $self, "Error: In alias definition for '$alias' the value may not"
+                  . " contain ':' unless it is symbolic.")
+        if !$is_symbolic and $value=~/:/;
+
```
tonycoz, line 1337: ParseXS is also used for C++ where a constant name containing `::` is normal, though that would be an enhancement.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
demerphq: Ack, ill clean it up.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
demerphq: I can add support for them, but as you stated the only thing they would be used for would be to detect one without a corresponding starting 'if'.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
demerphq: Really not sure what you want me to do here. Add the enhancement to the PR since we are doing a version bump anyway? Or create a new PR? As you said, currently we dont support this form of value name at all, it would have to be wrapped in a define.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
demerphq: Added support for it.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
demerphq: Done.

dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
tonycoz: You can ignore this.

I could see ALIAS supporting C++ symbols, or even general constant expressions, eg:
```
ALIAS:
  get_alpha: TAG_ALPHA * 2
  set_alpha: TAG_ALPHA * 2 + 1
...
```
but there's no particular reason to include it in this PR.

demerphq: FWIW, this PR includes doc changes that show how to work around any limitations in this area. As long as these expressions are hidden behind a define they are fine.

demerphq:
@tonycoz @jkeenan what do you think? Mergable yet?

jkeenan:
> @tonycoz @jkeenan what do you think? Mergable yet?

In general, I like what I see here, but I have two principal concerns.

**First:**  On CPAN there are over 2600 distributions that contain XS code.  That's less than 10 percent of the total, but I suspect that two things are simultaneously true: (i) they're more idiosyncratic than CPAN as a whole; and (ii) some of them are vastly more important than 99% of what's on CPAN.  Many distros with XS have been broken and unmaintained for years, so getting a clean baseline for the impact of this p.r. is not easy.

If we release this as part of 5.37.6 on November 20, the heavy-duty CPANtesters will start testing "all" of CPAN, but it takes time for them to look at FAILs and determine whether the failures are both new and attributable to the code that has just been merged into blead.  We have to be prepared for the possibility that flaws in this code which would warrant a revert might not appear for two week or more.  When all is said and done this p.r. is only adding a very modest amount of new functionality.  If we get a lot of BBC reports we should not burden CPAN authors with the need to do work-arounds to accommodate that functionality.

**Second:** ExtUtils-ParseXS is a dual-release, blead-first distribution.  That means that at some point after this p.r. gets merged, someone will have to do a new CPAN release (the most recent was done by @xsawyerx in January of this year) and, as part of preparation for that release, will have to test this against older versions of perl.  Based on [CPANtesters results](http://matrix.cpantesters.org/?dist=ExtUtils-ParseXS) for the latest version, I'd say that we would want EU:PXS to PASS going back to perl-5.8.9.  Has there been any assessment of how well this code will work with those older versions?  I may have missed something, but I don't think this concern has been discussed so far in this ticket.

demerphq:
@jkeenan I take your point. But the more popular cpan module are usually tested early in the process right? So I think we would get pretty decent feedback pretty fast.

Also, i think its worth noting here, nothing I have done (except one warning) changes the behavior of anything that isn't already throwing an exception. For instance the '#else' patch actually makes things that used to throw an error not throw errors. So it is unlikely (albeit possible!) that it will break anything. I think if it does we will see signal very quickly. If it does cause BBC then I would say we would revert it on first sign of trouble. 

As for back-compat we test that it compiles with older perls as far as I understand it. I had a breakage because i used `s///r` which i fixed already. So I think it should be good. Note https://github.com/Perl/perl5/actions/runs/3464003421/jobs/5785197085#step:5:575 test that it builds and passes test on 5.8

jkeenan:
> @jkeenan I take your point. But the more popular cpan module are usually tested early in the process right? So I think we would get pretty decent feedback pretty fast.
> 

I don't actually know the order in which people like @andk, @eserte, @bingos test CPAN distros after a monthly dev release.  I *do* know that the process of confirming that some CPANtesters failure is a BBC (and, beyond that, bisecting to a particular commit to blead) is a process that cannot be completely automated as it requires human judgment, hence takes time.

But, assuming @tonycoz is okay with your more recent commits, I'd say this is good to go.  Thanks.

tonycoz:
> Has there been any assessment of how well this code will work with those older versions?

The github workflow tests all of the modules in dist/ against 5.8, 5.10, 5.24 and the most recent non-development release, 5.36 right now.

This doesn't mean it's tested against all of the XS modules on CPAN.

tonycoz (approved):

demerphq:
Ill merge it and lets find out.

iabyn:
On Wed, Nov 16, 2022 at 01:09:15AM -0800, Yves Orton wrote:
> Ill merge it and lets find out. 

I'm seeing lots of new warnings on STDERR since the merge. Is this
expected? Is it the XS distributions that are wrong and need fixing?
And if so have it been reported upstream?

Warning: Aliases 'Digest::SHA::sha1' and 'sha1' have identical values of 0 - the base function in SHA.xs, line 118
    (If this is deliberate use a symbolic alias instead.)
Warning: Aliases 'Digest::SHA::hmac_sha1' and 'hmac_sha1' have identical values of 0 - the base function in SHA.xs, line 174
Warning: Aliases 'Digest::SHA::hashsize' and 'hashsize' have identical values of 0 - the base function in SHA.xs, line 235
Warning: Aliases 'Digest::SHA::digest' and 'digest' have identical values of 0 - the base function in SHA.xs, line 272
Warning: Aliases 'B::OP::next' and 'next' have identical values of 0 - the base function in B.xs, line 816
    (If this is deliberate use a symbolic alias instead.)
Warning: Aliases 'min' and 'min' have identical values of 0 - the base function in ListUtil.xs, line 274
    (If this is deliberate use a symbolic alias instead.)
Warning: Aliases 'sum' and 'sum' have identical values of 0 - the base function in ListUtil.xs, line 326
Warning: Aliases 'reduce' and 'reduce' have identical values of 0 - the base function in ListUtil.xs, line 545
Warning: Aliases 'none' and 'any' have identical values of 0 - the base function in ListUtil.xs, line 702
Warning: Aliases 'B::PADNAMELIST::MAX' and 'PadlistMAX' have identical values of 0 - the base function in B.xs, line 2167
Warning: Aliases 'head' and 'head' have identical values of 0 - the base function in ListUtil.xs, line 769
Warning: Aliases 'uniqint' and 'uniq' have identical values of 0 - the base function in ListUtil.xs, line 1325
Warning: Aliases 'bytes2str' and 'decode' have identical values of 0 - the base function in Encode.xs, line 982
    (If this is deliberate use a symbolic alias instead.)
Warning: Aliases 'str2bytes' and 'encode' have identical values of 0 - the base function in Encode.xs, line 1004
Warning (mostly harmless): No library found for -lposix
Warning (mostly harmless): No library found for -lcposix
Warning: Aliases 'isNFC_NO' and 'isComp_Ex' have identical values of 0 - the base function in Normalize.xs, line 814
    (If this is deliberate use a symbolic alias instead.)


-- 
Wesley Crusher gets beaten up by his classmates for being a smarmy git,
and consequently has a go at making some friends of his own age for a
change.
    -- Things That Never Happen in "Star Trek" #18

demerphq:
On Wed, 16 Nov 2022 at 10:26, iabyn ***@***.***> wrote:

> On Wed, Nov 16, 2022 at 01:09:15AM -0800, Yves Orton wrote:
> > Ill merge it and lets find out.
>
> I'm seeing lots of new warnings on STDERR since the merge. Is this
> expected?


In the sense that I thought we might see a few of these yes.


> Is it the XS distributions that are wrong and need fixing?
>

In the most prescriptive technical sense yes. In the practical sense it is
a bit more debatable. We have always warned when two aliases were created
to the same value. However a bug meant that this wasn't applied to the
value 0. I fixed that bug, on the assumption that this would likely be
rare.  I have to admit I did not notice the warnings from our own dists, or
I would have disabled the new warning.

The simplest solution is I can push a PR to disable this new warning. But
if we disable it for 0 why do we have it at all? Why should 0 be special
but the others no? Eg, if you do this:

void
foo (arg)
        SV * arg
  ALIAS: bar = 0

there is no warning that foo and bar are aliased to the same value. But if
you do the equivalent

void
foo (arg)
        SV * arg
  ALIAS: foo = 0 bar = 0

there will be, and if you override the base functions ix value like this:

void
foo (arg)
        SV * arg
  ALIAS: foo = 1 bar = 1

 there also will be a warning about bar.

BTW the above is why i considered the original behavior of allowing one
alias to 0 without warning that it creates dupes to be a bug. The code did
not consider the base function to be "an alias" unless it
was explicitly listed.



> And if so have it been reported upstream?
>

No I had not noticed this and was hoping that we would get little to no BBC
reports from it (it is a warning). My bad. I will prepare a PR to disable
this warning while we decided what to do. I had expected this to be only a
very rare case, to realize there are so many in core itself suggests that
is very wrong and this warning will be unhelpful. I might argue we simply
should warn about this at all then, i found the warnings sufficiently
unhelpful that I added support for new syntax to avoid them.

Note this merge was done about 20 minutes ago. This has not been in the
wild long.



> Warning: Aliases 'Digest::SHA::sha1' and 'sha1' have identical values of 0
> - the base function in SHA.xs, line 118
> (If this is deliberate use a symbolic alias instead.)
> Warning: Aliases 'Digest::SHA::hmac_sha1' and 'hmac_sha1' have identical
> values of 0 - the base function in SHA.xs, line 174
> Warning: Aliases 'Digest::SHA::hashsize' and 'hashsize' have identical
> values of 0 - the base function in SHA.xs, line 235
> Warning: Aliases 'Digest::SHA::digest' and 'digest' have identical values
> of 0 - the base function in SHA.xs, line 272
> Warning: Aliases 'B::OP::next' and 'next' have identical values of 0 - the
> base function in B.xs, line 816
> (If this is deliberate use a symbolic alias instead.)
> Warning: Aliases 'min' and 'min' have identical values of 0 - the base
> function in ListUtil.xs, line 274
> (If this is deliberate use a symbolic alias instead.)
> Warning: Aliases 'sum' and 'sum' have identical values of 0 - the base
> function in ListUtil.xs, line 326
> Warning: Aliases 'reduce' and 'reduce' have identical values of 0 - the
> base function in ListUtil.xs, line 545
> Warning: Aliases 'none' and 'any' have identical values of 0 - the base
> function in ListUtil.xs, line 702
> Warning: Aliases 'B::PADNAMELIST::MAX' and 'PadlistMAX' have identical
> values of 0 - the base function in B.xs, line 2167
> Warning: Aliases 'head' and 'head' have identical values of 0 - the base
> function in ListUtil.xs, line 769
> Warning: Aliases 'uniqint' and 'uniq' have identical values of 0 - the
> base function in ListUtil.xs, line 1325
> Warning: Aliases 'bytes2str' and 'decode' have identical values of 0 - the
> base function in Encode.xs, line 982
> (If this is deliberate use a symbolic alias instead.)
> Warning: Aliases 'str2bytes' and 'encode' have identical values of 0 - the
> base function in Encode.xs, line 1004
> Warning (mostly harmless): No library found for -lposix
> Warning (mostly harmless): No library found for -lcposix
> Warning: Aliases 'isNFC_NO' and 'isComp_Ex' have identical values of 0 -
> the base function in Normalize.xs, line 814
> (If this is deliberate use a symbolic alias instead.)
>

IMO what this says is that people want to duplicate aliases a lot more than
people think, and warning when they do is unhelpful.

I think maybe the best option would be to introduce a DEV_MODE type var to
ExtUtils::ParseXS so if people want extended warnings they can get them,
but that normally we would not warn about duplicate aliases.

cheers,
Yves

-- 

demerphq:
On Wed, 16 Nov 2022 at 10:44, demerphq ***@***.***> wrote:
> IMO what this says is that people want to duplicate aliases a lot more than people think, and warning when they do is unhelpful.

Or that it means that there is a lot of buggy code that we found with
this change. Its hard to say for sure, but occams razor suggests that
most of it is intentional.

Yves


-- 

demerphq:
@iabyn I have created https://github.com/Perl/perl5/pull/20517 to address this once we decide what we should do. We can also revert the patch that "fixed" this warning, but IMO that would be going backwards. This patch needs more tests before merging I guess, but if we want to silence this immediately we apply this and add the tests afterwards.

PR #20505, opened 2022-11-12
Decompose regcomp.c

demerphq:
This decomposes regcomp.c and moves about 40% or so of it into one of several different companion files, and moves some of the run time logic it contained into regexec.c as well. 

Previously:
```
$ wc -l reg*.[ch]
   3858 regcharclass.h
  26289 regcomp.c
   1431 regcomp.h
  11747 regexec.c
    947 regexp.h
     64 reginline.h
   2946 regnodes.h
  47282 total
```

Changed:

```
git diff --stat origin/blead reg*
 regcomp.c                  | 14105 ++++++++++++----------------------------------------------------------------------
 regcomp.h                  |     2 +-
 regcomp_debug.c            |  1627 ++++++++++
 regcomp_internal.h         |  1196 +++++++
 regcomp_invlist.c          |  1628 ++++++++++
 regcomp_study.c            |  3805 ++++++++++++++++++++++
 regcomp_trie.c             |  1688 ++++++++++
 regen/unicode_constants.pl |     4 +-
 regexec.c                  |   424 ++-
 regexp.h                   |     2 +-
 10 files changed, 12289 insertions(+), 12192 deletions(-)
```

Final:

```
$ wc -l reg*.[ch]
   3858 regcharclass.h
  16023 regcomp.c
   1627 regcomp_debug.c
   1434 regcomp.h
   1196 regcomp_internal.h
   1628 regcomp_invlist.c
   3805 regcomp_study.c
   1688 regcomp_trie.c
  12169 regexec.c
    947 regexp.h
     64 reginline.h
   2946 regnodes.h
  47385 total
```

This is a work in progress for now. Feedback welcomed.


hvds:
I'm not going to have time to go through this in detail any time soon. I think I'm happy with the overall direction though.

rwp0:
Awesome, but how about creating a `regex` directory and putting the related regular expression source files into (instead of prefixing in root)?

tonycoz:
> Awesome, but how about creating a `regex` directory and putting the related regular expression source files into (instead of prefixing in root)?

This was discussed on the list, see https://www.nntp.perl.org/group/perl.perl5.porters/2022/11/msg265047.html

win32/GNUmakefile
```diff
@@ -1154,7 +1159,15 @@ ifeq ($(CCTYPE),)
 endif
 
 
-..\regcomp$(o) : ..\regnodes.h ..\regcharclass.h
+..\regcomp$(o) : ..\regcomp.h ..\regcomp_internal.h ..\regnodes.h ..\regcharclass.h
```
tonycoz, line 1162: This doesn't require any action on your part, but this screams "automate me" to me.

tonycoz:
It looks good to me so far.  I might look at at the dependencies myself.

win32/GNUmakefile
demerphq: Oh god, me too. I already did some similar automation for the re module. But automating it in the win32 makefiles seemed a touch ambitious for me right now. :-)

demerphq:
>It looks good to me so far. I might look at at the dependencies myself.

Please do. Any advice, suggestions or feedback is appreciated. This was some crude shaving down of a large body of code, there are more things that could move, and more things we can do.

BTW, I have started writing a script that auto-analyzes and normalizes embed.fnc so it can be cleaned up and managed more easily than now. embed.fnc and related defines and macros are one of the biggest barriers to splitting our code.

demerphq:
@khwilliamson if i get this rebased properly are you ok with it?

khwilliamson (approved):
Each new file should be merely a move of the code extracted from regcomp.c, with tiny requisite changes.  I did not check this.  I trust the author to have properly done this.

But everything else looks fine to me.  I think this is a great step in the right direction.

demerphq:
@khwilliamson yes indeed, very few changes made, mostly related to scope and exports.

demerphq:
@tonycoz anything further?

embed.fnc
```diff
@@ -2064,14 +2065,16 @@ EXpRT	|I16	|do_uniprop_match|NN const char * const key|const U16 key_len
 EXpRT	|const char * const *|get_prop_values|const int table_index
 EXpR	|SV *	|get_prop_definition|const int table_index
 EXpRT	|const char *|get_deprecated_property_msg|const Size_t warning_offset
-#if defined(PERL_IN_REGCOMP_C)
-EiRT	|bool	|invlist_is_iterating|NN const SV* const invlist
-EiR	|SV*	|invlist_contents|NN SV* const invlist		    \
+#if defined(PERL_IN_REGCOMP_ANY)
+EpRT	|bool	|invlist_is_iterating|NN const SV* const invlist
+EpR	|SV*	|invlist_contents|NN SV* const invlist		    \
```
tonycoz: You've changed some of the inline invlist functions to non-inline global functions, should they instead be moved to invlist_inline.h and remain inline?

@khwilliamson might have an opinion here.

khwilliamson:
On 12/8/22 16:33, Tony Cook wrote:
> 
> 
> In embed.fnc 
> <https://github.com/Perl/perl5/pull/20505#discussion_r1043919085>:
> 
>> -#if defined(PERL_IN_REGCOMP_C)
> -EiRT	|bool	|invlist_is_iterating|NN const SV* const invlist
> -EiR	|SV*	|invlist_contents|NN SV* const invlist		    \
> +#if defined(PERL_IN_REGCOMP_ANY)
> +EpRT	|bool	|invlist_is_iterating|NN const SV* const invlist
> +EpR	|SV*	|invlist_contents|NN SV* const invlist		    \
> 
> You've changed some of the inline invlist functions to non-inline global 
> functions, should they instead be moved to invlist_inline.h and remain 
> inline?
> 
> @khwilliamson <https://github.com/khwilliamson> might have an opinion here.
> 

I missed that detail.  They should remain inline

embed.fnc
demerphq: That is not quite correct. I change some file level inline functions to be available to the regex engine. Calling them "global" is misleading. They are still only visible in the regex engine.

If converting them from inline is a problem i will move them. Thanks for the review.

demerphq: Wont moving them to invlist_inline.h make them API implicitly? Not what we mark as 'A', but effectively there will be no protections for loading them IIUIR, as all the headers will be available to any module during build right? Anyway, ill make a stab at it and see how it works out.

demerphq: ohisee! @khwilliamson actually exercised discipline when he set up invlist_inline.h and it includes define guards for the files which use it. Nice. That is a bit different than the other inline files I have encounterd. Well done. ++

demerphq: K, that should all be sorted now. Sorry about that.

tonycoz: > Calling them "global" is misleading

I struggled with the wording a bit.  They (were) global to the perl binary, but not API.  Perhaps I should have just said "no longer inline".

demerphq: Ah well, never mind. I was just feeling defensive. However you worded it, you were correct that it should be done a different way.

PR #20507, opened 2022-11-13
SvPVutf8_nomg - assign string length to the len parameter

mbarbon:


jkeenan:
Can you provide a description ... perhaps something along the lines of "For consistency with other \#defines added in 1ef9039bccb?

In our GH continuous integration, one porting test is failing in a way that suggests you need to add your name to AUTHORS.  Can you do so?  (But don't worry if you can't; we'll do that for a first-timer.)  Thanks.

@khwilliamson can you take a look at this one?

tonycoz:
The only test for SvPVutf8_nomg() doesn't test the length is returned :(

khwilliamson:
Thank you for finding and fixing this bug

This commit should be merged as soon as the authors is resolved.  I would have done so myself, but the author's github profile doesn't include an email address.  So we need to how it should be handled.

demerphq:
On Mon, 14 Nov 2022, 00:45 James E Keenan, ***@***.***> wrote:

> Can you provide a description ... perhaps something along the lines of
> "For consistency with other #defines added in 1ef9039
> <https://github.com/Perl/perl5/commit/1ef9039bccbfe64f47f201b6cfb7d6d23e0b08a7>
> ?
>
> In our GH continuous integration, one porting test is failing in a way
> that suggests you need to add your name to AUTHORS. Can you do so? (But
> don't worry if you can't; we'll do that for a first-timer.) Thanks.
>
It would be better if Mattias does it. His tests should have failed locally
with instructions on what to do: run Porting/updateAUTHORS.pl and then
committ the results.

Thanks a lot for the bugfix Mattia. Hope life is treating you well, long
time no see.

Cheers,
Yves

demerphq (approved):

demerphq:
> The only test for SvPVutf8_nomg() doesn't test the length is returned :(

@tonycoz we can add a patch for that to ext/XS-APItest as a follow up patch? It feels like your comment deserves more follow up than we should expect from @mbarbon in this context. Also given @khwilliamson's feedback we shouldn't make this wait on more tests...

tonycoz:
> @tonycoz we can add a patch for that to ext/XS-APItest as a follow up patch?

That's fine.

demerphq:
Thanks @mbarbon, merged!

mbarbon:
> His tests should have failed locally

Indeed, I have been lazy... I just tested the change against the CPAN module that was failing

Thanks for the quick turnaround on the patch!

demerphq:
> I just tested the change against the CPAN module that was failing

I know the feeling. the porting tests trip people up often enough that we have a specific target for it: 
```
TEST_JOBS=8 make -j8 test_porting
``` 
is a good sanity check before you create any PR's. Porting tests tend to catch things other tests don't.

>Thanks for the quick turnaround on the patch!

My pleasure. :-)

PR #20512, opened 2022-11-14
Add `PL_infix_plugin`, a mechanism for pluggable infix operators

leonerd:
This set of changes adds a new variable `PL_infix_plugin` to support a new pluggable infix operator mechanism. In many ways it is similar to the existing `PL_keyword_plugin`, however by necessity has to operate a little differently due to its much tighter integration with the parser, particularly in view of how operator precedence has to work.

Points of discussion to consider:

 * Is `perl.h` the right place to store the struct + enum definitions?
 * Do more precedence levels need considering? It's hard to imagine too many more of them being useful without having actual examples to write for them. They'd be easy enough to add later on so maybe this is fine.

(It also currently needs squashing down to a single commit)


abraxxa:
If the implementation of [XS::Parse::Infix](https://metacpan.org/pod/XS::Parse::Infix) is tightly tied to the core implementation and is the de-facto API for using it, it should be shipped with core.

leonerd:
@abraxxa 
> If the implementation of [XS::Parse::Infix](https://metacpan.org/pod/XS::Parse::Infix) is tightly tied to the core implementation and is the de-facto API for using it, it should be shipped with core.

Perhaps so. Slightly complicated by the fact that XS::Parse::Infix is one module in the distribution of XS-Parse-Keyword; being a more general API for doing keyword parsing. Would the whole thing move?

Part of the reason I like it *not* being core is that means I have a much faster iteration veloctiy on it. I can edit and change things much faster than I could if it was core. Moving it into core would remove that ability.

abraxxa:
Dual-life then.
But if it‘s tightly coupled keeping backcompat with older Perl versions would be added burden for the maintainer (you ;-)).

Grinnz:
Dual life modules have the same burden of back compat with older Perl versions.

leonerd:
> But if it‘s tightly coupled keeping backcompat with older Perl versions would be added burden for the maintainer (you ;-)).

"back-compat with older Perl versions" is literally not a problem here. This won't work *at all* on any perl that predates PL_infix_plugin, so that is all fine. I'd have the same compat problems whether it was CPAN-only or dual CPAN/core.

The fact remains that by *not* putting it in core I can iterate on it hundreds of times faster.

abraxxa:
My backcompat comment was regarding future Perl versions.

Not shipping something with core thats required to use a new core feature doesn‘t sound right.

Are there modules that are only shipped with Perl and not dual-life on CPAN? Where their version needs to match Perls’ version to make them work?

leonerd:
> Not shipping something with core thats required to use a new core feature doesn‘t sound right.

Thing is, this isn't the typical kind of "write some special syntax in your main.pl or -E" sort of feature, or even anything that could be visible from pure perl at all. You have to write an XS module to make use of it. Marking that XS module as depending on another XS module is not really an inconvenience.

> Are there modules that are only shipped with Perl and not dual-life on CPAN? Where their version needs to match Perls’ version to make them work?

Yeah there's loads. Mostly in `lib/` itself. They rarely ever change though. Whereas, this module is still being actively developed in terms of new features added, existing features changed. I *could* just copy a static snapshot of it into core perl but it would become outdated almost immediately and anyone wanting to actually write an infix operator module would want to grab the latest of CPAN rather than use their old out-of-date one from core anyway.

This is what I mean about "development velocity". I'm just not sure I see the point *yet* in adding that XS module to core, when it's still going to keep being developed so much.

Leont:
> This is what I mean about "development velocity". I'm just not sure I see the point yet in adding that XS module to core, when it's still going to keep being developed so much.

I can see the point while this is an experimental feature, but do think it should be included in core before it's declared stable.

leonerd:
> I can see the point while this is an experimental feature, but do think it should be included in core before it's declared stable.

Most definitely. It's what I said above in the PR description:

> We'd likely want to consider moving that XS module into core, or at least providing something similar from core itself once the overall idea seems a bit more stable.

demerphq:
> This is what I mean about "development velocity". I'm just not sure I see the point yet in adding that XS module to core, when it's still going to keep being developed so much.

FWIW, it isn't that clear why having the required module be in dist/ or cpan/ would have much recurring impact on your development velocity. Seems to me that you don't have to make the Perl version leading if you don't want to, if you put it in cpan/ then the published version on CPAN becomes leading and you don't have to update it at all, the RM will. Am I missing something about this?

leonerd:
> FWIW, it isn't that clear why having the required module be in dist/ or cpan/ would have much recurring impact on your development velocity. Seems to me that you don't have to make the Perl version leading if you don't want to, if you put it in cpan/ then the published version on CPAN becomes leading and you don't have to update it at all, the RM will. Am I missing something about this?

Eh; I mean sure I could put it there - I've no idea how the import process works so that's a practical question that needs answering as well.

But it's also the fact that in practice anyone would just pull the CPAN version anyway so I didn't really see a need to drop a stale old "you shouldn't use this" copy in core. But if everyone wants it there then fine, I can pop one in.

Any advice on how?

demerphq:
On Tue, 15 Nov 2022, 16:09 Paul Evans, ***@***.***> wrote:

> FWIW, it isn't that clear why having the required module be in dist/ or
> cpan/ would have much recurring impact on your development velocity. Seems
> to me that you don't have to make the Perl version leading if you don't
> want to, if you put it in cpan/ then the published version on CPAN becomes
> leading and you don't have to update it at all, the RM will. Am I missing
> something about this?
>
> Eh; I mean sure I could put it there - I've no idea how the import process
> works so that's a practical question that needs answering as well.
>
> But it's also the fact that in practice anyone would just pull the CPAN
> version anyway so I didn't really see a need to drop a stale old "you
> shouldn't use this" copy in core. But if everyone wants it there then fine,
> I can pop one in.
>
> Any advice on how?
>

If you publish it on cpan I'll take the time to import it.

Yves

>

leonerd:
> If you publish it on cpan I'll take the time to import it.

https://metacpan.org/dist/XS-Parse-Keyword

jkeenan:
> On Tue, 15 Nov 2022, 16:09 Paul Evans, ***@***.***> wrote: FWIW, it isn't that clear why having the required module be in dist/ or cpan/ would have much recurring impact on your development velocity. Seems to me that you don't have to make the Perl version leading if you don't want to, if you put it in cpan/ then the published version on CPAN becomes leading and you don't have to update it at all, the RM will. Am I missing something about this? Eh; I mean sure I could put it there - I've no idea how the import process works so that's a practical question that needs answering as well. But it's also the fact that in practice anyone would just pull the CPAN version anyway so I didn't really see a need to drop a stale old "you shouldn't use this" copy in core. But if everyone wants it there then fine, I can pop one in. Any advice on how?
> If you publish it on cpan I'll take the time to import it. Yves
> […](#)

I have to admit that I'm finding this discussion confusing.  Over the past ten years our general tendency has been to try to keep the Perl 5 core distribution from expanding unnecessarily.  I cannot recall the last time we decided to begin including a CPAN-maintained library in the core distribution (this could be researched) but I know we haven't done it that often.  The rationale for doing so now seems murky, and I think bringing a new CPAN distro into core falls into the range of policy decisions that should be discussed on list first.

demerphq:
On Tue, 15 Nov 2022, 19:52 James E Keenan, ***@***.***> wrote:

> On Tue, 15 Nov 2022, 16:09 Paul Evans, *@*.***> wrote: FWIW, it isn't
> that clear why having the required module be in dist/ or cpan/ would have
> much recurring impact on your development velocity. Seems to me that you
> don't have to make the Perl version leading if you don't want to, if you
> put it in cpan/ then the published version on CPAN becomes leading and you
> don't have to update it at all, the RM will. Am I missing something about
> this? Eh; I mean sure I could put it there - I've no idea how the import
> process works so that's a practical question that needs answering as well.
> But it's also the fact that in practice anyone would just pull the CPAN
> version anyway so I didn't really see a need to drop a stale old "you
> shouldn't use this" copy in core. But if everyone wants it there then fine,
> I can pop one in. Any advice on how?
> If you publish it on cpan I'll take the time to import it. Yves
> … <#m_-7331375143909927819_m_8730965078696932106_>
>
> I have to admit that I'm finding this discussion confusing. Over the past
> ten years our general tendency has been to try to keep the Perl 5 core
> distribution from expanding unnecessarily. I cannot recall the last time we
> decided to begin including a CPAN-maintained library in the core
> distribution (this could be researched) but I know we haven't done it that
> often. The rationale for doing so now seems murky, and I think bringing a
> new CPAN distro into core falls into the range of policy decisions that
> should be discussed on list first.
>

I think you are mischaracterizing the situation. Paul has said he wants to
add a feature to toke.c to support pluggable infix operators. He wants to
create an api around this feature which would live in an XS module. Paul
wants this module to be the one and only api to the core feature. Given the
problem domain this is not unreasonable.

Normally a core dev in this situation would put that logic either in
universal.c and export into the namespace or he would create a new
distribution in ext/.  However in this case Paul wants to continue
developing that module and api independently from the perl release cycle
(and by implication he is committing to back support for new features once
this is released). Essentially he wants to provide a core api with extra
service.

Thus it is NOT analogous to "bring a module into core from CPAN",  this
would be more akin to us releasing the regex engine independently from
core. If we were to do so we would move it from ext/ to dist/ or maybe even
cpan/.

Since the feature in core is designed to work specifically with this one
wrapper it makes sense that wrapper is maintained either in core entirely
or in some kind of dual life way. Thus cpan/ is the natural location for it.

To me the only controversial thing in this thread was the suggestion of
adding a feature to core that could only be accessed via a module *only*
distributed on cpan. How would we (core devs) even test things if we needed
to change toke.c?

Note, we aren't talking about importing every gee-whiz new package that
Paul has built on top of the api, just the api itself.

I hope that clears this up for you. I really don't think this is analogous
to cases I think you have in mind. I generally am not in favor of adding
cpan stuff to core, but core api's should definitely be bundled with core.

Yves

abraxxa:
Thanks Yves for wording it so perfect! That‘s what I tried to say.

leonerd:
Another observation: `XS::Parse::Infix` uses `Module::Build` so it wouldn't be easy to just throw that into core as it stands. It *might* be possible to adjust it around that but offhand I don't know how to make EU:MM do half the things I regularly make M:B do, so I don't think I'm able to do that. Plus, the point of `XS::Parse::Infix::Builder` is to support -other- modules implementing infix operators that use M:B in their build system. Again I'm not sure even if EU:MM can do the same sort of thing; and if so, how. It would be quite the engineering effort to get it that far.


So, what's the plan here? I think in terms of the docs + interface module, there's three options on the table:

## 0. Merge the PR as it is

Just accept this PR as it stands.

## 1. Add documentation

Add docs on how the internals work; even though I will strongly discourage anyone else trying to implement it, because the central registry of operators is kindof the entire point of the XS::Parse::Infix module. Most of the advantages of this approach disappear as soon as a second module tries to do what XS::Parse::Infix itself is doing (and the same goes for X:P:Keyword). My intention here is very very much that other XS module authors who want to implement real infix operators (or keywords) do so via XPI and XPK, not via the perl core interpreter interfaces directly.

## 2. Write a new in-core registry module

This would be a lot more effort and basically involve duplicating most of the code out of XPI and XPK again, into core. Having done so the new core-only mechanism would only get updated once a year and therefore have much slower development velocity than my CPAN modules.

## Any other thoughts?

Something else I might have missed?


Also, aside from this question on the docs and interface module, does anyone want to actually review the *actual* code? ;)

leonerd:
My opinions here: I strongly favour option 0. Neither options 1 or 2 seem overly appealing for reasons I've (hopefully) laid out above, and just seem to be adding more work for no real benefit. There's still a possibility I missed some other option though. 
Thoughts welcome.

leonerd:
Actually, another idea that's a sortof variant of option 2

## 3. Implement the Registry in Perl Core

I guess what might work is merging this PR now, and then recreating the same API shape that lives in src/infix.c, natively in perl core... So actually XPI would just be a tiny trampoline onto that and then core itself would implement the registry and various bits and pieces.

demerphq:
Basically i am fine with 0 provided it is followed quickly by a 2/3. The point about how often it gets updated is IMO not really relevant if the code lives in cpan/, people can just update locally if they want (presumably peoples dependencies will do that), and more importantly the core dev can test that their changes arent breaking the "standard api". 

For the short term very early experimental stage however I personally can live with merging this as is.

I also dont mind putting some effort into converting the api to MakeMaker to help you get this moving.

leonerd:
New plan: It seems like it should be possible to reïmplement the registry parts of XPI in actual perl core without too much trouble and then *that* becomes the interface that XPI uses to talk to perl core. That can be documented a bit more in that case.

I can probably write this up in the next few days anyway, so for now I'll revert this PR to a draft.

It sounds like everyone is broadly happy with the concept of pluggable infix operators and how the parsing is done on them, so that part seems stable.

leonerd:
On a further stare, it wasn't easily possible to extract the central registry part without basically moving over most of the still-experimental inner parts of XPI's design, which would have made further development much more difficult. Instead, I've gone with writing some documentation on this particular variable. In particular, it notes the temporary nature of this arrangement, in the hope of eventually moving the central registry into core once the overall API shape becomes more stable.

perlvars.h
```diff
@@ -235,10 +235,107 @@ Instead, use the function L</wrap_keyword_plugin>.
...
+discovery/reflection, and these abilities only work because it knows all the
+registered operators. Other modules should not use this interpreter variable
+directly to implement them because then those central features would no longer
+work properly.
+
+Furthermore, it is likely that this (experimental) API will be replaced in a
+future Perl version by a more complete API that fully implements the central
+registry and other semantics currently provided by C<XS::Parse::Infix>, once
+the module has had sufficient experimental testing time. This current
+mechanism exists only as an interrim measure to get to that stage.
```
jkeenan: `s/interrim/interim/`

leonerd: Ahyes, thanks.

demerphq:
@leonerd if you gotta do it that way then do it that way.

rjbs:
I promised @leonerd that I would chime in, and I am finally doing so.  I hope it is not anti-climactic when I just say, "This all sounds like a good plan to me, and I look forward to it.

PR #20513, opened 2022-11-14
Yves/implicit hek key

demerphq:
This is a POC for removing the char[0] style members from our structs. (It currently includes another commit i will push in a separate PR).

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -69,12 +69,21 @@
 : macros can't be generated in such situations.
 :
 : WARNING: Any macro created in a header file is visible to XS code, unless
-: care is taken to wrap it within something like #ifdef PERL_CORE..#endif.
-: This has had to be done with things like MAX and MIN, but nearly everything
-: else has been created without regard to the namespace pollution problem.
+: care is taken to wrap it within something like #ifdef PERL_CORE..#endif or
+: more commonly PERL_IN_FILE_C (with FILE_C appropriately replaced).  Most (if
+: not all) C files define such a symbol before importing perl.h.  In general you
+: should restrict the exposure of your exports as much as possible, although
+: older code may not do so.  Be aware that non-static exports can be "over
+: exported" and things work out fine, but inline and static macros will cause
+: errors unless restricted to the specific file they are intended for, and the
+: generated PERL_ARGS_ macros will only be available to inline functions in the
+: appropriate context.  This why you will see sections listing each file in
+: turn, and is also why the file favours use of '#if defined()' form of
+: conditions instead of '#ifdef' as the latter only works with one symbol and
+: the former can be combined.  It is also why you will often see functions with
+: an 's' or 'i' export type grouped together into a single conditional block
+: separate from the functions from the same file with 'p' in them.
 :
-: Here's what else you need to know about using this file with regards to name
-: space pollution:
 :
 : The A flag is used to make a function and its short name visible everywhere
 :	     on all platforms.  This should be used to make it part of Perl's
--- a/hv.c
+++ b/hv.c
@@ -3379,8 +3379,7 @@ S_share_hek_flags(pTHX_ const char *str, STRLEN len, U32 hash, int flags)
            HE directly from the HEK.
         */
 
-        Newx(k, STRUCT_OFFSET(struct shared_he,
-                                shared_he_hek.hek_key[0]) + len + 2, char);
+        Newx(k, sizeof(struct shared_he) + len + 2, char);
         new_entry = (struct shared_he *)k;
         entry = &(new_entry->shared_he_he);
         hek = &(new_entry->shared_he_hek);
--- a/hv.h
+++ b/hv.h
@@ -57,7 +57,8 @@ struct hek {
      * dont work on aligned platforms like HPUX
      * Also beware, the last byte of the hek_key buffer is a
      * hidden flags byte about the key. */
-     char       hek_key[1];        /* variable-length hash key */
+     /* IMPLIED:
+     char       hek_key[1];        *//* variable-length hash key */
     /* the hash-key is \0-terminated */
     /* after the \0 there is a byte for flags, such as whether the key
        is UTF-8 or WAS-UTF-8, or an SV */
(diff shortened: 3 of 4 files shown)


jkeenan (commented):
In commit message, `"... which is the overallocated to provide ..."`, something needs to be added (or perhaps deleted) there to make sense.

Also, should it be `bounds checking` rather than `bound checking`?

tonycoz:
I think it's practically safe in this particular case, since char has no alignment requirements, but in the general case it may leave the array after the fixed members unaligned, ie.

```
struct foo {
   char a;
   char b;
  /* implicit double[] */
};
...
  struct foo * f = malloc(sizeof(struct foo) + n * sizeof(double));
  double *d = (double *)(f + 1);
  *d; # unaligned
```
so that final actual declaration of the array at the end ensures that accesses to the array are aligned.

WRT standard C (or C++) I think with the change the accesses have undefined behaviour, but that's true of the trailing `Type foo[1];` [member](https://www.open-std.org/Jtc1/sc22/wg14/www/docs/dr_051.html) [too](https://wiki.sei.cmu.edu/confluence/display/c/DCL38-C.+Use+the+correct+syntax+when+declaring+a+flexible+array+member).

I do tend to think the trailing `Type foo[1]` is safer though - we're accessing the elements of the array through an expression of that type, which may avoid (further) aliasing issues.

I wish we could use the C99 flexible array members (FAM), but these types are typically in headers which also need to compile as C++, and they aren't valid C++.

gcc and clang allow FAM in C++ mode, but these headers need to compile with C++ XS modules, so they need work with every C++ compiler.

khwilliamson:
On 11/14/22 22:25, Tony Cook wrote:
> I think it's practically safe in this particular case, since char has no 
> alignment requirements, but in the general case it may leave the array 
> after the fixed members unaligned, ie.
> 
> |struct foo { char a; char b; /* implicit double[] */ }; ... struct foo 
> * f = malloc(sizeof(struct foo) + n * sizeof(double)); double *d = 
> (double *)(f + 1); *d; # unaligned |
> 
> so that final actual declaration of the array at the end ensures that 
> accesses to the array are aligned.
> 
> WRT standard C (or C++) I think with the change the accesses have 
> undefined behaviour, but that's true of the trailing |Type foo[1];| 
> member <https://www.open-std.org/Jtc1/sc22/wg14/www/docs/dr_051.html> 
> too 
> <https://wiki.sei.cmu.edu/confluence/display/c/DCL38-C.+Use+the+correct+syntax+when+declaring+a+flexible+array+member>.
> 
> I do tend to think the trailing |Type foo[1]| is safer though - we're 
> accessing the elements of the array through an expression of that type, 
> which may avoid (further) aliasing issues.
> 
> I wish we could use the C99 flexible array members (FAM), but these 
> types are typically in headers which also need to compile as C++, and 
> they aren't valid C++.
> 
> gcc and clang allow FAM in C++ mode, but these headers need to compile 
> with C++ XS modules, so they need work with every C++ compiler.

Why couldn't we define a symbol, say PERL_FAM to be empty on C, and 1 on 
C++, and change our declarations to be

foo[PERL_FAM]

jkeenan:
Is this p.r. intended to reduce or eliminate build-time warnings (*e.g.,* those discussed in https://github.com/Perl/perl5/issues/20410)?  If so, my own testing suggests that that is not being achieved yet.

demerphq:
On Tue, 15 Nov 2022, 20:44 James E Keenan, ***@***.***> wrote:

> Is this p.r. intended to reduce or eliminate build-time warnings (*e.g.,*
> those discussed in #20410 <https://github.com/Perl/perl5/issues/20410>)?
> If so, my own testing suggests that that is not being achieved yet.
>

It is a POC to demonstrate how one class of warning you are seeing might be
tackled. There are about 5 or 6 datastructures that do this, distributed
across 4 or 5 files. This was just the one that was referenced in leonerds
dupe bug report, and likely one of the more common cases, but not a
complete fix by any means. Do you see warnings about struct rcpv for
instance?

Yves

>

jkeenan:
> On Tue, 15 Nov 2022, 20:44 James E Keenan, ***@***.***> wrote: Is this p.r. intended to reduce or eliminate build-time warnings (*e.g.,* those discussed in #20410 <#20410>)? If so, my own testing suggests that that is not being achieved yet.
> It is a POC to demonstrate how one class of warning you are seeing might be tackled. There are about 5 or 6 datastructures that do this, distributed across 4 or 5 files. This was just the one that was referenced in leonerds dupe bug report, and likely one of the more common cases, but not a complete fix by any means. Do you see warnings about struct rcpv for instance? Yves
> […](#)

```
$ ~/bin/perl/report-build-warnings a2682da36c.freebsd.threaded.gcc12.hek.maketp.output.txt.gz
File:  a2682da36c.freebsd.threaded.gcc12.hek.maketp.output.txt.gz

  Waddress                                 104
  Warray-bounds                              3
  Wstringop-overflow=                        2
```
... but, no, I don't see any mention of the string `rcpv` in the `make test_prep` output.

demerphq:
On Tue, 15 Nov 2022 at 21:58, James E Keenan ***@***.***>
wrote:

>
>   Waddress                                 104
>
>   Warray-bounds                              3
>
>   Wstringop-overflow=                        2
>
>
What were those numbers /before/ this patch?



> ... but, no, I don't see any mention of the string rcpv in the make test_prep output.
>
>
Ok, interesting.  It would be nice if you could assemble the files/lines
where there is something like: char str[1];  in the error output.

cheers,
Yves
--

jkeenan:
> Ok, interesting. It would be nice if you could assemble the files/lines where there is something like: char str[1]; in the error output. 

This is as close as I get to that:
```
$ ~/bin/perl/parse-one-warnings-group  a2682da36c.freebsd.threaded.gcc12.hek.maketp.output.txt.gz 'Warray-bounds'
File:  a2682da36c.freebsd.threaded.gcc12.hek.maketp.output.txt.gz

[
  {
    char   => 26,
    group  => "Warray-bounds",
    line   => 441,
    source => "hv.h",
    text   => "array subscript [0, 9223372036854775807] is outside array bounds of 'unsigned char[0]'",
  },
  {
    char   => 26,
    group  => "Warray-bounds",
    line   => 441,
    source => "hv.h",
    text   => "array subscript [0, 9223372036854775807] is outside array bounds of 'unsigned char[0]'",
  },
  {
    char   => 26,
    group  => "Warray-bounds",
    line   => 441,
    source => "hv.h",
    text   => "array subscript [0, 9223372036854775807] is outside array bounds of 'unsigned char[0]'",
  },
]
```

tonycoz:
Is this still needed/desirable?

demerphq:
@tonycoz - nope. thanks to your efforts elsewhere. this was just a POC anyway, closing.

PR #20514, opened 2022-11-15
clarify that every tag is required to be annotated

andk:
- while the instructions were already clear at this point, accidents with lightweight (unannotated) tags have happened


Leont (approved):

jkeenan (approved):
+1

PR #20516, opened 2022-11-15
embed.fnc: Enhance docs, normalize and group guard clauses and sort function lists

demerphq:
Introduce a new standardized module for regen code that needs to parse header file content.

This is then used to implement embed.pl, embed_lib.pl, a tidy program for embed.fnc called regen/tidy_embed.pl, regen/regcomp.pl and regen/features.pl. 

It should produce more stable and more consistent output, and it has some neat features to tidy and handle preprocessor content properly. It handles 'elif' in embed.fnc properly, and can actually reconstitute an if/elif/elif/else chain from constituent parts if they are provided. @khwilliamson complained about the old code not supporting this in a comment in embed.fnc.

It does slow down make regen somewhat, it takes a few seconds with this PR. I wasn't really optimizing for speed considering make regen is a manual action only required when specific changes are made. If we really care I'm sure the normalization can be optimized more, I just didn't see the point. The other point is there are a bunch of further changes that could be made to improve the code even more, switching to using methods instead of hash lookups. I would like to leave that for future patches as every time someone changes embed.fnc this PR conflicts and has to be manually recreated.  Once it is merged that wont be an issue anymore.  

Reviewers Note: it may be helpful to start by reviewing the new test file t/porting/header_parser.t first. Its tests show what this PR is trying to achieve in the simplest form.


demerphq:
I have not heard any feedback on this, and it requires manual updates every time someone pushes an embed.fnc change. I will merge soon unless I hear someone object. I checked with @iabyn and merging this shouldnt cause him too much trouble in his refcount branch. It might cause some trouble to other large changes, but it shouldn't be difficult to resolve.  Long term this patch is intended to eliminate as much as possible future such issues, or at the least make them easier to resolve, when embed.fnc gets changed in the future.  @leonerd I know you have a large patch pending, would you prefer I wait until you merge? Or do you think you can handle the turbulence? @leont ? @khwilliamson ? @tonycoz ?

If there is clear feedback that i should wait I dont mind, i dont want to ruin anyones plans, but i would like to get this merged.

leonerd:
Don't worry too much about my change. I'm just adding a few functions to the list; I can add them around whatever other change happens. I've been dragging that branch along for four months now around lots of conflicts in embed.fnc, perly.y, various files in regen/, ... I've got quite used to it ;)

demerphq:
Ok, I just wanted to say im find with merging and then tweaking the process after if necessary and if you have  bad user experience.

embed.fnc
```diff
@@ -69,50 +69,58 @@
...
+: older code may not do so.  Be aware that non-static exports can be "over
+: exported" and things work out fine, but inline and static macros will cause
+: errors unless restricted to the specific file they are intended for, and the
+: generated PERL_ARGS_ macros will only be available to inline functions in the
+: appropriate context.  This why you will see sections listing each file in
+: turn, and is also why the file favours use of '#if defined()' form of
+: conditions instead of '#ifdef' as the latter only works with one symbol and
+: the former can be combined.  It is also why you will often see functions with
+: an 's' or 'i' export type grouped together into a single conditional block
+: separate from the functions from the same file with 'p' in them.
```
tonycoz: I think this would be more readable in point form.

regen/HeaderParser.pm
```diff
@@ -0,0 +1,1556 @@
...
+        @cmpop,    # include the numerical comparison operators.
+        '<< >>',
+        '+ -',
+        '* / %',    # highest prcedence operators.
+    );
+
+    my @unop= qw( ! ~ + - );
+    %unop= map  { $_ => 1 } @unop;
+    %cmpop= map { $_ => 1 } map { split /\s+/, $_ } @cmpop;
+    %binop= map { $_ => 1 } map { split /\s+/, $_ } @binop;
```
tonycoz, line 78: Minor nit, repeated in many places:  I like a space before the '=' and I think we mostly do that.

regen/embed_lib.pl
```diff
@@ -1,164 +1,98 @@
...
     {
         my %syms;
 
-        while (<IN>) {
-            chomp;
-            next unless $_;
-            next if /^#/;
-            my $check = (split /\t+/, $_)[2];
+        my $line_num = 0;
+        while (defined(my $line= <$in_fh>)) {
```
tonycoz: defined() is unneeded.

tonycoz (approved):
I've marked some minor issues, but I think it's fine to merge.

I haven't been able to do a deep look into it.

regen/HeaderParser.pm
demerphq: This comes up every now and then. I left hug assignment operators for my own good reasons. I include perltidy rules in the docs of my files for following my style. We don't have an official style I can use to format other folks code, but I provide one so folks can use it to format changes to my code. I also "follow the style of the file" when I am modifying existing files. I think that is a reasonable compromise.  

FWIW, I would welcome us adopting a standard perltidy ruleset for our test files and porting code and the like. But until then I would prefer that files I author and will maintain be in the style I prefer to use and in which I created them. Call it my artistic license. (Although I have good technical reasons for my preferences. I *never* mistakenly use assignment instead of equality operators.) I don't think that is unreasonable, especially as I would be happy to switch to the "approved" perltidy ruleset if we could agree on it. (There are tons of files I *long* to perltidy.)

embed.fnc
demerphq: Will do.

regen/embed_lib.pl
demerphq: I know, but I prefer to be explicit, and it doesn't hurt anything. (usually I write readline actually).

demerphq:
@tonycoz - thanks for your feedback, patches coming up. merry christmas!

regen/HeaderParser.pm
Grinnz: If there's any standard to follow for code format it would be https://perldoc.perl.org/perlstyle

demerphq: Until we agree on a set of official coding standards such that I can run perltidy with those rules on any file I modify (to fix my mistakes) then I will continue to follow the tradition of the project, which is to respect that others may have differing style choices and to follow the style of the file, granting the author of the file the choice of its style.

Again, I would vote in favor of any sane set of perltidy rules to mechanize this process. FWIW, I feel some of the guidance on the perltstyle doc is controversial and at least shouldn't be part of a style guide /unless/ it can be machine assisted.

I am in fact in favor of any machine assisted style management, indeed this patch is intended to introduce machine assisted tidying to embed.fnc itself. :-) So long as I can run a tool to fix things I dont care what we use, until then I'd prefer the things I author and intend to maintain follow my preferences. At least when I work on those I don't have to worry about my style, any more than anybody else does, as I just run my perltidy rules afterwards.

regen/embed_lib.pl
demerphq: Oh, wait, that was embed_lib.pl, yeah, ill revert that. Its a nervous habit adding the explicit defined. but there was no good purpose to do so here.

demerphq: Removed.

embed.fnc
demerphq: I broke it out into paragraphs and made it breathe a bit. Check the latest push.

PR #20517, opened 2022-11-16
ParseXS - Disable alias value collision warnings by default

demerphq:
We always had a bug with the value 0, when fixing it, we discovered a lot of people were doing duplicate aliases with 0. This patch disables the warnings except when in a special dev mode, as it makes no sense to show warnings like thus to end users, its not a system issue its a code issue, and only the XS files author can fix it. It would seem most times its not an issue and is entirely deliberate. So showing such warnings only on request also makes sense.

### DIFF
--- a/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
+++ b/dist/ExtUtils-ParseXS/lib/ExtUtils/ParseXS.pm
@@ -11,7 +11,7 @@ use Symbol;
 
 our $VERSION;
 BEGIN {
-  $VERSION = '3.48';
+  $VERSION = '3.49';
   require ExtUtils::ParseXS::Constants; ExtUtils::ParseXS::Constants->VERSION($VERSION);
   require ExtUtils::ParseXS::CountLines; ExtUtils::ParseXS::CountLines->VERSION($VERSION);
   require ExtUtils::ParseXS::Utilities; ExtUtils::ParseXS::Utilities->VERSION($VERSION);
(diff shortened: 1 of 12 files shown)


demerphq:
Those last pushes include tests and a fix for perl 5.8

jkeenan:
> We always had a bug with the value 0, when fixing it, we discovered a lot of people were doing duplicate aliases with 0. 

Could you be a bit more specific in this commit message, *e.g.,* a reference to a bug ticket or a user complaint?  How far back in time does *always* cover?

> This patch disables the warnings except when in a special dev mode, as it makes no sense to show warnings like thus to end users, its not a system issue its a code issue, and only the XS files author can fix it. It would seem most times its not an issue and is entirely deliberate. So showing such warnings only on request also makes sense.

jkeenan:
> Those last pushes include tests and a fix for perl 5.8

What about perl-5.8 needed to be fixed?

jkeenan:
I can confirm that this does turn off the warnings that begin with `Warning:`:
```
$ zgrep -c 'Warning:' 40db0ddf69.freebsd.threaded.clang10.dev_mode_parsexs.maketp.output.txt.gz
0
```
There were 15 such warnings being generated in blead at eae8123147.

demerphq:
> How far back in time does always cover

I don't specifically. I haven't checked. It wouldn't surprise me if it either dates back to Larrys code on xubpp, or the refactor that produced ParseXS, or alternatively I would not be surprised if the warning was added after the fact, and was misimplemented when it was.

> What about perl-5.8 needed to be fixed?

In the initial push of this patch I used `//` (defined-or), which isnt supported in 5.8. Nothing big.

demerphq:
> e.g., a reference to a bug ticket or a user complaint? 

I said before I noticed it while I working on silencing the non-buggy part of the warning. 

> How far back in time does always cover?

Long enough that we have multiple files in core that exploit the bug. See https://github.com/Perl/perl5/pull/20496#issuecomment-1316671123

demerphq:
I checked, it is present in the original version of ParseXS from 2005:

commit 6b09c1601036c61459334bdedef5d7e29e07fcaf
Author: Yitzchak Scott-Thoennes <sthoenna@efn.org>
Date:   Wed May 18 00:13:40 2005 -0700

demerphq:
@tonycoz any thoughts on this? id like to merge tomorrow.

demerphq:
Im going ahead with this. I dont think it makes anything worse.

PR #20520, opened 2022-11-16
test SvPVbyte*() and SvPVutf8*() variants set len

tonycoz:
These macros have limited use in core:

- only Socket uses SvPVbyte_nomg()
- nothing uses SvPVutf8_nomg()

Update the existing tests to ensure len is set properly.

Related to #20507


demerphq (approved):

demerphq:
Just a thought but a single function using aliases would probably be less hassle to maintain.  Bit late now, but maybe next time.

PR #20526, opened 2022-11-19
Add default-if-undef and default-if-false signature parameters

leonerd:
Adds syntax using the `//=` and `||=` operators to create signature parameters with default expressions that apply whenever the value is missing, undefined or false.

```perl
use v5.37;

sub greet($whom //= "world") {
   say "Hello, $whom";
}

greet();       # Prints "Hello, world"
greet(undef);  # same
```

Discussed on perl5-porters: https://www.nntp.perl.org/group/perl.perl5.porters/2022/11/msg265077.html


leonerd:
As this change adds only syntax that was previously a compiletime error on older perls, I didn't feel it necessary to add an "experimental" warning.

```
$ perl5.36.0 -E 'sub greet($whom //= "world") { say "Hello, $whom" }  greet(undef);'
Illegal operator following parameter in a subroutine signature at -e line 1, near "($whom //= "world""
syntax error at -e line 1, near "($whom //= "world""
```

lib/B/Op_private.pm
```diff
@@ -600,6 +600,8 @@ our %defines = (
     OPpARGELEM_HV            =>   4,
     OPpARGELEM_MASK          =>   6,
     OPpARGELEM_SV            =>   0,
+    OPpARG_IF_FALSE          =>  64,
```
jkeenan, line 603: Is it okay for `OPpARG_IF_FALSE` to be assigned the same value as `OPpASSIGN_BACKWARDS` a few lines below?

leonerd: Yeah this is fine. `OPpARG_IF_FALSE` is for `OP_ARGDEFELEM` whereas `OPpASSIGN_BACKWARDS` is for `OP_SASSIGN`.

The `OPp*` flags are for the `op_private` field, which is only 8 bits wide. The meanings of the bits are "private" to each individual optype. As such there's many many such overlaps and reuses of bits across the different ops. In fact, the entire `regen/op_private` file is used to manage them all and ensure no actual individual op type has a collision.

pod/perldelta.pod
```diff
@@ -27,6 +27,14 @@ here, but most should go in the L</Performance Enhancements> section.
 
 [ List each enhancement as a =head2 entry ]
 
+=head2 Defined-or and truth-or default expressions to signatures
```
JRaspass: perlop calls the operator `logical-or` not `truth-or`, but also I feel the word assignment ought to be in there to differentiate the `defined-or //` operator from the `defined-or assignment operator //=`

leonerd: A good idea. Reworded and force-pushed

iabyn:
LGTM. I don't understand what's been done in the parser and toker (because I didn't take the time to understand), but nothing leapt out at me that looked wrong.

leonerd:
Thanks @iabyn 
In brief: the `sigdefelem` rule used to yield an optree of just its expression fragment and have the higher-up rule wrap that in the `OP_ARGDEFELEM`. But now we need to remember if it was `=`, `//=` or `||=`. So I changed it to have that rule create the entire `OP_ARGDEFELEM` part of the tree, where it can set the private flags. This means less work for the higher-up rule to perform. But because that rule had two branches I decided to make a common function for them both to use.

PR #16, opened 2018-07-16
skip shadow call when euid > 0 on linux

atoomic:
maybe __linux__ is not restrictive enough
and we could consider using a hint sh file
to enable it only on some specific distro?

PR #20530, opened 2022-11-19
Replaced SSGROW with SSCHECK inside regcppush (Win32 slowdown, recursive patterns)

AnFunctionArray:
Which drastically improves performance on large matches.

[donuts.zip](https://github.com/Perl/perl5/files/10048106/donuts.zip)

For the following file for example:

(On https://github.com/AnFunctionArray/cperllexer/commit/5980b216792cd79f88eaa289ef314295a0ebc4d8)

With the following env vars:

MINQUEUECOUNT=50
SILENT=1

```
Z:\cllvmbackend\lexer>powershell -Command "Measure-Command {perl ./parse.pl ./donuts.pp}"
joinning 30, 834632 for real


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 471
Ticks             : 4713919
TotalDays         : 5.45592476851852E-06
TotalHours        : 0.000130942194444444
TotalMinutes      : 0.00785653166666667
TotalSeconds      : 0.4713919
TotalMilliseconds : 471.3919




Z:\cllvmbackend\lexer>powershell -Command "Measure-Command {Z:\perl5orig\perl.exe ./parse.pl ./donuts.pp}"
joinning 30, 834632 for real


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 25
Milliseconds      : 489
Ticks             : 254896534
TotalDays         : 0.000295019136574074
TotalHours        : 0.00708045927777778
TotalMinutes      : 0.424827556666667
TotalSeconds      : 25.4896534
TotalMilliseconds : 25489.6534




Z:\cllvmbackend\lexer>
```

Where the second run is without the patch.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -261,7 +261,7 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
             );
     );
 
-    SSGROW(total_elems + REGCP_FRAME_ELEMS);
+    SSCHECK(total_elems + REGCP_FRAME_ELEMS);
 
     /* memcpy the offs inside the stack - it's faster than for loop */
     memcpy(&PL_savestack[PL_savestack_ix], rex->offs + parenfloor + 1, paren_bytes_to_push);


jkeenan:
The intent of this p.r. appears to be performance improvement. @iabyn, @demerphq can you take a look?  Thanks.

richardleach:
For anyone else reading this who isn't familiar with `SSGROW` and `SSCHECK`, the difference boils down to growing the save stack from a size of `PL_savestack_ix` to a size of:
SSGROW:  `PL_savestack_ix + need`
SSCHECK: `PL_savestack * 3/2 `

So if there is a need to push a lot of things to the save stack, `SSGROW` will trigger a lot of reallocations for tiny size increases, but (in theory) not have the stack allocating a lot more memory than is needed. In comparison, by resizing by much more at a time. `SSCHECK` will need to do fewer reallocations but could result in the save stack being allocated a good deal more memory (in the worst case) than it actually needs.

I'm not familiar enough with the save stack to know what the best balance is.

Both functions would seem to benefit from using `malloc_usable_size` etc. where the `malloc()` implementation supports it. For other cases, perhaps `SSGROW` could be somehow tuned to better appreciate the actual allocation sizes it's going to be given. (IIRC modern languages (e.g. Go) and browsers (certainly Firefox) often go for power-of-two allocations in general and find that works well for them.)

AnFunctionArray:
On a second thought it should probably not grow so much thus we are probably leaving some stack somewhere I could only assumr.

AnFunctionArray:
Maybe if we add a check on regex start end if the stack pointer is the same.

AnFunctionArray:
Ah no it's fine - it's my code is ultra recursive:

```
(?<globaloutter>(?<globalinner>(?&fasterdecls))*+((?&globalinner)(?&globaloutter))?+)
```
Not perl issue - nevertheless in that specific scenario it's faster with this change.

demerphq:
This patch makes sense to me. Speculatively overallocating seems pretty reasonable to me. The growth factor of 1.5 seems a touch high, id expect something closer to 1.25 or something like that. What i don't understand is why the patch doesn't change SSGROW and instead introduces a new macro. If pre-overallocating is a good idea (i think it is) then why not do it always?

AnFunctionArray:
@demerphq I have not introduced a new macro - the SSCHECK is an existing one.

demerphq:
@AnFunctionArray sorry, misunderstood. I see. FWIW, and totally as an aside, both of these macros should probably change or be replaced. The api doesnt allow us to check for overflow and wraparound as far as I can tell. Consider what happens when total_elems + REGCP_FRAME_ELEMS wraps to 0.

tonycoz:
I'd tend to go towards making both entry points use exponential growth.

I suspect this is is more an issue on Windows than on Linux - on Linux malloc() tends to allocate large allocations with mmap() and then use mremap() to extend them for realloc(), but I don't see any API support for that in the Windows API.

AnFunctionArray:
@tonycoz That oughta be correct - I mainly tested on windows with this one - on mac for example no such issues were had.

demerphq:
> I'd tend to go towards making both entry points use exponential growth.

I think exponential growth needs to be bounded in a case like this. But otherwise I agree.

Yves

tonycoz:
> I think exponential growth needs to be bounded in a case like this. But otherwise I agree.

Bounded how?

It's already bounded by the maximum size of the scope stack (32GB for 64-bit).

If it switches from exponential growth to linear growth, then the cost goes from $O(N \log N)$ to $O(N^2)$.

We already use exponential growth for PVs (in Perl_sv_grow) with a default of 1.25.

demerphq:
On Mon, 21 Nov 2022 at 23:19, Tony Cook ***@***.***> wrote:

> I think exponential growth needs to be bounded in a case like this. But
> otherwise I agree.
>
> Bounded how?
>
> It's already bounded by the maximum size of the scope stack (32GB for
> 64-bit).
>
> If it switches from exponential growth to linear growth, then the cost
> goes from $O(N \log N)$ to $O(N^2)$.
>
Yeah, I know the math.  Keep in mind that is an approximation of the most
extreme case as it grows to infinity, whereas in practice the operations
are always bounded (we have finite memory available), and the actual
scenario is much less costly than the worst case.

So for instance if we made it grow exponentially on a sliding scale,
eventually bottoming out at some (large) fixed chunk size then yes, the
very few programs out there that build 4GB  strings using naive
concatenation might see something that feels like N*2 growth, but most
people would never notice. Eg, consider if we used an exponent of 4 until
we 1MB, then used an exponent of 2 until we hit 32MB, and then an exponent
of 1.25 until we hit 100MB, and then proceeded in chunks of 100MB (or
multiples thereof if the stack needs to grow by that amount). How many
*real* programs would be affected? And how much memory would be saved on
average?  Would it be faster for most programs?

I'd argue that choosing an optimization that is potentially extremely
wasteful for many users to benefit the extreme few who actually possess
alternative ways to solve the problem (as far as string allocation goes) is
not a good trade off, whatever the raw math says underneath. For instance
those people who might actually notice that building 4GB strings by naive
concatenation could always preallocate the string and then fill it in
afterwards. We could even expose an API to make that easier.

As for the case of sizing the stack, exponential growth represents a
probabilistic analysis that for any size we reach we are likely to reach
some exponentially larger size still. This is obviously not true. The stack
doesnt grow unbounded, it eventually will level off, and if we were to use
a sizing strategey like above when it goes linear it would be growing in
very large chunks, and I doubt that anyone would notice that it is growing
linearly in the extreme case.

We already use exponential growth for PVs (in Perl_sv_grow) with a default
> of 1.25.
>
Yeah, I know. That exponent is IMO too small when the string is very small,
and too large when the string is very large.  That is my point. Using a
fixed exponent doesn't make as much sense in practice with bounded memory
as it does in the rawest of big O analysis.

In the worst case the exponent means we would fail to allocate the
requested memory, even though if we allocated less we would have enough
space to concatenate in the desired string.

Anyway, my point is that a naive single exponent is actually a poor choice.
Using a larger exponent when the size is small, and a smaller exponent when
it is large, and a fixed size when it is very large seems like a better
balance for our users, with few (if any) who would notice the change. I
also think that it is likely that the exponents should be different for
different purposes, and could take into account data like "how more are
being requested". For instance it might be better in some cases for the
stack to grow by a large exponent on the requested size, as compared to a
small exponent on the total size. Eg, consider if the stack is 100 elements
large, and we request 1000, we would end up 1100 * 1.25 = 1375. But maybe
it would be better to say "1.25 of the total, or twice the requested size",
in which case we would end up at 2200.

Anyway, this is a complex discussion. Using a fixed exponent is not
*wrong*, in the theoretical sense anyway, but it is not necessarily the
best balance in the practical sense. In practice there is a probabilistic
question of "when will growth stop" and "how large will we grow by each
step". We don't know the answers to these questions, and they will vary by
workload. We might end up making this toy program run faster

use integer;
my $s= "";
my $i=0;
$s.="x" while ++$i; # loop till integer rollover.

while burning memory such that our developers can only fit half the real
data in memory that they could if we werent so wastefully over-allocating.

cheers,
Yves

-- 

iabyn:
First a quick observation about the commit message. Please write messages
whose first line describes what part of perl is being changed. When
looking through a series of commits, seeing a commit like

    Replaced SSGROW with SSCHECK

sounds like this has been done in all the perl core and maybe all the
bundled distributions. Something like

    regcppush(): for more regex speed, use SSCHECK not SSGROW

tells the reader what's affected and why, without having to read the whole
commit.

Now, some general observations about SSGROW / SSCHECK in the regex engine.

regcppush() is used to save the current state of captures when pushing
a new regex state, and is only used for a small number of operations:
CURLYX, EVAL, GOSUB (IIRC).

It's a vestigial part of when the engine used to recurse (and thus saved its
state on the C stack). Some variable-sized stuff was saved on the save
stack instead.

With the addition of chains of regmatch_state blocks and removal of
recursion, it no longer makes sense to use the save stack.

I had a vague plan that such state saves could be done using a couple of
new struct regmatch_state union members. For simple patterns with a small
number of captures, all the capture info could be saved in a single block
that has space for the various vars like maxopenparen and a few capture
states; for larger numbers, additional bocks containing just (start,
start_tmp, end) triplets would be used in addition. The blocks would be
skipped / processed as appropriate during regex state unwinding.

-- 
    -- The Diary of Samuel Pepys

AnFunctionArray:
@iabyn 

In my defence I'll say there weren't many instances of SSGROW in the source tree (IRC the only other instance was in scope.c).

Also what do you mean by "With the addition of chains of regmatch_state blocks and removal of
recursion, it no longer makes sense to use the save stack."?

AnFunctionArray:
I was about to otherwise suggest querying the available memory - specifically for Win32 API (where the problem occurs).

There are several APIs available in Windows (like https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex).

demerphq:
>Also what do you mean by "With the addition of chains of regmatch_state blocks and removal of
recursion, it no longer makes sense to use the save stack."?

I am not sure if I fully understand what @iabyn means (I get it generally but I'm not sure I quite follow how this can work with EVAL or scenarios where we want to save "all the buffers"), but at a high level, the concept is that the information we put on the save stack can be stored in the regmatch_state blocks, and restored appropriately as those state blocks are unwound. Essentially the state stack unwinding is the thing that would trigger the restore operations in the first place, and if those save states contain the require data then there is no need to have a save stack.

The general idea is that when you enter a new state you record the things that will change in the state record. So for instance the way pattern recursion is implemented is by having a "stop at close paren"  (close_paren) variable, which when zero means "not in recursion", and when non-zero means, "when execution reaches the close paren with this number 'return' from the sub". When we enter recursion we store the old value of the variable in the state. So the initial recursion would store 0, and then change the global state var to be X. If further recursion occurs, we push a new state frame, and it would store X, and we would set the var to Y. Eventually we hit `CLOSE #Y`, which causes us to unwind the state stack, so extract the old value,  so for instance restore the var back to X, and continue matching at the appropriate regop from the "caller". Later when we it the `CLOSE #X`, we again unwind the state stack, set the close to 0, and continue with the top level pattern. (NOTE: I reviewed the code after i wrote the above, and while what I said above about a global variable (in the execution engine) /was/ true for the open/close paren, and is /generally/ true of the state operations, it turns out that this particular case was optimized and the variable *just* lives in the regexp state stack, there is no global var for this one anymore, but for most cases what is described here is correct. `close_paren` is exceptional however and we do not use an explicit global var for it and instead just use the most recent stack frame, but most cases are as I described. Search for `EVAL_CLOSE_PAREN` in regexp.c)

Every regex operation that can backtrack creates or updates the "current regexp state frame", and it is this data which allows backtracking to occur. For instance the TRIE code will find all the possible matches in a single "pass", create a list of them, and then store that list in its framedata so that when "the pattern to the right" fails and it needs to backtrack to one of the other matching options it can step through the list, when the list is exhausted it "fails" and unwinds the stack to backtrack even further. Things that can't backtrack dont add new frames, but may modify the topmost in some cases.

I am glossing over a number of important details here, such as that the stack generally isn't unwound on success (GOSUB and EVAL are notable exceptions), only on failure, but at the same time there is a "success unwind" and a "failure unwind" which do different things.  But if you think a bit about how regexps are defined to behave and spend some time with `use re DEBUG=>'ALL'`, which IIRC should show the stack operations during matching, it will come clear.  Here is an abbreviated example of what you might see.

```
$ perl -Mre=Debug,EXECUTE,STACK,STATE -e'"aaba"=~/(?<a>a)(?<b>(?&a)b(?&a))/'
Matching REx "(?<a>a)(?<b>(?&a)b(?&a))" against "aaba"
Intuit: trying to determine minimum start position...
  doing 'check' fbm scan, [0..4] gave 0
  Found anchored substr "aaba" at offset 0 (rx_origin now 0)...
  (multiline anchor test skipped)
Intuit: Successfully guessed: match at offset 0
                             |   0| Setting an EVAL scope, savestack=3,
   0 <> <aaba>               |   0| 1:OPEN1 'a'(3)
   0 <> <aaba>               |   0| 3:EXACT <a>(5)
   1 <a> <aba>               |   0| 5:CLOSE1 'a'(7)
   1 <a> <aba>               |   0| 7:OPEN2 'b'(9)
   1 <a> <aba>               |   0| 9:GOSUB1[-8:1] 'a'(12)
                             |   0| entering GOSUB, prev_recurse_locinput=0 recurse_locinput[1]=564b45c3feb1
                             |   0| Setting an EVAL scope, savestack=15,
                             |   0| push #0   EVAL_postponed_AB yes
   1 <a> <aba>               |   1|  1:OPEN1 'a'(3)
   1 <a> <aba>               |   1|  3:EXACT <a>(5)
   2 <aa> <ba>               |   1|  5:CLOSE1 'a'(7)
                             |   1|  FAKE-END[before] GOSUB1 ce=564b45c28d20 recurse_locinput=0
                             |   1|  Setting an EVAL scope, savestack=25,
                             |   1|  END: EVAL trying tail ... (cur_eval=0)
                             |   1|  push #1   EVAL_postponed_AB yes
                             |   1|       #0   EVAL_postponed_AB yes
   2 <aa> <ba>               |   2|   12:EXACT <b>(14)
   3 <aab> <a>               |   2|   14:GOSUB1[-13:1] 'a'(17)
                             |   2|   entering GOSUB, prev_recurse_locinput=0 recurse_locinput[1]=564b45c3feb3
                             |   2|   Setting an EVAL scope, savestack=35,
                             |   2|   push #2   EVAL_postponed_AB yes
                             |   2|        #1   EVAL_postponed_AB yes
                             |   2|        #0   EVAL_postponed_AB yes
   3 <aab> <a>               |   3|    1:OPEN1 'a'(3)
   3 <aab> <a>               |   3|    3:EXACT <a>(5)
   4 <aaba> <>               |   3|    5:CLOSE1 'a'(7)
                             |   3|    FAKE-END[before] GOSUB1 ce=564b45c28e40 recurse_locinput=0
                             |   3|    Setting an EVAL scope, savestack=45,
                             |   3|    END: EVAL trying tail ... (cur_eval=0)
                             |   3|    push #3   EVAL_postponed_AB yes
                             |   3|         #2   EVAL_postponed_AB yes
                             |   3|         #1   EVAL_postponed_AB yes
   4 <aaba> <>               |   4|     17:CLOSE2 'b'(19)
   4 <aaba> <>               |   4|     19:END(0)
   4 <aaba> <>               |   4|     pop (yes) EVAL_postponed_AB[Y]
                             |   3|    EVAL_AB cur_eval=0 prev_eval=564b45c28e40
                             |   3|    EVAL_AB[after] GOSUB1 ce=564b45c28e40 recurse_locinput=564b45c3feb3
   4 <aaba> <>               |   3|    pop (yes) EVAL_postponed_AB[Y]
                             |   2|   EVAL_AB cur_eval=564b45c28e40 prev_eval=0
                             |   2|   EVAL_AB[before] GOSUB1 ce=564b45c28e40 recurse_locinput=0
   4 <aaba> <>               |   2|   pop (yes) EVAL_postponed_AB[Y]
                             |   1|  EVAL_AB cur_eval=0 prev_eval=564b45c28d20
                             |   1|  EVAL_AB[after] GOSUB1 ce=564b45c28d20 recurse_locinput=564b45c3feb1
   4 <aaba> <>               |   1|  pop (yes) EVAL_postponed_AB[Y]
                             |   0| EVAL_AB cur_eval=564b45c28d20 prev_eval=0
                             |   0| EVAL_AB[before] GOSUB1 ce=564b45c28d20 recurse_locinput=0
Match successful!

```

Anyway, the idea that @iabyn has is that when we enter an open, or perform an operation that might affect the capture buffers we do the bookkeeping required to unwind it in the regexp state vector in the same way I describe above. If the frame doesnt get unwound we dont need to restore the old value, etc. etc.  @iabyn hopefully can expand on some of the finer points himself.

iabyn:
On Wed, Nov 23, 2022 at 09:12:33AM -0800, AnFunctionArray wrote:
> Also what do you mean by "With the addition of chains of regmatch_state blocks and removal of
> recursion, it no longer makes sense to use the save stack.".

Perl's regexp engine is implemented using backtracking. In something like

    "aaaaab" =~ /^a*aab/,

then (ignoring optimisations), the a* will initially match 5 a's, then the
engine will fail to match "aab". It will backtrack, this time matching 4
a's and still failing on "aab"; then backtrack again, and this time, match
against 3 a's and 'aab'.

To do backtracking, the regex engine has to save its current state at any
decision point, so that it can be restored when backtracking.

Before my work for perl 5.10.0 (released in 2007) the regex engine
used to keep most of its state in local variables and recurse when trying
the next step. Backtracking involved returning from some levels of
S_regmatch() recursion, with the local variables being automatically
restored due to the fact that multiple copies of the variables were stored
on the CPU stack, one set per depth of recursion.

One problem with this was that for something like:

    /(?: ...(.)...(.)...(.)...)*..../

The state to be stored for each depth of '*' recursion had to include
the current matching state for each of $1,$2,$3. Since patterns have a
variable (and not capped) number of captures, this state couldn't be
stored in a fixed number of local variables.

So instead, the regex engine (mis)used the save stack as a sort of
temporary storage to store a variable-sized chunk of data for each new
level of recursion. This was only needed for the few regex op types, like
CURLYX/WHILEM, which needed to save and possibly restore all the current
capture info.

In perl-5.10.0, I changed the regex engine from recursive to iterative. So
rather than storing the current state on the CPU stack (in the form of
function local variables), it now saves state in malloc()ed chunks of
memory. This has an important advantage in that, for long strings, the
old regex engine could recurse deeply and blow the CPU stack (especially
in threaded environments, where each thread's max stack size was small by
default).

In particular, I created a structure called regmatch_state (which is
mainly a union) to save the current state for each type of regex op.
The regex engine would allocate a block of memory which could hold an
array of regmatch_state's. When that became full, a new block would be
malloc()ed and the blocks chained together. (These blocks are struct
regmatch_slab).

However, I left the capture state saving alone, so it is being done on the
save stack, even though I knew it wasn't ideal. My intention was to
revisit at some point, and instead of saving capture state on the save
stack, save it by allocating one or more extra regmatch_state structs to
hold the capture indices at the same time as pushing a WHILEM-type
regmatch_state.

I'm not saving that you need to do this; I was just mentioning it as a
general observation in case you or anyone else intended to do any further
work in this area.

-- 
"You may not work around any technical limitations in the software"
    -- Windows Vista license

demerphq:
@iabyn the thing i dont get is how you would handle a vector of parens? Eg, if we recurse into a sub pattern, we want to save the paren's it touches, not all the parens in the parent (iiuir). Would you keep using the save stack for eval/gosub and just use your alternative for things where we can always preserve the old values safely? Or am i mistaken in thinking they are different cases and I am missing something?

iabyn:
On Thu, Nov 24, 2022 at 07:24:34AM -0800, Yves Orton wrote:
> @iabyn the thing i dont get is how you would handle a vector of parens?
> Eg, if we recurse into a sub pattern, we want to save the paren's it
> touches, not all the parens in the parent (iiuir). Would you keep using
> the save stack for eval/gosub and just use your alternative for things
> where we can always preserve the old values safely? Or am i mistaken in
> thinking they are different cases and I am missing something?

The idea is that the code will save as many parens as need saving by
allocating a suitable number of regmatch_state blocks. For a small number
of parens a single regmatch_state would be used; for larger, use multiple
structs. There would be two new types of union within the regmatch_state
struct:  one that holds maxopenparen, lastcloseparen etc plus a small
array of (start, start_tmp, end) triplets.  The number would be chosen to
make this element of the union no larger than the existing largest union
member. Then there would be second new union type, not containing
maxopenparen et, but holding a bigger fixed set of triplets. S_regcppush()
or its equivalent would push at least one new regmatch_state block, and if
there were more than a few parens needing saving, additional blocks as
required.

My intention is that the save stack would not be used for *anything* in
the regex engine, and functions like S_regcppush() (no matter whether
called from EVAL or GOSUB or WHILEM), would save capture state using
regmatch_state blocks rather than the save stack.

-- 
Counsellor Troi states something other than the blindingly obvious.
    -- Things That Never Happen in "Star Trek" #16

demerphq:
> The idea is that the code will save as many parens as need saving by allocating a suitable number of regmatch_state blocks

I see, thank @iabyn it makes sense to me now. That was the bit that eluded me.

demerphq:
@AnFunctionArray were you going to push a fix for the commit message?

AnFunctionArray:
> @AnFunctionArray were you going to push a fix for the commit message?

I might.

Done.

demerphq (approved):

iabyn:
On Tue, Feb 07, 2023 at 05:23:04AM -0800, Yves Orton wrote:
> Merged #20530 into blead.

This patch is flawed.

by replacing SSGROW(n) - which grows the savestack by n,
with SSCHECK(n) - which grows the savestack by 50% and ignores n,
it's possible to overrun the end of the savestack when the stack is small
and the number of captures to save is large.

For example, the following code fails under valgrind:

    $_ = "x" x 1000;
    my $pat = '(.)' x 200;
    $pat = qr/($pat)+/;
    m/$pat/;


-- 
Never work with children, animals, or actors.

demerphq:
@iabyn ack. im looking at it now.

demerphq:
@iabyn hopefully fixed in https://github.com/Perl/perl5/pull/20824

PR #20531, opened 2022-11-21
t/op/readline.t: skip error handling checks on solaris with PERLIO=stdio

### LINKED ISSUE #20528: t/op/readline.t:  Recent commit breaks two tests in stdio runs on certain Solaris
Since May of this year we have been receiving smoke-testing reports for Solaris 2.11 -- more specifically, from Joyent SmartOS 18.3.0.  For the last three months, those reports have been indicating two unit test failures in t/op/readline.t.  In these recent reports, t/op/readline.t (35-36) fails on all 'stdio' and 'stdio DEBUGGING' runs but appears to generally pass on 'perlio' and 'perlio DEBUGGING'.

First report of failure in t/op/readline.t on perlsmoke.srv.customer.skylime.net:
https://perl5.test-smoke.org/report/5022710
Smoke test run at v5.37.3-255-g14580ae6b9 on 2022-09-09T10:27:09Z

All tests in t/op/readline.t were PASSing at this earlier commit:
https://perl5.test-smoke.org/report/5021501
Smoke test run at v5.37.2-154-gbf2a3dae9f on 2022-08-13T10:59:07Z

Relevant part of t/op/readline.t:
```
292 SKIP:
293 {
294     skip_without_dynamic_extension("IO", 4);
295     my $tmpfile = tempfile();
296     open my $fh, ">", $tmpfile
297         or die "Cannot open $tmpfile: $!";
298     ok(!$fh->error, "no error before we try to read");
299     ok(!<$fh>, "fail to readline file opened for write");
300     ok($fh->error, "error after trying to readline file opened for write");
301     ok(!close($fh), "closing the file should fail");
302 }
```
Tests 35-36 in t/op/readline.t were added to core in 0b602161d2f4c5e6b22144e9f8b9f8128a35f440, whose commit date falls in between the last good report and the first bad report:
```
commit 0b602161d2f4c5e6b22144e9f8b9f8128a35f440
Author:     Tony Cook <tony@develop-help.com>
AuthorDate: Thu Aug 18 11:10:51 2022 +1000
Commit:     Tony Cook <tony@develop-help.com>
CommitDate: Wed Aug 31 10:51:09 2022 +1000

    test that <> doesn't clear the stream error state
```
Currently we are getting smoke-test reports from only one other host whose operating system is described as 'Solaris 2.11'.  t/op/readline.t is passing all its tests on that host, but that host is *not* also described as 'Joyent SmartOS 18.3.0'.  Example report:  https://perl5.test-smoke.org/report/5026248.  I have no expertise in, or access to, Solaris, so I can't diagnose this further.

@tonycoz, can you take a look?

tonycoz:
In this test, readline() calls down into PerlIOStdio_read(), which uses fgetc() for a single character read, which fails as expected.

But fgetc() doesn't set the error flag on the stream, so the error flag test and the close() test fail.

So skip the tests.

Fixes #20528

Confirmed with the following code:
```
#include <stdio.h>
#include <stdlib.h>

int main() {
  FILE *f = fopen("test.txt", "w");
  if (!f) {
    perror("fopen");
    exit(1);
  }
  int c = fgetc(f);
  if (c != EOF) { /* should fail */
    printf("Expected fgetc() failure didn't occur\n");
    exit(1);
  }
  perror("fgetc");
  if (!ferror(f)) {
    printf("error flag not set\n");
    exit(1);
  }
  return 0;
}
```
Run:
```
tony@openindiana:~/dev/perl/git$ gcc -ostdio-error stdio-error.c
tony@openindiana:~/dev/perl/git$ ./stdio-error 
fgetc: Bad file number
error flag not set
```
On linux:
```
tony@venus:.../perl/git$ gcc -ostdio-error stdio-error.c
tony@venus:.../perl/git$ ./stdio-error 
fgetc: Bad file descriptor
tony@venus:.../perl/git$
```

### DIFF
--- a/t/op/readline.t
+++ b/t/op/readline.t
@@ -295,6 +295,9 @@ SKIP:
     my $tmpfile = tempfile();
     open my $fh, ">", $tmpfile
         or die "Cannot open $tmpfile: $!";
+    my @layers = PerlIO::get_layers($fh);
+    skip "fgetc doesn't set error flag on failure on solaris likes", 4
+        if $^O eq 'solaris' && $layers[-1] eq 'stdio';
     ok(!$fh->error, "no error before we try to read");
     ok(!<$fh>, "fail to readline file opened for write");
     ok($fh->error, "error after trying to readline file opened for write");


jkeenan:
I don't have access to Solaris to test this patch explicitly.  The C-code example works on FreeBSD as it does on Linux, so I can assume that a simple SKIP on Solaris is sufficient to resolve the problem. +1

khwilliamson (approved):

PR #20541, opened 2022-11-25
locale.c: Fix memory leak

khwilliamson:
PL_C_locale_obj is a global variable, and should be allocated just once per program.  Prior to this commit it could leak under MULTIPLICITY.  I was unable to get LSAN to notice this leak.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -4827,7 +4827,9 @@ Perl_init_i18nl10n(pTHX_ int printwarn)
 #  endif
 #  ifdef USE_POSIX_2008_LOCALE
 
-    PL_C_locale_obj = newlocale(LC_ALL_MASK, "C", (locale_t) 0);
+    if (! PL_C_locale_obj) {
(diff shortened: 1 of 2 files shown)


tonycoz (approved):

tonycoz:
> I was unable to get LSAN to notice this leak.

I suspect `newlocale(LC_ALL_MASK, "C", (locale_t) 0);` returns a pointer to a static object on most implementations.  It certainly does on Linux/glibc:
```
Breakpoint 1, Perl_init_i18nl10n (my_perl=0x5555559d62a0, printwarn=1) at locale.c:4831
4831        if (! PL_C_locale_obj) {
(gdb) p PL_C_locale_obj 
$1 = (locale_t) 0x7ffff7df6960 <_nl_C_locobj>
```

khwilliamson:
Merged via 67f006ee23a6331e90c3c0429a81e23e0680d8ee

PR #20542, opened 2022-11-25
Fix compilation of -DPERL_MEM_LOG

khwilliamson:
This no longer compiled, giving the error

locale.c:2878:15: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]

The simple solution is to change memlog to explicitly have a const parameter.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -5821,7 +5821,7 @@ ST	|void	|mem_log_common |enum mem_log_type mlt			\
 				|const UV typesize			\
 				|NN const char *type_name		\
 				|NULLOK const SV *sv			\
-				|Malloc_t oldalloc			\
+				|const Malloc_t oldalloc		\
 				|Malloc_t newalloc			\
 				|NN const char *filename		\
 				|const int linenumber			\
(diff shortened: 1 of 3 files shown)


jkeenan:
> This no longer compiled, giving the error
> 
> locale.c:2878:15: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]

I can't recall seeing this (build-time, presumably) error.  What configuration options do you use to trigger it.

Also, at what point in the past *did* it compile?  At what commit did it start to fail to compile?

tonycoz:
> This no longer compiled, giving the error
> 
> locale.c:2878:15: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
> 
> The simple solution is to change memlog to explicitly have a const parameter.

The problem here is passing a const pointer to realloc().

Safefree() blindly (and incorrectly) casts away the const, so we don't see similar issues there.

I was working on fixing -Wcast-qual warnings, and had to remove const from a few dynamically allocated pointers.

demerphq (approved):

demerphq:
@tonycoz @khwilliamson is this ok to merge?

tonycoz:
> @tonycoz @khwilliamson is this ok to merge?

No, the dynamically allocated const pointers should be non-const.

demerphq:
@tonycoz does that mean we should just close this PR? there doesnt seem to be much else in the PR *but* adding const decls.

demerphq (changes requested):
@khwilliamson tonycoz seems to think this wrong. Can you respond or close the PR?

khwilliamson:
over taken by events

PR #20544, opened 2022-11-25
util.c: Remove unnecessary #ifdef

khwilliamson:
putenv now always expands to PelEnv_putenv; no need to explicitly state so.

### DIFF
--- a/util.c
+++ b/util.c
@@ -2494,12 +2494,7 @@ Perl_my_setenv(pTHX_ const char *nam, const char *val)
         char *new_env = S_env_alloc(NULL, nlen, vlen, 2, 1);
         /* all that work just for this */
         my_setenv_format(new_env, nam, nlen, val, vlen);
-#    ifndef WIN32
         putenv(new_env);
-#    else
-        PerlEnv_putenv(new_env);
-        safesysfree(new_env);
-#    endif
 
 #  endif /* HAS_SETENV */
 }


jkeenan:
> putenv now always expands to PelEnv_putenv; no need to explicitly state so.

I see that the code you're deleting in the patch was mostly added in April of this year.
```
471430840e1 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2497) #    ifndef WIN32
f28aedcca64 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2498)         putenv(new_env);
471430840e1 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2499) #    else
471430840e1 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2500)         PerlEnv_putenv(new_env);
471430840e1 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2501)         safesysfree(new_env);
471430840e1 (Tomasz Konojacki         2022-04-19 16:41:23 +0200 2502) #    endif
```
What subsequent commit rendered that code unnecessary?

util.c
```diff
@@ -2494,12 +2494,7 @@ Perl_my_setenv(pTHX_ const char *nam, const char *val)
         char *new_env = S_env_alloc(NULL, nlen, vlen, 2, 1);
         /* all that work just for this */
         my_setenv_format(new_env, nam, nlen, val, vlen);
-#    ifndef WIN32
         putenv(new_env);
-#    else
-        PerlEnv_putenv(new_env);
-        safesysfree(new_env);
-#    endif
```
tonycoz, line 2502: This has three problems.

- On Win32 this will leak memory, POSIX-like* putenv holds onto the pointer and includes it in the environment, but this isn't true for Win32.
- this bypasses the thread safety of PerlEnv_putenv() added in 24f3e849b5c on POSIX-likes
- the putenv() isn't replaced on Win32, so it calls the CRT putenv() rather than PerlEnv_putenv() aka win32_putenv(), which probably explains the Win32 test failures.  Replacing the putenv() with PerlEnv_putenv() made one of the failing tests pass.


* putenv() hasn't been POSIX for a while

khwilliamson:
Well, I certainly blew this one

PR #20545, opened 2022-11-25
loc_tools.pl: Add sub to classify locales UTF-8 vs non

khwilliamson:
The new sub taks a list of locales available on the platform and separates the UTF-8 ones from the non-UTF-8 ones, returning two sub-lists


jkeenan:
LGTM

tonycoz (approved):

PR #20546, opened 2022-11-25
Pod-Html auxiliary subroutines now imported only from Pod::Html::Util

jkeenan:
As part of the refactoring of Pod-Html conducted during the 5.35/5.36 development cycle, three subroutines were moved from ext/Pod-Html/lib/Pod/Html.pm to ext/Pod-Html/lib/Pod/Html/Util.pm. To allow for the possibility that someone might have been importing any of these subroutines from Pod::Html, in 5.36 we imported them from Pod::Html::Util and allowed their re-exporting and noted in the documentation that this proviso would be for 5.36 only (68327975be0).

As of 5.38 we can dispense with this proviso and remove the test file used to exercise it as well.

Update pod/perldelta.pod as needed.  Keep porting tests happy.


tonycoz (approved):

tonycoz:
@craigberry did you mean to push those here?

jkeenan:
I made some errors in rebasing my branch on blead, so instead I manually applied the two patches in the p.r. (bcbede66ea9b2d1ecf2388cdeaa9690250ef7aee and 4d597d86ca65c369680ced9b942edbde4a6ec0a7) to blead and pushed them to origin.

Issue #20538, opened 2022-11-23
Collected issues related to @INC hooks.


demerphq:
**Description**
Perl supports "hooks" in the `@INC` array. The original idea was to allow for source filters, but because they were exposed as sub/method calls to the user the API opens up a number of troublesome issues which unfortunately the code does not account for. We have had bugs related to this previously, but we only fixed them superficially, we never addressed the fundamental issues involved. 

The main problem is that `@INC` hooks can and do modify the `@INC` array, and the behavior when they do so is pretty much accidental as opposed to intended, and can even result in accesses freed memory. 

So the first issue is that code like this:

```
perl -le'BEGIN { push @INC, sub { warn "overwriting INC array in glob"; *INC=[sub{warn "C"}]; }, sub{ warn "in B" } }; eval "require Frobnitz" or die $@'
```

will access the original `@INC` array after it is freed. This just happens to work out to just not do anything as it sees the freed array as having 0 elements and stops processing entirely. (It may be that the sub itself is also destroyed before it finishes executing, i havent checked) Arguably this code should never access freed memory and should output this:

```
$ ./perl -le'BEGIN { push @INC, sub { warn "overwriting INC array in glob"; *INC=[sub{warn "in C"}]; }, sub{ warn "in B" } }; eval "require Frobnitz" or die $@' 
overwriting INC array in glob at -e line 1.
in C at -e line 1.
Can't locate Frobnitz.pm in @INC (you may need to install the Frobnitz module) (@INC contains: CODE(0x556edbddb900)) at (eval 1) line 1.
```

Another interesting case is this:

```
perl -le'my $counter; BEGIN { push @INC, sub { warn "unshifting INC"; unshift @INC, sub{warn "in shift ",++$counter}; }, sub{ warn "in B" } }; eval "require Frobnitz" or die $@' 2>&1 | head
```

which outputs:

```
perl -le'my $counter; BEGIN { push @INC, sub { warn "unshifting INC"; unshift @INC, sub{warn "in shift ",++$counter}; }, sub{ warn "in B" } }; eval "require Frobnitz" or die $@' 2>&1 | head 
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
unshifting INC at -e line 1.
....
```

[...]

PR #20547, opened 2022-11-25
Harden @INC handling

demerphq:
This hardens `@INC` hook handling in a bunch of ways. It ensures error message show the truth, and it fixes segfaults, and other bugs, and it adds new facilities to INC hooks, ($SKIP and INCDIR)

Features (not bugfixes) in this PR:

- Add support for `$SKIP` this variable is set to the index of @INC that a hook exists in before the hook is executed. The hook may alter the value to set it to the integer predecessor of the index they want perl to look at next. Thus a hook that rewrites `@INC` can set `$SKIP` to -1 and have perl process the replaced contents. undef is supported to represent -1.
- Ensure that errors show the `@INC` entries inspected or called, and not the latest state of `@INC` which may not match at all.
- Add an INCDIR method, which is called just like an INC hook would be, but which can return a list of directories that perl will search. Returning zero directories is allowed.
- Add support for providing an object as the first argument of array style hook definitions. In this case we will pass the array in as the third argument so we can execute the call_method() with the object as the first argument.

Bugfixes:
- Fixes a segfault when an INC hook does something like this: `*INC=[ "foo", "bar", "baz"];`
- Error message being incorrect about what directories were inspected. For instance `push @INC, sub { @INC=() }`, will trigger right at the end of the @INC search, and the error message will say that we searched no directories, when in fact we searched the normal list.
- Allow blessed subrefs with no INC or INCDIR methods defined to be treated as unblessed coderefs.
- Allow blessed refs with no INC or INCDIR methods to be used in `@INC` so their overloads are triggered.

This should resolve all the conceptual issues raised in https://github.com/Perl/perl5/issues/20538

Edited: removed dedupe logic discussed below.

### DIFF
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4294,7 +4294,7 @@ S_require_file(pTHX_ SV *sv)
                         SvSetSV_nosteal(nsv,sv);
                     }
 
-                    ENTER_with_name("call_INC");
+                    ENTER_with_name("call_INC_hook");
                     SAVETMPS;
                     EXTEND(SP, 2);
 
@@ -4374,7 +4374,7 @@ S_require_file(pTHX_ SV *sv)
 
                     PUTBACK;
                     FREETMPS;
-                    LEAVE_with_name("call_INC");
+                    LEAVE_with_name("call_INC_hook");
 
                     /* Now re-mortalize it. */
                     sv_2mortal(filter_cache);
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4260,16 +4260,16 @@ S_require_file(pTHX_ SV *sv)
      * For searchable paths, just search @INC normally
      */
     if (!tryrsfp && !(errno == EACCES && !path_searchable)) {
-        AV * const ar = GvAVn(PL_incgv);
-        SSize_t i;
+        AV * const inc_ar = GvAVn(PL_incgv);
+        SSize_t inc_idx;
 #ifdef VMS
         if (vms_unixname)
 #endif
         {
             SV *nsv = sv;
             namesv = newSV_type(SVt_PV);
-            for (i = 0; i <= AvFILL(ar); i++) {
-                SV * const dirsv = *av_fetch(ar, i, TRUE);
+            for (inc_idx = 0; inc_idx <= AvFILL(inc_ar); inc_idx++) {
+                SV * const dirsv = *av_fetch(inc_ar, inc_idx, TRUE);
 
                 SvGETMAGIC(dirsv);
                 if (SvROK(dirsv)) {
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4259,6 +4259,7 @@ S_require_file(pTHX_ SV *sv)
      *
      * For searchable paths, just search @INC normally
      */
+    AV *inc_checked = (AV*)sv_2mortal((SV*)newAV());
     if (!tryrsfp && !(errno == EACCES && !path_searchable)) {
         SSize_t inc_idx;
 #ifdef VMS
@@ -4269,9 +4270,21 @@ S_require_file(pTHX_ SV *sv)
             namesv = newSV_type(SVt_PV);
             AV *inc_ar = GvAVn(PL_incgv);
             for (inc_idx = 0; inc_idx <= AvFILL(inc_ar); inc_idx++) {
-                SV * const dirsv = *av_fetch(inc_ar, inc_idx, TRUE);
+                SV *dirsv = *av_fetch(inc_ar, inc_idx, TRUE);
 
-                SvGETMAGIC(dirsv);
+                if (SvGMAGICAL(dirsv)) {
+                    SvGETMAGIC(dirsv);
+                    dirsv = newSVsv_nomg(dirsv);
+                } else {
+                    /* on the other hand, since we aren't copying we do need
+                     * to increment */
+                    SvREFCNT_inc(dirsv);
+                }
+                if (!SvOK(dirsv))
+                    continue;
+
+                av_push(inc_checked, dirsv);
+
                 if (SvROK(dirsv)) {
                     int count;
                     SV **svp;
@@ -4536,14 +4549,15 @@ S_require_file(pTHX_ SV *sv)
                 DIE(aTHX_ "Can't locate %s:   %s: %s",
                     name, tryname, Strerror(saved_errno));
             } else {
-                if (path_searchable) {		/* did we lookup @INC? */
-                    AV * const ar = GvAVn(PL_incgv);
+                if (path_searchable) {          /* did we lookup @INC? */
                     SSize_t i;
                     SV *const msg = newSVpvs_flags("", SVs_TEMP);
                     SV *const inc = newSVpvs_flags("", SVs_TEMP);
-                    for (i = 0; i <= AvFILL(ar); i++) {
+                    for (i = 0; i <= AvFILL(inc_checked); i++) {
+                        SV **svp= av_fetch(inc_checked, i, TRUE);
+                        if (!svp || !*svp) continue;
                         sv_catpvs(inc, " ");
-                        sv_catsv(inc, *av_fetch(ar, i, TRUE));
+                        sv_catsv(inc, *svp);
                     }
                     if (memENDPs(name, len, ".pm")) {
                         const char *e = name + len - (sizeof(".pm") - 1);
@@ -4597,7 +4611,7 @@ S_require_file(pTHX_ SV *sv)
 
                     /* diag_listed_as: Can't locate %s */
                     DIE(aTHX_
-                        "Can't locate %s in @INC%" SVf " (@INC contains:%" SVf ")",
+                        "Can't locate %s in @INC%" SVf " (@INC entries checked:%" SVf ")",
                         name, msg, inc);
                 }
             }
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4307,10 +4307,13 @@ S_require_file(pTHX_ SV *sv)
                         sv_setsv_nomg(l, loader);
                         loader = l;
                     }
-                    if (sv_isobject(loader))
-                        count = call_method("INC", G_LIST);
-                    else
-                        count = call_sv(loader, G_LIST);
+                    const char *method = NULL;
+                    if (sv_isobject(loader)) {
+                        method = "INC";
+                        count = call_method(method, G_LIST|G_EVAL);
+                    } else {
+                        count = call_sv(loader, G_LIST|G_EVAL);
+                    }
                     SPAGAIN;
 
                     if (count > 0) {
@@ -4367,6 +4370,30 @@ S_require_file(pTHX_ SV *sv)
                                                   PERL_SCRIPT_MODE);
                         }
                         SP--;
+                    } else {
+                        SV *errsv= ERRSV;
+                        if (SvTRUE(errsv) && !SvROK(errsv)) {
+                            STRLEN l;
+                            char *pv= SvPV(errsv,l);
+                            /* Heuristic to tell if this error message
+                             * includes the standard line number info:
+                             * check if the line ends in digit dot newline.
+                             * If it does then we add some extra info so
+                             * its obvious this is coming from a hook.
+                             * If it is a user generated error we try to
+                             * leave it alone. l>12 is to ensure the
+                             * other checks are in string, but also
+                             * accounts for "at ... line 1.\n" to a
+                             * certain extent. Really we should check
+                             * further, but this is good enough for back
+                             * compat I think.
+                             */
+                            if (l>=12 && pv[l-1] == '\n' && pv[l-2] == '.' && isDIGIT(pv[l-3]))
+                                sv_catpvf(errsv, "%s %s hook died--halting @INC search",
+                                          method ? method : "INC",
+                                          method ? "method" : "sub");
+                            croak_sv(errsv);
+                        }
                     }
 
                     /* FREETMPS may free our filter_cache */
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4319,11 +4319,6 @@ S_require_file(pTHX_ SV *sv)
                         SvGETMAGIC(loader);
                     }
 
-                    Perl_sv_setpvf(aTHX_ namesv, "/loader/0x%" UVxf "/%s",
-                                   diruv, name);
-                    tryname = SvPVX_const(namesv);
-                    tryrsfp = NULL;
-
                     if (SvPADTMP(nsv)) {
                         nsv = sv_newmortal();
                         SvSetSV_nosteal(nsv,sv);
@@ -4369,6 +4364,11 @@ S_require_file(pTHX_ SV *sv)
                             }
                         }
                     }
+
+                    Perl_sv_setpvf(aTHX_ namesv, "/loader/0x%" UVxf "/%s",
+                                   diruv, name);
+                    tryname = SvPVX_const(namesv);
+                    tryrsfp = NULL;
 
                     ENTER_with_name("call_INC_hook");
                     SAVETMPS;
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4307,6 +4307,9 @@ S_require_file(pTHX_ SV *sv)
                         SvSetSV_nosteal(nsv,sv);
                     }
 
+                    SV * inc_idx_sv = save_scalar(PL_incgv);
+                    sv_setiv(inc_idx_sv,inc_idx);
+
                     ENTER_with_name("call_INC_hook");
                     SAVETMPS;
                     EXTEND(SP, 2);
@@ -4411,6 +4414,21 @@ S_require_file(pTHX_ SV *sv)
 
                     /* FREETMPS may free our filter_cache */
                     SvREFCNT_inc_simple_void(filter_cache);
+
+                    /*
+                     Let the hook override which @INC entry we visit
+                     next by setting $INC to a different value than it
+                     was before we called the hook. If they have
+                     completely rewritten the array they might want us
+                     to start traversing from the beginning, which is
+                     represented by -1. We use undef as an equivalent of
+                     -1. This can't be used as a way to call a hook
+                     twice, as we still dedupe.
+                     We have to do this before we LEAVE, as we localized
+                     $INC before we called the hook.
+                    */
+                    inc_idx_sv = GvSVn(PL_incgv);
+                    inc_idx = SvOK(inc_idx_sv) ? SvIV(inc_idx_sv) : -1;
 
                     PUTBACK;
                     FREETMPS;
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4266,11 +4266,32 @@ S_require_file(pTHX_ SV *sv)
         if (vms_unixname)
 #endif
         {
-            SV *nsv = sv;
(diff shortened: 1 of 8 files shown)


jkeenan:
> This hardens `@INC` hook handling in a bunch of ways. It ensures error message show the truth, and it fixes segfaults, and other bugs, and it adds new facilities to INC hooks, ($SKIP and INCDIR)
> 
> Features (not bugfixes) in this PR:
> 
[snip]
> 
> Bugfixes:
> 

Someone is bound to ask (so let's get this question out of the way first) ... 

This p.r. does a lot.  Could it be broken down into smaller p.r.'s?  For example, could the bugfixes be handled before the new features?

demerphq:
@jkeenan  i dont plan to break up this PR. I originally did this as a single patch. I took an extra day to break it up into a bunch of smaller commits. All of the logic involved is in a single call site, and basically interacts with each other. So i see it as all one thing, the `@INC` api is broken, segfaults, and is really easy to get very very wrong.  The patches here ameliorate those issues and provide solutions. For instance @ap said he can make the code in base.pm, which is fragile, not be fragile anymore with the features added here.  Breaking it up into separate PR's would just be more work for everyone involved, for no real benefit to the project. So, no, i dont plan to break this up into smaller PR's. If someone has strong objections to one of the commits I might rebase that patch out. but i see no benefit to adding this as multiple PR's.

Note, i left the pod and tests as a follow up patch so that if we want to drop part of the commit sequence, I just have to update the one commit, and not deal with conflicts through the resolution chain. I was also thinking maybe once the details are ironed out I will rebase it so the docs and tests are applied first, the tests as TODO tests, and then untodo the relevant test in each patch as it goes. But right now id rather keep them as a single corpus at the end until I get more feedback.

haarg:
I'm happy to see the fixes here, but have some concerns.

First, and most minor, the error when failing to load a module includes the string `@INC contains:`. With this patch and using INCDIR, it instead reports the directories it searched. This is, strictly speaking, incorrect. The extra directories are not and never were in `@INC`. This could be misleading because it makes it appear that the dynamically checked directories were part of the static search list.

Second, de-duplicating both directories and hook calls seems wrong. I'm not sure what type of issue only calling a given hook once is meant to protect against, and there isn't a test for it. And I can think if a number of reasons to have both duplicate hooks and duplicate directories in `@INC`. It seems like a needless limitation.

demerphq:
> First, and most minor, the error when failing to load a module includes the string `@INC` contains:. With this patch and using INCDIR, it instead reports the directories it searched. This is, strictly speaking, incorrect. The extra directories are not and never were in `@INC`. This could be misleading because it makes it appear that the dynamically checked directories were part of the static search list.

Yes. Indeed. I left the "`@INC contains`" part of the error message alone, as I feel like there is likely a large amount of code that looks for it in error messages. But at the same time, I feel pretty strongly that we should report where we looked, not what `@INC` looks like after we have looked. 

Previously we reported unreliable data. FOr instance:

```
push @INC, sub { @INC= (); }
```

Would cause the error message to show an empty list of directories searched, even though we have searched many. Similarly, a tied variable might return one thing to the code that does the directory lookup, but another when we build the error message.

This leaves us in a sticky place in terms of fixing this. Reporting the state of `@INC` after the call is not helpful if that data can be corrupted, and changing the constant text in the error message would mean that we have a lot more BBC fallout (and BBB fallout frankly - I am pretty confident B has code that looks for that string).

We either continue to have an ineffective and corruptable error message, or we change the text and break lots of stuff. I decided the least worse choice was to leave the constant text unchanged, but to show the correct data about what "things" were inspected.

I don't mind changing the constant text if people think that makes sense. I really do object to us producing error messages which are unhelpful or confusing however. 

> Second, de-duplicating both directories and hook calls seems wrong. I'm not sure what type of issue only calling a given hook once is meant to protect against, and there isn't a test for it.

I will have to double check that later, I think if you look carefully you will see that it is tested implicitly. 

> And I can think if a number of reasons to have both duplicate hooks and duplicate directories in `@INC`. It seems like a needless limitation.

I think you need to outline more on that. How would having the same directory search twice be a desirable and good thing? Are you thinking of a hook that writes to disk, and relies on the double listing to find it twice?  Seems far fetched to me, but ok, i guess. I dont mind excluding the text directories from the deduper, although I think its a shame to do so. If someone had exerted this constraint from the very beginning I dont think anyone would be complaining.

However as far as *hooks* go, I do not agree. I think calling the same hook more than once is far more likely to be an accident than it is deliberate, and the deduper does not prevent its deliberate use. It is meant to trap this:

```
push @INC, sub { unshift @INC, "."};
```
will infinite loop in a part of our code where we will eventually exhaust the stack and segfault. I dont think we should be able to write perl code that can segfault or infinite loop that easily, especially as it is essentially an action-at-a-distance bomb that can blow up in totally unrelated code. Another case would be this:

```
push @INC, sub { @INC= shuffle @INC };
```
Would we want that to shuffle `@INC` a random number of times?

It is also meant to make it easier to rewrite `@INC` and preserve your hook in the result, without having to guard against it getting called twice during the same require. Eg consider this:

```
push @INC, sub { @INC = grep { !ref($_) || ($_ != __SUB__) } @INC; push @INC, @stuff, __SUB__; () }
```
With the deduper I can write the above and not worry about the callback getting called twice. It makes life nice and easy. But at the same time if someone really wants to create code that infinite loops like the above they can do this:

```
unshift @INC, [ sub { $INC[$INC]= [ $INC[$INC][0] ]; $INC-- } ];
```

I think its a prudent guard that will make life easier, and because of the array API does not make anything impossible which used to be possible. IOW, we do our best to keep people from blowing their foot off, but if they really want to do so they can.
 
If you can think of a number of reasons to have a hook called more than once can you please outline what they are, and explain why you couldn't use the array interface to hooks to achieve your result?

For instance, with the protection in place, if you wanted the same hook called at say the beginning and end of `@INC` you could do this:

```
$callback= sub { ... };
unshift @INC, [$callback];
push @INC, [$callback];
```

However it would stop this:

```
my $callback= [ sub { ... } ];
unshift @INC, $callback;
push @INC, $callback;
```

I don't think this is an unreasonable restriction. It makes it much harder to shoot yourself in the foot in a way that is super hard to debug, but it still allows power users to do what they want. Specifically it does NOT prevent anyone from creating a hook that gets called twice during the same require, if that is what they want, but it forces them to do it in a way where it must be deliberate and intentional, not accidental.

Note from the point of view of the internals:

```
$callback = sub { };
push @INC, [ $callback ], [ $callback ];
```
is two separate hooks, even though it resolves down to the same callback.

PS: Sorry for the lag, somehow I didnt see any notification that you had responded to this.

haarg:
> Yes. Indeed. I left the "`@INC contains`" part of the error message alone, as I feel like there is likely a large amount of code that looks for it in error messages. But at the same time, I feel pretty strongly that we should report where we looked, not what `@INC` looks like after we have looked.
> 
> ...
> 
> This leaves us in a sticky place in terms of fixing this. Reporting the state of `@INC` after the call is not helpful if that data can be corrupted, and changing the constant text in the error message would mean that we have a lot more BBC fallout (and BBB fallout frankly - I am pretty confident B has code that looks for that string).

I think we probably have to bite the bullet on that breakage. Showing the list of items actually searched seems like a decent change, but using an incorrect error doesn't seem viable to me. A quick scan over CPAN doesn't show too many things that are likely to be broken by this.


> I think you need to outline more on that. How would having the same directory search twice be a desirable and good thing? Are you thinking of a hook that writes to disk, and relies on the double listing to find it twice?

This is pretty much exactly the case I was thinking of.

> However as far as _hooks_ go, I do not agree. I think calling the same hook more than once is far more likely to be an accident than it is deliberate, and the deduper does not prevent its deliberate use. It is meant to trap this:
> 
> ```
> push @INC, sub { unshift @INC, "."};
> ```
> 
> will infinite loop in a part of our code where we will eventually exhaust the stack and segfault.

Does it actually do that? It leads to an infinite loop, but there's no recursion. It doesn't seem like it would exhaust the stack.

> I dont think we should be able to write perl code that can segfault or infinite loop that easily

There are pretty much infinite ways to create an infinite loop in perl. I don't see the value in trying to prevent this type, unless they lead to things like segfaults. Especially since `@INC` hooks are not common, and problems with them will usually show up at process start.

> Another case would be this:
> 
> ```
> push @INC, sub { @INC= shuffle @INC };
> ```
> 
> Would we want that to shuffle `@INC` a random number of times?

Honestly, yes. This really seems like a "you get to keep both pieces" problem. If someone does something that silly, we don't need to protect them.

As for a useful case for calling a hook multiple times, this PR provides a feature that would make them useful while debugging:

```
my $hook = sub { warn "reached $INC"; return; };
unshift @INC, $hook;
push @INC, $hook;
splice @INC, 5, 0, $hook;
```

> For instance, with the protection in place, if you wanted the same hook called at say the beginning and end of `@INC` you could do this:
> 
> ```
> $callback= sub { ... };
> unshift @INC, [$callback];
> push @INC, [$callback];
> ```
> 
> However it would stop this:
> 
> ```
> my $callback= [ sub { ... } ];
> unshift @INC, $callback;
> push @INC, $callback;
> ```
> 

This just feels like something that is confusing to document and not at all an obvious restriction to me.

ap:
> > Are you thinking of a hook that writes to disk, and relies on the double listing to find it twice?
> 
> This is pretty much exactly the case I was thinking of.

FWIW it is also the first thing that came to mind when I thought about plausible examples where it might be useful. 🙂

demerphq:
@haarg and @ap:

> I think we probably have to bite the bullet on that breakage.

Ok. What I did was a punt, I dont mind changing it to something like "`@INC` entries checked:" although suggestions welcome. Just so long as you take some of the heat if it breaks stuff. :-)

> There are pretty much infinite ways to create an infinite loop in perl. I don't see the value in trying to prevent this type, unless they lead to things like segfaults.

I have a pretty strong view on this. Maybe I am wrong about the segfault, it is possible I am confusing different bugs I encountered. However the infinite loop on something that should be a bounded process, combined with the action at a distance nature and the fact that allowing dupe calls makes it harder to write subs that rewrite `@INC` says to me that the block is sensible. I think its pretty reasonable to make it difficult to casually write inc bombs.

The example you gave is IMO perfectly served by wrapping the subs in fresh arrays.  I didnt find documenting the trick with arrays to be at all difficult, I have already done so in the PR.  Check the wording, it is pretty straight forward. https://github.com/Perl/perl5/pull/20547/commits/547f010f39d78b59e6342c0c96b2562106d4c198#diff-230e04648f168b0b966360f70f52dadc6e058290f63cb249d8539d43432112f6R6901

> This is pretty much exactly the case I was thinking of.
>>FWIW it is also the first thing that came to mind when I thought about plausible examples where it might be useful. slightly_smiling_face

I don't really see why a hook that does this would not just write the file to disk and then return a filehandle to the file it just created via the standard INC machinery. Why depend on the second directory trick? It starts to sound convoluted and contrived.

I feel much less strongly about deduping directories compared to my view on hooks, but it also seems totally reasonable to dedup at the app level. What about slow disk scenarios?

Anyway, since the deduping is the only controversial part of the PR I will separate out that part to a separate PR (at the same time changing the error message) and we can debate the deduping separately.  We can get the stuff we do agree on out of the way first.

ap:
> I don't really see why a hook that does this would not just write the file to disk and then return a filehandle to the file it just created via the standard INC machinery. Why depend on the second directory trick?

If you happen to have an open handle to the file already, sure, but maybe you called an external program to extract an archive or check out from a repo, or maybe you used a module that doesn’t expose the guts of the extraction process in this way, or whatever.

Basically whatever answer you give for “hooks can already return handles, what’s the point of `INCDIR`?” is also going to be an answer to your own question.

> Maybe I am wrong about the segfault, it is possible I am confusing different bugs I encountered.

FWIW I am entirely in agreement that nothing code does to `@INC` should cause a segfault, even as it’s being iterated. But I don’t see the point of trying to prevent an infinite loop here. For hooks, I get what would compel you, but even then I think it’s not just unnecessary but an active obstacle to people who might have use cases you aren’t imagining right now. I don’t like that kind of restriction; `@INC` hooks already aren’t everyday stuff so it’s not hard to imagine that someone having a weird use case you or I aren’t going to be able to come up with on the spot in this thread.

(Certainly “I need to record the stack depth so I can tell the top-level `require` from nested ones” is not a solution that just popped into my head at day 1 of working on the base.pm patch.)

Generally for this sort of thing I have a strong preference to operational models that are primitive (“`@INC` will be iterated by index”) but robust (“no segfaults, no matter what you try”). Those are easiest to use in unforeseen ways. The more restrictions you add to the model, the harder that gets.

(Note that `$INC` is an extension to this model which makes it both *easier*, *and* more robust, to employ in novel ways. ++ all around to that.)

I *am* glad to see you added an escape hatch to allow duplicate uses of the same hook.

And making mistakes difficult to make by default is generally also a principle I subscribe to. Maybe I’m not giving this enough weight.

But my subjective sense is that this may be a case of overcorrection – something I see where people react to a problem by solving it from multiple angles at once, when just one of those solutions would have perfectly sufficiently balanced out the overall situation by taking away knock-on effects, where the extra solutions create their own undesirable knock-on effects instead. Maybe I’m wrong, as I said, this is just my subjective sense. But I’d make that case to try one thing at a time and see if that suffices, because it will be easier to judge its necessity this way vs doing everything at once and trying to discern whether it was too much.

Maybe de-duping hooks by default to prevent mistakes but allowing duplicate calls if explicitly requested will turn out to be the correct balance. My bias at this time is of course that it will prove unnecessary – but I’m open to finding out otherwise.

ap:
> Anyway, since the deduping is the only controversial part of the PR I will separate out that part to a separate PR (at the same time changing the error message) and we can debate the deduping separately. We can get the stuff we do agree on out of the way first.

++

demerphq:
I have created https://github.com/Corion/parent/pull/12 as parent is one of the modules which depends on the test wording.

demerphq:
Ok, so I have removed the dedupe logic from the PR, and changed the error message to say `@INC entries checked:` instead of `@INC contains:`. This broke the tests for parent.pm, so i pushed a PR for that, and include a patch to fix the tests in this PR. @Corion see https://github.com/Perl/perl5/pull/20547/commits/956cdd5211f15f6926e667bc9a27faa3cfc2358c

Ill push a separate PR with the dedupe logic so we can continue that discussion there.

demerphq:
@haarg and @ap you ok with this now?

demerphq:
And the dedupe logic has been moved to https://github.com/Perl/perl5/pull/20588

pp_ctl.c
```diff
@@ -4284,40 +4319,113 @@ S_require_file(pTHX_ SV *sv)
...
+                                }
+                                else if (SvGMAGICAL(arg)) {
+                                    arg = newSVsv_nomg(arg);
+                                }
+                                else {
+                                    SvREFCNT_inc(arg);
+                                }
+                                av_push(incdir_av, arg);
+                            }
+                            /* We copy $SIG into incdir_continue_inc_idx
```
haarg: Should this be `$INC`?

t/op/require_errors.t
```diff
@@ -9,7 +9,12 @@ BEGIN {
 use strict;
 use warnings;
 
-plan(tests => 59);
+plan(tests => 71);
+
+
+# Dedupe @INC. As of 5.37.7 we won't inspect a directory twice, nor show a
```
haarg: This comment is out of date.

pp_ctl.c
```diff
@@ -4284,40 +4319,113 @@ S_require_file(pTHX_ SV *sv)
...
+                            method = "INC";
+                        } else {
+                            gv = gv_fetchmethod_pvn_flags(pkg, "INCDIR", 6, 0);
+                            if (gv && isGV(gv)) {
+                                method = "INCDIR";
+                                is_incdir = TRUE;
+                            }
+                        }
+                        if (!method) {
+                            if (SvTYPE(SvRV(loader)) != SVt_PVCV) {
```
haarg, line 4359: Should this allow `&{}` overloads?

haarg (changes requested):

t/op/require_errors.t
demerphq: FIxed.

pp_ctl.c
demerphq: Hrm. I dunno. On one hand, yeah from a completeness point of view we probably should. On the other hand, we don't support objects with overloaded `@{}` to implement the array API either, and in fact historically we would assume any object in @INC would have an INC method, and blindly call it as a method, with the resulting errors when it was a blessed sub with no methods. Its a tricky call. Prior to this patch you couldn't put an object in the array anyway, so nothing is going to break...

pp_ctl.c
demerphq: fixed

pp_ctl.c
haarg: If we only allow unblessed array refs, it feels like we should only allow unblessed code refs.

demerphq: That doesn't make sense. If the array ref is blessed it is treated as an object, not as a hook wrapper. If it is unblessed then we treat it as a hook wrapper.  That is existing logic and can't change.

We used to ONLY allow subrefs in the zero slot of the wrapper, which made some sense because the only hook method we offered was INC. However we now have two methods, INCDIR and INC.  If we wish to have the wrapper approach work with both hooks we need to be able to supply an object in the zero slot.

The thread here doesnt show the full context. (I will update the comment below, it is misleading.)

```
                    if (sv_isobject(loader)) {
                        /* if it is an object and it has an INC method, then
                         * call the method. Otherwise assume its a blessed coderef.
                         */
                        HV *pkg = SvSTASH(SvRV(loader));
                        GV * gv = gv_fetchmethod_pvn_flags(pkg, "INC", 3, 0);
                        if (gv && isGV(gv)) {
                            method = "INC";
                        } else {
                            gv = gv_fetchmethod_pvn_flags(pkg, "INCDIR", 6, 0);
                            if (gv && isGV(gv)) {
                                method = "INCDIR";
                                is_incdir = TRUE;
                            }
                        }
                        if (!method) {
                            if (SvTYPE(SvRV(loader)) != SVt_PVCV) {
                                if (dirsv != loader)
                                    croak("Object with arguments in @INC does not support a hook method");
                                else
                                    goto treat_as_string;
                            }
                        }
                    }
```

So, if the loader is an object (either wrapper or unwrapped), we check to see if it has either hook methods. If it does not then we check if it is a blessed sub, and if it is then we call it, if it is not then we treat the object as a string.  In the special case where we have a wrapper (`dirsv != loader`), we are bit more careful, it doesn't make a lot of sense to use the wrapper api and include a non-coderef, non-hook bearing object. Unlike the unwrapped case, it doesn't make any sense to ignore the arguments (the wrapper), and fallback to strignifying the ref. Why would you provide a wrapper if it is just going to be thrown away to be stringified? 

In the old code, this would have worked differently. We would have seen the item was an object, tried to call the INC method and died with a cryptic error message.

With the new implementation, we expand things so that this:

```
push @INC, bless(sub { warn "@INC" }, "Me");
```

work the same as

```
push @INC, sub { warn "@INC" };
```

I believe that this is no bad thing. It helps debugging `@INC` hooks, and it doesn't break anything existing, and IMO it is more naturally DWIM anyway, the rules can say "An object with an INC or INCDIR hook, or a coderef". That isn't inconsistent with a blessed coderef with no method.

demerphq:
Ok, after discussion with haarg i am going ahead.

PR #20548, opened 2022-11-25
POSIX::localeconv: Return empty values

khwilliamson:
This function returns a hash allowing Perl access to the localeconv() data structure, with the keys being the structure's field names, and the values being their corresponding value in the current locale.

Prior to this commit, it did not populate the hash with any keys whose values are the empty string.  But this is wrong.  Those fields exist, and are explicitly allowed to be empty, except for the decimal point. For example, the symbol indicating a number is positive is empty in many locales.  Someone might want to look up that value, and discover it is undefined.  Or they might want to iterate over all the keys in the hash.

I couldn't find a reason in the history why these are omitted, and it seems to me to be wrong to omit them.

### DIFF
--- a/ext/POSIX/t/posix.t
+++ b/ext/POSIX/t/posix.t
@@ -373,10 +373,9 @@ SKIP: {
 		currency_symbol mon_decimal_point mon_thousands_sep
 		mon_grouping positive_sign negative_sign)) {
     SKIP: {
-	    skip("localeconv has no result for $_", 1)
-		unless exists $conv->{$_};
-	    unlike(delete $conv->{$_}, qr/\A\z/,
-		   "localeconv returned a non-empty string for $_");
+	    my $value = delete $conv->{$_};
+	    skip("localeconv '$_' may be empty", 1) if $_ ne 'decimal_point';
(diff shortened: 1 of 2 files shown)


tonycoz (approved):

tonycoz:
It looked ok, but then I noticed it's failing in CI.  From the log (there was more than this):
```
#   Failed test 'localeconv returned an integer for int_frac_digits'
#   at t/posix.t line 401.
#                   undef
#     doesn't match '(?^:\A-?\d+\z)'

#   Failed test 'localeconv returned an integer for frac_digits'
#   at t/posix.t line 401.
#                   undef
#     doesn't match '(?^:\A-?\d+\z)'

#   Failed test 'localeconv returned an integer for p_cs_precedes'
#   at t/posix.t line 401.
#                   undef
#     doesn't match '(?^:\A-?\d+\z)'
```
I wasn't able to reproduce it locally.

khwilliamson:
The problem was it still was omitting certain numeric fields, depending on the locale.  The CI box used a locale that manifested this; my local one did not

tonycoz (approved):

PR #20549, opened 2022-11-25
loc_tools.pl: Add ability to skip known bad locales

khwilliamson:
Some platforms have locales that shouldn't be used.  This adds code to
avoid using such when looking at all the locales on a platform.

The next commit will add the first use.

### DIFF
--- a/t/loc_tools.pl
+++ b/t/loc_tools.pl
@@ -17,7 +17,8 @@ use feature 'state';
 
 my %known_bad_locales = (   # XXX eventually will need version info if and
                             # when these get fixed.
-                        );
+    solaris => [ 'vi_VN.UTF-8', ],  # Use of U+A8 segfaults: GH #20578
+);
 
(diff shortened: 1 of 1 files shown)


khwilliamson:
@bram-perl Please review

bram-perl:
The code change looks acceptable but it would be good to alter/expand the second commit message.
Something like:
```
On OmniOS r151042 (based on Solaris) the `strfxm_l` function segfaults when:
- LC_COLLATE_MASK is set to 'vi_VN.UTF-8'
- the input contains "\xA8"
Other locales and other characters appear to be unaffected.
```
Personally I would also consider creating a GH issue for it and adding more info in that (for example the crashing perl-code example, the C reproducer, ...).

(Ideally I suppose we would avoid the segfault/provide a warning to the user/... but I don't think it's worth the effort since it appears to be very specific...)

PR #20552, opened 2022-11-27
perlsub: normalize whitespace to use 4 space indents

haarg:
The file was previously using a mix of indent styles. Some parts used 4 space indents, some used 4 space indents mixed with tabs. The file would not be properly viewable with a tab width other than 8. This is normal for many files in perl core, but they should all be fixed.

Normalize all of the indents to use 4 space indents consistently, with no tabs.

### DIFF
--- a/pod/perlsub.pod
+++ b/pod/perlsub.pod
@@ -8,32 +8,32 @@ perlsub - Perl subroutines
 To declare subroutines:
 X<subroutine, declaration> X<sub>
 
-    sub NAME;			  # A "forward" declaration.
-    sub NAME(PROTO);		  #  ditto, but with prototypes
-    sub NAME : ATTRS;		  #  with attributes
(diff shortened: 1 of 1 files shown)


Grinnz:
```
#   "Verbatim line length including indents exceeds 78 by" 
#     +*** (including 0 from =over's) near line ***824 of pod/perlsub.pod
```

demerphq:
I think if you want to reindent the verbatim stuff reindent it to consistent to 1 space. @khwilliamson has pointed out several times that to a pod scanner a 1 space indent is sufficient for rendering tools to identify the verbatim and do what they want, and it leaves more space for examples in the code.

haarg:
I'm not going to change the verbatim text to use 1 space indents, because that is painful to work with in an editor.

demerphq:
Ok.

Grinnz (approved):

jkeenan (approved):

PR #20558, opened 2022-11-29
Move assert from perl.c to perlmain.c and miniperlmain.c

### LINKED ISSUE #20557: BBC: Blead breaks MLEHMANN/Coro-6.57.tar.gz on debugging perls
**Description**

With v5.37.4-198-gdd66b1d793 tests started to fail for MLEHMANN/Coro-6.57.tar.gz on perls that have been built with -DEBUGGING=both. 

Sample fail report: http://www.cpantesters.org/cpan/report/46c3cf2e-6e27-11ed-9260-1e68f759ba8a

Failure diagnostics:
perl.c:2713: perl_run: Assertion `!PL_restartop' failed.

Bisect leads me to:

``` text
dd66b1d793c73fea9309c1d12a879369bf55bb83 is the first bad commit
commit dd66b1d793c73fea9309c1d12a879369bf55bb83
Date:   Wed Oct 12 18:50:25 2022 +0200

    pp_ctl.c - in try_yyparse do not leak PL_restartop from compile that dies

```

**Steps to Reproduce**

cpan -i MLEHMANN/Coro-6.57.tar.gz

**Expected behavior**

Should build, test and install the pkg

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 37 subversion 6) configuration:
  Commit id: d569e7cc1171d70b4ea183bc7c257dcf50f0ab12
  Platform:
    osname=linux
    osvers=5.19.0-2-amd64
    uname='linux k93msid 5.19.0-2-amd64 #1 smp preempt_dynamic debian 5.19.11-1 (2022-09-24) x86_64 gnulinux '
    config_args='-Dprefix=/home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.6/29fb -Dmyhostname=k93msid -Dinstallusrbinperl=n -Uversiononly -Dusedevel -des -Ui_db -Dlibswanted=cl pthread socket inet nsl gdbm dbm malloc dl ld sun m crypt sec util c cposix posix ucb BSD gdbm_compat -Duseithreads -Uuselongdouble -DEBUGGING=both'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2 -g'
    ccversion=''
    gccversion='12.2.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
    libc=/lib/x86_64-linux-gnu/libc.so.6
    so=so
    useshrplib=false
    libperl=libperl.a
    gnulibc_version='2.35'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'


  Compile-time options:
    DEBUGGING
    HAS_TIMES
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_HASH_FUNC_SIPHASH13
    PERL_HASH_USE_SBOX32
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_TRACK_MEMPOOL
    PERL_USE_DEVEL
    PERL_USE_SAFE_PUTENV
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
    USE_REENTRANT_API
    USE_THREAD_SAFE_LOCALE
  Built under linux
  Compiled at Nov 20 2022 13:54:18
  %ENV:
    PERL5LIB=""
    PERL5OPT=""
    PERL5_CPANPLUS_IS_RUNNING="2346572"
    PERL5_CPAN_IS_RUNNING="2346572"
    PERL_CANARY_STABILITY_NOPROMPT="1"
    PERL_MM_USE_DEFAULT="1"
    PERL_USE_UNSAFE_INC="1"
  @INC:
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.6/29fb/lib/site_perl/5.37.6/x86_64-linux-thread-multi
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.6/29fb/lib/site_perl/5.37.6
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.6/29fb/lib/5.37.6/x86_64-linux-thread-multi
    /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.37.6/29fb/lib/5.37.6
    .
```

demerphq:
The assert breaks Coro. Whether Coro is doing something dodgy is up to its author and future investigation, the assert was placed with an assumption that perl_run() would get called once after perl_parse(), and PL_restartop shouldnt leak, and Coro shows that isnt true, so moving the assert makes sense. 

Also adds miniperlmain.pl to regen.pl so that it gets run by make regen, it is fast so there is no point in it being manual.

This should fix #20557

### DIFF
--- a/ext/ExtUtils-Miniperl/lib/ExtUtils/Miniperl.pm
+++ b/ext/ExtUtils-Miniperl/lib/ExtUtils/Miniperl.pm
@@ -5,7 +5,7 @@ use Exporter 'import';
 use ExtUtils::Embed 1.31, qw(xsi_header xsi_protos xsi_body);
 
 our @EXPORT = qw(writemain);
-our $VERSION = '1.12';
+our $VERSION = '1.13';
 
 # blead will run this with miniperl, hence we can't use autodie or File::Temp
 my $temp;
@@ -135,8 +135,29 @@ main(int argc, char **argv, char **env)
 	PL_perl_destruct_level = 0;
     }
     PL_exit_flags |= PERL_EXIT_DESTRUCT_END;
-    if (!perl_parse(my_perl, xs_init, argc, argv, (char **)NULL))
+    if (!perl_parse(my_perl, xs_init, argc, argv, (char **)NULL)) {
+
+        /* perl_parse() may end up starting its own run loops, which
+         * might end up "leaking" PL_restartop from the parse phase into
+         * the run phase which then ends up confusing run_body(). This
+         * leakage shouldn't happen and if it does its a bug.
+         *
+         * Note we do not do this assert in perl_run() or perl_parse()
+         * as there are modules out there which explicitly set
+         * PL_restartop before calling perl_run() directly from XS code
+         * (Coro), and it is conceivable PL_restartop could be set prior
+         * to calling perl_parse() by XS code as well.
+         *
+         * What we want to check is that the top level perl_parse(),
+         * perl_run() pairing does not allow a leaking PL_restartop, as
+         * that indicates a bug in perl. By putting the assert here we
+         * can validate that Perl itself is operating correctly without
+         * risking breakage to XS code under DEBUGGING. - Yves
+         */
+        assert(!PL_restartop);
+
         perl_run(my_perl);
+    }
 
 #ifndef PERL_MICRO
     /* Unregister our signal handler before destroying my_perl */
--- a/miniperlmain.c
+++ b/miniperlmain.c
@@ -109,8 +109,29 @@ main(int argc, char **argv, char **env)
 	PL_perl_destruct_level = 0;
     }
     PL_exit_flags |= PERL_EXIT_DESTRUCT_END;
-    if (!perl_parse(my_perl, xs_init, argc, argv, (char **)NULL))
+    if (!perl_parse(my_perl, xs_init, argc, argv, (char **)NULL)) {
+
+        /* perl_parse() may end up starting its own run loops, which
+         * might end up "leaking" PL_restartop from the parse phase into
+         * the run phase which then ends up confusing run_body(). This
+         * leakage shouldn't happen and if it does its a bug.
+         *
+         * Note we do not do this assert in perl_run() or perl_parse()
+         * as there are modules out there which explicitly set
+         * PL_restartop before calling perl_run() directly from XS code
+         * (Coro), and it is conceivable PL_restartop could be set prior
+         * to calling perl_parse() by XS code as well.
+         *
+         * What we want to check is that the top level perl_parse(),
+         * perl_run() pairing does not allow a leaking PL_restartop, as
+         * that indicates a bug in perl. By putting the assert here we
+         * can validate that Perl itself is operating correctly without
+         * risking breakage to XS code under DEBUGGING. - Yves
+         */
+        assert(!PL_restartop);
+
         perl_run(my_perl);
+    }
 
 #ifndef PERL_MICRO
     /* Unregister our signal handler before destroying my_perl */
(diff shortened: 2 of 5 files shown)


jkeenan:
My testing indicates that this fixes the breakage in Coro in debugging builds.

tonycoz:
If this is a post-condition for perl_parse(), wouldn't it be better for perl_parse() to assert() before returning?

Especially since any perl embedding (per perlembed) won't include this assert.

(This is an invitation to comment, not a rejection.)

khwilliamson (approved):

demerphq:
@tonycoz I tried to answer that in the last commits message:

In https://github.com/Perl/perl5/commit/dd66b1d793c73fea9309c1d12a879369bf55bb83 we added an assert to perl_run() that PL_restartop should
never be true when perl_run() is called after perl_parse(). Looked at
from the point of the internals, which calls perl_parse() and perl_run()
exactly once, this made sense.

It turns out however that there is at least one XS module out there that
expects to be able to set PL_restartop and then call perl_run(). If that
works out for them then we shouldn't block it, as we aren't really
trying to say "perl_run() should never be called with PL_restartop set"
(at least this assert wasn't trying to say that really), we are trying
to assert "between the top level transition from perl_parse() to
perl_run() we shouldnt leak any PL_restartop".

One could argue the assert maybe should go at the end of perl_parse(),
but I chose to put it in Miniperl.pm and thus into perlmain.c and
miniperlmain.c as I am not certain that perl_parse() should never be
called with PL_restartop set already, and putting it in the main code
really does more closely reflect the intent of this assert anyway.

demerphq:
@tonycoz fwiw, I dont mind moving it to perl_parse(). I just have no more confidence that someone is setting PL_restartop before calling perl_parse() than I do that someone is calling perl_run(). And really the intent of the assert so that in our tests, if we break exception handling during the compile phase that the assert triggers and the situation is made clear to the developer, as the result can be really confusing. I am not sure that we need to check this in embedded builds and what not, I think if our tests pass with this assert under debugging we are good to go.

But as I said, if you would like to find out if it is safe to put it in perl_parse(), I do agree it makes some sense. However if some XS code calls perl_parse() inside of an eval, then PL_restartop would be set before it was entered I think, so IMO there is a small risk that there will be some Coro style case that might also break. Since I just want to validate that this never happens during our tests in the "standard model" it seems the right approach.

demerphq:
For now I'm gunna merge.

PR #20568, opened 2022-12-01
locale.c: Add missing mutex locks

khwilliamson:
This adds locks around mbrtowc and reading the environment that a code review indicated were missing.  The mbrtowc lock is currently a no-op, but a future commit will change that.  The other lock is in code rarely compiled.


demerphq (approved):

jkeenan:
This p.r., opened on Dec 01, has 1 Approved code review but was subsequently re-classified as Draft.  Can we get some clarification as to its status?  Thanks.

khwilliamson:
Will add this later in a different PR

PR #20569, opened 2022-12-01
locale.c: Prepare for and  Rewrite localeconv() handling

khwilliamson:
localeconv() returns a structure contaiing fields that are associated
     with two different categories: LC_NUMERIC and LC_MONETARY.  Perl via
     POSIX::localeconv() reutrns a hash containing all the fields.
     
     Testing on Windows showed that if LC_CTYPE is not the same locale as
     LC_MONETARY for the monetary fields, or isn't the same as LC_NUMERIC for
     the numeric ones, mojibake can result.
     
     The solution to similar situations elsewhere in the code is to toggle
     LC_CTYPE into being the same locale as the one for the returned fields.
     But those situations only have a single locale that LC_CTYPE has to
     match, so it doesn't work here when LC_NUMERIC and LC_MONETARY are
     different locales.   Unlike Schrödinger's cat, LC_CTYPE has to be one or
     the other, not both at the same time.
     
     The previous implementation did not consider this possibility, and
     wasn't easily changeable to work.
     
     Therefore, this rewrites a bunch of it.  The solution used is to call
     localeconv() twice when the LC_NUMERIC locale and the LC_MONETARY locale
     don't match (with LC_CTYPE toggled to the corresponding one each time).
     (Only one call is made if the two categories have the same locale.)
     This one vs two complicated the code, but I thought it was worth it
     given that the one call is the most likely case.
     
     Another complication is that on platforms that lack nl_langinfo(),
     (Windows, for example), localeconv() is used to emulate portions of it.
     Previously there was a separate function to handle this, using an SV()
     cast as an HV() to avoid using a hash that wasn't actually necessary.
     That proved to lead to extra duplicated code under the new scheme, so
     that function was collapsed into a single one and a real hash is used in
     all circumstances, but is only populated with the one or two fields
     needed for the emulation.
     
     The only part of this commit that I thought could be split off from the
     rest concerns the fact that localeconv()'s return is not thread-safe,
     and so must be copied to a safe place (the hash) while in a critical
     section, locking out all other threads.  Before this commit, that
     copying was accompanied by determining if each string field needed to be
     marked as UTF-8.  That determination isn't necessarily trivial, so
     should really not be in the critical section.  This commit does that.
     And, with some effort, that part could have been split into a separate
     commit. but I didn't think it was worth the effort.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -3336,6 +3336,15 @@ S	|utf8ness_t|get_locale_string_utf8ness_i			\
 				|NULLOK const char * locale		\
 				|const unsigned cat_index
 S	|bool	|is_locale_utf8	|NN const char * locale
+#  ifdef HAS_LOCALECONV
+S	|HV *	|my_localeconv	|const int item
+S	|void	|populate_hash_from_localeconv				\
+				|NN HV * hv				\
+				|NN const char * locale			\
+				|const U32 which_mask			\
+                                |NN const lconv_offset_t * strings[2]	\
+				|NULLOK const lconv_offset_t * integers
+#  endif
 #  ifdef USE_LOCALE
 iR	|const char *|mortalized_pv_copy|NULLOK const char * const pv
 ST	|const char *|save_to_buffer|NULLOK const char * string	\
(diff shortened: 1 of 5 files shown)


demerphq (approved):

jkeenan:
Will this patch address the BBC breakage associated with `localeconv` and reported in https://github.com/Perl/perl5/issues/20571?

khwilliamson:
No it won't.  That is orthogonal to this.  I'm working on that now

Issue #20574, opened 2022-12-02
tied FETCH unexpectedly invoked twice when returning an object with overload


leonerd:
The following code unexpectedly invokes the tied `FETCH` function twice on every iteration line. If instead the code returned a non-overloaded object then this only happens once.

```
use v5.14;
use warnings;

package BoldStars {
    use overload
        '""' => sub {
            my $self = shift;               ## <-- THIS is line 7
            my $str = $self->[0];
            # Wrap every word in *stars*
            return $str =~ s/(\S+)/*$1*/gr;
        },
        fallback => 1;
    sub new {
        my $class = shift;
        my $self = bless [ $_[0] ], $class;
        return $self;
    }
}

package TiedCounter {
    use Carp;

    my $dashes = "";
    sub TIESCALAR {
        my $class = shift;
        return bless [], $class;
    }
    sub FETCH {
        carp "FETCHing TiedCounter";
        $dashes .= "-";
        return BoldStars->new( $dashes );
    }
}

tie my $tiecounter, "TiedCounter";

say $tiecounter for 1 .. 5;                 ## <-- THIS is line 37
```

I'd expect it to print 5 rows of output with each one having *one* more dash than the prior:

```
*-*
*--*
*---*
*----*
*-----*
```

but instead it prints:

```
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
FETCHing TiedCounter at tie-fetch-twice.pl line 7.
*--*
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
FETCHing TiedCounter at tie-fetch-twice.pl line 7.
*----*
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
FETCHing TiedCounter at tie-fetch-twice.pl line 7.
*------*
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
FETCHing TiedCounter at tie-fetch-twice.pl line 7.
*--------*
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
FETCHing TiedCounter at tie-fetch-twice.pl line 7.
*----------*
```

The `FETCH` function really was run twice for each call. Each row has two more dashes than the prior. The `FETCH` call from line 37 was expected, the one from lines 7 was most certainly not.

If you remove the `use overload ...` declaration then this goes away and only runs once (and overloading no longer happens of course):

```
FETCHing TiedCounter at tie-fetch-twice.pl line 37.
BoldStars=ARRAY(0x556a4784bf70)
[...]

PR #20575, opened 2022-12-03
Fix double FETCH from tied overloaded scalar

demerphq:
Use a copy of the tied sv in the case where the result is an overloaded object.

This likely breaks assignment to $_[0] in overloaded methods. Whether that is a bad thing is debatable. 

See: https://github.com/Perl/perl5/issues/20574

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -395,6 +395,12 @@ Also, embedded C<NUL> characters are now allowed in the input.
 If locale collation is not enabled on the platform (C<LC_COLLATE>), the
 input is returned unchanged.
 
+=item *
+
+Double FETCH during stringification of tied scalars returning an
+overloaded object have been fixed. The FETCH method should only be
+called once, but prior to this release was actually called twice.
+
 =back
 
 =head1 Known Problems
--- a/sv.c
+++ b/sv.c
@@ -2830,21 +2830,28 @@ char *
 Perl_sv_2pv_flags(pTHX_ SV *const sv, STRLEN *const lp, const U32 flags)
 {
     char *s;
+    bool done_gmagic = FALSE;
 
     PERL_ARGS_ASSERT_SV_2PV_FLAGS;
 
     assert (SvTYPE(sv) != SVt_PVAV && SvTYPE(sv) != SVt_PVHV
          && SvTYPE(sv) != SVt_PVFM);
-    if (SvGMAGICAL(sv) && (flags & SV_GMAGIC))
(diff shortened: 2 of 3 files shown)


leonerd:
I'm happy that fix looks to be doing the right thing.

The one test that it breaks honestly just looks like "eh, we observe it runs FETCH twice in this situation" without really justifying why it should. The test simply asserts current behaviour without rationale. I vote it quite reasonable to change that test.

I do wonder if we should put a little note in the perldelta though, just on the slightest offchance someone happened to be relying on a double-FETCH in this circumstance.

demerphq:
>I do wonder if we should put a little note in the perldelta though, just on the slightest offchance someone happened to be relying on a double-FETCH in this circumstance.

Yeah, this should go in the bug fix section. Ill take care of it.

demerphq:
I have repushed, but i have second thoughts. What do we expect to happen when someone writes to `$_[0]` in the overload callback? If it should trigger a store on the object then this patch really is the wrong thing. We should have a flag that get magic can use that say `SvGETMAGIC(sv)` was called, and does not need to be called again, and the code should stay unchanged and use the flag to prevent the double call.

Having said that, maybe this is the best thing we can do until we can figure out such a flag.

pod/perldelta.pod
```diff
@@ -395,6 +395,12 @@ Also, embedded C<NUL> characters are now allowed in the input.
 If locale collation is not enabled on the platform (C<LC_COLLATE>), the
 input is returned unchanged.
 
+=item *
+
+Double FETCH during strigification of tied scalars returning an
```
tonycoz: strigification?

sv.c
```diff
@@ -2830,21 +2830,28 @@ char *
...
+
     if (SvROK(sv)) {
         if (SvAMAGIC(sv)) {
             SV *tmpstr;
+            SV *nsv= (SV *)sv;
             if (flags & SV_SKIP_OVERLOAD)
                 return NULL;
-            tmpstr = AMG_CALLunary(sv, string_amg);
+            if (done_gmagic)
+                nsv = sv_2mortal(newSVsv_nomg(sv));
```
tonycoz: This could be `nsv = sv_mortalcopy_flags(sv, 0);`

pod/perldelta.pod
demerphq: thanks, fixed.

sv.c
demerphq: Thanks, done.

leonerd (approved):
LGTM

PR #20576, opened 2022-12-03
dist/PathTools - fix possible VMS breakage

demerphq:
In #19645 I added a fix for empty paths under taint mode. This fix may have changed VMS behavior, which this patch should fix.


hvds:
Thanks, it looks to me like this correctly restores the previous VMS behaviour.

Issue #20577, opened 2022-12-03
When @INC contains a blessed object, modifying %INC may modify @INC


djerius:
**Description**
An object in `@INC` is used to load packages from a hash (i.e. this is a fat-packed executable).  The main code loads `Test2` which loads `Test1`.  `Test1` modifies `%INC` in a particular fashion, which somehow removes the object from `@INC`.  This causes further hijinks and hilarity when the rest of the code no longer can be loaded via the `@INC` hook.

**Steps to Reproduce**

Run [bug.txt](https://github.com/Perl/perl5/files/10147128/bug.txt) as

`SHOW_BUG=1 perl bug.txt`

The output is:

```
$ SHOW_BUG=1 perl bug.txt
@INC is intact
@INC is modified
```
**Expected behavior**

Run the same code as
`SHOW_BUG=0 perl bug.txt`

The output is:

```
$ SHOW_BUG=0 perl bug.txt
@INC is intact
@INC is intact
```

**Perl configuration**

I see this under Perl v5.28.3 as well as v5.36.0 as listed below.

```
Site configuration information for perl 5.36.0:

Configured by dj at Tue May 31 10:48:33 EDT 2022.

Summary of my perl5 (revision 5 version 36 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=5.10.0-14-amd64
    archname=x86_64-linux
    uname='linux leafhopper 5.10.0-14-amd64 #1 smp debian 5.10.113-1 (2022-04-29) x86_64 gnulinux '
    config_args='-Dprefix=/home/dj/.plenv/versions/5.36.0 -de -Dversiononly -A'eval:scriptdir=/home/dj/.plenv/versions/5.36.0/bin''
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2'
    optimize='-O2'
    cppflags='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='10.2.1 20210110'
[...]

PR #20579, opened 2022-12-04
pp_ctl.c - copy hook into %INC not alias it

demerphq:
I am still not entirely sure how the bug reported in GH #20577 happens, but it is clear it is related to this logic storing the hook into the %INC hash as an alias of the sv it got from @INC. Changing this to be a copy makes the problem go away and seems to be more reasonable.

The fact we store the hook into %INC at all seems to be undocumented, and seems to be of limited utility.

This should fix GH #20577.


atoomic (approved):

demerphq:
Rewrote commit message, and updated test to skip miniperl.

PR #20581, opened 2022-12-04
pod2html:  Group '*' and 'no*' switches in documentation

jkeenan:
As the next step toward unifying the documentation in bin/pod2html and lib/Pod/Html.pm, group the '*' and 'no*' command-line switches to the pod2html utility and then adjust the descriptions of the switches accordingly.  There are 5 such instances (example: 'poderrors' and 'nopoderrors') already grouped in this manner in lib/Pod/Html.pm.  This commit brings this grouping into bin/pod2html as well.

No changes in functionality.

### DIFF
--- a/ext/Pod-Html/bin/pod2html
+++ b/ext/Pod-Html/bin/pod2html
@@ -28,15 +28,10 @@ pod2html takes the following arguments:
 =item backlink
 
   --backlink
-
-Turn =head1 directives into links pointing to the top of the HTML file.
-
-=item nobacklink
-
   --nobacklink
 
-Do not turn =head1 directives into links pointing to the top of the HTML file
-(default behaviour).
+Turn =head1 directives into links pointing to the top of the HTML file.
+--nobacklink (which is the default behavior) does not create these backlinks.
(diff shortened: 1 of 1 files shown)


ext/Pod-Html/bin/pod2html
```diff
@@ -155,27 +145,19 @@ Specify the base directory for finding library pods.
...
-
-=item noquiet
-
   --noquiet
 
-Display mostly harmless warning messages (default behaviour). But this is not
-the same as "verbose" mode.
+Don't display mostly harmless warning messages.  --noquiet -- which is the
+default behavior does not display these mostly harmless warning messages (but
+this is not the same as "verbose" mode).
```
tonycoz: The reverses the documented meaning of --noquiet.

jkeenan: Thanks for spotting that.  Should be corrected in latest commit.

tonycoz (approved):

carlbigdata (commented):
I have reviewed, looks good

PR #20588, opened 2022-12-06
pp_ctl.c - dedupe @INC entries, especially hooks.

demerphq:
It is all too easy to write code that will OOM with the current `@INC` hook mechanism. Consider this:

```
unshift @INC, sub { push @INC, "." };
```

this will run until it OOM's, which i believe is an untrappable and difficult error to deal with.

I think it should be harder to do this accidentally, and that a simple solution is to ensure that a given hook executes only once. I do not think this is much of a burden, as the array hook api offers a simple way to wrap hooks so that they get called an arbitrary number of times. (But not infinitely many times.)

Yes it is true that someone can write `1 while 1`; in a hook, but they are likely to do so accidentally whereas creating an OOM inducing infinite loop with an @INC hook is. 

In my experimentation with this in place, I have it makes writing `@INC` hooks which rewrite `@INC` easier, as you do not have to defend against the double call inside of the `@INC` hook. On the other hand it is still possible to intentionally cause a hook to be called more than once, by wrapping or currying.

See discussion in https://github.com/Perl/perl5/pull/20547 where @haarg and @ap have expressed some disagreement about whether this is a good thing, this patch was originally part of that PR and has been separated out. This patch is intended to be applied on top it as well.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6963,6 +6963,20 @@ or:
         ...
     }
 
+Note, that a given hook will be executed only once, no matter how many
+times it is pushed into C<@INC>. This is a precaution to make it harder
+to create infinite loops in the require machinery. For anyone deliberately
+wanting to do this, you can use the ARRAY hook representation to pass in
+the same object or sub more than once, or you use currying with a sub:
+
+    my $callback = sub { warn "In \@INC position $INC" };
+    @INC = map { sub{ $callback->() }, $_ } @INC; # curry callback
+    #or
+    @INC = map { [ $callback ], $_ } @INC;    # wrap in ARRAY
+
+would set up inc so the same callback was called before every existing
+lookup the C<@INC> array.
+
 These hooks are also permitted to set the L<C<%INC>|perlvar/%INC> entry
 corresponding to the files they have loaded.  See L<perlvar/%INC>.
 Should an C<INC> hook not do this then perl will set the C<%INC> entry
@@ -7012,6 +7026,15 @@ will be unrolled after the hook exits, and its value only has meaning
 immediately after execution of the hook, thus setting C<$INC> to some
 value prior to executing a C<require> will have no effect on how the
 require executes at all.
+
+As of 5.37.7 any given hook in the C<@INC> array no matter how it got
+there will be executed at most once, the deduping is done using the
+refaddr of the hook and cannot be bypassed with overloads (although the
+array interface makes it easy to create multiple distinct copies if you
+really really want to do this). Likewise any given textually distinct
+directory will be inspected only once. Eg, "foo/../bar/" and "bar/"
+would be treated as distinct, but inserting "foo/" twice would not result
+in us checking the filesystem twice.
 
 As of 5.37.7 C<@INC> values of undef will be silently ignored.
 
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4271,6 +4271,7 @@ S_require_file(pTHX_ SV *sv)
             namesv = newSV_type(SVt_PV);
             AV *inc_ar = GvAVn(PL_incgv);
             SSize_t incdir_continue_inc_idx = -1;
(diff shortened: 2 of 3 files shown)


Tux:
I am ++ to de-dup adjacent entries, but oposed to de-dup entries that have a different call in between.

```
@INC = ((grep { lenght && -d } split m/:/ => $ENV{AINC}), \&hook,
        ((grep { lenght && -d } split m/:/ => $ENV{BINC}), \&hook);
```
where `hook` can do checks on changes since its previous call.
I base my opinion on how perl's `Configure` uses callback units

haarg:
I still don't like this. I think it's fixing a problem that doesn't exist, while making the model for `@INC` hooks more complex and harder to reason about.

demerphq:
@haarg - have you actually tried using it? My experience with playing with it and without it, is that it makes it *easier* to reason about, and much easier to write hooks that rewrite `@INC` safely.  Various operations that without it result in infinite loops or other bad things suddenly become safe and easy. And it doesn't stop a determined and clever person doing whatever they want. 

I would really like to see an example of useful code that was harder with this than it is without it. I am pretty confident that many things will be simpler with it, and the rest will be no more complex.

demerphq:
@Tux - this patch doesn't prevent you from doing what you want, but you would need to spell it differently:

```
@INC = ((grep { lenght && -d } split m/:/ => $ENV{AINC}), [\&hook],
        ((grep { lenght && -d } split m/:/ => $ENV{BINC}), [\&hook]);
```
or
```
@INC = ((grep { lenght && -d } split m/:/ => $ENV{AINC}), sub{ hook(@_) },
        ((grep { lenght && -d } split m/:/ => $ENV{BINC}), sub { hook(@_) });
```

So the underlying `hook()` sub will be called twice, but each hook entry in `@INC`  processed only once, **no matter what happens to the `@INC` array during the lookup**. Consider what happens if there is another hook in place which reorders `@INC`, would you expect hook() to be called potentially an infinite number of times without your knowledge?

Consider what happens if someone does this:

```
push @INC, sub { unshift @INC, "." };
```

it will execute until it runs out of memory, and then fail in a way that cant be trapped by perl. I think *that* is way too easy to do via the `@INC` api.

BTW, I think it is worth mentioning that in a typical piece of perl code that tried to do this, the need for a guard would be eliminated as we would copy the array to the stack, eg if we did a perl level for loop

```
foreach my $inc_entry (@INC) {
}
```

so hooks modifying `@INC` wouldn't be a problem, and there would always be a bounded number of times that a hook could execute (as it would be executed how many times it was @INC at most). But in the internals we are using a C for loop on the *actual* array, not a copy on the stack.  This is nice because we can add items to the list during `@INC`'s lifetime, but it also means that something in normal circumstances would be a bounded operation is actually an unbounded operation. This logic is involved in loading error reporting modules, and the like, and the failure modes are horrible, C level infinite loop while slowly chewing up all ram available to the process (potentially all ram on the box). Which means you can't trap it, it might lead to overall system stability issues, and it is a nightmare to debug. 

All I have done with this restriction is try to get us to have most of the cake and eat most of the cake. We can do in flight modifications to `@INC`, and we can expect bounded behavior and run time for loading a module.  It behaves like the `for ()` loop equivalent would, more or less, but also allows new stuff to be added to the list of things to be processed or checked. IMO in any *sane* production code of a similar nature this kind of guard would be in place.  It is only because this logic is in a programming language where you can blow out memory or infinite loop in all sorts of interesting ways that it is even debatable that we should not prevent these cases. But my view is doing those tends to require deliberate effort, and most of programming is about writing programs that have bounded run times, so having an easy way to create an OOM bomb in the language in a place where it is not obvious that you have some so, and where the action is at a distance, is something we should eliminate.

@ap @haarg ^^^ i think the point above about how "normal" code wouldn't have this is pretty relevant to the expectations discussions. For instance, now that we have `INCDIR` we could make it so we copy `@INC` at the start, and modifications are only visible in the next call. That would also eliminate the OOM infinite loop trap nicely. (I think this patch is a better approach, but whatever.)

PR #20589, opened 2022-12-06
makedef.pl: Consistently use constant, make consistent with duplicated parts of perl.h

khwilliamson:
This script defines a constant 'PLATFORM', but often uses the rhs of
that constant.  Change to always use the constant.

The second commit changes to use the same paradigm  as perl.h for the part that has to manually be synced with that file


demerphq (approved):

demerphq:
I merged as i noticed this issue myself. I am doing some significant changes to the regen/ tree, and also to makedef.pl, see 

https://github.com/Perl/perl5/pull/20516

PR #20592, opened 2022-12-06
Define a `newPADxVOP()` convenience function

leonerd:
Avoids having to create a `OP_PAD{SV,AV,HV,CV}` into a temporary variable just to set the pad index of it before building it into a larger optree fragment. This can be written more neatly inline.


tonycoz (approved):
I do wonder if this would be better as an inline function, since the body is so trivial.

The code size with the call to this function vs the inlined call is [about the same](https://godbolt.org/z/3TaPGrdP7).

leonerd:
Given how tiny it is, it could well be inline. I've pushed another commit to test that out. I'll squash it if it works.

PR #20597, opened 2022-12-10
libnet: remove CUSTOMIZED entries and properly sync with CPAN

haarg:
The customization in core was included in libnet 3.14, so the CUSTOMIZED entries aren't needed. However, they weren't removed before 3.14 was synced to core.

Remove the CUSTOMIZED entries and properly sync the changes from 3.14.

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -711,23 +711,6 @@ use File::Glob qw(:case);
             qr(^demos/),
             qr(^t/external/),
         ],
-         # https://github.com/steve-m-hay/perl-libnet/pull/45
-        'CUSTOMIZED' => [
-            'lib/Net/Cmd.pm',
-            'lib/Net/Config.pm',
-            'lib/Net/Domain.pm',
-            'lib/Net/FTP.pm',
-            'lib/Net/FTP/A.pm',
-            'lib/Net/FTP/E.pm',
-            'lib/Net/FTP/I.pm',
-            'lib/Net/FTP/L.pm',
-            'lib/Net/FTP/dataconn.pm',
-            'lib/Net/NNTP.pm',
-            'lib/Net/Netrc.pm',
-            'lib/Net/POP3.pm',
-            'lib/Net/SMTP.pm',
-            'lib/Net/Time.pm',
-        ],
     },
 
     'Locale::Maketext' => {
--- a/cpan/libnet/lib/Net/Cmd.pm
+++ b/cpan/libnet/lib/Net/Cmd.pm
@@ -1,7 +1,7 @@
 # Net::Cmd.pm
 #
 # Copyright (C) 1995-2006 Graham Barr.  All rights reserved.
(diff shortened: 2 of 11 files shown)


jkeenan:
Since this p.r. includes changes in `*.pm` files, we need to have `$VERSION` implemented therein to keep `t/porting/cmp_version.t` happy.

haarg:
I'm not sure what the appropriate solution to that is in this case. The previous version was set incorrectly. cpan modules that are modified in core are meant to use underscores to indicate that they are modified. This PR fixes the files to properly match the CPAN release for the version given. Using a different version number doesn't seem appropriate.

Leont:
`sync-to-cpan` really should come with a warning with regard to customized files, that would have prevented this from happening.

jkeenan:
> I'm not sure what the appropriate solution to that is in this case. The previous version was set incorrectly. cpan modules that are modified in core are meant to use underscores to indicate that they are modified. This PR fixes the files to properly match the CPAN release for the version given. Using a different version number doesn't seem appropriate.

I would argue that the benefit of getting your work eliminating the CUSTOMIZED entry in `Porting/Maintainers.pl` outweighs the annoyance of having `$VERSION` in the `.pm` files underneath `cpan/libnet/` ahead of the corresponding files on CPAN by `_001`.

If @steve-m-hay were to have occasion to issue a new CPAN release sometime between now and May, then we could synch that version into the core prior to 5.38.0 and the temporary $VERSION difference would be cured.

If that makes sense, we can pick up this patch:  http://thenceforward.net/perl/misc/0001-cpan-libnet-Bump-VERSION-to-3.14_001-in-.pm-files.patch

jkeenan:
> > I'm not sure what the appropriate solution to that is in this case. The previous version was set incorrectly. cpan modules that are modified in core are meant to use underscores to indicate that they are modified. This PR fixes the files to properly match the CPAN release for the version given. Using a different version number doesn't seem appropriate.
> 
> I would argue that the benefit of getting your work eliminating the CUSTOMIZED entry in `Porting/Maintainers.pl` outweighs the annoyance of having `$VERSION` in the `.pm` files underneath `cpan/libnet/` ahead of the corresponding files on CPAN by `_001`.
> 
> If @steve-m-hay were to have occasion to issue a new CPAN release sometime between now and May, then we could synch that version into the core prior to 5.38.0 and the temporary $VERSION difference would be cured.
> 
> If that makes sense, we can pick up this patch: http://thenceforward.net/perl/misc/0001-cpan-libnet-Bump-VERSION-to-3.14_001-in-.pm-files.patch

@haarg, are you willing to accept the proposal above?  Otherwise, should we just wait for a new version of libnet to be released upstream on CPAN and deal with sync-ing then?  (Also, there are some merge conflicts now.)  Thanks.

demerphq:
This was resolved as part of the 5.37.10 release that I did. Steve Hay rolled a new libnet release, and I removed the customized entries. See 6755d87e212e8b23e6975cb6ed741f3cd84e16b3

I am closing this as it is already done.

PR #20598, opened 2022-12-10
Define OP_HELEMEXISTSOR, a handy LOGOP shortcut for HELEM existence tests

leonerd:
This op is constructed using an OP_HELEM as the op_first and any scalar expression as the op_other.

It is roughly equivalent to the following perl code:

```perl
exists $hv{$key} ? $hv{$key} : OTHER
```

except that the HV and the KEY expression are evaluated only once, and only one hv_* function is invoked to both test and obtain the value. It is therefore smaller and more efficient.

Likewise, adding the OPpHELEMEXISTSOR_DELETE flag turns it into the equivalent of

```perl
exists $hv{$key} ? delete $hv{$key} : OTHER
```

It is not currently exposed via operator syntax to pure perl code. It exists purely to support an upcoming core perl feature, though it could be useful for XS modules too. It could be exposed as perl syntax at some future time, if it was thought to be useful and a suitable spelling for the operator name was found.

An as-yet-unanswered question is whether core should (yet) contain any tests for this op. I happen to know it works because it is heavily exercised in a lot of my `feature-class` branch tests, but until that branch gets merged there's no other in-core testing of this op alone. I don't feel it should be added to XS::APITest because it's not really XS API, as such. We don't have any other opcode tests in there. But then I don't know where else to test it from in that case.


leonerd:
A proposed core use for this can be found later in my `feature-class` branch:

https://github.com/leonerd/perl5/blob/feature-class/class.c#L765-L772

A similar use-case in an XS module that inspired this:

https://metacpan.org/release/PEVANS/Object-Pad-0.74/source/src/field.c#L329-343

richardleach:
Bikeshedding, but should it be something like `OP_HVEXISTSOR` if it only works on HVs? (Especially if e.g. it turns out to be super useful and then people decide they want an AV variant as well.)

leonerd:
Oh, probably yes. I was going to say "but OP_EXISTS only works on HVs" but then of course I often forget that no - it works on AVs too. Much to everyone's annoyance.
OK, so better possible names: `OP_HVEXISTSOR`? But then it's not the HV we're checking if it exists, but an element inside it. That might suggest `OP_HELEMEXISTSOR`. Perhaps on the long side, but it's not really a "user-facing" op so the name length doesn't matter. There's precedent in long names from things like `OP_AELEMFASTLEX_STORE` and `OP_METHOD_REDIR_SUPER` anyway so this seems fine.

I'll rename it thusly.

leonerd:
Righty. Renamed, force-pushed, etc.

I can't rename the branch itself because github doesn't let you change the branch name of an open PR, but we can ignore that. :)

pp.c
```diff
@@ -5380,6 +5380,31 @@ PP(pp_exists)
     RETPUSHNO;
 }
```
wolfsage, line 5382: Some documentation above this about how to use it and its limitations would be nice. It seems to me that this won't handle magic in an expected way. The commit says it's roughly equivalent to `exists .... ? ... : ... `but I don't think EXISTS magic would ever be called since the code is doing just a fetch or a delete.

wolfsage (commented):
How will B::Concise, etc, handle these optrees? (Or is that to come or ..?)

pp.c
leonerd: Oooh, fun. I hadn't thought about magic (well, overload) since in my use-case I know that the HV is a plain one because it's internal to the surrounding code. But yes, perhaps an overload check for that and calling something else on it would be useful indeed.

leonerd:
@wolfsage 
> How will B::Concise, etc, handle these optrees? (Or is that to come or ..?)

These ops have normal names so `B::Concise` will be just fine with it. You possibly meant to ask about `B::Deparse` - in that case it would get upset because it doesn't know how to print it. I did think about that, but that's already the case with quite a few optree shapes that you couldn't possibly have compiled out of regular Perl code, but would only have made with C code calling `newOP...` directly. The same is true here.

In particular, since there isn't a pureperl operator having the same semantics (because it avoids multiple evaluation of the HV and key), there isn't really any pureperl code that `B::Deparse` could emit for this. I don't currently have a good feel for how we want to answer that.

wolfsage:
> In particular, since there isn't a pureperl operator having the same semantics (because it avoids multiple evaluation of the HV and key), there isn't really any pureperl code that `B::Deparse` could emit for this. I don't currently have a good feel for how we want to answer that.

Yeah, I suppose if this is being used in feature-class, and feature-class deparses sensibly, then that's Good Enough

leonerd:
@wolfsage 
> feature-class deparses sensibly, then that's Good Enough

Heh; that doesn't really apply though. :/ `B::Deparse` operates at the granularity of a single code reference. So it could be called upon to deparse an individual `method` out of a class. But methods don't really make sense on their own without the fields they capture. Already `B::Deparse` is known to make a mess of lexical closures so this is really just a worse variation of that. We don't *have* a thing to deparse an entire class, much like we don't have one for an entire package now.

leonerd:
Heh. Turns out that in the process of trying to write an external test for this by using XS::Parse::Infix, I discover a) that the core pluggable infix operators don't have precedence levels for logical operators, and b) this code doesn't work because it's missing a `PUTBACK`. So I'll draft this for now until I have some testing.

leonerd:
OK bugs fixed, docs and magic handling added.

pp.c
wolfsage: Looks like you fixed this up. Nice. It would be cool to document how/when various magic would be called so it's not a surprise, but I think this is fine as is if you don't want to.

I think it's something like:

If the hash is magical, this will invoke EXISTS on it. If the entry is found and the OPpHELEMEXISTSOR_DELETE flag is true, DELETE will be invoked. Otherwise FETCH will be.

leonerd: Added extra paragraph about tied hashes.

wolfsage: Perfect, thanks

leonerd:
Rebased and ready to merge. Any more comments anyone?

leonerd:
@wolfsage, @richardleach - this good now?

wolfsage (approved):
seems good to me though I have not tested the code at all.

PR #20599, opened 2022-12-10
gv.c - move English.pm into gv_fetchpvn. See RFC 0014

demerphq:
This implements the English name aliases for the magic punctuation variables listed in English.pm

It does not include the "strict" part of the RFC.

Currently no tests. Hoping @haarg or someone else might help contribute some.

### DIFF
--- a/caret_vars.inc
+++ b/caret_vars.inc
@@ -1,0 +1,742 @@
+case 1:
+    switch ((U8)name[0]) {
+    case 1:
+        /* $^A */
+        /* FALLTHROUGH */;
+    case 3:
+        /* $^C */
+        /* FALLTHROUGH */;
+    case 4:
+        /* $^D */
+        /* FALLTHROUGH */;
+    case 5:
+        /* $^E */
+        /* FALLTHROUGH */;
+    case 6:
+        /* $^F */
+        /* FALLTHROUGH */;
+    case 8:
+        /* $^H */
+        /* FALLTHROUGH */;
+    case 9:
+        /* $^I */
+        /* FALLTHROUGH */;
+    case 12:
+        /* $^L */
+        /* FALLTHROUGH */;
+    case 13:
+        /* $^M */
+        /* FALLTHROUGH */;
+    case 14:
+        /* $^N */
+        /* FALLTHROUGH */;
+    case 15:
+        /* $^O */
+        /* FALLTHROUGH */;
+    case 16:
+        /* $^P */
+        /* FALLTHROUGH */;
+    case 18:
+        /* $^R */
+        /* FALLTHROUGH */;
+    case 19:
+        /* $^S */
+        /* FALLTHROUGH */;
+    case 20:
+        /* $^T */
+        /* FALLTHROUGH */;
+    case 22:
+        /* $^V */
+        /* FALLTHROUGH */;
+    case 23:
+        /* $^W */
+        /* FALLTHROUGH */;
+    case 24:
+        /* $^X */
+        return name;
+    }
+    break;
+case 2:
+    switch ((U8)name[0]) {
+    case 14:
+        /* ${^NR} -> $.  (ALL) */
+        if (memEQs(name,*lenp,"\016R")) {
+            tmp_buf[0] = '.';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 18:
+        /* ${^RS} -> $/  (ALL) */
+        if (memEQs(name,*lenp,"\022S")) {
+            tmp_buf[0] = '/';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    }
+    break;
+case 3:
+    switch ((U8)name[0]) {
+    case 1:
+        /* ${^ARG} -> $_  (ALL) */
+        if (memEQs(name,*lenp,"\001RG")) {
+            tmp_buf[0] = '_';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 7:
+        /* ${^GID} -> $(  (ALL) */
+        if (memEQs(name,*lenp,"\007ID")) {
+            tmp_buf[0] = '(';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 15:
+        /* ${^OFS} -> $,  (ALL) */
+        if (memEQs(name,*lenp,"\017FS")) {
+            tmp_buf[0] = ',';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        /* ${^ORS} -> $\  (ALL) */
+        if (memEQs(name,*lenp,"\017RS")) {
+            tmp_buf[0] = '\\';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 16:
+        /* ${^PID} -> $$  (ALL) */
+        if (memEQs(name,*lenp,"\020ID")) {
+            tmp_buf[0] = '$';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 21:
+        /* ${^UID} -> $<  (ALL) */
+        if (memEQs(name,*lenp,"\025ID")) {
+            tmp_buf[0] = '<';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    }
+    break;
+case 4:
+    switch ((U8)name[0]) {
+    case 5:
+        /* ${^EGID} -> $)  (ALL) */
+        if (memEQs(name,*lenp,"\005GID")) {
+            tmp_buf[0] = ')';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        /* ${^EUID} -> $>  (ALL) */
+        if (memEQs(name,*lenp,"\005UID")) {
+            tmp_buf[0] = '>';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 15:
+        /* ${^OPEN} */
+        if (memEQs(name,*lenp,"\017PEN"))
+            return name;
+        break;
+    }
+    break;
+case 5:
+    switch ((U8)name[0]) {
+    case 5:
+        /* ${^ERRNO} -> $!  (ALL) */
+        if (memEQs(name,*lenp,"\005RRNO")) {
+            tmp_buf[0] = '!';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 13:
+        /* ${^MATCH} -> $&  (ALL) */
+        if (memEQs(name,*lenp,"\015ATCH")) {
+            tmp_buf[0] = '&';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 20:
+        /* ${^TAINT} */
+        if (memEQs(name,*lenp,"\024AINT"))
+            return name;
+        break;
+    }
+    break;
+case 6:
+    switch ((U8)name[0]) {
+    case 15:
+        /* ${^OSNAME} -> $^O  (ALL) */
+        if (memEQs(name,*lenp,"\017SNAME")) {
+            tmp_buf[0] = '\017';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 16:
+        /* ${^PERLDB} -> $^P  (ALL) */
+        if (memEQs(name,*lenp,"\020ERLDB")) {
+            tmp_buf[0] = '\020';
+            tmp_buf[1] = 0;
+            *lenp = 1;
+            return tmp_buf;
+        }
+        break;
+    case 19:
+        /* ${^SUBSEP} -> $;  (ALL) */
+        if (memEQs(name,*lenp,"\023UBSEP")) {
+            tmp_buf[0] = ';';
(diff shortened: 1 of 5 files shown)


demerphq:
One downside of this strategy is that is the worst from the point of view of making the "caret" namespace strict.

leonerd:
I put a bunch of comments on IRC but to copy here in brief: The principle seems fine but I'd love to see it implemented in terms of storing the name->var alias mapping in some sort of static data structure, rather than only becoming apparent by actually running the code.

E.g. if we started off with a data structure like:

```c
static const struct {
  const char *longname;
  char alias;
} english_varnames[] = {
  { "NR", '_' },
  { "GID", '(' },
  { "OFS", '\\' },
  { "PID", '$' },
  ...
};
```

then the entire body of this switch block becomes a query into that data structure. The structure is much easier to maintain in the future, more obvious to read. It contains the "real" names of the vars. Plus also in future it easily admits things like a strictness check to query if a mapping exists, or code that can enumerate all of the names from it. Perhaps this would eventually become the basis of the strictness checking overall.

(Details such as whether the data structure also stores the string lengths, or what order it's sorted into probably come out of how easily it can be queried most efficiently... but likely a good solution exists somehow).

I don't mind drastically if you want to make that change here or not. But if not I'll probably come along and fix it up in that direction afterwards anyway.. ;) I'm on a long crusade to try to assist static parsers of perl code as much as possible, and I think this kind of thing would be a great help towards that goal.

demerphq:
Well, just to summarize my response, IMO this comes down to the operations we want to do. If we want to be able to enumerate the set of known internal caret vars, then we can use an HV. But it means hashing a bunch of stuff at startup, and blah.  Personally i dont see why we would want to do that. We lazily create the variables we need, so the core never does anything like that. 

If we just need a fast way to check "is this an builtin caret var" then the code in this patch, or future versions of this PR, will do it efficiently.  The code to do this right needs to know all the legal options. I dont see the point in building a slow run time lookup with a custom data structure, when we can use code and make it fast. Especially if YAGNI says you dont need to enumerate the list in code. 

You can get a list of perlvars from pod/perlvar.pod, so we can always set up something to parse that for the list of known variables, I don't see why it needs to be baked into the core.

Anyway, i dont really care. But it sounds like you are trying to address a broader issue (having a manifest of all know internally used perl vars) with the English names patch. That *only* needs concern itself with *caret* vars, which is a subset of the whole.

Another issue to consider is that there are module load issues involved. For instance `$!` and `${^ERRNO}` load the Errno module. That means it has to be lazily populated fairly late. Similar story with some of the regex magic vars. I think there is a bigger can of worms involved in what you are talking about than you realize. I definitely think there is value in having a manifest of perl internal vars, but what we have on top of that should be addressed in a different PR.

For now, i am going to focus on the RFC, and i dont think we need an enumerable data structure to accomplish it.

demerphq:
So this works. Its not polished or wrapped up. Tests fail with it (lot so things use caret vars). It should only warn under warnings. It doesn't need me to finish it off. Needs a lot of test fixes, docs, and the like. But I have broken its back if someone else wants to pick it up. I probably wont look into this much for the next days.

haarg:
The way this is handling `${^MATCH}`, `${^PREMATCH}`, and `${^POSTMATCH}` is incomplete. Those variables already exist without this patch. The documentation states that since perl 5.20, the `/p` flag is ignored and `${^MATCH}` is equivalent to `$&`. This is incorrect. The `/p` flag is not ignored, and still controls the population of `${^MATCH}` etc. The simplest resolution to this is likely to make the behavior match the documentation, which is what this PR already implements. However in that case, the existing implementation of those vars should be removed, as well as changing `/p` to truly be ignored.

The documentation for the match variables and `/p` needs to be fixed no matter what this PR does, so that can probably be addressed separately.

This currently seems to be issuing two warnings for every use of a non-existent caret variable. It also doesn't warn for variables using an inappropriate sigil, like `@{^EXECUTABLE_NAME}`. I would think we would want to error, not warn.

There needs to be some way to disable or otherwise avoid the warnings/errors from a non-existent caret variable. If a new caret variable was added in the future, writing code to use it conditionally should be possible:
```perl
no strict "vars";
if ($] >= 5.040) {
    do_something(${^NEW_MAGIC_VAR});
}
```
This isn't possible with the current implementation, as you can't disable the warnings in any way. My initial thought is to have use of these variables an error, but tied to strict vars.

Leont:
> For now, i am going to focus on the RFC, and i dont think we need an enumerable data structure to accomplish it.

I agree.

demerphq:
On Sun, 11 Dec 2022, 03:58 Graham Knop, ***@***.***> wrote:

> The way this is handling ${^MATCH}, ${^PREMATCH}, and ${^POSTMATCH} is
> incomplete. Those variables already exist without this patch. The
> documentation states that since perl 5.20, the /p flag is ignored and
> ${^MATCH} is equivalent to $&. This is incorrect. The /p flag is not
> ignored, and still controls the population of ${^MATCH} etc. The simplest
> resolution to this is likely to make the behavior match the documentation,
> which is what this PR already implements. However in that case, the
> existing implementation of those vars should be removed, as well as
> changing /p to truly be ignored.
>
I'll take a look. I vaguely remember that in fact sawampersand is still an
issue. It may be we need an alternative solution for these vars.

The documentation for the match variables and /p needs to be fixed no
> matter what this PR does, so that can probably be addressed separately.
>
> This currently seems to be issuing two warnings for every use of a
> non-existent caret variable. It also doesn't warn for variables using an
> inappropriate sigil, like @{^EXECUTABLE_NAME}. I would think we would
> want to error, not warn.
>
I made it warn so it is reasonable to fix the tests that this new rule
breaks. What it does in the long run is TBD. Right now the proposal breaks
our test suite, so we need to address that first, and the implications of
that as well. I would really welcome someone else picking that up.

If we additionally want to check type requested in gv_fetch, and error when
its not special then we can probably do that, but I suspect creating a
globref would defeat it. Are we going to do that also with the non caret
vars like %ARG or @STDOUT? That feels like a slippery slope.

There needs to be some way to disable or otherwise avoid the
> warnings/errors from a non-existent caret variable. If a new caret variable
> was added in the future, writing code to use it conditionally should be
> possible:
>
> no strict "vars";if ($] >= 5.040) {
>     do_something(${^NEW_MAGIC_VAR});
> }
>
> This isn't possible with the current implementation, as you can't disable
> the warnings in any way.
>
Sig warn?

> My initial thought is to have use of these variables an error, but tied to
> strict vars.
>
We can check if strict is enabled I suspect.

Cheers
Yves

caret_vars.inc
```diff
@@ -0,0 +1,742 @@
+case 1:
+    switch ((U8)name[0]) {
+    case 1:
+        /* $^A */
+        /* FALLTHROUGH */;
+    case 3:
```
tonycoz, line 6: I assume (hope) this code is generated - where is the generator?

gv.c
```diff
@@ -2600,9 +2610,27 @@ Perl_gv_fetchpvn_flags(pTHX_ const char *nambeg, STRLEN full_len, I32 flags,
     if (!stash && !find_default_stash(&stash, name, len, is_utf8, add, sv_type)) {
         return NULL;
     }
+    char tmp[30];
+    char *use_name = (char*)name;
```
tonycoz, line 2614: use_name could be `const char *` if you change the name parameter and return type of S_is_internal_caret_name() to match, avoiding this cast.

gv.c
```diff
@@ -2600,9 +2610,27 @@ Perl_gv_fetchpvn_flags(pTHX_ const char *nambeg, STRLEN full_len, I32 flags,
...
+    STRLEN use_len = len;
+
+    if (((U8)name[0]) < 32) {
+        use_name = S_is_internal_caret_name(use_name, &use_len, tmp);
```
tonycoz, line 2618: Why doesn't S_is_internal_caret_name just return a pointer to a static string?

caret_vars.inc
demerphq: I can add it to the patch, but it will fail tests and etc. I just whipped this patch up to get this subject moving, it is not a high priority for me, I have some complex regex bugs on my TODO list. Are you just horrified at the generated code or is outright wrong?  Fwiw, the file is result of a two step process. First this:

```
git grep -P "[@%$]\{?\^." pod/perlvar.pod | perl -lne'/(?<!\S)[\@\%\$]\{?(\^.\w*)\}?/ and print "$1"' | sort | uniq > caret_var.txt
```

And then this:

```
# cat ~/scratch/gen.pl 
use strict;
use warnings;
use warnings FATAL => 'all';
open my $fh, "<", "lib/English.pm";
my %symbol;
while (<$fh>) {
    chomp;
    if (/^\s*\*(\w+)\s+=\s+\*(\^\S|\S)(?:\{(\w+)\})?\s+;/) {
        my ( $long, $short, $type ) = ( $1, $2, $3 );
        $type = "ALL" unless $type;
        my $fc = ord( substr( $long, 0, 1 ) ) - 64;
        $symbol{ length $long }{$fc}{$long} =
          { short => $short, type => $type };
    }
}
open $fh, "<", "caret_var.txt";
while (<$fh>) {
    chomp;
    next unless /^\^(\S+)/;
    my $long= $1;
    my $fc = ord( substr( $long, 0, 1 ) ) - 64;
    if ($fc < 1) {
        warn "ignoring: $long";
        next;
    }
    if ($symbol{length $long}{$fc}{$long}) {
        next;
    }
    $symbol{length $long}{$fc}{$long} = { short => $long, type => "STD" };
}

foreach my $length ( sort { $a <=> $b } keys %symbol ) {
    print "case $length:\n";
    my $nkeys = 0 + keys %{ $symbol{$length} };
    print "    switch ((U8)name[0]) {\n";
    my @fc= sort { $a <=> $b } keys %{ $symbol{$length} }; 
    foreach my $fc ( @fc ) {
        print "    case $fc:\n";
        foreach my $long ( sort keys %{ $symbol{$length}{$fc} } ) {
            my $short           = $symbol{$length}{$fc}{$long}{short};
            my $type            = $symbol{$length}{$fc}{$long}{type};

            my $longname_dquoted = $long;
            $longname_dquoted =~ s/^(.)/sprintf "\\%03o",ord($1)-64/e;
            my $shortname_dquoted = $short;
            my $shortname_squoted;
            my $shortlen;

            if ( length($shortname_dquoted)>1 and $shortname_dquoted =~ s/^\^// )
            {
                if (length($shortname_dquoted) == 1) {
                    $shortlen = 1;
                    $shortname_squoted = sprintf "'\\%03o'",ord($shortname_dquoted)-64;
                    $shortname_dquoted = sprintf qq("\\%03o"),ord($shortname_dquoted)-64;
                } else {
                    $shortlen = length($shortname_dquoted);
                    $shortname_dquoted = sprintf qq("\\%03o%s"),
                      ord($shortname_dquoted) - 64, length($shortname_dquoted)>1 ? substr($shortname_dquoted,1) : "";
                }
            }
            else {
                $shortlen = 1;
                $shortname_squoted = $shortname_dquoted;
                $shortname_squoted =~ s/([\\'])/\\$1/;
                $shortname_dquoted =~ s/([\\"])/\\$1/;
                $shortname_squoted = sprintf q('%s'), $shortname_squoted;
                $shortname_dquoted = sprintf q("%s"), $shortname_dquoted;
            }
            $longname_dquoted = qq("$longname_dquoted");

            my $ok = 1;    #$short ne "!" ? "1" : "0";
            0 and print <<~"EOF";
                if ($ok) {
                    /* \${^$long} -> \$$short  ($type) */
                    SV *gv_as_sv= (SV*)gv_fetchpvs($shortname_dquoted,GV_ADDMG|GV_ADDMULTI,SVt_PVGV);
                    if (hv_stores(PL_defstash,$longname_dquoted,gv_as_sv))
                        SvREFCNT_inc(gv_as_sv);
                }
            EOF
            if ($type ne "STD") {
                my $len = length $long;
                if ($shortname_squoted) {
                    print <<~"EOF";
                            /* \${^$long} -> \$$short  ($type) */
                            if (memEQs(name,*lenp,$longname_dquoted)) {
                                tmp_buf[0] = $shortname_squoted;
                                tmp_buf[1] = 0;
                                *lenp = 1;
                                return tmp_buf;
                            }
                    EOF
                } else {
                    print <<~"EOF";
                            /* \${^$long} -> \$$short  ($type) */
                            if (memEQs(name,*lenp,$longname_dquoted)) {
                                Copy($shortname_dquoted,tmp_buf,$shortlen,char);
                                *lenp = $shortlen;
                                tmp_buf[*lenp] = 0;
                                return tmp_buf;
                            }
                    EOF
                }
            } elsif ($length > 1)  {
                print <<~"EOF";
                        /* \${^$long} */
                        if (memEQs(name,*lenp,$longname_dquoted))
                            return name;
                EOF
            } elsif ($fc == $fc[-1]) {
                print <<~"EOF";
                        /* \$^$long */
                        return name;
                EOF
            } else {
                print <<~"EOF";
                        /* \$^$long */
                        /* FALLTHROUGH */;
                EOF
            }
        }
        print "        break;\n" if $length>1;
    }
    print "    }\n";
    print "    break;\n";
}
```

gv.c
demerphq: As in a global static var? Probably because I just simply didn't think of it. :-)  I suspect I was influenced by the fact that not all names get translated, and this "works". IOW: no good reason.

caret_vars.inc
tonycoz: I was thinking more of what happens when someone wants to add a new name?

For the generated code it's mostly "maybe I wouldn't do it quite that way" with the advantage of having seen this generated code.

demerphq: >I was thinking more of what happens when someone wants to add a new name?

Perhaps it was naive of me, but I figured adding one name to this type of code wouldn't be too difficult.  That was partly why i structured it like this. A bit of copy and paste and you are good.

> For the generated code it's mostly "maybe I wouldn't do it quite that way" with the advantage of having seen this generated code.

Reading it again with fresh eyes, I think I would agree with you and would certainly be open to suggestions. :-)

jkeenan:
> This implements the English name aliases for the magic punctuation variables listed in English.pm
> 
> It does not include the "strict" part of the RFC.
> 
> Currently no tests. Hoping @haarg or someone else might help contribute some.

This pull request is failing its porting tests in the "Sanity" run in our GitHub CI.  Could those be resolved so that discussion in this ticket can proceed?  Thanks.

PR #20604, opened 2022-12-11
Add function amagic_find(sv, method, flags)

ericherman:
Returns the CV pointer to the overloaded method, which will be needed by join to detect concat magic.

Co-authored-by: @book

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -3377,6 +3377,48 @@ Perl_try_amagic_un(pTHX_ int method, int flags) {
 }
 
 
+/*
+=for apidoc amagic_find
+
+Check C<sv> for the overloaded (active magic) operation C<method>, and
+return the C<CV *> or C<NULL>.
+
+C<method> is one of the values found in F<overload.h>.
+
+C<flags> are available for future use.
+
+=cut
+*/
+CV *
+Perl_amagic_find(pTHX_ SV *sv, int method, int flags)
+{
+    PERL_ARGS_ASSERT_AMAGIC_FIND;
+    PERL_UNUSED_VAR(flags);
+
+    assert(method >= 0 && method < NofAMmeth);
+
+    if (!SvAMAGIC(sv))
+        return NULL;
+
+    HV *stash = SvSTASH(SvRV(sv));
+    if (!Gv_AMG(stash))
+        return NULL;
+
+    MAGIC *mg = mg_find((const SV *)stash, PERL_MAGIC_overload_table);
+    if (!mg)
+        return NULL;
+
+    CV **cvp = NULL;
+    if (AMT_AMAGIC((AMT *)mg->mg_ptr))
+        cvp = ((AMT *)mg->mg_ptr)->table;
+    if (!cvp)
+        return NULL;
+
+    CV *cv = cvp[method];
+    return cv;
+}
+
+
 /* Implement tryAMAGICbin_MG macro.
    Do get magic, then see if the two stack args are overloaded and if so
    call it.
--- a/embed.fnc
+++ b/embed.fnc
@@ -635,6 +635,7 @@ XEop	|bool	|try_amagic_bin	|int method|int flags
 XEop	|bool	|try_amagic_un	|int method|int flags
 Apd	|SV*	|amagic_call	|NN SV* left|NN SV* right|int method|int dir
 Apd	|SV *	|amagic_deref_call|NN SV *ref|int method
+Apd	|CV *	|amagic_find	|NN SV *sv|int method|int flags
 p	|bool	|amagic_is_enabled|int method
 Apd	|int	|Gv_AMupdate	|NN HV* stash|bool destructing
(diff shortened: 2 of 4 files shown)


ericherman:
@demerphq : @book mentioned that this function would be useful to some of your work independent of #20503 . In this patch, the function is currently stand-alone and without tests; do you have thoughts about how it should be tested in isolation?

There are perhaps opportunities to reduce some duplication, as there are similar uses in Perl_amagic_call, however to make that sort of refactoring readable, I think we'd want to introduce a struct to hold the pointers to the mg, stash, cvp, etc. ... this patch does not attempt that.

gv.c
```diff
@@ -3376,6 +3376,46 @@ Perl_try_amagic_un(pTHX_ int method, int flags) {
...
+C<flags> are available for future use.
+
+=cut
+*/
+CV *
+Perl_amagic_find(pTHX_ SV *sv, int method, int flags)
+{
+    PERL_ARGS_ASSERT_AMAGIC_FIND;
+    assert(flags == 0);
+    (void)flags;
```
leonerd: Wants to be `PERL_UNUSED_VAR(flags);`

gv.c
```diff
@@ -3376,6 +3376,46 @@ Perl_try_amagic_un(pTHX_ int method, int flags) {
...
+    if (!mg)
+        return NULL;
+
+    CV **cvp = NULL;
+    if (AMT_AMAGIC((AMT *)mg->mg_ptr))
+        cvp = ((AMT *)mg->mg_ptr)->table;
+    if (!cvp)
+        return NULL;
+
+    CV *cv = cvp[method];
```
leonerd, line 3417: There's no bounds check here, but perhaps at the top of the function it might be a good idea to add

```c
assert(method >= 0 && method < max_amg_code);
```

leonerd: Actually on further inspection there's a better macro for that.

```c
assert(method >= 0 && method < NofAMmeth);
```

demerphq:
On Sun, 11 Dec 2022, 11:42 Eric Herman, ***@***.***> wrote:

> @demerphq <https://github.com/demerphq> : @book <https://github.com/book>
> mentioned that this function would be useful to some of your work
> independent of #20503 <https://github.com/Perl/perl5/pull/20503> . In
> this patch, the function is currently stand-alone and without tests; do you
> have thoughts about how it should be tested in isolation?
>
Basically it should agree with overload::Method. Usually you test this sort
of thing in ext/XSAPI-test (however it is spelled) by creating an XS
function to wrap the function you want to test, and then add a perl .t test
file to call it. For instance, if I have an object with certain overloads
the wrapper and o::Method should agree in a boolean context about whether
an object implements an overload method.

FWIW, The XSAPI-test xs file is huge, you can add new sections to create
new test modules and new tests fairly easily. Check it's changelog for
examples.

For instance, I'd expect a simple wrapper that takes an sv and an integer,
and returns a bool whether that sv is a reference with that integer
overload method enabled.

You then cross check that with overload::Method, you just need to account
for it using a string for the method, and the internals needs an integer
constant. I would just bake the constant into the test. Eg, array
drreferencing is '@{}' in overload::Method.

There are perhaps opportunities to reduce some duplication, as there are
> similar uses in Perl_amagic_call, however to make that sort of refactoring
> readable, I think we'd want to introduce a struct to hold the pointers to
> the mg, stash, cvp, etc.
>
I am not so much thinking about that level. I was thinking about maybe
rewriting overload.pm to use it.


and but this patch does not attempt that.
>

Sure, makes sense. Thanks for this, I'm looking forward to making use of it.

Yves

>

leonerd:
A thought on testing: Would it be possible to adjust the code in `amagic_call()` to use this function to find the magic first..? That way you wouldn't need to test this function specifically, as it would already be implicitly tested by anything that tests `amagic_call()`, for which there's already quite a lot of test code.

ericherman:
> A thought on testing: Would it be possible to adjust the code in `amagic_call()` to use this function to find the magic first..?

As I mentioned, there is an opportunity to do a bigger refactoring, however the callers stash a lot of intermediate pointers (mg, stash, cvp, etc), thus it is not a straight-forward refactoring.

We could change amagic_find to receive pointers-to-pointers as parameters, or introduce a parameter struct to hold the pointers, however it is not clear to me that either of these would be a great improvement.

I think it is possible, yes, however I do not have a good vision of what a true improvement of extracting the duplication of amagic_call would look like.

demerphq:
@ericherman is there any reason this is still marked draft? Id be fine with applying it as is.

leonerd:
> @ericherman is there any reason this is still marked draft? Id be fine with applying it as is.

+1. It looks fine to me.

demerphq:
Once its merged we can add tests after. Probably faster. :-)

ericherman:
Only that we didn't have tests in place yet. However, I'm happy to see it tested indirectly, as will happen with RFC0013 and your work @demerphq

PR #20605, opened 2022-12-11
Add `forbid_outofblock_ops()` to op.c; detect forbidden flow at compiletime of `defer` or `finally`

leonerd:
Add `forbid_outofblock_ops()` to op.c; detect forbidden flow at compiletime of `defer` or `finally`

Adds a new function to statically detect forbidden control flow out of a block.

Many kinds of forbidden control flow out of a `defer` or `finally` block can be detected statically with this function by analysing the optree, rather than leaving it until runtime when the attempt is actually made.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -66,6 +66,18 @@ C<INCDIR> methods will no longer trigger an exception, and instead will
 be treated the same as unblessed coderefs are, and executed as though
 they were an C<INC> hook.
 
+=head2 Forbidden control flow out of C<defer> or C<finally> now detected at compile-time
+
+It is forbidden to attempt to leave a C<defer> or C<finally> block by means
+of control flow such as C<return> or C<goto>. Previous versions of perl could
+only detect this when actually attempted at runtime.
+
+This version of perl adds compile-time detection for many cases that can be
+statically determined. This may mean that code which compiled successfully on
+a previous version of perl is now reported as a compile-time error with this
+one. This only happens in cases where it would have been an error to actually
+execute the code anyway; the error simply happens at an earlier time.
+
 =head1 Security
 
 XXX Any security-related notices go here.  In particular, any security
--- a/embed.fnc
+++ b/embed.fnc
@@ -1113,6 +1113,7 @@ Cp	|I32	|foldEQ_utf8_flags |NN const char *s1|NULLOK char **pe1|UV l1 \
 				|bool u1|NN const char *s2|NULLOK char **pe2 \
 				|UV l2|bool u2|U32 flags
 Cip	|I32	|foldEQ_latin1	|NN const char* a|NN const char* b|I32 len
+Apdx	|void	|forbid_outofblock_ops|NN OP *o|NN const char *blockname
 #if defined(PERL_IN_DOIO_C)
 SR	|bool	|ingroup	|Gid_t testgid|bool effective
 #endif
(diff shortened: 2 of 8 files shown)


richardleach:
Sounds cool. Since the API function is intended to be generally useful, please could you put the function name in the commit message?

leonerd:
Commit message and PR description rewritten.

leonerd:
(I had to push an empty commit to force CI to retest stuff; for some reason it didn't rerun the tests and I don't know why. I'll remove it before merge)

rwp0 (approved):
I don't claim to understand the implementation but the description (a switch from runtime to static analysis) is an improvement to the feature.

wolfsage:
Nice. Is it worth mentioning in the docs that "You should still have checks at run time" for cases like executing a subroutine or evaling a string that tries to goto out? That is, this static function alone isn't enough, it's just useful for finding compile time errors

leonerd:
> Is it worth mentioning in the docs that "You should still have checks at run time" for cases like executing a subroutine or evaling a string that tries to goto out?

Ahyes, definitely. I'll add some words to that effect.

wolfsage (approved):

pod/perldelta.pod
```diff
@@ -66,6 +66,17 @@ C<INCDIR> methods will no longer trigger an exception, and instead will
...
 
+=head2 Forbidden control flow out of C<defer> or C<finally> now detected at compile-time
+
+It is forbidden to attempt to leave a C<defer> or C<finally> block by means
+of control flow such as C<return> or C<goto>. Previous versions of perl could
+only detect this when actually attempted at runtime.
+
+This version of perl adds compile-time detection for many cases that can be
+statically determined. This may mean that code which previously compiled
+successfully is now reported as an error, but only in cases where it would in
```
rwp0: > but only in cases where it would in any case have been an error to actually execute the code anyway.

I re-read this sentence many times to understand it.
Perhaps, we can supply an example or leave just the main point?

leonerd: I'll see if I can reword it somehow. Maybe:

"This version of perl adds compile-time detection for many cases that can be statically determined. This may mean that code which compiled successfully on a previous version of perl is now reported as a compile-time error with this one. This only happens in cases where it would in any case have been an error to actually execute the code anyway."

rwp0: > in cases where it would in any case have been an error to actually execute the code anyway.

The **clause** still remains unchanged :-) 

The new compile-time out-of-block error only happens where it was already an (runtime) error?

What's the meaning of this, why do we need to mention that, or maybe it's just me not getting the point yet (sorry, then) :-)

rwp0: I think we can say something like: out-of-block was already an error in previous Perl versions. But the error is now determined at compile-time rather than runtime as previously hence improving the speed of execution or error determination (if I got that right).

leonerd: It's no faster to execute as we still have to detect those conditions anyway, because we can't *always* detect it statically. It just means that some kinds of programs that were always going to be an error are determined to be so at compiletime, rather than waiting until runtime to find out.

leonerd: I've rewritten the words in perldelta now.

The user-facing reason is "because compiletime is better than runtime".

Additionally I wanted to use this new function in the feature-class branch, for protecting the expressions in `field $var = EXPR` so I needed the code around anyway. Since it's here (because I copied it from `Syntax::Keyword::Defer`) we might as well use it for `defer` and `finally` blocks too.

Issue #20601, opened 2022-12-10
Build warning "warning: pointer ‘where’ may be used after ‘realloc’" in util.c


leonerd:
Recent builds are provoking a gcc warning:

```
In file included from util.c:26:
util.c: In function ‘Perl_safesysrealloc’:
util.c:329:17: warning: pointer ‘where’ may be used after ‘realloc’ [-Wuse-after-free]
  329 |         DEBUG_m(PerlIO_printf(Perl_debug_log, "0x%" UVxf ": (%05ld) rfree\n",PTR2UV(where),(long)PL_an++));
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
perl.h:4899:37: note: in definition of macro ‘DEBUG_m’
 4899 |                                     a;                                  \
      |                                     ^
util.c:290:15: note: call to ‘realloc’ here
  290 |         ptr = (Malloc_t)PerlMem_realloc(where,size);
```

Having looked into the code a bit I don't think this is really actually a problem. We're using a pointer after realloc, but only to convert it into a numerical representation to print it in a debugging message. This should be fine.

To fix it: Either we could convert the number into a UV before realloc, or we could somehow quiet the warning on this one line - though I don't know how to do that.

jkeenan:
On FreeBSD-12, building threaded, debugging builds with `gcc12` -- but **not** `gcc10` -- I am getting the build-time warning you report plus one other one.
```
$ report-build-warnings 84fa78b483.freebsd.threaded.gcc12.debugging.maketp.output.txt.gz 
 84fa78b483.freebsd.threaded.gcc12.debugging.maketp.output.txt.gz

  Warray-bounds                              1
  Wuse-after-free                            1

$ parse-build-warnings 84fa78b483.freebsd.threaded.gcc12.debugging.maketp.output.txt.gz
 84fa78b483.freebsd.threaded.gcc12.debugging.maketp.output.txt.gz

[
  {
    char   => 17,
    group  => "Wuse-after-free",
    line   => 329,
    source => "util.c",
    text   => "pointer 'where' may be used after 'realloc'",
  },
  {
    char   => 26,
    group  => "Warray-bounds",
    line   => 442,
    source => "hv.h",
    text   => "array subscript [0, 9223372036854775807] is outside array bounds of 'char[0]'",
[...]

PR #20607, opened 2022-12-14
Avoid a use-after-free warning

leonerd:
... by converting a pointer to a UV for debug print purposes before we call realloc() on it


jkeenan:
Per [comment by Tony Cook in GH 20601](https://github.com/Perl/perl5/issues/20601#issuecomment-1352390170), this p.r. is not accomplishing all that is intended to do, but that may be due to a bug in gcc (which @tonycoz) has reported.

demerphq:
This patch seems to be stalled on a gcc bug that folks are arguing and bickering about. I am going to merge it as a "better than nothing" and we can address it again in the future.

PR #20612, opened 2022-12-15
Define some more PL_infix_plugin precedence levels

leonerd:
Assignment operators (`==`) were missing, as were both the logical and the low-precedence shortcutting OR and AND operators (`&&`, `||`, `and`, `or`)

Also renumbered them around somewhat to even out the spacing. This is fine during a development cycle.


perl.h
```diff
@@ -6003,14 +6003,23 @@ typedef void (*XSINIT_t) (pTHX);
...
 
-/* TODO: find somewhere to store this */
 enum Perl_custom_infix_precedence {
-    INFIX_PREC_LOW  =  10, /* non-associative */
-    INFIX_PREC_REL  =  30, /* non-associative, just below `==` */
-    INFIX_PREC_ADD  =  50, /* left-associative, same precedence as `+` */
-    INFIX_PREC_MUL  =  70, /* left-associative, same precedence as `*` */
-    INFIX_PREC_POW  =  90, /* right-associative, same precedence as `**` */
-    INFIX_PREC_HIGH = 110, /* non-associative */
+    /* These numbers are spaced out to give room to insert new values as
```
wolfsage, line 6007: If the values here don't have much or any negative impact on memory or speed of the compiled program it might be nice to space them out even more. That way, maybe the actual values themselves would never need to be changed because there'd be enough room inbetween for plugins to do things?

perly.y
```diff
@@ -987,8 +984,12 @@ sigsubbody:	remember optsubsignature PERLY_BRACE_OPEN stmtseq PERLY_BRACE_CLOSE
 /* Ordinary expressions; logical combinations */
 expr	:	expr[lhs] ANDOP expr[rhs]
 			{ $$ = newLOGOP(OP_AND, 0, $lhs, $rhs); }
+	|	expr[lhs] PLUGLOWANDOP[op] expr[rhs]
+			{ $$ = build_infix_plugin($lhs, $rhs, $op); }
 	|	expr[lhs] OROP[operator] expr[rhs]
 			{ $$ = newLOGOP($operator, 0, $lhs, $rhs); }
+	|	expr[lhs] PLUGLOWOROP[op] expr[rhs]
```
wolfsage: What is `OROP[operator]` doing above here that you aren't and does it matter?

perly.y
```diff
@@ -1154,8 +1157,12 @@ termbinop:	term[lhs] PLUGHIGHOP[op] term[rhs]
 			{ $$ = newRANGE($DOTDOT, scalar($lhs), scalar($rhs)); }
 	|	term[lhs] ANDAND term[rhs]                       /* $x && $y */
 			{ $$ = newLOGOP(OP_AND, 0, $lhs, $rhs); }
+	|	term[lhs] PLUGLOGICALANDOP[op] term[rhs]
+			{ $$ = build_infix_plugin($lhs, $rhs, $op); }
 	|	term[lhs] OROR term[rhs]                         /* $x || $y */
 			{ $$ = newLOGOP(OP_OR, 0, $lhs, $rhs); }
+	|	term[lhs] PLUGLOGICALOROP[op] term[rhs]
```
wolfsage: Does this handle DORDOR too?

perly.y
leonerd: The newly-added rules are all the ones consuming a `PLUG*` token, which means it's a plugin infix operator handled by `PL_infix_plugin`. The code blocks all have to run `build_infix_plugin()` to invoke the code properly. As compared the non-`PLUG*` ones which just build internal native optrees directly by calling e.g. `newLOGOP` in this case.

perly.y
leonerd: These are the custom infix plugin ops. There isn't a `PLUGDOR...` because it would just be at the same precedence level as regular OR ops anyway.

perly.y
leonerd: Plus: Yes it's annoying that this is basically the same exact rule copied nine times with nine different token types. I did try squashing them all together with a single rule inbetween, but that causes a huge amount of grammar conflict, because that loses operator precedence information. :(

perl.h
leonerd: Yeah there's plenty of space, and they only have to fit in an `enum`, which is probably `int`-sized. I'll space them out more.

leonerd: Done. Though I've kept them within the range of a U8, in case in future we want to split the field up and add some flags in the remaining bits.

leonerd:
I decided I didn't like the names, so I've done a bunch of mass renaming so now things look nicer. It makes the diffs look worse, but that's only a temporary problem ;)

All the plugin operators now use token names `PLUGIN_*_OP` rather than `PLUG*OP` so it's easier to read.

wolfsage (approved):

Issue #20565, opened 2022-11-30
Perl_sync_locale stopped working somewhere after 5.37


dk:
I'm having a problem in my XS code that sync_locale() stopped working in 5.37.5 -
the issue is similar to what was already once talked about back in 2019 ( https://rt.perl.org/Ticket/Display.html?id=133945 )

Please find attached a proper test case, but shortly,  it happens under LC_NUMERIC=de_DE.UTF-8 when C code
```
    sync_locale();
    SvNV(newSVpv("4.8888", 0))
```
returns 4.0 .

Important: In order to reproduce the bug your system needs to have the de_DE.UTF-8 locale installed.

dk:
[sync_locale_bug.zip](https://github.com/Perl/perl5/files/10127115/sync_locale_bug.zip)

unpack, then perl Makefile.PL, make, ./test

jkeenan:
> I'm having a problem in my XS code that sync_locale() stopped working in 5.37.5 

Can you provide the configuration arguments you used to build the `perl` executable at which you observe this problem?  Perhaps `./perl -Ilib -V:config_args`

You say this stopped working in 5.37.5.  Does that mean that it was working in 5.37.4?

jkeenan:
I am having trouble reproducing your failure.
```
$ uname -mrs
FreeBSD 12.3-RELEASE-p1 amd64

$ locale -a | grep -E 'de_DE\.UTF-8'
de_DE.UTF-8

$ export THISDIR="$TESTINGDIR/v5.37.6"
$ export THISPERL="$THISDIR/bin/perl -I$THISDIR/lib"
$ $THISPERL -v | head -2 | tail -1
This is perl 5, version 37, subversion 6 (v5.37.6) built for amd64-freebsd-thread-multi
$ $THISPERL Makefile.PL && make
v5.37.6
Generating a Unix-style Makefile
Writing Makefile for T
Writing MYMETA.yml and MYMETA.json
Skip blib/lib/T.pm (unchanged)
Running Mkbootstrap for T ()
chmod 644 "T.bs"
"/usr/home/jkeenan/testing/v5.37.6/bin/perl" -MExtUtils::Command::MM -e 'cp_nonempty' -- T.bs blib/arch/auto/T/T.bs 644
[...]

PR #20618, opened 2022-12-17
Fix broken API: sync_locale()

khwilliamson:
This series of commits fixes GH https://github.com/Perl/perl5/issues/20565.

Lack of tests allowed sync_locale() to get broken until CPAN testing
showed it so.

Basically, I blew it in https://github.com/Perl/perl5/commit/9f5a615be674d7663d3b4719849baa1ba3027f5b.  
Most egregiously, I forgot to turn back on when a sync_locale() is executed,
the toggling for locales whose radix character isn't a dot.  And this
needs a way to tell the other code that it needs to recompute things at
this time, since our records don't reflect what happened before the
sync.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -3388,13 +3388,13 @@ ST	|const char *|save_to_buffer|NULLOK const char * string	\
 				    |NULLOK Size_t *buf_size
 ST	|unsigned int|get_category_index|const int category|NULLOK const char * locale
 #    ifdef USE_LOCALE_CTYPE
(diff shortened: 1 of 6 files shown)


jkeenan:
This pull request is doing fine on *nix (http://perl.develop-help.com/?b=smoke-me%2Fkhw-dimitry) but failing on 3 CI tests on Windows.

khwilliamson:
Actually, this PR is OK on Windows.  What was failing was #20619 (now fixed)

PR #20623, opened 2022-12-18
Speed up test suite execution on a many core system

khwilliamson:
This series of commits does better parallelization of the test suite, especially when PERL_TEST_HARNESS_ASAP is set.  That leads to an improvement on my box of 30-40% in elapsed time for executing the suite

### DIFF
--- a/t/harness
+++ b/t/harness
@@ -194,9 +194,10 @@ if (@ARGV) {
                     \z !x)
             {
                 my $path = $1;
+                my $name = $2;
 
                 $all_dirs{$path} = 1;
-                $serials{$path} = 1 if $2 =~ / \A 0 /x;
+                $serials{$path} = 1 if $name =~ / \A 0 /x;
             }
         }
 
--- a/t/harness
+++ b/t/harness
@@ -8,6 +8,9 @@ BEGIN {
     @INC = '../lib';              # pick up only this build's lib
 }
 
+my %must_be_executed_serially = map { $_ => 1 }
+                                    qw(../cpan/IO-Zlib/t ../ext/File-Find/t);
+
 my $torture; # torture testing?
 
 use TAP::Harness 3.13;
@@ -200,25 +203,21 @@ if (@ARGV) {
                 my $name = $2;
 
                 $all_dirs{$path} = 1;
-                $serials{$path} = 1 if $name =~ / \A 0 /x;
                 $map_file_to_dir{$file} = $path;
+
+                if ($name =~ / \A 0 /x) {
+                    $serials{$path} = 1;
+                }
+                elsif (defined $must_be_executed_serially{$path}) {
+                    $serials{$path} = 1;
+                    delete $must_be_executed_serially{$path};
+                }
             }
         }
 
         # We assume that the reason a test file's name begins with a 0 is to
         # order its execution among the tests in its directory.  Hence, a
         # directory containing such files should be tested in serial order.
-        #
-        # Add exceptions to the above rule
-        for (qw(../cpan/IO-Zlib/t ../ext/File-Find/t)) {
-            $serials{$_} = 1;
-        }
-
-        my @nonexistent_serials = grep { not exists $all_dirs{$_} } keys %serials;
-        if (@nonexistent_serials) {
-            die "These directories to be run serially don't exist."
-              . "  Check your spelling:\n" . join "\n", @nonexistent_serials;
-        }
 
         for my $file (@last) {
             my $file_dir = $map_file_to_dir{$file};
@@ -249,6 +248,12 @@ if (@ARGV) {
                                                                      keys %dir
                             ]
                    };
+
+        if (%must_be_executed_serially) {
+            die "These directories to be run serially don't exist.  Check your"
+            . " spelling:\n"
+            . join "\n", sort keys %must_be_executed_serially;
+        }
 
         $rules = { seq => \@seq };
     }
--- a/t/harness
+++ b/t/harness
@@ -178,6 +178,7 @@ if (@ARGV) {
         my %dir;
         my %serials;
         my %all_dirs;
+        my %map_file_to_dir;
 
         # Preprocess the list of tests
         for my $file (@last) {
@@ -199,6 +200,7 @@ if (@ARGV) {
 
                 $all_dirs{$path} = 1;
                 $serials{$path} = 1 if $name =~ / \A 0 /x;
+                $map_file_to_dir{$file} = $path;
             }
         }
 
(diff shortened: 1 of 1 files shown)


khwilliamson:
I am sometimes getting "You already have a parser for (../cpan/Pod-Simple/t/00about.t). Perhaps you have run the same test twice. at ../lib/TAP/Harness.pm line 597".  That directory is only mentioned once in the rules, and that file only once in the tests array

t/harness
```diff
@@ -90,6 +72,150 @@ if ($ENV{HARNESS_OPTIONS}) {
...
+                $serials{$path} = 1;
+            }
+            elsif (defined $must_be_executed_serially{$path}) {
+                $serials{$path} = 1;
+                delete $must_be_executed_serially{$path};
+            }
+        }
+    }
+
+    # We assume that the reason a test file's name begins with a 0 is to
```
demerphq: shouldnt this be "begins with a digit"? I think this heuristic is too picky or the comment is wrong :-)

t/harness
```diff
@@ -90,6 +72,150 @@ if ($ENV{HARNESS_OPTIONS}) {
...
+    # order its execution among the tests in its directory.  Hence, a
+    # directory containing such files should be tested in serial order.
+    #
+    # Add exceptions to the above rule
+    for (qw(../cpan/IO-Zlib/t ../ext/File-Find/t)) {
+        $serials{$_} = 1;
+    }
+
+    my @nonexistent_serials = grep { not exists $all_dirs{$_} } keys %serials;
+    if (0 && @nonexistent_serials) {
```
demerphq: 0 && ?

t/harness
```diff
@@ -292,71 +335,71 @@ my $h = TAP::Harness->new({
     verbosity   => $Verbose,
     timer       => $ENV{HARNESS_TIMER},
     exec        => sub {
-	my ($harness, $test) = @_;
```
demerphq, line 295: isnt our policy to replace tabs with spaces?

demerphq (commented):
Mostly looks good, a few minor nits, but I am not approving it yet as you have said it has issues.

demerphq:
> I am sometimes getting "You already have a parser..."

Yeah, it is running tests more than once, apparently when they begin with a 0.
Try:

```
TEST_JOBS=16 make test_harness -j16 TEST_ARGS='-re cpan/Archive-Tar'
```
which gives this:

```
....
TESTFILE=harness  ./runtests choose
../cpan/Archive-Tar/t/01_use.t .............. ok   
../cpan/Archive-Tar/t/02_methods.t .......... ok     
../cpan/Archive-Tar/t/03_file.t ............. ok     
../cpan/Archive-Tar/t/04_resolved_issues.t .. ok    
../cpan/Archive-Tar/t/05_iter.t ............. ok    
../cpan/Archive-Tar/t/06_error.t ............ ok    
../cpan/Archive-Tar/t/08_ptargrep.t ......... ok   
../cpan/Archive-Tar/t/09_roundtrip.t ........ ok   
../cpan/Archive-Tar/t/90_symlink.t .......... skipped: Skipping tests on this platform
../cpan/Archive-Tar/t/99_pod.t .............. skipped: Test::Pod v0.95 required for testing POD
../cpan/Archive-Tar/t/01_use.t .............. ok   
All tests successful.
Files=10, Tests=1240,  3 wallclock secs ( 0.35 usr  0.02 sys +  2.78 cusr  0.16 csys =  3.31 CPU)
Result: PASS
You already have a parser for (../cpan/Archive-Tar/t/01_use.t). Perhaps you have run the same test twice. at ../lib/TAP/Harness.pm line 597.
make: *** [makefile:865: test_harness] Error 25
```

also

```
TEST_JOBS=16 make test_harness -j16 TEST_ARGS='-re cpan/IO-Socket-IP'
```

shows the same thing. It is somewhat random which one is triggered first what I can tell. For instance I did 

```
TEST_JOBS=16 make test_harness -j16 TEST_ARGS='-re cpan'
```

and I found the IO-Socket-IP case. Something wrong with the scheduler logic I guess. Ill try to find the bug, just posting this if you are still awake.

Dumping `@tests` at the end gives this:

```
../cpan/Archive-Tar/t/01_use.t
../cpan/Archive-Tar/t/02_methods.t
../cpan/Archive-Tar/t/03_file.t
../cpan/Archive-Tar/t/04_resolved_issues.t
../cpan/Archive-Tar/t/05_iter.t
../cpan/Archive-Tar/t/06_error.t
../cpan/Archive-Tar/t/08_ptargrep.t
../cpan/Archive-Tar/t/09_roundtrip.t
../cpan/Archive-Tar/t/90_symlink.t
../cpan/Archive-Tar/t/99_pod.t
../cpan/Archive-Tar/t/01_use.t
../cpan/Archive-Tar/t/02_methods.t
../cpan/Archive-Tar/t/03_file.t
../cpan/Archive-Tar/t/04_resolved_issues.t
../cpan/Archive-Tar/t/05_iter.t
../cpan/Archive-Tar/t/06_error.t
../cpan/Archive-Tar/t/08_ptargrep.t
../cpan/Archive-Tar/t/09_roundtrip.t
../cpan/Archive-Tar/t/90_symlink.t
../cpan/Archive-Tar/t/99_pod.t
Died at harness line 405.
make: *** [makefile:865: test_harness] Error 255
```

its scheduling the module twice.

t/harness
```diff
@@ -90,6 +72,150 @@ if ($ENV{HARNESS_OPTIONS}) {
...
+            my $name = $2;
+
+            $all_dirs{$path} = 1;
+            $map_file_to_dir{$file} = $path;
+
+            # We assume that the reason a test file's name begins with a 0
+            # is to order its execution among the tests in its directory.
+            # Hence, a directory containing such files should be tested in
+            # serial order, with some exceptions hard-coded in.
+            if ($name =~ / \A 0 /x) {
```
demerphq, line 153: I think this should be `\d`

t/harness
```diff
@@ -114,133 +240,50 @@ if (@ARGV) {
...
-            die "These directories to be run serially don't exist."
-              . "  Check your spelling:\n" . join "\n", @nonexistent_serials;
+        # This is a bit of a game, because we only want to sort these tests in
+        # speed order. base/*.t wants to run first, and ext,lib etc last and in
+        # MANIFEST order
+        push @last,
+            _tests_from_manifest($Config{extensions}, $Config{known_extensions});
+        push @tests, @last;
+
+        push @tests, @last;
```
demerphq, line 252: This line is duplicated, and is the cause of the test fails.

demerphq (changes requested):
With the changes I mentioned all the tests pass for me locally.

demerphq:
With this diff the harness errors go away:
```
$ git diff
diff --git a/t/harness b/t/harness
index 9da972a20d..fef1c83a70 100644
--- a/t/harness
+++ b/t/harness
@@ -132,7 +132,7 @@ sub _compute_tests_and_ordering($) {
         };
 
         # Keep a list of the distinct directory names, and another list of
-        # those which contain a file whose name begins with a 0
+        # those which contain a file whose name begins with a digit
         if ($file =~ m! \A ( (?: \.\. / )?
                                 .*?
                             )             # $1 is the directory path name
@@ -146,11 +146,11 @@ sub _compute_tests_and_ordering($) {
             $all_dirs{$path} = 1;
             $map_file_to_dir{$file} = $path;
 
-            # We assume that the reason a test file's name begins with a 0
+            # We assume that the reason a test file's name begins with a digit
             # is to order its execution among the tests in its directory.
             # Hence, a directory containing such files should be tested in
             # serial order, with some exceptions hard-coded in.
-            if ($name =~ / \A 0 /x) {
+            if ($name =~ / \A \d /x) {
                 $serials{$path} = 1;
             }
             elsif (defined $must_be_executed_serially{$path}) {
@@ -274,7 +274,6 @@ if (@ARGV) {
             _tests_from_manifest($Config{extensions}, $Config{known_extensions});
         push @tests, @last;
 
-        push @tests, @last;
         push @seq, _compute_tests_and_ordering(\@last)->@*;
 
         if (%must_be_executed_serially) {
```

demerphq:
BTW, i love the idea behind this. It would be nice if it was easy to make the scheduler to have a callback, and then have half the workers read from one end of the queue and the other half from the front. (Eg, half do the fastest first, half do  the slowest first, and they meet in the middle).  I notice with your patch there is a long pause after startup, which would go way if some of the fast tests were being run early. Alternating fast and slow in the rules doesn't really work, the workers need to know the queue is ordered and which end to pull their job from.

t/harness
khwilliamson: a rebasing error

t/harness
khwilliamson: It has long been 0 without problems.  There are a bunch of test files whose names begin with their ticket number, and have nothing to do with ordering.

t/harness
khwilliamson: rebase error

t/harness
khwilliamson: rebase error

t/harness
khwilliamson: The paradigm is that if a directory has a test file whose name begins with a 0, to execute in numeric order all the test files in that directory whose names begin with a digit.  But at least one file has to have a leading 0.  This PR doesn't change the way this has worked for quite a while now.

t/harness
demerphq: Well, it feels wrong to me. We should just sort the files with leading digits in them into order, and then not care. Anyway, we can take care of that in another PR, i thought you added that logic, but on reflection you didnt.

t/harness
demerphq: Since this PR doesn't change it, then  changing it to be less fragile can wait for another PR.

demerphq (approved):

khwilliamson:
The reason for the pause is not because there is any idle CPU time, but because the first tests started take a long time to finish.   It would be a net loss to intermix the other end of the queue.

demerphq:
>  It would be a net loss to intermix the other end of the queue.

Maybe. Id like to see anyway.

FWIW, I have been working to a) fix all the parallelism blockers in our tests (basically done as far as I can tell) and breaking up large test files like 099_binary. (Done but not pushed upstream).

See https://github.com/Perl/perl5/pull/20643

PR #20624, opened 2022-12-18
loc_tools.pl: Always do normalized locale name check

khwilliamson:
Because of variances in capitilization and use of dashes, we should only compare locale names that have been normalized to a common syntax.  This was the remaining outlier, and caused issues on some platforms.

### DIFF
--- a/t/loc_tools.pl
+++ b/t/loc_tools.pl
@@ -147,7 +147,8 @@ sub _trylocale ($$$$) { # For use only by other functions in this file!
     my $list = shift;
     my $allow_incompatible = shift;
 
-    return if ! $locale || grep { $locale eq $_ } @$list;
+    my $normalized_locale = lc ($locale =~ s/\W//gr);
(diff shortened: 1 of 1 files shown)


jkeenan (approved):
1. Spelling error in commit message: `s/capitil/capital/`.
2. The statement starting with `return if`:  I had to stop and think, *"Which has higher precedence, the `!` or the `||`?"* Could you clarify with some parentheses?

demerphq:
>Could you clarify with some parentheses?

Normally i am sympathetic to that kind of request, but honestly with "!" I dont see the point, it has the highest precedence of all boolean operators, and parenthesizing everything inverted with it would be extremely tedius. I might have written "or" instead of "||" to emphasize the precedence difference, but i might not have either. but I likely wouldn't normally parenthesize a simple !x expression, it just adds line noise.

demerphq:
@jkeenan do you still object to this patch as is?

jkeenan:
> @jkeenan do you still object to this patch as is?

No, but the spelling error in the commit message still needs correction.

khwilliamson:
typo in commit message now corrected

PR #20626, opened 2022-12-18
More graceful and helpful handling of overloaded objects in @INC.

demerphq:
This PR adds logic to detect if objects in @INC have a string overload if they are not blessed CODE refs, and do not support a valid INC hook method (INC or INCDIR). If they do have an overload then we call it to stringify the hook. Otherwise we throw an exception. It does not make sense to actually look on disk for ARRAY(0x....) and in theory it could be a security issue if we do. (The practical likelyhood of this being a real issue is very low, but still.)

In older versions of perl we would throw an exception about not finding the INC method. This is more informative.

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -1183,6 +1183,14 @@ to load "%s"?)
 could not be found in UNIVERSAL.  This often means that a method
 requires a package that has not been loaded.
 
+=item Can't locate object method "INC", nor "INCDIR" nor string overload via package %s %s in @INC
+
+(F) You pushed an object, either directly or via an array reference hook,
+into C<@INC>, but the object doesn't support any known hook methods, nor
+a string overload and is also not a blessed CODE reference. In short the
+C<require> function does not know what to do with the object.
+See also L<perlfunc/require>.
+
 =item Can't locate package %s for @%s::ISA
 
 (W syntax) The @ISA array contained the name of another package that
@@ -4460,14 +4468,6 @@ report.  This limit is typically 2GB.
 and you mentioned a variable that starts with 0 that has more than one
 digit. You probably want to remove the leading 0, or if the intent was
 to express a variable name in octal you should convert to decimal.
-
-=item Object with arguments in @INC does not support a hook method
-
-(F) You pushed an array reference hook into C<@INC> which has an object
-as the first argument, but the object doesn't support any known hooks.
-Since you used the array form of creating a hook, you should have supplied
-an object that supports either the C<INC> or C<INCDIR> methods. You
-could also use a coderef instead of an object.
 
 =item Octal number > 037777777777 non-portable
 
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4316,7 +4316,12 @@ S_require_file(pTHX_ SV *sv)
                         && !SvOBJECT(SvRV(loader)))
                     {
                         loader = *av_fetch(MUTABLE_AV(SvRV(loader)), 0, TRUE);
-                        SvGETMAGIC(loader);
+                        if (SvGMAGICAL(loader)) {
+                            SvGETMAGIC(loader);
+                            SV *l = sv_newmortal();
+                            sv_setsv_nomg(l, loader);
+                            loader = l;
+                        }
                     }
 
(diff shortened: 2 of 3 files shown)


demerphq:
@book and @ericherman - this uses amagic_find().

haarg:
This doesn't seem to handle fallback or nomethod.

Granted, falling back to a numeric overload for an inc hook would never be an actual useful behavior. But the inconsistency doesn't seem like a good idea.

demerphq:
edited: removed crotchety pre-caffeine response to @haarg. he is right of course.

demerphq:
Regarding fallback and overload. Arguably that logic should be in amagic_find(). amagic_call() implements the equivalencies and IMO so should amgic_find(). 

I have to prepare for moving in the next weeks, I likely will not have a lot of time for perl.

demerphq:
@haarg sorry for being crufty this morning. See the new patchset, i made amagic_find() DTRT.

book:
Instead of having to keep `amagic_find` and `amagic_call` in sync, would it make sense for `amagic_call` to call `amagic_find`?

demerphq:
> Instead of having to keep amagic_find and amagic_call in sync, would it make sense for amagic_call to call amagic_find?

I think that is a matter of taste, but I think a close read of amagic_call() suggests a relatively firm *no*. amagic_call() deals with a lot more than a single SV. amagic_find() deals with one SV. So amagic_find necessarily is never going to be replace the internals of amagic_call(), at best you might get to a point where amagic_call makes two calls to amagic_find().

Consider if you have a binary operation. The left side could be overloaded, or the right side could be overloaded.  amagic_call() knows about this and knows how to deal with it efficiently. Unrolling that logic would be time-consuming and painful, and IMO would result in much less efficient logic. overloads are already slow, we dont want to make them slower. There is also the fact that this is Ilya code. His style is *extremely* dense, I have had to refactor his code before and it is extremely troublesome to do so. I have never seen code with such a high "density" of logic to characters typed. Even just ripping out the logic for this was a bit trouble some. He does a LOT in the conditions of an if statement. I have to meditate a bit on the code just to get my head wrapped around it.

Honestly, after a close review of amagic_call() my only thought for refactoring their logic would be to migrate amagic_call into something like amagic_call_or_find(), have it support  a "do not execute mode", and then make both amagic_call() and amagic_find() wrappers around the new function with the appropriate flags. Then they would both use the same internals (from amagic_call), but do slightly different things with the result.

I really think the future of this function comes down to the intent of use. I dont entirely understand the point of having it return a CV, but i tried to stay compatible with what I assume is the spirit of the function. For the use cases I am interested in, and also the use case I think this will be used most for is to determine if some other function would trigger an overload, i dont care what overload it would trigger, thus  I dont care about CV, I dont care which methods were called, and i dont care if overload needs to use more than one method. I would be happy with a function amagic_can() which returns a boolean.  The only reason I added that stuff was because it returns a CV and it seems to me that if you want a CV you probably want to know what method it is (IN/OUT method), and if you want an CV in the rare case you need two there needs to be a way to say so (method2).  But really all of that is contingent on returning a CV. If we returned a bool then we wouldnt care about which method was returned, nor if overload needed to execute two method to complete the operation.

So what was your intent with this function? Are you trying to save time by pre-extracting the CV so you can execute it in advance? I suspect if you want to do that then a generalized function like find_amagic is not what you want, as likely there are only a few cases where that is even possible. It might be better to have a specialized function for your purpose.

book:
Our intent was to answer the question "does this SV have a concat overload?" to figure out how the actual concat in `join` should be done. We didn't need the CV, but we figured that it might be useful for later use. If I remember correctly, we mostly needed this for the delimiter. 

Your description of a "return a boolean instead of running the code" version of `amagic_call` sounds closer to our original intent.

ericherman:
> Your description of a "return a boolean instead of running the code" version of `amagic_call` sounds closer to our original intent.

As can be seen explicitly here: https://github.com/book/perl5/blob/book/rfc0013-join/doop.c#L637-L646

Assuming https://github.com/Perl/perl5/pull/20635 is merged, then the above wrapper function vanishes.

haarg (approved):

PR #20628, opened 2022-12-19
Inline savepv() and related functions, and my_strlcat(), my_strlcpy()

khwilliamson:
These short functions are moved from util.c to inline.h.
savesharedpv() and savesharedpvn() aren't moved because there is a
complication of needing also to move croak_no_mem(), which could be
done, but are these called enough to justify it?


demerphq:
I thought savesharedpvn was mostly redundant after my patches to switch the file and warning data to use RCPV's.

demerphq:
Merging.

Issue #20627, opened 2022-12-19
amagic_find() doesn't handle method fallback properly.


demerphq:
**Description**
The new amagic_find() only checks for one method, and includes a bunch of overhead before it does. This means it isn't really suitable  currently for determining if something "has a string overload" as it does not consult the fallback logic as amagic_call() would. I think we need to change amagic_find() so the method parameter is a *int and we need to change it so it can detect fallback methods as amagic_call() does. I don't think that we should reimplement amagic_call() with amagic_find(), as I think the code in amagic_call() takes into account a lot of independent issues that aren't relevant  to amagic_find(). (eg, amagic_call() knows about two variables for binary operations, and how to reorder arguments, etc, things which arent relevant for single SV api like amagic_find().  However we should ensure that amagic_find() returns true when amagic_call() would actually trigger a method. 

@ericherman, @book, @leonerd and @haarg may be interested in this ticket.

**Steps to Reproduce**
See https://github.com/Perl/perl5/pull/20626 for an example of internals logic using amgic_find() and getting the wrong answer for fallbacks.

**Expected behavior**
amagic_find() should return a CV for an object with fallback =1, a numeric overload, when it is asked for the string_amg method of the SV, it probably should also write the actual method it is returning back into the method argument.

**Perl configuration**
agmagic_find was added in 5.37.7 (very recently: bab261c9114a5ca4aee6acea592e42b95068af70 via https://github.com/Perl/perl5/pull/20604) and has not be used in anger yet. It is marked API. We should not let it be released in a production release with its current interface.

ericherman:
I'd like help creating a failing test case, something a bit more focused than https://github.com/Perl/perl5/pull/20626 

Until then, it would be trivial to 
* add an unused parameter to the API,
* change the `int method` parameter to be an `int *`,
* or simply revert the patch.

While not as focused, maybe it can be tested indirectly. The code was extracted from https://github.com/Perl/perl5/pull/20503 ... perhaps there is a way to trigger the fallback oversight with the join tests?

@book pointed me to https://perldoc.perl.org/overload#Magic-Autogeneration as a way to possibly get fallback for `concat_amg`, but perhaps there is something more straight-forward.

Guidance welcome.

demerphq:
On Mon, 19 Dec 2022, 13:16 Eric Herman, ***@***.***> wrote:

> I'd like help creating a failing test case, something a bit more focused
> than #20626 <https://github.com/Perl/perl5/pull/20626>
>
[...]

PR #20629, opened 2022-12-19
Change function signature to allow for fallback

ericherman:
By changing `int method` to `int *method`, this opens the door for the function to honor fall-back, and inform the caller of which method the CV * is for.

By changing the embed from API to PERL_EXT, this signals that uses are internal, so far. When the fallback issues are resolved it can become an API function again.

Partially addresses #20627

This patch follows the guidance from @demerphq who wrote "This is something we should iterate on, not throw away because it's not perfect."

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -635,7 +635,7 @@ XEop	|bool	|try_amagic_bin	|int method|int flags
 XEop	|bool	|try_amagic_un	|int method|int flags
 Apd	|SV*	|amagic_call	|NN SV* left|NN SV* right|int method|int dir
 Apd	|SV *	|amagic_deref_call|NN SV *ref|int method
-Apd	|CV *	|amagic_find	|NN SV *sv|int method|int flags
+Epd	|CV *	|amagic_find	|NN SV *sv|NN int *method|int flags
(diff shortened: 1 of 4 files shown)


ericherman:
@demerphq, does this seem like a reasonable stop-gap measure?

@book, @leonerd and @haarg

demerphq:
Really sorry about the timing of this. See https://github.com/Perl/perl5/commit/3ecba84330636061a8a97638a4a741a882dba26c for what i think we need. I ended up having an hour or so to work on this, i guess i should have said.

ericherman:
> Really sorry about the timing of this. See [3ecba84](https://github.com/Perl/perl5/commit/3ecba84330636061a8a97638a4a741a882dba26c) for what i think we need. I ended up having an hour or so to work on this, i guess i should have said.

I'm happy to close this PR and go with that one.
I'll note that we still lack real tests for this in either case.

demerphq:
Well that PR implicitly checks the string_amg case in its tests for @INC. But true.

ericherman:
Superseded by: https://github.com/Perl/perl5/pull/20635

PR #20630, opened 2022-12-19
harness: handle directories serial/parallel mixture

khwilliamson:
A few test file directories have some test files that must be executed
     serially, while others may be in parallel.  Prior to this commit, all
     such were executed serially.
     
     This commit changes to look for and handle such directories.  It makes
     the safe assumption that the serial files act as setup for the parallel
     ones, so all such must be done first
     
     Here are two cases where this helps.
     
     IO::Compress has a large number of serial test files, then a few that
     can be executed in parallel.  It often is the last directory to finish.
     And the suite isn't over until each file in it is finished.  Those last
     few, prior to this commit, executed one after another, when the other
     cores were idle.  Now they can execute in parallel.
     
     Test::Simple has many tests, but just a few that need to execute
     serially.  Prior to this commit, they all had to execute serially.

### DIFF
--- a/t/harness
+++ b/t/harness
@@ -8,8 +8,14 @@ BEGIN {
     @INC = '../lib';              # pick up only this build's lib
 }
 
-my %must_be_executed_serially = map { $_ => 1 }
-                                    qw(../cpan/IO-Zlib/t ../ext/File-Find/t);
+my %must_be_executed_serially =
+    map { $_ => 1 } (
+                     '../cpan/IO-Zlib/t',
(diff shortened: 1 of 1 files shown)


demerphq (approved):
Love it. Thanks for taking care off the 0 dependency like I asked in the other PR.

demerphq:
I am itching to merge this @khwilliamson :-)

demerphq:
Ok. I cant wait any longer. :-)

PR #20631, opened 2022-12-20
Minor updates for release documentation

richardleach:
These are mostly additions for clarity or extra examples.

Some advice has been removed from "update INSTALL", because the wording within INSTALL has been made less version-specific, requiring fewer updates per-release.

### DIFF
--- a/Porting/release_managers_guide.pod
+++ b/Porting/release_managers_guide.pod
@@ -257,9 +257,12 @@ F<Porting/Maintainers.pl>, as well as checksums updated via:
 
 In most cases, once a new version of a distribution shipped with core has been
 uploaded to CPAN, the core version thereof can be synchronized automatically
-with the program F<Porting/sync-with-cpan>.  (But see the comments at the
-beginning of that program.  In particular, it has not yet been exercised on
-Windows as much as it has on Unix-like platforms.)
+with the program F<Porting/sync-with-cpan>. For example:
+
+ perl Porting/sync-with-cpan Archive::Tar
+
+(But see the comments at the beginning of that program.  In particular, it has
+not yet been exercised on Windows as much as it has on Unix-like platforms.)
 
 If, however, F<Porting/sync-with-cpan> does not provide good results, follow
 the steps below.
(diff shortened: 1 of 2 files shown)


jkeenan:
> These are mostly additions for clarity or extra examples.
> 
> Some advice has been removed from "update INSTALL", because the wording within INSTALL has been made less version-specific, requiring fewer updates per-release.

Can we get some feedback on these suggestions from other recent monthly dev release managers?

Porting/release_managers_guide.pod
```diff
@@ -491,10 +494,6 @@ version number.
 
 Review and update INSTALL to account for the change in version number.
 INSTALL for a BLEAD-POINT release should already contain the expected version.
-The lines in F<INSTALL> about "is not binary compatible with" may require a
```
Corion, line 494: This part applies to MAINT releases (e.g. 5.36.0 and 5.36.1 ) for which we strive to keep binary compatibility. So this part should be kept. But maybe a rephrasing like the following is better?

For MAINT releases, the lines in F<INSTALL> ...

richardleach: I'm not clear on what lines in INSTALL may require a correct choice of earlier version. 
The two sections that seem to touch on it are:
```
=head3 Compatibility with earlier versions

B<WARNING:> This version is not binary compatible with earlier versions
of Perl.  If you have built extensions (i.e. modules that include C code)
using an earlier version of Perl, you will need to rebuild and reinstall
those extensions.
```
^ which doesn't mention specific versions at all. (Although I feel like it used to do?)

```
=head1 Coexistence with earlier versions of perl 5

Perl 5.37.8 is not binary compatible with earlier versions of Perl.
In other words, you will have to recompile your XS modules.

In general, you can usually safely upgrade from one stable version of Perl
(e.g. 5.30.0) to another similar minor version (e.g. 5.30.1) without
re-compiling all of your extensions.  You can also safely leave the old
version around in case the new version causes you problems for some
reason.
```
^ which only mentions the version number being released and then gives
5.30.x versions as an example.

My concerns about retaining this section are:
* As per above, it's now not clear to me where any changes are necessary
* If there is still a requirement for the release manager to select a correct choice of earlier version to declare incompatibility with, there is no guidance or script for the release manager to use in order to actually do this.

richardleach: @Corion / @karenetheridge - any thoughts on the above, please?

demerphq:
@Corion and @karenetheridge, please note that @richardleach is waiting on your feedback on this PR.

@richardleach note that the PR is not passing CI. did you want to modernize it?

Porting/release_managers_guide.pod
Corion: I'm not sure how to properly rephrase my comment above to make it clearer. The sections you mentioned are the sections that require a correct choice of earlier (stable) version.

The versions in the example can remain unchanged but they could also be bumped to the latest stable major version (5.36 currently). I'd consider that makework though.

I still think that this is only relevant for a MAINT release and irrelevant for a BLEAD-POINT release as our attempts of binary compatibility only cover stable releases and their subversions.

richardleach: I'm still unclear about what exactly should be changed from the templatein the "Compatibility with earlier versions" section, other than removing it completely or totally freestyling something. Regardless, I've reinstated that text with the minor leading tweak as suggested.

jkeenan:
> @Corion and @karenetheridge, please note that @richardleach is waiting on your feedback on this PR.
> 
> @richardleach note that the PR is not passing CI. did you want to modernize it?

@Corion, @karenetheridge, @richardleach , can you respond to the questions above?  (If we get no response I think the ticket should be closed.)

Corion:
The changes look good to me now

PR #20636, opened 2022-12-21
sv.c - add support for HvNAMEf and HvNAMEf_QUOTEDPREFIX formats

demerphq:
They are similar to SVf and SVf_QUOTEDPREFIX but take an HV * argument and use HvNAME() and related macros to extract the string. This is helpful as it makes constructing error messages from a stash (HV *) easier.

See https://github.com/Perl/perl5/pull/20626 where this is used.

### DIFF
--- a/ext/XS-APItest/APItest.pm
+++ b/ext/XS-APItest/APItest.pm
@@ -4,7 +4,7 @@ use strict;
 use warnings;
 use Carp;
 
-our $VERSION = '1.29';
+our $VERSION = '1.30';
 
 require XSLoader;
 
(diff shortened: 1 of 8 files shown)


demerphq:
@leonerd @tonycoz, anyone else? any objections on this? I was thinking of calling it HvNf but that was pretty ugly to type. 

this is trivial and useful so ill merge soon unless i hear some feedback suggesting the contrary.

leonerd (approved):
LGTM. And I think the name is fine - it's not so long it feels awkward. `HvN` is a bit too short to be obvious what it means.

demerphq:
ta! merging.

demerphq:
Repushed with a version bump. Conflict resolution didn't notice that two patches bumping a module to the same version arent actually the same change, so we would have failed the t/porting/cmp_version.t test if i merged directly.

PR #20637, opened 2022-12-21
Add a new %{^HOOK} hash, similar to %SIG, and add support for "require__before" and "require__after" hooks. 

demerphq:
This PR adds a new `%{^HOOK}` hash, similar to `%SIG`, but for hooking keywords. The hash is magic in a similar way as `%SIG`, in that it validates its contents on store and actually maps to internal variables which store the values instead.  The first two members of the hash are `require__before` and `require__after`. Members are expected to be of the form `$keyword . "__" . $phase`, where `$keyword` is the keyword that executes the hook, and `$phase` is when the hook is executed (Currently either "before" or "after"). The initial user visible implementation includes support for a hook which is executed before require, and a hook which is executed after require. The `require__before` hook may return a callback that is also executed after the require, but can be constructed in such a way that it shares state with the before hook that executed. The before hooks execute before nearly any other `require` related logic, including checking if the required item was already `required`, and the after hooks executed after the `require` completed, regardless as to whether the require died or was successful.

With the `required__before` hook we could fix Devel::TraceUse to not alter the stack when it is in use. It could also be used to detect circular dependencies, log all used modules, or track the memory and time taken to require a given module (with or without its dependencies).

In order to implement this there are some nice goodies from an internal point of view. `mortal_destructor_sv()` and `mortal_svfunc_x()`  (and their macro wrappers `SAVEMORTALDESTRUCTOR_SV()` and `SAVEMORTALSVFUNC_X()`) can be used to execute a callback at the end of the current statement (either Perl or C callback respective). Also there is the `SvREFCNT_dec_set_NULL()` macro which can be used to decrement an `SV*` pointer and set it to NULL in a single statement, and a related function `SvREFCNT_dec_ret_NULL()` which decrements an SV's refcount and returns NULL, which is used by the `SvREFCNT_dec_set_NULL()` macro but can also be helpful in cases where a function wants to do this to an argument and then return NULL. There are also various other minor improvements to `regen/mg_vtable.pl` (trimming trailing whitespace on generated documentation and renaming of some confusingly named internal variables) and improvements to the documentation of the related functions in `pod/perlguts.pod`.

Everything has documentation, including the new utility functions and macros.

@haarg @ap @book


perl.c
```diff
@@ -938,6 +938,10 @@ perl_destruct(pTHXx)
     PL_warnhook = NULL;
     SvREFCNT_dec(PL_diehook);
     PL_diehook = NULL;
+    if (PL_requirehook) {
+        SvREFCNT_dec(PL_requirehook);
+        PL_requirehook = NULL;
+    }
```
tonycoz: SvREFCNT_dec() correctly handles NULL here.

pp_ctl.c
```diff
@@ -4089,6 +4089,19 @@ S_require_version(pTHX_ SV *sv)
...
 }
 
+static void
+S_call_post_require_hook(pTHX_ SV *hook_sv) {
+    dSP;
+    sv_2mortal(hook_sv);
+    ENTER_with_name("call_POST_REQUIRE");
+    SAVETMPS;
+    PUSHMARK(SP);
+    int count = call_sv(hook_sv, G_VOID);
```
tonycoz: count is otherwise unused and gcc warns about it

demerphq: fixed, thanks.

perl.c
demerphq: Fixed, thanks.

leonerd:
I've always thought that the abuse of `%SIG` which `__DIE__` and `__WARN__` make is incredibly ugly; I would much have preferred those two to have been written `${^DIEHOOK}` and `${^WARNHOOK}` so that every element of `%SIG` really does relate to a real POSIX signal.

It's rather too late to change those now, but can we avoid adding any more? I.e. can this please be spelled `${^REQUIREHOOK}` instead?

demerphq:
> I.e. can this please be spelled ${^REQUIREHOOK} instead?

FWIW, my original implementation worked like that. But it was more work and it felt unperlish to me so i switched to `%SIG`. While I agree with your sentiment, and likely I would agree that if `$SIG{__WARN__}` was `${^WARN}` or something it might be better, but that boat has sailed and I think that having /some/ hooks in magic caret vars and /some/ in `%SIG` is worse than having them all in `%SIG`, the element of surprise is reduced at the very least.  There is also a small advantage with `%SIG` that you can enumerate all the hooks that are in effect by walking `%SIG`, which has some level of utility and affords a certain sense of future portability. For instance if its in %SIG then a generic "disable all hooks" is just `local %SIG=();` (ignoring `@INC` hooks), with hooks all over the place that becomes rather more difficult and impossible to future proof against.

There is another advantage at the implementations level that adding support for validating a new special key in %SIG is nearly trivial, but adding a new var with new magic to front an internal state is rather more complex and invasive. I found it interesting how much simpler the code was when it was based on %SIG.

So frankly I'd prefer that we don't use a caret var for this.  If PSC decides we must then so be it, it can be done, but I don't think its as good an idea as it sounds at first. And I mean that literally, my unpushed original code for this used `${^REQUIRE_HOOK}` much as you suggested, but after playing with it I came to the conclusion that it just felt weird to have `$SIG{__WARN__}` on one hand and `${^REQUIRE_HOOK}` on the other.

demerphq:
@leonerd one option might be to introduce a `%{^HOOK}` variable. We could mirror warn and die between the two hashes. Then use that in new hooks.

leonerd:
Ahyes; that sounds like a better forward plan; a new place to put hooks would be nice.

In that case, since we'd want to move WARN and DIE out of the current place anyway, it probably makes little difference whether we add REQUIRE in to SIG for now anyway. Might as well continue on this PR as it is for now and move all of them later on.

ap:
> I've always thought that the abuse of `%SIG` which `__DIE__` and `__WARN__` make is incredibly ugly

I agree it would be cleaner to only have `%SIG` concerned with POSIX signals, and have those other hooks live elsewhere. But they do make a sort of sense to me as well, considering they too are a program interruption *somewhat* akin to a signal.

A `__REQUIRE__` hook in `%SIG` is a very weird place for it though. The only reason to put it there is that `__DIE__` and `__WARN__` happen to be there already, which also happens to make the implementation easier. Without that path dependence, it would seem a bizarre design decision.

> One option might be to introduce a `%{^HOOK}` variable. We could mirror warn and die between the two hashes. Then use that in new hooks.

This, however, I have a severe allergic reaction to. Please don’t. Code which uses the new interface would not work on older perls, not because older perls lack capabilities to support the feature being used, but simply because the way to access the capability changed name. To put it another way: because fashion. Worse, dealing with these hooks is relevant inside some rather plumbing-level modules, where forcing churn is just going to impose more forced perl version obsolescence on the ecosystem.

The same reasoning is going to apply in the future, of course, so adding a `%{^HOOK}` variable now – but for `require` only, with no mirroring of the `%SIG` hooks – seems like a good idea, if that seems like something there will be more use for later. (Btw, also, can we have lowercase keys in that case? (Or rather, simply use the name of the function being hooked, verbatim.) That `%SIG` keys are all uppercase makes sense, but IMO `%{^HOOK}` should not be beholden to that precedent.)

demerphq:
@ap: to you it is a category error to put __REQUIRE__ with __DIE__ and __WARN__ in %SIG. But i dont agree with you really, I see a "signal" generated when we are about to compile more code to be compatible with the notion of a generic signal, as much as __DIE__ or __WARN__ do anyway.  Remember that require is triggered *invisibly* by use, and internally by using of various functionality or variables.  So when someone "uses" code, having a "signal" triggered at the start of that process isn't that weird to me. Certainly to me it is not so obviously wrong that it seems like a "bizarre design decision". To me it seems a design decision well within the general concept of a signal triggered when perl does something invisibly under the hood. Especially as the require hook really is a true /signal/, unlike DIE or WARN, in that it cannot affect the execution of the require other than to throw an exception, both DIE and WARN allow you to modify the behavior of the die/warn functions.  So they arent just signal handlers but also mutators.

The only reason you seem to be distinguishing DIE and WARN from REQUIRE is that DIE and WARN are part of an exception handling process. To me the thing that the %SIG keys have in common is that they can all be triggered automatically by perl or the system itself when things happen. Perl requires modules behind the scenes, just as it throws exceptions, and thus it seems to me that they have more in common than they differ, and in fact, that REQUIRE, not being a mutator, actually has more in common with other signal handlers than it does with DIE or WARN and that it is really those two which are exceptional. [1]

With regard to %{^HOOK} or whatever we call it (if we go ahead with it), i was thinking we would just "tie" the __DIE__ and __WARN__ keys in %SIG to their equivalents in %{^HOOK}, not necessarily removing the keys from %SIG until there has been a very very long deprecation cycle. The only thing that might break back compat is someone using only %{^HOOK} instead of %SIG, but that is the same as someone using defer{}. It would also be possible to create a shim to make %{^HOOK} work properly on older perls if we wanted.

But i take your point that introducing a %{^HOOK} variable may just end up making things worse not better, that was part of my reason for saying "the only thing worse than putting this in %SIG while WARN and DIE are there is to have a different hook not in SIG." The precedent has been established, no matter how explicable we find that precedent, and we should just stick with it. 

[1] the return behavior of the __REQUIRE__ hook is a bit exceptional. but the actual location of the hook to me not so much.

demerphq:
@ap: Consider this case. opening a scalar file buffer auto-requires PerlIO::scalar. 

```
$ ./perl -Ilib -le'$SIG{__REQUIRE__}= sub { print "__REQUIRE__: @_" }; open my $fh,"<", \"" or die;'
__REQUIRE__: PerlIO.pm
__REQUIRE__: PerlIO/scalar.pm
__REQUIRE__: XSLoader.pm
__REQUIRE__: strict.pm
```

To me this feels very much like what perl level %SIGnals do.  Tell us when something internal has happened that we might want to know about.

demerphq:
So are there any objections to this moving forward?

demerphq:
This feels warnocked. I would like to merge this for the next release.

khwilliamson (approved):

tonycoz (approved):

haarg:
> @ap: to you it is a category error to put `__REQUIRE__` with `__DIE__` and `__WARN__` in %SIG. But i dont agree with you really, I see a "signal" generated when we are about to compile more code to be compatible with the notion of a generic signal, as much as `__DIE__` or `__WARN__` do anyway. Remember that require is triggered _invisibly_ by use, and internally by using of various functionality or variables. So when someone "uses" code, having a "signal" triggered at the start of that process isn't that weird to me. Certainly to me it is not so obviously wrong that it seems like a "bizarre design decision". To me it seems a design decision well within the general concept of a signal triggered when perl does something invisibly under the hood. Especially as the require hook really is a true /signal/, unlike DIE or WARN, in that it cannot affect the execution of the require other than to throw an exception, both DIE and WARN allow you to modify the behavior of the die/warn functions. So they arent just signal handlers but also mutators.

But all of the other %SIG entries do change the execution, just by being installed. For POSIX signals, installing a %SIG entry prevents the system default signal handler from being executed.

> The only reason you seem to be distinguishing DIE and WARN from REQUIRE is that DIE and WARN are part of an exception handling process. To me the thing that the %SIG keys have in common is that they can all be triggered automatically by perl or the system itself when things happen. Perl requires modules behind the scenes, just as it throws exceptions, and thus it seems to me that they have more in common than they differ, and in fact, that REQUIRE, not being a mutator, actually has more in common with other signal handlers than it does with DIE or WARN and that it is really those two which are exceptional. [1]
> 
> With regard to %{^HOOK} or whatever we call it (if we go ahead with it), i was thinking we would just "tie" the `__DIE__` and `__WARN__` keys in %SIG to their equivalents in %{^HOOK}, not necessarily removing the keys from %SIG until there has been a very very long deprecation cycle. The only thing that might break back compat is someone using only %{^HOOK} instead of %SIG, but that is the same as someone using defer{}. It would also be possible to create a shim to make %{^HOOK} work properly on older perls if we wanted.
> 
> But i take your point that introducing a %{^HOOK} variable may just end up making things worse not better, that was part of my reason for saying "the only thing worse than putting this in %SIG while WARN and DIE are there is to have a different hook not in SIG." The precedent has been established, no matter how explicable we find that precedent, and we should just stick with it.
> 
> [1] the return behavior of the `__REQUIRE__` hook is a bit exceptional. but the actual location of the hook to me not so much.

I don't think that require being used "behind the scenes" is really justification for it being handled differently from hooking into any other `CORE::` function. And I think the fact that `__REQUIRE__` functions entirely differently from all the other `%SIG` entries means it doesn't belong there. The magic return value is different from any other `%SIG` entry, and I don't think the desire to wrap require justifies the special case.

The original problem leading to this PR is that wrapping `CORE::GLOBAL::require` is problematic. This also applies to a number of other CORE functions, such as `open` and `system`. Maintaining context and correct parsing is hard or impossible. So I'd like to see a more general solution for attaching before/after actions to core subs. A separate variable from `%SIG` seems more appropriate. Since the hooks that this PR wants to implement work differently from the existing `__DIE__` and `__WARN__` handlers work, I don't think they need to be moved or aliased into the new feature.

haarg:
I'll add that while I'd like to see a more general solution for this type of hook, I think it's fine if there's an initial implementation that only works for require. But I'd like the shape of it to allow future extension, and I don't think `%SIG` is the right place for that.

demerphq:
Hi @haarg. I have outlined why i dont think a separate variable is a good idea. So while I respect your opinion, if that is the strongest point you have here, then I think we will have to just agree to disagree.  IMO The only thing worse than having `__WARN__`, `__DIE__` and `__REQUIRE__` be in %SIG is having `__WARN__` and `__DIE__` be in a different data structure from `__REQUIRE__ `itself. Having this live in %SIG means that %SIG=(); clears the hooks and restores "normal behavior".  If we introduce a new variable for this, then coders need to know about both. I think that is clearly worse than whatever conceptual friction you have with putting this stuff alongside the other hooks in %SIG. 

I actually do agree that it would have been better if Larry had put the __DIE__ and __WARN__ hooks somewhere else, and he had I would have put the __REQUIRE__ hook in the same place, but that boat has sailed. Lets not make the problem worse.

Also, I do agree that I would not be pushing __REQUIRE__ at all if it was sane to override require and other core ops properly. But it isn't and it seems likely it wont be for some time, if ever.  So I think we should go ahead with __REQUIRE__ as is.

Leont:
I don't quite like it in `%SIG` either (and hate the existing exceptions anyway).

demerphq:
@leont yeah, I understand. I dont like a number of design decisions in Perl. :-) But once they are baked in we have to consider this from the point of view of "least worst" options, not as though we have a clear field.  And I think that the least worst option here is just swallow our dislike of this. The other option opens up its own can of portability worms. Specifcally that code that currently exists and does:

```
local %SIG;
```
wll continue to work as it used to: disabling all signal handlers, OS level and Perl level.

If we introduce a new var that is no longer true, and it becomes yet another thing people need to deal with.

So the question here isnt "is this the ideal way to do this", it is "what is the least sucky way to do this". And IMO the objections about it being in %SIG are a matter of taste, which is important, but it lacks technical foundation.

Grinnz:
I'm not sure what you mean. @haarg 's objections were entirely technical (and seem reasonable to me).

wchristian:
Gotta agree that rather than messing more with %SIG, which should be posix signals, the optimal solution here would be a completely separate mechanism for hooking perl core function, which could then also include die and warn and permit future hooking on those without the issues %SIG replacements present, maybe allowing phasing that out in some far future.

demerphq:
@grinzz I disagree. I dont think @haarg made a technical argument, most of the discussion here about %SIG is one of taste, not of technical reasons.  There is NO technical reason this hook can't be in %SIG, the proof is in the patch itself. Not liking it being there is a matter of personal preference not of technical merit.  The focus on %SIG being only about POSIX signals is belied by the fact that Larry put other "signals" in that hash, and is also not a technical justification. The API signature of the callback also isn't a technical reason.  Having said that I am not ignoring this feedback, and I am looking at making a `%{^HOOKS}` hash which contains `__WARN__` and `__DIE__` and `__REQUIRE__` in such a way that if you set `$SIG{__WARN__}` it is mirrored in `${^HOOKS}{__WARN__}` and vice versa.  However @ap has raised what I consider are serious objections to this approach which ARE technical which IMO are fatal to that alternative, and I think he is right, but I am going to try to do it anyway so we can discuss it, but i still think this PR is the least worst approach.

@wchristian the best expert on this subject that i know of is @iabyn and when i consulted him his answer was roughly "goto is already *terrifying* code, I wouldn't want to make it worse". I paraphrased, but he did say "terrifying code", and when HE says something is terrifying I take notice, I reckon he has forgotten more internals than most of us ever knew, and every time he has said it in the past and I ended up looking at the code he was right. So it seems that until someone resolves that we have a problem. require is especially problematic because so much code related to require inspects the stack to find who called them for importing, thus you cant shim require safely without breaking stuff.

@ether this PR isn't a draft. I don't think marking it as a draft without speaking to me first is appropriate. Please don't do that again. Thanks.

Grinnz:
> @Grinzz I disagree. I dont think @haarg made a technical argument, most of the discussion here about %SIG is one of taste, not of technical reasons. There is NO technical reason this hook can't be in %SIG, the proof is in the patch itself. Not liking it being there is a matter of personal preference not of technical merit. The focus on %SIG being only about POSIX signals is belied by the fact that Larry put other "signals" in that hash, and is also not a technical justification. The API signature of the callback also isn't a technical reason.

I'm not sure how objections on a purely language design basis are "not technical". Obviously language design is largely opinion, but it's not just personal preference.

demerphq:
>I'm not sure how objections on a purely language design basis are "not technical".

When they come down to "I dont like that" and don't take into account the larger set of interlocking issues that are already present in the language.  I have already explained the problem that i see as devastating for the alternative, as has @ap.  In short, code that does:

```
local %SIG;
```

or the equivalent to get rid of all the signals and handler hooks would be broken because we have added a new hook in a different place. That is a backwards compatibility nightmare.  To quote @ap in response to the idea of %{^HOOKS}:

> This, however, I have a severe allergic reaction to. Please don’t. Code which uses the new interface would not work on older perls, not because older perls lack capabilities to support the feature being used, but simply because the way to access the capability changed name. To put it another way: because fashion. Worse, dealing with these hooks is relevant inside some rather plumbing-level modules, where forcing churn is just going to impose more forced perl version obsolescence on the ecosystem.

We aren't in a green field situation here where have unlimited freedom of movement where taste can prevail. We are renovating a building in a built up space, we don't have the luxury of saying "lets put a fountain here" as someone already put a car-park there.  I havent seen anyone address this point. They just say "i dont like that", but they dont address the alternatives, and IMO the alternatives all suck more than the original proposal.

Its fine to say `__WARN__` and `__DIE__` shouldn't be in %SIG.  If someone had said that when those hooks were introduced we wouldnt be having this conversation. But they didn't and now we have to live with that decision. The technical argument *against* putting it somewhere else is IMO stronger, actually devastating to the alternative, and since the people arguing for the alternative have not responded to that devastating argument it seem reasonable to dismiss them as personal preference not justified by technical merit. When someone like @ap dismisses something with "because fashion" I listen.

ap:
@demerphq

> to you it is a category error to put `__REQUIRE__` with `__DIE__` and `__WARN__` in `%SIG`.

You are misstating my argument. I said `__REQUIRE__` makes any sense at all in `%SIG` only because `__DIE__` and `__WARN__` are already present there.

> Remember that require is triggered invisibly by use, and internally by using of various functionality or variables.

By that logic, any internal function call inside a built-in function is a signal. Is that your argument? Or is there some additional condition you didn’t spell out? If that’s the argument, I don’t think that’s a useful idea of what a signal is.

> Certainly to me it is not so obviously wrong that it seems like a "bizarre design decision".

Again, that was my assertion about a hypothetical present in which `%SIG` contained only POSIX signal handlers and not `__DIE__` and `__WARN__`. In that scenario I contend that it would have never occurred to you to add a `__REQUIRE__` in there.

> The only reason you seem to be distinguishing DIE and WARN from REQUIRE is that DIE and WARN are part of an exception handling process.  To me the thing that the `%SIG` keys have in common is that they can all be triggered automatically by perl or the system itself when things happen.

But any internal function call inside perl qualifies for the criterion you propose. I think `__DIE__` does make a kind of sense to think of as a signal insofar as many signals will, unless handled, abort the program. On that count and on second thought, then, even `__WARN__` only makes sense as a signal in light of `__DIE__`.

> REQUIRE, not being a mutator, actually has more in common with other signal handlers than it does with DIE or WARN

A `TERM` handler can prevent the program from terminating on a `TERM` signal same as a `__DIE__` handler can prevent it from terminating on exception. Not sure why you claim a lack of commonality there.

> It would also be possible to create a shim to make `%{^HOOK}` work properly on older perls if we wanted.

Right. You have a point, and that does lessen my concern. It wasn’t backward compatibility I was worried about so much as forward compatibility, i.e. old perls being able to run newly written code, which requires that even newly written code use `%SIG` rather than `%{^HOOK}`. (“Because fashion” was my shorthand for making interfaces “prettier” based on hindsight without adding capability or other benefits while breaking compatibility in the process.) But you are right, it can also shim in a `%{^HOOK}` on old perls (a polyfill, in Javascript lingo).

> I take your point that introducing a `%{^HOOK}` variable may just end up making things worse not better

Actually I don’t mind that at all! In fact I think it’s a good idea. I am just averse to the immediate impulse of “well obviously we should rearrange all the furniture for consistency now that we have this new knob in place”. I suppose my “this I have a severe allergic reaction to” may have given the impression I was talking about `%{^HOOK}` as such but I really was just talking about moving `__DIE__` and `__WARN__` to it or even just mirroring them; in the same comment I went on to say that introducing `%{^HOOK}` now is probably best.

> IMO The only thing worse than having `__WARN__`, `__DIE__` and `__REQUIRE__` be in `%SIG` is having `__WARN__` and `__DIE__` be in a different data structure from `__REQUIRE__` itself.

If `__REQUIRE__` is to be the only hook ever added I’ll go along with this.

Otherwise I don’t think this is bad at all, quite the opposite. Yes, it will be a bit weird early on when `require` is the only thing for which hooks exist – but if we do go on adding hooks for `open` and `system` etc., it will make a lot more sense for them all to live together in their own separate spot dedicated to hooks for builtins rather than to have more and more entries in `%SIG` that have less and less in common with signals. Imagining a `%{^HOOK}` with a dozen different entries, it doesn’t feel weird at all to me to also keep having `__DIE__` and `__WARN__` over in `%SIG` (even without any mirroring) – at best kind of appropriate, at worst a bit of trivia to memorize (which no evolving system can entirely escape). Whereas if `__REQUIRE__` is added to `%SIG` now, the argument that we should follow precedent will only be that much stronger for the next hook, making @leonerd ever unhappier. 🙂

> I dont think @haarg made a technical argument, most of the discussion here about `%SIG` is one of taste, not of technical reasons. There is NO technical reason this hook can't be in `%SIG`

How is “the magic return value is very different from everything else” not clearly a technical argument?

Is “is it technically possible” actually the only criterion for what constitutes a technical argument? Because there is also NO technical reason we can’t, I dunno, move all global variables into `%SIG` as keys, say. That would then also be just a matter of taste. In which case taste clearly matters. 🙂

> `local %SIG`

I’m surprised you see this as an argument in favour of the proposal! To me this is *by a long shot* the clearest and cleanest argument that hooking `require` should go through a different mechanism than POSIX signal handlers. In fact to me this even somewhat strengthens the argument in favour of `__DIE__` in `%SIG` (and in fact in `%SIG` only) even after `%{^HOOK}` exists (though not for `__WARN__`, or at least that one is much less clear-cut to me).

Particularly once you start thinking about hooks for `open`, `system`, etc., it seems both very surprising and very annoying to me not to be able to easily disable all POSIX signal handlers without also disabling *every single builtin’s* hooks (as well as vice versa – though that seems a less likely use case).

ap:
tl;dr: I never argued against `%{^HOOK}` itself, just against moving `__DIE__` and `__WARN__` there.

ap:
> How is “the magic return value is very different from everything else” not clearly a technical argument?

In fact, I find this point much more important than all of the other arguments so far.

We have been talking about “the `__REQUIRE__` hook” this whole time but really this is two hooks: a before-hook and an after-hook. They just masquerade as a single hook, since this interface doesn’t allow you to write an after-hook without wrapping it inside a before-hook which returns it (and, if you didn’t need a before-hook anyway, otherwise does nothing). Then the caller of the before-hook examines its return value in order to trigger additional behavior down the line if it is a coderef.

This design is a mistake. I say this because I made this mistake. This is exactly how the [Plack::Middleware::Rewrite 1.x](https://metacpan.org/release/ARISTOTLE/Plack-Middleware-Rewrite-1.101/view/lib/Plack/Middleware/Rewrite.pm) interface worked – it let you pass a callback to rewrite the request, and that callback could return a coderef which would be called to rewrite the response. It was *awful* – both on the caller side and the callee side. In 2.x I split this into two independent callbacks, one for the request, one of the response, and that has made it much nicer to use.

The same will be true here.

Of course once you start contemplating two hooks, suddenly `%SIG` looks very clearly like the wrong place…

For that matter, maybe we should give some thought to being able to have multiple callbacks assigned to the same hook. You could always do this by closure-chaining of course (save the current callback to a variable, call it in the new callback you are adding).

But given a chance to look at this fresh, maybe it is worth considering some kind of interface that allows multiple callbacks natively, without closure-chaining them, so they can also be removed or disabled in arbitrary order.

Note this is another design aspect where having independent before- and after-hooks will make things *a lot* less grungy to use, even if we do leave it at closure-chaining. Just think about how you would write an after-hook in such a way that it is chainable in the combined before-but-also-after-hook interface.

In sum:

1. Make the two hooks actually independent
2. Give some thought to a multi-callback interface for each hook

Ovid:
I don't have any particular comments on whether this should be `$SIG{__REQUIRE__}`, `${^REQUIRE}`, `${^REQUIREHOOK}`, or so on, but this feature would be incredibly useful. It's one of those features which, when you need it, it's very frustrating not to have it. I think `${^HOOK}` is nice for no other reason than it's clearer (to me) what's happening here.

Nonetheless, regardless of the implementation, I'm looking forward to seeing this go in. What needs to be resolved to make that happen?

demerphq:
I have repushed this to use `%{^HOOK}`. Specifically `${^HOOK}{require__before}` and `${^HOOK}{require__after}`. 

I have defined that the naming convention for these hooks is "${keyword}__${phase}" (that is two underbars) where $phase is either "before" or "after". I chose to put  the phase after the keyword even though it sounds a bit "yoda-ish", because I believe the important part of the key name is the keyword being hooked, and I think it makes more sense if the keys are grouped together (when sorted) by the keyword they hook not by the phase they are executed in. Also the original proposal used the key `__REQUIRE__` which imo was very easy to grep for and eye-catching. I wanted something similar, and i believe "require__before" and "require__after" are easy to grep for. I chose to use two underbars because conceivable one day we might want to hook something that has underbars in its name already (although I am hard pressed to think of an example. Something in builtin perhaps?) so using two underbars keeps the phase distinct from the thing being wrapped. 

Since the original version of this patch I have made one significant change in that the `require__before` hook can change the name of the file that is loaded by altering  the `$_[0]` it was provided.

I have **not** removed the "post-action closure" for the `require__before` hook as suggested by @ap, although I have added support for a `require__after` hook.  While I personally do not see any value in a `require__after` hook I am open minded that maybe someone out there really would like to only do something after each require statement completes without doing anything before each require statement starts. So if people really want to have fully separate before and after hooks then they have them.  However for people who DO want to do something statefully coupled both before and after I believe very strongly that the "return a closure" behavior is the **best** way to do so. Anything else brings up awkward and painful edge cases that simply don't exist with the "return a closure" behavior I have implemented.

1. My original use case for this was to measure the time to load and memory used by executing a given require statement.  This necessitates that state is shared between the before and after actions in a tightly coupled way. Forcing people to use two separate callbacks for before and after complicates things in a variety of ways, especially in light of cyclic requires.  Having the before callback return a closure that contains the state it needs to compute time elapsed or memory conditions prior to the require means that this is safe and reliable and easy, without contortions.

2. I want to be able to detect cyclic dependency chains. Perl handles this fine at certain levels, but it can also horrible break code at other levels, and it should be easy to detect it with an appropriate `require__before` hook.

3. With regard to the comparison with Plack::Middleware::Rewrite that @ap made I do not consider it particularly apposite. Plack requests are not recursive, and do not form a tree.  Middleware layers may form a tree, but a given plack request does not generate an arbitrary number of additional plack requests like a require statement very often does and it definitely shouldn't create a cycle, so the comparison to me falls down. Also "rewrite" implies that data is changed, the post-action require hooks, and `require__after` hooks do not modify anything. The most that can happen is that the `require__before` hook can change the filename that will be loaded, or it can die. The post-action hook does not modify anything, and it executes inside of an eval, so it cannot alter what has already happened.

mg.c
```diff
@@ -1883,6 +1885,79 @@ Perl_magic_setsigall(pTHX_ SV* sv, MAGIC* mg)
...
 }
 
+int
+Perl_magic_gethook(pTHX_ SV *sv, MAGIC *mg)
+{
+    PERL_ARGS_ASSERT_MAGIC_GETHOOK;
+    PERL_UNUSED_ARG(sv);
+    PERL_UNUSED_ARG(mg);
+    return 0;
+}
```
tonycoz: If you're not doing anything on get you can leave it out (see the hints hooks for example.)

demerphq: Ack. Ill remove this. I was thinking maybe it would come in useful and so it might as well be there for when it was needed, but i guess if someone wants it they can add it themselves.

demerphq: Anyway, done now. Thanks for the feedback!

tonycoz:
You need a clear handler for the aggregate too:
```
$ ./perl -Ilib -E '${^HOOK}{require__before} = sub { say "hello"; () }; %{^HOOK} = (); require Cwd'
hello
hello
hello
hello
hello
hello
hello
hello
hello
$ ./perl -Ilib -E '%{^HOOK} =(require__before => sub { say "before"; () }); %{^HOOK} = ( require__after => sub { say "after" }); require warnings'
before
after
$
```
Also, I don't think sethookall is ever called (we don't SvSETMAGIC() the hash, just elements, while hv_clear() calls mg_clear() on the hash):
```
$ gdb --args ./perl -Ilib -E '%{^HOOK} =(require__before => sub { say "hello"; () }); require Cwd'
...
Reading symbols from ./perl...
(gdb) b Perl_magic_sethookall
Breakpoint 1 at 0x22f61c: file mg.c, line 1945.
(gdb) r
Starting program: /home/tony/dev/perl/git/perl6/perl -Ilib -E %\{\^HOOK\}\ =\(require__before\ =\>\ sub\ \{\ say\ \"hello\"\;\ \(\)\ \}\)\;\ require\ Cwd
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
hello
hello
hello
hello
hello
hello
hello
hello
hello
[Inferior 1 (process 2574055) exited normally]
```
Sorry this didn't occur to me when I looked at it yesterday.

demerphq:
> Sorry this didn't occur to me when I looked at it yesterday.

No problem. I was cribbing^Wcargo-culting from %SIG. I am not sure what i got wrong tho. If you have more details they would be appreciated.

tonycoz:
> No problem. I was cribbing^Wcargo-culting from %SIG. I am not sure what i got wrong tho. If you have more details they would be appreciated.

I think just adding the non-elem clear handler will fix the problems, and remove the non-elem set handler (unless you know a case where it is called)

demerphq:
> (unless you know a case where it is called)

its called from local.

```
gdb --args ./perl -Ilib -E '%{^HOOK} =(require__before => sub { say "hello"; () }); local %{^HOOK}; require Cwd'
```

same for %SIG. I just haven't worked out what to put inside of it, or Perl_magic_clearhookall. :-( In the output below the dump comes from me adding an hv_dump() to Perl_magic_clearhookall. But I havent worked out what i need to put inside of it. The hash seems to have no elements in it.  I am getting really confused about what i should do.

I'm struggling here TBH, I cant find any docs on this stuff to explain what is going on.

```
$ gdb --args ./perl -Ilib -E '%{^HOOK} =(require__before => sub { say "hello"; () }); local %{^HOOK}; require Cwd'
...
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./perl...
(gdb) b Perl_magic_sethookall
Breakpoint 1 at 0x25bcb1: file mg.c, line 1944.
(gdb) r
Starting program: /home/yorton/git_tree/perl/perl -Ilib -E %\{\^HOOK\}\ =\(require__before\ =\>\ sub\ \{\ say\ \"hello\"\;\ \(\)\ \}\)\;\ local\ %\{\^HOOK\}\;\ require\ Cwd
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
SV = PVHV(0x555555c7db48) at 0x555555c90320
  REFCNT = 4
  FLAGS = (OOK,SHAREKEYS)
  MAGIC = 0x555555c9d538
    MG_VIRTUAL = &PL_vtbl_hook
    MG_TYPE = PERL_MAGIC_hook(Z)
  AUX_FLAGS = 0
  ARRAY = 0x555555ca11c8
  KEYS = 0
  FILL = 0
  MAX = 7
  RITER = -1
  EITER = 0x0
  RAND = 0x318a3d68

Breakpoint 1, Perl_magic_sethookall (my_perl=0x555555c5a2a0, sv=0x18, mg=0x555555c5d760) at mg.c:1944
1944	{
(gdb) bt
#0  Perl_magic_sethookall (my_perl=0x555555c5a2a0, sv=0x18, mg=0x555555c5d760) at mg.c:1944
#1  0x00005555557a8b17 in Perl_mg_set (my_perl=0x555555c5a2a0, sv=0x555555c5d760) at mg.c:293
#2  0x00005555557a96a6 in Perl_mg_localize (my_perl=0x555555c5a2a0, sv=0x555555c90320, nsv=0x555555c5d760, 
    setmagic=true) at mg.c:507
#3  0x00005555558bcca8 in Perl_save_hash (my_perl=0x555555c5a2a0, gv=0x555555c90278) at scope.c:543
#4  0x00005555557fccfe in Perl_pp_rv2av (my_perl=0x555555c5a2a0) at pp_hot.c:2081
#5  0x0000555555790dcb in Perl_runops_debug (my_perl=0x555555c5a2a0) at dump.c:2861
#6  0x00005555555e9d9f in S_run_body (my_perl=0x555555c5a2a0, oldscope=1) at perl.c:2805
#7  0x00005555555e908d in perl_run (my_perl=0x555555c5a2a0) at perl.c:2720
#8  0x000055555559b30d in main (argc=4, argv=0x7fffffffd9a8, env=0x7fffffffd9d0) at perlmain.c:127
(gdb) 
```

demerphq:
@tonycoz I think that last patch sorts the localization issues. but I would be very much grateful for a second set of eyes.

While working on this I realized there is a timing issue with the after hooks, they fire a bit too late when executed via a require statement as opposed to a begin statement. I need to figure that out tomorrow.

t/op/hook/require.t
```diff
@@ -0,0 +1,215 @@
+#!perl
```
tonycoz, line 1: I hadn't noticed the extra directory level here - this test isn't run by `make test_harness`, I searched the first three CI runs that actually run tests and none ran this file.

I added a bare `die;` to this file and all tests passed.

mg.c
```diff
@@ -1891,6 +1893,98 @@ Perl_magic_setsigall(pTHX_ SV* sv, MAGIC* mg)
...
+
+    if (PL_hook__require__before) {
+        SvREFCNT_dec(PL_hook__require__before);
+        PL_hook__require__before = NULL;
+    }
+
+    if (PL_hook__require__after) {
+        SvREFCNT_dec(PL_hook__require__after);
+        PL_hook__require__after = NULL;
+    }
```
tonycoz: SvREFCNT_dec() already checks for NULL, you could skip the if() or use SvREFCNT_dec_NN().

pod/perlvar.pod
```diff
@@ -779,6 +779,67 @@ and use an C<END{}> or CORE::GLOBAL::die override instead.
...
+
+This hash contains coderefs which are called when various perl keywords
+which are hard or impossible to wrap are called. The keys of this hash
+are named after the keyword that is being hooked, followed by two
+underbars and then a phase term; either "before" or "after".
+
+It is an error to attempt to a value into this hash under a key which is
+not known to perl, and it is an error to store anything other than a
+CODEREF or undef into the hash. If you wish to use an object to implement
+a hook you can use currying to embed the object into the subroutine.
```
tonycoz: "attempt to XXX a value into" - missing word, presumably "store"

"It is an error" - with what consequences?

Perhaps instead:
```
Perl will throw an error if you attempt to modify a key which is not
known to perl, or to store anything other than a CODEREF or undef
into the hash.
```

tonycoz: Also, we seem to mostly spell out "code refererence" rather than "coderef" or "CODEREF" in prose in pod/

pod/perlvar.pod
```diff
@@ -779,6 +779,67 @@ and use an C<END{}> or CORE::GLOBAL::die override instead.
...
+C<require__before> hook will be silently ignored.
+
+C<require__before> hooks are called in FIFO order, and if they return
+a coderef those coderefs are called in FILO order.  In other words if A requires
+B requires C, then C<require__before> will be called for A, B and C in turn,
+and then the post-action coderef will execute for C, B and then finally A.
+
+Well behaved code should ensure that when setting up a
+C<require__before> hook that any prior installed hook will be called,
+and that their return, if a code-reference, will be called as well.  See
```
tonycoz: "return value"

pod/perlvar.pod
```diff
@@ -779,6 +779,67 @@ and use an C<END{}> or CORE::GLOBAL::die override instead.
...
+=item require__after
+
+The routine indicated by C<${^HOOK}{require__after}> is called by C<require>
+B<after> the require completes.  It is called with a single argument, the filename
+for the item being required (package names are converted to paths).
+It is executed when the C<require> completes, either via exception or via
+completion of the require statement, and you can check %INC to determine
+if the require was successful.
+
+C<require__after> hook are called in FILO order.  In other words if A requires
```
tonycoz: "C<require__after> hook are called in FILO order."  the "hook are" seems ungrammatical to me.

Perhaps:
```
The C<require__after> hook is called for each file in FILO order.
```
or
```
The C<require_after> hooks are called in FILO order.
```
(but there's only one such hook)

tonycoz:
Only the test filename causing the test not to run is critical.

mg.c
demerphq: OK. For what its worth I kinda wonder if we shouldn't have a SvREFCNT_dec() that updates the pointer to NULL.

t/op/hook/require.t
demerphq: Fixed in https://github.com/Perl/perl5/pull/20937

mg.c
demerphq: done.

pod/perlvar.pod
demerphq: Fixed, and i used your suggested wording with a few minor changes.

pod/perlvar.pod
demerphq: Fixed.

pod/perlvar.pod
demerphq: Fixed.

regen/mg_vtable.pl
```diff
@@ -212,6 +212,9 @@ BEGIN
                    vtable => 'debugvar' },
      lvref => { char => '\\', vtable => 'lvref',
                   desc => "Lvalue reference constructor" },
+     destruct => { char => "X", vtable => 'destruct',
+                        desc => "destruct callback" },
```
Leont: This should probably also set «value_magic => 1», otherwise localizing such a value will cause the free handler to be run twice

scope.c
```diff
@@ -1846,6 +1846,63 @@ Perl_cx_dump(pTHX_ PERL_CONTEXT *cx)
...
+        SV   *sv;
+        AV   *av;
+        char *pv;
+    } args_any;
+    IV nargs = 0;
+    if (PL_phase == PERL_PHASE_DESTRUCT) {
+        Perl_warn(aTHX_ "Can't call destructor for 0x%p in global destruction\n", sv);
+        return 1;
+    }
+    /* PUSHSTACKi(PERLSI_MAGIC); */
```
Leont: The `PUSHSTACK` (and a `POPSTACK` at the end) is well-adviced, otherwise any code that doesn't expect the stack to move from under their feet can get confused by an unexpected free handler.

scope.c
```diff
@@ -1846,6 +1846,63 @@ Perl_cx_dump(pTHX_ PERL_CONTEXT *cx)
 #endif	/* DEBUGGING */
 }
 
+void
+Perl_create_sv_destructor(pTHX_ SV *subref, SV *args) {
+    PERL_ARGS_ASSERT_CREATE_SV_DESTRUCTOR;
+    SV *variable = newSV_type_mortal(SVt_IV);
+    MAGIC* magic = sv_magicext(variable, subref, PERL_MAGIC_destruct,
+                               &PL_vtbl_destruct, (char *)args, args ? HEf_SVKEY : 0);
+    magic->mg_flags |= MGf_LOCAL;
```
Leont: The MGf_LOCAL line isn't needed if no localization handler is set, and no localization handler is needed if «value_magic => 1» is set in `mg_vtable.pl` (and then `MAGIC* magic`  doesn't have to be saved either).

regen/mg_vtable.pl
demerphq: Done.

scope.c
demerphq: Done.

scope.c
demerphq: done.

scope.c
```diff
@@ -1846,6 +1846,145 @@ Perl_cx_dump(pTHX_ PERL_CONTEXT *cx)
...
+    SV *variable = newSV_type_mortal(SVt_IV);
+    (void)sv_magicext(variable, coderef, PERL_MAGIC_destruct,
+                      &PL_vtbl_destruct, (char *)args, args ? HEf_SVKEY : 0);
+}
+
+
+void
+Perl_mortal_destructor_x(pTHX_ DESTRUCTORFUNC_t f, void *p) {
+    PERL_ARGS_ASSERT_MORTAL_DESTRUCTOR_X;
+    mortal_destructor_sv((SV*)f,(SV *)p);
```
Leont: This will treat `f` and `p` as refcounted, which is probably a bad idea.

Just using a different vtable is probably the easiest solution, this should be the simpler case anyway.

demerphq: I reworked the code a little bit to use SVFUNC_t which requires a SV * argument instead of DESTRUCTORFUNC_t which supports a void * argument.  Also the code you quoted is/was just broken, the function needed to be wrapped in a SVt_IV and the f argument needed to be converted into an IV with PTR2IV(). But once I did that it seems work as expected and passes the tests I added to XS-APITest. 

I am not really sure what we win by creating a new vtable to handle this and it would cost us one of the magic chars. It seems to me with the adjustments I just made it is fine as it is. 

Can you review once more please?

scope.h
```diff
@@ -195,6 +195,12 @@ scope has the given name. C<name> must be a literal string.
 #define SAVEDESTRUCTOR_X(f,p) \
           save_destructor_x((DESTRUCTORFUNC_t)(f), (void*)(p))
 
+#define SAVEMORTALSVFUNC_X(f,sv) \
+          mortal_svfunc_x((SVFUNC_t)(f), sv)
+
+#define SAVEMORTALDESTRUCTOR_SV(coderef,args) \
+          mortal_destructor_sv(coderef,args)
+
```
leonerd: A small point on the naming: I don't think these should be called "SAVE...", because they don't push anything to the save stack. Every other "SAVEfoo" macro is a wrapper for "push some values to the save stack".

I like the idea of them, I just think the name is misleading.

leonerd (changes requested):

scope.h
demerphq: The match SAVEMORTALIZESV(), which where the naming comes from.

leonerd: Oh I see. Boo. In that case, all the names are annoying and we should fix them all up at some point in the future, but for now yours isn't making it (much) worse, so might as well stick to that precedent then.

leonerd (approved):

haarg:
I generally like this. I'm not entirely certain about the interface - having two types of after hooks feels like it might indicate there's a better option. But that isn't a rejection.

But is this something that should be experimental?

demerphq:
> But is this something that should be experimental?

I don't see how that would be advantageous. This doesn't get in the way and its not a core part of the language nor does it affect code that already exists. (Like most use of experimental features.)

>  having two types of after hooks feels like it might indicate there's a better option

There aren't two types of after hook. The is a post-before hook and an after hook.  They aren't the same thing.  I explained why we have both in previous post, sharing state with /just/ an after hook is error prone and unreliable. On the other hand, @ap requested we provide a separate after hook for people who don't want to use a before hook and just want to do something after each require. I just gave him what he requested.

iabyn:
On Fri, Mar 17, 2023 at 06:37:57AM -0700, Yves Orton wrote:
> @demerphq commented on this pull request.
> 
> 
> 
> > +#define SAVEMORTALSVFUNC_X(f,sv) \
> +          mortal_svfunc_x((SVFUNC_t)(f), sv)
> +
> +#define SAVEMORTALDESTRUCTOR_SV(coderef,args) \
> +          mortal_destructor_sv(coderef,args)
> +
> 
> The match SAVEMORTALIZESV(), which where the naming comes from.

Yeah, but SAVEMORTALIZESV() still uses the save stack. It's action is to
mortalise the SV in leave_scope() on scope exit

So I don't think SAVE should be in the names.

-- 
The Enterprise successfully ferries an alien VIP from one place to another
without serious incident.
    -- Things That Never Happen in "Star Trek" #7

scope.h
demerphq: Well, @iabyn pointed out the flaw in my reasoning, so I have removed the SAVE from the macro wrappers as you requested originally. Sorry for not fully understanding SAVEMORTALIZESV() and confusing matters.

haarg:
> I don't see how that would be advantageous. This doesn't get in the way and its not a core part of the language nor does it affect code that already exists. (Like most use of experimental features.)

It isn't syntax, but it's still a part of the language. There are plenty of other experimental things that have had warnings but no feature flag, because they didn't interfere with any existing syntax. `postderef`, `const_attr`, and `builtin` for example. `smartmatch` as well (not `switch`).

> > having two types of after hooks feels like it might indicate there's a better option
> 
> There aren't two types of after hook. The is a post-before hook and an after hook. They aren't the same thing.

Those are both "after" hooks, and I never said they were the same thing. I understand the need for carrying state between the hooks. Having both types feels a bit awkward, but I'm not sure there's a better option.

Leont:
> Yeah, but SAVEMORTALIZESV() still uses the save stack. It's action is to
mortalise the SV in leave_scope() on scope exit
>
> So I don't think SAVE should be in the names.

I think that is a strong argument for renaming them before the next major release.

demerphq:
> I think that is a strong argument for renaming them before the next major release.

Done already. I noted in reply to @leonerd's original request, which davem was replying to, but because he replied via email it was not logged in the right place.

(This is just for the record, you and I discussed this already in IRC.)

PR #20639, opened 2022-12-22
Revert "Inline strlcat(), strlcpy()"

demerphq:
This reverts commit 5be23dd97e360d19c8abb4c0c534f5d5f9d3691a.

The original patch seems to break DBM on clang.


demerphq:
Clang has been broken since the patch this reverts, and is fixed here. I am going to apply this unilaterally as I was responsible for the merge earlier today. @khwilliamson fyi.

PR #20642, opened 2022-12-23
Revert "harness: handle directories serial/parallel mixture"

demerphq:
Reverts Perl/perl5#20630

It turns out that the numbers in tests really dont help with deciding if they need to be mutexted against other tests.

For now we will revert this.

### DIFF
--- a/t/harness
+++ b/t/harness
@@ -8,14 +8,8 @@ BEGIN {
     @INC = '../lib';              # pick up only this build's lib
 }
 
-my %must_be_executed_serially =
-    map { $_ => 1 } (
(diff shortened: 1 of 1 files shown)


demerphq:
FWIW, I am not sure we should revert https://github.com/Perl/perl5/pull/20642/commits/4d1aef8a5580a43a0ebd6eeeabff27094639cb17 but i cant manually fix this PR while my mutex tests are running.

demerphq:
I discovered the reason my tests were taking so long was a single file in the JSON::PP codebase, so i forced it to fail and the rest finished quickly afterwards, so i have repushed the revert to leave the config change alone.

PR #20643, opened 2022-12-23
Re-enable full blown parallel tests without using heuristics to detect special cases.

demerphq:
Unrevert @khwilliamson's parallelism patch, and then modify it to not use heuristics to decide what should be run in series, and at the same time introduce a new class of tests which must be run as singletons. This seems to make `dist/threads-shared/t/waithires.t` run successfully on Win32, when it can hang and die in the middle when it is run alongside other tests.

This also changes the model of listing special cases, instead of listing directories we list test files. Except for cases like waithires.t I think the list should eventually be empty, and the tests patched. The IO-Zlib exceptions hopefully can be removed in the near future.  See: https://github.com/tomhughes/IO-Zlib/pull/4

### DIFF
--- a/t/harness
+++ b/t/harness
@@ -8,14 +8,39 @@ BEGIN {
     @INC = '../lib';              # pick up only this build's lib
 }
 
-my %must_be_executed_serially =
-    map { $_ => 1 } (
-                     '../cpan/IO-Zlib/t',
-                     '../ext/File-Find/t',
-                     '../lib/DBM_Filter/t',  # Multiple test files in this
-                                             # directory use the same
-                                             # hard-coded temp file name
-                    );
+##############################################################################
+# Test files which cannot be executed at the same time.
+#
+# List all files which might fail when executed at the same time as another
+# test file from the same test directory. Being listed here does not mean
+# the test will be run by itself, it just means it won't be run at the same
+# time as any other file in the same test directory, it might be run at the
+# same time as a file from a different test directory.
+#
+# Ideally this is always empty.
+#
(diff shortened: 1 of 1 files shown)


demerphq:
I have also filed this: https://github.com/makamaka/JSON-PP/pull/80

demerphq:
@khwilliamson I think this is ready to go now. I removed the file name heuristics and numbered files stuff and changed the exceptions to be test file based instead of directory based, and added a new special class for files that should run as true singletons, which seems to make win32 tests happy with waithires.t This unreverts your directory split patch, and exploits it for the "per directory" mutex requirements.  Also the patches to external modules have been merged, it seems the only true parallism issue left is IO-Zlib, and i have a PR pending for that.

I have had this running in a loop for a while now, and I havent seen any fails in CI since I split out waithires.t

demerphq:
There is something curious that this PR reveals. When I let waithires.t run in the middle, or even separated from other win32 tests, I got CI failures. And when I separated it out to run at the very end, those CI failures seem to have stopped. But when I check the CI config for Windows, we dont run the windows CI tests in parallel at all. So why would running the waithires.t as the very last test make a difference. I don't get it. Is that just plain dumb luck?

demerphq:
I will merge this soon unless there are objections.

demerphq:
With https://github.com/Perl/perl5/pull/20655 we should have no tests at all that have ordering dependencies or cannot be run in parallel, with the weird exception of waithires.t on win32. As such I am going to merge this and see how it goes.

PR #20647, opened 2022-12-24
First round of `feature 'class'`

leonerd:
## Overall Notes To Reviewers

This is an **epic**-sized PR. I am not expecting that any one person can review everything all at once. It is perfectly fine to decide to just review some part of it - maybe just the docs, or the tests, or some small part of the implementation. Whatever you feel most confident in. When you leave a review comment please explain which part(s) of the PR you read and are commenting on, so it's clear.

*Reminder*: You do not have to be an internals expert, or even read any C code at all, to be able to review and comment on the tests, documentation, and overall shape of the idea.

Additionally, anyone who does wish to read and comment on the actual implementation is advised to start by reading the new documentation file `pod/perlclassguts.pod` which explains many details of how the internals work. Please start there to try to understand how the pieces all fit together. If there's something still unclear after reading it then that in itself is a valuable review comment. :) So please let me know.

## PR Specifics

I have tried to squash down my original list of about 60 commits into larger pieces here, but I've still kept various stages split apart by empty-commit markers, as well as leaving several of them separate generally. I'm open to suggestion on whether to further squash them down before merging.

In particular the very first commit, titled "Define the concept of a suspended compcv", could stand alone as its own PR to be merged directly first as a prerequisite, although it would have no real utility in core *other* than to support the rest of this work here so it seems less useful to bother doing so in its own PR.


pad.c
```diff
@@ -2844,6 +2844,91 @@ Perl_padname_dup(pTHX_ PADNAME *src, CLONE_PARAMS *param)
 
 #endif /* USE_ITHREADS */
 
+/*
+=for apidoc_section $lexer
+=for apidoc suspend_compcv
+
+Implements part of the concept of a "suspended complication CV"; which can be
```
jkeenan: Will the concept of a "suspended complication CV" be explained in user-facing documentation in a subsequent commit (*i.e.,* `pod/perlsomething.pod`)?

Also, what is the point of the semicolon in the line above.  Shouldn't it either be a comma or rewritten as `"suspended complication CV." This can be`

jkeenan:
Some perl-level warnings are being emitted during `make test_prep`:
```
$ git describe
v5.37.7-49-g5d3f7324c3
$ zcat 5d3f7324c3.freebsd.threaded.clang.feature-class.maketp.output.txt.gz | tail -n 6
./miniperl -Ilib autodoc.pl
Function 'resume_compcv_and_save', documented in pad.c, not listed in embed.fnc at autodoc.pl line 1819.
Function 'resume_compcv_final', documented in pad.c, not listed in embed.fnc at autodoc.pl line 1819.
./miniperl -Ilib pod/perlmodlib.PL -q
./perl -Ilib -I. -f pod/buildtoc -q
cd t && (rm -f perl; /bin/ln -s ../perl perl)
```

pod/perlclass.pod
```diff
@@ -0,0 +1,310 @@
...
+
+An extension of the metaprogramming API (currently proposed by RFC0022) which
+adds knowledge of classes, methods, fields, ADJUST blocks, and other such
+class-related details.
+
+=item * Extension Customisation
+
+Ways in which out-of-core modules can interact with the class system,
+including an ability for them to provide new class or field attributes.
+
```
Ovid, line 301: Isn't this part of the previous item, "Metaprogramming," and thus redundant, or am I missing something?

pod/perldiag.pod
```diff
@@ -1707,6 +1759,33 @@ is better written as simply itself, perhaps preceded by a backslash
 for non-word characters.  Doing it the way you did is not portable
 between ASCII and EBCDIC platforms.
 
+=item Class already has a superclass, cannot add another
+
+(F) You attempted to specify a second superclass for a C<class> by using
+the C<:isa> attribute, when one is already specified.  Classes may have
+at most one superclass.
```
Ovid: Perhaps a bit of clarification of the difference between old and new style classes?

```
(F) You attempted to specify a second superclass for a C<class> by using
the C<:isa> attribute, when one is already specified.  Unlike classes
whose instances are created with C<bless>, classes created via the
C<class> keyword cannot have more than one superclass.
```

pod/perlclass.pod
```diff
@@ -0,0 +1,310 @@
...
+Classes may inherit from B<one> superclass, by using the C<:isa> class
+attribute.
+
+    class Example::Base { ... }
+
+    class Example::Subclass :isa(Example::Base) { ... }
+
+Inherited methods are visible and may be invoked. Fields are always lexical
+and therefore not visible by inheritence.
```
lindleyw: spelling: 'inheritance'

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
+=head1 NAME
+
+perlclassguts - Internals of how C<feature 'class'> and class syntax works
+
+=head1 DESCRIPTION
+
+This document intends to provide in-depth information about the way in which
```
cjac: Can we switch the words around a bit here?  This is a bit easier to wrap my head around:

This document intends to provide in-depth information about the syntax and overall behaviour of the perl interpreter internals with regard to the C<feature 'class'>.

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+the perl interpreter internals around the C<feature 'class'> syntax and overall
+behaviour are provided. It is not intended as an end-user guide on how to use
+the feature; for that see L<perlclass>.
```
cjac: for that, see L<perlclass>.

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+
+The reader is assumed to be generally familiar with the perl interpreter
+internals overall. For a more general overview of these details, see also
+L<perlguts>.
+
+=head1 DATA STORAGE
+
+=head2 Classes
+
+A class is still fundamentally a package, and exists in the symbol table as an
```
cjac: is the word "still" necessary?

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+HV with an aux structure in exactly the same way. It is distinguished from a
```
cjac: in exactly the same way as what?

cjac (changes requested):
This is just the start of my review.  I would like to check this out and build the interpreter to test the documentation.  Do we have a docker with the image and deps installed yet?

cjac:
I'm building from source on my Debian WSL container with 

```
perl -p -e 'print "deb-src $1$/" if /^deb (.*)$/' /etc/apt/sources.list  | sudo dd of=/etc/apt/sources.list
sudo apt-get update
sudo apt-get build-dep -y perl
cd ~/src/github
mkdir perl5
cd perl5
git clone git@github.com:leonerd/perl5.git
cd perl5
git checkout -t origin/feature-class
eval "$(perl -I$HOME/perl5/lib/perl5 -Mlocal::lib | tee -a ~/.bashrc)"
$ ./Configure -des -Dprefix=$HOME/perl5
$ make
$ make install
$ which perl5.37.8
/home/cjac/perl5/bin/perl5.37.8
$ /home/cjac/perl5/bin/perl5.37.8 --version

This is perl 5, version 37, subversion 8 (v5.37.8 (v5.34.0-RC1-3603-g5d3f7324c3)) built for x86_64-linux

Copyright 1987-2022, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at https://www.perl.org/, the Perl Home Page.
```

Ovid:
For what it's worth, [I just ported HTML::TokeParser::Simple to use `feature 'class'`](https://github.com/Ovid/html-tokeparser-corinna). It was dead simple to do and everything behaved exactly as expected. The code is also a hell of a lot cleaner, though there are technical differences which make it an unfair comparison.

Ovid:
Getting some odd behavior with `Data::Dumper`.

```perl
#!/usr/bin/env perl

use 5.37.8;
use experimental 'class';
use Data::Dumper;

class Foo {}
my $foo = Foo->new;
say Dumper($foo);
```
Running that prints:

```
cannot handle ref type 16 at /Users/ovid/perl5/perlbrew/perls/feature-class/lib/5.37.8/darwin-2level/Data/Dumper.pm line 212.
$VAR1 = bless( , 'Foo' );
```

I assume this is expected, but it might be nice to have something a touch cleaner for this.

Ovid:
**Update**: [there is now a branch of `Perl::Tidy` which handles the new syntax](https://rt.cpan.org/Ticket/Display.html?id=145706)

Perl::Tidy is also struggling. Since it's not core, it probably is less important, but it's widely used. Leaving this here for others.

```perl
#!perl

use 5.37.8;
use experimental 'class';

class Point {
    field $x :param;
    field $y :param;

    method to_string () {
        return "($x, $y)";
    }
}
```

Running `Perl::Tidy` on that produces:

```
tidy.pl: Begin Error Output Stream
tidy.pl:
tidy.pl: There is no previous '?' to match a ':' on line 7
tidy.pl: 7: field $x :param;
tidy.pl:             ^
tidy.pl:
tidy.pl: There is no previous '?' to match a ':' on line 8
tidy.pl: 8: field $y :param;
tidy.pl:             ^
tidy.pl:14: To save a full .LOG file rerun with -g
```

I tried reading through `Perl::Tidy::Tokenizer` to see what's going on. On line 1860, we can change this:

```perl
    my %is_my_our_state;
    @q = qw(my our state);
    @is_my_our_state{@q} = (1) x scalar(@q);
```

To this:

```perl
    my %is_my_our_state;
    @q = qw(my our state field);
    @is_my_our_state{@q} = (1) x scalar(@q);
```

And that fixes the issue. I'm sure there will be others in Perl::Tidy. You can also add `class` to that list to let it properly handle attributes (`:isa(...)`) on class declarations, but that seems a nasty hack.

`Perl::Tidy` also messes up the indentation if you use a postfix block on a class. This can be fixed in `Perl::Tidy::Tokenizer`'s `code_block_type` function starting at line 5993 by including `class` with `sub` and `package`.

```perl
    # or a sub or package BLOCK
    elsif ( ( $last_nonblank_type eq 'i' || $last_nonblank_type eq 't' )
        && $last_nonblank_token =~ /^(sub|package|class)\b/ )
    {
        return $last_nonblank_token;
    }

    # or a sub alias
    elsif (( $last_nonblank_type eq 'i' || $last_nonblank_type eq 't' )
        && ( $is_sub{$last_nonblank_token} ) )
    {
        return 'sub';
    }

    elsif ( $statement_type =~ /^(sub|package|class)\b/ ) {
        return $statement_type;
    }
```

(I really should make these diffs and stop hacking my local perltidy code)

I've filed a report for Perl::Tidy at https://rt.cpan.org/Ticket/Display.html?id=145706&results=44284c8317c4fa0392f65a81614b024c

Ovid:
`Perl::Tidy` also messes up the indentation if you use a postfix block on a class. This can be fixed in `Perl::Tidy::Tokenizer`'s `code_block_type` function starting at line 5993 by including `class` with `sub` and `package`.

```perl
    # or a sub or package BLOCK
    elsif ( ( $last_nonblank_type eq 'i' || $last_nonblank_type eq 't' )
        && $last_nonblank_token =~ /^(sub|package|class)\b/ )
    {
        return $last_nonblank_token;
    }

    # or a sub alias
    elsif (( $last_nonblank_type eq 'i' || $last_nonblank_type eq 't' )
        && ( $is_sub{$last_nonblank_token} ) )
    {
        return 'sub';
    }
```

(I really should make these diffs and stop hacking my local perltidy code)

pod/perlclassguts.pod
cjac: `as a non-class package` ?

cjac:
> Getting some odd behavior with `Data::Dumper`.
> 
> ```perl
> #!/usr/bin/env perl
> 
> use 5.37.8;
> use experimental 'class';
> use Data::Dumper;
> 
> class Foo {}
> my $foo = Foo->new;
> say Dumper($foo);
> ```
> 
> Running that prints:
> 
> ```
> cannot handle ref type 16 at /Users/ovid/perl5/perlbrew/perls/feature-class/lib/5.37.8/darwin-2level/Data/Dumper.pm line 212.
> $VAR1 = bless( , 'Foo' );
> ```
> 
> I assume this is expected, but it might be nice to have something a touch cleaner for this.

Should this be added to dist/Data-Dumper/t/ ?

```
cjac@build0:/usr/src/git/github/perl5/perl5/dist/Data-Dumper$ cat t/class.t
#!perl

use 5.37.8;
use experimental 'class';
use Test::Exception;
use Test::More;

use 'Data::Dumper';

BEGIN { plan tests => 1; }

class Foo {}
my $foo = Foo->new;

lives_ok { say Dumper( $foo ) } 'expecting to live';
cjac@build0:/usr/src/git/github/perl5/perl5/dist/Data-Dumper$ prove5.37.8 -b -v t/class.t
t/class.t .. 
1..1
$VAR1 = bless( , 'Foo' );
ok 1 - expecting to live
cannot handle ref type 16 at /usr/src/git/github/perl5/perl5/dist/Data-Dumper/blib/lib/Data/Dumper.pm line 212.
ok
All tests successful.
Files=1, Tests=1,  0 wallclock secs ( 0.07 usr  0.02 sys +  0.13 cusr  0.03 csys =  0.25 CPU)
Result: PASS

cjac@build0:/usr/src/git/github/perl5/perl5/dist/Data-Dumper$ cat -n /usr/src/git/github/perl5/perl5/dist/Data-Dumper/blib/lib/Data/Dumper.pm | head -215 | tail -9
   207
   208  sub Dump {
   209    return &Dumpxs
   210      unless $Data::Dumper::Useperl || (ref($_[0]) && $_[0]->{useperl})
   211              # Use pure perl version on earlier releases on EBCDIC platforms
   212          || (! $IS_ASCII && $] lt 5.021_010);
   213    return &Dumpperl;
   214  }
   215
```

Ovid:
```perl
#!perl

use 5.37.8;
use experimental 'class';
use Test::Exception;
use Test::More;

BEGIN { use_ok('Data::Dumper') };

BEGIN { plan tests => 1; }

class Foo {}
my $foo = Foo->new;

lives_ok { say Dumper( $foo ) } 'expecting to live';
```

A few comments on testing. First, you have `use_ok` and `lives_ok`, so those are two tests, even though you only specify one in the plan.

Second, unless there's a standard for it in the Perl core, I wouldn't use `use_ok`. It's fragile and you have to remember how to use it correctly. Instead, just `use Data::Dumper;` and if that dies, the test fails. Much simpler and easier.

Third, `lives_ok` kinda helps to verify that the code lives, but we also want to know that there are no warnings and that the output is correct. For my tests, I use `Capture::Tiny`. For core, I have no idea what best practice is.

class.c
```diff
@@ -0,0 +1,1032 @@
...
+ *
+ *    You may distribute under the terms of either the GNU General Public
+ *    License or the Artistic License, as specified in the README file.
+ *
+ */
+
+/* This file contains the code that implements perl's new `use feature 'class'`
+ * object model
+ */
+
```
demerphq, line 13: This should include:
```
#define PERL_IN_CLASS_C
```
and likely at least some definitions in embed.fnc should be restricted with it. See all our other .c files for similar boilerplate.

demerphq:
This is missing the build changes for win32 to compile class.c, you have to do that manually in the makefiles in win32. See the changes i did recently to add regcomp_debug.c:

```
$ git grep regcomp_debug win32
win32/GNUmakefile:              ..\regcomp_debug.c      \
win32/GNUmakefile:..\regcomp_debug$(o) : ..\regcomp.h ..\regcomp_internal.h ..\regnodes.h ..\regcharclass.h
win32/Makefile:         ..\regcomp_debug.c      \
win32/Makefile:..\regcomp_debug$(o) : ..\regcomp.h ..\regcomp_internal.h ..\regnodes.h ..\regcharclass.h
```

demerphq:
> Third, lives_ok kinda helps to verify that the code lives, but we also want there's no warning and that the output is correct. For my tests, I use Capture::Tiny. For core, I have no idea what best practice is

I tend to use fresh_perl_is() and fresh_perl_like().

pod/perlclassguts.pod
wollmers: `This document provides ... `

Or even skip longish introductions at all.

pod/perlclass.pod
```diff
@@ -0,0 +1,310 @@
...
+            say $x;
+        }
+    }
+
+    My::Example->new->print_message;
+
+=head1 DESCRIPTION
+
+This document describes the syntax of the Perl's C<class> feature, which
+provider native keywords supporting object oriented programming paradigm.
```
cjac: provides native keywords

cjac:
There are no examples of calling object methods.  From within a method of the class, can we just call method() or do we need to call $self->method()?  Are they equivalent?

Ovid:
@leonerd I think we need an implicit `require` for `:isa(...)`. I just tried this:

```perl
#!/usr/bin/env perl

use lib 'lib';
use v5.37.8;

use HTML::TokeParser::Corinna::Token::Text;

my $token = HTML::TokeParser::Corinna::Token::Text->new( token => [ 'T', 'This is my text' ] ),
```

That fails with:

> Class :isa attribute requires a class but "HTML::TokeParser::Corinna::Token" is not one at lib/HTML/TokeParser/Corinna/Token/Text.pm line 3.

The class I'm trying to instantiate looks like this:

```perl
use experimental 'class';

class HTML::TokeParser::Corinna::Token::Text : isa(HTML::TokeParser::Corinna::Token) {
    no warnings 'experimental::builtin';
    use builtin 'true';
    method is_text { true }
}

1;
```

My expectation is that, this should be like using `parent` or `base`: if the class is not loaded, `require $class` will be called.

demerphq:
On Fri, 30 Dec 2022 at 10:44, Ovid ***@***.***> wrote:

>
> class HTML::TokeParser::Corinna::Token::Text : isa(HTML::TokeParser::Corinna::Token) {
>
> Maybe I have been overinfluenced by Rust, but I sure wish that lone ':'
token was as '->' or '=>' instead. It would read much nicer. There are a
*lot* of colons on that line.

cheers,
Yves

Ovid:
> On Fri, 30 Dec 2022 at 10:44, Ovid ***@***.***> wrote: 
> ```perl
> class HTML::TokeParser::Corinna::Token::Text : isa(HTML::TokeParser::Corinna::Token) {
> ```
> Maybe I have been overinfluenced by Rust, but I sure wish that lone ':'
> token was as '->' or '=>' instead. It would read much nicer. There are a *lot* of colons on that line. cheers, Yves

Agreed. It looks kinda ugly, but [the KIM grammar](https://ovid.github.io/articles/language-design-consistency.html) we've adopted for the keywords is:

```
KEYWORD IDENTIFIER MODIFIERS? DEFINITION?
```

Modifiers are always attributes. So Corinna only offers four new keywords (at this time) and all provide a consistent syntax:

```perl
# KEYWORD IDENTIFIER         MODIFIERS?        DEFINITION?
class     Foo                :isa(Bar)         { ... }
field     $foo               :param;
method    foo                :private          { ... }
role      Role::Serializable :does(Role::JSON) { ... }
```

Once Damian suggested this syntax, it was a unanimous decision by the Corinna team to adopt it (I was pleased, but surprised, by the unanimity). Plus, it mostly leverages what we already have in Perl.

demerphq:
On Fri, 30 Dec 2022 at 12:32, Ovid ***@***.***> wrote:

> On Fri, 30 Dec 2022 at 10:44, Ovid *@*.***> wrote:
>
> class HTML::TokeParser::Corinna::Token::Text : isa(HTML::TokeParser::Corinna::Token) {
>
> Maybe I have been overinfluenced by Rust, but I sure wish that lone ':'
> token was as '->' or '=>' instead. It would read much nicer. There are a
> *lot* of colons on that line. cheers, Yves
>
> Agreed. It looks kinda ugly, but the KIM grammar
> <https://ovid.github.io/articles/language-design-consistency.html> we've
> adopted for the keywords is:
>
> KEYWORD IDENTIFIER MODIFIERS? DEFINITION?
>
> Modifiers are always attributes. So Corinna only offers four new keywords
> (at this time) and all provide a consistent syntax:
>
> # KEYWORD IDENTIFIER         MODIFIERS?        DEFINITION?
> class     Foo                :isa(Bar)         { ... }
> field     $foo               :param;
> method    foo                :private          { ... }
> role      Role::Serializable :does(Role::JSON) { ... }
>
>
Things often look fine when you have a single or double word class, but
dont when you have a deeply nested one. Our errors for require statements
have this problem, we duplicate the class name twice in the message, which
looks fine when its a single or maybe double word example, but becomes too
much When::It::Is::A::Really::Deeply::Nested::Class::Name. Oh well.


> Once Damian suggested this syntax, it was a unanimous decision by the
> Corinna team to adopt it (I was pleased, but surprised, by the unanimity).
> Plus, it mostly leverages what we already have in Perl.
>
I can see the line of thought here. But playing devils advocate for a
moment, it doesn't seem like it would be a parse conflict to allow a =>
instead of colon to identify the modifiers. But I can live with "that boat
has sailed already". It reads very dense however. :-(

No reply necessary, just sharing my thoughts, I can live with this.

cheers,
Yves

-- 

Ovid:
@leonerd I don't know if this is intended, but I thought we were also injecting a `$class` variable into methods. This is failing:

```perl
    method normalize_tag {
        my $new_tag = lc $tag;
        return $class->new( token => [ 'E', $new_tag, sprintf "</%s>" => $new_tag ] );
    }
```

Changing `$class` to `(ref $self)` works, but it's ugly.

[Section 7.2 of the spec reads](https://github.com/Ovid/Cor/blob/master/rfc/methods.md#72-instance-methods):

> Instance methods are defined via `method $identifier (@args) { ... }`. They automatically have immutable `$class` and `$self` variables injected into them. `$class` contains the name of the class from which this method was called. `$self` is an instance of the current class.

Ovid:
**Update**: Overload works fine. It's the `explain` that was killing this, damn it :)

Also, is it known that we cannot yet handle overloading?

```perl
use experimental 'class';
class HTML::TokeParser::Corinna::Exception {
    use overload '""' => 'to_string', fallback => 1;
    use Devel::StackTrace;

    field $message : param;
    field $stack_trace = Devel::StackTrace->new->as_string;

    method message ()     {$message}
    method stack_trace () {$stack_trace}

    method to_string () {
        return join "\n" => $message, $stack_trace;
    }
}

1;
```

Later, I have this:

```perl
eval { die HTML::TokeParser::Corinna::Exception->new( message => 'foo' ) };
explain $@;
```

And it gives this warning:

```
cannot handle ref type 16 at /Users/ovid/perl5/perlbrew/perls/feature-class/lib/5.37.8/darwin-2level/Data/Dumper.pm line 212.
```

demerphq:
On Fri, 30 Dec 2022 at 16:45, Ovid ***@***.***> wrote:

> Also, is it known that we cannot yet handle overloading?
>
> use experimental 'class';
> class HTML::TokeParser::Corinna::Exception {
>     use overload '""' => 'to_string', fallback => 1;
>     use Devel::StackTrace;
>
>     field $message : param;
>     field $stack_trace = Devel::StackTrace->new->as_string;
>
>     method message ()     {$message}
>     method stack_trace () {$stack_trace}
>
>     method to_string () {
>         return join "\n" => $message, $stack_trace;
>     }
> }
>
> 1;
>
> Later, I have this:
>
> eval { die HTML::TokeParser::Corinna::Exception->new( message => 'foo' ) };
> explain $@;
>
> And it gives this warning:
>
> cannot handle ref type 16 at /Users/ovid/perl5/perlbrew/perls/feature-class/lib/5.37.8/darwin-2level/Data/Dumper.pm line 212.
>
>  Please change the explain command to use Devel::Peek::Dump and let us
know what it says.

Yves
-- 

Ovid:
@demerphq  Doing this:

```perl
use Devel::Peek;
eval { die HTML::TokeParser::Corinna::Exception->new( message => 'foo' ) };
Dump($@);
```

Resulted in this:

```
SV = PV(0x13200b6a0) at 0x13200adc8
  REFCNT = 1
  FLAGS = (ROK)
  RV = 0x1327dd228
  SV = PVOBJ(0x13303d000) at 0x1327dd228
    REFCNT = 1
    FLAGS = (OBJECT)
    STASH = 0x1327dd1f8	"HTML::TokeParser::Corinna::Exception"
    MAXFIELD = 1
    FIELDS = 0x600003216430
    Field No. 0 ($message)
    SV = PV(0x132852e20) at 0x133046030
      REFCNT = 1
      FLAGS = (POK,pPOK)
      PV = 0x600003216280 "foo"\0
      CUR = 3
      LEN = 16
    Field No. 1 ($stack_trace)
    SV = PV(0x132852700) at 0x1326b1090
      REFCNT = 1
      FLAGS = (POK,pPOK)
      PV = 0x131ebc390 "Trace begun at lib/HTML/TokeParser/Corinna/Exception.pm line 9\nHTML::TokeParser::Corinna::Exception::__ANON__('HTML::TokeParser::Corinna::Exception=OBJECT(0x1327dd228)', '** argument not available anymore **') called at t/exceptions.t line 16\neval {...} at t/exceptions.t line 16\n"\0
      CUR = 280
      LEN = 282
  PV = 0x1327dd228 ""
  CUR = 0
  LEN = 0```

**Comment by demerphq** [comment]
On Fri, 30 Dec 2022 at 17:03, Ovid ***@***.***> wrote:

> @demerphq <https://github.com/demerphq> Doing this:
>
> use Devel::Peek;eval { die HTML::TokeParser::Corinna::Exception->new( message => 'foo' ) };
> Dump($@);
>
> Resulted in this:
>
> SV = PV(0x13200b6a0) at 0x13200adc8
>   REFCNT = 1
>   FLAGS = (ROK)
>   RV = 0x1327dd228
>   SV = PVOBJ(0x13303d000) at 0x1327dd228
>     REFCNT = 1
>     FLAGS = (OBJECT)
>     STASH = 0x1327dd1f8	"HTML::TokeParser::Corinna::Exception"
>     MAXFIELD = 1
>     FIELDS = 0x600003216430
>     Field No. 0 ($message)
>     SV = PV(0x132852e20) at 0x133046030
>       REFCNT = 1
>       FLAGS = (POK,pPOK)
>       PV = 0x600003216280 "foo"\0
>       CUR = 3
>       LEN = 16
>     Field No. 1 ($stack_trace)
>     SV = PV(0x132852700) at 0x1326b1090
>       REFCNT = 1
>       FLAGS = (POK,pPOK)
>       PV = 0x131ebc390 "Trace begun at lib/HTML/TokeParser/Corinna/Exception.pm line 9\nHTML::TokeParser::Corinna::Exception::__ANON__('HTML::TokeParser::Corinna::Exception=OBJECT(0x1327dd228)', '** argument not available anymore **') called at t/exceptions.t line 16\neval {...} at t/exceptions.t line 16\n"\0
>       CUR = 280
>       LEN = 282
>   PV = 0x1327dd228 ""
>   CUR = 0
>   LEN = 0``
>
>
Ok, so the feature class patch breaks Data::Dumper and likely anything XS
that expects to handle a code ref (including the perl internals).  We
havent taught various bits of code about this new type SvPVOBJ. I guess
there is a good reason we aren't generating a PVCV, but this means a lot of
code will have to be changed to deal with this. I have to admit, I find
this surprising. This means a lot more code has to change than is necessary
if this was just a special type of PVCV. I think this might be a bigger
blocker than Paul realized. We have 233 mentions of PVCV just in the core
repo alone, and god knows how many out there in the wild. A lot of code
will expect that the only form of a code ref will be a PVCV.

Yves

-- 

**Comment by demerphq** [comment]
On Fri, 30 Dec 2022 at 17:23, demerphq ***@***.***> wrote:

> On Fri, 30 Dec 2022 at 17:03, Ovid ***@***.***> wrote:
>
>> @demerphq <https://github.com/demerphq> Doing this:
>>
>> use Devel::Peek;eval { die HTML::TokeParser::Corinna::Exception->new( message => 'foo' ) };
>> Dump($@);
>>
>> Resulted in this:
>>
>> SV = PV(0x13200b6a0) at 0x13200adc8
>>   REFCNT = 1
>>   FLAGS = (ROK)
>>   RV = 0x1327dd228
>>   SV = PVOBJ(0x13303d000) at 0x1327dd228
>>     REFCNT = 1
>>     FLAGS = (OBJECT)
>>     STASH = 0x1327dd1f8	"HTML::TokeParser::Corinna::Exception"
>>     MAXFIELD = 1
>>     FIELDS = 0x600003216430
>>     Field No. 0 ($message)
>>     SV = PV(0x132852e20) at 0x133046030
>>       REFCNT = 1
>>       FLAGS = (POK,pPOK)
>>       PV = 0x600003216280 "foo"\0
>>       CUR = 3
>>       LEN = 16
>>     Field No. 1 ($stack_trace)
>>     SV = PV(0x132852700) at 0x1326b1090
>>       REFCNT = 1
>>       FLAGS = (POK,pPOK)
>>       PV = 0x131ebc390 "Trace begun at lib/HTML/TokeParser/Corinna/Exception.pm line 9\nHTML::TokeParser::Corinna::Exception::__ANON__('HTML::TokeParser::Corinna::Exception=OBJECT(0x1327dd228)', '** argument not available anymore **') called at t/exceptions.t line 16\neval {...} at t/exceptions.t line 16\n"\0
>>       CUR = 280
>>       LEN = 282
>>   PV = 0x1327dd228 ""
>>   CUR = 0
>>   LEN = 0``
>>
>>
> Ok, so the feature class patch breaks Data::Dumper and likely anything XS
> that expects to handle a code ref (including the perl internals).  We
> havent taught various bits of code about this new type SvPVOBJ. I guess
> there is a good reason we aren't generating a PVCV, but this means a lot of
> code will have to be changed to deal with this. I have to admit, I find
> this surprising. This means a lot more code has to change than is necessary
> if this was just a special type of PVCV. I think this might be a bigger
> blocker than Paul realized. We have 233 mentions of PVCV just in the core
> repo alone, and god knows how many out there in the wild. A lot of code
> will expect that the only form of a code ref will be a PVCV.
>

HMM, actually i just realized PVOBJ is distinct from PVCV, so some of the
above is totally wrong, apologies.  So long as it passes is_object() most
code will be fine. Except for code that expects to introspect the object
via the normal type system. What should Data::Dumper do witha
PVOBJ reference? Dump it like it would dump a hash? Same question for
Sereal and Storable and other serialization tools.  So while I was wrong
that the scope of this problem is proportional to PVCV checks, but anything
that expects to introspect on the type of an object will be broken by this
new type.

Yves

-- 

File: pod/perlclass.pod (hunk 10 above)
Comment (rwp0) [reply pod/perlclass.pod]: And better spelled as [*object-oriented programming*](https://en.wiktionary.org/wiki/object-oriented_programming#English) (OOP) with a dash `-`

File: t/class/class.t [hunk 11]
```diff
@@ -0,0 +1,86 @@
...
+    isa_ok($obj, "Test1", '$obj');
+
+    is($obj->hello, "hello, world", '$obj->hello');
+}
+
+# Classes are still regular packages
+{
+    class Test2 {
+        my $ok = "OK";
+        sub NotAMethod { return $ok }
```
rwp0, line 29: Will it (having `my` and `sub` within classes) go away in the future?

What's the use of it? Is it the same as static methods as private methods will have `method :private` syntax I believe.

It's currently a bit confusing.

Ovid: @rwp0 Not sure if I understand your question, but I'll try to answer what I think you're asking.

`sub` will still be allowed in classes. There's no problem with that, but eventually we'll need work to ensure that subroutines cannot be called as methods.

Also, when the MOP is online, we'll need debugger work to ensure that typing `m` only shows subroutines, not methods (might be possibly now, but I haven't looked into it).

`my` will also be allowed in a top-level scope in a `class` (no sense arbitrarily disallowing it). It will effectively be class data, but as a `my` variable would be an internal implementation detail rather than a structural part of the class, its use should be _strongly_ discouraged if the anything outside the class is able to use that data.

Ovid:
@leonerd It appears that declaring a method as a forward declaration causes a segfault:

```
$ perl -Mfeature=class -E 'class Foo { method forward_declaration; }'
class is experimental at -e line 1.
method is experimental at -e line 1.
Segmentation fault: 11
```

pad.c
leonerd: > Will the concept of a "suspended complication CV" be explained in user-facing documentation in a subsequent commit (_i.e.,_ `pod/perlsomething.pod`)?

Not sure what you mean "user-facing"... it is explained in the rest of the sentence after the semicolon (to be a comma, as per below)

> Also, what is the point of the semicolon in the line above. Shouldn't it either be a comma or rewritten as `"suspended complication CV." This can be`

Ahyes a comma probably reads better.

pod/perlclass.pod
leonerd: "Metaprogramming" == a pureperl-accessible API for operating on classes/fields/etc... perhaps under the `meta::` namespace. I.e. the core copy of the various `Object::Pad::MOP::*` APIs.

"Extension customisation" == an XS-visible API for creating new attributes, etc... I.e. a copy of `ObjectPad_register_class_attribute()` and `ObjectPad_register_field_attribute()`.

leonerd:
> Do we have a docker with the image and deps installed yet?

You say "yet" - I'm not sure why. When have we ever done that for any feature branch..?

pod/perlclassguts.pod
leonerd: Yeah I'll reword it better.

leonerd:
> There are no examples of calling object methods. From within a method of the class, can we just call method() or do we need to call $self->method()? Are they equivalent?

Exactly the same as ever they were:

```perl
class C {
   method m1 { ... }
   method m2 { $self->m1(args...); }
}
```

I'll add some words to the doc if you can suggest some.

leonerd:
@Ovid 
> Getting some odd behavior with `Data::Dumper`.

Yeah; this is a known issue - https://github.com/leonerd/perl5/issues/18

An acceptable solution to it has yet to reveal itself. It isn't clear what behaviour `Data::Dumper` et.al. should take. See the above issue, and various other issues linked from it.

leonerd:
@Ovid 
> @leonerd I don't know if this is intended, but I thought we were also injecting a `$class` variable into methods. This is failing:
> 
> Changing `$class` to `(ref $self)` works, but it's ugly.

Hasn't been implemented so far. I was thinking of naming it `__CLASS__` so it's less surprising when it becomes visible in field initialiser expressions:

```perl
field $wibbler = __CLASS__->build_wibbler;
```

See also https://metacpan.org/pod/Object::Pad#__CLASS__

pad.c
```diff
@@ -2844,6 +2844,91 @@ Perl_padname_dup(pTHX_ PADNAME *src, CLONE_PARAMS *param)
...
+
+Resumes the parser state previously saved using the C<suspend_compcv> for a
+final time before being compiled into a full CV.  This should be used within
+an C<ENTER>/C<LEAVE> scoped pair.
+
+=for apidoc A|void|resume_compcv_and_save|struct suspended_compcv *buffer
+
+Resumes a buffer previously suspended by the C<suspend_compcv> function, in a
+way that will be re-suspended at the end of the scope so it can be used again
+later.  This should be used within an C<ENTER>/C<LEAVE> scoped pair.
```
tonycoz:
```
./miniperl -Ilib autodoc.pl
Function 'resume_compcv_and_save', documented in pad.c, not listed in embed.fnc at autodoc.pl line 1819.
Function 'resume_compcv_final', documented in pad.c, not listed in embed.fnc at autodoc.pl line 1819.
```

Presumably these need m entries in embed.fnc

tonycoz:
There appears to be some mishandling of unicode class names, I was looking for a call on the wrong class name, but got a crash instead:
```
tony@venus:.../git/perl6$ cat ../feature-class.pl
use v5.32;
use feature 'class';
use utf8;
use Devel::Peek;

class Ł {
    method foo { say "foo" }
}

Ł->new()->foo();
Dump(Ł->new);
tony@venus:.../git/perl6$ gdb --args ./perl -Ilib ../feature-class.pl 
...
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./perl...
(gdb) r
Starting program: /home/tony/dev/perl/git/perl6/perl -Ilib ../feature-class.pl
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
class is experimental at ../feature-class.pl line 6.

Program received signal SIGSEGV, Segmentation fault.
0x00005555555e2933 in Perl_get_av (name=0x555555ca6d60 "Ł::ISA", 
    flags=536870912) at perl.c:2863
2863            return GvAVn(gv);
(gdb) p gv
$1 = (GV * const) 0x0
(gdb) bt
#0  0x00005555555e2933 in Perl_get_av (name=0x555555ca6d60 "Ł::ISA", 
    flags=536870912) at perl.c:2863
#1  0x000055555579b58d in Perl_class_setup_stash (stash=0x555555c9bd88)
    at class.c:324
#2  0x000055555566da2d in Perl_yyparse (gramtype=258) at perly.y:612
#3  0x00005555555e1856 in S_parse_body (env=0x0, 
    xsinit=0x55555559d1f9 <xs_init>) at perl.c:2614
#4  0x00005555555dfb91 in perl_parse (my_perl=0x555555be72a0, 
    xsinit=0x55555559d1f9 <xs_init>, argc=3, argv=0x7fffffffe7e8, env=0x0)
    at perl.c:1923
#5  0x000055555559d114 in main (argc=3, argv=0x7fffffffe7e8, 
    env=0x7fffffffe808) at perlmain.c:106
(gdb) q
A debugging session is active.

        Inferior 1 [process 2018987] will be killed.

Quit anyway? (y or n) y

```

tonycoz:
This appears to be caused by the overloading of the second parameter to get_av() - it's both a boolean, and flags passed onwards.

After a closer look I didn't see a problem with unicode class names.

class.c
```diff
@@ -0,0 +1,1035 @@
...
+    padnamelist_store(aux->xhv_class_fields, PadnamelistMAX(aux->xhv_class_fields)+1, pn);
+    PadnameREFCNT_inc(pn);
+}
+
+static void
+apply_field_attribute_param(pTHX_ PADNAME *pn, SV *value)
+{
+    if(!value)
+        /* Default to name minus the sigil */
+        value = newSVpvn(PadnamePV(pn) + 1, PadnameLEN(pn) - 1);
```
tonycoz: Found the problem purely by testing, then tracked down the probable cause.

```
tony@venus:.../git/perl6$ cat ../feature-class2.pl
use v5.36;
use utf8;
use feature 'class';

class Foo {
    field $Ł :param;
};

Foo->new("Ł" => 1);
tony@venus:.../git/perl6$ ./perl -Ilib ../feature-class2.pl 
class is experimental at ../feature-class2.pl line 5.
field is experimental at ../feature-class2.pl line 6.
Required parameter 'Ł' is missing for "Foo" constructor at ../feature-class2.pl line 9.
```

class.c
```diff
@@ -0,0 +1,1035 @@
...
+
+        /* TODO: We're only obeying ASCII whitespace here */
+
+        /* Trim whitespace at the start */
+        while(value_at < value_max && isSPACE(*value_at))
+            value_at += 1;
+        while(value_max > value_at && isSPACE(*value_max))
+            value_max -= 1;
+
+        if(value_max >= value_at)
```
tonycoz, line 485: Can this be false?  If so, what ends up setting `*valp`?

class.c
```diff
@@ -0,0 +1,1035 @@
...
+            *valp = sv_2mortal(newSVpvn(value_at, value_max - value_at + 1));
```
tonycoz: while the current (and probably future) attribute names are always ASCII ids, the values should support Unicode,

Especially for :isa

The discarding of the flag on the name means that the UTF8 isn't set on the error message, which isn't a problem when the error goes to the terminal, but means `$@` isn't upgraded on error, though it is for other cases:
```
tony@venus:.../git/perl6$ ./perl -Ilib ../feature-check4.pl 
SV = PV(0x559e69b2b050) at 0x559e69b2a780
  REFCNT = 1
  FLAGS = (POK,pPOK,UTF8)
  PV = 0x559e69b3fa20 "Global symbol \"$\xC5\x81\" requires explicit package name (did you forget to declare \"my $\xC5\x81\"?) at (eval 2) line 1.\n"\0 [UTF8 "Global symbol "$\x{141}" requires explicit package name (did you forget to declare "my $\x{141}"?) at (eval 2) line 1.\n"]
  CUR = 109
  LEN = 241
tony@venus:.../git/perl6$ cat ../feature-check4.pl
use strict;
use utf8;
use Devel::Peek;

eval <<'EOS';
$Ł = 1;
EOS

Dump($@);
tony@venus:.../git/perl6$ ./perl -Ilib ../feature-class3.pl 
class is experimental at (eval 2) line 1.
field is experimental at (eval 2) line 2.
SV = PV(0x561f96ddf050) at 0x561f96dde780
  REFCNT = 1
  FLAGS = (POK,pPOK)
  PV = 0x561f96e0bfc0 "Unrecognized field attribute \xC5\x81 at (eval 2) line 3.\n"\0
  CUR = 52
  LEN = 56
tony@venus:.../git/perl6$ cat ../feature-class3.pl
use v5.32;
use feature 'class';
use utf8;
use Devel::Peek;

eval <<'EOS';
class A {
    field $x :Ł(1)
}
EOS

Dump($@);

```

class.c
```diff
@@ -0,0 +1,1035 @@
...
+    call_method("VERSION", G_VOID);
+
+    LEAVE;
+}
+
+#define split_attr_nameval(sv, namp, valp)  S_split_attr_nameval(aTHX_ sv, namp, valp)
+static void S_split_attr_nameval(pTHX_ SV *sv, SV **namp, SV **valp)
+{
+    STRLEN svlen = SvCUR(sv);
+    char *paren_at = (char *)memchr(SvPVX(sv), '(', svlen);
```
tonycoz: I believe every `char *` in this function could be `const char *` since you don't modify the SV.

Unfortunately the cast would still be required for C++ builds.

Ovid:
High level summary.

So far, aside from some POD and C issues, it looks like these are the major concerns:

1. `:isa(Some::Class)` doesn't `require Some::Class`
1. Forward declarations of methods cause segfaults
1. `__CLASS__` and or `$class` need to be available in classes
1. `Data::Dumper` and related tools need to be updated to handle `PVOBJ`
1. Unicode identifiers need to be addressed
1. Miscellaneous segfaults and memory wrap errors when doing strange things with non-scalar fields

Did I miss anything?

Given the size of this PR, that's an impressively low number of issues.

Ovid:
[Section 2.6.1 of the Corinna spec](https://github.com/Ovid/Cor/blob/master/rfc/overview.md#261-syntax) states at the end:

> Note that we also intend for the block to have `strict` and `warnings`, along with disabling indirect method calls.

Is this overlooked, or just not implemented yet?

Ovid:
I was playing around and trying to see what sort of weird behavior I can get. This completely non-sensical code throws a `memory wrap` error. (Changing `%y` to `$y` makes the error go away).

```
$ perl -Mfeature=class -E 'class Foo { use DDP; my $x = 3; field %y = p $x }; Foo->new'
class is experimental at -e line 1.
field is experimental at -e line 1.
panic: memory wrap at -e line 1.
```

[perldiag](https://perldoc.perl.org/perldiag#panic:-memory-wrap) says:

> (P) Something tried to allocate either more memory than possible or a negative amount.

haarg:
> [Section 2.6.1 of the Corinna spec](https://github.com/Ovid/Cor/blob/master/rfc/overview.md#261-syntax) states at the end:
> 
> > Note that we also intend for the block to have `strict` and `warnings`, along with disabling indirect method calls.
> 
> Is this overlooked, or just not implemented yet?

I'd just call that a bug in the spec. There is no reason to tie that to the class feature. In the future, strict, warnings, and class, will all be enabled by `use v5.40;`.

Ovid:
> I'd just call that a bug in the spec. There is no reason to tie that to the class feature. In the future, strict, warnings, and class, will all be enabled by use v5.40;.

I can see that, but code that simply has `use feature 'class';` without an explicit `use v$version;` may be rather surprising to people. Caused me some grief in the code I'm porting before I realized the spec and the implementation don't match.

pod/perlclass.pod
```diff
@@ -0,0 +1,310 @@
...
+
+A field behaves like a normal lexically scoped variable. It has a sigil and is
+private to the class (though creation of an accessor method will make it
+accessible from the outside). The main difference is that different instances
+access different values in the same scope.
+
+    class WithFields {
+        field $scalar = 42;
+        field @array  = qw(this is just an array);
+        field %hash   = (species => 'Marsian', planet => 'Mars');
```
Ovid: Minor nit: `s/Marsian/Martian/` for the accepted term :)

Ovid:
I do not know if this is intended or not, but assigning a default value to a hash attribute doesn't appear to work:

```perl
use DDP;
use experimental 'class';
class Some::Class {
    field $token : param;
    field %attr = $token->[0]->%*;
    field $attr = $token->[0];

    method hash ()    { return %attr }
    method hashref () { return $attr }
}

my $object = Some::Class->new(
    token => [
        { bgcolor => 'white' },
    ]
);
my %attributes = $object->hash;
p %attributes;
my $attributes = $object->hashref;
p $attributes;
```

The above prints:

```
{}
{
    bgcolor   "white"
}
```

I then tested that with an array attribute and got a segmentation fault:

```perl
use DDP;
use experimental 'class';
class Some::Class {
    field $token : param;
    field @attr = keys $token->[0]->%*;
}

my $object = Some::Class->new(
    token => [
        { bgcolor => 'white' },
    ]
);
```

Output:

```
/bin/bash: line 1: 45756 Segmentation fault: 11  perl -Ilib class.pl
```

For now, the above can be worked around with an `ADJUST` block.

```perl
class Some::Class {
    field $token : param;
    field %attr;
    field @attr;
    field $attr;
    ADJUST {
        %attr = $token->[0]->%*;
        @attr = keys $token->[0]->%*;   # better: @attr = keys %attr;
        $attr = $token->[0];
    }

    method hash ()    { return %attr }
    method array ()   { return @attr }
    method hashref () { return $attr }
}
```

Ovid:
Minor nit:

```perl
use v5.37.8;
use DDP;
use experimental 'class';
class Some::Class {
    method foo () {}
    method foo () {}
}
```

That generated `Subroutine foo redefined at class.pl line 5.`. Should be `Method foo redefined at class.pl line 5.`.

pad.c
leonerd: But if I do that then it generates new lines in proto.h:

```
+/* PERL_CALLCONV void
+resume_compcv_and_save(pTHX_ struct suspended_compcv *buffer); */
+#define PERL_ARGS_ASSERT_RESUME_COMPCV_AND_SAVE
+
+/* PERL_CALLCONV void
+resume_compcv_final(pTHX_ struct suspended_compcv *buffer); */
+#define PERL_ARGS_ASSERT_RESUME_COMPCV_FINAL
+
```

Admittedly those function prototypes are commented out, but the assert macros remain.

class.c
```diff
@@ -0,0 +1,1035 @@
...
+            SAVETMPS;
+            SPAGAIN;
+
+            EXTEND(SP, 2);
+
+            PUSHMARK(SP);
+            PUSHs(self);  /* I don't believe this needs to be an sv_mortalcopy() */
+            PUTBACK;
+
+            call_sv((SV *)cvp[i], G_VOID);
```
wolfsage, line 205: Don't we need `SPAGAIN` *after* call_sv?

class.c
leonerd: Ahyes, that needs some `newSVpvn_utf8`. I'll pop that in.

class.c
leonerd: Yeah I'll add newSVpvf_utf8() in here too.

class.c
leonerd: They can indeed. Now edited.

embed.fnc
```diff
@@ -3661,6 +3666,33 @@ XEop	|char  *|dup_warnings	|NULLOK char *warnings
 
 Amd	|const char * const|phase_name					\
 				|enum perl_phase
+
```
demerphq, line 3682: I am confused why none of the new symbols are guarded properly as requested in embed.fnc's doc. Functions that are not public should not be added with no restrictions, otherwise they can be used directly by code that shouldn't on *nix but wont work on Win32 and more restrictive architectures. 

Id expect at least some of the non 'A' functions to be guarded with something like:

```
#if defined(PERL_IN_CLASS_C)
#endif
```
Or possibly with other file name defines listed.  Note I previously requested you add such a define to be compatible with our other source files, this is why it is needed.

demerphq (changes requested):

embed.fnc
tonycoz: > otherwise they can be used directly by code that shouldn't on *nix

This is no longer true for ELF systems where the compiler supports the visibility attribute.

Hiding the names is still useful, since it avoids namespace pollution by the macros.

class.c
leonerd: Ahyes, added.

leonerd:
@Ovid 
> @leonerd I don't know if this is intended, but I thought we were also injecting a `$class` variable into methods.

Not currently. I'll probably add a `__CLASS__` as per Object::Pad, but that can be done in a separate commit later. I just want to get the minimal possible that's useful in first, before expanding out to other things later.

pad.c
leonerd: Ah actually I think that's all working now.

class.c
leonerd: @Ovid:
> 1. `:isa(Some::Class)` doesn't `require Some::Class`

WFM:

```
$ cat BaseClass.pm 
use v5.36;
use feature 'class';
class BaseClass { }
say "Now loaded BaseClass.pm";
1;

$ ./perl -Ilib -I. -E 'use v5.36; use feature "class"; class Derived :isa(BaseClass) { }'
class is experimental at -e line 1.
class is experimental at BaseClass.pm line 3.
Now loaded BaseClass.pm
```

> 2. Forward declarations of methods cause segfaults

Fixed.

> 3. `__CLASS__` and or `$class` need to be available in classes

`__CLASS__` to be added later.

> 4. `Data::Dumper` and related tools need to be updated to handle `PVOBJ`

Much thinking is ongoing about what's best there. There's no conclusion yet.

> 5. Unicode identifiers need to be addressed

I believe that's now fixed.

> 6. Miscellaneous segfaults and memory wrap errors when doing strange things with non-scalar fields

I'll look into this one now.

pod/perlclass.pod
leonerd: Fixed.

leonerd:
@Ovid:
> 1. `:isa(Some::Class)` doesn't `require Some::Class`

WFM:

```
$ cat BaseClass.pm 
use v5.36;
use feature 'class';
class BaseClass { }
say "Now loaded BaseClass.pm";
1;

$ ./perl -Ilib -I. -E 'use v5.36; use feature "class"; class Derived :isa(BaseClass) { }'
class is experimental at -e line 1.
class is experimental at BaseClass.pm line 3.
Now loaded BaseClass.pm
```

> 2. Forward declarations of methods cause segfaults

Fixed.

> 3. `__CLASS__` and or `$class` need to be available in classes

`__CLASS__` to be added later.

> 4. `Data::Dumper` and related tools need to be updated to handle `PVOBJ`

Much thinking is ongoing about what's best there. There's no conclusion yet.

> 5. Unicode identifiers need to be addressed

I believe that's now fixed.

> 6. Miscellaneous segfaults and memory wrap errors when doing strange things with non-scalar fields

I'll look into this one now.

class.c
leonerd: Ahyes. Now added.

embed.fnc
leonerd: @demerphq 
> I am confused why none of the new symbols are guarded properly as requested in embed.fnc's doc.

Oh partly laziness on my part. ;) I just left them all unguarded because I wasn't sure yet where I'd need to call them from. None of the functions added to `embed.fnc` are there *only* for class.c; they're all there because some other part of core needs to call it. I just didn't want to get into working out exactly which parts yet until it was a bit more stable. Also some of them probably should be externally-visible API; I can imagine wanting to call those from XS modules.

For example: `class_prepare_method_parse` and `class_wrap_method_body`. At the moment, the parser calls them directly from action blocks in `perly.y` when compiling `method` syntax. But I could imagine an XS module wanting to do that too. Perhaps it would be better to fold their actions into the behaviours of `start_subparse` and `newATTRSUB`. It's a conversation to be had, I just wasn't ready to have it *yet*. Hence leaving my options open for now.

leonerd:
@Ovid 
> That generated `Subroutine foo redefined at class.pl line 5.`. Should be `Method foo redefined at class.pl line 5.`.

Hmm. Potentially. But then what if you tried to replace a `method` with a `sub` or vice versa? Does the identity of the thing being replaced matter, or the thing replacing it? Both subroutines and methods live in the same namespace.

leonerd:
@Ovid 
> > 6\. Miscellaneous segfaults and memory wrap errors when doing strange things with non-scalar fields
> 
> I'll look into this one now.

Ahyes, there were a bunch of bugs here, but that should now be fixed and more tests added.

tonycoz:
> This appears to be caused by the overloading of the second parameter to get_av() - it's both a boolean, and flags passed onwards.

@leonerd asked in IRC what I meant by this.  Yes, it's still a problem.
```
tony@venus:.../git/perl6$ gdb --args ./perl -Ilib ../feature-class.pl
GNU gdb (Debian 10.1-1.7) 10.1.90.20210103-git
...
Program received signal SIGSEGV, Segmentation fault.
0x00005555555e295f in Perl_get_av (name=0x555555c5db70 "Ł::ISA", 
    flags=536870912) at perl.c:2863
2863            return GvAVn(gv);
(gdb) bt
#0  0x00005555555e295f in Perl_get_av (name=0x555555c5db70 "Ł::ISA", 
    flags=536870912) at perl.c:2863
#1  0x000055555579ceec in Perl_class_setup_stash (stash=0x555555c9d960)
    at class.c:327
#2  0x000055555566da59 in Perl_yyparse (gramtype=258) at perly.y:612
#3  0x00005555555e1882 in S_parse_body (env=0x0, 
    xsinit=0x55555559d1f9 <xs_init>) at perl.c:2614
#4  0x00005555555dfbbd in perl_parse (my_perl=0x555555be92a0, 
    xsinit=0x55555559d1f9 <xs_init>, argc=3, argv=0x7fffffffe7a8, env=0x0)
    at perl.c:1923
#5  0x000055555559d114 in main (argc=3, argv=0x7fffffffe7a8, 
    env=0x7fffffffe7c8) at perlmain.c:106
(gdb) !git describe
v5.37.7-143-g2c22917844
```
It's a bug in get_av() complicated by its interface.

From the get_av() documentation:
> If "flags" is zero and the variable does not exist then NULL is returned.

But in this case flags is non-zero (it's `SVf_UTF8`), but there's no `GV_ADD` so the GV isn't created and gv_fetchpv() returns a NULL GV.

Since flags is non-zero, get_av() blindly calls GvAVn(gv) on the NULL gv and we crash.

I expect changing Perl_get_av() (and get_hv()) to check both flags *and* gv would solve this, but it may mean for the `SVf_UTF8` case that we end up creating the AV on an existing glob when that wasn't desired.

So using the flags parameter as both boolean and flags causes a problem here.

leonerd:
@tonycoz 
> So using the flags parameter as both boolean and flags causes a problem here.

OK, so a much smaller reproducing case:

```
$ ./perl -Ilib -Mfeature=class -cE 'use utf8; class Ĉapelo { }'
class is experimental at -e line 1.
Segmentation fault
```

Seems to be easily solved by:

```diff
--- a/perl.c
+++ b/perl.c
@@ -2859,7 +2859,7 @@ Perl_get_av(pTHX_ const char *name, I32 flags)
 
     PERL_ARGS_ASSERT_GET_AV;
 
-    if (flags)
+    if (flags & ~SVf_UTF8)
         return GvAVn(gv);
     if (gv)
         return GvAV(gv);
```

I'll test with that and see if that has any collateral damage... if not maybe it's a simple bugfix we can split out to its own PR.

tonycoz:
> Seems to be easily solved by:

That looks good to me.

class.c
```diff
@@ -0,0 +1,1033 @@
...
+    PADOFFSET fieldix = aux[0].uv;
+
+    SV *val = NULL;
+
+    switch(PL_op->op_private & (OPpINITFIELD_AV|OPpINITFIELD_HV)) {
+        case 0:
+            if(PL_op->op_flags & OPf_STACKED)
+                val = newSVsv(POPs);
+            else
+                val = newSV(0);
```
richardleach, line 68: `val = newSV(0)` basically just returns a new SVt_NULL. Probably more efficient to do `val = newSV_type(SVt_NULL)`;

class.c
```diff
@@ -0,0 +1,1033 @@
...
+            break;
+
+        case OPpINITFIELD_AV:
+        {
+            AV *av = newAV();
+            if(PL_op->op_flags & OPf_STACKED) {
+                SV **svp = PL_stack_base + POPMARK + 1;
+                STRLEN count = SP - svp + 1;
+
+                av_extend(av, count);
```
richardleach, line 80: More efficient to directly create the AV with SV** array of size count. Maybe like this:
```
    AV *av;
    if(PL_op->op_flags & OPf_STACKED) {
￼        SV **svp = PL_stack_base + POPMARK + 1;
￼        STRLEN count = SP - svp;
￼
￼        av = newAV_alloc_x(count);
￼        while(svp <= SP) {
￼            av_push_simple(av, newSVsv(*svp));
￼            svp++;
￼        }
    } else
        av = newAV();
```
(I haven't had coffee yet this morning, `count` size may need adjusting.)

embed.fnc
jkeenan: Can this conversation be marked resolved now?

leonerd: Nope because I haven't done anything requested in it yet.

t/class/method.t
```diff
@@ -0,0 +1,92 @@
...
+{
+    class Test6 {
+        method forwarded;
+
+        method forwarded { return "OK" }
+    }
+
+    is(Test6->new->forwarded, "OK", 'forward-declared method works');
+}
+
```
JRaspass, line 91: How about another test case checking that class method invocation dies?
```perl
# methods can't be invoked on a non-instance
{
    class Test7 { method meth {} }

    ok(!eval { Test7->meth },
        'Class method invocation of an instance method fails');

    like($@, qr/^Cannot invoke method on a non-instance at /,
        'Failure from class method invocation');
}
```

Also would it be better to have the method name in the error message?

JRaspass (changes requested):
Finally got round to building this PR locally and having a poke (great work btw :tada:). Haven't managed to break anything yet so just trying to think up test cases to improve coverage.

t/class/method.t
leonerd: > How about another test case checking that class method invocation dies?

Ahyes; I thought I added one but apparently not. I'll shamelessly steal your one ;)

> Also would it be better to have the method name in the error message?

Hmm, probably a good idea yes.

leonerd: > Ahyes; I thought I added one but apparently not. I'll shamelessly steal your one ;)

Ohwait I remember now. They're all in t/lib/croak/class.

But your other comment about including the method name still stands.

class.c
leonerd: Ahyes; that looks a bit better. Count was fine - it's count, not max index. So no off-by-one adjustment needed.
Have force-pushed.

t/class/method.t
leonerd: > Also would it be better to have the method name in the error message?

Now added. Force-pushed.

JRaspass (approved):

embed.fnc
leonerd: @demerphq 
> I am confused why none of the new symbols are guarded properly as requested in embed.fnc's doc.

I finally got around to fixing this. It's now ifdef'ed on any of class.c, perly.y, toke.c or pad.c (because some of the field adjustments happen from there).

leonerd:
Does anyone have any feeling about how much I should squash down these commits further? I've kept them split apart for possible interest during review, but I don't know if we'd want to merge it in this much detail. I could squash them down to one big change per stage perhaps.

wolfsage:
IMO if each individual commit contains interesting and useful information on its own, I'd leave them separate and not squash.

demerphq:
> IMO if each individual commit contains interesting and useful information on its own, I'd leave them separate and not squash.

In theory I am +1 on this. The main thing I would go by is "does it compile and pass test at each patch". So I would squash fixup patches (if there are any). And I would squash to achieve the "compile and pass test". 

For things like this I tend to do the following to make sure each patch compiles and passes tests in the patch sequence:

`git rebase origin/blead --exec 'make test'`

adjust the make test to taste, for instance in practice I would spell it as 

`git rebase origin/blead --exec 'make -j16 test_harness'`

Not sure who it was on p5p irc who pointed out the --exec argument to rebase, but its super useful for this kind of thing.  Maybe it was you @wolfsage ? Or @ilmari possibly? Anyway, super useful tool and advice IMO.

Alternatively I would merge it on the command line as a --no-ff merge (I do this so rarely i cant remember how to do it off the top of my head), so that the sequence is kept as a topic branch and thus bisecting doesn't break on the individual patches.

leonerd:
@demerphq 
> In theory I am +1 on this. The main thing I would go by is "does it compile and pass test at each patch". So I would squash fixup patches (if there are any). 

Ahyes. I usually aim to keep them in that state by squashing fixups as I go, so hopefully the branch should already be in that state.

> And I would squash to achieve the "compile and pass test".
> 
> For things like this I tend to do the following to make sure each patch compiles and passes tests in the patch sequence:
...

I'll use that trick to doublecheck now then.

demerphq:
@leonerd I forgot to mention to set TEST_JOBS=16 in the exec. 

`git rebase origin/blead --exec 'TEST_JOBS=16 make -j16 test_harness'`

I have a 4 core cpu, with 8 virtual cores, but I find using 16 processes is the sweet spot for performance. Adjust to your local sweet spot. I found for instance on my box that using 24 processes actually slows things down as compared to 16.

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+If the op has the C<OPf_STACKED> flag it expects to find an initialising value
+on the stack. For SVs this is the topmost SV on the data stack. For AVs and
+HVs it expects a marked list.
+
+=head1 COMPILE-TIME BEHAVIOUR
+
+=head2 C<ADJUST> Phasers
+
+During compiletime, parsing of an C<ADJUST> phaser is handled in a
+fundamentally different way to the the existing perl phasers (C<BEGIN>,
```
rwp0: > fundamentally different way to **the the** existing perl phasers (C<BEGIN>,

Repeated "the"

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+
+The CVs are stored directly, not via RVs. Each CV will be an anonymous real
+function.
+
+=item *
+
+C<xhv_class_fields> will point to a C<PADNAMELIST> containing C<PADNAME>s,
+each being one defined field of the class. They are stored in order of
+declaration. Note however, that the index into this array will not necessarily
+be equal to the fieldix of each field, because in the case of a subclass, the
```
rwp0: better written as: `C<fieldix>`

pod/perlclassguts.pod
```diff
@@ -0,0 +1,410 @@
...
+just before wrapping it in the eventual CV. This function inserts extra ops
+into the optree to make the method work correctly.
+
+=head2 Object Instances
+
+=head3 SVt_PVOBJ
+
+    #define SVt_PVOBJ
+
+An SV type constant used for comparision with the C<SvTYPE()> macro.
```
rwp0: comparis**i**on -> comparison

extra **i**

pod/perlclass.pod
```diff
@@ -0,0 +1,310 @@
...
+Equivalent to
+
+    class X {
+        field $name;
+        method name { return $name; }
+    }
+
+=item * Metaprogramming
+
+An extension of the metaprogramming API (currently proposed by RFC0022) which
```
rwp0: Maybe link RFC's in POD for better visibility esp. by newcomers (or those who are unaware of the RFC) since googling them is hard.

But, that's unfortunately not an RFC yet, only a PR.

https://github.com/Perl/RFCs/pull/25

rwp0 (commented):
Also maybe worthy to note is that the docs stand out in using British English spelling while the rest of Perl POD is using American English spelling.

cf. initialise -> initialize which is caught by spellcheckers, GitHub editors included.

But I think that's okay because TMTOWTDI (ie. variation in written English).

Spelling difference

leonerd:
Thanks @rwp0 for the docs review. I believe I've fixed those all up now.

If we're all happy with this I'd like to start a merge sometime soon. My attack plan is probably going to be three rounds of commandline `merge --no-ff` on each of the three stages, so as to make three new merge commits into blead. Can omit the various "empty" commits while I'm doing that.

Edit: @demerphq's review is still currently in "changes requested" state because of the `#ifdef` guarding in `embed.fnc` but I asked on IRC and he's happy for me to ahead.

leonerd:
This has now been manually merged in stages, so I can close this PR now.

demerphq:
Congrats @leonerd

abraxxa:
@leonerd sorry for the late response but while reading 5.37.9s perlclass, I wondered why you wrote `use v5.36` and not `use v5.38`?

leonerd:
@abraxxa 
`use v5.36` enables the signatures that my example uses, but there's nothing else to be added by `use v5.38` that would make a difference anyway.

abraxxa:
So Perl 5.38 adds the class and field keywords even without use v5.38?

rwp0:
No. Feature bundles have no relation to feature 'class' yet.

Though 5.38 could add module_true which the class keyword probably provides
by default.

28 fev 2023, Ç.a. 20:15 tarixində Alexander Hartmaier <
***@***.***> yazdı:

ilmari:
> So Perl 5.38 adds the class and field keywords even without use v5.38?

No, `use feature 'class';` (which is experimental, so not in any version bundle) does.

PR #20652, opened 2022-12-29
POSIX::fmax - Correct the variable used in example

puck:
The variable in the example should be $max, not $min as we're finding the maximum value.


jkeenan:
Handled manually in f4ae97d054.  Thanks for spotting that.

PR #20653, opened 2022-12-29
regcomp.c etc - rework branch reset (?|: ... ) so it interoperates properly with named capture and recursion

demerphq:
Branch reset was hacked in without much thought about how it might interact with other features. Over time we added named capture and recursive patterns with GOSUB, but I guess because branch reset is somewhat esoteric we didnt notice the accumulating issues related to it.

The main problem was my original hack used a fairly simple device to give multiple OPEN/CLOSE opcodes the same target buffer id. When it was introduced this was fine. When GOSUB was added later however, we overlooked at that this broke a key part of the book-keeping for GOSUB.

A GOSUB regop needs to know where to jump to, and which close paren to stop at. However the structure of the regexp program can change from the time the regop is created. This means we keep track of every OPEN/CLOSE regop we encounter during parsing, and when something is inserted into the middle of the program we make sure to move the offsets we store for the OPEN/CLOSE data. This is essentially keyed and scaled to the number of parens we have seen. When branch reset is used however the number of OPEN/CLOSE regops is more than the number of logical buffers we have seen, and we only move one of the OPEN/CLOSE buffers that is in the branch reset. Which of course breaks things.

Another issues with branch reset is that it creates weird artifacts like this: `/(?|(?<a>a)|(?<b>b))(?&a)(?&b)/` where the `(?&b)` actually maps to the `(?<a>a)` capture buffer because they both have the same id. Another case is that you cannot check if `$+{b}` matched and `$+{a}` did not, because conceptually they were the same buffer under the hood.

These bugs are now fixed. The "aliasing" of capture buffers to each other is now done virtually, and under the hood each capture buffer is distinct. We introduce the concept of a "logical parno" which is the user visible capture buffer id, and keep it distinct from the true capture buffer id. Most of the internal logic uses the "true parno" for its business, so a bunch of problems go away, and we keep maps from logical to physical parnos, and vice versa, along with a map that gives use the "next physical parno with the same logical parno". Thus we can quickly skip through the physical capture buffers to find the one that matched. This means we also have to introduce a logical_total_parens as well, to complement the already existing total_parens. The latter refers to the true number of capture buffers. The former represents the logical number visible to the user.

It is helpful to consider the following table:

```
  Logical:    $1      $2     $3       $2     $3     $4     $2     $5
  Physical:    1       2      3        4      5      6      7      8
  Next:        0       4      5        7      0      0      0      0
  Pattern:   /(pre)(?|(?<a>a)(?<b>b)|(?<c>c)(?<d>d)(?<e>e)|(?<f>))(post)/
```

The names are mapped to physical buffers. So $+{b} will show what is in physical buffer 3. But $3 will show whichever of buffer 3 or 5 matched. Similarly @{^CAPTURE} will contain 5 elements, not 8. But %+ will contain all 6 named buffers.

Since the need to map these values is rare, we only store these maps when they are needed and branch reset has been used, when they are NULL it is assumed that physical and logical buffers are identical.

Currently the way this change is implemented will likely break plug in regexp engines because they will be missing the new logical_total_parens field at the very least. Given that the perl internals code is somewhat poorly abstracted from the regexp engine, with parts of the abstraction leaking out, I think this is acceptable. If we want to make plug in regexp engines work properly IMO we need to add some more hooks that they need to implement than we currently do. For instance mg.c does more work than it should. Given there are only a few plug in regexp engines and that it is specialized work, I think this is acceptable. We can work with the authors to refine the API properly.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -2811,8 +2811,14 @@ Cp	|SV *	|reg_named_buff_all					\
 				|const U32 flags
 
 : FIXME - is anything in re using this now?
+EXp	|void	|reg_numbered_buff_fetch_flags				\
+				|NN REGEXP * const re			\
+				|const I32 paren			\
+				|NULLOK SV * const sv			\
+				|U32 flags
+: FIXME - is anything in re using this now?
 EXp	|void	|reg_numbered_buff_fetch				\
-				|NN REGEXP * const rx			\
+				|NN REGEXP * const re			\
 				|const I32 paren			\
 				|NULLOK SV * const sv
 : FIXME - is anything in re using this now?
--- a/t/test.pl
+++ b/t/test.pl
@@ -357,6 +357,16 @@ sub is ($$@) {
     unless ($pass) {
 	unshift(@mess, "#      got "._qq($got)."\n",
 		       "# expected "._qq($expected)."\n");
+        if (defined $got and defined $expected and
+            (length($got)>20 or length($expected)>20))
+        {
+            my $p = 0;
+            $p++ while substr($got,$p,1) eq substr($expected,$p,1);
+            push @mess,"#  diff at $p\n";
+            push @mess,"#    after "._qq(substr($got,$p-40<0 ? 0 : $p-40,40))."\n";
+            push @mess,"#     have "._qq(substr($got,$p,40))."\n";
+            push @mess,"#     want "._qq(substr($expected,$p,40))."\n";
+        }
     }
     _ok($pass, _where(), $name, @mess);
 }
@@ -1093,6 +1103,7 @@ sub fresh_perl {
     # it feels like the least-worse thing is to assume that auto-vivification
     # works. At least, this is only going to be a run-time failure, so won't
     # affect tests using this file but not this function.
+    my $trim= delete $runperl_args->{rtrim_result}; # hide from runperl
     $runperl_args->{progfile} ||= $tmpfile;
     $runperl_args->{stderr}     = 1 unless exists $runperl_args->{stderr};
 
@@ -1104,7 +1115,7 @@ sub fresh_perl {
     my $results = runperl(%$runperl_args);
     my $status = $?;    # Not necessary to save this, but it makes it clear to
                         # future maintainers.
-
+    $results=~s/[ \t]+\n/\n/g if $trim;
     # Clean up the results into something a bit more predictable.
     $results  =~ s/\n+$//;
     $results =~ s/at\s+$::tempfile_regexp\s+line/at - line/g;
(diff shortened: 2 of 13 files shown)


t/test.pl
```diff
@@ -1130,6 +1141,10 @@ sub fresh_perl {
 sub _fresh_perl {
     my($prog, $action, $expect, $runperl_args, $name) = @_;
 
+    local $Level = $Level + 1;
+
+    $expect=~s/[ \t]+\n/\n/g if $runperl_args->{rtrim_result};
```
khwilliamson: Maybe \h instead to account for future tests using Unicode?

hvds:
I will try to make time to look at this in the next couple of weeks. I'd prefer to do so after tests have been added, though.

demerphq:
It already has tests. I updated the text and pr from the first push.

Happy new years!

Yves

demerphq:
FWIW, I would appreciate a review of this sooner than later. I have other work that will definitely conflict with this patch, and id like to avoid having to do the work over and over. @hvds, i understand it will be some time till have a chance to review, but if others are ok with it I will likely merge soon.

So please speak up if you have any issues with this.

t/test.pl
khwilliamson: Actually, should be [:blank:]

embed.fnc
```diff
@@ -2811,8 +2811,14 @@ Cp	|SV *	|reg_named_buff_all					\
 				|const U32 flags
 
 : FIXME - is anything in re using this now?
+EXp	|void	|reg_numbered_buff_fetch_flags				\
```
khwilliamson, line 2814: Is the above FIXME now obsolete?

embed.fnc
```diff
@@ -2811,8 +2811,14 @@ Cp	|SV *	|reg_named_buff_all					\
...
+				|NN REGEXP * const re			\
+				|const I32 paren			\
+				|NULLOK SV * const sv			\
+				|U32 flags
+: FIXME - is anything in re using this now?
```
khwilliamson, line 2819: And this one?

khwilliamson (approved):
This approach is the sanest I can think of.  In looking at the individual lines, , they also seemed sane, but I didn't try to find areas that should be changed that weren't, though the ones I did think about were covered.

t/test.pl
demerphq: Good catch, ill fix before i merge.

embed.fnc
demerphq: I was planning on reviewing these in a separate commit, there is a bunch of comments that I mean to go through. Fixing the curlyx -> curlym discrepancies with capture buffers will entail further changes to the logic related to this function I think, so I will leave it for that.

embed.fnc
demerphq: Same comment as above. ILl review these in the curlyx -> curlym fixups, which i think will impact these functions as well.

t/test.pl
demerphq: fixed

demerphq:
Those pushes have just been to rebase onto blead as other patches are applied. Ill merge once this test round completes.

Issue #7754, opened 2005-01-17
Regex bug: auto-reference to a capture used in a conditional expression


p5pRT:
Migrated from [rt.perl.org#33820](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=33820) (status was 'open')

Searchable as RT33820$

p5pRT:
### From philippe.verdret@xps-pro.com

#### Created by pverdret@xps-pro.com



Hi\,

In a capture a reference to this capture is perfectly possible&#8203;:

&nbsp;  "baaa" =~ \(\\1|a\)\+

the result of this match is b\<aaa>\.
If I replace the back\-reference \\1 by a conditional expression like this&#8203;:

&nbsp;  "baaa" =~ \(\(?\(1\)a|a\)\)\+

the result is the same\. The tested condition is
independant of the capture content so if i change the regex like this&#8203;:

&nbsp;    "baaa" =~ \(\(?\(1\)a|b\)\)\+

the result should be "\<baaaa>"\. But the matched string is \<b>aaaa\.

By introducing a \(?\{\}\) behind "b" the result is now right&#8203;:

&nbsp;    "baaa" =~ \(\(?\(1\)a|b\(?\{\}\)\)\)\+

use re 'debug' indicates that the empty code block changes the
generated opcodes\.

Philippe Verdret

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=medium

This perlbug was built using Perl v5.8.6 - Mon Dec 13 09:51:32 2004
It is being executed now by  Perl v5.8.4 - Mon Jun  7 08:31:41 2004.

Site configuration information for perl v5.8.4:

Configured by CRB at Mon Jun  7 08:31:41 2004.

Summary of my perl5 (revision 5 version 8 subversion 4) configuration:
  Platform:
    osname=MSWin32, osvers=4.0, archname=MSWin32-x86-multi-thread
    uname=''
    config_args='undef'
    hint=recommended, useposix=true, d_sigaction=undef
    usethreads=undef use5005threads=undef useithreads=define 
usemultiplicity=define
    useperlio=define d_sfio=undef uselargefiles=define usesocks=undef
    use64bitint=undef use64bitall=undef uselongdouble=undef
    usemymalloc=n, bincompat5005=undef
  Compiler:
    cc='cl', ccflags ='-nologo -Gf -W3 -MD -Zi -DNDEBUG -O1 -DWIN32 
[...]

PR #20654, opened 2022-12-29
regcomp.c - make sure conditional match (GROUPP) is not broken by CURLYM

demerphq:
This disables the CURLY->CURLYM optimization when the contents contains a GROUPP. 

This PR also includes a prior commit which fixes this a different, less efficient way, and then corrects that to a better way afterwards.

See https://github.com/Perl/perl5/issues/7754

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -9050,6 +9050,14 @@ NULL
                           depth, (IV) ST.count, (IV)ST.alen)
             );
 
+            if (ST.me->flags) {
+                /* emulate CLOSE: mark current A as captured */
+                U32 paren = (U32)ST.me->flags;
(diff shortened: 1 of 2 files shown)


demerphq:
CURLYM fakes an OPEN/CLOSE paren, and was deferring updating its close paren data until after it finished looping. This meant that tests against the capture buffer would not work. Changed it to close the parens each time. An alternative solution would be to exclude cases containing conditional logic from the CURLYX->CURLYM conversion, just like EVAL is excluded.

demerphq:
I will merge this tomorrow unless there are objections.

khwilliamson:
I've never understood this area of the code very well, but it looks plausible to me

PR #20660, opened 2022-12-31
DBM_Filter/t - Use a private temporary directory for tests

demerphq:
I missed a parallelism issue in the DBM_Filter tests. This patch makes 01error.t and 02core.t use temporary directories for their tests so they don't collide.


demerphq:
Merging immediately so this doesn't block anyones work. I can see CI breaking because of this for unrelated changes.

PR #20662, opened 2023-01-02
ext/File-Find - use tempdir() for file and dir creation tests

demerphq:
I thought this distribution was safe since all the tests seem to be run inside of dedicated per test file root dirs. But Karl has seen test failures. Using File::Temp::tempdir() to create a private root dir should harden us to any issues that are hidden in the code, even if I have as yet been unable to replicate the test fails that Karl has experienced.

### DIFF
--- a/ext/File-Find/t/find.t
+++ b/ext/File-Find/t/find.t
@@ -35,6 +35,7 @@ use Testing qw(
     file_path
 );
 use Errno ();
+use File::Temp qw(tempdir);
 
 my %Expect_File = (); # what we expect for $_
 my %Expect_Name = (); # what we expect for $File::Find::name/fullname
@@ -235,6 +236,9 @@ sub my_postprocess {
 *file_path_name = \&file_path;
 
 ##### Create directories, files and symlinks used in testing #####
+my $root_dir = cwd();
+my $test_dir = tempdir("FF_find_t_XXXXXX",CLEANUP=>1);
+chdir $test_dir;
 
 mkdir_ok( dir_path('for_find'), 0770 );
 ok( chdir( dir_path('for_find')), "Able to chdir to 'for_find'")
@@ -1111,5 +1115,5 @@ if ($^O eq 'MSWin32') {
     like($@, qr/invalid top directory/,
         "find() correctly died due to undefined top directory");
 }
-
+chdir $root_dir; # let File::Temp cleanup - Switch to `defer {}` one day
 done_testing();
--- a/ext/File-Find/t/taint.t
+++ b/ext/File-Find/t/taint.t
@@ -8,7 +8,6 @@ BEGIN {
         # May be doing dynamic loading while @INC is all relative
         @INC = map { $_ = File::Spec->rel2abs($_); /(.*)/; $1 } @INC;
     }
-
     if ($^O eq 'MSWin32' || $^O eq 'cygwin' || $^O eq 'VMS') {
         # This is a hack - at present File::Find does not produce native names
         # on Win32 or VMS, so force File::Spec to use Unix names.
@@ -32,6 +31,7 @@ use Testing qw(
 );
 use Errno ();
 use Config;
+use File::Temp qw(tempdir);
 
 BEGIN {
     plan(
@@ -45,7 +45,6 @@ my %Expect_File = (); # what we expect for $_
 my %Expect_Name = (); # what we expect for $File::Find::name/fullname
 my %Expect_Dir  = (); # what we expect for $File::Find::dir
 my ($cwd, $cwd_untainted);
-
 
 BEGIN {
     if ($^O ne 'VMS') {
@@ -167,6 +166,16 @@ sub simple_wanted {
 
 *file_path_name = \&file_path;
 
+##### Create directories, files and symlinks used in testing #####
+my $root_dir_tainted = cwd();
+my $root_dir;
+if ($root_dir_tainted=~/^(.*)$/) {
+    $root_dir = $1;
+} else {
+    die "Failed to untaint root dir of test";
+}
+my $test_dir = tempdir("FF_taint_t_XXXXXX",CLEANUP=>1);
+chdir $test_dir;
 
 mkdir_ok( dir_path('for_find_taint'), 0770 );
 ok( chdir( dir_path('for_find_taint')), 'successful chdir() to for_find_taint' );
@@ -333,3 +342,4 @@ SKIP: {
 
     chdir($cwd_untainted);
 }
+chdir $root_dir; # let File::Temp cleanup - Switch to `defer {}` one day
--- a/ext/File-Find/t/taint.t
+++ b/ext/File-Find/t/taint.t
@@ -1,5 +1,6 @@
 #!./perl -T
 use strict;
+use lib qw( ./t/lib );
 
 BEGIN {
     require File::Spec;
@@ -18,16 +19,10 @@ BEGIN {
     require File::Find;
     import File::Find;
 }
-
 use Test::More;
-BEGIN {
-    plan(
-        ${^TAINT}
-        ? (tests => 45)
-        : (skip_all => "A perl without taint support") 
-    );
-}
-use lib qw( ./t/lib );
+use File::Find;
+use File::Spec;
+use Cwd;
 use Testing qw(
     create_file_ok
     mkdir_ok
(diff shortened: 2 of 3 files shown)


demerphq:
For what its worth, I can fake up a parallel test file from before this commit with this:

```
for i in {1..2}; do TESTFILE=harness TEST_JOBS=1 TEST_ARGS='-re File-Find' ./runtests choose & done
```

after this patch the test fails stop. It seems reasonable to me that if we can show that an ext, cpan, dist test file is parallel safe to itself then it is is parallel safe overall, with the caveat of tests that might be inspecting the global disk state (like manifest checks), but we have those separated out to run before tests that are in ext, cpan or dist.

demerphq:
Interesting, this fix exposed a longstanding bug in taint.t.

jkeenan:
I, too, would like to see actual evidence of "test flapping" before applying this pull request.

Why in your commit messages are you spelling `@INC` as `@inc`?  That creates a hyperlink to a GH user whose handle is 'inc'.

demerphq:
> I, too, would like to see actual evidence of "test flapping" before applying this pull request.

As far as I am concerned I have seen enough to want to fix it. Arguably we should be able to run any test at the same time as itself except for some very specific exceptions. But this fails every time I run it:

```
for i in {1..2}; do TESTFILE=harness TEST_JOBS=1 TEST_ARGS='-re File-Find' ./runtests choose & done
```

with output like this:

```
cat test.out
../ext/File-Find/t/correct-absolute-path-with-follow.t .. ok
../ext/File-Find/t/correct-absolute-path-with-follow.t .. ok

#   Failed test 'able to mkdir: for_find'
#   at t/lib/Testing.pm line 29.
Unable to mkdir File exists: for_find at t/lib/Testing.pm line 29.
# Tests were run but no plan was declared and done_testing() was not seen.

# Looks like your test exited with 17 just after 5.
#   Failed test 'able to create file: fa/fa_ord'
#   at t/lib/Testing.pm line 22.
Unable to create file: fa/fa_ord at t/lib/Testing.pm line 22.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 2 just after 13.
../ext/File-Find/t/find.t ............................... 
Dubious, test returned 17 (wstat 4352, 0x1100)
Failed 1/5 subtests 
../ext/File-Find/t/find.t ............................... 
Dubious, test returned 2 (wstat 512, 0x200)
Failed 1/13 subtests 

#   Failed test 'able to mkdir: for_find_taint'
#   at t/lib/Testing.pm line 29.
Unable to mkdir File exists: for_find_taint at t/lib/Testing.pm line 29.
/bin/pwd: couldn't find directory entry in ‘..’ with matching i-node
# Looks like your test exited with 17 just after 3.

#   Failed test 'able to mkdir: fa_taint'
#   at t/lib/Testing.pm line 29.
Unable to mkdir No such file or directory: fa_taint at t/lib/Testing.pm line 29.
# Looks like your test exited with 2 just after 5.
../ext/File-Find/t/taint.t .............................. 
Dubious, test returned 17 (wstat 4352, 0x1100)
Failed 43/45 subtests 

Test Summary Report
-------------------
../ext/File-Find/t/find.t                             (Wstat: 4352 (exited 17) Tests: 5 Failed: 1)
  Failed test:  5
  Non-zero exit status: 17
../ext/File-Find/t/taint.t                            (Wstat: 4352 (exited 17) Tests: 3 Failed: 1)
  Failed test:  3
  Non-zero exit status: 17
  Parse errors: Bad plan.  You planned 45 tests but ran 3.
Files=3, Tests=10,  1 wallclock secs ( 0.11 usr  0.00 sys +  0.35 cusr  0.01 csys =  0.47 CPU)
Result: FAIL
Finished test run at Mon Jan  2 17:56:15 2023.
../ext/File-Find/t/taint.t .............................. 
Dubious, test returned 2 (wstat 512, 0x200)
Failed 41/45 subtests 

Test Summary Report
-------------------
../ext/File-Find/t/find.t                             (Wstat: 512 (exited 2) Tests: 13 Failed: 1)
  Failed test:  13
  Non-zero exit status: 2
../ext/File-Find/t/taint.t                            (Wstat: 512 (exited 2) Tests: 5 Failed: 1)
  Failed test:  5
  Non-zero exit status: 2
  Parse errors: Bad plan.  You planned 45 tests but ran 5.
Files=3, Tests=20,  1 wallclock secs ( 0.10 usr  0.00 sys +  0.34 cusr  0.03 csys =  0.47 CPU)
Result: FAIL
Finished test run at Mon Jan  2 17:56:15 2023.
[1]-  Exit 2                  TESTFILE=harness TEST_JOBS=1 TEST_ARGS='-re File-Find' ./runtests choose >> test.out 2>&1
[2]+  Exit 2                  TESTFILE=harness TEST_JOBS=1 TEST_ARGS='-re File-Find' ./runtests choose >> test.out 2>&1
```

With my patch I have never seen it fail. While the fact that something is racy against itself is not necessarily a sign that it is racy against other test files, it does mean it *could* be, and I think that is good enough for this fix.

Also with this PR we fix a bug in taint.t anyway, it does the wrong thing to `@INC` on win32, which just happens to work because we don't try to load anything after it has been corrupted. 

> Why in your commit messages are you spelling @INC as @inc?

That is a github artifact. If you check on the command line it shows this:
```
commit fd13e08419fc6a3a32d46307dc160154ef730871
Date:   Mon Jan 2 13:28:18 2023 +0100

    File-Find/t/taint.t - do not use rel2abs after we have forced to Unix mode
    
    taint.t corrupts @INC, by using rel2abs twice on its values, once after
    we have forced File::Spec to think we are on a *nix style platform. This
    breaks use statements that occur after the second rel2abs. Comparing to
    File-Find/t/find.t which also does this we can see that find.t only does
    this rel2abs on @INC behavior once. Presumably the second call was a
    rebase error of some sort.
    
    This reorganizes the use statements and ensures that we only use
    rel2abs() to change @INC prior to overriding it to use *nix semantics.
```

demerphq:
BTW, you can see evidence that taint.t is broken on win32 here: https://github.com/Perl/perl5/actions/runs/3822084395/jobs/6502028162#step:9:2918

```
../ext/File-Find/t/correct-absolute-path-with-follow.t ............... ok
# 0: D:\a\perl5\perl5\lib
# 1: D:\a\perl5\perl5\t
Can't locate File/Temp.pm in @INC (you may need to install the File::Temp module) (@INC entries checked: D:/a/perl5/perl5/ext/File-Find/t/lib D:/a/perl5/perl5/ext/File-Find/D:\a\perl5\perl5\lib D:/a/perl5/perl5/ext/File-Find/D:\a\perl5\perl5\t) at t/taint.t line 181.
BEGIN failed--compilation aborted at t/taint.t line 181.
# Looks like your test exited with 22 before it could output anything.
VOLUME = D:
../ext/File-Find/t/find.t ............................................ ok
../ext/File-Find/t/taint.t ........................................... 
Dubious, test returned 22 (wstat 5632, 0x1600)
Failed 45/45 subtests 
```

notice the path `D:/a/perl5/perl5/ext/File-Find/D:\a\perl5\perl5\lib` in the error message. This is because `@INC` gets munged twice when it should only be munged once.

I also have to admit I don't see why you would object to this patch /unless/ you can observe what it fixes. Surely if the code is more correct than it used to be, and is more hardened by using a temp dir that is enough and sufficient to justify the PR?

ext/File-Find/t/find.t
```diff
@@ -1111,5 +1115,5 @@ if ($^O eq 'MSWin32') {
     like($@, qr/invalid top directory/,
         "find() correctly died due to undefined top directory");
 }
-
+chdir $root_dir; # let File::Temp cleanup.
```
leonerd: BTW if there's no back-compat concerns in this file, you could use a `defer` block instead to ensure this always happens even after a failure.

```perl
my $orig_dir = cwd();
defer { chdir $orig_dir; }
```

at the top.

ext/File-Find/t/taint.t
```diff
@@ -342,3 +342,4 @@ SKIP: {
 
     chdir($cwd_untainted);
 }
+chdir $root_dir; # let File::Temp cleanup
```
leonerd: Same here re `defer`

leonerd (approved):
LGTM.

You can apply the `defer` if you want, or just note for another time.

ext/File-Find/t/find.t
demerphq: Enabling defer seems to require me dealing with a bunch of other stuff id rather not look into just now. So I added a comment. Ill revisit it another time.

ext/File-Find/t/taint.t
demerphq: Added a comment for now.

PR #19163, opened 2021-10-02
Simplize copyright claim

Yahasana:
why we don't make it simple and clean

PR #20663, opened 2023-01-02
Simplify copyright years

rwp0:
PR originally submitted by @Yahasana in #19163

This change seems to be *appropriate* as was stated in the comment below:

https://github.com/Perl/perl5/pull/19163#issuecomment-1065839753


demerphq:
So I take it you want to supersede #19163 with this?  ON the assumption this is yes, then I have the following question/comment:

@codesections @StuartJMackintosh - i think there is a related question that should be resolved before we change this. We have year based boiler plate in many of our files. Does that boiler plate have to reflect the history of the file itself? Can it be omitted and referenced back to the README? I don't think we should change one of our files without having a clear and common policy across all our files.

rwp0:
> So I take it you want to supersede #19163 with this? ON the assumption this is yes, then I have the following question/comment:
> 
> @codesections @StuartJMackintosh - i think there is a related question that should be resolved before we change this. We have year based boiler plate in many of our files. Does that boiler plate have to reflect the history of the file itself? Can it be omitted and referenced back to the README? I don't think we should change one of our files without having a clear and common policy across all our files.

Yes. The main reason is that the `README` is special, being a main user-facing file that most visitors see first on this GitHub repository, and it looks really ugly with listing each individual year instead of simply using a year range with a dash as is commonly done in many other projects worldwide for good.

demerphq:
@rwp0 this PR has conflicts. Will you resolve them?

rwp0:
Resolved through the GitHub web interface.

It created two commits now, is squeezing them required here?

demerphq:
@rwp0 better to resolve them on a command line and force push them as a single patch please.

rwp0:
> force

Okay, will try.

rwp0:
> error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.

Getting the above error when cloning my `Perl/perl5` fork repository.

Will look at this later then.

BTW the repository always takes a very long time to clone (and many times unsuccessfully, probably due to heavy size), any suggestion to accelerate that with `git clone`?

Tried two times today, one with IDE, one with the CLI, and both failed

---

Anyway, it was a very simple change.

To prevent further delaying on this I'm creating another PR through the interface linking this one to that :-)

PR #20667, opened 2023-01-03
Revert "Revert "Inline strlcat(), strlcpy()""

khwilliamson:
This reverts commit 81620fbedd5f5cb87f240d7809dc669cd60d0139. The original commit was wrongly blamed for breakage, when it was the harness parallelism that was at fault; since fixed, mostly.


tonycoz (approved):

demerphq (approved):
It was me who wrongly blamed it for breakage. In my defense, it did look like it worked. At least for a while.

PR #20670, opened 2023-01-04
locale.c: Don't SAVEFREEPV a void

khwilliamson:
This appears to have been working, but Newx is documented as returning a void, so that SAVEFREEPV() would have been doing nothing.


richardleach:
LGTM

PR #20671, opened 2023-01-04
Inline several functions from util.c

khwilliamson:
These trivial functions will likely get optimized out.  But I'm unsure if inlining all of them is a good idea

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -2820,8 +2820,8 @@ Cdp	|void	|save_pushptr	|NULLOK void * const ptr		\
 Cp	|void	|save_pushptrptr|NULLOK void * const ptr1		\
 				|NULLOK void * const ptr2		\
 				|const int type
-Aads	|char * |savepv 	|NULLOK const char *pv
-Aads	|char * |savepvn	|NULLOK const char *pv			\
+Aadip	|char * |savepv 	|NULLOK const char *pv
+Aadip	|char * |savepvn	|NULLOK const char *pv			\
 				|Size_t len
 Cdp	|void	|save_rcpv_free |NN char **ppv
 Cp	|void	|save_re_context
@@ -2837,13 +2837,13 @@ Aadp	|char * |savesharedpvn	|NULLOK const char * const pv		\
 				|const STRLEN len
 Cdp	|void	|save_shared_pvref					\
 				|NN char **str
-Aads	|char * |savesharedsvpv |NN SV *sv
+Aadip	|char * |savesharedsvpv |NN SV *sv
 Cp	|void	|save_sptr	|NN SV **sptr
 Cp	|void	|savestack_grow
 Cp	|void	|savestack_grow_cnt					\
 				|I32 need
 Xp	|void	|save_strlen	|NN STRLEN *ptr
-Aads	|char * |savesvpv	|NN SV *sv
+Aadip	|char * |savesvpv	|NN SV *sv
 Adhp	|SV *	|save_svref	|NN SV **sptr
 Aopx	|void	|savetmps
 Cdp	|void	|save_vptr	|NN void *ptr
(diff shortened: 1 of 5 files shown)


inline.h
```diff
@@ -3698,6 +3740,46 @@ Perl_my_strlcpy(char *dst, const char *src, Size_t size)
...
+*/
+
+PERL_STATIC_INLINE bool
+Perl_sv_destroyable(pTHX_ SV *sv)
+{
+    PERL_UNUSED_CONTEXT;
+    PERL_UNUSED_ARG(sv);
+    return TRUE;
+}
+
```
tonycoz: These are only called via function pointers, there's no point to inlining them

khwilliamson: PR updated to not include these two.

embed.fnc
```diff
@@ -627,7 +627,7 @@ CRTopd	|Malloc_t|realloc	|Malloc_t where 			\
 				|MEM_SIZE nbytes
 CTopd	|Free_t |mfree		|Malloc_t where
 
-CTpdR	|void * |get_context
+CTsdR	|void * |get_context
```
tonycoz: Shouldn't these have the `i` flag?

inline.h
```diff
@@ -3624,6 +3624,48 @@ Perl_savesharedsvpv(pTHX_ SV *sv)
     return savesharedpvn(pv, len);
 }
 
+#ifndef PERL_GET_CONTEXT_DEFINED
+
+/*
+=for apidoc_section $embedding
+=for apidoc get_context
+
+Implements L<perlapi/C<PERL_GET_CONTEXT>>, which you should use instead.
```
tonycoz, line 3659: > ..\av.c(1208): error C2129: static function 'void *Perl_get_context(void)' declared but not defined

The version from `win32/win32thread.c` should probably be inlined too, possibly in `win32/win32thread.h`

jkeenan:
> These trivial functions will likely get optimized out. But I'm unsure if inlining all of them is a good idea

@khwilliamson, this pull request is failing to compile during `make` on Win32.  Could you investigate?

demerphq:
'f' has to be the most minimal commit message I have seen so far. ;-)

Typo?

embed.fnc
```diff
@@ -529,7 +529,9 @@
 :   's'  Static function, but function in source code has a Perl_ prefix:
 :
 :	 This is used for functions that have always had a Perl_ prefix, but
-:	 have been moved to a header file and declared static.
+:	 have been moved to a header file and declared static.  If those
+:	 functions are also inlined, the 'i' or 'I' flags should also be
```
demerphq: FWIW, this is not correct. the 's', 'S', 'i' and 'I' (capital I) flags are mutually exclusive. You can only have one. I think you have been mislead as there is a bug in embed.pl that uses the first one in the flags definition. I just happened to stumble on this bug myself independently of this ticket, and will push a PR to fix it in the next minutes. Note the code that looks like this: 

```
                $type = {
                    'S' => 'STATIC',
                    's' => 'STATIC',
                    'i' => 'PERL_STATIC_INLINE',
                    'I' => 'PERL_STATIC_FORCE_INLINE'
                }->{$1};
```

demerphq (changes requested):

embed.fnc
demerphq: See https://github.com/Perl/perl5/pull/20755

embed.fnc
khwilliamson: Yes

embed.fnc
```diff
@@ -6108,11 +6107,13 @@ Cp	|int	|do_spawn	|NN char *cmd
 Cp	|int	|do_spawn_nowait|NN char *cmd
 #endif /* defined(VMS) || defined(WIN32) */
 #if defined(WIN32)
+CTsdR	|void * |get_context
```
tonycoz: I suspect the Win32 failures are because the Win32 Perl_get_context() isn't static.

demerphq: FWIW, I fixed that as part of the conflict resolution on embed.fnc

demerphq:
@khwilliamson I took the liberty of resolving the embef.fnc conflicts and resolved the win32 get_context not being static that was  raised by @tonycoz above at the same time and forced pushed.

demerphq (commented):
Looks ok to me now, but lets see how the tests work out.

demerphq (approved):
Approving to clear the "request changes" status.

tonycoz (approved):

PR #20674, opened 2023-01-04
pv_escape(): Fix compiler warning

khwilliamson:
A parameter to this function was declared const in embed.fnc, but it
isn't const

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -3159,7 +3159,7 @@ Apd	|char * |pv_escape	|NULLOK SV *dsv 			\
 				|const STRLEN count			\
 				|const STRLEN max			\
 				|NULLOK STRLEN * const escaped		\
-				|const U32 flags
+				|U32 flags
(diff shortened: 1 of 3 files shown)


richardleach:
The fourth parameter (`max`) also seems to be declared as a `const` in _embed.fnc_ but isn't `const` in the actual function.

Is the the _locale.c_ change supposed to be part of this PR?

demerphq (approved):

jkeenan:
> The fourth parameter (`max`) also seems to be declared as a `const` in _embed.fnc_ but isn't `const` in the actual function.
> 
> Is the the _locale.c_ change supposed to be part of this PR?

@khwilliamson, have you had a chance to respond to the question from @richardleach?  (The p.r. has garnered one Approved code review.)  Thanks.

demerphq:
Both changes seem reasonable to me, lets apply this and we can address any other embed.fnc issues in a follow up commit.

PR #20677, opened 2023-01-06
fixup many regexp bugs - not to be merged yet

### LINKED ISSUE #8267: [regex] backref problem with quantified groups
Migrated from [rt.perl.org#38133](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=38133) (status was 'open')

Searchable as RT38133$

### LINKED ISSUE #10073: Recursive back references do not always work
Migrated from [rt.perl.org#72020](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=72020) (status was 'open')

Searchable as RT72020$

### LINKED ISSUE #13619: Inconsistent behavior with backreferences nested inside subpattern references
Migrated from [rt.perl.org#121299](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=121299) (status was 'open')

Searchable as RT121299$

### LINKED ISSUE #14848: Regex bug with optional capturing group (<...>)? and backreference
Migrated from [rt.perl.org#125798](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=125798) (status was 'open')

Searchable as RT125798$

### LINKED ISSUE #18865: Backreferences in repeated clusters are inconsistently reset
**Description**
Certain patterns featuring backreferences in a repeated cluster/group fail to match correctly. This seems to be due to an overly-aggressive optimization.

For instance, 
* `"aaa" =~ /(?:(a)?\1)+/` results in `aa` being matched, not `aaa`;
* `"aaa" =~ /(?:((?{})a)?\1)+/` results in `aaa` being matched, the only difference is a no-op `(?{})` assertion added;
* `"ababab" =~ /(?:(?:(ab))?\1)+/` matches `abab`;
* `"ababab" =~ /(?:(?:((?{})ab))?\1)+/` matches `ababab`, same difference as before;
* `"XaaXbbXb" =~ /(?:X([ab])?\1)+/` matches `XaaXbb`;
* `"XaaXbbXb" =~ /(?:X((?{})[ab])?\1)+/` matches `XaaXbbXb`, same difference as before.

I am not actually sure which behavior is correct -- should a non-matching capture group reset or not reset the corresponding backreference, when used inside a repeated cluster? The inconsistency created by introducing a no-op seems to point out to a regexp optimization problem.

All credits for this issue go to ND and to Zoltán Herczeg, who found and analyzed it here: https://lists.exim.org/lurker/thread/20210606.140907.3a4b25d0.en.html . You can find some more examples of the issue in the linked thread.

Thanks!

**Perl configuration**
```
Summary of my perl5 (revision 5 version 30 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=4.19.0
    archname=x86_64-linux-gnu-thread-multi
    uname='linux localhost 4.19.0 #1 smp debian 4.19.0 x86_64 gnulinux '
    config_args='-Dmksymlinks -Dusethreads -Duselargefiles -Dcc=x86_64-linux-gnu-gcc -Dcpp=x86_64-linux-gnu-cpp -Dld=x86_64-linux-gnu-gcc -Dccflags=-DDEBIAN -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -fdebug-prefix-map=/build/perl-Wfb2Cd/perl-5.30.0=. -fstack-protector-strong -Wformat -Werror=format-security -Dldflags= -Wl,-Bsymbolic-functions -Wl,-z,relro -Dlddlflags=-shared -Wl,-Bsymbolic-functions -Wl,-z,relro -Dcccdlflags=-fPIC -Darchname=x86_64-linux-gnu -Dprefix=/usr -Dprivlib=/usr/share/perl/5.30 -Darchlib=/usr/lib/x86_64-linux-gnu/perl/5.30 -Dvendorprefix=/usr -Dvendorlib=/usr/share/perl5 -Dvendorarch=/usr/lib/x86_64-linux-gnu/perl5/5.30 -Dsiteprefix=/usr/local -Dsitelib=/usr/local/share/perl/5.30.0 -Dsitearch=/usr/local/lib/x86_64-linux-gnu/perl/5.30.0 -Dman1dir=/usr/share/man/man1 -Dman3dir=/usr/share/man/man3 -Dsiteman1dir=/usr/local/man/man1 -Dsiteman3dir=/usr/local/man/man3 -Duse64bitint -Dman1ext=1 -Dman3ext=3perl -Dpager=/usr/bin/sensible-pager -Uafs -Ud_csh -Ud_ualarm -Uusesfio -Uusenm -Ui_libutil -Ui_xlocale -Uversiononly -DDEBUGGING=-g -Doptimize=-O2 -dEs -Duseshrplib -Dlibperl=libperl.so.5.30.0'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    bincompat5005=undef
  Compiler:
    cc='x86_64-linux-gnu-gcc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -DDEBIAN -fwrapv -fno-strict-aliasing -pipe -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2 -g'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -DDEBIAN -fwrapv -fno-strict-aliasing -pipe -I/usr/local/include'
    ccversion=''
    gccversion='9.3.0'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='x86_64-linux-gnu-gcc'
    libpth=/usr/local/lib /usr/include/x86_64-linux-gnu /usr/lib /lib/x86_64-linux-gnu /lib/../lib /usr/lib/x86_64-linux-gnu /usr/lib/../lib /lib
    libs=-lgdbm -lgdbm_compat -ldb -ldl -lm -lpthread -lc -lcrypt
    perllibs=-ldl -lm -lpthread -lc -lcrypt
    libc=libc-2.31.so
    so=so
    useshrplib=true
    libperl=libperl.so.5.30
    gnulibc_version='2.31'
  Dynamic Linking:
    dlsrc=dl_dlopen.xs
    dlext=so
    d_dlsymun=undef
    ccdlflags='-Wl,-E'
    cccdlflags='-fPIC'
    lddlflags='-shared -L/usr/local/lib -fstack-protector-strong'


  Compile-time options:
    HAS_TIMES
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_IMPLICIT_CONTEXT
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    USE_64_BIT_ALL
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
    USE_REENTRANT_API
    USE_THREAD_SAFE_LOCALE
  Locally applied patches:
    DEBPKG:debian/cpan_definstalldirs - Provide a sensible INSTALLDIRS default for modules installed from CPAN.
    DEBPKG:debian/db_file_ver - https://bugs.debian.org/340047 Remove overly restrictive DB_File version check.
    DEBPKG:debian/doc_info - Replace generic man(1) instructions with Debian-specific information.
    DEBPKG:debian/enc2xs_inc - https://bugs.debian.org/290336 Tweak enc2xs to follow symlinks and ignore missing @INC directories.
    DEBPKG:debian/errno_ver - https://bugs.debian.org/343351 Remove Errno version check due to upgrade problems with long-running processes.
    DEBPKG:debian/libperl_embed_doc - https://bugs.debian.org/186778 Note that libperl-dev package is required for embedded linking
    DEBPKG:fixes/respect_umask - Respect umask during installation
    DEBPKG:debian/writable_site_dirs - Set umask approproately for site install directories
    DEBPKG:debian/extutils_set_libperl_path - EU:MM: set location of libperl.a under /usr/lib
    DEBPKG:debian/no_packlist_perllocal - Don't install .packlist or perllocal.pod for perl or vendor
    DEBPKG:debian/fakeroot - Postpone LD_LIBRARY_PATH evaluation to the binary targets.
    DEBPKG:debian/instmodsh_doc - Debian policy doesn't install .packlist files for core or vendor.
    DEBPKG:debian/ld_run_path - Remove standard libs from LD_RUN_PATH as per Debian policy.
    DEBPKG:debian/libnet_config_path - Set location of libnet.cfg to /etc/perl/Net as /usr may not be writable.
    DEBPKG:debian/perlivp - https://bugs.debian.org/510895 Make perlivp skip include directories in /usr/local
    DEBPKG:debian/squelch-locale-warnings - https://bugs.debian.org/508764 Squelch locale warnings in Debian package maintainer scripts
    DEBPKG:debian/patchlevel - https://bugs.debian.org/567489 List packaged patches for 5.30.0-9ubuntu0.2 in patchlevel.h
    DEBPKG:fixes/document_makemaker_ccflags - https://bugs.debian.org/628522 [rt.cpan.org #68613] Document that CCFLAGS should include $Config{ccflags}
    DEBPKG:debian/find_html2text - https://bugs.debian.org/640479 Configure CPAN::Distribution with correct name of html2text
    DEBPKG:debian/perl5db-x-terminal-emulator.patch - https://bugs.debian.org/668490 Invoke x-terminal-emulator rather than xterm in perl5db.pl
    DEBPKG:debian/cpan-missing-site-dirs - https://bugs.debian.org/688842 Fix CPAN::FirstTime defaults with nonexisting site dirs if a parent is writable
    DEBPKG:fixes/memoize_storable_nstore - [rt.cpan.org #77790] https://bugs.debian.org/587650 Memoize::Storable: respect 'nstore' option not respected
    DEBPKG:debian/makemaker-pasthru - https://bugs.debian.org/758471 Pass LD settings through to subdirectories
    DEBPKG:debian/makemaker-manext - https://bugs.debian.org/247370 Make EU::MakeMaker honour MANnEXT settings in generated manpage headers
    DEBPKG:debian/kfreebsd-softupdates - https://bugs.debian.org/796798 Work around Debian Bug#796798
    DEBPKG:fixes/autodie-scope - https://bugs.debian.org/798096 Fix a scoping issue with "no autodie" and the "system" sub
    DEBPKG:fixes/memoize-pod - [rt.cpan.org #89441] Fix POD errors in Memoize
    DEBPKG:debian/hurd-softupdates - https://bugs.debian.org/822735 Fix t/op/stat.t failures on hurd
    DEBPKG:fixes/math_complex_doc_great_circle - https://bugs.debian.org/697567 [rt.cpan.org #114104] Math::Trig: clarify definition of great_circle_midpoint
    DEBPKG:fixes/math_complex_doc_see_also - https://bugs.debian.org/697568 [rt.cpan.org #114105] Math::Trig: add missing SEE ALSO
    DEBPKG:fixes/math_complex_doc_angle_units - https://bugs.debian.org/731505 [rt.cpan.org #114106] Math::Trig: document angle units
    DEBPKG:fixes/cpan_web_link - https://bugs.debian.org/367291 CPAN: Add link to main CPAN web site
    DEBPKG:debian/hppa_op_optimize_workaround - https://bugs.debian.org/838613 Temporarily lower the optimization of op.c on hppa due to gcc-6 problems
    DEBPKG:debian/installman-utf8 - https://bugs.debian.org/840211 Generate man pages with UTF-8 characters
    DEBPKG:fixes/getopt-long-4 - https://bugs.debian.org/864544 [rt.cpan.org #122068] Fix issue #122068.
    DEBPKG:debian/hppa_opmini_optimize_workaround - https://bugs.debian.org/869122 Lower the optimization level of opmini.c on hppa
    DEBPKG:debian/sh4_op_optimize_workaround - https://bugs.debian.org/869373 Also lower the optimization level of op.c and opmini.c on sh4
    DEBPKG:debian/perldoc-pager - https://bugs.debian.org/870340 [rt.cpan.org #120229] Fix perldoc terminal escapes when sensible-pager is less
    DEBPKG:debian/prune_libs - https://bugs.debian.org/128355 Prune the list of libraries wanted to what we actually need.
    DEBPKG:debian/mod_paths - Tweak @INC ordering for Debian
    DEBPKG:debian/configure-regen - https://bugs.debian.org/762638 Regenerate Configure et al. after probe unit changes
    DEBPKG:debian/deprecate-with-apt - https://bugs.debian.org/747628 Point users to Debian packages of deprecated core modules
    DEBPKG:debian/disable-stack-check - https://bugs.debian.org/902779 [perl #133327] Disable debugperl stack extension checks for binary compatibility with perl
    DEBPKG:fixes/eumm-usrmerge - https://bugs.debian.org/913637 Avoid mangling /bin non-perl shebangs on merged-/usr systems
    DEBPKG:debian/perlbug-editor - https://bugs.debian.org/922609 Use "editor" as the default perlbug editor, as per Debian policy
    DEBPKG:fixes/gid-parsing - [79e302e] https://bugs.debian.org/941985 [perl #134169] (perl #134169) mg.c reset endptr after use
    DEBPKG:fixes/CVE-2020-10543.patch - [PATCH v530 1/4] regcomp.c: Prevent integer overflow from nested regex quantifiers.
    DEBPKG:fixes/CVE-2020-10878-1.patch - [PATCH v530 2/4] study_chunk: extract rck_elide_nothing
    DEBPKG:fixes/CVE-2020-10878-2.patch - [PATCH v530 3/4] regcomp: use long jumps if there is any possibility of overflow
    DEBPKG:fixes/CVE-2020-12723.patch - [PATCH v530 4/4] study_chunk: avoid mutating regexp program within GOSUB
  Built under linux
  Compiled at Oct 19 2020 10:56:54
  @INC:
    /etc/perl
    /usr/local/lib/x86_64-linux-gnu/perl/5.30.0
    /usr/local/share/perl/5.30.0
    /usr/lib/x86_64-linux-gnu/perl5/5.30
    /usr/share/perl5
    /usr/lib/x86_64-linux-gnu/perl/5.30
    /usr/share/perl/5.30
    /usr/local/lib/site_perl
    /usr/lib/x86_64-linux-gnu/perl-base

```

### LINKED ISSUE #19615: CURLYX v CURLYM disagreement
As part of an unrelated experiment I tried disabling the CURLYX -> CURLYM regexp optimization with the following patch:
```
diff --git a/regcomp.c b/regcomp.c
index 169272e1dd..2dac74fecd 100644
--- a/regcomp.c
+++ b/regcomp.c
@@ -5807,7 +5807,7 @@ S_study_chunk(pTHX_
               nogo:
 
                 /* Try optimization CURLYX => CURLYM. */
-                if (  OP(oscan) == CURLYX && data
+                if (0 && OP(oscan) == CURLYX && data
                       && !(data->flags & SF_HAS_PAR)
                       && !(data->flags & SF_HAS_EVAL)
                       && !deltanext    /* atom is fixed width */
```

I expected this to disable a pure optimization, but was surprised to find that with this change we get some test failures.

The first failure is t/re/pat.t test 104 "monster, length = 300000 at re/pat.t line 373" with the pattern `/(?^:b(?:a|b)+=)/`; I think this one is understandable if CURLYM can match more repeats than CURLYX can (though I hope we can lift the restriction some day).

The remainder are three tests (two newly added for the *ACCEPT work) failing in each of t/re/regexp.t and its variants. I think these show that CURLYX and CURLYM disagree on how captures inside a quantifier should be filled (which I think we also recently discussed). I infer that one or other case has a bug:
```
not ok 966 () ^(aa(bb)?)+$:aabbaa:y:-$1-$2-:-aa-- => '-aa-bb-', match=1
# $subject = "aabbaa";
# 
#                 
#                 $match = ($subject =~ m'^(aa(bb)?)+$') while $c--;
#                 $got = "-$1-$2-";
# 
# $got = "-aa-bb-";
# 
# $expected = "-aa--";

not ok 2080 - ACCEPT with CurlyM optimization GH #19484 () /(A(A|B(*ACCEPT)|C)+D)(E)/:ABDE:y:$1-$2:AB-B => '-B', match=1
# $subject = "ABDE";
# 
#                 
#                 $match = ($subject =~ m/(A(A|B(*ACCEPT)|C)+D)(E)/) while $c--;
#                 $got = "$1-$2";
# 
# $got = "-B";
# 
# $expected = "AB-B";

not ok 2082 - ACCEPT with CurlyM optimization GH #19484 () /(A(A|B(*ACCEPT)|C)+D)(E)/:ABCDE:y:$1-$2:AB-B => '-B', match=1
# $subject = "ABCDE";
# 
#                 
#                 $match = ($subject =~ m/(A(A|B(*ACCEPT)|C)+D)(E)/) while $c--;
#                 $got = "$1-$2";
# 
# $got = "-B";
# 
# $expected = "AB-B";
```

### LINKED ISSUE #20661: MetaBug: Capture Buffer, Quantifiers, and Backtracking Behavior and Optimization
**Description**
We have various tickets reporting issues with backtracking, quantifiers, and backtracking. They all relate to how capture buffers work when quantified, or inside of quantified structures.  Currently we have a lot complex and scatted code to deal with the issues. We should introduce a single comprehensive solution to this, we should define exactly the expected behavior, and we should ensure that we produce that behavior under all situations. Currently we definitely do not.

This issue is intended to be a gathering point for all such bugs and discussion so it can be viewed and managed comprehensively. We   have a variety of issues where @Iabyn has suggested a general model for resolving this, and I would like to over time collect all that discussion here so it can be easily referenced.  Please feel free to comment with more, or to edit this issue if you can and add more tickets here.

Tickets:
https://github.com/Perl/perl5/issues/19615
https://github.com/Perl/perl5/issues/18865
https://github.com/Perl/perl5/issues/14848
https://github.com/Perl/perl5/issues/13619
https://github.com/Perl/perl5/issues/10073

Reviewing the code I notice that we do a lot of things based on the maxopenparen. I wonder if that is an issue with recursion.

We also have to think about the interactions of backrefs, or even forward "backrefs". We have a test case like this: `/((\3|a)(b)?)+/`, which are expected to "see" the previous iterations value for the capture buffer, for instance against a string like "abbbb". Which suggests we need some kind of concept of transactions in terms of the capture buffer values.

I tried zeroing out the capture buffers each iteration, and it breaks $^N, and also various backref code like the above example.

I think the answer might be to not regcppush() the capture buffer state on the 0th iteration. But so far when I try that it segfaults, so i am doing it wrong I guess.

**Steps to Reproduce**
A very simple example is the behavior of these two patterns:
```
"aabbaa"=~/(aa(bb)?)+/ and print "-$1-$2-\n"; # prints "-aa--";
"aabbaa"=~/(aa(?:(bb)|))+/ and print "-$1-$2-\n"; # print "-aa-bb-";
```

We should get the same results as the pattern (?:bb|) and (?:bb)? should be identical. (`?` is just syntactic sugar, as is `+`, in true regular expressions you only have alternation and the `*` (kleene closure), from that you can build + and ? and character classes as syntactic sugar). 

**Expected behavior**
What this suggests is that the expected behavior should be "-aa--", and that "-aa-bb-" is a bug. From there I think we can generalize up and define that the contents of a set of  capture buffers should match the state after the most recent iteration of the last successful iteration. 

**Perl configuration**
N/A. This was created during 5.37 but some of these bugs go back years, if not decades.

demerphq:
This fixes a number of longstanding regexp bugs and issues. It includes debugging changes and other stuff i will rebase away.

The main point is that we use a two layer model to represent capture buffers, and implement a concept of "commiting" the changes, and we store additional data in the CURLY nodes so we can manage the parens they contain properly. This is likely not the most efficient way to do this. The idea was to be correct first, then optimize.

The intended semantics are that: when we have a quantified group we keep track of the parens that it contains, and we ensure that all capture buffers it does contain come from a single iteration of the quantifier.

Consider 

```
./perl -le'"barfoobar"=~/((foo)|(bar))+/ and print "$1-$2-$3=$&";'
bar--bar=barfoobar
```

In older perls:
```
perl -le'"barfoobar"=~/((foo)|(bar))+/ and print "$1-$2-$3=$&";'
bar-foo-bar=barfoobar
```
You can see that in older perls that $2 is from the second iteration, and $3 is from the third.

This patch currently doesn't entirely fix this (yet):
```
./perl -le'"abcaba"=~/((a)(b)(c)|(a)(b)|(a))+/ and print "$1|$2-$3-$4|$5-$6|$7=$&";'
a|a--|a-|a=abcaba
```
where arguably $2 and $5 should be empty, as they werent part of the last succesful match. Currently I have only fixed CURLYX -> CURLYM. And what should happen here is a bit controversial, you could argue this is correct.

However even for this case we have a change from older perls:

```
perl -le'"abcaba"=~/((a)(b)(c)|(a)(b)|(a))+/ and print "$1|$2-$3-$4|$5-$6|$7=$&";'
a|a-b-c|a-b|a=abcaba
```

which really highlights the underlying bug, we have capture buffers here from the first, second and third iteration of the +. IMO that doesn't make sense, and it is inconsistent with the behavior of capture buffers in other contexts.

This PR patch also fixes some other related issues, for instance the min and max have been changed and REG_INFINITY is now I32_MAX, as opposed to U16_MAX. This means that CURLYM and CURLYX now behave the same.

This is intended to fix https://github.com/Perl/perl5/issues/20661 and related bugs. 

Fixes #20661
Fixes #8267 
Fixes #19615
Fixes #18865
Fixes #14848
Fixes #13619
Fixes #10073

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -273,12 +273,14 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen comma_pDEP
 	I32 p;
         for (p = parenfloor + 1; p <= (I32)maxopenparen; p++) {
             Perl_re_exec_indentf(aTHX_
-                "    \\%" UVuf " %" IVdf " (%" IVdf ") .. %" IVdf " (regcppush)\n",
+                "    \\%" UVuf " %" IVdf " (%" IVdf ") .. %" IVdf " new %" IVdf " .. %" IVdf " (regcppush)\n",
                 depth,
                 (UV)p,
                 (IV)RXp_OFFSp(rex)[p].start,
                 (IV)RXp_OFFSp(rex)[p].start_tmp,
-                (IV)RXp_OFFSp(rex)[p].end
+                (IV)RXp_OFFSp(rex)[p].end,
+                (IV)RXp_OFFSp(rex)[p].start_new,
+                (IV)RXp_OFFSp(rex)[p].end_new
             );
         }
     });
@@ -322,8 +324,8 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen comma_pDEP
 
 /* set the start and end positions of capture ix */
 #define CLOSE_ANY_CAPTURE(rex, ix, s, e)                                    \
-    RXp_OFFSp(rex)[(ix)].start = (s);                                       \
-    RXp_OFFSp(rex)[(ix)].end = (e)
+    RXp_OFFSp(rex)[(ix)].start_new = (s);                                   \
+    RXp_OFFSp(rex)[(ix)].end_new = (e)
 
 #define CLOSE_CAPTURE(rex, ix, s, e)                                        \
     CLOSE_ANY_CAPTURE(rex, ix, s, e);                                       \
@@ -336,8 +338,8 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen comma_pDEP
         PTR2UV(rex),                                                        \
         PTR2UV(RXp_OFFSp(rex)),                                             \
         (UV)(ix),                                                           \
-        (IV)RXp_OFFSp(rex)[ix].start,                                       \
-        (IV)RXp_OFFSp(rex)[ix].end,                                         \
+        (IV)RXp_OFFSp(rex)[ix].start_new,                                   \
+        (IV)RXp_OFFSp(rex)[ix].end_new,                                     \
         (UV)RXp_LASTPAREN(rex)                                              \
     ))
 
@@ -363,11 +365,54 @@ S_unwind_paren(pTHX_ regexp *rex, U32 lp, U32 lcp comma_pDEPTH) {
     ));
     for (n = RXp_LASTPAREN(rex); n > lp; n--) {
         RXp_OFFSp(rex)[n].end = -1;
+        RXp_OFFSp(rex)[n].end_new = -1;
     }
     RXp_LASTPAREN(rex) = n;
     RXp_LASTCLOSEPAREN(rex) = lcp;
 }
 #define UNWIND_PAREN(lp,lcp) unwind_paren(rex,lp,lcp)
+
+#define CAPTURE_COMMIT(from_ix, to_ix, force, str) STMT_START {         \
+    U16 my_ix;                                                          \
+    if (force || from_ix) {                                             \
+        for ( my_ix = from_ix; my_ix <= to_ix; my_ix++ ) {              \
+            DEBUG_BUFFERS_r(Perl_re_exec_indentf( aTHX_                 \
+                    "CAPTURE_COMMIT " str ": \\%" IVdf                  \
+                    " std %" IVdf " .. %" IVdf                          \
+                    " ; new: %" IVdf " .. %" IVdf                       \
+                    " ; tmp: %" IVdf                                    \
+                    "\n",                                               \
+            depth, (IV)my_ix,                                           \
+            (IV)RXp_OFFSp(rex)[my_ix].start,                            \
+            (IV)RXp_OFFSp(rex)[my_ix].end,                              \
+            (IV)RXp_OFFSp(rex)[my_ix].start_new,                        \
+            (IV)RXp_OFFSp(rex)[my_ix].end_new,                          \
+            (IV)RXp_OFFSp(rex)[my_ix].start_tmp));                      \
+                                                                        \
+            RXp_OFFSp(rex)[my_ix].start = RXp_OFFSp(rex)[my_ix].start_new;  \
+            RXp_OFFSp(rex)[my_ix].end = RXp_OFFSp(rex)[my_ix].end_new;      \
+        }                                                               \
+    }                                                                   \
+} STMT_END
+
+#define CAPTURE_DUMP(from_ix, to_ix, str) STMT_START {                  \
+    U16 my_ix;                                                          \
+    DEBUG_BUFFERS_r(if (from_ix) {                                      \
+        for ( my_ix = from_ix; my_ix <= to_ix; my_ix++ ) {              \
+            Perl_re_exec_indentf( aTHX_                                 \
+                    "CAPTURE_DUMP " str ": \\%" IVdf " std %" IVdf      \
+                    " .. %" IVdf " ; new %" IVdf " .. %" IVdf           \
+                    " : tmp %" IVdf "\n",                               \
+            depth, (IV)my_ix,                                           \
+            (IV)RXp_OFFSp(rex)[my_ix].start,                            \
+            (IV)RXp_OFFSp(rex)[my_ix].end,                              \
+            (IV)RXp_OFFSp(rex)[my_ix].start_new,                        \
+            (IV)RXp_OFFSp(rex)[my_ix].end_new,                          \
+            (IV)RXp_OFFSp(rex)[my_ix].start_tmp);                       \
+        }                                                               \
+    });                                                                 \
+} STMT_END
+
 
 PERL_STATIC_INLINE void
 S_capture_clear(pTHX_ regexp *rex, U16 from_ix, U16 to_ix, const char *str comma_pDEPTH) {
@@ -382,13 +427,13 @@ S_capture_clear(pTHX_ regexp *rex, U16 from_ix, U16 to_ix, const char *str comma
                 "%" IVdf "(%" IVdf ") .. %" IVdf
                 "\n",
             depth, str, (IV)my_ix,
-            (IV)RXp_OFFSp(rex)[my_ix].start,
+            (IV)RXp_OFFSp(rex)[my_ix].start_new,
             (IV)RXp_OFFSp(rex)[my_ix].start_tmp,
-            (IV)RXp_OFFSp(rex)[my_ix].end,
+            (IV)RXp_OFFSp(rex)[my_ix].end_new,
             (IV)-1, (IV)-1, (IV)-1));
-        RXp_OFFSp(rex)[my_ix].start = -1;
+        RXp_OFFSp(rex)[my_ix].start_new = -1;
         RXp_OFFSp(rex)[my_ix].start_tmp = -1;
-        RXp_OFFSp(rex)[my_ix].end = -1;
+        RXp_OFFSp(rex)[my_ix].end_new = -1;
     }
 }
 
@@ -451,13 +496,15 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p comma_pDEPTH)
     DEBUG_BUFFERS_r(
         for (; paren <= *maxopenparen_p; ++paren) {
             Perl_re_exec_indentf(aTHX_
-                "    \\%" UVuf " %" IVdf "(%" IVdf ") .. %" IVdf " %s (regcppop)\n",
+                "    \\%" UVuf " %" IVdf " (%" IVdf ") .. %" IVdf " new %" IVdf " .. %" IVdf " (regcppop)\n",
                 depth,
                 (UV)paren,
                 (IV)RXp_OFFSp(rex)[paren].start,
                 (IV)RXp_OFFSp(rex)[paren].start_tmp,
                 (IV)RXp_OFFSp(rex)[paren].end,
-                (paren > RXp_LASTPAREN(rex) ? "(skipped)" : ""));
+                (IV)RXp_OFFSp(rex)[paren].start_new,
+                (IV)RXp_OFFSp(rex)[paren].end_new
+            );
         }
     );
 #if 1
@@ -473,8 +520,10 @@ S_regcppop(pTHX_ regexp *rex, U32 *maxopenparen_p comma_pDEPTH)
     for (i = RXp_LASTPAREN(rex) + 1; i <= rex->nparens; i++) {
         if (i > *maxopenparen_p) {
             RXp_OFFSp(rex)[i].start = -1;
+            RXp_OFFSp(rex)[i].start_new = -1;
         }
         RXp_OFFSp(rex)[i].end = -1;
+        RXp_OFFSp(rex)[i].end_new = -1;
         DEBUG_BUFFERS_r( Perl_re_exec_indentf( aTHX_
             "    \\%" UVuf ": %s   ..-1 undeffing (regcppop)\n",
             depth,
@@ -4366,7 +4415,7 @@ S_regtry(pTHX_ regmatch_info *reginfo, char **startposp)
 
     reginfo->cutpoint=NULL;
 
-    RXp_OFFSp(prog)[0].start = *startposp - reginfo->strbeg;
+    RXp_OFFSp(prog)[0].start_new = *startposp - reginfo->strbeg;
     RXp_LASTPAREN(prog) = 0;
     RXp_LASTCLOSEPAREN(prog) = 0;
 
@@ -4399,13 +4448,15 @@ S_regtry(pTHX_ regmatch_info *reginfo, char **startposp)
             ++pp;
             pp->start = -1;
             pp->end = -1;
+            pp->start_new = -1;
+            pp->end_new = -1;
         }
     }
 #endif
     REGCP_SET(lastcp);
     result = regmatch(reginfo, *startposp, progi->program + 1);
     if (result != -1) {
-        RXp_OFFSp(prog)[0].end = result;
+        RXp_OFFSp(prog)[0].end_new = result;
         return 1;
     }
     if (reginfo->cutpoint)
@@ -6582,14 +6633,14 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
         case KEEPS: /*   \K  */
             /* update the startpoint */
             st->u.keeper.val = RXp_OFFS_START(rex,0);
-            RXp_OFFSp(rex)[0].start = locinput - reginfo->strbeg;
+            RXp_OFFSp(rex)[0].start_new = locinput - reginfo->strbeg;
             PUSH_STATE_GOTO(KEEPS_next, next, locinput, loceol,
                             script_run_begin);
             NOT_REACHED; /* NOTREACHED */
 
         case KEEPS_next_fail:
             /* rollback the start point change */
-            RXp_OFFSp(rex)[0].start = st->u.keeper.val;
+            RXp_OFFSp(rex)[0].start_new = st->u.keeper.val;
             sayNO_SILENT;
             NOT_REACHED; /* NOTREACHED */
 
@@ -8033,9 +8084,13 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
             if (RXp_LASTPAREN(rex) < n)
                 sayNO;
 
-            ln = RXp_OFFSp(rex)[n].start;
-            endref = RXp_OFFSp(rex)[n].end;
-            if (ln == -1 || endref == -1)
+            ln = RXp_OFFSp(rex)[n].start_new;
+            endref = RXp_OFFSp(rex)[n].end_new;
+            if (ln < 0 || endref < 0) {
+                ln = RXp_OFFSp(rex)[n].start;
+                endref = RXp_OFFSp(rex)[n].end;
+            }
+            if (ln < 0 || endref < 0)
                 sayNO;			/* Do not match unless seen CLOSEn. */
 
             if (ln == endref)
@@ -8162,6 +8217,7 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
             /* Save all the positions seen so far. */
             ST.cp = regcppush(rex, 0, maxopenparen);
             REGCP_SET(ST.lastcp);
+            /* CAPTURE_COMMIT(0,maxopenparen,1,"GOSUB commit"); */
 
             /* and then jump to the code we share with EVAL */
             goto eval_recurse_doit;
@@ -8316,7 +8372,7 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
                 DEBUG_STATE_r( Perl_re_printf( aTHX_
                     "  re EVAL PL_op=0x%" UVxf "\n", PTR2UV(nop)) );
 
-                RXp_OFFSp(rex)[0].end = locinput - reginfo->strbeg;
+                RXp_OFFSp(rex)[0].end_new = locinput - reginfo->strbeg;
                 if (reginfo->info_aux_eval->pos_magic)
                     MgBYTEPOS_set(reginfo->info_aux_eval->pos_magic,
                                   reginfo->sv, reginfo->strbeg,
@@ -8667,7 +8723,7 @@ S_regmatch(pTHX_ regmatch_info *reginfo, char *startpos, regnode *prog)
 
         case GROUPP:  /*  (?(1))  */
             n = ARG1u(scan);  /* which paren pair */
-            sw = cBOOL(RXp_LASTPAREN(rex) >= n && RXp_OFFS_END(rex,n) != -1);
+            sw = cBOOL(RXp_LASTPAREN(rex) >= n && RXp_OFFS_END(rex,n) >= 0 );
             break;
 
         case GROUPPN:  /*  (?(<name>))  */
@@ -8839,20 +8895,23 @@ NULL
             /* see the discussion above about CURLYX/WHILEM */
             I32 n;
             int min, max;
-            /* U16 first_paren, last_paren; */
+            U16 first_paren, last_paren;
             regnode *A;
 
             assert(cur_curlyx); /* keep Coverity happy */
 
             min = ARG1i(cur_curlyx->u.curlyx.me);
             max = ARG2i(cur_curlyx->u.curlyx.me);
-            /* first_paren = ARG3a(cur_curlyx->u.curlyx.me); */
-            /* last_paren = ARG3b(cur_curlyx->u.curlyx.me);  */
+            first_paren = ARG3a(cur_curlyx->u.curlyx.me);
+            last_paren = ARG3b(cur_curlyx->u.curlyx.me);
             A = REGNODE_AFTER(cur_curlyx->u.curlyx.me);
             n = ++cur_curlyx->u.curlyx.count; /* how many A's matched */
             ST.save_lastloc = cur_curlyx->u.curlyx.lastloc;
             ST.cache_offset = 0;
             ST.cache_mask = 0;
+
+            if (n)
+                CAPTURE_COMMIT(first_paren,last_paren,0,"WHILEM");
 
             DEBUG_EXECUTE_r( Perl_re_exec_indentf( aTHX_  "WHILEM: matched %ld out of %d..%d\n",
                   depth, (long)n, min, max)
@@ -8863,6 +8922,9 @@ NULL
                 ST.cp = regcppush(rex, cur_curlyx->u.curlyx.parenfloor, maxopenparen);
                 cur_curlyx->u.curlyx.lastloc = locinput;
                 REGCP_SET(ST.lastcp);
+
+                if (n)
+                    CAPTURE_CLEAR(first_paren,last_paren,"WHILEM min");
 
                 PUSH_STATE_GOTO(WHILEM_A_pre, A, locinput, loceol,
                                 script_run_begin);
@@ -8875,6 +8937,7 @@ NULL
                 DEBUG_EXECUTE_r( Perl_re_exec_indentf( aTHX_  "WHILEM: empty match detected, trying continuation...\n",
                    depth)
                 );
+                CAPTURE_DUMP(first_paren,last_paren,"empty");
                 goto do_whilem_B_max;
             }
 
@@ -8984,6 +9047,7 @@ NULL
                             maxopenparen);
                 cur_curlyx->u.curlyx.lastloc = locinput;
                 REGCP_SET(ST.lastcp);
+                CAPTURE_CLEAR(first_paren,last_paren,"WHILEM max");
                 PUSH_STATE_GOTO(WHILEM_A_max, A, locinput, loceol,
                                 script_run_begin);
                 NOT_REACHED; /* NOTREACHED */
@@ -9283,7 +9347,7 @@ NULL
                         locinput - reginfo->strbeg);
                 }
                 else
-                    RXp_OFFSp(rex)[paren].end = -1;
+                    RXp_OFFSp(rex)[paren].end_new = -1;
 
                 if (EVAL_CLOSE_PAREN_IS_TRUE(cur_eval,(U32)FLAGS(ST.me)))
                 {
@@ -9327,7 +9391,7 @@ NULL
                                  locinput - reginfo->strbeg);               \
         }                                                                   \
         else {                                                              \
-            RXp_OFFSp(rex)[paren].end = -1;                                 \
+            RXp_OFFSp(rex)[paren].end_new = -1;                             \
             RXp_LASTPAREN(rex)  = ST.lastparen;                             \
             RXp_LASTCLOSEPAREN(rex) = ST.lastcloseparen;                    \
         }                                                                   \
@@ -9647,6 +9711,8 @@ NULL
 #undef ST
 
         case END: /*  last op of main pattern  */
+            if (rex->nparens)
+                CAPTURE_COMMIT(0,(int)rex->nparens,1,"END commit");
           fake_end:
             if (cur_eval) {
                 /* we've just finished A in /(??{A})B/; now continue with B */
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -45,6 +45,51 @@ XXX For a release on a stable branch, this section aspires to be:
 
 [ List each incompatible change as a =head2 entry ]
 
+=head2 Fixed regex engine capture buffer reset bugs and inconsistencies.
+
+Historically the exact rules of what should happen with quantified expressions
+that contain capture buffers were not clear, and the regex engine has been
+somewhat inconsistent in how it reset capture buffers inside of quantified
+groups. This meant that the state of the capture buffer variables $1, $2 could
+change when making supposedly minor and inconsequential changes to a pattern.
+
+After this release the capture buffers in the different branches of a match
+will be consistently mutually exclusive with each other, with the sole exception
+of the case of branch reset C</(?|...)/> which has its own special rules. In
+addition the final state of the capture buffers defined within a quantified
+group will consistently reflect the last successful iteration of the quantifier.
+
+For instance, after executing
+
+    "abcaba" =~ / ( (a) (b) (c) | (a) (b) | (a) )+ /x
+
+the pattern will match and leave C<$1> and C<$7> set to C<"a">, and C<$2>
+through C<$6> will be C<undef>. C<$&> will be the expected C<"abcaba">.
+
+In older perls this expression would result in C<$1> through C<$7> being
+set, with values from the first, second and third iterations of the quantifier
+set at the same time.
+
+An example of the discrepancy in behavior in older perls would be this:
+
+ "ababab" =~ /(?:(?:(ab))?\1)+/      and print "$&";
+ # output 'abab'
+ "ababab" =~ /(?:(?:((?{})ab))?\1)+/ and print "$&";
+ # outputs 'ababab'
+
+Another is this:
+
+ "A" =~ /(((?:A))?)+/;
+ my $first = $2;
+
+ "A" =~ /(((A))?)+/;
+ my $second = $2;
+
+in older perls $first and $second would not be the same. In newer perls they
+will.
+
+This may break code that was depending on the older inconsistent behaviour.
+
 =head1 Deprecations
 
 XXX Any deprecated features, syntax, modules etc. should be listed here.
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -382,9 +382,9 @@ Perl_rxres_save(pTHX_ void **rsp, REGEXP *rx)
     /* deal with regexp_paren_pair items */
     if (!p || p[1] < RX_NPARENS(rx)) {
 #ifdef PERL_ANY_COW
-        i = 7 + (RX_NPARENS(rx)+1) * 2;
+        i = 7 + (RX_NPARENS(rx)+1) * 4;
 #else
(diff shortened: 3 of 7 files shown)


demerphq:
https://github.com/Perl/perl5/issues/20661

demerphq:
This now works as I believe is expected, and matches PCRE etc. I am pretty sure it is nothing as elegant and efficient as it could be using the regexp state machine, but I do believe it is now correct.  You have to start somewhere. :-)

The important gist of the changes are:

The  offset data has been doubled to hold a "start/end" and "start_new/end_new" pairs. OPEN/CLOSE etc update start_new/end_new.  All of the buffer read logic has been changed to check both, first the "_new" variants, then the standard ones. At the and of a successful iteration of the quantifier we "commit" the state of the _new data to the standard data. Prior to a new iteration we clear the _new data.  All CURLY style operators now store the min/max as I32's and they store the "first paren" and "last paren" of their contents.

All branches now store the number of parens before them, and contained by them. As do tries. After a failed branch we clear the _new data for the parens it contains. (As well as whatever else we did historically).

There are a few other minor changes embedded into these patches which i will tease out, but that is the main point. @iabyn if you have suggestions on how to make it more efficient I would love to hear them.

demerphq:
@jkeenan why did you add a "do not merge" to this /draft/ PR? Doesn't that kinda defeat the point of it being draft?

jkeenan:
On 1/7/23 17:04, Yves Orton wrote:
> @jkeenan <https://github.com/jkeenan> why did you add a "do not merge" 
> to this /draft/ PR? Doesn't that kinda defeat the point of it being draft?
> 
>

I don't think you need "do not merge" in the Subject.  But do whatever 
you want.

demerphq:
@kid51 the point is that is a note to me, before I change it from draft ill probably split some of those patches apart, and I will certainly remove a bunch of debug crap. I wont change it from draft until I have done so.I thought do-not-merge is for ready PR's that we are sitting on till a later release cycle.  Anyway, it doesn't matter, it just struck me as redundant.

regcomp_study.c
```diff
@@ -3469,7 +3469,7 @@ Perl_study_chunk(pTHX_
             }
         }
         else if (OP(scan) == EVAL) {
-            if (data)
+            if (data && !(scan->flags & 128) )
```
tonycoz: This magic number deserves a name.

pod/perlre.pod
```diff
@@ -1990,6 +1990,18 @@ keep track of the number of nested parentheses. For example:
...
+Use a double block to create an optimistic form of this construct.
+C<(?{{ ... }})> should not disable any optimisations.
+
+=item C<(?{{ I<code> }})>
+X<(?{{}})> X<regex, optimistic code>
+
+This is *exactly* the same as C<(?{ I<code> })> with the exception
+that it does not disable any optimisations in the regex engine. It
+may be even more volatile than the normal form. Introduced in 5.37.7
+it is backwards compatible with older perls where it is a no-op.
```
tonycoz: Note quite exactly the same:
```
tony@venus:.../git/perl6$ ./perl -le '"aaa" =~ /(?:.(?{ local $x = $x+1; print $x }))+/'
1
2
3
tony@venus:.../git/perl6$ ./perl -le '"aaa" =~ /(?:.(?{{ local $x = $x+1; print $x }}))+/'
1
1
1
```
But changing this would be backward incompatible, since local in `(?:{})` works like this historically.

tonycoz:
I've had a look over it, but I'm not especially familiar with the regexp engine.

regcomp_study.c
demerphq: Agreed. Will fix.

pod/perlre.pod
demerphq: BAH. That sucks. I had forgotten about that.  Back to the drawing board on that one.

demerphq:
> I've had a look over it, but I'm not especially familiar with the regexp engine.

Your feedback is already super useful, thanks. I need to think on what to do about `(??{{ }})` and friends. I had hoped this trick with the extra block would Just Work. Now I need to think about this.  How important is the local thing? If someone is using (?{{}}) maybe they dont want to use local. I have a feeling this is hopeless optimism and I just need to figure out an alternative. Which means messing with the toker, which I had really wanted to avoid. That code is gnarly.

iabyn:
On Sat, Jan 07, 2023 at 10:01:15AM -0800, Yves Orton wrote:
> The  offset data has been doubled to hold a "start/end" and
> "start_new/end_new" pairs. OPEN/CLOSE etc update start_new/end_new.  All
> of the buffer read logic has been changed to check both, first the
> "_new" variants, then the standard ones. At the and of a successful
> iteration of the quantifier we "commit" the state of the _new data to
> the standard data. Prior to a new iteration we clear the _new data.  All
> CURLY style operators now store the min/max as I32's and they store the
> "first paren" and "last paren" of their contents.

I have zero enthusiasm to review this currently. This is because
its a big patch, with practically no explanation of what fundamental
problems it's trying to fix, and what its doing to fix them.

-- 
The optimist believes that he lives in the best of all possible worlds.
As does the pessimist.

demerphq:
@iabyn - fair enough. I thought I had explained well enough in the PR documentation.  I didnt bother in the commits because I am fairly certain I am not doing things efficiently and I would like your advice on how to do them efficiently so i expect to rip them up after I get some of your advice. What I have written in the text to this PR is more relevant generally than the code. But clearly I have not explained myself sufficiently.

I am trying to solve the issues of clearing capture buffers properly, and setting them properly inside of quantified expressions, especially I am trying to resolve the issues that come from CURLYX and CURLYN and CURLYM not being equivalent.  The code says that converting CURLYX to CURLYM or CURLYN is a "powerful optimization", which is correct in a sense, but it is not purely an optimization as it turns out CURLYM and CURLN do things "right" as far as capture buffers go, CURLYX does not. Which is problematic, as we disable these optimizations under various circumstances. Meaning a minor change to the pattern can result in the state of the capture buffers afterwards being different. 

Essentially there are two core bugs here. They both are solved using the same underlying machinery, teaching the relevant regops to know how many parens are bfore them, and how many are defined within them, a macro called CAPTURE_CLEAR() and a macro called CAPTURE_COMMIT(), and the addition of "transactional transparent capture buffers" which are committed only when a quantifier iteration is successful.  

**BUG 1** 
```
perl -le'"foobarfoo"=~/((foo)|(bar))+/ and print "$1-$2-$3=$&"' #wrong
foo-foo-bar=foobarfoo 
```

This should not leave $2 AND $3 set at the same time. Either $3 should be undef, and $1 should equal $2. Or, $2 should be undef and $1 should equal $3. It should work like the branch does:

```
./perl -le'"foobarfoo"=~/((foo)|(bar))+/ and print "$1-$2-$3=$&"' #right
foo-foo-=foobarfoo
```

**SOLUTION 1**
I have added the idea of "temporary offset data" which is "committed" when a WHILEM  or END executes.. And I have taught CURLY style regops to store the number of parens before them, and the number which are contained by them so that the CURLY/WHILEM can reset them appropaitely each loop.

For 1 this means that struct regexp_paren_pair  gets two new members, start_new and end_new. Like this:

```
typedef struct regexp_paren_pair {
    SSize_t start;
    SSize_t end;

    SSize_t start_new;
    SSize_t end_new;

    /* 'start_tmp' records a new opening position before the matching end
     * has been found, so that the old start and end values are still
     * valid, e.g.
     *    "abc" =~ /(.(?{print "[$1]"}))+/
     *outputs [][a][b]
     * This field is not part of the API.  */
    SSize_t start_tmp;
} regexp_paren_pair;
```

The logic that reads offset data has been changed to know about the start_new,end_new variants, and to check them first. When they are -1 the code then looks at the start,end pair. Thus the _new buffers are "transparent", when not set you can see the old value.  They are transactional because the OPEN/CLOSE logic writes into the _new buffers, and the value of the _new buffer is only "committed" to the standard buffer at certain times (WHILEM, END and GOSUB). 

Each loop of a quantifier is treated as a transaction. Either it succeeds in which case the state is committed, or it fails in which case it is cleared. The overall pattern is treated as a loop as with 1 iteration, so END is also a commit.  

So for instance in `"foobarfoo"=~/((foo)|(bar))+/` we first match foo, so offs[2].start_new = 0, offs[2].end_new=3, we hit the end of the + loop, and commit the data for $2 and $3, so now offs[2].start = 0, and offs[2].end = 3 as well, offs[3] stays -1 in both data. We then clear offs[2].start_new and offs[2].end_new by setting them to -1. We execute the loop again, and this time we match 'bar', so, we end up committing $2=(-1,-1) $3=(3,7);,  We execute the loop again, and this time we mach 'foo' again, and so we end up committing $2=(7,10), $3=(-1,-1). We execute it a third time, which fails to match either, and we are done.

With this CURLYX and CURLYM and CURLYN behave the same.

**BUG 2**
The other problem I am trying to solve is capture buffers from multiple branches of a pattern being set after an alternation. The capture buffers defined in different branches of an alternation should be mutexed against each other (leaving aside branch reset patterns). So this is wrong:

```
perl -le'"abcaba"=~/((a)(b)(c)|(a)(b)|(a))+/ and print "$1|$2-$3-$4|$5-$6|$7=$&"'; # wrong
a|a-b-c|a-b|a=abcaba
```
This should result in either $2 $3 and $4 being set, or $5 and $6, or $7. And $1 should either equal "$2$3$4" or "$5$6" or $7. Like this:

```
./perl -le'"abcaba"=~/((a)(b)(c)|(a)(b)|(a))+/ and print "$1|$2-$3-$4|$5-$6|$7=$&"'; # right
a|--|-|a=abcaba
```

**SOLUTION 2**
I have added the idea that a BRANCH node should clear the buffers it contains when the branch fails. So I have taught the BRANCH logic to store the number of parens before them, and then number which are contained by them as well.

For problem 2, we do something similar. Each BRANCH knows how many capture buffers are contained in its branch, and on failure clears those capture buffers (again by using CAPTURE_CLEAR() and clearing the _new state before they execute their payload. so in /(a)(b)(c)|(a)(b)|(a)/ we end up with: BRANCH(0,3) OPEN1 ... OPEN2 ... OPEN3 BRANCH(3,5) OPEN4 ... OPEN5 ... BRANCH(5,6) OPEN6 ... END

So the first branch when it fails clears buffer 1..3, the second when it fails clears 4..5, and when the third branch fails it clears 6. 

The end result is that the state of capture buffers in branches is mutexed.

**WHAT I NEED FROM YOU**
Everything else in this PR is a minor fix and I don't need your help with. What I really need is your insight into how to exploit the state machine to do the above stuff efficiently, and not have to loop over all the capture buffers for instance to clear or commit them. Or any other suggestions you might have to resolve the above bugs. I am satisfied my PR has the correct semantics and consistent behavior, I just need advice on how to make it efficient. I think maybe I can get rid of a bunch of this patch, maybe even most of it by using the information I have added to the regops in terms of buffers before and contained, to update maxopenparen properly. But I am not entirely sure if i can, GOSUB and other non-linear pattern constructs seem to suggest that wont work. Consider "abac" /((a)|(b))+(c)/, you cant just set maxopenparen to 2 after the last iteration of the loop, as when (c) matches it will get set back to 4, and if the buffer for $3 has not been cleared the value from the previous iteration of the quantifier loop will be visible again.

demerphq:
I have updated the PR by breaking up the main changes in the PR. The two patches I would really like some feed back from @iabyn are https://github.com/Perl/perl5/pull/20677/commits/ff43150a169745a295340ff18960efbbfa839be4
and https://github.com/Perl/perl5/pull/20677/commits/99429446d08cb71cbcb3eb05d36d88e6a95a0c86

pod/perlre.pod
demerphq: Changed it to `(**{ })` and `(*{...})`

demerphq:
@PhilipHazel could you please check this over? This is an attempt to make Perls capture buffers behave more consistently. It hopefully brings PCRE and Perl into agreement in more cases.

pod/perlre.pod
demerphq: changed to `(*{...})` and `(**{...})` in the most recent patch.

regcomp.h
```diff
@@ -216,13 +216,13 @@ struct regnode_2L {
     I32 arg2;
 };
 
-/* 'Two field' -- Two 16 bit unsigned args */
+/* 'Two field' -- Two 32 bit signed args */
```
khwilliamson: This makes regnode_2 and regnode_2L be the same size; with the only difference being the signedness of one argument.  That means the names are no longer valid.  Is it the case that the 2L needs a U32? or could all the ones that need 32 bit operands use the same regnode?

regcomp.h
```diff
@@ -327,12 +327,12 @@ struct regnode_ssc {
...
 #if SHORTSIZE > 2
 #  ifndef REG_INFTY
-#    define REG_INFTY  nBIT_UMAX(16)
+#    define REG_INFTY  nBIT_IMAX(32)
 #  endif
 #endif
 
 #ifndef REG_INFTY
-#  define REG_INFTY U16_MAX
+#  define REG_INFTY I32_MAX
```
khwilliamson: I believe that changing REG_INFTY is a bigger deal than what is done here.   Certainly, raising it is a good idea.  I changed it from I16 to U16 some releases back, and it was harder than this.  I believe the lower value is in our documentation.   But I think there are other gotchas involved.

regcomp.h
khwilliamson: Looking at a later commit, I see that this becomes unused.  But the name is still misleading.  And my question about combining the 2L and this remains.

regexec.c
```diff
@@ -3728,10 +3731,11 @@ Perl_regexec_flags(pTHX_ REGEXP * const rx, char *stringarg, char *strend,
              * Let @-, @+, $^N know */
             prog->lastparen = prog->lastcloseparen = 0;
             RXp_MATCH_UTF8_set(prog, utf8_target);
-            prog->offs[0].start = s - strbeg;
-            prog->offs[0].end = utf8_target
+            SSize_t match_start = s - strbeg;
+            SSize_t match_end = utf8_target
                 ? (char*)utf8_hop_forward((U8*)s, prog->minlenret, (U8 *) strend) - strbeg
                 : s - strbeg + prog->minlenret;
+            CLOSE_ANY_CAPTURE(prog, 0, match_start, match_end);
```
khwilliamson: This change is beyond what the commit message says, and I think should be in a separate commit with an explanation

regexec.c
```diff
@@ -328,11 +330,70 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
...
         PTR2UV(rex->offs),                                                  \
         (UV)(ix),                                                           \
-        (IV)(rex)->offs[ix].start,                                          \
-        (IV)(rex)->offs[ix].end,                                            \
+        (IV)(rex)->offs[ix].start_new,                                      \
+        (IV)(rex)->offs[ix].end_new,                                        \
         (UV)(rex)->lastparen                                                \
     ))
 
+#define CAPTURE_CLEAR(from_ix, to_ix, str)  STMT_START {                \
```
khwilliamson: I am unqualified to review changes to this area of the regex engine.  I have managed to avoid having to deal with these things, except minimally

regcomp.h
demerphq: I hadn't really thought about it. But yeah, probably you are right, we can merge 2L and 2 together, and use a union to have one argument be signed. I am not sure I want to deal with it until most of the important parts of the patch sequence are resolved. 

Related to this is that I would like to get rid of BRANCHJ and LONGJMP as they wont be needed anymore by the time this PR sequence is done.

regexec.c
demerphq: Hrm. Ill take a look.

regexec.c
demerphq: Well you can review the code for sanity. Right now that is much better than nothing. Also it doesnt help much if only me and Dave M know this code. What is the point of having code review if everybody says "I am not competent to review this?". 

Also, frankly i want to restructure how we store the paren offsets. I dont think using a single array of structs actually fits our needs very well. If it were stored as simple arrays it would be much more efficient to clear the paren buffers, or push them into the stack. To clear we really only need to set the ends to -1. (This is the kind of review that would be helpful.) 

We should organize the structure the operations of copying the paren data to the save stack, and clearing the parens (in two different ways) is efficient.  It seems to me if the paren data were organized lineararly, with starts, end, start_new, end_new and start_tmp all being separate chunks, then we would be able to clear the end_new and the end arrays independently with memset() (as we use -1 for the value), and we would be able to copy chunks of them into the save stack with memcpy.

regcomp.h
demerphq: If there are I didn't notice any beyond the tests making assumptions they can create strings that are REG_INFTY long. Ill have a look in the docs. Can you find the commit where you changed it last?

demerphq: FWIW, i found the patch where you did this. And it changes about as much as mine does.  See: 4288c5b93b3823465d8a11e569547afd41b5d1b8

I notice that there has been a lot of work related to REG_INFNTY done since then. I suspect my patch benefits from much of that. Anyway, i will push a fix to the comment. I couldnt find anything else but that.

regexec.c
demerphq: I dont get it. This is added in ddeba638cfb50b03af2a8d68f2573a2fb6c18e64 and is the *entire* reason for the patch.  This is the only place in the code which does a CLOSE_CAPTURE() like behavior which does not use CLOSE_CAPTURE(). Because of when it executes it cannot use CLOSE_CAPTURE() as it does not have some of debug data available to implement that. Thus this patch splits CLOSE_CAPTURE() into CLOSE_ANY_CAPTURE() and CLOSE_CAPTURE() as a wrapper.

The whole point is harden all the places in the code which close a capture and which might operate on offs[i].start and offs[i].end, which will be replaced in a follow up patch.

So I don't get it, and I don't agree. If you want I can add more to the commit message, but this line is the entire reason the code was changed. So that I could change the definition of CLOSE_ANY_CAPTURE() and fix all the places that try to close capture buffers.

demerphq: I have updated the commit message. Also I have pushed this commit as part of https://github.com/Perl/perl5/pull/20703. I think it would be better to provide any further feedback you have there.

demerphq:
Patches from this PR have been pushed independently as:
https://github.com/Perl/perl5/pull/20703
https://github.com/Perl/perl5/pull/20704
https://github.com/Perl/perl5/pull/20705
https://github.com/Perl/perl5/pull/20706
https://github.com/Perl/perl5/pull/20707

As they get merged I will rebase them away from this PR. For now they are all bunched together.

regcomp.h
demerphq: This patch was pushed as https://github.com/Perl/perl5/pull/20706

regcomp.h
demerphq: I pushed a patch to the end of the sequence which merges the regnode_2L and regnode_2 into a single struct, and introduces a consistent set of macros to access fields called "arg". With this patch we have a regnode_1, regnode_2, regnode_3, each of which can hold one more 32 bit argument as the previous. This storage is set up so it can store both U32 and I32 values, or two U16 values. Macros of the form ARG1i() ARG1u() and ARG1a() and ARG1b() can be used to access each arg field.  ANy other regnode types that also had arg1 or arg2 fields received the same treatment.

khwilliamson (approved):
This looks like a really nice enhancement.

The only thing I saw is that some of the values are cast, either redundantly, or inconsistently.  So
 `  arg= (U32)ARG1u(scan);
`
would be neater without the cast, and 
` utmp = (U32)ARG2i(scan);`

might be the right thing, or it might not be

demerphq:
> This looks like a really nice enhancement.
> 
> The only thing I saw is that some of the values are cast, either redundantly, or inconsistently. So ` arg= (U32)ARG1u(scan);` would be neater without the cast, and ` utmp = (U32)ARG2i(scan);`

The former can obviously be removed. The latter however I am not sure about. So on one hand, why not just use the struct api. On the other hand, we have cases where something is being stored as a signed value, and then read back as a signed value and cast to match an external variable, presumably because they can rule out the negative cases. It is not clear to me that we should change these or not.  You could argue that if the creator of the object thought the value was signed, or unsigned, then the fetcher should use the corresponding accessor, and the cast is actually a good thing. 

I cant decide really. Maybe in the end it doesn't matter much.

florian-pe:
@demerphq 
> WHAT I NEED FROM YOU
> Everything else in this PR is a minor fix and I don't need your help with. What I really need is your insight into how to exploit the state machine to do the above stuff efficiently, and not have to loop over all the capture buffers for instance to clear or commit them. Or any other suggestions you might have to resolve the above bugs. I am satisfied my PR has the correct semantics and consistent behavior, I just need advice on how to make it efficient.

There is a way to reset the capturing buffers in O(1) upon backtracking and avoid calling `memset()`. Instead of writing the start and end positions of a captured substring in a random access manner at their "final" index slot in the capture array, it would simpler to use the capture array using a stack discipline, sort of.

Besides the capture array base and capacity fields, it would require a `top` field and a `current` field. Both of them would need to be saved in a newly pushed choice point/backtracking frame.

OPEN could simply push a new capture struct on the catpure stack and store the start position in it as if every alternative was inside a branch reset group `(?| () | () )`. Additionally, OPEN would also write the real capture number in the newly pushed capture struct. The real capture number would be stored in the OPEN regop struct.

Upon backtracking, it would suffice to reset `top` and `current` to their previous values. Then new capturing groups could safely override previous ones.

At the end of a successful match, the array `@{^CAPTURE}` would be constructed with each capture group stored at their correct array indexes. Similarly with `$1`, `$2`, etc..

The description above is incomplete and awkward to fully put into words so here is the complete semantic description:
```
# working example
1 2 2 3 3   4 4 1  # capture group real indexes
( ( ) ( ) | ( ) )
0 1 1 2 2   1 1 0  # capture array indexes

# order in which capture structs are written (considering no backtracking occurs)
( 0 1 2
) 1 2 0

REGEX TRY INIT {
    top = -1;
    current = -1;
}

// The variable pos refers to the current position in the string being matched.

OPEN(n) {
    top++;
    capture[top].start = pos;
    capture[top].end = -1; // in order for (*ACCEPT) to know that the capturing group has not been closed yet
    capture[top].num = n;
    capture[top].prev_current = current;
    current = top;
}

CLOSE {
    capture[current].end = pos;
    current = capture[current].prev_current;
}

BACKTRACK {
    top = last_choice_point->top;
    current = last_choice_point->current;
}

ACCEPT { // (*ACCEPT)
    for (int i=0; i <= top; i++) {
        if (capture[i].end == -1) {
            capture[i].end = pos;
        }
    }
}

```

PR #20680, opened 2023-01-09
Makefile.SH - fix 'reonly' Makefile target to test ext/re/t/*.t properly

demerphq:
The .. in front of the ext/ required as the list is constructed relative to the t/ directory of the repo.

This also enables "full steam ahead" mode when parallel jobs are enabled. This target only tests a subset of our functionality, running in normal mode and separating core tests from ext/ tests just slows things down for no value.


khwilliamson (approved):

jkeenan:
LGTM.  This will enhance the utility of this `make` target.

PR #20681, opened 2023-01-09
regcomp.pl - fixup intflags debug data to handle gaps properly

demerphq:
We were not handling gaps in the sequence properly, and effectively showing the wrong flag names or missing the last flag. Now we die if there are any collisions or if any of the PREGf defines set more than one bit. This also adds some crude tests to validate that intflags serialization is working properly.

Note, extflags handles more complex scenarios and seems to handle this gracefully already, hence the reason I haven't touched it as well.

### DIFF
--- a/ext/re/t/intflags.t
+++ b/ext/re/t/intflags.t
@@ -1,0 +1,25 @@
+#!./perl
+
+BEGIN {
+    require Config;
+    if (($Config::Config{'extensions'} !~ /\bre\b/) ){
+	print "1..0 # Skip -- Perl configured without re module\n";
+	exit 0;
+    }
+}
+
+use strict;
+
+# must use a BEGIN or the prototypes wont be respected meaning
+# tests could pass that shouldn't.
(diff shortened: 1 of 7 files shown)


ext/re/t/intflags.t
```diff
@@ -0,0 +1,25 @@
...
+    if (($Config::Config{'extensions'} !~ /\bre\b/) ){
+	print "1..0 # Skip -- Perl configured without re module\n";
+	exit 0;
+    }
+}
+
+use strict;
+
+# must use a BEGIN or the prototypes wont be respected meaning
+    # tests could pass that shouldn't
```
khwilliamson: Is there some reason this is indented?

khwilliamson (approved):

ext/re/t/intflags.t
demerphq: It was indented like that in the file I cribbed it from lexical_debug.t, i fixed it in both and repushed.

PR #20682, opened 2023-01-10
Provide '-Ilib' to @INC in programs called by regen.pl

jkeenan:
In the course of attempting correction of a one-character typo in documentation in lib/feature.pm (a generated file), an inadequacy in regen.pl was discovered.  The programs which regen.pl invokes need to be provided with '-Ilib' in their invocations to be able to locate, e.g., strict.pm.

Modification of lib/feature.pm's docs also required increment to its $VERSION inside regen/feature.pl.

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -4,7 +4,7 @@
 # Any changes made here will be lost!
 
 package feature;
-our $VERSION = '1.78';
+our $VERSION = '1.79';
 
 our %feature = (
     fc                      => 'feature_fc',
@@ -366,7 +366,7 @@ warn when you use the feature, unless you have explicitly disabled the warning:
     no warnings "experimental::declared_refs";
 
 This allows a reference to a variable to be declared with C<my>, C<state>,
-our C<our>, or localized with C<local>.  It is intended mainly for use in
+or C<our>, or localized with C<local>.  It is intended mainly for use in
 conjunction with the "refaliasing" feature.  See L<perlref/Declaring a
 Reference to a Variable> for examples.
 
--- a/regen.pl
+++ b/regen.pl
@@ -15,7 +15,7 @@ use strict;
 
 my $tap = $ARGV[0] && $ARGV[0] eq '--tap' ? '# ' : '';
 foreach my $pl (map {chomp; "regen/$_"} <DATA>) {
-  my @command =  ($^X, '-I.', $pl, @ARGV);
+  my @command =  ($^X, '-I.', '-Ilib', $pl, @ARGV);
   print "$tap@command\n";
   system @command
     and die "@command failed: $?" 
--- a/regen/feature.pl
+++ b/regen/feature.pl
@@ -500,7 +500,7 @@ read_only_bottom_close_and_rename($h);
 
 __END__
 package feature;
-our $VERSION = '1.78';
+our $VERSION = '1.79';
 
 FEATURES
 
@@ -788,7 +788,7 @@ warn when you use the feature, unless you have explicitly disabled the warning:
     no warnings "experimental::declared_refs";
 
 This allows a reference to a variable to be declared with C<my>, C<state>,
-our C<our>, or localized with C<local>.  It is intended mainly for use in
+or C<our>, or localized with C<local>.  It is intended mainly for use in
 conjunction with the "refaliasing" feature.  See L<perlref/Declaring a
 Reference to a Variable> for examples.
 


jkeenan:
This is, FBOW, simultaneously a pull request and a bug report.

Working in a checkout of blead I called `perldoc lib/feature.pm` and spotted a one-character typo.  When I went to correct that, at the top of `lib/feature.pm` I read:
```
# -*- buffer-read-only: t -*-
# !!!!!!!   DO NOT EDIT THIS FILE   !!!!!!!
# This file is built by regen/feature.pl.
# Any changes made here will be lost!
```
So I made the change in `regen/feature.pl` instead.  When I ran `make test_porting` I got this:
```
$ make -j1 test_porting
...
'lib/feature.pm-new' and 'lib/feature.pm' differ
/usr/home/jkeenan/gitwork/perl/perl -I. regen/feature.pl --tap failed: 65280 at regen.pl line 20, <DATA> line 11.
    Hint:  A failure in this file can often be corrected by running:
     ./perl -Ilib regen.pl
porting/regen.t ........... 
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 24/44 subtests 

Test Summary Report
-------------------
porting/regen.t         (Wstat: 65280 (exited 255) Tests: 21 Failed: 1)
  Failed test:  21
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 44 tests but ran 21.
Files=33, Tests=43502, 23 wallclock secs ( 2.38 usr  0.20 sys + 57.84 cusr  2.40 csys = 62.81 CPU)
Result: FAIL
Finished test run at Tue Jan 10 15:01:50 2023.
*** [test_porting] Error code 1

1 error

```
Okay, let's try the command suggested in the Hint:
```
$ ./perl -Ilib regen.pl
/usr/home/jkeenan/gitwork/perl/perl -I. regen/embed.pl
Can't locate strict.pm in @INC (you may need to install the strict module) (@INC entries checked: . /usr/local/lib/perl5/site_perl/5.37.8/amd64-freebsd-thread-multi /usr/local/lib/perl5/site_perl/5.37.8 /usr/local/lib/perl5/5.37.8/amd64-freebsd-thread-multi /usr/local/lib/perl5/5.37.8) at regen/embed.pl line 23.
BEGIN failed--compilation aborted at regen/embed.pl line 23.
/usr/home/jkeenan/gitwork/perl/perl -I. regen/embed.pl failed: 512 at regen.pl line 20, <DATA> line 11.
```
This compilation failure surprised me at first.  If (i) I'm running a newly built `perl` from the top-level directory, (ii) adding `./lib` to `@INC`, and (iii) there exists `lib/strict.pm`, then why is the program failing to locate `strict.pm`?

However, the (partial?) explanation comes in the next line of output above.  `regen.pl` calls several programs from within itself, but adds only the current working directory (`-I.`) to the invocation; it does **not** add the directory one-level down (`-Ilib`) where `strict.pm` is found.

So I modified `regen.pl` to include `-Ilib` in `@INC`.  I believe this is safe -- `make test_porting` is passing -- but review is needed.

(Along the way I also discovered that I have to increment feature.pm's version by bumping `$VERSION` inside `regen/feature.pl`.)

demerphq:
I dont think this is the correct fix to be honest. The output from the porting script is just wrong. You typically aren't supposed to run regen scripts with `./perl`. Only the test suite does that and that is for testing. You are supposed to use a preinstalled perl, and then run it like any other perl script `perl regen.pl` and even that is kinda wrong, the correct thing to do is run`'make regen`.  

If you look at the Makefile for the 'regen' target you will see this. When these scripts are called from an uninstalled perl they need the `-Ilib`, and when it is run like that we ensure that an -I parameter is provided, but adding it will-nily like that will break the script when it is run as it is meant to be as it will use the uninstalled lib, which will break binary compatibility, and not the preinstalled lib.

Honestly I don't think you should try to fix this by altering regen.pl unless you want to mess-about detecting whether it is running in an uninstalled perl or not, instead just change the test output to show the correct commands.  The use of an *installed* (and presumably tested) perl to run regen is a deliberate bootstrapping measure, if the perl you run regen with is broken then it will just compound the problems that you have in your uninstalled perl using its regen output.

jkeenan:
> I dont think this is the correct fix to be honest. The output from the porting script is just wrong. You typically aren't supposed to run regen scripts with `./perl`. Only the test suite does that and that is for testing. You are supposed to use a preinstalled perl, and then run it like any other perl script `perl regen.pl` and even that is kinda wrong, the correct thing to do is run`'make regen`.
> 

I'll try something along the lines you are suggesting later today.

jkeenan:
Superseded by https://github.com/Perl/perl5/pull/20683, which has been merged into blead.

PR #20683, opened 2023-01-10
Hint make regen 20230110

jkeenan:


jkeenan:
This p.r. is proposed to supersede https://github.com/Perl/perl5/pull/20682. @demerphq

@atoomic, @khwilliamson, the original Hint was committed to blead here:
```
commit 1fb364e9e8c8d678834e5555a02711e6d8be5524
Author:     Nicolas R <atoomic@cpan.org>
AuthorDate: Mon Aug 3 14:26:55 2020 -0500
CommitDate: Mon Aug 10 22:29:43 2020 -0600

    Add a hint to t/porting/regen.t
    
    Also use 'lib' instead of '.'
    when running the regen script with tap output
    abort on the first failure so we advertise
    the command to run to fix the issue.

diff --git a/regen/regen_lib.pl b/regen/regen_lib.pl
index cbe51eda71..82c5c7a8e7 100644
--- a/regen/regen_lib.pl
+++ b/regen/regen_lib.pl
@@ -96,7 +96,7 @@ sub close_and_rename {
         }
         if ($fail) {
             print STDOUT "not ok - $0 $final_name\n";
-            print STDERR "$fail\n";
+            die "$fail\n";
         } else {
             print STDOUT "ok - $0 $final_name\n";
         }
@@ -204,6 +204,8 @@ sub read_only_bottom_close_and_rename {
     print $fh "\n$comment\n";
 
     close_and_rename($fh);
+
+    return;
 }
 
 sub tab {
diff --git a/t/porting/regen.t b/t/porting/regen.t
index 2c3fa21998..5c57967222 100644
--- a/t/porting/regen.t
+++ b/t/porting/regen.t
@@ -105,7 +105,12 @@ OUTER: foreach my $file (@files) {
 }
 
 foreach (@progs) {
-    my $command = "$^X -I. $_ --tap";
+    my $args = qq[-Ilib $_ --tap];
+    diag("./perl $args");
+    my $command = "$^X $args";
     system $command
-        and die "Failed to run $command: $?";
+        and die <<~"HINT";
+    Hint:  A failure in this file can often be corrected by running:
+     ./perl -Ilib $_
+HINT
 }
```
Would you like to comment?

t/porting/regen.t
```diff
@@ -111,6 +111,6 @@ foreach (@progs) {
     system $command
         and die <<~"HINT";
```
JRaspass: Now we're not interpolating `$_` we should single quote the heredoc.

khwilliamson (approved):

t/porting/regen.t
demerphq: agree

jkeenan: Fixed in 1044e25ef2; please re-review.

demerphq (approved):

JRaspass (approved):

PR #20684, opened 2023-01-11
regexec engine - wrap and replace RX_OFFS() with better abstractions

demerphq:
RX_OFFS() exposes a bit too much about how capture buffers are represented. This adds RX_OFFS_START() and RX_OFFS_END() and RX_OFFS_VALID() to replace most of the uses of the RX_OFFS() macro or direct access to the rx->off[] array. (We add RX_OFFSp() for those rare cases that should have direct access to the array.) This allows us to replace this logic with more complicated macros in the future. Pretty much anything using RX_OFFS() is going to be broken by future changes, so changing the define allows us to track it down easily.

Not all use of the rx->offs[] array are converted; some uses are required for the regex engine internals, but anything outside of the regex engine should be using the replacement macros, and most things in the regex internals should use it also.

### DIFF
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -3189,7 +3189,7 @@ PP(pp_match)
     if (global && (gimme != G_LIST || (dynpm->op_pmflags & PMf_CONTINUE))) {
         if (!mg)
             mg = sv_magicext_mglob(TARG);
-        MgBYTEPOS_set(mg, TARG, truebase, RXp_OFFS(prog)[0].end);
+        MgBYTEPOS_set(mg, TARG, truebase, RXp_OFFS_END(prog,0));
         if (RXp_ZERO_LEN(prog))
             mg->mg_flags |= MGf_MINMATCH;
         else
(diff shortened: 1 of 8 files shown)


pp_hot.c
```diff
@@ -3211,20 +3211,16 @@ PP(pp_match)
...
             {
-                const I32 len = RXp_OFFS(prog)[i].end - RXp_OFFS(prog)[i].start;
-                const char * const s = RXp_OFFS(prog)[i].start + truebase;
-                if (UNLIKELY(  RXp_OFFS(prog)[i].end   < 0
-                            || RXp_OFFS(prog)[i].start < 0
-                            || len < 0
-                            || len > strend - s)
+                const I32 len = RXp_OFFS_END(prog,i) - RXp_OFFS_START(prog,i);
+                const char * const s = RXp_OFFS_START(prog,i) + truebase;
+                if ( UNLIKELY( len < 0 || len > strend - s)
```
leonerd, line 3218: This condition seems to have lost the logic relating to `start/end < 0`. Maybe it wanted a `RXp_OFFS_VALID()` test as well?

demerphq: Its validated just above, at line 3214.

demerphq: The previous code validated twice. :-)

leonerd: Ohright, yes there it is. :)

leonerd (approved):
LGTM

pp_hot.c
demerphq: In a follow up code ill simplify that to extract the start offset and end offset just once for that code. its a bit repetitive.

PR #20686, opened 2023-01-11
Replace FreeBSD URL's with new HTTPS ones

rwp0:
Replace the old FreeBSD bugs HTTP URL's with the newer BugZilla HTTPS ones
in `README.freebsd` file as link, and `caretx.c` file as comment.

It's also consistent with other distribution files linking to FreeBSD issues.


leonerd (approved):
LGTM

PR #20688, opened 2023-01-12
Syntax highlight configpm

rwp0:
For enhancing the source readability in GitHub:

cf. https://raw.githubusercontent.com/Perl/perl5/blead/PACKAGING

and the various README files where `# vim: syntax=` was added for a similar purpose.

### DIFF
--- a/configpm
+++ b/configpm
@@ -1,4 +1,5 @@
 #!./miniperl -w
+# vim: syntax=perl
 #
 # configpm
 #


rwp0:
# Contrast

## Current
https://github.com/Perl/perl5/blob/fe5492d916201ce31a107839a36bcb1435fe7bf0/configpm

## Highlighted
https://github.com/Perl/perl5/blob/33c8167aa54a48d90b07d8c848774458e8c1874d/configpm

demerphq (approved):

demerphq:
thanks, applied.

PR #20689, opened 2023-01-13
bisect-runner.pl: When did warnings clear up

jkeenan:
Illustrate bisection to identify commit at which run-time warnings ceased being emitted from a program.

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -1094,6 +1094,59 @@ L<GH issue 20261|https://github.com/Perl/perl5/issues/20261>
 
 =back
 
+=head2 When did perl stop emitting warnings when running on certain code?
+
+=over 4
+
+=item * Background
+
+Most of the time, we bisect in order to identify the first "bad" commit:  the
+first time code failed to compile; the first time the code emitted warnings;
+and so forth.
+
+Some times, however, we want to identify the first "good" commit:  the point
+where the code began to compile; the point where the code no longer emitted
+warnings; etc.
+
+We can use this program for that purpose, but we have to reverse our sense of
+"good" and "bad" commits.  In this case, F<git> will label the healing commit
+as "bad" (or label a breaking commit as "good"), so we have to invoke the
+program such that we do a C<die> when we reach the healing commit.
+
+=item * Problem
+
+It was reported that in an older version of Perl, a warning was being emitted
+when a program was using the F<bigrat> module and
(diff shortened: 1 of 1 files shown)


tonycoz (approved):

Porting/bisect-runner.pl
```diff
@@ -1094,6 +1094,59 @@ =head2 When did perl stop segfaulting on certain code?
...
+and so forth.
+
+Some times, however, we want to identify the first "good" commit:  the point
+where the code began to compile; the point where the code no longer emitted
+warnings; etc.
+
+We can use this program for that purpose, but we have to reverse our sense of
+"good" and "bad" commits.  In this case, F<git> will label the healing commit
+as "bad" (or label a breaking commit as "good"), so we have to invoke the
+program such that we do a C<die> when we reach the healing commit.
```
hvds, line 1114: The usual way to "reverse our sense of good and bad commits" is with the `--expect-fail` option. While the example you add below is not wrong, I think it risks making things more confusing.

I would be inclined to rewrite the example to invert the sense of the `die()` and add `--expect-fail`, then work out how the surrounding text would have to change to accommodate the new code, then step back and consider whether this entire new section is actually resulting in an improvement of the documentation.

In particular, consider what the documentation would have needed to say for you to consider that `--expect-fail` - which is highlighted at lines 7-8 of this 829-line document (as rendered) - coiuld be a possible solution here.

Maybe the answer should be to rename `--expect-fail`, or even to replace `bisect` with two programs such as `bisect-what-fixed` and `bisect-what-broke`.

jkeenan: > The usual way to "reverse our sense of good and bad commits" is with the `--expect-fail` option. While the example you add below is not wrong, I think it risks making things more confusing.
> 
> I would be inclined to rewrite the example to invert the sense of the `die()` and add `--expect-fail`, then work out how the surrounding text would have to change to accommodate the new code, then step back and consider whether this entire new section is actually resulting in an improvement of the documentation.
> 
> In particular, consider what the documentation would have needed to say for you to consider that `--expect-fail` - which is highlighted at lines 7-8 of this 829-line document (as rendered) - coiuld be a possible solution here.
> 
> Maybe the answer should be to rename `--expect-fail`, or even to replace `bisect` with two programs such as `bisect-what-fixed` and `bisect-what-broke`.

@hv, since this pull request has been merged and since there's likely to be some back-and-forth on this, I'd like to move this discussion to a fresh ticket.  Please see https://github.com/Perl/perl5/issues/20724.

Issue #20690, opened 2023-01-13
regexp quantifier `{0}` causes wrong substitution against UTF-8 on string


agguser:
```sh
$ echo "abc" | perl -lne '
        print "string: ", "abc" =~ s/b{0}//gr;
        print "stdin : ", s/b{0}//gr;
        print "{0}?  : ", s/b{0}?//gr;
        print "no g  : ", s/b{0}//r;
        print "first : ", s/a{0}//r;
        '
string: abc
stdin : ac     # expect: abc
{0}?  : abc
no g  : abc
first : bc     # expect: abc
```

<details>
<summary>

```sh
$ perl -V
```
</summary>

```
Summary of my perl5 (revision 5 version 36 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=5.12.15-arch1-1
    archname=x86_64-linux-thread-multi
    uname='archlinux'
    config_args='-des -Dusethreads -Duseshrplib -Doptimize=-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -flto=auto -Dprefix=/usr -Dvendorprefix=/usr -Dprivlib=/usr/share/perl5/core_perl -Darchlib=/usr/lib/perl5/5.36/core_perl -Dsitelib=/usr/share/perl5/site_perl -Dsitearch=/usr/lib/perl5/5.36/site_perl -Dvendorlib=/usr/share/perl5/vendor_perl -Dvendorarch=/usr/lib/perl5/5.36/vendor_perl -Dscriptdir=/usr/bin/core_perl -Dsitescript=/usr/bin/site_perl -Dvendorscript=/usr/bin/vendor_perl -Dinc_version_list=none -Dman1ext=1perl -Dman3ext=3perl -Dcccdlflags='-fPIC' -Dlddlflags=-shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -Dldflags=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
[...]

PR #20694, opened 2023-01-13
regexec.c - avoid calling regrepeat when the max is 0

demerphq:
When we have a max quantifier of 0, then the quantified item is essentially a NOTHING reop. Regardless, we do not need to call regrepeat, and doing so confuses some of the logic it contains. Simply avoiding calling regrepeat() fixes the underlying issue, and avoids the broken code.

This fixes GH Issue #20690.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -9316,7 +9316,10 @@ NULL
                 /* avoid taking address of locinput, so it can remain
                  * a register var */
                 char *li = locinput;
-                ST.count = regrepeat(rex, &li, ST.A, loceol, reginfo, ST.max);
+                if (ST.max)
+                    ST.count = regrepeat(rex, &li, ST.A, loceol, reginfo, ST.max);
+                else
+                    ST.count = 0;
                 if (ST.count < ST.min)
                     sayNO;
                 SET_locinput(li);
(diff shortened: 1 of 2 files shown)


ilmari (approved):

ilmari (commented):
How about other callers? Would it make sense to make `regrepeat` be a no-op if the max count is zero?

demerphq:
I reviewed them, and as far as I could tell only this case calls it with a possible max of 0, all the other cases call it with a max of 1 as a constant, or guard the call for non-zero counts.

demerphq:
@ilmari - I repushed with an assert() on max so we will notice if anything does call it with a zero max. Makes me wonder if embed.fnc should have an NZ parameter similar to NN to do this automagically.

ilmari:
I think the `assert(max)` is enough for now. The `NZ` parameter might be worth it if there's lots, but that can be a separate patch.

demerphq:
I have a patch to embed.fnc and etc. But yeah, ill do it in a separate patch.

jkeenan:
> When we have a max quantifier of 0, then the quantified item is essentially a NOTHING reop. Regardless, we do not need to call regrepeat, and doing so confuses some of the logic it contains. Simply avoiding calling regrepeat() fixes the underlying issue, and avoids the broken code.
> 
> This fixes GH Issue #20690.

+1

PR #20696, opened 2023-01-13
embed.pl - add a way to declare a parameter should be non-zero

demerphq:
This autogenerates the required asserts to validate a parameter is non-zero. It uses it to replace the check in regrepeat() as a first example.


richardleach:
LGTM

khwilliamson (approved):

PR #20697, opened 2023-01-13
get_{av,hv} should ignore SVf_UTF8 when testing if flags == 0

leonerd:
Otherwise, if you call with simply `SVf_UTF8` on a name that does not exist, it will try to call `GvAVn()` or `GvHVn()` on a NULL value and segfault.

(See also https://github.com/Perl/perl5/pull/20647#issuecomment-1381252416)

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -2845,7 +2845,8 @@ Returns the AV of the specified Perl global or package array with the given
 name (so it won't work on lexical variables).  C<flags> are passed
 to C<gv_fetchpv>.  If C<GV_ADD> is set and the
 Perl variable does not exist then it will be created.  If C<flags> is zero
-and the variable does not exist then NULL is returned.
+(ignoring C<SVf_UTF8>) and the variable does not exist then C<NULL> is
+returned.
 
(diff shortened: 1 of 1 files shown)


perl.c
```diff
@@ -2859,7 +2860,7 @@ Perl_get_av(pTHX_ const char *name, I32 flags)
 
     PERL_ARGS_ASSERT_GET_AV;
 
-    if (flags)
+    if (flags & ~SVf_UTF8)
```
demerphq, line 2863: Maybe add a comment that this is logically the same as  `(flags && flags != SVf_UTF8)`

demerphq (approved):
Other than the point about the comment this looks fine.

Leont (approved):

perl.c
leonerd: I could do but I think I'd find that style more confusing than currently :)

demerphq: I don't understand how adding a *comment* that the two are equivalent could be more confusing than not having the comment. A lot of people dont think bit flips properly. But whatever, ill do a follow up PR.

PR #20701, opened 2023-01-14
Override *.h files as C with Linguist

rwp0:
GitHub classifies 23 files as C++ for some reason.

https://github.com/Perl/perl5/search?q=language%3AC%2B%2B&type=code

I believe Perl doesn't contain C++ code, and C++ header files can have distinguishable `.hh`, `.hpp`, `.hxx`, and `.h++` [extensions](https://en.wikipedia.org/wiki/C%2B%2B) in addition.

image

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,3 +21,4 @@ warnings.h linguist-generated
 *.pl linguist-language=Perl
 *.pm linguist-language=Perl
 *.t linguist-language=Perl
+*.h linguist-language=C


rwp0:
## References

### Linguist C++ extensions list

https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L741

### Linguist overrides documentation

https://github.com/github/linguist/blob/master/docs/overrides.md

xenu:
>I believe Perl doesn't contain C++ code, and C++ header files can have distinguishable .hh, .hpp, .hxx, and .h++ [extensions](https://en.wikipedia.org/wiki/C%2B%2B) in addition.

99% of our code is written in C, but we do have a few C++ files: `win32/perllib.c`, `win32/perlhost.h`, `win32/vdir.h` and `win32/vmem.h`.

I guess it would make sense to rename them because it's confusing.

PR #20702, opened 2023-01-14
Porting/updateAUTHORS.pl: Suggested `git config` command is wrong

bsmith:
When you run tests in a repository that has local modifications `t/porting/authors.t` checks if your git is correctly configured with your identity.  However, the suggested commands don't work, because `git config` doesn't need a `--set` flag to set options.

This patch removes the `--set` from the suggested commands which is not necessary to set the variables in the local git repository. For comparison, if you run `git commit` without your identity set up, it suggests:

```
Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.
```

### DIFF
--- a/Porting/updateAUTHORS.pl
+++ b/Porting/updateAUTHORS.pl
@@ -285,8 +285,8 @@ sub error_advice_for_uncommitted_changes {
     else {
         $extra .=
               "\nYou do not have any git user config set up, consider using\n\n"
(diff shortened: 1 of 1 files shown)


demerphq (approved):

demerphq:
Good catch, thanks.

PR #20711, opened 2023-01-15
regcomp.c - add optimistic eval (*{ ... }) and (**{ ... })

demerphq:
This adds (*{ ... }) and (**{ ... }) as equivalents to (?{ ... }) and (??{ ... }). The only difference being that the star variants are "optimisitic" and are defined to never disable optimisations. This is especially relevant now that use of (?{ ... }) prevents important optimisations anywhere in the pattern, instead of the older and inconsistent rules where it only affected the parts that contained the EVAL.

It is also very useful for injecting debugging style expressions to the pattern to understand what the regex engine is actually doing. The older style (?{ ... }) variants would change the regex engines behavior, meaning this was not as effective a tool as it could have been.

Similarly it is now possible to test that a given regex optimisation works correctly using (*{ ... }), which was not possible with (?{ ... }).

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -1902,7 +1902,7 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
             !sawlookahead &&
             (OP(first) == STAR &&
             REGNODE_TYPE(OP(REGNODE_AFTER(first))) == REG_ANY) &&
-            !(RExC_rx->intflags & PREGf_ANCH) && !pRExC_state->code_blocks)
+            !(RExC_rx->intflags & PREGf_ANCH) && !(RExC_seen & REG_PESSIMIZE_SEEN))
         {
             /* turn .* into ^.* with an implied $*=1 */
             const int type =
@@ -1915,7 +1915,7 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
         }
         if (sawplus && !sawminmod && !sawlookahead
             && (!sawopen || !RExC_sawback)
-            && !pRExC_state->code_blocks) /* May examine pos and $& */
+            && !(RExC_seen & REG_PESSIMIZE_SEEN)) /* May examine pos and $& */
             /* x+ must match at the 1st pos of run of x's */
             RExC_rx->intflags |= PREGf_SKIP;
 
@@ -2167,20 +2167,27 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
     }
     if (RExC_seen & REG_GPOS_SEEN)
         RExC_rx->intflags |= PREGf_GPOS_SEEN;
+
+    if (RExC_seen & REG_PESSIMIZE_SEEN)
+        RExC_rx->intflags |= PREGf_PESSIMIZE_SEEN;
+
     if (RExC_seen & REG_LOOKBEHIND_SEEN)
         RExC_rx->extflags |= RXf_NO_INPLACE_SUBST; /* inplace might break the
                                                 lookbehind */
     if (pRExC_state->code_blocks)
         RExC_rx->extflags |= RXf_EVAL_SEEN;
-    if (RExC_seen & REG_VERBARG_SEEN)
-    {
+
+    if (RExC_seen & REG_VERBARG_SEEN) {
         RExC_rx->intflags |= PREGf_VERBARG_SEEN;
         RExC_rx->extflags |= RXf_NO_INPLACE_SUBST; /* don't understand this! Yves */
     }
+
     if (RExC_seen & REG_CUTGROUP_SEEN)
         RExC_rx->intflags |= PREGf_CUTGROUP_SEEN;
+
     if (pm_flags & PMf_USE_RE_EVAL)
         RExC_rx->intflags |= PREGf_USE_RE_EVAL;
+
     if (RExC_paren_names)
         RXp_PAREN_NAMES(RExC_rx) = MUTABLE_HV(SvREFCNT_inc(RExC_paren_names));
     else
@@ -2944,6 +2951,7 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
     I32 num; /* numeric backreferences */
     SV * max_open;  /* Max number of unclosed parens */
     I32 was_in_lookaround = RExC_in_lookaround;
+    I32 fake_eval = 0; /* matches paren */
 
     /* The difference between the following variables can be seen with  *
      * the broken pattern /(?:foo/ where segment_parse_start will point *
@@ -3000,6 +3008,16 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
             goto parse_rest;
         }
         else if ( *RExC_parse == '*') { /* (*VERB:ARG), (*construct:...) */
+            if (RExC_parse[1] == '{') {
+                fake_eval = '{';
+                goto handle_qmark;
+            }
+            else
+            if ( RExC_parse[1] == '*' && RExC_parse[2] == '{' ) {
+                fake_eval = '?';
+                goto handle_qmark;
+            }
+
             char *start_verb = RExC_parse + 1;
             STRLEN verb_len;
             char *start_arg = NULL;
@@ -3310,7 +3328,9 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
             return ret;
         }
         else if (*RExC_parse == '?') { /* (?...) */
-            bool is_logical = 0;
+          handle_qmark:
+            ; /* make sure the label has a statement associated with it*/
+            bool is_logical = 0, is_optimistic = 0;
             const char * const seqstart = RExC_parse;
             const char * endptr;
             const char non_existent_group_msg[]
@@ -3323,8 +3343,14 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
             }
 
             RExC_parse_inc_by(1);   /* past the '?' */
-            paren = *RExC_parse;    /* might be a trailing NUL, if not
-                                       well-formed */
+            if (!fake_eval) {
+                paren = *RExC_parse;    /* might be a trailing NUL, if not
+                                           well-formed */
+                is_optimistic = 0;
+            } else {
+                is_optimistic = 1;
+                paren = fake_eval;
+            }
             RExC_parse_inc();
             if (RExC_parse > RExC_end) {
                 paren = '\0';
@@ -3705,10 +3731,13 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
                 }
                 pRExC_state->code_index++;
                 nextchar(pRExC_state);
+                if (!is_optimistic)
+                    RExC_seen |= REG_PESSIMIZE_SEEN;
 
                 if (is_logical) {
                     regnode_offset eval;
                     ret = reg_node(pRExC_state, LOGICAL);
+                    FLAGS(REGNODE_p(ret)) = 2;
 
                     eval = reg2Lanode(pRExC_state, EVAL,
                                        n,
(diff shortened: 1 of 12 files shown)


regcomp.c
```diff
@@ -3310,7 +3328,9 @@ S_reg(pTHX_ RExC_state_t *pRExC_state, I32 paren, I32 *flagp, U32 depth)
             return ret;
         }
         else if (*RExC_parse == '?') { /* (?...) */
-            bool is_logical = 0;
+            bool is_logical, is_optimistic;
+          handle_qmark:
+            is_logical = 0;
```
tonycoz: Why not put the label above the bool declarations so is_logical can be initialized in the same step?

You will need a `;` after it to give the label a statement, but I think it would be better.

The declaration for `is_optimistic` could be moved closer to first use too (immediately before the switch, I think).

Pretty much all of the declarations at the top of the block could be moved down (or out), that's outside the scope of the PR but it would let you avoid adding that semi-colon to the label.  But that's probably out of scope for this PR.

demerphq: Thanks. I have changed it as you suggested. I didn't know i could "just" stick a semicolon in there. :-) I pushed what i think is what you wanted me to do.

demerphq:
Folks in a previous patch in order to fix various bugs I had to disable optimizations in a regex whenever there is an EVAL group in the pattern *anywhere*, and not in just a localized areas as in earlier perls.  

Thus I really want to see this merged so that people have an alternative that does not alter the performance of the pattern. It has also already proved very useful in testing the regex engine. 

So unless i hear an objection very soon I will merge this unilaterally.

regcomp.c
tonycoz: I can see some other possible changes there, but they're unrelated to this PR.

tonycoz (approved):

regcomp.c
demerphq: Please let me know what you suggest so I can queue up the second pr.

iabyn:
On Wed, Jan 18, 2023 at 04:06:26AM -0800, Yves Orton wrote:
> Folks in a previous patch in order to fix various bugs I had to disable optimizations in a regex whenever there is an EVAL group in the pattern *anywhere*, and not in just a localized areas as in earlier patches.  
> 
> Thus I really want to see this merged so that people have an alternative that does not alter the performance of the pattern. It has also already proved very useful in testing the regex engine. 

I don't understand this. I thought the reason why certain optimisations
were disabled in the presence of an EVAL node was that its presence
invalidated the assumptions which made the optimisation possible.

For example with the super-linear cache, it assumes that if a pattern like
/A*B/ matches A at position p in the string after n iterations of
'*' but then continues on and fails B, then it will *always* fail at p,
regardless of n. So note any previous failure at point p, and if after
backtracking we reach the same WHILEM node while the string is at position
p again, fail immediately, without executing B.

This assumption doesn't hold for certain types of thing in B, such a \1 or
(??{...}), so the super-linear cache is disabled in such cases.

-- 
This email is confidential, and now that you have read it you are legally
obliged to shoot yourself. Or shoot a lawyer, if you prefer. If you have
received this email in error, place it in its original wrapping and return
for a full refund. By opening this email, you accept that Elvis lives.

demerphq:
On Tue, 24 Jan 2023 at 18:15, iabyn ***@***.***> wrote:

> On Wed, Jan 18, 2023 at 04:06:26AM -0800, Yves Orton wrote:
> > Folks in a previous patch in order to fix various bugs I had to disable
> optimizations in a regex whenever there is an EVAL group in the pattern
> *anywhere*, and not in just a localized areas as in earlier patches.
> >
> > Thus I really want to see this merged so that people have an alternative
> that does not alter the performance of the pattern. It has also already
> proved very useful in testing the regex engine.
>
> I don't understand this. I thought the reason why certain optimisations
> were disabled in the presence of an EVAL node was that its presence
> invalidated the assumptions which made the optimisation possible.
>

Well, that might apply to (??{ ... }), but it can't explain disabling
optimizations for (?{ ... }), which shouldnt affect the match process,
except when used in a

(?(?{ ...})yes|no)

pattern, and even then the eval shouldn't change how the pattern matches
anymore than it would if it were

(?(1)yes|no).

which now also disables some optimisations. Maybe not enough actually given
what you have pointed out in this mail.

For example with the super-linear cache, it assumes that if a pattern like
> /A*B/ matches A at position p in the string after n iterations of
> '*' but then continues on and fails B, then it will *always* fail at p,
> regardless of n. So note any previous failure at point p, and if after
> backtracking we reach the same WHILEM node while the string is at position
> p again, fail immediately, without executing B.
>
> This assumption doesn't hold for certain types of thing in B, such a \1 or
> (??{...}), so the super-linear cache is disabled in such cases.
>

I don't mind removing (**{ ... }) given your feedback here. But I think (*{
... }) should be here to stay.

I think over time there has been a mashup of functionality and
expectations. both (?{ ... }) and (??{ ... }) are implemented via an EVAL,
and both disable optimisations.

Personally i view (??{ ... }) as a relic that we would ditch if only we
could (but we can't so I am not arguing we should), its primary utility was
in solving problems that I think most people these days would use pattern
recursion to solve, IOW, these days most older uses of EVAL can be replaced
by GOSUB and people do not use it that much as the alternative is much
superior in multiple ways.

But (?{ ... }) is a different beast. Its primary usefulness is
understanding what the regex does, or for utilizing the backtracking engine
to solve certain string decomposition problems where you are more
interested in forcing the regex engine to backtrack than you are in the
final result of the pattern. Sure there is a tiny sliver of special purpose
utility in that it sets $^R (how many people use that really?), and it is
also potentially useful for solving some non-regular pattern match cases
like /(.*)(?(?{ length($1) % 4 != 0 })(*FAIL))/ but id say the main
practical reason people use it is to a) debug their patterns, and b)
exploit the backtracking nature of the regex engine to do string
decomposition problem solving where you are probably forcing the pattern to
fail at the end anyway.

Both of these aspects of its usefulness are broken by optimisations but in
opposite ways. Using it to debug your pattern isn't that helpful when using
it changes how the pattern is executed by disabling optimizations, and
optimisations get in the way of using it effectively as a string
decomposition tool.  There is an inherent tension between those two uses.

So since (?{ ... }) disabled many optimisations I introduced (*{ ...}) to
be the "optimistic" version. I am also minded to make (?{ ... }) disable
even more optimisations in the future, so that for instance the substring
optimisation is disabled by its presence, as the substring optimisation has
to be actively worked around if you want to use the regex engine for string
decomposition.

Consider these two tests:

    {
        my $message = 'Match is quadratic due to eval; See Bug 22395';
        our $count;
        for my $l (10, 100, 1000) {
            $count = 0;
            ('a' x $l) =~ /(.*)(?{ $count++ })[bc]/;
            is($count, $l*($l+3)/2+1, $message);
        }
    }

The [bc] is there because we want the pattern to fail and force
backtracking. When I originally wrote this test I used "b" instead, however
the substring optimization kicks in and prevents the match from even
starting. So I had to hide the "fail" inside of something that the compiler
could not resolve at compile time. Anyway, because of the optimisations
that (?{ ... }) does disable the above pattern executes in quadratic time
as you can see above.

On the other hand, when it is rewritten as (*{ ... }) it does what I
expected when I original wrote the code above, it shows that the pattern is
executing in linear time:

    {
        my $message = 'Match is linear, not quadratic; Bug 22395.';
        our $count;
        my $ok= 0;
        for my $l (10, 100, 1000) {
            $count = 0;
            ('a' x $l) =~ /(.*)(*{ $count++ })[bc]/;
            $ok += is($count, $l + 1, $message);
        }
        is($ok,3, "Optimistic eval does not disable optimisations");
    }

Based on your feedback I will queue up a patch to remove (**{ ... }), but i
think that (*{ ... }) has real utility and should stay, and I also am very
much tempted to make (?{ ... }) actually disable even more optimisations
than it currently does so that it can be used for string decomposition type
problems without fighting the optimizer and hiding parts of the pattern
inside of character classes to prevent the optimisations from kicking in.

Another example of why i think we need both (?{ ... }) and (*{ ... }).

/(.*)(?{ ... })(*FAIL)/ -> the code block should execute even though the
compiler can determine the pattern never matches.

/(.*)(*{ ... })(*FAIL)/ -> the optimiser should be free to mark this
pattern as "never matches" and simply return false when it is used, and the
code block should never execute. (Just like it would if we replaced (*FAIL)
with a substring that isnt present in the string being matched).

cheers,
Yves


-- 

hvds:
> Personally i view (??{ ... }) as a relic that we would ditch if only we could (but we can't so I am not arguing we should), its primary utility was in solving problems that I think most people these days would use pattern recursion to solve, IOW, these days most older uses of EVAL can be replaced by GOSUB and people do not use it that much as the alternative is much superior in multiple ways.

Gosh, I quite strongly disagree with this characterization. I feel that `(??{ ... })` enables very different types of computation such as Regexp::Grammar for example.

I am also quite dubious that your characterization of the "primary usefulness" of `(?{...})` is what everyone else finds it primarily useful for.

I have suggested before that it should be possible for the programmer to say explicitly "[enable|disable] [all|some] optimization for this [pattern|fragment]", and you were strongly against that. I would still rather see that than see a new variant of `(?{...})` differing only in the optimization.

demerphq:
> enables very different types of computation such as Regexp::Grammar for example.

Regexp::Grammar uses GOSUB and DEFINE heavily (there are 102 mentioned of `(?&...)` and 11 mentions of `(??{...})`, and and 45 mentions of `(?{ ... })`. As far as I can tell it uses `(?{ ... })` as an action that is triggered when part of the pattern is matched, which is basically what I meant by "for debugging" (I could have been more artful in my description). So it uses GOSUB more than it uses (??{ ... }) and (?{ ... }) put together.

Sure there are a few tasks that (??{ ... }) can perform that GOSUB can't. But anything that GOSUB can do it does better. It doesn't need external vars and it gets optimized and there is no issue embedding it in a pattern or string. 

I do mostly see (??{ ... }) as a relic, most of the main uses of it are replaced by GOSUB and made much easier (eg nested constructs, reusing pattern snippets, etc), and the things left are IMO mostly toy tasks (Damian Conway code aside). I do realize there are some things you can do with it that you just can't use GOSUB for and I know some folks are fond of it which is why I /wouldn't/ argue that we should remove it. But let me put it this way, I don't think we would add it if we didn't have it and we did have GOSUB.  

GOSUB combined with DEFINE is a much more useful and powerful way to do most grammar related tasks. It is easy to write, easy to embed, has no traps and zaps like the closures issues with (??{ ... }) and etc it gets optimized, and it doesnt require patterns to be recompiled each time it executes. I am quite confident for everything that both can do GOSUB is the better choice. 

> I am also quite dubious that your characterization of the "primary usefulness" of (?{...}) is what everyone else finds it primarily useful for.

Please provide examples. I think if you are going to say something like that, and worded like that, the onus is on you to put more effort into your argument so that I have something to engage with.

> I have suggested before that it should be possible for the programmer to say explicitly "[enable|disable] [all|some] optimization for this [pattern|fragment]", and you were strongly against that. I would still rather see that than see a new variant of (?{...}) differing only in the optimization.

I think you missed the point. You can't disable optimizations for a fragment because of EVAL. A  postponed EVAL anywhere in the pattern can affect the rest of the pattern. Patterns arent linear anymore, they never really were, and are even less so these days. We have "forward backreferences", we have forward and backwards recursion, etc. All of these things affect the overall pattern. The idea that you could disable optimisations just for the grouped constructs that contained EVAL was wrong. An EVAL, especially a postponed eval *anywhere* in the pattern can affect the entire match.

And yes I am generally against users "controlling optimizations". There is only one place it would/should matter and that is in EVAL. Any other place where someone might even know there is an optimization at play is IMO just a bug. We should not be providing escape hatches/workarounds for bugs, we should be fixing them. If we start providing options for this stuff we will just end up with a huge list of options over time. That isn't manageable or scalable.  

I think if you are going to make these kinds of comments the onus is on you to provide a forward path. How would you manage this optimization control? What would the syntax look like? How would it work? How would we test it?  How would we deal with testing the permutations of options that might exist? How would we avoid the proliferation of such flags as we modernize and improve the regex engine? What would we do when we remove an optimization entirely?   It is easy to throw out an argument that we should do "blah", but a lot harder to put the flesh on the bones to turn it into a completed initiative.

My view is that `(??{ ... })` and `(?{ ... })` should disable pretty much ALL optimizations. I have already commited to removing `(**{ ... })` in a future patch, @iabyn made a good argument about why, so there is no problem there. On the other hand `(*{ ... })` has already proved its usefulness, and I am firmly of the view that removing it would be a bad thing.  Having an `(?{ ... })` that disables optimizations does serve some utility (although that utility is impacted by the fact that it doesn't disable all of them), having an `(*{ ... })` that doesn't disable any optimizations also has some utility.  There are things in the regex engine you can't test without it except by counting the number of lines of output you get from `use re debug`, and that is inefficient and painful. So I am firmly of the opinion that `(*{...})` is here to stay.

PR #20718, opened 2023-01-17
Configure: Add various C99 compatibility improvements

fweimer-rh:
Two C99 compatibility issues are fixed by these changes: Return types are made explicit where they previously defaulted to int, and all called functions are now declared explicitly (either by including additional headers, or by adding prototypes manually).

This avoids implict ints and implicit function declarations, both legacy C language features removed in the 1999 revision of the language.

Verified with an instrumented GCC compiler on GNU/Linux.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/Configure
+++ b/Configure
@@ -4655,7 +4655,7 @@ esac
 # so we will take those off from locincpth.
 case "$gccversion" in
 3.*)
-    echo "main(){}">try.c
+    echo "int main(){}">try.c
     for incdir in $locincpth; do
        warn=`$cc $ccflags -I$incdir -c try.c 2>&1 | \
 	     grep '^c[cp]p*[01]: warning: changing search order '`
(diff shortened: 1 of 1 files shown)


rwp0:
```sh
# Generated using [metaconfig 3.5 PL0]
# with additional metaconfig patches from https://github.com/Perl/metaconfig
```

Good, but isn't `Configure` a generated file?

fweimer-rh:
The instructions say to send patches?

```
# Though this script was generated by metaconfig from metaunits, it is
# OK to send patches against Configure itself (but not to commit them
# to blead). It's up to
# the Configure maintainers to backport the patch to the metaunits if it
# is accepted. Exceptions to this rule, and more information, is in
# Porting/pumpkin.pod.
```

khwilliamson:
Thank you for the patch

PR #20725, opened 2023-01-19
Bisection documentation (Gh20724)

hvds:
1) Modify recently-added example to use `expect-fail`.
2) Add section "ABOUT BISECTION" to explain the concept and git terminology a bit better.

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -1109,9 +1109,8 @@ where the code began to compile; the point where the code no longer emitted
 warnings; etc.
 
 We can use this program for that purpose, but we have to reverse our sense of
-"good" and "bad" commits.  In this case, F<git> will label the healing commit
-as "bad" (or label a breaking commit as "good"), so we have to invoke the
-program such that we do a C<die> when we reach the healing commit.
+"good" and "bad" commits.  We use the C<--expect-fail> option to do that
+reversal.
 
 =item * Problem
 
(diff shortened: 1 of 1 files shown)


tonycoz (approved):

Porting/bisect-runner.pl
```diff
@@ -844,6 +844,29 @@ =head1 OPTIONS
 
 =back
 
+=head1 ABOUT BISECTION
+
+The process is all about identifying the commit that caused some change
+in behaviour - maybe good, maybe bad. But it is built around the
+C<git-bisect> tool, which is much more specifically oriented around
+finding "what broke the build". C<git> terminology embeds that assumption -
+commits earlier than the target commit are "good" commits, those later
```
jkeenan: My impression is that we generally spell names of other programs with `F<>`, so I would recommend `F<git>` here, and `F<bisect.pl>` in two locations below.

jkeenan (changes requested):
Other than POD formatting, LGTM

hvds:
James E Keenan ***@***.***> wrote:
:My impression is that we generally spell names of other programs with `F<>`, so I would recommend `F<git>` here, and `F<bisect.pl>` in two locations below.

I've changed it to use `F<bisect.pl>`.

The phrase "C<git> terminology" is talking about the git ecosystem rather
than the executable, so I'm not convinced that change is as appropriate.

I've also tightened up the wording of the first paragraph some.

hvds:
Now merged.

PR #20731, opened 2023-01-20
regcomp.c - fix fencepost error duping a regex

demerphq:
In fe5492d916201ce31a107839a36bcb1435fe7bf0 I made a fencepost error copying the logical_to_parno and related data structures. They all needed a +1 on their size as the paren counts (logical and physical) as they have to account for capture buffer 0 which is always present which represents the entire match.


demerphq:
I don't see that there is anything to approve or review here. Merging.

PR #20736, opened 2023-01-23
check the IO object exists when writing to IO magic variables

### LINKED ISSUE #20733: writing to magic variables connected to the selected output GV when its IO object has been cleared crashes perl
**Description**

Tested at 5.15.78-2.el7.3.x86_64 (RED OS release MUROM 7.3.2 - Centos based system), but also reproduced this bug on other platforms x86_64.

**Steps to Reproduce**

Fonded several examples were found using endless recursion and causing falls.
Insert this string and run with ```perl -e``` or run string from file.
```perl
while(()){$0=();int""}((unpack("p>",pack("P"))))
```
```perl
for(0){$0=('',unpack('P>',pack('P')))}
```
```perl
sub o{}*STDOUT=0;{{$~=0}{()}}
```
**Expected behavior**

Is it correct that the interpreter does not processing such samples (even with the flag ```-w```) and just calling SEGFAULT when trying to address the memory areas inaccessible to recording?

**Perl configuration**

```
perl version: v5.30.1
```

tonycoz:
pp_select() ensures that the GV in PL_defoutgv has an IO object when it the default output is set, but can't prevent that GV being cleared afterwards, resulting in a seg fault when the variable is written.

To prevent this, check PL_defoutgv has an IO object before trying to write to it.

Fixes #20733

### DIFF
--- a/mg.c
+++ b/mg.c
@@ -3072,25 +3072,55 @@ Perl_magic_set(pTHX_ SV *sv, MAGIC *mg)
             IoLINES(GvIOp(PL_last_in_gv)) = SvIV(sv);
         break;
     case '^':
-        Safefree(IoTOP_NAME(GvIOp(PL_defoutgv)));
-        IoTOP_NAME(GvIOp(PL_defoutgv)) = savesvpv(sv);
-        IoTOP_GV(GvIOp(PL_defoutgv)) =  gv_fetchsv(sv, GV_ADD, SVt_PVIO);
+        {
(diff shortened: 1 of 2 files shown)


jkeenan:
Following a TDD approach, I applied a patch of `t/io/defout.t` *only* to blead and ran the test program.  It failed due to segfault.  I then tested the branch; `make test_harness` PASSed.

LGTM, but someone with more C-foo should review the changes to `mg.c`.

demerphq (approved):

PR #20737, opened 2023-01-24
Update `push()` documentation to include examples and be more novice friendly

scottchiefbaker:
The `push()` documentation is very developer heavy. I added some examples and simplified the wording so it's more novice friendly.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6162,13 +6162,12 @@ Adds one or more items to the B<end> of an array.
 	push(@animals, "mouse"); # ("cat", "mouse")
 
 	my @colors = ("red");
-	push(@colors, ("blue", "green"); # ("red", "blue", "green")
+	push(@colors, ("blue", "green")); # ("red", "blue", "green")
 
 Returns the number of elements in the array following the completed
 L<C<push>|/push ARRAY,LIST>.
 
-	# Return value is the number of items in the updated array
-	my $color_count = push(@colors, ("yellow", "purple");
+	my $color_count = push(@colors, ("yellow", "purple"));
 
 	say "There are $color_count colors in the updated array";
 
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6172,18 +6172,6 @@ L<C<push>|/push ARRAY,LIST>.
 
 	say "There are $color_count colors in the updated array";
 
-A more technical explanation:
-
-Treats ARRAY as a stack by appending the values of LIST to the end of
-ARRAY.  The length of ARRAY increases by the length of LIST.  Has the same
-effect as
-
-    for my $value (LIST) {
-        $ARRAY[++$#ARRAY] = $value;
-    }
-
-but is more efficient.
-
 Starting with Perl 5.14, an experimental feature allowed
 L<C<push>|/push ARRAY,LIST> to take a
 scalar expression. This experiment has been deemed unsuccessful, and was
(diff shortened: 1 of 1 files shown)


haarg:
I like the new additions, but I don't like the existing example using a loop. I feel like the entire "A more technical explanation:" to "but is more efficient." section could be eliminated. It seems unneeded and inaccurate.

pod/perlfunc.pod
```diff
@@ -6164,8 +6182,7 @@ effect as
         $ARRAY[++$#ARRAY] = $value;
```
demerphq: Shouldn't this be `$ARRAY[@ARRAY] = $value;`?

demerphq (changes requested):
Mostly looks ok to me. One minor nit, I don't think we should use `$#ARRAY++` in docs for push `@ARRAY` is more natural, and doesn't require understanding that you can change the size if the array by assigning to $#ARRAY. (Frankly that operation is rare enough that I had forgotten it was legal, presizing strings and hashes is rarely done but is still more common IME than presizing arrays.).

pod/perlfunc.pod
Grinnz: I think the intent is to show (to someone intimately familiar with how `$#prefix` works) that the operation increases the size of the array and then assigns the element there. But I agree with @haarg  that this entire example doesn't give any useful information.

demerphq: I'm good with removing, but if we don't we should restate as I diid. The original is cryptic, and unnecessary (as you know there is no need to presize the array to store into an index) and we shouldn't explain one concept by invoking a separate and even more complex concept. Assigning a value to the next available index is often done the way I wrote it (ive done it myself  many times [although usually in two statements] and ive seen others do it as well), but I can't recall ever having seen it done the way it's shown in that example.

pod/perlfunc.pod
```diff
@@ -6156,6 +6156,24 @@ X<push> X<stack>
 
 =for Pod::Functions append one or more elements to an array
 
+Adds one or more items to the B<end> of an array.
+
+	my @animals = ("cat");
+	push(@animals, "mouse"); # ("cat", "mouse")
+
+	my @colors = ("red");
+	push(@colors, ("blue", "green"); # ("red", "blue", "green")
```
JRaspass: Two open parens, one closing.

JRaspass: There's a few of these, and in the unshift examples. Not sure I'd bother with the inner parens, the subs are variadic and the parens might confuse people? YMMV

scottchiefbaker: Good catch... I updated the PR

scottchiefbaker:
It was pretty unanimous that the original technical definition didn't add a lot to the documentation so I removed it. The diff looks gnarly now because of all the text moving around, but the wording is MUCH cleaner now.

scottchiefbaker:
What's next on this PR? 

I'd like to work on `shift()` and `pop()` doc fixes next, but I want to focus on this until it lands.

haarg:
lgtm

scottchiefbaker:
Ping

iabyn:
I don't like the return value example. You effectively say the same thing 3 times ("returns the number of elements") without adding any more information or clarity. Showing the return value being printed out doesn't add anything. Perheps get rid of the 'say' and change the comment line just above the 'push' to
# sets $color_count to 5

demerphq:
This ticket was stalled. I have removed the comment that it was stalled by.

demerphq:
And I will merge once CI passes successfully.

demerphq (approved):

demerphq:
Thanks. Applied.

PR #20738, opened 2023-01-24
Update `unshift()` documentation to include examples and be more novice friendly

scottchiefbaker:
The `unshift()` documentation is very developer focused. I added some examples and simplified the wording so it's more novice friendly.


haarg (approved):

demerphq:
Thanks. Merged

PR #20741, opened 2023-01-26
pp_hot.c - fix branch reset matches in list context

demerphq:
I am kinda surprised this issue was not picked up by one of our other test files. I would have expected one of the t/re/regexp.t based patches to validate list context matches populating the list properly. But apparently not! So when I fixed branch reset in fe5492d916201ce31a107839a36bcb1435fe7bf0 I missed the list context logic. This fixes the oversight. Thanks to Andreas Koenig for the BBC report on this.

### DIFF
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -3205,30 +3205,86 @@ PP(pp_match)
     /* push captures on stack */
 
     {
-        const I32 nparens = RXp_NPARENS(prog);
-        I32 i = (global && !nparens) ? 1 : 0;
+        const I32 logical_nparens = RXp_LOGICAL_NPARENS(prog);
+        /* This following statement is *devious* code. If we are in a global
+           match and the pattern has no parens in it we should return $&
+           (offset pair 0). So we set logical_paren to 1 when we should return
+           $&, otherwise we set it to 0.
+
+           This allows us to simply add logical_nparens to logical_paren to
+           compute the number of elements we are going to return.
+
+           In the loop intit we "not" it with: logical_paren = !logical_paren
+           which results in it being 0 inside the loop when we want to return
+           $&, and results in it being 1 when we want to return the parens.
+           Thus we either loop over 1..logical_nparens, or just over 0.
+
+           This is an elegant way to do this code wise, but is super devious
+           and potentially confusing. When I first saw this logic I thought
+           "WTF?". But it makes sense after you poke it a while.
+
+           Frankly I probably would have done it differently, but it works so
+           I am leaving it. - Yves */
+        I32 logical_paren = (global && !logical_nparens) ? 1 : 0;
+        I32 *l2p = RXp_LOGICAL_TO_PARNO(prog);
+        /* this is used to step through the physical parens associated
+         * with a given logical paren. */
+        I32 *p2l_next = RXp_PARNO_TO_LOGICAL_NEXT(prog);
 
         SPAGAIN;			/* EVAL blocks could move the stack. */
-        EXTEND(SP, nparens + i);
-        EXTEND_MORTAL(nparens + i);
-        for (i = !i; i <= nparens; i++) {
-            if (LIKELY(RXp_OFFS_VALID(prog,i)))
-            {
-                const I32 len = RXp_OFFS_END(prog,i) - RXp_OFFS_START(prog,i);
-                const char * const s = RXp_OFFS_START(prog,i) + truebase;
-                if ( UNLIKELY( len < 0 || len > strend - s)
-                )
-                    DIE(aTHX_ "panic: pp_match start/end pointers, i=%ld, "
-                        "start=%ld, end=%ld, s=%p, strend=%p, len=%" UVuf,
-                        (long) i, (long) RXp_OFFS_START(prog,i),
-                        (long)RXp_OFFS_END(prog,i), s, strend, (IV) len);
-                PUSHs(newSVpvn_flags(s, len,
-                    (DO_UTF8(TARG))
-                    ? SVf_UTF8|SVs_TEMP
-                    : SVs_TEMP)
-                );
-            } else {
-                PUSHs(sv_newmortal());
+        EXTEND(SP, logical_nparens + logical_paren);    /* devious code ... */
+        EXTEND_MORTAL(logical_nparens + logical_paren); /* ... see above */
+
+        /* loop over the logical parens in the pattern. This may not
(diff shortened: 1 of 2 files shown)


pp_hot.c
```diff
@@ -3205,31 +3205,42 @@ PP(pp_match)
...
+                            (IV)paren, (IV)offs_start, (IV)offs_end, s, strend, (IV)len);
+                        NOT_REACHED; /* NOTREACHED */
+                    }
+                    PUSHs(newSVpvn_flags(s, len,
+                        (DO_UTF8(TARG))
+                        ? SVf_UTF8|SVs_TEMP
+                        : SVs_TEMP)
+                    );
+                    break;
+                    NOT_REACHED; /* NOTREACHED */
```
tonycoz: This and the similar below aren't useful.

The macro/comments are useful for telling the compiler/linter that this code can't be reached due to some sort of logic, but in both of these cases they're immediately preceded by an unconditional `break;` - the compiler or linter already knows the code can't be reached.

pp_hot.c
```diff
@@ -3205,31 +3205,42 @@ PP(pp_match)
...
+                if (((offs_start = RXp_OFFS_START(prog, paren)) != -1) &&
+                    ((offs_end   = RXp_OFFS_END(prog, paren))   != -1))
+                {
+                    const I32 len = offs_end - offs_start;
+                    const char * const s = offs_start + truebase;
+                    if ( UNLIKELY( len < 0 || len > strend - s) ) {
+                        DIE(aTHX_ "panic: pp_match start/end pointers, paren=%" IVdf ", "
+                            "start=%" IVdf ", end=%" IVdf ", s=%p, strend=%p, len=%" IVdf,
+                            (IV)paren, (IV)offs_start, (IV)offs_end, s, strend, (IV)len);
+                        NOT_REACHED; /* NOTREACHED */
```
tonycoz: As with the NOT_REACHED below, I don't think this is useful.

DIE() is defined as `#define DIE return Perl_die` so the next line of code can't execute anyway.

I think DIE is intended to be like `RETURN` and its variants in pp funcs - it marks the end of the execution, and is written so it can easily be used there (hence the OP* return type of Perl_die()) and in such a way that the compiler knows execution doesn't continue past that point.  (This paragraph took ~5 attempts to write, please pardon any clumsiness)

pp_hot.c
```diff
@@ -3205,31 +3205,42 @@ PP(pp_match)
...
-            }
+        EXTEND(SP, logical_nparens + i);
+        EXTEND_MORTAL(logical_nparens + i);
+        for (i = !i; i <= logical_nparens; i++) {
+            paren = l2p ? l2p[i] : i;
+            do {
+                if (((offs_start = RXp_OFFS_START(prog, paren)) != -1) &&
+                    ((offs_end   = RXp_OFFS_END(prog, paren))   != -1))
+                {
+                    const I32 len = offs_end - offs_start;
```
tonycoz: Not a new issue: can a single parens match >2GB of bytes?

pp_hot.c
```diff
@@ -3205,31 +3205,42 @@ PP(pp_match)
     /* push captures on stack */
 
     {
-        const I32 nparens = RXp_NPARENS(prog);
-        I32 i = (global && !nparens) ? 1 : 0;
+        const I32 logical_nparens = RXp_LOGICAL_NPARENS(prog);
+        I32 i = (global && !logical_nparens) ? 1 : 0;
+        I32 *l2p = RXp_LOGICAL_TO_PARNO(prog);
+        I32 *p2l_next = RXp_PARNO_TO_LOGICAL_NEXT(prog);
+        I32 paren, offs_start, offs_end;
```
tonycoz: I expect the match start/end positions can be more than 2GB into the string.

pp_hot.c
demerphq: Offsets are Ssize_t so presumably yes. Good catch. Will fix.

pp_hot.c
demerphq: Yep

pp_hot.c
demerphq: I mostly put that in for human eyes, and then pu the others in for 'completeness' but if you think they are unhelpful ill remove them.

pp_hot.c
demerphq: Fixed. I also did a quick audit over similar code and it all seemed to be using SSize_t instead of I32.  Thanks!

pp_hot.c
```diff
@@ -3205,30 +3205,60 @@ PP(pp_match)
     /* push captures on stack */
 
     {
-        const I32 nparens = RXp_NPARENS(prog);
-        I32 i = (global && !nparens) ? 1 : 0;
+        const I32 logical_nparens = RXp_LOGICAL_NPARENS(prog);
+        I32 logical_paren = (global && !logical_nparens) ? 1 : 0;
+        I32 *l2p = RXp_LOGICAL_TO_PARNO(prog);
+        I32 *p2l_next = RXp_PARNO_TO_LOGICAL_NEXT(prog);
+        SSize_t offs_start, offs_end;
```
tonycoz: These are only used inside the while loop and so the definitions could be moved there.

pp_hot.c
```diff
@@ -3205,30 +3205,60 @@ PP(pp_match)
...
+                 * while the end is -1. */
+                if (((offs_end   = RXp_OFFS_END(prog, phys_paren))   != -1) &&
+                    ((offs_start = RXp_OFFS_START(prog, phys_paren)) != -1))
+                {
+                    const SSize_t len = offs_end - offs_start;
+                    const char * const s = offs_start + truebase;
+                    if ( UNLIKELY( len < 0 || len > strend - s) ) {
+                        DIE(aTHX_ "panic: pp_match start/end pointers, paren=%" IVdf ", "
+                            "start=%" IVdf ", end=%" IVdf ", s=%p, strend=%p, len=%" IVdf,
+                            (IV)phys_paren, (IV)offs_start, (IV)offs_end, s, strend, (IV)len);
```
tonycoz: You could avoid the casts here by using I32df for the I32 value (phys_paren), and `%zd` for the SSize_t values.

tonycoz (approved):

pp_hot.c
demerphq: willdo.

pp_hot.c
demerphq: ack. willdo also.

PR #20744, opened 2023-01-27
Expose {optimize,finalize}_optree() as real API functions

### LINKED ISSUE #20743: BBC: 0351a629e7 prevents List-Keywords from building
On IRC #p5p yesterday it was mentioned that CPAN distribution List-Keywords was no longer building against blead.  [Sample failure report](http://www.cpantesters.org/cpan/report/115d8d2e-01ae-11ed-9835-846ddf0dbbbc).  Bisection of failure reports with the following command:
```
perl Porting/bisect.pl \
--module=List::Keywords \
--start=fecef29147d516d89233c43e3a9c9800e6c7f9f4 \
--end=1e45e087ebebb524df3c2247e485d7d6aebcd6f0
```
... pointed to 0351a629e71de127cbfd1b142e9eaa6069deabf5
```
0351a629e71de127cbfd1b142e9eaa6069deabf5 is the first bad commit
commit 0351a629e71de127cbfd1b142e9eaa6069deabf5
Author: Tomasz Konojacki <me@xenu.pl>
Date:   Sat Jun 18 07:26:58 2022 +0200
Commit:     Tomasz Konojacki <me@xenu.pl>
CommitDate: Sat Jun 18 08:51:14 2022 +0200

    hide private functions with __attribute__((visibility("hidden")))
    
    This allows us to enforce API boundaries and potentially enables
    compiler optimisations.
    
    We've been always hiding non-public symbols on Windows. This commit
    brings that to the other platforms.
```
@xenu, @LeoNerd, can you take a look?

leonerd:
These are required by XS modules which want to create custom LOGOP-shaped optrees, to ensure that both sides of the tree get optimised and finalised.

See also
  https://github.com/Perl/perl5/issues/20743

Fixes #20743

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -1802,8 +1802,6 @@ Axmd	|OP *	|op_lvalue	|NULLOK OP *o				\
 poX	|OP *	|op_lvalue_flags|NULLOK OP *o				\
 				|I32 type				\
 				|U32 flags
-Apd	|void	|optimize_optree|NN OP *o
-Apd	|void	|finalize_optree|NN OP *o
 : Used in op.c and pp_sys.c
 p	|int	|mode_from_discipline					\
 				|NULLOK const char *s			\
@@ -3900,6 +3898,10 @@ EeiT	|void * |my_memrchr	|NN const char *s			\
 				|const STRLEN len
 # endif /* !defined(HAS_MEMRCHR) */
 #endif /* defined(PERL_CORE) || defined(PERL_EXT) */
+#if defined(PERL_CORE) || defined(PERL_USE_VOLATILE_API)
+Apd	|void	|optimize_optree|NN OP *o
+Apd	|void	|finalize_optree|NN OP *o
+#endif /* defined(PERL_CORE) || defined(PERL_USE_VOLATILE_API) */
 #if defined(PERL_DEBUG_READONLY_COW)
 p	|void	|sv_buf_to_ro	|NN SV *sv
 #endif /* defined(PERL_DEBUG_READONLY_COW) */
--- a/embed.fnc
+++ b/embed.fnc
@@ -1802,7 +1802,8 @@ Axmd	|OP *	|op_lvalue	|NULLOK OP *o				\
 poX	|OP *	|op_lvalue_flags|NULLOK OP *o				\
 				|I32 type				\
 				|U32 flags
-pd	|void	|finalize_optree|NN OP *o
+Apd	|void	|optimize_optree|NN OP *o
+Apd	|void	|finalize_optree|NN OP *o
 : Used in op.c and pp_sys.c
 p	|int	|mode_from_discipline					\
 				|NULLOK const char *s			\
@@ -4585,7 +4586,6 @@ p	|void	|check_hash_fields_and_hekify				\
 				|NULLOK SVOP *key_op			\
 				|int real
 p	|SV *	|op_varname	|NN const OP *o
-pd	|void	|optimize_optree|NN OP *o
 p	|void	|warn_elem_scalar_context				\
 				|NN const OP *o 			\
 				|NN SV *name				\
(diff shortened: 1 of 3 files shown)


demerphq:
Id feel a lot better if these were guarded by a define check that people had to enable to access them.

Something like PERL_VOLATILE_API maybe.

I think we need something more than "fully public" and "hidden". Something like PERL_EXT, but more substem specific. We already have precedent for some of the regex engine related functionality.

leonerd:
> Id feel a lot better if these were guarded by a define check that people had to enable to access them.
> 
> Something like PERL_VOLATILE_API maybe.

Ooh that's a fun thought, yes. Do we like that name?

demerphq:
Maybe PERL_USE_VOLATILE_API is a bit better. Anything that signifies clearly that the user is utilizing functionality we don't encourage people to use and that they understand the implications of doing so.

leonerd:
OK, how's that now?

demerphq (approved):

iabyn:
On Fri, Jan 27, 2023 at 07:31:06AM -0800, Paul Evans wrote:
> These are required by XS modules which want to create custom
> LOGOP-shaped optrees, to ensure that both sides of the tree get
> optimised and finalised.

The trouble with just exposing random bits of the core as API for XS code
is that we're now forever promising not to break such XS code, even
though we haven't specified in the documentation what exactly the function
does, or when it's safe to use it, etc. It's another of those little 'we
can't change anything for the better ever again' nails in the coffin of
perl.

C.f. one of my bugbears: Perl_peep() and Perl_rpeep() are both hookable,
meaning that perl's entire peephole optimiser process - the way and order
in which it processes optrees - is now basically frozen, and we can never do
anything different or better.

-- 
Please note that ash-trays are provided for the use of smokers,
whereas the floor is provided for the use of all patrons.
    -- Bill Royston

iabyn:
On Fri, Jan 27, 2023 at 06:42:18PM -0800, Yves Orton wrote:
> Maybe PERL_USE_VOLATILE_API is a bit better. Anything that signifies
> clearly that the user is utilizing functionality we don't encourage
> people to use and that they understand the implications of doing so.

Based on our experience of how otherwise competent programmers seem to
interpret "use experimental foo" as meaning "this will magically make my
code never break when using this new feature", I wouldn't hold my breath.

-- 
    -- Churchill on Montgomery

demerphq:
@iabyn I take your point, but its better than /just/ marking it as API. 

Experimental stuff at the perl level is a bit different from saying an API is volatile. We had stuff we have no intention or removing and which we treat as API which was marked as experimental one way or another for *ages*. So we kinda mudied the water there.  Also "experimental" kinda implies "will eventually become not experimental (or will be removed)". With PERL_USE_VOLATILE_API what we want to say is "you can use this, but we reserve the right to change it any way and the best you are going to get is some BBC support if we change it".

For what its worth, I strongly support your view that we shouldn't let users bake-into-stone our internals when we don't absolutely have to. I have similar concerns about some of our other internals and current proposals (RFC 13 in particular concerns me in this regard). Same for the `@INC` changes.

> C.f. one of my bugbears: Perl_peep() and Perl_rpeep() are both hookable, meaning that perl's entire peephole optimiser process - the way and order in which it processes optrees - is now basically frozen, and we can never do anything different or better.

Has somebody actually complained about this or are you applying the letter of the backward compatibility law? Personally I'd have no problem at all with also marking them as PERL_USE_VOLATILE_API and saying "nope, we can change this if we need to, and the best you'll get is basic BBC support".  

BTW, I have similar issues with the regexp changes. Whatever I do will break the regex engines that are out there. I dont treat that as something I can't do, I treat it as something that the engine maintainers need to work with me to fix. IMO our first priority is to our language users, not XS module authors. If we have to shake the XS tree occasionally to make the internals better then that is what we should do. 

Anyway, I take your point, no matter what our policy is someone will freak out when their code doesn't work anymore. But we have to balance our primary audience (perl programmers) from our secondary audience (XS module authors). So long as we provide reasonable BBC support we should be ok. I hope. :-)

Tux:
And it needs to be added to the signature (Config::Perl::V) as IMHO this indicates binary incompatability

demerphq:
@Tux I think you misunderstand. Only an XS author who wants use these functions needs to define PERL_USE_VOLATILE_API. This has nothing to do with Config::Perl::V.

Tux:
@demerphq thanks for the clarification. Now I get it.

PR #20746, opened 2023-01-27
Update autodie to CPAN version 2.35

toddr:
[DELTA]

2.35      2023-01-27 16:00:26+00:00 UTC

        * Prepare for 5.38 changes to deprecate smartmatch

        * Remove +x bit from pm and t files

        * CI - Turn off Pod coverage and critic tests below 5.12


book (approved):

jkeenan:
@toddr, with a p.r. like this, one way that I QA it is to build blead, run the module's tests (which should PASS), apply a patch for just the `.t` files in the p.r., run those tests (which should FAIL), then apply the entire p.r., all of whose tests should now PASS.  In short, TDD.

Doing that here I get a PASS when patching in only `t/exceptions.t`.  Am I correct in thinking that that is because we do not yet have `deprecated::smartmatch` in blead?

jkeenan (approved):

toddr:
> Doing that here I get a PASS when patching in only `t/exceptions.t`. Am I correct in thinking that that is because we do not yet have `deprecated::smartmatch` in blead?

That is correct.

PR #20747, opened 2023-01-28
WIP: split out matched offset data from regexp structure; create SVt_RXMO for it.

demerphq:
This is a work in progress. It is on top of the patches in yves/curlyx_curlym. The idea is to speed up matching by detaching the match results from the regexp itself, ultimately getting rid of the mother_re concept.

This means we have to "reverse" the relationship between pmops' and regexp and the last successful match data they contain so that the pmop contains a RXMO structure, which points at the regexp it is for.

Very much a work in progress.


dump.c
```diff
@@ -70,11 +70,23 @@ static const char* const svshorttypenames[SVt_LAST] = {
     "IO"
 };
 
+static const char* const unknowntypename = "UNKNOWN";
```
tonycoz: Better as
```
static const char unknowntypename[] = "UNKNOWN";
```

tonycoz:
You might already be heading there, but I was wondering if this could/would lead to having a user visible match result object?
```
my $re = qr/.../;
if (my $match = $re->match("somestring")) { # optional start parameter
  # illustrative, not normative
  # perhaps look at existing 
  print $re->text; # $&
  print $re->text(1); # $1 etc
  print $re->start; # $-[0]
  # and more
}
```
Python does something similar: https://docs.python.org/3/library/re.html#match-objects

demerphq:
It has definitely occurred to me yes but for the short term I am not planning to do this as it introduces a layer of additional complexity related to SvPVLV which I see in the regexp code which frankly I don't entirely understand and have been trying to avoid needing to understand. :-) As you can see from the failing tests here this is already a bit hairy. 

But yes, I have that it would be a natural next step to look into once I can get this working. I haven't spent anytime at all thinking about what the api would look like tho. I guess a modifier that caused the match to return an object would be possible, as would a function which returned the match results for PL_curpm. If you or anyone else has ideas maybe we could open an issue to collect them? (I dont think comments on this PR would be the right place.) I guess we could look at python for precedent there. It also implies we probably need to bless these objects and etc, which I also haven't thought much about. Right now having them be purely internal objects like SVt_INVLIST saves me having to worry about a lot of possible issues that might come up, but I certainly recognize the potential these new objects have.

dump.c
demerphq: ack. thanks!

tonycoz:
The RXMO doesn't necessarily need to be blessed, or even visible to the perl user, it could be done as a wrapper that hides the RXMO.

But as you say, that comes later.

iabyn:
On Sun, Jan 29, 2023 at 04:56:15PM -0800, Yves Orton wrote:
> It has definitely occurred to me yes but for the short term I am not
> planning to do this as it introduces a layer of additional complexity
> related to SvPVLV which I see in the regexp code which frankly I don't
> entirely understand and have been trying to avoid needing to understand.

You may find useful, the 'Background' section in the commit message for

    commit df6b4bd56551f2d39f7c0019c23f27181d8c39c4
    Author:     David Mitchell ***@***.***>
    AuthorDate: Fri Jul 7 14:13:32 2017 +0100

    give REGEXP SVs the POK flag again


-- 
My get-up-and-go just got up and went.

PR #20748, opened 2023-01-29
Update Memoize to 1.16

ap:


demerphq:
FWIW, in the future please document the changes made in the commit message. (Some folks object to title only commits). But im merging anyway. Thanks for the update.

ap:
Thanks. FTR I don’t mind reworking my PRs to suit preferences at all, so feel free to bounce them back until they do.

The only real reason I even submitted a PR for this myself was that I was aware of the now-obsolete entry in `EXCLUDED` in `Porting/Maintainers.pl` and wanted to ensure its removal would be part of the sync commit. I guess that ought to have been noted in the commit message.

demerphq:
Probably would have been better to note that, but IMO it's not that big a deal. If it hadn't been a module sync I would have asked for a repush, but for a module sync it's kinda a Grey zone. I'm not sure I've created very good body messages for those either. Shrug.

PR #20750, opened 2023-01-29
Configure - fix handling of quoted gcc output

### LINKED ISSUE #20606: Build Failure - Configure script incorrectly parses gcc's output when it has quotes
**Description**
The perl source's `Configure` script fails when the output of `gcc -v` contains entries that are enclosed in quotation marks.

When Configure is run, prior to building perl, one of the checks it performs is determining pre-defined compiler flags.  One way it does this is to run `gcc -v` and parse its output. However it is possible when gcc is built to have passed it configuration options in such a way as its gcc  will produce output that has single quotes.  For example:

```
# gcc -v
Using built-in specs.
COLLECT_GCC=/usr/bin/arm-openwrt-linux-muslgnueabi-gcc_orig
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-openwrt-linux-muslgnueabi/11.2.0/lto-wrapper
Target: arm-openwrt-linux-muslgnueabi
Configured with: /builder/shared-workdir/build/sdk/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/gcc-11.2.0/configure --target=arm-openwrt-linux --host=arm-openwrt-linux --build=x86_64-pc-linux-gnu --program-prefix= --program-suffix= --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --libexecdir=/usr/lib --sysconfdir=/etc --datadir=/usr/share --localstatedir=/var --mandir=/usr/man --infodir=/usr/info --disable-nls 'CXXFLAGS_FOR_TARGET=-g -O2 -D_GLIBCXX_INCLUDE_NEXT_C_HEADERS' --build=x86_64-pc-linux-gnu --host=arm-openwrt-linux-muslgnueabi --target=arm-openwrt-linux-muslgnueabi --enable-languages=c,c++ --with-bugurl=https://dev.openwrt.org/ --with-pkgversion='OpenWrt GCC 11.2.0' --enable-shared --disable-__cxa_atexit --with-default-libstdcxx-abi=gcc4-compatible --enable-target-optspace --with-gnu-ld --disable-nls --disable-libsanitizer --disable-libvtv --disable-libcilkrts --disable-libmudflap --disable-libmpx --disable-multilib --disable-libgomp --disable-libquadmath --disable-libssp --disable-decimal-float --disable-libstdcxx-pch --with-host-libstdcxx=-lstdc++ --prefix=/usr --libexecdir=/usr/lib --with-local-prefix=/usr --with-stage1-ldflags=-lstdc++ --with-float=hard
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.2.0 (OpenWrt GCC 11.2.0) 
```
Note the single quotes starting at `CXXFLAGS_FOR_TARGET` and ending at `-D_GLIBCXX_INCLUDE_NEXT_C_HEADERS`

This causes Configure to pass on a single quote into config.sh, which creates mis-matched quotes with predictable results.  For example the above output from `gcc -v` makes the following line in the resulting `config.sh` 
`ccsymbols='_GLIBCXX_INCLUDE_NEXT_C_HEADERS'=1'`

The resulting mismatched quotes cause problems with everything that tries to invoke `config.sh`

Affected systems include OpenWrt and any other device using ENTWARE as its package system.

**Steps to Reproduce**
On an affected system:

1. wget https://www.cpan.org/src/5.0/perl-5.37.6.tar.gz
2. tar -xvzf perl-5.37.6.tar.gz
3. cd perl-5.37.6
4. ./Configure

**Expected behavior**
Should configure properly

**Perl configuration**
N/A

**Patch**
The following patch (made against Configure itself) will strip quotes from the output of "gcc -v" before further processing:
```
--- Configure.orig	2022-11-14 14:52:18.000000000 -0400
+++ Configure	2022-12-11 16:47:27.642795981 -0400
@@ -23770,11 +23770,11 @@
 extern int foo;
 EOF
 for i in \`$cc -v -c tmp.c 2>&1 $postprocess_cc_v\`
 do
 	case "\$i" in
-	-D*) echo "\$i" | $sed 's/^-D//';;
+	-D*) echo "\$i" | $sed 's/^-D//;s/['\''\"]//g';;
 	-A*) $test "$gccversion" && echo "\$i" | $sed 's/^-A//' | $sed 's/\(.*\)(\(.*\))/\1=\2/';;
 	esac
 done
 $rm_try
 EOS
```

demerphq:
This patch was submitted in GH issue #20606. When gcc output contains quoted elements we fail to handle it properly. This tweaks the sed command to do so.

Fixes #20606.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -810,6 +810,7 @@ Kragen Sitaker                 <kragen@pobox.com>
 Krishna Sethuraman             <krishna@sgi.com>
 Kriton Kyrimis                 <kyrimis@princeton.edu>
 Kurt D. Starsinic              <kstar@wolfetech.com>
+Kurt Fitzner                   <kurt@va1der.ca>
 Kyriakos Georgiou
 Lajos Veres                    <vlajos@gmail.com>
 Larry Parmelee                 <parmelee@CS.Cornell.EDU>
--- a/Configure
+++ b/Configure
@@ -23778,7 +23778,7 @@ EOF
 for i in \`$cc -v -c tmp.c 2>&1 $postprocess_cc_v\`
 do
 	case "\$i" in
-	-D*) echo "\$i" | $sed 's/^-D//';;
+	-D*) echo "\$i" | $sed 's/^-D//;s/['\''\"]//g';;
 	-A*) $test "$gccversion" && echo "\$i" | $sed 's/^-A//' | $sed 's/\(.*\)(\(.*\))/\1=\2/';;
 	esac
 done


demerphq:
@VA1DER please review this PR and verify your details in the AUTHORS file.

VA1DER (approved):

demerphq:
As per @Tux's direction in https://github.com/Perl/perl5/issues/20606#issuecomment-1398561339 I am merging this.

PR #20752, opened 2023-01-29
[Pod] UNIVERSAL: Add lexical variables in Synopsis

rwp0:
Add `my`, and name variables differently to avoid confusion.

### DIFF
--- a/lib/UNIVERSAL.pm
+++ b/lib/UNIVERSAL.pm
@@ -1,6 +1,6 @@
 package UNIVERSAL;
 
-our $VERSION = '1.14';
+our $VERSION = '1.15';
 
 # UNIVERSAL should not contain any extra subs/methods beyond those
 # that it exists to define. The existence of import() below is a historical
@@ -24,21 +24,21 @@ UNIVERSAL - base class for ALL classes (blessed references)
 
 =head1 SYNOPSIS
 
-    $is_io    = $fd->isa("IO::Handle");
-    $is_io    = Class->isa("IO::Handle");
+    my $obj_is_io    = $fd->isa("IO::Handle");
+    my $cls_is_io    = Class->isa("IO::Handle");
 
-    $does_log = $obj->DOES("Logger");
-    $does_log = Class->DOES("Logger");
+    my $obj_does_log = $obj->DOES("Logger");
+    my $cls_does_log = Class->DOES("Logger");
 
-    $sub      = $obj->can("print");
-    $sub      = Class->can("print");
+    my $obj_sub      = $obj->can("print");
+    my $cls_sub      = Class->can("print");
 
-    $sub      = eval { $ref->can("fandango") };
-    $ver      = $obj->VERSION;
+    my $eval_sub     = eval { $ref->can("fandango") };
+    my $ver          = $obj->VERSION;
 
     # but never do this!
-    $is_io    = UNIVERSAL::isa($fd, "IO::Handle");
-    $sub      = UNIVERSAL::can($obj, "print");
+    my $is_io        = UNIVERSAL::isa($fd, "IO::Handle");
+    my $sub          = UNIVERSAL::can($obj, "print");
 
 =head1 DESCRIPTION
 


bsmith:
Would it make more sense to group the three uses of `can` together? (ie move the blank line 35 down one line).

Would it help to show an example of the other usage of VERSION?

rwp0:
> Would it make more sense to group the three uses of `can` together? (ie move the blank line 35 down one line).
> 
> Would it help to show an example of the other usage of VERSION?

Good, but maybe best done in a different pull request I think, since I intended this one as adding `my`-only.

Grinnz:
Please indicate in the title and commit message that this is for UNIVERSAL.

rwp0:
Sure I can, but isn't it apparent in the modified file list?

Grinnz:
Clearly, but that doesn't help when looking at a list of PRs or commits of which this project has numerous.

jkeenan:
On 1/29/23 15:37, Elvin Aslanov wrote:
> Sure I can, but isn't it apparent in the modified file list?
> 

You are correct that the modified files list in a commit message 
presents this information.  However, people working on the core 
distribution are often in the position of scrolling through the output 
of `git log` for dozens or hundreds of commits.  If a particular commit 
is very focused on one particular file, including that file's name (or 
its basename) in the commit message functions as a semantic tag to catch 
the eye.

There's no hard-and-fast rule about this; it's always a judgment call.

demerphq:
If there is only one file modified it should be named. If there are more than one then pick the one most likely to be relevant and indicative of the nature of the change.

rwp0:
> If there is only one file modified it should be named. If there are more than one then pick the one most likely to be relevant and indicative of the nature of the change.

As a comitter/maintainer, could you please amend my commit message to include that?

I think that should be possible to do through GitHub since I left the tick to allow edits, but not very sure on that.

If for any reason not, then I will try to do that later myself.

demerphq:
@rwp0 done.

rwp0:
> @rwp0 done. 

Thanks!

demerphq:
Just a minor note, it is risky practice to name your topic branches with just one level. It massively increases the chance that you will end up with a collision with an existing branch, especially when the branch name is something like "patch-29" ;-). 

I would recommend you always name your topic branches $your_name/$topic_name so that you don't have conflicts with someone else who wants to use $topic_name as well. Eg, all of my patches are yves/$topic_name. Since your are pushing to your own repo it probably isn't a big deal, and most people who push patches to the master repo use the binomial names, but I would recommend you develop the practice of always using binomial branch names just to avoid any issues.  For instance if someone who has access to the master repo pushed a "patch-29" there would be a conflict that can be annoying to deal with.

Another thing is that as far as I know there is no way in github itself to modify someones PR. To do so I had to do this:

```
git remote add -f rwp0 git@github.com:rwp0/perl5.git
git co -b patch-29 # here is where conflicts can come up
git commit --amend # and edit the patch
git push -f
```
HTH.

demerphq (approved):
Personally I am not super sure adding `my` to everything is a good idea, but the consensus in the core dev group seems to be the opposite so I am approving. Thanks for the contribution!

rwp0:
> Just a minor note, it is risky practice to name your topic branches with just one level. It massively increases the chance that you will end up with a collision with an existing branch, especially when the branch name is something like "patch-29" ;-). 
> 
> I would recommend you always name your topic branches $your_name/$topic_name so that you don't have conflicts with someone else who wants to use $topic_name as well. Eg, all of my patches are yves/$topic_name. Since your are pushing to your own repo it probably isn't a big deal, and most people who push patches to the master repo use the binomial names, but I would recommend you develop the practice of always using binomial branch names just to avoid any issues.  For instance if someone who has access to the master repo pushed a "patch-29" there would be a conflict that can be annoying to deal with.
> 
> Another thing is that as far as I know there is no way in github itself to modify someones PR. To do so I had to do this:
> 
> ```
> git remote add -f rwp0 git@github.com:rwp0/perl5.git
> git co -b patch-29 # here is where conflicts can come up
> git commit --amend # and edit the patch
> git push -f
> ```
> HTH.

Thanks for the suggestions, it was very interesting to read this text, but I actually used
GitHub's Edit button on file to modify it and submit as PR, so the branch name was auto-generated.

> git co -b patch-29 # here is where conflicts can come up

I hope GitHub reckons for this, and is not letting the branch name conflicts happen
by always providing a unique branch name to make its PR-by-Edit-button feature safe.

PR #20755, opened 2023-01-31
embed.pl - the 's', 'S', 'i' and 'I' flags are mutually exclusive

demerphq:
We had a bug where we processed the first one in the flags definition. Sorting the flags or rearranging them changes the output, which shouldn't happen.

### DIFF
--- a/regen/embed.pl
+++ b/regen/embed.pl
@@ -128,8 +128,21 @@ sub generate_proto_h {
         die_at_end "$plain_func: u flag only usable with m" if $flags =~ /u/
                                                             && $flags !~ /m/;
 
+        my ($static_flag, @extra_static_flags)= do {
+            # the seen filter can be removed once flag dedupe
+            # is done in tidy_embed.pl
+            my %seen;
+            grep !$seen{$_}++, $flags =~/([SsIi])/g;
+        };
+
+        if (@extra_static_flags) {
+            my $flags_str = join ", ", $static_flag, @extra_static_flags;
+            $flags_str =~ s/, (\w)\z/ and $1/;
+            die_at_end "$plain_func: flags $flags_str are mutually exclusive\n";
+        }
+
         my $static_inline = 0;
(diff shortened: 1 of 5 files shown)


regen/embed.pl
```diff
@@ -128,24 +128,34 @@ sub generate_proto_h {
         die_at_end "$plain_func: u flag only usable with m" if $flags =~ /u/
                                                             && $flags !~ /m/;
 
+        my %static_flags;
+        while ($flags=~/([SsIi])/g) {
```
JRaspass: Replacing the entire loop with a slice would be simpler and cleaner:
```perl
@static_flags{ $flags =~/[SsIi]/g } = ();
```

I'm not sure what version of Perl introduced hash slices (as opposed to hash key/value slices) but I note that despite this file claiming to support `5.004` it contains at least one use of defined-or `//=` so clearly 5.10 is at least acceptable for this file in this day and age!

demerphq: Thanks for the suggestion. After playing with this a bit I went a slightly different route that allows me to minimize having to call keys on %static_flags so many times.

demerphq: BTW, I would love to see someone (maybe even myself) cleanup embed.pl so that the flags validation is done via a data structure that specifies the mutually exclusive flags and the mutually required flags in a data driven approach. Then a whole bunch of code in there can "disappear" and be replaced by a single function that validates all the flags at once.

khwilliamson:
On 1/31/23 03:40, James Raspass wrote:
> I'm not sure what version of Perl introduced hash slices (as opposed to 
> hash key/value slices) but I note that despite this file claiming to 
> support |5.004| it contains at least one use of defined-or |//=| so 
> clearly 5.10 is at least acceptable for this file in this day and age!

I don't see any reason for this file to have to worry about any 
backwards compatibility whatsoever.

demerphq:
I dont see that there is anything controversial with this PR, so I am merging.

PR #20757, opened 2023-01-31
embed.fnc - sort functions by name (in their guard groups) and sort flags

demerphq:
This sorts the functions in embed_fnc by name by group. It uses a library sort then key sort, then line number sort to group similar functionality together. It also dedupes the flags list and sorts them so we do not have a mismash of different ordering of the flags.

This is based on top of https://github.com/Perl/perl5/pull/20755, if this is applied we can drop that, or vice versa when that is applied I can rebase away the duplicate patch.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -616,20 +616,20 @@ px	|LOGOP *|alloc_LOGOP	|I32 type				\
 p	|PADOFFSET|allocmy	|NN const char * const name		\
 				|const STRLEN len			\
 				|const U32 flags
-Xpd	|bool	|amagic_applies |NN SV *sv				\
+Xdp	|bool	|amagic_applies |NN SV *sv				\
 				|int method				\
 				|int flags
-Apd	|SV *	|amagic_call	|NN SV *left				\
+Adp	|SV *	|amagic_call	|NN SV *left				\
 				|NN SV *right				\
 				|int method				\
 				|int dir
-Apd	|SV *	|amagic_deref_call					\
+Adp	|SV *	|amagic_deref_call					\
 				|NN SV *ref				\
 				|int method
 p	|bool	|amagic_is_enabled					\
 				|int method
 
-EXiTp	|void	|append_utf8_from_native_byte				\
+ETXip	|void	|append_utf8_from_native_byte				\
 				|const U8 byte				\
 				|NN U8 **dest
 : FIXME - this is only called by pp_chown. They should be merged.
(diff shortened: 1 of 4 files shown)


Leont:
I'm not sure (in the literal sense) if this is a good idea. Take for example the «: peephole optimiser» section. Previously that contained `peep` and `rpeep`, now it only contains the former. Not all of embed.fnc is chaos, some of it is order and that order is useful.

demerphq:
I'll take a look. It wouldn't be too hard to create groups that stay together. This is already done for define groups after all. Also, this was already approved on the original merge, its just in a last minute cleanup I accidentally disabled the sort callback by removing the leading undebar.  I really want to get this and other build tooling to be automated. While there may be a few parts of it that aren't a chotic mess, most of it is. It's a nightmare trying to find out if there is already a function that does something that you want to do.

demerphq:
Hi @Leont I looked into this a bit, and I have some code I'm working on to allow people to define groups with comments in the code. But the more I have looked at it the less comfortable I am with it. It feels like it opens up a bunch of issues.

Before I go on I guess I should make it clear I am coming at this from the point of view of embed.fnc being a living document, and that circumstances change over time, for instance splitting up our files into smaller compilation units. A big part of my motivation with this new functionality is to remove all doubt and overhead in managing the file, and to remove frustration when performing activities like refactoring and splitting large C files. 

So with regard to peep/rpeep, they aren't API, and they are only used in op.c and peep.c, so really if they should be grouped they should be grouped by putting them inside of a 

```
#if defined(PERL_IN_OP_C) || defined(PERL_IN_PEEP_C) || defined(PERL_USE_VOLATILE_API)
```

guard clause so they don't get used unknowingly in a different file. Not sure if the PERL_USE_VOLATILE_API part should be there, but since they aren't marked to be hidden maybe that would be good idea.

These type of guard clauses are a big reason why I am reluctant to add enforce grouping rules. A big motivation with this new functionality is to make it dead simple to move functions between files, and restructure our code. So whatever grouping may have made sense at one point of time might no longer make sense later, and the existing grouping logic would regroup the files by guard clause anyway. Eg, what happens if someone wants to group a set of functions for one reason, but the guard clauses want them to be in another set of groups?    

We could add data to the file, or extract it from the code itself, to group files as much as possible by the functions they contain. For instance peep.c contains:

```
Perl_optimize_optree(pTHX_ OP* o)
Perl_finalize_optree(pTHX_ OP* o)
Perl_rpeep(pTHX_ OP *o)
Perl_peep(pTHX_ OP *o)
```

currently optimize_optree and finalize_optree arent grouped with peep and rpeep, surely they should be? And if they were then we probably would end up with:

```
Perl_finalize_optree(pTHX_ OP* o)
Perl_optimize_optree(pTHX_ OP* o)
Perl_peep(pTHX_ OP *o)
Perl_rpeep(pTHX_ OP *o)
```

One thing that occured to me is that we could add metadata "tags" to the function defs so that we can generate a reference file (or files) that has things grouped according to the tags so that we can represent multiple different groupings at once. For instance something like "SvREFCNT_inc" and SvREFCNT_dec  might be grouped both under "memory management" and also under "SV related functions". But if we try to do grouping directly in embed.fnc we can only have one grouping,  and choosing the right grouping to apply is really an arbitrary decisions based on personal preference, your preferred ordering and mine might differ, and why should mine prevail over yours (or vice versa)? On the other hand "sorted" is generic, and deterministic and equally fair (or unfair) to everyone's preference. 

Anyway, regardless of the question about grouping and sorting order, (and I am keen to hear your thoughts), this PR contains other changes we should integrate anyway IMO. The bug with `[SsIi]` flags demonstrated that there is real value in having a canonical flag order, and the extra validation is IMO pretty useful, I've been playing around to see what different ways I can break embed.fnc by messing up a change to it and running the `make regen`. Maybe I should just extract that part and submit it in a different PR.

Cheers, and thanks for the feedback.

UPDATE:
ps: BTW, I think there is another thing to consider, I am not sure that "peephole optimizer" is a "section". It doesnt include two important functions that are part of the peephole optimizer framework, and many lines in embed.fnc have comments saying where a file is defined and used. IMO that kind of thing should be automatically generated not managed with comments. Comments get out of date, they require manual updates etc. If we want to know that kind of thing we should write a tool that generates the data. I actually tried, but I found it difficult to determine where or if a function is used. For instance we have functions that are only used in certain builds, we have functions which are used underneath macro so the code only uses a macro to access them, and some of those macros are no-ops under one build option (eg threads) and used under another. Perl_peep and Perl_rpeep are good examples. Neither are really used directly under their name. Both are used primarily via the macros CALL_PEEP() and CALL_RPEEP(). I tried a bunch of things to overcome this, for instance getting the compiler to dump the macro-expanded form, but there are too many build options to do that reasonably.

demerphq:
@leont I have poked into this a bit, and while I can and did add a way to have comments that group content, I don't think it makes that much sense as the primary grouping primitive in the file is the guard clauses around the definitions, it also means someone needs to go through the file and add group comments, and who is going to do that?

If we want to have some way to have to grouped lists of functions I think we should implement the "tags" concept, and then generate the lists. But again, until someone sets up the tags, there wouldn't be much to output. 

Having a standard "personal preference free" ordering principal in the file makes a lot of things easier. IMO it reduces the chance of conflict (one of my objectives here), and at least provides a canonical ordering that is easy to understand and discover. [Note I realize that right after this gets merged there will be some conflicts, but its a one time cost that should eliminate or reduce such costs in the future.]

This patch ensures that comments are "sticky" to the function that follows them (with the exception of the comments at the top and bottom of the file), so when they are sorted the comment will go with them. It also introduces validation and ensures the flags are consistently ordered.  If people want to add comments that add some kind of "tag information" that shows that functions are related, then they can. But right now there really aren't that many "groups" in the file. It is a mismash of different styles and ordering. So I think this should be merged as is.

Leont:
Still not entirely convinced, but it's more a -0 than a -1 so feel free to merge it if no one else feels the same way about it.

demerphq:
No further feedback, so I am going to go ahead and merge this.

PR #20765, opened 2023-02-04
Assert that an op_free()'ed op has an op_first if it has OPf_KIDS

### LINKED ISSUE #20764: `op_free()` silently does nothing if `OPf_KIDS` but `op_first == NULL`
Found while debugging a test failure of `feature-class` branch.

If you want to free the former container of an optree during reshaping, it is insufficient simply to save its child and free the parent by doing this:

```c
OP *kid = cUNOPo->op_first;
cUNOPo->op_first = NULL;
op_free(o);
```

What happens here is that the op still has the `OPf_KIDS` flag set, which causes `Perl_op_free()` to want to recurse into the children. It sets up the pointer to be the `->op_first` of the op, then goes around the while loop again. The loop finds that the pointer is now NULL, decides its job is complete, and the whole function returns having done nothing.

While this is technically the caller's fault (because no op should ever have the `OPf_KIDS` flag while its `op_first` is NULL), I feel that `op_free()` ought to do something about it. Just unsure quite what.

leonerd:
Fixes #20764

### DIFF
--- a/op.c
+++ b/op.c
@@ -898,6 +898,12 @@ Perl_op_free(pTHX_ OP *o)
         /* free child ops before ourself, (then free ourself "on the
          * way back up") */
 
+        /* Ensure the caller maintains the relationship between OPf_KIDS and
+         * op_first != NULL when restructuring the tree
+         *   https://github.com/Perl/perl5/issues/20764
+         */
+        assert(!(o->op_flags & OPf_KIDS) || cUNOPo->op_first);
+
         if (!went_up && o->op_flags & OPf_KIDS) {
             next_op = cUNOPo->op_first;
             continue;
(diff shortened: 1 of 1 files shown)


leonerd:
Now with docs added

op.c
```diff
@@ -898,6 +898,11 @@ Perl_op_free(pTHX_ OP *o)
         /* free child ops before ourself, (then free ourself "on the
          * way back up") */
 
+        /* Detect the caller being silly
+         *   https://github.com/Perl/perl5/issues/20764
+         */
```
tonycoz: I think the comment could be improved.  Perhaps:

> Ensure the caller maintains the relationship between OPf_KIDS and op_first when restructuring the tree.

demerphq:
Other than @tonycoz's suggestion (which I agree with), this LGTM.

op.c
leonerd: Ahyes; message rewritten and force-pushed.

tonycoz (approved):

demerphq (approved):

PR #20766, opened 2023-02-05
allow AUTOLOAD for the INC method of objects in @INC

### LINKED ISSUE #20665: @INC hooks ignore AUTOLOAD
`AUTOLOAD` methods on objects in `@INC` are not called by perl 5.37.7.

I'm not sure what the best behavior here is, but I see two possible solutions. First is to not change anything and just document that `AUTOLOAD` is ignored for this case. Second is checking for an `AUTOLOAD` method after checking for `INC`, `INCDIR`, and string overloads, and if it exists calling it for `INC`.

**Steps to Reproduce**
```
perl -E'package AutoHook { sub DESTROY {} sub AUTOLOAD { say our $AUTOLOAD; return; } } unshift @INC, bless {}, "AutoHook"; require strict;'
```
On perl 5.36.0 this produces:
```
AutoHook::INC
```
On perl 5.37.7 this produces no output.

tonycoz:
This matches the behaviour in 5.36.

This does not allow AUTOLOAD for INCDIR, since if there is an AUTOLOAD the check for INC would have already succeeded.

Fixes #20665

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6851,6 +6851,9 @@ Optional state for the subroutine.  The state is passed in as C<$_[1]>.
 
 =back
 
+C<AUTOLOAD> cannot be used to resolve the C<INCDIR> method, C<INC> is
+checked first, and C<AUTOLOAD> would resolve that.
+
 If an empty list, L<C<undef>|/undef EXPR>, or nothing that matches the
 first 3 values above is returned, then L<C<require>|/require VERSION>
 looks at the remaining elements of L<C<@INC>|perlvar/@INC>.
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4338,10 +4338,11 @@ S_require_file(pTHX_ SV *sv)
                          * call the method.
                          */
                         HV *pkg = SvSTASH(SvRV(loader));
-                        GV * gv = gv_fetchmethod_pvn_flags(pkg, "INC", 3, 0);
+                        GV * gv = gv_fetchmethod_pvn_flags(pkg, "INC", 3, GV_AUTOLOAD);
                         if (gv && isGV(gv)) {
                             method = "INC";
                         } else {
+                            /* no point to autoload here, it would have been found above */
                             gv = gv_fetchmethod_pvn_flags(pkg, "INCDIR", 6, 0);
                             if (gv && isGV(gv)) {
                                 method = "INCDIR";
--- a/t/op/inccode.t
+++ b/t/op/inccode.t
@@ -21,7 +21,7 @@ unless (is_miniperl()) {
 
 use strict;
 
-plan(tests => 68 + !is_miniperl() * (4 + 14 * $can_fork));
+plan(tests => 71 + !is_miniperl() * (4 + 14 * $can_fork));
 
 sub get_temp_fh {
     my $f = tempfile();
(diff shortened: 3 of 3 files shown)


demerphq:
I think if we do this we need to document that AUTOLOAD precludes the use of INCDIR.

I was under the impression we didn't support autoloader previously, but given your comment about 5.36 I guess I was wrong.

I wish we had AUTOLOADSUB as an alternative to AUTOLOAD, which would return a coderef instead of acting as a proxy for the behaviour. Then we could do this kind of thing right.

tonycoz:
> I think if we do this we need to document that AUTOLOAD precludes the use of INCDIR.

I'll add that.

> I was under the impression we didn't support autoloader previously, but given your comment about 5.36 I guess I was wrong.

The old version used call_method(), which does support like AUTOLOAD like any normal method call (via pp_method's call to gv_fetchmethod_sv_flags() with `GV_AUTOLOAD`).

> I wish we had AUTOLOADSUB as an alternative to AUTOLOAD, which would return a coderef instead of acting as a proxy for the behaviour. Then we could do this kind of thing right.

I considered calling can() on the object, which can be used to implement something like that, but that wouldn't return to the old behaviour.

But perhaps that should change.

Or just document that AUTOLOAD is no longer supported for INC in perldelta and perlfunc/require

demerphq (approved):

demerphq:
> I considered calling can() on the object, which can be used to implement something like that, but that wouldn't return to the old behaviour.

Does that work when the method is handled by AUTOLOAD directly?  I thought it didn't which is why i muttered about a hypothetical AUTOLOADSUB. Anyway, I think what you have here is fine for now. Much obliged that you picked this up, I have been avoiding this subject for a bit as I have not had time nor emotional energy to address any of it while I am visiting family. (Eg the require hook stuff.)

tonycoz:
> Does that work when the method is handled by AUTOLOAD directly?

can() could return an anon sub `return sub { local $AUTOLOAD="INC"; $_[0]->AUTOLOAD(@_[1 .. $#_]; }` or something else horrible like that.

(/me waits for the bug report from someone using AutoLoader to load the definition of INC.)

demerphq:
@tonycoz - is there a reason you haven't merged this? I would have merged it already if you weren't a committer.

demerphq:
> can() could return an anon sub return `sub { local $AUTOLOAD="INC"; $_[0]->AUTOLOAD(@_[1 .. $#_]; }` or something else horrible like that.

Unless can() must do so, it seems like it would be unreliable to me. That is why i like the AUTOLOADSUB idea, if we had one we would define it in a way that it would work with can(), and would only be called once per "name", etc. Eg, it would have stable semantics. With AUTOLOAD there is no guarantee what it might do. Or am i missing your point and we could change can() to do this when someone has an AUTOLOAD method? AUTOLOAD is cute, but it isnt very easy to reason about.

Grinnz:
> I wish we had AUTOLOADSUB as an alternative to AUTOLOAD, which would return a coderef instead of acting as a proxy for the behaviour. Then we could do this kind of thing right.

You may consider this a prototype: https://metacpan.org/pod/Autoload::AUTOCAN - there are a number of funky details that come up documented there.

demerphq:
@Grinnz thank you.

PR #20767, opened 2023-02-06
regcomp.c - remove (**{ ... }) from the regex engine

demerphq:
Dave M pointed out that this idea was flawed, and after some testing I have come to agree with him. This removes it. It was only available for 5.37.8, so no deprecation cycle involved.

### DIFF
--- a/pod/perl5378delta.pod
+++ b/pod/perl5378delta.pod
@@ -28,6 +28,9 @@ included in a pattern, so that patterns which are O(N) in normal use become
 O(N*N) with a C<(?{ ... })> pattern in them. Switching to C<(*{ ... })> means
 the pattern will stay O(N).
 
+B<NOTE> the C<(**{ ... })> was removed in 5.37.9 as it didn't quite work out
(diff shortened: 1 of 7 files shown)


demerphq:
@iabyn as promised I have pushed this to remove `(**{ ... })`

bram-perl:
Disclaimer: I didn't look at the code and I also didn't look at what `(**{... })` is/was supposed to do.

I think it would be good to improve the commit message:
- in what way was it flawed?
- what testing did you do and how did it behave? I.e. what was the unexpected behavior?
- ...

I think it should be documented somewhere; ideally in the commit message but if it gets too long then a github issue for it or a mail to p5p about it (which can then be linked in the commit) would be OK too (IMO).

The main points I like to see addressed (somewhere):
- if you/someone decides to implement this what are the cases they should look out for?
- will you/someone remember in 2 months/2 years/5 years what the issues there were with this implementation?

tonycoz (approved):

PR #20770, opened 2023-02-07
util.c - move was_here to be initialized earlier

demerphq:
otherwise the goto out_of_memory breaks g++ as reported by Karl

### DIFF
--- a/util.c
+++ b/util.c
@@ -242,6 +242,7 @@ Perl_safesysrealloc(Malloc_t where,MEM_SIZE size)
     }
     else {
         dSAVE_ERRNO;
+        PERL_DEB(UV was_where = PTR2UV(where)); /* used in diags below */
 #ifdef USE_MDH
         where = (Malloc_t)((char*)where-PERL_MEMORY_DEBUG_HEADER_SIZE);
         if (size + PERL_MEMORY_DEBUG_HEADER_SIZE < size)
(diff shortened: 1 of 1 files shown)


khwilliamson (approved):

bram-perl:
Would be good to amend the commit message to include the actual error message:
```
util.c:335:11: error: jump to label 'out_of_memory crosses initialization of 'UV was_where'
```

Also I **suspect** - not yet confirmed - that this a g++ v12 thing..
There is a CI jobs using g++ which did succeed but that appears to be using g++ v11.3.0, so it's either the compiler version or one of the extra flags khw is using..

bram-perl:
> Also I **suspect** - not yet confirmed - that this a g++ v12 thing.. There is a CI jobs using g++ which did succeed but that appears to be using g++ v11.3.0, so it's either the compiler version or one of the extra flags khw is using..

It's unrelated to g++ v12.
To reproduce it needs `-Duseithreads` and `-DDEBUGGING`
Without `-Duseithreads` the offending code isn't reached.

Just for completeness: the **original** error was:
```
 util.c: In function ‘void* Perl_safesysrealloc(void*, size_t)’:
util.c:335:11: error: jump to label ‘out_of_memory’
  335 |           out_of_memory:
      |           ^~~~~~~~~~~~~
util.c:248:18: note:   from here
  248 |             goto out_of_memory;
      |                  ^~~~~~~~~~~~~
util.c:277:12: note:   crosses initialization of ‘UV was_where’
  277 |         UV was_where = PTR2UV(where);
      |            ^~~~~~~~~
make: *** [makefile:260: util.o] Error 1
```

PR #20776, opened 2023-02-08
perl.h: Make sure PERL_IMPLICIT_CONTEXT doesn't come back

khwilliamson:
This is an obsolete name, retained for back compat with cpan.  Make sure the core doesn't have it defined.

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -79,13 +79,17 @@ is used in functions that take a thread context initial parameter.
 #endif
 
 /* PERL_IMPLICIT_CONTEXT is a legacy synonym for MULTIPLICITY */
-#ifdef MULTIPLICITY
-#  ifndef PERL_IMPLICIT_CONTEXT
-#    define PERL_IMPLICIT_CONTEXT
-#  endif
+#if defined(MULTIPLICITY)               \
+ && ! defined(PERL_CORE)                \
+ && ! defined(PERL_IMPLICIT_CONTEXT)
+#  define PERL_IMPLICIT_CONTEXT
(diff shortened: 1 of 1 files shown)


perl.h
```diff
@@ -79,14 +79,17 @@ is used in functions that take a thread context initial parameter.
...
+#if defined(MULTIPLICITY)               \
+ && ! defined(PERL_CORE)                \
+ && ! defined(PERL_IMPLICIT_CONTEXT)
+#  define PERL_IMPLICIT_CONTEXT
 #endif
 #if defined(PERL_IMPLICIT_CONTEXT) && !defined(MULTIPLICITY)
 #  define MULTIPLICITY
 #endif
+#ifdef PERL_CORE    /* Make sure doesn't creep back in to core code */
+#  undef PERL_IMPLICIT_CONTEXT
```
demerphq, line 92: maybe this should use pragma warn?

demerphq: Sorry i mean something like:

```
#ifdef PERL_IMPLICIT_CONTEXT
#praga message("PERL_IMPLICIT_CONTEXT was removed from core perl. It does not do anything. undeffing it for compilation")
#undef PERL_IMPLICIT_CONTEXT
#endif
```

demerphq (changes requested):

perl.h
khwilliamson: done

demerphq: Cool.

demerphq (approved):

PR #20780, opened 2023-02-08
Update `pop()` and `shift()` docs with examples

scottchiefbaker:
Add some simple examples to `pop()` and `shift()` and simplify some wording.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -7574,6 +7574,9 @@ is C<undef>.
     my @arr  = (undef, 'two', 'three');
     my $item = shift(@arr); # undef
 
+C<shift> will operate on the C<@ARGV> array in C<eval STRING>,
+C<BEGIN {}>, C<INIT {}>, C<CHECK {}> blocks.
+
 Starting with Perl 5.14, an experimental feature allowed
 L<C<shift>|/shift ARRAY> to take a
 scalar expression. This experiment has been deemed unsuccessful, and was
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6055,6 +6055,9 @@ is C<undef>.
     my @arr  = ('one', 'two', undef);
     my $item = pop(@arr); # undef
 
+C<pop> will operate on the C<@ARGV> array in C<eval STRING>,
+C<BEGIN {}>, C<INIT {}>, C<CHECK {}> blocks.
+
 Starting with Perl 5.14, an experimental feature allowed
 L<C<pop>|/pop ARRAY> to take a
 scalar expression. This experiment has been deemed unsuccessful, and was
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -6045,9 +6045,7 @@ one element.
 
     # @arr is now ('cat', 'dog')
 
-Returns C<undef> if the array is empty. If ARRAY is omitted, C<pop> operates on
-the L<C<@ARGV>|perlvar/@ARGV> array in the main program, but the
-L<C<@_>|perlvar/@_> array in subroutines.
+Returns C<undef> if the array is empty.
 
 B<Note:> C<pop> may also return C<undef> if the last element in the array
 is C<undef>.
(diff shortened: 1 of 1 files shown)


pod/perlfunc.pod
```diff
@@ -7547,14 +7556,23 @@ X<shift>
 
 =for Pod::Functions remove the first element of an array, and return it
 
-Shifts the first value of the array off and returns it, shortening the
-array by 1 and moving everything down.  If there are no elements in the
-array, returns the undefined value.  If ARRAY is omitted, shifts the
-L<C<@_>|perlvar/@_> array within the lexical scope of subroutines and
-formats, and the L<C<@ARGV>|perlvar/@ARGV> array outside a subroutine
-and also within the lexical scopes
-established by the C<eval STRING>, C<BEGIN {}>, C<INIT {}>, C<CHECK {}>,
```
demerphq, line 7556: I think this part of the documentation was added because many people view these special blocks as special forms of subroutine. Also there is ambiguity as to what happens when a BEGIN (for instance) is embedded inside of a sub. So I don't think it should be entirely removed, but at the same time I think your text is better overall. Maybe we can add a special subsection that covers both shift and unshift and explains this? Eg, consider

```
perl -le'sub test { BEGIN { print "BEGIN:", pop } print "sub test: ", pop } test(42)' argv
BEGIN:argv
sub test: 42
```

demerphq (changes requested):

scottchiefbaker:
@demerphq good call. The original text was kind of convoluted so I removed the section on `BEGIN`, `INIT`, and `CHECK`. I went ahead and added back that section on the **bottom** so it's less confusing to novices.

demerphq:
@scottchiefbaker i think you only added it for shift, but not for pop. But im a bit tired and maybe I am misreading it.

scottchiefbaker:
@demerphq updated with `BEGIN` stuff for `pop()` as well.

scottchiefbaker:
What else needs to change for this to land?

pod/perlfunc.pod
```diff
@@ -7547,14 +7559,26 @@ X<shift>
...
+array by one and moves everything down.
+
+    my @arr  = ('cat', 'dog');
+    my $item = shift(@arr); # 'cat'
+
+    # @arr is now ('dog');
+
+Returns C<undef> if the array is empty. If ARRAY is omitted, C<shift>
+operates on the C<@ARGV> array in the main program, and the C<@_> array
+in subroutines.
```
demerphq: I think i  would move the "If ARRAY is ommitted" part down to join the part about eval and the special blocks. 

But what I was really wondering is if you can somehow eliminate the need to duplicate this bit of docs, by maybe making one of the two refer to the other.

Eg, Perhaps "shift" would say 

```
If ARRAY is omitted, C<shift> operates on the the C<@_> array
in normal subroutines, and on the C<@ARGV> array in the main program, and 
in C<eval STRING>, C<BEGIN {}>,  C<INIT {}>, and C<CHECK {}> blocks.
```

And then in the docs for pop it could say

```
If ARRAY is omitted, C<pop> defaults to using either C<@_> or C<@ARGV>, 
exactly as C<shift> does, see L</shift> for details.
```
(module a correct L<> link for shift of course).

Sorry it took me so long to get back to you, I was out of action for a while

Grinnz: I don't think it's necessarily worth consolidating with a cross link just to save a line of description. For small things it can be better to just write what people need to know in the place they need to know it.

demerphq: I dont feel very strongly about cross linking or not, but i do feel strongly that the different bits of docs for the "no array specifed" case should be in one sentence, and not part of a sentence referencing what happens if the array is empty.  

The cross linking suggestion was mostly to avoid the two descriptions getting out of sync with each other as I think they were previously, and to avoid double bookkeeping, but I don't feel that strongly about it.

demerphq:
btw, @scottchiefbaker if I have annoyed you with my feedback on this please feel free to return the favour in https://github.com/Perl/perl5/pull/20823 :-) Docs are hard. but its valuable to get them right! Your feedback on readability appreciated.

demerphq:
This has been stalled for quite a while. It would be a shame if 5.38 didnt include this doc tweak. @scottchiefbaker do you want me to pick up the loose ends?

scottchiefbaker:
@demerphq thanks for reminding me this was pending. Re-reading things, I agree with your conclusion. I moved the "if ARRAY is omitted" sections to the bottom as they're not the important part of that documentation and shouldn't be the first thing a novice reads. PR has been updated.

pod/perlfunc.pod
scottchiefbaker: I agree with moving things down... I updated the PR.

demerphq (approved):

PR #20781, opened 2023-02-09
document deprecation of ' as the package separator

tonycoz:
7e2d91e6d3a0 and a36fec492e made use on ' instead of :: as deprecated, but I forgot to include documentation, document this change in the reference documentation.

perldelta was updated in dcf36a4fb9625.

### DIFF
--- a/pod/perldata.pod
+++ b/pod/perldata.pod
@@ -21,8 +21,8 @@ structure it refers.  The rest of the name tells you the particular
 value to which it refers.  Usually this name is a single I<identifier>,
 that is, a string beginning with a letter or underscore, and
 containing letters, underscores, and digits.  In some cases, it may
-be a chain of identifiers, separated by C<::> (or by the slightly
-archaic C<'>); all but the last are interpreted as names of packages,
+be a chain of identifiers, separated by C<::> (or by the deprecated C<'>);
+all but the last are interpreted as names of packages,
 to locate the namespace in which to look up the final identifier
 (see L<perlmod/Packages> for details).  For a more in-depth discussion
 on identifiers, see L</Identifier parsing>.  It's possible to
@@ -137,11 +137,12 @@ That is, any word character in the ASCII range, as long as the first
 character is not a digit.
 
 There are two package separators in Perl: A double colon (C<::>) and a single
-quote (C<'>).  Normal identifiers can start or end with a double colon, and
-can contain several parts delimited by double colons.
-Single quotes have similar rules, but with the exception that they are not
-legal at the end of an identifier: That is, C<$'foo> and C<$foo'bar> are
-legal, but C<$foo'bar'> is not.
+quote (C<'>).  Use of C<'> as the package separator is deprecated and will be
+removed in Perl 5.40.  Normal identifiers can start or end with a double
+colon, and can contain several parts delimited by double colons.  Single
+quotes have similar rules, but with the exception that they are not legal at
+the end of an identifier: That is, C<$'foo> and C<$foo'bar> are legal, but
+C<$foo'bar'> is not.
 
 Additionally, if the identifier is preceded by a sigil --
 that is, if the identifier is part of a variable name -- it
--- a/pod/perldeprecation.pod
+++ b/pod/perldeprecation.pod
@@ -37,6 +37,14 @@ disabled I<even if you had previously written> C<use strict>. This was not
 the previous behaviour of C<use VERSION>, which at present will track
 explicitly-enabled strictness flags independently.
 
+=head3 Use of C<'> as a global name separator.
+
+Perl allows use of C<'> instead of C<::> to replace the parts of a
+package or global variable name, for example C<A::B> and C<A'B> are
+equivalent.
+
+C<'> will no longer be recognized as a name separator in Perl 5.40.
+
 =head2 Perl 5.38
 
 =head3 Pod::Html utility functions
--- a/pod/perlmod.pod
+++ b/pod/perlmod.pod
@@ -81,6 +81,9 @@ the $s variable in package C<owner>, which is probably not what you meant.
 Use braces to disambiguate, as in C<"This is ${owner}'s house">.
 X<::> X<'>
 
+Using C<'> as a package separator is deprecated and will be removed in
+Perl 5.40.
+
 Packages may themselves contain package separators, as in
 C<$OUTER::INNER::var>.  This implies nothing about the order of
 name lookups, however.  There are no relative packages: all symbols


scottchiefbaker:
This looks fine to me with one clarifying question:

> Normal identifiers can start or end with a double colon, and can contain several parts delimited by double colons.

Does that mean `$::foo::bar` and `$foo::bar::` are somehow valid? Maybe that's syntax I'm not aware of?

Grinnz:
Yes, `$::foo::bar` is equivalent to `$main::foo::bar`, and `$foo::bar::` is valid syntax for something or other.

demerphq:
> Does that mean `$::foo::bar` and $foo::bar:: are somehow valid?

Yes, both are. 

`$::foo::bar` is shorthand for the package variable `$main::foo::bar`.

The second one refers to the SCALAR slot of the `*foo::bar::` glob, which is used as part of the "stash" for the package `foo::bar`. It's a weird one, you don't see that much in scalar form, although the hash form is a little more common as it holds all the globs which represent symbols for the package. 

Globs contain one slot per type that perl can represent, and can be dereferenced somewhat like a hash. You can use it to tell if a package contains a specific symbol with a given type (except for scalar where for historic reasons it will always say it exists).

```
perl -MData::Dumper -E'$Data::Dumper::Sortkeys=1; print Dumper(\%Data::Dumper::); 
say "SCALAR=", ${*Data::Dumper::Sortkeys{SCALAR}}; 
say "*whatever{SCALAR}=", *whatever{SCALAR};'
$VAR1 = {
          'BEGIN' => *Data::Dumper::BEGIN,
          'Bless' => *Data::Dumper::Bless,
          'DESTROY' => *Data::Dumper::DESTROY,
          'Deepcopy' => *Data::Dumper::Deepcopy,
          'Deparse' => *Data::Dumper::Deparse,
          'Dump' => *Data::Dumper::Dump,
          'Dumper' => *Data::Dumper::Dumper,
          'DumperX' => *Data::Dumper::DumperX,
          'Dumpperl' => *Data::Dumper::Dumpperl,
          'Dumpxs' => *Data::Dumper::Dumpxs,
          'EXPORT' => *Data::Dumper::EXPORT,
          'EXPORT_OK' => *Data::Dumper::EXPORT_OK,
          'Freezer' => *Data::Dumper::Freezer,
          'ISA' => *Data::Dumper::ISA,
          'IS_PRE_516_PERL' => \'',
          'Indent' => *Data::Dumper::Indent,
          'Maxdepth' => *Data::Dumper::Maxdepth,
          'Maxrecurse' => *Data::Dumper::Maxrecurse,
          'Names' => *Data::Dumper::Names,
          'Pad' => *Data::Dumper::Pad,
          'Pair' => *Data::Dumper::Pair,
          'Purity' => *Data::Dumper::Purity,
          'Quotekeys' => *Data::Dumper::Quotekeys,
          'Reset' => *Data::Dumper::Reset,
          'Seen' => *Data::Dumper::Seen,
          'Sortkeys' => *Data::Dumper::Sortkeys,
          'Sparseseen' => *Data::Dumper::Sparseseen,
          'Terse' => *Data::Dumper::Terse,
          'Toaster' => *Data::Dumper::Toaster,
          'Trailingcomma' => *Data::Dumper::Trailingcomma,
          'Useperl' => *Data::Dumper::Useperl,
          'Useqq' => *Data::Dumper::Useqq,
          'VERSION' => *Data::Dumper::VERSION,
          'Values' => *Data::Dumper::Values,
          'Varname' => *Data::Dumper::Varname,
          '_bad_vsmg' => \'',
          '_compose_out' => *Data::Dumper::_compose_out,
          '_dump' => *Data::Dumper::_dump,
          '_quote' => *Data::Dumper::_quote,
          '_refine_name' => *Data::Dumper::_refine_name,
          '_sortkeys' => *Data::Dumper::_sortkeys,
          '_vstring' => *Data::Dumper::_vstring,
          'bootstrap' => *Data::Dumper::bootstrap,
          'format_refaddr' => *Data::Dumper::format_refaddr,
          'import' => *Data::Dumper::import,
          'init_refaddr_format' => *Data::Dumper::init_refaddr_format,
          'new' => *Data::Dumper::new,
          'post' => *Data::Dumper::post,
          'qquote' => *Data::Dumper::qquote
        };
SCALAR=1
*whatever{SCALAR}=SCALAR(0x56101ee905e8)
```

No doubt one of our documentation experts will follow up with a pointer the specific docs for this.

```
perl -E'$foo::bar::= "blorp"; say ${*foo::bar::{SCALAR}}; say $foo::bar::'
blorp
blorp
```

pod/perldata.pod
```diff
@@ -137,11 +137,12 @@ That is, any word character in the ASCII range, as long as the first
 character is not a digit.
 
 There are two package separators in Perl: A double colon (C<::>) and a single
-quote (C<'>).  Normal identifiers can start or end with a double colon, and
-can contain several parts delimited by double colons.
-Single quotes have similar rules, but with the exception that they are not
-legal at the end of an identifier: That is, C<$'foo> and C<$foo'bar> are
-legal, but C<$foo'bar'> is not.
+quote (C<'>).  Use of C<'> as the package separator is deprecated and will be
+removed in Perl 5.40.  Normal identifiers can start or end with a double
```
demerphq, line 141: Maybe it would be better to say "Normal identifiers can start with a double colon, (in which case the leading double colon is short hand for 'main::'), and trailing colons reference the glob or slots of the glob which represents the package of the name without the trailing colons."

Or perhaps something a little less clumsy that gets to the same point.

IOW, I would not consider a variable with a trailing colon to be a "normal identifier". It isnt. `$foo::bar::` is another way to access `${*foo::bar::{SCALAR}}`, and doesn't have a normal identifier name. Afaik the only way to address this variable is via its trailing double colons. Compare with `$foo::bar::baz`, which represents the SCALAR slot of the `baz` glob in `%foo::bar::`, which itself represents the hash slot of the glob `*foo::bar::`.

```
perl -le'$foo::bar::baz = "baz"; $foo::bar::= "BLORP"; print for keys %foo::bar::; print "---"; print for keys %foo::; print "---"; print ${*foo::bar::{SCALAR}}'
baz
---
bar::
---
BLORP
```

demerphq (approved):
I'll leave it to you if you want to change the wording on this, but i think you should to call out that a  name ending in `::` is not a normal identifier, it refers to a glob used to represent a stash, unlike a glob which is used to represent a symbol in a stash.

scottchiefbaker:
Today I learned that `$::foo` is a shortcut for main!

rwp0:
> Today I learned that `$::foo` is a shortcut for main!

Yeah main is the default package name.

pod/perldata.pod
tonycoz: I made no changes to that wording, I only word-wrapped the resulting paragraph after adding the deprecation text.

PR #20785, opened 2023-02-09
File-Find - set up tempdir early and test that we could chdir to it

demerphq:
If for some reason we die very early in the test script the cleanup() function would get called before we had set up $test_root_dir or $test_temp_dir. This then lead to further errors being generated by trying to chdir into an undefined directory. This patch ensures that the various setup behavior worked correctly, and that if it does not that we have some clear diagnostics about it.

### DIFF
--- a/ext/File-Find/t/taint.t
+++ b/ext/File-Find/t/taint.t
@@ -28,6 +28,7 @@ use Testing qw(
     symlink_ok
     dir_path
     file_path
+    _cleanup_start
 );
 use Errno ();
 use Config;
@@ -36,7 +37,7 @@ use File::Temp qw(tempdir);
 BEGIN {
     plan(
         ${^TAINT}
-        ? (tests => 45)
+        ? (tests => 48)
         : (skip_all => "A perl without taint support")
     );
 }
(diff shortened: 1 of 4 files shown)


ext/File-Find/t/taint.t
```diff
@@ -87,22 +98,37 @@ is($found, 1, 'taint.t found once again');
...
-    $test_root_dir = $1;
-} else {
-    die "Failed to untaint root dir of test";
+my $chdir_error = "";
+chdir($test_temp_dir)
+    or $chdir_error= "Failed to chdir to '$test_temp_dir': $!";
+is($chdir_error,"","chdir to temp dir '$test_temp_dir' successful")
+    or die $chdir_error;
+
+sub _something_wrong {
```
jkeenan: sub `_something_wrong()` appears to be defined in each of two test files.  Could it be moved to `t/lib/Testing.pm` instead?

demerphq: Thanks. Good suggestion. Done. :-)

demerphq: And actually I went a step further and moved the identical code from both into Testing.pm as well. Good catch.

atoomic (approved):

PR #20786, opened 2023-02-09
Add examples and clarify wording for `ucfirst()` docs

scottchiefbaker:


pod/perlfunc.pod
```diff
@@ -9611,17 +9611,21 @@ X<ucfirst> X<uppercase>
...
-If EXPR is omitted, uses L<C<$_>|perlvar/$_>.
+    my $str = ucfirst("hello world!"); # "Hello world!"
 
 This function behaves the same way under various pragmas, such as in a locale,
 as L<C<lc>|/lc EXPR> does.
 
+B<Note:> This is the internal function implementing the C<\u> escape in
+double-quoted strings.
+
+    my $str = "\uperl\E is great"; # "Perl is great"
```
atoomic, line 9627: excellent!

pod/perlfunc.pod
```diff
@@ -9611,17 +9611,23 @@ X<ucfirst> X<uppercase>
...
 =item ucfirst
 
-=for Pod::Functions return a string with just the next letter in upper case
+=for Pod::Functions return a string with the first letter in upper case
 
-Returns the value of EXPR with the first character in uppercase
-(titlecase in Unicode).  This is the internal function implementing
-the C<\u> escape in double-quoted strings.
+Returns the value of EXPR with the B<first> character in uppercase
+(titlecase in Unicode).
```
khwilliamson: How about

"Unicode calls this "titlecase"

khwilliamson (approved):

pod/perlfunc.pod
scottchiefbaker: That wording is definitely better. I updated the PR.

Do we **really** even need that clarification? I've never heard of titlecase before looking at these docs.

khwilliamson:
I do think we need to educate people with regard to Unicode speak; or for those who already do know it, to point out what we do that conforms to it.

pod/perlfunc.pod
demerphq: Afaik we do. Uppercase as term doesn't make sense in all scripts, some don't have an upper/lower distintiction, but do have a "titlecase" that is used to distinguish headings and titles from normal script. It's a bit like English and grammar, English grammar is fairly simple compared to Greek or German or what not, so it's not a great model to understand language constructs. Same goes for scripts, the English alphabrt has relatively few characters and relatively simple rules. Other scripts are much more complex, Arabic for instance. Hence unicode uses a more complex set of terms than western speakers are used to. "Foldcase" for instance is the correct way to normalize text for case insensitive comparisons which should work for any script (but turkish - for annoying historical reasons), whereas in English you can uppercase or lowercase everything to do the same task. Do that to Arabic and it doesn't work out correctly. Another example comes from Germanic languages, \xDF (latin-sharp-ess) matches "ss" case insensitively, but the uppercase of "s" is "S", and the lowercase of "S" is "s", and these days \xDF has its own uppercase (iirc used in Austrian German but not in Hochdeutsche). Anything that involves humans is messy. English is greatly simplified due to its relatively recent (500 years or so) historical origin as a patois of Norman French and Old Saxon. Most languages that originate as pidgin/patois are relatively simple compared to their donor languages.

PR #20787, opened 2023-02-10
sv.h - Update type comments to reflect implementation realities

### LINKED ISSUE #20698: sv.h: outdated comment block re: what each SV type can store
**Description**

A comment block towards the top of _sv.h_ says:
```
...
C<SVt_NULL> can only hold undef.  C<SVt_IV> can hold undef, an integer, or a
reference.  (C<SVt_RV> is an alias for C<SVt_IV>, which exists for backward
compatibility.)  C<SVt_NV> can hold any of those or a double.  C<SVt_PV> can only
hold C<undef> or a string.  C<SVt_PVIV> is a superset of C<SVt_PV> and C<SVt_IV>.
C<SVt_PVNV> is similar.  C<SVt_PVMG> can hold anything C<SVt_PVNV> can hold, but it
can, but does not have to, be blessed or magical.
```
Two things seems amiss when reading that:
1. PVs can hold `undef`, a string, or a **reference**. It seems like a simple comment update is needed here?
2. It reads like an NV can hold undef, an integer, a reference, or a double. However, `Perl_sv_setsv_flags` seems to only account for it holding either undef or a double.

e.g. `perl -MDevel::Peek -e 'my $y = 3.3; $y = {}; Dump($y)'` reveals that `$y` is upgraded to a PVNV to store the reference, but if NV's can store references, surely no upgrade should be required?

So I'm unclear here whether the comment block needs updating to match the realities of the implementation, or the implementation needs updating so that NVs actually *can* store integers and references.

**Perl configuration**
Applies to blead and at least the past few stable releases.

richardleach:
Closes #20698

### DIFF
--- a/sv.h
+++ b/sv.h
@@ -67,10 +67,13 @@ PVMG, we save memory by allocating smaller structs when possible.  All the
 other types are just simpler forms of C<SVt_PVMG>, with fewer internal fields.
 C<SVt_NULL> can only hold undef.  C<SVt_IV> can hold undef, an integer, or a
 reference.  (C<SVt_RV> is an alias for C<SVt_IV>, which exists for backward
-compatibility.)  C<SVt_NV> can hold any of those or a double.  C<SVt_PV> can only
-hold C<undef> or a string.  C<SVt_PVIV> is a superset of C<SVt_PV> and C<SVt_IV>.
-C<SVt_PVNV> is similar.  C<SVt_PVMG> can hold anything C<SVt_PVNV> can hold, but it
-can, but does not have to, be blessed or magical.
+compatibility.)  C<SVt_NV> can hold undef or a double. (In builds that support
+headless NVs, these could also hold a reference via a suitable offset, in the
+same way that SVt_IV does, but this is not currently supported and seems to
+be a rare use case.) C<SVt_PV> can hold C<undef>, a string, or a reference.
+C<SVt_PVIV> is a superset of C<SVt_PV> and C<SVt_IV>. C<SVt_PVNV> is similar.
+C<SVt_PVMG> can hold anything C<SVt_PVNV> can hold, but it can, but does not
+have to, be blessed or magical.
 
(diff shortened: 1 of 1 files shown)


sv.h
```diff
@@ -67,10 +67,13 @@ PVMG, we save memory by allocating smaller structs when possible.  All the
...
-hold C<undef> or a string.  C<SVt_PVIV> is a superset of C<SVt_PV> and C<SVt_IV>.
-C<SVt_PVNV> is similar.  C<SVt_PVMG> can hold anything C<SVt_PVNV> can hold, but it
-can, but does not have to, be blessed or magical.
+compatibility.)  C<SVt_NV> can hold undef or a double. (In builds that support
+headless NVs, these could also hold a reference via a suitable offset, in the
+same way that SVt_IV does, but this is not currently supported and seems to
+be a rare use case.) C<SVt_PV> can hold C<undef>, a string, or a reference.
+C<SVt_PVIV> is a superset of C<SVt_PV> and C<SVt_IV>. C<SVt_PVNV> is similar.
+C<SVt_PVMG> can hold anything C<SVt_PVNV> can hold, but it can, but does not
+have to, be blessed or magical.
```
demerphq: The last two sentences read a little bit awkward. Maybe:

`C<SVt_PVNV> is similar as it is a superset of C<SVt_PV> and C<SVt_NV>.
C<SVt_PVMG> can hold anything that a C<SVt_PVNV> can, but may also be blessed or magical as well.`

im not sure that "hold" is the best word for this doc, but its not the worst either. I think id use "represent" or "store" but its a matter of preference I guess.

demerphq (changes requested):

sv.h
richardleach: They are both the pre-existing text, but I'm happy to make changes.

demerphq (approved):

khwilliamson (approved):

Issue #899, opened 1999-11-28
The value of pattern match vars. after `next'


p5pRT:
Migrated from [rt.perl.org#1833](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=1833) (status was 'open')

Searchable as RT1833$

p5pRT:
### From inaba@st.rim.or.jp

<details><summary>Inline Patch</summary>

```diff
--- cop.h.org	Sun Mar 28 02:57:30 1999
+++ cop.h	Thu Nov 25 00:54:29 1999
@@ -205,8 +205,9 @@
 	PL_stack_sp	 = PL_stack_base + cx->blk_oldsp,		\
 	PL_markstack_ptr = PL_markstack + cx->blk_oldmarksp,		\
 	PL_scopestack_ix = cx->blk_oldscopesp,				\
-	PL_retstack_ix	 = cx->blk_oldretsp,				\
-	PL_curpm         = cx->blk_oldpm
+	PL_retstack_ix	 = cx->blk_oldretsp
+
+#define RESTOREPM(cxix) PL_curpm = cxstack[cxix+1].blk_oldpm
 
 /* substitution context */
 struct subst {
--- pp_ctl.c.org	Sun Mar 28 02:56:24 1999
+++ pp_ctl.c	Thu Nov 25 01:13:20 1999
@@ -1757,9 +1757,10 @@
 	if (cxix < 0)
 	    DIE("Label not found for \"next %s\"", cPVOP->op_pv);
     }
-    if (cxix < cxstack_ix)
+    if (cxix < cxstack_ix) {
 	dounwind(cxix);
-
+	RESTOREPM(cxix);
+    }
     TOPBLOCK(cx);
     oldsave = PL_scopestack[PL_scopestack_ix - 1];
     LEAVE_SCOPE(oldsave);
@@ -1782,9 +1783,10 @@
 	if (cxix < 0)
 	    DIE("Label not found for \"redo %s\"", cPVOP->op_pv);
     }
-    if (cxix < cxstack_ix)
+    if (cxix < cxstack_ix){
 	dounwind(cxix);
-
+	RESTOREPM(cxix);
+    }
     TOPBLOCK(cx);
     oldsave = PL_scopestack[PL_scopestack_ix - 1];
     LEAVE_SCOPE(oldsave);
@@ -1890,8 +1892,10 @@
 	    cxix = dopoptosub(cxstack_ix);
 	    if (cxix < 0)
[...]

PR #20791, opened 2023-02-10
perlvar.pod - add "Scoping Rules of Regex Variables" section

demerphq:
This section is used to document the majority of the regex variables, and previous language referring to them as "dynamically scoped" has been changed or simplified and a link to the new section provided to centralize the explanation.

Strictly speaking $1 is globally scoped, but the data it access is dynamically scoped such that successful matches behave as though they localize a regex match state variable. (Maybe one day we will actually have such a variable exposed to the user.)

This patch addss links to the relevant docs in perlsyn and perlvar to various places where it seemed appropriate, and also cleans up the wording for most cases to be similar or identical across all uses. It also cleans up a bit of related language in nearby paragraphs where it seemed to improve the readability of the docs.

It also replaces the older kind of confusing example code for understanding the behavior and documents that "goto LABEL" does not play nicely with the dynamic scoping.

This fixes Github Issue #899.

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -888,49 +888,77 @@ command or referenced as a file.
 =head2 Variables related to regular expressions
 
 Most of the special variables related to regular expressions are side
-effects.  Perl sets these variables when it has a successful match, so
-you should check the match result before using them.  For instance:
+effects. Perl sets these variables when it has completed a match
+successfully, so you should check the match result before using them.
+For instance:
 
     if( /P(A)TT(ER)N/ ) {
         print "I found $1 and $2\n";
     }
 
-These variables are read-only and dynamically-scoped, unless we note
-otherwise.
+These variables are read-only and behave similarly to a dynamically
+scoped variable, with only a few exceptions which are explicitly
+documented as behaving otherwise.  See the following section for more
+details.
 
-The dynamic nature of the regular expression variables means that
-their value is limited to the block that they are in, as demonstrated
-by this bit of code:
+=head3 Scoping Rules of Regex Variables
 
-    my $outer = 'Wallace and Grommit';
-    my $inner = 'Mutt and Jeff';
+Regular expression variables allow the programmer to access the state of
+the most recent I<successful> regex match in the current dynamic scope.
 
-    my $pattern = qr/(\S+) and (\S+)/;
+The variables themselves are global and unscoped, but the data they
+access is scoped similarly to dynamically scoped variables, in that
+every successful match behaves as though it localizes a global state
+object to the current block or file scope.
+(See L<perlsyn/"Compound Statements"> for more details on dynamic
+scoping and the C<local> keyword.)
 
-    sub show_n { print "\$1 is $1; \$2 is $2\n" }
+A I<successful match> includes any successful match performed by the
+search and replace operator C<s///> as well as those performed by the
+C<m//> operator.
 
-    {
-    OUTER:
-        show_n() if $outer =~ m/$pattern/;
+Consider the following code:
 
-        INNER: {
-            show_n() if $inner =~ m/$pattern/;
-        }
-
-        show_n();
+    my @state;
+    sub matchit {
+        push @state, $1;            # pushes "baz"
+        my $str = shift;
+        $str =~ /(zat)/;            # matches "zat"
+        push @state, $1;            # pushes "zat"
     }
 
-The output shows that while in the C<OUTER> block, the values of C<$1>
-and C<$2> are from the match against C<$outer>.  Inside the C<INNER>
-block, the values of C<$1> and C<$2> are from the match against
-C<$inner>, but only until the end of the block (i.e. the dynamic
-scope).  After the C<INNER> block completes, the values of C<$1> and
-C<$2> return to the values for the match against C<$outer> even though
-we have not made another match:
+    {
+        $str = "foo bar baz blorp zat";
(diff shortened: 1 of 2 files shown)


pod/perlvar.pod
```diff
@@ -984,14 +1012,14 @@ find uses of these problematic match variables in your code.
...
-parentheses from the last successful pattern match, not counting patterns
-matched in nested blocks that have been exited already.
+parentheses from the last successful pattern match in the current
+dynamic scope. (See L</Scoping Rules of Regex Variables>.)
 
 Note there is a distinction between a capture buffer which matches
 the empty string a capture buffer which is optional. Eg, C<(x?)> and
 C<(x)?> The latter may be undef, the former not.
 
-These variables are read-only and dynamically-scoped.
```
Grinnz, line 994: I feel like there isn't much benefit to remove "and dynamically-scoped" from all of these. It provides additional quick information for someone who has seen the definition before, and something to look up otherwise.

Grinnz: I suppose you're removing it because the variable itself is global, it just gets set in a dynamic scope, so I'm fine with it either way.

Grinnz: The reason I think it's fine to refer to these variables as "dynamically-scoped" themselves is because all globals exist by being used, so their effective scope is just where they are usable. But it's not the variables themselves that cause the dynamic scoping of their values.

demerphq:
On Fri, 10 Feb 2023, 18:02 Dan Book, ***@***.***> wrote:

>
> In pod/perlvar.pod
> <https://github.com/Perl/perl5/pull/20791#discussion_r1102542736>:
>
> >
>  Note there is a distinction between a capture buffer which matches
>  the empty string a capture buffer which is optional. Eg, C<(x?)> and
>  C<(x)?> The latter may be undef, the former not.
>
> -These variables are read-only and dynamically-scoped.
>
> The reason I think it's fine to refer to these variables as
> "dynamically-scoped" themselves is because all globals exist by being used,
> so their effective scope is just where they are usable. But it's not the
> variables themselves that cause the dynamic scoping of their values.
>

Check the discussion on the ticket.  I'd say that "dynamically scoped" is
an unsatisfactory description, although a least worst description in a
pinch, campared to others, but i think a better description is warranted.

The vars themselves are actually globals, and read only, so they can't be
directly localized, and since they only change on successful match they
don't really play the same way as a package variable would in the face of
local. The closest would be something like

local $1 = /.../ ? $1_new : $1;

The other thing is that "dynamically scoped" is a bit ambiguous in that a
package var is dynamically scoped, but the dynamic effects only kick in
when the user explicitly uses local.

Reading the ticket this caused confusion to others and I agree that linking
through to a proper description is better than repeating an only somewhat
correct description.

Another point was that many of the vars had different descriptions etc, but
they all have the same behaviour. (All that I changed anyway). Some of them
said "block scoped", some of them mentioned eval and various edge cases
etc. I think having them all references the same rules and avoid half
correct terms is a better approach.

Cheers
Yves

khwilliamson (approved):

pod/perlvar.pod
demerphq: @grinnz: I decided to go with a compromise, and changed the wording to "and its value is dynamically scoped". (Do I need the hyphen?)

demerphq:
@khwilliamson I tweaked the verbiage a touch more, are you still ok with this?

pod/perlvar.pod
Grinnz: I havent reviewed the patch fully yet (at a glance it looks fine to me) but that works for me.

pod/perlvar.pod
```diff
@@ -1209,18 +1214,24 @@ Mnemonic: the (possibly) Nested parenthesis that most recently closed.
...
-submatches in the currently active dynamic scope.
-(See L</Scoping Rules of Regex Variables>).
+match and any matching capture buffers that the pattern contains.
+(See L</Scoping Rules of Regex Variables>)
+
+The number of elements it contains will be one more than the number
+of capture buffers in the pattern, regardless of which capture buffers
+actually matched. You can use this to determine how many capture
+buffers there are in the pattern. (As opposed to C<@-> which may
+have less elements.)
```
khwilliamson: strict English grammar is 'fewer elements", though fewer and fewer people care

pod/perlvar.pod
```diff
@@ -1257,28 +1268,33 @@ surprising.
...
 =item @-
 X<@-> X<@LAST_MATCH_START>
 
+This array holds the offsets of the beginnings of the last successful
+match and any capture buffers it contains.
+(See L</Scoping Rules of Regex Variables>).
+
+The number of elements it contains will be one more than the number of
+the highest capture buffers (also called a subgroup) that actually
+matched something. (As opposed to C<@+> which may have less elements.)
```
khwilliamson: strict English grammar is 'fewer elements", though less and less people care

Note typo "addss" in first commit message.

khwilliamson (approved):

pod/perlvar.pod
demerphq: I shall fix both so I have fewer typos credited to my name.

pod/perlvar.pod
demerphq: done. :-)

pod/perlvar.pod
demerphq: Done.

PR #20793, opened 2023-02-10
Prefer SvREFCNT_dec() to calling sv_free()

richardleach:
Nowadays, `Perl_sv_free()`  just does `SvREFCNT_dec()`, which is a macro wrapper for the simple inline
function `Perl_SvREFCNT_dec()`.

The first commit in this PR updates the apidoc for `Perl_sv_free` to reflect the reality that `SvREFCNT_dec`
is no longer a wrapper for `Perl_sv_free` and is probably now the preferred option.

The second commit replaces a few uses of `sv_free()` with `SvREFCNT_dec` in toke.c. Some `ASSUME()`
statements have been used to help compilers strip out some unnecessary checks in `Perl_SvREFCNT_dec`.
There may be some scope to improve some of the `ASSUMES()`.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -6584,8 +6584,7 @@ and free the body itself.  The SV's head is I<not> freed, although
 its type is set to all 1's so that it won't inadvertently be assumed
 to be live during global destruction etc.
 This function should only be called when C<REFCNT> is zero.  Most of the time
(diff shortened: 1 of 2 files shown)


leonerd (approved):
LGTM

PR #20797, opened 2023-02-10
S_do_chomp: remove unused svrecode SV

richardleach:
Some compilers might well be able to optimize this out, but might as well take it out completely.

### DIFF
--- a/pp.c
+++ b/pp.c
@@ -738,7 +738,6 @@ S_do_chomp(pTHX_ SV *retval, SV *sv, bool chomping)
     if (chomping) {
         if (s && len) {
             char *temp_buffer = NULL;
-            SV *svrecode = NULL;
             s += --len;
             if (RsPARA(PL_rs)) {
                 if (*s != '\n')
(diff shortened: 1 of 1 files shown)


pp.c
```diff
@@ -771,7 +770,7 @@ S_do_chomp(pTHX_ SV *retval, SV *sv, bool chomping)
                                is the buffer inside PL_rs, so don't free it.
                              */
                             assert (temp_buffer == rsptr);
-                            goto nope_free_sv;
+                            goto nope_free_nothing;
```
khwilliamson, line 773: I don't think this line belongs in this PR.  I haven't worked it out, but the comments seem to me to indicate that the current code is wrong, and the change is correct.  But I don't understand fully.

The rest of the PR is fine

demerphq: He removed svrecode, so nope_free_sv no longer makes sense (I think) as it seems to exist only to free svrecode. Am I missing something?

demerphq (approved):

pp.c
richardleach: Yes, that was the rationale and I thought that compilers might warn about an unused `nope_free_sv`.

FWIW `svrevcode` did used to get used, but that stopped being the case with https://github.com/Perl/perl5/commit/8df0e7a28b22521824d9a5d0d44f2b27d2955298

demerphq: Nod. I think its fine to merge this. 

Yves

Issue #7579, opened 2004-11-04
grep {/PATTERN/} is slow


p5pRT:
Migrated from [rt.perl.org#32331](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=32331) (status was 'open')

Searchable as RT32331$

p5pRT:
### From @abigail

#### Created by @abigail



I would expect

&nbsp;    grep \{EXPR\} LIST

to be somewhat slower than 

&nbsp;    grep  EXPR\, LIST

For instance\, the following benchmark shows the block version to
be 5% slower&#8203;:

&nbsp;    use Benchmark qw /cmpthese/;

&nbsp;    our @&#8203;array = 101 \.\. 200;

&nbsp;    cmpthese \-2 => \{
&nbsp;        grep\_eq\_expr => '$&#8203;::a1 = grep  $\_ == 50\, @&#8203;&#8203;::array'\,
&nbsp;        grep\_eq\_blck => '$&#8203;::a2 = grep \{$\_ == 50\} @&#8203;&#8203;::array'\,
&nbsp;    \};

&nbsp;    \_\_END\_\_
&nbsp;                    Rate grep\_eq\_blck grep\_eq\_expr
&nbsp;    grep\_eq\_blck 28902/s           \-\-          \-5%
&nbsp;    grep\_eq\_expr 30282/s           5%           \-\-


If\, however\, the expression is a regexp\, things change drastically&#8203;:

&nbsp;    use Benchmark qw /cmpthese/;

&nbsp;    our @&#8203;array = 101 \.\. 200;

&nbsp;    cmpthese \-2 => \{
&nbsp;        grep\_re\_expr => '$&#8203;::b1 = grep  /^50$/\, @&#8203;&#8203;::array'\,
&nbsp;        grep\_re\_blck => '$&#8203;::b2 = grep \{/^50$/\} @&#8203;&#8203;::array'\,
&nbsp;    \};

&nbsp;    \_\_END\_\_
&nbsp;                    Rate grep\_re\_blck grep\_re\_expr
&nbsp;    grep\_re\_blck  6455/s           \-\-         \-69%
&nbsp;    grep\_re\_expr 20848/s         223%           \-\-

This difference in speed surprises me\.

It's not the fact that it's a regexp that's run inside a block\,
as a similar for loop doesn't show the huge difference&#8203;:

&nbsp;    use Benchmark qw /cmpthese/;

[...]

PR #20800, opened 2023-02-11
Perl_leave_adjust_stacks: don't make mortal copies of SvIMMORTAL SVs

richardleach:
SvIMMORTAL SVs cannot be prematurely freed and so there is no need
to create a mortal copy of them. They also will not leak, so there
is no need to add the SV* to the temp stack.

This was observed while looking at https://github.com/Perl/perl5/issues/7579,
where every scope exit from the regex block in `$​::b2 = grep {/^50$/} @​​::array`
currently involves creating a mortal copy of &PL_sv_no. With this PR, that
snippet doesn't create any such mortals and therefore has better performance.

Note: I'm only 95% sure about this one, so it needs careful review!

### DIFF
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -5041,6 +5041,13 @@ Perl_leave_adjust_stacks(pTHX_ SV **from_sp, SV **to_sp, U8 gimme, int pass)
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


tonycoz:
I think it's reasonable.

PR #20805, opened 2023-02-13
sv.c: fix compilation with g++ <= 7

xenu:
This compiler predates the addition of designated initializers to the C++ standard, so it's not surprising its implementation isn't fully compliant.

Omitting a field in a designated initializer list causes the following error:
```
sv.c: In function ‘void Perl_sv_upgrade(SV*, svtype)’:
sv.c:1088:13: sorry, unimplemented: non-trivial designated initializers not supported
```


xenu:
This fixes these failures:

https://perl5.test-smoke.org/report/5030563
https://perl5.test-smoke.org/report/5030557

PR #20806, opened 2023-02-13
Configure: don't probe for the malloc()/free() return type

tonycoz:
We require a C99 compiler, or in the case that led to this commit, a C++ compiler, and for those the malloc() and free() return types are well defined.

This probe broke on Solaris when building with g++.

Unlike glibc, the libc headers on Solaris, when building using a C++ compiler, define the stdlib.h functions within the std:: namespace, then imports those names into the top level namespace with 'using std::malloc'.

This conflicts with the declarations used in the probe, causing the probe to fail to build, despite malloc() actually returning a void *.

Since these two functions have well defined return types according to the standard, assume their return values match.

Configure can still be invoked with different definitions for malloctype and freetype, or hints can override them, so someone on a non-standard system can at least get past this if they really need to (such a system will likely not build perl anyway.)

### DIFF
--- a/Configure
+++ b/Configure
@@ -7917,46 +7917,20 @@ esac
 
 : compute the return types of malloc and free
 echo " "
-$cat >malloc.c <<END
-#$i_malloc I_MALLOC
-#$i_stdlib I_STDLIB
-#include <stdio.h>
-#include <sys/types.h>
-#ifdef I_MALLOC
-#include <malloc.h>
-#endif
-#ifdef I_STDLIB
-#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


khwilliamson (approved):

Tux (approved):

Leont (approved):

ilmari:
Hardcoding `malloctype` and `freetype` seems like a bit of a half-way house. Could we not get rid of `malloctype`/`Malloc_t` and `freetype`/`Free_t` completely?

Issue #20749, opened 2023-01-29
Multiple bugs in GNUMakefile


khwilliamson:
**Description**

miniperl is first made.  If a file that depends on is changed, it doesn't know to recompile miniperl.  
It only runs to completion the first time a work space is compiled.  This is because it does a `rename ..\config.sh.tmp config.sh` . `rename` fails if the destination file exists.

It runs to completion in spite of something failing.  For example, if the dll can't be generated (say because of a missing symbol), gmake will continue to go through the modules, compiling each one.  The original error, like a needle in a haystack,  ends up in the midst of the output.  This may not be a problem if the -j option isn't specified.

**Steps to Reproduce**

1st bug: touch locale.c; recompile; miniperl isn't remade.

2nd bug: gmake; gmake

3rd bug: Introduce a syntax error in a file.  Then gmake -jN  where N is some number, like the number of cores on your system

**Expected behavior**

None of these should happen
**Perl configuration**

```
This is a longstanding bug, and the configuration is irrelevant

**Comment by jkeenan** [comment]
The Subject of this ticket refers to `GNUMakefile`.  I don't see that file in the core distribution.  I do see:
```
$ find . -type f -iname '*GNUmakefile*'
./win32/GNUmakefile
```
Is that the file you're referring to?  If so, is this bug (and bug ticket) Win32-specific?

> **Description**
> 
> miniperl is first made. If a file that depends on is changed, it doesn't know to recompile miniperl. It only runs to completion the first time a work space is compiled. This is because it does a `rename ..\config.sh.tmp config.sh` . `rename` fails if the destination file exists.
> 
> It runs to completion in spite of something failing. For example, if the dll can't be generated (say because of a missing symbol), gmake will continue to go through the modules, compiling each one. The original error, like a needle in a haystack, ends up in the midst of the output. This may not be a problem if the -j option isn't specified.
> 
> **Steps to Reproduce**
> 
> 1st bug: touch locale.c; recompile; miniperl isn't remade.
> 

I see that happening on Linux, which implies that this problem is *not* Win32-specific.  Is that correct?

> 2nd bug: gmake; gmake
> 

What problem am I supposed to see in this case?

> 3rd bug: Introduce a syntax error in a file. Then gmake -jN where N is some number, like the number of cores on your system
> 
What problem am I supposed to see in this case?

demerphq:
> I see that happening on Linux, which implies that this problem is not Win32-specific. Is that correct?

[...]

PR #20811, opened 2023-02-15
re-work miniperl build on Win32

tonycoz:
These commits re-work the miniperl build:
- previously the miniperl config.h was overwritten by the full perl config, which, if the objects for miniperl were rebuilt, could lead to the newly compiled miniperl objects not matching the existing objects.  This also caused a problem @khwilliamson encountered where a Perl_putenv() definition was confused about whether the build was threaded or not.  Using a separate config.h (now in mini/ for miniperl) avoids this confusion.
- previously the miniperl object files didn't depend on their corresponding source files, presumably with the intent to avoid the above problem.  This could be frustrating if the developer was testing core changes by rapidly iterating on miniperl.  Since the miniperl object are now generated from a distinct config.h, this has been switched to a pattern (GNU make) or implicit rule (nmake) which adds an implicit dependency on the source file.

I believe this fixes some of the problems in #20749

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -1247,7 +1247,7 @@ $(MINIDIR)\.exists : $(CFGH_TMPL)
 #
 # MINIDIR generates config.h so miniperl.exe is not rebuilt when the 2nd
 # config.h is generated in CONFIGPM target, see also the comments for $(MINI_OBJ).
-	copy $(CFGH_TMPL) config.h
+	copy $(CFGH_TMPL) $(MINIDIR)\config.h
 	@(echo.&& \
 	echo #ifndef _config_h_footer_&& \
 	echo #define _config_h_footer_&& \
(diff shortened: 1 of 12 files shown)


win32/GNUmakefile
```diff
@@ -1420,23 +1420,27 @@ else
...
+	echo #endif)>> $(MINIDIR)\config.h
 else
 	@(echo #undef USE_CPLUSPLUS&& \
-	echo #endif)>> config.h
+	echo #endif)>> $(MINIDIR)\config.h
 endif
 #separate line since this is sentinal that this target is done
 	@rem. > $(MINIDIR)\.exists
 
+$(MINIDIR)\\%$(o): %.c
```
demerphq, line 1443: is the double backslash required here? just above it doesnt seem to be required.  Would a comment be in order?

Anyway, other than that LGTM.

demerphq (approved):

win32/GNUmakefile
tonycoz: A single `\` quotes the `%`, so an extra `\` is needed to quote the `\`.

https://www.gnu.org/software/make/manual/html_node/Static-Usage.html#index-_005c-_0028backslash_0029_002c-to-quote-_0025-1

(edit: better reference)

tonycoz:
Draft since I want to do some further checks.  Mostly I'm worried about `#include "config.h"` for miniperl checking the current directory first. (edit: which turns out to be a problem)

tonycoz:
@khwilliamson could you please see if this solves any of your problems from #20749

khwilliamson:
Yes it works AFAICT

tonycoz:
Applied manually.

PR #20813, opened 2023-02-16
Calculate conflict markers instead of hardcoding

afresh1:
It turns out that some tooling (Game of Trees) has special handling for conflict markers and will continually mark files that have them.

Since the test doesn't actually need them to be expanded, calculate them instead to not trigger the detection.

https://marc.gameoftrees.org/mail/1676484047.9764_0.html

### DIFF
--- a/t/comp/parser.t
+++ b/t/comp/parser.t
@@ -580,16 +580,13 @@ eval 'qq{@{0]}${}},{})';
 is(1, 1, "RT #124207");
 
 # RT #127993 version control conflict markers
+my @conflict_markers = map { $_ x 7 } qw( < = > );
 " this should keep working
-<<<<<<<
+$conflict_markers[0]
 " =~ /
->>>>>>>
+$conflict_markers[2]
 /;
-for my $marker (qw(
-<<<<<<<
-=======
->>>>>>>
-)) {
+for my $marker (@conflict_markers) {
     eval "$marker";
     like $@, qr/^Version control conflict marker at \(eval \d+\) line 1, near "$marker"/, "VCS marker '$marker' at beginning";
     eval "\$_\n$marker";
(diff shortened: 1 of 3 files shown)


afresh1:
Apparently there is another instance that has less obvious workarounds in t/lib/croak/toke.  If I find a fix for that before this merges I'll push another commit.

khwilliamson (approved):

afresh1:
Both because I don't see an obvious "fix" for the the croak/toke test and because [the folks over at GoT have decided that having conflict markers in files might be necessary](https://marc.gameoftrees.org/mail/1676548247.31257_0.html), I am going to close this and leave it be.  I don't think the code is significantly improved by obfuscating those markers.

demerphq:
On Fri, 17 Feb 2023, 01:29 Andrew Fresh, ***@***.***> wrote:

> Closed #20813 <https://github.com/Perl/perl5/pull/20813>.
>

I'm inclined to say we should keep this open regardless, I don't think we
should have conflict markers in our files, even for tests. I haven't looked
closely yet at the croak test andrew got stuck on but the general intent of
this PR seemed pretty reasonable to me.

Yves

afresh1:
That seems fine by me.  I will reopen and look forward to seeing ideas on fixing the croak/toke test.

demerphq:
I think the toke test depends on t/test.pl run_multiple_progs() (or maybe setup_multiple_progs) which needs to be patched to convert whatever replacement we use for conflict markers to the correct marker. Eg it could convert '<conflict<', '=conflict=' and '>conflict>' to the correct strings (and/or back, if necessary) it should work. I'm stuck on a phone just now so I may have overlooked something, but this is the general approach I'd take.

demerphq:
I updated the branch with code to fix the croak/toke issue.

PR #20814, opened 2023-02-16
Update perlre.pod

jidanni:
Users need to understand that their $2 etc. might not be defined always. I don't think this has been explicitly documented anywhere. Indeed, one could go on to mention (((.?)?)? ...

### DIFF
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -1313,6 +1313,9 @@ array you must use demarcated curly brace notation:
 See L<perldata/"Demarcated variable names using braces"> for more on
 this notation.
 
+B<NOTE>: (.?) will always create a $<digit> (incrementing $#{^CAPTURE}.)
+(.)? will only create a $<digit> if it matches.
+
 B<NOTE>: Failed matches in Perl do not reset the match variables,
 which makes it easier to write code that tests for a series of more
 specific cases and remembers the best match.


demerphq:
On Thu, 16 Feb 2023, 15:41 積丹尼 Dan Jacobson, ***@***.***>
wrote:

> Users need to understand that their $2 etc. might not be defined always. I
> don't think this has been explicitly documented anywhere. Indeed, one could
> go on to mention (((.?)?)? ...
>
I haven't reviewed your patch before saying this  but we document these
things in perlvar and keep perlre relatively light on the regexp vars as
there are a lot of them and we need to document them in perlvar regardless,
and perlre is already a very large doc so we put the meat of the docs on
them in perlvar.

Maybe you could review the latest state of perlvar and see whether your
patch is aligned or duplicative or in your opinion still relevant.

Yves


------------------------------

Grinnz:
I would agree this is a bit out of place. The documentation for `$1` etc in perlvar states that they are only set on successful matches.

demerphq:
I'm pretty sure the perlvar docs do explain the difference between `(X?)` and `(X)?` -- it comes down to `(X|)` vs `(?:(X)|)`

jidanni:
Well I was just thinking beginners will be reading perlre before
perlvar.

 ```
       /^(?:([\d.])+ mi w of )?SE cr Sec\s+(\d+)\s+T129N\s+R(\d+)W\s+N\s+Dak$/;
        my $u = $1 ? $1 : 0; # And will have added this safety line of code...
        print "$input: Mon ", ( $3 - 48 ) * 6 + 4 + ( 36 - $2 ) + $u, "\n";

```
...because they have been stung too many times about this thing that has
never been extra clearly documented.

But sure, somebody please mention something more than there currently
is somewhere.

Grinnz:
If anything, I think several references in this section should be updated with links to the appropriate perlvar sections.

demerphq:
Id prefer if you reworked this to reference perlvar.pod instead. I might give it a go myself if you dont want to, but im not comfortable with this wording and i think the best thing would be to direct people to the other doc.

jidanni:
Yes, I can guarantee my wording is bad (I failed perl 101. Don't even know what objects are.) 
I'm just hoping someone will rework it and mention something somewhere.

demerphq:
This is supplanted by https://github.com/Perl/perl5/pull/20823.

Thanks for the contribution, but I am going with my own more comprehensive rewrite. Please review the changes there. Thanks for the motivation to improve this.

PR #20823, opened 2023-02-19
perlre.pod - rework documentation of capture groups and backreferences

demerphq:
The documentation for capture groups and backreferences was getting somewhat awkward in terms of wording and structure and was inconsistent in terminology, difficult to read in raw form, and had omissions and mistakes in various places.

This is an attempt to clean it up, make it more readable, comprehensive, correct the mistakes, enhance what was there, and provide more links from it to perlvar.

### DIFF
--- a/pod/perlre.pod
+++ b/pod/perlre.pod
@@ -114,7 +114,7 @@ thing">> or the sequence S<C<"that thing">>.  The portions of the string
 that match the portions of the pattern enclosed in parentheses are
 normally made available separately for use later in the pattern,
 substitution, or program.  This is called "capturing", and it can get
-complicated.  See L</Capture groups>.
+complicated.  See L</Capture Groups and Backreferences>.
 
 The first alternative includes everything from the last pattern
 delimiter (C<"(">, C<"(?:"> (described later), I<etc>. or the beginning
@@ -874,7 +874,46 @@ inconsistencies (bugs) with the C</d> modifier, where Unicode rules
 would be used when inappropriate, and vice versa.  C<\p{}> did not imply
 Unicode rules, and neither did all occurrences of C<\N{}>, until 5.12.
 
-=head2 Regular Expressions
+=head2 Regexen, Regexps and Regular Expressions
+
+The term "regular expression" has a long history and a formal
+mathematical definition.  Unfortunately this formal definition is rather
+restricted in functionality: according to the strict definition of the
+term a regular expression I<must> be capable of being represented as a
+I<DFA>, which means "Deterministic Finite-state Automata", or via a
+mathematically equivalent but more flexible type of machine called an
+I<NFA>, which means "Non-Deterministic Finite-state Automata". Properly
+implemented neither a DFA nor an NFA will backtrack during matching,
+and they should match a string in linear time proportional to the length
+of the string being matched.
+
+Unfortunately this strict requirement precludes supporting many useful
+and powerful features, and many years ago the author of Perl, Larry
+Wall, decided that those additional useful features were more important
+than strict conformance to the requirements of a "regular expression"
+engine. As such he invented (or helped invent) the concept of a "regexp"
+(sometimes spelled without the "p"), which is very similar to a "regular
+expression" but with looser implementation requirements and much more
+powerful capabilities. (Many people still use the term "regular
+expression" for perls regexps, and the term can be found in use in this
+document, in books on Perl and in other Perl documentation. While this use
+is considered acceptable by the Perl programming community it is not strictly
+speaking correct and from time to time it can be helpful to distinguish
+between "regular expression" and "regexp" (or "regex" or "regexen").
+
+What this means is that while Perl's Regex Engine uses syntax very
+similar to that of a classic regular expression engine it is capable of
+much much more.  It does not use a DFA or NFA to do its matching,
+although it might use something similar in some cases. Instead it is
+classed as a "backtracking regex engine". In terms of computational
+theory it is essentially a "push down automata", and is capable of
+recursive patterns, matching backreferences, dynamic pattern matching
+and many other things that a normal regular expression engine is not.
+For instance Abigail (a legendary perl hacker) has shown that it is
+possible to use the Perl Regex Engine to solve Soduko puzzles.
+
+The following sections describe the breadth of features that the Perl
+Regex Engine supports.
 
 =head3 Quantifiers
 
@@ -882,33 +921,41 @@ Quantifiers are used when a particular portion of a pattern needs to
 match a certain number (or numbers) of times.  If there isn't a
 quantifier the number of times to match is exactly one.  The following
 standard quantifiers are recognized:
-X<metacharacter> X<quantifier> X<*> X<+> X<?> X<{n}> X<{n,}> X<{n,m}>
+X<metacharacter> X<quantifier> X<*> X<+> X<?>
 
     *           Match 0 or more times
     +           Match 1 or more times
     ?           Match 1 or 0 times
+
+Note that the C<+> and C<?> quantifiers are syntactic sugar. C<X+> is defined
+as C<XX*> and C<X?> is defined as C<(?:X|)>.
+
+Perl also supports numeric quantifiers to make things easier:
+X<{n}> X<{n,}> X<{n,m}>
+
     {n}         Match exactly n times
     {n,}        Match at least n times
     {,n}        Match at most n times
     {n,m}       Match at least n but not more than m times
 
-(If a non-escaped curly bracket occurs in a context other than one of
-the quantifiers listed above, where it does not form part of a
+If a non-escaped curly bracket occurs in a context other than one of the
+numeric quantifiers listed above, where it does not form part of a
 backslashed sequence like C<\x{...}>, it is either a fatal syntax error,
 or treated as a regular character, generally with a deprecation warning
 raised.  To escape it, you can precede it with a backslash (C<"\{">) or
-enclose it within square brackets  (C<"[{]">).
-This change will allow for future syntax extensions (like making the
-lower bound of a quantifier optional), and better error checking of
-quantifiers).
+enclose it within square brackets (C<"[{]">).  This change will allow for
+future syntax extensions (like making the lower bound of a quantifier
+optional), and better error checking of numeric quantifiers.
 
-The C<"*"> quantifier is equivalent to C<{0,}>, the C<"+">
-quantifier to C<{1,}>, and the C<"?"> quantifier to C<{0,1}>.  I<n> and I<m> are limited
-to non-negative integral values less than a preset limit defined when perl is built.
-This is usually 65534 on the most common platforms.  The actual limit can
+The C<"*"> quantifier is equivalent to C<{0,}>, the C<"+"> quantifier
+to C<{1,}>, and the C<"?"> quantifier to C<{0,1}>.  I<n> and I<m> are
+limited to non-negative integral values less than a preset limit
+defined when perl is built.  Prior to Perl 5.37.8 this was 65,534
+(U16_MAX) on the most common platforms, as of Perl 5.37.8 it is
+2,147,483,647 (I32_MAX) on most common platforms.  The actual limit can
 be seen in the error message generated by code such as this:
 
-    $_ **= $_ , / {$_} / for 2 .. 42;
+    $_ **= $_ , / x{$_} / for 2 .. 42;
 
 By default, a quantified subpattern is "greedy", that is, it will match as
 many times as possible (given a particular starting location) while still
@@ -1019,7 +1066,7 @@ X<\g> X<\k> X<\K> X<backreference>
   \pP       [3]  Match P, named property.  Use \p{Prop} for longer names
   \PP       [3]  Match non-P
   \X        [4]  Match Unicode "eXtended grapheme cluster"
-  \1        [5]  Backreference to a specific capture group or buffer.
+  \1        [5]  Backreference to a specific capture group.
                    '1' may actually be any positive integer.
   \g1       [5]  Backreference to a specific or previous group,
   \g{-1}    [5]  The number may be negative indicating a relative
@@ -1057,7 +1104,7 @@ See L<perlrebackslash/Misc> for details.
 
 =item [5]
 
-See L</Capture groups> below for details.
+See L</Capture Groups and Backreferences> below for details.
 
 =item [6]
 
@@ -1152,13 +1199,7 @@ string:
     print; 	# prints 1234X6789, not XXXXX6789
 
 
-=head3 Capture groups
-
-The grouping construct C<( ... )> creates capture groups (also referred to as
-capture buffers). To refer to the current contents of a group later on, within
-the same pattern, use C<\g1> (or C<\g{1}>) for the first, C<\g2> (or C<\g{2}>)
-for the second, and so on.
-This is called a I<backreference>.
+=head3 Capture Groups and Backreferences
 X<regex, capture buffer> X<regexp, capture buffer>
 X<regex, capture group> X<regexp, capture group>
 X<regular expression, capture buffer> X<backreference>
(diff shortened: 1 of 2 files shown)


pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+Capture bufers may also be given a name by appending C<< ?<name> >>
+immediately after the open paren, for example C<< (?<blah> ... ) >>
+creates a capture group called C<blah>. This will be covered in more
+detail a bit later on in this section.
+
+After a successful match the contents of a capture group may be accessed
+via a variety of different L<special variables|perlvar/Regex Variables>
+which can be used to access the content of the buffers by number or
+name, as well as to extract the offsets into the string that was
+matched, there is also a special variable that allows access to the
```
khwilliamson: Either a semi-colon instead of the comma; or better IMO, a separate sentence: 

> matched.  There is also ...

pod/perlre.pod
```diff
@@ -1593,35 +1674,41 @@ following this construct will be numbered as though the construct
...
-numbered groups holding the captures, and that interferes with the
-implementation of the branch reset pattern. If you are using named
-captures in a branch reset pattern, it's best to use the same names,
-in the same order, in each of the alternations:
+Prior to Perl 5.37.8 it was necessary to be careful when using the
+branch reset pattern in combination with named captures and named
+recursion. Named captures were implemented as being aliases to the
+underlying numbered capture group, which meant that if the same name
+were used in different alternations inside of a branch reset construct
+that the named buffer might end up referecing the numbered buffer in a
```
khwilliamson: s/that/then/
and "referencing"

pod/perlre.pod
```diff
@@ -1593,35 +1674,41 @@ following this construct will be numbered as though the construct
...
+confusing way. In older perls if you are using named captures in a
+branch reset pattern, it's best to use the same names, in the same
+order, in each of the alternations:
 
    /(?|  (?<a> x ) (?<b> y )
       |  (?<a> z ) (?<b> w )) /x
 
-Not doing so may lead to surprises:
+not doing so could lead to surprises:
```
khwilliamson: s/not/Not

pod/perlre.pod
```diff
@@ -1593,35 +1674,41 @@ following this construct will be numbered as though the construct
...
 
   "12" =~ /(?| (?<a> \d+ ) | (?<b> \D+))/x;
   say $+{a};    # Prints '12'
   say $+{b};    # *Also* prints '12'.
 
 The problem here is that both the group named C<< a >> and the group
-named C<< b >> are aliases for the group belonging to C<< $1 >>.
+named C<< b >> are aliases for the group belonging to C<< $1 >>. As of
```
khwilliamson: This doesn't make sense to me.  I think you mean

> named C<< b >>, prior to Perl 5.37.8, were aliases for the group belonging to C<< $1 >>. As of

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+capture groups as a whole in a single array or the named captures via a
+hash.
+
+The oldest, most common and traditional way to access the contents of a
+capture buffer after a successful match is via a
+L<digit var|perlvar/Digit Vars> such as C<$1> or C<$2> I<etc>, which refer
+to the first and second capture groups in the pattern. You may refer to
+a named capture group by using C<$+{name}>. Multiple capture groups may
+be given the same name, and you may access the contents of all of the
+capture groups with the same name via the array returned by C<$+{name}>.
```
khwilliamson: > The same name may be used for more than one capture group.  You can use C<$+{name}> ...

but it doesn't make sense to me that an array is accessed via a dollar sigil, unless it is a reference, and you need to say $+{name}->[0] etc

khwilliamson (approved):

pod/perlre.pod
demerphq: Yes it is a reference. I was more thinking of code like

```
my $all_captures_for_foo = $+{foo};
print "@$all_captures_for_foo";
```

but ill figure out a better wording.

pod/perlre.pod
demerphq: Willdo.

pod/perlre.pod
demerphq: Willdo.

pod/perlre.pod
demerphq: Willdo.

pod/perlre.pod
demerphq: Yep, that is better wording.

pod/perlvar.pod
```diff
@@ -1008,6 +1009,8 @@ find uses of these problematic match variables in your code.
 
 =over 8
 
+=item Digit Vars
```
Grinnz: I understand you're trying to make more linkable headings with this and Branch Reset, and I don't have a particular problem with doing it this way, just noting it's unorthodox for these documents.

pod/perlvar.pod
```diff
@@ -885,7 +885,8 @@ command or referenced as a file.
 
 =back
 
-=head2 Variables related to regular expressions
+=head2 Regex Variables
```
Grinnz, line 887: I don't see the need for changing this heading when it would break any existing links to this section.

pod/perlre.pod
```diff
@@ -1152,13 +1152,7 @@ string:
...
 
 
-=head3 Capture groups
-
-The grouping construct C<( ... )> creates capture groups (also referred to as
-capture buffers). To refer to the current contents of a group later on, within
-the same pattern, use C<\g1> (or C<\g{1}>) for the first, C<\g2> (or C<\g{2}>)
-for the second, and so on.
-This is called a I<backreference>.
+=head3 Capture Groups and Backreferences
```
Grinnz, line 1202: This would also break existing links to this section and I'm not sure how much value it brings to add the word backreferences to the heading.

pod/perlvar.pod
Grinnz: (The perlvar addition in particular may require me to special case it in Perldoc Browser to avoid considering "Digit Vars" a new variable name, but that's my problem.)

Grinnz: Personal preference would be to spell out "Variables" in headings.

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+The grouping construct C<( ... )> creates a capture group (also
+referred to as a "capture buffer" or sometimes simply "capture" or
+"buffer"). There is no limit to the number of capture groups that you
+may use. Capture groups are numbered with the leftmost open parenthesis
+being number 1, the second from the left being 2, I<etc>, with one
+exception, capture groups inside of the "branch reset" construct
+C<(?| ... )> are numbered differently. (See the section on
+L<branch reset|/Branch Reset> later in this file.)
+
+Capture bufers may also be given a name by appending C<< ?<name> >>
```
Grinnz: typo: "buffers"

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+which can be used to access the content of the buffers by number or
+name, as well as to extract the offsets into the string that was
+matched, there is also a special variable that allows access to the
+capture groups as a whole in a single array or the named captures via a
+hash.
+
+The oldest, most common and traditional way to access the contents of a
+capture buffer after a successful match is via a
+L<digit var|perlvar/Digit Vars> such as C<$1> or C<$2> I<etc>, which refer
+to the first and second capture groups in the pattern. You may refer to
```
Grinnz: This suggests that digit vars as a whole only refer to the first two capture groups, suggested rephrase: "which refer to the corresponding numbered capture groups."

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+referred to as a "capture buffer" or sometimes simply "capture" or
+"buffer"). There is no limit to the number of capture groups that you
+may use. Capture groups are numbered with the leftmost open parenthesis
+being number 1, the second from the left being 2, I<etc>, with one
+exception, capture groups inside of the "branch reset" construct
+C<(?| ... )> are numbered differently. (See the section on
+L<branch reset|/Branch Reset> later in this file.)
+
+Capture bufers may also be given a name by appending C<< ?<name> >>
+immediately after the open paren, for example C<< (?<blah> ... ) >>
```
Grinnz: spell out "parenthesis"

pod/perlre.pod
Grinnz: This doesn't seem accurate? From perlvar: "If multiple distinct capture groups have the same name, then $+{NAME} will refer to the leftmost defined group in the match."

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+capture buffer after a successful match is via a
+L<digit var|perlvar/Digit Vars> such as C<$1> or C<$2> I<etc>, which refer
+to the first and second capture groups in the pattern. You may refer to
+a named capture group by using C<$+{name}>. Multiple capture groups may
+be given the same name, and you may access the contents of all of the
+capture groups with the same name via the array returned by C<$+{name}>.
+See L<%+|perlvar/"%+"> and L<%-|perlvar/"%-"> for details.
+
+It is important to understand that there is a distinction between a
+capture group that did I<not match> (even though the overall match
```
Grinnz: I'm not sure about the phrasing "didn't match" for this, when it's more that the capture group was not chosen as part of the overall successful match. I could take a stab at rephrasing this section that way if you agree.

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+pattern, use C<\g1> (or C<\g{1}>) for the first, C<\g2> (or C<\g{2}>)
+for the second, and so on. This is called a I<backreference>. If a group
+did not match, the associated backreference won't match either. This can
+happen if the group is optional, or in a different branch of an
+alternation. When using a backreference in your pattern you may omit the
+C<"g">, and write C<"\1">, I<etc>, but there are some issues with this
+form, described a bit further below.
+
+X<forward backreference>
+Notice that even though the name for matching the same thing as another
```
Grinnz: "Notice that" unnecessary here

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+capture group in the current pattern is called a "backreference" it is
```
Grinnz: comma after "backreference"

pod/perlre.pod
```diff
@@ -1193,54 +1257,68 @@ would match the same as C</(Y) ( (X) \g3 \g1 )/x>.  This allows you to
...
+when creating a regex by concatenating smaller strings. For example if
+you have C<qr/$x$y/>, and C<$x> contained C<"\g1">, and C<$y>
+contained C<"37">, you would get C</\g137/> which is probably not what
+you intended and will likely throw an exception during compilation (unless
+the pattern actually contained 137 capture buffers).
 
 If you use braces, you may also optionally add any number of blank
 (space or tab) characters within but adjacent to the braces, like
-S<C<\g{ -1 }>>, or S<C<\k{ I<name> }>>.
+S<C<\g{ -1 }>>, or S<C<\k{ I<name> }>>. Notice that C<\g{ name }> and
```
Grinnz: I think it would be better to just stick to `\g` in examples after explaining what `\k` is once, since it's rarely used.

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+via a variety of different L<special variables|perlvar/Regex Variables>
+which can be used to access the content of the buffers by number or
+name, as well as to extract the offsets into the string that was
+matched, there is also a special variable that allows access to the
+capture groups as a whole in a single array or the named captures via a
+hash.
+
+The oldest, most common and traditional way to access the contents of a
+capture buffer after a successful match is via a
+L<digit var|perlvar/Digit Vars> such as C<$1> or C<$2> I<etc>, which refer
```
Grinnz: comma before etc

pod/perlre.pod
```diff
@@ -1167,18 +1161,88 @@ X<\g{1}> X<\g{-1}> X<\g{name}> X<relative backreference> X<named backreference>
...
+to the first and second capture groups in the pattern. You may refer to
+a named capture group by using C<$+{name}>. Multiple capture groups may
```
Grinnz: I don't think named capture groups need to be brought up in this part, as they're discussed later - the reader hasn't been told how to declare them yet so doesn't need the explanation how to access them yet. So the remaining part of this paragraph could be added to where those are introduced.

pod/perlre.pod
```diff
@@ -1193,54 +1257,68 @@ would match the same as C</(Y) ( (X) \g3 \g1 )/x>.  This allows you to
 interpolate regexes into larger regexes and not have to worry about the
 capture groups being renumbered.
 
-You can dispense with numbers altogether and create named capture groups.
-The notation is C<(?E<lt>I<name>E<gt>...)> to declare and C<\g{I<name>}> to
-reference.  (To be compatible with .Net regular expressions, C<\g{I<name>}> may
-also be written as C<\k{I<name>}>, C<\kE<lt>I<name>E<gt>> or C<\k'I<name>'>.)
+As mentioned earlier you can dispense with numbers altogether and create
```
Grinnz: comma after "earlier" (or remove if you remove the earlier reference to named capture vars)

pod/perlre.pod
```diff
@@ -1593,35 +1674,41 @@ following this construct will be numbered as though the construct
...
     / ( a )  (?| x ( y ) z | (p (q) r) | (t) u (v) ) ( z ) /x
     # 1            2         2  3        2     3     4
 
-Be careful when using the branch reset pattern in combination with
-named captures. Named captures are implemented as being aliases to
-numbered groups holding the captures, and that interferes with the
-implementation of the branch reset pattern. If you are using named
-captures in a branch reset pattern, it's best to use the same names,
-in the same order, in each of the alternations:
+Prior to Perl 5.37.8 it was necessary to be careful when using the
```
Grinnz: comma after "5.37.8"

pod/perlre.pod
```diff
@@ -1193,54 +1257,68 @@ would match the same as C</(Y) ( (X) \g3 \g1 )/x>.  This allows you to
...
+Capture group contents are dynamically scoped and available to you
+outside the pattern until the end of the enclosing block or until the
+next successful match in the same scope, whichever comes first. See
+L<perlsyn/"Compound Statements"> and
 L<perlvar/"Scoping Rules of Regex Variables"> for more details.
 
-You can access the contents of a capture group by absolute number (using
-C<"$1"> instead of C<"\g1">, I<etc>); or by name via the C<%+> hash,
-using C<"$+{I<name>}">.
+=head4 Use of curly braces in backreference notation
```
Grinnz, line 1353: I like the idea of splitting this into head4 subheadings (it's pretty sorely needed), but this one isn't sufficient on its own: it would imply that the whole rest of the "Capture groups" section is under this subheading which isn't the case. Suggested other subheadings: "Backreferences", "Named capture groups" (moving the last paragraph about capture group scoping before the section on backreferences), and maybe "Backreference notation ambiguity", "Examples", and "Match variables"? some reorganization might be needed.

pod/perlre.pod
demerphq: > I'm not sure about the phrasing "didn't match" for this

Im open to suggestions for sure! Keep in mind that `/(x)?/` and `/(?:(x)|)/` are equivalent. The wording "did not match" was more inspired by the former. Maybe "did not match, or were not involved in the matching pathway through the pattern". But that reads a bit clunky even if it is more correct.

pod/perlre.pod
demerphq: Actually i introduce them at line 1174.

pod/perlre.pod
demerphq: > This doesn't seem accurate?

the `$+{NAME}` is a typo, it should say `$-{NAME}` in that example. Good catch.

pod/perlre.pod
demerphq: Ack.

pod/perlvar.pod
demerphq: I couldn't find any. Do you mean those outside of the core?

pod/perlre.pod
demerphq: I am going to park this one until the rest of the feedback is cleared.

pod/perlre.pod
demerphq: This is related to the other point I am parking for now. I want to get the other stuff cleaned up and then ill thinking about further restructuring.

pod/perlre.pod
demerphq: I did some rewording myself. SO I am marking this as resolved for now.

pod/perlvar.pod
Grinnz: In general, any documentation may link to sections in core docs

pod/perlvar.pod
demerphq: Removed.

pod/perlvar.pod
demerphq: I'd like to rename it. See latest version of the patch.

demerphq:
@Grinnz you available for further review?

pod/perlre.pod
```diff
@@ -874,41 +874,88 @@ inconsistencies (bugs) with the C</d> modifier, where Unicode rules
 would be used when inappropriate, and vice versa.  C<\p{}> did not imply
 Unicode rules, and neither did all occurrences of C<\N{}>, until 5.12.
 
-=head2 Regular Expressions
+=head2 Regexen, Regexps and Regular Expressions
```
Grinnz, line 877: I don't particularly agree with the relevance of this section here (in addition to my previous objections to renaming sections unnecessarily)

Grinnz: To try to be more specific: this seems like a subject for a blog post, rather than Perl documentation on regexes.

Grinnz: (at the very least, I would prefer this be a separate task from this PR's useful reorganization)

PR #20824, opened 2023-02-19
scope.c - rework SSGROW() and SSCHECK() macros and undelying functions

demerphq:
Prior to this patch SSCHECK() took a "needs" parameter, but did not actually guarantee that the stack would be sufficiently large to accomodate that many elements after the call. This was quite misleading. Especially as SSGROW() would not do geometric preallocation, but SSCHECK() would, so much of the time SSCHECK() would appear to be a better choice, but not always.

This patch makes it so SSCHECK() is an alias for SSGROW(), and it makes it so that SSGROW() also geometrically overallocates. The underlying function that used to implement SSCHECK() savestack_grow() now calls the savestack_grow_cnt() which has always implemented SSGROW(). Anything in the internals that used to call SSCHECK() now calls SSGROW() instead.

At the same time the preallocation has been made a little bit more aggressive, ensuring that we always allocate at least SS_MAXPUSH elements on top of what was requested as part of the "known" size of the stack, and additional SS_MAXPUSH elements which are not part of the "known" size of the stack. This "hidden extra" is used to simply some of the other macros which are used a lot internally. (I have beefed up the comment explaining this as well.)

See https://github.com/Perl/perl5/pull/20530 where Dave M points out the problem with this logic. Currently I don't think we have a good way to test for the error that he showed so unfortunately this patch lacks a test.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -262,7 +262,7 @@ S_regcppush(pTHX_ const regexp *rex, I32 parenfloor, U32 maxopenparen _pDEPTH)
             );
     );
 
-    SSCHECK(total_elems + REGCP_FRAME_ELEMS);
+    SSGROW(total_elems + REGCP_FRAME_ELEMS);
     assert((IV)PL_savestack_max > (IV)(total_elems + REGCP_FRAME_ELEMS));
 
     /* memcpy the offs inside the stack - it's faster than for loop */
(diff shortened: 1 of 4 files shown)


demerphq:
@iabyn this is in response to your comment in https://github.com/Perl/perl5/pull/20530.

demerphq:
Merging as this fixes a bug I dont think we should let loose in a released perl.

PR #20825, opened 2023-02-19
Smoke me/davem/rc stack1

iabyn:
First cut of the first half of making the argument stack ref counted.
Functionally correct, all tests pass, Moose installs ok, but everything is very slow due to most ops still being wrapped.
The second half, to come later, will be to unwrap all the small ops (like pp_const()) and add other optimisations to make it run at blead speed.

By default the RC stack isn't enabled, and this branch works almost identically to blead. Build it with -Accflags=-DPERL_RC_STACK to get the RC stack. Then run something like perl -DstR -e'$x=1' to see the ref counting in action.

I intend for general discussions around this (such as when to merge into blead) to take place on p5p rather than here. I'll be starting a thread at some point soon.


khwilliamson:
On 2/19/23 13:47, iabyn wrote:
> First cut of the first half of making the argument stack ref counted.
> Functionally correct, all tests pass, Moose installs ok, but everything 
> is very slow due to most ops still being wrapped.
> The second half, to come later, will be to unwrap all the small ops 
> (like pp_const()) and add other optimisations to make it run at blead speed.
> 
> By default the RC stack isn't enabled, and this branch works almost 
> identically to blead. Build it with -Accflags=-DPERL_RC_STACK to get the 
> RC stack. Then run something like perl -DstR -e'$x=1' to see the ref 
> counting in action.
> 
> I intend for general discussions around this (such as when to merge into 
> blead) to take place on p5p rather than here. I'll be starting a thread 
> at some point soon.
> 
> 
>

It would be nice to take a few randomly selected ref-counted bugs and 
see that this brasnch indeed fixes them

PR #20828, opened 2023-02-20
perldelta for my_snprintf

khwilliamson:
a683fa5b7e816ae5c10d246c9a0b1f3ea743274b


demerphq (approved):
You could put a touch more effort into create a commit message body. But approved nonetheless. :-)

PR #20833, opened 2023-02-20
sv.h - make unused bit more obvious with a more obvious define (Variant 1/4)

demerphq:
Dave M complained about the previous version of this.

I want to see if this would be acceptable to him as a compromise.

Unfortunately his expectations and mine are more or less diametrically opposed. I really do not think gaps in these types of definitions is helpful. I find it really confusing when I find a gap in a bit sequence like this.

The fact that we can't nest C comments makes it worse.

### DIFF
--- a/sv.h
+++ b/sv.h
@@ -2085,7 +2085,7 @@ Returns the hash for C<sv> created by C<L</newSVpvn_share>>.
 #define SV_IMMEDIATE_UNREF      (1 <<  0) /* 0x0001 -     1 */
 #define SV_GMAGIC               (1 <<  1) /* 0x0002 -     2 */
 #define SV_COW_DROP_PV          (1 <<  2) /* 0x0004 -     4 */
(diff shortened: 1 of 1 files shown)


tonycoz:
I prefer 2 or 3, I don't think we need to give the unused bit a name, even commented out.

demerphq:
> I prefer 2 or 3, I don't think we need to give the unused bit a name, even commented out.

Ok. Just to clarify tho, are you taking into account how difficult it would be to write tooling to correct or change the list? For instance with the change you requested what would the script look like to add the octal value of the defines?

tonycoz:
> > I prefer 2 or 3, I don't think we need to give the unused bit a name, even commented out.
> 
> Ok. Just to clarify tho, are you taking into account how difficult it would be to write tooling to correct or change the list? For instance with the change you requested what would the script look like to add the octal value of the defines?

That wasn't listed as a criteria, I didn't see Dave's comment that lead to these PRs.

The unused entries could have a `/* NOT USED (1<<3) */` entry that the tool could pick up.

PR #20834, opened 2023-02-20
perldelta.pod - note that REG_INF has been raised

demerphq:
It used to be U16_MAX and it is now I32_MAX.


demerphq:
Merging to meet release deadline.

Issue #20839, opened 2023-02-21
perl-5.36.0 test lib/Benchmark.t has no exit.


blastwave:
**Description**
```
perl-5.36.0 test lib/Benchmark.t has no exit.
```

**Steps to Reproduce**
```
Perform a compile of 5.36.0 and then run the testsuite.
```

**Expected behavior**
```
 ? unknown ?

I suspect that the test supposedly will exit after it performs
enough abusive Fibonacci loops.
```

**Perl configuration**
```
$ /usr/local/bin/perl -V
Summary of my perl5 (revision 5 version 36 subversion 0) configuration:
   
  Platform:
    osname=solaris
    osvers=2.10
    archname=sun4-solaris-64-ld
    uname='sunos pollux 5.10 generic_150400-44 sun4u sparc sunw,ultraax-i2 '
    config_args='-Dcc=/opt/solarisstudio12.4/bin/c99 -Dprefix=/usr/local -Dmksymlinks -Accflags=-Xc -g -xO0 -errfmt=error -erroff=%none -xmemalign=8s -errshort=full -xstrconst -xildoff -m64 -xnolibmil -xcode=pic32 -xregs=no%appl -xlibmieee -ftrap=%none -xchip=ultra -mc -xs -xbuiltin=%none -xdebugformat=dwarf -xunroll=1 -L/usr/local/lib -R/usr/local/lib -I/usr/local/include -D_XOPEN_SOURCE=600 -D__EXTENSIONS__'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=define
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='/opt/solarisstudio12.4/bin/c99'
[...]

PR #20842, opened 2023-02-22
Benchmark.t - sanity check for a broken times()

demerphq:
If time() or times() is broken then Benchmark can infinite loop. This adds a sanity check that will die early if it appears that either are broken.

This fixes the infinite loop part of GH Issue #20839

### DIFF
--- a/lib/Benchmark.t
+++ b/lib/Benchmark.t
@@ -40,6 +40,89 @@ sub cmp_delta {
     return $max/$min <= (1+$delta);
 }
 
+sub splatter {
+    my ($message) = @_;
+    my $splatter = <<~'EOF_SPLATTER';
+    Please file a ticket to report this. Our bug tracker can be found at
+
+        https://github.com/Perl/perl5/issues
+
+    Make sure you include the full output of perl -V, also uname -a,
+    and the version details for the C compiler you are using are
+    very helpful.
+
+    Please also try compiling and running the C program that can
+    be found at
+
+        https://github.com/Perl/perl5/issues/20839#issuecomment-1439286875
+
+    and provide the results (or compile errors) as part of your
+    bug report.
+
+    EOF_SPLATTER
+
+    if ( $message =~ s/\.\.\.//) {
+        $splatter =~ s/Please/please/;
+    }
+    die $message, $splatter;
+}
+
+{
+    # Benchmark may end up "looping forever" if time() or times() are
+    # broken such that they do not return different values over time.
+    # The following crude test is intended to ensure that we can rely
+    # on them and be confident that we won't infinite loop in the
+    # following tests.
+    #
+    # You can simulate a broken time or times() function by setting
+    # the appropriate env var to a true value:
+    #
+    #   time()    -> SIMULATE_BROKEN_TIME_FUNCTION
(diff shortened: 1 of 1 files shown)


lib/Benchmark.t
```diff
@@ -40,6 +40,73 @@ sub cmp_delta {
...
+        $splatter =~ s/Please/please/;
+    }
+    die $message, $splatter;
+}
+
+{
+    # Benchmark may end up "looping forever" if time() or times() are
+    # broken such that they do return changing values over time. The
+    # following crude test is intended to ensure that we can reply on
+    # them sufficiently we don't infinite loop.
```
wolfsage: * "such that they do" -> "such that they do not"
* "we can reply" -> "we can rely"
* "sufficiently we don't infinite loop" - "so that we don't infinite loop"

lib/Benchmark.t
```diff
@@ -40,6 +40,73 @@ sub cmp_delta {
...
+
+    my @times0;
+    for (1..3) {
+        my $end_time = time + 1;
+        my $count= 0;
+        my $scale = $ENV{FAST_CPU} || 1;
+        my $count_threshold = 100_000;
+        while ( time < $end_time ) {
```
wolfsage: In the case where time does not work here (forcing while (1)), this entire block from for (1..3) { ... } takes 2 minutes and 3 seconds on my pretty good hardware. I suppose this should be rare, but it will still "look" like hanging. I wonder if it can be adjusted down, but maybe it doesn't matter.

wolfsage (commented):
A few spelling / grammar fixes perhaps, and also a question about speed in the worst case

lib/Benchmark.t
demerphq: Will fix. Thanks!

lib/Benchmark.t
demerphq: Im open to pushing it down. This was purely a finger in the air kind of estimate. So ill try out a lower number.

I do think at a certain level it doesn't matter. The main objective is to error eventually, instead of never. And so far we haven't seen any broken time() functions afaik, just times(), this was just a sanity defense if there /was/ a brokeb time() function out there so that our sanity check wouldn't recreate the bug it was supposed to protect against. :-(

But we can try a lower number I think.  Ill queue something up in the next few minutes.

lib/Benchmark.t
demerphq: Done.

lib/Benchmark.t
demerphq: And also done. I added some ENV vars to make it easier to simulate time() or times() failing. 

```
+    #   time()    -> SIMULATE_BROKEN_TIME_FUNCTION
+    #   times()   -> SIMULATE_BROKEN_TIMES_FUNCTION
```

wolfsage: Nice

wolfsage:
LGTM with one nit: You have `my $x= 0.0;` and `my $count=` ... with no spacing before the `=`. If you could fix those that'd be great

wolfsage (approved):

demerphq:
> If you could fix those that'd be great

done.

PR #20845, opened 2023-02-22
.github/workflows/testsuite.yml - switch from secrets to vars

### LINKED ISSUE #20843: Github secrets contain values that they should not, which then get elided in our tests output
**Description**
We have put some kind of configuration variables in our github secrets data which do not need to be secret. Any output of a secret is elided. So right now I am looking at a test result like this:

```
../lib/B/Deparse.t ................................................... ok

#   Failed test 'expecting $count < (***0_000 * ***)'
#   at ../lib/Benchmark.t line 90.
#     '***000***'
#         <
#     '***0000'
Either this system is extremely fast, or the time() function
is broken.
```
I do not have access to the secrets data (afaik) to check, but i surmise that it now contains the value "100". These values need to be removed from the secrets file. We should not store any simple integers in there at all. EVER. (Github refuses to fix this, I have already contacted them about it and they refused.)

See: https://github.com/Perl/perl5/actions/runs/4238660267/jobs/7365962545#step:10:2648

**Steps to Reproduce**
N/A.

**Expected behavior**
I expect to see useful data from my test failures, and not a bunch of '*'s instead of the test number of value.

**Perl configuration**
NA.

demerphq:
We have been using secrets to store boolean variables. This results in our CI output being annoyingly corrupted by the security eliding.

Github now supports "variables" as well as secrets, which are not elided in CI output. So switch our test framework controls to use the variable. Note the EXTENDED_TESTING variable was switched from to DO_EXTENDED_TESTING so that we can have both the EXTENDED_TESTING /secret/ and the new DO_EXTENDED_TESTING variable available at the same time as this commit is processed.

Fixes #20843

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -9,8 +9,8 @@
 # When actions are enabled then it will - by default - run the 'Sanity Check' job.
 #
 # Testing other build configurations is controlled by the 'Secrets' configured
-# in 'Settings' -> 'Secrets' -> 'Actions' -> 'Repository Secrets':
-#   - EXTENDED_TESTING: when this secret exists and is set to a true value then
+# in 'Settings' -> 'Secrets and Variables' -> 'Actions' -> 'Variables':
+#   - DO_EXTENDED_TESTING: when this var exists and is set to a true value then
 #     all build configurations are tested;
 #   - CI_SKIP_SANITY: if set to a true value then most of the 'Sanity Check' job is skipped;
 #   - CI_FORCE_LINUX: if set to a true value: run the 'linux' job;
@@ -33,18 +33,11 @@
 #      * string containing only zeroes (i.e. '0', '0000')
 #      * string containing only spaces and zeroes (i.e. '0  0 0 0')
 #
-# When setting a value it's best to set a long and complex value
-# since GitHub will mask the secrets in the output. Which means if you
-# set (for example) EXTENDED_TESTING to 1 then all occurrences of '1'
-# are replaced with '*'. (e.g. `s/1/*/g;`). So it's recommended to set
-# something that is unlikely to occur in the job(/build/test) output
-# (for example: '111111111111111111', 'Rumpelstiltskin', ...).
-#
 # Example: if you only want to run tests on cygwin then you can set:
-#   - EXTENDED_TESTING=00000000000000000000000000000000000000
-#   - CI_SKIP_SANITY=1111111111111111111111111111111111111111
-#   - CI_FORCE_CYGWIN=111111111111111111111111111111111111111
-#   - (and all other CI_FORCE_... secrets to 0000000000000000)
+#   - DO_EXTENDED_TESTING=0
+#   - CI_SKIP_SANITY=1
+#   - CI_FORCE_CYGWIN=1
+#   - (and all other CI_FORCE_... vars to 0)
 #
 
 name: testsuite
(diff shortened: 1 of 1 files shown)


bram-perl:
Preliminary note: I haven't used GitHub Variables yet; so what follows might be incorrect.

Some thoughts:

Looking at https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows shows: `Note: Configuration variables for GitHub Actions are in beta and subject to change.`. No idea if it will change or if changes are planned or ...; it's merely a note I saw so we may need to use it with care.

I think with variables the workflow can be simplified. If I remember correctly with secrets we ran into two issues:
- unable to directly use it in a `if` condition; and
- needing the shell to allow for "false" values such as "00000" (because it would otherwise mask '0' in the output)

With variables I think both of these go away.
Looking at the variables page they have the example:
```
    if: ${{ vars.USE_VARIABLES == 'true' }}
```

so it should be possible to use them directly in the if-conditions.

i.e.:
```
    if: (! cancelled() && (needs.sanity_check.outputs.run_all_jobs == 'true' || needs.sanity_check.outputs.ci_force_linux_i386 == 'true'))
```
could become something like (**UNTESTED**!)
```
    if: (! cancelled() && (needs.sanity_check.outputs.run_all_jobs == 'true' || vars.CI_FORCE_LINUX_I386 == 'true'))
```

(this would then remove the need for the "Check what-to-do" step in the sanity check run)

atoomic (approved):

rwp0 (approved):

demerphq:
> I think with variables the workflow can be simplified.

@bram-perl I would prefer we tackle that as a follow up, not as part of this PR. Changing the least possible at one time means it will be easier to debug later, and strictly speaking we don't need to change anything, it works as is, it is just more complex than it needs to be. I will add a comment saying that we should simply and pointing at your comment. 

> if: (! cancelled() && (needs.sanity_check.outputs.run_all_jobs == 'true' || vars.CI_FORCE_LINUX_I386 == 'true'))

Do we still need "needs.sanity_check.outputs.run_all_jobs == 'true'"  at all? Can't we just look at the var DO_EXTENDED_TESTING?

Anyway, lets leave those questions for afterwards.

bram-perl:
> > I think with variables the workflow can be simplified.
> 
> @bram-perl I would prefer we tackle that as a follow up, not as part of this PR. Changing the least possible at one time means it will be easier to debug later, and strictly speaking we don't need to change anything, it works as is, it is just more complex than it needs to be. I will add a comment saying that we should simply and pointing at your comment.

I kind of disagree.
The new vars were an opportunity to go for something consistent/restrictive.
I.e. only 'true' is true and all the rest (0, 1, "foobar", ...) is false.

Now that this is merged it means someone can configure the vars as '1' and it will work but if we later change the workflow to only recognize 'true' as true then it will (again) fail for them and they (again) need to update their repo settings..

demerphq:
> I kind of disagree.

@bram-perl I am sorry about that. I feel that this is a "perfect is the enemy of good enough" situation and that getting rid of the secret var from the master repo is the priority. Anyway, I haven't deleted the secret var /yet/ (although I changed its value to "1_true_1_True_1_TRUE_1_TrUe_1_tRuE_1" or something like that), so if we want we can revert, but I don't think we should. Better to just get the changes you think need to be made done in a timely fashion so that the problem you describe doesn't occur.

Having said that, while I am with you that simplifying our recipe is a good thing, I am not so sure I agree that it is a good thing making only "true" be "true" and everything else be "false", this is the repo for perl after all, it doesn't seem /that/ weird that our config follows perl semantics, if anything it seems like a feature.  I guess the "only 'true' is true" definition makes the recipe more understandable to someone more familiar with the github infra itself, but i am not sure it makes it more approachable or usable for perl people who just want to hack in their fork and flip some knobs to control testing.

Finally, I really don't want to see someone set EMBEDDED_TESTING back to 0 or 1 again and I'd like to get it deleted from the secrets ASAP so that for sure it wont happen again. It is very frustrating when a test fails and you cannot tell the test number that failed, or read the test output because it was elided.  Any inconvenience caused by people needing to update their settings in their private repo is IMO a minor issue compared to having the master repo produce elided test output for simple digits like 1 or 0.

Anyway, I am happy to work with you to get something better set up so that have something concrete to debate over. Personally id be fine leaving it as it is and just deleting the old secret var so that we have no secrets at all in the repo.

PR #20846, opened 2023-02-22
Make more use of the various RETURN macros

JRaspass:
We have them to reduce boilerplate, may as well use them.

### DIFF
--- a/op.c
+++ b/op.c
@@ -15497,8 +15497,7 @@ const_av_xsub(pTHX_ CV* cv)
         Perl_croak(aTHX_ "Magical list constants are not supported");
     if (GIMME_V != G_LIST) {
         EXTEND(SP, 1);
-        ST(0) = sv_2mortal(newSViv((IV)AvFILLp(av)+1));
-        XSRETURN(1);
+        XSRETURN_IV((IV)AvFILLp(av)+1);
     }
     EXTEND(SP, AvFILLp(av)+1);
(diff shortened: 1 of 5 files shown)


demerphq:
@iabyn what would you prefer with patches like this? Would you prefer we hold off until the stack refcounting stuff is merged or does this kind of thing help you?

@JRaspass in theory LGTM but I want to make sure it's not going to cause unnecessary extra work for @iabyn .

rwp0 (approved):

iabyn:
On Wed, Feb 22, 2023 at 02:56:18PM -0800, Yves Orton wrote:
> @iabyn what would you prefer with patches like this? Would you prefer we hold off until the stack refcounting stuff is merged or does this kind of thing help you?
> 
> @JRaspass in theory LGTM but I want to make sure it's not going to cause unnecessary extra work for @iabyn .

I very much do not want this patch to be applied, sorry!

Eventually core PP code will not be using PUSH/RETURN macros of any
variety, and this patch is likely to conflict with any work I've done so
far on my RC branch.  (But it may be helpful as a guide to me to spot
things that can be converted to the new reference-counted regime more
efficiently.)

-- 
Technology is dominated by two types of people: those who understand what
they do not manage, and those who manage what they do not understand.

demerphq:
@JRaspass sorry mate, I'm going to close this.

JRaspass:
No worries, makes sense.

PR #20849, opened 2023-02-24
Update Trig.pm

jidanni:
Yes, "or about 69 N 89 E, in the frozen wastes of Siberia." but must reveal the crucial missing step involved!

### DIFF
--- a/dist/Math-Complex/lib/Math/Trig.pm
+++ b/dist/Math-Complex/lib/Math/Trig.pm
@@ -669,9 +669,11 @@ straight north being zero, straight east being pi/2).
 
 The midpoint between London and Tokyo being
 
-    use Math::Trig qw(great_circle_midpoint);
+    use Math::Trig qw(great_circle_midpoint rad2deg);
 
     my @M = great_circle_midpoint(@L, @T);
+    sub SWNE { rad2deg( $_[0] ), 90 - rad2deg( $_[1] ) }
+    my @lonlat = SWNE(@M);
 
 or about 69 N 89 E, in the frozen wastes of Siberia.
 


tonycoz:
Could you improve the commit subject please?

Otherwise I think it's good, the committer will need to update `$VERSION`, though you could do that too.

jidanni:
I can't change any of that. It's too late. Please feel welcome to change it for me.

tonycoz:
You can modify the commit locally with `git commit --amend` and force push it, that will update the PR.

khwilliamson:
Although not a part of this PR, I object to the adjacent  term dismissing this spot as "frozen wastes"  Looking at Google Maps, this area in the summer is verdant, with many rivers and lakes.  This point is close to the juncture of two rushing rivers, and less than 20 km from  Lake Kata with an area of 452 square km.   It is part of the https://en.wikipedia.org/wiki/Putorana_Plateau which contains the second largest amount of fresh water in Russia, after Lake Baikal.  And for history buffs, the plateau contains part of the Siberian Traps, implicated by many to the largest known extinction event, at the end of the Permian.

jidanni:
@tonycoz Um, I'm actually just using GitHub.com in a browser, and often on a tiny cellphone.
Wait, I'll try https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/amending-a-commit . Wait, that says
> After you push a commit, the option to amend it is disabled in GitHub Desktop.

So GitHub Browser is probably even more restricted than GitHub Desktop. Also I'm not sure what to change them to anyway, so be my guest.

tonycoz:
Done in e15e200786b11858b8b4d82e91f6a3eeed934f15 and 1bc6b32c00831fcbf45413bd1e7d5ea44bff81ab

PR #20850, opened 2023-02-24
utf8.h: Rmv improper (STRLEN) cast

khwilliamson:
An enum is supposed to be an int.  STRLEN may be too large for that.  (I suspect the cast was inadvertently left in converting from an earlier implementation that was never released.)

Spotted by H. Merijn Brand.


demerphq (approved):

demerphq:
Needs to be rebased to pass test

Issue #20851, opened 2023-02-25
[MSWin32] Devel-Peek failing its test suite.


sisyphus:
**Description**

The issue arises with the current devel release (5.37.9) if and only if I am doing a -Duse64bitint build of 32-bit (x86) perl on MS Windows.
And the failure occurs (in the cleanup) only during the running of the test suite (ext/Devel-Peek/Peek.t).
That failure is reported as:
```
../ext/Devel-Peek/t/Peek.t ...... Dubious, test returned 5 (wstat 1280, 0x500)
All 33 subtests passed
```
If that test script is run outside of the testing harness, it runs fine - though the cleanup, after all tests have been reported as passing, typically takes around 4 seconds.
It makes no difference whether we build with a Microsoft compiler, or with a MinGW compiler.
It makes no difference whether the build is threaded (multi-thread) or unthreaded (perlio) - except that the cleanup at the end of Peek.t happens instantly when it is run outside of the harness on an unthreaded perl.

I would think that this issue is related to:
```
commit a405f1587e7ebdf2d76c4158dee7f80e14bb4003
Author: Yves Orton <demerphq@gmail.com>
Date:   Fri Jan 20 16:19:36 2023 +0100
```
AFAICS, the only change between Devel-Peek-1.32 (perl-5.37.8) and Devel-Peek-1.33 (perl-5.37.9) is in t/Peek.t.
(So I'm thinking that the issue is probably in dump.c.)

Cheers,
Rob
```
Summary of my perl5 (revision 5 version 37 subversion 9) configuration:

  Platform:
    osname=MSWin32
    osvers=10.0.22000.1574
    archname=MSWin32-x86-multi-thread-64int
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -D_WIN32_WINNT=0x0a00 -fdiagnostics-color=never -DPERL_TEXTMODE_SCRIPTS -DMULTIPLICITY -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STDIO -fwrapv -fno-strict-aliasing -mms-bitfields'
    optimize='-O2'
    cppflags='-DWIN32'
    ccversion=''
[...]

PR #20853, opened 2023-02-25
Fix a variety of build warnings, some clang related, some 32-bit build related, some from the ppport.h enable

demerphq:
I originally set out to see if I could replicate https://github.com/Perl/perl5/issues/20851 by building with 32 bit clang. But I could replicate that issue, but I did find these build warnings.

```
git diff --stat origin/blead
 class.c                   |  4 ++--
 dist/Time-HiRes/HiRes.pm  |  2 +-
 dist/Time-HiRes/HiRes.xs  |  2 ++
 dump.c                    |  2 +-
 ext/XS-APItest/APItest.pm |  2 +-
 ext/XS-APItest/APItest.xs | 12 ++++++------
 regcomp.c                 |  4 ++--
 7 files changed, 15 insertions(+), 13 deletions(-)
```

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -2279,10 +2279,6 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
             RExC_parno_to_logical_next[parno]= RExC_logical_to_parno[logical_parno];
             RExC_logical_to_parno[logical_parno] = parno;
         }
-        if (0)
-        for( int parno = 1; parno < RExC_total_parens ; parno++ )
-            PerlIO_printf(Perl_debug_log,"%d -> %d -> %d\n",
-                    parno, RExC_parno_to_logical[parno], RExC_parno_to_logical_next[parno]);
         RExC_rx->logical_to_parno = RExC_logical_to_parno;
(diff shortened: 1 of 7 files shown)


regcomp.c
```diff
@@ -2281,8 +2281,8 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
         }
         if (0)
         for( int parno = 1; parno < RExC_total_parens ; parno++ )
-            PerlIO_printf(Perl_debug_log,"%d -> %d -> %d\n",
-                    parno, RExC_parno_to_logical[parno], RExC_parno_to_logical_next[parno]);
+            PerlIO_printf(Perl_debug_log,"%" IVdf " -> %" IVdf " -> %" IVdf "\n",
```
Leont: Wouldn't it be better to use `I32df` instead of casting?

demerphq: It is debug code that I left in for myself to debug this feature if needed, notice the `if(0)`. I dont think it matters at all.

demerphq: Anyway, i changed the patch to just remove the entire if(0) function. Devel::Peek has this now.

PR #20858, opened 2023-02-26
Switch to Refcounted Stack

iabyn:
second attempt. Not for merge yet.


jkeenan:
> second attempt. Not for merge yet.

Thanks for announcing this.  Can you remove the `smoke-me/davem/rc_stack1` branch from the repository now?

demerphq:
You don't need to create a new PR Dave, in fact, you should not do so. You can and should force push to the old one. Note, you shouldn't really use a smoke-me branch for your PR, keep your topic branch for your PR  separate and distinct from the smoke-me version.

So for instance let's say you had followed the advice about keeping smoke-me branches distinct from PR branches and you had pushed your original PR as davem/rc_stack you might do this:

```
# update to latest
git fetch
# checkout topic branch
git checkout davem/rc_stack
# rebase topic branch to latest blead
git rebase origin/blead
#hack hack hack
# push the topic branch and overwrite old status, start github CI 
git push -f
# push the latest state of topic branch for smoking
git push origin HEAD:smoke-me/davem/rc_stack2
# delete the previous smoke branch
git push origin :smoke-me/davem/rc_stack
```
The reason I say don't use a smoke-me for your PR is I believe that the *smoke-me* infra doesn't deal with force pushes, but i know that the PR infra on github *does*. The recipe above  will avoid spamming github with new PRs, but still let you push new versions for smoking.

demerphq:
And actually, I wouldn't bother with smoke-me branches at all until the github Ci is green, even then these days most folks dont bother. Bit if the Ci fails then certainly so shall the smoke-me, but CI will complete in about an hour, two max. The feedback loop on smoke-me branches is much slower.

demerphq:
@iabyn do you think maybe we should edit the title of this PR to something like "Switch to Refcounted Stack"? I realize that you already publicized it by this name, but really you should publicize the *branch*, not the PR title. Github autosets the title from the topic branch, but for perl that usually isnt a great default.

demerphq:
I have been running some tests to install modules with this branch with PERL_RC_STACK enabled, and with latest blead as a comparison. 

So far I see the following modules broken with PERL_RC_STACK:

- Cpanel::JSON::XS  (Cpanel-JSON-XS-4.35) multiple tests fail due to: perl5.37.10: scope.c:1524: Perl_leave_scope: Assertion `0' failed.
- IO::Socket::SSL (IO-Socket-SSL-2.081) multiple tests SEGV
- HTML::Tree (HTML-Tree-5.07)  multiple tests fail due to: perl5.37.10: sv.c:6629: Perl_sv_clear: Assertion `SvTYPE(sv) != (svtype)SVTYPEMASK' failed.
- Text::CSV_XS (Text-CSV_XS-1.49) Multiple segfaults
- indirect (http://www.cpan.org/authors/id/V/VP/VPIT/indirect-0.39.tar.gz) Lots of
```
panic: attempt to copy freed scalar 5651dd619e80 to 5651dd5c2b18 at /home/yorton/perl5/perlbrew/perls/perl_rc_stack/lib/5.37.10/Test/More.pm line 1555, <DATA> chunk 2.
Attempt to free unreferenced scalar: SV 0x563120be0c48, Perl interpreter: 0x563120bab2a0
Attempt to free temp prematurely: SV 0x55ecc6e23870, Perl interpreter: 0x55ecc6d03a50.
perl5.37.10: sv.c:6629: Perl_sv_clear: Assertion `SvTYPE(sv) != (svtype)SVTYPEMASK' failed.
```

haarg:
indirect uses a bunch of custom OPs, so failures there are expected.

dur-randir:
I've started some preliminary testing and together with the usual suspects I've found the following interesting failures:

- Test::Trap
- Net::Server
- Number::Format
- Sub::Attribute
- Cache::Memcached::Fast
- Devel::Cover

demerphq:
> Text::CSV_XS

this was a bug in the XS code that the branch helped bring to light. I expect that for non PP related XS code this will be a pattern.

Tux:
Text-CSV_XS-1.50 has been uploaded to CPAN. This is the performance graph comparing speed of a single release over different versions of perl. It clearly shows the performance hit (up is fast, down is slow)

demerphq:
@iabyn we can set it up so in this branch CI runs with the define and without so you dont need that last patch. Ill look into a patch for you. (Until then it makes sense to keep it!)

iabyn:
On Wed, Mar 01, 2023 at 06:07:29AM -0800, H.Merijn Brand wrote:
> 
> Text-CSV_XS-1.50 has been uploaded to CPAN. This is the performance
> graph comparing speed of a single release over different versions of
> perl. It clearly shows the performance hit (up is fast, down is slow)

It's a bit difficult to judge without vertical axis labels and/or an
indication of where the origin is.


-- 
   -- Larry Wall

demerphq:
@iabyn I just dug into the IO::Socket::SSL errors a bit, and realized that the module is pure perl, yet it still segfaults. Obviously it depends on several XS modules, Net::SSLeay for instance, but they dont SEGV in test, and it does. 

```
Program received signal SIGSEGV, Segmentation fault.
0x000055555580d1cc in Perl_leave_adjust_stacks (my_perl=0x555555c602a0, from_sp=0x555555d2e758, to_sp=0x555555d2e750, 
    gimme=3 '\003', pass=1) at pp_hot.c:5417
warning: Source file is more recent than executable.
5417	                if (SvTEMP(sv))
(gdb) bt
#0  0x000055555580d1cc in Perl_leave_adjust_stacks (my_perl=0x555555c602a0, from_sp=0x555555d2e758, 
    to_sp=0x555555d2e750, gimme=3 '\003', pass=1) at pp_hot.c:5417
#1  0x00005555558d6e23 in Perl_pp_leave (my_perl=0x555555c602a0) at pp_ctl.c:2430
#2  0x000055555578e229 in Perl_runops_debug (my_perl=0x555555c602a0) at dump.c:2867
#3  0x00005555557ea756 in Perl_runops_wrap (my_perl=0x555555c602a0) at run.c:112
#4  0x00005555555eb69c in Perl_call_sv (my_perl=0x555555c602a0, sv=0x555555c959f8, flags=13) at perl.c:3151
#5  0x00005555555f611b in Perl_call_list (my_perl=0x555555c602a0, oldscope=2, paramList=0x555555c95ab8) at perl.c:5262
#6  0x00005555555ca547 in S_process_special_blocks (my_perl=0x555555c602a0, floor=102, 
    fullname=0x555555ca2f88 "BEGIN", gv=0x555555c95a70, cv=0x555555c959f8) at op.c:11107
#7  0x00005555555c9ae6 in Perl_newATTRSUB_x (my_perl=0x555555c602a0, floor=102, o=0x555555ca2e40, proto=0x0, 
    attrs=0x0, block=0x555555ca2e08, o_is_gv=false) at op.c:10944
#8  0x00005555555ba553 in Perl_utilize (my_perl=0x555555c602a0, aver=1, floor=102, version=0x0, idop=0x555555ca2498, 
    arg=0x0) at op.c:7843
#9  0x0000555555682230 in Perl_yyparse (my_perl=0x555555c602a0, gramtype=258) at perly.y:461
#10 0x00005555555e8a5c in S_parse_body (my_perl=0x555555c602a0, env=0x0, xsinit=0x55555559b3ec <xs_init>)
    at perl.c:2611
#11 0x00005555555e67f4 in perl_parse (my_perl=0x555555c602a0, xsinit=0x55555559b3ec <xs_init>, argc=3, 
    argv=0x7fffffffd928, env=0x0) at perl.c:1917
#12 0x000055555559b2e8 in main (argc=3, argv=0x7fffffffd928, env=0x7fffffffd948) at perlmain.c:106
```
(you can ignore the warning about source code being newer than executable. I build this from your branch a while back and have since changed branch a few times.)

Tux:
[Text-CSV_XS-1.50-perl.txt](https://github.com/Perl/perl5/files/10862423/Text-CSV_XS-1.50-perl.txt)
Here are the raw numbers that were used for the graph

demerphq:
Cpanel::JSON::XS uses SAVESTACK_POS(), from your comment in scope.c:

```
#ifdef PERL_RC_STACK
            /* DAPM Jan 2023. I don't think this save type is used any
             * more, but if some XS code uses it, fail it for now, as
             * it's not clear to me what perl should be doing to stack ref
             * counts when arbitrarily resetting the stack pointer.
             */
            assert(0);
#endif
```

I decided to see what would happen if I simply removed it, and all tests passed. 

All of the uses have the same form:

```
$ grep SAVESTACK_POS XS.xs 
          ENTER; SAVETMPS; SAVESTACK_POS (); PUSHMARK (SP);
              ENTER; SAVETMPS; SAVESTACK_POS (); PUSHMARK (SP);
          ENTER; SAVETMPS; SAVESTACK_POS (); PUSHMARK (SP);
    ENTER; SAVETMPS; SAVESTACK_POS (); PUSHMARK (SP);
```

I have never seen that before, and I dont know what we should do about it. Cpanel::JSON::XS is a fork of code (JSON::XS) by someone who has a history of not reacting well to internals changes, so i think we can assume there will be some gnashing of teeth over this. :-( Any idea what kind of issues we should expect by removing this? @rurban maintains the Cpanel::JSON::XS fork, maybe he has some insight here.

Tux:
If it will effectively be a no-op, it will be easy to put in `ppport.h` and all XS using it will be safe instantly :)

haarg:
JSON::XS added calls to `SAVESTACK_POS` in version 3.02 and removed them in version 4.0.

Those calls were brought into Cpanel::JSON::XS in version 3.0212 by merging in changes from JSON::XS. But apparently the removals have never been brought over.

demerphq:
Regarding HTML::Tree. The error is inconsistent, sometimes it does not occur. But when it occurs it always occurs in the same test file and test, construct_tree.t just after test 44. It also seems to require that HTML::FormatText LWP::UserAgent are installed, although i am less certain of that.

```
gdb --args perl -Mblib t/construct_tree.t
....
ok 44 - found Gisle
perl: sv.c:6629: Perl_sv_clear: Assertion `SvTYPE(sv) != (svtype)SVTYPEMASK' failed.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c27859 in __GI_abort () at abort.c:79
#2  0x00007ffff7c27729 in __assert_fail_base (fmt=0x7ffff7dbd588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x555555bcd850 "SvTYPE(sv) != (svtype)SVTYPEMASK", file=0x555555bc1b10 "sv.c", line=6629, 
    function=<optimized out>) at assert.c:92
#3  0x00007ffff7c38fd6 in __GI___assert_fail (assertion=0x555555bcd850 "SvTYPE(sv) != (svtype)SVTYPEMASK", 
    file=0x555555bc1b10 "sv.c", line=6629, function=0x555555bd51f8 <__PRETTY_FUNCTION__.25418> "Perl_sv_clear")
    at assert.c:101
#4  0x000055555584d56d in Perl_sv_clear (my_perl=0x555555c602a0, orig_sv=0x555556cb6f78) at sv.c:6629
#5  0x0000555555850af8 in Perl_sv_free2 (my_perl=0x555555c602a0, sv=0x555556cb6f78, rc=1) at sv.c:7233
#6  0x00005555558c7a07 in Perl_SvREFCNT_dec (my_perl=0x555555c602a0, sv=0x555556cb6f78) at sv_inline.h:689
#7  0x00005555558d0691 in Perl_pp_mapwhile (my_perl=0x555555c602a0) at pp_ctl.c:1271
#8  0x000055555578e229 in Perl_runops_debug (my_perl=0x555555c602a0) at dump.c:2867
#9  0x00005555557ea756 in Perl_runops_wrap (my_perl=0x555555c602a0) at run.c:112
#10 0x00005555555e9bbf in S_run_body (my_perl=0x555555c602a0, oldscope=1) at perl.c:2799
#11 0x00005555555e909e in perl_run (my_perl=0x555555c602a0) at perl.c:2719
#12 0x000055555559b32d in main (argc=3, argv=0x7fffffffd928, env=0x7fffffffd948) at perlmain.c:127

```

pp_ctl.c: 1271 is the last line of this code:
```
        if (SvPADTMP(src)) {
            SV *newsrc = sv_mortalcopy(src);
            PL_stack_base[PL_markstack_ptr[-1]] = newsrc;
#ifdef PERL_RC_STACK
            SvREFCNT_inc_simple_void_NN(newsrc);
            SvREFCNT_dec(src);
#endif
            src = newsrc;
        }
        if (SvPADTMP(src)) {
            src = sv_mortalcopy(src);
        }
        SvTEMP_off(src);
        DEFSV_set(src);
```

DEFSV_set is defined as 
```
# define DEFSV_set(sv) \
    (SvREFCNT_dec(GvSV(PL_defgv)), GvSV(PL_defgv) = SvREFCNT_inc(sv))
```

iabyn:
As regards SAVESTACK_POS() and Cpanel::JSON::XS(), I think all their uses in that distribution are effective NOOPs. They are all of the form: remember the current stack position, then do call_sv(), process any return values, then LEAVE() (which then restores  the old stack pos). But in every case, before the LEAVE the code manually resets the stack anyway, e.g. the 'SP -= items;' in encode_rv(). And if the code dies, perl does stack unwinding anyway. So NOOP.

One possibility is to just remove SAVESTACK_POS() from that distribution. According to grep.metacpan.org, the only other distribution on CPAN which uses that macro is B-Hooks-XSUB-CallAsOp-0.02, which hasn't had a release since 2009, and doesn't seem to be used anywhere.

The other possibility is to fix perl. When I disabled SAVESTACK_POS(),  I didn't yet have any easy way to fix it. Later I added
rpp_obliterate_stack_to() to the API, which is kind of exactly what leave_scope() needs to implement SAVESTACK_POS(). But I suspect that the concept of SAVESTACK_POS() is kind of broken, and making perl silently and correctly help an XS module to hang itself is just prolonging the agony.

So I'm kind of inclined to say keep SAVESTACK_POS() doing assert(0) and remove it from any disgtributions which still use it.

rurban:
Thanks. I'll happily remove it also. Old cruft

iabyn:
On Wed, Mar 01, 2023 at 07:03:27AM -0800, Yves Orton wrote:
> @iabyn I just dug into the IO::Socket::SSL errors a bit, and realized
> that the module is pure perl, yet it still segfaults. Obviously it
> depends on several XS modules, Net::SSLeay for instance, but they dont
> SEGV in test, and it does. 

I've reduced it to a small pure perl test code. Looking into it now.

-- 
The Enterprise is involved in a bizarre time-warp experience which is in
some way unconnected with the Late 20th Century.
    -- Things That Never Happen in "Star Trek" #14

iabyn:
On Wed, Mar 01, 2023 at 12:06:04PM -0800, iabyn wrote:
> I've reduced it to a small pure perl test code. Looking into it now.

Fixed by the commit: Perl_runops_wrap((): don't mortalise NULLs

-- 
    -- Churchill on Montgomery

iabyn:
The HTML::Tree failure was due to croak() trying to copy freed elements from @DB::args, which in turn was caused by pp_caller()'s trick of using elements between AvALLOC() and AvARRRAY() on a shift'ed @_. These two new commits make pp_caller() skip copying such elements to @DB:caller if they have RC==0:

- caller(): don't copy freed SV pointer to @DB::args
- pp_caller(): set @DB::args earlier.

I couldn't reduce HTML::Tree failure to a simple case, but I assume its due to the ref-counted stack altering the timings of when exactly a shifted array element gets freed and/or reallocated.
These two commits make @DB::args generally more robust regardless of PERL_RC_STACK or not.

leonerd:
Since we're now at user-visible-behaviour change freeze, I'll tag this PR with the "defer-next-dev" label to remind us we're not doing it this round, but to take a look next time.

khwilliamson (approved):
I have read through these commits, and they look reasonable to me.  I think it unlikely that someone will have the time to scrutinize them all carefully.  Therefore, at this point in the development cycle, I think that after a suitable period to give people a chance to comment, that these be merged

iabyn:
On Sat, Aug 12, 2023 at 04:40:30PM -0700, Karl Williamson wrote:
> @khwilliamson approved this pull request.
> 
> I have read through these commits, and they look reasonable to me.  I
> think it unlikely that someone will have the time to scrutinize them all
> carefully.  Therefore, at this point in the development cycle, I think
> that after a suitable period to give people a chance to comment, that
> these be merged

Yeah, most of the commits are unchanged from back in March, apart from
tweaks to rebase against blead. The last few commits fix up some failures
triggered by CPAN modules (e.g. edge cases with @_ and/or @DB::args), then
there are few commits to address specific points raised during earlier
review. In particular,

Added: t/op/refstack.t - this test file contains the failing code samples
from all 68 GH tickets which were part of the RT 'stack not ref counted
issues' meta ticket. All these tests pass under PERL_RC_STACK (and are
skipped on default builds).

Added support for calling RC-aware XS subs. Note that this doesn't add
support to the XS parser for flagging XS subs as RC-aware, it merely adds
a mechanism to pp_entersub() etc to call those flagged XS functions
without wrapping them.

After merging, my pain will become everyone else's pain: any new pp()
functions you write (I'm looking at you, Paul!) should be in the new
style, using rpp_push_1() rather than PUSHs() etc.

Other than that, in theory blead shouldn't be much affected on default
builds - just a few edge cases around the handling of @DB::args etc have
been tweaked.

My next major piece of work on this project will be to unwrap many of the
lightweight pp() functions, such as pp_const(). This should claw back a
lot of the performance drop.

I propose that we should add an additional GH build check thinggy
configuration (i.e. all the things that are run when you push or update a
PR):

    -Duseithreads -DDEBUGGING -Accflags='-DPERL_RC_STACK'

to ensure it doesn't bitrot.  I don't know who is responsible for updating
such things.

-- 
Hofstadter's Law: It always takes longer than you expect, even when you
take into account Hofstadter's Law.

demerphq:
On Mon, 14 Aug 2023, 22:07 iabyn, ***@***.***> wrote:

> On Sat, Aug 12, 2023 at 04:40:30PM -0700, Karl Williamson wrote:
> > @khwilliamson approved this pull request.
> >
> > I have read through these commits, and they look reasonable to me. I
> > think it unlikely that someone will have the time to scrutinize them all
> > carefully. Therefore, at this point in the development cycle, I think
> > that after a suitable period to give people a chance to comment, that
> > these be merged
>
> Yeah, most of the commits are unchanged from back in March, apart from
> tweaks to rebase against blead. The last few commits fix up some failures
> triggered by CPAN modules (e.g. edge cases with @_ and/or @DB::args), then
> there are few commits to address specific points raised during earlier
> review. In particular,
>
> Added: t/op/refstack.t - this test file contains the failing code samples
> from all 68 GH tickets which were part of the RT 'stack not ref counted
> issues' meta ticket. All these tests pass under PERL_RC_STACK (and are
> skipped on default builds).
>
> Added support for calling RC-aware XS subs. Note that this doesn't add
> support to the XS parser for flagging XS subs as RC-aware, it merely adds
> a mechanism to pp_entersub() etc to call those flagged XS functions
> without wrapping them.
>
> After merging, my pain will become everyone else's pain: any new pp()
> functions you write (I'm looking at you, Paul!) should be in the new
> style, using rpp_push_1() rather than PUSHs() etc.
>
> Other than that, in theory blead shouldn't be much affected on default
> builds - just a few edge cases around the handling of @DB::args etc have
> been tweaked.
>
> My next major piece of work on this project will be to unwrap many of the
> lightweight pp() functions, such as pp_const(). This should claw back a
> lot of the performance drop.
>
> I propose that we should add an additional GH build check thinggy
> configuration (i.e. all the things that are run when you push or update a
> PR):
>
> -Duseithreads -DDEBUGGING -Accflags='-DPERL_RC_STACK'
>
> to ensure it doesn't bitrot. I don't know who is responsible for updating
> such things.
>

It's easy, patch the workflow file in .github. you should do it yourself
imo.

Yves

>

hvds:
@iabyn if you think this is stable enough, it may be worth inviting @dur-randir to throw some fuzzing at it either before or after it is merged.

iabyn:
Now merged into blead.

inline.h
```diff
@@ -380,6 +398,400 @@ Perl_POPMARK(pTHX)
...
+Perl_rpp_extend(pTHX_ SSize_t n)
+{
+    PERL_ARGS_ASSERT_RPP_EXTEND;
+
+    EXTEND_HWM_SET(PL_stack_sp, n);
+#ifndef STRESS_REALLOC
+    if (UNLIKELY(_EXTEND_NEEDS_GROW(PL_stack_sp, n)))
+#endif
+    {
+        (void)stack_grow(PL_stack_sp, PL_stack_sp, n);
```
bulk88, line 421: This is wrong, and the prototype of `PERL_STATIC_INLINE void Perl_rpp_extend(pTHX_ SSize_t n)` wrong. The retval of `stack_grow()` is being ignored, and all `pp_*()` ops that use this static inline have to re-fetch the ptr inside `PL_stack_sp`, either from `my_perl` or the OS symbol tables (no thds perl) if this grow branch executes, instead of using the new `SV**` ptr delivered in a register from `stack_grow()`.

PR #20867, opened 2023-02-28
Attribute-Handlers: Add and fix Makefile.PL

haarg:


dist/Attribute-Handlers/Makefile.PL
```diff
@@ -0,0 +1,24 @@
+use 5.006;
+use strict;
+use ExtUtils::MakeMaker;
+WriteMakefile(
+  NAME          => 'Attribute::Handlers',
+  VERSION_FROM  => 'lib/Attribute/Handlers.pm',
+  ABSTRACT_FROM => 'lib/Attribute/Handlers.pm',
+  INSTALLDIRS   => INSTALLDIRS => ( "$]" >= 5.011 ? 'site' : 'perl' ),
```
demerphq: i think INSTALLDIRS is duplicated.

haarg:
I think this should really be done for everything in `dist`. As it is, many don't have a Makefile.PL or other packaging files, requiring external repositories to exist to track them. This makes patching the files a horrible pain in the ass, since you have to find the repository that contains the files you care about. But also often those repositories are long dead as the authors have moved on.

dist/Attribute-Handlers/.gitignore
```diff
@@ -0,0 +1 @@
+!/Makefile.PL
```
demerphq, line 1: Why the leading slash?

demerphq (changes requested):

demerphq:
Can you squash that down @haarg?  5 commits for this seems a lot.  Why not just one?

dist/Attribute-Handlers/Makefile.PL
haarg: Fixed.

dist/Attribute-Handlers/.gitignore
haarg: It will only match the file in this immediate directory, rather than recursively. And I only care about the one file. I just always write gitignore statements like this unless there's a good reason not to.

demerphq: Ok, makes sense to me.

demerphq (approved):

demerphq:
> I think this should really be done for everything in dist

Concur. How can I help?

jkeenan:
`MANIFEST` needs updating; then `make test_porting` and it should be good to go.

demerphq:
See https://github.com/Perl/perl5/issues/20874

demerphq:
I squashed this down myself.

PR #20868, opened 2023-02-28
dist/constant - synchronize Changes and Makefile.PL into core

demerphq:
This requires a version bump of course, thus then necessitating a new CPAN release.

See https://github.com/Perl/perl5/pull/20867#issuecomment-1448422564 where @haarg got this ball rolling. 

Ideally we should have everything in dist that we need to recreate the release. constant.pm doesnt seem to have a stand alone repo.

Fixes this distribution for https://github.com/Perl/perl5/issues/20874

### DIFF
--- a/dist/constant/Makefile.PL
+++ b/dist/constant/Makefile.PL
@@ -1,0 +1,30 @@
+use 5.006;
+use strict;
+use ExtUtils::MakeMaker;
+
+my %prereq = (
+    "Test::More" => 0,
+);
+
+WriteMakefile(
+    NAME            => 'constant',
+    LICENSE         => 'perl',
+    AUTHOR          => 'Sebastien Aperghis-Tramoni <sebastien@aperghis.net>',
+    VERSION_FROM    => 'lib/constant.pm',
+    ABSTRACT_FROM   => 'lib/constant.pm',
+    INSTALLDIRS     => ( $] < 5.011 ? 'perl' : 'site' ),
+    PL_FILES        => {},
+    PREREQ_PM       => \%prereq,
+    META_MERGE          => {
+        resources       => {
+            repository  => 'https://github.com/Perl/perl5.git',
+            license     => 'http://dev.perl.org/licenses/',
+            homepage    => 'https://metacpan.org/module/constant',
+            irc         => 'irc://irc.perl.org/#p5p',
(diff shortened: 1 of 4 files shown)


dist/constant/Makefile.PL
```diff
@@ -0,0 +1,30 @@
...
+    PL_FILES        => {},
+    PREREQ_PM       => \%prereq,
+    META_MERGE          => {
+        resources       => {
+            repository  => "git://perl5.git.perl.org/perl.git",
+            license     => 'http://dev.perl.org/licenses/',
+            homepage    => 'https://metacpan.org/module/constant',
+            irc         => 'irc://irc.perl.org/#p5p',
+            mailinglist => 'http://lists.perl.org/list/perl5-porters.html',
+            bugtracker  => "https://rt.perl.org/rt3/Search/Results.html?Query=Queue='perl5' AND Content LIKE 'module=constant' AND (Status='open' OR Status='new' OR Status='stalled')",
```
tonycoz: Out of date repository and bugtracker URLs.

demerphq: Fixed, I think. Please scrutinize carefully. I am not sure I did it right.

dist/constant/Makefile.PL
```diff
@@ -0,0 +1,30 @@
...
+    PL_FILES        => {},
+    PREREQ_PM       => \%prereq,
+    META_MERGE          => {
+        resources       => {
+            repository  => 'https://github.com/Perl/perl5.git',
+            license     => 'http://dev.perl.org/licenses/',
+            homepage    => 'https://metacpan.org/module/constant',
+            irc         => 'irc://irc.perl.org/#p5p',
+            mailinglist => 'http://lists.perl.org/list/perl5-porters.html',
+            bugtracker  => 'https://github.com/Perl/perl5/Issues',
```
tonycoz: Lower case "I" in issues, this one 404s.

The rest look fine.

It's possible to expand the resources key out by switching to meta-spec 2, (see `cpan/Scalar-List-Utils/Makefile.PL` for example) but I'm mostly trying to pick up obvious errors.  Refactoring to make them all meta-spec 2 seems out of scope for me.

demerphq: Fixed. Re: meta-spec 2, I think in the future we should look at making these Makefile.PL's more consistent, but I agree it is not in scope now.

PR #20870, opened 2023-02-28
dist/Filter-Simple - synchronize Makefile.PL and Changes from CPAN

demerphq:
No version bump was required apparently (I am a bit surprised, but not that much)

### DIFF
--- a/dist/Filter-Simple/Makefile.PL
+++ b/dist/Filter-Simple/Makefile.PL
@@ -1,0 +1,16 @@
+require 5.006; # uses 'our'
+use strict;
+use ExtUtils::MakeMaker;
+WriteMakefile(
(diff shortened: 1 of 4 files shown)


dist/Filter-Simple/Makefile.PL
```diff
@@ -0,0 +1,16 @@
+use strict;
```
tonycoz, line 2: This should probably have require 5.006, since the module code uses `our`, which eliminates the need to test for ABSTRACT_FROM below.

demerphq: Fixed. Also standardized the INSTALLDIRS and whitespace layout.

tonycoz (approved):

PR #20872, opened 2023-02-28
dist/Locale-Maketext - synch Makefile.PL from CPAN

demerphq:

### DIFF
--- a/dist/Locale-Maketext/Makefile.PL
+++ b/dist/Locale-Maketext/Makefile.PL
@@ -1,0 +1,45 @@
+require 5.006; # uses 'our'
+use strict;
+use ExtUtils::MakeMaker;
+
+WriteMakefile(
+    NAME          => 'Locale::Maketext',
+    VERSION_FROM  => 'lib/Locale/Maketext.pm',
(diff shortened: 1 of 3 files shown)


toddr:
@demerphq is there a reason we're wanting to pull this file in?

haarg:
See https://github.com/Perl/perl5/pull/20867#issuecomment-1448422564 for the reason I want to do this for everything in `dist`.

demerphq:
> See https://github.com/Perl/perl5/pull/20867#issuecomment-1448422564 for the reason I want to do this for everything in dist.

And I second that sentiment. I am going to create an issue for this in fact.

dist/Locale-Maketext/Makefile.PL
```diff
@@ -0,0 +1,45 @@
+require 5.004;
```
tonycoz: The module uses `our` and `no warnings;`, so this should be 5.006.

demerphq: Fixed, also made the INSTALLDIRS similar to the other examples you provided.

demerphq:
@tonycoz - is this ok?

tonycoz (approved):

Issue #20874, opened 2023-03-01
dist/ modules have inconsistent build processes and missing build artifacts in the repo


demerphq:
Module:
```
$ find dist -mindepth 1 -maxdepth 1 -type d '!' -exec test -e '{}/Makefile.PL' ';' -print | sort
```
dist/Attribute-Handlers
dist/autouse
dist/constant
dist/Devel-SelfStubber
dist/Dumpvalue
dist/encoding-warnings
dist/Env
dist/Exporter
dist/Filter-Simple
dist/FindBin
dist/I18N-Collate
dist/I18N-LangTags
dist/Locale-Maketext
dist/Math-Complex
dist/Net-Ping
dist/SelfLoader
dist/Term-Complete
dist/Term-ReadLine
dist/Test
dist/Text-Abbrev
dist/Thread-Queue
dist/Thread-Semaphore
dist/Tie-File

Cross referencing with CPAN data we get this:

```
Attribute-Handlers               R/RJ/RJBS/Attribute-Handlers-0.99.tar.gz
autouse                          R/RJ/RJBS/autouse-1.11.tar.gz   
constant                         R/RJ/RJBS/constant-1.33.tar.gz  
Devel-SelfStubber                F/FL/FLORA/Devel-SelfStubber-1.05.tar.gz
Dumpvalue                        F/FL/FLORA/Dumpvalue-1.17.tar.gz
encoding-warnings                A/AU/AUDREYT/encoding-warnings-0.11.tar.gz
Env                              F/FL/FLORA/Env-1.04.tar.gz      
Exporter                         T/TO/TODDR/Exporter-5.77.tar.gz 
Filter-Simple                    S/SM/SMUELLER/Filter-Simple-0.91.tar.gz
FindBin                          T/TO/TODDR/FindBin-1.53.tar.gz  
I18N-Collate                     F/FL/FLORA/I18N-Collate-1.02.tar.gz
I18N-LangTags                    S/SB/SBURKE/I18N-LangTags-0.35.tar.gz
Locale-Maketext                  T/TO/TODDR/Locale-Maketext-1.32.tar.gz
Math-Complex                     Z/ZE/ZEFRAM/Math-Complex-1.59.tar.gz
Net-Ping                         R/RU/RURBAN/Net-Ping-2.75.tar.gz
SelfLoader                       S/SM/SMUELLER/SelfLoader-1.24.tar.gz
Term-Complete                    F/FL/FLORA/Term-Complete-1.402.tar.gz
[...]

PR #20876, opened 2023-03-01
dist/Math-Complex - add missing files from CPAN dist

demerphq:
This fixes this module for https://github.com/Perl/perl5/issues/20874

### DIFF
--- a/dist/Math-Complex/Makefile.PL
+++ b/dist/Math-Complex/Makefile.PL
@@ -1,0 +1,22 @@
+{ use 5.006; }
+
+use ExtUtils::MakeMaker;
+
+WriteMakefile(
+	      NAME            => 'Math::Complex',
+	      VERSION_FROM    => 'lib/Math/Complex.pm',
+              INSTALLDIRS     => ( $] < 5.011 ? 'perl' : 'site' ),
+	      PREREQ_PM	      =>
+	      	{
+		    'Config'		=> 0,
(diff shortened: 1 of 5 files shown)


dist/Math-Complex/Makefile.PL
```diff
@@ -0,0 +1,22 @@
+{ use 5.006; }
+
+use ExtUtils::MakeMaker;
+
+WriteMakefile(
+	      NAME            => 'Math::Complex',
+	      VERSION_FROM    => 'lib/Math/Complex.pm',
+	      INSTALLDIRS     => 'perl',
```
tonycoz: INSTALLDIRS should be conditional on perl version:
```
INSTALLDIRS => ( $] < 5.011 ? 'perl' : 'site' ),
```

tonycoz: Sorry for the duplication, but I thought a comment on each PR would make it easier to track.

tonycoz (approved):

dist/Math-Complex/Makefile.PL
demerphq: Fixed.

demerphq: > Sorry for the duplication, but I thought a comment on each PR would make it easier to track.

Nothing to be sorry about, you were right. It did make it easier to track. :-)

PR #20877, opened 2023-03-01
dist/Net-Ping - add missing build artifacts

demerphq:
Files copied from https://github.com/rurban/Net-Ping

This fixes the module for https://github.com/Perl/perl5/issues/20874

### DIFF
--- a/dist/Net-Ping/Makefile.PL
+++ b/dist/Net-Ping/Makefile.PL
@@ -1,0 +1,75 @@
+use strict;
+use warnings;
+use v5.002; # not using our in the CPAN release
+use ExtUtils::MakeMaker;
+
+my @extras = ();
+my $EUMM_VER = $ExtUtils::MakeMaker::VERSION;
+my @AUTHORS = (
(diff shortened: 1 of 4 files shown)


dist/Net-Ping/Makefile.PL
```diff
@@ -0,0 +1,75 @@
...
+      'meta-spec' => { version => 2 },
+        resources   => {
+            # TODO: 26 old issues still open at RT
+            # https://rt.cpan.org/Public/Dist/Display.html?Name=Net-Ping
+            bugtracker  => 'https://github.com/rurban/Net-Ping/issues',
+            repository  => {
+                type => 'git',
+                url => 'https://github.com/rurban/Net-Ping.git',
+                web => 'https://github.com/rurban/Net-Ping',
+            },
```
tonycoz: These should point to the perl5 github if we're upstream

demerphq: Fixed, also made the INSTALLDIRS consistent with the others.

tonycoz (approved):

PR #20887, opened 2023-03-02
Update Config::Perl::V to version 0.36

demerphq:
Also fix the sync script to deal with .tgz files. 

Adds cpan/Config-Perl-V/t/43_plv53710rcs.t


Tux (approved):
New lines in MANIFEST should add the module/distribution name

PR #20889, opened 2023-03-02
perly.y - Allow :isa in class declaration without a block

johannessen:
My assumption is that the “unit” (or “statement”) syntax of `class` is supposed to be equivalent to the block syntax even in the presence of attributes:

```perl
# Equivalent syntax:
class B :isa(A) ;
class B :isa(A) { }
```

This is currently not in fact the case, see GH issue #20888.

This change copies @leonerd’s addition of `subattrlist` to the relevant _perly.y_ definition in [`ae79058`](https://github.com/Perl/perl5/pull/20647/commits/ae790582414fc0cbc81a9f1a71fc36e055d34d86#diff-0f6ed78c83e75dc7b07a379943afb2b8b62906eda1ec5f44a6b4b854b8d7e2b3) from the block syntax and duplicates it to the “unit” syntax.


leonerd (approved):
LGTM

leonerd:
Oops, I managed to overlook that case. Good catch.

Thanks.

PR #20892, opened 2023-03-02
Fixups to Config::Perl::V update from earlier, and improvements to sync-with-cpan

demerphq:
MANIFEST/Maintainers.pl - updates missed from previous Config::Perl::V update

The MANIFEST file was not updated very artfully in older versions of the
sync script, and the porting tests took so long I bailed out of them
without letting the Maintainers.pl file get updated. And there is no
test that the Maintainers.pl file is up to date (yet) so I didn't
notice.

The Porting/sync-with-cpan script has been updated to be nicer and
faster and avoid these problems in the future.

Porting/sync-with-cpan - Many improvements

- Clean up cpan directory before starting (after giving the user a
  chance to bail out first).
- Use the "make" (now run_make) infra for all testing, this ensures that
  it is all logged in a standard way, and that it all respects the
  various parallelism flags that are passed in. We also extract the
  test summary data if there is a failure. This massively speeds up the
  process when TEST_JOBS is set.
- Always update the MANIFEST so that the lines for the files we have
  synced show that they were synced from this tarball.
- Hint to the user that they can use `tail -F mail.log` to watch what
  is going on if they wihs.
- Update Maintainers.pl early. A common issue with the old version of
  the script was that if people bailed out early the maintainer data
  would not be updated.

Various other minor tweaks not worth mentioning here.

### DIFF
--- a/Porting/sync-with-cpan
+++ b/Porting/sync-with-cpan
@@ -97,7 +97,40 @@ from the filename -- but can be overwritten by the C<--version> option.
 
 =item C<--jobs> I<N>
 
-When running C<make>, pass a C<< -jI<N> >> option to it.
+When running C<make>, pass a C<< -jI<N> >> option to it to enable
+parallel building.
+
+Note that you can also set C<< TEST_JOBS=I<N> >> in the environment
+to enable parallel *testing* on top of parallel *building*.
+
+=item C<--yes>
+
+Just continue at all places where we would normally ask for the user
+to hit enter or hit CTL-C, with the exception of cases related to
+CUSTOMIZED distributions, where this option will cause the update to
+exit immediately unless the C<--force> option has also been used.
+
+=item C<--force>
+
+Do things we normally would refuse to do.
+
+=item C<--tarball>
+
+Use a predownloaded tarball and not one from CPAN.
+
+=item C<--version>
+
+Sync with a specific version, not the latest on CPAN.
+
+=item C<--no-test>
+
+=item C<--nt>
+
+Do not run tests. This is helpful for bulk updates.
+
+=item C<--help>
+
+Show help.
 
 =back
 
@@ -154,14 +187,17 @@ die "This does not look like a top level directory"
 die "Please run Configure before using $0\n"
     if !WIN32 && !-f "Makefile";
 
+#these are populated by Porting/Maintainers.pl
 our @IGNORABLE;
 our %Modules;
+our %DistName;
 
 use autodie;
 
 require "./Porting/Maintainers.pl";
 
 my $MAKE_LOG = 'make.log';
+unlink $MAKE_LOG if -e $MAKE_LOG;
 
 my %IGNORABLE    = map {$_ => 1} @IGNORABLE;
 
@@ -186,11 +222,13 @@ sub usage
 GetOptions ('tarball=s'  =>  \my $tarball,
             'version=s'  =>  \my $version,
             'jobs=i'     =>  \my $make_jobs,
-             force       =>  \my $force,
-             help        =>  sub { usage 0; },
-             ) or  die "Failed to parse arguments";
+            'yes'        =>  \my $yes_to_all,
+            'force'      =>  \my $force,
+            'no-test|nt' =>  \my $no_test,
+            'help'       =>  sub { usage 0; },
+        ) or  die "Failed to parse arguments";
 
-usage 1 unless @ARGV == 1 || @ARGV == 2;
+usage 1 unless @ARGV == 1;
 
 sub find_type_f {
     my @res;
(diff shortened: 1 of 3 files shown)


Porting/sync-with-cpan
```diff
@@ -162,6 +162,7 @@ use autodie;
 require "./Porting/Maintainers.pl";
 
 my $MAKE_LOG = 'make.log';
+unlink $MAKE_LOG if -e $MAKE_LOG;
```
Tux, line 200: does it need the test? just unlink it and ignore any error.

Tux (approved):

Porting/sync-with-cpan
demerphq: the code uses autodie. If it errors out it will die.

demerphq: FWIW, what you suggested is what i did originally. I am not used to using autodie. :-)

demerphq:
@tux I made a minor last update to make sync-with-cpan try to preserve customized description strings and only autoupdate ones that it added itself. I noticed that /some/ cpan modules have some kind of text, might as well leave that alone.

Porting/sync-with-cpan
```diff
@@ -538,23 +600,38 @@ if ($$info {CUSTOMIZED}) {
...
     open my $new, ">", $MANIFEST_NEW
         or die "Failed to open $MANIFEST_NEW for writing: $!\n";
+    my %keep = map +("cpan/$pkg_dir/$_" => 1), keys %new_files;
     my %gone = map +("cpan/$pkg_dir/$_" => 1), @gone;
     while (my $line = <$orig>) {
-        my ($file) = $line =~ /^(\S+)/
-            or die "Can't parse MANIFEST line: $line";
-        print $new $line if !$gone{$file};
+        chomp $line;
+        my ($file, $descr) = split /\t+/, $line;
```
haarg, line 727: Would this be better as `split /\s+/, $line, 2`? While the current MANIFEST file always uses tabs, the format of the file allows any whitespace to divide the file names and the comments.

demerphq: >  the format of the file allows any whitespace to divide the file names and the comments.

No, that isn't correct for the core MANIFEST, it isn't the same format specification as ExtUtils::Manifest uses (assuming you are right and it allows it - I don't necessarily see why it would, on some platforms having spaces in file names in a manifest might actually be desirable). The core MANIFEST MUST be tabs and it cannot be spaces. We have tests to validate this. See https://github.com/Perl/perl5/blob/blead/t/porting/manifest.t#L58

Also for this case it doesn't matter if there are additional tabs in the file as the $descr is only used to check if there is /something/ there at all. FWIW, there is a single line in the current MANIFEST which has additional tabs after the separator tabs, which was obviously a case of duplication:

```
perl -lne'/^(\S+)\t+([^\t]+)\t(.*)/ and print' MANIFEST
cpan/podlators/t/data/basic.cap			podlators test			podlators test
```

I have a patch to fix this. I am working on a patch to improve MANIFEST sorting and  formatting anyway, which will deal with this.

haarg: ExtUtils::Manifest allows spaces, and allows using single quotes around a file name if it contains spaces.

In core, most of the programs that read MANIFEST accept spaces as a separator. A couple use ExtUtils::Manifest. A number also require there to be whitespace or a tab character to read the line properly.

demerphq: >In core, most of the programs that read MANIFEST accept spaces as a separator.

That may be. I would argue that they are just doing it wrong. *shrug*.

I don't plan to change this in this commit. If you want to address this discrepancy I think you should do so in another commit/PR which changes the test file and all the other tools that assume this file uses tabs.  As far as I can tell this is something where EU::Manifest was changed and core was not. IMO that is out of scope for this change.

demerphq: BTW, I don't see why this bothers you. If EU::Manifest and other tools assume that the file is "any whitespace" separated then that is compatible with the core being stricter and assuming it is tab separated. And if EU::Manifest has adopted a more complex file format with quotes around file names with spaces that just means it is harder to process and manage the manifest and your suggested change could actually break things as it would not deal with quoted files. 

On the other hand with core using a simpler set of rules it is easier to deal with, and is still compatible with EU::Manifest. So nothing will break so long as someone doesn't assume you can use EU::Manifest to *write* the manifest file out, and we don't and shouldn't use EU::Manifest to write the manifest file, or sort it, or anything like that (or it would break our tests). So it seems to me that EU::Manifest is just enforcing a super set of the core format. There is no incompatibility there.

haarg: The previous code split on white space, and most of the code dealing with the MANIFEST does the same, so it seemed reasonable behavior to maintain. And having two different MANIFEST formats is a bit weird.

I don't really care much though - my second comment was meant to be informational. And I did find some problems which will be worth addressing after this. Specifically, that a number of the scripts reading MANIFEST expect there to be whitespace or a tab on every entry, which currently isn't true. Not relevant to this PR though.

demerphq: > The previous code split on white space

The previous code did it wrong. :-) I happen to have intimate knowledge of the file format as I have worked on it over the years, including on the tests for it, so to me splitting on spaces was just wrong.  

Which goes to show that an individuals preferred context matters here. I come from the point of view of that core was where the MANIFEST concept started, so it "owns" the format. You have been working on the toolchain so you see ExtUtils::Manifest as the "owner" of the format. At some point they were the same, but have over time drifted apart. But for me I think core is distinct from modules, and it doesnt matter if the rules are slightly different[1]. There are different constraints and objectives for the two environments - eg in core it is easy to say "no spaces in files - *EVER*", but you cant say that out in the wild. 

[1] I have noticed that toolchain gang people tend to take the view that the core should be treated the same as distributions, which I only 2/3rds agree with, i think the problems aren't entirely the same.

> And having two different MANIFEST formats is a bit weird.

When one is a stricter and simplified variant of the other it isn't /that/ weird. The core file came first, and never uses quotes around the files, and forbids spaces in file names, and uses tabs as a separator. EU::Manifest at some point was changed to use quoted filenames and spaces as a separator. But the EU::Manifest format is read-compatible with the core format, and since we don't use it to write there is no problem. I would hope that EU::Manifest writes tabs, and then it could be used for core, but that is a different matter.

> Specifically, that a number of the scripts reading MANIFEST expect there to be whitespace or a tab on every entry, which currently isn't true. 

That is interesting. Please file an issue for that and I'll look into getting that fixed.

> Not relevant to this PR though.

Although perhaps this PR is relevant to solving that problem in that one of the main ways such files end up in the MANIFEST (that I can see anyway) is via sync-with-cpan, and this patch ensures that such lines will have a description. So this PR is part of the solution.

The other thing is I made it so this PR includes functionality so we can "touch up" existing cpan/ distros even though we are technically up to date. I plan to try that on every module in the cpan directory soon.  So it is very possible a bunch of the offending manifest lines will go away.

demerphq: > The previous code split on white space

FWIW, that isn't actually correct. The previous code didn't split at all, it just extracted the filename as a set of non-whitespace characters and ignored the description. So I'd say what it did is more compatible with the core-subset of the EU::Manifest format than it is with the EU::Manifest format itself. If it wanted to emulate that it should handle quoted filenames no?  Regardless I think you could argue that what we are doing now and before isn't /strictly/ speaking correct as \S matches \0. :-) 

The old logic was:
```
        my ($file) = $line =~ /^(\S+)/
            or die "Can't parse MANIFEST line: $line";
```

PR #20893, opened 2023-03-03
dist/if: add Makefile.PL

haarg:


demerphq (commented):
Please squash the commits. We don't need a log of the old state of the Makefile, a single commit will do.

demerphq:
I resolved the conflict on MANIFEST (since I caused it), but i have not squashed the files. I would prefer you did that, or that you ask me to do it.

PR #20895, opened 2023-03-03
tweak MANIFEST sort order, better description indentation

demerphq:
P/manifest_lib.pl - tweak MANIFEST sort order, better description indent

This patch ensures that all files in a common sub-directory or preferably sub-directory pair will have their descriptions "indented" to the same level. Eg, all files in cpan/Foo should have the description line up.

I did some analysis and something like 95% of our files are less than 40 characters.  So short files mostly have their descriptions start on column 40, with the caveat that if they are in a subdirectory

This also changes the sort order so files at the top level of the repo, which are mostly C files, are sorted together, and not with things in cpan/ dist/ or ext/, which are grouped separately.

It has been pointed out that the format for MANIFEST differs from that used by ExtUtils::Manifest, and that maybe we should migrate. That would allow us to use normal spaces as separators instead of tabs, and be more fine grained about how we line up the files. IMO this is a good idea, but out of scope for this patch.

This patch includes an application of the new logic to MANIFEST.


leonerd:
This feels like quite a lot of work has gone into it but what *problem* is it actually attempting to solve? I personally don't find myself ever reading the MANIFEST files, other than when that annoying `t/Porting/` test script tells me I forgot to add a line, and I add it then go on with my day without considering it further. Is there some deeper usecase to the MANIFEST file that I have overlooked?

demerphq:
On Fri, 3 Mar 2023 at 20:40, Paul Evans ***@***.***> wrote:

> This feels like quite a lot of work has gone into it but what *problem*
> is it actually attempting to solve? I personally don't find myself ever
> reading the MANIFEST files, other than when that annoying t/Porting/ test
> script tells me I forgot to add a line, and I add it then go on with my day
> without considering it further. Is there some deeper usecase to the
> MANIFEST file that I have overlooked?
>
Well, I read the MANIFEST file from time to time, for instance for
descriptions of files and the layout inconsistencies bug my eyes.

This change was motivated by recent work getting the build artifacts for
dist/ distributions into the core. Such changes are tedious and manual, and
IMO it is made worse when everything is randomly indented. It was also
motivated because I want to make sync-with-cpan a more powerful and easier
to use tool, and it uses make manisort under the hood. If we sync a cpan
distro and it introduces a new long file id like all the files for that
cpan distro to line up nicely. It just bugs my eyes when it does not.

Anyway, there is no deeper reason other than "we can do better than that",
and "it bugs me when stuff like that isnt clean and managed
automatically".  I like an ordered workspace. I dont like code
like this being messy unnecessarily, and I especially dont like things
being manually managed that can and should be automated.  If someone adds a
long filename, I want a tool to keep the file pretty, I dont want it to bug
me when I look at a diff.  If someone opens the file they should see
something orderly that we are proud of, not a mess that says that we dont
care.  In short, if it is worth having a manifest then it is worth having
one that is easy on the eyes.  (To steal from the old line that if
something is worth doing it is worth doing well.)

Also, right now we are in a controversial code freeze, and all the other
things I want to work on are controversial at some level or another, so I
basically have a few months before I can merge them. So I am knocking out
some of the tasks that I had saved up for a rainy day. I gave up smoking
recently and this keeps my fingers busy. :-)

cheers,
Yves

-- 

wchristian:
While i think a PR should either be in response to clearly motivated bug report, or describe the /achieve that is being foiled in the first post, this looks good to me, and the follow-up explanation seems sensible. :)

leonerd (approved):
Eh.. I guess fine? I don't overly feel it's a huge problem that needs addressing but evidently you do because you've spent the time on it.

LGTM

Issue #20861, opened 2023-02-27
BBC: Blead Breaks Scope::Upper


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.37.6.

-----------------------------------------------------------------

BBC: Blead Breaks Scope::Upper

Please see http://fast-matrix.cpantesters.org/?dist=Scope::Upper

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.37.10:

Configured by cpan at Sun Feb 26 21:41:46 EST 2023.

Summary of my perl5 (revision 5 version 37 subversion 10) configuration:
  Commit id: 10c7b2021d05bc5b69caa39f144a06954fa66d68
  Platform:
    osname=linux
    osvers=5.15.96-0-lts
    archname=x86_64-linux-thread-multi
    uname='linux cjg-alpine3 5.15.96-0-lts #1-alpine smp sat, 25 feb 2023 21:02:54 +0000 x86_64 linux '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall -Duseithreads'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='12.2.1 20220924'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
[...]

PR #20900, opened 2023-03-04
Improve/fix RCPV macro support and docs

demerphq:
scope.c - improved RCPV support, SAVERCPV SAVEFREERCPV

I misnamed some of the RCPV stuff when it was introduced. The macro which I called SAVERCPVFREE should have been called SAVEFREERCPV, and there should also have been a SAVERCPV macro.

There also should have been RCPV_REFCNT_dec() and RCPV_REFCNT_inc() macros.

There was also missing documentation. This patch should fix all of these issues. Since this functionality has never been released in a production perl it should be fine to do these renames, nothing out there should use these macros yet.

I noticed these oversights while fixing Scope::Upper, see also:

https://github.com/Perl/perl5/issues/20861
https://rt.cpan.org/Ticket/Display.html?id=146897


pod/perlguts.pod
```diff
@@ -1889,6 +1889,20 @@ and back.
...
+
+This macro arranges to restore the value of a C<char *> which was
+allocated with a call to C<rcpv_new()> to its previous state when the
+current pseudo block is completed. The pointer stored in C<pv> at the
+time of the call will be refcount incremented and stored on the save
+stack. Later when the current pseudoblock is completed the value stored
+in C<&rcpv> will be refcount decremented, and the previous value
+restored from the savestack and also be refcount decremented.
+
+This is the C<RCPV> equivalent of C<SAVEFREESV()>.
```
leonerd: Did you mean to put this line in the docs for `SAVEFREERCPV()`? It doesn't seem to make sense here

leonerd (commented):
Overall seems good but one question

pod/perlguts.pod
demerphq: Ah, good catch. Thanks. (The naming conventions and layout for these docs is really confusing and irritating.)

demerphq: Anyway, the latest patch should fix this, and add docs for what that should have said SAVEGENERICSV(). (Like I said the naming convention for that is sucky.)

tonycoz (approved):

PR #20902, opened 2023-03-05
locale.c: Remove one use of nl_langinfo_l()

khwilliamson:
The limited POSIX guarantees of thread safety for nl_langinfo_l() aren't
enough for our uses, and I was naive to think that a simple Configure probe
could rule out all possible thread-safety issues that might exist in a
libc call.  I don't remember what the platforms were that falsely tested
ok for the probe, but if it were necessary to find out, revert this
patch, and start a smoke-me test.

What that Configure probe did was find one particular point of
non-safety.  And it turns out various platforms pass that, but don't
have a thread-safe nl_langinfo_l() generally.

There are two calls to nl_langinfo_l() in the code.  This commit removes
one, where the major advantage of using nl_langinfo_l() over plain
nl_langinfo() was efficiency.  There still had to be an alternate
implementation available that used plain nl_langinfo().  Since we can't
guarantee that the _l implementation doesn't have bugs, simply remove
it, and the existing alternative gets automatically used.

The remaining use of nl_langinfo_l() is only when using glibc, and is
disabled by default, requiring an explicit Configure parameter to
enable.  I have never seen a case where the glibc implementation failed
to be thread-safe.  This use may be enabled by default at some point,
but not until early in a development cycle.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -751,7 +751,8 @@ S_less_dicey_bool_setlocale_r(pTHX_ const int cat, const char * locale)
       * all instances of that have been removed */
 #    define QUERYLOCALE_ASSERT(index)                                       \
                         __ASSERT_(isSINGLE_BIT_SET(category_masks[index]))
-#    if ! defined(HAS_QUERYLOCALE) && defined(_NL_LOCALE_NAME)
+#    if ! defined(HAS_QUERYLOCALE) && (   defined(_NL_LOCALE_NAME)          \
+                                       && defined(HAS_NL_LANGINFO_L))
 #      define querylocale_l(index, locale_obj)                              \
             (QUERYLOCALE_ASSERT(index)                                      \
              mortalized_pv_copy(nl_langinfo_l(                              \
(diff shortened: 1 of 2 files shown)


locale.c
```diff
@@ -748,7 +748,8 @@ S_less_dicey_bool_setlocale_r(pTHX_ const int cat, const char * locale)
       * all instances of that have been removed */
 #    define QUERYLOCALE_ASSERT(index)                                       \
                         __ASSERT_(isSINGLE_BIT_SET(category_masks[index]))
-#    if ! defined(HAS_QUERYLOCALE) && defined(_NL_LOCALE_NAME)
+#    if ! defined(HAS_QUERYLOCALE) && (   defined(_NL_LOCALE_NAME)          \
+                                       && defined(HAS_NL_LANGINFO_L))       \
```
tonycoz: final \ makes the following `#define` part of the `#if`, which I think is the cause of the mac os build failure.

tonycoz:
Please link to a ticket or post or something that illustrates the ongoing nl_langinfo_l() thread unsafety.

Another option is to just say the limited POSIX guarantees of thread safety for nl_langinfo_l() aren't enough for our uses, if that's the problem.

locale.c
khwilliamson: thanks

tonycoz (approved):

PR #20903, opened 2023-03-05
locale.c: Split a static function in two

khwilliamson:
A future commit will want to use just the first part of the function.


leonerd (approved):
LGTM

PR #20904, opened 2023-03-05
loc_tools: Print $^E when upon a libc failure

khwilliamson:
This makes debugging easier.

### DIFF
--- a/t/loc_tools.pl
+++ b/t/loc_tools.pl
@@ -206,8 +206,12 @@ sub _trylocale ($$$$) { # For use only by other functions in this file!
 
         # We definitely don't want the locale set to something that is
         # unsupported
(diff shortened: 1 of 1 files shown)


leonerd (approved):
LGTM

(though to be honest I'd have also been quite fine with just one long line. The line-split here looks a bit awkward)

t/loc_tools.pl
```diff
@@ -206,7 +206,8 @@ ($$$$)
 
         # We definitely don't want the locale set to something that is
         # unsupported
-        die "Couldn't restore locale '$save_locale', category $category"
+        die "Couldn't restore locale '$save_locale', category $category;"
+          . " \$^E=$^E"
```
tonycoz: Why `$^E` but not `$!`?

tonycoz (approved):

Issue #101, opened 1999-06-23
Installation Question


p5pRT:
Migrated from [rt.perl.org#905](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=905) (status was 'resolved')

Searchable as RT905$

p5pRT:
### From Eoghan.Maher@motorola.com

and secondly\, 
rather than delete them I moved them to a different directory and then ran
sh Configure but system returns

First let's make sure your kit is complete\.  Checking\.\.\.
Policy\_sh\.SH&#8203;: No such file or directory
config\_h\.SH&#8203;: No such file or directory

THIS PACKAGE SEEMS TO BE INCOMPLETE\.

You have the option of continuing the configuration process\, despite the
distinct possibility that your kit is damaged\, by typing 'y'es\.  If you
do\, don't blame me if something goes wrong\.  I advise you to type 'n'o
and then contact perlbug@&#8203;perl\.com\.

Continue? \[n\] 
ABORTING\.\.\.

I tried to rename the file in the directory but it always looks for the 2\.
Should I just hit yes at the above prompt or should I not delete the files\.
Thanks in Advance for any help u can offer\.
Regards
Eoghan

PR #20916, opened 2023-03-09
Add deprecation subcategories, test deprecations properly including in perldiag and perlderprecation.pod

demerphq:
Currently we use a single warning category for all deprecations. This is dangerous in that it allows a new deprecation to sneak in when the user only wanted to allow a different deprecation, and IMO counterproductive as it means there is no way to know /which/ deprecation someone is silencing in their code and give them good feedback when the feature is no longer supported.  

Having only one category essentially says "if you want to risk using a deprecation then you have to risk not knowing about *any* deprecation", but this does not make sense. Why should I have to take maximal risk when I want to take a known risk?It makes even less sense when you consider we have at least three categories which are long standing and still not scheduled for removal. If a user wants to silence one of these long standing deprecation types they essentially must risk not being informed about any other category at all. That seems unhelpful and unnecessarily risky.

By having all deprecations be subcategorized in regen/warnings.pl we also get some built in support for tracking the version that a deprecation was added, which can then be used to test that our various files provide sufficient dev cycles before they are removed.

It also makes it more obvious that we have unscheduled deprecations that have been around for a very long time, and which will likely always be present (or so it seems).

This PR adds support for deprecation subcategories, it converts all existing deprecations to use subcategories, it fixes the API's of all deprecation related functionality that I could find to require a category, and it ensures that diag.t knows about these functions so it can test the diagnostics produced, it also validates that perldeprecation.pod metions all existing deprecation categories and that they are documented with sufficient version cycles before they are removed. 

I understand that the PSC has "decided" that it is a bad idea to have deprecation subcategories. I believe that this decision is wrong, and should not have been made without wider discusion. This patch demonstrates why it is a good idea to have these subcategories. It facilitates testing and documenting the categories, it allows our users to manage their risk at a more granular level, and it allows us quick access as to which deprecations are "live" at any given point in time. The counter arguments that I have seen expressed, for instance here: https://github.com/Perl/perl5/pull/20855#issuecomment-1445196362 and https://github.com/Perl/perl5/pull/20916https://github.com/Perl/perl5/pull/20855 do not seem very compelling compared to the advantages enumerated here, and demonstrated by this PR.


handy.h
```diff
@@ -2955,18 +2955,39 @@ last-inclusive range.
...
+
+=for apidoc_section $warning
+=for apidoc Cdm||deprecate|U32 category|"construct"
+Wrapper around Perl_ck_warner_d() to produce a deprecated warning in the given
+category with an appropriate message. The C<construct> argument must be a C string.
+
+=for apidoc Cdm||deprecate_disappears_in|U32 category|"when"|"construct"
+Wrapper around Perl_ck_warner_d() to produce a deprecated warning in the given
+category with an appropriate message that the construct will disappear in a specific
+release. The C<when> and C<construct> arguments must be a C string.
```
leonerd: It might be useful to document or give an example of the expected format of the `"when"` string, otherwise I can foresee all manner of different styles like `v5.40`, `5.40`, `5.40.0`, `version 5.040`, etc...

regen/warnings.pl
```diff
@@ -74,7 +74,15 @@ BEGIN
...
                                 'malloc'        => [ 5.008, DEFAULT_ON],
                            }],
-        'deprecated'    => [ 5.008, DEFAULT_ON],
+        'deprecated'    => [ 5.008, DEFAULT_ON, {
+                                'deprecated::goto_construct'           => [ 5.011003, DEFAULT_ON],
+                                'deprecated::unicode_property_name'    => [ 5.011003, DEFAULT_ON],
+                                'deprecated::dot_in_inc'               => [ 5.025011, DEFAULT_ON],
+                                'deprecated::version_downgrade'        => [ 5.035009, DEFAULT_ON],
+                                'deprecated::delimiter_will_be_paired' => [ 5.035010, DEFAULT_ON],
+                                'deprecated::apostrophe_as_pack_sep'   => [ 5.037009, DEFAULT_ON],
```
leonerd: "pack" might be confused with the `pack()` function. Perhaps better to use "pkg" or "package" spelled out in full?

t/porting/deprecation.t
```diff
@@ -0,0 +1,145 @@
...
+open my $fh, "<", $pod_file
+    or die "failed to open '$pod_file': $!";
+my $removed_in_version;
+my $subject;
+my %category_seen;
+my %subject_has_category;
+my $in_legacy;
+
+while (<$fh>) {
+    if (/=head2 (?|Perl (5\.\d+)(?:\.\d+)?|(Unscheduled))/) { # ignore minor version
```
leonerd: Tiniest of nitpicks: Should these be `^`-anchored?

t/porting/deprecation.t
```diff
@@ -0,0 +1,145 @@
...
+        $removed_in_version = lc $1;
+        if ($removed_in_version eq "5.38") {
+            $in_legacy = 1;
+        }
+    }
+    if (/=head3 (.*)/) {
```
leonerd: And also maybe `elsif`, since we don't expect any one line to match more than one of them

leonerd (commented):
Overall this appears to be going the right way; give or take some minor comments below.

Would appreciate a second opinion from @rjbs or @book first though, to check we're aligned with today's re-thinking on PSC.

handy.h
demerphq: Good suggestion. Will do. I might even make diag.t test this for us.

regen/warnings.pl
demerphq: I wanted to avoid it getting too long, but given we kind /want/ to discourage people from disabling deprecation warnings (especially this one), I think in hindsight spelling it out makes a lot of sense.

t/porting/deprecation.t
demerphq: In practice I dont think it matter that much, but strictly speak yes probably. :-)

t/porting/deprecation.t
demerphq: sure

handy.h
demerphq: Added verbiage to document the form, and a test to enforce it.

t/porting/deprecation.t
demerphq: Done.

t/porting/deprecation.t
demerphq: Done.

regen/warnings.pl
demerphq: Done.

demerphq:
@leonerd you said someone objected to this, but nobody has comment on this PR, nor on the thread on p5p.  Can you ask them to comment here please?

demerphq:
> We discussed offering split-up deprecation categories, so you can "no warnings 'deprecated::.xyz" and re-affirmed that we want to do this.

Based on the PSC "PSC #101, 2023-03-17" comment above, and the fact noone has raised any objections, I am merging this.

PR #20918, opened 2023-03-09
Regex bug fixes, refactoring (macroization) and code improvments

### LINKED ISSUE #10073: Recursive back references do not always work

demerphq:
This PR sequence includes the non-controversial parts of #20677 and #20747.

It includes multiple clean up patches, regex engine bug fixes, and other improvements, including an important patch to simplify and standardize the structures we use to represent regexp opcodes (regops). 

It does *not* include the controversial parts of the above PR's: neither the start_new/end_new pairing, nor the creation of an RXMO object (intended to separate out the compiled form of a regex from the results from that form). It does include many changes which are intended to facilitate the latter, such as wrapping the various regex structure accessors in macros so that they can be used to "hide" the split in the future. 

The patch marked "regexec.c - incredibly inefficient solution to backref problem" and its much more efficient successor "regexec.c - make REF into a backtracking state" fixes #10073 - RT72020

The patch marked "regexec.c - teach BRANCH and BRANCHJ nodes to reset capture buffers" adds tests and todo tests for #18865 

The patch marked "regcomp.c - track parens related to CURLYX and CURLYM" includes prep work for tracking how parens are before and within a quantified group, and tests for various bug reports we have related to inconsistencies between CURLYM and CURLYX.

This PR either fixes or contains preparation to fix the following bugs: #20661, #19615, #18865, #14848, #13619, #10073, #8267 

I would like to get this merged soon so that i can rebase the other two branches on top. @iabyn and I have been in correspondance about the changes in #20677 and for now I would like to move forward with most, but not all of that PR, this PR excludes the parts which @iabyn found to be controversial.  I believe that nothing left in this PR is or should be controversial.


khwilliamson (commented):
The two commits that mention fix up should be squashed.  I thought I was commenting on individual commits, but things showed up here.

khwilliamson (commented):
Typos in commit message

> This patch is very crude. It simple pushes the state of the parens and creates
     and unwind point
should be 'simply'
and should be 'an'

khwilliamson:
I spent an hour-ish looking at this.  The changes to the parts of the engine I'm familiar with look reasonable.  I have never spent much time looking at the engine execution stack.  Everything looks plausible, so I think this can be merged.

demerphq:
@khwilliamson fixups applied. Thanks for the review.

iabyn:
On Mon, Mar 13, 2023 at 06:26:20AM -0700, Yves Orton wrote:
> Merged #20918 into blead.

Sorry, I'm a bit late with a code review. In short, it looks fine, but
with the following niggles, listed by commit title.  [By fine, I mostly
mean the runtime parts - I'm just assuming the regcomp.c parts are ok too
because I'm scared of anything in that file.]


* regcomp.c - track parens related to CURLYX and CURLYM

In Perl_rxres_save(), it changes (twice)
    Newx()s RX_NPARENS(rx)_1) * 2
to '* 4', but I assume that that is vestigial from the removed
start/end changes, and needs to be changed back to '* 2' for now.

t/re/re_tests:
    adds a comment: 'was bar-foo-bar prior to 5.37.7'
    presumably this should say 5.37.10 ?


* regexec.c - teach BRANCH and BRANCHJ nodes to reset capture buffers

IMHO, the new CAPTURE_CLEAR() macro would be better implemented as an
inline static function. Easier to debug, single step through etc.


* regexec.c - incredibly inefficient solution to backref problem

In the curlyx struct part of the union, both cp and lastcp are commented
as: /* remember current savestack index */
Presumably it should be only one or the other?
(Other structs in the union comment on either cp or lastcp or neither.
Its all a bit inconsistent, and doesn't make it clear what the difference
is between cp and lastcp fields)


* regexec.c - make REF into a backtracking state

In regcomp.h, the comment above RE_PESSIMISTIC_PARENS doesn't make it
clear enough (to me, anyway, without also reading the commit message) that
enabling it *guarantees correct* (albeit very slow) behaviour, while
disabling it *optimises away* unnecessary saving of paren state, but might
contain bugs. Or to put it another way, its not clear what "aggressive"
means in that context.

In regcomp_internal.h, this added define has no comment to explain what
it is or does:

        #define VOLATILE_REF 1

-- 

demerphq:
> Sorry, I'm a bit late with a code review. In short, it looks fine, but with the following niggles, listed by commit title. [By fine, I mostly mean the runtime parts - I'm just assuming the regcomp.c parts are ok too because I'm scared of anything in that file.]

See https://github.com/Perl/perl5/pull/20940 which addresses the issues you have raised here.

PR #20919, opened 2023-03-09
Porting/manifest_lib.pl - autofix space/tab issues on `make manisort`

demerphq:
These days we can autofix space issues, so why not?


demerphq:
Ill merge this in the next day or two unless there are objections. (I dont know why there would be, it removes pain from manifest updates.)

PR #20926, opened 2023-03-11
Porting/cmpVERSION.pl - handle missing tags more gracefully

demerphq:
Today Mauke had a surprise where he was on the latest commit but cmp_version.t kept failing. It turned out that his personal github clone was missing the most recent tags, and thus confusing the test.

Since we now release a new tag monthly we can detect that this is happening and skip the test with some helpful diagnositcs on how to fix it.

In my testing this produced the following output (I run make test_porting with TEST_JOBS=16).

```
===(     854;1  0/?  0/?  1/?  0/?  0/?  0/?  0/1   429/6481  0/... )===
# Tag 'v5.31.8' is very old compared to the most recent commit.
# We normally release a new version every month, and this one is 38.20 months
# older than the current commit.  You probably have not synchronized your tags.
# This is common with github clones.  You can try the following steps:
# 
#   git remote add -f upstream git@github.com:Perl/perl5.git
# 
# to fix your checkout.
porting/cmp_version.t ..... skipped: Tag 'v5.31.8' is 38.20 months old. Update your tags!
```

@mauke hopefully this would have helped reduce your confusion and frustration?

### DIFF
--- a/Porting/cmpVERSION.pl
+++ b/Porting/cmpVERSION.pl
@@ -82,6 +82,29 @@ unless ($tag_exists eq $tag_to_compare) {
     exit 0;
 }
 
+my $commit_epoch = `git log -1 --format="%ct"`;
+chomp($commit_epoch);
+my $tag_epoch = `git for-each-ref --format="%(taggerdate:unix)" refs/tags/$tag_to_compare`;
+chomp($tag_epoch);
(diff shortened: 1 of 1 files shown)


demerphq:
Ill merge this tomorrow unless I hear an objection to it.

PR #20930, opened 2023-03-13
mg.c - add support for ${^LAST_SUCCESSFUL_PATTERN}

demerphq:
This exposes the "last successful pattern" as a variable that can be printed, or used in patterns, or tested for definedness, etc. Many regex magical variables relate to PL_curpm, which contains the last successful match.  We never exposed the *pattern* directly, although it was implicitly available via the "empty pattern".  With this patch it is exposed explicitly.  This means that if someone embeds a pattern as a match operator it can then be accessed after the fact much like a qr// variable would be.

@karenetheridge  asked if we had this, and I had to say "no", which was a shame as obviously the code involved isn't very complicated (the docs from this patch are far larger than the code involved!).  At the very least this can be useful for debugging and probably testing. It can also be useful to test if the /is/ a "last successful pattern", by checking if the var is defined.

This one is for you @karenetheridge . Happy hacking. :-).

NOTE: the one question I have about this is whether the name is right. Its a bit of a mouthful, and I have more than once typo'ed it as `${^LAST_SUCCESSFUL_MATCH}` instead of `${^LAST_SUCCESSFUL_PATTERN}`.  I lean away from the `_MATCH` variant, as we already give `${^MATCH}` special meaning (it is the same as $& under /p).

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -2219,8 +2219,9 @@ S_gv_magicalize(pTHX_ GV *gv, HV *stash, const char *name, STRLEN len,
                 if (memEQs(name, len, "\007LOBAL_PHASE"))
                     goto ro_magicalize;
                 break;
-            case '\014':	/* ${^LAST_FH} */
-                if (memEQs(name, len, "\014AST_FH"))
+            case '\014':
+                if ( memEQs(name, len, "\014AST_FH") ||               /* ${^LAST_FH} */
+                     memEQs(name, len, "\014AST_SUCCESSFUL_PATTERN")) /* ${^LAST_SUCCESSFUL_PATTERN} */
                     goto ro_magicalize;
                 break;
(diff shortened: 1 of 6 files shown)


karenetheridge:
How does this differ from `$&`? is it simply the scope?

answer: this is defining the last matched _pattern_, not the string that last matched.
so, we can expect that `defined($&) xor defined(${^LAST_SUCCESSFUL_PATTERN}) == 0`, always

Tux:
@karenetheridge if I understand correctly, `$&` is the matching part in the string matched against, `${^LAST_SUCCESSFUL_PATTERN}` is the regex (part) that lead to `$&`

karenetheridge (approved):
I can't speak to the implementation, but the docs look good and the tests cover a good range of usecases.

thank you very much!

Issue #15938, opened 2017-04-01
${^MATCH} not set without /p (contrary to docs)


p5pRT:
Migrated from [rt.perl.org#131087](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=131087) (status was 'open')

Searchable as RT131087$

p5pRT:
### From @ikegami

#### Created by @ikegami



perlre says&#8203;:

> p
>
> Preserve the string matched such that $\{^PREMATCH\} \, $\{^MATCH\} \, and
> $\{^POSTMATCH\} are available for use after matching\.
>
> In Perl 5\.20 and higher this is ignored\. Due to a new copy\-on\-write
> mechanism\, $\{^PREMATCH\} \, $\{^MATCH\} \, and $\{^POSTMATCH\} will be
> available after the match regardless of the modifier\.

perlvar says&#8203;:

> $\{^MATCH\}
>
> This is similar to $& \($MATCH \) except that it does not incur the
> performance penalty associated with that variable\.
>
> In Perl v5\.18 and earlier\, it is only guaranteed to return a defined
> value when the pattern was compiled or executed with the /p modifier\.
> In Perl v5\.20\, the /p modifier does nothing\, so $\{^MATCH\} does the same
thing as $MATCH \.

test \(without /p\)

&nbsp;    echo 'Lorem ipsum dolor sit ut dicta qui dolores\.' |\\
&nbsp;     perl \-nE 'say "$&|$\{^MATCH\}" while m/dolor/g'

output&#8203;:

&nbsp;    dolor|
&nbsp;    dolor|

test \(with /p\)&#8203;:

&nbsp;    echo 'Lorem ipsum dolor sit ut dicta qui dolores\.' |\\
&nbsp;     perl \-nE 'say $\{^MATCH\} while m/dolor/gp'

output&#8203;:

&nbsp;    dolor|dolor
&nbsp;    dolor|dolor

The code's behaviour differs from both perlre and perlvar\.

\- Eric

In reference to
http&#8203;://stackoverflow\.com/questions/43159304/the\-match\-and\-the\-p\-modifier\-in\-perl\-v24

Sorry if you have received this twice\.

<details><summary>Perl Info</summary>

```

Flags:
    category=core
    severity=low

Site configuration information for perl 5.24.0:

Configured by ikegami at Wed May 25 21:22:43 PDT 2016.

Summary of my perl5 (revision 5 version 24 subversion 0) configuration:

  Platform:
[...]

Issue #20728, opened 2023-01-19
perlvar and perlre incorrectly state that /p regex modifier is a noop


haarg:
perlvar states:

> In Perl v5.18 and earlier, it is only guaranteed to return a defined value when the pattern was compiled or executed with the `/p` modifier. In Perl v5.20, the `/p` modifier does nothing, so `${^MATCH}` does the same thing as `$MATCH`.

> In Perl v5.18 and earlier, it is only guaranteed to return a defined value when the pattern was compiled or executed with the `/p` modifier. In Perl v5.20, the `/p` modifier does nothing, so `${^PREMATCH}` does the same thing as `$PREMATCH`.

> In Perl v5.18 and earlier, it is only guaranteed to return a defined value when the pattern was compiled or executed with the `/p` modifier. In Perl v5.20, the `/p` modifier does nothing, so `${^POSTMATCH}` does the same thing as `$POSTMATCH`.

perlre states:

> As another workaround for this problem, Perl 5.10.0 introduced `${^PREMATCH}`, `${^MATCH}` and `${^POSTMATCH}`, which are equivalent to `` $` ``, `$&` and `$'`, except that they are only guaranteed to be defined after a successful match that was executed with the `/p` (preserve) modifier. The use of these variables incurs no global performance penalty, unlike their punctuation character equivalents, however at the trade-off that you have to tell perl when you want to use them. As of Perl 5.20, these three variables are equivalent to `` $` ``, `$&` and `$'`, and `/p` is ignored. 

These are incorrect. The `/p` flag is not a no-op, and it is requires for the `${^PREMATCH}`, `${^MATCH}`, and `${^POSTMATCH}` variables to be populated.

demerphq:
Ill review the code and fix. Thanks for the reminder.

haarg:
This got tagged as documentation, but we should probably fix the behavior to match the docs (and fix the version numbers in the docs).

demerphq:
On Fri, 20 Jan 2023 at 17:21, Graham Knop ***@***.***> wrote:

> This got tagged as documentation, but we should probably fix the behavior
> to match the docs (and fix the version numbers in the docs).
>

I will follow up on this tomorrow. I am not sure we can fix the behavior. I
have a vague memory that we thought that COW would replace the need for the
/p flag, but that it didnt work out as well as we had hoped. COW has a
refcount limit of 254 or something like that, and I think that in certain
cases we make potentially thousands of COW copies, so it ends up being
[...]

PR #20931, opened 2023-03-13
perlvar.pod - fixup docs for ${^PREMATCH}, ${^MATCH} and ${^POSTMATCH}

demerphq:
At some point someone documented they work just like `` $` ``, `$&` and `$'` but that is not the case, they still require the /p flag or they won't be populated. Until we can sort out the code and be sure that we want to make them exactly the same this patch fixes the docs.


haarg (approved):

haarg:
This fixes #15938 and #20728

PR #20933, opened 2023-03-14
Name unnamed unions

### LINKED ISSUE #20932: regcomp.h uses C11 feature (anonymous union/struct)
Since commit 17e3e02ad120eabda2bdb6c297a70d53294437ef, regcomp.h makes use of anonymous unions and structs, e.g.:
```c
struct regnode_anyofhs { /* Constructed this way to keep the string aligned. */
    U8	str_len;
    U8  type;
    U16 next_off;
    union {
        U32 arg1u;
        I32 arg1i;
        struct {
            U16 arg1a;
            U16 arg1b;
        };
    };
    char string[1];
};
```
This feature is only available in C11 (and previously as a GNU extension): https://en.cppreference.com/w/c/language/union

I don't think perl wants to require a C11 compiler yet, so these anonymous members should get names.

mauke:
Anonymous unions/structs are a C11 feature (previously a GNU extension) and not available in C90 or C99.

Fixes #20932.

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -11927,7 +11927,7 @@ S_optimize_regclass(pTHX_
         &&   *anyof_flags == 0
         &&   start[0] < (1 << ANYOFR_BASE_BITS)
         &&   end[0] - start[0]
-                < ((1U << (sizeof(((struct regnode_1 *)NULL)->arg1u)
+                < ((1U << (sizeof(ARG1u_LOC(NULL))
                                * CHARBITS - ANYOFR_BASE_BITS))))
 
     {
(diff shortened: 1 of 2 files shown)


Tux:
+1 from me

Tux (approved):

Tux (approved):

Tux (approved):

demerphq:
I have renamed "split" to hi_lo, otherwise this is great, thanks. and sorry for the trouble.

demerphq (approved):

PR #20739, opened 2023-01-24
Add IO::Socket::SSL and Net::SSLeay to the Perl core

atrodo:
On p5p, the question was asked what would it look like to just add IO::Socket::SSL and Net::SSLeay to the core. It appears that all that was needed is to include the distribution in cpan/, add information to Porting/Maintainers.pl, and include an explicit dependency between the two packages in Makefile.SH.

The tests all pass for me except for cmp_version.t. I think that's because it is confused by the newly added .c/.h files and thinking the versions need to be bumped instead of recognizing them as new. I am hoping to dig more into what is causing that failure and either fixing the files to pass or find a fix for cmpVERSION.pl.

PR #20936, opened 2023-03-16
Porting/cmpVERSION.pl - Handle new modules as not an error

atrodo:
Checking that a module version was bumped when .c/.h files are updated has an edge case where the previous version was unknown, notably when adding a new module to cpan/. The fix for this is to mark new modules as a new version in those cases

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -732,6 +732,7 @@ John W. Krahn
 John Wright                    <john@johnwright.org>
 Johnny Lam                     <jlam@jgrind.org>
 Jon Eveland                    <jweveland@yahoo.com>
+Jon Gentle                     <atrodo@atrodo.org>
 Jon Gunnip                     <jongunnip@hotmail.com>
 Jon Orwant                     <orwant@oreilly.com>
 Jonathan Biggar                <jon@sems.com>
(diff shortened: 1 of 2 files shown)


demerphq (approved):
LGTM, was there a specific test case that come up that I can use to test this with?

atrodo:
This change came up when I was working on #20739. After I added Net::SSLeay, `t/porting/cmp_version.t` would fail saying that there was no versions bumped but files were modified. The error is not specific to Net::SSLeay, it is for any new package added that has .c/.h files included, even though I suspect that occurrence would be rare.

I tried to find a simple way to add Net::SSLeay to have see the error, but the most reliable instructions appear to be cherry-picking the first commit of #20739, 0a1795aef, and running `Porting/cmpVERSION.pl`. For me, running that displays:

```
$ Porting/cmpVERSION.pl --tap
1..5
#
# Comparing against v5.37.8 ....
#
ok 1 - SKIP Can't parse $VERSION in cpan/Net-SSLeay/inc/Test/Net/SSLeay.pm
ok 2 - SKIP Can't parse $VERSION in cpan/Net-SSLeay/inc/Test/Net/SSLeay/Socket.pm
ok 3 - SKIP Can't parse $VERSION in cpan/Net-SSLeay/lib/Net/SSLeay.pm
ok 4 - SKIP Can't parse $VERSION in cpan/Net-SSLeay/lib/Net/SSLeay/Handle.pm
# No versions bumped in cpan/Net-SSLeay but 1 file was modified
# cpan/Net-SSLeay/constants.c
not ok 5 - in cpan/Net-SSLeay 1 file was modified but no versions were bumped
# not ok 5 - in cpan/Net-SSLeay 1 file was modified but no versions were bumped
```

Adding in this PR, that last error becomes `ok`.

PR #20940, opened 2023-03-18
Various regex engine minor changes for @iabyn

demerphq:
IN https://github.com/Perl/perl5/pull/20918#issuecomment-1474405156 @iabyn asked for some followups.

* Remove vestigal change in Perl_rxres_save()
* Fixes comment  'was bar-foo-bar prior to 5.37.7' to say '5.37.10' ?
* Rework CAPTURE_CLEAR() and UNWIND_PAREN() macros to be static inline.
* In regexp.h - document waht cp and lastcp are used for. 
* Document RE_PESSIMISTIC_PARENS and VOLATILE_REF better

Also includes two other changes I noticed when working on the above:

* Change _pDEPTH and _aDEPTH macros to not use leading underscores. Changed it to comma_pDEPTH and comma_aDEPTH, also added debug_pDEPTH and debug_aDEPTH for completeness (if we ever add a function with no normal arguments that wants to do recursion depth tracking under DEBUGGING).
* Throw an error if a pattern has more than U16_MAX open parens (aka capture groups in it). The code for some time has assumed that the id of all capture groups can fit in an U16, but we weren't validating that assumption during the regexp compile process.  Long term we should revisit all of the logic related to this and either switch it to use U32, or change the things that use U32 members for parent count related logic to use U16 instead.

### DIFF
--- a/dist/Devel-PPPort/parts/embed.fnc
+++ b/dist/Devel-PPPort/parts/embed.fnc
@@ -523,7 +523,7 @@
 :
 :   U  autodoc.pl will not output a usage example
 :
-:   W  Add a _pDEPTH argument to function prototypes, and an _aDEPTH
+:   W  Add a comma_pDEPTH argument to function prototypes, and a comma_aDEPTH
 :      argument to the function calls. This means that under DEBUGGING
 :      a depth argument is added to the functions, which is used for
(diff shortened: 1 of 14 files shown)


iabyn:
LGTM

PR #20941, opened 2023-03-18
t/porting/test_testlist.t - test that we test the same thing with t/harness and t/TEST

demerphq:
Also that we test everything expected in MANIFEST.

Also includes some fixups to t/TEST to deal with the fact that List/Util is not anymore the name of a distribution even though it is the name of an extension. Same for Cwd.

### DIFF
--- a/t/TEST
+++ b/t/TEST
@@ -121,12 +121,9 @@ $| = 1;
 # remove empty elements due to insertion of empty symbols via "''p1'" syntax
 @ARGV = grep($_,@ARGV) if $^O eq 'VMS';
 
-# String eval to avoid loading File::Glob on non-miniperl.
-# (Windows only uses this script for miniperl.)
-@ARGV = eval 'map glob, @ARGV' if $^O eq 'MSWin32';
-
 my $is_win32 = $^O eq "MSWin32";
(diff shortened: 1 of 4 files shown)


demerphq:
@tonycoz, @xenu, @craigberry can you guys check this on your Win32 builds and on VMS please? I suspect this fixes 'make test' on win32. (Thanks to @xenu's feedback).

t/TEST
```diff
@@ -119,35 +119,53 @@ $| = 1;
 #BEGIN { require '../lib/warnings.pm'; "warnings"->import() };
 
 # remove empty elements due to insertion of empty symbols via "''p1'" syntax
```
xenu: This comment should be moved to line 128.

xenu (changes requested):

tonycoz:
> 'make test' on win32

I see the hooks and class tests in CI, it looks fine except for the comment xenu mentioned.

t/TEST
demerphq: Thanks, done.

demerphq:
@xenu, i updated as requested. You ok with this now?

xenu (approved):

PR #20944, opened 2023-03-19
Simplify `use VERSION behaviour` on 5.37 onwards

leonerd:
(this description was all wrong but in any case I've now reverted it to a draft. We'll wait for 5.39 to apply it properly. I'll write something better by then)

### DIFF
--- a/op.c
+++ b/op.c
@@ -7902,7 +7902,7 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
 
         U16 shortver = S_extract_shortver(aTHX_ use_version);
 
-        if (shortver >= SHORTVER(5, 37)) {
+        if (shortver >= SHORTVER(5, 39)) {
             PL_hints |= HINT_STRICT_REFS | HINT_EXPLICIT_STRICT_REFS |
                 HINT_STRICT_SUBS | HINT_EXPLICIT_STRICT_SUBS |
                 HINT_STRICT_VARS | HINT_EXPLICIT_STRICT_VARS;
--- a/op.c
+++ b/op.c
@@ -7902,8 +7902,12 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
 
         U16 shortver = S_extract_shortver(aTHX_ use_version);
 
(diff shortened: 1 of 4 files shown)


Leont:
There are two «if (shortver >= SHORTVER(5, 35))»s in there, can't they be merged?

op.c
```diff
@@ -7899,20 +7899,18 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
 
         U16 shortver = S_extract_shortver(aTHX_ use_version);
 
-        /* If a version >= 5.11.0 is requested, strictures are on by default! */
-        if (shortver >= SHORTVER(5, 11)) {
+        if (shortver >= SHORTVER(5, 37)) {
+            PL_hints |= HINT_STRICT_REFS | HINT_EXPLICIT_STRICT_REFS |
+                HINT_STRICT_SUBS | HINT_EXPLICIT_STRICT_SUBS |
+                HINT_STRICT_VARS | HINT_EXPLICIT_STRICT_VARS;
+        } else if (shortver >= SHORTVER(5, 11)) {
```
tonycoz: The style guide in perlguts says to use uncuddled elses (also below)

tonycoz (approved):

op.c
leonerd: Ohyes, I just copied that from Zefram's patch. Shall fix.

leonerd:
> There are two «if (shortver >= SHORTVER(5, 35))»s in there, can't they be merged?

They could, but that might make the logic harder to read. At the moment, there's a clear split between "fiddling with strict bits", followed by "apply warnings".

leonerd:
Actually this doesn't fix any back-compat. It just adds new behaviour. Probably best to save it for .39 after all.

jkeenan:
@leonerd, this p.r. now has merge conflicts.  Do you want to resolve those -- or would it be better to start fresh with a new p.r.?

leonerd:
Resolved conflicts and repushed; retargeted for 5.39 instead.

jkeenan:
@leonerd, this pull request has been in draft status for 12 months and has merge conflicts for 6 months.  If you intend to move forward with it, could you resolve conflicts, remove the draft status and re-push?  Thanks.

Issue #20945, opened 2023-03-20
Blead breaks MAUKE/HTML-Blitz-0.07


mauke:
Sample report: https://www.cpantesters.org/cpan/report/e1513e54-c6b1-11ed-a7e8-4514ef0028d7

This is due to the following change:

5.36.0:
```
(v5.36.0) $ perl -we 'my $flork; ($flork = undef) =~ s/a/b/'
Use of uninitialized value $flork in substitution (s///) at -e line 1.
```
Blead:
```
(667b5e4) $ ./perl -we 'my $flork; ($flork = undef) =~ s/a/b/'
Use of uninitialized value in substitution (s///) at -e line 1.
```
The variable name is gone from the error message.

A simple bisect points to

> commit 9fdd7fc4796d89d16dceea42f2af91e4fde296ed
> Author: Richard Leach <rich+perl@hyphen-dash-hyphen.info>
> Date:   2022-07-08 14:52:19 +0000
> 
>    Implement OP_PADSV_STORE - combined sassign/padsv OP

as the source of the regression.

jkeenan:
@richardleach can you take a look? Thanks.

mauke:
@jkeenan I think he already did: https://github.com/Perl/perl5/pull/20946#issuecomment-1477120018

PR #20946, opened 2023-03-20
report uninit variable name even in optimized cases

mauke:
Code reviews very much welcome. Most of this patch is based on aping the existing structure of the code, not any deep understanding.

See issue #20945.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -16772,8 +16772,19 @@ S_find_uninit_var(pTHX_ const OP *const obase, const SV *const uninit_sv,
     switch (obase->op_type) {
 
     case OP_UNDEF:
-        /* undef should care if its args are undef - any warnings
+        /* the optimizer rewrites '$x = undef' to 'undef $x' for lexical
+         * variables, which can occur as the source of warnings:
+         *   ($x = undef) =~ s/a/b/;
+         * The OPpUNDEF_KEEP_PV flag indicates that this used to be an
+         * assignment op.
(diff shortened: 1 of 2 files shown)


demerphq (approved):
LGTM, might be good to get @iabyn to have a quick look.

sv.c
```diff
@@ -16870,6 +16887,20 @@ S_find_uninit_var(pTHX_ const OP *const obase, const SV *const uninit_sv,
         }
         return varname(NULL, '$', obase->op_targ,
                        NULL, (I8)obase->op_private, FUV_SUBSCRIPT_ARRAY);
+
+    case OP_AELEMFASTLEX_STORE:
+        if (match) {
+            SV **svp;
```
xenu, line 16893: We're using C99 now, there's no need to declare `svp` at the beginning of the block.

mauke: Hey, tell that to OP_AELEMFAST_LEX. I was only copying the existing code. :-)

Should I move the declarations in both places?

demerphq: Personally I dont think it matters much, that C99 allows us to do otherwise doesn't mean we should always do otherwise. We have some much older style code anyway it might be better to follow the existing style of the function.

richardleach:
mauke: thanks for putting this PR together. LGTM.

iabyn:
LGTM. I assume the second commit will be squashed?

PR #20951, opened 2023-03-21
Quiet hpux compiler warnings from `class.c`

leonerd:
Recent code provokes a few new warnings. This silences them.


demerphq (approved):
Yep, that seems to have silenced the warnings.

PR #20958, opened 2023-03-23
Fixes and changes related to silencing HPUX build warnings

demerphq:
This is a collection of fixes that came up while trying to get things to build properly on HPUX Itanium. We were producing hundreds of warnings of different types, some actually valid warnings and issues, other cruft from HPUX being super picky and not understanding some of the tricks we use. 

In regard to false positive warnings, two notable examples are our OO in C type logic that we do with casting types with similar layouts from one type to another makes the HPUX compiler produce loads of "possible misalignment" warnings.  Since we test thoroughly and Itanium would segfault if we actually did use misaligned pointers these warnings can be silenced. Another case that produces literally hundreds of warnings is the memory wrap logic which produces a warning about a constant comparisons always being false. We hard silence these warning categories.

Another case is that the XS generated code from ParseXS.pm can produce code that is unreachable, so we silence those warnings on hpux as well.

Other examples are legitimate, for instance using && against a constant where we should be using &.

Another case is that the HPUX compiler whines about assignment in a while conditional, so I rewrote the code to not do that.

This includes some changes to the regnode structures used by the regex engine which I did as an initial response to the misalignment problem. I kept these changes as IMO they actually improve the code by making it more clear what parts of the regnode might have multiple meanings, and by ensuring all access to the contents of the regnode srtuctures is via macros which know the correct struct layout. This also means some of the declaration code can be refactored to share a common set of named unions.

There is a bugfix to cmp_version.t to deal with older git's that don't support taggerdate:unix.  

With all of these patch @Tux's HPUX box passes tests under use64bitall. (It fails GDBM tests when built without the use64bitall).

### DIFF
--- a/dist/threads-shared/lib/threads/shared.pm
+++ b/dist/threads-shared/lib/threads/shared.pm
@@ -8,7 +8,7 @@ use Config;
 
 use Scalar::Util qw(reftype refaddr blessed);
 
-our $VERSION = '1.67'; # Please update the pod, too.
+our $VERSION = '1.68'; # Please update the pod, too.
 my $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
@@ -196,7 +196,7 @@ threads::shared - Perl extension for sharing data structures between threads
 
 =head1 VERSION
 
-This document describes threads::shared version 1.67
+This document describes threads::shared version 1.68
 
 =head1 SYNOPSIS
 
--- a/hv.c
+++ b/hv.c
@@ -480,7 +480,7 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
 
     if (!hv)
         return NULL;
-    if (SvTYPE(hv) == (svtype)SVTYPEMASK)
+    if (SvIS_FREED(hv))
         return NULL;
 
     assert(SvTYPE(hv) == SVt_PVHV);
--- a/op.c
+++ b/op.c
@@ -2469,8 +2469,7 @@ Perl_scalarvoid(pTHX_ OP *arg)
         }
         o = next_kid;
     }
-
-    return arg;
+    NOT_REACHED;
 }
 
 
--- a/peep.c
+++ b/peep.c
@@ -2690,7 +2690,7 @@ Perl_rpeep(pTHX_ OP *o)
 {
     OP* oldop = NULL;
     OP* oldoldop = NULL;
-    OP** defer_queue[MAX_DEFERRED]; /* small queue of deferred branches */
+    OP** defer_queue[MAX_DEFERRED] = { NULL }; /* small queue of deferred branches */
     int defer_base = 0;
     int defer_ix = -1;
 
--- a/perl.h
+++ b/perl.h
@@ -11,6 +11,22 @@
 #ifndef H_PERL
 #define H_PERL 1
 
+#if defined(__HP_cc) || defined(__HP_aCC)
+/* The HPUX compiler for Itanium is very picky and warns about
+ * things that gcc doesn't and that we would prefer it does not.
+ * So on that platform silence certain warnings unlaterally. */
+
+/* silence "relational operator ">" always evaluates to 'false'"
+ * warnings. We get a LOT of these from the memwrap checks. */
+#pragma diag_suppress 4276
+
+/* silence "may cause misaligned access" warnings from our "OO in C"
+ * type logic. we do this a lot and if it was broken we would fail tests
+ * all over the place */
+#pragma diag_suppress 4232
+
+#endif /* end HPUX warning disablement */
+
 #ifdef PERL_FOR_X2P
 /*
  * This file is being used for x2p stuff.
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -2989,7 +2989,13 @@ PP(pp_stat)
             Stat_t s;
             CLANG_DIAG_IGNORE_STMT(-Wtautological-compare);
             GCC_DIAG_IGNORE_STMT(-Wtype-limits);
+#if defined(__HP_cc) || defined(__HP_aCC)
+#pragma diag_suppress 2186
+#endif
             neg = PL_statcache.st_ino < 0;
+#if defined(__HP_cc) || defined(__HP_aCC)
+#pragma diag_default 2186
+#endif
             GCC_DIAG_RESTORE_STMT;
             CLANG_DIAG_RESTORE_STMT;
             if (neg) {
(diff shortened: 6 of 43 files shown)


peep.c
```diff
@@ -2690,7 +2690,7 @@ Perl_rpeep(pTHX_ OP *o)
 {
     OP* oldop = NULL;
     OP* oldoldop = NULL;
-    OP** defer_queue[MAX_DEFERRED]; /* small queue of deferred branches */
+    OP** defer_queue[MAX_DEFERRED] = { NULL, NULL, NULL, NULL }; /* small queue of deferred branches */
```
mauke: This could be shortened to
```c
OP** defer_queue[MAX_DEFERRED] = { NULL };
```
This decouples the initializer from the value of `MAX_DEFERRED`.

demerphq: Ack. Thanks.

demerphq: Fixed with the most recent push.

demerphq:
@Tux: with this patch sequence applied all tests pass on HPUX with use64bitall and uselongdouble.

IMO we should merge this sequence ASAP so that HPUX builds clean and passes test on 5.38. 

Reviews appreciated.

demerphq:
Id like to merge this soon, so unless there are objections registered ill merge it in the next day or so.

op.c
```diff
@@ -2469,8 +2469,8 @@ Perl_scalarvoid(pTHX_ OP *arg)
         }
         o = next_kid;
     }
-
-    return arg;
+    /* we should never get here */
```
tonycoz: Given the code at the end of the loop I don't think it's possible for that code to be reached.

Does simply changing this to `NOT_REACHED;` prevent the warning?

pp_sys.c
```diff
@@ -2989,7 +2989,13 @@ PP(pp_stat)
...
             GCC_DIAG_IGNORE_STMT(-Wtype-limits);
+#if defined(__HP_cc) || defined(__HP_aCC)
+#pragma diag_suppress 2186
+#endif
             neg = PL_statcache.st_ino < 0;
+#if defined(__HP_cc) || defined(__HP_aCC)
+#pragma diag_default 2186
+#endif
             GCC_DIAG_RESTORE_STMT;
             CLANG_DIAG_RESTORE_STMT;
```
tonycoz, line 3000: We can avoid all of this - config.h defines `ST_INO_SIGN` which has a negative value if `st_ino` is signed.

Especially since "fsblkcnt_t, fsfilcnt_t, and ino_t shall be defined as unsigned integer types" per POSIX.

hv.c
```diff
@@ -480,7 +480,7 @@ Perl_hv_common(pTHX_ HV *hv, SV *keysv, const char *key, STRLEN klen,
 
     if (!hv)
         return NULL;
-    if (SvTYPE(hv) == (svtype)SVTYPEMASK)
+    if ((U8)SvTYPE(hv) == (U8)SVTYPEMASK)
```
tonycoz: These should all use `SvIS_FREED(sv)`.

(I started typing "Maybe we should have a macro...".)

dist/threads-shared/lib/threads/shared.pm
```diff
@@ -8,7 +8,7 @@ use Config;
 
 use Scalar::Util qw(reftype refaddr blessed);
 
-our $VERSION = '1.67'; # Please update the pod, too.
```
tonycoz, line 11: "prgagma" in commit subject.

perl.h
```diff
@@ -5723,6 +5723,12 @@ EXTCONST char PL_bincompat_options[] =
 #  ifdef DEBUG_LEAKING_SCALARS_FORK_DUMP
                              " DEBUG_LEAKING_SCALARS_FORK_DUMP"
 #  endif
+#  ifdef HAS_LONG_DOUBLE
+                             " HAS_LONG_DOUBLE"
+#  endif
+#  ifdef HAS_STRTOLD
```
tonycoz: I think `HAS_STRTOLD` belongs in `non_bincompat_options[]` in perl.c, I don't see how it has an effect on binary compatibility.

That said, some of the existing `non_bincompat_options` probably belong in `PL_bincompat_options`.

op.c
demerphq: Not sure what you mean by "changing this". I tried adding a NOT_REACHED before the `return arg;` but it still threw warnings. The HPUX compiler doesn't seem to recognize various conventions that gcc does.

pp_sys.c
demerphq: Ok. Ill change it to use that instead. Are you thinking i should remove the DIAG_IGNORE statements as well?  I have to admit i tend to think that it would be better to merge this with the HPUX diag ignores, then afterward do a follow up that removes all of them. What do you think?

hv.c
demerphq: Hmm.  I didnt know we had that one. But the code would be a little different:

```
sv.h:#define SvIS_FREED(sv)     UNLIKELY(((sv)->sv_flags == SVTYPEMASK))
sv.h:#define SvTYPE(sv) ((svtype)((sv)->sv_flags & SVTYPEMASK))
```

Are you confident that would be safe?

dist/threads-shared/lib/threads/shared.pm
demerphq: ack, thanks.

perl.h
demerphq: Ok, ill take a look.

pp_sys.c
demerphq: I have to admit I am a /little/ hesitant about this one.  The comment just above this says:
```
           /*
             * We try to represent st_ino as a native IV or UV where
             * possible, but fall back to a decimal string where
             * necessary.  The code to generate these decimal strings
             * is quite obtuse, because (a) we're portable to non-POSIX
             * platforms where st_ino might be signed; (b) we didn't
             * necessarily detect at Configure time whether st_ino is
             * signed; (c) we're portable to non-POSIX platforms where
             * ino_t isn't defined, so have no name for the type of
             * st_ino; and (d) sprintf() doesn't necessarily support
             * integers as large as st_ino.
             */
```

Which seems to suggest that we don't reliably detect that st_ino is signed. I would rather we don't change this as part of this PR, and do a separate PR where we try to change it.

op.c
tonycoz: Sorry, I meant `NOT_REACHED;` instead of the return.

Though I'm surprised we don't see warnings for the missing return with the change as it is.

pp_sys.c
tonycoz: Ok

op.c
demerphq: dmq: i think i prefer while(1) + NOT_REACHED.
TonyC: I'm fine with that

demerphq:
@tux: HPUX smokes should be ok now. Please let me know if that is not the case.

Issue #20948, opened 2023-03-20
refaliasing disrespects readonlyness of array or hash


leonerd:
(copied from email)

Refaliasing can be used to edit the structure of a readonly array:

```
$ perl5.37.9 -MInternals=SetReadOnly -Mexperimental=refaliasing -lwe '@a = (11,22,33); SetReadOnly(\@a); \$a[1] = \555; print "@a"; splice(@a, 1, 1, 666)'
11 555 33
Modification of a read-only value attempted at -e line 1.
```

Observe that the refaliasing operation is able to replace the scalar at index 1, whereas the splice operation attempting to do the same thing croaks because of the readonlyness. The equivalent also happens with a readonly hash:

```
$ perl5.37.9 -MInternals=SetReadOnly -Mexperimental=refaliasing -lwe '%a = (aa=>11,bb=>22,cc=>33); SetReadOnly(\%a); \$a{bb} = \555; print $a{bb}; delete $a{bb}'
555
Attempt to delete readonly key 'bb' from a restricted hash at -e line 1.
```

These readonly flags ought to be respected. Refaliasing should not provide an ability to bypass them

---

https://www.nntp.perl.org/group/perl.perl5.porters/2023/03/msg266062.html

demerphq:
The hash part is not-a-bug.  Marking a hash as readonly doesn't make its values readonly, they need to be marked readonly independently.

PR #20963, opened 2023-03-23
pp.c/pp_hot.c - prevent aliasing or localizing read only value in a restricted hash

demerphq:
This also teaches t/porting/diag.t to handle hv_notallowed() errors. It does not currently include tests, that can come tomorrow.

$ ./perl -Ilib -Mexperimental=refaliasing,declared_refs -le'my %h=(k=>1); Internals::SvREADONLY(%h,1);
    Internals::SvREADONLY($h{k},1); my $v2=2; local \$h{k} = \$v2'
Attempt to alias readonly key "k" in restricted hash at -e line 1.

$ ./perl -Ilib -Mexperimental=refaliasing,declared_refs -le'my %h=(k=>1); Internals::SvREADONLY(%h,1);
    Internals::SvREADONLY($h{k},1); my $v2=2; local $h{k} = $v2'
Attempt to localize readonly key "k" in restricted hash at -e line 1.

### DIFF
--- a/dist/Storable/t/downgrade.t
+++ b/dist/Storable/t/downgrade.t
@@ -109,7 +109,7 @@ sub test_locked_hash {
         'trying to change a locked key' );
   is ($hash->{$key}, $value, "hash should not change?");
   eval {$hash->{use} = 'perl'};
-  like( $@, "/^Attempt to access disallowed key 'use' in a restricted hash/",
+  like( $@, q{/^Attempt to access disallowed key ["']use["'] in(?: a)? restricted hash/},
         'trying to add another key' );
   ok (eq_array([keys %$hash], \@keys), "Still the same keys?");
 }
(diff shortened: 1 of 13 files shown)


iabyn:
a general comment - that first commit also makes minor cosmetic changes to the warning messages, which generates  large amounts of trivial diffs which swamp out the main purpose of the commit. It would be better to split off into two commits.

demerphq:
> a general comment - that first commit also makes minor cosmetic changes to the warning messages

Yeah, ill split it out.  The diag tests weren't picking up the hv_notallowed() error messages and there were various inconsistencies that were annoying me so I cleaned it up.  I'll split that out.

demerphq:
>It would be better to split off into two commits.

@iabyn i have split them up as requested.

tonycoz:
Is this meant to be fixing #20948?

demerphq:
> Is this meant to be fixing https://github.com/Perl/perl5/issues/20948?

Only the part that concerns restricted hashes.

PR #20972, opened 2023-03-26
Fix tests when -DSILENT_NO_TAINT_SUPPORT is defined

DrHyde:
This PR adds automatic testing with this option enabled, and updates the tests that fail in the `lib/` and `t/` directories. There are also currently some failures in `cpan/` - I have raised PRs for those. You can see in the history for this branch that I fixed them in this repo to make sure the fix _worked_ then reverted when I raised the PRs upstream. Once they've been merged upstream I will edit history to make them disappear from here:

* https://github.com/dankogai/p5-encode/pull/172
* https://github.com/Dual-Life/Scalar-List-Utils/pull/124
* https://github.com/Perl-Toolchain-Gang/Test-Harness/pull/118
* https://github.com/neilb/Text-ParseWords/pull/2

This PR shouldn't be merged - and will fail its tests - until those have been merged upstream and then also updated in perl5-blead.

The fix in the tests is quite straightforward. Checking `$Config{taint_support}` is insufficient for -DSILENT_NO_TAINT_SUPPORT, so it adds checks that look for that in `$Config{ccflags}`.

### DIFF
--- a/lib/locale.t
+++ b/lib/locale.t
@@ -36,7 +36,9 @@ my $is_ebcdic = ord("A") == 193;
 my $os = lc $^O;
 
 # Configure now lets you build a perl that silently ignores taint features
-my $NoTaintSupport = exists($Config{taint_support}) && !$Config{taint_support};
+my $NoTaintSupport =
+    (exists($Config{taint_support}) && !$Config{taint_support}) ||
+    $Config{ccflags} =~ /-DSILENT_NO_TAINT_SUPPORT/;
 
 no warnings 'locale';  # We test even weird locales; and do some scary things
                        # in ok locales
--- a/lib/overload.t
+++ b/lib/overload.t
@@ -1,17 +1,18 @@
 #!./perl -T
 
+use Config;
 BEGIN {
     chdir 't' if -d 't';
     @INC = '../lib';
-    require Config;
-    if (($Config::Config{'extensions'} !~ m!\bList/Util\b!) ){
+    if (($Config{'extensions'} !~ m!\bList/Util\b!) ){
 	print "1..0 # Skip -- Perl configured without List::Util module\n";
 	exit 0;
     }
 }
 
-my $no_taint_support = exists($Config::Config{taint_support})
-                     && !$Config::Config{taint_support};
+my $NoTaintSupport =
+    (exists($Config{taint_support}) && !$Config{taint_support}) ||
+    $Config{ccflags} =~ /-DSILENT_NO_TAINT_SUPPORT/;
 
 my %skip_fetch_count_when_no_taint = (
     '<${$ts}> RT57012_OV' => 1,
@@ -24,7 +25,7 @@ my %skip_fetch_count_when_no_taint = (
 
 sub is_if_taint_supported {
     my ($got, $expected, $name, @mess) = @_;
-    if ($expected && $no_taint_support) {
+    if ($expected && $NoTaintSupport) {
         return skip("your perl was built without taint support");
     }
     else {
@@ -2052,7 +2053,7 @@ foreach my $op (qw(<=> == != < <= > >=)) {
 		    my $exp_fetch = ($var eq '$ts') ?
 			    $exp_fetch_s : $exp_fetch_a;
 		    SKIP: {
-			if ($skip_fetch_count_when_no_taint{$desc} && $no_taint_support) {
+			if ($skip_fetch_count_when_no_taint{$desc} && $NoTaintSupport) {
 			    skip("your perl was built without taint support");
 			}
 			else {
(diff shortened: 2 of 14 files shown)


Leont:
> The fix in the tests is quite straightforward. Checking $Config{taint_support} is insufficient for -DSILENT_NO_TAINT_SUPPORT, so it adds checks that look for that in $Config{ccflags}.

I'm not sure that's the right conclusion. The whole point of the taint_support option was that we don't have to dive into ccflags.

DrHyde:
That's the only trace I could find of the option that's available to perl after it's been built. `taint_support` doesn't _seem_ to get set by anything at all - but I assume I'm missing something.

lib/overload.t
```diff
@@ -1,17 +1,18 @@
...
-    if (($Config::Config{'extensions'} !~ m!\bList/Util\b!) ){
+    if (($Config{'extensions'} !~ m!\bList/Util\b!) ){
 	print "1..0 # Skip -- Perl configured without List::Util module\n";
 	exit 0;
     }
 }
 
-my $no_taint_support = exists($Config::Config{taint_support})
-                     && !$Config::Config{taint_support};
+my $NoTaintSupport =
```
JRaspass, line 13: Why the change of naming style? All the rest of this file is snake case.

DrHyde: For consistency with the `$NoTaintSupport` in other files

Leont:
> That's the only trace I could find of the option that's available to perl after it's been built. taint_support doesn't seem to get set by anything at all - but I assume I'm missing something.

I think that question is somewhat answered by 28003a95818e4d096e18c687a8bddd8584656154

demerphq:
> taint_support doesn't seem to get set by anything at all - but I assume I'm missing something.

@DrHyde, @Leont: Hopefully when that patch gets resurrected, next dev cycle I guess (we/I forgot about it this dev cycle), the var wont be named "taint_support", but rather "taint_disabled". 

It should never have been expressed as a positive config value given the history that we always supported taint, and that we historically had no setting to say so. Thus introducing a new setting that says we do have it now, when we always had it makes code awkward. Having two types of disablement just makes it even worse. 

I believe the thinking was "in the future we will have no taint support at all", but that is an unforced error where the false state of the config value has a different meaning depending on the version of perl (eg it introduces the notion that exists-and-false != not-exists). It also makes the interim logic where there are two forms of disablement more painful, as proved to be the case from the get-go, as part of the reason the patch was not merged was if you did not use Configure's Q&A  to set the define and instead set it via ccflags with something like '-Accflags=-DNO_TAINT_SUPPORT', the taint_support config value would not be set correctly. Much better to keep the setting negtive (as in taint_disabled) and when/if we remove taint entirely we just hardwire that setting to "taint_disabled=yes".  

For this reason I don't think these patches should check "taint_support" at all.  If anything I would check "taint_disabled" and if it is not set then I would check the ccflags. In the future 'taint_disabled' /might/ get set up without updating ccflags, but it is also conceivable it is set up AND the ccflags are setup, and historical perls will not have either "taint_support" nor "taint_disabled" and will only have the ccflags set up, so checking ccflags as a fallback makes sense no matter what.

FWIW, The taint configure patch was released fairly late in the 35.x cycle, and we didn't have a lot of time to find and fix the conceptual or implementation issues in the patch sequence until it was released. FWIW, I had meant to follow up on it in 5.37.x but it slipped my mind until I saw this PR.

lib/locale.t
```diff
@@ -36,7 +36,9 @@ my $is_ebcdic = ord("A") == 193;
 my $os = lc $^O;
 
 # Configure now lets you build a perl that silently ignores taint features
-my $NoTaintSupport = exists($Config{taint_support}) && !$Config{taint_support};
+my $NoTaintSupport =
+    (exists($Config{taint_support}) && !$Config{taint_support}) ||
```
demerphq, line 40: I would lose these checks entirely.  I am very hopeful that we will never have a config variable called "taint_support".

lib/locale.t
```diff
@@ -36,7 +36,9 @@ my $is_ebcdic = ord("A") == 193;
...
+    $Config{ccflags} =~ /-DSILENT_NO_TAINT_SUPPORT/;
```
demerphq, line 41: Why are you only checking SILENT_NO_TAINT_SUPPORT? Shouldn't this be 
```
$Config{ccflags} =~ /-D(?SILENT_)?NO_TAINT_SUPPORT/;
```

demerphq (changes requested):

DrHyde:
> @DrHyde, @Leont: Hopefully when that patch gets resurrected, next dev cycle ...

In that case I'll close this PR for now. @demerphq would you like a reminder after 5.38 is out so it doesn't get forgotten again? :-)

demerphq:
> would you like a reminder after 5.38 is out so it doesn't get forgotten again? :-)

Sure, can't hurt.

yves

PR #20975, opened 2023-03-28
Add a step to the RMG about podcheck

khwilliamson:
It's quick to skim the podcheck summary of errors, and some things, like broken links should really be fixed before the release.

### DIFF
--- a/Porting/release_managers_guide.pod
+++ b/Porting/release_managers_guide.pod
@@ -397,6 +397,21 @@ colon-delimited versions to use for comparison.  For example:
 
 L<http://analysis.cpantesters.org/beforemaintrelease?pair=5.20.2:5.22.0%20RC1>
 
+=head3 check pod errors
+
+F<t/porting/podcheck.t> is a porting test that will fail if it finds new
+problems in pods.  However, it can be taught to ignore problems, and
+sometimes people do so for problems that really should be fixed before
+release.  To see what it is ignoring, run
+
+ perl t/porting/podcheck.t --counts
+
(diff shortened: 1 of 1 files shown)


Porting/release_managers_guide.pod
```diff
@@ -397,6 +397,21 @@ colon-delimited versions to use for comparison.  For example:
...
+release.  To see what it is ignoring, run
+
+ perl t/porting/podcheck.t --counts
+
+Any problems listed as pedantic aren't worth your time investigating.
+These have a C<?> at the beginning of the text, or are for the too-long
+verbatim lines.
+
+But other warnings should be.  In particular, a broken link can well
+mean that someone clicking on the pod in a web page will get a 404.
```
demerphq, line 414: There should be a blank line after this one and before the next =head3

demerphq (changes requested):
I don't entirely get the need for this.  As part of the release process we tell people to run test_porting, which runs these checks. 

I don't object to it either, I am just not sure why it is needed.

Porting/release_managers_guide.pod
khwilliamson: thanks

khwilliamson:
The impetus for this was, IIRC, that 5.36.0 shipped with broken links.  As the text says, people wrongly silence podcheck warnings, and this adds a quick double check that that hasn't happened.

khwilliamson:
I've added to the text more, but can't get github to accept it without a merge commit.  So I'm closing this request, and opening a replacement

PR #20976, opened 2023-03-28
     Implement source::encoding

khwilliamson:
This implements https://github.com/Perl/PPCs/commit/2c5e4698d3aaf9114c631c819323e5b

### DIFF
--- a/lib/source/encoding.pm
+++ b/lib/source/encoding.pm
@@ -1,0 +1,71 @@
+package source::encoding;
+
+use v5.40;
+
+our $VERSION = '0.01';
+
+our $ascii_hint_bits = 0x00000010;
+
+sub import {
+    unimport();     # Get rid of any 'use utf8'
+    my (undef, $arg) = @_;
+    if ($arg eq 'utf8') {
+        require utf8;
+        utf8->import;
+        return;
+    }
+    elsif ($arg eq 'ascii') {
+        $^H |= $ascii_hint_bits;
+        return;
+    }
+
+    die "Bad argument for source::encoding: '$arg'";
+}
+
+sub unimport {
+    $^H &= ~$ascii_hint_bits;
+    utf8->unimport;
+}
+
+1;
+__END__
+
+=head1 NAME
+
+source::encoding -- Declare Perl source code encoding
+
+=head1 SYNOPSIS
+
+ use source::encoding 'ascii';
+ use source::encoding 'utf8';
+ no source::encoding;
+
+=head1 DESCRIPTION
+
+These days, Perl code either generally contains only ASCII characters with
+C<\x{}> and similar escapes to represent non-ASCII, or C<S<use utf8>> is used
+to indicate that the source code itself contains characters encoded as UTF-8.
+
+That means that a character in the source code not meeting these criteria is
+often a typographical error.  This pragma is used to tell Perl to raise an
+error when this happens.
+
+S<C<use source::encoding 'utf8'>> is a synonym for S<C<use utf8>>.  They may
+be used interchangeably.
+
+S<C<use source::encoding 'ascii'>> turns off any UTF-8 expectations, and
+raises a fatal error if any character within its scope in the input source
+code is not ASCII (or ASCII-equivalent on EBCDIC systems).
+
+S<C<no source::encoding>> turns off any UTF-8/ASCII expectations for the
+remainder of its scope.  The meaning of non-ASCII characters is then
+undefined.
+
+S<C<use source::encoding 'ascii'>> is automatically enabled within the
+lexical scope of a S<C<use v5.41.0>> or higher.
+
+=head1 SEE ALSO
+
+L<utf8>
+
+=cut
--- a/toke.c
+++ b/toke.c
@@ -88,7 +88,6 @@ Individual members of C<PL_parser> have their own documentation.
 #  define PL_nexttype		(PL_parser->nexttype)
 #  define PL_nextval		(PL_parser->nextval)
 
-
 #define SvEVALED(sv) \
     (SvTYPE(sv) >= SVt_PVNV \
     && ((XPVIV*)SvANY(sv))->xiv_u.xivu_eval_seen)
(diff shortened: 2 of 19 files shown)


khwilliamson:
If no one reviews this, I'll merge it in a week

toke.c
```diff
@@ -821,6 +823,43 @@ S_cr_textfilter(pTHX_ int idx, SV *sv, int maxlen)
...
 #endif
 
+STATIC void
+S_yyerror_non_ascii_message(pTHX_ const U8 * const s)
+{
+    PERL_ARGS_ASSERT_YYERROR_NON_ASCII_MESSAGE;
+
+    yyerror_pv(Perl_form(aTHX_ "Use of non-ASCII character 0x%02X"
+                               " illegal when 'use source::encoding"
+                               " \"ASCII\"' is in effect", *s), 0);
```
mauke: This message is inconsistent with the POD of `source::encoding`, which spells the import argument `ascii` (in lowercase). I think we should stick to one spelling throughout.

lib/source/encoding.pm
```diff
@@ -0,0 +1,72 @@
+package source::encoding;
+
+use strict;
+use warnings;
```
mauke: Why not `use v5.40;`?

lib/source/encoding.pm
```diff
@@ -0,0 +1,72 @@
...
+
+our $VERSION = '0.01';
+
+our $ascii_hint_bits = 0x00000010;
```
mauke, line 7: Why isn't this a constant?

lib/source/encoding.pm
```diff
@@ -0,0 +1,72 @@
...
+sub import {
+    unimport();     # Get rid of any 'use utf8'
+    my (undef, $arg) = @_;
+    if (lc $arg eq 'utf8') {
+        require utf8;
+        utf8->import;
+        return;
+    }
+    elsif (lc $arg eq 'ascii') {
+        $^H |= $source::encoding::ascii_hint_bits;
```
mauke: `$ascii_hint_bits` is in scope here (declared via `our`). Why use a fully-qualified name?

lib/source/encoding.pm
```diff
@@ -0,0 +1,72 @@
...
+use warnings;
+
+our $VERSION = '0.01';
+
+our $ascii_hint_bits = 0x00000010;
+
+sub import {
+    unimport();     # Get rid of any 'use utf8'
+    my (undef, $arg) = @_;
+    if (lc $arg eq 'utf8') {
```
mauke: Technically speaking, shouldn't case-insensitive comparisons look something like `fc($arg) eq fc('utf8')`?

khwilliamson: Yes.  Do you want this to be changed?

lib/source/encoding.pm
khwilliamson: probably I cargo-culted this code from somewhere else without changing everything/

lib/source/encoding.pm
khwilliamson: I don't see that v5.40 includes warnings.  Where is that documented?

toke.c
khwilliamson: thanks

lib/source/encoding.pm
khwilliamson: It could be; whatever I cargo-culted this from didn't have its corresponding variable be a constant.  I don't know how to make it constant without also using Internals::SvREADONLY

lib/source/encoding.pm
mauke: https://perldoc.perl.org/perl5360delta#use-v5.36

> Furthermore, `use v5.36` will also enable warnings as if you'd written `use warnings`.

https://perldoc.perl.org/functions/use#use-VERSION

> If the specified Perl version is 5.35.0 or higher, [warnings](https://perldoc.perl.org/warnings) are enabled.

lib/source/encoding.pm
mauke: How about `use constant { ASCII_HINT_BITS => 0x00000010 };`?

lib/source/encoding.pm
mauke: TBH I'd prefer if it were case sensitive. :-)

But I'll leave the decision to you.

lib/source/encoding.pm
```diff
@@ -0,0 +1,72 @@
...
+    $^H &= ~$ascii_hint_bits;
+    utf8->unimport;
+}
+
+1;
+__END__
+
+=head1 NAME
+
+source::encoding -- Declare Perl script's encoding
```
jkeenan: Back in the day, we used to get annoyed when people referred to "Perl *scripts*" rather than "Perl *programs*."  I recommend using *program* rather than *script* unless there's a reason to do otherwise in a specific context.

guest20: Is this only in a script? Does it also work in modules?

Grinnz: I think "Perl file" or "Perl source file" would be better descriptive indeed, as this is neither restricted to script files nor applies to the whole program.

guest20: Maybe go with something cute / monastic, like "Tome"?

khwilliamson: I used `source code`, to see how you like that

lib/source/encoding.pm
khwilliamson: changed to require all lc

lib/source/encoding.pm
khwilliamson: changed

lib/source/encoding.pm
khwilliamson: It's been years since I wrote a module; I thought there was a reason the bit variable was a package variable.

lib/source/encoding.pm
guest20: It has a nice symmetry to it

Grinnz: Works for me. Perhaps the description could mention explicitly that the pragma has a lexical effect on the current scope, so there is no ambiguity and no need to mention files.

lib/source/encoding.pm
khwilliamson: I brought this up at the hackathon at the TPRC.  The bottom line is that we couldn't figure out why these things need to be package variables, but they all are.  Changing this one alone would make it the single exception.   So, it's better to leave it as-is.  

A future effort could be to change them all at once.  When I first joined the project I was horrified that we had these magic values that relied on manual synchronization between C constants and perl code.  But it has worked.  It just occurred to me that Configure could be extended to furnish values to both.

So, I will merge this PR as-is by July 3 unless there is protest

PR #20978, opened 2023-03-28
threads.xs: Set non thread context in ithread_set()

khwilliamson:
This generally was getting set anyway, but a test yet to be committed found a case where it wasn't.


khwilliamson:
closing because I don't like merge commits

PR #20984, opened 2023-03-29
Perl RFC have been renamed to PPC

book:
See https://github.com/Perl/PPCs/pull/35


leonerd (approved):
LGTM.
As it's just a docs change this is fine to go in despite the change freeze.

PR #20986, opened 2023-03-29
t/porting/bincompat.t - test the code itself not just the output

demerphq:
Our checks on the define info we expose via Internals::V(), especially the sorted part, did not really work properly as it only checked defines that are actually exposed in our standard builds.  Many of the defines that are exposed in this list are special cases that would not be enabled in a normal build we test under CI, and indeed prior to this patch it was possible for us to produce unsorted output if certain defines were enabled.

This patch adds checks that reads the actual code.  It checks that the define and the string are the same, and it checks that strings would be output in sorted order assuming every define was enabled.

There are two historical exceptions where the string we show and the define use internally are different, but we work around these two cases with as special case hash.

### DIFF
--- a/t/porting/bincompat.t
+++ b/t/porting/bincompat.t
@@ -1,19 +1,58 @@
 #!./perl -w
+BEGIN {
+    chdir ".." if -e "./test.pl";
+    push @INC, "lib";
+}
 use strict;
-require './test.pl';
+require './t/test.pl';
 skip_all("Sorting order differs under EBCDIC") if $::IS_EBCDIC || $::IS_EBCDIC;
 
 use Config;
 
-plan(2);
-
-# Defiantly a white box test...
+my %legacy_different = (
+    # define                       # string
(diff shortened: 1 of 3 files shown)


t/porting/bincompat.t
```diff
@@ -1,19 +1,57 @@
...
-require './test.pl';
+chdir ".." if -e "./test.pl";
+push @INC, "lib";
+require './t/test.pl';
 skip_all("Sorting order differs under EBCDIC") if $::IS_EBCDIC || $::IS_EBCDIC;
 
 use Config;
 
-plan(2);
+plan('no_plan');
```
tonycoz: This is the old style "no plan" testing setup.

It's safer to not use a plan and call `done_testing();` after all of the tests are done.

demerphq: Hrn. I could have sworn that when i recently tried done_testing() in a test.pl based test file it didnt work. But now I see it in there. Sure, I'll use it, the only reason I didn't was because I thought I couldn't.

demerphq: Done. Thanks, TIL. :-)

tonycoz (approved):

PR #20988, opened 2023-03-29
op.c: remove unneeded variable

mauke:
`is_module_install_hack` is a local variable in S_process_special_blocks that is only used to silently treat INIT blocks (in package Module::Install::DSL) as BEGIN blocks. But instead of setting the variable and jumping before the `if` block that checks the variable, we can just jump into the block and get rid of the variable.

### DIFF
--- a/op.c
+++ b/op.c
@@ -11179,12 +11179,11 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
             if (strEQ(name, "INIT")) {
 #ifdef MI_INIT_WORKAROUND_PACK
                 {
-                    HV *hv= CvSTASH(cv);
+                    HV *hv = CvSTASH(cv);
                     STRLEN len = hv ? HvNAMELEN(hv) : 0;
-                    char *pv= (len == sizeof(MI_INIT_WORKAROUND_PACK)-1)
+                    char *pv = (len == sizeof(MI_INIT_WORKAROUND_PACK)-1)
                             ? HvNAME_get(hv) : NULL;
-                    if ( pv && strEQ(pv,MI_INIT_WORKAROUND_PACK) )
-                    {
+                    if ( pv && strEQ(pv, MI_INIT_WORKAROUND_PACK) ) {
                         /* old versions of Module::Install::DSL contain code
                          * that creates an INIT in eval, which expect to run
                          * after an exit(0) in BEGIN. This unfortunately
--- a/op.c
+++ b/op.c
@@ -11037,19 +11037,18 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
 {
     const char *const colon = strrchr(fullname,':');
     const char *const name = colon ? colon + 1 : fullname;
-    int is_module_install_hack = 0;
 
     PERL_ARGS_ASSERT_PROCESS_SPECIAL_BLOCKS;
 
     if (*name == 'B') {
-        module_install_hack:
-        if (strEQ(name, "BEGIN") || is_module_install_hack) {
+        if (strEQ(name, "BEGIN")) {
+            /* can't goto a declaration, but a null statement is fine */
+            module_install_hack: ;
             const I32 oldscope = PL_scopestack_ix;
             SV *max_nest_sv = NULL;
             IV max_nest_iv;
             dSP;
             (void)CvGV(cv);
-            is_module_install_hack = 0;
             if (floor) LEAVE_SCOPE(floor);
             ENTER;
 
@@ -11196,7 +11195,6 @@ S_process_special_blocks(pTHX_ I32 floor, const char *const fullname,
                          */
                         Perl_warn(aTHX_ "Treating %s::INIT block as BEGIN block as workaround",
                                 MI_INIT_WORKAROUND_PACK);
-                        is_module_install_hack = 1;
                         goto module_install_hack;
                     }
 


tonycoz (approved):

demerphq (changes requested):
This includes a bunch of unnecessary whitespace changes that should be removed and applied as a separate commit.  Please dont make whitespace changes just because you don't like the whitespace layout used previously, and if you must, please apply it as a whitespace only commit where you can prove that no functionality changes are mixed in with a diff -w.

mauke:
I have split the commit into two parts.

I didn't make whitespace changes just because I didn't like the previous layout; I added three spaces because there was no consistent layout before. (I added two spaces because the first and third variable declarations in the block randomly had no space between the variable name and the `=`, which was inconsistent with the second declaration (and the rest of the file). The third space I added was after a comma in an argument list, again to make the formatting consistent with other function calls  in the same file.)

demerphq:
> I didn't make whitespace changes just because I didn't like the previous layout; I added three spaces because there was no consistent layout before.

The reasons don't really matter. You altered code for no reason other than to adjust the whitespace, and you did so in a patch that also modified other code. This obscures what has actually been changed in the immediate diff and it obscures the history of the code in blame logs. Generally we frown on making unforced whitespace changes to code in .c files.   (IMO comments, pod, and .h files are a different matter.) As a general rule if you don't need to touch a line of code resist the urge to touch it.

The smaller the change the more important this convention is.  If you tweak a few lines of whitespace in a hundred line patch then no-one is really going to care, but when the percentage of whitespace changes to code changes starts to get into high percentages (eg >5% or so) you should either just not bother, or if you really can't resist (I sympathize) do it as a separate commit that has a message that says something like "whitespace only, use -w to view".  FWIW, there is TONS of code I would happily adjust because it is inconsistent or otherwise annoying, but I resist the urge. You can and should do the same.

demerphq (approved):

mauke:
What's a blame log?

Also, why would people use `-w` to look at a whitespace-only commit? It's just going to look empty.
Also also, if people don't want to see whitespace changes in the history, why don't they just use `git log -pw`?

demerphq:
On Thu, 30 Mar 2023 at 10:40, mauke ***@***.***> wrote:

> What's a blame log?
>

git blame $file

Follow a line of code through the blame log and you'll see why this rule
exists.


> Also, why would people use -w to look at a whitespace-only commit? It's
> just going to look empty.
>

Well, no. If the whitespace change adds or removes lines then it will show
those changes. But the whole point is that using -w to see the patch is
empty is a good way to reassure yourself that the change is in fact a pure
whitespace change.


> Also also, if people don't want to see whitespace changes in the history,
> why don't they just use git log -pw?
>
The point is that when someone looks at a normal diff they dont' want to
see a bunch of whitespace changes at the same time, and they dont want to
*have* to use -w to ignore the changes, as -w is not a universal panacea.
As I said above it does not hide whitespace changes that add or remove
lines, and it can make things look different than they actually are. (I
don't have examples for you, but i have seen it produce very confusing
diffs that only made sense when I removed the -w.)

Anyway, it is a simple rule of thumb: if you don't need to change a line
try to resist the urge to change the line, and if you really must then do
it in a patch by itself. It is a long standing convention, probably a
decade or more, I am actually a bit surprised it isn't documented somewhere
explicitly. I had hoped to be able to point you to the rule in some doc
file but I couldn't find it.  I know it was well established long before we
migrated to git.

Yves


-- 

mauke:
> > What's a blame log?
>
> git blame $file
>
> Follow a line of code through the blame log and you'll see why this rule exists.

`git blame` is mostly useless. At minimum you want `git blame -w` so it skips over whitespace-only changes, but for real work I prefer `git log -p --follow $file`. Not all changes that happen to touch a particular line are meaningful in context and I nearly always want to look further back in history.

> Anyway, it is a simple rule of thumb: if you don't need to change a line try to resist the urge to change the line

I completely disagree, of course: If you're tidying up a block of code (and you should do so when it simplifies the code or makes it more readable), you should also fix formatting issues. Readable code in the working directory right now beats a superficially "nice" history.

> I know it was well established long before we migrated to git.

... which is why I suspect this convention might just be an attempt to work around deficient tooling. Git can track code changes to an arbitrary depth (even as code moves between different files) without major contortions, so why are we optimizing for `git blame`?

demerphq:
On Thu, 30 Mar 2023 at 11:27, mauke ***@***.***> wrote:

> What's a blame log?
>
> git blame $file
>
> Follow a line of code through the blame log and you'll see why this rule
> exists.
>
> git blame is mostly useless.
>

That is a very extreme statement. I use it fairly regularly for code
archaeology. That /you/ find something to be useless is not an indicator
that we all do.


> At minimum you want git blame -w so it skips over whitespace-only
> changes, but for real work I prefer git log -p --follow $file.
>

I am familiar with blame -w.  I am also familiar with log -p --follow.
Familiar enough to believe that the latter is not a replacement for git
blame except in the simplest of cases. I am actually pretty surprised you
even brought it up.


> Not all changes that happen to touch a particular line are meaningful in
> context and I nearly always want to look further back in history.
>
git blame can help you jump past many irrelevant changes in a few simple
checks and is a very powerful tool for code archaeology on files with a lot
of code-churn. If I want to follow the history of a specific piece of
functionality I use git blame not git log. git log would be full of
false-positives.


> Anyway, it is a simple rule of thumb: if you don't need to change a line
> try to resist the urge to change the line
>
> I completely disagree, of course: If you're tidying up a block of code
> (and you should do so when it simplifies the code or makes it more
> readable), you should also fix formatting issues. Readable code in the
> working directory right now beats a superficially "nice" history.
>
It isn't up to you to pick and choose which of our norms apply to you, or
to decide unilaterally that these norms don't apply generally anymore.
Avoiding whitespace-only changes, and isolating them when they do happen,
is a long established convention in the project. All of the old hands try
their best to follow it as much as possible, and so should you.

If you want post to perl5porters and argue that the convention should be
changed then feel free to do so, but debating it with me in the comments on
a now closed PR doesn't achieve anything except to annoy me personally: I
didn't make this convention, and in fact when I was first introduced to it
I felt much the same as you seem to,  but over the years I have come to
appreciate the wisdom of avoiding unnecessary changes, and I now feel it is
a good policy that we should continue to follow. Arguing with me about it
in this context is like arguing with a cop that the speed limit is too low
when they give you a ticket for speeding: the cop didn't make the rule,
they just enforce it. Same thing here.

FWIW, If it were up to me we would have tooling to format all of our code,
and we could and would validate if a patch followed the rules by using the
tool *before* we merged the change. Without such a tool claims about
consistency are IMO unfounded, humans simply are not consistent in how they
perform complex tasks, and readability is a subjective matter. What you
find more readable I might not, and vice-versa. If every one of us changed
the style of code to suit our preferences (even under the guise of
"consistency or readability") then the chance of a mistake creeping in goes
up, and the more difficult it would be to understand our history of
changes. Tools like -w do make it easier, but they don't solve the problem
entirely.

cheers,
Yves



-- 

PR #21008, opened 2023-04-08
.PHONY not portable in dist/Locale-Maketext/Makefile.PL

craigberry:
It breaks the build on VMS because .PHONY is a syntax error in the native make-like utilities.  It probably also breaks nmake builds on Windows.

The solution here is to exclude the entire postamble with the .PHONY directive and the developer convenience targets when the build is running in core.  This may be a slightly bigger hammer than necessary but seems safe and simple enough.

### DIFF
--- a/dist/Locale-Maketext/Makefile.PL
+++ b/dist/Locale-Maketext/Makefile.PL
@@ -28,8 +28,15 @@ WriteMakefile(
 );
 
 sub MY::postamble {
-    return <<'MAKE_FRAG';
-.PHONY: tags critic
+    # .PHONY is not portable
+    my $self = shift;
+    my $phony_line = $self->can('is_make_type')
+                     && ($self->is_make_type('gmake')
+                         || $self->is_make_type('bsdmake'))
+                     ? '.PHONY: tags critic'
+                     : '';
+
+    return "$phony_line\n\n" . <<'MAKE_FRAG';
 
 tags:
 	ctags -f tags --recurse --totals \
--- a/dist/Locale-Maketext/lib/Locale/Maketext.pm
+++ b/dist/Locale-Maketext/lib/Locale/Maketext.pm
@@ -25,7 +25,7 @@ BEGIN {
 }
 
 
-our $VERSION = '1.32';
+our $VERSION = '1.33';
 our @ISA = ();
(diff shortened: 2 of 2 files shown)


Leont:
It's in `dist/` so shouldn't be marked as `CUSTOMIZED` because we are the upstream. You should probably update the version of the dist though.

dist/Locale-Maketext/Makefile.PL
```diff
@@ -28,6 +28,10 @@ WriteMakefile(
 );
 
 sub MY::postamble {
+    # .PHONY is not portable
+    my $self = shift;
+    return '' if $self->{PERL_CORE};
+
```
mauke, line 38: I'd go for something more like:
```perl
my $self = shift;
# .PHONY is not portable
my $phony_line = $self->can('is_make_type') && $self->is_make_type('gmake')
    ? '.PHONY: tags critic'
    : '';
return "$phony_line\n\n" . <<'MAKE_FRAG';
tags:
...
```

craigberry:
> It's in `dist/` so shouldn't be marked as `CUSTOMIZED` because we are the upstream. You should probably update the version of the dist though.

Thanks.  For some reason I thought `CUSTOMIZED` was a bidirectional indicator that "something's different."  I'll fix that shortly.

dist/Locale-Maketext/Makefile.PL
craigberry: I thought about something like that, but how can we be sure only GNU make supports it?  At least [BSD make](https://wiki.netbsd.org/tutorials/bsd_make/) claims to have .PHONY as well.

mauke: For one, I think false negatives are OK here. These extra targets are only relevant for developers (of the module), not people just installing off CPAN.

And we can easily extend the test:
```perl
my $phony_line = $self->can('is_make_type') && ($self->is_make_type('gmake') || $self->is_make_type('bsdmake'))
...
```

craigberry: thanks for the feedback -- I've followed your suggestion

Leont (approved):

PR #21009, opened 2023-04-09
perldelta: the v5.38.0 delta WIP

rjbs:


pod/perl5380delta.pod
```diff
@@ -0,0 +1,1658 @@
...
+C<readline()>, also spelled C<< <> >>, would clear the handle's error
+flag after an error occurred on the stream.
+
+The error flag is now only cleared when an error occurs when reading
+from the child process for glob() in F<miniperl>.  This allows it to
+correctly report errors from the child process on close().
+
+Since the error flag is no longer cleared calling close() on the
+stream may fail and if the stream was not explicitly closed, the
+implicit close of the stream may produce a warning.
```
tonycoz, line 160: One additional behaviour that came out of this in BBCs is that the eof flag is also not being cleared (PerlIO_clearerr() cleared both flags), which is probably worth mentioning.

pod/perl5380delta.pod
```diff
@@ -0,0 +1,1658 @@
...
+=item *
+
+If a thread was created the allocated string would be freed twice.
+
+=item *
+
+If two C<-F> switches were supplied the memory allocated for the first
+switch wouldn't be freed.
+
+=item *
```
tonycoz: There should be a `=back` here, the following `=item`s are unrelated to `PL_splitstr`.

pod/perl5380delta.pod
```diff
@@ -0,0 +1,1658 @@
...
+
+C<make -j6 minitest> could fail due to a build conflict in building
+C<$(MINIPERL_EXE)> between the main make process and a child process.
+[github #19829]
+
+=item *
+
+Properly populate osvers on Dragonfly BSD when the hostname isn't set.
```
tonycoz, line 977: [github #19951]  Do you want ticket numbers for the others that don't have them?

pod/perl5380delta.pod
```diff
@@ -0,0 +1,1658 @@
...
+Fix definition of C<Perl_atof>.
+
+=item *
+
+Fix undefined behavior with overflow related OPTIMIZE_INFTY and delta
+in C<regcomp.c>.
+
+=item *
+
+Fix regnode pointer alignment issue in <regcomp.h>.
```
tonycoz: Presumably C<> around the header name, not just <>, though this and immediately previous mention of regcomp.c should probably be F<> instead of C<>.

pod/perl5380delta.pod
rjbs: I've added a sentence, but am limited in my writing my by understanding.  I wrote:

> The eof flag is also not cleared as often, which may cause disruption in code relying on the old behavior.

pod/perl5380delta.pod
rjbs: Fixed, thanks.

pod/perl5380delta.pod
rjbs: Done, thanks.

pod/perl5380delta.pod
tonycoz: See https://github.com/rjbs/perl5/pull/1

pod/perl5380delta.pod
```diff
@@ -0,0 +1,1663 @@
...
+
+Creating an anonymous sub no longer generates an C<srefgen> op, the
+reference generation is now done in the C<anoncode> or C<anonconst>
+op, saving runtime.  [github #20290]
+
+=item *
+
+Temporary ("mortal") copies are no longer created during context exit for
+internal static SVs that are in no danger of being prematurely freed.
+[L<GH #20800|https://github.com/Perl/perl5/issues/20800>|https://github.com/Perl/perl5/issues/20800]
```
tonycoz, line 262: Assuming https://github.com/Perl/perl5/pull/21064 is applied, this will also need to be removed.

pod/perl5380delta.pod
```diff
@@ -0,0 +1,1663 @@
...
+same as L<C<POSIX::localeconv>|POSIX/localeconv> (returning a hash of
+the C<localeconv()> fields), but directly callable from XS code.
+
+=item *
+
+A new API function L<perlapi/C<my_strftime8>> is added.  This is the
+same as plain L<perlapi/C<my_strftime>>, but with an extra parameter
+that allows the caller to simply and reliably know if the returned
+string is UTF-8.
+
```
tonycoz, line 1331: my_strftime8() was removed in 5f951aa0d5804e69ae58fd3c1619b26a5abb3c13 (or at least made non-API)

pod/perl5380delta.pod
rjbs: That would be great, I will email p5p saying people can push such fixes directly.

Issue #20782, opened 2023-02-09
BBC: Blead Breaks Data::Clone and others


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.37.6.

-----------------------------------------------------------------

BBC: Blead Breaks Data::Clone

Please see http://fast-matrix.cpantesters.org/?dist=Data::Clone

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.37.9:

Configured by cpan at Wed Feb  8 22:16:32 EST 2023.

Summary of my perl5 (revision 5 version 37 subversion 9) configuration:
  Commit id: fe7859256d3a8e646ec90c7313a105c8d273db32
  Platform:
    osname=linux
    osvers=5.15.90-0-lts
    archname=x86_64-linux-thread-multi
    uname='linux cjg-alpine3 5.15.90-0-lts #1-alpine smp wed, 25 jan 2023 08:18:30 +0000 x86_64 linux '
    config_args='-des -Dprefix=~/bin/perl-blead -Dscriptdir=~/bin/perl-blead/bin -Dusedevel -Duse64bitall -Duseithreads'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='12.2.1 20220924'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
[...]

Issue #20863, opened 2023-02-27
PRE-BBC - CPAN distributions which will break when apostrophe as package separator is removed.


demerphq:
Module:
- Sub::Name
- Clone

**Description**
This is a bug to collect the names of CPAN distributions that will be broken when apostrophe as a package separator is removed, but which are not broken directly right now and merely warn. There are also other tickets where people are reporting CPAN distributions that are actually broken by the new deprecations.

**Steps to Reproduce**
Build any of the listed modules with the latest code in blead.

**Expected behavior**
These modules should not warn about use of deprecated features.

**Perl configuration**
Strictly speaking this is not-applicable, any sufficiently recent perl with apostrophe as a package separator deprecations is affected.
```
perl -V
Summary of my perl5 (revision 5 version 37 subversion 10) configuration:
  Commit id: 10c7b2021d05bc5b69caa39f144a06954fa66d68
  Platform:
    osname=linux
    osvers=5.14.0-1057-oem
    archname=x86_64-linux-thread-multi
    uname='linux oncidium 5.14.0-1057-oem #64-ubuntu smp mon jan 23 17:02:19 utc 2023 x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dcc=ccache gcc -Dld=gcc -Dprefix=/home/yorton/perl5/perlbrew/perls/latest_blead -Doptimize=-g -Dusedevel -Dman1dir=none -Dman3dir=none -DDEBUGGING -Dusemultiplicity -Duse64bitall -Dusethreads -Dusedevel -Aeval:scriptdir=/home/yorton/perl5/perlbrew/perls/latest_blead/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-g'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -fwrapv -DDEBUGGING -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
[...]

PR #21011, opened 2023-04-11
undeprecate ' in symbol and make it a default on feature instead

tonycoz:
This PR removes the deprecation of ' as package separator and make it into a default on feature instead, and disabled from 5.37 onwards.

If code tries to use `'` in symbol with the feature disabled then perl throws an error.  I considered leaving the `'` for the next token to pick up but this produced non-obvious messages.


tonycoz:
Attn: @rjbs @leonerd @book

This should fix #20782 #20863

PR #20990, opened 2023-03-30
Support a value stack with over 2**31 entries on 64-bit machines

tonycoz:
This series changes the mark stack and various macros to use SSIze_t instead of I32 for marks.

It also fixes several places where perl uses an I32 as an index into the stack, or an index into a list on the stack, or as a count of items returned on the stack.

It doesn't attempt to fix all I32 problems, but I do have a little list I'll work through of potential problems to check out.

Fixes #20917

PR #21012, opened 2023-04-11
fix a collection of other I32 bugs

tonycoz:
These commits fix a few of the other I32 bugs I suspected when I was reviewing the code for the #20990 work.

I have a few more I haven't investigated yet.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5696,6 +5696,7 @@ t/bigmem/read.t				Check read() handles large offsets
 t/bigmem/regexp.t			Test regular expressions with large strings
 t/bigmem/str.t				Test string primitives with large strings
 t/bigmem/subst.t			Test s/// with large strings
+t/bigmem/subst2.t			Test s//EXPR/e with large strings
 t/bigmem/vec.t				Check vec() handles large offsets
 t/charset_tools.pl			To aid in portable testing across platforms with different character sets
 t/class/class.t				See if class declarations work
--- a/MANIFEST
+++ b/MANIFEST
@@ -5694,6 +5694,7 @@ t/bigmem/index.t			Check that index() handles large offsets
 t/bigmem/pos.t				Check that pos() handles large offsets
 t/bigmem/read.t				Check read() handles large offsets
 t/bigmem/regexp.t			Test regular expressions with large strings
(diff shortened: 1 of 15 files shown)


demerphq (approved):

khwilliamson (approved):

tonycoz:
squashed the "squash me" commit and rebased

jkeenan:
@tonycoz, to properly test this pull request, should we be building a perl with any particular configuration arguments in order to get meaningful results?

(This p.r. has 2 approvals, but probably has not yet been tested other than GH actions.)

tonycoz:
To test this properly you need to set PERL_TEST_MEMORY set in the environment to 8 or more, and have that amount of available memory (in GB).

They only make sense for 64-bit platforms, ie. where ptrsize is 8 (or more if that exists.)

demerphq:
IN the perlreapi.pod commit you say "though I suspect it could use some, some members mentioned no longer exist,", which might be the case as this is  a copy of the actual structure. Ill take a look, but i guess your patch will need to be applied first.

Actually, even better maybe would be to drop the perlreapi.pod patch and let me take care of it.

tonycoz:
> Actually, even better maybe would be to drop the perlreapi.pod patch and let me take care of it.

I've removed it.

I also added another fix for a bug that came up while I was testing the rebased code.  Unfortunately I expect we have a number of similar bugs, I think it means we need to add `-Wconversion` to our warnings and follow-up on them, since it caught this one:
```
op.c: In function ‘Perl_check_hash_fields_and_hekify’:
op.c:2745:64: warning: conversion from ‘ssize_t’ {aka ‘long int’} to ‘I32’ {aka ‘int’} may change value [-Wconversion]
 2745 |             SV *nsv = newSVpvn_share(key, SvUTF8(sv) ? -keylen : keylen, 0);
```

This is a big job:
```
$ make opmini.o 2>&1 | grep 'warning:' | wc -l
247
```
and it takes more than just adding a bunch of casts.

demerphq:
> I've removed it.

Ok, I have just pushed https://github.com/Perl/perl5/pull/21241 as promised.

tonycoz:
Pushed as merge commit 03f6a7924c84dfae6ebc442d73072b6e26ae3896

PR #21017, opened 2023-04-12
perlfunc: add general description of goto

### LINKED ISSUE #21003: [doc] Documentation for 'goto &NAME' should mention that it does not return
**Where**
perlfunc -f goto

**Description**
'goto &NAME' does not return. The documentation should mention this explicitly.

mauke:
Fixes #21003.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -3597,6 +3597,9 @@ X<goto> X<jump> X<jmp>
 
 =for Pod::Functions create spaghetti code
 
+C<goto> transfers execution to a specified point in the program. Unlike a
+function call, execution does not return to C<goto>.
+
 The C<goto LABEL> form finds the statement labeled with LABEL and
 resumes execution there.  It can't be used to get out of a block or
 subroutine given to L<C<sort>|/sort SUBNAME LIST>.  It can be used to go


pod/perlfunc.pod
```diff
@@ -3582,6 +3582,9 @@ X<goto> X<jump> X<jmp>
 
 =for Pod::Functions create spaghetti code
 
+C<goto> transfers execution to a specified point in the program. Unlike a
+subroutine call, it does not return.
```
demerphq: Am i being excessively pedantic when I think that `goto &sub;` does return, just not to where it was goto'ed from? I think this should be 'goto LABEL' only.

jkeenan: @mauke, can you respond to @demerphq's comment above?  Thanks.

mauke: `goto` is a control transfer. Execution flows into `goto`, but does not come out again. That's why I think it is correct to say "goto does not return".
(If you use `goto &sub`, the subroutine does return eventually, of course, but `goto` itself does not.)

mauke: I've reworded it, hopefully making it less ambiguous.

khwilliamson (approved):

Issue #21020, opened 2023-04-14
Unescaped left brace in regex is illegal here in regex


KES777:
**Description**
According to documentation: https://perldoc.perl.org/perlrequick#Matching-repetitions
```
a{n,m} = match at least n times, but not more than m times.
a{n,}  = match at least n or more times
a{,n}  = match n times or fewer
a{n}   = match exactly n times
```
But when I tried to use `a{,n}` syntax, then I got the error above.

**Steps to Reproduce**
```console
$ perl -e 'm/a{1,2}/'
$ perl -e 'm/a{1,}/'
$ perl -e 'm/a{,2}/'
Unescaped left brace in regex is illegal here in regex; marked by <-- HERE in m/a{ <-- HERE ,2}/ at -e line 1.
$ perl -e 'm/a{2}/'
```

This issue occurs on 5.22.1, 5.28.1, 5.30.3, 5.32.0

5.34.0 and 5.37.6 Ok.

Probably this error belongs to: https://perldoc.perl.org/5.26.0/perldelta#Unescaped-literal-%22%7B%22-characters-in-regular-expression-patterns-are-no-longer-permissible

But if so, should we change documentation for `perlrequick` and possible other places.

**Expected behavior**
No error. 

**Perl configuration**
```
Summary of my perl5 (revision 5 version 30 subversion 3) configuration:
   
  Platform:
    osname=linux
    osvers=5.4.0-37-generic
    archname=x86_64-linux
    uname='linux work 5.4.0-37-generic #41~18.04.1-ubuntu smp mon jun 8 13:37:29 utc 2020 x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dprefix=/home/kes/perl5/perlbrew/perls/perl-5.30.3 -Aeval:scriptdir=/home/kes/perl5/perlbrew/perls/perl-5.30.3/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
    bincompat5005=undef
  Compiler:
    cc='cc'
    ccflags ='-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
[...]

PR #21021, opened 2023-04-14
perlrequick: Note {,n} added in 5.34

khwilliamson:
This fixes GH #21020

### DIFF
--- a/pod/perlrequick.pod
+++ b/pod/perlrequick.pod
@@ -363,7 +363,7 @@ C<a{n,}> = match at least C<n> or more times
 
 =item *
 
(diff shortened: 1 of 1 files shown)


ilmari:
The page doesn't document when any other feature was added (such as `/r`, which is uses liberally). Why is this one special?

Grinnz:
Currently because it shows up in the docs you find at perldoc.perl.org and is not available in most Perl versions in use. But we note when features are added in many places and I think it is always helpful and at worst not much clutter.

tonycoz (approved):

jkeenan (approved):

PR #21023, opened 2023-04-15
Unbreak Perl build on legacy Darwin systems

### LINKED ISSUE #18902: Perl 5.34.0 build breaks on Darwin gcc4
The Perl 5.6.34.0 distribution build employs 

-Wno-error:implicit-function-declaration

without configure checking whether -Wno-error: is supported.

This breaks building Perl on Darwin 8 / Mac OS X 10.4, whose native toolchain (Xcode) sports gcc v4.

The pkgsrc build can work around this by conditionallly stripping the option, but shouldn't have to.

sevan:
On modern macOS the compiler defaults to '-Werror,-Wimplicit-function-declaration' This breaks the build on code which calls a function without a prototype (invalid in C99).
Turning the check off with '-Wno-error=implicit-function-declaration' breaks the build of Perl itself on legacy versions of Darwin with compilers which do not recognise the flags.
Removing it allows Perl to once again build on Mac OS X as old as 10.4 with GCC 4.0.1.

Resolves #18902

### DIFF
--- a/dist/ExtUtils-CBuilder/Changes
+++ b/dist/ExtUtils-CBuilder/Changes
@@ -1,5 +1,17 @@
 Revision history for Perl extension ExtUtils::CBuilder.
 
+0.280243 - 2026-02-23
+
+  Fix:
+  - Unbreak Perl build on legacy Darwin systems.
+    On modern macOS the compiler defaults to
+    '-Werror,-Wimplicit-function-declaration'. This breaks the build on code
+    which calls a function without a prototype (invalid in C99).
+    Turning the check off with '-Wno-error=implicit-function-declaration' breaks
+    the build of Perl itself on legacy versions of Darwin with compilers which do
+    not recognise the flags.
+    https://github.com/Perl/perl5/issues/18902
+
 0.280242
 
  Fix:
--- a/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder.pm
+++ b/dist/ExtUtils-CBuilder/lib/ExtUtils/CBuilder.pm
@@ -7,7 +7,7 @@ use Perl::OSType qw/os_type/;
 
 use warnings;
 use strict;
-our $VERSION = '0.280242'; # VERSION
+our $VERSION = '0.280243'; # VERSION
 our @ISA;
 
 # We only use this once - don't waste a symbol table entry on it.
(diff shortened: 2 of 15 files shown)


Leont:
I see you've already made a PR for the MakeMaker part of change. This PR should be marked as not mergeable until that has been applied, released and synced back into blead.

Otherwise I approve of this change.

sevan:
Moved PR into draft mode until other PR for MakeMaker is merged.

haarg:
The changes to EUMM should be removed from this PR, since that will be handled separately.

Making this change, along with the EUMM change, is likely to break a number of XS modules on macOS. I agree that we should change this, but I'm not sure about the timing.

sevan:
I've removed `cpan/ExtUtils-MakeMaker/lib/ExtUtils/MM_Darwin.pm`

sevan:
[Smoke test result with unpatched blead](https://perl5.test-smoke.org/report/5033206)

sevan:
[Smoke test result with patched blead](https://perl5.test-smoke.org/file/log_file/5033391)

Ignore the failures, I didn't bump the rev so t/porting/cmp_version.t failed, cpan/ExtUtils-MakeMaker/t/04-xs-rpath-darwin.t  fails but the test shouldn't be running on Darwin 8.
There is an issue with cpan/Test-Simple/t/Test2/behavior/ipc_wait_timeout.t hanging which is why the smoke test took so long.  I had to run 'kill -9' several times over the past few days.

jkeenan:
> [Smoke test result with patched blead](https://perl5.test-smoke.org/file/log_file/5033391)
> 
> Ignore the failures, I didn't bump the rev so t/porting/cmp_version.t failed, cpan/ExtUtils-MakeMaker/t/04-xs-rpath-darwin.t fails but the test shouldn't be running on Darwin 8. There is an issue with cpan/Test-Simple/t/Test2/behavior/ipc_wait_timeout.t hanging which is why the smoke test took so long. I had to run 'kill -9' several times over the past few days.

There has been no discussion in this ticket in more than a year, and in that time the code has acquired several merge conflicts.  Those will have to be resolved before we can proceed.

sevan:
I'll take a look at updating over the weekend.

khwilliamson:
@sevan any chance of moving forward on this

sevan:
@khwilliamson working on resolving the conflict atm, hopefully I wont butcher it. :)

haarg (approved):

PR #21029, opened 2023-04-16
t/harness: avoid brace expansion for -dumptests

craigberry:
Brace expansion is not available in a POSIX shell, is handled slightly differently by the shells that do support it, and is unlikely to work when the underlying implementation for Perl's glob() function is not a Unix shell.  So instead of doing:

    {foo,bar,baz}/*.t

just accumulate the results of simpler glob operations:

    foo/*.t
    bar/*.t
    baz/*.t

This also allows us to dispense with the recursive function _extract_tests() and its fancy dispatch based on reference type; we would only ever be calling it with a simple string argument, so we might as well just call glob() directly.

### DIFF
--- a/t/harness
+++ b/t/harness
@@ -116,28 +116,6 @@ if ($ENV{HARNESS_OPTIONS}) {
 
 $jobs ||= 1;
 
-sub _extract_tests;
-sub _extract_tests {
-    # This can probably be done more tersely with a map, but I doubt that it
-    # would be as clear
-    my @results;
-    foreach (@_) {
-        my $ref = ref $_;
-        if ($ref) {
-            if ($ref eq 'ARRAY') {
-                push @results, _extract_tests @$_;
-            } elsif ($ref eq 'HASH') {
-                push @results, _extract_tests values %$_;
-            } else {
-                die "Unknown reference type $ref";
-            }
-        } else {
-            push @results, glob $_;
-        }
-    }
-    @results;
-}
-
 my %total_time;
 sub _compute_tests_and_ordering($) {
     my @tests = $_[0]->@*;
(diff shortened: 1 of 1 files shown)


demerphq (approved):

demerphq:
@craigberry I think this applies to t/harness in general, not just for -dumptests, although it might be the case that you only see it on your target platform via -dumptests (do you use `make test` or `make test_harness` normally?)

craigberry:
> @craigberry I think this applies to t/harness in general, not just for -dumptests, although it might be the case that you only see it on your target platform via -dumptests (do you use `make test` or `make test_harness` normally?)

Ooh, good point.  I noticed this running the porting test and hadn't actually tried to run the full test suite with harness.  I will revise the commit message and do a full run of `make test_harness` before merging (not sure if the CI has already done that).

demerphq:
> and do a full run of `make test_harness` before merging

I am not sure about the VMS process, but on *nix, test_porting is an alias for test_harness with some arguments. The porting tests however include the new tests I added recently which use -dumptests to ensure that `make test` and `make test_harness` run the same test files as each other.  So assuming VMS works like *nix, if you ran `make test_porting` then you already tested at both levels. :-)

t/harness
```diff
@@ -366,15 +344,21 @@ if (@ARGV) {
...
             push @seq, _compute_tests_and_ordering(\@next)->@*;
         }
 
-        my $last = { par => '{' . join (',', @last) . '}/*.t' };
-        @last = _extract_tests ($last);
+        @extracted = ();
+        foreach my $last (@last) {
+            push @extracted, glob ("$last/*.t");
+        }
+        @last = @extracted;
```
demerphq: doesnt this end up as: `@last = map { glob("$_/*.t") } @last;` (same above with `@next`?)

craigberry: Yes, thanks.  There seem to be other places where foreach is preferred in this file, so I thought I was going with the flow, but there are a number of other usages of map as well.  So I've pushed a revision that just uses map, which does look simpler overall.

Issue #20999, opened 2023-03-31
Warning: Cast truncates Constant Value


mubinskt:
Module:

**Description**

The inline.h file reports warning C4310 
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4310?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(C4310)%26rd%3Dtrue&view=msvc-170

![image](https://user-images.githubusercontent.com/48523534/229121937-2e76feac-cd0c-4252-94f0-57aa521d53ac.png)

**Steps to Reproduce**

Rebuild in a MFC Windows application referring perl files.

**Expected behavior**

**Perl configuration**

```

This is perl 5, version 32, subversion 1 (v5.32.1) built for MSWin32-x64-multi-thread

Copyright 1987-2021, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at http://www.perl.org/, the Perl Home Page.
```

demerphq:
As far as I can tell this is a bug in the MSVC, which is getting confused. There is a fairly long set of defines involved here:

```
#define NATIVE_BYTE_IS_INVARIANT(c)     UVCHR_IS_INVARIANT(c)
#define UVCHR_IS_INVARIANT(cp)  (OFFUNI_IS_INVARIANT(NATIVE_TO_UNI(cp)))
#define NATIVE_TO_UNI(ch)        ((UV) ASSERT_NOT_PTR(ch))
```

I believe the above define is where the trouble starts, the U8 argument is cast to a UV.

```
#define OFFUNI_IS_INVARIANT(c)                                              \
                        (((WIDEST_UTYPE)(c)) < UTF_MIN_CONTINUATION_BYTE)
```

So the UV is then cast to WIDEST_UTYPE, which might be smaller than a UV. It is defined as:

```
#define WIDEST_UTYPE PERL_UINTMAX_T
```

I assume that you are building a 32 bit perl, and that this:

```
    typedef U32TYPE PERL_UINTMAX_T;
```

is how PERL_UINTMAX_T is defined. (I make this assumption because if it were a 64 bit perl PERL_UINTMAX_T would be U64 which is what UV should be defined to be.)

[...]

PR #21031, opened 2023-04-17
eliminate or suppress various header warnings under MSVC with /W4

tonycoz:
As illustrated by #20999 it's common for projects to be built with the maximum warnings, and it's a common recommendation.

It's also something we follow with gcc, g++ and clang on POSIX-ish systems.

This change allows embedders of perl using MSVC to enable /W4 and include the perl headers, eliminating most warnings.

With these changes I get no warnings from the headers building `..\toke.obj` with `/W4`, which is at least a good start.

I haven't tried to eliminate all warnings in the perl build, that would be a much larger change.

https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md#compilers
https://www.learncpp.com/cpp-tutorial/configuring-your-compiler-warning-and-error-levels/
https://www.reddit.com/r/programming/comments/1m263m/turn_on_your_damn_warnings/

### DIFF
--- a/win32/win32.h
+++ b/win32/win32.h
@@ -274,7 +274,7 @@ typedef struct
     {
         FILE  _public_file;
         char* _ptr;
-    };
+    } u;
 
     char*            _base;
     int              _cnt;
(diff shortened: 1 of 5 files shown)


win32/win32.h
```diff
@@ -274,7 +274,7 @@ typedef struct
     {
         FILE  _public_file;
         char* _ptr;
-    };
+    } _u;
```
khwilliamson: This is not the commit to fix this, but it is undefined behavior for any symbol to begin with an underscore.  So, in theory all the similar symbols should be changed to like `u_`, which likely will never happen.

khwilliamson (approved):
The reason I can think of to not remove unused context parameters is that it affects code that uses the Perl_ form to call functions, such as when -DPERL_NO_SHORT_NAMES
 is chosen.  I don't know if anybody actually does that

win32/win32.h
tonycoz: "All identifiers that begin with an underscore are always reserved for use as identifiers
with file scope in both the ordinary and tag name spaces."

This means they're reserved a names at file scope (variables, functions, typedefs at file scope) and tag names, like `struct this`, `union that` or `enum andthis`.

They aren't reserved universally like names starting with underscore followed by an uppercase letter or starting with two underscores. (Or any name containing two underscores in a row in C++.)

But I've changed it for the new name anyway.

tonycoz:
> The reason I can think of to not remove unused context parameters is that it affects code that uses the Perl_ form to call functions, such as when -DPERL_NO_SHORT_NAMES is chosen. I don't know if anybody actually does that

I wasn't sure myself which was best here, I've changed them to `PERL_UNUSED_CONTEXT;` rather than removing the context parameter.

tonycoz:
Applied manually.

PR #21032, opened 2023-04-17
Correct the example in the documentation of splitdir in File::Spec::Win32

### LINKED ISSUE #20758: [doc] A wrong example in the splitdir method in the File::Spec::Win32 module
**Where**

 The splitdir method in the File::Spec::Win32 module

https://github.com/Perl/perl5/blob/14733d8a61afacead361cbb71efb4046e6dbe16a/dist/PathTools/lib/File/Spec/Win32.pm#L252

**Description**

`"/a/b/c"` is maybe wrong. `"/a/b//c/"` is maybe correct. 

```
Unlike just splitting the directories on the separator, leading empty and 
trailing directory entries can be returned, because these are significant
on some OSs. So,

    File::Spec->splitdir( "/a/b/c" );

Yields:

    ( '', 'a', 'b', '', 'c', '' )
```

tonycoz:
Fixes #20758

### DIFF
--- a/dist/PathTools/lib/File/Spec/Win32.pm
+++ b/dist/PathTools/lib/File/Spec/Win32.pm
@@ -5,7 +5,7 @@ use strict;
 use Cwd ();
 require File::Spec::Unix;
 
-our $VERSION = '3.88';
+our $VERSION = '3.90';
 $VERSION =~ tr/_//d;
 
 our @ISA = qw(File::Spec::Unix);
--- a/dist/PathTools/lib/File/Spec/Win32.pm
+++ b/dist/PathTools/lib/File/Spec/Win32.pm
@@ -249,7 +249,7 @@ Unlike just splitting the directories on the separator, leading empty and
 trailing directory entries can be returned, because these are significant
 on some OSs. So,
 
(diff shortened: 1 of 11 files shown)


dist/PathTools/lib/File/Spec/Win32.pm
```diff
@@ -249,7 +249,7 @@ Unlike just splitting the directories on the separator, leading empty and
 trailing directory entries can be returned, because these are significant
 on some OSs. So,
 
-    File::Spec->splitdir( "/a/b/c" );
+    File::Spec->splitdir( "/a/b//c" );
```
mauke: That needs to be
```perl
File::Spec->splitdir( "/a/b//c/" );
```
(with a trailing slash) to match the output given.

PS: That whole method body could be replaced by `return split( m|[\\/]|, $directories, -1 );`

PPS: I'm not sure what the `\Z(?!\n)` dance is for. Compatibility with perl-5.004, which didn't have `\z` maybe?

tonycoz: Oops, thank you and fixed.

xenu (approved):

PR #21046, opened 2023-04-22
Add missing ccflags on VMS to dist/Time-HiRes/Makefile.PL

craigberry:
The absence of these flags has never previously caused a problem for running the probes, but on OpenVMS x86, warnings about long symbols cause the probes to fail, probably because those symbols are not getting optimized out as early in the new compiler. This in turn causes the build to fail due to defining a homegrown clockid_t which conflicts with the system version. So add the flags that specify symbol handling (as well as everything else being used to build Perl).


leonerd (approved):
LGTM

demerphq (approved):

Leont (approved):

khwilliamson (approved):

PR #21047, opened 2023-04-23
configpm/INSTALL - hide POD content in configpm, add docs on NO_TAINT_SUPPORT to INSTALL

### LINKED ISSUE #21045: configpm: POD formatting problems and more
The documentation for `configpm` has a number of POD formatting problems.

First, if I run it through the `podchecker` utility shipped with `perl`, I get this output:
```
$ podchecker configpm
*** ERROR: '=item' outside of any '=over' at line 1108 in file configpm
*** ERROR: You forgot a '=back' before '=head1' at line 1145 in file configpm
*** ERROR: unresolved internal link 'taint_disabled' at line 1112 in file configpm
configpm has 3 pod syntax errors.
```
Similarly, if I run this file through the `pod2text` utility, I get this output:
```
$ pod2text < configpm > /tmp/configpm.txt
IO::File=IO(0x5563352d28f0) around line 1108: '=item' outside of any '=over'
IO::File=IO(0x5563352d28f0) around line 1145: You forgot a '=back' before '=head1'
POD document had syntax errors at /home/jkeenan/perl5/perlbrew/perls/perl-5.36.0/bin/pod2text line 75.
```
Second, it appears that several `heredoc` markers "leak" into the generated output.  From the plain-text file generated above:
```
    "taint_support"
        From define: "SILENT_NO_TAINT_SUPPORT" or "NO_TAINT_SUPPORT"

        If this perl is compiled with support for taint mode this variable
        will be set to 'define', if it is not it will be set to the empty
        string. Either of the above defines will result in it being empty.
        This property was added in version 5.37.11. See also
        "taint_disabled".

        EOF_TEXT

        $glossary{t}{taint_disabled} //= <<EOF_TEXT; =item "taint_disabled"

```
... and ...
```
        being a true value. This property was added in 5.37.11. See also
        "taint_support".

        EOF_TEXT

        if ($Opts{glossary}) { foreach my $fc (sort keys %glossary) { print
        CONFIG_POD "=head2 $fc\n\n=over 4\n\n"; foreach my $item (sort keys
        %{$glossary{$fc}}) { print CONFIG_POD $glossary{$fc}{$item}; } print
        CONFIG_POD "=back\n\n"; } }

        print CONFIG_POD <<'ENDOFTAIL';
```
Since this is newly added documentation, I'm not exactly sure what the intended output should be.

Third, the POD formatting errors are not being detected by any of the tests run via `make test_porting`.

Fourth (and this is somewhat separate from the foregoing), I only stumbled on the formatting problems because I was searching for documentation on `taint_disabled` (and doing that only because of a pull request filed against a CPAN distribution for which I am caretaker).  I would have expected to find such documentation in some file underneath `pod/`, but the only place I could find it was in the generated file `pod/perltoc.pod`.
```
$ ack -l 'taint_(support|disabled)' ./pod
pod/perltoc.pod
```
Even if `configpm` had its POD correctly formatted, I doubt that I would have thought to look there for documentation on these two properties.

demerphq:
At least some of the POD in configpm relates to NO_TAINT_SUPPORT. Hide the POD in comments, and make the file easier to read as the code that it outputs is visually distinct from the code it executes. THe modification to INSTALL just documents the NO_TAINT_SUPPORT defines in INSTALL.

Fixes https://github.com/Perl/perl5/issues/21045

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -420,6 +420,38 @@ the compiler dependency, you may need to also add "-Dcc=...".
 At C level the type is called C<__float128> (note, not "long double"),
 but Perl source knows it as NV.  (This is not "long doubles".)
 
+=head3 Taint Support
+
+Traditional perl has provided a security mechanism based on marking
+input data as untrusted unless it has been validated by a regex. This
+mechanism is called tainting and is enabled with the -T or -t options on
+the command line. This support has a performance cost on all code
+executed. It is possible to disable this support by providing the
+setting C<-Accflags=-DNO_TAINT_SUPPORT> or the setting
+C<-Accflags=-DSILENT_NO_TAINT_SUPPORT> to Configure.
+
+The former option C<NO_TAINT_SUPPORT> is more secure and disables taint
+support while making the use of the C<-T> or C<-t> options which
+normally enable taint support into an untrappable exception to ensure
+that no-one uses them while expecting taint checks to run. With this
+build mode there can be no confusion if a script supports taint or not.
+
+The latter option, C<SILENT_NO_TAINT_SUPPORT> is less secure but more
+flexible in that it silently disables taint support while ignoring the
+C<-T> and C<-t> command line options entirely. In this build mode it is
+possible to run a script with the C<-T> or C<-t> option and expect taint
+support but not get it, so this mode is only recommended to people who
+really know what they are doing.
+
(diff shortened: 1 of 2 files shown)


demerphq:
I'll merge this tomorrow unless there are objections.

PR #21053, opened 2023-04-28
 op/sub.t: Cancel watchdog when done

khwilliamson:
This test was flapping on some boxes
      (sample: http://perl.develop-help.com/raw/?id=292891 )

### DIFF
--- a/t/op/sub.t
+++ b/t/op/sub.t
@@ -426,6 +426,7 @@ eval '
    CORE::state sub b; sub d { sub b {} sub d }
  ';
 eval '()=%e; sub e { sub e; eval q|$x| } e;';
+watchdog 0;
 
 fresh_perl_like(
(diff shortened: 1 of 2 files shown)


t/op/sub.t
```diff
@@ -426,6 +426,7 @@ eval '
    CORE::state sub b; sub d { sub b {} sub d }
```
tonycoz, line 426: The commit message should refer to https://www.nntp.perl.org/group/perl.daily-build.reports/2023/04/msg292891.html  (linked from the top of the page you linked) rather than perl.develop-help.com, while I don't plan to remove the site or change the URL structure, I might get hit by a bus, or run out of money or whatever.

tonycoz (approved):
The url thing is minor.

Issue #20373, opened 2022-10-08
VS2022 Test op/magic.t, uni/gv.t, ext/POSIX/t/time.t failure in perl 5.37.4 on Windows 8.1


twata1:
**Description**

Hello,

I'am building Perl 5.37.4 on Windows 8.1 64 bit using Visual Studio 2022 Community Edition (Microsoft(R) C/C++ Optimizing Compiler Version 19.33.31630 for x64).

The compilation itself succeeds, although there are some warnings, such as C4244 and C4267.
Several tests fail when running nmake test.

```
C:\VCPerl64\perl-5.37.4\win32>nmake

(snip)

        if exist ..\..\lib\auto\threads\shared\shared.dll.manifest mt -nologo -manifest ..\..\lib\auto\threads\shared\shared.dll.manifest -outputresource:..\..\lib\auto\threads\shared\shared.dll;2
        if exist ..\..\lib\auto\threads\shared\shared.dll.manifest del ..\..\lib\auto\threads\shared\shared.dll.manifest        "C:\VCPerl64\perl-5.37.4\miniperl.exe" "-I..\..\lib" -MExtUtils::Command -e chmod -- 755 ..\..\lib\auto\threads\shared\shared.dll
Everything is up to date. 'nmake test' to run test suite.

C:\VCPerl64\perl-5.37.4\win32>nmake test

(snip)

Test Summary Report
-------------------
op/magic.t                                                         (Wstat: 0 Tests: 208 Failed: 4)
  Failed tests:  190, 193, 197, 199
uni/gv.t                                                           (Wstat: 0 Tests: 206 Failed: 1)
  Failed test:  78
../dist/Net-Ping/t/450_service.t                                   (Wstat: 0 Tests: 26 Failed: 0)
  TODO passed:   9
../ext/IPC-Open3/t/IPC-Open3.t                                     (Wstat: 0 Tests: 45 Failed: 0)
  TODO passed:   25
../ext/POSIX/t/time.t                                              (Wstat: 512 (exited 2) Tests: 19 Failed: 2)
  Failed tests:  9-10
  Non-zero exit status: 2
Files=2730, Tests=1119663, 4274 wallclock secs (94.50 usr + 13.05 sys = 107.55 CPU)
Result: FAIL
NMAKE : fatal error U1077: '.\perl.exe' : return code '0x7'
Stop.

C:\VCPerl64\perl-5.37.4\win32>cd ..\t

[...]

PR #21063, opened 2023-05-01
Remove my_strftime8()

khwilliamson:
The comment from Tony Cook

https://github.com/Perl/perl5/issues/20373#issuecomment-1524256091

made me realize that this function doesn't fully work.  It was added as public API earlier in the 5.37 series, but we don't want it making it into a stable release.  This commit renames it so that the original name will no longer work, but POSIX.xs can still, by changing to use the new name, which is undocumented

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -2076,7 +2076,8 @@ Adfp	|char * |my_strftime	|NN const char *fmt			\
 				|int wday				\
 				|int yday				\
 				|int isdst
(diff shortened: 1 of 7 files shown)


tonycoz (approved):

tonycoz:
Marked this a blocker, I think this is needed before 5.38.0 and we're in a full code freeze according to the schedule.

Issue #21040, opened 2023-04-20
Blead (9e254b0) Broke on Alpine Linux under g++ with Threads Enabled


cjg-cguevara:
Blead (9e254b0) broke on Alpine Linux under g++ with threads enabled.

Example:
https://perl5.test-smoke.org/report/5033265
https://perl5.test-smoke.org/file/log_file/5033265

Appears to have started at 9e254b0.

Note that Alpine Linux uses musl instead of glibc:
```
$ ldd --version
musl libc (x86_64)
Version 1.2.3
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
```

jkeenan:
@cjg-cguevara, can you paste the tail of the output from `make` on one or more of these threaded builds?

@khwilliamson, can you take a look?

cjg-cguevara:
```
g++ -c -DPERL_CORE -D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wwrite-strings -Wno-use-after-free caretx.c
g++ -c -DPERL_CORE -D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wwrite-strings -Wno-use-after-free dquote.c
g++ -c -DPERL_CORE -D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wwrite-strings -Wno-use-after-free time64.c
[...]

PR #21072, opened 2023-05-03
switch_locale_context: Add aTHX

khwilliamson:
This fixes GH #21040

Instead of a dTHX, this passes aTHX automatically, and skips calling this function if there is no valid context.

It moves that decision into the macro itself, avoiding some #ifdef directives.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -7210,19 +7210,13 @@ S_my_setlocale_debug_string_i(pTHX_
 #ifdef USE_PERL_SWITCH_LOCALE_CONTEXT
 
 void
-Perl_switch_locale_context()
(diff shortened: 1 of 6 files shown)


locale.c
```diff
@@ -7219,10 +7219,7 @@ Perl_switch_locale_context()
 
     /* Can't use pTHX, because we may be called from a place where that
```
mauke: Why does this comment say "Can't use pTHX" if the function now uses `pTHX`?

khwilliamson: Because I made a mistake; now fixed

khwilliamson:
This doesn't work on Windows

PR #21076, opened 2023-05-05
PACKAGING: Format POD

### LINKED ISSUE #21078: Mention of "Perl 5" in POD
Many POD documents such as `PACKAGING` and `INSTALL` contain the mention of `perl 5` in `NAME` manpage/POD section.

It'd be good and more helpful for the expected future changes to generilize the mention by removing 5, and also capitalizing the P where appropriate in order to refer to the language in general, and not to a specific major version of it.

We can start by working on the core Pod files (the ones without the `.pod` extension), for instance.

rwp0:
Format POD, adding environment variable and file formatting where appropriate.

Also, remove the number 5 from "Perl" in the NAME section to generalize the application of the file.

### See also

- https://github.com/Perl/perl5/pull/21077
- https://github.com/Perl/perl5/issues/21078

### DIFF
--- a/PACKAGING
+++ b/PACKAGING
@@ -1,29 +1,29 @@
 # vim: syntax=pod
 
 If you read this file _as_is_, just ignore the funny characters you
-see.  It is written in the POD format (see pod/perlpod.pod) which is
+see.  It is written in the POD format (see F<pod/perlpod.pod>) which is
 specifically designed to be readable as is.
 
 =head1 NAME
 
-PACKAGING - notes and best practice for packaging perl 5
+PACKAGING - Notes and best practice for packaging Perl
 
 =head1 SYNOPSIS
 
 This document is aimed at anyone who is producing their own version of
-perl for distribution to other users. It is intended as a collection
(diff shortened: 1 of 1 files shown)


rwp0:
If we agree on removing "5" from Perl onward, I can also apply the same idea (and POD formatting) to the `INSTALL` document as well.

PACKAGING
```diff
@@ -1,37 +1,37 @@
...
+C<perl> for distribution to other users. It is intended as a collection
 of useful tips, advice and best practice, rather than being a complete
-packaging manual. The starting point for installing perl remains
+packaging manual. The starting point for installing C<perl> remains
 F<INSTALL>.
 
 =head1 Customizing test running
 
-A small number of porting tests (those in t/porting) are not well suited
+A small number of porting tests (those in F<t/porting)> are not well suited
```
haarg: The parenthesis should be outside the `F<>`.

rwp0: Correct, will fix that

PR #21084, opened 2023-05-10
PerlIO::via: sanity check READ method return value

tonycoz:
Coverity complained that the range of SvIV(result) could be large and well outside the bounds of the buffer at SvPVX().

So sanity check that the value of result is within range, if it is out of range warn and return an error.

Also document the bottom of the range of expected return values from READ, -1 is less than $len and resulted in a crash.

Ideally we'd just use the length of the modified $buffer to avoid such range errors, but the API is already designed, so we can't.

CID 453853

### DIFF
--- a/ext/PerlIO-via/t/via.t
+++ b/ext/PerlIO-via/t/via.t
@@ -13,7 +13,7 @@ use warnings;
 
 my $tmp = "via$$";
 
-use Test::More tests => 26;
+use Test::More tests => 32;
 
 my $fh;
 my $a = join("", map { chr } 0..255) x 10;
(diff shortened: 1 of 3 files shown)


Tux:
As PerlIO::via is CORE (and CORE only), it is hard to test against existing PerlIO::via::* modules from CPAN, as they have their own requirements.

How hard is it to de-couple it from CORE (as if it were dual-life) so I could test against PerlIO::via::Bzip2, PerlIO::via::LineNumber, PerlIO::via::MD5, PerlIO::via::Timeout, PerlIO::via::Unidecode, PerlIO::via::gzip, and PerlIO::via::xz (all installed on my development box)

Overall I *think* that I agree and approve, but uncertainty

Tux:
Tested on blead 2023-05-09.12:40:13 +0200 a12bd18fcb7a0a71c07c21e4028d7ee45474a8f9 v5.37.11-38-ga12bd18
for PerlIO::via::xz = PASS

tonycoz:
`rd` isn't an IV, these were:
```
                        Perl_warn(aTHX_ "Invalid return from %s::READ = %" IVdf ", beyond end of the returned buffer at %zu",
                                  HvNAME(s->stash), rd, buflen);
```

PR #21085, opened 2023-05-11
op/sprintf2.t: skip extended precision test on AIX

tonycoz:
This was failing in smokes on AIX

### DIFF
--- a/t/op/sprintf2.t
+++ b/t/op/sprintf2.t
@@ -1191,6 +1191,8 @@ if($Config{nvsize} == 8) {
                                    or
                                $^O eq 'hpux'
                                    or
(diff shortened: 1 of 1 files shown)


khwilliamson:
@sisyphus ISTR you did work on an issue at least similar to this on AIX

khwilliamson (approved):

sisyphus:
> @sisyphus ISTR you did work on an issue at least similar to this on AIX

I think the issue cropped up not long back wrt @Tux's AIX system.
IIRC, @Tux established that both the gcvt() and sprintf() implementations provided by libc were buggy - which means there's no straightforward fix for it.

Cheers,
Rob

Issue #21073, opened 2023-05-04
C++ builds fail with modern C++ compilers


tonycoz:
Module: Compress::Raw::Bzip2

Other modules may also be a problem.

**Description**

With a modern C++ compiler supporting and defaulting to C++ 17, the `register` storage class modifier is no longer supported, which causes the build to fail:
```
 c++ -DNO_MATHOMS -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings -O2   -DVERSION=\"2.204\" -DXS_VERSION=\"2.204\" -fPIC "-I../.."  -DBZ_NO_STDIO  compress.c
compress.c:190:13: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
            register UChar  rtmp;
            ^~~~~~~~~
```

Note that at least one C99 construct we use isn't valid in C++:

```
In file included from globals.c:25:
In file included from ./perl.h:6191:
./opcode.h:1004:2: warning: array designators are a C99 extension [-Wc99-designator]
        [OP_NULL]               = Perl_pp_null,
        ^~~~~~~~~

```
(designated struct initialization is supported in C++20 but not arrays)

**Steps to Reproduce**

```
# requires clang-16, installed here from the llvm deb archive
./Configure -des -Dusedevel -Dcc=clang++-16 -Accflags='-g -fno-omit-frame-pointer -fsanitize=address -fno-common -x c++' -Aldflags=-fsanitize=address -Accflags=-DNO_MATHOMS && make
```

**Expected behavior**

Successful build.

**Workaround**

Build with `-std=c++11`, which changes the message to a warning:
```
[...]

PR #21090, opened 2023-05-12
Revert usage of C99 array designators

Leont:
This fixes the MSVC issue mentioned in #21073, by reverting the usage of C99 array designators in `opcode.h`


tonycoz:
This fixed building `globals.obj`, once I added `/std:c++20`, which is needed for structure designators.

tonycoz (approved):

PR #21093, opened 2023-05-14
perlstyle.pod: Fix wording

rwp0:
Spelling-only.

### DIFF
--- a/pod/perlstyle.pod
+++ b/pod/perlstyle.pod
@@ -23,7 +23,7 @@ well as several other useful L<named features|feature/AVAILABLE FEATURES>).
 
     use v5.36;
 
-Regarding aesthetics of code lay out, about the only thing Larry
+Regarding aesthetics of code layout, about the only thing Larry
 cares strongly about is that the closing curly bracket of
 a multi-line BLOCK should line up with the keyword that started the construct.
 Beyond that, he has other preferences that aren't so strong:


pod/perlstyle.pod
```diff
@@ -8,7 +8,7 @@ Each programmer will, of course, have his or her own preferences in
 regards to formatting, but there are some general guidelines that will
 make your programs easier to read, understand, and maintain.
 
-The most important thing is to use L<strict> and L<warnings> in all your
+The most important thing is to use L<strict> and L<warnings> in all you
```
khwilliamson: 'your' was correct

rwp0:
Please squash before merging 

GitHub UI should allow this with the merge button

pod/perlstyle.pod
```diff
@@ -23,7 +23,7 @@ well as several other useful L<named features|feature/AVAILABLE FEATURES>).
 
     use v5.36;
 
-Regarding aesthetics of code lay out, about the only thing Larry
+Regarding aesthetics of code layout, the only thing Larry
```
khwilliamson: including 'about' is more correct than removing it, as its use indicates there may be other things not listed here, which gives wiggle room to us if something comes up

khwilliamson (changes requested):

pod/perlstyle.pod
rwp0: But the double "about" sounds weird

pod/perlstyle.pod
rwp0: Yes, see commit 2

It was a technical error from my side.

pod/perlstyle.pod
rwp0: @khwilliamson ok I added the "about" back, now it's only "lay out" that's wrong.

Could you squash and merge via GitHub or should I squash the three commits myself?

khwilliamson: I can do the squash, but it's a pain for me.  It's should be easier for you.  If you don't know how to do it, it would be good to learn for future use.  Someone can guide you through.  email me to get started.

rwp0: Done (using GitHub Desktop). I can do via CLI too, but it's just easier.

Sorry, why it's a pain for you, because of GitHub UI?

There's no such a button in GitHub?:

https://docs.gitlab.com/ee/user/project/merge_requests/squash_and_merge.html

Thanks

---

This looks like it:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits

rwp0:
Since now it's just one character change, can we merge this already?

PR #21095, opened 2023-05-14
perlhacktips: C99, but not array designators nor compound literals

tonycoz:
Addresses #21073 (but isn't a fix)


tonycoz:
Update added compound literals, which C++ also doesn't support (C++ has something similar with a different syntax.)

rwp0 (approved):

Leont (approved):

Issue #20812, opened 2023-02-15
feature 'class' appears to leak opslabs


leonerd:
The first merge of feature 'class' turned out to have many memory leaks. Most of them are fixed by #20809.

With that in place most of the more basic tests no longer leak, but any field definition that includes a `:param` attribute or a `= EXPR` defaulting expression still does. In both cases the leak is visible in `valgrind` as:

```
==3139978== 16,136 (552 direct, 15,584 indirect) bytes in 1 blocks are definitely lost in loss record 6 of 6
==3139978==    at 0x48407B4: malloc (vg_replace_malloc.c:381)
==3139978==    by 0x153171: S_new_slab (op.c:254)
==3139978==    by 0x1535DA: Perl_Slab_Alloc (op.c:346)
==3139978==    by 0x16D01B: Perl_newSVOP (op.c:7572)
==3139978==    by 0x225D01: S_pending_ident (toke.c:9967)
==3139978==    by 0x22380B: Perl_yylex (toke.c:9592)
==3139978==    by 0x23DF02: Perl_yyparse (perly.c:357)
==3139978==    by 0x19E9DF: S_parse_body (perl.c:2614)
==3139978==    by 0x19C54F: perl_parse (perl.c:1923)
==3139978==    by 0x14E285: main (perlmain.c:106)
```

which suggests it's because an op gets allocated somewhere and isn't released again. But so far I haven't been able to find it.

tonycoz:
I see a leak from LSAN even without an initializer.  In the following I've rebuilt perl to add class to the default feature set to reduce noise:
```
$ LSAN_OPTIONS=report_objects=1 PERL_DESTRUCT_LEVEL=2 ./perl -DS -Ilib -e 'class Foo { field $x; }'
class is experimental at -e line 1.
allocated new op slab sz 0x40, 616000000c80, head slab 616000000c80 at -e line 1.
allocating op at 616000000e78, slab 616000000c80, head slab 616000000c80 at -e line 1.
freeing    op at 616000000e78, slab 616000000c80, head slab 616000000c80 at -e line 1.
field is experimental at -e line 1.
realloced  op at 616000000e78, slab 616000000c80, head slab 616000000c80 at -e line 1.
allocating op at 616000000e48, slab 616000000c80, head slab 616000000c80 at -e line 1.
allocating op at 616000000de8, slab 616000000c80, head slab 616000000c80 at -e line 1.
[...]

PR #21104, opened 2023-05-19
skip leaking test in class feature

iabyn:
This test is known to leak: see GH #20812. Skip it for now so that ASAN smokes don't fail. Start failing again on the next development branch so that the issue isn't forgotten

### DIFF
--- a/t/lib/croak/class
+++ b/t/lib/croak/class
@@ -98,6 +98,10 @@ EXPECT
 Global symbol "$self" requires explicit package name (did you forget to declare "my $self"?) at - line 5.
 Execution of - aborted due to compilation errors.
 ########
+# This test is known to leak: see GH #20812. Skip it for now so that ASAN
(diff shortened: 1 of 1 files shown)


leonerd (approved):
LGTM

jkeenan:
Configuring in blead with these arguments:
```
$ sh ./Configure -des -Dusedevel -Dcc=clang -Accflags="-g -fno-omit-frame-pointer -fsanitize=address  -fno-common -fsanitize-blacklist=`pwd`/asan_ignore" -Aldflags="-fsanitize=address"
```
... I then ran the test file:
```
$ cd t;PERL_DESTRUCT_LEVEL=2 ./perl harness ./lib/croak.t; cd -
```
... and confirmed the test failure.

I then switched to `davem/skip_croak`, repeated the process, and confirmed that the failing test is now skipped.  Merging.

PR #21107, opened 2023-05-19
Davem/deparse

iabyn:
Fix up regressions in Deparse.pm and update the list of files to skip/ignore for ./TEST -deparse.
I propose this gets merged before 3.38.0-RC1.

This is based on  the idea that about once per year I run 'cd t; ./TEST -deparse', which runs all the test files through Deparse before executing them, and see what test files have started failing or unexpectedly succeeding. (I normally do this not quite so close to the release date!)

The main fixes to Deparse.pm are:
* honour signatures under 'use v5.36' in addition to honouring under 'use feature "signatures"'.
* handle the new OP_PADSV_STORE op in list context. This was failing to deparse: @a = (my $x = 1).

And quite a few files have been added to Porting/deparse-skips.txt to skip new failing things for now, such as class/*.t

### DIFF
--- a/lib/B/Deparse.pm
+++ b/lib/B/Deparse.pm
@@ -7,7 +7,7 @@
 # This is based on the module of the same name by Malcolm Beattie,
 # but essentially none of his code remains.
 
-package B::Deparse 1.73;
+package B::Deparse 1.74;
 use strict;
 use Carp;
 use B qw(class main_root main_start main_cv svref_2object opnumber perlstring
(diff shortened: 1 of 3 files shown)


rwp0 (approved):

lib/B/Deparse.pm
```diff
@@ -3878,6 +3879,9 @@ sub pp_list {
 	    if ($padname->FLAGS & PADNAMEf_TYPED) {
 		$newtype = $padname->SvSTASH->NAME;
 	    }
+	} elsif ($lopname eq 'padsv_store') {
```
druud, line 3882: Indentation seems off?

iabyn:
On Mon, May 22, 2023 at 01:51:26AM -0700, druud wrote:
> @druud commented on this pull request.
> 
> 
> 
> > @@ -3878,6 +3879,9 @@ sub pp_list {
>  	    if ($padname->FLAGS & PADNAMEf_TYPED) {
>  		$newtype = $padname->SvSTASH->NAME;
>  	    }
> +	} elsif ($lopname eq 'padsv_store') {
> 
> Indentation seems off?

It's within a larger if block which looks like

	if (...) {
	    ...
	} elsif ($lopname eq 'padsv_store') {
            ...
	} elsif (...) {
	    ....
	} elsif (...)
	{
            ...
	} elsif (...)
	{
            ....
	}

and I just kept the same style when adding the extra elsif.

-- 
Never do today what you can put off till tomorrow.

Issue #19890, opened 2022-06-25
Commit 996e6b9e17 breaks build with -Accflags="-DNO_LOCALE_NUMERIC"


jkeenan:
When I compile with this `./Configure` switch:
```
-Accflags="-DNO_LOCALE_NUMERIC"
```
... compilation fails on both OpenBSD and FreeBSD (using `clang10` on both OSes) at this point:
```
$ sh ./Configure -des -Dusedevel -Accflags="-DNO_LOCALE_NUMERIC" && make miniperl | tee $P5P_DIR/make-output/`extract-sha`.openbsd.clang10.unthreaded.onlynolocalenumeric.makeminiperl.output.txt
...
cc -c -DPERL_CORE -DNO_LOCALE_NUMERIC -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings locale.c
locale.c:2293:13: error: use of undeclared identifier 'toggled'
        if (toggled) {
            ^
1 error generated.
*** Error 1 in /home/jkeenan/gitwork/perl (makefile:246 'locale.o': @`sh  cflags "optimize='-O2'" locale.o`  locale.c)
```
The above was observed on OpenBSD-6.9 at commit 7b92fe48f0def2bea3db4aa70f352ccf9d83b43a.  I have reproduced it on FreeBSD-12 as well.  I originally spotted this problem in this smoke-test report:  https://perl5.test-smoke.org/report/5019634.  I customarily smoke-test on OpenBSD with `-Accflags="-DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE"` because that is closer to the "system perl" configuration options used by @afresh1.

On an older Linux installation, with the same configuration I get:
```
$ uname -mrs
Linux 5.10.13-x86_64-linode141 x86_64
$ ./Configure -des -Dusedevel -Accflags="-DNO_LOCALE_NUMERIC" && make miniperl
...
[...]

PR #21108, opened 2023-05-20
v5.38.0: release branch

rjbs:


pod/perldelta.pod
```diff
@@ -142,25 +140,48 @@ These functions are closely tied to the internals of how the interpreter
...
+=item *
+
+If your code reads to end of file, and then rebinds the handle to a
+new file descriptor, previously since the eof flag wasn't set you
+could continue to read from the stream.  You now need to clear the eof
+flag yourself with C<< $handle->clearerr() >> to continue reading.
+
+=item *
+
+If your code encounters and error on the stream while reading with
```
khwilliamson: s/and error/an error/

pod/perldelta.pod
```diff
@@ -1324,10 +1334,9 @@ the C<localeconv()> fields), but directly callable from XS code.
 
 =item *
 
-A new API function L<perlapi/C<my_strftime8>> is added.  This is the
-same as plain L<perlapi/C<my_strftime>>, but with an extra parameter
-that allows the caller to simply and reliably know if the returned
-string is UTF-8.
+A new API function my_strftime8 is added.  This is the same as plain
```
khwilliamson: This function has been removed

pod/perldelta.pod
```diff
@@ -1473,13 +1482,13 @@ Avoid signed integer overflow in C<use integer> ops.
 
 Avoid adding an offset to a NULL pointer in C<hv_delete_common>.
 
-=item * PerlIO::get_layers will now accept IO references too
+=item *
```
khwilliamson, line 1619: Earlier line

> Fix querying locale without LC_NUMERIC on platform [L<github #19890|https://github.com/Perl/perl5/issues/19890>]
better written
> Fix a bug with querying locales on platforms that don't have LC_NUMERIC ...

khwilliamson (changes requested):

pod/perldelta.pod
```diff
@@ -407,7 +1632,7 @@ died, add a short obituary here.
 
 XXX Generate this with:
 
-  perl Porting/acknowledgements.pl v5.37.11..HEAD
+  perl Porting/acknowledgements.pl v5.37.0..HEAD
```
jkeenan, line 1802: Shouldn't that line read `perl Porting/acknowledgements.pl v5.39.0..HEAD`?

jkeenan (changes requested):

pod/perldelta.pod
rjbs: Yes.  This text is from some autogeneration code, which tends to not be good at blead-final releases.

rjbs: No, it is correct.

This document is the v5.38.0 delta, so should credit everyone with changes since v5.36.0 or v5.37.0.  As I am the only person in the acknowledgements for v5.37.0, it doesn't matter which we pick.

tonycoz:
When I did the deprecation patch for `'` in symbols I followed the (then) RFC and documented it would be removed in 5.40, and that's now in the 5.38.0 perldelta.

There were later updates to perldeprecation indicating it would be removed in 5.42.

Which is correct?

rjbs:
Deprecation warnings are nearly always in place for 2y.  I have updated the perldelta to say `'` for package sep comes out in v5.42.

PR #21125, opened 2023-05-25
Configure: Add distribution-friendly make message

rwp0:
### Table

| OS | File | Package | Command |
|--------|--------|--------|--------|
| Debian | `/etc/debian_version` | [`build-essential`](https://packages.debian.org/sid/build-essential) | `apt install build-essential` |
| OpenSUSE | `/etc/SUSE-brand` | [`patterns-devel-C-C++`](https://packagehub.suse.com/packages/patterns-devel-C-C++/) | `zypper install 'pattern:devel_C_C++'` |
| Red Hat | `/etc/redhat-release` | `Development Tools` | `yum groupinstall 'Development Tools'`|
| Fedora | `/etc/fedora-release` | `Development Tools` | `dnf group install 'Development Tools'` |

View the raw diff here (esp. if the conversations overshadow the code in the `Files changed` tab):

- https://patch-diff.githubusercontent.com/raw/Perl/perl5/pull/21125.diff

*Couldn't find a link for `Development Tools` Yum package group searching* 😢 

-----

Straight to the point as
"Public domain implementation" wasn't very helpful.

Came across this message building Perl on a freshly
installed Ubuntu Linux doing:

```
git clone https://github.com/Perl/perl5
sh Configure -des -Dusedevel
```


rwp0:
https://packages.ubuntu.com/focal/build-essential contains everything (`gcc`/`make`/`libc-dev`) needed to compile a program from its source code.

Leont:
I'm not sure it it's appropriate to have such a platform specific error message to all platforms.

That said, you can detect debian-likeness with `test -e /etc/debian_version`

rwp0:
> I'm not sure it it's appropriate to have such a platform specific error message to all platforms.
> 
> That said, you can detect debian-likeness with `test -e /etc/debian_version`

Agreed, thanks for the suggestion. How about now?

```sh
gmake)
	echo "I can't find make or gmake, and my life depends on it." >&4
	echo "Go find a public domain implementation or fix your PATH setting!" >&4
	test -e /etc/debian_version && echo "Run \"apt install build-essential\" to install make/gcc/libc-dev packages" >&4
	exit 1
	;;
	esac
```

Tux:
I can live with that. Can you extend it to other systems? openSUSE/zypper, */dnf (CentOS 8+, Rockylinux, …), */yum (CentOS ..7, RHEL, …), */apt-get (debian-like before apt), more?

rwp0:
> I can live with that. Can you extend it to other systems? openSUSE/zypper, */dnf (CentOS 8+, Rockylinux, …), */yum (CentOS ..7, RHEL, …), */apt-get (debian-like before apt), more?

Yes, but we need to determine and agree upon how to detect those systems, and also what package(s) to suggest for installation.

Tux:
openSUSE:

`[ -f /etc/SUSE-brand ] || [ grep -q -i opensuse /etc/os-release ]`
command: `zypper in 'pattern:devel_C_C++'`

I do not use that myself, as that also installs `.info` packages, which I experience as utterly useless. Worst way ever to write documentation and manuals

rwp0:
> [ grep -q -i opensuse /etc/os-release ]

@Tux 
Why did you surround grep with brackets? :-)
I think that's the package in question:
https://packagehub.suse.com/packages/patterns-devel-C-C++/

For RHEL-based ones we can do the following I believe:
```
yum groupinstall 'Development Tools'
```

### See

- https://unix.stackexchange.com/a/196910/455788

Tux:
I should have checked. You're right, brackets around grep not wanted

Configure
```diff
@@ -2673,6 +2673,7 @@ make)
 	gmake)
 	echo "I can't find make or gmake, and my life depends on it." >&4
 	echo "Go find a public domain implementation or fix your PATH setting!" >&4
```
rwp0: Not sure if it is a good idea to replace "public domain" with "open source" as strictly speaking there might be no current PD implementation of such tools at all.

rwp0:
Amended the commit to include OpenSUSE, Red Hat, Fedora.
See the [table](https://github.com/Perl/perl5/pull/21125#issue-1726036507) I made to map the associations.
More distribution suggestions welcome.
Thanks.

Configure
```diff
@@ -2673,6 +2673,12 @@ make)
...
+
+	test -f /etc/debian_version && echo "Run \"apt install build-essential\" to install make and friends" >&4
```
rwp0: `test -e` would be true for directories as well, so using `-f` here to limit the detection to files.

rwp0:
If no objections, can we merge this before 5.38?

This can be considered a kind of a minor documentation, I believe.

We don't have to add all the distributions, we can (and should in my opinion) grow that gradually possibly referencing this PR.
Having the 4 major Linux distros now should be enough for the start of such a work for adding distro-firendly messages.

Configure
rwp0: I can change the command to `apt-get` to include the legacy systems as well, but I don't think it's necessary since `apt` is the way to go, and users are expected to update their package manager anyway.

- https://itsfoss.com/apt-vs-apt-get-difference/

Targetting old stuff to the harm of the new is a bad thing to do IMHO.

Leont (approved):

rwp0:
@khwilliamson can we merge this too now after the 5.38.0 release?

If still not please elaborate so I understand and improve if possible?

Tux:
I'd say go for it.

One tiny nit: align the `&&`'s

Alternatively, I can commit that to the metaunit and merge, but that would omit you as author

rwp0:
> I'd say go for it.
> 
> One tiny nit: align the `&&`'s
> 
> Alternatively, I can commit that to the metaunit and merge, but that would omit you as author

I agree, I was also thinking about aligning, will finish it today.

@Tux is it okay if I add that as second commit (can it be squashed after as one)?

rwp0:
@Tux please go on and add commit to the metaunit and merge because I messed things up with my IDE :-)

```sh
	test -f /etc/debian_version && echo "Run \"apt install build-essential\" to install make and friends" >&4
	test -f /etc/SUSE-brand     && echo "Run \"zypper install 'pattern:devel_C_C++'\" to install make and friends" >&4
	test -f /etc/redhat-release && echo "Run \"yum groupinstall 'Development Tools'\" to install make and friends" >&4
	test -f /etc/fedora-release && echo "Run \"dnf group install 'Development Tools'\" to install make and friends" >&4
```

Whenever I need to sync with upstream Perl (after a long time and changes) this hell happens for me (as the IDE adds "new files" for some unknown reason to me). That's why I get mad at modifying delayed PR's 😄

rwp0:
@Tux anyway created the below PR (with operators aligned and whitespace removed) once had free time:

- https://github.com/Perl/perl5/pull/21207

PR #21136, opened 2023-06-01
warnings doc improvements

hvds:
- fix a couple of typos;
- clarify that 'use warnings' enables even warnings that have not yet been registered;
- rename function name in an example, since a later reference relates to the same function from a previous example;
- link to a more specific point from warnings::register

### DIFF
--- a/regen/warnings.pl
+++ b/regen/warnings.pl
@@ -16,7 +16,7 @@
 #
 # This script is normally invoked from regen.pl.
 
-$VERSION = '1.64';
+$VERSION = '1.65';
 
 BEGIN {
     require './regen/regen_lib.pl';
--- a/regen/warnings.pl
+++ b/regen/warnings.pl
@@ -1438,7 +1438,7 @@ X<warning, reporting> X<warning, registering>
 
 The C<warnings> pragma provides a number of functions that are useful for
 module authors.  These are used when you want to report a module-specific
-warning to a calling module has enabled warnings via the C<warnings>
+warning to a calling module that has enabled warnings via the C<warnings>
 pragma.
 
 Consider the module C<MyMod::Abc> below.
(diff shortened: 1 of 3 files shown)


hvds:
@demerphq does the example use of 'deprecated' in a module need updating in the light of your recent improvements? See near `warnings.pm` line 1260.

rwp0 (approved):

demerphq:
> does the example use of 'deprecated' in a module need updating in the light of your recent improvements?

@hvds: not that I can see. The old usage still works, and you can still use 

```
no warnings 'deprecated';
```

to disable all 'deprecated' warnings, the only difference is that you can now do this at a per warning level, and vice versa.

hvds:
@demerphq thanks for the confirmation.

regen/warnings.pl
```diff
@@ -1476,26 +1487,26 @@ package name.  The C<open> function in the module will display a warning
...
             warnings::warn("deprecated",
-                           "open is deprecated, use new instead");
+                           "open2 is deprecated, use open instead");
         }
-        new(@_);
+        open(@_);
     }
 
     sub new
     ...
```
tonycoz: I don't think the rename adds anything, I think the story the docs are trying to tell is the open() in the original version of the module (above) is being deprecated in favour of new(), which has a stub at the bottom of the chunk.

hvds: I failed to spot the stub - is removing that stub (on top of the current patch) enough to fix the problem and make the resulting docs coherent?

The next code chunk in the doc suggests that `use warnings FATAL => 'MyMod::Abc';` will fatalize a warning triggered by `MyMod::Abc::open()`. But if `sub open { ... }` has only a "deprecated" warning, that fatalization won't trigger. So something needs to change for the fatalization example to work. (I have not checked, but I'm guessing that the 'deprecated' example was added later, missing the fact that the fatalization example was relying on the original definition of `sub open`.)

tonycoz: I think you're right.

The fatalization example could call new() instead, but then I think the stub would need to be expanded to include the old open() implementation.

tonycoz (approved):

PR #21146, opened 2023-06-07
Errno: also escape osvers

### LINKED ISSUE #21135: Errno archname/osvers too strict.
Module: Errno

**Description**
The architecture check generated by [ext/Errno/Errno_pm.PL at line 300](https://github.com/Perl/perl5/blob/697eaf8/ext/Errno/Errno_pm.PL#L300) is too strict, relying on identical operating system identifiers that may have changed since Perl was installed.

The resulting file ends up at e.g. `.../perl/5.36/x86_64-linux/Errno.pm`

If the kernel version string changed, but the architecture and OS are otherwise consistent, the process will `die` on a failed comparison.

**Steps to Reproduce**
No simple script, but it should be reproducible by changing insignificant version information (like kernel minor version) after installing Perl, when running build scripts of Perl modules.

Two cases reported [here](https://forums.gentoo.org/viewtopic-t-1147768.html?sid=785ec7cded64da3e16986bc1ac0175e7).

**Expected behavior**
The check should validate on `uname --kernel-name` (Linux) and not `uname --kernel-release` (6.3.5-whatever-00013-g0c0ffee).

**Perl configuration**
Perl 5.36

tonycoz:
Partly fixes #21135

Tested by Configuring with

  ... -Darchname='x86_64-linux@alpha' -Dosvers='4.19.0@beta'

### DIFF
--- a/ext/Errno/Errno_pm.PL
+++ b/ext/Errno/Errno_pm.PL
@@ -2,7 +2,7 @@ use ExtUtils::MakeMaker;
 use Config;
 use strict;
 
-our $VERSION = "1.37";
+our $VERSION = "1.38";
 
 my %err = ();
 
@@ -271,9 +271,9 @@ sub write_errno_pm {
 	close(CPPO);
     }
 
-    # escape $Config{'archname'}
-    my $archname = $Config{'archname'};
-    $archname =~ s/([@%\$])/\\$1/g;
+    # escape $Config{'archname'}, $Config{'osvers'}
+    my ($archname, $osvers) = @Config{'archname', 'osvers'};
+    $_ = quotemeta for $archname, $osvers;
 
     # Write Errno.pm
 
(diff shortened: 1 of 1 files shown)


khwilliamson (approved):

ext/Errno/Errno_pm.PL
```diff
@@ -271,9 +271,9 @@ sub write_errno_pm {
 	close(CPPO);
     }
 
-    # escape $Config{'archname'}
-    my $archname = $Config{'archname'};
-    $archname =~ s/([@%\$])/\\$1/g;
+    # escape $Config{'archname'}, $Config{'osvers'}
+    my ($archname, $osvers) = @Config{'archname', 'osvers'};
+    s/([@%\$])/\\$1/g for $archname, $osvers;
```
mauke: `%` isn't special in double-quoted strings and doesn't need to be escaped. On the other hand, `"` and `\` probably should be escaped. (Or just `quotemeta` the whole thing?)

tonycoz: Unfortunately adding " to archname or osvers breaks the build when compiling perl, since neither ARCHNAME nor OSVER is escaped in config_h.PL.  I expect `\` would have similar problems.

I've changed this use quotemeta, so if we fix `config_h.SH` these will be handled correctly.

Leont (approved):

PR #21150, opened 2023-06-09
perl: Fix CVE_2023_31486

SoumyaWind:
HTTP::Tiny 0.082, a Perl core module since 5.13.9 and available standalone on CPAN, has an insecure default TLS configuration where users must opt in to verify certificates.

Reference: https://salsa.debian.org/perl-team/interpreter/perl/-/commit/1490431e40e22052f75a0b3449f1f53cbd27ba92.patch

### DIFF
--- a/cpan/HTTP-Tiny/lib/HTTP/Tiny.pm
+++ b/cpan/HTTP-Tiny/lib/HTTP/Tiny.pm
@@ -40,7 +40,7 @@ sub _croak { require Carp; Carp::croak(@_) }
 #pod * C<timeout> — Request timeout in seconds (default is 60) If a socket open,
 #pod   read or write takes longer than the timeout, an exception is thrown.
 #pod * C<verify_SSL> — A boolean that indicates whether to validate the SSL
-#pod   certificate of an C<https> — connection (default is false)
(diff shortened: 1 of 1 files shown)


Grinnz:
HTTP::Tiny is maintained on CPAN and changes must be submitted there (and have been already). See https://github.com/chansen/p5-http-tiny/pull/153

cpan/HTTP-Tiny/lib/HTTP/Tiny.pm
```diff
@@ -112,7 +112,7 @@ sub new {
         max_redirect => 5,
         timeout      => defined $args{timeout} ? $args{timeout} : 60,
         keep_alive   => 1,
-        verify_SSL   => $args{verify_SSL} || $args{verify_ssl} || 0, # no verification by default
+        verify_SSL   => $args{verify_SSL} // $args{verify_ssl} // 1, # verification by default
```
mauke, line 115: That's an interesting change in behavior. Consider:
```perl
my $http = HTTP::Tiny->new(
    verify_SSL => 0,
    verify_ssl => 1,
);
```
With the old code, this turns host verification on. With your change, this turns host verification **off**.

PR #21152, opened 2023-06-13
getpgrp: getpgid used for PID != 0 is now a lot more portable

### LINKED ISSUE #19648: getpgrp is now more portable than docs claim
**Description**

According to the docs for `getpgrp`:

> Note that the POSIX version of "getpgrp" does not accept a PID argument, so only "PID==0" is truly portable.

But the POSIX spec made the old behavior available in the `getpgid` function, which was moved from XSI to Base in Issue 7 (2018):
https://pubs.opengroup.org/onlinepubs/9699919799/functions/getpgid.html

This comment in perl.h is 26 years old:
https://github.com/Perl/perl5/blob/2808c4d89767b5122c7c116ebeb5679e006dc0df/perl.h#L865-L866

But it shows that `getpgid` is being used under the hood for the implementation of Perl's `getpgrp` function. So the docs should probably be updated to state that it's portable on systems conforming to the relatively recent POSIX standard.

tonycoz:
POSIX.1 2004 lists getpgid as an XSI extension, 2017 edition lists it as BASE.

Removed the comments since it now seems irrelevant.

Fixes #19648

### DIFF
--- a/perl.h
+++ b/perl.h
@@ -1027,9 +1027,6 @@ violations are fatal.
 #  define HAS_SETPGRP  /* Well, effectively it does . . . */
 #endif
 
-/* getpgid isn't POSIX, but at least Solaris and Linux have it, and it makes
-    our life easier :-) so we'll try it.
-*/
 #ifdef HAS_GETPGID
 #  define BSD_GETPGRP(pid)		getpgid((pid))
 #elif defined(HAS_GETPGRP) && defined(USE_BSD_GETPGRP)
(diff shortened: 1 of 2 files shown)


jkeenan:
> POSIX.1 2004 lists getpgid as an XSI extension, 2017 edition lists it as BASE.
> 
> Removed the comments since it now seems irrelevant.
> 
> Fixes #19648

Looks good to me, but it would be good if this were reviewed by someone much more knowledgeable than me concerning POSIX.

tonycoz:
I found the references via google:

https://pubs.opengroup.org/onlinepubs/009696799/functions/getpgid.html (getpgid() as XSI in IEEE Std 1003.1, 2004 Edition)

https://pubs.opengroup.org/onlinepubs/9699919799/functions/getpgid.html (getpgid() as base in IEEE Std 1003.1-2017)

PR #21161, opened 2023-06-20
no bareword_filehandes: don't look up class barewords as handles

### LINKED ISSUE #19426: Method resolution still searches inside bareword handles despite `no feature "bareword_filehandles"`
```perl
use v5.35;
use Future;
{ use feature 'bareword_filehandles'; open Future, "<", "/dev/null"; }     
my $f = Future->done(123);
```

Fails with:
```
Name "main::Future" used only once: possible typo at - line 3.
Can't locate object method "done" via package "IO::File" at - line 4.
```

This indicated that, despite the `no feature "bareword::filehandles"` that was in scope at toplevel, it nonetheless tried to resolve the `->done` method call via that bareword handle that was previously opened earlier, taking precedence over the otherwise-expected resolve via named package.

tonycoz:
This prevents SomeBareword from being looked up as a filehandle:

  no feature "bareword_filehandles";
  SomeBareword->method();

Since "bareword_filehandles" is explicitly about bareword handles, it does not prevent other string to object/class look ups from resolving the class as a filehandle, eg for the following:

  "SomeLiteral"->method();
  my $x = "SomeVariable";
  $x->method();

both are looked up as file handles per normal.

Note that in any of these cases, with or without the bareword_filehandles feature, if the name is a bareword that has been resolved as a class name since the last time the stash cache was cleared, it will be resolved as a class name, not a filehandle.

Fixes #19426


tonycoz:
I don't believe this belongs in 5.38 at this stage, it may change the behaviour of working code.

jkeenan:
> I don't believe this belongs in 5.38 at this stage, it may change the behaviour of working code.

@tonycoz, do you still want to pursue this p.r. (now that we're past 5.38.0)?  If so, you will need to resolve some conflicts.  Thanks.

tonycoz:
Rebased, waiting for review, preferably on both the behaviour and the implementation.

op.c
```diff
@@ -14839,6 +14852,12 @@ Perl_ck_subr(pTHX_ OP *o)
                 STRLEN len;
                 const char* str = SvPV(*const_class, len);
                 if (len) {
+                    if (!FEATURE_BAREWORD_FILEHANDLES_IS_ENABLED
+                        && !is_standard_filehandle_name(str)
+                        && (const_op->op_private & OPpCONST_BARE)) {
```
leonerd, line 14903: Is there a possibility that `const_op` remains NULL by this point? Should there be a test on it? If not, should it be `assert()`ed?

leonerd:
Added one comment about the implementation, but I think overall the intent of the behaviour is good. It means that with that feature enabled, the behaviour of "Name->foo" is simpler to explain and reason about, because it doesn't depend on (dynamic) filehandle setup any more.

op.c
tonycoz: That's not possible, this code is conditional on `const_class` and `const_class` is set from the op used to set `const_op` at the same point `const_op` is set.

If this code wasn't also conditional on `SvPV(*const_class)` I would have made the if() conditional on `const_op` instead and moved initializing `const_class` within the if() block.

I'll add an assert().

leonerd (approved):
LGTM

PR #21165, opened 2023-06-22
tick in symbols: rename new prototype scan_word and provide a forwarder

### LINKED ISSUE #21162: BBC 5.37.x breaks B::Hooks::Parser
E.g. 

```
Parser.xs:175:73: error: macro "scan_word" requires 6 arguments, but only 5 given
  175 |         return scan_word (base_s, dest, destlen, handle_package, res_len);
      |                                                                         ^
```
(from http://www.cpantesters.org/cpan/report/62c5a506-04c9-11ee-8cc4-947bc154cd3f)

Seems like it is using an internal API function, whose signature was changed recently.

tonycoz:
The change in prototype of the non-API scan_word() function caused B::Hooks::Parser to fail to build.

Instead, rename scan_word() with the new prototype to scan_word2() and provide a scan_word() with the old signature.

The scan_word2() entry may be removed once tick in symbol becomes an error.

Fixes #21162

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -2903,6 +2903,11 @@ EXpx	|char * |scan_word	|NN char *s				\
 				|NN char *dest				\
 				|STRLEN destlen 			\
 				|int allow_package			\
+				|NN STRLEN *slp
+EXpx	|char * |scan_word6	|NN char *s				\
+				|NN char *dest				\
+				|STRLEN destlen 			\
+				|int allow_package			\
 				|NN STRLEN *slp 			\
 				|bool warn_tick
 Cp	|U32	|seed
--- a/embed.h
+++ b/embed.h
@@ -1703,7 +1703,8 @@
 #   define reg_temp_copy(a,b)                   Perl_reg_temp_copy(aTHX_ a,b)
 #   define report_uninit(a)                     Perl_report_uninit(aTHX_ a)
 #   define scan_str(a,b,c,d,e)                  Perl_scan_str(aTHX_ a,b,c,d,e)
-#   define scan_word(a,b,c,d,e,f)               Perl_scan_word(aTHX_ a,b,c,d,e,f)
+#   define scan_word(a,b,c,d,e)                 Perl_scan_word(aTHX_ a,b,c,d,e)
+#   define scan_word6(a,b,c,d,e,f)              Perl_scan_word6(aTHX_ a,b,c,d,e,f)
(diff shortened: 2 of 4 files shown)


leonerd:
Tiniest comment: It's common in e.g. Linux kernel, when adding new functions with more arguments, to use the number of arguments to name the function. So here it'd be `scan_word6`. Do we want to consider doing similar in this case?

ilmari:
> Tiniest comment: It's common in e.g. Linux kernel, when adding new functions with more arguments, to use the number of arguments to name the function. So here it'd be `scan_word6`. Do we want to consider doing similar in this case?

There's precedent for that in perl too:

```
~/src/perl (blead $)$ rg '^Perl_\w+\d+\(' proto.h |rg -vi 'utf8|latin1|[iu](8|16|32|64)'
Perl_csighandler1(int sig);
Perl_csighandler3(int sig, Siginfo_t *info, void *uap);
Perl_do_aexec5(pTHX_ SV *really, SV **mark, SV **sp, int fd, int do_report)
Perl_gv_efullname4(pTHX_ SV *sv, const GV *gv, const char *prefix, bool keepmain);
Perl_gv_fullname4(pTHX_ SV *sv, const GV *gv, const char *prefix, bool keepmain);
Perl_my_atof2(pTHX_ const char *orig, NV *value);
Perl_my_atof3(pTHX_ const char *orig, NV *value, const STRLEN len);
Perl_pregfree2(pTHX_ REGEXP *rx);
Perl_sighandler1(int sig)
Perl_sighandler3(int sig, Siginfo_t *info, void *uap)
Perl_sys_init3(int *argc, char ***argv, char ***env);
Perl_gv_efullname3(pTHX_ SV *sv, const GV *gv, const char *prefix);
Perl_gv_fullname3(pTHX_ SV *sv, const GV *gv, const char *prefix);
Perl_do_exec3(pTHX_ const char *incmd, int fd, int do_report)
```

tonycoz:
> Tiniest comment: It's common in e.g. Linux kernel, when adding new functions with more arguments, to use the number of arguments to name the function. So here it'd be `scan_word6`. Do we want to consider doing similar in this case?

Good idea, done.

leonerd (approved):
LGTM

As it's a BBC issue, do we want to consider putting this in before 5.38.0-RC2?

ilmari (approved):

ilmari:
> As it's a BBC issue, do we want to consider putting this in before 5.38.0-RC2?

I thinks so, but I guess it's up to @rjbs to decide.

PR #21166, opened 2023-06-22
Use C<> for code example in perldelta.pod

dboehmer:


pod/perldelta.pod
```diff
@@ -180,7 +180,7 @@ the pattern will stay O(N).
 =head2 REG_INF has been raised from 65,536 to 2,147,483,647
 
 Many regex quantifiers used to be limited to U16_MAX in the past, but are
-now limited to I32_MAX, thus it is now possible to write /(?:word){1000000}/
+now limited to I32_MAX, thus it is now possible to write C</(?:word){1000000}/>
```
ilmari, line 183: I think the `U16_MAX` and `I32_MAX` should have `C<>` too.

dboehmer: Thanks for the fast reply! I thought about that, too, and actually compared to previous POD. I found in https://perldoc.perl.org/5.37.9/perldelta#REG_INF-has-been-raised-from-65,536-to-2,147,483,647 it wasn’t highlighted as code.

Should I amend the PR?

rjbs:
I will perform the second change described in the comments.

PR #21169, opened 2023-06-23
perldelta.pod: Link PPC's

rwp0:
People might not get what PPC is.

Let's make them visible by linking to the Perl/PPCs repository's previewable markdown files.

Noticed while reading the `perldelta` for today's release (RC2):

- https://metacpan.org/release/RJBS/perl-5.38.0-RC2/view/pod/perldelta.pod

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -402,11 +402,11 @@ Added the C<is_tainted()> builtin function. [L<github #19854|https://github.com/
 
 =item *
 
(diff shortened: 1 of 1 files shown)


rwp0:
Could also link to a permalink, but I think that's not needed, since those PPC's shouldn't change I believe.

rwp0:
@rjbs FYI

This relates and should be merged _in addition to_ with https://github.com/Perl/perl5/pull/21172 if so

PR #21177, opened 2023-06-26
Minor tweaks to perldelta.pod

esabol:
None of the other headers end with a period, so remove the period from the end of this one header. Also, the previous header says just "no longer", so change "will no longer" in this same header to just "no longer" to match the style of previous header. And remove an unnecessary comma.


esabol:
@rjbs FYI

rwp0 (approved):

PR #21180, opened 2023-06-28
Time-HiRes: don't warn in C++ builds

tonycoz:
g++ warns here:

warning: option ‘-Wc++-compat’ is valid for C/ObjC but not for C++ [-Wpragmas]

so don't try to suppress this warning for C++ builds.

I wasn't able to find a platform where the first argument to setitimer() or getitimer() is or was an enum.

There's a similar use in ODBM_File, but ODBM_File is never built as C++ since it uses a function with the same name as a C++ keyword.

There's also similar use in Encode, but we don't control that, and I don't think this warning justifies customization.

### DIFF
--- a/dist/Time-HiRes/Changes
+++ b/dist/Time-HiRes/Changes
@@ -17,6 +17,8 @@ Revision history for the Perl extension Time::HiRes.
    https://github.com/Perl/perl5/issues/19321
  - don't use C++ guards around the perl header files, it caused C++
    build failures with MSVC.
+ - don't try to suppress C++ compatibility warnings in C++ builds, since
+   that warns.
 
 1.9764 [2020-08-10]
  - Fix a bunch of repeated-word typos
--- a/dist/Time-HiRes/HiRes.pm
+++ b/dist/Time-HiRes/HiRes.pm
@@ -50,7 +50,7 @@ our @EXPORT_OK = qw (usleep sleep ualarm alarm gettimeofday time tv_interval
                  stat lstat utime
                 );
 
-our $VERSION = '1.9775';
+our $VERSION = '1.9776';
 our $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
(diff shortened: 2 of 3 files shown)


xenu:
>I wasn't able to find a platform where the first argument to setitimer() or getitimer() is or was an enum.

glibc when `_GNU_SOURCE` is defined, but **not** in the C++ mode, so that warning is indeed bogus:

```C
enum __itimer_which
  {
    /* Timers run in real time.  */
    ITIMER_REAL = 0,
#define ITIMER_REAL ITIMER_REAL
    /* Timers run only when the process is executing.  */
    ITIMER_VIRTUAL = 1,
#define ITIMER_VIRTUAL ITIMER_VIRTUAL
    /* Timers run when the process is executing and when
       the system is executing on behalf of the process.  */
    ITIMER_PROF = 2
#define ITIMER_PROF ITIMER_PROF
  };
```

```C
#if defined __USE_GNU && !defined __cplusplus
/* Use the nicer parameter type only in GNU mode and not for C++ since the
   strict C++ rules prevent the automatic promotion.  */
typedef enum __itimer_which __itimer_which_t;
#else
typedef int __itimer_which_t;
#endif

#ifndef __USE_TIME_BITS64
/* Set *VALUE to the current setting of timer WHICH.
   Return 0 on success, -1 on errors.  */
extern int getitimer (__itimer_which_t __which,
                      struct itimerval *__value) __THROW;

/* Set the timer WHICH to *NEW.  If OLD is not NULL,
   set *OLD to the old value of timer WHICH.
   Returns 0 on success, -1 on errors.  */
extern int setitimer (__itimer_which_t __which,
                      const struct itimerval *__restrict __new,
                      struct itimerval *__restrict __old) __THROW;
```

xenu (approved):

tonycoz:
> glibc when `_GNU_SOURCE` is defined, but **not** in the C++ mode, so that warning is indeed bogus

Thanks for that.

The warning produced by -Wc++-compat wouldn't be bogus here, since with normal `-Wc++-compat` we're building as C, it sees the 'passing int to enum' and rightly warns.

I'm not sure the alternate type for the `which` parameter is POSIX compliant, but it's not worth an upstream bug report.

tonycoz:
Changed to only restore diagnostics if we suppress them.

Also removed the "I wasn't able to find a platform" since xenu found one for me :)

PR #21186, opened 2023-06-30
Allow git to handle the conversion CR/LF > LF automatically

sevan:
From the gitattributes documentation:
This attribute marks the path as a text file, which enables end-of-line conversion. When text is set to "auto", Git decides by itself whether the file is text or binary. If it is text and the file was not already in Git with CRLF endings, line endings are converted on checkin and checkout as described above. Otherwise, no conversion is done on checkin or checkout.


Grinnz:
Whether the committers think this is desirable or not, it must exclude the cpan/ directory as these files are copied from CPAN.

sevan:
Added a `cpan/.gitattributes` file to ignore the auto conversion going forward.

Leont (approved):

Leont (changes requested):
The change itself looks good, but various tests need adaptation to keep working on Windows.

sevan:
Is your comment based on the testsuite failures on msvc142 and mingw64 from the github actions?

tonycoz:
I see similar failures with default line ending configuration (my default was autocrlf=input, eol=lf which had no problem).

The text=auto appears to override the `git config --global core.autocrlf false` set in the workflow, so some files are checked out with CRLF, since the core.eol default is "native".

sevan:
Not going to get to look at this and address the issues.

PR #21199, opened 2023-07-03
test.pl: Clarify watchdog() sub comments, best practice is to clear any you set

khwilliamson:
Failure to cancel a timer can lead to failing (or worse: flapping) tests


khwilliamson:
easiest for me to resubmit, as things got mixed up

PR #21201, opened 2023-07-03
Release schedule for 5.39 dev cycle

jkeenan:
Skeleton only; no assignments yet.

NOTE: The 5.38.0 release manager opted not to do a "same day" or "next day" of perl-5.39.0.  Does that mean that we start this year's dev cycle with a 5.39.0 release on July 20 --or a 5.39.1?  (Assuming the latter.)

NOTE: As of the time of creation of this branch, having no 5.39.0 release we also lack a v5.39.0 annotated tag.

This has implications for CPANtesters reports.  Task::CPAN::Reporter, for example, by default does not transmit multiple PASS reports for a given version of a given distro for a given Perl release.  If I tested WWW-Mechanize-2.17 against one of the perl-5.38.0-RC* releases, I can't transmit a new report until we're at 5.39.*.

### DIFF
--- a/Porting/release_schedule.pod
+++ b/Porting/release_schedule.pod
@@ -11,15 +11,15 @@ release schedules for the next, current and previous stable versions
 of Perl.  Dates with two or more question marks will only be releases if
 deemed necessary by the Steering Council.
 
+=head2 Perl 5.38
+
+  2023-07-01  5.38.0  ✓       Ricardo Signes
+  2023-??-??  5.38.1
+
 =head2 Perl 5.36
 
   2022-05-27  5.36.0  ✓       Ricardo Signes
   2023-04-23  5.36.1  ✓       Steve Hay
-
-=head2 Perl 5.34
-
-  2021-05-20  5.34.0  ✓       Sawyer X
-  2022-03-13  5.34.1  ✓       Steve Hay
 
 =head1 DEVELOPMENT RELEASE SCHEDULE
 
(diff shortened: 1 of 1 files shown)


Leont:
IMO we should push an annotated tag ASAP

jkeenan:
> IMO we should push an annotated tag ASAP

Which is what @rjbs has done.
```
commit aec03eb052e5786cf01d615cc5eb8e3d528c0fcd (tag: v5.39.0)
Merge: 76298ae68a e897d1e87b
Author:     Ricardo Signes <rjbs@semiotic.systems>
AuthorDate: Sun Jul 2 19:38:44 2023
Commit:     Ricardo Signes <rjbs@semiotic.systems>
CommitDate: Sun Jul 2 19:38:44 2023

    Merge branch 'release-5.39.0' into blead
```

jkeenan:
I intend to merge this pull request into blead within the next 24 hours so that we can start off the Perl and Raku Conference (NA) with the skeleton framework for perl-5.39 development in place.  The Perl Steering Council (yet to be elected) will be free to revise these dates and solicit release managers, so nothing is being written in stone yet.

However, the projected release date for perl-5.39.1 is July 20 -- a mere 10 days off.  The point-one release will be quite different from perl-5.38.0 or from the v5.39.0 tag, since a lot of work has gone into blead that was deliberately deferred from the last dev cycle.  Hence, it will embody a lot of new information and code, which would benefit from an on-time dev release.  Any volunteers?

PR #21204, opened 2023-07-04
Fix typo in pumpkin.pod

bekuno:


iabyn:
'cow orker' wasn't a typo, it's a joke. (It also appears in perlhist). Whether we want to keep the joke is a choice.

bekuno:
> 'cow orker' wasn't a typo, it's a joke. (It also appears in perlhist). Whether we want to keep the joke is a choice.

Upps.
It was in the file from the first commit.
`Perl 5 Porters authored and Chip Salzenberg committed on Feb 21, 1997.`

@khwilliamson You can revert, if you want.

khwilliamson:
I am inclined to not revert.  Maybe someone who is a real pumpkin would get it immediately; but we don't have any new ones of those coming along.  And for the remainder of us, it looks like an un-funny typo just waiting to be fixed and someone down the line might submit a P.R. and so on.

PR #21214, opened 2023-07-06
perlapi: 'resume_compcv' doesn't exist

khwilliamson:
I presume what was meant here is 'resume_compcv_final'


leonerd (approved):
Yes, I renamed it midway through and forgot to update that.

LGTM

PR #21223, opened 2023-07-09
A couple of small fixes to `pod/perlclass pod`

leonerd:


leonerd:
mingw64 is predictably unhappy, about predictably entirely unrelated things to this PR. I think we can ignore that red X.

khwilliamson (approved):

PR #21228, opened 2023-07-09
locale.c: Move setting errno out of critical section

khwilliamson:
I doesn't matter much, but its better practice to make critical sections as small as possible.


khwilliamson:
Applied via e38e5c71ea2260e88d7c7c9729f9d5893a989fbf

PR #21230, opened 2023-07-09
Add ability to emulate mingw locales on other boxes

khwilliamson:
This commit allows the ability to Configure a perl to use much of the locale handling code that MingW executes.  It is an aid to development on more POSIX-like boxes to be able to compile and test while partially simulating a MingW environment.  This can catch many bugs earlier without having to access a MingW box.  Of course this is a partial simulation, so it misses some things,

This currently fails a couple of tests in the suite; there are commits in the pipeline to fix those.

Remember this only gets compiled if explicitly requested.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -147,6 +147,95 @@ static int debug_initialization = 0;
 #define PERL_IN_LOCALE_C
 #include "perl.h"
 
+#ifdef WIN32_USE_FAKE_OLD_MINGW_LOCALES
+
+   /* Use -Accflags=-DWIN32_USE_FAKE_OLD_MINGW_LOCALES on a POSIX or *nix box
+    * to get a semblance of pretending the locale handling is that of a MingW
+    * that doesn't use UCRT (hence 'OLD' in the name).  This exercizes code
+    * paths that are not compiled on non-Windows boxes, and allows for ASAN.
+    * This is thus a way to see if locale.c on Windows is likely going to
+    * compile, without having to use a real Win32 box.  And running the test
+    * suite will verify to a large extent our logic and memory allocation
+    * handling for such boxes.  And access to ASAN and PERL_MEMLOG Of course the underlying calls are to the POSIX
+    * libc, so any differences in implementation between those and the Windows
+    * versions will not be caught by this. */
+
+#  define WIN32
+#  undef P_CS_PRECEDES
+#  undef CURRENCY_SYMBOL
+#  define CP_UTF8 -1
+#  undef _configthreadlocale
+#  define _configthreadlocale(arg) NOOP
+
+#  define MultiByteToWideChar(cp, flags, byte_string, m1, wstring, req_size) \
+                    (mbsrtowcs(wstring, &(byte_string), req_size, NULL) + 1)
+#  define WideCharToMultiByte(cp, flags, wstring, m1, byte_string,          \
+                              req_size, default_char, found_default_char)   \
+                    (wcsrtombs(byte_string, &(wstring), req_size, NULL) + 1)
+
+#  ifdef USE_LOCALE
+
+static const wchar_t * wsetlocale_buf = NULL;
+static Size_t wsetlocale_buf_size = 0;
+static PerlInterpreter * wsetlocale_buf_aTHX = NULL;
+
+STATIC
+const wchar_t *
+S_wsetlocale(const int category, const wchar_t * wlocale)
+{
+    /* Windows uses a setlocale that takes a wchar_t* locale.  Other boxes
+     * don't have this, so this Windows replacement converts the wchar_t input
+     * to plain 'char*', calls plain setlocale(), and converts the result back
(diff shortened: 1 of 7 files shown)


locale.c
```diff
@@ -147,6 +147,81 @@ static int debug_initialization = 0;
...
+
+    const wchar_t * wresult = Win_byte_string_to_wstring(CP_UTF8, byte_result);
+
+    if (! wresult) {
+        return NULL;
+    }
+
+    /* Emulate a global static memory return from wsetlocale().  This currently
+     * leaks at process end; would require changing LOCALE_TERM to fix that */
+    Size_t string_size = wcslen(wresult) + sizeof(wchar_t);
```
tonycoz: Should this be `+ 1` rather than `+ sizeof(wchar_t)` ?

As is I expect the Copy() below copies from past the end of wresult.

locale.c
```diff
@@ -147,6 +147,81 @@ static int debug_initialization = 0;
...
+#  undef CURRENCY_SYMBOL
+#  define CP_UTF8 -1
+#  undef _configthreadlocale
+#  define _configthreadlocale(arg) NOOP
+
+#  define MultiByteToWideChar(cp, flags, byte_string, m1, wstring, req_size) \
+                    (mbsrtowcs(wstring, &byte_string, req_size, NULL) + 1)
+#  define WideCharToMultiByte(cp, flags, wstring, m1, byte_string,          \
+                              req_size, default_char, found_default_char)   \
+                    (wcsrtombs(byte_string, &wstring, req_size, NULL) + 1)
```
tonycoz: () around macro arguments.

tonycoz:
I thought you had access to a Windows box?

I recently did some debugging against the WIP strawberry perl, and gdb there worked reasonably well.

gdb is included with the winlibs builds of mingw64 too.

khwilliamson:
I do have access to such a box, but not all others may.  And it allows me to use symbolic links to have a bunch of Configurations available that can be tested in parallel on a single box, including with ASAN, much more conveniently than any other method I can think of.

I could pull out the changes to embed.fnc so as to confine the changes to a few lines in perl.h and locale.c., where they would be more easily.  This is a case where the new automatic alphabetic sorting of embed.fnc is harmful, because adding this causes some definitions to move a long distance in the file.  I suppose that could be alleviated in this case by renaming the variable to WIN32_FAKE_OLD_MINGW, for example.

tonycoz (changes requested):

perl.h
```diff
@@ -44,6 +44,25 @@
...
+ * in this file (and hence throughout the code that isn't expecting it), this
+ * doesn't define that, but defines the appropriate things that would otherwise
+ * be defined later in the file.  Hence those and here must be kept in sync */
+#ifdef USE_FAKE_OLD_MINGW_LOCALES
+#  define UINT  unsigned int
+#  undef USE_THREAD_SAFE_LOCALE
+#  define NO_POSIX_2008_LOCALE
+#  undef HAS_NL_LANGINFO
+#  undef HAS_NL_LANGINFO_L
+#  undef UCRT
```
tonycoz: Should this be:
```
#  undef _UCRT
```
?

tonycoz:
You call it mingw locales, is it mingw or msvcrt?

If you use a UCRT version of mingw64, you can build perl with mingw64 with UCRT now (and that's what CI currently tests)

locale.c
khwilliamson: I don't think so.  But maybe I'm not understanding what's going on.  The extra space is for a wide NULL.  That is bigger than 1.  Since the data structure is a wchar_t, the size allocated must be a multiple of sizeof(wchar_t).

locale.c
khwilliamson: I think only the arguments that get replaced with a leading '&' would need parentheses.  I'll do that.  But please explain why other arguments need to be parenthesized

tonycoz: You're right, but it is needed for those two.

locale.c
tonycoz: wcslen() returns the number of wide characters in the string, which this code then adds a number of bytes to, if this is space for a wide `NUL`, then it should be one, since you're only adding a single wide character.

Newx() and Copy() below then correctly turn that wide character count into a byte count.

It could be you intend string_size to be a byte count, but that doesn't seem likely given Newx() and Copy() use them as wide character counts.

khwilliamson: Now I understand

khwilliamson:
It is mingw.  A future commit will change the handling when UCRT is used, not compiling some workarounds we have for bugs that UCRT fixes.  In MSVC all but one of these has been fixed since a long time; and the final one was fixed in version 1900.  (Note there are still bugs in both that we still use workarounds for.)

perl.h
khwilliamson: yes, now changed

khwilliamson:
requested changes completed

PR #21232, opened 2023-07-10
regen/HeaderParser.pm: require perl >= 5.10

khwilliamson:
This won't compile when run on earlier perls, and the reason isn't apparent without some digging.  It turns out that it uses named capture groups which came in 5.9.5..

### DIFF
--- a/regen/HeaderParser.pm
+++ b/regen/HeaderParser.pm
@@ -1,6 +1,7 @@
 package HeaderParser;
 use strict;
 use warnings;
+use v5.10;      # Needs named capture groups
 
 # these are required below in BEGIN statements, we cant have a
 # hard dependency on them as they might not be available when


Leont:
Why `BEGIN { require ` when we can just use `use`?

regen/HeaderParser.pm
```diff
@@ -1,6 +1,7 @@
 package HeaderParser;
 use strict;
 use warnings;
+BEGIN { require v5.9.5; }
```
ilmari: Why not just `use v5.10;` since that's the stable version that introduced named captures?

jkeenan:
> This won't compile when run on earlier perls, and the reason isn't apparent without some digging. It turns out that it uses named capture groups which came in 5.9.5..

AFAICT files under `regen/` are core-only, not CPAN.  Unless I'm missing something -- why should we care?

demerphq:
> AFAICT files under regen/ are core-only, not CPAN. Unless I'm missing something -- why should we care?

Because regen is supposed to run with an installed perl, and that might be quite old. We discussed how far back we support at one point, and it is much more recent than 5.10, but at the same time, if someone uses a sufficiently old enough perl we should give them a useful error message.

regen/HeaderParser.pm
demerphq: I agree with @ilmari.

jkeenan:
> > AFAICT files under regen/ are core-only, not CPAN. Unless I'm missing something -- why should we care?
> 
> Because regen is supposed to run with an installed perl, and that might be quite old. 

Understood; thanks for the clarification.

khwilliamson:
I ran into this on a real box where the installed perl was old enoughSent from my iPhoneOn Jul 10, 2023, at 2:47 PM, James E Keenan ***@***.***> wrote:﻿


AFAICT files under regen/ are core-only, not CPAN. Unless I'm missing something -- why should we care?

Because regen is supposed to run with an installed perl, and that might be quite old.

Understood; thanks for the clarification.
We discussed how far back we support at one point, and it is much more recent than 5.10, but at the same time, if someone uses a sufficiently old enough perl we should give them a useful error message.

demerphq:
@khwilliamson were you going to change this to use 'use v5.10' ?

demerphq:
I have pushed https://github.com/Perl/perl5/pull/21242 as a replacement for this PR with a more friendly error message if the perl is too old.

PR #21234, opened 2023-07-11
allow mingw64 ci to succeed

### LINKED ISSUE #21202: dist/ExtUtils-ParseXS/t/002-more.t failing on Win32 (mingw64)
GH Actions tests on "Windows mingw64" are failing on multiple pull requests on different branches.  See, *e.g.,*

https://github.com/Perl/perl5/actions/runs/5447805747/jobs/9910433855?pr=21200

https://github.com/Perl/perl5/actions/runs/5448721934/jobs/9912356883?pr=21201

https://github.com/Perl/perl5/actions/runs/5447537853/jobs/9909828212?pr=21199

Sample failures:
```
...
../dist/ExtUtils-ParseXS/t/001-basic.t ............................... ok

#   Failed test 'No error message recorded, as expected'
#   at t/002-more.t line 65.
#          got: 'Can't load 'D:\a\perl5\perl5\dist\ExtUtils-ParseXS\t/XSMore.dll' for module XSMore: load_file:Invalid access to memory location at D:\a\perl5\perl5\lib/DynaLoader.pm line 206.
../dist/ExtUtils-ParseXS/t/002-more.t ................................ 
# '
Failed 26/33 subtests 
#     expected: ''

#   Failed test 'the BOOT keyword'
#   at t/002-more.t line 68.
#          got: undef
#     expected: '100'
Undefined subroutine &XSMore::include_ok called at t/002-more.t line 70.
# Looks like your test exited with 2 just after 9.
../dist/ExtUtils-ParseXS/t/003-usage.t ............................... ok
...

Test Summary Report
-------------------
../dist/ExtUtils-ParseXS/t/002-more.t                              (Wstat: 512 (exited 2) Tests: 9 Failed: 2)
  Failed tests:  7, 9
  Non-zero exit status: 2
  Parse errors: Bad plan.  You planned 33 tests but ran 9.
../dist/Net-Ping/t/450_service.t                                   (Wstat: 0 Tests: 26 Failed: 0)
  TODO passed:   9
../ext/IPC-Open3/t/IPC-Open3.t                                     (Wstat: 0 Tests: 45 Failed: 0)
  TODO passed:   25
Files=2767, Tests=1128665, 1839 wallclock secs (103.00 usr +  8.31 sys = 111.31 CPU)
Result: FAIL
Finished test run at Mon Jul  3 19:19:37 2023.
gmake: *** [GNUMakefile:1840: test] Error 2
Error: Process completed with exit code 2
```
I have no access to Windows, so I can't diagnose further.

tonycoz:
mingw64 has been failing due to and apparent compiler bug, where it was generating what appears to be an invalid DLL, which failed to load.

This series of commits does two basic things:
- allows us to build with ucrt versions of mingw64 rather than just msvcrt (which is still supported)
- in CI, install a fresh mingw64 and test/build with that instead of the gcc bundled with strawberry perl.

This still uses the gmake from strawberry perl.

I expect some existing code that currently treats non-MSVC WIN32 builds can be updated to now check for PERL_WIN32_UCRT, but I wanted to get CI fixed first.

Fixes #21202

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -521,27 +521,18 @@ jobs:
     steps:
       - run: git config --global core.autocrlf false
       - uses: actions/checkout@v3
-      - name: Set up Perl build environment
-        run: |
-          # skip installing perl if it is already installed.
-          if (!(Test-Path "C:\strawberry\perl\bin")) {
-            choco install strawberryperl
-          }
-          echo @"
-          C:\strawberry\c\bin
-          C:\strawberry\perl\site\bin
-          C:\strawberry\perl\bin
-          "@ |
-            Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Install mingw-64
+        uses: egor-tensin/setup-mingw@v2
       - name: Host perl -V
-        run: perl -V
+        run: |
+          perl -V
       - name: gcc --version
         run: gcc --version
       - name: Build
         shell: cmd
         run: |
           cd win32
-          gmake CCHOME=C:\strawberry\c CFG=Debug -f GNUMakefile -j2
+          gmake CCHOME=C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64 CFG=Debug -f GNUMakefile -j2
       - name: Show Config
         shell: cmd
         run: |
@@ -552,7 +543,7 @@ jobs:
         run: |
           cd win32
           set HARNESS_OPTIONS=j2
-          gmake CCHOME=C:\strawberry\c CFG=Debug -f GNUMakefile test
+          gmake CCHOME=C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64 CFG=Debug -f GNUMakefile test
 
   #                            _
   #   ___ _   _  __ ___      _(_)_ __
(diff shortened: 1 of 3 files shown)


xenu:
> The installation of mingw64 is slow, hopefully a future change will let us use the github workflow caching mechanism.

It seems this github action downloads and extracts a 7z archive. LZMA is a very slow algorithm. I wonder if it would be faster to manually download and extract winlibs zip binaries. Windows comes with a tar binary that can handle zip archives (and it's much faster than e.g. PowerShell's Expand-Archive).

win32/GNUmakefile
```diff
@@ -487,6 +487,15 @@ GCCVER2		:= $(shell for /f "delims=. tokens=1,2,3" %%i in ('$(GCCBIN) -dumpversi
...
 
+# are we using UCRT or MSVCRT?
+ifeq ($(CCTYPE),GCC)
+CRT_TYPE := $(shell gcc --version 2>&1 | find "msvcrt" >nul && ( echo msvcrt) || ( echo ucrt) )
+else ifeq ($(CCTYPE),MSVC120)
+CRT_TYPE := msvcrt
+else
+CRT_TYPE := ucrt
+endif
+
```
xenu: You can check for `_UCRT` define instead. Both Visual C++ and mingw define it.

tonycoz: Thanks, my google-fu failed me when I tried finding such a macro.

demerphq:
@xenu said:
> LZMA is a very slow algorithm. I wonder if it would be faster

IMO the target should be correctness first, and only when correctness is achieved should we worry about performance.  Right now every push we have a failure from mingw64, which is very distracting for people who are not working on anything related to mingw64.  So I am going to apply this as is, and we can address making it faster later on.

demerphq (approved):

PR #21236, opened 2023-07-11
Remove microperl

ilmari:
It hasn't built in ages, and nobody has bothered fixing it, even after the threat of removal in perl5180delta and @nwc10's repeated threat in 2021, so let's make good on those.


jkeenan:
All tests pass for me on FreeBSD and Linux. +1

tonycoz (approved):

PR #21241, opened 2023-07-16
Synchronize struct regexp information in perlreapi.pod and regexp.h 

demerphq:
[perlreapi.pod](https://github.com/Perl/perl5/blob/0c77b7235984453280b85b181aecb8e757ee8480/pod/perlreapi.pod?plain=1#L13) has a copy of the struct regexp from [regexp.h](https://github.com/Perl/perl5/blob/0c77b7235984453280b85b181aecb8e757ee8480/regexp.h#L133), but over time the two copies have fallen out of sync with each other. When @nwc10 converted regexps to be first order SV types there were many changes that invalidated old structure members (typically replaced with RX_ macros), and there have been various other changes to the structure which were not captured in the docs. 

This PR includes two commits. The first modifies t/porting/podcheck.t to support validating that a verbatim block can be found in the file it was copied from. This can be enabled by including the text "copied from: FILENAME" or "source file: FILENAME" on the first line of the verbatim text where the FILENAME actually exists on disk. This text can be commented however you like, or uncommented, whatever works for you, but must be on one line. If this special annotation is found then the podcheck logic will convert the verbatim block into a pattern (whitespace insensitve) and check to see if the named file contains the pattern. If it does not then it will throw an error. Thus we should notice if the docs and source get out of sync with each other. In this mode the normal verbatim text line length checks are *disabled*, as we want to be able to take the original source, copy it and then indent it into the pod file.

The second commit makes some minor changes to the structure in regexp.h, and copies the result into perlreapi.pod, following the convention established above. It also includes additional documentation on a few of the columns that were missing previously, and I also fixed some of the docs that were out of date, and added some that are new as of 5.38. For instance the old "precomp" value is now actually the pv value of the regexp, so the documentation for "precomp" is not documented as the macro RX_PRECOMP. 

Thanks to @tonycoz for noticing this issue and calling my attention to it.

### DIFF
--- a/pod/perlreapi.pod
+++ b/pod/perlreapi.pod
@@ -620,6 +620,7 @@ The REGEXP struct is defined in F<regexp.h>.
 All regex engines must be able to
 correctly build such a structure in their L</comp> routine.
 
+=for apidoc Ayh||struct regexp
 =for apidoc Ayh||REGEXP
 
 The REGEXP structure contains all the data that Perl needs to be aware of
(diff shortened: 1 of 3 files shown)


t/porting/podcheck.t
```diff
@@ -1125,19 +1125,68 @@ package My::Pod::Checker {      # Extend Pod::Checker
...
+            my $file = $1;
+            $copied = 1;
+            my $pat = "";
+            foreach my $line (@lines[1..$#lines]) {
+                # convert each line into a pattern fragment
+                $line =~ s/(\s+)|(\S+)/$1 ? "\\s+" : quotemeta($2)/ge;
+                $pat .= $line;
+            }
+            # merge adjacent \s+ sequences
+            1 while $pat =~ s/\\s\+\\s\+/\\s+/;
```
tonycoz: why not:
```
$pat =~ s/(?:\\s\+){2,}/\\s+/g;
```
?

pod/perlreapi.pod
```diff
@@ -791,21 +859,24 @@ C<< ->offs[0].start/end >> represents C<$&> (or
...
+checks if the last compiled C<REGEXP>'s C<RX_PRECOMP> and C<RX_PRELEN>
 are equivalent to the new one, and if so uses the old pattern instead
 of compiling a new one.
 
 The relevant snippet from C<Perl_pp_regcomp>:
 
-	if (!re || !re->precomp || re->prelen != (I32)len ||
-	    memNE(re->precomp, t, len))
+	if (!re || !RX_PRECOMP(re) || RX_PRELEN(re) != (I32)len ||
+	    memNE(RX_PRECOMP(re), t, len))
```
tonycoz: I don't see that code in pp_regcomp.

I suspect this logic in re_op_compile replaces it:
```
    if (   old_re
        && !recompile
        && cBOOL(RX_UTF8(old_re)) == cBOOL(RExC_utf8)
        && ( RX_COMPFLAGS(old_re) == ( orig_rx_flags & RXf_PMf_FLAGCOPYMASK ) )
        && RX_PRELEN(old_re) == plen
        && memEQ(RX_PRECOMP(old_re), exp, plen)
        && !runtime_code /* with runtime code, always recompile */ )
```

regexp.h
```diff
@@ -170,10 +181,9 @@ typedef struct regexp {
      */
 
     U32 lastparen;           /* highest close paren matched ($+) */
+    U32 lastcloseparen;      /* last close paren matched ($^N) */
     regexp_paren_pair *offs; /* Array of offsets for (@-) and (@+) */
     char **recurse_locinput; /* used to detect infinite recursion, XXX: move to internal */
-    U32 lastcloseparen;      /* last close paren matched ($^N) */
-
```
tonycoz, line 188: This makes struct regexp 8 bytes larger on 8 byte alignment platforms.

Whether the smaller size is worth the loss of readability is a matter of taste, you can ignore this comment if you wish.

t/porting/podcheck.t
demerphq: No good reason, I'll change it. Thanks!

pod/perlreapi.pod
demerphq: I guess I'll just remove the expression, not sure if it's really that helpful.

regexp.h
demerphq: Oh, I think I miscounted then. I'll revert that change, I had the opposite intent. Good catch.

tonycoz: It doesn't require much counting:
- `void *pprivate;` is 8-byte aligned,
- the `U32 intflags;` and `U32 lastparen;` take us to the next 8 byte boundary
- so adding another U32 here means we need padding before `offs`
- we have only the two bitfields after `recurse_input` which should combine to a single U32
- so without `lastcloseparen` we need another 4 bytes of padding to 8 byte align `subbeg`.

The change you've made instead (moving `extflags` as well as `lastcloseparan` looks good.

tonycoz (approved):

regexp.h
demerphq: Agreed. I just wondered if you had some "trick" to make this easier. I actually started wondering if we should have a test which shows the size of the structs.

tonycoz: A little research found `pahole`:
```
$ pahole -C regexp ./perl
struct regexp {
        HV *                       xmg_stash;            /*     0     8 */
        union _xmgu                xmg_u;                /*     8     8 */
        STRLEN                     xpv_cur;              /*    16     8 */
...
        /* size: 224, cachelines: 4, members: 32 */
        /* sum members: 216, holes: 1, sum holes: 4 */
        /* sum bitfield members: 13 bits, bit holes: 1, sum bit holes: 19 bits */
        /* last cacheline: 32 bytes */
};
$ pahole -P ./perl
pmop    96      88      8
custom_op       40      32      8
block_sub       48      40      8
RExC_state_t    416     408     8
_IO_FILE        216     208     8
```
I think this is the tool Richard used

richardleach: Yeah, that's what I've used previously. Was only manually eyeballing though, I do not have any code around it.

PR #21246, opened 2023-07-17
utf8_hop_forward: Don't go over edge of buffer

khwilliamson:
even in the presence of malformed UTF-8.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -2659,7 +2659,7 @@ start of the next character.
 
 C<off> must be non-negative.
 
-C<s> must be before or equal to C<end>.
+C<s> must be before or equal to C<end>.  If after, the function panics.
 
 When moving forward it will not move beyond C<end>.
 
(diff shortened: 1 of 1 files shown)


tonycoz:
"edige" in commit subject

tonycoz:
Approved once the commit subject is fixed.

jkeenan:
> even in the presence of malformed UTF-8.

Can we add something to the test suite to demonstrate the change in behavior?

demerphq:
@jkeenan yes we can. I assume what you really are asking is for @khwilliamson to do so. :-)

jkeenan:
> @jkeenan yes we can. I assume what you really are asking is for @khwilliamson to do so. :-)

Yes.

khwilliamson:
I don't see a way to add a test here.  This protects against running off the end in two situations

1. If the memory at the end of the string happens to match continuation bytes.  This means the test would have to populate the memory, and somehow free it, but still point to just before it.
2. If the input parameters are such that the start is  after the end.  But this is asserted against in DEBUGGING builds.  We don't generally test for just non-debugging cases

tonycoz:
> I don't see a way to add a test here.

You could add an interface to XS::APItest that takes a SV buffer and initial and end offsets, and returns the final offset, and test that with various strings and offsets.

The commit message is still "edige"y.

khwilliamson:
There already are such tests in APItest.xs

inline.h
```diff
@@ -2677,19 +2676,24 @@ Perl_utf8_hop_forward(const U8 *s, SSize_t off, const U8 *end)
...
 
-    assert(s <= end);
     assert(off >= 0);
 
+    if (UNLIKELY(s >= end)) {
+        GCC_DIAG_IGNORE(-Wcast-qual)
+        return (U8 *)end;
+        GCC_DIAG_RESTORE
+    }
+
```
tonycoz: Shouldn't this just `assert(s <= end)`?

khwilliamson: My thinking on this was that where there is a possibility of accessing outside the buffer that the guards should be always enabled, and not just under DEBUGGING.  Also, it makes it behave sanely without crashing when called with a bad input value

tonycoz: We could panic in both DEBUGGING and non-DEBUGGING builds, which is something we do in a variety of places.

tonycoz:
> There already are such tests in APItest.xs

But none of the tests (I assumed in `ext/XS-APItest/t/utf8.t:1186`) were failing, and this doesn't introduce any new tests.

khwilliamson:
I couldn't figure out any reasonable new test that fails before the fix.  That's why this bug is unlikely to be a problem in the wild.   It requires the byte(s) immediately after the string in memory to be one of 32 bit patterns, 0x80..0xBF.  The best I could come up with was to end a string with a bunch of continuation bytes; then turn on the utf8 flag and set SvCUR to a smaller value, and then test that the hop didn't go past the new artificially lower limit.  But that seemed overly artificial.

PR #21247, opened 2023-07-17
utf8_hop_back: Check before derefencing

khwilliamson:
This conditional derefenced before checking if it is safe to dereference


tonycoz (approved):
I don't think there's any safety difference since s must be >= start at entry, and nothing we do will make s < start, but it may save an unnecessary memory access when `s == start`

khwilliamson:
Agreed

PR #21256, opened 2023-07-17
Batch of revisions to locale.c

khwilliamson:
This is a batch of commits to locale.c that together fix bugs and improve efficiency and code readability.  A significant set of tests will be added at the conclusion of all the batches


locale.c
```diff
@@ -1345,7 +1345,18 @@ S_emulate_setlocale_i(pTHX_
              line, new_obj));
 
 #ifdef MULTIPLICITY
+
+    if (DEBUG_Lv_TEST) {
+        if (PL_cur_locale_obj != new_obj) {
+            PerlIO_printf(Perl_debug_log,
+                          "(%" LINE_Tf "): PL_cur_locale_obj"
+                          " was %p, now is %p\n",
+                          line, PL_cur_locale_obj, new_obj);
```
tonycoz, line 1411: I don't think `locale_t` is required to be a pointer type.

I've probably missed this elsewhere.  I think it can be assumed to be a scalar type (pointer or arithmetic type), since you can cast `0` to `locale_t`.

khwilliamson: For example, in the man page
``locale_t newlocale(int category_mask, const char *locale,
                           locale_t base);``
it says

> The  newlocale()  function  creates a new locale object, or modifies an
        existing object, returning a reference to the new or modified object as
        the function result.

Am I wrong to think that one can always print a reference with `%p`?

khwilliamson: Actually POSIX 2017, says 

> Upon successful completion, the newlocale() function shall return a handle

so I suppose that doesn't have to be a pointer; I've just never seen an implementation that isn't.  There are plenty of `%p` instances in blead for printing locale objects.
I see three possibilities:

1.  Continue using `%p` until and if it becomes a problem some day.
2. Add a Configure probe and a `LOCALE_Tf` define and convert to use it now
3. Just add the `LOCALE_Tf` define and set it to `%p`, convert to use it, and add a Configure probe if and when it becomes needed

tonycoz: I went [by](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/locale.h.html):
```
The <locale.h> header shall define the locale_t type, representing a locale object.
```
which doesn't place any constraints on the type of `locale_t`, and being able to cast `0` to to `locale_t`.

I think 1 is fine.

Technically the argument of `%p` is required to be pointer to void (rather than any pointer) but I expect we break this rule generally.

locale.c
```diff
@@ -251,225 +251,282 @@ static const char C_thousands_sep[] = "";
...
-/* A few categories require additional setup when they are changed.  This table
- * points to the functions that do that setup */
-STATIC void (*update_functions[]) (pTHX_ const char *, bool force) = {
 #  ifdef USE_LOCALE_CTYPE
-                                S_new_ctype,
+#    define LC_CTYPE_ENTRY_                      LC_CTYPE,
+#    define LC_CTYPE_MASK_                       LC_CTYPE_MASK,
+#    define LC_CTYPE_STRING_                    "LC_CTYPE",
+#      define LC_CTYPE_UPDATE_          S_new_ctype,
+#  else
```
tonycoz: Would this series of commits be more maintainable if it worked something like `intrpvar.h`?

It would significantly reduce the repetition: how many times is `LC_CTYPE` repeated here, including as substrings of macro names?

For example locale_table.h could be:
```
#ifdef USE_LOCALE_CTYPE
  LOCALE_TABLE_ENTRY(LC_CTYPE, LC_CTYPE_MASK, "LC_CTYPE", S_new_ctype)
#endif
#ifdef USE_LOCALE_NUMERIC
  LOCALE_TABLE_ENTRY(LC_NUMERIC, LC_NUMERIC_MASK, "LC_NUMERIC", S_new_numeric)
#endif
... and more
```
Then in locale.c:
```
#undef LOCALE_TABLE_ENTRY
#define LOCALE_TABLE_ENTRY(index, mask, name, cb) index,
STATIC const int categories[] = {
#include "locale_table.h"
  FAKE_LCALL+1
}
#undef LOCALE_TABLE_ENTRY
#defined LOCALE_TABLE_ENTRY(index, mask, name, cb) name,
STATIC const char * const category_names[] = {
#include "locale_table.h"
   "Locale category unknown to Perl; ..."
}
... etc
```
though the mask and name could be synthesized from the index parameter.

khwilliamson: This sounds like a good idea.  I'll work on it.

tonycoz (approved):

PR #21267, opened 2023-07-19
perldelta.pod - document changes to UNIVERSAL.pm

demerphq:
We now define a UNIVERSAL::import and UNIVERSAL::unimport this explains why and what the consequences. Also tweaks the existing perldiag documentation of it.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -64,6 +64,40 @@ as with C<m?pattern?>.
 
 [GH #20763]
 
+=head2 Calling the import method of an unknown package produces an error
+
+Historically it has been possible to call the import() or unimport()
+method of any class, including ones which have not been defined, with
+an argument and not experience an error. For instance this code will
+not throw an error in 5.38:
+
+    Class::That::Does::Not::Exist->import("foo");
+
+however, as of 5.39.1 this will throw an exception.  Note that calling
+these methods with no arguments continues to silently succeed and do
+nothing. For instance
+
+    Class::That::Does::Not::Exist->import();
+
+will continue to not throw an error. This is because every class
+implicitly inherits from the class UNIVERSAL which now defines an import
+method. In older perls there was no such method defined, and instead the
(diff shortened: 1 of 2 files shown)


pod/perldelta.pod
```diff
@@ -64,6 +64,43 @@ as with C<m?pattern?>.
...
+platforms with case-insensitive file systems on older perls the
+following code
+
+    use STRICT 'refs';
+
+would silently do nothing as the module is actually called 'strict.pm'
+not 'STRICT.pm', so it would be loaded, but its import method would
+never be called.
+
+This change will also detect cases where people attempt to import subs
```
mauke: I would `s/subs/symbols/g` in this paragraph because subs are not the only thing that can be exported.

pod/perldelta.pod
```diff
@@ -64,6 +64,43 @@ as with C<m?pattern?>.
...
+    Class::That::Does::Not::Exist->import();
+
+will continue to not throw an error. This is because every class
+implicitly inherits from the class UNIVERSAL which now defines an import
+method. In older perls there was no such method defined, and instead the
+method calls for C<import> and C<unimport> were special cased to not
+throw errors if there was no such method defined.
+
+This change has been added because it makes it easier to detect case
+typos in C<require> and C<use> statements when running on
```
mauke: Why mention `require` here? AFAIK it does not interact with `import`.

demerphq: Ah, good catch.

pod/perldelta.pod
demerphq: True. I'll fix.

pod/perldelta.pod
demerphq: Removed the mention of require in the latest push.

pod/perldelta.pod
demerphq: I completely removed and replaced this.

leonerd (approved):
LGTM

PR #21273, opened 2023-07-20
Use strict and warnings in B::Deparse tests

ilmari:
This would have caught the duplicate (and it turns out, useless) definition of $expr in do_infix_keyword().

### DIFF
--- a/lib/B/Deparse-core.t
+++ b/lib/B/Deparse-core.t
@@ -34,6 +34,7 @@ BEGIN {
     }
 }
 
+use warnings;
 use strict;
 use Test::More;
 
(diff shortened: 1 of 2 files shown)


jkeenan:
> This would have caught the duplicate (and it turns out, useless) definition of $expr in do_infix_keyword().

The way you have phrased that suggests that this has been previously reported.  Do you have a link for that (or can you provide an illustration of the problem)?

Thank you very much.

ilmari:
> > This would have caught the duplicate (and it turns out, useless) definition of $expr in do_infix_keyword().
> 
> The way you have phrased that suggests that this has been previously reported. Do you have a link for that (or can you provide an illustration of the problem)?

If warnings had been in effect, the second `my $expr = …` [here](https://github.com/Perl/perl5/pull/21273/files#diff-7e7aad4c9946b0e9a31879f943c58b4cda0407c256f76c1addc365028218767eL135-L137) would have issued a `"my" variable $expr masks earlier declaration in same scope` warning.

I noticed this while looking at the file for other reasons, because Emacs' `flycheck-mode` uses `perl -wc $file` to show errors and warnings inline.  Then I looked closer at it and realised the `$expr` variable wasn't used at all, so the could _both_ go.

Issue #21272, opened 2023-07-20
t/porting/authors.t fails in symlink directory


yangmingtai:
Module:

**Description**
During the construction of perl-5.38.0, there was a test case failure

**Steps to Reproduce**
```
#!/bin/sh
# Assume a git checkout is in /tmp/perl-git
mkdir build
cd build
sh /tmp/perl-git/Configure -Dmksymlinks -des
make 
make test
cd t
./perl harness -v porting/authors.t 
exit

```
![image](https://github.com/Perl/perl5/assets/17870505/c2365a18-b03c-4a49-aaf3-908f745d6bc5)

Others have the same problem
https://www.nntp.perl.org/group/perl.perl5.porters/2023/06/msg266520.html

jkeenan:
A "first draft" p.r. for this problem is [here](https://github.com/Perl/perl5/pull/21274).

However, if the people who worked on `Porting/updateAUTHORS.p[lm]` want to provide something else, that would be okay.

demerphq:
> want to provide something else, that would be okay.

We could also check if .git is present if we dont already, but your patch should work fine.

PR #21274, opened 2023-07-20
Skip this test when .mailmap is not visible

jkeenan:
... which appears to be the case when you configure with -Dmksymlinks.

For: https://github.com/Perl/perl5/issues/21272 and https://www.nntp.perl.org/group/perl.perl5.porters/2023/06/msg266520.html

@doughera88

### DIFF
--- a/t/porting/authors.t
+++ b/t/porting/authors.t
@@ -19,6 +19,10 @@ skip_all(
     "This is a shallow clone, this test requires history.")
   if (-e "$source_dir/.git/shallow");
 
+skip_all(
+    "Cannot locate .mailmap: GH#21272.")
+  unless (-f ".mailmap");
+
 my $revision_range = ''; # could use 'v5.22.0..' as default, no reason to recheck all previous commits...
 if ( $ENV{TRAVIS} && defined $ENV{TRAVIS_COMMIT_RANGE} ) {
 	# travisci is adding a merge commit when smoking a pull request


demerphq (approved):

t/porting/authors.t
```diff
@@ -19,6 +19,10 @@ skip_all(
     "This is a shallow clone, this test requires history.")
   if (-e "$source_dir/.git/shallow");
 
+skip_all(
+    "Cannot locate .mailmap: GH#21272.")
+  unless (-f ".mailmap");
+
```
ilmari, line 25: 1. Please don't just reference the GH# in the message, explain the reason directly
2. Should this (and whatever actually uses `.mailmap`) not be looking in `$source_dir`, which has made the effort to follow CI environment variables and symlinks?

Leont: Agreed, this can be fixed instead of skipped

jkeenan: > Agreed, this can be fixed instead of skipped

Okay, but since I was skeptical of the work put into Porting/updateAUTHORS.p[lm] in the last dev cycle, someone other than me should provide the fix.  Have at it!

demerphq: I'll pick it up.

demerphq: @Leont fwiw, I don't really agree. The authors.t is an authors test, it only needs to run when someone is using a git repo, and is doing a "standard" build in the git tree. We can make it work with -Dmksymlinks, but IMO that isn't a normal dev mode where someone would be expected to create patches from.  I am not sure if supporting this test really makes sense. I do have an alternative patch, but honestly I think @jkeenan's patch was fine. This test is only interesting when someone is actually writing patches.

demerphq:
This is replaced by https://github.com/Perl/perl5/pull/21276

doughera88:
On Thu, Jul 20, 2023 at 11:19:34AM -0700, Yves Orton wrote:

> > +skip_all(
> +    "Cannot locate .mailmap: GH#21272.")
> +  unless (-f ".mailmap");
> +
> 
> @Leont fwiw, I don't really agree. The authors.t is an authors test, it only needs to run when someone is using a git repo, and is doing a "standard" build in the git tree. We can make it work with -Dmksymlinks, but IMO that isn't a normal dev mode where someone would be expected to create patches from.  I am not sure if supporting this test really makes sense. I do have an alternative patch, but honestly I think @jkeenan's patch was fine. This test is only interesting when someone is actually writing patches.

I agree it's not worth a lot of effort to make it work in this situation.  I had looked 
for an easy way to pass the source .mailmap file along, but didn't find one.
A more complex approach is unlikely to be worth the effort.

I would have reported this directly on github, but I don't have a working github ID at the moment.

-- 
    Andy Dougherty              ***@***.***

demerphq:
Ah, its ok, i already fixed it with yves/fix_author_t_under_mksymlinks_builds aka https://github.com/Perl/perl5/pull/21276

demerphq:
I just merged https://github.com/Perl/perl5/pull/21276 so I am closing this as redundant.

PR #21280, opened 2023-07-22
perlapi: Document PERL_STATIC_INLINE

khwilliamson:


perl.h
```diff
@@ -179,7 +179,14 @@ Otherwise ends a section of code already begun by a C<L</START_EXTERN_C>>.
 /* Fallback definitions in case we don't have definitions from config.h.
    This should only matter for systems that don't use Configure and
    haven't been modified to define PERL_STATIC_INLINE yet.
+
+=for apidoc CmU|void|PERL_STATIC_INLINE|token
```
tonycoz: I don't think this should be "C" core only.

The name as used is prefixed and has to be visible to other code for `inline.h` to work, so I think it's safe for it to be documented as API.

tonycoz (approved):

tonycoz:
I approved it, but then noticed the test failure, which appears to be connected to the change.

khwilliamson:
I submitted this as a result of a failure in regenerating Devel::PPPort.  It turns out that autodoc.pl generates an entry from the description in config.h (and it is better than the pod this PR would have created).  And that is where the duplicate came from.  Somehow D:P isn't getting that, so I will have to fix it there.  And autodoc should have failed the earlier version of this PR due to an entry in perlintern with the same name as one in perlapi.  I will fix that too

Issue #21275, opened 2023-07-20
If the 'scalar' IO layer cant be loaded by the IO system Perl will create files on disk named after a reference.


demerphq:
Module:
PerlIO-scalar 

**Description**
If for some reason the PerlIO logic can't load the 'scalar' IO layer, then Perl will happily create files named after a reference on disk.

While working on https://github.com/Perl/perl5/pull/21274 and https://github.com/Perl/perl5/issues/21272 I discovered this.

**Steps to Reproduce**
This is a bit tricky. Configure a perl with -Dmksymlinks, do not install it, then write a script that does something like this:

```
chdir $out_of_the_symlink_tree;
open my $fh, ">", \$scalar or die "Failed to write to scalar buffer:$!";
print $fh "this is a test";
close $fh
```
where $out_of_the_symlink_tree is a directory not in the symlink tree created by -Dmksymlinks, for instance tmp. Run it from the symlink directory. It should produce an error like this:

```
Unknown PerlIO layer "scalar" at ...
```
and then it will leave a file in the directory like this: `SCALAR(0x559c23463250)`

**Expected behavior**
It should die during the open if the layer cant be found I think, or this Just Shouldn't Happen. Why do we load the scalar layer late?

**Perl configuration**
```
./perl -Ilib -V
Summary of my perl5 (revision 5 version 39 subversion 0) configuration:
  Local Commit: 286a358bf243d37f5e09f9ae3eba6c2e0239c65c
  Ancestor: ccfa6b533228ad41e4dbc5576cd43081b8fd2a13
  Platform:
    osname=linux
    osvers=5.15.0-76-generic
    archname=x86_64-linux-thread-multi
    uname='linux oncidium 5.15.0-76-generic #83~20.04.1-ubuntu smp wed jun 21 20:23:31 utc 2023 x86_64 x86_64 x86_64 gnulinux '
    config_args='-Dmksymlinks -des -Dusethreads -Doptimize=-g3 -ggdb3 -d -Dusedevel -Dcc=ccache gcc -Dld=gcc -DDEBUGGING'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='gcc'
[...]

PR #21282, opened 2023-07-22
Move PerlIO::scalar into core

Leont:
This moves PerlIO::scalar into core. This prevents unpredictable behavior when the module can't be loaded (e.g. #21275), and also allows miniperl to run tests that need :scalar handles.


demerphq:
Nice. The commit message could be cleaned up tho. "prevents" is duplicated and "make run tests" doesn't make sense.

demerphq (approved):

demerphq (approved):

haarg:
There are a number of CPAN modules that explicitly load PerlIO::scalar. It may be good to leave a stub module.

Leont:
> There are a number of CPAN modules that explicitly load PerlIO::scalar. It may be good to leave a stub module.

Fair, I've reintroduced a stub.

Grinnz:
I can't tell from the tests, but does `open my $fh, '<:scalar', \$x` still work with this change? Not that I think it was commonly used explicitly

Leont:
> I can't tell from the tests, but does open my $fh, '<:scalar', \$x still work with this change? Not that I think it was commonly used explicitly

Yes it does. The only difference is that :scalar is now already included in the main binary, same way `:unix`/`:perlio`/`:crlf`/`:raw`/etc… have always been.

PR #21284, opened 2023-07-24
make sure "use VERSION;" sets the flag needed for "used only once"

### LINKED ISSUE #21271: [doc] use v5.36, only partially enables warnings
perl5360delta says:

`Furthermore, use v5.36 will also enable warnings as if you'd written use warnings.`

but the 'once' warning is an exception:

```
perl -e'use v5.36; no strict; print $i' 
Use of uninitialized value $i in print at -e line 1.
```

vs:

```
perl -e'use v5.36; no strict; use warnings; print $i'
Name "main::i" used only once: possible typo at -e line 1.
Use of uninitialized value $i in print at -e line 1.
```

tonycoz:
There's two parts to producing the "used only once" warning:

1. In a lexical scope with WARN_ONCE enabled, new GVs are created with the GVf_MULTI flag off, a second mention of such a name will set that flag.

2. After compilation, if G_WARN_ONCE is set in PL_dowarn, the entire package tree is walked to report names which don't have GVf_MULTI set.

In this case G_WARN_ONCE wasn't being set, so the second part didn't happen.

This flag is global, so using any other module that happened to enable the WARN_ONCE flag (anything that does C<use warnings;>) would allow warnings to be produced after compilation.

Fixes #21271

### DIFF
--- a/op.c
+++ b/op.c
@@ -7907,8 +7907,10 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
             if (!(PL_hints & HINT_EXPLICIT_STRICT_VARS))
                 PL_hints |= HINT_STRICT_VARS;
 
-            if (shortver >= SHORTVER(5, 35))
+            if (shortver >= SHORTVER(5, 35)) {
                 free_and_set_cop_warnings(&PL_compiling, pWARN_ALL);
+                PL_dowarn |= G_WARN_ONCE;
+            }
         }
         /* otherwise they are off */
         else {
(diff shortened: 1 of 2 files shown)


leonerd (approved):
LGTM

I wonder if we want to put this into a 5.38.1 and a 5.36.2 in the near future? Otherwise, people will be writing code that happens not to provoke `once` warnings accidentally.

Leont (approved):

tonycoz:
I've added this to both votes-5,36.xml and votes-5.38.xml in maint-votes.

Vote early, vote often.

PR #21293, opened 2023-07-25
universal.c - change import with args error to be a deprecated warning

demerphq:
We have had a LOT of fallout from 2dcf3cf50d7 as it turns out that there various common cases of people accidentally exploiting the silent use of missing import with arguments.  The premise of detecting this was that it would allow us to detect scenarios where users on a case-insensitive file-system had a typo. But it turns that there are a lot of accidental cases of this as well, a lot more than we expected.

Common mistaken cases are:

    * Via use_ok() where people think the second argument is a test
      name instead of an argument to the import of the module being
      used.

    * Using a quoted version number in a version check instead of a true
      number. IE:

        use Foo '1.234';

      gets passed to Foo->import() and not Foo->VERSION(). This is
      compounded by the fact that Exporter->import() *does* support a
      version check via import(). Arguably if we allow string versions
      (IMO we should) then we should deal with this in the parser, not
      in import() (although the latter is a little more debatable IMO).

    * Explicit calls to import() with a version number.

    * Mistakenly requesting an export from a module that doesn't
      actually export anything. One possible explanation for this case
      is that a module was at one point in its lifetime not a class, but
      then was later on turned into a class and the arguments to use
      statements weren't changed even though the main code was
      changed to use class syntax.

Because of the scope of the breakage of the original error this patch downgrades the error to a depecation warning instead, with the intention that in 5.44 we will make it a hard error.


haarg (approved):

jkeenan:
> [snip] 
> Because of the scope of the breakage of the original error this patch downgrades the error to a depecation warning instead, with the intention that in 5.44 we will make it a hard error.

I built and installed this branch on FreeBSD-12, installed `cpanm` against it, then used `cpanm` to install a list of 500 CPAN distributions (from an older "CPAN River" list) as well.  AFAICT, the only distros that were graded `FAIL` were distros that have been failing on this OS for years.  I got one instance of the `Attempt to call` warning -- which I think was from `cpanm` itself.

I then assembled a list of modules reported @cjg-cguevara, @andk and others in https://github.com/Perl/perl5/issues/21269 and used `cpanm` to try to install them against this branch.  The following modules FAILed to install:
```
App::adler
App::PerlXLock
App::Prove::Plugin::SetEnv
App::Sandy
Assert::Refute
Catalyst::Stuff
Data::Printer
Devel::TraceRun
Encode::IMAPUTF7
Exporter::NoWork
JIP::Spy::Events
MarpaX::Database::Terminfo
MarpaX::Java::ClassFile
maybe
Protocol::FIX
thanks
UNIVERSAL::DOES
Xporter
```
I should note, however, that these modules are (to me, at least) very obscure.  I can't recall trying to install them even when I was doing "CPAN River 3000" testing several years ago.  I haven't examined the `build.log` to see why they failed, but will do so upon request for particular distros.

As one might expect, during the second run of `cpanm` I recorded 289 instances of the `Attempt to call` warning.

demerphq:
@jkeenan thanks, that is promising feedback. Regarding cpanm it fatpacks a module that has the "version check as a string" bug, so that explains what you see. 

Fwiw, the reasons I haven't merged this today is that I have been quite busy, and because I think that the perldiag testing is broken as it hasn't complained about this change and I want to hunt down why before I merge. Perhaps that is not the right approach given how much trouble the fatal exception causes. Thoughts are welcome from you and @haarg.

PR #21294, opened 2023-07-26
A batch of locale handling patches

khwilliamson:


locale.c
```diff
@@ -1730,6 +1725,16 @@ S_calculate_LC_ALL_string(pTHX_ const char ** category_locales_list)
      * categories, adding new ones as they show up on obscure platforms.
      */
 
+    const char * my_category_locales_list[LOCALE_CATEGORIES_COUNT_];
+    const char ** locales_list = category_locales_list;
+    if (locales_list == NULL) {
+        locales_list = (const char **) &my_category_locales_list;
```
tonycoz: This should just be:
```
        locales_list = my_category_locales_list;
```
An array of T  (T = `const char *` here) decays to a pointer to T in most cases in C (`const char **` in this case).

The `&` on the array name is one of the few exceptions\*, which would make `&my_category_locales_list` a pointer to an array of LOCALE_CATEGORIES_COUNT_ pointers to const char *, which I suspect led you to adding the cast.

\* the other one I can think of is sizeof().

hints/darwin.sh
```diff
@@ -357,8 +357,7 @@ EOM
         esac
     fi
 
-    # The OS is buggy with respect to this.
```
tonycoz, line 360: commit message, perhaps:

Make sure *it* is executed while a mutex locks out other accesses.

Perhaps also "has *a* race" in the subject

locale.c
```diff
@@ -768,6 +768,200 @@ Perl_locale_panic(const char * msg,
...
+     * caller needs to arrange for each to be freed.
+     *
+     */
+
+    DEBUG_Lv(PerlIO_printf(Perl_debug_log,
+                           "Entering parse_LC_ALL_string; called from %"    \
+                           LINE_Tf "\nnew='%s'\n", caller_line, string));
+
+    const char separator[] = ";";
+    const Size_t separator_len = 1;
```
tonycoz, line 815: Are these ever not ";"? Or not a single character?

I do see this has changed from the code it replaces which uses character checks for ';'.

locale.c
```diff
@@ -1401,114 +1595,48 @@ S_setlocale_from_aggregate_LC_ALL(pTHX_ const char * locale, const line_t line)
...
-            p++;
-        }
+      case full_array:
+        break;
+    }
 
-        /* Find the index of the category name in our lists */
-        for (PERL_UINT_FAST8_T i = 0; i < LC_ALL_INDEX_; i++) {
+    /* Change each category to the value returned for it */
+    for (unsigned int i = 0; i < LC_ALL_INDEX_; i++) {
```
tonycoz: I do prefer the more general types here.  While it's not relevant here (LC_ALL_INDEX_ is a small integer), in some other cases CodeQL complains about using the possibly-small-PERL_UINT_FAST8_T variable as a loop index when the loop limit is a bigger type, for example:
```
regexec.c:4952
        for (i = 0; i < folds_to_count
                      + UNLIKELY(multi_fold_from != 0); i++)
Comparison between i of type PERL_UINT_FAST8_T and ... + ... of wider type unsigned long.
```
(this is as https://github.com/tonycoz/perl5/security/code-scanning/27 which I'm not sure you can see)

locale.c
khwilliamson: A future commit will take advantage of this being a variable, and with different potential values.

I have not seen any instances of more than a single character, but it seemed less brittle to allow a string for any such platforms, and with no extra logic needed.

locale.c
khwilliamson: A later commit will change to declare `i` to be an enum; I'm unsure if that is a good idea, but we can defer discussion until then.

tonycoz: You can't increment an enum type in C++:
```
tony@venus:.../perl/git$ g++ -oenuminc enuminc.cpp 
enuminc.cpp: In function ‘int main()’:
enuminc.cpp:9:33: error: no match for ‘operator++’ (operand type is ‘foo’)
    9 |     for (foo i = a; i < nomore; ++i)
      |                                 ^~~
tony@venus:.../perl/git$ cat enuminc.cpp
#include <iostream>

enum foo {
    a, b, nomore
};

int
main() {
    for (foo i = a; i < nomore; ++i)
        std::cout << i << '\n';
}
```
unless you provide an overloaded operator:
```
tony@venus:.../perl/git$ g++ -oenuminc enuminc.cpp 
tony@venus:.../perl/git$ cat enuminc.cpp
#include <iostream>

enum foo {
    a, b, nomore
};

foo operator++(foo &x) {
    x = static_cast<foo>(x+1);
    return x;
}

int
main() {
    for (foo i = a; i < nomore; ++i)
        std::cout << i << '\n';
}
tony@venus:.../perl/git$ ./enuminc
0
1
```

hints/darwin.sh
khwilliamson: done

locale.c
khwilliamson: done

locale.c
khwilliamson: Right, but I think it is better to wait to discuss this until we get to a commit that actually might use this

tonycoz (approved):

PR #21295, opened 2023-07-26
Regenerate Configure

khwilliamson:
These 3 commits each regenerate Configure.  The first two bring it up to date, and the third figures out the syntax of the string that represents LC_ALL when not all individual categories are set to the same locale.


Tux:
Can I have/get the metaunits you created to get to this regen?

Tux:
Based purely on the commits and the diffs, I see no objection, but I still like to see you commit the metaunits for me to review

khwilliamson:
Superseded by #21316

PR #21296, opened 2023-07-26
regexec.c - fix Clang compiler warning constant-logical-operand

Logikable:
This should fix the warning:

```
regexec.c:3465:21: error: use of logical '&&' with constant operand [-Werror,-Wconstant-logical-operand]
 3464 |                 if ((PL_sawampersand & SAWAMPERSAND_RIGHT)
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 3465 |                     && min >  prog->offs[0].end
      |                     ^
regexec.c:3465:21: note: use '&' for a bitwise operation
 3465 |                     && min >  prog->offs[0].end
      |                     ^~
      |                     &
regexec.c:3465:21: note: remove constant to silence this warning
```

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -3585,7 +3585,7 @@ S_reg_set_capture_string(pTHX_ REGEXP * const rx,
                     }
                     n++;
                 }
-                if ((PL_sawampersand & SAWAMPERSAND_RIGHT)
+                if ((PL_sawampersand & SAWAMPERSAND_RIGHT) > 0
                     && min >  RXp_OFFS_END(prog,0)
                 )
                     min = RXp_OFFS_END(prog,0);


jkeenan:
> This should fix the warning:
> 
> ```
> regexec.c:3465:21: error: use of logical '&&' with constant operand [-Werror,-Wconstant-logical-operand]
>  3464 |                 if ((PL_sawampersand & SAWAMPERSAND_RIGHT)
>       |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>  3465 |                     && min >  prog->offs[0].end
>       |                     ^
> regexec.c:3465:21: note: use '&' for a bitwise operation
>  3465 |                     && min >  prog->offs[0].end
>       |                     ^~
>       |                     &
> regexec.c:3465:21: note: remove constant to silence this warning
> ```

Can you provide the output of the following for the build-time warning you are proposing to repair?
```
$ uname -mrs

$ perl -v

$ perl -V
```
Thank you very much.

jkeenan:
Also, `clang --version`.

Leont:
That warning looks wrong to me?

demerphq:
Curious. SAWAMPERSAND_RIGHT is 0 I guess...

Leont:
> Curious. SAWAMPERSAND_RIGHT is 0 I guess...

That was also my first thought, but it's `4`: https://github.com/Perl/perl5/blob/blead/perl.h#L5946.

The real cause is a few lines below, PL_saw_ampersand is `(SAWAMPERSAND_LEFT|SAWAMPERSAND_MIDDLE|SAWAMPERSAND_RIGHT)` unless `PERL_SAWAMPERSAND` is defined.

demerphq:
Leont: yeah i was just going to say the same thing. So it ends up being true always. Kinda annoying that clang warns about this.

jkeenan:
This may be a devel version of `clang`.  Searching the net for that warning, I find references to: *A proposed update to clang's -Wconstant-logical-operand ...* (https://lore.kernel.org/all/20230718-nsecs_to_jiffies_timeout-constant-logical-operand-v1-0-36ed8fc8faea@kernel.org/).

I'm not getting any build-time warning in, *e.g.*, `clang-14` on Ubuntu 22.04 LTS on blead.

demerphq:
Something strange with this error. I cant find a version of regexec.c where line 3464/3465 has this code. The current code looks different.

demerphq:
I think this PR should be replaced by https://github.com/Perl/perl5/pull/21297

When we do not build with PERL_SAWAMPERSAND we turn PL_sawampersand into a constant, which means the code that contains the code that Clang complained about here is never used. So we can just ifdef it all away in a more simple way.

tonycoz:
> This may be a devel version of `clang`. Searching the net for that warning, I find references to: _A proposed update to clang's -Wconstant-logical-operand ..._ (https://lore.kernel.org/all/20230718-nsecs_to_jiffies_timeout-constant-logical-operand-v1-0-36ed8fc8faea@kernel.org/).

I see it with clang-17:
```
$ make regexec.o
clang-17 -c -DPERL_CORE -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings regexec.c
regexec.c:3589:21: warning: use of logical '&&' with constant operand [-Wconstant-logical-operand]
 3588 |                 if ((PL_sawampersand & SAWAMPERSAND_RIGHT)
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 3589 |                     && min >  RXp_OFFS_END(prog,0)
      |                     ^
regexec.c:3589:21: note: use '&' for a bitwise operation
 3589 |                     && min >  RXp_OFFS_END(prog,0)
      |                     ^~
      |                     &
regexec.c:3589:21: note: remove constant to silence this warning
1 warning generated.
tony@venus:.../git/perl6$ clang-17 --version
Debian clang version 17.0.0 (++20230722112248+ff380ced0519-1~exp1~20230722112304.100)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
tony@venus:.../git/perl6$
```
(clang installed from apt.llvm.org)

tonycoz:
Given my own testing and the fix above, I think clang is warning because we're combining the result of `&` with `&&`.  The patch here prevents that and suppresses the warning.

demerphq:
@tonycoz: even with your recent feedback (which i agree with) I do not like this patch and would prefer we use an alternative. 

I have repushed #21297 to use a different approach to solving this issue which I believe addresses your concerns.

tonycoz:
Fixed by 2a8fc21e747997d36fac7dceac7fa67680868e09.

Thanks for the patch.

PR #21297, opened 2023-07-26
regexec.c - restructure code that boils down to a constant conditional

demerphq:
When PERL_SAWAMPERSAND is not defined then Pl_sawampersand is defined
to be (SAWAMPERSAND_LEFT|SAWAMPERSAND_MIDDLE|SAWAMPERSAND_RIGHT).
Which means that the clause in the first if block:

```
    && !(PL_sawampersand & SAWAMPERSAND_RIGHT)
```

and the clause in the second if block:

```
    && !(PL_sawampersand & SAWAMPERSAND_LEFT)
```

will never be true. Thus neither of these blocks will execute in a
normal build where PERL_SAWAMPERSAND is not defined.

Nevertheless some versions of Clang notice that a related expression
guarded by these clauses is also constant, and mixes '&' and '&&'
together and then warns about it:

```
    $ clang-17 ... regexec.c
    regexec.c:3589:21: warning: use of logical '&&' with constant operand
        [-Wconstant-logical-operand]
     3588 |                 if ((PL_sawampersand & SAWAMPERSAND_RIGHT)
          |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     3589 |                     && min >  RXp_OFFS_END(prog,0)
          |                     ^
    regexec.c:3589:21: note: use '&' for a bitwise operation
     3589 |                     && min >  RXp_OFFS_END(prog,0)
          |                     ^~
          |                     &
    regexec.c:3589:21: note: remove constant to silence this warning
```

Restructuring the clauses into a nested if should fix this. Along the
way we fix a related I32/SSize_t mismatch, and eliminate an unnecessary
duplicate use of RXp_OFFS_END(prog,0) call, which these days is a more
complex and expensive macro than it once was.

Thanks to the github user @Logikable for calling this to our attention
with a different patch. See also https://github.com/Perl/perl5/pull/21296.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -3545,6 +3545,10 @@ S_reg_set_capture_string(pTHX_ REGEXP * const rx,
             SSize_t max = strend - strbeg;
             SSize_t sublen;
 
+            /* NOTE: the following if block is not used or tested
+             * in standard builds. It is only used when PERL_SAWAMPERSAND is
+             * defined */
+
             if (    (flags & REXEC_COPY_SKIP_POST)
                 && !(prog->extflags & RXf_PMf_KEEPCOPY) /* //p */
                 && !(PL_sawampersand & SAWAMPERSAND_RIGHT)
@@ -3567,6 +3571,9 @@ S_reg_set_capture_string(pTHX_ REGEXP * const rx,
                 assert(max >= 0 && max <= strend - strbeg);
             }
 
+            /* NOTE: the following if block is not used or tested
+             * in standard builds. It is only used when PERL_SAWAMPERSAND is
+             * defined */
             if (    (flags & REXEC_COPY_SKIP_PRE)
                 && !(prog->extflags & RXf_PMf_KEEPCOPY) /* //p */
                 && !(PL_sawampersand & SAWAMPERSAND_LEFT)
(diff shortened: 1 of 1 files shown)


Leont:
AFAICT we could simplify this by using `#ifdef PERL_SAWAMPERSAND` instead of defining a new macro `SAWAMPERSAND_ALWAYS`

demerphq:
@leont I thought about that, but I didnt like the way it read. I found this less confusing. PERL_SAWAMPERSAND isnt very self-documenting compared to SAWAMPERSAND_ALWAYS. I plan to review this code in more detail this cycle anyway. Id like to keep it more intuitively self-documenting for later.

tonycoz:
An optimized build doesn't generate code for these blocks anyway.

By making this code invisible to the vast majority of builds, ie. builds without -DPERL_SAWAMPERSAND, changes outside of here that would break compiling -DPERL_SAWAMPERSAND builds won't be detected.

I prefer something like the original commit (or just remove PERL_SAWAMPERSAND support)

demerphq:
> I prefer something like the original commit 

I really don't like the original commit, I think it just obfuscates what is going on. However I do agree with your analysis here, and have updated my counter-proposal in a way which i believe resolves your and @Leont's concerns while also satisfying my own, while at the same time giving @Logikable credit.

> (or just remove PERL_SAWAMPERSAND support)

I am not sure we *should* remove PERL_SAWAMPERSAND. There was this premise that COW fixed the issues with use of $&, but it doesn't really, not comprehensively anyway, so there are cases where code goes quadratic that should and would be linear if we were building under PERL_SAWAMPERSAND.

tonycoz (approved):

Issue #21187, opened 2023-07-01
close() no longer returns success after a failed read on a write-only filehandle


steve-m-hay:
The program below reproduces the problem.

With 5.36.1 the final close() call returns 1 (success). The output is as expected:
```
open(): 1
print(): 1
<>: undef
close(): 1
```

With 5.38.0-RC2 it returns an empty string (failure). The output is:
```
open(): 1
print(): 1
<>: undef
close():
$!=Bad file descriptor, $^E=
```

It works (i.e. close returns 1) if I remove the deliberate failing attempt at reading from the write-only filehandle, so presumably this is somehow interfering with the return value of the close() call.

(If I insert a pause (e.g. a read on <STDIN>) after the close() call then I find that it is possible to manually delete the file that was opened, which suggests that the close() operation has succeeded; it is just the return value from it that is wrong. By contrast, if I pause before the close() call then it is not possible to manually delete the file at that point, so hopefully that conclusion is correct.)

I have tested this with and without the recent CloseHandle() fix (commit 149c2e48b61bd93a9e4ee06b920ca68dd97b717c), but I have the same result (failure) either way. In fact, it also fails with 5.37.9, which is prior to the original CloseHandle() change (commit 8552f09f5cfe61a536a65f11290ef026f7aa0356) so the problem must be something else.

Testing further, I've narrowed it down a bit more: This worked in 5.37.3, but fails in 5.37.4.

```
use strict;
use warnings;
my $file = 'test.txt';
my $ret = open my $fh, '>', $file;
print "open(): $ret\n";
$ret = print $fh "WTF\n";
print "print(): $ret\n";
seek $fh, 0, 0;
{
# Suppress "Filehandle $fh opened only for output" warning
no warnings 'io';
$ret = <$fh>;
print "<>: " . (defined $ret ? $ret : "undef") . "\n";
}
$ret = close $fh;
print "close(): $ret\n";
print "\$!=$!, \$^E=$^E\n" unless $ret;
```

This is on Windows, with Visual Studio 2022 v17.4. perl -V as follows (all default build options):

Summary of my perl5 (revision 5 version 38 subversion 0) configuration:
   
  Platform:
    osname=MSWin32
    osvers=10.0.19045.3086
    archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
[...]

PR #21298, opened 2023-07-27
close: document recent-ish changes to closing file handles

tonycoz:
Where "recent" includes 5.22.0

I think this documents some of the things that confused @steve-m-hay in #21187.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -1356,8 +1356,15 @@ X<close>
 
 Closes the file or pipe associated with the filehandle, flushes the IO
 buffers, and closes the system file descriptor.  Returns true if those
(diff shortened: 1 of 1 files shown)


bram-perl:
I think I would like the message more when it was:

>  Returns true if those operations succeed, **and** if no error was reported by any PerlIO layer, and there was no existing error on the filehandle.

Without the 'and' it feels a bit less clear (to me).

Or maybe in list form?:

> Returns true if:
> - those operations succeeded; and
> - no error was reported by any PerlIO layer; and
> - there was no existing error on the filehandle.

tonycoz:
> I think I would like the message more when it was:

Updated to add the "and" back

khwilliamson (approved):

PR #21304, opened 2023-07-27
testuite.yml - add PERL_SAWAMPERSAND testing

demerphq:
Make sure we test this build mode.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -265,6 +265,7 @@ jobs:
           - "-Dcc='g++'"
           - "-Accflags=-DSILENT_NO_TAINT_SUPPORT"
           - "-Accflags=-DNO_TAINT_SUPPORT"
+          - "-Accflags=-DPERL_SAWAMPERSAND"
 
     steps:
       - name: Install System dependencies


bram-perl:
Can you elaborate (in the commit message) on why it's important to test this partciular configuration all the time?

demerphq:
@bram-perl I think we should be testing all of our optional build modes all the time, so to me this is like asking "why do you bother with CI". But I have updated the commit regardles.

khwilliamson:
On 7/27/23 13:48, Yves Orton wrote:
> @bram-perl <https://github.com/bram-perl> I think we should be testing 
> all of our optional build modes all the time, so to me this is like 
> asking "why do you bother with CI". But I have updated the commit regardles.
> 
>

Testing all is impractical.  For my locale changes, I have over 200 
different combinations of build options.  This is because platforms vary 
greatly in their locale handling

demerphq:
@khwilliamson The ideal is often impractical. My point was that I shouldn't have to explain why I am adding one to be tested. If we had infinite resources we would test them all. It is your call if you want to test your options or not. That you choose not to test some doesn't mean I should have to explain why I am testing another.

bram-perl:
The point of explaining it (in the commit message) is so that people later know **why** it was done.

Let's **assume** that there is no reason in the commit message and that someone in the future decides to remove PERL_SAWAMPERSAND: then they'll notice that the testsuite fails and just remove the test configuration. After all there is no one that remembers why it was added in the first place.

tonycoz:
I know there have been some efforts to reduce github workflow minute usage in the past, but I haven't been able to find any official limitations on public repositories.

See the testsuite.yml re-work from 430c0c0cde3

@atoomic do you know if there are any limits?

rwp0 (approved):

PR #21306, opened 2023-07-28
Add `__CLASS__` keyword

leonerd:
Also a couple of small neatenings of code/test; kept as separate commits in case of subsequent revert


leonerd:
Hrm; everything passes except the `g++` one which fails:

```
/usr/bin/ld: globals.o:(.data.rel.local+0xd48): undefined reference to `Perl_ck_classname'
/usr/bin/ld: miniperl: hidden symbol `Perl_ck_classname' isn't defined
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
```

probably because the newly-added opchecker function isn't done right, somehow. Hrm.

leonerd:
I can reproduce this locally, and it is fixed by simply moving the `Perl_ck_classname()` function out of class.c into op.c. Presumably therefore, something isn't expecting to find it in that particular file. Possibly a list of files to expose needs updating? But weird that it only happens under g++ and not regular gcc.

leonerd:
g++ now fixed. Turns out it needed some extra poking in `regen/embed_lib.pl`, which @ilmari found. Thanks :)

pod/perlfunc.pod
```diff
@@ -6059,6 +6060,37 @@ X<__PACKAGE__>
 
 A special token that returns the name of the package in which it occurs.
 
+=item __CLASS__
+X<__CLASS__>
+
+=for Pod::Functions the class name of the current instance.
+
+Invoked within a C<method> or similar location, such as a field initializer
```
xenu: I think `method` and `__PACKAGE__` should be linkified.

leonerd: Added; forcepushed.

class.c
```diff
@@ -1059,6 +1059,29 @@ Perl_class_add_ADJUST(pTHX_ HV *stash, CV *cv)
     av_push(aux->xhv_class_adjust_blocks, (SV *)cv);
 }
 
+OP *
+Perl_ck_classname(pTHX_ OP *o)
+{
+    if(!CvIsMETHOD(PL_compcv))
+        croak("Cannot use __CLASS__ outside of a method or field initializer expression");
```
Ovid, line 1066: Pardon my ignorance 😃 I think `ADJUST` is currently considered a method, but eventually will become a phaser. It is intended that `__CLASS__` will work there?

pod/perlclass.pod
```diff
@@ -120,6 +120,27 @@ constructor at all. Expressions using C<//=> will also apply if the caller did
...
+    class WithCustomField {
+        use constant DEFAULT_X => 10;
+        field $x = __CLASS__->DEFAULT_X;
+    }
+
+This allows subclasses to override the method with different behaviour.
+
+    class DifferentCustomField :isa(WithCustomField) {
+        sub DEFAULT_X { rand > 0.5 ? 20 : 30 }
+    }
```
Ovid, line 138: The initial design of Corinna was to separate methods and subroutines. Thus, you can't call subs like you can call methods. If we had the `:common` attribute, this could be written as:

```perl
    class WithCustomField {
        method DEFAULT_X :common {10}
        field $x = __CLASS__->DEFAULT_X;
    }
```

This allows subclasses to override the method with different behaviour.

```perl
    class DifferentCustomField :isa(WithCustomField) {
        method DEFAULT_X :common { rand > 0.5 ? 20 : 30 }
    }
```

t/class/class.t
```diff
@@ -14,12 +14,16 @@ no warnings 'experimental::class';
...
+
+        method classname { return __CLASS__; }
     }
 
     my $obj = Test1->new;
     isa_ok($obj, "Test1", '$obj');
 
     is($obj->hello, "hello, world", '$obj->hello');
+
+    is($obj->classname, "Test1", '$obj->classname yields __CLASS__');
```
Ovid, line 26: Can we add a test verifying that `__CLASS__` is available in `ADJUST` blocks?

class.c
leonerd: Yeah definitely. The ADJUST blocks will become individual blocks that are all still part of one overall containing method, so `__CLASS__` still still work fine there. It'll just stop being one method *per block*.

pod/perlclass.pod
leonerd: Indeed so. This is the best we can write currently, but once we have common methods we can change the documentation to fit the better design.

t/class/class.t
leonerd: Added, re-testing now.

pod/perlclass.pod
Ovid: Maybe we should add a note about this in the documentation? If people start writing code relying on this, future releases will break it.

leonerd: It's all very new and experimental now anyway. Any of the docs should be read in that light anyway, so I don't think adding specific wording to basically every claim and sugestion is going to be all that helpful.

rwp0 (approved):

class.c
```diff
@@ -1059,6 +1059,29 @@ Perl_class_add_ADJUST(pTHX_ HV *stash, CV *cv)
...
+PP(pp_classname)
+{
+    dSP;
+
+    SV *self = PAD_SVl(PADIX_SELF);
+    assert(SvTYPE(SvRV(self)) == SVt_PVOBJ);
+
+    EXTEND(SP, 1);
+    PUSHs(sv_newmortal());
+    sv_ref(*SP, SvRV(self), true);
```
tonycoz: Wouldn't it be better to make this use TARG?

leonerd: Ahyes, good call. Have added in a new commit for now; needs squashing before merge.

tonycoz (approved):

PR #21310, opened 2023-07-28
hints/linux: Add additional expression when matching clang

### LINKED ISSUE #21099: configure script is incorrectly checking for clang
https://github.com/Perl/perl5/blob/1515f807817872eac26fdb0ff7469e2b71ebc6b5/hints/linux.sh#L169

The configure script runs the following code:
```
# Ubuntu 11.04 (and later, presumably) doesn't keep most libraries
# (such as -lm) in /lib or /usr/lib.  So we have to ask gcc to tell us
# where to look.  We don't want gcc's own libraries, however, so we
# filter those out.
# This could be conditional on Unbuntu, but other distributions may
# follow suit, and this scheme seems to work even on rather old gcc's.
# This unconditionally uses gcc because even if the user is using another
# compiler, we still need to find the math library and friends, and I don't
# know how other compilers will cope with that situation.
# Morever, if the user has their own gcc earlier in $PATH than the system gcc,
# we don't want its libraries. So we try to prefer the system gcc
# Still, as an escape hatch, allow Configure command line overrides to
# plibpth to bypass this check.
if [ -x /usr/bin/gcc ] ; then
    gcc=/usr/bin/gcc
# clang also provides -print-search-dirs
elif ${cc:-cc} --version 2>/dev/null | grep -q '^clang ' ; then
    gcc=${cc:-cc}
else
    gcc=gcc
fi
```

When running on a system that only has clang, and not gcc. i.e., `cc=x86_64-pc-linux-gnu-clang`, we hit the `elif` branch. The grep command is a bit too strict when determining if the compiler is clang.

```
$ clang --version
Debian clang version 14.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
```
# x86_64-pc-linux-gnu-clang --version
Chromium OS 17.0_pre496208_p20230501-r1 clang version 17.0.0 (/mnt/host/source/src/third_party/llvm-project 98f5a340975bc00197c57e39eb4ca26e2da0e8a2)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

It looks like 'clang' doesn't appear at the beginning of the line like expected, but in the middle. Changing the check to `grep -q '\bclang\b'` for me works as expected.

ismell:
Newer versions of clang actually print the distro before `clang` when
calling `$CC --version`. This changes fixes the regex so it can match
this new pattern.

i.e.,

	$ clang --version
	Debian clang version 14.0.6

	$ x86_64-pc-linux-gnu-clang --version
	Chromium OS 17.0_pre496208_p20230501-r1 clang version 17.0.0

Fixes #21099

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1158,6 +1158,7 @@ Randy Stauner                  <rwstauner@cpan.org>
 Randy W. Sims
 Raphael Manfredi               <Raphael.Manfredi@pobox.com>
 Raul Dias                      <raul@dias.com.br>
+Raul E Rangel                  <rrangel@chromium.org>
 Raymund Will                   <ray@caldera.de>
 Redvers Davies                 <red@criticalintegration.com>
 Reini Urban                    <rurban@cpan.org>
--- a/hints/linux.sh
+++ b/hints/linux.sh
@@ -166,7 +166,7 @@ esac
 if [ -x /usr/bin/gcc ] ; then
     gcc=/usr/bin/gcc
 # clang also provides -print-search-dirs
-elif ${cc:-cc} --version 2>/dev/null | grep -q '^clang ' ; then
+elif ${cc:-cc} --version 2>/dev/null | grep -q -e '^clang version' -e ' clang version'; then
     gcc=${cc:-cc}
 else
     gcc=gcc


jkeenan:
You've modified code *only* in `hints/linux.sh`.  Is Linux the *only* OS where this would be a concern?  What about, say, FreeBSD, where `clang` is the default C-compiler?

jkeenan:
Also, is there some particular reason why you have marked this pull request as a 'draft'?

ismell:
It doesn't look like any of the other OS' rely on a check like this. So it's linux specific. I had it as a draft because I had a build going to verify it was working. It just passed, so I'm marking it ready for review.

tonycoz (approved):

ismell:
I pushed a new version that fixes the AUTHORS test failure.

jkeenan:
Self-assigning this ticket for the purpose of merging it to blead when CI has completed.

PR #21316, opened 2023-08-01
Regenerate Configure from latest metaconfig

khwilliamson:
This includes the LC_ALL notation unit, not yet merged in metaconfig.

It turns out that in order to get quite a few elements from being dropped, I had to add them to metaconfig.h


khwilliamson:
approved by @Tux via irc

PR #21323, opened 2023-08-02
Start handling platforms whose representation of LC_ALL is postional when not all categories are the same 

khwilliamson:
Some platforms use a positional notation for LC_ALL when not all
 categories are the same, like
     
         C/English/C.UTF-8/ ...
     
These commits take advantage of the new Configure probe to now handle  those when parsing an LC_ALL string.

Prior to this sequence, we only properly handled platforms where the representation is

 name1=value1;name2=value2; ...

Further commits are needed (and will be supplied) to fully handle the positional notation, but these get us started


tonycoz:
It looks reasonable, but fails to build on Windows and Mac.

PR #21326, opened 2023-08-04
Pod simple 3.45 sync 20230802

### LINKED ISSUE #21311: Pod::Simple 3.45 breaks Pod::Html test suite
Module: Pod::Html

**Description**

As a result, the attempt to bring the latest Pod::Simple into blead was reverted: 643dfa470afd0bc8cec06c90184bc5c131035fa4.

I tracked down the break to changing this line in  XHTML.pm from
` s/&[^;]+;//g;            # Strip entities.`

to 
`s/[<>&'"]//g;            # Strip HTML special characters`

 from Pod::Simple  commit b074ee34e12279aeba3f5a3edc9968a6127c3de0

>  *Author: Graham Knop <haarg@haarg.org>
> Date:   Tue May 24 14:59:08 2022 +0200
> refactor link handling to encode in correct locations

@haarg is of the opinion that the bug is actually in Pod::Html:

> definitely seems to be a Pod::Html bug. it's trying to use the idify function, but isn't calling it properly. it's also relying on very specific internals. and the behavior the test wants is different from the documented behavior of htmlify.
i'd say that the logic from Pod::Simple::XHTML should just be copied into Pod::Html and modified to be what it desires.
or possibly change the test to expect quotes to be stripped

**Steps to Reproduce**

Revert 643dfa470afd0bc8cec06c90184bc5c131035fa4; run test suite
**Expected behavior**

No errors

**Perl configuration**

```
# perl -V output goes here
```
``

jkeenan:
This branch completes the synching into blead of Pod-Simple-3.45.  This synching had to be reverted because of a side effect failure in `ext/Pod-Html/t/*.t`.  (It's admittedly kludgy -- but so is most of Pod-Html.)

Fixes https://github.com/Perl/perl5/issues/21311


haarg (approved):
I've reviewed this again, and realized that the `anchorify` function only has exactly one user, the `installhtml` script, which uses it to generate file names for the `--splithead` option. The `--splithead` option is only used for a single file: `perlipc.pod`. It is not used by anything on CPAN.

Since it is generating file names, not internal anchors, I think it makes sense for it to have its own implementation, not being tied directly to what Pod::Simple::XHTML uses for anchors.

This looks like the correct solution to me.

khwilliamson (approved):

jkeenan:
Merged, but unfortunately I forgot to squash commits before I did the merge.  (Those GH buttons are sometimes *too* easy and too tempting.)

Issue #19764, opened 2022-05-21
[EXPERIMENT] the builtin:: namespace


rjbs:
In Perl 5.36.0, we added the `builtin` namespace for new functions that are not always available in every namespace.  (They must be imported or called by fully-qualified name.)

rwp0:
```
> perl -E 'say for keys %builtin::'
trim
weaken
refaddr
import
is_weak
reftype
created_as_string
is_bool
floor
unweaken
true
ceil
created_as_number
blessed
false
indexed
```

that's how to list them for those who wonder what functions are available in the new namespace

leonerd:
The module itself is considered experimental, as is additionally every individual function that it shipped with. Those functions are:

    true false is_bool
    weaken unweaken is_weak
    blessed refaddr reftype
    created_as_string created_as_number
    ceil floor
    trim
    indexed

No individual function can be made non-experimental before the builtin namespace mechanism itself has been. However, it may turn out that individual functions can have a longer experimental status once the overall mechanism and most of its functions are declared stable. If that becomes the case, new Experiment issues should be created for each of those functions.

Any newly-added functions should be considered as separate experiments beyond this, with their own lifetimes.

leonerd:
Additional abilities that would need adding before deëxperimentalizing:

 - [ ] builtin bundles - a.la. `use builtin ':v5.37'`
 - [ ] integration with `use VERSION` - `use v5.37` to imply `use builtin ':v5.37'`

PR #21330, opened 2023-08-07
Deëxperiment most of the `builtin` functions

leonerd:
This PR removes the experimental status and associated `experimental::builtin` warning from *most* of the functions in the `builtin::` space.

I have been somewhat conservative in which functions to downgrade and which remain. The now-stable functions are the true/false constants, the weakref and reference query functions (copied from `Scalar::Util`), the numercal rounding functions, `indexed` and `trim`.

Still remaining experimental are:

 * `is_bool` - because stable boolean tracking was only recently introduced and may not have been thoroughly tested yet
 * `created_as_string`, `created_as_number` - similar
 * `export_lexically` - because lexical export of subs is equally still quite new and may need more testing

See also https://github.com/Perl/perl5/issues/19764

### DIFF
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,4 +1,4 @@
-package builtin 0.008;
+package builtin 0.009;
 
 use strict;
 use warnings;
@@ -40,12 +40,9 @@ can be requested for convenience.
 Individual named functions can be imported by listing them as import
 parameters on the C<use> statement for this pragma.
 
-The overall C<builtin> mechanism, as well as every individual function it
-provides, are currently B<experimental>.
-
-B<Warning>:  At present, the entire C<builtin> namespace is experimental.
-Calling functions in it will trigger warnings of the C<experimental::builtin>
-category.
+B<Warning>:  At present, many of the functions in the C<builtin> namespace are
+experimental.  Calling them will trigger warnings of the
+C<experimental::builtin> category.
 
 =head2 Lexical Import
 
@@ -105,6 +102,8 @@ This gives an equivalent value to expressions like C<!!0> or C<!1>.
 =head2 is_bool
 
     $bool = is_bool($val);
+
+This function is currently B<experimental>.
 
 Returns true when given a distinguished boolean value, or false if not. A
 distinguished boolean value is the result of any boolean-returning builtin
@@ -167,6 +166,8 @@ C<ARRAY> for array references, or C<HASH> for hash references.
 
     $bool = created_as_string($val);
 
+This function is currently B<experimental>.
+
 Returns a boolean representing if the argument value was originally created as
 a string. It will return true for any scalar expression whose most recent
 assignment or modification was of a string-like nature - such as assignment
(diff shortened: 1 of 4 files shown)


leonerd:
I am leaving this PR as a draft currently as such a downgrade likely warrants a decision from the (next) PSC, and currently there is no PSC due to the voting being in progress. I'll undraft it once there is a new council.

lib/builtin.pm
```diff
@@ -40,12 +40,9 @@ can be requested for convenience.
 Individual named functions can be imported by listing them as import
 parameters on the C<use> statement for this pragma.
 
-The overall C<builtin> mechanism, as well as every individual function it
-provides, are currently B<experimental>.
-
-B<Warning>:  At present, the entire C<builtin> namespace is experimental.
-Calling functions in it will trigger warnings of the C<experimental::builtin>
-category.
+B<Warning>:  At present, the many functions in the C<builtin> namespace are
```
ilmari: Should this not be "many of the functions"?

leonerd: Ahyes; fixed.

Leont:
> is_tainted - because taint mode

I don't understand why this would need to remain experimental.

leonerd:
> I don't understand why this would need to remain experimental.

I felt there were vague thoughts about maybe making taint-mode optional at some point in the future..? If so we might not want to go adding stable bits of API if we're just going to be removing it again anyway.

Other than that there's no huge reason why...

JRaspass:
I presumed it would just return always false on such builds

Leont:
> I felt there were vague thoughts about maybe making taint-mode optional at some point in the future..? If so we might not want to go adding stable bits of API if we're just going to be removing it again anyway.
>
> Other than that there's no huge reason why...

It may or may not disappear on the long run, but in the near future it would be a build option anyway so it'd still exist.

> I presumed it would just return always false on such builds

Exactly.

leonerd:
> > I presumed it would just return always false on such builds
> 
> Exactly.

Fair. Alright then; I'll remove the warning from that one too. Added to the PR.

rwp0 (approved):

tonycoz (approved):

leonerd:
Squashed to a single commit with a better message. Last chance to object now, or I'll merge it tomorrow morning if CI is happy.

EvanCarroll:
@leonerd separately is it possible to get boolean tracking on `AV*`? Such that

```perl
perl -we'sub foo { undef }; my @foo = foo(); die @foo ? "list is true" : "list is false";
```

dies with "list is false". It's awkward that `undef` evaluates to `false` in scalar context, but not list context when assigned to an array. If boolean tracking was on `AV*s` then we could on list-construction we could set the AV to false in such a situation, and simply flip the flag on all other operations (to true). This would close the gap and allow things like `my @methods = package->can("nothing"); return false regardless of context, without having to write custom implementations with `wantarray` on everything to cover this.

haarg:
That doesn't have anything to do with builtin, so this ticket is not the right place for it.

It also would be a huge compatibility break. It is not going to happen.

PR #21334, opened 2023-08-08
locale.c: Revamp S_stdize_locale()

khwilliamson:
We are now able to parse and construct non-uniform LC_ALL strings.  A
big part of this function was trying to deal with these, but now that
can be farmed out.

It's now extremely unlikely this function will now actually change
anything, based on recent field experience.  To save copying that
therefore won't likely ever be needed, this function now returns its
input if unchanged, and a mortalized copy if it unlikely does need to be
changed.  That means it no longer needs a buffer reserved for it.

The input is going to be the static global buffer pointed to by the
return of setlocale().  That means calls to this function need to be
protected by mutexes in threaded systems (the same as before this
commit), and the result safely copied during the critical section.


hints/aix.sh
```diff
@@ -694,5 +694,10 @@ EOF
   ;;
 esac # not gcc
 
+case "$osvers" in
```
tonycoz, line 697: Commit message: "hints page"?  should this be "hints file"?

tonycoz (approved):

hints/aix.sh
khwilliamson: yes, fixed

PR #21340, opened 2023-08-08
A batch of locale-related commits

khwilliamson:
These mostly are for fixing up  S_find_locale_from_environment().

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -2650,6 +2650,7 @@ S_find_locale_from_environment(pTHX_ const unsigned int index)
      */
 
     const char * const lc_all = PerlEnv_getenv("LC_ALL");
+    const char * locale_names[LC_ALL_INDEX_] = { NULL };
 
     /* Use any "LC_ALL" environment variable, as it overrides everything else.
      * */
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -2485,9 +2485,9 @@ S_find_locale_from_environment(pTHX_ const unsigned int index)
 
         /* If no corresponding environment variable, see if LANG exists.  If
          * so, use it. */
-        const char * default_name = PerlEnv_getenv("LANG");
-        if (default_name && strNE(default_name, "")) {
-            return default_name;
+        const char * LANG = PerlEnv_getenv("LANG");
+        if (LANG && strNE(LANG, "")) {
+            return LANG;
```
tonycoz: I don't think the variable name should be all upper-case, it screams "macro" to me.

Maybe `env_lang` or `env_LANG` (I prefer the first).

khwilliamson: changed to env_lang

locale.c
```diff
@@ -2592,46 +2592,68 @@ S_calculate_LC_ALL_string(pTHX_ const char ** category_locales_list,
...
+     *      setting ourselves, using the documented algorithm specified by the
+     *      POSIX standard (assuming the platform follows the Standard) rather
+     *      than use "" as the locale.  This will lead to results that differ
+     *      from native behavior if the native behavior differs from the
+     *      Standard's documented value, but khw believes it is better to know
+     *      what's going on, even if different from native, than to just guess.
+     *
+     *      Debian systems differ from this standard in having a LANGUAGE
+     *      environment variable used for just LC_MESSAGES.  This function does
+     *      NOT handle that.
```
tonycoz: I don't think this is Debian specific, on Fedora 38:
```
[tony@fedora git]$ cat lc_messages.c
#include <stdio.h>
#include <locale.h>
#include <errno.h>

int main() {
  setlocale(LC_ALL, "");
  errno = EINVAL;
  perror("");
}
[tony@fedora git]$ gcc -olc_messages lc_messages.c
[tony@fedora git]$ ./lc_messages 
Invalid argument
[tony@fedora git]$ LANGUAGE=de_DE.utf8 ./lc_messages 
Das Argument ist ungültig
```
Debian does have a patch against glibc to [include the LANGUAGE environment variable in the `locale` tool's output](https://sources.debian.org/patches/glibc/2.36-9%2Bdeb12u1/locale/locale-print-LANGUAGE.diff/) but I don't see anything to indicate Debian are patching this.

The glibc sources indicate this is is a [GNU extension](https://sourceware.org/git/?p=glibc.git;a=blob;f=intl/dcigettext.c;h=7886ac9545e33367eacd8d8ba1497bc45ffbe367;hb=HEAD#l1517), which is presumably why Debian patched `locale` to improve diagnostic reports.

So I'd expect all glibc systems to use `LANGUAGE`.

(This started as "Is this just Debian too, or Ubuntu as well", and down the rabbit hole I went.)

embed.fnc
```diff
@@ -4388,6 +4388,14 @@ S	|const char *|my_langinfo_i					\
 				|NULLOK Size_t *retbuf_sizep		\
```
tonycoz: commit message:

Is this even valid?
```
export LC_ALL="LC_CTYPE=de_DE;LC_NUMERIC=C;LC_TIME=de_DE;..."
```
If I try with the glibc setlocale on Debian stable:
```
tony@venus:.../perl/git$ cat locale-disp.c
#include <locale.h>
#include <stdio.h>

int main() {
    printf("setlocale %s\n", setlocale(LC_ALL, ""));
}
tony@venus:.../perl/git$ gcc -olocale-disp locale-disp.c 
tony@venus:.../perl/git$ ./locale-disp 
setlocale en_AU.UTF-8
tony@venus:.../perl/git$ LC_MESSAGES=de_DE.utf8 ./locale-disp 
setlocale LC_CTYPE=en_AU.UTF-8;LC_NUMERIC=en_AU.UTF-8;LC_TIME=en_AU.UTF-8;LC_COLLATE=en_AU.UTF-8;LC_MONETARY=en_AU.UTF-8;LC_MESSAGES=de_DE.utf8;LC_PAPER=en_AU.UTF-8;LC_NAME=en_AU.UTF-8;LC_ADDRESS=en_AU.UTF-8;LC_TELEPHONE=en_AU.UTF-8;LC_MEASUREMENT=en_AU.UTF-8;LC_IDENTIFICATION=en_AU.UTF-8
tony@venus:.../perl/git$ LC_ALL="en_AU.UTF-8;LC_NUMERIC=en_AU.UTF-8;LC_TIME=en_AU.UTF-8;LC_COLLATE=en_AU.UTF-8;LC_MONETARY=en_AU.UTF-8;LC_MESSAGES=de_DE.utf8;LC_PAPER=en_AU.UTF-8;LC_NAME=en_AU.UTF-8;LC_ADDRESS=en_AU.UTF-8;LC_TELEPHONE=en_AU.UTF-8;LC_MEASUREMENT=en_AU.UTF-8;LC_IDENTIFICATION=en_AU.UTF-8" ./locale-disp 
setlocale (null)
```

khwilliamson: Try s/utf8/UTF-8/.  That works for me.  Some platforms require the uppercase and dash

locale.c
khwilliamson: I did s/Debian/glibc/ here

embed.fnc
tonycoz: The generated LC_ALL value for the last test used the `.UTF-8` name, so I don't expect any difference, but I tried it anyway:
```
tony@venus:.../perl/git$ LC_MESSAGES=de_DE.UTF-8 ./locale-disp 
setlocale LC_CTYPE=en_AU.UTF-8;LC_NUMERIC=en_AU.UTF-8;LC_TIME=en_AU.UTF-8;LC_COLLATE=en_AU.UTF-8;LC_MONETARY=en_AU.UTF-8;LC_MESSAGES=de_DE.UTF-8;LC_PAPER=en_AU.UTF-8;LC_NAME=en_AU.UTF-8;LC_ADDRESS=en_AU.UTF-8;LC_TELEPHONE=en_AU.UTF-8;LC_MEASUREMENT=en_AU.UTF-8;LC_IDENTIFICATION=en_AU.UTF-8
tony@venus:.../perl/git$ LC_ALL="LC_CTYPE=en_AU.UTF-8;LC_NUMERIC=en_AU.UTF-8;LC_TIME=en_AU.UTF-8;LC_COLLATE=en_AU.UTF-8;LC_MONETARY=en_AU.UTF-8;LC_MESSAGES=de_DE.UTF-8;LC_PAPER=en_AU.UTF-8;LC_NAME=en_AU.UTF-8;LC_ADDRESS=en_AU.UTF-8;LC_TELEPHONE=en_AU.UTF-8;LC_MEASUREMENT=en_AU.UTF-8;LC_IDENTIFICATION=en_AU.UTF-8" ./locale-disp 
setlocale (null)
```
so no change.

Do you have a platform where the libc `setlocale(LC_ALL, "")` allows the mixed result from setlocale() to be used in  LC_ALL in the environment?

khwilliamson:
I misunderstood your point.  The commit message says, "There is no test added because I didn't find a shell that allows setting
LC_ALL to such a disparate value."  This is because the values that LC_ALL  can be set to in the environment are under control of the shell and the libc functions it calls.

The first call to your program in your example shows that setlocale() can be called with a disparate LC_ALL.  

But it falls apart in the next call.  Now [https://pubs.opengroup.org/onlinepubs/9699919799/](url) says setenv() is supposed to fail if the string contains an equals sign.  But if you echo $LC_ALL, even after exporting, it does contain an equals, so the system here is disobeying the POSIX standard (though maybe it isn't using setenv(), but some non-standard way of putting things into the environment.

So I tried on a box that uses a positional notation for disparate LC_ALL strings.  There is no equals sign, but I get the same results.  LC_ALL can be set to a disparate value and exported, but it still returns null for the second call

So, should we include this commit?  My take when developing it was that it's not illegal (certainly the positional notation platforms don't violate the Standard) , the code is straight forward, and short, and it would be better to anticipate that such a system could exist, and prepare for that eventuality.

tonycoz:
> The first call to your program in your example shows that setlocale() can be called with a disparate LC_ALL.

The first example has only a single locale in the LC_ALL environment variable, not the more complex result that setlocale() returns or accepts for its second parameter.

So yes, setlocale()s second parameter can (typically\*) be a disparate locale string

> But it falls apart in the next call. Now [https://pubs.opengroup.org/onlinepubs/9699919799/](url) says setenv() is supposed to fail if the string contains an equals sign. But if you echo $LC_ALL, even after exporting, it does contain an equals, so the system here is disobeying the POSIX standard (though maybe it isn't using setenv(), but some non-standard way of putting things into the environment.

Your link is to the top frame set of the POSIX spec, not a specific page, and has the text and url fields swapped.

I assume you meant the setenv page: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setenv.html

which says:

> The function shall fail if envname points to a string which contains an '=' character.

but it's not the name here (LC_ALL) that contains the `=` character.

> So, should we include this commit?

My main issue is it might mislead someone into thinking you can set LC_ALL in the environment to the result of setlocale(), which isn't the case.

\* since that's the simplest implementation, as opposed to some signed base64 encoding of the locale state, as an extreme example

khwilliamson:
This came about as I was code reading, and realized that the code just assumed that the environment variable LC_ALL contained a single value.  And, since there is now a function that will reliably parse any LC_ALL string known, I just changed to call it and process its result as a safety measure.  I could add a comment that we don't know of any instances where this actually happens, but put it there as a safety measure

khwilliamson:
I'm splitting this PR into two, so I can submit the final slightly controversial final commit later, and merge the remaining uncontroversial ones now, which are blocking other commits in the pipeline from being considered

PR #21341, opened 2023-08-08
locale.c: debug with aTHX

khwilliamson:
This commit automatically adds an aTHX element to debugging statements in this file.  I found this quite useful when debugging locale behavior in threaded perls.  

It would be better to use the thread id, but I don't know how to get that efficiently from C..

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -276,13 +276,14 @@
  * initialization.  This is done before option parsing, and before any thread
  * creation, so can be a file-level static.  (Must come before #including
  * perl.h) */
+#include "config.h"
 #ifdef DEBUGGING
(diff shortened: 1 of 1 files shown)


tonycoz:
> It would be better to use the thread id, but I don't know how to get that efficiently from C..

Yeah, the thread id is internal to threads.xs, to make it available in core we'd need spot for the id in the interpreter struct, which might be worth doing.

My initial impulse was to use the OS/libc thread handle (as returned by $thread->_handle) but that's a large integer (translated from a pointer with glibc) on both Win32 and pthreads.

tonycoz (approved):

PR #21344, opened 2023-08-08
Revamp S_bool_setlocale_2008_i()

khwilliamson:
This changes this function to avoid recursively calling it.  

It turns out that some platforms implement newlocale() in such a way that it can take an LC_ALL string where not all categories are the same, just as setlocale() does.  Other platforms do not allow this.  One commit in this series guesses as to which do, namely those that use the name=value form for this in setlocale().  I haven't run across any that this isn't true for, and at this point in the development cycle, I think its ok to try it this way.  A Configure probe could be written if necessary.


tonycoz (approved):

khwilliamson:
Pushed via a67bd85b3a75060213043b4e663b97272ec33f8a (in order to get a summary commit message)

PR #21347, opened 2023-08-08
Rework initialization of locale handling

khwilliamson:
This code has always been difficult to understand, with a self-modifying loop.  This series of commits changes things to use a switch() statement instead, with ancillary changes as well.

@tonycoz, this code uses casts in order to increment an enum because of that not working without them on C++.  We could define a ++ operator for this enum, as you mentioned in a previous PR.  But that would be our first place where we fairly explicitly account for the core accommodating C++.  I like that idea, in part because I use C++ as my normal development compiler, but the decision needs to be vetted.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4344,6 +4344,11 @@ Ri	|const char *|mortalized_pv_copy				\
 				|NULLOK const char * const pv
 S	|void	|new_LC_ALL	|NULLOK const char *unused		\
 				|bool force
+void
+S	|void	|output_check_environment_warning			\
+				|NULLOK const char * const language	\
+				|NULLOK const char * const lc_all	\
+				|NULLOK const char * const lang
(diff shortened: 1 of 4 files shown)


khwilliamson:
It turns out this PR is premature; #21344 needs to be merged first

jkeenan:
> It turns out this PR is premature; #21344 needs to be merged first

You (manually) merged #21344 yesterday.  Does that mean this p.r. is now reviewable?

khwilliamson:
Yes, it is reviewable

tonycoz:
> C++

I wouldn't mind being able to use C++ natively in core\* but I don't think it will ever happen (beyond the little in Win32 builds.)

\* I suspect using `std::sort()` would improve sort performance, and `std::wstring` would be nice for manipulating paths in `win32/win32.c`.

tonycoz (approved):

PR #21333, opened 2023-08-07
Fix Safe clobbering %SIG

ehuelsmann:
Fixes https://rt.cpan.org/Ticket/Display.html?id=112092

I followed pod/perlhack.pod; please let me know if I need to follow additional process: it's my first submission.

PR #21349, opened 2023-08-08
Update Safe metadata

ehuelsmann:
In PR #21333, I proposed a fix that was reported 7 years ago in RT. At the time of report, Safe was already maintained in the Perl tree for 7 years. As a result, nobody was watching the RT queue for Safe when the reporter of the problem reported it. However the reporter did what CPAN meta data told him to: report to the RT queue.

With the submission of this meta data, I'd like to request one last release to CPAN so as to update the metadata and point to this repository. I asked irc.perl.org/#metacpan if there's another way, but there is no other way than a release...

### DIFF
--- a/dist/Safe/Makefile.PL
+++ b/dist/Safe/Makefile.PL
@@ -7,4 +7,26 @@ WriteMakefile(
     VERSION_FROM => 'Safe.pm',
     INSTALLDIRS => ($] < 5.011 ? 'perl' : 'site'),
     ($core || $] >= 5.011) ? () : (INST_LIB => '$(INST_ARCHLIB)'),
+    AUTHOR => [
+        'Malcolm Beattie',
+        'Perl 5 Porters <perl5-porters@perl.org>'
+    ],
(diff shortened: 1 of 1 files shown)


ehuelsmann:
I feel stupid, but my setup does not generate META.json with this configuration. But since I'm unable to build blead Perl, it might be because it's the Perl that comes with Ubuntu.

haarg:
`META.json` and `META.yml` are generated when using `make distdir` or `make dist`.

ehuelsmann:
Ok. Thanks for the help. `perl Makefile.PL` generated META.yml which confused me into thinking it should have generated META.json too. It's working now.

ehuelsmann:
What - if anything - can I do to expedite this PR?

ehuelsmann:
Is there anything I can do to help this PR forward?

haarg (approved):

ehuelsmann:
Thanks for merging this PR!

The next thing is the meta data to be updated on (Meta)CPAN by uploading a release. Should I open an issue to request that, specifically?

PR #21354, opened 2023-08-09
locale.c: Avoid unnecessary freeing and reallocating

khwilliamson:
Check first if an interpreter variable has actually changed before going to the trouble of freeing its memory and reallocating and populating it.


khwilliamson:
This PR is premature

PR #21358, opened 2023-08-09
variant_under_utf8_count()  Use full words more

khwilliamson:
I'm unsure if this is a good idea or not.

The previous code, when counting UTF-8 variants in a string, processed byte-by-byte up to a word boundary, then switched to doing per-word as long as there is a full word left, then switched back to per-byte for the final partial word.

This commit does the same, except that any partial word at the beginning is processed per-word by reading the whole word in, and masking out the bytes in it that precede the string beginning, so that their contents are not actually looked at (though they are ANDed with 00).  I don't know if this is legal, but both valgrind and asan don't complain about it.

Thus a loop to process per-byte is eliminated, but the per-word loop is exectued an extra time, and it has an extra store to a variable each time through.  I think this is better than the alternative of creating two nearly identical copies of the loop interior.

I would think it should be possible eliminate the final per-byte loop by reading in the final whole word, and masking out any bytes that come after the end of the string.  But when I did this, asan complained about heap buffer overflow.  I think this is spurious, likely based on how they implement their checking.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -1082,41 +1082,103 @@ C<L<perlapi/is_utf8_invariant_string_loc>>,
 PERL_STATIC_INLINE Size_t
 S_variant_under_utf8_count(const U8* const s, const U8* const e)
 {
-    const U8* x = s;
-    Size_t count = 0;
-
     PERL_ARGS_ASSERT_VARIANT_UNDER_UTF8_COUNT;
 
-#  ifndef EBCDIC
+    Size_t count = 0;
 
-    /* Test if the string is long enough to use word-at-a-time.  (Logic is the
-     * same as for is_utf8_invariant_string()) */
+#  ifdef EBCDIC
+
+    const U8* x = s;
+
+#  else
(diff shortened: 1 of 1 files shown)


tonycoz:
> This commit does the same, except that any partial word at the beginning is processed per-word by reading the whole word in, and masking out the bytes in it that precede the string beginning, so that their contents are not actually looked at (though they are ANDed with 00). I don't know if this is legal, but both valgrind and asan don't complain about it.

A char array isn't required to be word aligned, so if `s` points to such an unaligned string this will adjust the starting pointer before the beginning of the string which is UB.

On x86_64, malloc() will align the returned pointer on a 16-byte boundary, so for SvPVX() the rounding down will point `x` at no lower than the address returned by malloc(), which is safe, so valgrind and asan have nothing to complain about.

Passing a pointer to an unaligned char[] array inside a struct is would also be unsafe AFAIK, though I don't know if ASAN would pick it up (valgrind wouldn't).

Issue #21360, opened 2023-08-10
Severe regex/concatenation performance regression on Windows


xenu:
This was originally reported on PerlMonks: https://www.perlmonks.org/index.pl?node_id=11153747

A performance regression introduced in Perl 5.28 causes this script to execute ~20x-40x slower on Windows:

```
use strict;
use warnings;
use feature 'say';
use Time::HiRes 'time';

my $x = 'a' x 1e6;
my $y = 'b' . $x;

my $t = time;
$x =~ s/./$&-/g;
say time - $t;

$t = time;
$y =~ s/./$&-/g;
say time - $t;

say $^V;

```
Here are the results:

```
12.7103209495544
0.850119113922119
v5.32.1

12.7171239852905
0.765337944030762
v5.28.2

0.345196962356567
0.314804077148438
v5.26.3
```

I can confirm the issue is still present in blead.

xenu:
I played with Windows Performance Analyzer a bit and it seems that most of the time is spent inside `Perl_pp_multiconcat` (which was introduced in Perl 5.28):

![wpa](https://github.com/Perl/perl5/assets/352038/ca48ea0d-d082-48f6-871b-5b0cde7a7840)

More specifically, it seems most of the time is spent on reallocations inside `Perl_tmps_grow_p`. And indeed, the following (crude) patch makes the script execute ~5x faster on my machine:

```diff
diff --git a/scope.c b/scope.c
index 28c767f128..4af8c4ffe2 100644
--- a/scope.c
+++ b/scope.c
@@ -231,8 +231,8 @@ Perl_tmps_grow_p(pTHX_ SSize_t ix)
 {
     SSize_t extend_to = ix;
 #ifndef STRESS_REALLOC
-    if (ix - PL_tmps_max < 128)
-        extend_to += (PL_tmps_max < 512) ? 128 : 512;
+    if (ix - PL_tmps_max < 10000)
+        extend_to += 10000;
 #endif
     Renew(PL_tmps_stack, extend_to + 1, SV*);
     PL_tmps_max = extend_to + 1;
```

I suspect the difference between Windows and Linux is caused by the fact that Windows has a slower `realloc`. On Windows it's implemented naively (`malloc` + `memcpy` + `free`), while on Linux it avoids copies thanks to the `mremap` syscall.

Tux:
Not only Windows.  Time::Hires was added in 5.7.3, so I cannot use this test code before that
```
use strict;
use warnings;
[...]

PR #21371, opened 2023-08-13
pp_multiconcat: in Phase 7 (magic SVs) reuse an available targ pad

richardleach:
When the op's own padslot is the targ for the op, using the SV in it for the string being built up, rather than creating a new mortal SV for the purpose, reduces the number of mortal SVs created during operation.

This is of relevance when the op will be called many times over within the same statement, such as when performing repeated concatenations during a substitution operation on a very large string. See GH #21360.

### DIFF
--- a/ext/B/B.pm
+++ b/ext/B/B.pm
@@ -20,7 +20,7 @@ sub import {
 # walkoptree comes from B.xs
 
 BEGIN {
-    $B::VERSION = '1.88';
+    $B::VERSION = '1.89';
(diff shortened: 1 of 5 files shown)


richardleach:
I've added a second commit such that Phase 7 uses an internal "tmps stack". I have not had time to benchmark this yet.

richardleach:
@iabyn properly knows how multiconcat works and is going to look into #21360.

PR #21372, opened 2023-08-14
test.pl: avoid unnecessary work in like_yn()

richardleach:
`like_yn()` underpins `like()` and `unlike()` in _test.pl_, so is a fairly hot subroutine when running perl's test suite.

This commit is a light refactor that does two main things:
* Moves the `$display_got` and `$display_expected` definitions to only the branch where they will actually be used.
* Simplifies the definitions of `$pass`, `$display_got`, and `$display_expected` to make them clearer and more efficient.

### DIFF
--- a/t/test.pl
+++ b/t/test.pl
@@ -467,17 +467,17 @@ sub like_yn ($$$@) {
     # definitely not like(..., '/.../') like
     # Test::Builder::maybe_regex() does.
     unless (re::is_regexp($expected)) {
-	die "PANIC: The value '$expected' isn't a regexp. The like() function needs a qr// pattern, not a string";
+        die "PANIC: The value '$expected' isn't a regexp. The like() function needs a qr// pattern, not a string";
     }
 
     my $pass = ($flip) ? $_[1] !~ /$expected/ : $_[1] =~ /$expected/;
     unless ($pass) {
         my $display_got = display($_[1]);
         my $display_expected = display($expected);
-	unshift(@mess, "#      got '$display_got'\n",
-		$flip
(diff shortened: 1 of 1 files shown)


demerphq (approved):

ilmari (approved):

jkeenan:
In the affected section of `t/test.pl` there is a mixture of spaces and tabs setting indents.
```
sub like_yn ($$$@) {$
    my ($flip, undef, $expected, $name, @mess) = @_;$
$
    # We just accept like(..., qr/.../), not like(..., '...'), and$
    # definitely not like(..., '/.../') like$
    # Test::Builder::maybe_regex() does.$
    unless (re::is_regexp($expected)) {$
^Idie "PANIC: The value '$expected' isn't a regexp. The like() function needs a qr// pattern, not a string";$
    }$
$
    my $pass = ($flip) ? $_[1] !~ /$expected/ : $_[1] =~ /$expected/;$
    unless ($pass) {$
        my $display_got = display($_[1]);$
        my $display_expected = display($expected);$
^Iunshift(@mess, "#      got '$display_got'\n",$
^I^I$flip$
^I^I? "# expected !~ /$display_expected/\n"$
                : "# expected /$display_expected/\n");$
    }$
    local $Level = $Level + 1;$
    _ok($pass, _where(), $name, @mess);$
}$  
```
Would this be a good time to change those tabs to spaces, so that the indentation in the `unless` block is clearly displayed?

richardleach:
> Would this be a good time to change those tabs to spaces, so that the indentation in the `unless` block is clearly displayed?

Sure thing. I'll do this before merging.

PR #21375, opened 2023-08-14
Fix strftime() and create public functions that handle UTF-8ness

khwilliamson:
These fix some bugs and create functions that properly handle UTF8ness.  These return the results in an SV, so the caller doesn't have to think about it.

Time::Piece doesn't currently handle UTF-8 very well.  This will allow it to easily change to be correct.


locale.c
```diff
@@ -5881,48 +5888,68 @@ the program, giving results based on that locale.
...
 
-    /* An empty format yields an empty result */
-    const int fmtlen = strlen(fmt);
-    if (fmtlen == 0) {
-        char *ret;
-        Newxz (ret, 1, char);
-        return ret;
-    }
+STATIC struct tm *
+S_ints_to_tm(pTHX_ int sec, int min, int hour, int mday, int mon, int year,
```
tonycoz: Why return a pointer to an allocated `struct tm`?

Either passing in a pointer to a `struct tm` to initialize, or simply returning a `struct tm` itself (not a pointer) would be more efficient and safer -- no need for memory management in the function nor the caller.

locale.c
```diff
@@ -5881,48 +5888,68 @@ the program, giving results based on that locale.
...
     /* use libc to get the values for tm_gmtoff and tm_zone on platforms that
      * have them [perl #18238] */
-#  if  defined(HAS_MKTIME)                                      \
-   && (defined(HAS_TM_TM_GMTOFF) || defined(HAS_TM_TM_ZONE))
+#if  defined(HAS_MKTIME)                                      \
+ && (defined(HAS_TM_TM_GMTOFF) || defined(HAS_TM_TM_ZONE))
     struct tm mytm2;
-    mytm2 = mytm;
+    StructCopy(mytm, &mytm2, struct tm);
+    mytm2 = *mytm;
```
tonycoz: These two lines do the same thing, though the second is safer.

khwilliamson: Yes; I forgot to delete the first line.  Will fix

locale.c
khwilliamson: I understand your first suggestion which I agree is better than this.  But I don't get the second.  Without a pointer to the caller's struct, I don't see how this function could return a full structure.  If it declared it as an automatic, it would go off the stack upon return.  If it declared it as static, then it would be shared memory that could have races.  If it did a Newx, then memory management is involved

tonycoz: > But I don't get the second.

You can return structures by value:
```
struct tm mybuildtm(int sec, int min, int hour, int mday, int mon, int year,
                   int wday, int yday, int isdst)
{
    struct tm mytm;
    mytm.tm_sec = sec;
    mytm.tm_min = min;
    mytm.tm_hour = hour;
    mytm.tm_mday = mday;
    mytm.tm_mon = mon;
    mytm.tm_year = year;
    mytm.tm_wday = wday;
    mytm.tm_yday = yday;
    mytm.tm_isdst = isdst;
    mktime(&mytm);
    return mytm;
}
```
is perfectly valid code, though in this case the pass-in-a-pointer seems to [produce better code on gcc](https://godbolt.org/z/h8WbW9z36), clang produces much better code and avoids the copy on return.

khwilliamson: If I ever knew that, I'd long forgotten.

pushed with the passed pointer version

locale.c
khwilliamson: fixed

locale.c
```diff
@@ -6031,45 +6104,109 @@ S_strftime_tm(pTHX_ const char *fmt, const struct tm *mytm)
...
-char *
-Perl_my_strftime8_temp(pTHX_ const char *fmt, int sec, int min, int hour, int mday,
-                         int mon, int year, int wday, int yday, int isdst,
-                         utf8ness_t * utf8ness)
-{   /* Documented above */
-    char * retval = my_strftime(fmt, sec, min, hour, mday, mon, year, wday,
-                                yday, isdst);
+STATIC char *
+S_strftime8(pTHX_ const char * fmt,
+                  struct tm * mytm,
```
tonycoz: I don't believe this modifies whatever mytm points at, so this could be `const struct tm *` (which matches the POSIX strftime.

Similarly for `sv_strftime_tm()`

tonycoz:
Looks good except for the const.

khwilliamson:
consts added

PR #21379, opened 2023-08-15
various: timespec/timeval.tv_sec is time_t, not long, according to POSIX

kanavin:
This matters on 32 bit systems configured with 64 bit time_t (so they survive beyond 2038). Casting them to long is causing loss of bits and badly mismatched time data.

### DIFF
--- a/dist/threads-shared/lib/threads/shared.pm
+++ b/dist/threads-shared/lib/threads/shared.pm
@@ -8,7 +8,7 @@ use Config;
 
 use Scalar::Util qw(reftype refaddr blessed);
 
-our $VERSION = '1.68'; # Please update the pod, too.
+our $VERSION = '1.69'; # Please update the pod, too.
 my $XS_VERSION = $VERSION;
 $VERSION = eval $VERSION;
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -72,6 +72,7 @@ Alexander Foken
 Alexander Gernler              <alexander_gernler@genua.de>
 Alexander Gough                <alex-p5p@earth.li>
 Alexander Hartmaier            <abraxxa@cpan.org>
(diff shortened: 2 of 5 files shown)


richardleach:
I've approved the workflow to run, but don't know how portable this change is, since `tv_sec` might still be defined as a long in some platforms.

tonycoz:
I think this is fine.

For maximum consistency it could use `Time_t` instead of `time_t` but there's already plenty of core code that uses `time_t`.

kanavin:
Can you help me understand the single failure?

jkeenan:
> Can you help me understand the single failure?

1. 

> Can you help me understand the single failure?

1. You have made changes to `dist/threads-shared/shared.xs`.  This requires you to increment `$VERSION` in the corresponding `.pm` file, *i.e.*, `dist/threads-shared/lib/threads/shared.pm`.
2. You need to add your name and email address to `AUTHORS`.  I believe you can do this by running `perl Porting/updateAUTHORS.pl`.
After the above steps, run `make test_porting` to confirm that those errors have been corrected, then run the full suite (`make test_harness`) once more, then commit and re-push.

kanavin:
Thanks, I've done those changes.

tonycoz (approved):

dist/threads-shared/lib/threads/shared.pm
```diff
@@ -8,7 +8,7 @@ use Config;
 
 use Scalar::Util qw(reftype refaddr blessed);
 
-our $VERSION = '1.68'; # Please update the pod, too.
+our $VERSION = '1.69'; # Please update the pod, too.
```
bsmith, line 11: The version number is reiterated in the POD on line 1.99, so probably (as per the comment) should be updated to 1.69.

PR #21384, opened 2023-08-17
Set builtin::indexed proto to "@"

JRaspass:
I can't see it making much difference in practice but I feel for consistency if nothing else it should be this.


leonerd (approved):
LGTM

PR #21386, opened 2023-08-17
Add a CI variant for -DPERL_RC_STACK

JRaspass:


demerphq (changes requested):
Please add a brief explanation of what this setting controls and when it can be removed in the future. (In the commit body)

JRaspass:
@demerphq I've updated the commit message saying the purpose of this change and importantly a link back to the PR which introduced the built mode so people can read what it does.

demerphq:
> @demerphq I've updated the commit message

Thanks. One question, maybe better addressed to @iabyn  but he ased that we test

```
-Duseithreads -DDEBUGGING -Accflags='-DPERL_RC_STACK
```

but your patch only tests PERL_RC_STACK by itself. I suspect that @iabyn was deliberate in that request and wants the combination tested as it probably exercizes the code most completely. It wouldn't surprise me if some PERL_RC_STACK testing *only* happens under DEBUGGING. Btw, sorry I didn't notice this before

I am approving this for now as my original concern is resolved but let's give @iabyn some time to follow up before we merge.

.

demerphq (approved):

jkeenan:
Prior to the merge I tested Dave M's branch on FreeBSD, first with:
```
-Accflags=-DPERL_RC_STACK
```
... with and without `-DDEBUGGING`, various other permutations, but always with `-Duseithreads`.  Results here:  https://perl5.test-smoke.org/report/5038202

I then repeated the smoke run of the same branch, but without `-DPERL_RC_STACK`.  Results here:  https://perl5.test-smoke.org/report/5038233

The two runs took exactly the same length of time!  So for our day-to-day CI I'd go simply with:
```
-Duseithreads -DDEBUGGING -Accflags=-DPERL_RC_STACK
```
**Note:** pull request now has merge conflicts.

demerphq:
@iabyn merged e179186c5615abd615cc50f3b6cd0a7210026f54 which is why @JRaspass closed it I guess, as Dave's patch does the same thing.

iabyn:
On Thu, Aug 17, 2023 at 01:46:34PM -0700, Yves Orton wrote:
> @iabyn merged e179186c5615abd615cc50f3b6cd0a7210026f54 which is why
> @JRaspass closed it I guess, as Dave's patch does the same thing.

Yeah, sorry I should have got up to date with my emails etc before
pushing, then I would have spotted this PR.

-- 
Monto Blanco... scorchio!

JRaspass:
No worries, yours was better with `-DDEBUGGING`

PR #21388, opened 2023-08-17
Enforce builtin.pm version bump when builtin.c is modified

ilmari:
universal.c is not listed, since it powers many different modules besides UNIVERSAL.pm, and we can't know which one would need its version bumped when it's modified.


leonerd (approved):
Seems good. Plus I like the structure - it easily shows how to add more in future (e.g. if we get around to adding a `meta.c` for the metaprogramming API)

PR #21390, opened 2023-08-19
OP_EMPTYAVHV: create ops in newANONHASH/newANONLIST rather than rpeep

richardleach:
The current implementation of the OP_EMPTYAVHV optimization does not optimize instances of empty anonymous hashes/lists on the RHS of LOGOP conditions, such as in `my $x = $y || {}`.

This commit addresses this by removing the peephole optimization in favour of having EMPTYAVHV OPs produced directly whenever a call is made to `newANONHASH` or `newANONLIST` with an empty child OP `o`. The standard TARGMY optimization in `S_maybe_targlex` has been slightly adapted to also optimize the `LVINTRO` case for this OP.

This commit also addresses some feedback nits from the last dev cycle: EMPTYAVHV is no longer a LISTOP, now just a BASEOP, and it does not use the OPf_SPECIAL flag. (Both were unnecessary carryovers from the ANONHASH and ANONLIST OPs.)

### DIFF
--- a/op.c
+++ b/op.c
@@ -2440,6 +2440,12 @@ Perl_scalarvoid(pTHX_ OP *arg)
         case OP_SCALAR:
             scalar(o);
             break;
+        case OP_EMPTYAVHV:
+            if (!(o->op_private & OPpTARGET_MY))
+                useless = (o->op_private & OPpEMPTYAVHV_IS_HV) ?
+                           "anonymous hash ({})" :
+                           "anonymous array ([])";
+            break;
         }
 
         if (useless_sv) {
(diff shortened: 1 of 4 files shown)


op.c
```diff
@@ -11661,13 +11667,18 @@ Perl_newFORM(pTHX_ I32 floor, OP *o, OP *block)
...
+    return (o) ? op_convert_list(OP_ANONLIST, OPf_SPECIAL, o)
+               : newOP(OP_EMPTYAVHV, 0);
 }
 
 OP *
 Perl_newANONHASH(pTHX_ OP *o)
 {
-    return op_convert_list(OP_ANONHASH, OPf_SPECIAL, o);
+    OP * anon = (o) ? op_convert_list(OP_ANONHASH, OPf_SPECIAL, o)
+                    : newOP(OP_EMPTYAVHV, OPf_SPECIAL);
```
tonycoz: Is the OPf_SPECIAL needed here, given "and it does not use the OPf_SPECIAL flag." in the commit message?

regen/opcodes
```diff
@@ -261,7 +261,7 @@ list		list			ck_null		m@	L
 lslice		list slice		ck_null		2	H L L
 anonlist	anonymous array ([])	ck_fun		ms@	L
 anonhash	anonymous hash ({})	ck_fun		ms@	L
-emptyavhv	empty anon hash/array	ck_fun		sT@	L
+emptyavhv	empty anon hash/array	ck_fun		sT0	L
```
tonycoz: Does this still take a list argument?

Though I'm not sure it matters anywhere.

op.c
richardleach: Good point, no it doesn't. It was accidentally left there after an experiment.

regen/opcodes
richardleach: I'll remove the argument.

richardleach:
Comments addressed in a squashed commit, rebased and pushed.

tonycoz (approved):

PR #21392, opened 2023-08-19
switch_locale_context: Add aTHX

khwilliamson:
This fixes GH #21040

Instead of a dTHX, this passes aTHX automatically, and skips calling this function if there is no valid context.

It moves that decision into the macro itself, avoiding some #ifdef directives.

And it adds explanation

### DIFF
--- a/dist/threads/threads.xs
+++ b/dist/threads/threads.xs
@@ -193,9 +193,6 @@ S_ithread_set(pTHX_ ithread *thread)
 {
     dMY_CXT;
     MY_CXT.context = thread;
-#ifdef PERL_SET_NON_tTHX_CONTEXT
-    PERL_SET_NON_tTHX_CONTEXT(thread->interp);
-#endif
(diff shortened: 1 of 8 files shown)


dist/threads/threads.xs
```diff
@@ -193,9 +193,6 @@ S_ithread_set(pTHX_ ithread *thread)
 {
     dMY_CXT;
     MY_CXT.context = thread;
-#ifdef PERL_SET_NON_tTHX_CONTEXT
-    PERL_SET_NON_tTHX_CONTEXT(thread->interp);
-#endif
 }
```
jkeenan, line 196: Will a change to this file necessitate an increment to `$VERSION` in `threads.pm`?

khwilliamson: It probably will unless it gets included in this month's development release.

tonycoz (approved):

PR #21394, opened 2023-08-19
     Fix ignoring locale categories

khwilliamson:
There have long been Configure options like
     
        `./Configure -DNO_LOCALE_TIME`
     
     to tell perl to not pay attention to the LC_TIME category.  This feature
     has never fully worked.  This commit fixes it.
     
     One problem is that the when the libc does know about LC_TIME, but perl
     is supposed to ignore that, it didn't have in place the logic to ignore
     it in many circumstances.
     
     This commit creates a boolean array for each locale category known to
     the platform's libc, indicating whether that category is to be ignored
     or not by perl.  If no category is to be ignored, the array and
     ancillary code are not compiled, so perl works like it did before this
     commit.
     
     But if one or more categories are to be ignored, the code comes into
     play.  A problem is that such categories really do exist on the
     platform, and so ignoring them just doesn't work properly.  On a
     platform with positional LC_ALL notation when not all categories have
     the same locale, LC_ALL legally could look something like
     
         `C/de_DE/nl_NL/en_GB/zh_TW.UTF-8/he_IL`
     
     Perl has to be smart enough to know which of those are categories to be
     ignore.  Prior to this commit, it wasn't.  In fact, all ignored
     categories do exist and do have a locale.  Those should all be C.
     Therefore this commit doesn't actually ignore any locales; it instead
     causes perl to refuse to change such ones away from C.
     
     We don't currently have tests for these configurations.  I have
     personally tested a couple hundred different combinations over and over,
     but that's way too much for CI.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4457,6 +4457,7 @@ S	|void	|give_perl_locale_control				\
 S	|parse_LC_ALL_string_return|parse_LC_ALL_string 		\
 				|NN const char *string			\
 				|NN const char **output 		\
+				|const parse_LC_ALL_STRING_action	\
 				|bool always_use_full_array		\
 				|const bool panic_on_error		\
 				|const line_t caller_line
--- a/embed.h
+++ b/embed.h
@@ -1315,7 +1315,7 @@
 #       endif
 #       if defined(LC_ALL)
 #         define give_perl_locale_control(a,b)  S_give_perl_locale_control(aTHX_ a,b)
-#         define parse_LC_ALL_string(a,b,c,d,e) S_parse_LC_ALL_string(aTHX_ a,b,c,d,e)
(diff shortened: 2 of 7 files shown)


embed.fnc
```diff
@@ -4457,6 +4457,7 @@ S	|void	|give_perl_locale_control				\
 S	|parse_LC_ALL_string_return|parse_LC_ALL_string 		\
```
tonycoz, line 4457: From the commit message:
```
 ./Configure -DNO_LOCALE_TIME
```
Should this be:
```
 ./Configure -Accflags=-DNO_LOCALE_TIME
```
?

tonycoz:
Good except for the Configure option.

embed.fnc
khwilliamson: Yes; fixed

Issue #21366, opened 2023-08-11
When embedding Perl in C, the locale is switched to C/ASCII


trygveaa:
**Description**

When embedding Perl in a C program in order to run Perl scripts in it, the locale is set to C/ANSI_X3.4-1968 (ASCII). This causes issues with using non-ascii characters.

This is a regression introduced in commit 7af2d2037375d58e700f9e1b217efb2c4db66133. Before this commit, the chosen locale was kept.

**Steps to Reproduce**

This script is one of the examples of embedding Perl taken from https://perldoc.perl.org/perlembed. The only changes are that it sets the locale first, and prints the current charset before and after loading Perl.

```
#include <EXTERN.h>
#include <locale.h>
#include <langinfo.h>
#include <perl.h>

static PerlInterpreter *my_perl;

int main(int argc, char **argv, char **env) {
  char *current_charset = NULL;
  setlocale(LC_ALL, "");

  current_charset = strdup (nl_langinfo (CODESET));
  printf("charset before loading perl: %s\n", current_charset);

  PERL_SYS_INIT3(&argc, &argv, &env);
  my_perl = perl_alloc();
  perl_construct(my_perl);
  PL_exit_flags |= PERL_EXIT_DESTRUCT_END;
  perl_parse(my_perl, NULL, argc, argv, (char **)NULL);
  perl_run(my_perl);

  current_charset = strdup (nl_langinfo (CODESET));
  printf("charset after loading perl: %s\n", current_charset);

  perl_destruct(my_perl);
  perl_free(my_perl);
  PERL_SYS_TERM();
  exit(EXIT_SUCCESS);
}
```

When ran it prints:

```
charset before loading perl: UTF-8
charset after loading perl: ANSI_X3.4-1968
```

If run with Perl before commit 7af2d2037375d58e700f9e1b217efb2c4db66133, it prints `UTF-8` in both lines.

After running `PERL_SYS_TERM()` the locale is back to `UTF-8` again, but the documentation says that it should only be called once after freeing the last interpreter. My use case (Perl scripts for extending functionality in WeeChat) is having long running Perl scripts that often are kept running for the whole lifetime of the application, so this doesn't help.

**Expected behavior**

That the locale is preserved after loading Perl and running Perl code.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 38 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=5.12.15-arch1-1
    archname=x86_64-linux-thread-multi
[...]

PR #21396, opened 2023-08-19
S_bool_setlocale_2008_i, S_querylocale_2008_i: Add entry assertions

khwilliamson:


tonycoz:
Are these necessarily true in an embedded perl? See #21366 for example.

jkeenan:
> Are these necessarily true in an embedded perl? See #21366 for example.

@khwilliamson, can you comment on @tonycoz's concerns here with respect to an embedded perl?

khwilliamson:
I believe they are true in embedded perl.  I need to go through the code and change a bunch of things that say `#ifdef THREADS` to instead be `ifdef MULTIPLICITY`.  Just now I ran a branch with these enabled on OpenBSD, and they caught the bug (in libc I'm pretty sure).

jkeenan:
> I believe they are true in embedded perl. I need to go through the code and change a bunch of things that say `#ifdef THREADS` to instead be `ifdef MULTIPLICITY`. Just now I ran a branch with these enabled on OpenBSD, and they caught the bug (in libc I'm pretty sure).

@khwilliamson there hasn't been any activity on this pull request in 8 months.  Do you intend to move forward with it?  Thanks.

khwilliamson:
I'm confident these assertions are true, and have helped me find bugs in branches I've smoked that included them.  I believe this branch should be merged

PR #21398, opened 2023-08-19
S_find_locale_from_environment: Handle disparate LC_ALL

khwilliamson:
This PR was split off of https://github.com/Perl/perl5/pull/21340.   There are comments in that that apply to this.

When calling setlocale(foo, ""), there is a standardized result.  If there is an environment variable LC_ALL set, it is used regardless of what 'foo' is.

In general, LC_ALL may be set to a heterogenous set of categories and locales, so this could be equivalently

 setlocale(LC_TIME, "de_DE/C/en_US/zh_TW.UTF-8/he_IL");

Prior to this commit, the code did not handle this possibility. It turns out that on all boxes and shells we've tried, they prohibit the LC_ALL environment variable from being set to anything but a single locale name.  But it's easy to handle this eventuality with an already-existing function whose return must be examined and processed.  And doing so makes this code less prone to future changes in those boxes.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -3244,7 +3244,47 @@ S_find_locale_from_environment(pTHX_ const locale_category_index index)
     /* Use any "LC_ALL" environment variable, as it overrides everything else.
      * */
     if (lc_all && strNE(lc_all, "")) {
-        return lc_all;
+        if (index == LC_ALL_INDEX_) {
+            return lc_all;
+        }
+
+        /* Here, there is an LC_ALL environment variable, but we want the value
+         * for just one category from it.  This is fine if LC_ALL is a single
+         * locale that applies to all categories, but if not, we need just the
+         * component that is for the category the caller is asking for.  It
+         * turns out that no libc/shell that khw knows about permits such a
+         * disparate LC_ALL environment variable, but this easily handles such
+         * a case should it ever arise. */
+        switch (parse_LC_ALL_string(lc_all,
+                                    (const char **) &locale_names,
+                                    no_override,  /* Handled by other code */
+                                    false,    /* Return only [0] if suffices */
+                                    false,      /* Don't panic on error */
+                                    __LINE__))
+        {
+          case invalid:
+            return NULL;
+
+          case no_array:
+            return lc_all;
+
+          case only_element_0:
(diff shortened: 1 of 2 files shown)


jkeenan:
> This PR was split off of #21340. There are comments in that that apply to this.
> 
> When calling setlocale(foo, ""), there is a standardized result. If there is an environment variable LC_ALL set, it is used regardless of what 'foo' is.
> 
> In general, LC_ALL may be set to a heterogenous set of categories and locales, so this could be equivalently
> 
> setlocale(LC_TIME, "de_DE/C/en_US/zh_TW.UTF-8/he_IL");
> 
> Prior to this commit, the code did not handle this possibility. It turns out that on all boxes and shells we've tried, they prohibit the LC_ALL environment variable from being set to anything but a single locale name. But it's easy to handle this eventuality with an already-existing function whose return must be examined and processed. And doing so makes this code less prone to future changes in those boxes.

Can we get a unit test(s) that will exercise the various blocks in this p.r.?

jkeenan:
> > This PR was split off of #21340. There are comments in that that apply to this.
> > When calling setlocale(foo, ""), there is a standardized result. If there is an environment variable LC_ALL set, it is used regardless of what 'foo' is.
> > In general, LC_ALL may be set to a heterogenous set of categories and locales, so this could be equivalently
> > setlocale(LC_TIME, "de_DE/C/en_US/zh_TW.UTF-8/he_IL");
> > Prior to this commit, the code did not handle this possibility. It turns out that on all boxes and shells we've tried, they prohibit the LC_ALL environment variable from being set to anything but a single locale name. But it's easy to handle this eventuality with an already-existing function whose return must be examined and processed. And doing so makes this code less prone to future changes in those boxes.
> 
> Can we get a unit test(s) that will exercise the various blocks in this p.r.?

@khwilliamson, ^^.

khwilliamson:
I figured out how to do a test, and found that glibc on Linux did not handle the case properly.  Therefore I'm closing this request, and will add a comment in the appropriate place in locale.c in case we ever want to revisit this.

khwilliamson:
A patch file containing this is attached
[disparate.txt](https://github.com/user-attachments/files/16826296/disparate.txt)

PR #21403, opened 2023-08-20
`core-cpan-diff`: Add option bundling

rwp0:
To enable `./perl -Ilib ./Porting/core-cpan-diff -ax`

https://perldoc.perl.org/Getopt::Long#bundling-(default:-disabled)

### DIFF
--- a/Porting/core-cpan-diff
+++ b/Porting/core-cpan-diff
@@ -9,7 +9,7 @@ use warnings;
 
 use 5.010;
 
-use Getopt::Long;
+use Getopt::Long qw(:config bundling);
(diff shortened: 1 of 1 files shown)


rwp0:
The command above appears here:

https://metacpan.org/dist/perl/view/Porting/release_managers_guide.pod#dual-life-CPAN-module-synchronisation

**dual-life CPAN module synchronisation** section of RMG

tonycoz (approved):

jkeenan:
Works for the way I customarily run `Porting/core-cpan-diff`.

PR #21405, opened 2023-08-20
Configure: Fix swapped responses for SETLOCALE_ACCEPTS_ANY_LOCALE_NAME

khwilliamson:
Answering y gave the opposite of what was intended; and vice-versa

### DIFF
--- a/Configure
+++ b/Configure
@@ -17452,14 +17452,14 @@ $rm_try
     esac
     case "$d_setlocale" in
         $define)
-            rp="When you set your locale to something besides C$c_utf8_locale, does it do so, or just pretend to?" >&4
+            rp="When you set your locale to something besides C$c_utf8_locale, does it actually do so? (Answer no if it just pretends to set it)" >&4
             dflt=n
             . ./myread
             case "$ans" in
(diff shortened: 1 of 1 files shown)


mauke:
I suggest rewording the question, too. From
```sh
"When you set your locale to something besides C$c_utf8_locale, does it do so, or just pretend to?"
```
to something like
```sh
"When you set your locale to something besides C$c_utf8_locale, does it actually do so (and not just pretend to)?"
```
in order to make it clear what a "yes" answer represents.

tonycoz (approved):

khwilliamson:
I haven't pushed this because mauke found some other issues.

Tux:
What is the state now?

jkeenan:
> I haven't pushed this because mauke found some other issues.

@khwilliamson , @mauke, can we get an update on the state of this merge request?  Thanks.

mauke:
I think the other issues is #21406.

khwilliamson:
This fixes #21406.  I found the other issues through #irc scrollback, and wrote https://github.com/Perl/perl5/issues/21715

Issue #21393, opened 2023-08-19
pp_ctl.c: new build-time warning on non-debugging builds


jkeenan:
A new build-time warning has begun to be emitted from `pp_ctl.c` in **non-debugging** builds.  On Ubuntu Linux 22.04 LTS, using `gcc-11` as the C-compiler, I configured with:
```
sh ./Configure -des -Dusedevel
```
... and built `perl` at b78c98a142.  I recorded the following warning during `make test_prep`:
```
pp_ctl.c: In function ‘S_require_file’:
pp_ctl.c:4454:13: warning: unused variable ‘count’ [-Wunused-variable]
 4454 |         int count = call_sv(PL_hook__require__before, G_SCALAR);
      |             ^~~~~
```
I recorded the same warning when building with `clang14` on FreeBSD-13 in a threaded build.

The statement at which the warning is being emitted was added in the previous development cycle:
```
commit 93f6f9654a81b66c47c07ca982e4c00558bd4159
Author:     Yves Orton <demerphq@gmail.com>
AuthorDate: Mon Dec 19 19:32:03 2022 +0100
Commit:     Yves Orton <demerphq@gmail.com>
CommitDate: Sat Mar 18 20:57:59 2023 +0800

    pp_ctl.c - add support for hooking require.
```
However, until recently it was not emitting any warning!  Further investigation showed that there were recent changes in `pp_ctl.c` just after the statement in question.  Specifically, in 5abffa7ff8a, committed to blead on Aug 16, we see:
```
commit 5abffa7ff8a6cee3e13e061f8332cf567427a21c
Author:     David Mitchell <davem@iabyn.nospamdeletethisbit.com>
AuthorDate: Tue Feb 14 12:57:30 2023 +0000
Commit:     David Mitchell <davem@iabyn.nospamdeletethisbit.com>
CommitDate: Wed Aug 16 17:16:59 2023 +0100

    make RC-stack-aware: require,eval
    
    Update pp_entereval() et al to handle a reference-counted stack
    environment in the presence of PERL_RC_STACK.
    
    But in the presence of PERL_XXX_TMP_NORC, don't actually manipulate
    reference counts yet. This will be turned off in a few commits' time.

diff --git a/pp_ctl.c b/pp_ctl.c
index 62f2b8a8f0..488d337fde 100644
--- a/pp_ctl.c
+++ b/pp_ctl.c

...

@@ -4383,15 +4388,17 @@ S_require_file(pTHX_ SV *sv)

...
[...]

PR #21408, opened 2023-08-21
Smoke me/jkeenan/gh 21393 unused var

jkeenan:
This suppresses the `-Wunused_variable` warning reported in https://github.com/Perl/perl5/issues/21393 -- but since I'm just monkey-patching, please review.

(You may get a porting test failure because (I think) Module-CoreList needs to be updated after today's release.)

### DIFF
--- a/pp_ctl.c
+++ b/pp_ctl.c
@@ -4451,9 +4451,8 @@ S_require_file(pTHX_ SV *sv)
         SAVETMPS;
         PUSHMARK(PL_stack_sp);
         rpp_xpush_1(name_sv); /* always use the object for method calls */
-        int count = call_sv(PL_hook__require__before, G_SCALAR);
+        call_sv(PL_hook__require__before, G_SCALAR);
         SV *rsv = *PL_stack_sp;
-        assert(count == 1); /* scalar context */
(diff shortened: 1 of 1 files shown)


pp_ctl.c
```diff
@@ -4452,6 +4452,7 @@ S_require_file(pTHX_ SV *sv)
         PUSHMARK(PL_stack_sp);
         rpp_xpush_1(name_sv); /* always use the object for method calls */
         int count = call_sv(PL_hook__require__before, G_SCALAR);
```
mauke: While we're at it, the type of `count` should be changed from `int` to `I32` because that's what `call_sv` returns.

iabyn:
Really the count variable and assertion should just be removed, and call_sv() be called without using it's return value. The code I changed had 'if (count && ...)', but call_sv(...,G_SCALAR)  always returns 1. The assertion was just a temporary crutch to check that I wasn't changing behaviour when I removed the 'count > 0' check.

jkeenan:
> Really the count variable and assertion should just be removed, and call_sv() be called without using it's return value. The code I changed had 'if (count && ...)', but call_sv(...,G_SCALAR) always returns 1. The assertion was just a temporary crutch to check that I wasn't changing behaviour when I removed the 'count > 0' check.

See latest version of this pull request.

iabyn:
LGTM

PR #21411, opened 2023-08-21
Add a note to the RMG about how to guess the web-accesible NNTP post URL

leonerd:
I found this step wasn't quite obvious when I made the 5.39.2 release, it needed some guesswork.

### DIFF
--- a/Porting/release_managers_guide.pod
+++ b/Porting/release_managers_guide.pod
@@ -1293,6 +1293,16 @@ You can include the customary link to the release announcement even before your
 message reaches the web-visible archives by looking for the X-List-Archive
 header in your message after receiving it back via perl5-porters.
 
+The exact link in the header won't actually work, but by guessing the year and
+month and adjusting the URL to fit the form of the others you should be able
+to work it out. For example the header
+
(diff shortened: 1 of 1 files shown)


leonerd:
Hm. When I tried this yesterday during the releaes the link didn't work. But it now works fine for me and actually the first one redirects me to the second anyway. Maybe it was just a timing issue?

rwp0:
Why these RMG/POD tests always fail? :-)

tonycoz:
> Hm. When I tried this yesterday during the releaes the link didn't work. But it now works fine for me and actually the first one redirects me to the second anyway. Maybe it was just a timing issue?

Yes, I think there's a delay.  You can also [link via msgid](https://www.nntp.perl.org/group/perl.perl5.porters/;msgid=20230821000139.4d5c71ec@shy.leonerd.org.uk)\*, but like the simple numeric links they aren't available immediately.

\* I use this on perl.develop-help.com to link to the NNTP source for reports

rwp0:
Can this PR be closed then?

jkeenan:
> Can this PR be closed then?

I think it is closable, particularly since it has developed merge conflicts.  But I'll leave it to @leonerd to perform last rites.

PR #21414, opened 2023-08-21
newSTATEOP(): copy cop_features of PL_curcop into returned COP

### LINKED ISSUE #21413: newSTATEOP() does not copy cop_features
`newSTATEOP()` mostly clones the various bits of interpreter state out of `PL_curcop` into a newly-allocated COP and returns it. It doesn't do anything about the `cop_features` field; that field is always zero in newly-returned COPs.

This doesn't appear to *actually* matter in practice, however, because the feature bits prettymuch all only apply at compile-time (enabling things like `say`, `fc`, the new `try` or `class` syntax, etc...) and thus it doesn't hugely matter that they've gone missing. However it is very confusing when debugging certain esoteric issues, that inspecting the values of `PL_curcop` at runtime doesn't truely reflect how the code had been compiled.

leonerd:
This doesn't currently break any tests, nor is it easily possible to write one, because current feature bits only apply at compiletime. But it's still confusing and very hard to debug subtle issues, and this would be necessary if we ever add a featurebit that needs runtime inspection.

Fixes #21413

### DIFF
--- a/op.c
+++ b/op.c
@@ -8550,6 +8550,8 @@ Perl_newSTATEOP(pTHX_ I32 flags, char *label, OP *o)
 #endif
     CopSTASH_set(cop, PL_curstash);
 
+    cop->cop_features = PL_curcop->cop_features;
+
     if (cop->op_type == OP_DBSTATE) {
         /* this line can have a breakpoint - store the cop in IV */
         AV *av = CopFILEAVx(PL_curcop);


tonycoz:
This is already done a few lines above:
```
    CopFEATURES_setfrom(cop, PL_curcop);
```

leonerd:
@tonycoz Ugh, you're absolutely right. I had been doing most of my debugging here reading the v5.36.0 source as that matched what perl I was actually running, and failed to notice that had been added in v5.38.

PR #21416, opened 2023-08-22
Don't try to free an already-freed SV.

Quipyowert2:
This fixes a segfault with Perl v5.39.1-256 compiled from source on WSL2 when running `perl -e 'grep a,@a=b,@a=c'`, and a **double free or corruption (out)** error when running the following script:

`perl -e 'grep%agc=ION2.ljt,s_re_deljt,%agc=Idbdeljt,splitxre_deljt,%agc=Idb-openON2..%xre_deljt,splitxt,d'`

After applying this PR, the double free message changed to `Attempt to free unreferenced scalar: SV 0x56218767b558, Perl interpreter: 0x5621876782a0 at AFL++/output/default/crashes/id:000077,sig:11,src:036114,time:106667371,execs:30357718,op:havoc,rep:4 line 1.`

WSL2 version: openSUSE 15.5
```
WSL version: 1.2.5.0
Kernel version: 5.15.90.1
WSLg version: 1.0.51
MSRDC version: 1.2.3770
Direct3D version: 1.608.2-61064218
DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Windows version: 10.0.19045.3324
```

Perl version:
```
This is perl 5, version 39, subversion 2 (v5.39.2 (v5.39.1-256-g6a90d96b0f)) built for x86_64-linux-thread-multi

Copyright 1987-2023, Larry Wall

Perl may be copied only under the terms of either the Artistic License or the
GNU General Public License, which may be found in the Perl 5 source kit.

Complete documentation for Perl, including FAQ lists, should be found on
this system using "man perl" or "perldoc perl".  If you have access to the
Internet, point your browser at https://www.perl.org/, the Perl Home Page.
```

I was going to check if this crash also happens on Perl 5.38.0, but `perlbrew` said there was no makefile found.

### DIFF
--- a/mg.c
+++ b/mg.c
@@ -544,6 +544,9 @@ Perl_mg_free(pTHX_ SV *sv)
 
     PERL_ARGS_ASSERT_MG_FREE;
 
+    if (SvIS_FREED(sv))
+        return 0;
+
     for (mg = SvMAGIC(sv); mg; mg = moremagic) {
         moremagic = mg->mg_moremagic;
         mg_free_struct(sv, mg);
--- a/sv.c
+++ b/sv.c
@@ -5932,6 +5932,8 @@ S_sv_unmagicext_flags(pTHX_ SV *const sv, const int type, const MGVTBL *vtbl, co
 
     if (SvTYPE(sv) < SVt_PVMG || !SvMAGIC(sv))
         return 0;
+    if (SvIS_FREED(sv))
+        return 0;
     mgp = &(((XPVMG*) SvANY(sv))->xmg_u.xmg_magic);
     for (mg = *mgp; mg; mg = *mgp) {
         const MGVTBL* const virt = mg->mg_virtual;
@@ -6938,6 +6940,9 @@ Perl_sv_clear(pTHX_ SV *const orig_sv)
         {
             U32 arena_index;
             const struct body_details *sv_type_details;
+
+            if (SvIS_FREED(sv))
+                goto get_next_sv;
 
             if (type == SVt_PVHV && HvHasAUX(sv)) {
                 arena_index = HVAUX_ARENA_ROOT_IX;


iabyn:
On Mon, Aug 21, 2023 at 10:34:37PM -0700, Nathan Mills wrote:
> This fixes a segfault with Perl v5.39.1-256 compiled from source on WSL2 when running `perl -e &#39;grep ***@***.******@***.***=c&#39;`, and a **double free or corruption (out)** error when running the following script:
> 
> `perl -e &#39;grep%agc=ION2.ljt,s_re_deljt,%agc=Idbdeljt,splitxre_deljt,%agc=Idb-openON2..%xre_deljt,splitxt,d&#39;`
> 
> After applying this PR, the double free message changed to `Attempt to free unreferenced scalar: SV 0x56218767b558, Perl interpreter: 0x5621876782a0 at AFL++/output/default/crashes/id:000077,sig:11,src:036114,time:106667371,execs:30357718,op:havoc,rep:4 line 1.`
> 

Thanks for the effort, but I don't think this is the right approach to
fixing things.

The bug is that the perl interpreter's stack is not reference-counted, so
sometimes it's possible that a value gets prematurely freed. At that
point, bad things can happen, including error messages about 'freeing a
freed scalar' or similar, or even a SEGV.

The real fix is to make the stack reference-counted, and although that is
a big task, there is an ongoing project to achieve it. The just-released
perl-5.39.2 includes a build option, DPERL_RC_STACK, which does just that.
When I tested your failing code on a perl built under that configuration,
it didn't crash.

Your patch is just detecting some places where things have already gone
wrong with ref counting, and trying to catch the mess. By that time it's
too late already.

-- 

Quipyowert2:
I thought the bug here might have been of the "stack is not reference counted" variety. When I searched on GitHub, I found a bunch of old bug reports, and I wasn't sure if it was actually that kind of bug or not. Reference counting the stack sounds like a much better solution than trying to avoid segfaulting after a scalar has already freed.

These two crashing scripts are from a run of AFL++ which found 113 crashing inputs, 71 of which were the same exact crash (passing the core dump flag `-u` to perl). With the stack being reference counted under that build option, I would expect much fewer crashes. Some of the crashes are of the `sleep` or `1..[some large number]` variety, so I have removed sleep from my local fuzzing dictionary.

Glad to hear that this is fixed with the PERL_RC_STACK compile option. I should've read the latest perldelta before opening this PR. I guess I can run AFL++ again later with the 5.39.2 release with that option enabled and see how few crashing inputs it finds.

PR #21419, opened 2023-08-23
Allow Perl_setlocale to work fully on platforms with positional LC_ALL notation

khwilliamson:
This series of commits  effectively does is to change Perl_setlocale() to return its value in the native format which the libc functions are expecting.

This differs from what it used to return only on platforms which use positional notation and only for LC_ALL when not all categories are set  to the same locale.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -195,7 +195,8 @@
  *      dot.  The macro query_nominal_locale_i() can be used to get the nominal
  *      locale that an external caller would expect, for all categories except
  *      LC_ALL.  For that, you can use the function
- *      S_calculate_LC_ALL_string().
+ *      S_calculate_LC_ALL_string().  Or S_native_querylocale_i() will operate
+ *      on any category.
  *
  * The underlying C API that this implements uses category numbers, hence the
  * code is structured to use '_r' at the API level to convert to indexes, which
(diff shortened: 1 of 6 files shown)


locale.c
```diff
@@ -2683,8 +2692,16 @@ S_calculate_LC_ALL_string(pTHX_ const char ** category_locales_list,
      *  2) NULL, to indicate to use querylocale_i() to get each individual
      *     value.
      *
-     * This function returns a mortalized string containing the locale name(s)
-     * of LC_ALL.
+     * The caller sets 'returning' to
+     *      WANT_MORTAL_TEMP        the function returns the calculated string
+     *                              as a mmortalized temporary, so the caller
```
tonycoz, line 2698: double m "mmortalized"

Not new: "mortal" implies an SV to me, and a mortal SV and a SAVEFREEPV() PV may be released at different times (SAVEFREESV() in perlguts talks about that).

Unfortunately we don't seem to have a short-hand name for this.

locale.c
```diff
@@ -2854,13 +2865,25 @@ S_calculate_LC_ALL_string(pTHX_ const char ** category_locales_list,
...
+
+        /* If returning to PL_setlocale_buf, set up to write directly to it,
+         * being sure it is resized to be large enough */
+        if (returning == WANT_PL_setlocale_buf) {
+            if (total_len > PL_setlocale_bufsize) {
+                PL_setlocale_bufsize = total_len;
+                Renew(PL_setlocale_buf, PL_setlocale_bufsize, char);
+            }
+
+            writable_alias = (char *) PL_setlocale_buf;
```
tonycoz: I think the bug here is that `PL_setlocale_buf` is `const  char *`.

save_to_buffer() gets away with no warning because Copy() and Renew() brokenly cast away the const.

locale.c
khwilliamson: How about WANT_TEMP_PV

locale.c
khwilliamson: I've been looking at the trees here and not the forest.  This doesn't lead to incorrect results.  The goal is to be able to return in a variable that XS code shouldn't bet able to write to, but we have to be able to initialize and manipulate the string before returning.  What do you suggest?

locale.c
tonycoz: That works.

locale.c
tonycoz: They can modify it in the same way you are, by using Copy(), or assigning a newly allocated pointer.

Or by casting as you do.

You can't protect it, I think we're better off not getting any more const unsafe.

khwilliamson:
On 8/27/23 18:33, Tony Cook wrote:
> 
> 
> In locale.c 
> <https://github.com/Perl/perl5/pull/21419#discussion_r1306765065>:
> 
>>  
>           /* In all cases here, there's nothing we create that needs to be freed,
>            * so leave 'free_if_void_return' set to the default 'false'. */
>       }
>       else {  /* Here, not all categories have the same locale */
> -        char * writable_alias;  /* XXX This name and indentation will make
> -                                   sense in a few commits */
> +        char * writable_alias;
> +
> +        /* If returning to PL_setlocale_buf, set up to write directly to it,
> +         * being sure it is resized to be large enough */
> +        if (returning == WANT_PL_setlocale_buf) {
> +            if (total_len > PL_setlocale_bufsize) {
> +                PL_setlocale_bufsize = total_len;
> +                Renew(PL_setlocale_buf, PL_setlocale_bufsize, char);
> +            }
> +
> +            writable_alias = (char *) PL_setlocale_buf;
> 
> They can modify it in the same way you are, by using Copy(), or 
> assigning a newly allocated pointer.
> 
> Or by casting as you do.
> 
> You can't protect it, I think we're better off not getting any more 
> const unsafe.

Please explain further.

Sure the protections are easily worked around, but the point is to lower 
the odds of it being inadvertently modified.

tonycoz:
There's plenty of other state that shouldn't be modified incautiously, we document it and move on.

My ideal would be for perl to be `-Wcast-qual` clean as part of const safeness, and we're far from that, but this only makes that worse.

If you really want to prevent inadvertent access you could hide the fields behind an incomplete struct that's only defined in locale.c, and add APIs for any fields that need read access.

locale.c
```diff
@@ -3823,21 +3823,61 @@ Perl_warn_problematic_locale()
...
      * called just after a change, so uses the actual underlying locale just
      * set, and not the nominal one (should they differ, as they may in
      * LC_NUMERIC). */
 
+#  ifdef LC_ALL
+
+    const char * individ_locales[LC_ALL_INDEX_] = { NULL };
+
+    switch (parse_LC_ALL_string(lc_all,
+                                (const char **) &individ_locales,
```
tonycoz: The `(const char **) &individ_locales` can just be `individ_locales` due to the way arrays decay into pointers.

tonycoz (approved):

PR #21421, opened 2023-08-24
eval_sv: add a G_USEHINTS flag

### LINKED ISSUE #21415: eval_sv() doesn't preserve caller hints/features
The Perl API function `eval_sv()` is conceptually the XS version of the Perl `eval()` function. But it doesn't really match very well, because it doesn't preserve the hints bits or enabled features of its caller; whereas the Perl function does.

`eval_sv()` isn't normally accessible from Perl but we can abuse `XS::APItest` in a built source checkout to find a callable copy of it. See the following examples.

The hints bits include the `strict` flags; these are not preserved:

```
$ ./perl -Ilib -e 'use strict; eval("\$x + \$y") or die $@'
Global symbol "$x" requires explicit package name (did you forget to declare "my $x"?) at (eval 1) line 1.
Global symbol "$y" requires explicit package name (did you forget to declare "my $y"?) at (eval 1) line 1.

$ ./perl -Ilib -MXS::APItest -e 'use strict; eval_pv("print \$x + \$y, \"DONE\"", 0)'
0DONE
```

The features bits include the enabled keywords; these are not preserved either:

```
$ ./perl -Ilib -e 'use feature "say"; eval("say q(HELLO)") or die $@'
HELLO

$ ./perl -Ilib -MXS::APItest -e 'use strict; eval_pv("say q(HELLO)", 0) or die $@'
syntax error at (eval 6) line 1, near "say q(HELLO)"
Execution of (eval 6) aborted due to compilation errors.
```

By comparison, the warnings bits *are* preserved, suggesting that at least this general testing method is valid:

```
$ ./perl -Ilib -MXS::APItest -e 'use warnings; eval_pv("print 123 + undef", 0) or die $@'
Use of uninitialized value in addition (+) at (eval 6) line 1.
123

$ ./perl -Ilib -MXS::APItest -e 'eval_pv("print 123 + undef", 0) or die $@'
123
```

Ultimately, I don't think this behaviour could be changed by default without breaking a *huge* amount of existing code. However, I think it would be useful to provide a new flag of some name I don't have a good suggestion for, to ask it to do this.

tonycoz:
Fixes #21415


rwp0 (approved):

op.c
```diff
@@ -12227,7 +12227,6 @@ Perl_ck_eval(pTHX_ OP *o)
         /* Store a copy of %^H that pp_entereval can pick up. */
         HV *hh = hv_copy_hints_hv(GvHV(PL_hintgv));
         OP *hhop;
-        STOREFEATUREBITSHH(hh);
```
leonerd, line 12230: Hmm, took me a little while to work out why this bit, until I saw that we're now reconstructing the feature bits from the entire hinthash. Is that OK? In general the hh might contain a lot more keys (e.g. other bits of enabled syntax and options), so iterating the whole thing might take a little while, as compared just storing/fetching the UV as we used to.

leonerd (commented):
Overall idea (and name) look fine.

Though I'm unsure about the idea of dropping the `"feature/bits"` hh key in favour of this reconstruction mechanism.

tonycoz:
I've changed this to iterate over the possible keys instead

leonerd:
I'm confused though. Why is all the code necessary to reconstruct the features bits, when you removed the call to the `STOREFEATUREBITSHH()` macro. I would have thought the original scheme of just storing the bitflags in one extra UV in the HH was the way to do this..? Is there something about it that doesn't work here?

ext/XS-APItest/t/call.t
```diff
@@ -340,6 +340,16 @@ for my $fn_type (qw(eval_pv eval_sv call_sv)) {
...
+{
+    use feature "fc";
+    # the XS eval_sv() returns the count of results
+    is(eval_sv('fc("A") eq fc("a"); 1', G_LIST), 0,
+       "don't inherit hints by default (so the eval fails)");
+    is(eval_sv('fc("A") eq fc("a"); 1', G_LIST | G_USEHINTS), 1,
+       "inherit hints when requested (so the eval succeeds)")
+      or diag($@);
+}
+
```
leonerd, line 357: Possibly also wants a test that the `strict` flags get inherited. Maybe check some code like `5 + $notavariable` and check it's 5 without the hints but doesn't even compile with.

leonerd:
Oh, now I read the commit message in the first of the two commits, that explains the situation more.

In that case this is probably fine, though I think adding a `use strict` test would be useful too.

tonycoz:
Unfortunately the XS::APItest eval_sv() wrapper returns the result count rather than the result, so I can't check for 5, but I can check whether a result is returned.

Which I've done.

leonerd (approved):
LGTM

leonerd:
+ don't forget a perldelta

iabyn:
ext/XS-APItest/t/call.t is now producing noise on stderr. I'm not confident enough about what exactly is being tested to just add in a 'my $unused =''  etc.

```
ok 537 - call_sv('d', G_EVAL|G_LIST|G_KEEPERR) - the correct error message
Useless use of string eq in void context at (eval 85) line 1.
ok 538 - don't inherit hints by default (so the eval fails)
ok 539 - inherit hints when requested (so the eval succeeds)
ok 540 - don't inherit hints (strict) by default, so the eval succeeds
Variable "$x" is not imported at (eval 88) line 1.
ok 541 - inherit hints (strict) when requested, so the evail fails
```

tonycoz:
Thanks fixed in e78b109c9815d

regen/feature.pl
```diff
@@ -392,14 +392,7 @@ sub longest {
...
 
-#define STOREFEATUREBITSHH(hh) \\
-  (hv_stores((hh), "feature/bits", newSVuv(PL_compiling.cop_features)))
-
-#define FETCHFEATUREBITSHH(hh)                              \\
-  STMT_START {                                              \\
-      SV **fbsv = hv_fetchs((hh), "feature/bits", FALSE);   \\
-      PL_compiling.cop_features = fbsv ? SvUV(*fbsv) : 0;   \\
-  } STMT_END
+#define FETCHFEATUREBITSHH(hh) S_fetch_feature_bits_hh(aTHX_ (hh))
```
bulk88, line 395: This commit turned 1 `hv_fetch()`, into 26 always failing `hv_fetch()` calls, on a `HV*` with only 1 key/1 HE* in it, with a key name of "CORE/prevailing_version".

regen/feature.pl
```diff
@@ -491,6 +484,49 @@ sub longest {
...
+PERL_STATIC_INLINE void
+S_fetch_feature_bits_hh(pTHX_ HV *hh) {
+    PL_compiling.cop_features = 0;
+
+    const struct perl_feature_bit *fb = PL_feature_bits;
+    while (fb->name) {
+        SV **svp = hv_fetch(hh, fb->name, (I32)fb->namelen, 0);
+        if (svp && SvTRUE(*svp))
+               PL_compiling.cop_features |= fb->mask;
+        ++fb;
```
bulk88, line 524: This while loop has 26 hash key name entries to test/look for. All 26 hash keys, are never found in typical production perl code. The `HV*` called `hh`, has only 1 HE*/key in it total called "CORE/prevailing_version". Pretty bad O(n) problems here with this PR.

PR #21422, opened 2023-08-24
Add newLISTOPn()

leonerd:
A convenient shortcut when constructing optree fragments in C code, making the optree shape easier to read for the lack of machinery noise.

### DIFF
--- a/class.c
+++ b/class.c
@@ -627,16 +627,6 @@ Perl_class_apply_attributes(pTHX_ HV *stash, OP *attrlist)
     op_free(attrlist);
 }
 
-static OP *
-S_newCROAKOP(pTHX_ SV *message)
-{
-    OP *o = newLISTOP(OP_LIST, 0,
-            newOP(OP_PUSHMARK, 0),
-            newSVOP(OP_CONST, 0, message));
-    return op_convert_list(OP_DIE, 0, o);
-}
-#define newCROAKOP(message)  S_newCROAKOP(aTHX_ message)
-
 void
 Perl_class_seal_stash(pTHX_ HV *stash)
 {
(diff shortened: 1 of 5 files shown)


leonerd:
Ideally I wanted the `newLISTOPn()` macro to automatically append the final `NULL` after the `__VA_ARGS__` part, so callers couldn't forget to put it there. But I couldn't see an easy way to achieve that.

Plus I'm not sure if it would confuse people reading the code elsewhere, wondering why that was missing. Too subtle perhaps?

rwp0 (approved):

djerius:
> Ideally I wanted the `newLISTOPn()` macro to automatically append the final `NULL` after the `__VA_ARGS__` part, so callers couldn't forget to put it there. But I couldn't see an easy way to achieve that.
> 
> Plus I'm not sure if it would confuse people reading the code elsewhere, wondering why that was missing. Too subtle perhaps?

I think it would. When I read your example code on the p5p list I immediately wondered where the final NULL was; it's an ingrained behavior; va_args **always** end in a NULL.

druud:
The macro-name could explicitly mention (somehow) that it also appends that NULL, so you could add two names, where the append-NULL one calls the other. Ideally the inner macro could then be flagged as 'deprecated' from the start.

leonerd:
> The macro-name could explicitly mention (somehow) that it also appends that NULL, so you could add two names, where the append-NULL one calls the other. Ideally the inner macro could then be flagged as 'deprecated' from the start.

Hmm. It's possible, but I can't think of a good name for the macro that doesn't basically just append "_NULL" anyway; at which point it's barely any better than just leaving it in the callsite.

tonycoz (approved):

PR #21426, opened 2023-08-25
Implement `no builtin ...`

leonerd:
Defines a new type of pad name entry, a "tombstone", that declares that a name does not exist.

Uses these tombstones to implement lexical unimport, thus ensuring that localized unimport within an inner scope gets correctly undone at end of that inner scope, allowing the name to become visible again in an outer scope.

### DIFF
--- a/pad.c
+++ b/pad.c
@@ -541,8 +541,8 @@ pad (via L<perlapi/pad_alloc>) and
 then stores a name for that entry.  C<name> is adopted and
 becomes the name entry; it must already contain the name
 string.  C<typestash> and C<ourstash> and the C<padadd_STATE>
-flag get added to C<name>.  None of the other
-processing of L<perlapi/pad_add_name_pvn>
+and C<padadd_TOMBSTONE> flags get added to C<name>.
+None of the other processing of L<perlapi/pad_add_name_pvn>
 is done.  Returns the offset of the allocated pad slot.
 
 =cut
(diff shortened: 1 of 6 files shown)


rwp0 (approved):

pad.c
```diff
@@ -982,8 +986,12 @@ Perl_pad_findmy_pvn(pTHX_ const char *namepv, STRLEN namelen, U32 flags)
 
     offset = pad_findlex(namepv, namelen, flags,
                 PL_compcv, PL_cop_seqmax, 1, NULL, &out_pn, &out_flags);
-    if (offset != NOT_IN_PAD)
+    if (offset != NOT_IN_PAD) {
+        /* is this a lexical import that has been deleted? */
+        if (PadnameIsTOMBSTONE(out_pn))
+            return NOT_IN_PAD;
```
richardleach: Just wondering why this check occurs here, rather than in `S_pad_findlex`. Is that latter function essentially doing unnecessary work when `PadnameIsTOMBSTONE` is true and could bail out earlier if the check was implemented there instead?

richardleach:
As far as I understand what's going on, no bugs are apparent to me.

pad.c
leonerd: An excellent question. As it turns out it would be better to put it in `S_pad_findlex` so as not to perform additional outer captures of tombstones or other weirdness.

I've edited it and force-pushed.

richardleach: Cool, I don't have any more comments.

tonycoz (approved):

PR #21428, opened 2023-08-27
Fix leak when smartmatching against a coderef

Leont:
Previously, when smartmatching against a code reference, the return value would be leaked. This fixes that leak.


tonycoz:
Could you add a test?  `t/op/svleak.t` has infrastructure for testing SV leaks.

Leont:
> Could you add a test? t/op/svleak.t has infrastructure for testing SV leaks.

Sure

tonycoz (approved):

PR #21429, opened 2023-08-27
`make-rmg-checklist`: Make script runnable from anywhere

rwp0:
Also correct usage message indicating POD format output which is the case without the `--html` flag

### DIFF
--- a/Porting/make-rmg-checklist
+++ b/Porting/make-rmg-checklist
@@ -1,7 +1,10 @@
 #!/usr/bin/perl
+
 use strict;
 use warnings;
+
 use Getopt::Long qw< :config no_ignore_case >;
+use FindBin qw< $Bin >;
 
 sub pod {
     my $filename = shift;
@@ -26,7 +29,7 @@ sub _help {
     print << "_END_HELP";
 $0 --version VERSION
 
-This script creates a release checklist as a simple HTML document. It accepts
+This script creates a release checklist as a POD or HTML document. It accepts
 the following arguments:
 
   --version     The version you are working on. This will infer the type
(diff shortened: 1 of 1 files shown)


rwp0:
This is to prevent the below cases from happening:

### Executing the script within the directory as the script

```
$ ~/git/perl5/Porting$ perl make-rmg-checklist  --version 5.39.3
Cannot open file (Porting/release_managers_guide.pod): No such file or directory
```

### Executing the script with its full path

```
$ perl ~/git/perl5/Porting/make-rmg-checklist --version 5.39.3
Cannot open file (Porting/release_managers_guide.pod): No such file or directory
```

Where [open](https://github.com/Perl/perl5/blob/e00e0004183c0e8553a72b49251b7382a309e1c5/Porting/make-rmg-checklist#L12C12-L12C12) fails and `perl` dies complaining.

- https://github.com/Perl/perl5/blob/e00e0004183c0e8553a72b49251b7382a309e1c5/Porting/make-rmg-checklist#L12C12-L12C12

Porting/make-rmg-checklist
```diff
@@ -140,7 +143,7 @@ if ($html) {
 
 my $type = _type_from_version($version);
 
-chomp( my @pod_lines = @{ pod('Porting/release_managers_guide.pod') } );
```
rwp0, line 143: I think this was not a good solution (hardcoding the `Porting` string).

Porting/make-rmg-checklist
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/perl
+
 use strict;
 use warnings;
+
 use Getopt::Long qw< :config no_ignore_case >;
+use FindBin qw< $Bin >;
```
rwp0, line 7: `qw<>` follows the existing style above.

rwp0:
I think many other core scripts can leverage FindBin since it's there for us shipped with Perl.

tonycoz (approved):

PR #21433, opened 2023-08-29
perldelta for newLISTOPn()

leonerd:


pod/perldelta.pod
```diff
@@ -27,6 +27,15 @@ here, but most should go in the L</Performance Enhancements> section.
 
 [ List each enhancement as a =head2 entry ]
 
+=head2 Added C<newLISTOPn()>
+
+A new API function has been added that simplifies C (or XS) code that creates
+LISTOP optree fragments.  C<newLISTOPn()> is a variadic function that takes a
```
rwp0: Thinking `C<LISTOP>` could read better in line 33

pod/perldelta.pod
```diff
@@ -27,6 +27,15 @@ here, but most should go in the L</Performance Enhancements> section.
...
+C<NULL>-terminated list of child op pointers, and constructs a new checked
+LISTOP to contain them all.  This is simpler than creating a new plain
```
rwp0: ditto (line 35)

rwp0 (approved):

pod/perldelta.pod
leonerd: I would think so, but it seems common practice in documentation about optree-manipulating functions that the kinds of ops being worked on - BASEOP, UNOP, BINOP, etc... - don't get special formatting.

pod/perldelta.pod
```diff
@@ -27,6 +27,15 @@ here, but most should go in the L</Performance Enhancements> section.
 
 [ List each enhancement as a =head2 entry ]
 
+=head2 Added C<newLISTOPn()>
```
tonycoz: This isn't a core language feature "XXX New core language features go here.  Summarize user-visible core language
enhancements.".

Doesn't this belong under "=head1 Internal Changes", described as "XXX Changes which affect the interface available to C<XS> code go here."?

pod/perldelta.pod
rwp0: Yes, but I think this can be corrected in the general documentation later.

Formatting helps the eyes a lot to separate those if done correctly.

pod/perldelta.pod
leonerd: Ahyes, oops. I put it in the wrong section.
force-pushed.

PR #21442, opened 2023-08-31
pp_(i_)negate: TARGMY & favour-IV/NV optimizations

richardleach:
The `negate` and `i_negate` functions are modified in a series of commits such that they:
* Support the generic TARGMY optimization, such that `op_targ` points directly to the SV to be assigned into, rather than assigning into a PADTMP for subsequent assignment by a different OP.
* Are RC-stack aware.
* Don't always call `S_negate_string` without first checking if the argument even looks like a string.


iabyn:
On Thu, Aug 31, 2023 at 02:54:30PM -0700, Richard Leach wrote:
> The `negate` and `i_negate` functions are modified in a series of commits such that they:
> * Support the generic TARGMY optimization, such that `op_targ` points directly to the SV to be assigned into, rather than assigning into a PADTMP for subsequent assignment by a different OP.
> * Are RC-stack aware.

Ugh, sorry but I've been spending the last few days working on a fairly
massive "make lots of common PP functions RC-stack aware" branch
which I'm about to turn into a PR.

Would you be able to hold off and rebase this once my work is merged?

-- 
O Unicef Clearasil!
Gibberish and Drivel!
    -- "Bored of the Rings"

richardleach:
> Would you be able to hold off and rebase this once my work is merged?

Yeah, that's no problem.

It was still useful for me to get familiar with the new macros.

richardleach:
Just running tests locally before pushing rebased commits. This PR will now do the following for `negate` and `i_negate`:

* Support the generic TARGMY optimization, such that `op_targ` points directly to the SV to be assigned into, rather than assigning into a PADTMP for subsequent assignment by a different OP.
* Don't always call `S_negate_string` without first checking if the argument even looks like a string.

pp.c
```diff
@@ -2669,7 +2669,8 @@ S_negate_string(pTHX)
     }
     else return FALSE;
     SvSETMAGIC(TARG);
-    rpp_replace_1_1(TARG);
+    if (targ != sv)
```
tonycoz: I think this and the others below could use:
```
    if (LIKELY(targ != sv))
```
since I expect:
```
$x = -$x;
```
to be relatively uncommon vs:
```
  foo(-$x);
```
though I could be convinced otherwise with measurements.

richardleach: Yeah, that's fair enough. I'll make those changes.

tonycoz (approved):

PR #21445, opened 2023-09-01
libnet: deal with /etc/hostname containing fqdn

seanyoung:
If /etc/hostname contains the fqdn then domainname() returns "domainname" rather than domainname from /etc/hostname

### DIFF
--- a/cpan/libnet/lib/Net/Domain.pm
+++ b/cpan/libnet/lib/Net/Domain.pm
@@ -107,6 +107,12 @@ sub _hostname {
   $host =~ s/(\A\.+|\.+\Z)//go;
   $host =~ s/\.\.+/\./go;
 
+  # deal with hostname returning fqdn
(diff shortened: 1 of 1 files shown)


haarg:
libnet is CPAN upstream. Patches should be sent to https://github.com/steve-m-hay/perl-libnet

jkeenan:
[Net::Domain](https://metacpan.org/pod/Net::Domain) is part of the [libnet](https://metacpan.org/dist/libnet) distribution on CPAN.  Modification requests should first be filed in libnet's bug tracker, which appears to be [here](https://rt.cpan.org/Dist/Display.html?Name=libnet). Please submit a patch upstream at that location. (@steve-m-hay) Once the upstream maintainer has taken action, we can synch it into the Perl 5 core distribution.

jkeenan:
Closing this p.r. as the fix first needs to be applied upstream on CPAN.

PR #21446, opened 2023-09-01
Implement builtin::stringify

leonerd:
Adds a new function to the `builtin::` space, to wrap `OP_STRINGIFY`.

It's a neater way to request the current oddball behaviours that come out of special rules of `qq($var)` or `join "", $var`. By providing this as its own function it's more convenient and easier to find, and may pave the way to removing some of those specialcase handling rules at some point in the future.

(As discussed in this mailing list thread: https://www.nntp.perl.org/group/perl.perl5.porters/2023/08/msg266989.html)

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -157,6 +157,10 @@ XS(XS_builtin_func1_scalar)
             Perl_pp_is_tainted(aTHX);
             break;
 
+        case OP_STRINGIFY:
+            Perl_pp_stringify(aTHX);
+            break;
+
         default:
             Perl_die(aTHX_ "panic: unhandled opcode %" IVdf
                            " for xs_builtin_func1_scalar()", (IV) ix);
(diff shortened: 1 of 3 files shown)


rwp0 (approved):

rwp0:
Great, but what's the real difference between `"$var"` and `stringify $var`? Is it just the **readability** or is there performance improvements (perhaps it takes a shorter path for stringification underneath the C code)?

Grinnz:
It is generally an identical optree, the difference is a function can be called in some contexts where double quotes can't (ex: `stringify foo->bar($baz + 1)`

iabyn:
Could you merge this *after* my rc_stack2 branch is merged? Your branch should apply cleanly after mine but not vice versa.

demerphq:
What is the difference between this and overload::StrVal()?

Yves

leonerd:
> Could you merge this _after_ my rc_stack2 branch is merged? Your branch should apply cleanly after mine but not vice versa.

@iabyn - I see you've merged #21449 now. Does this unblock mine?

leonerd:
> What is the difference between this and overload::StrVal()?

@demerphq `overload::StrVal()` can see past `'""'` overloading. It's a special weird function.

`builtin::stringify($x)` is exactly the same as the specialcased `"$x"` or `join "", $x` now - it specifically takes that stringify overloading into account if present.

tonycoz (approved):

iabyn:
> Does this unblock mine?
Yes.

PR #21447, opened 2023-09-02
Perldeltas for GH#21351 & GH#21265

richardleach:


pod/perldelta.pod
```diff
@@ -387,7 +387,12 @@ manager will later use a regex to expand these into links.
 
 =item *
 
-XXX
+Declared references can now be used with C<state> variables. [GH #21351]
+
+=item *
+
+Trailing elements in an unshifted and resized array will now always be
```
leonerd: Should `unshift` be `C<...>`-wrapped?

richardleach: I wasn't sure, but it certainly can be.

leonerd (approved):
LGTM

PR #21448, opened 2023-09-02
Make ExtUtils::CBuilder throw an exception on Windows on linker errors

Leont:
It already had this behavior for compiling and on Unix.

The lack of this is causing some very confusing CPAN Testers results because those will fail at runtime (because it can't find a loadable object) instead of at build time.

@ambs


ambs:
@Leont Hellows! It looks good to me. I am not sure how dual-life modules are supported nowadays. Been out of the loop for some time.
Thanks

Leont:
> @Leont Hellows! It looks good to me. I am not sure how dual-life modules are supported nowadays. Been out of the loop for some time.

Nothing has really changed, for an upstream=blead dist like this we merge it into blead and then the maintainer releases it to CPAN.

PR #21449, opened 2023-09-02
Davem/rc stack2

iabyn:
This series of commits "unwraps" most of the common nullary, unary and binary op pp() functions so that, under PERL_RC_STACK  builds, the pp() functions run directly rather than being called via a slow wrapper.

I've gone for the low-hanging fruit, which gives the greatest speed-up from the least effort, so far. I've generally avoided functions in pp_sys.c - they are typically complex in terms of ifdefs to run on different platforms, and the I/O they trigger make the slowdown from the wrapper less noticeable.

I've skipped pp_eq, pp_i_eq, pp_seq and pp_match for now as they have a dependency on the smartmatch op.

Haven't looked at list ops yet.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -552,6 +552,16 @@ Perl_boot_core_builtin(pTHX)
         CV *cv = newXS_flags(builtin->name, builtin->xsub, __FILE__, proto, 0);
         XSANY.any_i32 = builtin->ckval;
 
+        if (   builtin->xsub == &XS_builtin_func1_void
+            || builtin->xsub == &XS_builtin_func1_scalar)
+        {
+            /* these XS functions just call out to the relevant pp()
+             * functions, so they must operate with a reference-counted
+             * stack if the pp() do too.
+             */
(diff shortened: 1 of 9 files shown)


jkeenan:
> This series of commits "unwraps" most of the common nullary, unary and binary op pp() functions so that, under PERL_RC_STACK builds, the pp() functions run directly rather than being called via a slow wrapper.
> 

Can you remind me/us:  What is the best way to call `./Configure` to test the functionality of this branch? Thx

jkeenan:
> > This series of commits "unwraps" most of the common nullary, unary and binary op pp() functions so that, under PERL_RC_STACK builds, the pp() functions run directly rather than being called via a slow wrapper.
> 
> Can you remind me/us: What is the best way to call `./Configure` to test the functionality of this branch? Thx

To answer my own question ... add the following to the configuration:
```
-Accflags='-DPERL_RC_STACK'
```

tonycoz (approved):

PR #21453, opened 2023-09-04
Bump Devel-PPPort and recover patches from repo

atoomic:
Apply to blead multiple patches only available on the D-PPP GitHub repo.

* Bump INT_MAX_PERL to version 5038000
* Fix sv_2pv and sv_2pv_flags for Perl < 5.17.2
* Change github clone URL from git:// to https://
* Teach scanprov about --skip-devels
* Use __REDEFINE__ in more parts/inc/misc locations

### DIFF
--- a/dist/Devel-PPPort/parts/inc/SvPV
+++ b/dist/Devel-PPPort/parts/inc/SvPV
@@ -14,6 +14,7 @@
 __UNDEFINED__
 SvPVbyte
 sv_2pvbyte
+sv_2pv
 sv_2pv_flags
 sv_pvn_force_flags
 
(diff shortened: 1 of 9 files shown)


dist/Devel-PPPort/parts/inc/SvPV
```diff
@@ -82,6 +83,26 @@ __UNDEFINED__  SV_SMAGIC                0
 __UNDEFINED__  SV_HAS_TRAILING_NUL      0
 __UNDEFINED__  SV_COW_SHARED_HASH_KEYS  0
 
+#if { VERSION < 5.7.2 }
+#
+/* Fix sv_2pv for Perl < 5.7.2 */
```
khwilliamson: There needs to be a better comment about what this fixes.  A link to [https://github.com/Dual-Life/Devel-PPPort/issues/231#issuecomment-1685327026](url) is sufficient.

dist/Devel-PPPort/parts/inc/SvPV
```diff
@@ -82,6 +83,26 @@ __UNDEFINED__  SV_SMAGIC                0
...
+
+# ifdef sv_2pv
+#  undef sv_2pv
+# endif
+
+# if defined(__GNUC__) && !defined(PERL_GCC_BRACE_GROUPS_FORBIDDEN)
```
khwilliamson: This should instead be`#if defined(PERL_USE_GCC_BRACE_GROUPS)`

dist/Devel-PPPort/parts/inc/SvPV
```diff
@@ -82,6 +83,26 @@ __UNDEFINED__  SV_SMAGIC                0
...
+  __UNDEFINED__ sv_2pv(sv, lp) ({ SV *_sv_2pv = (sv); STRLEN *_lp_2pv = (lp); _lp_2pv = _lp_2pv ? : &PL_na; SvPOKp(_sv_2pv) ? ((*(_lp_2pv) = SvCUR(_sv_2pv)), SvPVX(_sv_2pv)) : Perl_sv_2pv(aTHX_ _sv_2pv, (_lp_2pv)); })
```
khwilliamson: Symbol names with a leading underscore are reserved for the C++ compiler's use, and this code is supposed to be compilable as C++.  We violate that in many places, but we should not be introducing new such symbols.  perlhacktips has relevant sections: [https://perldoc.perl.org/perlhacktips#Symbol-Names-and-Namespace-Pollution](url) and [https://perldoc.perl.org/perlhacktips#Writing-safer-macros](url)

It is best not to use PL_na, and it isn't necessary with brace groups, so it would be better to declare a local variable, like `dummy`.  The reasons are set out in the safer-macros link.

Without brace groups, PL_na is as good as it can get.

khwilliamson (changes requested):
@pali, there are things for you here.

dist/Devel-PPPort/parts/inc/SvPV
atoomic: comment added

dist/Devel-PPPort/parts/inc/SvPV
atoomic: changed as requested

dist/Devel-PPPort/parts/inc/SvPV
atoomic: Thanks I've adjusted it with your suggestion as
```c
__UNDEFINED__ sv_2pv(sv, lp) ({ SV *_sv_2pv = (sv); STRLEN sv_2pv_dummy_; STRLEN *_lp_2pv = (lp); _lp_2pv = _lp_2pv ? : [&sv_2pv_dummy_](https://www.irccloud.com/irc/ircperl/channel/%26sv_2pv_dummy_); SvPOKp(_sv_2pv) ? ((*(_lp_2pv) = SvCUR(_sv_2pv)), SvPVX(_sv_2pv)) : Perl_sv_2pv(aTHX_ _sv_2pv, (_lp_2pv)); })
```

dist/Devel-PPPort/parts/inc/SvPV
```diff
@@ -90,6 +111,22 @@ __UNDEFINED__  SV_COW_SHARED_HASH_KEYS  0
...
+#elif { VERSION < 5.17.2 }
+
+/* Fix sv_2pv_flags for Perl < 5.17.2 */
+
+# ifdef sv_2pv_flags
+#  undef sv_2pv_flags
+# endif
+
+# if defined(__GNUC__) && !defined(PERL_GCC_BRACE_GROUPS_FORBIDDEN)
+  __UNDEFINED__ sv_2pv_flags(sv, lp, flags) ({ SV *_sv_2pv = (sv); const I32 _flags_2pv = (flags); STRLEN *_lp_2pv = (lp); _lp_2pv = _lp_2pv ? : &PL_na; ((!(_flags_2pv & SV_GMAGIC) || !SvGMAGICAL(_sv_2pv)) && SvPOKp(_sv_2pv)) ? ((*(_lp_2pv) = SvCUR(_sv_2pv)), SvPVX(_sv_2pv)) : Perl_sv_2pv_flags(aTHX_ _sv_2pv, (_lp_2pv), (_flags_2pv)); })
```
khwilliamson: Use the same mechanism as before to avoid PL_na

khwilliamson (changes requested):

dist/Devel-PPPort/parts/inc/SvPV
atoomic: thanks updated `sv_2pv_flags` and `sv_2pv_flags` with the same technique for the `PERL_USE_GCC_BRACE_GROUPS` part
submitted to d-ppp repo via https://github.com/Dual-Life/Devel-PPPort/pull/236

khwilliamson (approved):

PR #21456, opened 2023-09-05
class.c: Minor av microoptimization tweaks

richardleach:
This PR does two things:
* In `pp_initfield`, it removes an unnecessary `av_extend`. (Probably my fault that it didn't get removed in a previous PR.)
* In `Perl_class_wrap_method_body`, it's possible to use the inline `av_push_simple` function rather than `av_push` on the newly created `fieldmap` AV.

### DIFF
--- a/class.c
+++ b/class.c
@@ -78,7 +78,6 @@ PP(pp_initfield)
 
                 av = newAV_alloc_x(count);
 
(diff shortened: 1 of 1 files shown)


leonerd (approved):
LGTM

PR #21457, opened 2023-09-06
class: don't leak the default initialiser ops if they're forbidden

### LINKED ISSUE #20812: feature 'class' appears to leak opslabs

tonycoz:
Previously if forbid_outofblock_ops() here threw an error the ops from defop would leak, including leaking the slab(s) containing those ops.

The other callers to forbid_outofblock_ops() left the ops being checked on the parser stack when performing this check, so the parse stack clean up would release the ops, but the field initaliser code removes the OP from the parse stack before
class_set_field_defop() so the OP and its children leaked.

To prevent that, populate the defop for the field with the supplied ops before calling forbid_outofblock_ops(), then as the stack rewinds class_seal_stash() will check the error count and free the ops.

Fixes #20812

### DIFF
--- a/class.c
+++ b/class.c
@@ -634,8 +634,8 @@ Perl_class_seal_stash(pTHX_ HV *stash)
     assert(HvSTASH_IS_CLASS(stash));
     struct xpvhv_aux *aux = HvAUX(stash);
 
-    /* generate initfields CV */
-    {
+    if (PL_parser->error_count == 0) {
+        /* generate initfields CV */
         I32 floor_ix = PL_savestack_ix;
         SAVEI32(PL_subline);
         save_item(PL_subname);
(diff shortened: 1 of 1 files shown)


class.c
```diff
@@ -795,6 +795,16 @@ Perl_class_seal_stash(pTHX_ HV *stash)
 
         aux->xhv_class_initfields_cv = initfields;
     }
+    else {
+        /* we had errors, clean up and don't populate initfields */
+        PADNAMELIST *fieldnames = aux->xhv_class_fields;
+        for(SSize_t i = 0; fieldnames && i <= PadnamelistMAX(fieldnames); i++) {
```
leonerd: One small thought: usually cleanup happens from the end backwards; so maybe

```c
for(SSize_t i = PadnamelistMAX(fieldnames); i >= 0; i--)
```

would be better here?

tonycoz: That seems reasonable, updated.

leonerd (approved):
LGTM

PR #21463, opened 2023-09-07
`release_managers_guide.pod`: Capitalize section headings

rwp0:
=head3 specifically


rwp0:
Un-titlecased `Dual-Life` as per the suggestion by @tonycoz

PR #21466, opened 2023-09-08
`GetOpt::Std`: Fix `import` methods' indirect syntax

### LINKED ISSUE #21465: `Getopt::Std`: Indirect object syntax in the `import` methods
- https://github.com/Perl/perl5/blob/92ddeac03bb34a9c34a0187468d02fbb0a9cc4dd/lib/Getopt/Std.pm#L149

- https://github.com/Perl/perl5/blob/92ddeac03bb34a9c34a0187468d02fbb0a9cc4dd/lib/Getopt/Std.pm#L307

Should these not be `Getopt::Std->import;` instead?

https://perldoc.perl.org/perlobj#Indirect-Object-Syntax

rwp0:
Also some minor documention, formatting, styling

Fixes https://github.com/Perl/perl5/issues/21465

---

Add some tests for the scoped syntax

### DIFF
--- a/lib/Getopt/Std.pm
+++ b/lib/Getopt/Std.pm
@@ -14,11 +14,11 @@ Getopt::Std - Process single-character switches with switch clustering
     use Getopt::Std;
 
     getopts('oif:');  # -o & -i are boolean flags, -f takes an argument
-		      # Sets $opt_* as a side effect.
-    getopts('oif:', \%opts);  # options as above. Values in %opts
-    getopt('oDI');    # -o, -D & -I take arg.
-                      # Sets $opt_* as a side effect.
-    getopt('oDI', \%opts);    # -o, -D & -I take arg.  Values in %opts
+		      # Sets $opt_* global variables as a side effect
+    getopts('oif:', \my %opts);  # Options as above, values in %opts
+    getopt('oDI');    # -o, -D & -I take arguments
+                      # Sets $opt_* global variables as a side effect
+    getopt('oDI', \my %opts);    # -o, -D & -I take arg, values in %opts
 
 =head1 DESCRIPTION
 
@@ -39,8 +39,8 @@ The C<getopts()> function returns true unless an invalid option was found.
 The C<getopt()> function is similar, but its argument is a string containing
 all switches that take an argument.  If no argument is provided for a switch,
 say, C<y>, the corresponding C<$opt_y> will be set to an undefined value.
-Unspecified switches are silently accepted.  Use of C<getopt()> is not
-recommended.
+Unspecified switches are silently accepted.  Use of C<getopt()> is B<not
+recommended>.
 
 Note that, if your code is running under the recommended C<use strict
 vars> pragma, you will need to declare these package variables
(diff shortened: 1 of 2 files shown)


lib/Getopt/Std.pm
```diff
@@ -14,11 +14,12 @@ Getopt::Std - Process single-character switches with switch clustering
...
 
     getopts('oif:');  # -o & -i are boolean flags, -f takes an argument
-		      # Sets $opt_* as a side effect.
-    getopts('oif:', \%opts);  # options as above. Values in %opts
-    getopt('oDI');    # -o, -D & -I take arg.
-                      # Sets $opt_* as a side effect.
-    getopt('oDI', \%opts);    # -o, -D & -I take arg.  Values in %opts
+		      # Sets $opt_* global variables as a side effect
+    getopts('oif:', \my %opts);  # Options as above, values in %opts
+
```
rwp0: This blank line helps to separate `getopts` and `getopt` visually since `s` is the only difference in their names.

lib/Getopt/Std.pm
```diff
@@ -39,8 +40,8 @@ The C<getopts()> function returns true unless an invalid option was found.
 The C<getopt()> function is similar, but its argument is a string containing
 all switches that take an argument.  If no argument is provided for a switch,
 say, C<y>, the corresponding C<$opt_y> will be set to an undefined value.
-Unspecified switches are silently accepted.  Use of C<getopt()> is not
-recommended.
+Unspecified switches are silently accepted.  Use of C<getopt()> is B<not
```
rwp0, line 42: `getopt` is not recommended, make this bold, so that it's not somewhat hidden at the end.

lib/Getopt/Std.pm
```diff
@@ -146,7 +147,7 @@ sub getopt (;$$) {
     }
     unless (ref $hash) { 
 	local $Exporter::ExportLevel = 1;
-	import Getopt::Std;
+	Getopt::Std->import;
```
rwp0, line 149: I'm still not [very sure](https://perldoc.perl.org/functions/import), it might be due to some magic by the `Exporter`, but its "argument" is a bareword, so looks like the reverse method syntax.

Leont: Yeah this is indirect syntax.

rwp0:
If this is approved I can begin a general work of converting other indirect object notations in the Core distributions/utilities.

lib/Getopt/Std.pm
```diff
@@ -14,11 +14,11 @@ Getopt::Std - Process single-character switches with switch clustering
     use Getopt::Std;
 
     getopts('oif:');  # -o & -i are boolean flags, -f takes an argument
-		      # Sets $opt_* as a side effect.
-    getopts('oif:', \%opts);  # options as above. Values in %opts
-    getopt('oDI');    # -o, -D & -I take arg.
-                      # Sets $opt_* as a side effect.
-    getopt('oDI', \%opts);    # -o, -D & -I take arg.  Values in %opts
+		      # Sets $opt_* global variables as a side effect
+    getopts('oif:', \my %opts);  # Options as above, values in %opts
```
jkeenan, line 18: While recommending `\my %opts` is probably better guidance for a Synopsis than a reference to a global variable, I don't see any tests for that syntax in `lib/Getopt/Std.t`.  Can you add unit tests that exercise this syntax?  (I suspect the p.r. will be good to go thereafter).  Thanks.

jkeenan (changes requested):

lib/Getopt/Std.pm
jkeenan: Looks good.  Unless some new objection emerges, let's merge it into blead tomorrow.

jkeenan (approved):

jkeenan:
> If this is approved I can begin a general work of converting other indirect object notations in the Core distributions/utilities.

Recommend posting to mailing list a list of the distros/utilities which will be part of this effort.  Clear out any objections in advance.

rwp0:
> If this is approved I can begin a general work of converting other indirect object notations in the Core distributions/utilities.

Okay, thanks for the tip!

PR #21473, opened 2023-09-12
Version bundles for `use builtin`

leonerd:
Adds the ability to import an entire collection of builtin functions at once by doing

    use builtin ':5.40';

This will import everything that was considered stable (i.e. not marked experimental) at the given perl version.

Does not (currently) integrate with `use VERSION` syntax but that would be the obvious final step to be added later.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -24,13 +24,12 @@ struct BuiltinFuncDescriptor {
     bool is_experimental;
 };
 
-#define warn_experimental_builtin(name, prefix) S_warn_experimental_builtin(aTHX_ name, prefix)
-static void S_warn_experimental_builtin(pTHX_ const char *name, bool prefix)
+#define warn_experimental_builtin(name) S_warn_experimental_builtin(aTHX_ name)
+static void S_warn_experimental_builtin(pTHX_ const char *name)
 {
     /* diag_listed_as: Built-in function '%s' is experimental */
     Perl_ck_warner_d(aTHX_ packWARN(WARN_EXPERIMENTAL__BUILTIN),
-                     "Built-in function '%s%s' is experimental",
-                     prefix ? "builtin::" : "", name);
+                     "Built-in function 'builtin::%s' is experimental", name);
 }
 
 /* These three utilities might want to live elsewhere to be reused from other
@@ -125,7 +124,7 @@ XS(XS_builtin_func1_scalar)
 
     switch(ix) {
         case OP_IS_BOOL:
-            warn_experimental_builtin(PL_op_name[ix], true);
+            warn_experimental_builtin(PL_op_name[ix]);
             Perl_pp_is_bool(aTHX);
             break;
 
@@ -266,7 +265,7 @@ XS(XS_builtin_export_lexically)
 {
     dXSARGS;
 
-    warn_experimental_builtin("export_lexically", true);
+    warn_experimental_builtin("export_lexically");
 
     if(!PL_compcv)
         Perl_croak(aTHX_
@@ -395,7 +394,7 @@ static OP *ck_builtin_func1(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
     const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
 
     if(builtin->is_experimental)
-        warn_experimental_builtin(builtin->name, false);
+        warn_experimental_builtin(builtin->name);
 
     SV *prototype = newSVpvs("$");
     SAVEFREESV(prototype);
@@ -476,7 +475,7 @@ static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
     const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
 
     if(builtin->is_experimental)
-        warn_experimental_builtin(builtin->name, false);
+        warn_experimental_builtin(builtin->name);
 
     SV *prototype = newSVpvs("@");
     SAVEFREESV(prototype);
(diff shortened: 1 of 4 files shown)


builtin.c
```diff
@@ -491,33 +492,87 @@ static const char builtin_not_recognised[] = "'%" SVf "' is not recognised as a
...
-    { "reftype",    &XS_builtin_func1_scalar, &ck_builtin_func1, OP_REFTYPE,    false },
-    { "ceil",       &XS_builtin_func1_scalar, &ck_builtin_func1, OP_CEIL,       false },
-    { "floor",      &XS_builtin_func1_scalar, &ck_builtin_func1, OP_FLOOR,      false },
-    { "is_tainted", &XS_builtin_func1_scalar, &ck_builtin_func1, OP_IS_TAINTED, false },
-    { "trim",       &XS_builtin_trim,         &ck_builtin_func1, 0,             false },
-    { "stringify",  &XS_builtin_func1_scalar, &ck_builtin_func1, OP_STRINGIFY,  true },
-
-    { "created_as_string", &XS_builtin_created_as_string, &ck_builtin_func1, 0, true },
-    { "created_as_number", &XS_builtin_created_as_number, &ck_builtin_func1, 0, true },
+    { "is_bool",       0, &XS_builtin_func1_scalar, &ck_builtin_func1, OP_IS_BOOL,    true  },
```
ilmari: Instead of `0` and having to have an explicit check in `S_import_builtin_bundle`, could this be a named constant (e.g. `NO_BUNDLE`) that's greater than any valid version (e.g. `INT_MAX`)?

builtin.c
```diff
@@ -535,14 +590,24 @@ XS(XS_builtin_import)
...
 
-        CV *cv = get_cv(SvPV_nolen(fqname), SvUTF8(fqname) ? SVf_UTF8 : 0);
-        if(!cv)
-            Perl_croak(aTHX_ builtin_not_recognised, sym);
+            int want_ver = 1000*vmajor + vminor;
+
+            if(want_ver < 5039 ||
+                    /* round up devel version to next major release; e.g. 5.39 => 5.40 */
+                    want_ver > (1000*PERL_REVISION + PERL_VERSION + (PERL_VERSION % 2)))
+                Perl_croak(aTHX_ "Builtin version bundle %s is not supported by this Perl", sympv);
```
ilmari: For consistency with feature bundles, this should be `"Builtin bundle \"%s\" is not supported by Perl %vd"`.

builtin.c
```diff
@@ -476,7 +476,7 @@ static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
     const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
 
     if(builtin->is_experimental)
-        warn_experimental_builtin(builtin->name, false);
+        warn_experimental_builtin(builtin->name, true);
```
ilmari: All callers pass `true` for the `prefix` parameter now, so it can go.

builtin.c
leonerd: Makes the table wider, but code simpler.
Force-pushed.

builtin.c
leonerd: Fixed, forcepushed.

builtin.c
leonerd: Fixed, forcepushed.

builtin.c
ilmari: I think it should also include the current Perl version, like `feature.pm` does.

rwp0:
> Does not (currently) integrate with use VERSION syntax but that would be the obvious final step to be added later.

Sorry, but what is meant by this sentence?
Any examples to illustrate the point?
Thanks

leonerd:
> > Does not (currently) integrate with use VERSION syntax but that would be the obvious final step to be added later.
> 
> Sorry, but what is meant by this sentence? Any examples to illustrate the point? Thanks

@rwp0 The point is that eventually `use VERSION` will imply a `use builtin` bundle

```perl
use v5.40;  # implies `use builtin ':5.40`
say "ok" if true;
```

builtin.c
```diff
@@ -489,35 +489,93 @@ static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
...
 
+static bool S_parse_version(const char *vstr, int *vmajor, int *vminor)
+{
+    /* Parse a string like "5.35" to yield 5 and 35. Ignores an optional
+     * trailing third component e.g. "5.35.7". Returns false on parse errors.
+     */
+
+    size_t len;
+
+    if(sscanf(vstr, "%d.%d%zn", vmajor, vminor, &len) < 2)
```
richardleach, line 534: Since you know you only want 3 trailing digits max, can the `sscanf` patterns be `%d.%3d%n` and `%3d%n`? That would hopefully allow simplifying to something like this (not tested):

```
static bool S_parse_version(const char *vstr, int *vmajor, int *vminor)
{
    /* Parse a string like "5.35" to yield 5 and 35. Ignores an optional
     * trailing third component e.g. "5.35.7". Returns false on parse errors.
     */

    size_t len;

    if(sscanf(vstr, "%d.%d%zn", vmajor, vminor, &len) < 2)
        return FALSE;

    vstr += len;

    if (!vstr[0])
        return TRUE;
    
    if(vstr[0] == '.') {
        vstr++;

        int _dummy;
        if(sscanf(vstr, "%3d%zn", &_dummy, &len) < 1)
            return FALSE;
        vstr += len;
    }

    if(vstr[0])
        return FALSE;

    return TRUE;
}
```

leonerd:
I've realised that this code will be simpler to integrate with `use VERSION` later if we use the same U16 `(maj << 8) | min` format for version numbers than it uses for version bundles, so I've adjusted it and force-pushed again.

rwp0 (approved):

builtin.c
```diff
@@ -489,35 +492,93 @@ static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
...
+}
+
+#define import_builtin_bundle(ver)  S_import_builtin_bundle(aTHX_ ver)
+static void S_import_builtin_bundle(pTHX_ U16 ver)
+{
+    for(int i = 0; builtins[i].name; i++) {
+        if(builtins[i].since_ver <= ver)
+            import_sym(newSVpvn_flags(builtins[i].name, strlen(builtins[i].name), SVs_TEMP));
+    }
+}
```
richardleach, line 580: This seems like a lot of busywork to take char *s and put them into new mortal SVs in order to:
* Do some expensive/convoluted concatenations (via `Perl_newSVpvf`, compared with manually building concatenated strings from a pointer to `builtins[i].name` within `S_import_sym`), then:
* Just use the char * of one concatenation (`fqname`) in one call (to `get_cv`)
* Pass the other SV to `S_export_lexical`, which calls `pad_add_name_sv`, which calls `pad_add_name_pvn`, which strips out the char * to use.

Instead:
* Could `S_import_builtin_bundle` pass the char * and strlen as arguments to `S_import_sym`
* If there's no realistic danger of `get_cv` and `export_lexical` croaking (there doesn't seem to be), could `S_import_sym` either manually allocate a buffer to use for the first concatenation and to pass to `get_cv`, then reuse the buffer for the second concatenation and to pass to `export_lexical`, or
* If `get_cv` and `export_lexical` could feasibly croak, `S_import_sym` could create one mortal SV, use it for the `get_cv` call, then reuse it for the second concatenation and `export_lexical` call?
* Could `S_export_lexical` be changed to take a char * and length, then call `pad_add_name_pvn` directly?

PS If your answer is "yes, in theory, but I'm busy", I'm happy to properly look at this and submit a follow-up PR if feasible once this PR is merged.

builtin.c
```diff
@@ -488,35 +492,93 @@ static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
...
 
+static bool S_parse_version(const char *vstr, int *vmajor, int *vminor)
+{
+    /* Parse a string like "5.35" to yield 5 and 35. Ignores an optional
+     * trailing third component e.g. "5.35.7". Returns false on parse errors.
+     */
+
+    size_t len;
+
+    if(sscanf(vstr, "%d.%d%zn", vmajor, vminor, &len) < 2)
```
tonycoz, line 534: How liberal is the parsing meant to be?  The following all compile:
```
tony@venus:.../git/perl6$ ./perl -Ilib -e 'use builtin ":5.40.10"'
tony@venus:.../git/perl6$ ./perl -Ilib -e 'use builtin ":   +5.+40"'
tony@venus:.../git/perl6$ ./perl -Ilib -e 'use builtin ":   +5.  +40. -10"'
```

builtin.c
leonerd: @richardleach I've had a go at implementing this, and while it does reduce the churn of temporary SVs, my current attempt goes for passing around `char *`s instead of `SV *`s, which means they forget the length and UTF8 flag. Not a huge deal currently because of the nature of the symbol names we're working on, but it feels like conceptually a backwards step.

Since all these extra temporary SVs only happen once at import time, I'm tempted to suggest it's not a huge deal - already at compiletime there's so many other temporary SVs and optree fragments floating around that eventually get tidied up before runtime, I don't think an extra few here are going to make a huge difference,

I'm tempted to say merge it as it stands, and if anyone feels particularly enthused they could look into reducing it here, but any sort of work of that nature is likely better spent in other places to bigger impact, honestly.

builtin.c
leonerd: Ohyeah, hrmm.. I forgot that `%d` was quite so liberal in its parsing. We should probably tighten that up quite a bit in that case. Shall I work on it here, or merge and do a separate commit for that?

builtin.c
richardleach: Fair enough, thanks for looking into it.

builtin.c
tonycoz: Either works.  I can do the separate commit if you like.

leonerd: If you've an idea on how to make it more strict, I think best to merge this as it stands and you can tighten it up separately.

PR #21476, opened 2023-09-12
locale.c: Change to use enum instead of int

khwilliamson:
Knowing that the domain is just a few values can lead the compiler to generate better code, and to catch programmer errors.

This also creates #defines for the interior of a C 'for' loop header, which I think are more readable.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -442,20 +442,20 @@ S_wsetlocale(const int category, const wchar_t * wlocale)
 #endif  /* WIN32_USE_FAKE_OLD_MINGW_LOCALES */
 
 /* 'for' loop headers to hide the necessary casts */
-#define all_individual_category_indexes(i)                                  \
-    locale_category_index i = (locale_category_index) 0;                    \
-    i < LC_ALL_INDEX_;                                                      \
-    i = (locale_category_index) ((int) i + 1)
+#define for_all_individual_category_indexes(i)                              \
+    for (locale_category_index i = (locale_category_index) 0;               \
+         i < LC_ALL_INDEX_;                                                 \
+         i = (locale_category_index) ((int) i + 1))
 
-#define all_but_0th_individual_category_indexes(i)                          \
-    locale_category_index i = (locale_category_index) 1;                    \
-    i < LC_ALL_INDEX_;                                                      \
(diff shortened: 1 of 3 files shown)


locale.c
```diff
@@ -441,6 +441,22 @@ S_wsetlocale(const int category, const wchar_t * wlocale)
...
+#define all_but_0th_individual_category_indexes(i)                          \
+    locale_category_index i = (locale_category_index) 1;                    \
+    i < LC_ALL_INDEX_;                                                      \
+    i = (locale_category_index) ((int) i + 1)
+
+#define all_category_indexes(i)                                             \
+    locale_category_index i = (locale_category_index) 0;                    \
+    i <= LC_ALL_INDEX_;                                                     \
+    i = (locale_category_index) ((int) i + 1)
+
```
ilmari: This looks to be redundant with `all_individual_category_indexes`, and unused.

locale.c
```diff
@@ -441,6 +441,22 @@ S_wsetlocale(const int category, const wchar_t * wlocale)
 #  endif
 #endif  /* WIN32_USE_FAKE_OLD_MINGW_LOCALES */
 
+/* 'for' loop headers to hide the necessary casts */
```
ilmari, line 444: Since these are only usable inside the control block of a `for` loop, I think it makes more sense to include that in the macro definition, e.g. 

```
#define for_all_individual_category_indexes(i) \
    for (locale_category_index i = … 
```

rwp0 (approved):

ilmari (changes requested):

locale.c
khwilliamson: Maybe the name should be changed to something like `for_every_category`.  The difference is that the upper limit is `<=` which includes `LC_ALL`, whereas the other includes only the individual categories, and the upper limit is `<`.  Future commits in the pipeline use this one

locale.c
khwilliamson: Done

tonycoz (approved):

locale.c
khwilliamson: I'll come up with a better name when the first use of this macro is added

PR #21481, opened 2023-09-13
Add Test2-Suite

leonerd:
This would make `use Test2::V0` possible on core.

Also adds `Term::Table` as that's a dependency.


jkeenan:
> This would make `use Test2::V0` possible on core.

For reference:  There was an extensive discussion of this idea on the perl5-porters mailing list starting at the end of April: 
https://www.nntp.perl.org/group/perl.perl5.porters/2023/04/msg266270.html
> 
> Also adds `Term::Table` as that's a dependency.
> 

Hmm, I've never heard of that module.  I will have to read up on it!

leonerd:
Un-drafted as it now seems to work, so ready for review.

/cc @exodist

exodist:
+1 from me. I am in favor of this. I do not see anything suspicious in the diff.

jkeenan:
> Un-drafted as it now seems to work, so ready for review.
> 

It's been a long time since we added new CPAN libraries to those that we ship under `cpan/` in the Perl 5 core -- so long, in fact, that I cannot name the last new CPAN inclusion in core.  And here we're adding not one but two CPAN distros to core.  The p.r. indicates that 262 files are either newly added or have changes.  It's therefore unrealistic to expect a full-scale code review of this p.r.

Having followed the development of Test2::* at a distance for a decade, I'm inclined to say, "Let's trust @exodist and @leonerd that now is the right time to bring these two distros into core."  But I think we would also benefit from two things:

1. An explanation -- largely for the benefit of people who haven't been following Test2::* closely -- of (a) what the overall objectives of Test2::* are and (b) what parts of the larger Test2::* project are **not** coming into core.

2. My impression is that since the mid-2010s (parts of) Test2::* have become widely used in enterprise testing.  Could we get some **user testimonials** about these modules?  **Note:** I'm not saying we should require such testimonials, but if we had them they would be a confidence builder and we could use them in conference talks and blog posts over the next year.

Leont:
> It's been a long time since we added new CPAN libraries to those that we ship under cpan/ in the Perl 5 core -- so long, in fact, that I cannot name the last new CPAN inclusion in core

Technically it was ExtUtils::PL2Bat, but that doesn't really count since it's really just a refactor of code already in core. Before that the last was probably `experimental` in 5.20. The last large addition was CPAN::Meta and friends in 5.14.

jjatria:
I'm not sure how helpful this will be, but here goes a testimonial.

I started using Test2 when I started at my current job, and I was immediately sold. The diagnostics are definitely an improvement, and the test suite runs faster thanks to preloading, etc. All of that was good. But more than anything else, we found that writing tests was widely seen as more pleasant: this is a common comment we get from new people who join and are exposed to Test2. The checks are more expressive, there are fewer hoops to jump through (`is` is much smarter, for example), and there's a lot of modern tooling around the most common test tasks. The mocking tools, for example, are particularly useful and versatile.

Since then, I've started using it also for my personal projects, and have even written a couple of Test2::Tools modules, so I've had the chance to see that Test2 also makes things better for the developer of _test libraries_. The toolkit for writing tests for the tests is again worth a particular mention.

I'm personally very glad Test2 exists, and I look forward to it being included in core because I think this will allow a lot more programmers to be exposed to it and realise how much better things can be.

leonerd:
Aside from any of the actual advantages that `Test2-Suite` has over previous tools, one fact remains that `Test::More` and friends are a developmental dead-end. Because they've been stagnant for so long it's now basically impossible to add any new features or abilities. Any new things (e.g. refcount-based tests) have either been written as entire new CPAN modules (my `Test::Refcount`), or more lately been added as features in `Test2-Suite` instead.

Bringing `Test2-Suite` into core like this is a way to resolve that deadlock and allow new development on testing functions once more in core perl, so that core and CPAN modules alike can make use of all the new features.

MANIFEST
```diff
@@ -2309,6 +2309,21 @@ cpan/Term-ANSIColor/t/module/true-color.t
 cpan/Term-ANSIColor/t/taint/basic.t
 cpan/Term-Cap/Cap.pm			Perl module supporting termcap usage
 cpan/Term-Cap/test.pl			See if Term::Cap works
+cpan/Term-Table/lib/Term/Table.pm		Term-Table
+cpan/Term-Table/lib/Term/Table/Cell.pm		Term-Table
+cpan/Term-Table/lib/Term/Table/CellStack.pm	Term-Table
+cpan/Term-Table/lib/Term/Table/HashBase.pm	Term-Table
```
tonycoz, line 2315: Shouldn't this be done for the Win32 makefiles too?

Similarly for the second commit.

tonycoz: Urr, wrong chunk

Makefile.SH
```diff
@@ -1418,8 +1418,8 @@ _cleaner2:
 	-rmdir lib/Test2/EventFacet lib/Test2/Event/TAP lib/Test2/Event
 	-rmdir lib/Test2/API/InterceptResult lib/Test2/API lib/Test2
 	-rmdir lib/Test/use lib/Test/Tester lib/Test/Builder/Tester
-	-rmdir lib/Test/Builder/IO lib/Test/Builder lib/Test lib/Term
-	-rmdir lib/TAP/Parser/YAMLish lib/TAP/Parser/SourceHandler
+	-rmdir lib/Test/Builder/IO lib/Test/Builder lib/Test lib/Term/Table
+	-rmdir lib/Term lib/TAP/Parser/YAMLish lib/TAP/Parser/SourceHandler
```
tonycoz: Shouldn't this be in the win32/ makefile too?

leonerd: Yes almost certainly. All of the edits to any files here were made automatically by the `sync-with-cpan` tool. It's possible that has missed some things. I'll have a look and see if I can guess around it manually.

leonerd: Huh, interesting. I don't know why, but both the `win32` Makefile templates contain just a list of the toplevel lib dir directories, not every path inside. Since they already had `$(LIBDIR)\Term` and `$(LIBDIR)\Test2` in the list, it doesn't appear that anything new needs adding there.

tonycoz (approved):

PR #21468, opened 2023-09-12
Enable thread-safe locales on openbsd

khwilliamson:
This had been disabled in the hints file due to bugs, which perl now works around, so reenable.  For example, commit
0f3830f3997cf7ef1531bad26d2e0f13220dd862, "locale.c: Work around some libc POSIX 2008 bugs"

PR #21485, opened 2023-09-15
locale_table.h: Allow perl to compile with -DNO_LOCALE...

khwilliamson:
Previously, there would be a compilation error if any of -DNO_LOCALE_CTYPE, -DNO_LOCALE_NUMERIC, or -DNO_LOCALE_COLLATE were specified in the Configure options.  This is fixed here by fiddling with the preprocessor options.

### DIFF
--- a/locale_table.h
+++ b/locale_table.h
@@ -23,40 +23,50 @@
  */
 
 #ifdef LC_CTYPE
-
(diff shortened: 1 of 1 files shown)


khwilliamson:
This needs to be be applied before #21468 will compile on OpenBSD

PR #21486, opened 2023-09-15
Remove non-null checks before op_free()

ilmari:
The argument to op_free() is declared NULLOK, and the first thing it does is return if the op is null or already freed, so there's no point in callers checking for it.

### DIFF
--- a/class.c
+++ b/class.c
@@ -800,9 +800,7 @@ Perl_class_seal_stash(pTHX_ HV *stash)
         if (fieldnames) {
             for(SSize_t i = PadnamelistMAX(fieldnames); i >= 0 ; i--) {
                 PADNAME *pn = PadnamelistARRAY(fieldnames)[i];
-                OP *valop = PadnameFIELDINFO(pn)->defop;
-                if (valop)
-                    op_free(valop);
+                op_free(PadnameFIELDINFO(pn)->defop);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


richardleach:
I guess the only argument against this is that it's cheaper to test for NULL within a function than calling another function just to test and return. I don't know how common the NULL vs non-NULL cases are in each of these functions to know whether that's much of a concern though.

tonycoz:
I think moving the NULL check to an inline function would resolve that argument.

richardleach:
> I think moving the NULL check to an inline function would resolve that argument.

Sounds good to me.

tonycoz:
I worked on putting such an inline op_free() patch together, then I tried profiling the code, including #21486, op_free():
```
35103768:  871:Perl_op_free(pTHX_ OP *o)
        -:  872:{
        -:  873:    OPCODE type;
 35103768:  874:    OP *top_op = o;
 35103768:  875:    OP *next_op = o;
 35103768:  876:    bool went_up = FALSE; /* whether we reached the current node by
        -:  877:                            following the parent pointer from a child, and
        -:  878:                            so have already seen this node */
        -:  879:
 35103768:  880:    if (!o || o->op_type == OP_FREED)
   651728:  881:        return;
        -:  882:
 34452040:  883:    if (o->op_private & OPpREFCOUNTED) {
        -:  884:        /* if base of tree is refcounted, just decrement */

```
so I don't think it's worth inlining it.

tonycoz (approved):

Issue #21477, opened 2023-09-13
BBC: Blead Breaks Lingua::EN::Inflexion


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.39.3.

-----------------------------------------------------------------

BBC: Blead Breaks Lingua::EN::Inflexion

Please see http://fast-matrix.cpantesters.org/?dist=Lingua::EN::Inflexion

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.39.3:

Configured by cpan at Wed Sep 13 00:00:03 EDT 2023.

Summary of my perl5 (revision 5 version 39 subversion 3) configuration:
  Commit id: 4e1743c01c56b892b7a354ce13b44a570edbff00
  Platform:
    osname=freebsd
    osvers=12.4-release-p5
    archname=amd64-freebsd-thread-multi
    uname='freebsd cjg-freebsd12 12.4-release-p5 freebsd 12.4-release-p5 generic amd64 '
    config_args='-des -Dprefix=/home/cpan/bin/perl -Dscriptdir=/home/cpan/bin/perl/bin -Dusedevel -Duse64bitall -Duseithreads'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    optimize='-O2'
    cppflags='-DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion=''
    gccversion='FreeBSD Clang 13.0.0 (git@github.com:llvm/llvm-project.git llvmorg-13.0.0-0-gd7b669b3a303)'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
[...]

PR #21487, opened 2023-09-15
smartmatch: $obj ~~ $scalar: one item in list cxt

iabyn:
My recent commit, v5.39.2-101-g987819ee34,
    "make RC-stack-aware: unwrap startmatch [sic] and misc"
introduced a bug which broke Lingua::EN::Inflexion.

This was because the code path for handling overloaded objects of the form ($obj ~~ $scalar) was re-pushing the original args onto the stack before calling the overload method, which was unnecessary. This meant that in list context, rather than returning the one-element list (Y/N), it was returning the three-element list ($obj, $scalar, Y/N).

This commit fixes that.

### DIFF
--- a/lib/overload.t
+++ b/lib/overload.t
@@ -71,7 +71,7 @@ package main;
 
 $| = 1;
 BEGIN { require './test.pl'; require './charset_tools.pl' }
-plan tests => 5363;
+plan tests => 5367;
 
 use Scalar::Util qw(tainted);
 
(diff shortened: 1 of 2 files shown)


lib/overload.t
```diff
@@ -3228,3 +3228,29 @@ package RT33789 {
...
+        'eq'  => sub { $_[0][0] eq $_[1] },
+    ;
+
+    my $o = bless ['cat'];
+
+    # smartmatch is deprecated and will be removed in 5.042
+    no warnings 'deprecated';
+
+    my @result = ($o ~~ 'cat');
+    ::is(scalar(@result), 1, "GH #21477: return one arg");
```
mauke, line 3250: Nit: I wouldn't call the return value an "arg".

jkeenan:
> My recent commit, v5.39.2-101-g987819ee34, "make RC-stack-aware: unwrap startmatch [sic] and misc" introduced a bug which broke Lingua::EN::Inflexion.
> 
> This was because the code path for handling overloaded objects of the form ($obj ~~ $scalar) was re-pushing the original args onto the stack before calling the overload method, which was unnecessary. This meant that in list context, rather than returning the one-element list (Y/N), it was returning the three-element list ($obj, $scalar, Y/N).
> 
> This commit fixes that.

This fixes https://github.com/Perl/perl5/issues/21477 for me, so once @iabyn respond's to @mauke's question, this should be good to merge.

Issue #21350, opened 2023-08-08
Peculiar debugger 'list' command behavior: "floating point" line numbers


joemcmahon:
Module: perl5db.pl

**Description**

Accidentally entering a line range with one or more periods on a list command causes the debugger to produce anomalous line numbers

**Steps to Reproduce**

1. `perl -d` any script.
2. At the debugger prompt, enter `l 1.10`.
3. The debugger will list line 1 with the line number 1.10.
4. Enter a bare `l` command. 
5. The debugger will list lines 2 to 11 as 2.10, 3.10,...11.10.
6. if you enter `l 1.2.3`, the debugger will list line 1 as line 1.2.3.
7. A subsequent list command will list lines as 2.2, 2.3,...
8. Entering `l 127.0.0.1` will list line 127 as `127.0.0.1 ...`. 
9. Subsequent `l` commands will _not_ show decimal points, so the logic is detecting that floating point numbers with no decimals can be displayed as integers

**Expected behavior**

The debugger should either reject numbers containing periods on a list command; automatically truncate them to integers (probably the best option); or interpret the first `.` as a comma, then truncate anything after a second `.`,  and print the resulting range of lines.

**Perl configuration**

```
# perl -V output goes here
Summary of my perl5 (revision 5 version 38 subversion 0) configuration:
   
  Platform:
    osname=darwin
    osvers=22.6.0
    archname=darwin-2level
    uname='darwin tekili-li.local 22.6.0 darwin kernel version 22.6.0: wed jul 5 22:22:05 pdt 2023; root:xnu-8796.141.3~6release_arm64_t6000 arm64 '
    config_args='-de -Dprefix=/Users/joemcmahon/perl5/perlbrew/perls/perl-5.38.0 -Aeval:scriptdir=/Users/joemcmahon/perl5/perlbrew/perls/perl-5.38.0/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-fno-common -DPERL_DARWIN -mmacosx-version-min=13.5 -DNO_POSIX_2008_LOCALE -fno-strict-aliasing -pipe -fstack-protector-strong'
    optimize='-O3'
[...]

PR #21488, opened 2023-09-16
gh-21350: [FIX] Make debugger `l` range specs sane

joemcmahon:
The pattern which matches valid line number ranges in the debugger is far too complaisant when it comes to strings it will accept. A number of completely nonsensical values are cheerfully accepted and produce results that range from nothing at all to very odd results indeed:

 - specifying a float for a line number appends the fractional part of the float to ALL line numbers until something else sets the starting line number to an integer again.[1]
 - Strings like `.$.$.$`, `.....`, and `22.$` are all acccepted. Some do nothing, some generate errors.

This tracks back to a catchall line range pattern that was added in the Perl *3* debugger at commit a687059cbaf. This pattern looks as if it were added to eventually implement a number of things, including lines relative to the current line, negative indexes into the magic source code array, variable references, and more. Several of these were implemented elsewhere in the debugger, but the regex was never simplified. This change breaks up that regex and cleans up a few others (including code that internally generates possible negative line numbers).

Changes to the old catchall regex:
 - Remove the conditional match of a leading `-`. Negative line numbers are not permitted.
 - Remove `$` as a valid character in a linespec in the catchall regex. `$scalar` is handled explicitly earlier in the _cmd_l code. There is no other documented/legitimate use for '$' in a list linespec.
 - The linespec character class treated `.`,  and digits, as equally valid characters in a linespec. This led to the linespecs matching floating point numbers, IPv4 addresses, and various nonnumeric nonsense that did not translate to a valid line number. The combined character set was split into either one period (handled already as "the current line") or a series of digits _only_ (a possible line number). This had to be done for both the starting and ending linespec.
 - \A and \z were added to the catchall regex to ensure that it matched the whole of the remaining line, or failed. This prevents things like `...` from being matched as equivalent to `.`.
 - Make sure the 'v' command creates good linespecs.  As previously coded, this could generate linespecs that started with a negative number. Since we've outlawed negative numbers in linespecs, a `v` too close to the start of the file (as occurs in the tests for perl5db.pl) would cause the more-stringent range check to throw an error.  Instead, the 'v' command now checks the generated step back and sets it to 1 if it is less than 1. The existing test is sufficient to confirm that the new code in cmd_v fixes the issue.
 - Verify that float line numbers are illegal: This was the original bug that triggered this change: the original regex accepted floating-point numbers, and because indexing an array in Perl with a float works because of implicit type conversions, and the increment of a float by (integer) 1 doesn't downgrade the float to an int, the line numbers continued to be displayed with the fractional portion of the original number until something intervened with an explicit or implicit integer linespec.  We simply don't permit them to be specified as line numbers now.
 - Use \w for valid variable name characters: The debugger should eventually be UTF-8 compatible, but is not; this is a step along the way to get there. It makes the variable match in the cmd_l reference parsing code match proper Perl variable names, including UTF-8 "word" characters, and now also properly matches regex result variables (`$1`, `$2`, etc.). It explicitly rules out bare \$, which the old regex would match, triggering an error.
 - A new perldb/t test file was added to verify that all the range cleanups work properly, along with new stanzas in the perl5db.pl test suite.

### DIFF
--- a/lib/perl5db.pl
+++ b/lib/perl5db.pl
@@ -532,7 +532,7 @@ BEGIN {
 use vars qw($VERSION $header);
 
 # bump to X.XX in blead, only use X.XX_XX in maint
-$VERSION = '1.77';
+$VERSION = '1.78';
 
 $header = "perl5db.pl version $VERSION";
 
@@ -2609,7 +2609,7 @@ sub _cmd_l_handle_var_name {
 
 sub _cmd_l_handle_subname {
 
-    my $s = $subname;
+    my $s = my $subname = shift;
 
     # De-Perl4.
     $subname =~ s/\'/::/;
@@ -2620,9 +2620,9 @@ sub _cmd_l_handle_subname {
     # Put it in CORE::GLOBAL if t doesn't start with :: and
     # it doesn't live in this package and it lives in CORE::GLOBAL.
     $subname = "CORE::GLOBAL::$s"
-    if not defined &$subname
-        and $s !~ /::/
-        and defined &{"CORE::GLOBAL::$s"};
+        if not defined &$subname
+           and $s !~ /::/
+           and defined &{"CORE::GLOBAL::$s"};
 
     # Put leading '::' names into 'main::'.
     $subname = "main" . $subname if substr( $subname, 0, 2 ) eq "::";
@@ -2691,34 +2691,25 @@ sub _cmd_l_plus {
 }
 
 sub _cmd_l_calc_initial_end_and_i {
-    my ($spec, $start_match, $end_match) = @_;
+    my ($current_line, $start_match, $end_match) = @_;
 
-    # Determine end point; use end of file if not specified.
-    my $end = ( !defined $start_match ) ? $max :
-    ( $end_match ? $end_match : $start_match );
-
-    # Go on to the end, and then stop.
+    my $end = $end_match // $start_match // $max;
+    # Clean up the end spec if needed.
+    $end = $current_line if $end eq '.';
     _minify_to_max(\$end);
 
-    # Determine start line.
-    my $i = $start_match;
-
-    if ($i eq '.') {
-        $i = $spec;
-    }
-
-    $i = _max($i, 1);
-
-    $incr = $end - $i;
+    # Determine the loop start point.
+    my $i = $start_match // 1;
+    $i = $current_line if $i eq '.';
 
     return ($end, $i);
 }
 
 sub _cmd_l_range {
-    my ($spec, $current_line, $start_match, $end_match) = @_;
+    my ($current_line, $start_match, $end_match) = @_;
 
     my ($end, $i) =
-        _cmd_l_calc_initial_end_and_i($spec, $start_match, $end_match);
+        _cmd_l_calc_initial_end_and_i($current_line, $start_match, $end_match);
 
     # If we're running under a client editor, force it to show the lines.
     if ($client_editor) {
@@ -2780,18 +2771,15 @@ sub _cmd_l_range {
 sub _cmd_l_main {
     my $spec = shift;
 
-    # If this is '-something', delete any spaces after the dash.
-    $spec =~ s/\A-\s*\z/-/;
-
     # If the line is '$something', assume this is a scalar containing a
     # line number.
     # Set up for DB::eval() - evaluate in *user* context.
-    if ( my ($var_name) = $spec =~ /\A(\$.*)/s ) {
-        return _cmd_l_handle_var_name($var_name);
+    if ( $spec =~ /\A(\$(?:[0-9]+|[^\W\d]\w*))\z/ ) {
+        return _cmd_l_handle_var_name($spec);
     }
     # l name. Try to find a sub by that name.
     elsif ( ($subname) = $spec =~ /\A([\':A-Za-z_][\':\w]*(?:\[.*\])?)/s ) {
-        return _cmd_l_handle_subname();
+        return _cmd_l_handle_subname($subname);
     }
     # Bare 'l' command.
     elsif ( $spec !~ /\S/ ) {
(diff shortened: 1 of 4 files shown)


lib/perl5db.pl
```diff
@@ -2786,7 +2786,7 @@ sub _cmd_l_main {
     # If the line is '$something', assume this is a scalar containing a
     # line number.
     # Set up for DB::eval() - evaluate in *user* context.
-    if ( my ($var_name) = $spec =~ /\A(\$.*)/s ) {
+    if ( my ($var_name) = $spec =~ /\A(\$[\w]+)\z/s ) {
```
mauke: - Technically you don't need `(` `)` and `$var_name` here. The regex is anchored to match the whole string, so `$var_name` is redundant as a copy of `$spec`.
- `[\w]+` could be written as `\w+`.
- `\w+` matches invalid variable names such as `$2b`. If you want to be more strict, you could use something like `\$(?:[0-9]+|[^\W\d]\w*)`.
- The `/s` flag is redundant now that the regex doesn't contain `.` anymore.

lib/perl5db.pl
```diff
@@ -2786,7 +2786,7 @@ sub _cmd_l_main {
...
         return _cmd_l_handle_var_name($var_name);
```
mauke: This would have to be `return _cmd_l_handle_var_name($spec)` if you decide to get rid of `$var_name` and the capturing group above.

lib/perl5db.pl
```diff
@@ -2802,9 +2802,13 @@ sub _cmd_l_main {
         return _cmd_l_plus($new_start, $new_incr);
     }
     # l start-stop or l start,stop
-    elsif (my ($s, $e) = $spec =~ /^(?:(-?[\d\$\.]+)(?:[-,]([\d\$\.]+))?)?/ ) {
+    elsif (my ($s, $e) = $spec =~ /\A(?:(\.|\d+)(?:[-,](\.|\d+))?)?\z/ ) {
```
mauke: I would suggest adding the `/a` flag to restrict `\d` to match ASCII digits only.

I'm not sure `.` (the current line, presumably?) is handled correctly by `_cmd_l_range`. In particular, this part in `_cmd_l_calc_initial_end_and_i` makes no sense:
```perl
    my $i = $start_match;

    if ($i eq '.') {
        $i = $spec;
    }
```
What does `_cmd_l_calc_initial_end_and_i` need `$spec` for, anyway?

I would've expected the call in `_cmd_l_range` to pass `$current_line` instead, like this:
```perl
    my ($end, $i) =
        _cmd_l_calc_initial_end_and_i($current_line, $start_match, $end_match);
```

Then `_cmd_l_calc_initial_end_and_i` could do:
```perl
    if ($end eq '.') {
        $end = $current_line;
    }
    _minify_to_max(\$end);

    my $i = $start_match // 1;
    if ($i eq '.') {
        $i = $current_line;
    }
```

lib/perl5db/t/gh-21350
```diff
@@ -0,0 +1,3 @@
+#!perl
+
+print "minimal program to allow the debugger to launch\n";
```
mauke, line 3: New files should be added to `MANIFEST`.

lib/perl5db.pl
joemcmahon: I didn't change that because it wasn't critical to getting the patch working; I'd rather address style fixes as a separate patch. Agreed on the `[\w]+`, will change that, and the more-strict regex is definitely better.

lib/perl5db.pl
joemcmahon: I'll leave that for now; I think that's better cleaned up separately.

lib/perl5db.pl
joemcmahon: Yeah, I passed over the `$i = $spec;` since the tests were passing; your suggestion is, I think, better. I will add the `/a`. At some point the debugger really should be able to debug UTF-8 source code, but this patch is not the pace to do it.

lib/perl5db/t/gh-21350
joemcmahon: Definitely, missed that.

Thank you for the review! Very helpful.

lib/perl5db.pl
joemcmahon: Going back, you're right, that's just silly and redundant. Removing `$var_name`.

lib/perl5db.pl
joemcmahon: See above. You're definitely right.

lib/perl5db.pl
joemcmahon: I've slightly rearranged the ordering, and cleaned up that confusing nested tertiary, but I'm switching this to your logic and explicitly passing `$current_line`.

lib/perl5db.pl
```diff
@@ -2691,25 +2691,19 @@ sub _cmd_l_plus {
...
-    # Determine end point; use end of file if not specified.
-    my $end = ( !defined $start_match ) ? $max :
-    ( $end_match ? $end_match : $start_match );
+    my ($spec, $current_line, $start_match, $end_match) = @_;
 
-    # Go on to the end, and then stop.
+              # If we have...      then use...
+    my $end = defined $end_match   ? $end_match
+            : defined $start_match ? $start_match
+            : $max;
```
tonycoz: This could be:
```
    my $end = $end_match // $start_match // $max;
```

lib/perl5db.pl
```diff
@@ -2691,25 +2691,19 @@ sub _cmd_l_plus {
 }
 
 sub _cmd_l_calc_initial_end_and_i {
-    my ($spec, $start_match, $end_match) = @_;
-
-    # Determine end point; use end of file if not specified.
-    my $end = ( !defined $start_match ) ? $max :
-    ( $end_match ? $end_match : $start_match );
+    my ($spec, $current_line, $start_match, $end_match) = @_;
```
tonycoz: `$spec` is no longer used, and if this function no longer needs `$spec` then _cmd_l_range() also no longer needs `$spec` that I can see.

tonycoz:
If you add a "Fixes #someissuenumber" to the commit message (or to the discussion for the PR) then github will close the issue once this is merged.

I don't think the commit message needs to include `gh-21350: [FIX]`, a `Fixes #` in the commit message would let us know the issue being fixed, and the rest of the subject "makes ... sane" implies a fix.

lib/perl5db.pl
joemcmahon: Oh, I like that. Changing it.

joemcmahon:
Thanks, I'll add that as well. (Review much appreciated.)

lib/perl5db.pl
joemcmahon: Yes, definitely doesn't. I've cleaned that up and fixed _cmd_l_subname so it gets passed the subname instead of reaching out into the caller's namespace.

tonycoz (approved):

PR #21489, opened 2023-09-18
Fix eval in package DB

### LINKED ISSUE #11286: closure keeps containing scope alive
Migrated from [rt.perl.org#89544](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=89544) (status was 'resolved')

Searchable as RT89544$

### LINKED ISSUE #19370: eval '$foo' in package DB called from lexical sub can not see closure values or globals
Perl v5.30.0

eval "string" in package DB is supposed to be evaluated in the nearest non-DB context, useful for debug tools.  However when code in package DB is called from a nested lexical ("my") sub, *only* the lexicals in the innermost sub are visible; globals and lexicals in enclosing scopes can no longer be accessed.

(This may seem obscure, but it is important to me because I use some debugging tools which depend on eval in package DB being able to see user-defined lexicals).

The output from the following script is:
```
$w_from_outer = 41;
$x_from_outer = 42;
$y_from_outer = 43;
$w_from_inner = undef;
$x_from_inner = undef;
$y_from_inner = undef;
$z_from_inner = 44;
```
and here is the script:
```
!/usr/bin/perl
use strict; use warnings; use v5.18; use Data::Dumper;

our $w = 41;
my $x = 42;

my ($w_from_outer, $x_from_outer, $y_from_outer, 
    $w_from_inner, $x_from_inner, $y_from_inner, $z_from_inner);

sub outer {
  my $y = 43;

  $w_from_outer = &DB::doeval('$w');
  $x_from_outer = &DB::doeval('$x');
  $y_from_outer = &DB::doeval('$y');

  my sub inner {
    my $z = 44;

    $w_from_inner = &DB::doeval('$w'); # gets undef
    $x_from_inner = &DB::doeval('$x'); # gets undef
    $y_from_inner = &DB::doeval('$y'); # gets undef
    $z_from_inner = &DB::doeval('$z');
  }
  inner();
}
outer();

say Data::Dumper->Dump(
      [
        $w_from_outer, $x_from_outer, $y_from_outer,
        $w_from_inner, $x_from_inner, $y_from_inner, $z_from_inner
      ],
      [qw(
        w_from_outer x_from_outer y_from_outer
        w_from_inner x_from_inner y_from_inner z_from_inner
      )] );
      
package DB;

sub doeval($) {
  return eval $_[0];
}
```

tonycoz:
This largely reverts a0d2bbd5c4 and adds a text for the documented behaviour of eval in package DB.

Fixes #19370.

a0d2bbd5c4 was originally added to fix #11286, but it also broke long documented behaviour as reported in #19370 that's used by the debugger.

The original eval-in-DB was implemented by @iabyn

### DIFF
--- a/cv.h
+++ b/cv.h
@@ -129,7 +129,7 @@ See L<perlguts/Autoloading with XSUBs>.
 #endif
 #define CVf_DYNFILE	0x1000	/* The filename is malloced  */
 #define CVf_AUTOLOAD	0x2000	/* SvPVX contains AUTOLOADed sub name  */
-#define CVf_HASEVAL	0x4000	/* contains string eval  */
+/* 0x4000 previously CVf_HASEVAL  */
 #define CVf_NAMED	0x8000  /* Has a name HEK */
 #define CVf_LEXICAL	0x10000 /* Omit package from name */
 #define CVf_ANONCONST	0x20000 /* :const - create anonconst op */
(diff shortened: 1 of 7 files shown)


iabyn:
The "ignore current lexical scope for evals in package DB" behaviour was actually present before my time, albeit in a slightly borked form. I accidentally broke the behaviour with my "jumbo closure fix", perl-5.8.0-1527-gb5c19bd7c1, then fixed it properly with perl-5.8.0-519-gd819b83ae9 (with that fix, it then completely ignored the current lexical scope, while originally it only partly did)

My feeling is that NULLing CvOUTSIDE() was the wrong fix for leaking closures, and reverting it is the correct thing to do, especially as the restored DB/eval behaviour is documented in perlfunc.

I haven't looked closely, but my gut feeling is that the CvOUTSIDE() leak thing might be fixable by a suitable application of the  CvWEAKOUTSIDE flag, or in the alternative, applying a similar measure.

PR #21490, opened 2023-09-18
Fix two coverity scan reported issues

tonycoz:
1) a loop with an off-by-one ending index

2) uninitialized structure members being passed along to standard library functions

### DIFF
--- a/regcomp_debug.c
+++ b/regcomp_debug.c
@@ -320,7 +320,7 @@ S_regdump_intflags(pTHX_ const char *lead, const U32 flags)
 
     ASSUME(REG_INTFLAGS_NAME_SIZE <= sizeof(flags)*8);
 
-    for (bit=0; bit<=REG_INTFLAGS_NAME_SIZE; bit++) {
+    for (bit=0; bit < REG_INTFLAGS_NAME_SIZE; bit++) {
         if (flags & (1<<bit)) {
             if (!set++ && lead)
                 Perl_re_printf( aTHX_  "%s", lead);
--- a/regcomp_debug.c
+++ b/regcomp_debug.c
@@ -318,7 +318,7 @@ S_regdump_intflags(pTHX_ const char *lead, const U32 flags)
     int bit;
     int set=0;
 
-    ASSUME(REG_INTFLAGS_NAME_SIZE <= sizeof(flags)*8);
+    STATIC_ASSERT_STMT(REG_INTFLAGS_NAME_SIZE <= sizeof(flags) * CHARBITS);
 
     for (bit=0; bit < REG_INTFLAGS_NAME_SIZE; bit++) {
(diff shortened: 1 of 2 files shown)


regcomp_debug.c
```diff
@@ -320,7 +320,7 @@ S_regdump_intflags(pTHX_ const char *lead, const U32 flags)
 
     ASSUME(REG_INTFLAGS_NAME_SIZE <= sizeof(flags)*8);
```
mauke: This line doesn't make sense to me. Both `REG_INTFLAGS_NAME_SIZE` and `sizeof flags` are compile-time constants. There is no need for the compiler to "assume" anything to optimize the code; it already knows all the values involved. This is just an `assert`.

(And for maximum portability, this `8` should be `CHAR_BIT` instead.)

khwilliamson: Off-topic, concerning `CHAR_BIT`.  Without looking, this appears to be defined starting in C99.  We have it defined ourselves in places like pp_pack.c, and there is `CHARBITS`  from Configure, which could be removed and replaced with `CHAR_BIT`

khwilliamson (approved):

regcomp_debug.c
mauke: `CHAR_BIT` is from C89, actually: https://en.cppreference.com/w/c/types/limits

tonycoz: It's not directly related to either of the fixes here, but it is in the spirit of the little changes in this PR.

I've changed that `ASSUME()` and another in the same file to `STATIC_ASSERT_STMT()`.

Perl in general uses `CHARBITS` so I followed along with that rather than using `CHAR_BIT`.

Other uses of ASSUME seem reasonable, though I haven't looked at them in detail.

mauke: I see your commit, but not in this pull request. Did you accidentally commit to the wrong branch?

It looks like there are 3 branches:

1. `Perl:blead`, the target of this PR
2. `tonycoz:cid464357-using-uninitialized`, the source of this PR, which contains the `struct tm` and off-by-one fixes
3. `Perl:cid464357-using-uninitialized`, which is like [#2](https://github.com/tonycoz/perl5/tree/cid464357-using-uninitialized) but also contains the `STATIC_ASSERT_STMT` changes

[Branch #3](https://github.com/Perl/perl5/tree/cid464357-using-uninitialized) is unrelated to this PR because it exists in the Perl repository, not your fork of it (like [branch #2](https://github.com/tonycoz/perl5/tree/cid464357-using-uninitialized)).

tonycoz: Doh,  it's not the first time I've made that mistake when pushing from `git gui`.  Thanks and fixed.

mauke (approved):

PR #21496, opened 2023-09-19
Remove pointless use-after-free assert

ilmari:
While op_free() doesn't actually free the op, but adds it to the slab's freelist and marks it as freed, that's not obvious when reading the code. Also, if op_free() fails to do its job, we have bigger problems than just here.


leonerd (approved):
LGTM

I can't remember what I was thinking when I wrote that, but you're right it's entirely pointless here.

PR #21498, opened 2023-09-19
`t`: Convert indirect `Config` syntax in core's tests

rwp0:
We have a lot of lines throughout the distribution with this particular indirect object notation.

This PR deals with the tests in `t/` directory which includes core files only.

Convert:

```perl
require Config; import Config;
```

To:

```perl
require Config; Config->import;
```

---

Two test files are excluded in this PR:

- `t/op/lc.t`: Unicode encoding issues 
- `t/re/bigfuzzy.t`: Binary data issues

### DIFF
--- a/t/TEST
+++ b/t/TEST
@@ -938,7 +938,7 @@ SHRDLU_1
 ###   ./perl harness
 ### in the 't' directory since most (>=80%) of the tests succeeded.
 SHRDLU_2
-	if (eval {require Config; import Config; 1}) {
+	if (eval {require Config; Config->import; 1}) {
 	    if ($::Config{usedl} && (my $p = $::Config{ldlibpthname})) {
 		warn <<SHRDLU_3;
 ### You may have to set your dynamic library search path,
--- a/t/io/errnosig.t
+++ b/t/io/errnosig.t
@@ -6,7 +6,7 @@ BEGIN {
     set_up_inc( qw(. ../lib) );
 }
 
-require Config; import Config;
+require Config; Config->import;
 plan(tests => 1);
 
 SKIP: {
--- a/t/io/msg.t
+++ b/t/io/msg.t
@@ -5,7 +5,7 @@ BEGIN {
 
   require "./test.pl";
   set_up_inc( '../lib' ) if -d '../lib' && -d '../ext';
-  require Config; import Config;
+  require Config; Config->import;
 
   if ($ENV{'PERL_CORE'} && $Config{'extensions'} !~ m[\bIPC/SysV\b]) {
     skip_all('-- IPC::SysV was not built');
--- a/t/io/perlio.t
+++ b/t/io/perlio.t
@@ -2,7 +2,7 @@ BEGIN {
 	chdir 't' if -d 't';
     require './test.pl';
     set_up_inc('../lib');
-	require Config; import Config;
+	require Config; Config->import;
 	skip_all_without_perlio();
 }
 
--- a/t/io/pipe.t
+++ b/t/io/pipe.t
@@ -4,7 +4,7 @@ BEGIN {
     chdir 't' if -d 't';
     require './test.pl';
     set_up_inc('../lib');
-    require Config; import Config;
+    require Config; Config->import;
 }
 if (!$Config{'d_fork'}) {
     skip_all("fork required to pipe");
--- a/t/io/sem.t
+++ b/t/io/sem.t
@@ -5,7 +5,7 @@ BEGIN {
 
   require "./test.pl";
   set_up_inc( '../lib' ) if -d '../lib' && -d '../ext';
-  require Config; import Config;
+  require Config; Config->import;
 
   if ($ENV{'PERL_CORE'} && $Config{'extensions'} !~ m[\bIPC/SysV\b]) {
     skip_all('-- IPC::SysV was not built');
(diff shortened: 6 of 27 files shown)


rwp0:
I wonder if using `use Config;` is a better idea than running two statements in tests (except in `eval`).

Leont:
> I wonder if using use Config; is a better idea than running two statements in tests.

Making it compile-time would defeat the point of putting it in an eval or running it after `@INC` has been set up

rwp0:
> > I wonder if using use Config; is a better idea than running two statements in tests.
> 
> Making it compile-time would defeat the point of putting it in an eval or running it after `@INC` has been set up

Thanks. I think it's good to go then!

nicomen:
It would be interesting if the various approaches could be made more similar, like should it always be in a BEGIN block, or should it be bundled together with require of test.pl and setting lib somehow.

t/op/lc.t
```diff
@@ -212,12 +212,12 @@ for ("a\x{100}", "xyz\x{100}") {
...
     is(substr(lc($_), 0), lc($_), "[perl #38619] lc");
 }
-for ("a\x{100}", "�yz\x{100}") { # � to Ss (different length)
+for ("a\x{100}", "ßyz\x{100}") { # ß to Ss (different length)
     is(substr(ucfirst($_), 0), ucfirst($_), "[perl #38619] ucfirst");
 }
 
 #fc() didn't exist back then, but coverage is coverage.
-for ("a\x{100}", "�yz\x{100}", "xyz\x{100}", "XYZ\x{100}") { # � to Ss (different length)
+for ("a\x{100}", "ßyz\x{100}", "xyz\x{100}", "XYZ\x{100}") { # ß to Ss (different length)
```
tonycoz: It looks like you've changed the encoding of this file from latin-1 to utf-8.

This makes the test not test what it's supposed to be testing - that the change in length from ucfirst of "ß" is properly handled, since perl is (as per default) reading the file as Latin-1, with the change it's reading `\xC3\x9F` instead of `\xBF`.

(github's side-by-side diff was really confusing for this.)

rwp0:
> It would be interesting if the various approaches could be made more similar, like should it always be in a BEGIN block, or should it be bundled together with require of test.pl and setting lib somehow.



> BEGIN

Yes, thanks for the comment, though that's a matter of another PR, I'd like to limit this to reverting the indirect syntax only🙂

jkeenan:
@rwp0, GitHub diff does not display the differences between blead and your branch with respect to `
t/re/bigfuzzy_not_utf8.t` because it says it's a binary file.

Command-line `diff` also identifies this file as a binary file and says it differs from the version of that file in blead.
```
[perl2] 2066 $ diff ../perl/t/re/bigfuzzy_not_utf8.t ./t/re/bigfuzzy_not_utf8.t
Binary files ../perl/t/re/bigfuzzy_not_utf8.t and ./t/re/bigfuzzy_not_utf8.t differ
```
(In the above, blead is checked out into `../perl` and your branch into `./perl2` and rebased it on blead.)

But are the differences limited to just your substitution of `Config->import` for `import Config`?  When I examine the two files with `vimdiff`, I see that line 37 in blead has been broken into two lines in your branch.  According to `wc -l`, the file is 39 lines long in blead but 40 in your branch.  Per [superuser](https://superuser.com/questions/125376/how-do-i-compare-binary-files-in-linux#125408), you can also get a diff of the hexdumps of the two versions of the file with something like this:
```
$ diff <(xxd blead.bigfuzzy_not_utf8.t) <(xxd branch.bigfuzzy_not_utf8.t)
```
So before merging we have to get bigfuzzy to contain only your intended changes.  Can you investigate?

rwp0:
> @rwp0, GitHub diff does not display the differences between blead and your branch with respect to ` t/re/bigfuzzy_not_utf8.t` because it says it's a binary file.
> 
> Command-line `diff` also identifies this file as a binary file and says it differs from the version of that file in blead.
> 
> ```
> [perl2] 2066 $ diff ../perl/t/re/bigfuzzy_not_utf8.t ./t/re/bigfuzzy_not_utf8.t
> Binary files ../perl/t/re/bigfuzzy_not_utf8.t and ./t/re/bigfuzzy_not_utf8.t differ
> ```
> 
> (In the above, blead is checked out into `../perl` and your branch into `./perl2` and rebased it on blead.)
> 
> But are the differences limited to just your substitution of `Config->import` for `import Config`? When I examine the two files with `vimdiff`, I see that line 37 in blead has been broken into two lines in your branch. According to `wc -l`, the file is 39 lines long in blead but 40 in your branch. Per [superuser](https://superuser.com/questions/125376/how-do-i-compare-binary-files-in-linux#125408), you can also get a diff of the hexdumps of the two versions of the file with something like this:
> 
> ```
> $ diff <(xxd blead.bigfuzzy_not_utf8.t) <(xxd branch.bigfuzzy_not_utf8.t)
> ```
> 
> So before merging we have to get bigfuzzy to contain only your intended changes. Can you investigate?

That file indeed contains binary data passed as a function argument.

This fact unfortunately seem to mess with my IDE and tools.

I excluded the `t/re/bigfuzzy.t` file from my changes as well, and noted that in the commit message.

I'm noting to work on these two excluded files later in another PR, perhaps individually.

Thanks for the review

tonycoz (approved):

rwp0:
I think this is good to go too now.

`Encode` fixed its tests as well in https://github.com/dankogai/p5-encode/pull/176 upstream.

PR #21503, opened 2023-09-21
feature: document feature bundle ':all' and why you don't use it

### LINKED ISSUE #17209: Tweak Travis CI configuration to limit the smoker outside of blead to a single flavor
The idea is to speedup Travis CI smoke run time.

Most of the time when submitting a case, you do not really need to run the test suite on all Perl Flavors.

I'm suggesting to limit it to a single flavor on all PR or commit outside of blead,
and let blead do the complete smoking.

Another thing we should consider is abort the Travis CI build on the first failure.

tonycoz:
Fixes #17209

### DIFF
--- a/lib/feature.pm
+++ b/lib/feature.pm
@@ -4,7 +4,7 @@
 # Any changes made here will be lost!
 
 package feature;
-our $VERSION = '1.84';
+our $VERSION = '1.85';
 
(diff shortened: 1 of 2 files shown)


leonerd (approved):
LGTM

PR #21507, opened 2023-09-21
Fix perl version macros

mauke:


Grinnz:
I don't know how regression tests of XS code are done but that seems like a good idea

mauke:
@Grinnz I put a sanity check inline in `handy.h`. Maybe that should go in XS::APItest?

khwilliamson:
Regressions in modules are handled by ppport.h.  There have been plenty over the years.  So the fix should go in Devel::PPPort.  The code there looks correct.  What is needed is to undefine the incorrect macro, so that the correct one will be substituted.

mauke:
> Regressions in modules are handled by ppport.h. There have been plenty over the years.

I don't know what this means.

> So the fix should go in Devel::PPPort.

Agreed. I was planning to submit an equivalent patch there (when/if this PR is accepted).

> The code there looks correct.

.. huh? If the code in Devel:::PPPort is correct, why would it need fixing?

Besides, looking at https://github.com/Perl/perl5/blob/cd622ffb09d80e3663c61b6a4b620438507526af/dist/Devel-PPPort/parts/inc/version#L151-L154 it seems like it uses the same incorrect logic as `handy.h`.

> What is needed is to undefine the incorrect macro, so that the correct one will be substituted.

No action needed there. Devel::PPPort always redefines `PERL_VERSION_LE` to use its own version, so once that's fixed, it will be preferred automatically.

Leont (approved):

mauke:
Equivalent change to Devel::PPPort: https://github.com/Dual-Life/Devel-PPPort/pull/237

PR #21469, opened 2023-09-12
Enable thread-safe locales on some freebsd versions

khwilliamson:
This had been disabled in the hints file due to bugs, which perl now works around, so reenable.  For example, commit
0f3830f3997cf7ef1531bad26d2e0f13220dd862, "locale.c: Work around some libc POSIX 2008 bugs"

PR #21513, opened 2023-09-23
Test simultaneous threads running different locales

khwilliamson:
This adds an elaborate test to lib/locale_threads.t, designed to stress test perl's (and the platform's) ability to properly handle multiple threads, each running with a different combination of locales.

Locales originally were global to the whole process, but POSIX 2008 and Windows (even earlier) introduced per-thread locales.  This test is skipped on platforms that don't offer either of these.  But otherwise it verifies that they actually work and that perl correctly uses them.

Each thread in the test runs a tight loop of calling locale-sensitive operations, and checking that their result is as expected (which is computed before starting the thread).  This checks that other threads aren't interfering with this one.

Each thread is started with LC_CTYPE in locale A, LC_NUMERIC in locale B, etc for each category on the system for which perl has an operation that uses that category.  Most platforms do not properly handle such a situation, returning mojibake if, for example, LC_CTYPE is in an incompatible locale with LC_TIME.  perl takes extra actions to make this work, and the test verifies that.  (In reading code, it appears that some recent libc's do handle this without a perl workaround being needed, but that is a minority capability, and I have chosen to leave the workaround in on those platforms, due in part to not being able to write a Configure probe that reliably finds more than surface-level failures.)

Every so many iterations, each thread will shift to using a different set of disparate locales, and then check that the new expected values are the ones now returned.  Doing this in the tight loop should surface races that would arise when changing locales, and verify that each thread is independent of the others.

If two locales are almost the same, it could be that a thread getting its expected result was in fact just getting an identical result from another thread.  To minimize this, the test goes to some lengths to try to find locales as different from each other as possible to run simultaneously.

Experience with smoking this led me to choose 30 threads with 100 iterations each, as higher values lead to it taking too long, and running out of memory on some of our smoke farm boxes.  On my Linux box, I've used 500 threads and 1000 iterations, which takes a long time.  On my Windows box, something like 100 threads and 1000 iterations.  And on some other boxes I have access to, something like 50 threads and 1000 iterations.

I started writing this test a couple of years ago, and only now, after a couple hundred locale-related commits to blead, is it ready itself to be put into blead without generating failures.  I also have submitted tickets against some platforms.  Just this week, two such tickets were resolved, https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=269375 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=255646

I did not file tickets on platforms where I thought they would be ignored, so no tickets against Windows nor Darwin, but workarounds remain for them in perl.

### DIFF
--- a/lib/locale_threads.t
+++ b/lib/locale_threads.t
@@ -4,27 +4,442 @@ use warnings;
 # This file tests interactions with locale and threads
 
 BEGIN {
+    $| = 1;
+
     chdir 't' if -d 't';
     require './test.pl';
     set_up_inc('../lib');
-    require './loc_tools.pl';
-    skip_all("No locales") unless locales_enabled();
+
     skip_all_without_config('useithreads');
-    $| = 1;
-    eval { require POSIX; POSIX->import(qw(locale_h  unistd_h)) };
+
+    require './loc_tools.pl';
+
+    eval { require POSIX; POSIX->import(qw(errno_h locale_h unistd_h )) };
     if ($@) {
-	skip_all("could not load the POSIX module"); # running minitest?
+        skip_all("could not load the POSIX module"); # running minitest?
     }
 }
 
+use Time::HiRes qw(time usleep);
+
+use Devel::Peek;
+$Devel::Peek::pv_limit = 0; $Devel::Peek::pv_limit = 0;
+use Data::Dumper;
+$Data::Dumper::Sortkeys=1;
+$Data::Dumper::Useqq = 1;
+$Data::Dumper::Deepcopy = 1;
+
+my $debug = 0;
+
+my %map_category_name_to_number;
+my %map_category_number_to_name;
+my @valid_categories = valid_locale_categories();
+foreach my $category (@valid_categories) {
+    my $cat_num = eval "&POSIX::$category";
+    die "Can't determine ${category}'s number: $@" if $@;
+
+    $map_category_name_to_number{$category} = $cat_num;
+    $map_category_number_to_name{$cat_num} = $category;
+}
+
+my $LC_ALL;
+my $LC_ALL_string;
+if (defined $map_category_name_to_number{LC_ALL}) {
+    $LC_ALL_string = 'LC_ALL';
+    $LC_ALL = $map_category_name_to_number{LC_ALL};
+}
+elsif (defined $map_category_name_to_number{LC_CTYPE}) {
+    $LC_ALL_string = 'LC_CTYPE';
+    $LC_ALL = $map_category_name_to_number{LC_CTYPE};
+}
+else {
+    skip_all("No LC_ALL nor LC_CTYPE");
+}
+
 # reset the locale environment
-local @ENV{'LANG', (grep /^LC_/, keys %ENV)};
+delete local @ENV{'LANGUAGE', 'LANG', keys %map_category_name_to_number};
 
+my @locales = find_locales($LC_ALL);
+skip_all("Couldn't find any locales") if @locales == 0;
+
+plan(2);
+
+my ($utf8_locales_ref, $non_utf8_locales_ref)
+                                    = classify_locales_wrt_utf8ness(\@locales);
+
+my $official_ascii_name = 'ansi_x341968';
+
+my %lang_code_to_script = (     # ISO 639.2, but without the many codes that
+                                # are for latin (but the few western European
+                                # ones that are latin1 are included)
+                            am          => 'amharic',
+                            amh         => 'amharic',
+                            amharic     => 'amharic',
+                            ar          => 'arabic',
+                            be          => 'cyrillic',
+                            bel         => 'cyrillic',
+                            ben         => 'bengali',
+                            bn          => 'bengali',
+                            bg          => 'cyrillic',
+                            bul         => 'cyrillic',
+                            bulgarski   => 'cyrillic',
+                            bulgarian   => 'cyrillic',
+                            c           => $official_ascii_name,
+                            cnr         => 'cyrillic',
+                            de          => 'latin_1',
+                            deu         => 'latin_1',
+                            deutsch     => 'latin_1',
+                            german      => 'latin_1',
+                            div         => 'thaana',
+                            dv          => 'thaana',
+                            dzo         => 'tibetan',
+                            dz          => 'tibetan',
+                            el          => 'greek',
+                            ell         => 'greek',
+                            ellada      => 'greek',
+                            en          => $official_ascii_name,
+                            eng         => $official_ascii_name,
+                            american    => $official_ascii_name,
+                            british     => $official_ascii_name,
+                            es          => 'latin_1',
+                            fa          => 'arabic',
+                            fas         => 'arabic',
+                            flamish     => 'latin_1',
+                            fra         => 'latin_1',
+                            fr          => 'latin_1',
+                            heb         => 'hebrew',
+                            he          => 'hebrew',
+                            hi          => 'hindi',
+                            hin         => 'hindi',
+                            hy          => 'armenian',
+                            hye         => 'armenian',
+                            ita         => 'latin_1',
+                            it          => 'latin_1',
+                            ja          => 'katakana',
+                            jpn         => 'katakana',
+                            nihongo     => 'katakana',
+                            japanese    => 'katakana',
+                            ka          => 'georgian',
+                            kat         => 'georgian',
+                            kaz         => 'cyrillic',
+                            khm         => 'khmer',
+                            kir         => 'cyrillic',
+                            kk          => 'cyrillic',
+                            km          => 'khmer',
+                            ko          => 'hangul',
+                            kor         => 'hangul',
+                            korean      => 'hangul',
+                            ku          => 'arabic',
+                            kur         => 'arabic',
+                            ky          => 'cyrillic',
+                            latin1      => 'latin_1',
+                            lao         => 'lao',
+                            lo          => 'lao',
+                            mk          => 'cyrillic',
+                            mkd         => 'cyrillic',
+                            macedonian  => 'cyrillic',
+                            mn          => 'cyrillic',
+                            mon         => 'cyrillic',
+                            mya         => 'myanmar',
+                            my          => 'myanmar',
+                            ne          => 'devanagari',
+                            nep         => 'devanagari',
+                            nld         => 'latin_1',
+                            nl          => 'latin_1',
+                            nederlands  => 'latin_1',
+                            dutch       => 'latin_1',
+                            por         => 'latin_1',
+                            posix       => $official_ascii_name,
+                            ps          => 'arabic',
+                            pt          => 'latin_1',
+                            pus         => 'arabic',
+                            ru          => 'cyrillic',
+                            russki      => 'cyrillic',
+                            russian     => 'cyrillic',
+                            rus         => 'cyrillic',
+                            sin         => 'sinhala',
+                            si          => 'sinhala',
+                            so          => 'arabic',
+                            som         => 'arabic',
+                            spa         => 'latin_1',
+                            sr          => 'cyrillic',
+                            srp         => 'cyrillic',
+                            tam         => 'tamil',
+                            ta          => 'tamil',
+                            tg          => 'cyrillic',
+                            tgk         => 'cyrillic',
+                            tha         => 'thai',
+                            th          => 'thai',
+                            thai        => 'thai',
+                            ti          => 'ethiopian',
+                            tir         => 'ethiopian',
+                            uk          => 'cyrillic',
+                            ukr         => 'cyrillic',
+                            ur          => 'arabic',
+                            urd         => 'arabic',
+                            zgh         => 'arabic',
+                            zh          => 'chinese',
+                            zho         => 'chinese',
+                        );
+my %codeset_to_script = (
+                            88591  => 'latin_1',
+                            88592  => 'latin_2',
+                            88593  => 'latin_3',
+                            88594  => 'latin_4',
+                            88595  => 'cyrillic',
+                            88596  => 'arabic',
+                            88597  => 'greek',
+                            88598  => 'hebrew',
+                            88599  => 'latin_5',
+                            885910 => 'latin_6',
+                            885911 => 'thai',
+                            885912 => 'devanagari',
+                            885913 => 'latin_7',
+                            885914 => 'latin_8',
+                            885915 => 'latin_9',
+                            885916 => 'latin_10',
+                            cp1251 => 'cyrillic',
+                            cp1255 => 'hebrew',
+                      );
+
+my %script_priorities = (       # In trying to make the results as distinct as
+                                # possible, make the ones closest to Unicode,
+                                # and ASCII lowest priority
+                            $official_ascii_name => 15,
+                            latin_1 => 14,
+                            latin_9 => 13,
+                            latin_2 => 12,
+                            latin_4 => 12,
+                            latin_5 => 12,
+                            latin_6 => 12,
+                            latin_7 => 12,
+                            latin_8 => 12,
+                            latin_10 => 12,
+                            latin   => 11,  # Unknown latin version
+                        );
+
+my %script_instances;   # Keys are scripts, values are how many locales use
+                        # this script.
+
+sub analyze_locale_name($) {
+
+    # Takes the input name of a locale and creates (and returns) a hash
+    # containing information about that locale
+
+    my %ret;
+    my $input_locale_name = shift;
+
+    my $old_locale = setlocale(LC_CTYPE);
+
+    # Often a locale has multiple aliases, and the base one is returned
+    # by setlocale() when called with an alias.  The base is more likely to
+    # meet the XPG standards than the alias.
+    my $new_locale = setlocale(LC_CTYPE, $input_locale_name);
+    if (! $new_locale) {
+        diag "Unexpectedly can't setlocale(LC_CTYPE, $new_locale);"
+           . " \$!=$!, \$^E=$^E";
+        return;
+    }
+
+    $ret{locale_name} = $new_locale;
+
+    # XPG standard for locale names:
+    #   language[_territory[.codeset]][@modifier]
+    # But, there are instances which violate this, where there is a codeset
+    # without a territory, so instead match:
+    #   language[_territory][.codeset][@modifier]
+    $ret{locale_name} =~ / ^
+                                      ( .+? )          # language
+                              (?:  _  ( .+? ) )?       # territory
+                              (?: \.  ( .+? ) )?       # codeset
+                              (?: \@  ( .+  ) )?       # modifier
+                            $
+                         /x;
+
+    $ret{language}  = $1 // "";
+    $ret{territory} = $2 // "";
+    $ret{codeset}   = $3 // "";
+    $ret{modifier}  = $4 // "";
+
+    # Normalize all but 'territory' to lowercase
+    foreach my $key (qw(language codeset modifier)) {
+        $ret{$key} = lc $ret{$key};
+    }
+
+    # Often, the codeset is omitted from the locale name, but it is still
+    # discoverable (via langinfo() ) for the current locale on many platforms.
+    # We already have switched locales
+    use I18N::Langinfo qw(langinfo CODESET);
+    my $langinfo_codeset = lc langinfo(CODESET);
+
+    # Now can switch back to the locale current on entry to this sub
+    if (! setlocale(LC_CTYPE, $old_locale)) {
+        die "Unexpectedly can't restore locale to $old_locale from"
+          . " $new_locale; \$!=$!, \$^E=$^E";
+    }
+
+    # Normalize the codesets
+    foreach my $codeset_ref (\$langinfo_codeset, \$ret{codeset}) {
+        $$codeset_ref =~ s/\W//g;
+        $$codeset_ref =~ s/iso8859/8859/g;
+        $$codeset_ref =~ s/\b65001\b/utf8/;     # Windows synonym
+        $$codeset_ref =~ s/\b646\b/$official_ascii_name/;
+        $$codeset_ref =~ s/\busascii\b/$official_ascii_name/;
+    }
+
+    # The langinfo codeset, if found, is considered more reliable than the one
+    # in the name.  (This is because libc looks into the actual data
+    # definition.)  So use it unconditionally when found.  But note any
+    # discrepancy as an aid for improving this test.
+    if ($langinfo_codeset) {
+        if ($ret{codeset} && $ret{codeset} ne $langinfo_codeset) {
+            diag "In $ret{locale_name}, codeset from langinfo"
+               . " ($langinfo_codeset) doesn't match codeset in"
+               . " locale_name ($ret{codeset})";
+        }
+        $ret{codeset} = $langinfo_codeset;
+    }
+
+    $ret{is_utf8} = 0 + ($ret{codeset} eq 'utf8');
+
+    # If the '@' modifier is a known script, use it as the script.
+    if (    $ret{modifier}
+        and grep { $_ eq $ret{modifier} } values %lang_code_to_script)
+    {
+        $ret{script} = $ret{nominal_script} = $ret{modifier};
+        $ret{modifier} = "";
+    }
+    elsif ($ret{codeset} && ! $ret{is_utf8}) {
+
+        # The codeset determines the script being used, except if we don't
+        # have the codeset, or it is UTF-8 (which covers a multitude of
+        # scripts).
+        #
+        # We have hard-coded the scripts corresponding to a few of these
+        # non-UTF-8 codesets.  See if this is one of them.
+        $ret{script} = $codeset_to_script{$ret{codeset}};
+        if ($ret{script}) {
+
+            # For these, the script is likely a combination of ASCII (from
+            # 0-127), and the script from (128-255).  Reflect that in the name
+            # used (for distinguishing below)
+            $ret{script} .= '_' . $official_ascii_name;
+        }
+        elsif ($ret{codeset} =~ /^koi/) {   # Another common set.
+            $ret{script} = "cyrillic_${official_ascii_name}";
+        }
+        else {  # Here the codeset name is unknown to us.  Just assume it
+                # means a whole new script.  Add the language at the end of
+                # the name to further make it distinct
+            $ret{script} = $ret{codeset};
+            $ret{script} .= "_$ret{language}"
+                                    if $ret{codeset} !~ /$official_ascii_name/;
+        }
+    }
+    else {  # Here, the codeset is unknown or is UTF-8.
+
+        # In these cases look up the script based on the language.  The table
+        # is meant to be pretty complete, but omits the many scripts that are
+        # ASCII or Latin1.  And it omits the fullnames of languages whose
+        # scripts are themselves.  The grep below catches those.  Defaulting
+        # to Latin means that a non-standard language name is considered to be
+        # latin -- maybe not the best outcome but what else is better?
+        $ret{script} = $lang_code_to_script{$ret{language}};
+        if (! $ret{script}) {
+            $ret{script} = (grep { $ret{language} eq $_ }
+                                                    values %lang_code_to_script)
+                            ? $ret{language}
+                            : 'latin';
+        }
+    }
+
+    # If we have @euro, and the script is ASCII or latin or latin1, change it
+    # into latin9, which is closer to what is going on.  latin9 has a few
+    # other differences from latin1, but it's not worth creating a whole new
+    # script type that differs only in the currency symbol.
+    if (  ($ret{modifier} && $ret{modifier} eq 'euro')
+        && $ret{script} =~ / ^ ($official_ascii_name | latin (_1)? ) $ /x)
+    {
+        $ret{script} = 'latin_9';
+    }
+
+    #  Look up the priority of this script.  All the non-listed ones have
+    #  highest (0 or 1) priority.  We arbitrarily make the ones higher
+    #  priority (0) that aren't known to be half-ascii, simply because they
+    #  might be entirely different than most locales.
+    $ret{priority} = $script_priorities{$ret{script}};
+    if (! $ret{priority}) {
+        $ret{priority} = (   $ret{script} ne $official_ascii_name
+                          && $ret{script} =~ $official_ascii_name)
+                         ? 0
+                         : 1;
+    }
+
+    # Script names have been set up so that anything after an underscore is a
+    # modifier of the main script.  We keep a counter of which occurence of
+    # this script this is.  This is used along with the priority to order the
+    # locales so that the characters are as varied as possible.
+    my $script_root = ($ret{script} =~ s/_.*//r) . "_$ret{is_utf8}";
+    $ret{script_instance} = $script_instances{$script_root}++;
+
+    return \%ret;
+}
+
+# Prioritize locales that are most unlike the standard C/Latin1-ish ones.
+# This is to minimize getting passes for tests on a category merely because
+# they share many of the same characteristics as the locale of another
+# category simultaneously in effect.
+sub sort_locales ()
+{
+    my $cmp =  $a->{script_instance} <=> $b->{script_instance};
+    return $cmp if $cmp;
+
+    $cmp =  $a->{priority} <=> $b->{priority};
+    return $cmp if $cmp;
+
+    $cmp =  $a->{script} cmp $b->{script};
+    return $cmp if $cmp;
+
+    $cmp =  $a->{modifier} cmp $b->{modifier};
+    return $cmp if $cmp;
+
+    $cmp =  $a->{codeset} cmp $b->{codeset};
+    return $cmp if $cmp;
+
+    $cmp =  $a->{territory} cmp $b->{territory};
+    return $cmp if $cmp;
+
+    return lc $a cmp lc $b;
+}
+
+# Find out extra info about each locale
+my @cleaned_up_locales;
+for my $locale (@locales) {
+    my $locale_struct = analyze_locale_name($locale);
+
+    next unless $locale_struct;
+
+    my $name = $locale_struct->{locale_name};
+    next if grep { $name eq $_->{locale_name} } @cleaned_up_locales;
+
+    push @cleaned_up_locales, $locale_struct;
+}
+
+@locales = @cleaned_up_locales;
+
+# Without a proper codeset, we can't really know how to test.  This should
+# only happen on platforms that lack the ability to determine the codeset.
+@locales = grep { $_->{codeset} ne "" } @locales;
+
+# Sort into priority order.
+@locales = sort sort_locales @locales;
+
+# First test
 SKIP: { # perl #127708
-    my @locales = grep { $_ !~ / ^ C \b | POSIX /x } find_locales('LC_MESSAGES');
-    skip("No valid locale to test with", 1) unless @locales;
-
-    local $ENV{LC_MESSAGES} = $locales[0];
+    my $locale = $locales[0];
+    skip("No valid locale to test with", 1) if $locale->{codeset} eq
+                                                          $official_ascii_name;
+    local $ENV{LC_MESSAGES} = $locale->{locale_name};
 
     # We're going to try with all possible error numbers on this platform
     my $error_count = keys(%!) + 1;
@@ -33,11 +448,13 @@ SKIP: { # perl #127708
         use threads;
         use strict;
         use warnings;
+        use Time::HiRes qw(usleep);
 
         my \$errnum = 1;
 
         my \@threads = map +threads->create(sub {
-            sleep 0.1;
+            usleep 0.1;
+            'threads'->yield();
 
             for (1..5_000) {
                 \$errnum = (\$errnum + 1) % $error_count;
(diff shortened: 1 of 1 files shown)


lib/locale_threads.t
```diff
@@ -1,30 +1,445 @@
 use strict;
+#23456789112345678921234567893123456789412345678951234567896123456789712345678981
```
jkeenan: What is the purpose of the line above?

jkeenan:
On a threaded build on Linux, I ran `make test_prep && ./perl -Ilib lib/locale_threads.t`.  That reported:
```
1..2
ok 1 - Didn't segfault
ok 2 - Verify there were no failures with simultaneous running threads
```
On a similar threaded build on FreeBSD-13, I did likewise, the run ended with:
```
1..2
ok 1 - Didn't segfault
ok 2 # skip Unsafe locale threads
```
So, the test file PASSed, but only by skipping "unsafe" locale threads.  How do I find out what those are?  What should I do about them?

jkeenan:
In [this failure on MacOS (big sur)](https://pipelinesghubeus22.actions.githubusercontent.com/F8yobFedDMKReLD2lUI49NJpt8TNtEKo3DptPtnjfqw2qokl1r/_apis/pipelines/1/runs/35615/signedlogcontent/43?urlExpires=2023-09-23T20%3A50%3A14.3009393Z&urlSigningMethod=HMACV1&urlSignature=VybDC9Sn57W8wqK7WllO5ZtFDOKm%2BaourflA%2FXgcOyY%3D), we get an enormous amount of failure output -- perhaps too much?  How would you advise using this to debug the failure?

lib/locale_threads.t
```diff
@@ -54,67 +471,1290 @@ SKIP: { # perl #127708
...
+    return $category_number if $category_number != $LC_ALL;
+
+    # If this was LC_ALL, the next one won't be
+    return &get_next_category();
+}
+
 SKIP: {
-    skip("POSIX version doesn't support thread-safe locale operations", 1)
-                                                unless ${^SAFE_LOCALES};
+    skip("Unsafe locale threads", 1) unless ${^SAFE_LOCALES};
```
jkeenan, line 796: Can you add a comment in here as to how ${^SAFE_LOCALES} gets determined ("set") on a given machine?  (I can't figure that out from existing documentation in `perllocale` and `perlvar`.  Thanks.

khwilliamson: perlvar says

> ${^SAFE_LOCALES}
             Reflects if safe locale operations are available to this perl
             (when the value is 1) or not (the value is 0). This variable is
             always 1 if the perl has been compiled without threads. It is
             also 1 if this perl is using thread-safe locale operations. Note
             that an individual thread may choose to use the global locale
             (generally unsafe) by calling "switch_to_global_locale" in
             perlapi. This variable currently is still set to 1 in such
             threads.
  
 This makes sense to me but patches or suggestions are welcome.  If using locales is safe to do on this platform without fear of another thread coming along and screwing this thread up, then the value is 1.  You shouldn't be using locales on platforms where the value is 0, as, you can get segfaults, etc.

lib/locale_threads.t
khwilliamson: It is a relic that I should have removed, and will do so before any push

lib/locale_threads.t
khwilliamson: github is not allowing me to reply to some of the separate comments individually, so here is a conglomeration.

I don't know how to shorten the failure output.  What I did was to write the got vs expected out and then used a diff tool that has per-character granularity.  This particular test is failing in the output of strxfrm() which is not designed to be human-readable.  So, looking at the diff showed me a few slight differences, but in internal strxfrm format, so unclear as to the significance.  But what it does show is either the perl code is not doing the right thing on Darwin, or the Darwin libc has a race.  I think its the latter, as the code runs fine on OpenBSD.  And this  test file usually passes on Darwin, with the occasional flap.  I've never seen strxfrm() fail on it.  I have looked at the libc code in Darwin before, and it is based on early *BSD code, that hasn't been synced with newer BSD releases; those releases have fixed bugs in this area, and continue to do so.  What this means is that Darwin cannot be considered to have thread-safe locales.  I'll turn off their use in another commit and repush this PR

khwilliamson: FreeBSD 13.1 does not have safe threaded locales, as  the failure in #21469 indicates.  Hence the test is skipped.

jkeenan: Would it then be fair and accurate to say, *"Whether or not a particular machine has safe threaded locales is a characteristic of the machine over which the person compiling, building or using Perl has no control whatsoever?"*

Of course, that begs the question, *"How does Perl discover or calculate what the value of ${^SAFE_LOCALES} is for a particular machine?"*

khwilliamson: Not quite.  You have no control over whether a particular machine is capable of having safe threaded locales, except indirectly by filing bug reports when the machine wrongly claims such capability, and waiting if and until it gets fixed.

But you can Configure perl to not use the safe interface even if it is available.  I see no good reason to do so unless you discover that the safe interface is buggier than the unsafe one.

Perl knows that ${^SAFE_LOCALES} is true of non-threaded perls, as there is no other thread than your own.  Otherwise, Configure probes if the POSIX safe interface is available, and assumes it works as advertised.  But experience has shown  that this is a lie in many cases, so hints files are used to override this for particular versions.  Perl knows when Microsoft introduced safe threaded locales, and uses preprocessor conditionals to act on that.  Similarly, Perl knows that MingW needs to have been compiled with a modern libc dll, and checks for that

dur-randir:
Fails on MacOS Monterey with 
```
1..2
ok 1 - Didn't segfault
sh: line 1: 76162 Abort trap: 6           /Users/dur-randir/Projects/perl-git/perl "-I../lib" "-I." "tmp_DHQC_A" 2>&1
not ok 2 - Verify there were no failures with simultaneous running threads
# Failed test 2 - Verify there were no failures with simultaneous running threads at lib/locale_threads.t line 1444
#      got "Assertion failed: (strNE(retval, \"\")), function S_querylocale_2008_i, file locale.c, line 2193."
# expected "1"
#  diff at 0
#    after ""
#     have "Assertion failed: (strNE(retval, \"\")), f"
#     want "1"
# PROG:
#
#         BEGIN { $| = 1; }
#         use threads;
#         use strict;
#         use warnings;
#         use POSIX qw(locale_h);
#         use utf8;
#         use Time::HiRes qw(time usleep);
#         $|=1;
#
#         use Data::Dumper;
#         $Data::Dumper::Sortkeys=1;
#         $Data::Dumper::Useqq = 1;
#         $Data::Dumper::Deepcopy = 1;
#
#         # Get the tests stored for us by the setup process
#         use Storable;
#         my $all_tests_ref = retrieve('tmp_DHQC_B');
#         if (! defined $all_tests_ref) {
#             die 'Could not restore the built-up data structure';
#         }
#
#         my %corrects;
#
#         sub output_test_failure_prefix {
#             my ($iteration, $category_name, $test) = @_;
#             my $tid = threads->tid();
#             print STDERR "\nthread ", $tid,
#                          " failed in iteration $iteration",
#                          " for locale $test->{locale_name}",
#                          " codeset='$test->{codeset}'",
#                          " $category_name",
#                          "\nop='$test->{op}'",
#                          "\nafter getting ",
#                          ($corrects{$category_name}
#                                     {$test->{locale_name}}
#                                     {all} // 0),
#                          ' previous correct results for this category and',
#                          ' locale,
# including ',
#                          ($corrects{$category_name}
#                                     {$test->{locale_name}}
#                                     {$tid} // 0),
#                          ' in this thread
# ';
#         }
#
#         sub output_test_result($$$) {
#             my ($type, $result, $utf8_matches) = @_;
#
#             no locale;
#
#             print STDERR "$type";
#
#             my $copy = $result;
#             if (! $utf8_matches) {
#                 if (utf8::is_utf8($copy)) {
#                     print STDERR " (result already was in UTF-8)";
#                 }
#                 else {
#                     utf8::upgrade($copy);
#                     print STDERR " (result wasn't in UTF-8; converted for",
#                                  " easier comparison)";
#                 }
#             }
#             print STDERR ":\n";
#
#             use Devel::Peek;
#             Dump $copy;
#         }
#
#         sub iterate {       # Run some chunk of iterations of the tests
#             my ($tid,                  # Which thread
#                 $initial_iteration,    # The number of the first iteration
#                 $count,                # How many
#                 $tests_ref)            # The tests
#              = @_;
#
#             my $iteration = $initial_iteration;
#             $count += $initial_iteration;
#
#             # Repeatedly ...
#             while ($iteration < $count) {
#                 my $errors = 0;
#
#                 use locale;
#
#                 # ... execute the tests
#                 foreach my $test ($tests_ref->@*) {
#
#                     # We know what we are expecting
#                     my $expected = $test->{expected};
#
#                     my $category_name = $test->{category_name};
#
#                     # And do the test.
#                     my $got = eval $test->{op};
#
#                     if (! defined $got) {
#                         output_test_failure_prefix($iteration,
#                                                    $category_name,
#                                                    $test);
#                         output_test_result('expected', $expected,
#                                            1 # utf8ness matches, since only one
#                                           );
#                         $errors++;
#                         next;
#                     }
#
#                     my $utf8ness_matches = (   utf8::is_utf8($got)
#                                              == utf8::is_utf8($expected));
#
#                     my $matched = ($got eq $expected);
#                     if ($matched) {
#                         if ($utf8ness_matches) {
#                             no warnings 'uninitialized';
#                             $corrects{$category_name}
#                                       {$test->{locale_name}}
#                                       {all}++;
#                             $corrects{$category_name}
#                                       {$test->{locale_name}}
#                                       {$tid}++;
#                             next;   # Complete success!
#                         }
#                     }
#
#                     $errors++;
#                     output_test_failure_prefix($iteration,
#                                                $category_name,
#                                                $test);
#
#                     if ($matched) {
#                         print STDERR "Only difference is UTF8ness",
#                                      " of results\n";
#                     }
#                     output_test_result('expected', $expected,
#                                        $utf8ness_matches);
#                     output_test_result('got', $got,
#                                        $utf8ness_matches);
#
#                 } # Loop to do the remaining tests for this iteration
#
#                 return 0 if $errors;
#
#                 $iteration++;
#
#                 # A way to set a gdb break point pp_study
#                 #study if $iteration % 10 == 0;
#
#                 threads->yield();
#             }
#
#             return 1;
#         } # End of iterate() definition
#
#         my $map_category_number_to_name = {
#                                  "0" => "LC_ALL",
#                                  "1" => "LC_COLLATE",
#                                  "2" => "LC_CTYPE",
#                                  "3" => "LC_MONETARY",
#                                  "4" => "LC_NUMERIC",
#                                  "5" => "LC_TIME",
#                                  "6" => "LC_MESSAGES"
#                                };
# ;
#
#         sub setlocales {
#             # Set each category to the appropriate locale for this test group
#             my ($categories, $locales) = @_;
#             for (my $i = 0; $i < $categories->@*; $i++) {
#                 if (! setlocale($categories->[$i], $locales->[$i])) {
#                     my $category_name =
#                             $map_category_number_to_name->{$categories->[$i]};
#                     print STDERR "\nthread ", threads->tid(),
#                         " setlocale($category_name ($categories->[$i]),",
#                         " $locales->[$i]) failed
# ";
#                     return 0;
#                 }
#             }
#
#             return 1;
#         }
#
#         my $startup_insurance = 1;
#         my $future = $startup_insurance
#                     + 15 * 0.18;
#         my $starting_time = time() + $future;
#
#         sub wait_until_time {
#
#             # Sleep until the time when all the threads are due to wake up, so
#             # they run as simultaneously as we can make it.
#             my $sleep_time = ($starting_time - time());
#             #printf STDERR "thread %d started, sleeping %g sec\n",
#             #              threads->tid, $sleep_time;
#             if ($sleep_time < 0 && 1) {
#                 # What the start time should have been
#                 my $a_better_future = $future - $sleep_time;
#
#                 my $better_per_thread =
#                     ($a_better_future - $startup_insurance) / 15;
#                 printf STDERR '\$per_thread_startup would need to be "%g'
#                             . ' for thread %d to have started
# in sync with'
#                             . ' the other threads
# ',
#                               $better_per_thread, threads->tid;
#                 die 'Thread started too late';
#             }
#             else {
#                 usleep($sleep_time * 1_000_000) if $sleep_time > 0;
#             }
#         }
#
#         # Create all the subthreads: 1..n
#         my @threads = map +threads->create(sub {
#             $SIG{'KILL'} = sub { threads->exit(); };
#
#             my $thread = shift;
#
#             # Start out with the set of tests whose number is the same as the
#             # thread number
#             my $which_test_group = $thread;
#
#             wait_until_time();
#
#             # Loop through all the iterations for this thread
#             my $this_iteration_start = 1;
#             do {
#                 # Set up each category with its locale;
#                 my $this_ref = $all_tests_ref->[$which_test_group];
#                 return 0 unless setlocales($this_ref->{categories},
#                                            $this_ref->{locales});
#                 # Then run one batch of iterations
#                 my $result = iterate($thread,
#                                       $this_iteration_start,
#                                       20,
#                                       $this_ref->{tests});
#                 return 0 if $result == 0;  # Quit if failed
#
#                 # Next iteration will shift to use a different set of locales
#                 # for each category
#                 $which_test_group++;
#                 $which_test_group = 0 if $which_test_group >= 15;
#                 $this_iteration_start += 20;
#             } while ($this_iteration_start <= 100);
#
#             return 1;   # Success
#
#         }, $_), (1..15 - 1);    # For each non-0 thread
#
#         # Here is thread 0.  We do a smaller chunk of iterations in it; then
#         # join whatever threads have finished so far, then do another chunk.
#         # This tests for bugs that arise as a result of joining.
#
#         my %thread0_corrects = ();
#         my $this_iteration_start = 1;
#         my $result = 1;    # So far, everything is ok
#         my $which_test_group = -1;  # Start with 0th test group
#
#         wait_until_time();
#         alarm(600);    # Guard against hangs
#
#         do {
#             $which_test_group++;
#
#             # Next time, we'll use the next test group
#             $which_test_group = 0 if $which_test_group >= 15;
#
#             my $this_ref = $all_tests_ref->[$which_test_group];
#
#             # set the locales for this test group.  Do this even if we are
#             # going to bail, so that it will be set correctly for the final
#             # batch after the loop.
#             $result &= setlocales($this_ref->{categories},
#                                    $this_ref->{locales});
#
#             if (0 > 1) {
#                 my @joinable = threads->list(threads::joinable);
#                 if (@joinable) {
#                     print STDERR 'In thread 0, before iteration ',
#                                     $this_iteration_start,
#                                     ' these threads are done: ',
#                                     join(', ', map { $_->tid() } @joinable),
#                                     '
# ';
#                 }
#             }
#
#             # Join anything already finished.
#             for my $thread (threads->list(threads::joinable)) {
#                 my $thread_result = $thread->join;
#                 if (0 > 1) {
#                     print STDERR 'In thread 0, before iteration ',
#                                  $this_iteration_start,
#                                  ' joining thread ', $thread->tid(),
#                                  '; result=', ((defined $thread_result)
#                                                ? $thread_result
#                                                : 'undef'),
#                                   '
# ';
#                 }
#
#                 # If the thread failed badly, stop testing anything else.
#                 if (! defined $thread_result) {
#                     $_->kill('KILL')->detach() for threads->list();
#                     print 0;
#                     exit;
#                 }
#
#                 # Update the status
#                 $result &= $thread_result;
#             }
#
#             # Do a chunk of iterations on this thread 0.
#             $result &= iterate(0,
#                                 $this_iteration_start,
#                                 20,
#                                 $this_ref->{tests},
#                                 \%thread0_corrects);
#             $this_iteration_start += 20;
#
#           # And repeat as long as there are other tests
#         } while (threads->list(threads::all));
#
#         print $result
# STATUS: 34304
```
built with
```
./perl -Ilib -V|grep config_args
    config_args='-de -Dusedevel -DDEBUGGING -Dusethreads'
```

khwilliamson:
@dur-randir I believe it should pass now (because it should be being skipped through a new commit)

lib/locale_threads.t
khwilliamson: I used this test to investigate failures, deciding that either it was perl's fault and fixed that, or that it was the platform's fault.  Usually in the latter case, I or others built a C reproducer program not involving perl.  So the answer to how the answer was discovered is that it it was mostly a manual process resulting in bug fixes to perl, or to hints files changes

dur-randir:
I'm on commit ae541d257e, which seems to be the last one, and the test isn't skipped.

khwilliamson:
[55c4444](https://github.com/Perl/perl5/pull/21513/commits/55c4444a9a6143bdfbf099fd39aa487557119e20) is supposed to skip it.  I don't understand why it isn't.  Perhaps you could eyeball it, and it might be obvious.  Otherwise attach a perl -V output, please.

lib/locale_threads.t
```diff
@@ -4,27 +4,441 @@ use warnings;
...
+                            latin_7 => 12,
+                            latin_8 => 12,
+                            latin_10 => 12,
+                            latin   => 11,  # Unknown latin version
+                        );
+
+my %script_instances;   # Keys are scripts, values are how many locales use
+                        # this script.
+
+sub analyze_locale_name($) {
```
tonycoz, line 224: While not an issue for this commit, I wonder how much of this could go in a module.

lib/locale_threads.t
```diff
@@ -54,67 +470,1290 @@ SKIP: { # perl #127708
...
+
+            # Use a variable to avoid runtime bugs being hidden by constant
+            # folding
+            add_trials('LC_NUMERIC', 'my $in = 4.2; sprintf("%g", $in)');
+            next;
+        }
+
+        if ($category eq 'LC_TIME') {
+            add_trials('LC_TIME', "POSIX::strftime($strftime_args)");
+            add_trials('LC_TIME',
```
tonycoz: This might be more readable as a heredoc.

lib/locale_threads.t
```diff
@@ -54,67 +470,1290 @@ SKIP: { # perl #127708
...
+    # start a bunch of threads.  Each sets the locale of each category being
+    # tested to the value determined in the code above.  Then each sleeps
+    # to a common start time, at which point they awaken and iterate their
+    # respective loops.  Each iteration runs a set of tests and checks that
+    # the results are as expected.  This should catch any instances of other
+    # threads interfering.  The code above has set up things so that each
+    # thread has as disparate results from the other threads as possible, so
+    # to more likely catch any bleed-through.
     fresh_perl_is("
+        BEGIN { \$| = 1; }
```
tonycoz: Rather than a huge string literal here, I think it would be better to build this up in pieces, with heredocs for the literal parts.

Right now it's mostly literal text with many backslashes, making it hard to pick out the small number of replacements (like `$iterations_per_test_group`, `$category_number_to_name`).

It would significantly reduce the number of `\` in the code.

lib/locale_threads.t
```diff
@@ -54,67 +470,1290 @@ SKIP: { # perl #127708
     pass("Didn't segfault");
 }
 
+# Second test setup
+my %locale_name_to_object;
+for (my $i = 0; $i < @locales; $i++) {
```
tonycoz: Is there a reason this is a C style loop instead of:
```
for my $i (0 .. $#locales) { # or @locales-1
```
There's other cases further down.

dur-randir:
@khwilliamson I've re-run configure and it's now skipped. Looks like some artifact of a partial rebuild.

jkeenan:
@khwilliamson, @tonycoz has posed several questions about this p.r.  Can you comment?

lib/locale_threads.t
khwilliamson: I don't understand how you mean this to be

lib/locale_threads.t
khwilliamson: I don't know of other uses for this.

lib/locale_threads.t
khwilliamson: Done

lib/locale_threads.t
khwilliamson: Just habit from mostly programming in C these days.  Changed.

lib/locale_threads.t
mauke: Presumably something like:
```perl
add_trials('LC_TIME', <<'END_OF_CODE');
use I18N::Langinfo qw(langinfo
    ABDAY_1 ABDAY_2 ABDAY_3 ABDAY_4 ABDAY_5 ABDAY_6 ABDAY_7
    ABMON_1 ABMON_2 ABMON_3 ABMON_4 ABMON_5 ABMON_6
    ABMON_7 ABMON_8 ABMON_9 ABMON_10 ABMON_11 ABMON_12
    DAY_1 DAY_2 DAY_3 DAY_4 DAY_5 DAY_6 DAY_7
    MON_1 MON_2 MON_3 MON_4 MON_5 MON_6
    MON_7 MON_8 MON_9 MON_10 MON_11 MON_12
    D_FMT D_T_FMT T_FMT);
no warnings "uninitialized";
join "|",
     map { langinfo($_) }
           ABDAY_1,ABDAY_2,ABDAY_3,ABDAY_4,ABDAY_5,
           ABDAY_6,ABDAY_7,
           ABMON_1,ABMON_2,ABMON_3,ABMON_4,ABMON_5,
           ABMON_6, ABMON_7,ABMON_8,ABMON_9,ABMON_10,
           ABMON_11,ABMON_12,
           DAY_1,DAY_2,DAY_3,DAY_4,DAY_5,DAY_6,DAY_7,
           MON_1,MON_2,MON_3,MON_4,MON_5,MON_6, MON_7,
           MON_8,MON_9,MON_10,MON_11,MON_12,
           D_FMT,D_T_FMT,T_FMT;
END_OF_CODE
```

khwilliamson: thanks; changed to @mauke suggestion

lib/locale_threads.t
```diff
@@ -54,67 +471,1305 @@ SKIP: { # perl #127708
...
+
+    # Then just the errnos.
+    my @msg_catalog = sort { $a <=> $b } keys %msg_catalog;
+
+    # Remove the excess ones.
+    splice @msg_catalog, $max_message_catalog_entries
+                                          if $max_message_catalog_entries >= 0;
+    my $msg_catalog = join ',', @msg_catalog;
+
+    eval "POSIX::localeconv()->{currency_symbol}";
```
tonycoz: Why is this a string eval?  Wouldn't:
```
eval  { POSIX::localeconv()->{currency_symbol}; };
```
be equivalent?

lib/locale_threads.t
```diff
@@ -54,67 +471,1305 @@ SKIP: { # perl #127708
...
+        my @cat_tests = $combinations_ref->@*;
+
+        # This sorts the test cases so that the ones with the least overlap
+        # with other cases are first.
+        sub sort_test_order {
+            my $a_tests_count = scalar $a->{in_common_locale_counts}->@*;
+            my $b_tests_count = scalar $b->{in_common_locale_counts}->@*;
+            my $tests_count = ($a_tests_count <= $b_tests_count)
+                              ? $a_tests_count
+                              : $b_tests_count;
```
tonycoz: You define min() above, this could use `min($a_tests_count, $b_tests_count)`

lib/locale_threads.t
khwilliamson: Yes, thanks; changed

lib/locale_threads.t
khwilliamson: Thanks, done

tonycoz (approved):

sisyphus:
For me on Windows 11, running current blead (at commit 5314ff9f) built with a gcc-13.2.0 UCRT compiler, the second test of lib/locale_threads.t takes just on 3 minutes to (successfully) terminate.
Is that reasonable and expected ?

The threading model of this gcc-13.2.0 compiler is MCF threads, and <code>perl -V</code> reports:
```
.\perl -I.\lib -V
Summary of my perl5 (revision 5 version 39 subversion 5) configuration:
  Derived from:
  Platform:
    osname=MSWin32
    osvers=10.0.22621.2428
    archname=MSWin32-x64-multi-thread-quadmath
    uname=''
    config_args='undef'
    hint=recommended
    useposix=true
    d_sigaction=undef
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=undef
    uselongdouble=undef
    usemymalloc=n
  Compiler:
    cc='gcc'
    ccflags =' -DWIN32 -DWIN64 -fdiagnostics-color=never -DPERL_TEXTMODE_SCRIPTS -DMULTIPLICITY -DPERL_IMPLICIT_SYS -DUSE_PERLIO -D__USE_MINGW_ANSI_STDIO -fwrapv -fno-strict-aliasing -mms-bitfields'
    optimize='-Os'
    cppflags='-DWIN32'
    ccversion=''
    gccversion='13.2.0'
    gccosandvers=''
    intsize=4
    longsize=4
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long long'
    ivsize=8
    nvtype='__float128'
    nvsize=16
    Off_t='long long'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='g++'
    ldflags ='-s -L"D:\perl-5.39.41-1320-ucrt-mcf\lib\MSWin32-x64-multi-thread-quadmath\CORE" -L"C:\winlibs-ucrt-1320\mingw64\lib" -L"C:\winlibs-ucrt-1320\mingw64\x86_64-w64-mingw32\lib" -L"C:\winlibs-ucrt-1320\mingw64\lib\gcc\x86_64-w64-mingw32\13.2.0"'
    libpth=C:\winlibs-ucrt-1320\mingw64\lib C:\winlibs-ucrt-1320\mingw64\x86_64-w64-mingw32\lib C:\winlibs-ucrt-1320\mingw64\lib\gcc\x86_64-w64-mingw32\13.2.0 D:\_64\msys_1320\1.0\local\lib
    libs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladvapi32 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lversion -lodbc32 -lodbccp32 -lcomctl32 -lquadmath
    perllibs= -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladvapi32 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lversion -lodbc32 -lodbccp32 -lcomctl32 -lquadmath
    libc=
    so=dll
    useshrplib=true
    libperl=libperl539.a
    gnulibc_version=''
  Dynamic Linking:
    dlsrc=dl_win32.xs
    dlext=dll
    d_dlsymun=undef
    ccdlflags=' '
    cccdlflags=' '
    lddlflags='-shared -s -L"D:\perl-5.39.41-1320-ucrt-mcf\lib\MSWin32-x64-multi-thread-quadmath\CORE" -L"C:\winlibs-ucrt-1320\mingw64\lib" -L"C:\winlibs-ucrt-1320\mingw64\x86_64-w64-mingw32\lib" -L"C:\winlibs-ucrt-1320\mingw64\lib\gcc\x86_64-w64-mingw32\13.2.0"'


  Compile-time options:
    HAS_LONG_DOUBLE
    HAS_STRTOLD
    HAS_TIMES
    HAVE_INTERP_INTERN
    MULTIPLICITY
    PERLIO_LAYERS
    PERL_COPY_ON_WRITE
    PERL_DONT_CREATE_GVSV
    PERL_HASH_FUNC_SIPHASH13
    PERL_HASH_USE_SBOX32
    PERL_IMPLICIT_SYS
    PERL_MALLOC_WRAP
    PERL_OP_PARENT
    PERL_PRESERVE_IVUV
    PERL_USE_SAFE_PUTENV
    USE_64_BIT_INT
    USE_ITHREADS
    USE_LARGE_FILES
    USE_LOCALE
    USE_LOCALE_COLLATE
    USE_LOCALE_CTYPE
    USE_LOCALE_NUMERIC
    USE_LOCALE_TIME
    USE_PERLIO
    USE_PERL_ATOF
    USE_QUADMATH
    USE_THREAD_SAFE_LOCALE
  Locally applied patches:
    uncommitted-changes
  Built under MSWin32
  Compiled at Nov  2 2023 16:21:23
  @INC:
    .\lib
    D:/comp-1320-ucrt/perl-5.39.41/lib
```
Cheers,
Rob

khwilliamson:
Yes it takes a long time.  And I wish I could add even more simultaneous threads to it.  Probably we should lower the number; perhaps depending on platform.  You could experiment by changing `$thread_count` and/or the iteration count.  For now, I think we should keep the number where it is, to better shake out races, but we may need to speed it up for 5.40.  Please open a ticket to that effect

tonycoz:
The new test is failing on threaded builds on dragonflybsd and openbsd:

https://perl5.test-smoke.org/report/5042914
https://perl5.test-smoke.org/report/5042917
https://www.nntp.perl.org/group/perl.daily-build.reports/2023/11/msg299114.html (no smokedb report for this one)

jkeenan:
> The new test is failing on threaded builds on dragonflybsd and openbsd:
> 
> https://perl5.test-smoke.org/report/5042914 https://perl5.test-smoke.org/report/5042917 https://www.nntp.perl.org/group/perl.daily-build.reports/2023/11/msg299114.html (no smokedb report for this one)

List of smoke-test failures as of 2023-11-11:
https://perl5.test-smoke.org/submatrix?test=../lib/locale_threads.t&pversion=5.39.5

Moreover, the output from a test failure is overwhelming (to me, at least).  From a build today on OpenBSD-6.9, clang-10 as the compiler.
```
$ sh ./Configure -des -Dusedevel -Duseithreads -Accflags="-DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE -fno-delete-null-pointer-checks"
...
$ cd t; ./perl harness -v ../lib/locale_threads.t; cd -

ok 1 # skip No valid locale to test with
not ok 2 - Verify there were no failures with simultaneous running threads
# Failed test 2 - Verify there were no failures with simultaneous running threads at ../lib/locale_threads.t line 1768
#      got "\nthread 0 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $string = chr 0x100; utf8::encode($string); no warnings \"uninitialized\"; POSIX::mblen(undef); POSIX::mblen($string)\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PV(0x2b4c2e0b2a0) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b53891bc30 \"2\"\\0\n  CUR = 1\n  LEN = 16\ngot:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 1\n  PV = 0x2b53891bc30 \"1\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 0 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 1\n  PV = 0x2b53891bc30 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b53891bc30 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 10 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PV(0x2b4fe40ea30) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b563a784b0 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b5863fc338) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b563a784b0 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 10 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PVIV(0x2b5863fc338) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a784b0 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b5863fc338) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a784b0 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 10 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 19 previous correct results for this category and locale,\nincluding 19 in this thread\nexpected:\nSV = PVIV(0x2b5863fc338) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a784b0 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b563a570d0) at 0x2b552fed2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b563a784b0 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 0 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b53891bc30 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b53891bc30 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 12 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PV(0x2b564fb6a30) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b4f29110f0 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b564fcd338) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b4f29110f0 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 12 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PVIV(0x2b564fcd338) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4f29110f0 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b564fcd338) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4f29110f0 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 12 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 19 previous correct results for this category and locale,\nincluding 19 in this thread\nexpected:\nSV = PVIV(0x2b564fcd338) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4f29110f0 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b4c2df10d0) at 0x2b4e68422b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b4f29110f0 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 0 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 18 previous correct results for this category and locale,\nincluding 18 in this thread\nexpected:\nSV = PVIV(0x2b5154b3320) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b53891bc30 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b4ee2083a0) at 0x2b4ee1f0238\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b53891bc30 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 8 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $string = chr 0x100; utf8::encode($string); no warnings \"uninitialized\"; POSIX::mblen(undef); POSIX::mblen($string)\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PV(0x2b532b80ab0) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b515a07370 \"2\"\\0\n  CUR = 1\n  LEN = 16\ngot:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 1\n  PV = 0x2b515a07370 \"1\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 8 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 1\n  PV = 0x2b515a07370 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b515a07370 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 8 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b515a07370 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b515a07370 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 8 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 18 previous correct results for this category and locale,\nincluding 18 in this thread\nexpected:\nSV = PVIV(0x2b4cc82d320) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b515a07370 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b5159e80d0) at 0x2b5565702b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b515a07370 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 4 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 2 previous correct results for this category and locale,\nincluding 2 in this thread\nexpected:\nSV = PV(0x2b4b8cb9b10) at 0x2b5713982b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b4f2b012d0 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b588f9a338) at 0x2b5713982b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 0\n  PV = 0x2b4f2b012d0 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 7 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PV(0x2b4efde3a30) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b4b8cca6b0 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b4efdd3338) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b4b8cca6b0 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 7 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PVIV(0x2b4efdd3338) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4b8cca6b0 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b4efdd3338) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4b8cca6b0 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 7 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 19 previous correct results for this category and locale,\nincluding 19 in this thread\nexpected:\nSV = PVIV(0x2b4efdd3338) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b4b8cca6b0 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b4c2e1f0d0) at 0x2b52a24c2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b4b8cca6b0 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 9 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PV(0x2b583547a30) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b499ffd1d0 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b583560338) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b499ffd1d0 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 9 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 1 previous correct results for this category and locale,\nincluding 1 in this thread\nexpected:\nSV = PVIV(0x2b583560338) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b499ffd1d0 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b583560338) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b499ffd1d0 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 9 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 19 previous correct results for this category and locale,\nincluding 19 in this thread\nexpected:\nSV = PVIV(0x2b583560338) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b499ffd1d0 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b49a00c0d0) at 0x2b56bce32b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b499ffd1d0 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n\nthread 14 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $string = chr 0x100; utf8::encode($string); no warnings \"uninitialized\"; POSIX::mblen(undef); POSIX::mblen($string)\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PV(0x2b4f291aab0) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  PV = 0x2b563a78950 \"2\"\\0\n  CUR = 1\n  LEN = 16\ngot:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 1\n  PV = 0x2b563a78950 \"1\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 14 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; my $str = \"\\x{100}\"; utf8::encode($str); no warnings \"uninitialized\"; POSIX::mbtowc(undef, undef); POSIX::mbtowc($value, $str); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 1\n  PV = 0x2b563a78950 \"256\"\\0\n  CUR = 3\n  LEN = 16\ngot:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (IOK,pIOK,pPOK)\n  IV = 196\n  PV = 0x2b563a78950 \"196\"\\0\n  CUR = 3\n  LEN = 16\n\nthread 14 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'my $value; no warnings \"uninitialized\"; POSIX::wctomb(undef, undef); POSIX::wctomb($value, 0xFF); $value;\'\nafter getting 0 previous correct results for this category and locale,\nincluding 0 in this thread\nexpected:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a78950 \"\\xC3\\xBF\"\\0\n  CUR = 2\n  LEN = 16\ngot:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a78950 \"\\xFF\"\\0\n  CUR = 1\n  LEN = 16\n\nthread 14 failed in iteration 1 for locale C.UTF-8 codeset=\'utf8\' LC_CTYPE\nop=\'use I18N::Langinfo qw(langinfo CODESET); no warnings \"uninitialized\"; langinfo(CODESET);\'\nafter getting 18 previous correct results for this category and locale,\nincluding 18 in this thread\nexpected:\nSV = PVIV(0x2b4f28e6320) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  PV = 0x2b563a78950 \"UTF-8\"\\0\n  CUR = 5\n  LEN = 16\ngot:\nSV = PVMG(0x2b563a820d0) at 0x2b550ffb2b0\n  REFCNT = 1\n  FLAGS = (POK,pPOK)\n  IV = 196\n  NV = 0\n  PV = 0x2b563a78950 \"US-ASCII\"\\0\n  CUR = 8\n  LEN = 16\n0"
# expected "1"
#  diff at 0
#    after ""
#     have "\nthread 0 failed in iteration 1 for loca"
#     want "1"
# PROG: 
# 
#     BEGIN { $| = 1; }
#     my $debug = 0;
#     my $thread_count = 15;
#     my $iterations_per_test_group = 20;
#     my $iterations = 100;
#     my $die_on_negative_sleep = 1;
#     my $per_thread_startup = 0.18;
#     my $all_tests_file = tmp_VSJ_B;
#     my $alarm_clock = 600;
#     use threads;
#     use strict;
#     use warnings;
#     use POSIX qw(locale_h);
#     use utf8;
#     use Time::HiRes qw(time usleep);
#     $|=1;
# 
#     use Data::Dumper;
#     $Data::Dumper::Sortkeys=1;
#     $Data::Dumper::Useqq = 1;
#     $Data::Dumper::Deepcopy = 1;
# 
#     # Get the tests stored for us by the setup process
#     use Storable;
#     my $all_tests_ref = retrieve($all_tests_file);
#     if (! defined $all_tests_ref) {
#         die "Could not restore the built-up data structure";
#     }
# 
#     my %corrects;
# 
#     sub output_test_failure_prefix {
#         my ($iteration, $category_name, $test) = @_;
#         my $tid = threads->tid();
#         print STDERR "\nthread ", $tid,
#                      " failed in iteration $iteration",
#                      " for locale $test->{locale_name}",
#                      " codeset='$test->{codeset}'",
#                      " $category_name",
#                      "\nop='$test->{op}'",
#                      "\nafter getting ", ($corrects{$category_name}
#                                           {$test->{locale_name}}
#                                           {all} // 0),
#                      " previous correct results for this category and",
#                      " locale,\nincluding ", ($corrects{$category_name}
#                                               {$test->{locale_name}}
#                                               {$tid} // 0),
#                      " in this thread\n";
#     }
# 
#     sub output_test_result($$$) {
#         my ($type, $result, $utf8_matches) = @_;
# 
#         no locale;
# 
#         print STDERR "$type";
# 
#         my $copy = $result;
#         if (! $utf8_matches) {
#             if (utf8::is_utf8($copy)) {
#                 print STDERR " (result already was in UTF-8)";
#             }
#             else {
#                 utf8::upgrade($copy);
#                 print STDERR " (result wasn't in UTF-8; converted for easier",
#                              " comparison)";
#             }
#         }
#         print STDERR ":\n";
# 
#         use Devel::Peek;
#         Dump $copy;
#     }
# 
#     sub iterate {       # Run some chunk of iterations of the tests
#         my ($tid,                  # Which thread
#             $initial_iteration,    # The number of the first iteration
#             $count,                # How many
#             $tests_ref)            # The tests
#             = @_;
# 
#         my $iteration = $initial_iteration;
#         $count += $initial_iteration;
# 
#         # Repeatedly ...
#         while ($iteration < $count) {
#             my $errors = 0;
# 
#             use locale;
# 
#             # ... execute the tests
#             foreach my $test ($tests_ref->@*) {
# 
#                 # We know what we are expecting
#                 my $expected = $test->{expected};
# 
#                 my $category_name = $test->{category_name};
# 
#                 # And do the test.
#                 my $got = eval $test->{op};
# 
#                 if (! defined $got) {
#                     output_test_failure_prefix($iteration,
#                                                $category_name,
#                                                $test);
#                     output_test_result("expected", $expected,
#                                         1 # utf8ness matches, since only one
#                                       );
#                     $errors++;
#                     next;
#                 }
# 
#                 my $utf8ness_matches = (   utf8::is_utf8($got)
#                                         == utf8::is_utf8($expected));
# 
#                 my $matched = ($got eq $expected);
#                 if ($matched) {
#                     if ($utf8ness_matches) {
#                         no warnings 'uninitialized';
#                         $corrects{$category_name}{$test->{locale_name}}{all}++;
#                         $corrects{$category_name}{$test->{locale_name}}{$tid}++;
#                         next;   # Complete success!
#                     }
#                 }
# 
#                 $errors++;
#                 output_test_failure_prefix($iteration, $category_name, $test);
# 
#                 if ($matched) {
#                     print STDERR "Only difference is UTF8ness of results\n";
#                 }
#                 output_test_result("expected", $expected, $utf8ness_matches);
#                 output_test_result("got", $got, $utf8ness_matches);
# 
#             } # Loop to do the remaining tests for this iteration
# 
#             return 0 if $errors;
# 
#             $iteration++;
# 
#             # A way to set a gdb break point pp_study
#             #study if $iteration % 10 == 0;
# 
#             threads->yield();
#         }
# 
#         return 1;
#     } # End of iterate() definition
# 
# my $map_category_number_to_name = {
#                                  "0" => "LC_ALL",
#                                  "2" => "LC_CTYPE",
#                                  "3" => "LC_MONETARY",
#                                  "5" => "LC_TIME",
#                                  "6" => "LC_MESSAGES"
#                                };
# 
#     sub setlocales {
#         # Set each category to the appropriate locale for this test group
#         my ($categories, $locales) = @_;
#         for my $i (0 .. $categories->@* - 1) {
#             if (! setlocale($categories->[$i], $locales->[$i])) {
#                 my $category_name =
#                             $map_category_number_to_name->{$categories->[$i]};
#                 print STDERR "\nthread ", threads->tid(),
#                              " setlocale($category_name ($categories->[$i]),",
#                              " $locales->[$i]) failed\n";
#                 return 0;
#             }
#         }
# 
#         return 1;
#     }
# 
#     my $startup_insurance = 1;
#     my $future = $startup_insurance + $thread_count * $per_thread_startup;
#     my $starting_time = time() + $future;
# 
#     sub wait_until_time {
# 
#         # Sleep until the time when all the threads are due to wake up, so
#         # they run as simultaneously as we can make it.
#         my $sleep_time = ($starting_time - time());
#         #printf STDERR "thread %d started, sleeping %g sec\n",
#         #              threads->tid, $sleep_time;
#         if ($sleep_time < 0 && $die_on_negative_sleep) {
#             # What the start time should have been
#             my $a_better_future = $future - $sleep_time;
# 
#             my $better_per_thread =
#                         ($a_better_future - $startup_insurance) / $thread_count;
#             printf STDERR "$per_thread_startup would need to be %g",
#                           " for thread %d to have started\nin sync with",
#                           " the other threads\n",
#                           $better_per_thread, threads->tid;
#             die "Thread started too late";
#         }
#         else {
#             usleep($sleep_time * 1_000_000) if $sleep_time > 0;
#         }
#     }
# 
#     # Create all the subthreads: 1..n
#     my @threads = map +threads->create(sub {
#         $SIG{'KILL'} = sub { threads->exit(); };
# 
#         my $thread = shift;
# 
#         # Start out with the set of tests whose number is the same as the
#         # thread number
#         my $which_test_group = $thread;
# 
#         wait_until_time();
# 
#         # Loop through all the iterations for this thread
#         my $this_iteration_start = 1;
#         do {
#              # Set up each category with its locale;
#             my $this_ref = $all_tests_ref->[$which_test_group];
#             return 0 unless setlocales($this_ref->{categories},
#                                         $this_ref->{locales});
#             # Then run one batch of iterations
#             my $result = iterate($thread,
#                                  $this_iteration_start,
#                                  $iterations_per_test_group,
#                                  $this_ref->{tests});
#             return 0 if $result == 0;   # Quit if failed
# 
#             # Next iteration will shift to use a different set of locales for
#             # each category
#             $which_test_group++;
#             $which_test_group = 0 if $which_test_group >= $thread_count;
#             $this_iteration_start += $iterations_per_test_group;
#         } while ($this_iteration_start <= $iterations);
# 
#         return 1;   # Success
# 
#     }, $_), (1..$thread_count - 1);     # For each non-0 thread
# 
#     # Here is thread 0.  We do a smaller chunk of iterations in it; then
#     # join whatever threads have finished so far, then do another chunk.
#     # This tests for bugs that arise as a result of joining.
# 
#     my %thread0_corrects = ();
#     my $this_iteration_start = 1;
#     my $result = 1;    # So far, everything is ok
#     my $which_test_group = -1;  # Start with 0th test group
# 
#     wait_until_time();
#     alarm($alarm_clock);    # Guard against hangs
# 
#     do {
#         $which_test_group++;
#         # Next time, we'll use the next test group
#         $which_test_group = 0 if $which_test_group >= $thread_count;
# 
#         my $this_ref = $all_tests_ref->[$which_test_group];
# 
#         # set the locales for this test group.  Do this even if we are
#         # going to bail, so that it will be set correctly for the final
#         # batch after the loop.
#         $result &= setlocales($this_ref->{categories}, $this_ref->{locales});
# 
#         if ($debug > 1) {
#             my @joinable = threads->list(threads::joinable);
#             if (@joinable) {
#                 print STDERR "In thread 0, before iteration ",
#                              $this_iteration_start,
#                              " these threads are done: ",
#                              join (", ", map { $_->tid() } @joinable),
#                              "\n";
#             }
#         }
# 
#         # Join anything already finished.
#         for my $thread (threads->list(threads::joinable)) {
#             my $thread_result = $thread->join;
#             if ($debug > 1) {
#                 print STDERR "In thread 0, before iteration ",
#                              $this_iteration_start,
#                              " joining thread ", $thread->tid(),
#                              "; result=", ((defined $thread_result)
#                                            ? $thread_result
#                                            : "undef"),
#                              "\n";
#             }
# 
#             # If the thread failed badly, stop testing anything else.
#             if (! defined $thread_result) {
#                 $_->kill('KILL')->detach() for threads->list();
#                 print 0;
#                 exit;
#             }
# 
#             # Update the status
#             $result &= $thread_result;
#         }
# 
#         # Do a chunk of iterations on this thread 0.
#         $result &= iterate(0,
#                            $this_iteration_start,
#                            $iterations_per_test_group,
#                            $this_ref->{tests},
#                            \%thread0_corrects);
#         $this_iteration_start += $iterations_per_test_group;
# 
#         # And repeat as long as there are other tests
#     } while (threads->list(threads::all));
# 
#     print $result;
# STATUS: 0
Failed 1/2 subtests 
	(less 1 skipped subtest: 0 okay)

Test Summary Report
-------------------
../lib/locale_threads.t (Wstat: 0 Tests: 2 Failed: 1)
  Failed test:  2
Files=1, Tests=2, 15 wallclock secs ( 0.02 usr  0.01 sys + 15.48 cusr  3.97 csys = 19.48 CPU)
Result: FAIL
```
**I believe this commit should be reverted until we get a better grasp on the problem.**

khwilliamson:
@tonycoz and I are looking at the issue.  If these two platforms are indeed a problem, I'd rather skip the test on them rather than reverting the test, so as to gain experience with the other platforms that aren't showing problems as of now.

jkeenan:
> @tonycoz and I are looking at the issue. If these two platforms are indeed a problem, I'd rather skip the test on them rather than reverting the test, so as to gain experience with the other platforms that aren't showing problems as of now.

https://github.com/Perl/perl5/pull/21633

PR #21516, opened 2023-09-25
builtin: avoid the looseness of sscanf() parsing

tonycoz:


leonerd (approved):
Code/test changes LGTM

The `g++` smoker looks unhappy but at first glance it really doesn't look like it's anything caused by this change. Might just be a suprious failure in Test2 itself. Perhaps try rerunning it?

leonerd:
`g++` happy now on a rerun

tonycoz:
> Might just be a suprious failure in Test2 itself. Perhaps try rerunning it?

Unfortunately there's been a few of those, I've had them on some pushes, and some smokers are getting them intermittently.

PR #21519, opened 2023-09-26
say that `ord` returns code points

dakkar:
all the other uses of "numeric value" are in the `atoi` sense, where "numeric value" of '1' is 1, but `ord('1')` is 49; "code point" is more correct/precise, and already used in many places

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -5196,9 +5196,9 @@ X<ord> X<encoding>
 
 =item ord
 
-=for Pod::Functions find a character's numeric representation
+=for Pod::Functions find a character's code point
 
-Returns the numeric value of the first character of EXPR.
+Returns the code point of the first character of EXPR.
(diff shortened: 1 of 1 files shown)


Leont (approved):

Grinnz:
Perhaps "numeric code point" would be even clearer?

khwilliamson:
Either way is fine with me.  @dakkar what do you thinkZ?

dakkar:
all other places just say "code point" https://github.com/search?q=repo%3APerl%2Fperl5+%22code+point%22+path%3A.pod&type=code
with the exception of the documentation for `\N{}` which says "named code point"
so I feel that "numeric" is not helping, here

khwilliamson:
@dakkar  Can you squash the second commit into the first one?  Then I will merge this into blead

dakkar:
done!

PR #21520, opened 2023-09-26
`File-Find`: Minor tests/docs improvements

### LINKED ISSUE #21577: `File-Find`: Duplicated `use strict` line
Unnecessary duplication of using the pragma.

https://github.com/Perl/perl5/blob/4dc42bd5cef7cf68d9c7161d91338ff4a4173bc6/ext/File-Find/lib/File/Find.pm#L13

---

https://github.com/Perl/perl5/blob/4dc42bd5cef7cf68d9c7161d91338ff4a4173bc6/ext/File-Find/lib/File/Find.pm#L3-L13

rwp0:
There are two tests with indirect syntax within the File-Find core distribution,
which are not using it for the purpose of testing indirect syntax.

This commit aims to fix those since that syntax is deprecated (or obsolescent).

Also bumps the version to 1.44

Adjust per @ilmari's suggestions

Remove extra `use strict;` in `Find.pm`

---

Fixes https://github.com/Perl/perl5/issues/21577

### DIFF
--- a/ext/File-Find/lib/File/Find.pm
+++ b/ext/File-Find/lib/File/Find.pm
@@ -1,21 +1,21 @@
 package File::Find;
+
 use 5.006;
 use strict;
 use warnings;
 use warnings::register;
-our $VERSION = '1.43';
+
 use Exporter 'import';
 require Cwd;
+require File::Basename;
+require File::Spec;
 
+our $VERSION = '1.44';
 our @EXPORT = qw(find finddepth);
 
-
-use strict;
 my $Is_VMS = $^O eq 'VMS';
 my $Is_Win32 = $^O eq 'MSWin32';
 
-require File::Basename;
-require File::Spec;
 
 # Should ideally be my() not our() but local() currently
 # refuses to operate on lexicals
@@ -812,7 +812,7 @@ File::Find - Traverse a directory tree.
 =head1 DESCRIPTION
 
 These are functions for searching through directory trees doing work
-on each file found similar to the Unix I<find> command.  File::Find
+on each file found similar to the Unix L<find(1)> command.  C<File::Find>
 exports two functions, C<find> and C<finddepth>.  They work similarly
 but have subtle differences.
 
@@ -846,14 +846,14 @@ where C<find()> works from the top of the tree down.
 Despite the name of the C<finddepth()> function, both C<find()> and
 C<finddepth()> perform a depth-first search of the directory hierarchy.
 
-=head2 %options
+=head2 C<%options>
 
 The first argument to C<find()> is either a code reference to your
 C<&wanted> function, or a hash reference describing the operations
 to be performed for each file.  The
 code reference is described in L</The wanted function> below.
 
-Here are the possible keys for the hash:
+Here are the possible B<keys> for the hash:
 
 =over 4
 
@@ -879,7 +879,7 @@ function. It is called with a list of strings (actually file/directory
 names) and is expected to return a list of strings. The code can be
 used to sort the file/directory names alphabetically, numerically,
 or to filter out directory entries based on their name alone. When
-I<follow> or I<follow_fast> are in effect, C<preprocess> is a no-op.
+C<follow> or C<follow_fast> are in effect, C<preprocess> is a no-op.
 
 =item C<postprocess>
 
@@ -887,7 +887,7 @@ The value should be a code reference. It is invoked just before leaving
 the currently processed directory. It is called in void context with no
 arguments. The name of the current directory is in C<$File::Find::dir>. This
 hook is handy for summarizing a directory, such as calculating its disk
-usage. When I<follow> or I<follow_fast> are in effect, C<postprocess> is a
+usage. When C<follow> or C<follow_fast> are in effect, C<postprocess> is a
 no-op.
 
 =item C<follow>
@@ -897,15 +897,15 @@ links (followed) may contain files more than once and may even have
 cycles, a hash has to be built up with an entry for each file.
 This might be expensive both in space and time for a large
 directory tree. See L</follow_fast> and L</follow_skip> below.
-If either I<follow> or I<follow_fast> is in effect:
+If either C<follow> or C<follow_fast> is in effect:
 
 =over 4
 
 =item *
 
-It is guaranteed that an I<lstat> has been called before the user's
+It is guaranteed that an C<lstat> has been called before the user's
 C<wanted()> function is called. This enables fast file checks involving C<_>.
-Note that this guarantee no longer holds if I<follow> or I<follow_fast>
+Note that this guarantee no longer holds if C<follow> or C<follow_fast>
 are not set.
 
 =item *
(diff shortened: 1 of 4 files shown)


ext/File-Find/t/find.t
```diff
@@ -20,7 +20,7 @@ BEGIN {
         @File::Spec::ISA = 'File::Spec::Unix';
     }
     require File::Find;
-    import File::Find;
+    File::Find->import;
```
ilmari: Could this not just be `use File::Find;` after the `BEGIN` block?

ext/File-Find/t/taint.t
```diff
@@ -16,7 +16,7 @@ BEGIN {
         @File::Spec::ISA = 'File::Spec::Unix';
     }
     require File::Find;
-    import File::Find;
+    File::Find->import;
```
ilmari: This test already does `use File::Find;` below, so this can just be removed.

ilmari (changes requested):

ext/File-Find/t/find.t
rwp0: Yes, I also thought about that but didn't "dare" to change 😄

ext/File-Find/lib/File/Find.pm
```diff
@@ -1,21 +1,21 @@
...
 use Exporter 'import';
 require Cwd;
+require File::Basename;
+require File::Spec;
 
+our $VERSION = '1.44';
 our @EXPORT = qw(find finddepth);
 
-
-use strict;
```
rwp0, line 13: Wow, I didn't expect to discover a **repeated** `use strict;` in the module.

ext/File-Find/lib/File/Find.pm
```diff
@@ -1,21 +1,21 @@
...
+
 use 5.006;
 use strict;
 use warnings;
 use warnings::register;
-our $VERSION = '1.43';
+
 use Exporter 'import';
 require Cwd;
+require File::Basename;
```
rwp0, line 10: Not sure why it's using `require` instead of `use`, just to avoid imports (which could be achieved by passing empty list to `use` as well)?

ext/File-Find/t/find.t
```diff
@@ -1,6 +1,4 @@
 #!./perl
-use strict;
```
ilmari, line 2: This should still be here, so it applies to the code in  the `BEGIN` block below.

ilmari (changes requested):

hvds:
I think it's a fine idea to remove or reduce usage of indirect notation in code. I'm less convinced that doing so in tests is a good idea, at least as a blanket effort. At the least, I recommend prefacing such pull requests with an assertion that these are incidental uses of indirect notation - i.e. that the test being changed isn't _intentionally_ testing indirect notation.

rwp0:
> I think it's a fine idea to remove or reduce usage of indirect notation in code. I'm less convinced that doing so in tests is a good idea, at least as a blanket effort. At the least, I recommend prefacing such pull requests with an assertion that these are incidental uses of indirect notation - i.e. that the test being changed isn't _intentionally_ testing indirect notation.

What preface (or **prefix** for PR/commit titles) do you suggest? 😊

This actually does more than just that and is meant to be a distribution-wise change (hoping to receive such a label by the maintainers 🏷️).

I think having so in test files at least gives bad examples to users reading distribution tests, and also being a bad reference for test writers ✍️.

ext/File-Find/t/find.t
rwp0: Noted with thanks

rwp0: Added above `BEGIN` with comment

hvds:
> > I think it's a fine idea to remove or reduce usage of indirect notation in code. I'm less convinced that doing so in tests is a good idea, at least as a blanket effort. At the least, I recommend prefacing such pull requests with an assertion that these are incidental uses of indirect notation - i.e. that the test being changed isn't _intentionally_ testing indirect notation.
> 
> What prefix do you suggest? 😊

I may be misunderstanding what you are asking here, but I suggest rather than starting the pull request with text like:

_There are two tests with indirect syntax within the File-Find core distribution_

instead start it with text like:

_There are two tests with indirect syntax within the File-Find core distribution, which are not using it for the purpose of testing indirect syntax._

> I think having so in test files at least gives bad examples to users reading distribution tests, and also being a bad reference for test writers ✍️.

Having good examples is great. But we need to leave the door open for tests to test the things they are trying to test, which may include testing that indirect syntax works for their particular context.

If at some point we actually talk about _removing_ support for indirect syntax, then part of the process for that would be removing anything in the tests that relies on it. But as long as support is not yet removed we need to have tests that it is actually working as intended. The further you reduce accidental testing by removing current uses, the more important it is that intentional tests are maintained or enhanced.

nicomen:
I think the clean up work of removing undiscouraged use of indirect syntax is great! I hope also that you gradually feel confident enough to simplify the code where possible (like require+import => use and similar). These will be harder to spot after fixing indirect calls only, but even without those I think this is great. Thanks.

I wonder if perlcritic would've picked up some of the modernation needs for this entire module.

rwp0:
> > > I think it's a fine idea to remove or reduce usage of indirect notation in code. I'm less convinced that doing so in tests is a good idea, at least as a blanket effort. At the least, I recommend prefacing such pull requests with an assertion that these are incidental uses of indirect notation - i.e. that the test being changed isn't _intentionally_ testing indirect notation.
> > 
> > 
> > What prefix do you suggest? 😊
> 
> I may be misunderstanding what you are asking here, but I suggest rather than starting the pull request with text like:
> 
> _There are two tests with indirect syntax within the File-Find core distribution_
> 
> instead start it with text like:
> 
> _There are two tests with indirect syntax within the File-Find core distribution, which are not using it for the purpose of testing indirect syntax._
> 
> > I think having so in test files at least gives bad examples to users reading distribution tests, and also being a bad reference for test writers ✍️.
> 
> Having good examples is great. But we need to leave the door open for tests to test the things they are trying to test, which may include testing that indirect syntax works for their particular context.
> 
> If at some point we actually talk about _removing_ support for indirect syntax, then part of the process for that would be removing anything in the tests that relies on it. But as long as support is not yet removed we need to have tests that it is actually working as intended. The further you reduce accidental testing by removing current uses, the more important it is that intentional tests are maintained or enhanced.

I'm not planning to remove intentional tests, and I think that would be a wrong thing to do.

I changed the commit/pull request description, hope that helps.

hvds:
> I'm not planning to remove intentional tests, and I think that would be a wrong thing to do.

Ah good, that hadn't been clear to me up to now.

> I changed the commit/pull request description, hope that helps.

That's super, thank you.

ext/File-Find/lib/File/Find.pm
jkeenan: Let's use `git blame` to investigate.  For line 9 above, that points to this commit:
```
commit 6280b799bc1eabee6c39553ae157386070a217cd
Author:     unknown <perl5-porters.nicoh.com>
AuthorDate: Wed Jan 10 01:46:24 1996 +0000
Commit:     Andy Dougherty <doughera@lafayette.edu>
CommitDate: Wed Jan 10 01:46:24 1996 +0000

    Fixed exporting of symbols to work.

diff --git a/lib/File/Find.pm b/lib/File/Find.pm
index c151bcc891..780dfea469 100644
--- a/lib/File/Find.pm
+++ b/lib/File/Find.pm
@@ -1,9 +1,10 @@
 package File::Find;
 require 5.000;
 require Exporter;
-use Config;
-use Cwd;
-use File::Basename;
+require Config;
+require Cwd;
+require File::Basename;
...
```
Now, the commit message doesn't fully describe the context that led to this commit, but, IMO, if it was good for Andy Dougherty (@doughera88) in 1996, it's still good for me today!

ext/File-Find/lib/File/Find.pm
```diff
@@ -954,12 +954,13 @@ C<$_> will be the same as C<$File::Find::name>.
...
-a regular expression I<untaint_pattern>.  Note that all names passed to the
-user's C<wanted()> function are still tainted. If this option is used while not
-in taint-mode, C<untaint> is a no-op.
+If find is used in L<taint-mode|perlsec/Taint mode> (C<-T> command line
+switch or C<if EUID != UID> or C<if EGID != GID>), then internally
+directory names have to be untainted before they can be C<chdir>'d to.
+Therefore they are checked against a regular expression C<untaint_pattern>.
+Note that all names passed to the user's C<wanted()> function
+are still tainted.
+If this option is used while not in taint-mode, C<untaint> is a no-op.
```
jkeenan: You might as well run line 963 into line 962, assuming they're both in the same paragraph.

jkeenan (changes requested):
Please look at line 963 in Find.pm as noted above; otherwise good to merge.  Thanks.

PR #21521, opened 2023-09-26
`Sys-Hostname`: Few minor improvements


rwp0:
- Correct indirect syntax in test
- Correct indentation on `BEGIN`
- Enrich documentation
- Spelling and grammar
- Capitalize comment sentences

Bump version

### DIFF
--- a/ext/Sys-Hostname/Hostname.pm
+++ b/ext/Sys-Hostname/Hostname.pm
@@ -1,49 +1,45 @@
 package Sys::Hostname;
 
 use strict;
+use warnings ();
 
 use Carp;
-
 use Exporter 'import';
 
 our @EXPORT  = qw/ hostname /;
-
 our $VERSION;
-
-use warnings ();
-
 our $host;
 
 BEGIN {
-    $VERSION = '1.25';
-    {
-	local $SIG{__DIE__};
-	eval {
+  $VERSION = '1.26';
+  {
+	  local $SIG{__DIE__};
+	  eval {
 	    require XSLoader;
 	    XSLoader::load();
-	};
-	warn $@ if $@;
-    }
+	  };
+	  warn $@ if $@;
+  }
 }
 
 
 sub hostname {
-  @_ and croak("hostname() does not accepts arguments (it used to silently discard any provided)");
+  @_ and croak("hostname() does not accept arguments (it used to silently discard any provided)");
 
-  # method 1 - we already know it
+  # Method 1 - We already know it
   return $host if defined $host;
 
-  # method 1' - try to ask the system
+  # Method 1' - Try to ask the system
   $host = ghname() if defined &ghname;
   return $host if defined $host;
 
   if ($^O eq 'VMS') {
 
-    # method 2 - no sockets ==> return DECnet node name
+    # Method 2 - No sockets ==> return DECnet node name
     eval { local $SIG{__DIE__}; $host = (gethostbyname('me'))[0] };
     if ($@) { return $host = $ENV{'SYS$NODE'}; }
 
-    # method 3 - has someone else done the job already?  It's common for the
+    # Method 3 - Has someone else done the job already?  It's common for the
     #    TCP/IP stack to advertise the hostname via a logical name.  (Are
     #    there any other logicals which TCP/IP stacks use for the host name?)
     $host = $ENV{'ARPANET_HOST_NAME'}  || $ENV{'INTERNET_HOST_NAME'} ||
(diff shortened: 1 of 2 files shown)


cpan/libnet/t/hostname.t
```diff
@@ -36,7 +36,7 @@ else {
  print "not ok 1\n";
 }
 
-# This checks thats hostanme does not overwrite $_
+# This checks that hostname does not overwrite $_
```
Grinnz: This change is in an unrelated distribution which is maintained on CPAN and should not be included.

ext/Sys-Hostname/Hostname.pm
```diff
@@ -1,64 +1,60 @@
...
+    # Method 4 - Does hostname happen to work?
     my($rslt) = `hostname`;
     if ($rslt !~ /IVVERB/) { ($host) = $rslt =~ /^(\S+)/; }
     return $host if $host;
 
-    # rats!
+    # Rats!
     $host = '';
-    croak "Cannot get host name of local machine";  
+    croak "Cannot get the hostname of the local machine";
```
Grinnz, line 57: Checked CPAN and the only instances which depend on the current wording of this error message are distributions which have copied Sys::Hostname and its tests themselves. So I am wary of darkpan code depending on it, but otherwise seems fine to change.

ext/Sys-Hostname/Hostname.pm
```diff
@@ -67,54 +63,54 @@ sub hostname {
...
+    # Method 5 - SysV uname command (may truncate)
     || eval {
 	local $SIG{__DIE__};
 	$host = `uname -n 2>/dev/null`; ## sysVish
     }
 
-    # bummer
-    || croak "Cannot get host name of local machine";  
+    # Bummer
+    || croak "Cannot get the hostname of the local machine";
```
Grinnz, line 111: same error message change

jkeenan:
While there are some things I am okay with in this pull request, it is trying to do too many things at once -- most of which don't need to be done at all.

There is no real benefit to changing the casing within comments, *e.g.,*
```
# method 1 - we already know it
```
to
```
# Method 1 - We already know it```
```
There are approximately 15 such changes in this patch.  Not only are these changes useless, they also distract one's attention from the one truly substantive change in the patch, namely, the grammatical correction in the error message (from `does not accepts` to `does not accept`).  I am in favor of correcting mistakes in our error messages, even at the cost of incurring some downstream breakage.

As @Grinnz has noted, we will also face some downstream breakage in changing `Cannot get host name of local machine` to `Cannot get the hostname of the local machine`.  The only reason I'm not completely opposed to this change is that `hostname` is well known as a standard Unix command.

Until this p.r. is trimmed down to what we can agree is substantive, it should not be accepted.

rwp0:
> While there are some things I am okay with in this pull request, it is trying to do too many things at once -- most of which don't need to be done at all.
> 
> There is no real benefit to changing the casing within comments, *e.g.,*
> ```
> # method 1 - we already know it
> ```
> to
> ```
> # Method 1 - We already know it```
> ```
> 

The intended benefit is that when one reads the source code, they can sectionalize the "methods" easier.

rwp0:
> While there are some things I am okay with in this pull request, it is trying to do too many things at once -- most of which don't need to be done at all.
> 
> There is no real benefit to changing the casing within comments, *e.g.,*
> ```
> # method 1 - we already know it
> ```
> to
> ```
> # Method 1 - We already know it```
> ```
> There are approximately 15 such changes in this patch.  Not only are these changes useless, **they also distract one's attention from the one truly substantive change** in the patch, namely, the grammatical correction in the error message (from `does not accepts` to `does not accept`).  I am in favor of correcting mistakes in our error messages, even at the cost of incurring some downstream breakage.
> 
> As @Grinnz has noted, we will also face some downstream breakage in changing `Cannot get host name of local machine` to `Cannot get the hostname of the local machine`.  The only reason I'm not completely opposed to this change is that `hostname` is well known as a standard Unix command.
> 
> Until this p.r. is trimmed down to what we can agree is substantive, it should not be accepted.

So basically you're saying that the PR can be accepted if I revert the comment case changes?

---

I wonder why PR's should include only substantive changes, and why are unsubstantive changes distracting reviewers (might just ignore harmless "trivial cosmetics")?

Super minimal changes in individual PR's will result in multiple version bumps for the distribution, and probably multiple CPAN syncing.

If it's neutral to you, neither good nor bad with no benefit or harm, let it be, no?

And I already mentioned above the reasoning behind those _unsubstantive_ comment changes.


---

I also think this PR should have a `Sys-Hostname` GitHub label on it since it deals with one Perl distribution only 🏷️. I searched and it doesn't seem to exist, not sure why.

Grinnz:
No opinion on the substantiveness of the changes, but regardless multiple version bumping is not necessary, since version bumps are only required when there has been a change from an existing release (in this case, since it is not dual life, a release of Perl).

PR #21528, opened 2023-09-26
util.c: Better memory panic messages

khwilliamson:
Before this commit, the readily-available details of the problem weren't included in the panic messages.

I was struggling to find a bug in this area, and spent a bunch of time, until I hit on adding this detail to the message; the answer became quickly apparent.


richardleach:
LGTM

tonycoz (approved):

PR #21529, opened 2023-09-26
 perlclib: Add extensive advice for interfacing with libc

khwilliamson:
We should decide where this all should go if not perlxs.

### DIFF
--- a/dist/ExtUtils-ParseXS/lib/perlxs.pod
+++ b/dist/ExtUtils-ParseXS/lib/perlxs.pod
@@ -14,7 +14,13 @@ or statically linked into perl.  The XS interface description is
 written in the XS language and is the core component of the Perl
 extension interface.
 
-Before writing XS, read the L</CAVEATS> section below.
+This documents the XS language, but it's important to first note that XS
+code has full access to system calls including C library functions.  It
+thus has the capability of interfering with things that the Perl core or
+other modules have set up, such as signal handlers or file handles.  It
+could mess with the memory, or any number of harmful things.  Don't.
+Further detail is in L<perlclib>, which you should read before actually
+writing any production XS.
 
 An B<XSUB> forms the basic unit of the XS interface.  After compilation
 by the B<xsubpp> compiler, each XSUB amounts to a C function definition
@@ -2110,30 +2116,6 @@ Note that these macros will only work together within the I<same> source
 file; that is, a dMY_CTX in one source file will access a different structure
 than a dMY_CTX in another source file.
 
-=head2 Thread-aware system interfaces
-
-Starting from Perl 5.8, in C/C++ level Perl knows how to wrap
-system/library interfaces that have thread-aware versions
-(e.g. getpwent_r()) into frontend macros (e.g. getpwent()) that
-correctly handle the multithreaded interaction with the Perl
-interpreter.  This will happen transparently, the only thing
-you need to do is to instantiate a Perl interpreter.
-
-This wrapping happens always when compiling Perl core source
-(PERL_CORE is defined) or the Perl core extensions (PERL_EXT is
-defined).  When compiling XS code outside of the Perl core, the wrapping
-does not take place before Perl 5.28.  Starting in that release you can
-
- #define PERL_REENTRANT
-
-in your code to enable the wrapping.  It is advisable to do so if you
-are using such functions, as intermixing the C<_r>-forms (as Perl compiled
-for multithreaded operation will do) and the C<_r>-less forms is neither
-well-defined (inconsistent results, data corruption, or even crashes
-become more likely), nor is it very portable.  Unfortunately, not all
-systems have all the C<_r> forms, but using this C<#define> gives you
-whatever protection that Perl is aware is available on each system.
-
 =head1 EXAMPLES
 
 File C<RPC.xs>: Interface to some ONC+ RPC bind library functions.
@@ -2208,10 +2190,11 @@ In Makefile.PL add -ltirpc and -I/usr/include/tirpc.
 
 =head1 CAVEATS
 
-XS code has full access to system calls including C library functions.
-It thus has the capability of interfering with things that the Perl core
-or other modules have set up, such as signal handlers or file handles.
-It could mess with the memory, or any number of harmful things.  Don't.
+=head2 Use of standard C library functions
+
+See L<perlclib>.
+
+=head2 Event loops and control flow
 
 Some modules have an event loop, waiting for user-input.  It is highly
 unlikely that two such modules would work adequately together in a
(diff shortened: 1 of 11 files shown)


tonycoz:
None of this is "XS language".

I think it belongs in perlguts or perlhacktips, or possibly a new pod.

The existing text under CAVEATS should probably move to the same place.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+thread may overwrite those variables before you get a chance to use
+them.  To overcome this, use an exclusive lock, unlocking only after you
+copy to safety the contents of all the variables you need.
+
+Ideally, you would create a mutex for these, but you may be able to use
+the C<ENV_LOCK> for this purpose, provided that using it doesn't unduly
+slow down actual environment accesses.
+
+=item C<Ꞧ>
+
```
tonycoz: This is horribly mangled by the github pod viewer, and similarly in the table.

https://github.com/khwilliamson/perl5/blob/perlxs/dist/ExtUtils-ParseXS/lib/perlxs.pod

I don't know if adding an `=encoding` directive would help.  (perlpod says "...if your encoding isn't US-ASCII, then put a C<=encoding I<encodingname>> command very early in the document"

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+another thread is calling C<tzset>.  Therefore wrap them using
+C<ENVr_LOCALEr_LOCK>.
+
+=item C<T>
+
+These time-related functions may do writes; perhaps through an indirect
+call to C<tzset>, which writes to time-related global variables.
+Therefore they require an exclusive lock.  See C</L</M>>.
+
+=item C<×>
```
tonycoz: The `×` shows here fine in the github viewer, but is mangled in the table.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+related one.
+
+Some platforms may allocate the buffer per-thread, so it is thread-safe,
+but you have to use the value (or copy it) before the next call.
+
+There may be a re-entrant function available in which the caller
+furnishes the buffer, and which is thread-safe.  Other flags on the
+entry should indicate this.
+
+=item E
```
mauke: Missing `C< >`.

dist/ExtUtils-ParseXS/lib/perlxs.pod
tonycoz: Also, I expect it won't display properly by default in a terminal on Windows.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
-become more likely), nor is it very portable.  Unfortunately, not all
-systems have all the C<_r> forms, but using this C<#define> gives you
-whatever protection that Perl is aware is available on each system.
+This must follow your
+
+ #include <perl.h>
+
+line.  If you don't include F<perl.h>, you must instead place this line
+
+ #include <regen/reentr.pl>
```
mauke: Shouldn't this be `<reentr.h>`?

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+ ENV_LOCK
+ ...
+ ENV_UNLOCK
+
+then the program is safe with regards to these.  If this is done, then
+read-only accesses can be protected by instead using
+
+ ENV_READ_LOCK
+ ...
+ ENV_READ_UNLOCK
```
tonycoz: It might be worth mentioning here that the return value of getenv() won't be valid after the `ENV_READ_UNLOCK;`

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+
+=item E
+
+This function accesses the program's environment.
+
+Writing to the environment can create a race with other threads either
+reading or writing the environment at the same time.  If all writers (as
+the core does) protect their accesses by surrounding them with
+
+ ENV_LOCK
```
tonycoz: All the `*_LOCK`/`*_UNLOCK` lines should have a terminating `;` since that's how they will appear in code.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+
+These variables must not be accessed and functions not called when
+another thread is calling C<tzset>.  Therefore wrap them using
+C<ENVr_LOCALEr_LOCK>.
+
+=item C<T>
+
+These time-related functions may do writes; perhaps through an indirect
+call to C<tzset>, which writes to time-related global variables.
+Therefore they require an exclusive lock.  See C</L</M>>.
```
tonycoz: Is the first `/` in `C</L</M>>` intended here?  It displays like:
image

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2112,27 +2114,410 @@ than a dMY_CTX in another source file.
...
+ dbm_delete()
+ dbm_error()
+ dbm_fetch()
+ dbm_firstkey()
+ dbm_nextkey()
+ dbm_open()
+ dbm_store()
+ dirname()          B
+ dlerror()          B
+ drand48()          Ꞧ
```
tonycoz: Perl provides Perl_drand48() (not just on POSIX-ish)

Unfortunately it provides none of the other *rand48() functions.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2427,3 +2814,5 @@ details so they are only useful to the maintainer of the XS code itself.
 Originally written by Dean Roehrich <F<roehrich@cray.com>>.
 
 Maintained since 1996 by The Perl Porters <F<perl5-porters@perl.org>>.
+=item C<×>
+ getlogin()         × å
```
tonycoz: I expect this should be removed.

khwilliamson:
On 10/4/23 17:10, Tony Cook wrote:
> 
> 
> In dist/ExtUtils-ParseXS/lib/perlxs.pod 
> <https://github.com/Perl/perl5/pull/21529#discussion_r1346569440>:
> 
>> +=item C<Ꞧ>
> +
> 
> This is horribly mangled by the github pod viewer, and similarly in the 
> table.
> 
> https://github.com/khwilliamson/perl5/blob/perlxs/dist/ExtUtils-ParseXS/lib/perlxs.pod <https://github.com/khwilliamson/perl5/blob/perlxs/dist/ExtUtils-ParseXS/lib/perlxs.pod>

The first line already is an =encoding line.  Things appear to display 
fine to me except in the verbatim lines, which must be some bug that I'm 
surprised hasn't been found before,.

tonycoz:
I was looking at https://github.com/khwilliamson/perl5/blob/perlxs/dist/ExtUtils-ParseXS/lib/perlxs.pod

image
image

jkeenan:
@khwilliamson, do you wish us to proceed on this p.r.?  Thanks.

khwilliamson:
after it works

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2208,11 +2184,949 @@ In Makefile.PL add -ltirpc and -I/usr/include/tirpc.
...
+L</Dealing with embedded perls and threads>.
+
+A problem with changing the locale of a single category is that mojibake
+can arise on some platforms if the C<LC_CTYPE> category and the changed one
+are not the same.  The preprocessor directive
+C<LIBC_HANDLES_MISMATCHED_CTYPE> is defined if perl believes this
+platform doesn't have this issue.  Otherwise, you may have to change
+more than one category to accomplish your task.
+
+The exception to you being called with the numeric parts of the locale
```
jkeenan: "The exception to you being called ..."  Wording is awkward.

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2208,11 +2184,949 @@ In Makefile.PL add -ltirpc and -I/usr/include/tirpc.
...
+
+=item *
+
+raises a signal that functions in other instances are sensitive to
+
+=back
+
+If your code doesn't do any of these things, nor depends on any of their
+values, then Congratulations!!, you don't have to worry about MULTIPLICITY
+nor threading.  But wait, a surprising number of libc functions do
```
jkeenan: Stylistic nits:  I expect the word `nor` to be paired with `neither`.  I would word this sentence:
```
If your code doesn't do any of these things or depend upon any of their
values, then Congratulations!!, you don't have to worry about MULTIPLICITY
or threading.
```

dist/ExtUtils-ParseXS/lib/perlxs.pod
```diff
@@ -2208,11 +2184,949 @@ In Makefile.PL add -ltirpc and -I/usr/include/tirpc.
...
+Therefore, they must be called at least once before switching to
+multi-threads:
+
+ getutent()  getutline()  getutxid()    mallinfo()  valloc()
+ getutid()   getutxent()  getutxline()  pvalloc()
+
+=head3 Functions that are thread-safe when called with appropriate arguments
+
+Some of the functions are thread-safe if called with arguments that
+comply with certain (easily-met) restrictions.  These are:
```
jkeenan: We don't need the hyphen there, just a wordspace.

jkeenan (dismissed):
"The exception to you being called ..."  Wording is awkward.

(In general I can make only small stylistic comments.  The C and POSIX discussed in these docs is something I've never learned.)

khwilliamson:
I think this should be a new pod, because it can be generated (and most of the lists were generated)  But what to name it?

dist/ExtUtils-ParseXS/lib/perlxs.pod
tonycoz: "nor" without "neither" is rare, but it is valid. See method 2 at https://www.wikihow.com/Use-Nor

If it were to be rephrased to avoid "nor" I think it would need to be:
```
If your code doesn't do any of these things and does not depend on any of their
values, then Congratulations!!, you don't have to worry about MULTIPLICITY
nor threading.
```

tonycoz:
I could the thread parts going into perlthrtut (which has a small section on reentrancy) or perlguts (or a new perlthrguts)

The locale parts could go into perlguts or a new perllocguts.

dist/ExtUtils-ParseXS/lib/perlxs.pod
jkeenan: Well, that still retains a 'nor'.  Shouldn't the last phrase be:
```
you don't have to worry about MULTIPLICITY
or threading.
```

khwilliamson: I still like the original.  Being a dinosaur, I consulted Strunk and White "The Elements of Style" and Fowler's "Modern English Usage", and neither is not required for nor.  I prefer nor here because it emphasizes the negative aspect.

khwilliamson:
Much of the data for this PR came from a __DATA__ section for generating another file for 5.41.  I think it would be best to make the contents of this PR be able to be generated from that same DATA section, to avoid redundancy, and make it easier to update as needed.  I'm ok with separating threads vs non-threads, but I think the locale stuff shouldn't be separated from the non-locale stuff, because it is too easy to think, "I'm not using locales" and skip reading a separate file, whereas they may very well apply to you.  Note that the text of this applies to core development as well as XS modules.

tonycoz:
> Note that the text of this applies to core development as well as XS modules.

It could all go in perlguts, which is generally things relevant to both XS and core.

dist/ExtUtils-ParseXS/lib/perlxs.pod
khwilliamson: changed

khwilliamson:
I started to put this in perlguts, but the description of that file doesn't really fit this: 

> This document attempts to describe how to use the Perl API, as well as
to provide some info on the basic workings of the Perl core.

I had forgotten about `perlclib`, which some of this PR duplicates, but perlclib had more extensive advice.  So, I've modified the PR to instead add things to that pod.

PR #21532, opened 2023-09-28
Add builtin function load_module

mfontani:
As per PPC006[1]

[1]: https://github.com/Perl/PPCs/blob/main/ppcs/ppc0006-load-module.md

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -474,6 +474,23 @@ XS(XS_builtin_indexed)
     XSRETURN(retcount);
 }
 
+XS(XS_builtin_load_module);
+XS(XS_builtin_load_module)
+{
+    dXSARGS;
+    if (items != 1)
+        croak_xs_usage(cv, "arg");
+    SV *module_name = newSVsv(ST(0));
+    if (!SvPOK(module_name)) {
+        SvREFCNT_dec(module_name);
+        croak_xs_usage(cv, "defined string");
+    }
+    load_module(PERL_LOADMOD_NOIMPORT, module_name, NULL, NULL);
+    /* The loaded module's name is left intentionally on the stack for the
+     * caller's benefit, and becomes load_module's return value. */
+    XSRETURN(1);
+}
+
 static OP *ck_builtin_funcN(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
 {
     const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
@@ -515,6 +532,8 @@ static const struct BuiltinFuncDescriptor builtins[] = {
 
     { "created_as_string", NO_BUNDLE, &XS_builtin_created_as_string, &ck_builtin_func1, 0, true },
     { "created_as_number", NO_BUNDLE, &XS_builtin_created_as_number, &ck_builtin_func1, 0, true },
+
+    { "load_module", NO_BUNDLE, &XS_builtin_load_module, &ck_builtin_func1, 0, true },
 
     /* list functions */
     { "indexed",          SHORTVER(5,39), &XS_builtin_indexed,          &ck_builtin_funcN, 0, false },
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -27,6 +27,7 @@ builtin - Perl pragma to import built-in utility functions
         trim
         is_tainted
         export_lexically
+        load_module
     );
 
     use builtin ':5.40';  # most of the above
@@ -370,6 +371,25 @@ Z<>
 This must be called at compile time; which typically means during a C<BEGIN>
 block. Usually this would be used as part of an C<import> method of a module,
 when invoked as part of a C<use ...> statement.
+
+=head2 load_module
+
+    load_module($module_name);
+
+This function is currently B<experimental>.
+
+Loads a named module from the inclusion paths (C<@INC>).  C<$module_name> must
+be a string that provides a module name.  It cannot be omitted, and providing
+an invalid module name will result in an exception.  Not providing any argument
+results in a compilation error.  Returns the loaded module's name on success.
+
+The effect of C<load_module>-ing a module is mostly the same as C<require>-ing,
+down to the same error conditions when the module does not exist, does not
+compile, or does not evaluate to a true value.  See also
+L<the C<module_true> feature|feature/"The 'module_true' feature">.
+
+C<load_module> can't be used to require a particular version of Perl, nor can
+it be given a bareword module name as an argument.
 
 =head1 SEE ALSO
 
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -444,6 +444,170 @@ TODO: {
     is($HASH{key}, "val", 'Lexically exported hash is accessible');
 }
 
+# load_module
+{
+    use builtin qw( load_module );
+    use feature qw( try );
+    no warnings 'experimental::try';
+    my ($ok, $e);
+
+    # Can't really test this sans string eval, as it's a compilation error:
+    eval 'load_module();';
+    $e = $@;
+    ok($e, 'load_module(); fails');
+    like($e, qr/^Not enough arguments for builtin::load_module at/, 'load_module(); fails with correct error');
+    eval 'load_module;';
+    $e = $@;
+    ok($e, 'load_module; fails');
+    like($e, qr/^Not enough arguments for builtin::load_module at/, 'load_module; fails with correct error');
+
+    # Failure to load module croaks
+    try {
+        load_module(undef);
+    } catch ($e) {
+        ok($e, 'load_module(undef) fails');
+        like($e, qr/^Usage: builtin::load_module\(defined string\)/, 'load_module(undef) fails with correct error');
+    };
+    try {
+        load_module(\"Foo");
+    } catch ($e) {
+        ok($e, 'load_module(\"Foo") fails');
+        like($e, qr/^Usage: builtin::load_module\(defined string\)/, 'load_module(\"Foo") fails with correct error');
+    };
+    try {
+        load_module(["Foo"]);
+    } catch ($e) {
+        ok($e, 'load_module(["Foo"]) fails');
+        like($e, qr/^Usage: builtin::load_module\(defined string\)/, 'load_module(["Foo"]) fails with correct error');
+    };
+    try {
+        load_module('5.36');
+    }
+    catch ($e) {
+        ok($e, 'load_module("5.36") fails');
+        like($e, qr/^Can't locate 5[.]36[.]pm in \@INC/, 'load_module("5.36") fails with correct error');
+    };
+    try {
+        load_module('v5.36');
+    }
+    catch ($e) {
+        ok($e, 'load_module("v5.36") fails');
+        like($e, qr/^Can't locate v5[.]36[.]pm in \@INC/, 'load_module("v5.36") fails with correct error');
+    };
+    try {
+        load_module("Dies");
+        fail('load_module("Dies") succeeded!');
+    }
+    catch ($e) {
+        ok($e, 'load_module("Dies") fails');
+        like($e, qr/^Can't locate Dies[.]pm in \@INC/, 'load_module("Dies") fails with correct error');
+    }
+    my $module_name = 'Dies';
+    try {
+        load_module($module_name);
+        fail('load_module($module_name) $module_name=Dies succeeded!');
+    }
+    catch ($e) {
+        ok($e, 'load_module($module_name) $module_name=Dies fails');
+        like($e, qr/^Can't locate Dies[.]pm in \@INC/, 'load_module($module_name) $module_name=Dies fails with correct error');
+    }
+    $module_name =~ m!(\w+)!;
+    try {
+        load_module($1);
+        fail('load_module($1) from $module_name=Dies succeeded!');
+    }
+    catch ($e) {
+        ok($e, 'load_module($1) from $module_name=Dies fails');
+        like($e, qr/^Can't locate Dies[.]pm in \@INC/, 'load_module($1) from $module_name=Dies fails with correct error');
+    }
+    "Dies" =~ m!(\w+)!;
+    try {
+        load_module($1);
+        fail('load_module($1) from "Dies" succeeded!');
+    }
+    catch ($e) {
+        ok($e, 'load_module($1) from "Dies" fails');
+        like($e, qr/^Can't locate Dies[.]pm in \@INC/, 'load_module($1) from "Dies" fails with correct error');
+    }
+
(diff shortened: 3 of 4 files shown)


lib/builtin.t
```diff
@@ -444,6 +444,36 @@ TODO: {
     is($HASH{key}, "val", 'Lexically exported hash is accessible');
 }
 
+# load_module
+{
+    use builtin qw( load_module );
+    my ($ok, $e);
+
+    # Failure to load module croaks
+    $ok = eval 'load_module("Foo");';
```
leonerd: No need to string-eval this one. For simply capturing exceptions, you probably want to `eval { ... }` or even consider using the `try` feature.

And the same for the remaining cases in the test file.

lib/builtin.pm
```diff
@@ -371,6 +372,21 @@ This must be called at compile time; which typically means during a C<BEGIN>
 block. Usually this would be used as part of an C<import> method of a module,
 when invoked as part of a C<use ...> statement.
 
+=head2 load_module
+
+    load_module($module_name);
+
+This loads a named module from the inclusion paths (C<@INC>).  C<$module_name>
```
leonerd: No need for the particle here; just begin the sentence "Loads a named module ..."

lib/builtin.pm
```diff
@@ -371,6 +372,21 @@ This must be called at compile time; which typically means during a C<BEGIN>
...
+must be a string that provides a module name.  It cannot be omitted, and
+providing an invalid module name will result in an exception.
+
+The effect of C<load_module>-ing a module is the same as C<require>-ing, down
+to the same error conditions when the module does not exist, does not compile,
+or does not evalute to a true value.
```
leonerd: I'd word this more, "does not yield a true value at the end of its scope". Plus maybe a mention of the new `yield_true` feature of `use v5.38`.

builtin.c
```diff
@@ -474,6 +474,25 @@ XS(XS_builtin_indexed)
...
+    SV *st0 = ST(0);
+    SV *module_name = newSVsv(st0);
+    ENTER;
+    lex_start(NULL, NULL, 0);
+    I32 floor = start_subparse(FALSE, 0);
+    OP *op = newSVOP(OP_CONST, 0, module_name);
+    op->op_private |= OPpCONST_BARE;
+    utilize(0, floor, NULL, op, NULL);
+    LEAVE_SCOPE(floor);
+    LEAVE;
```
leonerd: My apologies - I think I sent you down the wrong path in my original email.

Having looked more into this since then I've remembered we have a `load_module` function at the C level as well. That would be a far simpler way to do all of this.

```c
load_module(PERL_LOADMOD_NOIMPORT, module_name, NULL, NULL);
```

This is the one I was thinking of when I mentioned `utilize()`. Sorry for the misdirection.

lib/builtin.pm
```diff
@@ -371,6 +372,21 @@ This must be called at compile time; which typically means during a C<BEGIN>
 block. Usually this would be used as part of an C<import> method of a module,
 when invoked as part of a C<use ...> statement.
 
+=head2 load_module
+
+    load_module($module_name);
```
leonerd, line 377: As it's new (and therefore experimental) don't forget to copy the "This function is experimental" text from one of the other function docs.

leonerd (changes requested):

lib/builtin.t
Grinnz: Seems somewhat fragile to rely on the system not having a Foo module to load, but maybe that is acceptable for core tests?

demerphq: It should just load something we have that dies during load. Eg, t/lib/Dies.pm

mfontani:
@leonerd looks like this should be all, please have another look.

lib/builtin.pm
mfontani: Done that in latest commit.

builtin.c
```diff
@@ -474,6 +474,17 @@ XS(XS_builtin_indexed)
     XSRETURN(retcount);
 }
 
+XS(XS_builtin_load_module);
+XS(XS_builtin_load_module)
+{
+    dXSARGS;
+    if (items != 1)
+        croak_xs_usage(cv, "arg");
+    SV *st0 = ST(0);
```
Leont: This line seems unnecessary, why not go straight for `SV *module_name = newSVsv(ST(0));`?

mfontani: Fair point. Rebased and squashed, thanks.

lib/builtin.pm
```diff
@@ -371,6 +372,24 @@ This must be called at compile time; which typically means during a C<BEGIN>
...
+
+This function is currently B<experimental>.
+
+Loads a named module from the inclusion paths (C<@INC>).  C<$module_name> must
+be a string that provides a module name.  It cannot be omitted, and providing
+an invalid module name will result in an exception.
+
+The effect of C<load_module>-ing a module is the same as C<require>-ing, down
+to the same error conditions when the module does not exist, does not compile,
+or does not evalute to a true value. See also
```
leonerd: Typo "evaluate"

lib/builtin.t
```diff
@@ -444,6 +444,50 @@ TODO: {
...
+    }
+    $module_name = 'strict';
+    try {
+        load_module($module_name);
+        pass('load_module($module_name) $module_name=strict worked');
+    }
+    catch ($e) {
+        fail('load_module($module_name) $module_nam=strict errored: ' . $e);
+    }
+}
```
leonerd, line 609: Another test worth adding is the usual magic one; something like:

```perl
"Dies" =~ m/(\w+)/;
load_module($1);
...
```

and then check that it still behaves correctly. I believe your implementation currently won't, but it can be fixed by adding a call to `SvGETMAGIC(ST(0));` on a line before the `newSVsv`.

leonerd (changes requested):

lib/builtin.pm
mfontani: Well spotted. Updated.

lib/builtin.t
mfontani: I've added four of the type `load_module($1)`, two for failure and two for success: `$module_name =~ m!(\w+)!` and `"..." =~ m!(\w+)!` and all four seem to pass (on my machine) without needing to `SvGETMAGIC(...)`.  I'll watch test suite failures in case I can find other ways to reliably repro the requirement for having to add that call.

mfontani: Looks like those passed everywhere despite the lack of `SvGETMAGIC(..)`. Would you know why it should in principle be required, or have a suggestion for a test what would highlight it being required? Looks otherwise to not be required...

leonerd: Ooohright; yes. Because you're doing `newSVsv()` that already invokes get magic on its source argument.

In that case the code is fine but remember to add those tests in case of future edits.

mfontani: Tests for that are there:
- https://github.com/Perl/perl5/pull/21532/files#diff-3b3171c64e546f883b025fda758c78654dc5e73380370bf557acd6d9a10f7991R472-R489 for the failure to load a module (`$module_name = "Dies"` vs "Dies", then perform `=~` on either)
- https://github.com/Perl/perl5/pull/21532/files#diff-3b3171c64e546f883b025fda758c78654dc5e73380370bf557acd6d9a10f7991R507-R522 for the success (ditto, for `strict` instead of `Dies`

haarg:
This appears to segfault when given unexpected input, like undef or a reference.

Also, the `Perl_load_module` being used internally uses a `BEGIN` block. This isolates some bits of hints etc and impacts the caller stack. I wouldn't say that is specifically bad, but it is a difference from `require`, which the documentation implies shouldn't be the case.

The return value from this function is not tested, but seems to just be the same as its input module name. This is not obviously intentional, and not documented. The PPC this is meant to implement also doesn't state what the return value should be.

leonerd:
Ahyes; PPC0006 doesn't talk about the return value. I'd suggest it should either be literally nothing, or just `PL_sv_yes`.

mfontani:
Thanks @haarg & @leonerd, updated:

- croak if `!SVpOK(...)` so it croaks on literal undef or reference, etc & add tests for that
- return `XSRETURN_YES` (i.e. "a true value"), document that and test that
- add a test for tied scalar, and for `load_module()` failing to compile (as string eval...)

Not handled, "really":

> Also, the Perl_load_module being used internally uses a BEGIN block. This isolates some bits of hints etc and impacts the caller stack. I wouldn't say that is specifically bad, but it is a difference from require, which the documentation implies shouldn't be the case.

I'm unsure how to be able to both call the `load_module` _and_ **not** make it do that... so I've amended the documentation stating it's pretty much the same as `require` _but_ it "internally uses a `C<BEGIN>` block to perform its operations".  The difference is subtle, and there's corner cases, but it's _mostly_ like require.

Let me know your thoughts.

builtin.c
```diff
@@ -474,6 +474,19 @@ XS(XS_builtin_indexed)
...
 
+XS(XS_builtin_load_module);
+XS(XS_builtin_load_module)
+{
+    dXSARGS;
+    if (items != 1)
+        croak_xs_usage(cv, "arg");
+    SV *module_name = newSVsv(ST(0));
+    if (!SvPOK(module_name))
+        croak_xs_usage(cv, "defined string");
```
tonycoz: The call to load_module() will release `module_name`, but `module_name` isn't released if the croak is called.

mfontani: Thanks! I've added a `SvREFCNT_dec(module_name);` prior to calling `croak_xs_usage(cv, "defined string");`. That should do it, right?

tonycoz (approved):

mfontani:
Is there anything else I should be doing for this?

lib/builtin.pm
```diff
@@ -371,6 +372,26 @@ This must be called at compile time; which typically means during a C<BEGIN>
...
+
+Loads a named module from the inclusion paths (C<@INC>).  C<$module_name> must
+be a string that provides a module name.  It cannot be omitted, and providing
+an invalid module name will result in an exception.  Not providing any argument
+results in a compilation error.  Returns a true value on success.
+
+The effect of C<load_module>-ing a module is mostly the same as C<require>-ing,
+down to the same error conditions when the module does not exist, does not
+compile, or does not evaluate to a true value.  It internally uses a C<BEGIN>
+block to perform its operations.  See also
```
leonerd: Probably best not to explicitly document that the `BEGIN` block is used, in case we change the implementation at another time. Keeping it somewhat vague gives us the freedom to change it in future.

Moreover probably better to downplay the "it's like require but..." in favour of simply describing what it does in its own terms. At the end you could say that it's a bit similar to require, without specifically detailing exactly the differences.

leonerd (commented):
I think this is getting close to ready now.

We discussed this at the PSC meeting today. I've put some comments about the docs above.

We further thought about what the return value ought to be. The original PPC document doesn't discuss it. The implementation you've gone for is `return true`, which seems fine. But also a valid choice could be returning the module name itself. It would permit things like

```perl
my @objects = map { load_module($_)->new } @class_names;

BEGIN { load_module("Some::Thing")->import(qw( some symbol names )); }
```

Neither of these is particularly difficult to do as it stands, but that might be an extra convenience. Either way, probably not worth blocking this PR for that. If we decide to change it we can alter it later down the line.

lib/builtin.pm
mfontani: Thanks. Fair point. See if the updated version - which removes that part - is better.

It merely states: "the effect of ... is _mostly the same as_ requiring it".

mfontani:
> But also a valid choice could be returning the module name itself. It would permit things like
> 
> ```perl
> my @objects = map { load_module($_)->new } @class_names;
> 
> BEGIN { load_module("Some::Thing")->import(qw( some symbol names )); }
> ```

I had that initially. Pushed a separate commit which does that.

If approved, I'd like to squash them though.

mfontani:
Added a useful test for `load_module(...)->import(...)`.

mfontani:
Re-pushed (and squashed) after conflicts were found in `pod/perldelta.pod`.

lib/builtin.pm
```diff
@@ -371,6 +372,25 @@ This must be called at compile time; which typically means during a C<BEGIN>
...
+=head2 load_module
+
+    load_module($module_name);
+
+This function is currently B<experimental>.
+
+Loads a named module from the inclusion paths (C<@INC>).  C<$module_name> must
+be a string that provides a module name.  It cannot be omitted, and providing
+an invalid module name will result in an exception.  Not providing any argument
+results in a compilation error.  Returns a true value on success.
```
leonerd: Since the implementation and tests now work to the idea of returning the module name, these docs need fixing.

builtin.c
```diff
@@ -474,6 +474,20 @@ XS(XS_builtin_indexed)
...
+    dXSARGS;
+    if (items != 1)
+        croak_xs_usage(cv, "arg");
+    SV *module_name = newSVsv(ST(0));
+    if (!SvPOK(module_name)) {
+        SvREFCNT_dec(module_name);
+        croak_xs_usage(cv, "defined string");
+    }
+    load_module(PERL_LOADMOD_NOIMPORT, module_name, NULL, NULL);
+}
```
leonerd, line 492: I'd add a comment here specifically pointing out we leave the module name on the stack, so that becomes its return value. It's intentional. Also probably good to put an `XSRETURN(1)` to ensure we return exactly one item even in list context.

leonerd (changes requested):
Generally good, a couple more little comments need adding.

lib/builtin.pm
mfontani: Thanks, updated with:

> Returns the loaded module's name on success.

builtin.c
mfontani: Thanks! Updated with:

```c
    /* The loaded module's name is left intentionally on the stack for the
     * caller's benefit, and becomes load_module's return value. */
    XSRETURN(1);
```

leonerd (approved):
LGTM

If there's no other objections to this over the wekeend, I say we merge it on Monday.

Leont:
[PerlIO::import](https://github.com/Perl/perl5/blob/blead/lib/PerlIO.pm#L22) may be a good first user. It's core-only so is free to use such a feature.

mfontani:
> [PerlIO::import](https://github.com/Perl/perl5/blob/blead/lib/PerlIO.pm#L22) may be a good first user. It's core-only so is free to use such a feature.

Sounds like a good idea. I will get on it once this is merged, though, and not as part of this PR.

iabyn:
I think the POD for builtin::load_module should state (briefly) its raison d'etre, i.e. what problem does it solve which require can't do.  Which is that require doesn't accept module names (apart from barewords) - it only accepts filenames.

PR #21536, opened 2023-09-30
regexec.c - make find_byclass() work with (*SKIP) and friends

### LINKED ISSUE #21534: Optimization bug preventing a regex containing (*SKIP)(*FAIL) to work correctly
The following regex doesn't match while it should.
```
"dir/file.mp3" =~ m{ ([^/]+) (?: / (*SKIP)(*FAIL) | \z ) }x
```
Disabling the optimization (I don't know which one causes this) with `(?{})` makes the match succeed as it should.
```
say $& if "dir/file.mp3" =~ m{ ([^/]+) (?: (?{}) / (*SKIP)(*FAIL) | \z ) }x
```

Additionally, Using a possessive quantifier makes the bug go away.
```
$ perl -E 'say $& if "dir/file.mp3" =~ m{ ([^/]++) (?: / (*SKIP)(*FAIL) | \z ) }x'
file.mp3
```

When running the faulty regex with `use re "debug";` the message `Contradicts stclass... [regexec_flags]` appears.
```
$ perl -E 'use re "debug"; say $& if "dir/file.mp3" =~ m{ ([^/]+) (?: / (*SKIP)(*FAIL) | \z ) }x'
Compiling REx " ([^/]+) (?: / (*SKIP)(*FAIL) | \z ) "
Final program:
   1: OPEN1 (3)
   3:   PLUS (6)
   4:     NANYOFM[/] (0)
   6: CLOSE1 (8)
   8: BRANCH (buf:1/1) (16)
  10:   EXACT </> (12)
  12:   SKIP (14)
  14:   OPFAIL (20)
  16: BRANCH (buf:1/1) (FAIL)
  18:   EOS (20)
  19: TAIL (20)
  20: END (0)
stclass NANYOFM[/] plus minlen 1
Matching REx " ([^/]+) (?: / (*SKIP)(*FAIL) | \z ) " against "dir/file.mp3"
Matching stclass NANYOFM[/] against "dir/file.mp3" (12 bytes)
   0 <> <dir/file.m>         |   0| 1:OPEN1(3)
   0 <> <dir/file.m>         |   0| 3:PLUS(6)
                             |   0| NANYOFM[/] can match 3 times out of 2147483647...
   3 <dir> </file.mp3>       |   1|  6:CLOSE1(8)
   3 <dir> </file.mp3>       |   1|  8:BRANCH (buf:1/1)(16)
   3 <dir> </file.mp3>       |   2|   10:EXACT </>(12)
   4 <dir/> <file.mp3>       |   2|   12:SKIP(14)
   4 <dir/> <file.mp3>       |   3|    14:OPFAIL(20)
                             |   3|    failed...
                             |   2|   failed...
Contradicts stclass... [regexec_flags]
Match failed
Freeing REx: " ([^/]+) (?: / (*SKIP)(*FAIL) | \z ) "
```

demerphq:
This fixes https://github.com/Perl/perl5/issues/21534

When the previous_occurence_end logic was added it did not account for patterns where "s" is updated by regtry(). See 21d1ed5 which introduced this logic. That patch seemed to assume that regtry() would not alter s, but it does. This fix seems to work around the problem for the verb's, but it feels like a bodge instead of a proper fix, which I think would involve reworking the previous_occurrence_end logic to take into account that regtry() can modify s. But for now this should work.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -1934,13 +1934,16 @@ STMT_START {                                                                    
         s++;                                                   \
     }
 
+#define LAST_REGTRY_SKIPPED_FORWARD(reginfo) (reginfo->cutpoint)
+
 /* We keep track of where the next character should start after an occurrence
  * of the one we're looking for.  Knowing that, we can see right away if the
  * next occurrence is adjacent to the previous.  When 'doevery' is FALSE, we
  * don't accept the 2nd and succeeding adjacent occurrences */
 #define FBC_CHECK_AND_TRY                                           \
         if (   (   doevery                                          \
-                || s != previous_occurrence_end)                    \
+                || s != previous_occurrence_end                     \
+                || LAST_REGTRY_SKIPPED_FORWARD(reginfo) )           \
             && (   reginfo->intuit                                  \
                 || (s <= reginfo->strend && regtry(reginfo, &s))))  \
         {                                                           \
@@ -4356,6 +4359,12 @@ Perl_regexec_flags(pTHX_ REGEXP * const rx, char *stringarg, char *strend,
 
 /*
  - regtry - try match at specific point
+
+ NOTE: *startpos may be modifed by regtry() to signal to the caller
+       that the next match should start at a specific position in the
+       string. The macro LAST_REGTRY_SKIPPED_FORWARD(reginfo) can be
+       used to detect when this has happened.
+
  */
 STATIC bool			/* 0 failure, 1 success */
 S_regtry(pTHX_ regmatch_info *reginfo, char **startposp)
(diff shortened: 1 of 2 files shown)


khwilliamson:
Bisecting this issue found that it likely never worked properly in its history, long before the `previous_occurrence_end` logic was added,  I think that indicates as well that this is a bodge, and this bisection result might aid in finding the more underlying fix

demerphq:
It might be that this bodge is the minimal patch to accomplish the goal, even though it doesn't feel quite right. The only alternatives i can think to this are: 1. change regtry() so that when s is modified we signal back to the caller, which is much the same as checking reginfo, or 2. actually check if s was changed by regtry on failure, and then set previous_occurence_end to 0, instead of to s when it was.

demerphq:
I've reworked this patch to hide the check inside of a macro and to include a comment about it in the docs for regtry().

khwilliamson (approved):

PR #21537, opened 2023-10-01
Get /o off of perlop.pod

jidanni:
Elsewhere in the page we see
> The bottom line is that using "/o" is almost never a good idea.

So I'll get rid of /o here too.

P.S., I also get rid of a newline, based on
```
$ echo 'while(<>){}if 0;'|perltidy
while (<>) { }
if 0;
```


Grinnz (approved):
Agreed; there is no reason to show this in a code example.

PR #21540, opened 2023-10-04
Modernise Class::Struct-using core modules

ilmari:
These modules couldn't inherit from Exporter because they use Class::Struct, and were written before Exporter could export its import method.

Now that it can, update them to do that, and remove the pointless BEGIN block around the Exporter setup.

In passing, bump the use VERSION statement to 5.38, but disable signatures, since the modules use prototypes.

### DIFF
--- a/lib/File/stat.pm
+++ b/lib/File/stat.pm
@@ -1,6 +1,5 @@
 package File::stat 1.14;
 use v5.38;
-no feature 'signatures'; # we use prototypes
 
 use warnings::register;
 use Carp;
@@ -185,7 +184,7 @@ struct 'File::stat' => [
      }
 ];
 
-sub populate (@) {
+sub populate {
     return unless @_;
     my $stob = new();
     @$stob = (
@@ -195,9 +194,9 @@ sub populate (@) {
     return $stob;
 } 
 
-sub lstat ($)  { populate(CORE::lstat(shift)) }
+sub lstat :prototype($) { populate(CORE::lstat(shift)) }
 
-sub stat ($) {
+sub stat :prototype($) {
     my $arg = shift;
     my $st = populate(CORE::stat $arg);
     return $st if defined $st;
--- a/lib/File/stat.pm
+++ b/lib/File/stat.pm
@@ -1,19 +1,13 @@
-package File::stat;
-use 5.006;
+package File::stat 1.14;
+use v5.38;
+no feature 'signatures'; # we use prototypes
 
-use strict;
-use warnings;
 use warnings::register;
 use Carp;
 use constant _IS_CYGWIN => $^O eq "cygwin";
 
 BEGIN { *warnif = \&warnings::warnif }
 
-our(@EXPORT, @EXPORT_OK, %EXPORT_TAGS);
-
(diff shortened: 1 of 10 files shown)


lib/Time/gmtime.pm
```diff
@@ -1,27 +1,24 @@
...
 );
- 
-BEGIN { 
-    use Exporter   ();
-    @ISA         = qw(Exporter Time::tm);
-    @EXPORT      = qw(gmtime gmctime);
-    @EXPORT_OK   = qw(  
+
+use Exporter 'import';
+our @ISA         = qw(Exporter Time::tm);
```
JRaspass: Surely you don't want this anymore? `parent` will do it's thing and we're importing Exporter's import now rather than subclassing it.

lib/File/stat.pm
```diff
@@ -1,38 +1,32 @@
...
-use strict;
-use warnings;
 use warnings::register;
 use Carp;
 use constant _IS_CYGWIN => $^O eq "cygwin";
 
 BEGIN { *warnif = \&warnings::warnif }
 
-our(@EXPORT, @EXPORT_OK, %EXPORT_TAGS);
+our $VERSION = '1.14';
```
JRaspass: Might as well move to this to the package line if we're going modern.

ilmari: Good point, done.

haarg:
Why not use `:prototype()` so we don't have to disable signatures?

ilmari:
> Why not use `:prototype()` so we don't have to disable signatures?

Because I was being lazy and didn't want to make too many unrelated changes to the primary objective.

ilmari:
> > Why not use `:prototype()` so we don't have to disable signatures?
> 
> Because I was being lazy and didn't want to make too many unrelated changes to the primary objective.

On second thought, I wanted too see how it would look like with signatures, so I switched to the `:prototype` syntax as a prerequisite. Actually adding signatures would technically be a change of API, so I've put that off for now.

lib/Time/gmtime.pm
```diff
@@ -34,8 +28,8 @@ sub populate (@) {
     return $tmob;
 } 
 
-sub gmtime (;$)    { populate CORE::gmtime(@_ ? shift : time)}
-sub gmctime (;$)   { scalar   CORE::gmtime(@_ ? shift : time)} 
+sub gmtime  :prototype(;$) { populate CORE::gmtime(@_ ? shift : time) }
+sub gmctime :prototype(;$) { scalar   CORE::gmtime(@_ ? shift : time) }
 
 1;
```
JRaspass: We can also drop all the `1;` lines now that we have the `5.38` feature bundle with `module_true`.

ilmari: good point, will fix

jkeenan (approved):
My impression is that the two commits in this p.r. are each substantive on their own terms and therefore should *not* be squashed before merging.  I tested each to ensure that all tests PASS.

haarg (approved):

rwp0 (approved):

PR #21544, opened 2023-10-05
Test2-Suite: remove problematic test

haarg:
Test2-Suite's t/acceptance/Workflow-Acceptance.t test is causing intermittent failures. Exodist has said:

> the test is not one that is necessary to run all the time, it should probably be an author test.

So for now, remove the test.


jkeenan:
> Test2-Suite's t/acceptance/Workflow-Acceptance.t test is causing intermittent failures. Exodist has said:
> 
> > the test is not one that is necessary to run all the time, it should probably be an author test.
> 
> So for now, remove the test.

Github CI failures suggest file needs to be removed from MANIFEST as well.

jkeenan (approved):
List of failures on our smoke-testing rigs [here](https://perl5.test-smoke.org/submatrix?test=../cpan/Test2-Suite/t/acceptance/Workflow-Acceptance.t&pversion=5.39.4).

PR #21545, opened 2023-10-06
Get Test2 tests working on VMS

craigberry:
List assignment to %ENV is not supported, and mongrel filenames produced by concatenating hard-coded slashes with the return values from File::Spec functions don't work.

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -1144,6 +1144,13 @@ our %Modules = (
             # https://github.com/Test-More/Test2-Suite/issues/208
             't/acceptance/Workflow-Acceptance.t',
         ],
+        'CUSTOMIZED'   => [
+            qw( t/modules/Plugin/SRand.t
+                t/modules/Require/AuthorTesting.t
+                t/modules/Require/EnvVar.t
+                t/modules/Tools/GenTemp.t
+              )
+        ],
     },
 
     'Text::Abbrev' => {
--- a/cpan/Test2-Suite/t/modules/Plugin/SRand.t
+++ b/cpan/Test2-Suite/t/modules/Plugin/SRand.t
@@ -33,9 +33,8 @@ sub intercept_2(&) {
 }
 
 {
-    local %ENV = %ENV;
-    $ENV{HARNESS_IS_VERBOSE} = 1;
-    $ENV{T2_RAND_SEED} = 1234;
+    local $ENV{HARNESS_IS_VERBOSE} = 1;
+    local $ENV{T2_RAND_SEED} = 1234;
 
     my ($events, $warning);
     my $reseed_qr = qr/SRand loaded multiple times, re-seeding rand/;
--- a/cpan/Test2-Suite/t/modules/Require/AuthorTesting.t
+++ b/cpan/Test2-Suite/t/modules/Require/AuthorTesting.t
@@ -1,8 +1,7 @@
 use Test2::Bundle::Extended -target => 'Test2::Require::AuthorTesting';
 
 {
-    local %ENV = %ENV;
-    $ENV{AUTHOR_TESTING} = 0;
+    local $ENV{AUTHOR_TESTING} = 0;
     is($CLASS->skip(), 'Author test, set the $AUTHOR_TESTING environment variable to run it', "will skip");
 
     $ENV{AUTHOR_TESTING} = 1;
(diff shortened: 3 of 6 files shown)


demerphq:
FWIW, I don't think Test2 should be using %ENV for state at all. What happens if someone needs to test something that uses %ENV? Its an unforced design decision as well, it doesn't have to work like that, it just does.  Now that it is in core it shouldn't mess with %ENV IMO, I objected once to this when it was not part of core, but was ignored, but now it is in core IMO its a problem waiting to be reported. (And yes I have written tests that were broken by Test2 using %ENV for state.)

Leont:
Any reason for using `delete local` and not just localized assignment?

craigberry:
> FWIW, I don't think Test2 should be using %ENV for state at all. What happens if someone needs to test something that uses %ENV? Its an unforced design decision as well, it doesn't have to work like that, it just does. Now that it is in core it shouldn't mess with %ENV IMO, I objected once to this when it was not part of core, but was ignored, but now it is in core IMO its a problem waiting to be reported. (And yes I have written tests that were broken by Test2 using %ENV for state.)

That's interesting.  I was not aware of this history and was just chasing test failures, not really thinking about the design of what to me is a new module. Do you have thoughts on a preferable state mechanism, and do you think that should be addressed as part of this PR or as a new issue?

craigberry:
> Any reason for using `delete local` and not just localized assignment?

Not really.  I guess I thought it might make it clearer that a potential conflict with the real environment was being avoided, but as a practical matter, ```delete local $ENV{'FOO'}``` really gets you the same thing as ```local $ENV{'FOO'}```; do you think the latter is more readable?

Leont:
> do you think the latter is more readable?

I would prefer either
~~~
    local $ENV{AUTHOR_TESTING} = 0;
~~~
or 
~~~
    delete local $ENV{AUTHOR_TESTING};
~~~
over
~~~
    delete local $ENV{AUTHOR_TESTING};
    $ENV{AUTHOR_TESTING} = 0;
~~~

demerphq:
> Do you have thoughts on a preferable state mechanism, and do you think that should be addressed as part of this PR or as a new issue?

IMO the code should use a private hash in its own namespace for state, and copy anything relevant from env into that private hash very very very early. But no, I do not think that needs to be part of this patch. I was just grumbling aloud. It would not surprise me if the use of %ENV for state causes issues on platforms where the env is sensitive as it is on VMS.

I will note that I think @Leont has a point. In the case where you are doing:

```
    delete local @ENV{'HARNESS_IS_VERBOSE', 'T2_RAND_SEED'};
    $ENV{HARNESS_IS_VERBOSE} = 1;
    $ENV{T2_RAND_SEED} = 1234;
```
I think you should just do:

```
    local $ENV{HARNESS_IS_VERBOSE} = 1;
    local $ENV{T2_RAND_SEED} = 1234;
```

I wouldn't combine delete and local in the cases you have here, there doesn't seem to be any need to do so.

craigberry:
> I wouldn't combine delete and local in the cases you have here, there doesn't seem to be any need to do so.

Thanks for the feedback.  I have re-pushed with the simplified localization of %ENV entries.

haarg:
Poking @exodist so he's aware of this. I don't think he has much availability at the moment, so patching this in core first makes sense.

exodist:
If someone wants to make a PR to put these changes into the Test2::Suite dist, I can at least make time to merge and mint a release within the next week. @haarg @craigberry, @Leont and @leonerd I would be happy to give any of the 4 of you a commit bit to Test2-Suite and a merge bit on pause for simple fixes like this, to keep things moving even while I am in family emergency mode. Any of you 4 can ping or email me directly if you want the bits.

@demerphq Your request to change state, if there was already a ticket that I have neglected or ignored please reply with a link to it. If there is no ticket please open one on the test2-suite github repo and we can discuss it there. I do not want to muck up this tickets thread with that discussion.

Leont:
> If someone wants to make a PR to put these changes into the Test2::Suite dist, I can at least make time to merge and mint a release within the next week.

https://github.com/Test-More/Test2-Suite/pull/278

> @haarg @craigberry, @Leont and @leonerd I would be happy to give any of the 4 of you a commit bit to Test2-Suite and a merge bit on pause for simple fixes like this, to keep things moving even while I am in family emergency mode. Any of you 4 can ping or email me directly if you want the bits.

I'll skip on that one, too many other toolchain obligations already.

craigberry:
I've created an equivalent PR upstream:

https://github.com/Test-More/Test2-Suite/pull/279

I don't think I can be much help with ongoing maintenance, though -- I've got a ton of other test failures in other places that need tracking down.

Issue #21546, opened 2023-10-06
Intermittent failure of Perl ptest t/op/sigsystem


wlyu2:
This bug is found and tracked in the Yocto Project (Bug 15136):
https://bugzilla.yoctoproject.org/show_bug.cgi?id=15136

Alexandre Belloni provided the following error output in the above link:
> `# Failed test 3 - Reaped only one process at op/sigsystem.t line 42`
> `FAIL: t/op/sigsystem`

Note that this ptest failure is intermittent. 

This issue is potentially related to the following two currently open issues in Perl:
- https://github.com/Perl/perl5/issues/12802
- https://github.com/Perl/perl5/issues/12896

jkeenan:
p.r. https://github.com/Perl/perl5/pull/21547 has been submitted with reference to this.

jkeenan:
Closed by manual processing of https://github.com/Perl/perl5/pull/21547 in 75d974a58c

PR #21547, opened 2023-10-06
Fix intermittent failure of test t/op/sigsystem.t

wlyu2:
This fix addresses the intermittent failure of the test t/op/sigsystem.t by improving its robustness. Before the fix, this test waits a hard-coded amount of time in the parent process for the child process to exit, and the child process may not be able to exit soon enough. With this fix, the parent process in this test polls for whether the SIGCHLD handler reaped the child process for at most 25 seconds.

### DIFF
--- a/t/op/sigsystem.t
+++ b/t/op/sigsystem.t
@@ -37,7 +37,15 @@ SKIP: {
     test_system('with reaper');
 
     note("Waiting briefly for SIGCHLD...");
-    Time::HiRes::sleep(0.500);
+
+    # Wait at most 50 * 0.500 = 25.0 seconds for the child process to be 
+    # reaped. If the child process exits and gets reaped early, this polling
+    # loop will exit early. 
+
+    for (1..50) {
+	last if @pids;
+	Time::HiRes::sleep(0.500);
+    }
 
     ok(@pids == 1, 'Reaped only one process');
     ok($pids[0] == $pid, "Reaped the right process.") or diag(Dumper(\@pids));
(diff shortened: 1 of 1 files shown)


t/op/sigsystem.t
```diff
@@ -19,10 +19,12 @@ SKIP: {
     require Time::HiRes;
 
     my @pids;
+    my $reaped_child = 0;
```
mauke: I don't think you need this variable. Instead of testing `if $reaped_child`, you could just test `if @pids`.

t/op/sigsystem.t
```diff
@@ -50,8 +60,11 @@ sub test_system {
     my $got_zeroes      = 0;
 
     # This test is looking for a race between system()'s waitpid() and a
-    # signal handler.    Looping a few times increases the chances of
-    # catching the error.
+    # signal handler. The system() call is expected to not interfere with the 
+    # SIGCHLD signal handler. In particular, the wait() called within system() 
+    # is expected to reap the child process forked by system() before the 
+    # SIGCHLD signal handler is called. 
+    # Looping a few times increases the chances of # catching the error.
```
jkeenan: You should remove the second '`#`' character in line 67 above.

t/op/sigsystem.t
wlyu2: Good suggestion. Thanks!

t/op/sigsystem.t
wlyu2: Will fix this typo.

wlyu2:
The updated branch now addresses the above comments.

jkeenan:
> The updated branch now addresses the above comments.

Please run `perl Porting/updateAUTHORS.pl` so that we get your name and email address into `AUTHORS`.  Verify that all the metadata is correct by running `make test_porting` before re-pushing.  Thanks.

List:  apart from metadata questions, is this p.r. correct?

tonycoz:
It looks fine to me.

Fixes #21546 <- so merging closes the ticket

jkeenan:
> > The updated branch now addresses the above comments.
> 
> Please run `perl Porting/updateAUTHORS.pl` so that we get your name and email address into `AUTHORS`. Verify that all the metadata is correct by running `make test_porting` before re-pushing. Thanks.
> 

I took care of this manually and merged in commit 75d974a58c.  Thanks for your contribution.

Issue #21523, opened 2023-09-26
BBC: Blead Breaks Inline::CPP


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.39.4.

-----------------------------------------------------------------

BBC: Blead Breaks Inline::CPP

Please see http://fast-matrix.cpantesters.org/?dist=Inline::CPP

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.39.4:

Configured by cpan at Tue Sep 26 14:11:09 EDT 2023.

Summary of my perl5 (revision 5 version 39 subversion 4) configuration:
  Commit id: b5f24b1f49b3a977a4b039f4bf3a593941351661
  Platform:
    osname=linux
    osvers=6.1.55-0-lts
    archname=x86_64-linux
    uname='linux cjg-alpine3 6.1.55-0-lts #1-alpine smp preempt_dynamic sun, 24 sep 2023 23:14:02 +0000 x86_64 linux '
    config_args='-des -Dprefix=/home/cpan/bin/perl -Dscriptdir=/home/cpan/bin/perl/bin -Dusedevel -Duse64bitall'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='12.2.1 20220924'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
    Off_t='off_t'
    lseeksize=8
    alignbytes=8
    prototype=define
  Linker and Libraries:
    ld='cc'
[...]

PR #21552, opened 2023-10-10
default mark indexes back to I32 but allow selection of SSize_t too

tonycoz:
This also defines a new type to use for the mark stack which can be used by code to always use the correct size for pointers into the mark stack, and provides a fallback for that type in ppport.h for use with older perls.

If we are using I32 for mark indexes, then limit the size of the value stack to the limit of that type.

The idea here is to allow for a gradual migration to SSize_t stack indexes, though I expect most simple XS won't need any changes.


Leont (approved):

tonycoz:
Fixes #21523

PR #21557, opened 2023-10-12
Bisection: Document use of '--one-liner' switch

jkeenan:
... in the context of searching for first commit when a warning was emitted.

Allow for coding example to exceed 78-char line length.

### DIFF
--- a/t/porting/known_pod_issues.dat
+++ b/t/porting/known_pod_issues.dat
@@ -1,4 +1,4 @@
-# This file is the data file for t/porting/podcheck.t.
+# This file is the data file for porting/podcheck.t.
 # There are three types of lines.
 # Comment lines are white-space only or begin with a '#', like this one.  Any
 #   changes you make to the comment lines will be lost when the file is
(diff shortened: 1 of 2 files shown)


khwilliamson (approved):

t/porting/known_pod_issues.dat
```diff
@@ -1,4 +1,4 @@
-# This file is the data file for t/porting/podcheck.t.
+# This file is the data file for porting/podcheck.t.
```
hvds, line 1: Is this an intentional change? It doesn't appear related to the rest of the PR.

khwilliamson:
On 10/12/23 17:17, Hugo van der Sanden wrote:
> 
> 
> In t/porting/known_pod_issues.dat 
> <https://github.com/Perl/perl5/pull/21557#discussion_r1357539308>:
> 
>> -# This file is the data file for t/porting/podcheck.t.
> +# This file is the data file for porting/podcheck.t.
> 
> Is this an intentional change? It doesn't appear related to the rest of 
> the PR.

This is caused by regenerating the podcheck exceptions data file. 
Presumably this was necessary to do because of something else in the 
patch.  Depending on which directory you are in at the time of the 
regeneration, it has the initial 't/'.  I suppose it ought to be changed 
to not include that, but it's never been questioned before.

jkeenan:
> On 10/12/23 17:17, Hugo van der Sanden wrote: ***@***.**** commented on this pull request. ------------------------------------------------------------------------ In t/porting/known_pod_issues.dat <[#21557 (comment)](https://github.com/Perl/perl5/pull/21557#discussion_r1357539308)>: > -# This file is the data file for t/porting/podcheck.t. +# This file is the data file for porting/podcheck.t. Is this an intentional change? It doesn't appear related to the rest of the PR.
> This is caused by regenerating the podcheck exceptions data file. Presumably this was necessary to do because of something else in the patch. Depending on which directory you are in at the time of the regeneration, it has the initial 't/'. I suppose it ought to be changed to not include that, but it's never been questioned before.

@khwilliamson is correct.  It is simply the result of running `cd t; ./perl -I../lib porting/podcheck.t --regen; cd -`.  I noticed that before creating the p.r., but since it was not my doing, I let it stand.

PR #21566, opened 2023-10-17
don't clear IOK on @{"_<$file"} entries when a dbstate is freed

### LINKED ISSUE #21564: `${"_<a.pl"}[$i]` no longer numerical for pragmas
See [https://www.perlmonks.org/?node_id=11154956](https://www.perlmonks.org/?node_id=11154956)

perldebguts says:

> Each array `@{"_<$filename"}` holds the lines of $filename for a file
> compiled by Perl. [...] Values in this array are magical in numeric
> context: they compare equal to zero only if the line is not breakable.

The last bit is not always true since 5.36.

Take the program

```perl
#!/usr/bin/perl

use strict;
use warnings;

use v5.10;

say 'Hello, world!';
```

In 5.36 and 5.38, we see:

```
  DB<1> use Devel::Peek;

  DB<2> Dump( ${"_<a.pl"}[8] );
SV = PVMG(0x55d8627958e0) at 0x55d86251ff80
  REFCNT = 1
  FLAGS = (IOK,POK,IsCOW,pIOK,pPOK)
  IV = 94387850872920
  NV = 0
  PV = 0x55d8628074b0 "say 'Hello, world!';\n"\0
  CUR = 21
  LEN = 32
  COW_REFCNT = 1

  DB<3> Dump( ${"_<a.pl"}[6] );
SV = PVMG(0x55d862786d50) at 0x55d86251fec0
  REFCNT = 1
  FLAGS = (POK,pIOK,pNOK,pPOK)
  IV = 0
  NV = 0
  PV = 0x55d8627d7d50 "use v5.10;\n"\0
  CUR = 11
  LEN = 16
```
Note that lack of `IOK` for line 6? It should look like line 5.

```
  DB<4> use Devel::Peek;Dump( ${"_<a.pl"}[5] );
SV = PVMG(0x55856a4198d0) at 0x55856a4e9630
  REFCNT = 1
  FLAGS = (IOK,POK,pIOK,pPOK)
  IV = 0
  NV = 0
  PV = 0x55856a486840 "\n"\0
  CUR = 1
  LEN = 16
```

This also means that only the lines with pragmas are broken.

This is a deviation from 5.34. In 5.34, we see

```
      DB<3> Dump( ${"_<a.pl"}[6] );
    SV = PVMG(0x55874ddaf680) at 0x55874db508a8
      REFCNT = 1
      FLAGS = (IOK,POK,pIOK,pPOK)
      IV = 94039612799016
      NV = 0
      PV = 0x55874db739a0 "use v5.10;\n"\0
      CUR = 11
      LEN = 16
```

(Same goes at least as far back as 5.10.)

So they were breakable in 5.34, but not in 5.36. Sounds like a valid bug fix. But it sounds like the person that fixed this forgot to set `IOK` when making that change.

tonycoz:
These are meant to be IOK even when non-breakable, and I broke that when fixing #19198.

Fixes #21564

### DIFF
--- a/lib/perl5db.t
+++ b/lib/perl5db.t
@@ -3467,6 +3467,30 @@ EOS
     );
 }
 
+{
+    # https://github.com/Perl/perl5/issues/21564
+    # not a debugger bug, but with the way the fix for #19198 was broken
+    # this needs to be tested with a debugger of some sort (even a no-op
+    # debugger) so test it here.
+    my $wrapper = DebugWrap->new(
+        {
+            cmds =>
+            [
+                'c', # just run it, we check the output of the code
+                'q'
+            ],
+            prog => \<<'EOS',
+use v5.12;
+no strict;
+use B qw(svref_2object SVf_IOK);
+my $sv = svref_2object(\(${"_<$0"}[3])); # the "use B;" line
+say +($sv->FLAGS & SVf_IOK) ? "OK" : "FAIL";
+EOS
+        }
+    );
+    $wrapper->output_like(qr/\bOK\b/, "check the line is IOK");
+}
+
 done_testing();
 
 END {
--- a/op.c
+++ b/op.c
@@ -1361,7 +1361,6 @@ S_cop_free(pTHX_ COP* cop)
         if (av) {
             SV * const * const svp = av_fetch(av, CopLINE(cop), FALSE);
             if (svp && *svp != &PL_sv_undef && SvIVX(*svp) == PTR2IV(cop) ) {
-                (void)SvIOK_off(*svp);
                 SvIV_set(*svp, 0);
             }
         }


tonycoz:
Rebased

lib/perl5db.t
```diff
@@ -3467,6 +3467,30 @@ EOS
...
+say +($sv->FLAGS & SVf_IOK) ? "OK" : "FAIL";
+EOS
+        }
+    );
+    $wrapper->output_like(qr/\bOK\b/, "check the line is IOK");
+}
+
 done_testing();
 
 END {
```
jkeenan, line 3496: From a TDD point-of-view, this looks good, *i.e.,* applying just the test file, I get a FAIL.  Then patching op.c, I get a PASS.

jkeenan (approved):

PR #21567, opened 2023-10-19
Fix compiling without LC_ALL defined

khwilliamson:
It is legal to have a platform without the aggregate LC_ALL locale category defined, though I don't know of any such.  Recent changes broke compilation for that case.  This commit restores it.

(To check this, one has to munge perl.h slightly to force it to be #undef, and I hadn't tried that recently)

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4416,10 +4416,19 @@ Ri	|const char *|mortalized_pv_copy				\
 				|NULLOK const char * const pv
 S	|const char *|native_querylocale_i				\
 				|const locale_category_index cat_index
+S	|void	|new_LC_ALL	|NN const char *lc_all			\
+				|bool force
 S	|void	|output_check_environment_warning			\
 				|NULLOK const char * const language	\
 				|NULLOK const char * const lc_all	\
 				|NULLOK const char * const lang
+S	|parse_LC_ALL_string_return|parse_LC_ALL_string 		\
+				|NN const char *string			\
+				|NN const char **output 		\
+				|const parse_LC_ALL_STRING_action	\
+				|bool always_use_full_array		\
+				|const bool panic_on_error		\
+				|const line_t caller_line
 So	|void	|restore_toggled_locale_i				\
 				|const locale_category_index cat_index	\
 				|NULLOK const char *original_locale	\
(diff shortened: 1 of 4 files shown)


tonycoz:
Is it legal?

From C99, 7.11:
```
The macros defined are NULL (described in 7.17); and

   LC_ALL
...
```
with the same text in the [C23 working draft](https://open-std.org/JTC1/SC22/WG14/www/docs/n3096.pdf) (the 7.17 is 7.21 now). 

From [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/locale.h.html):
```
The <locale.h> header shall define NULL (as described in <stddef.h> and at least the following as macros:

    LC_ALL
...
```

tonycoz:
C89 also requires LC_ALL

khwilliamson:
I was unaware of this.  But bisecting shows that since at least 5.002, there have been `#ifdef LC_ALL` directives in the perl core.  I imagine that I thought those were there for a good reason, and have simply followed existing practice.

Were they there because the authors also thought `LC_ALL` was optional, or because there were non-compliant compilers, or because there were buggy compilers that needed it to be undefined as a workaround?  I don't know.

So what to do now?  I think we should get things to work before considering deleting this code.  Since it has been a feature for a very long time, would there need to be a deprecation cycle?  Note also that since the POSIX 2008 interface doesn't allow for querying `LC_ALL` (even in the proposed update to the Standard), that there are places in the code that effectively have to deal with `LC_ALL` not being available.  And the recent AIX workaround effectively ignores `LC_ALL` because it is broken.

tonycoz:
In 5.002 we (or they, I wasn't involved then) still used K&R function definitions, presumably for pre-C89 compatibility which we no longer support.  I'd expect any build that doesn't support LC_ALL to be broken in several other ways.

I didn't recall the AIX, but I think I found it after some reading through `locale.c`.  Do you mean 28b5b3c676dc7 ?  If it is, I don't think it's relevant to whether LC_ALL is *defined*, it's still defined, we simply need to do some extra work to get a sane "opaque string" from setlocale().

My issue is `locale.c` is very complex in terms of pre-processor conditions, I understand why it's complex, but I think it's worth considering eliminating "never happens" conditions to simplify it if we can.

If nothing else if an implementation is so broken it doesn't define LC_ALL, the user can build with `-DNO_LOCALE` to get a mostly working perl.

I'll approve this, but I do think it's removing as an option.

As to deprecation: we required C89, we require C99, both require that LC_ALL be defined, I don't think we need a deprecation cycle.

tonycoz (approved):

khwilliamson:
The first mention of C89 is in 5.6, contrasting it to C99.  By then there were quite a few #ifdef LC_ALL lines in our code.  Every release has had them. 

I intended only to point out that LC_ALL has had issues even this year, on AIX.  The workaround did not involve undef'ing it.

That's a good point about using `-DNO_LOCALE`, which I hadn't thought of.

PR #21571, opened 2023-10-19
locale.c: Avoid an infinite loop

khwilliamson:
This was doing a 'continue' under some rare Configurations before doing an increment.  No test because it's really hard to get such a Configuration in the wild.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5179,12 +5179,8 @@ S_my_localeconv(pTHX_ const int item)
 #  endif
 
         /* Examine each string */
-        while (1) {
-            const char * name = strings[i]->name;
-
-            if (! name) {   /* Reached the end */
-                break;
-            }
+        for (const lconv_offset_t *strp = strings[i]; strp->name; strp++) {
+            const char * name = strp->name;
 
             /* 'value' will contain the string that may need to be marked as
              * UTF-8 */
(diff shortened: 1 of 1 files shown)


jkeenan:
> This was doing a 'continue' under some rare Configurations before doing an increment. 

Can you cite any examples you've come across?

khwilliamson:
Actually, if one Configures blead with -Accflags=-DNOL_LOCALE_NUMERIC and runs the test suite, version/t/07locale.t with valgrind or asan, one gets invalid reads.

tonycoz (approved):
I see what you're fixing here, I do wonder if it would be more understandable as a for loop:
```
        for (const locale_offset_t *strp = strings[i]; strp->name; ++strp) {
            const char * name = strp->name;

            /* 'value' will contain the string that may need to be marked as
             * UTF-8 */
            SV ** value = hv_fetch(hv, name, strlen(name), true);
            if (! value) {
                continue;
            }

            /* Determine if the string should be marked as UTF-8. */
            if (UTF8NESS_YES == (get_locale_string_utf8ness_i(SvPVX(*value),
                                                  locale_is_utf8,
                                                  NULL,
                                                  (locale_category_index) 0)))
            {
                SvUTF8_on(*value);
            }
        }
```

I think the integers loop immediately below has the same possible bug.

khwilliamson:
I think your method is clearer; so changed to use that.  And yes, the integer loop has the same bug.  New version pushed.

When I used gdb on this, it turns out that the Configuration with NO_LOCALE_NUMERIC is unrelated to the infinite loop, and I have issued another PR to fix that #21573.  I don't remember the Configuration I stumbled upon (or maybe it was a different platform) that caused the loop

tonycoz (approved):

Issue #21402, opened 2023-08-20
`Archive-Tar`: Core/CPAN Version Mismatch 


rwp0:
CPAN's [`Archive-Tar`](https://metacpan.org/dist/Archive-Tar) distribution is at version 3.02 while core [`Archive::Tar`](https://perldoc.perl.org/Archive::Tar) module is version 2.40.

[Version 3.02](https://metacpan.org/release/BINGOS/Archive-Tar-3.02) was released on _April 13, 2023_

[Version 2.40](https://metacpan.org/release/BINGOS/Archive-Tar-2.40) was released on _July 27, 2021_

This means it hasn't been updated for almost [2 years](https://www.timeanddate.com/date/durationresult.html?d1=13&m1=04&y1=2023&d2=27&m2=07&y2=2021).

The [RMG](https://metacpan.org/dist/perl/view/Porting/release_managers_guide.pod) instructs [Dual-Life](https://metacpan.org/dist/perl/view/Porting/release_managers_guide.pod#dual-life-CPAN-module-synchronisation) Core distributions to be synced with CPAN ones before releasing a new Perl version, either development or production.

Here's `core-cpan-diff -x -a` command (_run as per RMG_) output:

```
$ ./perl -Ilib Porting/core-cpan-diff -x -a

Archive::Tar:
    Perl: BINGOS/Archive-Tar-2.40.tar.gz
    CPAN: BINGOS/Archive-Tar-3.02.tar.gz
```

This was noticed by @leonerd while working on his [5.39.2 release](https://github.com/Perl/perl5/blob/b13b7a24d7ed748e05da63140d039ec1ed370441/Porting/release_schedule.pod#L43C36-L43C41).

---

CC @bingos (the distribution maintainer)

iabyn:
On Sun, Aug 20, 2023 at 06:39:50AM -0700, Elvin Aslanov wrote:
> CPAN's [Archive-Tar](https://metacpan.org/dist/Archive-Tar) is at version 3.02 while core [Archive::Tar](https://perldoc.perl.org/Archive::Tar) is version 2.40.
> 
> [Version 3.02](https://metacpan.org/release/BINGOS/Archive-Tar-3.02) was released on April 13, 2023
> 
> [Version 2.40](https://metacpan.org/release/BINGOS/Archive-Tar-2.40) was released on July 27, 2021

There was an attempt to upgrade core to 3.02 in July 2023, but it was
[...]

PR #21575, opened 2023-10-21
cpan/Archive-Tar - Update to version 3.02

haarg:
3.02  13/04/2023 (Manfred Stock)
- Test block sizes up to (2**31 - 1) bytes only (fix for 32bit perls)
- Don't match on message from exception in symlink test
- Improve formatting of $Archive::Tar::RESOLVE_SYMLINK documentation


jkeenan:
Have the problems discussed in https://github.com/Perl/perl5/issues/21402 been addressed?  So far those have discouraged us from synching this version of Archive-Tar into blead.

jkeenan:
> Have the problems discussed in #21402 been addressed? So far those have discouraged us from synching this version of Archive-Tar into blead.

Apparently not.  CI tests on Windows are failing.

rwp0:
This if completed fixes https://github.com/Perl/perl5/issues/21402

May add this wording in the PR description to link those two properly together

haarg:
I'm going to close this for now. Either Archive::Tar needs to be updated, and we'll pull that in, or we'll need to bring Archive::Tar in with a patch. I'm not going to address that right now.

PR #21576, opened 2023-10-21
cpan/Test-Harness - Update to version 3.48

haarg:
3.48 2023-10-02
        - Accept TAP version 14


jkeenan:
> 3.48 2023-10-02 - Accept TAP version 14

+1

(I didn't synch this into blead earlier on the assumption that @Leont would be better positioned to determine when that should happen.)

Leont (approved):

PR #21578, opened 2023-10-21
perldelta for locale changes

khwilliamson:


pod/perldelta.pod
```diff
@@ -415,6 +415,46 @@ Don't clear the integer flag C<IOK> from lines in the C<<
 @{"_<$sourcefile"} >> array when a C<dbstate> op is removed for that
 line.  This was broken when fixing [GH #19198].  [GH #21564]
 
+=item *
+
+Many bug fixes have been made for using locales under threads and in
+embedded perls.  And workarounds for libc bugs have been added.  As a
+result thread-safe locale handling is now the default under OpenBSD, and
+MingW when compiled with UCRT.
```
jkeenan, line 423: It would be good to explain what `UCRT` is.  In the core distribution, it's only mentioned in comments in C-level files.
```
$ ack -l UCRT .
win32/win32.h
locale.c
perl.h
```
Should it get an entry in `pod/perlglossary.pod`?

jkeenan (changes requested):

pod/perldelta.pod
khwilliamson: Changed to spell it out.  I would think that knowing this is a Microsoft term, people would be able to quickly find it if they don't already know what it is.  My guess is that any MS developer would know immediately.

PR #21579, opened 2023-10-22
Fix a bug in VMS configure script and add trailing quotes

jhamby:
Here's a bug fix for the "installation prefix" prompt in the VMS configure script, which currently thinks it's asking a yes/no question and not a generic string. I also added some missing trailing quotes as a separate commit, to fix the syntax highlighting in vim.

My future pull requests will be from individual topic branches with related fixes in each branch, so that I can make separate pull requests for each. I figure these two commits are non-controversial enough that I can put them in a single pull request.


tonycoz:
The changes look sane to me but I'm VMS-clueless, hopefully @craigberry can look over it.

jhamby:
omg, I didn't realize I had an email address from circa 1996 listed as a contributor. I'll run the suggested command to change my email address in the authors list. That should fix the failing check.

jhamby:
I added both my current and past email address to .mailmap since the authors test that failed must be searching the entire Perl history. I'd run the suggested Perl script except it calls "git", and the VMS port of git is so limited that it fails half the time when Perl tries to run it, so I've defined git as the "vgit" command so that Perl doesn't find it at all (for now).

jhamby:
The porting tests cared about the order of the two email addresses of mine that it found, so I reversed them.

craigberry (approved):
Looks good to me.  I wonder when I broke prefix, because that used to work :-).

PR #21580, opened 2023-10-22
make a new stub to clone into when pushing a new pad

### LINKED ISSUE #18606: Can't undef active subroutine
Should this produce: "deep recursion" not: "Can't undef active subroutine" ?

```
#!/usr/bin/perl
sub a
 {my sub b
   {&a; # Can't undef active subroutine
   }
  b
 }
a
```

tonycoz:
This previously put the same CV into the inner pad, so on a recursive call into the owning sub, from this sub, this CV would still be active, and the attempt to clone into the still busy CV would throw.

Fixes #18606

### DIFF
--- a/pad.c
+++ b/pad.c
@@ -2468,8 +2468,19 @@ Perl_pad_push(pTHX_ PADLIST *padlist, int depth)
                         || PadnameIsSTATE(names[ix])
                         || sigil == '&')
                 {
-                    /* outer lexical or anon code */
-                    sv = SvREFCNT_inc(oldpad[ix]);
+                    SV *tmp = oldpad[ix];
+                    if (sigil == '&' && SvTYPE(tmp) == SVt_PVCV
+                        && !PadnameOUTER(names[ix])
+                        && CvLEXICAL(tmp) && CvCLONED(tmp)
+                        && !PadnameIsOUR(names[ix])
+                        && !PadnameIsSTATE(names[ix])) {
+                        /* lexical sub that needs cloning */
+                        sv = newSV_type(SVt_PVCV);
+                    }
+                    else {
+                        /* outer non-cloning lexical or anon code */
+                        sv = SvREFCNT_inc(tmp);
+                    }
                 }
                 else {		/* our own lexical */
                     if (sigil == '@')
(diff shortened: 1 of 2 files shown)


tonycoz:
push to fix minitest

pad.c
```diff
@@ -2460,8 +2460,19 @@ Perl_pad_push(pTHX_ PADLIST *padlist, int depth)
...
+                        && !PadnameOUTER(names[ix])
+                        && CvLEXICAL(tmp) && CvCLONED(tmp)
+                        && !PadnameIsOUR(names[ix])
+                        && !PadnameIsSTATE(names[ix])) {
+                        /* lexical sub that needs cloning */
+                        sv = newSV_type(SVt_PVCV);
+                    }
+                    else {
+                        /* outer non-cloning lexical or anon code */
+                        sv = SvREFCNT_inc(oldpad[ix]);
```
mauke:
```c
sv = SvREFCNT_inc(tmp);
```
?

richardleach:
I've been added as a reviewer, but don't understand the operation of lexical subs well enough to meaningfully contribute. At my level of understanding, it is apparent that different instances of the same lexical sub trying to operate on the same pad will cause problems and this PR adds some separation.

pad.c
tonycoz: Thanks, fixed.

mauke:
I don't fully understand the context of this change, but:

- It is a very small change.
- The change in behavior only affects lexical subs (`my sub`).
- The gist of it makes sense to me (if lexical sub, then clone, don't reuse/reassign).
- The current behavior of lexical subs is broken (#18606, #21987).
- This change fixes those bugs (and it comes with tests for the correct behavior).

Therefore I think it should be merged.

leonerd (approved):
LGTM

jimav:
What is the first perl release which includes this fix? 
(or point me to how to answer such questions in general: What release has a particular commit?)

I'm still getting "Can't undef active subroutine" with 5.36.0

xenu:
```
> git tag --contains 4fc2379
v5.39.10
v5.39.8
v5.39.9
```

This is also visible in GitHub's interface under the commit message.

mauke:
> I'm still getting "Can't undef active subroutine" with 5.36.0

5.36.0 was released on 2022-05-27 (according to `perldoc perlhist`).
This PR was created on 2023-10-22 and merged on 2024-02-20.
Barring the invention of time travel, it is very unlikely for 5.36.0 to include this fix.

PR #21581, opened 2023-10-22
remove C++20 deprecated volatile argument

tonycoz:
This produced warnings like:
```
g++ -c   -D_REENTRANT -D_GNU_SOURCE -DDEBIAN -fwrapv -fno-strict-aliasing \
  -pipe -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 \
  -xc++ -I/usr/lib/llvm-18/include -std=c++17 -fno-exceptions \
  -funwind-tables -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS \
  -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  -fexceptions \
  -I/usr/lib/x86_64-linux-gnu/perl5/5.36/Imager/include -std=c++23 \
  -O2 -g   -DVERSION=\"1.000\" -DXS_VERSION=\"1.000\" -fPIC \
  "-I/usr/lib/x86_64-linux-gnu/perl/5.36/CORE"   LLVM.c
In file included from /usr/lib/x86_64-linux-gnu/perl/5.36/CORE/perl.h:5798,
                 from LLVM.xs:15:
/usr/lib/x86_64-linux-gnu/perl/5.36/CORE/proto.h:392:65: warning: ‘volatile’-qualified parameter is deprecated [-Wvolatile]
  392 | PERL_CALLCONV I32       Perl_call_sv(pTHX_ SV* sv, volatile I32 flags);
      |                                                    ~~~~~~~~~~~~~^~~~~
```
I haven't been able to produce this when building perl itself as C++ since it requires a modern -std argument to reproduce, and that breaks the build in other ways.

### DIFF
--- a/perl.c
+++ b/perl.c
@@ -3060,7 +3060,7 @@ See L<perlcall>.
 */
 
 SSize_t
-Perl_call_sv(pTHX_ SV *sv, volatile I32 flags)
+Perl_call_sv(pTHX_ SV *sv, I32 arg_flags)
                         /* See G_* flags in cop.h */
 {
     LOGOP myop;		/* fake syntax tree node */
@@ -3070,6 +3070,10 @@ Perl_call_sv(pTHX_ SV *sv, volatile I32 flags)
     bool oldcatch = CATCH_GET;
     int ret;
     OP* const oldop = PL_op;
+    /* Since we don't modify flags after setjmp() we don't really need to make
+       flags volatile, but gcc complains that it could be clobbered anyway.
+     */
+    volatile I32 flags = arg_flags;
     dJMPENV;
 
     PERL_ARGS_ASSERT_CALL_SV;
--- a/embed.fnc
+++ b/embed.fnc
@@ -772,7 +772,7 @@ CTadop	|Malloc_t|calloc	|MEM_SIZE elements			\
 AOdp	|SSize_t|call_pv	|NN const char *sub_name		\
 				|I32 flags
 AOdp	|SSize_t|call_sv	|NN SV *sv				\
-				|volatile I32 flags
+				|I32 flags
 : Used in several source files
 Rp	|bool	|cando		|Mode_t mode				\
 				|bool effective 			\
--- a/proto.h
+++ b/proto.h
@@ -444,7 +444,7 @@ Perl_call_pv(pTHX_ const char *sub_name, I32 flags);
         assert(sub_name)
 
 PERL_CALLCONV SSize_t
-Perl_call_sv(pTHX_ SV *sv, volatile I32 flags);
+Perl_call_sv(pTHX_ SV *sv, I32 flags);
 #define PERL_ARGS_ASSERT_CALL_SV                \
         assert(sv)
 


jkeenan:
> This produced warnings like:
> 
> ```
> g++ -c   -D_REENTRANT -D_GNU_SOURCE -DDEBIAN -fwrapv -fno-strict-aliasing \
>   -pipe -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 \
>   -xc++ -I/usr/lib/llvm-18/include -std=c++17 -fno-exceptions \
>   -funwind-tables -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS \
>   -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  -fexceptions \
>   -I/usr/lib/x86_64-linux-gnu/perl5/5.36/Imager/include -std=c++23 \
>   -O2 -g   -DVERSION=\"1.000\" -DXS_VERSION=\"1.000\" -fPIC \
>   "-I/usr/lib/x86_64-linux-gnu/perl/5.36/CORE"   LLVM.c
> In file included from /usr/lib/x86_64-linux-gnu/perl/5.36/CORE/perl.h:5798,
>                  from LLVM.xs:15:
> /usr/lib/x86_64-linux-gnu/perl/5.36/CORE/proto.h:392:65: warning: ‘volatile’-qualified parameter is deprecated [-Wvolatile]
>   392 | PERL_CALLCONV I32       Perl_call_sv(pTHX_ SV* sv, volatile I32 flags);
>       |                                                    ~~~~~~~~~~~~~^~~~~
> ```
> 
> I haven't been able to produce this when building perl itself as C++ since it requires a modern -std argument to reproduce, and that breaks the build in other ways.

@tonycoz, I don't understand your final paragraph above.  You said, "I haven't been able to produce this" -- *this* presumably referring to the warnings output -- but *something* produced that output.  Can you clarify?  Provide `uname -mrs` and `perl -V:config_args` perhaps?

tonycoz:
You can't produce it building perl itself, or at least I wasn't able to, due to the way our build process works (or doesn't.)

You can produce it building a C++ XS module, which is what the output is from (though unfinished.)  Since it's a C++20 deprecation you probably need to request C++ 20 or later from the compiler, which the `-std=c++23` option does in that g++ command line.

mauke (approved):

Leont:
Why is that variable even volatile? It doesn't make sense to me.

mauke:
Good question. The `volatile` was introduced in commit 001d637e39d61293255c8b6c91da77ebdca4e4d6 ("FW: [PATCH] RE: [PATCH] volatile, avoid clobbered"), which doesn't tell us much. Presumably the intent was to make sure that `flags` is preserved across a `longjmp`, but local variables that are not modified between `setjmp` and `longjmp` (like `flags`) should be safe anyway.

AnFunctionArray:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1152r0.html#C89

Basically:

The C89 Committee also considered requiring that a call to longjmp restore the calling environment fully, that is, that upon execution of longjmp....

part - the last [...].

So yeah this commit is the correct solution as suggested in:

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1152r0.html#parret

But also the more you learn.

mauke:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf

7.13.2.1/3:

> All accessible objects have values, and all other components of the abstract machine have state, as of the time the longjmp function was called, except that the values of objects of automatic storage duration that are local to the function containing the invocation of the corresponding setjmp macro that do not have volatile-qualified type *and have been changed between the setjmp invocation and longjmp call* are indeterminate.

`flags` is not modified after `setjmp`, so does not need to be `volatile`.

tonycoz:
Unfortunately removing the volatile entirely produces:
```
perl.c: In function ‘Perl_call_sv’:
perl.c:3063:32: warning: argument ‘flags’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Wclobbered]
 3063 | Perl_call_sv(pTHX_ SV *sv, I32 flags)
      |                            ~~~~^~~~~

```

Given this I expect the 001d637 patch was to suppress this warning rather than to meet the requirements of the language/library.

So I think the PR is good as it is.

<details><summary>Patch against this PR I used to check for the warning</summary>
<p>


```
diff --git a/perl.c b/perl.c
index 32c1b7945e..4784cc3c1e 100644
--- a/perl.c
+++ b/perl.c
@@ -3060,7 +3060,7 @@ See L<perlcall>.
 */
 
 SSize_t
-Perl_call_sv(pTHX_ SV *sv, I32 arg_flags)
+Perl_call_sv(pTHX_ SV *sv, I32 flags)
                         /* See G_* flags in cop.h */
 {
     LOGOP myop;                /* fake syntax tree node */
@@ -3071,7 +3071,6 @@ Perl_call_sv(pTHX_ SV *sv, I32 arg_flags)
     int ret;
     OP* const oldop = PL_op;
     /* keep a volatile copy since C++ deprecated volatile function arguments */
-    volatile I32 flags = arg_flags;
     dJMPENV;
 
     PERL_ARGS_ASSERT_CALL_SV;
```

</p>
</details>

tonycoz:
Though I'll edit the comment

perl.c
```diff
@@ -3070,6 +3070,10 @@ Perl_call_sv(pTHX_ SV *sv, volatile I32 flags)
     bool oldcatch = CATCH_GET;
     int ret;
     OP* const oldop = PL_op;
+    /* Since we don't modify flags after setjmp() we don't really to make
```
mauke: "we don't really to" - missing "need"?

tonycoz: Thanks, fixed.

PR #21582, opened 2023-10-23
use the supplied handle in IO::Handle::blocking() on Win32

### LINKED ISSUE #17455: IO::Handle::blocking does not work under windows 7 64bit [rt.cpan.org #61074]
Migrated from [rt.cpan.org#61074](https://rt.cpan.org/Ticket/Display.html?id=61074) (status was 'new')

Requestors:
* deathwolf+bitcard@gmail.com

Attachments:
* [issue.pl](https://rt.cpan.org/Ticket/Attachment/827112/428517/issue.pl)

From deathwolf+bitcard@gmail.com on 2010-09-05 02:12:17:
```
IO::Handle::blocking seems to break the socket entity under windows 7
64bit(possible other OS'es).

Attached is an example of breaking code: please adjust the port and
address to something that is open on your side, and then try
uncommenting either of the IO::Handle::blocking calls. The connect will
fail with "unknown error" or "A non-blocking socket operation could not
be completed immediately." depending on the perl distribution.

Please note that $sock->blocking(0) or $sock->blocking(1) will also fail
the same way.

Issue present in both activeperl 5.10.1 and strawberry perl 5.12.1.
```

### From deathwolf+bitcard@gmail.com on 2010-09-05 11:39:54:
Additional bug reference for a similar bug that might be related #17448

tonycoz:
The refactor in 375ed12a meant that fd was initialized to -1,
rather than PerlIO_fileno(f) on Win32.

fixes #17455

### DIFF
--- a/dist/IO/IO.xs
+++ b/dist/IO/IO.xs
@@ -86,12 +86,12 @@ io_blocking(pTHX_ InputStream f, int block)
     if (RETVAL >= 0) {
 	int mode = RETVAL;
 	int newmode = mode;
-#ifdef O_NONBLOCK
+#  ifdef O_NONBLOCK
 	/* POSIX style */
 
-# ifndef O_NDELAY
-#  define O_NDELAY O_NONBLOCK
-# endif
+#    ifndef O_NDELAY
+#      define O_NDELAY O_NONBLOCK
+#    endif
 	/* Note: UNICOS and UNICOS/mk a F_GETFL returns an O_NDELAY
 	 * after a successful F_SETFL of an O_NONBLOCK. */
 	RETVAL = RETVAL & (O_NONBLOCK | O_NDELAY) ? 0 : 1;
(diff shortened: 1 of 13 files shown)


dist/IO/IO.xs
```diff
@@ -129,8 +129,9 @@ io_blocking(pTHX_ InputStream f, int block)
     if (block >= 0) {
 	unsigned long flags = !block;
 	/* ioctl claims to take char* but really needs a u_long sized buffer */
-	const int ret = ioctl(fd, FIONBIO, (char*)&flags);
-	if (ret != 0)
+
+        RETVAL = ioctl(fd, FIONBIO, (char*)&flags);
```
mauke: Why use `RETVAL` here? The value isn't actually used/returned in the `WIN32` branch of the code. (In fact, we could skip the variable altogether and write `if (ioctl(...) != 0) { return -1; }`.)

If this is just to avoid "unused variable" warnings on Win32, IMHO the better fix would be to move the `int RETVAL` definition into the `HAS_FCNTL` branch.

We could save one level of (preprocessor) indentation/nesting by changing `... #else #ifdef WIN32 ... #else ... #endif #endif` to `... #elif defined WIN32 ... #else ... #endif`. It would even be symmetrical with the original condition, which is `#if defined HAS_FCNTL` (not `#ifdef`).

tonycoz: I don't remember why I used RETVAL, I've cleaned that up.

I've also rearranged the #if tree and cleaned up indentation

mauke (approved):

PR #21584, opened 2023-10-25
sv.c: Initialize intrpvar.h variable in cloning

khwilliamson:
This was previously overlooked.  But it only affects rare Configuration, so hadn't shown up so far in our testing.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -16117,6 +16117,13 @@ perl_clone_using(PerlInterpreter *proto_perl, UV flags,
     PL_less_dicey_locale_buf = NULL;
     PL_less_dicey_locale_bufsize = 0;
 #endif
+#if ! defined(HAS_NL_LANGINFO)          \
+ &&   defined(USE_LOCALE_CTYPE)         \
+ && ! defined(WIN32)                    \
+ && ! defined(HAS_MBTOWC)               \
+ && ! defined(HAS_MBRTOWC)
+    PL_langinfo_recursed = 0;
+#endif
 
     /* Unicode inversion lists */
 


tonycoz:
Is it a valid configuration? C99 requires mbrtowc().

khwilliamson:
It is not a valid configuration, but was in C89 when this code was mostly written.  And experience has shown that it is best to have fallbacks where possible with respect to vendor libc locale handling functions, because they appear to be buggier than other portions of their code base.  I long have known, for example, that querylocale() was buggy in Darwin, but only recently was it confirmed that the whole POSIX 2008 locale interface there was buggy.  Also recently, we discovered bugs in the most basic C89 locale function, setlocale(), in AIX.  I had to skip tests in 5.38 on that platform, but was able to write a workaround in 5.39.  We also have workarounds in place for glibc, and for *BSD libc bugs, and FreeBSD has only recently fixed some bugs I reported years ago; some may remain.

We still check for mbr?towc() in Configure, along with lots of other C99 functions.  I suspect these are required to avoid breaking XS code that may test for their existence.

I am unaware of problems with the mbr?towc functions, but again I wouldn't have imagined there being problems with basic setlocale().  If this code didn't already exist, I would think we would just say we don't support this Configuration should the situation this covers arise, but since it does exist, I think it's worth it, for now, to keep it from bit rotting.  After the locale handling is stable for a while, or when the bus factor drops to 0, is when I would think it would be ok to rip this out.  It occupies neither time nor space since it's #ifdef'd.

tonycoz (approved):

PR #21585, opened 2023-10-25
locale.c: Note another Configuration option

khwilliamson:
This is about another Configuration option that affects locale handling.

I'm thinking we should make this option the default for the remainder of the 5.39 development cycle, to shake out any issues.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -252,6 +252,15 @@
  *          querylocale() is called only while the locale mutex is locked, and
  *          the result is copied to a per-thread place before unlocking.
  *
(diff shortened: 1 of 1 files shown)


tonycoz (approved):

khwilliamson:
Any comments about making this the default for now?

tonycoz:
> Any comments about making this the default for now?

I think it's fine if guarded by a version check, so it only applies for 5.39.now - 5.39.9 or thereabouts, just in case we forget to remove it or bus factor becomes a problem.

PR #21594, opened 2023-10-26
[doc] `Benchmark.pm`: Scope variables

rwp0:
This enhances the module's documentation.

### DIFF
--- a/lib/Benchmark.pm
+++ b/lib/Benchmark.pm
@@ -5,13 +5,13 @@ use strict;
 
 =head1 NAME
 
-Benchmark - benchmark running times of Perl code
+Benchmark - Benchmark running times of Perl code
 
 =head1 SYNOPSIS
 
-    use Benchmark qw(:all) ;
+    use Benchmark qw(:all);
 
-    timethis ($count, "code");
+    timethis($count, "code");
 
     # Use Perl code in strings...
     timethese($count, {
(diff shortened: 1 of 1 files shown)


rwp0:
Can this be merged now that I fixed the conflicts?

lib/Benchmark.pm
```diff
@@ -494,7 +494,7 @@ init();
 
 sub BEGIN {
     if (eval 'require Time::HiRes') {
-	Time::HiRes->import('time');
+	import Time::HiRes qw(time);
```
haarg: This is indirect object syntax, which we are trying to move away from.

rwp0: @haarg I fixed that

lib/Benchmark.pm
```diff
@@ -494,7 +494,7 @@ init();
 
 sub BEGIN {
     if (eval 'require Time::HiRes') {
-	Time::HiRes->import('time');
+	 Time::HiRes->import('time');
```
haarg: This now has an extraneous space.

rwp0: I see, I fixed that as well, though the inconsistent indentation is on the file-level.

rwp0:
@haarg can we merge already if all good now since hasConflicts is removed

rwp0:
Any updates yet?

mauke (approved):

book (approved):

PR #21595, opened 2023-10-26
release announcement: perl is the language, perl 5 is the version

haarg:
Update the release announcement template to refer to the language as Perl, and use the full version rather than omitting the 5.

Also update the metacpan link to match its new URLs.


leonerd (approved):
LGTM

jkeenan (approved):

PR #21599, opened 2023-10-27
Revert "locale.c: Don't use SvPVX .." and replace

khwilliamson:
Commit 2f1617ef716ae12369475fa6c11cbefd7ba7dfca works but under Configurations compiled with -DNO_LOCALE_NUMERIC or -DNO_LOCALE_MONETARY, "uninitialized" errors are emitted.

Revert that commit, and instead test for SvPOK before getting to the spot where the value would be used.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5194,12 +5194,12 @@ S_my_localeconv(pTHX_ const int item)
             /* 'value' will contain the string that may need to be marked as
              * UTF-8 */
             SV ** value = hv_fetch(hv, name, strlen(name), true);
-            if (! value) {
+            if (! value || ! SvPOK(*value)) {
                 continue;
             }
 
             /* Determine if the string should be marked as UTF-8. */
-            if (UTF8NESS_YES == (get_locale_string_utf8ness_i(SvPV_nolen(*value),
+            if (UTF8NESS_YES == (get_locale_string_utf8ness_i(SvPVX(*value),
                                                   locale_is_utf8,
                                                   NULL,
                                                   (locale_category_index) 0)))


jkeenan:
> Commit [2f1617e](https://github.com/Perl/perl5/commit/2f1617ef716ae12369475fa6c11cbefd7ba7dfca) works but under Configurations compiled with -DNO_LOCALE_NUMERIC or -DNO_LOCALE_MONETARY, "uninitialized" errors are emitted.
> 
> Revert that commit, and instead test for SvPOK before getting to the spot where the value would be used.

Just now I configured perl at e97277e136147cec7536b0a1d1a97208add9d0eb with the following arguments:
```
sh ./Configure -des -Dusedevel -DNO_LOCALE_NUMERIC
```
I then ran `make test_prep` and recorded the output.  (`gcc-11` is the compiler here on Linux.)  I then ran the program I use to extract build-time warnings from that output, but got none.
```
$ report-build-warnings e97277e136.linux.unthreaded.gcc-11.no-locale-numeric.maketp.output.txt.gz 
File:  e97277e136.linux.unthreaded.gcc-11.no-locale-numeric.maketp.output.txt.gz

```
How or where are you getting those uninitialized value warnings?

khwilliamson:
On 10/27/23 15:27, James E Keenan wrote:
>     Commit 2f1617e
>     <https://github.com/Perl/perl5/commit/2f1617ef716ae12369475fa6c11cbefd7ba7dfca> works but under Configurations compiled with -DNO_LOCALE_NUMERIC or -DNO_LOCALE_MONETARY, "uninitialized" errors are emitted.
> 
>     Revert that commit, and instead test for SvPOK before getting to the
>     spot where the value would be used.
> 
> Just now I configured perl at e97277e 
> <https://github.com/Perl/perl5/commit/e97277e136147cec7536b0a1d1a97208add9d0eb> with the following arguments:
> 
> |sh ./Configure -des -Dusedevel -DNO_LOCALE_NUMERIC |
> 
> I then ran |make test_prep| and recorded the output. (|gcc-11| is the 
> compiler here on Linux.) I then ran the program I use to extract 
> build-time warnings from that output, but got none.
> 
> |$ report-build-warnings 
> e97277e136.linux.unthreaded.gcc-11.no-locale-numeric.maketp.output.txt.gz File: e97277e136.linux.unthreaded.gcc-11.no-locale-numeric.maketp.output.txt.gz |
> 
> How or where are you getting those uninitialized value warnings?

In looking back on it, it appears that they are coming from a test file 
not yet in blead, https://github.com/Perl/perl5/pull/21513

tonycoz (approved):

PR #21601, opened 2023-10-28
Announce 5.42 removal of goto to jump into construct

### LINKED ISSUE #16275: inward goto deprecation has no removal date
Migrated from [rt.perl.org#132492](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=132492) (status was 'open')

Searchable as RT132492$

jkeenan:
Fixes #16275

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -3626,7 +3626,8 @@ Also, unlike most named operators, this has the same precedence as
 assignment.
 
 Use of C<goto LABEL> or C<goto EXPR> to jump into a construct is
-deprecated and will issue a warning.  Even then, it may not be used to
+deprecated and will issue a warning; it will become a fatal error in
+Perl 5.42. While still available, it may not be used to
 go into any construct that requires initialization, such as a
 subroutine, a C<foreach> loop, or a C<given>
 block.  In general, it may not be used to jump into the parameter
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -7664,7 +7664,7 @@ repeatedly, the C</g> has no effect.
 =item Use of "goto" to jump into a construct is deprecated
 
 (D deprecated::goto_construct) Using C<goto> to jump from an outer scope into an inner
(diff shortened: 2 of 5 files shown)


pod/perlfunc.pod
```diff
@@ -3633,8 +3633,8 @@ block.  In general, it may not be used to jump into the parameter
 of a binary or list operator, but it may be used to jump into the
 I<first> parameter of a binary operator.  (The C<=>
 assignment operator's "first" operand is its right-hand
-operand.)  It also can't be used to go into a
-construct that is optimized away.
+operand.)  It also can't be used to go into a construct that is optimized
+away.  Finally, it will be made into a fatal exception in Perl 5.42.
```
hvds: I think this would be better following on from the first sentence of the paragraph:

```
Use of C<goto LABEL> or C<goto EXPR> to jump into a construct is
deprecated and will issue a warning; it will become a fatal error in
Perl 5.42. While still available, it may not be used to go into any construct [...]
```

hvds (approved):
I have not checked if there are other docs that need updating.

pod/perlfunc.pod
jkeenan: Agreed.  Please check that I've got your wording correct.

jkeenan:
> I have not checked if there are other docs that need updating.

What would be helpful is if you could review the documentation of `goto` in `pod/perlfunc.pod` (e.g., https://github.com/Perl/perl5/blob/blead/pod/perlfunc.pod#goto-LABEL) to see if there's anything else that would be affected by this change.  I myself don't think there is, but then again I've never written `goto` in Perl code.

rwp0 (approved):

hvds:
> What would be helpful is if you could review the documentation of `goto` in `pod/perlfunc.pod` to see if there's anything else that would be affected by this change.

I don't see anything else there that would be affected.

jkeenan:
> Fixes #16275

@leonerd can you check to see if this p.r. is satisfactory?  Thanks.

Issue #16815, opened 2019-01-19
Carp.pm useless warning "Use of uninitialized value $error[1] in join or string"


p5pRT:
Migrated from [rt.perl.org#133776](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133776) (status was 'open')

Searchable as RT133776$

p5pRT:
### From wagnerc@plebeian.com

Message RFC822:
From: wagnerc@plebeian.com
Reply-To: wagnerc@plebeian.com
Subject: Carp.pm useless warning "Use of uninitialized value $error[1] in
 join or string"
To: perlbug@perl.org
Message-ID: <5.22.4_13508_1547759937@applejack>
CC: wagnerc@plebeian.com
X-RT-Original-Encoding: ascii
Content-Type: text/plain; charset="ascii"
Content-Length: 5281


This is a bug report for perl from wagnerc@plebeian.com,
generated with the help of perlbug 1.40 running under perl 5.22.4.


-----------------------------------------------------------------
The Carp 1.50 main functions will print a useless and confusing warning about the use of an uninitialized value if one of the arguments to the given function is undef.  This breaks the encapsulation principle of Carp to "report the caller's errors, not the ones it might happen to tickle while doing so."

Examples:
$ perl -MCarp -e 'Carp::carp "foo", undef, "bar";'
Use of uninitialized value $error[1] in join or string at /usr/lib/perl5/site_perl/5.22/Carp.pm line 588.
foobar at -e line 1.

$ perl -MCarp -e 'sub{Carp::confess "foo", undef, "bar"}->();'
Use of uninitialized value $error[1] in join or string at /usr/lib/perl5/site_perl/5.22/Carp.pm line 588.
foobar at -e line 1.
        main::__ANON__() called at -e line 1

$ perl -MCarp -e 'package X; sub{Carp::croak "foo", undef, "bar"}->();'
Use of uninitialized value $error[1] in join or string at /usr/lib/perl5/site_perl/5.22/Carp.pm line 588.
foobar at -e line 1.
        X::__ANON__() called at -e line 1

$ perl -MCarp -e 'package X; sub{Carp::cluck "foo", undef, "bar"}->();'
Use of uninitialized value $error[1] in join or string at /usr/lib/perl5/site_perl/5.22/Carp.pm line 588.
foobar at -e line 1.
        X::__ANON__() called at -e line 1


Fix:
[...]

PR #21603, opened 2023-10-29
Carp croak gh16827 20231029

jkeenan:
Rebases an older branch on blead.  Failing tests; do not merge; for research only.

See: https://github.com/Perl/perl5/issues/16815


jkeenan:
In https://github.com/Perl/perl5/commit/97e737c99f9f2eb32f9911a554db7ab97c09a7b9 I added code changes suggested by the OP back in 2019 when we were still using RT for issue tracking (https://github.com/Perl/perl5/issues/16815).  Test failures resulted.  I was only looking at the problem from the point of view of cleaning up loose ends from our RT->GH conversion.  I had, and have, no strong feelings about the problems in Carp, so I'm not going to invest more time in this pull request.  Closing.

PR #21607, opened 2023-10-31
Mention that the regex used by the empty m// is scoped

choroba:


pod/perlop.pod
```diff
@@ -2055,15 +2055,15 @@ The bottom line is that using C</o> is almost never a good idea.
...
-I<successfully> matched regular expression is used instead. In this
-case, only the C<g> and C<c> flags on the empty pattern are honored; the
-other flags are taken from the original pattern. If no match has
-previously succeeded, this will (silently) act instead as a genuine
-empty pattern (which will always match).  Using a user supplied string as
-a pattern has the risk that if the string is empty that it triggers the
-"last successful match" behavior, which can be very confusing. In such
-cases you are recommended to replace C<m/$pattern/> with
-C<m/(?:$pattern)/> to avoid this behavior.
+I<successfully> matched regular expression in the current scope is
```
mauke: "current dynamic scope", surely? All the match variables are dynamically scoped.

choroba: Force pushed the change.

demerphq:
FWIW perlvar already contains the following text:

> Notice that each successful match in the exact same scope overrides the
match context of the previous successful match, but that unsuccessful
matches do not. Also note that in an inner nested scope the previous
state from an outer dynamic scope persists until it has been overriden
by another successful match, but that when the inner nested scope exits
whatever match context was in effect before the inner successful match
is restored when the scope concludes.
>
> It is a known issue that `goto LABEL` may interact poorly with the
dynamically scoped match context. This may not be fixable, and is
considered to be one of many good reasons to avoid `goto LABEL`.

pod/perlop.pod
```diff
@@ -2055,15 +2055,16 @@ The bottom line is that using C</o> is almost never a good idea.
...
-case, only the C<g> and C<c> flags on the empty pattern are honored; the
-other flags are taken from the original pattern. If no match has
-previously succeeded, this will (silently) act instead as a genuine
-empty pattern (which will always match).  Using a user supplied string as
-a pattern has the risk that if the string is empty that it triggers the
-"last successful match" behavior, which can be very confusing. In such
-cases you are recommended to replace C<m/$pattern/> with
-C<m/(?:$pattern)/> to avoid this behavior.
+I<successfully> matched regular expression in the current dynamic
+scope is used instead. In this case, only the C<g> and C<c> flags on
```
demerphq: IMO this should include a reference to the documentation in perlvar.

demerphq (changes requested):

pod/perlop.pod
choroba: Added a fixup. Let me know if it's OK to squash it.

demerphq: lgtm. once you squash it I will approve the merge.

choroba: Done.

PR #21612, opened 2023-11-03
Add ability to emulate thread-safe locale operations

khwilliamson:
Locale information was originally global for an entire process.  Later,
it was realized that different threads could want to be running in
different locales.  Windows added this ability, and POSIX 2008 followed
suit (though using a completely different API).  When available, perl
automatically uses these capabilities.

But many platforms have neither, or their implementation, such as on
Darwin, is buggy.  This commit adds the capability for Perl programs to
operate as if the platform were thread-safe.

This implementation is based on the observation that the underlying
locale matters only to relatively few libc calls, and only during their
execution.  It can be anything at all at any other time.  perl keeps
what the proper locale should be for each category in a a per-thread
array.  Each locale-dependent operation must be wrapped in mutex
lock/unlock operations.  The lock additionally compares what libc knows
the locale to be, and what it should be for this thread at this time,
and changes the actual locale to the proper value if necessary.  That's
all that is needed.

This commit adds macros to perl.h, for example "MBTOWC_LOCK_", that
expand to do the mutex lock, and change the global locale to the
expected value.  On perls built without this emulation capability, they
are no-ops.  All code in the perl core (unless I've missed something),
are changed to use these macros (there weren't actually many places that
needed this).  Thus, any pure perl program will automatically become
locale-thread-safe under this Configuration.

In order for XS code to also become locale-thread-safe, it must use
these macros to wrap calls to locale-dependent functions.  Relatively
few modules call such functions.  For example, the only one I found that
ships with the perl core is Time::Piece, and it has more fundamental
issues with running under threads than this.  I am preparing pull
requests for it.

Thus, this is not completely transparent to code like native-thread-safe
locale handling is.  Therefore ${^SAFE_LOCALES} returns 2 (instead of 1)
for this type of thread-safety.

Another deficiency compared to the native thread safety is when a thread
calls a non-perl library that accesses the locale.  The typical example is
Gtk (though this particular application can be configured to not be
problematic).  With the native safe threads, everything works as long as
only one such thread is used per Perl program.  That thread would then
be the only one operating in the global locale, hence there are no
conflicts.  With this emulation, all threads are operating in the global
locale, and mutexes would have to be used to prevent conflicts.  To
minimize those, the code added in this commit restores the global locale
when through to the state it was in when started.

A major concern is the performance impact.  This is after all trading
speed for accuracy.  lib/locale_threads.t is noticeably slower when this
is being used.  But that is doing multiple threads constantly using
locale-dependent operations.  I don't notice any change with the rest of
the test suite.  In pure perl, this only comes into play while in the
scope of 'use locale' or when using some of the few POSIX:: functions
that are locale-dependent.  And to some extent when formatting, but the
regular overhead there should dwarf what this adds.

This commit leaves this feature off by default.  The next commit changes
that for the next few 5.39 development releases, so we can see if there
is actually an issue.


khwilliamson:
The MingW failure is because `makedef.pl` doesn't know about this being compiled with `UCRT`. Does anyone know a fix for that?

sisyphus:
> The MingW failure is because makedef.pl doesn't know about this being compiled with UCRT. Does anyone know a fix for that?

I don't know - but does setting <code>LIBC</code> (in the win32/GNUmakefile) to 'libucrt.a' address this problem ?

For the MSVC builds, the GNUmakefile does set <code>LIBC</code> to either msvcrt.lib or ucrt.lib, as appropriate - and <code>$Config{libc}</code> is set to the same value.
However, for the mingw builds, both <code>LIBC</code> and <code>$Config{libc}</code> remain empty.

I've checked, and the mingw-w64 ports of gcc do contain <code>libucrt.a</code>, along with <code>libucrtapp.a</code> and <code>libucrtbase.a</code>

Cheers,
Rob

sisyphus:
If I build this source leaving <code>LIBC</code> unset then I get the same build failure as reported by the CI test.
If I build with <code>LIBC=-lucrt</code> then I get past that failure, but then quickly get clobbered with:
```
gmake[1]: Leaving directory 'D:/comp-1320-ucrt/perl-5.39.42/dist/threads-shared'
g++ -shared -o ..\perl539.dll  -s -L"D:\perl-5.39.42-1320-ucrt-mcf\lib\MSWin32-x64-multi-thread\CORE" -L"C:\winlibs-ucrt-1320\mingw64\lib" -L"C:\winlibs-ucrt-1320\mingw64\x86_64-w64-mingw32\lib" -L"C:\winlibs-ucrt-1320\mingw64\lib\gcc\x86_64-w64-mingw32\13.2.0" \
   ..\toke.o ..\regcomp.o ..\regcomp_trie.o ..\regcomp_debug.o ..\regcomp_invlist.o ..\regcomp_study.o ..\regexec.o ..\op.o ..\sv.o ..\pp.o ..\pp_ctl.o ..\pp_sys.o ..\pp_pack.o ..\pp_hot.o ..\gv.o ..\perl.o ..\utf8.o ..\dump.o ..\hv.o ..\av.o ..\builtin.o ..\caretx.o ..\class.o ..\deb.o ..\doio.o ..\doop.o ..\dquote.o ..\globals.o ..\mro_core.o ..\locale.o ..\keywords.o ..\mathoms.o ..\mg.o ..\numeric.o ..\pad.o ..\peep.o ..\perly.o ..\pp_sort.o ..\reentr.o ..\run.o ..\scope.o ..\taint.o ..\time64.o ..\universal.o ..\util.o perllib.o ..\perlio.o .\win32.o .\win32sck.o .\win32thread.o .\fcrypt.o ..\DynaLoader.o ..\lib\auto\Win32CORE\Win32CORE.a  -lmoldname -lkernel32 -luser32 -lgdi32 -lwinspool -lcomdlg32 -ladvapi32 -lshell32 -lole32 -loleaut32 -lnetapi32 -luuid -lws2_32 -lmpr -lwinmm -lversion -lodbc32 -lodbccp32 -lcomctl32 ..\lib\CORE\perl539.exp
C:/winlibs-ucrt-1320/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ..\lib\CORE\perl539.exp:fake:(.edata+0x504): undefined reference to `Perl_category_lock_i'
C:/winlibs-ucrt-1320/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ..\lib\CORE\perl539.exp:fake:(.edata+0x508): undefined reference to `Perl_category_unlock_i'
make: *** [GNUmakefile:1500: ..\perl539.dll] Error 1
```
I'll look into this further.

Cheers,
Rob

khwilliamson:
I don't know how much you know about makedef.pl, so here is some info in case you don't know.

MingW should work fine when loaded with UCRT and not need the functions added in this commit.  

But, if it doesn't have UCRT, it does need these functions.  They must be listed in makedef.pl in order for them to be loaded.

Windows prior to its use of UCRT also needs these functions; and doesn't afterwards.  We already have determined that MSVC80 and above have this.  Hence the code in makedef.pl
````
if ($define{USE_LOCALE_THREADS} && ! $define{NO_THREAD_SAFE_LOCALE}) {
    if (    $define{USE_POSIX_2008_LOCALE}
        || ($define{WIN32} && (   $cctype !~ /\D/
                               && $cctype >= 80)))
    {
        $define{USE_THREAD_SAFE_LOCALE} = 1;
    }
    else {
        $define{EMULATE_THREAD_SAFE_LOCALES} = 1;
    }
}
````
and then further down
````
unless ($define{EMULATE_THREAD_SAFE_LOCALES})
{
    ++$skip{$_} foreach qw(
        PL_restore_locale
        PL_restore_locale_depth
        Perl_category_lock_i
        Perl_category_unlock_i
    );
}
````
The goal would be to change the first section of code to also account for MingW.

**Comment by khwilliamson** [comment]
I added a commit that just assumes that UCRT is being used for MingW, and it now works.  And I looked at every value that  is currently read or passed in to makedef.pl.  Nothing appeared to me to be relevant.  That would mean that GNUMakefile would need to be modified somehow to pass this information when it calls makedef.pl

**Comment by sisyphus** [comment]
It looks to me that it's the LIBC setting in GNUmakefile that should be set to <code>-lucrt</code> when (and only when) gcc uses UCRT,
<code>makedef.pl</code> can then detect that setting by examining <code>$Config{libc}</code> which will be set to whatever GNUmakefile's <code>LIBC</code> is set to.

We can set LIBC to <code>-lucrt</code> by specifying <code>LIBC=-lucrt</code> as one of gmake's command line arguments - without any need to amend the GNUmakefile.
However, it should probably be ammended to allow manual alteration, along with some explanatory documentation.

For the moment, I'm just specifying <code>LIBC=-lucrt</code> as a command line arg.
And I've applied the following patch to the latest <code>makedef.pl</code>:
```
--- makedef.pl_orig	2023-11-08 09:47:32.578126900 +1100
+++ makedef.pl	2023-11-08 11:22:27.539482300 +1100
@@ -167,7 +167,7 @@
 
 if ($define{USE_LOCALE_THREADS} && ! $define{NO_THREAD_SAFE_LOCALE}) {
     if (    $define{USE_POSIX_2008_LOCALE}
-        || ($define{WIN32} && ($cctype eq "GCC" || (   $cctype !~ /\D/
+        || ($define{WIN32} && (($cctype eq "GCC" && $Config{libc} eq '-lucrt') || (   $cctype !~ /\D/
                                                     && $cctype >= 80))))
     {
         $define{USE_THREAD_SAFE_LOCALE} = 1;
```
With the gcc-13.2.0 UCRT compiler, that builds fine and the perl test suite passes.
As regards <code>lib/locale_threads.t</code> specifically:
```
>..\perl -I..\lib ../lib/locale_threads.t
1..2
ok 1 - Didn't segfault
ok 2 - Verify there were no failures with simultaneous running threads
```
With a gcc-13.2.0 MSVCRT compiler, $Confgi{libc} will be the empty string (as has always previously been the case) and the condition changes to FALSE.
I gather that's what we want to happen ?
Do I need to run additional testing ?

Is there a better way of going about it ?

UPDATE: I've just tested using a gcc-13.2.0 MSVCRT compiler - $Config{libc} is set to the empty string. Test 2 of t/lib/locale_threads.t failed <strike>, whereas I had expected it to be skipped</strike> (silly expectation, me now thinks). All other tests passed.

Cheers,
Rob

**Comment by khwilliamson** [comment]
Earlier this year, when I was testing this code, there were two platforms that had issues with it.  This, and AIX (early versions, the only ones we get smokes for).  My suspicion is it is a bug in each platform.  On MingW, it succeeded sometimes.  My plan is then to just not do the emulation on MingW, since using UCRT makes it natively thread-safe, among other fixes.

So I think we should do something like @sisyphus suggests, to add a line (plus comments) to GNUMakefile to set LIBC to`-lucrt` by default, and indicate to comment that out if you don't want it.

@tonycoz @xenu @steve-m-hay  Any other suggestions?

**Comment by tonycoz** [comment]
> Any other suggestions?

Before @xenu pointed out the `_UCRT` macro, I parsed the output of `gcc --version` in `GNUmakefile` to detect which CRT was being used:
```
C:\Users\Tony\dev\perl\git\perl\win32>c:\strawberry-perl-5.38.0.1-64bit-portable\c\bin\gcc --version
gcc (MinGW-W64 x86_64-msvcrt-posix-seh, built by Brecht Sanders) 13.1.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.


C:\Users\Tony\dev\perl\git\perl\win32>c:\mingw64-ucrt\bin\gcc --version
gcc (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders) 13.1.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

This should be easier to do in `makedef.pl` that is was in `GNUmakefile`.

**Comment by xenu** [comment]
`gcc -dumpmachine` is a bit easier to parse than `gcc --version`

**Comment by xenu** [comment]
Scratch that, I was sure it contains the same information, but apparently not.

```
> gcc  -dumpmachine
x86_64-w64-mingw32
> gcc --version
gcc.exe (MinGW-W64 x86_64-ucrt-mcf-seh, built by Brecht Sanders) 13.1.0
```

**Comment by sisyphus** [comment]
I don't have any UCRT builds from vendors other than Brecht Sanders, so I can't quickly check that they also specify "ucrt" somewhere inside <code>gcc --version</code>,

Setting LIBC inside GNUmakefile has the added advantage that the standard <code>perl -V</code> output then automatically displays that <code>libc=-lucrt</code> (for UCRT builds only) - and that information might be handy in bug reports at some stage.

The GNUmakefile currently prepends LIBC to LIBFILES, and LIBC therefore also appears at the beginning of <code>$Config{perllibs}</code> and <code>$Config{libs}</code>.
That's why I had to specify <code>-lucrt</code>.
However, I don't see any need for LIBC to be inserted at the start of LIBFILES. If we didn't do that we could set LIBC to some other string, if it's so desired.

I've just completed a build with LIBC set to '-lucrt', but <b>not</b> prepended to <code>$Config{perllibs}</code> and <code>$Config{libs}</code>.
All tests passed.

Cheers,
Rob

**Comment by sisyphus** [comment]
> I don't have any UCRT builds from vendors other than Brecht Sanders, so I can't quickly check that they also specify "ucrt" somewhere inside <code>gcc --version</code>

The ones provided by niXman apparently don't.
I've grabbed https://github.com/niXman/mingw-builds-binaries/releases/download/13.2.0-rt_v11-rev0/x86_64-13.2.0-release-posix-seh-ucrt-rt_v11-rev0.7z.
Going by the filename, I expect it's UCRT.
```
>gcc --version
gcc (x86_64-posix-seh-rev0, Built by MinGW-Builds project) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Another download option is https://github.com/niXman/mingw-builds-binaries/releases/download/13.2.0-rt_v11-rev0/x86_64-13.2.0-release-posix-seh-msvcrt-rt_v11-rev0.7z.
I expect that's the MSVCRT alternative.
It reproduces the same <code>gcc --version</code> info:
```
>gcc --version
gcc (x86_64-posix-seh-rev0, Built by MinGW-Builds project) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
I've sent an email off to mingw-w64 mailing list, seeking enlightenment.
I'll report back when the enlightenment has been delivered.

Cheers,
Rob

**Comment by sisyphus** [comment]
LHMouse says:
<quote> 
```
echo '#include <stddef.h>' | x86_64-w64-mingw32-gcc -E -dM -xc - | fgrep _UCRT
```
gives <code>#define _UCRT</code> if the headers have been configured to use UCRT by default.
</quote>

Not sure where that leaves us ... especially those of us still using the cmd.exe shell.

**Comment by sisyphus** [comment]
The following  achieves what I would like to see - without any need to alter the GNUmakefile and without any need for user input.
First, I created a new directory in the win32 folder named "configure", and placed a file name "rt.c" inside that new directory.
This new file contains:
```
/*
 Print the default runtime. $Config{libc}
 will be set to this specified value.
*/
#include <stdio.h>
#include <stddef.h>

int main(void) {
#if defined(_UCRT)
  printf("-lucrt\n");
#else
  printf("-lmsvcrt\n");
#endif
  return 0;
}
```
I then patched win32/config_sh.PL as follows:
```
     }
     elsif ($opt{cc} =~ /\bgcc\b/) {
         chomp($opt{gccversion} = `$opt{cc} -dumpversion`);
+        chomp($opt{libc} = `$opt{cc} -o rt.exe configure/rt.c 1>nul 2>&1 && rt`);
+        `del /F /Q rt.exe`; # rt.exe no longer needed
     }
 }
```
I then built current bleadperl in the usual way (without specifying a value for <code>LIBC</code>).
All test pass - including test 2 of lib/locale_threads.t which was actually run (ie not skipped).
<code>$Config{libc} </code> was set to "-lucrt", but $Config{perllibs} and $Config{libs} do not contain "-lucrt".

I guess we could instead build and run configure/rt.exe from inside <code>makedef.pl</code>.
However, we would then end up with a <code>perl -V</code> that tells us nothing about the default runtime.

khwilliamson:
@xenu, @tonycoz  any comments on @sisyphus proposal

tonycoz:
I think it's reasonable.

The test is reasonable and putting it in `config_sh.PL` means that ext/cpan/dist can use it if needed.

PR #21621, opened 2023-11-07
Fix up locale.c when no mbtowc() nor mbrtowc()

khwilliamson:
This code was written when we still required only a C89 compiler for non-Windows systems without nl_langinfo().  Now that we require C99, mbtowc() and mbrtowc() should be available, and this code could be jettisoned.  But I'd like to get it in good shape before I forget the nuances, so that if it needs to be brought back (say because of a non-conforming compiler or buggy implementation), that it easily can.

I therefore re-examined the code and this series of commits is the result.  Notably the lines affected by its inclusion or not are reduced to just a single area of `locale.c`, to make it easier to revert the removal.  And I figured out a few ways to improve it and save some cycles.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5131,6 +5131,21 @@ S_my_localeconv(pTHX_ const int item)
      * corrections determined at hash population time, at an extra maintenance
      * cost which khw doesn't think is worth it
      */
+
+#  ifndef HAS_SOME_LANGINFO
+
+    /* We are done when called with an individual item.  There are no integer
+     * items to adjust, and it's best for the caller to determine if this
+     * string item is UTF-8 or not.  This is because the locale's UTF-8ness is
+     * calculated below, and in some Configurations, that can lead to a
+     * recursive call to here, which could recurse infinitely. */
+
+    if (item != 0) {
+        return hv;
+    }
+
+#  endif
+
     for (unsigned int i = 0; i < 2; i++) {  /* Try both types of strings */
         if (! strings[i]) {     /* Skip if no strings of this type */
             continue;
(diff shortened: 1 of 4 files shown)


locale.c
```diff
@@ -6300,28 +6300,37 @@ S_my_langinfo_i(pTHX_
         utf8ness_t is_utf8 = UTF8NESS_UNKNOWN;
         char * scratch_buf = NULL;
         Size_t scratch_buf_size = 0;
+        const U8 * first_variant;
+        Size_t len;
```
tonycoz: I think these belong within the loops that use them.

There's no harm in having separate declarations of `len` within each loop, and it makes it easier to reason about the use of `len`, since it would not available outside those inner blocks.

Similarly for `first_variant`.

I don't see any uses of `len` or `first_variant` outside of their respective loops, or in the rest of the commit series, but I may have missed them.

I believe this would also allow `len` to be const.

locale.c
```diff
@@ -6283,21 +6283,100 @@ S_my_langinfo_i(pTHX_
...
+         * could think of is to see if the codeset is part of the locale name.
+         * This is very less than ideal; often there is no code set in the
+         * name; and at other times they even lie.
+         *
+         * But there is an XPG standard syntax, which many locales follow:
+         *
+         * language[_territory[.codeset]][@modifier]
+         *
+         * So we take the part between the dot and any '@' */
+        retval = (const char *) strchr(locale, '.');
```
tonycoz: Why cast here?

In C strchr() returns `char *` which trivially converts to `const char *`.

In C++, since `locale` is `const char *` so the `const char *` [overload from](https://en.cppreference.com/w/cpp/string/byte/strchr)\*:
```
const char* strchr( const char* str, int ch ); 
char* strchr( char* str, int ch );
```
is selected here, which already returns `const char *`.

There's no cast for the similar assignment to `modifier` below.

\* `#include <string.h>` in C++ behaves like including `<cstring.h>` but adds the names that would normally be added to the `std::` namespace to the [global namespace](https://en.cppreference.com/w/cpp/header#C_compatibility_headers)

khwilliamson: It was old code that I simply moved from elsewhere in the file without changing it.  I imagine there were different types at one time that required the case.  But now changed.

locale.c
khwilliamson: It's because not both are loops, so the 2nd  does conflict with the first.

But I'm not happy with the duplicated logic; it was ok before these commits, as it was just a line or two.  But it has expanded.  I've now pushed a revised version which has one loop that iterates over all of the tested strings.  It makes a portion more complicated, but the actual logic is squashed into one instance

locale.c
```diff
@@ -6294,129 +6255,307 @@ S_my_langinfo_i(pTHX_
...
+        locale_category_index  cat_index = LC_TIME_INDEX_;
+        const locale_category_index  follow_on_cat_index = LC_TIME_INDEX_;
+
+#          else
+
+        /* Effectively out-of-bounds, as here there are no trial entries at
+         * all.  This allows this code to compile, but there are no strings to
+         * test, and so the answer will always be non-UTF-8. */
+        const locale_category_index  cat_index =
+                                     follow_on_cat_index = LC_ALL_INDEX_;
```
tonycoz: What declares `follow_on_cat_index` in this case?

Also, `cat_index` is const here, the loop below may never actually execute but it still tries to modify cat_index.

I suspect this should have been:
```
        locale_category_index  cat_index = LC_ALL_INDEX_;
        const locale_category_index  follow_on_cat_index = LC_ALL_INDEX_;
```

tonycoz (approved):

locale.c
khwilliamson: Yes, good catch.  I hadn't bothered to actually build this configuration; but I did with your fixes, and it passed its tests

khwilliamson:
Applied via a15d3a085745f320ab9b75bdf56b647a9f9914f9

PR #21624, opened 2023-11-09
win32: retire visual studio 2013 support

tonycoz:
I recently (tried) to test VS 2013 builds, but while vs2013 in theory supports mixed declarations and code, in practice that support is broken, and failed to build toke.obj:

The one case I looked at in detail, VS 2013 required an extra ; after a conditional statement to accept the following declaration, for example:

```
C:\Users\Tony\dev\perl\git>cl /c mixed.c
Microsoft (R) C/C++ Optimizing Compiler Version 18.00.21005.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

mixed.c
mixed.c(8) : error C2143: syntax error : missing ';' before 'const'
mixed.c(9) : error C2065: 'pend' : undeclared identifier
mixed.c(9) : error C2100: illegal indirection

C:\Users\Tony\dev\perl\git>type mixed.c
/* adapted from S_intuit_more() in toke.c.
   don't run this code
*/
int g(const char *p) {
  if (!p)
    return (0);

  const char * const pend = p+1;
  return *pend;
}
```

regcomp.obj also failed to build due to the use of the "inline" keyword in regcomp.h.  This is fairly trivially fixable but inline is also C99.

Despite these build issues we haven't received any error reports for VS2013, so I don't think it's worth re-working any mixed declarations to support it.

It might be worth updating regcomp.h to use PERL_STATIC_INLINE but that's out of scope for this commit.

### DIFF
--- a/README.win32
+++ b/README.win32
@@ -41,7 +41,7 @@ additional software to run (other than what came with your operating
 system).  Currently, this port is capable of using one of the
 following compilers on the Intel x86 and x86_64 architectures:
 
-      Microsoft Visual C++    version 12.0 or later
+      Microsoft Visual C++    Visual C++ 2015 (version 14.0) or later
       Intel C++ Compiler      (experimental)
       Gcc by mingw.org        gcc version 3.4.5-5.3.0
       Gcc by mingw-w64.org    gcc version 4.4.3 or later
@@ -65,14 +65,8 @@ that are also supported by perl's makefile.
 =back
 
 The Microsoft Visual C++ compilers are also now being given away free. They
-are available as "Visual C++ 2013-2022 Community Edition" and are the same
-compilers that ship with "Visual C++ 2013-2022 Professional".
-
-Visual C++ 2013 is capable of B<targeting> XP and Windows Server 2003 but the
-build host requirement is Windows 7/Windows Server 2012. For more details see
-https://docs.microsoft.com/en-us/visualstudio/productinfo/vs2013-compatibility-vs
-and
-https://docs.microsoft.com/en-us/visualstudio/productinfo/vs2013-sysrequirements-vs
+are available as "Visual C++ 2015-2022 Community Edition" and are the same
+compilers that ship with "Visual C++ 2015-2022 Professional".
 
 The MinGW64 compiler is available at L<https://mingw-w64.org>.
 The latter is actually a cross-compiler targeting Win64. There's also a trimmed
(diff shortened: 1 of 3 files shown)


README.win32
```diff
@@ -41,7 +41,8 @@ additional software to run (other than what came with your operating
 system).  Currently, this port is capable of using one of the
 following compilers on the Intel x86 and x86_64 architectures:
 
-      Microsoft Visual C++    version 12.0 or later
+      Microsoft Visual C++    version 14.0 or later
+                              Visual C++ 2015 or later
```
xenu: Perhaps it would be better to phrase it as "2015 (14.0) or later". The current wording kind of implies those are two different products.

xenu (approved):

xenu:
Yeah, if the build is broken and nobody noticed, it's a no-brainer. VS2013 was the odd one out of the supported versions, the only one without UCRT.

rwp0 (approved):

khwilliamson:
What then is the earliest supported version after this is merged?

sisyphus:
> What then is the earliest supported version after this is merged?

Earliest supported will then be version 14.0, also referred to as "MSVC140", and "Visual C++ 2015" in the GNUmakefile and Makefile.

README.win32
tonycoz: I've adjusted that and removed a 2013 mention from `README.win32`

khwilliamson:
I believe we can also remove anything that special cases any _MSC_VER < 1900

mauke:
Not a lot of those around:
```
cpan/Compress-Raw-Zlib/zlib-src/zutil.c:#if (!defined(_MSC_VER) || (_MSC_VER <= 600))
cpan/Compress-Raw-Zlib/zlib-src/zutil.h:#if (defined(_MSC_VER) && (_MSC_VER > 600)) && !defined __INTERIX
cpan/Time-Piece/Piece.xs:#if !(_MSC_VER >= 1500)
dist/Time-HiRes/HiRes.xs:#  if((defined(_MSC_VER) && _MSC_VER < 1900) || \
perl.h:#      elif _MSC_VER < 1900
```

jkeenan:
On 11/14/23 21:02, mauke wrote:
> Not a lot of those around:
> 
> |cpan/Compress-Raw-Zlib/zlib-src/zutil.c:#if (!defined(_MSC_VER) || 
> (_MSC_VER <= 600)) cpan/Compress-Raw-Zlib/zlib-src/zutil.h:#if 
> (defined(_MSC_VER) && (_MSC_VER > 600)) && !defined __INTERIX 
> cpan/Time-Piece/Piece.xs:#if !(_MSC_VER >= 1500)

The two files above will require coordination with their upstream 
maintainers.

> dist/Time-HiRes/HiRes.xs:# if((defined(_MSC_VER) && _MSC_VER < 1900) || 
> \ perl.h:# elif _MSC_VER < 1900 |
> 

This can be handled within blead.

sisyphus:
>  dist/Time-HiRes/HiRes.xs:# if((defined(_MSC_VER) && _MSC_VER < 1900) ||
 \ perl.h:# elif _MSC_VER < 1900 |

We can remove more than just that from HiRes.xs.
That quoted line is part of a section that pertains solely to unsupported Visual Studio and unsupported mingw.org compilers.
AFAICS,, the following piece from HiRes.xs can (and should) be removed in its entirety:

```
/* Visual C++ 2013 and older don't have the timespec structure.
 * Neither do mingw.org compilers with MinGW runtimes older than 3.22. */
#  if((defined(_MSC_VER) && _MSC_VER < 1900) || \
      (defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR) && \
      defined(__MINGW32_MAJOR_VERSION) && (__MINGW32_MAJOR_VERSION < 3 || \
      (__MINGW32_MAJOR_VERSION == 3 && __MINGW32_MINOR_VERSION < 22))))
struct timespec {
    time_t tv_sec;
    long   tv_nsec;
};
#  endif
```

khwilliamson:
Also, now that MingW can be compiled with UCRT, things that are under WIN32 and `#ifndef _MSC_VER` should be checked to see if `_UCRT` is defined, and then we can assume a modern run time library, but I don't know how modern

PR #21628, opened 2023-11-11
`Deparse.t`: Fix `new` indirect syntax

rwp0:
In two places

### DIFF
--- a/lib/B/Deparse.t
+++ b/lib/B/Deparse.t
@@ -62,7 +62,7 @@ while (<DATA>) {
     # parse options if necessary
     my $deparse = $meta{options}
 	? $deparse{$meta{options}} ||=
-	    new B::Deparse split /,/, $meta{options}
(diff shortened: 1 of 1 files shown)


jkeenan:
I recommend that you **not** use the term "Revert" in the first line of your commit message, as that incorrectly suggests that we are using `git revert` to undo a commit to the repository which we now believe to be a mistake.  The code you're replacing is **not** a mistake, merely discouraged.

rwp0:
> I recommend that you **not** use the term "Revert" in the first line of your commit message, as that incorrectly suggests that we are using `git revert` to undo a commit to the repository which we now believe to be a mistake. The code you're replacing is **not** a mistake, merely discouraged.

Okay, will amend that.
Currently yes, but why to use something ourselves that we discourage? 🙂

rwp0:
Can this be merged please since we have many similar PR's merged into blead?

PR #21629, opened 2023-11-11
cpan/Getopt-Long - Update to version 2.57

jkeenan:
Changes in version 2.57
-----------------------

* Adapt tests for perl core, thanks to James Keenan. Packaging changes only, no functional changes.

Changes in version 2.56
-----------------------

* Do not use Getopt::Long to establish Getopt::Long::Parser version. Packaging changes only, no functional changes.

Changes in version 2.55
-----------------------

* Move Getopt::Long::Parser to a separate module and add documentation. Note that Getopt::Long uses Getopt::Long::Parser, so programs that use Getopt::Long and call the oo interface will still work.

* Allow periods in option names, e.g. to indicate levels in a hierarchical configuration. Since periods in option names have never been valid, they are now allowed by default unless pass_through is enabled.

* Fix long standing bug that duplicate options were not detected when the options differ in case while ignore_case is in effect. This will now yield a warning and become a fatal error in a future release.

Committer: Reflect changes in makefiles due to Getopt-Long reorganization.

In the course of synching Getopt-Long-2.57 into blead there were changes to MANIFEST et al. which led to regeneration of metadata.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -1150,12 +1150,15 @@ cpan/Filter-Util-Call/t/call.t			See if Filter::Util::Call works
 cpan/Filter-Util-Call/t/rt_101033.pm
 cpan/Filter-Util-Call/t/rt_101033.t
 cpan/Filter-Util-Call/t/rt_54452-rebless.t
-cpan/Getopt-Long/lib/Getopt/Long.pm	Fetch command options (GetOptions)
-cpan/Getopt-Long/t/gol-basic.t		See if Getopt::Long works
-cpan/Getopt-Long/t/gol-linkage.t	See if Getopt::Long works
-cpan/Getopt-Long/t/gol-oo.t		See if Getopt::Long works
-cpan/Getopt-Long/t/gol-xargv.t		See if Getopt::Long works
-cpan/Getopt-Long/t/gol-xstring.t	See if Getopt::Long works
+cpan/Getopt-Long/lib/Getopt/Long.pm		Fetch command options (GetOptions)
+cpan/Getopt-Long/lib/Getopt/Long/Parser.pm	Getopt-Long
+cpan/Getopt-Long/t/gol-basic.t			See if Getopt::Long works
+cpan/Getopt-Long/t/gol-linkage.t		See if Getopt::Long works
+cpan/Getopt-Long/t/gol-load1.t			Getopt-Long
+cpan/Getopt-Long/t/gol-load2.t			Getopt-Long
+cpan/Getopt-Long/t/gol-oo.t			See if Getopt::Long works
+cpan/Getopt-Long/t/gol-xargv.t			See if Getopt::Long works
+cpan/Getopt-Long/t/gol-xstring.t		See if Getopt::Long works
 cpan/HTTP-Tiny/corpus/auth-01.txt	Test file related to HTTP::Tiny
 cpan/HTTP-Tiny/corpus/auth-02.txt	Test file related to HTTP::Tiny
 cpan/HTTP-Tiny/corpus/auth-03.txt	Test file related to HTTP::Tiny
--- a/Makefile.SH
+++ b/Makefile.SH
@@ -1441,8 +1441,8 @@ _cleaner2:
 	-rmdir lib/IO/Compress/Zlib lib/IO/Compress/Zip lib/IO/Compress/Gzip
 	-rmdir lib/IO/Compress/Base lib/IO/Compress/Adapter lib/IO/Compress
 	-rmdir lib/IO lib/I18N/LangTags lib/I18N lib/Hash/Util lib/Hash
-	-rmdir lib/HTTP lib/Filter/Util lib/Filter lib/File/Spec
-	-rmdir lib/ExtUtils/Typemaps lib/ExtUtils/ParseXS
+	-rmdir lib/HTTP lib/Getopt/Long lib/Filter/Util lib/Filter
+	-rmdir lib/File/Spec lib/ExtUtils/Typemaps lib/ExtUtils/ParseXS
 	-rmdir lib/ExtUtils/MakeMaker/version lib/ExtUtils/MakeMaker
 	-rmdir lib/ExtUtils/Liblist lib/ExtUtils/Constant lib/ExtUtils/Command
 	-rmdir lib/ExtUtils/CBuilder/Platform/Windows
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -612,7 +612,8 @@ our %Modules = (
     },
 
     'Getopt::Long' => {
-        'DISTRIBUTION' => 'JV/Getopt-Long-2.54.tar.gz',
+        'DISTRIBUTION' => 'JV/Getopt-Long-2.57.tar.gz',
+        'SYNCINFO'     => 'jkeenan on Sat Nov 11 13:09:21 2023',
         'FILES'        => q[cpan/Getopt-Long],
         'EXCLUDED'     => [
             qr{^examples/},
--- a/cpan/Getopt-Long/lib/Getopt/Long.pm
+++ b/cpan/Getopt-Long/lib/Getopt/Long.pm
@@ -3,31 +3,24 @@
 # Getopt::Long.pm -- Universal options parsing
 # Author          : Johan Vromans
 # Created On      : Tue Sep 11 15:00:12 1990
-# Last Modified By: Johan Vromans
-# Last Modified On: Thu Nov 17 17:45:27 2022
-# Update Count    : 1777
+# Last Modified On: Sat Nov 11 17:48:41 2023
+# Update Count    : 1808
 # Status          : Released
 
 ################ Module Preamble ################
 
-# There are no CPAN testers for very old versions of Perl.
-# Getopt::Long is reported to run under 5.8.
-use 5.004;
+# Getopt::Long is reported to run under 5.6.1. Thanks Tux!
+use 5.006001;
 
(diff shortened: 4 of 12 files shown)


jkeenan:
This synchronization with CPAN was more complex than usual, so I'm pushing it to a smoke-me branch.  I'll merge it into blead in two or three days so that it's ready for a November 20 monthly development release.

jkeenan:
This pull request is failing in two of or GH continuous integration tests:  those for Windows Mingw64 pull request and push:

* push request: https://github.com/Perl/perl5/actions/runs/6835969949/job/18590440823?pr=21629
* pull: https://github.com/Perl/perl5/actions/runs/6835967877/job/18609019538?pr=21629

I've rerun one of the two runs with no improvement.  I believe this is a Perl build-time error on these machines -- not an error in Getopt-Long-2.57.  The failing code looks like this:
```
2023-11-13T02:04:34.4411123Z ..\miniperl.exe -I..\lib ..\make_ext.pl "MAKE=gmake" --dir=..\cpan --dir=..\dist --dir=..\ext lib
2023-11-13T02:04:34.4424505Z Generating a gmake-style Makefile
2023-11-13T02:04:34.4805196Z Writing Makefile for lib
2023-11-13T02:04:34.4809737Z gmake[1]: Entering directory 'D:/a/perl5/perl5/dist/lib'
2023-11-13T02:04:34.5832401Z ..\miniperl.exe -I..\lib ..\mkppport
2023-11-13T02:04:34.6011152Z Can't locate base.pm in @INC (you may need to install the base module) (@INC entries checked: D:\a\perl5\perl5\lib D:\a\perl5\perl5\cpan\AutoLoader\lib D:\a\perl5\perl5\dist\Carp\lib D:\a\perl5\perl5\dist\PathTools D:\a\perl5\perl5\dist\PathTools\lib D:\a\perl5\perl5\cpan\ExtUtils-Install\lib D:\a\perl5\perl5\cpan\ExtUtils-MakeMaker\lib D:\a\perl5\perl5\cpan\ExtUtils-Manifest\lib D:\a\perl5\perl5\cpan\File-Path\lib D:\a\perl5\perl5\ext\re D:\a\perl5\perl5\dist\Term-ReadLine\lib D:\a\perl5\perl5\dist\Exporter\lib D:\a\perl5\perl5\ext\File-Find\lib D:\a\perl5\perl5\cpan\Text-Tabs\lib D:\a\perl5\perl5\dist\constant\lib D:\a\perl5\perl5\cpan\version\lib D:\a\perl5\perl5\cpan\Getopt-Long\lib D:\a\perl5\perl5\cpan\Text-ParseWords\lib D:\a\perl5\perl5\cpan\ExtUtils-PL2Bat\lib D:\a\perl5\perl5\dist\ExtUtils-ParseXS\lib D:\a\perl5\perl5\cpan\parent\lib D:\a\perl5\perl5\cpan\ExtUtils-Constant\lib .) at D:\a\perl5\perl5\cpan\Getopt-Long\lib/Getopt/Long.pm line 23.
2023-11-13T02:04:34.6022680Z BEGIN failed--compilation aborted at D:\a\perl5\perl5\cpan\Getopt-Long\lib/Getopt/Long.pm line 23.
2023-11-13T02:04:34.6027850Z Compilation failed in require at ..\mkppport line 6.
2023-11-13T02:04:34.6030220Z BEGIN failed--compilation aborted at ..\mkppport line 6.
2023-11-13T02:04:34.6043558Z gmake: *** [GNUMakefile:1544: MakePPPort] Error 2
2023-11-13T02:04:34.6044270Z gmake: *** Waiting for unfinished jobs....
2023-11-13T02:04:34.8117353Z "D:\a\perl5\perl5\miniperl.exe" "-I..\..\lib" lib_pm.PL lib.pm 
2023-11-13T02:04:34.8486546Z Extracting lib.pm (with variable substitutions)
2023-11-13T02:04:35.0128788Z gmake[1]: Leaving directory 'D:/a/perl5/perl5/dist/lib'
```
I should note that this version of Getopt-Long is (a) doing fine on CPANtesters; (b) doing fine on our smoke-testing rigs; and (c) doing fine in all other runs on our GH CI rig, **including two Windows MSVC 142 rigs.**  The upstream author (@sciurius) is the long-time maintainer of this distro.

The MingW failures at a simple (new) invocation of `use base qw(Exporter);`.  I don't see why that should cause a Perl build-time failure.  The mixture of forwards- and backwards-path separators in the CI output is suspicious.

I don't know much about our CI rigs and have never personally used MingW.  Could some people who are knowledgeable about either please take a look?

Thank you very much.

jkeenan:
@tonycoz, @xenu, @atoomic, could you take a look at this CI problem?  Thanks.

tonycoz:
Getopt::Long uses `base`, which hasn't been built at the point the test fails, and `dist/base/lib` isn't in the `@toolchain` array in `write_buildcustomize.pl` so isn't available even for bootstrapping.

I expect the simplest correct fix is to add `"dist/base/lib"` to `@toolchain` along with the [other MSWin32 conditional additions.](https://github.com/Perl/perl5/blob/6be4da51a41be558ff16958665830c24f90ec2cc/write_buildcustomize.pl#L52)

Changing Getopt::Long to use `parent` instead of the discouraged `base` would also fix this, since `cpan/parent/lib` is added to `@INC` by `write_buildcustomize.pl` on Windows.  Getopt::Long would need to depend on parent for this to work on CPAN.

Even better would be to import 'import' from Exporter instead of using inheritance:
```
use Exporter qw(import);
```
though it would need to depend on Exporter version 5.57 or later for this to work on CPAN.

jkeenan:
> Getopt::Long uses `base`, which hasn't been built at the point the test fails, and `dist/base/lib` isn't in the `@toolchain` array in `write_buildcustomize.pl` so isn't available even for bootstrapping.
> 
> I expect the simplest correct fix is to add `"dist/base/lib"` to `@toolchain` along with the [other MSWin32 conditional additions.](https://github.com/Perl/perl5/blob/6be4da51a41be558ff16958665830c24f90ec2cc/write_buildcustomize.pl#L52)
> 

Let me try this first before I ask the maintainer to revise his code.  Thanks.

jkeenan:
> > Getopt::Long uses `base`, which hasn't been built at the point the test fails, and `dist/base/lib` isn't in the `@toolchain` array in `write_buildcustomize.pl` so isn't available even for bootstrapping.
> > I expect the simplest correct fix is to add `"dist/base/lib"` to `@toolchain` along with the [other MSWin32 conditional additions.](https://github.com/Perl/perl5/blob/6be4da51a41be558ff16958665830c24f90ec2cc/write_buildcustomize.pl#L52)
> 
> Let me try this first before I ask the maintainer to revise his code. Thanks.

Done.  Let's see how this does on our Windows smokers as well as in our CI.  Then I'll squash and merge.

sciurius:
I have no problems reverting the 'use base' change in Getopt::Long, just let me know which of the alternatives is preferred. However,

> I expect the simplest correct fix is to add `"dist/base/lib"` to `@toolchain` along with the [other MSWin32 conditional additions.](https://github.com/Perl/perl5/blob/6be4da51a41be558ff16958665830c24f90ec2cc/write_buildcustomize.pl#L52)

If I understand correctly, this indicates there is a difference in build environment between MSWindows and other platforms. That sounds like a bad idea.

jkeenan:
> I have no problems reverting the 'use base' change in Getopt::Long, just let me know which of the alternatives is preferred. However,
>

See below.
 
> > I expect the simplest correct fix is to add `"dist/base/lib"` to `@toolchain` along with the [other MSWin32 conditional additions.](https://github.com/Perl/perl5/blob/6be4da51a41be558ff16958665830c24f90ec2cc/write_buildcustomize.pl#L52)
> 
> If I understand correctly, this indicates there is a difference in build environment between MSWindows and other platforms. That sounds like a bad idea.

Perhaps so, but that opens up a much wider scope of problems than is appropriate for this ticket.  If we want to go there, someone should open an issue (not a pull request, yet) on GH or raise this on the mailing list.

The change suggested by @tonycoz has enabled us to get PASS on all our GH CI rigs, including the two Mingw64 rigs.  So right now I've squashed the two commits in the p.r. and force-pushed.  When CI completes once again, I'll merged to blead.  At this point, the more extensive smoke-testing blead gets will have a better chance of flushing out any problems with Getopt-Long-2.57.

@sciurius, the only thing that I would suggest to you right now is to simply open an Issue in Getopt-Long's GH repo asking for an investigation of the forward and backwards compatibility of `use base` versus `use parent` versus any other invocation.  Then if you recommend a revision you can include it in your next CPAN release and P5P can resume the discussion then.

tonycoz:
> If I understand correctly, this indicates there is a difference in build environment between MSWindows and other platforms. That sounds like a bad idea.

Windows is in general its own special snowflake, and there is a difference in the build process here.

On Windows the makefile is static, so it's difficult to parallelize the build within the sets of XS and non-XS extensions for the gmake makefile.  To provide *some* parallel building the XS and non-XS extensions are built in parallel with each other.  In this case it means that `mkppport` may try to run before `base.pm` has been built (and copied to `lib/`) and so `mkppport` fails, as seen in the original version of this PR.

On POSIX-likes, we can generate separate rules for each extension, which allows the make utility to process building the extensions in parallel within each group (XS and non-XS), so we can use a more sensible dependency chain, allowing XS extensions to depend on running `mkppport` which depends on building the non-XS extensions, so `base` is guaranteed to be built before `mkppport` is run.

I've considered re-working the Win32 build to generate a separate makefile for building the extensions, but it's a lot of work for not much gain.

PR #21633, opened 2023-11-12
Skip threaded locale test on OpenBSD and Dragonfly

jkeenan:
Due to repeated testing failures on threaded builds subsequent to e2934ffc33.

### DIFF
--- a/lib/locale_threads.t
+++ b/lib/locale_threads.t
@@ -11,6 +11,8 @@ BEGIN {
     set_up_inc('../lib');
 
     skip_all_without_config('useithreads');
+    skip_all("Fails on threaded builds on OpenBSD and DragonFly BSD")
+        if ($^O =~ m/^(openbsd|dragonfly)$/);
 
     require './loc_tools.pl';
 


lib/locale_threads.t
```diff
@@ -11,6 +11,8 @@ BEGIN {
     set_up_inc('../lib');
 
     skip_all_without_config('useithreads');
+    skip_all("Fails on threaded builds on OpenBSD and Dragonfly")
```
rwp0: The styling is DragonFly with a capital "F"

Or *DragonFly BSD* (spelled fully)

https://en.wikipedia.org/wiki/DragonFly_BSD

Leont:
The commit title is less helpful than possible, otherwise it looks good to me

lib/locale_threads.t
jkeenan: Agreed, per https://www.dragonflybsd.org/

jkeenan:
> The commit title is less helpful than possible, otherwise it looks good to me

I've reworded the commit message.  If it's not yet okay, please let me know.  Otherwise, I'll merge this once we have some satisfactory smoke reports from DragonFly.

rwp0 (approved):

PR #21634, opened 2023-11-12
`valgrindpp.pl`: Fix `new` indirect syntax

rwp0:
In two places
Also format POD in few places

### DIFF
--- a/Porting/valgrindpp.pl
+++ b/Porting/valgrindpp.pl
@@ -1,4 +1,5 @@
 #!/usr/bin/perl
+
 use IO::File ();
 use File::Find qw(find);
 use Text::Wrap qw(wrap);
@@ -49,7 +50,7 @@ my $hidden = do { local $"='|'; $opt{hide} ? qr/^(?:@{$opt{hide}})$/o : '' };
 # (do it early, as it may fail)
 my $fh = \*STDOUT;
 if (exists $opt{'output-file'}) {
-  $fh = new IO::File ">$opt{'output-file'}"
+  $fh = IO::File->new($opt{'output-file'}, 'w')
         or die "$0: cannot open $opt{'output-file'} ($!)\n";
 }
 
@@ -197,7 +198,7 @@ sub filter {
 
   # Get all the valgrind output lines
   my @l = do {
-    my $fh = new IO::File $_ or die "$0: cannot open $_ ($!)\n";
+    my $fh = IO::File->new($_, 'r') or die "$0: cannot open $_ ($!)\n";
     # Process outputs can interrupt each other, so sort by pid first
     my %pid; local $_;
     while (<$fh>) {
@@ -269,7 +270,7 @@ __END__
 
 =head1 NAME
 
-valgrindpp.pl - A post processor for make test.valgrind
+valgrindpp.pl - A post processor for C<make test.valgrind>
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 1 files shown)


Porting/valgrindpp.pl
```diff
@@ -49,7 +50,7 @@
 # (do it early, as it may fail)
 my $fh = \*STDOUT;
 if (exists $opt{'output-file'}) {
-  $fh = new IO::File ">$opt{'output-file'}"
```
tonycoz, line 52: Changing these to use 3-arg open and eliminating the use of `IO::File` would be better.

rwp0: Why `open` is preferred here?

tonycoz: I don't know of any formal guidance on using IO::File over open(), but open has the 3-arg form, so you can 
```
open my $fh, ">", $opt{'output-file'}
```
safely, to get the equivalent from IO::File you need the more C style (see fopen):
```
my $fh = IO::File->new(${'output-file'}, "w")
```
The current code has the same issues as the discouraged 2 arg open, where the type of open is mixed with the filename.

rwp0: > I don't know of any formal guidance on using IO::File over open(), but open has the 3-arg form, so you can
> 
> ```perl
> open my $fh, ">", $opt{'output-file'}
> ```
> 
> safely, to get the equivalent from IO::File you need the more C style (see fopen):
> 
> ```perl
> my $fh = IO::File->new(${'output-file'}, "w")
> ```
> 
> The current code has the same issues as the discouraged 2 arg open, where the type of open is mixed with the filename.

Thanks, I didn't know of this difference.

I've always been using the native `IO::File->new('FILE', 'w')` or the `Path-Tiny` distribution instead of `open` whenever possible for object oriented-ness and readability.

Force pushed to have the nicer looking letter syntax.

Porting/valgrindpp.pl
```diff
@@ -197,7 +198,7 @@ sub filter {
 
   # Get all the valgrind output lines
   my @l = do {
-    my $fh = new IO::File $_ or die "$0: cannot open $_ ($!)\n";
+    my $fh = IO::File->new($_) or die "$0: cannot open $_ ($!)\n";
```
haarg: Should probably be `IO::File->new($_, 'r')`.

rwp0: The second argument is optional and defaults to `r` as with `open`.
https://github.com/Perl/perl5/blob/b2bbe7d8e2b196d341cd8c3e1d3b7600b81c1a41/dist/IO/lib/IO/File.pm#L154-L165

rwp0:
Can this be merged already please?

Porting/valgrindpp.pl
mauke: The second argument does not default to `r`. The linked code shows that it acts like 2-arg open, which tries to parse an open mode and a filename from the second argument. That's not the same as always opening the file for reading. For correctness, we never want to reinterpret filenames; i.e. we should always use 3-arg open or the corresponding `IO::File->new($file, $mode)` form.

rwp0: Thanks @mauke, will amend by adding the `r` second argument.

rwp0: Force-pushed

mauke (approved):

PR #21638, opened 2023-11-16
Don't use POSIX 2008 locale on Dragonfly

khwilliamson:
Tony Cook has shown with a C program that it leaks.


tonycoz:
https://bugs.dragonflybsd.org/issues/3361

tonycoz (approved):

PR #21641, opened 2023-11-16
Neater lists of header/source/object files in Makefile.SH

leonerd:
There doesn't seem to be any need to create indirect lists of c1, c2, c3,... just to concatenate them all together. It just makes maintainting the list a bit harder.

Might as well just concatente them all into a single list.

Also while I was there I sorted them alphabetically, again for ease of maintaining / adding new ones.


khwilliamson:
Without having looked carefully for missing items, LGTM

tonycoz:
This looks correct to me, though I wonder why we don't use `$(c_base:.c=$(OBJ_EXT))`

Presumably ancient makes.

tonycoz (approved):

leonerd:
> This looks correct to me, though I wonder why we don't use `$(c_base:.c=$(OBJ_EXT))`
> 
> Presumably ancient makes.

Yes I considered that, but I didn't want to adjust too much in one go. We could look at that as a next step though.

PR #21642, opened 2023-11-16
unwrap and optimise pp_aassign() and friends

iabyn:
This branch:

- makes pp_aassign() RC-stack-aware

- optimises pp_aassign() on PERL_RC_STACK builds.

- also unwraps pp_emptyavhv() pp_refassign() and pp_undef(), chiefly to avoid skewing list assignment benchmark results.

- documents and adds tests for ref alias assignment in list context

- adds rpp_replace_at_norc() and several rpp_foo_NN() variants to the API

The main intent of this branch is to unwrap and optimise pp_aassign; all the other commits came about as side-effects of that work.

The main optimisation work on pp_aassign() consists of: firstly, skipping all the temps stack juggling that it used to do to avoid prematurely freeing SVs, in such cases as @a = ($_a[0], ...). With an RC stack this is no longer necessary. Secondly, trying to avoid manipulating the stack as much as possible, now that every rpp_replace_1_1() or similar requires two ref count adjustments, one of which requires a conditional.

The Porting/bench.pl results on the expr::aassign:: subset of the benchmarks gave an averages of around 96% on Ir, COND etc compared with a non-PERL_RC_STACK blead build.

The work in this branch turned out to be a lot harder than I expected, and took around calendar months to get fully working.


sv.h
```diff
@@ -2415,7 +2415,17 @@ that already have a PV buffer allocated, but no SvTHINKFIRST.
...
    /* exercise the immortal resurrection code in sv_free2() */
-#  define SvREFCNT_IMMORTAL 1000
+#  ifdef PERL_RC_STACK
+     /* When the stack is ref-counted, the code tends to take a lot of
+      * short cuts with immortals, such as skipping the bump of the ref
+      * count of PL_sv_undef when pushing it on the stack. Exercise that
+      * this doesn't cause problems, especially on code which
+      * special-cases RC==1 etc.
+      */
+#    define SvREFCNT_IMMORTAL 10
```
tonycoz, line 2425: Is this 10 still meant to be here?

The comment this commit adds to `t/op/hash.t` says `SvREFCNT_IMMORTAL` is 1000 for DEBUGGING builds, so I suspect this is a temporary debugging change that stuck.

iabyn:
On Sun, Nov 19, 2023 at 08:57:21PM -0800, Tony Cook wrote:
> Is this 10 still meant to be here?
> 
> The comment this commit adds to `t/op/hash.t` says `SvREFCNT_IMMORTAL` is 1000 for DEBUGGING builds, so I suspect this is a temporary debugging change that stuck.

It's intentional. The original 'make it 1000 under DEBUGGING' was supposed
to trigger reanimating immortals at least once or twice, in order to
exercise the reanimating code in in sv_free2().  

As the new code comment tries to explain, under PERL_RC_STACK, it's now
much more likely that the refcount of immortals will get messed with,
so I want to test in many places what happens when the RC reaches a small
number. With it at 1000, it was just triggering a single failure in a
cpan/ module.

-- 
All wight. I will give you one more chance. This time, I want to hear
no Wubens. No Weginalds. No Wudolf the wed-nosed weindeers.
    -- Life of Brian

t/op/array.t
```diff
@@ -265,21 +265,29 @@ is ($got, '');
 }
 
 sub test_arylen {
-    my $ref = shift;
+    my ($ref, , $fixed, $desc) = @_;
```
tonycoz, line 268: Minor: the double comma doesn't seem useful and may confuse people looking over the test in the future.

tonycoz (approved):

iabyn:
Now merged.

PR #21643, opened 2023-11-16
Fix a compiler warning for 32-bit build in pp.c

t-a-k:
This should fix the warning:
```
pp.c: In function ‘Perl_pp_divide’:
pp.c:1626:3: warning: label ‘ret’ defined but not used [-Wunused-label]
 1626 |   ret:
      |   ^~~
```
This warning is triggered when `PERL_TRY_UV_DIVIDE` is left undefined, e.g. on typical 32-bit configuration with `NV_PRESERVES_UV` defined, and not triggered on typical 64-bit build where IV and NV are both 64-bit.

### DIFF
--- a/pp.c
+++ b/pp.c
@@ -1621,6 +1621,7 @@ PP(pp_divide)
 #endif
             DIE(aTHX_ "Illegal division by zero");
         TARGn(left / right, 1);
+        goto ret;               /* redundant, but silence -Wunused-label */
     }
 
   ret:


tonycoz:
I'd really rather `#ifndef` out the label.

You can get the same warning with `-Duselongdouble` on a 64-bit build.

jkeenan:
> I'd really rather `#ifndef` out the label.
> 
> You can get the same warning with `-Duselongdouble` on a 64-bit build.

Confirmed.
```
$ sh ./Configure -des -Dusedevel -Duselongdouble && make test_prep
...
cc -c -DPERL_CORE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings pp.c
pp.c: In function ‘Perl_pp_divide’:
pp.c:1626:3: warning: label ‘ret’ defined but not used [-Wunused-label]
 1626 |   ret:
      |   ^~~
```

Tux:
> I'd really rather `#ifndef` out the label.
> 
> You can get the same warning with `-Duselongdouble` on a 64-bit build.

IRC
```
Nov 09 09:26:29 <[Tux]> should the ret: label on pp.c#1626 be in #ifdef PERL_TRY_UV_DIVIDE to prevent unused warnings?
```

jkeenan:
> > I'd really rather `#ifndef` out the label.
> > You can get the same warning with `-Duselongdouble` on a 64-bit build.
> 
> IRC
> 
> ```
> Nov 09 09:26:29 <[Tux]> should the ret: label on pp.c#1626 be in #ifdef PERL_TRY_UV_DIVIDE to prevent unused warnings?
> ```

I don't have the C foo to code that up, but I'd encourage you to create a (smoke-me) branch so that I could try it out.  Thanks.

t-a-k:
I thought that such use of redundant `goto` might be controversial, but I think there are some good reasons to use `goto` here, aren't there?
* Some functions such as `pp_multiply` and `pp_subtract` in `pp.c` already (happened to?) have similar constructs.
* using `#ifdef` just for compiling out an unused label tends to make the code lengthy, IMHO.

tonycoz:
I think you're right, though it's a close thing.

The `goto ret;` was added as part of the reference counted stack changes, replacing what was `RETURN;`.

I'll approve this, but if coverity complains (it probably won't) I expect I'll change it to a `#ifndef`.

tonycoz (approved):

PR #21644, opened 2023-11-17
Supported MSCV no longer has a broken localeconv()

khwilliamson:
We no longer have to have an ugly, slow workaround that was required for VS 2013 and before.

We still need the workaround for MingW when not used with UCRT, but the 2nd commit removes it from when UCRT is used.


tonycoz (approved):

perl.h
tonycoz, line 1: perhaps `s/MSCV/MSVC/` in the commit message

PR #21647, opened 2023-11-19
extend "Possible precedence issue" warning, cover more cases

mauke:
Previously, code like

    return $a or $b;

would emit a warning of the form "Possible precedence issue with control flow operator" because it parses as (return $a) or $b.

However, the equally misleading

    exit $a != 0;      # exit($a) != 0
    exit $a ? 0 : $b;  # exit($a) ? 0 : $b

were not diagnosed.

This patch moves the check higher up in the compiler: Now any binary or ternary operator (not just logical operators) warns if its first operand is an unparenthesized control flow operator that does not terminate normally (i.e. next, last, redo, goto, exit, return, die).

### DIFF
--- a/op.c
+++ b/op.c
@@ -4299,6 +4299,52 @@ Perl_invert(pTHX_ OP *o)
     return newUNOP(OP_NOT, OPf_SPECIAL, scalar(o));
 }
 
+/* Warn about possible precedence issues if op is a control flow operator that
+   does not terminate normally (return, exit, next, etc).
+*/
+static void
+S_check_terminates(pTHX_ OP *op)
+{
+    assert(op != NULL);
+
+    /* [perl #59802]: Warn about things like "return $a or $b", which
+       is parsed as "(return $a) or $b" rather than "return ($a or
+       $b)".
+    */
+    switch (op->op_type) {
+    case OP_NEXT:
+    case OP_LAST:
+    case OP_REDO:
+        /* XXX: Perhaps we should emit a stronger warning for these.
+           Even with the high-precedence operator they don't seem to do
+           anything sensible.
+
+           But until we do, fall through here.
+         */
+    case OP_EXIT:
+    case OP_RETURN:
+    case OP_DIE:
+    case OP_GOTO:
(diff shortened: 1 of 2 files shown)


op.c
```diff
@@ -4299,6 +4299,52 @@ Perl_invert(pTHX_ OP *o)
...
+           The former can occur with e.g.
+
+                use constant FEATURE => ( $] >= ... );
+                sub { not FEATURE and return or do_stuff(); }
+         */
+        if (!op->op_folded && !(op->op_flags & OPf_PARENS))
+            Perl_ck_warner(aTHX_ packWARN(WARN_SYNTAX),
+                           "Possible precedence issue with control flow operator");
+        /* XXX: Should we optimze this to "return $a;" (i.e. remove
+           the "or $b" part)?
```
richardleach, line 4342: Might be worth doing - someone who deparses their code might notice the discrepancy even if they missed the warning.

khwilliamson: +1

mauke: This is part of the original code that I just moved into a separate function. I don't think this PR is the right place to make this kind of change.

mauke: I have now taken a stab at this idea in https://github.com/mauke/perl5/commits/fold-flow-ops. Comments welcome.

I'm still not sure if it should be added to this PR.

jkeenan:
On 11/19/23 12:52, mauke wrote:
> 
> 
> In op.c <https://github.com/Perl/perl5/pull/21647#discussion_r1398466379>:
> 
>> +        /* XXX: Should we optimze this to "return $a;" (i.e. remove
> +           the "or $b" part)?
> 
> I have now taken a stab at this idea in 
> https://github.com/mauke/perl5/commits/fold-flow-ops 
> <https://github.com/mauke/perl5/commits/fold-flow-ops>. Comments welcome.
> 
> I'm still not sure if it should be added to this PR.
> 

When in doubt, separate PRs.

op.c
richardleach: Oh yeah, fair enough, separate PR sounds good.

tonycoz (approved):

PR #21648, opened 2023-11-19
perldelta for locale stress test addition

khwilliamson:
I wonder if this goes into unnecessary detail, @karenetheridge


karenetheridge (approved):
I like verbosity :) it can always be trimmed down later for perl5400delta.

PR #21653, opened 2023-11-20
elide right operand if left operand transfers control

mauke:
This commit contains the following changes:

- The "possible precedence issue" warning now mentions (in parentheses) the name of the control flow operator that triggered the warning:

      $ ./perl -cwe 'return $a or $b'
      Possible precedence issue with control flow operator (return) at -e line 1.

- CORE::dump now counts as a control flow operator:

      $ ./perl -cwe 'CORE::dump or f()'
      Possible precedence issue with control flow operator (dump) at -e line 1.

- Any binary operator whose left operand is a control flow operator is optimized out along with any right operand it may have, even if no warning is triggered:

      $ ./perl -Ilib -MO=Deparse -we '(die $a) + $b'
      BEGIN { $^W = 1; }
      die $a;
      -e syntax OK

  (No warnings because explicit parentheses are used in the preceding example.)

      $ ./perl -Ilib -MO=Deparse -we 'exit $a ? 0 : 1'
      Possible precedence issue with control flow operator (exit) at -e line 1.
      BEGIN { $^W = 1; }
      exit $a;
      -e syntax OK

   (As discussed in PR #21647.)

### DIFF
--- a/lib/B/Deparse.t
+++ b/lib/B/Deparse.t
@@ -1857,7 +1857,7 @@ package foo;
 CORE::do({});
 CORE::do({});
 ####
-# [perl #77096] functions that do not follow the llafr
+# [perl #77096] functions that do not follow the looks-like-a-function rule
 () = (return 1) + time;
 () = (return ($1 + $2) * $3) + time;
 () = (return ($a xor $b)) + time;
@@ -1871,6 +1871,26 @@ CORE::do({});
 () = (last 1) + 3;
 () = (next 1) + 3;
 () = (redo 1) + 3;
+() = (-R $_) + 3;
+() = (-W $_) + 3;
+() = (-X $_) + 3;
+() = (-r $_) + 3;
+() = (-w $_) + 3;
+() = (-x $_) + 3;
+>>>>
+() = (return 1);
+() = (return ($1 + $2) * $3);
+() = (return ($a xor $b));
+() = (do 'file') + time;
+() = (do ($1 + $2) * $3) + time;
+() = (do ($1 xor $2)) + time;
+() = (goto 1);
+() = (require 'foo') + 3;
+() = (require foo) + 3;
+() = (CORE::dump 1);
+() = (last 1);
+() = (next 1);
+() = (redo 1);
 () = (-R $_) + 3;
 () = (-W $_) + 3;
 () = (-X $_) + 3;
@@ -1900,6 +1920,13 @@ $_ = ($a xor not +($1 || 2) ** 2);
 () = warn;
 () = warn() + 1;
 () = setpgrp() + 1;
+>>>>
+() = (eof) + 1;
+() = (return);
+() = (return, 1);
+() = warn;
+() = warn() + 1;
+() = setpgrp() + 1;
 ####
 # loopexes have assignment prec
 () = (CORE::dump a) | 'b';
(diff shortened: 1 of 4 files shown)


hvds:
I'm slightly surprised there's been no discussion about the possibility of changing the precedence of `exit` - if someone writes `exit $a ? 0 : 1` it seems pretty obvious what the intent is. Is it inconceivable that we could change the behaviour to match the intent, rather than warning and ignoring it? We don't see the same behaviour, for example, with `die $a ? 0 : 1`, nor with `exit $a + 1`.

If it is not inconceivable, we should have that discussion before applying this patch.

mauke:
Why before applying this patch? This patch does not depend on the precedence of `exit` (or any other operator). It's an entirely orthogonal issue.

That said, `exit` has the precedence it has because it is takes a single scalar argument, like `length` or `int`, so it gets precedence just below arithmetical operators but above comparisons. `exit $a + 1` works because `exit` is below `+` (arithmetic). `die $a ? 0 : 1` works because `die` is a list operator, which has normal function call precedence.

Grinnz:
This is actually consistent behavior, it can be surprising only because it looks like a user defined function and it's not common to define those with the `($)` prototype for similar behavior. (and it's not common because this precedence is surprising for things that look like functions)

hvds:
> Why before applying this patch?

It would seem counterproductive to start warning that particular uses of `exit` will not do what the author intended if we were shortly thereafter going to change the behaviour so that the warned behaviour actually does do what the author intended.

I am not questioning whether it is consistent behaviour, I am questioning whether it is useful behaviour; and, should it turn out there is some different behaviour that would be more useful, whether we have a pathway to get there from here. My guess for the second part is that the pathway is available, because there is no useful way for people to have knowingly taken advantage of the current interpretation (specifically with `exit`) except for the express purpose of obfuscation.

Generally it seems more perlish to do what the programmer wants rather than to tell them you know what they want but they can't have it because they didn't ask the right way.

Grinnz:
> > Why before applying this patch?
> 
> It would seem counterproductive to start warning that particular uses of `exit` will not do what the author intended if we were shortly thereafter going to change the behaviour so that the warned behaviour actually does do what the author intended.

If the behavior were changed, it would no longer hit the warning case, and this patch is not specific to the exit keyword.

> My guess for the second part is that the pathway is available

Unfortunately not, because exit can be overriden to do whatever the programmer desires (and is in some high profile cases like mod_perl)

mauke:
> It would seem counterproductive to start warning that particular uses of `exit` will not do what the author intended if we were shortly thereafter going to change the behaviour so that the warned behaviour actually does do what the author intended.

That's not what this patch does.

tonycoz:
> I'm slightly surprised there's been no discussion about the possibility of changing the precedence of `exit` - if someone writes `exit $a ? 0 : 1` it seems pretty obvious what the intent is. Is it inconceivable that we could change the behaviour to match the intent, rather than warning and ignoring it? We don't see the same behaviour, for example, with `die $a ? 0 : 1`, nor with `exit $a + 1`.
> 
> If it is not inconceivable, we should have that discussion before applying this patch.

exit() has that precedence because it's a named unary operator - are you also suggesting we change the precedence of all such operators?

tonycoz (approved):

PR #21657, opened 2023-11-22
harness: fix operator precedence in call to exit

mauke:
Previously:

    $ ./perl -w -c t/harness
    Possible precedence issue with control flow operator at t/harness line 498.
    t/harness syntax OK


leonerd (approved):
LGTM

PR #21660, opened 2023-11-22
sv_isa_sv() no longer considered experimental

leonerd:
Removing the `x` flag will remove the "experimental" wording from the generated perlapi.pod page.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -3146,7 +3146,7 @@ Adp	|void	|sv_insert_flags|NN SV * const bigstr			\
 Adp	|IO *	|sv_2io 	|NN SV * const sv
 Adp	|int	|sv_isa 	|NULLOK SV *sv				\
 				|NN const char * const name
-ARdpx	|bool	|sv_isa_sv	|NN SV *sv				\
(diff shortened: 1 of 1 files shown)


rwp0 (approved):

jkeenan:
> Removing the `x` flag will remove the "experimental" wording from the generated perlapi.pod page.

@leonerd, when I grep the repository for `sv_isa_sv` (*e.g.,* by `ack -C3 sv_isa_sv .`), I don't find any discussion of the experimental status of this API function other than the `x` you're proposing to delete from `embed.fnc`.

Could you provide a brief explanation of this API, how it came to be, and why it has been deemed experimental up until now?  Thanks.

johannessen:
It’s the XS equivalent of Perl’s new-ish `isa` operator. Since `isa` underwent an experimental phase, so did `sv_isa_sv()`.

Apparently it wasn’t de-experimentalised along with `isa` in 5.36 due to an oversight.

leonerd:
Yup, all of what @johannessen just said.

Introduced in 5.32, was due to un-experimenting in 5.36 when `isa` itself was, but I just forgot about it.

jkeenan (approved):

PR #21663, opened 2023-11-23
bisect: explicitly set diff path prefixes

mauke:
The patch code in Porting/bisect-runner.pl assumes `git show` will always display filenames with "a/..." and "b/..." prefixes. This is not necessarily the case. For example, the config settings diff.mnemonicPrefix and diff.noprefix can cause different prefixes or no prefixes at all to be used. In that case (e.g. after `git config --global diff.noprefix true`), bisect will halt like this:

    HEAD is now at 95388f2eb2 perlhist: Move 5.26.0 to another table to fix line length
    can't find file to patch at input line 24
    Perhaps you used the wrong -p or --strip option?
    The text leading up to this was:
    --------------------------
    |commit 46bfb3c49f22629a21d99ff2d02461a2a8a9aac4
    |Author: Nicholas Clark <nick@ccl4.org>
    |Date:   2022-08-02 16:40:39 +0200
    |
    |    Configure should avoid looping infinitely repeating the same question
    |
    |    Configure's helper function ./myread is intended to loop until it gets an
    |    acceptable answer. For a couple of cases, an empty string is not acceptable
    |    (eg 'Where is your C library?', if all automated attempts at answering this
    |    have failed). In these cases, if Configure's standard input is /dev/null (or
    |    closed), the shell read returns an empty string, and ./myread repeats the
    |    question.
    |
    |    Before this commit, it would loop infinitely (generating continuous terminal
    |    output). With this commit, we add a retry counter - if it asks the same
    |    question to many times, it aborts Configure. This means that unattended
    |    ./Configure runs should now always terminate, as termination with an error
    |    is better than spinning forever.
    |
    |diff --git Configure Configure
    |index 9d7fc39419..9d3b5f73b1 100755
    |--- Configure
    |+++ Configure
    --------------------------
    File to patch:

... which seems silly, seeing as how the filename is right there ("diff --git Configure Configure", "--- Configure", "+++ Configue") and yet bisect has to ask which file to patch ("it's Configure, duh!").

This commit tells git explicitly to use "a/" and "b/" as prefixes, which works no matter what the local config settings are.

### DIFF
--- a/Porting/bisect-runner.pl
+++ b/Porting/bisect-runner.pl
@@ -1606,23 +1606,27 @@ sub apply_patch {
     die_255("Can't $what$files: $?, $!");
 }
 
+sub patch_from_commit {
+    my ($revert, $commit, @files) = @_;
+    my $flags = $revert ? '-R ' : '';
+    my $patch = `git show --src-prefix=a/ --dst-prefix=b/ $flags$commit @files`;
+    if (!defined $patch) {
+        my $thing = $revert ? 'revert commit' : 'commit';
+        die_255("Can't get $thing $commit for @files: $?") if @files;
+        die_255("Can't get $thing $commit: $?");
+    }
+    return $patch;
+}
+
 sub apply_commit {
     my ($commit, @files) = @_;
-    my $patch = `git show $commit @files`;
-    if (!defined $patch) {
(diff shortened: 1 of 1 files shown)


jkeenan (approved):
I tried it out. LGTM, but would appreciate one other person to confirm.

PR #21664, opened 2023-11-23
grow the tmps (mortal) stack exponentially rather than linearly

### LINKED ISSUE #21654: [win32] Both IO and global match are VERY SLOW since 5.12 and up to 5.38 unless pre-heated
Hello,

in [this thread](https://www.perlmonks.org/index.pl?node_id=11155604) at perlmonks an anonymous contributor spotted a severe performance drop in perl under windows for both IO and global match.

I have tested the code against different strawberry distros in [my answer](https://www.perlmonks.org/index.pl?node_id=11155616) and seems something starting getting worst after 5.12 and still present in 5.38

Pre-heating (predimensionate the array) seems to mitigate the issue after 5.26

Someone in the chat supposed to be something related to malloc on win32.

I'm not in the position to dig it up further.

Thanks for looking and, if something is found, I'll take care to update the above thread.

L*

tonycoz:
As with the value stack and the save stack, this gives us constant amortized growth per element.

After this patch the profiler shows the "SvPV_shrink_to_cur(sv)" and "sv = sv_2mortal(newSV(80))" calls in do_readline as the hotspots for the io unheated test case, using 55% of the measured time in total.

Fixes #21654

### DIFF
--- a/scope.c
+++ b/scope.c
@@ -246,8 +246,12 @@ Perl_tmps_grow_p(pTHX_ SSize_t ix)
 {
     SSize_t extend_to = ix;
 #ifndef STRESS_REALLOC
-    if (ix - PL_tmps_max < 128)
-        extend_to += (PL_tmps_max < 512) ? 128 : 512;
+    SSize_t grow_size = PL_tmps_max < 512 ? 128 : PL_tmps_max / 2;
+    if (extend_to > SSize_t_MAX - grow_size - 1)
+        /* trigger memwrap message or fail allocation */
+        extend_to = SSize_t_MAX-1;
+    else
+        extend_to += grow_size;
 #endif
     Renew(PL_tmps_stack, extend_to + 1, SV*);
     PL_tmps_max = extend_to + 1;


scope.c
```diff
@@ -246,8 +246,12 @@ Perl_tmps_grow_p(pTHX_ SSize_t ix)
 {
     SSize_t extend_to = ix;
 #ifndef STRESS_REALLOC
-    if (ix - PL_tmps_max < 128)
-        extend_to += (PL_tmps_max < 512) ? 128 : 512;
+    SSize_t grow_size = PL_tmps_max < 512 ? 128 : PL_tmps_max / 5;
```
leonerd: Why divide 5? This seems like it grows the stack only by +20% each time - which feels like a fairly small growth ratio to me.

tonycoz: It's the same as AVs and the value stack.  The save stack effectively uses 2 `((old) * 3 / 2)`.

I'm happy with some other factor since this is a small part of the cost.

tonycoz:
Draft since I think I should add a test, hopefully one with not too many false positives (it will be a timing test)

scope.c
tonycoz: I've changed this to +50%

tonycoz:
Adding a performance test was too flakey, back to untested.

PR #21668, opened 2023-11-24
locale_threads.t: Better handle weird locales

khwilliamson:
The previous code was generating bunches of uninitialized variable warnings, due to 1) not checking for definedness early; 2) the loop termination needs to be reevaluated each time because there is a potential splice, shortening the array.

This only happens, I believe, on MingW not using UCRT.

### DIFF
--- a/lib/locale_threads.t
+++ b/lib/locale_threads.t
@@ -845,18 +845,16 @@ SKIP: {
     # and then a setlocale of the resulting one.  Discard locales which have
     # any unacceptable name
     if (${^O} eq "MSWin32" && $Config{'libc'} !~ /ucrt/) {
-        for my $i (0 .. $#locales) {
-            my $locale_name = $locales[$i]->{locale_name};
+        @locales = grep {
+            my $locale_name = $_->{locale_name};
             my $underlying_name = setlocale(&LC_CTYPE, $locale_name);
 
             # Defeat any attempt to skip the setlocale if the same as current,
             # by switching to a locale very unlikey to be the current one.
             setlocale($LC_ALL, "Albanian");
 
-            next if setlocale(&LC_CTYPE, $underlying_name);
-
-            splice @locales, $i, 1;
-        }
+            defined($underlying_name) && setlocale(&LC_CTYPE, $underlying_name)
+        } @locales;
     }
 
     # Create a hash of the errnos:


lib/locale_threads.t
```diff
@@ -845,15 +845,17 @@ SKIP: {
...
-            setlocale($LC_ALL, "Albanian");
+                # Defeat any attempt to skip the setlocale if the same as current,
+                # by switching to a locale very unlikey to be the current one.
+                setlocale($LC_ALL, "Albanian");
 
-            next if setlocale(&LC_CTYPE, $underlying_name);
+                next if setlocale(&LC_CTYPE, $underlying_name);
+            }
 
             splice @locales, $i, 1;
```
mauke: This still looks like it both splices and increments `$i` if `setlocale` fails, so it skips over some elements.

I think this loop should either be a `grep` ...
```perl
@locales = grep {
    my $locale_name = $locales[$i]->{locale_name};
    my $underlying_name = setlocale(&LC_CTYPE, $locale_name);

    # Defeat any attempt to skip the setlocale if the same as current,
    # by switching to a locale very unlikey to be the current one.
    setlocale($LC_ALL, "Albanian");

    defined($underlying_name) && setlocale(&LC_CTYPE, $underlying_name)
} @locales;
```
... or count backwards:
```perl
for my $i (reverse 0 .. $#locales) {
   my $locale_name = $locales[$i]->{locale_name};
    my $underlying_name = setlocale(&LC_CTYPE, $locale_name);

    # Defeat any attempt to skip the setlocale if the same as current,
    # by switching to a locale very unlikey to be the current one.
    setlocale($LC_ALL, "Albanian");

    if (!(defined($underlying_name) && setlocale(&LC_CTYPE, $underlying_name))) {
        splice @locales, $i, 1;
    }
}
```
... so `splice` does not affect the indices of not-yet-visited elements.

khwilliamson: I used your grep solution

jkeenan:
> The previous code was generating bunches of uninitialized variable warnings, due to 1) not checking for definedness early; 2) the loop termination needs to be reevaluated each time because there is a potential splice, shortening the array.
> 
> This only happens, I believe, on MingW not using UCRT.

Could we get someone who can build this branch on Windows/MingW without using UCRT and give us a shout-out?  Thanks.

khwilliamson:
I have tried it on MingW, and they no longer appear

PR #21676, opened 2023-11-27
make generic "Out of memory!" error more specific

### LINKED ISSUE #21672: Say "Perl: Out of memory", not just "Out of memory"
Please have perl print
> Perl: out of Memory

not just

> Out of memory.

Else it is hard to tell what printed that message.

https://github.com/szabgab/perlmaven.com/issues/583

Yes, I know perl doesn't like to say "Perl: " before its
messages. But it would be nice if it was like bash, and did.

mauke:
Before:
```
$ perl -e '$_ = "a" x ~0'
Out of memory!
panic: fold_constants JMPENV_PUSH returned 2 at -e line 1.
```
After:
```
$ perl -e '$_ = "a" x ~0'
Out of memory in perl:util:safesysrealloc
panic: fold_constants JMPENV_PUSH returned 2 at -e line 1.
```

Fixes #21672.


guest20:
Cool patch, @mauke !

Could the `x` / `unpack` / `[]->[~0] ++` cases also become catchable or is that too much visible plumbing?

mauke:
@guest20 I don't really see how. Something like `[]->[[]]++` could genuinely succeed on a machine with enough memory, so it has to try to allocate. The lvalue `vec` cases mentioned above are catchable because they exceed what perl can represent internally, so it doesn't even try to allocate anything.

Changing the way memory allocation is handled in general would be a much bigger effort.

guest20:
>  lvalue vec cases [] are catchable because they [] doesn't even try to allocate anything.

That's a cool thing I just learnt. Thanks.

tonycoz (approved):

PR #21678, opened 2023-11-29
Add a work-around for older Perl's that need `@{^CAPTURE}`

scottchiefbaker:
I was bit by a bug wherein I was accessing `@{^CAPTURE}` on Perl v5.16 which does not have this variable. This PR adds some documentation to allow a user a work-around if they are in the same situation.

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -1096,8 +1096,8 @@ An array which exposes the contents of the capture buffers, if any, of
 the last successful pattern match, not counting patterns matched
 in nested blocks that have been exited already.
 
-Note that the 0 index of @{^CAPTURE} is equivalent to $1, the 1 index
-is equivalent to $2, etc.
+Note that the 0 index of C<@{^CAPTURE}> is equivalent to C<$1>, the 1 index
+is equivalent to C<$2>, etc.
 
     if ("foal"=~/(.)(.)(.)(.)/) {
         print join "-", @{^CAPTURE};
@@ -1119,6 +1119,19 @@ see L<perldata/"Demarcated variable names using braces"> for more
 information on this form and its uses.
 
 This variable was added in 5.25.7
+
+If you need access to this functionality in older Perls you can use this
+function immediately after your regexp.
+
+	sub get_captures {
+		no strict 'refs';
+
+		my $last_idx = scalar(@-) - 1;
+		my @arr      = 1 .. $last_idx;
+		my @ret      = map { $$_; } @arr;
+
+		return @ret;
+	}
 
 =item $MATCH
 


scottchiefbaker:
I'm not sure if this is the type of thing we want/need in the documentation, but it would have saved me about an hour of hair pulling if it were documented.

scottchiefbaker:
@mauke pointed out on IRC that this functionality is also in `Data::Munge`. Their implementation is a little hacky, shorter, and harder to read. For completeness I'm including it here:

```
sub get_captures {
    no strict 'refs';
    map $$_, 1 .. $#+
}
```

We could include this in the documentation if my function is too long/wordy.

pod/perlvar.pod
```diff
@@ -1120,6 +1120,26 @@ information on this form and its uses.
 
 This variable was added in 5.25.7
 
+If you need access to this functionality in older Perl's you can use this
+function immediately after your regexp.
+
+    sub get_captures {
+        my $str      = $&;
+        my @captures = ();
```
mauke: This declares an empty array `@captures`, then immediately assigns an empty list to it. This assignment is redundant.

pod/perlvar.pod
```diff
@@ -1120,6 +1120,26 @@ information on this form and its uses.
 
 This variable was added in 5.25.7
 
+If you need access to this functionality in older Perl's you can use this
```
mauke: `s/Perl's/Perls/`

pod/perlvar.pod
```diff
@@ -1120,6 +1120,26 @@ information on this form and its uses.
...
+function immediately after your regexp.
+
+    sub get_captures {
+        my $str      = $&;
```
mauke: At least in older Perls, the use of `$&` slows down all regex matches everywhere in the program.

pod/perlvar.pod
```diff
@@ -1120,6 +1120,26 @@ information on this form and its uses.
...
+        my @captures = ();
+
+        # Build array using the offsets from @- and @+
+        for (my $i = 1; $i < @-; $i++) {
```
mauke: Why use a C-style for loop here? This is a simple counting loop, so
```perl
for my $i (1 .. $#-) {
```
expresses the idea more directly (and slightly more efficient).

pod/perlvar.pod
```diff
@@ -1120,6 +1120,26 @@ information on this form and its uses.
...
+            my $start = $-[$i];
+            my $end   = $+[$i];
+            my $len   = $end - $start;
+
+            my $str = substr($str, $start, $len);
+            push(@captures, $str);
```
mauke:
```perl
my @captures;
for (RANGE) {
    push(@captures, ...)
}
```
could also be written as
```perl
my @captures = map {
    ...
} RANGE;
```

pod/perlvar.pod
scottchiefbaker: Agree to disagree here... `$#-` is too much punctuation and not as easily understood.

I've been using Perl for 20+ years and I just learned about `#` syntax in arrays today.

pod/perlvar.pod
scottchiefbaker: I always try and be extra clear what the intent of my code is. Especially if a novice comes across it.

pod/perlvar.pod
scottchiefbaker: Good catch... I will update the PR tomorrow when I'm in front of the correct machine.

pod/perlvar.pod
mauke: If you don't like `$#-`, `@- - 1` is the same thing. (Doesn't really look better with `@-`, though, IMHO).

scottchiefbaker:
Upon further reflection I think it makes sense to use the format that @mauke proposed in `Data::Munge`. I would however like to expand it slightly so it's more clear and readable for novices. Since the intention is to put it in the documentation I think readability is key here. Here is my expanded version:

```
sub get_captures {
    no strict 'refs';

    my $last_idx = scalar(@-) - 1;
    my @arr      = 1 .. $last_idx;
    my @ret      = map { $$_; } @arr;

    return @ret;
}
```

khwilliamson:
I have quibbles about the commit messages, in particular the titles.  A title should have enough information for someone scanning a log of commits to be able to quickly determine if this is potentially a commit that has bearing on what they're scanning for.   For that, the title should include what the commit pertains to.  For small specific changes like this, noting in the title that it is perlvar that is affected will tell me immediately if I should look further into it or not.  I have adopted the convention for myself to place the file name at the beginning of the message, like 

perlvar: Use a faster/shorter version of get_captures()

Something that affects more things wouldn't have the file name, but should indicate what area, as explicitly as possible, is affected.  

A  commit message should answer the classic questions "Who, what, when, where, and why".  The who and when are automatically filled in for you by the system, and appear adjacent to the title, but they are less important than the other three.  A title should give at least a hint  towards those.  So "perlvar: Fix a typo" is better than "perlvar: Fix a word" because it more explicitly says what is wrong that needs to be fixed.

scottchiefbaker:
@khwilliamson all that makes sense to me. Usually that all gets corrected on the squash and merge phase. Unless you'd like me to rebase and squash all these down to a single commit before then.

khwilliamson:
On 12/6/23 09:11, Scott Baker wrote:
> @khwilliamson <https://github.com/khwilliamson> all that makes sense to 
> me. Usually that all gets corrected on the squash and merge phase. 
> Unless you'd like me to rebase and squash all these down to a single 
> commit before then.

Just as long as it doesn't get merged with uncorrected wording is fine 
with me

scottchiefbaker:
Ping... what's next on this? Is this something we need in the docs?

khwilliamson:
I think now is a good time to squash and make sure the commit msg is good.  Then I would look at the full result.

scottchiefbaker:
@khwilliamson I have squashed

scottchiefbaker:
@khwilliamson Thank you sir!

PR #21691, opened 2023-12-05
Archive-Tar sync from CPAN with modification

jkeenan:


leonerd (approved):
LGTM

rwp0:
Fixes https://github.com/Perl/perl5/issues/21402

jkeenan:
@haarg, does this approach look okay to you (at least as a temporary measure)?

PR #21692, opened 2023-12-05
unwrap pp_readline(), pp_glob() and a few others

iabyn:
unwrap:

    pp_index()
    pp_readline()
    pp_rcatline()
    pp_glob()
    pp_rv2gv()

and some associated tweaking to other dependant stuff.

### DIFF
--- a/dist/IO/IO.xs
+++ b/dist/IO/IO.xs
@@ -553,7 +553,11 @@ PPCODE:
     if (!ix && GIMME_V != G_LIST)
         Perl_croak(aTHX_ "Can't call $io->getlines in a scalar context, use $io->getline");
     Zero(&myop, 1, UNOP);
+#if defined(PERL_VERSION_GE) && PERL_VERSION_GE(5,39,5)
+    myop.op_flags = (ix ? (OPf_WANT_SCALAR | OPf_STACKED) : OPf_WANT_LIST);
+#else
     myop.op_flags = (ix ? OPf_WANT_SCALAR : OPf_WANT_LIST ) | OPf_STACKED;
+#endif
     myop.op_ppaddr = PL_ppaddr[OP_READLINE];
     myop.op_type = OP_READLINE;
     myop.op_next = NULL; /* return from the runops loop below after 1 op */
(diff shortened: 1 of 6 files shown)


dist/IO/IO.xs
```diff
@@ -553,18 +553,24 @@ PPCODE:
     if (!ix && GIMME_V != G_LIST)
         Perl_croak(aTHX_ "Can't call $io->getlines in a scalar context, use $io->getline");
     Zero(&myop, 1, UNOP);
+#if defined(PERL_VERSION_GE) && PERL_VERSION_GE(5,39,5)
```
tonycoz, line 556: `ppport.h` defines `PERL_VERSION_GE` so there's no need to check this here and below.

pp_hot.c
```diff
@@ -3979,6 +4135,7 @@ Perl_do_readline(pTHX)
         }
     }
     else {
+        /* XXX on RC builds, push on stack rather than moralise ? */
```
tonycoz, line 4138: "moralize" -> "mortalize" here and below.

Even if we don't push it to the value stack, it might be good to re-use the mortal stack entry, but I can see that being out of scope for this series of commits.

See #21654 for what led to this comment.  fea90cfbe1f reduced the CPU cost of such incremental tmps stack extensions, but we still do a lot of avoidable clean up of the mortals.

Only the spelling needs to be fixed to resolve this comment.

tonycoz:
Except for the minor things above I think it's good.

I'd forgotten the magic behind CALLRUNOPS() so it took me a while to figure out how the reference counted stacking handling by pp_readline/do_readline got translated into the non-RC handling expected by Perl_xs_wrap.

iabyn:
Now merged.
I fixed up the PERL_VERSION_GE, and the  'moralise' typos, as suggested by Tony. Removing the mortalising for list context readline is on my list of things to look at again, but I decided it was out of scope for this PR (which had already grown from 'unwrap pp_rv2gv' to 'unwrap all the readline and glob processing machinery'.)

PR #21693, opened 2023-12-05
allow "used only once" warnings to be fatal

### LINKED ISSUE #13814: "Name "%s::%s" used only once" warnings are not fatalized
Migrated from [rt.perl.org#121834](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=121834) (status was 'open')

Searchable as RT121834$

tonycoz:
"used only once" warnings are special, instead of being emitted at the code where the name in question is used, they are emitted during a scan of the symbol table done after parsing has finished.

This meant that any FATAL flags set in the COP for the parse point of the name is no longer in scope, so the warnings we emit can't be treated as fatal.

To make them behave as FATAL set a new flag on the name if fatal WARN_ONCE warnings are enabled and use that to dispatch the warnings as normal or fatally when we do the symbol table scan.

I originally approached the dispatch as fatal or non-fatal by messing around with cop_warnings, but that was dumb, and I went for a much simpler change.

Fixes #13814


gv.c
```diff
@@ -2819,11 +2823,20 @@ Perl_gv_check(pTHX_ HV *stash)
                 CopFILEGV(PL_curcop)
                     = gv_fetchfile_flags(file, HEK_LEN(GvFILE_HEK(gv)), 0);
 #endif
-                Perl_warner(aTHX_ packWARN(WARN_ONCE),
-                        "Name \"%" HEKf "::%" HEKf
-                        "\" used only once: possible typo",
-                            HEKfARG(HvNAME_HEK(stash)),
-                            HEKfARG(GvNAME_HEK(gv)));
+                if (GvONCE_FATAL(gv)) {
+                    Perl_fatal_warner(aTHX_ packWARN(WARN_ONCE),
```
leonerd: Does this need to use the full `Perl_...(aTHX_...)` version of a function call? Is the macro generated in `embed.h` visible here, so this could be just `fatal_warner(packWARN...)`?

gv.c
```diff
@@ -2819,11 +2823,20 @@ Perl_gv_check(pTHX_ HV *stash)
...
+                                      "Name \"%" HEKf "::%" HEKf
+                                      "\" used only once: possible typo",
+                                      HEKfARG(HvNAME_HEK(stash)),
+                                      HEKfARG(GvNAME_HEK(gv)));
+                }
+                else {
+                    Perl_warner(aTHX_ packWARN(WARN_ONCE),
```
leonerd: Ditto here

util.c
```diff
@@ -2135,6 +2135,16 @@ any of the categories are by default enabled.
 =for apidoc vwarner
 This is like C<L</warner>>, but C<args> are an encapsulated argument list.
 
+=for apidoc fatal_warner
+
+This outputs a warning as if fatal warnings were enabled.  This may
+return in some cases.
```
leonerd: I can't quite follow from these words what it is that the function does. Can you perhaps use more words to describe it?

gv.c
tonycoz: New-fangled varargs macros 😵 

I've updated both of these.

util.c
tonycoz: I've added more words. :smile:

leonerd (approved):
LGTM

PR #21695, opened 2023-12-06
Fcntl: export O_TMPFILE on request (Linux specific)

### LINKED ISSUE #21694: Add O_TMPFILE to Fcntl?
`Fcntl` is currently lacking `O_TMPFILE` which is available in recent Linux versions (3.11 and newer, possibly backported to some 3.10 e.g. CentOS 7). `O_TMPFILE` could be useful to implement temporary files which are guaranteed to be cleaned up after process exit (something which is not true for temporary files created by `File::Temp`). Currently one needs to use a CPAN module (e.g. `Linux::TempFile`) to make use of it, or use somehow h2ph, or hardcode the constant value.

Maybe it could also be systematically checked if there are other useful constants from `fcntl.h` which could be added to `Fcntl`.

mauke:
On Linux, several filesystems support the O_TMPFILE flag in open(2), which allows creating an unlinked (unnamed) file. Check for the O_TMPFILE flag and if present, export it from the Fcntl module on request.

Fixes #21694.


khwilliamson (approved):

leonerd (approved):
LGTM

Don't forget to perldelta

rwp0 (approved):

Issue #21697, opened 2023-12-08
[Win32] lib/locale.t test failures in perl-5.39.2 and onwards


sisyphus:
**Description**

Firstly, when I build perl using <code>gcc version 13.2.0 (MinGW-W64 x86_64-ucrt-mcf-seh, built by Brecht Sanders)</code> there is no problem.
But when I build perl using <code>gcc version 13.1.0 (MinGW-W64 x86_64-msvcrt-posix-seh, built by Brecht Sanders)</code> (as shipped with StrawberryPerl-5.38.0.1) I get the test failures detailed below.

Is this a known issue ? ... or am I the only person seeing it ?
Has it already been reported ?

**Steps to Reproduce**

Use the gcc toolchain provided with SP-5.38.0 to build current blead (or any devel release of perl back to 5.39.2) and then, from the 't' directory run:
```
..\perl -I..\lib -T ../lib/locale.t
```
In the output you should then find:
```
....
not ok 381 Verify that /[[:upper:]]/ matches all alpha X for which uc(X) == X and lc(X) != X
not ok 382 Verify that /[[:lower:]]/i matches all alpha X for which uc(X) == X and lc(X) != X
not ok 383 Verify that /[[:lower:]]/ matches all alpha X for which lc(X) == X and uc(X) != X
not ok 384 Verify that /[[:upper:]]/i matches all alpha X for which lc(X) == X and uc(X) != X
not ok 385 Verify that \w and [:word:] are identical
not ok 386 Verify that \d and [:digit:] are identical
not ok 387 Verify that \s and [:space:] are identical
not ok 388 Verify that [:posix:] and [:^posix:] are mutually exclusive
not ok 389 Verify that [:lower:] contains at least a-z
not ok 390 Verify that [:lower:] is a subset of [:alpha:]
not ok 391 Verify that [:upper:] contains at least A-Z
not ok 392 Verify that [:upper:] is a subset of [:alpha:]
not ok 393 Verify that /[[:lower:]]/i is a subset of [:alpha:]
not ok 394 Verify that [:alpha:] is a subset of [:alnum:]
not ok 395 Verify that [:digit:] contains at least 0-9
not ok 396 Verify that [:digit:] is a subset of [:alnum:]
not ok 397 Verify that [:digit:] matches either 10 or 20 code points
not ok 398 Verify that if there is a second set of digits in [:digit:], they are consecutive
[...]

PR #21701, opened 2023-12-08
lib/locale.t: Fix#21697

khwilliamson:
This was caused by two failures; the two commits fix one each.

### DIFF
--- a/lib/locale.t
+++ b/lib/locale.t
@@ -1039,6 +1039,8 @@ foreach my $Locale (@Locale) {
     debug "\n";
     debug "Locale = $Locale\n";
 
+    setlocale(&POSIX::LC_ALL, "C");     # Clear out old code set for a fresh
+                                        # start
     unless (setlocale(&POSIX::LC_ALL, $Locale)) {
         $setlocale_failed{$Locale} = $Locale;
 	next;


sisyphus:
It does indeed fix #21697.
LGTM.

tonycoz:
Shouldn't the use of `_wsetlocale()` prevent the the problem you describe in the second commit?

I'm lost in the macros, but I thought on Windows we always treated the locale name as UTF-8 (via S_wrap_wsetlocale()) and ignore the current code page.

khwilliamson:
I don't know if @tonycoz meant to imply that this is a bug, and the commit is just a work around for it.  But it got me to thinking that this is the case.  

So I used gdb to examine what's going on, and it appears to be in `_wsetlocale()` or perhaps our misunderstanding of it; the documentation may or may not be defective.

In locale.c, `S_wrap_wsetlocale()` is getting a UTF-8 encoded string containing the sequence `\xC3\xBC`  This is U+FC, an umlauted `u` which replaces the ASCII `u` in "Turkish".  A function is called to turn that into a `wchar_t` string.  That string looks like
``T 00 u 00 r 00 k 00 i 00 s 00 h 00 _ 00 T 00 FC 00 r 00 k 00 i 00 y 00 e 00 . 00 1 00 2 00 5 00 4 00``

`_wsetlocale` sets errno to 42 and returns NULL.  The MS errno page says 42 means 

``EILSEQ 	Illegal sequence of bytes (for example, in an MBCS string).``

I don't understand why

tonycoz:
> I don't know if @tonycoz meant to imply that this is a bug, and the commit is just a work around for it. But it got me to thinking that this is the case.

That's what I was thinking, whether that's a Perl or _wsetlocale() bug I don't know.

> In locale.c, `S_wrap_wsetlocale()` is getting a UTF-8 encoded string containing the sequence `\xC3\xBC` This is U+FC, an umlauted `u` which replaces the ASCII `u` in "Turkish". A function is called to turn that into a `wchar_t` string. That string looks like `T 00 u 00 r 00 k 00 i 00 s 00 h 00 _ 00 T 00 FC 00 r 00 k 00 i 00 y 00 e 00 . 00 1 00 2 00 5 00 4 00`
> 
> `_wsetlocale` sets errno to 42 and returns NULL. The MS errno page says 42 means
> 
> `EILSEQ Illegal sequence of bytes (for example, in an MBCS string).`
> 
> I don't understand why

setlocale()/_wsetlocae() are [documented](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170) to not set `errno` when the locale name string is invalid, so I suspect the `EILSEQ` is noise, since my own testing below showed `errno`=0 for failures that appear to be on the locale name string.

I tried the following:
```
#include <locale.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  //wchar_t *p =_wsetlocale(LC_ALL, L"Turkish_T\xFCrkiye.1254");
  //wchar_t *p =_wsetlocale(LC_ALL, L".ACP");
  //wchar_t *p =_wsetlocale(LC_ALL, L"Turkish_Turkiye.1254");
  wchar_t *p =_wsetlocale(LC_ALL, L"T\xFCrkish_T\xFCrkiye.1254");
  if (!p) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(p);
}
```
Except for the .ACP case they all failed with "wsetlocale: No error" (tested with MSVC 2022, gcc-msvcrt 13.2.1)

I used [godbolt.org](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:20,endLineNumber:3,positionColumn:20,positionLineNumber:3,selectionStartColumn:20,selectionStartLineNumber:3,startColumn:20,startLineNumber:3),source:'%23include+%3Clocale.h%3E%0A%23include+%3Cstdio.h%3E%0A%23include+%3Cstdlib.h%3E%0A%0Aint+main()+%7B%0A++wchar_t+*p+%3D_wsetlocale(LC_ALL,+L%22Turkish_T%5CxFCrkiye.1254%22)%3B%0A++if+(!!p)+%7B%0A++++perror(%22wsetlocale%22)%3B%0A++++exit(1)%3B%0A++%7D%0A++_putws(p)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:47.762191048764194,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:vc_v19_37_x64,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x64+msvc+v19.37+(Editor+%231)',t:'0')),header:(),k:50,l:'4',m:80.48533872598584,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x64+msvc+v19.37+(Compiler+%231)',t:'0')),header:(),l:'4',m:19.514661274014156,n:'0',o:'',s:0,t:'0')),k:52.2378089512358,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) to check whether the generated wide strings were what I expected (a UTF-8 encoded source file with a literal `ü` did not produce good results.)

khwilliamson:
I don't think that locale is supported on MSVC.  Instead, I tested your program on MingW, modified as follows

````
#include <locale.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  wchar_t *q =_wsetlocale(LC_ALL, L"Turkish_T\xFCrkiye.1254");
  if (!q) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(q);
  wchar_t *r =_wsetlocale(LC_ALL, L"Thai_Thailand.874");
  if (!r) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(r);
  wchar_t *p =_wsetlocale(LC_ALL, L"Turkish_T\xFCrkiye.1254");
  if (!p) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(p);
}
````

And got the following:

````
Turkish_Türkiye.1254
Thai_Thailand.874
wsetlocale: Illegal byte sequence
````
so the bug is in `_wsetlocale` and the uestion becomes what are we going to do about it.  The bug is there, and all the remaining patch does is work around it in our test file.   Some Perl program may come along and innocently happen to switch from Thai to Turkish and hit this bug.  Note that this is a MingW not built with UCRT.  This Configuration should be going away soon.

**Comment by sisyphus** [comment]
Does the proposed removal of the <code>%setlocale_failed</code> part of the condition weaken the tests on the non-problematic systems ?
If so, the <code>%setlocale_failed </code> could instead be replaced with <code>(%setlocale_failed && $Config{libc} ne '-lmsvcrt')</code>.
Similarly, I take it that the newly inserted <code>setlocale(&POSIX::LC_ALL, "C");</code> call really only needs to be made if <code>$Config{libc} eq '-lmsvcrt'</code>.
I don't know if there's anything to be gained by introducing any of that extra clutter.

I ran @khwilliamson's C script on various 32-bit and 64-bit mingw-w64 compilers, and also on 32-bit and 64-bit VS 2022.
The only time I saw the <code>wsetlocale: Illegal byte sequence</code> complaint was on the MSVCRT mingw-w64 compilers.
Otherwise, it performed correctly.

**Comment by tonycoz** [comment]
Frustratingly I get different results:

```
C:\Users\Tony\dev\perl\git>type wsetlocale2.c
#include <locale.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  wchar_t *q =_wsetlocale(LC_ALL, L"Turkish_T\xFCrkiye.1254");
  if (!q) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(q);
  wchar_t *r =_wsetlocale(LC_ALL, L"Thai_Thailand.874");
  if (!r) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(r);
  wchar_t *p =_wsetlocale(LC_ALL, L"Turkish_T\xFCrkiye.1254");
  if (!p) {
    perror("wsetlocale");
    exit(1);
  }
  _putws(p);
}

C:\Users\Tony\dev\perl\git>gcc --version | find "gcc"
gcc (MinGW-W64 x86_64-msvcrt-posix-seh, built by Brecht Sanders) 13.1.0

C:\Users\Tony\dev\perl\git>gcc -owsetlocale2.exe wsetlocale2.c

C:\Users\Tony\dev\perl\git>.\wsetlocale2
wsetlocale: No error

C:\Users\Tony\dev\perl\git>ver

Microsoft Windows [Version 10.0.19045.3693]
```
`dumpbin /imports` shows the executable is linked to `msvcrt.dll`.

I was hoping we could `setlocale(LC_ALL, "C");` to get consistent encoding handling there, but it's not working for me by default.

sisyphus:
> Frustratingly I get different results

For me, the Windows version is <code>Microsoft Windows [Version 10.0.22621.2861]</code>, and that's the only difference I can see between our 2 environments.
I inserted <code>printf("%s\n%x\n", __MINGW64_VERSION_STR, _WIN32_WINNT);</code> into the script, in case either of those might be of some relevance.
It showed that _WIN32_WINNT is set to 0x0601, and the mingw runtime is 11.0.0.

jkeenan:
@khwilliamson @sisyphus @tonycoz : Can we get an update on the status of this ticket?

https://github.com/Perl/perl5/issues/21697 -- which is referred to in this ticket's Subject line and commit message -- was merged Dec 7 2023, but the discussion in this ticket carried onward.  If this ticket is *not* yet closable, we should at least provide a better subject line for it.  Thanks.

jkeenan:
> @khwilliamson @sisyphus @tonycoz : Can we get an update on the status of this ticket?
> 
> #21697 -- which is referred to in this ticket's Subject line and commit message -- was merged Dec 7 2023, but the discussion in this ticket carried onward. If this ticket is _not_ yet closable, we should at least provide a better subject line for it. Thanks.

Having heard no reason to keep this ticket open, I'm closing it now.  Please open a **new** ticket if you have further problems with this test file.

Thank you very much.

PR #21702, opened 2023-12-08
locale.c: is_locale_utf8(): C locale cant be UTF-8

khwilliamson:
This adds a quick check to rule out "C" and "POSIX" as being possible UTF-8 locales.  This shortcuts the more expensive checking needed for non-trivial cases.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -4551,6 +4551,10 @@ S_is_locale_utf8(pTHX_ const char * locale)
         return PL_in_utf8_CTYPE_locale;
     }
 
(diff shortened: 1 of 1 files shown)


jkeenan:
> This adds a quick check to rule out "C" and "POSIX" as being possible UTF-8 locales. This shortcuts the more expensive checking needed for non-trivial cases.

This could use a test.

khwilliamson:
There is no change in external behavior; only a shortcut from the existing code which arrives at the same answer more lugubriously.  Hence the change is effectively not testable.

tonycoz (approved):

PR #21704, opened 2023-12-08
Simplify finding if a locale is UTF-8

khwilliamson:
This series of commits improves the handling of determining whether or not a locale is UTF-8.  Comments are added, as well as a shortcut that works in some circumstances to avoid work.  But the biggest change is providing a cleaner implementation of two functions, moving some code from one that really belongs in the other function, to that other function

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -4580,8 +4580,14 @@ S_is_locale_utf8(pTHX_ const char * locale)
 
 #      else
 
-        /* Without those functions, we use our emulation of nl_langinfo(),
-         * which works quite well, but isn't fully definitive */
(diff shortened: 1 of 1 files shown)


jkeenan:
> Often in perl, we just want to know if a locale is UTF-8 or not; we don't care otherwise if it is CP1252 or whatever.
> 

Could we get some tests that demonstrate this?

khwilliamson:
I meant often in locale.c.

khwilliamson:
commit and summary messages have now been cleaned up

tonycoz (approved):

PR #21705, opened 2023-12-08
locale.c: Quit now if strftime() fails with EINVAL

khwilliamson:
It turns out that some platforms set errno to EINVAL when the libc function strftime() finds the input format to be in error, though this is not specified in the Standard (a defect there in my opinion). Windows in particular does set errno.  Quit when an EINVAL errno is detected.  On platforms that don't set EINVAL, the current behavior is retained where the loop is executed again and again to rule out that the error isn't a buffer size problem.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6846,7 +6846,7 @@ value is a pointer to the formatted result (which MUST be arranged to be FREED
 BY THE CALLER).  This allows these functions to increase the buffer size as
 needed, so that the caller doesn't have to worry about that.
 
-On failure they return NULL, and set errno to C<EINVAL>.
+On failure, they return NULL, and set C<errno> to C<EINVAL>.
 
 C<sv_strftime_tm> and C<sv_strftime_ints> are preferred, as they transparently
 handle the UTF-8ness of the current locale, the input C<fmt>, and the returned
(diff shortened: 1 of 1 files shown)


tonycoz:
This should save errno (`dSAVE_ERRNO`) and restore it on the success path.

This change is a bit risky on non-Windows (which documents the EINVAL behaviour you're using): the C standard allows any library function that isn't documented to use errno to set errno to a non-zero value whether or not it succeeds.

So a conforming implementation of strftime() might set errno to EINVAL even if the problem isn't the format string.

khwilliamson:
I wondered if this would be too risky

khwilliamson:
So what to do?  I can withdraw this request, or ifdef Windows

jkeenan:
> So what to do? I can withdraw this request, or ifdef Windows

*My two cents from an administrative/process point of view:*

I cannot speak to the specifics of this pull request or to Tony's concerns.  However I do note that we have a lot of locales-related changes in process right now, the latest of which (https://github.com/Perl/perl5/pull/21714) involves 17 individual commits.  That's a lot to review, and we don't have a large number of people qualified to conduct such reviews.

If this particular p.r. is not really essential, I'd suggest withdrawing it for the time being so that we can focus on others that you would consider both more important and less problematic.

tonycoz:
I think it's reasonable to make it use `#ifdef WIN32`.

I look at POSIX, but unfortunately that doesn't specify that strftime() sets errno, and the fallback as described for errno is roughly the same as the C standard.

jkeenan:
@tonycoz, @khwilliamson has forced-pushed some changes.  Could you re-review?  Thanks.

tonycoz (approved):

PR #21714, opened 2023-12-12
Revamp localeconv() handling

khwilliamson:
localeconv() is problematic.  It isn't thread-safe on any platform except modern Windows.  It returns lots of values, and from two different locale categories (which don't have to be set to the same locale), and it is buggy on Windows. And, it's used a lot in Windows, to find what the decimal point character is.

Thus there's lots of compensation locale.c does to accommodate this, and for the variances in Configurations.  This series of commits begins the process of simplifying it and speeding it up in some cases.

Often, the locale is C.  When this is the case, all the values are the same with one exception.  And those values are documented by the Standards, so are known at compile time, and thus no mutex locking is needed to make it thread-safe.  One commit introduces a stripped down function to use when the locale is C instead of the more general one.

Other commits simplify the code, in part by making greater use of tables.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4415,12 +4415,12 @@ Sf	|char * |strftime_tm	|NN const char *fmt			\
 				|NN const struct tm *mytm
 # if defined(HAS_LOCALECONV)
 S	|HV *	|my_localeconv	|const int item
-S	|void	|populate_hash_from_localeconv					\
-				|NN HV *hv					\
-				|NN const char *locale				\
-				|const U32 which_mask				\
-				|NN const lconv_offset_t *strings[2]		\
-				|NULLOK const lconv_offset_t *integers[2]
+S	|void	|populate_hash_from_C_localeconv			\
+				|NN HV *hv				\
+				|NN const char *locale			\
+				|const U32 which_mask			\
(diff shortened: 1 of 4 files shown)


tonycoz (approved):

tonycoz:
I'll admit with all the workarounds we do for various buggy implementations (and the performance implications of those workaround), whether we'd be better off with our own locale implementation.  Floating point conversions make that harder though.

khwilliamson:
I don't think it is feasible for our own implementation.  If you think floating point is bad, consider collation.  We go to great lengths to use strxfrm() for that, kludging in a not-totally satisfactory way of handling embedded-NULs, which strxfrm(), doesn't deal with.  Similarly we would have to be able to read and parse the locale definitions on each platform.  I doubt that these are fully Standardized, much less their positions in the file system, or even their accessibility.

tonycoz:
My initial original thought for this was providing our own locale definitions, but that causes problems with systems that define locales we don't.  Then I thought we could generate definitions from the system definitions at build time, but that leads to copyright issues for Strawberry perl and other binary distributions (and I forgot about collations).

Then I thought we could extract them at runtime, but again I'd forgotten about collations.

My dreams of locale sanity, shattered.

PR #21719, opened 2023-12-14
 locale.c: Avoid some locale switching

khwilliamson:
In our emulation of nl_langinfo(), prior to this commit, we always
toggled the locales of two categories to the necessary ones.  But it
turns out that most paths through the code don't require this toggling.
This commit narrows the scope of the toggling to just those cases that
need it.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6091,17 +6091,6 @@ S_my_langinfo_i(pTHX_
      * but they are not exposed.  Also calling setlocale(), then calling
      * GetThreadLocale() doesn't work, as the former doesn't change the
      * latter's return.  Therefore we are stuck using the mechanisms below. */
-
-#    ifdef WE_MUST_DEAL_WITH_MISMATCHED_CTYPE
-
-    const char * orig_CTYPE_locale = toggle_locale_c(LC_CTYPE, locale);
-
-#    endif
-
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -6105,68 +6105,66 @@ S_my_langinfo_i(pTHX_
 #    if      defined(HAS_SNPRINTF)                                          \
        && (! defined(HAS_LOCALECONV) || defined(TS_W32_BROKEN_LOCALECONV))
 
-        {
```
tonycoz: I expect removing the braces here would break C++ builds, assuming both this case and the following case can be compiled.

Removing the braces changes the scope of the declarations in the block from being limited to the block to include past the next case, which allows the initializations (like for floatbuf) to be bypassed, which isn't valid in C++.

tonycoz: I wasn't able to find a case where both blocks are compiled on Linux, but they don't seem to be entirely excluded.

tonycoz:
The second commit looks fine.

locale.c
khwilliamson: Sorry, I had already figured this out, but forgot I already had an open pull request.  And, it turns out that there was another place that needed the toggling.  PR now updated

tonycoz (approved):

PR #21720, opened 2023-12-14
locale.c: Move code to a separate function

khwilliamson:
Preprocessor directives will likely cause this code to be skipped and not compiled on modern libc's.  It is a fallback to when there is no mbtowc() nor mbrtowc() available on the platform, both C99 functions.

But, we've had too many examples of buggy libc functions, varying by platform, and so this is being kept for the time being so that perl will still work on a platform where these two functions are found to be buggy, and Configure is overridden to forbid their use.

Moving the code to a separate function isolates it from the mainline code, making that more understandable, and makes it easier to remove this code if we decide to.

Using the -b or -w options to diff will make the deltas in this commit more understandable, as the code was outdented in the new function.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -4567,6 +4567,8 @@ S_is_locale_utf8(pTHX_ const char * locale)
 
     return FALSE;
 
+     /* Definitively, can't be UTF-8 */
+#    define HAS_DEFINITIVE_UTF8NESS_DETERMINATION
 #  else
 
     /* If the input happens to be the same locale as we are currently setup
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -6513,305 +6525,310 @@ S_my_langinfo_i(pTHX_
...
 
-#          endif
-        };
+    /* For this portion of the file to compile, some C99 functions aren't
+     * available to us, even though we now require C99.  So, something must be
+     * wrong with them.  The code here should be good enough to work around
+     * this issue, but should the need arise, comments in S_is_locale_utf8()
+     * list some alternative C99 functions that could be tried.
+     *
+     * But MB_CUR_MAX is a C99 construct that helps a lot, is simple for a
```
tonycoz, line 6595: `MB_CUR_MAX` is C89

tonycoz:
This looks fine, but all the indentation and comment re-wrapping made this difficult to review.

tonycoz (approved):

khwilliamson:
Just moving it without changing the white space caused diff to spew out many, many misleading differences.  So I went ahead and did the white space changes, and in the commit message recommended using the diff options `-w` or `-b`; either of which stopped the misleading differences.

locale.c
khwilliamson: And so it is, even though it isn't in K&R v2, based on draft C89.  And the Linux man pages say C99.  Changed via a93c9128eb1306fc15c7b25040919300ce889d03.

I see now that `mbtowc()` is also C89.  Again the man page says C99 and it is missing from K&R.

PR #21722, opened 2023-12-14
Use reentrant API when MULTIPLICITY

khwilliamson:
Prior to this commit it was only done when using threads.  You can have MULTIPLICITY without threads.  And when you do, there is shared global data that can get clobbered, and which the reentrant API automagically greatly reduces the possibility of.

I claim that the prior situation indicates that it is pretty rare in practice to have MULTIPLICITY defined without threads.  Otherwise, we would be getting tickets about race conditions.  I believe @leont thinks that many programs would not use the affected libc calls, so this wouldn't be occurring much in the field.


Leont:
>  I believe @Leont thinks that many programs would not use the affected libc calls, so this wouldn't be occurring much in the field.

I suspect is usually overkill, but it's harmless overkill.

Leont (approved):

PR #21723, opened 2023-12-14
Fallback to use BitScan(Forward|Reverse) on MingW

khwilliamson:
Perl has functions that return the positions of the least or most significant one bit in a word.  gcc has builtins that do these specific operations, and they get compiled to a single instruction on many hardware platforms.  Windows instead has various BitScan functions that do the same thing.

MingW has both gcc and Windows, so it should have access to both the builtins and to the BitScan functions, but prior to this commit, if we didn't find gcc builtins on MingW, we wouldn't bother to look for the BitScan functions.

Reading the code caused me to pause to try to figure out why not use these BitScan functions, and asked on #irc.  The answer from @xenu was that on MingW (as opposed to MSVC) they're emulated, and not as efficient as the gcc builtins.

Then I looked at the code again, and realized we don't try them unless the gcc builtins aren't available, unlikely these days.  But we should do something so that people who read this code in the future don't also end up spending time unnecessarily wondering why we treat MingW specially.

We could add a comment, but it seems to me better to just not make MingW a special case.  It's likely that the MingW emulation of the BitScan functions is better than our further fallback, so I see that as better and cleaner than special casing MingW.

### DIFF
--- a/inline.h
+++ b/inline.h
@@ -1401,7 +1401,7 @@ Perl_is_utf8_invariant_string_loc(const U8* const s, STRLEN len, const U8 ** ep)
 #  endif
 #endif
 
-#if defined(_MSC_VER)
+#if defined(WIN32)
 #  include <intrin.h>
 #  pragma intrinsic(_BitScanForward)
 #  pragma intrinsic(_BitScanReverse)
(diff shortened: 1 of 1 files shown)


tonycoz:
They compile down to [single instructions](https://godbolt.org/z/7fsTjoTGz) [godbolt.org], which seems like decent emulation:
```
#include <intrin.h>

unsigned long bsf(unsigned long mask) {
    unsigned long pos;
    _BitScanForward(&pos, mask);

    return pos;
}

unsigned long bsf64(unsigned __int64 mask) {
    unsigned long pos;
    _BitScanForward64(&pos, mask);

    return pos;
}
```

Produces:
```
bsf:
/APP
        bsfl %ecx,%eax 
/NO_APP
        ret
bsf64:
/APP
        bsfq %rcx,%rax 
/NO_APP
        ret
```

tonycoz (approved):

PR #21726, opened 2023-12-15
hints/openbsd: Fix up some locale related hints

khwilliamson:
These two commits tell perl that the OpenBSD locale system is very restricted, and so the code to handle more general situations need not be compiled


jkeenan:
@fresh1, can you take a look at this OpenBSD-related p.r.?  Thanks.

tonycoz (approved):

PR #21728, opened 2023-12-17
`perllol`: Remove indirect object syntax in `Dumpvalue` example

rwp0:
Fix `dumpValue` method call syntax as well.

Change the variable name to `$prettify`
because the adverb doesn't make sense
without the indirect syntax.

The documentation shows correct examples:

https://perldoc.perl.org/Dumpvalue

### DIFF
--- a/pod/perllol.pod
+++ b/pod/perllol.pod
@@ -227,7 +227,7 @@ Hmm... that's still a bit ugly.  How about this:
 	}
     }
 
-When you get tired of writing a custom print for your data structures,
+When you get tired of writing a custom C<print> for your data structures,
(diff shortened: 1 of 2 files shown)


mauke:
I'd suggest rewording the commit message as "perllol: remove indirect object syntax in Dumpvalue example" or similar.

rwp0:
> I'd suggest rewording the commit message as "perllol: remove indirect object syntax in Dumpvalue example" or similar.

Done, though I'd prefer shorter commit messages.

tonycoz (approved):

rwp0:
@mauke can we merge it now (after the change and an approval) ?

PR #21729, opened 2023-12-18
return doesn't allow an indirect object

### LINKED ISSUE #21716: Return with a bareword compiles under strict
**Description**

`return sum map { ... } @list` seems to compile OK under strict without the import of `List::Util::sum` and returns a filehandle and a mapped list.

**Steps to Reproduce**
This program shows the behavior:
```
use warnings;
use strict;

use Data::Dumper;
# use List::Util qw{ sum };

sub xx { return sum map { $_ + 1 } 1 .. 5 }
print Dumper(xx);
```

The actual error occurs only after you remove `return`. Removing `map` causes it to report a warning, but it still runs.

**Expected behavior**

I expected to get an error about invalid bareword under strict mode, same as without `return`.

**Perl configuration**
Behavior is consistent on both perl 5.10.1 and 5.38.2.

tonycoz:
Since return isn't actually a function, I didn't think the "function" part of the original message applied.

Fixes #21716

### DIFF
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -4003,6 +4003,12 @@ follow the C<\N>.
 (F) While certain functions allow you to specify a filehandle or an
 "indirect object" before the argument list, this ain't one of them.
 
+=item Missing comma after first argument to return
+
+(F) While certain operators allow you to specify a filehandle or an
+"indirect object" before the argument list, C<return> isn't one of
+them.
+
 =item Missing command in piped open
 
 (W pipe) You used the C<open(FH, "| command")> or
--- a/t/lib/croak/op
+++ b/t/lib/croak/op
@@ -300,3 +300,11 @@ LOOP: {
 }
 EXPECT
 Can't "last" out of a "finally" block at - line 4.
+########
+# NAME return HANDLE LIST isn't valid [github #21716]
+sub xx {
+  return sum map { $_+1} 1 .. 5;
+}
+EXPECT
+Missing comma after first argument to return at - line 2, near "5;"
+Execution of - aborted due to compilation errors.
--- a/op.c
+++ b/op.c
@@ -13654,6 +13654,13 @@ Perl_ck_return(pTHX_ OP *o)
 
     PERL_ARGS_ASSERT_CK_RETURN;
 
+    if (o->op_flags & OPf_STACKED) {
+        kid = cUNOPx(OpSIBLING(cLISTOPo->op_first))->op_first;
+        if (kid->op_type != OP_SCOPE && kid->op_type != OP_LEAVE)
+            yyerror("Missing comma after first argument to return");
+        o->op_flags &= ~OPf_STACKED;
+    }
+
     kid = OpSIBLING(cLISTOPo->op_first);
     if (PL_compcv && CvLVALUE(PL_compcv)) {
         for (; kid; kid = OpSIBLING(kid))
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -73,7 +73,25 @@ XXX For a release on a stable branch, this section aspires to be:
     If any exist, they are bugs, and we request that you submit a
     report.  See L</Reporting Bugs> below.
 
-[ List each incompatible change as a =head2 entry ]
+=head2 C<return> no longer allows an indirect object
+
+The C<return> operator syntax now rejects indirect objects, in most
+cases this would compile and even run, but wasn't documented and could
+produce confusing results, for example:
+
+  # note that sum hasn't been defined
+  sub sum_positive {
+    return sum grep $_ > 0, @_;
+  }
+  say for sum_positive(-1, 2, 3)
+
+produced:
+
+  *main::sum
+  2
+  3
+
+[github #21716]
 
 =head1 Deprecations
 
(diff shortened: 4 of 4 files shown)


pod/perldiag.pod
```diff
@@ -4003,6 +4003,11 @@ follow the C<\N>.
 (F) While certain functions allow you to specify a filehandle or an
 "indirect object" before the argument list, this ain't one of them.
 
+=item Missing comma after first argument to return
+
+(F) While certain operators allow you to specify a filehandle or an
+"indirect object" before the argument list, return isn't one of them.
```
leonerd: `C<return>`

t/lib/croak/op
```diff
@@ -300,3 +300,11 @@ LOOP: {
 }
 EXPECT
 Can't "last" out of a "finally" block at - line 4.
+########
+# NAME return HANDLE LIST isn't valid
```
leonerd: Is it worth mentioning https://github.com/Perl/perl5/issues/21716 in the comment?

leonerd (changes requested):
As this change makes a few things become compiletime errors that previously had been either runtime errors or potentially fine if subs were defined late, it's probably worth at least a mention in the perldelta so if people find code breaks they know why.

pod/perldiag.pod
tonycoz: Added

t/lib/croak/op
tonycoz: Done (potentially saves someone from code archaeology)

tonycoz:
> As this change makes a few things become compiletime errors that previously had been either runtime errors or potentially fine if subs were defined late, it's probably worth at least a mention in the perldelta so if people find code breaks they know why.

If the sub is defined late the indirect object has already been compiled into an `rv2gv` op:
```
# system perl without this patch
$ perl -MO=Concise,xx -e 'sub xx { return sum map { $_ } 1..5 } sub sum(@); '
main::xx:
a  <1> leavesub[1 ref] K/REFC,1 ->(end)
-     <@> lineseq KP ->a
1        <;> nextstate(main 2 -e:1) v ->2
-        <@> return KS ->-
-           <0> pushmark s ->2
3           <1> rv2gv sKR/1 ->4
2              <#> gv[*sum] s ->3
8           <|> mapwhile(other->9)[t4] K ->a
7              <@> mapstart K ->8
4                 <0> pushmark s ->5
-                 <1> null lK/1 ->5
-                    <1> null lK/1 ->8
-                       <@> scope lK ->-
-                          <;> ex-nextstate(main 3 -e:1) v ->9
-                          <1> ex-rv2sv sK/1 ->-
9                             <#> gvsv[*_] s ->8
6                 <1> rv2av lKPM/1 ->7
5                    <$> const[AV ARRAY] s ->6
-e syntax OK
```
It even runs and returns a GV:
```
# system perl without this patch
$ perl -MDevel::Peek -e 'sub xx { return sum map { $_ } 1..5 } sub sum(@); Dump([xx()])'
SV = IV(0x561df481e1a8) at 0x561df481e1b8
  REFCNT = 1
  FLAGS = (TEMP,ROK)
  RV = 0x561df481d750
  SV = PVAV(0x561df47f2f50) at 0x561df481d750
    REFCNT = 1
    FLAGS = ()
    ARRAY = 0x561df482da10
    FILL = 5
    MAX = 5
    FLAGS = (REAL)
    Elt No. 0
    SV = PVGV(0x561df48497b0) at 0x561df481d738
      REFCNT = 1
      FLAGS = (FAKE,MULTI)
      NAME = "sum"
      NAMELEN = 3
      GvSTASH = 0x561df47f1518  "main"
      FLAGS = 0x2
      GP = 0x561df482e180
        SV = 0x0
        REFCNT = 3
        IO = 0x0
        FORM = 0x0  
        AV = 0x0
        HV = 0x0
        CV = 0x561df481e3e0
        CVGEN = 0x0
        GPFLAGS = 0x0 ()
        LINE = 1
        FILE = "-e"
        EGV = 0x561df481e0c8    "sum"
    Elt No. 1
    SV = IV(0x561df481d7d0) at 0x561df481d7e0
      REFCNT = 1
      FLAGS = (IOK,pIOK)
      IV = 1
    Elt No. 2
    SV = IV(0x561df481d7e8) at 0x561df481d7f8
      REFCNT = 1
      FLAGS = (IOK,pIOK)
      IV = 2
    Elt No. 3
    SV = IV(0x561df481d878) at 0x561df481d888
      REFCNT = 1
      FLAGS = (IOK,pIOK)
      IV = 3
```
So perhaps it should be a deprecation instead.

That would make me sad for such obviously broken code, since it seems more likely that the intent is to call a function rather than return a glob.

tonycoz:
> That would make me sad for such obviously broken code, since it seems more likely that the intent is to call a function rather than return a glob.

which my perldelta illustrates.

leonerd (approved):
LGTM

I think this looks fine, though we should keep an eye out for BBC reports in case it turns out there's modules around that break on it - or even that actually rely on it to work.

tonycoz:
I also considered a parser approach to fixing this (splitting LSTOP into list ops that take an indirob and those that don't), but that's more complex and riskier.

PR #21731, opened 2023-12-18
Add `av_unshift_sv()`

leonerd:
A new API function to handle this case, avoiding the need to unshift + assign separately.

A not-very-optimal implementation at present. A better one could be provided perhaps by merging the logic into `Perl_av_unshift` itself, but that is left for a later exercise.

### DIFF
--- a/av.c
+++ b/av.c
@@ -923,6 +923,24 @@ Perl_av_unshift(pTHX_ AV *av, SSize_t num)
 }
 
 /*
+=for apidoc av_unshift_sv
+
+Unshift a single SV onto the beginning of the array.  The array will grow
+automatically to accommodate the addition.
+
+=cut
+*/
+
+void
+Perl_av_unshift_sv(pTHX_ AV *av, SV *sv)
+{
+    PERL_ARGS_ASSERT_AV_UNSHIFT_SV;
+
+    av_unshift(av, 1);
+    av_store(av, 0, sv);
+}
+
+/*
 =for apidoc av_shift
 
 Removes one SV from the start of the array, reducing its size by one and
--- a/embed.fnc
+++ b/embed.fnc
@@ -708,6 +708,8 @@ ARdm	|SSize_t|av_top_index	|NN AV *av
 Adp	|void	|av_undef	|NN AV *av
 Adp	|void	|av_unshift	|NN AV *av				\
 				|SSize_t num
+Adp	|void	|av_unshift_sv	|NN AV *av				\
+				|NN SV *sv
 : Used in perly.y
 Rp	|OP *	|bind_match	|I32 type				\
 				|NN OP *left				\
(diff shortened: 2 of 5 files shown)


tonycoz (approved):

Leont:
My intuition is that should should also have a `av_unshift_svs` that takes an array of `SV*`,

tonycoz:
Thinking about it, av_store()'s return value is significant - and in this case that can be a problem if the av is tied

For normal arrays av takes ownership of a reference count of the supplied val, but when av is tied av_store() returns NULL and the caller is responsible for that reference count, and this implementation doesn't handle that, and doesn't allow the caller to handle that.

Perhaps it should:
```
assert(mg_find((const SV *)av, PERL_MAGIC_tied) == NULL);
```
Similarly for the proposed `av_unshift_svs()`  -- which I don't think needs to be part of this ticket.

jkeenan:
> A new API function to handle this case, avoiding the need to unshift + assign separately.
> 
> A not-very-optimal implementation at present. A better one could be provided perhaps by merging the logic into `Perl_av_unshift` itself, but that is left for a later exercise.

@leonerd, both @Leont and @tonycoz have raised questions about this p.r.  Do you wish to pursue it further?  Thanks.

jkeenan:
> > A new API function to handle this case, avoiding the need to unshift + assign separately.
> > A not-very-optimal implementation at present. A better one could be provided perhaps by merging the logic into `Perl_av_unshift` itself, but that is left for a later exercise.
> 
> @leonerd, both @Leont and @tonycoz have raised questions about this p.r. Do you wish to pursue it further? Thanks.

@leonerd, ^^

jkeenan:
> A new API function to handle this case, avoiding the need to unshift + assign separately.
> 
> A not-very-optimal implementation at present. A better one could be provided perhaps by merging the logic into `Perl_av_unshift` itself, but that is left for a later exercise.

This p.r. got an approval from one reviewer nearly a year ago -- but second thoughts soon emerged.  There has been little substantive discussion in nearly twelve months.  Consequently, the p.r. has not been merged into blead.

My sense is that we're not enthusiastic enough about this to put it into blead and that the ticket should be closed.  @leonerd, @Leont, @tonycoz ?

jkeenan:
> > A new API function to handle this case, avoiding the need to unshift + assign separately.
> > A not-very-optimal implementation at present. A better one could be provided perhaps by merging the logic into `Perl_av_unshift` itself, but that is left for a later exercise.
> 
> This p.r. got an approval from one reviewer nearly a year ago -- but second thoughts soon emerged. There has been little substantive discussion in nearly twelve months. Consequently, the p.r. has not been merged into blead.
> 
> My sense is that we're not enthusiastic enough about this to put it into blead and that the ticket should be closed. @leonerd, @Leont, @tonycoz ?

No one has spoken up for keeping this p.r. active, so I'm closing it now.  Please provide a rationale for re-opening if that is someone's wish.  Thanks.

PR #21735, opened 2023-12-20
locale.c: Avoid unnecessary malloc

khwilliamson:
Here an SV is being created with a size, but immediately sv_usepvn_flags() is called to place an already allocated and populated PV into it, so that initial size is wasted.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5969,9 +5969,8 @@ Perl_sv_strftime_tm(pTHX_ SV * fmt, const struct tm * mytm)
                              );
     SV * sv = NULL;
     if (retval) {
-        STRLEN len = strlen(retval);
-        sv = newSV(len);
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -5969,9 +5969,8 @@ Perl_sv_strftime_tm(pTHX_ SV * fmt, const struct tm * mytm)
                              );
     SV * sv = NULL;
     if (retval) {
-        STRLEN len = strlen(retval);
-        sv = newSV(len);
-        sv_usepvn_flags(sv, retval, len, SV_HAS_TRAILING_NUL);
+        sv = newSV(0);
```
richardleach: `Perl_sv_usepvn_flags` is going to immediately do `SvUPGRADE(sv, SVt_PV)`, calling `Perl_sv_upgrade`. Would be more efficient still to assign the SV using `sv = newSV_type(SVt_PV);`

Otherwise, LGTM.

Issue #21732, opened 2023-12-19
[doc] if Configure'd to "Use Strict by default", 5.38.2 build fails because assorted build scripts break under "use strict"


gsteemso:
At a minimum, offending scripts appear to include "regen/embed_lib.pl" (as called by "autodoc.pl") and "base/cond.t" (as called by minitest).  Each of them failed hard with a slew of "did you forget to declare variable $such-and-such?" messages.

If it turns out that these do in fact run successfully with "use strict", I will be very confused, because disabling that option in Configure allowed compilation and installation to succeed.

jkeenan:
> At a minimum, offending scripts appear to include "regen/embed_lib.pl" (as called by "autodoc.pl") and "base/cond.t" (as called by minitest). Each of them failed hard with a slew of "did you forget to declare variable $such-and-such?" messages.
> 
> If it turns out that these do in fact run successfully with "use strict", I will be very confused, because disabling that option in Configure allowed compilation and installation to succeed.

Can you post the `./Configure` command you used in this investigation?

tonycoz:
```
$ ./Configure -des -Dusedevel -Dusedefaultstrict
...
$ make -j2 test-prep
...
$ ./miniperl -Ilib autodoc.pl
...
Global symbol "$n" requires explicit package name (did you forget to declare "my $n"?) at ./regen/embed_lib.pl line 3.
Global symbol "$n" requires explicit package name (did you forget to declare "my $n"?) at ./regen/embed_lib.pl line 4.
Global symbol "$n" requires explicit package name (did you forget to declare "my $n"?) at ./regen/embed_lib.pl line 5.
Global symbol "$n" requires explicit package name (did you forget to declare "my $n"?) at ./regen/embed_lib.pl line 6.
Global symbol "$n" requires explicit package name (did you forget to declare "my $n"?) at ./regen/embed_lib.pl line 7.
BEGIN not safe after errors--compilation aborted at ./regen/embed_lib.pl line 10.
Compilation failed in require at autodoc.pl line 89.
```

jkeenan:
> At a minimum, offending scripts appear to include "regen/embed_lib.pl" (as called by "autodoc.pl") and "base/cond.t" (as called by minitest). Each of them failed hard with a slew of "did you forget to declare variable $such-and-such?" messages.
> 
> If it turns out that these do in fact run successfully with "use strict", I will be very confused, because disabling that option in Configure allowed compilation and installation to succeed.

[...]

PR #21737, opened 2023-12-20
allow some basic infrastructure to load with -Dusedefaultstrict

tonycoz:
The changes to t/test.pl appear to fix real bugs.

This allows `make test_harness` to run, but many tests will still fail under -Dusedefaultstrict

This addresses #21732 but does not fix it.  I'm unsure how supported that build option is.

### DIFF
--- a/TestInit.pm
+++ b/TestInit.pm
@@ -37,7 +37,7 @@
 
 package TestInit;
 
-$VERSION = 1.04;
+our $VERSION = 1.05;
 
 # Let tests know they're running in the perl core.  Useful for modules
 # which live dual lives on CPAN.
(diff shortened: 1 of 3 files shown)


Leont (approved):

jkeenan:
I'm okay with these changes for the purpose of having more of our infrastructure being "strict-compliant" in the sense of PASSing under `./perl -Ilib -Mstrict -c %`.  But I don't think we should tie that objective too closely to *have perl build/test under `-Dusedefaultstrict`*.  So while I'm okay with the code and the commit message, I'm uneasy with the Subject of this p.r.

rwp0 (approved):

PR #21739, opened 2023-12-20
locale.c: Move three functions within file

khwilliamson:
These are moved to where the other functions operating on the same thing are.


jkeenan:
It looks like you're actually repositioning **3** functions within this file -- correct?  If so, can you change the Subject line?  Thanks.

tonycoz:
The commit message needs a similar update.

Otherwise it's fine.

PR #21740, opened 2023-12-21
S_my_langinfo_i: Always return both a buffer and a value

khwilliamson:
This function returns a value, and also copies that value into a buffer. Certain rare Configurations failed to fill in the buffer in some circumstances.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6060,6 +6060,7 @@ S_my_langinfo_i(pTHX_
      * override if necessary */
     utf8ness_t is_utf8 = UTF8NESS_IMMATERIAL;
     const char * retval = NULL;
+    bool retval_saved = false;
 
     switch (item) {
       default:
@@ -6140,6 +6141,7 @@ S_my_langinfo_i(pTHX_
             if (LIKELY(s < e)) {
                 *s = '\0';
                 retval = save_to_buffer(item_start, retbufp, retbuf_sizep);
+                retval_saved = true;
                 Safefree(floatbuf);
 
                 if (utf8ness) {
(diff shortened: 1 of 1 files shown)


tonycoz:
It's difficult to tell from the description, but does the interface of the function demand that the buffer be filled, as opposed to the buffer being an optional scratch space>

If it is required, this seems more like a workaround for the cases where save_to_buffer() should be called but isn't.

If it isn't, then the cases where retval is a string literal this seems like a waste.

khwilliamson:
`S_my_langinfo_i()` is designed to present a uniform, care-free interface to the libc function `nl_langinfo`.  That means it returns a stable value that is valid until its next call, and that was generated without interfering with any other running thread or perl instance in embedded applications.  `nl_langinfo` only furnishes these qualities on unthreaded builds.

Further, Windows doesn't even have `nl_langinfo`, so in that case, this hides that fact from the higher level code, so it all looks from that point of view that we have a well-behaved implementation, no matter the underlying platform or Configuration.

I think there's likely to be inefficiencies as a result.

This function is reentrant, so a a buffer parameter is required; that might be going away later in 5.39.  A typical call to `nl_langinfo` under threads must be done in a critical section, and the result copied before releasing the mutex.  Again a buffer is required.

Literals are only returned in minority platforms (such as OpenBSD) or Configurations, so they should be less of a concern.

I have been planning to look at this interface (since it is internal, it can be changed) after things settled down, with an eye to improving the efficiency.  #21796 goes quite a ways towards that settling.  Ideas are welcome.

But in the meantime, this commit fixes some bugs on minority Configurations, and will hopefully become unnecessary later.

khwilliamson:
I'm working on an alternative idea

PR #21754, opened 2023-12-26
locale.c: Add clarifying comment

khwilliamson:
I wonder if it is too "cute"

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6149,6 +6149,21 @@ S_emulate_langinfo(pTHX_ const nl_item item,
 #  else
 
       case CRNCYSTR:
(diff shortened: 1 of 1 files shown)


jkeenan:
> I wonder if it is too "cute"

It is.   By the time I got to the end of the comment, I was baffled as to your point or how to verify it.

tonycoz:
A new paragraph at "The POSIX standard..." might make this easier to read, making the real point of the comment stand out.

I don't mind a bit of whimsy, but I've had others disagree.

tonycoz (approved):

PR #21756, opened 2023-12-26
cpan/Math-BigInt - Update to version 2.003001

book:
2.003001 2023-12-26

 * Add configuration methods trap_inf() and trap_nan(). Previously it was only possible to modify these properties by using the config() method.

 * Fix CPAN RT #150796 so that config() no longer modifies the input when it is a hash ref.

 * Fix CPAN RT #150797 so that both accuracy and precision can be set simultaneously with config().

 * Add the following methods to Math::BigInt

   - bilog2()   base 2 logarithm rounded downwards, i.e., int(log2(x))
   - bilog10()  base 10 logarithm rounded downwards, i.e., int(log10(x))
   - bclog2()   base 2 logarithm rounded upwards, i.e., ceil(log2(x))
   - bclog10()  base 10 logarithm rounded upwards, i.e., ceil(log10(x))

 * Add the following backend library methods to Math::BigInt::Lib. These methods do the core computations for the corresponding methods in Math::BigInt (see above).

   - _ilog2()   base 2 logarithm rounded downwards
   - _ilog10()  base 10 logarithm rounded downwards
   - _clog2()   base 2 logarithm rounded upwards
   - _clog10()  base 10 logarithm rounded upwards

### DIFF
--- a/Porting/Maintainers.pl
+++ b/Porting/Maintainers.pl
@@ -759,7 +759,8 @@ our %Modules = (
     },
 
     'Math::BigInt' => {
-        'DISTRIBUTION' => 'PJACKLAM/Math-BigInt-2.002001.tar.gz',
+        'DISTRIBUTION' => 'PJACKLAM/Math-BigInt-2.003001.tar.gz',
+        'SYNCINFO'     => 'book on Tue Dec 26 22:44:58 2023',
         'SYNCINFO'     => 'jkeenan on Mon Dec 11 21:10:38 2023',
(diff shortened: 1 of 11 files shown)


jkeenan:
Assuming the [CPANtesters results](http://fast-matrix.cpantesters.org/?dist=Math-BigInt%202.003001) for 2.003001 continue to be good for, say, the next 24 hours, I think this is good to merge.

book:
The [CPANtesters results](http://fast-matrix.cpantesters.org/?dist=Math-BigInt%202.003001) for 2.003001 are still green, so I'll merge.

PR #21760, opened 2023-12-29
Fix a wrong sigil used in a perlre example

mkende:
In the description of the @^CAPTURE array, an example of demarcated variable name uses the @ sigil. The $ sigil is probably more correct in that case and raises less questions when reading this section (although the @ sigil will give the same result in that case).


mauke:
Good change.
It looks like this is your first commit to perl. Would you like to add yourself to the `AUTHORS` file? (Running `perl Porting/updateAUTHORS.pl` should do it, although it might not know your full name.) That should fix the failing test.

mkende:
Thanks for the suggestion @mauke. I have done so (and also updated the commit to use my cpan email address).

mkende:
I’m not sure where is the other email address leaking from as the commit now has only my cpan address, but I also updated the `.mailmap` file as it seemed required by the check.

mauke (approved):

PR #21763, opened 2023-12-29
[POD] `diagnostics.pm`: Remove indirect syntax

rwp0:
Convert the below indirect syntax in documentation:

```
enable diagnostics;
disable diagnostics;
```

Since `enable` and `disable` are only methods

It's confusing to see barewords in the documentation.

As a user I thought `diagnostics` should actually be a string argument
to exported-by-default separate `enable` and `disable` subroutines.

### DIFF
--- a/lib/diagnostics.pm
+++ b/lib/diagnostics.pm
@@ -11,8 +11,8 @@ Using the C<diagnostics> pragma:
     use diagnostics;
     use diagnostics -verbose;
 
-    enable  diagnostics;
-    disable diagnostics;
+    diagnostics->enable;
+    diagnostics->disable;
 
 Using the C<splain> standalone filter program:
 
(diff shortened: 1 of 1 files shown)


mauke:
I like it, but you need to bump the version number of diagnostics.pm.

rwp0:
> I like it, but you need to bump the version number of diagnostics.pm.

Thanks, version bumped.

lib/diagnostics.pm
```diff
@@ -186,7 +186,7 @@ use 5.009001;
 use Carp;
 $Carp::Internal{__PACKAGE__.""}++;
 
-our $VERSION = '1.39';
+our $VERSION = '1.40';
```
rwp0, line 189: Here's the version bump per @mauke

rwp0:
@mauke checking the source code for `diagnostics.pm` I see that both `import` (`enable` is aliased to it) and `disable` subroutines `shift` the first argument to void.

- https://github.com/Perl/perl5/blob/29dc33f80fb2c20cbfe5828b67d1809e909db4a2/lib/diagnostics.pm#L504
- https://github.com/Perl/perl5/blob/29dc33f80fb2c20cbfe5828b67d1809e909db4a2/lib/diagnostics.pm#L548

Although the documentation mentions `enable` and `disable` as **methods**, these seem to be nothing more than regular functions, and there's no `bless` statement in the entire module to think of them as class methods rather than package functions.

Given that, I'm thinking maybe we change `->` to `::` sometime in the future removing void `shift`'s from the functions.

Seems to me like the entire "show" here is made to have `enable diagnostics` work 🙂

PR #21765, opened 2023-12-29
locale.c: Slightly change wording of DEBUG messages

khwilliamson:
This makes them more consistent with the other messages in this file.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -8686,10 +8686,10 @@ S_toggle_locale_i(pTHX_ const locale_category_index cat_index,
     const char * locale_to_restore_to = querylocale_i(cat_index);
 
     DEBUG_Lv(PerlIO_printf(Perl_debug_log,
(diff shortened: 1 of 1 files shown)


jkeenan:
> This makes them more consistent with the other messages in this file.

Can you provide an example of how (assuming I've built perl with `-DDEBUGGING`) I would do some work that would show me before/after contrasts for the debugging messages?

khwilliamson:
You could do 
'./perl -Ilib -T -DLv lib/locale.t`
with and without the patch, and compare the outputs

jkeenan:
> You could do './perl -Ilib -T -DLv lib/locale.t` with and without the patch, and compare the outputs

Be careful what you asked for, Jim; you get a million-plus-line diff!

jkeenan (approved):

khwilliamson:
Sorry I did not think about the size of the output.  What you could do is look at the beginning of the post-commit file until you get to one with the word "toggle", note the line number, and then look in the other file near that line number.

PR #21768, opened 2023-12-29
Remove references to nl_langinfo_l(), add comment

khwilliamson:
This was missed when we earlier removed the use of this function.  I have since run experiments using it, and found using it slows things down considerably in the context where we planned to use it.


jkeenan (approved):
LGTM (but IANACP).

PR #21772, opened 2023-12-29
locale.c: handle codesets GB18030, EUC-TW

khwilliamson:
These two codesets are multi-byte, like UTF-8.  Previously we sort of-of looked for them, but not fully.  This commit rectifies that.  If the code set matches one of those two, we don't create a bias towards the locale being UTF-8.

I discovered this by testing in an unusual Configuration.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6629,19 +6629,46 @@ S_override_codeset_if_utf8_found(pTHX_ const char * codeset,
 
     /* Here, the name doesn't indicate UTF-8, but MB_CUR_MAX indicates it could
      * be.  khw knows of only two other locales in the world, EUC-TW and GB
-     * 18030, that legitimately require this many bytes (4).  In both, the
-     * single byte characters are the same as ASCII.  No multi-byte character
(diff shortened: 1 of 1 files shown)


jkeenan:
> These two codesets are multi-byte, like UTF-8. Previously we sort of-of looked for them, but not fully. This commit rectifies that. If the code set matches one of those two, we don't create a bias towards the locale being UTF-8.
> 
> I discovered this by testing in an unusual Configuration.

Can you say what that unusual configuration was?

khwilliamson:
Add to Configure options
` -Ud_nl_langinfo -Ui_langinfo -Ud_nl_langinfo_l -Ud_thread_safe_nl_langinfo_l -Ud_mbtowc -Ud_mbrtowc -Accflags=-DNO_LOCALE_TIME -Accflags=-DNO_LOCALE_MONETARY `

But #21771 needs to be applied first if you want to reproduce this, as it fixes a bug.  No tests actually fail, but `lib/locale_threads.t` prints a warning about these

tonycoz (approved):

PR #21775, opened 2023-12-29
locale.c: Move declaration to first uses

khwilliamson:
C99 allows the clearer code of declaring a variable close to its use.

This also will be helpful in a future commit.


jkeenan (approved):
LGTM (but IANACP).

PR #21776, opened 2023-12-29
locale.c: C is the only locale under NO_LOCALE

khwilliamson:
It is possible to Configure perl to not pay attention to locales at all. Effectively that means the only permissible locale is "C", which underlies all C programs at startup.

Thus, when asked what the current locale is, the answer is always going to be "C"; and we can define the macro that computes this info to just return "C" instead of doing any lookup.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -2668,9 +2668,11 @@ S_bool_setlocale_2008_i(pTHX_
 #  define query_nominal_locale_i(i)                                         \
       (__ASSERT_(i != LC_ALL_INDEX_)                                        \
        ((i == LC_NUMERIC_INDEX_) ? PL_numeric_name : querylocale_i(i)))
-#else
+#elif defined(USE_LOCALE)
 #  define query_nominal_locale_i(i)                                         \
       (__ASSERT_(i != LC_ALL_INDEX_) querylocale_i(i))
+#else
+#  define query_nominal_locale_i(i)  "C"
 #endif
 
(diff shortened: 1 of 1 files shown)


jkeenan:
> It is possible to Configure perl to not pay attention to locales at all. Effectively that means the only permissible locale is "C", which underlies all C programs at startup.
> 
> Thus, when asked what the current locale is, the answer is always going to be "C"; and we can define the macro that computes this info to just return "C" instead of doing any lookup.

**blead**
```
$ sh ./Configure -des -Dusedevel -Accflags=-DNO_LOCALE && make test_prep
...
$ ./perl -Ilib -MPOSIX=setlocale -E 'say setlocale( LC_ALL );'
C

$ git clean -dfxq; sh ./Configure -des -Dusedevel && make test_prep
...
$ ./perl -Ilib -MPOSIX=setlocale -E 'say setlocale( LC_ALL );'
en_US.UTF-8
```
**branch**
```
$ git fetch origin pull/21776/head:gh-21776-khw-locale-20231229
$ git checkout gh-21776-khw-locale-20231229
$ git clean -dfx; sh ./Configure -des -Dusedevel -Accflags=-DNO_LOCALE && make test_prep
...
$ ./perl -Ilib -MPOSIX=setlocale -E 'say setlocale( LC_ALL );'
C
```
Branch DWIMs for me (Linux).

tonycoz (approved):

PR #21777, opened 2023-12-29
locale.c: toggling locales is a no-op if no locales

khwilliamson:
When the only legal locale is C, toggling to another locale doesn't make sense.  By #defining the macro that implements toggling this to do nothing in this case, we can avoid some #ifdefs


jkeenan:
> When the only legal locale is C, toggling to another locale doesn't make sense. By #defining the macro that implements toggling this to do nothing in this case, we can avoid some #ifdefs

Do we have any way of writing a unit test that exercises this code before/after?

khwilliamson:
No.  This doesn't change existing external behavior, but it reduces the need for ifdef's in future commits

tonycoz (approved):

Issue #21746, opened 2023-12-22
Strange interaction with locale and late loading Encode module


ailin-nemui:
Module:

**Description**

Trying to `use Encode` after `POSIX::setlocale` has been changed to Polish, causes the following error:
```
Constants from lexical variables potentially modified elsewhere are no longer permitted at constant.pm line 41.
BEGIN failed--compilation aborted at constant.pm line 45.
Compilation failed in require at Encode.pm line 7.
BEGIN failed--compilation aborted at Encode.pm line 7.
```

**Steps to Reproduce**

(in case you have any of that set)
```
unset LC_ALL LC_CTYPE LC_COLLATE LC_MESSAGES LC_MONETARY LC_NUMERIC LC_TIME LC_ADDRESS LC_IDENTIFICATION LC_MEASUREMENT LC_NAME LC_PAPER LC_TELEPHONE
```

```
LANG=pl_PL.UTF-8 perl -E'
   BEGIN{
      use POSIX;
      setlocale(LC_ALL,"")
   }
   use Encode
'
```

you can also get the same error with:
```
LANG=pl_PL.utf8 perl -MPOSIX -E'
   setlocale(LC_ALL,"");
   eval q{
      use constant X => $]
   };
   print $@
'
```

**Expected behavior**

`use Encode` should not crash depending on the locale

**Perl configuration**

```
Summary of my perl5 (revision 5 version 39 subversion 6) configuration:
  Snapshot of: c5f88297e0985be19f66e995b527ee4a001fc028
  Platform:
    osname=linux
    osvers=6.6.3-1-default
    archname=x86_64-linux-thread-multi
    uname='linux d5421s.localdomain 6.6.3-1-default #1 smp preempt_dynamic wed nov 29 05:06:07 utc 2023 (d766c57) x86_64 x86_64 x86_64 gnulinux '
    config_args='-de -Dprefix=~/perl5/perlbrew/perls/perl-blead -Duseshrplib -Dusethreads -Dusemultiplicity -Dusedevel -Aeval:scriptdir=~/perl5/perlbrew/perls/perl-blead/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
[...]

PR #21786, opened 2023-12-31
locale.c: Be sure to toggle into dot radix locale

khwilliamson:
This fixes GH #21746

Perl keeps the LC_NUMERIC category in a locale where the radix character is a dot, regardless of what the user has requested.  This is because much XS code has been written with the dot assumption.  When the user's actual radix character is desired, the locale is briefly toggled to that one for the duration of the operation.

When the user changes the LC_NUMERIC locale, the new one is noted, but the attempted change is otherwise ignored unless its radix is a dot. The new one will be briefly toggled into when appropriate.

The blamed commit contains a logic error

commit 818cdb7aa9f85227c1c7313257c6204c872beb94
AuthorDate: Sun Apr 11 05:57:07 2021 -0600
CommitDate: Thu Sep 1 09:02:04 2022 -0600

    locale.c: Skip code if will be a no-op

It decided it was a no-op if the new locale that the user is changing to is the same as the previous locale.  But it didn't consider that what actually happens is that the new locale does actually get changed, and this code is supposed to make sure that, before returning control to the user, that a dot radix locale is in effect.

If the new locale is a dot radix locale, then no harm is done by skipping the code, but otherwise things can go wrong.

I am chagrined that I made this logic error without noticing before it got pushed, and am surprised that it took this long for the error to surrface.  There must be something else intervening to make this not a problem in most circumstances, but I haven't analyzed what it might be.

The details as to why it happened in this test case are pretty obscure. The locale in effect is looking for a comma radix, but what is being checked for is a Perl version number, like 5.0936.  When converting that to a floating point number, the dot is not recognized, and only the initial '5' is found.  The failing code in a module has different actions depending on the current perl version it is being called from, and the conditional got the answer wrong because 5 is less than 5.0936, whereas the actual version is above that.  So it did the wrong thing and caused an error.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -3386,10 +3386,26 @@ S_new_numeric(pTHX_ const char *newnum, bool force)
                            "Called new_numeric with %s, PL_numeric_name=%s\n",
                            newnum, PL_numeric_name));
 
-    /* If not forcing this procedure, and there isn't actually a change from
-     * our records, do nothing.  (Our records can be wrong when sync'ing to the
-     * locale set up by an external library, hence the 'force' parameter) */
+    /* We keep records comparing the characteristics of the LC_NUMERIC catetory
+     * of the current locale vs the standard C locale.  If the new locale that
+     * has just been changed to is the same as the one our records are for,
+     * they are still valid, and we don't have to recalculate them.  'force' is
+     * true if the caller suspects that the records are out-of-date, so do go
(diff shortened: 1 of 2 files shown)


jkeenan (commented):
Test (in branch) passes for me where locale is `en_DK.utf8`.  Someone else should review the C changes and the comment.

tonycoz (approved):

PR #21788, opened 2024-01-02
Fix an error in `perlvar`: `@+` may have *more* elements than `@-`.

mkende:
The `@+` variable is documented with (emphasis mine):

> You can use this to determine how many capture buffers there are in the pattern. (**As opposed to C<@-> which may have fewer elements.**)

While `@-` had:

> The number of elements it contains will be one more than the number of the **highest capture buffers [...] that actually matched something**. (As opposed to C<@+> which may have fewer elements.)

So the parenthesis in the `@-` documentation is probably just a erroneous copy/paste.

### DIFF
--- a/pod/perlvar.pod
+++ b/pod/perlvar.pod
@@ -1358,7 +1358,7 @@ match and any capture buffers it contains.
 
 The number of elements it contains will be one more than the number of
 the highest capture buffers (also called a subgroup) that actually
-matched something. (As opposed to C<@+> which may have fewer elements.)
+matched something. (As opposed to C<@+> which may have more elements.)
 
 C<$-[0]> is the offset of the start of the last successful match.
 C<$-[I<n>]> is the offset of the start of the substring matched by


jkeenan:
Could you provide a little code which demonstrates the validity of this documentation change?  (As a comment here, not as further doc changes). Thanks.

demerphq:
```
perl -le'"a"=~/(a)?(b)?(c)?/; print for 0+@+, 0+@-;'
4
2
```

demerphq (approved):

mkende:
Thanks @demerphq.

@jkeenan, in addition to the example provided by @demerphq, the reasoning for this change can also come from the surrounding documentation: The `@+` variable is documented with (emphasis mine):

> You can use this to determine how many capture buffers there are in the pattern. (**As opposed to C<@-> which may have fewer elements.**)

While `@-` had:

> The number of elements it contains will be one more than the number of the **highest capture buffers [...] that actually matched something**. (As opposed to C<@+> which may have fewer elements.)

I have added this context in the PR description too, for future reference.

demerphq:
Thanks for the fixup, i recently added this verbiage, I appreciate you noticed the error and fixed it.

PR #21791, opened 2024-01-03
handle/optimise some ops in void context

iabyn:
This branch:

Fixes a couple of ops which weren't getting their context set correctly.
This doesn't make any change to the correct operation of those ops (they
don't change their behaviour based on context), but could allow for
future optimisations.

Skips a couple of portmanteau OP optimisations ({padv,elemfast}_store) in
non-void context. This allows those two be ops to be slightly faster in void
context (which is overwhelmingly how they are likely to be used), and
falls back to the old separate padsv/sassign op pair otherwise.

Updates a few other ops to handle void context slightly more efficiently.

### DIFF
--- a/gv.c
+++ b/gv.c
@@ -4018,6 +4018,15 @@ Perl_amagic_call(pTHX_ SV *left, SV *right, int method, int flags)
     }
   }
 
+  /* If there's an optimised-away assignment such as $lex = $a + $b, where
+   * the  operator sets the targ lexical directly and skips the sassign,
+   * treat the op as scalar even if its marked as void */
+  if (   PL_op
+      && (PL_opargs[PL_op->op_type] & OA_TARGLEX)
+      && (PL_op->op_private & OPpTARGET_MY)
+  )
+      force_scalar = 1;
+
   switch (method) {
     /* in these cases, we're calling '+' or '-' as a fallback for a ++ or --
      * operation. we need this to return a value, so that it can be assigned
--- a/peep.c
+++ b/peep.c
@@ -3934,6 +3934,11 @@ Perl_rpeep(pTHX_ OP *o)
                   * child (PADSV), and gets to it via op_other rather
                   * than op_next. Don't try to optimize this. */
                  && (lval != rhs)
+                 /* For efficiency, pp_padsv_store() doesn't push its
+                  * result onto the stack. For the relatively rare case of
+                  * the $lex assignment not in void context, we just do it
+                  * the old slow way. */
+                 && OP_GIMME(o,0) == G_VOID
                ) {
                 /* SASSIGN's bitfield flags, such as op_moresib and
                  * op_slabbed, will be carried over unchanged. */
--- a/peep.c
+++ b/peep.c
@@ -3985,6 +3985,11 @@ Perl_rpeep(pTHX_ OP *o)
             if (!(o->op_private & (OPpASSIGN_BACKWARDS|OPpASSIGN_CV_TO_GV))
                 && (lval->op_type == OP_NULL) && (lval->op_private == 2) &&
                 (cBINOPx(lval)->op_first->op_type == OP_AELEMFAST_LEX)
+                 /* For efficiency, pp_aelemfastlex_store() doesn't push its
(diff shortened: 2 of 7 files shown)


peep.c
```diff
@@ -3934,6 +3934,11 @@ Perl_rpeep(pTHX_ OP *o)
                   * child (PADSV), and gets to it via op_other rather
                   * than op_next. Don't try to optimize this. */
                  && (lval != rhs)
+                 /* For efficiency, pp_padsv_store() doesn't push its
+                  * result onto the stack. For the relatively rare case of
+                  * the $lex assignment not in void context, we just do it
+                  * the old slow way. */
+                 && OP_GIMME(o,0) == G_VOID
```
tonycoz, line 3941: I wondered how rare, so I measured it, first, add some code so we can distinguish the cases:
```
--- a/pp_hot.c
+++ b/pp_hot.c
@@ -307,6 +307,8 @@ PP(pp_and)
  *    (PL_op->op_private & OPpASSIGN_BACKWARDS) {or,and,dor}assign
 */
 
+unsigned void_stores;
+
 PP(pp_padsv_store)
 {
     OP * const op = PL_op;
@@ -336,6 +338,8 @@ PP(pp_padsv_store)
         );
     SvSetMagicSV(targ, val);
 
+    if (GIMME_V == G_VOID)
+        ++void_stores;
     rpp_replace_1_1_NN(targ);
     return NORMAL;
 }
```
build and test with `-fprofile-arcs -ftest-coverage`, and review the result:
```
382305756:  341:    if (GIMME_V == G_VOID)
320134726:  342:        ++void_stores;
```
which is 83.7% void context, or 1 in 6 are non-void context.

peep.c
```diff
@@ -3985,6 +3985,11 @@ Perl_rpeep(pTHX_ OP *o)
             if (!(o->op_private & (OPpASSIGN_BACKWARDS|OPpASSIGN_CV_TO_GV))
                 && (lval->op_type == OP_NULL) && (lval->op_private == 2) &&
                 (cBINOPx(lval)->op_first->op_type == OP_AELEMFAST_LEX)
+                 /* For efficiency, pp_aelemfastlex_store() doesn't push its
+                  * result onto the stack. For the relatively rare case of
+                  * the array assignment not in void context, we just do it
+                  * the old slow way. */
+                 && OP_GIMME(o,0) == G_VOID
```
tonycoz, line 3992: This was much closer:
```
   102025:  389:    SvSetMagicSV(targ, val);
        -:  390:
   102023:  391:    if (GIMME_V == G_VOID)
   101925:  392:        ++void_counter;
        -:  393:
```
or 99.9% void context.

tonycoz (approved):

gv.c
```diff
@@ -4018,6 +4018,15 @@ Perl_amagic_call(pTHX_ SV *left, SV *right, int method, int flags)
...
   }
 
+  /* If there's an optimised-away assignment such as $lex = $a + $b, where
+   * the  operator sets the targ lexical directly and skips the sassign,
+   * treat the op as scalar even if its marked as void */
+  if (   PL_op
+      && (PL_opargs[PL_op->op_type] & OA_TARGLEX)
+      && (PL_op->op_private & OPpTARGET_MY)
+  )
+      force_scalar = 1;
```
tonycoz, line 4028: Combined with:
```
    U8 gimme = (force_scalar || PL_op->op_type == OP_MULTICONCAT)
                    ? G_SCALAR : GIMME_V;
```
also in this function, Coverity scan complains that we check whether `PL_op` is non-NULL in this newly added code, but don't check in the assignment to `gimme`.

Is it valid for `PL_op` to be NULL for a call to Perl_amagic_call()?

iabyn:
On Mon, Jan 15, 2024 at 02:08:46PM -0800, Tony Cook wrote:
> Is it valid for `PL_op` to be NULL for a call to Perl_amagic_call()?

Yes. I've fixed up the branch with doesn't test for non-null PL_op via
v5.39.6-159-ged79937e38.

-- 
I before E. Except when it isn't.

PR #21796, opened 2024-01-03
Consolidate handling of emulated langinfo items into a single low-level function

khwilliamson:
When nl_langinfo() is missing from a platform, as notably it isn't present in Windows, perl tries to emulate it.

I have long been unhappy with the situation where the default values get returned at a higher level function, and those that get calculated are done in a lower level one.  The default values get returned for some Configurations where there is no variance in results, or where missing platform abilities preclude our calculating things.

This series of mostly bite-size commits rectifies this, by moving everything into a low-level function.  This has a couple of maintainability benefits

- All the code for an item is in the same place, leading to better understandability 
- The low level function has always been the one that gets used internally.  Prior to this, it couldn't be called without an `#ifdef` to make sure it was usable.  Most such calls needed the same `#ifdef` anyway, but it still had to be taken into consideration.  After these commits, that function can be called without needing to think about it.
- There was some duplication, now eliminated

The commits related to `LC_TIME` have been split into tiny related chunks, as one big one would lead to not being able to understand what's going on so easily


embed.fnc
```diff
@@ -4413,6 +4413,18 @@ S	|void	|ints_to_tm	|NN struct tm *my_tm			\
...
 S	|bool	|is_locale_utf8 |NN const char *locale
+Sr	|void	|setlocale_failure_panic_via_i				\
+				|const locale_category_index cat_index	\
+				|NULLOK const char *current		\
+				|NN const char *failed			\
+				|const line_t proxy_caller_line 	\
+				|const line_t immediate_caller_line	\
+				|NN const char *higher_caller_file	\
+				|const line_t higher_caller_line
+S	|void	|set_save_buffer_min_size				\
```
tonycoz, line 4420: The commits message mentions save_to_buffer(), but this moves only set_save_buffer_min_size() and setlocale_failure_panic_via_i() 

The change to locale.c moves set_save_buffer_min_size() and save_to_buffer() so there's some inconsistency here.

The commit message mentions only save_to_buffer(), I assume you're treating set_save_buffer_min_size() and an implementation detail there, but it doesn't explain setlocale_failure_panic_via_i().

khwilliamson: I think this is a result of `embed.fnc` now being a generated file, which has some downsides.  I moved the two functions `save_to_buffer` and `set_save_buffer_min_size` inside `embed.fnc`, and then the regen phase alphabetically sorts things within each grouping, and the `diff` looks for the minimal diff, which  doesn't necessarily reflect the actual changes from a human perspective.

tonycoz: I don't think that explains it.  If I build with `-Accflags=-DNO_LOCALE` and `make locale.i` then the preprocessed result only includes the definition of save_to_buffer() from locale.c and doesn't include the declaration from proto.h.

I can edit embed.fnc and move save_to_buffer() outside of the `#ifdef USELOCALE` (which appears to be the intent here, similar to set_save_buffer_min_size(), make regen and then locale.i and I see both the save_to_buffer() and set_save_buffer_min_size() declarations from proto.h

Note I inserted save_to_buffer() into it's new position in alphabetical order and the make regen didn't move it.

Note that the setlocale_failure_panic_via_i() declaration is also visible in locale.i when built with `-DNO_LOCALE` as above, which doesn't seem correct.

khwilliamson:
Then the problem was in `git rebase -i`; either my mistake, or the not uncommon mistake of `rerere`.  I redid that commit and looked at all of them in this sequence.  I found several other things where the rebase had been screwed up; though none as consequential as the one @tonycoz ++ spotted.

tonycoz (approved):

PR #21799, opened 2024-01-04
perlfunc: remove some unneeded snark from the docs

rjbs:
Calling them unnecessary is one thing, but "and we do mean backward" seems a bit much.


jkeenan (changes requested):
If you're going to remove that entire paragraph, then why not close up the space between the entry for `%A` and that for `%i`?

However, I think that if de-snarkifying the documentation is your essential goal, we could accomplish this better by inserting this in place of that paragraph:
```
Finally, for backward compatibility, Perl permits these widely-supported conversions:
```

book:
Thanks, applied as 7cf19d62030109daf1492d5e49fc187c592366dc.

PR #21800, opened 2024-01-05
locale.c: Remove redundant lock

khwilliamson:
This is a relic from a past implementation.  The operation is locked by a different macro to lock it, so this is redundant.


khwilliamson:
This is premature

PR #21801, opened 2024-01-05
locale.c Use macro to avoid conditionals

khwilliamson:
We know at compile time what these conditionals always evaluate to.  Use a macro to replace the confitional with nothing when its always going to be true.  (The section of code doesn't even get compiled when its going to always be false.)

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5594,6 +5594,19 @@ S_populate_hash_from_localeconv(pTHX_ HV * hv,
      * global static buffer.  Some locks might be no-ops on this platform, but
      * not others.  We need to lock if any one isn't a no-op. */
 
+    /* If the call could be for either or both of the two categories, we need
+     * to test which one; but if the Configuration is such that we will never
+     * be called with one of them, the code for that one will be #ifdef'd out
+     * below, leaving code for just the other category.  That code will always
+     * want to be executed, no conditional required.  Create a macro that
+     * replaces the condition with an always-true value so the compiler will
+     * omit the conditional */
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -5495,6 +5495,16 @@ S_populate_hash_from_localeconv(pTHX_ HV * hv,
...
+    /* If the call could be for either of the two categories, we need to test
+     * which one; but if the Configuration is such that we will never be called
+     * with one of them, the code for that one won't get compiled below, and
+     * the test is useless, and so omit it by using a macro that expands to
+     * nothing. */
+#    if defined(USE_LOCALE_NUMERIC) && defined(USE_LOCALE_MONETARY)
+#      define IF_CALL_IS_FOR(x)  if (which_mask & OFFSET_TO_BIT(x ## _OFFSET))
+#    else
+#      define IF_CALL_IS_FOR(x)
+#    endif
```
tonycoz: I'm not really comfortable with replacing the if() here, I think I'd prefer something like:
```
#    if defined(USE_LOCALE_NUMERIC) && defined(USE_LOCALE_MONETARY)
#      define CALL_IS_FOR(x)  (which_mask & OFFSET_TO_BIT(x ## _OFFSET))
#    else
#      define CALL_IS_FOR(x) 1
#    endif
...
    if (CALL_IS_FOR(NUMERIC))
...
```

Given this macro is specific to populate_hash_from_localeconv() it could be `#undef`ed at the end of the function.

tonycoz (approved):

PR #21807, opened 2024-01-09
perldelta: Add API function sv_langinfo()

khwilliamson:
This performs the same task as Perl_langinfo8(), but returns an SV. This is typically more convenient for the caller.

It turns out that nl_langinfo() has some weird returns that aren't yet exposed by perl, but future commits will.  An SV makes these easier to cope with.


khwilliamson:
This is premature

Issue #21811, opened 2024-01-09
locale.c: Recent change triggers build failure on OpenBSD


jkeenan:
A recent commit appears to have incited build failures on OpenBSD.  Note how builds are failing before `miniperl` at [this smoketesting page](https://perl5.test-smoke.org/search?selected_osnm=openbsd&report_osnm=openbsd&selected_perl=all).

Opening up the log for [one of these reports](http://perl.develop-help.com/dblog/5047110), I see:
```
3 warnings generated.
In file included from locale.c:385:
In file included from ./perl.h:7828:
./sv_inline.h:75:28: warning: unused parameter 'file' [-Wunused-parameter]
S_new_SV(pTHX_ const char *file, int line, const char *func)
                           ^
./sv_inline.h:75:38: warning: unused parameter 'line' [-Wunused-parameter]
S_new_SV(pTHX_ const char *file, int line, const char *func)
                                     ^
./sv_inline.h:75:56: warning: unused parameter 'func' [-Wunused-parameter]
S_new_SV(pTHX_ const char *file, int line, const char *func)
                                                       ^
locale.c:6088:1: error: conflicting types for 'S_emulate_langinfo'
S_emulate_langinfo(pTHX_ const nl_item item,
^
./proto.h:7087:1: note: previous declaration is here
S_emulate_langinfo(pTHX_ const int item, const char *locale, char **retbufp, Size_t *retbuf_sizep, utf8ness_t *utf8ness);
^
locale.c:6502:34: warning: format specifies type 'int' but the argument has type 'nl_item' (aka 'long') [-Wformat]
                                 item);
                                 ^~~~
```
Building OpenBSD myself (albeit on an out-of-date OpenBSD-6.9), `make miniperl` terminated here:
```
cc -c -DPERL_CORE -DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE -DNO_LOCALE_MONETARY -DNO_LOCALE_TIME -DNO_LOCALE_MESSAGES -DLIBC_HANDLES_MISMATCHED_CTYPE -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings locale.c
locale.c:6088:1: error: conflicting types for 'S_emulate_langinfo'
S_emulate_langinfo(pTHX_ const nl_item item,
^
[...]

PR #21814, opened 2024-01-10
locale.c: Change S_emulate_langinfo formal arg type

khwilliamson:
The type of this parameter changed recently from an enum to an int, and embed.fnc was updated, but not the actual function definition.

No compilers except the OpenBSD one complained so far.

Fixes GH #21811

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6080,7 +6080,7 @@ STATIC const char * S_override_codeset_if_utf8_found(pTHX_
  || ! defined(LC_MESSAGES)
 
 STATIC const char *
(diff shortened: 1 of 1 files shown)


tonycoz:
This fixes the compilation error, but not the format string mismatch also mentioned in #21811

> No compilers except the OpenBSD one complained so far.

Probably because `nl_item` is `long` on OpenBSD while it is `int` on Linux and most others (FreeBSD, darwin at least), there was only a type mismatch on OpenBSD.

khwilliamson:
I missed that part of the ticket.  Now fixed

tonycoz (approved):

PR #21815, opened 2024-01-10
Fix blead not compiling with -DNO_LOCALE

khwilliamson:
This recently introduced failure was due to yet another problem with my doing a 'rebase -i'.  embed.fnc nowadays gets re-sorted as part of the build process, and I need to manually verify each time that the new ordering didn't screw anything up.  Anyway, this function declaration got placed in the wrong bucket of #ifdefs in embed.fnc.  This commit fixes that.

### DIFF
--- a/embed.fnc
+++ b/embed.fnc
@@ -4428,6 +4428,15 @@ S	|char * |strftime8	|NN const char *fmt			\
 				|const bool came_from_sv
 Sf	|char * |strftime_tm	|NN const char *fmt			\
 				|NN const struct tm *mytm
+# if  defined(HAS_IGNORED_LOCALE_CATEGORIES_) || !defined(HAS_NL_LANGINFO) || \
+     !defined(LC_MESSAGES)
+S	|const char *|emulate_langinfo					\
+				|const int item 			\
+				|NN const char *locale			\
+				|NN char **retbufp			\
+				|NULLOK Size_t *retbuf_sizep		\
+				|NULLOK utf8ness_t *utf8ness
+# endif
 # if defined(HAS_LOCALECONV)
 S	|HV *	|my_localeconv	|const int item
 S	|void	|populate_hash_from_C_localeconv			\
@@ -4491,15 +4500,6 @@ RS	|char * |my_setlocale_debug_string_i				\
 				|NULLOK const char *locale		\
 				|NULLOK const char *retval		\
 				|const line_t line
-#   endif
-#   if defined(HAS_IGNORED_LOCALE_CATEGORIES_) || \
-       !defined(HAS_NL_LANGINFO) || !defined(LC_MESSAGES)
-S	|const char *|emulate_langinfo					\
-				|const int item 			\
-				|NN const char *locale			\
-				|NN char **retbufp			\
-				|NULLOK Size_t *retbuf_sizep		\
-				|NULLOK utf8ness_t *utf8ness
 #   endif
 #   if defined(HAS_NL_LANGINFO)
 S	|const char *|my_langinfo_i					\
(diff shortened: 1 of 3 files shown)


ilmari:
Should we add an `-Accflags=-DNO_LOCALE` entry to the [`CONFIGURE_ARGS` matrix](https://github.com/Perl/perl5/blob/blead/.github/workflows/testsuite.yml#L258)?

tonycoz (approved):

khwilliamson:
@ilmari.  I don't know.  It has a bunch of differences from others, but is uncommon in the field.  And don't we have to pay for added work flows?  I personally test these locale changes I've been making in more than 200 different Configurations, but I didn't expect a rebase to break things, so I skipped all but the basic ones thus causing this bug.

jkeenan:
I checked out https://github.com/Perl/perl5/pull/21814 onto my OpenBSD-6.9 VM, configured and built it as far as `make test_prep`.  It compiled; I did not run the test suite.

I then checked out https://github.com/Perl/perl5/pull/21815, configured and attempted to build that branch, also on OpenBSD-6.9.  I was not surprised when it quickly failed.
```
...
cc -c -DPERL_CORE -DNO_LOCALE_NUMERIC -DNO_LOCALE_COLLATE -DNO_LOCALE_MONETARY -DNO_LOCALE_TIME -DNO_LOCALE_MESSAGES -DLIBC_HANDLES_MISMATCHED_CTYPE -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -std=c99 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wno-declaration-after-statement -Wc++-compat -Wwrite-strings locale.c
locale.c:6068:1: error: conflicting types for 'S_emulate_langinfo'
S_emulate_langinfo(pTHX_ const nl_item item,
^
./proto.h:7021:1: note: previous declaration is here
S_emulate_langinfo(pTHX_ const int item, const char *locale, char **retbufp, Size_t *retbuf_...
^
locale.c:6476:34: warning: format specifies type 'int' but the argument has type 'nl_item'
      (aka 'long') [-Wformat]
                                 item);
                                 ^~~~
1 warning and 1 error generated.
*** Error 1 in /home/jkeenan/gitwork/perl (makefile:260 'locale.o': @`sh  cflags "optimize='-O2'" locale.o`  locale.c)
```
I then decided to see what happens when I checkout the 21814 p.r., then merge the 21815 p.r. into the 21814.  I call the resulting branch `smoke-me/jkeenan/gh-21814-21815-combined`.
```
$ uname -mrs
OpenBSD 6.9 amd64
$ gitcurr
smoke-me/jkeenan/gh-21814-21815-combined
$ ./perl -Ilib -V:config_args
config_args='-des -Dusedevel';
[openbsd69: perl] $ gcc --version | head -n 1
gcc (GCC) 4.2.1 20070719 
```
Since I got as far as `make test_prep`, I pushed it to origin for further smoke-testing, especially by our (handful of) OpenBSD smokers -- preferably with newer compilers.  @afresh1, @cjg-cguevara, ^^.

afresh1:
Is there a way to configure Test-Smoke to run smoke-me branches?

jkeenan:
> Is there a way to configure Test-Smoke to run smoke-me branches?

The branch which Test-Smoke runs is the first entry in the file `smokecurrent.gitbranch`.  So when I want to run a smoke-me branch I simply manually edit that file and kick off a run with `sh ./smokecurrent.sh`.  People with more sysadmin-foo than me presumably write `cron` jobs that listen for the birth of new branches whose names begin with `smoke-me/`, then use that to rewrite the first line of `smokecurrent.gitbranch` and kick off a run.

tonycoz:
> Is there a way to configure Test-Smoke to run smoke-me branches?

You might try asking in #smoke on irc.perl.org.

I used to use https://github.com/tonycoz/smoke-me but that was written before Test::Smoke supported git itself, and someone in #smoke will hopefully have something more modern and less of a hack.

afresh1:
I may ask, for now I just used the trick to add the branch to `smokecurrent.gitbranch` (which I wish was a list of regexes for branches to match).  In any case, tests passed -- https://perl5.test-smoke.org/report/5048191

(although I notice that I don't have quite as many `-DNO_LOCALE_*` flags as the other tests.  I can do a manual test if you provide more details on what you want though.

PR #21820, opened 2024-01-11
 locale.c: Change to use SVs for buffers for langinfo()

khwilliamson:
This sequence of commits improves the interface between perl and `nl_langinfo()`  

Prior to this sequence, the code implementing the langinfo interface used
malloc'd buffers to store results in.

However, I discovered that perl has never implemented one langinfo item
properly, because its documentation is misleading.  And that item is not
a standard C string, but instead contains embedded NUL characters.

I started to implement an extension to the existing buffer interface to
handle that.  But then I realized it would need the equivalent of SvCUR
and SvLEN.

And then, I realized that the glibc extensions to nl_langinfo() return
values that are better interpreted as UVs than PVs.

To cope with all that, my implementation would be starting to smell like
perl's already-existing SV implementation.

So these commits change to use SVs behind the scenes.

This presents some advantages, due to the
maturity of the perl SV code.  The same SV can be reused for scratch
with potential for PV stealing, COW, fewer mallocs, etc.  The explicit
freeing of that SV can be delayed until global destruction.


locale.c
```diff
@@ -6421,30 +6421,28 @@ S_emulate_langinfo(pTHX_ const int item,
...
-                char * code_set_name;
-                const Size_t name_len = modifier - retval;
-                Newx(code_set_name, name_len + 1, char);    /* +1 for NUL */
-                my_strlcpy(code_set_name, retval, name_len + 1);
-                SAVEFREEPV(code_set_name);
-                retval = code_set_name;
+                sv_setpvn(sv, name, modifier - name);
+            }
+            else {
+                sv_setpv(sv, name);
```
tonycoz, line 6541: Thank you, thank you, thank you, this made my eye bleed when I came across it in passing, I'd considered suggesting a save_to_bufffern(), but this works too.

locale.c
```diff
@@ -6242,12 +6321,13 @@ S_emulate_langinfo(pTHX_ const int item,
             /* Everything in between is the radix string */
             if (LIKELY(s < e)) {
                 *s = '\0';
```
tonycoz, line 6340: This is no longer needed.

tonycoz (approved):

PR #21821, opened 2024-01-11
perlfunc: document relationship between wait() and waitpid()

### LINKED ISSUE #15108: perldoc for waitpid missing description for argument value  -1
Migrated from [rt.perl.org#127080](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=127080) (status was 'open')

Searchable as RT127080$

tonycoz:
Finally fixes #15108, which was mostly fixed by mauke in a6b6b8e and 237516c.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -10580,6 +10580,8 @@ L<C<$SIG{CHLD}>|perlvar/%SIG>, it may accidentally wait for the child
 created by L<C<qx>|/qxE<sol>STRINGE<sol>> or L<C<system>|/system LIST>.
 See L<perlipc> for details.
 
+Equivalent to L<C<waitpid(-1, 0)>|/waitpid PID,FLAGS>.
+
 Portability issues: L<perlport/wait>.
 
 =item waitpid PID,FLAGS


pod/perlfunc.pod
```diff
@@ -10580,6 +10580,8 @@ L<C<$SIG{CHLD}>|perlvar/%SIG>, it may accidentally wait for the child
 created by L<C<qx>|/qxE<sol>STRINGE<sol>> or L<C<system>|/system LIST>.
 See L<perlipc> for details.
 
+Equivalent to L<C<waitpid(-1, 0)>|/waitpid>.
```
mauke: Isn't that a broken link? I would've expected `L<...|/waitpid PID,FLAGS>` to match the `=item` directive following below.

jkeenan: I agree with @mauke.  Some crude experimentation suggests that:
```
diff --git a/../perl/pod/perlfunc.pod b/./pod/perlfuncx.pod
index 3364c1fd5e..fbeb88c462 100644
--- a/../perl/pod/perlfunc.pod
+++ b/./pod/perlfuncx.pod
@@ -10580,6 +10580,8 @@ L<C<$SIG{CHLD}>|perlvar/%SIG>, it may accidentally wait for the child
 created by L<C<qx>|/qxE<sol>STRINGE<sol>> or L<C<system>|/system LIST>.
 See L<perlipc> for details.
 
+Equivalent to L<C<waitpid(-1, 0)>|/waitpid PID,FLAGS>.
+
 Portability issues: L<perlport/wait>.
 
 =item waitpid PID,FLAGS
```
... will produce a correct link to the following `waitpid` entry.

tonycoz: I was expecting podcheck to pick up a bad link, but I guess not here.

khwilliamson: It's probably because waitpid is also a kernel call, and there are these lines in `t/porting/known_pod_issues.dat`

> waitpid(2)
waitpid(3)

tonycoz: podcheck deliberately allows such links, if I disallow them there are 39 other similar links, an example is a link to perldoc/recv from IO::Socket.

Both perldoc.perl.org and metacpan.org resolve that (though metacpan [scrolls away a little](https://github.com/metacpan/metacpan-web/issues/2972).

I'll fix this link.

jkeenan:
> Finally fixes #15108, which was mostly fixed by mauke in [a6b6b8e](https://github.com/Perl/perl5/commit/a6b6b8ec1b4f13c49a1d7ea81f89a87a654406cb) and [237516c](https://github.com/Perl/perl5/commit/237516c9c19dbadbcd600eb1e3e8bf856a48ee2a).

This looks okay to me -- but since I've never used `wait` or `waitpid` my opinion doesn't count for much.  @mauke @iabyn, care to comment?  Thanks.

mauke (approved):
looks good to me

PR #21832, opened 2024-01-15
Add API function sv_langinfo()

khwilliamson:
This performs the same task as Perl_langinfo8(), but returns an SV. This is typically more convenient for the caller.

It turns out that nl_langinfo() has some weird returns that aren't yet exposed by perl, but future commits will.  An SV makes these easier to cope with.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5795,11 +5795,18 @@ the same information.  But it is more thread-safe than regular
 C<nl_langinfo()>, and hides the quirks of Perl's locale handling from your
 code, and can be used on systems that lack a native C<nl_langinfo>.
 
-However, you should instead use the improved version of this:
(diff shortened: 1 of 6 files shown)


locale.c
```diff
@@ -5885,6 +5907,16 @@ L<I18N::Langinfo>.
                                     emulate_langinfo(item, "C", sv, utf8p)
 #endif
 
+SV *
+Perl_sv_langinfo(pTHX_ const nl_item  item) {
+    utf8ness_t dummy;
```
tonycoz: Is `dummy` needed?  The places I looked the utf8ness pointer can be NULL.

tonycoz (approved):

PR #21833, opened 2024-01-15
 Fix langinfo(ALT_DIGITS)

khwilliamson:
This has never worked properly before in Perl.  The code is returning
the result of the libc function nl_langinfo().  The documentation for it
that I have found (and presumably my predecessors) is very unclear.  But
what actually happens (from using gdb) is that the return is very C
unfriendly.

Instead of returning a NUL-terminated string, it returns 100 (perhaps
fewer) NUL-terminated strings in a row.  When it is fewer (given the
few examples I've seen), the final one ends with two NULs in a row.  (I
can't think of a way for it to work and be otherwise).  The 100th one
doesn't necessarily have two terminating NULs.

Prior to this commit, only the string for the zeroth digit was returned;
now the entire ALT_DIGIT string sequence is returned, forcing a double
NUL at the end of the final one.

This information is accessible in several ways.  Via XS, one can use any
of several functions, including the newly introduced sv_langinfo(),
returning an SV, which allows for easier handling of embedded NULs.
(Otherwise in XS, using the functions that return a char*, one has to
look for the double-NUL.)

From Perl-space, the access is via I18N::Langinfo, which behind the
scenes also uses an SV.  The documentation added in this commit gives
advice for how to turn the return into an `@array` for more convenient
access.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -7095,7 +7095,7 @@ S_emulate_langinfo(pTHX_ const int item,
 #    endif
 #    if defined(WIN32) || ! defined(USE_LOCALE_TIME) || ! defined(HAS_STRFTIME)
 
-          case ALT_DIGITS: retval = "0"; break;
+          case ALT_DIGITS: retval = ""; break;
 #    else
           case ALT_DIGITS:
             format = "%Ow"; /* Find the alternate digit for 0 */
--- a/locale.c
+++ b/locale.c
@@ -6301,9 +6301,97 @@ S_langinfo_sv_i(pTHX_
 
         retval = nl_langinfo(item);
         Size_t total_len = strlen(retval);
+        char separator;
+
+        if (UNLIKELY(item == ALT_DIGITS) && total_len > 0) {
+
+            /* The return from nl_langinfo(ALT_DIGITS) is specified by the
+             * 2017 POSIX Standard as a string consisting of "semicolon-
+             * separated symbols. The first is the alternative symbol
+             * corresponding to zero, the second is the symbol corresponding to
+             * one, and so on.  Up to 100 alternative symbols may be
+             * specified".  Infuriatingly, Linux does not follow this, and uses
+             * the least C-language-friendly separator possible, the NUL.  In
+             * case other platforms also violate the standard, the code below
+             * looks for NUL and any graphic \W character as a potential
+             * separator. */
+            const char * sep_pos = strpbrk(retval,
+                                        " !\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~");
+            if (sep_pos) {
+                separator = *sep_pos;
+            }
+            else if (UNLIKELY(total_len >
+                                        2 * UVCHR_SKIP(PERL_UNICODE_MAX) * 4))
+            {   /* But as a check against the possibility that the separator is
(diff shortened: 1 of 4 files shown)


locale.c
```diff
@@ -6118,13 +6118,66 @@ S_langinfo_sv_i(pTHX_
...
         Size_t total_len = strlen(retval);
+        char separator;
+
+        if (UNLIKELY(item == ALT_DIGITS) && total_len > 0) {
+
+            char * sep_pos =
+                    (char *) strpbrk(retval, "!\"#$%&'()*+,-./.@[\\]^_`{|}~");
+            if (sep_pos) {
+                separator = retval[sep_pos - retval];
+            }
```
tonycoz: You don't dereference `sep_pos` here, and even if the:
```
separator = retval[sep_pos - retval];
```
became:
```
separator = *sep_pos;
```
`sep_pos` can still be `const char *` and , so I don't see the need for the cast here.

Where does the magic string of separators come from?

So four things:
1. the type of sep_pos (const vs non-const pointer and it could be const itself)
2. the cast seems unnecessary with 1.
3. `*sep_pos` vs `retval[sep_pos-retval]`
4. document the source of the magic string

tonycoz:
I found https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html when looking today:
> ALT_DIGITS
>    The alternative symbols for digits, corresponding to the %O conversion specification modifier. The value consists of &lt;semicolon&gt;-separated symbols. The first is the alternative symbol corresponding to zero, the second is the symbol corresponding to one, and so on. Up to 100 alternative symbols may be specified.

though glibc doesn't follow this. (I don't expect any action on this comment)

tonycoz (approved):

khwilliamson:
I changed the comments in`locale.c` based on @tonycoz findings, and I cleaned up the pod the Langinfo module.  Otherwise there were no changes since his approval.

locale.c
```diff
@@ -6141,10 +6141,106 @@ S_langinfo_sv_i(pTHX_
...
+        char separator;
+
+        if (UNLIKELY(item == ALT_DIGITS) && total_len > 0) {
+
+            /* The format of the return from nl_langinfo(ALT_DIGITS) is not
+             * specified by the POSIX Standard, nor much by vendors' man pages.
+             * (although the Standard does specify how they should be formatted
+             * internally in <langinfo.h>, and a conforming implementation
+             * would likely try to do the least work possible in transforming
+             * them for output).  There are up to 100 different values, the
```
tonycoz: The section of that page I was referring to is headed "LC_TIME C-Language Access".

The internal format is described further up and is different from the format documented under  "LC_TIME C-Language Access".

The internal format would be documented under "LC_TIME Locale Definition" is:
```
"0";"1";...
```
while the value documented under "LC_TIME C-Language Access" is:
```
0;1;...
```
though burying it in this page is misleading 😒.

tonycoz:
The glibc ALT_DIGITS bug is mentioned in passing [here](https://sourceware.org/bugzilla/show_bug.cgi?id=31030) though this mostly talks about a similar bug with ERA.

tonycoz (approved):

PR #21835, opened 2024-01-15
Configure: Handle case of no localeconv() found

khwilliamson:
If not found, there certainly won't be a subset of its returns


khwilliamson:
This is related to GH #21803

Tux (approved):
Applied to meta. Go ahead

PR #21836, opened 2024-01-15
Change some USE_ITHREADS to USE_THREADS

khwilliamson:
These are where the underlying implementation should not matter.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -359,7 +359,7 @@ static int debug_initialization = 0;
 /* Automatically include the caller's file, and line number in debugging output;
  * and the errno (and/or extended errno) if non-zero.  On threaded perls add
  * the aTHX too. */
-#  if defined(USE_ITHREADS) && ! defined(NO_LOCALE_THREADS)
+#  if defined(USE_THREADS) && ! defined(NO_LOCALE_THREADS)
 #    define DEBUG_PRE_STMTS                                                 \
         DEBUG_ERRNO;                                                        \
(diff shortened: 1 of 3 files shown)


jkeenan:
My sense is that the discussion of ITHREADS versus THREADS took place so long ago that most current users or readers of this list would not be able to recall the similarities and/or differences.

Could you elaborate a bit as why this is a case where "the underlying implementation should not matter" and what would be a case where the underlying implementations *would* matter?  Thanks.

Leont:
> Could you elaborate a bit as why this is a case where "the underlying implementation should not matter" and what would be a case where the underlying implementations _would_ matter? Thanks.

There is only one implementation left, and that has been the case for some 20 years I think. The distinction is purely theoretical. Consistently using one of the two makes sense, I'm not sure it really matters which one.

Tux:
Commit: 3db8f154c4c6e098a5a0bdf7932e8f86fbd2c451
Author: H.Merijn Brand <h.m.brand@xs4all.nl>  2002-10-19 16:10:21

    Happy chainsaw stories; The removal of the 5005 threads
    Still imcomplete. Configure will follow
    
    p4raw-id: //depot/perl@18030

khwilliamson:
I was following the dicates of config.h which says

> USE_THREADS
This symbol, if defined, indicates that Perl should
be built to use threads.  At present, it is a synonym for
and C<USE_ITHREADS>, but eventually the source ought to be
changed to use this to mean C<_any_> threading implementation.

Leont:
Your logic is correct, but very optimistic (that we'll ever get a new threading model)

Leont (approved):

PR #21839, opened 2024-01-15
loc_tools: Generally return at least C locale

khwilliamson:
If a category exists on the system, but we are supposed to keep it in the C locale, it still is testable as C or POSIX.

Prior to this commit, it wasn't considered testable, and as a result tests in our suite were wrongly skipped.

### DIFF
--- a/t/loc_tools.pl
+++ b/t/loc_tools.pl
@@ -490,7 +490,24 @@ sub find_locales ($;$) {
     my @categories = (ref $input_categories)
                       ? $input_categories->@*
                       : $input_categories;
-    return unless locales_enabled(\@categories);
+
+    # If we can't use at least one of these categories, investigate further
+    if (! locales_enabled(\@categories)) {
+
+        # Not usable at all if system doesn't have locales
+        return unless locales_enabled();
+
+        # Nor if any of the required categories isn't on the system
+        my @on_platform = platform_locale_categories();
+        for my $category (@categories) {
+            return unless grep { $category eq $_ } @on_platform;
+        }
+
+        # Otherwise the category is on the system, but not generally usable.
+        # But the two always-present locales should be usable
+        return ( "C", "POSIX" );
+    }
+
 
     # Note, the subroutine call above converts the $categories into a form
     # suitable for _trylocale().


jkeenan:
> If a category exists on the system, but we are supposed to keep it in the C locale, it still is testable as C or POSIX.
> 
> Prior to this commit, it wasn't considered testable, and as a result tests in our suite were wrongly skipped.

How would I be able to see whether, once this branch had been applied, certain tests would no longer be wrongly skipped?

I built both blead and your branch on the same machine, then ran the tests that `require t/loc_tools.pl`.
```
cd t; TEST_JOBS=1 ./perl harness -v \
../ext/POSIX/t/mb.t \
../ext/POSIX/t/posix.t \
../ext/POSIX/t/time.t \
../ext/POSIX/t/wrappers.t \
../ext/re/t/reflags.t \
../ext/re/t/re_funcs_u.t \
2>1 1>../out; cd -
```
I compared the outputs and the only differences were the names of tempdirs created during `ext/POSIX/t/wrappers.t`.

khwilliamson:
I think the easiest would be to apply this patch on your OpenBSD box, and run it there

jkeenan:
> I think the easiest would be to apply this patch on your OpenBSD box, and run it there

Unfortunately, that didn't show any meaningful differences.
```
$ diff -w blead.out branch.out                                                              
52c52
< ok 29 # 23:36, on 01/15/24
---
> ok 29 # 23:49, on 01/15/24
216c216
< ok 65 - ./HiC_988 does not exist
---
> ok 65 - ./HiC_62991 does not exist
218c218
< ok 67 - ./HiC_988 now exists
---
> ok 67 - ./HiC_62991 now exists
224c224
< ok 73 - ./HiC_988 does not exist
---
> ok 73 - ./HiC_62991 does not exist
356c356
< Files=6, Tests=331,  5 wallclock secs ( 0.06 usr  0.03 sys +  1.00 cusr  0.34 csys =  1.43 CPU)
---
> Files=6, Tests=331,  5 wallclock secs ( 0.05 usr  0.02 sys +  1.04 cusr  0.33 csys =  1.44 CPU)
358c358
< Finished test run at Mon Jan 15 23:36:53 2024.
---
> Finished test run at Mon Jan 15 23:49:26 2024.
```

khwilliamson:
Thanks for finding this

I was calling the wrong function internally.  Try this on blead or the old version of this patch on OpenBSD

`./perl -T -Ilib lib/locale.t | wc -l`

and then run the same with this new version, the count should be 300ish higher than before

jkeenan:
> Thanks for finding this
> 
> I was calling the wrong function internally. Try this on blead or the old version of this patch on OpenBSD
> 
> `./perl -T -Ilib lib/locale.t | wc -l`
> 
> and then run the same with this new version, the count should be 300ish higher than before

Confirmed.
```
blead
$ ./perl -T -Ilib lib/locale.t | wc -l                       
     381
branch
$ ./perl -T -Ilib lib/locale.t | wc -l                       
     682
```

jkeenan (approved):

PR #21842, opened 2024-01-16
win32/GNUmakefile: enable warnings for gcc

tonycoz:
Warnings are already enabled for MSVC, so enable them for GCC too.

-Wformat is disabled since the default "__printf__" format checker doesn't understand the C99-isms like "%zd", resulting in many spurious warnings.  The __gnu_printf__ checker does understand them, but it doesn't understand the Windows specific "%I64d" and similar formats.

I'd hoped to detect `-Wformat` issues, but gcc is horribly broken here.

### DIFF
--- a/win32/GNUmakefile
+++ b/win32/GNUmakefile
@@ -619,9 +619,11 @@ OPTIMIZE	= -Os
 LINK_DBG	= -s
 endif
 
-EXTRACFLAGS	=
+EXTRACFLAGS	= -Wall -Wextra -Werror=pointer-arith -Wno-format -Wno-long-long -Werror=vla
 ifeq ($(USE_CPLUSPLUS),define)
 EXTRACFLAGS	+= $(CXX_FLAG)
+else
+EXTRACFLAGS	+= -std=c99
 endif
 CFLAGS		= $(EXTRACFLAGS) $(INCLUDES) $(DEFINES) $(LOCDEFS) $(OPTIMIZE)
 LINK_FLAGS	= $(LINK_DBG) -L"$(INST_COREDIR)" -L"$(subst ;," -L",$(CCLIBDIR))"


khwilliamson:
LGTM

sisyphus:
Note that it makes the build of perl rather noisy, with 680 "-Wunknown-pragmas" warnings and 211 "-Wunused-parameter" warnings being the main offenders.
The "-Wunknown-pragmas" come mainly (perhaps entirely) from Inline.h, which spits out the following whenever Inline.h is loaded:
```
../inline.h:1507: warning: ignoring '#pragma intrinsic ' [-Wunknown-pragmas]
 1507 | #  pragma intrinsic(_BitScanForward)
      | 
../inline.h:1508: warning: ignoring '#pragma intrinsic ' [-Wunknown-pragmas]
 1508 | #  pragma intrinsic(_BitScanReverse)
      | 
../inline.h:1510: warning: ignoring '#pragma intrinsic ' [-Wunknown-pragmas]
 1510 | #    pragma intrinsic(_BitScanForward64)
      | 
../inline.h:1511: warning: ignoring '#pragma intrinsic ' [-Wunknown-pragmas]
 1511 | #    pragma intrinsic(_BitScanReverse64)
```
Are those pragmas only ever required for MS compilers ?

khwilliamson:
You earlier found that MingW new about those `BitScan` functions and compiled them down to a single instruction.  One could remove those `#pragma` statements and see what gets compiled.

sisyphus:
The troublesome block of code in Inline.h is:
```
#if defined(WIN32)
#  include <intrin.h>
#  pragma intrinsic(_BitScanForward)
#  pragma intrinsic(_BitScanReverse)
#  ifdef _WIN64
#    pragma intrinsic(_BitScanForward64)
#    pragma intrinsic(_BitScanReverse64)
#  endif
#endif
```
I don't think _MSC_VER is complaining about the <code>#pragma</code> statements, so I'll try the following;
```
#if defined(WIN32)
#  include <intrin.h>
#  ifdef _MSC_VER
#    pragma intrinsic(_BitScanForward)
#    pragma intrinsic(_BitScanReverse) 
#    ifdef _WIN64
#      pragma intrinsic(_BitScanForward64)
#      pragma intrinsic(_BitScanReverse64)
#    endif
#  endif
#endif
```
And do a PR if it tests ok.
(If the warnings are being truthful, then that alteration ought to be fine.)

sisyphus:
That removes the <code>-Wunknown-pragmas</code>warnings.
I might as well attack the remaining 363 warnings while I'm at it.
(It might take me a while.)

PR #21847, opened 2024-01-17
Add some locale tests to APItest

khwilliamson:


jkeenan:
The commit message seems to be ... a bit *sparse*.  Could you elaborate a bit, particularly for people who are not very familiar with APItest.pm?  Thanks.

tonycoz:
It could be "Add some C locale test cases to APItest" but I don't think it adds much.

I don't think it needs a bunch of prose explaining APItest, `APItest.pm` describes the purpose of XS::APItest.

tonycoz (approved):

PR #21850, opened 2024-01-17
Get `use VERSION` to import the corresponding `builtin` bundle

leonerd:
A few months ago we added `use builtin :VER` bundles, specifically with the plan that we'd eventually get `use VERSION` to imply it. This PR finally completes the last step of the process, allowing a simple `use VERSION` declaration to import all of the (stable) builtin symbols at that version.

Now the full user convenience is

```perl
use v5.40;
say "True is ", true;
weaken( my $weakref = [] );
# etc...
```

This also applies to the `-E` commandline switch, allowing short oneliners like

```
$ perl -E 'say reftype []'
ARRAY
```

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -596,12 +596,37 @@ static void S_import_sym(pTHX_ SV *sym)
     export_lexical(ampname, (SV *)cv);
 }
 
-#define import_builtin_bundle(ver)  S_import_builtin_bundle(aTHX_ ver)
-static void S_import_builtin_bundle(pTHX_ U16 ver)
+#define cv_is_builtin(cv)  S_cv_is_builtin(aTHX_ cv)
+static bool S_cv_is_builtin(pTHX_ CV *cv)
 {
+    char *file = CvFILE(cv);
+    return file && strEQ(file, __FILE__);
+}
+
+#define import_builtin_bundle(ver, do_unimport)  S_import_builtin_bundle(aTHX_ ver, do_unimport)
+static void S_import_builtin_bundle(pTHX_ U16 ver, bool do_unimport)
+{
+    SV *ampname = sv_newmortal();
+
     for(int i = 0; builtins[i].name; i++) {
-        if(builtins[i].since_ver <= ver)
+        sv_setpvf(ampname, "&%s", builtins[i].name);
+
+        bool want = (builtins[i].since_ver <= ver);
+
+        bool got = false;
+        PADOFFSET off = pad_findmy_sv(ampname, 0);
+        CV *cv;
+        if(off != NOT_IN_PAD &&
+                SvTYPE((cv = (CV *)PL_curpad[off])) == SVt_PVCV &&
+                cv_is_builtin(cv))
+            got = true;
+
+        if(!got && want) {
             import_sym(newSVpvn_flags(builtins[i].name, strlen(builtins[i].name), SVs_TEMP));
+        }
+        else if(do_unimport && got && !want) {
+            pad_add_name_sv(ampname, padadd_STATE|padadd_TOMBSTONE, 0, 0);
+        }
     }
 }
 
@@ -636,7 +661,7 @@ XS(XS_builtin_import)
                 Perl_croak(aTHX_ "Builtin version bundle \"%s\" is not supported by Perl " PERL_VERSION_STRING,
                         sympv);
 
-            import_builtin_bundle(want_ver);
+            import_builtin_bundle(want_ver, false);
 
             continue;
         }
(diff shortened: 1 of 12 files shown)


pod/perlfunc.pod
```diff
@@ -10177,6 +10183,21 @@ file (possibly after a C<package> statement or any amount of whitespace or
 comment), so that its effects happen first, and other pragmata are applied
 after it.
 
+Furthermore, once the version is 5.39.0 or higher, no later C<use VERSION>
+declaration within the same scope is permitted to lower the version again.
+This is because any imported functions from C<builtin> would need to be made
+invisible again.  If you wish to use a higher version only in a small
+section of a file, enclose that section in its own lexical scope.  Once
+outside the scope again afterwards, any imported effects are no longer
```
Grinnz: perhaps "any lexically imported effects" or "any imported effects from use VERSION" so as not to be misleading

t/comp/use.t
```diff
@@ -157,6 +157,12 @@ like $@, qr/^Can't use string/,
     like $warnings, qr/^Downgrading a use VERSION declaration to below v5.11 is deprecated, and will become fatal in Perl 5.40 at /,
         'use 5.01 after use 5.012 provokes deprecation warning';
 }
+eval 'use 5.039; use 5.020;';
+like $@, qr/^Downgrading a use VERSION declaration when above v5\.39 is not permitted at /,
+    'downgrading prevailing version in same scope is not permitted';
+eval '{ use 5.039; } use 5.020;';
+ok !$@, 'downgrading prevailing version by falling out of scope is permitted';
+
```
happy-barney: shouldn't this test also availability of symbols ?

happy-barney: also it would be nice to test behaviour when code already has exported symbol, eg:
```
use Scalar::Util qw (weaken);
use v5.40;
```
(and vice versa)

leonerd: Symbol visibility is tested by a different test; see the one on line 181.

happy-barney: can you move them closer?

also related test:
```
eval "{ use v5.39 } use v5.20; true";
ok $@, ...
```

pod/perlfunc.pod
leonerd: Ahyes; reworded in 3fadae8c81 (will fixup before merge)

tonycoz (approved):

leonerd:
Having discussed this with PSC today, @haarg expressed concern about the new restriction on downgrading version declarations. I won't merge this yet then, and see what can be done differently in order to remove that restriction.

leonerd:
Logic rewritten to perform some unexport of symbols at `use VERSION` time, removing the restriction on downgrading. PTAL.

Once merged this really needs a perldelta, which I'll write in a separate commit.

haarg:
Looks good to me.

leonerd:
@tonycoz - I rewrote quite a lot of this since your last review. This still good with you?

builtin.c
```diff
@@ -683,8 +683,7 @@ XS(XS_builtin_unimport)
          * tombstone and instead COP_SEQ_MAX_HIGH_set() on the padname to
          * clear it.
          */
-        off = pad_add_name_sv(ampname, padadd_STATE|padadd_TOMBSTONE, 0, 0);
-        SvREFCNT_dec(PL_curpad[off]);
+        pad_add_name_sv(ampname, padadd_STATE|padadd_TOMBSTONE, 0, 0);
```
tonycoz, line 712: Does the removal of the SvREFCNT_dec() fix some other problem?

If so the commit message should mention it.

I suspect this was resulting in freeing of a freed scalar further down the line but the commit message should mention it if so.

tonycoz:
> @tonycoz - I rewrote quite a lot of this since your last review. This still good with you?

Looks fine except for wondering about that SvREFCNT_dec() removal.

builtin.c
leonerd: Actually it was indirectly *causing* a problem. Nothing ought to be looking in the pad at the SVs in tombstone entires, but if anything did then previously they'd see the address of an SV that has longsince been freed (or perhaps more likely reused). By not freeing it, we instead leave it as a simple SVt_NULL. I thought that was safer than freeing it then NULL'ing it out, in case something tried to dereference it anyway.

leonerd: I've now made some further adjustments to the code in `pad.c`, and expanded on the wording in the commit message to explain the situation.

tonycoz (approved):

PR #21851, opened 2024-01-18
nostdio.h: Properly indent #nested lines

khwilliamson:


mauke:
I'd prefer "properly indented" lines like this:
```c
#if defined(A)
  #define FOO 1
  #define BAR 2
#endif
```
Does anyone know why we insist on putting the `#` in column 1?
I assume it's for compatibility with ancient compilers, but then we use (some) C99 features, and the ability to put arbitrary whitespace (and even comments) before the `#` in a preprocessor directive was already part of C89, so this seems incongruous.

mauke (approved):

Tux:
IIRC AIX' xlc prefers
```
#ifdef
#  ifdef
#  endif
#endif
```
over
```
#ifdef
  #ifdef
  #endif
#endif
```

PR #21854, opened 2024-01-20
locale.c: Simplify some preprocessor directives

khwilliamson:
This bit of common code is to be compiled if any combination of three other bits get compiled.  Have each of them #define a symbol that it got compiled, so that the common bit only needs to check for that #define


khwilliamson:
closed as redundant

PR #21855, opened 2024-01-20
locale.c: Use hv_deletes instead of hv_delete.

khwilliamson:
I stumbled upon this macro where the parameter must be a literal.

It is hardly used in core.  Should it be API?


tonycoz (approved):
I think it's reasonable to make hv_deletes() and the other similar macros in hv.h API.

Leont:
Yeah, they should be API.

PR #21860, opened 2024-01-20
Fix not compiling when NO_LOCALE

khwilliamson:
This function is used when locale handling on the system doesn't exist or is turned off; therefore it must be compiled under that circumstance.


khwilliamson:
A different solution that fixes more things will replace this PR

PR #21861, opened 2024-01-20
Don't mortalize my_localeconv() for internal use

khwilliamson:
For external consumption, retain the mortalization; but for internal explicitly decrement the reference count when done.

I don't know the area of code dealing with temporaries very well.  But it seems to me more desirable to leave a value around for as short a time as possible and then explicitly get rid of it.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -4999,7 +4999,7 @@ fields), but directly callable from XS code.
 HV *
 Perl_localeconv(pTHX)
 {
-    return my_localeconv(0);
+    return (HV *) sv_2mortal((SV *) my_localeconv(0));
 }
 
 HV *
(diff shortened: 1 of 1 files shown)


Leont:
Do note that mortalizing is exception-safe (in the sense that it will not leak the SV if an exception is thrown), I'd guess it also needs a `SvREFCNT_dec_NN` before the panic, and maybe elsewhere too.

khwilliamson:
I'm trying to understand this.  I don't write panics unless it is something that is unexpected and prevents proceeding in a reasonable manner.  So that if a panic does show up, it means it is a bug that should be fixed.  A leak under those circumstances is the least of our worries.   This particular panic shouldn't happen, because the hash was just populated with the key in question.  Did some change happen where the populating got skipped?  It's a fail safe that the code remains consistent in this regard.

Leont (approved):

PR #21871, opened 2024-01-22
locale.c: Add "struct lconv" then remove definition as a fallback

khwilliamson:
This is almost certainly available on your platform; but if it is not, a comment is added so that you can revert a commit to get it for you.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -5003,6 +5003,50 @@ Perl_localeconv(pTHX)
     return my_localeconv(0);
 }
 
+#ifndef HAS_LOCALECONV
+#if 0
+
+/* If localeconv() is unavailable, the functions below simulate its presence as
+ * if the locale were the C one.
+ *
+ * If this structure definition is also unavailable, those functions won't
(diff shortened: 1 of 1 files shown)


tonycoz:
Is there any real value to this?  localeconv() and struct lconv are C89, let alone C99.

khwilliamson:
I don't know if there is any value; the net result is just a comment saying revert a commit and get this structure without having to type it.  I added the structure when checking that things worked on platforms without it; and I thought maybe it would be useful sometime to someone; but don't clutter up the code with it now. 

So, I'm fine either way

jkeenan:
> I don't know if there is any value; the net result is just a comment saying revert a commit and get this structure without having to type it. I added the structure when checking that things worked on platforms without it; and I thought maybe it would be useful sometime to someone; but don't clutter up the code with it now.
> 
> So, I'm fine either way

Given the above discussion, closing this p.r.  We can always open a new ticket if warranted.  Thanks.

Issue #16607, opened 2018-07-03
stack checks make -DDEBUGGING builds incompatible with normal ones


p5pRT:
Migrated from [rt.perl.org#133327](https://rt-archive.perl.org/perl5/Ticket/Display.html?id=133327) (status was 'open')

Searchable as RT133327$

p5pRT:
### From @ntyni


This is a bug report for perl from Niko Tyni \<ntyni@&#8203;debian\.org>\,
generated with the help of perlbug 1\.41 running under perl 5\.28\.0\.




---
The stack 'high water mark' checks introduced in v5\.27\.1\-66\-g87058c31e9
seem to have introduced an incompatibility when XS modules are built
with a normal Perl but run with a \(otherwise similarly configured\)
\-DDEBUGGING Perl\.

We're using this combination in Debian as we offer a \-DDEBUGGING
interpreter \(called /usr/bin/debugperl\) in a separate package for the
benefit of our users\, but we don't rebuild all the XS modules with that\.

It would be useful for us to keep these builds binary compatible\.
Last time an incompatibility happened it was deemed worth fixing&#8203;:
see \[perl \#127212\] and Aaron's related commit message in 5ca5a62834a7f \.

Some symptoms of the breakage&#8203;:

 \# debugperl \-MDateTime \-e 'print DateTime\->today'
 panic&#8203;: XSUB DateTime&#8203;::\_rd2ymd \(DateTime\.c\) failed to extend arg stack&#8203;: base=55b63e6b3b48\, sp=55b63e6b3b80\, hwm=5
5b63e6b3b68

 \# printf '\<AAA>\\n\<DDD/>\<CCC>\<DDD>\\n\</DDD>\</CCC>\\n\</AAA>' | debugperl \-MXML&#8203;::LibXML \-e 'XML&#8203;::LibXML\->new\->parse\_f
h\(\*STDIN\)\->documentElement\->childNodes; '
 panic&#8203;: XSUB XML&#8203;::LibXML&#8203;::Node&#8203;::\_childNodes \(LibXML\.c\) failed to extend arg stack&#8203;: base=561827c61b48\, sp=561827c6
1b68\, hwm=561827c61b60

These are just examples\, and I believe the fault is not with the
modules\.

If I grok this correctly\, the problem is that EXTEND only updates
the high\-water mark when the DEBUGGING preprocessor symbol is defined
\(i\.e\. the choice is done at compile time\.\) If the XS module is built
[...]

PR #21872, opened 2024-01-22
Add builtin::inf and builtin::nan

leonerd:
Will take opinions on whether "nan" should be capitalised as "NaN" to match its general style elsewhere.

Also if anyone can suggest better documentation than "numerical not-a-number" that'd be great ;)

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -89,14 +89,39 @@ XS(XS_builtin_false)
     XSRETURN_NO;
 }
 
+XS(XS_builtin_inf);
+XS(XS_builtin_inf)
+{
+    dXSARGS;
+    if(items)
+        croak_xs_usage(cv, "");
+    EXTEND(SP, 1);
+    XSRETURN_NV(NV_INF);
+}
+
+XS(XS_builtin_nan);
+XS(XS_builtin_nan)
+{
+    dXSARGS;
+    if(items)
+        croak_xs_usage(cv, "");
+    EXTEND(SP, 1);
+    XSRETURN_NV(NV_NAN);
+}
+
 enum {
     BUILTIN_CONST_FALSE,
     BUILTIN_CONST_TRUE,
+    BUILTIN_CONST_INF,
+    BUILTIN_CONST_NAN,
 };
 
 static OP *ck_builtin_const(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
 {
     const struct BuiltinFuncDescriptor *builtin = NUM2PTR(const struct BuiltinFuncDescriptor *, SvUV(ckobj));
+
+    if(builtin->is_experimental)
+        warn_experimental_builtin(builtin->name);
 
     SV *prototype = newSVpvs("");
     SAVEFREESV(prototype);
@@ -109,6 +134,8 @@ static OP *ck_builtin_const(pTHX_ OP *entersubop, GV *namegv, SV *ckobj)
     switch(builtin->ckval) {
         case BUILTIN_CONST_FALSE: constval = &PL_sv_no; break;
         case BUILTIN_CONST_TRUE:  constval = &PL_sv_yes; break;
+        case BUILTIN_CONST_INF:   constval = newSVnv(NV_INF); break;
+        case BUILTIN_CONST_NAN:   constval = newSVnv(NV_NAN); break;
         default:
             DIE(aTHX_ "panic: unrecognised builtin_const value %" IVdf,
                       builtin->ckval);
@@ -518,6 +545,8 @@ static const struct BuiltinFuncDescriptor builtins[] = {
     /* constants */
     { "true",  SHORTVER(5,39), &XS_builtin_true,   &ck_builtin_const, BUILTIN_CONST_TRUE,  false },
     { "false", SHORTVER(5,39), &XS_builtin_false,  &ck_builtin_const, BUILTIN_CONST_FALSE, false },
+    { "inf",        NO_BUNDLE, &XS_builtin_inf,    &ck_builtin_const, BUILTIN_CONST_INF,   true },
+    { "nan",        NO_BUNDLE, &XS_builtin_nan,    &ck_builtin_const, BUILTIN_CONST_NAN,   true },
 
     /* unary functions */
     { "is_bool",         NO_BUNDLE, &XS_builtin_func1_scalar, &ck_builtin_func1, OP_IS_BOOL,    true  },
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -51,6 +51,25 @@ package FetchStoreCounter {
     is(prototype(\&builtin::is_bool), '$', 'is_bool prototype');
 }
 
+# float constants
+{
+    use builtin qw( inf nan );
+
+    ok(inf, 'inf is true');
+    ok(inf > 1E10, 'inf is bigger than 1E10');
+    ok(inf == inf, 'inf is equal to inf');
+    ok(inf == inf + 1, 'inf is equal to inf + 1');
+
+    # Invoke the real XSUB
+    my $inf = ( \&builtin::inf )->();
+    ok($inf == $inf + 1, 'inf returned by real xsub');
+
+    ok(nan != nan, 'NaN is not equal to NaN');
+
+    my $nan = ( \&builtin::nan )->();
+    ok($nan != $nan, 'NaN returned by real xsub');
+}
+
 # weakrefs
 {
     use builtin qw( is_weak weaken unweaken );
--- a/lib/builtin.pm
+++ b/lib/builtin.pm
@@ -1,4 +1,4 @@
-package builtin 0.012;
+package builtin 0.013;
 
 use strict;
 use warnings;
@@ -18,6 +18,7 @@ builtin - Perl pragma to import built-in utility functions
 
     use builtin qw(
         true false is_bool
+        inf nan
         weaken unweaken is_weak
         blessed refaddr reftype
         created_as_string created_as_number
@@ -142,6 +143,22 @@ or any variable containing one of these results.
 
 This function used to be named C<isbool>. A compatibility alias is provided
 currently but will be removed in a later version.
+
+=head2 inf
+
+    $num = inf;
+
+This function is currently B<experimental>.
+
+Returns the floating-point infinity value.
+
+=head2 nan
+
+    $num = nan;
+
+This function is currently B<experimental>.
+
+Returns the floating-point "Not-a-Number" value.
 
 =head2 weaken
 


lib/builtin.t
```diff
@@ -51,6 +51,17 @@ package FetchStoreCounter {
     is(prototype(\&builtin::is_bool), '$', 'is_bool prototype');
 }
 
+# float contants
+{
+    use builtin qw( inf nan );
+
+    ok(inf, 'inf is true');
+    ok(inf > 1E10, 'inf is bigger than 1E10');
+    ok(inf == inf, 'inf is equal to inf');
```
Grinnz, line 60: A good test here would be inf == inf + 1 or similar

Grinnz:
Also of note is that POSIX provides "Inf", "NaN", and "nan" (which I guess can contain more information about which number it is not), so perhaps this is not necessary, the only reason I don't generally reach for those is because they only exist since 5.22 and POSIX is not dual-life.

lib/builtin.t
```diff
@@ -51,6 +51,17 @@ package FetchStoreCounter {
     is(prototype(\&builtin::is_bool), '$', 'is_bool prototype');
 }
 
+# float contants
```
skington: Trivial: typo, constants.

johannessen:
Prior art: [Data::Float](https://metacpan.org/pod/Data::Float) also uses `nan`.

lib/builtin.t
```diff
@@ -51,6 +51,17 @@ package FetchStoreCounter {
...
+{
+    use builtin qw( inf nan );
+
+    ok(inf, 'inf is true');
+    ok(inf > 1E10, 'inf is bigger than 1E10');
+    ok(inf == inf, 'inf is equal to inf');
+
+    ok(nan != nan, 'NaN is not equal to NaN');
```
Ovid, line 67: I would like to see more tests here. `>=`, `>`, `<=`, `<`, and `==`, should all return false.

lib/builtin.t
Ovid, line 60: [Infinity should be larger than all other values except itself and NaN](https://www.gnu.org/software/libc/manual/html_node/Infinity-and-NaN.html). Tests for that would be nice.

haarg: `inf == inf / 2` might be a reasonable test.

haarg:
This should probably be explicit about what "type" of NaN value it is returning. See [POSIX's nan function](https://metacpan.org/pod/POSIX#nan).

POSIX also has the functions `isinf` and `isnan`. I'm unsure if they should be included here.

ilmari:
> This should probably be explicit about what "type" of NaN value it is returning. See [POSIX's nan function](https://metacpan.org/pod/POSIX#nan).

Should it usurp the entire functionality of `POSIX::nan`, or should it just say to use that if you need it?

> POSIX also has the functions `isinf` and `isnan`. I'm unsure if they should be included here.

It would make sense, but they should probably be spelled `is_inf` and `is_nan` to match `is_bool`.

lib/builtin.t
leonerd: I could do a couple more tests but I didn't want to go overboard. I'm not intending to test the concept of "inf" itself, I'm sure that's handled elsewhere. I'm just checking that I've wrapped the right thing.

leonerd:
I could consider adding the `is_...` checks as well, but perhaps in a later commit? This one was supposed to be a quick 5 minutes addition of the two constants.

leonerd:
> This should probably be explicit about what "type" of NaN value it is returning. See [POSIX's nan function](https://metacpan.org/pod/POSIX#nan).

@haarg It would be useful to know, but I don't think we portably always do know. This constant is just using `PL_nan` from `perl.h`, and that isn't clear on what kind of NaN it is. Also note that the distinction between signalling and quiet versions is specifically an IEEE'ism, and might not portably exixst in all platforms.

Grinnz:
> > This should probably be explicit about what "type" of NaN value it is returning. See [POSIX's nan function](https://metacpan.org/pod/POSIX#nan).
> 
> Should it usurp the entire functionality of `POSIX::nan`, or should it just say to use that if you need it?

It is probably best to keep this version as simple as possible for the vast majority of needs.

perlpunk:
++
I as a YAML module author will be happy to see builtin inf and nan in perl :)

Leont:
We should probably have a test that checks how this interacts with bigint/bigrat/bignum.

rwp0 (approved):

lib/builtin.t
tonycoz: There are already comprehensive tests for handling of inf and nan values in `t/op/infnan.t`

tonycoz:
I think the lower-case names are reasonable and consistent.

> Also if anyone can suggest better documentation than "numerical not-a-number" that'd be great ;)

Maybe "floating point" rather than "numerical" (neither is an actual number) as a clue that things aren't quite "real".

I think the inf/nan value tests in `t/op/infnan.t` cover much of the extra tests covered here.

I do think it could use a better test for inf (inf /2 == inf works I think) and any test for nan (nan != nan I guess).

hvds:
> I do think it could use a better test for inf (inf /2 == inf works I think) and any test for nan (nan != nan I guess).

`$x/2 == $x` would also be satisfied by 0 (and infinitesimals, if they were supported); `$x + 1 == $x` would also be satisfied by large values; but I think the two in combination should uniquely be satisfied by +/- inf, and `$x > 0` could decide between the two.

leonerd:
I see the request to also add `is_inf` and `is_nan`, but I think I'd like to do those in a separate follow-up PR as they're a bit more involved and would poke at things in opcode.h and so on as well. For now I'd like to leave this PR small.

builtin.c
```diff
@@ -87,15 +87,40 @@ XS(XS_builtin_false)
     XSRETURN_NO;
 }
 
+XS(XS_builtin_inf);
+XS(XS_builtin_inf)
+{
+    dXSARGS;
+    if(items)
+        croak_xs_usage(cv, "");
+    mPUSHn(PL_inf.nv);
```
wolfsage: Huh, don't we need EXTEND(SP, 1) before PUSH? mPUSHn says the stack must have room for the item being pushed. Or is there some other magic going on?

wolfsage: I guess... the stack always has 1 spot free for simple returns but anything beyond that you have to EXTEND for?

leonerd: Oops, we do need that; but apparently I was just lucky in all the tests here that it didn't matter. But yes, I should add it.

skington: Is there any test coverage for this? Here and elsewhere? If it's possible to accidentally break the stack, shouldn't we test that we haven't?

wolfsage: I couldn't find a test to trip it up, even compiling with -Accflags=-DSTRESS_REALLOC

lib/builtin.t
leonerd: Added `inf == inf + 1`. There's no need to go overboard on them; most of the value tests are indeed elsewhere. This is just testing that the `builtin::inf` function appears to have grabbed the correct value there.

builtin.c
leonerd: This kind of thing is quite hard to reliably test for, because you need to ensure that the stack already has exactly zero elements spare when you enter the function. It just grows over time as different things end up using it, so it's quite hard to arrange a size already.

leonerd: Actually, the existing true/false also have a bug. They should be fixed first - see https://github.com/Perl/perl5/pull/21888

skington: Is there any existing linting that could be extended to check for this sort of thing?

iabyn:
On Thu, Jan 25, 2024 at 02:13:14PM -0800, Paul Evans wrote:
> +XS(XS_builtin_inf);
> +XS(XS_builtin_inf)
> +{
> +    dXSARGS;
> +    if(items)
> +        croak_xs_usage(cv, "");
> +    mPUSHn(PL_inf.nv);
> 
> This kind of thing is quite hard to reliably test for, because you need
> to ensure that the stack already has exactly zero elements spare when
> you enter the function. It just grows over time as different things end
> up using it, so it's quite hard to arrange a size already.

Under DEBUGGING builds, stuff like this should be automatically detected.
Look at the code in pp_entersub which checks si_stack_hwm.

Either you didn't test with DEBUGGING. or that XS code isn't pushing onto
an area of stack which hasn't reecently been extended for, or there's a
bug or conceptual problem with the si_stack_hwm mechanism

-- 
    -- Churchill on Montgomery

wolfsage:
> Under DEBUGGING builds, stuff like this should be automatically detected. Look at the code in pp_entersub which checks si_stack_hwm. Either you didn't test with DEBUGGING. or that XS code isn't pushing onto an area of stack which hasn't reecently been extended for, or there's a bug or conceptual problem with the si_stack_hwm mechanism

I tried under DEBUGGING and couldn't trigger a problem, but maybe my tests were bad. If you can trip it up Dave I'd love to see it!

leonerd:
> Either you didn't test with DEBUGGING. or that XS code isn't pushing onto an area of stack which hasn't reecently been extended for, or there's a bug or conceptual problem with the si_stack_hwm mechanism

I expect that since these are only tested by little calls at the main level of `lib/builtin.t`, by that point in time the stack will have already been extended quite a bit past the single level they'll need to push to.

Perhaps a better test could be something like

```perl
eval("builtin::true, " x 1024);
```

Annoying eval, but I couldn't offhand think of a way to get a lot of calls onto the real stack that didn't involve map or someother mess.

iabyn:
On Fri, Jan 26, 2024 at 07:27:22AM -0800, Paul Evans wrote:
> > Either you didn't test with DEBUGGING. or that XS code isn't pushing onto an area of stack which hasn't reecently been extended for, or there's a bug or conceptual problem with the si_stack_hwm mechanism
> 
> I expect that since these are only tested by little calls at the main
> level of `lib/builtin.t`, by that point in time the stack will have
> already been extended quite a bit past the single level they'll need to
> push to.

I think you are misunderstanding what the si_stack_hwm high-water-mark
debugging mechanism does. The si_stack_hwm field is set to the current
PL_stack_sp at the start of *each* op execution. The EXTEND macros
and similar update the hwm.  At the end of each op or XS call, the code in
runops and pp_entersub checks whether PL_stack_sp > si_stack_hwm. If so, it
means that the op pushed something on the stack without calling EXTEND()
first.

So even if the stack has already grown from previous activity, it will
croak if *that* op or XS call did a PUSHs without a preceding POPs or
EXTEND(1) within the same function.

In the specific case of builtin::true() and similar XS functions, the
stack has already just been extended by at least 1 to push the
gv(builtin::true) just before calling pp_entersub(). So XS functions
inherit at least one pre-extended slot, at least when invoked the normal
way (but not if called directly from other C code).

Which is why builtin::true didn't trigger a croak.

-- 
"I do not resent criticism, even when, for the sake of emphasis,
it parts for the time with reality".
    -- Winston Churchill, House of Commons, 22nd Jan 1941.

tonycoz:
> In the specific case of builtin::true() and similar XS functions, the stack has already just been extended by at least 1 to push the gv(builtin::true) just before calling pp_entersub(). So XS functions inherit at least one pre-extended slot, at least when invoked the normal way (but not if called directly from other C code).
> 
> Which is why builtin::true didn't trigger a croak.

I was aware of this from writing tests for #16607, and more recently looked at making this stricter.

It turns out EU::PXS is generating code that doesn't EXTEND for single item returns, eg:
```
NV
time()
    PREINIT:
        struct timeval Tp;
    CODE:
        int status;
        status = gettimeofday (&Tp, NULL);
        if (status == 0) {
            RETVAL = Tp.tv_sec + (Tp.tv_usec / NV_1E6);
        } else {
            RETVAL = -1.0;
        }
    OUTPUT:
        RETVAL
```
becomes:
```
XS_EUPXS(XS_Time__HiRes_time); /* prototype to pass -Wmissing-prototypes */
XS_EUPXS(XS_Time__HiRes_time)
{
    dVAR; dXSARGS;
    if (items != 0)
       croak_xs_usage(cv,  "");
    {
#line 1209 "HiRes.xs"
        struct timeval Tp;
#line 1631 "HiRes.c"
	NV	RETVAL;
	dXSTARG;
#line 1211 "HiRes.xs"
        int status;
        status = gettimeofday (&Tp, NULL);
        if (status == 0) {
            RETVAL = Tp.tv_sec + (Tp.tv_usec / NV_1E6);
        } else {
            RETVAL = -1.0;
        }
#line 1642 "HiRes.c"
	XSprePUSH;
	PUSHn((NV)RETVAL);
    }
    XSRETURN(1);
}
```
which doesn't extend.

The question is whether we treat this as an error.

Some other core XS which does its own stack handling has similar problems, such as Cwd::getcwd().

Very little code calls XSUBs directly from what I can see:
https://metacpan.org/dist/SWF/source/Bitmap.xs#L44 (2007)
https://metacpan.org/release/NUFFIN/B-XSUB-Dumber-0.01/source/Dumber.xs#L14 (2008, module marked as "don't use this seriously")

leonerd:
So I think the summary is that scalar-returning XSUBs probably *should* extend the stack but in practice most situations they'd be called don't matter if they don't because of the CV to `pp_entersub` being there anyway. There's no harm in us doing it anyway because it'll basically be a noöp if we do, but at least this'll be one fewer place to fix a bug if that situation ever changes.

With that said, this feels like it's ready to merge now?

tonycoz (approved):

lib/builtin.t
haarg: $x == $x + 1 will also be true for large (not infinite) floats.

PR #21874, opened 2024-01-23
Add newSVpvz() to create a sized SvPV.

khwilliamson:
There's no current trivial method to create a non-zero length  SV that will hold a PV.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -5795,6 +5795,28 @@ Perl_newSV(pTHX_ const STRLEN len)
     }
     return sv;
 }
+
+/*
+=for apidoc newSVpvz
+
+Creates a new SV initialized with an empty string, like C<newSVpvs("")>, but
+with enough available space to hold a string of C<len> bytes (plus a trailing
+NUL) without needing to grow.
+
+The reference count for the new SV is set to 1.
+
+=cut
+*/
+SV *
+Perl_newSVpvz(pTHX_ const STRLEN len)
+{
+    SV *sv = newSV_type(SVt_PV);
+    sv_grow_fresh(sv, len + 1);
+    (void) sv_setpv_freshbuf(sv);
+
+    return sv;
+}
+
 /*
 =for apidoc sv_magicext
 
--- a/embed.fnc
+++ b/embed.fnc
@@ -2273,6 +2273,7 @@ ARdp	|SV *	|newSVpvn_share |NULLOK const char *s			\
 				|U32 hash
 ARdp	|SV *	|newSVpv_share	|NULLOK const char *s			\
 				|U32 hash
+ARdp	|SV *	|newSVpvz	|const STRLEN len
 ARdp	|OP *	|newSVREF	|NN OP *o
 Adp	|SV *	|newSVrv	|NN SV * const rv			\
 				|NULLOK const char * const classname
--- a/embed.h
+++ b/embed.h
@@ -440,6 +440,7 @@
 # define newSVpvn(a,b)                          Perl_newSVpvn(aTHX_ a,b)
 # define newSVpvn_flags(a,b,c)                  Perl_newSVpvn_flags(aTHX_ a,b,c)
 # define newSVpvn_share(a,b,c)                  Perl_newSVpvn_share(aTHX_ a,b,c)
+# define newSVpvz(a)                            Perl_newSVpvz(aTHX_ a)
 # define newSVrv(a,b)                           Perl_newSVrv(aTHX_ a,b)
 # define newSVsv_flags(a,b)                     Perl_newSVsv_flags(aTHX_ a,b)
 # define newSVuv(a)                             Perl_newSVuv(aTHX_ a)
--- a/proto.h
+++ b/proto.h
@@ -3056,6 +3056,11 @@ Perl_newSVpvn_share(pTHX_ const char *s, I32 len, U32 hash)
 #define PERL_ARGS_ASSERT_NEWSVPVN_SHARE
 
 PERL_CALLCONV SV *
+Perl_newSVpvz(pTHX_ const STRLEN len)
+        __attribute__warn_unused_result__;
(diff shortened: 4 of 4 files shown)


sv.c
```diff
@@ -5795,6 +5795,30 @@ Perl_newSV(pTHX_ const STRLEN len)
...
+
+=cut
+*/
+SV *
+Perl_newSVpvz(pTHX_ const STRLEN len)
+{
+    SV *sv = newSV_type(SVt_PV);
+    sv_grow_fresh(sv, len + 1);
+    SvPOK_only(sv);
+    SvPVX(sv)[0] = '\0';
```
leonerd: Small thought: I wonder if this should also `SvPVX(sv)[len] = '\0'` so as to ensure that if the caller only initialises the `[0 .. len-1]` bytes to whatever was wanted, there's definitely still a trailing NUL at the end.

Leont: Instead of it merely being preallocated, what if it actually was a string of that length? Either zeroed out or undefined value? Because when you preallocate it like this you probably don't want to have to mess around with `SvCUR` right after.

leonerd: As it stands, the new SV will have `SvCUR == 0`, so it's up to the caller to do the right thing. If it returned `SvCUR == len` then it would be good to zero out the entire buffer so it's determined; I wouldn't like the idea of having uninitialised bytes visible there.

khwilliamson: This seems like a good idea.

khwilliamson: It could use `sv_setpv_freshbuf(sv)` to initialize (which handles tainting possibilities).  But why does that function call `SvPOK_only_UTF8`?  I think it is irrelevant for a fresh PV, but it seems to me that it would be more robust to clear the flag.

I'm inclined to say something like

> Creates a new SV initialized with a zero-length string, and sized with enough
preallocated space to hold a string of len bytes (plus a trailing NUL)
without needing to grow.

Leont: That does raise a good point, this function could use a flags argument for `SVf_UTF8`

Leont: >  But why does that function call SvPOK_only_UTF8? I think it is irrelevant for a fresh PV, but it seems to me that it would be more robust to clear the flag.

It disables typeflags except POK and UTF8, that does sound like the right thing to do in that particular case.

Leont:
I'd love to see some examples of how this would be used, because I doubt there are many use-cases where it's being used in any way other than immediately setting `SvCUR(sv) = len;` right after calling this function.

khwilliamson:
I'm calling a libc function which is passed a buffer and its size.  It fills the buffer if it fits, and returns the actual number of bytes used;  if it couldn't fit, it returns 0, and you have to increase the buffer size by realloc()  and try again.

Currently each call to the function starts with a a Newx of a buffer based on a guess as to its needed size.  In many cases, the populated buffer is then attached to an SV via some sv_usepv()-ish call.  But other cases don't lend themselves to that, and the buffer gets examined, processed, and eventually freed.

It would be more efficient to pass the SvPVX of the SV that is to eventually be returned, sized according to the guess.  And for the cases where the value needs processing, to use a scratch SV whose life ends only on global destruction, and populate it fresh each time.  That way the SV ends up quickly getting large enough to hold the typical request of the current program, and you don't have to free and start fresh each time.

I have had other cases in the past where this would have been a more natural solution than whatever I ended up doing; but I don't remember now what they were.

sv.c
khwilliamson: I don't understand your point.  This is a fresh SV.  Creating it does not turn on the UTF-8 flag.  Therefore there is no need to preserve it being on.  Keeping it at zero is a no-op.  And hence using a construct that preserves it makes me stop and question what is going on here?  And the answer after wasting my time is nothing is going on.  Or do I misunderstand?

And I don't see the need for a flags argument.  If we knew what is going to be placed into the PV, we would be using something like 'newSVpv()` and its kin.  I would be using this function when I don't know what the string is going to be, much less if it is going to be UTF-8 or not, but I do have a guess as to how big to make it so as to avoid in most circumstances the need for additional reallocs.  Changing a flag setting is a lot cheaper than a malloc potentially is.

Maybe my situation is so specialized that a special function isn't warranted, but I have wanted this very functionality in the past.

tonycoz:
This could use a mention in perlguts around "If you want to write to an existing SV's buffer ..." which discusses writing directly to the buffer of existing SVs.

Leont:
> I'm calling a libc function which is passed a buffer and its size. It fills the buffer if it fits, and returns the actual number of bytes used; if it couldn't fit, it returns 0, and you have to increase the buffer size by realloc() and try again.
> 
> Currently each call to the function starts with a a Newx of a buffer based on a guess as to its needed size. In many cases, the populated buffer is then attached to an SV via some sv_usepv()-ish call. But other cases don't lend themselves to that, and the buffer gets examined, processed, and eventually freed.

That's a valid use-case, but I don't think this function is a particularly good use-case for it because you'd have to keep growing it anyway.

sv.c
Leont: > Maybe my situation is so specialized that a special function isn't warranted, but I have wanted this very functionality in the past.

I think we just have different situations in mind.

khwilliamson: Could you elaborate on the situation you are thinking of?

khwilliamson:
I think you misunderstand; the initial modest size is enough that growth isn't required except under exceptional circumstances.

jkeenan:
@khwilliamson @Leont @leonerd @tonycoz  ... Discussion in this pull request petered out more than 6 months ago.  It doesn't appear that a consensus ever emerged as to whether or not we need this functionality.  Can we get your updated thoughts on both necessity/value and implementation?  Thanks.

Leont:
> I think you misunderstand; the initial modest size is enough that growth isn't required except under exceptional circumstances.

Fair enough. I'm fine with merging this as-is (even if I'd prefer the other behavior).

tonycoz (approved):

Issue #21867, opened 2024-01-22
BBC: blead breaks Safe when `use`-ing modules without an `import` method.


mkende:
Module: Safe

**Description**

`use` of a module inside a `Safe` container fails if the module does not expose an `import` method.

This seems related to https://perldoc.perl.org/5.39.1/perldelta#Calling-the-import-method-of-an-unknown-package-produces-an-error. However, the module I’m using exists and I’m calling `use` without any argument, so this is not the case described in this `perldelta` entry.

**Steps to Reproduce**

```
perl -MSafe -e '$s = Safe->new(); $s->deny_only(); $s->reval("use Tie::Scalar;"); print $@;'
```

**Expected behavior**

No output is written as the `reval` call should succeed and `$@` should be empty, which is what happens on Perl 5.38.0 as well as all older versions that I tried.

**Actual behavior**

With Perl 5.39.1 and later, the `reval` call fails and `$@` is printed with the following content:

> Can't locate object method "import" via package "Tie::Scalar" at (eval 7) line 1.
> BEGIN failed--compilation aborted at (eval 7) line 1.

Note, that just `perl -e 'use Tie::Scalar'` still compiles successfully. So `use`-ing a module that does not expose an `import` method is still allowed, as per the [documentation](https://perldoc.perl.org/5.39.6/functions/use).

**Perl configuration**

I tested this on a default perl 5.39.8 built using `perlbrew install blead` under Debian (running in WSL).

See the attached [perl_version.txt](https://github.com/Perl/perl5/files/14015527/perl_version.txt) for the full output of `perl -V`.

```
Summary of my perl5 (revision 5 version 39 subversion 8) configuration:
  Snapshot of: ea7611b58ba1e09cf59415d24fe10105fe55ea8c
  Platform:
    osname=linux
    osvers=5.15.133.1-microsoft-standard-wsl2
    archname=x86_64-linux
    uname='linux mathias-2022 5.15.133.1-microsoft-standard-wsl2 #1 smp thu oct 5 21:02:42 utc 2023 x86_64 gnulinux '
    config_args='-de -Dprefix=/home/mathias/perl5/perlbrew/perls/perl-blead -Dman1dir=none -Dman3dir=none -Dusedevel -Aeval:scriptdir=/home/mathias/perl5/perlbrew/perls/perl-blead/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=undef
    usemultiplicity=undef
    use64bitint=define
[...]

Issue #17494, opened 2020-01-19
Tie::File ruins UTF-8 encoded files [rt.cpan.org #96014]


toddr:
Migrated from [rt.cpan.org#96014](https://rt.cpan.org/Ticket/Display.html?id=96014) (status was 'new')

Requestors:
* mjd@plover.com

From mjd@plover.com on 2014-05-27 21:10:00
:

It is not safe to use Tie::File to operate on a file with a multibyte encoding, such as UTF-8. This has been a defect in Tie::File since it was first written. I was aware at the time that this would be a problem, but at the time it seemed too hard to fix.  I know a little bit more now, and the problem has become considerably more urgent since 2005.  It should be fixed.  I would like to fix this, but I need some help.

The essential problem is as follows.  Tie::File holds an array that records the tell() position at which each record in the file begins, as reported by seek().  Tie::File makes the not-completely-warranted assumption that these offsets are actually byte counts, but I think this is a minor issue.

The major issue this this: Suppose that record 3 begins at tell() offset 100, and record 4 begins at tell() offset 150. In some places Tie::File then concludes from this that record 3 is 50 bytes long, which I think is generally correct.

But in other places Tie::File will calculate the length of record 3 by reading it in and calling length() on the result.  Since what is wanted is the length in bytes, this is wrong except for legacy encodings where each character is one byte.

Now suppose Tie::File is asked to replace record 3 with some string $s.  Tie::File uses length() on $s to find out how many bytes $s is, which is wrong, and it uses length() on record 3 similarly wrongly. Then it compares these two wrong lengths to decide whether record 3 can be overwritten with $s in-place, or whether the tail of the file needs to be copied upwards (if $s is shorter than the old record 3) or downwards (if it is longer).  All these calculations are being done with character lengths, but they should be done with byte lengths. Since they are done wrongly, Tie::File mangles the data file when it modifies the record.

On review, I see the following relevant ten-year-old comment:

     # length($oldrec) here is not consistent with text mode  TODO XXX BUG   

The enclosed program demonstrates the problem.  It writes a correct UTF-8 encoded file, copies it to stdout, then uses Tie::File to modify a record in the middle, then copies the resulting carbled file to stdout.

[...]

PR #21875, opened 2024-01-23
Add option to specify binmode, bump version.

brainbuz:
I encountered problems with characters above 128 and added an option to specify the binmode.

### DIFF
--- a/dist/Tie-File/lib/Tie/File.pm
+++ b/dist/Tie-File/lib/Tie/File.pm
@@ -1,7 +1,5 @@
 package Tie::File;
 
-require 5.005;
-
 use strict;
 use warnings;
 
@@ -10,15 +8,14 @@ use POSIX 'SEEK_SET';
 use Fcntl 'O_CREAT', 'O_RDWR', 'LOCK_EX', 'LOCK_SH', 'O_WRONLY', 'O_RDONLY';
 sub O_ACCMODE () { O_RDONLY | O_RDWR | O_WRONLY }
 
-
-our $VERSION = "1.07";
+our $VERSION = "1.08";
 my $DEFAULT_MEMORY_SIZE = 1<<21;    # 2 megabytes
 my $DEFAULT_AUTODEFER_THRESHHOLD = 3; # 3 records
 my $DEFAULT_AUTODEFER_FILELEN_THRESHHOLD = 65536; # 16 disk blocksful
 
-my %good_opt = map {$_ => 1, "-$_" => 1}
-                 qw(memory dw_size mode recsep discipline 
-                    autodefer autochomp autodefer_threshhold concurrent);
+my %good_opt = map { $_ => 1, "-$_" => 1 }
+  qw(memory dw_size mode recsep discipline
+  autodefer autochomp autodefer_threshhold concurrent binmode);
 
 our $DIAGNOSTIC = 0;
 our @OFF; # used as a temporary alias in some subroutines.
@@ -55,7 +52,8 @@ sub TIEARRAY {
   }
   $opts{dw_size} = $opts{memory} unless defined $opts{dw_size};
   if ($opts{dw_size} > $opts{memory}) {
-      croak("$pack: dw_size may not be larger than total memory allocation\n");
+      croak(
+        "$pack: dw_size may not be larger than total memory allocation\n");
   }
   # are we in deferred-write mode?
   $opts{defer} = 0 unless defined $opts{defer};
@@ -113,7 +111,11 @@ sub TIEARRAY {
 	$fh = Symbol::gensym();
     }
     sysopen $fh, $file, $opts{mode}, 0666 or return;
-    binmode $fh;
+    if ( $opts{binmode} ) {
+      binmode( $fh, $opts{binmode} );
+    } else {
+      binmode $fh 
+    }
     ++$opts{ourfh};
   }
   { my $ofh = select $fh; $| = 1; select $ofh } # autoflush on write
@@ -2051,6 +2053,15 @@ Changes to the array are reflected in the file immediately.
 
 Lazy people and beginners may now stop reading the manual.
 
+=head2 unicode
+
+To open a unicode encoded file, use the L<"binmode"> option or provide a 
+file handle that open with the desired encoding.
+While Tie::File can read Unicode files, it mangles them when attempting
+to write. A unicode string can consume more bytes than its' string
+length and Tie::File miscalculates non-byte oriented
+strings when writing. 
+
 =head2 C<recsep>
 
 What is a 'record'?  By default, the meaning is the same as for the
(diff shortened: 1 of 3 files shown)


dist/Tie-File/t/44_binmode.t
```diff
@@ -0,0 +1,46 @@
...
+use 5.008;
+
+# option to specify binmode
+
+use Tie::File;
+use Test::More tests => 3;
+use Test::More;
+
+use utf8;
+use feature 'unicode_strings';
```
haarg: This feature doesn't appear to be needed in this test. For now I think it could be removed.

haarg:
Could you update the commit message to note that this is a change in Tie-File?

dist/Tie-File/lib/Tie/File.pm
```diff
@@ -2128,6 +2131,16 @@ then the array C<@gifts> would appear to contain C<("Gold\n",
...
 
+=head2 C<binmode>
+
+Specify the binmode to set for files opened by Tie::File.
+
+  # make sure Perl is expecting unicode files and strings
+  use utf8;
+  use feature 'unicode_strings';
+  # tie a file with utf8 binmode
+  tie my @array, 'Tie::File', 'array.file', 'binmode' => ':utf8';
```
Grinnz: Please suggest ':encoding(UTF-8)' rather than ':utf8' particularly for reading files as the :utf8 layer can result in corrupted strings.

```suggestion
  tie my @array, 'Tie::File', 'array.file', 'binmode' => ':encoding(UTF-8)';
```

Grinnz:
The porting tests indicate dist/Tie-File/t/44_binmode.t must be added to the MANIFEST. I'm not sure if there is an automated tool for this but it should be straightforward.

jkeenan:
Assuming this p.r. is intended to fix https://github.com/Perl/perl5/issues/21867, please add that to the commit message.  Thanks.

Leont:
This commit header should probably mention `Tie::File`

brainbuz:
I switched the pod to :encoding(UTF-8), updated the commit message, also deleted the line require 5.005.
The issue James mentioned applies to a different module. I did find a related issue #17494 that I missed last night when I was looking for open issues, unfortunately this PR does not fix it, when I put this new version in the code I was working on where I discovered the problem was affected by the issue from 17494. I'll look at that issue and update this PR if I can fix it, if I can't fix it I would recommend adding a note to the doc on Tie::File that it should be considered read only on unicode files.

dist/Tie-File/lib/Tie/File.pm
```diff
@@ -2128,6 +2129,16 @@ then the array C<@gifts> would appear to contain C<("Gold\n",
 "Frankincense\n", "Myrrh\n")>, or (on Win32 systems) C<("Gold\r\n",
 "Frankincense\r\n", "Myrrh\r\n")>.
 
+=head2 C<binmode>
+
+Specify the binmode to set for files opened by Tie::File.
+
+  # make sure Perl is expecting unicode files and strings
+  use utf8;
+  use feature 'unicode_strings';
```
jkeenan: Your documentation for the `binmode` option states that to use it you have to ensure that the program "is expecting unicode file and strings", and that this is ensured by calling `use utf8; use feature 'unicode_strings';`.  However, the test program you've added to the repository, does not call `use feature 'unicode_strings';`.  Which is correct?  The documentation or the test program?

dist/Tie-File/lib/Tie/File.pm
```diff
@@ -2128,6 +2129,16 @@ then the array C<@gifts> would appear to contain C<("Gold\n",
...
+  # tie a file with UTF-8 binmode
+  tie my @array, 'Tie::File', 'array.file', 'binmode' => ':encoding(UTF-8)';
```
jkeenan: One of our porting tests, `t/porting/podcheck.t`, imposes a 78-character line length in documentation and calculates tab expansion as part of that process.  This line is slightly over that calculated maximum.  A simple way to fix that would be to break the line in two at an appropriate point, *e.g.,*:
```
diff --git a/dist/Tie-File/lib/Tie/File.pm b/dist/Tie-File/lib/Tie/File.pm
index 9b41efa08a..b45aea9685 100644
--- a/dist/Tie-File/lib/Tie/File.pm
+++ b/dist/Tie-File/lib/Tie/File.pm
@@ -2137,7 +2137,8 @@ Specify the binmode to set for files opened by Tie::File.
   use utf8;
   use feature 'unicode_strings';
   # tie a file with UTF-8 binmode
-  tie my @array, 'Tie::File', 'array.file', 'binmode' => ':encoding(UTF-8)';
+  tie my @array,
+    'Tie::File', 'array.file', 'binmode' => ':encoding(UTF-8)';
 
 =head2 C<mode>
 
```
Here, too, run `make test_porting` before re-pushing.  Thanks.

jkeenan (changes requested):
Our "porting" tests impose QA standards and have to be placated before our GH CI programs can run to completion.  Please adjust as needed.  This will facilitate evaluation of the actual code changes you're proposing.  Thanks.

dist/Tie-File/lib/Tie/File.pm
jkeenan: Also, I suggest changing the Subject of this GH pull request to be: "Tie-File: add option to specify binmode".  You can trim the first line of the commit message to simply that as well.

dist/Tie-File/lib/Tie/File.pm
brainbuz: Those 2 lines are recommended, one of the earlier comments requested I remove unicode_strings from the test because it isn't necessary for the test to work. I can make the doc clearer that these two lines are recommended.

dist/Tie-File/lib/Tie/File.pm
brainbuz: Once I delved into the open issue #17494, I determined that this pull request only works for reading files and have work in another branch https://github.com/brainbuz/perl5/tree/trytofix17494, I was able to improve behavior but writes of new data are still 1 character off.
I'd like to wait a few days to see if I can solve this problem, if I can't I'll make the minor changes and add to the doc a warning that with non-byte files Tie::File should be used read only. If I can fix it I'll merge it to this pull request.

Tux: I see *no* added value in adding "recommended" `use` statements in documentation if there is no relation to the stuff being documented.
`use utf8;` is about the source code and has no relation whatsoever to the file being tied, so it distracts from what is documented. My first though is "why do I need this?"
I'd say leave out what is not required.

brainbuz:
I'm closing this and submitting a new PR for just a documentation patch. I tried to fix 17494, without that it isn't safe to write to unicode file handles with Tie::File.

brainbuz:
I'm closing this and submitting a new PR for just a documentation patch. I tried to fix 17494, without that it isn't safe to write to unicode file handles with Tie::File.

PR #21879, opened 2024-01-24
[doc] fix perlre \K example

yuu-no:
**Where**

https://perldoc.perl.org/perlre#%5CK

**Description**

The example seems wrong.

```
 For instance

s/(foo)bar/$1/g;

can be rewritten as the much more efficient

s/foo\Kbar//g;
```

Perhaps the below is correct.

```
s/(?<=foo)bar/$1/g;
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1463,6 +1463,7 @@ Yaroslav Kuzmin                <ykuzmin@rocketsoftware.com>
 Yary Hluchan
 Yasushi Nakajima               <sey@jkc.co.jp>
 Yitzchak Scott-Thoennes        <sthoenna@efn.org>
+Yusei Ueno                     <y.say1125@gmail.com>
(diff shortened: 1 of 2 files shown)


haarg:
The original example is confusing, but this change isn't really correct.

The proposed change to `s/(?<=foo)bar/$1/g` doesn't really make sense. Look around assertions are not capturing, so `$1` will always be undef. That search and replace would produce warnings if they are enabled.

I think the intent of the original is to say that `s/(foo)bar/$1/g` could be rewritten as `s/(?<=foo)bar//g`, but that is inefficient, and would be better rewritten as `s/foo\Kbar//g`.

mauke:
What's wrong with the original example?

yuu-no:
Oh sorry, this is my misstake.

example is not wrong.

PR #21883, opened 2024-01-25
Fix #21867: Safe not able to use modules without an import method.

### LINKED ISSUE #21867: BBC: blead breaks Safe when `use`-ing modules without an `import` method.

mkende:
This is done by adding &UNIVERSAL::import and unimport to Safe default_share array following the suggestion of @haarg.

See issue #21867 for reference.

### DIFF
--- a/dist/Safe/t/safeload.t
+++ b/dist/Safe/t/safeload.t
@@ -18,7 +18,7 @@ BEGIN {
 use strict;
 use Test::More;
 use Safe;
-plan(tests => 3);
+plan(tests => 4);
 
 my $c = new Safe;
 $c->permit(qw(require caller entereval unpack rand));
@@ -35,3 +35,8 @@ $r or diag $@;
 # perl version in 5.10-.
 ok !$c->reval(q{use 5.012; $undeclared; 1}),
    'reval does not prevent use 5.012 from enabling strict';
+
+# "use Tie::Scalar" depends on UNIVERSAL::import as Tie::Scalar does not have
+# its own import method.
+$r = $c->reval(q{ use Tie::Scalar; 1 });
+ok( defined $r, "Can load Tie::Scalar.pm in a Safe compartment" ) or diag $@;
--- a/dist/Safe/Safe.pm
+++ b/dist/Safe/Safe.pm
@@ -3,7 +3,7 @@ package Safe;
 use 5.003_11;
 use Scalar::Util qw(reftype refaddr);
 
-$Safe::VERSION = "2.45";
+$Safe::VERSION = "2.46";
 
 # *** Don't declare any lexicals above this point ***
 #
@@ -77,8 +77,10 @@ my $default_root  = 0;
 my $default_share = [qw[
     *_
     &PerlIO::get_layers
+    &UNIVERSAL::import
     &UNIVERSAL::isa
     &UNIVERSAL::can
+    &UNIVERSAL::unimport
     &UNIVERSAL::VERSION
     &utf8::is_utf8
     &utf8::valid
@@ -822,4 +824,3 @@ Reworked to use the Opcode module and other changes added by Tim Bunce.
 Currently maintained by the Perl 5 Porters, <perl5-porters@perl.org>.
 
 =cut
-


dist/Safe/t/safetie.t
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
+
+use strict;
+use warnings;
+use Test::More;
+use Safe;
+plan(tests => 1);
+
+my $c = new Safe;
```
jkeenan: We've been trying to get away from indirect object notation.  Could this be rewritten as `my $c = Safe->new();`?

dist/Safe/t/safetie.t
```diff
@@ -0,0 +1,23 @@
...
+$c->deny_only();
+
+# Tie::Scalar has no "import" method, we test that it can still be used
+# within a safe.
+ok($c->reval("use Tie::Scalar; 1"));
```
jkeenan: Do we really need to create a new test file for a single unit test of this functionality?  I recommend seeing whether this could go into one of the `t/safe?.t` files.

jkeenan (changes requested):

dist/Safe/t/safetie.t
mkende: I removed this file now. Let me know if you want me to update the existing tests.

dist/Safe/t/safetie.t
mkende: Done. `t/safeload.t` seemed to be the most appropriate for this.

dist/Safe/t/safetie.t
jkeenan: > I removed this file now. Let me know if you want me to update the existing tests.

No, there are others who are more concerned with this than I am.  We'll let them follow up in a subsequent pull request. :-)

jkeenan:
While all the nits I was picking have been addressed, I'd like someone who is more familiar with `Safe`, `import` methods, etc., to have a look at this.  (I've never used `Safe.pm` myself.)  @haarg, @tonycoz, can you take a look?

tonycoz (approved):

haarg (approved):

mkende:
@jkeenan I’m not super familiar with GitHub code reviews but it seems that your review is still needed (and there is something about "requested changes" from you, but I’m not sure what given that all conversations are resolved). Can you take a look?

dist/Safe/t/safeload.t
```diff
@@ -35,3 +35,8 @@ $r or diag $@;
 # perl version in 5.10-.
 ok !$c->reval(q{use 5.012; $undeclared; 1}),
    'reval does not prevent use 5.012 from enabling strict';
+
+# "use Tie::Scalar" depends on UNIVERSAL::import has Tie::Scalar does not have
```
xenu: I think this was meant to be "as" instead of "has".

xenu (changes requested):

dist/Safe/t/safeload.t
mkende: Fixed, thanks for noticing.

tonycoz:
> @jkeenan I’m not super familiar with GitHub code reviews

It just needed someone to hit the merge button, which I've done.

PR #21884, opened 2024-01-25
locale.c: Fix ALT_DIGITS on Alpine

khwilliamson:
Alpine, for the C and equivalent locales, doesn't return anything like what the Posix Standard says it should for nl_langinfo(ALT_DIGITS). It's supposed to be a string of up to 99 semi-colon-separated values. Instead, it is a string "0123456789".  There are no separators and these are standard digits, not alternate.  This commit causes any platform that returns a standard digit when alternates are requested to instead return the empty string, which is used elsewhere, and now here, to indicate there are no alternate digits for this locale.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -6321,6 +6321,22 @@ S_langinfo_sv_i(pTHX_
             if (sep_pos) {
                 separator = *sep_pos;
             }
+            else if (strpbrk(retval, "123456789")) {
+
+                /* Alternate digits, with the possible exception of 0,
+                 * shouldn't be standard digits, so if we get any back, return
+                 * that there aren't alternate digits.  0 is an exception
+                 * because there may be locales that do not have a zero, such
+                 * as Roman numerals.  It could therefore be that alt-0 is 0,
+                 * but alt-1 better be some multi-byte Unicode character(s)
+                 * like U+2160, ROMAN NUMERAL ONE.  This clause is necessary
+                 * because the total length of the ASCII digits won't trigger
+                 * the conditional in the next clause that protects against
+                 * non-Standard libc returns, such as in Alpine platforms, but
+                 * multi-byte returns will trigger it */
+                retval = "";
+                total_len = 0;
+            }
             else if (UNLIKELY(total_len >
                                         2 * UVCHR_SKIP(PERL_UNICODE_MAX) * 4))
             {   /* But as a check against the possibility that the separator is


jkeenan:
@cjg-cguevara, Can you smoke this branch on our Alpine Linux rig?  Thanks.

khwilliamson:
I have started a smoke-me branch

locale.c
```diff
@@ -6321,6 +6321,14 @@ S_langinfo_sv_i(pTHX_
             if (sep_pos) {
                 separator = *sep_pos;
             }
+            else if (strpbrk(retval, "0123456789")) {
```
tonycoz: I could see a locale returning 0 for the first entry and more localized strings for the later values, which this will match.

Wouldn't it be better to strEQ() here?

khwilliamson: Why would such a locale exist?  All I can think of is a system that has no zero.  What would you think would be the smallest number that had to be a non-ASCII digit in order for it to be considered an alternate digit?  I think it would have to be one.

A strEQ of '"0123456789" would work for Alpine, but given the possibility that a platform can use NUL as a separator, or in Alpine's case, with no separators at all, I'm reluctant to give too much leeway for fear of it trying to read past the end of the buffer.

One possibility, if the ALT_DIGIT return begins with 0, is to call strftime with the `%O` format on a time value that is expected to return a non-0 value.  But not all strftime()s recognize that format, though it is in POSIX 2017, and the history there doesn't mention that it was a late addition.  But my Linux man page says it is "Single Unix" specification.

If the smallest number that I mentioned is indeed 1, as it would be for Roman Numerals, then we could use a 'strBEGINs(retval, "01")` which would catch the Alpine case and hopefully most others.

tonycoz: I was considering a numeric system (such as roman numerals) that have no zero, so fallback to "0" to represent it.

I also see a problem with this and the glibc workarounds: every system is paying for this, and we have limited visibility in bug reports as to what's going on.

I suspect this and the glibc issue should be detected in Configure so if we do get a bug report we can ask for `config.sh` to see what sort of brokenness is present, and the code that handles that brokenness is only compiled on the platforms that need it.

But I also realize that's a pain, both because `Configure` is a bit arcane, and because detecting the glibc issue at least requires a locale with alt_digits defined, which might not be available.

khwilliamson: It would be a big pain to write a Configure probe.  There is perl language infrastructure to find all the locales on a system, but I haven't ever seen Configure calling perl instead of shell.  And there might not be a perl at all on the system.

Even if possible, Configure could be run before such a locale was later installed.

And the penalty here is only on systems that don't follow the POSIX Standard for ALT_DIGIT returns, except that if we knew what the platform's syntax is, we wouldn't have to search the libc return for it.  But keep in mind this is only executed if we know that requested item is indeed ALT_DIGITS.

I believe the revision I just pushed handles the Alpine case while allowing for alt-0 to be zero.

I believe we only have to worry about alt-0 being zero.  alt-1 etc better not be ASCII digits

tonycoz (approved):

PR #21885, opened 2024-01-25
Run only one of mini_mktime(), mktime()

khwilliamson:
mini_mktime() is a faster, stripped down version of mktime() that ignores locale-dependent information.   But it doesn't work when the structure it fills in contains fields that are locale-related.  A patch 20 some years ago called mktime() to fill in the missing fields.

But that ends up executing similar algorithms twice. If we need the locale-related info, why not just call mktime() once and be done with it.  I figured I must be missing something, so started a smoke-me branch with this change. [https://perl.develop-help.com/?b=smoke-me%2Fkhw-mktime](url).  So far everything is passing.

There is one glitch on my Linux glibc.  mktime() did not accept 60 for the second number (a leap second) , and mini_mktime() does.  And there is a test that includes a leap second.

The right thing to do to calculate the number of seconds since the epoch is to have a table of when actual leap seconds happened, and take that into account.  I guess there haven't been enough of them to make a difference. [https://en.wikipedia.org/wiki/Leap_second](url) has an extended discussion on this.

This commit uses plain mktime() when the fields exist that are not populated by min_mktime

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -7877,22 +7877,75 @@ S_ints_to_tm(pTHX_ struct tm * mytm,
     mytm->tm_wday = wday;
     mytm->tm_yday = yday;
     mytm->tm_isdst = isdst;
+
+/* If no mktime() can't call it; and if it doesn't have the extra fields that
+ * mini_mktime() doesn't handle, no need to call it; use our faster internal
+ * mini_mktime() instead */
+#if ! defined(HAS_MKTIME) || (   ! defined(HAS_TM_TM_GMTOFF)    \
+                              && ! defined(HAS_TM_TM_ZONE))
+
     mini_mktime(mytm);
 
-    /* use libc to get the values for tm_gmtoff and tm_zone on platforms that
-     * have them [perl #18238] */
-#if  defined(HAS_MKTIME)                                      \
- && (defined(HAS_TM_TM_GMTOFF) || defined(HAS_TM_TM_ZONE))
-    struct tm mytm2 = *mytm;
+#else
+
+    /* Otherwise we have to use the (slower) libc mktime(), which does take
+     * locale into consideration. [perl #18238] */
     MKTIME_LOCK;
-    mktime(&mytm2);
+
+    (void) mktime(mytm);
+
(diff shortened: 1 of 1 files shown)


Tux:
Nice analysis!

tonycoz:
> ... this change. [https://perl.develop-help.com/?b=smoke-me%2Fkhw-mktime](url). So far ...

When you use the link creation tool in the github editor it creates a link using the highlighted text as the text of the link, *not* the url of the link. You need to replace the "url" in the produced markdown to make it a useful link.

eg. from your markdown `[https://perl.develop-help.com/?b=smoke-me%2Fkhw-mktime](url)`, which is something you've done a few times in other issues and pull requests (including the wikipedia link)

If you want the link text like https://perl.develop-help.com/?b=smoke-me%2Fkhw-mktime to just be both the text and url of a link don't bother with the link creation tool, github will linkify it for you and I can stop visiting `https://github.com/Perl/perl5/pull/url` 🙂

ext/POSIX/t/posix.t
```diff
@@ -292,7 +292,9 @@ SKIP: {
 		    `cl --version 2>&1` =~ /^.*Version\s+([\d.]+)/ and $1 >= 14)
 		or (Win32::GetOSVersion())[1] >= 6));
 
-    try_strftime("Thu Feb 29 00:00:60 1996 060", 60,0,-24, 30,1,96);
+    # This is an actual time a leap second occurred.
+    # https://en.wikipedia.org/wiki/Leap_second */
+    try_strftime("Mon Jun 30 23:59:60 1997 181", 60,59,23, 30,5,97);
```
tonycoz: strftime() works in local time, but leap seconds are added at midnight UTC, not localtime.

So this change only matters when the current time zone is UTC (Britain in winter, some parts of Africa I assume)

khwilliamson: So my analysis was flawed.  

Now updated to detect `mktime`  not handling leap seconds like `mini_mktime` does and to override with the latter

tonycoz (approved):

PR #21886, opened 2024-01-25
INSTALL: clarify usethreads/useithreads

sevan:
Add a description that they mean the same thing, in case one wondered if there was a difference (perl shipped with system is configured with useithreads, packaging system defines usethreads).

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -318,6 +318,7 @@ enable this, run
 	sh Configure -Dusethreads
 
 The default is to compile without thread support.
+Defining usethreads is synonymous with useithreads currently.
 
 Perl used to have two different internal threads implementations.  The
 current model (available internally since 5.6, and as a user-level module


tonycoz (approved):
It might be worth adding "currently", but I don't think this is ever going to change realistically.

sevan:
Done.

jkeenan:
@tonycoz, @sevan took your suggestion, but I reworded the doc slightly for better flow.  Pushed as 2a2fbe60b3.

matthewpersico:
Shouldn't "perl" be "Perl" at https://github.com/Perl/perl5/commit/2a2fbe60b3ca61da41a701fa082a961d493bc218#diff-b83b327f0b62e8168163c61f15fb0ac2c14169729e4f4184678337a15cd04decL332 ?

PR #21887, opened 2024-01-25
Bugfixes to `builtin` import and tombstone behaviour

leonerd:
* Make multiple imports idempotent, matching the behaviour of things like strict and feature flags
 * Fix various bugs to do with making previously-shadowed symbols visible again when unimporting named builtins

### DIFF
--- a/lib/builtin-unimport.t
+++ b/lib/builtin-unimport.t
@@ -34,6 +34,17 @@ no warnings 'experimental::builtin';
     }
 }
 
+# multiple imports are idempotent
+{
+    use builtin 'true';
+    use builtin 'true';
+    is(true(), "1", 'imported true() from two use lines');
+
+    no builtin 'true';
+    ok(!defined eval "true(); 1", 'true() is no longer visible after a single no');
+    like($@, qr/^Undefined subroutine &main::true called at /, 'Failure from missing function');
+}
+
 # vim: tabstop=4 shiftwidth=4 expandtab autoindent softtabstop=4
 
 done_testing();
--- a/lib/builtin-unimport.t
+++ b/lib/builtin-unimport.t
@@ -17,13 +17,13 @@ no warnings 'experimental::builtin';
 {
     package UnimportTest;
 
-    sub true() { return "true" };
+    sub true() { return "yes" };
 
     {
         use builtin 'true';
         no builtin 'true';
 
-        ::is(true(), "true", 'no builtin can remove lexical import');
+        ::is(true(), "yes", 'no builtin can remove lexical import');
     }
 
     {
--- a/lib/builtin-unimport.t
+++ b/lib/builtin-unimport.t
@@ -1,0 +1,39 @@
+#!./perl
+
+BEGIN {
+    chdir 't' if -d 't';
+    require './test.pl';
+    set_up_inc('../lib');
+}
+
+# This file mostly contains tests of the PADNAMEf_TOMBSTONE mechanism.
+# Currently the only way it is accessible from perl code is via unimport from
+# builtin.
+
+use v5.36;
+no warnings 'experimental::builtin';
+
+# imported builtins can be unexported
+{
+    package UnimportTest;
+
+    sub true() { return "true" };
+
+    {
+        use builtin 'true';
+        no builtin 'true';
+
+        ::is(true(), "true", 'no builtin can remove lexical import');
+    }
+
+    {
+        use builtin 'true';
+        { no builtin 'true'; }
+
+        ::is(true(), 1, 'no builtin is lexically scoped');
+    }
+}
+
+# vim: tabstop=4 shiftwidth=4 expandtab autoindent softtabstop=4
+
+done_testing();
--- a/lib/builtin-unimport.t
+++ b/lib/builtin-unimport.t
@@ -45,6 +45,31 @@ no warnings 'experimental::builtin';
     like($@, qr/^Undefined subroutine &main::true called at /, 'Failure from missing function');
 }
 
+# tombstone reveals earlier lexicals of the same name
+{
+   my sub true() { return "yes" }
+
+   {
+      is(true(), "yes", 'lexical true() before import+unimport');
+
+      use builtin 'true';
+      no builtin 'true';
+
+      is(true(), "yes", 'lexical true() after import+unimport');
+   }
+
+   {
+       use builtin 'true';
+       {
+           use builtin 'true';
+           no builtin 'true';
+       }
+       no builtin 'true';
+
+       is(true(), "yes", 'lexical true() after double nested import+unimport');
+   }
+}
+
 # vim: tabstop=4 shiftwidth=4 expandtab autoindent softtabstop=4
 
 done_testing();
--- a/pad.c
+++ b/pad.c
@@ -1114,6 +1114,44 @@ S_unavailable(pTHX_ PADNAME *name)
                          PNfARG(name));
 }
 
+/* Given an offset in a pad, walks backwards to return the offset of the
+ * previous pad entry that it negates.
+ */
+
+#define pad_skip_tombstone(namepv, namelen, pnl, seq, offset)  S_pad_skip_tombstone(aTHX_ namepv, namelen, pnl, seq, offset)
+STATIC PADOFFSET
+S_pad_skip_tombstone(pTHX_ const char *namepv, STRLEN namelen, const PADNAMELIST *pnl, U32 seq, PADOFFSET offset);
+STATIC PADOFFSET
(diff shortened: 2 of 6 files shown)


lib/builtin-unimport.t
```diff
@@ -0,0 +1,75 @@
...
+no warnings 'experimental::builtin';
+
+# imported builtins can be unexported
+{
+    package UnimportTest;
+
+    sub true() { return "yes" };
+
+    {
+        use builtin 'true';
```
happy-barney, line 23: you mentioned examples where `use builtin` should produce warning - this is one example

lib/builtin-unimport.t
```diff
@@ -0,0 +1,75 @@
...
+        ::is(true(), "yes", 'no builtin can remove lexical import');
+    }
+
+    {
+        use builtin 'true';
+        { no builtin 'true'; }
+
+        ::is(true(), 1, 'no builtin is lexically scoped');
+    }
+}
```
happy-barney, line 35: shouldn't `true` be unimported at block end ?
```
{
  use builtin 'true';
}
::is (true(), 'yes')
```

happy-barney (commented):
commenting only tests

pad.c
```diff
@@ -1156,8 +1194,19 @@ S_pad_findlex(pTHX_ const char *namepv, STRLEN namelen, U32 flags, const CV* cv,
...
+                    continue;
+
+                if (!PadnameIsTOMBSTONE(name))
                     break;
+
+                /* TODO(leonerd 2024-01-25): For better performance, we could
+                 * have performed this lookup at add time and just skip
+                 * immediately now, but there's currently no spare field in 
+                 * the padname struct to store the result.
+                 */
```
Ovid, line 1207: 😢 I think we've hit these issues before, where we have various internal structures in Perl which are exposed directly instead of hiding behind APIs, leading to things like this.

happy-barney: do you mean opaque types?

Tux: What if one bit in the padname indicates indirection (for the non-critical tasks) as in inode tables or UTF-8?
Or am I talking bullshit here?

happy-barney: @Tux what about extending it with common (in C word in observed) generic structure `callback_data` + `callback_cleanup` ?

Instead of callback cleanup, `name` can be extended to:
```
int symbol_table;
char * name_without_sigil;
```

so `callback_cleanup` will be provided by `symbol_table` ? (I'm already working on `symbol_table` extension, though slowly)

tonycoz: > or UTF-8?

padnames are always UTF-8.

I don't see any problems with extending the `padname` structure with extra fields, as long as we don't go wild.

The `xpadn_fieldinfo` could be turned into a union of the field info pointer and whatever other space we need if we want some "free" space unused for sub padnames.

leonerd: > I don't see any problems with extending the `padname` structure with extra fields, as long as we don't go wild.
> 
> The `xpadn_fieldinfo` could be turned into a union of the field info pointer and whatever other space we need if we want some "free" space unused for sub padnames.

Yeah that's probably going to be a fix I'll do at some point soon. I mostly didn't want the vast majority (~99%) of Padname structures to have to pay the memory cost of those unused fields.

Thinking more, at the time you create the Padname struct you always know if it's going to be for a field or a tombstone. Perhaps in that case it could actually allocate a "PadnameFull" structure with those extra fields; then the bitfllags indicate if those are valid or not. It means that normal (small) Padname structs don't need those fields, so we get some memory savings.

haarg:
Something that I'm not sure about is the following:
```
sub true { "true" }

use builtin qw(true);

{
  use builtin qw(true);
  no builtin qw(true);

  print true(), "\n";
}
```
This will print `true`, which means the `use`/`no` pair didn't return to the previous state.

leonerd:
> This will print `true`, which means the `use`/`no` pair didn't return to the previous state.

Mmmm. Yes that's a fun one indeed. But compare it to features, strict or warnings flags:

```
$ perl -e 'use feature "say"; { use feature "say"; no feature "say"; say "Boo" }'
String found where operator expected (Do you need to predeclare "say"?) at -e line 1, near "say "Boo""

$ perl -ce 'use strict; { use strict; no strict; @{"main::ISA"} = (1); }'
-e syntax OK

$ perl -e 'use warnings; { use warnings; no warnings; 1 + undef }'
```

In each of these cases, the things are simple on/off switches, they don't have activation counts. One `no` is enough to undo within that scope the effect of any number of preceeding `use`s.

It's hard to correctly see how to simultaneously make both of these tests pass:

```perl
sub true { "yes" }
use builtin 'true';  # <-- THIS LINE
{
  use builtin 'true'; no builtin 'true';
  is( true(), "1", 'true is builtin' );
}
```

```perl
sub true { "yes" }
{
  use builtin 'true';  # <-- THIS LINE
  use builtin 'true'; no builtin 'true';
  is( true(), "yes", 'true is package sub' );
}
```

The only thing different is the scope position of the marked line. Locally from the perspective of the `is()` call, the same things are visible in its scope in both cases, so why should it see two different behaviours.

A question we have to ask is are both of these tests correct? Or is only one of them correct and the other *should* fail? Which one?

By comparison to feature / strict / warning flags, it is at least consistent to say that lexical imports of builtins don't have a counting behaviour, and thus multiple activations can be cancelled by a single deäctivation. It may not be exactly what is expected in all cases, but it's the same as the other things that we already have.

happy-barney:
re multiple `use builtin 'true'` ... behaviour can be considered wrong from two points of view:
- second `use` should produce warning `redefine` if this usage of `use` is stackable (ie, treating as a function)
- second `use` should be noop if `use` treated as declaration of expected state

From my perception of programming, `use <pragma>` should be treated as declarative state of what is expected how language should behave in scope of validity of given pragma.

lib/builtin-unimport.t
leonerd: Warning implemented - see #21915

lib/builtin-unimport.t
leonerd: Yes, that's correct. That's how scopes work. Are you suggesting I should add a line of test?

happy-barney: yes, that will be nice
question raised in meanwhile: should it return to previously declared "sub true" as well ?

haarg:
> In each of these cases, the things are simple on/off switches, they don't have activation counts. One `no` is enough to undo within that scope the effect of any number of preceeding `use`s.

I think "no builtin" only really makes sense if we think of it as just removing a lexical. This means it should result in either the previous lexical being unmasked or no lexical being available. So this is in a way "counting".

> It's hard to correctly see how to simultaneously make both of these tests pass:
> 
> ```perl
> sub true { "yes" }
> use builtin 'true';  # <-- THIS LINE
> {
>   use builtin 'true'; no builtin 'true';
>   is( true(), "1", 'true is builtin' );
> }
> ```
> 
> ```perl
> sub true { "yes" }
> {
>   use builtin 'true';  # <-- THIS LINE
>   use builtin 'true'; no builtin 'true';
>   is( true(), "yes", 'true is package sub' );
> }
> ```

The second of these should fail. true should also be returning "1". And it should also issue a warning about masking an existing lexical sub.

`use builtin 'true';` should be exactly equivalent to `my sub true () { !!1 }`. This means that doing it multiple times in the same scope should warn. That would also mean `use v5.40; use builtin 'true';` would warn.

If that seems wrong, then I think lexical exports are the wrong tool for this.

leonerd:
This entire PR and the wider "tombstone" mechanism are now being abandoned, in favour of the simpler model of "you can't get rid of lexicals". Which means `use VERSION` will have to interact slightly differently with builtin bundles. This is the new model implemented by #22002.

Once that and the removal of `builtin::unimport` have been applied, this PR can be closed.

jkeenan:
1. For those who haven't been following along ... what is the "tombstone" behavior that is being eliminated?  That term doesn't appear in our documentation.
2. Branch now has merge conflicts.

leonerd:
> 1. For those who haven't been following along ... what is the "tombstone" behavior that is being eliminated?  That term doesn't appear in our documentation.

It's all really new, having just been created to attempt to implement `no builtin ...`, which we're about to remove anyway. A short-lived experiment that has already been decided failed.

>     2. Branch now has merge conflicts.

Yup but don't worry, I'm going to delete it anyway.

happy-barney:
@leonerd preventing multiple `use VERSION` will break possibility gradually upgrade parts of code.

Why to even install builtins as functions? wouldn't it be easier to:
- `use builtin X` will just turn on flag for X
- `toke.c` will return X as keyword only if flag for X is enabled

(I already posted poc for data structure for "unlimited" number of feature flags ~3 years ago - https://pastebin.com/Lnuq3Nec)

leonerd:
This can all be abandoned now. We've already removed all the tombstone related logic from builtin and `use VERSION`, and #22063 will remove the entire concept.

Issue #21880, opened 2024-01-24
ext/XS-APItest/t/locale.t failure in perl 5.39.7 on Windows 8.1 using gcc-13.2.0


twata1:
**Description**

I have been building Perl 5.39.4 and 5.39.6 on Windows 8.1 64 bit using gcc version 13.2.0 (MinGW-W64 x86_64-msvcrt-posix-seh, built by Brecht Sanders), which can be found on https://winlibs.com/ .
The tests have passed.

This time I am building Perl 5.39.7.
Several types of warnings are output (e.g. `ignoring '#pragma intrinsic ' [-Wunknown-pragmas]`,  `unused parameter 'piPerl' [-Wunused-parameter]`, etc.), but the compilation itself is successful.

When I run `gmake test`, I get a message box at `../ext/XS-APItest/t/locale.t` that says "perl.exe has stopped working".
I have also run `.\perl harness ../ext/XS-APItest/t/locale.t` as follows with similar results.

```
C:\Perl64-gcc13\perl-5.39.7\t>.\perl harness ../ext/XS-APItest/t/locale.t
../ext/XS-APItest/t/locale.t .. Dubious, test returned 255 (wstat 65280, 0xff00)
All 63 subtests passed

Test Summary Report
-------------------
../ext/XS-APItest/t/locale.t (Wstat: 65280 (exited 255) Tests: 63 Failed: 0)
  Non-zero exit status: 255
  Parse errors: No plan found in TAP output
Files=1, Tests=63, 18 wallclock secs ( 0.00 usr +  0.00 sys =  0.00 CPU)
Result: FAIL
Finished test run at Wed Jan 24 01:00:16 2024.

C:\Perl64-gcc13\perl-5.39.7\t>
```

The results of `mode con` are as follows.
```
C:\Perl64-gcc13\perl-5.39.7\t>mode con

Status for device CON:
----------------------
    Lines:          9999
    Columns:        130
    Keyboard rate:  31
    Keyboard delay: 1
    Code page:      1252


C:\Perl64-gcc13\perl-5.39.7\t>
```

**Steps to Reproduce**

Download perl-5.39.7.tar.gz from metacpan (https://metacpan.org/release/CORION/perl-5.39.7).
Extract from perl-5.39.7.tar.gz to C:\Perl64-gcc13.
Launch Command Prompt.
On that Command Prompt, run the following.
_(Note: I use gmake.exe, which is a renamed mingw32-make.exe.)_

```
[...]

PR #21891, opened 2024-01-27
locale.c: Add missing check for error return

khwilliamson:
When strftime() fails, it returns NULL; the check for this was missing, which led to a segfault on Windows that didn't appear on other platforms because the particular test worked on other platforms.

Fixes GH #21880

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -7232,6 +7232,12 @@ S_emulate_langinfo(pTHX_ const int item,
 
         restore_toggled_locale_c(LC_TIME, orig_TIME_locale);
 
+        /* If strftime() returned an error, we return an empty string */
+        if (! temp) {
+            retval = "";
+            break;
+        }
+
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -7232,6 +7232,11 @@ S_emulate_langinfo(pTHX_ const int item,
 
         restore_toggled_locale_c(LC_TIME, orig_TIME_locale);
 
+        /* If strftime() returned an error, we return an empty string */
+        if (! temp) {
+            retval = "";
```
mauke, line 7237: This doesn't actually return. It keeps running the following blocks, which want to dereference `temp`.
Missing `break;` here?

khwilliamson: The `break` got lost between the box I tested with and the box I submitted with, and I'm embarrassed I didn't notice

tonycoz (approved):

PR #21894, opened 2024-01-27
locale.c: Fix bug when -DNO_LOCALE_TIME

khwilliamson:
The logic for these preprocessor conditionals is wrong.  We must change LC_CTYPE to match LC_TIME (if the platform requires it) even when LC_TIME is confined to be the C locale, as happens when -DNO_LOCALE_TIME is in effect in a Configuration.

But doing so allows for simplification and hence the complete removal of these preprocessor conditionals, as the logic is already available in the DEALING macros that this commit replaces the previous code with

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -388,7 +388,8 @@ static int debug_initialization = 0;
  * looking for.  XXX This still presumes that we have to match COLLATE and
  * CTYPE even on platforms that apparently handle this. */
 #if defined(USE_LOCALE_CTYPE) && ! defined(LIBC_HANDLES_MISMATCHED_CTYPE)
-#  define WE_MUST_DEAL_WITH_MISMATCHED_CTYPE
+#  define WE_MUST_DEAL_WITH_MISMATCHED_CTYPE    /* no longer used; kept for
+                                                   possible future use */
 #  define start_DEALING_WITH_MISMATCHED_CTYPE(locale)                          \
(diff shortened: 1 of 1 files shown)


jkeenan:
In your p.r. you are deleting two statements that include `WE_MUST_DEAL_WITH_MISMATCHED_CTYPE` (lines 7900 and 8010 in blead).  Why then do we still retain the definition of that term at line 391?
```
$ grep -n WE_MUST_DEAL_WITH_MISMATCHED_CTYPE locale.c
391:#  define WE_MUST_DEAL_WITH_MISMATCHED_CTYPE
7900:#  if defined(WE_MUST_DEAL_WITH_MISMATCHED_CTYPE) && defined(USE_LOCALE_TIME)
8010:#  if defined(WE_MUST_DEAL_WITH_MISMATCHED_CTYPE) && defined(USE_LOCALE_TIME)
```

khwilliamson:
Yes that `#define` can be removed, as it is no longer used; but I'm inclined to keep it in in case a situation arises where it is needed.  Unused C-language macro definitions have no effect on runtime size or performance.

jkeenan:
> Yes that `#define` can be removed, as it is no longer used; but I'm inclined to keep it in in case a situation arises where it is needed. 

If that is the case, then I would recommend adding a comment to that effect.

> Unused C-language macro definitions have no effect on runtime size or performance.

khwilliamson:
Comment added

PR #21896, opened 2024-01-27
Langinfo.t: Set locale categories if no LC_ALL

khwilliamson:
This is a minor thing, but all these Linux extra categories need to be set to "C" for testing, when LC_ALL is not available on the platform.

This was overlooked in 3c45ca2e154f5fdc7ce2a0fdf51b192da3a1ae98

### DIFF
--- a/ext/I18N-Langinfo/t/Langinfo.t
+++ b/ext/I18N-Langinfo/t/Langinfo.t
@@ -204,10 +204,16 @@ if (locales_enabled('LC_ALL')) {
     setlocale(LC_ALL, "C");
 }
 else { # If no LC_ALL, make sure the categories used in Langinfo are in C
-    setlocale(LC_CTYPE, "C")    if locales_enabled('LC_CTYPE');
-    setlocale(LC_MONETARY, "C") if locales_enabled('LC_MONETARY');
-    setlocale(LC_NUMERIC, "C")  if locales_enabled('LC_NUMERIC');
-    setlocale(LC_TIME, "C")     if locales_enabled('LC_TIME');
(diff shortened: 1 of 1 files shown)


jkeenan:
> This is a minor thing, but all these Linux extra categories need to be set to "C" for testing, when LC_ALL is not available on the platform.
> 
> This was overlooked in [3c45ca2](https://github.com/Perl/perl5/commit/3c45ca2e154f5fdc7ce2a0fdf51b192da3a1ae98)

You say that this patch is relevant "when LC_ALL is not available on the platform".  By that do you mean, *when LC_ALL has no value shown when I call `locale`*?  Or do you mean *when LC_ALL does not even appear when I call `locale`?

I ask because on each of Linux, FreeBSD and OpenBSD, calling `locale` for me yields:
```
LC_ALL=
```
... (*i.e.,* the first case above).

khwilliamson:
One has to hack things internally in order to create a system where LC_ALL is not available.  The code to handle this was originally to try to workaround perl not understanding how to deal with platforms that represent LC_ALL differently than what is done on Linux and Windows.  *BSD boxes are examples.  But it turns out that that workaround was buggy, and now perl does handle those boxes without needing the workaround.

Still, one might want to have perl avoid the use of LC_ALL in case there are bugs in the platform's handling of it.  AIX earlier versions have such a bug, and for 5.38 perl avoided a failing test because of it.  Now, there is a workaround in place that doesn't involve turning off LC_ALL.  I have an understanding with @tonycoz to keep LC_ALL emulation intact until the locale handling stabilizes (which we are close to).  And then we can remove it, maybe for 5.40, in one commit that can be reverted if necessary.

tonycoz (approved):

PR #21898, opened 2024-01-28
Add basic example of qx// on perlop.pod

jidanni:
The page never gives a basic example. Only non-basic examples.

Yes, date will look different per locale. Please use a better example if you want.

### DIFF
--- a/pod/perlop.pod
+++ b/pod/perlop.pod
@@ -2467,6 +2467,8 @@ list of lines (however you've defined lines with C<$/> or
 C<$INPUT_RECORD_SEPARATOR>), or an empty list if the shell (or command)
 could not be started.
 
+    print qx/date/; # prints "Sun Jan 28 06:16:19 CST 2024"
+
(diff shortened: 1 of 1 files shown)


pod/perlop.pod
```diff
@@ -2467,6 +2467,8 @@ list of lines (however you've defined lines with C<$/> or
 C<$INPUT_RECORD_SEPARATOR>), or an empty list if the shell (or command)
 could not be started.
 
+    print qx/date/; # prints "Sun Jan 28 06:16:19 CST 2024"
```
jkeenan, line 2470: This section of `perlop` documents both the `qx` operator and its backticks variant.  If we're going to add an example of the first, then perhaps we should also add one of the second:
```
    print `date`; # also invokes the 'date' command
```

jidanni: Yes, be my guest, the more examples the merrier! Thanks.

PR #21899, opened 2024-01-28
State clearly /bin/sh cannot be changed, in perlop.pod

jidanni:
Else end users will search for hours for the variable that controls this.

https://stackoverflow.com/questions/55010798/temporarily-change-default-shell-inside-a-perl-system-call

### DIFF
--- a/pod/perlop.pod
+++ b/pod/perlop.pod
@@ -2456,7 +2456,7 @@ X<qx> X<`> X<``> X<backtick>
 =item C<`I<STRING>`>
 
 A string which is (possibly) interpolated and then executed as a
-system command, via F</bin/sh> or its equivalent if required.  Shell
+system command, via F</bin/sh> (not changable) or its equivalent if required.  Shell
 wildcards, pipes, and redirections will be honored.  Similarly to
 C<system>, if the string contains no shell metacharacters then it will
 executed directly.  The collected standard output of the command is


pod/perlop.pod
```diff
@@ -2456,7 +2456,7 @@ X<qx> X<`> X<``> X<backtick>
 =item C<`I<STRING>`>
 
 A string which is (possibly) interpolated and then executed as a
-system command, via F</bin/sh> or its equivalent if required.  Shell
+system command, via F</bin/sh> (not changable) or its equivalent if required.  Shell
```
jkeenan, line 2459: `s/changable/changeable/`

jidanni: I actually looked that up several times first I thought it needed an i, and I almost got it right. Well please go ahead and fix it for me because I'll  only screw it up more thanks!

Leont:
I think this would be confusing even if it had been true, but it's also not true. You absolutely can override what it does (though I wouldn't recommend it).

```
perl -e 'BEGIN { *CORE::GLOBAL::readpipe = sub { die } }; print `echo hi`'
```

jidanni:
I'm very glad that it is changeable after all.

Some people might want to use a different shelf for some reason, without using e g., bash -c ... for the one line command of a shell.

Anyway I'm trying to say is the man page says /bin/sh and many users would be curious "well how can I change that if I wanted to?" But they just can't figure it out. I'm glad you figured it out even though I can't figure out what that line you posted means... I'm not really sure what shell it is using. But anyways I'm just hoping that the man page will provide a few clues. The first thing users usually would do is look at the perlvar man page and see if the SHELL environment variable is the thing that can control it. Well of course that doesn't work, probably for security reasons. But my point is users are going to be curious and they can't figure it out. So there should be some kind of hint about it somehow. Thanks.

And of course if you mean via recompiling perl, well then the man page might say it cannot be changed except by recompiling...

mauke:
@jidanni 
>  I'm glad you figured it out even though I can't figure out what that line you posted means... I'm not really sure what shell it is using.

Backticks can also be spelled as `readpipe` (a built-in function). (Try `print readpipe("date")`, for example.) The `readpipe` builtin can be overridden, either locally in a package:
```perl
use subs qw(readpipe);
sub readpipe :prototype(_) { ... }
print `echo hi`;  # actually calls readpipe("echo hi")
```
... or globally everywhere (by assigning to `*CORE::GLOBAL::readpipe`).

Since backticks just turn into a call to `readpipe()`, you can do whatever you want in the function body, including not using any shell at all (e.g. Leont's example just throws an exception).

jidanni:
Well that's all very well and good, and over my head.

I'm just thinking there should be a practical example somewhere, not necessarily on this man page but on some other man page.

Say the user likes the bash '<<<' operator, not available in the Bourne shell.

So the example could show how to use could switch the shell, without using the obvious bash -c.

PR #21902, opened 2024-01-28
[doc] `perldelta`: Minor template fixes

rwp0:
- http -> https links
- Linking modules and sections
- Apply formatting

### DIFF
--- a/Porting/perldelta_template.pod
+++ b/Porting/perldelta_template.pod
@@ -58,7 +58,7 @@ future release, and will at that time need to be installed from CPAN.
 Distributions on CPAN which require these modules will need to list them as
 prerequisites.
 
-The core versions of these modules will now issue C<"deprecated">-category
+The core versions of these modules will now issue C<deprecated>-category
 warnings to alert you to this fact.  To silence these deprecation warnings,
 install the modules in question from CPAN.
 
@@ -96,7 +96,7 @@ XXX
 =head1 Modules and Pragmata
 
 XXX All changes to installed files in F<cpan/>, F<dist/>, F<ext/> and F<lib/>
-go here.  If Module::CoreList is updated, generate an initial draft of the
+go here.  If L<Module::CoreList> is updated, generate an initial draft of the
 following sections using F<Porting/corelist-perldelta.pl>.  A paragraph summary
 for important changes should then be added by hand.  In an ideal world,
 dual-life modules would have a F<Changes> file that could be cribbed.
@@ -115,7 +115,7 @@ release manager will have to investigate the situation carefully.)
 
 =item *
 
-XXX Remove this section if Porting/corelist-perldelta.pl did not add any content here.
+XXX Remove this section if F<Porting/corelist-perldelta.pl> did not add any content here.
 
 =back
 
@@ -137,7 +137,7 @@ XXX If there was something important to note about this change, include that her
 
 =item *
 
-XXX Remove this section if Porting/corelist-perldelta.pl did not add any content here.
+XXX Remove this section if F<Porting/corelist-perldelta.pl> did not add any content here.
 
 =back
 
@@ -187,8 +187,8 @@ include any changes in L<perldiag> that reconcile it to the C<C> code.
 
 =head2 New Diagnostics
 
-XXX Newly added diagnostic messages go under here, separated into New Errors
-and New Warnings
+XXX Newly added diagnostic messages go under here, separated into L</New Errors>
+and L</New Warnings>
 
 =head3 New Errors
 
@@ -393,7 +393,7 @@ XXX
 =item *
 
 XXX Add anything here that we forgot to add, or were mistaken about, in
-the perldelta of a previous release.
+the F<perldelta> of a previous release.
 
 =back
 
@@ -412,7 +412,7 @@ XXX Generate this with:
 
 If you find what you think is a bug, you might check the perl bug database
 at L<https://github.com/Perl/perl5/issues>.  There may also be information at
-L<http://www.perl.org/>, the Perl Home Page.
+L<https://www.perl.org/>, the Perl Home Page.
 
 If you believe you have an unreported bug, please open an issue at
 L<https://github.com/Perl/perl5/issues>.  Be sure to trim your bug down to a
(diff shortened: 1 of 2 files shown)


rwp0:
Finally managed to make a cleaner PR.

Please ignore the previous ones (which I closed).

pod/perldelta.pod
```diff
@@ -58,7 +58,7 @@ future release, and will at that time need to be installed from CPAN.
 Distributions on CPAN which require these modules will need to list them as
 prerequisites.
 
-The core versions of these modules will now issue C<"deprecated">-category
+The core versions of these modules will now issue C<deprecated>-category
```
rwp0, line 61: "" in addition to C<> seems redundant here

pod/perldelta.pod
```diff
@@ -96,7 +96,7 @@ XXX
 =head1 Modules and Pragmata
 
 XXX All changes to installed files in F<cpan/>, F<dist/>, F<ext/> and F<lib/>
-go here.  If Module::CoreList is updated, generate an initial draft of the
+go here.  If L<Module::CoreList> is updated, generate an initial draft of the
```
rwp0, line 99: Link to the module so it's documentation could be accesses

pod/perldelta.pod
```diff
@@ -115,7 +115,7 @@ release manager will have to investigate the situation carefully.)
 
 =item *
 
-XXX Remove this section if Porting/corelist-perldelta.pl did not add any content here.
+XXX Remove this section if F<Porting/corelist-perldelta.pl> did not add any content here.
```
rwp0, line 118: Mark it a file as with previous paragraph

pod/perldelta.pod
```diff
@@ -142,7 +142,7 @@ before worked properly in Perl.
...
```
rwp0, line 145: Ditto

pod/perldelta.pod
```diff
@@ -192,8 +192,8 @@ include any changes in L<perldiag> that reconcile it to the C<C> code.
 
 =head2 New Diagnostics
 
-XXX Newly added diagnostic messages go under here, separated into New Errors
-and New Warnings
+XXX Newly added diagnostic messages go under here, separated into L</New Errors>
+and L</New Warnings>
```
rwp0, line 196: Link to the sections, a practice done previously within the file with other sections too

pod/perldelta.pod
```diff
@@ -362,7 +362,7 @@ XXX
 XXX Important bug fixes in the core language are summarized here.  Bug fixes in
 files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
-XXX Include references to GitHub issues and PRs as: [GH #12345] and the release
+XXX Include references to GitHub issues and PR's as: I<[GH #12345]> and the release
```
rwp0: Emphasize the GH part, add ' to PR to better indicate the plural

pod/perldelta.pod
```diff
@@ -398,7 +398,7 @@ XXX
 =item *
 
 XXX Add anything here that we forgot to add, or were mistaken about, in
-the perldelta of a previous release.
+the F<perldelta> of a previous release.
```
rwp0, line 401: perldelta's are files, and each release has a file called `perldelta` by default (without version infix)

pod/perldelta.pod
```diff
@@ -417,7 +417,7 @@ XXX Generate this with:
 
 If you find what you think is a bug, you might check the perl bug database
 at L<https://github.com/Perl/perl5/issues>.  There may also be information at
-L<http://www.perl.org/>, the Perl Home Page.
+L<https://www.perl.org/>, the Perl Home Page.
```
rwp0, line 420: unfortunately all previous delta's went with http://, so fix this in the general template to affect all future perldeltas

rwp0 (commented):
Made reviews on lines to clarify my points for modification

Grinnz:
Can you also apply these changes to Porting/perldelta_template.pod?

pod/perldelta.pod
Grinnz: Apostrophes indicate possessive, not plural

rwp0:
@Grinnz thanks, that's done

pod/perldelta.pod
khwilliamson: People sometimes use apostrophes in weird situations to indicate plural; I'm neutral about the practice.  In this case, I have a slight preference to not add it.

I however would like other feedback about italicizing the GH references.  I don't think it's worth the bother.

rwp0: > I however would like other feedback about italicizing the GH references. I don't think it's worth the bother.

What about applying C<> instead?

rwp0: > Apostrophes indicate possessive, not plural

Hmm, could use `-` here instead, it'd at least help separate the abbreviation

Grinnz: Normally these tags get turned into links in the final release perldelta. Adding any other formatting here may encourage people to add that formatting when writing the perldelta and confuse things, so I would also side on omitting the formatting here.

hvds: > People sometimes use apostrophes in weird situations to indicate plural; I'm neutral about the practice. In this case, I have a slight preference to not add it.

FWIW I've noticed fonts in which it is very hard to see that the 's' at the end of a pluralized abbreviation is lower-case, so it reads as PRS. An additional indication that a plural was intended could reduce confusion in such situations.

As far as formal correctness is concerned, I think classically the plural of PR (pull request) would be PR (pull requests).

> What about applying C<> instead?

That sounds like an excellent alternative.

Porting/perldelta_template.pod
```diff
@@ -357,7 +357,7 @@ XXX
 XXX Important bug fixes in the core language are summarized here.  Bug fixes in
 files in F<ext/> and F<lib/> are best summarized in L</Modules and Pragmata>.
 
-XXX Include references to GitHub issues and PRs as: [GH #12345] and the release
+XXX Include references to GitHub issues and PR's as: I<[GH #12345]> and the release
```
jkeenan: I, too, believe italicizing the GH references is unnecessary (here and below).

jkeenan (changes requested):

rwp0:
Thanks for the feedback.

Reverted the changes concerning the apostrophe and formatting in regards to PR notation.

Please merge if all is good now.

rwp0:
Not sure why reviews show a red X when I resolved the review myself. @jkeenan could you please re-review?

PR #21906, opened 2024-01-29
[inline.h] Silence "[-Wunknown-pragmas]" warnings in MinGW builds.

sisyphus:
These warnings appeared first in perl-5.39.7, as a result of https://github.com/Perl/perl5/pull/21842.
MinGW does not recognize the "intrinsic" pragma, so it's simplest just to hide this pragma from MinGW's view.


khwilliamson (commented):
LGTM

PR #21910, opened 2024-01-30
XS::APItest: don't allow a NULL AV * through

tonycoz:
e6c95c59c changed how `av` is initialised in multicall_return, leaving it possible for it to be NULL after the switch.

Of course the values of "context" for that are invalid, so throw an error.  This should also resolve this for Coverity and perhaps other static analysis.

cid 480208


khwilliamson (approved):

richardleach:
Thanks Tony

PR #21915, opened 2024-01-31
Warn on lexical sub shadowing

leonerd:
Adds a new warning that prints (in the `shadow` category) whenever a new lexical sub is added whose name shadows over an existing package sub. This warning is useful in cases involving lexical import - especially of things like `use builtin ...` but is likely beneficial to know in other situations too.

Examples:

```
$ ./perl -Mwarnings -e 'sub true { "ok" }; use builtin "true";'
Lexical subroutine &true masks previously declared package subroutine at -e line 1.

$ ./perl -Mwarnings -e 'use constant true => 1; use builtin "true";'
Lexical subroutine &true masks previously declared package subroutine at -e line 1.
```

In particular this now helps warn users of a particular surprise they may encounter:

```
$ ./perl -Ilib -e 'use warnings; use Test::Deep; use builtin ":5.39";'
Lexical subroutine &blessed masks previously declared package subroutine at -e line 1.
Lexical subroutine &reftype masks previously declared package subroutine at -e line 1.
```

### DIFF
--- a/pad.c
+++ b/pad.c
@@ -954,6 +954,15 @@ S_pad_check_dup(pTHX_ PADNAME *name, U32 flags, const HV *ourstash)
             --off;
         }
     }
+    /* check the package for my sub &x vs. sub PACKAGE::x collisions */
+    if (PadnamePV(name)[0] == '&') {
+        /* PadnamePV is always in UTF-8 */
+        CV *cv = get_cvn_flags(PadnamePV(name)+1, PadnameLEN(name)-1, GV_NOTQUAL|SVf_UTF8);
+        if(cv)
+            warner(packWARN(WARN_SHADOW),
+                "Lexical subroutine %" PNf " masks previously declared package subroutine",
+                PNfARG(name));
+    }
 }
 
 
--- a/lib/builtin.t
+++ b/lib/builtin.t
@@ -276,6 +276,7 @@ package FetchStoreCounter {
 # imported builtins can be unexported
 {
     package UnimportTest;
+    no warnings 'shadow';
 
     sub true() { return "true" };
 
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -239,7 +239,11 @@ XXX L<message|perldiag/"message">
 
 =item *
 
-XXX L<message|perldiag/"message">
+L<Lexical subroutine %s masks previously declared package subroutine|perldiag/"Lexical subroutine %s masks previously declared package subroutine">
+
+The current package already contains a subroutine whose name matches that of
+a newly-declared lexical or lexically imported subroutine.  The latter will
+take precedence and make the package one inaccessible by name.
 
 =back
 
--- a/pod/perldiag.pod
+++ b/pod/perldiag.pod
@@ -3573,6 +3573,12 @@ The number of items in a hash can be obtained by doing:
 
     scalar(keys %hash);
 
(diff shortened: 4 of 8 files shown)


leonerd:
It should be noted however that this warning still does not save you from shadowing in the other order: Namely, if a lexical sub is created (or imported) first, and the a package sub is imported after it, no warning is printed.

```
$ ./perl -Ilib -ce 'use warnings; use builtin ":5.39"; use Test::Deep'
-e syntax OK
```

It is *much* harder (impossible?) to even implement such a warning, because the creation of new package subs in this case happens during the `Exporter` behaviour, and when it runs it isn't even aware of, or has access to, the caller's lexical scope, so it could not check for colliding names.

This PR does not attempt to address that particular side of the problem; it can be attacked at a later time.

leonerd:
It is likely that this change will cause collateral damage on #21887; which may have to be adjusted, perhaps with the addition of some `no warnings 'shadow'` lines.

pad.c
```diff
@@ -954,6 +954,14 @@ S_pad_check_dup(pTHX_ PADNAME *name, U32 flags, const HV *ourstash)
             --off;
         }
     }
+    /* check the package for my sub &x vs. sub PACKAGE::x collisions */
+    if (PadnamePV(name)[0] == '&') {
+        CV *cv = get_cvn_flags(PadnamePV(name)+1, PadnameLEN(name)-1, GV_NOTQUAL);
```
tonycoz: padnames are always UTF8, this needs the `SVf_UTF8` flag:
```
tony@venus:.../git/perl6$ git log -n1 --pretty=oneline
5db4884c69252a594467022d9fd3fcc1b79af84b (HEAD -> 21915-lexical-shadow) perldelta for new lexical sub warning
tony@venus:.../git/perl6$ cat ../21915.pl
use utf8;
use strict;
use warnings;

sub π {}
sub pi {}
my sub π {}
my sub pi {}
tony@venus:.../git/perl6$ ./perl -Ilib ../21915.pl
Lexical subroutine &pi masks previously declared package subroutine at ../21915.pl line 8.
tony@venus:.../git/perl6$
```

tonycoz:
I don't see any issues beyond the UTF-8 one.

pad.c
leonerd: Fixed and forcepushed. Your example now warns on both for me.

tonycoz:
It could use a test for shadowing on a UTF-8 name.

leonerd:
> It could use a test for shadowing on a UTF-8 name.

Test added, re-pushed.

tonycoz (approved):

leonerd:
Discussing this in PSC meeting, on further observation it feels like we might want to revert this commit.

It is causing various test collateral damage (e.g. #21995).

It only warns in this one specific order - package _then_ lexical; it would not warn in the arguably much-more-likely case of lexical then package (e.g. because of `use VERSION` implicit imports). So it's not very useful to users.

The analogous case using variables does not cause a warning (other than the 'once' warning):

```
$ perl -Mwarnings -M-warnings=once -cE '$main::var = 123; my $var = 456'
-e syntax OK

$ perl -ce'use strict; use warnings; use vars q[$foo]; my $foo = 1;'
-e syntax OK
```

All of these reasons suggest that maybe this wants to be removed. More discussion may be required.

Issue #21916, opened 2024-02-01
[WIN32] silence warnings emitted by doio.c during mingw-w64 builds.


sisyphus:
As of perl-5.39.7, doio.c is emitting the following warnings when building perl with a mingw-w64 port of gcc-13.2.0
```
..\doio.c: In function 'Perl_setfd_cloexec':
..\doio.c:64:24: warning: unused parameter 'fd' [-Wunused-parameter]
   64 | Perl_setfd_cloexec(int fd)
      |                    ~~~~^~
..\doio.c: In function 'Perl_setfd_inhexec':
..\doio.c:73:24: warning: unused parameter 'fd' [-Wunused-parameter]
   73 | Perl_setfd_inhexec(int fd)
      |                    ~~~~^~
..\doio.c: In function 'Perl_cando':
..\doio.c:2940:36: warning: unused parameter 'effective' [-Wunused-parameter]
 2940 | Perl_cando(pTHX_ Mode_t mode, bool effective, const Stat_t *statbufp)
      |                                    ^
..\doio.c: At top level:
..\doio.c:3002:1: warning: 'S_ingroup' defined but not used [-Wunused-function]
 3002 | S_ingroup(pTHX_ Gid_t testgid, bool effective)
      | ^~~~~~~~~
```
This patch silences the first 3 warnings - and perl-5.39.7 still builds and tests fine for me on Windows 11:
```
--- doio.c_orig	2024-01-15 03:40:35.000000000 +1100
+++ doio.c	2024-02-01 22:41:43.055727300 +1100
@@ -66,6 +66,8 @@
     assert(fd >= 0);
 #if defined(HAS_FCNTL) && defined(F_SETFD) && defined(FD_CLOEXEC)
     (void) fcntl(fd, F_SETFD, FD_CLOEXEC);
+#elif !defined(DEBUGGING)
+    PERL_UNUSED_ARG(fd);
 #endif
 }
 
@@ -75,6 +77,8 @@
     assert(fd >= 0);
 #if defined(HAS_FCNTL) && defined(F_SETFD) && defined(FD_CLOEXEC)
     (void) fcntl(fd, F_SETFD, 0);
+#elif !defined(DEBUGGING)
+    PERL_UNUSED_ARG(fd);
 #endif
 }
 
@@ -2967,6 +2971,9 @@
      /* Atari stat() does pretty much the same thing. we set x_bit_set_in_stat
       * too so it will actually look into the files for magic numbers
       */
[...]

PR #21918, opened 2024-02-02
Silence some warnings in doio.c

sisyphus:
This PR aims to silence the first 3 of the 4 warnings presented in https://github.com/Perl/perl5/issues/21916.
See that Issue for details and discussion ... or discuss here if that's preferable.
At this stage there is no fix to silence the 4th warning.
Feel free to merge this PR anyway, if it looks ok.

### DIFF
--- a/doio.c
+++ b/doio.c
@@ -66,6 +66,8 @@ Perl_setfd_cloexec(int fd)
     assert(fd >= 0);
 #if defined(HAS_FCNTL) && defined(F_SETFD) && defined(FD_CLOEXEC)
     (void) fcntl(fd, F_SETFD, FD_CLOEXEC);
+#elif !defined(DEBUGGING)
+    PERL_UNUSED_ARG(fd);
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


khwilliamson (commented):
lgtm

doio.c
```diff
@@ -2967,6 +2971,9 @@ Perl_cando(pTHX_ Mode_t mode, bool effective, const Stat_t *statbufp)
      /* Atari stat() does pretty much the same thing. we set x_bit_set_in_stat
       * too so it will actually look into the files for magic numbers
       */
+# ifdef __MINGW32__
+    PERL_UNUSED_ARG(effective);
+# endif
```
mauke: Does the `#ifdef __MINGW32__` guard make sense here? If `effective` is unused in this branch (`DOSISH`), we could just mark it as `PERL_UNUSED_ARG` unconditionally, even if other compilers don't warn about it (yet?).
Or am I missing something?

sisyphus: > Does the <code>#ifdef __MINGW32__ </code> guard make sense here?

I don't think it serves a purpose, and I've now removed it.
I was bit thrown by the fact that I don't know what the "DOSISH" systems are ... and lost sight of the fact that it doesn't actually matter.

mauke (approved):

doio.c
Leont: > I was bit thrown by the fact that I don't know what the "DOSISH" systems are

At this point it's really only WinNT, but historically it also included Win9x, DOS and OS/2

PR #21920, opened 2024-02-03
perlapi: Add detail to sv_setpv_bufsize()

khwilliamson:
suggested edits welcome

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -4972,9 +4972,14 @@ Perl_sv_setsv_cow(pTHX_ SV *dsv, SV *ssv)
 /*
 =for apidoc sv_setpv_bufsize
 
-Sets the SV to be a string of cur bytes length, with at least
-len bytes available. Ensures that there is a null byte at SvEND.
+Sets the SV to be a string of C<cur> bytes length, with at least
+C<len> bytes available.   Ensures that there is a null byte at C<SvEND>.
+
 Returns a char * pointer to the SvPV buffer.
+
+The caller must set the first C<cur> bytes of C<sv> before the first use of its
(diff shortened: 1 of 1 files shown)


sv.c
```diff
@@ -4972,8 +4972,9 @@ Perl_sv_setsv_cow(pTHX_ SV *dsv, SV *ssv)
 /*
 =for apidoc sv_setpv_bufsize
 
-Sets the SV to be a string of cur bytes length, with at least
-len bytes available. Ensures that there is a null byte at SvEND.
+Sets the SV to be a string of C<cur> bytes length, with at least
+C<len> bytes available.   If <cur> is not zero, the string contents are
+unspecified.  Ensures that there is a null byte at C<SvEND>.
```
tonycoz: the second `<cur>` is missing the `C`

Maybe:
```
If C<cur> is not zero the string contents are not initialized.
```
Or (this is my preferred choice):
```
The string contents will not be initialized.
```
Since `cur == 0` is just a degenerate case and "will not be" makes it less passive.

If we want to make fewer promises, so we could change it to initialize in some cases (`-DDEBUGGING` builds might initialize it to a filler byte.)
```
The string contents currently will not be initialized.  This may change.
```

khwilliamson:
I want to make it plain that you can use this SV as-is, safely; it is an empty string with available space that we don't bother changing from whatever was there before the call.  I'm fine with saying this might change.

sv.c
khwilliamson: I changed it to be more explicit of what the caller needs to do.  I think that obviates the need to say we could change it, because any change we make would not be harder for the caller than what is now called for.

tonycoz (approved):

PR #21927, opened 2024-02-05
Recognise a :reader attribute on class fields

leonerd:
Generates simple reader accessors.

Not yet addressed - any kind of writer or mutator.

perldelta to be written in a separate commit

### DIFF
--- a/class.c
+++ b/class.c
@@ -949,6 +949,88 @@ apply_field_attribute_param(pTHX_ PADNAME *pn, SV *value)
     (void)hv_store_ent(aux->xhv_class_param_map, value, newSVuv(PadnameFIELDINFO(pn)->fieldix), 0);
 }
 
+static void
+apply_field_attribute_reader(pTHX_ PADNAME *pn, SV *value)
+{
+    if(value)
+        SvREFCNT_inc(value);
+    else
+        /* Default to name minus the sigil */
+        value = newSVpvn_utf8(PadnamePV(pn) + 1, PadnameLEN(pn) - 1, PadnameUTF8(pn));
+
+    PADOFFSET fieldix = PadnameFIELDINFO(pn)->fieldix;
+
+    I32 floor_ix = start_subparse(FALSE, 0);
+    SAVEFREESV(PL_compcv);
+
+    I32 save_ix = block_start(TRUE);
+
+    PADOFFSET padix;
+
+    padix = pad_add_name_pvs("$self", 0, NULL, NULL);
+    assert(padix == PADIX_SELF);
+
+    padix = pad_add_name_pvn(PadnamePV(pn), PadnameLEN(pn), 0, NULL, NULL);
+    intro_my();
+
+    OP *methstartop;
+    {
+        UNOP_AUX_item *aux;
+        Newx(aux, 2 + 2, UNOP_AUX_item);
+
+        UNOP_AUX_item *ap = aux;
+        (ap++)->uv = 1;       /* fieldcount */
+        (ap++)->uv = fieldix; /* max_fieldix */
+
+        (ap++)->uv = padix;
+        (ap++)->uv = fieldix;
+
+        methstartop = newUNOP_AUX(OP_METHSTART, 0, NULL, aux);
+    }
+
+    OP *argcheckop;
+    {
+        UNOP_AUX_item *aux;
+        Newx(aux, 3, UNOP_AUX_item);
+
+        aux[0].iv = 0; /* params */
+        aux[1].iv = 0; /* opt_params */
+        aux[2].iv = 0; /* slurpy */
+
+        argcheckop = newUNOP_AUX(OP_ARGCHECK, 0, NULL, aux);
+    }
+
+    OP *retop;
+    {
+        OPCODE optype = 0;
+        switch(PadnamePV(pn)[0]) {
+            case '$': optype = OP_PADSV; break;
+            case '@': optype = OP_PADAV; break;
+            case '%': optype = OP_PADHV; break;
+            default: NOT_REACHED;
+        }
+
+        retop = newLISTOP(OP_RETURN, 0,
+            newOP(OP_PUSHMARK, 0),
+            newPADxVOP(optype, 0, padix));
+    }
+
+    OP *ops = newLISTOPn(OP_LINESEQ, 0,
+            newSTATEOP(0, NULL, NULL),
+            methstartop,
+            argcheckop,
+            retop,
+            NULL);
+
+    SvREFCNT_inc(PL_compcv);
+    ops = block_end(save_ix, ops);
+
+    OP *nameop = newSVOP(OP_CONST, 0, value);
+
+    CV *cv = newATTRSUB(floor_ix, nameop, NULL, NULL, ops);
+    CvIsMETHOD_on(cv);
+}
+
 static struct {
     const char *name;
     bool requires_value;
@@ -957,6 +1039,10 @@ static struct {
     { .name           = "param",
       .requires_value = false,
       .apply          = &apply_field_attribute_param,
+    },
+    { .name           = "reader",
+      .requires_value = false,
+      .apply          = &apply_field_attribute_reader,
     },
     { NULL, false, NULL }
 };
--- a/class.c
+++ b/class.c
@@ -489,6 +489,8 @@ static void S_split_attr_nameval(pTHX_ SV *sv, SV **namp, SV **valp)
 
         if(value_max >= value_at)
             *valp = sv_2mortal(newSVpvn_utf8(value_at, value_max - value_at + 1, do_utf8));
+        else
+            *valp = NULL;
     }
     else {
         *namp = sv;
--- a/pod/perlclass.pod
+++ b/pod/perlclass.pod
@@ -227,6 +227,36 @@ If there is no defaulting expression then the parameter is required by the
 constructor; the caller must pass it or an exception is thrown. With a
 defaulting expression this becomes optional.
 
+=head3 :reader
+
+A field with a C<:reader> attribute will generate a reader accessor method
+automatically.  The generated method will have an empty (i.e. zero-argument)
+signature, and its body will simply return the value of the field variable.
+
+    field $s :reader;
+
+    # Equivalent to
+    field $s;
+    method s () { return $s; }
+
+By default accessor method will have the same name as the field (minus the
+leading sigil), but a different name can be specified in the attribute's value.
+
+    field $x :reader(get_x);
+
+    # Generates a method
+    method get_x () { return $x; }
+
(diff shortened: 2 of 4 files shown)


Leont (approved):
Look good to me (but I'm no expert on OPcodes)

leonerd:
> Look good to me (but I'm no expert on OPcodes)

The opcode stuff was mostly copy-pasted from Object::Pad's implementation anyway ;)

leonerd:
If @tonycoz has a moment I'd appreciate a look at the various UTF-8 handling around the padnames in particular.

leonerd:
On a casual commandline test using Espernato (and therefore non-Latin1 Unicode), it appears to work for me:

```
$ ./perl -Ilib -CD -E 'use utf8; use feature "class"; class X { field $sandviĉon :param :reader } say X->new( sandviĉon => "fromaĝo" )->sandviĉon'
class is experimental at -e line 1.
field is experimental at -e line 1.
Wide character in say at -e line 1.
fromaĝo
```

pod/perlclass.pod
```diff
@@ -227,6 +227,30 @@ If there is no defaulting expression then the parameter is required by the
 constructor; the caller must pass it or an exception is thrown. With a
 defaulting expression this becomes optional.
 
+=head3 :reader
+
+A field with a C<:reader> attribute will generate a reader accessor method
+automatically. By default accessor method will have the same name as the field
+(minus the leading sigil), but a different name can be specified in the
+attribute. The generated method will have an empty (i.e. zero-argument)
```
rwp0: > but a different name can be specified in the attribute

How? If by `:reader(NAME)` let's document that too please (or give example thereof).

class.c
```diff
@@ -949,6 +949,88 @@ apply_field_attribute_param(pTHX_ PADNAME *pn, SV *value)
...
+
+    I32 save_ix = block_start(TRUE);
+
+    PADOFFSET padix;
+
+    padix = pad_add_name_pvs("$self", 0, NULL, NULL);
+    assert(padix == PADIX_SELF);
+
+    /* TODO: UTF-8 flag? */
+    padix = pad_add_name_pvn(PadnamePV(pn), PadnameLEN(pn), 0, NULL, NULL);
```
tonycoz: No UTF-8 is needed, the name is always UTF-8, if you look at the definition of pad_add_name_sv():
```
    namepv = SvPVutf8(name, namelen);
    return pad_add_name_pvn(namepv, namelen, flags, typestash, ourstash);
```
though this should be documented.

tonycoz:
> If @tonycoz has a moment I'd appreciate a look at the various UTF-8 handling around the padnames in particular.

It looks good to me, but I think it should have some UTF-8 name tests.

t/class/accessor.t
```diff
@@ -0,0 +1,46 @@
...
+        'Failure from argument to accessor');
+}
+
+# Alternative names
+{
+    class Test2 {
+        field $f :reader(get_f) = "value";
+    }
+
+    is(Test2->new->get_f, "value", 'accessor with altered name');
```
perigrin: Should `Test2->new->f` also work here? The POD suggests not but doesn't state it explicitly and there's no test confirming it doesn't.

t/class/accessor.t
```diff
@@ -0,0 +1,46 @@
...
+    # Read accessor does not permit arguments
+    ok(!eval { $o->s("value") },
+        'Reader accessor fails with argument');
+    like($@, qr/^Too many arguments for subroutine \'Test1::s\' \(got 1; expected 0\) at /,
+        'Failure from argument to accessor');
+}
+
+# Alternative names
+{
+    class Test2 {
```
Grinnz: not that I expect it to be an issue here, but perhaps Test2 isn't the most unique class name? :)

pod/perlclass.pod
leonerd: Example added to docs.

class.c
leonerd: Righty - I'll remove the `TODO` comment.

t/class/accessor.t
leonerd: Test added

t/class/accessor.t
leonerd: Fair. I'll rename these to `Testcase1` etc, and also fix up the names in the other `.t` files in a separate commit.

leonerd:
> It looks good to me, but I think it should have some UTF-8 name tests.

@tonycoz Yes certainly, though I have a whole bunch I plan to steal from https://metacpan.org/release/PEVANS/Object-Pad-0.808/source/t/95utf8.t which test all of the class stuff. I'll do that in a separate test-only PR. (At least, I hope it'll be test-only. Maybe it will uncover a bug :) )

t/class/accessor.t
leonerd: I did this directly in 4650b6afe4

class.c
```diff
@@ -949,6 +949,87 @@ apply_field_attribute_param(pTHX_ PADNAME *pn, SV *value)
...
+    }
+
+    OP *retop;
+    {
+        OPCODE optype = 0;
+        switch(PadnamePV(pn)[0]) {
+            case '$': optype = OP_PADSV; break;
+            case '@': optype = OP_PADAV; break;
+            case '%': optype = OP_PADHV; break;
+        }
```
wolfsage, line 1013: should we have a default case that throws an error?

wolfsage:
This causes a segfault, is this known:
```
use v5.36;
use feature 'class';
no warnings 'experimental::class';

class Testcase1 {
  field $s :reader();
}

```

```
alh@emondsfield:~/p5/perl$ ./perl -Ilib nc.pl
Use of uninitialized value at nc.pl line 6.
Use of uninitialized value at nc.pl line 6.
Segmentation fault (core dumped)
```

Looks like we get a null or freed sv in `apply_field_attribute_reader(pTHX_ PADNAME *pn, SV *value)` for value maybe?

(Likewise, this crashes in a different way - attempting to eat all memory or ... some other bug?
```
use v5.36;
use feature 'class';
use warnings;
no warnings 'experimental::class';

class Testcase1 {
  field $s :param();
}

```
```
alh@emondsfield:~/p5/perl$ ./perl -Ilib nc.pl
Use of uninitialized value at nc.pl line 6.
Out of memory in perl:util:safesysrealloc
```

It can also crash like this:
```
alh@emondsfield:~/p5/perl$ ./perl -Ilib nc.pl
Use of uninitialized value at nc.pl line 7.
perl: sv_inline.h:928: Perl_SvPV_helper: Assertion `PL_valid_types_PVX[SvTYPE(_svcur) & SVt_MASK]' failed.
Aborted (core dumped)

```

Also, what happens in this case?
```
use v5.36;
use feature 'class';
no warnings 'experimental::class';

class Testcase1 {
  field $s :reader("foo");
}
```

I guess we create a method call `"foo"`. It compiles fine, gives no warnings, but `$obj->foo` isn't usable which might surprise some folks.

Also if you do `:reader(DESTROY)` it looks like we will call the reader in the destructor - no harm done but definitely weird. So these magic attributes that can generate CVs can do some weird things....

class.c
leonerd: I think the only case that would happen is if the sigil in the field name wasn't one of these three options, which would likely imply much more serious problems have already happened. But I have added a `default: NOT_REACHED;` to cover that case, at least under `-DDEBUGGING`

leonerd:
> This causes a segfault, is this known:
...
> Looks like we get a null or freed sv in `apply_field_attribute_reader(pTHX_ PADNAME *pn, SV *value)` for value maybe?

Hrm; I don't seem to be able to reproduce that:

```
$ ./perl -Ilib -Mexperimental=class -E 'class Point { field $x :reader() = 123; } say Point->new->x'
123
```

> Also, what happens in this case?
...
>   field $s :reader("foo");
...
> I guess we create a method call `"foo"`. It compiles fine, gives no warnings, but `$obj->foo` isn't usable which might surprise some folks.

Surprising perhaps, yes. Similar would already happen with `:param("foo")`. I suppose we could add a warning similar in style to e.g. this one:

```
$ perl -Mwarnings -E 'qw( foo, bar )'
Possible attempt to separate words with commas at -e line 1.
```

> Also if you do `:reader(DESTROY)` it looks like we will call the reader in the destructor - no harm done but definitely weird. So these magic attributes that can generate CVs can do some weird things....

That feels like a comment to make in documentation - "you *can* do this but you very likely should not."

leonerd:
> Hrm; I don't seem to be able to reproduce that:

Ah but wait a moment, maybe something more subtle going on:

```
$ valgrind ./perl -Ilib -Mexperimental=class -E 'class Point { field $x :param() :reader() = 123; } say Point->new( x => 456 )->x'
==164071== Memcheck, a memory error detector
==164071== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==164071== Using Valgrind-3.20.0 and LibVEX; rerun with -h for copyright info
==164071== Command: ./perl -Ilib -Mexperimental=class -E class\ Point\ {\ field\ $x\ :param()\ :reader()\ =\ 123;\ }\ say\ Point-\>new(\ x\ =\>\ 456\ )-\>x
==164071== 
==164071== Conditional jump or move depends on uninitialised value(s)
==164071==    at 0x1D09F9: apply_field_attribute_param (class.c:925)
==164071==    by 0x1D1327: S_class_apply_field_attribute (class.c:1066)
==164071==    by 0x1D149C: Perl_class_apply_field_attributes (class.c:1091)
==164071==    by 0x298355: Perl_yyparse (perly.y:1574)
==164071==    by 0x1A0360: S_parse_body (perl.c:2609)
==164071==    by 0x19DED0: perl_parse (perl.c:1912)
==164071==    by 0x14F285: main (perlmain.c:106)
==164071== 
==164071== Conditional jump or move depends on uninitialised value(s)
==164071==    at 0x1D0C49: apply_field_attribute_reader (class.c:955)
==164071==    by 0x1D1327: S_class_apply_field_attribute (class.c:1066)
==164071==    by 0x1D149C: Perl_class_apply_field_attributes (class.c:1091)
==164071==    by 0x298355: Perl_yyparse (perly.y:1574)
==164071==    by 0x1A0360: S_parse_body (perl.c:2609)
==164071==    by 0x19DED0: perl_parse (perl.c:1912)
==164071==    by 0x14F285: main (perlmain.c:106)
==164071== 
...
```

leonerd:
Turns out to have been a good ol' uninitialised pointer. Fixed by 4b75b723bf

leonerd:
Accepting that there should be a (warning? error?) when attempting to create a `:reader` with a name that's not a valid perl symbol, and also that there should be another test file to check some UTF-8 names, I'll do both of those in a subsequent PR.
That aside, I think this looks good to go?

tonycoz (approved):

rwp0 (approved):

Issue #17949, opened 2020-07-07
[EXPERIMENT] defer {} syntax


leonerd:
(copied from https://www.nntp.perl.org/group/perl.perl5.porters/2020/06/msg257611.html )

I'm trying to simplify the design of a `try/catch` syntax out of [Syntax::Keyword::Try](https://metacpan.org/pod/Syntax::Keyword::Try) so it can be moved into core perl. One thing I think we can get rid of is the `finally` keyword, in favour of instead inventing a new `LEAVE` phaser block, similar to the Raku ones.

In summary: I'd like to add `LEAVE` with syntax which looks like another phaser such as `END`, but runs at the time you leave the block it is placed in:

```perl
  sub f {
    print 1;
    LEAVE { print 3; }

    print 2;
  }

  f();
```

Will output

    123

Furthermore, a LEAVE phaser inside something like a `foreach` loop would run every time you leave the block in order to run a new one:

```perl
  foreach (qw( a b c )) {
    print "Start($_)";
    LEAVE { print "End($_)"; }

    more code here...
  }
```

will output

    Start(a)End(a)Start(b)End(b)Start(c)End(c)

Seems simple enough.

It is important to note that a LEAVE block is similar to an END block, in that the mere presence of that syntax is enough to queue it for being run, regardless of whether the line of code it's on was actually "reached". Thus, just as in

```perl
  exit 0;
  END { say "Ending the program now" }
```

still prints, so too would

```perl
  sub f {
    return;
    LEAVE { say "Leaving the function now" }
  }
```

If we had a LEAVE phaser (totally independent of any thoughts of try/catch), it reduces the scope of that work considerably and makes it easier to work out how to implement.

Thoughts?

Grinnz:
FTR, prior work on CPAN includes https://metacpan.org/pod/Scope::OnExit, https://metacpan.org/pod/Scope::Cleanup, and various scope guards like https://metacpan.org/pod/Scope::Guard

leonerd:
Further discussions about LEAVE have lead to an interesting question - What is the behaviour of exceptions thrown from them?

In a simple case such as

```perl
  sub f {
    return "123";
    LEAVE { die "456" }
  }
```

It is relatively easy to argue the case that this should die with "456 at ...", but complications come when the LEAVE block is being executed during stack unwind because another exception is being thrown.

```perl
  sub g {
    die "123";
    LEAVE { die "456" }
  }
```

[...]

PR #21929, opened 2024-02-05
Document appearance of "defer" feature in perlsyn.pod

dboehmer:
Perl version that introduced `defer` taken from https://github.com/Perl/perl5/issues/17949#issuecomment-1168728607

### DIFF
--- a/pod/perlsyn.pod
+++ b/pod/perlsyn.pod
@@ -787,8 +787,8 @@ is executed on the way past. It provides a means of deferring execution until
 a later time. This acts similarly to syntax provided by some other languages,
 often using keywords named C<try / finally>.
 
-This syntax is available if enabled by the C<defer> named feature, and is
-currently experimental. If experimental warnings are enabled it will emit a
+This syntax is available since Perl 5.36 if enabled by the C<defer> named feature,
+and is currently experimental. If experimental warnings are enabled it will emit a
 warning when used.
 
     use feature 'defer';


leonerd:
Locally, this change appears to be fine. But on a wider perspective, we don't currently document when various syntax features first appeared. If we add it here is raises the question of whether many more of the recent additions should be similarly documented.

I think they probably should, but before we merge this one in particular we should decide on a wider view, if that's the sort of thing we want to do.

dboehmer:
I actually tried to find out when `defer` became available and didn’t find a list. `perldoc feature` does also only mention that it is experimental.

Both `perldoc feature` and `perldoc -f defer` refer to `perldoc perlsyn`. So I looked there and several syntax features are already described with the Perl version required:
* `when` https://github.com/Perl/perl5/blob/07a53dca6ee0b9c82acfb8d777fe4e7cc6aae661/pod/perlsyn.pod#219
* `for` lists with multi-value assignments https://github.com/Perl/perl5/blob/07a53dca6ee0b9c82acfb8d777fe4e7cc6aae661/pod/perlsyn.pod#285
* yadda-yadda operator `...` https://github.com/Perl/perl5/blob/07a53dca6ee0b9c82acfb8d777fe4e7cc6aae661/pod/perlsyn.pod#995
*

khwilliamson:
Other commits that add similar information have been approved in recent releases, in spite of such concerns.

I think the de-facto rule is "If this bothered someone enough that they created a PR, then accept it". 

I think that is reasonable, so this LGTM

leonerd:
Fair enough. In that case we'll merge this one now and we can make a more thorough sweep of perlsyn, etc.. to add similar notes about when other recent additions appeared separately.

leonerd (approved):

PR #21934, opened 2024-02-06
perlapi, av.h: Need "apidoc_item"

khwilliamson:
Commit 0d564a3d013287cf0880fdc3264f30c20fe6188d added docs for newAV_mortal, but it needed to use "apidoc_item", not plain "apidoc".

Not doing so resulted in this build time warning:

Empty pod for newAV (from av.h) at autodoc.pl line 1265.

and the resulting pod/perlapi.pod was incorrect.

Plain "apidoc" is used to begin a new entry.  "apidoc_item" is for related subsidiary functions to the main entry.

Suggestions welcome as to how to make this relationship more obvious. Perhaps it should be renamed "apidoc_sub_item"

### DIFF
--- a/av.h
+++ b/av.h
@@ -113,7 +113,7 @@ Note that there are both real and fake AVs; see the beginning of this file and
 'av.c'
 
 =for apidoc newAV
-=for apidoc newAV_mortal
+=for apidoc_item newAV_mortal
 =for apidoc_item newAV_alloc_x
 =for apidoc_item newAV_alloc_xz
 


leonerd (approved):
LGTM

khwilliamson:
@richardleach, please note

richardleach:
Thanks @khwilliamson. I don't think a rename is necessarily needed. I'd noted the correct label but then bodged it in haste.

Having an easy way to spot new warnings would be most useful to me personally. Are there any obvious recommendations, or is this a Porting/ tooling gap?

khwilliamson:
It is a porting tool gap; which I had never considered before.  Patches welcome.

When changing pod, it's always a good idea to eyeball the result by running `perlodc`, though it is easy to forget, so I think a porting test is a good idea.

khwilliamson:
Such a tool would be easy to create; just run autodoc.pl and fail if stderr is non-empty.

But it's still a good idea to eyeball `perldoc` output, as unintended results are easily generated without an error or warning being raised.

Issue #21930, opened 2024-02-06
Misleading warning "Use of uninitialized value $x in numeric eq (==)" when comparing value of $x from core function


dboehmer:
Module: —

**Description**
I am comparing the length of a string. It happened to be undefined but the warning was misleading and made me think the logic of my comparison was broken. The warning reads like my code was missing parenthesis and compiled into `length( $x == 0 )` which it didn’t.

I think this was probably discussed before and I am not the first to bring this up but googling this yields a ton of posts about low quality code which produced the warning and users trying to figure out its basic meaning.

**Steps to Reproduce**

```console
$ perl -wE 'my $x; say( (length $x) == 0 )'
Use of uninitialized value $x in numeric eq (==) at -e line 1.
1
```

The warning is incorrect: not `$x` is compared but `length($x)`. I confirmed this behavior for

* `abs($x)`
* `length($x)`
* `delete()` on hash keys:
```console
$ perl -wE 'my %h; say( delete $h{foo} == 0 )'
Use of uninitialized value $h{"foo"} in numeric eq (==) at -e line 1.
1
```

**Expected behavior**

It’s probably hard to implement a generic representation of the expression. So any of those:
1. most informative: `Use of uninitialized value length($x) in numeric eq (==)` (I think, it might be possible as it looks like there is already special handling for core functions.)
1. more precise: `Use of uninitialized value $x in length`
1. just `Use of uninitialized value in numeric eq (==)`  like for user-defined functions:

```console
$ perl -wE 'sub f{shift} my $x; say( f($x) == 0 )'
Use of uninitialized value in numeric eq (==) at -e line 1.
1
```

**Perl configuration**

```
Summary of my perl5 (revision 5 version 38 subversion 0) configuration:
   
  Platform:
    osname=linux
    osvers=5.10.0-25-amd64
    archname=x86_64-linux-thread-multi
    uname='linux berlin 5.10.0-25-amd64 #1 smp debian 5.10.191-1 (2023-08-16) x86_64 gnulinux '
    config_args='-de -Dprefix=/home/daniel/perl5/perlbrew/perls/perl-5.38.0-threads-bullseye -Dusethreads -Aeval:scriptdir=/home/daniel/perl5/perlbrew/perls/perl-5.38.0-threads-bullseye/bin'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
[...]

PR #21936, opened 2024-02-07
Improve "uninitialized value" warnings for length/pop/shift/splice

mauke:
See #21930.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -17487,6 +17487,15 @@ S_find_uninit_var(pTHX_ const OP *const obase, const SV *const uninit_sv,
         goto do_op;
     }
 
+    case OP_LENGTH:
+        o = cUNOPx(obase)->op_first;
+        sv = find_uninit_var(o, uninit_sv, match, desc_p);
+        if (sv) {
+            Perl_sv_insert_flags(aTHX_ sv, 0, 0, STR_WITH_LEN("length("), 0);
+            sv_catpvs_nomg(sv, ")");
+        }
+        return sv;
+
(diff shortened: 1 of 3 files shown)


haarg:
`readline` also suffers from this problem.

sv.c
```diff
@@ -17487,6 +17492,47 @@ S_find_uninit_var(pTHX_ const OP *const obase, const SV *const uninit_sv,
         goto do_op;
     }
 
+    case OP_LENGTH:
+        o = cUNOPx(obase)->op_first;
+        sv = find_uninit_var(o, uninit_sv, match, desc_p);
+        if (sv) {
+            Perl_sv_insert_flags(aTHX_ sv, 0, 0, STR_WITH_LEN("length("), 0);
```
leonerd, line 17499: Why the `Perl_...(aTHX_ ...)` form? `sv_insert_flags` is available directly as a macro-wrapped function.

mauke: Because the macro form takes 6 arguments, but I'm only passing 5. `STR_WITH_LEN("...")` expands to two comma-separated items, but that happens too late.

leonerd: Ugh, ohgod yes that particular issue. I wonder if we should make some conveneince wrappers for this. Perhaps:

```
sv_prepend_pvn(sv, pv, len, flags)
sv_prepend_pvs(sv, "str", flags)
```

A separate issue though; this use here is fine given that.

leonerd (approved):
LGTM

PR #21938, opened 2024-02-07
 locale.c: Avoid copies

khwilliamson:
Refactoring slightly as suggested by Tony Cook avoids having to make two
copies in some circumstances.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -2236,8 +2236,7 @@ S_querylocale_2008_i(pTHX_ const locale_category_index index,
         else {
 
             QUERYLOCALE_LOCK;
-            retval = savepv(my_querylocale(index, cur_obj));
-            QUERYLOCALE_UNLOCK;
+            retval = my_querylocale(index, cur_obj);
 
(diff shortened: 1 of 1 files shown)


locale.c
```diff
@@ -2251,10 +2251,11 @@ S_querylocale_2008_i(pTHX_ const locale_category_index index,
              * system distinugishes "C" from "POSIX", we do too.) */
             if (cur_obj == PL_C_locale_obj && ! isNAME_C_OR_POSIX(retval)) {
                 Safefree(retval);
-                retval = savepv("C");
+                retval = "C";
+            }
+            else {
+                SAVEFREEPV(retval);
             }
```
tonycoz: The test `cur_obj == PL_C_locale_obj && ! isNAME_C_OR_POSIX(retval)` seems very cheap to me, certainly a lot cheaper than savepv(), wouldn't it be better to do this test within the lock and avoid the copy entirely?

tonycoz (approved):

PR #21940, opened 2024-02-07
sv_inline.h: Use Sv*OK (rather than Sv*OK_nog) in Perl_SvUV_nomg, Perl_SvNV_nomg

t-a-k:
It seems that some errors were made when original macros (`SvUV_nomg` and `SvNV_nomg` in `sv.h`) were translated to inline functions.

### DIFF
--- a/sv_inline.h
+++ b/sv_inline.h
@@ -852,7 +852,7 @@ PERL_STATIC_INLINE UV
 Perl_SvUV_nomg(pTHX_ SV *sv) {
     PERL_ARGS_ASSERT_SVUV_NOMG;
 
-    if (SvIOK_nog(sv))
+    if (SvUOK(sv))
         return SvUVX(sv);
     return sv_2uv_flags(sv, 0);
 }
@@ -861,7 +861,7 @@ PERL_STATIC_INLINE NV
 Perl_SvNV_nomg(pTHX_ SV *sv) {
     PERL_ARGS_ASSERT_SVNV_NOMG;
 
-    if (SvNOK_nog(sv))
+    if (SvNOK(sv))
         return SvNVX(sv);
     return sv_2nv_flags(sv, 0);
 }


richardleach:
The move from macros to inline functions was in https://github.com/Perl/perl5/commit/dae8ab819b9ffd6d540355efffacf225aff00a0f. The intention was that `sv` would be evaluated only once.

richardleach:
@t-a-k - do you have test cases that demonstrate erroneous behaviour as a result of this change please?

richardleach:
@khwilliamson Do you happen to remember the ins & outs of the original commit?

sv_inline.h
```diff
@@ -861,7 +861,7 @@ PERL_STATIC_INLINE NV
 Perl_SvNV_nomg(pTHX_ SV *sv) {
     PERL_ARGS_ASSERT_SVNV_NOMG;
 
-    if (SvNOK_nog(sv))
+    if (SvNOK(sv))
```
khwilliamson, line 864: Prior to the commit that made this inline, it was:
`#define SvNV_nomg(sv) (SvNOK(sv) ? SvNVX(sv) : sv_2nv_flags(sv, 0))`
`so that commit was erroneous, and this new one fixes it.

khwilliamson (approved):

sv_inline.h
```diff
@@ -852,7 +852,7 @@ PERL_STATIC_INLINE UV
 Perl_SvUV_nomg(pTHX_ SV *sv) {
     PERL_ARGS_ASSERT_SVUV_NOMG;
 
-    if (SvIOK_nog(sv))
+    if (SvUOK(sv))
```
khwilliamson, line 855: Prior to the commit that made this inline, it was:
`#define SvUV_nomg(sv) (SvUOK(sv) ? SvUVX(sv) : sv_2uv_flags(sv, 0))`
so that commit was erroneous, and this new one fixes it.

khwilliamson (approved):
It looks like I blew it in that commit.   It would be nice to have test cases, but we should merge this in any event.

khwilliamson:
@t-a-k thanks for digging this up

t-a-k:
> @t-a-k - do you have test cases that demonstrate erroneous behaviour as a result of this change please?

I'm afraid that making test cases will be difficult for these cases, because these changes will not change the result of `Sv*V_nomg`.  Even prior to this patch, there may be redundant call to `sv_2*v_flags`, but these functions will eventually do the right thing.  I think there had been some performance loss, but this loss would be too subtle to be observed from Perl test cases.

PR #21946, opened 2024-02-08
locale.c: Each toggle locale must have matching restore

khwilliamson:
Future commits will want this symmetry, which until now hasn't mattered.


khwilliamson:
I forgot that this is part of #21908

PR #21949, opened 2024-02-08
locale.c: use type-agnostic code for storing tm_zone

### LINKED ISSUE #21948: Blead breaks build in g++, clang++ builds on some platforms
Compiling `perl` using `g++` or `clang++` has started to fail on certain platforms, such as several of the BSDs.

For example, building on FreeBSD-13 with the following configuration:
```
$ g++ --version | head -n1
g++ (FreeBSD Ports Collection) 12.2.0

$ sh ./Configure -des -Dusedevel -Dcc=g++ && make test_prep
```
... the build stops here:
```
g++ -c -DPERL_CORE -DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -DNO_POSIX_2008_LOCALE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_FORTIFY_SOURCE=2 -O2 -Wall -Werror=pointer-arith -Werror=vla -Wextra -Wno-long-long -Wwrite-strings -Wno-use-after-free locale.c
locale.c: In function 'void S_ints_to_tm(tm*, int, int, int, int, int, int, int, int, int)':
locale.c:8072:25: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
 8072 |         mytm->tm_zone = zone;
      |                         ^~~~
      |                         |
      |                         const char*
*** Error code 1

Stop.
```
Line 8072 in `locale.c` was last modified in blead yesterday.
```
commit 36e41d53ff6a5995c16234f8027ef34b5e6a34cc
AuthorDate: Wed Jan 10 11:30:11 2024 -0700
CommitDate: Wed Feb 7 10:59:05 2024 -0700

    Don't use both mini_mktime() and mkttime()
```
`g++` builds on Linux so far appear to be unaffected.

mauke:
On some platforms, struct tm has a tm_zone member, but its type is not consistent:

 - Linux: const char *tm_zone;
 - FreeBSD (& probably other BSDs): char *tm_zone;

In order to save/restore tm_zone, we can't just use a "const char *" or "char *" variable because different parts of this code would always be a const violation on one platform or the other.

Workaround: Use the tm_zone member of a full struct tm, which has the right type no matter the platform.

Fixes #21948.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -8044,11 +8044,8 @@ S_ints_to_tm(pTHX_ struct tm * mytm,
 
         /* But since mini_mktime() doesn't handle these fields, save the
          * results we already have for them */
-#  ifdef HAS_TM_TM_GMTOFF
-        long int gmtoff = mytm->tm_gmtoff;
-#  endif
-#  ifdef HAS_TM_TM_ZONE
-        const char *zone = mytm->tm_zone;
+#  if defined(HAS_TM_TM_GMTOFF) || defined(HAS_TM_TM_ZONE)
+        struct tm mytm2 = *mytm;
 #  endif
 
         /* Repeat the entire process with mini_mktime */
@@ -8066,10 +8063,10 @@ S_ints_to_tm(pTHX_ struct tm * mytm,
 
         /* And use the saved libc values for tm_gmtoff and tm_zone */
 #  ifdef HAS_TM_TM_GMTOFF
-        mytm->tm_gmtoff = gmtoff;
+        mytm->tm_gmtoff = mytm2.tm_gmtoff;
 #  endif
 #  ifdef HAS_TM_TM_ZONE
-        mytm->tm_zone = zone;
+        mytm->tm_zone = mytm2.tm_zone;
 #  endif
 
     }


jkeenan:
@mauke, could you create a `smoke-me/` branch in the main repository so that our BSD smokers can pick up these fixes and test them?  Thanks.

mauke:
@jkeenan Done (I think).

locale.c
```diff
@@ -8044,11 +8044,8 @@ S_ints_to_tm(pTHX_ struct tm * mytm,
 
         /* But since mini_mktime() doesn't handle these fields, save the
          * results we already have for them */
-#  ifdef HAS_TM_TM_GMTOFF
-        long int gmtoff = mytm->tm_gmtoff;
```
khwilliamson, line 8048: Is this general enough that we don't have to use your `struct tm` method for this field too?  When writing this, I was surprised that this was a `long`, since I can't imagine it ever exceeding 14 * 60 * 60 (for double daylight time 12 timezones away from gmt).

khwilliamson (approved):

locale.c
mauke: 14 * 60 * 60 = 50400, which exceeds the range of a 16-bit signed integer type (which is all C requires from `int`). To get a type that is guaranteed to be at least 32 bits wide, you need `long int` (or `int_least32_t`, but that's only C99, I believe).

I removed the `gmtoff` local variable both for symmetry with `zone` and because I'd already "paid" for an entire `struct tm` on the stack, so why not make use of its fields?

jkeenan:
On 2/8/24 11:18, mauke wrote:
> @jkeenan <https://github.com/jkeenan> Done (I think).
> 

Thanks.  Smoking now on the machine from which I filed ticket.

locale.c
khwilliamson: somehow I misread this until I downloaded it and looked on my box.  You already had done it the way I was suggesting.

mauke:
This is yours, I think? https://perl5.test-smoke.org/report/5049091
Looks good.

leonerd (commented):
I don't have a BSD box to test on, but from eyeballing it alone this looks like a reasonable change to make and if it fixes a bug then go for it.

jkeenan:
> This is yours, I think? https://perl5.test-smoke.org/report/5049091 Looks good.

Yes, it is from my machine.  However, I really wish that you hadn't been so quick to merge this branch into blead.  The overwhelming majority of our *BSD smoke-test reports are coming from Carlos, including the ones where compilation was done with g++ or clang++ and where I spotted these build-time failures in the first place.  My impression is that most of our non-GH smokers run on a 24- or 48-hour cycle, so I generally try to give a `smoke-me/` branch that length of time to accrue results before I make a recommendation to merge or not.

mauke:
I blame this one on:

> &lt;khw> mauke, I have tested https://github.com/Perl/perl5/pull/21949 on FreeBSD.  Since blead is broken, I think it should be applied sooner than later

Also:
- It doesn't make things worse.
- It fixes all the BSD builds I saw on https://perl.develop-help.com, including yours.

But I'll try to keep it mind for next time.

khwilliamson:
I disagree that we should have waited.

My assertion is that getting blead back to being unbroken is higher priority than waiting for smoke reports that the fix works everywhere.  blead breakages are a  potential issue in perpetuity for future bisections, condemning our future selves and our successors to have to do extra quite unexpected, unwelcome research  should a bisect land in just the wrong place.   This effort dwarfs the consequences of the current fix being incomplete.

In this case, looking at the fix, and the analysis of the problem, it was quite unlikely that it would work for just one and not the other boxes.

And then there is the immediate problem.  I have smokes that I started today that were ruined by this problem, and had to be restarted.  (I realize this is ironic, since I was the one who created the problem.)  And I refreshed blead and restarted them; now they are ruined again by #21957 not being yet merged.  To get anything like timely results, I will have to apply that fix to my smokes and restart them.

The bottom line is that minimizing the period that blead is broken is a big deal.

PR #21954, opened 2024-02-08
Fatalize downgrading a `use VERSION` past v5.11

leonerd:
In Perl 5.36 we made this a deprecation warning, due to be removed in 5.40. Now it is being removed.

This removal means that the three shadow hints bits used to implement implicit vs explicit strict hints can be reclaimed at a later date, for use in other features.

### DIFF
--- a/op.c
+++ b/op.c
@@ -8018,6 +8018,15 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
 
         U16 shortver = S_extract_shortver(aTHX_ use_version);
 
+        if (shortver && PL_prevailing_version) {
+            /* use VERSION while another use VERSION is in scope
+             * This should provoke at least a warning, if not an outright error
+             */
+            /* downgrading from >= 5.11 to < 5.11 is now fatal */
+            if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11))
+                croak("Downgrading a use VERSION declaration to below v5.11 is not permitted");
+        }
+
         /* If a version >= 5.11.0 is requested, strictures are on by default! */
         if (shortver >= SHORTVER(5, 11)) {
             if (!(PL_hints & HINT_EXPLICIT_STRICT_REFS))
@@ -8034,10 +8043,6 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
         }
         /* otherwise they are off */
         else {
-            if(PL_prevailing_version >= SHORTVER(5, 11))
-                deprecate_fatal_in(WARN_DEPRECATED__VERSION_DOWNGRADE, "5.40",
-                    "Downgrading a use VERSION declaration to below v5.11");
-
             if (!(PL_hints & HINT_EXPLICIT_STRICT_REFS))
                 PL_hints &= ~HINT_STRICT_REFS;
             if (!(PL_hints & HINT_EXPLICIT_STRICT_SUBS))
--- a/t/lib/feature/implicit
+++ b/t/lib/feature/implicit
@@ -62,37 +62,34 @@ EXPECT
 # lower version after higher version
 sub evalbytes { print "evalbytes sub\n" }
 sub say { print "say sub\n" }
+use 5;
+say "no";
 use 5.015;
 evalbytes "say 'yes'";
 use 5.014;
 evalbytes;
-use 5;
-say "no"
 EXPECT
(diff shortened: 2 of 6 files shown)


op.c
```diff
@@ -8018,6 +8018,15 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
...
 
+        if (shortver && PL_prevailing_version) {
+            /* use VERSION while another use VERSION is in scope
+             * This should provoke at least a warning, if not an outright error
+             */
+            /* downgrading from >= 5.11 to < 5.11 is now fatal */
+            if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11))
+                croak("Downgrading a use VERSION declaration to below v5.11 is not permitted");
+        }
+
```
leonerd, line 8029: I should add, as a pre-review comment, that the reason for this slightly-odd code structure is that I have some more commits that add other warnings or errors in this same section. So it's a weird shape at the moment but it'll make more sense later.

leonerd:
Rebased and edited title for clarity

khwilliamson (approved):
lgtm

t/lib/feature/implicit
```diff
@@ -58,41 +58,3 @@ Helloworld
...
-evalbytes;
-use 5;
-say "no"
-EXPECT
-Downgrading a use VERSION declaration to below v5.11 is deprecated, and will become fatal in Perl 5.40 at - line 8.
-yes
-evalbytes sub
-say sub
-########
-# Implicit unicode_string feature
```
wolfsage, line 77: I think all three of these tests are supposed to be testing a change before/after v5.15 and the addition of the downgrade warning was added to the tests when we added the warning. But removing this warning and making it fatal isn't a reason to get rid of these tests - they are here to test other behaviour. I think the real solution is to reorder the tests like so:
```
# lower version after higher version
sub evalbytes { print "evalbytes sub\n" }
sub say { print "say sub\n" }
use 5;
say "no";
use 5.015;
evalbytes "say 'yes'";
use 5.014;
evalbytes;
EXPECT
say sub
yes
evalbytes sub
########
# Implicit unicode_string feature
use v5.8.8;
my $sharp_s = chr utf8::unicode_to_native(0xdf);
print 'ss' =~ /$sharp_s/i ? "ok\n" : "nok\n";
use v5.14;
print 'ss' =~ /$sharp_s/i ? "ok\n" : "nok\n";
EXPECT
nok
ok
########
# Implicit unicode_eval feature
require '../../t/charset_tools.pl';
my $long_s = byte_utf8a_to_utf8n("\xc5\xbf");
use v5.8.8;
print eval "use utf8; q|$long_s|" eq "\x{17f}" ? "ok\n" : "nok\n";
use v5.15;
print eval "use utf8; q|$long_s|" eq $long_s ? "ok\n" : "nok\n";
EXPECT
ok
ok

```

leonerd: Ahyes; that looks a much better idea. I've done that instead in 517020f627

wolfsage (approved):

op.c
```diff
@@ -8018,6 +8018,15 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
 
         U16 shortver = S_extract_shortver(aTHX_ use_version);
 
+        if (shortver && PL_prevailing_version) {
+            /* use VERSION while another use VERSION is in scope
+             * This should provoke at least a warning, if not an outright error
+             */
+            /* downgrading from >= 5.11 to < 5.11 is now fatal */
+            if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11))
+                croak("Downgrading a use VERSION declaration to below v5.11 is not permitted");
```
wolfsage, line 8027: Why croak() instead of Perl_croak()? (We use Perl_croak() elsewhere in this function).

Does it matter?

leonerd: `croak()` is now just a macro wrapper around `Perl_croak()`.

Until recently, we didn't have a macro wrapper for `croak()` that could be used by Perl core. It only became possible when we adopted C99 and its extra `__VA_ARGS__` macro ability. There's lots of old code that has not yet been updated, which we should go around and fix up sometime.

wolfsage: Ah, croak() is Perl_croak_nocontext but it seems like it doesn't matter too much if you use one or the other? Fine

leonerd: `croak()` used to be `Perl_croak_nocontext` and still is for non-core Perl.

But when we adopted C99, we made these variadic macros safe to use properly from Perl core. For more context (pardon the pun) see 13e5ba49b2cfe0add44db552ecbebb2f785aecbc

PR #21957, opened 2024-02-08
class.c: Correct allocation of OP_ARGCHECK aux structure

leonerd:
The original code was wrong on two counts:
 * Using Newx() instead of PerlMemShared_malloc()
 * Creating a generic UNOP_AUX_item array instead of the special struct type

### DIFF
--- a/class.c
+++ b/class.c
@@ -992,14 +992,14 @@ apply_field_attribute_reader(pTHX_ PADNAME *pn, SV *value)
 
     OP *argcheckop;
     {
-        UNOP_AUX_item *aux;
-        Newx(aux, 3, UNOP_AUX_item);
+        struct op_argcheck_aux *aux = (struct op_argcheck_aux *)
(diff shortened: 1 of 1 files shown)


class.c
```diff
@@ -992,14 +992,14 @@ apply_field_attribute_reader(pTHX_ PADNAME *pn, SV *value)
 
     OP *argcheckop;
     {
-        UNOP_AUX_item *aux;
-        Newx(aux, 3, UNOP_AUX_item);
+        struct op_argcheck_aux *aux = (struct op_argcheck_aux *)
+            PerlMemShared_malloc(sizeof(struct op_argcheck_aux));
```
mauke: I'd prefer `sizeof *aux`.

leonerd: Good call. Force-pushed.

khwilliamson:
The first smokes are showing this fixes it

PR #21961, opened 2024-02-09
perlclass: improve wording

mauke:
At least I think it's an improvement.

### DIFF
--- a/pod/perlclass.pod
+++ b/pod/perlclass.pod
@@ -29,10 +29,12 @@ native keywords for object-oriented programming.
 =head2 History
 
 Since Perl 5, support for objects revolved around the concept of I<blessing>
-references with a package name. Such reference could then be used to call
-subroutines from the package it was blessed with (or any of its parents). This
-system, while bare-bones, was flexible enough to allow creation of multiple
-more advanced, community-driven systems for object orientation.
+references with a package name (see L<perlfunc/"bless REF,CLASSNAME">). Such a
+reference could then be used to call subroutines from the package it was
+blessed with (or any of its parents). This system, while bare-bones, was
+flexible enough to allow creation of multiple more advanced, community-driven
+systems for object orientation. For more information, see L<perlmod> and
+L<perlobj>.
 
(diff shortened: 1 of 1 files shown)


pod/perlclass.pod
```diff
@@ -57,9 +58,9 @@ the scope of current package:
 
     class NAME VERSION : ATTRIBUTES...;
 
-The C<class> keyword declares a new package which is intended to be a class.
-All other keywords from the C<class> feature should be used in scope of this
-declaration.
+The C<class> keyword declares a new package that is intended to be a class.
```
jkeenan: Is there some point in the documentation where we explicitly distinguish between `package` and `class`?  (If so, perhaps we should refer to it here.  If not, then let's add something here.)

pod/perlclass.pod
```diff
@@ -270,22 +271,40 @@ named C<new> and is invoked like a method call on the class name:
...
-populated where possible.
+During object construction, class fields are looked up in the C<%arguments>
+hash and populated where possible.
 
 =head2 Adjustment
 
-Object adjustment can be performed during the construction to run user-defined
-code. It is done with the help of C<ADJUST> blocks, which are called in order
-of declaration.
+Object adjustment is way to run arbitrary user-defined code during object
```
jkeenan: should be `is a way`

khwilliamson (approved):

khwilliamson:
My approval is based on resolving @jkeenan 's comments

pod/perlclass.pod
mauke: Isn't that what the "History" section is about? There's more in perlmod ("Packages", "Perl Classes"), but I'm not sure how to reference that here.

leonerd (approved):
LGTM

pod/perlclass.pod
mauke: I've now added some links to perlfunc/bless, perlmod, and perlobj to the "History" section. In particular, perlmod explains how packages work in general. Do you think this provides enough background information or should we add something to the description of the `class` keyword, too?

PR #21966, opened 2024-02-09
t/run/locale.t: Several improvements

khwilliamson:
Most notable is that on platforms where some category is restricted to always being in the C locale, prior to these commits, only the C locale would be tested for any category.  Now, that decision for one category doesn't affect the decision for other categories.

### DIFF
--- a/t/run/locale.t
+++ b/t/run/locale.t
@@ -511,15 +511,12 @@ EOF
     }
 
 SKIP: {
-        # Note: the setlocale Configure probe could be enhanced to give us the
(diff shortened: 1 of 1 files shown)


jkeenan:
> Most notable is that it may find more locales to test with when not all locale categories can vary from being C

The commit message is somewhat confusing.  What is the *it* which the fifth word refers to?  Does the message as a whole refer to *"those locale categories which can vary from being C"* -- or to some other set of locale categories?

khwilliamson:
comment updated

PR #21968, opened 2024-02-10
Update perldsc.pod adding vital deep slice shortcut

jidanni:
Without showing this beginners always try
`${HoH}{flintstones}{"wife", "pet"} = ( "wilma", "dino" );` racking their brains for why it won't work. Plus such _deeper_ slices are not documented, so clearly, anywhere else!

### DIFF
--- a/pod/perldsc.pod
+++ b/pod/perldsc.pod
@@ -652,6 +652,12 @@ X<hash of hashes> X<HoH>
      $HoH{flintstones}{$what} = $new_folks{$what};
  }
 
+ # same, but using a hash slice
+ @{ $HoH{flintstones} }{ keys %new_folks } = values %new_folks;
+
+ # same, but without a %new_folks variable
+ @{ $HoH{flintstones} }{ "wife", "pet" } = ( "wilma", "dino" );
+
 =head2 Access and Printing of a HASH OF HASHES
 
  # one element


pod/perldsc.pod
```diff
@@ -652,6 +652,10 @@ X<hash of hashes> X<HoH>
      $HoH{flintstones}{$what} = $new_folks{$what};
  }
 
+ # same, but using a slice
+
+ @{ ${HoH}{flintstones} }{ "wife", "pet" } = ( "wilma", "dino" );
```
mauke: I'd prefer calling it a "hash slice" in the comment (as in `same, but using a hash slice`).

For consistency with the other examples, the empty line between the comment and the code should be removed.

Similarly, the `{ }` around `HoH` shouldn't be there (`@{ $HoH{flintstones} }`).

If you want to completely mimic the previous example, including the use of the `%new_folks` variable, it should look like:
```perl
@{ $HoH{flintstones} }{ keys %new_folks } = values %new_folks;
```

jidanni:
1. Please go ahead with your recommendations. I am very thankful. I'll only mess it up if I reedit it.
2. As far as the "completely mimic" version, could you please add that as a second example, as I want users to see both the simple minded version and slick version.

mauke (approved):

khwilliamson (approved):

PR #21972, opened 2024-02-11
Tie::File docpatch warning unicode writes are unsafe.

brainbuz:
I looked at #17494, until that is fixed Tie::File writing to unicode file handles is unsafe, added a warning to the Tie::File documentation.


khwilliamson:
This looks good to me, but @brainbuz , the file `dist/Tie-File/lib/Tie/File.pm` contains a line that sets `VERSION`.  Bump that number and the tests should then pass.

PR #21973, opened 2024-02-11
Free up HINT bit

khwilliamson:
There are two locale-related hint bits, one for each of the two 'use locale' forms.  This commit combines the bits into one.  There is no slowdown in checking if any of the forms is in effect, but deciding which one is slightly slower.  Most perl code doesn't use either form, so the slight slowdown should be acceptable.

But if there are other free bits available, there isn't any need for this slowdown, however rare.  Commit
4210e234cb45c0ad3ecd56b85152559b52131e16 just gave us the ability to free up 3 other HINTS bits.  Those should be consumed first before using this one.

But this code is already written and isn't trivial, so shouldn't be abandoned without a trace.  I did some research before writing it (and before those three bits became available), and this looked to be the lowest hanging fruit to get another bit, should it become necessary.

My intent is to revert this commit immediately, while adding a comment noting its commit number, so that a future maintainer needing a bit can reinstate it.

### DIFF
--- a/lib/locale.pm
+++ b/lib/locale.pm
@@ -3,7 +3,7 @@ package locale;
 use strict;
 use warnings;
 
-our $VERSION = '1.11';
+our $VERSION = '1.12';
 use Config;
 
 $Carp::Internal{ (__PACKAGE__) } = 1;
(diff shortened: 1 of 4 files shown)


leonerd:
> But if there are other free bits available, there isn't any need for this slowdown, however rare. Commit
https://github.com/Perl/perl5/commit/4210e234cb45c0ad3ecd56b85152559b52131e16 just gave us the ability to free up 3 other HINTS bits. Those should be consumed first before using this one.

I'd prefer not to use up those newly-freed ones right away, especially as that commit has had almost no realworld testing yet. If we have to revert it we'll need to keep those bits free for a while yet. I think it best to launch 5.40 with them in the current state, and see into the 5.41 development cycle if we can reuse them.

PR #21975, opened 2024-02-11
Fix warning in run/locale.t

khwilliamson:
A test in this file checks that garbage in the environment is properly handled.  But some shells (notably bash) warn when LC_ALL is set to a non-existent locale.  This commit avoids using a shell to do the setting up of the environment.

This was spotted by Dave Mitchell who also provided the outlines of the fix.


khwilliamson:
@iabyn also tested this fix.

PR #21978, opened 2024-02-12
Update perlop.pod to finally say if one can do ... if $x with here-docs

jidanni:
By the way **I am only guessing** what I wrote is true. It would be wonderful if I were wrong.

Please in that case rewrite my patch, finally documenting how to do it. 

Better yet would be to pull it out of the indebted here-doc example, making a more general example.

In any case, the page needs to finally say, one way or the other, if one can use the ... if $x form with here-docs.

### DIFF
--- a/pod/perlop.pod
+++ b/pod/perlop.pod
@@ -2922,6 +2922,9 @@ This will print...
 
 ...with no leading whitespace.
 
+By the way, one must use if(){}. There is no way to
+do "... if $some_var;" with here-docs.
+
 The line containing the delimiter that marks the end of the here-doc
 determines the indentation template for the whole thing.  Compilation
 croaks if any non-empty line inside the here-doc does not begin with the


Grinnz:
I don't know what you're trying to say here. The line with the `<<` operator can contain any normal Perl syntax, the following lines are strings with potential interpolation. Is this what you mean? This has always been valid syntax.

```
print <<'EOF' if $x;
foo
EOF
```

jidanni:
OK, I will make a new patch. Thanks.

jidanni:
Good. A better example would be just
```
print <<EOF if $x;
bla
EOF
```
But where is a good place to mention it on perlop?

Maybe someone can make a better patch.

Anyway without an example, very few beginners can figure out how the tricky here-doc syntax interacts with "... if $x". Thanks.

mauke:
What's so special about `if`? A here-doc can be used anywhere a normal string literal can appear:
```perl
# normal strings
$x = "a\n" . foo("b\n", "c\n") if $y eq "d\n";
```
Equivalent code using here-docs:
```perl
$x = <<_A_ . foo(<<_B_, <<_C_) if $y eq <<_D_;
a
_A_
b
_B_
c
_C_
d
_D_
```

jidanni:
Well it is great that you all have mastered here-doc syntax.

However, trust me, that people who haven't will try syntax errors like
```
print <<EOF; if $x
bla
EOF
```
because they can't find one single example of how to combine ...if $x with here-docs! So please somebody add one to the docs. Thanks.

mauke:
The documentation does contain

> ```perl
>              myfunc(<< "THIS", 23, <<'THAT');
>           Here's a line
>           or two.
>           THIS
>           and here's another.
>           THAT
> ```

If someone can't generalize from that, should they be using these "advanced" constructs in their code?

We can't list every combination of quoting constructs, operators, and statements. What makes postfix (not prefix) `if` (not `unless`, `while`, `until`, or `for`) so special? The `EXPR if EXPR;` syntax is fully general; no restrictions are placed on the subexpressions. If you can use a here-doc in some expression, you can add an `if` modifier without changing anything.

Or starting with an example from perlop:
```perl
  print <<EOF;
The price is $Price.
EOF
```

We turn to perlsyn, which says:
> Any simple statement may optionally be followed by a SINGLE modifier, just before the terminating semicolon (or block ending).

So let's add an `if` modifier "just before the terminating semicolon":
```perl
  print <<EOF if $cond;
The price is $Price.
EOF
```

That is indeed the correct syntax.

Similarly, if you start with:
```perl
  print <<EOF
The price is $Price.
EOF
  ;
```
Then the modified form is:
```perl
  print <<EOF
The price is $Price.
EOF
  if $cond;
```

jidanni:
You folks could file your taxes just looking at the IRS code, no forms or instructions
needed.

OK, I'll stay away from the advanced stuff and just leave this behind.

https://unix.stackexchange.com/questions/769272/how-to-combine-if-x-and-a-here-doc-in-perl/769273#769273

PR #21984, opened 2024-02-12
pad_add_name_pvn(): the name is in UTF-8

tonycoz:
This came up in discussion for #21927

### DIFF
--- a/pad.c
+++ b/pad.c
@@ -594,7 +594,8 @@ variable.  Stores the name and other metadata in the name part of the
 pad, and makes preparations to manage the variable's lexical scoping.
 Returns the offset of the allocated pad slot.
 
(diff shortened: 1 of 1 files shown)


leonerd (approved):
LGTM

leonerd:
If a padname is *definitely* UTF-8 now, perhaps the docs for `PadnamePV` and `PadnameUTF8` need updating as well?

tonycoz:
It's worth reading the commit message for 2502ffdfca07fac6972c9b2da7ae160d011c2877

The documentation for PadnameUTF8() mentions:
```
Whether PadnamePV is in UTF-8.  Currently, this is always true.
```
Do we want to commit to never changing that?  (I have no problem with that.)

PR #21994, opened 2024-02-15
Remove forced proto-changes in HP-UX hints for old pthreads Draft 4

Tux:


hints/hpux.sh
```diff
@@ -789,17 +789,10 @@ d_isnan='define'
...
-# interfaces.
-case "$d_oldpthreads" in
-    ''|$undef)
-	d_crypt_r_proto='undef'
-	d_getgrent_r_proto='undef'
-	d_getpwent_r_proto='undef'
-	d_strerror_r_proto='undef'
-	;;
-    esac
+# Old versions of pthreads (Draft 4) might require the following vaiables
```
jkeenan: Typo: `s/vaiables/variables/`

khwilliamson:
After doing some more research, I don't believe that it is C99 that allows us to make this change.  There is nothing about threads in C99; it is from POSIX.  Instead, I think the following gives the rationale.  The patch that created the prototype undefinitions is

>  commit 90e831dcaea7bb76a1f6fd13b3d99e8db4a40ec1
 Author: Spider Boardman <spider@orb.nashua.nh.us>
 Date:   Thu Mar 21 04:23:45 2002 -0500
  
     More HP-UX/IA64 work
     Message-Id: <200203211423.JAA61178@leggy.zk3.dec.com>
     
     p4raw-id: //depot/perl@15388

The comment (removed in this pull request) said that it was because of `pthreads Draft 4`. 

https://engineering.purdue.edu/~smidkiff/KKU/files/PThreads.pdf says this:

> DCE threads are based on draft 4 (an early draft) of
the POSIX threads standard (which was originally
named 1003.4a, and became 1003.1c upon
standardization).

I couldn't find a date for either draft 4 or  DCE threads, but DCE are described in the  [HP Journal Dec 1995](https://www.hpl.hp.com/hpjournal/95dec/dec95a1.pdf)

There is this quote also from https://engineering.purdue.edu/~smidkiff/KKU/files/PThreads.pdf

>  "Real" POSIX threads, based on the IEEE POSIX
1003.1c-1995 (also known as the ISO/IEC
9945-1:1996) standard, part of the ANSI/IEEE
1003.1, 1996 edition, standard. POSIX
implementations are, not surprisingly, the standard on
Unix systems. POSIX threads are usually referred to
as Pthreads.

So the draft 4 was still around in 2002 when the commit was added.  But
https://docstore.mik.ua/manuals/hp-ux/en/B2355-60130/pthread.3T.html mentions the 1003.1c standard.  It is for

> HP-UX 11i Version 3: February 2007

So it appears that by then that HP had converted to Standardized pthreads.  Though this document doesn't mention them 
https://support.hpe.com/hpesc/public/docDisplay?docId=pdbPHCO_42326&docLocale=en_US
perhaps because the change predated it.

I think that the manual from 2007 mentioning 1003.1c as what is used, and not mentioning a 12+ year old draft indicates that "real" pthreads are in use as of at least then

khwilliamson (approved):

PR #21997, opened 2024-02-16
Issue deprecation warnings on subsequent `use VERSION` declarations

leonerd:
We'll want to keep an eye on tests for this one because it seems quite likely it may invoke new warnings in existing code.


jkeenan:
> We'll want to keep an eye on tests for this one because it seems quite likely it may invoke new warnings in existing code.

IMO we *should* set a sunset version for this.  Several years ago @Abigail went through the code base, identified all deprecations and got us to set dates (generally 1 or 2 years out) when those warnings got converted into exceptions.  I think that has worked well; we have had few, if any, bug tickets complaining that we should have kept something in deprecated status longer than we did.  I further doubt that the practice of mixing `use VERSION` declarations in the same scope is widespread -- but people who have done so should speak up now!

leonerd:
On account of the fact it might well provoke plenty of warnings around places, I decided to set it at 3 versions out, rather than the usual 2. It's not really blocking any fundamentally new features or changes in the language, it's just a big lot of cleanup, so I think the extra time is justified.

Though I'm kindof hoping that we won't have a "5.46" and instead it'll be a "7.x" for some x ;)

pod/perldiag.pod
```diff
@@ -7852,6 +7852,30 @@ even if it requests a later version.
...
 
+=item use VERSION while another use VERSION is in scope is now deprecated
+
+(W deprecated) Once you have a C<use VERSION> statement in scope, any other
+C<use VERSION> statement is now deprecated, due to the increasing complexity
+of swapping from one prevalent version to another.
+
+It is suggested that you do not try to mix multiple different version
+declarations within the same file as it leads to complex behaviours about the
+visibiity of features and builtin functions, as well as confusing human
```
jkeenan: `s/visibiity/visibility/`

op.c
```diff
@@ -8030,6 +8030,10 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
             /* downgrading from >= 5.11 to < 5.11 is now fatal */
             if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11))
                 croak("Downgrading a use VERSION declaration to below v5.11 is not permitted");
+
+            /* OK lets at least warn */
```
jkeenan: `s/lets/let's/`

jkeenan (changes requested):
Small stuff.

op.c
leonerd: fixed

pod/perldiag.pod
leonerd: fixed

leonerd:
@jkeenan is this good now?

pod/perldeprecation.pod
```diff
@@ -60,6 +60,18 @@ in deprecation warnings.
...
 
+=head2 Perl 5.46
+
+=head3 C<use VERSION> while another C<use VERSION> is in scope
+
+A C<use VERSION> declaration has many implicit effects on the surrounding
+scope, such as L<strict> and L<feature> flags, or importing L<builtin>
+functions. Once you have a C<use VERSION> statement in scope, any other
+C<use VERSION> statement is now deprecated since Perl 5.39.8, due to the
+increasing complexity of swapping from one prevalent version to another.
```
ilmari: Typo: "prevalent" should be "prevailing". Also in `perldiag.pod`.

ilmari (changes requested):

pod/perldeprecation.pod
leonerd: Oops; yes I think I was even the one who came up with the word "prevailing" in the first place.

ilmari (approved):

jkeenan (approved):

PR #21993, opened 2024-02-15
Yves/fix static build regex engine

demerphq:
This fixes build issues in the regex engine with -Uusedl. It fixes make test_reonly in this build mode, and it fixes the issues reported in Issue #21558 regarding reg_mesg.t failing under -Uusedl. 

@tonycoz this should do it I think.

PR #22000, opened 2024-02-17
t/re/reg_mesg.t - escape non-visible codepoints in test names

demerphq:
In GH #21993 James Keenan pointed out that this test scrambles the terminal by outputting raw utf8, possibly malformed as part of the test names. This makes it hard to understand what is going on with the test. This patch fixes the test names so that the code snippets we output are escaped. We use a custom sub to do the escaping so that we do not use the regex engine at all to do the escaping.

See https://github.com/Perl/perl5/pull/21993#issuecomment-1945981506 for details of the problems James observed.

With this patch the output should not break any terminals.

### DIFF
--- a/t/re/reg_mesg.t
+++ b/t/re/reg_mesg.t
@@ -20,6 +20,19 @@ use open qw(:utf8 :std);
 use warnings;
 BEGIN { ${^WARNING_BITS} = undef }  # Kludge to restore default warnings
 
+# NOTE:
+# This file uses display_rx() from t/test.pl, which escapes
+# non-printable codepoints and percentage symbols as non-standard
+# "%x{01+aa+bc+02}" style sequences.  We do this as we may be testing
+# malformed utf8 sequences or other non-printable characters which we
+# want to distinguish from any actual regex pattern escape sequences,
+# while at the same time we don't want to emit any non-printable
+# sequences to a terminal as literals where they may cause problems.
+# Thus "\x{01}" means the string contained six codepoints making up a
+# normal "\x{...}" style escape sequence whereas "%x{01}" means the string
+# contained a single codepoint and "%x{01+02}" means it contained two
+# codepoints, etc. - Yves
+
 # Kind of a kludge to mark warnings to be expected only if we are testing
 # under "use re 'strict'"
 my $only_strict_marker = ':expected_only_under_strict';
@@ -804,14 +817,14 @@ for my $strict ("", "use re 'strict';") {
                     my $eval_string = "$strict $regex";
                     $_ = "x";
                     eval "$eval_string; \$meaning_of_life = 42";
-                    ok (! defined $meaning_of_life, "$eval_string died");
+                    ok(! defined $meaning_of_life, display_rx("$eval_string died"));
                     my $error= $@;
                     if ($error =~ qr/\Q$expect/) {
                         ok(1, "... and gave expected message");
                     } else {
-                        ok(0,$eval_string);
-                        diag("Have: " . _q(add_markers($error)));
-                        diag("Want: " . _q($death[$i+1]));
+                        ok(0, display_rx($eval_string));
+                        diag("Have: " . display_rx(add_markers($error)));
+                        diag("Want: " . display_rx($death[$i+1]));
                     }
                 }, undef, "... and no other warnings");
         }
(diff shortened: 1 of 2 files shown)


t/re/reg_mesg.t
```diff
@@ -20,6 +20,25 @@ use open qw(:utf8 :std);
...
+# written to not use any regex logic. Currently it assumes that we are
+# using ASCII/Unicode encoding. Maybe one day we will extend it to
+# handle EBCDIC, although I am not sure it is necessary. - Yves
+sub escape_string {
+    my ($str) = @_;
+    my $escaped = "";
+    for my $i (0 .. length($str)-1) {
+        my $char = substr($str,$i,1);
+        my $ord = ord($char);
+        if (32 <= $ord <= 127) {
```
mauke: Should be `126` (or `< 127`) for printable ASCII.

For unambiguous output, there should be another condition: `... && $char ne "\\"`

jkeenan:
> In GH #21993 James Keenan pointed out that this test scrambles the terminal by outputting raw utf8, possibly malformed as part of the test names. This makes it hard to understand what is going on with the test. This patch fixes the test names so that the code snippets we output are escaped. We use a custom sub to do the escaping so that we do not use the regex engine at all to do the escaping.
> 
> See [#21993 (comment)](https://github.com/Perl/perl5/pull/21993#issuecomment-1945981506) for details of the problems James observed.
> 
> With this patch the output should not break any terminals.

Thanks.  I tested this branch on Linux (both locally and over SSH), FreeBSD and OpenBSD.  I got no spewing anywhere, nor did I freeze and terminals.  Subject to @mauke's concerns, this p.r. looks good.

t/re/reg_mesg.t
demerphq: Good catch. I changed the end range to 126 as you suggested, but I'm not sure why we need the second part with `"\\"` as it is codepoint  92 (0x5c) , so it shouldn't be escaped anyway.

mauke: In case you want to distinguish `escape_string("\\\0")` from `escape_string("\\x{00}")`. Currently they produce `\\x{00}`.

demerphq: Regarding the backslash, did you mean to escape backslashes anyway? I guess at a certain level that makes sense, but  if we go down that road maybe we should do what the regex debugger does and use % instead of backslash so that actual backslashes in the input stay as backslashes. I doubt if any of our tests include a percentage symbol. Having said that I doubt its necessary, we just need enough info to work out which test failed.

demerphq: oh weird, the page didnt update automatically, so i didnt see your most recent comment. I guess ill change it to use %.

demerphq: Ok, i just updated it. `escape_string("\\\0")` should output `\%x{00}` and `escape_string("\\x{00}")` should output `\x{00}`

t/re/reg_mesg.t
```diff
@@ -20,6 +20,36 @@ use open qw(:utf8 :std);
...
+sub escape_string {
+    my ($str) = @_;
+    my $escaped = "";
+    for my $i (0 .. length($str)-1) {
+        my $char = substr($str,$i,1);
+        my $ord = ord($char);
+        if ($char eq "%") {
+            $escaped .= "%%";
+        }
+        elsif (32 <= $ord <= 126) {
```
khwilliamson: This won't work on non-ASCII platforms.  Instead write 
`elsif (32 <= utf8::native_to_uni($ord) <= 126) {`
The extra function call is optimized out on ASCII platforms.

t/re/reg_mesg.t
```diff
@@ -20,6 +20,36 @@ use open qw(:utf8 :std);
 use warnings;
 BEGIN { ${^WARNING_BITS} = undef }  # Kludge to restore default warnings
 
+# Escape a string in a similar fashion to how the regex debugger does,
+# using "x" style escapes in the form "%x{..}" to show the codepoint of
+# the escaped value and "%%" to show a literal percentage sign.  We use
+# percentage instead of backslash so that it is trivial to distinguish
+# backslash based sequences that are in the pattern from escaped octets.
+# Currently we assume ASCII/Unicode semantics, and escape anything that
```
khwilliamson: After the change below, this is no longer true.  I suggest something like

> We escape anything that isn't an ASCII printable character

The 32 and 126 are less important

khwilliamson (changes requested):

t/re/reg_mesg.t
demerphq: Thanks. Done. BTW, the function is utf8::native_to_unicode().  I believe the C version is "uni", but in the perl level we spell it out.

t/re/reg_mesg.t
demerphq: Comment has been updated.

PR #22002, opened 2024-02-17
Simplify `use VERSION` behaviour with respect to imported builtins

leonerd:
Having `use VERSION` attempt to unimport builtin functions has proven to be an overly-complex and terrible design, which lead to the "tombstone" pad idea.

We've subsequently decided that was a mistake and wish to remove it. This is the first step of doing so. As it was only recently added in a development version and has not been in any stable release, there's no need to deprecate or warn about it - we can simply remove it.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -636,7 +636,7 @@ static bool S_cv_is_builtin(pTHX_ CV *cv)
 }
 
 void
-Perl_import_builtin_bundle(pTHX_ U16 ver, bool do_unimport)
+Perl_import_builtin_bundle(pTHX_ U16 ver)
 {
     SV *ampname = sv_newmortal();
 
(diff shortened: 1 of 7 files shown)


tonycoz:
Do the problems with `use VERSION` downgrades removing subs also apply to `no builtin "somefunc";`?

haarg:
Yes. We also intend to remove `no builtin`.

leonerd:
> Yes. We also intend to remove `no builtin`.

In fact I already have a mostly-finished branch locally on my laptop to do just this. I just don't like having too many large (in terms of behaviour) changes up in the air in PRs at once, so there's a reasonably small surface for folks to review. I'll rebase + PR that one once this one is merged.

ilmari (approved):
LGTM

PR #22004, opened 2024-02-18
class.c: Add a comment explaining how class_wrap_method_body works

leonerd:
After a recent email discussion with @iabyn I thought it best to better explain how this part works.


tonycoz:
This looks reasonable, but I didn't (don't) see the discussion that led to it.

PR #22006, opened 2024-02-19
Silence inconsequential warnings in win32/perlhost.h.

sisyphus:
I don't know why these functions that are emitting the warnings are even declared in perlhost.h.
And I don't know why they insist on taking an unused arg.
Maybe there's a better approach to silencing these warnings than the one I have chosen.

Anyway, they silence the warnings for me, and don't break anything in the process (AFAICS).
Hopefully, the CI tests will confirm this.


tonycoz:
These functions are all called via a table of function pointers, and the callers supply that first parameter.

They aren't used currently in perl, but that could change in the future, it's possible they're used by the ActivePerl ISAPI plugin.

tonycoz (approved):

PR #22007, opened 2024-02-19
add OPpCONST_LINE flag to aid deparsing

iabyn:
Code like

    my $line = __LINE__;

was being deparsed as

    my $line = '42';

because __LINE__ is converted by the lexer into an ordinary OP_CONST whose SV is just a plain string holding the line number.

This commit adds a flag, OPpCONST_LINE, to OP_CONST, which is set to indicate that this OP_CONST was created from a __LINE__ construct.

Deparse.pm then uses this new flag to print out the literal string "__LINE__" rather than the constant's actual value.

This fixes one failing test file under
    cd t; ./TEST -deparse
and removes one (but not all) sources of failure from three others.

### DIFF
--- a/lib/B/Deparse.t
+++ b/lib/B/Deparse.t
@@ -3391,3 +3391,8 @@ $x = (return 'ok');
 $x //= (return 'ok');
 $x = exit 42;
 $x //= exit 42;
+####
+# preserve __LINE__ etc
+my $x = __LINE__;
+my $y = __FILE__;
+my $z = __PACKAGE__;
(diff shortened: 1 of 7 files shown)


tonycoz (approved):
The change itself looks file.

Does `__FILE__` need a similar change?

iabyn:
I've now updated the commit so that it uses two bits of OP_CONST's  op_private to deparse all three of __LINE__, __FILE__, __PACKAGE__.

lib/B/Deparse.t
```diff
@@ -3391,3 +3391,6 @@ $x = (return 'ok');
 $x //= (return 'ok');
 $x = exit 42;
 $x //= exit 42;
+####
+# preserve __LINE__ etc
+my $x = __LINE__;
```
tonycoz, line 3396: This could use tests for `__FILE__` and `__PACKAGE__` now that you're adding support for them.

iabyn:
Yeah, I meant to update the test but forgot. Done now, and rebased and repushed.

tonycoz (approved):

PR #22009, opened 2024-02-19
Remove the `builtin::unimport` function and `no builtin ...`

leonerd:
Following on from #22002, this removes the last place that `PADNAMEf_TOMBSTONE` is used. Once this is removed we can get rid of all the tombstone logic, and close a few of the related bugs / PRs with it.

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -674,7 +674,7 @@ XS(XS_builtin_import)
         SV *sym = ST(i);
         STRLEN symlen;
         const char *sympv = SvPV(sym, symlen);
-        if(strEQ(sympv, "import") || strEQ(sympv, "unimport"))
+        if(strEQ(sympv, "import"))
             Perl_croak(aTHX_ builtin_not_recognised, sym);
 
         if(sympv[0] == ':') {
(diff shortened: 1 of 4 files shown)


builtin.c
```diff
@@ -707,44 +707,7 @@ XS(XS_builtin_import)
...
-         * same time as this tombstone would, we could not bother adding the
-         * tombstone and instead COP_SEQ_MAX_HIGH_set() on the padname to
-         * clear it.
-         */
-        pad_add_name_sv(ampname, padadd_STATE|padadd_TOMBSTONE, 0, 0);
-    }
-
-    COP_SEQMAX_INC;
-    finish_export_lexical();
+    croak("builtin::unimport has now been removed");
```
ilmari: `builtin::unimport` was never in a stable release, so there's no point in leaving this tombstone here, just remove it completely.

pod/perldiag.pod
```diff
@@ -657,12 +657,24 @@ is currently being compiled. Since this method is used to introduce new
...
+and introduce many subtle questions about the nature of hiding lexical
+functions, so was reomved in Perl 5.39.8.
+
+If you wish to limit the scope of visiblity of an imported C<builtin>
+function, put it inside its own scope:
+
+    {
+      use builtin 'refaddr';
+      ...
+    }
```
ilmari: Instead of mentioning that workaround here, it should be in the main `builtin` docs, where the `no builtin` was mentioned previously.

ilmari (changes requested):

builtin.c
leonerd: Indeed - I'll do that before I undraft the PR. I just wanted to see the errors during some smoke-tests first.

leonerd: done

pod/perldiag.pod
leonerd: done

ilmari (approved):

tonycoz (approved):

PR #22011, opened 2024-02-20
perldelta 4fc2379: make a new stub to clone into when pushing a new pad

tonycoz:
If anyone has better wording for the second issue I'd love to see it.

### DIFF
--- a/pod/perldelta.pod
+++ b/pod/perldelta.pod
@@ -476,7 +476,50 @@ manager will later use a regex to expand these into links.
 
 =item *
 
-XXX
+Lexical subs now have a new stub in the pad for each recursive call
+into the containing function.  This fixes two problems:
+
+=over
+
+=item *
+
+If the lexical sub called the containing function, a "Can't undef
+active subroutine" error would be thrown.  For example:
+
+    use v5.36.0;
+    sub outer($oc) {
+        my sub inner ($c) {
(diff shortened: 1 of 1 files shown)


pod/perldelta.pod
```diff
@@ -476,7 +476,50 @@ manager will later use a regex to expand these into links.
...
+inner recursive call:
+
+    use v5.36.0;
+    sub outer ($x) {
+        my sub inner ($label) {
+            say "$label $x";
+        }
+        inner("first");
+        outer("inner") if $x eq "outer";
+        # $x visible inside this call to inside is incorrect
```
mauke: Should be `call to inner`. But maybe rephrase as `this call to inner() sees the wrong $x`?

pod/perldelta.pod
```diff
@@ -476,7 +476,50 @@ manager will later use a regex to expand these into links.
 
 =item *
 
-XXX
+Lexical subs now have a new stub in the pad for each recursive call
+into the containing function.  This fixed two problems:
```
mauke: Not sure about tenses in English. Is it "This fixed" or "This fixes"?

mauke:
The rest looks good to me.

pod/perldelta.pod
khwilliamson: s/fixed/fixes/

mauke (approved):

Issue #22015, opened 2024-02-21
new 'use VERSION' warnings in some core modules


iabyn:
**Description**
Since v5.39.7-198-g351de214d2, with the  the addition this warning:
```
use VERSION while another use VERSION is in scope is deprecated
```
a number of core modules have started emitting this warning on stderr during perl's 'make test'. It would be good to get these silenced before 5.40.
Here's the relevant output from 'make test':
```
lib/B/Deparse .................................................... use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 211) line 17, <DATA> chunk 157.


dist/Devel-SelfStubber/t/Devel-SelfStubber ....................... use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.
use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.
use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.
use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.
use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.



dist/SelfLoader/t/01SelfLoader ................................... use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.

dist/SelfLoader/t/02SelfLoader-buggy ............................. use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.

dist/SelfLoader/t/03taint ........................................ use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 5) line 1.
```

leonerd:
I can possibly find time to look at these some evening of my vacation but if not I'll be happy to review anyone else's PR if they get there first ;)

jkeenan:
> **Description** Since v5.39.7-198-g351de214d2, with the the addition this warning:
> 
> ```
> use VERSION while another use VERSION is in scope is deprecated
> ```
> 
> a number of core modules have started emitting this warning on stderr during perl's 'make test'. It would be good to get these silenced before 5.40. Here's the relevant output from 'make test':

[...]

PR #22016, opened 2024-02-21
allow multiple use v5.x declarations in some cases where it's harmless

haarg:
allow multiple use v5.x declarations in some cases where it's harmless

We want to disallow using multiple "use v5.x" declarations in the same scope, but there are some cases we know to be harmless.

For versions before 5.10, the use declaration was only a version check with no side effects. If there was a previous version declarations that was below 5.10, we can safely ignore it without a warning or error.

### DIFF
--- a/op.c
+++ b/op.c
@@ -8022,18 +8022,25 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
             /* use VERSION while another use VERSION is in scope
              * This should provoke at least a warning, if not an outright error
              */
-            if (shortver >= SHORTVER(5, 39))
+            if (PL_prevailing_version < SHORTVER(5, 10)) {
+                /* if the old version had no side effects, we can allow this
+                 * without any warnings or errors */
+            }
+            else if (shortver >= SHORTVER(5, 39)) {
                 croak("use VERSION of 5.39 or above is not permitted while another use VERSION is in scope");
-            if (PL_prevailing_version >= SHORTVER(5, 39))
+            }
+            else if (PL_prevailing_version >= SHORTVER(5, 39)) {
                 croak("use VERSION is not permitted while another use VERSION of 5.39 or above is in scope");
-
-            /* downgrading from >= 5.11 to < 5.11 is now fatal */
-            if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11))
+            }
+            else if (PL_prevailing_version >= SHORTVER(5, 11) && shortver < SHORTVER(5, 11)) {
+                /* downgrading from >= 5.11 to < 5.11 is now fatal */
                 croak("Downgrading a use VERSION declaration to below v5.11 is not permitted");
-
-            /* OK let's at least warn */
-            deprecate_fatal_in(WARN_DEPRECATED__SUBSEQUENT_USE_VERSION, "5.46",
+            }
+            else {
+                /* OK let's at least warn */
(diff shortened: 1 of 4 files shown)


jkeenan:
I checked out this branch and built it.  It has the advantage of cleaning up the warnings emitted during `make test` for all of the test files reported by @iabyn in https://github.com/Perl/perl5/issues/22015 with the exception of `lib/B/Deparse.t`.
```
export TEST_FILES="../dist/Devel-SelfStubber/t/Devel-SelfStubber.t ../dist/SelfLoader/t/02SelfLoader-buggy.t ../dist/SelfLoader/t/01SelfLoader.t ../dist/SelfLoader/t/03taint.t ../lib/B/Deparse.t"; cd t; TEST_JOBS=1 ./perl harness "$TEST_FILES"; cd -
../dist/Devel-SelfStubber/t/Devel-SelfStubber.t .. ok     
../dist/SelfLoader/t/02SelfLoader-buggy.t ........ ok   
../dist/SelfLoader/t/01SelfLoader.t .............. ok     
../dist/SelfLoader/t/03taint.t ................... ok     
../lib/B/Deparse.t ............................... 1/? use VERSION while another use VERSION is in scope is deprecated, and will become fatal in Perl 5.46 at (eval 211) line 17, <DATA> chunk 157.
../lib/B/Deparse.t ............................... ok     
All tests successful.
```
I have not reviewed the code changes *per se*, nor have I yet tested its impact on those 2 BBCs.

mauke:
Why is v5.11 the cut-off point? `use v5.10;` already had semantic effects like enabling `state` or `say`.

haarg:
> Why is v5.11 the cut-off point? `use v5.10;` already had semantic effects like enabling `state` or `say`.

Yeah, I was wrong here. I was thinking about strict, which wasn't applied until 5.12

Leont (approved):

Leont:
Logic looks good to me (but also what mauke said)

haarg:
I've updated this to only allow versions before 5.10, but also it now only checks that the previous version is < 5.10. So `use v5.8; use v5.12;` is allowed.

op.c
```diff
@@ -8022,18 +8022,25 @@ Perl_utilize(pTHX_ int aver, I32 floor, OP *version, OP *idop, OP *arg)
             /* use VERSION while another use VERSION is in scope
              * This should provoke at least a warning, if not an outright error
              */
-            if (shortver >= SHORTVER(5, 39))
+            if (PL_prevailing_version < SHORTVER(5, 10)) {
```
mauke, line 8025: Why not `... || shortver < SHORTVER(5, 10)`, too? It seems a bit weird that `use v5.8; use v5.12;` is accepted because `use v5.8` has no side effects, but `use v5.12; use v5.8;` is an error.

haarg: Perl starts with implicit 5.8 semantics. So changing from 5.8 to 5.8 does nothing. `use` doesn't just add behaviors, but changes all of the semantics to match the requested version. Based on the existing behavior of use, the `use v5.8` in `use v5.12; use v5.8;` would have side effects, and disabling the 5.12 features and turning off strict. This behavior is weird enough (and I don't believe was ever implemented) that it has already been deprecated for 2 releases, and was already intended to be fatal in 5.40 even before we started looking at other changes.

mauke: Oh, looks like I had misunderstood `use VERSION` semantics entirely. :-)
Thanks for setting me straight.

mauke (approved):

PR #22024, opened 2024-02-25
perl.h: Move code to later in file

khwilliamson:
This moves the portion of the code dealing with toggling LC_NUMERIC to after the portion dealing with the locale and environment mutexes; thus making those related pieces adjacent.

The diff listing however shows that env mutex code being moved, due to the 'diff' algorithm.


tonycoz (approved):
WRT diff, you've effectively swapped those segments, so diff is just showing the smaller of the two reasonable diffs. (around 350 lines for the code you moved, around 160 for the block diff picked up).

PR #22025, opened 2024-02-25
perl.h: Move some code to earlier in the file

khwilliamson:
This code for defining the ENV mutex is more logically placed ahead of the locale stuff, just after the other mutex code, because the locale stuff depends on it.


khwilliamson:
Further investigation showed I was wrong about this.

PR #22030, opened 2024-02-26
Automatically regenerate locale_table.h

khwilliamson:
I got tired of editing each entry when something new comes along.

Doing this opens up new possibilities of simplification, but this commit
tries to keep the generated header as close to the original as possible.


regen/locale.pl
```diff
@@ -0,0 +1,123 @@
...
+    require './regen/regen_lib.pl';
+}
+
+use strict;
+use warnings;
+
+sub open_print_header {
+    my ($file, $quote) = @_;
+    return open_new($file, '>',
+                    { by => 'regen/reentr.pl',
```
tonycoz: `by` should be `regen/locale.pl`

tonycoz:
otherwise fine

PR #22031, opened 2024-02-26
Tie::File: use File::Temp in two test scripts

iabyn:
t/29_downcopy.t and t/29a_upcopy.t intermittently timeout on some smokers. This is speculated to be due to the current directory being on a very slow USB drive, when lots of small test files are created, modified and deleted by these tests.

Update these two test scripts to use File::Temp to create a temp subdir in the OS's normal temp directory as a place for the temp files.

This has two potential advantages. First the OS's normal tempdir (e.g.  /tmp or wherever $TMPDIR points to) may be mounted as tmpfs or similar, and thus won't actually write to a slow USB or network drive.  Second, each test file will be created in directory that's normally empty, so there's potentially less manipulating of the directory information on disk.


khwilliamson (approved):

t/harness
```diff
@@ -29,14 +29,6 @@ my @_must_be_executed_serially = (
...
     '../ext/File-Find/t/taint.t',
     '../ext/File-Find/t/find.t',
-
-    # CJKT.t creates some temporary files called "t/$$.utf", while
-    # Unicode.t expects to find a fixed number of *.utf files in that
-    # directory. Running the tests in parallel can cause failures like:
-    # "Bad plan.  You planned 56 tests but ran 57."
-    # Reported as https://rt.cpan.org/Ticket/Display.html?id=151983
-    'cpan/Encode/t/CJKT.t',
-    'cpan/Encode/t/Unicode.t',
```
mauke: Why remove these lines from `t/harness`?

mauke (approved):

jkeenan (approved):

iabyn:
On Mon, Feb 26, 2024 at 10:57:13AM -0800, mauke wrote:
> Why remove these lines from `t/harness`?

They're from an unrelated commit that accidentally got included in the PR
branch.

-- 
I before E. Except when it isn't.

PR #22033, opened 2024-02-26
Tie::File: update code for v5.8 compatibility

mauke:
- Ever since 5926ffcd2cf97, Tie::File has been using warnings.pm and 'our', making the code incompatible with 5.005 ('use warnings' wouldn't even compile anymore), so remove the old compatibility code.
- Update code for v5.8 (the "Unknown discipline" error from v5.6 was changed to "IO layers (...) unavailable" then).
- 'use constant' to declare constants.
- Remove redundant 'use strict "vars"' (the whole file is under 'use strict').


dist/Tie-File/lib/Tie/File.pm
```diff
@@ -1451,14 +1442,15 @@ package Tie::File::Cache;
...
-#sub REQ  () { 6 } # Total number of cache requests 
-use strict 'vars';
+use constant {
+    HEAP  => 0,
+    HASH  => 1,
+    MAX   => 2,
+    BYTES => 3,
+    #STAT  => 4, # Array with request statistics for each record
+    #MISS  => 5, # Total number of cache misses
+    #REQ   => 6, # Total number of cache requests
```
jkeenan, line 1452: Why are we retaining these 3 commented-out lines?

jkeenan:
What is the rationale for the use of `use constant XXX`?

tonycoz:
> What is the rationale for the use of `use constant XXX`?

`use constant` tells the reader you're creating constants, `sub X () { num }` is a "magical" invocation that happens to create a constant.

dist/Tie-File/lib/Tie/File.pm
tonycoz: They're used by other commented out code that records stats.

That code could be removed, or it could be uncommented (and maybe controllable by a constant), but such a change would serve a different purpose than this PR.

tonycoz (approved):

PR #22035, opened 2024-02-26
Preserve split " " behaviour when whitespace modifiers are present

### LINKED ISSUE #18032: use re '/aansx'; breaks split " ", ...
$ perldoc -f split
[...] As another special case, "split" emulates the default behavior of the command line tool awk when the PATTERN is either omitted or a string composed of a single space character
[...] In this case, any leading whitespace in EXPR is removed before splitting occurs, and the PATTERN is instead treated as if it were "/\s+/";
[...] in Perl 5.18.0 and later this special case is triggered by any expression which evaluates to the simple string " ".
[...] If omitted, PATTERN defaults to a single space, " ", triggering the previously described awk emulation.
[...]

```
$ perl -Mstrict -wE'

  my $s= "  foo  bar  ";

  say for split " ", $s;

  { use re qw( /aansx );
    say for split " ", $s;
  }
'
```
```
foo
bar
 
 
f
o
o
 
 
b
a
r 



```
```
---
Site configuration information for perl 5.24.3:

Configured by root at Tue Feb 13 14:35:15 CET 2018.
[...]
Locally applied patches:
    fix for #132063
    fix for #132227
[...]
```

richardleach:
Prior to this commit, the " " pattern would be collapsed into the empty "" pattern if the split was within scope of `use re` whitespace modifiers. Since this split pattern is documented as being a special case, that behaviour is undesirable.

Fixes #18032

### DIFF
--- a/regcomp.c
+++ b/regcomp.c
@@ -2222,9 +2222,20 @@ Perl_re_op_compile(pTHX_ SV ** const patternp, int pat_count,
         /* It's safe to read through *next only if OP(first) is a regop of
          * the right type (not EXACT, for example).
          */
-        if (REGNODE_TYPE(fop) == NOTHING && nop == END)
-            RExC_rx->extflags |= RXf_NULL;
-        else if ((fop == MBOL || (fop == SBOL && !FLAGS(first))) && nop == END)
+        if (REGNODE_TYPE(fop) == NOTHING && nop == END) {
+            if ((RExC_rx->extflags & RXf_SPLIT)
+                && pat_count == 1
+                && OP_TYPE_IS(expr, OP_CONST)
+                && SvPOK(cSVOPx_sv(expr)) && SvCUR(cSVOPx_sv(expr)) == 1
+                && SvPV_nolen(cSVOPx_sv(expr))[0] == ' '
+            ){
+                /* GH 18032: split " " behaviour should remain unchanged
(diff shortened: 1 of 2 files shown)


demerphq (commented):
Can you please hold off merging this, the change doesn't look right to me. I spent a fair bit of time removing logic very much like this in the past as it is very fragile and we really should have a special metapattern for this as well. I will post a more detailed review in a day or two.

tonycoz:
I think it could use a behaviour test - ie. testing the results of `split " "` where it was broken rather than just testing the regexp compiler dump.

richardleach:
I'll wait for @demerphq's review before making any changes.

demerphq:
@richardleach I pushed https://github.com/Perl/perl5/pull/22069

I think we should just be exempting `split STRING` from `use re '/x'`. IMO it doesn't make sense to allow /x to apply to string arguments to split. We dont't apply the same metacharacter rules in split STRING and split PATTERN, and i feel like /x is about changing how a pattern is parsed, not how a string literal is parsed, an interpretation that means the special case of split " " is preserved. THis has the advantage that something like:

```
my $prefix = "";
my $infix= " ";
my $suffix = "";

my @parts= split $prefix . $Infix . $suffix;
```

triggers the expected awklike emulation under `use re "/x"`;

richardleach:
Ok, I'll close this PR in favour of #22069.

PR #22036, opened 2024-02-27
Update perlref.pod to warn $f{B} = $f{A} is only copying a pointer

jidanni:
Else people wonder how do things get injected into their data.

Please edit if needed.

### DIFF
--- a/pod/perlref.pod
+++ b/pod/perlref.pod
@@ -117,6 +117,16 @@ brackets:
         'Clyde' => 'Bonnie',
     };
 
+But be aware:
+
+    $folks{A} = {
+        'Adam'  => 'Eve',
+        'Clyde' => 'Bonnie',
+    };
+    $folks{B} = $folks{A};           # just pointers to the anonymous hash
+    %{ $folks{C} } = %{ $folks{A} }; # deep copy
+    $folks{A}{Clyde} = 'Wilma';      # also updates B! C is immune
+
 Anonymous hash and array composers like these can be intermixed freely to
 produce as complicated a structure as you want.  The multidimensional
 syntax described below works for these too.  The values above are


pod/perlref.pod
```diff
@@ -117,6 +117,16 @@ brackets:
         'Clyde' => 'Bonnie',
     };
 
+But be aware:
+
+    $folks{A} = {
+        'Adam'  => 'Eve',
+        'Clyde' => 'Bonnie',
+    };
+    $folks{B} = $folks{A};           # just pointers to the anonymous hash
```
guest20, line 126: It's a reference, not a pointer.

pod/perlref.pod
```diff
@@ -117,6 +117,16 @@ brackets:
...
+    %{ $folks{C} } = %{ $folks{A} }; # deep copy
```
guest20, line 127: This is a shallow copy, not a deep copy.

pod/perlref.pod
```diff
@@ -117,6 +117,16 @@ brackets:
...
+    $folks{A}{Clyde} = 'Wilma';      # also updates B! C is immune
```
guest20, line 128: "$folks{B}" isn't a _thing_, it's the name of a _reference_ to a thing.

It might be worth adding a whole paragraph up top about what it means for two references to point to the same structure (even if the reference is stored in different places) instead of including a 6 line code example in the middle of a sentence.

pod/perlref.pod
jidanni: I think it's great you folks could put something better into that page than my version.

I even had my first accident today hitting the resolve button.

jkeenan:
Closing this ticket, as there has been no further discussion in five months and no one has expressed support for the change.

PR #22037, opened 2024-02-27
Update perlfunc.pod to document die newline workarounds

jidanni:
Else the user cannot find where the program died!

Welcome to enhance my patch.

### DIFF
--- a/pod/perlfunc.pod
+++ b/pod/perlfunc.pod
@@ -1826,7 +1826,23 @@ and concatenated to make the exception.
 
 If the string exception does not end in a newline, the current
 script line number and input line number (if any) and a newline
-are appended to it.  Note that the "input line number" (also
+are appended to it.  
+
+Thus to avoid ending in a newline one might sometimes need
+workarounds, e.g.,
+
+    # Workaround for a function call that ends in a newline:
+    use Data::Dumper;
+    die Dumper(\%h)," ";
+
+    # Here-doc workaround:
+    my $m = <<EOF;
+    Huss, buss,
+    what a fuss.
+    EOF
+    die "$m "; #or chomp $m; die $m;
+
+Note that the "input line number" (also
 known as "chunk") is subject to whatever notion of "line" happens to
 be currently in effect, and is also available as the special variable
 L<C<$.>|perlvar/$.>.  See L<perlvar/"$/"> and L<perlvar/"$.">.


mauke:
What does a user care where a program died?

jidanni:
Because s/he might have 50 files, each with 50 die statements, and finding which one died would be a nightmare.

khwilliamson (approved):

guest20:
It's only a nightmare to find which `die` is which if your `die` messages don't contain unique/useful messages.

You really have two problems bumping up against each other here and it mostly goes away if, as a point of order, you include a label before you `Dumper` something to an output. Pick something unique so you can `git grep` for it. 

I usually sub-class which ever logger the project uses with a `labeled_dump` method that dies if it doesn't get a label.

```perl
$app->log->labelled_dump( customer_credit_carrds => \@cards );
# Tue 27 [9203] request=b4dc0ffee [debug] customer_credit_cards [{type => amex....
```

Even in a single-file script using `Data::Dumper` directly is a bit of a pest, so I end up wrapping it in a similar sub (adding time, setting formatting etc)

jidanni:
The very very very very main source of problems, is the person way back one who thought it would be smart to make things act different depending on a final new line.

It might have seen like a cool idea way back when. But sure enough ...

Also one can I just use $. In here-documents, and then  getting the file name is a big drag too.

So okay you guys are welcome to put your ideas into the die documentation too.

guest20:
1. Most programs don't include line numbers when showing error messages to users... Try `ls --nonsense` or `docker evengreen` or `halt --hammerzeit` 

2. `die` is how one raises exceptions, and you don't want line numbers in the string your callers get when catching and checking your exception. 

3. If you want to blame the caller for an exception (say when they pass invalid arguments) you need to be able to `die` without the site where you call `die` being in the message (like carp)

4. Changing this behaviour breaks compat with existing code. 

Maybe you can write a perlcritic role to prevent this being done on accident if your own code?

jidanni:
I don't want to change the code.

But workarounds should be documented for when
we want to keep getting the line numbers, despite
having a newline in our back pocket.

Leont:
Yeah I don't think this is a good idea either

jkeenan:
> Yeah I don't think this is a good idea either

There has been no further conversation in this ticket in five months.  Closing.

PR #22039, opened 2024-02-27
Update Dumper.pm to hint at \( .... ) two birds one stone

jidanni:
Welcome to edit my patch.

### DIFF
--- a/dist/Data-Dumper/Dumper.pm
+++ b/dist/Data-Dumper/Dumper.pm
@@ -1416,8 +1416,9 @@ distribution for more examples.)
 =head1 BUGS
 
 Due to limitations of Perl subroutine call semantics, you cannot pass an
-array or hash.  Prepend it with a C<\> to pass its reference instead.  This
-will be remedied in time, now that Perl has subroutine prototypes.
+array or hash.  Prepend it with a C<\> to pass its reference instead:
+Dumper C<\>(%a, @b, $c), C<\>%e, C<\>@f, C<\>$g.
+This will be remedied in time, now that Perl has subroutine prototypes.
 For now, you need to use the extended usage form, and prepend the
 name with a C<*> to output it as a hash or array.
 
@@ -1459,6 +1460,6 @@ Version 2.189
 
 =head1 SEE ALSO
 
-perl(1)
+Data::Dump
 
 =cut


khwilliamson (approved):

khwilliamson:
This needs a version bump

dist/Data-Dumper/Dumper.pm
```diff
@@ -1416,8 +1416,9 @@ distribution for more examples.)
 =head1 BUGS
 
 Due to limitations of Perl subroutine call semantics, you cannot pass an
-array or hash.  Prepend it with a C<\> to pass its reference instead.  This
-will be remedied in time, now that Perl has subroutine prototypes.
+array or hash.  Prepend it with a C<\> to pass its reference instead:
+Dumper C<\>(%a, @b, $c), C<\>%e, C<\>@f, C<\>$g.
+This will be remedied in time, now that Perl has subroutine prototypes.
```
guest20, line 1421: Who is going to remedy this? 
At what time? 
How do prototypes remedy with this?
Are you thinking of signatures?

nicomen: Perhaps that's a question for the original commit including that statement? Althought history seems to stop here: https://github.com/Perl/perl5/commit/823edd999e1a252d36bcee7f4b8cc4bb197530e3

It seems however to refer to standard prototypes and not the new signatures.

guest20: My point was more that the vibes are wrong. Optimistic use of future tense and prototypes really don't mix in `$current_year`

dist/Data-Dumper/Dumper.pm
```diff
@@ -1416,8 +1416,9 @@ distribution for more examples.)
 =head1 BUGS
 
 Due to limitations of Perl subroutine call semantics, you cannot pass an
-array or hash.  Prepend it with a C<\> to pass its reference instead.  This
-will be remedied in time, now that Perl has subroutine prototypes.
+array or hash.  Prepend it with a C<\> to pass its reference instead:
+Dumper C<\>(%a, @b, $c), C<\>%e, C<\>@f, C<\>$g.
```
guest20, line 1420: `\( ... )` is a list of references to each of the items in the parens. `\( @things )` is not a reference to `@things`, nor a reference to an anon-array with the items from `@things`:

```
% perl -MData::Dumper -E '
  say Dumper \( @ARGV ) 
'  -- two items
$VAR1 = \'two';
$VAR2 = \'items';
```

`\( %things )` does the same with the key/value pairs from the hash:

```
 % perl  -MData::Dumper -E '
  my %arg = @ARGV; 
  say Dumper \( %arg ) 
' -- one pair
$VAR1 = \'one';
$VAR2 = \'pair';
```

`Dumper \(%a, @b, $c)` is not a good example of how to avoid list flattening.

dist/Data-Dumper/Dumper.pm
haarg: It's definitely not a change we will be making. That sentence should be removed.

jidanni: Oh yes please remove it or whatever and also tweak my version if you want I'm very welcome.

I guess I'll learn what a version bump is a little later. But just go right ahead.

jkeenan:
There has been no further discussion in this ticket in five months.  Closing.

Issue #21822, opened 2024-01-11
BBC: Blead Breaks Variable::Magic


cjg-cguevara:
This is a bug report for perl from "Carlos Guevara" <carlos@carlosguevara.com>,
generated with the help of perlbug 1.43 running under perl 5.39.6.

-----------------------------------------------------------------

BBC: Blead Breaks Variable::Magic

Please see http://fast-matrix.cpantesters.org/?dist=Variable::Magic

---
**Flags**
- category=core
- severity=low
---
**Perl configuration**
```
Site configuration information for perl 5.39.6:

Configured by cpan at Thu Jan 11 13:07:43 EST 2024.

Summary of my perl5 (revision 5 version 39 subversion 6) configuration:
  Commit id: 142fad4314e422f85799a7991e48b413c3dd9840
  Platform:
    osname=linux
    osvers=6.6.10-0-lts
    archname=x86_64-linux-thread-multi
    uname='linux cjg-alpine3 6.6.10-0-lts #1-alpine smp preempt_dynamic mon, 08 jan 2024 10:08:57 +0000 x86_64 linux '
    config_args='-des -Dprefix=/home/cpan/bin/perl -Dscriptdir=/home/cpan/bin/perl/bin -Dusedevel -Duse64bitall -Duseithreads'
    hint=recommended
    useposix=true
    d_sigaction=define
    useithreads=define
    usemultiplicity=define
    use64bitint=define
    use64bitall=define
    uselongdouble=undef
    usemymalloc=n
    default_inc_excludes_dot=define
  Compiler:
    cc='cc'
    ccflags ='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64'
    optimize='-O2'
    cppflags='-D_REENTRANT -D_GNU_SOURCE -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong'
    ccversion=''
    gccversion='13.2.1 20231014'
    gccosandvers=''
    intsize=4
    longsize=8
    ptrsize=8
    doublesize=8
    byteorder=12345678
    doublekind=3
    d_longlong=define
    longlongsize=8
    d_longdbl=define
    longdblsize=16
    longdblkind=3
    ivtype='long'
    ivsize=8
    nvtype='double'
    nvsize=8
[...]

PR #22041, opened 2024-02-27
Revert "OP_PADSV_STORE: only in void context"

iabyn:
This is merely an optimisation, which is breaking some CPAN modules due to expectations about particular optree structures.

Revert for now and worry about it again after 5.40 is released.

See https://github.com/Perl/perl5/issues/21822


khwilliamson (approved):

jkeenan:
> This is merely an optimisation, which is breaking some CPAN modules due to expectations about particular optree structures.
> 
> Revert for now and worry about it again after 5.40 is released.
> 
> See #21822

I applaud the idea of backing this commit out so as to avoid some BBC problems in the lead-up to 5.40.0.  But I don't know enough about the code format in `ext/B/t/optree_varinit.t` to evaluate it myself.

Leont (approved):

PR #22042, opened 2024-02-27
regen/regen_lib.pl: Stop compile time warning

khwilliamson:
This suppresses the warning about negative repeat counts


jkeenan:
> This suppresses the warning about negative repeat counts

Can you provide an example of how you encountered this warning?

khwilliamson:
PERL5OPT=-w make regen

jkeenan (approved):
Did before/after and, yes, before I got the warnings, after I didn't.

PR #22045, opened 2024-02-27
Lock ENV mutex while changing the environment

khwilliamson:
The comments here read:

/* only the parent thread can clobber the process environment, so no need
 * to use a mutex */

But this isn't true.  What is true about it is that only one thread can change the environment.  But the POSIX Standard explicitly states (in section 2.9.1):

    "Since multi-threaded applications are not allowed to use the
     environ variable to access or modify any environment variable while
     any other thread is concurrently modifying any environment
     variable, any function dependent on any environment variable is not
     thread-safe if another thread is modifying the environment"

In other words, the results are undefined if a thread is trying to read the environment while another thread is modifying it.

Thus we need a write lock when performing these operations.  Threads merely reading the environment can use a read lock.  Some code already uses one; future commits will fix others to do so.

Since I'm unaware of any tickets involving this, I infer that it is rare for a process to change the environment once initialization is done, and/or we have heisenbugs which turn out to have been caused by this.

### DIFF
--- a/util.c
+++ b/util.c
@@ -2398,30 +2398,36 @@ void
 Perl_my_setenv(pTHX_ const char *nam, const char *val)
 {
 #  if defined(USE_ITHREADS) && !defined(WIN32)
-    /* only parent thread can modify process environment, so no need to use a
-     * mutex */
+    /* only parent thread can modify process environment */
     if (PL_curinterp != aTHX)
         return;
 #  endif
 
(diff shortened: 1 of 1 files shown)


tonycoz:
Calls to getenv() can cause the problems you describe, but read access through `%ENV` doesn't actually read the environment, it is roughly a non-magical hash read. There's some special casing for %ENV in `Perl_hv_common`, but the process environment isn't accessed on read.

tonycoz (approved):

PR #22050, opened 2024-02-28
Update perldata.pod but not sure I'm really adding any value

jidanni:
So feel free to edit of just plain reject.


ilmari:
I don't see how these examples add any value over the one that's already there. Closing.

PR #22052, opened 2024-02-28
Add CI entry for static Perl

Leont:


tonycoz:
The change itself looks fine, if it was inspired by some sort of issue, like the `re` build/test issues, this could use a mention in the commit message.

Leont:
> The change itself looks fine, if it was inspired by some sort of issue, like the re build/test issues, this could use a mention in the commit message.

No it was the other way around, the re build/test issues were discovered when I wrote this patch.

tonycoz:
> No it was the other way around, the re build/test issues were discovered when I wrote this patch.

Ok, I though it might have been inspired by #21550

tonycoz (approved):

PR #22057, opened 2024-02-29
Permit subsequent `use VERSION` of same version

leonerd:
The recently-added warning about subsequent `use VERSION` statements even triggered on duplicate requests of the same version. It turns out there shouldn't be problems with doing that, so we can relax the message a little to not warn in that particular case.

### DIFF
--- a/pod/perldeprecation.pod
+++ b/pod/perldeprecation.pod
@@ -62,13 +62,14 @@ Category: "deprecated::unicode_property_name"
 
 =head2 Perl 5.46
 
-=head3 C<use VERSION> while another C<use VERSION> is in scope
+=head3 Changing C<use VERSION> while another C<use VERSION> is in scope
 
 A C<use VERSION> declaration has many implicit effects on the surrounding
 scope, such as L<strict> and L<feature> flags, or importing L<builtin>
-functions. Once you have a C<use VERSION> statement in scope, any other
-C<use VERSION> statement is now deprecated since Perl 5.39.8, due to the
-increasing complexity of swapping from one prevailing version to another.
+functions. Once you have a C<use VERSION> statement in scope, another
+C<use VERSION> statement with a different version is now deprecated since
+Perl 5.39.8, due to the increasing complexity of swapping from one
+prevailing version to another.
 
 Category: "deprecated::subsequent_use_version"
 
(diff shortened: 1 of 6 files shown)


jkeenan:
Doesn't this section of `pod/perldeprecation.pod` have to change as well?
```
 63 =head2 Perl 5.46
 64 
 65 =head3 C<use VERSION> while another C<use VERSION> is in scope
 66 
 67 A C<use VERSION> declaration has many implicit effects on the surrounding
 68 scope, such as L<strict> and L<feature> flags, or importing L<builtin>
 69 functions. Once you have a C<use VERSION> statement in scope, any other
 70 C<use VERSION> statement is now deprecated since Perl 5.39.8, due to the
 71 increasing complexity of swapping from one prevailing version to another.
 72 
 73 Category: "deprecated::subsequent_use_version"
 74 
```
Also, there may be CPAN authors/maintainers who have already changed some of their code in anticipation of the current version of the warning being schedule to appear in 5.40.  If we apply this p.r., then those people will have to change their code once again.  I think the current p.r. is okay, but it may be an irritant to people who have been diligent enough to update their distros.  So we may have to acknowledge that we may be inconveniencing some people.

leonerd:
> Doesn't this section of `pod/perldeprecation.pod` have to change as well?

Ahyes, indeed.

> Also, there may be CPAN authors/maintainers who have already changed some of their code in anticipation of the current version of the warning being schedule to appear in 5.40. If we apply this p.r., then those people will have to change their code once again

This change only makes the warning happen *less* often. If people have already changed code then this PR won't require further change. If they haven't then this PR may cause fewer warnings and mean less other change needed overall.

haarg (approved):

pod/perldeprecation.pod
```diff
@@ -62,13 +62,14 @@ Category: "deprecated::unicode_property_name"
...
-=head3 C<use VERSION> while another C<use VERSION> is in scope
+=head3 Changing C<use VERSION> while another C<use VERSION> is in scope
 
 A C<use VERSION> declaration has many implicit effects on the surrounding
 scope, such as L<strict> and L<feature> flags, or importing L<builtin>
-functions. Once you have a C<use VERSION> statement in scope, any other
-C<use VERSION> statement is now deprecated since Perl 5.39.8, due to the
-increasing complexity of swapping from one prevailing version to another.
+functions. Once you have a C<use VERSION> statement in scope, another
+C<use VERSION> statement with a diffeerent version is now deprecated since
```
jkeenan: s/diffeerent/different/

PR #22059, opened 2024-03-01
Add is_inf and is_nan functions to builtin::

leonerd:
As it is not expected that these two functions are likely to be called very often, I didn't bother making inlined-opcode variants of them as per e.g. is_bool(). Instead these are just regular XSUBs like trim() is.

Testing-wise, I'm a little unsure about the use of `1 * (inf - inf)` to get a calculated NaN value, but I really couldn't think of anything else that actually works. Is it portable enough?

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -99,6 +99,20 @@ XS(XS_builtin_inf)
     XSRETURN_NV(NV_INF);
 }
 
+XS(XS_builtin_is_inf);
+XS(XS_builtin_is_inf)
+{
+    dXSARGS;
+    SV *val = ST(0);
+    if(items != 1)
+        croak_xs_usage(cv, "val");
+    SvGETMAGIC(val);
+    if(SvNOK(val) && Perl_isinf(SvNV(val)))
+        XSRETURN_YES;
+    else
+        XSRETURN_NO;
+}
+
 XS(XS_builtin_nan);
 XS(XS_builtin_nan)
 {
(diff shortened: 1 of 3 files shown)


builtin.c
```diff
@@ -99,6 +99,20 @@ XS(XS_builtin_inf)
...
 
+XS(XS_builtin_is_inf);
+XS(XS_builtin_is_inf)
+{
+    dXSARGS;
+    SV *val = ST(0);
+    if(items != 1)
+        croak_xs_usage(cv, "val");
+    SvGETMAGIC(val);
+    if(SvNOK(val) && Perl_isinf(SvNV(val)))
```
tonycoz, line 110: The NOK check seems non-perlish to me, it also breaks overloads:
```
tony@venus:.../git/perl6$ ./perl -Ilib ../22059.pl 
Built-in function 'builtin::inf' is experimental at ../22059.pl line 6.
Built-in function 'builtin::is_inf' is experimental at ../22059.pl line 11.
Built-in function 'builtin::is_inf' is experimental at ../22059.pl line 13.

1

1
tony@venus:.../git/perl6$ cat ../22059.pl
use v5.38;
use builtin qw(is_inf inf);
use POSIX();

package Foo {
  use overload '0+' => sub { inf };
}

my $x = bless {}, "Foo";

say is_inf($x);
say POSIX::isinf($x);
say is_inf("Inf");
say POSIX::isinf("Inf");
```

tonycoz:
Another issue with this (and ceil(), floor() for that matter), is it doesn't support overloads in the same way other numeric functions do, e.g.:

```
use overload sin => sub { ... };
```

It's less of an issue if the `0+` overload returns a Inf/Nan as needed (which this code also doesn't support), but I can see it as an issue for ceil/floor with bignum-ish overloads.

jkeenan:
> Another issue with this (and ceil(), floor() for that matter), is it doesn't support overloads in the same way other numeric functions do, e.g.:
> 
> ```
> use overload sin => sub { ... };
> ```
> 
> It's less of an issue if the `0+` overload returns a Inf/Nan as needed (which this code also doesn't support), but I can see it as an issue for ceil/floor with bignum-ish overloads.

@leonerd, @tonycoz has noted some significant problems with this pull request.  Do you wish to pursue it further?  Thanks.

PR #22063, opened 2024-03-04
Remove the tombstone-related code from pad.h + pad.c

leonerd:
This was a short-lived experimental feature intended to implement removal of lexical symbols, in order to provide syntax like `no builtin ...` or to remove implied imported builtins on a change of prevailing `use VERSION`. The model of removing a lexical has proven to be too subtle and complex to implement as well as raising various awkward questions about the semantics, so we decided to remove it again in Perl 5.39.8. There is now no longer any code that uses `PADNAMEf_TOMBSTONE`.

As `PADNAMEf_TOMBSTONE` was only added in earlier in the 5.39.x developent series, there is no need to retain the constants in .h files for compatibility or to reserve the numbers for them. They have never appeared in a stable release of Perl.

### DIFF
--- a/Porting/deparse-skips.txt
+++ b/Porting/deparse-skips.txt
@@ -279,11 +279,6 @@ uni/variables.t
 ../lib/dumpvar.t
 ../lib/English.t
 
-# 18 Feb 24. This is probably failing due to the new PADNAMEf_TOMBSTONE
-# thing, which is likely to be removed again before 5.40. The test failure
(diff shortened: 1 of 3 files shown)


tonycoz:
`../lib/builtin.t` can be removed from `Porting/deparse-skips.txt`

leonerd:
> `../lib/builtin.t` can be removed from `Porting/deparse-skips.txt`

Ahyes. Now removed and re-pushed.

tonycoz (approved):

PR #22069, opened 2024-03-07
Exclude split STRING from default pattern modifier /x

### LINKED ISSUE #18032: use re '/aansx'; breaks split " ", ...

demerphq:
regcomp.c - Exclude split STRING from /x default modifiers.
    
We have a magic special case with split STRING where if the STRING
is a single space character we trigger awk emaulation mode, which
causes split to ignore leading whitespace, and to split on \s+.
    
When default regex modifier support was added to perl we didn't notice
that it doesn't make sense for the /x modifiers to change the behavior
of a STRING pattern in split. STRING patterns are NOT parsed using the
normal pattern parsing rules, for instance "\s+" would split on 's'
characters, not on whitepace. The /x modifier is designed to change the
rules for *pattern* parsing, most notably by adding support for comments
and ignoring whitespace in pattern definitions. This feels to me like a
feature that should only apply to cases where a true pattern is parsed,
such as where \s+ would be treated as the special pattern for matching
whitespace and not a literal 's' character. Historically there was NO way
to create a string pattern with the /x pattern enabled, to enable /x the
string MUST have been fed into a true pattern parsing construct like
qr// m// or s///.
    
By excluding STRING patterns from /x notation, we ensure that the
special case of awk emulation for C<split " "> is preserved.
    
Note that a default /x modifier *does* change the behavior of
C<split / />, as the pattern in that case is parsed the same as
C<split qr/ /> or C<m/ />.
    
    
Fixes #18032

Note this is a replacement for https://github.com/Perl/perl5/pull/22035


t/op/split.t
```diff
@@ -7,7 +7,7 @@ BEGIN {
     require './charset_tools.pl';
 }
 
-plan tests => 197;
+plan tests => 213;
```
mauke: Should be 219.

t/op/split.t
```diff
@@ -725,3 +725,37 @@ SKIP: {
 		}, "special-case pattern for $prog");
 	}
 }
+
+# gh18032: check that `split " "` does not get converted to `split ""`
+SKIP: {
+    skip_if_miniperl("special-case patterns: need dynamic loading", 5);
```
mauke: The `5` should be `5*2 + 6*2`.

t/op/split.t
demerphq: thanks, fixed.

t/op/split.t
demerphq: thanks, fixed with dynamic computation along these lines.

Leont:
Do I understand correctly that this affects all string-based split and not just `" "`? (I think that's reasonable, but it wasn't clear from your description)

demerphq:
@Leont yes you understand correctly. I should update the commit message. NOTE: Done now.

demerphq:
FWIW, it turns out this patch tickles some interesting issues under static builds (-Uusedl), I am digging into it, but there seems to be a rather large amount of horribleness involved so i don't know exactly what i will have to fix.

richardleach:
Does this change also affect non-perl engines, like the comment & block at https://github.com/Perl/perl5/blob/12d9d972e1e6d0f79c915e80e43b1fe2619c2538/regcomp.c#L2209 do?

demerphq:
@richardleach short answer no.

What that comment is saying is that other regex engines can get the special awk emulation behavior by setting SKIPWHITE in the flags for the pattern. Historically, notably before other engines were supported, we would inspect the pattern string (much like your patch did) OUTSIDE the regex engine to trigger the special mode. When I introduced the regex engine plugin framework i changed it so that it was the regex engines that would set the flags, not the split code, and that the flags would be the controlling factor not the spelling of the pattern.

So, that means that if an alternative regex engine wants to set SKIPWHITE on a pattern they compile Perl will continue to respect that decision. However it would be up to the regex engine to emulate this behavior of having RXf_SPLIT be mutexed with RXf_PMf_EXTENDED. We can probably fix that with a bit of research. The way the default modifers are merged into the flags is a bit convoluted so i punted and went with a fairly late mutex, but we can probably do it much earlier.

demerphq:
NOTE, I had to add some extra code to ensure that the test would not die under -Uusedl. Compiling the string "\N{SPACE}" under -Uusedl was triggering broken regex engine output. I havent added a specific test for this as the new tests in t/op/split.t test it as part of the tests for skipwhite. Perhaps we should have a specific test for it. I am actually a bit unclear on what we expect. The patterns compiled to process "\N{SPACE}" are "internal" regexes, and it is not clear if we should produce regex debugging output for them or not.  I might revisit this general subject in more detail later on.

richardleach:
Cheers for the explanation, @demerphq.

demerphq:
Anyone object to me merging this?

Leont (approved):

druud:
Maybe this is useful for test cases:
```
perl -Mstrict -wE'
  say "<", join("|", split " ", " a b c   "), ">";
  say "<", join("|", split " ", " a b c   ", -1), ">";
  { use re "/x";
    say "<", join("|", split " ", " a b c   "), ">";
    say "<", join("|", split " ", " a b c   ", -1), ">";
  }
'
<a|b|c>
<a|b|c|>
< |a| |b| |c| | | >
< |a| |b| |c| | | |>
```
(perl v5.38.2)

PR #22071, opened 2024-03-08
Fcntl: export File Sealing flags on request

richardleach:
On Linux, anonymous files can be created with memfd_create(2). File sealing provides a mechanism for a process to ensure that the shared memory cannot be unexpectedly tampered with.

This commit adds the flags for this mechanism to the Fcntl module, from where they can be optionally exported.

See: https://man7.org/linux/man-pages/man2/memfd_create.2.html


jkeenan:
> On Linux, anonymous files can be created with memfd_create(2). File sealing provides a mechanism for a process to ensure that the shared memory cannot be unexpectedly tampered with.
> 

I have to admit that I have a negative reaction whenever I see a pull request whose commit message reads like, *On **Linux** we can do blah, blah, blah; therefore we should modify core like so.*  We're not supporting a Linux application; we're supporting a language on dozens of operating systems.

I'm particularly put off by *this* p.r.'s commit message, because in less than 60 seconds I was able to confirm that `memfd_create` appears on FreeBSD and NetBSD.  (I'm not sure yet about OpenBSD.)

So I think should investigate other platforms, revise the commit message to reflect what we learn, then make the revised p.r. from a `smoke-me/` branch so that our non-Linux smoke testing rigs give meaningful results.  Thanks.

richardleach:
> On **Linux**

I used https://github.com/Perl/perl5/pull/21695, since it had already been merged, as a template for both the change and the commit message. This seemed like a reasonable thing to do at the time, but I apologise for any offence caused and am happy to reword the commit message.

I'm not sure what a `smoke-me` branch would tell us, since Fnctl wraps all flag definitions in `#ifdef` directives, so the module will get the flag if the platform it is built on supports it, and won't if it doesn't.

jkeenan:
> 
> I'm not sure what a `smoke-me` branch would tell us, since Fnctl wraps all flag definitions in `#ifdef` directives, so the module will get the flag if the platform it is built on supports it, and won't if it doesn't.

Precisely.  A `smoke-me` branch would tell us whether we had done no harm (or not).

nicomen:
…

> I'm not sure what a `smoke-me` branch would tell us, since Fnctl wraps all flag definitions in `#ifdef` directives, so the module will get the flag if the platform it is built on supports it, and won't if it doesn't.

Here's some info on the intention behind and concept of a smoke-me branch if that's what you were wondering about. https://perldoc.perl.org/perlgit#Using-a-smoke-me-branch-to-test-changes

richardleach:
@nicomen - I am familiar with `smoke-me` branches, but thanks for the link nonetheless.

@jkeenan - I've pushed `smoke-me/hydahy/fcntl_seals`

richardleach:
I've also updated the commit message within this PR.

rwp0:
> I've also updated the commit message within this PR.

Thanks, could you please update the PR description as well?

richardleach:
> Thanks, could you please also update the PR description as well?

For what benefit? It would make the subsequent flow of comments more difficult to understand.

tonycoz (approved):

PR #22074, opened 2024-03-12
Note that __VA_OPT__ is now in C++20

khwilliamson:
whereas before we said it had never been standardized


tonycoz:
It's also in [C23](https://open-std.org/JTC1/SC22/WG14/www/docs/n3220.pdf)

tonycoz (approved):

PR #22081, opened 2024-03-15
Slightly optimize IV range check in S_sv_2iuv_common()

t-a-k:
The code was like `iv != IV_MIN && CONST > abs(iv)`, but more straightforward equivalent `-CONST < iv && iv < CONST` does not require special treatment of IV_MIN and will be compiled to a bit better code (lesser total number of comparisons).

I believe this is a performance optimization without any behavioral change.

### DIFF
--- a/sv.c
+++ b/sv.c
@@ -2066,6 +2066,15 @@ S_sv_setnv(pTHX_ SV* sv, int numtype)
     }
 }
 
+#ifndef NV_PRESERVES_UV
+#  define MAX_UV_PRESERVED_IN_NV (((UV)1 << NV_PRESERVES_UV_BITS) - 1)
+#  define MAX_IV_PRESERVED_IN_NV ((IV)MAX_UV_PRESERVED_IN_NV)
+#  define MIN_IV_PRESERVED_IN_NV (-MAX_IV_PRESERVED_IN_NV)
+/* We presume that (IV)MAX_UV_PRESERVED_IN_NV and (-MAX_IV_PRESERVED_IN_NV)
+   above will not overflow if the condition below holds true:  */
+STATIC_ASSERT_DECL(MAX_UV_PRESERVED_IN_NV <= (UV)IV_MAX);
+#endif
+
 STATIC bool
 S_sv_2iuv_common(pTHX_ SV *const sv)
 {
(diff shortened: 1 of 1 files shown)


sv.c
```diff
@@ -2098,9 +2098,10 @@ S_sv_2iuv_common(pTHX_ SV *const sv)
             SvIV_set(sv, I_V(SvNVX(sv)));
             if (SvNVX(sv) == (NV) SvIVX(sv)
 #ifndef NV_PRESERVES_UV
-                && SvIVX(sv) != IV_MIN /* avoid negating IV_MIN below */
-                && (((UV)1 << NV_PRESERVES_UV_BITS) >
-                    (UV)(SvIVX(sv) > 0 ? SvIVX(sv) : -SvIVX(sv)))
+                /* Optimizing compilers might merge two comparisons below
+                   into single comparison */
+                && -(IV)((UV)1 << NV_PRESERVES_UV_BITS) < SvIVX(sv)
```
mauke: If `(IV)(1 << NV_PRESERVES_UV_BITS)` ends up being `IV_MIN`, this code will try to negate `IV_MIN` again. Actually, the real problem is one step before that: If `(UV)1 << NV_PRESERVES_UV_BITS` exceeds the range of `IV`, the result of the `IV` cast is implementation defined (or might raise a signal).
I believe this would happen if there is a platform where `NV_PRESERVES_UV_BITS` is equal to `IVSIZE * CHAR_BIT - 1`.

t-a-k: > If `(UV)1 << NV_PRESERVES_UV_BITS` exceeds the range of `IV`,
[...]
> I believe this would happen if there is a platform where `NV_PRESERVES_UV_BITS` is equal to `IVSIZE * CHAR_BIT - 1`.

I think these would practically never happen (there seems to be no combination of such floating-point format and IV/UV format on platforms supported by perl), and even if that is the case, [static assertion in `sv.c`](https://github.com/Perl/perl5/blob/0e9cc8e5aaebeef20004deb2c6fdf911e60ba482/sv.c#L2292-L2293) will fail anyway.

tonycoz: Maybe:
```
        && (IV)(((UV)1 << NV_PRESERVES_UV_BITS)-1) <= SvIVX(sv)
        && SvIVX(sv) <= (IV)-(((UV)1 << NV_PRESERVES_UV_BITS)-1)
```
instead, which avoids negating `IV_MIN`.

Defining symbols for the limits makes this a lot more readable:
```
#define MAX_NV_PRESERVED (IV)(((UV)1 << NV_PRESERVES_UV_BITS)-1)
#define MIN_NV_PRESERVED (IV)-(((UV)1 << NV_PRESERVES_UV_BITS)-1)

...
        && MIN_NV_PRESERVED <= SvIVX(sv)
        && SvIVX(sv) <= MAX_NV_PRESERVED
```
See https://godbolt.org/z/ThGbvYav4 for some generated code comparisons.

t-a-k: That makes sense.  I will modify this PR.

t-a-k:
I force-pushed 1a23fdc2b84fa8c2bfe4703f30475f0acc9e56bf.

I adopted @tonycoz's proposal in https://github.com/Perl/perl5/pull/22081#discussion_r1529616381, but I renamed `(MAX|MIN)_NV_PRESERVED` to `(MAX|MIN)_IV_PRESERVED_IN_NV` to clarify that they are IV constants.

tonycoz (approved):

Issue #21784, opened 2023-12-31
BBC: Blead breaks MLEHMANN/Coro-6.57.tar.gz


eserte:
If EV or Event are installed, then the test suite fails since perl 5.39.3 like this (in the Event subdirectory):
```
...
t/01_unblock.t .. 1/12 Attempt to free unreferenced scalar: SV 0x1102708 at /home/cpansand/.cpan/build/2023123113/Coro-6.57-0/Event/../blib/lib/Coro.pm line 442.
t/01_unblock.t .. Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 2/12 subtests 
...
```
or like this (in the EV subdirectory):
```
t/01_unblock.t .. 1/12 Attempt to free unreferenced scalar: SV 0x1e45f30 at /home/cpansand/.cpan/build/2023123113/Coro-6.57-0/EV/../blib/lib/Coro.pm line 442.
t/01_unblock.t .. Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 2/12 subtests 
```

CPAN Testers <http://matrix.cpantesters.org/?dist=Coro+6.57> show also pass reports. In these cases probably neither Event nor EV were installed.

jkeenan:
@eserte, it's not clear to me how you got these results.  On Linux, I twice attempted to bisect this problem, using these invocations:
```
 2095  perl Porting/bisect.pl --start=v5.39.2 --end=v5.39.3 --with-module=Event --module AnyEvent
 2097  perl Porting/bisect.pl --start=v5.39.2 --end=v5.39.3 --with-module=EV --module AnyEvent
```
In both cases, no failure was reported:
```
Runner returned 0 for end revision at Porting/bisect.pl line 233.
```
Could you clarify?  Thanks.

jkeenan:
> @eserte, it's not clear to me how you got these results. On Linux, I twice attempted to bisect this problem, using these invocations:
> 
> ```
>  2095  perl Porting/bisect.pl --start=v5.39.2 --end=v5.39.3 --with-module=Event --module AnyEvent
>  2097  perl Porting/bisect.pl --start=v5.39.2 --end=v5.39.3 --with-module=EV --module AnyEvent
> ```
> 
> In both cases, no failure was reported:
> 
> ```
> Runner returned 0 for end revision at Porting/bisect.pl line 233.
> ```
> 
> Could you clarify? Thanks.

@eserte, could you clarify how you got these BBCs?  Thanks.

[...]

PR #22082, opened 2024-03-17
Perl_tied_method(): *conditionally* use RC stack

iabyn:
Fixes GH #21784.

Perl_tied_method() is used to call tied methods for filehandles.  One of its actions is to push a new stackinfo (and hence a new argument stack) to run the method under. When the  method returns, any returned values are copied back to to original stack and the stackinfo popped.  But stackinfos (and the AV pointed to by si_stack) aren't freed at that point; they are kept around until global destruction, in case they're needed again.

Before this commit, Perl_tied_method() was incorrectly *always* making the new stack AvREAL: this should in fact be done only on PERL_RC_STACK builds.

The effect of this was that, on non-PERL_RC_STACK builds, any temp SVs created as the return value(s) from the method call, would be left on the abandoned stack, while the SVs themselves have long since been freed. Normally this doesn't matter much - the array will be finally emptied only during global destruction, which doesn't care if an already-freed SV is attempted to be freed again.

But this was causing Coro to fail, because it maintains a stackinfo chain per green thread, and manages and frees them itself, so they can end up being freed before global destruction, triggering an 'Attempt to free unreferenced scalar' warning.

For completeness, this commit now also resets PL_stack_sp to PL_stack_base before abandoning the stack, so that that on PERL_RC_STACK builds, there's no chance of the copied return value SVs being double-counted later on,

### DIFF
--- a/pp_sys.c
+++ b/pp_sys.c
@@ -704,7 +704,11 @@ Perl_tied_method(pTHX_ SV *methname, SV **mark, SV *const sv,
      * our caller wasn't, then we'll need to adjust when copying the args
      * and results between the two stacks.
      */
+#ifdef PERL_RC_STACK
     push_stackinfo(PERLSI_MAGIC, 1);
+#else
+    push_stackinfo(PERLSI_MAGIC, 0);
+#endif
 
     /* extend for object + args. If argc might wrap/truncate when cast
      * to SSize_t and incremented, set to -1, which will trigger a panic in
@@ -761,6 +765,10 @@ Perl_tied_method(pTHX_ SV *methname, SV **mark, SV *const sv,
     }
     ret_args = call_sv(methname, (flags & G_WANT)|G_METHOD_NAMED);
     SV **orig_sp = PL_stack_sp;
+    /* Ensure that the old stack is marked as empty before we abandon it.
+     * Otherwise, if it's AvREAL(), items could be double-freed some time
+     * after we copy the return SVs to the caller's stack. */
+    PL_stack_sp = PL_stack_base;
     pop_stackinfo();
     /* pop and free the spare SV (the 'X' in the comments above */
     if (flags & TIED_METHOD_ARGUMENTS_ON_STACK) {


jkeenan:
@iabyn I'm glad that you were able to look into this problem, but it seems as if the branch is failing some of our ASAN-related CI tests.  Can you investigate? Thanks.

mauke:
The ASAN stuff is failing on blead, too (helloworld.c is crashing). It's unrelated to this patch.

jkeenan:
> The ASAN stuff is failing on blead, too (helloworld.c is crashing). It's unrelated to this patch.

@mauke, thanks.  On my own machine I'm getting segfaults even during `Configure` when I try those `ASAN`-related config options.  So I can't bisect.  Have you been able to identify where those crashes began?

iabyn:
On Sun, Mar 17, 2024 at 12:44:04PM +0000, James E Keenan wrote:
> > The ASAN stuff is failing on blead, too (helloworld.c is crashing). It's unrelated to this patch.
> 
> @mauke, thanks.  On my own machine I'm getting segfaults even during
> `Configure` when I try those `ASAN`-related config options.  So I can't
> bisect.  Have you been able to identify where those crashes began?

Do you get a SEGV if you try the same simple compilation that Configure is
trying to do? I.e. create this file and try to compile and run it:

    $ cat ./try.c

    #include <stdio.h>
    int main() { printf("Ok\n"); return(0); }

    $ clang -fsanitize=address -o try -O2 -g -DDEBUGGING -fno-strict-aliasing -fstack-protector-strong -I/usr/local/include -fstack-protector-strong -L/usr/local/lib try.c -lpthread -lgdbm -ldb -ldl -lm -lcrypt -lutil -lc

    $ ./try
    Ok

    $ 

That works ok for me on Linux with a somewhat outdated version of clang:

    $ clang --version
    clang version 13.0.1 (Fedora 13.0.1-1.fc35)



-- 
It's not that I'm afraid to die, I just don't want to be there when it
happens.
    -- Woody Allen

jkeenan:
> clang -fsanitize=address -o try -O2 -g -DDEBUGGING -fno-strict-aliasing -fstack-protector-strong -I/usr/local/include -fstack-protector-strong -L/usr/local/lib try.c -lpthread -lgdbm -ldb -ldl -lm -lcrypt -lutil -lc

```
$ uname -mrs
Linux 6.5.0-25-generic x86_64

$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ clang -fsanitize=address -o try -O2 -g -DDEBUGGING -fno-strict-aliasing -fstack-protector-strong -I/usr/local/include -fstack-protector-strong -L/usr/local/lib gh-22082-try.c

$ ./try
Ok
```

jkeenan:
> > clang -fsanitize=address -o try -O2 -g -DDEBUGGING -fno-strict-aliasing -fstack-protector-strong -I/usr/local/include -fstack-protector-strong -L/usr/local/lib try.c -lpthread -lgdbm -ldb -ldl -lm -lcrypt -lutil -lc
> 
> ```
> $ uname -mrs
> Linux 6.5.0-25-generic x86_64
> 
> $ clang --version
> Ubuntu clang version 14.0.0-1ubuntu1.1
> Target: x86_64-pc-linux-gnu
> Thread model: posix
> InstalledDir: /usr/bin
> 
> $ clang -fsanitize=address -o try -O2 -g -DDEBUGGING -fno-strict-aliasing -fstack-protector-strong -I/usr/local/include -fstack-protector-strong -L/usr/local/lib gh-22082-try.c
> 
> $ ./try
> Ok
> ```

In contrast, here's what I get with those config args in `./Configure`:
```
$ cat 0e9cc8e5aa.clang.sanitize-equals-address.configure.output.txt 
First let's make sure your kit is complete.  Checking...
Would you like to see the instructions? [n]  
Locating common programs...
Checking compatibility between /usr/bin/echo and builtin echo (if any)...
Symbolic links are supported.
Checking how to test for symbolic links...
You can test for symbolic links with 'test -h'.
Checking for cross-compile
No targethost for running compiler tests against defined, running locally
Good, your tr supports [:lower:] and [:upper:] to convert case.
Using [:upper:] and [:lower:] to convert case.
I see a config.sh file.  Shall I use it to set the defaults? [y]  
Fetching default answers from your old config.sh file...
Operating system name? [linux]  
Operating system version? [6.5.0-25-generic]  
Installation prefix to use? (~name ok) [/usr/local]  
AFS does not seem to be running...
What installation prefix should I use for installing files? (~name ok)
[/usr/local]  
Build a threading Perl? [n]  
Build Perl for multiplicity? [n]  
Use which C compiler? [clang -fsanitize=address]  
Checking for GNU cc in disguise and/or its version number...
Now, how can we feed standard input to your C preprocessor...
Directories to use for library searches?
[/usr/local/lib /usr/lib/x86_64-linux-gnu /usr/lib /usr/lib64]  
What is the file extension used for shared libraries? [so]  
Make shared library basenames unique? [n]  
Build Perl for SOCKS? [n]  
Try to use long doubles if available? [n]  
Checking for optional libraries...
What libraries to use? [-lpthread -ldl -lm -lcrypt -lutil -lc]  
What optimizer/debugger flag should be used? [-O2]  
Any additional cc flags?
[-fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -DPURIFY]  
Let me guess what the preprocessor flags are...
Any additional ld flags (NOT including libraries)?
[ -fstack-protector-strong -L/usr/local/lib]  
Checking your choice of C compiler and flags for coherency...
I've tried to compile and run the following simple program:

#include <stdio.h>
int main() { printf("Ok\n"); return(0); }

I used the command:

	clang -fsanitize=address -o try -O2 -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -DPURIFY -fstack-protector-strong -L/usr/local/lib try.c -lpthread -ldl -lm -lcrypt -lutil -lc
	 ./try

and I got the following output:

Segmentation fault (core dumped)
The program compiled OK, but exited with status 139.
You have a problem.  Shall I abort Configure [y]  
Ok.  Stopping Configure.
```

tonycoz:
If you try:
```
sudo sysctl -w vm.mmap_rnd_bits=28
```
on your machine does it fix the ASAN segv?

There's a reported problem It's not a perl issue, but a problem with github/ubuntu/ASAN, see https://github.com/actions/runner-images/issues/9491

Thanks to @ilmari for finding that link

jkeenan:
> If you try:
> 
> ```
> sudo sysctl -w vm.mmap_rnd_bits=28
> ```
> 
> on your machine does it fix the ASAN segv?
> 

Yes, it enables `./Configure` to complete; now running `make test_prep`.

Is that something I should set permanently -- or just when I want to do an ASAN build?

tonycoz:
> Is that something I should set permanently -- or just when I want to do an ASAN build?

It's only needed for ASAN.

You might also be able to fix it by using an updated clang from apt.llvm.org.

Workaround for this ASAN issue in #22084

tonycoz (approved):

PR #22083, opened 2024-03-17
OP_METHSTART, OP_INITFIELD: use shared mem for aux

iabyn:
Allocate the aux structure for these two ops using PerlMemShared_malloc() rather then Newx().

OPs (and auxiliary structures) are shared between threads, and there's no guarantee that an OP will be freed by the same thread as the one which created it. Some OSes (Windows IIRC) get upset if memory is malloc()ed and free()d by different threads, since each thread has a separate malloc() pool.

Hence we usually use PerlMemShared_malloc()/_free() to allocate such shared structs. These two recently-added ops were just using Newx().

This commit fixes them.

No tests failures as far as I'm aware - this is more a theoretical thing where one thread compiles some code, creates a child, then exits. The child lives on, and when the child finally exits, the aux struct is freed to the wrong pool.

### DIFF
--- a/class.c
+++ b/class.c
@@ -771,7 +771,8 @@ Perl_class_seal_stash(pTHX_ HV *stash)
             }
 
             UNOP_AUX_item *aux;
-            Newx(aux, 2, UNOP_AUX_item);
+            aux = (UNOP_AUX_item*)PerlMemShared_malloc(
+                                    sizeof(UNOP_AUX_item) * 2);
 
             aux[0].uv = fieldix;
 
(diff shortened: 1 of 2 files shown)


class.c
```diff
@@ -771,7 +771,8 @@ Perl_class_seal_stash(pTHX_ HV *stash)
             }
 
             UNOP_AUX_item *aux;
-            Newx(aux, 2, UNOP_AUX_item);
+            aux = (UNOP_AUX_item*)PerlMemShared_malloc(
```
leonerd, line 774: Most minor of nitpicks: Style-wise I tend to put a space between`UNOP_AUX_item` and `*`

leonerd (approved):
LGTM give or take a whitespace.
LMK if you don't get time to do this before Wed. morning and I can fix+merge instead. Would be good to get this in for 5.39.9

iabyn:
Nitpick fixed, and merged.

PR #22084, opened 2024-03-17
testsuite.yml: fix ASAN build tests

tonycoz:
Apparently a recent Ubuntu update changed the ASLR entropy to use more bits which is incompatible with ASAN.

This adjusts the entropy bits to be compatible with ASAN.


tonycoz:
adjusted the commit subject to make it less opaque

khwilliamson:
I don't know enough to review this, but it works, so it should be merged sooner than later

PR #22086, opened 2024-03-18
Use C99 named initialisers in core structs that define MGVTBLs

leonerd:
Since most Magic vtables do not provide every function, using C99 named initialisers allows us to more easily specify which ones are present, so readers don't have to count commas. Additionally provides a layer of robustness in case more functions are added in future.

(No rush before 5.39.9 but I would like to test this out at least, sometime before too long)


regen/mg_vtable.pl
```diff
@@ -538,7 +525,7 @@ BEGIN
     print $vt "  { $funcs }$comma\n";
     print $vt <<"EOH" if $data->{cond};
 #else
-  { 0, 0, 0, 0, 0, 0, 0, 0 }$comma
+  { .svt_get = 0 }$comma
```
ilmari: Can this not be `{}` like the other all-zero `MGVTBL`s were changed to? Or should they also have at least one member initialised? And in that case, could they both be just `{0}`?

Leont:
I thought we couldn't use this in public headers because C++?

regen/mg_vtable.pl
ilmari: Or is the difference because `static` vs not?

leonerd: Oh, yes it can probably be the same. I had a few different attempts at appeasing the compiler.

leonerd:
> I thought we could use this in public headers because C++?

Ugh; it's `mg_vtable.h`, indeed. Does that still count? Even if we can't use it there, I'd still like to do the .c files.

regen/mg_vtable.pl
ilmari: Looks like MSVC [doesn't like](https://github.com/Perl/perl5/actions/runs/8333990114/job/22806953844?pr=22086#step:4:3002) `{}` even for `static` definitions, so I think going for `{0}` consistently is the way to go.

ilmari:
> > I thought we could use this in public headers because C++?
> 
> Ugh; it's `mg_vtable.h`, indeed. Does that still count? Even if we can't use it there, I'd still like to do the .c files.

It gets included in `perl.h`, so yes. But having it for the hand-written ones in `.c` files is still a big win.

tonycoz:
> > I thought we could use this in public headers because C++?
> 
> Ugh; it's `mg_vtable.h`, indeed. Does that still count? Even if we can't use it there, I'd still like to do the .c files.

Another option would be to generate a separate `mg_vtable.c` with the initializations.

The `DOINIT` code doesn't appear usable outside of the executable/libperl.so since the magic functions are marked hidden.

jkeenan:
@leonerd,  Given where we are in our annual production cycle, are you thinking of this for 5.40 or 5.42?

leonerd:
> @leonerd, Given where we are in our annual production cycle, are you thinking of this for 5.40 or 5.42?

Oh it doesn't need to be in 5.40 at all; it isn't doing any kind of user-benefit change. It's purely tidying up the code internally for our benefit. It can wait for 5.42, I just happen to have started writing it now.

ext/PerlIO-encoding/encoding.xs
```diff
@@ -55,7 +55,7 @@ typedef struct {
 
 #define NEEDS_LINES	1
 
-static const MGVTBL PerlIOEncode_tag = { 0, 0, 0, 0, 0, 0, 0, 0 };
+static const MGVTBL PerlIOEncode_tag = {};
```
ilmari: This still needs to be `{0}` [per MSVC](https://github.com/Perl/perl5/actions/runs/8340362405/job/22824615617?pr=22086#step:4:3002).

ilmari:
> The `DOINIT` code doesn't appear usable outside of the executable/libperl.so since the magic functions are marked hidden.

Looking more closely at things, `DOINIT` is only defined in `INTERN.h`, which is only included in `globals.c`, so even though `mg_vtable.h` is included in `perl.h`, that only causes the `extern` declarations to be visible outside `globals.c`, so we could change the generated code after all.

mg_vtable.h
```diff
@@ -105,42 +105,42 @@ enum {		/* pass one of these to get_vtbl */
...
+    [want_vtbl_arylen_p] = "arylen_p",
+    [want_vtbl_backref] = "backref",
+    [want_vtbl_checkcall] = "checkcall",
+    [want_vtbl_collxfrm] = "collxfrm",
+    [want_vtbl_dbline] = "dbline",
+    [want_vtbl_debugvar] = "debugvar",
+    [want_vtbl_defelem] = "defelem",
+    [want_vtbl_destruct] = "destruct",
+    [want_vtbl_env] = "env",
+    [want_vtbl_envelem] = "envelem",
```
tonycoz, line 118: Array designated initializers are not C++ compatible, and since we have people who build perl with a C++ compiler, it's excluded by the guide in `perlhacktips.pod`.

g++ supports it as an extension. but MSVC doesn't.

jkeenan: @leonerd, @tonycoz identified some problems with this p.r. back in March.  Do you expect to move forward with the p.r.?  (It has begun to acquire merge conflicts.)  Thanks.

PR #22090, opened 2024-03-21
prepare_export_lexical: save PL_comppad safely

### LINKED ISSUE #21981: panic: invalid pad in intro_my with perl -MO=... and use v5.39*
**Description**

With -DDEBUGGING:
```
$ ./perl -Ilib -MO=Concise -e 'use v5.39'
panic: invalid pad in intro_my: 0x55ef8a17ca20[0x55ef8a1945b0] at -e line 1
```

Also
```
$ ./perl -Ilib -MO=Deparse -e 'use v5.39'
panic: invalid pad in intro_my: 0x55cd471baa20[0x55cd471d2630] at -e line 1.
sub BEGIN {
    require v5.39;
    ()
}
```

Note this doesn't happen with anything `use v5.38` or less, only 5.39 and higher.

According to a bisect this appears to have been introduced here:
```
HEAD is now at 6d65c70ff0 Import from `builtin` version bundles should be idempotent on existing symbols
good - zero exit from ./perl -Ilib -MO=Concise -e use v5.39;
48d382667948f6a16c732c47791ca4be891a9bdb is the first bad commit
commit 48d382667948f6a16c732c47791ca4be891a9bdb
Author: Paul "LeoNerd" Evans <leonerd@leonerd.org.uk>
Date:   Fri Jan 19 15:00:35 2024 +0000

    Have `use VERSION` and `perl -E` import a corresponding `use builtin` bundle
```

tonycoz:
When the pad being saved and the pad for PL_compcv is the same, in some cases the actual exports would result in reallocating the AvARRAY() for the saved PL_comppad.

The LEAVE in finish_export_lexical() would restore the old PL_comppad (which is fine) and the pre-reallocation PL_curpad (which isn't fine).

This would later panic.

SAVECOMPPAD; restores PL_comppad on LEAVE and then restores PL_curpad from PL_comppad, preventing the desync between those values.

It's unclear to me why only the save_BEGINs; causes this, but the fix does fix a real problem and prevents the panics that I'm aware of here.

Fixes #21981

### DIFF
--- a/builtin.c
+++ b/builtin.c
@@ -48,8 +48,9 @@ Perl_prepare_export_lexical(pTHX)
     /* We need to have PL_comppad / PL_curpad set correctly for lexical importing */
     ENTER;
     SAVESPTR(PL_comppad_name); PL_comppad_name = PadlistNAMES(CvPADLIST(PL_compcv));
-    SAVESPTR(PL_comppad);      PL_comppad      = PadlistARRAY(CvPADLIST(PL_compcv))[1];
-    SAVESPTR(PL_curpad);       PL_curpad       = PadARRAY(PL_comppad);
+    SAVECOMPPAD();
+    PL_comppad      = PadlistARRAY(CvPADLIST(PL_compcv))[1];
+    PL_curpad       = PadARRAY(PL_comppad);
 }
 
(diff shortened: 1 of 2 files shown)


leonerd (approved):
LGTM

Looks like I may have to apply a similar fix in some of my CPAN XS modules then...

PR #22091, opened 2024-03-21
add CVf_XS_RCSTACK to dump output

iabyn:
This was missed when I added the flag


leonerd (approved):
LGTM

I wonder if we can include a trailing comma on these lines, now we're all C99. Would make the next addition a bit neater.

iabyn:
I eventually decided not to add a comma this time round, as I have another commit in progress which builds on this commit.  I'll look into doing for the next commit.

PR #22092, opened 2024-03-21
Keep using undocumented gcc locale feature

khwilliamson:
Since 5.39.7, this feature has been enabled experimentally by default. It is similar to querylocale(), found on *BSD systems.  That function returns the current locale, and was created because this ability is often required, and POSIX neglected to define anything for this purpose. I'm told that the next POSIX Standard update will include a function for this purpose.

I have been using the gcc feature successfully for years in my local work spaces; we started this experiment to see if it worked generally. And there have been zero reports of problems with it.

The experiment was designed to refuse to compile in 5.39.10, so as to make sure it didn't slip into 5.40 without it being explicitly decided to do so.  Now, the final steps in publishing 5.39.9 include bumping the release to 5.39.10, and those can't be done with the code as-is.

This commit declares the experiment a success, and enables this feature non-experimentally as the default, with the ability to turn it off by passing
    -Accflags=-DNO_USE_NL_LOCALE_NAME
to Configure.

### DIFF
--- a/locale.c
+++ b/locale.c
@@ -251,14 +251,13 @@
  *          querylocale() is called only while the locale mutex is locked, and
  *          the result is copied to a per-thread place before unlocking.
  *
- *      -Accflags=-DUSE_NL_LOCALE_NAME
- *          glibc has an undocumented equivalent function to querylocale().  It
- *          currently isn't used by default because it is undocumented.  But
- *          testing hasn't found any problems with it.  Using this Configure
- *          option enables it on systems that have it (with no effect on
- *          systems lacking it).  Enabling this removes the need for perl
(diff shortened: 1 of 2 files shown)


leonerd:
OK - this switches the sense of the default and the arugment itself, but it doesn't remove the version-conditional `#error` line:

https://github.com/perl/perl5/commit/6eb3b0c7c89d29c37927b53a5142f06f8bd1621e#diff-f7d64c37bffc437cd59ec53833a08ce5adec3e7b03436c7ed8cb11c789b26ac0R395

khwilliamson:
Oops; fixed

leonerd (approved):
LGTM

PR #22097, opened 2024-03-22
add CvEVAL_COMPILED() flag and fix closure bug.

iabyn:
[NB: I'm not expecting to merge this until after 5.40. Too much chance that something subtle changes in closure behaviour.]

EVAL CVs are treated a bit weirdly: their CvROOT() and CvSTART() fields don't get populated; instead the current values are stored in the PL_eval_root and PL_eval_start variables while they are being executed.

This caused a bug in closures and nested evals when an inner eval was repeated twice. The first inner eval accessed an outer lexical, which caused a fake cache entry to be added to the outer eval's pad. The second inner eval finds this cached entry, but incorrectly concludes that the outer eval is in fact an anon sub prototype and issues a 'variable is not available' warning. This is due to this simplistic definition in pad.c:

    #define CvCOMPILED(cv) CvROOT(cv)

This commit adds a new flag, CvEVAL_COMPILED(), to indicate a fully-compiled EVAL CV. This allows us to work around the limitation.

In an ideal world this would have been fixed instead by making EVAL CVs first-class citizens with CvROOT() etc, but plenty of stuff seems to assume otherwise. So I took the path of least resistance.

See https://www.perlmonks.org/?node_id=11158351

### DIFF
--- a/cv.h
+++ b/cv.h
@@ -140,6 +140,7 @@ See L<perlguts/Autoloading with XSUBs>.
                                    CVf_METHOD; now CVf_NOWARN_AMBIGUOUS */
 #define CVf_XS_RCSTACK  0x200000 /* the XS function understands a
                                     reference-counted stack */
+#define CVf_EVAL_COMPILED 0x400000 /* an eval CV is fully compiled */
 
 /* This symbol for optimised communication between toke.c and op.c: */
 #define CVf_BUILTIN_ATTRS	(CVf_NOWARN_AMBIGUOUS|CVf_LVALUE|CVf_ANONCONST)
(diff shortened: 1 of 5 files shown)


tonycoz (approved):

jkeenan:
For the purpose of getting smoke-testing data, today I checked out this pull request, rebased it on (post-perl-5.40.0) blead and pushed it to origin as branch `smoke-me/jkeenan/davem/eval-compiled-gh-22097-20240610`.  @tonycoz has approved the branch but, since @iabyn has mentioned the possibility of subtle changes in closure behavior, additional eyeballs would be helpful.

PR #22108, opened 2024-03-28
testsuite.yml: workaround cygwin base address conflict in 5.39.10

tonycoz:
This also adds a note to perldelta for the 5.39.10 release on how to avoid the problem.


jkeenan (approved):
This looks reasonable, and since the only way we can be sure is to get it into smoke-testing/CI, I say go for it.

PR #22114, opened 2024-04-01
functionise and simplify TRIE_BITMAP_SET_FOLDED

iabyn:
These two commits turn the private TRIE_BITMAP_SET_FOLDED() macro into a static function and then simplify it a bit, mainly in a quest to silence a Coverity warning.

### DIFF
--- a/regcomp_trie.c
+++ b/regcomp_trie.c
@@ -517,21 +517,28 @@ is the recommended Unicode-aware way of saying
            : ( state==1 ? special : 0 )                                 \
       )
 
-#define TRIE_BITMAP_SET_FOLDED(trie, uvc, folder)           \
-STMT_START {                                                \
-    TRIE_BITMAP_SET(trie, uvc);                             \
-    /* store the folded codepoint */                        \
-    if ( folder )                                           \
-        TRIE_BITMAP_SET(trie, folder[(U8) uvc ]);           \
-                                                            \
-    if ( !UTF ) {                                           \
-        /* store first byte of utf8 representation of */    \
-        /* variant codepoints */                            \
-        if (! UVCHR_IS_INVARIANT(uvc)) {                    \
-            TRIE_BITMAP_SET(trie, UTF8_TWO_BYTE_HI(uvc));   \
-        }                                                   \
-    }                                                       \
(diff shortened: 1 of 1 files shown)


regcomp_trie.c
```diff
@@ -517,21 +517,28 @@ is the recommended Unicode-aware way of saying
...
+ * if any */
+
+STATIC void
+S_trie_bitmap_set_folded(pTHX_ RExC_state_t *pRExC_state,
+    reg_trie_data *trie, U8 uvc, const U8 * folder)
+{
+    TRIE_BITMAP_SET(trie, uvc);
+    /* store the folded codepoint */
+    if ( folder )
+        TRIE_BITMAP_SET(trie, folder[(U8) uvc ]);
```
hvds, line 531: The `(U8)` cast is now redundant.

hvds (approved):
LGTM; it may be worth putting a short version of the second commit message as a comment in the code - my expectation is that Coverity will still complain, and that could save the next investigator some effort.

regcomp_trie.c
demerphq: IMO it shouldn't be called uvc when this is a function. I think the argument should be changed to cp or ch instead. "uvc" means "UV Codepoint", if its not a uv it should have a different name, like uch or something like that.

iabyn:
Added the suggested code comments and merged.

iabyn:
On Tue, Apr 02, 2024 at 03:03:18AM -0700, Yves Orton wrote:
> IMO it shouldn't be called uvc when this is a function.

Now fixed up with v5.39.9-28-ga73aa540c8.

-- 

khwilliamson:
Note that the added comment is wrong w.r.t. EBCDIC.   The first byte is neither C2 nor C3

iabyn:
On Tue, Apr 02, 2024 at 09:02:25AM -0700, Karl Williamson wrote:
> Note that the added comment is wrong w.r.t. EBCDIC.   The first byte is neither C2 nor C3

Sigh! Fixed up with v5.39.9-29-gbcd28b1883.

-- 
   -- Larry Wall

Issue #21877, opened 2024-01-23
regular expressions matching lines read from an in-memory scalar is extremely slow in cygwin/MSYS2 perls 


Old-Green-Man:
**Description**
With cygwin/MSYS2 perl when you do regular expression matching on lines read from an in-memory scalar and the regex matches something it takes orders of magnitude longer than matching the same lines read from a disk file.

**Steps to Reproduce**
Below is code to demonstrate the behavior.  An example file is attached to run with the code. In this file about 16% of the lines match.
```
#!/usr/bin/env perl

use warnings;
use strict;
use Time::HiRes qw( time );

my $file = shift @ARGV;
my ($fh, $time, $n);

open $fh, "<", $file;
$time = time;
$n = 0;
while(<$fh>) {
  /^ ?Query/;
  $n++;
}
printf "%f read lines from disk and do RE; n=$n.\n", time - $time;

seek $fh, 0, 0;
my $s = "";
while(<$fh>) {
  $s .= $_;
}
# my $s = do {
#   local $/;
#   <$fh>;
# };

open $fh, "<", \$s;
$time = time;
$n = 0;
while(<$fh>) {
  /^ ?Query/;
  $n++;
}
printf "%f read lines from in-memory file and do RE; n=$n.\n", time - $time;
```

On my cygwin system this prints:

```
0.122725 read lines from disk and do RE; n=570694.
27.238712 read lines from in-memory file and do RE; n=570694.
```
So the in-memory file is about 300 times slower.

**Expected behavior**
I would expect the times to be roughly in the same ball park.

**Perl configuration**

```
Summary of my perl5 (revision 5 version 36 subversion 3) configuration:

  Platform:
    osname=cygwin
    osvers=3.4.10-1.x86_64
    archname=x86_64-cygwin-threads-multi
    uname='cygwin_nt-10.0-22631 walter 3.4.10-1.x86_64 2023-11-29 12:12 utc x86_64 cygwin '
[...]

Issue #20206, opened 2022-08-31
pp_subst + COW leaks memory


richardleach:
**Description**
Since v5.19.1, perl has leaked memory when built with COW and a successful match occurs.

**Steps to Reproduce**
The following is based upon https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/regexredux-perl-1.html and takes the same input of FASTA string. Comments show the approximate memory usage at each step.

```
# approx 50 MB input file.

my $seq = do { local $/; <STDIN> };
# 5.16.3         approx 102 MB RAM used at this point
# 5.18.0         approx 102 MB RAM used at this point
# 5.19.1 & later approx 55 MB RAM used at this point

$seq =~ s/>.*\n//g;
# 5.16.3         102 MB used
# 5.18.4:        102 MB used
# 5.19.1 & later 102 MB used

$seq =~ s/t/<4>/g;
# 5.16.3         175 MB used
# 5.18.4         175 MB used
# 5.19.1 & later 176 MB used

$seq =~ s/DS/<3>/g;
# 5.16.3         198 MB used
# 5.18.4         198 MB used
# 5.19.1 & later 248 MB used

$seq =~ s/NS/<2>/g;
# 5.16.3         197 MB used
# 5.18.4         197 MB used
# 5.19.1 & later 321 MB used

$seq =~ s/2/1/g;
# 5.16.3         197 MB used
# 5.18.4         197 MB used
# 5.19.1 & later 395 MB used

$seq =~ s/3/1/g;
# etc etc
```
**Expected behavior**
Memory use reaches a steady level (e.g. 100 - 200 MB here), rather than going up by approx the size of the input string with each successful match.

**Perl configuration**
Any perl since v5.19.1 with COW enabled.

richardleach:
Stepping through `pp_subst` with the script above:
* `TARG` is never `SvIsCOW` at the start of `pp_subst` (`was_cow` == 0)
* `force_on_match` == 0
* `r_flags` == 1
* `CALLREGEXEC(rx, orig, strend, orig, 0, TARG, NULL, r_flags)` returns true. This function also sets `SvIsCOW_on(TARG)`, and the cow ref count ==1
* `dstr` does not get upgraded
* Some of the subst patterns start down the in-place branch, but immediately jump to the `got_a_cow` label in the not-in-place branch, so basically the same path is followed for all patterns in the script above.
* The `do` blocks occur, doing the actual substitutions (I think, didn't look too closely at that bit)
[...]

PR #22120, opened 2024-04-04
sv_setsv_cow: only succeed if sv_setsv() would also COW

tonycoz:
This improved performance of the test case I was using from:

```
real    1m7.690s
user    0m2.593s
sys     1m4.984s
```
to
```
real    0m0.576s
user    0m0.453s
sys     0m0.093s
```
on cygwin.

The problem that is fixed here is that sv_setsv_cow() would COW SVs even though they had a very short string in a very large buffer, which had some sort of unclear to me interaction with the Win32 virtual memory system that had a painful performance impact when the large SV was made COW.

It may also improve memory handling on non-Win32 systems, since the SV was made COW, other SV copying functions wouldn't check if it was suitable for COW, so further copies would retain a COW reference for the large buffer.  In the example test case from #21877 if we pushed the SV to an array after a successful match memory use would balloon out, running even a system with a generous amount of memory, out of memory.

Fixes part of #21877

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -5190,6 +5190,7 @@ ext/XS-APItest/t/sv_numeq.t			Test sv_numeq
 ext/XS-APItest/t/sv_streq.t			Test sv_streq
 ext/XS-APItest/t/svcat.t			Test sv_catpvn
 ext/XS-APItest/t/svcatpvf.t			Test sv_catpvf argument reordering
+ext/XS-APItest/t/svcow.t			Test COW
 ext/XS-APItest/t/sviscow.t			Test SvIsCOW
 ext/XS-APItest/t/svpeek.t			XS::APItest extension
 ext/XS-APItest/t/svpv.t				More generic SvPVbyte and SvPVutf8 tests
--- a/embed.fnc
+++ b/embed.fnc
@@ -4144,7 +4144,7 @@ RTp	|MEM_SIZE|malloc_good_size					\
 #endif
 #if defined(PERL_ANY_COW)
 : Used in regexec.c
-EXpx	|bool	|sv_setsv_cow	|NN SV **pdsv				\
+EXdpx	|bool	|sv_setsv_cow	|NN SV **pdsv				\
 				|NN SV *ssv
 #endif
 #if defined(PERL_CORE)
--- a/embed.fnc
+++ b/embed.fnc
@@ -4144,7 +4144,7 @@ RTp	|MEM_SIZE|malloc_good_size					\
 #endif
 #if defined(PERL_ANY_COW)
 : Used in regexec.c
-EXpx	|SV *	|sv_setsv_cow	|NULLOK SV *dsv 			\
+EXpx	|bool	|sv_setsv_cow	|NN SV **pdsv				\
 				|NN SV *ssv
 #endif
 #if defined(PERL_CORE)
--- a/ext/XS-APItest/APItest.xs
+++ b/ext/XS-APItest/APItest.xs
@@ -3159,6 +3159,17 @@ bool
 sv_setsv_cow_hashkey_notcore()
 
 void
+sv_grow(SV *sv, UV len)
+    CODE:
+        sv_force_normal(sv);
+        SvGROW(sv, len);
+
+void
+sv_force_normal(SV *sv)
+    CODE:
+        sv_force_normal(sv);
+
+void
 sv_set_deref(SV *sv, SV *sv2, int which)
     CODE:
     {
--- a/ext/XS-APItest/t/svcow.t
+++ b/ext/XS-APItest/t/svcow.t
@@ -1,0 +1,46 @@
+#!perl
+use strict;
+use warnings;
+use XS::APItest;
+use B;
+
+use Test::More tests => 11;
+
+{
+    # github #21877
+    # the regexp engine would COW an SV that had a large
+    # SvLEN() in cases where sv_setsv() wouldn't.
+    # This led to some surprises.
+    # - On cywgin this produced some strange performance problems
+    # - In general it meant the (large) buffer of the SV remained
+    #   allocated for longer than it otherwise would.
+    # Also, since the SV became CoW, further copies would also
+    # be CoW, for example, code like:
+    #
+    # while (<>) { # sv_getsv() currently allocates a large-ish buffer
+    #    /regex that (captures)/; # CoW large buffer
+    #    push @save, $_; # copy in @save still has that large buffer
+    # }
+    my $x = "Something\n" x 1000;
+    cmp_ok(length $x, '>=', 1250,
+           "need to be at least 1250 to be COWed");
+    sv_grow($x, 1_000_000);
+    my $ref = B::svref_2object(\$x);
+    cmp_ok($ref->LEN, '>=', 1_000_000,
+           "check we got it longer");
+    ok(!SvIsCOW($x), "not cow before");
+    is($ref->REFCNT, 1, "expected reference count");
+    ok($x =~ /me(.)hing/, "match");
+    ok(!SvIsCOW($x), "not cow after");
+
+    # make sure reasonable SVs are COWed
+    my $y = "Something\n" x 1000;
+    sv_force_normal($y);
+    cmp_ok(length $y, '>=', 1250,
+           "need to be at least 1250 to be COWed");
+    my $ref2 = B::svref_2object(\$y);
+    ok(!SvIsCOW($y), "not cow before");
+    is($ref2->REFCNT, 1, "expected reference count");
+    ok($y =~ /me(.)hing/, "match");
+    ok(SvIsCOW($y), "is cow after");
+}
--- a/proto.h
+++ b/proto.h
@@ -6047,10 +6047,10 @@ Perl_uvuni_to_utf8(pTHX_ U8 *d, UV uv)
 
 #endif /* !defined(NO_MATHOMS) */
 #if defined(PERL_ANY_COW)
-PERL_CALLCONV SV *
-Perl_sv_setsv_cow(pTHX_ SV *dsv, SV *ssv);
+PERL_CALLCONV bool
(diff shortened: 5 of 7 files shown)


iabyn (commented):
Looks plausible. It would be nice for the sv_setsv_cow() fn  to have a few lines of code comments at the top explaining what it's for/does.

jkeenan:
@tonycoz this p.r. was filed back in April, but is still in draft status and has acquired merge conflicts.  Assuming you want to move forward with it, could you resolve those conflicts and un-draft the ticket?  Thanks.

tonycoz:
I'll rebase it, I'm looking into some behaviour I didn't expect with a variant of this change.

bulk88:
@tonycoz 

https://github.com/Perl/perl5/issues/22623

Your performance problem is related to the ticket above, and HeapReAlloc() either refuses to shrink the memory allocation as a MS perf optimization, or all "freed" memory by HeapReAlloc() from large (any alloc too large for Low Fragmention Heap), is donated to the Low Fragmentation Heap. HeapReAlloc() does a ```VirtualProtect(ptr, len, MEM_RESERVE);```, then puts ```ptr``` into a linked list in Low Fragmentation Heap. ```VirtualProtect(ptr, len, MEM_RESERVE);``` makes the memory counts for the process go down in task manager, for "Private Working Set" and "Commit Size" but you will eventually have problems with running out of virtual address space because of 100s MBs or 10s of GBs of memory marked as "reserved".

I agree with logic, is SvCUR vs SvLEN gap is too big at some threshold, do not COW. Or shrink the PV buffer in the parent SV * (maybe a very bad perf idea for PAD vars).

edit:

very little can be found on this bug, and im using win7,  but look at this 2011 post, nobody in that thread knows why the code exists, but someone in the mid 1990s knew the secret bug/problem but it was lost to time


https://forum.powerbasic.com/forum/user-to-user-discussions/powerbasic-for-windows/48415-refresh-heap?p=569065#post569065

bulk88:
> very little can be found on this bug, and im using win7, but look at this 2011 post, nobody in that thread knows why the code exists, but someone in the mid 1990s knew the secret bug/problem but it was lost to time
> 
> https://forum.powerbasic.com/forum/user-to-user-discussions/powerbasic-for-windows/48415-refresh-heap?p=569065#post569065


https://stackoverflow.com/questions/9164837/realloc-does-not-correctly-free-memory-in-windows

Above shows the bug with Windows architecture. I found the correct explanation, NT Kernel has no function called ```VirtualReAlloc```. Only 2 functions exist, ```VirtualAlloc``` and ```VirtualFree```. A kernel memory object, once it is born, it has 2 fields. ```start_address``` and ```length```. They are constant for the rest of that kernel object's lifespan. A user can use ```VirtualProtect``` to change single 4096 byte units between ```start_address``` and ```start_address+length```. The changes are RWX flags and decommit (discard data, SEGV ON, all process memory counters instantly drop), and turning on any of the 3 RWX flags does "SEGV OFF".  ```VirtualUnlock``` is a command to "move from phy ram to paging file now and delete from phy ram, next RWX will read back from paging file . 

The only way to shrink the reserved address space of kernel memory handle is to VirtualAlloc a new kernel memory object, then call ```memcpy()```, then call ```VirtualFree``` on the old kernel memory object. That is a pathological perf degrade, therefore ```HeapReAlloc``` will not do that. Now how to fix that in perl when msvcrt/libc ```realloc``` does not release virtual address space, and instead only deletes memory contents from phy ram and from paging file and switch the address to SEGV ON?

Side trivia ```VirtualLock``` is obsolete/noop now. It used to mean pin 4096 byte unit in phy ram until further notice. MS either made the command always fail and return an error code, or it silently succeeds but does nothing. ```VirtualLock``` was a security problem from user mode. Its only purpose was to talk to hardware I/O devices (Windows 95), but that is privilege for kernel mode drivers. If it makes sense, the kernel mode driver can take a 4096 unit of user mode phy ram, and make it unpage/pin in phy ram forever or however long is needed for hardware DMA/MMIO.

tonycoz:
> I'll rebase it, I'm looking into some behaviour I didn't expect with a variant of this change.

The behaviour I was concerned about was a leak if sv_setsv_cow() didn't cow copy an already COW SV, I eventually tracked this down to pp_subst, which seems to assume force_normal() will always free the buffer of a CoW SV: it doesn't, if the SV is the final owner the SV retains the buffer and CoW is turned off.

Fixing that leak (freeing the memory in pp_subst) lead to a use after free in the regular expression engine that I'll be tracking down next.

tonycoz:
The problem I ran into was #20206, if sv_setsv_cow() didn't CoW copy a previously CoW SV the [code from pp_subst](https://github.com/Perl/perl5/issues/20206#issuecomment-1233402714) would receive a CoW SV with only a single refcount, and that code doesn't correctly handle that case.

Unfortunately fixing that results in use after free in code like:
```
m!someregexp (?{ s/xxx/yyy/ })!
```
(this failed in a test, but I've lost of track of where the test was)

which I believe is a known problem in some cases with the regexp engine not updating its pointers to match changes in the underlying SV.

bulk88:
> The problem I ran into was #20206, if sv_setsv_cow() didn't CoW copy a previously CoW SV the [code from pp_subst](https://github.com/Perl/perl5/issues/20206#issuecomment-1233402714) would receive a CoW SV with only a single refcount, and that code doesn't correctly handle that case.

A max 255 COW SVPV with a refcount of 1 shouldn't ever exist in the first place. Whatever RC--ed RC # 2  down to RC # 1 on the PV buffer shouldve stripped the SV head flags saying this SVPV is a COW, and shouldve `= '\0';`-ed the RC byte at the end of the PV buffer.

Someone correct me if this thinking is wrong.

tonycoz:
> Someone correct me if this thinking is wrong.

The problem is if you have two SVs pointing at the same PV, if sv1 releases its reference to the PV, sv1 gets marked as non-COW, but it has no way to mark sv2 as non-COW, so you're left with a COW PV with a reference count of 1.

bulk88:
> > Someone correct me if this thinking is wrong.
> 
> The problem is if you have two SVs pointing at the same PV, if sv1 releases its reference to the PV, sv1 gets marked as non-COW, but it has no way to mark sv2 as non-COW, so you're left with a COW PV with a reference count of 1.

Can't some random location inside libperl `SV_CHECK_THINKFIRST_COW_DROP(sv2);`, that will execute in the near future realize `sv2` is a `SvISCOW_on()` but `COWREFCNT(sv2) == 0` and quickly remove SV head flag 'SvISCOW_on()`? And by doing that `sv2` is now a non-any-COW, no-COW-255, plain old `Newx()` block that can be `Renew()`ed or PAD_STEAL ed or MORTAL_STEAL ed ?

tonycoz:
> > The problem is if you have two SVs pointing at the same PV, if sv1 releases its reference to the PV, sv1 gets marked as non-COW, but it has no way to mark sv2 as non-COW, so you're left with a COW PV with a reference count of 1.
> 
> Can't some random location inside libperl `SV_CHECK_THINKFIRST_COW_DROP(sv2);`, that will execute in the near future realize `sv2` is a `SvISCOW_on()` but `COWREFCNT(sv2) == 0` and quickly remove SV head flag 'SvISCOW_on()`? And by doing that `sv2`is now a non-any-COW, no-COW-255, plain old`Newx()`block that can be`Renew()`ed or PAD_STEAL ed or MORTAL_STEAL ed ?

How does the modification of sv1 find sv2?

What triggers this  `SV_CHECK_THINKFIRST_COW_DROP(sv2);`?

Or are you thinking every COW SV is periodically checked to see if its PV has more than one reference?

But there's no real need, S_sv_uncow() will just turn the flag off if something needs to modify the PV and the SV has the only reference to it.

bulk88:
> > > The problem is if you have two SVs pointing at the same PV, if sv1 releases its reference to the PV, sv1 gets marked as non-COW, but it has no way to mark sv2 as non-COW, so you're left with a COW PV with a reference count of 1.
> > 
> > 
> > Can't some random location inside libperl `SV_CHECK_THINKFIRST_COW_DROP(sv2);`, that will execute in the near future realize `sv2` is a `SvISCOW_on()` but `COWREFCNT(sv2) == 0` and quickly remove SV head flag 'SvISCOW_on()`? And by doing that `sv2`is now a non-any-COW, no-COW-255, plain old`Newx()`block that can be`Renew()`ed or PAD_STEAL ed or MORTAL_STEAL ed ?
> 
> How does the modification of sv1 find sv2?

It can't and doesn't need to. When sv1 RC-- and discards the COW PVX, sv2 is on its own.
> 
> What triggers this `SV_CHECK_THINKFIRST_COW_DROP(sv2);`?

Natural runloop code. sv_catpvn()/etc.  `sv2 as non-COW, so you're left with a COW PV with a reference count of 1.` sv2 logically already is a non-COW if RC is 1. Its just a question of how big or small or slow or fast Perl wants its core macros to be.

SVfIsCOW + SvLEN() +last byte==RC==0  is logically the same as `if((flags & (SVfIsCOW |SVfPOK)) == SVfPOK) {}`, but 3x the work/3x the branches to find the answer.

> Or are you thinking every COW SV is periodically checked to see if its PV has more than one reference?

Probably. A no COW Newx() buf can always upgrade to a COW Newx() RC == 1 buf, and back down to a no COW Newx() buf at any time no side effects.

> But there's no real need, S_sv_uncow() will just turn the flag off if something needs to modify the PV and the SV has the only reference to it.

Adding a variant of `S_sv_uncow()`/`SV_CHECK_THINKFIRST_COW_DROP(sv2);` that can RECOVER the Newx() block if RC == 1 is needed badly, along with a `SV_CHECK_THINKFIRST_COW_GROW_DROP(sv2, extra_len);` for `sv_catpvn()` and `sv_catpvf()` to skip the 2nd `realloc()` call if the `HEK*` COW or `SVs_STATIC` COW or COW 255 had RC == 2 or RC == Inf (aka `SVs_STATIC` and `HEK*` COWs).

